PR #1271, opened 2025-08-28
Fix DNS migration to handle UseDNS=no in network files

jardahrazdera:
Fixes #1246 

## Problem
Migration 1755109182.sh only fixes resolved.conf but doesn't remove `UseDNS=no` from network files. When `UseDNS=no` persists in network configs, it blocks DHCP DNS and forces fallback to Cloudflare after reboot.

## Solution
Remove `UseDNS=no` from all network files in the migration, matching what `omarchy-setup-dns DHCP` does.

## Testing
- Reproduced the issue by setting Cloudflare DNS + UseDNS=no in network files
- Confirmed DNS gets stuck on Cloudflare (1.1.1.1)  
- Verified fixed migration removes UseDNS=no and restores DHCP DNS (router IP)
- DNS switches back to router DHCP correctly after reboot

PR #1314, opened 2025-08-30
Fix DNS migration to handle UseDNS=no in network files

### LINKED ISSUE #1246: DNS provider script does not persist DHCP preference after reboot
First off, thank you for the work on Omarchy - the included tooling makes setup much smoother.

I’ve noticed an issue with the DNS provider script (`/usr/bin/.`.. - the one that prompts with gum choose between Cloudflare, DHCP, or Custom). On my system, the script does not persist the DHCP choice across reboots.

### Steps to reproduce
1. Run the script manually and choose DHCP.
2. Confirm resolvectl status shows my LAN DNS (from DHCP) in use.
3. Reboot.
4. After boot, DNS has reverted to Cloudflare, and I have to re-run the script and choose DHCP again.

### Expected behavior
Selecting DHCP should persist across reboots and updates.
My system should consistently use the DNS servers provided by my network’s DHCP without requiring manual intervention.

### Actual behavior
After reboot, DNS settings revert to Cloudflare.
Running the script again and choosing DHCP restores expected behavior until the next reboot.

### Notes
I noticed the script adds or removes UseDNS=no under [DHCPv4] in .network files depending on the selection.
Even if I lock `/etc/systemd/resolved.conf` with `chattr +i`, I still need to run the script after every reboot to re-enable DHCP DNS.
This makes sense on a laptop that roams between networks, but for a desktop on a fixed LAN it causes unnecessary breakage (I rely on local DNS. My security policy dictates local DNS).

### Suggested fix
Make the script’s choice persist by default (maybe using drop-in .conf overrides instead of rewriting `/etc/systemd/resolved.conf`).
Or, provide a way to set DHCP as a permanent default so it isn’t reset to Cloudflare on reboot/update.

jardahrazdera:
Fixes #1246

Reopening the DNS fix since the previous PR #1271 got closed accidentally.

Migration 1755109182.sh only fixed resolved.conf but didn't remove UseDNS=no from network files, causing DNS to revert to Cloudflare after reboot.

This adds:
- Fix to migration 1755109182.sh to remove UseDNS=no from network files
- New migration 1756491748.sh for users who already ran the old version

Both now match what omarchy-setup-dns DHCP does.

Tested and confirmed it fixes the issue.

### DIFF
--- a/migrations/1756491748.sh
+++ b/migrations/1756491748.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+
+# Migration to fix DNS issue for existing users who already ran 1755109182.sh
+# Remove UseDNS=no from all network files to allow DHCP DNS
+# This matches what omarchy-setup-dns DHCP does
+
+echo "Removing UseDNS=no from network files to fix DNS issue..."
+
+for file in /etc/systemd/network/*.network; do
+  [[ -f "$file" ]] || continue
+  if grep -q "^UseDNS=no" "$file"; then
+    echo "Removing UseDNS=no from $file"
+    sudo sed -i '/^UseDNS=no/d' "$file"
+  fi
+done
+
+echo "DNS migration completed - DHCP DNS should work after reboot"
--- a/migrations/1756491748.sh
+++ b/migrations/1756491748.sh
@@ -1,17 +1,6 @@
-#!/bin/bash
-
-# Migration to fix DNS issue for existing users who already ran 1755109182.sh
-# Remove UseDNS=no from all network files to allow DHCP DNS
-# This matches what omarchy-setup-dns DHCP does
-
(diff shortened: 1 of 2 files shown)


dhh:
Let's just combine the entire cleanup in the new migration. Just wanted to catch that some folks will have run the old one. But once you add a new name, you'll get everyone.

migrations/1756491748.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+# Migration to fix DNS issue for existing users who already ran 1755109182.sh
+# Remove UseDNS=no from all network files to allow DHCP DNS
+# This matches what omarchy-setup-dns DHCP does
```
dhh: Remove both bash and comments. Style is just a single echo at the top. No need for other echos during progress either unless things fail.

PR #881, opened 2025-08-18
Change Ollama install from AUR package to official website

valbertoenoc:
It seems that the official Ollama AUR package does not support using the GPU out of the box. When I ran into this issue, it was solved by installing from the official website command. So I added an option to install software from the System Menu with a custom command instead of defaulting to `yay`.

PR #1319, opened 2025-08-30
Ensure that Ollama can run on NVIDIA GPUs

roberto-aguilar:
When installing Ollama via the `Install > AI > Ollama` menu, I noticed that language models defaulted to running on the CPU instead of the GPU.

This happens because Ollama requires the `ollama-cuda` package to enable CUDA support for NVIDIA GPUs.

Additionally, I’ve extracted the installation logic into a dedicated `omarchy-install-ollama` script. This keeps the menu code simpler and easier to maintain.

### DIFF
--- a/bin/omarchy-install-ollama
+++ b/bin/omarchy-install-ollama
@@ -1,0 +1,10 @@
+#!/bin/bash
+
+echo -e "Installing Ollama...\n"
+sudo pacman -S --noconfirm ollama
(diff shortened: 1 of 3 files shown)


fals:
would be nice to merge this,  LLMs without GPU if possible should be avoided.

dhh:
Duplicate of #881

PR #1320, opened 2025-08-30
Rename install to add in walker menu

kodnin:
This PR renames the _Install_ entry in the `walker` menu to _Add_. It's about linguistic consistency. _Add_ is a better complement to _Remove_. Alternatively _Uninstall_ could be a better complement to _Install_. Some screenshots below, because it matters how something looks and feels.

Before:

screenshot-2025-08-30_09-34-54

After:

screenshot-2025-08-30_09-35-28

### DIFF
--- a/bin/omarchy-menu
+++ b/bin/omarchy-menu
@@ -184,7 +184,7 @@ show_setup_config_menu() {
 }
 
 show_install_menu() {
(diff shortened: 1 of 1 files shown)


naxels:
I agree with your thinking, however personally I like Install / Uninstall more than Add / Remove in terms of Software and packages

abenz1267:
One could even think about just creating a submenu for the 3 items (install,remove,update) to create more context.

dhh:
It's "Remove" because it also includes stuff like fingerprint sensor and fido2 key that comes from Setup. I like what we have now. Install / Setup and then Remove to do from either. Appreciate the suggestion though!

PR #1321, opened 2025-08-30
new wallpaper for catppuccin

inffy:
This was requested in discord as a new wallpaper


dhh:
I think this is quite nice, but on second review, we already have 3 backgrounds for catppuccin, and this one also isn't quite high resolution enough. So going to skip on it as a default for now. Thanks for the suggestion, though!

PR #1323, opened 2025-08-30
chore: Add Emacs to editor install menu

spdawson:
## Synopsis

As well as installing the `emacs-wayland` package, we enable and start the `emacs` systemd service. A GUI Emacs client can then be started using e.g. `emacsclient -c`.

### DIFF
--- a/bin/omarchy-menu
+++ b/bin/omarchy-menu
@@ -208,12 +208,13 @@ show_install_service_menu() {
 }
 
 show_install_editor_menu() {
-  case $(menu "Install" "  VSCode\n  Cursor [AUR]\n  Zed\n  Sublime Text\n  Helix") in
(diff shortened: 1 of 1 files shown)


JordanAnthonyKing:
Since Omarchy default ships with lazyvim, perhaps the emacs install should default to Doom?

spdawson:
> Since Omarchy default ships with lazyvim, perhaps the emacs install should default to Doom?

I don't think so. We could add Doom Emacs as a separate editor option, but I don't think there is anything like a majority consensus that Doom is the "standard" or "best" way to run Emacs.

JordanAnthonyKing:
> I don't think there is anything like a majority consensus that Doom is the "standard" or "best" way to run Emacs.

Nor is there for lazyvim (^:

PR #1324, opened 2025-08-30
Add private browser key binding

kodnin:
This PR adds a private (window) browser keybinding available under `SUPER SHIFT + B` as an alternative to the normal (window) browser keybinding of `SUPER + B`. It's setup to be compatible with other default web browsers like Firefox. It's easy to extend it for more browsers.

To test you can (for example) install Firefox and use `xdg-settings set default-web-browser firefox.desktop`. Then use `SUPER + B` and `SUPER SHIFT + B` to see it in action.

screenshot-2025-08-30_11-53-00

As a developer I often use a private window for testing (Rails :wink:) web apps. And occasionally to keep my (normal) browsing history clean.

### DIFF
--- a/bin/omarchy-launch-browser
+++ b/bin/omarchy-launch-browser
@@ -1,3 +1,19 @@
 #!/bin/bash
 
-exec setsid uwsm app -- $(sed -n 's/^Exec=\([^ ]*\).*/\1/p' {~/.local,~/.nix-profile,/usr}/share/applications/$(xdg-settings get default-web-browser) 2>/dev/null | head -1) ${args[@]} $@
+browser_exec=$(sed -n 's/^Exec=\([^ ]*\).*/\1/p' {~/.local,~/.nix-profile,/usr}/share/applications/$(xdg-settings get default-web-browser) 2>/dev/null | head -1)
+
+case "$browser_exec" in
+*firefox*)
+  private_flag="--private-window"
+  ;;
+*)
+  private_flag="--incognito"
+  ;;
+esac
+
+args=("$@")
+for i in "${!args[@]}"; do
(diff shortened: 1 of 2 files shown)


abenz1267:
Does this really need a keybind? why not open a private window via launcher?

bin/omarchy-launch-browser
```diff
@@ -1,3 +1,19 @@
 #!/bin/bash
 
-exec setsid uwsm app -- $(sed -n 's/^Exec=\([^ ]*\).*/\1/p' {~/.local,~/.nix-profile,/usr}/share/applications/$(xdg-settings get default-web-browser) 2>/dev/null | head -1) ${args[@]} $@
+browser_exec=$(sed -n 's/^Exec=\([^ ]*\).*/\1/p' {~/.local,~/.nix-profile,/usr}/share/applications/$(xdg-settings get default-web-browser) 2>/dev/null | head -1)
+
+case "$browser_exec" in
+*firefox*)
```
dhh: Should expand to include all the firefox-based browsers we also cover in windows.conf: windowrule = tag +firefox-based-browser, class:(Firefox|zen|librewolf)

dhh:
I like this. Often need a private browser for separate cookie testing. Natural hotkey to have it on. Just need to expand the firefox-like browser capture.

kodnin:
@dhh, thanks for the feedback! I've included the other Firefox-based browsers.

Please keep in mind I've only tested this with Chromium and Firefox. I couldn't install Zen or Librewolf because of AUR issues.

kodnin:
Now AUR works again I've also tested it with Zen and it works! :relaxed:

PR #1330, opened 2025-08-30
feat: walker vim motions

joaomendoncaa:
we should start making an effort to configure things with vim motions in mind since this is a distribution heavily skewed towards developers


guilhermetk:
There is already a pending PR for that: https://github.com/basecamp/omarchy/pull/489

PR #1343, opened 2025-08-31
Add keyd keyboard remapping for sysrq shortcut on Mac mode keyboards

acrogenesis:
This PR adds support for the `keyd` keyboard remapping daemon to fix print screen functionality on Mac mode keyboards by mapping `leftmeta+leftshift+4` to the `sysrq` key.
You **can** still use `left super + left shift + 4` to move a window to the 4th workspace.

### Changes Made

**New Files:**
- keyd.sh - Hardware configuration script that:
  - Installs the `keyd` package from AUR
  - Creates `/etc/keyd/default.conf` with keyboard mappings
  - Enables and starts the keyd service

- 1756583646.sh - Migration script that sources the hardware config for automatic setup

### Why This Change?
- **Mac Mode Keyboards**: Many Mac-style keyboards like my Keychron K3 send `leftmeta+leftshift+4` at the firmware level to simulate a print screen for mac instead of the button being mapped to a true print screen.
- **Universal Compatibility**: Uses `default.conf` without device-specific IDs, so it works with any keyboard.

### DIFF
--- a/install/config/hardware/keyd.sh
+++ b/install/config/hardware/keyd.sh
@@ -1,0 +1,27 @@
+#!/bin/bash
+
+# Install keyd and configure Keychron K3 keyboard remapping
+
+# Install keyd if not already installed
+if ! yay -Qq keyd &>/dev/null; then
+  echo "Installing keyd (keyboard remapping daemon)"
+  yay -S --noconfirm --needed keyd || true
+fi
+
+# Create keyd config directory
+sudo mkdir -p /etc/keyd
+
+# Create keyd config file
+if [[ ! -f /etc/keyd/default.conf ]]; then
+  echo "Creating keyd config for keyboard remapping"
+  cat <<EOF | sudo tee /etc/keyd/default.conf >/dev/null
+[main]
+leftmeta+leftshift+4 = sysrq
+EOF
+fi
+
+# Enable and start keyd service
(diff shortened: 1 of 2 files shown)


doctorguile:
keyd is powerful but to achieve what you want i think its more straight forward to just add this to

~/.config/hypr/bindings.conf

`
bindd = SUPER ALT SHIFT, 4, Screenshot of region, exec, omarchy-cmd-screenshot
`

Although I appreciate your post as I was trying to find a way to enable sticky shift key 

keyd + the following config did the trick for me

`
shift = oneshot(shift)
`

acrogenesis:
Hi @doctorguile, I recently changed keyboards to the flow84 and tried your proposed method and it doesn't work.
I believe it has something to do with how the keyboards send the keypress automatically. 

In the new commit 75cdeff I added `leftmeta+leftshift+3 = sysrq` which covers this keyboard "mac" print screen

dhh:
If you're on the Flow84, you just need to hit FN + N to be in Windows mode. I'm almost certain that this is possible with all the Keychrons too.

acrogenesis:
Yeah that works too but the super key moves from the command/alt to option/windows button which feels less natural for me

dhh:
Appreciate the suggestion for this, but I think it's too much for the default installation.

Issue #1269, opened 2025-08-28
Screensaver activates while running games


JordanAnthonyKing:
When playing games with a controller the screensaver will activate as no mouse/keyboard input has been detected.

Specifically I'm using the dolphin emulator, I'll update if I find this is more widespread.

JordanAnthonyKing:
The lock screen does too

alexperreault:
This would be a problem upstream, with Hypridle

Tho I find it odd that it doesn't detect dolphin.

I'm pretty sure someone had this issue before you, you should search for Hypridle + dolphin

JordanAnthonyKing:
> This would be a problem upstream, with Hypridle

Looking through hypridle issues looks like it’s an issue with our config:
https://github.com/hyprwm/hypridle/issues/53

I guess we need to add a .* rule for anything fullscreen to inhibit idle?

JordanAnthonyKing:
Well I tried these two solutions to be a bit more intelligent about not activating the lock/screensaver unless the thing that was fullscreen was the screensaver itself but neither seemed to work. I'm using the plain `.*` rule for now. If anyone can show me what I had wrong that'd be greatly appreciated.
```
# Prevent screensaver/lock while running apps fullscreen
windowrule = idleinhibit fullscreen, class:.*
# windowrule = idleinhibit fullscreen, class:^(?!Screensaver$).*
# windowrule = idleinhibit fullscreen, class:^(?!.*Screensaver).*$
```

dhh:
Solved via #1303.

lukepighetti:
still having this issue on omarchy `4.0.2-1` 

clanker added this line to `~/.config/hypr/hyprland.lua`

```lua
-- Keep the screensaver/lock from kicking in while gaming: inhibit idle for
-- any window that goes fullscreen (covers Steam games, emulators, etc. that
-- Omarchy doesn't already special-case).
o.window({ fullscreen = true }, { idle_inhibit = "fullscreen" })
```
which does work when playing skyrim full screen but obviously would cause issues if we left any fullscreen app visible. clanker found the window name and offered this more tailored solution


```bash
# move steam to desktop 2
# launch skyrim
# SUPER+1 to go back to your terminal
# list all client class and title
> hyprctl clients -j | jq -r '.[] | "\(.initialClass) | \(.initialTitle)"'
steam | Steam
steam_app_72850 | Skyrim # <<< use this one
foot | foot
chromium | New Tab - Chromium
[...]

PR #1348, opened 2025-08-31
Omarchy 2.1.0

dhh:
# What changed?

* Add Omarchy Mirror as the default place to get Arch core/extra/multilib packages by @dhh 
* Remove chaotic-aur as a default setup now that we have our own package repo by @dhh 
* Add omarchy-pkg-add/drop/present/missing helper commands for migrations in particular by @dhh
* Add omarchy-cmd-present/missing helper commands for migrations in particular by @dhh
* Add Install > Service > Bitwarden as an alternative to 1pw by @crbelaus
* Add Install > Editor > Emacs by @spdawson
* Add Samba drive support to file manager by @chantastic
* Fix omarchy-migrate when migrations fail so they can be skipped by @Gazler
* Fix bring back ttf-ia-writer now that the package has been sorted by @dhh + @ryanrhughes 
* Fix trailing slash in Omarchy pkg repo declarations by @dhh
* Fix battery monitor math by @pipetogrep
* Fix screensaver should pause for 2 seconds after each completed animation by @oppegard 
* Fix Install > AI > Ollama to install GPU-accelerated versions where possible by @valbertoenoc
* Fix Omarchy Update: AUR should only upgrade AUR packages by @dhh 
* Fix Alacritty adding its own opacity controls on top of Hyprland by @dhh
* Fix Kanagawa transparency being too high with default background on the terminal by @dhh
* Fix missing icons for OS Age + Uptime in About by @theamit-969 
* Fix qt5-wayland should be installed so KeepassXC, Nextcloud, Sqlite Browser, and other apps display correctly by @roelandmoors
* Fix JetBrains font installed by default should be Nerd Font version by @ankurkotwal
* Fix steam + retroarch could idle during gaming when using joypads by @JordanAnthonyKing
* Fix wrong navigation icons used by Nautilus by @pipetogrep


JordanAnthonyKing:
My fix https://github.com/basecamp/omarchy/pull/1348/commits/ed3e1f901b0f683f1a9747c565ce9f76dd37f719 probably shouldn’t be included as it prevents screensaver from transitioning to lock.
Issue discussion here: https://github.com/basecamp/omarchy/issues/1269

Gazler:
@dhh Maybe we should make a simple script for `omarchy-confirm` that defaults to `gum` but falls back to simple terminal `y/n` prompt for users that have uninstalled it? We use it in a few migrations as well as the migration script and preflight now.

Gazler:
Added a PR for omarchy-confirm here: https://github.com/basecamp/omarchy/pull/1353

dhh:
Not a bad idea, but I'd rather just conclude that gum is part of the preconditions for Omarchy. Only reason I made an exception in that trap error is that it'll run before packages are installed.

dhh:
> My fix [ed3e1f9](https://github.com/basecamp/omarchy/commit/ed3e1f901b0f683f1a9747c565ce9f76dd37f719) probably shouldn’t be included as it prevents screensaver from transitioning to lock. Issue discussion here: #1269

I shipped this instead: https://github.com/basecamp/omarchy/commit/256a27ca8e8e8851ae6786704f1ae9af6bb7ca75

JordanAnthonyKing:
> > My fix [ed3e1f9](https://github.com/basecamp/omarchy/commit/ed3e1f901b0f683f1a9747c565ce9f76dd37f719) probably shouldn’t be included as it prevents screensaver from transitioning to lock. Issue discussion here: #1269
> 
> I shipped this instead: [256a27c](https://github.com/basecamp/omarchy/commit/256a27ca8e8e8851ae6786704f1ae9af6bb7ca75)

Ah, I see. I guess though this is going to still have issues with mpv, dolphin, and basically any other fullscreen app? I’ll investigate and see if I can find a better solution.

PR #1353, opened 2025-08-31
Add omarchy-confirm script which defaults to gum

Gazler:
The omarchy-confirm script is partially API compatible with gum, which will use gum If available, otherwise it will fall back to a simple y/n prompt.


dhh:
I'd rather just depend on gum always being there. It's OK to have preconditions on packages.

PR #1354, opened 2025-08-31
Add Crystal lang to dev installs

frayzil:
# Add [Crystal lang](https://crystal-lang.org/) to dev installs.
Files changed:
[bin/omarchy-menu](https://github.com/frayzil/omarchy/blob/f1ec359a8e985b9e77fbf3c1e77a7a7fece2fdd0/bin/omarchy-menu#L262)
[bin/omarchy-install-dev-env](https://github.com/frayzil/omarchy/blob/f1ec359a8e985b9e77fbf3c1e77a7a7fece2fdd0/bin/omarchy-install-dev-env#L128)

### DIFF
--- a/bin/omarchy-install-dev-env
+++ b/bin/omarchy-install-dev-env
@@ -125,4 +125,8 @@ dotnet)
   echo -e "Installing .NET...\n"
   mise use --global dotnet@latest
   ;;
+crystal)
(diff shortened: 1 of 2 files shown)


dhh:
We can't fit every language out there onto the list, so I think we're going to have to cut the line here. Thankfully mise makes it so easy to install whatever you want!

PR #1365, opened 2025-08-31
Enable Google Signin / sync

hjanuschka:
this allows chromium to sign in with google account and sync, like u usually do in chrome!
screenshot-2025-08-31_23-35-10


krymancer:
Seems like these values come from [here](https://chromium.googlesource.com/experimental/chromium/src/+/b08bf82b0df37d15a822b478e23ce633616ed959/google_apis/google_api_keys.cc)

hjanuschka:
@Krymancer - https://stackoverflow.com/a/67459416 it came from here.
we could compile/build it with those values embedded, keeping cli flags, aligns it more with chromium, but still be able to signin/sync

OmarSkalli:
@hjanuschka I've been using this for the last week, and it's been amazing to sync my google account. I've run into the following issue however, and curious to know if there's an easy way to fix it/existing flag maybe? 

When I open a new browser window (e.g. SUPER + B) and select a profile, if there's already another browser window open with that profile, it simply redirects me to that window instead of spawning a new browser with the selected profile. This is particularly disruptive when trying to open a new window in a blank workspace as it will automatically switch workspace and focus another browser instance.

teoruiz:
@OmarSkalli I have this behaviour in `main`. I agree it's a pain but it's not related to this PR.

OmarSkalli:
> @OmarSkalli I have this behaviour in `main`. I agree it's a pain but it's not related to this PR.

Ah you're right, thanks for pointing this out. I wasn't using multiple profiles until this change and wrongly assumed it was related.

PR #1370, opened 2025-09-01
added a check that the shell is bash before running mise activate

mroboff:
In **omarchy/default/bash** and **omarchy/config/uwsm/env**, I added a check that `$SHELL == "/bin/bash"` before running `eval "$(mise activate bash)".` Without this check, Mise may not function properly when the user's interactive shell is not bash (and will absolutely not function properly if the user's shell is fish.)

The issue occurs because mise keeps its own environment variable under MISE_SHELL, which it sets to whatever is supplied as the argument to _mise activate_. 

Note: this change does not enable mise to run in other shells. (Since Omarchy assumes bash, the idea is that a user will need to manually setup mise activate in their new shell if they want to run a different shell,) for instance in _zsh.rc_ or in _.config/fish/config.fish_.

### DIFF
--- a/default/bash/init
+++ b/default/bash/init
@@ -1,4 +1,4 @@
-if command -v mise &> /dev/null && [[ "$SHELL" == "/bin/bash" ]]; then
+if command -v mise &> /dev/null; then
   eval "$(mise activate bash)"
 fi
 
--- a/default/bash/init
+++ b/default/bash/init
@@ -1,4 +1,4 @@
-if command -v mise &> /dev/null; then
+if command -v mise &> /dev/null && [[ "$SHELL" == "/bin/bash" ]]; then
   eval "$(mise activate bash)"
 fi
 
--- a/config/uwsm/env
+++ b/config/uwsm/env
@@ -2,6 +2,6 @@ export OMARCHY_PATH=$HOME/.local/share/omarchy
 export PATH=$OMARCHY_PATH/bin/:$PATH
 export TERMINAL=alacritty
 
-if command -v mise &> /dev/null; then
+if command -v mise &> /dev/null && [[ "$SHELL" == "/bin/bash" ]]; then
   eval "$(mise activate bash)"
 fi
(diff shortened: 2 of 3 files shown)


dhh:
Don't we need to do the same for starship and zoxide?

mroboff:
> Don't we need to do the same for starship and zoxide?

I don't think so. Having checked both, it appears only mise breaks as described above. 

Having starship and zoxide load into bash, even when a user's primary shell is not bash, is useful for when they jump into bash from their primary shell. (For example, when debugging shell scripts or when trying to run a complex bash command that they don't want to convert by hand.) 

In testing, when opening a bash shell from fish, both zoxide and starship load automatically and operate as expected. Similarly, I can load zoxide or starship into fish without either breaking or behaving abnormally. Therefore, I recommend making the change only to mise, which breaks when the shell is not bash.  
 
Again, this is under the assumption that Omarchy intends to only configure and support bash "out of the box." Otherwise, you'll want to add elifs to mise, starship, and zoxicide to invoke their initialization for other popular shells (namely zsh and fish.)

dhh:
The config change is going to need a migration to catch folks up.

default/bash/init
```diff
@@ -1,4 +1,4 @@
-if command -v mise &> /dev/null; then
+if command -v mise &> /dev/null && [[ "$SHELL" == "/bin/bash" ]]; then
```
andresatierf: Here seems a bit redundant because this file is called by .bashrc

dhh: Agree. I think only the uwsm makes sense. If you're switching shell, you shouldn't be running the bash scripts.

mroboff: agreed

mroboff:
> The config change is going to need a migration to catch folks up.

Done.

dhh:
Hmm, on second thought, you need to do this mise activation anyway. This env lives in config land, which is yours to edit. I think maybe the better path is actually just to document that this needs updating if you switch shell?

dhh:
Yeah, I think that's the way to go. Appreciate you exploring this!

PR #1156, opened 2025-08-27
View the release notes of a current or past release from the terminal

ardavis:
After updating Omarchy sometimes I thought it could be useful to view the release notes from the terminal.

image

Also supports a specific version to be passed in.

image

This uses the 'glow' package to display the markdown content.

PR #1374, opened 2025-09-01
On update-icon right-click open tags list on github

jheuing:
When updating omarchy, I am curious on what has been updated. This merge request adds a right click on the update icon, which will open the latest tag on github.

### DIFF
--- a/bin/omarchy-update-notes
+++ b/bin/omarchy-update-notes
@@ -1,0 +1,22 @@
+#!/bin/bash
+
+# Get remote tag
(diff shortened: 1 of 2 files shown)


ardavis:
This is a cool idea! It's quite related to https://github.com/basecamp/omarchy/pull/1156 but I don't see why we couldn't have both.

1156 will have it so when you run update it'll show you the release notes and then you confirm whether you'd like to update.

jheuing:
#1156 is also great @ardavis ! I can see both, too. We'll see if we have common ground here and this gets merged or not :-)

dhh:
We now have a link to the release notes in the notice before the update confirmation. I think that's enough and more discoverable. But appreciate the suggestion!

PR #1380, opened 2025-09-01
Add locale to the waybar clock format

arni1981:
Icelandic was not being picked up by the clock on waybar, this issue contains more information.

https://github.com/Alexays/Waybar/issues/2694#issuecomment-1837118050

### DIFF
--- a/config/waybar/config.jsonc
+++ b/config/waybar/config.jsonc
@@ -57,8 +57,8 @@
     "on-click": "alacritty -e btop"
   },
   "clock": {
-    "format": "{:%A %H:%M}",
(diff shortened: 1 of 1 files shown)


dhh:
actually, could you add a migration that uses grep/sed to change the existing .config/waybar/ file to match the new default too? Thanks!

rez1-dev:
@dhh you should add `:L` in the FAQ too

https://learn.omacom.io/2/the-omarchy-manual/67/faq#how-do-i-change-the-clock-format-to-12-hour

dhh:
We are just making :L the default, since it doesn't change anything for English. So don't need to document it.

PR #1385, opened 2025-09-01
Install opencode using AUR

d4vsanchez:
The `opencode-bin` package doesn't exist in Arch official repositories. It's using the name of the package in AUR.

In the current version, the package can't be found:
image

With this change applied the package will try to get installed:
image


arafays:
lgtm 👍🏽

dhh:
We have the `opencode` package in the OPR. Fixed to use the right name. Thanks!

PR #1400, opened 2025-09-01
Add cursor selection menu

CobyPowers:
This commit features a new style menu that allows the user to change their cursor theme.

Three new scripts were created:
- omarchy-cursor-set (sets cursor theme)
- omarchy-cursor-list (lists all available cursors)
- omarchy-cursor-current (displays active cursor theme)

In addition, the file "omarchy-menu" was also updated to accommodate the new menu option.

The cursor is set by searching /usr/share/icons and ~/.local/share/icons for the provided theme name and applies the changes by 1) updating ~/.config/hypr/envs.conf with two environment variables "HYPRCURSOR_THEME" and "XCURSOR_THEME", 2) changing the active cursor theme using hyprctl, and 3) updating the GTK cursor theme using gsettings.

More information can be found here: https://wiki.hypr.land/Hypr-Ecosystem/hyprcursor/

The cursor theme selection list is built by searching the two directories mentioned previously by finding directories that contain a child directory called "cursors" (which indicates a cursor theme).

The current cursor theme is detected by searching the user's envs.conf for the value of "HYPRCURSOR_THEME".

### DIFF
--- a/bin/omarchy-cursor-current
+++ b/bin/omarchy-cursor-current
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+CONFIG_FILE="$HOME/.config/hypr/envs.conf"
+
+[[ ! -f "$CONFIG_FILE" ]] && exit
+
+grep 'env = HYPRCURSOR_THEME' "$CONFIG_FILE" | cut -d ',' -f2 | tr -d ' ' | sed -E 's/-/ /g'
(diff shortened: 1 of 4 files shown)


dhh:
Appreciate this, but in my opinion there's not enough value in the cursor changes from an aesthetic perspective. Best we just pick the best one we can have and go with that.

PR #1409, opened 2025-09-02
Enable numlock by default in input configuration

bftanase:
I believe this is a sensible default for those that have keyboards with num pads. Whenever I want to enter a number I reach for the numpad, just to realize that numlock is off and it doesn't do what I expect it.


bftanase:
Closing. I'll reopen with a MR to dev.

PR #1411, opened 2025-09-02
Fix spacing between icons on Waybar

brink-lab:
Fixes the uneven spacing between icons on waybar by updating styling.css with Propo font family. However, Propo introduces a small issue as the volume icons don't have the same width. As a consequence, icons left of volume will move based on the displayed volume icon. To solve this, I kept the monospace font for the pulseaudio module only - this way the volume icons have fixed widths but eats a bit into the spacing towards the CPU icon. But I see this as the best trade-off.

Issue described in more detail here: 
https://github.com/basecamp/omarchy/discussions/1254
https://github.com/basecamp/omarchy/issues/1394

Additionally, I deleted a space after the tray icon in the waybar conf to align space correctly between tray and bluetooth icons as well.

Related to this PR I have proposed to update the mute icon as well so it has same size as the two other volume icons plus an updated wifi icon in the same style as the other used wifi-icons: https://github.com/basecamp/omarchy/pull/1386

I would assume, that a change like this should go into the default configs under .local/ so it goes out to everyone - but I will let you decide. I am not 100% sure which file I proposed these changes to, so I need you to review.

### DIFF
--- a/config/waybar/config.jsonc
+++ b/config/waybar/config.jsonc
@@ -1,83 +1,138 @@
-@import "../omarchy/current/theme/waybar.css";
+{
+  "reload_style_on_change": true,
+  "layer": "top",
+  "position": "top",
+  "spacing": 0,
+  "height": 26,
+  "modules-left": ["custom/omarchy", "hyprland/workspaces"],
+  "modules-center": ["clock", "custom/update", "custom/screenrecording-indicator"],
+  "modules-right": [
+    "group/tray-expander",
+    "bluetooth",
+    "network",
+    "pulseaudio",
+    "cpu",
+    "battery"
+  ],
+  "hyprland/workspaces": {
+    "on-click": "activate",
+    "format": "{icon}",
+    "format-icons": {
+      "default": "",
+      "1": "1",
+      "2": "2",
+      "3": "3",
+      "4": "4",
+      "5": "5",
+      "6": "6",
+      "7": "7",
+      "8": "8",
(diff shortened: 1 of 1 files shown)


brink-lab:
Hmm. After a restart, my waybar looks like this. 
image

Not sure if there are some conflicts with the .local/ config. We need to test this and make sure.

brink-lab:
The end result with all updates listed above looks like this: 

image

versus current:

image

brink-lab:
Beware: This PR has become outdated after the Omarchy 3.0 release, which updated waybar css and conf files with the new Screenrecord feature. Merging this PR in its current state interferes with the Screenrecord feature. 

@dhh I'll update this PR (or file a new) if you are interested. Just give a heads up when you have decided on a release, so I can update towards the latest waybar files.

dhh:
The problem is that this needs to be able to work with the font selector. And I don't think most of the fonts come in Propo styles. We'd need to account for that.

brink-lab:
Right. Haven't found a solution that isn't hacky. I guess we could also add margin individually to the right hand side of the different icons and eye-ball even spacing. 
But I'm unsure if this could create new issues if icons are reorganized differently from the default.

brink-lab:
@dhh I synced the config plus css files and updated margin modifiers to add spacing instead. 
Seems to work with all the standard fonts in Omarchy.

p2ndemic:
Nice! Thanks!

PR #1412, opened 2025-09-02
feat: add DDEV development environment support with mkcert integration

grzegorzbartman:


roberto-aguilar:
Wouldn't this be redundant considering that PHP, Node and databases are available through the existing development environment options?

What DDEV seems to offer can already composed.

dannevang:
> What DDEV seems to offer can already composed.

DDEV is a lot more than that. It is easy switching between typical environments (apache, nginx), different php versions for different projects, different database types, extensions with elasticsearch, redis ect, snapshotting and for some deployment. 
Of course everything can be manually set up - but that also go for a lot of the other things added.

grzegorzbartman:
Omarchy provides a ready operating system for new developers in companies. 
DDEV extends this philosophy to individual projects - ensuring new team members can start contributing immediately.
Just as Omarchy eliminates OS setup friction, DDEV eliminates project setup friction. 
Both serve the same mission: getting developers productive faster.
Just run git clone && ddev start.
This saved us (Droptica) a lot of time (we are agency that support 100+ php projects for out clients).

Additional Benefits:
- Project isolation: Multiple PHP versions simultaneously
- Team consistency: Identical environments eliminate "works on my machine" issues  
- Framework optimization: Laravel, Drupal, WordPress, TYPO3, Magento, Craft CMS pre-tuned
- Production parity: Development mirrors deployment environment

You can manually compose the tools, but experienced teams choose abstractions that work - it's why I use Omarchy instead of configuring Arch manually.

dhh:
Mise now has support for PHP. I'd rather we get that setup, so it matches every other development environment we have on Omarchy. If you'd like to look into that, please do with: https://docs.serv00.com/PHP/

PR #1413, opened 2025-09-02
Support current wallpaper on lock screen

gkurts:
Adds support for showing the current theme's wallpaper on the lock screen.

This change adds a 'path = $background' entry to the hyprlock.conf file.

The theme must add a '$background = ~/.config/omarchy/current/background' line to the hyprlock.conf file in the theme folder for this to work.

If $background is not set, hyprlock simply ignores the line.


gkurts:
Closing to retarget dev branch

Issue #105, opened 2025-07-09
Add support for hibernate


dhh:
We need to do some setup to enable hibernate. Currently, this is the error message:

```
 systemctl hibernate
Call to Hibernate failed: Not enough suitable swap space for hibernation available on compatible block devices and file systems
```

We need a swap file that's the same size as the memory on the system for this to work.

sgloutnikov:
The following may be useful for computing swapfile size:

```
SWAP_SIZE=$(free | awk '/Mem/ {x=$2/1024/1024; printf "%.0fG", (x<2 ? 2*x : x<8 ? 1.5*x : x) }')
```

CashWilliams:
Should probably be optional during install. I don't want a 64G swap file 🤣

ryanyogan:
This may not be the ideal solution; however, you may suspend first and then hibernate.  This shouldn't use nearly as much swap?

`systemctl suspend-then-hibernate`

### Sharp Edges
* Btrfs swap does not support hibernate mode unless compression is disabled:
https://wiki.archlinux.org/title/Power_management/Suspend_and_hibernate#Hibernation_with_a_swap_file_on_Btrfs


### Potential solution

We could create dynamic swap based on the current memory usage, mirror that size, and hibernate?  Here is a rough example

```bash
#!/usr/bin/env bash
#
# dynswap: Create a swapfile sized to current RAM usage (+ margin) for hibernation
# Tested on Arch Linux.  Run as root.

set -euo pipefail

MARGIN_PCT=20           # % safety buffer above current used RAM
SWAPFILE=/swapfile      # path to swap file
FSTAB=/etc/fstab        # fstab path

## 1. Gather current RAM usage (in MiB) ########################################
# 'free --mebi' prints MiB units; field 3 in Mem row = used
USED_MIB=$(free --mebi | awk '/^Mem:/ {print $3}')
if [[ -z "$USED_MIB" ]]; then
  echo "Could not determine used memory" >&2; exit 1
fi

## 2. Calculate swap size with margin ##########################################
SWAP_SIZE_MIB=$(( USED_MIB + (USED_MIB * MARGIN_PCT / 100) ))
echo "Current RAM used: ${USED_MIB} MiB, creating ${SWAP_SIZE_MIB} MiB swapfile."

## 3. Disable and remove any existing swapfile #################################
if swapon --summary | grep -q "$SWAPFILE"; then
  swapoff "$SWAPFILE"
fi
rm -f "$SWAPFILE"

[...]

PR #1417, opened 2025-09-02
Added hibernation support

killeik:
Resolves: #105
- Creates btrfs subvolume for swap file, to not break snapshots
- Creates swap file with size of ram
- Adds swap file to `/etc/fstab` to mount swap file on boot
- Adds resume module via `HOOKS+=(resume)` in `/etc/mkinitcpio.conf.d/omarchy_resume.conf`
- Adds `Hibernate` option to Power menu  

I'd love to get some feedback 😄

### DIFF
--- a/bin/omarchy-hibernation-available
+++ b/bin/omarchy-hibernation-available
@@ -1,0 +1,11 @@
+#!/bin/bash
+# Checks if swap is big enough to hibernate
+
+SWAPSIZE_KB=$(cat /proc/swaps |grep -v -e "Filename" -e "zswap"| awk '{print $3}' );
+MEMTOTAL_KB=$(cat /proc/meminfo |grep MemTotal |awk '{print $2}' );
+if [ "${SWAPSIZE_KB:-0}" -gt "$MEMTOTAL_KB" ]; then
+  exit 0;
+else
+  exit 1;
+fi
+  
--- a/bin/omarchy-menu
+++ b/bin/omarchy-menu
@@ -446,10 +446,15 @@ show_update_password_menu() {
 }
 
 show_system_menu() {
-  case $(menu "System" "  Lock\n󱄄  Screensaver\n󰤄  Suspend\n  Relaunch\n󰜉  Restart\n󰐥  Shutdown") in
+  case $(
+  if omarchy-hibernation-available;
+  then menu "System" "  Lock\n󱄄  Screensaver\n󰤄  Suspend\n󰤁  Hibernate\n  Relaunch\n󰜉  Restart\n󰐥  Shutdown";
+  else menu "System" "  Lock\n󱄄  Screensaver\n󰤄  Suspend\n  Relaunch\n󰜉  Restart\n󰐥  Shutdown";
+  fi) in
   *Lock*) omarchy-lock-screen ;;
   *Screensaver*) omarchy-launch-screensaver force ;;
   *Suspend*) systemctl suspend ;;
+  *Hibernate*) systemctl hibernate;;
   *Relaunch*) uwsm stop ;;
   *Restart*) systemctl reboot ;;
   *Shutdown*) systemctl poweroff ;;
--- a/install/config/all.sh
+++ b/install/config/all.sh
@@ -15,6 +15,7 @@ run_logged $OMARCHY_INSTALL/config/docker.sh
 run_logged $OMARCHY_INSTALL/config/mimetypes.sh
 run_logged $OMARCHY_INSTALL/config/localdb.sh
 run_logged $OMARCHY_INSTALL/config/sudoless-asdcontrol.sh
+run_logged $OMARCHY_INSTALL/config/hibernation.sh
 run_logged $OMARCHY_INSTALL/config/hardware/network.sh
 run_logged $OMARCHY_INSTALL/config/hardware/set-wireless-regdom.sh
 run_logged $OMARCHY_INSTALL/config/hardware/fix-fkeys.sh
--- a/install/config/hibernation.sh
+++ b/install/config/hibernation.sh
@@ -1,0 +1,48 @@
+#!/bin/bash
+
+# Creates a swap file in the btrfs subvolume, adds the swap file to /etc/fstab, and adds a resume hook to mkinitcpio.
+
+SWAP_SUBVOLUME="/swap"
+if ! sudo btrfs subvolume show $SWAP_SUBVOLUME &>/dev/null ; then
+  echo 'Creating Btrfs subvolume'
+  sudo btrfs subvolume create "$SWAP_SUBVOLUME" 
+else
+  echo "Btrfs subvolume $SWAP_SUBVOLUME already exists"
+fi
+
+SWAP_FILE="$SWAP_SUBVOLUME/swapfile"
+if ! sudo swaplabel "$SWAP_FILE" &>/dev/null; then
+  echo 'Creating swapfile in Btrfs subvolume'
+  MEM_TOTAL_KB="$(grep MemTotal /proc/meminfo | awk '{print $2}')k"
+  sudo btrfs filesystem mkswapfile -s "$MEM_TOTAL_KB" "$SWAP_FILE"
+else
+  echo "File $SWAP_FILE already exists"
+fi
+
+
+if ! grep -Fq "$SWAP_FILE" /etc/fstab; then
+  echo "Adding swapfile to /etc/fstab"
+  sudo cp -a /etc/fstab "/etc/fstab.$(date +%Y%m%d%H%M%S).back"
+  printf "\n# Btrfs swapfile for system hibernation\n%s none swap defaults,pri=0 0 0\n" "$SWAP_FILE" | sudo tee -a /etc/fstab >/dev/null
+else
+  echo "Swapfile $SWAP_FILE already exists in /etc/fstab"
+fi
+
+
+echo "Enabling swap on $SWAP_FILE with priority 0"
+sudo swapon -p 0 "$SWAP_FILE"
+
+sudo mkdir -p /etc/mkinitcpio.conf.d
+
+MKINITCPIO_CONF="/etc/mkinitcpio.conf.d/omarchy_resume.conf"
+HOOKS_LINE='HOOKS+=(resume)'
+if [ ! -f "$MKINITCPIO_CONF" ] || ! grep -q "^${HOOKS_LINE}$" "$MKINITCPIO_CONF"; then
+  echo "Adding resume hooks to $MKINITCPIO_CONF"
+  echo "$HOOKS_LINE" | sudo tee -a "$MKINITCPIO_CONF" >/dev/null
+
+  echo "Regenerating initramfs..."
+  sudo mkinitcpio -P
+  echo "Hibernation enabled"
+else
+  echo "Hibernation hooks already enabled"
+fi
--- a/migrations/1758205457.sh
+++ b/migrations/1758205457.sh
@@ -1,0 +1,10 @@
+
+if omarchy-hibernation-available; then
+  echo "Swap already created, skipping migration";
+  exit 0;
+fi
+
+MEM_TOTAL_HUMAN="$(free --human |grep "Mem" |awk '{print $2}')"
+if gum confirm "Use $MEM_TOTAL_HUMAN on drive to create swap and make hibernation available?"; then
+  source  $OMARCHY_PATH/install/config/hibernation.sh
+fi


nielpattin:
Not sure about install step, but I can confirm that the hibernate work like it should be.
I'm also on Nvidia which most the problem come from and doesn't have any issue. Hope this get merge soon because I have to create a copy of Omarchy menu just to test it out or just run systemctl hibernate in terminal, but I prefer it under Suspend like this.

nielpattin:
@killeik ok, I got some problem with the hyprlock crash(not related), so I tried to reinstall and the first thing I do is to run the hibernate.sh command to test.
Run `systemctl hibernate` like usual.
But when power on, it brings me to the Omarchy Bootloader screen. So I'm not sure what wrong.

Boot up to the linux only open a new fresh just like when reboot, others app open or the last state doesn't return.

killeik:
@nielpattin 
>it brings me to the Omarchy Bootloader screen

It's okay. The boot sequence from hibernation starts like a fresh boot. It should decrypt the disk and start the kernel. Later on, however, it finds a RAM dump on the disk and loads that instead of loading fresh.

>Boot up to the linux only open a new fresh just like when reboot, others app open or the last state doesn't return.

Let's debug
I need output of that one (putted them in one command, to not make you Ctrl+C/Ctrl+V all the way  :D ) 
```bash
bash -c 'set -x ;
swapon ;
free -h ;
sudo btrfs subvolume show /swap ;
sudo btrfs inspect-internal map-swapfile /swap/swapfile ;
cat /sys/power/resume_offset ;
cat /sys/power/resume ;
cat /etc/fstab ;
cat /etc/mkinitcpio.conf.d/omarchy_resume.conf ;'
```

And kernel logs after hibernation (they are not so small, so send them as file, maybe) 
 `sudo dmesg -TH` 
 
 
For convenience, you can send them to me in DM in Discord, I'm @killeik on omarchy server

killeik:
After some debugging in DM, it was discovered that there was a magical Nvidia, for which the standard mode of operation on Linux is to torment the user. 
Other than that, everything worked fine.

brennandunn:
I've tested this on my Framework laptop / AMD Ryzen and it works great 👍 Thanks!

dhh:
Need to update for current branch, make the Hibernate option conditional on the swap being available, and add a migration with a gum confirm something like "Use XX GB on drive to make hibernate available?".

woopstar:
> Need to update for current branch, make the Hibernate option conditional on the swap being available, and add a migration with a gum confirm something like "Use XX GB on drive to make hibernate available?".

While you add it. Also make a check if swap is already configured. I have an encrypted swap on my machine so `hibernation.sh` would break my machine with custom partitions / swap / encryption.

killeik:
My apologies. I updated my fork branch to the current active branch, accidentally deleting the changes I had added, and GitHub automatically closed the pull request because the branches started to match. 

By the way, I added migration and made the hibernation option in omarchy-menu optional.

killeik:
Tested conditional hibernation in power menu, and migration - everything works correctly.

The only pitfall is that if the user manually created a swap and didn't enable the resume hook, the boot sequence will not resume from the saved hibernation image and will start fresh. 

I found a way to check if the hooks are enabled on the current initramfs with limine:
`lsinitcpio /boot/EFI/Linux/*.efi | grep 'hooks/resume'`
However, I'm unsure if it will work on GRUB and other systems.

woopstar:
Did you try the PR? I tried multiple PR in my own fork, and the menu does not work. I suspect this PR might break the omarchy menu system

killeik:
> Did you try the PR? I tried multiple PR in my own fork, and the menu does not work. I suspect this PR might break the omarchy menu system

I've tested this PR on both of my computers and it works correctly. 
I actually don't have any ideas how it can break menu, because my change for it is very simple. So can you please give more info what's going on in your fork?

woopstar:
> I've tested this PR on both of my computers and it works correctly.
> I actually don't have any ideas how it can break menu, because my change for it is very simple. So can you please give more info what's going on in your fork?

https://github.com/woopstar/omarchy/commits/master/

I will check tomorrow which PR that broke the app launcher then :)

woopstar:
It was #1661 that broke the menu. I tested this. Works as it should. But it shows the lock menu that "hangs" while it waits for hibernation to be ready. Should we consider showing something else if possible? Would it be possible to have it say "Entering hibernation" instead of showing the lock screen, which you cannot type or do anything in?

killeik:
>It was https://github.com/basecamp/omarchy/pull/1661 that broke the menu. I tested this. Works as it should.

Thank you!

> But it shows the lock menu that "hangs" while it waits for hibernation to be ready. Should we consider showing something else if possible? Would it be possible to have it say "Entering hibernation" instead of showing the lock screen, which you cannot type or do anything in?

Neither hypridle, which invokes the lock screen before sleep or hibernation, nor hyprlock, which works as a lock screen, has any difference between sleep and hibernation. Therefore, it's not possible at this level.

However, systemd has different targets: suspend.target, hibernate.target, hybrid-sleep.target, and suspend-then-hibernate.target.
Technically, we could write a service that runs before hibernate.target and changes the hyprlock configuration to display something like "Hibernating, wait" and then changes it back after hibernation. However, it feels like a bad hack, changing user-space config on every hibernation is kinda crazy. 

To do that the normal way, we would probably need to create an issue in Hyprildle/Hyprlock.

woopstar:
> However, systemd has different targets: suspend.target, hibernate.target, hybrid-sleep.target, and suspend-then-hibernate.target.
> Technically, we could write a service that runs before hibernate.target and changes the hyprlock configuration to display something like "Hibernating, wait" and then changes it back after hibernation. However, it feels like a bad hack, changing user-space config on every hibernation is kinda crazy.

Yeah, let's not walk down that road. But I can confirm the PR works as expected :)

DominicBoettger:
I am thinking about moving to OMARCHY from my customized ARCH and i think this was the script I used to enable the Swap. I used a swap file instead of a volume (i think the subvolume is the better approach). I drop my config / code as it may be helpful.
I am using hibernate when closing the lid.
I checked the issues for hibernate as this is an important feature for me.

````
#!/bin/sh
# check if su or root
if [ "$(id -u)" != "0" ]; then
  echo "This script must be run as root" 1>&2
  exit 1
fi
# check if btrfs
if [ ! -d /sys/fs/btrfs ]; then
  echo "This script is only for btrfs filesystems" 1>&2
  exit 1
fi
# check if swapfile exists if not create it with 96GB
if [ ! -f /swap/swapfile ]; then
  echo "Creating swapfile with 96GB"
  fallocate -l 96G /swap/swapfile
  chmod 600 /swap/swapfile
  mkswap /swap/swapfile
  swapon /swap/swapfile
fi

cd /tmp
# :: btrfs_map_physical :: #
wget https://raw.githubusercontent.com/osandov/osandov-linux/master/scripts/btrfs_map_physical.c
gcc -O2 -o btrfs_map_physical btrfs_map_physical.c

# :: edit grub :: #
offset=$(./btrfs_map_physical /swap/swapfile)
offset_arr=($(echo ${offset}))
offset_pagesize=($(getconf PAGESIZE))
offset=$((offset_arr[25] / offset_pagesize))
btrfsroot=`findmnt / -no UUID`

echo "btrfsroot=$btrfsroot"
echo "offset=$offset"

# create a backup of /etc/default/grub
cp /etc/default/grub /etc/default/grub.bak

# edit /etc/default/grub
sed -i "s#loglevel=3#resume=/dev/mapper/root loglevel=3#" /etc/default/grub
sed -i "s/loglevel=3/resume_offset=$offset loglevelbtrfs property set \/swap\/swapfile compression none=3/" /etc/default/grub
````

My /etc/systemd/logind.conf

````
[Login]
#NAutoVTs=6
#ReserveVT=6
#KillUserProcesses=no
#KillOnlyUsers=
#KillExcludeUsers=root
#InhibitDelayMaxSec=5
#UserStopDelaySec=10
#SleepOperation=suspend-then-hibernate suspend
HandlePowerKey=lock
#HandlePowerKeyLongPress=ignore
#HandleRebootKey=reboot
#HandleRebootKeyLongPress=poweroff
#HandleSuspendKey=suspend
#HandleSuspendKeyLongPress=hibernate
#HandleHibernateKey=hibernate
#HandleHibernateKeyLongPress=ignore
HandleLidSwitch=suspend-then-hibernate
HandleLidSwitchExternalPower=suspend-then-hibernate
#HandleLidSwitchDocked=ignore
#PowerKeyIgnoreInhibited=no
#SuspendKeyIgnoreInhibited=no
#HibernateKeyIgnoreInhibited=no
#LidSwitchIgnoreInhibited=yes
#RebootKeyIgnoreInhibited=no
#HoldoffTimeoutSec=30s
#IdleAction=ignore
#IdleActionSec=30min
#RuntimeDirectorySize=10%
#RuntimeDirectoryInodesMax=
#RemoveIPC=yes
#InhibitorsMax=8192
#SessionsMax=8192
#StopIdleSessionSec=infinity
````

**Comment by woopstar** [comment]
@killeik  I've been thinking if this should also enable the suspend-then-hibernate support.

```
$ cat /etc/systemd/sleep.conf.d/omarchy.conf
[Sleep]
HibernateDelaySec=60min
AllowSuspendThenHibernate=yes
```

```
$ cat /etc/systemd/logind.conf.d/omarchy.conf
[Login]
HandleLidSwitch=suspend-then-hibernate
HandleLidSwitchExternalPower=suspend-then-hibernate
HandleLidSwitchDocked=suspend-then-hibernate
```

Unsure if we also must:

```
systemctl enable systemd-suspend-then-hibernate.service
```

References:
https://man.archlinux.org/man/systemd-sleep.conf.5
https://man.archlinux.org/man/logind.conf.5.en

**Comment by killeik** [comment]
If we want to do something like this, I suggest using `sleep` instead of straight-forward suspend-then-hibernate for HandleLidSwitch:

> sleep
> Put the system to sleep, through suspend, hibernate, hybrid-sleep, or suspend-then-hibernate. The sleep operation to use is automatically selected by [systemd-logind.service(8)](https://man.archlinux.org/man/systemd-logind.service.8.en). By default, suspend-then-hibernate is used, and falls back to suspend and then hibernate if not supported. Refer to SleepOperation= setting in [logind.conf(5)](https://man.archlinux.org/man/logind.conf.5.en) for more details. This command is asynchronous, and will return after the sleep operation is successfully enqueued. It will not wait for the sleep/resume cycle to complete.
> 
> Added in version 256.

It will work if the system has or does not have hibernation mode. 

I'm also not sure if we really want to change HibernateDelaySec. 

Setting the `AllowSuspendThenHibernate=yes` may actually break users without hibernation. 
This flag, as I understand it, is intended to turn off the features, not to enable them.
Systemd doesn't need any changes to the configuration to find system support for hibernation. 

And the `systemd-suspend-then-hibernate.service` - it's not a daemon. It's a one-shot command to start the suspend-then-hibernate sequence.   
> Note that systemd-suspend.service, systemd-hibernate.service, systemd-hybrid-sleep.service, and systemd-suspend-then-hibernate.service should never be executed directly. Instead, trigger system sleep with a command such as systemctl suspend or systemctl hibernate.


But I'm not sure if it's a good idea to talk about these improvements in this PR. My idea was just to add support for hibernation.

**Comment by yovko** [comment]
There is an easier way to find `resume=UUID=<UUID>` and `resume_offset` cmdline parameters:

```
findmnt -no UUID -T /swap/swapfile
btrfs inspect-internal map-swapfile -r /swap/swapfile
```

And I add `resume=UUID=<UUID> resume_offset=<the-number-from-inspect-internal>` at the end of my `cmdline` in my `limine.conf` file. But I'm on my custom Arch Linux setup. Not sure about Omarchy 

https://wiki.archlinux.org/title/Power_management/Suspend_and_hibernate#Acquire_swap_file_offset

@DominicBoettger

File: bin/omarchy-hibernation-available [hunk 1]
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+# Checks if swap is big enough to hibernate
+
+SWAPSIZE=$(( 1024 * $(cat /proc/swaps |grep -v -e "Filename" -e "zram"| awk '{print $3}') ));
```
chrismazanec: this line returns math error when the swap doesn't exist. default 0 should be moved somewhere here before the multiplication.

killeik: Fixed it, thank you

Prajwal-Prathiksh:
This would be a really cool feature to extend the battery life!

lukehowson:
Great job. Worked for me too.

I prefer suspend-then-hibernate with a 1 hour delay. So I changed systemctl suspend to systemctl suspend-then-hibernate in the menu.

And similarly for /etc/systemd/logind.conf:
HandleLidSwitch=suspend-then-hibernate

And /etc/systemd/sleep.conf:
HibernateDelaySec=3600.

lukehowson:
Why no merge? dhh likes his bag toasty?

pa-philos:
Does this needs more testing reports from more people before it gets merged?

dhh:
Migration has an interactive prompt we need to sort:

image

Also, some tweaks to the migration:

```
echo "Offer hibernation if missing as a system-menu option"

if omarchy-hibernation-available; then
  exit 0
fi

MEM_TOTAL_HUMAN="$(free --human |grep "Mem" |awk '{print $2}')"
if gum confirm "Use $MEM_TOTAL_HUMAN on boot drive to make hibernation available?"; then
  source  $OMARCHY_PATH/install/config/hibernation.sh
fi
```

dhh:
It unfortunately doesn't work consistently on my machine either. There's no feedback while it's trying to write the hibernation file, but even when I left it for a few minutes, it never fully shut down the computer. I manually had to do that. 

Then on reboot, it's also stuck on the decryption screen without feedback for a long time. Long enough that you think it might be broken. Eventually it did boot up.

On the second run, I must not have waited long enough for the hibernation write to actually go through, because I shut the computer off after maybe 20 seconds, and when I reboot again, there was no hibernation restored.

I totally get why this is desirable to have, but if we can't get it to work consistently, I don't think we can offer it out of the box.

This was on a 128GB AMD 395+ in the Beelink GTR9 Pro.

killeik:
@dhh, I made the suggested changes. Now, there won't be an interactive prompt. Previously, I had to use mkinitcpio because omarchy wasn't defaulting to use only limine uki.

Regarding the issue with your PC not shutting down, do you have time to debug it?
Try adding these parameters to sleep conf:
https://wiki.archlinux.org/title/Power_management/Suspend_and_hibernate#System_does_not_power_off_when_hibernating
If this solves the problem, I will add the parameters to the installation script 

Also, if the system doesn't restore the correct state after hibernation, can you share the output of the command `sudo dmesg -TH`? There will probably be information explaining why that happened.

pch:
From my limited understanding (arch noob here), subvolumes aren’t recommended for hibernation and may lead to weird results if CoW and compression aren’t disabled for the subvolume.

Can anyone confirm?

killeik:
> From my limited understanding (arch noob here), subvolumes aren’t recommended for hibernation and may lead to weird results if CoW and compression aren’t disabled for the subvolume.

From https://btrfs.readthedocs.io/en/latest/btrfs-man5.html#swapfile-support
> subvolume - cannot be snapshotted if it contains any active swapfiles

So, to create snapshots, it's required to have a separate subvolume for swap. 

About compression - newly created subvolume doesn't contain compression flags. 
And command `btrfs filesystem mkswapfile` -  disables CoW for swapfile 
```
sudo lsattr /swap
---------------C------ /swap/swapfile
```

braintornapart:
I've taken the newest .iso of Omarchy, installed it on my ThinkPad X220, and updated it. Then I've copied the changed files from this repo to the folders located in "~/.local/share/omarchy/." and gave them the executable flag. Then I've run the "1758205457.sh" file and rebooted the laptop. Then I've added "HandleLidSwitch=suspend-then-hibernate" to "/etc/systemd/login.conf" and "HibernationDelaySec=2min" (to test it) and "AllowSuspendThenHibernate=yes" in "/etc/systemd/sleep.conf". The "Hibernation" option in Power Menu, and suspend then hibernate works flawlessly, but after entering the password to the "OMARCHY" decryption screen, I have to put the password again into the user login screen (one with the fingerprint icon).

ThinkPad X220 specs:
processor: Intel i5-2540M
memory: 16GB DDR3 1866MHz
disk: 240GB SSD

I am using GitHub for the first time in my life, so if I've done something wrong by leaving this comment, please tell me. I've given a lot of information (this may help in some way).

killeik:
> ... The "Hibernation" option in Power Menu, and suspend then hibernate works flawlessly, but after entering the password to the "OMARCHY" decryption screen, I have to put the password again into the user login screen (one with the fingerprint icon).

@braintornapart So, the problem is that you had to enter your password twice?

That's how hypridle/lock works. I wrote about that before.
> Neither hypridle, which invokes the lock screen before sleep or hibernation, nor hyprlock, which works as a lock screen, has any difference between sleep and hibernation. 

If we want to keep the login screen after sleep mode, we also have to keep it after hibernation mode. 
The first prompt on the decryption screen is required because of the encrypted disks, and we can't disable it.

It's a bit annoying, but fixing it requires changes to hypridle to differentiate between hibernation and sleep.

braintornapart:
Yes, I've thought entering password twice was a bug. Dhh had some issues with hibernation and Shirkit told that the commit probably needs more testing before merging so I thought I may help (thinking I've had a bug). I haven't read the whole discussion before commenting and I apologize for doing that. At least I've learned something today :). BTW thanks for implementing this  feature!

ExonakiDev:
@killeik 

It did not work for me, ran the script to set it up, then tried hibernating. It does seem to be writing to disk and then it turns off but then when I restart it, it acts as if it is a fresh boot. There is no saved state it resumes from.

On a thinkpad t480. Here is the output of the debug commands you had provided earlier:

```
+ swapon
NAME           TYPE       SIZE USED PRIO
/swap/swapfile file      31.1G   0B    0
/dev/zram0     partition    4G   0B  100
+ free -h
               total        used        free      shared  buff/cache   available
Mem:            31Gi       4.3Gi        24Gi       613Mi       3.3Gi        26Gi
Swap:           35Gi          0B        35Gi
+ sudo btrfs subvolume show /swap
Place your finger on the fingerprint reader
@/swap
	Name: 			swap
	UUID: 			75a4b469-29ea-f745-9f6d-32df707bcf55
	Parent UUID: 		-
	Received UUID: 		-
	Creation time: 		2025-10-28 20:13:58 -0400
	Subvolume ID: 		315
	Generation: 		31584
	Gen at creation: 	31584
	Parent ID: 		288
	Top level ID: 		288
	Flags: 			-
	Send transid: 		0
	Send time: 		2025-10-28 20:13:58 -0400
	Receive transid: 	0
	Receive time: 		-
	Snapshot(s):
	Quota group:		n/a
+ sudo btrfs inspect-internal map-swapfile /swap/swapfile
Physical start: 144453935104
Resume offset:      35267074
+ cat /sys/power/resume_offset
35267074
+ cat /sys/power/resume
253:0
+ cat /etc/fstab
# Static information about the filesystems.
# See fstab(5) for details.

# <file system> <dir> <type> <options> <dump> <pass>
# /dev/mapper/root
UUID=63d09745-51fc-4093-ba65-2b165621b32b	/         	btrfs     	rw,relatime,compress=zstd:3,ssd,space_cache=v2,subvol=/@	0 0

# /dev/mapper/root
UUID=63d09745-51fc-4093-ba65-2b165621b32b	/home     	btrfs     	rw,relatime,compress=zstd:3,ssd,space_cache=v2,subvol=/@home	0 0

# /dev/mapper/root
UUID=63d09745-51fc-4093-ba65-2b165621b32b	/var/cache/pacman/pkg	btrfs     	rw,relatime,compress=zstd:3,ssd,space_cache=v2,subvol=/@pkg	0 0

# /dev/mapper/root
UUID=63d09745-51fc-4093-ba65-2b165621b32b	/var/log  	btrfs     	rw,relatime,compress=zstd:3,ssd,space_cache=v2,subvol=/@log	0 0

# /dev/nvme0n1p1
UUID=0D45-91EB      	/boot     	vfat      	rw,relatime,fmask=0022,dmask=0022,codepage=437,iocharset=ascii,shortname=mixed,utf8,errors=remount-ro	0 2

/dev/mapper/wwanssd  /mnt/wwanssd  btrfs  defaults  0  2


# Btrfs swapfile for system hibernation
/swap/swapfile none swap defaults,pri=0 0 0
+ cat /etc/mkinitcpio.conf.d/omarchy_resume.conf
HOOKS+=(resume)
```

yovko:
@killeik 

From my modest experience, hibernation doesn’t work reliably or long-term on encrypted Btrfs unless I add to the `cmdline` in my `/boot/EFI/limine/limine.conf` file the kernel parameters `resume=UUID=<UUID-of-unlocked-/dev/mapper/root>` and `resume_offset=<number>`, which I find using the commands:

```
findmnt -no UUID -T /swap/swapfile
btrfs inspect-internal map-swapfile -r /swap/swapfile
```

`HOOKS+=(resume)` alone is not enough.

killeik:
@ExonakiDev it seems that everything configured correctly. Please send the kernel logs after hibernation resume using the command:
`sudo dmesg -TH` (perhaps as a file, as they are quite large)

@yovko on UEFI systems, the systemd-sleep - backend for `systemd hibernate` command, sets EFI variable `HibernateLocation` which contains the disk ID and resume offset. This is somehow not specified in man page for systemd-sleep, but it is in the code:
https://github.com/systemd/systemd/blob/main/src/sleep/sleep.c#L147

However, it is specified in the man that the systemd-hibernate-resume-generator reads the EFI variable and enables hibernation resume.
> When  HibernateLocation EFI variable is detected, it enables the systemd-hibernate-resume.service(8) unit, which will instruct the kernel to resume the system from the hibernation image.
https://man.archlinux.org/man/systemd-hibernate-resume-generator.8

Therefore, on modern systems with systemd and EFI, the `systemd hibernate` command does not require manual setting of the disk id and offset. 

All of the above tests confirm that.

---

Btw, all debugged cases were the Nvidia issue that looks like this in `dmesg`:
```
NVRM: GPU 0000:01:00.0: PreserveVideoMemoryAllocations module parameter is set. System Power Management attempted without driver procfs suspend interface. Please refer to the 'Configuring Power Management Support' section in the driver README.
nvidia 0000:01:00.0: PM: pci_pm_freeze(): nv_pmops_freeze [nvidia] returns -5
nvidia 0000:01:00.0: PM: dpm_run_callback(): pci_pm_freeze returns -5
nvidia 0000:01:00.0: PM: failed to quiesce async: error -5
PM: hibernation: Failed to load image, recovering.
PM: hibernation: resume failed (-5)
```
It's an Nvidia-only issue and not a common problem with the setup; some of these users encountered issues with not only hibernation but also regular sleep :D

ExonakiDev:
```
 [Wed Oct 29 20:57:49 2025] [  +0.000043] PM: Image loading done
 961 │ [Wed Oct 29 20:57:49 2025] [  +0.000004] PM: hibernation: Read 9742696 
     │ kbytes in 7.27 seconds (1340.12 MB/s)
 962 │ [Wed Oct 29 20:57:49 2025] [  +0.000798] PM: Image successfully loaded
 963 │ [Wed Oct 29 20:57:49 2025] [  +0.000006] printk: Suspending console(s) 
     │ (use no_console_suspend to debug)
 964 │ [Wed Oct 29 20:57:49 2025] [  +0.011851] NVRM: GPU 0000:01:00.0: Preser
     │ veVideoMemoryAllocations module parameter is set. System Power Manageme
     │ nt attempted without driver procfs suspend interface. Please refer to t
     │ he 'Configuring Power Management Support' section in the driver README.
 965 │ [Wed Oct 29 20:57:49 2025] [  +0.000004] nvidia 0000:01:00.0: PM: pci_p
     │ m_freeze(): nv_pmops_freeze [nvidia] returns -5
 966 │ [Wed Oct 29 20:57:49 2025] [  +0.001211] nvidia 0000:01:00.0: PM: dpm_r
     │ un_callback(): pci_pm_freeze returns -5
 967 │ [Wed Oct 29 20:57:49 2025] [  +0.000005] nvidia 0000:01:00.0: PM: faile
     │ d to quiesce async: error -5
 968 │ [Wed Oct 29 20:57:49 2025] [  +0.285883] nvme nvme1: 8/0/0 default/read
     │ /poll queues
 969 │ [Wed Oct 29 20:57:49 2025] [  +0.001820] nvme nvme0: 8/0/0 default/read
     │ /poll queues
 970 │ [Wed Oct 29 20:57:49 2025] [  +0.001983] nvme nvme0: Ignoring bogus Nam
     │ espace Identifiers
 971 │ [Wed Oct 29 20:57:50 2025] [  +0.777942] PM: hibernation: Failed to loa
     │ d image, recovering.
 972 │ [Wed Oct 29 20:57:50 2025] [  +0.273055] PM: hibernation: Basic memory 
     │ bitmaps freed
 973 │ [Wed Oct 29 20:57:50 2025] [  +0.000309] OOM killer enabled.
 974 │ [Wed Oct 29 20:57:50 2025] [  +0.000001] Restarting tasks: Starting
 975 │ [Wed Oct 29 20:57:50 2025] [  +0.000137] Restarting tasks: Done
 976 │ [Wed Oct 29 20:57:50 2025] [  +0.000018] PM: hibernation: resume failed
     │  (-5)
 977 │ [Wed Oct 29 20:57:50 2025] [  +0.094818] BTRFS: device fsid 63d09745-51
     │ fc-4093-ba65-2b165621b32b devid 1 transid 31921 /dev/mapper/root (253:0
     │ ) scanned by mount (470)
```

seems like nvidia is causing issues. I have a mx150 dgpu, always a pain

[dmesg-debug.txt](https://github.com/user-attachments/files/23222897/dmesg-debug.txt)
@killeik

ExonakiDev:
@killeik 
Good news! got it working for my nvidia dgpu. I propose this to be added to the script as part of the setup if a Nvidia gpu/hybrid gpu setup is detected. I would be happy to make the changes myself!

The `PreserveVideoMemoryAllocations` flag needs to be set to 0, and I guess it is set to 1 by default.

need to do this, and it started working for me

```
sudo nvim /etc/modprobe.d/nvidia.conf
```

Append this line at the end
```
options nvidia NVreg_PreserveVideoMemoryAllocations=0
```

Rebuild initramfs (omarchy prompts to run limine-mkinitcpio automatically)
```
sudo mkinitcpio -P
```

OR specifically for those using limine:
```
sudo limine-mkinitcpio -P
```

Then finally, you can try hibernate
```
systemctl hibernate
```

it resumes just fine, still have to enter my password twice, once for decryption, once for hyprlock

killeik:
Hi, @nielpattin! Could you test this fix?
If anyone else is having Nvidia hibernation problems, please try this fix and let me know the results. If it works for the most users, I will add it.

nielpattin:
> Hi, @nielpattin! Could you test this fix?
> If anyone else is having Nvidia hibernation problems, please try this fix and let me know the results. If it works for the most users, I will add it.
> 

Bro, I have been switch to Windows 1 month ago... Might go back to Omarchy later in near future

ali-afk:
~~So I've configured hibernation on my system before coming across this thread through creating a swap subvolume, setting its swap priority to -1 (to prevent using it as swap in favor of zram swap), and configured the resume kernel params. However, this does not resume my hyprland session. Is this intended? or is there anything I can do?~~

---EDIT
NOW FIXED
If anyone faced a hibernation issue similar to this:
```
aug 21 18:56:04 feds kernel: NVRM: GPU 0000:05:00.0: PreserveVideoMemoryAllocations module parameter is set. System Power Management attempted without driver procfs suspend interface. Please refer to the 'Configuring Power Management Support' section >
aug 21 18:56:04 feds kernel: PM: pci_pm_suspend(): nv_pmops_suspend+0x0/0x20 [nvidia] returns -5
aug 21 18:56:04 feds kernel: PM: dpm_run_callback(): pci_pm_suspend+0x0/0x160 returns -5
aug 21 18:56:04 feds kernel: nvidia 0000:05:00.0: PM: failed to suspend async: error -5`
```
Then all you have to do is add the following to to /etc/modprobe.d (with modconf hook active in mkinitcpio.conf).
```
options nvidia NVreg_PreserveVideoMemoryAllocations=0
```
Note that nvidia-{hibernate,resume,suspend}.service must be on, nvidia-persistenced and nvidia-powerd can also be safely turned on if you want.
END EDIT---

On further note,
to allow for greater flexibilty for how the user wants their system to sleep (i.e suspend-then-hibernate, hybrid-sleep,etc), I've edited omarchy-menu as so:

```
show_system_menu() {
  case $(menu ... Sleep\n󰜉  ...) in
...
  *Sleep*) systemctl sleep ;; <-- This line here
...
  *) back_to show_main_menu ;;
  esac
}
```

Therefore, sleep can be changed from /etc/systemd/logind.conf:

```
[Login]
# Example config, systemctl sleep performs the first supported command.
SleepOperation=hybrid-sleep suspend-then-hibernate suspend
```

~~On testing, I discovered that hybrid-sleep in fact does not work. Calling systemctl hybrid-sleep does not actually suspend the system.~~
Now fixed through previous edit.

killeik:
The hyprland wiki says that setting `nvidia.NVreg_PreserveVideoMemoryAllocations=1` actually fixes 'suspend/wakeup issues': https://wiki.hypr.land/Nvidia/#suspendwakeup-issues

So, who should we trust? o.O

ali-afk:
Attention NVIDIA users, hibernation DOES work with the above patches. However, hybrid-sleep (hibernation+suspend, not to be confused with suspend-then-hibernate) has certain caveats.

I have an intel iGPU and nvidia dGPU. Resuming from hybrid-sleep disconnects the NVIDIA dGPU for some reason, as in cannot be used for processing during the remainder of your session and doesn't appear on btop/nvtop.

I've scoured the net for hours and rebooted my PC almost 15-20 times. Could not resolve the issue. Most likely a driver issue from the posts I've seen online.

Watch out for this log specifically:
```
Nov 02 21:15:31 archie kernel: nvidia-modeset: ERROR: GPU:0: Failed detecting connected display devices
```
I will be on the look out for more info

patroza:
hibernation works well on my machine; x870e, 9950x3d, 96gb, 5090.
I did change the script to generate a solid 96gb swap, my system does have an iGPU and was detecting about 91-92GB ram.
if I ever change that setting I don't want to regenerate swap.
I wonder if it's a good idea to lock to known memory sizes? 8/16/24/32/36/48/64/96/128/256 etc? or to detect the actual memory, minus vram reduced?

KamQb:
Hi, I managed to get it working on desktop with nvidia gpu only. I followed this thread in [nvidia forum](https://forums.developer.nvidia.com/t/preservevideomemoryallocations-systemd-services-causes-resume-from-hibernate-to-fail/233643/10)

- run the scripts from PR
- Enabled systemd services
```
sudo systemctl enable nvidia-suspend.service
sudo systemctl enable nvidia-hibernate.service
sudo systemctl enable nvidia-resume.service
```
- removed all the nvidia related stuff from `MODULES=(...)` in `/etc/mkinitcpio.conf` 
- added `options nvidia NVreg_PreserveVideoMemoryAllocations=1` in `/etc/modprobe.d/nvidia.conf`
- run `sudo mkinitcpio -P`

I'm on nvidia/580.105.08, 6.17.8-arch1-1, rtx 3070

Edit: yeeah so I actually advise against this, it's not reliable

woopstar:
Needs a new rebase :)

pomartel:
Just tested the script on a Lenovo Thinkpad X1 Gen 13 (intel processor) and it worked like a charm!

pomartel:
I don't think it's convenient to have a menu option to hibernate. The most sensible default config would be to sleep then hibernate after a certain time if the laptop is running on battery. To my knowledge, that's how it works on MacOS and Windows. I modified the @woopstar config to hibernate after 24 hours if running on battery :

I used the following config:
```bash
# /etc/systemd/sleep.conf.d/suspend-then-hibernate.conf
[Sleep]
AllowSuspendThenHibernate=yes
HibernateDelaySec=24h
```
```bash
# /etc/systemd/logind.conf.d/suspend-then-hibernate.conf
[Login]
HandleLidSwitch=suspend-then-hibernate
```

And here is the script to set the files:
````bash
#!/bin/bash

# --- logind (lid switch) ---
LOGIND_CONF_DIR="/etc/systemd/logind.conf.d"
LOGIND_CONF_FILE="$LOGIND_CONF_DIR/suspend-then-hibernate.conf"

if [[ ! -f "$LOGIND_CONF_FILE" ]]; then
    sudo mkdir -p "$LOGIND_CONF_DIR"

    sudo tee "$LOGIND_CONF_FILE" >/dev/null <<'EOF'
[Login]
HandleLidSwitch=suspend-then-hibernate
EOF

    sudo systemctl restart systemd-logind
    echo "Created $LOGIND_CONF_FILE and restarted systemd-logind"
fi

# --- sleep (hibernate delay) ---
SLEEP_CONF_DIR="/etc/systemd/sleep.conf.d"
SLEEP_CONF_FILE="$SLEEP_CONF_DIR/suspend-then-hibernate.conf"

if [[ ! -f "$SLEEP_CONF_FILE" ]]; then
    sudo mkdir -p "$SLEEP_CONF_DIR"

    sudo tee "$SLEEP_CONF_FILE" >/dev/null <<'EOF'
[Sleep]
AllowSuspendThenHibernate=yes
HibernateDelaySec=24h
EOF

    echo "Created $SLEEP_CONF_FILE"
fi
```

pa-philos:
>  To my knowledge, that's how it works on MacOS and Windows. I modified the @woopstar config to hibernate after 24 hours if running on battery :

If this was an OS for unified hardware I'd agree, but the hardware is too big to consider 24 hours sleep the only hibernation strategy.

pomartel:
@Shirkit I don't know what you think would be a good default but waking up from hibernation is just as slow as a cold boot and you have to enter your password twice. So I wouldn't be too agressive on the timeout unless this works better and faster.

killeik:
It seems that dhh plans to disable even the normal sleep mode:  https://github.com/basecamp/omarchy/pull/3852/changes/fc04525f032656ceb81f10045ae702e00356e8c5

It seems that this MR will never be merged.  This is a little frustrating, but I understand the maintainers motivation. Probably this MR should be turned into  some kind of script/plugin outside the omarchy ecosystem so that people can add hibernation to their system...

dhh:
I'd like to have hibernation support in Omarchy, but it's not as simple as just merging this PR. I tried it when it first was posted, and there were a number of confusing issues and things that didn't work quite right. Haven't had time to get back to it yet.

But normal sleep mode also has issues on a lot of machines leading to a confusing situation where the function either doesn't work or it only works partially (leaving the system in a degraded state on resume).

I'd rather not offer something than offer it in a state where it's confusing/broken for some folks.

michaelshmitty:
> @Shirkit I don't know what you think would be a good default but waking up from hibernation is just as slow as a cold boot and you have to enter your password twice. So I wouldn't be too agressive on the timeout unless this works better and faster.

There's a big difference in UX between cold boot and waking from hibernation though. All your state gets restored. In my opinion the the whole reason of having hibernation. Cold boot times have indeed become much shorter over the years, but not having to re-open all your work in progress is a big plus.

michaelshmitty:
> It seems that dhh plans to disable even the normal sleep mode: [fc04525](https://github.com/basecamp/omarchy/commit/fc04525f032656ceb81f10045ae702e00356e8c5)
> 
> It seems that this MR will never be merged. This is a little frustrating, but I understand the maintainers motivation. Probably this MR should be turned into some kind of script/plugin outside the omarchy ecosystem so that people can add hibernation to their system...

A pity. Thanks for your effort though. I think your approach is close to the one I used (and documented here: https://hacktheplanet.be/spells/Linux/4-suspend-then-hibernate-in-omarchy/) and it has been working well on my Framework 13 AMD Ryzen 7040 laptop for the past few months. Perhaps we'll be able to come back to this PR in the future.

Sammyrules7:
I like many other found this to work fine on my framework 13. But I also found that you can skip the LUKS decryption password when using the steps detailed in #1283  although this requires TPM. Would it be worth trying to get a pr for getting a TPM implementation merged?
The one issue that i still have is when turning on from showdown i don't have any lock screen but i imagine this to be easy to do with hyprlock.

dhh:
Ended up turning this into an optional feature that can be enabled/disabled (alongside suspend). Thanks to @killeik and everyone who worked on this! Please do open PRs if we can expand support to more systems (like more nvidia combinations). Closed via https://github.com/basecamp/omarchy/pull/4103

agworkgit:
I have recently had some success on my system with getting hibernation to actually work properly and have put together a guide for others that might have a similar system to mine.

OS:
Omarchy 3.3.3
Linux 6.18.3-arch1-1

Specs:
Intel(R) Core(TM) i5-10400F (12) @ 4.30 GHz
NVIDIA GeForce GTX 1650 [Discrete]
3840x2160 in 32", 60 Hz [External]
171.39 GiB / 463.74 GiB (37%) - btrfs
2.74 GiB / 15.51 GiB (18%)
15.08 MiB / 19.51 GiB (0%)

The whole process is documented here -> https://github.com/agworkgit/omarchy-hibernation-nvidia-fix

AbdSattout:
@agworkgit this seems to break suspend-then-hibernate

agworkgit:
> @agworkgit this seems to break suspend-then-hibernate

Yeah, sorry, I forgot to mention it, I mostly use just the hibernate functions as I'm on a PC and need resume after long periods away. If you want just suspend it usually works quite well out of the box but it won't power save if you're on a PC.
Anyway, few kinks to be ironed out still but overall loving my experience on Omarchy coming from both Windows and MacOS.

PR #1420, opened 2025-09-02
Add Zellij theme support to omarchy-set-theme Closes #1419

### LINKED ISSUE #1419: Zellij or tmux theming should be includede
I use Zellij all day everyday and it has a bunch of built in themes and it would be great if I change my theme in Omarchy the theme is changed there as well. I'm assuming something else can be done in tmux too.

fabean:


dhh:
I've actually found that I don't need Zellij when running on Omarchy because the tiling window setup offers enough support for multiple terminals that it's not worth the additional abstraction. So for now, we're not going to include one in the box by default, but of course lovely for folks to install their own on top 👌

PR #1421, opened 2025-09-02
Super + RETURN unresponsive fix when btop is active (launched with Super + T)

### LINKED ISSUE #1392: SUPER + return doesn't work when btop is open
Steps to reproduce:
1. SUPER + t -> opens btop
2. SUPER + return -> does nothing

After quitting btop or moving the focus to another windows, SUPER + return works again as intented.

manuel1618:
Fixes #1392 
As explained in the issue, this seems more like a dirty hack. 
But it solves the issue of Super + Return not being picked up when btop is active and was launched with Super + T


owensd:
This only fixes part of the issue. The waybar one needs to be fixed too.

**omarchy-launch-btop**
```
!/bin/bash
uwsm app -- alacritty -e sh -c 'btop; exec $SHELL'
```

I created a bin on my system with that script and updated both the waybar config and the bindings config to call that instead.

dhh:
We need to fix this in omarchy-cmd-terminal-cwd instead. It's getting something wrong when btop is the active terminal.

owensd:
Ah, yep. It's because the string is empty:

```
#!/bin/bash

# Go from current active terminal to its child shell process and run cwd there
terminal_pid=$(hyprctl activewindow | awk '/pid:/ {print $2}')
shell_pid=$(pgrep -P "$terminal_pid" | head -n1)

if [[ -n $shell_pid ]]; then
  working_dir=$(readlink -f "/proc/$shell_pid/cwd" 2>/dev/null)
  if [ -z "$working_dir" ]; then
    echo "$HOME"
  else
    echo "$working_dir"
  fi
else
  echo "$HOME"
fi
```

manuel1618:
Like this? I'm sorry I forgot to create a dedicated branch this time. 
After applying the migration, it worked on my system.

mohimirza:
@manuel1618  The issue is not limited to just btop. I have replaced my omarchy-launch-browser from chromium to zen-browser and it's the same issue  there as well. If I have zen-browser opened up, I am not able to launch terminal with Super + T in the same workspace.

Related to #1464

manuel1618:
@mohimirza 

I just tested this with LibreWolf from #1464 and can confirm:
- reproduced the error
- migration from this PR also fixes this

dhh:
Thanks for working on this. We never need migrations for bins that are included in Omarchy. I committed here: https://github.com/basecamp/omarchy/commit/05a5bfc9b43b016a497c999cefad403e25905f3f.

PR #1429, opened 2025-09-03
Fix UFW Persistence

martinmose:
I noticed UFW looked fine on first run but went inactive after a reboot.

## What’s fixed
- UFW was enabled with `ufw --force enable` but didn’t survive reboots because the systemd service wasn’t enabled
- Added `systemctl enable ufw` so the firewall stays active after restart
- Added a migration script (`migrations/1756911131.sh`) to fix this for existing installs

## Why
Without the systemd service `ufw status` shows inactive after a reboot, leaving systems unprotected

## Side note: SSH defaults
Right now [SSH on port 22 is open by default](https://learn.omacom.io/2/the-omarchy-manual/93/security). That feels a bit too open for Omarchy’s “sane defaults”?

Suggestion:
- Don’t allow SSH by default
- Recommend Tailscale first (`omarchy-install-tailscale`) before enabling port 22.  
  (SSH can also work without opening port 22 by running: `sudo tailscaled up --ssh`)
- Document how to enable SSH manually (e.g. `ufw allow 22/tcp`) for those who need it

### DIFF
--- a/migrations/1756911131.sh
+++ b/migrations/1756911131.sh
@@ -1,13 +1,11 @@
 echo "Enable UFW systemd service for existing installations"
 
-# Check if UFW is installed and configured but systemd service not enabled
-if command -v ufw >/dev/null 2>&1; then
-  # Check if UFW has rules configured (indicating it was set up by Omarchy)
-  if sudo ufw status | grep -q "Status: active\|22/tcp\|53317"; then
-    # Enable UFW systemd service if not already enabled
-    if ! systemctl is-enabled ufw >/dev/null 2>&1; then
-      sudo systemctl enable ufw
-      echo "UFW systemd service enabled"
+if omarchy-cmd-present ufw; then
+    if sudo ufw status | grep -q "Status: active\|22/tcp\|53317"; then
+        if ! systemctl is-enabled ufw >/dev/null 2>&1; then
+            sudo systemctl enable ufw --now
+            echo "UFW systemd service enabled"
+        fi
     fi
-  fi
 fi
+


migrations/1756911131.sh
```diff
@@ -0,0 +1,13 @@
+echo "Enable UFW systemd service for existing installations"
+
+# Check if UFW is installed and configured but systemd service not enabled
+if command -v ufw >/dev/null 2>&1; then
```
roberto-aguilar: Omarchy now has a [omarchy-cmd-present](https://github.com/basecamp/omarchy/blob/master/bin/omarchy-cmd-present) that highly improves the readability so probably is worth using it for consistency with [other migration files](https://github.com/basecamp/omarchy/blob/master/migrations/1754984734.sh#L3C4-L3C23).

```suggestion
if omarchy-cmd-present ufw; then
```

migrations/1756911131.sh
```diff
@@ -0,0 +1,13 @@
...
+  # Check if UFW has rules configured (indicating it was set up by Omarchy)
+  if sudo ufw status | grep -q "Status: active\|22/tcp\|53317"; then
+    # Enable UFW systemd service if not already enabled
+    if ! systemctl is-enabled ufw >/dev/null 2>&1; then
+      sudo systemctl enable ufw
```
roberto-aguilar: If `ufw` is being enabled in here, it would be best to start it as well.

```suggestion
      sudo systemctl enable ufw
      sudo systemctl start ufw
```

martinmose:
@roberto-aguilar thanks a lot for the review. Good feedback. I have commited both.

migrations/1756911131.sh
killeik: The code you wrote will work absolutely fine, but just so you know, there's a handy flag that you can use to start or stop the unit when you enable or disable it. 

```
> systemctl enable --help |grep now
     --now               Start or stop unit after enabling or disabling it
```

martinmose: I didn't know! I will update to use `--now` instead! Thanks

martinmose:
This issue https://github.com/basecamp/omarchy/issues/1423 was created before the pull request - I just wanted to link them here for reference.

PR #1433, opened 2025-09-03
Split Development menu; add Railway CLI; add Mise uninstall

JustinMissmahl:
What:

Split Development menu into submenus: Languages, CLI Tools, Databases.
Add Railway CLI installer to omarchy-install-dev-env.
Add Mise uninstall flow: new omarchy-mise-remove script and Remove menu entry.

Why:

Reduce menu clutter, group items logically.
Provide a common deploy CLI.
Let users remove mise tools interactively.

How:

bin/omarchy-menu: new show_install_dev_languages_menu, show_install_dev_cli_menu, show_install_dev_databases_menu; Remove menu adds “Mise”.
bin/omarchy-install-dev-env: new railway case using mise use --global railway@latest.
bin/omarchy-mise-remove: bash script; lists installed tools via mise ls --installed --json + jq; multi-select with gum (preferred) or fzf fallback; runs mise uninstall --yes (falls back to tool-only on version errors).

Notes:

No breaking changes; existing menu items unchanged.
Dependencies already used in repo: jq, gum optional, fzf fallback.

Test:

Launch omarchy-menu → Install → Development:
Languages submenu routes correctly.
CLI Tools → Railway installs via mise.
Databases → Docker DB unchanged.
Remove → Mise: shows installed mise tools; can multi-select and uninstall; shows “done” afterward.


dhh:
Railway is too specific to have as part of this, and I don't think adding an extra layer to the development menu brings enough value for the extra indirection. But I would take the mise uninstall option on it's own.

JustinMissmahl:
> Railway is too specific to have as part of this, and I don't think adding an extra layer to the development menu brings enough value for the extra indirection. But I would take the mise uninstall option on it's own.

The base idea was to have the possibility to extend the CLI menu, maybe with other services like Vercel, AWS, Fly.io, and so on. I was traveling, so sadly I didn't get to it yet, but if that is maybe a bit overengineered, it's understandable.

But yeah, I also liked the mise menu, as I thought about it, maybe even bringing install, remove, and a new manage tab into one tab to not only add and remove, but maybe also manage versions and so on, as mise gives helpful control over it.

JustinMissmahl:
please ignore the last commit i forgot to put it into its own branch i know ghossty is coming

dhh:
Can you clean this up to just focus on the mise uninstall?

JustinMissmahl:
> Can you clean this up to just focus on the mise uninstall?

Sure will redo it from the latest point on quickly i am back at my laptop in about three hours

dhh:
If you're still interested in working on the mise uninstall, please open a new PR.

PR #1436, opened 2025-09-03
Add Qwen-Code AI tool to installation menu

SOBANEJAZ:
- Uses AUR to install **qwen-code** CLI tool
- Added Qwen-Code installation option to **bin/omarchy-menu** file in the AI Installation function e.g. (show_install_ai_menu())
- Follows existing pattern for AUR-based AI tools

### DIFF
--- a/bin/omarchy-menu
+++ b/bin/omarchy-menu
@@ -227,7 +227,7 @@ show_install_ai_menu() {
       echo ollama
   )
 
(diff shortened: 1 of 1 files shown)


dhh:
I think we're well-enough covered with the majors that folks are most likely to use for now. Anyone can always add more directly via Install > AUR 👌

PR #1437, opened 2025-09-03
Feed the image folder to the Image Viewer so the next/previous image navigation works

bftanase:
When double clicking an image in Nautilus, my expectation is that I can navigate to the other images in the folder with the arrow keys.

This doesn't happen currently because the imv command receives only the filename as a parameter.

However, if we use:

`imv -n /myfolder/image.jpg /myfolder`

This works because we're opening  the folder starting at the image given by the -n parameter.

I couldn't find a more elegant way of doing this, so better ideas are welcome!

### DIFF
--- a/applications/imv.desktop
+++ b/applications/imv.desktop
@@ -1,6 +1,6 @@
 [Desktop Entry]
 Name=Image Viewer
-Exec=imv %F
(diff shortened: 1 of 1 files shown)


bftanase:
Closing. I'll reopen with a MR to dev.

PR #1438, opened 2025-09-03
Add env vars to hyprland for screen sharing

keithhubner:
Previously it was not possible to screen share via apps like Google Meet. These environment variables are used to tell the applications to use session type wayland and desktop environment to hyprland instead of X11 which is defaulted to by many applications.


dhh:
Very nice. Just needed to go in the default files instead of the generated one. Added with your credit 👌

keithhubner:
Ah yes! Of course! Thanks @dhh !

PR #1439, opened 2025-09-04
Support for wallpapers in hyprlock

gkurts:
(closed original #1413 and opened this one to target _from_ and _to_ dev branch)

Adds support for showing the current theme's wallpaper on the lock screen.

This change adds a 'path = $background' entry to the hyprlock.conf file.

The theme must add a '$background = ~/.config/omarchy/current/background' line to the hyprlock.conf file in the theme folder for this to work.

If $background is not set, hyprlock simply ignores the line.

### DIFF
--- a/config/hypr/hyprlock.conf
+++ b/config/hypr/hyprlock.conf
@@ -3,6 +3,7 @@ source = ~/.config/omarchy/current/theme/hyprlock.conf
 background {
     monitor =
     color = $color
+    path = $background
 }
 
 animations {
--- a/migrations/1757277657.sh
+++ b/migrations/1757277657.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+conf="$HOME/.config/hypr/hyprlock.conf"
+
+if ! grep -q "path = \$background" "$conf"; then
+  sed -i '/color = \$color/a \    path = \$background' "$conf"
+fi


gkurts:
Link to bug #1488

dhh:
This will need a migration to add it to existing hyprlock.conf files. Use grep + sed so we don't have to refresh the whole file 👌

gkurts:
Got chya. I'll get that added to this PR this evening.

gkurts:
Added migration, but to be clear, adding this line to the hyprlock.conf will not trigger the wallpaper to show on the lock screen.

The theme must add the `$background  = ~/.config/omarchy/current/background` value into its hyprlock.conf file. I can do that as a migration for all default themes or simply leave it as an exercise for the end user and/or theme designers.

dhh:
I like leaving it for the theme designers. Don't want the default themes to use this, but happy to let extra themes go to town!

mrpbennett:
Just for my learning can you explain the use of:

> The theme must add a '$background = ~/.config/omarchy/current/background' line to the hyprlock.conf file in the theme folder for this to work.

In my hyprlock file I only have the following and it changes for all themes and all wallpapers. 

```
background {
    monitor =
    path = $HOME/.config/omarchy/current/background
    ...
```

Just curious on the variable use.

D-Pagey:
> In my hyprlock file I only have the following and it changes for all themes and all wallpapers.
> 
> ```
> background {
>     monitor =
>     path = $HOME/.config/omarchy/current/background
>     ...
> ```

I added `blur_passes = 1` to this and it is :ok_hand:

gkurts:
Sure! If the theme designer desides to apply a different wallpaper for the lock screen or something like that, they can do it by specifying $background = $HOME/.config/omarchy/current/<path-to-desired-background. This way, the omarchy hyprlock.conf will respect whatever path the theme designer gives them. Although they will probably just use the standard $HOME/.config/omarchy/current/background symlink, they have that choice.

I really only even went down this path to set the lock screen wallpaper because my daughter likes seeing the [omarchy-all-hallows-eve](https://github.com/guilhermetk/omarchy-all-hallows-eve-theme) background. 😄 

Going to have to check out blur_passes!!

mrpbennett:
> Sure! If the theme designer desides to apply a different wallpaper for the lock screen or something like that, they can do it by specifying $background = $HOME/.config/omarchy/current/<path-to-desired-background. This way, the omarchy hyprlock.conf will respect whatever path the theme designer gives them. Although they will probably just use the standard $HOME/.config/omarchy/current/background symlink, they have that choice.
> 
> 
> 
> I really only even went down this path to set the lock screen wallpaper because my daughter likes seeing the [omarchy-all-hallows-eve](https://github.com/guilhermetk/omarchy-all-hallows-eve-theme) background. 😄 
> 
> 
> 
> Going to have to check out blur_passes!! 

Got you!! That makes sense now. Thanks for clarifying ￼

dhh:
I don't think this actually works, because often you need to run other background config changes like blur passes, but after playing around with this enough, I got convinced that we should just use the current background and a few blur passes. So that seems like what this was aiming for anyway 😄

gkurts:
Got chya! I do like the simpler is better approach. Thanks for the explanation on the decision!

D-Pagey:
@mrpbennett our suggestion made it into Omarchy v3 :grin:

mrpbennett:
> @mrpbennett our suggestion made it into Omarchy v3 :grin: 

Well now I guess I can say I contributed to Omarchy! 😂😂😂

PR #1444, opened 2025-09-04
Feed the image folder to the Image Viewer so the next/previous image navigation works

bftanase:
When double clicking an image in Nautilus, my expectation is that I can navigate to the other images in the folder with the arrow keys.

This doesn't happen currently because the imv command receives only the filename as a parameter.

However, if we use:

`imv -n /myfolder/image.jpg /myfolder`

works becasue we're opening  the folder starting at the image given by the -n parameter.

### DIFF
--- a/applications/imv.desktop
+++ b/applications/imv.desktop
@@ -1,6 +1,6 @@
 [Desktop Entry]
 Name=Image Viewer
-Exec=imv %F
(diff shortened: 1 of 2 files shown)


dhh:
This needs a migration to update the existing imv.desktop files that have been copied into .local/share/applications.

bftanase:
@dhh thanks for the feedback. I think I understand how it works now. See my latest commit.

Issue #1414, opened 2025-09-02
can we just use NetworkManager...


kromsam:
It is has become basically the standard in the Linux Desktop space. it has by far the best support and a lot of apps (such as VPNs) need it to work.

I have really like Omarchy so far, but Impala with IWD has been of the experiences about it I really don't like. Have already bumped into a lot of issues were all of a sudden DNS does not resolve or stuff like that.

NetworkManager 'just works'.

daanlenaerts:
I noticed similar issues while configuring Wireguard VPN. Configuring a custom DNS server doesn't seem to work.

blazskufca:
I second the proposal. I also did the switch from systemd-networkd to NetworkManager manually, I've also had issues with DNS until I've explicitly set the [DNS resolver to systemd-resolved](https://wiki.archlinux.org/title/NetworkManager#systemd-resolved). You could try if it works for you too

adam-coates:
For vpn's generally I have found the likes of openconnect-sso to work well. Particularly to connect to an eduroam vpn, such as a university vpn. However, I've yet to try with a proton vpn or any custom DNS for that matter.

leonardobetti:
I second this proposal.

381181295:
Yeah, having the option to connect to a wireguard vpn with NetworkManager would be pretty cool

jpdombrowski:
This is mainly an issue of not being used to systemd-networkd. Setting up wireguard is manageable, even w/ wg-quick and proton provided config files if you set DNS via resolvectl in the PostUp hook.

Switching to NetworkManager also is possible and can even be done cleanly w/o any manual work during archinstall.

systemd-networkd not supporting wifi standards does not make any sense. iwd handles wifi authentication. impala is a frontend for iwd. iwd supports owe/wpa3.

Omakub comes with NetworkManager. Arch defaults to networkd but makes the switch easy. That's a philosophical choice. Arch favours simplicity.

The Arch Wiki covers all of this. Check it out.

daanlenaerts:
> This is mainly an issue of not being used to systemd-networkd. Setting up wireguard is manageable, even w/ wg-quick and proton provided config files if you set DNS via resolvectl in the PostUp hook.

Oh yeah it absolutely is, this is my first experience with Arch. I did set up Wireguard with some PostUp and PreDown hooks, and that works fine. It's just not really user friendly.

[...]

PR #1446, opened 2025-09-04
[NetworkManager] - Attempt to switch from systemd-networkd to NetworkManager 

blazskufca:
Attempt to switch from `systemd-networkd`to `NetworkManager`as proposed in #1414. 

Sadly `impala`, the current network manager TUI [does not play nice with NetworkManager](https://github.com/pythops/impala/issues/24). Until upstream issues is resolved `nmtui` could be used instead, although it's not as good as `impala`.

### DIFF
--- a/migrations/1756659875.sh
+++ b/migrations/1756659875.sh
@@ -7,10 +7,15 @@ fi
 
 yay -S --noconfirm --needed networkmanager
 
-sudo systemctl disable --now systemd-networkd.service
-sudo systemctl disable --now systemd-networkd-wait-online.service
+if systemctl is-enabled --quiet systemd-networkd.service 2>/dev/null; then
+  sudo systemctl disable --now systemd-networkd.service
+fi
 
-if systemctl is-active --quiet wpa_supplicant.service 2>/dev/null; then
+if systemctl is-enabled --quiet systemd-networkd-wait-online.service 2>/dev/null; then
+  sudo systemctl disable --now systemd-networkd-wait-online.service
+fi
+
+if systemctl is-enabled --quiet wpa_supplicant.service 2>/dev/null; then
   sudo systemctl disable --now wpa_supplicant.service
   sudo systemctl mask wpa_supplicant.service
 fi


wixisse:
maybe it would be better to actually have the option to have a cli to change the wifi and blurttoth and also maybe having the wallpaper selector to be weseclector
and maybe also to have valent act as a phone link app

ghost:
I think this commit should address the DNS leak situation caused by NM when using vanilla VPN clients that don't combat NM quirks.

https://askubuntu.com/questions/1394798/ubuntu-20-04-network-manager-wireguard-dns-leak

jardahrazdera:
Omarchy uses systemd-networkd + iwd by design. This is a deliberate architectural choice for simplicity and minimal dependencies.

NetworkManager would:
- Add unnecessary complexity
- Conflict with existing systemd-networkd configuration
- Break impala (as you noted)
- Go against Omarchy's philosophy of using systemd components

cc @dhh - proposal to replace systemd-networkd with NetworkManager

blazskufca:
@jardahrazdera A counterargument I’d make is that sticking with networkd breaks compatibility with many user-facing networking applications. For instance, if someone wants to use a VPN app like ProtonVPN, they won’t be able to do so with networkd.

From what I understand, being user-friendly is a core goal of Omarchy. A lot of effort has gone into making things as simple and intuitive as possible. In that spirit, I believe using a networking daemon that just works with various VPNs and similar tools outweighs any theoretical concerns about “unnecessary complexity” or deviating from systemd components.

DHH has already been tagged in the original issue, so I believe he’s aware of it and will decide in due time which direction he wants to take.  We probably don't need to keep tagging him.

dhh:
I'm always open to ideas, but I'm not down for nmtui. So going to need another, great way of managing wifi selection before this could even be considered. If that emerges, feel free to open again!

kromsam:
because of all the changes in Omarchy in the mean time it felt more sensible to create a new PR:

https://github.com/basecamp/omarchy/pull/2909

PR #1452, opened 2025-09-04
fix: NetworkManager DNS conflicts with systemd-resolved

arafays:
I installed network manager while installing arch and was running into problem that i was unable to browse internet until i ran the setup dns script in omarchy this fixes it perminently updates the network manger and patches it configuration.

This pull request adds a migration script to address DNS conflicts between NetworkManager and systemd-resolved. The script checks for misconfigurations and hardcoded DNS entries, applies necessary fixes, and restarts NetworkManager if changes are made.

NetworkManager and systemd-resolved DNS integration:

* Added `migrations/1757005778.sh` to detect and fix DNS configuration conflicts between NetworkManager and systemd-resolved, ensuring proper integration.
* Checks for the presence and status of NetworkManager and systemd-resolved before proceeding with configuration changes.
* Detects missing or incorrect NetworkManager DNS configuration in `/etc/NetworkManager/conf.d/99-dns-systemd-resolved.conf` and creates or updates the file as needed.
* Identifies NetworkManager connections with hardcoded DNS settings and clears them to prevent conflicts.
* Restarts NetworkManager automatically if any DNS configuration changes are applied.This migration ensures NetworkManager and systemd-resolved cooperate by:
  - Enforcing `dns=systemd-resolved` in a dedicated NetworkManager conf.
  - Detecting and clearing hardcoded IPv4 DNS on active connections.
  - Restarting NetworkManager when changes are applied.
  
The goal is to prevent DNS resolution conflicts and flaky name lookups on systems using systemd-resolved alongside NetworkManager.

### DIFF
--- a/migrations/1757005778.sh
+++ b/migrations/1757005778.sh
@@ -1,0 +1,91 @@
+#!/bin/bash
+
+echo "Fix NetworkManager DNS conflicts with systemd-resolved"
+
+# Check if NetworkManager is installed and running
+if ! command -v nmcli >/dev/null 2>&1; then
+    echo "NetworkManager not found, skipping DNS configuration fix"
+    exit 0
+fi
+
+if ! systemctl is-active --quiet NetworkManager; then
+    echo "NetworkManager not running, skipping DNS configuration fix"
+    exit 0
+fi
+
+# Check if systemd-resolved is available
+if ! systemctl is-active --quiet systemd-resolved; then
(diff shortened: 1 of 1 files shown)


jardahrazdera:
Omarchy doesn't use NetworkManager. We use systemd-networkd + iwd by design.

This migration would be a no-op on standard Omarchy systems.

cc @dhh

See #1446 for context.

Issue #1423, opened 2025-09-03
ufw status


jonwsoto:
I am a linux newb here, just set this up on my Mac Mini 2014 🎉. 

I noticed by using `sudo ufw status` it was set to inactive. I was reading in the manual that it should be enabled by default. Any way I could have screwed this up on the initial install? I did have a slight hiccup as I used the first release IOS and obtained the black screen. I was able to get it up and running and now updated to v2.1.0.  Just kind of concerning that it wasn't enabled as it stated in the [documentation](https://learn.omacom.io/2/the-omarchy-manual/93/security). I used `sudo systemctl enable ufw --now`
to get it up. It appears to persist now after reboot. ChatGPT thinks maybe it was missing enablement step in the installer?

dhoppe:
I can confirm that ufw is not enabled by default. 

```
~ ❯ sudo ufw status
[sudo] password for dhoppe:
Status: inactive

~ ❯ sudo systemctl status ufw
○ ufw.service - CLI Netfilter Manager
     Loaded: loaded (/usr/lib/systemd/system/ufw.service; disabled; preset: disabled)
     Active: inactive (dead)

~ ✗ sudo ufw --force enable
Firewall is active and enabled on system startup

~ ❯ sudo ufw status
Status: active

To                         Action      From
--                         ------      ----
53317/udp                  ALLOW       Anywhere
53317/tcp                  ALLOW       Anywhere
22/tcp                     ALLOW       Anywhere
172.17.0.1 53/udp          ALLOW       172.16.0.0/12              # allow-docker-dns
53317/udp (v6)             ALLOW       Anywhere (v6)
53317/tcp (v6)             ALLOW       Anywhere (v6)
22/tcp (v6)                ALLOW       Anywhere (v6)

~ ❯ sudo systemctl status ufw
○ ufw.service - CLI Netfilter Manager
     Loaded: loaded (/usr/lib/systemd/system/ufw.service; disabled; preset: disabled)
     Active: inactive (dead)

~ ✗ sudo systemctl enable --now ufw
Created symlink '/etc/systemd/system/multi-user.target.wants/ufw.service' → '/usr/lib/systemd/system/ufw.service'.

~ ❯ sudo systemctl status ufw
● ufw.service - CLI Netfilter Manager
     Loaded: loaded (/usr/lib/systemd/system/ufw.service; enabled; preset: disabled)
     Active: active (exited) since Thu 2025-09-04 16:01:17 CEST; 4s ago
[...]

PR #1456, opened 2025-09-04
fix(ufw): Enable and start ufw.service by default

dhoppe:
Fixes: #1423

### DIFF
--- a/install/first-run/firewall.sh
+++ b/install/first-run/firewall.sh
@@ -15,7 +15,7 @@ sudo ufw allow 22/tcp
 sudo ufw allow in proto udp from 172.16.0.0/12 to 172.17.0.1 port 53 comment 'allow-docker-dns'
 
 # Turn on the firewall
-sudo ufw --force enable
+sudo systemctl enable --now ufw.service
 
 # Turn on Docker protections
 sudo ufw-docker install
--- a/migrations/1757023182.sh
+++ b/migrations/1757023182.sh
@@ -1,0 +1,9 @@
+#!/usr/bin/env bash
+
+echo "Enable and start the UFW service"
+if systemctl is-enabled --quiet ufw.service && systemctl is-active --quiet ufw.service; then
+  # UFW is already enabled and running, nothing to change
+  :
+else
+  sudo systemctl enable --now ufw.service
+fi


roberto-aguilar:
Duplicate of https://github.com/basecamp/omarchy/pull/1429

dhoppe:
@roberto-aguilar I was not aware of this pull request, since it has not been linked to the issue. However, the other pull request will start the ufw.service only during the migration, not after it has been installed.

martinmose:
@dhoppe well... it's because the pull request was created before the issue :sweat_smile: 
That is not correct. Please look at the pull request again?

roberto-aguilar:
No worries at all, @dhoppe, just though about leaving a reference so that there's no duplicate work as @martinmose has been iterating over his PR as well.

You're right, https://github.com/basecamp/omarchy/pull/1429 doesn't explicitly call `systemctl start` during the installation process but that may be because `sudo ufw --force enable` is being called above.

What `ufw enable` seems to be doing behind the scenes along starting the process is to apply and persists its configuration to `/etc/ufw/ufw.conf` which is not the same command being invoked by `systemctl start`.

The `/usr/lib/systemd/system/ufw.service` definition roughly looks like this:

```
[Service]
ExecStart=/usr/lib/ufw/ufw-init start
```

That's how deep I went the rabbit hole, but make sense to use `ufw enable` during installation and only `systemctl start` afterwards.

martinmose:
> @roberto-aguilar I was not aware of this pull request, since it has not been linked to the issue. However, the other pull request will start the ufw.service only during the migration, not after it has been installed.

@dhoppe and another thing: 
To avoid duplicate effort and respect the maintainers’ time it would be great if we could share feedback and improvements in one PRs instead of duplicating the work.

install/first-run/firewall.sh
```diff
@@ -15,7 +15,7 @@ sudo ufw allow 22/tcp
 sudo ufw allow in proto udp from 172.16.0.0/12 to 172.17.0.1 port 53 comment 'allow-docker-dns'
 
 # Turn on the firewall
-sudo ufw --force enable
+sudo systemctl enable --now ufw.service
```
dhh, line 18: Can't run --now in an installer that is executed in a chroot.

dhh:
Fixed via https://github.com/basecamp/omarchy/pull/1429.

PR #1458, opened 2025-09-04
Copy URL to clipboard from Chromium apps

jankeesvw:
This afternoon I was at Rails World, talking to @dhh, I was asking if there is a way to copy the current URL to the clipboard with a Omarchy web application. We came up with an idea to use a Chrome extension that would be bundled with Omarchy. In this PR there is a bundled Chrome extension that is loaded via the `/config/chromium-flags.conf`.

Below in the screenshot you'll see an example of how it works. I'm browsing with the preinstalled HEY application, when I press `ALT+SHIFT+U` the current URL is on my clipboard and I see a small notification. 

image

## How to test:

```bash
cd ~/.local/share/omarchy
gh pr checkout 1458
omarchy-update

pkill chromium # not required
```

You should see this:

```bash
# Running migration (1757021485)
# Install Copy URL extension for Chromium
```
Then start a chromium browser, et voila.


migrations/1757021485.sh
```diff
@@ -0,0 +1,50 @@
...
+          message: 'URL copied'
+        });
+      });
+    });
+  }
+});
+EOF
+
+chromium --headless --load-extension="$EXTENSION_DIR" --dump-dom about:blank &>/dev/null
+echo $EXTENSION_DIR
```
jankeesvw: This unfortunately doesn't work.

jankeesvw:
I just pushed an update to this PR, it is now wrapped in the launcher for web apps. Unfortunately it isn't consistently working yet. If you kill all Chromium sessions it works, but if there is still a session active it won't.

hjanuschka:
guess chromium needs atleast one full restart?!

jankeesvw:
> guess chromium needs atleast one full restart?!

My solution works when you don't have any other Chromium open, so it has to be the first. 

Right now I'm working on making a .crx, which should work.

hjanuschka:
i wonder if we should just patch chromium to have a global hotkey 🤨

jankeesvw:
> i wonder if we should just patch chromium to have a global hotkey 🤨

Is that easier to do? I think this extension could work.

hjanuschka:
shouldn't be that hard, extension for me feels a bit heavy, but well on the otherside if we can ship it so that it works reliable


so your PR works if the first chromium that launches has the config to load the extension?
if you'd publish the extension to the store, we could just enable it by default via policy (i think this is what is planned for 1-password; or done already)

jankeesvw:
> shouldn't be that hard, extension for me feels a bit heavy, but well on the otherside if we can ship it so that it works reliable
> 
> so your PR works if the first chromium that launches has the config to load the extension? if you'd publish the extension to the store, we could just enable it by default via policy (i think this is what is planned for 1-password; or done already)

Shipping it to the store is the easiest, but the disadvantage is that this is also an attack vector that could be exploited, if someone overwrites the extension (through my account) they could compromise the browser. That's why I started with making the extension locally.

The flag `--load-extension` works great, but only if you aren't running Chromium, otherwise it reuses the existing Chromium. 

I'll try to make some time this weekend. Just got back from the conference.

hjanuschka:
cant we somehow download it  and verify the hash? to protect against tempered extension?

> The flag --load-extension works great, but only if you aren't running Chromium, otherwise it reuses the existing Chromium.


yes this is because there can only be one i nstance running, we could add --load-extension=  to https://github.com/basecamp/omarchy/blob/master/config/chromium-flags.conf


so no matter how chromium is started it gets the flags

jankeesvw:
Thanks for the pointer! I'll take a look at it tomorrow.

jankeesvw:
Added the Omarchy icon to the extension so people recognize the origin of this extension:

screenshot-2025-09-06_19-23-24

config/chromium-flags.conf
```diff
@@ -1,3 +1,4 @@
 --ozone-platform=wayland
 --ozone-platform-hint=wayland
 --enable-features=TouchpadOverscrollHistoryNavigation
+--load-extension=$HOME/.local/share/omarchy/applications/chromium/copy-url
```
hjanuschka: i wonder if we could use "~/.local/share/omarchy/applications/chromium/copy-url" and then could totally avoid the $HOME changes in this PR

hjanuschka:
so - i am not the one deciding to merge (cc @dhh )  - but overall this looks good to me.

i would like to rename the extension to "omarchy" - and keep it as a place for further stuff - and not making it explicitly only for copy of url.

config/chromium-flags.conf
jankeesvw: @hjanuschka nice! This works.

I had tried $HOME which doesn't work. But ~ works fine, great suggestion. I dropped the change to the update script.

jankeesvw: @hjanuschka can you give it a try?

hjanuschka: it works for me! :) but i had to change the shortcut, as my customized conf + german macbook keyboard

dhh:
I'm digging this! But let's use `ctrl + alt + L` as the hotkey to mirror that `ctrl + L` is how you access the location bar. Let's refer to icon.png in the root of Omarchy folder too. And let's put this under `extensions/copy-url` at the top level instead of `applications`.

jankeesvw:
> I'm digging this! But let's use `ctrl + alt + L` as the hotkey to mirror that `ctrl + L` is how you access the location bar. Let's refer to icon.png in the root of Omarchy folder too. And let's put this under `extensions/copy-url` at the top level instead of `applications`.

Nice to hear that you like it, I agree with your suggestions. However there are two minor "issues";

1. `ctrl + alt + L` is not available, but `ctrl + shift + L` is
2. The icon in the root is an ascii file, can I include a PNG version of the logo in the root?

What do you think?

jankeesvw:
@dhh I just pushed an update to this branch. I went with `ctrl + shift + L`, since it's close to `ctrl + L` like you mentioned. I also added a PNG version of the logo in the root and created a symlink from the extension folder to that file. Plus, I moved everything to the suggested folder. :ok_hand:

config/chromium-flags.conf
jankeesvw: Cool, thanks for testing. I added a little explanation to the commit to make it clear you can change it.

jankeesvw:
I just rebased this pull request with `master` after the release of #1540

dhh:
Hmm, I can get this to load, but the hotkey is not getting picked up. Ctrl + Shift + L doesn't do anything. I checked that the extension is loaded, and it is. But yeah, no hotkey is exposed. Not in chromium itself, not in the webapps.

dhh:
Moving it to Alt + Shift + L worked 👌

jankeesvw:
@dhh ah, you noticed my changed. :-) 

I just asked some friends to test it. I'm glad it now works on your machine too.

jankeesvw:
`Alt+L` is nice, thanks! I tested it and it works for me too.

dhh:
I had Alt + L working, but then I couldn't get it working again?? It refused to allow anything but Alt + Shift + L.

jankeesvw:
> I had Alt + L working, but then I couldn't get it working again?? It refused to allow anything but Alt + Shift + L.

I had the same issue, strange. There is something funky with these shortcuts. It's not well documented, what should, and what shouldn't work.

But `Ctrl-Alt+L` works, notification with the world icon:

image

PR #1292, opened 2025-08-29
Respect user defined editor

vladkhard:
Allows to use default editor for menu/setup actions

PR #1465, opened 2025-09-05
Make omarchy-menu respect EDITOR env var

gigor:
Currently there is no way to change what editor is open from omarchy menu. This PR solves it by querying EDITOR variable with fallback to nvim.


spdawson:
Duplicate of #1292?

gigor:
yes. didn't notice

PR #1470, opened 2025-09-05
Adds VSCode theme update

Kasui92:
If VSCode is installed, you can update its theme by installing a dedicated extension and applying it, just like in Omakub.

The `vscode.theme` setting is optional. If it’s not defined, VSCode will fall back to its default dark or light theme.

I’ve added a corresponding VSCode theme for all the default themes, except for `osaka-jade`, which unfortunately has no matching color scheme available.

### DIFF
--- a/themes/catppuccin-latte/vscode.theme
+++ b/themes/catppuccin-latte/vscode.theme
@@ -1,0 +1,2 @@
+vscode[theme]="Catppuccin Latte"
+vscode[extension]="Catppuccin.catppuccin-vsc"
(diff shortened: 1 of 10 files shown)


Kasui92:
> Appreciate this, but I don't want to carry theme support for all the optional editors. I could see this existing as a separate package of some sort, though. We've talked about potentially doing plugins. But by default, themeing is only for the default apps.

I see [this](https://github.com/basecamp/omarchy/pull/719#issuecomment-3185050802) now.
Close.

PR #1474, opened 2025-09-06
add visual feedback for super + mute + fix

cluah:
- show visual feedback (device name, volume icon)
- prevents switching to the same sink, as it prevents a device from reacquiring the audio output when it reconnects
```
[ "$next_sink_name" != "$current_sink_name" ] && {
  pactl set-default-sink "$next_sink_name"
}
```

### DIFF
--- a/bin/omarchy-cmd-audio-switch
+++ b/bin/omarchy-cmd-audio-switch
@@ -24,7 +24,8 @@ fi
 next_sink=$(echo "$sinks" | jq -r ".[$next_sink_index]")
 next_sink_name=$(echo "$next_sink" | jq -r '.name')
 next_sink_description=$(echo "$next_sink" | jq -r '.description')
-next_sink_volume=$(echo "$next_sink" | jq -r '.volume["front-left"].value_percent | sub("%"; "")')
+next_sink_volume=$(echo "$next_sink" | jq -r \
+  '.volume | to_entries[0].value.value_percent | sub("%"; "")')
 next_sink_is_muted=$(echo "$next_sink" | jq -r '.mute')
 
 if [ "$next_sink_is_muted" = "true" ] || [ "$next_sink_volume" -eq 0 ]; then
--- a/bin/omarchy-cmd-audio-switch
+++ b/bin/omarchy-cmd-audio-switch
@@ -1,14 +1,16 @@
 #!/bin/bash
 
+focused_monitor="$(hyprctl monitors -j | jq -r '.[] | select(.focused == true).name')"
+
 sinks=$(pactl -f json list sinks)
 sinks_count=$(echo "$sinks" | jq '. | length')
 
-[ "$sinks_count" -eq 0 ] && {
+if [ "$sinks_count" -eq 0 ]; then
   swayosd-client \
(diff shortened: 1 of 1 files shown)


bin/omarchy-cmd-audio-switch
```diff
@@ -1,19 +1,47 @@
...
 
-# Find all the audio sinks but exit if there are none
-sinks=($(wpctl status | sed -n '/Sinks:/,/Sources:/p' | grep -E '^\s*│\s+\*?\s*[0-9]+\.' | sed -E 's/^[^0-9]*([0-9]+)\..*/\1/'))
-[ ${#sinks[@]} -eq 0 ] && exit 1
+sinks=$(pactl -f json list sinks)
+sinks_count=$(echo "$sinks" | jq '. | length')
 
-# Find current audio sink
-current=$(wpctl status | sed -n '/Sinks:/,/Sources:/p' | grep '^\s*│\s*\*' | sed -E 's/^[^0-9]*([0-9]+)\..*/\1/')
+[ "$sinks_count" -eq 0 ] && {
```
dhh: Use a full conditional here.

bin/omarchy-cmd-audio-switch
```diff
@@ -1,19 +1,47 @@
...
+  swayosd-client \
+    --monitor "$(hyprctl monitors -j | jq -r '.[] | select(.focused == true).name')" \
```
dhh: Let's extract this monitor selection since we're using it more than once.

bin/omarchy-cmd-audio-switch
```diff
@@ -1,19 +1,47 @@
...
+  icon_state="low"
+elif [ "$next_sink_volume" -le 66 ]; then
+  icon_state="medium"
+else
+  icon_state="high"
+fi
+
+next_sink_volume_icon="sink-volume-${icon_state}-symbolic"
+
+[ "$next_sink_name" != "$current_sink_name" ] && {
```
dhh: Use normal conditional.

bin/omarchy-cmd-audio-switch
```diff
@@ -1,19 +1,47 @@
...
+  next_sink_index=$(((current_sink_index + 1) % sinks_count))
+else
+  next_sink_index=0
+fi
 
+next_sink=$(echo "$sinks" | jq -r ".[$next_sink_index]")
+next_sink_name=$(echo "$next_sink" | jq -r '.name')
+next_sink_description=$(echo "$next_sink" | jq -r '.description')
+next_sink_volume=$(echo "$next_sink" | jq -r '.volume["front-left"].value_percent | sub("%"; "")')
+next_sink_is_muted=$(echo "$next_sink" | jq -r '.mute')
```
dhh, line 29: I like showing whether we're muted or not, but I find the custom volume icons too much. Just do mute or not.

cluah:
75% done. but let me advocate for the icons:

- swayosd uses the same icon for mute=true and volume=0
- there is no icon for "not muted"
- 4 loc
- looks nice
- harms nothing

https://github.com/ErikReider/SwayOSD/blob/main/src/server/osd_window.rs#L116-L127

PR #1479, opened 2025-09-06
Add clipboard manager

dz0ny:
This add **clipse** as a clipboard manger that binds to **super+shift+v**

screenshot-2025-09-06_12-51-58

ref: https://github.com/basecamp/omarchy/pull/563

### DIFF
--- a/default/hypr/apps.conf
+++ b/default/hypr/apps.conf
@@ -2,6 +2,7 @@
 source = ~/.local/share/omarchy/default/hypr/apps/1password.conf
 source = ~/.local/share/omarchy/default/hypr/apps/bitwarden.conf
 source = ~/.local/share/omarchy/default/hypr/apps/browser.conf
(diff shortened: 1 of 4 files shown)


abenz1267:
next time look for issues/prs: https://github.com/basecamp/omarchy/issues?q=clipboard

dhh:
Appreciate this, but we still haven't solved the underlying issue with wl-clipboard exposing secrets. When we do, we intend to use Walker's built-in clipboard management.

abenz1267:
> Appreciate this, but we still haven't solved the underlying issue with wl-clipboard exposing secrets. When we do, we intend to use Walker's built-in clipboard management.

i've actually already told ryan: this SHOULD work just fine with walker 1.0.0, because it's explicitly checking for a mimetype and not just the `SENSITIVE` envvar. At least with 1password it works fine.

PR #1482, opened 2025-09-06
feat(bin-env-tools): Update omarchy-menu to use default editor instead nvim

manuelgdlvh:
Currently, the Omarchy menu always uses nvim to open the config file shortcut.
This is strange if you don’t want to install nvim and prefer another editor, like helix in my case.

There is an environment variable EDITOR that already defines the default editor (set to nvim by default).
This change will use the EDITOR variable instead of hardcoding nvim, so users can choose their own text editor.

It may also be a good idea to review other tools that can be made flexible in the same way.
If a tool cannot be replaced (because of how the command works), then the system should warn the user when uninstalling it, showing that some features will break.

### DIFF
--- a/bin/omarchy-menu
+++ b/bin/omarchy-menu
@@ -71,8 +71,8 @@ show_style_menu() {
   *Theme*) show_theme_menu ;;
   *Font*) show_font_menu ;;
   *Background*) omarchy-theme-bg-next ;;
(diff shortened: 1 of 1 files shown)


rez1-dev:
duplicate of https://github.com/basecamp/omarchy/pull/1292

PR #1484, opened 2025-09-06
feat: Notification settings persist across theme changes

pipetogrep:
This PR separates the theme values from common settings in each theme's `mako.ini`. 
Common values are set in `~/config/mako/core.ini` and then included per theme.

The main motivation for this was having a central location to set app specific mako preferences, and having them persist when changing themes.

### DIFF
--- a/config/mako/core.ini
+++ b/config/mako/core.ini
@@ -1,0 +1,17 @@
+anchor=top-right
+default-timeout=5000
+max-icon-size=32
+width=420
+height=110
+
+[app-name=Spotify]
+invisible=1
+
+[mode=do-not-disturb]
+invisible=true
+
+[mode=do-not-disturb app-name=notify-send]
+invisible=false
+
+[urgency=critical]
+default-timeout=0
--- a/install/config/theme.sh
+++ b/install/config/theme.sh
@@ -24,5 +24,4 @@ ln -snf ~/.config/omarchy/current/theme/neovim.lua ~/.config/nvim/lua/plugins/th
 mkdir -p ~/.config/btop/themes
 ln -snf ~/.config/omarchy/current/theme/btop.theme ~/.config/btop/themes/current.theme
 
-mkdir -p ~/.config/mako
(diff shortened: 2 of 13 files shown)


ryanrhughes:
@pipetogrep This is nice to clean up some of those common elements. Looks like we just need a migration for this one though if you don't mind adding that.

dhh:
I can't believe I missed this! I had it in my head that mako didn't have includes? Or maybe it was AI that put that there!

migrations/1756371866.sh
```diff
@@ -0,0 +1,2 @@
+echo "Reload notifications"
```
dhh: We need to copy over the .config/mako/core.ini file in this migration.

dhh:
Given that we can do this, I think we should actually put all of this inside Omarchy. The `~/.config/mako/config` file should just have includes to the files inside of `~/.config/omarchy/current/theme/mako.ini`, then that file can include `~/.local/share/omarchy/default/mako/core.ini`.

pipetogrep:
I've tested this locally and ran the migration to simulate an upgrade, everything seems ok. Would appreciate further testing :)

My test involves seeing if my addition of silencing Plexamp notifications in `core.ini` works after a theme change.

Useful commands:
```
pkill mako || makoctl reload
notify-send "Test Title" "This is a test notification from Mako"
```
[Arch wiki entry on mako's supported configurations](https://man.archlinux.org/man/mako.5.en#GLOBAL_CONFIGURATION_OPTIONS)

migrations/1756371866.sh
```diff
@@ -0,0 +1,3 @@
...
+cp ~/.local/share/omarchy/default/mako/core.ini ~/.config/mako/
```
dhh: I don't think this is right? We are not supposed to copy the core. Only the current theme one.

dhh: Actually, don't even need to copy current theme. That's already in the right place.

PR #1489, opened 2025-09-07
Add example scaling configuration for 24" monitors

c4software:
Hi,

Since some fonts are a bit small at scale 1 with a 2560x1440 monitor, the scalling can be set at a higher scaling than 1. This PR adds an example scaling configuration for that kind of monitors.


dhh:
Appreciate this, but think it's a bit too specific. We can't have for everything in here. Although I would like to do some more config option management in the menu itself.

PR #1491, opened 2025-09-07
Add GLM-4.5 support for Claude Code

0xK1K3:
Z.ai has recently added an endpoint for its latest GLM-4.5 model to ensure compatibility with Claude Code. This pull request adds an installing option within Omarchy-Menu > Install > AI to apply the necessary changes so it can work without having to deal with configurations.

The model is worth a try, it's almost as capable as sonnet-4 but much faster and around 20 times cheaper

image
image


dhh:
Appreciate this, but it's too specific for this kind of installer. Might make a good manual entry in the discussion area though!

emrtnn:
Thanks for the advice! Will open a discussion to share this today 👍

PR #1493, opened 2025-09-07
imv->swayimg

sxl613:
Ref: https://github.com/basecamp/omarchy/issues/1442

`imv` is currently looking for maintainers and doesn't have webp support.
`swayimg` could be a suitable Wayland-friendly alternative.


dhh:
Not loving this one. Don't like the info overlay by default, don't like the odd background repeating in the viewer. We do need to get webp support, but I don't think this is it.

PR #1497, opened 2025-09-07
Add graceful Hyprland exit for shutdown/restart

hjanuschka:
- **New script**: `omarchy-cmd-hyprland-graceful-exit` that:
  - Sends SIGTERM to all applications first (graceful shutdown signal)
  - Waits 3 seconds for applications to close properly
  - Force kills any remaining processes with SIGKILL if needed
  - Then exits Hyprland cleanly

- **Updated omarchy-menu**: Both "Shutdown" and "Restart" options now use the graceful exit script before calling `systemctl poweroff`/`reboot`

## Benefits
- Prevents data loss during system shutdown
- Allows applications to save work and cleanup properly
- Better user experience with proper shutdown flow
- Maintains backward compatibility

### DIFF
--- a/bin/omarchy-cmd-hyprland-graceful-exit
+++ b/bin/omarchy-cmd-hyprland-graceful-exit
@@ -1,6 +1,6 @@
 #!/bin/bash
-# Graceful Hyprland exit script with proper app shutdown delay
-# Ensures applications have time to save data and cleanup before termination
+# Graceful application shutdown script for Hyprland
+# Ensures applications have time to save data and cleanup before system shutdown/restart
 
 echo "Gracefully shutting down applications..."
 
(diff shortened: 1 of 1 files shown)


abenz1267:
This is all handled by uwsm. https://github.com/Vladimir-csp/uwsm?tab=readme-ov-file#concepts-and-features

dhh:
Whatever uwsm has doesn't seem to be working for chrome. But maybe we can just make something targeted for Chrome?

hjanuschka:
@abenz1267 thanks for the hint, i'll take care of it and follow the uwsm trail!

woopstar:
Needs rebase @hjanuschka

dhh:
I ended up doing a crude version where we close all windows then just sleep 1. That has made the chrome issue go away.

PR #324, opened 2025-07-24
Omarchy Monitor Configuration

ryanyogan:
# Omarchy Monitor Settings

A terminal-based interface for managing monitor resolution and scaling in Hyprland/Wayland environments.

## Overview

This tool provides an interactive terminal interface for configuring monitor settings in Hyprland. It automatically detects connected displays and offers intelligent scaling recommendations based on resolution and DPI.

## Features

- Multi-monitor detection and configuration
- Intelligent scaling recommendations
- Real-time configuration updates
- Terminal-adaptive theming
- Demo mode for testing
- Comprehensive test coverage
- Git-based versioning system

##### Current Home
- https://github.com/ryanyogan/omarchy-monitor-settings
- Looking for ideas on future home, official package?  Anyone want to jump into that branch and CR?

## Quick Testing
```
# Install Go Lang if not installed
mise use -g go@latest

# Install the latest version
go install github.com/ryanyogan/omarchy-monitor-settings@latest

# Run from the terminal
omarchy-monitor-settings
```

## Usage Demonstration
screenshot-2025-07-24_12-11-36

Screenshot 2025-07-26 at 2 13 14 PM

#### Multiple Monitors
screenshot-2025-07-24_13-00-02
screenshot-2025-07-24_13-00-19

### 🎮 Controls

- **↑/↓ or k/j** - Navigate menu items
- **Enter/Space** - Select option
- **h or ?** - Show help screen
- **Esc** - Return to main menu
- **q or Ctrl+C** - Quit application

## 🔧 Monitor Detection

The application automatically detects monitors using:
1. **hyprctl** (Hyprland's native tool)
2. **wlr-randr** (Wayland fallback)
3. **Demo data** (for testing/development)

## 📊 Scaling Recommendations

Smart scaling based on resolution:
- **4K+ displays** (3840x2160+): 2x monitor scaling, 0.8x font scaling
- **1440p displays** (2560x1440): 1x monitor scaling, 0.9x font scaling  
- **1080p displays** (1920x1080): 1x monitor scaling, 0.8x font scaling

## Requirements

- Go 1.19 or later
- For production use: Hyprland with `hyprctl` or `wlr-randr`
- For development: Any Unix-like system

### Quick Install (Recommended)

```bash
# Install the latest version
go install github.com/ryanyogan/omarchy-monitor-settings@latest

# Install a specific version
go install github.com/ryanyogan/omarchy-monitor-settings@v1.1.2
```

### From Source

```bash
[...]

PR #1503, opened 2025-09-08
Add TUI to select a resolution

c4software:
Hi,

Simple TUI to allow users a quick resolution change using hyprctl, fzf and gum.

https://github.com/user-attachments/assets/9384fb5f-abb6-4bad-86c9-1d73a4f2ac87

The current selected value is in italics, and the script at the end explains how to persist this new configuration.

### DIFF
--- a/bin/omarchy-menu
+++ b/bin/omarchy-menu
@@ -132,7 +132,7 @@ show_toggle_menu() {
 }
 
 show_setup_menu() {
-  local options="  Audio\n  Wifi\n󰂯  Bluetooth\n󱐋  Power Profile\n󰍹  Monitors"
+  local options="  Audio\n  Wifi\n󰂯  Bluetooth\n󱐋  Power Profile\n󰍹  Resolution\n󰍹  Monitors"
   [ -f ~/.config/hypr/bindings.conf ] && options="$options\n  Keybindings"
   [ -f ~/.config/hypr/input.conf ] && options="$options\n  Input"
   options="$options\n󰱔  DNS\n  Config\n󰈷  Fingerprint\n  Fido2"
@@ -148,6 +148,7 @@ show_setup_menu() {
     blueberry
     ;;
   *Power*) show_setup_power_menu ;;
+  *Resolution*) present_terminal "omarchy-monitor-resolution-picker" ;;
   *Monitors*) edit_in_nvim ~/.config/hypr/monitors.conf ;;
   *Keybindings*) edit_in_nvim ~/.config/hypr/bindings.conf ;;
   *Input*) edit_in_nvim ~/.config/hypr/input.conf ;;
--- a/bin/omarchy-monitor-resolution-picker
+++ b/bin/omarchy-monitor-resolution-picker
@@ -1,0 +1,136 @@
+#!/bin/bash
+
+# Improved script to change a monitor's resolution and scale under Hyprland.
+# Uses fzf for interactive selection and gum only for final confirmation.
+
+set -euo pipefail
+
+# -- Dependency check --
+check_deps() {
+  for cmd in hyprctl jq fzf gum; do
+    if ! command -v "$cmd" &> /dev/null; then
+      echo "Error: command '$cmd' not found." >&2
+      echo "Please install it to use this script." >&2
+      exit 1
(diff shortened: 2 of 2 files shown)


axelfontaine:
Shouldn't the TUI offer to persist it for the user as well?

c4software:
> Shouldn't the TUI offer to persist it for the user as well?

It's easy to do, but if you go in this direction hyprmon is a larger tool but can persist and create profiles. 

The idea behind this script is just to offer a quick way to temporarily change a monitor resolution.

LucaNerlich:
> The idea behind this script is just to offer a quick way to temporarily change a monitor resolution.

but why would I want that? Honest question.

c4software:
> > The idea behind this script is just to offer a quick way to temporarily change a monitor resolution.
> 
> but why would I want that? Honest question.

In my case it is to reduce scaling/resolution when doing presentations (classrooms, meeting etc)

abenz1267:
I don't see the point of this being a TUI instead of just being a submenu.

c4software:
> I don't see the point of this being a TUI instead of just being a submenu.

A submenu is a nice idea, but the user can have multiple screens not just one. And it's just easier to make this evolve outside of the current code of the menu IMO.

abenz1267:
What exactly is the issue with multiple screens?

A bash script to maintain is a bash script to maintain and the benefits from a better UX outweights the cons of having to maintain slightly more. If at all.

Also: persistency is a must.

c4software:
Yes you right I will use Walker to render the list instead of fzf to match the style of the menu.

c4software:
There it is :

https://github.com/user-attachments/assets/f3aabffc-7120-442b-88e8-acf29358f387

abenz1267:
Way better, or not :P. Only minor nitpick: if there's only 1 monitor there's no reason to offer a selection.

c4software:
> Way better, or not :P. Only minor nitpick: if there's only 1 monitor there's no reason to offer a selection.

Nice idea.

c4software:
Done, I also update the PR description accordingly.

abenz1267:
Is there a good reason for the confirmation via gum? should be dropped imo. If you want confirmation that should also be done in walker.

Personally I'd remove the confirmation though.

c4software:
> Is there a good reason for the confirmation via gum? should be dropped imo. If you want confirmation that should also be done in walker.
> 
> Personally I'd remove the confirmation though.

It's just because the confirmation also explain how to persist the configuration, and display « Monitor, Resolution, scale ».

abenz1267:
> > Is there a good reason for the confirmation via gum? should be dropped imo. If you want confirmation that should also be done in walker.
> > Personally I'd remove the confirmation though.
> 
> It's just because the confirmation also explain how to persist the configuration, and display « Monitor, Resolution, scale ».

Yeah, but there's no real reason for it. It's not something system critical and i think it's fair to assume users aren't fully dense. It's also easy to correct, IF someone chose the wrong option. Having a confirmation screen for that... seems overblown.

You can ofc keep it if you feel like it's necessary. But then make it a selection in walker and not annoy the user by popping up a different UI.

For the persistency, i already wrote that above: it should just default to make it persistent. That's what users would expect.

c4software:
I'm updating the PR to handle :

- Saving to the monitors.conf with updating if already present.
- Automatic mode.
- Disable screen.

c4software:
Working great : 

It's now handle :

- Saving the configuration.
- Disable / Enable screen
- Multiple monitor.

https://github.com/user-attachments/assets/5a7174a7-dade-43f6-950d-6f7bf094fdea

c4software:
And I also added preferred resolution and auto scale.

c4software:
I also add options to save profiles (save, load, delete)

https://github.com/user-attachments/assets/04e72ade-dfdc-4d40-85bd-7184b2ae088d

woopstar:
Check #324 or #1764 . All seems related

dhh:
Appreciate the work that went into this, but I think we're actually better off keeping this in the config files, like everything else. I'd rather not start on the road where we seek to put a GUI on top of all the main config files. But do think this could make a good external package for those who want that!

PR #1506, opened 2025-09-08
feat: add support for launching flatpak applications

andresatierf:
### Overview
Add support for user switching to flatpak browsers.

### Motivation
Changing the default browser with `xdg-settings set default-web-browser <some-flatpak>`, causes `omarchy-launch-browser` to try to launch `/usr/bin/flatpak` instead of the proper app.

### Testing
Tested with the following flatpaks
- `app.zen_browser.zen`
- `com.vivaldi.Vivaldi`

### DIFF
--- a/bin/omarchy-launch-browser
+++ b/bin/omarchy-launch-browser
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-exec setsid uwsm app -- $(sed -n 's/^Exec=\([^ ]*\).*/\1/p' {~/.local,~/.nix-profile,/usr}/share/applications/$(xdg-settings get default-web-browser) 2>/dev/null | head -1) ${args[@]} $@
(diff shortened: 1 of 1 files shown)


dhh:
I'm not a big fan of flatpaks. I don't think we need them in Omarchy. We have a wonderful package manager in pacman that gives us access to all the browsers we would ever need. Thanks for the suggestion, though!

andresatierf:
@dhh I understand that, but shouldn't the user be able to choose to change their browser and not lose on some of the features of this distribution? This does not force or encourage the use of flatpaks, simply allows the launch-browser script to not break if the user decides to use them

dhh:
They should definitely be able to change their browser, but they should use the package system we have in Arch to do it. Vivaldi, Zen, everything is already available.

PR #1511, opened 2025-09-08
FEAT: added clambshell mode utils

DavidHoenisch:
## What changed?

This new feature adds in support for "clambshell mode" operation on omarchy. Clambshell mode allows a laptop to be closed while connected to an external monitor without locking the screen, thus, disabling the laptop screen and only using the external monitor. 

## How it works

The omarchy-toggle-clambshell-mode script takes in a single string argument of `open` or `close`. When open is passed to the script the external monitor is treated as an external monitor, allowing both the laptop screen and external monitor to work together. When close is passed to the script the laptop monitor is disabled, and the external monitor is treated as the sole and primary display. 

The monitors.conf file in ~/.config/hypr/monitors.conf file is what is used to call the script. When hyprland detects a state change in monitors, it calls the script with the appropriate arguments.

```plain
bindl=,switch:off:Lid Switch,exec,~/.local/share/omarchy/bin/omarchy-toggle-clambshell-mode open
bindl=,switch:on:Lid Switch,exec,~~/.local/share/omarchy/bin/omarchy-toggle-clambshell-mode close
```

Do note that I have not tested this with multiple external monitors. It reliably has been working for a laptop with a single external monitor.

### DIFF
--- a/bin/omarchy-toggle-clambshell-mode
+++ b/bin/omarchy-toggle-clambshell-mode
@@ -1,0 +1,9 @@
+#!/bin/bash
+if [[ "$(hyprctl monitors)" =~ [[:space:]]DP-[0-9]+ ]]; then # Check if external monitor is connected
+  if [[ $1 == "open" ]]; then
+    hyprctl keyword monitor "eDP-1,preferred,auto,auto" # Enable laptop screen
+  else
+    hyprctl keyword monitor "eDP-1,disable" # Disable laptop screen
+  fi
+fi
+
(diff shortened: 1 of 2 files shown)


netrusov:
setting scale to "auto" is not good because not everyone is using defaults. for me, "auto" (or 2) on 4K laptop display is so small I can barely see any text. same can be said about position while using multiple monitors.

also, there will be issues in laptop + external monitor setup when you change theme or do anything that requires "hyprctl reload". here's what I've found:
- https://github.com/basecamp/omarchy/issues/797
- https://github.com/hyprwm/Hyprland/issues/4090

spoiler: you'll need to persist these settings in a file and source it using hyprland in order for these settings to survive reload

PS: is it clambshell or clamshell?

dhh:
Now fully supported and automated.

PR #1515, opened 2025-09-08
add background image for everforest theme

D-Pagey:
For the issue https://github.com/basecamp/omarchy/issues/617

- [x] Must look good at 6K. Doesn't mean it HAS to be natively 6K, but it can't be like 1440p. Gotta be minimum 4K and not rely on lines that'll look blurry at 6K.
- [x] Must be 2mb or less. That means JPG for anything that's a picture but PNG for anything that's a graphical design.
- [x] Must fit with the tones and style of the theme.

### Screenshot on 4K 32"
screenshot-2025-09-08_22-46-36


dhh:
Not bad, but it has too much blue up top, and I don't think it's quite special enough. Thanks for suggesting it!

PR #1518, opened 2025-09-08
Add DuckDB installation and UI launch scripts

matsonj:
## Adds DuckDB to the available databases

While most other databases run inside of docker, this is not necessary for DuckDB, since it runs in-process.

This script adds the following: 

1) Install DuckDB from the DB installation screen in Omarchy via a helper script.
2) The helper script exists to also add DuckDB to the shell PATH so it can be ran from the CLI with `duckdb` and adds the shortcut `SUPER SHIFT D` for the DuckDB UI as a web app.
3) It also adds a `launch-duckdb-ui` script which allows the DuckDB UI (web app) to be launched in the background with `duckdb -ui`, which then closes it when the browser window is closed.


bin/omarchy-install-duckdb
matsonj, line 1: In general, it's not clear to me if we should modify the `bindings.conf` files of the user as part of this step, but this is my opinionated approach to how I would use DuckDB. We can simplify this and leave it as an exercise to reader instead but imo using the DuckDB web app from a keyboard shortcut is a sublime experience.

dhh:
Appreciate this, but I'd rather we follow the convention already setup with docker DBs. That way you can easily run multiple versions of DuckDB, if you need to. It also makes it easier to uninstall. And to have all DBs running in a similar way that we can inspect via the LazyDocker TUI 👌

mehd-io:
> Appreciate this, but I'd rather we follow the convention already setup with docker DBs. That way you can easily run multiple versions of DuckDB, if you need to. It also makes it easier to uninstall. And to have all DBs running in a similar way that we can inspect via the LazyDocker TUI 👌

DuckDB does run fine in Docker, but since it’s an in-process database, running it as a service in a container is usually overkill for local use. The other databases here (MySQL, Postgres, etc.) follow a client/server model, so the Dockerized setup makes sense to keep versions isolated and expose them as network services.

DuckDB is different: people often use it as a lightweight toolkit for working with local files (CSV, JSON, Parquet) directly from the terminal or scripts, more like `sqlite3` or even `jq` than like a traditional DB server. That’s why I think Docker is not the right default home for it. Maybe there’s a better place for tools of this type?

naxels:
I agree with both, I use DuckDB all the time for quick analytics outside Docker..

Maybe adding both options (and explicitly stating intent on Docker vs app) is a possibility if @dhh agrees?

I disagree on adding the SUPER SHIFT D shortcut. I believe beyond a couple of generic all round global shortcuts, users
should customize themselves.
Neither is there a global shortcut for Postgres/MySQL etc and for example, NeoVim is the default IDE, I'm not going to suggest adding another global keyboard shortcut for VSCode which I prefer,
if I want it, I will update/add it myself.

dhh:
If we're going to run it more like sqlite, it needs to be as slim as sqlite as a package. That means we can't have an installer that sets up all sorts of other stuff on top or changes how the system works in other ways. Then it's just a library. I'm open to that idea too.

But actually, I see that there's already an AUR package for that with `duckdb`. What else do we need here?

My initial invitation was actually based on the misconception that this was more like mysql than it was like sqlite.

matsonj:
Ok lets close this and I will show off how I use it on X, the everything app.

hannes:
We're going to start distributing a DuckDB Docker image soon, will update this issue when it's available.

szarnyasg:
We have an official Docker image now. The README/Dockerfile at at https://github.com/duckdb/duckdb-docker/

It works like so:

```bash
docker run --rm -it -v "$(pwd):/workspace" -w /workspace duckdb/duckdb
```

PR #1519, opened 2025-09-09
Enable and start Ollama as part of its installation

roberto-aguilar:
If Ollama has been installed but not manually started, the following error will be shown when running `ollama run ...`:

```
Error: ollama server not responding - could not connect to ollama server, run 'ollama serve' to start it
```

Rather than manually running `ollama serve` each time, this ensures that the service is automatically installed upon installation and restarts, significally improving the usage experience.

The `bin/omarchy-install-ollama` file was extracted to keep the menu logic as simple as possible.


dhh:
We would need an uninstaller for this too. I'm starting to having more issues with Ollama too. Compared to just using LM Studio. Not sure there's enough value there.

roberto-aguilar:
Thanks for the feedback, @dhh!

> We would need an uninstaller for this too.

I just created the uninstaller, ~however, I'm not sure where to wire it in the UI. Can you please give me some guidance in here?~. Figured it out, see comment below for the integration in the `Remove` menu.

> I'm starting to having more issues with Ollama too. Compared to just using LM Studio. Not sure there's enough value there.

I understand your position regarding Ollama, thanks for sharing it. If you’re open to keeping Ollama around a bit longer, one thing I really like is how it gives us a very Docker-like workflow: you can pull models and then run them right from the terminal. That makes it easy to script, automate, and keep the workflow close to the command line, which is pretty handy for devs. Nonethless, I trust your call on this 🙇

roberto-aguilar:
Done, @dhh, the uninstaller is now wired into the `Remove > AI > Ollama` menu entry that mimics the structure of the `Install > AI > Ollama` menu ✅ 

I updated the PR description with screenshots of it.

ghost:
This commit igoes against Arch etiquette, as installing shouldn't run the installed package.

roberto-aguilar:
Thanks for you feedback, @aspasia-patron!

I'll defer the call to @dhh as doesn't seem like he's taking a purist approach for the sake of it. This solves an actual problem while interacting with Ollama as it not feasible to start the service upon restarts and neither for people to have to research which service to enable and start upon every installation.

As for the etiquette, [there's literally a function](https://github.com/basecamp/omarchy/blob/master/bin/omarchy-menu#L56-L58) that installs and launches several packages like [the VS Code editor](https://github.com/basecamp/omarchy/blob/master/bin/omarchy-menu#L230), [the RetroArch emulator](https://github.com/basecamp/omarchy/blob/master/bin/omarchy-menu#L262), and so on. In regards to systemd services, it happens for [Docker](https://github.com/basecamp/omarchy/blob/master/install/config/docker.sh#L22), [UFW](https://github.com/basecamp/omarchy/blob/master/install/first-run/firewall.sh#L21), and so on.

I do appreciate your feedback, but there are precedents for this in Omarchy and it is something that will greatly improve the experience while interacting with language models in the terminal.

roberto-aguilar:
Friendly ping, @dhh, have you given additional though to this?

PR #1523, opened 2025-09-09
Add an LLM picker to provide access to multiple services with a single keybinding

c4software:
Hi,

I really like the Super+C shortcut, but it's limited to a single service. Since there are multiple interesting LLMs on the market, I’d like to propose this solution.

https://github.com/user-attachments/assets/0a868414-5314-4221-b314-bd8ccd1026d9


dhh:
That is a cool pattern! I don't want it in the main line on the AIs, though. I usually go to a specific model for a specific purpose. And I love having it just on a direct key. But you should write this up under Discussions. On how people could add their own.

c4software:
I will think about how to make this modular, to make it installable easily for users so.

c4software:
I know how. I will extract just the bare minimum and distribute it for users, as for example `omarchy-llm-menu` script.

PR #656, opened 2025-08-11
Configure grouping functionality

reshadman:
- Styles **groupbar** and **group**, the default styling is not optimized and accessible
- Adds bindings for managing groups
- For default styling black with alpha is used, looks o.k on dark mode themes.

PR #1529, opened 2025-09-09
Add tabbed group shortcuts and missing styles

mihailo-obradovic:
I use tabbed groups sometimes, so I made sure the usual border colors are also applied there. A couple of notes:

- I set the base style in `looknfeel.conf`, the same way it was done for general borders. I believe this matches the tokyo night theme in the newer versions of Omarchy, but on mine (installed pre-2.0) those colors were yellow.
- I never used locked groups but I saw in the Hypr docs that the variables were there - I assumed we would be using the same values.
- Set the font of group tabs to Caskaydia to match the rest, but this needs to be changed to be dynamic. I'm not sure how this is done - I tried with the `omarchy-font-set` file but there's probably a better way than adding another file to be modified.
- Added the missing colors (active group border, active locked group border, active tab) to all built-in themes, copying the general active border color.

I noticed that we don't override the inactive border color. The color is quite neutral, so that's probably fine. Modifying it for each theme would require extra work.

EDIT: Before and after screenshots on an active group/tab:
Before - active
After - active

### DIFF
--- a/bin/omarchy-font-set
+++ b/bin/omarchy-font-set
@@ -25,6 +25,27 @@ if [[ -n "$font_name" && "$font_name" != "CNCLD" ]]; then
       -v "$font_name" \
       ~/.config/fontconfig/fonts.conf
 
+    # Update Hyprland group tab font in both default and user config
+    if [[ -f ~/.local/share/omarchy/default/hypr/looknfeel.conf ]]; then
+      sed -i "s/font_family = .*/font_family = $font_name/g" ~/.local/share/omarchy/default/hypr/looknfeel.conf
+    fi
+    if [[ -f ~/.config/hypr/looknfeel.conf ]]; then
+      # Check if groupbar section exists in user config
+      if grep -q "groupbar" ~/.config/hypr/looknfeel.conf; then
+        sed -i "s/font_family = .*/font_family = $font_name/g" ~/.config/hypr/looknfeel.conf
+      else
+        # Add groupbar section if it doesn't exist
+        if ! grep -q "group {" ~/.config/hypr/looknfeel.conf; then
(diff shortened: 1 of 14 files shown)


dhh:
To carry this, we'd need to have groups exposed fully with the right default keybindings etc. Happy to take a look at that if you want to propose something.

mihailo-obradovic:
@dhh, I added all shortcuts that are relevant to groups. My attempt aimed to find the right balance between making them intuitive, easy to press and in line with the existing principles. Since there are so many, I think a small compromise was necessary. In my proposed solution, I used `[` and `]`, but pressing those with `super` requires two hands, so it's not ideal. I actually use a different setup myself because I don't have any shortcuts for apps and instead use letters for some tiling shortcuts. That, however, is not compatible with the base Omarchy setup. So rather than picking letters myself, I kept them like this hoping you would have a better idea.

Also, `super` + `G` would be the primary choice for toggling the group state, but since it's taken I used `super` + `R`.

For dropping a window out of a group I opted for `super` + `end` because it's close to the arrow keys, though it feels a bit weird, as the other keys from that area are unused. And some keyboards may not have end.

The `movewindoworgroup` bindings are undoubtedly useful, but I'm wondering if it's optimal to include them as they can confuse people and aren't really necessary.

I will have another look at the inactive and locked styles tomorrow.

mihailo-obradovic:
Ready for review. Here are some screenshots:

Active window (jade osaka theme):
screenshot-2025-09-22_09-51-55
Window with disabled group locking:
screenshot-2025-09-22_09-52-06
Active group with current tab highlighted:
screenshot-2025-09-22_09-52-23
Active locked group with current tab highlighted:
screenshot-2025-09-22_09-52-32

dhh:
Solved via #656

PR #1535, opened 2025-09-09
Add menu option for updating drive encryption and user passwords

dhh:
If you need to change the LUKS encryption password, it should be easy-peasy. And let's just make it easy to change the user password too.

### DIFF
--- a/bin/omarchy-drive-set-password
+++ b/bin/omarchy-drive-set-password
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+encrypted_drives=$(blkid -t TYPE=crypto_LUKS -o device)
+
+if [[ -n $encrypted_drives ]]; then
+  if [[ $(wc -l <<<"$encrypted_drives") -eq 1 ]]; then
+    drive_to_change="$encrypted_drives"
+  else
+    drive_to_change="$(omarchy-drive-select "$encrypted_drives")"
+  fi
+
+  if [[ -n $drive_to_change ]]; then
+    echo "Changing full-disk encryption password for $drive_to_change"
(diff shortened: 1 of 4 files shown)


bin/omarchy-drive-set-password
```diff
@@ -0,0 +1,21 @@
...
+if [[ -n $encrypted_drives ]]; then
+  if [[ $(wc -l <<<"$encrypted_drives") -eq 1 ]]; then
+    drive_to_change="$encrypted_drives"
+  else
+    drive_to_change="$(omarchy-drive-select "$encrypted_drives")"
+  fi
+
+  if [[ -n $drive_to_change ]]; then
+    echo "Changing full-disk encryption password for $drive_to_change"
+    sudo cryptsetup luksChangeKey --pbkdf argon2id --iter-time 2000 "$drive_to_change"
```
woopstar, line 14: Why hardcode the PBKDF and iter-time? This can cause issues with manual installs.

dhh: Want to make sure that earlier installs that had too high of an iter-time gets reset. What's the conflict? We could consider taking that somewhere else.

woopstar: > Want to make sure that earlier installs that had too high of an iter-time gets reset. What's the conflict? We could consider taking that somewhere else.

Some people (like myself) use a customised setting, and this would revert the personal settings. My Arch currently runs a higher iter-time (5000).

Also, some would run a whole different PBKDF, such as `pbkdf2`, if they use secure boot with GRUB2.

dhh: I think if you run a custom setting, you can also just run the cryptsetup command yourself. You're already very well-versed in the settings. Mostly curious about whether this could break/brick a setup.

PR #1536, opened 2025-09-09
Update omarchy-install-docker-dbs to install postgres v17

LucaNerlich:
The current default / stable postgres version is 17 https://www.postgresql.org/docs/ with v18 being in beta already.

I think the install script should install the latest stable version.


LucaNerlich:
my bad, missed the PR from just earlier today: https://github.com/basecamp/omarchy/pull/1524

PR #1539, opened 2025-09-09
Use background for selected items in Walker menus

nqst:
This PR updates the Walker menu styles so that the selected item is highlighted with a background in addition to the foreground color.

This way, it's much easier to distinguish the active item:

https://github.com/user-attachments/assets/8e99894b-546d-4366-a612-455f64bccc35

Current version (without the background):

https://github.com/user-attachments/assets/ee48dd2c-db62-473b-b284-b84b1973c52f

> [!NOTE]
> I used `@border` because it gives a theme‑appropriate highlight that works well out of the box. I tried `@text`, but it looked a bit dull. I'd prefer to define a new variable (e.g., `@selected-bg`) ideally, but it would require updating all themes.

### DIFF
--- a/default/walker/themes/omarchy-default.toml
+++ b/default/walker/themes/omarchy-default.toml
@@ -2,9 +2,9 @@
 width = 664
 min_width = 664
 max_width = 664
-height = 396
-min_height = 396
-max_height = 396
+height = 386
+min_height = 386
+max_height = 386
 
 # List constraints are critical - without these, the window shrinks when empty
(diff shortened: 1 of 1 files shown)


dhh:
This is pretty good, but I think it's too heavy for my aesthetics. It would make a great local change for folks who want that, though. Please do write it up under Discussions 👌

nqst:
@dhh done — see https://github.com/basecamp/omarchy/discussions/1573

PR #1021, opened 2025-08-23
Enable full mDNS resolution for .local domains

jardahrazdera:
## Summary
• Extends DHH's existing Avahi implementation to support full mDNS resolution
• Configures nsswitch.conf to properly resolve .local domains
• Completes the network discovery setup that was partially implemented

## Problem
The current Avahi setup in printer.sh enables network printer discovery but doesn't configure full mDNS resolution for .local domains. Users need to manually configure nsswitch.conf to resolve hostnames like `raspberrypi.local`.

## Solution
Adds one line to printer.sh that configures nsswitch.conf with proper mDNS resolution order:
```
hosts: mymachines mdns_minimal [NOTFOUND=return] resolve [!UNAVAIL=return] files myhostname dns
```

This follows the same simple approach as existing Omarchy scripts - no backups, direct sed operation.

## Testing
Tested on my system where I previously had to manually configure this. With this change, .local domains resolve correctly after running the printer setup.

Fixes #256

PR #1541, opened 2025-09-09
Enable full mDNS resolution for .local domains

### LINKED ISSUE #256: Add Avahi for mDNS network discovery as core package
Right now the Networks tab in Nautilus is confusingly redundant when you are on a network with other machines because it does not have any native discovery (mDNS) or Bonjour like functionality.

I would like to propose that Avahi is implemented as an out of the box package. It is a lightweight package in keeping with the ethos of the project.

https://wiki.archlinux.org/title/Avahi

jardahrazdera:
Extends existing Avahi setup to enable complete mDNS functionality.
Now handles not just printer discovery but also .local domain resolution,
network shares in Nautilus, and service discovery.

The nss-mdns package was already installed but not configured.
This adds the missing nsswitch.conf configuration.

Includes migration for existing users with Avahi already installed.

Fixes #256

Replaces #1021 (clean diff)


jardahrazdera:
Closing due to diff display issues.

PR #1542, opened 2025-09-09
Better Zoom Webapp Support for meeting links

ATeal:
### Support for exec and mime types. Zoom integration
---
In order to better support webapp zoom:
  - Added optional params for omarchy-webapp-install
    - exec, defaults to omarchy-launch-webapp
    - mimetypes, defaults blank
  - added zoom webapp launcher that parses meeting links and transforms them and calls launch webapp to join meeting links
  - migration to convert existing zoom installs to the new custom handler
  - updated the base installer to call new zoom handler and set mimetypes


ATeal:
I've been thinking on this more and I wonder if I couldn't expand things further to just get  tighter webapp integration all around. Have some sort of mechanism for intercepting links into other web apps (youtube links -> opens youtube webapp, x links -> opens x webapp etc.) Maybe something to explore when I have some more time to mess with it

dhh:
This is awesome!! I've been wanting this all week 😄. I made some tweaks, but I'd love to expand this to all the default web apps we have. And maybe even add some more? Like Microsoft Teams and other apps like that where you get meeting invites for. Let's make it all just work.

PR #1543, opened 2025-09-09
feat(security): Strengthen default Docker database credentials

randall-vx:
### Problem:

Currently, the databases deployed via Docker use insecure configurations:

* MySQL and MariaDB are started with empty root passwords.
* PostgreSQL uses the trust authentication method, which requires no password.
* MongoDB uses weak, hardcoded credentials (admin:admin123).
* Redis is started without password protection.

While convenient for local development, this is a security risk and promotes poor practices.

### Solution:

This PR modifies the script to automatically generate strong, random passwords for each database at installation time.

### Changes Made:

Uses openssl rand -base64 16 to generate robust passwords for MySQL, MariaDB, PostgreSQL, Redis, and MongoDB.

Removed environment variables that allowed for empty passwords or insecure authentication methods.

After each container is deployed, the script now clearly outputs the username (if applicable) and the generated password for the user to save.

Improved the script's output to be more informative and user-friendly.

### Result:

With these changes, Omarchy provides a much more secure and robust initial setup that aligns with security best practices.

### DIFF
--- a/bin/omarchy-install-docker-dbs
+++ b/bin/omarchy-install-docker-dbs
@@ -1,18 +1,66 @@
 #!/bin/bash
+# Description: Install popular databases using Docker with secure passwords.
+set -e
 
 options=("MySQL" "PostgreSQL" "Redis" "MongoDB" "MariaDB")
 choices=$(printf "%s\n" "${options[@]}" | gum choose --no-limit --header "Select databases (space to select, return to install, esc to cancel)") || main_menu
 
 if [[ -n "$choices" ]]; then
+  echo -e "\n Deploying selected databases..."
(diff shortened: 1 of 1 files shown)


bftanase:
Obviously I can't speak for everyone, but I disagree with this take. Since those are local development db servers, I want to be able to connect with minimum fuss.

I fail to see how defining passwords will improve security... If someone gains access to your local system, those containers are the least of your concern.

Also, by default those are listening only on localhost, so someone from outside cannot connect.

ryanrhughes:
These are local DBs intended for local development. We don't really need to give ourselves unnecessary hurdles to jump for things that'll never be exposed anyhow.

PR #481, opened 2025-08-04
Setting a relative path causes path resolution problems with dkms and potentially opens up a security issue.

precision:
I didn't test if this causes issues with the install, but I tracked this down while debugging why DKMS has been failing to install modules.

Using a leading relative path could potentially lead to unintended consequences if someone were to say send you a tarball with a `bin/ls` that just `rm -fr /`.

```
==> dkms install --no-depmod zenpower3/0.2.0 -k 6.15.9-zen1-1-zen
./bin/dkms: line 194: ./bin/date: No such file or directory
./bin/dkms: line 1462: ./bin/mkdir: No such file or directory
mktemp: failed to create directory via template ‘/var/lib/dkms/zenpower3/0.2.0/6.15.9-zen1-1-zen/.tmp_x86_64_XXXXXX’: No such file or directory

Error! Unable to make temporary directory.
./bin/dkms: line 1464: ./bin/mkdir: No such file or directory
./bin/dkms: line 1465: ./bin/cp: No such file or directory
./bin/dkms: line 1468: ./bin/mkdir: No such file or directory
./bin/dkms: line 1470: ./bin/cp: No such file or directory
./bin/dkms: line 1512: ./bin/cp: No such file or directory
```

PR #1544, opened 2025-09-09
Security - Remove relative bin from $PATH

wltechblog:
Having ./bin in your PATH is extremely dangerous, and even moreso when it's the first entry and will override base system tools like `ls` . A different mechanism should be used for in-project executables.

This PR removes the security hole without addressing ways to handle in-project executables.

### DIFF
--- a/default/bash/shell
+++ b/default/bash/shell
@@ -10,5 +10,5 @@ if [[ ! -v BASH_COMPLETION_VERSINFO && -f /usr/share/bash-completion/bash_comple
 fi
 
 # Set complete path
-export PATH="./bin:$HOME/.local/bin:$PATH"
+export PATH="$HOME/.local/bin:$PATH"
 set +h


DavidHoenisch:
my hunch is this is relative path that is assumed to be pointing to ~/.local/share/omarchy/bin

wltechblog:
> my hunch is this is relative path that is assumed to be pointing to ~/.local/share/omarchy/bin

 Per a previous thread, the intended goal was to make it easy to run in-project executables.... I hope I don't actually have to lay out the trivial exploit to get attention on this.

gerardroche:
It's a duplicate of https://github.com/basecamp/omarchy/pull/481

wltechblog:
> It's a duplicate of #481

Partially. #481 is trying to address providing the in-project bin and has completely stalled. The last comment from a developer is over a month ago.

This much simpler PR is only addressing the security issue, which is trivial and obvious and if I'm being honest it's frustrating that it isn't being taken seriously.

ryanrhughes:
This will be handled via #1654

PR #1546, opened 2025-09-09
fix: Add lock file to omarchy-cmd-tzupdate

guilhermetk:
This PR improves the timezone update script by adding:

- A lockfile mechanism to prevent concurrent runs
- Automatic cleanup with trap
- Handling of stale lockfiles (5 min expiry)

Without this, repeated right clicks on the clock may result in multiple Waybar instances:

screenshot-2025-09-09_15-15-42

### DIFF
--- a/bin/omarchy-cmd-tzupdate
+++ b/bin/omarchy-cmd-tzupdate
@@ -1,5 +1,27 @@
 #!/bin/bash
 
+lockfile="/tmp/omarchy-cmd-tzupdate.lock"
+max_age=300
+
+# If lockfile exists, check if the process is still running or if lock is fresh
+if [[ -e "$lockfile" ]]; then
+  pid=$(<"$lockfile")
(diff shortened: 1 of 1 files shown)


abenz1267:
Why is that script even a thing on right-click to begin with?

guilhermetk:
It’s a chef’s addition: https://github.com/basecamp/omarchy/commit/eace29af7d32ef43c84ebc2bd1f8ebeb3aacde5f. Maybe because he travels frequently, but that’s just a guess.

Edit: here it is: https://github.com/basecamp/omarchy/pull/314#issuecomment-3122035598

dhh:
Appreciate this, but it's too heavy. I recognize that it can take a while, though. So maybe we just add a notify-send that says "Updating time zone...".

dhh:
Added it that way. Thanks for the reminder 🙏

PR #1551, opened 2025-09-10
feat: enhance NVIDIA setup script with an option to use pre-compiled packages

MarianoMiguel:
My desktop machine has an Nvidia RTX 3080 (Ampere) GPU. On this machine, I’m never able to complete the installation with any of the Nvidia DKMS packages (module build fails or the entire installer just freezes). The pre-compiled driver path is the only one that consistently works.

The idea behind this PR is to provide some way for the user to choose between DKMS and pre-compiled drivers. The downside of it is that it introduces an interactive prompt in the middle of the setup. It unblocks machines like mine, but I’m not entirely sure this is the ideal or preferred UX.

I validated this on my RTX 3080 system with a custom Omarchy ISO.


dhh:
Not looking to have any questions being asked during install. I think we're also in a good place now with the drivers.

PR #1555, opened 2025-09-10
Update audio switch script to filter available sinks

elpritchos:
Filter audio sinks to include only available ports.

This fixes an issue where if the next sink is `not available` then the sink will not allow change even if there are other available sinks.

### DIFF
--- a/bin/omarchy-cmd-audio-switch
+++ b/bin/omarchy-cmd-audio-switch
@@ -2,7 +2,7 @@
 
 focused_monitor="$(hyprctl monitors -j | jq -r '.[] | select(.focused == true).name')"
 
-sinks=$(pactl -f json list sinks)
+sinks=$(pactl -f json list sinks | jq '[.[] | select([.ports[]? | .availability == "available"] | any)]')
 sinks_count=$(echo "$sinks" | jq '. | length')
 
 if [ "$sinks_count" -eq 0 ]; then


cluah:
i'm using my headphones and right now my laptop's sink says "availability unknown" but if i press super+mute it switches without any problems

elpritchos:
> i'm using my headphones and right now my laptop's sink says "availability unknown" but if i press super+mute it switches without any problems

Yep I've tested what you said locally and you're right so this update isn't ideal.

I'm still experiencing some sinks won't set which blocks cycling though sinks, however it doesn't seem specifically related to availability as like you said 'availability unknown' will set, and also I found a null sink with nothing listed in the port array will also set.

I'm looking into this more but at the moment I'm unsure why this is the case.

cluah:
what does your setup look like? i have a pair of headphones and i see two sinks. i checked both wpctl status and pactl list sinks

elpritchos:
> what does your setup look like? i have a pair of headphones and i see two sinks. i checked both wpctl status and pactl list sinks

I have a number of HDMI sinks from an external monitor, as well as a couple from bluetooth headphones. I've also added a test sink with `pactl load-module module-null-sink`. All appear as SUSPENDED except the one currently set as default, two of the HDMI sinks have a port availability of `not available`, and the test one has an empty port array.

A sink with a port availability of `not available` cannot be set, and what appears to happen is another valid sink is chosen so `pactl set-default-sink` exits without an error status. Therefore when it comes to this original switching script, because the next sink when set ends up defaulting back to the currently set valid sink, I cannot cycle through sinks.

What I'm thinking now is this should instead just filter out `not available` sinks, however I'm uncertain about other possible edge cases. I think there can still be issues cycling correctly if one sink defaults to another as the one that is set is not guaranteed.

elpritchos:
I think the following is better but also not ideal if there is ever a case where one sink defaults to another.

```
sinks=$(pactl -f json list sinks | jq '[.[] | select((.ports | length == 0) or ([.ports[]? | .availability != "not available"] | any))]')
```

cluah:
#1603 it sounds like removing the "not available" sinks should fix the cycling

to recap, since I'm not sure i got it all: an HDMI monitor has multiple audio sinks. and some of those sinks have a "not available" port. and those sinks cannot be set. are you currently able to use that monitor? if so, after removing the "not available" sinks, can you continue using it?

elpritchos:
> #1603 it sounds like removing the "not available" sinks should fix the cycling
> 
> to recap, since I'm not sure i got it all: an HDMI monitor has multiple audio sinks. and some of those sinks have a "not available" port. and those sinks cannot be set. are you currently able to use that monitor? if so, after removing the "not available" sinks, can you continue using it?

Yeah so the monitor has it's own speakers as well as some audio outputs like a headphone jack so I think this is why it shows as multiple possible sinks. One sink has an availability of 'available' which can be set as default and audio can be played through the monitor speakers, the others show as 'not available' as there is nothing plugged in and cannot be set. If the available sink is currently set, attempting to set one of the other  'not available' sinks with `pactl set-default-sink` exits with a 0 exit status, but the default sink is unchanged as it will default back to the 'available' sink.

In terms of this script, in that case the `next_sink` will always be the same but cannot be set, effectively blocking the ability to cycle though valid sinks.

Filtering out the `not available` sinks does fix this for me, however there may be other edge cases where this defaulting to another sink behavior may cause a loop where some valid sinks are missed.

cluah:
so this solves the issue you were facing and doesn't seem to break anything #1603

a couple of things i still don’t fully understand: if you try to set an "not available" sink, does the current default sink change or stay the same? and what happens if you swap `pactl set-default-sink` for `wpctl set-default { id }` (where id refers to .properties.["object.id"])?

```
pactl -f json list sinks | jq '[ .[] | select([.ports[] | .availability == "not available"] | all) | { name: .name, id: .properties["object.id"] } ]'
```

elpritchos:
> so this solves the issue you were facing and doesn't seem to break anything #1603

It does, **but** I can imagine edge cases which could still be an issue.

> a couple of things i still don’t fully understand: if you try to set an "not available" sink, does the current default sink change or stay the same? and what happens if you swap `pactl set-default-sink` for `wpctl set-default { id }` (where id refers to .properties.["object.id"])?

Trying to set a "not available" sink results in another sink being set **or** no change. For example, lets say the current sink is a headphone sink, trying to set a "not available" monitor sink when there is also a separate "available" monitor sink results in that "available" monitor sink being set. Similarly, trying to set a "not available" headphone sink, results in no change if the current default is the "available" headphone sink.

elpritchos:
@cluah I forgot to mention, same behavior when using `wpctl set-default`.

PR #669, opened 2025-08-11
Fix screen recording on Intel graphics

alansikora:
Screen recording with ALT+PrtScreen was failing on Intel graphics due to VAAPI compatibility issues with wl-screenrec. This change adds Intel graphics detection to use wf-recorder instead, which works properly with Intel hardware.

Changes:
- Add Intel graphics detection in omarchy-cmd-screenrecord  
- Use wf-recorder for Intel graphics instead of wl-screenrec
- Preserve existing behavior for NVIDIA (wf-recorder) and AMD/other (wl-screenrec)

PR #1559, opened 2025-09-10
revert: Fix screen recording on Intel graphics #669

eifr:
It looks like https://github.com/basecamp/omarchy/pull/1368 resolves the screen recording issue with Intel GPUs when using wl-screenrec.

I’ve tested it on my machine, and it works as expected.

Changes:
- Removed Intel GPU detection from omarchy-cmd-screenrecord

Sorry for the empty PR before — it was a misclick in the GitHub UI, lol.


dhh:
Can you confirm @eclecticc?

eclecticc:
Just tested on a Framework Laptop 12 with HDA graphics.  Confirmed that #1368 means that #669 is no longer needed.  wl-screenrec uses ~1/2 the memory and ~1/3 the CPU of wf-recorder using the video encode hardware, so makes sense to use it.

eifr:
Btw forgot to mention but this was tested on my framework 13 (12th) 😅
Thanks @eclecticc 🙏

PR #1561, opened 2025-09-10
feat(waybar): add recording indicator

eifr:
### Summary
Adds a custom Waybar module that shows an indicator when screen recording with recording is active.

### Details
- Introduced a `screen-rec-indicator.sh` script in `~/.config/waybar/scripts/`
- Waybar config updated with a `custom/screen-rec-indicator` module
- Indicator shows a red ⏺ while recording, hidden when idle

### Why
This makes it immediately clear when a screen recording is running, helping avoid accidental recordings and improving usability.

### Notes
- Styling for the indicator is defined in `style.css` (`.recording` in red, `.idle` hidden).
- Can be extended later to support clickable start/stop actions.

### DIFF
--- a/config/waybar/config.jsonc
+++ b/config/waybar/config.jsonc
@@ -7,6 +7,7 @@
   "modules-left": ["custom/omarchy", "hyprland/workspaces"],
   "modules-center": ["clock", "custom/update"],
   "modules-right": [
+    "custom/screen-rec-indicator",
     "group/tray-expander",
     "bluetooth",
     "network",
@@ -123,6 +124,12 @@
   "custom/expand-icon": {
     "format": " ",
     "tooltip": false
+  },
+  "custom/screen-rec-indicator": {
+    "on-click": "omarchy-cmd-screenrecord",
+    "exec": "~/.config/waybar/scripts/screen-rec-indicator.sh",
+    "interval": 1,
+    "return-type": "json"
   },
   "tray": {
     "icon-size": 12,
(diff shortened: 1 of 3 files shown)


config/waybar/config.jsonc
```diff
@@ -124,6 +125,12 @@
     "format": " ",
     "tooltip": false
   },
+  "custom/screen-rec-indicator": {
+    "on-click": "omarchy-cmd-screenrecord",
+    "exec": "~/.config/waybar/scripts/screen-rec-indicator.sh",
+    "interval": 1,
```
abenz1267: This is a bad solution.  Waybar supports calling scripts via signals. You don't want polling every second even though you aren't recording at all.

eifr: updated to use signals, wdyt?

config/waybar/style.css
```diff
@@ -61,3 +62,13 @@ tooltip {
 .hidden {
   opacity: 0;
 }
+
+#custom-screen-rec-indicator {
+  min-width: 12px;
+  margin-left: 8.75px;
```
eifr, line 68: i added the indicator in line 40 so no need for the props here.

also, why remove the animation?

dhh: Found the animation distracting. I think all we need is a fixed indicator.

dhh:
This is excellent! Much nicer than having to store state. Great suggestion re: signals, @abenz1267 👌

PR #1568, opened 2025-09-10
Omarchy 2.2.0

dhh:
# What changed?

* Add screen recording indicator to the top bar when recording (clicking will stop) by @eifr + @dhh 
* Fix audio output switch should only move between available outputs by @elpritchos 
* Fix setting gnome dark mode and themes on first install by @dhh
* Fix gum should be available to the install script as soon as possible so we can offer retry on failures by @dhh

### DIFF
--- a/bin/omarchy-cmd-audio-switch
+++ b/bin/omarchy-cmd-audio-switch
@@ -2,7 +2,7 @@
 
 focused_monitor="$(hyprctl monitors -j | jq -r '.[] | select(.focused == true).name')"
 
-sinks=$(pactl -f json list sinks)
+sinks=$(pactl -f json list sinks | jq '[.[] | select([.ports[]? | .availability == "available"] | any)]')
 sinks_count=$(echo "$sinks" | jq '. | length')
 
 if [ "$sinks_count" -eq 0 ]; then
--- a/bin/omarchy-cmd-first-run
+++ b/bin/omarchy-cmd-first-run
@@ -8,5 +8,6 @@ if [[ -f "$FIRST_RUN_MODE" ]]; then
   rm -f "$FIRST_RUN_MODE"
   $OMARCHY_PATH/install/first-run/battery-monitor.sh
   $OMARCHY_PATH/install/first-run/firewall.sh
+  $OMARCHY_PATH/install/first-run/gnome-theme.sh
   sudo rm -f /etc/sudoers.d/first-run
 fi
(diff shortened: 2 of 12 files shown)


brink-lab:
@dhh 
Good that we are deleting workspaces_swipe. It triggers the hyprland error message, now that hyprland doesn't support it anymore. 
Just want to highlight that another contributor has submitted a PR to include the workspaces swipe functionality following the new Gestures "syntax". 
Maybe worth merging it in as part of this release as well to retain the functionality 

Link to PR
https://github.com/basecamp/omarchy/pull/1618

elpritchos:
@dhh
Thanks for the merge however I've since found the audio sink filtering I submitted isn't ideal as not all valid sinks are listed as "available". I've submitted a new PR #1603 to temporarily improve the filter by instead only filtering out "not available" sinks to fix the cycling issues but a more robust solution is probably needed. There has been some further discussion in the original PR #1555 about this.

bin/omarchy-install-steam
```diff
@@ -1,8 +1,5 @@
 #!/bin/bash
 
-echo "Adding multilib repository for 32-bit compatibility"
-sudo sed -i '/^#\s*\[multilib\]/,/^#\s*Include/ s/^#\s*//' /etc/pacman.conf
```
dhh, line 4: @ryanrhughes We'll need a migration to ensure existing installations have multilib on.

michabbb:
How does the upgrade from 2 to 3 work?

jankeesvw:
> How does the upgrade from 2 to 3 work? 

Existing installations can upgrade as per-normal using Update > Omarchy. New installations should download the fresh, all-inclusive ISO before beginning.

stopar:
> > How does the upgrade from 2 to 3 work?
> 
> Existing installations can upgrade as per-normal using Update > Omarchy. New installations should download the fresh, all-inclusive ISO before beginning.

Followed Update > Omarchy, but Omarchy still at V2. 

screenshot-2025-09-17_09-19-30

screenshot-2025-09-17_09-20-10

Kasui92:
> 
> > > How does the upgrade from 2 to 3 work?
> > 
> > Existing installations can upgrade as per-normal using Update > Omarchy. New installations should download the fresh, all-inclusive ISO before beginning.
> 
> Followed Update > Omarchy, but Omarchy still at V2. 
> 


Omarchy v3 hasn't been released yet and is on the dev branch, while you're on master, where v2 is still available. You'd better wait for the official release.

PR #1570, opened 2025-09-10
Activate animation on manual resizing

eddiedale:
Seeing that Omarchy otherwise has animations on by default, and has a mainly keyboard-driven usage, I believe this animation adds delight.

The only downside is mouse-resizing not being as responsive because of the added animation-easing.

Still, I believe the improvements when using it keyboard-driven are more important.


dhh:
Appreciate the suggestion, but I actually do resize quite often with the mouse, and this feels too laggy with that. But good one to add to your own looknfeel if that's not something you use!

PR #1581, opened 2025-09-10
feat: enable vim keys for btop

martinmose:
Omarchy already ships with Vim-like keybindings so let's use vim keys for `btop` as well :pray:

### DIFF
--- a/migrations/1757535736.sh
+++ b/migrations/1757535736.sh
@@ -1,11 +1,9 @@
-#!/bin/bash
-
 echo "Enabling vim keys in btop"
 
 CONFIG_FILE=~/.config/btop/btop.conf
 
-if [ -f "$CONFIG_FILE" ]; then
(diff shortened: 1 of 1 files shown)


martinmose:
Do we need a migration for this? :thinking: 
Something like...
```
#!/bin/bash 

echo "Enabling vim keys in btop" 

CONFIG_FILE=~/.config/btop/btop.conf 
if [ -f "$CONFIG_FILE" ]; then 
    if grep -q "^vim_keys = " "$CONFIG_FILE"; then 
        sed -i 's/^vim_keys = False/vim_keys = True/' "$CONFIG_FILE" 
    fi
fi
```
?

dhh:
Yes, need migration for any files that live in .config 👍

martinmose:
> Yes, need migration for any files that live in .config 👍

Cool! Added now

PR #1587, opened 2025-09-10
fix: assigning correct variable for direct exit in menu navigation

LeoPazEs:
# Summary
Fixes menu navigation behavior from #1100 
# Problem
When opening the theme menu and the power menu, the behaviour of Theme menu → Style menu (was going to main menu) was still happening.
# Solution
Declaring the DIRECT_ACCESS instead of BACK_TO_EXIT


LeoPazEs:
Fix #1582 already pushed to master

PR #1589, opened 2025-09-11
Replace Omarchy-specific themes with themes for both Omarchy and Omakub

rplopes:
This PR proposes an initial extraction of Omarchy and Omakub content into a shared `omacom-core` repo. This should allow keeping a single source of truth for shared configuration, avoiding having to maintain two sources separately. Some tangible benefits:
- Omarchy can get configuration that hasn't been ported from Omakub yet, such as VS Code themes
- Omakub could benefit from Omarchy's faster pace of development without lagging so much behind
- Derivative distros by the community (e.g. a NixOS version) could also benefit from this shared repo

As a first step, this PR extracts themes. This makes it possible in Omarchy to use VS Code themes from Omakub, for instance.

Before this PR:

screenshot-2025-09-10_23-32-55

With this PR:

screenshot-2025-09-11_00-56-49

Future work, out of scope of this PR:
- Changing instructions on how to create an Omarchy theme to call it an Omacom theme, and including the Omakub-specific files such as the VS Code theme
- Back-filling missing theme data, i.e. adding Omakub configs to current Omarchy-only themes, so that they can start being used in Omakub too
- Doing a similar PR in the Omakub repo, to actually benefit from the new Omarchy themes and the new extraction there as well
- Extending the extraction to other parts of the codebase where it would make sense to share code between both distros

**Warning:** This PR fetches `omacom-core` from my own repo, https://github.com/rplopes/omacom-core, as it's not available under the Basecamp GitHub org. It should be changed to a repo under the Basecamp org before merging.

### DIFF
--- a/bin/omarchy-theme-install
+++ b/bin/omarchy-theme-install
@@ -15,7 +15,7 @@ if [ -z "$REPO_URL" ]; then
 fi
 
 THEMES_DIR="$HOME/.config/omarchy/themes"
-THEME_NAME=$(basename "$REPO_URL" .git | sed -E 's/^omarchy-//; s/-theme$//')
+THEME_NAME=$(basename "$REPO_URL" .git | sed -E 's/^(omarchy|omacom)-//; s/-theme$//')
 THEME_PATH="$THEMES_DIR/$THEME_NAME"
 
 # Remove existing theme if present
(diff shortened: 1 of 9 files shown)


rplopes:
In the meantime, @dhh announced Omakub development was winding down, and VS Code theming was added to Omarchy.

Closing this PR, as it didn't get any traction and looks like it doesn't align with the direction the project is headed towards.

PR #1592, opened 2025-09-11
hypr: remove deprecated workspace_swipe setting

jfernandez:
Hyprland 0.51 removed the workspace_swipe setting from gestures configuration. Remove with gestures block and add commented examples in user config template.

Fixes:
image


ryanrhughes:
Merged #1596 since the replacement defaults to false anyway. There was some extra stuff included here in the config that I don't think we really need to have.

Thank you for pinging me with this one though. :smiley:

PR #1599, opened 2025-09-11
workaround to fix error when launching neovim with any catppuccin default theme

dotsilva:
https://github.com/catppuccin/nvim/issues/923

this is a known issue

https://github.com/LazyVim/LazyVim/pull/6354#issuecomment-3202799735

that last link is the source of the workaround provided, i tested it here on my laptop, the error pop doesn't show anymore

### DIFF
--- a/themes/catppuccin/neovim.lua
+++ b/themes/catppuccin/neovim.lua
@@ -9,10 +9,10 @@ return {
 	{
     "akinsho/bufferline.nvim",
     init = function()
-      local bufline = require("catppuccin.groups.integrations.bufferline")
-      function bufline.get()
-        return bufline.get_theme()
-      end
-    end,
-  },
+        local bufline = require("catppuccin.groups.integrations.bufferline")
+        function bufline.get()
+          return bufline.get_theme()
+        end
+      end,
+    },
 }


ramtinJ95:
I actually tried this same approach and got it merged but DHH reverted it, it did not work on his machine it looked like. Because you need to write some migration also for this change, and I may have messed up that bit maybe, although it did work for me with the migration and everything.

dhh:
Yes, last we had this, it broke on my machine. Happy to give it another try though!

dotsilva:
Oh, sorry, i searched only in open PR about it, i forgot to search in closed PR too before trying this PR, then i didn't see this was already tried. It was my fault.

dhh:
Get the same error with this as I did last time:

image

ramtinJ95:
When did you run a lazy update last btw? Is it something you do frequently? Because I can now replicate this, if im on older versions the issue this tries to solve does not exist. Then when adding this fix creates I get the same error. However when on the latest on everything then this fix indeed works. Which makes it a bit harder to actually patch I guess so it works for all versions people might have

ramtinJ95:
Folke seems to be back now btw so this should get fixed shortly upstream anyway 🎆

dotsilva:
https://github.com/LazyVim/LazyVim/releases/tag/v15.0.0

It was patched in the recent Lazyvim release, it is in the release notes, so i think this PR is not necessary anymore: https://github.com/LazyVim/LazyVim/commit/02d0c9aebddb902d9c974a38bb44db9bae4eac6d

PR #1601, opened 2025-09-11
Improve git config defaults and structure

zachasme:
~~This PR aligns `git` configuration with other tools: a bare user config referencing an Omarchy default config.~~

It also adds additional rules, specifically around `git` terminal usage:

- `push.autoSetupRemote = true` Automatically set upstream branch on push
- `diff.algorithm = histogram` Clearer diffs on moved/edited lines
- `diff.colorMoved = plain` Highlight moved blocks in diffs
- `diff.mnemonicPrefix = true` More intuitive refs in diff output
- `commit.verbose = true` Include diff comment in commit message template
- `column.ui = auto` Output in columns when useful (not when piping into grep for example)
- `branch.sort = -committerdate` Sort branches by most recent commit first
- `tag.sort = -version:refname` Sort version numbers as you would expect (default is alphabetically)
- `rerere.enabled = true` Record and reuse conflict resolutions
- `rerere.autoupdate = true` Apply stored conflict resolutions automatically

Obviously a matter of taste, but Omarchy *is* opinionated! I've been running this config for a few months, prompted by the blog post [How Core Git Developers Configure Git](https://blog.gitbutler.com/how-git-core-devs-configure-git#clearly-makes-git-better).

### DIFF
--- a/config/git/config
+++ b/config/git/config
@@ -1,0 +1,28 @@
+# See https://git-scm.com/docs/git-config
+
+[alias]
+	co = checkout
+	br = branch
+	ci = commit
+	st = status
+[init]
+	defaultBranch = master
+[pull]
+	rebase = true 			 # Rebase (instead of merge) on pull
+[push]
+	autoSetupRemote = true   # Automatically set upstream branch on push
+[diff]
+	algorithm = histogram    # Clearer diffs on moved/edited lines
+	colorMoved = plain       # Highlight moved blocks in diffs
+	mnemonicPrefix = true    # More intuitive refs in diff output
+[commit]
+	verbose = true           # Include diff comment in commit message template
+[column]
+	ui = auto 			     # Output in columns when possible
+[branch]
+	sort = -committerdate    # Sort branches by most recent commit first
+[tag]
+	sort = -version:refname  # Sort version numbers as you would expect
+[rerere]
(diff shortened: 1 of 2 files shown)


woopstar:
Needs rebase @zachasme

zachasme:
Rebased @woopstar

config/git/config
```diff
@@ -0,0 +1,2 @@
+[include]
```
dhh: I think let's just expose all the settings directly to the user. These are often fairly personal, so makes sense that they're easy to change individually.

migrations/1757575120.sh
```diff
@@ -0,0 +1,5 @@
+echo "Add git defaults"
+
+omarchy-refresh-config git/config
+
+source $OMARCHY_PATH/install/config/git.sh
```
dhh: Don't think we want to overwrite someone's existing config. So let's just drop the migration and this is a new default going forward only.

foucist:
I was caught by surprise with the `column.ui = auto` setting.  I'm used to occasionally being able to select from the file listing in git status, but the wrapping of file names got in the way of that.

I'm fine with the setting for the other commands, but I disabled it for the status command specifically with `git config --global column.status never`.  Just leaving this message here as a reference in case anyone else stumbles across this issue.

PR #1611, opened 2025-09-11
Minor Firewall Setup Improvements

kashifkhan:
Two changes for ufw setup:
* Use ufw application profile - easier to read, similarly custom app profiles can be added for other apps.
* Enable ufw & reload after all the setup and config is done.


dhh:
Don't understand the point of moving the ufw-docker install bit?

kashifkhan:
This is just organizing the order so that all firewall rules & settings are configured before UFW is enabled at the end. Have seen that pattern in other places.

dhh:
Appreciate this, but prefer to have it exposed as we do now.

PR #1618, opened 2025-09-11
Feat: Workspace swipes back on

adschem:


dhh:
Add as a comment instead. We didn't have it on before. Thanks!

PR #1621, opened 2025-09-12
Add modifications to support Offline ISO

ryanrhughes:
This brings a few features with it aside from just adding support for the ISO so I'll detail those out.

- Add support for offline installation via offline ISO
- Add a new installer look / feel for all 3 install methods
- Embed app icons in Omarchy instead of downloading (compatible with all 3 install methods)
- Add installation logs
- Add `omarchy-upload-log` utility
- Remove some of the multiple builds of initramfs during install
- Add the ability to switch to master if your branch has been deleted
- Add universal trap with useful options (used in ISO and installer)
- Add helpers for install (used in ISO and installer)
- Create a packages registry (used in ISO and installer)
- Install packages for BCM4360 on Apple devices
- Add pre-built lazyvim
- Add pre-built Ruby 3.4.5 + Rails 8.0.2.1

Details about what / why certain things are the way they are can be found in comments in the files changed here to hopefully streamline.

**This has been tested on:**
- ✅ Manual Install on Limine
- ✅ Manual Install on Systemd-boot
- ✅ Online ISO in VM
- ✅ Offline ISO in VM

Offline ISO has been tested and confirmed by many in Discord. This has only minor tweaks from the state of RC2.

- Online ISO: http://iso.omarchy.org/omarchy-online-iso-RC3.iso
- Offline ISO: http://iso.omarchy.org/omarchy-offline-iso-RC3.iso
- Manual Install: `curl -fsSL https://omarchy.org/install | OMARCHY_REF=offline-iso-installer OMARCHY_INSTALL_MODE=online bash`
   - The change to `boot.sh` isn't picked up via URL so we have to pass it in for now.

**ISO build command (from offline-iso branch):**
```
OMARCHY_INSTALLER_REF=offline-iso-installer OMARCHY_CONFIGURATOR_REF=update-styling ./bin/omarchy-iso-make
```


bin/omarchy-cmd-first-run
```diff
@@ -6,8 +6,8 @@ FIRST_RUN_MODE=~/.local/state/omarchy/first-run.mode
 
 if [[ -f "$FIRST_RUN_MODE" ]]; then
   rm -f "$FIRST_RUN_MODE"
-  $OMARCHY_PATH/install/first-run/battery-monitor.sh
-  $OMARCHY_PATH/install/first-run/firewall.sh
-  $OMARCHY_PATH/install/first-run/gnome-theme.sh
+  bash "$OMARCHY_PATH/install/first-run/battery-monitor.sh"
```
ryanrhughes, line 9: Moved away from running anything directly that shouldn't be. This is partially due to how mkarchiso works in that it requires us to match the +x on **every file** individually that needs it because of how they normalize files. 

There's really no reason for these to be directly executable anyway and calling via `bash "file.sh"` still spawns a subshell so we have no net negative here.

bin/omarchy-tui-install
```diff
@@ -20,13 +20,16 @@ fi
...
 ICON_DIR="$HOME/.local/share/applications/icons"
 DESKTOP_FILE="$HOME/.local/share/applications/$APP_NAME.desktop"
-ICON_PATH="$ICON_DIR/$APP_NAME.png"
 
-mkdir -p "$ICON_DIR"
-
-if ! curl -sL -o "$ICON_PATH" "$ICON_URL"; then
-  echo "Error: Failed to download icon."
-  return 1
+if [[ ! "$ICON_URL" =~ ^https?:// ]] && [ -f "$ICON_URL" ]; then
```
ryanrhughes, line 24: We use local file paths now for any apps that ship with Omarchy to make offline compatible.

bin/omarchy-update-available
```diff
@@ -1,5 +1,12 @@
 #!/bin/bash
 
+# See if the branch has been deleted
+current_branch=$(git -C "$OMARCHY_PATH" branch --show-current)
```
ryanrhughes: Added as a bit of a safety to allow users to pivot back to master easily if they were running off of a build branch that eventually gets removed.

bin/omarchy-upload-log
ryanrhughes, line 1: Log uploader that supports uploading install logs (default), as well as useful journalctl logs for diagnosis when users are having issues.

ryanrhughes, line 1: Log uploader that supports uploading install logs (default), as well as useful journalctl logs for diagnosis when users are having issues.

install.sh
```diff
@@ -1,60 +1,122 @@
...
-OMARCHY_INSTALL="$OMARCHY_PATH/install"
+export OMARCHY_PATH="$HOME/.local/share/omarchy"
+export OMARCHY_INSTALL="$OMARCHY_PATH/install"
+export LOG_FILE="/var/log/omarchy-install.log"
 export PATH="$OMARCHY_PATH/bin:$PATH"
 
+# Save original stdout and stderr for trap to use
+exec 3>&1 4>&2
+
+run_logged() {
```
ryanrhughes: This function allows us to run our scripts in a way that just logs a bit of telemetry for us to the logs, as well as all of their output. The log reader that displays on screen reads this output to display it back to the user.

install.sh
```diff
@@ -1,60 +1,122 @@
...
+    echo "[$(date '+%Y-%m-%d %H:%M:%S')] Failed: $script (exit code: $exit_code)" >>"$LOG_FILE"
+  fi
+
+  return $exit_code
+}
+
+#################################
+# Helpers
+#################################
+source $OMARCHY_INSTALL/helpers/ansi-codes.sh
```
ryanrhughes: Added a set of helpers that contain common / reused methods that we find ourselves using in the installer, ISO, and configurator so they're able to be universal and prevent multiple updates.

install.sh
```diff
@@ -1,60 +1,122 @@
...
+source $OMARCHY_INSTALL/helpers/trap-errors.sh
+source $OMARCHY_INSTALL/helpers/tail-log-output.sh
+
+#################################
+# ENV Setup + Precheck
+#################################
+source $OMARCHY_INSTALL/preflight/set-size-vars.sh
+source $OMARCHY_INSTALL/preflight/set-gum-styling.sh
+source $OMARCHY_INSTALL/preflight/ensure-gum-installed.sh
+source $OMARCHY_INSTALL/preflight/guard.sh # Need to be able to prompt
```
ryanrhughes: These get sourced because we need them to effectively be includes in the shell script and not run as a subshell or log. In the case of something like `guard.sh`, users can override so we have to be able to present that option.

install/config/detect-keyboard-layout.sh
ryanrhughes, line 1: If a grep fails, not in a conditional statement, it will trigger pipefail.

install/config/theme.sh
```diff
@@ -3,7 +3,6 @@
 # Set links for Nautilius action icons
 sudo ln -snf /usr/share/icons/Adwaita/symbolic/actions/go-previous-symbolic.svg /usr/share/icons/Yaru/scalable/actions/go-previous-symbolic.svg
 sudo ln -snf /usr/share/icons/Adwaita/symbolic/actions/go-next-symbolic.svg /usr/share/icons/Yaru/scalable/actions/go-next-symbolic.svg
-sudo gtk-update-icon-cache /usr/share/icons/Yaru
```
ryanrhughes, line 6: This doesn't like being ran in chroot

install/first-run/theme.sh
ryanrhughes, line 1: All of these result in a bunch of errors if ran in chroot so moved to first-run instead

install/helpers/ansi-codes.sh
ryanrhughes, line 1: Helpers to make code using ansi codes far more readable.

install/helpers/tail-log-output.sh
ryanrhughes, line 1: This is the function that does the work for the installer log output. It's a simple read, print, move cursor setup so it's dead simple. The pid for the job is set  so it can be killed via `cleanup` or `stop_log_output`.

install/helpers/trap-errors.sh
ryanrhughes, line 1: This is a universal trap used in the installer, as well as the ISO during the archinstall process.

install/login/limine-snapper.sh
```diff
@@ -5,10 +5,25 @@ if command -v limine &>/dev/null; then
 HOOKS=(base udev plymouth keyboard autodetect microcode modconf kms keymap consolefont block encrypt filesystems fsck btrfs-overlayfs)
 EOF
 
-  [[ -f /boot/EFI/limine/limine.conf ]] && EFI=true
+  [[ -f /boot/EFI/limine/limine.conf ]] || [[ -f /boot/EFI/BOOT/limine.conf ]] && EFI=true
```
ryanrhughes, line 6: Fixes an issue where Limine would be missing the cmdline in the config if you installed to USB drive

install/omarchy-base.packages
ryanrhughes, line 1: For the offline ISO in particular, we need a list of packages to build the repo that gets embedded. Previously, this was a separate list in the ISO repo itself but given how close this already is, that doesn't make any sense to me. It makes a lot more sense that we should maintain `omarchy-base.packages` and `omarchy-other.packages` here so that as we merge anything that may install a new package, we can easily add it to the appropriate manifest list and the next ISO build will pick it up automatically.

`omarchy-base.packages` represents the packages that are installed all at the same time in `install/packages.sh` whereas `omarchy-other.packages` are the ones that could be installed later or are conditional in nature.

install/packaging/lazyvim.sh
```diff
@@ -1,8 +1,5 @@
 #!/bin/bash
 
 if [[ ! -d "$HOME/.config/nvim" ]]; then
-  git clone https://github.com/LazyVim/starter ~/.config/nvim
-  cp -R ~/.local/share/omarchy/config/nvim/* ~/.config/nvim/
-  rm -rf ~/.config/nvim/.git
-  echo "vim.opt.relativenumber = false" >>~/.config/nvim/lua/config/options.lua
```
ryanrhughes, line 7: This is handled in our lazyvim repo now

install/post-install/pacman.sh
ryanrhughes, line 1: Set the pacman.conf after install, regardless of what method we took to get here so that everyone ends in the same config.

install/post-install/stop-logs.sh
ryanrhughes, line 1: Stops logs and prints out the telemetry we show on the final screen by just pulling it out of the log. This makes it live after reboot and universal to install methods.

install/preflight/disable-mkinitcpio.sh
ryanrhughes, line 1: Prevent at least some of the unnecessary rebuilds

install/preflight/ensure-gum-installed.sh
ryanrhughes, line 1: No gum, no go. 😎

install/preflight/set-gum-styling.sh
ryanrhughes, line 1: Common gum styling preflight utility to make them all look the same and prevent having to repeat these options in every instance.

bin/omarchy-update-available
dhh: When would they ever be running off a branch that got deleted? This seems like something that's only going to happen to us as we test things. I'd probably rather just have us deal with it directly, then. So we don't have to carry extra guards in the main code.

bin/omarchy-update-git
```diff
@@ -1,5 +1,19 @@
...
+if ! git -C "$OMARCHY_PATH" ls-remote --heads origin "$current_branch" | grep -q .; then
+  echo -e "\e[33mWarning: Branch '$current_branch' no longer exists at origin.\e[0m"
+
+  if gum confirm "Would you like to switch to the main branch (master) before updating?"; then
+    git -C "$OMARCHY_PATH" checkout master
+    echo "Switched to master branch"
+  else
+    echo "Continuing with current branch configuration"
+  fi
+fi
```
dhh: Same with this. I think this is too defensive.

bin/omarchy-update-available
ryanrhughes: The current scenario is this scenario where the offline ISO RC is built from the `offline-iso-installer` branch. We can't cut to `master` right after install in case there are utilities or things that are expected to exist in the test.

Thought was that this gives a path back but I'm also aligned that maybe it's just better to have a more direct utility to handle that scenario if someone installed an RC of some branch and doesn't want to reinstall.

Maybe like an `omarchy-change-branch` utility that allows you to pick from origin branches or something.

install.sh
```diff
@@ -1,60 +1,122 @@
 #!/bin/bash
 
 # Exit immediately if a command exits with a non-zero status
-set -eE
+set -eEo pipefail
```
dhh, line 4: I had some errors that were caused, falsely, by o and pipefail. But I take you've tested that there's nothing?

install.sh
```diff
@@ -1,60 +1,122 @@
...
+#################################
+source $OMARCHY_INSTALL/helpers/ansi-codes.sh
+source $OMARCHY_INSTALL/helpers/chroot.sh
+source $OMARCHY_INSTALL/helpers/logo.sh
+source $OMARCHY_INSTALL/helpers/trap-errors.sh
+source $OMARCHY_INSTALL/helpers/tail-log-output.sh
+
+#################################
+# ENV Setup + Precheck
+#################################
```
dhh: I prefer the shorter commenting style without the banner surround. Don't think it's necessary.

install.sh
```diff
@@ -1,60 +1,122 @@
...
 
 # Exit immediately if a command exits with a non-zero status
-set -eE
+set -eEo pipefail
 
-OMARCHY_PATH="$HOME/.local/share/omarchy"
-OMARCHY_INSTALL="$OMARCHY_PATH/install"
+export OMARCHY_PATH="$HOME/.local/share/omarchy"
+export OMARCHY_INSTALL="$OMARCHY_PATH/install"
+export LOG_FILE="/var/log/omarchy-install.log"
```
dhh: Should scope this as `OMARCHY_INSTALL_LOG_FILE`  to match the others.

install.sh
dhh: Let's extract this into a helper like the rest. I would just call it `log()`.

dhh: Actually maybe just match it with the others: `logged_source`.

install/config/detect-keyboard-layout.sh
```diff
@@ -4,13 +4,12 @@
...
-
-if [[ -n "$layout" ]]; then
+if grep -q '^XKBLAYOUT=' "$conf"; then
+  layout=$(grep '^XKBLAYOUT=' "$conf" | cut -d= -f2 | tr -d '"')
   sed -i "/^[[:space:]]*kb_options *=/i\  kb_layout = $layout" "$hyprconf"
 fi
 
-if [[ -n "$variant" ]]; then
+if grep -q '^XKBVARIANT=' "$conf"; then
+  variant=$(grep '^XKBVARIANT=' "$conf" | cut -d= -f2 | tr -d '"')
```
dhh, line 11: What are these changes for?

install/config/hardware/fix-apple-bcm4360.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+# Look for BCM4360 802.11ac chipset on Apple
+echo "Checking for Apple BCM4360 wireless chipset..."
```
dhh: Would cut this.

install/config/hardware/fix-apple-bcm4360.sh
```diff
@@ -0,0 +1,10 @@
...
+if lspci -nnv | grep -A2 "14e4:43a0" | grep -q "106b:"; then
+  echo "Apple BCM4360 detected, installing broadcom-wl driver..."
+  sudo pacman -S --noconfirm --needed broadcom-wl dkms linux-headers
+else
+  echo "Apple BCM4360 not detected, skipping driver installation"
```
dhh: Don't think we need the negative here either.

install/config/hardware/fix-apple-bcm4360.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+# Look for BCM4360 802.11ac chipset on Apple
+echo "Checking for Apple BCM4360 wireless chipset..."
+if lspci -nnv | grep -A2 "14e4:43a0" | grep -q "106b:"; then
+  echo "Apple BCM4360 detected, installing broadcom-wl driver..."
```
dhh: I think just "Apple BCM4360 detected" is necessary here. The pacman command is already going to output the driver install.

install/config/hardware/nvidia.sh
```diff
@@ -30,9 +30,12 @@ if [ -n "$(lspci | grep -i 'nvidia')" ]; then
     KERNEL_HEADERS="linux-hardened-headers"
   fi
 
-  # Enable multilib repository for 32-bit libraries
-  if ! grep -q "^\[multilib\]" /etc/pacman.conf; then
-    sudo sed -i '/^#\s*\[multilib\]/,/^#\s*Include/ s/^#\s*//' /etc/pacman.conf
+  # Don't do this for offline installs
+  if [ "${OMARCHY_INSTALL_MODE:-offline}" = "online" ]; then
```
dhh: Could we just add the multilib for all? As we talked about, it doesn't really matter to have it there, even if you aren't using it yet.

install/config/detect-keyboard-layout.sh
ryanrhughes: Strange..apparently my comment didn't post on this one. 

A grep not in a conditional that gets piped to something else without a proper fallback or something can actually trigger an error since it exits 1. This wraps them up in an if condition instead of trying to set them straight to a variable to prevent this error from happening.

install/config/mise-ruby.sh
```diff
@@ -5,3 +5,27 @@ mise settings set ruby.ruby_build_opts "CC=gcc-14 CXX=g++-14"
 
 # Trust .ruby-version
 mise settings add idiomatic_version_file_enable_tools ruby
+
+# Check if architecture is x86_64
+if [ "$(uname -m)" != "x86_64" ]; then
```
dhh: Usually prefer to make it a positive conditional. So instead of an early exit, just wrap the work that should be done on x86 in the conditional.

install.sh
ryanrhughes: I went with run_logged to distinguish that these will be executed in a subshell and not actually sourced. Are we OK losing that distinction?

ryanrhughes: It would technically be viable that we could have `logged_source` and `logged_run` that execute logged actions via `source` vs `bash` (subshell)

install/config/theme.sh
dhh: We can move it to first-run.

dhh: Ah, you did.

install/first-run/theme.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
```
dhh: Already have this file on dev as gnome-theme.sh

install/helpers/ansi-codes.sh
```diff
@@ -0,0 +1,33 @@
...
+ANSI_CYAN="\033[36m"
+
+# Functions for backwards compatibility and dynamic values
+ansi_save_cursor() { printf "${ANSI_SAVE_CURSOR}"; }
+ansi_restore_cursor() { printf "${ANSI_RESTORE_CURSOR}"; }
+ansi_clear_line() { printf "${ANSI_CLEAR_LINE}"; }
+ansi_clear_below() { printf "${ANSI_CLEAR_BELOW}"; }
+ansi_move_up() { printf "\033[${1}A"; }
+ansi_move_down() { printf "\033[${1}B"; }
+ansi_move_to_column() { printf "\033[${1}G"; }
```
dhh: None of these moves are currently used anywhere.

install.sh
ryanrhughes: Gone

install/helpers/ansi-codes.sh
```diff
@@ -0,0 +1,33 @@
...
+ANSI_RESTORE_CURSOR="\033[u"
+ANSI_CLEAR_LINE="\033[2K"
+ANSI_CLEAR_BELOW="\033[J"
+ANSI_HIDE_CURSOR="\033[?25l"
+ANSI_SHOW_CURSOR="\033[?25h"
+ANSI_CLEAR_SCREEN="\033[H\033[2J"
+ANSI_RESET="\033[0m"
+
+# ANSI color codes
+ANSI_GRAY="\033[90m"
```
dhh: Only the ANSI_GRAY is used and only in one spot. Not sure worth defining?

install/helpers/logo.sh
```diff
@@ -0,0 +1,7 @@
+clear_logo() {
+  ansi_clear_screen
```
dhh: This function is also only used here. I'd inline to reduce coupling.

install/helpers/tail-log-output.sh
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash
+
+start_log_output() {
+  ansi_save_cursor
+
+  ansi_hide_cursor
```
dhh: These two ansi functions are also only used here. I say we put them here as well then.

install.sh
ryanrhughes: I found that without -o and pipefail, the installer would actually allow some errors to continue and not be trapped. I was able to recreate a number of these silent errors, especially if anything spawned a subshell of any kind. This makes it much more sensitive and stops if there's any unhandled error anywhere in the chain.

install/helpers/trap-errors.sh
```diff
@@ -0,0 +1,155 @@
...
+▀ ▀   ▀ █ ▀▄  ▀▀█▀▀▀█▄▀  
+█▀▀▀▀▀█ ▀█  ▄▀▀ █ ▀ █▄▀██
+█ ███ █ █▀▄▄▀ █▀███▀█▄██▄
+█ ▀▀▀ █ ██  ▀ █▄█ ▄▄▄█▀ █
+▀▀▀▀▀▀▀ ▀ ▀ ▀▀▀  ▀ ▀▀▀▀▀▀'
+
+# Cleanup function - always stop monitoring and restore cursor
+cleanup() {
+  stop_log_output
+  ansi_show_cursor
```
dhh: Only used here also.

install/helpers/trap-errors.sh
```diff
@@ -0,0 +1,155 @@
...
+# Track if we're already handling an error to prevent double-trapping
+ERROR_HANDLING=false
+
+# Error handler
+catch_errors() {
+  # Store exit code immediately before it gets overwritten
+  local exit_code=$?
+
+  # Prevent recursive error handling
+  if [ "$ERROR_HANDLING" = "true" ]; then
```
dhh: Should be the first guard here if we're going to do that.

install/helpers/trap-errors.sh
```diff
@@ -0,0 +1,155 @@
...
+
+  if [ "$error_type" = "interrupted" ]; then
+    gum style --foreground 1 --padding "1 0 1 $PADDING_LEFT" "Omarchy installation interrupted!"
+  else
+    gum style --foreground 1 --padding "1 0 1 $PADDING_LEFT" "Omarchy installation failed!"
+  fi
+
+  LOG_LINES=$(($TERM_HEIGHT - $LOGO_HEIGHT - 35))
+
+  # Show the last lines of the log to help debug
```
dhh: Whenever we have an explaining comment like this, it usually makes sense to extract an explaining function.

install/helpers/trap-errors.sh
```diff
@@ -0,0 +1,155 @@
...
+        truncated_line="$line"
+      fi
+      gum style "$truncated_line"
+    done
+    echo
+  fi
+
+  gum style "This command halted with exit code $exit_code:"
+
+  # Show the failed script if available, otherwise show the command
```
dhh: Here too.

install/packages.sh
```diff
@@ -1,130 +1,17 @@
...
-  yay \
-  zoxide
+# Read packages from the unified package list file
+PACKAGE_LIST="$OMARCHY_INSTALL/omarchy-base.packages"
+
+# Check if package list exists
+if [ ! -f "$PACKAGE_LIST" ]; then
+  echo "Error: Package list not found at $PACKAGE_LIST"
+  exit 1
+fi
```
dhh: Too defensive. We put the list there. It'll be there. (Or stuff afterwards will blow up anyway)./

install/packages.sh
```diff
@@ -1,130 +1,17 @@
...
+
+# Read packages from file (skip comments and empty lines)
+mapfile -t packages < <(grep -v '^#' "$PACKAGE_LIST" | grep -v '^$')
+
+# Install all packages
+sudo pacman -S --noconfirm --needed "${packages[@]}"
+
```
dhh: :scissors:

install/preflight/ensure-gum-installed.sh
```diff
@@ -0,0 +1,5 @@
+# Ensure gum is installed before we use it
```
dhh: On dev we have this in trap-errors.sh.

install/preflight/set-gum-styling.sh
```diff
@@ -0,0 +1,13 @@
+# Tokyo Night theme for gum confirm
```
dhh: Would just turn this into preflight/gum.sh and then do the install needed here too. So you have package and coloring together.

install/reboot.sh
```diff
@@ -1,14 +1,39 @@
...
-echo "You're done! So we're ready to reboot now..." | tte --frame-rate 640 wipe
+echo
+echo "You're done!" | tte --canvas-width 0 --anchor-text c --frame-rate 640 wipe
+
+# Display installation stats if available
+if [ -f "$LOG_FILE" ] && grep -q "Installation Time Summary" "$LOG_FILE" 2>/dev/null; then
+  echo
+  echo
+  # Build the complete stats block and pass to tte all at once
+  STATS=$(tail -n 20 "$LOG_FILE" | sed -n '/=== Installation Time Summary ===/,/^===/p' | grep -E "^(Archinstall|Omarchy|Total):")
```
dhh: Let's just reduce this to a single number.

install/config/hardware/nvidia.sh
ryanrhughes: I think so. Interestingly enough, this is how the ISO has been running the whole time I've been working on it effectively so no negatives noted.

install/preflight/set-gum-styling.sh
ryanrhughes: Like this. Don't know why I didn't think of that.

install.sh
dhh: Ahh, yes, I see. It's in the subshell. Fine as-is for now.

ryanrhughes: Pulled it out to a helper so we have that now.

ryanrhughes:
@mrlarsendk I think I'd just add it to install/config/hardware/fix-apple-bcm4360.sh with the proper conditions and we can rename that to fix-apple-bcm.sh or something that covers both.

PR #1629, opened 2025-09-12
Add health to Waybar battery tooltip

shveeram:
Changed the Waybay config for the battery to have capacity and health in the tool tip:

1. tooltip-format-full (added new)
image

2. tooltip-format-discharging (update)

image

3. tooltip-format-charging (update)

### DIFF
--- a/config/waybar/config.jsonc
+++ b/config/waybar/config.jsonc
@@ -85,8 +85,9 @@
       "default": ["󰁺", "󰁻", "󰁼", "󰁽", "󰁾", "󰁿", "󰂀", "󰂁", "󰂂", "󰁹"]
     },
     "format-full": "󰂅",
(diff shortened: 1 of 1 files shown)


dhh:
That's a really interesting addition, but for me it's too much information on something you can't really act on. I don't personally need to see this health all the time when I just want to see how much battery is left.

leoplanet:
> That's a really interesting addition, but for me it's too much information on something you can't really act on. I don't personally need to see this health all the time when I just want to see how much battery is left.

I agree with you, I personally added it to fastfetch as I am on older hardware where it is something I keep an eye on. 
However, more than the %ge, it would be great to have an estimate of how much time before battery dies. The command upower -b gives this information as 'time to empty: xx hours'. This is more actionable information than battery percentage in my view. Since I don't know how to do it, its up to you.

PR #1637, opened 2025-09-12
feat: configurable window close confirmation

cassiozen:
Allows the user to configure apps that require a confirmation before closing with `Super + W`.

Apps that require confirmation will show an OSD message on the first `Super + W`, then closed on a second `Super + W`.

https://github.com/user-attachments/assets/0dc42d6e-7874-4bb9-ab86-3b4d3663824b


dhh:
That's an interesting idea, but don't like having config options for stuff like this. Also wouldn't use it. Prefer for Super + W to just fire. But this might make a good Manual in the Discussions section for folks who want that.

PR #1646, opened 2025-09-13
feat: Add `SUPER+L` shortcut for session lock

eifr:
This PR introduces a `SUPER+L` keyboard shortcut to quickly lock the current session, providing a familiar and convenient way to secure the workspace.


KornienkoDev:
there is "omarchy-lock-screen" already for that purpose, better to use it

dhh:
I prefer keeping this just to Super + Esc then Return. Fast enough and doesn't require a new binding. But you can just add that even more direct keybinding on your end!

PR #1648, opened 2025-09-13
update to use `$TERMINAL` instead of `alacitty`

eifr:
This update replaces the hardcoded `alacritty` command with the `$TERMINAL` environment variable.

The `$TERMINAL` variable is already set in uwsm, which means users can easily change their preferred terminal (e.g. Ghostty, Kitty, etc.) without modifying configuration files. This makes the setup more flexible and consistent across environments.


antoniodavid (approved):
LGTM

dhh:
The main problem with using the uwsm TERMINAL is that to change it, you have to relaunch Hyprland. But since 0.51, Hyprland supports live reloading of envs. So we should explore whether we can do `env = TERMINAL, alacritty` instead. Then if you change that, it'll immediately apply.

eifr:
@dhh I moved the env variables into env.conf. I didn’t use the Omarchy default config since this one needs to remain editable.

dhh:
Unfortunately hyprland's envs don't reach into all the places, like when things are run by uwsm. So think we do have to put the env in uwsm, and just live with the fact that you'll have to relaunch Hyprland when you change this setting.

dhh:
This is great. Only thing missing is converting the screensaver setup such that it'll work with kitty. And then as soon as Ghostty 1.2 is available as an Arch package, we need that integrated as well.

bin/omarchy-tui-remove
```diff
@@ -6,7 +6,7 @@ DESKTOP_DIR="$HOME/.local/share/applications/"
 if [ "$#" -eq 0 ]; then
   # Find all TUIs
   while IFS= read -r -d '' file; do
-    if grep -q '^Exec=.*alacritty.*-e' "$file"; then
+    if grep -q '^Exec=.*$TERMINAL.*-e' "$file"; then
```
hegjon, line 9: This looks like a bug, `$TERMINAL` will not be expanded in single quotes.

dhh: def a bug, please pr

hegjon: When looking closer it seems like it is not a bug after all, seems like the intentions is actually to grep for the string `$TERMINAL` unexpanded.

PR #1653, opened 2025-09-13
add command to open chromium webapp on localhost

LeonardoTrapani:
Devs work all the time in browses with ports

localhost:3000 etc...

But they are not yet able to use our beautiful chromium webapps

This pr has to goal to be able to easily open a chromium webapp on localhost

# Current state:

just the bin command

I've been trying to integrate it with walker, but it's harder than it seemed, so I wanted confirmation that people want this, and how to implement this. Should we do it with walker, or maybe on the omarchy-menu, or a separate tui?...

Running:
```
omarchy-launch-localhost 3000
```

Has this result:
image

(Currently nothing is running on localhost 3000 on my computer)

### DIFF
--- a/bin/omarchy-launch-localhost
+++ b/bin/omarchy-launch-localhost
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+if [ $# -eq 0 ]; then
+    echo "Usage: omarchy-launch-localhost <port>"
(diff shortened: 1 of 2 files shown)


dhh:
Hmm, I don't think this adds enough on top of `omarchy-launch-webapp` to warrant another setup. Thanks for suggesting, though.

PR #1657, opened 2025-09-14
Add T2 MacBook support

ryanrhughes:


DerAlbertCom:
Is this "just" MacBook or also T2 MacPro 2019?

yuriteixeira:
I would love to see some comments over the changes to better understand why they were needed. Would be great to learn :pray:

PR #730, opened 2025-08-12
Add an ssh-agent systemd user service

r3ap3r2004:
This PR adds an ssh-agent systemd user service and a toggle menu item to easily enable or disable the agent.  It also adds a migration to enable the service by default.

PR #1661, opened 2025-09-14
add ssh-agent support to store SSH key passphrases

sgruendel:
ssh-agent support so git or VSCode passphrases are stored as described in https://www.lorenzobettini.it/2023/09/hyprland-and-ssh-agent/


dhh:
I use 1password as my ssh agent, so this wouldn't be a good fit for a default setup, but I'm open to the idea that we could have a way to add this on as an optional service installation.

sgruendel:
> I use 1password as my ssh agent, so this wouldn't be a good fit for a default setup, but I'm open to the idea that we could have a way to add this on as an optional service installation.

Where would you see it then?  "Install > Service", like Bitwarden?

dhh:
That's what I'm thinking for now?

bin/omarchy-install-ssh-agent
```diff
@@ -0,0 +1,30 @@
...
+WantedBy=default.target
+EOF
+
+# Ensure Hyprland picks up the SSH_AUTH_SOCK variable
+echo 'env = SSH_AUTH_SOCK,$XDG_RUNTIME_DIR/ssh-agent.socket' | tee -a ~/.config/hypr/envs.conf >/dev/null
+
+# Start ssh-agent automatically for user
+systemctl enable --user ssh-agent.service
+
+echo 'SSH Agent has been installed and enabled. Please reboot to apply changes.'
```
sgruendel: Maybe create `$HOME/.local/state/omarchy/reboot-required` and call `omarchy-update-restart` @dhh ?

bin/omarchy-install-ssh-agent
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+
+# add ssh keys to ssh-agent
+mkdir -p ~/.ssh
+echo 'AddKeysToAgent yes' | tee -a ~/.ssh/config >/dev/null
+
+# Create systemd service for ssh-agent
+mkdir -p ~/.config/systemd/user
+tee ~/.config/systemd/user/ssh-agent.service >/dev/null <<EOF
+[Unit]
```
spdawson: Any reason not to use the default-installed `/usr/lib/systemd/user/ssh-agent.service` service configuration?

sgruendel: Ah hadn't seen it's already there @spdawson , but seems to be missing the `DISPLAY=:0` env.  Will try to reuse the existing one then.

Traap:
I cannot use 1password for various reasons on all my machine, so I hook a bash script into my .bashrc  with the following.
sshkeys script.

```bash
#!/bin/bash

sshkeys() {
  local agentName agentCheck=1 cmpTo
  local AGENT_ENV_FILE="${HOME}/.ssh/agent_env"

  case "$OSTYPE" in
    msys*)   agentName="/usr/bin/ssh-agent"; cmpTo=0 ;;
    linux*)  agentName="ssh-agent";          cmpTo=1 ;;
    *)       echo "Unsupported OS: $OSTYPE"; return 1 ;;
  esac

  # Load existing agent vars if the file exists
  if [[ -f "$AGENT_ENV_FILE" ]]; then
    source "$AGENT_ENV_FILE" > /dev/null

    # Check if agent is alive and socket exists
    if ! kill -0 "$SSH_AGENT_PID" 2>/dev/null || [[ ! -S "$SSH_AUTH_SOCK" ]]; then
      rm -f "$AGENT_ENV_FILE"
      unset SSH_AGENT_PID SSH_AUTH_SOCK
    fi
  fi

  # Start a new agent if necessary
  if [[ -z "$SSH_AGENT_PID" || -z "$SSH_AUTH_SOCK" ]]; then
    eval "$(ssh-agent -s)" > /dev/null
    echo "export SSH_AUTH_SOCK=$SSH_AUTH_SOCK" > "$AGENT_ENV_FILE"
    echo "export SSH_AGENT_PID=$SSH_AGENT_PID" >> "$AGENT_ENV_FILE"

    # Add keys only when a new agent is started
    if [[ -d $SSHHOME ]]; then
      for key in "$SSHHOME"/*.pub; do
        if [[ ! -f "$key" ]]; then
          echo "ERROR: $SSHHOME does not have any public keys."
          continue
        else
          echo "ssh-add: ${key%.*}"
          ssh-add "${key%.*}" 2>/dev/null
        fi
      done
    else
      echo "ERROR: SSHHOME is not defined or not a directory."
    fi
  fi
}

sshkeys "$@"

```

bin/omarchy-install-ssh-agent
sgruendel: works with existing service config now

bin/omarchy-menu
```diff
@@ -254,6 +254,7 @@ show_install_service_menu() {
   *Tailscale*) present_terminal omarchy-install-tailscale ;;
   *Bitwarden*) install_and_launch "Bitwarden" "bitwarden bitwarden-cli" "bitwarden" ;;
   *Chromium*) present_terminal omarchy-install-chromium-google-account ;;
+  *SSH Agent*) present_terminal omarchy-install-ssh-agent ;;
```
woopstar: Just tested. Spaces are NOT allowed and result in failure. This needs to be fixed.

woopstar (changes requested):

bin/omarchy-menu
woopstar: You also miss the item in the case:

` case $(menu "Install" "  Dropbox\n  Tailscale\n󰟵  Bitwarden\n  Chromium Account") in`

Which essentially should say "SSH Agent"

sgruendel: Ah you're right of course @woopstar , should have tested the menu, too ...  Fixed now, also added a menu icon so it aligns with the other items.

woopstar:
Does this also fix #730 ?

bin/omarchy-install-ssh-agent
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+# add ssh keys to ssh-agent
```
sudomateo: This doesn't actually add SSH keys to the agent. It just creates the client configuration for it.

gregid:
That would be a nice addition as a service to install - are there blockers for the change? If not just bumping it.

woopstar:
Is this still relevant?

sgruendel:
> Is this still relevant?

I'm still running it like this on my local installations :)

yaubara:
Would be very useful

ggounot:
This PR is missing the following line in `.config/hypr/hyprland.conf`:

```
source = ~/.config/hypr/envs.conf
```

Once that's added, I believe this is a very relevant addition. Not everyone uses 1Password, and ssh-gent is a much more standard approach. Moving forward, we could even add options for other agents like Bitwarden to the install menu.

bin/omarchy-install-ssh-agent
```diff
@@ -0,0 +1,15 @@
...
+mkdir -p ~/.ssh
+echo 'AddKeysToAgent yes' | tee -a ~/.ssh/config >/dev/null
+
+# Ensure Hyprland picks up the SSH_AUTH_SOCK variable
+echo 'env = SSH_AUTH_SOCK,$XDG_RUNTIME_DIR/ssh-agent.socket' | tee -a ~/.config/hypr/envs.conf >/dev/null
```
patrick-boc: Since SSH_AUTH_SOCK has nothing to do with hyprland, it should go to `~/.config/uwsm/env` instead.

bin/omarchy-install-ssh-agent
sgruendel: clarified the comment, thanks for pointing this out

bin/omarchy-install-ssh-agent
sgruendel: fixed, thanks

sgruendel:
> This PR is missing the following line in `.config/hypr/hyprland.conf`:
> 
> ```
> source = ~/.config/hypr/envs.conf
> ```

fixed by usingf `~/.config/uwsm/env` as suggested by @patrick-boc

sgruendel:
> Once that's added, I believe this is a very relevant addition. Not everyone uses 1Password, and ssh-gent is a much more standard approach. Moving forward, we could even add options for other agents like Bitwarden to the install menu.

Yeah I've switched to Bitwarden as SSH agent some weeks ago, just created #5024 to make this available.

bin/omarchy-install-ssh-agent
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+# configure SSH to add keys to ssh-agent automatically
+mkdir -p ~/.ssh
+echo 'AddKeysToAgent yes' | tee -a ~/.ssh/config >/dev/null
+
+# Ensure Hyprland picks up the SSH_AUTH_SOCK variable
+echo 'env = SSH_AUTH_SOCK,$XDG_RUNTIME_DIR/ssh-agent.socket' | tee -a ~/.config/uwsm/env >/dev/null
```
dhh: This is not the correct syntax for uwsm/env, needs to be normal export.

sgruendel: fixed

csfh:
Hi!

Please update the PR description and double check the source branch which will allow us to review it.

PR #1662, opened 2025-09-14
Add preview images to system theme folders

tahayvr:
This PR adds preview images (taken from the Omarchy manual) to the system theme folders.

These are useful when we implement the Walker feature to show previews in the theme-changer UI.

Omarchist also can use these for previews in its UI.


dhh:
Some of these are a bit heavy. Can you try to run them through tinypng.com or similar to make them as small as they can be? Also perhaps shrink the size a bit if possible.

tahayvr:
I think 1500px wide is enough. Right?

I'll go with that if you give me the go.

tahayvr:
Changed image size to 1500px wide and ran through squoosh.app for smaller file size.

dhh:
Let's conventionalize this as tokyo-night/preview.png. Then we can tell theme makers just to have that preview.png file in there too.

PR #1668, opened 2025-09-15
Add Prism Launcher in the Install > Gaming menu

RedsonBr140:
Prism Launcher is an already consolidated, popular, and in many ways better than the original Minecraft Launcher. I created a new menu so that we can keep both.


dhh:
Appreciate the suggestion, but I don't really want two options here for Minecraft. If someone is so into it that they know about Prism, they can use Install > Package to set it up imo.

PR #1671, opened 2025-09-15
allow windows conf + base defaults

JustinMissmahl:


dhh:
We have that now in the looknfeel.conf file.

PR #1673, opened 2025-09-15
Waybar: Simple media controls via built‑in MPRIS

iccodes:
- Add built‑in mpris next to workspaces (/ icons; left=play/pause, right=next).
- Separate commit: Walker toggle (Toggle → “Media Controls”) to enable/disable mpris.
- Migration: enable mpris only if missing; no‑op otherwise.
- Rationale: keep controls simple and native; the toggle helps users who prefer a cleaner bar or want to hide controls for screenshots/recording.

### DIFF
--- a/.nvimlog
+++ b/.nvimlog
@@ -1,0 +1,1 @@
+WRN 2025-09-14T20:30:49.344 ?.112200   server_start:193: Failed to start server: operation not permitted: /run/user/1000/nvim.112200.0
--- a/bin/omarchy-menu
+++ b/bin/omarchy-menu
@@ -139,11 +139,12 @@ show_screenrecord_menu() {
 }
 
 show_toggle_menu() {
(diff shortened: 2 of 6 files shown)


dhh:
Thanks for suggestion, but I prefer to just keep the media controls on the keyboard, and then let folks control it via hotkeys. Having to move the mouse to change the song doesn't seem right to me. But the waybar is free for anyone to customize, of course!

iccodes:
You're welcome. Makes sense to push for a keyboard-centric experience like the rest of Omarchy. What about the visibility benefits of being able to see whats playing regardless of what workspace you're in? I was toying around with a simple "whats playing" tui + keybind but it felt too hidden.

dhh:
I don't love anything that moves in the top bar (besides the clock!), so for the default, I think that'll be a pass. But it's a great option to add in for folks who want that 👌

iccodes:
Yeah I get that distraction issue, trade-offs :man_shrugging: . Happy to use it personally, thanks for the consideration.

PR #1674, opened 2025-09-15
Fix the issue of kernel update detection

ikegam:
Changed to a method that compares boot time with package update time. https://github.com/basecamp/omarchy/issues/1652


ikegam:
This does not work if the system is configured to remove old module files. It seems that not in omarchy-update-restart, but in its caller omarchy-update-perform, we should keep the current Linux version and verify it after yay completes.

PR #1678, opened 2025-09-15
feat: add `format-bluetooth` option to Waybar’s PulseAudio module

eifr:
Adds a `format-bluetooth `option to display a bluetooth icon when audio is routed through a Bluetooth device.

### DIFF
--- a/config/waybar/config.jsonc
+++ b/config/waybar/config.jsonc
@@ -103,6 +103,7 @@
   },
   "pulseaudio": {
     "format": "{icon}",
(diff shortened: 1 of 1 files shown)


dhh:
I had tried this in the past, and I didn't find that the extra icon added something I was looking for. Appreciate the suggestion though!

eifr:
@dhh When using headphones, it's just easier to know that they're automatically selected as the default output.
Nevertheless, I'll just keep it locally :)

PR #1679, opened 2025-09-15
Add fullscreen toggle (monocle) bindings for up and down keys

c4software:
This PR allows the user to put a window in monocle mode by pressing Super ↓ or ↑ to toggle a window in monocle mode Fullscreen but still with a waybar, just like you only have a single window.

Video incoming.

### DIFF
--- a/default/hypr/bindings/tiling.conf
+++ b/default/hypr/bindings/tiling.conf
@@ -7,6 +7,8 @@ bindd = SUPER, J, Toggle split, togglesplit, # dwindle
 bindd = SUPER, P, Pseudo window, pseudo, # dwindle
 bindd = SUPER, V, Toggle floating, togglefloating,
 bindd = SHIFT, F11, Force full screen, fullscreen, 0
+bindd = SUPER, up, Toggle full screen, fullscreen, 1 # Monocle mode
+bindd = SUPER, down, Toggle full screen, fullscreen, 1 # Monocle mode
 
 # Move focus with SUPER + arrow keys
 bindd = SUPER, left, Move focus left, movefocus, l


c4software:
https://github.com/user-attachments/assets/eba93bb8-2e72-49a4-b540-c9b14f1af398

dhh:
When do you use this vs just going fullscreen or moving the window to another workspace? This strikes me a bit as "one option too many".

c4software:
I use it to do a temporary fullscreen to use the whole width of my screen. Less than a full fullscreen (like the F11, no waybar hiding for exemple), for me it's really something I always do (on GNOME, macOS, Windows, etc.).

Edit : And Super ↓/↑ is something widely use.

dhh:
Super up/down is already used to navigate between windows, so that wouldn't fly.

c4software:
> Super up/down is already used to navigate between windows, so that wouldn't fly.

Oh you're right, I rarely split horizontally and I miss that part! It could be moved to Alt (but it will probably cause trouble in some ide), and it wouldn't be as standard as Super.

I really like the monocle view, and I feel it should be included, but like you said too many options kill the option.

guilhermetk:
I use this functionality when I group multiple apps in a single workspace by their category, e.g: chatting apps go to workspace 1 (discord, whatsapp, telegram). Often I need to use the whole screen (preview image, video) so I toggle fullscreen with a personalized shortcut `SUPER + E`. That would bug me sometimes because I forgot that fullscreen was enabled so I added a rule to change the border color when that is the case:

`
windowrule = bordercolor 0xffeb6f92,fullscreen:1
`

https://github.com/user-attachments/assets/9d3b61e3-6a26-4f43-8101-a4268dd67d98

Ps: I'm not advocating for this feature to be added, just showing how I use it

c4software:
Nice idea the dynamic color!

dhh:
Added as `ALT + F11` 👍

PR #1683, opened 2025-09-15
Add a new Ristretto background

bjarneo:
Suggestion for a high res Ristretto wallpaper.
5120x2880

screenshot-2025-09-15_20-31-27
4-ristretto


dhh:
Not bad, but also not brown enough for me for ristretto. Appreciate the pitch though!

PR #1685, opened 2025-09-15
Add a keybinding to open the capture menu

c4software:
Hi,

PR that adds a keybinding (Alt + Shift + 4) to open the capture menu (Screenshot, Video capture).


dhh:
Now that screenrecording opens the menu and we've collapsed screenshots into a single smart selection key, I don't think we need this any more. Thanks for suggesting, though!

PR #1686, opened 2025-09-15
exit if screensaver not in focus

### LINKED ISSUE #1672: Screensaver persists in workspace after using other workspaces
### System details

omarchy 2.1.2 framework 13 12th gen

### What's wrong?

Current behavior: 
Step 1: Start in a given workspace, such as 1
Step 2: Idle long enough for the screensaver to trigger
Step 3: Switch to a different workspace via hotkey, such as 2, and use applications
Step 4: Switch to workspace 1, and the screensaver is still running

Expected behavior:
- -
Step 4: Screensaver in workspace 1 will have been terminated.

manuel1618:
Fixes #1672 

Workspace switching (e.g. super+1) does not send input the the terminal which runs the screensaver. Hyperlandland keybindings never reach the terminal. So the following condition for screensaver exit has been added:

- whenever something else than the terminal which runs the screensaver is in focus -> exit 

This runs well on my local testing (single monitor). Cant speak about multi monitors - would love some support here.

### DIFF
--- a/bin/omarchy-cmd-screensaver
+++ b/bin/omarchy-cmd-screensaver
@@ -21,5 +21,9 @@ while true; do
     if read -n 1 -t 3; then
       exit_screensaver
     fi
+    # Exit if screensaver is not in focus 
+    if ! hyprctl activewindow -j | jq -e '.class == "Screensaver"' >/dev/null 2>&1; then
+      exit_screensaver
+    fi
   done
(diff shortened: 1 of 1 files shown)


manuel1618:
sorry for the duplicate, renamed the branch, that closed the PR - opened a new one  #1687

PR #1690, opened 2025-09-16
feat: Add “Boot Lock” Option to Toggle Menu

5kyguy:
This PR introduces a new Boot Lock option in the Toggle Menu, enabling users to decide whether the system should automatically lock after boot.

Currently, systems with auto-login enabled can be accessed simply by restarting, which poses a security risk — especially for devices used in public spaces.

With this change, users can enable the Boot Lock option, which automatically locks the screen once the system finishes booting. This improves security and protects user sessions on shared or public machines.

### DIFF
--- a/bin/omarchy-lock-screen
+++ b/bin/omarchy-lock-screen
@@ -1,5 +1,13 @@
 #!/bin/bash
 
+# Check if this is being called from autostart (boot lock)
+if [[ "$1" == "--boot" ]]; then
+  # Exit early if boot lock is disabled
+  if ! [[ -f ~/.local/state/omarchy/toggles/boot-lock-enabled ]]; then
+    exit 1
+  fi
+fi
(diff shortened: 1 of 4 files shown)


XY55bqg6iAw8ttP:
Auto login doesn’t trump LUKS, which handles this, though… so seems redundant for the most part, no?

5kyguy:
You're right that LUKS provides strong protection against physical access.

This is useful for users who don't use LUKS encryption. The feature is optional and can be toggled off if not needed (default state is disabled).

dhh:
This unfortunately isn't real security. It doesn't provide what encryption does, and encryption is on by default. If you go off that track to run it without, that's of course totally up to you, it's your computer, but the official recommendation is that you don't, so we won't ship with controls to accommodate that situation.

PR #1693, opened 2025-09-16
bug: figma icon link is broken

woopstar:
The icon link for the Figma webapp is broken. Use homarr-labs instead.


dhh:
On dev we've pulled the Figma file into Omarchy itself 👌

PR #1697, opened 2025-09-16
Allow custom editor, like we do custom terminal

dhh:
Set $EDITOR in `~/.config/uwsm/default`, relaunch Hyprland, and it'll be available in all the places Omarchy asks to open something in an editor.


bin/omarchy-launch-editor
```diff
@@ -0,0 +1,13 @@
...
+
+case "$EDITOR" in
+nvim | vim | nano | micro | hx)
+  exec setsid uwsm app -- "$TERMINAL" -e "$EDITOR" "$@"
+  ;;
+code | codium | subl | gedit | kate | zeditor)
+  exec setsid uwsm app -- "$EDITOR" "$@"
+  ;;
+*)
+  exec setsid uwsm app -- "$TERMINAL" -e nvim "$@"
```
spdawson, line 11: I have my `EDITOR` set to `emacsclient -c`, so will be forced into this branch. I wonder if this might be a bit too clever, with trying to differentiate GUI from TUI editors? Or maybe the fallback should just be the second branch in the above; e.g.

```shell
case "$EDITOR" in
nvim | vim | nano | micro | hx)
  exec setsid uwsm app -- "$TERMINAL" -e "$EDITOR" "$@"
  ;;
*)
  exec setsid uwsm app -- "$EDITOR" "$@"
  ;;
esac
```

So anything we don't definitively know to be TUI is assumed to be a GUI editor.

dhh: I like that idea. But we also have to deal with the case where no $EDITOR is set and make sure it then opens neovim. Can you open a PR for that?

spdawson: Yes, will do 👀

spdawson: Opened #1702

PR #1700, opened 2025-09-16
Fix default animation values

gkurts:
The default looknfeel.conf has a couple of issues in it around animations. This corrects them.

I'm not sure if this is desired or not, so feel free to delete this if it isn't in line with the original goal.

https://github.com/user-attachments/assets/5310697f-e61a-4ccd-93e9-6f16194c951a

### DIFF
--- a/default/hypr/looknfeel.conf
+++ b/default/hypr/looknfeel.conf
@@ -43,7 +43,7 @@ decoration {
 
 # https://wiki.hyprland.org/Configuring/Variables/#animations
 animations {
-    enabled = yes, please :)
+    enabled = true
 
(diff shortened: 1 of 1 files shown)


gkurts:
These can, of course, be completely overridden by the user in their .config/hypr/hyprland.conf file.

dhh:
I don't like workspace shift animations, so that was intentional as it is 👌

gkurts:
No problem! That's honestly more of a user pref anyway. The only thing that caught my eye was the "yes, please :)" value for animations being an invalid value instead of true or false. Maybe it interprets that as false, though? 

edit - it sees the "yes" part as true

PR #1701, opened 2025-09-16
Themeing for ghostty

dhh:
Prepare Omarchy for having Ghostty as the new default terminal. We hit a bug with live theme switching for ghostty 1.2.0, but that should be fixed in 1.2.1.

### DIFF
--- a/bin/omarchy-cmd-screensaver
+++ b/bin/omarchy-cmd-screensaver
@@ -7,7 +7,7 @@ screensaver_in_focus() {
 exit_screensaver() {
   hyprctl keyword cursor:invisible false
   pkill -x tte 2>/dev/null
-  pkill -f "$TERMINAL --class Screensaver" 2>/dev/null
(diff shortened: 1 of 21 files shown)


tahayvr:
Can use color values for the Ghostty config instead of the `theme= <name>` method.

for example Matte Black colors can be set for ghostty like:
# normal colors
palette = 0=#333333
palette = 1=#D35F5F
palette = 2=#FFC107
palette = 3=#B91C1C
palette = 4=#E68E0D
palette = 5=#D35F5F
palette = 6=#BEBEBE
palette = 7=#BEBEBE

# bright colors
palette = 8=#8A8A8D
palette = 9=#B91C1C
palette = 10=#FFC107
palette = 11=#B90A0A
palette = 12=#F59E0B
palette = 13=#B91C1C
palette = 14=#EAEAEA
palette = 15=#FFFFFF

I can do this for other themes as well.

PR #1703, opened 2025-09-16
feat(hyprland): add 3-finger swipe support

dhiman1675:


dhh:
I'm not a fan of having these swipes on by default. I think you should be using your keyboard for these. But all good to add them on your own config!

PR #1706, opened 2025-09-16
Prevent huge diffs with editorconfig

ahowardm:
To prevent huge `diffs` in PRs, we add an `.editorconfig` file to enforce styling for code. Editorconfig [documentation](https://editorconfig.org/)

The first commit adds the style rules and the second one applies them.


dhh:
Got this merged in 👍

PR #1708, opened 2025-09-16
Add 1Password extension policy configuration

hjanuschka:
## Summary
- Added 1Password browser extension policy configuration for Chromium and Brave browsers
- Created migration script to install the policies during system setup

## Changes
- Added `static/onepassword.json` with 1Password extension ID and installation policies
- Created migration `1757683389.sh` to:
  - Create policy directories for both Chromium (`/etc/chromium/policies/managed/`) and Brave (`/etc/brave/policies/managed/`)
  - Copy the policy file to both browser policy directories
  - Set appropriate permissions (755 for directories, 644 for policy files)

## Test plan
- [ ] Run the migration script on a test system
- [ ] Verify directories are created with correct permissions
- [ ] Confirm 1Password extension is automatically installed in Chromium
- [ ] Confirm 1Password extension is automatically installed in Brave


markstos:
Why force 1Password extension to be installed when not everyone uses 1Password?

dhh:
The default install of Omarchy is my setup. It is and should be possible to remove anything and everything after you're up and running, but we start from the base of "this is what I use". You can read more about the principles powering Omarchy here: https://learn.omacom.io/3/omacom/76/omakase-computing

dhh:
@hjanuschka This isn't actually working. The extension is not getting installed on a new setup.

hjanuschka:
you removed the migration, it needs to copy the onepassword.json to /etc/chromium/..... 
have found no other way to put it outside the repo into /etc/

dhh:
This was on a fresh installation. Those files are copied over automatically.

dhh:
My bad. They weren't copied to the right place!

dhh:
Now that it's in the right place, the machine just locks up when you boot chromium:

image

dhh:
It starts on the second launch, but the 1pw extension is not installed.

dhh:
Here is the installer we're using: https://github.com/basecamp/omarchy/blob/dev/install/config/browser-policies.sh

hjanuschka:
gosh! sorry! will try this, can we revert that in dev branch?

using a VM (with iso) is the fastest way to debug/try fresh install?

hjanuschka:
initially when i did the PR -  had browser running, copied file over there, it showed up in chrome://policy and installed extension - strange

dhh:
Yeah, fastest way to test fresh install is by running ./bin/omarchy-iso-make-dev from https://github.com/omacom-io/omarchy-iso

dhh:
Hmm. I can get it working just fine in proxmox. Must be an issue with the qemu flow. Everything is working now 😄

dhh:
Hmm, no, this isn't going to work, since folks can't remove the extension then:

image

PR #1712, opened 2025-09-17
Add auto-changing colors for Neovim

ryanrhughes:
Very MVP oriented approach to changing colors in Neovim.

This taps into the auto-config reload capabilities of Lazy and just utilizes a simple `colorscheme` command to allow Lazy to do the work since it automatically lazy-loads colorschemes.

Because we're registering all of the themes in `all-themes.lua` so that they're installed, we technically could remove them from the individual theme files. I didn't in the event we decide to pivot this approach at all.

The other key thing is that this requires us to effectively prevent any themes that become a part of main line from using any config functions like Catppuccin and Ristretto were using before this change. If we all those to happen, this gets far more complex clearing out highlights and other remnants of shit. -- Not worth it.

### DIFF
--- a/config/nvim/lua/plugins/all-themes.lua
+++ b/config/nvim/lua/plugins/all-themes.lua
@@ -1,0 +1,56 @@
+return {
+	-- Load all theme plugins but don't apply them
+	-- This ensures all colorschemes are available for hot-reloading
+	{
+		"ribru17/bamboo.nvim",
+		lazy = true,
+		priority = 1000,
+	},
+	{
+		"catppuccin/nvim",
+		name = "catppuccin",
+		lazy = true,
+		priority = 1000,
+	},
+	{
+		"sainnhe/everforest",
+		lazy = true,
+		priority = 1000,
+	},
+	{
+		"ellisonleao/gruvbox.nvim",
+		lazy = true,
+		priority = 1000,
+	},
+	{
+		"rebelot/kanagawa.nvim",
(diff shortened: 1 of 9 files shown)


dhh:
I can't this working as-is.

tahayvr:
This is an early version of my PR that was closed. That PR became big as I troubleshooted the problems and factored in support for new installed themes. Don't repeat the road I went down for 2 months 😀

ryanrhughes:
This is working now. I had a version of Pixel still in my local that was kind of shimming the previous non-working state. :face_with_peeking_eye:

ryanrhughes:
@tahayvr as someone who's gone down this road and with a design eye; could be great to have you pull this and see if it looks right to you. Intent is that this is a very narrow spec.

It doesn't handle for hotreloading installed themes, doesn't support using configs to set themes, etc. Just straight colorschemes for themes in our registry.

nickjj:
I have a few questions about this.

- I'm not using Omarchy but I removed these lines and it's still working `lazy = false, priority = 1000,`, maybe it can be lazy?
- What is that transparency file?

nickjj:
If anyone sees this, that transparency file seems to be in a different repo here: https://github.com/omacom-io/omarchy-pkgs/blob/master/pkgbuilds/omarchy-nvim/plugin/after/transparency.lua

Looks like theme related Neovim configuration is in a separate repo.

Issue #1607, opened 2025-09-11
Screenshot not working


Luquatic:
### System details

NVIDIA 5090, Omarchy latest

### What's wrong?

omarchy-cmd-screenshot
config file not found
Error: Unrecognized image file format
Error: Unrecognized image file format

Luquatic:
running `Update > Config > Plymouth` makes it so I can run it once and then it stops working again..

```
omarchy-cmd-screenshot
config file not found

(satty:51784): Adwaita-WARNING **: 16:20:52.495: Using GtkSettings:gtk-application-prefer-dark-theme with libadwaita is unsupported. Please use AdwStyleManager:color-scheme instead.

omarchy-cmd-screenshot
\config file not found
Error: Unrecognized image file format
Error: Unrecognized image file format
```

Egge21M:
I got the same issue on a Framework 13 AMD.

matheorism:
I have the same issue. I can select a region once I press Print Screen , but it doesn't do anything after that.

OS: Omarchy v2.1.2
Kernel: Linux 6.16.6-arch1-1
Hyprland 0.51.0

emourad:
Not exactly the same issue, but screenshot works only the second time.

tuladhar:
Screenshot works for me without issues, no changes made on top of the default. Here's "screenshot" of fastfetch.

<img width="1109" height="647" alt="Image" src="https://github.com/user-attachments/assets/a128a13a-2ba2-494d-9915-755df378f1ac" />

ghost:
Same here. Same errors when running omarchy-cmd-screenshot from terminal as @Luquatic mentioned.

OS: Omarchy v2.1.2
Kernel: Linux 6.16.7-arch1-1
Hyprland 0.51.0

ctarx:
> Not exactly the same issue, but screenshot works only the second time.

same here

ryanrhughes:
By happenstance, I've tracked this down back to the update of adding `--freeze` here. https://github.com/basecamp/omarchy/blob/d09b6aa42a25e71e270ac8ac19e5ce3c17feb878/bin/omarchy-cmd-screenshot#L11C47-L11C55

On my system, it's hit and miss. Sometimes the snapshot works, sometimes it doesn't. Without `--freeze`, I've gone hours with no failures.

alexandru-calinoiu:
[...]

PR #146, opened 2025-07-12
Add omarchy-screenshot

ryanrhughes:
Throughout the conversation yesterday in Discord, @dhh mentioned _"...But would prefer if there was just one hotkey and then you could toggle. Don't recall the mac one I was using, but on space you could toggle between region/window/screen..."_ which got me thinking about ways to get as close to that experience as we can. `omarchy-screenshot` aims to provide that flexibility, and a bit of the best of all worlds.

By default, `omarchy-screenshot` invokes a smart selection (via slurp) that allows you to click a window, a whole display, or drag a region and presents a menu of options of what to do after with the default location being Copy allowing for a **Print Screen, Enter** command sequence to net you whatever you chose to your clipboard.

untitled

### Defaults
The default state is that you get smart selection, a menu after you snap your screenshot, and the edit action opens in Gradia. Those can be changed by setting the mode, passing in an editor to `--editor`, or passing in an action to `--actions`.

The border and overlay colors attempt to extract from `~/.config/omarchy/current/theme/wofi.css` with a fallback if that fails allowing them to better match the active theme but also not fail if something didn't line up.

### Usage Examples
```
  omarchy-screenshot                              # Smart selection, show menu
  omarchy-screenshot --action=copy                # Smart selection, auto-copy
  omarchy-screenshot --editor="satty --filename"  # Use Satty for editing
  omarchy-screenshot smart --action=copy          # Same as above (explicit mode)
  omarchy-screenshot windows                      # Window selection, show menu
```

### Keybinds
Keybinds were kept matched to what the targets are now, but I believe it's feasible to reduce to two but would like feedback prior to making that change since it'll change expected behaviors.

Since the smart mode effectively combines 3, while not locking you into any one path from the outset, we don't need multiple specialized keybinds.

I would recommend we go with this.

```
# Screenshots
bindd = , PRINT,Take a screenshot (smart selection) , exec, ~/.local/share/omarchy/bin/omarchy-screenshot smart
bindd = SHIFT, PRINT,Take a screenshot (fullscreen), exec, ~/.local/share/omarchy/bin/omarchy-screenshot fullscreen

# Eliminate these (changing fullscreen from ctrl + print scr --> shift + print scr)
# bindd = SHIFT, PRINT,Take a screenshot (window), exec, ~/.local/share/omarchy/bin/omarchy-screenshot windows
# bindd = CTRL, PRINT,Take a screenshot (fullscreen), exec, ~/.local/share/omarchy/bin/omarchy-screenshot fullscreen
[...]

PR #1714, opened 2025-09-17
Add smart region selection to omarchy-cmd-snapshot

ryanrhughes:
Since flipping away from the snapshot tool I proposed in #146, there's one piece that I find myself consistently missing and that's that smart region selection. So I've extracted it to re-propose just that bit here.

I hate having to deal with multiple keybinds or menus to make decisions _before_ taking a screenshot when in reality, I may change my mind on the fly. By extracting this one piece, we enable the ability to have Print Screen handle region, window, or output selection without any user prompts, inputs, or finger gymnastics. 

There is technically a decently verbose area we could remove being this bit. It's kind of hefty but serves a decent purpose in that there are occasions where when moving quickly, you'll click a window to snap but technically drag 2-3 pixels and wind up with a 2px snapshot. This insulates against that, and then figures out if the area is too small to be a viable region, what region it was within to determine your intended action.

```
# If the selction area is L * W < 20, we'll assume you were trying to select whichever
# window or output it was inside of to prevent accidental 2px snapshots
if [[ "$SELECTION" =~ ^([0-9]+),([0-9]+)[[:space:]]([0-9]+)x([0-9]+)$ ]]; then
  if (( ${BASH_REMATCH[3]} * ${BASH_REMATCH[4]} < 20 )); then
    click_x="${BASH_REMATCH[1]}"
    click_y="${BASH_REMATCH[2]}"

    while IFS= read -r rect; do
      if [[ "$rect" =~ ^([0-9]+),([0-9]+)[[:space:]]([0-9]+)x([0-9]+) ]]; then
        rect_x="${BASH_REMATCH[1]}"
        rect_y="${BASH_REMATCH[2]}"
        rect_width="${BASH_REMATCH[3]}"
        rect_height="${BASH_REMATCH[4]}"

        if (( click_x >= rect_x && click_x < rect_x+rect_width && click_y >= rect_y && click_y < rect_y+rect_height )); then
          SELECTION="${rect_x},${rect_y} ${rect_width}x${rect_height}"
          break
        fi
      fi
    done <<< "$RECTS"
  fi
fi
```

### DIFF
--- a/bin/omarchy-cmd-screenshot
+++ b/bin/omarchy-cmd-screenshot
@@ -8,7 +8,58 @@ if [[ ! -d "$OUTPUT_DIR" ]]; then
   exit 1
 fi
 
-pkill slurp || hyprshot -m ${1:-region} --raw |
+pkill slurp && exit 0
+
+MODE="${1:-smart}"
+
+get_rectangles() {
+  hyprctl monitors -j | jq -r '.[] | "\(.x),\(.y) \((.width / .scale) | floor)x\((.height / .scale) | floor)"'
+  hyprctl clients -j | jq -r '.[] | "\(.at[0]),\(.at[1]) \(.size[0])x\(.size[1])"'
+}
+
+# Select based on mode
+case "$MODE" in
+  region)
+    SELECTION=$(slurp 2>/dev/null)
+    ;;
+  windows)
+    SELECTION=$(get_rectangles | slurp -r 2>/dev/null)
+    ;;
+  fullscreen)
+    SELECTION=$(hyprctl monitors -j | jq -r '.[] | select(.focused == true) | "\(.x),\(.y) \((.width / .scale) | floor)x\((.height / .scale) | floor)"')
+    ;;
+  smart|*)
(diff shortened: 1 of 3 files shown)


abenz1267:
Personally i'd add in [wayfreeze](https://github.com/Jappie3/wayfreeze) to freeze the screen when taking a screenshot.

I can guarantee you people will ask for this.

sspaeti:
> Personally i'd add in [wayfreeze](https://github.com/Jappie3/wayfreeze) to freeze the screen when taking a screenshot.
> 
> I can guarantee you people will ask for this.

Oh, freezing the screen would be so needed and great 😍 Thanks for mentioning it. It's exactly what I was missing coming from Snagit. Usually the moment the printscreen is hit, is what I want to capture, but usually it's already over as the thing is playing on.

ryanrhughes:
@abenz1267 that's a really good solution! I hadn't added it since Hyprshot and others use Hyprpicker to do the freezing and it has a bug that sometimes causes an endless loop. 

By happenstance, this fixes #1607

PR #1725, opened 2025-09-17
feat: Keybinding to switch to the next theme

SamrudhNelli:
Bind SUPER + SHIFT + CTRL + N to toggle the next theme.

As per the hotkey is concerned LEFT or W could be used as an alt to N.
PS : Previous theme functionality could be added if required (would require a custom omarchy-theme-previous script).


abenz1267:
dunno. i don't think everything has to be a keybind.

samrudhnelli:
> dunno. i don't think everything has to be a keybind.

When we have a keybinding to switch to the next wallpaper, why not to toggle themes?
when you need some new fresh look, imo it would be easier rather than using waybar.

samrudhnelli:
Closing this since the keybinding might be just a personal preference, and not useful for all of us..

PR #1735, opened 2025-09-17
Add Automatic theme changing support for VSCodium

samrudhnelli:
Modified the omarchy-theme-set-vscode file to allow automatic theme changing both in vscodium and vscode if they exist.


samrudhnelli:
Added support when json contains comments..
Credits to : @OmarSkalli #1756

PR #1740, opened 2025-09-17
use helix instead of hx in omarchy-launch-editor

### LINKED ISSUE #1739: Setting default editor to "helix" does not work
### System details

AMD Ryzen 9 9950X, Omarchy 3.0.1

### What's wrong?

We can now set the default editor in `~/.config/uwsm/default`. 

Looking at https://github.com/basecamp/omarchy/blob/master/bin/omarchy-launch-editor, it expects `$EDITOR` to be `nvim | vim | nano | micro | hx`. 

When installing the helix editor through the package installer on Omarchy, the `hx` command is not available, and needs to added to the user's bashrc file under an alias. This bashrc does not get loaded on time to use the `hx` alias. 

I will create a fix to use helix instead of hx.

Cammisuli:
fixes #1739

### DIFF
--- a/bin/omarchy-launch-editor
+++ b/bin/omarchy-launch-editor
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 case "${EDITOR:-nvim}" in
-nvim | vim | nano | micro | hx)
+nvim | vim | nano | micro | helix)
   exec setsid uwsm app -- "$TERMINAL" -e "$EDITOR" "$@"
   ;;
 *)


bin/omarchy-launch-editor
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 case "${EDITOR:-nvim}" in
-nvim | vim | nano | micro | hx)
+nvim | vim | nano | micro | helix)
```
lukehsiao: **suggestion**: I'd support both. `hx` is the default in other distros, meaning `hx` has muscle memory. In my case, for example, I symlink `hx` to `helix` in Arch. I still type `hx`, and setting it to `hx` with that setup works. If this merges, then my setup would break.

Cammisuli: Yea I get it. But the thing is that `hx` doesn't get set unless you add an alias to your bashrc or symlinked it like you did. 

Setting `hx` as the `EDITOR` in uwsm/default will just not work by default on a fresh install.

Cammisuli: It honestly wouldn't hurt to have both, so I can add it either way.

lukehsiao: Yes, I think better to add it even if unused by a fresh install, than mysteriously break things for users on an upgrade.

astralaster: This happens because arch renames the hx binary to helix to avoid collision with hex: https://gitlab.archlinux.org/archlinux/packaging/packages/helix/-/blob/main/PKGBUILD?ref_type=heads#L48

lukehsiao: Yes.

PR #1747, opened 2025-09-18
Float and size Calculator

gkurts:
When tiled, the calculator looks goofy. This will float it and limit its size to make it look a bit better. 

image


gkurts:
Another one of those "delete if it doesn't make sense" PRs. 🤣

PR #1748, opened 2025-09-18
Adds omarchy-cyclenext and bindings for custom cycling

fabiokr:
I like to run some apps in fullscreen mode in the same workspace, but that makes it hard to cycle through tiles when you are looking for a specific one. This changes the ALT+tab binding to temporarily leave fullscreen so you can cycle through the tiles, and upon releasing ALT it will focus on the selected tile and go back to fullscreen.

https://github.com/user-attachments/assets/46ca5ee5-1a54-421c-8b94-51507822abd4

### DIFF
--- a/bin/omarchy-cyclenext
+++ b/bin/omarchy-cyclenext
@@ -1,0 +1,38 @@
+#!/bin/bash
+
+STATE_FILE="/tmp/hypr_alt_tab_state"
+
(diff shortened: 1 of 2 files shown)


fabiokr:
I just started using https://github.com/H3rmt/hyprshell for this as it is much better.

PR #1753, opened 2025-09-18
Include `helix` along with `hx` in `bin/omarchy-launch-editor`

Xoorgh:
When [building from source](https://docs.helix-editor.com/install.html), Helix will use the binary name `hx`. When installing from [some package managers](https://docs.helix-editor.com/package-managers.html), Helix will sometimes use the binary name `hx`, `helix` or both.

This change ensures both may be used.


Xoorgh:
Duplicate of #1740

Issue #1624, opened 2025-09-12
Improve docker daemon.conf defaults to work better for containers using resources over tailscale tunnel on host


cyppe:
### What do you need?

# Docker MTU Optimization for Tailscale Networks

## Problem Statement

The current Docker configuration in `install/config/docker.sh` works well for standard networks, but causes significant performance issues when using **Tailscale VPN** (which Omarchy promotes as the default VPN solution).

### Root Cause
- **Tailscale MTU**: 1280 bytes (due to WireGuard overhead)
- **Docker default MTU**: 1500 bytes
- **Result**: Packet fragmentation causing 2-3x slower performance

### Current Configuration
```json
{
    "log-driver": "json-file",
    "log-opts": { "max-size": "10m", "max-file": "5" },
    "dns": ["172.17.0.1"],
    "bip": "172.17.0.1/16"
}
```

### Proposed Configuration
```json
{
    "log-driver": "json-file",
    "log-opts": { "max-size": "10m", "max-file": "5" },
    "dns": ["172.17.0.1"],
    "bip": "172.17.0.1/16",
    "mtu": 1280,
    "default-network-opts": {
        "bridge": {
            "com.docker.network.driver.mtu": "1280"
        }
    }
}
```

### Performance Impact
- **Before**: Docker containers 2-3x slower over Tailscale
- **After**: Docker containers equal to host performance
- **Improvement**: 50-200% performance gain for Tailscale connections

### Technical Details
1. **`"mtu": 1280`**: Sets MTU for default Docker bridge network
2. **`"default-network-opts"`**: Applies MTU 1280 to ALL Docker networks (including docker-compose)
   - This feature was added in Docker 23.0+ (March 2023)
   - Solves the limitation where daemon MTU only applied to default bridge

### Compatibility
- **Docker Version**: Requires Docker 23.0+ (released March 2023)
- **Backward Compatibility**: Gracefully ignored by older Docker versions
- **Network Impact**: No impact on non-Tailscale networks (1280 MTU works fine everywhere)
- **VPN Compatibility**: Benefits any VPN with MTU < 1500 (common with WireGuard-based VPNs)

## Real-World Testing Results

Tested with:
- **Environment**: Arch Linux + Docker + DDEV + Tailscale
- **Workload**: Database connections, web requests, file transfers
- **Result**: All containers achieved MTU 1280, eliminating packet fragmentation

```bash
# Verification command
for container in $(docker ps --format "{{.Names}}"); do
[...]

PR #1755, opened 2025-09-18
Improve Docker daemon configuration

cyppe:
# Docker MTU Optimization for Tailscale/WireGuard VPNs

This change optimizes Docker daemon configuration to eliminate packet fragmentation when using Tailscale or other WireGuard-based VPNs, providing **50-200% performance improvement** for containerized workloads.

- Sets MTU to 1280 bytes (matching Tailscale's WireGuard overhead)
- Applies MTU to ALL Docker networks including docker-compose via `default-network-opts`
- Increases ulimits (nofile/nproc) from 1024 to 65536 for better container performance

**Fixes:** [#1624 - Improve docker daemon.conf defaults to work better for containers using resources over tailscale tunnel on host](https://github.com/basecamp/omarchy/issues/1624)

### DIFF
--- a/migrations/1758181225.sh
+++ b/migrations/1758181225.sh
@@ -1,0 +1,21 @@
+echo "Update Docker daemon configuration with MTU optimization for Tailscale/WireGuard VPNs and improved ulimits"
+
+# Only update if Docker is installed
+if command -v docker >/dev/null 2>&1; then
+    echo "Updating Docker daemon.json with MTU and ulimits optimization..."
+    
+    # Backup existing daemon.json if it exists (only during migration, not fresh install)
+    if [[ -f /etc/docker/daemon.json ]]; then
+        backup_timestamp=$(date +"%Y%m%d%H%M%S")
+        sudo cp /etc/docker/daemon.json "/etc/docker/daemon.json.bak.${backup_timestamp}"
+        echo "Backed up existing daemon.json to daemon.json.bak.${backup_timestamp}"
+    fi
+    
+    # Run the Docker configuration script to apply the updated settings
+    bash $OMARCHY_PATH/install/config/docker.sh
+    
+    echo "Docker daemon configuration updated. Restart Docker service to apply changes:"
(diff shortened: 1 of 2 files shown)


dhh:
cc @jeremy

migrations/1758181225.sh
```diff
@@ -0,0 +1,21 @@
...
+# Only update if Docker is installed
+if command -v docker >/dev/null 2>&1; then
+    echo "Updating Docker daemon.json with MTU and ulimits optimization..."
+    
+    # Backup existing daemon.json if it exists (only during migration, not fresh install)
+    if [[ -f /etc/docker/daemon.json ]]; then
+        backup_timestamp=$(date +"%Y%m%d%H%M%S")
+        sudo cp /etc/docker/daemon.json "/etc/docker/daemon.json.bak.${backup_timestamp}"
+        echo "Backed up existing daemon.json to daemon.json.bak.${backup_timestamp}"
+    fi
```
jeremy, line 12: Would be nice if we could backup the original daemon.json but preserve its config, e.g. using jq to apply the new settings.

migrations/1758181225.sh
```diff
@@ -0,0 +1,21 @@
...
+    
+    # Run the Docker configuration script to apply the updated settings
+    bash $OMARCHY_PATH/install/config/docker.sh
```
jeremy, line 15: Safe to re-run without clobbering stuff?

migrations/1758181225.sh
```diff
@@ -0,0 +1,21 @@
+echo "Update Docker daemon configuration with MTU optimization for Tailscale/WireGuard VPNs and improved ulimits"
+
+# Only update if Docker is installed
+if command -v docker >/dev/null 2>&1; then
```
jeremy, line 4: Should prob update the config if it's present, but only re-run the config script if Docker is installed.

jeremy (commented):
Gave this a spin locally and it's working fine for me.

PR #1756, opened 2025-09-18
Improve theme-set-vscode to support jsonc and respect symlinks

OmarSkalli:
Improvements to the VS Code theme setup script to support

* Config files with comments, e.g. https://github.com/basecamp/omarchy/issues/1733
* Maintain symlinks 

Jsonc is tricky to parse, so there are edge cases that might not work -- but they should be fairly uncommon.


OmarSkalli:
@dhh apologies for the ping on this PR, but there’s a bit of time sensitivity as changing themes right now potentially destroys VS Code configs for some users.

PR #1757, opened 2025-09-18
Faster pacman calls: added --needed flag where it was missing for consistency

koenvaneijk:
My first pull request to Omarchy 🎉! Thank you for this project and this gift to the community.

The pacman `--needed` command/flag tells the Arch Linux package manager to only install packages if they are not already present or if the installed version is older than the one available. This should speed things up in cases where the latest package is already installed. 

The flag `--needed` was already present in quite a few places, but not consistently throughout all scripts.

Let me know what you think and hoping I can contribute more in the future.

### DIFF
--- a/bin/omarchy-install-steam
+++ b/bin/omarchy-install-steam
@@ -1,5 +1,5 @@
 #!/bin/bash
 
 echo "Now pick dependencies matching your graphics card"
-sudo pacman -Syu --noconfirm steam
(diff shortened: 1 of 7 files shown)


dhh:
Thanks for the suggestion, but I don't think these actually end up making sense. You've added them to mostly the index update calls. Or to places where we know whether it's needed or not. So going to pass at this time.

PR #1758, opened 2025-09-18
Performance tuning

cyppe:
# Add Conservative System Performance Optimizations

Adds essential performance tuning for development workloads - fixes common Docker, VS Code, and build issues.

## What it improves:

- **Fixes file watcher errors** in VS Code/webpack (`fs.inotify.max_user_watches=524288`)
- **Prevents PID exhaustion** with containers (`kernel.pid_max=4194304`)
- **Better network performance** for Docker/databases (`somaxconn=65536`, `tcp_fastopen=3`)
- **Proper file descriptor limits** for development tools (`DefaultLimitNOFILE=65536:524288`)
- **Less aggressive swapping** - reduces swap usage when RAM is available (`vm.swappiness=30` vs default 60)

## Implementation:
- Appends settings to existing `/etc/sysctl.d/99-sysctl.conf` (no new files)
- Compatible with existing SSH flakiness configuration
- Conservative settings safe for all system types
- Auto-applied on fresh installs and via migration for existing users

### DIFF
--- a/default/bash/init
+++ b/default/bash/init
@@ -18,3 +18,10 @@ if command -v fzf &> /dev/null; then
     source /usr/share/fzf/key-bindings.bash
   fi
 fi
+
+# Performance tuning: Fix for Arch Linux systemd user session file descriptor limit inheritance bug
+# Arch Linux systemd user sessions don't properly inherit global ulimit settings
+# This shell-level workaround ensures file descriptor limits are set correctly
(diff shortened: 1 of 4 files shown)


cyppe:
Will update this PR tomorrow as I see now that we have 

install/config/ssh-flakiness.sh
install/config/all.sh

So will make sure this update aligns with it.

cyppe:
Updated.

PR #346, opened 2025-07-26
Auto-load themes for neovim

tahayvr:
This PR adds an automatic reload feature for NeoVim themes when you change your Omarchy theme.

https://github.com/user-attachments/assets/bc552ab4-6733-4687-90d2-2127504fbae1

- It uses a new LazyVim starter repo that includes the required config for this solution to work. 

[Omarchy lazyvim starter repo](https://github.com/tahayvr/omarchy-lazyvim-starter)
(feel free to fork this repo into your own github and switch the git clone command in install/development/nvim.sh)

-  I also needed to add a specific flavour for the regular Catppuccin theme (mocha) for this to work as intended.
-  Finally, I added a line in the omarchy-theme-set that triggers the load of the new theme in NeoVim.

## How does the automatic reload feature work?:

In order to enable the automatic reload feature, each theme in Omarchy must have a neovim.lua file that defines the themes like this:

```lua
-- .config/omarchy/themes/gruvbox/neovim.lua
return {
    { "morhetz/gruvbox", lazy = false },
    {
        "LazyVim/LazyVim",
        opts = {
            colorscheme = "gruvbox",
        },
    },
}
```

Additionally, we must make sure that LazyVim is configured to load every theme on startup. This is done by setting the `spec` and `install` options in the LazyVim configuration.

```lua
-- .config/nvim/lua/config/lazy.lua
spec = {
        -- add LazyVim and import its plugins
        { "LazyVim/LazyVim",           import = "lazyvim.plugins" },
        -- import/override with your plugins
        { import = "plugins" },
        -- import themes on load
        { "ellisonleao/gruvbox.nvim", lazy =false },
        { "catppuccin/nvim",           name = "catppuccin", lazy = false },
        { "folke/tokyonight.nvim", lazy = false },
        { "neanias/everforest-nvim", lazy = false },
        { "rebelot/kanagawa.nvim", lazy = false },
        { "tahayvr/matteblack.nvim", lazy = false },
        { "EdenEast/nightfox.nvim", lazy = false },
        { "rose-pine/neovim",          name = "rose-pine", lazy = false },
    },
install = { colorscheme = { "catppuccin-latte", "catppuccin-mocha", "tokyonight", "habamax", "gruvbox", "everforest", "kanagawa", "nordfox", "matteblack", "rose-pine-dawn" },
```

Then, we need a function to reload the theme based on Omarchy's current theme symlink target. This function will check if the theme has changed and apply the new theme without restarting Neovim.

```lua
[...]

PR #1759, opened 2025-09-18
New neovim plugin solution for auto-loading themes

tahayvr:
As we discussed in #346 I moved all the logic for Neovim theme switching to a plugin called "Chameleon".

Now all we need is:

1. Adding the plugin to our nvim plugins folder

```lua
-- ~/.config/nvim/lua/plugins/chameleon.nvim
return {
  {
    "tahayvr/chameleon.nvim",
    lazy = false,
    priority = 1000,
    config = function()
      require("chameleon").setup({
        system = {
          symlink_path = vim.fn.expand("~/.config/omarchy/current/theme"),
          debounce_ms = 100,
          notify = false,
        },
      })
    end,
  },
}
```

2. theme `neovim.lua` files now only need to assign colors to a set of variables. Plugin looks at these variables in the `current/theme` symlink, and colors the theme based on them.

```lua
-- ~/.config/omarchy/themes/matte-black/neovim.lua

-- for reference check https://github.com/tahayvr/chameleon.nvim
return {
  -- BACKGROUND LAYERS (dark → lighter).
  bg1 = "#121212",
  bg2 = "#1b1b1b",
  bg3 = "#242424",
  -- FOREGROUND LAYERS (bright → muted).
  fg1 = "#bebebe",
  fg2 = "#a3a3a3",
  fg3 = "#8a8a8d",
  -- SELECTION
  selbg = "#333333",
  selfg = "#bebebe",
  -- COMMENTS
  comment = "#8a8a8d",
  -- ACCENT COLORS (syntax + UI)
  color1 = "#D35F5F",
  color2 = "#FFC107",
  color3 = "#b91c1c",
  color4 = "#e68e0d",
  color5 = "#D35F5F",
  color6 = "#bebebe",
  -- UI / DECORATIVE
  uic1 = "#333333",
}
```

3. Add SIGUSR command to the `omarchy-theme-set` script

```bash
pkill -SIGUSR1 nvim
```
4. Finally, we can remove the old symlink method that added a theme.lua symlink in the `nvim/lua/plugins` folder.

It works great. Let me know how it goes. 

more info about the config and how the plugin works in the repo: [chameleon.nvim](https://github.com/tahayvr/chameleon.nvim)

### DIFF
--- a/bin/omarchy-theme-set
+++ b/bin/omarchy-theme-set
@@ -28,6 +28,7 @@ omarchy-restart-waybar
 omarchy-restart-swayosd
 hyprctl reload
 pkill -SIGUSR2 btop
+pkill -SIGUSR1 nvim
 makoctl reload
 
 # Change gnome, browser, vscode themes
--- a/config/nvim/lua/plugins/chameleon.nvim
+++ b/config/nvim/lua/plugins/chameleon.nvim
@@ -1,0 +1,16 @@
+return {
+  {
+    "tahayvr/chameleon.nvim",
+    lazy = false,
+    priority = 1000,
+    config = function()
+      require("chameleon").setup({
+        system = {
+          symlink_path = vim.fn.expand("~/.config/omarchy/current/theme"),
+          debounce_ms = 100,
+          notify = false,
+        },
+      })
+    end,
+  },
+}
--- a/config/nvim/lua/plugins/theme.lua
+++ b/config/nvim/lua/plugins/theme.lua
@@ -1,8 +1,0 @@
-return {
-	{
-		"LazyVim/LazyVim",
-		opts = {
-			colorscheme = "tokyonight",
-		},
-	},
-}
--- a/install/config/theme.sh
+++ b/install/config/theme.sh
@@ -11,9 +11,6 @@ mkdir -p ~/.config/omarchy/current
 ln -snf ~/.config/omarchy/themes/tokyo-night ~/.config/omarchy/current/theme
 ln -snf ~/.config/omarchy/current/theme/backgrounds/1-scenery-pink-lakeside-sunset-lake-landscape-scenic-panorama-7680x3215-144.png ~/.config/omarchy/current/background
 
-# Set specific app links for current theme
-ln -snf ~/.config/omarchy/current/theme/neovim.lua ~/.config/nvim/lua/plugins/theme.lua
-
 mkdir -p ~/.config/btop/themes
 ln -snf ~/.config/omarchy/current/theme/btop.theme ~/.config/btop/themes/current.theme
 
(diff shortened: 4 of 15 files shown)


ryanrhughes:
@tahayvr I'm seeing some similar things, albeit a little better due to the ability to more clearly define hex colors, as encountered in the Pixel test branch. The colors are close in the themes but not quite right. These are most evident in ones like Tokyo Night, which is really our flagship and has to be pretty well perfect I think.

It may be worth continuing to tweak on this but I also fear that at some point with this sort of approach, you wind up with a situation where you're eventually playing whack a mole and there's now way to solve all. 

This could be an alternative approach / fallback option for themes that don't have actual in-spec Neovim themes though.

image

screenshot-2025-09-20_00-00-48

I also am seeing some weirdness when it comes to applying the transparency and highlights. As you cycle through, those aren't quite right.
https://github.com/user-attachments/assets/69ef400c-354e-4d8d-8c97-f2b333d1b2d6

tahayvr:
@ryanrhughes There's no way to replicate all themes completely but I don't think we have to. As long as we use the right colors, we should be good. I will tinker with the color variables in the `neovim.lua` files for the system themes to make them look better (Tokyo Night was ported incorrectly, will fix).

As for the highlights, I am still making updates to the Chameleon plugin to account for specific languages. Update your plugins and try again. see if it's better now.

For transparency, the updated plugin doesn't set a background color, so what is shown now is terminal bg (thanks for pointing it out).

In general, I think theming is much easier if users (and theme makers) can just set color variables and get a new look.

ryanrhughes:
@tahayvr do you think it's possible to merge Chameleon into this stack?

https://github.com/basecamp/omarchy/pull/1898

What I'm thinking here is that we have a bit of order of operations where if a colorscheme is defined by a theme, we'll use that. This is what I have in https://github.com/basecamp/omarchy/pull/1712. 

If the theme designer has elected to define specific colors in the Chameleon format, it uses that. 

If neither of those are true, it falls back to Pixel and utilizes standard ANSI colors.

This effectively gives us 3 layers for theme designers to work with ranging from no work to a lot of work depending on their specific wants, needs, and desires.

tahayvr:
@ryanrhughes It probably is possible. 

However, I'm holding off on spending more time on this till we get some input from @dhh. My understanding from his comments on #346 is that he wants something super simple. Chameleon was my solution based on the comments on that PR. 

Personally, I think any approach that works is fine.

dhh:
I do want something simple. I think we can get the best of all worlds with: Built-in themes get to specify a "real nvim theme plugin", extra themes uses this approach if they want to theme neovim, and if there's nothing, it'll use the pixel approach.

tahayvr:
@ryanrhughes I'm going to close this PR since there are a lot of file changes that are unnecessary here now.

PR #1760, opened 2025-09-18
Expanded mime types for better out-of-the-box experience

koenvaneijk:
Hey Omarchy team! 👋 As another small contribution to make Omarchy feel even more "ready to rock" right after install, I've added to the MIME type defaults in `install/config/mimetypes.sh`. This builds on the existing setup (like images with imv and videos with mpv) by adding sensible handlers for a few more common file types. 

Nothing groundbreaking, just some quality-of-life tweaks to save users a bit of manual config. 

#### What's New?
- **Audio Files**: Now open with mpv (e.g., MP3, WAV, OGG, FLAC)
- **Basic Text Files**: Default to nvim for quick editing (text/plain).
- **Subtitles**: Also nvim (e.g., SRT, MicroDVD).
- **Markdown & Code Files**: nvim again (Markdown, Python, JS, JSON, CSV)
- **Archives**: Handled by pcmanfm (ZIP, TAR, GZIP, 7Z)

#### Files Changed
- `install/config/mimetypes.sh` (added some lines of `xdg-mime` commands under new comment blocks)

#### Testing Notes
- Run the installer and check defaults: e.g., `xdg-mime query default text/markdown` should return `nvim.desktop`.
- No conflicts with other scripts; refreshes via `omarchy-refresh-applications` as before.

Thanks for checking it out and I'm looking forward to making more contributions.

### DIFF
--- a/install/config/mimetypes.sh
+++ b/install/config/mimetypes.sh
@@ -33,3 +33,15 @@ xdg-mime default mpv.desktop video/x-ms-asf
 xdg-mime default mpv.desktop video/x-ogm+ogg
 xdg-mime default mpv.desktop video/x-theora+ogg
 xdg-mime default mpv.desktop application/ogg
+
+# Open audio files with mpv
+xdg-mime default mpv.desktop audio/mpeg
+xdg-mime default mpv.desktop audio/x-wav
+xdg-mime default mpv.desktop audio/ogg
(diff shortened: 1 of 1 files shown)


dhh:
mpv already picks up the audio files, since it's the only one. I think the others are actually also fine as-is.

PR #1763, opened 2025-09-18
Add script to generate torrent from iso

patcito:
Downloading the iso was a bit slow for me so I created a torrent like in the good old days. 

It is using cloudflare as webseed so it should never be slower than cloudflare in worst

Second commit adds scripts to automatically generate and store torrents
for Omarchy ISO releases in the repository.

Changes:
  - Add omarchy-generate-torrent: Core script for creating torrents from ISOs
  - Add omarchy-build-iso: Complete ISO build script with integrated torrent generation
  - Add omarchy-build-iso-with-torrent: Wrapper to add torrent generation to existing builds
  - Add torrents/.gitattributes to mark torrent files as binary
  - Remove previous torrent (will be regenerated with proper workflow)

Workflow:
  - Torrents are generated into torrents/ directory for commit to repository
  - Each release will have its torrent stored permanently in git
  - Provides version history and easy access without external infrastructure
  - Small file size (~72KB) makes repository storage practical

The torrents are configured with:
  - Web seed: [https://iso.omarchy.org/<iso-name](https://iso.omarchy.org/%3Ciso-name)>
  - Trackers: OpenBittorrent and OpenTracker


JaxonWright:
What would be the best way to seed new versions as a seeder automatically? I'd love to dedicate some of my bandwidth to it!

patcito:
it seeds automatically from the webseed iso.omarchy.org upon creation of the torrent file, if you want to seed yourself just use your local torrent client and stay connected I guess.

kevcube:
Thanks, this is great, gives me a much faster download!

dhh:
We have an existing setup for building the ISO over here: https://github.com/omacom-io/omarchy-iso. Have a look at adding the torrent stuff over there.

PR #1773, opened 2025-09-18
add password manager options

robertobarreda:
This update introduces seamless support for selecting a password manager between 
1Password (the current default) and the newly integrated [KeepassXC](https://keepassxc.org/) and [Bitwarden](https://bitwarden.com/) options.

### DIFF
--- a/bin/omarchy-install-password-manager
+++ b/bin/omarchy-install-password-manager
@@ -7,14 +7,14 @@ fi
 
 package="$1"
 
-case $packages in
+case $package in
 "bitwarden")
   omarchy-pkg-add bitwarden-desktop bitwarden-cli
   # Bitwarden package installs the executable with a different name
   ln -s /usr/bin/bitwarden-desktop /usr/bin/bitwarden
   ;;
 "keepassxc")
-  omarchy-pkg-add keepassxc
+  omarchy-pkg-add $package
   if [[ ! -f ~/.config/keepassxc/keepassxc.ini ]]; then
     mkdir -p ~/.config/keepassxc/
     cp -Rpf $OMARCHY_PATH/config/keepassxc/keepassxc.ini ~/.config/keepassxc/keepassxc.ini


muzzwood:
Would love to see a Bitwarden option if possible!

robertobarreda:
> Would love to see a Bitwarden option if possible!

Do you mean [bitwarden](https://archlinux.org/packages/extra/x86_64/bitwarden/) (desktop) or [bitwarden-cli](https://archlinux.org/packages/extra/x86_64/bitwarden-cli/)

bin/omarchy-lock-screen
```diff
@@ -3,9 +3,13 @@
 # Lock the screen
 pidof hyprlock || hyprlock &
 
-# Ensure 1password is locked
-if pgrep -x "1password" >/dev/null; then
-  1password --lock &
+# Ensure password manager is locked
+if pgrep -x "$PASSWORD_MANAGER" >/dev/null; then
+  $PASSWORD_MANAGER --lock &
```
andresatierf: is this needed twice?

bin/omarchy-install-password-manager
```diff
@@ -0,0 +1,19 @@
...
+
+if (($# == 0)); then
+  echo "Usage: omarchy-install-password-manager [1password|bitwarden|keepassxc]"
+  exit 1
+fi
+
+package="$1"
+
+# Install package
+omarchy-pkg-add $package
```
andresatierf: you could install both bitwarden and bitwarden-cli as it was previously, if the user chooses bitwarden

bin/omarchy-lock-screen
robertobarreda: oops, no sorry. fixed

bin/omarchy-install-password-manager
robertobarreda: thx for the suggestion. I've added it.

bin/omarchy-lock-screen
```diff
@@ -3,9 +3,12 @@
...
 pidof hyprlock || hyprlock &
 
-# Ensure 1password is locked
-if pgrep -x "1password" >/dev/null; then
-  1password --lock &
+# Ensure password manager is locked
+if pgrep -x "$PASSWORD_MANAGER" >/dev/null; then
+  case "$PASSWORD_MANAGER" in
+  "1password" | "keepassxc") $PASSWORD_MANAGER --lock & ;;
+  *) ;;
```
andresatierf:
```suggestion
  "bitwarden") $PASSWORD_MANAGER lock & ;;
  *) ;;
```

andresatierf: actually, here you want to lock the cli, not sure if it will lock the gui

robertobarreda: No, it won't. I tried myself to find a way to lock the ui with a command and I couldn't find it.

Anyway, I'll add the `bw lock` to at least lock the cli version.

migrations/1758207683.sh
```diff
@@ -0,0 +1,6 @@
+echo "Copy configs for keepassxc as alternative password manager options"
+
+if [[ ! -f ~/.config/keepassxc/keepassxc.ini ]]; then
```
andresatierf: Couldn't we just create this if $TERMINAL is keepass? We could then copy it if the user changes to keepass?

edit: I don't know how it's handled by other applications. do we always copy the configuration, even if the user is not using it, or do we copy it only if the user starts using it?

bin/omarchy-install-password-manager
```diff
@@ -0,0 +1,25 @@
...
+  exit 1
+fi
+
+package="$1"
+package_extras="$2"
+
+# Install package
+omarchy-pkg-add $package $package_extras
+
+if [ $package = "bitwarden" ]; then
```
andresatierf: since we already need to check things for individual managers, I think the extra_packages param could be removed and we just check in this file if it's bitwarden and add the cli to the install command. that will cleanup the usage in the menu

this is only my opinion, let me know what you think

migrations/1758207683.sh
robertobarreda: It was my initial idea. After all the feedback I realized is much simple to handle it during the install process.
Let me know your thoughts on the latest changes.

bin/omarchy-install-password-manager
robertobarreda: Indeed. Initially I wanted to keep the list of required packages in one place (on the menu as they where initially).
But since other settings are required I definitely moved everything to the password-manager installer.

muzzwood:
> > Would love to see a Bitwarden option if possible!
> 
> Do you mean [bitwarden](https://archlinux.org/packages/extra/x86_64/bitwarden/) (desktop) or [bitwarden-cli](https://archlinux.org/packages/extra/x86_64/bitwarden-cli/)

Sorry for the delay getting back to you! I was meaning the desktop app, but if you've gone with the cli version no probs. 🙏🏻

robertobarreda:
> > > Would love to see a Bitwarden option if possible!
> > 
> > 
> > Do you mean [bitwarden](https://archlinux.org/packages/extra/x86_64/bitwarden/) (desktop) or [bitwarden-cli](https://archlinux.org/packages/extra/x86_64/bitwarden-cli/)
> 
> Sorry for the delay getting back to you! I was meaning the desktop app, but if you've gone with the cli version no probs. 🙏🏻

I went with both actually.

dcunited001:
IIRC, omarchy has some special handling set up for `gnome-keyring` to launch. so idk what the official way to replace it would be. i assumed that would be custom for awhile. I do plan on doing that myself.

If the intent is to have `keepassxc` register itself as `org.freedesktop.secrets` then you could use the `Secret.Service.Lock` method via dbus to lock the password manager.

Here's one way to discover that, via Bustle and QtDbus, though some of these applications pull in dependencies on Arch, so i'd check the package lists before you install. 

2025-10-17T11:03:41,988685023-04:00

Usually for this desktop-integration stuff, D-Bus is the way to go. It can be complex to support across a lot of different installations, though. IMO, I would prefer some other means of determining what the `PASSWORD_MANAGER` is other than via a UWSM or hyprland environment variable. (idk though, i just found this thread)

Where does `1password` fit into Omarchy? Is it just set up for browser integration?  Or does `omarchy-menu` help access it in various ways?

bin/omarchy-lock-screen
```diff
@@ -3,10 +3,12 @@
...
 pidof hyprlock || hyprlock &
 
-# Ensure 1password is locked
-if pgrep -x "1password" >/dev/null; then
-  1password --lock &
-fi
+# Ensure password manager is locked
+case "$PASSWORD_MANAGER" in
+"1password" | "keepassxc") pgrep -x "$PASSWORD_MANAGER" >/dev/null && $PASSWORD_MANAGER --lock & ;;
+"bitwarden") bw lock & ;;
```
sgruendel, line 9: I don't think this will work as `bw` sessions use an env variable to store session id, and you won't be able to access it here

PR #1778, opened 2025-09-18
Fix terminal classes to be compatible with ghostty

andresatierf:
This PR updates the class names of several bash terminal commands to
comply with the GTK `Application.id_is_valid` specification, as
described here: https://docs.gtk.org/gio/type_func.Application.id_is_valid.html

These changes ensure that the class names are compatible with ghostty, as this requires class to follow the above requirements, as described here: https://ghostty.org/docs/config/reference#class.

Some of the applications were also moved to `omarchy-launch-*` to make sure they are launched with the correct classes from anywhere.

### DIFF
--- a/bin/omarchy-launch-about
+++ b/bin/omarchy-launch-about
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-exec setsid uwsm app -- alacritty --class=Omarchy -o font.size=9 -e bash -c 'fastfetch; read -n 1 -s'
+exec setsid uwsm app -- "$TERMINAL" --class=omarchy.About -o font.size=9 -e bash -c 'fastfetch; read -n 1 -s'
--- a/bin/omarchy-launch-audio
+++ b/bin/omarchy-launch-audio
@@ -1,0 +1,3 @@
+#!/bin/bash
+
+exec setsid uwsm app -- "$TERMINAL" --class=omarchy.Wiremix -e wiremix "$@"
--- a/bin/omarchy-launch-bluetooth
+++ b/bin/omarchy-launch-bluetooth
@@ -1,0 +1,3 @@
+#!/bin/bash
+
+exec setsid uwsm app -- blueberry "$@"
--- a/bin/omarchy-launch-floating-terminal-with-presentation
+++ b/bin/omarchy-launch-floating-terminal-with-presentation
@@ -1,4 +1,4 @@
 #!/bin/bash
 
 cmd="$*"
-exec setsid uwsm app -- alacritty --class=Omarchy --title=Omarchy -e bash -c "omarchy-show-logo; $cmd; omarchy-show-done"
+exec setsid uwsm app -- "$TERMINAL" --class=omarchy.Omarchy --title=Omarchy -e bash -c "omarchy-show-logo; $cmd; omarchy-show-done"
--- a/bin/omarchy-launch-wifi
+++ b/bin/omarchy-launch-wifi
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-exec setsid uwsm app -- "$TERMINAL" --class=Impala -e impala "$@"
+exec setsid uwsm app -- "$TERMINAL" --class=omarchy.Impala -e impala "$@"
--- a/bin/omarchy-menu
+++ b/bin/omarchy-menu
@@ -37,11 +37,11 @@ menu() {
 }
 
 terminal() {
-  alacritty --class=Omarchy -e "$@"
+  $TERMINAL --class=omarchy.Omarchy -e "$@"
 }
 
 present_terminal() {
-  omarchy-launch-floating-terminal-with-presentation $1
+  omarchy-launch-floating-terminal-with-presentation "$1"
 }
 
 open_in_editor() {
@@ -178,14 +178,14 @@ show_setup_menu() {
   options="$options\n  Defaults\n󰱔  DNS\n  Security\n  Config"
 
   case $(menu "Setup" "$options") in
-  *Audio*) $TERMINAL --class=Wiremix -e wiremix ;;
+  *Audio*) omarchy-launch-audio ;;
   *Wifi*)
     rfkill unblock wifi
     omarchy-launch-wifi
     ;;
   *Bluetooth*)
     rfkill unblock bluetooth
-    blueberry
+    omarchy-launch-bluetooth
     ;;
   *Power*) show_setup_power_menu ;;
   *Monitors*) open_in_editor ~/.config/hypr/monitors.conf ;;
@@ -291,7 +291,7 @@ show_install_ai_menu() {
   *OpenAI*) aur_install "OpenAI Codex" "openai-codex-bin" ;;
   *Gemini*) aur_install "Gemini" "gemini-cli" ;;
   *Studio*) install "LM Studio" "lmstudio" ;;
-  *Ollama*) install "Ollama" $ollama_pkg ;;
+  *Ollama*) install "Ollama" "$ollama_pkg" ;;
   *Crush*) install "Crush" "crush-bin" ;;
   *opencode*) install "opencode" "opencode" ;;
   *) show_install_menu ;;
--- a/config/waybar/config.jsonc
+++ b/config/waybar/config.jsonc
@@ -100,11 +100,11 @@
     "format-disabled": "󰂲",
     "format-connected": "",
     "tooltip-format": "Devices connected: {num_connections}",
-    "on-click": "blueberry"
+    "on-click": "omarchy-launch-bluetooth"
   },
   "pulseaudio": {
     "format": "{icon}",
-    "on-click": "$TERMINAL --class=Wiremix -e wiremix",
+    "on-click": "omarchy-launch-audio",
     "on-click-right": "pamixer -t",
     "tooltip-format": "Playing at {volume}%",
     "scroll-step": 5,
--- a/default/hypr/apps/system.conf
+++ b/default/hypr/apps/system.conf
@@ -3,7 +3,7 @@ windowrule = float, tag:floating-window
 windowrule = center, tag:floating-window
 windowrule = size 800 600, tag:floating-window
 
-windowrule = tag +floating-window, class:(blueberry.py|Impala|Wiremix|org.gnome.NautilusPreviewer|com.gabm.satty|Omarchy|About|TUI.float)
+windowrule = tag +floating-window, class:(blueberry.py|omarchy.Impala|omarchy.Wiremix|org.gnome.NautilusPreviewer|com.gabm.satty|omarchy.Omarchy|omarchy.About|TUI.float)
 windowrule = tag +floating-window, class:(xdg-desktop-portal-gtk|sublime_text|DesktopEditors|org.gnome.Nautilus), title:^(Open.*Files?|Open Folder|Save.*Files?|Save.*As|Save|All Files)
 
 # Fullscreen screensaver


andresatierf:
Alternatively, we can just apply the `TUI.float` class, which follows the GTK specification and will work with ghostty. @dhh what do you think?

abenz1267:
Sorry, but the ghostty docs are simply wrong. The class name does NOT need to follow GTKs application (!!!)-name format. A class != application id.

No application does btw. Check f.e. with `hyprctl` clients:

```
Window 56023641d9b0 -> #omarchy | Omacom - Discord:
	class: discord
	initialClass: discord


Window 560236482840 -> Gio.Application.id_is_valid — Firefox Developer Edition:
	class: firefox-developer-edition
	initialClass: firefox-developer-edition


Window 5602364dd610 -> ~:
	class: footclient
	initialClass: footclient

	```

**Comment by andresatierf** [comment]
@abenz1267 I believe we are not talking about the same thing. This PR is relative to ghostty windows opened like `ghostty --class=XYZ -e some-app`. In these instances, the class is only properly applied if the class name follows the GTK standard. If it does not, ghostty will not apply the class

**Comment by abenz1267** [comment]
We talk about the same thing. The problem is ghostty treating the class incorrectly and not omarchy setting a wrong class.

Again: what ghostty tries to force here is NOT a standard and also NOT required by GTK.

Edit: i'm investigating.. if firefox can do it, why can't ghostty....

**Comment by abenz1267** [comment]
Ok, i've looked into this for a few minutes:

there is several problems here:
1. GTK4 simply sets the class = application id and there's no way to control this (lol)
2. This literally breaks Ghostty's `single instance mode` ! You can't have a single ghostty instance and start another one with a different class... BECAUSE: ghostty doesn't set the actual WM class with `--class`, but rather the whole application id, WHICH BREAKS DBUS COMMUNICATION.

I feel like this is something Ghostty has to look into and not omarchy.

**Comment by andresatierf** [comment]
I don't know that much about it to do such an in depth investigation. This PR was only to make the ghostty windows float, as after the latest update, if you change the $TERMINAL to ghostty, they no longer float.

**Comment by abenz1267** [comment]
Yes, the problem simply is that this "fix" breaks it for everyone using ghostty's single instance mode, which is a ghostty feature, see [here](https://ghostty.org/docs/help/gtk-single-instance)

**Comment by abenz1267** [comment]
Ok, this is not the whole story:

apparently ghostty has a new service you can enable with `systemctl enable --user app-com.mitchellh.ghostty.service`.

Prior to that you needed to run ghostty with `+new-window` to open a new window (NOT INSTANCE).

Apparently ghostty now does not need `+new-window` anymore (if the service is running). Running `ghostty --class` will open a new instance though and not just a new window. So this is slower than opening a new ghostty window. Not sure.. personally i don't think it's a biiiig biggy for opening non-frequent apps like f.e. wiremix or whatever.

**Comment by abenz1267** [comment]
This gets even weirder:

IF you run the ghostty service, you also need `quit-after-last-window-closed = false`, otherwise the service is pointless.

The catch: now you need end up with dangling ghostty instances every time you launch it with `--class`. So you need to additional run `--quit-after-last-window-closed=true`.

**Comment by naxels** [comment]
I run into the same issue with Ghostty as my default Terminal.

and as mentioned, applying class=TUI.float directly works.

Doesn't work:
`uwsm app -- ghostty --class=Wiremix -e wiremix`

Works with floating:
`uwsm app -- ghostty --class=TUI.float -e wiremix`

Same is true for Impala and others.

Is this issue open with Ghostty or are they expecting Omarchy to change?
Looking at the docs mentioned above, https://ghostty.org/docs/config/reference#class, it seems Omarchy has to adapt..

**Comment by abenz1267** [comment]
From what i gathered, this is nothing Ghostty can fix, since it seems to be a GTK limitation. So omarchy has to adapt i guess. And when running ghostty with single instance mode (which one should do): remember to call `--quit-after-last-window-closed=true` every time you call `--class` so you don't end up with dangling Ghostty instances that eat 300mb memory.

This also means that omarchy scripts need to identify when $TERMINAL is ghostty and otherwise NOT use that flag. Meh.

**Comment by naxels** [comment]
What would it take to make it simple?

maybe we can discuss on Ghostty an auto close after last window closed or some other way like a config to make it less complicated?

I'm not that familiar with the internals of terminals/GTK myself unfortunately

**Comment by abenz1267** [comment]
The problem is:

- ghostty has a service you can enable so it starts faster
- this service is a bit pointless, if you DON'T have `quit-after-last-window-closed=false` in your config, because it SHOULD stay open
- but IF you have that config set => you run `ghostty --class=my.cass` => you close that window again => ... the instance is STILL running, AS EXPECTED because of the flag. "expected".
- so you NEED to do `ghostty --class=my.cass --quit-after-last-window-closed=true`
- this flag is ONLY a ghostty thing => if you have a different $TERMINAL => breaks the keybind

I'm afraid there isn't actually something one can do about this, other than scripting terminal flags based on the terminal.

Foot f.e. also has the ability to run a server/client mode. Without these issues. None of them. :shrug:

**Comment by Woosah** [comment]
Please excuse me for not finding the PR earlier and creating a duplicate (#2109)... But to put in my 2 cents, I think that several issues seem to be mixed up here.

A quick look at [GitHub](https://github.com/transmission/transmission/issues/5067) and [beyond](https://docs.gtk.org/gtk4/index.html) gives me the impression that ghostty may be very strict in this regard, but not entirely wrong, especially if they refer exclusively to GTK 4 in their definition. However, in doing so, they violate common expectations. In GTK 4, there is apparently no longer a formal WM_CLASS, as this actually refers to X11 and not to Wayland. This means that, from the perspective of GTK 4, there is an application ID, but no longer a WM_CLASS. Looking at it from the other perspective, i.e., from the window manager's point of view, the class is practical and should be customizable if necessary to enable any window identification. 

To summarize, ghostty asks quite rightly in their GTK ***4*** implementation, what should WM_CLASS be? And Omarchy, as a “user,” asks quite rightly, why shouldn't I decide what that should be, it makes my life easier! The closest concept that I could find in GTK 4 is the the one of an accessible_role, see [here](https://docs.gtk.org/gtk4/class_method.Widget.set_accessible_role.html) and [here](https://docs.gtk.org/gtk4/class_method.Widget.get_accessible_role.html). But I am no expert in this, especially when it comes to questions about using "--quit-after-last-window-closed=true" or not, or the consequences of this systemd-service and so on. This is over my head. 

Therefore, for me the "correct" implementation (on the ghostty side) is a purely philosophical question that I personally cannot solve. But it is really annoying that I would like to use ghostty as my daily driver because it feels good, and suddenly I cannot get a floating Impala window anymore. Therefore, if there can be a way to get this going again and this is not too much a burden on the Omarchy side, like a convention to add "xxx.float" to all windows that should be floating (maybe not always "TUI.float" to be able to be more precise in the applied windowrules), that, from my personal perspective, would be pracmatic and great! ;-)

**Comment by abenz1267** [comment]
There's nothing Ghostty can do about this really, as it's bound to do what GTK4 wants.

The problem is that Omarchy doesn't know HOW people use Ghostty. Service enabled? No? 

And depending on that AND the `quit-after-last-window-closed` setting set by users, it is mandatory to add `--quit-after-last-window-closed=true` to the commands calling ghostty, because as i already said: otherwise you have dangling background ghostty processes.

Not only that, but ultimately, you'd always need to call ghostty with `+new-window` with the regular "open terminal" keybind. Otherwise the service won't be triggered!

The point is: if people use Ghostty as a terminal, Ghostty is very much special needs and would require special treatment in a few places.

**Comment by Woosah** [comment]
I investigated this a little bit further and wanted to share this, but, as always, I can only speak definitely for my own installation, which is tuned to my needs, and about my personal experience with ghostty.

I installed ghostty through pacman, did nothing out of the ordinary, only adjusted the config in regard of ghostty's look. That I did from day one on omarchy, which was 2 weeks after its first relaease, I think, i.e., on version 1. I always used (and use) zsh as my shell, but I do not think that this makes a difference. This setup was my daily driver even before on KDE since ghostty was released. I am running this setup on my 7 year old desktop at home (i7-7700K and GTX 1060), a 5 year old AMD-based Laptop without a dedicated GPU) and my work desktop with an i7 12th Gen and RTX 3070 Ti. I never had any problems regarding CPU, RAM, performance or anything, neither on KDE nor on Omarchy. It was (and is) a completely snappy experience. 
The only adjustement I did since my first Omarchy install is to delete my own theming and to use Omarchy's "Install Terminal" function to get a really cool and way more far-reaching integration of ghostty into Omarchy's theming than before. Works like a charm! 

When I activate ghostty on purpose with many instances, this happens:

ghostty_instances
ghostty_processes

All deliberately called ghostty instances are showing up as independent processes with no gtk-single-instance mode activated. And on window closing, the processes are closed respectively, too. The same happens, when I start ghostty once, and then start "ghostty -e wiremix" from alacritty. All come up as separate processes.

ghostty_instances_2
ghostty_processes_2

And when I look at `ghostty +show-config`, I have:

ghostty_config

So, `gtk-single-iunstance=false` in the config. After that, I looked into my available services and installed unit files as well as any ghostty process at idle, I find this:

ghostty_service
ghostty_process_on_idle

So, there is the dbus acitvated service available, and at idle we have the process running with --gtk-single-instance=true and --initial-window=false. I really do not know what triggered the service to run while doing this testing, because I did not manually start it and the user systemd unit is listed as disabled. But although now I have this single instance process running in the current session (and I expect it to be gone again after a relaunch), it does not interfere with any other ghostty instances opened from now on. 

To summarize: I hope that gives a little impression about how a user (i.e., me) is using ghostty as a terminal, and that my feeling is, these possible ghostty-GTK4-pitfalls are actually not that important while using Omarchy. 

That leaves the question, how do I get back the floating TUIs? We can certainly agree about the class naming convention being strange, but how to proceed from here? How can I help to implement a solution?

Best regards, and sorry for the long long long comment! :-)

**Comment by abenz1267** [comment]
As I already mentioned multiple times: Ghostty handling is not as simple as "just change the class". There's a few things to consider.

**Comment by Woosah** [comment]
Maybe the dangers or important consequences of this are still not clear to me, but if there are so many things that can go wrong depending on different user setups and depending on the use or not-use of the dbus service, then the most pragmatic course of action after "just change the class" might be to keep alacritty in the line? 

That is, builtin features where Omarchy needs a predictable way of doing things, especially when they touch on the look and feel of the graphical environment, should maybe still be done via alacritty? I mean, ghostty is great, but if this issue causes such a headache, then it might be not worth it to use ghostty in these cases... 

At the end of the day, Omarchy should provide a great look, cause no issues for the average user and should be predictable for the maintainers. If this can be achieved by NOT using $TERMINAL in all and every case, so be it. Because the main benefit of having ghostty as the $TERMINAL is to have it following the Omarchy theme. I am living half of my time in Omarchy in ghostty, and that theming feature really makes my day, while at the same time, I did not recognize any downsides. Besides the floating issue. And I will not loose any sleep if these windows are called using alacritty like before.

At least that would be my proposal.

**Comment by Naelpuissant** [comment]
> There's nothing Ghostty can do about this really, as it's bound to do what GTK4 wants.
> 
> The problem is that Omarchy doesn't know HOW people use Ghostty. Service enabled? No?
> 
> And depending on that AND the `quit-after-last-window-closed` setting set by users, it is mandatory to add `--quit-after-last-window-closed=true` to the commands calling ghostty, because as i already said: otherwise you have dangling background ghostty processes.
> 
> Not only that, but ultimately, you'd always need to call ghostty with `+new-window` with the regular "open terminal" keybind. Otherwise the service won't be triggered!
> 
> The point is: if people use Ghostty as a terminal, Ghostty is very much special needs and would require special treatment in a few places.

Hello ! 
First thanks for the gtk-single-instance, didn't knew about this option lol !
I might misunderstand something but I tried to monitor ghostty activity using btop and this is what I saw : 

## gtk-single-instance not set
- I open ghostty w btop (1 service running)
- I open ghostty (2 service)
- I open ghostty by setting a class com.omarchy.test (3 services)

## gtk-single-instance=true
- I open ghostty w btop (1 service running)
- I open ghostty (1 service)
- I open ghostty by setting a class com.omarchy.test (2 services)
- I kill ghostty that have the class (1 service)

## TL;DR
ghostty should be fine even with gtk-single-instance=true.

**Comment by abenz1267** [comment]
you need this config:

```
quit-after-last-window-closed = false
gtk-single-instance = true
```

otherwise ghostty will close and `gtk-single-instance` is pointless. Once you have this, you'll see the issues.

**Comment by Naelpuissant** [comment]
I tried with
```
quit-after-last-window-closed = false
gtk-single-instance = true
```
And got the same result that with just gtk-single-instace=true except that after I closed every windows, pop a new ghostty and it has a coherent elapsed time (I can see it stayed opened). 

Tell me if I still missed something, but if we good here too I guess we could fix the initial issue.

Ghostty version : Ghostty 1.2.2-arch1

**Comment by abenz1267** [comment]
> I tried with
> 
> ```
> quit-after-last-window-closed = false
> gtk-single-instance = true
> ```
> 
> And got the same result that with just gtk-single-instace=true except that after I closed every windows, pop a new ghostty and it has a coherent elapsed time (I can see it stayed opened).
> 
> Tell me if I still missed something, but if we good here too I guess we could fix the initial issue.
> 
> Ghostty version : Ghostty 1.2.2-arch1

No, you did not get the same results. If you use `quit-after-last-window-closed = false` and run `ghostty --class some.class` and close it again, it'll literally be still running. That's what that option does. And that's a problem.

**Comment by Naelpuissant** [comment]
Looks like the option doesn't quit after the last ghostty window was closed when set to false, which is what we want. I can't reproduce the bug/behavior you are talking about. Can you give more details on how you are testing this ?

**Comment by abenz1267** [comment]
> Looks like the option doesn't quit after the last ghostty window was closed when set to false, which is what we want. I can't reproduce the bug/behavior you are talking about. Can you give more details on how you are testing this ?

What do you mean "this is what we want?"????? YOU WANT dangling ghostty process? XD

**Comment by abenz1267** [comment]
I'll explain this AGAIN:

1. you have ghostty as a service to workaround the bad launch times
2. you have `quit-after-last-window-closed = false`, because otherwise the service is useless

Now you have:

ONE ghostty process already running.

Now you do `ghostty --class=some.class`:

this will launch a NEW ghostty process, because of the different app id

You "close" this ghostty window.

Now you have 2 ghostty processes STILL.

Thats what you want? :joy:

**Comment by Naelpuissant** [comment]
Maybe I’m not clear but this is not what I have, the 2nd process is cleared, no dangling process, just one « global » process as expected

**Comment by abenz1267** [comment]
Impossible. That's literally what `quit-after-last-window-closed = false` does.

**Comment by Woosah** [comment]
Maybe there is a misunderstanding regarding ghostty's behavior:

You have a single process with the normal terminal that stays open after the last terminal closes, when you use single instance mode with quit after last one set to false. The process stays.

But: If you open a second window with the "-e" option, than ghostty uses single instance mode set to false and quit after last as true automatically. See here:

https://ghostty.org/docs/config/reference#initial-command



Am 25. Oktober 2025 09:16:07 MESZ schrieb Andrej Benz ***@***.***>:

**Comment by abenz1267** [comment]
That seems to be the missing piece!

**Comment by Naelpuissant** [comment]
Oh nice catch ! So we should be fine with this fix

**Comment by Woosah** [comment]
Yes, it should be. Just added 
```
gtk-single-instance=true
quit-after-last-window-closed=false
```
to my config. After that, you see only one process for the actual terminal, regardless of how many terminal windows you spawn, and one extra process for any TUI started with the `-e` option, i.e., impala, btop, etc. Works really smooth, no problems. After closing every window inlcuding the last, one main process of ghostty stays to be ready to open a new terminal window immediately. And I can definitely see the difference to before, because the launch of a second "ordinary" terminal after the main process is present is reeeealy snappy. It works. Certainly, the ghostty devs did their job. 

So, what about the original problem? I would recommend changing the default settings of ghostty in omarchy to the ones above, and to adhere to the GTK4 convention of a class, i.e., change class impala to impala.float to let the windowrules do their magic? Any suggestions? What about the font size? Alacritty comes with 9 as a default, and because it's used in many places (about, install, etc.), the hardcoded value for the size of a terminal floating window in these cases might by a problem when a user changes the ghostty terminal font size.

dhh:
This has all been sorted now on dev 👍

PR #1779, opened 2025-09-18
Keyboard backlight keybindings 

ludagoo:
I added keybindings to control keyboard back light brightness. This works on my MacBook Air 2017. I am not sure how this will work with other computers or how to test. Surely this is not generic enough, but I do not know where to go from here, so am submitting PR to get feedback.

### DIFF
--- a/config/hypr/hypridle.conf
+++ b/config/hypr/hypridle.conf
@@ -6,6 +6,12 @@ general {
 }
 
 listener {
+    timeout = 30
+    on-timeout = sh -c 'brightnessctl -d smc::kbd_backlight g > /tmp/kbd_brightness && brightnessctl -d smc::kbd_backlight s 0'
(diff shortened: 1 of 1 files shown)


Xoorgh:
I did something similar for my MacBook Air 2013. Also added a timed shutoff that returns to the previous value on resume, if it's of any help:

```sh
listener {
    timeout = 30
    on-timeout = sh -c 'brightnessctl -d smc::kbd_backlight g > /tmp/kbd_brightness && brightnessctl -d smc::kbd_backlight s 0'
    on-resume = sh -c 'val=$(cat /tmp/kbd_brightness 2>/dev/null || echo 50); brightnessctl -d smc::kbd_backlight s "$val"'
}
```

ludagoo:
@Xoorgh  where would that go?

Xoorgh:
@ludagoo into config/hypr/hypridle.conf

PR #1781, opened 2025-09-18
Update omarchy-launch-about to use the default terminal

LucaNerlich:
Ive installed ghostty via "install -> terminal" - which also correctly set it as the "default" (for super+return e.g), also fast-fetch shows ghostty.

The omarchy launch about / button top left still uses alacritty, which doesnt work (for me), since Ive uninstalled it (after switching to ghostty)

### DIFF
--- a/bin/omarchy-launch-about
+++ b/bin/omarchy-launch-about
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-exec setsid uwsm app -- alacritty --class=Omarchy -o font.size=9 -e bash -c 'fastfetch; read -n 1 -s'
+exec setsid uwsm app -- $TERMINAL --class=Omarchy -o font.size=9 -e bash -c 'fastfetch; read -n 1 -s'


gkurts:
font.size isn't an option for ghostty, it uses font-size (a dash instead of a period). To support both would probably require some way of determining which shell is the default and applying the necessary parameter(s) based on that.

Also, when I applied your change and fixed the font size parameter, it launched, but not in a floating window like Alacritty did, so some work on that would need to happen as well.

gkurts:
image

Quick follow up... I had to change the --class parameter value to `com.Omarchy` to meet the ghostty / gtk [requirements](https://docs.gtk.org/gio/type_func.Application.id_is_valid.html) that class names have at least one dot in them, then I had to modify the system.conf file to change the windowrule for the Omarchy class to .*Omarchy (I suck at regexes so there may be a better way of doing this). 

Once those were done, it worked.

LucaNerlich:
let me check regarding the classes. Good point. In terms of font-size, maybe we can get away with not setting / passing one at all? Ill try.

LucaNerlich:
scratch this, there are way to many occurances of alacritty still in the omarchy setup, to worry about this 'about' feature atm. e.g see here https://github.com/basecamp/omarchy/pull/1795

gkurts:
There are a few other PRs related to the classes and fonts in this scope, too.

Issue #655, opened 2025-08-11
Laptops with dual batteries don't report dual charge level


rigo3t:
I have a Lenovo ThinkPad T480. Previously I was on KDE Plasma and by default I could see the two batteries and their separate charge levels. 

I attempted to play around with the Waybar configuration to add BAT0 and BAT1, and while I was able to display two battery icons, they were simply mirrored. I realize that probably it's necessary to do some changes to upower, or something else I don't know about.

I do not have the technical expertise to expose the second battery to the configuration, but it's not uncommon for ThinkPads to have two batteries and most of us nerds are on these anyway. :)

emourad:
Same with my thinkpad

raybarrera:
> Tested on my surface book 2, would be great if you could check it on your end as well https://github.com/basecamp/omarchy/pull/1783

Just realized I misread the issue--my PR only addresses the low battery notification error that comes from dual batteries, not the other issue. My bad!

raybarrera:
@rigo3t have you tried a config like this: https://github.com/Alexays/Waybar/wiki/Configuration#multiple-instances-of-a-module

landsman:
> [@rigo3t](https://github.com/rigo3t) have you tried a config like this: https://github.com/Alexays/Waybar/wiki/Configuration#multiple-instances-of-a-module

Thanks, this is working for me perfectly!

I put there:
BAT0
BAT1

<img width="471" height="288" alt="Image" src="https://github.com/user-attachments/assets/e47949e5-6bb2-4d2c-95db-4d0e62c567cc" />

PR #1783, opened 2025-09-18
Battery monitor misreporting charge on multiple battery devices

raybarrera:
On devices with multiple batteries the grep would return a string containing the various battery entries from `upower -e`, which would return a null entry when passed to the `upower -i` comand.

This change stores the multiple entries, the loops over them to get their average. It also handles trimming floats that may now occur as part of the average calculations.

Relates to #655

### DIFF
--- a/bin/omarchy-battery-monitor
+++ b/bin/omarchy-battery-monitor
@@ -6,31 +6,46 @@ BATTERY_THRESHOLD=10
 NOTIFICATION_FLAG="/run/user/$UID/omarchy_battery_notified"
 
 get_battery_percentage() {
-  upower -i "$(upower -e | grep 'BAT')" \
-  | awk -F: '/percentage/ {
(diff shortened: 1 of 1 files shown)


raybarrera:
WRT the tab spacing--that's just the default formatting choice with the default lsp in the lazyvim setup, so I opted to leave it iin

phantom-kali:
A couple quick thoughts on the diff:
- `get_battery_state()` still hits the original multi-battery snag (passes all paths to `upower -i`, causing a malformed arg). We could fix that by looping there too, but integrating state checks into the percentage func (like in #1990 ) avoids the duplication and keeps things DRY.
- Averaging across *all* batteries (regardless of state) might mask real risks—e.g., one discharging at 5% + one charging at 100% averages ~52%, skipping the alert. For safety, alerting on the *min* discharging battery feels more conservative. If we want a "combined" metric, maybe average just the discharging ones?

PR #1789, opened 2025-09-19
Fixed ghostty theme name for nord theme (nord must be lowercase.)

mroboff:


ryanrhughes:
This doesn't appear to be correct so I'm going to close this but if you're having troubles with it, ping me and I'll help figure out what's going on.

image

mroboff:
> This doesn't appear to be correct so I'm going to close this but if you're having troubles with it, ping me and I'll help figure out what's going on.
> 
> image

When launching Ghostty with the nord theme set,with the config set to theme = Nord as per default, Ghosty would launch with an error saying the Nord theme could not be found in any of the default paths. Looking at the installed themes, I noticed nord was installed but was titled in lowercase. Switching the theme = Nord to theme = nord fixed the issue. Ghostty was installed via the Omarcy install menu, so I suspect others may have this issue as well.

PR #1795, opened 2025-09-19
Make omarchy-menu support $TERMINAL

nmeylan:
# Changes
Now that `ghostty` and `kitty` can be used as default instead of `alacritty`, the user configured $TERMINAL should be used in omarchy-menu.

**Note**: The 3 "supported" terminals by omarchy, all handles `--class` (https://ghostty.org/docs/config/reference#class, https://sw.kovidgoyal.net/kitty/invocation/#options) , if new terminal would be added to supported terminals the script may be adapted:

**Reason**:
After setting up `ghostty` i have deleted `alacritty` and `omarchy-menu` was not working anymore

### DIFF
--- a/bin/omarchy-menu
+++ b/bin/omarchy-menu
@@ -37,7 +37,7 @@ menu() {
 }
 
 terminal() {
-  alacritty --class=Omarchy -e "$@"
+  $TERMINAL --class=Omarchy -e "$@"
 }
 
 present_terminal() {


abenz1267:
This can't be done as long as Ghostty doesnt handle `--class/--title/--working-directory` correctly. 

Window rules are not working, this is why omarchy still relies on alacritty.

dhh:
What @abenz1267 said. We need to make it clear exactly what's needed to get full support. Then open some issues over on the Ghostty GH. And I can work with @mitchellh to get us all the way to a place where Ghostty can fully replace all elements of Alacritty for a complete swap/default.

nmeylan:
Thanks for your comments, I guess it was too soon to uninstall alacritty

luposlip:
@abenz1267 does `kitty` support window rules?

PR #1796, opened 2025-09-19
Swiss keyboard layout patch

rixlabs:
Patch to support the Swiss keyboard layouts in hyprland config. Unfortunately the format of the layout coming from arch (linux in general) for Swiss keyboard is ugly.

I'm not the best with bash and I am absolutely not an expert with hypr so every suggestion is more than welcome.

### DIFF
--- a/install/config/detect-keyboard-layout.sh
+++ b/install/config/detect-keyboard-layout.sh
@@ -2,6 +2,7 @@
 conf="/etc/vconsole.conf"
 hyprconf="$HOME/.config/hypr/input.conf"
 
+#TODO add check for underscores in the layout from arch and set the hypr configs accordingly
 if grep -q '^XKBLAYOUT=' "$conf"; then
   layout=$(grep '^XKBLAYOUT=' "$conf" | cut -d= -f2 | tr -d '"')
   sed -i "/^[[:space:]]*kb_options *=/i\  kb_layout = $layout" "$hyprconf"
--- a/install/config/detect-keyboard-layout.sh
+++ b/install/config/detect-keyboard-layout.sh
@@ -17,10 +17,6 @@ if grep -q '^KEYMAP=' "$conf"; then
     layout="ch"
     variant="fr"
     ;;
-  it_CH*)
-    layout="ch"
-    variant="it"
-    ;;
   *)
     layout="$keymap"
     ;;
--- a/install/config/detect-keyboard-layout.sh
+++ b/install/config/detect-keyboard-layout.sh
@@ -2,13 +2,35 @@
 conf="/etc/vconsole.conf"
 hyprconf="$HOME/.config/hypr/input.conf"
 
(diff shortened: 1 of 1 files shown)


rixlabs:
This PR is linked to https://github.com/omacom-io/omarchy-iso/pull/40

ryanrhughes:
Resolved in [v3.0.2](https://github.com/basecamp/omarchy/releases/tag/v3.0.2)

dhh:
Thanks for the patch! A few concerns before this lands:

**Regression for users who currently work.** The previous script read `XKBLAYOUT`/`XKBVARIANT`, which are already XKB-formatted (`ch`, `us`, `de`, etc.) — exactly what Hyprland expects. The new script ignores them entirely and reads `KEYMAP=` instead, which is a console keymap, not an XKB layout. Anyone whose `vconsole.conf` already has correct XKB fields (the common archinstall path) loses that signal.

**The default fallback `layout="$keymap"` is wrong for most non-trivial keymaps.** Console keymap names ≠ XKB layout names:
- `de-latin1` → not a valid `kb_layout`; should be `de`
- `dvorak`, `colemak` → these are XKB *variants* of `us`, not layouts
- `uk` console keymap → XKB calls it `gb`
- `ru` console keymap → typically wants XKB `us,ru` with a toggle

**Missing `it_CH`** (Italian Swiss) — if you're translating Swiss keymaps, include the third one.

**Suggested approach:** keep XKB fields as the primary source and only translate `KEYMAP` as a fallback for known special cases:

```bash
if grep -q '^XKBLAYOUT=' "$conf"; then
  layout=$(grep '^XKBLAYOUT=' "$conf" | cut -d= -f2 | tr -d '"')
fi
if grep -q '^XKBVARIANT=' "$conf"; then
  variant=$(grep '^XKBVARIANT=' "$conf" | cut -d= -f2 | tr -d '"')
fi

if [[ -z "$layout" ]] && grep -q '^KEYMAP=' "$conf"; then
  keymap=$(grep '^KEYMAP=' "$conf" | cut -d= -f2 | tr -d '"')
  case "$keymap" in
    de_CH*) layout="ch"; variant="de" ;;
    fr_CH*) layout="ch"; variant="fr" ;;
    it_CH*) layout="ch"; variant="it" ;;
  esac
fi

[[ -n "$layout" ]]  && sed -i "/^[[:space:]]*kb_options *=/i\\  kb_layout = $layout" "$hyprconf"
[[ -n "$variant" ]] && sed -i "/^[[:space:]]*kb_options *=/i\\  kb_variant = $variant" "$hyprconf"
```

That narrowly fixes the Swiss case without breaking anyone else.

Minor: the `# Write to hypr config` comment is noise — drop it.

rixlabs:
Thanks for the feedback. I'll fix it

dhh:

PR #1801, opened 2025-09-19
Fix Matte Black theme ghostty and vscode configs

tahayvr:
I missed some important values in the Matte Black Ghostty theme (BG, FG, Cursor,...)


tahayvr:
I also fixed a vscode typo for the Name value. it was stopping the theme switch.

PR #1807, opened 2025-09-19
Quote Hyprland terminal cwd binding in migration

acrogenesis:
Added a migration that rewrites the Super+Return terminal binding to include the quoted $(omarchy-cmd-terminal-cwd) helper it also handles legacy bindings that lack the helper or omit quotes, while leaving already-migrated configs untouched

Fixes this error when changing to ghostty
screenshot-2025-09-19_09-08-00


dhh:
We moved on to a different setup with xdg-default-terminal.

PR #1812, opened 2025-09-19
Hide hyperland cursor while typing

js:
Set hypr config [`cursor:hide_on_key_press`](https://wiki.hypr.land/Configuring/Variables/#cursor) to true by default, to hide the cursor when you start typing so it is not in the way. I find this a nice quality of life improvement as the cursor never gets in the way when I'm typing. It's also happens to be how macOS behaves.


dhh:
TIL! This is super nice!

PR #1815, opened 2025-09-19
chore: run shellcheck on install/iso.sh

jmelahman:
I was reading the source and noticed these `gum.sh`, `logo.sh`, and `layout.sh` where combined into this `presentation.sh` in https://github.com/basecamp/omarchy/commit/5372393211373c1601ad257bc6f85d5cc237f36c which wasn't reflected in `iso.sh`.

The preflight scripts like `set-size-vars.sh` seemed to have been introduced in https://github.com/basecamp/omarchy/commit/8651f6d8b0f3bd03eba7d556fd07327c69360623 without any corresponding source and I couldn't find any existing/deleted/generated files with these names. Maybe that changed missed some untracked files?

At any rate, this makes shellcheck happy for this file.

Worth noting I didn't test this.

### DIFF
--- a/install/iso.sh
+++ b/install/iso.sh
@@ -1,14 +1,7 @@
 # Called by Omarchy ISO setup before starting configurator and archinstall
+# shellcheck shell=bash
 
-source "$OMARCHY_INSTALL/preflight/set-size-vars.sh"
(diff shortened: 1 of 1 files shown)


dhh:
We no longer use this file.

PR #1822, opened 2025-09-20
Make HEY the default mailto handler

ryanrhughes:


naxels:
Is there anything we can abstract so that people that don't have / use HEY can also have a default mailto other than HEY?

(like for example what we now have with $TERMINAL and $EDITOR)

ryanrhughes:
I think because of how they work, it might be tougher to just make a variable but I'm thinking if people want to add that; it'd be a matter of expanding the options that the email handler looks at.

At that point it could check a system variable for a different option to have been set and take a different route.

Just branch off of this one and PR back into it if you'd like to add one.

PR #1242, opened 2025-08-28
add themes from https://github.com/eza-community/eza-themes

sgruendel:
add theme support for eza, fixes #1135

PR #1827, opened 2025-09-20
Add eza themes for Catppuccin Latte, Matte Black and Nord

sgruendel:


tahayvr:
Why? What's wrong with using the ANSI colors like we are right now?

sgruendel:
> Why? What's wrong with using the ANSI colors like we are right now?

We've already added eza themes from https://github.com/eza-community/eza-themes with #1242 , this PR just adds some more.

dhh:
Hmm, I find these themes a bit much. Here's the new Nord:
image

Here's the old one:

image

And the Matte Black one also doesn't look right. I think maybe we're actually better off just relying on the ANSI colors all-in.

Appreciate your work for this, though!

PR #1830, opened 2025-09-20
Fix screen recording VAAPI fallback for older hardware

jpow18:
# Fix: Add graceful fallback for screen recording when VAAPI hardware acceleration fails

The current script crashes on systems where `wl-screenrec` cannot initialize VAAPI hardware acceleration, preventing screen recording entirely. This affects MacBooks and likely other systems with older Intel graphics.

## Problem

* `wl-screenrec` fails with VAAPI errors and panics, leaving users unable to record
* Script terminates instead of attempting alternative recording methods

## Solution

* Try `wl-screenrec` first (optimal for systems with working hardware acceleration)
* When `wl-screenrec` fails to start, automatically fall back to `wf-recorder` with software encoding
* Maintains existing NVIDIA handling and all original functionality

## Testing

Verified on MacBook Pro with Intel 3rd gen graphics:

* `wl-screenrec` fails with VAAPI connection errors and panics
* Fallback to `wf-recorder` works perfectly
* User experience: seamless recording regardless of hardware acceleration support

## Additional Context

Systems experiencing this issue often don't have complete VAAPI tooling installed:

```bash
$ vainfo
bash: command not found: vainfo
```

### DIFF
--- a/bin/omarchy-cmd-screenrecord
+++ b/bin/omarchy-cmd-screenrecord
@@ -1,5 +1,4 @@
 #!/bin/bash
-
 [[ -f ~/.config/user-dirs.dirs ]] && source ~/.config/user-dirs.dirs
 OUTPUT_DIR="${OMARCHY_SCREENRECORD_DIR:-${XDG_VIDEOS_DIR:-$HOME/Videos}}"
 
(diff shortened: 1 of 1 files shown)


dhh:
Hmm, not loving this where we just try to start something that's then going to blow up. Let's see if we can't find a test ala what we're doing with nvidia to guide the conditional.

pedro757:
I just copied this code as a temporary fix and it works for me in a old computer, I guess old hardware is why this is still an open issue #1858

jpow18:
Thanks for the tip, dhh. I updated the script to check VAAPI availability upfront and use that to determine which recorder to use. This way it doesn't try to start something that will potentially break.

Thanks for checking this out, pedro757. I appreciate you reviewing my work!

dhh:
@ryanrhughes do we still need this?

pedro757:
I updated to v3.1 and screen recording is still not working for older hardware

dec05eba:
Fallback is still needed but should now be done differently. It should be done by only adding `-fallback-cpu-encoding yes` option to the gpu screen recorder command

neheb:
Needs rebasing.

dhh:
We've revamped the screenrecording entirely. Please reopen with a new patch, if anyone wants to fix this up.

PR #1831, opened 2025-09-20
Add import for user walker css config

euandeas:
This will allow people to override the styling in a similar way to the lookandfeel.conf e.g. rounded corners. This will also allow dot files to change the walker styling more easily.


ryanrhughes:
Walker v1 changes this setup. If users want to override, it'd be best for them to just duplicate the theme to their config dir and update as they see fit.

PR #947, opened 2025-08-20
Theme Auto Switching based on Sunrise / Sunset

mazdak:
### 🌅 Theme Auto-Switching Based on Sunrise/Sunset

This PR introduces automatic theme switching functionality that seamlessly transitions
between light and dark themes based on sunrise and sunset times using the darkman
service.

✨ **What does it do?**

• Automatic Theme Switching: Integrates with `darkman` to automatically switch between
light and dark themes based on solar position
• Location-Based Configuration: Uses latitude/longitude coordinates for accurate
sunrise/sunset calculations
• Menu Integration: Added "Theme Auto-Switch" option to the Toggle menu for easy
access
• Configurable Themes: Supports custom light and dark theme selection via
`~/.config/omarchy/theme-auto-switch.conf`
• Optional Nightlight: Can integrate with `hyprsunset` for temperature-based nightlight
effects

screenshot-2025-08-20_18-52-08

### 🔧 Implementation Details

• New Script: `omarchy-toggle-theme-auto-switch` - Enables/disables the auto-switching
service
• Configuration File: `theme-auto-switch.conf` - Stores theme preferences and
location settings
• Migration Script: 1755702456.sh - Installs darkman and sets up necessary hooks
• Hook Scripts: Automatic theme switching via XDG data directories (~/.
local/share/dark-mode.d/ and ~/.local/share/light-mode.d/)

PR #1836, opened 2025-09-20
Add auto theming (day/night) in omarchy including a menu item

oluwatimio:
### Feature Addition

Omarchy is very beautiful, and comes with a lot of beautiful themes out the box some are light and some are dark themes. This PR adds functionality to automatically change themes (day/night) based on the user's config.

https://github.com/user-attachments/assets/5e8e977b-cc15-4abd-8995-4ded7b1fd750

### DIFF
--- a/config/auto-themer/theme-auto.conf
+++ b/config/auto-themer/theme-auto.conf
@@ -1,0 +1,15 @@
+# Auto theme switch configuration (source-of-truth)
+# This file is read directly by: ~/.local/share/omarchy/bin/omarchy-theme-auto
+# Edit this file directly: ~/.local/share/omarchy/config/auto-themer/theme-auto.conf
+
+# Enable the auto switcher (true/false)
+ENABLED=true
+
+# Theme slugs (directory names under ~/.config/omarchy/themes)
+# Examples: catppuccin-latte (light), tokyo-night (dark), rose-pine (light), nord (dark)
+LIGHT_THEME=catppuccin-latte
+DARK_THEME=tokyo-night
+
+# Time window for day vs night (24h HH:MM)
+DAY_START=07:00
+NIGHT_START=19:00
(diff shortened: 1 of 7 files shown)


config/systemd/user/omarchy-theme-auto.timer
```diff
@@ -0,0 +1,12 @@
+[Unit]
```
oluwatimio, line 1: Timer will run every minute and will triger the `omarchy-theme-auto apply`

config/auto-themer/theme-auto.conf
```diff
@@ -0,0 +1,15 @@
+# Auto theme switch configuration (source-of-truth)
+# This file is read directly by: ~/.local/share/omarchy/bin/omarchy-theme-auto
+# Edit this file directly: ~/.local/share/omarchy/config/auto-themer/theme-auto.conf
+
+# Enable the auto switcher (true/false)
+ENABLED=true
```
oluwatimio: This is true but if we want we can set this to FALSE by default. Opinions?

acrogenesis:
This is great. Would be nice to have the option to auto-turn on night light

oluwatimio:
> This is great. Would be nice to have the option to auto-turn on night light

@acrogenesis I agree this will be nice! I added it in this commit https://github.com/basecamp/omarchy/pull/1836/commits/636abefb8483367c5d852d7ec2cb1641c1256fc3. Its a config option so it  s optional

morgoth:
Is it the same as closed https://github.com/basecamp/omarchy/pull/947 ?
I think a small functionality when one could setup when daylight/nightlight should be switched would be a perfect fit for omarchy. I would use it personally ;-)

oluwatimio:
> Is it the same as closed #947 ? I think a small functionality when one could setup when daylight/nightlight should be switched would be a perfect fit for omarchy. I would use it personally ;-)

Thanks for pointing that out, had no idea this had been tried. Im okay closing this if we are really opinionated about not having this natively in Omarchy since an extension is provided here https://github.com/oluwatimio/omarchy-auto-themer-tui

PR #1845, opened 2025-09-21
Fix omarchy-tui-install

djordje:
- escape `$TERMINAL` to keep it as variable in generated desktop file
- add migration to fix Docker and Disk Usage TUIs (which are broken after fresh install)

ISSUE:
I have installed fresh Omarchy v3.0.1 and I was not able to run Disk Usage. After scanning repo I found out https://github.com/basecamp/omarchy/issues/1820 which explains the issue.

### DIFF
--- a/bin/omarchy-tui-install
+++ b/bin/omarchy-tui-install
@@ -43,7 +43,7 @@ cat >"$DESKTOP_FILE" <<EOF
 Version=1.0
 Name=$APP_NAME
 Comment=$APP_NAME
-Exec=\$TERMINAL --class $APP_CLASS -e $APP_EXEC
+Exec=\$TERMINAL --class=$APP_CLASS -e $APP_EXEC
 Terminal=false
 Type=Application
 Icon=$ICON_PATH


naxels:
I just found another issue that might be included in this PR..

When I create a TUI app for SSH for example, it generates (after correcting for the escape):
`Exec=$TERMINAL --class TUI.tile -e ssh x.x.x.x`

However, when I open the application, it shows and error saying:
```
class: value required
TUI.float: invalid field
```

However when I update the line like nvim.desktop, it becomes
`Exec=$TERMINAL --class=TUI.tile -e ssh x.x.x.x`

note the --class=

Afterwards it works..

So the correction needed in `bin/omarchy-tui-install` is the additional = after --class to fix it.

I tested this as well for Docker and Disk Usage, they both work now as expected for me.

NOTE! I use ghostty as my default Terminal

djordje:
Thanks @naxels. Can you check if `--class=<something>` works fine with Alacritty and Kitty? If not I'll test those today/tomorrow to confirm we have fix for all supported terminal emulators.

naxels:
Not at my computer today, but I know it works for Alacritty as you can also see in the Omarchy source code, lot of it is class= only some class(space)

djordje:
@naxels tested with Kitty and Ghostty successfully. Added suggested change as part of this PR.

PR #1851, opened 2025-09-21
alacritty.toml wity proper bg and fg

qasimsk20:
background = "#1e1e2e"
foreground = "#cdd6f4"

are the proper colors for terminal bg and fg when themeing with Catppuccin

### DIFF
--- a/themes/catppuccin/alacritty.toml
+++ b/themes/catppuccin/alacritty.toml
@@ -1,6 +1,6 @@
 [colors.primary]
-background = "#24273a"
-foreground = "#cad3f5"
+background = "#1e1e2e"
+foreground = "#cdd6f4"
 dim_foreground = "#8087a2"
 bright_foreground = "#cad3f5"
 


ryanrhughes:
I think this really highlights a bigger issue in the Catppuccin theme; it's not consistently Mocha or Macchiato. We're using Macchiato for VSCode and the Alacritty colors (not just FG / BG) but most other places, it's Mocha which I believe would have been the intended palette. 

Would you want to take a swing at updating all of them in this theme?

qasimstatic:
I think we should consistently keep this one as Mocha only as the Latte is ONLY latte I presume?

ryanrhughes:
Mocha and Macchiato are both dark with Mocha being the darkest variant. What I'm saying is that if you compare **all** of the colors to the official docs, you'll notice that all of Alacritty is Macchiato, not Mocha. Whereas most other theme items are actually Mocha. 

They're very similar so it's an easy oversight but we should probably seek to fix all of it.

qasimstatic:
Yeah I agree I have been using the theme for a long time now and they all are really similar but Mocha is kind of the standard one which everyone uses when it comes to Dark and Catppuccin. So it is only alacritty inside The Catppuccin theme of Omarchy that is Macchiato.

PR #1853, opened 2025-09-21
Add Godot to Install Editor Menu

justinhhorner:
Let's make it easy to get started with game development using Godot on Omarchy!

This change will make it quick and easy to install the latest version of Godot via the install editor menu. Since most developers use (or at least start with) GDScript, I think it makes sense for the non-mono version to be the default place to start.

https://github.com/user-attachments/assets/a8255f76-00c4-449c-b1e4-06dab33b8831

### DIFF
--- a/bin/omarchy-menu
+++ b/bin/omarchy-menu
@@ -259,13 +259,14 @@ show_install_service_menu() {
 }
 
 show_install_editor_menu() {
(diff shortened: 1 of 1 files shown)


dhh:
I think this is stretching "Editor" too far. I would not think "game engine" when going to Install > Editor. Maybe we'll come up with some other place where it makes sense in the future, but for now, I don't think it fits there. Appreciate the suggestion though!

Issue #1841, opened 2025-09-21
Can't get internet access on any captive portal for airport / hotel Wi-Fi.


coolaj86:
### System details

2025 Framework Laptop 13

### What's wrong?

I've got a buddy on the phone right now who would be reporting this issue, except that he can't get connected to do it.

He's been through two airports and two hotels yesterday and today and hasn't been able to access the internet after clicking through the captive portals.

4 failures in a row over two days (and no successes) seems like more than just bad luck.

Omarchy was installed from the 3.0.1 ISO - so it's a fairly fresh install and no custom network setup as far as I'm aware.

This seems weird to me because my understanding is that the captive portal usually is done through DNS hijacking and a mac address whitelist... so I don't know why it would work on an iPhone just fine and not on Omarchy. I'll update if we figure out what's going wrong.

dradrien:
Can confirm I had similar issue in a hotel last week. Had to play with dns config to finally get it to work.

ryanburnette:
I tried a few variations of dns settings with no success.

moonward:
The recent systemd 258-2 update contains a regression bug that affects DNS resolution. This issue impacts some DNS configurations, while others—like Cloudflare—remain unaffected. Even on an updated Omarchy 3.01 system, you might be affected, especially if you're pointing a local DNS server, such as one on a hotel network.

These three workarounds worked fine with my system, so you may try one of them:

1. Use limine to revert to a kernel with older systemd from a few days ago. limine can be accessed from app menu.
2. In Setup > DNS, switch your DNS to Cloudflare.
3. Edit the file /etc/systemd/resolved.conf with sudo nano and add the following line to the [Resolve] section:
       DNSSEC=no
    After saving the changes, reboot your system.

-----
Try one of the options above. If running ping google.com succeeds without an error, it works fine.

Since these are temporary solutions, options 1 and 2 are easier and safer to revert once a permanent fix for systemd is released in a few days, I guess.

good luck

gianmeg:
today i tried installing omarchy 3.0.1, it was successful, its normal  and can access internet

then when i try update and reboot, suddenly after reboot i cant access internet, other device can access 

is it same case like me?

gianmeg:
[...]

Issue #1850, opened 2025-09-21
Provide option to disable DNSSEC in resolved


nkasco:
### What do you need?

Recent changes to systemd-resolved changed default behavior of the DNSSEC setting to be enabled by default, for me this broke my Unifi DNS config (couldn't resolve A records from my internal DNS)

This is discussed here:
- https://bbs.archlinux.org/viewtopic.php?id=308364
- https://gitlab.archlinux.org/archlinux/packaging/packages/systemd/-/issues/57

It took me a while to find the root cause for this since it wasn't caused by Omarchy, but since this is an opinionated distro I think it would make sense to either disable it by default or provide a setting toggle for it with some minor guidance.

Necessary resultant change config:
- Add to `/etc/systemd/resolved.conf`: `DNSSEC=no`
- Ensure this isn't overwritten in any migrations
- Provide a menu option when selecting a custom DNS (i.e. Does your DNS support DNSSEC? (A records may not resolve properly if not))

iamenderst:
Don't edit that file, create a drop-in inside /etc/systemd/resolved.conf.d/ with that option to make it persistent.

sgruendel:
Same for me, could no longer ping or ssh to my pi-hole server "minis12" using the name, just the IP address, also FQDN "minis12.fritz.box" didn't work.  I also have a laptop running Omarchy that I didn't update yet to the new resolved, if I do "resolvectl status" it shows DNSSEC=no/unsupported for all interfaces. Got it working again as @iamenderst suggested, I suppose we should generally create that drop-in file to restore the old behavior, especially as they state in https://wiki.archlinux.org/title/Systemd-resolved#DNSSEC that "DNSSEC support in systemd-resolved is considered experimental and incomplete as of Jun 2023."

PR #1856, opened 2025-09-21
Ask user if DNSSEC should be disabled for DHCP or Custom setup

sgruendel:


sgruendel:
fixes #1841 and #1850

gkurts:
I wish I had seen this before I spent several hours yelling furiously at my machine while trying to get it to use my Pihole instead of Cloudflare directly. 🤦‍♂

ryanrhughes:
This is one hell of a find that has driven me nuts on occasion!

Looking at the details for this though; it looks to me like we should just set this to allow-downgrade and have it set that way by default. No need for toggling on / off since true is really only intended for environments where you **know** the DNS server supports DNSSEC. 

Do you mind making that update?

> If set to "allow-downgrade", DNSSEC validation is attempted, but if the server does not support DNSSEC properly, DNSSEC mode is automatically disabled. Note that this mode makes DNSSEC validation vulnerable to "downgrade" attacks, where an attacker might be able to trigger a downgrade to non-DNSSEC mode by synthesizing a DNS response that suggests DNSSEC was not supported.
> 
> If set to false, DNS lookups are not DNSSEC validated.
> 
> Note that DNSSEC validation requires retrieval of additional DNS data, and thus results in a small DNS lookup time penalty.
> 
> DNSSEC requires knowledge of "trust anchors" to prove data integrity. The trust anchor for the Internet root domain is built into the resolver, additional trust anchors may be defined with [dnssec-trust-anchors.d(5)](https://www.freedesktop.org/software/systemd/man/latest/dnssec-trust-anchors.d.html#). Trust anchors may change at regular intervals, and old trust anchors may be revoked. In such a case DNSSEC validation is not possible until new trust anchors are configured locally or the resolver software package is updated with the new root trust anchor. In effect, when the built-in trust anchor is revoked and DNSSEC= is true, all further lookups will fail, as it cannot be proved anymore whether lookups are correctly signed, or validly unsigned. If DNSSEC= is set to "allow-downgrade" the resolver will automatically turn off DNSSEC validation in such a case.
> 
> Client programs looking up DNS data will be informed whether lookups could be verified using DNSSEC, or whether the returned data could not be verified (either because the data was found unsigned in the DNS, or the DNS server did not support DNSSEC or no appropriate trust anchors were known). In the latter case, it is assumed that client programs employ a secondary scheme to validate the returned DNS data, should this be required.
> 
> It is recommended to set DNSSEC= to true on systems where it is known that the DNS server supports DNSSEC correctly, and where software or trust anchor updates happen regularly. On other systems it is recommended to set DNSSEC= to "allow-downgrade".

sgruendel:
"allow-downgrade" is the default and it's broken see https://gitlab.archlinux.org/archlinux/packaging/packages/systemd/-/issues/57 . That's why it gets removed completely for Cloudflare setup to stay with default there.

sgruendel:
Looks like this is no longer needed since systemd 258-3 or 258-4, according to `man resolved.conf` now

> DNSSEC= [...] Defaults to "no".

Issue #1695, opened 2025-09-16
Bluetooth headphones show as "(null)" when switching sources and no longer autoconnect


shawnyeager:
### System details

Framework 13" AMD 7640U w/ Omarchy 2.1.2 + AirPod Pro and AirPod Max

### What's wrong?

Somewhere between Omarchy 2.0 and now, two behaviors have changed:

1. Turning on either my AirPod Pro and my AirPod Max no longer automatically connect/switch sources.
2. When I do cycle to either of them using `SUPER`+`MUTE`, the banner at the bottom of the screen shows the source "(null"), instead of the paired device name as shown in the Bluetooth GUI.

dhh:
I believe we have a fix for 2 on dev already, but I haven't heard of any changes around 1. Do let us know if you find what's the root here!

shawnyeager:
@dhh Found a fix that works for me using `wpctl`:

```
next_sink_description=$(echo "$next_sink" | jq -r '.description')
if [ "$next_sink_description" = "(null)" ] || [ "$next_sink_description" = "null" ] || [ -z "$next_sink_description" ]; then
  sink_id=$(echo "$next_sink" | jq -r '.properties."object.id"')
  next_sink_description=$(wpctl status | grep -E "\s+\*?\s+${sink_id}\." | sed -E 's/^.*[0-9]+\.\s+//' | sed -E 's/\s+\[.*$//')
fi
```

ryanrhughes:
Resolved in [v3.0.2](https://github.com/basecamp/omarchy/releases/tag/v3.0.2)

shawnyeager:
Hey, @ryanrhughes. I saw my PR was reverted. Is there a different fix?

ryanrhughes:
@shawnyeager I had accidentally merged it straight to master so reverted and opened a new one to merge through dev but in looking at it now; I'm thinking they cancelled eachother out when dev merged into master. 🤦‍♂ -- Will have to open another one and include that in the next release.

Good catch.

shawnyeager:
Cheers!

PR #1859, opened 2025-09-21
Fix for '(null)' display on AirPod source selection

shawnyeager:
Possible resolution for #1695.

### DIFF
--- a/bin/omarchy-cmd-audio-switch
+++ b/bin/omarchy-cmd-audio-switch
@@ -23,7 +23,13 @@ fi
 
 next_sink=$(echo "$sinks" | jq -r ".[$next_sink_index]")
 next_sink_name=$(echo "$next_sink" | jq -r '.name')
+
 next_sink_description=$(echo "$next_sink" | jq -r '.description')
+if [ "$next_sink_description" = "(null)" ] || [ "$next_sink_description" = "null" ] || [ -z "$next_sink_description" ]; then
+  sink_id=$(echo "$next_sink" | jq -r '.properties."object.id"')
+  next_sink_description=$(wpctl status | grep -E "\s+\*?\s+${sink_id}\." | sed -E 's/^.*[0-9]+\.\s+//' | sed -E 's/\s+\[.*$//')
+fi
+
 next_sink_volume=$(echo "$next_sink" | jq -r \
   '.volume | to_entries[0].value.value_percent | sub("%"; "")')
 next_sink_is_muted=$(echo "$next_sink" | jq -r '.mute')


cluah:
You sent me the output of "pactl list sinks" and "wpclt status" via Discord. In the output of "pactl list sinks" there are three sinks and they all have a description. So when I run the code everything works fine and the condition included as a solution in the PR is never meet for any of the three sinks.

```
if [ "$next_sink_description" = "(null)" ] || [ "$next_sink_description" = "null" ] || [ -z "$next_sink_description" ]; then
# it never reaches this point
fi
```

My thoughts on this: 1. The headphones where the issue shows up weren't connected when you ran "pactl list sinks" and the output you sent me doesn't include them. 2. The issue was solved in another way, and the solution in the PR is no longer needed.

- Please try removing the fix and see if the issue comes back.
- Make sure the headphones where the issue was showing up are connected and send me the output so I can look at it again.

ryanrhughes (approved):

cluah:
As I said, this fixes nothing as the code is never reached.

https://github.com/user-attachments/assets/e0468cd6-8db2-4ff1-9ef6-032a12ca1633

It would be helpful if Shawn could send the output of "pactl -f json list sinks" so we can take another look at it.

shawnyeager:
@cluah I'm puzzled by your dogmatic pursuit here. It works for me. Reverting brings back the previous, broken behavior. 

If it doesn't work for you, submit a PR.

cluah:
Your code gets the description using `wpctl status` when it is not found in `pactl -f json list sinks`. But in the output you shared with me, all sinks have their description.

```
[
  {
    "index": 534,
    "state": "SUSPENDED",
    "name": "alsa_output.pci-0000_c1_00.6.analog-stereo",
    "description": "Family 17h/19h/1ah HD Audio Controller Analog Stereo"
  },
  {
    "index": 538,
    "state": "SUSPENDED",
    "name": "alsa_output.usb-Shure_Inc_Shure_MV7__MV7__11-bb65d660e9c731569bbc54cdcdf1c339-01.analog-stereo",
    "description": "Shure MV7+ Analog Stereo"
  },
  {
    "index": 591,
    "state": "SUSPENDED",
    "name": "alsa_output.pci-0000_c1_00.1.hdmi-stereo-extra2",
    "description": "Radeon High Definition Audio Controller [Rembrandt/Strix] Digital Stereo (HDMI 3)"
  }
]
```

So the fix is never reached as the condition is never met.

```
if [ "$next_sink_description" = "(null)" ] || [ "$next_sink_description" = "null" ] || [ -z "$next_sink_description" ]; then
# it never reaches this point
fi
```

How can the fix be the fix if it's never reached?

A few days ago, I helped @elpritchos catch a small flaw in his PR. Here, I'm trying to do the same thing.

ryanrhughes:
Seems like it resolves the issue to me @cluah but if you see something I don't, please let me know. This appears specific to Airpods, which I don't see in your output so be sure to test with a pair connected via BT.

**With this PR** (and echo for good measure)

https://github.com/user-attachments/assets/20fd800e-9d69-444e-adf3-412d32907087

**Without this PR**

https://github.com/user-attachments/assets/40e096a1-85ee-4e3d-96db-d0fbc71f3137

cluah:
I'm not sure why he sent me the output with the AirPods disconnected.

So it seems that this comes from `pactl -f json`. 
https://gitlab.freedesktop.org/pulseaudio/pulseaudio/-/blob/master/src/pulsecore/json.c#L767

This is what I wanted to know. Thanks.

d3vw (commented):
It works!

PR #631, opened 2025-08-10
Replace gnome-keyring with pass/pass-secret-service

jzetterman:
This PR addresses issue #352 by replacing gnome-keyring with [pass](https://www.passwordstore.org/) and pass-secret-service. In testing multiple options, Gnome Keyring, Kwallet, and KeePassXC all presented significant challenges to automating the creation of a keyring that could be provisioned from the command line in a way that didn't require any user interaction. Pass was the only solution I found capable of meeting the goal of no user interaction in the provisioning of a passwordless keystore that apps such as 1Password can use for storing secrets such as MFA tokens. 

I've tested this on a fresh VM inside Proxmox and on an older Thinkpad X1 Yoga and it worked flawlessly in both scenarios with saving 1Password MFA tokens. That said, this is my first PR ever (in any capacity) and I am not a developer by trade, but an IT systems engineer/manager for the last several years. If there is room for improvement I am totally open to it. For example, I made an educated guess about where the logic inside keyring.sh should go and considered a few options. Making it it's own file seemed like the most logical way to handle it. I'm also not sure if this warrants a migration and haven't developed one. My thought is that existing Omarchy users might not want their current keyring solution to be replaced. 

Thanks for your consideration!

PR #1860, opened 2025-09-21
Login and UKI updates

### LINKED ISSUE #352: Gnome-keyring not creating default keyvault, 1Password MFA issues
I use MFA with 1Password and by default it doesn't appear that gnome-keyring is set up to work with Hyprland fully. 1Password is unable to save the token after I enter my MFA code (on my personal account) or after I complete a Duo MFA (on my work account) challenge. 

Specifically, when logging in it's not (at least on my system) initializing a default keyring (I have no keyrings at all). I was able to prove this by checking ~/.local/share/keyrings/, which is empty as well as running `secret-tool store --label="Test" testkey testvalue` which prompts me to set a password for "Default keyring". I see that gnome-keyring was added to the Yay install command in xtras.sh at some point and that a migration file exists to add it to older installations. I've replicated this on multiple virtual machines as well. 

Creating a keyring with Seahorse and making it the default does allow 1Password to save the MFA tokens. However, that keyring does not automatically unlock when first logging into the system.

I'm investigating steps to effectively address this on an alternate and virtual machines. So far I've only succeeded in breaking LUKS encryption and bricking the install. That said, I hope to develop a solution this weekend.

### LINKED ISSUE #372: Better Default Configuration for Managing SSH Keys and Secrets
SSH keys are an excellent way to manage multiple accounts, such as separate GitHub profiles for work and personal use. Many distributions already integrate SSH key management with their system secret managers, such as [GNOME Keyring](https://wiki.archlinux.org/title/GNOME/Keyring), to simplify this process.

I propose offering a similar default configuration for handling secrets in this way. The system should unlock the "keyring" automatically upon user login. Also, there should be a way for prompting passwords when required, or when storing new secrets, by using a tool similar to askpass.

### LINKED ISSUE #1015: After installing VSCode and login in to GitHub there's a missing OS keyring error
The specific error says:

It says to [troubleshoot using this guide](https://code.visualstudio.com/docs/configure/settings-sync#_troubleshooting-keychain-issues) or use a weaker encryption.

### LINKED ISSUE #1619: Multiple Lenovo types not able to boot / get in to BIOS after installation
### System details

Lenovo m720q (M1UKT77A BIOS)

### What's wrong?

I installed Omarchy 2.0 on a Lenovo m720q succesfully. However, during the reboot the machine hangs on the Lenovo logo and I'm unable to get in to the BIOS even. Disconnecting power didn't help.
I reproduced this on 2 Lenovo m720q's, as I thought it might be a quirky machine I grabbed another to try on.

Looking at #1286 it might be there is a Lenovo specific issue?

I will try a CMOS reset and report back after.

Image

### LINKED ISSUE #1719: New installs sometimes end up with Limine instead of Omarchy as the #1 boot loader entry
### System details

All

### What's wrong?

We should always start with Omarchy instead of Limine as the #1 boot loader entry.

ryanrhughes:
- Changes from using Omarchy Seamless Login DM to SDDM with Autologin
- Update to use custom named UKI to (hopefully) resolve issues due to long filename (Fixes #1619)
- Update UKI install rules to always prioritize (Fixes #1719)
- Remove Archinstall created Limine EFI entry
- Create a default keyring to house 2FA tokens and other items (Fixes #352) (Fixes #372)
- Updates VSCode installer to ensure it uses the default keyring (Fixes #1015)
- Adds guards for BTRFS and Limine
- Drops install support for alternate bootloaders

### A note about the default keyring 
This keyring is created with no password so it can be unlocked on boot without any user intervention. This relies entirely on LUKS encryption to provide protection for this info but given the only thing that should exist in here are 2FA tokens and / or SSH keys that would otherwise persists on disk, this is fine to me. There's nothing stopping a user from adding a password to their keyring if they'd prefer that but because we don't present an actual login screen outside of LUKS, the password can't be captured to unlock the keyring.

The alternative path here would be #631 which I ran personally and did work well but is a bit of a non-standard approach.

### Dropping support for alternate bootloaders
At this point, there's no reason someone should be installing fresh with anything but Limine. Because of this, I'm proposing here that we drop it from the install so as to simplify the complexity and permutations of installs we have to deal with.

As a part of this, it was added to the guard in the event someone does a manual install with the wrong bootloader chosen. They could still technically do it but they're on the hook for their own hooks and configs.

### AMI Bios Fix
The change to the UKI name _should_ fix the issue we were having with AMI BIOS based on the testing I had some users do a while back. The actual install here has not been tested yet as we haven't found a volunteer.


abenz1267:
Why the change to SDDM?

ryanrhughes:
There are a few reasons:
1. Seamless Login is literally just the VT stuff from SDDM ripped out on our own. Thus, we have to maintain and compile it on install.
2. Seamless Login doesn't handle PAM things that SDDM does on its own with relation to starting Gnome Keyring and other items. Had work to do to solve this; is solved without doing work by SDDM.
3. SDDM handles transitions and interactions with Getty and Plymouth that we currently have to disable / mask and are brittle
4. We have a number of users with legitimate use cases for multiple users (still with LUKS). I don't, but I also don't mind them having a graceful offramp if it doesn't impact the default experience. 
5. The main line experience DOES NOT CHANGE -- if I didn't tell you, you wouldn't know SDDM even existed :upside_down_face: 

**TLDR; SDDM is battle tested, well maintained, and allows us to have the exact same experience as before (single LUKS PW w/ autologin) while leaving a graceful offramp to users who may wish to utilize it differently.**

mlombardi96:
@ryanrhughes Now that we're switching to SDDM, how come we didn't opt for KDE Wallet, which supports unlocking with the encryption password and SDDM autologin? https://wiki.archlinux.org/title/KDE_Wallet#Using_SDDM_autologin_and_LUKS_encryption

Like you pointed out, having a default keyring with no password isn't a big deal, but I'm just curious.

ryanrhughes:
@mlombardi96 Good question. The short answer is; I tried it, it blew up my system, not worth it. 😂

The longer answer is that the KDE Wallet appears to be less widely supported than Gnome / Libsecret, which was also the reason for using it over the Pass solution as well. It just kind of works with everything whereas KDE appears to not be quite as universal. 

While there is a path to unlock the KDE Wallet with a LUKS password, it has some specifics when it came to the encryption that was clearly more than just swapping the modules. I tried it and couldn't figure it out in a reasonable amount of time.

Given the default setup for us is 1Password + Gnome for 2FA tokens residing on a LUKS encrypted drive; the effort to make sure the keyring is behind a PW and unlocked just isn't worth the effort to me. But if someone else has experience with this and wants to take on the migration / intricacies, I'd be happy to review and merge in the solution.

sheikhlimon:
Nice! Finally, this is getting added. 🙌
Until now, I had to use a wrapper script just to run MongoDB Compass:
``` 
#!/bin/bash
export XDG_CURRENT_DESKTOP=GNOME
export XDG_SESSION_DESKTOP=gnome
export GNOME_KEYRING_CONTROL=/run/user/$(id -u)/keyring
export ELECTRON_OZONE_PLATFORM_HINT=wayland
exec mongodb-compass --ignore-additional-command-line-flags "$@"
```

screenshot-2025-09-27_01-11-35

I tried to work around it using PAM and autostart, but couldn’t quite get it to work. It’s not a huge deal, but it’s one of those little annoyances. GNOME and KDE handle this automatically, so it’s great to see it being addressed here too. Really appreciate it!

Woosah:
Question regarding your proposal about "Dropping support for alternate bootloaders":
Does this mean I would not be able to install omaarchy on fresh arch install using grub, i.e., if everything needed for omarchy is ready to go? Because that would be a real "pain-in-the-...." for me, and I would imagine for other ppl, too. The reasons for me to use grub are (1) to be able to have a dual-boot system, (2) to enable a multi-user system, (3) to implement snapshots using timeshift and grub-btrfs as I want, and (4) to have a complete custom btrfs subvolume layout that is not easily implemented using limine. If you want to have a look into what I mean, look here for my personal notes for a grub-based omarchy installation... 
[Custom-Omarchy-Install-using-GRUB.md](https://github.com/user-attachments/files/22605980/Custom-Omarchy-Install-using-GRUB.md)

ryanrhughes:
Good question @Woosah. It'd install and work just fine. But you'd be on your own to add the `splash quiet` hooks to your setup as well as adding the UKI config if you'd like to use that. 

Otherwise; there's really no difference for you since you're not using any of the bootloader config anyway.

Woosah:
That would be no problem, since I needed to do these things since v1...  Thank you for the clarification, @ryanrhughes!

bin/omarchy-install-vscode
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+echo "Installing VSCode..."
+omarchy-pkg-add visual-studio-code-bin
+
+mkdir -p ~/.vscode
+
+cat > ~/.vscode/argv.json << 'EOF'
```
dhh, line 8: This needs to be part of a migration for existing installations too, no?

PR #1863, opened 2025-09-22
Gaming improvements

ryanrhughes:
Creating this as draft to maintain visibility and contributions to bring this to life.

The goal here is to provide a first class experience for gaming out of the box (or at least if you run the gaming setup).

There are a number of issues tackled already but still plenty to tackle still.
## Things to Do
This is a non-exhaustive list of things that need to be handled here still.

- [x] Install Proton-GE
- [x] Install Gamescope
- [ ] Add setup scripts to install core launcher set
- [ ] Parse and set compatibility options in Steam automatically (or via utility)
- [ ] Parse and set compatibility options in Heroic automatically

## Omarchy Gaming Helper
One of the more interesting things that exists is a service that runs in the background. [omarchy-gaming-helper.sh](https://github.com/basecamp/omarchy/pull/1863/files#diff-96103c6764c26ad0061e54af74d04e84d683c654954309d367beeb470c1b784e)

In some situations, we have things that we need to adjust for gaming that just aren't possible via Hyprland rules. We could make it an action for the user to do but honestly; that sucks and computers are good at doing things so we should use it for that.

The service is simple. It listens to the Hyprland IPC socket for events and dispatches commands. 

`omarchy-setup-gaming` will install and run it as an actual service but if you'd like more telemetry in order to improve it, do the following:
- `systemctl --user stop omarchy-gaming-helper`
- Run `~/.local/share/omarchy/default/hypr/scripts/omarchy-gaming-helper.sh` from a terminal window

### Example: Battle.net
When running Battle.net inside of Gamescope, launching games doesn't actually open a new window but just changes the title. No problem; but Hyprland only evaluates window rules on creation, not title updates. Now we're able to fire off logical fullscreen / un-fullscreen commands logically so the transitions feel nice.

### Example: Fractional Scaling
When you launch a game with fractional scaling, things go a little funky. 

https://github.com/user-attachments/assets/338b47b2-db90-456a-9b91-3511698aabf9

Hyprland sits and re-evaluates the window rules over and over and over. It's wild. 

The solution is to change to using an integer for the scale, but it's annoying to do that every game launch so the gaming helper watches and does that for us resulting in a smooth experience and sets it back when we're done.

https://github.com/user-attachments/assets/3e61dca2-72a0-4297-81c8-c6f696a425be

## Hyprland Confs
A separate section of Hyprland confs have been created to capture and set a common set of items. There's also a [general-gaming.conf](https://github.com/basecamp/omarchy/pull/1863/files#diff-1a4902bf56e24a279588ef2baf656b3b02a0c9e9fcdcf24d26ff9962805fb74e) that can contain some things to apply to multiple games / launchers so that all we really need to do is tag them appropriately.

We'll need quite a lot of rules here over time I suspect to catch all of the combos.

## Launchers
Heroic Launcher, and Steam are shoe-ins. Lutris is here but I haven't had great experiences with it. A lot of the scripts fail and seem outdated. 

Heroic has a great experience, can install from Epic, Amazon, GoG, and Custom, and has a good amount of support for options. You can also set global wrappers so if you want to have a default command to run in gamescope to apply to everything, we can (probably will). 

## Gamescope
Gamescope really smooths out a lot of weirdness that can exist for non-native games. However, we do have to pass in some options and will want to have an elegant way of determining those on our own vs requiring users to enter them. As a for instance, this is mine populated in Steam or Heroic `gamescope -w 3840 -h 2160 -W 3840 -H 2160 -- %command%`

The downside is that you do have to define resolutions to have it render properly. I also had to set a framerate to get Overwatch to recognize high refresh as an option.

**Steam**
image

**Heroic** ( -- is at the end)
image


arielj:
> **Heroic** ( -- is at the end) image

just wanted to add that heroic has a settings tab specific for gamescope configuration

image

that can be set globally to be applied to all new games, and then it can be changed in a per-game basis too, to avoid manual input errors

ryanrhughes:
@arielj Does Heroic just launch everything in Gamescope by default? I'd seen this tab but wasn't really clear if it was wrapping the games automatically. Maybe I just needed to add flags to the additional options here.

arielj:
> @arielj Does Heroic just launch everything in Gamescope by default? I'd seen this tab but wasn't really clear if it was wrapping the games automatically. Maybe I just needed to add flags to the additional options here.

@ryanrhughes no, heroic does not use gamescope by default, only if enabled (enabling any of those options or adding something to the additional arguments will enable gamescope)

we actually have a PR to have an explicit `enable gamescope` option to make it more clear https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/4224

so, currently heroic enables gamescope if anything for gamescope is configured

ryanrhughes:
@arielj That's great to know! I found that confusing and wasn't ever really sure what purpose it served if I couldn't turn it on. 

Really, I think my grand vision for this v1 centers a lot around Heroic and Steam. Given it looks like you're one of the maintainers on Heroic; maybe you can gauge my ambition and see if it feels achievable. 

Lutris sucks. There are multiple scripts for the same things, they fail, the app is dated. I really just kind of don't want anything to do with it. But I do like the idea they had around scripted installs. 

Heroic presents a really fantastic experience that is basically what I'd want, and has the bonus of Epic, GoG, and Prime Gaming directly. Love that. 

I want to be able to build some sidecar set of installer scripts that effectively creates the game setup and Heroic settings **for** the user. 

For example; on install, I'd like to just go ahead and set the flags for Gamescope to be the proper resolution and framerate based on the User's setup. Technically I'd like these to be dynamic based on the screen it's launched from...but that's later. I'm thinking this could be achieved through modifying the config files.

For installing games, I see something like `omarchy-install-game-battle-net` inserting the proper config to setup the Battle.net installer. Same thing for other manual installs like Ubisoft and others. Then the user could effectively 1-click install, and just carry out the rest of the steps.

My real final-final thing is wanting to suck Steam games installed into the Heroic launcher so that they could be launched via `steam://{game_id}` or however they needed to be but so that I have a single, unified launcher experience for all of my games.

arielj:
> Really, I think my grand vision for this v1 centers a lot around Heroic and Steam. Given it looks like you're one of the maintainers on Heroic; maybe you can gauge my ambition and see if it feels achievable.

I don't really know what's possible from the omarchy/arch side of things cause I don't use neither Arch nor Omarchy, but I can share some things from heroic's point of view.

> For example; on install, I'd like to just go ahead and set the flags for Gamescope to be the proper resolution and framerate based on the User's setup. Technically I'd like these to be dynamic based on the screen it's launched from...but that's later. I'm thinking this could be achieved through modifying the config files.

The settings for heroic are plain json files and follow predefined paths (the only caveat is that we use the game's appname for the file's name that is not obvious how to get it from outside the launcher cause it's internal data of each store, a string or an id). But it should be possible to edit the config file before opening heroic (though the first time I guess it needs to run without that so it creates the files)

> For installing games, I see something like `omarchy-install-game-battle-net` inserting the proper config to setup the Battle.net installer. Same thing for other manual installs like Ubisoft and others. Then the user could effectively 1-click install, and just carry out the rest of the steps.

in heroic we have the `sideloaded` games (with the Add Game button) that could be used to add Ubisoft connect for example, technically it's just some content in a json file pointing to a wine prefix with the tool installed inside it. I guess those installers could edit the files?

> My real final-final thing is wanting to suck Steam games installed into the Heroic launcher so that they could be launched via `steam://{game_id}` or however they needed to be but so that I have a single, unified launcher experience for all of my games.

Heroic has a feature to add games to steam automatically, we use https://www.npmjs.com/package/steam-shortcut-editor to edit steam's shortcuts file and some other things to generate the images for it, and then we support the `heroic://` protocol that can be used to launch games.

I imagine some other tool could do something similar but in the other direction, reading the libraries' json files from heroic's config folder and add them to steam, if that's even needed I could go into more details, but we basically have one json file for each store + one for sideloaded games/app with all the info to generate the shortcut.

flozero:
Thank you !!! 

Maybe out of scope here but a way to imrpove drivers installation would be helpful specially for gaming

edit: Maybe worth having some sort of cleaning script like removing lutris for heroic as it seems you prefer to have this

jrdora31:
Feature request: ensure proper system keyboard layout (e.g. fr-FR AZERTY) is respected in-game under Proton/Wayland, since current input mapping falls back to en-US when launching via Faugus Launcher — needs confirmation whether Heroic behaves the same.
https://www.reddit.com/r/omarchy/comments/1ohd4rb/omarchy_world_of_warcraft_battlenet_keyboard/

jakegibson:
Some notes from my previous cachyOS / jakoolit hyprland setup:
- cachyos + cachyos gaming meta package required very minimal tweaking for most games to "just work" for me. https://wiki.cachyos.org/configuration/gaming/
- cachyos does go hard enough on forked packages / custom kernel tweaks to a level I don't think it makes the best basis for a general purpose distro stability/compatibility wise, but I think a similar gaming meta package that does sane defaults for env variables, optimal proton version (GE or cachyos), launch options etc would be great
- jakoolit was more opinionated on window rules than I think makes sense for omarchy (which types of apps go to which desktop for example). But forcing fullscreen at least seems like a sane default 
https://github.com/JaKooLit/Hyprland-Dots/blob/c862835221decf03b6640bb43e4044861cceaa04/config/hypr/UserConfigs/WindowRules.conf#L54

https://github.com/JaKooLit/Hyprland-Dots/blob/c862835221decf03b6640bb43e4044861cceaa04/config/hypr/UserConfigs/WindowRules.conf#L202

Slammed with work over the next month but if things slow down for the holidays hoping to spend some time on this plus taking a stab at a mapping db of app names/ids across stores for 

> the only caveat is that we use the game's appname for the file's name that is not obvious how to get it from outside the launcher cause it's internal data of each store, a string or an id

so we could build up some default launch configs for popular games.

hipsterusername:
@ryanrhughes - Wondering if there's still interest in pushing this forward. I do think it would benefit to separate concerns and push a Steam gaming solution forward first. There's growing interest in folks moving to Linux, and I think "seamless gaming" as a configurable option would be an incredible draw.

dhh:

PR #1866, opened 2025-09-22
Add fcitx5-configtool to enable keyboard configuration from GUI

### LINKED ISSUE #827: Waybar always shows US keyboard
Should be the chosen one.

woopstar:
Fixes #827

### DIFF
--- a/install/omarchy-base.packages
+++ b/install/omarchy-base.packages
@@ -25,6 +25,7 @@ evince
 eza
 fastfetch
 fcitx5
+fcitx5-configtool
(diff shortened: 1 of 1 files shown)


c4software:
It was some version ago, and was removed here.

https://github.com/basecamp/omarchy/commit/178f97a6e3334befe531ffbc7f87d193a09278fb

dhh:
Yeah, I'm not a fan of this tool. We have other ways of accomplishing the configurations. You're always free to install on your own, though!

PR #1867, opened 2025-09-22
Fix screen recording for multiple display setups

### LINKED ISSUE #1301: omarchy-cmd-screenrecord not working with multiple displays (wf-recorder)
As per the [OMARchy Manual](https://learn.omacom.io/2/the-omarchy-manual/53/hotkeys#screenrecordings), the hotkey for screen recording should be pressed a second time to stop the recording. This works correctly for 'record a specific area' (Alt + PrintScr). However, when attempting to stop a 'record the full monitor' session (Ctrl + Alt + PrintScr) I only see an notification saying that the record was started, but by pressing the hotkey again, a I see a new notification saying the same thing instead of the recording being stopped this causes so I never finish recording thus not saving any files, actually I checked btop if the process is running and it seems that is not. I tried to run the command in the terminal, because I thought that might be something crashing and worked fine, save the file and everything. I don't think that this can be a problem because it does start so should be recognize the input but I use fn + } as PrintScr button.

I did some digging and found out that is because of wf-record wants you to select an monitor when are multiple connected:

```
Please select an output from the list to capture (enter output no.):
1. Name: eDP-1 Description: Chimei Innolux Corporation 0x1522  (eDP-1)
2. Name: DP-1 Description: LG Electronics LG ULTRAWIDE 304AZXC9U386 (DP-1)
Enter output no.:
```

So I guess this fails eventually and I never get wf-recorder running so the script ended up not working.

For the time being I had this script that leverage hyprclt, gum, allacrity and a temp file to make it work for selecting an terminal. I use a custom smaller floating window that I have but I put the Omarchy class here if anyone want to use this

```bash
#!/bin/bash

[[ -f ~/.config/user-dirs.dirs ]] && source ~/.config/user-dirs.dirs
OUTPUT_DIR="${OMARCHY_SCREENRECORD_DIR:-${XDG_VIDEOS_DIR:-$HOME/Videos}}"

if [[ ! -d "$OUTPUT_DIR" ]]; then
  mkdir -p "$OUTPUT_DIR"
  notify-send "Screen recording directory created: $OUTPUT_DIR" -t 3000
fi

screenrecording() {
  filename="$OUTPUT_DIR/screenrecording-$(date +'%Y-%m-%d_%H-%M-%S').mp4"

  notify-send "Screen recording starting..." -t 1000
  sleep 1

  wf-recorder -f "$filename" -c libx264 -p crf=23 -p preset=medium -p movflags=+faststart "$@"
}

if pgrep -x wf-recorder >/dev/null; then
  pkill -x wf-recorder
  notify-send "Screen recording saved to $OUTPUT_DIR" -t 2000
  exit 0
fi

temp_file=$(mktemp)

alacritty --class Omarchy -e bash -c "hyprctl monitors | grep 'Monitor' | awk '{print \$2}' | gum filter --placeholder 'Select a monitor...' --prompt 'Monitor: ' > \"$temp_file\""

selected_monitor=$(cat "$temp_file")

rm "$temp_file"

if [[ -n "$selected_monitor" ]]; then
  # Start the recording on the selected monitor
  screenrecording -o "$selected_monitor"
else
  notify-send "Screen recording cancelled." -t 2000
  exit 1
fi
```

kyedavey:
Fixes #1301

ISSUE: Screen recording a display on a multiple display setup currently fails due to the output to be recorded not being specified.

FIX: Pull name of focused monitor from hyprctl and set output flag of wf-recorder / wl-screenrec

### DIFF
--- a/bin/omarchy-cmd-screenrecord
+++ b/bin/omarchy-cmd-screenrecord
@@ -47,7 +47,7 @@ screenrecording_active() {
 if screenrecording_active; then
   stop_screenrecording
 elif [[ "$SCOPE" == "output" ]]; then
-  start_screenrecording
+  start_screenrecording -o "$(hyprctl monitors -j | jq -r '.[] | select(.focused == true).name')"
 else
   region=$(slurp) || exit 1
   start_screenrecording -g "$region"


ryanrhughes:
@kyedavey Could you test https://github.com/basecamp/omarchy/pull/1920 and see if it resolves this issue as well?

kyedavey:
@ryanrhughes Tested, #1920 also resolves this issue

kyedavey:
Closing in favor of #1920

PR #1872, opened 2025-09-22
Update aliases

iqre8:
Add alias for vi


dhh:
Don't think this is a good idea. vi is a separate editor, and some people will install that.

iqre8:
How about if we test for vi then? Traditionally every *nix* system has an editor that's available via the 'vi' command. Not having an alias for vi will make people start wondering if it's pico, nano etc.

if ! command -v vi &> /dev/null; then
    alias vi="vim"
fi
    export EDITOR=vim

Issue #1875, opened 2025-09-22
Make done message close automatically after some time


crqch:
### What do you need?

Regarding
https://github.com/basecamp/omarchy/blob/master/bin/omarchy-show-done

In my opinion, when a user installs or removes their app, and doesn't press Enter in 5 minutes, we can assume that they already know about it's deletion or installation, therefore the window can close by itself.

The current implementation resulted in a case where 3 windows were left hanging open, just because I've started doing things in other workspaces.

dhh:
I disagree here. If you start an install, and you walk away, you should still be able to see the completion when you come back.

PR #1876, opened 2025-09-22
Done message timeout after 5 mins

skutaada:
Adds the enhancement mentioned in: #1875.


dhh:
I disagree here. If you start an install, and you walk away, you should still be able to see the completion when you come back.

PR #1878, opened 2025-09-22
Add umu-launcher to Gaming Menu

R1kaB3rN:
Adds umu-launcher, a tool to run non-Steam games using the same technologies as Steam, to the gaming menu. Featured in clients such as [Heroic Games Launcher](https://heroicgameslauncher.com/), [Lutris](https://lutris.net/), and [Drop](https://droposs.org/). Although intended to be used by clients, adding it to the gaming menu makes it available for users who get their games from non-Steam storefronts and who prefer not to install a full-blown gaming launcher to run their games, as well as for game developers who want a quick and easy way to test their games through Proton from their terminal.


dhh:
I think this is a tad too low-level for inclusion in this menu. Appreciate the suggestion though! ✌️

PR #1882, opened 2025-09-22
systemd-resolvconf has been added to support wg-quick out of the box

aquaspy:
wg-quick relies on the resolvconf. The recommended solution for arch linux seems to be using systemd-resolvconf

I was troubleshooting and found [this](https://superuser.com/questions/1500691/usr-bin-wg-quick-line-31-resolvconf-command-not-found-wireguard-debian) thread about  it.

It worked perfectly.

### DIFF
--- a/install/omarchy-base.packages
+++ b/install/omarchy-base.packages
@@ -129,3 +129,4 @@ xournalpp
 yaru-icon-theme
 yay
 zoxide
+systemd-resolvconf


dhh:
Thanks for the suggestion, but we don't ship with wg-quick in the box, so I don't think it makes sense to add accommodations for it with extra packages by default.

PR #1884, opened 2025-09-22
Fix Tailscale split DNS compatibility

jardahrazdera:
## Summary
Fixes Tailscale split DNS compatibility issues by removing `[!UNAVAIL=return]` from nsswitch.conf configuration while maintaining mDNS printer discovery functionality.

## Changes
- Updated printer setup script to use `resolve` instead of `resolve [!UNAVAIL=return]`
- Updated existing Avahi migration script  
- Added new migration to fix existing user configurations

## Background
This addresses the issue raised by @lewispb in https://github.com/basecamp/omarchy/pull/1021#issuecomment-3318650490 where the `[!UNAVAIL=return]` parameter was breaking Tailscale split DNS for multiple users at 37signals.

## Testing
- DNS resolution continues working properly
- mDNS printer discovery functionality preserved
- Migration will automatically fix existing user configurations

## Impact
- **New installs**: Will use the correct configuration from the start
- **Existing users**: Will be automatically fixed during next system update via migration
- **Tailscale users**: Split DNS will work properly without manual intervention

### DIFF
--- a/migrations/1758562032.sh
+++ b/migrations/1758562032.sh
@@ -1,0 +1,5 @@
+echo "Fix Tailscale split DNS compatibility by removing [!UNAVAIL=return] from nsswitch.conf"
+
+if grep -q '\[!UNAVAIL=return\]' /etc/nsswitch.conf; then
+  sudo sed -i 's/resolve \[!UNAVAIL=return\]/resolve/g' /etc/nsswitch.conf
+fi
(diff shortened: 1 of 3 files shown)


migrations/1758562032.sh
```diff
@@ -0,0 +1,5 @@
+echo "Fix Tailscale split DNS compatibility by removing [!UNAVAIL=return] from nsswitch.conf"
+
+if grep -q '\[!UNAVAIL=return\]' /etc/nsswitch.conf; then
+  sudo sed -i 's/resolve \[!UNAVAIL=return\]/resolve/g' /etc/nsswitch.conf
+fi
```
lewispb, line 5: I've tested and this works nicely.

PR #1890, opened 2025-09-23
Update fix-apple-spi-keyboard.sh

stallmer:
Added "MacBook10,1" to the if statement so that these changes will also apply to a 2017 12" MacBook when installing Omarchy.


ryanrhughes:
#1947 was already merged for this.

PR #1892, opened 2025-09-23
Fixes the nvim desktop entry exec command

### LINKED ISSUE #1852: Neovim not working in Nautilus
### System details

Omarchy v3.0.1

### What's wrong?

When I try to open a file with Neovim in Nautilus nothing happens.

Checked neovim.desktop and the line that needs to execute looks fine, though $TERMINAL seems not to work in a .desktop file according to Chatgpt, don't no if this is true.

```
~ ❯ cat .local/share/applications/nvim.desktop
[Desktop Entry]
Name=Neovim
GenericName=Text Editor
Comment=Edit text files
Exec=$TERMINAL --class=nvim --title=nvim -e nvim -- %F
Terminal=false
Type=Application
Keywords=Text;editor;
Icon=nvim
Categories=Utility;TextEditor;
StartupNotify=false
MimeType=text/english;text/plain;text/x-makefile;text/x-c++hdr;text/x-c++src;text/x-chdr;text/x-csrc;text/x-java;text/x-moc;text/x-pascal;text/x-tcl;text/x-tex;application/x-shellscript;text/x-c;text/x-c++;
```
I fixed it by changing it to alacritty

dharmavagabond:
The desktop entries doesn't expand any ennvars, so it needs to be executed via the shell.

Edit: It uses `omarchy-launch-editor` instead of `sh`.

resolves #1852

### DIFF
--- a/applications/nvim.desktop
+++ b/applications/nvim.desktop
@@ -2,7 +2,7 @@
 Name=Neovim
 GenericName=Text Editor
 Comment=Edit text files
-Exec=$TERMINAL --class=nvim --title=nvim -e nvim -- %F
+Exec=sh -c "$TERMINAL --class=nvim --title=nvim -e nvim -- %F"
 Terminal=false
 Type=Application
 Keywords=Text;editor;
--- a/applications/nvim.desktop
+++ b/applications/nvim.desktop
@@ -2,7 +2,7 @@
 Name=Neovim
 GenericName=Text Editor
 Comment=Edit text files
-Exec=sh -c "$TERMINAL --class=nvim --title=nvim -e nvim -- %F"
+Exec=omarchy-launch-editor %F
 Terminal=false
 Type=Application
 Keywords=Text;editor;


naxels:
I disagree with this PR for 2 reasons
- i can't replicate the issue myself with base Omarchy files/config
- It only fixes 1 terminal .desktop file if it is an actual issue

dharmavagabond:
To replicate, launch the entry in question:
`gio launch ~/.local/share/applications/nvim.desktop`. It won't open because the [Desktop Entries](https://specifications.freedesktop.org/desktop-entry-spec/latest/exec-variables.html) doesn't expand any envvar (not just $TERMINAL).

It fixes the only desktop entry that omarchy provides.

Rishabh-Sarang:
can't have this PR merged because of [this](https://github.com/basecamp/omarchy/issues/1852#issuecomment-3337113614)

dharmavagabond:
@Rishabh-Sarang I've updated the PR. It uses `omarchy-launch-editor` so it works no matter how you open the desktop entry.

NicolasDorier:
I tried this PR, after installing 3.0.2 and this didn't work for me. That said, I am new in omarchy, so probably unrelated.

I applied this PR on `~/.local/share/omarchy/applications/nvim.desktop`, but nothing change...
Actually even if I try to change the icon in this file, it has no impact in the app launcher for example...

Rishabh-Sarang:
> I tried this PR, after installing 3.0.2 and this didn't work for me. That said, I am new in omarchy, so probably unrelated.
> 
> I applied this PR on `~/.local/share/omarchy/applications/nvim.desktop`, but nothing change... Actually even if I try to change the icon in this file, it has no impact in the app launcher for example...

You have to change the file in `~/.local/share/application/nvim.desktop`

NicolasDorier:
Thank you @Rishabh-Sarang , I confirm that this work!

PR #1898, opened 2025-09-23
Add auto-changing colors for Neovim 

bjarneo:
This PR is a PoC based on this branch: https://github.com/basecamp/omarchy/pull/1712

Details

1. It will install all themes from the all-themes.lua on neovim load (run rm -rf ~/.local/share/nvim/lazy/* to test). Priority and lazy is removed. Lazy will load it only as soon as the theme is set, which caused issues.
2. It will automatically set the color scheme for the native themes on the omarchy theme change
3. It will automatically set the color scheme for the 3rd party installed themes that includes a neovim.lua file
4. It will create a neovim.lua file with the layzvim config setting it to pixel if neovim.lua does not exist (as if lazyvim don't trigger any events if nothing has changed)

I've tried to use remote calls as I hoped that could trigger neovim changes (like :colorscheme <theme_name>) automatically, but that does not work unless the nvim instance is started as a server first. https://neovim.io/doc/user/remote.html

One alternative is to send a dbus event using [dbus-send](https://dbus.freedesktop.org/doc/dbus-send.1.html) from the the bash script (omarchy-theme-set, when switching themes), then from the hot reload lua plugin you listen to events matching the one coming from the dbus-send, then auto reload. In that case we can  replace 4 with an event.

@ryanrhughes can you give it a spin?

### DIFF
--- a/bin/omarchy-theme-set-neovim
+++ b/bin/omarchy-theme-set-neovim
@@ -12,10 +12,6 @@ if [[ ! -f "$THEME_PATH/neovim.lua" ]]; then
     cat > "$THEME_PATH/neovim.lua" << 'EOF'
 return {
     {
-        "bjarneo/pixel.nvim",
-        name = "pixel",
-    },
-    {
         "LazyVim/LazyVim",
         opts = {
             colorscheme = "pixel",


bin/omarchy-theme-set
```diff
@@ -9,14 +9,17 @@ THEMES_DIR="$HOME/.config/omarchy/themes/"
...
+THEME_PATH="$THEMES_DIR$THEME_NAME"
 
 # Check if the theme entered exists
 if [[ ! -d "$THEME_PATH" ]]; then
   echo "Theme '$THEME_NAME' does not exist in $THEMES_DIR"
   exit 1
 fi
 
+# Run neovim before the symlinks to create the neovim.lua if not exists
+omarchy-theme-set-neovim "$THEME_PATH"
```
bjarneo, line 21: I think it is safe to put it next to the other script like vscode

bjarneo:
@ryanrhughes give me a ping if you want me to experiment with the dbus event signal, so we don't need start copying files to directories. (4.)

dhh:
Can we rebase this on top of what we have now? Process should be that we just use ansi colors if a theme doesn't have neovim.lua.

bjarneo:
We surely can. I'll look into it! I noticed I managed to delete the branch, so I'll reopen the entire thing.

PR #1904, opened 2025-09-23
Add cursor support in theme-set-vscode

adelplace:
Add support for Cursor in theme switch 🙂


dhh:
Appreciate the work here, but I ended up preferring the implementation in #2015. I've given you co-author credit, though, since you were first 👍

Issue #1168, opened 2025-08-27
Bash or ZSH?


hyprhex:
What do you like to use more?  Which one is more customized and updated?

ferrao:
`zsh` allows me to install [fish-like autosuggestions](https://github.com/zsh-users/zsh-autosuggestions), which is the only thing that I actually miss on the `bash` + `starship` combo.

ramtinJ95:
You can get highlighting and autosuggestions just like zsh/fish with ble.sh, check it out!

hyprhex:
Yeah, zsh is amazing. I wish @dhh could make it a default one with omzsh, or the powerlevel10k themes

lbonvarl:
Despite the warnings of limited support, try this one out: https://github.com/romkatv/zsh4humans. Way better in performance than omyzsh and excellent default config. See this: I have bash in its standard starship config and zsh with zsh4humans as my new standard shell (changed just via chsh). tested on an old HP6200 with i7 2600 and SSD.

```
❯ hyperfine "bash -l"
Benchmark 1: bash -l
  Time (mean ± σ):      62.0 ms ±  12.9 ms    [User: 37.0 ms, System: 33.2 ms]
  Range (min … max):    55.5 ms … 121.4 ms    24 runs

  Warning: The first benchmarking run for this command was significantly slower than the rest (121.4 ms). This could be caused by (filesystem) caches that were not filled until after the first run. You should consider using the '--warmup' option to fill those caches before the actual benchmark. Alternatively, use the '--prepare' option to clear the caches before each timing run.

❯ hyperfine "zsh -l"
Benchmark 1: zsh -l
  Time (mean ± σ):       1.8 ms ±   0.5 ms    [User: 1.0 ms, System: 0.8 ms]
  Range (min … max):     0.8 ms …   3.6 ms    1153 runs

  Warning: Command took less than 5 ms to complete. Note that the results might be inaccurate because hyperfine can not calibrate the shell startup time much more precise than this limit. You can try to use the `-N`/`--shell=none` option to disable the shell completely.
```

diogoodev:
> `zsh` allows me to install [fish-like autosuggestions](https://github.com/zsh-users/zsh-autosuggestions), which is the only thing that I actually miss on the `bash` + `Starship` combo.

I used Fish for a while, but about a month ago I migrated to Bash. Now I’m using ble.sh together with bash-completion, and I’m pretty satisfied with this setup.

[...]

PR #1907, opened 2025-09-23
Prompt user to choose shell during installation

vyrx-dev:
# Choose Shell 
- Automatically installs the selected shell if not already present.
- omarchy-meny > Setup > Shell

https://github.com/user-attachments/assets/835b97dd-71a0-4826-bea8-4b1659054ceb

### DIFF
--- a/bin/omarchy-choose-shell
+++ b/bin/omarchy-choose-shell
@@ -57,7 +57,3 @@ case $choice in
   exit 1
   ;;
 esac
-  echo "Invalid choice. Please try again."
-  exit 1
-  ;;
-esac
--- a/bin/omarchy-choose-shell
+++ b/bin/omarchy-choose-shell
@@ -1,8 +1,8 @@
 #!/usr/bin/env bash
 
 BOLD='\e[1m'
-GREEN='\e[92m'
 YELLOW='\e[93m'
+GREEN='\e[92m'
 RESET='\e[0m'
 
 print_info() {
@@ -21,34 +21,28 @@ install_shell() {
   fi
 }
 
-print_info "Choose your user's default shell:"
-echo "1. bash"
-echo "2. fish"
-echo "3. zsh"
-echo "4. quit"
+shell=$(gum choose --height 5 --header "Choose your usdefault shell" bash fish zsh quit)
 
-read -p "Enter the number of your preferred shell: " choice
-
-case $choice in
-1)
+case $shell in
+bash)
   install_shell bash
   chsh -s "$(command -v bash)"
   print_choice "Shell choice: bash"
   printf '%s\n' "If default shell was changed, you will need to logout and log back in for change to take effect."
   ;;
-2)
+fish)
   install_shell fish
   chsh -s "$(command -v fish)"
   print_choice "Shell choice: fish"
   printf '%s\n' "If default shell was changed, you will need to logout and log back in for change to take effect."
   ;;
-3)
+zsh)
   install_shell zsh
   chsh -s "$(command -v zsh)"
   print_choice "Shell choice: zsh"
   printf '%s\n' "If default shell was changed, you will need to logout and log back in for change to take effect."
   ;;
-4)
+quit)
   echo "User has chosen to quit program."
   exit 1
   ;;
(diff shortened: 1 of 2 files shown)


vyrx-dev:
Solves issue #1168

woopstar:
There are defaults for bash located here:

https://github.com/basecamp/omarchy/tree/master/default/bash

Should we have the same for the other shell options?

vyrx-dev:
> There are defaults for bash located here:
> 
> https://github.com/basecamp/omarchy/tree/master/default/bash
> 
> Should we have the same for the other shell options?

Ohhh yes, I completely missed that 😅 and as far as I know, there aren't any for the other shells right now.

woopstar:
> there aren't any for the other shells right now.

Which should then be created, so the implementation equals giving the shell more or less.

bin/omarchy-choose-shell
```diff
@@ -0,0 +1,68 @@
...
+}
+
+install_shell() {
+  shell=$1
+  if ! command -v "$shell" >/dev/null 2>&1; then
+    print_info "$shell not found. Installing..."
+    if command -v pacman >/dev/null 2>&1; then
+      sudo pacman -S --noconfirm "$shell"
+    elif command -v apt-get >/dev/null 2>&1; then
+      sudo apt-get update && sudo apt-get install -y "$shell"
```
abenz1267: why pollute this with non-arch package managers?

vyrx-dev: Should I remove it?

abenz1267: IMO yes. Omarchy is meant to be run on Arch, so there's no point in checking for other distro's package managers.

woopstar: Agreed. This is Arch only.

One could argue to use Yay instead of pacman, but remove apt-get

vyrx-dev: Done ✅

dhh:
Appreciate the idea, but we have a bunch of ties on bash that would break with this. So this PR isn't usable in its current form. And I'm not interested in maintaining any big changes to accommodate radically different shells, like fish or nushell. Happy to explore zsh as a direct extension to bash, though.

ctrl-alt-adrian:
> Appreciate the idea, but we have a bunch of ties on bash that would break with this. So this PR isn't usable in its current form. And I'm not interested in maintaining any big changes to accommodate radically different shells, like fish or nushell. Happy to explore zsh as a direct extension to bash, though.

Adding support for zsh would be big.

woopstar:
`bin/omarchy-choose-shell` is not executable in the PR. It needs a chmod +x

vyrx-dev:
> `bin/omarchy-choose-shell` is not executable in the PR. It needs a chmod +x

Is there a way to avoid needing chmod +x here? I saw the other scripts in bin don’t need it, so I’m trying to figure out what’s different with this one.

woopstar:
> Is there a way to avoid needing chmod +x here? I saw the other scripts in bin don’t need it, so I’m trying to figure out what’s different with this one.

I see them as executable. But another option is to modify the $cmd in this file:

https://github.com/basecamp/omarchy/blob/6e75222d8f88678061309bfcf108a708f10b8a18/bin/omarchy-launch-floating-terminal-with-presentation#L4C98-L4C102

To use a specific shell to launch applications, but I prefer not to take that approach. I think the best approach is to ensure that all bins are executable. We could make a script that takes care of that and runs on update.

pixincreate:
hey, i recently opened [this](https://github.com/basecamp/omarchy/discussions/1728) discussion where i added startship preset management with a state parameter to my configs. have been thinking of raising a pr adding support for omarchy. the thing is, transient prompt is well supported in zsh but not in bash. would love to see zsh being an option in omarchy.

edit: https://github.com/basecamp/omarchy/pull/2013

dhh:
Going to close this out, as the only path forward I'm currently looking to consider is switching to zsh. We're not going to build everything around whatever-shell-you'd-like. That remains a DIY adventure.

ktaraszk:
> Going to close this out, as the only path forward I'm currently looking to consider is switching to zsh. We're not going to build everything around whatever-shell-you'd-like. That remains a DIY adventure.

Whaat? Haven't you guys heard about shebang? Just use `#!/bin/bash` at every script that requires bash to work...

crmne:
Hey y'all. If you want to use Fish with Omarchy, I replicated @dhh's bash config here: https://github.com/crmne/omafish

Just install that.

PR #1908, opened 2025-09-24
feat: add relaunch-required and reboot-required to waybar

alexperreault:
This PR adds icons to waybar if a relaunch or a restart is required

When updating, it gives you the option to restart (or relaunch) right away.

If you say no to defer the restart, you never got any other reminder and the states weren't cleared.

This reminds the user that a restart (or relaunch) is pending and they can choose to proceed at a convenient time for them

### DIFF
--- a/bin/omarchy-reboot-required
+++ b/bin/omarchy-reboot-required
@@ -1,0 +1,10 @@
+#!/bin/bash
+
+# Get reboot state file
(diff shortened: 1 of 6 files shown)


dhh:
This is a nice implementation, but I don't think we need it. It's imo enough that we offer you the chance to act on this after a full Omarchy upgrade. I don't think there's enough value in having a nag icon on top of that.

PR #1917, opened 2025-09-24
Notification improvements

nqst:
### Changes:

- Move common settings from theme files to `core.ini`. Themes now only define colors
- Switch font to CaskaydiaMono (default monospaced font), which pairs better with the overall UI
- Set font size in pixels for reliability and easier control
- Increase padding and line height for better readability 
- Remove extra whitespace characters in some notifications (manual fine-tuning is no longer needed since a monospaced font is now used)
- Adjust wording in the Setup Wi-Fi notification
- Update Tokyo Night foreground color in notifications (reusing the color from Limine/Kitty/eza)

### Preview:

| Before | After |
|--------|--------|
|  |  |


tristonarmstrong:
personally, i prefer the before

andresatierf:
I think this is a nice addition, but would be great to have it follow the system font. When the user changes the font via the menu, it should also change the notification font

tristonarmstrong:
> I think this is a nice addition, but would be great to have it follow the system font. When the user changes the font via the menu, it should also change the notification font

The font might be whats throwing me off about it + the spacing between words

nqst:
> I think this is a nice addition, but would be great to have it follow the system font. When the user changes the font via the menu, it should also change the notification font

Good point, thanks! Addressed via https://github.com/basecamp/omarchy/pull/1917/commits/609d5ab0ec4a282a1de04cdf5460b15e7c6e272a. It now inherits the default system monospace font, like Walker does.

default/mako/core.ini
```diff
@@ -1,7 +1,13 @@
 anchor=top-right
 default-timeout=5000
-width=420
+width=430
 height=110
+outer-margin=20
+padding=12,15
+border-size=2
+max-icon-size=32
+font=monospace 13px
```
nqst, line 9: My personal preference for the default is a monospaced font, but if you prefer a proportional font, the change is as simple as:

```suggestion
font=sans-serif 14px
```

vaqMAD:
Text feels less squeezed actually I like it.

dhh:
Appreciate this, but I intentionally chose a regular font, not a monospace, for the notifications. So not looking to have that included in the monospace font changing. Happy to take a separate PR to cleanup the mako settings, though, if that makes sense.

PR #1921, opened 2025-09-24
Add Hyprlock faillock message

tarekbecker:
I ran into the problem that faillock kicked in and I couldn't understand why my password was rejected.

This adds an error message under the password input box if the account is logged and for how long.

If the account is not locked, no message is shown. The `<span> </span>` is required to ensure that not a single vertical black line is rendered.

If someone wants to test this, you can change with CTRL+ALT F1/F2 to a terminal and login as root and use ` faillock --user <username> --reset` to unlock manually the account.,

Arch ships by default with 10 min delay after 3 attempts. The new install routine changes these defaults.

### DIFF
--- a/bin/omarchy-hyprlock-error-message
+++ b/bin/omarchy-hyprlock-error-message
@@ -1,0 +1,72 @@
+#!/bin/bash
+
+USERNAME="$(whoami)"
+
+# Check if the user exists
+if ! id "$USERNAME" >/dev/null 2>&1; then
+  echo "User $USERNAME does not exist" >&1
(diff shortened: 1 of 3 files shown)


dhh:
We've raised the default to 10 attempts now. I'd still take a patch that makes it clear that the account has been locked, but it can't be this much code. I think all we need is something that fits with the existing designs (so can't use a red like this) and says "Account locked". That's enough of a clue for someone to figure out what to do. Feel free to open a PR that just focuses on that. Thanks for working on this!

PR #1923, opened 2025-09-24
Use QT6, drop QT5, update "kvantum-qt5" to "kvantum"

cevdetta:
Package changements:
- use QT6 version of `kvantum` instead of `kvantum-qt`
- drop QT5 for QT6 (up-to-date and better wayland integration)
in order to remove unnecessary dependencies of old packages.

### DIFF
--- a/install/omarchy-base.packages
+++ b/install/omarchy-base.packages
@@ -89,6 +89,7 @@ power-profiles-daemon
 python-gobject
 python-poetry-core
 python-terminaltexteffects
+qt5-wayland
 qt6-wayland
 ripgrep
 satty
--- a/install/omarchy-base.packages
+++ b/install/omarchy-base.packages
@@ -52,7 +52,7 @@ inetutils
 iwd
 jq
 kdenlive
-kvantum-qt5
+kvantum
 lazydocker
 lazygit
 less
@@ -89,7 +89,7 @@ power-profiles-daemon
 python-gobject
 python-poetry-core
 python-terminaltexteffects
-qt5-wayland
+qt6-wayland
 ripgrep
 satty
 signal-desktop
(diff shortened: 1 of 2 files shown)


woopstar:
Would this also require a migration for people who already have it installed?

cevdetta:
> Would this also require a migration for people who already have it installed?

If we want them to migrate to QT6 version of `kvantum` and get rid of unnecessary dependencies, yes it'd require.

dhh:
This would yank wayland support for qt5 apps, making them fallback to xwayland. Don't think that adds up.

woopstar:
> This would yank wayland support for qt5 apps, making them fallback to xwayland. Don't think that adds up.

I believe you can have Qt5-wayland and Qt6-wayland running side-by-side. Would that be beneficial for this PR?

cevdetta:
> > This would yank wayland support for qt5 apps, making them fallback to xwayland. Don't think that adds up.
> 
> I believe you can have Qt5-wayland and Qt6-wayland running side-by-side. Would that be beneficial for this PR?

The thing is `qt5-wayland` is needed for packages who are depended on `qt5` (`qt5-base`), but if we don't have any packages and use `qt6-wayland` version of them (which this PR does), though I'm not really against installing `qt5-wayland` default.

> This would yank wayland support for qt5 apps, making them fallback to xwayland. Don't think that adds up.

dhh:
But if someone installs a qt5 application, they're going to fall back to xwayland, right? But I'm happy to see us target where there's most support, as long as we don't break something big and obvious.

cevdetta:
> But if someone installs a qt5 application, they're going to fall back to xwayland, right? But I'm happy to see us target where there's most support, as long as we don't break something big and obvious.

I understood and made a PR to install `qt5-wayland` by default again.

woopstar:
Is this still relevant?

PR #1925, opened 2025-09-25
Add systemd-sleep hook to restore network connectivity after resume

segusantos:
This PR adds a `systemd-sleep` hook that restarts network services after the system resumes from sleep.

- Restarts `systemd-networkd` and `iwd` services when present and enabled.
- Resolves connectivity issues with captive portals after resume.

I've had the issue of having to fully reboot in order to reconnect to captive portals after suspend for a while now, and restarting network services on resume fixed it for me.

I’m not fully certain if restarting these services is the ideal solution or if there’s a more subtle fix, but I'm open to discussion in order to fix this in the best way possible.


dhh:
This is an interesting idea, but I think it's too big of a hammer to hit with. We should be finding the specific issue and try to solve it at the root. This is not something I've seen on the Framework 13, for example. But would be nice to have this as a documented path for folks with this problem until we have a root-cause solution. Please do add it to Discussions > Manual ✌️

PR #1933, opened 2025-09-25
feat(waybar): add privacy indicator

eifr:
Add privacy indicator for screenshare or mic usage

### DIFF
--- a/config/waybar/config.jsonc
+++ b/config/waybar/config.jsonc
@@ -4,8 +4,16 @@
   "position": "top",
   "spacing": 0,
   "height": 26,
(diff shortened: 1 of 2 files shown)


dhh:
I found that this isn't reliable. If you start Zoom, for example, but then turn off both mic and video, it's still going to show. Also, we already have the screenrecording indicator. So think we'll just stay with that for now. It's worse to have something unreliable than nothing at all.

orazioedoardo:
> If you start Zoom, for example, but then turn off both mic and video, it's still going to show.

This is likely due to the app not stopping engaging the mic and camera and just not sending media, so the indicator is actually correct. Many other apps wrongly do it this way, Signal Desktop being another example.

PR #1945, opened 2025-09-25
Add Helium browser to supported browsers list in omarchy-launch-webapp

Furyfree:
This PR adds Helium browser to the supported browsers list in `omarchy-launch-webapp`.

### DIFF
--- a/bin/omarchy-launch-webapp
+++ b/bin/omarchy-launch-webapp
@@ -3,7 +3,7 @@
 browser=$(xdg-settings get default-web-browser)
 
 case $browser in
-google-chrome* | brave-browser* | microsoft-edge* | opera* | vivaldi* | helium*) ;;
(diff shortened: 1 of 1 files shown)


bin/omarchy-launch-webapp
```diff
@@ -3,7 +3,7 @@
 browser=$(xdg-settings get default-web-browser)
 
 case $browser in
-google-chrome* | brave-browser* | microsoft-edge* | opera* | vivaldi* | helium*) ;;
+google-chrome* | brave-browser* | microsoft-edge* | opera* | vivaldi* | helium-browser*) ;;
```
d0ctr, line 6: why so? the expected name for the .desktop file is [helium.desktop](https://github.com/imputnet/helium-linux/blob/main/package/helium.desktop)

PR #1950, opened 2025-09-25
Add SSH session integration for Ghostty

mirzap:
The custom `xterm-ghostty` TERM value in Ghostty can break terminal functionality when connecting via SSH to systems that lack the matching terminfo entry.

```
$ tmux ls
missing or unsuitable terminal: xterm-ghostty
```

To address this we need to enable [shell integration](https://github.com/ghostty-org/ghostty/issues/7608) feature.

### DIFF
--- a/config/ghostty/config
+++ b/config/ghostty/config
@@ -21,4 +21,4 @@ shell-integration-features = no-cursor
 keybind = f11=toggle_fullscreen
 
 # SSH session terminfo
-shell-integration-features = ssh-env,ssh-terminfo
+shell-integration-features = ssh-env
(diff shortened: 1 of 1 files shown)


dhh:
What's the best default here @mitchellh?

mitchellh:
I recommend setting `shell-integration-features = ssh-env`. It is the safest option. It does override `ssh` (for compatible shells) so its default off, but its fairly benign (it just overrides the TERM env when using `ssh`). See here: https://ghostty.org/docs/config/reference#shell-integration-features 

There is a more correct but invasive one `ssh-terminfo`. I don't recommend it as a default for a distro. 

There's more general information on what's going on here: https://ghostty.org/docs/help/terminfo

mirzap:
@dhh @mitchellh Thanks for feedback. Updated.

mitchellh:
Note this is an area we're still actively trying to improve Ghostty on from multiple angles. My answer is just the current state of the union.

PR #1951, opened 2025-09-25
Add alias for caffeinate command

sarrietav-dev:
MacOS has this cool utility command, so instead of installing a new package, let's use this alias!

### DIFF
--- a/default/bash/aliases
+++ b/default/bash/aliases
@@ -26,6 +26,7 @@ alias ....='cd ../../..'
 # Tools
 alias d='docker'
 alias r='rails'
+alias caffeinate="systemd-inhibit sleep infinity"
 n() { if [ "$#" -eq 0 ]; then nvim .; else nvim "$@"; fi; }
 
 # Git


woopstar:
And what happens on non-Apple MacBook computers with this PR?

ctarx:
There is already a script that toggles hypridle on and off to disable idle behavior, so adding a caffeinate alias is redundant.

sarrietav-dev:
> And what happens on non-Apple MacBook computers with this PR?

This PR doesn't depend on the machine you are using. As long you are using Omarchy you're good to go.

sarrietav-dev:
> There is already a script that toggles hypridle on and off to disable idle behavior, so adding a caffeinate alias is redundant.

I think this PR is a nice addition especially for users coming from macOS. But thanks for pointing out that we can also use Ctrl+Super+I for this.

dhh:
We have the `Trigger > Toggle > Idle Lock` for this 👍

PR #1956, opened 2025-09-26
Add window theme to Ghostty config

scossar:
Adds `window-theme = ghostty` to `omarchy/config/ghostty/config`, so that the current Ghostty theme is applied to the tab bar.

The Ghostty documentation for the setting is here: https://ghostty.org/docs/config/reference#window-theme. 

Without the option:

screenshot-2025-09-25_20-11-51

With the option:

screenshot-2025-09-25_20-17-15

### DIFF
--- a/config/ghostty/config
+++ b/config/ghostty/config
@@ -7,6 +7,7 @@ font-style = Regular
 font-size = 9
 
 # Window
+window-theme = ghostty
 window-padding-x = 14
 window-padding-y = 14
 confirm-close-surface=false


Thundernirmal:
@scossar, I think better to add gtk-toolbar-style = flat as well, which looks better
image

scossar:
@Thundernirmal, I like that. I've made the change in my own config. Not setting `window-theme = ghostty` feels like a UI bug. Adding `gtk-toolbar-style = flat` feels more like a stylistic choice. I'll leave the PR as it is for now, but if others want the toolbar style change, I can update the PR.

PR #1957, opened 2025-09-26
Add more Mac models to SPI keyboard support

joelgaff:
Added:

MacBook 8,1
MacBook 9,1

MacBook 8,1 uses slightly different moduels from what I can tell, so I’ve added a case statment.  Please tweak as needed.


grin:
FYI, another PR with very similar changes: https://github.com/basecamp/omarchy/pull/1959 by @wey-gu 
probably worth combining

PR #1960, opened 2025-09-26
fix: change uwsm app to uwsm-app for performance

### LINKED ISSUE #1953: Use uwsm-app instead of 'uwsm app' for performance
### What do you need?

Hi,

Noticed that for example `omarchy-launch-webapp` and possibly others use `uwsm app`. Could it be changed to `uwsm-app`, as it is faster?

Image

Thanks!

woopstar:
Fixes #1953


woopstar:
This somehow breaks something. Need to check whatever it is.

default/hypr/autostart.conf
```diff
@@ -1,10 +1,10 @@
-exec-once = uwsm app -- hypridle
-exec-once = uwsm app -- mako
-exec-once = uwsm app -- waybar
-exec-once = uwsm app -- fcitx5
-exec-once = uwsm app -- swaybg -i ~/.config/omarchy/current/background -m fill
-exec-once = uwsm app -- swayosd-server
-exec-once = uwsm app -- walker --gapplication-service &
+exec-once = uwsm-app -- hypridle
```
woopstar: This apparently breaks stuff

woopstar: Could be, this is the issue:

https://github.com/Vladimir-csp/uwsm/blob/master/scripts/uwsm-app.sh#L9

woopstar:
https://github.com/Vladimir-csp/uwsm/issues/107

Seems related

woopstar:
Seems the fix from https://github.com/Vladimir-csp/uwsm/issues/107#issuecomment-2746316911 works. If I run uwsm-app /bin/true before launching Hyprland, everything works out.

Which must be related to that wayland-wm-app-daemon.service is not started.

alexjp:
Strange, I have been using `uwsm-app` for so long, and everything working along.

I do have hyprland `exec-once` call a autostart.nu script, instead of everything on `exec-once` (this way I can control better the autostart sequence, time it and display the loading time).

For reference, my autostart.nu https://gist.github.com/alexjp/c75547afd9e42ad165c05afbb6ed5c91

One solution might be to keep `uwsm app` on Hyprland autostart, but use `uwsm-app` for launching applications mid session.

Thanks!

woopstar:
> One solution might be to keep `uwsm app` on Hyprland autostart, but use `uwsm-app` for launching applications mid session.

Yeah. I'm considering to do a fix for https://github.com/Vladimir-csp/uwsm/issues/107#issuecomment-2746316911 to make it able to run in parallel.

alexjp:
> Yeah. I'm considering to do a fix for [Vladimir-csp/uwsm#107 (comment)](https://github.com/Vladimir-csp/uwsm/issues/107#issuecomment-2746316911) to make it able to run in parallel.

If it fixes/works without any side effects, I will do the same on my "autostart.nu" :)

Thank You!

Vladimir-csp:
While a bunch of `exec-once = uwsm-app -- ...` should work fine with the next uwsm release, I would still recommend using native systemd services for, well, services. That is, apps that are not instanced interactively, but autostarted as a single instance with the graphical session.

alexjp:
> While a bunch of `exec-once = uwsm-app -- ...` should work fine with the next uwsm release, I would still recommend using native systemd services for, well, services. That is, apps that are not instanced interactively, but autostarted as a single instance with the graphical session.

Hi, and thank you for uwsm!

Sorry if this is a dumb question, but would `uwsm-app -t service` also be "recommended" ?

Vladimir-csp:
> Sorry if this is a dumb question, but would `uwsm-app -t service` also be "recommended" ?

It's not so much about unit type, but having actual unit files in `*/systemd/user/` with dependencies and conditions, so systemd would just do what it does, no commands required.

As for uwsm, there is also `-u` option to set predictable unit name without random instancing part (if this supposed to be a proper single-instance service).

PR #1962, opened 2025-09-26
Notify on "warning" and "critical" battery levels according to waybar battery thresholds

ptsmonteiro:
Send notifications for "warning" and "critical" battery levels according to the current waybar battery module configuration (instead of the hardcoded 10%).

### DIFF
--- a/bin/omarchy-battery-monitor
+++ b/bin/omarchy-battery-monitor
@@ -1,13 +1,18 @@
 #!/bin/bash
 
 # Designed to be run by systemd timer every 30 seconds and alerts if battery is low
+# Uses warning and critical thresholds as defined in the waybar battery module config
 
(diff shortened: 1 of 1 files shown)


ptsmonteiro:
We could even consider ditching the omarchy-battery-monitor altogether because waybar/config.jsonc already supports events:

> "events": {
> 		"on-discharging-warning": "notify-send -u normal 'Low Battery'",
> 		"on-discharging-critical": "notify-send -u critical 'Very Low Battery'",
> 		"on-charging-100": "notify-send -u normal 'Battery Full!'"
> }

But I didn't want to go that far quite yet.

dhh:
I'd rather take a look at what the waybar setup could look like. Did not know that this was built in! Let's explore switching to that instead. What are the warning and critical thresholds?

PR #1964, opened 2025-09-26
Added show_setup_browser_menu

adriandersen:
https://github.com/basecamp/omarchy/discussions/985#discussion-8778271

Adds a simple menu for choosing default web browser from a list of installed and known web browsers.

### DIFF
--- a/bin/omarchy-menu
+++ b/bin/omarchy-menu
@@ -172,7 +172,7 @@ show_font_menu() {
 }
 
 show_setup_menu() {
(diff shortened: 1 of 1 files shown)


adriandersen:
I feel like this would make more sense under "Defaults" menu, but right now that option only opens the configfile for uwsm

dhh:
Appreciate this, but I don't quite think it's needed, given how every browser has a built-in way to set itself as the default. They typically scream about it when you launch them for the first time too!

PR #1967, opened 2025-09-26
Set waybar battery icon color according to warning and critical thresholds

ptsmonteiro:
Give minimal visual feedback about the battery state by changing the waybar battery icon color according to the current battery state.

- Normal (no color change)
- Warning (orange)
- Critical (red)


dhh:
Interesting suggestion, but I don't think this needs more attention than the monochrome icon that matches all themes gives. Good tweak to make on your own setup, though! ✌️

PR #1968, opened 2025-09-26
Added Fuji-Amethyst Theme

Shreyas-ITB:
I created a new purple color based theme, it currently has two background images which are not so high quality but they are pretty clear in 1080p resolution. I really like this theme and I hope you guys would love it too!


dhh:
You can add host your new extra theme like so: https://learn.omacom.io/2/the-omarchy-manual/92/making-your-own-theme

PR #1979, opened 2025-09-26
Kill correct screensaver process when locking screen  

vladkhard:
/bin/omarchy-launch-screensaver uses alacritty and /bin/omarchy-lock-screen uses $TERMINAL


woopstar:
Both should most likely use $TERMINAL then :)

vladkhard:
> Both should most likely use $TERMINAL then :)

looks like alacritty will stay for these things for a while. I'm all for replacing it with $TERMINAL everywhere

PR #1266, opened 2025-08-28
Add multi-profile support to web app installer

mulsori:
## Add multi-profile support for web app installer

This PR enhances the web app installer to support multiple browser profiles, allowing users to create separate web apps from different Chromium profiles.

### Changes:
- Installer: Added profile detection and selection
  - Automatically detects available Chromium profiles (Default, Profile 1, etc.)
  - Shows profile selection menu when multiple profiles exist
  - Uses single profile automatically when only one exists

### Use case:
Users can now install the same web app (e.g., Teams, Discord) multiple times with different browser profiles, allowing separate work/personal accounts to run simultaneously.

### Example:
- Teams (Work profile) 
- Teams (Personal profile)
screenshot-2025-08-28_21-55-10
screenshot-2025-08-28_21-56-31
screenshot-2025-08-28_21-58-50

- Both appear as separate desktop applications

PR #1980, opened 2025-09-26
Fix: Allow passing flags/arguments to omarchy-launch-or-focus-webapp script

### LINKED ISSUE #1977: omarchy-launch-or-focus-webapp does not accept flags for profile-directory
### System details

Omarchy 3.0.1

### What's wrong?

When launching webapps (via keybinding, console, etc) it typically accepts flags such as the chromium profile-directory flag for opening browser with specific user profile, 

The added omarchy-launch-or-focus-webapp does not accept the flags and will not open the webapp with   the correct user profile when trying to pass flag:

example:
bindd = SUPER, M, Music, exec, omarchy-launch-or-focus-webapp "https://music.youtube.com" --profile-directory="Default"

Expected: open the music.youtube.com url using the chromium profile "Default"
Actual: ignores flags and opens the chromiun with the last used profile, and does not navigate the the specified url

meirdick:
What it solves: 
This change addresses the reported issue where the omarchy-launch-or-focus-webapp script failed to correctly process and pass additional flags, such as --profile-directory, to the underlying web app launcher. This resulted in web apps not opening with the intended user profile or configuration. 

How it works: The script is modified to use shift to correctly separate the [window-pattern] from the subsequent arguments ([url] [flags...]). It then passes all remaining arguments as the single launch command to omarchy-launch-or-focus. 

Related Issue: Closes #1977

### DIFF
--- a/bin/omarchy-launch-or-focus-webapp
+++ b/bin/omarchy-launch-or-focus-webapp
@@ -1,8 +1,12 @@
 #!/bin/bash
 
 if (($# == 0)); then
-  echo "Usage: omarchy-launch-or-focus-webapp [window-pattern] [url]"
+  echo "Usage: omarchy-launch-or-focus-webapp [window-pattern] [url-and-flags...]"
   exit 1
 fi
 
-exec omarchy-launch-or-focus "$1" "omarchy-launch-webapp '$2'"
+WINDOW_PATTERN="$1"
+shift
+LAUNCH_COMMAND="omarchy-launch-webapp $@"
+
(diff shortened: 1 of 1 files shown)


Prajwal-Prathiksh:
#1266 - and this seem to be solving similar problems.

patroza:
I had to change the implementation like so for it to work well for me, without it, the -or-focus variant would not work correctly..

```
app=$1
rest=${@:2}
exec setsid uwsm-app -- $(sed -n 's/^Exec=\([^ ]*\).*/\1/p' {~/.local,~/.nix-profile,/usr}/share/applications/$browser 2>/dev/null | head -1) --app="$app" "$rest"
```

PR #1984, opened 2025-09-26
Ghostty theme sync

j-c-m:
Synchronize ghostty themes to their alacritty counterpart, or directly from their source of truth.

Use dynamic runtime override colors for cursor and selection for a more 'themed' experience.

### DIFF
--- a/themes/catppuccin-latte/ghostty.conf
+++ b/themes/catppuccin-latte/ghostty.conf
@@ -1,1 +1,28 @@
-theme = Catppuccin Latte
+# Ghostty theme generated from Alacritty
+# https://github.com/catppuccin/alacritty/blob/main/catppuccin-latte.toml
+# themed selection and cursor overrides
+
+palette = 0=#bcc0cc
+palette = 1=#d20f39
+palette = 2=#40a02b
+palette = 3=#df8e1d
+palette = 4=#1e66f5
+palette = 5=#ea76cb
+palette = 6=#179299
+palette = 7=#5c5f77
+palette = 8=#acb0be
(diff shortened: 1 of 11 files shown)


dhh:
Were you also looking at this, @tahayvr ?

tahayvr:
@dhh saw this because of your comment. What about it?

tahayvr:
@j-c-m Does this change the look? How is it different than using the `theme = name` method?

j-c-m:
It does change the look on some of the themes (to make them perfect rgb matches to their alacritty counterparts).

For example in catpuccin-late the Ghostty source does not match the official source on black,white,and bright black. catpuccin dark was using Catpuccin Mocha for ghostty, but Catpuccin Machiatto for Alacritty.

Additionally it adds the dynamic:
```
cursor-color = cell-foreground
cursor-text = cell-background
selection-foreground = cell-background
selection-background = cell-foreground
```
So cursor and selection are themed dyamically as of Ghostty 1.2.0.

The difference between using theme = name method is really about control, explicit definitions in Omarchy vs implied definitions from upstream.

I can also make these changes upstream if theme = name is preferred, although we'd still want to resolve catpuccin mocha vs machiatto etc here.

Bigger picture at why I am looking at explicit definitions inside Omarchy is to ultimately create a single theme definition and then generate all these app specific theme files.

dhh:
We have switch to Catpuccin Mocha for Alacritty now too. I think we're better off just leaning on the built-in themes, unless any of them are terribly wrong. Happy to take a specific fix to a specific problem, but otherwise let's stick with the named themes ✌️

Issue #1988, opened 2025-09-26
Wiremix displays generic endpoint name for GoXLRMini


iamobservable:
### System Information

<img width="1299" height="661" alt="Image" src="https://github.com/user-attachments/assets/201cffdf-a4c9-4db7-a760-46d429c665d7" />

### What is wrong?

Wiremix displays generic endpoint names for the GoXLRMini device connected as an external sound board. The GOXLRMini provides multiple audio devices where applications can have associated audio routed. The default wiremix configuration displays "GoXLRMini" for each of the audio devices, making it difficult to known which device is being selected. It would be helpful to read the GoXLRMini device description versus the nickname.

_I will be submitting a PR with an added wiremix configuration file._

**No wiremix configuration setup**
- Top left image show the ~/.config/wiremix/wiremix.toml file completely commented out as though a configuration files does not exist.
- Bottom left image shows the wiremix playback interface with endpoint selection
- Right image links to the wiremix configuration naming overrides documentation

<img width="2560" height="1440" alt="Image" src="https://github.com/user-attachments/assets/d138db58-00b0-45b5-808f-02daa8a92884" />

**Added wiremix configuration setup**

- Top left image show the ~/.config/wiremix/wiremix.toml file configured to override the naming for devices named "GoXLRMini"
- Bottom left image shows the wiremix playback interface with endpoint selection
- Right image links to the wiremix configuration naming overrides documentation

<img width="2560" height="1440" alt="Image" src="https://github.com/user-attachments/assets/8eeeb517-5c34-47e2-abf6-f05ec1b0a01b" />

iamobservable:
PR submitted -> https://github.com/basecamp/omarchy/pull/1989

iamobservable:
Closed given [PR feedback](https://github.com/basecamp/omarchy/pull/1989)

Created a [Discussion Manual](https://github.com/basecamp/omarchy/discussions/2690) to capture information about the configuration suggestion.

PR #1989, opened 2025-09-26
Create a wiremix configuration to display GoXLRMini names

iamobservable:
Suggested solution for #1988 

GoXLRMini device reports generic name for individual audio devices (Chat, Game, Music, Sample, etc). This patch allows the endpoint device name to be displayed uniquely "GoXLR Game" in the Wiremix playback tab instead of only as "GoXLRMini."

**Before**

screenshot-2025-09-26_13-25-56

**After**

screenshot-2025-09-26_13-25-36

This may relate to other audio devices that can also be configured over time.

### Wiremix configuration reference

https://crates.io/crates/wiremix#names

### DIFF
--- a/default/wiremix/wiremix.toml
+++ b/default/wiremix/wiremix.toml
@@ -1,0 +1,6 @@
+# Configuration reference: https://github.com/tsowell/wiremix/blob/05a40ca2a147ce3e015802acc62b2e7bb3711bde/wiremix.toml#L141
+[[names.overrides]]
+types     = [ "endpoint" ]
+property  = "node:node.nick"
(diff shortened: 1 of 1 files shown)


dhh:
Nice fix, but it's a bit too specific to something we don't ship with in the box. Would make a good addition to the Discussion > Manuals, though!

iamobservable:
> Nice fix, but it's a bit too specific to something we don't ship with in the box. Would make a good addition to the Discussion > Manuals, though!

For anyone following the thread, here is the Discussion > Manuals url -> https://github.com/basecamp/omarchy/discussions/2690

PR #1990, opened 2025-09-26
fix(battery): handle multiple batteries in low-battery notifier

phantom-kali:
The low-battery notification script fails on multi-battery laptops due to malformed `upower` args from multiple device paths.

**Changes:**
- Loop over batteries to compute min discharging percentage.
- Alert only if below threshold; skip if none discharging.
- No behavior change for single-battery setups.

Code handles edge cases (no batteries, all charging). Tested on simulated dual-battery env.


raybarrera:
https://github.com/basecamp/omarchy/pull/1783

dhh:
Duplicate of #1783

PR #1991, opened 2025-09-26
Keybind to real-time voice dictation in Omarchy

mrbende:
This PR proposes built-in dictation to Omarchy.

Inspired by [@ThePrimeagen 's stream](https://www.youtube.com/live/6V_NrgpPU_c?si=7Zaz38L1Bh74PZAw) and their talk to Cursor pedal, I wanted to see if there was an easy way to get voice dictation into an Omarchy keybind. I found [nerd-dictation](https://github.com/ideasman42/nerd-dictation/tree/main) by @ideasman42, installed ydotool, and after a bit of debugging it just worked.

Typing a long set of instructions into OpenCode or Cursor can get annoying, and I find myself often giving sub-par instructions just because I don't want to type a novel into each prompt... but now I can just speak it into my mic triggered by a hypr keybind.

Proposing this PR in case @dhh and others in the Omarchy community want to integrate this functionality into the out of the box Omarchy experience. I have not yet tested this on a fresh install, but I believe that all of the necessary configuration has been added. If anybody has a machine they don't mind hard-refreshing with a fresh Arch install, a test would be greatly appreciated. 

It leverages [nerd-dictation](https://github.com/ideasman42/nerd-dictation) to perform voice to text transcription with a small local model downloaded on OS install at `~/.config/nerd-dictation`, uses built-in `parec` for audio recording, and an added dependency `ydotool` to simulate input on the cursor with Wayland. This is added to the hypr autostart to launch the input emulation daemon in the background, but this could be optimized in the future to only launch the daemon on dictation. Notifications are sent via mako when recording begins and stops.

https://github.com/user-attachments/assets/7e045c68-6baf-41a7-9a1d-db3afa18b3ea

### DIFF
--- a/bin/omarchy-dictation
+++ b/bin/omarchy-dictation
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 # Voice dictation control script for Hyprland
-# Hold SUPER SHIFT F to start record, press again to stop and submit (ENTER)
+# Hold CTRL ALT, J to start record, press again to stop and submit (ENTER)
 
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 PID_FILE="/tmp/voice-dictation.pid"
--- a/config/hypr/bindings.conf
+++ b/config/hypr/bindings.conf
@@ -27,8 +27,8 @@ bindd = SUPER, X, X, exec, omarchy-launch-webapp "https://x.com/"
 bindd = SUPER SHIFT, X, X Post, exec, omarchy-launch-webapp "https://x.com/compose/post"
 
 # Map voice dictation - slight hold to start, press to stop and ENTER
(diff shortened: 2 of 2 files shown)


mrbende:
Keybind was changed from SUPER SHIFT F to CTRL ALT J... intended to keep same homerow easy access, but avoids SUPER conflict. Before, sometimes a bunch of windows would pop up if I was still holding SUPER when dictation was active.

jelenv:
This is great, I was looking for exactly such voice dictation option. I'll try it

mrbende:
> This is great, I was looking for exactly such voice dictation option. I'll try it

Nice! Let me know what you think. I tried some of the larger vosk models and found them to be a lot worse but your mileage may vary

dhh:
We now have voxtype built in.

PR #719, opened 2025-08-12
Add themes support for Zed Editor

melonamin:
## Summmary

This PR adds theme support for the Zed editor, allowing Omarchy's theme system to seamlessly integrate with Zed.

## What's Changed

  - 11 Zed theme files added:
    - Catppuccin (using official Macchiato variant)
    - Catppuccin Latte (using official theme)
    - Everforest
    - Gruvbox
    - Kanagawa
    - Matte Black
    - Nord
    - Osaka Jade
    - Ristretto
    - Rose Pine
    - Tokyo Night (using official theme)
  - Theme switching integration - Modified omarchy-theme-set to:
    - Create Zed theme symlinks when switching themes
    - Update Zed's settings.json for live theme reloading
    - Support both zed and zeditor command names
  - Installation support - Updated install/desktop/theme.sh to set up initial Zed theme links
  - Migration script - Added migration for existing installations to enable Zed support

## Technical Details

  - In Zed selected themes use standardized name "Omarchy Current" for consistency
  - Live reloading via settings.json modification
  - Official themes used where available for better quality (Catppuccin Macchiato, Catpuccin Latte Tokyo Night)
  - Other themes generated from Alacritty color definitions using custom converter

## Testing

  - Tested theme switching across all 11 themes
  - Verified live reloading works without restarting Zed
  - Confirmed migration script works for existing installations
  - Created test script for automated screenshot capture

### Catppuccin
Catppuccin

### Catppuccin Latte
Catppuccin_Latte

### Everforest
Everforest

### Gruvbox
Gruvbox

### Kanagawa
Kanagawa

### Matte Black
Matte_Black

### Nord
Nord

### Osaka Jade
Osaka_Jade

### Ristretto
Ristretto

### Rose Pine
Rose_Pine

### Tokyo Night
Tokyo_Night

PR #2006, opened 2025-09-27
Update Zed install to include Omazed (theme sync package) install and setup

APS6:
This PR adds automatic installation of [Omazed](https://github.com/APS6/omazed).

PS: This is my first time making a PR.

Omazed is a lightweight Bash + systemd package that:

- Automatically syncs Zed’s theme with your current Omarchy theme.
- Includes the 11 default Omarchy themes.
- Generates custom Zed themes from Alacritty configs on the fly (fast with bash).

As a Zed user myself I wanted to make the experience better for everyone and not just me. Since theme switching for all apps was previously considered out of scope in #719,  I decided to make a separate, opt-in package. The goal was to keep it as **lightweight** and **easy to set up** as possible.

### How it works

- A systemd service uses inotify-tools to watch the current Omarchy theme.
- On change, it runs a simple Bash script to update Zed’s theme.
- Resource usage is very low (about 5 MB of memory in my testing):
omazed in btop. using only 5MB memory

To make the install simple I added the package to **AUR** so that everything can be setup **in seconds** with `yay -S omazed && omazed setup`

### Demo
Video of theme sync in action, including external themes:
https://github.com/user-attachments/assets/4266c939-d992-4551-b87f-d8feddc78572

Auto-generated Sakura theme (from Alacritty) in Zed:
Automatically generated sakura theme

### Change in this PR

The Zed installer entry now also installs and sets up Omazed:
`*Zed*) present_terminal "echo 'Installing Zed...'; sudo pacman -S --noconfirm zed && yay -S --noconfirm omazed && omazed setup && setsid gtk-launch dev.zed.Zed" ;;`

That's it! Omazed handles everything else.

This way, users who choose Zed from the Omarchy installer get theme sync out of the box. Omarchy itself carries **no extra burden** since Omazed is fetched from the AUR.

Omazed is **MIT-licensed and open source**, so changes or improvements can be made freely.

### Proposal
To make Omazed even lighter, I’d like to propose either:
- A file that always contains the current theme name (not a symlink), so I can drop the inotify-tools dependency and switch to systemd.path, or
- Even better, a way to hook into omarchy-theme-set directly, removing the need for file watching entirely.

### DIFF
--- a/bin/omarchy-menu
+++ b/bin/omarchy-menu
@@ -262,7 +262,7 @@ show_install_editor_menu() {
   case $(menu "Install" "  VSCode\n  Cursor\n  Zed\n  Sublime Text\n  Helix\n  Emacs") in
   *VSCode*) install_and_launch "VSCode" "visual-studio-code-bin" "code" ;;
   *Cursor*) install_and_launch "Cursor" "cursor-bin" "cursor" ;;
-  *Zed*) install_and_launch "Zed" "zed" "dev.zed.Zed" ;;
+  *Zed*) present_terminal "echo 'Installing Zed...'; sudo pacman -S --noconfirm zed && yay -S --noconfirm omazed && omazed setup && setsid gtk-launch dev.zed.Zed" ;;
   *Sublime*) aur_install_and_launch "Sublime Text" "sublime-text-4" "sublime_text" ;;
   *Helix*) install "Helix" "helix" ;;
   *Emacs*) install "Emacs" "emacs-wayland" && systemctl --user enable --now emacs.service ;;


dhh:
Add a new hooks system that you can tap into: https://github.com/basecamp/omarchy/pull/2527

APS6:
> Add a new hooks system that you can tap into: #2527

I feel like the single theme-set file might get messy if multiple packages are trying to use the same file. Might be better to either call all hooks prefixed with `theme-set-*` or inside `hooks/theme/` so that extensions can create their own hooks without interfering with others. 
Or would you prefer I just append `omazed sync` to the theme-set file?

dhh:
Just append to the file.

APS6:
Done.
Omazed 1.1.1 now uses the theme-set hook.
Thanks for that!

morgoth:
@dhh I'm wondering if this PR should be merged, or maybe there should be a place in manual with all the tools that are integrating via hook?
This way there is no maintenance burden on omarchy side and a possibility to create powerful community integrations via hooks. 

BTW I tested `omazed` package and it works nicely. Thank you @APS6 for creating it.

dhh:
@APS6 I think we might have an easier route now: https://github.com/basecamp/omarchy/pull/4053. If we could reduce this to just a single zed.config.tpl file, I wouldn't mind including it by default.

PR #2007, opened 2025-09-27
Replace battery monitor upower and awk usage with simple file read

0xMarius:
Hey, a fan of these scripts


woopstar:
Conflicts with #1783

dhh:
This doesn't seem to work on all setups. Here it is for my Framework 13:

```
~ ✗ cat /sys/class/power_supply/BAT0/capacity
cat: /sys/class/power_supply/BAT0/capacity: No such file or directory
```

On my setup, it's on BAT1. This is why we need these tricks to find the right BAT first.

PR #2009, opened 2025-09-27
Add optional xpadneo driver prompt to Steam installer

ericjim:
Prompts users during Steam installation to optionally install the xpadneo driver, which fixes common Xbox/Bluetooth controller issues including detection problems and missing rumble support.

### DIFF
--- a/bin/omarchy-install-steam
+++ b/bin/omarchy-install-steam
@@ -1,5 +1,52 @@
 #!/bin/bash
 
+set -e
+
 echo "Now pick dependencies matching your graphics card"
+echo "Installing Steam..."
 sudo pacman -Syu --noconfirm steam
+
+# Ask about controller support
+echo ""
+echo "Do you want to install enhanced controller support?"
+echo "This installs xpadneo driver to fix common Xbox/Bluetooth controller issues."
+echo "Learn more: https://github.com/atar-axis/xpadneo"
+echo ""
+echo "Fixes reported by users:"
+echo "  • Steam not detecting Xbox controllers (Bazzite #602, EndeavourOS forums)"
+echo "  • Missing rumble/vibration support"
+echo "  • Bluetooth connectivity problems"
+echo ""
+read -p "Install controller fix? (y/N): " -n 1 -r
+echo ""
+
+if [[ $REPLY =~ ^[Yy]$ ]]; then
+    echo "Setting up xpadneo enhanced controller support..."
+
+    # Install kernel headers
+    KERNEL=$(uname -r)
+    if [[ $KERNEL == *"zen"* ]]; then
+        sudo pacman -S --noconfirm linux-zen-headers
+    elif [[ $KERNEL == *"lts"* ]]; then
+        sudo pacman -S --noconfirm linux-lts-headers
+    else
+        sudo pacman -S --noconfirm linux-headers
+    fi
+
+    # Install xpadneo driver
+    echo "Installing xpadneo from AUR..."
+    yay -S --noconfirm xpadneo-dkms
+
+    # Replace default driver
+    echo 'blacklist xpad' | sudo tee /etc/modprobe.d/blacklist-xpad.conf >/dev/null
+    sudo modprobe hid_xpadneo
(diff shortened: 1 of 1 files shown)


dhh:
I think we should just install this stuff to ensure that these common controllers just work out of the box. cc @ryanrhughes

bin/omarchy-install-steam
```diff
@@ -1,5 +1,52 @@
...
+echo "This installs xpadneo driver to fix common Xbox/Bluetooth controller issues."
+echo "Learn more: https://github.com/atar-axis/xpadneo"
+echo ""
+echo "Fixes reported by users:"
+echo "  • Steam not detecting Xbox controllers (Bazzite #602, EndeavourOS forums)"
+echo "  • Missing rumble/vibration support"
+echo "  • Bluetooth connectivity problems"
+echo ""
+read -p "Install controller fix? (y/N): " -n 1 -r
+echo ""
```
dhh: Cut all this. It should just add this directly.

bin/omarchy-install-steam
```diff
@@ -1,5 +1,52 @@
...
+
+if [[ $REPLY =~ ^[Yy]$ ]]; then
+    echo "Setting up xpadneo enhanced controller support..."
```
dhh: No output and no ask needed.

bin/omarchy-install-steam
```diff
@@ -1,5 +1,52 @@
...
+
+    # Install kernel headers
+    KERNEL=$(uname -r)
+    if [[ $KERNEL == *"zen"* ]]; then
+        sudo pacman -S --noconfirm linux-zen-headers
+    elif [[ $KERNEL == *"lts"* ]]; then
+        sudo pacman -S --noconfirm linux-lts-headers
+    else
+        sudo pacman -S --noconfirm linux-headers
```
dhh: We only support the default kernel by default, so just stick to that.

bin/omarchy-install-steam
```diff
@@ -1,5 +1,52 @@
...
+    echo "Installing xpadneo from AUR..."
+    yay -S --noconfirm xpadneo-dkms
+
+    # Replace default driver
+    echo 'blacklist xpad' | sudo tee /etc/modprobe.d/blacklist-xpad.conf >/dev/null
+    sudo modprobe hid_xpadneo
+    echo 'hid_xpadneo' | sudo tee /etc/modules-load.d/xpadneo.conf >/dev/null
+
+    # Add input group permission
+    sudo usermod -a -G input $USER
```
dhh: Won't this require a reboot to take effect?

ericjim: Yes, this will require a reboot to take effect. Looks like I can add

```
omarchy-state set relaunch-required
```

bin/omarchy-install-steam
```diff
@@ -1,5 +1,25 @@
...
+sudo modprobe hid_xpadneo
+
+if [[ ! -f /etc/modules-load.d/xpadneo.conf ]]; then
+  echo 'hid_xpadneo' | sudo tee /etc/modules-load.d/xpadneo.conf >/dev/null
+fi
+
+sudo usermod -a -G input $USER
+omarchy-state set relaunch-required
+
+setsid gtk-launch steam >/dev/null 2>&1 &
```
ericjim: Given that a restart is required, we can probably get rid of this line opening Steam

bin/omarchy-install-steam
ericjim: Done. I think this is ready for re-review @dhh

ericjim:
I tested this on a clean install of Omarchy.  After a restart, I was able to connect my Xbox One Elite Series controller, get vibration and buttons working.

I noticed an unrelated issue on my clean install, Nvidia drivers were automatically installed, instead of AMD drivers.  I am on a framework 16 with the dGPU.  I'd be happy to file an issue for this.

(EDIT: adjacent issue https://github.com/basecamp/omarchy/issues/1441 - re: AMD drivers - I can pick this after on a different branch)

vdtdg:
Just chiming in to confirm this worked for me as well. Dual controllers also behaved correctly. Hope that helps put some weight for this change.

dhh:
Moved this over to a dedicated installer.

PR #2010, opened 2025-09-27
Update reboot reminder message for clarity

ShrootBuck:
I found this a bit confusing when I was installing Omarchy, so I think a minor clarification would be helpful, especially to those who might be switching to linux for the first time


dhh:
Hmm, don't see how this makes things clearer.

PR #2013, opened 2025-09-27
feat(starship): add preset management system

pixincreate:
## description

switch between different starship prompt themes with a simple command

## features

- **easy switching**: `spreset <preset_name>` - changes apply instantly
- **zsh transient prompt**: minimal prompt after command execution (automatic)
- **persistent presets**: your selection is remembered across sessions

## usage

```bash
# switch presets
spreset p10k      # powerlevel10k-inspired theme
spreset starship  # default minimal theme

# list available presets
spreset
```

## available presets

- **`starship`** - default minimal configuration
- **`p10k`** - powerlevel10k-inspired with rich information display

## shell support

- **bash**: full preset switching support
- **zsh**: preset switching + transient prompt (shows minimal prompt after commands)
- **others**: basic fallback support

## migrations

```bash
migrations/1758994560.sh
```

### zsh

image

> [!NOTE]
> transient prompt is enabled by default
> to disable transient prompt, set `STARSHIP_TRANSIENT_PROMPT=false` env before starship is initiated (preferably in `~/.profile` or `~/.(bash|zsh)rc`

### bash

image

> [!NOTE]
> transient prompt is not supported since ble(.)sh is required which is not natively supported by bash


phucisstupid:
nice, please add preset catppuccin too: https://github.com/catppuccin/starship/blob/main/starship.toml

pixincreate:
<strike>
few things that running in my mind after i looked into catppuccin theme.

catppuccin relies on `palette` value to change the themes. guess i can add a validation to `spreset` that uses sed to change the `palette` on the go. 
</strike>

added the support!

dhh:
Appreciate the suggestion, but I think we're better off just letting someone explore all that starship has to offer by editing their own config with all the options available. The default should just be simple and easily extensible. So going to skip this for now.

PR #2014, opened 2025-09-27
feat(update): Add firmware update script to update process

eifr:
Introduce a script to handle firmware updates as part of the system update process.

### DIFF
--- a/bin/omarchy-update-firmware
+++ b/bin/omarchy-update-firmware
@@ -3,6 +3,9 @@
 set -e
 echo -e "\e[32mUpdate Firmare\e[0m"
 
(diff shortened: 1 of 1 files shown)


raphaelmso:
I think it would be better if this was a standalone option at walker rather than a entry in omarchy-update-perform.

dhh:
I like having this built in, but agree that it shouldn't be part of the main omarchy-update flow. But we can add it as Update > Firmware.

eifr:
The firmware menu item was added to the system and removed from the Omarchy update, but I believe an "Update All" button for easier maintenance should still be implemented.

PR #2015, opened 2025-09-27
Theme support for Cursor

ludagoo:
Just copied `omarchy-theme-set-vscode`. Reading the `vscode.json` file in the themes because VSCode themes will work in Cursor and then omarchy-themes do not need to explicitly add `cursor.json`.

### DIFF
--- a/bin/omarchy-theme-set
+++ b/bin/omarchy-theme-set
@@ -30,9 +30,10 @@ hyprctl reload
 pkill -SIGUSR2 btop
 makoctl reload
 
-# Change gnome, browser, vscode themes
+# Change gnome, browser, vscode, cursor themes
 omarchy-theme-set-terminal
 omarchy-theme-set-gnome
 omarchy-theme-set-eza
 omarchy-theme-set-browser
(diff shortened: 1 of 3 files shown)


dhh:
Appreciate the suggestion, but having played around with a lot of cursors, I don't actually think it's something that adds enough value to the experience that we need to push it. Of course, anyone can always tailor their own system as they see fit!

ludagoo:
@dhh Cursor the AI editor that is based on VSCode.

dhh:
Lol. I clearly scanned this too quickly 😂

dhh:
I'd like to see less duplication here. Let's make the omarchy-theme-set-vscode the authoritative version, but make it take the params we need to change for cursor, and then we can have omarchy-theme-set-cursor call the other one.

OmarSkalli:
Looks like #1904 (didn't review it, just a quick glance) also adds Cursor support through a refactor to avoid code duplication.

ludagoo:
@dhh is this what you want?

PR #2020, opened 2025-09-27
Waybar Themes support colors

thooams:
I had colors for every themes in the waybar.css

### DIFF
--- a/themes/catppuccin/waybar.css
+++ b/themes/catppuccin/waybar.css
@@ -1,2 +1,11 @@
 @define-color foreground #cdd6f4;
 @define-color background #181824;
+
+@define-color black #4C4F69;
+@define-color red #D20F39;
+@define-color green #40A02B;
+@define-color yellow #DF8E1D;
+@define-color blue #1E66F5;
+@define-color magenta #EA76CB;
+@define-color cyan #179299;
+@define-color white #ACB0BE;
--- a/themes/everforest/waybar.css
+++ b/themes/everforest/waybar.css
@@ -1,2 +1,11 @@
 @define-color foreground #d3c6aa;
 @define-color background #2d353b;
+
+@define-color black #343f44;
+@define-color red #e67e80;
+@define-color green #a7c080;
+@define-color yellow #dbbc7f;
+@define-color blue #7fbbb3;
+@define-color magenta #d699b6;
+@define-color cyan #83c092;
+@define-color white #859289;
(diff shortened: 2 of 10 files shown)


hcabel:
Yes please, more colors for waybar! :pray: 
That being said, I'm not sure your variable naming is meaningful enough. Your associating color value to color name, which isn't very intuitive for all theme. A better approach IMO would be use the purpose of the color instead (like foreground and background)

Examples:
- `accent`, would be for the theme main colour, or a color to highlight some element
- `warning` colour for middly important icon (battery running low) 
- `critical` for errors / critical information (out of battery very soon | notification) 

Something like: `light`, `lighter`, `dark`, `darker`, ... could also work. I reckon we should keep thing simple and not try to add all the colors, with foreground, background and the 3 I've mentioned above, there is already plenty of options for customization.

thooams:
Yep, I use status color in my waybar: https://github.com/thooams/omarchy-dotfiles/blob/main/waybar/theme.css

dhh:
Can you show with screenshots what this is changing?

thooams:
Ok, I moved colors to semantic colors (like @hcabel proposed) in a theme.css file. @dhh, nothing changed for the css files, I just propose semantic colors for every stylesheet files to allow more variables in our custom css. (I have no screenshots to show because there’s no changes)

hcabel:
@dhh The point of this PR is to provide extra theme colours that we can choose from when customizing waybar.

Currently we only have access to `@foreground` and `@background` which is quite limiting

dhh:
Hmm, every additional required setup is a burden for those making new themes. I'd like to see exactly how this is being put to good use before placing more such burdens on theme providers. Also need to see a clear plan for how this would work with existing extra themes that have not defined these variables.

dhh:
On further consideration, I think this is too much. You can already do whatever you want in a waybar.css to overwrite anything. Can't ask theme makers to have to consider even more custom colors.

hcabel:
Sorry for late respond (was pretty busy) here is what I've done for my waybar:

Tokyo-night:
image
Everforest:
image

This changed automatically, based on the accent color variable in the omarchy's waybar.css 
image
image

Without that there is no way (that I know of) to access other color from the theme.
Yes, we can use any static color, but you end up with something like this when switching theme:
image

dhh:
Appreciate this. I think that's too much for the default setup. If someone wants to go that far with the tweaks, they're always able to just overwrite their waybar style and config directly 👍

PR #2026, opened 2025-09-28
Create omarchy-doctor

ryanrhughes:
This is the 4th or 5th swing at something like this but the first one I think is actually viable.

All of the other attempts resulted in something that just felt difficult to maintain over time with a giant set of checks that are totally disconnected from the actual code they're checking. 

In this setup, we turn our install scripts into something more resembling modules with some different functionality. This helps keep the verification code close to the code it's verifying which should make it maintainable without becoming a chore.

- `$OMARCHY_DESCRIPTION` defines a nice name we can call it during install and in the `omarchy-doctor` run. 
- `should_run()` defines the run conditions for the script and is optional but useful for things like hardware fixes
- `omarchy_install()` defines any install commands (executed by `install.sh`)
- `omarchy_verify()` defines any verification commands (executed by `omarchy-doctor`)

This could even allow going a step further to convert things like the present day `omarchy-install-tailscale` to be an install script that isn't executed during the main install but has `omarchy_install` and `omarchy_verify` available such that you could execute the verification for it if the user had Tailscale installed.

**NOTE: This is pulled back into #1860 because there were some changes there to the whole login bit that I didn't want to redo.**

The output is simple and could be made more sophisticated over time but does the job. 

Interestingly enough, the creation of this script actually caught a couple of things on my new system that were already out of spec and I had no idea.

image


bin/omarchy-doctor
```diff
@@ -0,0 +1,161 @@
...
+# Request sudo upfront to avoid prompts during checks
+sudo -v
+
+# Keep sudo alive during the script
+(while true; do sudo -v; sleep 50; done) &
+SUDO_REFRESH_PID=$!
+trap "kill $SUDO_REFRESH_PID 2>/dev/null" EXIT
+
+# Define Omarchy locations
+export OMARCHY_PATH="/home/ryan/Work/omarchy/omarchy-installer"
```
pkordel, line 12: Maybe use $HOME here instead?

dhh:

PR #2032, opened 2025-09-28
feat(catppuccin): use mauve color for hyprland border

phucisstupid:
this pr use catppuccin mauve as borders color, this is default for most catppuccin mocha style and more comfy to look at


dhh:
Appreciate the suggestion, but I find that to be a bit too heavy. Especially around the chrome browser. I think the softer color works better.

PR #2034, opened 2025-09-28
Add Omarchy Theme Installer script

vyrx-dev:
## Auto-fetches themes from [omarchythemes.com](https://omarchythemes.com/) repo.
- Pulls live theme list no manual copy pasting links
- Always up-to-date with new community themes
- Interactive search

https://github.com/user-attachments/assets/51ccd70f-9e5e-40dc-8134-67024215f600

### DIFF
--- a/bin/omarchy-theme-switcher.sh
+++ b/bin/omarchy-theme-switcher.sh
@@ -1,68 +1,65 @@
 #!/usr/bin/env bash
 
-set -e
+echo "╔════════════════════════════════════════╗"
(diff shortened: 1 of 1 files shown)


dhh:
I appreciate wanting to do this, but I actually prefer using the explicitly copied theme URL over, since it makes it very clear that it's not a theme that's maintained by the Omarchy theme. This is especially important when we're dealing with various forms of upgrades where not all themes might be ready at the same time. So I'm going to skip this for default, but I think it could be a great extra AUR app that you distribute and folks can install 👌

PR #2048, opened 2025-09-28
Omarchy 3.0.2

ryanrhughes:


bin/omarchy-restart-pipewire
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+echo -e "Restarting pirewire audio service...\n"
```
shawnanuptra:
```suggestion
echo -e "Restarting pipewire audio service...\n"
```

Just a small typo :)

ryanrhughes: Thanks for catching that

PR #2052, opened 2025-09-28
Add support for older printers via gutenprint

jmargeta:
In the spirit of extending life of perfectly functional, just a bit older, hardware, gutenprint (previously gimp-print) allows CUPS to work with many older printers: https://gimp-print.sourceforge.io/p_Supported_Printers.php

After installing gutenprint, I was able to successfully find my printer (HP LaserJet 1320) and print a few pages:
image

Having out of the box printing experience using older printers and Omarchy would be very pleasant.
However, is gutenprint something that merits joining the base packages, or better to have it as an optional install via walker, or just as it is now - via normal Arch package installation?

### DIFF
--- a/install/omarchy-base.packages
+++ b/install/omarchy-base.packages
@@ -36,6 +36,7 @@ gnome-calculator
 gnome-keyring
 gnome-themes-extra
 gum
+gutenprint
 gvfs-mtp
 gvfs-smb
 hypridle


dhh:
I'm thinking that we should probably move all the printer installation stuff into `Install > Service > Printer`. There are tons of folks who never use a printer. Don't think we need all these packages and the setup we currently have for them. Please do open a PR that does that, if you'd like. I'll close this for now.

PR #2066, opened 2025-09-29
Zoxide: Use `--cmd cd` instead of custom script to replace `cd`

rubas:
Reference: https://github.com/ajeetdsouza/zoxide?tab=readme-ov-file#configuration

---

### Flags

When calling `zoxide init`, the following flags are available:

- `--cmd`
  - Changes the prefix of the `z` and `zi` commands.
  - `--cmd j` would change the commands to (`j`, `ji`).
  - `--cmd cd` would replace the `cd` command.


dhh:
I don't think these are equivalent. The version we currently has gives you a full-path indication when zoxide is doing a jump:

image

I've come to really like that. Since our starship prompt is so minimal. Unless we can recreate this using zoxide alone, I don't want to take a step backwards.

rubas:
Duplicate: https://github.com/basecamp/omarchy/pull/1360 - Next time I ll do a better search.

PR #2072, opened 2025-09-29
feat: add WirePlumber config for automatic Bluetooth audio switching

flaviolivolsi:
**The problem:**

On a fresh Omarchy install, when a Bluetooth audio device connects, it does not automatically become the active audio sink. The user must manually open a tool like `wiremix` and select the device to route audio to it.

This behavior seems to be caused by WirePlumber storing user preferences (`~/.local/state/wireplumber/default-nodes`), which can override the default device priorities on reconnection.

**The solution:**

This PR makes the auto-switch behavior explicit and reliable by adding a WirePlumber configuration that gives newly connected Bluetooth devices a higher priority (2000) than the default for internal audio (1000).

**The result:**

With this change, any newly connected Bluetooth headset will immediately and automatically become the default input/output device. This provides a seamless "it just works" experience, consistent with the Omarchy philosophy, regardless of stored preferences or specific hardware.

---

(description edited for clarity - thx @iamobservable and @cluah)

### DIFF
--- a/migrations/1759159162.sh
+++ b/migrations/1759159162.sh
@@ -3,17 +3,9 @@ echo "Enable automatic Bluetooth audio device switching"
 # Ensure WirePlumber is installed
 omarchy-pkg-add wireplumber
 
-# Create config directory if it doesn't exist
+# Copy over bluetooth switch config
 mkdir -p ~/.config/wireplumber/wireplumber.conf.d/
+cp $OMARCHY_PATH/config/wireplumber/wireplumber.conf.d/51-bluetooth-autoswitch.conf ~/.config/wireplumber/wireplumber.conf.d/
 
-# Only install if not already present (don't overwrite user modifications)
-if [[ ! -f ~/.config/wireplumber/wireplumber.conf.d/51-bluetooth-autoswitch.conf ]]; then
-  cp ~/.local/share/omarchy/config/wireplumber/wireplumber.conf.d/51-bluetooth-autoswitch.conf \
-     ~/.config/wireplumber/wireplumber.conf.d/
-
-  echo "  → Bluetooth auto-switch config installed"
-  echo "  → Restarting WirePlumber..."
-  systemctl --user restart wireplumber
-else
-  echo "  → Config already exists, skipping"
-fi
+# Restart to pickup new config
+systemctl --user restart wireplumber
--- a/migrations/1759159162.sh
+++ b/migrations/1759159162.sh
@@ -1,0 +1,19 @@
+echo "Enable automatic Bluetooth audio device switching"
+
+# Ensure WirePlumber is installed
+omarchy-pkg-add wireplumber
+
+# Create config directory if it doesn't exist
+mkdir -p ~/.config/wireplumber/wireplumber.conf.d/
+
+# Only install if not already present (don't overwrite user modifications)
+if [[ ! -f ~/.config/wireplumber/wireplumber.conf.d/51-bluetooth-autoswitch.conf ]]; then
+  cp ~/.local/share/omarchy/config/wireplumber/wireplumber.conf.d/51-bluetooth-autoswitch.conf \
+     ~/.config/wireplumber/wireplumber.conf.d/
+
+  echo "  → Bluetooth auto-switch config installed"
+  echo "  → Restarting WirePlumber..."
+  systemctl --user restart wireplumber
+else
+  echo "  → Config already exists, skipping"
+fi
--- a/config/wireplumber/wireplumber.conf.d/51-bluetooth-autoswitch.conf
+++ b/config/wireplumber/wireplumber.conf.d/51-bluetooth-autoswitch.conf
@@ -1,0 +1,36 @@
+# Automatically switch to Bluetooth audio devices when connected
+# This configuration ensures Bluetooth audio devices become the default
+# audio sink/source when they connect
+
+monitor.bluez.rules = [
+  {
+    # Bluetooth output devices (headphones/speakers)
+    matches = [
+      {
+        device.api = "bluez5"
+        media.class = "Audio/Sink"
+      }
+    ]
+    actions = {
+      update-props = {
+        priority.session = 2000   # Higher priority than internal audio (1000)
+        priority.driver  = 2000
+      }
+    }
+  }
+  {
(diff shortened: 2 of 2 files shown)


dhh:
Nice! We're going to need a migration to copy this over on existing installs.

flaviolivolsi:
> Nice! We're going to need a migration to copy this over on existing installs.

Whoops, right - thanks @dhh! I created a migration, I hope it's ok the way I did it

I'll ask a question since we're at it: are there any guidelines for contributing? E.g.: using the dev branch as base, ways to test migrations or changes or anything I might have missed (tried to search but didn't find them}

dhh:
We should make a contributors guide for sure. The branch thing is also something we should get cleaned up, so it's just the default is correct. There's a bit of backlog there!

migrations/1759159162.sh
```diff
@@ -0,0 +1,31 @@
+echo "Enable automatic Bluetooth audio device switching"
+
+# Only apply if WirePlumber 0.5+ is installed
+if command -v wireplumber &> /dev/null; then
```
dhh: I think we can just do `omarchy-pkg-add wireplumber` to ensure it's installed. Then we don't have to do the version test either.

flaviolivolsi:
> We should make a contributors guide for sure. The branch thing is also something we should get cleaned up, so it's just the default is correct. There's a bit of backlog there!

Cool! Yeah just wanted to make sure I got things right as it's my first PR and I'd like to send more. Since I didn't find an issue for the contributors guide I went ahead and [created one](https://github.com/basecamp/omarchy/issues/2077). I also amended the latest commit and applied the fix you mentioned about `omarchy-pkg-add` - thanks again!

cluah:
How does it compare to current behavior? I mean, my BT headphones already connect when I turn them on.

iamobservable:
Same question as @cluah.

My wiremix setup (default) automatically switches to my BT device (Airpod Pro 2) once connected.

What device(s), version, etc are you using @flaviolivolsi ?

flaviolivolsi:
@cluah @iamobservable thanks for pointing this out! In my case the headphones do connect automatically but they don't become the active audio sink - at least on my fresh install, I still have to open wiremix and hit 'd' on the device to set it manually. Maybe this already works seamlessly on some hardware/setups?

My PR just aims to automate that step making the behavior consistent with what worked for me. Here's some info about my setup:

- Laptop: Lenovo ThinkPad P14s AMD Gen 5
- Headset: Skullcandy Crusher ANC 2
- Kernel: 6.16.8-arch3-1
- PipeWire 1.4.8-2
- WirePlumber 0.5.11-1
- BlueZ 5.84-1

The key difference might be that on fresh installs, WirePlumber has no stored preferences in `~/.local/state/wireplumber/default-nodes`. Once you manually switch audio outputs via wiremix or other means, WirePlumber stores this as a user preference that overrides device priorities on reconnection (it might be helpful to know if you recall having done so).

What I'd like to achieve is consistent behavior by giving Bluetooth devices higher priority (2000 vs 1000 for internal audio), making the auto-switch work regardless of stored preferences - which matches what you're already experiencing.

iamobservable:
@flaviolivolsi 

Thanks for sharing more information about the expectations. In the initial message, I did not see a mention of not having previously selected the device. My experience IS HAVING already manually selected the BT device and it later automatically switches on connection sync. Sounds like this is not the goal you had, but to allow for switching of any device immediately after Bluetooth connect. Thanks for clarifying.

Now if I can just figure out how to use pipewire and airplay to connect to my devices on my home network! But that is another cliff to climb. :rofl:

flaviolivolsi:
> @flaviolivolsi
> 
> Thanks for sharing more information about the expectations. In the initial message, I did not see a mention of not having previously selected the device. My experience IS HAVING already manually selected the BT device and it later automatically switches on connection sync. Sounds like this is not the goal you had, but to allow for switching of any device immediately after Bluetooth connect. Thanks for clarifying.
> 
> Now if I can just figure out how to use pipewire and airplay to connect to my devices on my home network! But that is another cliff to climb. 🤣

You're right, my initial description was unclear and this discussion also helped me frame it properly, thank you for that! Updating the PR description right away to be more specific about the problem and solution.

And best of luck with your new quest! :D

cluah:
@flaviolivolsi i got it. do you know how i can get the current default behavior without reinstalling omarchy?

flaviolivolsi:
> @flaviolivolsi i got it. do you know how i can get the current default behavior without reinstalling omarchy?

Copying over `51-bluetooth-autoswitch.conf` from this PR inside your `~/.config/wireplumber/wireplumber.conf.d` should be enough! You likely won't have those dirs tho, so you'll have to create them with `mkdir -p ~/.config/wireplumber/wireplumber.conf.d`

cluah:
Sorry I mean I want this:

> On a fresh Omarchy install, when a Bluetooth audio device connects, it does not automatically become the active audio sink. The user must manually open a tool like wiremix and select the device to route audio to it.

I tried deleting “~/.local/state/wireplumber” and “/var/lib/bluetooth/,” but I'm not sure if I'm doing it.

cluah:
@flaviolivolsi I tried again using a fresh installation. 

Sinks: laptop's built-in speakers + two BT headphones.
Sources: I didn't check the sources.

My Sony WH-1000 headphones automatically acquire audio after connecting them using the UI (Bluetooth icon on the top right). And it's the one listed in `pactl get-default-sink`. I also tried other older Sony WH headphones with the same result. Both show priority 1010. When I connect the second one, the first one doesn't lose its place, and when I disconnect the first one, the second one takes its place.

Then I tried to test the PR, but I couldn't get it to work. It kept showing 1010. I had to change this:

```
matches = [
  {
    device.api = "bluez5"
    media.class = "Audio/Sink"
  }
]
```
To this: 
```
matches = [
  {
    node.name = "~bluez_output.*"
  }
]
```
I was able to see the higher priority values, but I couldn't see a change in behavior.

So at least in this case, where the priority of headphones is already higher than that of built-in speakers, increasing it seems to keep things the same.

flaviolivolsi:
@cluah thanks for testing! I'll try again with a new install and let you know if I incur in the same issue

dhh:

PR #2073, opened 2025-09-29
Set background image for current theme by index or filename

AlexZeitler:
This PR adds the ability to be specific about which background image for the current theme should be set:

You can set the background for the current theme by index:

```bash
omarchy-theme-bg-set-index 5
```

You can set the background for the current theme by filename:

```bash
omarchy-theme-bg-set-filename mybg.jpg
```

This came in handy for me when I was scripting my post Omarchy install setup

### DIFF
--- a/bin/omarchy-theme-bg-set-filename
+++ b/bin/omarchy-theme-bg-set-filename
@@ -1,0 +1,28 @@
+#!/bin/bash
+
+# set background image by filename for current theme
+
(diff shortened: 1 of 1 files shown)


dhh:
Appreciate the suggestion, but I think that's a pretty niche consideration, and would require substantial complication of the scripts. Think you're better off just having this in your tailoring script 👌

PR #2084, opened 2025-09-29
Improve UX for webapp removal selection

devzeth:
This PR improves the user experience when using the webapp removal tool:

- **Clearer instructions**: Added specific guidance in the selection header (Space to select, Enter to submit)
- **Better error messaging**: Improved the error message to explicitly explain how to select web apps for removal

These changes make the tool more intuitive for users who might not be familiar with the gum selection interface.


dhh:
In the latest Gum, we now have a description below to reveal toggle and submit. I don't actually think this adds enough on top of that to be worth the repetition:

image

PR #2098, opened 2025-09-30
Increase Application Not Responding delay

CaffeinatedTech:
Issue: When an application hangs temporarily, the Application Not Responding dialog appears, asking if you want to wait or terminate the application.  The default is to show the dialog when just one ping is missed.  This causes the dialog to appear while games load levels, or sometimes a window will dim momentarily while the dialog appears and disappears almost instantaneously.  This harms user experience and causes unnecessary google searches to solve it.

Fix: Include the `anr_missed_pings` setting and set to 10 instead of the default of 1. Also provide an easy way to disable the dialog all together `enable_anr_dialog = true` right in the config files.

I feel that this is a more useful default while also exposing the setting more conveniently to the user.

### DIFF
--- a/config/hypr/looknfeel.conf
+++ b/config/hypr/looknfeel.conf
@@ -21,3 +21,12 @@ dwindle {
     # Avoid overly wide single-window layouts on wide screens
     # single_window_aspect_ratio = 1 1
 }
+
+# https://wiki.hypr.land/Configuring/Variables/#misc
+misc {
+    # Enable or disable the ANR dialog
+    enable_anr_dialog = true
+    # Delay before showing Not Responding dialog
+    anr_missed_pings = 10
+  }
+


AnujCtrl:
much needed.

config/hypr/looknfeel.conf
```diff
@@ -21,3 +21,12 @@ dwindle {
...
     # single_window_aspect_ratio = 1 1
 }
+
+# https://wiki.hypr.land/Configuring/Variables/#misc
+misc {
+    # Enable or disable the ANR dialog
+    enable_anr_dialog = true
+    # Delay before showing Not Responding dialog
+    anr_missed_pings = 10
+  }
```
landsman, line 31:
```suggestion
}
```

woopstar:
I think this needs to be included in the defaults/hypr/looknfeel.conf file also

vlofgren:
Might be worth seeing if vaxry will accept a PR for an updated default value for this in the main Hyprland repo.  The default value is kinda insane everywhere and I don't know who would want it.

dhh:
Yes, I'd love to see this increased as a default in Hyprland. Have also see a million false positives on this. What do you think, @vaxerski?

dhh:
I've added this to the defaults for now and given you credit 👌

vaxerski:
iirc the default is 5s, if the app is hung for >5s it shows up... with your 5 it's 25s, quite slow.

CaffeinatedTech:
I wrote a little C GTK program to test the ANR - well I had AI write it.  With the `anr_missed_pings = 1` which is indicated as the hyprland default in the [docs](https://wiki.hypr.land/Configuring/Variables/#misc), it takes 2 seconds to show an ANR dialog.  2 seconds in my opinion is a bit too tight, and there seem to be plenty of people looking to disable ANR because they see the message too frequently.  It is pretty annoying to see it while on the load screen of a game - try Portal for example.

Anyhow, DHH's setting of 5 would make the new timeout 10 seconds, which might be a bit long, but not so long that a user might start scratching their head wondering what is going on.

I also propose adjusting the default `anr_missed_pings` value upstream, I'll see if there are already any open issues for it.

I've opened a discussion over at the hyprland repo https://github.com/hyprwm/Hyprland/discussions/11884

dhh:
Lowered our default to 3 pings. But yeah, when it's set to 1, it's definitely trickering faster than 5 seconds!

CaffeinatedTech:
Vaxerski increased the default to 5 pings upstream.  The ping rate is actually 1.5 seconds.

https://github.com/hyprwm/Hyprland/commit/378438ffe7306d20f0cd799c8c367cf4c9356c78

dhh:
We should remove this Omarchy setting when the next hyprland ships, then 👍

PR #2109, opened 2025-09-30
Fixes #2036: ghostty needs class names with a dot

### LINKED ISSUE #2036: When changing default terminal, waybar tui's (wifi, audio) is not floating
### System details

Thinkpad t480

### What's wrong?

When i changed my default terminal to ghostty, wifi tui and audio tui is not floating but when i change back to alacritty it is floating

Woosah:
Ghostty recently rewrote the complete [GTK application api](https://mitchellh.com/writing/ghostty-gtk-rewrite) in zig, and now it strictly adheres to the [GTK standard](https://docs.gtk.org/gio/type_func.Application.id_is_valid.html), where alacritty does this a little less. Therefore, when Impala or Wiremix are started with only Impala/Wiremix as a class parameter, ghostty ignores these as invalid. GTK classes must be strings with at least one period in it, and they are not allowed to start with one. Without a class set, the Omarchy windowrule does not match anymore and these windows are treated as tiles, not floats. The hard-coded font-size is used to match the alacritty-based Omarchy style but can be omitted to make this dynamic.

### DIFF
--- a/bin/omarchy-launch-wifi
+++ b/bin/omarchy-launch-wifi
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-exec setsid uwsm app -- "$TERMINAL" --class=Impala -e impala "$@"
+exec setsid uwsm app -- "$TERMINAL" --class=Impala.float --font-size=9 -e impala "$@"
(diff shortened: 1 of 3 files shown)


abenz1267:
https://github.com/basecamp/omarchy/pull/1778

dhh:
Closed as duplicate.

PR #2121, opened 2025-10-01
any browser in webapp bindings

crestanigui:
Firefox (and other browsers that doesn't support webapp) users have a bad exp when changing browser in ./config/hypr/bindings because it doesn't change the omarchy-launch-webapp flow. 

When uninstalling omarchy-chromium even worse. The bindings start returning errors.

This PR checks the default browser and if it's not on the webapp supported list, it opens the link from the parameter normally in a new tab. 
**The user just change the default browser and the bindings are now in the browser of choice!**

### DIFF
--- a/bin/omarchy-launch-webapp
+++ b/bin/omarchy-launch-webapp
@@ -2,13 +2,16 @@
 
 browser=$(xdg-settings get default-web-browser)
 
-case $browser in
-google-chrome* | brave-browser* | microsoft-edge* | opera* | vivaldi* | chromium*)
-  app_cmd=$(sed -n 's/^Exec=\([^ ]*\).*/\1/p' {~/.local,~/.nix-profile,/usr}/share/applications/$browser 2>/dev/null | head -1)
+webapp_browsers="google-chrome brave-browser microsoft-edge opera vivaldi chromium"
+
+resolve_exec() {
+  sed -n 's/^Exec=\([^ ]*\).*/\1/p' {~/.local,~/.nix-profile,/usr}/share/applications/$1 2>/dev/null | head -1
+}
+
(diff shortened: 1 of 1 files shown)


bartdorsey:
I'd kind of like to see firefoxpwa (a third party program and extension) integrated so that it could be used when a user has firefox/waterfox/zen as their browser.

g012:
That's not the behavior I'd like to see. Personnally I prefer keeping chromium for webapps, and opening any link from these webapps into firefox. I don't want to lose the webapp functionality just because I use firefox as default.

crestanigui:
Personally I think we should respect the user’s default browser choice. Many people choose Firefox (or others) specifically to avoid Google, and not everyone wants websites turned into "apps". If something truly deserves its own application feel, it should be packaged separately (like Discord, Slack, etc.), rather than forced into Chromium’s --app mode.

g012:
But nothing is forced - you don't have to use webapps, you can create a shortcut that launches the default browser instead.
However, if webapps are used (using chromium if default browser doesn't support webapps), links in them should open in the default browser.

Of course separate package is better, but if it doesn't exist, then we have to live with it.

crestanigui:
I'm convinced. Using omarchy-launch-webapp is already very specific. 
The solution should be altering the shortcut to open the link in the default browser.

Thanks!

MillerJ20:
> Awesome fix, ran into this with my buddy using Zen browser, fix in PR works great. Thanks!

Silly question: I am having a hard time figuring out where these commands are to edit them and try this flow for myself, since the PR was closed. What directory are they in?

EDIT: I spent two more minutes looking, they are in /home/{USERNAME}/.local/share/omarchy/bin

PR #2136, opened 2025-10-01
added an icon to waybar when the notifications are silenced

victoralensai:
This pull request introduces a do-not-disturb indicator to Waybar. The changes add a new visual indicator for do-not-disturb mode, update notification keybindings to refresh the indicator, and implement the supporting script.

**Waybar do-not-disturb indicator integration:**

* Added a new `custom/do-not-disturb-indicator` module to the center of Waybar, providing a visual cue when notifications are silenced. No indicators are visible when notifications are not silenced. (`config/waybar/config.jsonc`)
* Configured the `custom/do-not-disturb-indicator` module to toggle do-not-disturb mode and display a notification when clicked, using a custom script for status updates. (`config/waybar/config.jsonc`)
* Implemented the `do-not-disturb.sh` script to output indicator status in JSON format based on current notification mode. (`default/waybar/indicators/do-not-disturb.sh`)

**Notification keybinding improvements:**

* Updated the SUPER+CTRL+COMMA keybinding to trigger a Waybar refresh after toggling do-not-disturb mode, ensuring the indicator updates immediately. (`default/hypr/bindings/utilities.conf`)


victoralensai:
**Note**: This is my first PR, so please let me know if I missed any steps or if the procedure could be improved. 
I initially wrote a description, but GitHub suggested using Copilot to refine it. I kept the Copilot version as it was clearer, I hope there is no problem with that.

victoralensai:
This is the default look when notifications are disabled through `SUPER CTRL + COMMA`:
image

dhh:
This is great, but we need a migration that adds the new elements to the waybar config and style. Should use grep/sed, so we don't have to refresh the entire bar and overwrite people's configurations, if we can avoid it.

victoralensai:
Would some sort of update script for waybar that dynamically add the new elements do the trick? I wonder how to refresh the entire bar without overwriting peoples's configuration.

I haven't thought of the fact that adding elements to the defaults would overwrite people's configuration :(.

dhh:
You can check how some of the other migrations do it. Basically, you'll look for the string that's next to what you want to add, then dynamically add it.

victoralensai:
I have added a migration script that dynamically change ~/.config/waybar/style.css and ~/.config/waybar/config.jsonc. 

I have tested it on the files from the master branch and I get the same as my feature branch. If the sed does not find the lines to change, then nothing happen (I don't want peoples config to get overwrite!). Still it makes a backup file just in case.

I did not make a migration for the default bindings, because I assume people don't mess with the file in `defaults/`.

dhh:
Appreciate the work on this, but ended up with a smaller solution: https://github.com/basecamp/omarchy/pull/4525/changes/fcf883e827bc4c08383465ad3bc4a1bf5eb5899d. Thanks for working on it!

PR #2146, opened 2025-10-01
fix: no random key press on wake enter in hyprlock password field 

nicoestrada:
No longer have those random key presses to disable sleep enter the hyprlock password fields on wake. Also enable mouse movement to trigger wake


dhh:
This is nice. We need a migration that adds these new rules using sed, so folks with existing configurations don't have them overwritten, but still get these rules.

nicoestrada:
Done!

migrations/1759778278.sh
```diff
@@ -0,0 +1,18 @@
...
+CONFIG_FILE="$HOME/.config/hypr/hypridle.conf"
+
+# Ensure the configuration file exists
+if [ ! -f "$CONFIG_FILE" ]; then
+    echo "Configuration file not found at $CONFIG_FILE. Creating a new one."
+    touch "$CONFIG_FILE"
+fi
+
+if ! grep -q "key_press_enables_dpms = true" "$CONFIG_FILE"; then
+    sed -i '$a key_press_enables_dpms = true' "$CONFIG_FILE"
```
dhh: This will just append it to the end, so not inside the right listener?

nicoestrada: Fixed, per [docs](https://wiki.hypr.land/Configuring/Variables/), these must be in misc section

nicoestrada:
Since `key_press_enables_dpms` and `mouse_move_enables_dpms` are [misc variables](https://wiki.hypr.land/Configuring/Variables/#misc), I think it might be time to add a misc.conf file for the one place for us to configure misc vars. 

Another solution could just add misc{} to another file but think this might be more intuitive down the line

migrations/1759778278.sh
nicoestrada: These are misc variables and think these should be where the misc hyprland config lives

dhh:
This should actually just be the default, so added it straight to the default configs and gave you credit. Thanks!

PR #812, opened 2025-08-15
Use host's DNS resolver in Docker containers

kevinmcconnell:
Configures Docker to use the host's DNS resolver. With this configuration, Docker containers will be able to resolve DNS in the same way the host can.

We do this by making systemd-resolved listen on our Docker network, and then configure the Docker daemon to use it.

On example of when this setup would be required is if you have infrastructure on a Tailscale network that your buildx containers need to reach (like an internal registry); without it Tailscale's MagicDNS will not be available from inside a build container.

/cc @dhh

PR #752, opened 2025-08-13
Add omarchy-setup-cloudflare-dns script

jardahrazdera:
Addresses issue #745 by providing user choice for DNS configuration.

Features:
- Makes Cloudflare DNS opt-in instead of forced default
- Preserves local development environments with DHCP option
- Includes custom DNS configuration support
- DNS over TLS support for enhanced security
- Simplified implementation (58 lines) per review feedback
- Follows Omarchy naming conventions

Fixes #745

PR #2150, opened 2025-10-01
Use systemd-resolved's stub resolver

samuelpecher:
On first run, symlink systemd-resolved's resolver into /etc/resolv.conf
* use first-run script due to chroot restrictions
* add to password-less sudo first-run commands

Matches migration 1754984623

### DIFF
--- a/bin/omarchy-cmd-first-run
+++ b/bin/omarchy-cmd-first-run
@@ -9,6 +9,7 @@ if [[ -f "$FIRST_RUN_MODE" ]]; then
 
   bash "$OMARCHY_PATH/install/first-run/battery-monitor.sh"
   bash "$OMARCHY_PATH/install/first-run/firewall.sh"
+  bash "$OMARCHY_PATH/install/first-run/dns-resolver.sh"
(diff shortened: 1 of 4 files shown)


dhh:
What is this fixing or adding?

samuelpecher:
New installs weren't picking up proper DNS settings as `/etc/resolv.conf` didn't point to the stub resolver config created by systemd-resolved. This fixes that, in line with Arch's docs [instruction to enable to symlink](https://wiki.archlinux.org/title/Systemd-resolved).

This was enabled in #812, and taken out in preparation for #752, but without the symlink the new settings won't be picked up on a fresh install.

flavorjones:
👍 I needed to make this change manually this morning after an Omarchy upgrade before Tailscale DNS would work for me.

dhh:
Great. We probably need a migration to fix this for existing installations too, yeah?

dhh:
Will merge this for now. Please do a follow-up PR with the migration we need.

samuelpecher:
Migration PR is #2313

PR #2154, opened 2025-10-02
Add Midnight Commander theme support and default TUI

rafaelvzago:
## Summary
  - bundle Midnight Commander theme variants plus helper script for applying them
  - ship default Midnight Commander TUI assets and packaging hooks so it installs with omarchy
  - add migrations to roll out the new theme assets and ensure the TUI is registered


abenz1267:
omarchy already ships with a file browser. i don't think shipping another one makes much sense. 

The next PR will contain yazi... next one ranger... and so on.

rafaelvzago:
@abenz1267 I wasn't aware that there was already a text file browser, other then use nvim for this propose. I'm using mc for a long time and I thought that would be good to have a TUI file browser. Thanks for the feedback.

woopstar:
if this is to be merged, then the migrations should also install mc. Adding the package to the base file does not mean users who already have Omarchy installed will have it installed.

dhh:
Appreciate the suggestion and the work here, but I'm really happy with nautilus as my file manager, so I wouldn't use this, and therefore it can't be a default. But maybe you can add a omarchy-mc package that does all this work and sets things up with the matching themes 👌

PR #2171, opened 2025-10-02
Add TZ environment variable for timezone-aware applications

### LINKED ISSUE #2170: Waybar clock :L locale operator not respecting system timezone without TZ environment variable
## System details
Omarchy v3.0.1 (Linux 6.16.8-arch3-1), manual installation over arch, running the scripts from cloned repo

## Problem Description

Waybar clock `:L` locale operator not respecting system timezone without TZ environment variable

In some manual installations, the waybar clock format `{:L%A %H:%M}` displays UTC time instead of the configured system timezone, even when `timedatectl` shows the correct timezone is set.

## Root Cause

The `:L` locale operator in waybar relies on the `TZ` environment variable to determine the timezone. While `timedatectl` can show and set the system timezone correctly, applications launched through certain desktop environments or window managers may not inherit the `TZ` environment variable, causing them to default to UTC.

## Example

- System timezone: `timedatectl show -p Timezone --value` returns `America/Argentina/Tucuman`
- Waybar config: `"format": "{:L%A %H:%M}"`
- **Without TZ:** Clock shows UTC time
- **With TZ:** Clock shows correct local time

## Current Workaround

Users have to hardcode the timezone in waybar config:
```json
{
    "clock": {
        "timezone": "America/Argentina/Tucuman",
        "format": "{:%A %H:%M}"
    }
}
```

## Proposed Solution

Set the `TZ` environment variable dynamically in omarchy's default bash environment:

```bash
# In default/bash/envs
export TZ="$(timedatectl show -p Timezone --value)"
```

This ensures:
1. **Dynamic detection:** Works for all users worldwide without hardcoding
2. **Proper inheritance:** Applications launched from terminals get the correct TZ
3. **Waybar compatibility:** The `:L` locale operator works as expected
4. **System consistency:** Matches the timezone configured via `timedatectl`

matiasmoya:
This PR adds the TZ environment variable to ensure timezone-aware applications display the correct local time instead of UTC.

## Problem
- Waybar clock format `{:L%A %H:%M}` displays UTC time instead of local time in some systems where TZ env variable is not defined
- Applications launched from shell don't inherit system timezone without TZ variable
- Users have to hardcode timezone in application configs as workaround

More details about the problem here https://github.com/basecamp/omarchy/issues/2170 

## Solution  
- Dynamically set TZ environment variable based on system timezone via `timedatectl`
- Added to `default/bash/envs` so it's available in all bash sessions
- Works globally for all users without hardcoding timezone values

## Testing
- Verified TZ variable is set correctly: `echo $TZ` shows system timezone
- Waybar clock now displays local time with `:L` locale operator
- Applications launched from shell inherit correct timezone

Fixes #2170

### DIFF
--- a/default/bash/envs
+++ b/default/bash/envs
@@ -1,3 +1,4 @@
 # Editor used by CLI
 export SUDO_EDITOR="$EDITOR"
 export BAT_THEME=ansi
+export TZ="$(timedatectl show -p Timezone --value)"


dhh:
I can't replicate this and I don't need how adding a TZ export to bash would fix it since that export wouldn't be available to waybar. I have the correct TZ using :L. You can update the TZ based on your current IP location by right clicking on the clock or via Update > Timezone in the menu.

PR #2176, opened 2025-10-03
add chameleon plugin

tahayvr:
Adds the chameleon plugin to support easy neovim theming using a color palette. 

See [chameleon.nvim](https://github.com/tahayvr/chameleon.nvim) for more info.

### DIFF
--- a/bin/omarchy-theme-set
+++ b/bin/omarchy-theme-set
@@ -29,6 +29,7 @@ omarchy-restart-swayosd
 hyprctl reload
 pkill -SIGUSR2 btop
 makoctl reload
+pkill -SIGUSR1 nvim
 
 # Change gnome, browser, vscode themes
 omarchy-theme-set-terminal
--- a/config/nvim/lua/plugins/chameleon.lua
+++ b/config/nvim/lua/plugins/chameleon.lua
@@ -1,0 +1,16 @@
+return {
+  {
+    "tahayvr/chameleon.nvim",
+    lazy = false,
+    priority = 1000,
+    config = function()
+      require("chameleon").setup({
+        system = {
+          symlink_path = vim.fn.expand("~/.config/omarchy/current/theme"),
+          debounce_ms = 100,
+          notify = false,
+        },
+      })
+    end,
+  },
+}


dhh:
How does this work together with the new theme loader we added? What I'd love to see is that the default themes use neovim.lua, but extra themes, if they don't have neovim.lua, could add neovim.theme, which would use this. And then, finally, if neither are there, we just default to pixel, which grabs colors from terminal.

tahayvr:
Chameleon only responds to themes that have a neovim.lua with a  palette specified like below. It will ignore anything else (silently):

```lua
return {
  -- BACKGROUND LAYERS (dark → lighter).
  bg1 = "#0F0F19",
  bg2 = "#181824",
  bg3 = "#202034",

  -- FOREGROUND LAYERS (bright → muted).
  fg1 = "#EDEDFE",
  fg2 = "#BDBDE6",
  fg3 = "#8C8CB8",

  -- SELECTION
  selbg  = "#2B2B44",
  selfg  = "#EDEDFE",

  -- COMMENTS
  comment = "#7575A3",

  -- ACCENT COLORS (syntax + UI)
  color1 = "#FF3366",
  color2 = "#00F59B",
  color3 = "#FFEA00",
  color4 = "#33A1FF",
  color5 = "#FF66F6",
  color6 = "#3CFFED",

  -- UI / DECORATIVE
  uic1 = "#5A5A88",
}
```

The only issue is that Lazyvim is gonna look at the theme.lua symlink, and expects a theme so throws an error when seeing something like above. Still works, but the error comes and goes.

Maybe there is a way to silence that Lazyvim error/notification?

dhh:
I think we should use a different name for these, so we can reserve neovim.lua for the case where you're referencing a full theme plugin. So chameleon should only kick in if there's no neovim.lua, but if we have a neovim.theme, or something like that.

tahayvr:
Done! chameleon now responds to `neovim.theme` files instead of `neovim.lua`.

dhh:
Do you have some themes that I can use to test this on?

dhh:
Also, @ryanrhughes, does this mean we have to add the chameleon plugin in the omarchy-lazyvim bundle? I've already wanted to have a copy of all the default theme plugins in that bundle so they're preloaded there.

tahayvr:
Sure, here's matte black in this format:
```lua
return {
  -- BACKGROUND LAYERS (dark → lighter).
  bg1 = "#121212",
  bg2 = "#1b1b1b",
  bg3 = "#242424",
  -- FOREGROUND LAYERS (bright → muted).
  fg1 = "#bebebe",
  fg2 = "#a3a3a3",
  fg3 = "#8a8a8d",
  -- SELECTION
  selbg = "#333333",
  selfg = "#bebebe",
  -- COMMENTS
  comment = "#8a8a8d",
  -- ACCENT COLORS (syntax + UI)
  color1 = "#D35F5F",
  color2 = "#FFC107",
  color3 = "#b91c1c",
  color4 = "#e68e0d",
  color5 = "#D35F5F",
  color6 = "#bebebe",
  -- UI / DECORATIVE
  uic1 = "#333333",
}
```

and here's a monotone one for a theme called Void:
```lua
return {
	-- BACKGROUND LAYERS (dark → darker).
	bg1 = "#05010C",
	bg2 = "#0b0712",
	bg3 = "#120e19",
	-- FOREGROUND LAYERS (bright → muted).
	fg1 = "#FFFFFF",
	fg2 = "#DECCFF",
	fg3 = "#B59AE6",
	-- SELECTION
	selbg = "#382952",
	selfg = "#FFFFFF",
	-- COMMENTS
	comment = "#6B578F",
	-- ACCENT COLORS (syntax + UI)
	color1 = "#DECCFF",
	color2 = "#DECCFF",
	color3 = "#DDCCFF",
	color4 = "#B59AE6",
	color5 = "#B49AE6",
	color6 = "#B49AE6",
	-- UI / DECORATIVE
	uic1 = "#382952",
}
```
Unfortunately, Lazyvim is still going to throw an error at least once because the neovim.lua file doesn't exist.
2025-10-07-232314_hyprshot

ryanrhughes:
> Also, @ryanrhughes, does this mean we have to add the chameleon plugin in the omarchy-lazyvim bundle? I've already wanted to have a copy of all the default theme plugins in that bundle so they're preloaded there.

@dhh Yep. We would need that. In thinking about this more; we need to consider how we want to maintain this so that we pick up additions like this.

We either have to clone Omarchy in and pull those config files over during the build process or just move our Nvim configs over to that package instead. I think I prefer the latter since it's a little cleaner and just creates a single, self-contained module.

I'll mess with this tomorrow and see if there are any unforeseen consequences.

PR #2186, opened 2025-10-03
fix: expected .desktop filename for helium

d0ctr:
According to common sense and Helium browser's spec for .desktop file: [helium.desktop](https://github.com/imputnet/helium-linux/blob/main/package/helium.desktop)

The expected name here should be `helium*` not `helium-browser*`.

### DIFF
--- a/bin/omarchy-launch-webapp
+++ b/bin/omarchy-launch-webapp
@@ -3,7 +3,7 @@
 browser=$(xdg-settings get default-web-browser)
 
 case $browser in
-google-chrome* | brave-browser* | microsoft-edge* | opera* | vivaldi* | helium-browser*) ;;
+google-chrome* | brave-browser* | microsoft-edge* | opera* | vivaldi* | helium*) ;;
 *) browser="chromium.desktop" ;;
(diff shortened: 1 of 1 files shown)


dhh:
I just installed the helium browser using the AUR package and it's indeed `/usr/share/applications/helium-browser.desktop`.

d0ctr:
> I just installed the helium browser using the AUR package and it's indeed `/usr/share/applications/helium-browser.desktop`.

There are 2 AUR packages for this, none of them are maintained by the developers and they produce different `.desktop` files:

- [helium-browser-appimage](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=helium-browser-appimage)
- [helium-browser-bin](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=helium-browser-bin)

dhh:
The one I was using was [helium-browser-bin](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=helium-browser-bin). Which I would say is the default that folks would pick.

PR #2191, opened 2025-10-03
Set default terminal after removing one

hyprhex:
The default terminal on **Omarchy** is _Alacritty_, but if you install a new one like _Ghostty_, it will switch automatically as the default one. But when you remove _Ghostty_, you're stuck with no terminal, so that's why I added this short code to check removed pkg If it's terminl it check if for installed one and set it as a default

### DIFF
--- a/bin/omarchy-pkg-remove
+++ b/bin/omarchy-pkg-remove
@@ -13,9 +13,22 @@ fzf_args=(
 )
 
 pkg_names=$(yay -Qqe | fzf "${fzf_args[@]}")
+TERMINALS=("alacritty" "kitty" "ghostty")
 
 if [[ -n "$pkg_names" ]]; then
   # Convert newline-separated selections to space-separated for yay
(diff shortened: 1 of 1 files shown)


dhh:
Appreciate the consideration, but I don't think this is the right place to put it. I also don't think we need child-safety locks on everything. Removing the terminal that you've set as the default seems like a willful act of self-sabotage 😄. Also, you can just fix it manually if you realize that you've made that mistake.

hyprhex:
I understand you. I didn't create a specific file for it. We can also create a specific tab for set the default terminal on the Omarchy menu. What do you think? @dhh

dhh:
I have been thinking about that, but I also don't want to get in the habit of believing that we need to hide away every config file in Omarchy behind a menu. This is something that happens so rarely that it's imo fine that you need to go to Setup > Defaults to change a config file.

hyprhex:
Ah yeah, you let some space to make people play with config files and learn to customize their configurations.

PR #2211, opened 2025-10-04
Add omarchy-launch-sound script, change Waybar sound module, and improve Ghostty class handling for Impala/Wiremix to open with window rules

Furyfree:
This PR introduces a modular omarchy-launch-sound script for Wiremix, updates the Waybar sound module to use it, and refines Ghostty compatibility by using valid GTK application IDs (com.omarchy.*) for Impala and Wiremix.

These changes ensure Hyprland window rules correctly identify and float both apps when launched from Ghostty, or fall back to the previous class behavior when launched from other terminals.

### DIFF
--- a/config/waybar/config.jsonc
+++ b/config/waybar/config.jsonc
@@ -5,18 +5,14 @@
   "spacing": 0,
   "height": 26,
   "modules-left": ["custom/omarchy", "hyprland/workspaces"],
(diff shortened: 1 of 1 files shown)


abenz1267:
duplicate: https://github.com/basecamp/omarchy/pull/1778

PR #2217, opened 2025-10-04
Add keybinding for locking screen

morgoth:
This is a default keybinding on Ubuntu.
Very often I need a quick screen lock (going for a lunch in office, etc).
I think this might be useful for all, as it promotes good practice to not leave your machine unattended.


tonimarmol:
You already have SUPER + ESC then ENTER

morgoth:
Yes, but these are 2 key actions ;-).
I find it useful, but feel free to close.

dhh:
Appreciate the suggestion, but yes, I'm happy with what we have here.

PR #2230, opened 2025-10-05
added gnome-disk-utility to base-packages

andrepadez:
Added gnome-disk-utility next to nautilus in install/omarchy-base-packages.

Since I installed i have been somewhat frustrated to see this error message when i clicked "Open in Disks" from Nautilus.
image

As I'm a Linux Newb, I assume i'm not the only one going through this.

Just today my question on the Omacom Discord was answered by @pooparoo and i figured the Powers That Be would be happy to have this installed from scratch also.

### DIFF
--- a/default/hypr/apps/steam.conf
+++ b/default/hypr/apps/steam.conf
@@ -1,4 +1,4 @@
-# Float Steam, fullscreen RetroArch
+# Float Steam
 windowrule = float, class:steam
 windowrule = center, class:steam, title:Steam
 windowrule = opacity 1 1, class:steam
--- a/install/omarchy-base.packages
+++ b/install/omarchy-base.packages
@@ -68,6 +68,7 @@ mariadb-libs
 mise
 mpv
 nautilus
+gnome-disk-utility
 noto-fonts
 noto-fonts-cjk
 noto-fonts-emoji


andrepadez:
Next time I'll know to work from branch dev, sorry about that.

dhh:
No sweat. Thanks for adding this!

woopstar:
@dhh Wouldn't this require a migration also for existing users?

andrepadez:
> you can just run `sudo pacman -S gnome-disk-utility` for now

woopstar:
> you can just run `sudo pacman -S gnome-disk-utility` for now

Yes, I can, but some would most likely forget. And the intention for migrations is to make sure fixes are applied properly.

andrepadez:
when a new update comes and this pull request is merged into it, the
application will be installed

woopstar:
> when a new update comes and this pull request is merged into it, the application will be installed

I disagree. The merge of this PR into master will not trigger the application to be installed from a regular `Update -> Omarchy`

andrepadez:
Well i thought so, otherwise the existing users will never get rhe new
stuff which i don't think it's correct.

But I may be wrong

woopstar:
> But I may be wrong

You are.

[omarchy-base.packages](https://github.com/basecamp/omarchy/blob/master/install/omarchy-base.packages) is only touched by https://github.com/basecamp/omarchy/blob/master/install/packaging/base.sh which is only touched by https://github.com/basecamp/omarchy/blob/master/install/packaging/all.sh which is only touched by https://github.com/basecamp/omarchy/blob/master/install.sh

Which.... is only run during the first installation.

If we want to run this, then I'm more than happy to add a PR to fix this. Up to @dhh to decide.

dhh:
Yeah, we do need a migration here. Please do a PR for that.

andrepadez:
Awesome, i will follow that PR so next time I'll know what to do.

PR #2232, opened 2025-10-05
Fix DPI issue on VLC on high DPI monitors with QT_SCALE_FACTOR

NicolasDorier:
Added QT_SCALE_FACTOR environment variable for scaling on high DPI display. Without it, VLC UI  is ridiculously small.

# WARNING

This actually then break OBS Studio UI... everything becomes too big... unsure how I should proceed to have both OBS and VLC working as expected.


dhh:
OBS is part of the default installation, VLC is not, so we can't break the defaults to deal with something optional. You can make your own vlc.desktop file and insert this env there, though. Or you can just use the default mpv player instead of vlc.

PR #2233, opened 2025-10-05
Fix Neovim .desktop file to open files from GUI file manager

AlwaysRead:
Previously, Neovim could only be launched from a terminal, and opening files
from a GUI file manager failed. This change updates the .desktop file to use
`sh -c` with $TERMINAL, allowing files to be opened directly from the GUI.

### DIFF
--- a/applications/nvim.desktop
+++ b/applications/nvim.desktop
@@ -2,7 +2,7 @@
 Name=Neovim
 GenericName=Text Editor
 Comment=Edit text files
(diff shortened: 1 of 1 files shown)


dharmavagabond:
@dhh, this recently merged PR introduces a [previously reported](https://github.com/basecamp/omarchy/issues/1852#issuecomment-3337113614) bug/side-effect with walker and fixed in #1892.

dhh:
Yeah, fixed this. Thanks.

PR #2234, opened 2025-10-05
Add ncdu to base-packages

terrablue:
ncdu (https://dev.yorhel.nl/ncdu, https://archlinux.org/packages/extra/x86_64/ncdu) is a small (installed size 528.3 KB), well maintained, ncurses application to analyse and manage disk usage.

I believe it fits omarchy and would even suggest to add a keybinding for it.

### DIFF
--- a/install/omarchy-base.packages
+++ b/install/omarchy-base.packages
@@ -68,6 +68,7 @@ mariadb-libs
 mise
 mpv
 nautilus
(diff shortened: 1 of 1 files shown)


dhh:
Thanks for the suggestion, but we already have dust, so this wouldn't be something I'd use.

PR #2239, opened 2025-10-05
Support for changing the current player using a hotkey

killeik:
If there are multiple players, such as a Firefox tab with YouTube and Spotify, there is currently no way to change which player will receive commands sent by swayosd-client --playerctl.

Using the key combination `SUPER + Prev/Next`, playerctld switches the current player to the previous/next player in the list. 
I believe there should be built-in support for playerctld, but there is only a hacky way to change the player https://github.com/ErikReider/SwayOSD/issues/167


dhh:
Doesn't seem to be working here: 

```
~ ❯ playerctld shift
Cannot shift: GDBus.Error:org.freedesktop.DBus.Error.NameHasNoOwner: Destination does not exist
```

killeik:
> Doesn't seem to be working here:
> 
> ```
> ~ ❯ playerctld shift
> Cannot shift: GDBus.Error:org.freedesktop.DBus.Error.NameHasNoOwner: Destination does not exist
> ```

To run the `playerctld shift` command, the Playerctl Daemon must be running.
I added it to the startup configuration. When `playerctld` is running, the `playerctld shift` command will work.

dhh:

PR #2240, opened 2025-10-05
Update bindings.conf, Super+L for Screen Lock

atcriteria:
Mirrors the default Windows shortcut for locking the screen rather than needing to use super+esc, enter. Locking the screen on command is an important function and the consistent UX may be appreciated by migrating Windows users.


dhh:
Appreciate the suggestion, but I like keeping the hotkey scope small, and we have Super + Esc, Return 👌

PR #2242, opened 2025-10-05
Login and UKI updates

### LINKED ISSUE #352: Gnome-keyring not creating default keyvault, 1Password MFA issues

### LINKED ISSUE #372: Better Default Configuration for Managing SSH Keys and Secrets

### LINKED ISSUE #1015: After installing VSCode and login in to GitHub there's a missing OS keyring error

### LINKED ISSUE #1619: Multiple Lenovo types not able to boot / get in to BIOS after installation

### LINKED ISSUE #1719: New installs sometimes end up with Limine instead of Omarchy as the #1 boot loader entry

### LINKED ISSUE #2203: [Bug] Persistent Keyring prompt on every Brave browser launch.
### System details

AMD Ryzen 7 7800X3D, NVIDIA 4080 SUPER, Omarchy Latest (3.0.2)

### What's wrong?

On every launch, Brave prompts for the GNOME Keyring password.
**Brave belongs to the same keyring as the default Chromium**.  Chromium does not ask for access to keyring, but Brave does. I don't know why. I checked the flag settings, as well as the configuration files for both browsers - everything seems to be identical. 

Brave Version : 	1.83.109 Chromium: 141.0.7390.55 (Official Build) (64-bit) 
Chromium Version:  140.0.7339.207 (Official Build)

**Note**
From what I can see, I'm not the only struggling wit this problem: https://discord.com/channels/1390012484194275541/1421504252299186208

--- 

Image

ryanrhughes:
Original #1860

- Changes from using Omarchy Seamless Login DM to SDDM with Autologin
- Update to use custom named UKI to (hopefully) resolve issues due to long filename (Fixes #1619)
- Update UKI install rules to always prioritize (Fixes #1719)
- Remove Archinstall created Limine EFI entry
- Create a default keyring to house 2FA tokens and other items (Fixes #352) (Fixes #372)
- Updates VSCode installer to ensure it uses the default keyring (Fixes #1015) (Fixes #2203)
- Adds guards for BTRFS and Limine
- Drops install support for alternate bootloaders

### A note about the default keyring 
This keyring is created with no password so it can be unlocked on boot without any user intervention. This relies entirely on LUKS encryption to provide protection for this info but given the only thing that should exist in here are 2FA tokens and / or SSH keys that would otherwise persists on disk, this is fine to me. There's nothing stopping a user from adding a password to their keyring if they'd prefer that but because we don't present an actual login screen outside of LUKS, the password can't be captured to unlock the keyring.

The alternative path here would be #631 which I ran personally and did work well but is a bit of a non-standard approach.

### Dropping support for alternate bootloaders
At this point, there's no reason someone should be installing fresh with anything but Limine. Because of this, I'm proposing here that we drop it from the install so as to simplify the complexity and permutations of installs we have to deal with.

As a part of this, it was added to the guard in the event someone does a manual install with the wrong bootloader chosen. They could still technically do it but they're on the hook for their own hooks and configs.

### AMI Bios Fix
The change to the UKI name _should_ fix the issue we were having with AMI BIOS based on the testing I had some users do a while back. The actual install here has not been tested yet as we haven't found a volunteer.


install/login/default-keyring.sh
```diff
@@ -0,0 +1,20 @@
+KEYRING_DIR="$HOME/.local/share/keyrings"
+KEYRING_FILE="Default_keyring.keyring"
+DEFAULT_FILE="default"
+
+cat << EOF | tee "$KEYRING_FILE"
```
y0no, line 5: KEYRING_FILE need to be concatenated with KEYRING_DIR no ?

install/login/default-keyring.sh
```diff
@@ -0,0 +1,20 @@
...
+[keyring]
+display-name=Default keyring
+ctime=$(date +%s)
+mtime=0
+lock-on-idle=false
+lock-after=false
+EOF
+
+cat << EOF | tee "$DEFAULT_FILE"
```
y0no, line 14: Same here

ryanrhughes: Good catch. Thank you.

nightdevil00:
setting the keyring with no password makes SDDM unable to login to uwsm session and manual intervention is required. I had to enable multi-users.target.  I used this to go around the issue and be able to login. I was on dev branch, no option from sddm brought me to desktop, login password did not work as keyring was blank.  I have a Lenovo thinkbook 15p 
IMG_8039

[fix_sddm_autologin.sh](https://github.com/user-attachments/files/22784243/fix_sddm_autologin.sh)

celsobenedetti:
> [fix_sddm_autologin.sh](https://github.com/user-attachments/files/22784243/fix_sddm_autologin.sh)

Legit fixed my issue (default sddm login page, as shown in the picture above, and described here #2603).
Thank you sir.

PR #2246, opened 2025-10-05
Auto-detect icons and icon conversion support for webapp/tui install

parnoldx:
## Summary
  - Auto-detect icons from website URLs using Google's favicons service
  - Add ImageMagick conversion support for custom non-PNG icon URLs

 - Resolution for the auto-detected icons depends on what the website provides through favicons
  - Some websites may only offer low-resolution icons, but manifest/html parsing for higher quality icons was intentionally skipped to keep the implementation simple. It would add a lot of code and new problems like robot block with some site like e.g. facebook.
- Here an overview how the auto-detected icons look like for some test sites. Most are pretty good.
all-icons

### Icon Auto-Detection
  - Web apps no longer require manual icon URL input
  - Automatically fetches favicon from Google's favicon service when icon URL is omitted
  - Simplifies web app installation workflow

  ### ImageMagick Icon Conversion
  - Supports conversion of non-PNG image formats (ICO, JPG, SVG, etc.) to PNG
  - Downloads source icon and converts using ImageMagick
  - Falls back to `-flatten` for problematic formats

I also add https:// if no protocol specified for the url so the user can just type hey.com instead of https://hey.com

### DIFF
--- a/bin/omarchy-tui-install
+++ b/bin/omarchy-tui-install
@@ -5,7 +5,7 @@ if [ "$#" -ne 4 ]; then
   APP_NAME=$(gum input --prompt "Name> " --placeholder "My TUI")
   APP_EXEC=$(gum input --prompt "Launch Command> " --placeholder "lazydocker or bash -c 'dust; read -n 1 -s'")
   WINDOW_STYLE=$(gum choose --header "Window style" float tile)
-  ICON_URL=$(gum input --prompt "Icon URL> " --placeholder "See https://dashboardicons.com (must use PNG!)")
+  ICON_URL=$(gum input --prompt "Icon URL> " --placeholder "See https://dashboardicons.com")
 else
   APP_NAME="$1"
   APP_EXEC="$2"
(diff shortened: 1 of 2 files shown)


bin/omarchy-tui-install
```diff
@@ -23,11 +23,29 @@ DESKTOP_FILE="$HOME/.local/share/applications/$APP_NAME.desktop"
...
-  if ! curl -sL -o "$ICON_PATH" "$ICON_URL"; then
-    echo "Error: Failed to download icon."
+  # Extract file extension from URL
+  EXT="${ICON_URL##*.}"
+  EXT="${EXT%%\?*}"
+
+  SOURCE_ICON="$ICON_DIR/${APP_NAME}_source.${EXT}"
+
+  # Download icon
+  if curl -sL -o "$SOURCE_ICON" "$ICON_URL" 2>/dev/null; then
```
dhh: We're using the same logic here and in web app. Let's extract as `omarchy-icon-dl`.

parnoldx: Good point. Also made sure it will always extract the best resolution for multi format icons.

dhh:
Ended up with a simpler solution for this here: https://github.com/basecamp/omarchy/pull/4650. Thanks for working on this though ✌️

PR #2253, opened 2025-10-06
added search option in menu (fuzzy search)

KoushikScripts:
### Add Deep Search to Omarchy Menu
Adds a fuzzy search feature to quickly find any menu item across the entire menu tree without manual navigation. Instead of clicking through Install → Development → JavaScript → Node.js, type "node" and jump directly there.

### Why This Helps
Not everyone remembers exact menu paths, especially for infrequently-used options or when first learning the system. Search makes everything discoverable and faster to access.

### Implementation

- New show_deep_search() function with fuzzy matching
 
- "Search" option added to main menu

- Search window width increased to 600px for readability
 
- All existing menu navigation unchanged 

- Increased the width of the fuzzy search menu for better visibility of items

Usage: Open menu → Search → Type query → Enter
_`Search bar is the first item of the menu so that the user can just hit "Enter" after using "Super + ALT + Space" they will be able to access the fuzzy search with a single additional keystroke`_

### Maintenance Note
When adding menu items, update three locations:

1. The menu function itself (e.g., show_install_menu)
2. The full_menu_list array
4. The case statement in show_deep_search()

image

image

_**If you need any changes or improvements, please let me know.**_

### DIFF
--- a/bin/omarchy-menu
+++ b/bin/omarchy-menu
@@ -3,7 +3,7 @@
 export PATH="$HOME/.local/share/omarchy/bin:$PATH"
 
 # Set to true when going directly to a submenu, so we can exit directly
-BACK_TO_EXIT=false
+BACK_TO_EXIT=true
 
 back_to() {
   local parent_menu="$1"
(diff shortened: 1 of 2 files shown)


abenz1267:
@dhh do you actually consider this? Seems like maintenance hell to me.

I'd rather start migrating the whole thing to proper walker menus.

narra-skr:
Closing in favor of [#2258](https://github.com/basecamp/omarchy/pull/2258) which has a cleaner auto-extraction approach using AWK. Their `extract_menu_actions()` dynamically parses menu items from the code, eliminating the manual three-location maintenance requirement.

PR #2257, opened 2025-10-06
feat(bluetooth): switch desktop integration to bluetui

harshvsri:
replace blueberry launchers with a Bluetui runner, ship bluetui in the base packages, and align Hyprland rules and migrations so the Bluetooth UI opens in a consistent terminal interface

author: harshvsri <harshvsri@gmail.com>

### DIFF
--- a/bin/omarchy-launch-bluetooth
+++ b/bin/omarchy-launch-bluetooth
@@ -1,0 +1,3 @@
+#!/bin/bash
+
+exec setsid uwsm app -- "$TERMINAL" --class=Bluetui -e Bluetui "$@"
--- a/bin/omarchy-menu
+++ b/bin/omarchy-menu
@@ -185,7 +185,7 @@ show_setup_menu() {
     ;;
   *Bluetooth*)
     rfkill unblock bluetooth
-    blueberry
+    omarchy-launch-bluetooth
     ;;
   *Power*) show_setup_power_menu ;;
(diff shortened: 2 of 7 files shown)


guilhermetk:
This was suggested already, check: https://github.com/basecamp/omarchy/pull/967#issuecomment-3214083842

dhh:
This is the 4th PR trying for the same thing. Appreciate the suggestion, but answer still no 😄

harshvsri:
@dhh  @guilhermetk  may i ask what u dont really like about it, it seems quite consistent next to the impala tui
u can share your insights and maybe we could find/build something much better than blueberry.

since we need to rely on rfkill anyways to really "power off" bt n wifi and it also needs two services to run all the time
systemctl --user list-units --type=service --all | grep blue
  app-blueberry\x2dobex\x2dagent@autostart.service    loaded    inactive dead    Bluetooth OBEX Agent
  app-blueberry\x2dtray@autostart.service             loaded    inactive dead    blueberry
  
  and bluetui is a lightweight tui just like impala(iwd) which works with bluez.
  no extra overhead.
  
  btw thanks a lot @dhh for making such an amazing community where u value suggestions from someone like me, a rookie. Its feels like being herd by google or microsoft directly and interacting with their ceo directly.
  Huge love n respect for you sir. @dhh

PR #2258, opened 2025-10-06
Add new 'Search' option to main menu

RushiChaganti:
This PR adds a Search option to the Omarchy main menu, allowing users to quickly find and execute any menu action. The search interface is dmenu-based and integrates themes, fonts, and power profiles for faster navigation.

Note: Another contributor, [[KoushikScripts](https://github.com/KoushikScripts/pull/2253)](https://github.com/KoushikScripts/pull/2253), was also working on a similar fuzzy search feature. Their approach is slightly different, but both aim to improve menu navigation.

**Key Changes:**

* Added `extract_menu_actions()` to list all menu items with their commands.
* Added `show_search_menu()` for searchable menu selection.
* Integrated dynamic themes, fonts, and power profiles.
* Updated `go_to_menu()` and `show_main_menu()` to include the new Search option.

**Benefit:**

Users can quickly access any menu item without navigating multiple sub-menus.

**Main Menu** image

**Search Menu** image

### DIFF
--- a/bin/omarchy-menu
+++ b/bin/omarchy-menu
@@ -33,7 +33,7 @@ menu() {
     fi
   fi
 
-  echo -e "$options" | walker --dmenu --theme dmenu_250 -p "$prompt…" "${args[@]}"
+echo -e "$options" | walker --dmenu --theme dmenu_250 -p "$prompt…" -w 400 "${args[@]}"
 }
 
 terminal() {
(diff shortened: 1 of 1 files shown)


dhh:
Appreciate the suggestion, but I don't love the dynamics of that. If we're going to have a menu-wide search, it has to be fully integrated into the walker setup.

RushiChaganti:
Hey @dhh ,
Just to confirm — you’d like me to integrate this throughout the system, including the walker menus for all apps, and combine everything into a single unified search option, correct?

dhh:
I don't know if this is currently feasible. What I'd prefer is that there's one input field, and that it by default just filters the visible options we have, but, maybe, if you hold down a modifier, like alt, it'll search the global space. But I don't want this to be a big, invasive change.

RushiChaganti:
Hey @dhh ,

I’ve been working on implementing the feature you suggested but ran into a few roadblocks, so I wanted to share the details in case you have any other suggestions.

1.I integrated Walker (App Menu) into the global search, but the app icons aren’t rendering.

2.I’m replacing the entire current menu with the global search since adding a key triggers one of the existing menu items.

I’ve tried several approaches but haven’t been able to resolve these issues.
Would really appreciate your thoughts or any pointers.

Thanks — and sorry for the trouble!

PR #2260, opened 2025-10-06
fix(presentation-terminal): Ensure consistent ASCII logo alignment

vaqMAD:
The presentation terminal's ASCII logo, displayed by 'omarchy-show-logo', would misalign if the user had a non-default font size set in their global Alacritty configuration.

This change fixes the issue by explicitly setting the font size to 9 for this specific terminal instance using the  flag.

This approach guarantees correct rendering of the logo without overriding the user's personal font preferences for their regular terminal sessions.

Before (at 14 font size):
image

After: 
image

### DIFF
--- a/bin/omarchy-launch-floating-terminal-with-presentation
+++ b/bin/omarchy-launch-floating-terminal-with-presentation
@@ -1,4 +1,4 @@
 #!/bin/bash
 
 cmd="$*"
-exec setsid uwsm app -- alacritty --class=Omarchy --title=Omarchy -e bash -c "omarchy-show-logo; $cmd; omarchy-show-done"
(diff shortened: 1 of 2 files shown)


vaqMAD:
Note: the same schema is used in file `omarchy/bin/omarchy-launch-about`, but it was not used in installer / uninstaller (`omarchy/bin/omarchy-launch-floating-terminal-with-presentation`)

lukehsiao:
This is a win for consistency, sure. However, it also has the downside that in some user configurations, this results in _tiny_ font for the update terminal.

image

I personally prefer respecting the user-configured alacritty font size, and would suggest Omarchy revert the font size hardcoding.

PR #2262, opened 2025-10-06
Fix slow network speeds for BCM4360.

rplankenhorn:
After my initial install of Omarchy, my Wifi speeds were incredibly slow.  Less than 5Mbps down.  After switching to this driver, I was getting more normal speeds.

screenshot-2025-10-06_12-11-48

### DIFF
--- a/install/config/hardware/fix-apple-bcm43xx.sh
+++ b/install/config/hardware/fix-apple-bcm43xx.sh
@@ -4,8 +4,12 @@
 
 pci_info=$(lspci -nnv)
 
-if echo "$pci_info" | grep -q "106b:" && 
-  (echo "$pci_info" | grep -q "14e4:43a0" || echo "$pci_info" | grep -q "14e4:4331"); then
-  echo "Apple BCM4360 / BCM4331 detected"
-  sudo pacman -S --noconfirm --needed broadcom-wl dkms linux-headers
+if echo "$pci_info" | grep -q "106b:"; then
+  if echo "$pci_info" | grep -q "14e4:43a0"; then
+    echo "Apple BCM4360 (14e4:43a0) detected — installing broadcom-wl-dkms"
+    sudo pacman -S --noconfirm --needed broadcom-wl-dkms dkms linux-headers
+  elif echo "$pci_info" | grep -q "14e4:4331"; then
+    echo "Apple BCM4331 (14e4:4331) detected — installing broadcom-wl"
+    sudo pacman -S --noconfirm --needed broadcom-wl dkms linux-headers
+  fi
 fi


iamobservable:
@rplankenhorn 

I've experienced intermittent network dropouts with my MBP 13,2 (2016) non-T2 chip device. I found [another reference about supplying kernel parameters](https://wiki.archlinux.org/title/Broadcom_wireless#BCM43602_802.11ac_Wireless_LAN_SoC) that 
is specific to the 13,2 device but uses the 43602 chipset.

Would you be open another commit from me with an additional change for the 13,2 device in this PR? Also a question to @ryanrhughes and the Omarchy team. Would it be better to submit a separate PR?

```bash
cat /sys/class/dmi/id/product_name
```
MacBookPro13,2

```bash
lspci -nnd ::0280
```
02:00.0 Network controller [0280]: Broadcom Inc. and subsidiaries BCM43602 802.11ac Wireless LAN SoC [14e4:43ba] (rev 02)

image


Dropping [this gist link](https://gist.github.com/cristianmiranda/6f269797b62076c3414c3baa848dda67#file-brcmfmac43602-pcie-txt) as well in case it may be useful for other hardware configurations. It is related to Ubuntu, but also has hardware related information.

rplankenhorn:
@iamobservable When you say intermittent network dropouts, are you experiencing a complete Wifi disconnect or it's showing it's still connected and it's extremely slow?

For me, when I was trying to install this additional driver over WiFi, I had to temporarily configure Pacman to only download 1 package at a time instead of 5 in parallel because that was too much for the connection and Pacman would abort.

As I write this, it seems that my speed has dropped back down to the abysmal speed yet again.  

screenshot-2025-10-06_19-21-00

I won't speak for the Omarchy team but if you have a fix for your adapter, maybe do it in a separate PR.  I think I need more time to fully troubleshoot this issue.

rplankenhorn:
Been doing some more testing and I have been getting consistent speeds with the driver and connecting to a dedicated 5G WiFi SSID.  Before, I was connecting to a 2G/5G SSID and I think that was causing issues.

After doing some more research comparing the two drivers, I think we should be using the broadcom-wl-dkms in all cases given that it rebuilds automatically whenever the kernel updates.  Thoughts?

ryanrhughes:
The [broadcom-wl package](https://archlinux.org/packages/extra/x86_64/broadcom-wl/) has broadcom-wl-dkms included as a dependency so unless I'm missing something here; this is effectively just installing broadcom-wl in long form and not really doing anything differently. Any perceived performance gains would only be coincidental or due to change in circumstances.

@iamobservable if you've found a working fix, I'm happy to review a PR for it.

iamobservable:
> The [broadcom-wl package](https://archlinux.org/packages/extra/x86_64/broadcom-wl/) has broadcom-wl-dkms included as a dependency so unless I'm missing something here; this is effectively just installing broadcom-wl in long form and not really doing anything differently. Any perceived performance gains would only be coincidental or due to change in circumstances.
> 
> @iamobservable if you've found a working fix, I'm happy to review a PR for it.

Thanks @ryanrhughes. I don't have anything that presently is ready for a pull request. I'll create a new PR once I spend a little more time researching.

Question:

I did find a kernel parameter definition from the arch linux website, [mentioned above](https://wiki.archlinux.org/title/Broadcom_wireless#BCM43602_802.11ac_Wireless_LAN_SoC), that identifies my device specific chipset on the MacBookPro12,2. I added it to my local macbookpro12,2 testing installation, but have not really seen significant changes from it. It seems the focus is on proximity of the device to the wifi router. I have pretty good coverage across my house so this hasn't been a really testable situation.

Is this valuable to the Omarchy team? I am :100: open to creating a PR with this definition if so, but have not been able to verify through testing locally.

Also, I tried to installing the broadcom-wl package to verify what is in the modprobe file [usr/lib/modprobe.d/broadcom-wl.conf](https://archlinux.org/packages/extra/x86_64/broadcom-wl/files/) with yay and pacman, but have been receiving an error.

```bash
❯ sudo pacman -S broadcom-wl
resolving dependencies...
looking for conflicting packages...

Package (1)        New Version       Net Change

extra/broadcom-wl  6.30.223.271-649    1.61 MiB

Total Installed Size:  1.61 MiB

:: Proceed with installation? [Y/n]
:: Retrieving packages...
 broadcom-wl-6.30.223.271-649-x86_64 is up to date
error: failed retrieving file 'broadcom-wl-6.30.223.271-649-x86_64.pkg.tar.zst.sig' from mirror.omarchy.org : Exceeded the maximum allowed file size (16384) with 16384 bytes
warning: failed to retrieve some files
error: failed to commit transaction (failed to retrieve some files)
Errors occurred, no packages were upgraded.
```

I am a little confused by what is going on there, so it is making research into the broadcom-wl package difficult.

Let me know what you think!

rplankenhorn:
> The [broadcom-wl package](https://archlinux.org/packages/extra/x86_64/broadcom-wl/) has broadcom-wl-dkms included as a dependency so unless I'm missing something here; this is effectively just installing broadcom-wl in long form and not really doing anything differently. Any perceived performance gains would only be coincidental or due to change in circumstances.
> 
> @iamobservable if you've found a working fix, I'm happy to review a PR for it.

@ryanrhughes Thanks for the link.  That makes sense.  I'll roll back to the original driver and see if I'm still seeing issues.

PR #2270, opened 2025-10-07
Add interactive language setup menu and script

agata:
## Summary
- add `bin/omarchy-setup-language` to let users enable/disable locales from the CLI
- surface a new Language item under Setup -> Omarchy so the script is discoverable
- refresh the Omarchy menu wiring so the current locale can be changed or removed directly

## Screenshots
1. Omarchy menu showing the new Language entry

image

2. Setup top-level menu (for overall context)

image

3. Changing the current language

image

4. Removing a language

image

## Testing
- [x] ./bin/omarchy-setup-language (enable locale, switch locale, delete locale)
- [x] Omarchy menu navigation through the new Language entry

## Known issues
- Adding the Language entry makes the Setup menu tall enough that the final Config item is partially truncated (see screenshot 1). We may need a follow-up layout tweak.

### DIFF
--- a/bin/omarchy-menu
+++ b/bin/omarchy-menu
@@ -1,6 +1,7 @@
 #!/bin/bash
 
-export PATH="$HOME/.local/share/omarchy/bin:$PATH"
+script_dir=$(cd -- "$(dirname "${BASH_SOURCE[0]}")" && pwd)
+export PATH="$script_dir:$HOME/.local/share/omarchy/bin:$PATH"
 
 # Set to true when going directly to a submenu, so we can exit directly
 BACK_TO_EXIT=false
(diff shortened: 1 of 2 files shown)


kromsam:
This feels related to https://github.com/basecamp/omarchy/issues/1187 and https://github.com/basecamp/omarchy/pull/1248 . Both were dismissed by DHH.

I would be pro more locale flexibility in Omarchy, already earlier in the installation phase. Currently we allow users to pick a keyboard lay-out during installation, why not a locale too (as is possibile in the original archinstall program)? I want my system to speak my own language and standards, (apart from my CLI of course.) The XDG standards even allows for flexible Home directory naming (see aforementioned PR and issue).

agata:
I understand your point about offering more locale flexibility during installation, similar to how we handle keyboard layouts. My initial thought was also to provide these options upfront.

However, I felt there was an intention to keep the installation steps as minimal and simple as possible. That's why I opted for the approach in this pull request, which allows for easy language changes *after* installation.

I believe most Japanese users, and likely many non-English native speakers, end up changing their language settings post-installation anyway. This process can be a bit tedious, so I think this pull request aligns well with Omarchy concept of making things easy and automated.

kromsam:
The only thing is that when you set the language during installation, the default /home/ directories (like Documents / Downloads / Pictures, etc...) will be automatically named in the correct language. It is quite a hassle to change this after the installation. If Omarchy consistently uses the XDG base directories, instead of hard-coding directory names, this won't be an issue, and widely supported across linux.

https://wiki.archlinux.org/title/XDG_user_directories

agata:
This issue is discussed in #3161.

timlau:
Tested the script and changed the language from da_DK.UTF8 to en_DK.UTF8 it work like a charm, Thanks

It would be nice to have this included into Omarchy, I selected the wrong language at install and this solved it

dhh:
Closing because this targets the retired menu and installer structure. Language setup would need a fresh implementation in the current ISO/settings and Quattro flows. Thanks for the contribution.

Issue #617, opened 2025-08-10
Add up to 3 default backgrounds for every theme that match really well


dhh:
I'd like for every default theme to have 3 default backgrounds. Here are the rules:

- Must look good at 6K. Doesn't mean it HAS to be natively 6K, but it can't be like 1440p. Gotta be minimum 4K and not rely on lines that'll look blurry at 6K.
- Must be 2mb or less. That means JPG for anything that's a picture but PNG for anything that's a graphical design.
- Must fit with the tones and style of the theme.

Please just open 1 pull request per image you want to add. Then we can approve/reject them individually.

dhh:
We are now missing additional backgrounds for:

- [ ] Catppuccin Latte
- [ ] Everforest
- [ ] Gruvbox
- [ ] Kanagawa
- [ ] Matte Black

PR #2280, opened 2025-10-07
Add gruvbox background

Sergiogd112:
Add a background for gruvbox.
The requirements from issue #617  were:
- <2mb
- at least 4k resolution
- follow the colour palette

The image proposed:
- 377.3 kB
- 3840x2160
- follows the palette


dhh:
Thanks for the suggestion, but I don't think this matches the gruvbox colors. It's gotta be a lot more green-focused.

PR #2283, opened 2025-10-07
Add gruvbox background

Sergiogd112:
Add a background for gruvbox.
The requirements from issue #617  were:
- <2mb
- at least 4k resolution
- follow the colour palette (lean towards green based on #2280)

The image proposed:
- 432.6 kB
- 3840 x 2400
- leans towards green


dhh:
Hmm, interesting, but I don't love it. There's some 1990s adventure game nostalgia here, but I don't think this is up to the standard of the others. Thanks for suggesting, though!

PR #144, opened 2025-07-12
hypr: added custom script to use wofi as xdph share-picker

abenz1267:
I've added a script that let's you use any dmenu as a xdph share picker. It'll use wofi, as wofi is the default launcher (for now... ).

PR #2289, opened 2025-10-07
Add custom walker-based share picker with monitor descriptions

hoblin:
## Problem

  The current Qt-based `hyprland-share-picker` has several issues that don't align with Omarchy's design:

  1. **Generic labels**: Shows "Screen 0", "Screen 1", "Screen 2" instead of actual monitor names, making it hard to identify which monitor to share on multi-monitor setups
  2. **Takes half the screen**: The Qt dialog is unnecessarily large and intrusive
  3. **Doesn't support themes**: Uses Qt's default styling which looks out of place in Omarchy
  4. **No keyboard navigation**: In a keyboard-centric system like Omarchy, you can't search or quickly select with the keyboard

  ## Solution

  This PR adds a custom walker-based share picker that:

  - **Shows friendly monitor names** from Hyprland descriptions
  - **Compact and themeable** - Uses walker's dmenu mode, integrates with Omarchy themes
  - **Fully keyboard-driven** - Search and select with keyboard, matching Omarchy's workflow
  - **Supports all features** - Screen/Window/Region selection with restore token support

  ### Before (Qt picker)
 
screenshot-2025-10-07_20-20-41

  ### After (Walker picker)

screenshot-2025-10-07_20-14-30

screenshot-2025-10-07_20-15-33

screenshot-2025-10-07_20-16-12

  ## Implementation

  - **Script**: `bin/omarchy-hyprland-share-picker` - Custom picker using walker
  - **Config**: `default/hypr/xdph.conf` - Configures xdg-desktop-portal-hyprland
  - **Migration**: Copies config to `~/.config/hypr/xdph.conf`

  ## Benefits

  1. **No extra dependencies** - Uses `hyprctl`
  2. **Omarchy consistency** - Matches walker-based design language
  3. **Better UX** - Searchable, keyboard-driven, clear monitor identification

  ## Context

  PR #144 proposed a similar wofi-based picker but was closed

  The Qt picker comes from upstream `xdg-desktop-portal-hyprland`

### DIFF
--- a/bin/omarchy-hyprland-share-picker
+++ b/bin/omarchy-hyprland-share-picker
@@ -1,0 +1,91 @@
+#!/bin/bash
+
+ALLOW_TOKEN=false
+[[ "$1" == "--allow-token" ]] && ALLOW_TOKEN=true
+
+output_selection() {
+  local type="$1"
+  local value="$2"
+
+  echo -n "[SELECTION]"
+  [[ "$ALLOW_TOKEN" == "true" ]] && echo -n "r"
+  echo "/${type}:${value}"
+  exit 0
+}
+
+show_screen_tab() {
+  local options=()
+  local connectors=()
+
+  while IFS='|' read -r connector desc width height; do
+    local friendly_name=$(sed -E 's/ (0x[0-9A-F]+|[A-Z0-9]{10,})$//' <<<"$desc")
+    options+=("$friendly_name (${width}x${height}) - $connector")
+    connectors+=("$connector")
+  done < <(hyprctl monitors -j 2>/dev/null | jq -r '.[] | "\(.name)|\(.description)|\(.width)|\(.height)"' 2>/dev/null)
+
+  local selected=$(printf '%s\n' "${options[@]}" | walker --dmenu -w 350 -p "Share Screen…" 2>/dev/null)
+
+  for i in "${!options[@]}"; do
+    [[ "${options[$i]}" == "$selected" ]] && output_selection "screen" "${connectors[$i]}"
+  done
+
+  exit 1
+}
+
+show_window_tab() {
+  [[ -z "$XDPH_WINDOW_SHARING_LIST" ]] && exit 1
+
+  local options=()
+  local ids=()
+  local rolling="$XDPH_WINDOW_SHARING_LIST"
+
+  while [[ -n "$rolling" && "$rolling" == *"[HC>]"* ]]; do
+    local id="${rolling%%\[HC>\]*}"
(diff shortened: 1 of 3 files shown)


sspaeti:
This looks gorgeous! And just what I worked on today. I had the issue, that the screen picker always was showing up twice, whenever I shared the screen. I could work around by setting the `Allow a restore token` by default, which did the trick. I documented it here: https://www.ssp.sh/brain/screen-sharing-on-wayland-hyprland-with-chrome/

But my next stance was to try to change the eDP and DP-1, as I never know which one is my monitor. I couldn't figure it out. But This looks like a straight forward solution, and looks gorgeous. And everything with the keyboard. Thanks for sharing! If that does not have any side effects, so I couldn't share anymore in certain apps, I'll definitely use this.

hoblin:
In fact, it does fire several times when I tested it with Google Meet. I believe the reason is that it requests access the first time to render a preview in the screen sharing UI itself, and then the second time for the actual call. This is my theory because in Discord and OBS Studio, I did not have this problem. But I would lie if I said that I dug deep into this issue.

abenz1267:
I've created this package https://aur.archlinux.org/packages/xdph-dmenu-sharepicker a while back. it lets you use any dmenu not just walker.

vaqMAD:
Nice ! I was thinking thinking about doing similar PR :)

hoblin:
It broke after update to Omarchy 3.1.1 so I need to fix it.

t4t5:
Opened a PR to make it work with the latest version of Omarchy here 👍️  https://github.com/hoblin/omarchy/pull/1

hoblin:
Thanks @t4t5 
Merged and marked as ready for review.

abenz1267:
I don't really like the multi-step approach. Have you tested the script I shared above? Isn't that... less cumbersome?

hoblin:
> Have you tested the script I shared above?

No, I didn't. Sorry, a little busy. My primary concern is additional dependencies that make it less stable. In the way it is implemented, the only dependencies we use are those that already exist in Omarchy.

abenz1267:
> > Have you tested the script I shared above?
> 
> No, I didn't. Sorry, a little busy. My primary concern is additional dependencies that make it less stable. In the way it is implemented, the only dependencies we use are those that already exist in Omarchy.

it's not a dependency. it's a dmenu script. literally just bash.

NicolasDorier:
I wished to have this one in! The default `hyprland-share-picker` really kills the vibe.
It's like if grandma was dressed in a trash bag and showing up in a fashion show...

PR #2297, opened 2025-10-07
chore: better contrast for ristretto kitty conf

shelldandy:
Adds better contrast to know in which split and tab you might be in kitty:

image

Before:

image


shelldandy:
@dhh can I get a review on this one??

dhh:
Please don't prompt for reviews after a mere two days. Some times it'll take two hours, some times it'll take two weeks. But I'll get to everything in the backlog at some point.

PR #2299, opened 2025-10-07
feat: (alias) added 'cd -' support

alexhraber:
Mac users coming to Omarchy will likely expect `cd -` to work out of the box — it’s ingrained muscle memory for toggling between `$PWD` and `$OLDPWD`. I personally rely on it heavily, and without this change, `cd -` failed because `zoxide` doesn’t handle the `-` argument.

This patch adds a small, safe branch that delegates `cd -` to the built-in `cd`, restoring standard shell behavior while preserving all existing `zd` functionality.

screenshot-2025-10-07_20-02-15

### DIFF
--- a/default/bash/aliases
+++ b/default/bash/aliases
@@ -8,6 +8,14 @@ alias cd="zd"
 zd() {
   if [ $# -eq 0 ]; then
     builtin cd ~ && return
+  elif [[ "$1" == "-" ]]; then
+    if [[ -n "$OLDPWD" ]]; then
+      builtin cd -
+    else
+      print -u2 "zd: OLDPWD not set"
+      return 1
+    fi
+    return
   elif [ -d "$1" ]; then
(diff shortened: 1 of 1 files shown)


default/bash/aliases
```diff
@@ -8,6 +8,14 @@ alias cd="zd"
 zd() {
```
alexhraber, line 8: (unrelated note)

The definition of `zd` and others here in the `aliases` file probably belongs with other function definitions in `default/bash/functions`.

alexhraber: https://github.com/basecamp/omarchy/pull/2305

dhh:
I can't replicate the problem you're having with cd -:

image

alexhraber:
Strange! I'll close this PR and keep the change for my personal usage. Thanks for confirming.

PR #2300, opened 2025-10-07
mise remove fresh

JustinMissmahl:


JustinMissmahl:
Been thinking if a dedicated "mise install" menu point, making use of the `mise registry` command, would maybe make sense. Let me know if I should add it.

dhh:
What I'd like to see is a full _Remove > Development > *_ that's a mirror of _Install > Development_. So it'll need to not just deal with mise, but every other tool that we've made installable there too. Feel free to open a new PR with that pursuit if you'd like ✌️

JustinMissmahl:
> What I'd like to see is a full _Remove > Development > *_ that's a mirror of _Install > Development_. So it'll need to not just deal with mise, but every other tool that we've made installable there too. Feel free to open a new PR with that pursuit if you'd like ✌️

I am about to go to sleep in a couple of hours but I can take a look at it later on.

PR #2303, opened 2025-10-08
Add monitor swap functionality

tamayotchi:
## Description
Adds a new feature to swap active workspaces across all monitors in a circular pattern.

- New script: `omarchy-swap-monitors` for circular workspace rotation across all monitors
- Includes proper error handling and user notifications
- Validates monitor availability before attempting swap

https://github.com/user-attachments/assets/4b32620c-eb4c-4df2-9d8c-f2049f67515c


woopstar:
How does this work with a 3-monitor setup?

rez1-dev:
> How does this work with a 3-monitor setup?

This won't work because `MONITORS[0]` and `MONITORS[1]` are defined.

You'll need to add `MONITORS[2]` for the third monitor in the script. Similarly for 4+ monitors.

For a robust solution, dynamically detect monitors and iterate through them instead of hardcoding indices. No sane person uses more than 3 monitors anyway.

woopstar:
> or a robust solution, dynamically detect monitors and iterate through them instead of hardcoding indices.

I know :) I was just too lazy to point it out. Wanted contributor to think of this instead 👯

woopstar (changes requested):
Maybe this works:

```bash
#!/usr/bin/env bash
# Rotate workspaces across all monitors in Hyprland/Omarchy

readonly TMP_WORKSPACE=999
readonly APP_NAME="Omarchy"

mapfile -t MONITORS < <(hyprctl monitors -j | jq -r '.[].name')

if (( ${#MONITORS[@]} < 2 )); then
    notify-send "$APP_NAME" "Only one monitor detected — nothing to rotate."
    exit 0
fi

get_active_workspace() {
    hyprctl monitors -j | jq -r ".[] | select(.name==\"$1\") | .activeWorkspace.id"
}

# Store all active workspaces
declare -a WORKSPACES
for mon in "${MONITORS[@]}"; do
    ws=$(get_active_workspace "$mon")
    if [[ -z "$ws" ]]; then
        notify-send "$APP_NAME" "Could not detect workspace on $mon."
        exit 1
    fi
    WORKSPACES+=("$ws")
done

# Move current workspace to temp
hyprctl dispatch workspace "$TMP_WORKSPACE"
hyprctl dispatch moveworkspacetomonitor "$TMP_WORKSPACE" "${MONITORS[0]}"

# Rotate: each monitor gets the workspace from the previous monitor
for (( i=${#MONITORS[@]}-1; i>=0; i-- )); do
    prev_idx=$(( (i - 1 + ${#MONITORS[@]}) % ${#MONITORS[@]} ))
    hyprctl dispatch moveworkspacetomonitor "${WORKSPACES[$prev_idx]}" "${MONITORS[$i]}"
done

# Focus the workspace that moved to the first monitor
hyprctl dispatch workspace "${WORKSPACES[-1]}"

notify-send "$APP_NAME" "Rotated workspaces across ${#MONITORS[@]} monitors."
```

woopstar (changes requested):
omarchy-swap-monitors should go into the bin folder

woopstar (approved):
Tested this last night. Works perfect on my 2 and 3 monitor setup.

bin/omarchy-swap-monitors
```diff
@@ -0,0 +1,54 @@
...
+
+# Swap neighbouring monitors from right-to-left to rotate without creating temps.
+for (( i=${#MONITORS[@]}-1; i>=1; i-- )); do
+    dispatch swapactiveworkspaces "${MONITORS[$((i-1))]}" "${MONITORS[$i]}"
+done
+
+# Ensure focus returns to the monitor the user was on.
+dispatch focusmonitor "$ORIGINAL_MONITOR"
+
+notify-send "$APP_NAME" "Rotated workspaces across ${#MONITORS[@]} monitors."
```
woopstar, line 54: Consider to set timeouts on notifications

dhh:
Closing as outdated: this uses the old command/config conventions and the implementation is not robust beyond two monitors. The monitor-swap idea can be re-proposed against the current Quattro monitor stack.

PR #2305, opened 2025-10-08
optional: add 'cd -' and move functional aliases to functions file

alexhraber:
Please see #2299 first, specifically this comment:
https://github.com/basecamp/omarchy/pull/2299/files#r2412387383

This PR moves all functional aliases out of `default/bash/aliases` to `default/bash/functions` and also introduces support for `cd -` as described in the description of #2299.


dhh:
I know that these are technically functions, but they're there as aliases to commands, so I find leaving them where they are as a better fit. Thanks for the suggestion though!

PR #2315, opened 2025-10-08
Show keyboard layout in the Lock screen

hyprhex:
Sometime you have more than one language, and when you are at the lock screen you don't know which layout 

That's why I added to show you the keyboard layout that you are using.

### DIFF
--- a/config/hypr/hyprlock.conf
+++ b/config/hypr/hyprlock.conf
@@ -34,6 +34,18 @@ input-field {
     fade_on_empty = false
 }
 
+# Keyboard LAYOUT 
+label { 
+    monitor = 
+    text = $LAYOUT 
+    color =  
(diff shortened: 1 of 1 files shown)


dhh:
Can you show a screenshot of what this looks like?

hyprhex:
hyprlock

Show the keyboard layout under the box, and if you change it, it will change

hyprhex:
I set the position of the layout under the box, and you can also play with the style on how it looks, but I like it like that.

dhh:
Hmm, the white doesn't look great on lighter backgrounds, but worse, it's there even when you only have a single language defined. I don't think that's worth the aesthetic trade-off for most people. So I think we should add this to the Discussions > Manual explaining how to make this change on your own config when you need it 👌

PR #2327, opened 2025-10-09
Fix asdcontrol typo

sa-dd:
`asdcontrol-git` should be `asdcontrol` since the former doesn't exist. 

Screenshot_20251009_061441


ryanrhughes:
This is already fixed in dev and the package has been restored to https://pkgs.omarchy.org for now. Thanks for catching this and pushing it up.

PR #2347, opened 2025-10-10
feat: always make bins executable on update

woopstar:


dhh:
What problem are you trying to fix? The permissions should be correct in the source. Shouldn't have to retroactively apply anything like this.

dhh:
Actually, regardless of the problem, this isn't going to fly. Have to solve specific problems, not just plaster over them.

PR #2350, opened 2025-10-10
Disable opacity for Netflix

mvaled:


dhh:
We now have `Super + Backspace` to remove transparency. I'd rather stick to that for this.

PR #2353, opened 2025-10-10
screensaver: Use the monitor refresh rate

defer:
When launching, use the refresh rate of each monitor instead of the hardcoded 240.

My selfish motivation is just to stop hearing fans on my laptop, but it makes sense to avoid wasting cycles past the monitor refresh rate.


dhh:
The framerate just sets the speed of the animation. It doesn't have anything to do with the display itself. It's a bit confusing, but 240 is just there to determine the speed.

PR #2360, opened 2025-10-10
Added the Option to install Jetbrains

PhilIsHere:
I assume that a good chunk of devs use the JetBrains Suite.

The thing is, if you don't know that it not only requires the package but also the runtime, you spend unnecessary time to find what you're looking for. So I tried to fix this, and that is the solution. I hope you like it and merge it.

It basically asks the user what he wants to install (accepts a number) and then does its thing. :-)

### DIFF
--- a/bin/omarchy-install-jetbrains
+++ b/bin/omarchy-install-jetbrains
@@ -1,0 +1,94 @@
+#!/bin/bash
+if [[ -z "$1" ]]; then
+  echo "What would you like to install?"
+  echo " 0 - Nothing (exit)"
+  echo " 1 - full"
+  echo " 2 - clion"
+  echo " 3 - datagrip"
+  echo " 4 - dataspell"
+  echo " 5 - intellij-idea"
+  echo " 6 - phpstorm"
+  echo " 7 - pycharm"
+  echo " 8 - rider"
+  echo " 9 - webstorm"
+  while true; do
+    read -r -p "Enter the number of your choice: " selection
+    case "$selection" in
+      0)
+        echo "Exiting without installing anything."
+        exit 0
+        ;;
+      1) set -- full; break ;;
+      2) set -- clion; break ;;
+      3) set -- datagrip; break ;;
+      4) set -- dataspell; break ;;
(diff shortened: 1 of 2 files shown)


defer:
Interesting, before reading I was half expecting this to be just an yay install of jetbrains-toolbox. Is there a benefit on these packages when compared to their toolbox variants?

PhilIsHere:
> Interesting, before reading I was half expecting this to be just an yay install of jetbrains-toolbox. Is there a benefit on these packages when compared to their toolbox variants? 

In my case, when installing the Jetbrains Tools via the toolbox, the dimensions and resolution have been very weird. The file explorer was very big, the code field and text size small and so on.

I did not have had these issues with the packages from the AUR (I tested phpstorm and goland).

In my opinion this is just a more streamlined experience. I'm sure I could've somehow fixed the mentioned issue with the toolbox variants. But why bother? And now people don't need to bother to fix it either and can start quicker :)

histfak:
It seems that you missed RubyMine and RustRover.

dhh:
Thanks for the suggestion, but there are still a variety of display issues and incompatibilities with the JetBrains suite. So I don't want to have a direct path to installation in the menus before that's all sorted. JetBrains is working on it, though! When they're ready, we can have a menu item to install the Toolbox.

PhilIsHere:
> Thanks for the suggestion, but there are still a variety of display issues and incompatibilities with the JetBrains suite. So I don't want to have a direct path to installation in the menus before that's all sorted. JetBrains is working on it, though! When they're ready, we can have a menu item to install the Toolbox.

My commit doesn't include the toolbox at all and solves the display issues you are describing. I had them as well when trying the toolbox. I do not have the display issues with the individual packages which are used in my suggestion.

PhilIsHere:
@histfak thank you for pointing that out. I added the options.

I'm not quite sure why my Commit doesn't show up here, as far as I'm aware it should update automatically, no? Do I have to press "Create pull request" again? I didn't want to accidentally open a second one, but wanted this fix in, if @dhh reconsiders my solution and indeed merges it.

image

PR #2371, opened 2025-10-11
Show local branches in update branch menu

ludagoo:
I find it useful to also have any local branches that I am working on show up in the Update > Branch menu.


dhh:
I think if you have this level of sophistication, you're better off just changing branches in the terminal. This was just added such that people on beta releases had an easy way to get back on master. So appreciate the suggestion, but I don't really want to complicate this script for such a rare event.

PR #2381, opened 2025-10-11
fix: correct typo in command usage (verion -> version)

danielpavone:
The command name was misspelled as 'omarchy-verion-branch-set', which caused confusion when using the script. Updated it to 'omarchy-version-branch set' for consistency and correct usage output.

### DIFF
--- a/bin/omarchy-update-branch
+++ b/bin/omarchy-update-branch
@@ -3,7 +3,7 @@
 set -e
 
 if (($# == 0)); then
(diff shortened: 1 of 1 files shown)


ryanrhughes:
The update here is technically incorrect as well but I corrected it and gave you co-author credit. Thanks for pointing it out. https://github.com/basecamp/omarchy/commit/01a1e43b2584d2ef5ada68334437712597e3aec1

Issue #2184, opened 2025-10-03
Chrome crashes when moving tile between monitors on Hyprland


Tatuck:
### System details

AMD Ryzen 7 7840HS, NVIDIA GeForce RTX 4060 Max-Q, AMD Radeon 780M

### What's wrong?

Hello.
When running Google Chrome with Wayland (--ozone-platform=wayland) under Omarchy (Hyprland), Chrome crashes if the window is moved from one monitor to another.
The following error appears in the terminal:
```
~ ❯ google-chrome-stable --ozone-platform=wayland
[25566:25566:1003/120242.562772:ERROR:ui/ozone/platform/wayland/host/wayland_wp_image_description.cc:219] Incomplete image description info from compositor.
[25566:25566:1003/120242.562831:ERROR:ui/ozone/platform/wayland/host/wayland_wp_image_description.cc:219] Incomplete image description info from compositor.
[25566:25566:1003/120243.276944:ERROR:ui/ozone/platform/wayland/host/wayland_wp_image_description.cc:219] Incomplete image description info from compositor.
[25566:25566:1003/120244.619696:ERROR:ui/events/platform/wayland/wayland_event_watcher.cc:47] libwayland: wl_display#1: error 0: invalid object 63

[1003/120244.631334:ERROR:third_party/crashpad/crashpad/snapshot/elf/elf_dynamic_array_reader.h:64] tag not found
[1003/120244.631782:ERROR:third_party/crashpad/crashpad/snapshot/elf/elf_dynamic_array_reader.h:64] tag not found
Trace/breakpoint trap      (core dumped) google-chrome-stable --ozone-platform=wayland
```
I was able to do this without having to change anything, but now I am receiving that error. Could it be because of the update? (v3.0.2)

### Steps to reproduce:
1. Launch Chrome with Wayland backend: `google-chrome-stable --ozone-platform=wayland`
2. Move the Chrome window from one monitor to another using `SUPER + SHIFT + 2`
3. Chrome crashes immediately with the above error.

### Workarounds tried:
1. Running with `--ozone-platform=x11` works, but the scaling is broken (window contents appear oversized). The scaling can be fixed using `--force-device-scale-factor=1`.
2. Updating Chrome and Hyprland didn’t resolve the issue.

I know that running Chrome with `--ozone-platform=x11 --force-device-scale-factor=1` works as a workaround, but I’d like to understand why I didn’t need these flags before to move windows between monitors without crashing.

### System info:
[...]

PR #2394, opened 2025-10-12
fix: prevent Chromium crash when moving windows between monitors

### LINKED ISSUE #2384: Bug: Chromium crashes on Hyprland after update when moving chromium browsers between monitors
### System details

i5-1335U, Intel Iris Xe, Omarchy v3.0.2

### What's wrong?

**Started after upgrading to:**
omarchy-chromium 141.0.7390.76-2 (2025-10-12)

**What happens**
After updating Chromium, moving Chromium window between monitors causes it to crash and sometimes triggers a Hyprland restart.

**Temporary workaround I found**
Downgrading these packages resolves the issue:
```
sudo pacman -U /var/cache/pacman/pkg/omarchy-chromium-141.0.7390.54-1002-x86_64.pkg.tar.zst
```

**To prevent auto-upgrade until fix arrives**
```
/etc/pacman.conf
IgnorePkg = omarchy-chromium
```
**Do the same for Brave installed via Flatpak**
```
View commit history:
flatpak remote-info --log flathub com.brave.Browser

Downgrade to older commit:
sudo flatpak update --commit=6ab4e1add8dd29f7a94fe515e77e475f5145718b72314629a7900c0c68292e08 com.brave.Browser

Prevent auto-updates:
sudo flatpak mask com.brave.Browser
```

Now dragging or moving chromium or brave window doesn't cause a crash

d-cas:
# Fix Chromium Crash When Moving Windows Between Monitors

## Problem
Chromium-based browsers (Chrome, Brave, etc.) crash instantly when dragging windows between monitors or moving them to different workspaces using Hyprland keybinds (super+shift+number). This started after recent Chromium updates enabled Wayland color management, which conflicts with Hyprland's implementation.

**Note:** A fix has [landed upstream in Chromium](https://chromium-review.googlesource.com/c/chromium/src/+/7003036) but won't reach stable releases for approximately a month. This PR provides a temporary workaround in the meantime.

## Solution
Add `--disable-features=WaylandWpColorManagerV1` to default flag configs:
- `config/chromium-flags.conf`
- `config/brave-flags.conf`

These are deployed to `~/.config/` during Omarchy setup and apply automatically to all Chromium-based browsers.

## Impact
- ✅ Fixes crash when dragging between monitors
- ✅ Fixes crash when using super+shift+number to move windows
- ✅ Applies to all Chromium-based browsers
- ⚠️ Temporary workaround until upstream fix reaches stable

Users can edit `~/.config/chromium-flags.conf` to customize or remove the flag once the upstream fix lands in their Chromium version.

## Testing
```bash
# Move browser window between monitors or workspaces
# Should not crash

# Verify flag is active:
ps aux | grep chromium | grep WaylandWpColorManagerV1
```

## References
- Fixes #2384
- Refs #2184
- Chromium fix (merged, pending stable): https://chromium-review.googlesource.com/c/chromium/src/+/7003036
- Chromium bug tracker: https://issues.chromium.org/issues/40263183
- Hyprland issue: https://github.com/hyprwm/Hyprland/issues/11957
- Hyprland discussion: https://github.com/hyprwm/Hyprland/discussions/11961

### DIFF
--- a/migrations/1760401344.sh
+++ b/migrations/1760401344.sh
@@ -2,10 +2,14 @@ echo "Add Chromium crash workaround flag for Hyprland to existing configs"
 
 # Add flag to chromium-flags.conf if it exists and doesn't already have it
 if [[ -f ~/.config/chromium-flags.conf ]]; then
-  grep -q '\-\-disable-features=WaylandWpColorManagerV1' ~/.config/chromium-flags.conf || sed -i '$a # Chromium crash workaround for Wayland color management on Hyprland - see https://github.com/hyprwm/Hyprland/issues/11957\n--disable-features=WaylandWpColorManagerV1' ~/.config/chromium-flags.conf
+  if ! grep -qF -- "--disable-features=WaylandWpColorManagerV1" ~/.config/chromium-flags.conf; then
+    sed -i '$a # Chromium crash workaround for Wayland color management on Hyprland - see https://github.com/hyprwm/Hyprland/issues/11957\n--disable-features=WaylandWpColorManagerV1' ~/.config/chromium-flags.conf
+  fi
 fi
 
 # Add flag to brave-flags.conf if it exists and doesn't already have it
 if [[ -f ~/.config/brave-flags.conf ]]; then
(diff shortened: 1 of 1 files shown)


dhh:
Need a migration to add this line to existing configs (use grep to check it's not already there, then sed to add it, see other migrations for the pattern).

d-cas:
@dhh Added migration script to update existing user configs. The script:
   - Checks if configs exist before modifying
   - Uses grep to verify flag isn't already present  
   - Adds the workaround flag to both Chromium and Brave configs

Ready for review!

d-cas:
Updated migration script to use `sed` instead of `echo`, following the pattern from migrations/1752543801.sh. More concise and matches the existing migration style.

PR #2400, opened 2025-10-12
Add jetbrains suite

PhilIsHere:
This Adds the option in the Menu to install _individual_ JetBrains Packages from the AUR and the respective runtime.
This, by design, _circumvents_ the Toolbox to avoid display issues in the IDEs.

The individual packages do not, at least in my case, have the display issues. It's a install and start to build kind of experience.

I'm sorry if this accidentally opens a new PR instead of updating the existing one https://github.com/basecamp/omarchy/pull/2360
Right now I'm trying to add my second commit which adds RustRover and RubyMine to the options.

Edit: I sincerly apologize. I'm not sure how I'm supposed to update the existing PR. How I understood it, this should've have happened automatically, as soon as I pushed my commit to the branch, but it didn't so I tried to do the Pull Request thing again, expecting this step is needed to update the PR. It's not my intention to spam this feature request. If someone can point me to the right direction on how to Update existing PRs this would be highly appreciated so I can avoid this in the future. Thank you for your understanding.

### DIFF
--- a/bin/omarchy-install-jetbrains
+++ b/bin/omarchy-install-jetbrains
@@ -3,14 +3,16 @@ if [[ -z "$1" ]]; then
   echo "What would you like to install?"
   echo " 0 - Nothing (exit)"
   echo " 1 - full"
-  echo " 2 - clion"
-  echo " 3 - datagrip"
-  echo " 4 - dataspell"
-  echo " 5 - intellij-idea"
-  echo " 6 - phpstorm"
-  echo " 7 - pycharm"
-  echo " 8 - rider"
-  echo " 9 - webstorm"
+  echo " 2 - CLion"
+  echo " 3 - DataGrip"
+  echo " 4 - DataSpell"
+  echo " 5 - IntelliJ IDEA"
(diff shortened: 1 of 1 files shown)


jarlah:
i would say its essential to separate between community and ultimate of intellij. I use mostly community for my small projects on this computer. So default ultimate is a bit problematic given the license nagging

dhh:
My position hasn't changed from yesterday: https://github.com/basecamp/omarchy/pull/2360#issuecomment-3393227258

Appreciate the work, but a) I believe we should just be installing the JetBrains Toolbox and b) not until it's fully wayland/hyprland compatible out of the box.

jarlah:
> My position hasn't changed from yesterday: [#2360 (comment)](https://github.com/basecamp/omarchy/pull/2360#issuecomment-3393227258)
> 
> Appreciate the work, but a) I believe we should just be installing the JetBrains Toolbox and b) not until it's fully wayland/hyprland compatible out of the box.

totally agree, its also easier to sign into license servers etc cross products

Issue #2272, opened 2025-10-07
Chromium asked for password keyring and everything is gray?


Luquatic:
### System details

NVIDIA 5090, Omarchy latest

### What's wrong?

After just booting up my pc I got a pop-up when opening chromium about a password-keyring and after setting one and continue on my chromium browser is now black and white/gray? Any idea how to fix this?

Luquatic:
<img width="3838" height="2161" alt="Image" src="https://github.com/user-attachments/assets/94d49263-0f6c-4037-90e7-e7c22b18855a" />

prathamngundikere:
Even I got the same message. It was asking for Default Keyring password everytime I restart the desktop. I kind of disabled the password manager inside chrome, and removed some VPN extensions (for no reason). This time when I restart it was good. No pop up message like that anymore.

prathamngundikere:
But IDK about the white/grey thing...

sntsntsntsnt:
Had same issue. Also now themes on Chromium are malfunctioning and shades of color are off so now the tab bar doesn't blend with the waybar as it did.

Luquatic:
> Had same issue. Also now themes on Chromium are malfunctioning and shades of color are off so now the tab bar doesn't blend with the waybar as it did.

Had? You fixed it?

BernhardRode:
Same here. Interestingly, it occurs in Chrome and Chromium. Firefox is just fine.

<img width="3842" height="2162" alt="Image" src="https://github.com/user-attachments/assets/fb061ce9-198d-4d0e-ba78-b53d6d45acba" />


[2025-10-02T21:59:09+0200] [ALPM] upgraded google-chrome (140.0.7339.207-1 -> 141.0.7390.54-1)
[2025-10-07T00:06:37+0200] [ALPM] upgraded omarchy-chromium (140.0.7339.207-2 -> 141.0.7390.54-1002)

Checking the Google Chrome Release notes, i do not see any obvious changes.

https://developer.chrome.com/release-notes/141

sntsntsntsnt:
> > Had same issue. Also now themes on Chromium are malfunctioning and shades of color are off so now the tab bar doesn't blend with the waybar as it did.
> 
> Had? You fixed it?

Sorry, I wasn't very specific. I had the issue where it would require me to input my password and it went away once I did. Haven't got the gray screen bug.

Luquatic:
I fixed the password-keyring popup but everything is still gray

harusharu:
[...]

PR #2409, opened 2025-10-12
fix: chromium gray colors

Luquatic:
solves issue [#2272](https://github.com/basecamp/omarchy/issues/2272#issuecomment-3393103756)


dhh:
Duplicate of #2394

PR #2410, opened 2025-10-12
Waybar battery indicator: show capacity and remaining time when charging or discharging

jelenv:
When charging or discharging show the current capacity in % left to the battery icon. Additionally in the tooltip show the remaining time until full discharge or charge. Other battery states are kept unchanged.

This is an enhancement worthy of adding to the default waybar config. It helps with knowing the battery state, personally I was missing knowing how much time I have left when on battery.

### DIFF
--- a/config/waybar/config.jsonc
+++ b/config/waybar/config.jsonc
@@ -78,16 +78,16 @@
   },
   "battery": {
     "format": "{capacity}% {icon}",
-    "format-discharging": "{icon}",
-    "format-charging": "{icon}",
+    "format-discharging": "{capacity}% {icon}",
+    "format-charging": "{capacity}% {icon}",
     "format-plugged": "",
     "format-icons": {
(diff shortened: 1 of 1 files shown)


morgoth:
Personally I would only include battery capacity when very low, ie from 10% down.

dhh:
Thanks for the suggestion, but I like to keep the waybar free of any moving information.

jelenv:
That's understandable. Could we keep only the tooltip extension? Shall I edit this PR or create a new one?

dhh:
I don't actually love the time thing either because it's so variable to the power draw. And having it hop back and forth between 4h left and 6.5h left isn't very helpful imo. I think waybar needs something more like a moving average before this would be a useful addition.

jelenv:
I can see that. Thanks for taking the time to answer.

Rishabh-Sarang:
> Thanks for the suggestion, but I like to keep the waybar free of any moving information.

Why is that?

PR #2425, opened 2025-10-13
fix: make bcm43xx driver script vendor agnostic

sa-dd:
Remove the Apple vendor ID restriction to enable Broadcom driver installation on non-Apple hardware with compatible chipsets


sa-dd:
Sorry, forgot to change the file name inside `all.sh`

PR #2428, opened 2025-10-13
show the battery percentage

Rishabh-Sarang:


morgoth:
This was already proposed and rejected https://github.com/basecamp/omarchy/pull/2410

PR #2432, opened 2025-10-13
ghostty: Don't toggle fullscreen on ctrl+enter

defer:
Standard fullscreen toggles from omarchy work well across all apps, no need for an app specific one.


dhh:
I don't think this is necessary. Fine for the terminals to also have their own bindings on top of what Omarchy has as standard.

PR #2438, opened 2025-10-14
Size `tag:floating-window` relatively, not absolutely

lukehsiao:
Size `tag:floating-window` relatively, not absolutely

By using percentages, the default size of the `tag:floating-window`
windows look more consistent regardless of screen resolution.

The default of 45% and 60% are selected by simply rounding up to the
nearest 5% using the existing absolute values on a 1080p screen. (800 /
1920 = 41.7%, 600 / 1080 = 55.6%).

---

### DIFF
--- a/default/hypr/apps/system.conf
+++ b/default/hypr/apps/system.conf
@@ -1,7 +1,7 @@
 # Floating windows
 windowrule = float, tag:floating-window
 windowrule = center, tag:floating-window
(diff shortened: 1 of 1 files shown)


dhh:
Hmm, I think the 800x600 sizing is correct regardless of what the size of the screen is. It ensures we have enough space for all these terminal actions, but it doesn't need to have more/fill up the screen on a bigger display.

PR #2445, opened 2025-10-14
Revert "fix(presentation-terminal): Ensure consistent ASCII logo alignment (#2260)"

lukehsiao:
Revert "fix(presentation-terminal): Ensure consistent ASCII logo alignment (#2260)"

This reverts commit 6892304277e0045dee15c79d6fa7f02c35b37f62.

While a hard-coded font size of 9 likely is small enough to fix the logo
rendering on a large population of display configurations, it has the
downside of potentially making the update terminal font painfully small
since it ignores the user's configured preferences.

Rather than somewhat arbitrarily drawing a hard-coded line in the sand
(which still doesn't solve the issue for even smaller displays), just
respect the user's preference, allowing them to easily configure a size
that both looks good and doesn't warp the logo.

Signed-off-by: Luke Hsiao <luke@hsiao.dev>

---

## Example of the extremely small size-9 font on some displays

image

Ref: https://github.com/basecamp/omarchy/pull/2260

### DIFF
--- a/bin/omarchy-launch-floating-terminal-with-presentation
+++ b/bin/omarchy-launch-floating-terminal-with-presentation
@@ -1,4 +1,4 @@
 #!/bin/bash
 
 cmd="$*"
-exec setsid uwsm-app -- alacritty -o font.size=9 --class=Omarchy --title=Omarchy -e bash -c "omarchy-show-logo; $cmd; omarchy-show-done"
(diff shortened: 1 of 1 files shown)


dhh:
Going to need something that solves the original issue, if we're going to remove the 9pt fix. Maybe there's a way to prevent things from wrapping? But I'd rather have something that looks a little small than something that looks fully broken.

PR #2461, opened 2025-10-15
integrate  global search for applications and menu items

RushiChaganti:
This update introduces a  global search that integrates both applications and menu items into a single searchable interface.

**Details**:

Applications and menu entries are now combined into one seamless search experience.

Search results display inline counts for better clarity.

The global search can be accessed by pressing Esc from within any menu (as modifying other keybindings would require extensive structural changes — following @dhh  request for minimal modification).

It remains accessible from the main menu as well.

**Notes**:

This code was collaboratively developed with the assistance of various LLMs and has been thoroughly reviewed and tested by a human to ensure correctness and reliability.

Although it has been tested thoroughly, some minor bugs may still exist.

You can comment on this PR if you encounter any issues — I’ll address them promptly.

Open for suggestions and feedback for further improvements.

image
image

https://github.com/user-attachments/assets/e2a2cb00-ff10-415b-9bb7-8f94ed8bf0b3

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.yml
+++ b/.github/ISSUE_TEMPLATE/bug.yml
@@ -1,5 +1,5 @@
 name: Bug
-description: Report a problem
+description: Report a validated bug -- NOT FOR SUPPORT REQUESTS
 labels: [bug]
 body:
   - type: markdown
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,5 +1,5 @@
 blank_issues_enabled: false
 contact_links:
-  - name: 🛟 Community Support
+  - name: 🛟 Support 🛟
     url: https://omarchy.org/discord
     about: Need help? Join our Discord community for support with any issues. GitHub issues should be used for verified bugs, documentation, and enhancement proposals only.
--- a/.github/ISSUE_TEMPLATE/enhancement.yml
+++ b/.github/ISSUE_TEMPLATE/enhancement.yml
@@ -1,5 +1,5 @@
 name: Enhancement
-description: Feature or change
+description: Suggest a feature or change
 labels: [enhancement]
 body:
   - type: markdown


shyamenk:
Wow, it's a nice feature @RushiChaganti

bin/omarchy-menu
```diff
@@ -459,12 +459,28 @@ show_system_menu() {
...
+    "$LAUNCHER" unified
+    unset OMARCHY_MENU_ACTIVE
+}
+
+# --bindhold "Alt_L:launch_unified_search"
+
 go_to_menu() {
   case "${1,,}" in
   *apps*) walker -p "Launch…" ;;
+  *search*) /omarchy-global-search ;;
```
celsobenedetti: this is a typo, correct?
```suggestion
  *search*) omarchy-global-search ;;
```

bin/omarchy-menu
```diff
@@ -459,12 +459,28 @@ show_system_menu() {
 }
 
 show_main_menu() {
-  go_to_menu "$(menu "Go" "󰀻  Apps\n󰧑  Learn\n󱓞  Trigger\n  Style\n  Setup\n󰉉  Install\n󰭌  Remove\n  Update\n  About\n  System")"
+  local selection
+  selection=$(menu  "Go" "󰀻  Apps\n󰍉  Search\n󰧑  Learn\n󱓞  Trigger\n  Style\n  Setup\n󰉉  Install\n󰭌  Remove\n  Update\n  About\n  System")
+  if [[ -z "$selection" ]]; then
+    $HOME/test/test.sh
+  else
```
celsobenedetti: is this supposed to be commited?

bin/omarchy-menu
```diff
@@ -459,12 +459,28 @@ show_system_menu() {
...
+    go_to_menu "$selection"
+  fi
 }
 
+LAUNCHER="${LAUNCHER:-$HOME/test/test.sh}"
+launch_unified_search() {
+    export OMARCHY_MENU_ACTIVE=1
+    "$LAUNCHER" unified
+    unset OMARCHY_MENU_ACTIVE
+}
```
celsobenedetti: again, I'm supposing `test.sh` is some WIP?

RushiChaganti: No, it’s finished. I’ve been working on the file and renamed it to **omarchy-global-search**, but I’ve forgotten parts of the code.

RushiChaganti:
Hey @celsobenedetti  Fixed the issue,
Thanks for pointing it out!!

bin/omarchy-global-search
```diff
@@ -0,0 +1,572 @@
...
+                seen["$name"]=1
+                output+="APP|||${name}|||$(basename "$desktop_file")|||${exec_cmd}"$'\n'
+            fi
+        done
+    done
+    
+    # Save to cache
+    echo -n "$output" > "$APPS_LIST_CACHE"
+    echo -n "$output"
+}
```
celsobenedetti: Can we try and simplify this? 
I'm thinking we need no caching at all for listing available desktop apps.
The way `walker` does it is though [elephant](https://github.com/abenz1267/elephant#), which we could query.

I'm thinking however it can be even simpler here.
We could `fd --type file --extension desktop {dirs}` to find the desktop files.
We don't need to grep for the `Exec` of each file, as we can instead `xdg-launch $file`, which works for all files.

We only need then to find the `Name` of each file so we can present it as a label on the menu.
We could grep for it, but that too can be avoided for performance: the script below used pattern matching to find the name for each file.

```bash
get_applications_cached() {
    app_dirs=(
        "/usr/share/applications"
        "/usr/local/share/applications"
        "$HOME/.local/share/applications"
    )

    while IFS= read -r -d '' desktop_file; do
        name=""
        while IFS= read -r line; do
            case "$line" in
            Name=*)
                name="${line#Name=}"
                break
                ;;
            esac
        done <"$desktop_file"
        printf "APP|||%s|||%s|||xdg-launch %s\n" "$name" "$(basename "$desktop_file")" "$desktop_file"
    done < <(fd -t f -e desktop -0 . "${app_dirs[@]}" 2>/dev/null)
}
```

With the script above I see good performance. We can run it everytime without the need to cache anything.
We could remove all code related to the `APPS_LIST_CACHE` and `APPS_CACHE` variables.

I'd personally prefer to avoid the complexity of caching, and I do believe it can be avoided. 

### update

> [!NOTE]
I suggested `xdg-launch` above, but the correct command is `gtk-launch`, sorry!

celsobenedetti (changes requested):

bin/omarchy-global-search
RushiChaganti: That looks great — I really like the simplification.
I didn’t know about using fd for this earlier, so I went with the cache-based approach to avoid repeated lookups.
But yeah, with fd it’s fast enough that caching isn’t really needed anymore.

Let’s drop the cache logic and go with your version — it’s cleaner, more maintainable, and avoids unnecessary complexity.

bin/omarchy-global-search
```diff
@@ -0,0 +1,545 @@
...
+        name=""
+        while IFS= read -r line; do
+            case "$line" in
+            Name=*)
+                name="${line#Name=}"
+                break
+                ;;
+            esac
+        done <"$desktop_file"
+        printf "APP|||%s|||%s|||xdg-launch %s\n" "$name" "$(basename "$desktop_file")" "$desktop_file"
```
celsobenedetti: I had suggestioned `xdg-launch` here, which is incorrect. Sorry! (the correct is `gtk-launch`)

But from what I'm seeing this printf could be simpler.
We already launch the selection with `gtk-launch` in the `unified_search` function.

```suggestion
        printf "APP|||%s|||%s\n" "$name" "$(basename "$desktop_file")"
```

bin/omarchy-global-search
```diff
@@ -0,0 +1,545 @@
...
+        else
+            return 1
+        fi
+    fi
+}
+
+# -------------------------------
+# 🔹 Cached Application List
+# -------------------------------
+get_applications_cached() {
```
celsobenedetti: since we are not caching the applications, can we call it `get_applications`?
```diff
diff --git a/bin/omarchy-global-search b/bin/omarchy-global-search
index e928d42..6bbe23a 100755
--- a/bin/omarchy-global-search
+++ b/bin/omarchy-global-search
@@ -222,9 +222,9 @@ ensure_menu_json() {
 }
 
 # -------------------------------
-# 🔹 Cached Application List
+# 🔹 Get Application List
 # -------------------------------
-get_applications_cached() {
+get_applications() {
     app_dirs=(
         "/usr/share/applications"
         "/usr/local/share/applications"
@@ -408,7 +408,7 @@ unified_search() {
 
     # Load cached applications (fast - cached list)
     local app_opts
-    mapfile -t app_opts < <(get_applications_cached)
+    mapfile -t app_opts < <(get_applications)
     all_options+=("${app_opts[@]}")
     app_count=${#app_opts[@]}
 
```

bin/omarchy-global-search
```diff
@@ -0,0 +1,545 @@
...
+
+    local stored_hash=""
+    [[ -f "$cache_file" ]] && stored_hash=$(cat "$cache_file" 2>/dev/null)
+
+    if [[ "$current_hash" != "$stored_hash" ]]; then
+        echo "$current_hash" > "$cache_file"
+        return 0  # Changed
+    fi
+    return 1  # Not changed
+}
```
celsobenedetti: With the changes of 9a6e8645106f1067c3c619eb0cfa439b79745146 I think we can remove a lot of code here.
Can we remove all code related to caching?

```diff
diff --git a/bin/omarchy-global-search b/bin/omarchy-global-search
index e928d42..4bc1aef 100755
--- a/bin/omarchy-global-search
+++ b/bin/omarchy-global-search
@@ -2,64 +2,6 @@
 
 MENU_JSON="${MENU_JSON:-menu.json}"
 MENU_SCRIPT="${MENU_SCRIPT:-$HOME/.local/share/omarchy/bin/omarchy-menu}"
-APPS_CACHE="${APPS_CACHE:-$HOME/.local/share/omarchy/apps.cache}"
-APPS_LIST_CACHE="${APPS_LIST_CACHE:-$HOME/.local/share/omarchy/apps.list}"
-
-# Ensure data directory exists
-mkdir -p "$(dirname "$APPS_CACHE")"
-
-# -------------------------------
-# 🔹 Fast Application Change Detection
-# -------------------------------
-get_apps_hash_fast() {
-    local app_dirs=(
-        "/usr/share/applications"
-        "/usr/local/share/applications"
-        "$HOME/.local/share/applications"
-    )
-
-    local hash_input=""
-    for dir in "${app_dirs[@]}"; do
-        if [[ -d "$dir" ]]; then
-            # Use directory modification time instead of scanning all files
-            hash_input+="$(stat -c '%Y' "$dir" 2>/dev/null || stat -f '%m' "$dir" 2>/dev/null)|"
-        fi
-    done
-
-    echo -n "$hash_input" | md5sum 2>/dev/null | cut -d' ' -f1 || echo -n "$hash_input" | md5 2>/dev/null
-}
-
-check_apps_changed() {
-    local current_hash=$(get_apps_hash_fast)
-    local stored_hash=""
-
-    [[ -f "$APPS_CACHE" ]] && stored_hash=$(cat "$APPS_CACHE" 2>/dev/null)
-
-    if [[ "$current_hash" != "$stored_hash" ]]; then
-        echo "$current_hash" >"$APPS_CACHE"
-        return 0 # Changed
-    fi
-    return 1 # Not changed
-}
-
-check_menu_changed() {
-    local menu_script="$1"
-    local cache_file="${APPS_CACHE}.menu"
-
-    [[ ! -f "$menu_script" ]] && return 1
-
-    local current_hash
-    current_hash=$(stat -c '%Y' "$menu_script" 2>/dev/null || stat -f '%m' "$menu_script" 2>/dev/null)
-
-    local stored_hash=""
-    [[ -f "$cache_file" ]] && stored_hash=$(cat "$cache_file" 2>/dev/null)
-
-    if [[ "$current_hash" != "$stored_hash" ]]; then
-        echo "$current_hash" >"$cache_file"
-        return 0 # Changed
-    fi
-    return 1 # Not changed
-}
 
 # -------------------------------
 # 🔹 Background Menu JSON Generation
@@ -199,17 +141,6 @@ ensure_menu_json() {
     local script_path="${2:-$MENU_SCRIPT}"
     local force_regen="${3:-false}"
 
-    # Quick existence check for launch speed
-    if [[ -f "$json_file" ]] && [[ "$force_regen" != "true" ]]; then
-        # Trigger background regeneration if needed, but don't wait
-        (
-            if check_apps_changed || check_menu_changed "$script_path"; then
-                generate_menu_json "$script_path" "$json_file" 2>/dev/null
-            fi
-        ) &
-        return 0
-    fi
-
     # Force regeneration (blocking)
     if [[ "$force_regen" == "true" ]] || [[ ! -f "$json_file" ]]; then
         if [[ -f "$script_path" ]]; then
@@ -222,9 +153,9 @@ ensure_menu_json() {
 }
 
 # -------------------------------
-# 🔹 Cached Application List
+# 🔹 Get Application List
 # -------------------------------
-get_applications_cached() {
+get_applications() {
     app_dirs=(
         "/usr/share/applications"
         "/usr/local/share/applications"
@@ -406,9 +337,8 @@ unified_search() {
         menu_count=${#menu_opts[@]}
     fi
 
-    # Load cached applications (fast - cached list)
     local app_opts
-    mapfile -t app_opts < <(get_applications_cached)
+    mapfile -t app_opts < <(get_applications)
     all_options+=("${app_opts[@]}")
     app_count=${#app_opts[@]}
 
@@ -486,24 +416,6 @@ unified_search() {
     fi
 }
 
-regenerate_menu() {
-    local json_file="${1:-$MENU_JSON}"
-    local script_path="${2:-$MENU_SCRIPT}"
-
-    # Clear caches
-    rm -f "$APPS_LIST_CACHE" "$APPS_CACHE" "${APPS_CACHE}.menu" 2>/dev/null
-
-    ensure_menu_json "$json_file" "$script_path" "true"
-
-    if [[ $? -eq 0 ]]; then
-        notify-send "Menu Regenerated" "Menu JSON updated"
-        return 0
-    else
-        notify-send "Error" "Failed to regenerate menu"
-        return 1
-    fi
-}
-
 # -------------------------------
 # 🔹 Main
 # -------------------------------
@@ -511,9 +423,6 @@ case "${1:-unified}" in
 unified | search | -s | --search)
     unified_search "${2:-$MENU_JSON}" "$3"
     ;;
-regenerate | regen | -r | --regenerate)
-    regenerate_menu "${2:-$MENU_JSON}" "${3:-$MENU_SCRIPT}"
-    ;;
 help | -h | --help)
     cat <<'EOF'
 Usage: $0 [command] [options]
@@ -524,7 +433,6 @@ Commands:
   help                 Show this help
 
 Optimizations:
-  - Cached application list (persistent)
   - Background menu regeneration
   - Fast directory-based change detection
   - Non-blocking startup
@@ -532,12 +440,9 @@ Optimizations:
 Environment Variables:
   MENU_JSON            Path to menu JSON file
   MENU_SCRIPT          Path to menu script
-  APPS_CACHE           Apps hash cache
-  APPS_LIST_CACHE      Cached application list
 
 Examples:
   $0                   # Fast unified search (<1s)
-  $0 regenerate        # Force cache rebuild
 EOF
     ;;
 *)
```

bin/omarchy-global-search
```diff
@@ -0,0 +1,545 @@
...
+        done < <(omarchy-font-list)
+    fi
+
+    if command -v omarchy-powerprofiles-list &>/dev/null; then
+        while IFS= read -r profile; do
+            if [[ -n "$profile" ]]; then
+                json_output+=$',\n'
+                json_output+=$(printf '    {\n      "path": "Setup > Power Profile > %s",\n      "command": "powerprofilesctl set '\''%s'\''"\n    }' \
+                    "$(echo "$profile" | sed 's/"/\\"/g')" \
+                    "$(echo "$profile" | sed 's/"/\\"/g')")
```
celsobenedetti: for each of these `echo "font" | sed 's/....'` we could use bash search/replace instead.
Here is the suggestion from shell check

```
. shellcheck: See if you can use ${variable//search/replace} instead. [SC2001]
```

this is a minor suggestion, but could improve performance

```suggestion
                    "${theme//"/\\"/}" \
                    "${theme//"/\\"/}")
            fi
        done < <(omarchy-theme-list)
    fi

    if command -v omarchy-font-list &>/dev/null; then
        while IFS= read -r font; do
            if [[ -n "$font" ]]; then
                json_output+=$',\n'
                json_output+=$(printf '    {\n      "path": "Style > Font > %s",\n      "command": "omarchy-font-set '\''%s'\''"\n    }' \
                    "${font//"/\\"/}" \
                    "${font//"/\\"/}")
            fi
        done < <(omarchy-font-list)
    fi

    if command -v omarchy-powerprofiles-list &>/dev/null; then
        while IFS= read -r profile; do
            if [[ -n "$profile" ]]; then
                json_output+=$',\n'
                json_output+=$(printf '    {\n      "path": "Setup > Power Profile > %s",\n      "command": "powerprofilesctl set '\''%s'\''"\n    }' \
                    "${profile//"/\\"/}" \
                    "${profile//"/\\"/}")
```

RushiChaganti:
Thank you for the thorough review! I've learned a lot from your suggestions, especially about using bash native string replacement over sed for better performance. I've applied all the feedback and simplified the caching system as discussed.
@celsobenedetti

RushiChaganti:
The AWK script was unintentionally preserving shell-level double quotes in the extracted commands, which led to inconsistent JSON output.

**Example:**
Before:
`"command": "present_terminal \"omarchy-setup-fido2 --remove\""`
After:
`"command": "present_terminal omarchy-setup-fido2 --remove"`

**Fixes implemented:**

1. Remove all double quotes from command strings before JSON escaping:
   `gsub(/"/, "", cmd_line)`
2. Escape only backslashes for proper JSON formatting, since quotes are no longer present to escape.

This issue surfaced while testing all menu items, where I noticed the inconsistent behavior caused by the AWK script.

celsobenedetti:
I think if this is to be merged, we need to go for a different approach.

Currently we use `awk` to build a `menu.json` based on the contents of `omarchy-menu`.
As impressive as the awk code is, it is coupled to the structure of `omarchy-menu`, so if that were to change, the `omarchy-global-search` script would break, and some poor maintainer would need to figure out how to adapt it.

I understand this approach was done to minimize changes in the existing `omarchy-menu`, however I don't think it is maintainable. 

Personally I think this `omarchy-global-search` is a better experience to the existing `omarchy-menu`, and could even replace it.
However, if we do need to maintain both the existing `omarchy-menu` and the new `omarchy-global-search`, I think we should do the following:

1. **first define a `omarchy-menu.json` as the "source of truth"**
2. use this json file for `omarchy-menu` (would require restructuring of that file) 
3. use this json file for `omarchy-global-search` (should be easy, since we already do)

**Then, to add/remove items from the menu, we modify the json file, and it would apply to both `omarchy-menu` and `omarchy-global-search`**

This would require restructuring of the `omarchy-menu`, but I think it is really the only way to go.

RushiChaganti:
@celsobenedetti  Yeah, I agree with all your points.
Would it be okay if I take a shot at working on the omarchy-menu?

celsobenedetti:
> @celsobenedetti Yeah, I agree with all your points. Would it be okay if I take a shot at working on the omarchy-menu?

Good question, since that would be a more invasive change, we should ideally get some feedback from @ryanrhughes or @dhh.

The big question is this:
- do we want to keep both `omarchy-menu` and `omarchy-global-search`, or could `omarchy-global-search` replace the existing menu?



If the decision is that both should be kept, then yes, I think this is the only way to go:
>    first define a omarchy-menu.json as the "source of truth"
>    use this json file for omarchy-menu (would require restructuring of that file)
>    use this json file for omarchy-global-search (should be easy, since we already do)


I'd say yes, go for it. Ultimately we are not sure if this is the direction the maintainers want to go in, but If we could show a concrete implementation of this working, it would help move discussion along.

I'd love to help with this project too, let know if anything pops up :)

RushiChaganti:
Previously, I experimented with a similar approach, but it wasn’t implemented well at the time. DHH commented on it here:

https://github.com/basecamp/omarchy/pull/2258#issuecomment-3375421821

Based on that feedback, DHH made it clear that the existing menu needs to be preserved. So, to move forward, I think we’ll need to modify the current omarchy-menu accordingly. It’s likely to be an invasive change, and I’m not entirely sure whether DHH would approve it — but it seems to be the logical next step if we want both systems to coexist.

I’ll keep you updated if anything comes up. Thanks for the input and collaboration!

RushiChaganti:
This update implements a unified data source for the menu and global search components.

Created menu.json as the single source of truth for navigation data.

Refactored omarchy-menu to load its structure from this JSON file.

Updated omarchy-global-search to use the same data source for consistency.

Simplified data management and eliminated redundant definitions between the two components.

This change ensures both omarchy-menu and omarchy-global-search stay synchronized and easier to maintain going forward.

ryanrhughes:
@RushiChaganti This is a really great idea but I think we have a better way to accomplish it available to us. I've documented details about this in https://github.com/basecamp/omarchy/issues/2807 if you want to take a look at that. 

If you'd be up for it, I'd love to have you take a swing at accomplishing this that way instead of this current route that really requires us to reinvent the wheel a bit.

PR #2462, opened 2025-10-15
Add Windows VM

ryanrhughes:
In looking at #1333, this proved to be a really great idea but had some flaws that needed ironing out to be compatible with the Omarchy install. 

I branched off of that PR to create this one which is a fully-formed and functioning version that keeps the core, but refines it a bit and keeps in line with many of the conventions we've put in place, while also adding some QoL things to make the whole experience a little nicer.

### Stayin' Alive
The current desktop file shortcut **does not** launch with `--keep-alive` which means that the moment you exit the RDP session, it will shutdown the Docker container. This makes booting back up take a bit longer and is less desirable in situations where you don't really care about power consumption but if you're on something like a laptop, you probably don't want the VM just chilling in the background. This may be something we play with over time or just leave to the user to adjust the default state of based on their use case.

If you run with `--keep-alive`, the container will run in the background on RDP exit which makes your next connection almost instant.

### Smart Scaling
There are only a few scaling values available in the RDP app but just defaulting to 100 feels bad. I applied a reasonable scale to it that seems to work pretty well based on testing. This allows install on a device to boot a Windows env that feels about the same vs having to adjust first thing.

### Installer
The installer has a configurator-esque feel to it and prompts for the values to use to give users control. Once you answer the questions, it's automated to the finish-line. 

You monitor the results in the browser and once you arrive in a desktop, you're free to connect with the RDP app and do what you wish.
screenshot-2025-10-15_02-27-45
screenshot-2025-10-15_02-33-40

### VM Notifications
When starting VMs, it can feel like nothing is happening if you use the launcher shortcut so a notification is thrown just to let the user know something is indeed happening.
screenshot-2025-10-15_01-44-24

### Unified Tool
It's all just one, unified tool to handle a variety of tasks.
screenshot-2025-10-15_01-36-38

### Status
Status is just a helpful status to know the current state.
screenshot-2025-10-15_01-36-10

### A note about WinBoat
WinBoat was also explored and looks like an excellent project. However, it's still pretty early and the individual app launching proved to be very poor at the moment. Really the only experience I'd suggest using in it at the moment would be the full desktop launcher. 

There is also a bug within the app itself that makes it have compatibility issues with Nvidia and require running with some special flags.

I'm sure with time, this project will surpass what we have here and possibly be a viable replacement but it's just not there currently.

### DIFF
--- a/bin/omarchy-windows-vm
+++ b/bin/omarchy-windows-vm
@@ -170,6 +170,8 @@ EOF
     exit 1
   fi
 
+  mkdir -p $HOME/Windows
+
   # Create docker-compose.yml in user config directory
   cat << EOF | tee "$COMPOSE_FILE" > /dev/null
 services:
@@ -194,7 +196,7 @@ services:
       - 3389:3389/udp
     volumes:
       - $HOME/.windows:/storage
-      # - $HOME/Documents:/Documents # Uncomment to mount local volume in Windows. It will show in file explorer under "network -> host.lan -> Documents" or \\host.lan\Documents
+      - $HOME/Windows:/shared
     restart: always
     stop_grace_period: 2m
 EOF


bin/omarchy-windows-vm
```diff
@@ -0,0 +1,450 @@
...
+  windows:
+    image: dockurr/windows
+    container_name: omarchy-windows
+    environment:
+      VERSION: "11"
+      RAM_SIZE: "$SELECTED_RAM"
+      CPU_CORES: "$SELECTED_CORES"
+      DISK_SIZE: "$SELECTED_DISK"
+      USERNAME: "$USERNAME"
+      PASSWORD: "$PASSWORD"
```
sspaeti, line 187: Don't forget `KEY` under `environement` - or how do you handle the Windows key? I added this: https://github.com/basecamp/omarchy/pull/1333/files#diff-8b66d03485362890843f0851699ad2b837d4958ac9761a3643d93d197d87163cR9 as per [DHH request](https://github.com/basecamp/omarchy/pull/1333#issuecomment-3289657795). 

This is how it looks if you add arbitrary to force input: https://github.com/basecamp/omarchy/pull/1333#issuecomment-3293645724

dhh: Turns out that it's actually kosher to start Windows without a key, which will then just delay activation to when you're doing an activity that requires that (like changing background image etc).

Issue #2314, opened 2025-10-08
All windows close when dragging between monitors


TobbEEEEH:
### System details

AMD Ryzen 7 5800X 8-core, Radeon RX 6700

### What's wrong?

If I drag any window (dolphin, terminal, browser) with SUPER + Mouse1 to the edge where my monitors "meet" all windows crash on all workspaces. It almost looks like a sigkill on everything that's open right now/reboot of hyprland.

maietta:
By chance, are you dealing with any monitors that are DisplayPort over USB?

Because I am dealing with the same issue.

TobbEEEEH:
> By chance, are you dealing with any monitors that are DisplayPort over USB?
> 
> Because I am dealing with the same issue.

I run DP -> DP on main monitor and HDMI -> HDMI on my secondary. When trying it again it seems to be browser related as Dolphin don't act up right now. It's also closing at random points, but I guess this wasn't it. I have to check further if it's specific to Brave or if any other GUI apps act up.

adelplace:
Same issue here with chrome.
And I also have the problem using the keyboards shortcuts like SUPER + MAJ + 2

khanalpride:
Same issue with Microsoft edge browser as well.

71be7:
I've also had the same issue with Microsoft Edge, Chromium, and the included ChatGPT and YouTube webapps. I've seen it triggered by keyboard shortcuts, mouse dragging windows, and resizing.

guasam:
Same issue here with Edge and Zen Browser. I had to stop moving the window to another monitor, as it would cause the entire browser to just disappear during ongoing calls. This has happened multiple times.

bruhmux:
Same here with Brave browser. Same behavior where no matter the method used to transfer workspaces to, it crashes, but only when between monitors. Same monitor workspaces work fine. I was working just fine on it 7 hours ago, only after a restart it took affect which i did more than once yesterday.
This only seems to happen with Brave(browsers) and workspaces specifically, as any other app can be moved, including terminals and electron apps like discord. When I move another app into the same workspace, it resizes appropriately, only crashing on moving the browser to another monitor.
I'm using DP for both monitors, (main) DP-1: 2560x1440@165, DP-2: 1920x1080@60

wolfgangw:
[...]

PR #2480, opened 2025-10-16
Forward all omarchy-launch-browser args to the exec cmd

jelenv:
Forward all arguments added to `omarchy-launch-browser` to the exec command.

Current implementation does not use any additional args used in the `.desktop` file for the browser. The change allows to add any arguments you may want to use to the browser exec call. 

For example this is useful to work around the issue #2314. The PR fixing it (#2394), does not fix the issue when using `google-chrome-stable`. With this change, you can fix it yourself.

---
**[Not part of the PR changes]**
To fix issue #2314 when using Google Chrome, do this:
1. Copy the .desktop file (to create a personal override): `cp /usr/share/applications/google-chrome.desktop ~/.local/share/applications/`
2. In the desktop file to all the `Exec=` lines add `--disable-features=WaylandWpColorManagerV1` right next to the `/usr/bin/google-chrome-stable` (fixes launching via the desktop entry)
3. Edit `~/.config/hypr/bindings.conf`, add `--disable-features=WaylandWpColorManagerV1` to the `$browser` variable (fixes launching via `Super + B`). Like this:
```sh
$browser = omarchy-launch-browser --disable-features=WaylandWpColorManagerV1
```

### DIFF
--- a/bin/omarchy-launch-browser
+++ b/bin/omarchy-launch-browser
@@ -10,12 +10,9 @@ else
 fi
 
 args=()
-for arg in "$@"; do
-  if [[ $arg == "--private" ]]; then
-    args+=("$private_flag")
-  else
-    args+=("$arg")
-  fi
+for arg; do
+  [[ $arg == "--private" ]] && arg=$private_flag
+  args+=("$arg")
 done
 
 exec setsid uwsm app -- "$browser_exec" "${args[@]}"


jelenv:
Possibly more elegant (less boilerplate) implementation:
```sh
# ...
args=()
for arg; do
  [[ $arg == "--private" ]] && arg=$private_flag
  args+=("$arg")
done

exec setsid uwsm app -- "$browser_exec" "${args[@]}"
```

Let me know which one is preffered.

jelenv:
Closing the PR, as I realized now that the previous code already forwarded the arguments and in a more elegant way.

PR #2486, opened 2025-10-16
 Feature turn off backlight on launch screensaver

### LINKED ISSUE #2473: turn off keyboard backlight with screensaver
### What do you need?

To save some energy in battery.
Does it work to someone already? 
I am on Thinkpad T480.

manuel1618:
fixes #2473


dhh:
Interesting idea, but I think I'd be rather surprised by it. If you leave your laptop in a location where you want backlighting for a few minutes while away, I'd find it odd if it was off when I came back. Think I'm going to skip this one.

PR #2488, opened 2025-10-16
feature: User customization export / import

joelgaff:
Omakase is the preferred option in Omarchy - it's in the name after all. But sometimes you may want to order a la carte as well, a hand roll on the side. 

Then, when you're ready to move your set up to a new machine, you want to take your customizations with you, and not have to spend all day re-configuring.  A super simple way to be up and running on a fresh (or not so fresh) Omarchy install, with the comforts of your old set up. This feature helps achieve just that.

Simply select Export from the Omarchy > Settings > Config > Export/Import menu to automatically export your customizations, including any packages (including AUR packages), and custom web apps from your old machine.

Then, drop the created tar.gz archive in the home folder on your new machine and head to Omarchy > Settings > Config > Export/Import

Follow just a few prompts and your customizations will be imported for you.

### Export Screen
image

### Import Screen
image

### What Gets Exported / Imported
- Personal Dotfiles (~/)
- Configuration Files (~/.config/)
- Web Applications (~/.local/share/applications/)
- Package Information
	- pkglist.txt - Explicitly installed pacman packages
	- aurlist.txt - AUR packages
	- all_packages.txt - Complete package list
- Documentation
	- README.txt - Import instructions

### What Gets Excluded
- Browser Profiles
	- BraveSoftware/
	- chromium/
	- google-chrome*/
	- vivaldi*/
	- microsoft-edge*/

- Large Application Data
	- Cursor/ 
	- 1Password/ 
	- obsidian/
	- Microsoft/ 
	- Insync/
	- spotify/

- Cache & Temporary Files
	- All */Cache/, */cache/ directories
	- GPU caches, shader caches
	- Log files
	- Crash reports

- Security Items 
	- SSH keys
	- GPG keys
	- Fingerprint enrollment
	- FIDO2 settings
	---
There are lots of ways this set up could be tweaked, to include more, less, or other items.  I'm open to what DHH and Ryan and the greater community feel fits best.  Or, maybe this doesn't even fit well in Omarchy.  That's ok, too!  I've found it super useful, however, as I have been moving my Omarchy set up from machine to machine.

### DIFF
--- a/bin/omarchy-export
+++ b/bin/omarchy-export
@@ -48,7 +48,7 @@ if [ -d "$HOME/.config" ]; then
     done
     cd - > /dev/null
     
-    echo "  Backed up all .config files"
+    echo "  Backed up .config files"
 fi
 
 
@@ -106,9 +106,9 @@ if [ -d "$HOME/.local/share/applications" ]; then
                 fi
             fi
         done
-        echo "  Backed up $USER_APP_COUNT user-created web apps with icons"
+        echo "  Backed up user-created web apps with icons"
     elif [ $USER_APP_COUNT -gt 0 ]; then
-        echo "  Backed up $USER_APP_COUNT user-created web apps"
+        echo "  Backed up user-created web apps"
     else
         echo "  No user-created web apps found (all are Omarchy defaults)"
     fi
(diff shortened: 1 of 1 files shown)


dhh:
I think this is a really interesting idea, and it solves a real problem. But I'd rather get folks on the path of using something like GNU Stow to manage their dotfiles. It'll ultimately be more portable and easier to track when Omarchy is making updates to it. If you'd like to explore what that could look like, please do! Thanks for suggesting this ✌️

joelgaff:
Amen, and thanks for the feedback. I've actually started using GNU stow in the time since I submitted this PR and totally see your point.

PR #2496, opened 2025-10-17
Hide "Hardware Locality lstopo" from app menu

lukehsiao:
By default, we have this Hardware Locality lstopo app in the app launcher, which is unlikely to be useful to the typical user.

This hides it along with the other apps Omarchy intentionally excludes from the menu.

Ref: https://github.com/basecamp/omarchy/issues/1631

---

image

### DIFF
--- a/applications/hidden/lstopo.desktop
+++ b/applications/hidden/lstopo.desktop
@@ -1,0 +1,2 @@
+[Desktop Entry]
+Hidden=true


dhh:
This is not a default app in a new install. Not sure where it's coming from, but I don't think it makes sense to have hiding .desktop files for apps we don't directly install.

lukehsiao:
Noting that this eventually became part of the defaults, and was re-implemented here: https://github.com/basecamp/omarchy/commit/e01741afab11296fedbba5fae33e4141bb16db74

Issue #2502, opened 2025-10-17
Ghostty can't load theme when Everforest is selected - theme not found


gonzochic:
### System details

Omarchy 3.0.2

### What's wrong?

After an Omarchy update, I get an error when opening Ghostty.

<img width="913" height="526" alt="Image" src="https://github.com/user-attachments/assets/9ad6d65a-1b16-4e3f-8ebf-5c5518a4dec7" />

I run the Everforest Theme. To resolve the issue I removed two additinal blanks within the ghostty config.

morgoth:
already fixed in https://github.com/basecamp/omarchy/pull/2471

PR #2503, opened 2025-10-17
Fix theme line formatting in ghostty.conf

gonzochic:
Fixes my issue (#2502) with ghostty in Omarchy 3.0.2 when using Everforest theme.


kyedavey:
duplicate of #2471

PR #2511, opened 2025-10-17
Fixed problematic default scaling in hyprland

Rovetown:
Updated hyprland's monitor config to default to 1080p instead of 2k to fix general issues on low-res displays and inside Virtual Machines like VirtualBox.

The current config shipped with the iso makes the hyprland windows scale to 2k instead of 1080p, which results in broken scaling inside a terminal or application, which prevents alot of functionality on non-1080p displays and especially Virtual Machines like VirtualBox, where you have to trick around to get a working display resolution to install GuestBox Additions.

This is a fix for https://github.com/basecamp/omarchy/issues/2144 and also additionally resolves a temp workaround posted in the discussion: https://github.com/basecamp/omarchy/discussions/176

### DIFF
--- a/config/hypr/monitors.conf
+++ b/config/hypr/monitors.conf
@@ -3,17 +3,17 @@
 # Format: monitor = [port], resolution, position, scale
 # You must relaunch Hyprland after changing any envs (use Super+Esc, then Relaunch)
 
-# Optimized for retina-class 2x displays, like 13" 2.8K, 27" 5K, 32" 6K.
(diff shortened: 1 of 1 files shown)


dhh:
Appreciate the suggestion, but the settings are the way they are because I wouldn't want to run Omarchy on anything less than a 2x display.

Happy to entertain a path that detects low-resolution displays and does something accordingly. But can't take this as-is.

Issue #2295, opened 2025-10-07
Webapp remove gives "You must provide web app names" when trying to remove any webapp


lavavex:
### System details

Intel i7-6600u, Thinkpad T470s

### What's wrong?

omarchy-webapp-remove script will not allow me to remove any webapps, complaining about not providing webapp names.

zhydzhy:
At first I thought it was a bug as well, but you need to use "x" key to toggle the apps you want to remove and then hit "enter"

lavavex:
> At first I thought it was a bug as well, but you need to use "x" key to toggle the apps you want to remove and then hit "enter"

Ah, I see... so it's a "RTFM" issue and not a bug, cool. It would be nice to see that a bit more front and center then it currently is as most people (I think) would assume highlighting and pressing enter would select. Maybe adding an empty checkbox next to each option to make it more clear you need to select one or multiple before continuing would help... I'll take a look at the code later and see if it's easy to implement or at least add a more verbose error for not choosing one...

kroaton:
First time user here. I had the exact same experience. Maybe in cases where the user presses Enter and nothing was selected, have a message that says "You need to select a webapp using Space to remove it. Multiple can be selected at once."

PR #2516, opened 2025-10-18
Clarify error message when no web apps have been selected in the WebApp Remove menu

ZorudaRinku:
Resolves slightly obscure error when user doesn't see the small `x toggle` text (#2295)


brennancoslett:
I also was confused by the existing error message and I think this text would have helped me figure it out without the google.

PR #2518, opened 2025-10-18
Fix theming for Brave and add theming support for helium

CantC0unt:
Theming was broken for Brave and some* browsers

renamed `rgb` to `THEME_RGB_COLOR` since `rgb` was renamed in an earlier commit which broke this functionality.

copied brave theming code to make theming possible for helium


CantC0unt:
alternatively: calling `echo "{\"BrowserThemeColor\": \"$THEME_HEX_COLOR\"}" | tee "/etc/chromium/policies/managed/color.json" >/dev/null` unconditionally might fix theeming for more chromium based browsers

PR #2525, opened 2025-10-18
Restore normal Caps Lock behaviour by removing default compose:caps

QuarkOnRails:
Decided to remove default compose:caps because of 2 reasons:

1. **Consistency:** Mismatch between log in screen and system. Caps Lock works normally on log in screen, then switches to compose inside of the system.

2. **User experience:** As DHH said on X: "Omarchy wasn't built to win over existing linux users, It was built to grow the pie. To attract people on Windows and Mac who love computers to the joy of Hyprland..." 

Caps Lock on Win and Mac does not work compose by default, This can be confusing for users coming from those systems, while it’s easy for experienced Linux users to re-enable if they prefer.

### DIFF
--- a/config/hypr/input.conf
+++ b/config/hypr/input.conf
@@ -3,7 +3,7 @@
 input {
   # Use multiple keyboard layouts and switch between them with Left Alt + Right Alt
   # kb_layout = us,dk,eu
-  kb_options = compose:caps # ,grp:alts_toggle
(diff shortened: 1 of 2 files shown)


dhh:
Thanks for the suggestion, but I really like caps lock as the modifier. Never used the key in normal use, and it's perfectly placed to be the compose key, which I use constantly. Feel free to change it on your own setup, though!

PR #2529, opened 2025-10-18
Support environment variables in browser desktop files

jelenv:
The launch scripts `omarchy-launch-browser` and `omarchy-launch-webapp` now extract and forward env variables in browser desktop files correctly.

Fixes launching Chrome with custom hardware acceleration flags and similar configurations.

For example this configuration in `google-chrome.desktop` file now correctly works:
```sh
Exec=env LIBVA_DRIVER_NAME=radeonsi /usr/bin/google-chrome-stable --disable-features=WaylandWpColorManagerV1 %U
```

This is often needed on hybrid GPU laptops.

### DIFF
--- a/bin/omarchy-launch-browser
+++ b/bin/omarchy-launch-browser
@@ -1,27 +1,22 @@
 #!/bin/bash
 
 default_browser=$(xdg-settings get default-web-browser)
-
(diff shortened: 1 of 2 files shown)


dhh:
I prefer to see chromium-flags.conf used for this stuff. Appreciate the suggestion, though ✌️

PR #2532, opened 2025-10-18
clicking the copy-url  extension icon to copy the URL

### LINKED ISSUE #2493: Copy URL to clipboard enhancement to also work on click
### What do you need?

This PR is great https://github.com/basecamp/omarchy/pull/1458 and is primarily aimed at web apps with no location bar. Right now the only way to use it is via keyboard shortcut. Even though the icon shows up in the normal chrome browser, clicking on it doesn't do anything and the "copy url" action is grayed out. I've made a [small modification](https://github.com/davidguttman/omarchy/commit/3c6555f415d8f733956c8ddeed624b5360f2de26) to my local version to work on click and can open a PR if there's interest.

0xRichardH:
Closes: https://github.com/basecamp/omarchy/issues/2493

* Added support for copying the current tab's URL when the extension icon is clicked 
* Modified notifications to auto-close after 4 seconds, preventing them from lingering indefinitely


dhh:
Getting this error when I click the icon:

image

0xRichardH:
The error occurs because the extension tries to access chrome:// URLs, which Chrome extensions cannot access for security reasons.

This change introduces a check to prevent the extension from attempting to copy these restricted URLs.

PR #2533, opened 2025-10-18
Add window blur effect

vyrx-dev:
# ✨ Window Blur Effect

- Added window blur for a cleaner and more polished look than plain transparency.
image


dhh:
Fixed via cc2523ff8994 🤘

PR #2535, opened 2025-10-18
Fix invalid terminal working directory

ezradyck:
If a process doesn't have a valid working directory, for example when
launching a terminal when Firefox is the active window, this command was
passing an invalid working directory to the terminal, causing an error
to be shown in Alacritty.

Check to see if the symlink exists and is valid before reading it, and
default to $HOME if it doesn't.


ezradyck:
Looks like this was already fixed in dev, but hasn't been merged into master yet.

PR #2537, opened 2025-10-18
Add elephant config to support !bangs in walker

CantC0unt:
This adds a `websearch.toml` file to change the default websearch query to use `unduck`

This keeps the `default google search` but allows the option to use [`!bangs`](https://raw.githubusercontent.com/T3-Content/unduck/refs/heads/main/src/bang.ts) like in duckduckgo but without the server-side rerouting  (aka its ~~blazingly fast~~ faster than duckduckgo, this should be cached in browser on first call, but it doesn't looks like that is happening)

2025-10-1902-58-35-ezgif com-video-to-gif-converter

~~**Caveat**: This requires the config to be placed in `.config/elephant/` instead of `.config/omarchy/elephant/`, The existing `calc.toml` and `desktopapplications.toml` configs also probably require this.~~ (Fixed in 3.1.0 migration)

### DIFF
--- a/config/elephant/websearch.toml
+++ b/config/elephant/websearch.toml
@@ -1,0 +1,4 @@
+[[entries]]
+default = true
+name = "Web Search"
+url = "https://unduck.link?q=%TERM%"


CantC0unt:
@dhh thoughts on this ?

dhh:
I think it's a clever idea, but don't like the concept of routing everything through this domain. It's a nice idea for someone to set on their own though ✌️

CantC0unt:
> I think it's a clever idea, but don't like the concept of routing everything through this domain. It's a nice idea for someone to set on their own though ✌️

technically its possible to have this running fully locally (or even integrated properly with elephant/walker) but I don't have the time of the day to work on this for the next few months at least. 

Its also just a simple JS script that's getting cached along with the map for bangs. I understand the concern with the domain, but if that's the only concern then its also possible to host it locally or on Omarchy domain. Do [check it out](https://github.com/T3-Content/unduck) if you think its worth it.

Issue #1256, opened 2025-08-28
Feature Request: Hypridle different timeout actions for battery vs AC power


TomSchimana:
Hypridle should support different actions based on power source. On battery, screensavers drain power unnecessarily. Just turning off the display would be better. When plugged in, screensavers are fine.

Example:
```
listener {
    timeout = 150                                             # 2.5min
    on-timeout = sh -c 'if cat /sys/class/power_supply/BAT*/status 2>/dev/null | grep -q "Discharging"; then hyprctl dispatch dpms off; else pidof hyprlock || omarchy-launch-screensaver; fi'
    on-timeout = pidof hyprlock || omarchy-launch-screensav##er # start screensaver (if we haven't locked already)
}
```

dhh:
I think that idea. Should be extracted into a omarchy-* bin. Please do explore a PR.

PR #2573, opened 2025-10-19
feat: disabled screensaver in power-saver mode to save battery

dakshesh14:
## Description:

Disabled screensaver in power-saver mode to save battery by:
1. Early exit in `omarchy-launch-screensaver` when in `power-saver` mode.
2. Hiding the screensaver option from the "System" menu.

Demo:

https://github.com/user-attachments/assets/d17b8398-e99b-4f5a-aafc-1fe534273f0e

Addresses: #1256

### DIFF
--- a/bin/omarchy-launch-screensaver
+++ b/bin/omarchy-launch-screensaver
@@ -5,6 +5,11 @@ if ! command -v tte &>/dev/null; then
   exit 1
 fi
 
+# Exit early if we are using power-saver
+if [[ "$(powerprofilesctl get)" == "power-saver" ]]; then 
+  exit 1
+fi
+
 # Exit early if screensave is already running
 pgrep -f "alacritty --class Screensaver" && exit 0
 
--- a/bin/omarchy-menu
+++ b/bin/omarchy-menu
@@ -451,7 +451,15 @@ show_update_password_menu() {
 }
 
 show_system_menu() {
-  case $(menu "System" "  Lock\n󱄄  Screensaver\n󰤄  Suspend\n󰜉  Restart\n󰐥  Shutdown") in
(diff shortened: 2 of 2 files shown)


dakshesh14:
Hi team, One improvement I can think of for this PR is making a reusable function to check if we are in power-saver mode.

For example, in `~/.local/share/omarchy/bin/omarchy-utils.sh`:

```bash
#!/bin/bash

is_power_saver() {
  [[ "$(powerprofilesctl get)" == "power-saver" ]]
}
```

Then source it in the scripts and use is_power_saver instead of repeating the check. As I’m new to both Omarchy and Bash scripting, I’d appreciate feedback on whether this approach makes sense.

dhh:
That's an interesting idea, but I think it's a little too clever. We have a toggle for screensaver on/off already. I don't think it's a common enough consideration/case to carry the extra code. But appreciate the suggestion ✌️.

What I would consider, perhaps, is a hook for setting the powerprofile, like we have with setting the theme. Then it would be easy for you and others to add behavior like this after the fact.

dakshesh14:
Thanks for the review. Agreed — this might be an over-engineered way to handle that issue [#1256](https://github.com/basecamp/omarchy/issues/1256).

I’ll look into adding a hook for power profile changes instead. Just to confirm - this hook would be a user-modifiable script that runs when the power profile changes, allowing for custom behavior like disabling the screensaver, right?

Also, one thing I didn’t like about this PR is that it made the screensaver depend on the power profile directly. A hook-based approach should avoid that coupling and keep the features independent.

dhh:
Yes, exactly. See the new omarchy-hook feature for details.

PR #2574, opened 2025-10-19
You need 2 to make fullscreen

hyprhex:
1 is just for the width of the screen to make a full window, so you need to set 2. If not working, you need to remove or comment out the same line in the old `tiling.conf` file


ludo237:
Are you sure, it seems full screen to me with `0`

hyprhex:
I was using 2 on my old config when I did the update, not working, and I set it to 2 again and removed the old one.

dhh:
We have both full width (`Super + Alt + F`) and full screen (`Super + F`). They're both working fine here.

hyprhex:
I think with all people because for me was not working and when you validate that it's work with all I think the problem was my custome config because I add it in the first version of Omarchy

PR #2590, opened 2025-10-19
Edit config directly from terminal with one word

hyprhex:
You edit config with nvim, so the fastest way is to use one word to open your file without the need to go throw the menu.


dhh:
Interesting idea, but I think this is too niche to have permanent aliases for. Good to add to your own .bashrc if it's something you use constantly, though.

hyprhex:
THank you, Alwayse lookin to help Omarchy users

PR #2598, opened 2025-10-20
Add binding for tiled full screen toggle

brianblakely:
This is niche, but you might like it. Binds Super + Ctrl + F to force-tile a fullscreen app. It works nice for things including running games in a tile or watching YouTube in a pip tile without needing to configure these apps specifically to do so.

image


dhh:
Love it!! Didn't know this existed 😍

rez1coder:
Freaking awesome. Thanks for sharing @brianblakely

PR #2607, opened 2025-10-20
Fix ls when not using eza

wiehann:
The `ls` command results in an error `bash: command not found: eza`, instead of it reverting to the builtin `ls` if the `eza` package is removed.

The `ls` aliases are now only set if the `eza` package is installed.


dhh:
Should also use omarchy-cmd-present.

wiehann:
Resolved.

PR #2609, opened 2025-10-20
Fix cd command if zoxide is missing

wiehann:
Some `cd ` commands like `cd non-exisiting-directory` results in an error `bash: command not found: z`, it should revert to the builtin `cd` behavior if the `z` (zoxide) package is removed.

The `cd` aliases are now only set up if the `z` (zoxide) package is installed.


dhh:
You can use omarchy-cmd-present for a cleaner if here.

wiehann:
Done, thank you.

PR #2621, opened 2025-10-20
Surface: enable built-in keyboard at LUKS

brennancoslett:
Derives the list of required modules for the installer from Chris McLeod's manual installation steps. His instructions say to use the surface-linux kernel however the default kernel has the right set of modules already (as noted by the fact that the right modules are loaded for the keyboard to work during the install and after the LUKS unlock screen).

I only have a Surface Laptop 3 to test this on but based on the guide where Chris is using a newer laptop and the only difference is the pinctrl module which I attempt to autodetect I think it should "just" work for others but definitely "use at your own risk".

### DIFF
--- a/install/config/all.sh
+++ b/install/config/all.sh
@@ -30,3 +30,4 @@ run_logged $OMARCHY_INSTALL/config/hardware/fix-bcm43xx.sh
 run_logged $OMARCHY_INSTALL/config/hardware/fix-apple-spi-keyboard.sh
 run_logged $OMARCHY_INSTALL/config/hardware/fix-apple-suspend-nvme.sh
 run_logged $OMARCHY_INSTALL/config/hardware/fix-apple-t2.sh
+run_logged $OMARCHY_INSTALL/config/hardware/fix-surface-keyboard.sh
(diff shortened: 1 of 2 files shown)


brennancoslett:
@dhh @ryanrhughes sorry if directly tagging is the wrong way to get y'all's attention. I believe this will get LUKS decryption working for intel-based Surface devices without requiring the custom kernel other folks have been using. Hopefully I got all the testing context in the commit message and PR already, I tested a fresh install from the offline ISO with this fix and it just works.

plumpkid:
Hey @brennancoslett I've got the same model of laptop. This thread has been closed but I'm on the latest update and my in-built keyboard still doesn't work to decrypt at startup :')

brennancoslett:
> Hey @brennancoslett I've got the same model of laptop. This thread has been closed but I'm on the latest update and my in-built keyboard still doesn't work to decrypt at startup :')

Sorry on slow reply, this message got lost in my email inbox...

Did you already have omarchy installed and do the update on a live system?
The script I added only gets run by default one a new install via the ISO. There are instructions on [this thread](https://discordapp.com/channels/1390012484194275541/1425250295209791549/1449551403155456124) on the discord on how to generate the mkinicpio conf file and regenerate your initramfs with the dependency that someone has already tried out and solved their problem.

From discord for folks that don't want to deal with it:
```
If upgrading Omarchy on an existing install:

1) Check if the file exists
   ls /etc/mkinitcpio.conf.d/surface_device_modules.conf

2) If it does NOT exist, generate it
   curl -fsSL https://raw.githubusercontent.com/basecamp/omarchy/master/install/config/hardware/fix-surface-keyboard.sh | sudo bash

3) Rebuild initramfs so the change applies
   sudo mkinitcpio -P

4) Reboot
   sudo reboot
```

PR #2626, opened 2025-10-20
Search obsidian vault in directories

Hieu-Truong-6219:
Uses find rather than regex for searching for the .obsidian file.


jheuing:
Solves also my issue with a deeper nesting structure. Had a hickup with a space in the directory name, but just renamed it, which was no issue.

dhh:
We now extract from `~/.config/obsidian/obsidian.json`, so this is no longer needed.

PR #2631, opened 2025-10-20
fix: use gemini-cli-bin for Gemini installation

vyrx-dev:
#### Switched package installation from gemini-cli to gemini-cli-bin.
- The extra repo version is outdated.
- Even though Omarchy menu shows it's installing from AUR, it actually pulls from the extra repo.
image


dhh:
Where are you finding this `gemini-cli-bin` package? It's not on AUR or Arch repos.

PR #2637, opened 2025-10-20
Add user confirmation before updating Omarchy

ksalk:
**Added user confirmation prompt before updating Omarchy.**

Currently clicking on Omarchy update icon in Waybar causes update to start instantly. To prevent missclicks and starting update when not intended, simple confirmation prompt was added.

image


killeik:
Seems to be not relevant after this commit  https://github.com/basecamp/omarchy/commit/b75e70fb9b5a561c5880afe87f3c7ffdb1f2dc01

ksalk:
Seems to be not relevant after this commit b75e70fb9b5a561c5880afe87f3c7ffdb1f2dc01

PR #2644, opened 2025-10-20
Add optional extra install for hyprscrolling

dhh:
Experiment with optional extras, like hyprscrolling.


ofrades:
Can we include this extra @dhh ?

dhh:
Vaxery is working on a native version of this for full inclusion.

PR #2648, opened 2025-10-21
TTE go fullscreen: --no-eol (tte 0.12.2/0.13.0)

wolfgangw:
Add new Terminal Text Effects (https://github.com/ChrisBuilds/terminaltexteffects) option "--no-eol" and set "--canvas-height 0" to use the full screen.

"--no-eol" suppresses the trailing newline emitted when an effect animation completes.
"--canvas-height 0" lets TTE match terminal height.

Current TTE version in Arch packages is 0.12.0-1, I believe. The new option will be available in 0.12.2/0.13.0.


dhh:
Awesome. Please ping me on this when the package is available and we can merge 👌

PR #2660, opened 2025-10-21
chore: add powerprofile hook for custom behavior

dakshesh14:
## Description:

Implemented a script to set the power profile, enabling users to hook into power mode changes for custom behavior. This aims to address issues like #1256 and follows up on discussions from PR #2573.

### DIFF
--- a/bin/omarchy-menu
+++ b/bin/omarchy-menu
@@ -205,7 +205,7 @@ show_setup_power_menu() {
   if [[ "$profile" == "CNCLD" || -z "$profile" ]]; then
     back_to show_setup_menu
   else
-    powerprofilesctl set "$profile"
+    omarchy-powerprofile-set "$profile"
   fi
 }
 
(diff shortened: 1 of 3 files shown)


dakshesh14:
Raised this as a continuation of the discussion in PR #2573. I couldn’t fully test `omarchy-hook` on my machine — the hook itself didn’t execute as expected. However, I verified that the power profile is being set correctly, so the changes in this PR are functioning as intended in that regard.

I may explore `omarchy-manual` further to debug the hook, but based on my testing, this PR should work correctly since the power profile setup is verified.

dakshesh14:
Hey @ryanrhughes, could you review this PR when you get a chance?

dhh:
I think this is too much code for the value it provides, to be honest. If you need custom behavior, you're better off doing your own script, then using the new menu.sh overwrite for that function. Appreciate you working on this though!

dakshesh14:
Fair point. Thanks for reviewing!

PR #2664, opened 2025-10-21
elephant: Remove elephant-files from base packages

defer:
The functionality is interesting, but the implementation is too resource-
intensive. Several users have reported issues such as Chrome becoming
unresponsive or general system slowness related to I/O or memory usage.

The files provider currently indexes all non-hidden, non-VCS directories in the home directory. [1]

It also maintains a map of md5 -> (path, changed_time). [2]

On my system:

    $ fd . ~/ --ignore-vcs --type directory | wc -l
    698078

The default inotify watch limit in Omarchy is 524,288. Any system with more
dirs than that will run out of available inotify watches for other
applications, leading to performance degradation.

Additionally, the map alone would consume approximately 6 GB of memory in my
case.

Another expensive operation is `stat`-ing every file and directory in the
filesystem whenever anything is removed. [3]

[1] https://github.com/abenz1267/elephant/blob/v2.3.2/internal/providers/files/setup.go#L59  
[2] https://github.com/abenz1267/elephant/blob/v2.3.2/internal/providers/files/setup.go#L133  
[3] https://github.com/abenz1267/elephant/blob/v2.3.2/internal/providers/files/setup.go#L87

### DIFF
--- a/install/omarchy-base.packages
+++ b/install/omarchy-base.packages
@@ -27,7 +27,6 @@ elephant-bluetooth
 elephant-calc
 elephant-clipboard
 elephant-desktopapplications
-elephant-files
 elephant-menus
 elephant-providerlist
 elephant-runner
--- a/migrations/1761044056.sh
+++ b/migrations/1761044056.sh
@@ -1,0 +1,5 @@
+echo "Drop elephant-files"
+
+if omarchy-pkg-present elephant-files; then
+  omarchy-pkg-drop elephant-files
+fi


abenz1267:
Watchers will only be put on directories though not files. Additionally you can use `.ignore` files.

The in-memory map is bad ngl.

defer:
Thanks @abenz1267, I'll update the description. In my case I work with very large source trees, for which I use fd, so ignoring it would not be a good option.

Would you be open to having exclude paths in the file provider configs? I could work with that, would be similar to how other crawlers work.

Also, would there be a way to avoid the stat syscall on every file in the map? Maybe by keeping metadata on whether it's a dir and checking prefixes on the map.

abenz1267:
> Thanks @abenz1267, I'll update the description. In my case I work with very large source trees, for which I use fd, so ignoring it would not be a good option.

I see.

> Would you be open to having exclude paths in the file provider configs? I could work with that, would be similar to how other crawlers work.

Will add that as a workaround. The proper solution would be to store the data on disk and not in memory.

> Also, would there be a way to avoid the stat syscall on every file in the map? Maybe by keeping metadata on whether it's a dir and checking prefixes on the map.

I have no idea why i did this. I'll now just check for prefix and delete those lol.

defer:
> > Thanks @abenz1267, I'll update the description. In my case I work with very large source trees, for which I use fd, so ignoring it would not be a good option.
> 
> I see.
> 
> > Would you be open to having exclude paths in the file provider configs? I could work with that, would be similar to how other crawlers work.
> 
> Will add that as a workaround. The proper solution would be to store the data on disk and not in memory.

That would solve the memory issues which is very appreciated, something like sqlite would be great for this. However, it does not solve the inotify limit thing. I wonder how tracker/gnome handles this since inotify watches can't be recursive.

> 
> > Also, would there be a way to avoid the stat syscall on every file in the map? Maybe by keeping metadata on whether it's a dir and checking prefixes on the map.
> 
> I have no idea why i did this. I'll now just check for prefix and delete those lol.

abenz1267:
Can you please update to elephant 2.4.0 and give it a try?

davidguttman:
I ran into issues with cpu spikes. I wound up adding big directories to .ignore as a workaround. Wrote about it here https://discord.com/channels/1390012484194275541/1429826740741472341/1429826740741472341

Is omarchy using the file indexing of elephant for anything?

eugenialoli:
Some more info about the slowness here: https://github.com/basecamp/omarchy/issues/2571
Is there a workaround?

defer:
Closing this, @abenz1267 folder ignore configs works around the issue. However @abenz1267, people might be affected by inotify exhaustion without knowing about it. Would you be willing to write / accept a patch for handling watcher add failures (they're currently ignored) and notifying the user that the number of watches exceeds the limit?

abenz1267:
> Closing this, @abenz1267 folder ignore configs works around the issue. However @abenz1267, people might be affected by inotify exhaustion without knowing about it. Would you be willing to write / accept a patch for handling watcher add failures (they're currently ignored) and notifying the user that the number of watches exceeds the limit?

ofc! pr's are always welcome. although i'd really like to know how to handle this better.. maybe there's a better concept i'm missing atm.

eugenialoli:
Ok, but how do you use .ignore files? for someone who is stumbling on this bug report, what's the step by step work to make it go around the problem?

defer:
@eugenialoli elephant uses 'fd' to find folders for indexing. fd will not explore folders containing a .ignore file.

However, this is not a good solution because it prevent fd, fzf, ripgrep from working on those folders as well, so if you use those for your own purposes you'll be breaking your workflow.

The correct solution is to use the new files config option to ignore specific dirs (or the whole home) or `pacman -R elephant-files`

abenz1267:
> @eugenialoli elephant uses 'fd' to find folders for indexing. fd will not explore folders containing a .ignore file.
> 
> However, this is not a good solution because it prevent fd, fzf, ripgrep from working on those folders as well, so if you use those for your own purposes you'll be breaking your workflow.
> 
> The correct solution is to use the new files config option to ignore specific dirs (or the whole home) or `pacman -R elephant-files`

... did the new non-in-memory version no help without ignoring?

defer:
@abenz1267 it sure helps on memory usage. However, in my case I still need to ignore my largest folders because they change a lot during my workflow (compiling full operating systems / aosp) and elephant goes crazy handling inotify updates. That's, however, nothing particularly wrong with elephant, I had to do this in other crawlers (i.e. tracker as well).

Tracker does index the content of files though that's why it crawls the filesystem. In this case elephant is mostly keeping a stored version of the filesystem metadata. If it's now persisted it might be worth it to skip it and just use the filesystem?. I'm thinking something similar to typing `fzf` in the command line. Results are streamed incrementally and filtered as the user types. Would surely move the time cost of searching up.

abenz1267:
> @abenz1267 it sure helps on memory usage. However, in my case I still need to ignore my largest folders because they change a lot during my workflow (compiling full operating systems / aosp) and elephant goes crazy handling inotify updates. That's, however, nothing particularly wrong with elephant, I had to do this in other crawlers (i.e. tracker as well).
> 
> Tracker does index the content of files though that's why it crawls the filesystem. In this case elephant is mostly keeping a stored version of the filesystem metadata. If it's now persisted it might be worth it to skip it and just use the filesystem?. I'm thinking something similar to typing `fzf` in the command line. Results are streamed incrementally and filtered as the user types. Would surely move the time cost of searching up.

elephant does use the metadata to put freshly changed files on top without having to search.

abenz1267:
If it goes crazy because of the inotify updates, i might look into just batching them, same as being done when deleting => batches and handles it just every 2 seconds.

sounds like that'd solve the issue.

eugenialoli:
I just updated the system and it REINSTALLED elephant-files. I need to find a way to ban it?

I mean, removing it made things faster, but it didn't completely fixed my problems. Chromium and loading big apps is still quite slow. Just not as bad as they used to be.

abenz1267:
with elephant 2.4.1 all issues should be resolved.

that said: i haven't noticed any slowdown on any of my machines before-hand that were related to elephant, but i don't use omarchy.  With 2.4.1 everything is smoother anyways.

PR #2672, opened 2025-10-21
Rename gruvbox background filename

chrisroundhill:
Gruvbox backround image filename is named 1-grubox.jpg, renaming to 1-gruvbox.jpg


dhh:
This would break existing usage. So would need a migration. Don't think that's worth it. Nobody looks at these anyway. Appreciate the suggestion, though!

PR #2677, opened 2025-10-21
Update theme-set-vscode to support cursor and codium

samrudhnelli:
Updated theme-set-vscode to allow theme-changing in VSCode, VSCodium and Cursor without unnecessary extra files.


chaddestchad67:
@SamrudhNelli Aww is someone upset by free speech? 😂😂😂😂😂😂😂😂😂😂

basestchad69:
https://github.com/user-attachments/assets/a70f12a2-fbfc-4216-a695-df6622d113af

samrudhnelli:
@dhh added theme changing support for codium (which was previously the default editor and open-source version of code). We could add another file for omarchy-theme-set-codium, but it clutters the bin folder, so i merged those files into one.

dhh:
I've added vscodium already on the current setup, which I think is fine, so we're good 👍

PR #2679, opened 2025-10-21
chore: better contrast for kanagawa kitty tab colors

shelldandy:
Provides better contrast for tabs inside of kitty in kanagawa.

After
image

Before

image


shelldandy:
@dhh can i get a review on this one please?

shelldandy:
@dhh are 22 days enough to get a review?

dhh:
We now auto generate these, so feel free to resubmit for the new setup. I can't actually fully tell the difference, though?

PR #2682, opened 2025-10-21
feat:make bluetooth wifi and audio icons in waybar toggle visibility instead of opening new window

theajmalrazaq:
#### Previous Behavior
Clicking on the Bluetooth WiFi or Audio icons in Waybar always opened a new window instead of toggling their visibility  

#### New Behavior
Now clicking the icons toggles their visibility  
If an icon section is active it hides when clicked again  
If hidden it becomes visible  
This improves usability and prevents multiple unnecessary windows from opening  

#### Fix Summary
- Added toggle functionality for Bluetooth WiFi and Audio icons  
- Removed old logic that always opened new windows  
- Improved icon state handling for a smoother experience

### DIFF
--- a/config/waybar/config.jsonc
+++ b/config/waybar/config.jsonc
@@ -96,9 +96,9 @@
     }
   },
   "bluetooth": {
-    "format": "",
+    "format": "",
     "format-disabled": "󰂲",
-    "format-connected": "",
+    "format-connected": "",
     "tooltip-format": "Devices connected: {num_connections}",
     "on-click": "omarchy-launch-bluetooth"
   },


CantC0unt:
I don't think this should be the default behavior ever. 

If I ever click the WIFI icon, its much more likely I do it to change what WIFI I'm connected to, or connect to one if I'm not connected to any than it is that i did it to toggle it off. same goes for Bluetooth as well, If I ever click that it's much more likely I did that to connect or disconnect from a BT device than it is that I wanted toggle BT off.

dhh:
Interesting idea, but I don't think it's worth the extra code. You have to learn `Super + W` to be able to use Omarchy effectively. Thanks for the suggestion, though!

PR #2686, opened 2025-10-21
feat(defaults): easily set defaults/install additional browsers

LukasParke:
This PR adds a set of helper scripts to set defaults across all the needed places and install a couple browsers and properly set the default values needed for them to be used elsewhere when opening links.

Makes OAuth login much smoother


basestchad69:
https://github.com/user-attachments/assets/a70f12a2-fbfc-4216-a695-df6622d113af

LukasParke:
@dhh Any thoughts on this PR? 

Happy to make changes if you have any in mind.

I made these edits as these were things I ran through manually to get OAuth login flows to properly use the correct browser for cached sessions.

LukasParke:
Went ahead and swapped it to target Dev, and synced it.

LukasParke:
Synced the branch

dhh:
Appreciate your work on this, but I think it's too much. Browsers installed via Install > Package/AUR already offer to set themselves as default when you start them. I don't think we need to make this easier than what's already there. But thanks for exploring it!

LukasParke:
> Appreciate your work on this, but I think it's too much. Browsers installed via Install > Package/AUR already offer to set themselves as default when you start them. I don't think we need to make this easier than what's already there. But thanks for exploring it!

Hey @dhh,

They do but that functionality doesn't actually extend system wide in all situations, when doing things like logging into the slack desktop client it still attempts to open the OAuth login link in the normal chromium instance. 

That was really the driver behind this entire effort for me, as different apps use different open tools.

PR #2692, opened 2025-10-22
Adds nautilus backspace-nav extension

EstebanForge:
From here: https://github.com/EstebanForge/nautilus-backspace-nav

A Nautilus extension that enables (restores) backspace key navigation to go up one directory level, similar to other popular file managers.

- Enables Nautilus Python extensions support
- Required for Nautilus extensions like backspace navigation
- Added alongside other Python packages in base installation

Used for months on Fedora, latest Gnome, without issues. Now using it on Omarchy, without issues.
To provide some sanity for those moving out from Windows into Omarchy.

### DIFF
--- a/config/nautilus-extensions/backspace-nav.py
+++ b/config/nautilus-extensions/backspace-nav.py
@@ -1,0 +1,143 @@
+#!/usr/bin/env python
+
+# Author: Esteban Cuevas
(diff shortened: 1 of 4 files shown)


dhh:
Appreciate the suggestion, but I don't think this is worth the weight on the default installation.

PR #2694, opened 2025-10-22
shellcheck: fix various issues

neheb:


chaddestchad67:
https://github.com/user-attachments/assets/ad288fde-e35d-4ae1-94e1-7b299d98b175

dhh:
You can't just change a bunch of random stuff with no description or justification.

PR #2696, opened 2025-10-22
Avoid migrations issues as experienced in #2546

### LINKED ISSUE #2546: Walker and Elephant missing after system upgrade
### System details

Omarchy v3.1.0

### What's wrong?

After updating Omarchy through the update icon option in the topbar, walker and elephant were completely missing. I attempted to re-install these packages through the omarchy repo, and did so successfully. However, the walker menu and the system menu in the top left do not appear when I press `SUPER SPACE` and press the top left icon, respectively.

edit:
Installing `walker` instead of `walker-bin` using yay allows me to open the system menu and regular menu using the key bindings, but they are now transparent and what I assume not their default appeareance.

edit 2:
After doing the above and rebooting, the menus now seem to be fine. But why did this happen in the first place?

```
~ ✗ elephant v
2025/10/19 13:35:25 INFO elephant config="using default config"
2.2.3


~ ❯ walker -v
Running Service: true
0.13.26

~ ❯ cat /etc/pacman.conf
# See the pacman.conf(5) manpage for option and repository directives

[options]
Color
ILoveCandy
VerbosePkgLists
HoldPkg = pacman glibc
Architecture = auto
CheckSpace
ParallelDownloads = 5
DownloadUser = alpm

# By default, pacman accepts packages signed by keys that its local keyring
# trusts (see pacman-key and its man page), as well as unsigned packages.
SigLevel = Required DatabaseOptional
LocalFileSigLevel = Optional

# pacman searches repositories in the order defined here
[core]
Include = /etc/pacman.d/mirrorlist

[extra]
Include = /etc/pacman.d/mirrorlist

[multilib]
Include = /etc/pacman.d/mirrorlist

[omarchy]
SigLevel = Optional TrustAll
Server = https://pkgs.omarchy.org/$arch
```

NicolasDorier:
I tried to run `omarchy-update`, then elephant was not found

I noticed I could not install elephant manually on pacman without `-Sy`.
Without this, installing 3.1.1 was giving me errors about file not found on the repository.

This may fix #2546

### DIFF
--- a/bin/omarchy-pkg-add
+++ b/bin/omarchy-pkg-add
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 if omarchy-pkg-missing "$@"; then
-  sudo pacman -S --noconfirm --needed "$@" || exit 1
+  sudo pacman -Sy --noconfirm --needed "$@" || exit 1
 fi
 
 for pkg in "$@"; do


dhh:
`-Sy` is considered dangerous because it only updates the dbs, not the whole system. This may give you an inconsistent system state in a rolling distro. 

The omarchy-update already runs `pacman -Syu` before running migrations. If something isn't working with that, we need to fix it there,

PR #2697, opened 2025-10-22
Feat/enable vim style tiling

glauberdmo:
This PR requests to add  one commit:
_**feat: add support to vim style tiling**_ 

this config enables:
- SUPER + HJKL to focus on tiles
- SUPER SHIFT + HJKL swap tiles

And change conflicts:
- Show keybindings: from SUPER+K to SUPER+F1
  - F1 was chosen since is the main help menu to navigate through the
    system
- Toggle split: from SUPER+J to SUPER+S


dhh:
Appreciate the suggestion, but I wouldn't use this, and I don't want to move the hotkeys around any more. Good tweak to make to a local config though!

PR #2698, opened 2025-10-22
fix: alacrity not supporting emojis

glauberdmo:
This project contains noto-fonts-emoji, but it is not properly set. Was added a conf in fontconfig/conf.d


chaddestchad67:
https://github.com/user-attachments/assets/ad288fde-e35d-4ae1-94e1-7b299d98b175

PR #2721, opened 2025-10-22
Feature chromium srvc

brutalsam96:
### Problem

When logging out or shutting down the system, Chrome/Chromium shows "didn't shut down correctly" 
warnings because browsers are terminated abruptly (SIGKILL) instead of gracefully (SIGTERM).

### Solution

Added systemd user service that hooks into uwsm's `wayland-session-shutdown.target` to 
gracefully shutdown Chromium-based browsers before compositor exits.

Added service copy and enable script to migrations(needs testing)

### Changes

- Added `chrome-graceful-shutdown.service` to handle browser shutdown
- Supports Chrome, Chromium, Brave, Vivaldi, Opera
- 10-second grace period before force-kill
- Integrated with uwsm's session management

### Testing

- [x] Tested on Omarchy with uwsm 0.24.1
- [x] Chrome no longer shows crash warnings after Reboot or Log out
- [x] Browser Session state preserved correctly
- [x] Works with multiple browsers simultaneously (at least 2)
- [x] Works when browsers are closed(running in background) or when shutdown is triggered when browsers are running
- [ ] Testing with more browsers (only tested with chrome and chromium simultaneously)
- [ ] Testing migrations script it should copy and enable the service

### Installation

- To manually install it need to copy chrome-graceful-shutdown.service file to `~/.config/systemd/user/` and enable and start the service 

```bash
systemctl --user enable chrome-graceful-shutdown.service
systemctl --user start chrome-graceful-shutdown.service
```

### DIFF
--- a/config/systemd/user/chrome-graceful-shutdown.service
+++ b/config/systemd/user/chrome-graceful-shutdown.service
@@ -1,0 +1,56 @@
+[Unit]
+Description=Graceful Chromium-based browser shutdown
+Documentation=https://wiki.archlinux.org/title/Systemd/User
+DefaultDependencies=no
+Before=wayland-session-shutdown.target
+PartOf=wayland-session@.target
+StopWhenUnneeded=yes
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStart=/bin/true
+# Gracefully shutdown Chrome/Chromium/Brave and other chromium-based browsers
+ExecStop=/usr/bin/sh -c ' \
+    browsers="chrome chromium brave vivaldi opera"; \
+    running_browsers=""; \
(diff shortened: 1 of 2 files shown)


brutalsam96:
@hjanuschka 
Howdy,
This is a follow-up on the issue/PR you opened earlier. I’ve created a service that hooks into the uwsm shutdown target and gracefully closes browsers before the session ends.
Feel free to check it out — I’d love your feedback!

brutalsam96:
I have refactored the code a little bit, used dev branch to create migrations and tested migration scripts on a new Pull Request you can check it here:
[Link](http://github.com/basecamp/omarchy/pull/2841)

PR #2738, opened 2025-10-22
Synchronize the chromium background with omarchy

samrudhnelli:
Automatically change the chromium background whenever the desktop background is changed.

### DIFF
--- a/bin/omarchy-theme-bg-next
+++ b/bin/omarchy-theme-bg-next
@@ -4,7 +4,6 @@
 
 BACKGROUNDS_DIR="$HOME/.config/omarchy/current/theme/backgrounds/"
 CURRENT_BACKGROUND_LINK="$HOME/.config/omarchy/current/background"
-CHROMIUM_BACKGROUND_LINK="$HOME/.config/chromium/Default"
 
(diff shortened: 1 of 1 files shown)


basestchad69:
@dhh AAAHGHGHGHGH WOKE 🤬🤬🤬🤬🤬🤬🤬🤬🤬🤬

dhh:
Hmm, didn't seem to change anything for me when changing. Maybe this only works if you've set your chromium to use a background already, which isn't the default?

samrudhnelli:
> Hmm, didn't seem to change anything for me when changing. Maybe this only works if you've set your chromium to use a background already, which isn't the default?

oh yes, it works only for those who had set a custom background. I'll look into it..

samrudhnelli:
A migration would be needed that asks for approval to close chromium and run `omarchy-theme-bg-next.`
After this migration the chromium background would be set according to the local file `(.config/chromium/Default/background.jpg)`. We just replace the background.jpg with the current background when `omarchy-theme-bg-next` is run from then onwards.

samrudhnelli:
@dhh this needs a review..

dhh:
On further reflection, I actually prefer the look where it's just the solid color from the theme. It gets a bit much to repeat the background image on top of a desktop that already has the same. Appreciate the work here, though ✌️

Issue #2558, opened 2025-10-19
Major keybinds broke after omarchy update


rafa-alves6:
### System details

Omarchy 3.1.0

### What's wrong?

I updated omarchy by clicking the refresh button on the waybar, and a lot of keybinds broke, for example:

- Super + Space
- Super + Alt + Space
- Super + return
- Super + n
- Super + m
And many others...

Basically I've only tested and opening htop works, and opening file explorer with super + E. 
Now the main issue is I can't open the terminal (super + return).
I tried using ctrl + alt + f2 to open tty and tried chat gpt'ing my way out of it, but to no success.

I'm wondering if there is a way to reset everything to default configs or will I need to reinstall again:

wekauwau:
https://github.com/basecamp/omarchy/releases/tag/v3.1.0

deepu7d:
@wekauwau any solution?
<img width="477" height="110" alt="Image" src="https://github.com/user-attachments/assets/442fbf2a-465e-432a-a177-dad75b6c6d8d" />

<img width="982" height="666" alt="Image" src="https://github.com/user-attachments/assets/2a91a535-01d1-45aa-8bd8-88630ce2eac2" />

morgoth:
Looks like a dupliate of https://github.com/basecamp/omarchy/issues/2548

deepu7d:
> Looks like a dupliate of [#2548](https://github.com/basecamp/omarchy/issues/2548)

thanks

sepulworld:
reinstalling walker doesn't help

sepulworld:
```
omarchy-refresh-walker
bash: command not found: Omarchy-refresh-walker
```

looks like all got he Omarchy command except for omarchy-nvim-setup are gone. Odd.

deepu7d:
> ```
> omarchy-refresh-walker
> bash: command not found: Omarchy-refresh-walker
> ```
> 
> looks like all got he Omarchy command except for omarchy-nvim-setup are gone. Odd.

yay -S walker
omarchy-refresh-walker 
This worked for me
I guess  capital O is causing the issue :)

bscott:
Major keybinds are not working for me either, wsa able to reinstall and get the SUPER+Space menu to come up but bindings such as SUPER+B still are not.

umibear:
I lost all major keybinds aswell, omarchy-nvim-setup is only omarchy command i have anymore. I can't open any of the omarchy menus etc.

EDIT:
[...]

PR #2740, opened 2025-10-22
bug: if uwsm is installed we should cleanup

sepulworld:
The current logic will not cleanup because uwsm is installed during the update process

### DIFF
--- a/migrations/1752292967.sh
+++ b/migrations/1752292967.sh
@@ -1,6 +1,6 @@
 echo "Update to use UWSM and seamless login"
 
-if omarchy-cmd-missing uwsm; then
+if omarchy-cmd-present uwsm; then
   sudo rm -f /etc/systemd/system/getty@tty1.service.d/override.conf
   sudo rmdir /etc/systemd/system/getty@tty1.service.d/ 2>/dev/null || true
 


sedubois:
Thanks @sepulworld, I've tested this patch and can confirm this allows the migration script to run while the original version didn't (which resulted in a non-functional Omarchy: no Walker menus on super-Alt-space, no terminal on super-Enter, cfr. https://github.com/basecamp/omarchy/issues/2543).

sedubois:
NB: an extra improvement would be to ensure that `OMARCHY_PATH="$HOME/.local/share/omarchy"` and `PATH="$OMARCHY_PATH/bin:$PATH"` are also exported when using a plain TTY (ctrl-alt-F3), as it was needed to manually run the migration from there to restore a functional Omarchy, but not sure what's the best place to do that.

migrations/1752292967.sh
```diff
@@ -1,6 +1,6 @@
 echo "Update to use UWSM and seamless login"
 
-if omarchy-cmd-missing uwsm; then
+if command -v uwsm &> /dev/null; then
```
dhh: We can use `omarchy-cmd-present` here.

sepulworld: yes! good call. updated and squashed commits.

nkxxll:
Yes I can confirm that this works and actually closes this #2558 issue which was prematurely closed

ryanrhughes:
This migration is incredibly old (https://github.com/basecamp/omarchy/pull/140) and was used to get people from the pre-uwsm days to uwsm version in v1.4.0. We used the uwsm check as a trigger back then to determine if this should run or not because it predated the migrations state logging.

At the point you'd be running this migration as a user, you wouldn't have uwsm installed and would be launching into Omarchy without a Plymouth loader even.

I have intent to go back and consolidate some of these old migrations since currently we'll try to setup omarchy seamless login just to turn around and delete it in favor of SDDM which is wasteful and error prone if we still have some super-early builds floating around.

The Walker issues mentioned here are entirely unrelated to this. If you didn't have uwsm installed; nothing would work past v1.4.0.

nkxxll:
> The Walker issues mentioned here are entirely unrelated to this. If you didn't have uwsm installed; nothing would work past v1.4.0.

wasn't a walker issue that is my point... walker not working was just a symptom of uwsm environment not being loaded correctly (for some people in the issue including me). I wanted to link is PR in case someone is searching for a solution like me. Way forward is fine for me. Maybe pin one issue for the uwsm migration problem since some people will be searching for this.

ryanrhughes:
The only way that could happen is that the migration ran previously, failed, wasn't caught, and re-running passed. That is entirely possible as we have some holes we've plugged with the update / migration system since then.

We honestly might just be able to remove the check altogether since the migration state handles this now and it was really just to speed up `migrate all` runs which aren't even a thing anymore.

nkxxll:
that is a good idea! Thank you for your great work keep it up 🔥

PR #2748, opened 2025-10-22
Adding title to TUI App

tom-jagus:
Added a app title to a TUI App installer allowing better recognition of a installed app. This mainly comes from need to set specific hyprland windowrule to given App only.

### DIFF
--- a/bin/omarchy-tui-install
+++ b/bin/omarchy-tui-install
@@ -38,12 +38,14 @@ else
   APP_CLASS="TUI.tile"
 fi
 
+APP_TITLE="TUI.$APP_NAME"
+
 cat >"$DESKTOP_FILE" <<EOF
 [Desktop Entry]
 Version=1.0
 Name=$APP_NAME
 Comment=$APP_NAME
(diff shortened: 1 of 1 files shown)


csfh:
Closing — stale against `dev`, and the approach doesn't match how Omarchy launches terminals now.

On `quattro`, TUI desktop entries go through `xdg-terminal-exec --app-id=…`, not `$TERMINAL --class=…`. Foot in particular does not take `--class` (it wants `--app-id`), so a raw `$TERMINAL --title=… --class=…` line is not portable across the four approved terminals.

`--title` itself exists on Alacritty, Foot, Ghostty, and Kitty, and `xdg-terminal-exec --title=` would map it via each desktop file's `X-TerminalArgTitle=`. But title-based Hyprland rules are a weak fit here anyway: installed TUIs already share `TUI.float` / `TUI.tile`, and apps routinely overwrite the window title. If you need per-app rules, give each installed TUI its own `--app-id` (e.g. `TUI.lazydocker`) rather than relying on `--title`.

PR #2752, opened 2025-10-23
Fix auto-generated everforest obsidian theme readability issues

alexandersix:
## Summary

This PR updates the dynamic Obsidian theme logic to prevent menu icons and text from being unreadable due to contrast issues.

## More Details

When using the "Omarchy" theme in Obsidian with the `everforest` theme, the sidebar, title bar, and tab items are unreadable due to a low color contrast with the background.

To fix this, I have added in some very basic color contrast checking that ensures the text color (which defaults to the `$border_color`) can be read on top of the background color. If these values do _not_ have a decent contrast ratio, I set the two CSS variables responsible for the text (`--text-muted` and `--text-faint`) to the `foreground` color that comes from the Alacritty config file. Note that, in doing this, we keep the border color the same, since some themes are expecting their borders to not stand out from the background (like in the case of `everforest`, for example).

This fix does **NOT** impact the following:

- The ability to create custom `obsidian.css` files for themes
- Attempting to create a valid `$border_color` if one does not exist (though we _do_ check this new value for contrast issues)
- Any themes that both had a sufficiently contrasting `$border_color` and used the automatic Obsidian theme generation code

Happy to chat more about this if needed 👍

<details><summary>**Before**</summary>
<p>

screenshot-2025-10-22_20-41-58

</p>
</details> 

<details><summary>**After**</summary>
<p>

screenshot-2025-10-22_20-42-41

</p>
</details>

### DIFF
--- a/bin/omarchy-theme-set-obsidian
+++ b/bin/omarchy-theme-set-obsidian
@@ -258,38 +258,30 @@ extract_theme_data() {
     fi
   done
 
-  # Sort by distance and get the closest color for code background
-  local -a closest_to_bg
-  readarray -t closest_to_bg < <(printf '%s\n' "${bg_distances[@]}" | sort -n | head -1 | cut -d: -f2)
-
   # All background variations use the same as primary background
   local bg_primary_alt="$bg_color"
   local bg_secondary="$bg_color"
   local bg_secondary_alt="$bg_color"
 
-  # Code block background uses the closest different color
-  local code_bg="${closest_to_bg[0]}"
-
-  # If no different color available, create a subtle variant for code blocks
-  if [ -z "$code_bg" ]; then
-    read -r r g b <<<"$(hex_to_rgb "$bg_color")"
-    if [ $bg_brightness -gt 127 ]; then
-      r=$((r - 10))
-      g=$((g - 10))
-      b=$((b - 10))
-    else
-      r=$((r + 15))
-      g=$((g + 15))
-      b=$((b + 15))
-    fi
-    [ $r -lt 0 ] && r=0
-    [ $r -gt 255 ] && r=255
-    [ $g -lt 0 ] && g=0
-    [ $g -gt 255 ] && g=255
-    [ $b -lt 0 ] && b=0
-    [ $b -gt 255 ] && b=255
-    code_bg=$(printf "#%02x%02x%02x" "$r" "$g" "$b")
+  # Generate code background color that will contrast with foreground text
+  read -r r g b <<<"$(hex_to_rgb "$bg_color")"
+  if [ $bg_brightness -gt 127 ]; then
+    r=$((r - 10))
+    g=$((g - 10))
+    b=$((b - 10))
+  else
+    r=$((r + 15))
+    g=$((g + 15))
+    b=$((b + 15))
   fi
+  [ $r -lt 0 ] && r=0
+  [ $r -gt 255 ] && r=255
+  [ $g -lt 0 ] && g=0
+  [ $g -gt 255 ] && g=255
+  [ $b -lt 0 ] && b=0
+  [ $b -gt 255 ] && b=255
+  code_bg=$(printf "#%02x%02x%02x" "$r" "$g" "$b")
+
 
   # Find closest color to foreground for code block text
   local code_fg=""


dhh:
Yeah, we have several issues with the auto-generated themes for other of the default themes too. I'd like to see everything solved in one fell swoop. Either that we actually just go explicit for these default themes whenever there are issues or we fix the generator. Want to take a look at the more comprehensive fix?

alexandersix:
> Yeah, we have several issues with the auto-generated themes for other of the default themes too. I'd like to see everything solved in one fell swoop. Either that we actually just go explicit for these default themes whenever there are issues or we fix the generator. Want to take a look at the more comprehensive fix?

That's totally fair--I'm happy to take a look when I have a bit of time and see what I can come up with!

There's some part of me as well that wonders if it's actually the right decision to have the default themes come with explicit theme files for the included applications instead of defaulting to something auto-generated, but the rational part of my brain remembers that there are a LOT of themes included by default and that would be a ton of work to keep up-to-date.

alexandersix:
Ok, curiosity got the better of me and I did a bit of digging:

We're currently running into this particular bug due to the use of the `$border_color` variable supplied by `btop` being applied to the menu items and navigation. This isn't necessarily bad (for example, the `osaka jade` theme looks great due to its high-contrast borders), but for some themes like `everforest`, the border colors are left intentionally subtle.

What I'm thinking of doing here is to write a bit of code that checks the contrast level of the `$border_color` variable in the `omarchy-set-theme-obsidian` script and, if the colors don't contrast enough, set it to the `foreground` color that comes from the `alacritty` configuration for the given theme. Ideally, I'll be able to just set the text to this color and leave the actual borders alone (to preserve that nice low-contrast border look in some themes), and in themes where the contrast is already good, everything can stay as-is.

The only hiccup that I'm bumping into so far is that Bash doesn't handle decimals very well. It seems (from some limited research) that the going method of handling decimals is to pipe any decimal math to the `bc` command line tool, but I don't seem to have that installed in Omarchy by default. Instead, I'm just doing the good old "multiply by 1000 and do integer math" solution. However @dhh, if you're open to having `bc` included in Omarchy, I can very easily adjust the logic to use that tool instead.

I'm not a color expert by any means, but I'll get a "good enough" contrast detection system going and am more than happy for someone more in-tune with color contrast systems to give their input as well.

dhh:
Think best if we can keep it all in bash, but if it's crazy cumbersome, we can consider something else. Thanks for investigating this! 🤘

alexandersix:
For sure! I'd prefer not adding dependencies if possible, so I'll leave `bc` as a last resort. For what it's worth, using `bc` in our bash scripts would look something like this:

```bash
echo "20.5 + 5" | bc
```

I'll keep plugging away and report back when I have something 👍

alexandersix:
Alright, I have something written up that's working on my end!

I've updated the PR description (and can also update the title, if so desired) to match what is now happening in the code, but here's the TL;DR:

- Contrast checking has been added and implemented for `$border_color`
- Any `$border_color` values that _aren't_ sufficiently contrasting against the background not used for menu icons and text, instead these will use the `foreground` color that is used by the normal Obsidian text
- All validly contrasting themes are not affected and should remain the same as they were before.

dhh:
Flexoki still seems to struggle:

image

And Catppuccin Latte:

image

Everforest is also very hard to see:

image

Think we need something that sorts this across the entire default range of themes.

alexandersix:
@dhh that's really strange--all of these (and the rest of the default themes) were working fine on my end (everforest _for sure_, since that's the theme that I use every day). 

I'm noticing now that, with some recent update since I pushed the last commit to this PR, none of my changes are working on my local install and the Obsidian theme isn't getting changed correctly when I switch themes (it just stays as a purple default). When I get back to my desk today, I'm going to click through the changelog and see if anything has potentially changed that would be causing this.

It seems like, now, switching themes with the theme switcher causes the Obsidian theme to go back to default, but then running the `omarchy-theme-set-obsidian` script manually after the theme change fixes the problem. This feels like an issue with my local install, so I'll poke around.

alexandersix:
Ok, a quick update and full reboot seems to have sorted the theme switching issue.

I also found the issue--seems like a couple of lines from my changes didn't make it into my last push. They have been added now, and these themes are looking _much_ better.

However, I'm noticing that, for the two light themes, the background color of the selected item in the nav pane is the same color as the text itself, rendering that particular entry unreadable. I'm going to dig in a bit and see what might be causing that as well.

alexandersix:
Alright, just pushed up a "fix" for the strange code block and selected note background issue.

Long story short, it stems from the current generation logic selecting the "closest color to the background". This works really well for the dark themes, but for some reason (which I'm still trying to wrap my brain around), it causes some very strange problems with light themes.

The "fix" I've added is to remove the "closest color" logic altogether in favor of the backup color generation that styles the `code_bg` color with a computed color based off of Obsidian's background color.

This isn't a _perfect_ fix, since we're generating colors now that aren't necessarily in the color palette, but it _does_ quickly fix the legibility issue. I'd like to keep thinking through the logic here, but, given that the state of the contrast in Obsidian is pretty poor right now, it might be worthwhile to add this in as a stop-gap until we can figure out something more robust with the color generation. One thing to note is that this changes the backgrounds of the dark theme code blocks/selected notes as well--I've checked through them and they all look fine, but the new colors are no longer in the palette.

Would love some thoughts here. 

If we decide that we want to just merge this in as-is to fix some of the contrast issues, I'm happy to open up another PR once I can figure out a better system of using the existing palette colors, only falling back to the generated color if the contrast isn't up to snuff.

**EDIT: I just had a thought that I might be able to compare the contrast like I did for the border colors and only apply this fallback on light themes. Alternatively, we have an `is_light_theme` variable that I could leverage as well. That's not currently in the PR, but I'll add that in before we merge to lessen the fallout of this change to _just_ the light themes (which were the ones having trouble with code block contrast initially)**

dhh:
It would be great to stick to colors only in the theme. If you have a strategy here, we'll hold 3.1.4 until you have that in this patch 👌

alexandersix:
Definitely agree about sticking to colors in the theme.

I don't have a strategy of any sort at the moment. Lots of hunches, but nothing concrete. I have about an hour today to take a look, so let me see if I can come up with something that might work 🤔

At the very least, we know that we only really need to come up with something that will work for light themes, because this already works well for dark theme palettes.

dhh:
Another option is just to do the light themes manually for now?

dhh:
Actually, I just looked all the default themes through with this new version. It looks way better. Good enough to ship. Only problem is that the headings are all red or pink?

image

image

image

alexandersix:
> Another option is just to do the light themes manually for now?

That might not be a bad idea for the moment, but I'd hate to have two separate methods of applying Obsidian themes.

I think the main issue boils down to the fact that there are very few colors in a terminal color theme compared to the number of colors that can be used in Obsidian. I was looking at [Kepano's GH repo for the Flexoki obsidian theme](https://github.com/kepano/flexoki-obsidian), for example, and there are a lot more colors being used here than we have in his Alacritty theme. So maybe we really _should_ go fully custom to get a nice range of colors (or defer to an existing Obsidian theme, kind of like what we do with Neovim themes if I understand correctly)

> Actually, I just looked all the default themes through with this new version. It looks way better. Good enough to ship. Only problem is that the headings are all red or pink?

That seems strange -- things appear to work as intended on my end.

<details><summary>Screenshots</summary>
Everforest:

image

Matte Black:

image

Osaka Jade:

image

Catppuccin Latte:

image

</details>

dhh:
Ah, looks that then by design. The pinks just look a bit off to me. But let's just merge this for now and then continue tweaking in a follow up. Thanks for working on this!

alexandersix:
Fair enough haha. I think there's a world where we end up making custom themes for a handful of these default (if not all of them), so we can always cross that bridge when we need to.

And no problem, happy to help!

the-exodus:
Just weighing in here that the theming stuff seems to be affecting vscode too, if that's possible? I haven't manually set a theme for vscode, but using everforest, and the panel borders are the same color as the background.

dhh:
VSCode uses direct extensions on the VSCode repo so that's not related.

PR #2758, opened 2025-10-23
add cmd+a as a shortcut for select all

pachuc:
@dhh loving the new shortcuts. really helps since I have to be on mac all day for work, and then I come home and switch to omarchy. may I suggest this one as well?

### DIFF
--- a/default/hypr/bindings/clipboard.conf
+++ b/default/hypr/bindings/clipboard.conf
@@ -2,4 +2,5 @@
 bindd = SUPER, C, Copy, sendshortcut, CTRL, Insert,
 bindd = SUPER, V, Paste, sendshortcut, SHIFT, Insert,
 bindd = SUPER, X, Cut, sendshortcut, CTRL, X,
+bindd = SUPER, A, SELECT, sendshortcut, CTRL, A,
 bindd = SUPER CTRL, V, Clipboard, exec, omarchy-launch-walker -m clipboard


mythz:
I've also added "undo" to cover the most frequent CTRL shortcuts:

```
bindd = SUPER, Z, SELECT, sendshortcut, CTRL, Z,
```

pachuc:
I've been thinking about this a bit, and now that I have super+c/super+v/super+a shortcuts, I find myself naturally drifting towards other super based shorcuts (such as super+s for saving). I think this is because my hand is now naturally resting/positioned for super for these keybinds rather than ctrl, and using super for some and ctrl for others feels awkward. At the same time, mapping every shortcut to have ctrl & super equivalents in this way seems untenable. I think a better long term option would be for Omarchy to allow the user to pick "mac" or "windows" mode, and based on that it would configure every shortcut to use either super or ctrl. Maybe this needs to be done more at a system level, rather than at the hyperland level?

mythz:
I'd be happy with just the `x` `c` `v` `a` `z` bindings since I use them the most, and often in the same editing workflow, where having to switch back between CTRL/SUPER keys is felt the most.

I don't use CTRL+S nearly as much where my muscle memory still reaches for the correct key binding. If it's used in an editing workflow it's right at the end so I'm not going to immediately go back and switch to using SUPER again. So IMO `s` wouldn't be necessary.

pachuc:
My muscle memory is to spam save after every change lol. Another one I ran into is super+F for find, but in omarchy right now that toggles full screen. 

I do think for the best product experience for people transitioning from other OSs, we should provide a consistent experience where all hotkeys are mapped to either super or ctrl based on what you prefer. I think the current hodgepodge of most keys being ctrl, but then allowing some of the shortcuts on super as well is a bit of a weird product experience. However, that said, I also realize that needing a universal copy/paste for the terminal also complicates this a bit more.

Anyway, just my 2 cents. Would be happy to just have super a/x added for now, and I will customize my own setup to my preference.

dhh:
I think it's a fine line here, but I'm going to stick with just the clipboard universals, as those are the ones that differ between terminal and the rest. There are a lot of other selection keybindings we could try to map, but I think that's going too far. You're of course always free to add to your own config ✌️

PR #2769, opened 2025-10-23
Changing Window Focus keybindings to SUPER+<HJKL> similar to i3wm and VIM

joshuafouch:
Hello,

I would like to make a PR for changing the window focus keybinds to **Super + HJKL** as this is the standard for VIM and also similar to the keybinds in other tiling window-managers such as i3wm, dwm, bpswm, and so on. However, i3wm uses <j k l ;> for changing focus, but most people rebind this to HJKL. I have found this very helpful for me rather than moving my hand slightly over to the arrow keys or using ALT+TAB. 

I also changed the "Show Keybinding" keybind to SUPER+B (B for bindings) to account for the rebind of SUPER+K. 
Additionally, I changed the "Toggle Split" keybind to SUPER+S (S for split) to account for the rebind of SUPER+J.

I still kept the SUPER+<arrow keys> as well if people still like to use that.

Please let me know if you have any other questions!

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,9 @@
 # Omarchy
 
-Turn a fresh Arch installation into a fully-configured, beautiful, and modern web development system based on Hyprland by running a single command. That's the one-line pitch for Omarchy (like it was for Omakub). No need to write bespoke configs for every essential tool just to get started or to be up on all the latest command-line tools. Omarchy is an opinionated take on what Linux can be at its best.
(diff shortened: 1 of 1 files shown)


dhh:
Thanks for the suggestion, but I wouldn't use this, and I don't want to move keybindings again. But this would make for a nice tutorial in Discussion > Manuals for those who want this ✌️

PR #2771, opened 2025-10-23
switch background-color to window#waybar

coreydemarse:
allows for themes with semi-transparent waybar

with background-color on * the semi-transparency is applied to all elements which doesn't look right with themes that have a semi-transparent waybar


coreydemarse:
just realised the indentation is off on line 14, needs to be fixed before merging (oops)

coreydemarse:
going to open a new PR

PR #2772, opened 2025-10-23
switch background-color to window#waybar

coreydemarse:
allows for themes with semi-transparent waybar

with background-color on * the semi-transparency is applied to all elements which doesn't look right with themes that have a semi-transparent waybar

### DIFF
--- a/config/waybar/style.css
+++ b/config/waybar/style.css
@@ -1,7 +1,6 @@
 @import "../omarchy/current/theme/waybar.css";
 
 * {
-  background-color: @background;
   color: @foreground;
 
   border: none;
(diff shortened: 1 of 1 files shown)


coreydemarse:
if this change is accepted, people using themes with a semi-transparent waybar color could also add blur to their waybar by modifying their `~/.config/hypr/looknfeel.conf` and adding a few lines to their decoration section

```
decoration {
    # add these lines for blur
    layerrule = blur,waybar
    layerrule = ignorezero,waybar
}
```

image

coreydemarse:
closed as you can just do this in your theme

```
@define-color background rgba(0,0,0,0);

window#waybar {
    background-color: rgba(0, 0, 0, 0.1);
}
```

PR #2775, opened 2025-10-23
Show correct keybinding for omarchy-menu in waybar

alex-305:
Shows user's correct keybinding for omarchy-menu in waybar.

So this is what I set my keybinding to:
image

And this is now properly reflected in waybar's omarchy-menu tooltip:
image

### DIFF
--- a/bin/config.jsonc.backup
+++ b/bin/config.jsonc.backup
@@ -39,10 +39,9 @@
     }
   },
   "custom/omarchy": {
-    "exec": "$OMARCHY_PATH/default/waybar/indicators/omarchy-menu.sh",
-    "signal": 6,
-    "return-type": "json",
+    "format": "<span font='omarchy'>\ue900</span>",
     "on-click": "omarchy-menu",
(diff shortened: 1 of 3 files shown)


dhh:
Appreciate the suggestion, but I think it's too much code for the value. You can just change your config.jsonc if you want it to match.

(Very cool backgrounds, btw!)

alex-305:
> Appreciate the suggestion, but I think it's too much code for the value. You can just change your config.jsonc if you want it to match.
> 
> (Very cool backgrounds, btw!)

@dhh I accidentally included my backup waybar.config.jsonc in this commit. It should only be 40 lines added instead of 178. I've removed it in my most recent commit but it was after you closed this PR. I'm wondering if that changes your mind at all on this since it's a nice QOL thing with very little overhead.

dhh:
I'm thinking the judo version, now that we have a notification teaching you about keybindings pop up on new installs, is perhaps even just to remove that clue.

alex-305:
Yeah that seems pretty sensible. I haven't encountered those keybinding tips anywhere else.

Issue #2599, opened 2025-10-20
Windows VM: RDP connects too early, “Connection reset by peer”, VM auto-stops


nabeeladzan:
### System details

i5 8350u, Intel UHD 620, Omarchy 3.10

### What's wrong?

Launching the Windows VM from the desktop entry or via omarchy-windows-vm launch does not open an RDP window. The CLI attempt shows FreeRDP errors and the VM container is stopped automatically. Desktop entry only shows a notification and then appears to do nothing.

```
omarchy master ❯ omarchy-windows-vm launch
Starting Windows VM...
[+] Running 2/2
 ✔ Network windows_default    Created                                                                                                                                                           0.2s
 ✔ Container omarchy-windows  Started                                                                                                                                                           0.9s
Waiting for Windows VM to be ready...

 ┌─────────────────────────────────────┐
 │                                     │
 │      Connecting to Windows VM       │
 │                                     │
 │  VM will auto-stop when RDP closes  │
 │                                     │
 └─────────────────────────────────────┘

[14:14:38:312] [159106:00026d82] [WARN][com.freerdp.client.common.cmdline] - [warn_credential_args]: Using /p is insecure
[14:14:38:312] [159106:00026d82] [WARN][com.freerdp.client.common.cmdline] - [warn_credential_args]: Passing credentials or secrets via command line might expose these in the process list
[14:14:38:312] [159106:00026d82] [WARN][com.freerdp.client.common.cmdline] - [warn_credential_args]: Consider using one of the following (more secure) alternatives:
[14:14:38:312] [159106:00026d82] [WARN][com.freerdp.client.common.cmdline] - [warn_credential_args]:   - /args-from: pipe in arguments from stdin, file or file descriptor
[14:14:38:312] [159106:00026d82] [WARN][com.freerdp.client.common.cmdline] - [warn_credential_args]:   - /from-stdin pass the credential via stdin
[14:14:38:312] [159106:00026d82] [WARN][com.freerdp.client.common.cmdline] - [warn_credential_args]:   - set environment variable FREERDP_ASKPASS to have a gui tool query for credentials
[14:14:38:326] [159106:00026d84] [WARN][com.freerdp.client.x11] - [load_map_from_xkbfile]:     : keycode: 0x08 -> no RDP scancode found
[14:14:38:327] [159106:00026d84] [WARN][com.freerdp.client.x11] - [load_map_from_xkbfile]: ZEHA: keycode: 0x5d -> no RDP scancode found
[...]

PR #2780, opened 2025-10-23
Improve RDP readiness check before launching VM

3x3cut0r:
- add a helper that waits for the RDP port and performs an auth-only FreeRDP probe before launching the client
- extract Windows credentials up front and reuse them for the readiness check
- abort the launcher when RDP cannot be verified within 25 seconds and notify the user
(#2599)


cadmium621:
Request following add to line 315 in change c3ac950 ~/.local/share/omarchy.bin/omarchy-windows-vm file:
# Microsoft Account Login Prompt
WIN_PASS=$(zenity --title="Microsoft Account Password" --password)

ryanrhughes:
Merged via #4089. I couldn't push changes back up to the remote tracked here.

PR #2783, opened 2025-10-24
chore: better security and tooling

shelldandy:
I was reading this article https://マリウス.com/a-word-on-omarchy/ and thought that some things seemed pretty quick to address so this is the PR for it:

- Reduce finger sudo attempts from 10 to 5
- Add essential cli packages to the base install
  - bc - command-line calculator
  - cronie - cron daemon for scheduled tasks
  - gnu-netcat - network debugging
  - rsync - file sync/backup
  - screen - terminal multiplexer
  - tcpdump - network packet analyzer
- Add useful aliases (cat => bat, top =>btop, etc)


abenz1267:
Omarchy already installs `libqalculate` for calc, because it's a dependency for `elephant-calc`.

I don't know about `gnu-netcat` .... I'd recommend using `openbsd-netcat` as it has the `-U`-flag. That can be used to improve startup time of  Walker f.e.

dhh:
Appreciate the effort, but I'm not interested in any of the additional tools from that screed. Nor is the fingerprint count a security issue of note. We've discussed the cat issue before, and it doesn't work because so many scripts use cat and expect clean output. And we already have btop bound to Super + Shift + T.

PR #2784, opened 2025-10-24
Fix shellcheck linter warnings

iho:
Hi folks! 

You have many shell scripts and many contributors, and you need a linter to avoid rookie mistakes from people who don't write shell scripts often. I would recommend asking people to have a linter in their editor https://github.com/koalaman/shellcheck#user-content-in-your-editor

https://www.shellcheck.net/wiki/SC2046
https://www.shellcheck.net/wiki/SC2155
https://www.shellcheck.net/wiki/SC2162
https://www.shellcheck.net/wiki/SC2004
https://www.shellcheck.net/wiki/SC2086
https://www.shellcheck.net/wiki/SC2124
https://www.shellcheck.net/wiki/SC2128
https://www.shellcheck.net/wiki/SC1087
https://www.shellcheck.net/wiki/SC2005
https://www.shellcheck.net/wiki/SC2207
https://www.shellcheck.net/wiki/SC2183
https://www.shellcheck.net/wiki/SC2001
https://www.shellcheck.net/wiki/SC2143
https://www.shellcheck.net/wiki/SC2059
https://www.shellcheck.net/wiki/SC2207


dhh:
Appreciate the suggestion, but stylistically I disagree with most of these choices. Not against having a linter, but it should follow the style we already have.

PR #2786, opened 2025-10-24
Install Herd-parity PHP extensions for php/laravel/symfony

### LINKED ISSUE #2049: Install Development > Laravel - should include default php extentions
### What do you need?

Laravel typically includes the most common php extensions preinstalled for first party services like laravel cloud, laravel forge, laravel herd, and the php.new environment setup.

see: https://herd.laravel.com/docs/macos/technology/php-extensions
and php.new

The install Laravel should aim for parity with the recommended Laravel php.new and other first class Laravel services.

**Comparison of Enabled Modules**
php.new Installed | Omarchy Enabled | Status

bcmath | bcmath | ✅ Enabled
bz2 | (Missing) | ❌ Missing
calendar | (Missing) | ❌ Missing
ctype | ctype | ✅ Enabled
curl | curl | ✅ Enabled
dba | (Missing) | ❌ Missing
dom | dom | ✅ Enabled
exif | (Missing) | ❌ Missing
ffi | (Missing) | ❌ Missing
fileinfo | fileinfo | ✅ Enabled
filter | filter | ✅ Enabled
ftp | (Missing) | ❌ Missing
gd | (Missing) | ❌ Missing
gmp | (Missing) | ❌ Missing
hash | hash | ✅ Enabled
iconv | iconv | ✅ Enabled
imagick | (Missing) | ❌ Missing
imap | (Missing) | ❌ Missing
intl | intl | ✅ Enabled
json | json | ✅ Enabled
ldap | (Missing) | ❌ Missing
mbstring | mbstring | ✅ Enabled
openssl | openssl | ✅ Enabled
pcre | pcre | ✅ Enabled
pdo | PDO | ✅ Enabled
pdo_mysql | pdo_mysql | ✅ Enabled
pdo_sqlite | pdo_sqlite | ✅ Enabled
tokenizer | tokenizer | ✅ Enabled
xml | xml | ✅ Enabled
opcache | xdebug | ❌ Missing (Zend Module)
zip | zip | ✅ Enabled

meirdick:
Brings the `php`, `laravel`, and `symfony` dev-env installs closer to the extension set Laravel first-party tools ship (php.new, Herd, Forge).

- Installs `php-gd` and `php-imagick` from official repos via `omarchy-pkg-add`
- Enables `bz2 calendar dba exif ffi ftp gd gmp ldap` in php.ini, same mechanism as the existing six
- Enables Imagick via the `/etc/php/conf.d/imagick.ini` snippet its package ships
- OPcache: built in and on by default on php 8.5, nothing to do

Ref: https://herd.laravel.com/docs/macos/technology/php-extensions

Closes #2049

https://claude.ai/code/session_01NSB5ZvnEAFedvWqj4eJbvC


weppami:
would be nice to choose php version on install

meirdick:
sure - but project specific php typically managed in omarchy by mise?

csfh (changes requested):
Thanks for chasing Laravel/Herd parity — the gap is real, but this isn't the way we should close it.

### Blockers
1. **`--noconfir` typo** — the `pacman` line won't run as written.
2. **Don't call `pacman` / `yay` here** — `install_php` already uses `omarchy-pkg-add`. Stay on that path (official repos only).
3. **Drop AUR `php-imap`** — IMAP isn't a Laravel baseline requirement, and pulling `yay` into the default Laravel install is the wrong trade. Skip it unless there's a concrete Omarchy use case.
4. **Don't sed `/etc/php/php.ini`** — fragile across package upgrades, and Arch already has the right mechanism.

### Prefer `/etc/php/conf.d` (Arch's tip)
ArchWiki: *instead of editing `php.ini`, enable/configure extensions in `/etc/php/conf.d`*. PHP is built with `--with-config-file-scan-dir=/etc/php/conf.d`. There is no Debian-style `phpenmod` on Arch — **drop-in `.ini` files are the command.**

Suggested shape:

1. Fold packages into `install_php` (shared by `php` / `laravel` / `symfony`):
   ```bash
   omarchy-pkg-add php composer php-sqlite php-gd php-imagick xdebug
   ```
2. Write an Omarchy-owned drop-in, e.g. `/etc/php/conf.d/omarchy-extensions.ini`, listing only what we actually want:
   ```ini
   extension=bcmath
   extension=intl
   extension=iconv
   extension=openssl
   extension=pdo_sqlite
   extension=pdo_mysql
   extension=gd
   ; add bz2/exif/gmp only if we truly want them — not the full Herd kitchen sink
   ```
3. For packages that already ship snippets (`imagick.ini`, `xdebug.ini`): uncomment those package files (as we already do for xdebug). Don't also enable the same module in our drop-in — double-load errors.
4. **OPcache:** bundled with `php`; on current Arch it's default-on since 8.5. Don't invent a stub `opcache.ini` + sed, and don't force `opcache.enable_cli=1` without a strong reason (sticky bytecode surprises local `artisan` / Composer).

### Scope
Herd ships `dba` / `ldap` / `calendar` / `ffi` / `ftp` because it's a kitchen sink — that is not a Laravel requirement list. Aim for what Laravel apps actually need from official packages, not byte-for-byte Herd parity.

Please rewrite along those lines (or close and open a tighter PR against `install_php` on `quattro`).

Jamesking56:
Personally, I feel that `php-imap` is a bad addition. Firstly, its an AUR package that could be taken over by a malicious actor (see recent AUR issues). Secondly, Laravel doesn't require it for local development. If a developer needs to work with IMAP (which I would say is pretty rare), then they can install the AUR and enable the extension manually. Sending mail in Laravel uses SMTP or mail provider APIs, not IMAP, and locally you'd usually use Laravel's `log` driver instead of sending actual emails.

meirdick:
Rewrote per review, rebased on `quattro`.

- Fixed: typo, `pacman`/`yay` calls. Packages now go through `omarchy-pkg-add`, official repos only.
- Dropped: AUR `php-imap`.
- Dropped: the OPcache block. Checked on php 8.5.10: `php -m` lists Zend OPcache and `opcache.enable => On` with no `zend_extension` line. Built in, on by default. The issue table marked it missing from an older php. `opcache.enable_cli` without `file_cache` does not persist bytecode across CLI runs, so no gain for Composer/Artisan either.
- Folded into `install_php`, so `php`, `laravel`, `symfony` all get it.
- Imagick: uncomment the snippet `php-imagick` ships (`; extension = imagick`).

On the drop-in vs php.ini sed: kept the existing sed pattern on purpose. `install_php` already uncommented `bcmath intl iconv openssl pdo_sqlite pdo_mysql` in php.ini on every machine that ran it. A drop-in that lists the same modules gives `Module "bcmath" is already loaded` warnings on those installs. Moving to a drop-in means re-commenting php.ini lines in the same run. Happy to do that as a follow-up if you want the migration. Also, Arch marks php.ini as a backup file, upgrades write `php.ini.pacnew`, so the sed survives updates.

On scope: kept `bz2 calendar dba exif ffi ftp gd gmp ldap`. All ship in the base `php` package, only `gd` needs `php-gd`. They add no install weight, and Herd/php.new/Forge users hit them in real apps (ldap auth, ffi, ftp storage driver, exif on uploads). Open to trimming if you still want it tighter.

csfh:
We can run migrations in omarchy to solve the previous PHP ini files

PR #2801, opened 2025-10-24
Add the keybinding for the copy url extension in the SUPER+K Menu

manuel1618:
This is a proposal to list the important shortcut SHFT ALT + L (copying url of a web app into the clipboard) in the omarchy-menu-keybindings.

Pardon my ignorance if there is a better way to do this, happy to adapt. Also not sure, if I can simply change the manifest of the extension. 

I tested it locally on omarchy 3.1.1

image


dhh:
I definitely would like to see this keybinding included, but I think it's too much to try to dynamically extract from all extensions. We should just have a hardcoded list that we can add to for these cases.

manuel1618:
@dhh agreed

PR #2806, opened 2025-10-24
Fix: Alt + Tab keybinding is mapped to two different actions

opiero:
The Alt + Tab keybinding is current mapped to two different actions: `cyclenext` and and `bringactivetotop`.

As the bringactivetotop had two different keybindings: Alt + Tab and Alt + Shift + Tab, I removed the conflicting one.


axelfontaine:
Not sure about this. The current behavior is most certainly intentional. Just like the duplicate binding for alt shift tab.

opiero:
Oh, I had no idea about that. Should I close the PR then?

dhh:
This is indeed intentional. They go together.

PR #2814, opened 2025-10-24
fix: corrected initramfs hook ordering.

### LINKED ISSUE #2777: After the plymouth boot screen and before hyprland, I see my BIOS/UEFI splash screen momentarily.
### System details

ROG Strix G814JZ, Omarchy 3.1.1

### What's wrong?

I see the ROG boot splash screen logo after the plymouth loading bar loads and before hyprland launches. This started happening after the update. I don't know if this is related to #2699.

denisu14:
This fixes #2777.

I believe it was happening because kms was after plymouth in the hooks sequence. The [wiki](https://wiki.archlinux.org/title/Plymouth) says that plymouth relies on kms to display graphics. It also says plymouth should be before encrypt and I kept it that way.

### DIFF
--- a/install/login/limine-snapper.sh
+++ b/install/login/limine-snapper.sh
@@ -2,7 +2,7 @@ if command -v limine &>/dev/null; then
   sudo pacman -S --noconfirm --needed limine-snapper-sync limine-mkinitcpio-hook
 
   sudo tee /etc/mkinitcpio.conf.d/omarchy_hooks.conf <<EOF >/dev/null
-HOOKS=(base udev plymouth keyboard autodetect microcode modconf kms keymap consolefont block encrypt filesystems fsck btrfs-overlayfs)
+HOOKS=(base udev keyboard autodetect microcode modconf kms keymap consolefont block plymouth encrypt filesystems fsck btrfs-overlayfs)
 EOF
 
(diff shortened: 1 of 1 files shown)


denisu14:
Awaiting review.

ryanrhughes:
@denisu14 After moving this and rebuilding, you're not seeing the flash issue any longer?

nweger:
I tried this, and it didn't help.

Of course i did limine-update after.

EDIT: I tried it using limine-mkinitcpio, and it did actually work now.

denisu14:
It did fix it for me. However, I ended up wiping my disk and getting a fresh install of omarchy. Mainly because after the update a lot of things were not working for me. I don't think this PR is necessary since I don't have the issue now with the default omarchy installation. We can probably close the PR.

As far as I understand, omarchy uses migration scripts to facilitate configuration changes between new releases. I suspect that these scripts broke a lot of stuff for me. This is obviously natural as there are many different hardware in the wild and the scripts may not work for everyone. Also, people change their configurations between updates and this also results in the scripts not having the intended effect. For future updates, is there a way to make these migrations more reliable? For example by having tests and treating the entire update as a transaction and ensuring atomicity, as in if one thing fails the entire update is not applied.

@ryanrhughes 
If this sounds dumb, then probably my understanding is incorrect. I'm not an engineer, just a student. Thanks for taking your time to look into the issue.

PR #2820, opened 2025-10-24
Add script to toggle fingerprint as primary sudo authorization

ramtinJ95:
When fingerprint is setup it is always the "first" option for authorizing a sudo command which is super annoying when the laptop is in clamshell mode and closed far away. So I added this script to toggle it for that situation, because I still need the fingerprint sometimes and want to be able to bring it back easily. 

This is especially annoying when using the default super+alt+space menu installer then I had no other option but to open up the laptop and use the fingerprint or just run the pacman/yay command directly in the terminal which defeats the whole purpose of that menu then.

### DIFF
--- a/bin/omarchy-toggle-fingerprint
+++ b/bin/omarchy-toggle-fingerprint
@@ -1,0 +1,102 @@
+#!/bin/bash
+
+# Toggle fingerprint authentication on/off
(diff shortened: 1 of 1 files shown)


dhh:
We have another PR exploring fixing this via lid controls directly. Let's pursue that path so it's automatic.

PR #2829, opened 2025-10-25
feat: remember waybar hidden status through restart

alexperreault:
Currently, if waybar is toggled using ```omarchy-toggle-waybar```, its state is reset when waybar is restarted, which occurs ofter in Omarchy (i.e. changing themes, updating).

This fixes the problem using the start_hidden option in waybar's config file.

```omarchy-toggle-waybar``` now changes the config and restarts waybar, persisting the bar's hidden state

### DIFF
--- a/bin/omarchy-toggle-waybar
+++ b/bin/omarchy-toggle-waybar
@@ -1,7 +1,14 @@
 #!/bin/bash
 
-if pgrep -x waybar >/dev/null; then
-  pkill -x waybar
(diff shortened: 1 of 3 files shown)


dhh:
I can see the appeal here, but I don't like an implementation that needs to munge the config file. I think we're probably better off just exposing that option directly, and then folks can change it directly if they want the waybar hidden.

Alternatively, we can look into something using a state file in `.local/state/omarchy` that a launch wrapper can look at via autostart.

alexperreault:
I see your point. I'll take a crack at an implementation with a state and open another PR.

Thanks for the quick feedback!

PR #2830, opened 2025-10-25
feat: make waybar restart less jittery

alexperreault:
Currently, omarchy-restart-waybar just calls omarchy-restart-app

This kill the process and restarts it. When waybar is killed, Hyprland redraws the app to take the whole space. Waybar is then immediatly restarted, which redraws the app. This makes the top of apps jitter on a waybar restart

This PR fixes this using pkill -SIGUSR2, which tells waybar to reload the config without killing the process. This avoids the aforementioned jitter.

### DIFF
--- a/bin/omarchy-restart-waybar
+++ b/bin/omarchy-restart-waybar
@@ -1,3 +1,10 @@
 #!/bin/bash
 
-omarchy-restart-app waybar
+if pgrep -x waybar >/dev/null; then
+  # Waybar is running, reload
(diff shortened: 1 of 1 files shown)


dhh:
We used to use that, but unfortunately there's been a long-running bug with waybar where this causes stacking after walking from sleep in certain situations. So think we'll have to live with the jitter in the full restart case. But we've been using the internal signals of waybar to update specific modules.

PR #2841, opened 2025-10-25
Feature: Chromium browser safe shutdown service

brutalsam96:
### Problem
When logging out or shutting down, Chrome/Chromium shows "didn't shut down correctly" 
warnings because browsers are terminated abruptly (SIGKILL) instead of gracefully (SIGTERM).

### Solution
Added systemd user service that hooks into uwsm's `wayland-session-shutdown.target` to 
gracefully shutdown Chromium-based browsers before compositor exits.

Added service copy and enable script to migrations

### Changes
- Added `chromium-shutdown.service` to handle browser shutdown
- Added migration using omarchy-dev-migration script
- Supports Chrome, Chromium, Brave, Vivaldi, Opera
- 10-second grace period before force-kill
- Integrated with uwsm's session management

### Testing
- [x] Tested on Omarchy with uwsm 0.24.1
- [x] Chrome no longer shows crash warnings after Reboot or Log out
- [x] Browser Session state preserved correctly
- [x] Works with multiple browsers simultaneously (at least 2)
- [x] Works when browsers are closed(running in background) or when shutdown is triggered when browsers are running
- [ ] Testing with more browsers (I have only tested with chrome and chromium simultaneously)
- [x] Testing migrations script it should copy and enable the service

### Installation
- It should automatically install and enable with migrations, i managed to test in a vm.
- To manually install it need to copy chromium-shutdown.service file to `~/.config/systemd/user/` and enable and start the service 

```bash
systemctl --user enable chromium-shutdown.service
systemctl --user start chromium-shutdown.service
```

### DIFF
--- a/bin/omarchy-chromium-shutdown
+++ b/bin/omarchy-chromium-shutdown
@@ -1,22 +1,19 @@
 #!/usr/bin/bash
+echo "[$(date +%s.%N)] Script started, PID=$$"
 
-__browsers="chrome chromium brave vivaldi opera"
+__browsers="chrome|chromium|brave|vivaldi|opera|microsoft-edge|microsoft-edge-stable|microsoft-edge-beta|microsoft-edge-dev|ungoogled-chromium|iron|slimjet|yandex_browser|yandex-browser|comodo|iridium|coccoc"
 
-for __browser in $__browsers; do
-    if pgrep -x "$__browser" >/dev/null 2>&1; then
-        printf "Sending SIGTERM to %s...\n" "$__browser"
-        pkill -15 -x "$__browser" 2>/dev/null || true
-        pidwait "$__browser" || true
-        printf "%s exited.\n" "$__browser"
-    fi
-done
+__running=$(pgrep -x "$__browsers" 2>/dev/null || true)
 
-__still_running=$(pgrep -x -l -f "$(echo $__browsers | tr ' ' '|')" || true)
-
-if [ -z "$__still_running" ]; then
-    printf "All browsers have been terminated successfully.\n"
+if [ -z "$__running" ]; then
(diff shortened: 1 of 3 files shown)


brutalsam96:
Update:
Adjusted timings on the main script added 1 second delay before exitting script to account for previous instructions being executed to prevent inconsistency

brutalsam96:
UPD:
- Rewrote the service to use an external shell script instead of an inline script in the service file
- Refactored the script and it's logic, core logic now sends SIGTERM to all listed browser and checks if they are terminated if not waits 3 seconds, checks again if anything is not terminated  script exits leaving force shutdown to the session manager
- Updated migration script to copy both the service file and the shell script, installing them correctly. Script is stored in ~/.local/bin 

@dhh @ryanrhughes 
Would you mind reviewing this change when you get a chance? Thanks!

brutalsam96:
Moved to draft since inconsistentcy is still an issue, i rewrote the service but will need some more testing, i have also discovered why chrome is failing even if it is under uwsm slice will write detailed explanation and the push solution later if it passes testing.

brutalsam96:
Okay, last commit should take care of it all, for now shutdown script runs after uwsm's own cleanup meaning if browser is under uwsm's slice it might shutdown unexpectedly, moving browser out of uwsm lets my script handle the shutdown safely, this can be improved by moving the service sequence before the uwsm's termination after which will not require changing the browser startup script.

brutalsam96:
After several weeks of testing, I’m still seeing inconsistent shutdown behavior. It seems this issue needs to be handled within uwsm itself rather than by hooking into its systemd chains. I’ll investigate the uwsm code to explore a more universal fix later. Closing this PR since it doesn’t fully resolve the problem.

PR #2843, opened 2025-10-25
feat(omarchy-menu): add npm install helper and GitHub Copilot CLI option to AI menu

mghalix:
Adds `npm_install` helper for global npm packages and integrates GitHub Copilot CLI into AI tools install menu.

image

Currently Microsoft recommends installing it through npm and the package on AUR is currently outdated & not published by them, so I had to add an `npm_install` function which follows the exact pattern of `install` & `aur_install` where a window is presented with the installation process & no user input needed.

### DIFF
--- a/bin/omarchy-menu
+++ b/bin/omarchy-menu
@@ -73,6 +73,10 @@ aur_install_and_launch() {
   present_terminal "echo 'Installing $1 from AUR...'; yay -S --noconfirm $2 && setsid gtk-launch $3"
 }
 
+npm_install() {
+  present_terminal "echo 'Installing $1 via npm...'; npm install -g $2"
+}
+
 show_learn_menu() {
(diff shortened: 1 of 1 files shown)


mghalix:
Also do you think that tools that are commonly used with Coding Agents such as [`spec-kit`](https://github.com/github/spec-kit) belong here in this menu?

If so, should they be top-level alongside the currently available Coding Agents, or should they appear as a checkbox-like add-on when installing supported Coding Agents in a follow up menu?

gabectlz:
At this point, if we're transitioning to npm, we can definitely just do Gemini and Codex in npm as well (I don't know any other coders so probably Claude code as well)

dhh:
I'm not keen to do anything relying directly on npm. It wouldn't be included in the Omarchy updates. Also think we have probably enough recommendations for AI options at the moment. There'll always be more, but they're easy enough to install directly too.

PR #2850, opened 2025-10-25
Adding key bindings for screen lock and suspend to RAM

hirschnase:
I have added two new key bindings that I think are very helpful for everyone.

1. Lock the screen
SUPER ALT L  locks the screen

This feature is helpful for all people who work in an office where there is a need of locking and unlocking the machine several times a day. Having a shortcut for this should be in the standard key bindings imho.

2. Suspend the machine to RAM (aka Sleep mode)
SUPER ALT S suspends to RAM (put the machine to sleep mode)

This feature is helpful for people who'd like to put their machine to sleep mode when the work is done or when they leave the desk for a longer time period. I think this is superior to shutting the machine down, because the start up time is much quicker, when the machine needs to be used again. Also, all the open apps are there again instantly. I think this should be part of the standard key bindings as well.

### DIFF
--- a/default/hypr/bindings/utilities.conf
+++ b/default/hypr/bindings/utilities.conf
@@ -42,3 +42,8 @@ bindd = SUPER CTRL, S, Share, exec, omarchy-menu share
 # Waybar-less information
 bindd = SUPER CTRL, T, Show time, exec, notify-send "    $(date +"%A %H:%M  —  %d %B W%V %Y")"
 bindd = SUPER CTRL, B, Show battery remaining, exec, notify-send "󰁹    Battery is at $(omarchy-battery-remaining)%"
+
+# Lock Screen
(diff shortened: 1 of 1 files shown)


dhh:
I appreciate these, but I find that the system menu on `Super + Esc` does these jobs well enough. `Super + Esc, then Return` is quick enough for me for lock. But this is where it's easy enough to add your own bindings on top of that!

hirschnase:
@dhh thanks for your feedback. I did not know about the System menu - but yes, it makes sense to use it! Maybe it would be good to mention it somehow in the `keybindings` overview (SUPER K). I was searching there and found no mention of how to activate the screen lock or suspend the machine via a keyboard shortcut. Maybe other people will do a search there as well...

dhh:
Good point. We should say: System menu (lock, restart, etc). Or something like that.

Issue #1441, opened 2025-09-04
Zed fails to load on Intel GPU


ElBrodino:
Zed would not launch on a fresh Omarchy install.
I took me too much time but I found out the issue.
Zed uses Vulkan under the hood. The Vulkan plugins were simply not installed on my Intel based laptop.
to figure out that the Vulkan plugins were missing for my Intel processor.

A simple "vulkan-intel" package install fixed it

Maybe that's something that could be handled automatically when choosing Install / Editor / Zed.

dz0ny:
Same goes for AMD. `vulkan-radeon` is required to run zed.

fpigeonjr:
confirmed on 2015 macbook air, intel

kristofer:
same on MacBook Pro 2015, 15"

dhh:
If anyone wants to work on a omarchy-install-vulkan script that installs the right driver for the right graphics card, we can run that before we run the zed installer.

ElBrodino:
Trying to give it a go.
Managed to make a sh work on mine: autodetect intel GPU then install the package

> Same goes for AMD. `vulkan-radeon` is required to run zed.

hey @dz0ny, if you have an AMD, what kind of vendor does the lspci command give you?
could you kindly run this:
`lspci` and check the line that reads `VGA compatible controller`

it's a bit strange to test for hardware that I do not possess!
At least, Intel and AMD could be covered

whargrove:
@ElBrodino I am on a FW13 (latest AI350 mainboard)

```
> lspci | grep -i display
c1:00.0 Display controller: Advanced Micro Devices, Inc. [AMD/ATI] Krackan [Radeon 840M / 860M Graphics] (rev c2)
```

whargrove:
> A simple "vulkan-intel" package install fixed it

Confirm that installing `vulkan-radeon` also resolves the issue for me.

ElBrodino:
Hey guys, I didn't want this to turn out like a rabbit hole, but here we are!
_Thanks @whargrove for your support, btw! -- I'm on FW13 11th !_

I realize there a many configurations of Integrated GPU and discreet GPU out there. No wonder support if a little lacking. There's a lot of resource though!

### AMD
Help about your specific configuration is very much appreciated!
[...]

PR #2860, opened 2025-10-26
Vulkan install menu

### LINKED ISSUE #2706: Screen recording fails on hybrid Intel+NVIDIA systems due to missing VAAPI drivers – should be detected automatically
When trying to use screen recording on a fresh Omarchy installation, the recording failed silently — no video file was saved. After checking the logs, I found the following message:

`gsr_get_supported_video_codecs_vaapi: vaInitialize failed`

This error means VAAPI wasn’t initialized correctly because the proper driver was missing. And according to assada's comment, the solution depends on the GPU vendor:

> For anyone getting
> `gsr_get_supported_video_codecs_vaapi: vaInitialize failed`
> this means vaapi isn’t properly set up.
> 
> AMD: `sudo pacman -S libva-mesa-driver mesa-vdpau`
> Intel: `sudo pacman -S intel-media-driver`
> NVIDIA: `sudo pacman -S libva-nvidia-driver` 

 _Originally posted by @assada in [#2628](https://github.com/basecamp/omarchy/issues/2628#issuecomment-3424009321)_

My laptop has both an Intel integrated GPU and an NVIDIA GPU.
Even after installing libva-nvidia-driver, the issue persisted — it was only resolved after I also installed the Intel VAAPI driver (intel-media-driver).

So, the root cause seems to be that VAAPI tries to initialize the first available driver (often the wrong one) in hybrid systems. Without both drivers installed, it fails silently.

### Expected behavior:
Omarchy (or the underlying recording component) should detect when VAAPI initialization fails and show a clear message explaining the issue.

Ideally, it could suggest which driver to install based on GPU detection (via `lspci` or similar).

### Why this matters:
While not a core bug in Omarchy itself, this issue can affect many users — especially those on laptops with hybrid GPUs — right after a clean install. Improving the detection or error message would greatly improve UX and prevent confusion.

### Possible solution
A simple way to improve UX could be to add a small VAAPI check in the screen recording script before starting the recorder:

```bash
check_vaapi() {
  if ! vainfo >/dev/null 2>&1; then
    gpu=$(lspci | grep -E "VGA|3D" | awk -F: '{print $3}' | tr -d ' ')
    msg="VAAPI initialization failed. Please install the appropriate driver:\n"
    msg+="\nAMD: sudo pacman -S libva-mesa-driver mesa-vdpau"
    msg+="\nIntel: sudo pacman -S intel-media-driver"
    msg+="\nNVIDIA: sudo pacman -S libva-nvidia-driver"
    msg+="\n\nDetected GPU: ${gpu:-unknown}"
    notify-send "Screen recording error" "$msg" -u critical -t 10000
    exit 1
  fi
}

```

This would prevent the silent failure and guide the user to install the correct driver automatically.

ElBrodino:
A new script that handles GPU lookup and install Vulkan packages according to what lspci detects.
The script is launch by the Omarchy menu ( Install / Editor / Zed )
Could be repurposed to fix #2706


ElBrodino:
@ryanrhughes kindly review this humble script 😉

ElBrodino:
Addresses #1441

PR #2862, opened 2025-10-26
Omarchy 3.1.4

dhh:
# What changed?

* Add better presentation of keybindings in order of importance by @dhh
* Add `Super + S` to reveal scratchpad workspace overlay and `Super + Alt + S` to move an app there by @fldc
* Add `Ctrl + P` to print on the image viewer by @ctarx 
* Add a confirmation before starting the update process by @ryanrhughes 
* Add omarchy-keyring to install trusted signing keys by @ryanrhughes 
* Add gum confirm defaults to match the active theme by @ryanrhughes 
* Fix low-contrast issues in auto-generated Obsidian color schemes by @alexandersix
* Fix Brave dialogs to be floating like chromium by @Rishabh-Sarang
* Fix new windows should appear on top when in full-screen mode by @gyaaniguy
* Fix Surface keyboard compatibility with decryption screen by @brennancoslette
* Fix `Alt + Shift + L` hotkey for copying URL in a web app should be in list of keybindings by @manuel1618
* Fix post-install message to no longer include the message to remove USB key (not necessary) by @NeimadTL
* Fix install failure when passwords began with a "-" @dakshesh14
* Fix prevent showing sed error by mistake when encountering early issues in the installer by @ryanrhughes 
* Fix opening nvim via app launcher or file manager would actually open $EDITOR by @dhh
* Fix that the new-version check for Omarchy in the waybar should only run once every 6 hours not every hour by @dhh
* Fix slow application startups on some systems by setting systemd envs at startup by @ryanrhughes 
* Fix port 22 should not be open by default in ufw configuration by @ctarxctarx 
* Fix aliases for cd/ls should only be set if eza/zoxide is still on the system by @wiehann 
* Fix time should be synced before doing an Omarchy update by @dhh 
* Fix Android Studio window rules so dialogs are clickable by @AdamMusa
* Fix Ghostty default shell config to respect the block cursor by @jwkicklighter 
* Fix fcitx5 clipboard interfering with hotkeys in other apps by turning it off by @cairin

### DIFF
--- a/config/fcitx5/conf/clipboard.conf
+++ b/config/fcitx5/conf/clipboard.conf
@@ -1,2 +1,2 @@
-[Addon]
-Enabled=False
+TriggerKey=
+PastePrimaryKey=


applications/nvim.desktop
```diff
@@ -2,7 +2,7 @@
 Name=Neovim
 GenericName=Text Editor
 Comment=Edit text files
-Exec=omarchy-launch-editor %F
+Exec=$TERMINAL -e nvim %F
```
dharmavagabond: @dhh, this will reintroduce the bug with nautilus since its an invalid Desktop Entry, the env vars don't get expanded there (you can corroborate with `desktop-file-validate`). I really think omarchy should get rid of this entry since there is already one at `/usr/share/applications` or copy the one already mentioned, specifically:
```
Exec=nvim %F
Terminal=true
```

ryanrhughes: Resolved this via https://github.com/basecamp/omarchy/pull/2862/commits/ba348a6a15724f65d980539381e7e6e540adb871

After trying a few solutions and validating with folks in Discord, this feels like the best option that solves for all issues.

tomycostantino:
Looks awesome, just a quick question: How do I remove an app from the scratchpad, or that could be in the next version? @dhh

tomycostantino:
> Looks awesome, just a quick question: How do I remove an app from the scratchpad, or that could be in the next version? @dhh

All good, for those finding the same issue, just do `super` + `shift` + `workspace #`, it's like moving it to a workspace.

naxels:
Just tried the Scratchpad, interesting addition! Please update the manual to reflect this feature :)

PR #2872, opened 2025-10-26
imv: add default Ctrl+P binding to print current image via CUPS

ctarx:
- Add a default key binding for `imv` mapping <Ctrl+P> to printing the current image via `lp`.
- Matches common UX expectations for printing (Ctrl+P).


dhh:
Needs a migration to copy it over on existing installations 👍

PR #2887, opened 2025-10-26
Remove default SSH allow rule from UFW

ctarx:
### Summary
Refines the default UFW configuration for safer defaults without losing functionality.

### Changes
~~- Limit LocalSend (port 53317, TCP/UDP) to the local LAN instead of allowing from anywhere.~~
- Remove the default UFW rule that opens port 22 (SSH).

### DIFF
--- a/install/first-run/firewall.sh
+++ b/install/first-run/firewall.sh
@@ -1,27 +1,10 @@
-#!/bin/bash
-# Exit immediately if a command exits with a non-zero status
-set -eEo pipefail
-
-# Detect local network CIDR via routing table
-IFACE=$(ip route show default | awk '{print $5; exit}')
-if [[ -z "${IFACE:-}" ]]; then
-  echo "ERROR: Cannot detect default network interface." >&2
-  exit 1
-fi
-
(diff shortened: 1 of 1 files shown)


install/first-run/firewall.sh
```diff
@@ -1,13 +1,27 @@
...
 
-# Allow ports for LocalSend
-sudo ufw allow 53317/udp
-sudo ufw allow 53317/tcp
-
-# Allow SSH in
-sudo ufw allow 22/tcp
+# Allow ports for LocalSend — only from local network
+sudo ufw allow from "${NETWORK}" to any port 53317 proto udp comment 'LocalSend UDP (LAN only)'
+sudo ufw allow from "${NETWORK}" to any port 53317 proto tcp comment 'LocalSend TCP (LAN only)'
```
dhh: I don't think this is going to work if it hardcodes the current local network. I'd expect LocalSend to work both at home, at work, and at a friend's house.

But I'm fine closing off SSH by default.

ctarx: Makes sense. I’ll drop the LocalSend changes for now since I didn’t account for switching networks. I’ll update this PR so it only removes the default SSH allow rule. Thanks for pointing it out!

Issue #2878, opened 2025-10-26
On first install, adjust USB removal phrasing


ElBrodino:
### What do you need?

When the first install is completed, the user is notified to remove the USB thumb drive before rebooting the system.
If the user removes the thumb drive **before** rebooting as instructed, the temporary OS will throw a lot of errors on screen.
Maybe the OS needs rebooting instructions from the thumb drive which is no longer present.

_Suggestion_: Rephrase the message to : "Remove the thumb drive before the system boots **but not before the system has completed the shutdown sequence**."
If that is a bit hard to write or execute, maybe suggest a shutdown-removal-Boot sequence : 

1. Please shutdown system
2. Remove USB thumb drive
3. Turn on system

ElBrodino:
After a couple of omarchy installs, that one bit me a couple of times already! 
I've seen it happen on Youtube tutorials as well !!

NeimadTL:
@ElBrodino, thanks for your feedback. How about "Remember to remove USB installer after the has completely shutdown!" ?

@dhh, @ryanrhughes, any thoughts would be appreciated on this, thank you.

ryanrhughes:
I'm not super particular on it. In almost all cases, our EFI entries will get positioned above the USB so we honestly may not even need the notice.

ElBrodino:
Oh then that's why, I was not aware the order got changed. I might want to keep the EFI boot priority, what do you think?

ryanrhughes:
It's just how the install works. It adds an entry which on most machines floats to the top, however if you've set an order yourself, that usually takes priority. It varies from machine to machine.

NeimadTL:
Thanks @ryanrhughes. Should we consider closing this issue @ElBrodino ? so that we and others know that it's not something to be done.

ElBrodino:
@NeimadTL tahnks for your keen interest!! I believe something still needs to be done to avoid the wall of errors just before the first joys of exploring omarchy.
With what @ryanrhughes said about most machines, the removal of the text sits well with me !

gxaurab:
i too messed up at this part. nice issue raised

dhh:
We just removed it.

PR #2888, opened 2025-10-26
Remove after install usb installer message

NeimadTL:
Since in almost all cases, the EFI entries will get positioned above the USB, this notice is not needed.

Enhancement for #2878


NeimadTL:
Thank you for the review @dhh

PR #2891, opened 2025-10-27
Chromium URL Handler

HowieDuhzit:
Add Chromium URL Handler for installing WebApps from a website such as Omarchy.Apps

### DIFF
--- a/bin/omarchy-install-handler
+++ b/bin/omarchy-install-handler
@@ -1,86 +1,48 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
-# Usage check
-if [[ $# -lt 1 ]]; then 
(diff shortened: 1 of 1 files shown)


HowieDuhzit:
Changed to base to dev for pulling and testing

dhh:
What is this Omarchy.Apps website you're referring to that uses this URL format?

HowieDuhzit:
It's a website I'm working on that us a repo of WebApps that we can one click install using the handler.

Users can add new apps the same way they are added with the webapp installer and they get saved to the site and then enable one click install for everyone else after.

Right now its only setup for webapps but I plan to extended to TUIs.

https://omarchy.app/

image

dhh:
I think this is interesting, but going to skip for now. Thanks for exploring it!

Issue #2875, opened 2025-10-26
Walker clipboard history stops working on copying screenshot


andnig:
### System details

Reproduced on AMD AI MAX 395+ and a 2nd PC with AMD Ryzen 9 5900 + NVIDIA 4080; Omarchy 3.1.3

### What's wrong?

If you add a screenshot to the clipboard, the walker clipboard history "dies" in the sense of that no text can be added after the first image in the history. Copy/Pasting still works, but the copied text is not added to the history.

### Reproduction

1. Add text to the clipboard (works as expected)
2. Copy a picture (make a screenshot to clipboard) --> This image is still added to the clipboard history as expected
3. Now copy some text --> This text is NOT added to the clipboard history anymore.

To fix the clipboard history, it's enough to restart walker (or I guess just elephant as some users in discord reported).

I'm happy to provide some logs, just let me know from what.

abenz1267:
Can't reproduce this at all. What version are you using?

andnig:
Elephant  2.6.1-1 and Walker  2.6.2-1.

I just had a heart attack because I myself couldn't reproduce it, but then I found an additional data point: I needed to add considerable amount (sorry I don't know the exact number) of text copies first (I basically copied random text from this GH issue I guess about 30 times).

Then I used the omarchy screenshot utility (where it opens satty). There I pressed ctrl + c to get the image to the clipboard. The image was added, and now no new text is added to the clipboard history (but the clipboard itself works).

So:


    1. Add text to the clipboard (but there needs to be a considerable amount of text in the clipboard history, so ctrl+c 30 times or so) (works as expected)
    2. Copy a picture (make a screenshot to clipboard (satty with ctrl + c)) --> This image is still added to the clipboard history as expected
    3. Now copy some text --> This text is NOT added to the clipboard history anymore.

abenz1267:
Still can't reproduce that. Anyone else?

abenz1267:
Do you get any error? `journalctl --user -f | grep elephant`

andnig:
No errors, no.
These are the only journalctl results for the two sessions. In both the history didn't receive new elements, after copying an image.
```
Oct 26 16:17:30 archlinux uwsm_app-daemon[3884]: received: app -- elephant
[...]

PR #2894, opened 2025-10-27
Fix: Finish child wl-paste processes when restarting walker 

andnig:
Elephant spawns wl-paste for images and text. When we simply pkill
elephant, these childs stay running, meaning when we restart walker
multiple times we end up having multiple instances of dangling wl-paste
processes. Using systemd to stop the services allows elaphant to close
these.

This was recognized by @abenz1267 as part of #2875 .

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.yml
+++ b/.github/ISSUE_TEMPLATE/bug.yml
@@ -1,5 +1,5 @@
 name: Bug
-description: Report a problem
+description: Report a validated bug -- NOT FOR SUPPORT REQUESTS
 labels: [bug]
 body:
   - type: markdown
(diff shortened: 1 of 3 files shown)


abenz1267:
i'm not 100% sure, but i think @ryanrhughes wanted to change how elephant/walker are started anyways so this might be changed anyways.

I think the plan was to use an actual user-level systemd service for elephant. So restarting would simply be `systemctl --user restart elephant.service`.

andnig:
Ah I see, this would make things easier then. Thanks for clarifying! :+1: 
I'll leave the PR open for now, but check in later and close it, if none of the maintainers did.

ryanrhughes:
I do want to try to move over to that but need to do some testing with it to make sure that's smooth. Let's leave this as an alternative for now just in case we find an issue with that.

PR #2934, opened 2025-10-28
Add fullscreen border color support to all themes

dcadenas:
This PR adds distinct border colors for fullscreen/maximized windows (SUPER+ALT+F) to provide visual feedback when windows are covering others.

## Changes
- Added `$fullscreenBorderColor` variable to all 12 themes
- Added `windowrulev2 = bordercolor $fullscreenBorderColor, fullscreen:1` rule
- Each theme uses high-contrast accent colors from its palette
- Tokyo-night maintains gradient style (pink-to-orange vs cyan-to-green)

## Rationale
When a window is maximized, it can be hard to tell if other windows are underneath. The distinct border color provides immediate visual feedback that the window is in fullscreen mode.

## Testing
Tested across multiple themes by toggling SUPER+ALT+F. Each theme shows obvious color contrast between normal and fullscreen borders.

### DIFF
--- a/default/hypr/looknfeel.conf
+++ b/default/hypr/looknfeel.conf
@@ -3,6 +3,7 @@
 # Variables
 $activeBorderColor = rgba(33ccffee) rgba(00ff99ee) 45deg
 $inactiveBorderColor = rgba(595959aa)
+$fullscreenBorderColor = rgba(ffd700ee)
 
 # https://wiki.hyprland.org/Configuring/Variables/#general
 general {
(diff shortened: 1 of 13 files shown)


dhh:
Appreciate the suggestion, but I don't think it's worth the visual distraction. I prefer the clean full screen view. We've also recently just added the rule that if you start another app while you're in full screen, you'll get it on top. That's enough.

dcadenas:
Thanks for considering! Totally understand preferring the clean look.

One thing: the auto-bring-to-top only helps when opening new apps. Coming back to a workspace later, super+ALT+F maximized windows look identical to single-window workspaces - no visual hint. (Super+F is different as it removes borders entirely.) This just colors the existing border.

For future readers, easy to add manually to `looknfeel.conf` if wanted:

```
windowrulev2 = bordercolor rgb(YOUR_COLOR_HERE), fullscreen:1
```

dcadenas:
**Update for Hyprland 0.50+**: The window rule syntax has changed. If you're on a recent version, use:

```conf
# In ~/.config/hypr/looknfeel.conf
windowrule = border_color rgb(ff5500), match:fullscreen_state_internal 1
```

Key changes from the original PR:
- `windowrulev2` → `windowrule`
- `bordercolor` → `border_color` (underscore)
- `fullscreen:1` → `match:fullscreen_state_internal 1`

Works with Super+Alt+F (maximize). Adjust the color to match your theme.

PR #2935, opened 2025-10-28
Closes #2911: Added automatic theme-switcher

### LINKED ISSUE #2911: Suggestion: Automatic Theme Switching Feature
### What do you need?

Hi Omarchy team! 👋

This is my first issue on an open-source project, and I wanted to share an idea for improving Omarchy:

I created a simple automatic theme switcher using the existing Omarchy commands:

- omarchy-theme-list
- omarchy-theme-next
- omarchy-theme-bg-next

The script allows users to:

- Automatically switch to the next theme or next background.
- Optionally select a random theme.
- Configure intervals (minute, hour, day, week, month).

I also set up systemd service and timer files so the theme switching can run automatically in the background.

**Motivation:**

Omarchy already makes it easy to switch themes manually, which is great. However, I think most users would enjoy opening their PC to a fresh theme automatically, like every day or once a week, without having to run commands themselves. This small addition could make the user experience more dynamic and fun.

I’m open to renaming the files or moving them if they don’t fit the current Omarchy structure or naming conventions.

Thank you for considering this feature.I hope it can be useful for other Omarchy users!

LukasTrust:
This is my first pull request to the project. 

Summary of changes:
-  Added an automatic theme-switcher as described in issue #2911.

If you need further changes or explanations, please let me know. 

Related Issue: https://github.com/basecamp/omarchy/issues/2911

### DIFF
--- a/bin/omarchy-theme-switcher
+++ b/bin/omarchy-theme-switcher
@@ -1,0 +1,111 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+CONFIG="$HOME/.config/hypr/theme-switcher.conf"
+TIMER_PATH="$HOME/.config/systemd/user/omarchy-theme-switcher.timer"
+
+# --- Check for config file ---
+if [[ ! -f "$CONFIG" ]]; then
+    echo "No config file found at $CONFIG"
+    exit 1
+fi
+
+# --- Load config values safely ---
+mode=""
+interval=""
+while IFS='=' read -r key value; do
+    key=$(echo "$key" | xargs)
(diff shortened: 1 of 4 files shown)


dhh:
I appreciate the suggestion, but I don't love the amount of code required here. And I'm skeptical of the feature as is. I think this could make a good separate package for those who want that, though. Like a omarchy-timed-theme-switcher package on AUR ✌️

LukasTrust:
Thanks for the feedback, i might look into making it into a separate package!

PR #2936, opened 2025-10-28
check the drives filesystem and only use snapper on btrfs

ecruzolivera:
Hi, in my setup I have two drives, one `btrfs`  for root and another `ext4` for `home`, the current `omarchy-snapshot` script assumes that all drives are Btrfs.

### DIFF
--- a/bin/omarchy-snapshot
+++ b/bin/omarchy-snapshot
@@ -20,12 +20,14 @@ create)
 
   echo -e "\e[32mCreate system snapshot\e[0m"
 
-  # Get existing snapper config names from CSV output
-  mapfile -t CONFIGS < <(sudo snapper --csvout list-configs | awk -F, 'NR>1 {print $1}')
-
-  for config in "${CONFIGS[@]}"; do
-    sudo snapper -c "$config" create -c number -d "$DESC"
-  done
+  # Read snapper configs and their subvolume paths
+  while IFS=, read -r name path _; do
+    [[ $name == "Config" ]] && continue # Skip header
+    # Check if the path exists and is btrfs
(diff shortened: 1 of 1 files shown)


ecruzolivera:
Hi @ryanrhughes, just a friendly reminder of this.

ryanrhughes:
@ecruzolivera What's the issue this solves exactly? If you don't have btrfs for a drive, it won't have a snapper config so it'd already be ignored unless I'm misunderstanding.

ecruzolivera:
> @ecruzolivera What's the issue this solves exactly? If you don't have btrfs for a drive, it won't have a snapper config so it'd already be ignored unless I'm misunderstanding.

Hello, this script will limit snapper to only be called on btrfs partitions.

I have two drives, one for `/root` that has btrfs and another for `/home` that it is ext4. Without this script, the snapper [script](https://github.com/basecamp/omarchy/blob/7c0adf7d40206bd1b49305a2f470c0be300f99c2/bin/omarchy-snapshot#L24C26-L24C88) will report all partitions to be snapped including the ext4 ones.

```
sudo snapper --csvout list-configs | awk -F, 'NR>1 {print $1}'
home
root
```
And therefore the script will fail when trying to [snap](https://github.com/basecamp/omarchy/blob/7c0adf7d40206bd1b49305a2f470c0be300f99c2/bin/omarchy-snapshot#L26C3-L28C7) an ext4 partition.

My script will check if the partition is Btrfs before trying to snap it

```
 if [[ -d $path ]] && [[ "$(findmnt -n -o FSTYPE --target "$path")" == "btrfs" ]]; then
      sudo snapper -c "$name" create -c number -d "$DESC"
 fi
```

Please let me know if I was able to explain myself properly.

PR #2937, opened 2025-10-28
fix: improve catppuccin-latte group background colors

MangelMaxime:
Hello,

In this PR, I am trying to improve catpuccin-late theme for Hyprland groups:

- Improves contrast
- Make the colors consistent with the theme

**Before**

image

See how the background colors of the group is not from catpuccin-late color palettes and also that the inactive group text is almost unreadable.

**After**

image

### DIFF
--- a/themes/catppuccin-latte/hyprland.conf
+++ b/themes/catppuccin-latte/hyprland.conf
@@ -1,4 +1,7 @@
 $activeBorderColor = rgb(1e66f5)
+$text = rgb(4c4f69)
+$surface0 = rgb(ccd0da)
+$base = rgb(eff1f5)
 
 general {
     col.active_border = $activeBorderColor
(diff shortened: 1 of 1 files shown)


dhh:
I agree that the contrast is a little weak, but I don't love the suggested fix. We still need to retain some translucency to match with all the other themes. Whatever the fix, it has to be general. Feel free to take another stab at that if you can find something that's generic and that works with all the themes.

MangelMaxime:
When creating this PR, I though about the issue for the other themes.

From my POV, it is all the themes that need to customize their `groupbar` styling to match their color palette. What do you think?

dhh:
I think you'll get the color match just by retraining some translucency. We probably just have to tweak the contrast/text color a bit.

PR #2909, opened 2025-10-27
Switch to NetworkManager

kromsam:
Closes #1414.

The only thing that stood in the way for a switch to NetworkManager was a good TUI. Thanks to the work of @Zeus-Deus, we now have a viable replacement for Impala: [Gazelle](https://github.com/Zeus-Deus/gazelle-tui). AUR: https://aur.archlinux.org/packages/gazelle-tui

This PR sets the basis for a move towards NetworkManager. It makes sure we are all on the same page and we can start to test the configuration.

In this proposal I stay as close to the current Omarchy set-up as possible. Which means:

* We stop using systemd-networkd, and start using NetworkManager instead
* We replace Impala with Gazelle
* We keep IWD as the backend for WiFi. I have tested this, and together with NetworkManager this also works for Enterprise setups like eduroam.
* We keep using systemd-resolved for DNS resolution, as Omarchy already uses it: https://github.com/basecamp/omarchy/blob/7c0adf7d40206bd1b49305a2f470c0be300f99c2/bin/omarchy-setup-dns#L83

The only thing that will really change for the user is that there will be a new Network TUI. The introduction of NetworkManager will not be noticeable for most users.

It might come at the cost of having to recreate a connection for some networks, we will have to test this.

## Why NetworkManager?
See issue #1414 . Systemd-networkd lacks usability when it comes to VPN-connections and Enterprise (802.1x) Networks such as eduroam. NetworkManager is the de facto standard for Linux on the Desktop. It comes with practically every Desktop Environment, and therefore has a big user base and functionality for desktop users. Systemd-neworkd is more meant for use on systems where networking is a matter of set-and-forget, like a server. https://wiki.archlinux.org/title/NetworkManager

## Why stick with IWD, and not use wpa_supplicant?
Omarchy already uses IWD. IWD is the modern replacement of wpa_supplicant. This is the way linux is heading, it is a mature project (in version 3.10), and it has become a proper replacement. I have tested IWD and had no issues with logging in to Enterprise Networks or connection to VPNs. https://wiki.archlinux.org/title/Iwd

## How do I test this?
The instructions at https://github.com/basecamp/omarchy/issues/1414#issuecomment-4811643281 replicate the exact steps of this PR.

## To Do's
- [ ] Add Gazelle to Omarchy Packages https://github.com/omacom-io/omarchy-pkgs
- [ ] Test what happens with existing connections when migrating from networkd to nm

PR #2939, opened 2025-10-28
Add automatic detection of an eduroam network

jharlan-hash:
Hello! This is my first pull request, so let me know what I can do to improve this, or if it's completely unnecessary. 

I noticed that connecting to eduroam (the default wifi for many institutions worldwide) was unnecessarily difficult to do on Linux. Therefore, I added this small script to the wifi.sh file to detect if eduroam is available and suggest an installation method. 

I first detect the wireless interface using `iwctl`, display an error if none is found, and scan for available networks.
If the "eduroam" network is found, the script installs the `geteduroam` package, prompts the user to launch the authentication GUI, and removes the package if the user dismisses the prompt.

Thank you for taking the time to look at this, and let me know what we can change!

Thank you,
Jack Sovern

### DIFF
--- a/install/first-run/wifi.sh
+++ b/install/first-run/wifi.sh
@@ -1,3 +1,24 @@
 if ! ping -c3 -W1 1.1.1.1 >/dev/null 2>&1; then
+
+  INTERFACE=$(iwctl station list | tail -n +5 | awk '{print $2}' | tr -d " \t\n\r")
+  if [[ -z "$INTERFACE" ]]; then
+      notify-send "Error" "No wireless interface found" -u critical
+      exit 1
+  fi
+
+  iwctl station $INTERFACE scan
+  if iwctl station $INTERFACE get-networks | grep -qF "eduroam"; then
+      if ! yay -S --noconfirm geteduroam; then
+          notify-send "Error" "Failed to install geteduroam package" -u critical
+          exit 1
+      fi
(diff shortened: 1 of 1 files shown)


AndrewMast:
If this gets rejected, maybe it can be made into a script called `omarchy-setup-eduroam` or something like that. Seems useful, but maybe not universally applicable.

jharlan-hash:
> If this gets rejected, maybe it can be made into a script called `omarchy-setup-eduroam` or something like that. Seems useful, but maybe not universally applicable.

Yes, this totally makes sense. I wonder what percentage of people using Omarchy actually have to deal with eduroam?

This would keep everything simpler, too.

AndrewMast:
Also, it appears that #2909 can also fix the eduroam issue by using NetworkManager instead of systemd-networkd. Maybe wait on this to see where that PR goes?

Zeus-Deus:
FYI - Gazelle already handles eduroam (full 802.1X support). No extra tools needed.

I documented how to switch to NetworkManager + Gazelle here: https://github.com/basecamp/omarchy/issues/1414#issuecomment-3411454352

Works with PEAP/MSCHAPv2 (standard eduroam setup).

jharlan-hash:
Sounds good! I'll close this pull request and let's see if we can get #2909 and #1414 merged!

PR #2940, opened 2025-10-28
Update utilities.conf with keyboard shortcuts for zooming in and out on the display

pelephant2:
`Super+Shift+Z` to zoom in on the display, and `Super+Shift+Alt+Z` to return the display to the default 100% zoom.


dhh:
Doesn't work as presented:

image

PR #2945, opened 2025-10-28
Add optional biology/bioinformatics package

mmorri:
## Summary

Adds comprehensive biology and bioinformatics tooling as a standalone, opt-in package that can be installed with a single command:

```bash
omarchy-install-biology
```

This package includes:

**System packages** (via pacman):
- Genomics tools: samtools, bcftools, bedtools, bowtie2, bwa, minimap2
- Quality control: fastqc, trimmomatic, cutadapt
- Phylogenetics: mafft, muscle, clustalw
- Visualization: igv, pymol
- Utilities: parallel, bbmap, compression tools

**Python packages** (via uv):
- Scientific: numpy, scipy, pandas, matplotlib, seaborn
- Biology: biopython, pysam, scikit-bio, scanpy, napari
- Interactive: jupyter, jupyterlab

**R packages** (user-level):
- Tidyverse ecosystem
- Bioconductor: Biostrings, GenomicRanges, DESeq2, edgeR, limma
- Analysis: Seurat, phyloseq, SingleCellExperiment

## Implementation

- Creates `omarchy-install-biology` command for easy installation
- Automatically installs Python and R dev environments if needed
- All installation scripts check for dependencies and install them via `omarchy-install-dev-env`
- Adds R support to `omarchy-install-dev-env` for standalone R installation
- Completely separate from default installation - users must explicitly opt-in

See `docs/BIOLOGY.md` for complete documentation.

### DIFF
--- a/bin/omarchy-install-biology
+++ b/bin/omarchy-install-biology
@@ -8,11 +8,14 @@ echo "  - Python packages (biopython, scanpy, etc.)"
 echo "  - R packages (Bioconductor, tidyverse, etc.)"
 echo ""
 
+# Set OMARCHY_INSTALL environment variable
+export OMARCHY_INSTALL="$HOME/.local/share/omarchy/install"
+
 # Source the helpers
-source ~/.local/share/omarchy/install/helpers/all.sh
+source $OMARCHY_INSTALL/helpers/all.sh
 
 # Run the biology installation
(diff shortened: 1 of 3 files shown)


dhh:
I think we had a misunderstanding here. I was suggesting that you added a separate package that you can post to the AUR and then folks can install it via that. I don't see that we have something like this directly in Omarchy. Sorry for the confusion.

PR #2946, opened 2025-10-28
Fix steam install steam

mmorri:


dhh:
We already have multilib turned on by default in Omarchy. I don't know how you ended up without that.

PR #2954, opened 2025-10-29
Feature (hypr): add “Pop window” (float+pin) on SUPER+O

ryrobes:
## What & Why
Adds a “Pop window” toggle bound to **SUPER+L**:
- Floats the focused window (if tiled), pins it across all workspaces, and tags it `pop`.
- Toggling again unpins it and re-tiles it.
- Tag-driven styling adds rounded borders for active `pop` windows.
- Good for YouTube videos or any app you want to stay put on the monitor when you have lots of workspaces, etc.

https://github.com/user-attachments/assets/e0cebd8c-f5da-4430-9314-a1c742e162db

## Testing
- Hyprland: verified pop/un-pop across workspaces, floating state restored correctly (`pop` tag removed).
- No script deps beyond `hyprctl` and `jq`.

### DIFF
--- a/default/hypr/bindings/tiling-v2.conf
+++ b/default/hypr/bindings/tiling-v2.conf
@@ -9,7 +9,7 @@ bindd = SUPER, T, Toggle window floating/tiling, togglefloating,
 bindd = SUPER, F, Full screen, fullscreen, 0
 bindd = SUPER CTRL, F, Tiled full screen, fullscreenstate, 0 2
 bindd = SUPER ALT, F, Full width, fullscreen, 1
-bindd = SUPER, L, Pop window (float + pin), exec, omarchy-cmd-pop
+bindd = SUPER, O, Pop window out (float & pin), exec, omarchy-cmd-pop
 
 # Move focus with SUPER + arrow keys
 bindd = SUPER, LEFT, Move window focus left, movefocus, l


jkoestinger:
Just saw this PR and wanted to give a heads up: I think quite some people are mapping SUPER + L to lock the screen, because that is what is getting sent by the "lock" key on keyboards that have it usually (funnily, this would turn the lock screen key into a pop windows key). Maybe it would make sense to use another binding for it?

ryrobes:
Hah, fair point. Any suggestions on something unused and also isn't frequently re-bound by people? SUPER + O maybe?

imvaskii:
> Hah, fair point. Any suggestions on something unused and also isn't frequently re-bound by people? SUPER + O maybe?

`SUPER SHIFT, ESCAPE` maybe?

madebynoxc:
If `SUPER+T` toggles floating, it would make sense to float and pin it with `SUPER+SHIFT+T`

ryrobes:
Would prefer to keep it on a SUPER only keybind since it's a top-level base tile/system op? Personally, `SUPER+O` (for pOp) seems to make sense?

But if people really want SUPER+SHIFT+T, then so be it - let me know.

**Spicy take:** I honestly feel that it's more _broadly_ useful IMHO than Pseudo (`SUPER+P`), so locally I might make **it** `SUPER+P` and make PseudO `SUPER+O` - but that'd be a breaking change for the release.

ryrobes:
Changed to `SUPER-O` for now, follows the new keybind scheme better w/o an extra modifier. 
Works great with the default `SUPER` + left/right drag for move and resize. I use it for YT, Zoom, GoogleMeet, Messenger windows daily when I want a really flexible window that persists and doesn't mutate my tiles across workspaces.

dhh:
We already have picture-in-picture mode that does this automatically for video pop outs on YT etc. Could we try to reuse that with a trigger for tagging?

dhh:
The file is in default/hypr/apps/pip.conf

ryrobes:
Good idea, but after experimenting a bit I think they are likely better as 2 sep things since they serve 2 distinct use cases. Pip being special / specific side-effect and "Pop" being a generalized user-triggered window mutation useful for arbitrary tiles. Thoughts?

dhh:
I think the windowrules are the same though, no? Or at least should share a similar base. Just that I'd like to get rid of the cmd-pop. And just rely on adding/removing a tag to the active window, which is then ruled by static rules.

ryrobes:
Would love to be able to piggy back the existing pip.conf rules. The problem is the important rules for pip aren't _reactive_ on existing windows (only dynamic rules are noborder, opacity, etc do that). So they have to be dispatched explicitly in order to work (unfortunately). So while we _can_ dynamically change the tags on an existing window, hyprland it won't cascade & re-eval rules based on that new tag.

https://wiki.hypr.land/Configuring/Window-Rules/#static-rules
https://wiki.hypr.land/Configuring/Window-Rules/#dynamic-rules

Kind of a bummer, honestly. If I can find a workaround that's not ugly, I'll remedy - but for now it seems like this kind of thing needs it's own little script call.

PR #2955, opened 2025-10-29
Add missing hyprlock PAM configuration

### LINKED ISSUE #2094: Authentication failure via lock screen after some X amount of time
### System details

AMD 7 5800X, RTX 3070Ti, Omarchy v3.0.2

### What's wrong?

I've ran into this issue a handful of times where if I lock the screen and come back an hour or so later I am unable to log in and get the `Authentication failed` message.

related to nvda?

Image

d-cas:
Fixes #2094, #1469

## Problem
Hyprlock fails to authenticate after extended lock periods because `/etc/pam.d/hyprlock` doesn't exist in the current Arch base image. Users get "Authentication Failed" even with correct password and must reboot.

## Root Cause
Upstream Arch recently added `/etc/pam.d/hyprlock` to their hyprlock package (0.8.2-1), but Omarchy's current base predates this addition. Without a PAM policy, hyprlock doesn't know how to authenticate and fails silently.

## Solution
Add the missing PAM configuration file now, rather than waiting for the next Arch rebase. This provides:
- Immediate fix for existing users via migration script
- Proper PAM policy for new installations
- Forward compatibility (when Omarchy rebases to newer Arch, this config will already match upstream)

## Implementation
- **New installs:** Install script copies PAM config during setup
- **Existing users:** Migration script adds/repairs config on upgrade
- **Idempotent:** Safe to run multiple times, handles incomplete configs

## Testing
- Solution discovered and tested by @ItsMick in #2094
- Confirmed working by multiple users in the issue thread
- Migration script tested successfully on Omarchy 3.1.3 dev branch

### DIFF
--- a/default/pam.d/hyprlock
+++ b/default/pam.d/hyprlock
@@ -1,0 +1,5 @@
+# Omarchy PAM fix for Hyprlock (Arch-based)
+auth       include        system-local-login
+account    include        system-local-login
+password   include        system-local-login
+session    include        system-local-login
--- a/install/config/all.sh
+++ b/install/config/all.sh
@@ -3,6 +3,7 @@ run_logged $OMARCHY_INSTALL/config/theme.sh
 run_logged $OMARCHY_INSTALL/config/branding.sh
 run_logged $OMARCHY_INSTALL/config/git.sh
 run_logged $OMARCHY_INSTALL/config/gpg.sh
+run_logged $OMARCHY_INSTALL/config/hyprlock-pam.sh
 run_logged $OMARCHY_INSTALL/config/timezones.sh
 run_logged $OMARCHY_INSTALL/config/increase-sudo-tries.sh
 run_logged $OMARCHY_INSTALL/config/increase-lockout-limit.sh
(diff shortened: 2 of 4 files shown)


dhh:
If this is fixed in hyprlock 0.8.2+, we should have the correct package in both the latest ISO and when you update. I see that extra/hyprlock is already at 0.9.2. Why doesn't this fix it then?

d-cas:
You're totally right. I just checked and Arch's config includes `/etc/pam.d/login`, which itself includes `system-local-login` for all the same PAM modules my version has (plus `pam_nologin.so`).

They're functionally equivalent. The package already has the correct fix.

Closing this PR as unnecessary. Thanks for pushing me to verify.

dhh:
But may we still have a case where this hasn't been corrected for old systems? Or does the simple fact of upgrading hyprlock to 0.9.2 sort it out?

d-cas:
Yes, upgrading hyprlock automatically creates the file. Verified two ways:

1. Package ownership: `pacman -Qo /etc/pam.d/hyprlock` shows it's owned by hyprlock 0.9.2-3
2. Deleted the file, ran `pacman -S --overwrite '*' hyprlock`, file was recreated

My mistake: Saw the one-line config (`auth include login`) and didn't realize it chains to `/etc/pam.d/login`. Thought it looked incomplete.

No migration needed.

PR #2961, opened 2025-10-29
Replace swapwindow with movewindow

imvaskii:
~Move active windows around just by using keyboard.~

~We can also move windows with SUPER + left mouse click but having keyboard shortcuts could be more convenient?~

`swapwindow` is limiting when the target direction of workspace does not have a window to swap with. Functionality wise `movewindow` is similar to `swapwindow` with extra functionality to be able to move active window to active workspaces.

### DIFF
--- a/default/hypr/bindings/tiling-v2.conf
+++ b/default/hypr/bindings/tiling-v2.conf
@@ -16,12 +16,6 @@ bindd = SUPER, RIGHT, Move window focus right, movefocus, r
 bindd = SUPER, UP, Move window focus up, movefocus, u
 bindd = SUPER, DOWN, Move window focus down, movefocus, d
 
-# Move active windows SUPER + CTRL arrow keys
-bindd = SUPER CTRL, LEFT, Move window focus left, movewindow, l
-bindd = SUPER CTRL, RIGHT, Move window focus right, movewindow, r
-bindd = SUPER CTRL, UP, Move window focus up, movewindow, u
-bindd = SUPER CTRL, DOWN, Move window focus down, movewindow, d
-
 # Switch workspaces with SUPER + [1-9]
 bindd = SUPER, code:10, Switch to workspace 1, workspace, 1
 bindd = SUPER, code:11, Switch to workspace 2, workspace, 2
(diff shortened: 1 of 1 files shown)


joshuafouch:
I am pretty sure you can just do this with SUPER + arrow keys?

imvaskii:
> I am pretty sure you can just do this with SUPER + arrow keys?

Isn't that for just focusing windows? this change is for moving the window around.

onlyjuans:
I think you could do it already using 

> **SUPER + Shift + arrow keys**

Michallote:
Yes, this is already implemented with SUPER + Shift + Arrows

imvaskii:
@onlyjuans and @Michallote the `SUPER + SHIFT + Arrow Keys` is implemented for [swapping windows](https://github.com/basecamp/omarchy/blob/master/default/hypr/bindings/tiling-v2.conf#L50-L54). This PR is for moving active windows around.

dhh:
What's the functional difference between this and swapping? It looks like the same behavior?

imvaskii:
> What's the functional difference between this and swapping? It looks like the same behavior?

for swapping windows target direction should have an window to swap with but for moving it does not. Very simple use case is say if you have external monitor connected and lets say it is aligned on top of your primary screen and has no windows. and you want to move active window from your primary screen to this ext display, you'd just need to do `SUPER CTRL, UP arrow` or the `hyprctl dispatch movewindow u`

dhh:
I see. Maybe we should actually just replace the swaps with the move, then. Because they're functionally the same on single monitors, but then there's extra functionality on multi-monitor. I don't think we need this subtlety difference otherwise.

imvaskii:
> I see. Maybe we should actually just replace the swaps with the move, then. Because they're functionally the same on single monitors, but then there's extra functionality on multi-monitor. I don't think we need this subtlety difference otherwise.

I agree! we can just simply replace `swapwindow` with `movewindow` [here](https://github.com/basecamp/omarchy/blob/master/default/hypr/bindings/tiling-v2.conf#L50-L54) then?

dhh:
Yes, let's try that.

dhh:
Unfortunately the behavior is actually not quite the same. If you have a layout with 2 / 1 / 2 apps in columns, using move on the one in the middle will do a swap with the one on the left in the existing setup, and move it out of the middle column and into the left one with the new (leaving only two columns). That's a step back for how I use Hyprland.

But the great thing is that you can of course just add this to your own bindings file if you prefer move over swap ✌️

PR #2969, opened 2025-10-29
feat: add entry for editing mako config in menu

lmello0:
This commit adds two entries in menu:

- `Menu > Setup > Config > Mako` - this opens the mako config (`~/.config/mako/config`)
- `Menu > Update > Process > Mako` - this restart mako process


dhh:
The mako config is controlled by the theme. I don't think that makes it a good candidate for including and editing directly. Appreciate the suggestion, though ✌️

PR #2926, opened 2025-10-28
Zed FIX - Auto Install Vulkan drivers

ElBrodino:
This script handles Vulkan drivers for install GPU hardware
It is meant to be run when installing Omarchy

Issue #2321, opened 2025-10-08
Zed fails to start on Omarchy Linux due to missing Vulkan driver (AMD Barcelo APU)


ebaerwald:
### System details

AMD Ryzen 5 5625U, AMD Ryzen 7 Barcelo APU (integrated Radeon GPU), Omarchy v3.0.2

### What's wrong?

Zed fails to start on my system (Omarchy Linux, AMD APU). The process runs indefinitely without showing a window. The Zed.log only shows repeated “starting zed” messages with no further output.

`2025-10-08T19:08:53+02:00 INFO  [zed] ========== starting zed version 0.206.7, sha unknown ==========
(repeats multiple times)`

After investigation, the issue appears to be caused by missing Vulkan drivers for the integrated AMD GPU on this distribution’s default installation.

`vulkaninfo | less`

`ERROR: [Loader Message] Code 0 : setup_loader_term_phys_devs:  Failed to detect any valid GPUs in the current config
ERROR: vkEnumeratePhysicalDevices failed with ERROR_INITIALIZATION_FAILED
`

**Workaround**

Manually installing the AMD Vulkan stack fixes the issue:

`sudo pacman -S vulkan-radeon mesa mesa-vdpau lib32-vulkan-radeon lib32-mesa
`

After that, vulkaninfo detects the GPU correctly, and zeditor starts normally.

Omarchy’s installer seems to omit Vulkan driver packages (vulkan-radeon), even though the amdgpu kernel driver is present.

joeldeteves:
Similar issue with LMStudio - Radeon 680M

morgoth:
It's being addressed here https://github.com/basecamp/omarchy/pull/1639

ryanrhughes:
Related #2706

dhh:
Closing as fixed, resolved, or superseded by the current Quattro implementation based on the open-issue review.

PR #2986, opened 2025-10-29
Gpu toggle drivers

ElBrodino:
# Fixes those:
- Zed (missing GPU-related packages #2321 )
- Screen record (missing VAAPI packages #2706 )
There are Pull Request that are supposed to handle this when installing Omarchy (#2926 #1830 )
This PR makes checking and updating the correct packages according to what GPU is installed.
It's running at least 6 times faster as well 😇
# how to test
In Menu / Update / Hardware / there is a new item **GPU**
This will launch the script that detect GPU and install the missing drivers

### DIFF
--- a/install/config/hardware/vulkan.sh
+++ b/install/config/hardware/vulkan.sh
@@ -35,9 +35,9 @@ done
 if [[ -v VULKAN_DRIVER_PACKAGE ]]; then
   echo "Detected display : ${VULKAN_DRIVER_PACKAGE[@]}"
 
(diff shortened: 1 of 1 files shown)


ElBrodino:
check it out @dhh @ryanrhughes

ElBrodino:
For some reason, I made this branch dependent of both Vulkan and VAAPI branches. It does not have to be

dhh:
We got all these sorted now 👍

PR #2988, opened 2025-10-29
feat: Add MANPAGER environment variable for nvim

micahkepe:
This pull request makes a small configuration improvement by setting the manual page viewer to use `nvim` with the `+Man!` option. This enhances the experience of viewing man pages in the terminal by introducing syntax highlighting, table of contents Location List navigation via `gO`, and more.

### DIFF
--- a/default/bash/envs
+++ b/default/bash/envs
@@ -1,3 +1,4 @@
 # Editor used by CLI
 export SUDO_EDITOR="$EDITOR"
 export BAT_THEME=ansi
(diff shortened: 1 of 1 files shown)


dhh:
Interesting idea, but I don't think it's a good fit as a default, as plenty of people will be changing their editor to something other than nvim. Shouldn't have to learn vim to look at man pages. Nice local setting to make, though. We should document that this is possible in Discussion > Manuals 👍

PR #2864, opened 2025-10-26
Integrate global search in walker menu 

RushiChaganti:
With the help of @abenz1267, I was able to make significant progress. He assisted me with several doubts, provided pointers, and highlighted potential issues.

This is a draft PR.

PR #3005, opened 2025-10-30
Feature/global search

RushiChaganti:
This PR integrates global search into the Walker menu.

### Key Changes
- Adds a global search entry and trigger within Walker
- Allows navigating through menu items using arrow keys while global search is active
- Adds descriptions to each menu item to clarify functionality
- Converts keybindings to Lua and integrates them into the Walker menu
- Ensures consistent UX across the Walker interface
- Thoroughly tested against the latest `dev`

### Notes
- Replaces previous PR #2864 for a cleaner review history
- Rebased on `dev` with a clean branch
- Tested all menus and workflows
- Introduces one additional enhancement (global-search-aware keybindings)

Thanks to @abenz1267 for continuous guidance and helping debug issues throughout this process.


AndrewMast:
It appears that some changes undo what was made in recent releases, especially in `utilities.conf`. But I'd personally love to have a global search.

RushiChaganti:
https://github.com/user-attachments/assets/94afbf3b-cf11-4935-a865-8e68b80e4fb1

RushiChaganti:
> It appears that some changes undo what was made in recent releases, especially in `utilities.conf`. But I'd personally love to have a global search.
@AndrewMast 
Can you let me know which changes were rolled back? I haven’t really kept track.

AndrewMast:
Some of the names of the keybinds were recently changed. Not super important, but they are reverted here for seemingly unrelated reasons. Also some unneeded whitespace changes in `layout.xml` and removed comments, etc.

Also, dhh recently implemented prioritized keybindings in `omarchy-menu-keybindings` that are removed here.

Relatively minor things, but trying to keep things clean and help this get implemented.

RushiChaganti:
Thanks for pointing those out — I'll clean up the reverted keybind names, whitespace, and comments.
And yeah, I'll make sure the prioritized keybindings change is preserved as well.

Could you share the branch or commit where DHH pushed that update? I wasn't able to find it.

Also experimenting with reducing the subtext size — will push an update tomorrow.

AndrewMast:
I think the changes were made in #2862, at least the prioritized keybindings were done there. (edited, linked wrong PR)

dhh:
This is pretty interesting, but a few notes:

1) I don't like these icons at all. They include color, clash with the underlying theme, and just don't look very nice. We have to be able to use nerdfont glyphs.

2) I don't like the subtext either. It makes the launcher look messy. Much prefer the clean look of a single line.

dhh:
The sizing of the windows aren't right either. For this to fly, it has to be a pixel-by-pixel match of the existing launcher we have.

RushiChaganti:
Thanks for the feedback. Could you let me know how you'd like to handle cases where there are multiple theme or Waybar entries? I'd like to understand the selection logic before proceeding.

I'll replace the icons with appropriate Nerd Font glyphs and remove the subtext to match the clean, single-line style you prefer.

I'll also ensure the launcher window sizing matches the existing one pixel-perfectly.

RushiChaganti:
Would you also like me to integrate the additional features, such as the priority keybindings? They aren’t strictly necessary now since I’ve already connected them through hyprctl dispatch, but I can still add them if you want.

I’ll also work on the new scratchpad feature you introduced in version 3.1.4. I haven’t had a chance to update yet, but I’ll review it and incorporate the changes.

dhh:
Hmm. I just looked the full implementation. This is a mountain of code. I appreciate all the work you've put into it, but it's not something I'd want to maintain going forward. Maybe we can get @abenz1267 or @ryanrhughes to chime in on other ways to achieve what we're trying to do in less than 1500 lines of code.

abenz1267:
Who cares about LoC if its easier to maintain? A bunch of toml files is way less error prone than anything you will do in bash. Not to mention easier to customize and extend.

RushiChaganti:
Yeah, I agree with @abenz1267  here. The TOML-based approach is more maintainable and easier to extend compared to a large bash implementation, even if it results in a higher line count. And we have converted the whole menu to TOML, So the LOC will increase.

dhh:
I don't find the toml format an improvement. It's excessively verbose compared to what we have now. It also doesn't work with dynamic elements, like we use in the themes list or the fonts list, as I understand it?

But the verbosity of code I was talking about was actually all the rest that's in this setup.

Overall, the PR has to feel as light and succinct as what we have now. What's current here does not.

dhh:
Yeah, this is using hardcoded lists for themes and fonts. That'd be a no-go as well.

abenz1267:
you can create all the menus with lua if that floats your boat more i guess? Also allows creating dynamic menus. Walker users have already created whole browser bookmarks menus using it.

> Overall, the PR has to feel as light and succinct as what we have now. What's current here does not.

Your current impl also doesnt have proper global search. So there's that i guess :joy: .

mkbula:
> It also doesn't work with dynamic elements, like we use in the themes list or the fonts list, as I understand it?

I have a whole theming setup with lua elephant menus only. It's the native solution and doesn't need dmenu + bash.

dhh:
Happy to see other takes on this. But we are not giving up on dynamic menus or doubling the LOC size to get global search.

abenz1267:
I already told you dynamic menus are very much possible with Lua. That was not a joke. @RushiChaganti would just need to implement them.

But since this won't ever go forward, there's no point in doing that to begin with i guess.

Too bad one of the most requested features and most criticized points in every review out there isn't going forward because of LoC.

I guess this can be closed then.

abenz1267:
@dhh would this be better?

```lua
Name = "omarchy_learn"
NamePretty = "Learn"
Cache = true
Action = "%VALUE%"

function GetEntries()
    return {
        { Text = "  Keybindings", Value = "$HOME/.local/share/omarchy/bin/omarchy-safe-exec 'omarchy-menu-keybindings'" },
        { Text = "  Omarchy", Value = "$HOME/.local/share/omarchy/bin/omarchy-safe-exec 'omarchy-launch-webapp https://learn.omacom.io/2/the-omarchy-manual'" },
        { Text = " Hyprland", Value = "$HOME/.local/share/omarchy/bin/omarchy-safe-exec 'omarchy-launch-webapp https://wiki.hypr.land/'" },
        { Text = "󰣇 Arch", Value = "$HOME/.local/share/omarchy/bin/omarchy-safe-exec 'omarchy-launch-webapp https://wiki.archlinux.org/title/Main_page'" },
        { Text = "󱆃 Bash", Value = "$HOME/.local/share/omarchy/bin/omarchy-safe-exec 'omarchy-launch-webapp https://devhints.io/bash'" },
        { Text = " Neovim", Value = "$HOME/.local/share/omarchy/bin/omarchy-safe-exec 'omarchy-launch-webapp https://www.lazyvim.org/keymaps'" }
    }
end
```

This is your current bash:

```bash
show_learn_menu() {
  case $(menu "Learn" "  Keybindings\n  Omarchy\n  Hyprland\n󰣇  Arch\n  Neovim\n󱆃  Bash") in
  *Keybindings*) omarchy-menu-keybindings ;;
  *Omarchy*) omarchy-launch-webapp "https://learn.omacom.io/2/the-omarchy-manual" ;;
  *Hyprland*) omarchy-launch-webapp "https://wiki.hypr.land/" ;;
  *Arch*) omarchy-launch-webapp "https://wiki.archlinux.org/title/Main_page" ;;
  *Bash*) omarchy-launch-webapp "https://devhints.io/bash" ;;
  *Neovim*) omarchy-launch-webapp "https://www.lazyvim.org/keymaps" ;;
  *) show_main_menu ;;
  esac
}
```

and this is the toml from this pr:

```toml
name = "omarchy_learn"
name_pretty = "Learn"
action = "%VALUE%"

[[entries]]
text = "Keybindings"
value = "$HOME/.local/share/omarchy/bin/omarchy-safe-exec 'omarchy-menu-keybindings'"

[[entries]]
text = "Omarchy"
value = "$HOME/.local/share/omarchy/bin/omarchy-safe-exec 'omarchy-launch-webapp https://learn.omacom.io/2/the-omarchy-manual'"

[[entries]]
text = "Hyprland"
value = "$HOME/.local/share/omarchy/bin/omarchy-safe-exec 'omarchy-launch-webapp https://wiki.hypr.land/'"

[[entries]]
text = "Arch"
value = "$HOME/.local/share/omarchy/bin/omarchy-safe-exec 'omarchy-launch-webapp https://wiki.archlinux.org/title/Main_page'"

[[entries]]
text = "Bash"
value = "$HOME/.local/share/omarchy/bin/omarchy-safe-exec 'omarchy-launch-webapp https://devhints.io/bash'"

[[entries]]
text = "Neovim"
value = "$HOME/.local/share/omarchy/bin/omarchy-safe-exec 'omarchy-launch-webapp https://www.lazyvim.org/keymaps'"
```

ryanrhughes:
There are 3 requirements I think it important to outline here that we're not currently meeting and need to do a little bit of work on to make this a really viable addition. 

- Ease of maintenance - It should be easy to add / edit / remove items throughout the **entire** menu at one time.
- Consistent experience - The experience should be consistent with what we have today, while adding the ability to drill down with search.
- Handle themes and fonts (dynamic) - It needs to show themes and fonts as options in those respective places regardless of when a user has installed them.

I have some thoughts on each that can get us there.

### Ease of Maintenance
The current approach gives me pause for a few reasons from a sheer maintenance perspective and it's only partially to do with the verbosity of toml as a configuration. History tells me that this config will change. Therefore, we will be forced to make massive, error-prone, find and replace edits. I don't like this. 

In addition, it spreads a single concept (the omarchy-menu) across numerous locations, which I also don't like. That spread doesn't even provide a hierarchical benefit, it's just noise.

With these two things in mind; is it possible that we're jumping to the end configuration that Elephant expects too soon and should consider our part in this as more of a generator whereby we create / maintain a simple function or two that define a menu in a much easier to handle syntax and allow the generator to output the 15 toml files with 1k LoC? This would insulate us against configuration changes since we just change the generator, and make working on the menu much more concise since it could still be **the menu** not _the 20 files that make the menu_. We would need to determine what triggers re-generate the outputs if we went down this road.

The lua idea @abenz1267 pointed out as an option may also something worth exploring.

> I already told you dynamic menus are very much possible with Lua. That was not a joke. @RushiChaganti would just need to implement them.

When I think about adding an item, I just want to define an item simply like:
```
menu_item("Steam", "The steam launcher app", "steam-icon", omarchy-setup-steam)
```

Or even a JSON structure to map over would be nicer / easier to deal with. At least it provides hierarchy and consolidation; though I don't know if that's the solution here.

### Consistent Experience
The experience of this needs to be like the Omarchy menu we have today (Super + Alt + Space opens a menu to access these things). We should be able to change it out and no one even notice other than that their search works better and provides better results. 
image

As I've voiced before, I don't see any desire to merge all of the Omarchy menu items in with the app launcher like has been done here. It just results in a noisy mess. Here I'm just trying to launch Steam and have to deal with a host of irrelevant options, and options that are only even relevant to me as a user once which isn't great.
image

### Handle Fonts and Themes
Fonts and themes can be installed so this has to be somewhat dynamic or triggered to refresh from what the user has on install / remove of those items. We can't have it to where you can only select to change to themes that ship with Omarchy. That's counterintuitive.

ryanrhughes:
> would this be better?

@abenz1267 Yes...that is objectively a better approach, and in-line with my novel. It's actually even a slight bit better than today since it unifies the entry and function line.

ryanrhughes:
@abenz1267 pointed out to me that I forgot to run `omarchy-refresh-walker` after pulling this PR which made some of what I was seeing incorrect.

This is still inconsistent and needs some work but at least not merged with all other items in the system.
image

RushiChaganti:
> Or even a JSON structure to map over would be nicer / easier to deal with. At least it provides hierarchy and consolidation; though I don't know if that's the solution here.

I already did that mapping of JSON and made the menu consistent but the user needs to go to search option that's it 
https://github.com/basecamp/omarchy/commit/3a7134b202ee56ca0432150251d62b3ff38c2a76#diff-3cc1fc1c75dd7c43dcdc343c639cdd2ade579c8e8002c0ce25e444f83d0721e0

RushiChaganti:
> @dhh would this be better?

I will give this a shot

RushiChaganti:
@dhh & @ryanrhughes
How should I handle menu items like Theme or Waybar, which include multiple actions such as install, remove, and update?

For example, searching for “theme” would show three related options, or if someone selects “install,” they’d then see many choices like Install Elixir, Install Steam, etc.

Maybe we should include subtext or short descriptions for better clarity — or do you have other suggestions?
What are your thoughts, @abenz1267 ?

abenz1267:
I think the biggest problem is that the omarchy menu was not done with global search in mind. The requirements for such a menu are rather different. You want less duplication of items and more grouping by domain. Even IF you add subtext to indicate what menu an item is from, you end up with this:

```
------
Waybar
Some Menu
------

------
Waybar
Some Menu 2
------

------
Waybar
Some Menu 3
------
```

This is kinda bad. And since DHH doesn't not want the subtext, you'd rather have to do:

```
------
Install Waybar
------

------
Remove Waybar
------

------
Customize Waybar
------
```

But what happens if you actually enter the `customize` menu?

```
------
Customize Waybar
------

------
Customize Hyprland
------

------
Customize Hyprlock
------
```

This isn't an option either. One way to work around this would be by putting the subtext BEFORE the actual text in global search:

```
------
Customize Waybar
------
```

Because the menu's name will only be put into subtext IF it's global search. See here:

image

vs

image

So you could put subtext on the same line as text

I'd still rethink different things:
- do you really need install/remove? why cant this be handled in a single menu?
- do you need the "trigger" menu? f.e. a proper keybind search will bring up the desired keybind anyways ... and you can just activate it (if implemented).

I think having a good look at the current menu and optimizing it would be the proper approach.

RushiChaganti:
**The issue is that Omarchy’s menus are currently organized by action instead of by subject.**

Also, there are multiple separate files handling different theme-related actions — such as update, theme-set, install, theme-list, theme, and current.

A better approach would be to merge these into a single command file and use named sub-commands instead. For example:

To install a theme: `./theme install <theme-name>`

To remove a theme: `./theme remove <theme-name>`

This would make the structure cleaner, more consistent, and easier to maintain.

dhh:
Thanks for exploring it, but ultimately I don't think this path works well.

PR #3006, opened 2025-10-30
Add MeteorJS to JavaScript install options

StorytellerCZ:
A quick addition to add [MeteorJS](https://www.meteor.com/) as an installation option under JavaScript.
You can install MeteorJS either via NPM or via a curl. The second option is still favorite for many and the easiest one. Functionality wise, there isn't a difference.
https://docs.meteor.com/about/install.html

### DIFF
--- a/bin/omarchy-install-dev-env
+++ b/bin/omarchy-install-dev-env
@@ -64,6 +64,10 @@ deno)
   echo -e "Installing Deno...\n"
   mise use -g deno@latest
   ;;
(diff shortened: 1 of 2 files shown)


dhh:
The other JavaScript options are just runtimes. We haven't dug into offering JS web frameworks, as there are simply just too many. Also, the installer for meteor is already easy enough to do. Appreciate the suggestion, though!

PR #3009, opened 2025-10-30
feat: sound for low battery notificaition.

vander00:
This pull request adds a mp3 file and a line for playing the file when the battery is low.

I had a problem when you're fullscreen e.g. watching youtube video you cannot see the notification so the sound solves that.


felixzsh:
i would also like to have sound for notifications, maybe the approach would be to create a omarchy-notify-send to wrap notification along with optional sound.

also, there is no need to add mp3 files to the repo, i see omarchy comes with free desktop sounds, for this case usr/share/sounds/freedesktop/stereo/dialog-warning.oga would do the job.

also, a migration sh file needs to be done for this.

dhh:
That's a fun idea, but I'm not a big fan of interface sound effects. I always turn them off in every OS I use. But would be a fun one to document in Discussions > Manual ✌️

PR #1373, opened 2025-09-01
Add shortcuts for moving workspaces and disabling monitors

jheuing:

PR #3010, opened 2025-10-30
Add Super+Shift+~ keybinding to move workspace between monitors

bhood239:
This adds a keybinding for moving the currently focused workspace to the other monitor:

- **Super + Shift + ~** → Move workspace to the next display

This matches the existing Super + Shift pattern used for workspace management and improves
multi-monitor workflow efficiency, especially when docking or switching setups.

Tested locally on a multi-monitor configuration. No changes to existing behavior for users
without multiple displays.

### DIFF
--- a/default/hypr/bindings/tiling-v2.conf
+++ b/default/hypr/bindings/tiling-v2.conf
@@ -47,6 +47,9 @@ bindd = SUPER, TAB, Next workspace, workspace, e+1
 bindd = SUPER SHIFT, TAB, Previous workspace, workspace, e-1
 bindd = SUPER CTRL, TAB, Former workspace, workspace, previous
 
+# Move workspace between monitors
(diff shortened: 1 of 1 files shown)


jheuing:
This seems to be a duplicate of: https://github.com/basecamp/omarchy/pull/1373

bhood239:
> This seems to be a duplicate of: #1373

Thanks for pointing out that PR. I would say there is a key difference in that #1373 moves the workspace directionally, which depends on the monitor layout and requires 2 bindings to cover both directions (I use my monitor to the left of my laptop, which would mean I would need to use the opposite binding).
The binding I have added moves the workspace to the next monitor in the sequence, which I would say is a simpler and more general action and isn't dependent on the monitor orientation.

I would say this is a complementary option to the options added in #1373 rather than a duplicate

dhh:
Added https://github.com/basecamp/omarchy/commit/70d15921b992477ca383fb9c5e01885282ce1e74

PR #3015, opened 2025-10-31
Don't outline hidden windows when screenshotting fullscreen apps

johntheyoung:
When taking a screenshot with a fullscreen window active, slurp outlines all windows on the workspace including those hidden behind the fullscreen app.

**Before:**
screenshot-2025-10-31_12-38-18

screenshot-2025-10-31_12-35-46

screenshot-2025-10-31_12-31-01

**After:**
screenshot-2025-10-31_12-33-25

screenshot-2025-10-31_12-40-44

This fix detects fullscreen windows and only passes that window to slurp for outlining, preventing visual clutter while preserving normal outline behavior when no window is fullscreen.

### DIFF
--- a/bin/omarchy-cmd-screenshot
+++ b/bin/omarchy-cmd-screenshot
@@ -13,29 +13,34 @@ pkill slurp && exit 0
 MODE="${1:-smart}"
 PROCESSING="${2:-slurp}"
 
+get_active_workspace() {
+  hyprctl monitors -j | jq -r '.[] | select(.focused == true) | .activeWorkspace.id'
+}
+
 get_rectangles() {
-  local active_workspace=$(hyprctl monitors -j | jq -r '.[] | select(.focused == true) | .activeWorkspace.id')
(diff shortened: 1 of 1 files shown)


bin/omarchy-cmd-screenshot
```diff
@@ -19,6 +19,24 @@ get_rectangles() {
   hyprctl clients -j | jq -r --arg ws "$active_workspace" '.[] | select(.workspace.id == ($ws | tonumber)) | "\(.at[0]),\(.at[1]) \(.size[0])x\(.size[1])"'
 }
 
+get_fullscreen_rect() {
+  local ws="$1"
+  # Try multiple field shapes for robustness across Hyprland versions:
```
dhh: We should just make sure this works with 0.52. All Omarchy installs that'll get this new behavior will have the latest.

johntheyoung: Just retested on Omarchy 3.1.7 (Hyprland 0.52 via `hyprland --version`), and the fix still auto-selects the fullscreen window without outlining hidden clients. I also refactored the helper to reuse the cached workspace and collapsed the fullscreen jq fallback so the script matches the other omarchy-cmd-* tools. Let me know if you want extra logging, but this works on 0.52. Thanks!

dhh:
This is no longer a problem 👍

PR #3017, opened 2025-10-31
Add copy-url support for Brave browser

Blinkuu:
Adds support for the `copy-url` extension to Brave Browser and a required migration.


Blinkuu:
@dhh Can we land this? It's a small QoL. Currently, I have to keep patching it after every Omarchy release. There are likely more people that struggle with this :sweat_smile:

PR #3019, opened 2025-10-31
Add keybinding for hyprpicker color picker

osmansemir:
## Summary
Added a keybinding to launch hyprpicker, making it easier to quickly pick colors from the screen.

## Changes
- Added keybinding: [SUPER + SHIFT + P] to launch hyprpicker

## Motivation
Hyprpicker is already included in the distro, but there wasn't a convenient way to launch it. This keybinding makes color picking more accessible for users working with design, theming, or development.

## Testing
- Tested the keybinding launches hyprpicker successfully
- Verified it doesn't conflict with existing keybindings
- Color picker functions as expected


default/hypr/bindings/utilities.conf
sgruendel, line 1: it's already there in line 37 :)

dhh:
We already have it on `Super + PrtScr`.

PR #3020, opened 2025-10-31
Update icon design in icon.txt

Rishabh-Sarang:
The logo is way too big


dhh:
I like the size as it is. It fills out the gap left for it on fastfetch. But this is why we expose it. You can change it to whatever you like!

Rishabh-Sarang:
> I like the size as it is. It fills out the gap left for it on fastfetch. But this is why we expose it. You can change it to whatever you like!

image

this looks way better

dhh:
It's not the same logo, though. Proportions are entirely different. Appreciate the effort, but don't think this is an improvement.

PR #3025, opened 2025-10-31
Added functionality for selecting the Windows version in the omarchy-windows-vm

yetanotherdude2:


dhh:
Think we're better off leaving this as an advanced step for folks to do themselves. And keep the main path as simple as can be.

PR #3030, opened 2025-10-31
Use gdu disk usage analyzer

gyaaniguy:
The default 'disk usage' application - dust is slow, has no shortcuts, and well  quite nice as a du replacement, its quite basic to be considered as a disk usage analyzer.

I found this - `gdu` - https://github.com/dundee/gdu 

> Pretty fast disk usage analyzer written in Go.
>Gdu is intended primarily for SSD disks where it can fully utilize parallel processing. However HDDs work as well, but the performance gain is not so huge.

In my testing, gdu is really really fast! Its amazing actually. Crunches through 100s of GB like its nothing. Has sweet shortcuts, lots of functions - delete files, ability to jump to path, sorting options etc. So we can actually use it for removing old trash.

This PR 
- changes and sets gdu as the default omarchy disk usage application
- doesn't remove dust. Its still a kickass du replacement.
- uses `gdu /` as the default. but `gdu -d` might be a better alternative - In this it will first show all the drives
- **Is completely untested ! I don't know how to !**

Please let me if I can test this somehow, or if PR needs changes

gdu is on pacman `sudo pacman -S gdu` 

Cheers

### DIFF
--- a/bin/omarchy-tui-install
+++ b/bin/omarchy-tui-install
@@ -3,7 +3,7 @@
 if [ "$#" -ne 4 ]; then
   echo -e "\e[32mLet's create a TUI shortcut you can start with the app launcher.\n\e[0m"
   APP_NAME=$(gum input --prompt "Name> " --placeholder "My TUI")
-  APP_EXEC=$(gum input --prompt "Launch Command> " --placeholder "lazydocker or bash -c 'gdu'")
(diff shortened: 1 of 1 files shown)


dhh:
I think gdu is pretty cool, but I don't think it's a straight replacement for du, which doesn't rely on anything interactive. It's just a quick overview. I think that's a better fit for the default. Appreciate the suggestion though! ✌️

Issue #2059, opened 2025-09-29
System does not suspend when idle


colutti:
### System details

AMD Radeon RX 7900 XTX [Discrete], 3840x2160 @ 1.67x in 28", 60 Hz [External Monitor], Omarchy v3.0.1

### What's wrong?

My PC never suspends when idle when running Omarchy. I think, by default, omarchy should add:

listener {
    timeout = X
    on-timeout = systemctl suspend
}

to enable the system to suspend after X time.

woopstar:
I think you are right that this is missing. We would also need to make sure this is in #1417  because we might want to do suspend-then-hibernate if hibernate support is available

SuspendTV:
Im on desktop and added this to my hypridle.conf but it may be up for debate what behaviour people would want as default on a laptop, the safe default may be to add it as a later listener after 7 min?

gyaaniguy:
Shouldn't default  suspend  only happen if system is running on battery ?

SuspendTV:
> Shouldn't default suspend only happen if system is running on battery ?

My opinion is that i would always want my system to suspend after a while, but i can see arguments for the other way too.
In the spirit of "Keep it simple" i would still argue that the always suspend approach should be the default.

nnutter:
As some random user, my laptop always being dead when I go to use it is not ideal. I have some `hypridle.conf` setup I made following, I think, an Arch wiki page but it does not hibernate and maybe doesn't even suspend right because my laptop is usually dead when I open it up. It also doesn't replace the screen when locking so when it isn't dead and I open it up the screen shows content for a brief moment before showing the lock screen. I love in macOS having the option of having the lock kick in some seconds after the screensaver starts so that I can provide input and stop the lock when I'm reading or whatever and the timeout is about to happen. Sorry, this isn't more helpful.

<details>
<summary>~/.config/hypr/hyperidle.conf</summary>

```
general {
    lock_cmd = omarchy-lock-screen                         # lock screen and 1password
    before_sleep_cmd = loginctl lock-session               # lock before suspend.
    after_sleep_cmd = hyprctl dispatch dpms on             # to avoid having to press a key twice to turn on the display.
    on_unlock_cmd = omarchy-restart-waybar                 # prevent stacking of waybar when waking
}

# FIXME: Scrensaver depends on tte being installed
# listener {
#     timeout = 150                                             # 2.5min
[...]

Issue #2077, opened 2025-09-29
Create a contributor's guide (CONTRIBUTING.md)


flaviolivolsi:
### What correction or addition do we need?

As mentioned in PR #2072 a contributor's guide would be useful. This issue is to track that effort.

As a new contributor I ran into a few points of uncertainty that a guide could clarify. Having a `CONTRIBUTING.md` would make it easier for new people to submit high-quality PRs and reduce the review burden on maintainers.

Some potential topics to cover:

*   **Branching strategy:** clarification on the default branch for PRs (currently seems to be `dev` even though it might be just temporary).
*   **Migrations:** guidelines on creating and testing migrations for existing installations
*   **Code style:** clarification on any linters, formatters or general style guidelines in use if any
*   **Setting up a dev environment:** any specific steps needed to test changes locally

Happy to help contribute to a draft once a direction is decided!

victoralensai:
This would be greatly appreciated! :D

Maybe also add small descriptions of what is the purpose of some folders (for example i was confused why there is a `config/waybar` but also a `defaults/waybar`).

brunny-eth:
happy to help here as well if you need a pair of hands @flaviolivolsi

PR #3031, opened 2025-10-31
added suspend after 7 min

SuspendTV:
Proposed solution for #2059, Im unsure if a migration should be added.

### DIFF
--- a/migrations/1761664693.sh
+++ b/migrations/1761664693.sh
@@ -2,6 +2,7 @@ echo "Adding auto suspend after 7 min"
 
 # Add the auto suspend listener to hypridle.conf
 if ! grep -q 'on-timeout *= *systemctl suspend' ~/.config/hypr/hypridle.conf; then
+    cp ~/.config/hypr/hypridle.conf ~/.config/hypr/hypridle.bak.$(date +%s)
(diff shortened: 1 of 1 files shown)


dharmavagabond:
Yes, you need a migration that runs when we update.

SuspendTV:
I have now added a migration, but i dont understand the naming of them...
i also still dont know if we want to edit peoples configs? or is that part of the "opinionated" approach ;)

dharmavagabond:
Migrations should be created with `omarchy-dev-add-migration` (it uses Unix timestamps for the naming). As for the second part, it is advised to create a copy instead of overwriting. For example, take a look at this [migration](https://github.com/basecamp/omarchy/blob/master/migrations/1757783204.sh).

SuspendTV:
Thanks for the help, any more feedback?
This is my first pull request, and it has been a great experience, thanks for the help. #2077 would probably make new people like me take up less of peoples time ;)

dhh:
Suspend is unfortunately not reliable on all systems, so it's not something we can trigger by default. Thanks for the suggestion, though! Good one to add for folks who want it on systems that support it.

PR #3034, opened 2025-10-31
Nvidia script refactoring

killeik:
So, in #1417, I discovered that some shenanigans are needed to enable hibernation for Nvidia users.
Somehow I found myself refactoring this script :D

1. Installs headers to all installed kernels.
2. Adds modules to a separate file `/etc/mkinitcpio.conf.d/nvidia.conf` instead of using sed on `/etc/mkinitcpio.conf`.
3. Uses `limine-mkinitcpio` instead of `mkinitcpio -P` (this removes another interactive confirmation).
4. Creates `env =` for nvidia in `envs.conf` instead of `hyprland.conf`.
5. Exits early, so all the script not indented
6. Overall improvements

I also added migration to standardize the configurations for already existing users. 

@Kn0ax, as the author of the script, please let me know if you have any concerns.

### DIFF
--- a/install/config/hardware/nvidia.sh
+++ b/install/config/hardware/nvidia.sh
@@ -8,73 +8,52 @@
 #
 # ==============================================================================
 
-# --- GPU Detection ---
-if [ -n "$(lspci | grep -i 'nvidia')" ]; then
-  # --- Driver Selection ---
-  # Turing (16xx, 20xx), Ampere (30xx), Ada (40xx), and newer recommend the open-source kernel modules
-  if echo "$(lspci | grep -i 'nvidia')" | grep -q -E "RTX [2-9][0-9]|GTX 16"; then
-    NVIDIA_DRIVER_PACKAGE="nvidia-open-dkms"
-  else
-    NVIDIA_DRIVER_PACKAGE="nvidia-dkms"
-  fi
+# Exit early if no nvidia devices found
+if lspci | grep -qvi 'nvidia'; then
+   exit 0
+fi
 
-  # Check which kernel is installed and set appropriate headers package
-  KERNEL_HEADERS="linux-headers" # Default
-  if pacman -Q linux-zen &>/dev/null; then
-    KERNEL_HEADERS="linux-zen-headers"
-  elif pacman -Q linux-lts &>/dev/null; then
-    KERNEL_HEADERS="linux-lts-headers"
-  elif pacman -Q linux-hardened &>/dev/null; then
-    KERNEL_HEADERS="linux-hardened-headers"
-  fi
+# Driver Selection
+NVIDIA_DRIVER_PACKAGE="nvidia-dkms"
 
-  # force package database refresh
-  sudo pacman -Syu --noconfirm
+# Turing (16xx, 20xx), Ampere (30xx), Ada (40xx), and newer recommend the open-source kernel modules
+if lspci | grep -i 'nvidia' | grep -q -E "RTX [2-9][0-9]|GTX 16"; then
+  NVIDIA_DRIVER_PACKAGE="nvidia-open-dkms"
+fi
 
-  # Install packages
-  PACKAGES_TO_INSTALL=(
-    "${KERNEL_HEADERS}"
-    "${NVIDIA_DRIVER_PACKAGE}"
-    "nvidia-utils"
-    "lib32-nvidia-utils"
-    "egl-wayland"
-    "libva-nvidia-driver" # For VA-API hardware acceleration
-    "qt5-wayland"
-    "qt6-wayland"
-  )
+# Add headers to all supported kernels
+KERNEL_HEADERS="$( pacman -Q | grep -E '^linux(-zen|-lts|-hardened)? ' |awk '{print $1 "-headers"}' )"
 
-  sudo pacman -S --needed --noconfirm "${PACKAGES_TO_INSTALL[@]}"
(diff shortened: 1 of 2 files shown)


Kn0ax:
Hi,
Thanks for taking your time and trying to improve it. 

I will have a few comments.

install/config/hardware/nvidia.sh
```diff
@@ -8,73 +8,52 @@
...
+  "egl-wayland"
+  "libva-nvidia-driver" # For VA-API hardware acceleration
+  "qt5-wayland"
+  "qt6-wayland"
+)
+sudo pacman -Syu --needed --noconfirm "${PACKAGES_TO_INSTALL[@]}"
+
+# Configure modprobe for early KMS and hibernation
+cat <<EOF | sudo tee /etc/modprobe.d/nvidia.conf >/dev/null
+options nvidia_drm modeset=1
```
Kn0ax: It seems so we don't need that
https://gitlab.archlinux.org/archlinux/packaging/packages/nvidia-utils/-/issues/15

install/config/hardware/nvidia.sh
```diff
@@ -8,73 +8,52 @@
...
+EOF
 
-  # Remove any old nvidia modules to prevent duplicates
-  sudo sed -i -E 's/ nvidia_drm//g; s/ nvidia_uvm//g; s/ nvidia_modeset//g; s/ nvidia//g;' "$MKINITCPIO_CONF"
-  # Add the new modules at the start of the MODULES array
-  sudo sed -i -E "s/^(MODULES=\\()/\\1${NVIDIA_MODULES} /" "$MKINITCPIO_CONF"
-  # Clean up potential double spaces
-  sudo sed -i -E 's/  +/ /g' "$MKINITCPIO_CONF"
+# Configure mkinitcpio for early loading
+echo 'MODULES+=(nvidia nvidia_modeset nvidia_uvm nvidia_drm)' | sudo tee /etc/mkinitcpio.conf.d/nvidia.conf >/dev/null
```
Kn0ax: This is not a good idea, as we could have other modules that and this could cause problems.

install/config/hardware/nvidia.sh
```diff
@@ -8,73 +8,52 @@
...
 
-  # Define modules
-  NVIDIA_MODULES="nvidia nvidia_modeset nvidia_uvm nvidia_drm"
+# Turing (16xx, 20xx), Ampere (30xx), Ada (40xx), and newer recommend the open-source kernel modules
+if lspci | grep -i 'nvidia' | grep -q -E "RTX [2-9][0-9]|GTX 16"; then
+  NVIDIA_DRIVER_PACKAGE="nvidia-open-dkms"
+fi
 
-  # Create backup
-  sudo cp "$MKINITCPIO_CONF" "${MKINITCPIO_CONF}.backup"
```
Kn0ax, line 58: Some people still install Omarchy via the installation script, and it's always great to have a backup in case of any problems :)

install/config/hardware/nvidia.sh
killeik: For the context, `mkinitcpio` - is just the bash script:
```
> file /usr/bin/limine-mkinitcpio
/usr/bin/limine-mkinitcpio: Bourne-Again shell script, ASCII text executable
> file /usr/bin/mkinitcpio
/usr/bin/mkinitcpio: Bourne-Again shell script, Unicode text, UTF-8 text executable
```
And all the configuration in the `mkinitcpio.conf` - is just bash arrays:
```
> cat /etc/mkinitcpio.conf  |grep -v #
MODULES=(btrfs)
BINARIES=(/usr/bin/btrfs)
FILES=()
HOOKS=(base udev autodetect microcode modconf kms keyboard keymap consolefont block encrypt filesystems fsck resume)
```

In bash, there is a `+=` operator that pushes to an array. I used this so that, when mkinitcpio goes through the `/etc/mkinitcpio.conf.d/nvidia.conf` file, it adds the Nvidia modules to the `MODULES` array.

This makes maintenance easier because, if there are any changes to the desired configuration for nvidia, or if users want to remove it completely, this task will be straightforward, as that file is fully under the control of the nvidia script. 

If you meant something else, please describe what you mean in more detail.

install/config/hardware/nvidia.sh
killeik: Backing up is essential if the file is changed, but it's not in this script, because I moved the mkinitcpio configuration to a separate file for `MODULES+=(...)`

install/config/hardware/nvidia.sh
killeik: Btw, it's not really outdated - I just renamed `/etc/mkinitcpio.conf.d/nvidia.conf` to `/etc/mkinitcpio.conf.d/omarchy_nvidia.conf` to indicate that it's part of omarchy setup.

ryanrhughes: Using a drop-in like this is the preferred way to do it vs editing the main string.

Kn0ax: my bad, i missed it :) lgtm

install/config/hardware/nvidia.sh
Kn0ax: same mistake as above :)

Kn0ax:
i think its ready to merge :) 
thanks <3

dhh:

PR #3037, opened 2025-11-01
Feature: Screen rotation

guibbv2011:
### Screen Rotation

**Overview**  
Effortlessly switch your display orientation on the fly—perfect for 2-in-1 laptops like Framework in tent, tablet, or portrait modes. Seamlessly integrates with Hyprland for instant, hardware-aware flips without disrupting your workflow.

**How It Works**  
- **Quick Access**: Press **Super + CTRL + arrow** to cycle through orientations: Normal (up), Right (90° CW), Down (180°), Left (270° CW).  
- **Smart Behavior**: Applies only to your active monitor and rotates touch input/tablet stylus to match—everything feels natural, no lag or redraw glitches.  

**Note**
> You need to set transform default in .config/hypr/monitors.conf

### DIFF
--- a/bin/omarchy-cmd-rotate-screen
+++ b/bin/omarchy-cmd-rotate-screen
@@ -1,13 +1,12 @@
 #!/bin/bash
 
-# Exit if arguments are missing
 if [[ -z "$1" ]]; then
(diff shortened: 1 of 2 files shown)


dhh:

PR #3054, opened 2025-11-01
Add Binding to move window out of scratchpad

avinashjoshi:
PR #2669 adds support for special scratchpad workspace. This PR adds binding to move window out of scratchpad workspace if needed.


dhh:
We already have `Super + Shift + [n]` to move it. I'd prefer to stick with that.

PR #3058, opened 2025-11-02
feat: add window minimize and restore feature

MBMarsell:
## Window minimize feature
Got a little inspired by scratchpad, found it to be nice for overlaying and such, but figured being able to "minimize" something and bring it back easily would be nice too.

- Adds the option to minimize windows using a special workspace (special:minimized)
- Restore minimized windows
- Shows count of minimized windows with indicator

### Keybinds

Minimize current window with `SUPER ALT, M`
Open restore windows menu with `SUPER ALT SHIFT, M`

### In action

https://github.com/user-attachments/assets/faee4b47-554f-42b0-9a3b-e71d0948ff05

### DIFF
--- a/bin/omarchy-cmd-minimize
+++ b/bin/omarchy-cmd-minimize
@@ -1,0 +1,91 @@
+#!/bin/bash
+
+MINIMIZE_WORKSPACE="special:minimized"
+MENU_COMMAND="walker --dmenu -p 'Restore Minimized Window'"
+STATE_DIR="$HOME/.local/state/omarchy"
+LIST_FILE="$STATE_DIR/minimized_windows.list"
+
+# Ensure state directory exists
+mkdir -p "$STATE_DIR"
+
+MINIMIZE_ACTION() {
+    WINDOW_INFO=$(hyprctl activewindow -j)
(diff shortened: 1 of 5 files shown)


abenz1267:
Not sure how i feel about this. Tilers dont have a minimize function for a reason and this seems more like a hack.

This makes it way too easy to not adapt to a proper tiling wm workflow, by sticking to what you are used to from non-tilers.

Just because you can, doesn't mean you should.

MBMarsell:
Appreciate the feedback, thanks!

I also guess the term 'minimize' is a little misleading here as I see it more as an enhanced 'stash'. But by handling windows this way with the indicator and menu as well, makes them be out of sight, but not out of mind (I guess this need already exists as we have the possibility for 'special hidden workspaces' in tilers?).

I don't think this makes the adaptation harder either, as you already can move stuff into a 'throwaway' workspace or the scratchpad workspace and forget about that, although this is a little different of course as it is not overlaying, but you get an overview of what's been hidden and can easily bring back what you want.

Thanks again for your thoughts!😄

dhh:
Appreciate the suggestion here, but agree, I don't think it fits the tiling metaphor that well. And it's a lot of code. So don't think it's a good fit for the default, but you would wrap this up in a package for those who want it.

PR #3061, opened 2025-11-02
Fix: omarchy-cmd-screenrecord for regions in monitors left/above primary

ryrobes:
**Problem:** When trying to screenrecord a region to the left and/or above of the primary monitor - negative target coords are produced. `gpu-screen-record` does not allow this and fails. 

**Solution:** In order to safely capture these "out of bounds" regions - we detect this case and record the entire target parent display, then post-process crop it with the translated coords of the targeted area. While not ideal (until upstream `gpu-screen-record` allows negative coords), it ensures the clip is saved and works as the user expects.

**Additional Fix:** Changed the `screen-recording.sh` waybar check script to look for the new temp file instead of the `gpu-screen-recorder` process, this has the added benefit of allowing the "stop recording" icon to be cleared from _other_ monitors waybars other than the one that was clicked (all other displays would previously continue to show the icon indefinitely with the process check method). Using the tmp check file allows this to work consistently across all displays.

**Tests:** Tested all screen recording modes on all monitors and regions, no regressions seen - while still allowing the "bad coord" region recordings to function.

### DIFF
--- a/bin/omarchy-cmd-screenrecord
+++ b/bin/omarchy-cmd-screenrecord
@@ -11,6 +11,9 @@ fi
 SCOPE=""
 AUDIO="false"
 WEBCAM="false"
+# need a persistent file to store crop params for when stop is called 
+CROP_FILE="/tmp/omarchy_screenrecord.crop"
+INFO_FILE="/tmp/omarchy_screenrecord.info"
 
 for arg in "$@"; do
   case "$arg" in
@@ -22,6 +25,11 @@ done
 
 cleanup_webcam() {
   pkill -f "WebcamOverlay" 2>/dev/null
+}
+
+cleanup_crop_files() {
+  [[ -f "$CROP_FILE" ]] && rm -f "$CROP_FILE"
+  [[ -f "$INFO_FILE" ]] && rm -f "$INFO_FILE"
 }
 
 start_webcam_overlay() {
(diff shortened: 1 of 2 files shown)


dec05eba:
How do you select an out of bounds region? or how do you configure monitor(s) to have negative coordinates?

ryrobes:
It's only out of bounds for `gpu-screen-record`, not the user. Think of it this way...

- You have 3 screens - [1] [2] [3]
- Center screen 2 is the "primary" display
- You record a region on screen 2 or 3
   - No problem, `gpu-screen-record` gets the offset region coords starting from the primary monitor's 0x0 in the upper left - treating it as one big monitor of addressable pixels stretching _right and down_.

- You go to record a region on screen 1
   - Problem. The current system tries to target a negative X value to "reach" those pixels on the left of the primary 0x0 point (it's "relative pixel address" / region)
   - `gpu-screen-record` doesn't support negatives, so the command fails silently and nothing gets recorded.
   - Same issue with a screen ABOVE the primary's 0x0 waterline, since that would create negative a _Y_ value.

_Basically a workaround to allow people with odd / complex screen setups to still use region selection on all screens._

dec05eba:
@ryrobes I wasn't able to reproduce it that way. Wayland doesn't have a concept of a primary monitor (and neither does hyprland) so I wasn't able to get negative position with monitor setup. However I was able to manually set negative position with hyprland config, so I fixed it now. The fix is in gpu-screen-recorder aur version `r1189.a545fe4`. Unfortunately nobody bothered reporting the issue so I didn't know it existed, even though it only took a few seconds to fix it.
Try with that newest version of gpu screen recorder to see if you can remove that workaround.

ryrobes:
Ah, didn't realize you were the author. Nice. Apologies. Will give it a try. I didn't set the screens to have negative coords explicitly, it just worked out that way. 

i.e.
```
monitor=DP-1,3840x2160@60,auto-left,1.25
monitor=DP-2,3840x2160@60,auto-left,1.25
monitor=DP-3,3840x2160@60,auto-right,1.25
```

Region select on 1 and 2 was causing:

`gsr error: invalid value for option -region '795x516+-2745+597', expected width, height, x and y to be greater or equal to 0` etc.

_Likely an edge case since it wasn't reported anywhere._

ryrobes:
Confirmed fixed in `gpu-screen-recorder` r1189.a545fe4. Closing. Thanks @dec05eba!

PR #3063, opened 2025-11-02
Fix: adjust battery monitor logic to prevent warning on every boot

ZeroSegFault:
Several users reported that the “Time to charge!” battery warning appears on every boot, even when the battery is not low.

This issue occurs because omarchy-battery-remaining returns an empty value early in the boot process.
This patch adds a simple check to skip sending the notification if the returned value is empty.


dhh:
This has already been fixed on dev 👍

PR #3074, opened 2025-11-02
added command line options to run omarchy-update non-interactively

qollin:
I my opinion it should be possible to run omarchy-update without the interactive reminder at the start. To that end I added the command-line option "-y" (for "yes") for non-interactive mode which skips the initial gum dialog. This is implemented using the  getopts utility.


dhh:
Curious: When are you running this non-interactively? Not against a flag, but we don't need all this code to do it. Can just do `$1 == "-y"`.

qollin:
On vanilla Arch I ran "yay" every day after login. On Omarchy I run omarchy-update (via an "ou" alias). And having to manually choose "yes" every time is tedious ;-)

And yeah, the solution is a bit over engineered. I wanted to do it properly and set the groundwork for future flags... But I'm happy to change it to a simple "$1" == "-y"....

dhh:
We now have `omarchy-update -y` 👍

PR #3076, opened 2025-11-02
Add superfile. A beautiful, modern TUI for file browsing and advanced…

etmartinkazoo:
… file management.

A TUI for file management seems fitting for the developer focus of Omarchy. Superfile is very well maintained and offers advanced functionality for those who predominantly use the terminal.


dhh:
Thanks for the suggestion, but I'm happy with Nautilus as a default, and don't want to double up. Good one to add to your own setup, though!

PR #3082, opened 2025-11-02
Further instructions for auto nightlight

morgoth:
Follow up to https://github.com/basecamp/omarchy/pull/2446l

I noticed that it did not work because the process was not running by default, vide https://github.com/basecamp/omarchy/commit/d9efaac906cc291f27a242d66e940a6357658b3c

### DIFF
--- a/config/hypr/hyprsunset.conf
+++ b/config/hypr/hyprsunset.conf
@@ -5,7 +5,9 @@ profile {
     identity = true
 }
 
-# Enable auto switch to nightlight:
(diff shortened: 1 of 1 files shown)


config/hypr/hyprsunset.conf
```diff
@@ -5,7 +5,9 @@ profile {
     identity = true
 }
 
-# Enable auto switch to nightlight:
+# To enable auto switch to nightlight, set in your .config/hypr/autostart:
```
fcsonline, line 8: Maybe we can change this path `.config/hypr/autostart.conf`

PR #3083, opened 2025-11-02
Align mako font with other apps

AhmedCharfeddine:
Hello,

Unless I'm missing something, the specified mako font is inconsistent with the rest of omarchy. 
This fixes that :)

image


dhh:
This is intentional. I see notifications as more like a browser window. We don't use monospace fonts there.

PR #3087, opened 2025-11-02
Add progressbar trough color variable for swayosd

artfwo:
Currently swayosd progressbar has a gray background (called `trough` in swayosd) in every theme and it doesn't always look great.

This change makes `trough` themeable and adds trough colors for all default themes. The default colors are generated as midpoints between previously defined background and progress colors.

Attached are before and after examples for cobalt2 theme. Extra themes will need to define `@trough` in swayosd.css, or the progressbar will appear with the same background as the window background.

image
image

### DIFF
--- a/config/swayosd/style.css
+++ b/config/swayosd/style.css
@@ -23,6 +23,10 @@ progressbar {
   border-radius: 0;
 }
 
+trough {
+  background-color: @trough;
+}
+
 progress {
   background-color: @progress;
 }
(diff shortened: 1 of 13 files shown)


dhh:
Hmm, I actually prefer the grey in that example. Also just checked all the other standard themes, and imo it looks good as we have it now. Appreciate the suggestion, though, but prefer to stick with what we have.

artfwo:
@dhh no problem. the standard themes could retain grey color, but extra themes could still benefit from the customizable background here. should you decide in favor of including that, i'm happy to modify the PR accordingly.

PR #3107, opened 2025-11-03
Fix HDR omarchy-cmd-screenshot

jtaw5649:
### Summary

  - Switch the screenshot command over to hyprpicker so the selection overlay renders correctly on HDR monitors
  - Remove the legacy wayfreeze dependency from the base package manifest
  - Add migration 1762156000.sh to drop wayfreeze installs

### Notes 
- Compatible with [2953](https://github.com/basecamp/omarchy/pull/2953) and [3015](https://github.com/basecamp/omarchy/pull/3015)
- Satty Save as (Ctrl+Shift+S) not working is related to Satty, see: [#332](https://github.com/Satty-org/Satty/issues/332) and [152](https://github.com/Satty-org/Satty/issues/152)

### DIFF
--- a/bin/omarchy-cmd-screenshot
+++ b/bin/omarchy-cmd-screenshot
@@ -51,30 +51,27 @@ get_rectangles() {
 
 # Select based on mode
 case "$MODE" in
-region)
-  wayfreeze &
-  PID=$!
-  sleep .1
-  SELECTION=$(slurp 2>/dev/null)
-  kill $PID 2>/dev/null
-  ;;
-windows)
-  wayfreeze &
-  PID=$!
-  sleep .1
-  SELECTION=$(get_rectangles | slurp -r 2>/dev/null)
-  kill $PID 2>/dev/null
-  ;;
-fullscreen)
-  SELECTION=$(hyprctl monitors -j | jq -r '.[] | select(.focused == true) | "\(.x),\(.y) \((.width / .scale) | floor)x\((.height / .scale) | floor)"')
-  ;;
-smart | *)
-  RECTS=$(get_rectangles)
-  wayfreeze &
-  PID=$!
-  sleep .1
-  SELECTION=$(echo "$RECTS" | slurp 2>/dev/null)
-  kill $PID 2>/dev/null
+  region)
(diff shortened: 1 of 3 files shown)


jtaw5649:
Apologies, I was rushing when I made this.

michaelaguiar:
I am actually still having the HDR issue while the monitor is set to 10bit color. Screenshots are super bright and washed out even with this update. I am using an Nvidia card.

Anyone else still have this issue?

vitaliiorlov:
> I am actually still having the HDR issue while the monitor is set to 10bit color. Screenshots are super bright and washed out even with this update. I am using an Nvidia card.
> 
> Anyone else still have this issue?

Same for me.

michaelaguiar:
I found the issue with this.

When this freezes the frame with hyprpicker (wayfreeze has the same issue), and you have HDR enabled in hyprland, the screenshot is super bright, caused by some issue with hyprpicker grabbing a frame from the compositor.

If I comment out the hyprpicker lines, the screenshot is the correct color, it just doesn't freeze the frame.

```
# hyprpicker -r -z >/dev/null 2>&1 & PID=$!
# sleep .1
```

Anyone have an idea how to fix the color issue with the frozen frame screenshot?

jtaw5649:
I initially tried a few different ways to fix the issue with wayfreeze but its a tricky issue. 

> I found the issue with this.
> 
> When this freezes the frame with hyprpicker (wayfreeze has the same issue), and you have HDR enabled in hyprland, the screenshot is super bright, caused by some issue with hyprpicker grabbing a frame from the compositor.
> 
> If I comment out the hyprpicker lines, the screenshot is the correct color, it just doesn't freeze the frame.
> 
> ```
> # hyprpicker -r -z >/dev/null 2>&1 & PID=$!
> # sleep .1
> ```
> 
> Anyone have an idea how to fix the color issue with the frozen frame screenshot?

I did initially try a few different ways to fix the issue with wayfreeze. As you said one outcome is not freezing the frame (which was unacceptable for me). I provided a minimal fix that worked for me at the time. It's a tricky issue and I wish you the best of luck.

Hopefully more people can get on board with testing a more permanent solution so it doesn't pop back up.

PR #3150, opened 2025-11-04
Add system language option to update menu

LukasTrust:
This PR adds a system language change option to the update menu, similar to the existing timezone option. When a new language is selected, Waybar restarts automatically to apply it immediately.

I’m not entirely sure if a full reboot or additional steps are required for the change to fully take effect, but this makes the option visible and functional in the menu.


LukasTrust:
[Related discussion](https://github.com/basecamp/omarchy/discussions/3161)

Issue #1147, opened 2025-08-26
Turning off DPMS causes Hyprlock to crash, requiring hard reboot of Omarchy


nkasco:
Whenever my machine goes idle and the lockscreen displays, hypridle.conf has a configuration listener to turn off dpms.

This puts the screen to sleep, and for my LG ultrawide monitor when the light goes from white to amber when it doesn't detect input, this seems to be consistently crashing hyprlock each time.

**Repro Instructions:** Simulating the commands in omarchy-lock-screen allows me to repro this at will.
```loginctl lock-session;sleep 3;hyprctl dispatch dpms off;sleep 10;hyprctl dispatch dpms on && brightnessctl -r```
(Really what seems to be doing it is the hyprctl dispatch dpms off and having the monitor off for long enough that it doesn't detect input, if I do this with shorter intervals the issues doesn't repro since the monitor hasn't lost input detection)

**Workaround:** Comment out the listener in ~/.config/hypr/hypridle.conf that runs hyprctl dispatch dpms off
```
#listener {
#    timeout = 330                                            # 5.5min
#    on-timeout = hyprctl dispatch dpms off                   # screen off when timeout has passed
#    on-resume = hyprctl dispatch dpms on && brightnessctl -r # screen on when activity is detected
#}
```
While this will lead to the screen locking but next turning off the monitor, it's better than crashing.

**Proposed Omarchy Workaround:** Provide a setting in Omarchy to allow the user to choose if their screen should turn off or not when idle. This is really a hyprlock issue, but since it's provided by Omarchy we should improve the UX.

nkasco:
This was mentioned in #26 by @jonh-favo and @ethantang93 as well

nkasco:
Also - It looks like if you change your monitor configuration to go into standby instead of powering off (this is the verbiage on my LG monitor OSD at least) it seems to be preventing the issue.

I also run my peripherals through my monitor, so without monitor power I can't wake with a keyboard or mouse command anyway.

fumita:
I’m using the same LG monitor and experiencing the same issue, which I’m working around with the same workaround. 😢

imanubdesigner:
I was browsing the web to see if the developer or people like DHH had anything to say about the Hyprlock crash. From what has been gathered (and as @nkasco rightly pointed out), it's due to the monitor suspending, or Deep Sleep, which unfortunately I can't disable on my LG monitor. Replacing it just for this one feature seems a bit excessive.

[...]

PR #3158, opened 2025-11-05
Fix popped windows appearing over screensaver

d-cas:
## Problem
Popped windows (Super+O feature from v3.1.5) appear over the screensaver, defeating its privacy/security purpose.

**Root cause:** In Hyprland, pinned windows have higher z-order than fullscreen windows. The Super+O pop feature pins windows, and the screensaver uses fullscreen mode. Result: popped windows stay visible over the screensaver.

## Solution
Temporarily unpin all windows tagged with `pop` before launching the screensaver. Windows remain floating and centered (just not pinned), allowing the screensaver to cover them. This is a one-way operation—users can Super+O again to re-pop if needed.

## Testing
- ✅ Tested on multi-monitor setup (laptop + external monitor)
- ✅ Verified with natural hypridle activation
- ✅ Confirmed screensaver covers popped windows on both screens
- ✅ Multi-monitor dismiss works correctly

## Changes
- Modified `bin/omarchy-launch-screensaver` to unpin popped windows before screensaver launch
- Added `migrations/1762301611.sh` to patch existing user installations

## Note
Opening as draft to confirm this is unintended behavior. The Super+O pop feature is new in v3.1.5, and I want to verify that popped windows appearing over the screensaver isn't by design before requesting review.

### DIFF
--- a/migrations/1762301611.sh
+++ b/migrations/1762301611.sh
@@ -1,0 +1,22 @@
+echo "Fix screensaver to cover popped windows"
+
+SCREENSAVER_SCRIPT=~/.local/share/omarchy/bin/omarchy-launch-screensaver
+
+# Skip if already patched
+if grep -q "Temporarily unpin all popped windows" "$SCREENSAVER_SCRIPT" 2>/dev/null; then
+  exit 0
+fi
+
+# Only patch if the file exists and has the expected structure
+if [[ -f "$SCREENSAVER_SCRIPT" ]] && grep -q 'focused=$(hyprctl monitors' "$SCREENSAVER_SCRIPT"; then
+  # Insert unpinning logic before the focused monitor line using sed
+  sed -i '/focused=$(hyprctl monitors/i\
+# Temporarily unpin all popped windows so screensaver covers them\
+popped_windows=$(hyprctl clients -j | jq -r '\''.[] | select(.tags[] == "pop") | .address'\'')\
+if [[ -n "$popped_windows" ]]; then\
+  while IFS= read -r addr; do\
+    hyprctl dispatch pin address:"$addr"\
+  done <<< "$popped_windows"\
+fi\
+' "$SCREENSAVER_SCRIPT"
+fi


ryrobes:
Nice. Def not intended. Maybe lets look into getting it fixed in Hyprland perhaps? I know that this has come up in hypr discussions before, being able to tweak the zindex for things like this. IMO fullscreen should always have the highest zindex, but by default pinned windows are higher (maybe for a good reason, but I can't think of one).

d-cas:
Thanks for confirming! 

I looked into this and found that pinned > fullscreen z-order is intentional Hyprland design per @vaxerski ([#1147](https://github.com/hyprwm/Hyprland/issues/1147#issuecomment-1336071062)). Omarchy's screensaver uses fullscreen mode, so pinned windows (like Super+O popped windows) stay on top.

This fix solves it by unpinning popped windows before the screensaver launches.

I'm open to opening a Hyprland issue about it, but not sure how to frame it given vaxerski explicitly said the pinned > fullscreen hierarchy is intentional. Maybe something like "exception for security-critical fullscreen windows like screensavers/lockers?" Curious what you think.

Marking ready for review.

Killer feature btw

abenz1267:
> Maybe something like "exception for security-critical fullscreen windows like screensavers/lockers?" Curious what you think.

Not much! It's not Hyprlands fault omarchy doesnt implement a proper screensaver :P. You'd expect things like screensavers, lockscreens etc be on the appropriate wayland layer: overlay.

right now even walker is visible when the screensaver is running... which is correct! Because fullscreened windows are still on the same layer as every other regular window. And thats what omarchy's screensaver is: a regular window.

d-cas:
> Not much! It's not Hyprlands fault omarchy doesnt implement a proper screensaver :P. You'd expect things like screensavers, lockscreens etc be on the appropriate wayland layer: overlay.
> 
> right now even walker is visible when the screensaver is running... which is correct! Because fullscreened windows are still on the same layer as every other regular window. And thats what omarchy's screensaver is: a regular window.

I looked into this more. You're right. The proper fix is using Wayland layer shell instead of a fullscreen window.

This should be straightforward to fix with a minimal wrapper that creates a proper overlay layer and just pipes the existing terminaltexteffects animations into it. Basically keep all the current aesthetic but render it in the right layer. Currently testing Omarchy with systemd hooks on one system and Omarchy w/ dracut on a different system, but if I have some time I'll take a crack at implementing this properly. Also happy to collab if anyone is interested.

dhh:
Appreciate the suggestion here, but the screensaver is not there to provide a locked/secure setup. We go to the lock screen after a timeout for that. The screensaver is just there for a visual effect.

PR #3159, opened 2025-11-05
Prompt adornment in Nix env.

brianblakely:
Helpful indicator that you are working in a Nix shell env.
image

Not shown - when the package version is set, it will render as such: `shell-name@1.2.3`

### DIFF
--- a/config/starship.toml
+++ b/config/starship.toml
@@ -1,6 +1,6 @@
 add_newline = true
 command_timeout = 200
-format = "[$directory$git_branch$git_status]($style)$character"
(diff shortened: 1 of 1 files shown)


brianblakely:
I also considered having only the Nix icon, sans package info. However, I feel that did not disambiguate enough in a Nix-heavy workflow.

dhh:
I think this is a good addition to a user-specific config, but I don't want to open the door to the default starship config having a million env-specific configs, so going to skip this for the default. Thanks for the suggestion, though ✌️

PR #3174, opened 2025-11-05
Scale with the right monitor

y0no:
`omarchy-windows-vm` take the first monitor from `hyprtctl monitors` instead of the active one.

This PR choose the active monitor instead.


geoweir:
thanks

PR #3182, opened 2025-11-05
Implement language selection for Windows OS installation

y0no:
Added language selection feature to the installation script.

### DIFF
--- a/bin/omarchy-windows-vm
+++ b/bin/omarchy-windows-vm
@@ -183,7 +183,7 @@ Turkish
 Ukrainian
   '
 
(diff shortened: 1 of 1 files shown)


lypanov:
Doesn't it make more sense for this a) work automatically based on the selections that you've made already for Omarchy, and b) to set the keyboard correctly? A random note - you defaulted it to French, I'd probably go with English. I know this is biased, but it is what it is. P.S. I am just a random person. I am not a contributor. This is not an official response. ;)

y0no:
You are right, this will be clearly a better choice to implement automatic configuration from Omarchy settings (at least auto select settings). The only difficulty concerns the diff between available languages on Omarchy and the one supported by dockurr/windows. 

The default french selection  is a mistake from my configuration (woops).

PR #3184, opened 2025-11-05
Feat: Generate & apply Zed theme based on current theme

ryrobes:
`omarchy-theme-set-zed`

**Simple**: Creates & applies a Zed theme (if installed) when the system theme changes. Colors will be pulled from theme's walker.css file. 

https://github.com/user-attachments/assets/75b181f9-aad3-4464-8fdd-f1db2ac01e30

Will _optionally_ use a `zed.json` file in the theme directory as an override in this simple format, if it exists. `zed.json` can have all or partial keys, will be merged with the generated theme map.

```json
 {
   "colors": {
     "background": "#1e1e2e",
     "foreground": "#cdd6f4",
     "text": "#cdd6f4",
     "selected": "#89b4fa",
     "border": "#89b4fa",
     "base": "#1e1e2e"
   },
   "syntax": {
     "comment": "#6c7086",
     "keyword": "#cba6f7",
     "string": "#a6e3a1",
     "number": "#fab387",
     "type": "#f9e2af",
     "operator": "#f5c2e7",
     "function": "#89b4fa"
   }
 }
```

It will create the theme file format needed, "install" it, AND set Zed to use that theme. Should update immediately. A {THEME_NAME} theme will be seen the available Zed themes.

**Tests**: Installed a fresh copy of Omarchy 3.1.5, installed Zed, changed the theme (actual testing in the video above).

### DIFF
--- a/applications/theme-generator.desktop
+++ b/applications/theme-generator.desktop
@@ -1,0 +1,8 @@
+[Desktop Entry]
+Type=Application
+Name=Theme Generator
+Exec=omarchy-theme-generate --walker  # Replace with full path to your script
+Icon=preferences-desktop-theme
+Categories=Utility;
+Terminal=false
+StartupNotify=false
--- a/bin/omarchy-theme-generate
+++ b/bin/omarchy-theme-generate
@@ -1,0 +1,1759 @@
+#!/bin/bash
+
+# Omarchy Theme Generator
+# Generates a theme from an input image using ImageMagick for color extraction
+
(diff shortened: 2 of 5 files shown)


morgoth:
This was already proposed https://github.com/basecamp/omarchy/pull/719

It's pretty easy now to support this with omarchy hook integration.
There is https://github.com/APS6/omazed based on hooks

ryrobes:
Nice. Seems a lot more complicated though, no?

ryrobes:
Close for new PR.

PR #3190, opened 2025-11-05
Add minimal prompt indicator for ssh

slinn-01:
Adds a minimal and non-intrusive indicator to starship.toml for ssh connections.

### DIFF
--- a/config/starship.toml
+++ b/config/starship.toml
@@ -1,6 +1,6 @@
 add_newline = true
 command_timeout = 200
-format = "[$directory$git_branch$git_status]($style)$character"
+format = "[$directory$git_branch$git_status$hostname]($style)$character"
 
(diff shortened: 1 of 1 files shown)


slinn-01:
If you're anything like me, you've probably installed Omarchy on a handful of machines. I routinely ssh between them to copy things or check various things, and when looking at a terminal prompt I find it helpful to see a little indicator that I'm ssh'd otherwise it can be easy to forget which machine I'm on (I generally structure and mirror my setups from machine to machine).

dhh:
That's a pretty neat and subtle addition to starship, but I'm not sure it's something I'd rock myself. Find it a little confusing that it's only there when I'm connecting to another Omarchy machine. Great addition to your local setup, though ✌️

PR #3196, opened 2025-11-05
Add check if we are running on battery power

erans:
Check if we are indeed running on battery and the batter is discharging and show an appropriate alert.

If needed we can also check the amount of battery left and make a better informed decision on the text to display to the user.


dhh:
We now warn explicitly and require confirmation before starting the update. I think that's enough. But thanks for suggesting!

PR #3202, opened 2025-11-06
feat: tee screensaver automatically loads screen refresh rate

liaobinbin:
The default 240 will cause high CPU usage during screensaver, and my laptop fan will be noisy, which is not very friendly in the office, and some monitors only have a maximum frame rate of 60 frames per second


AndrewMast:
It appears like a lot of changes are being made to screensavers over on the [`dev`](https://github.com/basecamp/omarchy/compare/master...dev) branch, maybe this could be included over there?

liaobinbin:
Okay, I will submit a new modification to the dev branch

PR #3203, opened 2025-11-06
Fix screensaver animation offset between effects

rwwiv:
## Problem
The screensaver animations were offsetting upward after each effect completed.

## Solution
- Added `--no-eol` flag to the `tte` command to suppress the trailing newline emitted after each animation
- Changed canvas height from `$(tput lines) - 2` to `0` for automatic terminal height detection

## Testing
Verified that animations now transition cleanly without offset between effects.


AndrewMast:
Hey! This same edit was already merged into the `dev` branch with PR #2648, but it has yet to be released.

rwwiv:
Ah great! Didn't catch `dev` is the target branch for contrib, thanks for the heads up!

PR #3206, opened 2025-11-06
feat: tee screensaver automatically loads screen refresh rate

liaobinbin:
The default 240 will cause high CPU usage during screensaver, and my laptop fan will be noisy, which is not very friendly in the office, and some monitors only have a maximum frame rate of 60 frames per second


dhh:
We run at 240fps because I found the standard 60fps to feel much too slow for the effects. It is CPU intensive, though. There's no way of getting around that! But it is as designed right now. The TTE folks are considering ways to make it more efficient, though.

PR #3207, opened 2025-11-06
Feature: Add bluetui integration and migration

a-sologub:
## Summary
Add bluetui integration to replace blueberry as the Bluetooth TUI.

## What's New
- New `omarchy-launch-bluetooth` launch script
- Automatic migration from blueberry to bluetui
- Updated configurations for waybar and menu

### DIFF
--- a/migrations/1762417304.sh
+++ b/migrations/1762417304.sh
@@ -1,6 +1,9 @@
-echo "Install bluetui as new bluetooth selection TUI"
+echo "Replace bluetooth GUI with TUI"
 
-if omarchy-cmd-missing bluetui; then
-  omarchy-pkg-add bluetui
-  omarchy-refresh-waybar
+omarchy-pkg-add bluetui
+omarchy-pkg-drop blueberry
+
+if ! grep -q "omarchy-launch-bluetooth" ~/.config/waybar/config.jsonc; then
+  sed -i 's/blueberry/omarchy-launch-bluetooth/' ~/.config/waybar/config.jsonc
+  omarchy-restart-waybar
 fi


kromsam:
I hate to break it to you, but dhh does not like bluetui. You haven't been the first one to try this.

https://github.com/basecamp/omarchy/pull/1260#issuecomment-3238041519

a-sologub:
I appreciate the heads up. That’s unfortunate, I feel the TUI looks really authentic and fits Omarchy's design well

kromsam:
I agree... anything is better than blueberry honestly. pythops has taste :)

dhh:
I just gave bluetui another look. It has improved a lot (especially by having the keybindings exposed, like Impala now does too)! But one blocker for me is that when I go to pair, say, a set of airpods, it'll connect, but won't be "trusted" by default. So now I have to notice that and manually hit "t" to trust. That's pretty confusing imo. If @pythops changes that at some point, I think we're almost there.

pythops:
@dhh yep, that's something I can change for sure. 
anything else ?

I hear also from users that they want the device to be connected automatically after pairing, instead pairing, going to paired devices then pressing enter to connect. is that a behavior you want to have  as well ?

Rishabh-Sarang:
> @dhh yep, that's something I can change for sure. anything else ?
> 
> I hear also from users that they want the device to be connected automatically after pairing, instead pairing, going to paired devices then pressing enter to connect. is that a behavior you want to have as well ?

That would be great

dhh:
> @dhh yep, that's something I can change for sure. anything else ?
> 
> I hear also from users that they want the device to be connected automatically after pairing, instead pairing, going to paired devices then pressing enter to connect. is that a behavior you want to have as well ?

Yes, that would be excellent. Thanks for the quick response, and thanks for all your work on these tools!

pythops:
Cool, I will fix it then. 
You are welcome :)

pythops:
I have just released a new version `v0.8.0` that has the requested changes.
Give it a try guys and let me know how it works for you. 
Feedback is always welcome :) 
https://github.com/pythops/bluetui/releases/tag/v0.8.0

dhh:
@pythops I just installed 0.8.0, but when launching it on Omarchy now, I just see a blank terminal. Maybe some regression?

pythops:
:thinking: not that I am aware of. I will check
Meanwhile,  you can try this:

1. make sure bluetooth service is running
```
systemctl status bluetooth
```

2. Make sure `bluetooth` is not blocked in the kernel
```
sudo rfkill unblock bluetooth
```
3. Try to launch `bluetui` from a terminal instead

pythops:
I have doubled checked, run `bluetui` on different terminals, different machines, asked people to run it and it seems to be working fine. Maybe one of the points I mentioned above seems to be the issue for you.

dhh:
Got it working. I think I had a pending kernel update that was screwing things. But would be great if we can improve the error mode such that if it doesn't detect bluetooth underpinnings it'll show an error instead of a blank screen.

PR #3210, opened 2025-11-06
Feat: Full dynamic theme gen from images (+GTK, Zed, *VSCode - Walker UI)

ryrobes:
## End-to-end theme generation from the Omarchy menu
Takes a single image (or picks a random image from a folder), pulls colors, generates a full Omarchy theme from it, and sets it (as theme "Dynamic"). Very easy to use - fully runnable directly in the menu - it's CLI can be scripted, bound, cron scheduled, etc. 
i.e `omarchy-theme-generate ~/Wallpapers/ my-cool-theme-name`

https://github.com/user-attachments/assets/ae94545f-1c08-48c5-8d09-4fce44346967

- Adds GTK theming (either from image or alacritty.toml)
   - See Nautilus in the video
- Adds Zed theming (from image, read from an existing zed.json, or from walker.css)
- Simple Walker UI file/folder selector via the Omarchy menu
   - Default Walker usage writes to a theme called "Dynamic", that gets overwritten on each run
   - Walker script re-uses the main script with a `--walker` arg to keep the file sprawl down
- Uses a modified VSCode theme system (backwards compatible with current functionality) that generates the theme and applies it directly so it does not require a pre-installed VSCode theme
- Can be used to generate themes from images "offline" w/o switching the current theme - writing them to the themes directory under an arbitrary theme name (CLI usage)
- Shows a fancy notify-send toast with preview image while generating

**Testing:** Installed on a fresh 3.1.5 instance, all existing themes work fine (with added Zed and GTK support derived from existing base files).

> You can test it on your own by installing the modified scripts from this repo:
[https://github.com/ryrobes/omarchy-theme-generate](https://github.com/ryrobes/omarchy-theme-generate)
(simple install script, sample images, etc)

_Note: Yes, it's a lot of bash - but quite straightforward - fits with the "Beautiful & Modern" tagline well. No extra deps or weird complexity.  Besides, few distros have anything like this out of the box - just like few have good unified (and expanding) theme systems like Omarchy._

### DIFF
--- a/bin/omarchy-theme-generate
+++ b/bin/omarchy-theme-generate
@@ -33,7 +33,7 @@ build_folder_cache() {
     mkdir -p "$CACHE_DIR"
 
     echo "Scanning for image folders..."
-    notify-send "Omarchy Theme Generator" "Scanning for image folders..." -t 4000
+    notify-send "Omarchy Theme Generator" "Scanning for image folders..." -t 6000
 
     # Create temporary file for atomic write
     local temp_cache="$CACHE_FILE.tmp.$$"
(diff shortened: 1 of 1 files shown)


ethandpowers:
How cool is that??  Great work!

brianspragge:
I 100% support this addition!

dhh:
Very interesting exploration! But we can't carry 3,500 lines of code for this. I'd suggest making an AUR package for it. Thanks for working on it ✌️

PR #3213, opened 2025-11-06
Feature: Heroic Games Launcher (for Epic Games) to the Omarchy > Install > Gaming Menu

chancedomore:
When first installing omarchy i had to find a way to be able to play rocket league or an Epic Games Launcher equivalent for linux. This seemed like the most straight forward option. Heroic can handle as an equivalent for Epic Games, GOG, and Prime games

For more info on the app:
https://heroicgameslauncher.com/

Lutris could also be another option.

https://www.reddit.com/r/linux_gaming/comments/1ipqtel/what_are_the_pros_and_cons_of_lutris_vs_heroic/ 

image

### DIFF
--- a/bin/omarchy-menu
+++ b/bin/omarchy-menu
@@ -301,8 +301,9 @@ show_install_ai_menu() {
 }
 
 show_install_gaming_menu() {
-  case $(menu "Install" "  Steam\n  RetroArch [AUR]\n󰍳  Minecraft") in
+  case $(menu "Install" "  Steam\n  Heroic Games Launcher\n  RetroArch [AUR]\n󰍳  Minecraft") in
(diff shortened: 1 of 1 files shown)


dhh:
Appreciate the suggestion, but I just tried to install it, and I found it rather confusing. Logged in with my Epic Games account, wasn't able to install more from the store, just saw the Fortnite game (which won't run anyway). I get that this allows some people to play stuff they otherwise wouldn't have, but I think there are too many caveats to making this a general recommendation.

chancedomore:
@dhh After submitting this PR i found this already active thread around this. Feels like there are some good ideas with a fair amount of work to make it more seemless. 

https://github.com/basecamp/omarchy/pull/1863

PR #3215, opened 2025-11-06
Remove unnecessary prints

guibbv2011:
principally this piece of code:
`echo "${default_transform}" > ~/Work/output.txt`

that's was used for debug purpose only


guibbv2011:
Sorry for this, I'm will update a previous PR #3037

PR #3221, opened 2025-11-07
Add installation option for GitHub Copilot CLI

tomschlick:
Added an installation option for the GIthub Copilot CLI under the Install -> AI menu.

https://aur.archlinux.org/packages/github-copilot-cli
https://github.com/features/copilot/cli


dhh:
Appreciate the suggestion, but I think we're OK now in terms of recommendations (also this PR is incomplete, didn't change the menu, didn't use AUR install).

PR #3237, opened 2025-11-07
fix: use xdg-terminal-exec in TUI installer for terminal compatibility

shelldandy:
Replace hardcoded --class flag with xdg-terminal-exec --app-id to properly support kitty and other terminals. This aligns with the rest of the codebase and ensures floating window rules work correctly across different terminals.

Otherwise TUIs weren't floating windows ever when using with kitty or similar terminal apps


dhh:
Would be nice to also do a migration that moves all the existing TUIs with the old format over.

PR #3245, opened 2025-11-08
Modify clipboard copy binding for Zed application

bombless:
Updated clipboard bindings to make universal copy work on Zed editor.


dhh:
Interesting solution, but that's not something we can have as part of the default bindings. Too specific to Zed. I think a better option is to figure out how to configure Zed such that it responds to the Ctrl + Insert/Shift + Insert combo we're using for everything else.

bombless:
Maybe we can add an argument to zeditor and make them handle super+c differently. Not sure if it's possible to contribute to Zed in this way.

Also, universal copy don't work on nautilus either.

PR #3252, opened 2025-11-08
Fix: Add missing fullscreen keybind to tiling.conf

Diing54:
This PR adds the keybind for toggling fullscreen (SUPER + F), which was missing from the tiling.conf file.

I understand there is tiling-v2.conf which i believe is the version 2 but the bindd = SUPER, F, Full screen, fullscreen, 0  was non-functional or maybe it was not being loaded well


dhh:
`Super + F` was not part of the original keybinding set. That's why we have a v2. So we didn't override someone's personal bindings like that. The original key was F11.

Diing54:
Alright !! Understood

Issue #3254, opened 2025-11-08
Config errors on update to 3.1.6


Mayank-8127:
### System details

i5 12500h, intel iris xe integrated graphics, omarchy 3.1.6

### What's wrong?

On updating to the most recent version, I got many config errors that were constantly showing on display

<img width="1390" height="77" alt="Image" src="https://github.com/user-attachments/assets/b209026e-d16e-4912-a9b2-8c3e98dceffc" />

And on commenting only the line causing the first error itself made the errors disappear. I dont know if it broke any functionality or something, but there is nothing instantly obvious.

Line that i commented => `windowrule = tag +pip, title:(Picture.{0,1}in.{0,1}[Pp]icture)`
File path => `/home/jarvis/.local/share/omarchy/default/hypr/apps/pip.conf`

muhammedaksam:
Same issue on my end.

<img width="1908" height="90" alt="Image" src="https://github.com/user-attachments/assets/78172b4f-1164-46cc-b978-d77306722351" />

ankurkotwal:
I'm also having the same issue.

bartdorsey:
An alternative line 2 which does the same thing but doesn't trigger the parsing bug is:

```windowrule = tag +pip, title:(Picture.?in.?[Pp]icture)```

I believe this was caused by a change to the hyprlang parsing library.

mikailbayram:
Created a PR with the fix:  [PR #3255](https://github.com/basecamp/omarchy/pull/3255)

ankurkotwal:
Thank you @bartdorsey and @mikailbayram - the suggested fix works for me.

eerison:
Thanks @mikailbayram, it also fixed the issue on my side. (#3255)

baremarco:
> An alternative line 2 which does the same thing but doesn't trigger the parsing bug is:
> 
> `windowrule = tag +pip, title:(Picture.?in.?[Pp]icture)`
> 
> I believe this was caused by a change to the hyprlang parsing library.

Thank you this solution also works for me @bartdorsey

tlxq:
Thx for solution! @bartdorsey

sk4rrjin:
Tried to share it as a [discussion to hyprland](https://github.com/hyprwm/Hyprland/discussions/12238), not sure they wanted to break the regex parsing.

dhh:
Fixed in 3.1.7 👍

PR #3255, opened 2025-11-08
Fix: adjust windowrule regex for pip to avoid Hyprland parsing error

mikailbayram:
## Fix windowrule regex parsing in Hyprland
**Problem:**
The previous regex line in `pip.conf`:

`windowrule = tag +pip, title:(Picture.{0,1}in.{0,1}[Pp]icture)`

caused a parsing error in recent versions of Hyprland:

image

**Solution:**
Replaced `{0,1}` with `?`, which is equivalent in regex and compatible with Hyprland's parser:

PR #3258, opened 2025-11-08
Update pip.conf, fix Invalid rule/parameter errors after hyprland update

sk4rrjin:
Updated hyprland throws Invalid rule / Invalid parameter on the pip windowrule; reported in  #3254

### DIFF
--- a/default/hypr/apps/pip.conf
+++ b/default/hypr/apps/pip.conf
@@ -1,5 +1,5 @@
 # Picture-in-picture overlays
-windowrule = tag +pip, title:(Picture.{0,1}in.{0,1}[Pp]icture)
+windowrule = tag +pip, title:(Picture.?in.?[Pp]icture)
 windowrule = float, tag:pip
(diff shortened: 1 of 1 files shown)


d-cas:
Great work getting this PR up so quickly @sk4rrjin! 

This fix works, but FYI @vaxerski is investigating whether this is actually a parser bug in Hyprland 0.52.0

See https://github.com/hyprwm/Hyprland/discussions/12238. The comma in {0,1} might be getting treated as a delimiter. Either way, this PR should be merged as the immediate fix, and we can revisit if/when it's fixed upstream.

lukehsiao:
Duplicate of #3255

d-cas:
> Duplicate of #3255

Community was all over this one!

Muhiddin0:
Hi guys! I just want to say, this is the most amazing OS I've ever seen.
What do you think is better — installing the latest version of the OS, or using the installation script?

PR #3265, opened 2025-11-09
Feature: OAuth picker for walker

HappyNZ95:
OAuth is necessary but annoying to work with most of the time.

This PR adds a small script that lets you pick an OAuth/TOTP account from a .totp  file using walker --dmenu. Once you select an account, it generates the current TOTP code and copies it to your clipboard.

**Depends on**
oathtool (oath-toolkit)

**How it works:**
Type in your OAuth code into the .totp file in your home directory like so:

```
github SKDFHWOEFKH
microsoft KSAJDFHSDSDFKJ
google KJSGHKSDJFHHJDS
```

Shows all account names from your secrets file in walker's dmenu
Generates a 6-digit TOTP code with oathtool from your selection
Copies the code to the clipboard
Sends a confirmation notification with the name and code or an error message

My personal keybinding is SUPER CONTROL, O (OAuth)

**Why this is useful:**
It makes accessing TOTP codes super easy without having to open a seperate app, pull out your phone, etc.

screenshot-2025-11-09_21-42-01

If this is of interest I can integrate with the menu under Setup > Security > OAuth

### DIFF
--- a/bin/omarchy-oauth-menu
+++ b/bin/omarchy-oauth-menu
@@ -1,0 +1,29 @@
+#!/usr/bin/env bash
+
+SECRETS_FILE="$HOME/.totp"
+
+if [[ ! -f "$SECRETS_FILE" ]]; then
+    notify-send "Error" "$SECRETS_FILE not found"
+    exit 1
+fi
+
+CHOICE=$(awk '{print $1}' "$SECRETS_FILE" | walker --dmenu -p "OAuth account:")
+
+if [[ -n "$CHOICE" ]]; then
+    SECRET=$(awk -v acc="$CHOICE" '$1 == acc {print $2}' "$SECRETS_FILE")
+
+    if [[ -n "$SECRET" ]]; then
+        CODE=$(oathtool --totp -b "$SECRET" 2>/dev/null)
+
+        if [[ -n "$CODE" ]]; then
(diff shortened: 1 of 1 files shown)


abenz1267:
What tool are you using to get the `$HOME/.totp`-file to begin with?

Omarchy does ship with 1Password by default. I'm not sure if maybe @dhh wants to integrate the upcoming native 1Password provider for Walker instead.

HappyNZ95:
> What tool are you using to get the `$HOME/.totp`-file to begin with?
> 
> Omarchy does ship with 1Password by default. I'm not sure if maybe @dhh wants to integrate the upcoming native 1Password provider for Walker instead.

This was just entered manually by copying the setup code provided when setting up a new  'authenticator app' in Microsoft etc.
If there's an integration for 1Password's authenticator in walker that would be ideal.

dhh:
This is an interesting idea, and thanks for suggesting it, but not something I'd personally use. Maybe we can do this as an external package for those who want this?

abenz1267:
> This is an interesting idea, and thanks for suggesting it, but not something I'd personally use. Maybe we can do this as an external package for those who want this?

There's https://github.com/abenz1267/elephant-community where users can submit custom menus for elephant. They can be easily installed via `elephant community install`.

PR #3271, opened 2025-11-09
Ensure Linux kernel is running preempt=full to avoid audio/video jitter under load

dhh:
We never want audio/video playback or recording to crackle under load.

### DIFF
--- a/install/login/limine-snapper.sh
+++ b/install/login/limine-snapper.sh
@@ -33,7 +33,7 @@ TARGET_OS_NAME="Omarchy"
 ESP_PATH="/boot"
 
 KERNEL_CMDLINE[default]="$CMDLINE"
-KERNEL_CMDLINE[default]+="quiet splash"
+KERNEL_CMDLINE[default]+="quiet splash preempt=full"
 
 ENABLE_UKI=yes
 CUSTOM_UKI_NAME="omarchy"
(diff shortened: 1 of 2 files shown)


dhh:
cc @flavorjones

abenz1267:
Is this needed?  `sudo dmesg | grep Preempt` on Arch should output `full` by default anyways.

Have you checked? I'm 99% sure it's the default for Arch.

dhh:
You're right. Good catch, @abenz1267 👌. I was having some stuttering issues on a new Beelink GTR9P and also caught some when recording a video on the Framework Desktop, so it seemed to square up that this was the culprit, but just verified and we are indeed already in full. So problem must be elsewhere!

geldot:
> You're right. Good catch, @abenz1267 👌. I was having some stuttering issues on a new Beelink GTR9P and also caught some when recording a video on the Framework Desktop, so it seemed to square up that this was the culprit, but just verified and we are indeed already in full. So problem must be elsewhere!

Check your clocksource in dmesg logs. The Kernel is known to fall back from TSC to less efficient HPET timers on some AMD Beelink models due to a firmware quirk. The fallback can be overridden with `clocksource=tsc tsc=reliable` which fixes the stuttering at your own risk. My understanding is that the correct fix would be to measure the actual boot timing offset during boot and implement an adjustment in the kernel.

PR #3274, opened 2025-11-09
feat: new icon for connected bluetooth waybar

vander00:
## The problem:
I didn't know if I was connected to my bluetooth buds without opening blueberry.

## Solution:
I just added "!" symbol to connected bluetooth so it's possible to easily see if you're connected or not.

### DIFF
--- a/config/waybar/config.jsonc
+++ b/config/waybar/config.jsonc
@@ -99,7 +99,7 @@
   "bluetooth": {
     "format": "",
     "format-disabled": "󰂲",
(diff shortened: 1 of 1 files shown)


vander00:
image

How it looks.

brink-lab:
I would suggest to change the icon for "disabled" and "not connected" states instead. There are already established practices for this.

dhh:
Hmm, don't love the aesthetics of this

PR #3281, opened 2025-11-09
Dynamically set hyprlock font and placeholder text

mlombardi96:
Currently, hyprlock has a fixed font, and its placeholder text always includes the fingerprint icon even if fingerprint auth isn't setup.

This PR fixes this by:
1. Updating `omarchy-font-set` to change the hyprlock font
2. Updating `omarchy-setup-fingerprint` to add/remove the fingerprint icon from the hyprlock placeholder text

Dynamic font demo:

https://github.com/user-attachments/assets/c9254852-d7b0-449d-aba9-8ced00ab097d

Dynamic fingerprint icon demo:

https://github.com/user-attachments/assets/c4eb70a3-eb83-4df5-8332-c47ad84c7b41

### DIFF
--- a/config/hypr/hyprlock.conf
+++ b/config/hypr/hyprlock.conf
@@ -22,10 +22,10 @@ input-field {
     outer_color = $outer_color
     outline_thickness = 4
 
-    font_family = CaskaydiaMono Nerd Font Propo
+    font_family = CaskaydiaMono Nerd Font
     font_color = $font_color
 
-    placeholder_text = Enter Password 󰈷
+    placeholder_text = Enter Password
     check_color = $check_color
     fail_text = <i>$FAIL ($ATTEMPTS)</i>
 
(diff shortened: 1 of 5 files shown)


config/hypr/hyprlock.conf
```diff
@@ -22,10 +22,10 @@ input-field {
     outer_color = $outer_color
     outline_thickness = 4
 
-    font_family = CaskaydiaMono Nerd Font Propo
+    font_family = CaskaydiaMono Nerd Font
```
mlombardi96, line 25: I was the one that changed this to the `Propo` version of the font. I originally did this to fix spacing issues in the placeholder text.

However, now that we are dynamically changing the font, we can't rely on every font having a `Propo` variation.

In order to fix the spacing issues now, the placeholder text is wrapped in a `<span>` with whitespace at the beginning and end of the text. Visually, it looks the same before and after.

dhh:
Very nicely done!

Issue #3218, opened 2025-11-07
Omarchy screensaver triggers active app to exit fullscreen mode


rmusaev99:
### System details

msi laptop intel + nvidia, 2k display

### What's wrong?

So when I have alacritty opened in full screen mode when omarchy starts showing screensaver when I press esc or space to exit screensaver alacritty opens in basic mode, not full screen. This is a little but a bit annoying thing. The reason why it matters is because when I'm not in the full screen mode the gap between application and top bar starts annoyingly jumping - I have all the paddings set to 0 so UI parts are supposed to touch each other but instead I have gap that goes from 1px to 0px right when I type. Not sure why, maybe because my display is 2k

abenz1267:
Makes sense, because omarchy's "screensaver" is just a fullscreened window, so whatever other fullscreen window one has => not fullscreen anymore.

MGross21:
Can definitely do something that returns the state of the prior window. Currently screensaver just checks the focused monitor:

https://github.com/basecamp/omarchy/blob/8350b84eee10601efce64efef6fe044f03bfcd32/bin/omarchy-launch-screensaver#L16

Can add something to remember the prior window state, if full screen. Something like:
```bash
prior_fullscreen=$(hyprctl activewindow -j | jq -r '.fullscreen')
```

Then, to restore later if it was full screen:
```bash
hyprctl dispatch focuscurrentorlast # restore focus to prior window, not just monitor

if [ "$prior_fullscreen" -eq 1 ]; then
    hyprctl dispatch fullscreen
fi
````

Hopefully, this is what you were referring to.

dhh:
Happy to take a PR to restore the prior window.

SamCod3:
**Quattro-specific reproduction and implementation note**

I can confirm the same class of fullscreen-state restoration problem on
Omarchy Quattro `dev (c7f7e837)` with Hyprland `0.56.0`. This has not been
tested on stable Omarchy.

The case I tested uses Brave with Hyprland's split fullscreen state:

- before the idle/lock transition: `fullscreen = 2`,
  `fullscreenClient = 0`;
- after the screensaver/lock transition, the internal state can remain `2`
  while `fullscreenClient` becomes non-zero;
- Brave then behaves as if it entered real browser fullscreen and hides its
  chrome instead of returning to the exact previous `2/0` state.

Quattro's implementation differs from the one that existed when this issue
and #3284 were opened:

[...]

PR #3284, opened 2025-11-09
Fix: Restore former window fullscreen state after screensaver exit

MGross21:
# Summary

Addresses #3218 by @rmusaev99.

Currently the fullscreen launch of the screensaver will override any former full-screen windows and will not return them to their state upon closing.

## Changes:

* Capture the former window and full-screen status
* Upon shutdown, revert focus to the prior window and restore full-screen status (1 and 2)
* Added null-safe jq defaults (`//`) to try and prevent any errors from launching screensaver from blank workspaces.

## Testing

I was able to test the individual commands using my custom instance of Hyprland on a laptop, but I was not able to get an instance of Omarchy running in a VM. Would be very appreciative if someone could test the changes on their own instance. Happy to make whatever tweaks are required to confirm this works on single-monitor and multi-monitor systems as well as small taste-related code changes.

## Other Approach
I have also experimented with the idea of forcing the screensaver to run on uniquely-named special workspaces, such as iterating over all monitors and running:
```bash
hyprctl dispatch togglespecialworkspace "screensaver-${MONITOR-NAME}"
```
This way, there is no conflict or need for reverting the state of any former windows on teardown. If this is something that would like to be looked into, I would be happy to make additional changes or open another PR.

### DIFF
--- a/bin/omarchy-launch-screensaver
+++ b/bin/omarchy-launch-screensaver
@@ -24,32 +24,44 @@ hypr_focus_monitor() {
   hyprctl dispatch "hl.dsp.focus({ monitor = \"$1\" })" >/dev/null 2>&1 || hyprctl dispatch focusmonitor "$1" >/dev/null
 }
 
-hypr_exec() {
-  local command="$1"
+hypr_exec_ws() {
+  local workspace="$1"
+  local command="$2"
 
-  hyprctl dispatch "hl.dsp.exec_cmd([[$command]])" >/dev/null 2>&1 || hyprctl dispatch exec -- bash -lc "$command" >/dev/null
+  hyprctl dispatch "hl.dsp.exec_cmd([[[workspace special:$workspace] $command]])" >/dev/null 2>&1 ||
+    hyprctl dispatch exec "[workspace special:$workspace] $command" >/dev/null
 }
 
+# Launch Screensaver on each monitor in a unique special workspace
 for m in $(hyprctl monitors -j | jq -r '.[] | .name'); do
   hypr_focus_monitor "$m"
 
   case $terminal in
   *Alacritty*)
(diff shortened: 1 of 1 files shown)


dhh:
Hmm, I can't get this working. If I full screen an app, then start the screensaver, it'll still bring me back to the non-fullscreen case after I quit the screensaver.

MGross21:
I will work on getting an instance of Omarchy running in a VM to test/update the code accordingly. Thanks for giving it a test!

MGross21:
Okay! With a VM version of Omarchy, I was able to pinpoint the core issue of the code, which is:

* Manually calling the screensaver from the `omarchy-menu` means that the `former_window` and `former_fullscreen` states were likely registered to `walker`, not to the fullscreen window in question. The former code (ee4d386) likely would have worked if the screensaver was invoked automatically from an idle timer, but the behavior must work in both circumstances.

Instead of writing a more complicated window history backsearch, I ended up going with my second approach:

> ## Other Approach
> 
> I have also experimented with the idea of forcing the screensaver to run on uniquely-named special workspaces, such as iterating over all monitors and running:
> 
> ```shell
> hyprctl dispatch togglespecialworkspace "screensaver-${MONITOR-NAME}"
> ```
> 
> This way, there is no conflict or need for reverting the state of any former windows on teardown. If this is something that you would like to be looked into, I would be happy to make additional changes or open another PR.

This was a much better (and working) approach as it fully sandboxes the screensavers into unique special workspaces. This cleaned up a ton of code and simplified setup/teardown.

Let me know if this works on your end! 👍🏼

MGross21:
Note: Prior walker naming collision issues fixed in #3873.

MGross21:
@dhh Ready for re-review when you have a chance. 👍🏼

MGross21:
Resolved conflicts. Migrated the changes to use the new `/bin` helpers and `lua` hl features

MGross21:
Rebased onto Quattro and cut to 3 lines. It's just the per-monitor special workspace now, so the screensaver leaves an existing fullscreen window alone.

@omarchybot

PR #3295, opened 2025-11-10
Omarchy 3.2.0

dhh:
# What changed?

_Note: Not yet released._

## New keys and capabilities

* Add new stable Arch mirror that will lag the edge by a month, so we can catch any package issues before they roll out to everyone by @dhh
* Add Ghostty as the new default terminal (existing users can switch using _Install > Terminal_) by @dhh
* Add JetBrainsMono Nerd Font as the new default font (because Caskaydia doesn't render as well in Ghostty, existing users can switch using _Style > Font_) by @dhh
* Add Ethereal and Hackerman as new theme options by @bjarneo 
* Add visual theme picker by @tahayvr 
* Add [BlueTUI](https://github.com/pythops/bluetui) instead of GTK bluetooth configuration by @a-sologub + @dhh
* Add @tobi's excellent try command by @dhh
* Add font changing to the lock screen and only show fingerprint when fingerprint sensor is active by @mlombardi96
* Add `Super + Alt + Comma` to invoke/open the last notification by @czroth 
* Add `Super + Alt + Arrows` to move between tiling group tabs by @chrislewis
* Add `Super + Shift + Alt + Arrow Left/Right` to move workspaces between monitors by @jheuing
* Add `Super + Ctrl + Shift + Alt + Arrows` to move splits in Ghostty by 100 lines (instead of default 10) by @dhh
* Add AAC audio codec for screen recordings for better compatibility by @robzolkos 
* Add floating as default when launching 1password (use `Super + T` to tile it instead) by @dhh
* Add Xbox controller setup installer under _Install > Gaming_ by @ericjim
* Add optional size + positioning parameters to `omarchy-hyprland-window-pop` command by @DarrenVictoriano 

## Fixed stuff

* Fix application launcher would only show first 50 entries by @dhh
* Fix screensaver to work with all three default terminals by @dhh + @ryanrhughes 
* Fix ability to decline an Omarchy update by @dhh
* Fix PS1 setting no longer necessary given default starship installation by @dhh
* Fix DaVinci Resolve dialog window management by @innerviewer
* Fix class name match for Microsoft Edge in window rules by @patroza
* Fix Windows scaling should be based on the focused monitor instead of the first by @y0no
* Fix TUI installer should use xdg-terminal-exec to be in line with other changes by @shelldandy 
* Fix Helium browser match in omarchy-launch-webapp by @MatyiFKBT
* Fix switching to tiling group windows with numbers on other keyboard layouts by @adam-zluva 
* Fix low contrast with sway OSD overlays on the matte black theme by @MidnightSlicer
* Fix that screen should wake on both mouse move and keyboard when it's been sleeping by @nicoestrada
* Fix wifi settings should launch or focus instead of multiple by @dhh 
* Fix missing tab completion for mise by adding `usage` as default package by @tionichm
* Fix overly fast terminal scrolling on mouse wheel in Ghostty by @dhh
* Fix walker/elephant should run via systemd instead of exec-once by @abenz1267 
* Fix terminal cwd to only kick in when the active window is a shell by @daviewales 
* Fix using consistent `org.omarchy.*` app-ids for all TUIs by @patroza 
* Fix `format-disk` function should use exfat instead of ext4 for Windows/macOS compatibility by @dhh + @daviewales 
* Fix mouse focus issues for JetBrains window adjustments by @landsman 
* Fix "Allow restore token" checkbox in screensharing options should be checked by default by @larrydewey
* Fix workspace 10 that was left out earlier by @jelenv 
* Fix image viewing with imv and video viewing with mpv should float instead of tile by default by @dhh
* Fix Everforest color scheme in Ghostty + Kitty to be the lighter variant we use for Alacritty by @dhh
* Fix waybar tray icon to be more intuitive by @brink-lab 
* Fix Bluetooth icon in tray should use the connected style when any device is connected by @vander00
* Fix walker config for breaking changes in latest version by @abenz1267 
* Fix blinking cursor on screensaver in Ghostty by upgrading to TTE 13 by @ryanrhughes 
* Fix hitting the screenrecording hotkey during a recording should stop the recording by @ryanrhughes 
* Fix screenrecording on external gpus and with HDR by using the desktop-portal-capture screen by @dec05eba
* Fix screenrecording on systems without GPUs with CPU fallback by @dec05eba

### New stable Arch mirror

With hundreds of thousands of people having downloaded the Omarchy ISO, it's time to ensure that the bleeding edge of Arch package upgrades isn't automatically rolling out to everyone without vetting. So we're introducing a new stable mirror that'll lag the latest by a month. It'll grab a snapshot of the latest repo on the first of the month, then upload that on the last Monday at 10am UTC of the month. This will allow a smaller group of package pioneers to catch problems and fix them before everyone upgrades. That should dramatically lower the risk that we see incompatibility issues in the wild at a large scale.

If you're interested in being part of the Omarchy Package Pioneers, you can run `omarchy-refresh-pacman-mirrorlist edge`. That'll put you on the edge mirror where all new packages are immediately available. Please pull early and often, and help us catch future issues!

### Ghostty is the new default terminal

[Ghostty](https://ghostty.org/) is a fantastic terminal created by @mitchellh, which natively supports splits and tabs. We've made it the default for new installations. Existing users can switch using _Install > Terminal_. Alacritty + Kitty remains as optional installs on new installations.

ghostty

You can see all keybinds on `ghostty +list-keybinds`, but here's a quick cheatsheet:

* `Ctrl + Shift + O` split vertically.
* `Ctrl + Shift + E` split horizontally.
* `Ctrl + Alt + Arrows` move between splits.
* `Ctrl + Shift + Super + Arrows` resize splits.
* `Ctrl + Shift + T` open new tab
* `Ctrl + Shift + Arrows` move between tabs.

### New themes: Ethereal and Hackerman

preview

hackerman-6k-tc

### Visual theme picker

The theme picker new has visual previews of every included theme (and every extra theme can add preview.png in 16:9 to get the same):

image

### Bluetooth TUI

With improvements to [BlueTUI](https://github.com/pythops/bluetui), we're finally able to make it match the TUI used for wifi configuration. 

image

### Tobi's try

Managing all your quick code experiments is made much nicer with @tobi's try. Automatically namespace your attempts by the data under `~/Work/tries`, delete them from the overview, and jump back into the action. See the [try page for more](https://github.com/tobi/try).

image

### DIFF
--- a/migrations/1763561396.sh
+++ b/migrations/1763561396.sh
@@ -1,4 +1,4 @@
 echo "Switch to stable Omarchy Package Repository builds"
 
 omarchy-refresh-pacman
-sudo pacman -Syu
+sudo pacman -Syu --noconfirm


default/bash/functions
```diff
@@ -25,14 +25,21 @@ format-drive() {
...
-      sudo mkfs.ext4 -L "$2" "$([[ $1 == *"nvme"* ]] && echo "${1}p1" || echo "${1}1")"
-      sudo chmod -R 777 "/run/media/$USER/$2"
-      echo "Drive $1 formatted and labeled '$2'."
+      sudo parted -s "$1" mkpart primary 1MiB 100%
+
+      partition="$([[ $1 == *"nvme"* ]] && echo "${1}p1" || echo "${1}1")"
+      sudo partprobe "$1" || true
+      sudo udevadm settle || true
+
+      omarchy-pkg-add exfatprogs
```
daviewales: Should `exfatprogs` be added to `install/omarchy-base.packages`, rather than trying to install it when this function is run?

This will ensure that this function still works if the user has never run the function, and has no internet connection.

dhh: Yes, it actually should. It's small enough.

dhh: Please do a PR 👍

daviewales: Done: https://github.com/basecamp/omarchy/pull/3448

scoop:
Can't do a PR on the PR description I'm afraid :sweat_smile: 

> You can see all keybinds on ghostty +list-keybinds, but here's a quick cheatsheet:
> [...]
> Ctrl + Shift + Arrows move between splits.

This should be Ctrl + Alt + Arrows. Since:

> Ctrl + Shift + Arrows move between tabs.

Super + Ctrl + Brackets also cycles splits, which I find mildly more ergonomic anyway.

daviewales:
> Can't do a PR on the PR description I'm afraid 😅

You can! Just set the base branch to `dev`, instead of `master`. ([example](https://github.com/basecamp/omarchy/pull/3448))

**Edit**: missed the key words "PR description". Sorry.

PR #3300, opened 2025-11-10
Use the scaling for the current monitor when opening Omarchy-Windows

geoweir:
Added a jq filter to select the scaling factor for the currently active monitor. 

For multi-monitor users where screens can have different scales this will open the RDP client in a scale suitable for that screen. Tested while running monitors displays as well as with one.


y0no:
Hi,
This is already implemented here:  https://github.com/basecamp/omarchy/pull/3174 😁

geoweir:
closed, thanks

PR #3302, opened 2025-11-10
Replace removed qt5-remoteobjects with qt6-remoteobjects

SytexMC:
Use `qt6-remoteobjects` since `qt5-remoteobjects` has been removed from the official repositories.


dhh:
I actually just removed qt5-removeobjects entirely from dev. I could not find what we were using it for!

Issue #3301, opened 2025-11-10
omarchy-update: hitting escape does not abort the update


siebertm:
### System details

Framework 13, AMD

### What's wrong?

Just stumbled over this during the 3.1.7 upgrade:

* open the `omarchy-upgrade` UI (via the button in waybar)
* decide to not do the upgrade, hit `ESC`
* the update wants to run

Expected: Update should be aborted (as when saying "no")

Note: I did not check whether selecting "no" via the menu actually works. I'd need to wait until the next release :D

sleao:
coincidentally I came here to see if anyone mentioned that hitting "no" does not cancel the update. should be a simple fix, PR is open :)

dhh:
This is fixed in the upcoming 3.2 👍

PR #3311, opened 2025-11-10
fix update not cancelling

sleao:
This should take care of #3301 as the update script wasn´t cancelling when declined. We move the confirmation up so it doens´t fall in the trap and also change the return code from the `omarchy-update-confirm` script


dhh:
This has been fixed on dev 👍

PR #3312, opened 2025-11-10
feat: add custom imv keybindings for delete and rotate currently opened image

ctarx:
This PR adds two new imv keybindings:

- Ctrl+X deletes the currently opened image and closes the viewer
- Ctrl+R rotates the current image by 90 degrees

### DIFF
--- a/migrations/1762802472.sh
+++ b/migrations/1762802472.sh
@@ -1,23 +1,4 @@
-echo "Overwrite imv config with new keybindings; backup existing config if present"
+echo "Update imv config with new keybindings"
 
-if [ -f ~/.config/imv/config ]; then
-  cp ~/.config/imv/config ~/.config/imv/config.bak.$(date +%s)
-else
-  mkdir -p ~/.config/imv
-fi
-
-cat >~/.config/imv/config <<'EOF'
-[binds]
-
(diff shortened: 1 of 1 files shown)


dhh:
This is very nice, but ctrl + x doesn't seem to close the window in my testing. It just deletes the image and leaves a black imv window behind.

ctarx:
This happens only when you open a single file. 
If you select multiple files and open them in `imv`, then after deleting one it will automatically move to the next.
I can replace `close` with `quit`, but then when browsing multiple files it will exit the program entirely.


https://github.com/user-attachments/assets/9d34be1e-603d-4f76-be6b-ae9b138cd655

dhh:
I mostly just open a single image at the time. Think it's too confusing that it's just left as a black screen. Understand that there's a tradeoff if you have opened multiple.

ctarx:
I’m happy to adjust this so it works better when opening just one image. We can either use quit instead, or add an extra keybind for “delete and quit.”
Something like this;

- `Ctrl+x`  delete + quit
- `Ctrl+Shift+x`  delete + close

dhh:
I like that 👍

ctarx:
Done 👍

PR #3316, opened 2025-11-10
Add a new theme named Ethereal

bjarneo:
By request, here is a new theme called Ethereal. All open for a new name.

What is not part of this PR is the vscode theme. If anyone has a suggestion for something fitting, that would be awesome. I do have a feeling that a custom theme has to be created.

Screenshot
screenshot-2025-11-10_21-18-41

### DIFF
--- a/config/nvim/lua/plugins/all-themes.lua
+++ b/config/nvim/lua/plugins/all-themes.lua
@@ -53,9 +53,4 @@ return {
 		lazy = true,
 		priority = 1000,
 	},
-	{
-    	"bjarneo/aether.nvim",
-    	lazy = true,
-    	priority = 1000,
- 	}
 }


dhh:
I think we need to convert the neovim setup to something that works with the current switcher. Do you know what's required @ryanrhughes?

bjarneo:
The important part here is that lazyvim is reloading the colors when they change. If I remember it correctly, the hot reload implemented in Omarchy just checks if the color scheme is changing, then reloads. 

The difference here is that aether.nvim hot reload checks if the file is changed, not the color scheme, as Aether always uses the same theme with different colors.

The alternative is to create a standalone nvim plugin, but then it adds more maintenance.

TLDR: It works with the current switcher as Aether is using this kind of neovim.lua setup. Could it be better? Most likely.

**After I wrote all of this, we can simplify by removing the custom hot reload parts and renaming the theme to ethereal. Then it would work with the Omarchy hot reloading.**

```
return {
    {
        "bjarneo/aether.nvim",
        name = "ethereal",
        priority = 1000,
        opts = {
            disable_italics = false,
            colors = {
                -- Monotone shades (base00-base07)
                base00 = "#060B1E", -- Default background
                -- more colors

                -- Accent colors (base08-base0F)
                base08 = "#ED5B5A", -- Variables, errors, red
                -- more colors
            },
        },
    {
        "LazyVim/LazyVim",
        opts = {
            colorscheme = "ethereal",
        },
    },
}
```

bjarneo:
Updated and verified!

The next question: should it be added to `config/nvim/lua/plugins/all-themes.lua` then?

zigmoo:
Very cool

bjarneo:
> Updated and verified!
> 
> The next question: should it be added to `config/nvim/lua/plugins/all-themes.lua` then?

also done!

ryrobes:
Nice. 

Not directly related, but I've been playing with the idea locally of themes having their own optional hypr.conf file - which means we could package more far-reaching changes in theme swaps - and the just change the hypr defaults to also load `~/.config/omarchy/current/theme/hypr.conf`. 

Just an idea!

bjarneo:
Vscode theme added!

https://marketplace.visualstudio.com/items?itemName=Bjarne.ethereal-omarchy

screenshot-2025-11-19_07-28-17

dhh:
There's a rebase needed.

PR #3318, opened 2025-11-10
Fix screenrecord not saving

pukljak:
I had problem with screenrecord.
It seems like it worked, even when i stopped i get notification that recording is stopped and saved, but it was empty.
Pasted image
image

I debug it, and this is log:
image

After my changes:
image

So it uses default command bat if that fails it goes to next which fixed my problem.
If I get it right this problem is connected to NVIDIA gpu.

### DIFF
--- a/bin/omarchy-cmd-screenrecord
+++ b/bin/omarchy-cmd-screenrecord
@@ -62,7 +62,10 @@ start_screenrecording() {
   # Merge audio tracks into one - separate tracks only play one at a time in most players
   [[ "$AUDIO" == "true" ]] && audio_args="-a default_output|default_input"
 
-  gpu-screen-recorder -w "$@" -f 60 -c mp4 -o "$filename" $audio_args &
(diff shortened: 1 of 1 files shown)


dhh:
@ryanrhughes Seems like we could build the fallback into the gpu-screen-recorder itself?

pukljak:
@dhh if needed I can give it a try

dec05eba:
@dhh gpu screen recorder supports that, see: https://github.com/basecamp/omarchy/pull/1830#issuecomment-3501712049
but in this case it's also an issue of omarchy not installing vaapi drivers, which it should do for a proper system setup, as mentioned here: https://github.com/basecamp/omarchy/issues/2706 fallback for cpu recording should only be done on the devices that truly dont support gpu encoding. Almost all computers support gpu encoding these days, even 14 year old laptops, if properly setup.

dec05eba:
I made a pull request that uses -fallback-cpu-encoding option instead: https://github.com/basecamp/omarchy/pull/3401 and fixes other issues

dhh:
Thanks for taking a swing at this, but I think fixing the root is probably best here. So let's focus on #3401.

PR #3319, opened 2025-11-10
Add turn off keyboard backlight

pukljak:
Laptop keyboard have backlight turned on all the time, even Omarchy is locked and screen is off.

It is set that keyboard backlight turn off when screen is off (after 5.5 minutes).
It turns back on when some key or touchpad is pressed.


pomartel:
I have the same issue! Thanks for the patch.

GreekAdonis:
I think it should fade in/out, and restore the previously set brightness value

PR #3323, opened 2025-11-10
Change wait command to use wait without delay

pgmdpfcon:


dhh:
That's not going to wait for elephant to start? Which is the point. Elephant needs to start before walker.

abenz1267:
This shouldn't be needed at all.

Why was this done? I don't remember... can you re-check if this is even still needed?

dhh:
Before this, we would occasionally get in a state where walker would be stuck with "waiting on elephant" or whatever the error message was. It was intermittent, but happened often enough. After the wait, I haven't seen it.

abenz1267:
Hm... ok. This is nothing anyone ever reported except for Omarchy.

I can't tell what it would be caused by, other than Hyprland exec-once/environment timings. It's def. not a Walker/Elephant problem.

The ideal way to start Elephant would still be a user-level systemd service. For Walker using xdg-autostart or a systemd service work fine.

I think fixing the root of the issue would be better than having to hack around it, like it's currently being done.

Walker even has an "emergency" mode where you can hardcode list items, for situations where Elephant is truly dead.

image

dhh:
Whatever produced the state, Walker/Elephant wouldn't recover on its own. Elephant would be running, Walker would be running, but Walker would be stuck in "Waiting for Elephant". Happy to explore other solutions, but this fix provided the relief we needed to be able to ship. Feel free to try another take at it in a PR ✌️

PR #3325, opened 2025-11-11
overload lateral window navigation for grouped windows

chrislewis:
Overloads `changegroupactive` so that it's also mapped by `SUPER ALT LEFT/RIGHT`


dhh:
I'm not sure I understand what this is supposed to do?

chrislewis:
It enables navigating through the windows in a single tabbed window group using SUPER ALT and the left or right arrow key. This is already possible using SUPER ALT, TAB and SUPER ALT SHIFT, TAB, but using the arrow keys feels more natural and consistent with using arrow keys to change the focus of actual windows.

dhh:
Ah, but we are using super alt arrows to move windows in/out of the group. Super alt pg up/down also moves between windows inside the group.

chrislewis:
Oh? They def [move windows into a group](https://github.com/basecamp/omarchy/blob/master/default/hypr/bindings/tiling-v2.conf#L82-L83) but the only current binding to remove from a group looks to be [SUPER ALT G](https://github.com/basecamp/omarchy/blob/master/default/hypr/bindings/tiling-v2.conf#L79). Confirmed on my laptop/current omarchy, but maybe you're speaking to a planned change?

Otherwise it works well; it is overloaded which feels right to me (with the same binding I can join a window to a group and then continue navigating tabs in that group), but to your point maybe that raises the question of the most reasonable/natural way to pull a window out. Personally I don't think of removing a window from a group as a "directional" action, so to me a dedicated non-arrow key makes sense.

tiling

dhh:
I see what you mean. Yeah, this is actually nice. Better than pg up/down. I'll remove those.

PR #3328, opened 2025-11-11
update flatpaks if available

neheb:
Integrate into omarchy-update-system-pkgs to avoid manually having to run flatpak update.


dhh:
We don't have any official support for flatpaks, so I don't think this is right to add in the default update, but we now have hooks, so you can do this yourself in `~/.config/omarchy/hooks/post-update` 🤘

PR #3330, opened 2025-11-11
improve: add image/video share and run share commands silently

DankyoDev:
This PR adds two new features and improves the UX of the share menu:

1. Added 'image' and 'video' modes in omarchy-cmd-share to quickly send
   the latest screenshot or .mp4 video from ~/Pictures or ~/Videos.

2. Updated show_share_menu in omarchy-menu to run share commands
   silently in the background using 'nohup', avoiding the terminal popup
   when triggered.

All changes include explanatory comments for maintainability.

No breaking changes; backward compatible with previous 'clipboard', 'file', and 'folder' modes.

### DIFF
--- a/bin/omarchy-menu
+++ b/bin/omarchy-menu
@@ -124,11 +124,11 @@ show_screenrecord_menu() {
 
 show_share_menu() {
   case $(menu "Share" "  Clipboard\n  File\n  Folder\n  Latest Picture\n  Latest Video") in
-  *Clipboard*) nohup omarchy-cmd-share clipboard >/dev/null 2>&1 & ;;
-  *File*) nohup omarchy-cmd-share file >/dev/null 2>&1 & ;;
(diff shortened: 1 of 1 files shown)


bin/omarchy-menu
```diff
@@ -122,11 +122,15 @@ show_screenrecord_menu() {
   esac
 }
 
+# Updated to run share commands silently in the background using 'nohup' to avoid opening a terminal window.
+# This makes the menu smoother and non-blocking when triggered from Hyprland or Wayland.
```
dhh: I don't see any terminal issue opening without this? Never did.

dhh:
Living with this for a while, I think it's not frequent enough that this is needed to warrant the additional complexity to the menu. Very much appreciate the work, but it's easy for us to add too much, and I think this is just over that line.

PR #3336, opened 2025-11-11
fix: disable mouse focus for JetBrains

landsman:
Let's turn off mouse focus at all. In JetBrains IDEs it produces unexpected UX.

Fixing issue with lost focus: https://github.com/basecamp/omarchy/pull/3326#issuecomment-3516611112

### DIFF
--- a/default/hypr/apps/jetbrains.conf
+++ b/default/hypr/apps/jetbrains.conf
@@ -19,4 +19,4 @@ windowrule = size >50% >50%, class:^(jetbrains-.*), title:^()$, floating:1
 windowrule = noinitialfocus, class:^(jetbrains-.*)$, title:^(win.*)$, floating:1
 
 # Disable mouse focus
-windowrulev2 = nofollowmouse,class:^jetbrains-.*$
+windowrule = nofollowmouse, class:^(jetbrains-.*)$


landsman:
Please test @NicolasDorier @coendewit @AdamMusa @jorgemanrubia

NicolasDorier:
The issue this fix is when a popup appear from some mouse action, the popup wouldn't be auto focused. (no issue with keyboard shortcut)

I tested this, and this work as expected. Let me check if we can't just use the jetbrain tag.

---

The `jetbrain` tag we apply only apply to a subset of popups. (those which had problems)

Some other popups  (like create new branch), don't have the issues we solved in the previous commit... However it has issues related to autofocus when we open via mouse click.

This properly fix it. ACK for me.

default/hypr/apps/jetbrains.conf
```diff
@@ -17,3 +17,6 @@ windowrule = size >50% >50%, class:^(jetbrains-.*), title:^()$, floating:1
 
 # Disable window flicker when autocomplete or tooltips appear
 windowrule = noinitialfocus, class:^(jetbrains-.*)$, title:^(win.*)$, floating:1
+
+# Disable mouse focus
+windowrulev2 = nofollowmouse,class:^jetbrains-.*$
```
NicolasDorier: use `windowrule` instead, and fix the formatting to be the same as above.

cyppe:
Not related, but a lot of smart people with Jetbrins knowledge. Is there any ideas how to get the jetbrains-toolbox to scale properly in 2x too? Super small right now.

NicolasDorier:
@cyppe, I made a video where I had a popup with scaling issue (see the Find File in the video) on https://github.com/basecamp/omarchy/pull/3326 

If that's your issue, you just need to modify hpr/app/jetbrains.conf as from my PR.

cyppe:
The settings you did works amazing for the editors. 

I means the jetbrains-toolbox floating window/widget. Not a big deal as its mainly used for installing jetbrains editors. But still maybe someone had some idea how to get the scaling correct.

image

Syrent:
It’s a solid improvement, but it looks like tab drag-and-drop no longer works as a side effect


https://github.com/user-attachments/assets/ae2ec2f1-2b3f-4b58-a921-5cb8351e7544

PR #3339, opened 2025-11-11
Update terminal binding to use --dir option

tonysm:
In basecamp@4c97a31, a migration was added to fix this option on existing installs, but I think the default config was never updated with the new option.


dhh:
cc @ryanrhughes so we might need another migration to fix what was generated with this.

PR #3342, opened 2025-11-11
Change qt5-remoteobjects to qt6-remoteobjects

Rovel:
I was trying to use omarchy-iso to build a custom image and was failing due to q5-remoteobjects not found, searching found that qt5-remoteobjects was removed so I upgraded to qt6, not sure if needs more editing but the ISO was built successfully.

https://archlinux.org/packages/extra/x86_64/qt5-remoteobjects/

### DIFF
--- a/install/omarchy-other.packages
+++ b/install/omarchy-other.packages
@@ -35,7 +35,7 @@ pipewire
 pipewire-alsa
 pipewire-jack
 pipewire-pulse
-qt5-remoteobjects
+qt6-remoteobjects
 qt6-wayland
 sassc
 snapper


AndrewMast:
It looks like the `qt5-remoteobjects` package was removed from the list in commit https://github.com/basecamp/omarchy/commit/fceb7106d90aaea4fb8c97095168dc0ffc66da23 on the dev branch (#3295).

Is `qt6-remoteobjects` needed?

Rovel:
Commenting it out broke build, but qt6-remote worked and I flashed a new device with that usb and it works, reading https://github.com/basecamp/omarchy/pull/1923 and lookig at `kvantum` arch repo: https://archlinux.org/packages/extra/x86_64/kvantum/ looks like its already in qt6 so I see no problem in the change

SytexMC:
> Commenting it out broke build

The dev branch seems to be building just fine. I'm honestly not sure what it is/was needed for in the first place. No point in keeping something that isn't needed.

dhh:
This has already been removed from dev. Whatever needed it must have been removed a while ago.

PR #3369, opened 2025-11-12
Add NVIDIA Container Toolkit for Docker GPU support

Zeus-Deus:
### Problem
Omarchy currently installs Docker and NVIDIA drivers automatically, but doesn't configure Docker to use the GPU. Users have to manually install and configure `nvidia-container-toolkit` after installation.

### Solution
This PR automatically installs and configures the NVIDIA Container Toolkit when an NVIDIA GPU is detected, enabling GPU-accelerated Docker containers out of the box.

### Changes
- Added `nvidia-container-toolkit` installation to `nvidia.sh`
- Configured Docker runtime to recognize NVIDIA GPU

### DIFF
--- a/install/config/hardware/nvidia.sh
+++ b/install/config/hardware/nvidia.sh
@@ -77,4 +77,12 @@ env = LIBVA_DRIVER_NAME,nvidia
 env = __GLX_VENDOR_LIBRARY_NAME,nvidia
 EOF
   fi
+  # Install NVIDIA Container Toolkit for Docker GPU support
+  sudo pacman -S --needed --noconfirm nvidia-container-toolkit
+
+  # Configure Docker to use NVIDIA runtime
+  sudo nvidia-ctk runtime configure --runtime=docker
+
+  # Restart Docker to apply changes
+  sudo systemctl restart docker
 fi


install/config/hardware/nvidia.sh
```diff
@@ -77,4 +77,12 @@ env = LIBVA_DRIVER_NAME,nvidia
 env = __GLX_VENDOR_LIBRARY_NAME,nvidia
 EOF
   fi
+  # Install NVIDIA Container Toolkit for Docker GPU support
```
jmargeta, line 80: Running these 3 commands worked perfectly to set up my NVIDIA-enabled machine for Docker, thanks @Zeus-Deus

As this is not a conditional install, why not move nvidia-container-toolkit to the PACKAGES_TO_INSTALL list and have it installed together with the other packages?

Zeus-Deus: It should only install on devices with Nvidia GPU's.
This script already detects if the user has them and installs if found, adding it here will install with the Nvidia drivers.

dhh:
I think this is relatively niche, so adding another dependency for it is a bit much. I'd rather document that this is what needs to happen. Maybe we start with an entry under Manual in Discussions? Thanks for exploring!

PR #3370, opened 2025-11-12
Feature: added config options for nightlight

whitallee:
User can edit nightlight.conf to add or change temperature of hyprsunset to cycle through their own preferences.
- Default toggle options still set to @dhh 's preferences, but easily adjustable.
- Edited `omarchy-toggle-nightlight` to access values from `.config/omarchy/nightlight.conf`.
Updated `omarchy-menu` to include access to config file.

image

image

I felt like this aligned with @dhh 's views on defaults being easily accessed to change. The current nightlight felt lacking for anyone that likes to use night mode with any sort of customization. Seems like it'd be a good addition for future versions.

AI Disclosure: I did use AI to help me through the BASH scripting, as I don't work with it much. The code in this PR has been thoroughly massaged and nitpicked by yours truly, let me know if I missed anything and need to make adjustments.

Complete honesty, this is my first time contributing to open source, so anyone with the time to, please let me know if I need to go about any of this differently. Hope this can help someone :+1:

### DIFF
--- a/bin/omarchy-menu
+++ b/bin/omarchy-menu
@@ -210,11 +210,12 @@ show_setup_power_menu() {
 }
 
 show_setup_config_menu() {
-  case $(menu "Setup" "  Hyprland\n  Hypridle\n  Hyprlock\n  Hyprsunset\n  Swayosd\n󰌧  Walker\n󰍜  Waybar\n󰞅  XCompose") in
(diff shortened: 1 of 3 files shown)


dhh:
This is an interesting idea, but we already expose the full nightlight config in `.config/hypr/hyprsunset.conf`. I think it's better to just point folks to editing that 👍

PR #3377, opened 2025-11-13
feat: Add battery status of peripheral devices to waybar

andnig:
I find it quite helpful to know the battery state of my connected peripherals. 
This PR adds such an indicator to the very right of the waybar.
It uses upower under the hood and scans every 30 seconds for changes.

It adds this peripherals icon and on tooltip you get the status of the individual devices
image

The top-level icon on the waybar itself is white, if all devices are in the good state, gets yellow, if one or many of the devices are below the good threshold and red, if one or many of the devices are in critical state. 
This should give a nice early warning, to charge the battery.

The migration itself should update the waybar config without overwriting custom configurations. While this is quite more complex than simply refreshing with the default waybar config, I think especially the waybar configuration might be customized quite frequently, so we might try to be careful here.

### DIFF
--- a/migrations/1762723022.sh
+++ b/migrations/1762723022.sh
@@ -1,9 +1,9 @@
 #!/bin/bash
 
-# This migration script adds the custom/omarchy-bluetooth-battery-status module to Waybar configuration.
+# This migration script adds the custom peripherals battery monitoring module to Waybar configuration.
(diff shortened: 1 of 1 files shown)


dhh:
That's an interesting idea, but I find it a bit much for the default pack. Would make a nice omarchy-accessories-battery or the like package that folks who need this can install 👍

PR #3389, opened 2025-11-13
improve: uniquely identify terminal apps for window rules and focus. Unify waybar behaviour to use launch-or-focus

patroza:
enables targeting terminal apps specifically for `windowrule` and `-launch-or-focus` similar to `Wiremix`.

### DIFF
--- a/bin/omarchy-cmd-screensaver
+++ b/bin/omarchy-cmd-screensaver
@@ -1,13 +1,13 @@
 #!/bin/bash
 
 screensaver_in_focus() {
-  hyprctl activewindow -j | jq -e '.class == "com.omarchy.Screensaver"' >/dev/null 2>&1
+  hyprctl activewindow -j | jq -e '.class == "org.omarchy.screensaver"' >/dev/null 2>&1
 }
 
 exit_screensaver() {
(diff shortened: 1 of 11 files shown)


dhh:
What purpose does this serve? We don't identify it when starting it via the default bindings either.

patroza:
> What purpose does this serve? We don't identify it when starting it via the default bindings either.

enables targeting btop app specifically for `windowrule` and `-launch-or-focus` similar to `Wiremix`.
from my pov, all terminal apps should actually be targetable individually.

The motiviation is e.g trying to launch-or-focus another terminal window will instead focus the running btop "app"

dhh:
Need to do this consistently, then, in all the places where btop is invoked. Including bindings. And need migrations to get it all sorted for everyone.

patroza:
> Need to do this consistently, then, in all the places where btop is invoked. Including bindings. And need migrations to get it all sorted for everyone.

Sure. It's added to bindings and waybar, and added migrations.
**update** also added for the remaining; lazydocker and terminal editors.

patroza:
I have a better idea, one sec

PR #3390, opened 2025-11-13
fix: passing additional arguments to omarchy-launch-webapp

patroza:
before the change
`omarchy-launch-or-focus-webapp "chrome-outlook.office.com__mail_inbox-Profile_2" "https://outlook.office.com/mail/inbox" "--profile-directory=Profile 2"`
would still launch with the default profile,

with the change, it loads with the specified Profile 2.

### DIFF
--- a/bin/omarchy-launch-webapp
+++ b/bin/omarchy-launch-webapp
@@ -7,4 +7,7 @@ google-chrome* | brave-browser* | microsoft-edge* | opera* | vivaldi* | helium*)
 *) browser="chromium.desktop" ;;
 esac
 
-exec setsid uwsm-app -- $(sed -n 's/^Exec=\([^ ]*\).*/\1/p' {~/.local,~/.nix-profile,/usr}/share/applications/$browser 2>/dev/null | head -1) --app="$1" "${@:2}"
+app=$1
+rest=${@:2}
+
+exec setsid uwsm-app -- $(sed -n 's/^Exec=\([^ ]*\).*/\1/p' {~/.local,~/.nix-profile,/usr}/share/applications/$browser 2>/dev/null | head -1) --app="$app" "$rest"


dhh:
How does adding these local assigns fix things?

patroza:
> How does adding these local assigns fix things?

i'll be honest; beats me, but with the change, the following works:
`omarchy-launch-or-focus-webapp "chrome-outlook.office.com__mail_inbox-Profile_2" "https://outlook.office.com/mail/inbox" "--profile-directory=Profile 2"`

without it, it doesn't.

for the record, the following works fine in both situations:
`omarchy-launch-webapp "https://outlook.office.com/mail/inbox" "--profile-directory=Profile 2"`

dhh:
Hmm. I can't replicate the problem.

When I do:

`omarchy-launch-or-focus-webapp HEY "https://app.hey.com" "--profile-directory=Profile 2"`

And then run it again, it correctly focuses on the HEY app. Maybe you window pattern is too strict? Remember it's done with fuzzy search.

patroza:
> Hmm. I can't replicate the problem.
> 
> When I do:
> 
> `omarchy-launch-or-focus-webapp HEY "https://app.hey.com" "--profile-directory=Profile 2"`
> 
> And then run it again, it correctly focuses on the HEY app. Maybe you window pattern is too strict? Remember it's done with fuzzy search.

The problem is not the focusing, that works fine, but the passing of the profile directory when passed through the `omarchy-launch-or-focus` script; without the fix, it's not passing on that setting correctly. 
So you have to test with Hey closed. I'm not sure if it remembers your last profile either so you may try some random profile directories, say Profile 10.

dhh:
Looks like the problem is the space in the profile name for you. If you do: `omarchy-launch-or-focus-webapp HEY https://app.hey.com --profile-directory=Profile2` it appears to be correct?

pomartel:
I stumbled on the same issue and ended up changing the name of the default `Profile 1` directory to `Work` in `~/.config/BraveSoftware/Brave-Browser`. I guess the space in `Profile 1` could also be escaped but I didn't try.

Issue #3388, opened 2025-11-13
Installation failed with wrong checksum


hamkoroke:
### System details

Thinkpad X1 Carbon Gen

### What's wrong?

During the early installation step the Intel firmware package fails during checksum checks.
Tried with and without internet wired connection, same result.

![Image](https://github.com/user-attachments/assets/f7ea1d9d-6879-41dc-bbd8-4955076b149f)

dhh:
We fixed this. Was a problem with the mirror.

PR #3392, opened 2025-11-13
fixed wrong checksum issue/#3388

himangshusingh:
fixed issue [#3388 ](https://github.com/basecamp/omarchy/issues/3388)


dhh:
This is doing a million other things than addressing the checksum issue, including adding mirror lists that are counterproductive to fast downloads. We need something far more targeted here.

PR #3397, opened 2025-11-13
Add interactive dashboard icon selector for webapp installer

Grepsy:
Adds integration with dashboardicons.com to allow users to search and select icons interactively when creating web apps. Users can now choose between searching the icon database or entering a custom URL.

Features:
- Interactive icon search with gum filter using icon names and aliases
- Automatic handling of light/dark icon variants
- Metadata caching (1 week) in ~/.cache/omarchy

https://github.com/user-attachments/assets/82074e61-594d-482a-8367-bdb55477927d

### DIFF
--- a/bin/omarchy-webapp-install
+++ b/bin/omarchy-webapp-install
@@ -1,10 +1,109 @@
 #!/bin/bash
 
+select_dashboard_icon() {
+  METADATA_URL="https://raw.githubusercontent.com/homarr-labs/dashboard-icons/refs/heads/main/metadata.json"
+  BASE_URL="https://cdn.jsdelivr.net/gh/homarr-labs/dashboard-icons"
+  CACHE_DIR="$HOME/.cache/omarchy"
+  CACHE_FILE="$CACHE_DIR/dashboard-icons-metadata.json"
+  CACHE_MAX_AGE=604800  # 1 week
+  FORMAT="png"
+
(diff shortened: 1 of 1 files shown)


dhh:
I think this is really clever, but I don't quite think it adds up to a better experience when you can't see the icon you're selecting. I think you're better off doing the search on dashboardicons.com directly, so you can see what you're setting.

What I would consider, though, is a way of basically finding the correct icon from dashboardicons if the URL you've entered is a match. Not sure if that's possible but if you find a way please do open another PR with that 👍

Grepsy:
Thanks for the feedback! Appreciate it. I see that ghostty supports image rendering through the Kitty graphics protocol so might be able to use that for a live preview. I'll give it a shot and report back.

ghostty is new default terminal right? Does that mean we can assume it will always be used when running omarchy scripts?

dhh:
Ghostty will be the new default for new installations from 3.2. But we still support Alacritty. So we'd have to have a fallback that works okay with that as well. But I think if we find a way to do URL matching, we maybe don't even need it?

Grepsy:
I'm not sure URL matching will work because the dashboard-icons metadata doesn't include URL info. I did manage to get icon previews working with Ghostty and will check if I can find a solution for Alacritty.

https://github.com/user-attachments/assets/406cb559-2eae-48eb-9d9c-915199c01fc5

Issue #3303, opened 2025-11-10
screenrecording not showing up in /Videos


jordihoven:
### System details

Dell XPS 15 9350, Omarchy 3.1.7

### What's wrong?

Recording a video using Trigger/Capture/Screenrecord/Region no file is saved to my /Videos folder.

wqsaali:
After upgrade, I am facing following error on screenrecording
When i click on `Region` recording
<img width="327" height="379" alt="Image" src="https://github.com/user-attachments/assets/68b41d5b-0527-4249-9295-6d19fca6b600" />

It stored mp4 file in videos but recording video is corrupted.

https://github.com/user-attachments/assets/9833a558-0dce-4444-a7d5-9475f0b31e62

ElBrodino:
Check out this #2913

nemoralis:
https://github.com/basecamp/omarchy/issues/2628#issuecomment-3424009321

dhh:
Should be fixed now in 3.2.

jordihoven:
> Should be fixed now in 3.2.

Confirmed. Screenrecordings now show up under /Videos as expected.

PR #3401, opened 2025-11-14
Redesign screen capture option

### LINKED ISSUE #3367: Screen recording not working / skewed for external displays
### System details

Latitude 7390, Intel(R) Core(TM) i7-8650U (8) @ 4.20 GHz, Intel UHD Graphics 620 @ 1.15 GHz [Integrated]

### What's wrong?

Here is my system detail

Image

I have three external displays and one laptop display. Screen recording is behaving differently for these displays
### Laptop own display
It is working fine
#### Logs
Image

### External display 1
It is not able to find display
#### Logs
Image

### External display 2
It skewed the display when recording
#### Logs
Image

#### Recording
https://github.com/user-attachments/assets/3eb34f1d-59f8-4e07-a6a4-31239fc57dae

### External display 3
It is not able to find display
Image

I am using wdisplays (I used nwg-displays as well) but same issue

Image

dec05eba:
Use desktop portal capture option which allows capturing monitors connected to external gpus (common on laptops with dedicated gpus) and supports capturing properly when hdr is enabled.

Desktop portal capture allows selecting capture of a window, monitor or screen region.

Add -fallback-cpu-encoding yes to use cpu encoding if the system doesn't support gpu encoding. Ideally omarchy should install vaapi drivers which would fix this issue for most users, but some users have a gpu that dont support gpu encoding at all so this option is still needed.

Use pgrep with ^ as is done in waybar indicator screen-recording.sh to only grep for gpu-screen-recorder program.

Update omarchy-menu to reflect not selecting region/display capture and instead allow capturing only desktop audio or desktop + microphone audio.

Stop recording when pressing the record hotkey or navigating to the record menu instead of having to select
a recording option and then stopping the recording.

Fixes #3367, #3303 and #3118

### DIFF
--- a/bin/omarchy-cmd-screenrecord
+++ b/bin/omarchy-cmd-screenrecord
@@ -11,12 +11,14 @@ fi
 DESKTOP_AUDIO="false"
 MICROPHONE_AUDIO="false"
 WEBCAM="false"
+STOP_RECORDING="false"
 
 for arg in "$@"; do
   case "$arg" in
     --with-desktop-audio) DESKTOP_AUDIO="true" ;;
     --with-microphone-audio) MICROPHONE_AUDIO="true" ;;
     --with-webcam) WEBCAM="true" ;;
+    --stop-recording) STOP_RECORDING="true"
   esac
 done
 
@@ -109,8 +111,10 @@ if screenrecording_active; then
   else
     stop_screenrecording
   fi
-else
+elif [[ "$STOP_RECORDING" == "false" ]]; then
   [[ "$WEBCAM" == "true" ]] && start_webcam_overlay
 
   start_screenrecording || cleanup_webcam
+else
+  exit 1
 fi
--- a/bin/omarchy-menu
+++ b/bin/omarchy-menu
@@ -112,10 +112,12 @@ show_screenshot_menu() {
 }
 
 show_screenrecord_menu() {
+  omarchy-cmd-screenrecord --stop-recording && exit 0
+
   case $(menu "Screenrecord" "  With desktop audio\n  With desktop + microphone audio\n  With desktop + microphone audio + webcam") in
-  *"With desktop audio"*) omarchy-cmd-screenrecord --with-desktop-audio ;;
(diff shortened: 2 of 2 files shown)


dec05eba:
Note that I tested recording a window as well which works fine, as long as the monitor isn't rotated. This seems to be an issue in hyprland as obs studio also has the same issue and it works in other wayland compositors.

dec05eba:
@ryanrhughes

sohro-da:
@dec05eba, thank you for your screen capturing improvement efforts. I believe this feature is really handy and useful in any OS. Especially, when I need to quickly show my friends/coworkers how to do something and how exactly.

But do you think you can also do a small patch, so when in capturing mode, pressing the keys again would stop and save the video?

Because I can't believe, as of now, you need to press the keys, then select an item to stop the capturing. If it's already fixed in the upcoming version, please ignore this message.

dec05eba:
@sohro-da I added that now

neheb:
Lovely. That CPU fallback is needed. My GPU (7750) won't get hardware acceleration until kernel 6.19.

ryanrhughes:
I don't see any inherent downfall with this approach and it does resolve a few issues with things like HDR and skewing.

@dhh The only downside I see here is that we have to use the screen share portal selector which isn't the greatest experience. Though perhaps it's also motivation to explore better alternatives there as well such as https://github.com/WhySoBad/hyprland-preview-share-picker to enhance the screen sharing experience everywhere.

dhh:
Merged in https://github.com/basecamp/omarchy/commit/128a6125a92d

neheb:
@dec05eba software fallback is still broken.

dec05eba:
> @dec05eba software fallback is still broken.

Are you sure you are using the updated version? In that case run the `bin/omarchy-cmd-screenrecord` script and report the issue as mentioned here: https://git.dec05eba.com/?p=about with the terminal output

dec05eba:
There was an issue with proper fallback to cpu encoding. Can you try if version r1213.72c548c of gpu-screen-recorder (on aur) works for you now? if it does then this issue will get resolved when omarchy updates to that version of gpu-screen-recorder.

neheb:
Yep that fixed it.

PR #3404, opened 2025-11-14
hypr: Adding xdph token default of true

larrydewey:
The `allow_token_by_default` configuration option under `screencopy` within `xdg-dektop-portal-hyprland` (xdph) enables the "Allow restore token" checkbox by default in the screen sharing section dialog. Enabling it to `true` ensures the checkbox is pre-ticked, which can help avoid repeated selection prompts during screen sharing sessions, such as those reported with Slack, Google Meet, Zoom, etc. on Hyprland.

EDIT:
In order to see this work, I had to restart `xdph` via `systemctl --user restart xdg-desktop-portal-hyprland`

### DIFF
--- a/migrations/1763478660.sh
+++ b/migrations/1763478660.sh
@@ -1,0 +1,4 @@
+echo "Configure XDPH config for screensharing to remember token selection"
+
+cp $OMARCHY_PATH/config/hypr/xdph.conf ~/.config/hypr/
+systemctl --user restart xdg-desktop-portal-hyprland


dhh:
Can you record a video showing the difference? I couldn't see any difference trying this using the Zoom web app.

larrydewey:
Google Meet Videos:

https://github.com/user-attachments/assets/3ccc641b-65fb-4aac-8f16-33f37151170d

https://github.com/user-attachments/assets/e82498a4-0c80-4dbd-af69-3ec6f2ef5518

larrydewey:
Zoom Meet Videos:

https://github.com/user-attachments/assets/f7e056ce-6d9d-4ff9-b046-c45acfa7b64e


https://github.com/user-attachments/assets/36ed5e4b-3dfe-4027-8809-b2dc21176d37

larrydewey:
@dhh I rebooted after initially making the change because of a kernel update, but like I mentioned in the edit, if you restart xdph, you can see the token remember box is checked and it doesn't ask twice.

dhh:
Ahh, I didn't restart xdph, so that's probably why. Will try with that.

PR #3406, opened 2025-11-14
fix: add walker action hints

abenz1267:
I think Omarchy should expose the possible keybind actions available to users, as default Walker already does.

Hiding this information from users doesn't really serve any purpose and in fact: people don't even know what their Walker can do.

I just added the according xml section... of course styling can be adjusted. I think @ryanrhughes worked on this at some point.

I think you should really get this in.

For reference: i'm talking about these

image

### DIFF
--- a/default/walker/themes/omarchy-default/layout.xml
+++ b/default/walker/themes/omarchy-default/layout.xml
@@ -113,6 +113,35 @@
               </object>
             </child>
             <child>
+              <object class="GtkBox" id="KeybindsWrapper">
+                <property name="hexpand">true</property>
+                <style>
+                  <class name="keybinds-wrapper"></class>
+                </style>
+                <child>
+                  <object class="GtkBox" id="GlobalKeybinds">
+                    <property name="hexpand">true</property>
+                    <property name="spacing">10</property>
+                    <property name="margin-top">10</property>
+                    <style>
+                      <class name="global-keybinds"></class>
+                    </style>
+                  </object>
+                </child>
+                <child>
+                  <object class="GtkBox" id="Keybinds">
+                    <property name="hexpand">true</property>
+                    <property name="halign">end</property>
+                    <property name="spacing">10</property>
+                    <property name="margin-top">10</property>
+                    <style>
+                      <class name="keybinds"></class>
+                    </style>
+                  </object>
+                </child>
+              </object>
+            </child>
+            <child>
               <object class="GtkLabel" id="Error">
                 <style>
                   <class name="error"></class>


dhh:
Appreciate the suggestion, but aesthetically, I think that looks too rough. Wouldn't want to look at that all the time when I'm looking at the launcher. Also think the majority of those options are pretty rarely used, so doesn't feel worth the real estate to expose it all the time imo. I've never used any of these (beyond return). We don't use history or pinning either.

Would be nice to document that this is possible for folks to add to their own install somewhere though 👍

abenz1267:
So if I add:
- a config option to generally hide/show them
- a possible keybind to toggle the actions

You would add it?

ryanrhughes:
@abenz1267 Is it viable to have something like `?` hotkeyed to reveal them? That's a pretty common one across other things we have so it'd be consistent. 

While I don't know that we always need them, I do know there are a number of options available that I didn't even know, particularly with more commonly used elements like the calculator history.

abenz1267:
> @abenz1267 Is it viable to have something like `?` hotkeyed to reveal them? That's a pretty common one across other things we have so it'd be consistent.

I don't see how "consistency" is a factor? Omarchy is using 10000 different tools, it's not like you use Quickshell and have 1 UI for everything. You should try to provide a good experience for the tool in question.

> While I don't know that we always need them, I do know there are a number of options available that I didn't even know, particularly with more commonly used elements like the calculator history.

That is correct. And some of them are contextual, f.e. the Clipboard pause/unpause. So without always showing them, you don't even know if your clipboard is paused or not.

I've thought about this and i will add a simple config option to hide/show them, so one doesnt have to create a whole new theme to achieve that, but aside from that i won't add any toggle. It simply doesn't make sense.

For a quick "reality check":

- you actually don't even notice them THAT much when actually using
- since introducing them, there wasn't really any negative feedback
- functionality and discoverability shouldn't get crippled by aesthetics

If @dhh  doesn't use any of the 10000 features Walker has to offer: fine. But hiding that possibilities from Omarchy users... ok? Questionable.

I'm not even sure why you wouldn't just go for a simpler launcher like Fuzzel at this point. You don't have to deal with the complexities of Walker+Elephant. It's already minimal by default. I guess @markstos would gladly help you migrate.

- simple
- no CSS/XML
- you don't need the theming capabilities anyways
- you don't use any advanced features of Walker
- you seem to have no intention of providing the "best" Walker experience to your users (thats based on the feedback of people that use Walker because they actually want to use it)

Obviously @dhh has no clue what Walker does or has to offer, nor does he seem to be interested in even finding out. I'm pretty sure he hasn't read any readme even once. Nor looked at the configuration besides PRs. That's ok.

But whats the point? If you don't actually vet the software you ship to your users... how opinionated is it really? Can you even have an opinion on something you barely know anything about?

Walker is a complex tool with a lot of options. I'm not sure hiding crucial things like functionality is a good thing. Esp. with "aesthetics" being the reason. You aren't Apple and this isn't MacOSX.

And to make this clear: i've opened up this PR for a single reason => every time I show someone that he can enable the hints the feedback is the same: "why doesn't omarchy show them?!" Every. Single. Time.

Do what you want :shrug:

abenz1267:
I've actually looked at what providers Omarchy ships to it's users at the moment just to point out how ridiculous this actually is:

https://github.com/omacom-io/omarchy-pkgs/blob/master/pkgbuilds/omarchy-walker/PKGBUILD

- 'elephant-bluetooth' => unusable without knowing the keybinds
  - finding new devices
  - connect/disconnect
  - trust/untrust
  - pair
- 'elephant-calc' => usable, but you will not know about saving
  - save/delete results 
- 'elephant-clipboard' => hiding functionalities, some even crucial like clearing or pause/unpause
  - clearing
  - deleting
  - editing
  - send via localsend (not released yet...)
  - pause/unpause
- 'elephant-desktopapplications' => this depends on other configuration options, but again hiding functionality
  - even without any config => open app without closing Walker
- 'elephant-files' => hiding basically all functionality
  - open
  - open dir
  - copy path
  - copy file
  - send via localsend (not released yet...)
- 'elephant-runner' => hiding functionality
  - run in terminal
- 'elephant-todo' => unusable without knowing the keybinds
  - change category
  - create
  - delete
  - change status / start tracking
- 'elephant-websearch' => depends on configuration
  - engines can be configured as keybinds for faster access

This would also basically apply to providers you aren't shipping.

.... does this sound right? :shrug:

koibtw:
@abenz1267 bro ur crashing out its not that deep 😭

abenz1267:
> @abenz1267 bro ur crashing out its not that deep 😭

It's not my fault you had to nuke https://github.com/hyprutils/ (and hyprlauncher :joy: ) because nothing took off and everyone on hyprland was making fun of you and nyx. There's no need to stalk me publicly. You can't be THAT hurt.

koibtw:
bruh
ur mad lol
andrej

ryanrhughes:
@abenz1267 my question was a simple suggestion in response to yours. In my opinion, hotkeys are largely optional, should be intuitive enough to not need direction, and have a way to reference them as a part of a solid user experience. If you have to constantly put them in front of the user; that's a UX problem that should be explored.

That aside; there is an appropriate way to present a case for something where there's a difference of opinion but a page log insulting rant definitely isn't it. I'm really disappointed that's the direction this thread went. 

At this point, I suppose we'll just agree to disagree, and move on.

abenz1267:
> @abenz1267 my question was a simple suggestion in response to yours. In my opinion, hotkeys are largely optional, should be intuitive enough to not need direction, and have a way to reference them as a part of a solid user experience. If you have to constantly put them in front of the user; that's a UX problem that should be explored.

You have a misconception of what hotkeys are. There is no way of accessing the functionality otherwise. Those keys are the ONLY way for using the actions. They are not hotkeys. They ARE the functionality.

Keybinds/Hotkeys are fundamentally not intuitive. When talking UX there's different "levels" of how to use a program: beginner, intermediate and expert (simplified). Hotkeys fall under intermediate- and expert-level usage. They usually get adopted for more efficiency and are fundamentally NOT intuitive.

We don't talk about "enter" here.

If you don't know about functionality, keybinds are even LESS intuitive. Walker is HEAVILY keyboard driven.

The learning of hotkeys usually goes: my program can do X, is there a keybind to do X? Example: open a new tab in a browser => how? => ctrl t (firefox).

image

Ofc firefox has the luxury of being able to expose functionality via a mouse-accessible context menu.

The concept is the same: expose functionality => provide keybind. That's what Walker does. Hiding them is horrible UX Design.

I am somewhat UX Engineer btw.

> That aside; there is an appropriate way to present a case for something where there's a difference of opinion but a page log insulting rant definitely isn't it. I'm really disappointed that's the direction this thread went.

What insults? That I said DHH has no clue about Walker? That is just a fact, proven on multiple occasions. What else? I'm sorry, but if you feel offended by this ..... that's up to you. It wasn't meant to be offending at all. I was simply pointing out that as a vendor you ship a product you barely know anything about, nor put in the time to learn about it. That is your decision and if you feel offended by that, maybe you should address it with yourself? 

It's like someone saying: "Benz, you are 1.76m". If i get offended by that.. that's a me problem.

> At this point, I suppose we'll just agree to disagree, and move on.

You haven't addressed anything i said. :shrug: 

What are you disagreeing with?
- shipping modules DHH doesnt use => why isn't this opinionated?
- shipping modules that are fundamentally unusable without the hints => bad UX
- hiding functionality from users => bad UX

saying `we'll just agree to disagree` is kinda pointless. i don't even know what you disagree with.

PR #3410, opened 2025-11-15
Enhance window focus to seamlessly cycle between tiled and floating windows

DarrenVictoriano:
This PR introduces a new Hyprland focus helper script omarchy-hyprland-move-focus that improves window navigation by:
- Moving focus between tiled windows using directional keys (l, r, u, d).
- Seamlessly falling back to floating windows when no tiled window exists in the chosen direction.
- Cycling between all windows when the currently focused window is floating.

## Caveat:
- **no_focus_fallback** must be set to true in the Hyprland configuration for this logic to function as intended. This ensures that failing to focus a tiled window properly triggers the fallback path.
- floating window is always visually on top even if it is not the active focused window. I can't seem to find a way to override this behavior, maybe its hyperland's limitation?

## Demo

https://github.com/user-attachments/assets/15316d70-46c5-441f-9dce-6c50a1021dc0

### DIFF
--- a/bin/omarchy-hyprland-move-focus
+++ b/bin/omarchy-hyprland-move-focus
@@ -1,0 +1,19 @@
+#!/bin/sh
+
+DIRECTION="$1"
+
+ACTIVE_WINDOW_BEFORE=$(hyprctl activewindow -j | jq .address)
+IS_FLOATING=$(hyprctl activewindow -j | jq .floating)
+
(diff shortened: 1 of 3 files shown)


dhh:
This is an interesting idea, but we already have alt+tab to jump between windows in layers. I don't love the idea of rewriting basic Hyprland functionality like this with a script. Might be something to document under Discussions > Manual for those who want that though ✌️

PR #3411, opened 2025-11-15
feat(walker): set window_integration true

abenz1267:
`window_integration` makes the `desktopapplications` provider behave like this:

- if an application is already opened => focus window instead of opening a new instance

`window_integration_ignore_actions` will cause elephant to ignore this behaviour for desktop actions, f.e. `new private window` => actions always open a new instance.

I think this behaviour is more in line with "vanilla" omarchy, because it uses `launch-or-focus` everywhere in the keybinds to start/focus apps.

### DIFF
--- a/config/elephant/desktopapplications.toml
+++ b/config/elephant/desktopapplications.toml
@@ -1,3 +1,5 @@
 show_actions = false
 only_search_title = true
 history = false
+window_integration = true
+window_integration_ignore_actions = true


dhh:
I think it's going to be hard to do this consistently. Some apps you want as singletons, but others, like a terminal, you don't. That's why the default bindings in Omarchy are split about 50-50 between singletons which we focus and the rest which we don't. I don't know how many people are using their app launcher to focus on existing applications, but I could see us having this as a commented out but explained option in the default config to accommodate that. As a default, though, I think it's better to launch rather than focus.

abenz1267:
fair

PR #3415, opened 2025-11-15
Support opening web apps with a specified browser

rogeriopvl:
## Overview
This PR adds support for opening web apps with a specified browser.

This is useful in scenarios where instead of using different browser profiles for work and personal browsing (the most common approach), users use different browsers (in my case Brave for personal browsing, with strict privacy settings, and Chromium for work).

## Example
I want to open the Youtube web app with Brave instead of my default (work) browser.

With this change I can edit the desktop file at `~/.local/share/omarchy/applications/Youtube.desktop` and change the `Exec` param to specify a different browser:

```shell
[Desktop Entry]
Version=1.0
Name=YouTube
Comment=YouTube
Exec=omarchy-launch-webapp https://youtube.com/ brave-browser.desktop
Terminal=false
Type=Application
Icon=/home/rogeriopvl/.local/share/applications/icons/YouTube.png
StartupNotify=true
```

I understand this may be considered "edge-case enough" to be included in Omarchy and that's ok. I just wanted to share it in case more people are interested in this feature.

### DIFF
--- a/bin/omarchy-launch-webapp
+++ b/bin/omarchy-launch-webapp
@@ -1,10 +1,18 @@
 #!/bin/bash
 
-browser=$(xdg-settings get default-web-browser)
+# use browser passed as arg or the user default one
+browser="${2:-$(xdg-settings get default-web-browser)}"
 
 case $browser in
 google-chrome* | brave-browser* | microsoft-edge* | opera* | vivaldi* | helium-browser*) ;;
 *) browser="chromium.desktop" ;;
 esac
 
-exec setsid uwsm-app -- $(sed -n 's/^Exec=\([^ ]*\).*/\1/p' {~/.local,~/.nix-profile,/usr}/share/applications/$browser 2>/dev/null | head -1) --app="$1" "${@:2}"
+# if second arg was used as browser, use remaining args for flags
+if [[ $# -ge 2 && "$2" == "$browser" ]]; then
+  flags=("${@:3}")
+else
+  flags=("${@:2}")
+fi
+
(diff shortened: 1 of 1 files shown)


gorciu-official:
upvote

abenz1267:
Elephant has `elephant-bookmarks` which lets you define multiple browser you can assign to bookmarks and change them around as you want. Maybe that's something of interest for you.

rogeriopvl:
@abenz1267 I'm interested, but I'm not sure I understand how that would work. Where can I find an example or documentation about `elephant-bookmarks`?

abenz1267:
Eehh.. that's another victim of omarchy not showing the actual keybind hints in Walker.

image

If you've installed `elephant-bookmarks` and you access it via a prefix or the switcher, you can use all the actions provided at the bottom.

you can get documentation for the config by executing `elephant generatedoc bookmarks`. It will print out markdown.

Kyhze:
As a related additional feat, it would be great to have the custom web app wizard ask you which browser you want to use for that specific web app.

rogeriopvl:
> As a related additional feat, it would be great to have the custom web app wizard ask you which browser you want to use for that specific web app.

If there's interest in merging this, I can definitely add that as well.

dhh:
The `omarchy-launch-webapp` command is there to provide the indirection of browsers being selected. If you know you want a specific browser, you can and should probably just invoke that directly. Change your Exec line to: `Exec=brave --app=https://youtube.com/` 👍

PR #3428, opened 2025-11-16
Avoid invalid terminal working directory

### LINKED ISSUE #3427: Opening a new terminal with Super+Enter sometimes sets working directory to `/usr/bin` instead of $HOME
### System details

Thinkpad L380, Omarchy 3.1.7

### What's wrong?

If the current active window has a child process, then Super+Enter sets the working directory of the newly opened terminal to the cwd of the child process. This is great when the current active window is a terminal, and the child process is a shell. This is less great when the current active window is Chromium or Firefox, and the child process is some random web process, with cwd `/usr/bin`.

## Why

Looking at `~/.config/hypr/bindings.conf`, we see the following:

``` conf
bindd = SUPER, RETURN, Terminal, exec, $terminal --dir="$(omarchy-cmd-terminal-
cwd)"
```

Of particular interest, is the command `omarchy-cmd-terminal-cwd`.

We can reproduce the underlying issue by running the following, and ensuring that the active window is Chromium before the sleep finishes:

``` bash
sleep 3 && omarchy-cmd-terminal-cwd
```

If you managed to focus Chromium before the sleep finished, you will see the output `/usr/bin`.

Looking at the source for `omarchy-cmd-terminal-cwd`, it is immediately clear why this happens:

https://github.com/basecamp/omarchy/blob/8350b84eee10601efce64efef6fe044f03bfcd32/bin/omarchy-cmd-terminal-cwd#L1-L17

You can see that it assumes that the current active window is a _terminal_, and that the last child PID belongs to a shell spawned by that terminal.

But this is not a valid assumption, if the current active window is _not_ a terminal, and the last child PID of that window is _not_ a shell.

## Proposed fix

``` diff
diff --git a/bin/omarchy-cmd-terminal-cwd b/bin/omarchy-cmd-terminal-cwd
index 062c35f..b651148 100755
--- a/bin/omarchy-cmd-terminal-cwd
+++ b/bin/omarchy-cmd-terminal-cwd
@@ -6,8 +6,10 @@ shell_pid=$(pgrep -P "$terminal_pid" | tail -n1)

 if [[ -n $shell_pid ]]; then
   cwd=$(readlink -f "/proc/$shell_pid/cwd" 2>/dev/null)
+  shell=$(readlink -f "/proc/$shell_pid/exe" 2>/dev/null)

-  if [[ -d $cwd ]]; then
+  # Check if $shell is a valid shell and $cwd is a directory.
+  if grep -qs "$shell" /etc/shells && [[ -d $cwd ]]; then
     echo "$cwd"
   else
     echo "$HOME"
```

I'll link a PR with this fix.

daviewales:
Don't use the working directory of the last child of the active window when opening a new terminal with Super+Enter.

Fixes #3427.

### DIFF
--- a/bin/omarchy-cmd-terminal-cwd
+++ b/bin/omarchy-cmd-terminal-cwd
@@ -6,8 +6,10 @@ shell_pid=$(pgrep -P "$terminal_pid" | tail -n1)
 
 if [[ -n $shell_pid ]]; then
   cwd=$(readlink -f "/proc/$shell_pid/cwd" 2>/dev/null)
+  shell=$(readlink -f "/proc/$shell_pid/exe" 2>/dev/null)
 
-  if [[ -d $cwd ]]; then
+  # Check if $shell is a valid shell and $cwd is a directory.
+  if grep -qs "$shell" /etc/shells && [[ -d $cwd ]]; then
     echo "$cwd"
   else
     echo "$HOME"


daviewales:
Note: This PR does make `omarchy-cmd-terminal-cwd` slightly slower, as it runs two additional commands:

**master**:

``` hyperfine
hyperfine omarchy-cmd-terminal-cwd
Benchmark 1: omarchy-cmd-terminal-cwd
  Time (mean ± σ):      86.8 ms ±  15.4 ms    [User: 51.3 ms, System: 50.3 ms]
  Range (min … max):    59.9 ms … 127.7 ms    30 runs
```

**this PR**:

``` hyperfine
hyperfine omarchy-cmd-terminal-cwd
Benchmark 1: omarchy-cmd-terminal-cwd
  Time (mean ± σ):      98.9 ms ±  18.9 ms    [User: 58.1 ms, System: 56.3 ms]
  Range (min … max):    76.8 ms … 148.6 ms    19 runs
```

abenz1267:
I'm not fully sure this script is needed at all, now that Ghostty is the default terminal.

daviewales:
I'm assuming that the other terminals are still supported? So maybe only necessary for them?

PR #3432, opened 2025-11-16
Make gaps toggle also turn off rounded corners

wusher:
When you press the gaps toggle key, it now also turns rounded corners on and off. The script remembers your corner setting so it can put it back the way you had it.


wusher:
accidentally opened

PR #3449, opened 2025-11-18
Allow experimental arm64/aarch64 installs in preflight guard

Jeevan00-1:


dhh:
We don't have a fully verified setup for arm yet. So don't want to remove the guard until we do. You can always continue anyway.

PR #3451, opened 2025-11-18
fix(walker): update for breaking changes in latest version

abenz1267:
These changes are needed to avoid config breakage.

### DIFF
--- a/config/walker/config.toml
+++ b/config/walker/config.toml
@@ -1,7 +1,8 @@
-force_keyboard_focus = true # forces keyboard forcus to stay in Walker
-selection_wrap = true       # wrap list if at bottom or top
-theme = "omarchy-default"   # theme to use
+force_keyboard_focus = true                                                 # forces keyboard forcus to stay in Walker
+selection_wrap = true                                                       # wrap list if at bottom or top
+theme = "omarchy-default"                                                   # theme to use
 additional_theme_location = "~/.local/share/omarchy/default/walker/themes/"
+hide_action_hints = true                                                    # globally hide the action hints
 
 [placeholders]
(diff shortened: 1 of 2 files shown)


dhh:
What is the config breakage not having this will cause? Are we in need of an emergency release to deal with it?

abenz1267:
Uhm... i'm not sure, since you host the packagebuilds yourself? If the version is somehow pinned, no issues. In worst case it will fallback to default theme.

I wasn't sure when you want to release 3.2 so i just created the PR as early as possible.

dhh:
@ryanrhughes told me we're good. We've paused building our own pkgs so we don't pick up this incompatible version before we're ready to roll out 3.2.

philweber:
> What is the config breakage not having this will cause? Are we in need of an emergency release to deal with it?

Existing users who upgrade to the latest version of `walker` will see an error message at the bottom of menus and the launcher: “Required object ‘Keybinding’ not found”. Adding the new required objects to `layout.xml` will see unsightly shortcut key hints at the bottom of `walker` windows. I tried (unsuccessfully) to edit the omarchy-default theme to hide the shortcut hints, but ultimately rolled back to the previous version of `walker`.

abenz1267:
If you install the latest version from the AUR and not the one Omarchy ships, yeah, that will happen. This is what this PR is for.

You'd simply look at the commit (... or release notes...) to see how to hide them.

dhh:
Are these changes backwards compatible? So if we merge this now, before we have the latest package, will that also give an error message or just be ignored?

abenz1267:
Not backwards compatible due to changed element IDs.

dhh:
This still seems busted?

image

image

dhh:
Needed a migration to refresh the walker config and restart walker. All good.

PR #3461, opened 2025-11-19
first-run: Use balanced profile for desktop as well

defer:
My cpu (9550x3d) idles at 30w with the balanced governor, but 45w when in performance.

This patch changes the default to balanced for both because desktop cpus also benefit from scaling in the form of power savings (i.e. in my case this could be $15/year assuming a 80% idle scenario), lower part degradation, lower heat generation and fan noise.

The CPU will still reach higher clocks when under load, but we still reap the benefits when on idle.

### DIFF
--- a/install/first-run/battery-monitor.sh
+++ b/install/first-run/battery-monitor.sh
@@ -1,10 +1,8 @@
+# Set balanced profile to allow power savings when idle
+powerprofilesctl set balanced || true
+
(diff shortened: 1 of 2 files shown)


dhh:
Appreciate the suggestion, but this is not how I'd like to have my computers set up out of the box. There's a clear advantage to running in performance, and I'm personally fine with the higher wattage being drawn as a trade-off. But we have the selector exactly for this so if you differ, you can change it 👍

PR #3462, opened 2025-11-19
Add Google DNS option

Jamesking56:
Inspired by Cloudflare's latest major downtime incident, it would be nice if we had an option to quickly switch to using Google's Public DNS servers.

This PR adds Google as a DNS option to the Setup > DNS so that if Cloudflare goes down again, Omarchy users can quickly switch to Google's Public DNS (which does not use any of Cloudflare's infrastructure).

### DIFF
--- a/bin/omarchy-setup-dns
+++ b/bin/omarchy-setup-dns
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 if [[ -z $1 ]]; then
-  dns=$(gum choose --height 5 --header "Select DNS provider" Cloudflare DHCP Custom)
+  dns=$(gum choose --height 5 --header "Select DNS provider" Cloudflare Google DHCP Custom)
 else
   dns=$1
 fi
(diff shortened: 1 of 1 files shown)


Jamesking56:
This file has a lot of code duplication, I'm happy to reduce the code duplication either in this PR or another PR if you like?

dhh:
Let's clean up the code duplication before adding more to it 👍

Jamesking56:
> Let's clean up the code duplication before adding more to it 👍

#3495 

Will rebase this once #3495 is in

Niknu:
> > Let's clean up the code duplication before adding more to it 👍
> 
> #3495
> 
> Will rebase this once #3495 is in

Would it be difficult to understand the chain of changes if you rebase this already on the #3495 ? 
From my experience you shouldn't experience any git problems from that.

dhh:
This has been added.

PR #3464, opened 2025-11-19
Add  theme ethereal

bjarneo:
The previous PR, not able to reopen after I fixed the rebase: https://github.com/basecamp/omarchy/pull/3316

516101443-157aaa6b-9890-4581-b33b-2ed0fbbfb527
512370367-15068acb-2717-48e4-ae7d-9496694fb901 (1)

https://marketplace.visualstudio.com/items?itemName=Bjarne.ethereal-omarchy

### DIFF
--- a/themes/ethereal/neovim.lua
+++ b/themes/ethereal/neovim.lua
@@ -1,44 +1,13 @@
 return {
   {
-    "bjarneo/aether.nvim",
-    name = "aether",
+    "bjarneo/ethereal.nvim",
+    dependencies = { "bjarneo/aether.nvim" }, -- Ensure aether is loaded first
     priority = 1000,
-    opts = {
-      disable_italics = false,
-      colors = {
-        -- Monotone shades (base00-base07)
-        base00 = "#060B1E", -- Default background
-        base01 = "#6d7db6", -- Lighter background (status bars)
-        base02 = "#060B1E", -- Selection background
-        base03 = "#6d7db6", -- Comments, invisibles
-        base04 = "#F99957", -- Dark foreground
-        base05 = "#ffcead", -- Default foreground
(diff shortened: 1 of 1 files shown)


dhh:
Same nvim error message as with the other theme:

image

bjarneo:
Sorry about that. Regression fixed. Now both themes have the naming fix for neovim.lua.

dhh:
Also, I like the colors of the background, but it's too low resolution for a 6K display. We'll need to find something that doesn't show compression artifacts etc. Even though 5mb is already kinda big!

bjarneo:
I am still looking into the lazyvim issue for the aether.nvim for both of these themes.

bjarneo:
For both the hackerman and the ethereal theme, the neovim switch does work.

However, it is not working until the themes are downloaded, which does not happen automatically at my end if I switch from i.e. osaka-jade to hackerman (after a lazyvim cache deletion).

Is there any way to preload these themes? This is not an issue with the other native themes.

dhh:
I think we need to add these themes to the omarchy-nvim package for that to work correctly, actually. I'll have a look at that.

dhh:
BTW, the highlight color in nvim when you have uncommitted changes looks like it wasn't tuned for the color scheme. It's very bright cyan. Is that as intended?

image

bjarneo:
No, not intended. It's known, but have not gotten around to fix it yet. The issue is the same for hackerman. I'll get it fixed today in aether.nvim, so regardless of this PR it will update as soon as I push the change to that repo.

bjarneo:
> No, not intended. It's known, but have not gotten around to fix it yet. The issue is the same for hackerman. I'll get it fixed today in aether.nvim, so regardless of this PR it will update as soon as I push the change to that repo.

Fixed

PR #3470, opened 2025-11-20
feat: add changelog display feature using gum

RushiChaganti:
### TL;DR 

Adds a clean GUM-based TUI to show changelogs before updating, includes a new conflict warning step, and requires jq, gum, and grep to parse release data.

### **Changelog Display Feature**

This PR adds a feature that shows users the latest changelogs, helping them stay informed about updates such as keybinding changes or newly added functionality.

### **How It Works**

* When the user selects **Update**, the script fetches the latest release information.
* The changelog is displayed using `gum paper`, ensuring it fits cleanly within the floating terminal without hiding confirmation options.
* After reviewing the changelog, the user can exit the view using **q** or **Esc**.
* Once closed, the user can choose to **accept** or **reject** the update by selecting **Yes** or **No**.

### **Additional Improvements**

* A new confirmation step warns users when their current config may conflict with the updated version.

### **Requirements**

To extract and display release contents, the following tools are required:

* `jq`
* `gum`
* `grep`

### DIFF
--- a/bin/omarchy-update-confirm
+++ b/bin/omarchy-update-confirm
@@ -1,14 +1,80 @@
-#!/bin/bash
-
+#!/usr/bin/env bash
 set -e
 
+REPO="basecamp/omarchy"
+API_URL="https://api.github.com/repos/$REPO/releases/latest"
+
+# Grab latest release info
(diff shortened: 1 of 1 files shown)


dhh:
Thanks for looking at this, but we have https://github.com/basecamp/omarchy/pull/1156 in the backlog already to tackle this ✌️

PR #3471, opened 2025-11-20
hyprmon TUI

landsman:
Hyprmon is a user-friendly TUI that helps me configure my monitors. It’s useful even with a single monitor, especially for users coming from macOS or Windows.

Tested locally on my machine.

Could you tell me what I should adjust or update to finish this task?

### DIFF
--- a/bin/omarchy-launch-hyprmon
+++ b/bin/omarchy-launch-hyprmon
@@ -1,0 +1,3 @@
+#!/bin/bash
+
+exec setsid uwsm-app -- xdg-terminal-exec --app-id=com.omarchy.Hyprmon -e env HYPRLAND_CONFIG=~/.config/hypr/monitors.conf hyprmon "$@"
--- a/migrations/1763323173.sh
+++ b/migrations/1763323173.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+
+pacman -S --noconfirm hyprmon-bin
--- a/bin/omarchy-menu
+++ b/bin/omarchy-menu
@@ -177,7 +177,7 @@ show_setup_menu() {
   *Wifi*) omarchy-launch-wifi ;;
   *Bluetooth*) omarchy-launch-bluetooth ;;
   *Power*) show_setup_power_menu ;;
-  *Monitors*) open_in_editor ~/.config/hypr/monitors.conf ;;
+  *Monitors*) omarchy-launch-hyprmon ;;
   *Keybindings*) open_in_editor ~/.config/hypr/bindings.conf ;;
   *Input*) open_in_editor ~/.config/hypr/input.conf ;;
   *Defaults*) open_in_editor ~/.config/uwsm/default ;;


woopstar:
This will overwrite the monitors file on update.

The current monitors file contains defaults such as:

```
# Optimized for retina-class 2x displays, like 13" 2.8K, 27" 5K, 32" 6K.
env = GDK_SCALE,2
monitor=,preferred,auto,auto
```

This should be moved, maybe?

bin/omarchy-launch-hyprmon
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+exec setsid uwsm-app -- xdg-terminal-exec --app-id=com.omarchy.Hyprmon -e env HYPRLAND_CONFIG=$HOME/.config/hypr/monitors.conf hyprmon "$@"
```
woopstar: Maybe change to `~/.config/hypr/monitors.conf` to use same format as everything else.

woopstar:
We can also consider to make use of the profile functionality that hyprmon provides, which is super nice. Adding it to the Waybar to easily select between monitor profiles?

migrations/1763323173.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+sudo pacman -S hyprmon
```
woopstar: can you change to use yay, and i believe the package is `hyprmon-bin`

This also requires a migration to make sure it is installed, since we change the omarchy-menu

landsman:
@woopstar Thanks for suggestions!

> We can also consider to make use of the profile functionality that hyprmon provides, which is super nice. Adding it to the Waybar to easily select between monitor profiles?

I’d love to improve the monitor configuration UX for newcomers. We could tackle this in the next PR. From my perspective, just opening a TUI where you can quickly switch profiles would be enough.

migrations/1763323173.sh
landsman: > can you change to use yay

I saw that `pacman` is used in migration multiple times already, so I used that.

> and i believe the package is hyprmon-bin

replaced by bin package, thanks.

landsman:
> This will overwrite the monitors file on update.
> 
> The current monitors file contains defaults such as:
> 
> ```
> # Optimized for retina-class 2x displays, like 13" 2.8K, 27" 5K, 32" 6K.
> env = GDK_SCALE,2
> monitor=,preferred,auto,auto
> ```
> 
> This should be moved, maybe?

From my tests, Hyprmon successfully loaded this configuration and rewrote only the lines containing the word "monitor" after the first change. It did not modify anything else in the file. I did not use profiles for this test.

Do you notice any unexpected behavior?

bin/omarchy-launch-hyprmon
landsman: done

dhh:
I think hyprmon is a nice tool for folks with multiple monitors when you need to do layouts, but I prefer keeping the default config file editing for the default one-monitor case.

woopstar:
> From my tests, Hyprmon successfully loaded this configuration and rewrote only the lines containing the word "monitor" after the first change. It did not modify anything else in the file. I did not use profiles for this test.

You are absolutely right. This will only alter the monitor lines.

landsman:
It's even suitable for the scale configurations.

https://github.com/user-attachments/assets/b953532b-524b-4b5a-a215-d8a4a0bbe2e6

woopstar:
> It's even suitable for the scale configurations.

Yes, it's super nice. I'll merge it my self. Shame this has been closed.

migrations/1763323173.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+pacman -S --noconfirm hyprmon-bin
```
woopstar, line 3: Needs to be installed through yay and not packman

bin/omarchy-launch-hyprmon
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
```
woopstar, line 1: File is not executable

migrations/1763323173.sh
landsman: pushed to my fork

bin/omarchy-launch-hyprmon
landsman: fixed in my fork

PR #3478, opened 2025-11-20
Update omarchy-update-system-pkgs

PierruMagnus:
Add option to skip AUR packages update.

My last Omarchy update broke because of one AUR package that could not be downloaded "qt5-webengine", even manually i could not update it, issue on code.qt.io server.


dhh:
Hmm, I see the issue, but I don't want to inject a confirm like this for every update. Would be better if we did like we do with migrations where we detect a failure and give you the option to continue.

PR #3479, opened 2025-11-20
Update Waybar tray icon

brink-lab:
Updates Waybar tray icon to an arrow (and updates spacing for icon) in line with tray icons used on Mac and Windows. 
This follows established practices and thereby helps users cross over with less friction.
Also, the currently used icon is more commonly used as a "log-out" icon for e.g. online profiles.

Update looks like this: 
image

### DIFF
--- a/config/waybar/style.css
+++ b/config/waybar/style.css
@@ -53,7 +53,7 @@
 }
 
 #custom-expand-icon {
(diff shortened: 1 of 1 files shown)


brink-lab:
Wow, that was quick :D

PR #3481, opened 2025-11-20
Use the running Emacs to open the editor.

mvaled:
When we install Emacs, we also enable the emacs.service; to avoid having two competing Emacs instances, let's open a frame if the service is running.

### DIFF
--- a/bin/omarchy-launch-editor
+++ b/bin/omarchy-launch-editor
@@ -9,6 +9,14 @@ case "$EDITOR" in
 nvim | vim | nano | micro | hx | helix | fresh)
   exec omarchy-launch-tui "$EDITOR" "$@"
   ;;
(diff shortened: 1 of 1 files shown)


dhh:
Shouldn't this use emacsclient -c -n (-c creates a new frame, -n returns immediately) instead of emacsclient -r for "reuse an existing frame"?

mvaled:
> Shouldn't this use emacsclient -c -n (-c creates a new frame, -n returns immediately) instead of emacsclient -r for "reuse an existing frame"?

My usual setup is a single Emacs frame parked  in workspace 3.  So, `-c` would be kind of disruptive.

mvaled:
Superseded by https://github.com/omacom/omarchy/pull/12807

PR #3484, opened 2025-11-20
Auto Theme Switching (v2)

LukasTrust:
Hello, this is my second attempt at creating an auto theme-switcher. This version reduces code and complexity. If you’re not interested, I completely understand. If you are, there are two things that need to be done:

1. Migrate the .service and .timer files to $HOME/.config/systemd/user.
2. Remove the minute setting, which was only intended for testing.

Here’s a quick 2× speed video demonstrating it working:

https://github.com/user-attachments/assets/b8fbac42-6043-47c2-ae4b-c8a15e02bb94

### DIFF
--- a/systemd/user/omarchy-theme-auto-switcher.timer
+++ b/systemd/user/omarchy-theme-auto-switcher.timer
@@ -2,7 +2,7 @@
 Description=Run Auto Theme Switcher periodically
 
 [Timer]
-OnCalendar=*-*-* *:*:00
+OnCalendar=monthly
 Persistent=true
 
 [Install]


ColtonIdle:
whats the point of auth theme switching. for day and night (like macOS?)

LukasTrust:
> whats the point of auth theme switching. for day and night (like macOS?)

It’s not a day/night theme switcher. The idea is to automatically rotate themes on whatever schedule the user prefers: hourly, daily, weekly, or even monthly. Personally, I often forget to switch my theme and I thought it would be a nice quality-of-life feature to have it happen automatically. To use the cool theme you download more often

dhh:
I think it's an interesting idea, but I think we're better off putting it in a separate package, like omarchy-auto-theme-switching on AUR or something like that 👍

PR #3486, opened 2025-11-21
Add keyboard backlight toggle script and bindings

presidentecarter:
Add functionality for keyboard brightness up and down keys, as well as when a computer only has a toggle key it cycles through the brightness levels. 

I don't use bash often so this is AI-enabled, however I believe this uses the principles set forth by DHH when this was built.


dhh:
Got this included now. Thanks for working on it, though ✌️

PR #3495, opened 2025-11-21
Remove duplication in omarchy-setup-dns

Jamesking56:
https://github.com/basecamp/omarchy/pull/3462#issuecomment-3557744362

This PR removes the duplicated logic in `bin/omarchy-setup-dns` by moving the duplicated logic into a function.


dhh:
Sorted now.

PR #3499, opened 2025-11-21
tte --reuse-canvas

felixzsh:
edit: this pr was originally thought as help for making possible the fullscreen on the screensaver, now i made another pr that actually solves that #3510 but i will keep this open as i still think the --reuse-canvas option should be used for tte in the omarchy screensaver.


felixzsh:
related to #3501

felixzsh:
in the changeblog of the tte creator, he mention he created this option for the omarchy screensaver use case:

https://chrisbuilds.github.io/terminaltexteffects/changeblog/changeblog_0.13.0/

PR #3503, opened 2025-11-21
Disable plocate database updates on boot

jbnunn:
Add plocate.sh config script to prevent database updates from running on boot by setting Persistent=false for plocate-updatedb. This prevents the timer from running catch-up updates immediately after boot if the system was off during the scheduled daily update time. Avoids I/O load during startup while still allowing daily scheduled updates to run (saved between 400-500ms on bootup in my tests).


dhh:
I think that's an interesting optimization, and I do like trying to squeeze more performance where we can, but I don't think this is worth it. The idea that the database is guaranteed to be up to date after a boot seems to outweigh the idea that 400ms is added during boot. But please do keep looking for these optimization opportunities!

PR #3504, opened 2025-11-21
Remove bluetooth from omarchy-menu for PCs w/o the hardware

jondkinney:
Also standardize on Wi-Fi (the registered trademark) that the update hardware menu already had correct

This also helps VMs not have the option in the menu since they don't support bluetooth without additional steps like adding a hardware USB module that can pass through the capabilities.

### DIFF
--- a/bin/omarchy-menu
+++ b/bin/omarchy-menu
@@ -171,7 +171,7 @@ show_font_menu() {
 }
 
 show_setup_menu() {
-  local options="  Audio\n  Wifi"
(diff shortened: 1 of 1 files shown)


dhh:
I appreciate the sentiment, but don't love hiding it here in the menu. These menus should preferably be stable so we can see what the options are. Would rather consider an error state if you do something with bluetooth on a system without it. But I also kinda think it doesn't matter.

Happy to take a separate change to the Wi-Fi thing.

PR #3510, opened 2025-11-22
feat: fullscreen for ghostty/kitty screensaver

### LINKED ISSUE #3501: no fullscreen on screensaver after omarchy 3.2.0 and tte 0.13
### System details

thinkad x1 carbon gen 7

### What's wrong?

update: the problem seems to be ghostty/kitty, fullscreen works flawless in alacritty

felixzsh:
currently the screensaver view is fullscreen only in alacritty, there has been changes in tte to improve the omarchy screensaver, but those were when alacritty was the default terminal. With these little changes, the fullscreen will be also possible in ghostty and kitty.

this closes #3501

### DIFF
--- a/bin/omarchy-cmd-screensaver
+++ b/bin/omarchy-cmd-screensaver
@@ -23,7 +23,7 @@ while true; do
     --no-eol --no-restore-cursor random_effect &
 
   while pgrep -x tte >/dev/null; do
-    if read -n 1 -t 3 || ! screensaver_in_focus; then
+    if read -n 1 -t 1 || ! screensaver_in_focus; then
       exit_screensaver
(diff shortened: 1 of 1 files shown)


dhh:
Strange. I'm getting fullscreen kitty and Ghostty screensavers. Can you share a picture of what you're seeing?

felixzsh:
without these changes what im seeing for example with unstable effect and all others that uses the whole screen in some way is this :

image


as you can see, the effect doesnt actually touches the screen borders, where with these changes (setting paddings to zero), this can be solved:

image

dhh:
Ahhh. I see what you mean 👌

PR #3544, opened 2025-11-22
Fixes #3543: check for limine.conf in /boot as secondary fallback location

### LINKED ISSUE #3543: install fails on omarchy 3.2 with limine-snapper script expecting /boot/limine/limine.conf
### System details

Surface Book 2

### What's wrong?

Surface Book 2, so after archinstall using the setup from https://learn.omacom.io/2/the-omarchy-manual/96/manual-installation, I follow the guide from surface-linux here https://github.com/linux-surface/linux-surface/wiki/Installation-and-Setup.

Since I was hit by #3335 during my last install attempt some weeks ago I moved /boot/EFI/limine/limine.conf to /boot/limine.conf this time, even though the issue was closed. Update limine.conf to add entry with surface-linux and reboot with secure boot works fine.

During omarchy 3.2 install, I see

`Updated: /boot/limine.conf`

which makes me happy. However, it is followed by:

```
v Signed /boot/EFI/limine/limine_x64.efi
(5/7) Reloading system bus configuration
(6/7) Updating icon theme caches...
(7/7) Updating the desktop file MIME type cache...
Error: Limine config not found at /boot/limine/limine.conf

This command halted with exit code 0:
Failed script: /home/zark/.local/share/omarchy/install/login/limine-snapper.sh
```

There was no /boot/limine/limine.conf after archinstall and I did not move it there, so I guess that is correct - but then why does it fail on this path? 

I have no folder `/boot/limine/`

ZarK:
Use /boot/limine.conf as a secondary fallback path. 

Fixes #3543

### DIFF
--- a/install/login/limine-snapper.sh
+++ b/install/login/limine-snapper.sh
@@ -16,7 +16,11 @@ EOF
       limine_config="/boot/EFI/limine/limine.conf"
     fi
   else
-    limine_config="/boot/limine/limine.conf"
+    if [[ -f /boot/limine/limine.conf ]]; then
+      limine_config="/boot/limine/limine.conf"
+    else
+      limine_config="/boot/limine.conf"
+    fi
   fi
 
   # Double-check and exit if we don't have a config file for some reason


ryanrhughes:
Did this a little differently in https://github.com/basecamp/omarchy/commit/c37ef6d78ccd21b9f0fda583203de7447d9006ec but gave you co-author credit. Thank you for pushing this up!

PR #3553, opened 2025-11-22
Exclude new long screensaver effect

MirasMustimov:
Hello!

tte 0.13 introduced new effect called dev_worm which takes too long to complete compared to other effects and might still be in development based on its name.

this pr disables that effect from screensaver.

to do that I reverted usage of random_effect to how random effect was obtained before.

the effect
image


Graphical27:
ty man i was looking for this

dhh:
Agree. Just hit this, and it is indeed very slow to complete.

PR #3560, opened 2025-11-23
Adding toggle file to lock the screen on return from screensaver

riquemon:
This adds a check for a `lock-after-screensaver` toggle file.
If the toggle file is present, the `exit_screensaver` routine will also lock the session in addition to closing the screensaver.

Allows paranoid users like me to walk away from the gorgeous screensaver at the coffee shop to impress my friends,
while also protecting my login session from said friends.


dhh:
I think this is a bit too hidden. And don't want to train anyone that it's safe to walk away without actually locking. Many things could go wrong if that's not explicitly done. But appreciate the suggestion!

PR #3577, opened 2025-11-23
Add support for Firefox in Webapp Launcher.

ghost:


dhh:
Firefox doesn't have the `--app` flag for a frameless design that we use for web apps in Omarchy. So for a cohesive aesthetic, we stick with Chrome for those. But you can of course always just alter the .desktop files generated for the web apps to use Firefox directly.

PR #3588, opened 2025-11-24
Omarchy 3.2.1

dhh:
# What changed?
* Add release channel and last update date for packages to the About screen by @dhh
* Add copy-url-to-clipboard (hotkey `Alt + Shift + L`) for Brave by @Blinkuu 
* Fix screensaver to be truly fullscreen on Ghostty/Kitty as it is on Alacritty by @felixzsh
* Fix `omarchy-refresh-pacman` should backup `/etc/pacman.conf` and `/etc/pacman.d/mirrorlist` before overwriting by @dhh
* Fix PDF viewer should start in floating mode (like image and video viewers) by @dhh
* Fix Steam + Zed installers to give user a chance to pick the correct vulkan backend for their system by @dhh
* Fix screensaver should not include the new experimental (and very slow) `dev_worm` effect by @MirasMustimov
* Fix _Remove > TUI_ to work with new xdg-terminal setup by @neurapy
* Fix window pop wasn't applying the right tag and required double invocation to undo by @dhh
* Fix Bluetooth icon wouldn't switch to off when turning off the BT adapter in BlueTUI by @dhh
* Fix `omarchy-refresh-walker` should ensure that Elephant service is running + add Walker to autostart by @dhh

### DIFF
--- a/bin/omarchy-channel-set
+++ b/bin/omarchy-channel-set
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+if (($# == 0)); then
+  echo "Usage: omarchy-channel-set [stable|edge|dev]"
+  exit 1
+else
+  channel="$1"
+fi
+
+case "$channel" in
+"stable") omarchy-branch-set "master" && omarchy-refresh-pacman "stable" ;;
+"edge") omarchy-branch-set "edge" && omarchy-refresh-pacman "stable" ;;
+"dev") omarchy-branch-set "dev" && omarchy-refresh-pacman "edge" ;;
+*) echo "Unknown channel: $channel"; exit 1; ;;
+esac
--- a/bin/omarchy-branch-set
+++ b/bin/omarchy-branch-set
@@ -1,0 +1,14 @@
+#!/bin/bash
+
+if (($# == 0)); then
+  echo "Usage: omarchy-channel-set [master|dev]"
+  exit 1
+else
+  branch="$1"
+fi
+
+case "$branch" in
+  "master") git -C $OMARCHY_PATH switch master ;;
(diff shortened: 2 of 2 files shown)


bin/omarchy-channel-set
```diff
@@ -0,0 +1,15 @@
...
+if (($# == 0)); then
+  echo "Usage: omarchy-channel-set [stable|edge|dev]"
+  exit 1
+else
+  channel="$1"
+fi
+
+case "$channel" in
+"stable") omarchy-branch-set "master" && omarchy-refresh-pacman "stable" ;;
+"edge") omarchy-branch-set "edge" && omarchy-refresh-pacman "stable" ;;
```
Niknu: Would this throw a error as `omarchy-branch-set` only have `master` and `dev` ?

I can see this is already has been merged so I don't know if this make sense to handle this comment?
I assume `edge` would be handle later?

bin/omarchy-cmd-reboot
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+omarchy-state clear re*-required
+omarchy-hyprland-window-close-all
+sleep 1 # Allow apps like Chrome to shutdown correctly
+systemctl reboot --no-wall
```
Niknu: When we are using systemctl to `reboot` or `shutdown` the system. Isn't all the lines above actually need?
Because from my knowledge when we are using systemctl we using the `systemd` and `systemd` will handle correct shutdown of the different application so a sleep shouldn't be needed. It will just delay the shutdown with 1s.

It's in the lower end of optimization so I can also see why this isn't really important to change.

Issue #3592, opened 2025-11-24
AUR TUI encoded characters


amadeu01:
### System details

AMD Radeon RX 7600, AMD Ryzen 5 9600X, Omarchy 3.1.7

### What's wrong?

<img width="1481" height="1018" alt="Image" src="https://github.com/user-attachments/assets/d0e96574-a981-4121-999b-ee916707915c" />

TUI is showing enconded text

squishydal:
Yup it happen to me too.

grimkraut:
I can confirm this too, on 2 machines! On 3.2.0

dev-oskar:
I have the same issue

justindotdevv:
yep, same issue here
on `3.2.0`

gardnmi:
Same Issue:

<img width="868" height="586" alt="Image" src="https://github.com/user-attachments/assets/fd370890-1db0-45fc-addf-ae40b4ab1fb3" />

<img width="850" height="558" alt="Image" src="https://github.com/user-attachments/assets/877ec2b4-fe97-49f0-a572-b9b2b29bb1c4" />

3bit:
I just updated to 3.2.0 and have the same issue.

CGAlei:
Yup, same for me, everything works perfect, only AUR installer has encoded chars.

justindotdevv:
looks like the `yay -Slqa` from the `omarchy-pkg-aur-install` script is returning encoded characters instead of normal text. 

below is the output after running `yay -Slqa`

<img width="1067" height="1386" alt="Image" src="https://github.com/user-attachments/assets/eaa7c557-97c5-474c-9e3c-e7e44ca4fafe" />

lai2301:
I found that the arg of yay command stop working.
I just simply change the command in `~/.local/share/omarchy/bin/omarchy-pkg-aur-install`
Update the line 17 from `yay -Slqa` to `yay -qSs`
```
#!/bin/bash

fzf_args=(
  --multi
  --preview 'yay -Siia {1}'
  --preview-label='alt-p: toggle description, alt-b/B: toggle PKGBUILD, alt-j/k: scroll, tab: multi-select'
  --preview-label-pos='bottom'
  --preview-window 'down:65%:wrap'
  --bind 'alt-p:toggle-preview'
  --bind 'alt-d:preview-half-page-down,alt-u:preview-half-page-up'
  --bind 'alt-k:preview-up,alt-j:preview-down'
[...]

PR #3593, opened 2025-11-24
#3592: Fix omarchy-pkg-aur-install

LukasTrust:
Issue: https://github.com/basecamp/omarchy/issues/3592


dhh:
Don't think this is right. I only see 15K packages. Not 80K. I think AUR is having trouble (again).

dhh:
We are going to include the `yay-gzip-fix` package until yay fixes itself.

PR #3595, opened 2025-11-24
Convert r alias into a function that uses bin/rails if available (1 line change)

barisozmen:
Turns the `r` alias into a function that checks for `bin/rails` first and use it if available, otherwise falls back to `rails`

## How to test
- Run `r` in a Rails project (where `bin/rails` present) - should use `bin/rails`
- Run `r` from another path where `bin/rails` is not available - should fall back to `rails`

## Why?
I recently start learning Rails and if I understood correctly it is a better practice to use `bin/rails` over `rails`. This change makes it effortless to follow the better practice.

### DIFF
--- a/default/bash/aliases
+++ b/default/bash/aliases
@@ -32,7 +32,7 @@ alias ....='cd ../../..'
 
 # Tools
 alias d='docker'
-alias r='rails'
+r() {if [ -x "bin/rails" ]; then bin/rails "$@"; else rails "$@"; fi}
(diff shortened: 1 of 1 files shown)


dhh:
The r will automatically use bin/rails when it's being used from ~/Work because mise has been setup to add bin from that directory into path. So should already be good 👍

barisozmen:
Thanks for the clarification! I'm new to Omarchy and I'm glad to learn about this.

PR #3596, opened 2025-11-24
kernel: Keep modules functional after a kernel upgrade

defer:
When a kernel upgrade is taken in, we replace current modules with new ones and they are not available until we reboot. This can cause the system to appear broken, especially when plugging new hardware like a pen drive or odd input device.

This patch adds kernel-modules-hook which keeps the current modules even if the package is replaced. A one-off clean-up service runs on startup
  to remove older modules.


Traap:
@defer - I would benefit from this PR.  My work computer requires customization because of Microsoft Active Directory integration and Thinkpad docking station requirements needed to use USB, WiFI, and Ethernet ports - display and hdmi ports are not, sadly supported.

woopstar:
@defer  please rebase :)

PR #3655, opened 2025-11-27
Feat: Screensaver displays current date/time or a string of text using figlet

lucaengelhard:
Let the screensaver display the current time, date or a string of text based on a gnu date datetime format string.

image

# Features
- Set gnu date format string in config file and format it with figlet
- Set a string to display with figlet
- Change the font figlet uses

# Changes
- `config/omarchy/branding/screensaver.conf`
  - Created new screensaver config file
  - Defaults are already set and commented out, so legacy behavior stays the same
- `bin/omarchy-launch-screensaver`
  - Load configuration
  - Check if font file exists
  - Priority: **time > text > file**
  - Only execute if figlet is installed

# Further considerations
- Merging this would make figlet a needed dependency
- The priority is very opinionated
- Effects on already running omarchy installations

### DIFF
--- a/config/omarchy/branding/screensaver.conf
+++ b/config/omarchy/branding/screensaver.conf
@@ -4,3 +4,6 @@
 # Set the path to the ascii-font the screensaver should use
 # Uses figlet (https://en.wikipedia.org/wiki/FIGlet)
 # font=
+
+# Set the text content of the screensaver (datetime takes priority)
+# text=Omarchy


dhh:
That's a neat suggestion, but I prefer our Omarchy logo default.

lucaengelhard:
Understandable. The default still is the Omarchy logo and there is the option to change it through a config file

PR #3661, opened 2025-11-27
Update Plymouth theme to enable smoother progress bar

mlombardi96:
This PR updates the Omarchy Plymouth theme so that there is a smoother handoff between fake and real progress once the drive is decrypted. It also includes some formatting changes so that the script it easier to read.

Before (big jump in progress once the drive is unlocked):

https://github.com/user-attachments/assets/232664bd-e483-46aa-977a-07b7ee8bf400

After (smooth progress bar all the way through boot):

https://github.com/user-attachments/assets/2b69bdc2-0153-4b4e-8faf-cd789ef18428


default/plymouth/omarchy.script
mlombardi96, line 1: Due to all the formatting changes, it's easier to review this file with whitespace hidden.

default/plymouth/omarchy.script
```diff
@@ -214,44 +198,38 @@ progress_bar.y = progress_box.y + (progress_box.image.GetHeight() - progress_bar
...
-  {
-    global.real_progress = progress;
-    
-    # If real progress is above limit, stop fake progress and use real progress
-    if (progress > global.fake_progress_limit)
-      {
-        stop_fake_progress();
-        update_progress_bar(progress);
-      }
+fun progress_callback(duration, progress) {
```
mlombardi96, line 201: This is the main change. The `progress_callback` function is called several times a second throughout the boot process. 

However, while the drive is being decrypted, the `duration` and `progress` arguments stop updating and are frozen until the drive is unlocked.

To handle this, we record the `duration` value the moment fake progress starts. Once the drive is unlocked, duration/progress will start updating again. 

We switch to showing real progress once the fake progress values are exceeded so that there's a seamless handoff.

default/plymouth/omarchy.script
```diff
@@ -214,44 +198,38 @@ progress_bar.y = progress_box.y + (progress_box.image.GetHeight() - progress_bar
...
-Plymouth.SetBootProgressFunction(progress_callback);
-
-#----------------------------------------- Quit --------------------------------
+  global.real_progress = progress;
 
-fun quit_callback ()
-{
-  logo.sprite.SetOpacity (1);
+  # Use real progress once its unfrozen and exceeds fake progress
+  if (duration > global.fake_progress_start_time && progress > global.fake_progress) {
```
mlombardi96, line 211: I experimented with not tracking the starting duration at all and only having `progress > global.fake_progress` in this condition.

However, there are certain scenarios where the "frozen" real progress value is greater than the fake progress value as soon as the progress bar is shown, which results in only the real progress being shown the entire boot.

PR #3686, opened 2025-11-29
fix: reorder mise activation to prevent PROMPT_COMMAND hook overwrite

plgonzalezrx8:
- Moves `mise activate` to the end of `bash/init` so its `PROMPT_COMMAND` hook isn't overwritten by starship and zoxide
- Ensures mise-managed tools (Python, Node, etc.) are correctly available in new terminal sessions


dhh:
I haven't seen any problems with the current setup. Can you explain the problem you're seeing?

PR #3687, opened 2025-11-29
Let the screensaver run in Wezterm

db48x:
Because Wezterm is nice.

Alternatively, perhaps it would be better for the screensaver to always use a specific terminal emulator regardless of the user’s preference for interactive terminals.

### DIFF
--- a/bin/omarchy-launch-screensaver
+++ b/bin/omarchy-launch-screensaver
@@ -20,6 +20,11 @@ for m in $(hyprctl monitors -j | jq -r '.[] | .name'); do
   hyprctl dispatch focusmonitor $m
 
   case $terminal in
+  *wezterm*)
+    hyprctl dispatch exec -- \
+      wezterm start --class=org.omarchy.screensaver --no-auto-connect \
+        --always-new-process omarchy-cmd-screensaver
+    ;;
   *Alacritty*)
     hyprctl dispatch exec -- \
       alacritty --class=org.omarchy.screensaver \
@@ -41,7 +46,7 @@ for m in $(hyprctl monitors -j | jq -r '.[] | .name'); do
       -e omarchy-cmd-screensaver
     ;;
   *)
-    notify-send "✋  Screensaver only runs in Alacritty, Ghostty, or Kitty"
+    notify-send "✋  Screensaver only runs in Wezterm, Alacritty, Ghostty, or Kitty"
     ;;
   esac
 done


iltumio:
I'm using the same patch in my system to make the screensaver run in wezterm

omarchy launch screensaver

```bash
#!/bin/bash

# Launch the Omarchy screensaver in the default terminal on the system with the correct font configuration.

# Exit early if we don't have the tte show
if ! command -v tte &>/dev/null; then
  exit 1
fi

# Exit early if screensave is already running
pgrep -f org.omarchy.screensaver && exit 0

# Allow screensaver to be turned off but also force started
if [[ -f ~/.local/state/omarchy/toggles/screensaver-off ]] && [[ $1 != "force" ]]; then
  exit 1
fi

# Silently quit Walker on overlay
walker -q

focused=$(hyprctl monitors -j | jq -r '.[] | select(.focused == true).name')
terminal=$(xdg-terminal-exec --print-id)

for m in $(hyprctl monitors -j | jq -r '.[] | .name'); do
  hyprctl dispatch focusmonitor $m

  case $terminal in
  *Alacritty*)
    hyprctl dispatch exec -- \
      alacritty --class=org.omarchy.screensaver \
      --config-file ~/.local/share/omarchy/default/alacritty/screensaver.toml \
      -e omarchy-cmd-screensaver
    ;;
  *ghostty*)
    hyprctl dispatch exec -- \
      ghostty --class=org.omarchy.screensaver \
      --config-file=~/.local/share/omarchy/default/ghostty/screensaver \
      --font-size=18 \
      -e omarchy-cmd-screensaver
    ;;
  *kitty*)
    hyprctl dispatch exec -- \
      kitty --class=org.omarchy.screensaver \
      --override font_size=18 \
      --override window_padding_width=0 \
      -e omarchy-cmd-screensaver
    ;;
  *wezterm*)
    hyprctl dispatch exec -- \
      wezterm --config-file ~/.config/wezterm/screensaver.lua \
      start --always-new-process \
      --class org.omarchy.screensaver \
      -- omarchy-cmd-screensaver
    ;;
  *)
    notify-send "Screensaver only runs in Alacritty, Ghostty, Kitty, or WezTerm"
    ;;
  esac
done

hyprctl dispatch focusmonitor $focused

```

In addition to this, I'm loading a dedicated wezterm config file just for the screensaver:

```lua
local wezterm = require("wezterm")
local config = wezterm.config_builder()

-- Screensaver: black background, large font, no chrome
config.colors = {
  foreground = "#a9b1d6",
  background = "#000000",
  cursor_bg = "#000000",
  cursor_fg = "#000000",
  cursor_border = "#000000",
}

config.font = wezterm.font("JetBrainsMono Nerd Font", { weight = "Regular" })
config.font_size = 18

config.window_decorations = "NONE"
config.enable_tab_bar = false
config.window_background_opacity = 1.0

config.window_padding = {
  left = 0,
  right = 0,
  top = 0,
  bottom = 0,
}

config.window_close_confirmation = "NeverPrompt"
config.default_cursor_style = "SteadyBlock"

return config
```

iltumio:
I opened another PR to add full support to wezterm here #4836. This should solve this screensaver issue and provide full support to wezterm in omarchy

tuxclaw:
Tested a local WezTerm screensaver adaptation on Omarchy 4.0.3-1 with WezTerm using native Wayland.

The packaged launcher still rejects `org.wezfurlong.wezterm.desktop`, including when launched by the idle timer.

Adding `*wezterm*` to the allowlist and using this branch with the current launcher's existing per-monitor sequencing works:

```bash
hypr_exec wezterm --config-file "$HOME/.config/omarchy/wezterm-screensaver.lua" start \
  --always-new-process --no-auto-connect --class org.omarchy.screensaver -- omarchy-screensaver
```

Dedicated configuration used:

```lua
local wezterm = require 'wezterm'
return {
  font = wezterm.font 'Fira Code',
  font_size = 18.0,
  enable_wayland = true,
  enable_tab_bar = false,
  enable_scroll_bar = false,
  window_decorations = 'RESIZE',
  window_padding = { left = 0, right = 0, top = 0, bottom = 0 },
  window_background_opacity = 1.0,
  colors = { background = '#000000', foreground = '#ffffff' },
  check_for_updates = false,
  exit_behavior = 'Close',
}
```

Live validation confirmed separate fullscreen windows on landscape and portrait monitors, both with the exact `org.omarchy.screensaver` app ID, two running `ttfx` processes, and successful launcher exit. Focusing the original application dismissed both screensaver windows automatically; none remained after two seconds. The dedicated config also parsed successfully with `wezterm --config-file ... show-keys`.

This was a user-owned launcher adaptation, not a checkout/test of this PR's branch. Sharing results to help move the existing proposal forward.

PR #3690, opened 2025-11-29
Fix window matching for PWAs

### LINKED ISSUE #3669: Bug regarding "omarchy-launch-or-focus" and ghostty
### System details

i7-1280P, omarchy 3.2.1

### What's wrong?

Hey there,
somewhat new to Omarchy, Hyprland, and all that goodness.
I was trying to test/debug my autostart commands, having the unreasonable wish to start all my communication apps in a special workspace.

Now when running `omarchy-launch-or-focus-webapp Whatsapp "https://web.whatsapp.com"` the ghostty window just get's focused. This seems due to the fact that in `omarchy-launch-or-focus` this part of the **jq**-query `(.title|test("\\b" + $p  + "\\b";"i"))` always get's evaluated to the opened ghostty window.

This maybe related to b5002b1

FYI, here is the complete command I tried, when I first noticed this behavior:
`hyprctl dispatch exec [workspace special:communication silent] omarchy-launch-or-focus-webapp WhatsApp "https://web.whatsapp.com"`

[omarchy-debug.log](https://github.com/user-attachments/files/23822146/omarchy-debug.log)

MellowPhi:
The `\b` word boundaries in regex weren't matching browser PWA classes like `brave-web.whatsapp.com__-Default` Dots broke the boundaries, so the script skipped the real web app and grabbed whatever came first—like a terminal editing Whatsapp.desktop.
The fix: Swap `\b` for `(^|[\\.-])` before the pattern and `($|[\\.-])` after. Now it matches when the app name is at the start/end or next to dots/hyphens.
This keeps things precise (no false positives) while fixing PWAs. 
One line changed, focusing back to WhatsApp now works correctly.

Fixes: https://github.com/basecamp/omarchy/issues/3669

### DIFF
--- a/bin/omarchy-launch-or-focus
+++ b/bin/omarchy-launch-or-focus
@@ -7,7 +7,7 @@ fi
 
 WINDOW_PATTERN="$1"
 LAUNCH_COMMAND="${2:-"uwsm-app -- $WINDOW_PATTERN"}"
-WINDOW_ADDRESS=$(hyprctl clients -j | jq -r --arg p "$WINDOW_PATTERN" '.[]|select((.class|test("\\b" + $p + "\\b";"i")) or (.title|test("\\b" + $p + "\\b";"i")))|.address' | head -n1)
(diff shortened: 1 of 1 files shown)


dhh:
I can't actually replicate the problem any more. Maybe this got fixed some other way?

PR #3699, opened 2025-11-29
Add GIF option to screenrecord menu

bhood239:
- Saves flag and file name to tmp files, then if exists on stop_screenrecording call, converts and cleans up tmp files.

- Includes converting and saved notifications

- Saves to pictures folder if gif instead of videos

Tested this on my own system and works great

Menu option:
screenshot-2025-11-29_11-46-27


dhh:
I like the option of being able to convert to gif, but I'd rather we just do it as a function, like the other transcode functions. Please take a stab at that in a new PR 🙏

dhh:
Added a gif transcoding function 👍

PR #3701, opened 2025-11-29
Custom screenshot filename

birkirb:
Small tweak to allow for the customization of the screenshot filename through 3 additional environment variables. 
Lot's of other ways to do it of course.
This one doesn't change the defaults, and overrides can be put in the `.config/hypr/env.conf`, much like `OMARCHY_SCREENSHOT_DIR`

e.g.
```
env = OMARCHY_SCREENSHOT_DATE_FORMAT,%Y.%m.%dT%H.%M.%S
env = OMARCHY_SCREENSHOT_PREFIX,false
env = OMARCHY_SCREENSHOT_SUFFIX,_screenshot.png
```


dhh:
Appreciate the suggestion, but find this too much configuration for something like this.

PR #3706, opened 2025-11-30
fix: close walker menu when already open with keybind

ashvwinn:
When you open omarchy menu, system menu or keybinds list menu with the specified keybind, pressing the keybind again does not close it rather shows "No Results" in the same menu.

**Example:**
1. Open omarchy menu with `SUPER + ALT + SPACE`
2. Press same keybind again while it is open. It will show "No Results".

The intended behavior should be to close the already open menu.

### DIFF
--- a/bin/omarchy-menu
+++ b/bin/omarchy-menu
@@ -468,6 +468,12 @@ go_to_menu() {
   esac
 }
 
+# Close walker if it's already open
+if hyprctl layers | grep -q walker; then
+  walker -q
+  exit 0
+fi
+
 if [[ -n "$1" ]]; then
   BACK_TO_EXIT=true
   go_to_menu "$1"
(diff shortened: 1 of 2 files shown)


HegewaldK:
There is a parameter in walker/config.toml for this: `close_when_open = true`
This is not part of the default Omarchy config, but maybe it's what you want.

ashvwinn:
That is turned on by default in walker.
The issue probably has to do with how omarchy launches the menu because this is not a issue if you open the applications menu with `SUPER + SPACE`

iop098321qwe:
In my case, when opening the Omarchy Menu using `SUPER + ALT + SPACE` I am able to close it using `SUPER + SPACE` without issue. This is how I have been closing it at least, not sure if that helps you at all.

dhh:
This should already be working correctly now.

PR #3707, opened 2025-11-30
Add graphical tablet support

pgmdpfcon:


dhh:
Not useful like this without more information and follow up. It looks like a mystery assignment as-is.

PR #3711, opened 2025-11-30
Automatic fallback to Alacritty for legacy GPUs when install Omarchy

yaroslavrick:
## Description:

Closes https://github.com/basecamp/omarchy/issues/3581

## Context

Ghostty requires modern GPU acceleration (OpenGL/Vulkan) which is often unstable or missing on legacy hardware, specifically pre-GCN AMD GPUs using the `radeon` kernel driver. On these systems, Ghostty may crash immediately, rendering the system difficult to use post-install.

https://github.com/user-attachments/assets/dcf2f804-d903-4cfb-9d4e-a44e88470800

## Changes

This PR implements an automatic detection and fallback mechanism:

### Detection:

Added `bin/omarchy-detect-legacy-gpu` which checks for the presence of the `radeon` driver in `/sys/class/drm`. This script is designed to be extensible for other legacy drivers in the future.

### Fallback:

Added `install/config/hardware/legacy-gpu-terminal.sh` which runs during installation. If legacy hardware is detected, it calls `omarchy-install-terminal alacritty` to set Alacritty as the **default** terminal.

### Package:

Added `alacritty` to `install/omarchy-base.packages` to ensure the fallback terminal is always available.

### Hook:
Wired up the new configuration script in `install/config/all.sh`

### Verification

- Verified that the fallback script correctly returns legacy driver name (radeon in this case):

screenshot-2025-11-30_17-14-12

- Verified that the fallback script works with modern GPU (in Arch linux):

modern_gpu


yaroslavrick:
Because of I do not have not legacy laptop, it worth to run this script on modern machine to check:

```bash
for d in /sys/class/drm/card*/device/driver; do basename $(readlink -f $d); done
```

install/config/hardware/legacy-gpu-terminal.sh
```diff
@@ -0,0 +1,3 @@
+if omarchy-detect-legacy-gpu; then
```
dhh: Let's just inline the check here. I don't think this has any general utility outside this use case at the moment.

Also, it should do a full swap of Ghostty for Alacritty. Which will involve setting it as the default terminal etc. Not just installing it.

yaroslavrick: Thank you for the suggestion!

[Fixed](https://github.com/basecamp/omarchy/pull/3711/commits/fd5e69be2e234f9688c51bcf010d6a194b5f8089)

yaroslavrick: @dhh Just following up — I've simplified the GPU detection and added the full Ghostty→Alacritty swap as you suggested. Let me know if anything else is needed.

dhh:
Very nice!

neheb:
Here's the interesting thing:

Current arch uses the radeon driver for GCN 1.0, 1.1, 1.2. Kernel 6.19 changes that to amdgpu for those GPUs. This will eventually be fixed once arch starts shipping 6.19 but worth a note.

neheb:
This also doesn't fix https://github.com/basecamp/omarchy/issues/4062

I don't know how that would be handled.

PR #3738, opened 2025-12-02
feat: enable wayland for sddm

vander00:
This PR switches SDDM to run on Wayland by default.
The main benefit is avoiding the extra Xorg server that SDDM normally launches, saving roughly 200 MB of RAM.

KDE Plasma already uses Wayland for SDDM in its own setups, so this approach is expected to work smoothly. Still minor issues can’t be completely ruled out.

### DIFF
--- a/bin/omarchy-install-gaming-retroarch
+++ b/bin/omarchy-install-gaming-retroarch
@@ -21,7 +21,10 @@ omarchy-pkg-add \
   libretro-parallel-n64 libretro-picodrive libretro-play libretro-ppsspp \
   libretro-sameboy libretro-scummvm libretro-shaders-slang libretro-snes9x \
   libretro-yabause \
-  libretro-cap32-git libretro-fbneo-git libretro-uae-git libretro-vice-git \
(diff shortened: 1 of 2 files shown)


patroza:
Interestingly my sddm stopped working after nvidia driver changes, your change saved the day.

install/login/sddm.sh
```diff
@@ -11,5 +11,18 @@ Current=breeze
 EOF
 fi
 
+if [ ! -f /etc/sddm.conf.d/toWayland.conf ]; then
+  cat <<EOF | sudo tee /etc/sddm.conf.d/toWayland.conf
+[General]
+HaltCommand=/usr/bin/systemctl poweroff
+RebootCommand=/usr/bin/systemctl reboot
```
yevshev: No need to set these here, they're already set in `/usr/lib/sddm/sddm.conf.d/default.conf`:

https://man.archlinux.org/man/sddm.conf.5#HaltCommand=
https://man.archlinux.org/man/sddm.conf.5#RebootCommand=

PR #3741, opened 2025-12-02
New Backgrounds added to ethereal theme.

pr1hm:


dhh:
The first two are too low-res to work on a 6K screen, but the last one is very nice. I've added it and given you credit 👍

PR #3748, opened 2025-12-03
Override zoom webapp browser to chromium if default browser is brave

damirsehic:
Something changed with one of latest brave changes in the last couple of days. The share screen button does not show up in meetings. When wrapping zoom webapp in chromium, it works as expected.


slinlee:
This worked for me as a workaround. Thanks, @damirsehic !

dhh:
Appreciate the idea, but I don't really want to collect exceptions like this that are bound to be temporary. Best to alert the Brave team, then set Chrome as your default browser until fixed.

PR #3750, opened 2025-12-03
Small improvements to the 'update available' waybar module

pmpinto:
## Display the version number of the available update
I wanted to know ahead, which version number is available when the 'update available' module appears in the waybar.

Since the `omarchy-update-available` command already returns a string like `Omarchy update available (v3.2.2)`, might as well just use it instead of hard coding the string without the version number.

Preview:
screenshot-2025-12-03_11-20-07

## Open the GitHub releases page when right clicking the module
Also, since I always read the release notes (as you should) before updating, I wanted a quick way to do it. So I added a right-click function to the update module that opens `https://github.com/basecamp/omarchy/releases`.

Originally brought up here: https://github.com/basecamp/omarchy/discussions/3324

### DIFF
--- a/migrations/1765198150.sh
+++ b/migrations/1765198150.sh
@@ -1,18 +1,25 @@
 echo "Display update version number and allow right-click to open release notes"
 
-WAYBAR_CONFIG="$HOME/.config/waybar/config_test.jsonc"
+WAYBAR_CONFIG="$HOME/.config/waybar/config.jsonc"
 
-# Check only inside custom/update for on-click-right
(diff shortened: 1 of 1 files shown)


dhh:
I like it. But needs a migration. See 1764067068.sh for an example.

pmpinto:
@dhh Was unaware of the migration system. Please check again.
Tested on my config and does what I believe is expected:
* Adds `on-click-right` only when it's not already present inside `custom/update`;
* Adds `tooltip-format` only when it's missing inside `custom/update`, and replaces the whole line if it's set to the default value;

dhh:
Appreciate your work on this, but on second thought, I think we're better off with the general link to changes and the confirmation you get when you just click this. We have all the information there.

PR #3752, opened 2025-12-03
Walker: Add runner provider prefix for quick command execution

avinashjoshi:
This PR adds a `!` prefix for the runner provider in Walker, enabling quick command execution through the launcher interface. For example: `!omarblue` shows bluetooth-related commands

image


dhh:
Appreciate the suggestion, but I don't think it works well, as there's no way to add parameters or get feedback from the command that's launched.

PR #3753, opened 2025-12-03
add comment about seemingly stale submenu entries in main menu list

Dieterbe:


dhh:
Appreciate this, but I don't think it's necessary. Right under this block is the segment that shows the CLI method that's being invoked.

PR #3757, opened 2025-12-04
Add scroll actions for workspace navigation

Rishabh-Sarang:


dhh:
Interesting idea, but we already have Super + Mouse Scroll to move between workspaces.

PR #3759, opened 2025-12-04
Add Fizzy webapp

dcchambers:
Add a webapp for [Fizzy](https://www.fizzy.do/), the new kanban tool from 37Signals.

The icon is sourced from https://dashboardicons.com/community/fizzy (Seems official, by @jzimdars)


jbnunn:
Seems pretty niche. And given that a lot of people that use Fizzy may choose to run it on their own hardware, this one might be better left for the user to add themselves.

dhh:
Thanks @dcchambers, I've been meaning to add this! The collection of web apps preinstalled with Omarchy is the list of webapps that I personally use every day. You can remove any that you don't want under _Remove > Web Apps_.

PR #3760, opened 2025-12-04
Fix keybinds description

kratusch:
Should be Shrink window left and Expand window right respectively.


dhh:
I think what we have now is correct. If you have a window in the middle, it does what the description says it does.

kratusch:
But that doesn't make sense as the direction is opposite.

Look at these for example shrink window up and expand window down makes sense:
https://github.com/basecamp/omarchy/blob/8a37f3dce71b0764e3758cb86478c503a4a8746e/default/hypr/bindings/tiling-v2.conf#L84 and https://github.com/basecamp/omarchy/blob/8a37f3dce71b0764e3758cb86478c503a4a8746e/default/hypr/bindings/tiling-v2.conf#L85

So should Shrink window left and Expand window right, not Shrink window left and Expand window left.

PR #3761, opened 2025-12-04
Add releases link to check changes before updating 

hyprhex:
Make it easy for Omarchy users to check the latest changes before updating by clicking the Omarchy releases link


dhh:
This has already been added to dev 👍

PR #3764, opened 2025-12-05
Add branch support for theme variations

joaofelipegalvao:
This adds an optional branch parameter to `omarchy-theme-install`, letting theme creators maintain multiple variations in one repository instead of creating separate repos for each variant.

## Demo

https://github.com/user-attachments/assets/d3f4bcf3-3056-4caf-8107-4d607fb9aeaa

## Usage

Backward compatible - existing usage unchanged:
```bash
omarchy-theme-install https://github.com/user/omarchy-theme.git
```

New optional branch parameter:
```bash
omarchy-theme-install https://github.com/user/omarchy-theme.git dark-variant
```

Themes from branches are installed as `theme-name-branch` to avoid conflicts.

## Why

Creating separate repositories for each theme variation (light/dark/minimal) clutters GitHub profiles and splits maintenance across multiple repos. Using branches is cleaner and follows standard git workflow.

Working example: https://github.com/joaofelipegalvao/omarchy-oasis

### DIFF
--- a/bin/omarchy-theme-install
+++ b/bin/omarchy-theme-install
@@ -1,13 +1,15 @@
 #!/bin/bash
 
 # omarchy-theme-install: Install a new theme from a git repo for Omarchy
-# Usage: omarchy-theme-install <git-repo-url>
+# Usage: omarchy-theme-install <git-repo-url> [branch-name]
 
 if [ -z "$1" ]; then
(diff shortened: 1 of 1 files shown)


dhh:
Interesting idea, but I think it complicates the very simple system we have too much. Think folks are better off just keeping separate repositories.

PR #3772, opened 2025-12-05
Fix keybinding description for makoctl restore

wiehann:
The description for the `makoctl restore` binding is the same as the one used for `makoctl invoke`. This causes a duplicate in the show key bindings menu.

**Before:**
image

**After:**
image


dhh:
This has already been fixed on dev 👍

PR #3776, opened 2025-12-05
Feature/session management

LukasTrust:
I wanted a way to automatically reopen my applications after restarting my PC. So I created a session management system that can:

- Save/restore workspace snapshots (windows, positions, workspaces)
- Quick save/restore with 3 numbered slots (SUPER+ALT+1/2/3)
- Optional persistent mode: auto-save on shutdown, auto-restore on login

Question:
Is a systemd service the right way to hook into shutdown events, or is there a better Hyprland-native way?

Open to feedback and totally understand if this feature doesn't fit Omarchy's vision. Thanks for taking a look!

Edit Discussion:
https://github.com/basecamp/omarchy/discussions/2226

### DIFF
--- a/bin/omarchy-menu
+++ b/bin/omarchy-menu
@@ -483,7 +483,7 @@ show_session_slot_menu() {
 }
 
 show_main_menu() {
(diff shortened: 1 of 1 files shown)


dhh:
Thanks for working on this, but I think this might be better as a separate AUR package. It's a lot of code, and I don't want to be on the hook for maintaining it directly going forward. Appreciate the effort though!

PR #3785, opened 2025-12-06
Bugfix: For kernels different from "linux" Omarchy update always says kernel has changed even if it hasn't

mmsbrggr:
Fix kernel update detection for non-standard kernels

  The kernel update check in omarchy-update-restart was hardcoded to check the linux package, causing false reboot
  prompts for users with alternative kernels like linux-lts, linux-zen, or linux-g14.

  Changes:
  - Dynamically detect the running kernel's package using pacman -Qo on the modules directory
  - Compare pacman versions directly instead of parsing uname -r with sed

  This fixes the issue where custom kernels with version suffixes (e.g., -lts, -g14, etc.) would always trigger a reboot prompt
  due to version string mismatches.

Fixes: https://github.com/basecamp/omarchy/issues/1652


bin/omarchy-update-restart
```diff
@@ -1,6 +1,10 @@
 #!/bin/bash
 
-if [ "$(uname -r | sed 's/-arch/\.arch/')" != "$(pacman -Q linux | awk '{print $2}')" ]; then
+kernel_info=$(pacman -Qo "/usr/lib/modules/$(uname -r)" 2>/dev/null | grep -v headers | head -1)
```
ikegam, line 3: The script fails here if the old modules are gone after the update. I need to dig into how omarchy scripts pass values to each other so I can fix this properly. I haven't found any good precedents yet, which is why I've left my PR in draft.

mmsbrggr: Will the omarch-update ever remove the current linux kernel? Or is this script called in contexts where the currently running kernel has been removed?

mmsbrggr:
@ikegam As I don't think that an omarchy-update will ever remove the active kernel I still think that this is a valid fix and could be merged.
Otherwise, I am open to any suggestions to improve my approach and fix the issue for good.

PR #3798, opened 2025-12-07
Add example to enable three-finger drag in input configuration

daviewales:
Not sure if this needs a migration?

Commented out by default, similar to other examples in the input configuration.


daviewales:
@dhh, note that three finger drag emulates a left-click-drag, so in addition to dragging windows, it will also select text, draw, move drag handles, etc.

dhh:
Updated 👍

PR #3801, opened 2025-12-07
actually querying yay instead of searching through cache

adam-saudagar:
I was trying to install joplin, but i couldn't find it in yay install option (mod+alt+space > install > AUR)

so I looked into why, and it turned out we are actually simple searching through cache, and not actually querying the aur repository

I updated the script to make it work

please let me know if any changes or explanation is needed

### DIFF
--- a/bin/omarchy-pkg-aur-install
+++ b/bin/omarchy-pkg-aur-install
@@ -12,9 +12,10 @@ fzf_args=(
   --bind 'alt-b:change-preview:yay -Gpa {1} | tail -n +5'
   --bind 'alt-B:change-preview:yay -Siia {1}'
   --color 'pointer:green,marker:green'
+  --bind "change:reload(yay -qSs {q})"
 )
 
-pkg_names=$(yay -Slqa | fzf "${fzf_args[@]}")
+pkg_names=$(yay -qSs | fzf "${fzf_args[@]}")
 
 if [[ -n "$pkg_names" ]]; then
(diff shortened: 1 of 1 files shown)


bin/omarchy-pkg-aur-install
```diff
@@ -12,9 +12,10 @@ fzf_args=(
   --bind 'alt-b:change-preview:yay -Gpa {1} | tail -n +5'
   --bind 'alt-B:change-preview:yay -Siia {1}'
   --color 'pointer:green,marker:green'
+  --bind "change:reload(yay -qSs {q})"
 )
 
-pkg_names=$(yay -Slqa | fzf "${fzf_args[@]}")
+pkg_names=$(yay -qSs | fzf "${fzf_args[@]}")
```
dhh, line 18: These two are not the same:

```
omarchy dev ❯ yay -qSs | wc -l
15176

omarchy dev ❯ yay -Slqa | wc -l
102603
```

Need all 102K packages available.

adam-saudagar: I made this change looking this issue

https://github.com/basecamp/omarchy/issues/3592#issuecomment-3571266236

adam-saudagar: okay i replaced my -qSs to -Slqa and now its working fine, before I was getting weird symbols as in the issue linked, and i can find my packages as well through fzf

I'll close the PR, thanks for the review though @dhh

PR #3806, opened 2025-12-07
Add error handling to omarchy-hyprland-window-close-all

luojiyin1987:
- Add set -euo pipefail to fail fast on errors
  - Add xargs -r to suppress 'argument list empty' warning when no windows exist


dhh:
Appreciate the suggestion, but I don't think this is needed. The hotkey that triggers this doesn't output any errors anyway. Fine to just fail as-is, and less code.

PR #3808, opened 2025-12-07
Fix Limine config detection for omarchy manual installation (/boot/EFI/arch-limine)

madhurdhama:
## Summary
During a fresh Omarchy installation on Arch Linux using Limine, the Limine
installer places its configuration under:

    /boot/EFI/arch-limine/limine.conf

The current script only checks /EFI/BOOT/ and /EFI/limine/, so Omarchy manual
installation fails with the error:

    Error: Limine config not found at /boot/EFI/limine/limine.conf

## What this PR changes
- Adds detection for the new archinstall Limine path: /boot/EFI/arch-limine/
- Maintains support for USB EFI path and default EFI path
- Keeps BIOS detection unchanged

### DIFF
--- a/install/login/limine-snapper.sh
+++ b/install/login/limine-snapper.sh
@@ -5,13 +5,20 @@ if command -v limine &>/dev/null; then
 HOOKS=(base udev plymouth keyboard autodetect microcode modconf kms keymap consolefont block encrypt filesystems fsck btrfs-overlayfs)
 EOF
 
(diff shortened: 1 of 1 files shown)


kalinsky:
The issue is reproducible on my system as well. Applying the fix will now lead to properly detecting where the config file is and the installer finishing without any issues. Thank you for the patch!

ryanrhughes:
Implemented this in https://github.com/basecamp/omarchy/commit/64a8d7bb0405f661eb186095c5503ca10695906e and gave you co-author. Thanks for pushing this up!

PR #3814, opened 2025-12-08
feat: add omarchy-update-toggle to enable/disable updates

SoAp9035:
Adds a toggle feature for updates, giving users full control over whether to receive Omarchy and system updates, enhancing user experience and preventing unintended changes.


dhh:
Interesting suggestion, but I don't think we need this. If you don't want an update, just don't run the update? We already have a confirmation now asking you if you're sure you want to run it.

SoAp9035:
My main reason for this feature was to stop Omarchy updates when someone uses `sudo pacman -Syu`. Now, if a user runs this command, Omarchy gets new updates too. I wanted a simple way to turn off Omarchy updates, so people can update other packages without changing their Omarchy version.

@dhh

dhh:
I don't think that's a good idea. New packages might have new incompatibilities, which will require Omarchy changes to accommodate.

PR #3827, opened 2025-12-08
feat: update walker configuration to comment out layerrule in walker.conf, so it uses global hypr config

ronakgh97:
a minor change to the `walker.conf` configuration by commenting out the animation rule for the Walker application, so it uses hypr anim in user config


dhh:
This isn't going to fly. It's there intentionally to prevent the animation as that's the default design.

PR #3830, opened 2025-12-09
Make Mako config user-overridable

horaceko:
Mako (the notification app in Omarchy) uses a config file at `~/.config/mako/config` to configure itself. Currently, this symlinks to the Mako config in the active theme. There are two quirks to this approach:

* If a user edits that file, they are actually editing the theme config
* If a user makes some Mako config edits and changes themes, these changes are not persisted to the new theme

This PR switches things over to mirror the model used by a number of other components in Omarchy, which is to have a separate config file in `~/.config` that overrides the theme configs, and operates independently of the active theme.

This PR addresses #3629.


dhh:
After running this, I get:

```
omarchy horace--mako-user-config ❯ makoctl reload
Unable to parse configuration file (fr.emersion.Mako.InvalidConfig)
```

horaceko:
@dhh fixed; my checked-in config file was somehow missing an `=`.

horaceko:
Closing this PR in favor of a new PR that also adds overrides for Walker.

patrick-boc:
> in favor of a new PR that also adds overrides for Walker.

Which PR is that?

Issue #3066, opened 2025-11-02
Battery Low Warning


himanshu0805:
### What do you need?

For Laptop's with discharging battery.
When ever battery percentage gets low to 10% show a warning message.

I am not getting any warning message when battery falls below 10% and the laptop just shuts down after 5%
it will be great if there is a warning message that remind me to plug my laptop for charging,

harshvsri:
Check for omarchy-battery-monitor daemon, u can see via systemctl --user status.

morgoth:
There is already a notification about 10% battery. Doesn't it work for you?

himanshu0805:
The notification dosen't come when I watch YouTube in Firefox at full screen.

Also there is no sound for it.

morgoth:
The sound feature is proposed here https://github.com/basecamp/omarchy/pull/3009 so you may want to follow on it

mksonkar:
Happened the same to me, took a break from work only to come back and find the laptop shut down due to low battery and lost all my work. A warning notification for low battery would be great.

brutalsam96:
There was a notification service, it is still there but something broke it with a few updates

mksonkar:
> There was a notification service, it is still there but something broke it with a few updates

Would be great if someone fixes it. Sometimes when watching youtube in full screen, laptop just shuts down without warning... happened a couple of times with me.

brutalsam96:
> > There was a notification service, it is still there but something broke it with a few updates
> 
> Would be great if someone fixes it. Sometimes when watching youtube in full screen, laptop just shuts down without warning... happened a couple of times with me.

I have done some digging and found the commit that changed the script that pushes the notification, while i did some tests by moving up the battery threshold i was getting the notification, i will test more when my battery discharges meanwhile you can try to see if you don't have a file created at this path "/run/user/$UID/omarchy_battery_notified", 
if it is there and it is not removed by the script, notification simpy won't run because this file is supposed to be a flag that notification is pushed already, script should remove that file at some point, but it might failing for whatever reason and preventing from pushing a notification.

scale03:
Hi!  I think i found the root cause of this issue, it was the mako config the problem, even with a critical flag there weren't the configuration  needed on mako to allow the notification to be displayed on the top layer when on full screen i just draft a new pr to solve this.

brutalsam96:
[...]

PR #3832, opened 2025-12-09
Fix: Refactored the battery notification script with added feature and fix for inconsistent notifications

brutalsam96:
### Summary

Refactored the omarchy-battery-monitor script to improve reliability of battery notifications.

### Changes

Rewrote battery level fetching and parsing logic, removing dependency on _omarchy-battery-remaining_ call that may have caused inconsistent notifications mentioned in issue #3066 

Updated notification flow to use two levels:

A higher-threshold warning.

A lower-threshold critical alert.

### Outcome

Should ensure battery notifications trigger consistently and notify twice now.

P.S If omarchy-battery-remaining script is not used elsewhere it should be obselete and safe to remove.

### DIFF
--- a/bin/omarchy-battery-monitor
+++ b/bin/omarchy-battery-monitor
@@ -8,19 +8,11 @@ WARN_BATTERY_THRESHOLD=20
 CRIT_NOTIFICATION_FLAG="/run/user/$UID/omarchy_battery_notified_crit"
 WARN_NOTIFICATION_FLAG="/run/user/$UID/omarchy_battery_notified_warn"
 
-BATTERY_PATH=$(upower -e | grep 'BAT' | head -n 1)
+BATTERY_PATH=$(upower -e | grep 'DisplayDevice' | head -n 1)
 BATTERY_INFO=$(upower -i "$BATTERY_PATH")
 
 BATTERY_STATE=$(echo "$BATTERY_INFO" | grep "state" | awk '{print $2}')
 BATTERY_LEVEL=$(echo "$BATTERY_INFO" | grep "percentage" | awk '{print $2}' | tr -d '%')
-#BATTERY_LEVEL=$(echo "$BATTERY_INFO" | grep "percentage" \
-#  | awk -F: '/percentage/ {
-#    gsub(/[%[:space:]]/, "", $2);
-#    val=$2;
-#    printf("%d\n", (val+0.5))
-#    exit
-#    }'
-#  )
 
 send_notification() {
   local level=$1
--- a/bin/omarchy-battery-monitor
+++ b/bin/omarchy-battery-monitor
@@ -2,22 +2,58 @@
 
 # Designed to be run by systemd timer every 30 seconds and alerts if battery is low
 
-BATTERY_THRESHOLD=10
-NOTIFICATION_FLAG="/run/user/$UID/omarchy_battery_notified"
-BATTERY_LEVEL=$(omarchy-battery-remaining)
-BATTERY_STATE=$(upower -i $(upower -e | grep 'BAT') | grep -E "state" | awk '{print $2}')
+CRIT_BATTERY_THRESHOLD=10
+WARN_BATTERY_THRESHOLD=20
+
+CRIT_NOTIFICATION_FLAG="/run/user/$UID/omarchy_battery_notified_crit"
+WARN_NOTIFICATION_FLAG="/run/user/$UID/omarchy_battery_notified_warn"
+
+BATTERY_PATH=$(upower -e | grep 'BAT' | head -n 1)
+BATTERY_INFO=$(upower -i "$BATTERY_PATH")
+
+BATTERY_STATE=$(echo "$BATTERY_INFO" | grep "state" | awk '{print $2}')
+BATTERY_LEVEL=$(echo "$BATTERY_INFO" | grep "percentage" | awk '{print $2}' | tr -d '%')
+#BATTERY_LEVEL=$(echo "$BATTERY_INFO" | grep "percentage" \
+#  | awk -F: '/percentage/ {
+#    gsub(/[%[:space:]]/, "", $2);
+#    val=$2;
+#    printf("%d\n", (val+0.5))
+#    exit
+#    }'
+#  )
 
 send_notification() {
-  notify-send -u critical "󱐋 Time to recharge!" "Battery is down to ${1}%" -i battery-caution -t 30000
+  local level=$1
+  local title=$2
+  local msg=$3
+  local urgency=$4
+
+  notify-send -u "$urgency" -a "System Power" "$title" "$msg" -i battery-caution -t 50000
 }
 
+
 if [[ -n "$BATTERY_LEVEL" && "$BATTERY_LEVEL" =~ ^[0-9]+$ ]]; then
-  if [[ $BATTERY_STATE == "discharging" && $BATTERY_LEVEL -le $BATTERY_THRESHOLD ]]; then
-    if [[ ! -f $NOTIFICATION_FLAG ]]; then
-      send_notification $BATTERY_LEVEL
-      touch $NOTIFICATION_FLAG
+
+  if [[ $BATTERY_STATE == "discharging" ]]; then
+
+    if [[ "$BATTERY_LEVEL" -le "$CRIT_BATTERY_THRESHOLD" ]]; then
+      if [[ ! -f "$CRIT_NOTIFICATION_FLAG" ]]; then
+        send_notification "$BATTERY_LEVEL" " 🔴 Critical Battery" "Battery is critical (${BATTERY_LEVEL}%)" "critical"
+        touch "$CRIT_NOTIFICATION_FLAG"
+        # Also creates WARN flag to prevent duplicate flags
+        touch "$WARN_NOTIFICATION_FLAG"
+      fi
+
+    elif [[ "$BATTERY_LEVEL" -le "$WARN_BATTERY_THRESHOLD" ]] then
+      if [[ ! -f "$WARN_NOTIFICATION_FLAG" ]]; then
+        send_notification "$BATTERY_LEVEL" " 🟠 Low Battery" "Battery is Low (${BATTERY_LEVEL}%)" "critical"
+        touch "$WARN_NOTIFICATION_FLAG"
+      fi
     fi
+  # Remove notified flags when plugged in
   else
-    rm -f $NOTIFICATION_FLAG
+    if [[ -f "$WARN_NOTIFICATION_FLAG" || -f "$CRIT_NOTIFICATION_FLAG" ]]; then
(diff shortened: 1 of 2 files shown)


scale03:
Hi! 👋  
the dual-threshold logic (Warning vs Critical), it's a great UX improvement.                                                                                                                                                                                     I've been profiling this service casually with eBPF tracing tools, and I noticed that the current implementation produce significant overhead and context switch just to notify the battery level warning. Because the script spawns external binaries (upower, grep, awk, find, head) multiple times per cycle, it creates noticeable PID churn and syscall latency every 30 seconds.

I'm also seeing a lot of ENOENT noise in the systemd journal because of race conditions when he try tracking these short-lived processes.
<details> <summary><b>🔍 Click to see the trace log evidence (ENOENT storm)</b></summary>

 trace > [omarchy-battery.log](https://github.com/user-attachments/files/24061166/omarchy-battery.log)
</details>

Proposal: We can achieve the exact same logic with near-zero overhead (<1ms execution) by reading directly from the kernel interface (/sys/class/power_supply/) using pure Bash built-ins, instead of shelling out to upower.

brutalsam96:
> Hi! 👋 the dual-threshold logic (Warning vs Critical), it's a great UX improvement. I've been profiling this service casually with eBPF tracing tools, and I noticed that the current implementation produce significant overhead and context switch just to notify the battery level warning. Because the script spawns external binaries (upower, grep, awk, find, head) multiple times per cycle, it creates noticeable PID churn and syscall latency every 30 seconds.
> 
> I'm also seeing a lot of ENOENT noise in the systemd journal because of race conditions when he try tracking these short-lived processes.
> 
> 🔍 Click to see the trace log evidence (ENOENT storm)
> trace > [omarchy-battery.log](https://github.com/user-attachments/files/24061166/omarchy-battery.log)
> 
> Proposal: We can achieve the exact same logic with near-zero overhead (<1ms execution) by reading directly from the kernel interface (/sys/class/power_supply/) using pure Bash built-ins, instead of shelling out to upower.

Hello there, yes i have been thinking on reading directly from kernel but i choose not to stray too far from the original script just to whip up something fast and test while my battery was discharging to find out why exactly it was working inconsistently(which i couldn't exactly tell why), but i guess you are right it can be a lot cleaner and free of overhead, i might look into it later but you are free to contribute upon this pr i have sent you a collaboration request to my fork.

scale03:
Hi @brutalsam96 — thanks for the invite!

On second thought, instead of bundling the refactor here, I think it's safer to just fix the fullscreen notification issue (https://github.com/basecamp/omarchy/issues/3066) first while maintaining the current logic.   

Actually the problem is not the bash script alone ( notify-send -u critical) this flag is already there but cannot force the visual layer. It can only request "critical" status; the config file decides where "critical" things get drawn. Without the config change, the script fix is useless for fullscreen scenarios.

Im testing out if this could be the final fix for the issue you mentioned, maybe i will make a separate pr for this one, but i would like to collaborate for a refactor later as well, in orther to use sysfs if upower is not prefered istead for this use case.  But im not sure is gonna be  accepted (i already tried it locally and works fine on my thinkpad)

brutalsam96:
@scale03 no problem, i have found that we have status and capacity at the /sys/class/power_supply, this is much easier than parsing full info from upower, but now need to dive deeper and if all drivers provide same status and capacity files also the path /sys/class/power_supply/BAT0, need to also find out if BAT0 is always the name for the battery and maybe handle multiple battery sources.

scale03:
@brutalsam96 Nice, i just created a pr to fix the silent death while on full screen.     #3838                                                                                                                                             completely agree on dropping the upower parsing for direct sysfs reads could be a huge win in the log run.

Since BAT0 isn't universal (and some machines have multiple batteries like BAT0 + BAT1), we probably shouldn't hardcode the path.

I suggest we iterate through /sys/class/power_supply/* and check the type file to confirm it's a "Battery" before reading status and capacity. That way we support multi-battery setups and different driver naming conventions out of the box.

brutalsam96:
@scale03 
Solid work on finding and fixing disappearing notifications.
Exactly my thoughts, assuming BAT as a sole source of information can lead to problems with machines with multiple batteries or desktops with UPS might also be recognized this way.
Luckily for that upower comes in handy because it combines and calculates in such cases so even initial script that targeted bat was off the mark, so correct way to run DisplayDevice through the upower -i and parse that information
```
~ ❯ upower -e
/org/freedesktop/UPower/devices/line_power_AC0
/org/freedesktop/UPower/devices/DisplayDevice
```
Then again this shouldn't do much with overhead and context switch problems, i will research more regarding this matter, until then i will push updated commit that utilizes DisplayDevice instead of BAT.
@dhh @ryanrhughes would love to hear your thoughts regarding this problem as well.

scale03:
I'm not saying that this is a problem,  I'm just saying that for this monitor battery script and maybe something else too the logic is simple enough that could be refactored in bash only. I notice also that systemd is producing every 30 second A LOT of enoet while monitoring the battery monitor script's childs but is just noise not a problem, and the fact that runs every 30 sec to check the battery life could make sense to optimize it (is good for the battery itself) but probably upower is still necessary somehow so maybe is better open a discussion to discuss about this instead of commenting this pr.

Cheers

bin/omarchy-battery-monitor
```diff
@@ -2,22 +2,50 @@
...
+WARN_BATTERY_THRESHOLD=20
+
+CRIT_NOTIFICATION_FLAG="/run/user/$UID/omarchy_battery_notified_crit"
+WARN_NOTIFICATION_FLAG="/run/user/$UID/omarchy_battery_notified_warn"
+
+BATTERY_PATH=$(upower -e | grep 'DisplayDevice' | head -n 1)
+BATTERY_INFO=$(upower -i "$BATTERY_PATH")
+
+BATTERY_STATE=$(echo "$BATTERY_INFO" | grep "state" | awk '{print $2}')
+BATTERY_LEVEL=$(echo "$BATTERY_INFO" | grep "percentage" | awk '{print $2}' | tr -d '%')
```
ryanrhughes: I don't like pulling this logic in here vs relying on `omarchy-battery-remaining`. This leaves us in a position where we're using 2 different methods to calculate for notifications vs the keybind triggered notification.

I'd recommend we update the logic in `omarchy-battery-remaining` and continue utilizing here for consistency.

bin/omarchy-battery-monitor
```diff
@@ -2,22 +2,50 @@
...
+    elif [[ "$BATTERY_LEVEL" -le "$WARN_BATTERY_THRESHOLD" ]] then
+      if [[ ! -f "$WARN_NOTIFICATION_FLAG" ]]; then
+        send_notification "$BATTERY_LEVEL" " 🟠 Low Battery" "Battery is Low (${BATTERY_LEVEL}%)" "critical"
+        touch "$WARN_NOTIFICATION_FLAG"
+      fi
     fi
+  # Remove notified flags when plugged in
   else
-    rm -f $NOTIFICATION_FLAG
+    if [[ -f "$WARN_NOTIFICATION_FLAG" || -f "$CRIT_NOTIFICATION_FLAG" ]]; then
```
ryanrhughes: Can trim these surrounding if statements I think. `rm` will run just fine whether the file is there or not.

bin/omarchy-battery-monitor
```diff
@@ -2,22 +2,50 @@
...
+BATTERY_LEVEL=$(echo "$BATTERY_INFO" | grep "percentage" | awk '{print $2}' | tr -d '%')
 
 send_notification() {
-  notify-send -u critical "󱐋 Time to recharge!" "Battery is down to ${1}%" -i battery-caution -t 30000
+  local level=$1
+  local title=$2
+  local msg=$3
+  local urgency=$4
+
+  notify-send -u "$urgency" -a "System Power" "$title" "$msg" -i battery-caution -t 50000
```
ryanrhughes: Should send the message as `-u critical` always so that the message never disappears. This prevents the issue where it could appear and timeout. 

It looks like currently there's just a long time here but that's really not necessary when we can just rely on just not having a timeout at all.

bin/omarchy-battery-monitor
```diff
@@ -2,22 +2,50 @@
...
+
+BATTERY_PATH=$(upower -e | grep 'DisplayDevice' | head -n 1)
+BATTERY_INFO=$(upower -i "$BATTERY_PATH")
+
+BATTERY_STATE=$(echo "$BATTERY_INFO" | grep "state" | awk '{print $2}')
+BATTERY_LEVEL=$(echo "$BATTERY_INFO" | grep "percentage" | awk '{print $2}' | tr -d '%')
 
 send_notification() {
-  notify-send -u critical "󱐋 Time to recharge!" "Battery is down to ${1}%" -i battery-caution -t 30000
+  local level=$1
```
ryanrhughes: I would just skip this formality. We should only be passing 1-2 things here so we could slam it down to just string interpolated `${1}`, `${2}`, etc.

bin/omarchy-battery-monitor
```diff
@@ -2,22 +2,50 @@
...
-  if [[ $BATTERY_STATE == "discharging" && $BATTERY_LEVEL -le $BATTERY_THRESHOLD ]]; then
-    if [[ ! -f $NOTIFICATION_FLAG ]]; then
-      send_notification $BATTERY_LEVEL
-      touch $NOTIFICATION_FLAG
+
+  if [[ $BATTERY_STATE == "discharging" ]]; then
+
+    if [[ "$BATTERY_LEVEL" -le "$CRIT_BATTERY_THRESHOLD" ]]; then
+      if [[ ! -f "$CRIT_NOTIFICATION_FLAG" ]]; then
+        send_notification "$BATTERY_LEVEL" " 🔴 Critical Battery" "Battery is critical (${BATTERY_LEVEL}%)" "critical"
```
ryanrhughes: Changing how we set these notifications should allow us to simplify this call.

ryanrhughes (changes requested):
Left some feedback for a few small changes. I've only validated this on my laptop but to others' points; do not know what this might do in a situation where you have a UPS or something on a desktop. However, in those cases, perhaps you'd also want the same notice.

bin/omarchy-battery-monitor
brutalsam96: Appreciate the pointers, i have modified the script per your suggestions and also replaced omarchy-batter-remaining with more concise parsing method and most importantly we now get battery level from DisplayDevice instead of BAT0, which should in theory resolve all edge cases with multiple batteries or UPS cases, but still would be great if we could find someone to test these out.

brutalsam96:
Added requested changes. 
Aside from this i was working on different solution than 30 second timer and service that runs script each time.
Basically we got UPower Dbus service that send SIGNAL on PropertiesChange event, meaning it sends battery info each time it's properties change usually when you lose a percentage of battery charge, i am working on a python script that should be able to listen in on Dbus System Sessions and work out battery monitoring logic using dbus-next python library.

dhh:

PR #3840, opened 2025-12-10
Update browser launching to include Exec args

trap0xcc:
When launching the default browser, there is a sed that pulls out the binary's exec path. This drops all arguments, presumably to remove the %U present. This Change updates these scripts to include any args in the Exec line, while still ignoring the %U if present.


trap0xcc:
For some context, this was useful for me as I configured a user local chromium.desktop that has a `--proxy-pac-url` arg specified that I wanted applied to all browser launching in Omarchy.

PR #3842, opened 2025-12-11
enhancement : Enhanced the `os age` feature in fastfetch config

TheAhumMaitra:
The command to print `os age` is good but we can write this more clearly. I hope this helps to make Omarchy good!

Have a nice day!

**One suggestion for core devlopers:**

Please add a `contributing.md`


TheAhumMaitra:
Thank you

PR #3857, opened 2025-12-13
Add better portal share picker

ryanrhughes:


Ziryt:
Sorry for bothering, I tried to add  border-radius: 12px; into css, and doesn't make a difference, and there are other border-radiuses and it seems they also don't apply
@ryanrhughes

ryanrhughes:
@Ziryt there were a couple of other changes in the dev branch before taking this live that resolves that. It sounds like you don't have the config file for the share picker so it's not picking up the styling at all.

If you update to the latest Omarchy it should resolve it or you can 

```
mkdir -p ~/.config/hyprland-preview-share-picker
omarchy-refresh-config hyprland-preview-share-picker/config.yaml
```

Ziryt:
I have new styles applied, I can set border thickness for example, and it will change, but not the border radius for some reason, it just doesn't seem to be applied...

mozabes:
This PR broke my ability to share my screen for anything besides Chromium.  Removing the custom_binary_picker fixed the issue.  Just commenting for anyone else running into a similar issue

PR #3858, opened 2025-12-13
feat: Prevent Snapper snapshots from exhausting disk space

### LINKED ISSUE #3055: /.snapshots takes up all my space
### What do you need?

I found myself running out of disk space frequently (installed omarchy on a 100GB partition, which I assumed would be sufficient). It took me a lot of research to find out that the space was taken by the snapper snapshot under /.snapshots

I managed to free up the space, by deleting all snapshots. They took 50GB, half of all space. 

I wonder if there is a way to limit the space, or to better document this (my `du` skills were insufficient to identify this)

### LINKED ISSUE #3850: Snapshots using up all disk space
### System details

AMD, 250GB SSD

### What's wrong?

My disk space is at 99% and seems to be stuck there, i dont have anything heavy installed and when I list out things that are using disk space I dont get anything heavy.

After doing some investigation I can see there are 5 snapshots and these snapshots are what is using up all my space.

Im not sure if it is okay to remove them.

nacapulque:
## Summary

This change adds minimal, targeted safeguards to Omarchy’s Snapper setup to prevent snapshots—especially on `/home`—from silently consuming large portions of the filesystem.

It preserves Omarchy’s existing snapshot model (pre/post snapshots with Limine boot integration) while making disk usage more predictable and stable.

## Key improvements

- Prevents Snapper snapshots from consuming excessive disk space
- Makes snapshot cleanup proactive and predictable instead of reactive
- Reduces I/O overhead on `/home` by disabling unnecessary background comparisons
- Improves space accounting accuracy via Btrfs qgroups
- Preserves Omarchy’s snapshot and rollback philosophy (no timeline snapshots, no behavior change)

## Rationale

Omarchy intentionally disables Snapper timeline snapshots in favor of transactional, boot-safe recovery.  
However, Snapper’s default space governance can still allow snapshots to grow until disk pressure becomes user-visible.

This change introduces space and cleanup guardrails without altering when or how snapshots are created, aligning Snapper’s defaults with Omarchy’s desktop/workstation use case.

fixes #3850 
fixes #3055

### DIFF
--- a/install/login/limine-snapper.sh
+++ b/install/login/limine-snapper.sh
@@ -75,7 +75,6 @@ term_palette_bright: 414868;f7768e;9ece6a;e0af68;7aa2f7;bb9af7;7dcfff;c0caf5
 term_foreground: c0caf5
 term_foreground_bright: c0caf5
 term_background_bright: 24283b
-
 EOF
 
   # Remove the original config file if it's not /boot/limine.conf
@@ -83,6 +82,10 @@ EOF
     sudo rm "$limine_config"
   fi
 
+  # Enable Btrfs qgroups for accurate Snapper space accounting
+  if command -v btrfs &>/dev/null; then
+    sudo btrfs quota enable / 2>/dev/null || true
+  fi
 
   # Match Snapper configs if not installing from the ISO
   if [[ -z ${OMARCHY_CHROOT_INSTALL:-} ]]; then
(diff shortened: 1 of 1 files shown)


ryanrhughes:
I updated a little bit about this implementation in https://github.com/basecamp/omarchy/commit/9b78e3c571e19b99a6c166dd20ec5feb8a3f2e71 and gave you co-author credit. Thank you for pushing this up!

T1t4n25:
thanks for your hard work you both I as an omarchy user appreciate it.

However I think this problem appeared again as I had to remove 320GB of snapshots yesterday from my 512GB nvme to save space, I was updated and using omarchy just this year for 2-3 months, so it should be a new behaviour I think, I had it set to 
this might be the problem from my prespective
```
~ ✗ grep -E "TIMELINE_|NUMBER_" /etc/snapper/configs/root
NUMBER_CLEANUP="yes"
NUMBER_MIN_AGE="3600"
NUMBER_LIMIT="50"
NUMBER_LIMIT_IMPORTANT="10"
TIMELINE_CREATE="yes"
TIMELINE_CLEANUP="yes"
TIMELINE_MIN_AGE="3600"
TIMELINE_LIMIT_HOURLY="10"
TIMELINE_LIMIT_DAILY="10"
TIMELINE_LIMIT_WEEKLY="0"
TIMELINE_LIMIT_MONTHLY="10"
TIMELINE_LIMIT_QUARTERLY="0"
TIMELINE_LIMIT_YEARLY="10"
```

I changed them to a lower number limit 
```
NUMBER_CLEANUP="yes"
NUMBER_MIN_AGE="3600"
NUMBER_LIMIT="5"
NUMBER_LIMIT_IMPORTANT="5"
TIMELINE_CREATE="yes"
TIMELINE_CLEANUP="yes"
TIMELINE_MIN_AGE="3600"
TIMELINE_LIMIT_HOURLY="0"
TIMELINE_LIMIT_DAILY="5"
TIMELINE_LIMIT_WEEKLY="0"
TIMELINE_LIMIT_MONTHLY="0"
TIMELINE_LIMIT_QUARTERLY="0"
TIMELINE_LIMIT_YEARLY="0"
```

I'll be staying with this for a while tell me if you needed any updated or if you can give me tips on managing them in a more friendly way.

Thanks for your hard work I really appreciate it!

PR #3860, opened 2025-12-13
enforce clearer snapshot descriptions + tighter checks in omarchy-snapshot

Dieterbe:
main advantage is that now in your boot menu, instead of just seeing a bunch of version numbers, you actually see much more descriptive entries that make it clearer when it was taken and why.

also, omarchy-snapshot used to quietly exit when you gave it a bogus command. that's fixed now too.


Dieterbe:
Hi @ryanrhughes what's the verdict? Thanks

PR #3869, opened 2025-12-14
feat: add omarchy-launch-or-focus-special-workspace

CantC0unt:
### Why 
---

There are some apps that I like to open and forget like notes/todo or music apps but I still like to be able to quickly access it to change volume or skip songs or add/toggle a todo etc. which is not that quick if they are clustered with all my other apps on some random workspace. adding a windowrule helps but not all that much + I have to do it for every other app.

### What 
---

I added a <sup>*</sup>new script `omarchy-launch-or-focus-special-workspace` which does the following:

| Condition | Action |
|--------|--------|
| App is not open | Shift focus to special workspace and launch the app |
| App is open on workspace `x` | Silently move the app to special workspace and focus the app |
| App is open on special workspace | toggle the special workspace |
| Launch command is not given | uses `"uwsm-app -- $WINDOW_PATTERN"` as the launch command, same as `omarchy-launch-or-focus` | |

\* modified from `omarchy-launch-or-focus`

### Usage 
---

example for keybinds: 
```sh
bindd = SUPER, M, Music, exec, omarchy-launch-or-focus-special-workspace "YouTube Music" music omarchy-launch-webapp "https://music.youtube.com/"
```


Bucurenciu-Cristian:
This is very awesome. Thanks

pomartel:
Yep, I just tried it with Spotify and it's super neat!

```ini
bindd = SUPER SHIFT, M, Music (Spotify), exec, omarchy-launch-or-focus-special-workspace spotify music
```

pomartel:
Another way to make this work (if this PR is not merged) :
```ini
$switchoffspecialworkspace = hyprctl monitors -j | jq -r '.[] | select(.focused == true) | .specialWorkspace.name' | sed 's/special://' | xargs -I [] hyprctl dispatch togglespecialworkspace []
bind = SUPER SHIFT, Q, exec, $switchoffspecialworkspace 

windowrule = workspace special:todos,class:(todoist)
windowrule = workspace special:music,class:(spotify)
```
You define the special workspaces and associated apps in  window rules. Then you have a single key binding (SUPER SHIFT Q in this example) to hide the active special workspace.

CantC0unt:
> Another way to make this work (if this PR is not merged) :
> ```ini
> $switchoffspecialworkspace = hyprctl monitors -j | jq -r '.[] | select(.focused == true) | .specialWorkspace.name' | sed 's/special://' | xargs -I [] hyprctl dispatch togglespecialworkspace []
> bind = SUPER SHIFT, Q, exec, $switchoffspecialworkspace 
> 
> windowrule = workspace special:todos,class:(todoist)
> windowrule = workspace special:music,class:(spotify)
> ```
> You define the special workspaces and associated apps in  window rules. Then you have a single key binding (SUPER SHIFT Q in this example) to hide the active special workspace.

I considered this (similar) approach when I was making the script. 
Doing it this way, you need to add a windowrule for all the spaces you need, plus it takes away the ability to open it normally, ie. not on a special workspace (I believe you can still open and then move but it's an extra step)

ddVital:
What is the difference between this and scratchpad?

CantC0unt:
> What is the difference between this and scratchpad?

It's a scratchpad for specific apps (or app groups). 

At least in my case, I had too many things open on the scratchpad and that made it really cluttered and didn't leave me with enough room for all things I wanted. This way I can use the said apps full screen* if I want to, with a way to quick access it like I would with a scratchpad.

dcunited001:
The way I handle this is similar to the above, but if i want an app to launch while the workspace is empty, then i use `on-created-empty`.

```ini
$wsDocs = docs
$wsDocsKey = d
# $wsDocsMon = $m2port # can't quite set this to a specific monitor
workspace = special:$wsDocs, on-created-empty:[float] zeal
bind = SUPER, $wsDocsKey, togglespecialworkspace, $wsDocs
bind = SUPER ALT, $wsDocksKey, movetoworkspacesilent, special:$wsDocs
windowrule = workspace special:$wsDocs, class:(org.zealdocs.Zeal)
windowrule = float, class:(org.zealdocs.Zeal)
windowrule = size 1920 1080, class:(org.zealdocs.Zeal)
```

I have a directory `~/.config/hypr/workspaces` where the workspaces are defined.
If i have trouble with them on one computer, then I don't include that from the
main `~/.config/hypr/hyprland.conf`. I have special workspaces for things like:

- Sound (one for music and one for audio)
- Docs (one each for man, info and zeal/devdocs)
- For Dotfiles editor windows (one each for hyprland, nixos, emacs, org-agenda)

If you look at the `props` on [Workspace Rules](https://wiki.hypr.land/0.52.0/Configuring/Workspace-Rules/), you can specify some additional logic there, though it's a bit messy.

Some of the following is apropos to the PR/Issue. What I can't quite get, at least not without `hyprctl` scripting:

- associating special workspaces to a specific monitor, though I don't want this most of the time.
- jumping to a special workspace that's already open on the monitor (without pulling it to the current monitor)
- (From a special workspace with floating windows), using `SUPER+[left/right]` to jump to another workspace's windows & back will cause the special workspace to disappear. If it's tiled, the special workspace gets focused.
- Jumping `SUPER+[left/right]` between workspaces where one has mixed floating windows (especially where *one side of one monitor* connects to *two other monitors*). It's complex and "non-deterministic". 

If you only have two monitors, it simplifies things with standard dispatchers (e.g. `focusmonitor +1` with two monitors always alternates)

The special workspaces are a bit limited, since they're only supported on `movetoworkspace` and `movetoworkspacesilent`. Usually this means using `hyprctl`.

For my use case, maybe some of these should just be standard named workspaces. I
may be overusing these special workspaces. What are the main differences from a
UI/UX perspective? e.g. I can't really click through to interact with the lower
workspace when a special workspace is active.

CantC0unt:
> The way I handle this is similar to the above, but if i want an app to launch while the workspace is empty, then i use `on-created-empty`.

my changes don't "force" empty workspace, you can have multiple apps on one special workspace if you want. The idea was segregating things I would want to access quickly into one workspace (mostly related).

using windowrule for this feels overly restrictive to me specially because you might not always want to open the said app on a special workspace (which is not possible if you set a windowrule).

> Some of the following is apropos to the PR/Issue. What I can't quite get, at least not without `hyprctl` scripting:
> 
> * associating special workspaces to a specific monitor, though I don't want this most of the time.
> * jumping to a special workspace that's already open on the monitor (without pulling it to the current monitor)
> * (From a special workspace with floating windows), using `SUPER+[left/right]` to jump to another workspace's windows & back will cause the special workspace to disappear. If it's tiled, the special workspace gets focused.
> * Jumping `SUPER+[left/right]` between workspaces where one has mixed floating windows (especially where _one side of one monitor_ connects to _two other monitors_). It's complex and "non-deterministic".

for 1 and 2, I don't understand, why do you want to do this on a special workspace, normal workspaces work the way you seem to want/need. (If I'm not wrong this is possible with special workspaces as well, but imo, defeats the purpose of special workspaces)
3 can be done and there might be some use, but again, it kind of defeats the purpose of special workspaces, if thats what you want, its better to use normal workspaces, special workspaces are meant to be toggled off once you're done.
4 is out of scope for this PR.

> If you only have two monitors, it simplifies things with standard dispatchers (e.g. `focusmonitor +1` with two monitors always alternates)

also unrelated and out of scope for this PR, but you can check out kanshi for multiple monitor dynamic setups.

> The special workspaces are a bit limited, since they're only supported on `movetoworkspace` and `movetoworkspacesilent`. Usually this means using `hyprctl`.

that is the whole point of special workspaces - https://wiki.hypr.land/Configuring/Dispatchers/#special-workspace

> For my use case, maybe some of these should just be standard named workspaces. I may be overusing these special workspaces.

yes, these should just be standard named workspaces if you don't want them to be toggled on and off.

>  What are the main differences from a UI/UX perspective? e.g. I can't really click through to interact with the lower workspace when a special workspace is active.

hyprland defines it as this 
> A special workspace is what is called a “scratchpad” in some other places. A workspace that you can toggle on/off on any monitor.


The way I see it, it's meant to be toggled on to do something quick and then get back to doing whatever you were doing before. eg - quickly looking at docs or reference material while you're building something or changing the running song or volume (assuming you don't have physical controls on your listening device) etc.

pomartel:
@dhh you might also want to consider my take on special workspace which is a bit simpler and works just as well: https://sudomarchy.com/posts/hyprland-workspaces-are-special#2-script-it

PR #3875, opened 2025-12-15
Use Shift+Delete for cutting.

kratusch:
Shift+Delete is more universal. Look: https://en.wikipedia.org/wiki/Table_of_keyboard_shortcuts.


kratusch:
The only problem i faced with shift+delete is with a file manager, which is for permanently deleting. Closing as this could be a problem for many users.

PR #3876, opened 2025-12-15
Replace gum spinning globe with nerdfont checkmark

jbnunn:
The animated globe icon in omarchy-show-done (seen after actions like installing or removing packages from the Omarchy menu) looks out of place with the rest of the clean aesthetic across Omarchy. This replaces the animation with a nerdfont checkmark. Before and after below:

gum spin --spinner "globe"

show-done-with-nerdfont

### DIFF
--- a/bin/omarchy-show-done
+++ b/bin/omarchy-show-done
@@ -1,4 +1,5 @@
 #!/bin/bash
 
 echo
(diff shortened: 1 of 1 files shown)


dhh:
I agree that the globe isn't ideal, but I prefer to have something that's spinning, which calls your eye to the line. But maybe we could do that in another way? Open to ideas, but the checkbox alone isn't it.

PR #3880, opened 2025-12-15
feat: Add China region connectivity support

n1203:
- Add GitHub repository cloning mirror support in boot.sh
- Add China-friendly keyserver fallbacks in install/preflight/pacman.sh
- Add China-friendly keyserver configuration in default/gpg/dirmngr.conf
- Add Arch Linux package mirror list for China (mirrorlist-china)
- Create INSTALL_CN.md installation documentation for Chinese users
- Update README.md with reference to China installation guide
- Update omarchy-refresh-pacman to support china option

### DIFF
--- a/INSTALL_CN.md
+++ b/INSTALL_CN.md
@@ -1,0 +1,109 @@
+# Omarchy 中国地区安装指南
+
+本文档提供了针对中国大陆地区用户优化的安装方法，以解决网络连接问题。
+
+## 快速安装（使用中国镜像）
+
+如果您在中国大陆，可以使用以下命令进行安装，这将自动使用中国友好的镜像源：
+
+```bash
+curl -o- https://omarchy.org/install | OMARCHY_USE_CHINA_MIRRORS=1 bash
+```
+
+## 环境变量选项
+
+### OMARCHY_USE_CHINA_MIRRORS
+
+设置为 `1` 启用中国镜像源（清华、中科大、阿里云等）：
+
+```bash
+export OMARCHY_USE_CHINA_MIRRORS=1
+```
+
+### OMARCHY_GIT_MIRROR
+
+指定 Git 仓库镜像站点（如果 GitHub 访问困难）：
+
+```bash
+# 使用 Gitee 镜像（需要先在 Gitee 上创建镜像）
+export OMARCHY_GIT_MIRROR=gitee.com
+
+# 使用 GitCode 镜像（需要先在 GitCode 上创建镜像）
+export OMARCHY_GIT_MIRROR=gitcode.com
+```
+
+## 手动切换到中国镜像
+
+如果您已经安装了 Omarchy，可以随时切换到中国镜像：
+
+```bash
+omarchy-refresh-pacman china
+```
+
+切换回默认镜像：
+
+```bash
+omarchy-refresh-pacman stable
+```
+
+## 支持的中国镜像源
+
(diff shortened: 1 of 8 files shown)


jamessimo:
Just want to +1 this.

AndrewMast:
Does this PR simply provide better connectivity or does it actually allow Omarchy to work in this region? I don't love the idea of adding region-specific optimization (especially if it only makes things faster), feels like adding one will require us to add other regions too.

If this PR doesn't make it through, maybe the next step would be adding support to use a custom (3rd party source) set of mirrors and keyservers, maybe stored in a JSON file that can be kept up-to-date outside of the main repo.

Sen:
+1

n1203:
> Does this PR simply provide better connectivity or does it actually allow Omarchy to work in this region? I don't love the idea of adding region-specific optimization (especially if it only makes things faster), feels like adding one will require us to add other regions too.

Hi Andrew — this PR is about making installs actually work in mainland China, not just faster. The mirrors and keyserver fallbacks move pacman/GPG from frequent failure to reliable completion on common ISPs.

To avoid region creep, I can refactor it to a generic, opt‑in mechanism that loads mirrors/keyservers from an external JSON manifest. Default stays global; CN (and future regions) live outside the main repo. Happy to adjust the PR accordingly.

n1203:
> > Does this PR simply provide better connectivity or does it actually allow Omarchy to work in this region? I don't love the idea of adding region-specific optimization (especially if it only makes things faster), feels like adding one will require us to add other regions too.
> 
> Hi Andrew — this PR is about making installs actually work in mainland China, not just faster. The mirrors and keyserver fallbacks move pacman/GPG from frequent failure to reliable completion on common ISPs.
> 
> To avoid region creep, I can refactor it to a generic, opt‑in mechanism that loads mirrors/keyservers from an external JSON manifest. Default stays global; CN (and future regions) live outside the main repo. Happy to adjust the PR accordingly.

Also, my first run with the install script failed mid‑way in China—pretty frustrating—this PR addresses that failure mode.

AndrewMast:
> To avoid region creep, I can refactor it to a generic, opt‑in mechanism that loads mirrors/keyservers from an external JSON manifest. Default stays global; CN (and future regions) live outside the main repo. Happy to adjust the PR accordingly.

I think adjusting the PR to do this would be great. Maybe once the mechanism is implemented, you can provide the China mirrors/keyservers in a gist or in the description of the PR, but not in the actual code in the repo. Maybe we can add it to the documentation once the PR is merged.

n1203:
> > To avoid region creep, I can refactor it to a generic, opt‑in mechanism that loads mirrors/keyservers from an external JSON manifest. Default stays global; CN (and future regions) live outside the main repo. Happy to adjust the PR accordingly.
> 
> I think adjusting the PR to do this would be great. Maybe once the mechanism is implemented, you can provide the China mirrors/keyservers in a gist or in the description of the PR, but not in the actual code in the repo. Maybe we can add it to the documentation once the PR is merged.
> 

Sounds like a plan. I'll refactor the PR to implement the generic opt-in mechanism and ensure the specific CN mirrors are kept out of the source code. I'll provide the example config in the PR description or a Gist once the changes are pushed.

dhh:
I don't see how this can work given that Omarchy depends not just on a controlled mirror, but also on its own package repository. Both are hosted on Cloudflare. So if Cloudflare is not available, none of this is going to fly.

PR #3881, opened 2025-12-15
fix: Add confirmation prompt for empty password in Windows VM setup

n1203:
When user presses Enter without entering a password, prompt with confirmation dialog to either use default password or retry entering a custom password. This prevents accidental use of default password when user expects to enter password input mode.

- Wrap password input in validation loop with explicit confirmation
- On decline, re-prompt for password entry instead of proceeding
- Update placeholder text to clarify behavior

### DIFF
--- a/bin/omarchy-windows-vm
+++ b/bin/omarchy-windows-vm
@@ -140,13 +140,25 @@ EOF
     USERNAME="docker"
   fi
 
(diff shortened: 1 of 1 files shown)


dhh:
Hmm, appreciate the suggestion, but feel like we're pretty well covered with "Password (Press enter to use default: admin)" ✌️

PR #3658, opened 2025-11-27
Move between workspace windows by using vim motions

hyprhex:
Save few second by moving fast with your keyboards lazy boy.

PR #3888, opened 2025-12-16
Reduce keyboard fatigue for move window focus

huangeddie:
By adding VIM direction keys as another way to move window focus.

Moving focus windows is a common action. I believe some devs (myself included) will find it more comfortable to navigate with the VIM keys since their hands are already at that position rather than having to constantly shift their hands down and to the right to reach the arrow keys.


huangeddie:
Also verified those shortcuts don't seem to be used anywhere else (checked with SUPER + K shortcut search)

dharmavagabond:
FYI: #3658

PR #3909, opened 2025-12-16
feat(bidings): Add BlueTUI and Wiremix binding to bindings.conf

ddVital:
This pull request adds a new keyboard binding to the Hyprland configuration, allowing quick access to the BlueTUI application.

New key binding:

* Added a binding for `SUPER + B` to launch `BlueTUI` using the `omarchy-launch-tui` script.


brink-lab:
It should be SUPER + SHIFT + letter to follow the current conventions for bindings for launching apps. e.g. BTOP (also a TUI) is on SUPER + SHIFT + T.

ddVital:
> It should be SUPER + SHIFT + letter to follow the current conventions for bindings for launching apps. e.g. BTOP (also a TUI) is on SUPER + SHIFT + T.

Thanks for the feedback. I went with SUPER + B because SUPER + SHIFT + B is already taken by Chromium, but I’m fine changing it. SUPER + SHIFT + U could work (blUetui), or SUPER + SHIFT + R since Bluetooth is a radio.

I’m also thinking about adding SUPER + SHIFT + S for Wiremix, let me know if that makes sense.

dhh:
Added these on Super + Ctrl + A/B to fit with the rest of the system control commands 👌

brink-lab:
> Added these on Super + Ctrl + A/B to fit with the rest of the system control commands 👌

Super + ctrl + W for Impala (wifi) while we're at it? :) 
Then hotkeys for all waybar icons are sorted in one go. Unless it already has a hotkey ofc.

PR #3916, opened 2025-12-17
feat: add support for SSH git URLs when installing themes

Vargrul:
### What this does

Adds support for installing Omarchy themes from SSH (scp-style) Git repository URLs and clarifies the accepted URL formats in the prompt.

### Why

Previously, the theme installer only worked correctly with HTTPS Git URLs. Providing an SSH URL such as `git@github.com:org/omarchy-foo-theme.git` resulted in an incorrect theme name and install path.

Supporting SSH URLs makes it easier for contributors and users who rely on SSH-based Git workflows.

### Changes

- Adds support for scp-style SSH Git URLs (`git@host:org/repo.git`) when deriving the theme name.
- Normalizes SSH URLs before calling `basename` to ensure correct theme directory naming.
- Updates the `gum input` placeholder to explicitly document supported Git URL formats.
- Keeps existing theme naming rules intact (`omarchy-` prefix and `-theme` suffix removal).

### Examples

- `https://github.com/org/omarchy-foo-theme.git` → `~/.config/omarchy/themes/foo`
- `git@github.com:org/omarchy-foo-theme.git` → `~/.config/omarchy/themes/foo`

### Notes

- No behavior changes for existing HTTPS-based installs.
- This change only affects how the repository name is parsed and does not modify cloning behavior.

### DIFF
--- a/bin/omarchy-theme-install
+++ b/bin/omarchy-theme-install
@@ -5,7 +5,7 @@
 
 if [ -z "$1" ]; then
   echo -e "\e[32mSee https://manuals.omamix.org/2/the-omarchy-manual/90/extra-themes\n\e[0m"
-  REPO_URL=$(gum input --placeholder="Git repo URL for theme" --header="")
+  REPO_URL=$(gum input --placeholder="Git repo URL (https or git@host:org/repo.git)" --header="")
 else
   REPO_URL="$1"
 fi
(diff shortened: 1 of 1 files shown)


bin/omarchy-theme-install
```diff
@@ -15,7 +15,20 @@ if [ -z "$REPO_URL" ]; then
...
+REPO_PATH="$REPO_URL"
+if [[ "$REPO_PATH" != *"://"* && "$REPO_PATH" == *:*/* ]]; then
+  REPO_PATH="${REPO_PATH#*:}"
+fi
+REPO_PATH="${REPO_PATH%/}"
+
+REPO_BASENAME=$(basename "$REPO_PATH")
+REPO_BASENAME="${REPO_BASENAME%.git}"
+
+THEME_NAME=$(
```
neheb, line 29: shorter to use <<< instead of echo.

PR #3920, opened 2025-12-17
Add CTRL ALT SHIFT DEL to close all windows except current workspace

chriopter:
As addition to existing CTRL ALT DEL, i like to clean up everything except the current workspace / what im looking at currently. 

CTRL ALT DEL SHIFT closes everything except current workspace and moves it to workspace 1.


dhh:
Interesting idea! That wouldn't personally be something I'd use, so probably best as a separate extension. But love how easy it is to add these things. Thanks for suggesting.

PR #3929, opened 2025-12-18
Make `gitk` and `git gui` commands work

### LINKED ISSUE #3927: `gitk` and `git gui` commands fail due to missing dependency
### System details

Omarchy 3.2.3

### Issue In Brief

The `gitk` and `git gui` commands fail due to a missing dependancy; the package `tk`.

On a fresh install...

Image

### Why

Looks like `tk` is an optional dependency of `git` in arch, probably because it doesn't make much sense in a headless install.

Image

nagi:
The `gitk` & `git gui` commands fail on a fresh install due to `tk` not being installed.

gitgui

The `tk` package is flagged as an optional dependency of `git` in Arch's package manager, but without it the `gitk` command falls over.

I suspect it's optional by default as `gitk` makes no sense in a headless install, but it's really handy in headmore system.

This adds 26MB to the installation.

Fixes #3927

### DIFF
--- a/install/omarchy-base.packages
+++ b/install/omarchy-base.packages
@@ -114,6 +114,7 @@ sushi
 swaybg
 swayosd
 system-config-printer
+tk
 tldr
 tree-sitter-cli
 tobi-try


dhh:
The Omarchy recommendation for a gui for git is LazyGit. Git gui or gitk wouldn't be something I'd personally use, so adding another 26MB seems too much. Appreciate the suggestion, though!

PR #3936, opened 2025-12-18
Feat: Add hyprmon-bin - Closes #2858

### LINKED ISSUE #2858: Add a better display configuration tool
### What do you need?

There are a few options available to us to provide a better experience for users to configure their monitors. 

I think something like nwg-display is a little to heavy and doesn't fit the vibe of Omarchy but some candidates I do think could be fit are:
- [hyprdynamicmonitors](https://github.com/fiffeek/hyprdynamicmonitors)
- [display-tui](https://github.com/otto-bus-dev/display-tui)
- [hyprmon](https://github.com/erans/hyprmon)

It'd be great if people could explore these options and propose a solution for this that keeps things simple, still allows use of config files, and fits the overall Omarchy theme. It should also be able to accomplish this without needing a whole lot of rigging to support.

I don't utilize multiple displays, and I know @dhh doesn't either so it will likely require someone to pick up the torch on this one who does so we can ensure it's properly tested but also just have the motivation to tackle it.

shmall03:
Fixes #2858

### DIFF
--- a/bin/omarchy-reinstall
+++ b/bin/omarchy-reinstall
@@ -18,6 +18,7 @@ if gum confirm "Are you sure you want to reinstall and lose all config changes?"
   echo "Reinstalling missing Omarchy packages"
   mapfile -t packages < <(grep -v '^#' "$OMARCHY_PATH/install/omarchy-base.packages" | grep -v '^$')
   sudo pacman -Syu --noconfirm --needed "${packages[@]}"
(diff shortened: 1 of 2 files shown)


shmall03 (commented):
Added new lines at the ends on new files

dhh:
Appreciate the suggestion, but I actually prefer just to have this done through the config files. And then we can point folks who just don't want that to hyprmon as a separate option.

PR #3955, opened 2025-12-21
fix(aur): add --cleanafter to prevent omarchy-update failures on corrupted cache, and make AUR pkg updates idempotent

### LINKED ISSUE #3917: omarchy-update fails at Searching AUR for updates when folder already exists(?)
### System details

during 3.2.4 update, 13620H RTX4050 laptop 

### What's wrong?

Updating these manually after deleting the repos would work, but its unclear why i need to do that, am i doing something wrong?

```
Running as unit: run-p238361-i238362.service; invocation ID: 7a16142e075146ae80fa049fd3028406

Update AUR packages
:: Searching AUR for updates...
 -> Orphan (unmaintained) AUR Packages: activitywatch
:: 2 packages to upgrade/install.
2  aur/google-chrome     143.0.7499.109-1 -> 143.0.7499.146-1
1  aur/google-cloud-cli  549.0.1-1        -> 550.0.0-1
==> Packages to exclude: (eg: "1 2 3", "1-3", "^4" or repo name)
 -> Excluding packages may cause partial upgrades and break systems
==>
AUR Explicit (2): google-chrome-143.0.7499.146-1, google-cloud-cli-550.0.0-1
:: (0/2) Failed to download PKGBUILD: google-cloud-cli
:: (1/2) Downloaded PKGBUILD: google-chrome
 -> error fetching google-cloud-cli: fatal: destination path 'google-cloud-cli' already exists and is not an empty directory.
         context: exit status 128


Please review the output above carefully, correct the error, and retry the update.

If you need assistance, get help from the community at https://omarchy.org/discord

🌎 Done! Press any key to close...
```

scale03:
### WHY
This PR addresses a recurring failure during the `omarchy-update` process where `yay` crashes the system update.
Fixes #3917
### The Problem
When updating AUR packages, `yay` attempts to reuse existing git repositories in `~/.cache/yay/`. If a previous download was interrupted or the directory is corrupted, `git` fails because the "destination path already exists and is not an empty directory." 

Because the script uses the `--noconfirm` flag, `yay` cannot prompt the user to clean the directory and instead exits with a fatal error, having the user to manually remove the path.

### The Fix
Added the `--cleanafter` flag to the yay command. This ensures that:
1. Build directories are removed immediately after a successful system update.
2. The system state remains clean for future update cycles.
3. The update process becomes **idempotent**, preventing stale state conflicts.

### DIFF
--- a/bin/omarchy-update-system-pkgs
+++ b/bin/omarchy-update-system-pkgs
@@ -3,7 +3,7 @@
 set -e
 
 echo -e "\e[32m\nUpdate system packages\e[0m"
-sudo pacman -Syyu --noconfirm
+sudo pacman -Syyu --noconfirm --cleanafter
 
 # Update AUR packages if any are installed
 if pacman -Qem >/dev/null; then
--- a/bin/omarchy-update-system-pkgs
+++ b/bin/omarchy-update-system-pkgs
@@ -3,13 +3,13 @@
 set -e
 
 echo -e "\e[32m\nUpdate system packages\e[0m"
-sudo pacman -Syyu --noconfirm --cleanafter
+sudo pacman -Syyu --noconfirm 
 
 # Update AUR packages if any are installed
 if pacman -Qem >/dev/null; then
   if omarchy-pkg-aur-accessible; then
     echo -e "\e[32m\nUpdate AUR packages\e[0m"
-    yay -Sua --noconfirm
+    yay -Sua --noconfirm --cleanafter
     echo
   else
     echo -e "\e[31m\nAUR is unavailable (so skipping updates)\e[0m"


dhh:
I don't follow fully here. You write:

```
When updating AUR packages, yay attempts to reuse existing git repositories in ~/.cache/yay/. If a previous download was interrupted or the directory is corrupted, git fails because the "destination path already exists and is not an empty directory."
```

But the change you've made is to the pacman step, not to the yay step? AUR is only processed with the yay step, no?

scale03:
@dhh  I mistakenly applied the flag to the pacman command instead of yay.  I've just updated the PR to move the --cleanafter flag to the correct step, which is where the git cache conflicts occur.  Thanks for pointing it out!

dhh:
Merged via 88784781036137ccb116e962ea9bedafee6574fd 👌

Issue #3954, opened 2025-12-21
NVIDIA hardware install script installs unsupported driver on Pascal and older GPUs


ashddev:
### System details

GPU: NVIDIA GeForce GTX 1050 (affects GTX 10xx, 9xx, and older), Omarchy version: 3.2.3

### What's wrong?

As of NVIDIA driver version 590, `nvidia-dkms` no longer supports Pascal or older GPUs ([Arch Linux announcement](https://archlinux.org/news/nvidia-590-driver-drops-pascal-support-main-packages-switch-to-open-kernel-modules/)).

Omarchy's NVIDIA install script currently selects `nvidia-dkms` for Pascal GPUs, which results in an unsupported driver and a broken graphical environment after upgrade.

The script should instead install the legacy proprietary branch `nvidia-580xx-dkms` ([AUR](https://aur.archlinux.org/packages/nvidia-580xx-dkms)) for these GPUs.

I’m happy to submit a patch for this issue.

dhh:
Closing as fixed, resolved, or superseded by the current Quattro implementation based on the open-issue review.

PR #3956, opened 2025-12-21
fix(nvidia): install legacy NVIDIA driver on older GPUs

ashddev:
Fixes [#3954](https://github.com/basecamp/omarchy/issues/3954)

This PR updates the NVIDIA hardware install script to install the legacy NVIDIA driver for older GPUs (Pascal (10xx), Maxwell (9xx), Kepler (7xx)), as suggested in the recent [Arch Linux announcement](https://archlinux.org/news/nvidia-590-driver-drops-pascal-support-main-packages-switch-to-open-kernel-modules/).

### DIFF
--- a/install/config/hardware/nvidia.sh
+++ b/install/config/hardware/nvidia.sh
@@ -11,7 +11,11 @@ if [ -n "$NVIDIA" ]; then
     # Pascal (10xx) and Maxwell (9xx) use legacy branch that can only be installed from AUR
     PACKAGES=(nvidia-580xx-dkms nvidia-580xx-utils lib32-nvidia-580xx-utils)
   fi
-  # < GTX 9, See: https://wiki.archlinux.org/title/NVIDIA
+  # Bail if no supported GPU
+  if [ -z "${PACKAGES+x}" ]; then
+    echo "No compatible driver for your NVIDIA GPU. See: https://wiki.archlinux.org/title/NVIDIA"
+    exit 0
+  fi
 
   pacman -S --needed --noconfirm "$KERNEL_HEADERS" "${PACKAGES[@]}"
 


install/config/hardware/nvidia.sh
```diff
@@ -31,19 +32,29 @@ if [ -n "$(lspci | grep -i 'nvidia')" ]; then
...
-    "lib32-nvidia-utils"
     "egl-wayland"
-    "libva-nvidia-driver" # For VA-API hardware acceleration
     "qt5-wayland"
     "qt6-wayland"
   )
 
-  sudo pacman -S --needed --noconfirm "${PACKAGES_TO_INSTALL[@]}"
+  # Turing (16xx, 20xx), Ampere (30xx), Ada (40xx), and newer recommend the open-source kernel modules
+  # Pascal (10xx), Maxwell (9xx), Kepler (7xx) and older use legacy branch that can only be installed from AUR
```
Kn0ax: Kepler is not supported by nvidia-580xx
image

install/config/hardware/nvidia.sh
```diff
@@ -31,19 +32,29 @@ if [ -n "$(lspci | grep -i 'nvidia')" ]; then
...
+  if [ "$NVIDIA_DRIVER" = "$NVIDIA_DRIVER_CURRENT" ]; then
+    DRIVER_PACKAGES=(nvidia-open-dkms nvidia-utils lib32-nvidia-utils libva-nvidia-driver)
+    sudo pacman -S --needed --noconfirm "${GENERAL_PACKAGES[@]}" "${DRIVER_PACKAGES[@]}"
+  elif [ "$NVIDIA_DRIVER" = "$NVIDIA_DRIVER_LEGACY" ]; then
+    sudo pacman -S --needed --noconfirm "${GENERAL_PACKAGES[@]}"
+    if ! command -v yay &>/dev/null; then
```
Kn0ax: yay is a base package of Omarchy.
idk if we need this

install/config/hardware/nvidia.sh
```diff
@@ -31,19 +32,29 @@ if [ -n "$(lspci | grep -i 'nvidia')" ]; then
...
-    "libva-nvidia-driver" # For VA-API hardware acceleration
     "qt5-wayland"
     "qt6-wayland"
   )
 
-  sudo pacman -S --needed --noconfirm "${PACKAGES_TO_INSTALL[@]}"
+  # Turing (16xx, 20xx), Ampere (30xx), Ada (40xx), and newer recommend the open-source kernel modules
+  # Pascal (10xx), Maxwell (9xx), Kepler (7xx) and older use legacy branch that can only be installed from AUR
+  if [ "$NVIDIA_DRIVER" = "$NVIDIA_DRIVER_CURRENT" ]; then
+    DRIVER_PACKAGES=(nvidia-open-dkms nvidia-utils lib32-nvidia-utils libva-nvidia-driver)
```
Kn0ax: I don't know if it was always like this, but apparently Nvidia-utils is a dependency for Nvidia packages, meaning we don't need to install it separately.

Kn0ax:
Should we inform users whether their graphics card is currently supported by NVIDIA or by the Arch team?

Right now, the script assumes that if a user does not have a GPU newer than the Pascal architecture, it will get the 580xx driver, which is not supported by older cards, and users are likely going to be unaware of this.

Kn0ax:
this should handle migration for mentioned cards 
(sorry, i couldn't add this as a commit to current branch for some reason.)

```bash
echo "Migrate legacy NVIDIA GPUs to nvidia-580xx driver"

# Skip if no nvidia packages are installed
if ! pacman -Qq 2>/dev/null | grep -q '^nvidia'; then
  exit 0
fi

# Skip if its a "modern" gpu
if lspci | grep -iE 'nvidia. *(RTX [2-9][0-9]|GTX 16)'; then
  exit 0
fi

# Remove current nvidia driver packages
omarchy-pkg-drop nvidia nvidia-dkms nvidia-utils lib32-nvidia-utils

# Install 580xx driver from AUR
yay -S --needed --noconfirm nvidia-580xx-dkms nvidia-580xx-utils lib32-nvidia-580xx-utils

```

ashddev:
> Should we inform users whether their graphics card is currently supported by NVIDIA or by the Arch team?
> 
> Right now, the script assumes that if a user does not have a GPU newer than the Pascal architecture, it will get the 580xx driver, which is not supported by older cards, and users are likely going to be unaware of this.

Thanks for taking the time to review my PR!

Yeah it may be simplest to just inform the user if their GPU is legacy and unsupported based on [this status table](https://wiki.archlinux.org/title/NVIDIA) and direct them to the wiki if so.

install/config/hardware/nvidia.sh
ashddev: Fair it is a dependency, but I don't see a problem with being explicit here for the sake of clarity and in case dependencies change.

Kn0ax:
@ryanrhughes nvidia-580xx-dkms and related packages are only available on AUR. Therefore they need to be installed using yay.

Also, my intention with this script was not only for Omarchy; rather, it was meant to be a general script to help people who want to use Hyprland + Arch with Nvidia cards. (Also for those who wants to install omarchy manually) 
If it’s not a problem for you, I’d like it to stay that way. The current changes make the script too specific to Omarchy. [idc too much about credit at all, although it was cool to see in omarchy ngl :')]

Thanks for all your hard work again.

ryanrhughes:
@Kn0ax the 580xx packages were added to OPR. https://github.com/omacom-io/omarchy-pkgs/commit/d87dd64e077bffe6e2664089e2a6eebc08179bde We only install packages from there so that we can ensure they're available in the ISO for offline installations. They'll get built and released there so they're available before we release this. 🙂

> Also, my intention with this script was not only for Omarchy; rather, it was meant to be a general script to help people who want to use Hyprland + Arch with Nvidia cards. (Also for those who wants to install omarchy manually)

I definitely have an appreciation for this idea but generally speaking; this becomes an incredibly slippery slope and is outside of the scope of the install file, I believe. 

That would need to live in a different utility somewhere as the files in `/install` really only execute during initial install and we intentionally keep those incredibly trim and opinionated so they're fast, error-free, and predictable.

As it sits now, it would work well for anyone whether they were using the ISO or manual install method.

> idc too much about credit at all
You absolutely get the credit for the work here. We just have to keep the files clean from excess comment blocks or it'd get to be a bit of a mess in here. 😅

Kn0ax:
Thanks for the info about OPR. Last time I checked, OPR was for Omarchy-specific custom packages (like the Chromium fork, etc.), so good to know.

I’ll probably make a separate gist for the "generalized" script.

_Pulling it from here was a nice little party trick for me :')_
<3

ryanrhughes:
The original setup was just to host some extras but as Arch / AUR has faced continual attacks, we upped it to house **all** packages we rely on for Omarchy. This gives us more stability and control and even allows us to maintain our stable vs edge setup with the AUR packages as well.

PR #3962, opened 2025-12-22
feat: added power profile indicator and change notification

maxdollinger:
I tweaked the scripts a little to indicate which power profile is currently active and to provide a notification when the power profile is switched via the menu.

The goal was to change as little as possible, so I just removed the part where the power profile indicator is removed.

### DIFF
--- a/bin/omarchy-menu
+++ b/bin/omarchy-menu
@@ -194,7 +194,7 @@ show_setup_power_menu() {
   if [[ "$profile" == "CNCLD" || -z "$profile" ]]; then
     back_to show_setup_menu
   else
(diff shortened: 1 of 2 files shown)


dhh:
We already have highlighting of the active power profile? Do you have a bug where it's not showing?

image

maxdollinger:
Thank's for your reply.
I checked and realized that I had overlooked the ‘italic’ style of the current option.
Sorry for not looking more closely.

I'll just overwrite the ‘.current’ class in my walker config to make the element more visible to me and close the PR.

Thank you very much for creating Omarchy, which is a great user experience overall.

PR #3964, opened 2025-12-22
feat: add clipboard state management and integrate with waybar

sgruendel:
implementing @abenz1267's idea from [#3187 ](https://github.com/basecamp/omarchy/discussions/3187)

### DIFF
--- a/bin/omarchy-clipboard-state
+++ b/bin/omarchy-clipboard-state
@@ -26,4 +26,9 @@ toggle)
   fi
   pkill -RTMIN+9 waybar
   ;;
+*)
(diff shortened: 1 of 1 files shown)


dhh:
When would you need/want to do this?

sgruendel:
> When would you need/want to do this?

Might come in handy when you yank stuff in Neovim that you don't want to show up in your clipboard history, or in general when you want to "pin" your current clipboard entry.

dhh:
Appreciate the suggestion, but I don't think that's something I'd use, so maybe we can document this for those who want it in the Discussions 👍

PR #3965, opened 2025-12-22
fix: use correct blue for Alacritty

sgruendel:
use blue from https://github.com/kepano/flexoki/blob/main/alacritty/flexoki-light.toml which is already used by kitty.conf

### DIFF
--- a/migrations/1765836579.sh
+++ b/migrations/1765836579.sh
@@ -1,3 +1,0 @@
-echo "Fix blue color for Alacritty"
-
-sed -i 's/blue = "#205EA6"/blue = "#4385BE"/' $OMARCHY_PATH/themes/flexoki-light/alacritty.toml 


dhh:
The themes file is inside of Omarchy, so we don't need a migration.

sgruendel:
> The themes file is inside of Omarchy, so we don't need a migration.

ah was too eager, removed it

dhh:
Need to fix this for Ghostty too.

sgruendel:
> Need to fix this for Ghostty too.

`themes/flexoki-light/ghostty.conf` just has `theme = Flexoki Light`, so it's built in already and nothing to configure from Omarchy side.

dhh:
Then I'm not sure we can conclude that any blue is wrong. Because the one used by Ghostty and Alacritty is the same. It's Kitty that's different.

sgruendel:
> Then I'm not sure we can conclude that any blue is wrong. Because the one used by Ghostty and Alacritty is the same. It's Kitty that's different.

As linked in the initial commit, https://github.com/kepano/flexoki/blob/main/alacritty/flexoki-light.toml is what's also used in Omarchy's `themes/flexoki-light/alacritty.toml` -- except for the blue.  And that blue is also used in `themes/flexoki-light/kitty.conf`.

dhh:
I fixed this on the dynamic config branch 👍

PR #3967, opened 2025-12-22
feat: optional re-run scripts instead of force close

KenErikson:
### What

Add option for `omarchy-show-done` to run the script again. Hope it's usecase is clear enough. Especially when starting out and configuring omarchy, one can be forced to open up the menu quite a few times, and this should reduce friction a bit.

Hopefully someone finds it useful, at least.

### Why

It is tedious to open up the menu and find the right entry every time one:
- want to install multiple programs/themes (yes, one can tab/multi-select in some scripts, but not others)
- do a typo e.g. when creating a webapp, and want to restart

### Screenshots

image

https://github.com/user-attachments/assets/489dcbe9-4367-49e3-8356-303444bbf8f3

https://github.com/user-attachments/assets/c230991d-73ea-4419-8392-1128ce4334ad

### Considerations
- Is it worth it/useful enough for everyone?
- `r` could be swapped out with some other key/key-combo
- The messages are clear enough IMO, but UX is not my expertise
- Unclear for developers how it should be used (additional complexity)?
- Unclear for users what "run again" means?
- Confusing for existing users?

_Thanks for Omarchy_

### DIFF
--- a/bin/omarchy-launch-floating-terminal-with-presentation
+++ b/bin/omarchy-launch-floating-terminal-with-presentation
@@ -1,4 +1,11 @@
 #!/bin/bash
 
 cmd="$*"
-exec setsid uwsm-app -- xdg-terminal-exec --app-id=org.omarchy.terminal --title=Omarchy -e bash -c "omarchy-show-logo; $cmd; omarchy-show-done"
+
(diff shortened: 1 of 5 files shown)


dhh:
I think this is an interesting idea, but we use show-done in a bunch of cases where it does not make sense. I think it's going to be confusing to have as something like this in every case. Thanks for the suggestion, though!

PR #3972, opened 2025-12-23
feat: add support for single-label DNS resolution configuration

TheCodeOne:
Adds a prompt in `omarchy-setup-dns` to enable single-label DNS resolution when users configure DHCP or custom DNS servers. This allows single-label queries (e.g., `server/`) to work with local DNS servers instead of only using mDNS.

## Changes

- Added interactive prompt in `omarchy-setup-dns` when "DHCP" or "Custom" DNS is selected
- Creates `/etc/systemd/resolved.conf.d/15-single-label.conf` with `ResolveUnicastSingleLabel=yes` when enabled
- Removes the config file if user declines
- Automatically removes config file when "Cloudflare" is selected

### DIFF
--- a/bin/omarchy-setup-dns
+++ b/bin/omarchy-setup-dns
@@ -1,5 +1,29 @@
 #!/bin/bash
 
+configure_single_label_dns() {
+  local resolved_conf_d="/etc/systemd/resolved.conf.d"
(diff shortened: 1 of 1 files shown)


dhh:

PR #3976, opened 2025-12-23
Fix: missing mako notifications icons

brutalsam96:
### What this change does
Adds icon-path field to mako config file

### Why
Notifications that are called with notify-send or any other api that include icons such as battey-caution, dialog-information etc. were missing from notification, this was because mako config was missing path to icon libraries.

### Verification
Triggered a notification using notify-send
Verified icons are showing up properly
### Screenshots

**Before**
no-icon
**After**
with-icon

### DIFF
--- a/default/mako/core.ini
+++ b/default/mako/core.ini
@@ -6,6 +6,7 @@ padding=10,15
 border-size=2
 max-icon-size=32
 font=sans-serif 14px
+icon-path=icon-path=/usr/share/icons/Yaru-dark:/usr/share/icons/Yaru:/usr/share/icons/breeze-dark:/usr/share/icons/breeze:/usr/share/icons/Adwaita:/usr/share/icons/hicolor:/usr/share/pixmaps
 
 [app-name=Spotify]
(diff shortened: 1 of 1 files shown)


dhh:
Appreciate this, but I don't think those icons look good at all. We use nerdfont glyphs for a lot of the notifications instead. I think that better fits the themed aesethetic of Omarchy.

brutalsam96:
> Appreciate this, but I don't think those icons look good at all. We use nerdfont glyphs for a lot of the notifications instead. I think that better fits the themed aesethetic of Omarchy.

No problem, i noticed this while i was tweaking battery monitor notification i guess i can use same nerdfont glyphs they look much better, one more thing since we use Nerdfont now perhaps, global settings of mako needs a change to nerdfont as well ? currenty its set so sans-serif, also can turn off icons completely as well.

```
icons=0
font=JetBrainsMono Nerd Font 10

```

dhh:
I don't like using a monospace font for the notifications, so this is intentional.

PR #3977, opened 2025-12-23
Enable parallel compilation of packages

cover:
## What
Enable `make` to [parallel compiling](https://wiki.archlinux.org/title/Makepkg#Parallel_compilation) packages using all cpu cores.

## Why
I was updating an AUR package today and noticed that it was using just one core, and it was taking some time to finish. After enabling all cores it only took a few seconds. Some packages – like `qt5-webengine` – could take a very long time to compile on just 1 core.


defer:
I like this, I have this. But I wouldn't recommend it as a default. Some packages have broken dependency chains and just fail. But my biggest gripe is that some people don't have enough memory to have a per-cpu thread and would be clueless as to why the build is being killed.

dhh:
Agree with @defer. I'd rather have a  slower standard build that always works than a faster one that some times breaks. Appreciate the suggestion, though!

PR #3982, opened 2025-12-24
fix: remove +noidle tag from window after update

### LINKED ISSUE #3973: Bug: `omarchy-update-without-idle` adds `noidle` tag but never removes it
### System details

AMD 7640U (Framework 13), Omarchy 3.2.3

## Summary

The script `omarchy-update-without-idle` tags the focused window with `noidle` to prevent idle during updates, but never removes the tag when the update completes. Combined with the window rule `idleinhibit always, tag:noidle`, this permanently blocks hypridle from detecting idle until the window is closed or manually untagged.

## Steps to reproduce

1. Focus a terminal window
2. Run a command that calls `omarchy-update-without-idle` (or run it directly)
3. Complete the update
4. Leave system idle - screensaver/lock/DPMS never triggers

## Expected behavior

The `noidle` tag should be removed when the update completes.

## Actual behavior

The tag persists indefinitely, blocking all idle detection.

## Affected files

- `~/.local/share/omarchy/bin/omarchy-update-without-idle`
- `~/.local/share/omarchy/default/hypr/apps/system.conf` (defines `windowrule = idleinhibit always, tag:noidle`)

## Suggested fix

Wrap the update command and remove the tag on completion:

```bash
#!/bin/bash
hyprctl dispatch tagwindow +noidle &> /dev/null
# run update command here
# ...
hyprctl dispatch tagwindow -- -noidle &> /dev/null
```

Or create a corresponding `omarchy-update-idle-restore` script to call after updates.

## Workaround

Manually remove the tag:

```bash
hyprctl dispatch tagwindow -- -noidle
```

Or remove from all windows:

```bash
hyprctl -j clients | jq -r '.[] | select(.tags | index("noidle")) | .address' | \
  xargs -I{} sh -c 'hyprctl dispatch focuswindow address:{} && hyprctl dispatch tagwindow -- -noidle'
```

sgruendel:
fixes #3973

### DIFF
--- a/bin/omarchy-toggle-tagwindow-idle
+++ b/bin/omarchy-toggle-tagwindow-idle
@@ -1,0 +1,24 @@
+#!/bin/bash
+
+COMMAND="$1"
+
+if [[ -z "$COMMAND" ]]; then
+  echo "Usage: omarchy-toggle-tagwindow-idle <on|off>"
+  exit 1
+fi
+
+case "$COMMAND" in
+  on)
+    # Ensure screensaver/sleep doesn't set in during updates
+    hyprctl dispatch tagwindow +noidle &> /dev/null
+    ;;
+  off)
+    # Re-enable screensaver/sleep after updates
+    hyprctl dispatch tagwindow -- -noidle &> /dev/null 
+    ;;
+  *)
+    echo "Invalid command: $COMMAND"
+    echo "Usage: omarchy-toggle-tagwindow-idle <on|off>"
+    exit 1
+    ;;
+esac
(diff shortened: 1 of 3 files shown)


dhh:
I don't think we need this. When the update window is closed, the noidle tag automatically disappears.

dhh:
Actually maybe this applies noidle to the entire ghostty instance that stays persistent.

sgruendel:
> I don't think we need this. When the update window is closed, the noidle tag automatically disappears.

yeah it's an issue if you manually run `omarchy-update` from an existing terminal and leave it open

PR #3985, opened 2025-12-25
feat: add window transparency TUI command

sharifli4:
Adds omarchy-cmd-transparency, a simple TUI for adjusting window transparency in Hyprland. Features:
- Interactive window selector using gum filter
- Real-time transparency adjustment with arrow keys
- Visual progress bar showing current transparency level
- Option to persist the rule to hyprland.conf

------ 
Command: window-transparency


dhh:
That's an interesting idea. I think it could make a good AUR package for those who want it. But it's not something I've ever had a desire to tweak on the fly, so think it's best as an external package 👍

PR #3988, opened 2025-12-25
feat: add headphone battery indicator to waybar

sharifli4:
Shows Bluetooth headphone icon with battery percentage when headphones are the active audio sink. Uses UPower to retrieve battery level.

### DIFF
--- a/bin/omarchy-headphone-status
+++ b/bin/omarchy-headphone-status
@@ -1,0 +1,23 @@
+#!/bin/bash
+# Shows headphone icon with battery level if bluetooth audio device is the active sink
+
+# Get the active audio sink (marked with *)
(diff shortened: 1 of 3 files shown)


dannymcc:
I would prefer this to be a user added addition to waybar rather than the out-of-the-box experience.

pomartel:
There is an easier way to show the headphone icon built-in the pulseaudio waybar module: https://man.archlinux.org/man/extra/waybar/waybar-pulseaudio.5.en#ICONS:

Maybe this should be included by default.

CantC0unt:
This should definitely be using pulseaudio and upower


And similar stuff has been rejected in the past
https://github.com/basecamp/omarchy/pull/3377
https://github.com/basecamp/omarchy/pull/2410

PR #3990, opened 2025-12-25
Update PostgresSQL version in installation script from 17 to 18

Jan747:
Bumping PostgresSQL DB from 17 to 18.


JovanKanevche (approved):
good stuff

PR #3992, opened 2025-12-25
Fix omarchy-update failing outside Hyprland session

jkbdco:
Discovered this when trying to update my machine via ssh.

The omarchy-update-without-idle script fails when HYPRLAND_INSTANCE_SIGNATURE is not set (e.g., when running updates via SSH, TTY, or a non-Hyprland session).

Even though the hyprctl output is suppressed, set -e in the parent script
catches the non-zero exit code and triggers the error trap, causing the
update to fail with no visible error message.

Adding || true allows the update to proceed gracefully when the idle
inhibitor can't be set.

### DIFF
--- a/bin/omarchy-update-without-idle
+++ b/bin/omarchy-update-without-idle
@@ -1,4 +1,4 @@
 #!/bin/bash
 
 # Ensure screensaver/sleep doesn't set in during updates
-hyprctl dispatch tagwindow +noidle &> /dev/null
+hyprctl dispatch tagwindow +noidle &> /dev/null || true


dhh:
Target dev branch and rebase 👍

jkbdco:
> Target dev branch and rebase 👍

Hey @dhh — when I rebased onto dev, I hit a conflict. Looks like e44e937 already addressed this by moving idle management elsewhere and making this script a no-op. Should I still proceed, or is this PR no longer needed?

sgruendel:
> > Target dev branch and rebase 👍
> 
> Hey @dhh — when I rebased onto dev, I hit a conflict. Looks like [e44e937](https://github.com/basecamp/omarchy/commit/e44e937284ded70537f336c7f49cc09b6df0772c) already addressed this by moving idle management elsewhere and making this script a no-op. Should I still proceed, or is this PR no longer needed?

Looks like you should now apply your changes to `omarchy-update-perform` directly.

PR #4001, opened 2025-12-27
Add wezterm support

broccoli-robbie:
- Added WezTerm as an installable terminal via the Omarchy menu, installs from AUR by default.
 - Compatible with theme switching and screensaver.
 - Mostly default WezTerm config.

### DIFF
--- a/bin/omarchy-menu
+++ b/bin/omarchy-menu
@@ -261,11 +261,11 @@ show_install_editor_menu() {
 }
 
 show_install_terminal_menu() {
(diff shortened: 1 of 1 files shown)


dhh:
I don't want to have any of the default terminal options depend on a floating git version. We need to wait until there's a proper wayland-compatible release on the official arch repo.

dhh:
When we have a stable package release, we can also use the new dynamic config templates to provide the theme 👍

PR #4010, opened 2025-12-28
Fix gruvbox typo

morgoth:


dhh:
Unfortunately this would break existing gruvbox users backgrounds that point to this. So don't think it's worth fixing. It's all internal anyway.

PR #4034, opened 2025-12-31
bluetooth: disable AutoEnable by default

Rishabh-Sarang:
- Disable Bluetooth auto-enable at boot using BlueZ policy
- Prevents Bluetooth powering on by default
- Improves battery life and privacy
- Adds AutoEnable=false under [Policy] in main.conf
- Restarts bluetooth.service

### DIFF
--- a/migrations/bluetooth-off-by-default.sh
+++ b/migrations/bluetooth-off-by-default.sh
@@ -1,0 +1,25 @@
+#!/bin/bash
+set -e
+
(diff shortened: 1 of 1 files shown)


dhh:
Appreciate the suggestion, but I don't think this is a good default. Bluetooth should be on by default, as is common in most operating systems. The Bluetui tui already has a way to turn it off after you've booted if you don't want it.

PR #4036, opened 2025-12-31
Update qemu.conf

eososlinux:
Version de hyprland 0.53 regla de ventana archivo qemu.con

windowrule {  name = windowrule-1  opacity = 1.0 1.0  match:class = ^(qemu)$ }


dhh:
What we currently have has already been updated to 0.53 syntax and is throwing no errors in my testing. It's also the same rule we use in a bunch of other places.

eososlinux:
> What we currently have has already been updated to 0.53 syntax and is throwing no errors in my testing. It's also the same rule we use in a bunch of other places.

Faltaba ese archivo, según la documentación en tu rama no estaba corregido, espero te funcione bien

PR #4040, opened 2025-12-31
Update nvidia.sh

mwhagedorn:
change nvidiea setup to ensure nouveau display drivers are used as I have old hardware


dhh:
This was already fixed in https://github.com/basecamp/omarchy/pull/3956.

PR #4048, opened 2026-01-01
Zsh initial support

rsd:
Adds initial supports for users that wish to use zsh instead of bash.
Keeps all bash defaults style for zsh.
Shares the imports that are shell agnostic.

### DIFF
--- a/default/zsh/bindings
+++ b/default/zsh/bindings
@@ -1,0 +1,15 @@
+# Key bindings
+# Standard Emacs mode
+bindkey -e
+
+# Arrow keys match what you've typed so far against your command history
+bindkey '^[[A' history-search-backward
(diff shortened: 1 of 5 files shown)


neheb:
What's the relation between this and omarchy-zsh?

neheb:
https://github.com/omacom-io/omarchy-zsh/blob/master/functions/functions.zsh#L58

ryanrhughes:
[omarchy-zsh](https://github.com/omacom-io/omarchy-zsh) is the official package for zsh support. It maintains 1-1 parity with the bash setup.

rsd:
so, what else is official and not in omarchy?
I miss the list, so works dont clash again.

PR #4051, opened 2026-01-01
Set macOS-style screenshot keybindings

vadika:
Not every keyboard have the printscreen keys, so having macOS-screenshot keybindings is helpful and in overall omarchy style.


daviewales:
I'm assuming the actual Mac shortcuts (e.g Shift+Super+4) were not used because Shift+Super is usually used to launch an application in Omarchy?

vadika:
> I'm assuming the actual Mac shortcuts (e.g Shift+Super+4) were not used because Shift+Super is usually used to launch an application in Omarchy?

yes. was selected closest not to ruin muscle memory :-)

dhh:
Added `Super + Ctrl + C` to open the capture menu instead for keyboards without the printscr key 👍

vadika:
> Added `Super + Ctrl + C` to open the capture menu instead for keyboards without the printscr key 👍

Super + Ctrl + P, according to the commit

dhh:
I changed my mind after that commit 😄. Instead of adding another 4 keybindings, I just added once for the capture menu (`Super + Ctrl + C`).

PR #4053, opened 2026-01-01
Use theme config templates with singular color definition

dhh:
Most config templates just set explicit colors, which we can generate from a single definition instead of dealing with many individual setups.


themes/flexoki-light/colors.toml
```diff
@@ -0,0 +1,24 @@
+accent = "#205EA6"
+active_border_color = "#D14D41"
+active_tab_background = "#CECDC3"
+background = "#FFFCF0"
+color0 = "#100F0F"
+color1 = "#D14D41"
+color10 = "#879A39"
+color11 = "#D0A215"
+color12 = "#205EA6"
```
sgruendel: blue should be `#4385BE`, see Steph Ango's original definition of light colors under https://github.com/kepano/flexoki/tree/main?tab=readme-ov-file#light-colors

PR #4060, opened 2026-01-02
Add mouse movement threshold to screensaver exit

axelfontaine:
This fixes mouse movement exit for the screensaver by introducing a jitter busting threshold (at least 10 mouse move events within a 5 second window).

Mouse tested: Logi Mx Master 3 via 2.4 GHz dongle

Previous behavior: immediate exit
New behavior: works as expected, exits with mouse movement, but no longer due to jitter


axelfontaine:
Still unreliable, closing for now

PR #4061, opened 2026-01-02
Allow unsecure URLs to be copied

AlexZeitler:
Current version only allows URLs in a secure context like HTTPS to be copied.

The new version allows unsecure URLs to be copied.

It's useful when you're using it for URLs during development without HTTPS (e.g. via SSH)

It is using [chrome.offscreen API](https://developer.chrome.com/docs/extensions/reference/api/offscreen)


dhh:
Thanks for the contribution! This is no longer needed because the insecure-URL copy case is now handled by the offscreen-document implementation added in https://github.com/basecamp/omarchy/commit/28d853402e9d8726fc315eb18b7517421195ccd9. That implementation also includes reliability handling for the offscreen document and Chromium service-worker updates, so I'm closing this as superseded.

PR #4065, opened 2026-01-03
omarchy remove dev (install dev mirror)

JustinMissmahl:
follow up on https://github.com/basecamp/omarchy/pull/2300#issuecomment-3414278268

### DIFF
--- a/bin/omarchy-menu
+++ b/bin/omarchy-menu
@@ -366,15 +366,60 @@ show_install_elixir_menu() {
 }
 
 show_remove_menu() {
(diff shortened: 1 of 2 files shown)


JustinMissmahl:
one question:

Do we still want to have added a new "Custom" option or something like that in the install / remove dev sections for custom control of mise packages or is that not needed?

JustinMissmahl:
aaah yeah and another question was would it be wanted to maybe look into having a shared env config for the development options to not be in need to hold track of two separate lists or is it wanted to stay more modular?

dhh:
I think this is pretty good as is 👌

PR #4066, opened 2026-01-03
Introduce `omarchy-theme-bg-launch` as a unified wallpaper manager and add `awww` as an optional backend with transition effects.

Hyperbola-QAQ:
- Introduce `omarchy-theme-bg-launch` to centralize wallpaper management
- Add support for `awww` as a background backend with configurable transitions
- Expose new configuration options in `uwsm/default`:
  - `BACKGROUND_BACKEND`: choose between `'swaybg'` (default) or `'awww'`
  - `AWWW_TRANSITION_ARGS`: customize transition effects (e.g., `--transition-type wave`)
- Update Hyprland autostart config to use the new launcher script


dhh:
Appreciate the suggestion, but I don't think there's enough value in offering alternate background services for this to be worth the complexity ✌️

PR #4074, opened 2026-01-04
Add WezTerm terminal emulator support

S1M0N38:
## Summary
- Add WezTerm as an installable terminal option with full theme support
- Add WezTerm config with dynamic theme loading via dofile
- Add terminal icon and screensaver support for WezTerm

## Test plan
- [ ] Install WezTerm via menu and verify it works
- [ ] Verify theme switching applies to WezTerm
- [ ] Test screensaver with WezTerm as default terminal

---
**DO NOT MERGE**

This is just an experiment to explore the various quirks you encounter when adding WezTerm support to Omarchy. It might serve as a reference when a proper PR is created.

### DIFF
--- a/bin/omarchy-install-terminal
+++ b/bin/omarchy-install-terminal
@@ -69,7 +69,7 @@ EOF
 [Desktop Entry]
 Type=Application
 TryExec=wezterm
-Exec=wezterm start
+Exec=omarchy-wezterm-launcher start
 Icon=org.wezfurlong.wezterm
 Terminal=false
 Categories=System;TerminalEmulator;
@@ -86,7 +86,7 @@ X-TerminalArgDir=--cwd
 
 [Desktop Action New]
 Name=New Terminal
-Exec=wezterm start
+Exec=omarchy-wezterm-launcher start
 EOF
   fi
 
--- a/bin/omarchy-wezterm-launcher
+++ b/bin/omarchy-wezterm-launcher
@@ -1,0 +1,134 @@
+#!/bin/bash
+
+# omarchy-wezterm-launcher - Workspace-aware WezTerm launcher
+# Detects the current WezTerm workspace and spawns new windows in it
+# Falls back to standard `wezterm` behavior for non-WezTerm launches
+
+# Function to extract WEZTERM_PANE from a process environment
+get_wezterm_pane_from_pid() {
+  local shell_pid="$1"
+  if [[ -z "$shell_pid" || ! -f "/proc/$shell_pid/environ" ]]; then
+    return 1
+  fi
+
+  # Read the environ file and extract WEZTERM_PANE value
+  # environ uses null bytes as separators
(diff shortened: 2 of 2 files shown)


bin/omarchy-install-terminal
```diff
@@ -12,14 +12,28 @@ case "$package" in
...
   exit 1
   ;;
 esac
 
 # Install package
-if omarchy-pkg-add $package; then
+install_success=false
+if [[ $package == "wezterm" ]]; then
+  # Install wezterm-git from AUR
+  if yay -S --noconfirm wezterm-git; then
```
dhh, line 26: We should just be using the official wezterm package that's on the official Arch repo.

bin/omarchy-wezterm-launcher
```diff
@@ -0,0 +1,134 @@
+#!/bin/bash
+
+# omarchy-wezterm-launcher - Workspace-aware WezTerm launcher
```
dhh, line 3: We don't have this for any other term, so I don't think that makes a good default. You should be able to add a terminal to any workspace.

themes/catppuccin-latte/wezterm.lua
```diff
@@ -0,0 +1,39 @@
+-- Catppuccin Latte theme for WezTerm
```
dhh, line 1: We should use the new dynamic templates for this,

bin/omarchy-install-terminal
S1M0N38: It seems that the official Arch repo `wezterm` package is not compatible with Wayland (it’s an older version that does not include the Wayland fixes yet). Using `wezterm-git` gives access to the latest version of WezTerm, but it requires compilation, which can take quite some time depending on the system hardware. To be honest, I was surprised that I was able to compile it successfully on the first try (MacBook Air 2014 / Omarchy 3.2).

I completely agree that we should use the official package from the Arch repository:
1. Better security
2. Better user experience during installation
3. Consistency with other app installations

however, when I try to install from the official arch repo, I was not able to start it.

dhh:
Hmm, I don't see that we can add this as an official option until there's an official and up-to-date package available. Can't rely on AUR for core system packages like this. Let's pick it up again when there's an official package option that's compatible 👍

bin/omarchy-wezterm-launcher
S1M0N38: Yeah, reverting 403b345. It was also introducing some major issues in other parts of Omarchy. The reason this was created in the first place was to avoid the problem where new terminal windows spawned with SUPER + RETURN would always open in the `default` WezTerm workspace.

So, if the user has WezTerm open in a different WezTerm workspace (WezTerm workspace ≠ Hyprland workspace — they’re a bit like tmux sessions), the new terminal window would be spawned in the background (in the default WezTerm workspace).

PR #4078, opened 2026-01-04
Fix screensaver skip on timeout

devfros:


devfros:
It would go to lock screen right away instead of showing screensaver first

PR #4088, opened 2026-01-05
Add voice dictation with voxtype

dhh:
Add AI voice dictation using [Voxtype](https://voxtype.io/) that can be installed via _Install > AI > Dictation_, and which then shows up in the Waybar.

### DIFF
--- a/bin/omarchy-voxtype-config
+++ b/bin/omarchy-voxtype-config
@@ -1,0 +1,6 @@
+#!/bin/bash
+set -e
+
+# Used by Voxtype waybar module to open config on right click
+
+exec omarchy-launch-editor ~/.config/voxtype/config.toml
--- a/bin/omarchy-voxtype-help
+++ b/bin/omarchy-voxtype-help
@@ -1,5 +1,0 @@
-#!/bin/bash
-
-# Reminds the user of the dictation hotkey when the Voxtype mic in the Waybar is clicked.
-
-notify-send "    Voxtype Dictation Ready" "Hold Super + Ctrl + X to dictate.\nEdit ~/.config/voxtype/config.toml for options." -t 10000
--- a/bin/omarchy-voxtype-install
+++ b/bin/omarchy-voxtype-install
@@ -15,5 +15,5 @@ if gum confirm "Install Voxtype + AI model (~400MB) to enable dictation?"; then
   voxtype setup systemd
 
   omarchy-restart-waybar
-  omarchy-voxtype-help
+  notify-send "    Voxtype Dictation Ready" "Hold Super + Ctrl + X to dictate.\nEdit ~/.config/voxtype/config.toml for options." -t 10000
 fi
--- a/config/waybar/config.jsonc
+++ b/config/waybar/config.jsonc
@@ -150,8 +150,8 @@
       "transcribing": "󰔟"
     },
     "tooltip": true,
-    "on-click-right": "omarchy-voxtype-model",
-    "on-click": "omarchy-voxtype-help"
+    "on-click-right": "omarchy-voxtype-config",
+    "on-click": "omarchy-voxtype-model"
   },
   "tray": {
     "icon-size": 12,
--- a/migrations/1767685679.sh
+++ b/migrations/1767685679.sh
@@ -49,8 +49,8 @@ patch -N ~/.config/waybar/config.jsonc << 'EOF'
 +      "transcribing": "󰔟"
 +    },
 +    "tooltip": true,
-+    "on-click-right": "omarchy-voxtype-model",
-+    "on-click": "omarchy-voxtype-help"
++    "on-click-right": "omarchy-voxtype-config",
++    "on-click": "omarchy-voxtype-model"
 +  },
    "tray": {
      "icon-size": 12,


peteonrails:
I've been working to get `voxtype` ready enough for a PR to  Omarchy for inclusion, so this work makes my day. First class support for this distribution is a priority of mine: happy to take any feedback, coordinate, or offer help. There is a new release today that includes icon themes (and manpages), one of the themes uses the default Omarchy fonts in the Waybar.

bin/omarchy-install-voxtype
```diff
@@ -32,9 +43,14 @@ if ! grep -q '"custom/voxtype"' $waybar_config; then
   sed -i '/"custom\/screenrecording-indicator"[[:space:]]*:[[:space:]]*{/,/^[[:space:]]*},/{
   /^[[:space:]]*},/a\
   "custom/voxtype": {\
-    "exec": "voxtype status --follow --extended --format json",\
+    "exec": "omarchy-voxtype-status",\
```
peteonrails: I hope to have made this easier for you by adding this switch in voxtype 0.4.5:

`voxtype status --follow --format json --icon-theme omarchy` 

You can also specify via voxtype config (~/.config/voxtype/config.toml):
```
  [status]
  icon_theme = "omarchy"
```

But instead of the NerdFont FA icons, I used:   
- idle: U+F036C (mdi-microphone) 
- recording: U+F040A (mdi-record)
- transcribing: U+F04CE (mdi-sync)
- stopped: U+F036D (mdi-microphone-off)

 I'd be happy to update the theme to match whichever set y'all prefer. 

OR to better decouple yourself in case you update themes, you could do this in your config: 

```
  [status.icons]
  idle = "󰍬"           
  recording = "󰍬"     
  transcribing = "󰔟"  
  stopped = "󰍭"        
```

dhh: You're the best! That's exactly what I was looking for 😄

dhh: Seems like the latest release on AUR is bust, though:

image

peteonrails: Ouch. On it.

peteonrails: > Seems like the latest release on AUR is bust, though:

Problem in my packaging script. Fixed and uploaded as 0.4.5-2. Thanks for giving it a try!

dhh: Hmm, still seeing the same error:

image

peteonrails: I have had some trouble previously with yay caching and not catching pkgrel updates. I'll bump to 0.4.6 and that should clear it up.

peteonrails: I've tagged and pushed 0.4.6 - did that clear it up?

dhh: Hmm, weird, nope:

image

peteonrails: :bangbang:  If you delete local cache does the package install? I suspect the old PKGBUILD and checksums are hanging around from when they did not match the binaries. 

```
  rm -rf ~/.cache/yay/voxtype-bin
  yay -S voxtype-bin
```

dhh: Bingo!

dhh:
@peteonrails It would be nice if when we do "voxtype setup --download" that it doesn't give this scary sounding error after install:

image

dhh:
@ryanrhughes Would be nice to get the AUR package onto the OPR.

dhh:
@peteonrails Also would be nice if "voxtype setup model" would offer to adjust the config for a model that has already been downloaded. I've set that up on right click so it's easy to switch between models, but you have to redownload to be able to change the model.

peteonrails:
> @peteonrails Also would be nice if "voxtype setup model" would offer to adjust the config for a model that has already been downloaded. I've set that up on right click so it's easy to switch between models, but you have to redownload to be able to change the model.

@dhh Thanks for that feedback - I didn't consider that use case and it makes total sense. Implementing it in the next build.

dhh:
@peteonrails Would also be nice if we could hide these post-install instructions, since they don't apply to the Omarchy setup. Maybe --silence-post-install or something?

image

dhh:
@peteonrails Also, when I'm testing this in a VM, I get a core dump:

image

Would be nice if there was a guard against whether you need from the system and a user friendly error message if this can't be setup.

peteonrails:
> @peteonrails Would also be nice if we could hide these post-install instructions, since they don't apply to the Omarchy setup. Maybe --silence-post-install or something?
> image

I can do that. Thinking of removing all of the post install stuff into a `voxtype setup check` subcommand.

peteonrails:
> @peteonrails Also, when I'm testing this in a VM, I get a core dump:
> image
> 
> Would be nice if there was a guard against whether you need from the system and a user friendly error message if this can't be setup.

Illegal Instruction is usually triggered when running on a Zen3 or earlier CPU using the AVX-512 build. Is it possible that in the VM, that `/usr/bin/voxtype` is symlinked to `/usr/lib/voxtype/voxtype-avx512` instead of `/usr/lib/voxtype/voxtype-avx2`? 

The AVX-512 Zen4 optimized binaries have been a real headache and I am considering removing them in future packages of voxtype-bin. That would lead to about a 10% performance hit, but only on CPU work: if you have opinions on that I'd certainly take them under advisement.

dhh:
I'm seeing this on a MS-A2 with Zen5 cores running Proxmox. So it's probably the Proxmox layer that's causing problems? I don't actually need this to work from within Proxmox, but it would be nice if others who try don't just see a core dump.

peteonrails:
Following up here - I added a startup guard for the VM coredump seen above, and added `--no-post-install` and `--quiet` flags for the setup command (PR inbound for `--no-post-install`) 

Additionally: 
`voxtype setup model` will let you switch models without re-downloading. 
`voxtype setup model --set <modelname>` updates the config without using the menu (must be present already)
`voxtype setup model --set <modelname> --restart`  will restart the daemon after the update

Also updated the icon-theme to match the icons you set in your initial PR, expecting it to go unused by default.

Hopefully these enhancements are useful. Thanks again for giving  this a go.

PR #4091, opened 2026-01-05
Map physical screenshot key to screenshot

axelfontaine:
This key is present on all Logi MX Keys models and automatically sends SUPER+SHIFT+S to the OS.


axelfontaine:
Superseeded by #4176

PR #4095, opened 2026-01-06
Allow overriding webapp launch browser via environment variable

benhoman:
Allow the user to set an environment variable to override the browser temporarily (or semi-permanently) without changing the default browser.

I find this useful because I mainly use Zoom for work which I only use a specific browser for so I can either update the Zoom .desktop file or create a new one that will tell the webapp launcher which browser to use without changing by default browser.


dhh:
I think you're better off just changing the .desktop file in this case.

PR #4100, opened 2026-01-06
Add icons for additional audio profiles in Waybar

pomartel:
So first of all I know that similar PR have been rejected in the past but this does not add any moving parts to the waybar.

It is a builtin feature of the pulseaudio waybar module that replaces the default icon with a headset icon when a bluetooth headset is connected. If find it super useful to know if my headset actually connected to my computer or if it auto-connected to my phone instead.

screenshot-2026-01-06_11-26-55

### DIFF
--- a/config/waybar/config.jsonc
+++ b/config/waybar/config.jsonc
@@ -115,11 +115,6 @@
     "format-muted": "",
     "format-icons": {
       "headphone": "",
-      "hands-free": "",
-      "headset": "",
-      "phone": "",
-      "portable": "",
-      "car": "",
       "default": ["", "", ""]
     }
   },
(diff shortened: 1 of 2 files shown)


zigmoo:
Like!

config/waybar/config.jsonc
```diff
@@ -114,6 +114,12 @@
     "scroll-step": 5,
     "format-muted": "",
     "format-icons": {
+      "headphone": "",
+      "hands-free": "",
+      "headset": "",
+      "phone": "",
+      "portable": "",
+      "car": "",
```
dhh: I think we can just get by with the headphones.

pomartel:
@dhh Sure thing! I updated the PR to keep just the headphone.

PR #4104, opened 2026-01-06
feat: Update omarchy-menu-keybindings to work with submaps

JakePi3:
## Submaps

Submaps are a feature of hyprland where you enter an alternate mode that creates a new set of keybindings.

The keybindngs are associated with a corresponding submap name such as "remote" or "gamemode".

In this mode only newly defined keybinds are available while the default keybinds are disabled, until the user exits the submap and returns to the default keybinds.

## Why is this Useful?

Some keybindings interfere with GUI/TUI apps because they capture and interpret the key combinations instead of forwarding them down to the underlying application.  See discussion [here](https://github.com/basecamp/omarchy/discussions/3105).

Some keybindings may also interfere with gaming because combinations are too close to something that may terminate the application (consider ALT+W vs SUPER+W).

## Why is this necessary?

Although it's possible to make use of submaps without modifying `bin/omarchy-menu-keybindings`, the existing operation will print keybindings across all submaps, including duplicates, even when some key combinations are not available within the context of the currently selected submap.

## Example of Using a Submap

Easiest is to modify `~/.config/hypr/bindings.conf` as in the linked discussion.

### DIFF
--- a/bin/omarchy-menu-keybindings
+++ b/bin/omarchy-menu-keybindings
@@ -241,7 +241,7 @@ output_keybindings() {
 }
 
 print_usage() {
-  echo "Usage: omarch-menu-keybindings [--print|-p] [--submap|-s <submap>] [--help|-h]"
+  echo "Usage: omarchy-menu-keybindings [--print|-p] [--submap|-s <submap>] [--help|-h]"
 }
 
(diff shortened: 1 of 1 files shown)


dhh:

PR #4106, opened 2026-01-07
Voxtype: suppress generic post-install instructions during install / setup

peteonrails:
Voxtype v0.4.9 adds a `--no-post-install` flag that suppresses the "Next steps" instructions after downloading the model. Since Omarchy handles systemd setup, hotkey configuration, and displays its own success notification, these generic instructions are redundant and potentially confusing to users.

The flag still shows download progress, just suppresses the post-download instructions.

If the maintainers prefer 100% silent operation, this can be changed to `--quiet` instead.


dhh:
Hmm, didn't seem to work. I have 0.4.9, and it still posted the install notes here. Also, after installing on a F13, it seems like it's failing in this odd way:

image

dhh:
The --no-post-install does seem to work when you rerun the setup command. But didn't on the first run?

PR #4108, opened 2026-01-07
feat: Enhance battery monitoring with time estimates, configurable notifications, and power profile management.

Dan-Kingsley:
I have been having issues, not seeing any notifications for low battery, then being shut down completely when my battery dies on my laptop. So I have created a more comprehensive battery monitoring system setup. Allowing users to configure how their computer reacts to low battery, and when they should receive notifications. Additionally I have implemented "X time until full" or "X time until empty" on both the default waybar config and the low battery notifications.


dhh:
Appreciate the work on this, but it's too much for the default setup.

PR #4119, opened 2026-01-07
feat: Add Omarchian agent pack for OpenCode

IgorWarzocha:
Hi fellow Omarchians, OpenCode agents & skills nerd here, Howaboua. I present you with the ultimate pack to manage Omarchy. I've been using it for a while now on a copied Omarchy repo, and when I saw how the omarchy-skill was implemented, I decided to submit my version.

I don't code. I don't know anything about Arch. I didn't have Linux installed at all until maybe about half a year ago. This setup allowed me to install and maintain my 3 Omarchy machines with zero effort.

What this adds:
- `omarchy-ai` command - launches OpenCode with the Omarchian agent from ~/.config/
- Dedicated agent + 4 skills (core, admin, theme, controls)
- Highly customised AGENTS.md in ~/.config/ for Omarchy context

Why this approach?
- Keeps your global OpenCode installation clean
- Agent only has access to omarchy-specific skills
- Workspace is ~/.config/ - where the agent actually operates
- Copies, not symlinks - users can customize locally as their setup evolves

Assumes OpenCode is installed.

Why these files are structured like this? Because this style works. And it works not just with Opus, it will work with smaller models you throw at it.

Any feedback, please shout.


dhh:

PR #4126, opened 2026-01-08
feat: Change color picker binding to work with 60% keyboards

ddVital:
This PR updates the color picker keybinding.

The shortcut was changed from SUPER + PRINT to SUPER + CTRL + P because most 60% keyboard layouts do not have a dedicated Print Screen key (it’s usually behind an Fn layer), making the original binding unreliable or unusable.


dhh:
Can't change existing shortcuts on people. But you can just add `bindd = SUPER CTRL, P, Color picker, exec, pkill hyprpicker || hyprpicker -a` to your own ~/.config/hypr/bindings.conf` setup.

PR #4131, opened 2026-01-08
bring back lost local hyprlock theming and save it in repo 

JustinMissmahl:
slightly more modernized lockscreen + working a bit closer with the active theme

includes hotfix from master


dhh:
Thanks for the suggestion, but I'm happy with how the hyprlock screen looks for now. But feel free, as always, to personalize your own ✌️

PR #4150, opened 2026-01-08
Fix: Jetbrains rules were not working properly anymore

NicolasDorier:
I have not been able to pinpoint what was the problem, but rewriting rules in the new style
is more beautiful and fixed my issue.

Note: I think this PR is ready to merge BUT there is still an issue documented in [this comment](https://github.com/basecamp/omarchy/pull/4150#issuecomment-3724024940).
But I think this is a problem that will be [fixed in hyprland eventually](https://github.com/hyprwm/Hyprland/discussions/12936), so this PR is probably ready to go as is.

Current workaround  is to set the `min_size` to specific values (half of your monitor) `min_size = 720 450`

### DIFF
--- a/default/hypr/apps/jetbrains.conf
+++ b/default/hypr/apps/jetbrains.conf
@@ -1,22 +1,41 @@
 # Fix splash screen showing in weird places and prevent annoying focus takeovers
-windowrule = tag +jetbrains-splash, match:class ^(jetbrains-.*)$, match:title ^(splash)$, match:float 1
-windowrule = center on, match:tag jetbrains-splash
-windowrule = no_focus on, match:tag jetbrains-splash
-windowrule = border_size 0, match:tag jetbrains-splash
+windowrule {
+    name = jetbrains-splash
+    match:class = ^(jetbrains-.*)$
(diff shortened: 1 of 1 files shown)


NicolasDorier:
so https://github.com/basecamp/omarchy/commit/fdee76e2d250a177e5bd591091eb0f3dbdb2f7bc is problematic.
I was blaming @dhh vibe code on that.

But actually, the logic is correct `min_size = (monitor_w*0.5) (monitor_h*0.5)` is exactly what need to be done for fixing the popup.

Sadly, `min_size` doesn't support expressions. Thus the line is ignored. (And without warning)

image


image

And changing to `size` is better, but still unusable.

image

Opened https://github.com/hyprwm/Hyprland/discussions/12936

NicolasDorier:
Ping @dhh @ryanrhughes 

This PR is ready to be merged IMHO

`min_size` expression support has been fixed by hyprland (https://github.com/hyprwm/Hyprland/discussions/12936), and will work as expected in the next version. (while not crashing in the current version)

I am currently using those rules with the commit that fixed it and jetbrains works like a charm.

PR #4156, opened 2026-01-08
feat: automate package cache cleanup

night-slayer18:
Automates package cache cleaning (paccache -rk2) during system updates.
- Adds 'paccache' step to bin/omarchy-update-system-pkgs
- Ensures 'pacman-contrib' is installed (dependency for paccache)

### DIFF
--- a/bin/omarchy-update-system-pkgs
+++ b/bin/omarchy-update-system-pkgs
@@ -25,3 +25,10 @@ if [[ -n $orphans ]]; then
   done
   echo
 fi
+
+# Clean package cache if paccache is available
+if command -v paccache &>/dev/null; then
+  echo -e "\e[32m\nClean package cache\e[0m"
+  paccache -rk2
+  echo
+fi
--- a/install/omarchy-base.packages
+++ b/install/omarchy-base.packages
@@ -91,6 +91,7 @@ omarchy-chromium
 omarchy-nvim
 omarchy-walker
 opencode
+pacman-contrib
 pamixer
 pinta
 playerctl


night-slayer18:
@ryanrhughes @dhh requesting for a review

severin-richner:
Like the idea and selection menu is nice. For me this would also require to install `pacman-contrib` and `bc`, so I guess these packages would need to be added.

bin/omarchy-cleanup
```diff
@@ -0,0 +1,444 @@
...
+    echo -e "  \e[32mNothing to clean\e[0m"
+    return
+  fi
+  
+  echo -e "  Packages to remove: \e[1m$pkg_count\e[0m"
+  
+  if [[ "$DRY_RUN" == "true" ]]; then
+    echo -e "  \e[33m[DRY RUN] Would run: sudo paccache -rk2\e[0m"
+    return
+  fi
```
severin-richner: Ran it with `--dry-run --all` and got following output:
```bash
Package Cache
  Current size: 1.3G
./cleanup.sh: line 98: [[: 0
0: arithmetic syntax error in expression (error token is "0")
  Packages to remove: 0
0
  [DRY RUN] Would run: sudo paccache -rk2
```

Seems like it's trying to compare the string "0" to the integer 0.

Also I'm not sure where the 0 on the second to last line is coming from.

night-slayer18: Thanks for catching this!

On line 96, I had:   
```bash
pkg_count=$(echo "$dry_output" | grep -c "\.pkg\.tar" || echo "0")
```

When `grep -c` finds zero matches, it:  
1. Outputs `0` to stdout
2. Returns exit code `1` (indicating no matches found)

Because of the exit code `1`, the `|| echo "0"` part gets triggered, which adds another `0`.

So `pkg_count` ended up containing `"0\n0"` (two zeros on separate lines) instead of just `"0"`.

Then on line 98, when bash tries to evaluate `[[ 0\n0 -eq 0 ]]`, it fails with:
```
arithmetic syntax error in expression (error token is "0")
```

Changed line 110 to:
```bash
pkg_count=$(echo "$dry_output" | grep -c "\.pkg\.tar" || true)
```

Now when `grep -c` returns `0` with exit code `1`, the `|| true` just ensures the command succeeds without adding extra output.

**Also Added:**

- Dependency checks for `bc` and `gum` at the start of the script (lines 15-25)
- Added `bc` and `pacman-contrib` to `install/omarchy-base.packages`

night-slayer18:
@dhh  @ryanrhughes  
Would appreciate maintainer feedback when you have a chance. Thanks!

ryanrhughes:
I've never been a big fan of chore utilities like this vs just having things happen via good defaults.

For example; orphaned packages are pruned on update, BTRFS snapshots are limited and pruned properly by config, journal files are self-rotating, etc. those are good defaults that just happen.

If there are some in here that are worth adding such as `paccache` to run at the end of an update cycle to prevent that cache from getting out of hand; I'm down for that but I don't think it's worth having and maintaining a dedicated utility people have to run manually.

night-slayer18:
@ryanrhughes Thanks for the feedback! I checked the codebase and you're absolutely right about the existing automation:

**Already handled by good defaults:**
- Orphaned packages - auto-removed in `bin/omarchy-update-system-pkgs` (lines 20-27)
- BTRFS snapshots - limited to 5 with space management in `install/login/limine-snapper.sh` (lines 66-69)
- Systemd journal - using systemd's built-in rotation

**Not yet automated:**
- Package cache cleanup (paccache)
- AUR build cache (`~/.cache/yay/`)

I understand the preference for automation over manual utilities, but I think there's value in having both:

**Why this utility is still useful:**

1. **Visibility & Control** - Users can see exactly what's taking up space before deciding what to clean. Even with automation, people sometimes want to manually audit their system.

2. **Safe Defaults** - Automated cleanup needs to be conservative (can't be too aggressive). This utility lets users go further when they actually need the space.

3. **Educational** - Shows users what's consuming storage on their system. Good for learning what Arch/Omarchy is doing under the hood.

4. **Dry-run mode** - Users can preview before making changes, which automation can't provide.

5. **Selective cleanup** - Interactive menu lets users choose categories. Maybe I don't want to touch snapshots today, just cache.

6. **Handles user-level cache** - Things like `~/.cache/yay/` build files and user thumbnails can't easily be automated system-wide.

**What I propose:**

1. **Add the paccache automation** you suggested (pacman hook to run after updates) - I'll add this to the PR
2. **Keep the manual utility** for the use cases above - it complements automation rather than replacing it

The utility would be optional (run via `omarchy-cleanup` when needed), while the automation handles routine maintenance. This follows the pattern of having both `omarchy-update` (automated) and manual tools like `omarchy-pkg-remove` (on-demand).

If you still feel the manual utility doesn't fit Omarchy's philosophy, I can split this into two separate PRs:
- PR - 1: Just add paccache automation
- PR - 2: Add manual cleanup utility (you can decide if it's worth including)

Let me know what you prefer!

ryanrhughes:
@night-slayer18 that sounds like a plan to me. Let's keep 'em separate so we can take the automation easily and also analyze the utility as a smaller, separate addon.

night-slayer18:
@ryanrhughes Thanks for the feedback! I've split the work as discussed:

1. **This PR**: Now focused solely on the requested automation (`paccache` hook in `omarchy-update-system-pkgs`).
2. **#4354**: Contains the manual `omarchy-cleanup` utility and menu integration as a separate add-on.

Both are ready for review!

night-slayer18:
@ryanrhughes requesting for you review please

night-slayer18:
@ryanrhughes requesting for your review please

dhh:
It's great that you're working on this, but you gotta chill with the constant requests for review. We're all just volunteering our time to work on Omarchy. This isn't an emergency fix. It's a nice-to-have.

dhh:
It needs a rebase. It needs a migration to add pacman-utils to existing installations. It needs to use omarchy-cmd-present instead of the manual `command -v` check.

PR #4174, opened 2026-01-08
v3.3.3

ryanrhughes:
Hotfix to stop gcc14 updates


Egge21M:
It seems like these where not really required by anything on the system. Why not switch to gcc from the core repo and get rid of gcc14? At least that is what I did and so far nothing exploded

ryanrhughes:
@Egge21M I believe that to be true and the right path forward too. Ignoring for now allows us a little more time to ensure there's no hidden edge we're not considering. I've removed from my systems just fine as well.

PR #4180, opened 2026-01-09
feat: implement interactive .NET version selection

BirajMainali:
**Motivation:** Installing the latest version isn't always appropriate (compatibility, LTS, legacy projects). This change lets users pick the exact stable minor/patch they need rather than forcing latest.

**What I changed:**

- Behavior: Show interactive list of stable .NET versions and install the chosen one.
- Filtering: Exclude `rc` and `preview` releases.
- Deduplication: Keep one entry per minor version (latest patch).

**Artifacts** 
image

image

image

### DIFF
--- a/bin/omarchy-install-dev-env
+++ b/bin/omarchy-install-dev-env
@@ -133,8 +133,21 @@ ocaml)
   opam install ocaml-lsp-server odoc ocamlformat utop --yes
   ;;
 dotnet)
-  echo -e "Installing .NET...\n"
-  mise use --global dotnet@latest
+  
+  # List available stable .NET versions, filtering out RC and preview versions, and sort them in descending order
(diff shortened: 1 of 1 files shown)


AndrewMast:
Might be interesting to add `--latest` flag to skip the prompt, not sure if that's scope creep though. Maybe even `--version=[version]` would be nice.

BirajMainali:
> Might be interesting to add `--latest` flag to skip the prompt, not sure if that's scope creep though. Maybe even `--version=[version]` would be nice.

I’ve condensed the current environment setup/installation flow as Install > Development > *. I’m not sure how we might handle adopting explicit versions in the TUI. Personally, I like this flow; I took inspiration from the Remove > web app > * sequence.

PR #4183, opened 2026-01-09
fix: add wayland gpu-context to mpv for proper fullscreen support

### LINKED ISSUE #4182: mpv needs 'gpu-context=wayland' for proper fullscreen support
### System details

Intel Core i3-8130U; UHD Graphics 620; Omarchy 3.3.3

### What's wrong?

# Default mpv configuration

For mpv to render correctly on wayland while on fullscreen, one needs to add `gpu-context=wayland` to its config.
This happens because of mpv defaulting to a faulty way for the GPU to manage the image, which Wayland sets on it correctly while **not** on fullscreen.

## Demonstration

I have here a visual demonstration of this, where one can appreciate the pastel colors on monkeytype being barely distinct from each other:

### Without the config fix

Image

### And with the fix applied

Image

The screenshots were taken at roughly the same frame, yet are not edited in any manner.
The image could seem _washed away_, so to speak.

## How I intend to contribute

I have made the following changes expressed in a future PR.

- Created a config file with just that one line at `config/`
- Added a unix-timestamped migration script at `migrations/` that does:
  - Check if the file exists and copies it from `config/` otherwise.
  - If incorrect lines setting the same property exist, deletes them.
  - If the line is not there then, checks if to add a newline and adds the simple line in question.

EFrMG:
Closes #4182 

It is my hope this helps! Thank you for the technical blessing the distro is.


neheb:
Downgrading to 40.7 fixes this.

I don't think it's worth bothering working around this when it'll probably get fixed upstream.

neheb:
Ugh this is a mess:

https://github.com/mpv-player/mpv/issues/17204
https://github.com/hyprwm/Hyprland/discussions/12829
https://gitlab.freedesktop.org/wayland/wayland-protocols/-/issues/305

mpv: not our bug, either hyprland or wayland-protocols.

neheb:
Passing

```
--target-colorspace-hint=no
```

to mpv is another way to handle this.

EFrMG:
> Passing
> 
> ```
> --target-colorspace-hint=no
> ```
> 
> to mpv is another way to handle this.

But you are dropping HDR for SDR.

EFrMG:
Oh, it seems that using waylandvk context throws an error for me, which would also enable HDR and is the default:
`MESA-INTEL: warning: ../mesa-25.3.3/src/intel/vulkan/anv_formats.c:993: FINISHME: support more multi-planar formats with DRM modifiers`

Well, either way. X11 does not support HDR well as well.

EFrMG:
Doing a bit of research I found this:

In 0.41.0, `vo=gpu-next` became the default and `target-colorspace-hint` changed to `auto`. The `auto` part is what could be fixed more properly, rather than with `gpu-context` as I thought.

Bug: `auto` might be incorrectly detecting Wayland as HDR-capable when it is not (yet), or it is failing to send a "End HDR" signal, leaving the screen in a desaturated state.

The older version likely defaulted to SDR tone-mapping that just happened to look better, just as `gpu-context=wayland` instead of `gpu-context=waylandvk` also does.

EFrMG:
We have two options then:

Apply a temporary fix expressed in the PR; I can monitor in upcoming updates if 
removing it makes a difference on my hardware.

Or just close the PR and wait for it to get fixed, although some might see washed-out colors on fullscreen for a while.

Inconveniently enough, mpv depends on a lot of stuff as well.

neheb:
Apparently next version of libplacebo will fix this.

In my case, I have to use Wayland instead of waylandvk so that hardware decoding works (old GPU).

EFrMG:
A more consistent solution to this and other related patches in the future, also considering how extensive Wayland is, is to add a new walker menu entry where to toggle patches.

We might be leaving users with HDR-supporting monitors with a lesser quality image in this case, so a **toggle** is ideal.

There could also be an "Experiments" section which might be ambitious but useful nonetheless.

For instance, this is in the form of: `Walker Menu -> Update -> Patches -> mpv fullscreen: On / Off`.

The main advantage would be to have a more stable and easy-to-fix system regarding common eventualities. A secondary yet important thing could be to check what the userbase wants from new experiments quickly; granted that there is some reusable logic that makes it work both ways (for developers and curious users).

Including that if experiments or patches conflict or break things, toggling them could be safer to test in most cases, as it would keep the distro more immutable **and** forward-looking.

It is an idea that may flourish someday.

EFrMG:
There is a toggle gum script version accessible under "Updates" in the Walker menu.

screenshot-2026-01-18_07-10-02

screenshot-2026-01-18_07-11-11


I left the default config for new installs to have the patch, given that 3 users have the issue already. Although that might be an oversight.

The toggle works perfectly back and forth.

EFrMG:
My apologies for the edits!

EFrMG:
A bit of info on this from: [MPV's documentation](https://mpv.io/manual/stable/). Where even if Wayland is reported as supported, `auto` does not work maybe because of other factors.

screenshot-2026-01-21_15-26-22


Another option is to use `--target-colorspace-hint-mode=source-dynamic` which works for me and another user on Discord. Although again: "This is experimental and depends on the display's ability to react to metadata changes", from the docs.

What those two flags do is essentially the same, although one operates _prior_ to the other and is more consistent (the one in the PR changes).

evenromo:
https://r-htpc.github.io/wiki/hdr

neheb:
This seems to be fixed with the latest mpv.

edit: No it's not...

my display showed a warning: Wide color gamut is enabled but the display is not in 10bit mode

neheb:
I found another way to fix this:

```
render {
  cm_fs_passthrough = 0
  cm_auto_hdr = 0
}
```

in .config/hypr/monitors.conf

PR #4195, opened 2026-01-09
Fix WiFi not working on boot

### LINKED ISSUE #4190: WiFi broken after 3.2 update: systemd-networkd disabled + rfkill race condition
## Summary

After updating to omarchy 3.2, WiFi no longer works on boot. Two separate issues require manual intervention every restart.

## System Specs

- **Device:** Desktop (AMD Ryzen AI Max 300 Series)
- **WiFi:** MediaTek MT7925 (RZ717) Wi-Fi 7 160MHz
- **Kernel:** 6.18.3-arch1-1
- **Omarchy:** 3.2

## Symptoms

On every boot, WiFi fails to connect. Manual workaround required:
```bash
sudo rfkill unblock all
sudo systemctl restart iwd
```

### Issue 1: `systemd-networkd` not enabled

`iwd` is enabled and connects to WiFi, but no IP address is obtained because `systemd-networkd` is not running to handle DHCP.

- Config files exist in `/etc/systemd/network/` (20-wlan.network, etc.)
- `iwd.service` is enabled (symlink dated Dec 16 - install date)
- `systemd-networkd.service` was never enabled

**Evidence from logs:**
```
systemd-networkd[1011]: wlan0: Connected WiFi access point: BTB-KSCGJ7
# But no "Gained carrier" or DHCP - networkd wasn't running
```

**Fix:**
```bash
sudo systemctl enable --now systemd-networkd
```

### Issue 2: rfkill race condition with MT7925

Even with networkd enabled, WiFi is soft-blocked on boot due to a race condition:

1. `systemd-rfkill` runs at boot to restore rfkill state
2. MT7925 driver loads at approximately the same time
3. Driver registers its rfkill device *after* systemd-rfkill finishes
4. WiFi remains soft-blocked

**Evidence from logs:**
```
Jan 09 16:02:27 systemd[1]: Starting Load/Save RF Kill Switch Status...
Jan 09 16:02:27 mt7925e 0000:c0:00.0: ASIC revision: 79250000  # Same second - race
```

**Fix:** udev rule to unblock WiFi when interface appears:
```bash
echo 'ACTION=="add", SUBSYSTEM=="net", KERNEL=="wlan*", RUN+="/usr/bin/rfkill unblock wifi"' | sudo tee /etc/udev/rules.d/81-wifi-unblock.rules
```

## Suggested Fixes for Omarchy

1. Enable `systemd-networkd` in the installer/update script
2. Ship `/etc/udev/rules.d/81-wifi-unblock.rules` for WiFi 7 / MediaTek compatibility

## Additional Context

This setup (iwd + systemd-networkd) is the intended architecture based on the existing config files. The issue appears to be that networkd was either never enabled or got disabled during the 3.2 update.

Sebastian-Geocento:
## Summary

Fixes #4190 (my issue) - WiFi broken after 3.2 update.

After updating to omarchy 3.2, WiFi stopped working on boot. Two issues were identified:

1. **systemd-networkd not enabled** - iwd connects to WiFi but no DHCP/IP address
2. **rfkill race condition** - WiFi soft-blocked before systemd-rfkill can restore state

## Changes

- Enable `systemd-networkd.service` alongside iwd
- Add udev rule `/etc/udev/rules.d/81-wifi-unblock.rules` to unblock WiFi when interface appears

## Testing

Both fixes have been tested on my machine and confirmed working:
- **Hardware:** Desktop (AMD Ryzen AI Max 300 Series) with MediaTek MT7925 WiFi 7
- **Kernel:** 6.18.3-arch1-1
- WiFi now works automatically on boot without manual intervention

## Open to Alternatives

This is my proposed fix based on debugging the issue. If there are better or alternative solutions, I'm open to feedback and suggestions.

### DIFF
--- a/install/config/hardware/network.sh
+++ b/install/config/hardware/network.sh
@@ -1,6 +1,15 @@
 # Ensure iwd service will be started
 sudo systemctl enable iwd.service
 
+# Enable systemd-networkd to handle DHCP for iwd connections
+sudo systemctl enable systemd-networkd.service
+
 # Prevent systemd-networkd-wait-online timeout on boot
 sudo systemctl disable systemd-networkd-wait-online.service
 sudo systemctl mask systemd-networkd-wait-online.service
+
+# Fix rfkill race condition on boot (WiFi soft-blocked before systemd-rfkill restores state)
+# This ensures WiFi is unblocked when the interface appears, fixing issues with MediaTek MT7925 and similar
+if [[ ! -f /etc/udev/rules.d/81-wifi-unblock.rules ]]; then
+  echo 'ACTION=="add", SUBSYSTEM=="net", KERNEL=="wlan*", RUN+="/usr/bin/rfkill unblock wifi"' | sudo tee /etc/udev/rules.d/81-wifi-unblock.rules
+fi


GitUser-IamALIVE:
does "rfkill unblock wlan" doesnt work for you ?

Sebastian-Geocento:
I did try it and no, on boot I had the same issue

dhh:
Thanks for the detailed debugging and PR! A couple of concerns:

**On enabling `systemd-networkd`:** Omarchy uses iwd in standalone mode, which has its own built-in DHCP client. Enabling `systemd-networkd` alongside it introduces a second network management layer that could cause conflicts for other users. If iwd wasn't doing DHCP for you, the root cause is likely an iwd configuration issue (e.g. `EnableNetworkConfiguration=true` in `/etc/iwd/main.conf`) rather than a missing systemd-networkd service. I'd drop this change.

**On the rfkill udev rule:** This is a more reasonable fix, but force-unblocking wifi on every boot for all users overrides intentional rfkill state and applies a hardware-specific workaround universally. This feels better suited as a documented troubleshooting step rather than a default for everyone.

Would you be able to check whether your iwd config has `EnableNetworkConfiguration=true` set? That alone might solve the DHCP side without needing systemd-networkd at all.

PR #4198, opened 2026-01-09
Add ~/.bashrc to setup menu

Tsurgcom:
To Bash, or not to Bash, that is the question;


dhh:
Appreciate the suggestion, but think we should assume some basic familiarity with the system here. There are a ton of configs we could link up, but I'm trying just to focus on the ones that might be novel to folks and particular to Omarchy ✌️

PR #4201, opened 2026-01-10
Black Screen after laptop wake from sleep NVIDIA

### LINKED ISSUE #4200: Black screen after laptop wake from sleep on NVIDIA GPUs
### System details

Framework Overkill, AMD Ryzen AI 9 HX 370, NVIDIA Gefore RTX 5070

### What's wrong?

When I close and reopen the laptop lid (suspend/resume cycle), the screen stays black. Sometimes requires closing and reopening the lid 2-3 times before the display works again. Not super consistent

I think It's NVIDIA causing it. I haven't made a fix like this to something like an distro but I'll give it a shot

NVIDIA's driver defaults to only saving "essential" video memory allocations during suspend. When the system resumes, the GPU's video memory is in an undefined state, causing the black screen.

From NVIDIA's official driver documentation (Chapter 21 - Power Management)(https://download.nvidia.com/XFree86/Linux-x86_64/565.57.01/README/powermanagement.html)

> "The GPU state saved by the NVIDIA kernel drivers includes allocations made in video memory. However, these allocations are collectively large, and typically cannot be evicted. Since the amount of system memory available to drivers at suspend time is often insufficient to accommodate large portions of video memory, **the NVIDIA kernel drivers are designed to act conservatively, and normally only save essential video memory allocations.**"

> "The resulting loss of video memory contents is partially compensated for by the user-space NVIDIA drivers, and by some applications, but **can lead to failures such as rendering corruption and application crashes upon exit from power management cycles.**"

## Solution

Add `NVreg_PreserveVideoMemoryAllocations=1` to `/etc/modprobe.d/nvidia.conf`:

```
options nvidia NVreg_PreserveVideoMemoryAllocations=1
```

Then regenerate initramfs and reboot.

From NVIDIA's docs:

> "the NVIDIA Linux kernel module "nvidia.ko" needs to be loaded with the **"NVreg_PreserveVideoMemoryAllocations=1" module parameter. This changes the default video memory save/restore strategy to save and restore all video memory allocations.**"

## References

- [NVIDIA Driver Documentation - Chapter 21: Power Management](https://download.nvidia.com/XFree86/Linux-x86_64/435.17/README/powermanagement.html)
- [Arch Wiki - NVIDIA Tips and Tricks - Preserve video memory after suspend](https://wiki.archlinux.org/title/NVIDIA/Tips_and_tricks#Preserve_video_memory_after_suspend)

visch:
Closes #4200

Haven't made a PR like this before, in a Distro so forgive me if something is off here! Love what you are all doing with Omarchy this has been a blast using as my daily driver!

Enable full video memory preservation by adding `NVreg_PreserveVideoMemoryAllocations=1` and regenerating initramfs for existing NVIDIA machines

refs: https://download.nvidia.com/XFree86/Linux-x86_64/435.17/README/powermanagement.html , https://wiki.archlinux.org/title/NVIDIA/Tips_and_tricks#Preserve_video_memory_after_suspend

### DIFF
--- a/install/config/hardware/nvidia.sh
+++ b/install/config/hardware/nvidia.sh
@@ -19,9 +19,10 @@ if [ -n "$NVIDIA" ]; then
 
   omarchy-pkg-add "$KERNEL_HEADERS" "${PACKAGES[@]}"
 
-  # Configure modprobe for early KMS
+  # Configure modprobe for early KMS and suspend/resume support
   sudo tee /etc/modprobe.d/nvidia.conf <<EOF >/dev/null
 options nvidia_drm modeset=1
+options nvidia NVreg_PreserveVideoMemoryAllocations=1
 EOF
 
   # Configure mkinitcpio for early loading
--- a/migrations/1768012038.sh
+++ b/migrations/1768012038.sh
@@ -1,0 +1,30 @@
+echo "Add NVIDIA suspend/resume video memory preservation"
+
+# Only apply to systems with NVIDIA GPU
+NVIDIA="$(lspci | grep -i 'nvidia')"
+if [ -z "$NVIDIA" ]; then
+  exit 0
+fi
+
+NVIDIA_CONF="/etc/modprobe.d/nvidia.conf"
+
+# Check if the config already has the preserve setting
+if [ -f "$NVIDIA_CONF" ] && grep -q "NVreg_PreserveVideoMemoryAllocations=1" "$NVIDIA_CONF"; then
+  echo "NVIDIA video memory preservation already configured"
+  exit 0
+fi
+
+# Add or update the nvidia.conf file
+sudo tee "$NVIDIA_CONF" <<EOF >/dev/null
+options nvidia_drm modeset=1
+options nvidia NVreg_PreserveVideoMemoryAllocations=1
+EOF
+
+echo "Regenerating initramfs..."
+if omarchy-cmd-present limine-update; then
+  sudo limine-update
+elif command -v mkinitcpio &>/dev/null; then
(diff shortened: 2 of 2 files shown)


GitUser-IamALIVE:
which NVIDIA card are you using and what is the version of NVIDIA driver you are using ?
because on nvidia-open and rtx 4070 i dont have this issue on both my pc and laptop , also kargs and NVIDIA module settings are already handled by nvidia-utils on arch Linux and this karg is not needed.

visch:
> which NVIDIA card are you using and what is the version of NVIDIA driver you are using ?
> because on nvidia-open and rtx 4070 i dont have this issue on both my pc and laptop , also kargs and NVIDIA module settings are already handled by nvidia-utils on arch Linux and this karg is not needed.

GPU: NVIDIA GeForce RTX 5070 Laptop GPU
Driver Version: 590.48.01
Driver Package: nvidia-open-dkms 590.48.01-1

Maybe Blackwell specific?

This karg fixes my problem though? Maybe I'm missing something?

ryanrhughes:
See my comment here. https://github.com/basecamp/omarchy/issues/4200#issuecomment-3905313212 

This is already set by default by Arch so adding to the conf isn't necessary.

visch:
> See my comment here. [#4200 (comment)](https://github.com/basecamp/omarchy/issues/4200#issuecomment-3905313212)
> 
> This is already set by default by Arch so adding to the conf isn't necessary.

Sorry I was wrong here! I think some recent tweaks in Omarchy addressed this for me as I had some other fixes in place too.

If I get anything concrete again I'll try to be more specific. Sorry for the chasing here

PR #4217, opened 2026-01-10
Enable clipboard sharing for Windows VM

arcangelo7:
Currently clipboard sharing doesn't work when connecting to the Windows VM via RDP.

This adds the `/clipboard` flag to the xfreerdp3 command so you can copy/paste between Windows and Omarchy.

Tested on Omarchy 3.3.3, works in both directions.


ryanrhughes:
@arcangelo7  Were you able to successfully copy from Windows and paste in Omarchy?

I'm able to copy from Omarchy and paste in Windows, which is a plus, but not the other way around.

PR #4219, opened 2026-01-10
Add 5G/LTE modem support in waybar

mmorri:
## Summary
Adds waybar integration for 5G/LTE cellular modems using ModemManager.

- **omarchy-5g-status**: Waybar module showing connection status and signal strength
- **omarchy-launch-5g**: Fuzzel-based menu for connect/disconnect
- Auto-hides when no modem is detected (safe for all systems)

## Screenshots
The module appears between Bluetooth and WiFi icons, showing signal strength with cellular icons.

## Test plan
- [x] Tested on Samsung Galaxy Book 935QDC with MediaTek T700 5G modem
- [x] Module hides correctly when no modem present
- [x] Connect/disconnect workflow works via fuzzel menu
- [x] Signal strength updates every 5 seconds

## Dependencies
- `modemmanager` - for modem control
- `fuzzel` - for the connect/disconnect menu

### DIFF
--- a/bin/omarchy-5g-status
+++ b/bin/omarchy-5g-status
@@ -1,0 +1,57 @@
+#!/bin/bash
+# Waybar module for 5G/LTE modem status
+# Returns JSON for waybar custom module
+
+MODEM=$(mmcli -L 2>/dev/null | grep -oP '/Modem/\K\d+' | head -1)
+
+if [ -z "$MODEM" ]; then
+    echo '{"text": "", "tooltip": "No modem found", "class": "hidden"}'
+    exit 0
+fi
+
+# Strip ANSI color codes from output
+STATUS=$(mmcli -m "$MODEM" 2>/dev/null | sed 's/\x1b\[[0-9;]*m//g')
+STATE=$(echo "$STATUS" | grep -P '^\s+\|\s+state:' | grep -oP 'state:\s+\K\w+')
+SIGNAL=$(echo "$STATUS" | grep -oP 'signal quality:\s+\K\d+')
+OPERATOR=$(echo "$STATUS" | grep -oP 'operator name:\s+\K.+' | head -1)
+ACCESS=$(echo "$STATUS" | grep -oP 'access tech:\s+\K\w+')
+
+# Get bearer info if connected
+BEARER=$(echo "$STATUS" | grep -oP 'Bearer\s+\|.*paths:.*Bearer/\K\d+' | tail -1)
+if [ -n "$BEARER" ]; then
+    IP=$(mmcli -b "$BEARER" 2>/dev/null | sed 's/\x1b\[[0-9;]*m//g' | grep -oP 'address:\s+\K[\d.]+')
+fi
+
+# Default signal if not available
+[ -z "$SIGNAL" ] && SIGNAL=0
+
+# Select icon based on signal strength and connection state
+if [ "$STATE" = "connected" ]; then
+    if [ "$SIGNAL" -ge 80 ]; then
+        ICON="󰀳"
+    elif [ "$SIGNAL" -ge 60 ]; then
+        ICON="󰀵"
+    elif [ "$SIGNAL" -ge 40 ]; then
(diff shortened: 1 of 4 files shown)


jheuing:
Hey @mmorri ! Got any insights how you got your 5g working? I am on a Thinkpad X1 and wasn't actually successful. I also tried using modemmanager. I'll give it another try the other week (will check out fuzzel).

The waybar support is great. I'll add it then to my custom setup. We'll see if it makes it into the standard.

mmorri:
Just seeing this! So, the 5G was kind of working out of the box once I wrote the applet to activate it, I didn't have to fight with drivers. What is the issue you are seeing for the X1? happy to have a look, I have an X13s with 5G that I think might have the exact same chip

jheuing:
@mmorri actually it might be a sim issue I have figured. the system reports a "sim-pin2"-lock. which seems to be an "old" thing which the driver can not remove, but windows would remove automatically on first connect. checking family laptops at the weekend if they have sim slots and would change it automatically on windows - then I'll go back to see if it made a difference. thanks loads for coming back to me on this. I'll see after the weekend if I have any progress then :-) my carrier support said they have no button for removing it in their gui ;)

mmorri:
Oh I have the same issue with sim-lock pin2 on the X13s! It works without problems on windows, I never manage to get it to work on Linux, any distribution. Good luck!

Sent from [Proton Mail](https://proton.me/mail/home) for Android.

-------- Original Message --------

jheuing:
It said it is about disabling PIN2 / FDN, which can be done carrier side. No chance with my carrier, they don't know even what I am talking about.

There seem to be some android devices out there, which can potentially disable PIN2 /  FDN, but not the one testing device I have. 

So it seems... no chance :-) I see no option at all getting my sim card to work at this stage because of software issues taking SIM2 / FDN for granted, while windows and routers totally ignore that setting (sim works fine then)

skibidiandulka:
I am not sure if this is really needed, because I doubt that most of the users would use this. So it would probably just be excess code for most people. I think this is great tho and you should move it to discussions->Show & Tell.

mmorri:
@jheuing Hey! Got an update — I finally figured out the issue with the modem getting stuck.

**The problem:** The modem stays in `enabling` state with `power state: low` and throws `Invalid transition` errors. The `sim-pin2` lock that shows up is actually a red herring — the real blocker is **FCC regulatory authentication**. US-sold modems require an FCC unlock before they can power on fully, and ModemManager doesn't run it automatically unless you configure it.

**The fix:**

Since we have the same modem (Foxconn/Qualcomm, vendor ID `105b`), ModemManager already ships with the right unlock script. You just need to enable it:

```bash
sudo mkdir -p /etc/ModemManager/fcc-unlock.d
sudo ln -s /usr/share/ModemManager/fcc-unlock.available.d/105b /etc/ModemManager/fcc-unlock.d/105b
sudo systemctl restart ModemManager
```

That's it. The modem should unlock and connect automatically on every boot from now on — no manual intervention needed.

If you want to test it manually first before restarting:
```bash
sudo /usr/share/ModemManager/fcc-unlock.available.d/105b /org/freedesktop/ModemManager1/Modem/0 wwan0mbim0
```

Hope this unblocks you!

mmorri:
> I am not sure if this is really needed, because I doubt that most of the users would use this. So it would probably just be excess code for most people. I think this is great tho and you should move it to discussions->Show & Tell.

I will do this, thanks for your kind comment

mmorri:
@skibidiandulka Good idea — posted it as a Show & Tell discussion here: https://github.com/basecamp/omarchy/discussions/4823

mmorri:
Quick update: I moved this into a separate AUR package as well:

https://aur.archlinux.org/packages/samsung-galaxybook-5g

Since the need is niche, this should make any eventual Omarchy integration much easier: Omarchy could depend on or optionally install the package without carrying all the modem/FCC-unlock and UI helper code directly.

dhh:

PR #4220, opened 2026-01-10
Add Samsung Galaxy Book audio fix

mmorri:
## Summary
Fixes speaker audio on Samsung Galaxy Book laptops with Realtek ALC298 codec.

**The problem:** Samsung laptops with ALC298 have speaker amplifiers that aren't initialized by the Linux kernel, resulting in no audio from internal speakers.

**The solution:**
- `omarchy-samsung-audio-init`: HDA verb script that initializes the speaker amp
- `omarchy-setup-samsung-audio`: One-time setup that installs systemd services and WirePlumber config

## Usage
```bash
omarchy-setup-samsung-audio
```

This will:
1. Install `alsa-tools` if needed
2. Create systemd services for boot and suspend resume
3. Configure WirePlumber to prevent audio suspend (which loses amp init)
4. Initialize the audio immediately

## Test plan
- [x] Tested on Samsung Galaxy Book 935QDC
- [x] Audio works after boot
- [x] Audio persists after suspend/resume
- [x] Audio persists when pausing/resuming media playback


dhh:
We should put this is a separate AUR package, so all Arch distros could benefit. Then we can install it by default in Omarchy.

mmorri:
Thanks for the attention and direction, DHH. I moved this into a separate AUR package so other Arch users can benefit too:

https://aur.archlinux.org/packages/samsung-galaxybook-audio

I’d love to keep contributing to Omarchy.

PR #4221, opened 2026-01-10
Add Yoru Swirl Buck wallpaper to Tokyo Night theme

Maxteabag:
## Summary

Adds a new wallpaper for Tokyo night theme featuring abstract swirling concentric waves.

## Details

- **Resolution:** 7680x3215
- **Format:** 8-bit indexed PNG  
- **Size:** 7.9MB

Created in response to DHH's request for a high-quality version of the swirl buck design.


dhh:
I really like the image, but it still looks low quality when zoomed all the way up on my 6K screen. I take it that the root here is AI generated and then upscaled? I think we need a way to turn this into a vector graphics image first, like I'm sure the original is.

Here's a zoomed example of what I see at 6K:

image

You can see there's a lot of compression noise. The lines aren't straight.

Maxteabag:
> I really like the image, but it still looks low quality when zoomed all the way up on my 6K screen. I take it that the root here is AI generated and then upscaled? I think we need a way to turn this into a vector graphics image first, like I'm sure the original is.

Yeah It's gemini, I upscaled it but I messed up the compression on the first attempt. I updated the PR with a new version which should look much better, albeit I can see it's not perfect if you zoom all the way in.

Vectorization is tricky - the image has subtle gradients. I've been experimenting with different approaches, and it seems to be doable although it would change the overall style a bit.

For now, the updated PNG should look much better on your 6K, let me know if it's good enough. If not I'll delve into the vectorization.

dhh:
Bam! That worked. Even after I converted it from PNG to JPG to get the size down. It remains sharp (enough) to look great on my 6K screen. I've added it as the new first wallpaper for Tokyo Night. Excellent, beautiful work 🤘❤️

PR #4225, opened 2026-01-11
omarchy theme sync

beaterblank:
## Summary

When the active Omarchy theme changes, this automatically applies user-defined configuration files to a target directory, replacing placeholders with values from the current theme’s `colors.toml`, uses default `dark.toml`(catppuccin) `light.toml`(catppuccin-latte) if not found.

I’ve  made it as a background service in this [repository](https://github.com/beaterblank/omarchy-theme-sync-ctl). I originally created it so that my TUI application would stay in sync with theme colors and also to help for #1548

## Usage

1. Create a configuration directory under:

```
~/.config/omarchy-theme-sync/
```

2. Inside it, create a folder for each application.

3. Within the application folder:

   * Create a file named `dir` containing the destination path where the processed configuration files should be written.
   * Add one or more configuration files that include placeholders matching keys from `colors.toml`.

All placeholders in the form `${variable}` are automatically detected and replaced using values from the active theme’s `colors.toml`.

### Example

**Directory structure**

```
~/.config/omarchy-theme-sync/zen-cal/
├── dir
└── zen-cal.conf
```

**`dir`**

```
$HOME/.config/zen-cal
```

**`zen-cal.conf`**

```ini
# Zen-Cal color configuration
today    = ${selection_background}
headings = ${accent}
text     = ${foreground}
weekends = ${color2}
```

When the theme changes, the processed configuration is written to `$HOME/.config/zen-cal` with all color variables resolved.

### DIFF
--- a/bin/omarchy-theme-sync
+++ b/bin/omarchy-theme-sync
@@ -4,7 +4,6 @@ set -euo pipefail
 CONFIG_DIR="$HOME/.config"
 THEME_DIR="$CONFIG_DIR/omarchy/current"
 SYNC_DIR="$CONFIG_DIR/omarchy-theme-sync"
-THEME_STORE="$HOME/.local/share/omarchy/themes"
 
 COLORS_FILE=""
 


dhh:
Interesting idea, but we already look for custom themed config files in `~/.config/omarchy/themed`. You can add any .tpl file, and the replacement of variables will be made, so `~/.config/omarchy/themed/myconfig.cfg.tpl` ends up in `~/.config/omarchy/current/theme/myconfig.cfg`. And then we also have the `~/.config/omarchy/hooks/theme-set` hook that's called on theme change if you want something else 👍

beaterblank:
This approach is particularly useful for applications that require config files to be in a fixed location, such as Krita, where custom themes are stored in `~/.local/share/krita/colors`. I’m also working on a configuration for the Zen browser, which as well, so far as I know, similarly doesn’t allow specifying a custom theme location. That said, I understand this might be a more niche use case for the general repository.

Thank you!

PR #4227, opened 2026-01-11
Enable scanner support

daviewales:
Install the Gnome Document Scanner frontend (simple-scan), which automatically includes the required dependencies for scanner support.

Enable support for modern 'driverless' scanners with sane-airscan.

Note: This adds 47.23 MiB to the base install size.

See: https://wiki.archlinux.org/title/SANE

Fixes #4102.

### DIFF
--- a/install/omarchy-base.packages
+++ b/install/omarchy-base.packages
@@ -106,9 +106,11 @@ qt5-wayland
 ripgrep
 ruby
 rust
+sane-airscan
 satty
 sddm
 signal-desktop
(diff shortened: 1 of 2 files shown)


dhh:
Thanks for the suggestion, but I don't think enough people have dedicated scanners these days to warrant the extra 50mb on the default install. But would be good to have a page in the manual about adding these drivers for scanners. Maybe start something in Discussions that gather that and other common accessories like that 👌

PR #4231, opened 2026-01-11
Add Running Buck wallpaper to Catppuccin theme

Maxteabag:
## Summary

Adds a new wallpaper for Catppuccin theme featuring a buck silhouette running downhill with dramatic spiral vortex clouds.

## Details

- **Resolution:** 7680x3215
- **Format:** 8-bit indexed PNG
- **Size:** 8.8MB


dhh:
I really do like this buck theme for Omarchy, but I don't think the colors quite match Catppuccin. And we still haven't been able to recreated the vector sharpness of the original yet either. Thanks for continuing to try here!

PR #4245, opened 2026-01-13
add btm to floating TUI rule

neheb:
Alternative to btop.


neheb:
Main advantage over btop for me is GPU support with the amdgpu driver. The former requires rocm libraries and only works with newer GPUs.

dhh:
We only keep things in there that are part of the default toolkit. Otherwise it could really balloon. So probably best just to add it directly to your user config when you install btm ✌️

PR #4247, opened 2026-01-13
Add webcam source selection for screen recording

incpo:
I had trouble selecting my second webcam when I tried to start a screen recording. To fix it, I edited the code a bit and added a camera selection menu when more than one camera device is connected

- Add --webcam-device argument to omarchy-cmd-screenrecord
- Auto-detect first available webcam if none specified
- Add webcam selection UI in omarchy-menu when multiple cameras available
- Skip selection UI when only one webcam is detected

I don’t want to deal with how to record the screen without a capture card, because obs doesn’t show the window menu. So this is just a demo recorded on an iphone

https://github.com/user-attachments/assets/ad7a7b83-1e57-4d13-941d-009495357f02

### DIFF
--- a/bin/omarchy-cmd-screenrecord
+++ b/bin/omarchy-cmd-screenrecord
@@ -14,6 +14,7 @@ fi
 DESKTOP_AUDIO="false"
 MICROPHONE_AUDIO="false"
 WEBCAM="false"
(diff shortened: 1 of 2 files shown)


dhh:
Very nice!

PR #4250, opened 2026-01-13
auto toggle scratchpad on switching workspace

mitanjan:
The **default behaviour** when switching workspace from scratchpad should be to

- **auto toggle scratchpad and show the main workspace**

Reason

**why would anyone want to switch workspace if he still want to see the scratchpad after switching workspace, if the user switches the workspace the scratchpad should toggle and the main workspace should be visible.**

This PR uses an already defined variable in hyprland to acheive this.

### DIFF
--- a/default/hypr/looknfeel.conf
+++ b/default/hypr/looknfeel.conf
@@ -131,6 +131,11 @@ cursor {
     hide_on_key_press = true
 }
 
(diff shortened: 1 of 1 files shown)


pomartel:
I just tested it and it works great. A little detail that makes  hyprland more enjoyable!

PR #4255, opened 2026-01-13
fix invalid skill name

haoqixu:
The `name` field contains invalid characters (must be lowercase a-z, 0-9, hyphens only)

ref: https://agentskills.io/specification#name-field


dhh:
Already fixed on dev 👍

PR #4260, opened 2026-01-14
Fix Bluetooth device showing bluez ID instead of friendly name

shawnyeager:
## Summary

- Fixes regression where Bluetooth devices (e.g., AirPods Max) show raw bluez ID like `bluez_output.90:9C:4A:EE:01:1C` instead of friendly name when switching audio sources with `Super+Mute`

## Root Cause

For Bluetooth audio devices, the friendly name is stored on the **Device** entry in wpctl (accessible via `device.id`), not the **Sink** entry (accessible via `object.id`). The previous fix looked up `object.id`, which returns the sink line showing the raw bluez ID.

## Fix

Look up `device.id` first to get the friendly name from the Device entry, falling back to `object.id` for non-Bluetooth devices.

## Test plan

- [ ] Connect Bluetooth headphones (AirPods, etc.)
- [ ] Press `Super+Mute` to cycle audio outputs
- [ ] Verify the friendly device name appears instead of `bluez_output.XX:XX:XX...`

### DIFF
--- a/bin/omarchy-cmd-audio-switch
+++ b/bin/omarchy-cmd-audio-switch
@@ -28,8 +28,16 @@ next_sink_name=$(echo "$next_sink" | jq -r '.name')
 
 next_sink_description=$(echo "$next_sink" | jq -r '.description')
 if [ "$next_sink_description" = "(null)" ] || [ "$next_sink_description" = "null" ] || [ -z "$next_sink_description" ]; then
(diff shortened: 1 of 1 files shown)


dhh:
I can't replicate the problem with a pair of airpods via bluetooth. They already show with the right name. What are you seeing for that?

shawnyeager:
For my AirPods Max, the above: `bluez_output.90:9C:4A:EE:01:1C`.

I applied this fix and have had no problem since.

PR #4265, opened 2026-01-14
fix: starship git status

andrewf403:
Starship prompt was missing "up-to-dateness" statuses of the repository: ahead, behind, up to date.
Starship doc states that `$ahead_behind` should be included for that in addition to `$all_status`

Because formatting for `ahead`, `behind`, and `diverged` was included in config, I assumed that they were meant to be shown. 

_If I'm mistaken, and current behavior is intended, please close this PR._

Also, added "+" status when all changes are staged, because currently clean working tree and one with staged changes look the same in a prompt, IMO it is confusing. More so because just moment before staging them, unstaged changes had indicator in the prompt.

screenshot-2026-01-15_00-26-19

### DIFF
--- a/config/starship.toml
+++ b/config/starship.toml
@@ -17,7 +17,7 @@ format = "[$branch]($style) "
 style = "italic cyan"
 
 [git_status]
-format     = '[$all_status]($style)'
+format     = '[$all_status$ahead_behind]($style)'
 style      = "cyan"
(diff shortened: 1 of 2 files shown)


dhh:
The omission was actually intentional. I narrowed down what's being shown to just what I felt was more pertinent. The full set of all the permutations is imo too much. Appreciate the suggestion, though!

PR #4269, opened 2026-01-15
Sanitize prompt before starship init

foucist:
### Summary

Adds prompt sanitization to clear stale terminal line state after abnormal exits, preventing prompt artifacts.

In some cases, exiting a program abruptly can leave residual output on the current terminal line, causing the next prompt to render incorrectly. This change ensures the prompt is always drawn on a clean line.

---

### Example

```sh
iex
# press Ctrl-\
```

Exiting `iex` with `Ctrl-\` sends `SIGQUIT`, which terminates the process without restoring terminal display state. When control returns to the shell, leftover output may remain on the line.

This behavior is not specific to IEx; any program that exits without cleaning up the terminal line can trigger it.

---

### Fix

Starship assumes a clean terminal line when rendering the prompt and does not perform line hygiene itself. Before rendering the prompt, the shell now resets the cursor position and clears the current line, preventing prompt corruption regardless of how the previous process exited.

### DIFF
--- a/default/bash/init
+++ b/default/bash/init
@@ -3,6 +3,9 @@ if command -v mise &> /dev/null; then
 fi
 
 if command -v starship &> /dev/null; then
+  # clear stale readline state before rendering prompt (prevents artifacts in prompt after abnormal exits like SIGQUIT)
+  __sanitize_prompt() { printf '\r\033[K'; }
+  PROMPT_COMMAND="__sanitize_prompt${PROMPT_COMMAND:+;$PROMPT_COMMAND}"
   eval "$(starship init bash)"
 fi
 


dhh:
This has a worse side effect: #4783. Reverting.

foucist:
My bad! As it turns out, there’s a good reason for Starship’s default `add_newline = true`, which always prints a newline before the prompt.

With PS1-style prompts in bash and zsh, the line editor (Readline / ZLE) owns the line state and effectively “sanitizes” it, so there’s no need for an extra newline. I was trying to make Starship feel more like that, but since it just writes to stdout without access to the internal line state, that led me to a hacky workaround that only made things worse. 

Sorry about that!

It seems the only reliable way to guarantee a clean prompt is to accept the newline default (or switch back to PS1).

PR #4277, opened 2026-01-15
Toggle hybrid GPU mode via supergfxctl

dhh:
Very useful for Asus G14 and other laptops with NVIDIA + AMD iGPU combos. Turning off the DGPU can save a lot of power.


bin/omarchy-toggle-hybrid-gpu
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash
+
+# Toggle dedicated vs integrated GPU mode via supergfxd (for hybrid gpu laptops, like Asus G14).
+# Requires reboot to take effect.
+
+# Ensure supergfxctl has been installed
+if omarchy-cmd-missing supergfxctl; then
+  omarchy-pkg-aur-add supergfxctl
```
neheb, line 9: Would it make sense to add https://wiki.archlinux.org/title/Unofficial_user_repositories#g14 instead?

dhh: We will add this to OPR instead 👍

PR #4286, opened 2026-01-16
Add "Without audio" option to screenrecord menu

robzolkos:
Sometimes you just need a quick, silent screen recording - like creating a short demo video to attach to a GitHub pull request, documenting a UI bug, or recording a gif-like clip where audio isn't needed.  

Adds a "Without audio" option to the screenrecord menu for capturing silent video

This option existed before the screen capture redesign in https://github.com/basecamp/omarchy/commit/128a6125a92d but was removed as part of simplifying the menu. This PR brings it back as the last menu option.

no-audio


sean-imus:
i listen to lots of music and having to strip the audio everytime i want to share a small snippet is really annoying as its not possible everywhere to set the clip to mute from the upload interface. This is a fix for that :+1:

PR #4287, opened 2026-01-16
Add migration to rebuild python-terminaltexteffects for Python 3.14

### LINKED ISSUE #4263: missing module TerminalTextEffects after update
### System details

4060, i7, latest update

### What's wrong?

After updating omarchy to the latest v, FSR TerminalTextEffects module is missing when screensaver kicks in? I have no idea if this is a skill issue or if this happened because of the latest update.

edit:
`python-terminaltexteffects 0.14.2-1` is still installed, yet:

```
❯ tte
Traceback (most recent call last):
  File "/usr/bin/tte", line 5, in <module>
    from terminaltexteffects.__main__ import main
ModuleNotFoundError: No module named 'terminaltexteffects'
~
```

it's because of my py version. It upgraded it to 3.14 on the latest update, causing the module not to be found.

billabel:
Fixes issue where python-terminaltexteffects fails with ModuleNotFoundError after system upgrade to Python 3.14. The package was built for Python 3.13 and installed to /usr/lib/python3.13/site-packages, but Python 3.14 only searches /usr/lib/python3.14/site-packages.

The migration removes the outdated OPR package and reinstalls from AUR, forcing a rebuild with the current Python version.

Fixes #4263

### DIFF
--- a/migrations/1768585821.sh
+++ b/migrations/1768585821.sh
@@ -1,0 +1,6 @@
+echo "Rebuild python-terminaltexteffects for Python 3.14 compatibility"
+
+if omarchy-pkg-present python-terminaltexteffects; then
+  omarchy-pkg-drop python-terminaltexteffects
+  omarchy-pkg-aur-add python-terminaltexteffects
+fi


ryanrhughes:
TTE has already been rebuilt on edge mirror on `python 3.14`. On stable, python is still `python 3.13` so it should remain that way until next release.

PR #4292, opened 2026-01-17
Add patch to /etc/updatedb.conf to include /home directory

### LINKED ISSUE #3744: plocate does not index $HOME on Btrfs subvolume setups unless PRUNE_BIND_MOUNTS="no"
### System details

Omarchy 3.2

### What's wrong?

On Btrfs systems locate (plocate) fails to index anything in $HOME. Running updatedb completes without errors, but locate returns no results for files inside the user’s home directory.

The cause is that Arch (and Omarchy, inheriting from Arch) ships:

```
PRUNE_BIND_MOUNTS="yes"
```

in /etc/updatedb.conf

Btrfs subvolumes appear as bind mounts, so updatedb prunes them even though they are real subvolumes.

Setting it to:

```
PRUNE_BIND_MOUNTS="no"
```

immediately fixes the issue. This might be a bit crude and I'm not sure whether it might have unintended side-effects for some users. 

Another solution might be to create a different db for home (see https://discussion.fedoraproject.org/t/mlocate-does-not-find-files-in-home-on-fedora-33-btrfs/75073) ? 

I'd be happy to contribute a solution myself, but without a DISTRIBUTORS.md i'm a bit lost at the moment.

jorge-campo:
## Problem description
New Omarchy Linux installations currently ship with the following value in in `/etc/updatedb.conf`:

```bash
PRUNE_BIND_MOUNTS = "yes"
```

With this value, the `updatedb` run used by `locate`/`plocate` prunes bind mounts, which results in paths under `/home` being omitted from the locate database. As a consequence, searches via `locate`/`plocate` do not return matches for files located in users’ home directories.

This PR updates the default configuration to:

```bash
PRUNE_BIND_MOUNTS = "no"
```

so that /home contents are indexed and searchable by default after installation.

### Steps to reproduce the issue

1. Confirm current config: `grep -n 'PRUNE_BIND_MOUNTS' /etc/updatedb.conf`. Expected to show: `PRUNE_BIND_MOUNTS = "yes"`.
2. Create a test file under `/home`: `touch /home/$USER/locate-test-file`.
3. Rebuild the `locate`/`plocate` database: `sudo updatedb`.
5. Search for the file: `locate locate-test-file` (or `plocate locate-test-file`). Notice the file is missing from results.

### Summary of changes
During installation, before running `updatedb` in `install/config/localdb.sh`, the script now enforces `PRUNE_BIND_MOUNTS = "no"` in `/etc/updatedb.conf` (whitespace-tolerant: updates the existing setting if present, otherwise appends it), then rebuilds the locate database as usual.

### Testing
1. Verified the updated `localdb.sh` script on an existing Omarchy 3.3.3 installation: the script executed successfully with no errors and `updatedb` completed as expected.
2. Built and installed a new Omarchy ISO from the `omarchy-iso` repository using:

   ```bash
   OMARCHY_INSTALLER_REPO="jorge-campo/omarchy" OMARCHY_INSTALLER_REF="fix-localdb-prune-bind-mounts" ./bin/omarchy-iso-make
   ```
   Installation completed without issues using this ISO image, and after the install, `/etc/updatedb.conf` contained the updated `PRUNE_BIND_MOUNTS = "no"` value.

Screenshot from the installation test on a virtual machine:

image

> [!NOTE]
>
> This PR was prepared with support from GitHub Copilot.

closes #3744

### DIFF
--- a/install/config/localdb.sh
+++ b/install/config/localdb.sh
@@ -1,2 +1,17 @@
 # Update localdb so that locate will find everything installed
+
+UPDATEDB_CONF="/etc/updatedb.conf"
+
+if [[ -f "$UPDATEDB_CONF" ]]; then
+  # Ensure /home (and other non-pruned mounts) are included in the locate database by default.
+  # Be tolerant of whitespace: PRUNE_BIND_MOUNTS = "yes" / PRUNE_BIND_MOUNTS="yes" / etc.
+  if grep -qE '^[[:space:]]*PRUNE_BIND_MOUNTS[[:space:]]*=' "$UPDATEDB_CONF"; then
+    sudo sed -i -E 's|^[[:space:]]*PRUNE_BIND_MOUNTS[[:space:]]*=.*$|PRUNE_BIND_MOUNTS = "no"|' "$UPDATEDB_CONF"
+  else
+    echo 'PRUNE_BIND_MOUNTS = "no"' | sudo tee -a "$UPDATEDB_CONF" >/dev/null
+  fi
+else
+  echo "Warning: $UPDATEDB_CONF not found; skipping PRUNE_BIND_MOUNTS configuration" >&2
+fi
+
 sudo updatedb


install/config/localdb.sh
```diff
@@ -1,2 +1,17 @@
...
+  # Ensure /home (and other non-pruned mounts) are included in the locate database by default.
+  # Be tolerant of whitespace: PRUNE_BIND_MOUNTS = "yes" / PRUNE_BIND_MOUNTS="yes" / etc.
+  if grep -qE '^[[:space:]]*PRUNE_BIND_MOUNTS[[:space:]]*=' "$UPDATEDB_CONF"; then
+    sudo sed -i -E 's|^[[:space:]]*PRUNE_BIND_MOUNTS[[:space:]]*=.*$|PRUNE_BIND_MOUNTS = "no"|' "$UPDATEDB_CONF"
+  else
+    echo 'PRUNE_BIND_MOUNTS = "no"' | sudo tee -a "$UPDATEDB_CONF" >/dev/null
+  fi
+else
+  echo "Warning: $UPDATEDB_CONF not found; skipping PRUNE_BIND_MOUNTS configuration" >&2
+fi
```
dhh, line 15: We don't need any of this defensive guarding since this is all running against a known, fixed target on install. So you can just execute the grep/sed pair directly.

dhh:
Good catch. We also need a migration for existing installations 👍

shreyansh-malviya:
> Good catch. We also need a migration for existing installations 👍

On a soft note, the migration associated with [commit](https://github.com/basecamp/omarchy/commit/9f561e0123b0a2266ca35442fefa5a9290d1e898) misses on extension.

1771622309 -> 1771622309.sh

PR #4295, opened 2026-01-17
Feat: Combine package installations from pacman and yay

severin-richner:
# Summary
Removes the `AUR` option from the `Install` menu and handles all package installations through the `Package` option.

## Why
My current flow for installing is as follows:
Open menu > Install > Package > search package (not found) > exit
Reopen menu > Install > AUR > seach package (found) > install

With this PR the `Package` section lists also packages from the AUR and adds a warning for the user:
image

Official packages are still installed using `pacman` and only packages from the AUR use `yay`.

Pros:
- easier package search
- less options in the menu
- same functionality as before

Cons:
- maybe more likely to install something unofficial, but warning should make this clear

### DIFF
--- a/bin/omarchy-pkg-install
+++ b/bin/omarchy-pkg-install
@@ -48,10 +48,8 @@ if [[ -n "$pkg_list" ]]; then
         IFS=' ' read -r pkg repo <<< "$pkg_repo"
         # use yay for AUR, pacman for official repos (including omarchy)
         if [[ $repo == "aur" ]]; then
-            echo "using yay"
             yay -S --noconfirm "$repo/$pkg"
         else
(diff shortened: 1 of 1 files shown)


sean-imus:
this is great, i have the exact same flow as you when i try to install something, and this is the perfect solution. this gets a +1 from me, adds simplicity at no apparent cost :+1:

dhh:
This is how it used to be. But AUR and the official Arch and Omarchy repositories are actually very different in terms of availability and reliability. AUR has many times had long outages. And packages on the AUR are not vetted and are more likely to break. So I prefer having things separated, so folks know exactly when they start having dependencies on AUR packages.

PR #4300, opened 2026-01-18
feat: Add confirmation dialog for closing protected windows

ToniFeliudeC:
## Summary
Adds optional confirmation dialog when closing windows for user-defined "protected" applications.

## IMPORTANT
The "Close all windows" command (CTRL+ALT+DELETE) is not covered by this change. That binding is intentionally uncomfortable to press, making accidental triggers unlikely. However could be added too if wanted

## What does this "solve"
Some applications don't have built-in "unsaved changes" prompts, or users may simply want extra protection against accidentally closing critical windows. In a keyboard-driven OS like Omarchy, it's easy to hit SUPER+W by mistake (specially if you are learning it) and close something you didn't intend to a game mid-session, a chat app, a long-running process, etc.

With this PR, users can list window classes in protected-windows, and those windows will show a confirmation prompt before closing:

image image

Unprotected windows close immediately as before (killactive).

## Functionality
New script omarchy-hyprland-window-close checks if the active window's class matches any entry in protected-windows
If not protected → closes immediately
If protected → opens a floating terminal with gum confirm
SUPER+W binding updated to use this script

## Good things
No behavior change unless user configures protected windows
One window class per line
Empty config = same behavior as before
## Not too good
Tiny overhead checking the config file on every SUPER+W

### DIFF
--- a/bin/omarchy-hyprland-window-close
+++ b/bin/omarchy-hyprland-window-close
@@ -1,0 +1,52 @@
+#!/bin/bash
+
+# Close the active window, with optional confirmation for protected apps.
+# Protected app classes are listed in ~/.config/omarchy/protected-windows (one per line).
+
+set -euo pipefail
+
+PROTECTED_FILE="$HOME/.config/omarchy/protected-windows"
+
+# Get active window info
+win_json="$(hyprctl activewindow -j 2>/dev/null || true)"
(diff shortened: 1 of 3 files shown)


dhh:
This is an interesting idea, but not something I've personally felt a need for in my usage. Apps like Typora already offers a built-in "do you want to save before closing" option, as does many others. Feels like something that should be fixed at the application level.

PR #4307, opened 2026-01-19
Add global search to omarchy-menu

DonSqualo:
SUPER + Space now opens a custom menu that allows for quick navigation to anything within the current menu tree.

## Summary
- Rewrite the menu system with a declarative architecture that enables global search across all menu items
- Search results are displayed using breadcrumb paths (e.g., "Install › Development › JavaScript › Node.js")

## Changes
- refactor omarchy-menu to build from omarchy-menu.def
- elephant plugin registration of main menu and all others combined, also from omarchy-menu.def
- Hierarchical navigation preserved in submenus
- Legacy support for direct menu access (`omarchy-menu screenrecord`, etc.) maintained

## Test plan
- [ ] Open menu with Super key and verify top-level items appear
- [ ] Type to search - verify all nested items appear with breadcrumb format
- [ ] Select a search result and verify action executes correctly
- [ ] Navigate into submenus and verify hierarchical behavior works
- [ ] Test direct menu access: `omarchy-menu install`, `omarchy-menu system`, etc.
- [ ] Test conditional items (Suspend/Hibernate appear based on system state)

### DIFF
--- a/bin/omarchy-launch-walker
+++ b/bin/omarchy-launch-walker
@@ -12,4 +12,4 @@ if ! pgrep -f "walker --gapplication-service" > /dev/null; then
   setsid uwsm-app -- walker --gapplication-service &
 fi
 
-exec walker --width 644 --maxheight 300 --minheight 300 "$@"
+exec walker --width 644 --maxheight 530 --minheight 300 "$@"
--- a/bin/omarchy-menu
+++ b/bin/omarchy-menu
@@ -44,9 +44,6 @@ install_terminal() { present_terminal "omarchy-install-terminal $1"; }
 aur_install() { present_terminal "echo 'Installing $1 from AUR...'; yay -S --noconfirm $2"; }
 aur_install_and_launch() { present_terminal "echo 'Installing $1 from AUR...'; yay -S --noconfirm $2 && setsid gtk-launch $3"; }
 
-#=============================================================================
-# Condition evaluation for def-based menus
-#=============================================================================
 eval_condition() {
   local cond="$1"
   [[ -z "$cond" ]] && return 0  # No condition = always show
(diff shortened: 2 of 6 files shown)


severin-richner (changes requested):
Currently not working, initial menu doesn't make sense. Searching and confirming intermediate menus like "Style" and "Setup" don't do anything.

Overall I feel like this change just makes the menu experience more confusing, wouldn't recommend to add this.

DonSqualo:
Manually fixed it, Claude shouldnt have already made a pull request.

Intended behaviour:
pressing SUPER + Space opens dmenu style menu like before, but search is global. This avoids having to type Setup, Enter, Wifi to go open the Wi Fi settings, but more importantly it avoids having to memorize the entire submenu structure.
This makes reaching apps, settings and learnings a lot more convenient.

I also refactored omarchy-menu to be configurable from a .def file, which allows reading and editing the entire menu in one place

severin-richner (commented):
Did a manual test:
1. overwrote my local omarchy setup with the proposed changes
2. ran the migration script, but got a "Unable to restart Walker -- RESTART MANUALLY" from the `omarchy-restart-walker` script (turns out I have no `app-walker@autostart.service`, which this script tries to restart)
3. Tried to open menu and just got the same as the `Omarchy launch apps` menu
4. rebooted
5. Still when I launch the `omarchy-menu` it just lists applications. Turns out when I search for something it does start to list menu items (very confusing behavior), but in a strange ordering, for example for searching `install` I see results like this:

image

Also pressing enter on the options doesn't seem to work. Tried with `Update > Omarchy`, `Install > Package`, `Remove > Package` --> nothing happens.

6. Tried to run command `omarchy-menu install`, correctly opens install menu and seems functional. Same with `omarchy-menu system`.

I haven't really looked at the code, but it seems strange to have a `.def` file in the `bin` directory.

dhh:
Appreciate the suggestion, but I don't like seeing everything just all mushed together like this. If we were ever to do something like this, it would need to separate things found at the current level and anything else that's deeper. It would also need to respect depth, so if you're already down a level, it should only search from there.

PR #4312, opened 2026-01-20
Add locale-aware date to waybar clock format

timothygwright:
## Summary

- Adds the locale-aware date (`%x`) to the default waybar clock format
- New format: `{:L%A %H:%M %x}` displays as "Tuesday 17:47 01/20/2026"
- Uses the `:L` prefix with `%x` to respect system locale (MM/DD/YYYY for US, DD/MM/YYYY for others)

### DIFF
--- a/config/waybar/config.jsonc
+++ b/config/waybar/config.jsonc
@@ -61,7 +61,7 @@
     "on-click-right": "alacritty"
   },
   "clock": {
-    "format": "{:L%A %H:%M}",
+    "format": "{:L%A %H:%M %x}",
     "format-alt": "{:L%d %B W%V %Y}",
     "tooltip": false,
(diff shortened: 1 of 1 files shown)


config/waybar/config.jsonc
```diff
@@ -61,7 +61,7 @@
     "on-click-right": "alacritty"
   },
   "clock": {
-    "format": "{:L%A %H:%M}",
+    "format": "{:L%A, %H:%M, %x}",
```
severin-richner: Date gets already displayed in  the `format-alt` (when you click on it).

timothygwright: I was looking for time and date at the same time, instead of one at a time

dhh:
Appreciate the suggestion, but I prefer the clean look of just having the time up there by default. But this config is meant to be user editable!

PR #4317, opened 2026-01-20
Add user envs.conf override for environment variables

simon-ami:
Users had no way to override environment variables set in the default `envs.conf` (like `QT_STYLE_OVERRIDE`) without manually editing `hyprland.conf` to add a source line.

This adds `envs.conf` to the user config pattern, matching how other configs (monitors, input, bindings, looknfeel, autostart) already work.

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/config/hypr/envs.conf
+++ b/config/hypr/envs.conf
@@ -1,0 +1,2 @@
+# Override default Omarchy environment variables
+# env = XCURSOR_SIZE,32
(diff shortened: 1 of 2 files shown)


pomartel:
You will find env variables in the `~/.config/uwsm/` directory.

dhh:
You can actually just overwrite any ENV inside your existing .config/hypr/hyprland.conf. ENVs are made to be overwritten 👌

PR #4330, opened 2026-01-21
Add kb_variant to input.conf

manuel1618:
(Very) small Quality of Life Update. I happened to look up this parameter more than once and think that it might be useful for others as well:

Personally, I use the "kb_variant = intl" to be able to type letters like ä, ö, ü also on the us layout. There are other use cases where you want to specify a variant of a layout. This is just a hint that its quickly available to users.


dhh:
I like it, but it needs a line of documentation and separation from the other block before it.

PR #4331, opened 2026-01-22
Open Nautilus in cwd of terminal

schwepmo:
TLDR: `SUPER+SHIFT+F` now opens Nautilus in the current working directory. 

Similar to how `SUPER+SHIFT+T` opens a terminal in the latest directory of a terminal, I find it super useful for Nautilus to be opened in the directory I'm currently working in.
Luckily, just reusing the `omarchy-cmd-terminal-cwd` script seems to to the trick.

One could also consider adding `SUPER+SHIFT+ALT+F` for this functionality, or for opening Nautilus at the default directory.

### DIFF
--- a/migrations/1769543550.sh
+++ b/migrations/1769543550.sh
@@ -1,0 +1,6 @@
+echo "Add SUPER+ALT+SHIFT+F shortcut to open nautilus in cwd"
+
+# Add the new CWD binding if it doesn't exist
+if ! grep -q "SUPER ALT SHIFT, F" ~/.config/hypr/bindings.conf; then
+  sed -i '/bindd = SUPER SHIFT, F, File manager, exec, uwsm-app -- nautilus --new-window/a bindd = SUPER ALT SHIFT, F, File manager (cwd), exec, uwsm-app -- nautilus --new-window "$(omarchy-cmd-terminal-cwd)"' ~/.config/hypr/bindings.conf
+fi


pomartel:
I was trying to do the opposite. Open a terminal from Nautilus in the current working directory.

That said, having a global shortcut for an app-specific trigger is probably not a good idea.

schwepmo:
> I was trying to do the opposite. Open a terminal from Nautilus in the current working directory.
> 
> That said, having a global shortcut for an app-specific trigger is probably not a good idea.

There is https://github.com/Stunkymonkey/nautilus-open-any-terminal, for which you should be able to configure a keyboard shortcut.

dhh:
I like it, but yes, let's put it on ALT. Then we also need a migration to add this for existing bindings.conf files.

pomartel:
Turns out that is you install the `nautilus-python` package, it adds an "Open in Ghostty" option to Nautilus. https://github.com/ghostty-org/ghostty/discussions/2416

@dhh Should I turn this into a PR ?

screenshot-2026-01-23_09-24-40

schwepmo:
> I like it, but yes, let's put it on ALT. Then we also need a migration to add this for existing bindings.conf files.

Done!

dhh:
We need a migration to add this to existing binding.conf files too.

dhh:
Requires a rebase.

shafayetejaman:
The shortcut does not work inside tmux.
I think adding tmux config is also a good option

bind-key F run-shell "nohup nautilus #{pane_current_path} >/dev/null 2>&1 & disown"

PR #4334, opened 2026-01-22
feat: added omarchy-pkg-update to manage package updates

jacopofilonzi:
Added a script based of the other omarchy-pkg-* style to see packages that needs to be updated.
Allow updating single or multiple packages.
Added `a` shortcut to update all packages automatically

### DIFF
--- a/bin/omarchy-pkg-update
+++ b/bin/omarchy-pkg-update
@@ -1,0 +1,39 @@
+#!/bin/bash
+
+# Show a fuzzy-finder TUI for updating Arch and OPR packages
+
+fzf_args=(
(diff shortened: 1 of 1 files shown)


abenz1267:
I don't think this is a good idea.

1. Arch highly discourages partial upgrades
2. Package updates might need updated config files, which Omarchy handles through migrations

Adding ANOTHER option for how to update packages doesn't really bring any benefits i think.

dhh:
I get the impulse, but agree with benz. We really need to stick to Arch's model of world updates to prevent any shared library conflicts.

PR #4340, opened 2026-01-23
Allow overriding default browser via $BROWSER

thejesus567:
## Summary

When I first started using Omarchy, I had some issues trying to change the default browser.

Currently, the only documented way to do this is by running:

`xdg-settings set default-web-browser`

(as discussed here: https://github.com/basecamp/omarchy/discussions/985)

This simple PR adds a more straightforward option: allowing users to set the `$BROWSER` environment variable in the default config file.

## What changed

- `omarchy-launch-browser` now respects `$BROWSER` if it is set
- If `$BROWSER` is not set, it falls back to `xdg-settings` (current behavior)
- Added an example in the default config file to document this option

## Why

Setting an environment variable is simpler, more explicit, and consistent with common Linux conventions. It also avoids having to run `xdg-settings` manually just to override the browser.

### DIFF
--- a/config/uwsm/default
+++ b/config/uwsm/default
@@ -6,6 +6,9 @@ export TERMINAL=xdg-terminal-exec
 # Use code for VSCode
 export EDITOR=nvim
 
+# Use a different default browser (make sure browser is installed)
+# export BROWSER=firefox.desktop
+
 # Use a custom directory for screenshots (remember to make the directory!)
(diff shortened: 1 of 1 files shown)


dhh:
Appreciate the idea and the option, but I actually think this is more confusing for most users, as browsers already natively support the xdg-settings approach. They usually offer you "Make this your default browser" when you open them, and if you accept, they trigger that xdg-settings flow. So I prefer to stick to just one option here.

PR #4345, opened 2026-01-23
Add nautilus-python package for 'Open in Ghostty' shortcut in Nautilus

pomartel:
If Ghostty is installed, this will add an Open in Ghostty shortcut for Nautilus on right click.

screenshot-2026-01-23_09-24-40

### DIFF
--- a/install/omarchy-base.packages
+++ b/install/omarchy-base.packages
@@ -78,6 +78,7 @@ mariadb-libs
 mise
 mpv
 nautilus
+nautilus-python
 gnome-disk-utility
 noto-fonts
 noto-fonts-cjk
--- a/migrations/1769183359.sh
+++ b/migrations/1769183359.sh
@@ -1,0 +1,3 @@
+echo "Add nautilus-python package for 'Open in Ghostty' shortcut in Nautilus"
+
+omarchy-pkg-add nautilus-python


pomartel:
Some context : https://github.com/ghostty-org/ghostty/pull/4816

neheb:
Out of curiosity, does drag and drop of files from nautilus to ghostty work? Upstream tells me it should work with nautilus-python but it's not working here.

pomartel:
> Out of curiosity, does drag and drop of files from nautilus to ghostty work? Upstream tells me it should work with nautilus-python but it's not working here.

No, I just tried it and it doesn't work. Let me know if you figure it out. It could be something useful to add the PR!

rez1-dev:
Thank you for all the QoL features @pomartel

pomartel:
> Thank you for all the QoL features @pomartel

Thanks for the kind words. I'm having such a blast contributing to Omarchy! 🚀

rez1-dev:
I don't see the option on Omarchy v3.5.1

atsyplenkov:
Has this feature been removed? I cannot find it on Omarchy v3.8.2 either

pomartel:
You need to install this package:
```
omarchy-pkg-add ghostty-nautilus
```

PR #4346, opened 2026-01-23
Remove proprietary garbage

benja2998:
While I have no problem with the user installing and using proprietary software, the problem is that Omarchy ships with proprietary software as a Linux distribution. This pull request removes proprietary software from Omarchy.

### DIFF
--- a/install/omarchy-base.packages
+++ b/install/omarchy-base.packages
@@ -84,7 +84,6 @@ noto-fonts-extra
 nss-mdns
 nvim
 obs-studio
-obsidian
 omarchy-chromium
 omarchy-nvim
 omarchy-walker


guilhermetk:
https://learn.omacom.io/2/the-omarchy-manual/61/commercial-guis

> Omarchy is mostly focused on providing free, open source software, but it's not religious about it. Sometimes the best solution is a commercial offering, **and that's just fine**.

abenz1267:
This won't ever be merged.

jubiloso:
But nice to have a list of proprietary software to remove. At least most of them are only web apps that doesn't use disk space, are only entries in a menu that open the URL in the default browser, right?

I understand 37signals adding their software for promotion. With all the work they have done, it's really fine, but adding a lot of software in the ISO image is unnecessary. You could display a welcome splash screen with checkboxes for selecting common software to install, specially the one that takes disk space.

benja2998:
I don't think a Linux distro should ship with proprietary software that isn't essential to make the hardware work.

PhaedrusFlow:
https://github.com/basecamp/omarchy/pull/4346#issuecomment-3794808797

Then fork omarchy and remove what you think it shouldn't have to offer an alternative to folks who want it- I don't see omarchy giving up obsidian based on an opinion.

If you were to offer a pr that provided a FOSS alternative that clearly worked better, that would help your case.

PR #4350, opened 2026-01-24
Replace alacritty with xdg-terminal-exec in the waybar

pomartel:
Right-clicking on the waybar cpu icon opens a Alacritty. I changed it to use the user's preferred terminal instead.

Since right clicking the waybar Omarchy logo already opens a terminal, we could also remove the right click action for the cpu icon.

### DIFF
--- a/config/waybar/config.jsonc
+++ b/config/waybar/config.jsonc
@@ -58,7 +58,7 @@
     "interval": 5,
     "format": "󰍛",
     "on-click": "omarchy-launch-or-focus-tui btop",
-    "on-click-right": "alacritty"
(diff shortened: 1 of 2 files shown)


neheb:
I could be wrong but I think the intention here is to provide a fallback for systems that don't support ghostty.

pomartel:
xdg-terminal-exec opens alacritty if it is set as the preferred terminal app.

dhh:
As others have mentioned, this is essentially a backup for when Ghostty won't run on a system. It's not meant to be a general purpose launcher for the terminal. We already have that on `Super + Return` 👍

PR #4361, opened 2026-01-26
This PR introduces a concept of Omarchy-lite

KonTy:
There are people that want to use omarchy because it is a nice light distro but it has a lot of bloat, this PR attempts to address this issue, togeather with a omarchy-iso PR this allows you to add --lite switch and create light version of omarchy that you can use as a base and customize it the way you want.

the exact command is:

 ./bin/omarchy-iso-make --lite 

you can also run

 ./bin/omarchy-iso-make --lite +dev +com

this will get packges from dev and com packges files and install them allowing us to create many different  SKUs

I fully tested this in full mode, that way DHH gets all of his apps and will not get upset and --lite mode.

There might be some bugs but majority of functinality is working. 

omarchy-base-lite.packages - REMOVED bloat:
- Apps: 1password, obsidian, libreoffice, spotify, signal-desktop, evince, gnome-calculator
- Media: obs-studio, kdenlive, pinta, gpu-screen-recorder
- Dev: docker, nvim, omarchy-nvim, tree-sitter-cli (moved to +dev)
- Dev libs: dotnet, postgresql-libs, mariadb-libs, luarocks, python-poetry-core
- Terminals: ghostty, alacritty (replaced with kitty)
- Other: localsend, typora, xournalpp, ufw-docker

omarchy-other-lite.packages - REMOVED dev packages

+dev addon includes: nvim, omarchy-nvim, vscodium, docker, lazygit, etc.

Web apps: None installed by default

KEPT: All OS utilities including printing (cups)

### DIFF
--- a/bin/omarchy-cmd-screensaver-lite
+++ b/bin/omarchy-cmd-screensaver-lite
@@ -1,0 +1,36 @@
+#!/bin/bash
+
+# Omarchy Lite screensaver - Matrix effect only
+
+screensaver_in_focus() {
+  hyprctl activewindow -j | jq -e '.class == "org.omarchy.screensaver"' >/dev/null 2>&1
+}
+
+exit_screensaver() {
+  hyprctl keyword cursor:invisible false &>/dev/null || true
+  pkill -x tte 2>/dev/null
+  pkill -f org.omarchy.screensaver 2>/dev/null
+  exit 0
+}
+
+# Exit the screensaver on signals and input from keyboard and mouse
+trap exit_screensaver SIGINT SIGTERM SIGHUP SIGQUIT
+
+printf '\033]11;rgb:00/00/00\007'  # Set background color to black
+
+hyprctl keyword cursor:invisible true &>/dev/null
+
+tty=$(tty 2>/dev/null)
+
+while true; do
+  tte -i ~/.config/omarchy/branding/screensaver.txt \
+    --frame-rate 120 --canvas-width 0 --canvas-height 0 --reuse-canvas --anchor-canvas c --anchor-text c\
+    --effect matrix \
+    --no-eol --no-restore-cursor &
+
+  while pgrep -t "${tty#/dev/}" -x tte >/dev/null; do
+    if read -n1 -t 1 || ! screensaver_in_focus; then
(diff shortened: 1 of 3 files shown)


Kasui92:
There used to be a **"bare"** version that more or less served its purpose. It has been removed, as Omarchy respects these [Doctrines](https://learn.omacom.io/3/omacom/81/doctrine), established since the project's inception.

all defaults can be easily removed post-installation via the menu, btw 🫡

KonTy:
@Kasui92 I agree and this lite mode totally 100% in line with these principles. Everything here in this mode follows that "Doctrine". The only things removed are the apps that are very specific, like Node.JS, and specific things for Ruby developers. Not everyone is a web developer. I'm a KDE developer, for example, and I don't want to have to install web developer tools. Everything else is kept in the ISO, Aesthetics, keyboard shortcuts, etc. 

Also, this is a relatively simple change, maybe about 100 lines of code with a lot of flexibility gained from it.

ghrubi:
There have been other PRs before this with the same intent. Please see the comments https://github.com/basecamp/omarchy/pull/4346

KonTy:
> There have been other PRs before this with the same intent. Please see the comments #4346

What I did is completely different, they are saying they want to remove apps from the main image, what my PR is giving you is a choice, you can like always have a FULL distro, which has *all the apps* and is opinionated. But you  can also build a light weight ISO, and you can chose to publish it along side with the full iso or not, but by making a light distro you are attracting more people and community will grow because of it. 

you know like many distros have different images with KDE/XFCE/Cinammon etc. Same thing here, you can put on the web site a full version or a lite version and give people a choice.

dhh:
Appreciate the suggestion, but I'm just not interested in maintaining a lite version of Omarchy. The mission statement for Omarchy is detailed here: https://learn.omacom.io/3/omacom/81/doctrine#defaults-over-decisions.

But for folks who want what you're proposing here, I'd suggest making something like a post-install tailoring script. So you install Omarchy, then run `curl -fsSL https://your-tailoring/script | bash`, and you end up with your definition of a lite setup. This is how Omarchy itself started (and how it still works via the manual install).

PR #4362, opened 2026-01-26
Add copy-filepath-to-clipboard mode to screenshot tool

olivoil:
Adds a new processing mode that copies the screenshot's file path to clipboard instead of the image data. Useful for terminal applications like Claude Code or Open Code that can't paste images directly but accept file paths.

The keybinding `Ctrl+PrtScr` runs `omarchy-cmd-screenshot smart clipboard-path`. It still opens satty for annotation, saves to the same location with the same naming convention as other screenshots, and copies the full path to clipboard on save.

### DIFF
--- a/bin/omarchy-cmd-screenshot
+++ b/bin/omarchy-cmd-screenshot
@@ -81,7 +81,7 @@ grim -g "$SELECTION" - |
     --actions-on-enter save-to-clipboard \
     --save-after-copy \
     --copy-command 'wl-copy'
-elif [[ $PROCESSING == "clipboard-path" ]]; then
+elif [[ $PROCESSING == "copy-filepath-to-clipboard" ]]; then
   FILENAME="$OUTPUT_DIR/screenshot-$(date +'%Y-%m-%d_%H-%M-%S').png"
   grim -g "$SELECTION" - |
     satty --filename - \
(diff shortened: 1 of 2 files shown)


abenz1267:
I'd actually be interested in how often people missclick screenshot keybinds. With this added, that would be 5 different keybinds for the `print` key.

Personally I don't think this is very user-friendly state and I'm not sure if going further by adding yet another keybind is doing any good.

ryanrhughes:
I think this adds too much complexity and confusion to this process. If we wanted to just expand the screenshot utility to allow for users to configure this themselves, I think that'd be OK but this feels like a pretty specific flavor of screenshots that isn't for everyone and to @abenz1267's point; expands an already large menu of keybinds for screenshots.

Would you be interested in pairing this back to just adding an appropriate flag to the bin to make this possible?

olivoil:
That makes sense @ryanrhughes @abenz1267. I've removed the keybind and kept just the `copy-filepath-to-clipboard` flag in the bin script. 

Users who want it can add their own keybind.

Note Satty-org/Satty#319 recently added `copy-filepath-to-clipboard` as a native action. Once `0.21.0` is released, we can drop the `--copy-command` workaround and use `--actions-on-enter "save-to-file,copy-filepath-to-clipboard"`. Happy to open a follow-up PR when that lands.

dhh:
What's the workflow where you want the path on the clipboard instead of the image itself?

olivoil:
@dhh it's to paste a screenshot in claude code running in ghostty. The default screenshot copy method where it copies the image content does not work, neither does dragging the image from the file explorer. Pasting the path to an image is the only thing I've been able to use for this.

dhh:
Claude Code in Ghostty does work with image pasting, but unfortunately you have to use Ctrl + V directly. You can't use Super + V. But it does work. Do we still need this given that?

olivoil:
I did not realize Ctrl + V works, that's much better than having to remember a different screenshot keybind. Closing the PR, thank you!

PR #4374, opened 2026-01-27
Add Super + CTRL + M to toggle temporary Hyprland visual profile

jassem-manita:
As discussed in [discussion #4369](https://github.com/basecamp/omarchy/discussions/4369):

"I sometimes need to switch from my normal setup to the most minimal visual Hyprland config **temporarily**. That means manually editing multiple conf files and reloading the config just to be reverted later. Redoing this every time is slow and inconvenient. A visual mode conf that can be toggled on/off temporarily would be a better alternative."

## Solution

The existing keybindings are great for individual toggles but this PR solves a completely different problem
it is meant for scenarios where a combination of multiple tweaks "a visual profile" (more than 4-5 settings you have general options , decoration options , blur, animations, input, touchpad, misc....) need to be applied and reverted periodically depending on the situation not just for general tweaks.
This allows users to toggle multiple settings at once as a preset/profile rather than pressing 5+ different keybindings each time.
With this PR, it's one key to apply all/revert everything back to default.

Press **Super + CTRL + M** to instantly toggle between:

| Mode | Description |
|------|-------------|
| **Default mode** | Your regular Hyprland setup |
| **Custom mode** | Whatever visual tweaks you put in `mode.conf` |

### Screenshots

**Default Mode:**

default_mode
 

**Custom Mode (Minimal Visuals):**

custom_mode

---

## How It Works

**Press Super + CTRL + M:**

1.  `mode/toggle-mode.sh` script checks `hyprland.conf`
2.  Toggles the line that loads `mode/mode.conf`
3.  Reloads Hyprland automatically
4.  Shows a notification displaying:
   - ON/OFF status
   - Mode file location
   - Brief description

---

## Files Added/Changed

- **`bindings.conf`** - Added the Super + CTRL + M keybinding
- **`mode/toggle-mode.sh`** - The toggle script
- **`mode/mode.conf`** - Customizable visual config file

### DIFF
--- a/config/hypr/bindings.conf
+++ b/config/hypr/bindings.conf
@@ -28,4 +28,4 @@ bindd = SUPER SHIFT ALT, X, X Post, exec, omarchy-launch-webapp "https://x.com/c
 # bindd = SUPER, SPACE, Omarchy menu, exec, omarchy-menu
 
 # # Toggle Visual Mode on/off
-bindd = SUPER, B, Visual mode, exec, ~/.config/hypr/mode/toggle-battery-mode.sh
+bindd = SUPER CTRL, M, Visual mode, exec, ~/.config/hypr/mode/toggle-battery-mode.sh


sean-imus:
The example you showed in the pictures can already be achieved using existing keybindings (see `SUPER + K`), so adding a new script feels unnecessary.

For example:

 `SUPER + SHIFT + BACKSPACE` removes gaps
 `SUPER + BACKSPACE` disables transparency

You can also hide the waybar with:

 `SUPER + SHIFT + SPACE`

This already provides a quick path to a more minimal setup without introducing additional logic. If someone needs finer-grained control, they can extend this behavior themselves, but the default Omarchy keybindings cover the common "minimal view" use cases well.

Additionally, `SUPER + B` does not align with Omarchy's keybinding conventions, I believe a `SUPER + CTRL + [Letter]` would be more appropriate here, and the choice of `B` is not immediately intuitive.

jassem-manita:
Appreciate the feedback I understand your points about existing keybindings but let me clarify **the use case**.

The existing keybindings are great for individual toggles but this PR solves a completely different problem
it is meant for scenarios where a combination of multiple tweaks "a visual profile" (more than 4-5 settings you have general options , decoration options , blur, animations, input, touchpad, misc....) need to be applied and reverted periodically depending on the situation not just for general tweaks.
This allows users to toggle multiple settings at once as a preset/profile rather than pressing 5+ different keybindings each time.
With this PR, it's one key to apply all/revert everything back to default.

Regarding keybinding convention
You're absolutely right I'm going to change it to SUPER + CTRL + M

wacel-nacer:
I’ve actually run into this exact same issue myself several times.

While the individual keybindings mentioned are useful for quick tweaks, they don't quite solve the context switching problem. When I’m moving from a casual browsing setup to a deep focus or minimal work session, having to remember and hit 4 or 5 different key combos feels disjointed. It's not just about the toggles, it's about the state.

I'd much rather have one master toggle for a 'Visual Profile' than have to fumble with multiple individual shortcuts. It just makes the transition into a minimal setup feel way more seamless.

dhh:
Very interesting idea, but can't say it's something that I've personally found a need for. Would recommend documenting how to do this in the Discussions forum. Thanks for contributing!

PR #4375, opened 2026-01-28
Automatically switch power profile when plugged in and unplugged

pomartel:


dhh:
This would only be something we'd want for laptops. Need to make sure it doesn't apply to desktop machines.

pomartel:
Good catch! I updated the script with a function to test if a battery is present before setting the rules. I don't have a desktop running Omarchy to test this.

I don't know if this should be an optional install since we are forcing the `balanced` and `power-saver` modes, and this can only be overriden by editing the rules file. Or maybe it should be explained in the manual?

moso:
Woo, someone picked it up!

I'd prefer to run the `performance` profile on my end, if AC is on, to get the most out of my device. But that's up to personal preference, I guess.

I'd probably also prefer to run `balanced` until a certain battery level, eg `<= 10%`, when not on AC, and then make it set to the `power-saver` profile.

I'm aware not all laptops have a `performance` profile (I'm eyeing you, Lenovo). But you can apply a check with `powerprofilesctl list` to list available profiles, and `| grep -q "^performance$"` on the result.

Also, setting targets via a config file would be sweet. However, I'm not aware of how difficult that would be to implement.

I have a simple-but-working bash script that I can share for inspiration, which both detects different variable names for AC and Battery. It requires manual setup of the `udev` rules and a `systemd`-like service to also check on boot. Hence why I'd personally prefer something like this baked in 😉

pomartel:
That's also my concern that the defaults might not please everybody. It's just a matter of changing the values in `/etc/udev/rules.d/99-power-profile.rules`. But it's not an easily discoverable path! We could add an item in the power profile menu to launch the file in the default text editor or make a basic TUI to change the config.

install/config/powerprofilesctl-rules.sh
```diff
@@ -0,0 +1,19 @@
+battery_present() {
+  for bat in /sys/class/power_supply/BAT*; do
+    [[ -r "$bat/present" ]] &&
+    [[ "$(cat "$bat/present")" == "1" ]] &&
+    [[ "$(cat "$bat/type")" == "Battery" ]] &&
+    return 0
+  done
+  return 1
+}
```
dhh: Think we should extract this into `omarchy-battery-present` since we use a similar check in a few other places.

install/config/powerprofilesctl-rules.sh
```diff
@@ -0,0 +1,19 @@
...
+
+if battery_present; then
+  sudo mkdir -p /etc/udev/rules.d
+
+  cat <<'EOF' | sudo tee "/etc/udev/rules.d/99-power-profile.rules"
+SUBSYSTEM=="power_supply", ATTR{type}=="Mains", ATTR{online}=="0", RUN+="/usr/bin/powerprofilesctl set power-saver"
+SUBSYSTEM=="power_supply", ATTR{type}=="Mains", ATTR{online}=="1", RUN+="/usr/bin/powerprofilesctl set balanced"
```
dhh: I'd rather see a default where we're running in balanced on battery, but performance (when available) on mains. Not all laptops have the performance setting, though?

moso: As mentioned above, we can check for available profiles with `powerprofilesctl list`, and adjust if `performance` isn't available. Some Lenovo models are notoriously known for lacking a `performance` profile.

moso:
@pomartel https://gist.github.com/moso/dc24c05bb2b154c88da7ceab89c220c5 for inspiration

Keep in mind that it's "homemade" and a classic "it works on my end" kind of thing :slightly_smiling_face:

pomartel:
Coming from a Macbook Air, I'm really annoyed by the fan noise on performance mode! 😆

I extracted the battery logic to a separate script and modified the logic as follow:

```
# Default AC profile:
# 3 profiles → performance
# 2 profiles → balanced

# Default Battery profile:
# 3 profiles → balanced
# 2 profiles → balanced
```
Let me know if you prefer that it defaults to power-saver on a 2 profiles system.

It's just missing an easy way to override the defaults. Unfortunately ENV variables won't work at the system level.

install/config/powerprofilesctl-rules.sh
```diff
@@ -0,0 +1,28 @@
+if omarchy-battery-present; then
+  mapfile -t profiles < <(omarchy-powerprofiles-list)
+
+  if [[ ${#profiles[@]} -eq 0 ]]; then
+    echo "Error: No power profiles available"
+    exit 1
```
dhh: Since this is running during setup, we don't want any error message output or exits. So think just wrap everything in this conditional, rather than treating it as a precondition.

pomartel: done!

migrations/1769566732.sh
```diff
@@ -0,0 +1,3 @@
+echo "Set Power Profiles Rules when plugged in (balanced) and unplugged (power-saver)"
```
morgoth, line 1: That actually might be very annoying.
I work with unplugged laptop (F13) every day on balanced profile, as power-saver makes the performance really poor.

As suggested in https://github.com/basecamp/omarchy/pull/4375#issuecomment-3814062733 that would also be my preferred way for automatic switch.
BTW, this is how it works on Ubuntu
> I'd probably also prefer to run balanced until a certain battery level, eg <= 10%, when not on AC, and then make it set to the power-saver profile.

pomartel: I just realized that the PR was merged a few minutes before I updated that comment. The defaults are actually performance on AC and balanced on Battery. Not sure if it's worth opening a new PR to change the echo.

morgoth: I see. It's not obvious by looking at the code ;-)
I would say, it is worth to change it before release

pomartel: Ok, just did. Thanks for catching it before the release! https://github.com/basecamp/omarchy/pull/4388

ituri:
How do I turn this off? I prefer to run my ThinkPad in balanced mode as the performance mode keeps the fans running all the time. 

There's no `/etc/systemd/system/power-profile-auto.service`on my system as suggested in the PR.

dhh:
Rule is in /etc/udev/rules.d/99-power-profile.rules

PR #4391, opened 2026-01-30
Add screensaver launch binding

AbdSattout:


dhh:
Appreciate the suggestion, but I think we're well-enough served with Super + Esc for the system menu to start. I also don't know of many keyboards that have a dedicated button for this! But you can always add it to your own Bindings ✌️

PR #4392, opened 2026-01-30
Add ability to install different shells from menu

Mrid22:
added ability to install different omarchy shells (fish, zsh, reset to bash) from main menu


dhh:
There's a lot more to the compatibility for the shell than just installing it. We've had a lot of tickets from folks running alternative shells, and then not getting all the initialization and features running correctly. I appreciate the suggestion, but for now, we're sticking to Bash on the default path. If someone wants to switch the shell, they'll know they're on a DYI adventure that isn't supported in the default setup.

PR #4395, opened 2026-01-30
Add automatic background cycling 

Eggmansmile:
I added automatic background cycling (disabled by default). I think it's one of those easy to add features but they're good for people with big background libraries that want something fresh every once in a while.

screenshot-2026-01-31_00-13-11

image


brink-lab:
I love this! Maybe it should be added as a sub-menu to "Backgrounds"? "Theme" and "Font" already have sub-menus as well.

Eggmansmile:
So in style -> backgrounds -> cycle background menu with a trigger to switch background?

Eggmansmile:
image

image

What about this

dhh:
Nice idea! I don't personally desire this cycling, so I'd rather not take on the weight to maintain this solution, but maybe you could add it as a omarchy-cycling-backgrounds package or something like that on AUR? Thanks for working on it ✌️

PR #4401, opened 2026-01-31
fix(waybar/config.jsonc): remove extra character in clock's format-alt

d1nozavr:
"format-alt": "{:L%d %B W%V %Y}" -> "format-alt": "{:L%d %B %V %Y}"


dhh:
That character is intentional. It's W for Week. Signifying that it's a week number 👍

PR #4406, opened 2026-02-01
Fix silent suspend failures when sleep.target is masked

RyanLatimer:
Some systems have sleep.target masked, which causes systemctl suspend to silently fail despite the hardware having full suspend support. The system goes through pre-suspend motions (lock screen, etc.) but never actually enters a sleep state.

This PR adds a migration to unmask the target in the case that the device has it masked and, as a result, has sleep functionality inhibited. In addition, when a user tries to enable sleep or hibernate, the same check is run.

As far as I can tell, there are no side effects to unmasking this target, as all it does is allow systemctl suspend to function properly and not fail silently.


RyanLatimer:
This change may cause driver bugs in NVIDIA GPUs that cause them to not wake.

PR #4414, opened 2026-02-01
Fix hypridle timout comment

nicolasbartsch:
**Changes:**

- What: fix misleading comment (151s ≠ 5min)
- Why: prevents confusion when users tweak idle/lock timings
- Scope: comment-only, no behavior change

### DIFF
--- a/config/hypr/hypridle.conf
+++ b/config/hypr/hypridle.conf
@@ -11,7 +11,7 @@ listener {
 }
 
 listener {
-    timeout = 151                      # 5min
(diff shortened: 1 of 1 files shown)


pomartel:
The screensaver resets the hypridle timer (this is likely a bug!). The workaround is to  start the screensaver after to 2.5 minutes and the turn off the screen after another 2.5 minutes, so 5 minutes total. Leave your computer idled for 5 minutes and see if you experience the same behaviour.

ReloadingBee:
Note that if the screensaver is disabled, the timer doesn't get reset.

PR #4467, opened 2026-02-02
Add dynamic theming for gum

severin-richner:
# Summary
This adds dynamic theming for gum by using the colors from the current theme.

## Why
The `gum` cli tool is used quite extensively in omarchy scripts and it's default pink theme can look quite out of place.

## Remark
I don't like how this requires to add a lot of different environment variables. Sadly it doesn't seem like `gum` provides a "global theme" setting, so this might still be the best option.

To avoid having all those variables in the environment, we could add a bash function which shadows `gum` and adds the variables just for this scope.

## Examples
first image is before, second is after

### Catppuccin: Install > Web App
catppuccin_install_webapp_before
catppuccin_install_webapp_after

### Catppuccin Latte: Remove > TUI
catppuccin_latte_remove_tui_before
catppuccin_latte_remove_tui_after

### Matte Black: Update > Omarchy
matte_black_update_before
matte_black_update_after

### Rose Pine: Install > Web App
rose_pine_install_webapp_before
rose_pine_install_webapp_after


dhh:
Can you show some images of where this improves things? The gum styling we have now is already derived from the terminal colors that are theme dependent.

severin-richner:
> Can you show some images of where this improves things? The gum styling we have now is already derived from the terminal colors that are theme dependent.

Added some example screenshots.

migrations/1769993645.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+# reapply the current theme and restart hyprland to ensure the environment variables are set.
+THEME_NAME_PATH="$HOME/.config/omarchy/current/theme.name"
+
+if [[ -f $THEME_NAME_PATH ]]; then
+  omarchy-theme-set "$(cat $THEME_NAME_PATH)"
```
dhh, line 8: Let's turn the theme part into omarchy-theme-reset. I've needed it for other things, like updating background images 👍

default/themed/hyprland.conf.tpl
```diff
@@ -7,3 +7,141 @@ general {
 group {
     col.border_active = $activeBorderColor
 }
+
+# Gum Style (generic) Variables
```
dhh: Let's split this out into a separate dedicated gum.envs.tpl file.

dhh:
That's a clear improvement. Just a few notes and we can merge this 👌

migrations/1769993645.sh
severin-richner: Added an omarchy-theme-reset script

default/bash/rc
```diff
@@ -3,4 +3,5 @@ source ~/.local/share/omarchy/default/bash/aliases
 source ~/.local/share/omarchy/default/bash/functions
 source ~/.local/share/omarchy/default/bash/init
 source ~/.local/share/omarchy/default/bash/envs
+[[ -f $HOME/.config/omarchy/current/theme/gum.envs ]] && source $HOME/.config/omarchy/current/theme/gum.envs
```
dhh: I like the simplicity of this, but then it would only apply to new terminals opened. Not existing ones, right?

severin-richner: Yes, I just realized that then the theming also only applies to user bash shells, so scripts like omarchy-menu which don't source the `~/.bashrc` file don't set the environment variables. (If they are started using a keyboard shortcut.)

I think setting the variables via hyprland might be the easiest approach?

dhh: Yes, fine to set via hyprland. Then we actually get reloading for free. You shouldn't even have to call hyprctl reload.

severin-richner: Great, in this case I just reverted the last commit and keep it in the `default/themed/hyprland.conf.tpl` (since adding something like a `default/themed/gum.env.conf.tpl`,  which then also gets added as a config to hyprland will probably just we more confusing than helpful).

dhh: I prefer having this in a separate file. Hyprland doesn't have a problem with includes. So we should use use gum.env.conf and then link those from the default hyprland config inside omarchy.

migrations/1769993645.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+# reapply the current theme (also restarts hyprland to add the gum env variables)
```
dhh: We are no longer restarting.

default/bash/rc
severin-richner: I now added a separate `gum.env.conf.tpl` file for this.

migrations/1769993645.sh
severin-richner: ~~Updated the migration script to also add the new `gum.env.conf` to the user's `hyprland.conf`.~~

Migration script just runs `omarchy-theme-restart`. The `gum.env.conf` file gets sourced in the `default/hypr/envs.conf` file.

default/themed/gum.env.conf.tpl
```diff
@@ -0,0 +1,137 @@
...
+env = GUM_CHOOSE_SELECTED_FOREGROUND,#{{ selection_foreground }}
+env = GUM_CHOOSE_SELECTED_BACKGROUND,#{{ selection_background }}
+
+# Gum Filter Style Variables
+env = GUM_FILTER_PROMPT_FOREGROUND,#{{ accent }}
+env = GUM_FILTER_PROMPT_BACKGROUND,#{{ background }}
+env = GUM_FILTER_TEXT_FOREGROUND,#{{ foreground }}
+env = GUM_FILTER_TEXT_BACKGROUND,#{{ background }}
+env = GUM_FILTER_MATCH_FOREGROUND,#{{ accent }}
+env = GUM_FILTER_CURSOR_TEXT_FOREGROUND,#{{ cursor }}
```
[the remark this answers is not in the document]
severin-richner: This template gets populated with the values from the `colors.toml` files. These already contain a # at the start. Meaning a this filled in template file has lines like:
`env = GUM_FILTER_PROMPT_FOREGROUND,##81a1c1`

Double ## is the way to escape a comment in hyprland, thus this is expected.

default/themed/gum.env.conf.tpl
```diff
@@ -0,0 +1,137 @@
+# Gum Style (generic) Variables
+env = FOREGROUND,#{{ foreground }}
+env = BACKGROUND,#{{ background }}
+env = BORDER_FOREGROUND,#{{ accent }}
+env = BORDER_BACKGROUND,#{{ background }}
+
```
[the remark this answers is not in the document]
severin-richner: These values do get used by gum. See `gum style --help`.

migrations/1769993645.sh
```diff
@@ -0,0 +1,3 @@
+echo "Resetting theme"
+
+omarchy-theme-reset
```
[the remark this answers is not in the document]
severin-richner: Done.

severin-richner:
Rebased this PR onto `dev` and updated the migration sctipt. Should be ready to merge, let me know if there are any concerns with it. @dhh

PR #4469, opened 2026-02-02
Add minimal TUI timer

severin-richner:
# Summary
Adds a small timer TUI utility, which integrates with waybar and has the ability to do pomodoro-style timers.

## Why
I find it useful to have a little timer, which I can directly see in the waybar while working. This addition is meant to do exactly that and nothing more.

## Demo
https://github.com/user-attachments/assets/32c6ca64-f1b7-4dee-b690-78f9e78d06ba

## Implementation Details and Limitations
The timer functionality essentially consists of 2 scripts. 

The `omarchy-timer` script provides the user interface and the CLI for setting a timer and launches a secondary script, which plays the sound and sends the notification when the timer runs out. It also implements the pomodoro logic, in which the timer simply oscillates between work and break time.

The `timer.sh` script handles the waybar indicator for displaying the active timer.

These two scripts use the `/run/user/$UID/omarchy-active-timer` file to communicate between them. The `omarchy-timer` script writes to this file the current active timers type (`t`:  timer, `w`: work or `b`: break) and the time at which this timer expires in UNIX time. A timer can be canceled by deleting this file, which can be done by right-clicking on the waybar indicator.

### Notifications
- if system is active and timer expires: plays sound and shows notification
- if system is locked and timer expires: plays sound (and triggers notification, however this is not visible on lock screen)
- if system is suspended and timer expires: nothing happens, whenever the system wakes up, the sound and notification are triggered
- if timer is active and user logs out or system shuts down: `/run/user/$UID/` directory gets cleared, thus timer is deleted

[Related Discussion: Incorporate a tui Pomodoro timer #4068](https://github.com/basecamp/omarchy/discussions/4068#discussion-9304565)

### DIFF
--- a/bin/omarchy-timer
+++ b/bin/omarchy-timer
@@ -1,0 +1,217 @@
+#!/bin/bash
+
+ACTIVE_TIMER_FILE=$XDG_RUNTIME_DIR/omarchy-active-timer
+
+if omarchy-cmd-missing gum hyprctl paplay waybar; then
+  echo "Requires gum, hyprctl, paplay and waybar commands to be present."
+  exit 1
+fi
+
+[ $# -eq 0 ] && INTERACTIVE_MODE=true || INTERACTIVE_MODE=false
+
+initial_prompt() {
+  clear
+  gum style \
+  	--border thick \
+  	--padding "1" \
+  	"OMARCHY TUI TIMER"
+}
+
+# returns the valid duration in seconds or "" if invalid
+validate_duration() {
+  local input="$1"
+  local token value unit
+  local seconds_total=0
+
+  if [[ -z "$input" ]]; then
+    echo ""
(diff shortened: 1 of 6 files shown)


dhh:
Neat idea! I think this would be nice as a omarchy-timer AUR package. It's a bit too much to take on for Omarchy itself at 333 lines, and I'm not sure it's something I'd personally use. But would be a great extra option for those who want it ✌️

severin-richner:
> Neat idea! I think this would be nice as a omarchy-timer AUR package. It's a bit too much to take on for Omarchy itself at 333 lines, and I'm not sure it's something I'd personally use. But would be a great extra option for those who want it ✌️

That's fair and good idea, I might look into creating an AUR package 👍

PR #4471, opened 2026-02-02
fix: theme template rendering with yq v4.50+

### LINKED ISSUE #4470: Theme templates not rendering due to yq string interpolation bug
## Problem

Theme templates (ghostty.conf, hyprland.conf, etc.) are not being rendered correctly. Template variables like `{{ background }}`, `{{ accent_strip }}` are passed through literally instead of being replaced with color values.

**Error when running `omarchy-theme-set`:**
```
Error: 1:86: lexer: invalid input text "^#\";\"\"))|g\""
```

**Affected files in `~/.config/omarchy/current/theme/`:**
- ghostty.conf: `background = {{ background }}` (should be `background = #FFFCF0`)
- hyprland.conf: `rgb({{ accent_strip }})` (should be `rgb(205EA6)`)
- And other templated configs

## Root Cause

In `bin/omarchy-theme-set-templates`, the yq expression that generates both standard and `_strip` substitutions in a single call doesn't work correctly with recent versions of yq (v4.50.1):

```bash
# Current (broken)
sed 's/=/:/' "$COLORS_FILE" | yq -r 'to_entries[] | "s|{{ \(.key) }}|\(.value)|g", "s|{{ \(.key)_strip }}|\(.value | sub("^#";""))|g"' > "$sed_script"
```

The nested string interpolation with `sub()` causes:
1. A lexer error for the `sub("^#";"")` syntax
2. The `$k` variable not interpolating in the second output string (produces `{{ _strip }}` instead of `{{ accent_strip }}`)

## Fix

Split into two separate yq calls:

```bash
# Generate standard substitutions
sed 's/=/:/' "$COLORS_FILE" | yq -r 'to_entries[] | "s|{{ \(.key) }}|\(.value)|g"' > "$sed_script"

# Generate _strip substitutions (separate call)
sed 's/=/:/' "$COLORS_FILE" | yq -r 'to_entries[] | "s|{{ \(.key)_strip }}|\(.value | sub(\"#\"; \"\"))|g"' >> "$sed_script"
```

Note: The `sub()` regex also changed from `"^#"` to `\"#\"` (escaped quotes work, raw quotes with semicolon separator cause lexer errors).

## Environment

- yq version: v4.50.1 (https://github.com/mikefarah/yq/)
- Omarchy version: latest from git
- Theme tested: flexoki-light

## Workaround

Until fixed, users can manually apply the fix by running:

```bash
# Re-run theme-set after patching
omarchy-theme-set "Your Theme Name"
```

Or manually render templates using the corrected yq syntax.

justcarlson:
## Summary

Fixes theme templates not rendering due to yq string interpolation bug. Template variables like `{{ background }}` and `{{ accent_strip }}` were being passed through literally instead of being replaced with color values.

## Changes

Split the yq expression in `omarchy-theme-set-templates` into two separate calls:

```bash
# Before (broken with yq v4.50+)
sed 's/=/:/' "$COLORS_FILE" | yq -r 'to_entries[] | "s|{{ \(.key) }}|\(.value)|g", "s|{{ \(.key)_strip }}|\(.value | sub("^#";""))|g"' > "$sed_script"

# After (works)
sed 's/=/:/' "$COLORS_FILE" | yq -r 'to_entries[] | "s|{{ \(.key) }}|\(.value)|g"' > "$sed_script"
sed 's/=/:/' "$COLORS_FILE" | yq -r 'to_entries[] | "s|{{ \(.key)_strip }}|\(.value | sub(\"#\"; \"\"))|g"' >> "$sed_script"
```

## Root Cause

The nested string interpolation with `sub()` in a single yq expression causes:
1. A lexer error: `invalid input text "^#\";\"\"))|g\""`
2. Variable not interpolating in second output (produces `{{ _strip }}` instead of `{{ accent_strip }}`)

## Test Plan

1. Run `omarchy-theme-set "Any Theme"`
2. Verify `~/.config/omarchy/current/theme/ghostty.conf` has actual color values (e.g., `background = #FFFCF0`) not template variables
3. Verify `~/.config/omarchy/current/theme/hyprland.conf` has rendered `_strip` values (e.g., `rgb(205EA6)`)

Fixes #4470

### DIFF
--- a/bin/omarchy-theme-set-templates
+++ b/bin/omarchy-theme-set-templates
@@ -15,19 +15,22 @@ hex_to_rgb() {
 if [[ -f $COLORS_FILE ]]; then
   sed_script=$(mktemp)
 
-  while IFS='=' read -r key value; do
-    key="${key//[\"\' ]/}"                # strip quotes and spaces from key
-    [[ $key && $key != \#* ]] || continue # skip empty lines and comments
-    value="${value#*[\"\']}"
-    value="${value%%[\"\']*}" # extract value between quotes (ignores inline comments)
+  # Generate standard substitutions
+  # Note: Split into two yq calls to avoid nested interpolation bug in yq v4.50+
+  sed 's/=/:/' "$COLORS_FILE" | yq -r 'to_entries[] | "s|{{ \(.key) }}|\(.value)|g"' > "$sed_script"
+  
+  # Generate _strip substitutions (removes # prefix from hex colors)
+  sed 's/=/:/' "$COLORS_FILE" | yq -r 'to_entries[] | "s|{{ \(.key)_strip }}|\(.value | sub(\"#\"; \"\"))|g"' >> "$sed_script"
 
-    printf 's|{{ %s }}|%s|g\n' "$key" "$value"            # {{ key }} -> value
-    printf 's|{{ %s_strip }}|%s|g\n' "$key" "${value#\#}" # {{ key_strip }} -> value without leading #
+  # Generate _rgb substitutions for hex colors
+  while IFS='=' read -r key value; do
+    key=$(echo "$key" | xargs)
+    value=$(echo "$value" | xargs | tr -d '"')
     if [[ $value =~ ^# ]]; then
       rgb=$(hex_to_rgb "$value")
-      echo "s|{{ ${key}_rgb }}|${rgb}|g"
+      echo "s|{{ ${key}_rgb }}|${rgb}|g" >> "$sed_script"
     fi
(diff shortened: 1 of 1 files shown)


dhh:
The theme application has been changed such that yq is no longer in the mix.

dhh:
You can see the bash-only version here: https://github.com/basecamp/omarchy/blob/dev/bin/omarchy-theme-set-templates

justcarlson:
This is so much cleaner! I have adopted it. Thanks! :smile:

PR #4480, opened 2026-02-02
Improve update check to verify latest tag is newer

pomartel:
Running on the dev branch, which is currently set at v3.3.3, I see the update icon in the waybar.

The `omarchy-update-script` condition should also check if the latest tag is greater than the current tag. Currently, it just checks if it's equal.


pomartel:
This might be happening because I set the `OMARCHY_PATH` to my fork. I will have to look into this a little bit more.

PR #4493, opened 2026-02-04
Disable animations during screenshots

hattapauzi:
## Summary
- temporarily disable Hyprland animations while `omarchy-cmd-screenshot` runs to avoid mid-transition captures
- restore the previous animation state on exit


dhh:
How is the problem you're trying to solve manifesting itself? Why not just wait until whatever anim you're looking at is done?

hattapauzi:
The issue shows up when triggering the screenshot during an active animation (workspace switch or window open/close). The captured frame can be in mid‑transition, so the window appears partially offset/blurred or mid‑fade. Waiting isn’t always works out because the screenshot flow itself (slurp + wayfreeze) can be invoked while animations are still running. The change freezes a stable frame by temporarily disabling animations only for the capture, then restores them immediately after.

I could not reproduce this behavior (screenshot result that's messed up by the animation) in my other machine with the same custom animations config but the animation is still indeed rolling when invoking the `omarchy-cmd-screenshot` ... After applying this pull request that I did, I can confirm that the animation is not rolling therefore not messing up my screenshot anymore...


https://github.com/user-attachments/assets/88897d1b-0249-4360-9da0-f53be2e1661b
Before using custom animations (animations provided by themes)


https://github.com/user-attachments/assets/683219d9-a354-476e-8c58-643a7b0c7dac
After using my own custom animations


https://github.com/user-attachments/assets/8529abe5-cde0-43d7-bce9-78ffb8d7a3c6
Just some example script I wrote to toggle between different scripts (some of them messed up the screenshot)


I can provide the messed up screenshot later when I got back to that machine...

dhh:
I appreciate the work into this, but I haven't seen this happen once in all my screenshots, and I'm sensitive to carrying a bunch of extra code we have to maintain forever for it. So going to skip at this time, but thanks for putting in the time on it!

hattapauzi:
Sure! I havent encountered the bug after the recent update on dev channel... I suspected it might got something to do with wayfreeze behavior from what Ive gathered from the issues discussion...

dhh:
We switched to using hyprpicker!

PR #4494, opened 2026-02-04
Fix/screenshot disable animations

hattapauzi:
## Summary
- temporarily disable Hyprland animations while `omarchy-cmd-screenshot` runs to avoid mid-transition captures
- restore the previous animation state on exit


hattapauzi:
Closing duplicate PR so we can retarget #4493 to dev.

PR #4495, opened 2026-02-04
Fix clipboard loop caused by wl-x11-sync two-way sync

maralcbr:
## Problem

Some setups have `wl-x11-sync.sh` configured with bidirectional clipboard sync (clipboard → primary → clipboard), causing an infinite feedback loop that breaks paste functionality.

## Solution

This migration:
- Detects the problematic two-way sync pattern (`wl-paste --type text --watch wl-copy`)
- Creates a backup of the existing script before modifying
- Replaces with one-way PRIMARY → CLIPBOARD sync
- Restarts the systemd service to apply changes immediately

The new script uses:
```bash
wl-paste --primary --type text --watch sh -c 'cat | tee >(wl-copy) | xclip -selection clipboard -i -f >/dev/null'
```

This syncs PRIMARY selection to both Wayland clipboard (`wl-copy`) and X11 clipboard (`xclip`) in a one-way flow, preventing the feedback loop while maintaining full clipboard compatibility.

### DIFF
--- a/migrations/1770189394.sh
+++ b/migrations/1770189394.sh
@@ -1,0 +1,33 @@
+echo "Fix clipboard loop caused by wl-x11-sync"
+
+SCRIPT="$HOME/.local/bin/wl-x11-sync.sh"
+UNIT="$HOME/.config/systemd/user/wl-x11-sync.service"
+
+# Some setups end up with a two-way clipboard<->primary sync using wl-paste/wl-copy.
+# That can create a feedback loop (clipboard change -> set primary -> primary change
+# -> set clipboard ...) and effectively break pasting.
+
+if [[ -f "$SCRIPT" ]]; then
+  if grep -qE 'wl-paste --type text --watch wl-copy( --primary)?' "$SCRIPT"; then
+    BACKUP="$SCRIPT.bak.$(date +%s)"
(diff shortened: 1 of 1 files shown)


dhh:
Given that this loop is not part of the default setup, I don't think it makes sense for us to carry accommodations in the box for it. A migration would also only fix existing systems, not new ones. Appreciate the suggestion either way ✌️

maralcbr:
Before applying this fix there was no way I could copy and paste between ghostty and chromium.

Does it work for you?

I posted on reddit and apparently others have the same issue.

dhh:
Yes, no problem pasting between ghostty and chromium. Have you installed anything custom around the clipboard or fcitx or something like that?

maralcbr:
Not at all. It’s a 2015 MacBook Pro.

Since my first Omarchy experience it has always been a problem that put me
off. And the only way I could solve it after trying in and off a lot of
things was the code in this PR.

I even reinstalled Omarchy completely at some point abd ended up with the
same issue out of the fresh installation.

dhh:
That's very strange. Definitely not a widespread problem. This is the first I hear of it. But it does sound very annoying! Maybe would be good to move this set of fixes over to a discussion here on GH, then we can ask folks experiencing the problem to chime in with what might be common factors.

Issue #1026, opened 2025-08-24
[Touchpad bug]-  Synaptics touchpad not working


Sameer292:
It was not working.. It was being detected but didn't work..
This fixed the issue:
`sudo modprobe psmouse synaptics_intertouch=1`

geek-at:
Thanks! Also works with Huawei trackpads

dhh:
Would be nice with a PR that'll do this automatically for the confirmed machines that needs it at install time 👍

Sameer292:
@dhh 
please check the PR
https://github.com/basecamp/omarchy/pull/4499

dhh:
Closed — fixed by PR #4499 (merged 2026-02-07), which enables Synaptics InterTouch for confirmed touchpads.

Whats-the-time:
Machine: Lenovo P3

>  grep synaptics /proc/bus/input/devices
>  P: Phys=synaptics-pt/serio0/input0
>  lsmod | grep "psmouse"
>  psmouse 245760 0

Problem is I need psmouse synaptics_intertouch=0 to fix my issue which is, touchpad does not work after resuming from hibernate/suspend.
I add psmouse.synaptics_intertouch=0 to /etc/default/limine under KERNEL_CMDLINE (next to quiet splash) and then run limine-update to update the /boot entry.

So, this is not a the solution for every 'synaptic' + 'psmouse' probe?

PR #4499, opened 2026-02-04
Enable Synaptics InterTouch for confirmed touchpads if not already loaded

Sameer292:
please view #1026 for the issue
@dhh


install/config/hardware/synaptic-touchpad.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+# Enable Synaptics InterTouch for confirmed touchpads if not already loaded
+
+if ! lsmod | grep -q '^psmouse'; then
+    echo "Loading psmouse module for synaptics touchpad..."
+    modprobe psmouse synaptics_intertouch=1
+else
+    echo "psmouse module loaded."
```
dhh: We don't need any echo statements during the install. So can just boil this down to the modprobe. Also, no shebang line needed. And 2 spaces, no tabs.

Sameer292: Updated

install/config/hardware/synaptic-touchpad.sh
```diff
@@ -0,0 +1,5 @@
+# Enable Synaptics InterTouch for confirmed touchpads if not already loaded
+
+if ! lsmod | grep -q '^psmouse'; then
```
dhh: Hmm, this triggers on my machine which does not have synaptics intertouch. Can't be that indiscriminate. It has to only be enabled when we know it's on the machine.

Sameer292: Oh yeah nice catch. Updated to check for synaptics intertouch and enable it only if available

Sameer292:
@dhh Are you planning the release? Let's include this with it.

PR #4524, opened 2026-02-06
 Add Framework 16 RGB keyboard theme syncing

godlewski:
## Summary

Syncs the omarchy theme accent color to the Framework Laptop 16 RGB keyboard via `qmk_hid`, following the same pattern as the existing ASUS ROG integration with `asusctl`.

- Detects Framework 16 hardware via DMI data
- Auto-installs `qmk-hid` from AUR on Framework 16 machines
- Installs a udev rule so `qmk_hid` can access the keyboard without root
- Converts the theme's hex accent color to QMK's HSV format using Python's `colorsys`
- Sets solid color effect with the accent hue/saturation and persists via `--save`

## New files

- `bin/omarchy-hw-framework16` — hardware detection via DMI
- `bin/omarchy-theme-set-framework16` — applies accent color to keyboard RGB
- `default/themed/framework16-kbd.rgb.tpl` — accent color template
- `install/packaging/framework16.sh` — conditional `qmk-hid` AUR package install
- `install/config/hardware/framework16-qmk-hid.sh` — udev rule for unprivileged HID access

## Modified files

- `bin/omarchy-theme-set` — calls `omarchy-theme-set-framework16` during theme application
- `install/packaging/all.sh` — registers the Framework 16 packaging script
- `install/config/all.sh` — registers the udev rule setup during install

## Tested on hardware

All testing done on a Framework Laptop 16 (AMD Ryzen 7040 Series) with ANSI keyboard module (FW 0.3.1).

- **Hardware detection** — `omarchy-hw-framework16` correctly identifies the machine via DMI vendor and product name
- **Package install** — `qmk-hid` installs from AUR via `omarchy-pkg-aur-add` and `yay`
- **Keyboard communication** — `qmk_hid -l` finds the keyboard (`32ac:0012`, "Laptop 16 Keyboard Module - ANSI")
- **Hex-to-HSV conversion** — verified against known colors (red, green, blue, white, orange), all values match expected QMK HSV output
- **Theme switching** — visually confirmed keyboard color changes when switching themes (ristretto coral → hackerman green → pure blue)
- **Graceful no-op** — exits cleanly with no output when `qmk_hid` is not installed or the theme file is missing
- **Sleep/wake persistence** — color survives sleep/wake cycles via `--save` (persists to keyboard EEPROM)
- **Install flow simulation** — ran both scripts through `run_logged` with the same env vars and execution path as `install.sh` to confirm they resolve paths correctly

## DEMO

https://github.com/user-attachments/assets/9f020740-05fb-46fe-b321-d25b3cff027a

### DIFF
--- a/migrations/1770483021.sh
+++ b/migrations/1770483021.sh
@@ -1,11 +1,4 @@
 echo "Install Framework 16 keyboard RGB support"
 
-if omarchy-hw-framework16; then
-  omarchy-pkg-add qmk-hid
-
-  if [[ ! -f /etc/udev/rules.d/50-framework16-qmk-hid.rules ]]; then
-    sudo cp "$OMARCHY_PATH/default/udev/framework16-qmk-hid.rules" /etc/udev/rules.d/50-framework16-qmk-hid.rules
-    sudo udevadm control --reload-rules
-    sudo udevadm trigger
-  fi
-fi
+source $OMARCHY_PATH/install/packaging/framework16.sh
+source $OMARCHY_PATH/install/config/hardware/framework16-qmk-hid.sh


dhh:
Nice! Let's make omarchy-theme-set-keyboard and then have that call omarchy-theme-set-keyboard-asus-rog and omarchy-theme-set-keyboard-f16. We also need a migration to install this on existing machines.

godlewski:
@dhh I updated the PR with  `omarchy-theme-set-keyboard` as a dispatcher that calls `omarchy-theme-set-keyboard-asus-rog` and `omarchy-theme-set-keyboard-f16` as you suggested! 

Also added the migration for existing Framework 16 machines.

I retested on my Framework 16:
- Keyboard RGB changes color when switching themes
- Migration installs qmk-hid, creates the udev rule, and is idempotent on re-run

install/packaging/framework16.sh
```diff
@@ -0,0 +1,3 @@
+if omarchy-hw-framework16; then
+  omarchy-pkg-aur-add qmk-hid
```
dhh: We can't actually install yay packages during initial setup, as we're running an offline installer. So we need to get qmk-hid built for the Omarchy Package Repository. You can open a PR here: https://github.com/omacom-io/omarchy-pkgs/

install/config/hardware/framework16-qmk-hid.sh
```diff
@@ -0,0 +1,12 @@
+# Allow unprivileged access to the Framework 16 keyboard for RGB control via qmk_hid.
+
+if omarchy-hw-framework16; then
+  if [[ ! -f /etc/udev/rules.d/50-framework16-qmk-hid.rules ]]; then
+    cat <<EOF | sudo tee /etc/udev/rules.d/50-framework16-qmk-hid.rules
+SUBSYSTEM=="hidraw", ATTRS{idVendor}=="32ac", ATTRS{idProduct}=="0012", MODE="0660", TAG+="uaccess"
```
dhh: We should put this file into default/udev/framework16-qmk-hid.rules, then copy it over when installing. Because then we can reuse for migration and better track all the files.

install/packaging/framework16.sh
godlewski: Sounds good! 

I put the PR up here: https://github.com/omacom-io/omarchy-pkgs/pull/45

default/themed/framework16-kbd.rgb.tpl
```diff
@@ -0,0 +1 @@
+{{ accent }}
```
dhh: Let's actually just share a single keyboard.rgb file. Don't need per-keyboard templates then.

migrations/1770483021.sh
```diff
@@ -0,0 +1,14 @@
+echo "Install Framework 16 keyboard RGB support"
+
+if omarchy-hw-framework16; then
+  omarchy-pkg-aur-add qmk-hid
```
dhh: Can be changed to `omarchy-pkg-add` now that the package has been merged.

godlewski:
@dhh I updated the PR with your feedback and retested on my Framework 16!

migrations/1770483021.sh
```diff
@@ -0,0 +1,11 @@
+echo "Install Framework 16 keyboard RGB support"
+
+if omarchy-hw-framework16; then
+  omarchy-pkg-add qmk-hid
+
+  if [[ ! -f /etc/udev/rules.d/50-framework16-qmk-hid.rules ]]; then
```
dhh: You can call the install file from here so we don't need to duplicate this logic.

dhh:
Wonderful work on this, @godlewski 🤘

PR #4533, opened 2026-02-07
Add the Vantablack theme

bjarneo:
screenshot-2026-01-31_18-41-22
preview


dhh:
Seems like there are still some regular colors left:

image

image

Seems like this should be entirely monochrome, right?

bjarneo:
Something is missing it seems. At my end it covers everything. I'll create another PR fixing this!

PR #4541, opened 2026-02-07
Implement a new color system with inspiration from base24

bjarneo:
Adds a semantic color naming scheme for `colors.toml` (inspired by [base24](https://github.com/Base24/base24), but with readable names instead of `base0X`) so themes can expose more than the 8 ANSI accents without forcing every consumer through `color0..15`.

Targets `omarchy-shell` so it can build on the new `mix_color` template helper.

## What this changes

- `bin/omarchy-theme-set-templates`
  - Adds alias fallbacks (`bg`/`fg`/`light_fg`/`bright_fg`/`lighter_bg`/`dark_fg`/`muted`/`selection`/`orange`) so themes can ship either canonical or aliased keys.
  - Restores the ANSI `color0..15` -> semantic alias map for backward compat (`red`<-`color1`, ..., `bright_purple`<-`color13`).
  - Derives `brown` from `orange` via `mix_color` (no more `darken_color` helper).
  - Resolves a `mode = "dark"|"light"` field per theme (see below).
- `default/themed/{pi.json,shell.toml,helix.toml,gum_env.lua,foot.ini}.tpl` migrate off `{{ colorN }}` placeholders and pull from the semantic keys directly. Internal palette names inside those templates (helix's palette block, Pi's vars) keep their existing labels; only the values change.
- Themes that still used ANSI `color0..15` (`last-horizon`, `solitude`, `white`, `vantablack`) ported to the semantic keys.

## Light/dark mode

Each in-tree `colors.toml` now declares its mode explicitly:

```toml
mode = "dark"   # or "light"
```

Resolution precedence in `bin/omarchy-theme-set-templates`:

1. `mode` key in `colors.toml` (canonical)
2. legacy `light.mode` marker file shipped beside the theme (back-compat for user themes)
3. background luminance heuristic (last-resort fallback)

The resolved value is exposed in templates as both `{{ mode }}` and `{{ theme_type }}` (kept as an alias so the existing vscode template doesn't need to change).

`bin/omarchy-theme-set-gnome` reads the same field via `tomlq` and falls back to the `light.mode` file when the key is absent, so existing user themes keep their GNOME color-scheme behavior unchanged. The four in-tree light themes (`catppuccin-latte`, `flexoki-light`, `rose-pine`, `white`) drop their `light.mode` marker now that the key carries the same information.

## Color reference

Mapping between Omarchy color names and [base24](https://github.com/Base24/base24).

### Shade ramp

| Name | base24 | ANSI |
| --- | --- | --- |
| `bg` | base00 | - |
| `lighter_bg` | base01 | 0 |
| `selection` | base02 | 8 |
| `muted` | base03 | - |
| `dark_fg` | base04 | - |
| `fg` | base05 | - |
| `light_fg` | base06 | 7 |
| `bright_fg` | base07 | 15 |

### Accent colors

| Name | base24 | ANSI |
| --- | --- | --- |
| `red` | base08 | 1 |
| `orange` | base09 | - |
| `yellow` | base0A | 3 |
| `green` | base0B | 2 |
| `cyan` | base0C | 6 |
| `blue` | base0D | 4 |
| `purple` | base0E | 5 |
| `brown` | base0F | - |

### Extended colors

| Name | base24 | ANSI |
| --- | --- | --- |
| `dark_bg` | base10 | - |
| `darker_bg` | base11 | - |
| `bright_red` | base12 | 9 |
| `bright_yellow` | base13 | 11 |
| `bright_green` | base14 | 10 |
| `bright_cyan` | base15 | 14 |
| `bright_blue` | base16 | 12 |
| `bright_purple` | base17 | 13 |

### Semantic keys

Set independently per theme, not part of the base24 spec:

| Name | Purpose |
| --- | --- |
| `background` | Terminal/app background |
| `foreground` | Terminal/app foreground |
| `accent` | Primary accent color |
| `cursor` | Cursor color |
| `selection_background` | Selected text background |
| `selection_foreground` | Selected text foreground |
| `mode` | `"dark"` or `"light"` (drives GNOME color-scheme and vscode `type`) |

## Demo: Ayu Mirage

A base24 Ayu Mirage palette converted to the Omarchy names.

screenshot-2026-02-07_19-47-44

```toml
mode = "dark"
accent = "#73d0ff"
cursor = "#d9d7ce"
foreground = "#cccac2"
background = "#1f2430"
selection_foreground = "#d9d7ce"
selection_background = "#323844"

bg = "#1f2430"
lighter_bg = "#242936"
selection = "#323844"
muted = "#4A5059"
dark_fg = "#707a8c"
fg = "#cccac2"
light_fg = "#d9d7ce"
bright_fg = "#f3f4f5"

red = "#f28779"
yellow = "#ffad66"
orange = "#ffd173"
green = "#d5ff80"
cyan = "#95e6cb"
blue = "#73d0ff"
purple = "#d4bfff"
brown = "#f27983"
dark_bg = "#171b24"
darker_bg = "#1a1f29"
bright_red = "#ff6666"
bright_yellow = "#ffcc66"
bright_green = "#87d963"
bright_cyan = "#5ccfe6"
bright_blue = "#80b0ff"
bright_purple = "#b692ff"
```

## Compatibility

User themes that still ship `color0..15` keys keep working. The restored alias block maps each ANSI slot to its semantic name (`color1` -> `red`, `color9` -> `bright_red`, etc.) and only fills in gaps, so a theme that defines both an ANSI slot and its semantic counterpart keeps the explicit semantic value.

The four in-tree themes that used ANSI keys (`last-horizon`, `solitude`, `white`, `vantablack`) are ported in this PR, so the alias block exists purely for user-defined themes.

For context, the previous closed PRs: #4526, #4538. Discussion: https://discord.com/channels/1390012484194275541/1399365674832232448/1469128896166494289


default/themed/btop.theme.tpl
```diff
@@ -11,73 +11,83 @@ theme[title]="{{ foreground }}"
...
+
+# Graph gradient colors (spectrum shades from background to foreground)
+theme[gradient_color_0]="{{ bg }}"
+theme[gradient_color_1]="{{ lighter_bg }}"
+theme[gradient_color_2]="{{ selection }}"
+theme[gradient_color_3]="{{ muted }}"
+theme[gradient_color_4]="{{ dark_fg }}"
+theme[gradient_color_5]="{{ fg }}"
+theme[gradient_color_6]="{{ light_fg }}"
+theme[gradient_color_7]="{{ bright_fg }}"
```
bjarneo, line 93: Go through all of these one more time. Selection should probably not be part of the gradients.

bjarneo:
It is a bit much, isn't it.

dhh:
I like the thinking, but the only practical use I see of this is btop? And not sure it's worth asking everyone to come up with that many additional colors just for that. If btop already has a more specific format anyway.

Do you have other apps where base24 is actually used fully?

bjarneo:
I totally get that. 

All in all, it really depends on how much flexibility Omarchy should bring to the table.

It would be fully used if we add automatically 
1. Zed theme
2. VSCode theme
3. Nvim theme

For those using hooks, I know some need more colors for applications such as Spotify, which is outside Omarchy's scope.

By reading the base24 spec, you can map the 8 colors from the base16 spec to the base24 if you don't define the extra 8 (base24) colors.

| Base24 | Name | Fallback (Base16) | ANSI |
| :--- | :--- | :--- | :--- |
| `base10` | `dark_bg` | `base00` | - |
| `base11` | `darker_bg` | `base00` | - |
| `base12` | `bright_red` | `base08` | 9 |
| `base13` | `bright_yellow` | `base0A` | 11 |
| `base14` | `bright_green` | `base0B` | 10 |
| `base15` | `bright_cyan` | `base0C` | 14 |
| `base16` | `bright_blue` | `base0D` | 12 |
| `base17` | `bright_purple` | `base0E` | 13 |

This will usually work for btop. Unsure how well it will work with Zed or VS Code, though.

dhh:
I'd say it would be quite appealing if we can get full vscode themeing syncing via templates. How does that work at the moment, though? Why isn't the 16 we have enough? But I also very much like the idea that a base16, like we're using, can be automatically turned into a 24 algorithmicly. I want to be conscious of keeping the barrier to entry for theme makers low.

bjarneo:
Vscode/zed is a json file you would need to put in a directory. Straightforward to solve.

16 colors are enough for most themes. This is what I do using Aether. What I see for theming nvim, zed, and vscode is that there are not enough colors to make the user interface aesthetically pleasing, meaning certain contrasts do not work well and are a bit unreadable. I can find examples here.

24 algorithmically is also a good approach to this. Finding a shade of a color can be done using bash.
```
for i in {0..15}; do
  val=$((i * 255 / 15))
  printf "#%02x%02x%02x\n" $val $val $val
done
```

Saw this the other day, which was interesting, now, a bit overkill, but an example of creating 256 colors from a 16 color palette (16*16): https://x.com/mitchellh/status/2023823975143665918

Based on my comment: https://github.com/basecamp/omarchy/pull/4541#issuecomment-3902138147, I find it to be a bit much myself. 

One could keep it plain, use 16 colors (as is), and generate shades for the additional 8 colors, while allowing them to be defined by power users.

dhh:
Yeah, it's a good point. Because we're obviously also still using explicit nvim themes. If there was a path where we could get perfect nvim, vscode, and btop themes from the same color base, I'd say that's worth going to 24.

bjarneo:
Then I will continue testing this implementation, which will also include vscode and nvim!

dhh:
The holy grail is that literally everything is generated from 1 colors.toml file.

OldJobobo:
Is it also an explicit goal to preserve the ability for theme authors to override the generated colors when needed? From a design standpoint, generated palettes are often good, but they don’t always match the intended aesthetic or sometimes produce muddy results.

dhh:
We should definitely always allow per-config overrides.

bjarneo:
Step one. Make 8 of the colors optional.

With the 24 colors defined:
screenshot-2026-02-19_20-45-26

With 16 colors defined, and shade auto-generated for the optional 8 colors.
screenshot-2026-02-19_20-44-34

No visual difference.

Example of Ayu Mirage with the optional colors.
```toml
accent = "#73d0ff"
cursor = "#d9d7ce"
foreground = "#cccac2"
background = "#1f2430"
selection_foreground = "#d9d7ce"
selection_background = "#323844"

bg = "#1f2430"
lighter_bg = "#242936"
selection = "#323844"
muted = "#4A5059"
dark_fg = "#707a8c"
fg = "#cccac2"
light_fg = "#d9d7ce"
bright_fg = "#f3f4f5"

red = "#f28779"
yellow = "#ffad66"
orange = "#ffd173"
green = "#d5ff80"
cyan = "#95e6cb"
blue = "#73d0ff"
purple = "#d4bfff"
brown = "#f27983"

# Optional. Will be auto-generated if not defined.
dark_bg = "#171b24"
darker_bg = "#1a1f29"
bright_red = "#ff6666"
bright_yellow = "#ffcc66"
bright_green = "#87d963"
bright_cyan = "#5ccfe6"
bright_blue = "#80b0ff"
bright_purple = "#b692ff"
```

bjarneo:
Step two, add vscode theme.

Used the template from Aether. The difference is more color support coming from the colors.toml. In need of more testing for different color scheme variations. Including light mode.

Note: I do not find a clean way of forcing vscode to change the theme "Omarchy" when the colors are updated. Kill and reopen is scary as a developer might have unsaved work.

Preview
screenshot-2026-02-19_21-19-11

dhh:
Very promising!!

bjarneo:
Step three, add neovim.lua. 

Now, I would recommend we fork aether.nvim v3, move it to omacom, and rename it, and do potential adjustments to that theme.

Example using Ethereal colors.toml
screenshot-2026-02-19_21-45-05

bin/omarchy-theme-set-vscode
```diff
@@ -3,22 +3,68 @@
 # Sync Omarchy theme to VS Code, VSCodium, and Cursor
 
 VS_CODE_THEME="$HOME/.config/omarchy/current/theme/vscode.json"
+GENERATED_THEME="$HOME/.config/omarchy/current/theme/vscode-theme.json"
```
bjarneo, line 7: Look into this. See line 5.

bin/omarchy-theme-set-vscode
```diff
@@ -34,6 +80,5 @@ set_theme() {
   fi
 }
 
-set_theme "code" "$HOME/.config/Code/User/settings.json" "$HOME/.local/state/omarchy/toggles/skip-vscode-theme-changes"
-set_theme "codium" "$HOME/.config/VSCodium/User/settings.json" "$HOME/.local/state/omarchy/toggles/skip-codium-theme-changes"
```
bjarneo: Add back codium

default/themed/vscode-theme.json.tpl
```diff
@@ -0,0 +1,1285 @@
...
+        "sideBarSectionHeader.foreground": "{{ foreground }}",
+        "sideBarSectionHeader.border": "{{ muted }}40",
+
+        "minimap.findMatchHighlight": "{{ accent }}80",
+        "minimap.selectionHighlight": "{{ accent }}60",
+        "minimap.errorHighlight": "{{ red }}",
+        "minimap.warningHighlight": "{{ yellow }}",
+        "minimap.background": "{{ background }}",
+        "minimap.selectionOccurrenceHighlight": "{{ accent }}40",
+        "minimap.foregroundOpacity": "#000000c0",
```
bjarneo: Look into this one

default/themed/vscode-theme.json.tpl
```diff
@@ -0,0 +1,1285 @@
...
+        "editorWarning.border": "{{ yellow }}00",
+        "editorInfo.foreground": "{{ cyan }}",
+        "editorInfo.background": "{{ cyan }}15",
+        "editorInfo.border": "{{ cyan }}00",
+        "editorHint.foreground": "{{ green }}",
+        "editorHint.border": "{{ green }}00",
+        "problemsErrorIcon.foreground": "{{ red }}",
+        "problemsWarningIcon.foreground": "{{ yellow }}",
+        "problemsInfoIcon.foreground": "{{ cyan }}",
+        "editorUnnecessaryCode.opacity": "#00000080",
```
bjarneo: Look into this one

bin/omarchy-theme-set-vscode
```diff
@@ -3,22 +3,68 @@
...
+  cp "$GENERATED_THEME" "$ext_dir/themes/omarchy-color-theme.json"
+
+  cat > "$ext_dir/package.json" <<EOF
+{
+    "name": "omarchy-theme",
+    "displayName": "Omarchy",
+    "description": "Omarchy color theme",
+    "publisher": "local",
+    "version": "1.0.0",
+    "engines": { "vscode": "^1.70.0" },
```
bjarneo, line 81: Be certain about the version

bjarneo (commented):
Look into this

default/themed/neovim.lua.tpl
```diff
@@ -0,0 +1,56 @@
...
+        accent               = "{{ accent }}",
+        cursor               = "{{ cursor }}",
+        foreground           = "{{ foreground }}",
+        background           = "{{ background }}",
+        selection             = "{{ selection }}",
+        selection_foreground = "{{ selection_foreground }}",
+        selection_background = "{{ selection_background }}",
+      },
+    },
+    config = function(_, opts)
```
bjarneo: This is not needed

bjarneo (commented):
Add comment

dpan:
This may be relevant in case you haven't seen it: https://x.com/mitchellh/status/2023823975143665918

bjarneo:
> This may be relevant in case you haven't seen it: https://x.com/mitchellh/status/2023823975143665918

https://github.com/basecamp/omarchy/pull/4541#issuecomment-3926360969

Yes. Seen it. Cool approach. More or less 16 shades of color. 

The approach for this is Base16 + 8 shades. Though, using a better naming convention.

bjarneo:
Here are a few existing themes that use colors.toml to generate Neovim color schemes to get a feeling of how this would look.

## Rose pine
screenshot-2026-02-22_10-14-51


## Hackerman
screenshot-2026-02-22_10-02-38

## Ayu Mirage (only a local theme)
screenshot-2026-02-22_10-01-49

## Osaka jade
screenshot-2026-02-22_10-01-11

## Miasma
screenshot-2026-02-22_10-00-50

## Gruvbox
screenshot-2026-02-22_10-00-28

bjarneo:
Similar for vscode
screenshot-2026-02-22_10-19-59
screenshot-2026-02-22_10-19-32
screenshot-2026-02-22_10-18-27
screenshot-2026-02-22_10-17-54
screenshot-2026-02-22_10-17-10
screenshot-2026-02-22_10-14-51

bjarneo:
screenshot-2026-02-22_10-24-33
screenshot-2026-02-22_10-24-16
screenshot-2026-02-22_10-24-00
screenshot-2026-02-22_10-23-27
screenshot-2026-02-22_10-23-01
screenshot-2026-02-22_10-22-48

bjarneo:
Consideration: A more muted color for comments

bjarneo:
Muted comments

screenshot-2026-02-22_10-37-39
screenshot-2026-02-22_10-41-05

bjarneo:
For my own sake (AI):

# Theming

It uses a color system inspired by [Base24](https://github.com/Base24/base24), with human-readable names instead of the Base24 `base00`-`base17` identifiers. Each theme is defined in a single `colors.toml` file that drives every themed application through template files.

## Color reference

### Semantic keys

Set independently per theme. These are not part of the Base24 spec:

| Name | Purpose |
|---|---|
| `background` | Terminal/app background |
| `foreground` | Terminal/app foreground |
| `accent` | Primary accent color |
| `cursor` | Cursor color |
| `selection_background` | Selected text background |
| `selection_foreground` | Selected text foreground |

### Shade ramp

Grayscale tones from darkest background to brightest foreground:

| Name | Base24 | ANSI | Purpose |
|---|---|---|---|
| `bg` | base00 | - | Editor/panel background |
| `lighter_bg` | base01 | 0 | Subtle background variation |
| `selection` | base02 | 8 | Selection/highlight background |
| `muted` | base03 | - | Comments, inactive text |
| `dark_fg` | base04 | - | Dark foreground, line numbers |
| `fg` | base05 | - | Default foreground text |
| `light_fg` | base06 | 7 | Light foreground |
| `bright_fg` | base07 | 15 | Brightest foreground |

### Accent colors

| Name | Base24 | ANSI | Typical use |
|---|---|---|---|
| `red` | base08 | 1 | Errors, deletions |
| `yellow` | base09 | 3 | Warnings, modifications |
| `orange` | base0A | - | Constants, numbers |
| `green` | base0B | 2 | Strings, additions, success |
| `cyan` | base0C | 6 | Support functions, builtins |
| `blue` | base0D | 4 | Functions, links, accent |
| `purple` | base0E | 5 | Keywords, control flow |
| `brown` | base0F | - | Embedded language punctuation |

### Extended colors (bright variants)

These are optional. If omitted from `colors.toml`, they are auto-generated by lightening the base accent colors by 20% (dark backgrounds are darkened from `bg` at 75% and 50%):

| Name | Base24 | ANSI | Derived from |
|---|---|---|---|
| `dark_bg` | base10 | - | `bg` darkened to 75% |
| `darker_bg` | base11 | - | `bg` darkened to 50% |
| `bright_red` | base12 | 9 | `red` lightened 20% |
| `bright_yellow` | base13 | 11 | `yellow` lightened 20% |
| `bright_green` | base14 | 10 | `green` lightened 20% |
| `bright_cyan` | base15 | 14 | `cyan` lightened 20% |
| `bright_blue` | base16 | 12 | `blue` lightened 20% |
| `bright_purple` | base17 | 13 | `purple` lightened 20% |

bjarneo:
I think it is ready to review as is

shvedes:
@bjarneo  Genuine question: is there any specific reason to stick to the exact Base24 format instead of, say, Omarchy’s custom definitions? The initial `colors.toml` idea is technically very expandable, and extending it doesn’t introduce complications beyond a one-time template update. My point is that the more colors we potentially have in `colors.toml`, the higher quality the resulting theme can be.

I [implemented](https://github.com/shvedes/vibranium/blob/master/themes/aamis/colors-extended.toml) this in my personal setup, and it works like a charm:

screenshot
screenshot

Themes generated from such a template look as if they were manually adjusted: sleek and visually consistent.

While we won’t always use every color across all templates, programs, or configs, having them available becomes very useful when building larger or more complex things, like a CSS-heavy setup (for example, an Obsidian theme). Compared to the current Obsidian theme generated from `colors.toml`, the new ones look much more like they were designed from the ground up.

Considering that `omarchy-theme-set-templates` can also be expanded significantly to support not just RGB but other formats (see [this loop](https://github.com/shvedes/vibranium/blob/02a7060ae2fe23faad558c806874aeb42dd53379/bin/vb-theme-set-templates#L82)), this opens up a strong opportunity for high-quality automatic theming with little to no extra effort.

bjarneo:
@shvedes I'll come back to you

bjarneo:
> @bjarneo Genuine question: is there any specific reason to stick to the exact Base24 format instead of, say, Omarchy’s custom definitions? The initial `colors.toml` idea is technically very expandable, and extending it doesn’t introduce complications beyond a one-time template update. My point is that the more colors we potentially have in `colors.toml`, the higher quality the resulting theme can be.
> 
> I [implemented](https://github.com/shvedes/vibranium/blob/master/themes/aamis/colors-extended.toml) this in my personal setup, and it works like a charm:
> 
> screenshot screenshot
> 
> Themes generated from such a template look as if they were manually adjusted: sleek and visually consistent.
> 
> While we won’t always use every color across all templates, programs, or configs, having them available becomes very useful when building larger or more complex things, like a CSS-heavy setup (for example, an Obsidian theme). Compared to the current Obsidian theme generated from `colors.toml`, the new ones look much more like they were designed from the ground up.
> 
> Considering that `omarchy-theme-set-templates` can also be expanded significantly to support not just RGB but other formats (see [this loop](https://github.com/shvedes/vibranium/blob/02a7060ae2fe23faad558c806874aeb42dd53379/bin/vb-theme-set-templates#L82)), this opens up a strong opportunity for high-quality automatic theming with little to no extra effort.

Fair question, and worth clarifying first: the schema in this PR isn't strict Base24 anymore. The commit messages reference Base24 because the bright_* set was modeled on it, but the file already includes Omarchy-specific keys: bg, lighter_bg, selection, muted, dark_fg, light_fg, bright_fg, orange, brown. We're past Base24 in spirit, so the real question is how much further to go, not whether to leave it.

Where I'm hesitant to extend: every new key is a decision multiplied across 30 themes. If a value is copy-pasted or can be computed from existing colors, it should live in omarchy-theme-set-templates as a derivation rather than in colors.toml. That's exactly why bright_*, dark_bg, darker_bg, orange, and brown have lighten/darken or fallback rules. Keeps the per-theme surface small and consistent.

Supporting other formats isn't a bad thing. Something to consider.

Anyhow, this is extendable if you need more colors later on.

Could you share your colors.toml and the Obsidian/CSS templates that consume it?

bjarneo:
Here's an example where I removed the vscode and btop files for all native omarchy themes:

catppuccin
catppuccin-latte
ethereal
everforest
flexoki-light
gruvbox
hackerman
kanagawa
last-horizon
lumon
matte-black
miasma
nord
osaka-jade
retro-82
ristretto
rose-pine
solitude
tokyo-night
vantablack
white

shvedes:
@bjarneo

> Could you share your colors.toml and the Obsidian/CSS templates that consume it?

Yeah, sure!

So, first of all, I must say that I'm not using Omarchy itself, but my personal Omarchy-like setup, where I borrowed Omarchy's theme structure because it was well thought out and my previous implementation was pretty bad at the time. **Then** I improved it as I described above. So I'll speak from this point on.

First, I keep both `colors.toml` for compatibility with Omarchy themes and `colors-extended.toml` with extended colors. Here's a [`colors-extended.toml`](https://github.com/shvedes/vibranium/blob/master/themes/lumon/colors-extended.toml) example, which is a port of Omarchy's Lumen theme. And [here](https://github.com/shvedes/vibranium/blob/master/default/themed/obsidian.css.extended.tpl) is the `obsidian.css` example with CSS color calculations for smooth background shades.

[vb-theme-set-templates](https://github.com/shvedes/vibranium/blob/master/bin/vb-theme-set-templates) is the reimplementation of `omarchy-theme-set-templates` with added extra keywords and color formats, including HSL. At the top, it has a brief explanation of how it works, what custom keywords are available, and what formats it supports, with examples.

I should also point out that I replaced the temporary `sed` script with an [awk script](https://github.com/shvedes/vibranium/blob/master/awk/vb-theme-set-templates.awk) to move it out of the loop, making it faster.

Now, we cannot convert HEX colors defined in the TOML file to HSL format exactly, so there will always be a slight, barely noticeable color shift. I kept this because while we're losing a bit in subtlety, we're potentially gaining a lot of room for automatic color calculations in theme template files that support and need them.

I'm also considering refactoring it so the template generator understands what type of theme it generates configs for: light or dark, so it can properly handle background color calculations for light themes too. Right now, background shade calculations work much better for dark themes. In light themes, it can generate pretty awkward background shifts, like in this example:

screenshot-2026-05-15_21-05-01


I also have [this](https://github.com/shvedes/vibranium/blob/master/wiki/Themes%20Architecture.md) wiki page that describes how it works under the hood. And [here](https://github.com/shvedes/vibranium/blob/master/wiki/Making%20Your%20Own%20Theme.md) you can see how it works in action with **zero** preconfigured configs except for the metadata and the `colors-extended.toml` itself.

default/themed/vscode-theme.json.tpl
```diff
@@ -0,0 +1,1340 @@
...
+        "editorWarning.border": "{{ yellow }}00",
+        "editorInfo.foreground": "{{ blue }}",
+        "editorInfo.background": "{{ blue }}15",
+        "editorInfo.border": "{{ blue }}00",
+        "editorHint.foreground": "{{ cyan }}",
+        "editorHint.border": "{{ cyan }}00",
+        "problemsErrorIcon.foreground": "{{ red }}",
+        "problemsWarningIcon.foreground": "{{ yellow }}",
+        "problemsInfoIcon.foreground": "{{ blue }}",
+        "editorUnnecessaryCode.opacity": "{{ background }}80",
```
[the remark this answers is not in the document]
bjarneo: Nope https://code.visualstudio.com/api/references/theme-color#editor-colors

default/themed/vscode-theme.json.tpl
```diff
@@ -0,0 +1,1340 @@
...
+        "sideBarSectionHeader.foreground": "{{ foreground }}",
+        "sideBarSectionHeader.border": "{{ muted }}40",
+
+        "minimap.findMatchHighlight": "{{ accent }}80",
+        "minimap.selectionHighlight": "{{ accent }}60",
+        "minimap.errorHighlight": "{{ red }}",
+        "minimap.warningHighlight": "{{ yellow }}",
+        "minimap.background": "{{ background }}",
+        "minimap.selectionOccurrenceHighlight": "{{ accent }}40",
+        "minimap.foregroundOpacity": "{{ background }}c0",
```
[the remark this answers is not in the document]
bjarneo: Not according to the docs: https://code.visualstudio.com/api/references/theme-color#minimap-colors

bin/omarchy-theme-set-templates
```diff
@@ -165,12 +165,22 @@ if [[ -f $COLORS_FILE ]]; then
...
+  # Resolve theme mode (dark|light) with this precedence:
+  #   1. `mode` key in colors.toml
+  #   2. legacy `light.mode` file shipped beside the theme
+  #   3. background luminance auto-detect
+  # `theme_type` is kept as an alias so existing templates ({{ theme_type }}) keep working.
+  [[ ${THEME_COLORS[mode]} ]] || THEME_COLORS[mode]="${THEME_COLORS[theme_type]}"
+  if [[ ! ${THEME_COLORS[mode]} ]]; then
+    if [[ -f $NEXT_THEME_DIR/light.mode ]]; then
+      THEME_COLORS[mode]="light"
+    else
```
bjarneo: Keep or remove the auto-detect?

themes/ayu-mirage-dark/colors.toml
```diff
@@ -0,0 +1,36 @@
+mode = "dark"
```
bjarneo: Removing, committed by mistake

themes/hackerman2/colors.toml
```diff
@@ -0,0 +1,34 @@
+mode = "dark"
```
bjarneo: Removing, committed by mistake

themes/miasma2/colors.toml
```diff
@@ -0,0 +1,34 @@
+mode = "dark"
+
```
bjarneo: Removing, committed by mistake

bin/omarchy-theme-set-templates
shvedes: I'd keep this. It's simple yet convenient

config/omarchy/themed/alacritty.toml.tpl.sample
```diff
@@ -66,21 +70,21 @@ text = "{{ selection_foreground }}"
...
-cyan = "{{ color6 }}"
-white = "{{ color7 }}"
+color0 = "{{ bg }}"
+color1 = "{{ red }}"
+color2 = "{{ green }}"
+color3 = "{{ yellow }}"
+color4 = "{{ blue }}"
+color5 = "{{ purple }}"
+color6 = "{{ cyan }}"
+color7 = "{{ fg }}"
```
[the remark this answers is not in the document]
bjarneo: Fix coming

PR #4543, opened 2026-02-07
Remove the aether.nvim dependency

bjarneo:


themes/vantablack/neovim.lua
```diff
@@ -1,11 +1,7 @@
 return {
   {
-    "bjarneo/vantablack.nvim",
+    dir = "~/Code/vantablack.nvim",
```
bjarneo: I am getting tired

dhh:
That did it! No more colors. But thinking we should probably flip the brightness, so comments are grey and code is white:

image

PR #4555, opened 2026-02-08
Ensure Walker can't just die and stay dead

dhh:
Before, Walker could crash, and the service wouldn't be restarted, so user would be stuck with cold-boot lag.


dhh:
cc @abenz1267, I found out that my walker service had been dead for days due to a crash. No wonder Walker was sluggish to start 😄

abenz1267:
Do you have a clue why it even crashed?

dhh:
No. The crash happened on a machine with a hybrid gpu, but it was running in igpu mode.

Issue #4558, opened 2026-02-09
nvim crashes terminal when launched via `n` bash function/alias


grumpydumpty:
### System details

AMD Ryzen 9 PRO 8945Hz, AMD Radeon 780M Graphics, Omarchy 3.3.3

## Problem

`nvim` crashes terminal when launched via `n` bash function/alias.

When launched via `SUPER` + `SHIFT` + `N`, `nvim` opens fine.

## Suspected cause

nvim calls a function in `~/.local/share/omarchy/default/bash/aliases`:

`n() { if [ "$#" -eq 0 ]; then nvim .; else nvim "$@"; fi; }`

This causes a recursive function where nvim calls itself endlessly. I changed it to:

`n() { if [ "$#" -eq 0 ]; then command nvim; else command nvim "$@"; fi; }`

Which causes the function to exit its recursion and launch the actual editor.

## Proposed solution

```diff
-n() { if [ "$#" -eq 0 ]; then nvim .; else nvim "$@"; fi; }
+n() { if [ "$#" -eq 0 ]; then command nvim; else command nvim "$@"; fi; }
```

## Testing

I have been using for a number of days now without issue. I upgraded to Omarchy 3.3.3 this morning and the issue reappeared. I reapplied to proposed solution and the issue is gone.

## Attribution

Originally diagnosis and propose solution, taken from here: [nvim crashing terminal : r/omarchy](https://www.reddit.com/r/omarchy/comments/1pciz7g/nvim_crashing_terminal/).

avenkidur:
This doesn't make sense to me. There shouldn't be a difference between `nvim` and `command nvim`, but I also don't how your change is resolving your problem.

`command nvim` is the shell built-in function `command` which substitutes `nvim` the actual command that would be run after searching in order aliases, functions, built-in functions, and then $PATH

---
I did load bash and test the behavior you described and I did not experience any kind of loop or crash. (Omarchy 3.3.3)

Could you run `command -V nvim && command -V n` to check what is being executed?

dhh:
I think you must have added your own nvim function that's calling n. I can't replicate any crash behavior either. Never heard of it being a problem for others as well.

grumpydumpty:
I didn't touch the n function.

See the link attached for others having the problem.

dhh:
All good. I've merged it. Does no harm.

PR #4559, opened 2026-02-09
Fix for issue #4558

grumpydumpty:
Implementation of the proposed solution in the bug report for issue #4558

### DIFF
--- a/default/bash/aliases
+++ b/default/bash/aliases
@@ -35,7 +35,7 @@ alias ....='cd ../../..'
 alias c='opencode'
 alias d='docker'
 alias r='rails'
-n() { if [ "$#" -eq 0 ]; then command nvim ; else command nvim "$@"; fi; }
+n() { if [ "$#" -eq 0 ]; then command nvim . ; else command nvim "$@"; fi; }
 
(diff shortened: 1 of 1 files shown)


default/bash/aliases
```diff
@@ -34,7 +34,7 @@ alias ....='cd ../../..'
 alias c='opencode'
 alias d='docker'
 alias r='rails'
-n() { if [ "$#" -eq 0 ]; then nvim .; else nvim "$@"; fi; }
+n() { if [ "$#" -eq 0 ]; then command nvim ; else command nvim "$@"; fi; }
```
Daan5556: You should add the `.` after the nvim command so that if no path is provided, it will default to the current directoy.

`n() { if [ "$#" -eq 0 ]; then command nvim . ; else command nvim "$@"; fi; }`

grumpydumpty: I am not a fan of that behaviour but since that is the existing behave I added the `.` I removed.

dhh:
Appreciate the suggestion, but I don't think the original issue is reproducible on stock Omarchy.

PR #4561, opened 2026-02-09
Enable bluetooth at login screen

jmthvt:
Reproduce changes described in https://github.com/basecamp/omarchy/issues/1744#issuecomment-3452090960

Fixes https://github.com/basecamp/omarchy/issues/1744

### DIFF
--- a/migrations/1770646264.sh
+++ b/migrations/1770646264.sh
@@ -10,4 +10,4 @@ sudo tee /etc/mkinitcpio.conf.d/uhid_module.conf <<EOF >/dev/null
 MODULES+=(uhid)
 EOF
 
-sudo mkinitcpio -P
+sudo limine-mkinitcpio -P


pomartel:
I ran the migration and got this message when running `sudo mkinitcpio -P` :
```
==> ERROR: No presets found in /etc/mkinitcpio.d
==> WARNING: This does not update Limine boot entries.
    Use 'limine-mkinitcpio' or 'limine-update' instead.
==> Would you like to run 'limine-mkinitcpio' now? [Y/n]: Y
```
I don't know how this works but shouldn't the script run the `limine-mkinitcpio` instead?

pomartel:
Also, `mkinitcpio-bluetooth` does not seem to be an official package. It's in the AUR.

pomartel:
I just can't get this PR to work with the Lofree Flow 84 bluetooth keyboard. @dhh has the same keyboard, maybe he'll have more luck or better debugging ability!

jmthvt:
> I ran the migration and got this message when running `sudo mkinitcpio -P` :
> 
> ```
> ==> ERROR: No presets found in /etc/mkinitcpio.d
> ==> WARNING: This does not update Limine boot entries.
>     Use 'limine-mkinitcpio' or 'limine-update' instead.
> ==> Would you like to run 'limine-mkinitcpio' now? [Y/n]: Y
> ```
> 
> I don't know how this works but shouldn't the script run the `limine-mkinitcpio` instead?

I just checked my bash history, and you're right, I ran `sudo limine-mkinitcpio -P`, my mistake. Commit incoming.

jmthvt:
> Also, `mkinitcpio-bluetooth` does not seem to be an official package. It's in the AUR.

Is that an issue? I saw [omarchy-pkg-add](https://github.com/basecamp/omarchy/blob/ff234bba1f44f9ce0ee61dd1767dec03bb8fff8b/AGENTS.md?plain=1#L28) supports AUR.

pomartel:
You should use `omarchy-pkg-aur-add` for AUR packages. Also, `omarchy-base.packages` only work for official packages.

pomartel:
And most important, test your PR before submitting it!

jmthvt:
As per the [mkinitcpio-bluetooth](https://github.com/irreleph4nt/mkinitcpio-bluetooth/?tab=readme-ov-file#troubleshooting) repo, you might need to load another module depending on your keyboard, some commands that might help:

```
$ journalctl | grep bluetooth

$ sudo dmesg | grep bluetooth
```

jmthvt:
OK I see the package has to be added to https://github.com/omacom-io/omarchy-pkgs/, I will work on it. Switching the PR to draft for now.

dhh:
Please have a fully functional branch in your own fork before opening a PR. Otherwise we'll clog up the system here with things that don't work and aren't ready.

PR #4565, opened 2026-02-09
Handle network click on ethernet-only systems

skibidiandulka:
Waybar uses a single click handler for the network module (wifi and ethernet). On ethernet-only machines, the current handler launches the wifi TUI, which fails because no wifi adapter exists.

This change detects physical adapters via /sys/class/net.
- If a wifi adapter exists: launch impala as before
- Else if a physical non-wifi adapter exists: open an ethtui
- Else: show a notification

Because there was no ethernet TUI only for wifi or bluetooth I have created my own. The project repositroy is accessible at: https://github.com/skibidiandulka/ethtui . It is integrated in to OMARCHY the same way as bluetui and impala.

### DIFF
--- a/bin/omarchy-menu
+++ b/bin/omarchy-menu
@@ -207,7 +207,7 @@ show_font_menu() {
 }
 
 show_setup_menu() {
-  local options="  Audio\n  Wifi\n󰂯  Bluetooth\n󱐋  Power Profile\n  System Sleep\n󰍹  Monitors"
+  local options="  Audio\n  Wifi\n󰈀  Ethernet\n󰂯  Bluetooth\n󱐋  Power Profile\n  System Sleep\n󰍹  Monitors"
   [ -f ~/.config/hypr/bindings.conf ] && options="$options\n  Keybindings"
   [ -f ~/.config/hypr/input.conf ] && options="$options\n  Input"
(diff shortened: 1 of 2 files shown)


skibidiandulka:
I see that there is no TUI for ethernet and the impala doesnt have ethernet support. I am going to attempt to create a ethernet tui.

skibidiandulka:
I have created my own ethernet TUI called reasonably: ethtui. The project repository can by accessed here: https://github.com/skibidiandulka/ethtui

skibidiandulka:
I should probably add a option that detects if ethernet is active and it rather displays the ethernet TUI then the wifi TUI.

skibidiandulka:
I have a problem. Bluetui and impala are official arch packages and can be installed via pacman. What should I do, what are my options???

skibidiandulka:
image
I have created such TUI. (I am going to get rid of the white text there...)

dhh:
Appreciate the exploration here, but this is too small of an issue to accept anywhere near this amount of code. I like the idea of having basically Impala but for ethernet. Especially if it wraps up things like bonding multiple links. But that has to first be created as a separate package that we can then consider using if it's useful enough. Thanks again for suggesting, though!

PR #4573, opened 2026-02-10
Windows vm: automatically add kvm support based on CPU vendor

Mrid22:
Changed the script to automatically add kvm support if not found instead of exiting.


dhh:
This is sloppy work. There are comments left in, the conditional doesn't work.

PR #4574, opened 2026-02-10
Dropbox: dont install file explorer integration if explorer is not present

Mrid22:
In the dropbox install script, there is no need to install nautilus-dropbox if nautilus is not present on the system ( people might use other file explorers ), so I made it conditional and added integrations for other explorers as well


dhh:
Appreciate the suggestion, but nautilus is the Omarchy default, so I think it's fair that things are tied like that. It should be easy to change for anyone who switches. Don't really want to maintain a big list of this.

PR #4575, opened 2026-02-10
Fix dictation push-to-talk dropping stop event on early modifier release

jra3:
## Summary

- The `binddr` (release) for `SUPER CTRL + X` requires both modifiers to still be held when X is released. If you release Super or Ctrl before X, the modifier mask no longer matches — the stop event never fires and recording gets stuck.
- Uses a Hyprland submap instead: `SUPER+CTRL+X` starts recording and enters a `dictation` submap where the release of X alone (no modifier requirement) stops recording and exits the submap. This makes key release order irrelevant.

## Test plan

- [ ] Press `Super+Ctrl+X`, hold, release X first → recording stops (same as before)
- [ ] Press `Super+Ctrl+X`, hold, release Super first then X → recording stops (was broken)
- [ ] Press `Super+Ctrl+X`, hold, release Ctrl first then X → recording stops (was broken)
- [ ] Verify other keybindings still work normally while not in dictation submap


dhh:
This doesn't seem to be working for me. If I release X first, it doesn't stop recording. I have to manually push X again.

dhh:
Switched to just having it be a toggle instead of push to talk.

PR #4581, opened 2026-02-11
fix(aur): add -a flag to yay command to assume AUR packages

abenz1267:
Hey,

this should (!!!) fix https://github.com/basecamp/omarchy/issues/4577, but requires testing/confirmation.

Just added this flag: `-a --aur              Assume targets are from the AUR`

Regards


abenz1267:
i just saw there's another PR: https://github.com/basecamp/omarchy/pull/4578

different solutions for the problem.

dhh:
This didn't actually work. It's still pulling from our outdated version instead of directly from AUR.

PR #4519, opened 2026-02-06
Background selector

dhms013:
Details :
1. Added [default/elephant/omarchy_background_selector.lua](https://github.com/dhms013/omarchy/blob/ce2cb2c697490ae1e066f0bc15ec9a05967b4218/default/elephant/omarchy_background_selector.lua) as elephant's provider
2. Edit Style > background in [bin/omarchy-menu](https://github.com/dhms013/omarchy/blob/ce2cb2c697490ae1e066f0bc15ec9a05967b4218/bin/omarchy-menu#L191)  => access the elephant [menus:omarchyBackgroundSelector](https://github.com/dhms013/omarchy/blob/ce2cb2c697490ae1e066f0bc15ec9a05967b4218/bin/omarchy-menu#L203)
3. Edit [bin/omarchy-menu](https://github.com/dhms013/omarchy/blob/ce2cb2c697490ae1e066f0bc15ec9a05967b4218/bin/omarchy-menu#L597) => add background selector as default query (write 'background' in omarchy-menu's main page, then Enter)
4. Add migration file [1770380577.sh](https://github.com/dhms013/omarchy/blob/ce2cb2c697490ae1e066f0bc15ec9a05967b4218/migrations/1770380577.sh)
5. Edit [bin/omarchy-theme-set](https://github.com/dhms013/omarchy/blob/ce2cb2c697490ae1e066f0bc15ec9a05967b4218/bin/omarchy-theme-set#L54) => add omarchy-restart-walker
6. Edit [install/config/walker-elephant.sh](install/config/walker-elephant.sh) => add new [symlink](https://github.com/dhms013/omarchy/blob/2b3789f25829e825bd77d4cab477e970ff9eb2a6/install/config/walker-elephant.sh#L26)

The goal :
- Add option for user to select the background using default shortcut that exec omarchy-theme-bg-next, or using 'UI' selector from omarchy-menu > style > background

https://github.com/user-attachments/assets/6728aca0-d808-41f7-89b7-fb683aede4a6

PR #4584, opened 2026-02-12
background selector v2

dhms013:
it's the v2 of background selector in my PR of #4519 

Key of difference :
1. Using TUI
2. Add chafa as dependency => [migrations/1770846792.sh](https://github.com/dhms013/omarchy/blob/0546d18cc48c010ee0d7970dd7d28adc1afc4cfd/migrations/1770846792.sh)
3. Follow the previous method, omarchy-menu > style > background => access [omarchy-theme-bg-select](https://github.com/dhms013/omarchy/blob/0546d18cc48c010ee0d7970dd7d28adc1afc4cfd/bin/omarchy-theme-bg-select)
4. Follow DHH's request regarding [layout](https://github.com/basecamp/omarchy/pull/4519#issuecomment-3871373493) (not the same, but at least preview show multiple images)

Downside : 
1. Not as snappy as #4519 
2. Doesn't have 'search' feature 

https://github.com/user-attachments/assets/2bf7301b-0e5f-4044-a7ed-d738402edd33

### DIFF
--- a/bin/omarchy-menu
+++ b/bin/omarchy-menu
@@ -148,7 +148,10 @@ show_screenrecord_menu() {
   *"With desktop audio") omarchy-cmd-screenrecord --with-desktop-audio ;;
   *"With desktop + microphone audio") omarchy-cmd-screenrecord --with-desktop-audio --with-microphone-audio ;;
   *"With desktop + microphone audio + webcam")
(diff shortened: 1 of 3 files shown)


dhh:
Good exploration. But seeing this, I prefer your v1. Let's go with that.

PR #4586, opened 2026-02-12
Clean up icon spacing in Update menu

horaceko:
Before:
screenshot-2026-02-12_10-26-21

After:
screenshot-2026-02-12_10-28-46

This change makes use of the ["thin space" Unicode character](https://unicode-explorer.com/c/2009) to handle the spacing. A regular (full) space is too large, i.e.

screenshot-2026-02-12_10-37-22

### DIFF
--- a/bin/omarchy-menu
+++ b/bin/omarchy-menu
@@ -481,7 +481,7 @@ show_remove_elixir_menu() {
 }
 
 show_update_menu() {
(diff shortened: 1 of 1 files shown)


Mrid22:
what theme is that?

horaceko:
> what theme is that?

[Archwave](https://github.com/davidguttman/archwave)

PR #4588, opened 2026-02-12
Open current terminal on activity icon right click instead of Alacritty

nialsdelacruz:


pomartel:
This is on purpose. You can read DHH comment here : https://github.com/basecamp/omarchy/pull/4350#issuecomment-3804418864

nialsdelacruz:
> This is on purpose. You can read DHH comment here : https://github.com/basecamp/omarchy/pull/4350#issuecomment-3804418864

Oh, that makes sense. I didn't see that other PR before. Thanks.

PR #4596, opened 2026-02-13
Add caligula tui as replacement for dd

Mrid22:
Changed the iso2sd function to now calls caligula, which is a nice tui for disk burning, alot better experience imo, plus it now calls omarchy-disk-select to ease that part of the process as well


dhh:
I don't think we need to add an additional dependency here for iso2sd. The dd command is great as is. But I do like the idea of selecting the drive!

PR #4612, opened 2026-02-15
Add paste queue for sequential copy/paste

emmgeede:
## Summary

- Adds `omarchy-paste-queue`, a FIFO paste queue inspired by macOS Pastebot
- Copy multiple items sequentially, then paste them back one at a time in order
- A floating panel (ghostty window) shows queued items with live updates
- Bindings: **Super+Alt+C** (queue copy), **Super+Alt+V** (queue paste), **Super+Alt+X** (clear queue)

## Changes

- `bin/omarchy-paste-queue` — the paste queue script
- `default/hypr/bindings/clipboard.conf` — keybindings added
- `default/hypr/apps/paste-queue.conf` — window rules for the floating panel
- `default/hypr/apps.conf` — sources the new window rules
- `install/omarchy-base.packages` — adds `inotify-tools` dependency (for live panel updates)

## How it works

1. **Super+Alt+C** copies the current selection and pushes it onto the queue. A floating panel appears showing all queued items.
2. **Super+Alt+V** pastes the oldest item from the queue (FIFO) and removes it. When the queue is empty, the panel closes automatically.
3. **Super+Alt+X** clears the entire queue and closes the panel.

The queue is stored in `$XDG_RUNTIME_DIR` (tmpfs, cleared on reboot). The panel uses `inotifywait` to refresh in real time when items are added or removed.

## Test plan

- [ ] Copy several items with Super+Alt+C — panel should appear and list them
- [ ] Paste with Super+Alt+V — items should paste in FIFO order
- [ ] Panel should auto-close when last item is pasted
- [ ] Super+Alt+X should clear the queue and close the panel
- [ ] Verify no conflicts with existing Super+C/V/X clipboard bindings

### DIFF
--- a/bin/omarchy-paste-queue
+++ b/bin/omarchy-paste-queue
@@ -1,0 +1,129 @@
+#!/bin/bash
+# omarchy-paste-queue: Sequential copy/paste (FIFO) like macOS Pastebot
+# Usage: omarchy-paste-queue push|paste|clear|panel
+
+QUEUE_DIR="${XDG_RUNTIME_DIR:-/tmp}/paste-queue-${UID}"
+COUNTERFILE="${XDG_RUNTIME_DIR:-/tmp}/paste-queue-counter"
+PIDFILE="${XDG_RUNTIME_DIR:-/tmp}/paste-queue-panel.pid"
+PANEL_TITLE="paste-queue-panel"
+
+next_id() {
+  local n=0
+  [[ -f "$COUNTERFILE" ]] && n=$(cat "$COUNTERFILE")
+  n=$((n + 1))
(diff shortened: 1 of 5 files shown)


dhh:
I think this is pretty clear, but that it would be better as an AUR package, like omarchy-seq-paste or something like that. It's not something I've ever needed myself, but the example on the Pastebot site with the car stats for the table is indeed pretty cool. I could definitely see installing a package for that if I was ever in need of something like this. Thanks for working on it!

emmgeede:
Created an AUR package for this: [omarchy-seq-paste](https://aur.archlinux.org/packages/omarchy-seq-paste)

`yay -S omarchy-seq-paste`

PR #4620, opened 2026-02-16
Faster/Improved Shutdown Procedure

SerrpentDagger:
## Summary
The current shutdown procedure is simple, but inflexible, with the following drawbacks:
- If no windows are open, it waits 2 seconds unnecessarily;
- If a window brings up a confirmation dialogue (e.g: Pinta with unsaved changes), it will just be killed, leading to a potential loss of state or work.

This PR addresses those issues with the following approach:
- Ask all windows to close, asynchronously;
- Check if there are any left;
  - If so, shutdown immediately;
  - If not, bring up a TUI with a live list of hanging windows, and the option to proceed immediately or cancel shutdown;
  - Shutdown proceeds as soon as all windows close, or after max 15 seconds if no user input.

The TUI is presented with `omarchy-launch-floating-terminal-with-presentation` to look nice, but having to press a key to proceed with shutdown didn't seem sensible. As a solution to this use-case, I also added the `--no-done` flag to `omarchy-launch-floating-terminal-with-presentation` to optionally avoid `omarchy-show-done` after completion. It's current use in other scripts remains unchanged.

## Demo video
https://github.com/user-attachments/assets/36ed3ec0-1901-4f0e-a24b-f132428c2eac

### DIFF
--- a/bin/omarchy-cmd-shutdown-perform
+++ b/bin/omarchy-cmd-shutdown-perform
@@ -60,8 +60,12 @@ shutdown-query() {
   gum confirm "Force-shutdown immediately?" && shutdown-proceed
 }
 
-sleep 0.2 # Give quick-closing windows a chance
+sleep 0.1 # Give quick-closing windows a chance
 get-open-windows # For initial calculations
+if [ ${#windows[@]} -le 1 ]; then # Immediate shutdown if windows are closed
+  shutdown-proceed
+  exit 0
+fi
 
 # Prep for display, and spawn gum for user input
 tput civis


sean-imus:
This is really nice and helpful I believe, I never knew it waits an entire 2 seconds for no reason if nothing is open. Have you tested edge cases like a pinned window being in the middle of the screen thus potentially blocking the shutdown window? Otherwise I'm all for this change.

SerrpentDagger:
Thanks for the feedback. I actually wasn't aware of the popped/pinned window feature! The shutdown currently does open behind and out of focus if there's a pinned hanging window, which could lead to a 'surprise' shutdown when the timer runs out in the background.

One addition for unforeseen edge-cases is to check if the shutdown window is in focus, and cancel shutdown if not. I think that makes sense also for if someone clicks outside the TUI, at which point they probably don't want the automatic shutdown.
Potentially a better fix for opening-in-the-background specifically is to pin the shutdown TUI with a hyprland window rule. That makes it open on top of the other pinned window as one would hope.

I also tried out having pinta in a seperate workspace from where the shutdown is initiated. With the first change, that takes you to the workspace with the hanger/its popup, closing the shutdown TUI as it loses focus. I think that's a reasonable behaviour, letting the user try to deal with the hanger or else just re-initiate shutdown in that workspace and forcing it.

I've got these changes pretty well fleshed out in my testing environment now, so I should have time to update this PR tomorrow or the day after, if we think those are good solutions.

Also, a fun fact from testing: The screensaver too opens behind popped/pinned windows.

Prajwal-Prathiksh:
A very cool feature!
Amazing PR! <3

sean-imus:
> One addition for unforeseen edge-cases is to check if the shutdown window is in focus, and cancel shutdown if not. I think that makes sense also for if someone clicks outside the TUI, at which point they probably don't want the automatic shutdown. Potentially a better fix for opening-in-the-background specifically is to pin the shutdown TUI with a hyprland window rule. That makes it open on top of the other pinned window as one would hope.
> 
> I also tried out having pinta in a seperate workspace from where the shutdown is initiated. With the first change, that takes you to the workspace with the hanger/its popup, closing the shutdown TUI as it loses focus. I think that's a reasonable behaviour, letting the user try to deal with the hanger or else just re-initiate shutdown in that workspace and forcing it.

I think pinning the shutdown window is a good choice. 

Automatically switching workspaces to show windows that are blocking shutdown could be disorienting, especially if multiple hanging windows are spread across different workspaces. For such a core OS feature, we should generally avoid any steps in the shutdown process that aren’t strictly necessary.

It feels better to present the user with a live list of blocking windows and let them decide how to resolve the situation themselves. Most operating systems require an explicit user action to force shutdown when applications block, rather than relying on an automatic timeout. I do recognize that this can be annoying for some users who are accustomed to pressing the shutdown button and having the system shut down regardless.

I’m also a bit uneasy about treating the TUI losing focus as a signal to stop shutdown. There are many legitimate reasons the TUI could lose focus, for example switching workspaces to close hanging windows, in those cases the user may still fully intend to shut down.

I believe this is a really interesting discussion about UX, and I would love to hear more views on this!

dhh:
Appreciate the attempt at solving this, but I personally find the quickly appearing and then disappearing terminal window more obtrusive than the two-second wait. I also don't love the idea of using a full-size terminal for this stuff. The 2-second delay is actually just there to give Chromium a chance to save after shutdown, so we don't get an error on next boot.

I'm not opposed to something that would be try to deal with unsaved pinta, for example, but it should be using mako notifications, and it's gotta be a way lighter touch. Not 120-lines of code.

Thanks for taking a shot at this, though ✌️

SerrpentDagger:
Hmm, alright. I considered using mako originally, but it doesn't seem capable of gathering a choice from the user because it doesn't support the `--action` flag. That means a mako-based solution would be to just 'show a notification and cancel shutdown if something is hanging'. I decided against that because it blocks the user from shutting down if they can't close a window for some reason.

PR #4621, opened 2026-02-16
Add SUPER+M keybinding for maximize window

dinatih:
## Summary

- Adds `SUPER+M` as a maximize keybinding using `fullscreenstate, 1 0`
- Unlike `SUPER+F` (true fullscreen), this keeps waybar visible and apps keep their own UI (e.g. Chromium tabs/address bar remain visible)

## How it works

`fullscreenstate, 1 0` tells Hyprland to maximize the window at the compositor level (fills the screen, keeps waybar) while reporting "normal" to the client app — so browsers don't hide their toolbar.

## Test plan

- [ ] Press `SUPER+M` on a tiled window — should maximize while keeping waybar
- [ ] Press `SUPER+M` on Chromium — tabs and address bar should remain visible
- [ ] Press `SUPER+M` again — should toggle back to tiled
- [ ] Verify no conflict with existing keybindings (`SUPER+M` is currently unbound)

### DIFF
--- a/default/hypr/bindings/tiling-v2.conf
+++ b/default/hypr/bindings/tiling-v2.conf
@@ -9,6 +9,7 @@ bindd = SUPER, T, Toggle window floating/tiling, togglefloating,
 bindd = SUPER, F, Full screen, fullscreen, 0
 bindd = SUPER CTRL, F, Tiled full screen, fullscreenstate, 0 2
 bindd = SUPER ALT, F, Full width, fullscreen, 1
+bindd = SUPER, M, Maximize window, fullscreenstate, 1 0
 bindd = SUPER, O, Pop window out (float & pin), exec, omarchy-hyprland-window-pop
 
 # Move focus with SUPER + arrow keys


jheuing:
Is this the same as SUPER+ALT+F?

abenz1267:
> Is this the same as SUPER+ALT+F?

No, `SUPER+ALT+F` sets `bindd = SUPER ALT, F, Full width, fullscreen, 1`, so fullscreen to compositor and client, so the app thinks it's fullscreen and potentially hides bars etc.

This PR makes it so that it's fullscreen for the compositor, but NOT for the client.

abenz1267:
I'd rethink the binding, because right now every "fullscreen" related binding in omarchy is mapped to `F`. Having this one suddenly be on `M` isn't very intuitiv.

Maybe `Super Ctrl Alt F`?

dhh:
I don't see a difference here between `fullscreenstate, 1` (which we already have on `Super + Alt + F`) and this new `fullscreenstate, 1 0`. Tried with Ghostty, Chrome, Obsidian. All take up the full workspace, but leaves the waybar in place.

Can someone demonstrate an app where the two modes differ?

abenz1267:
The difference is what the app thinks it is:

- `bindd = SUPER ALT, F, Full width, fullscreen, 1` => maximized with gaps etc => but client thinkgs its fullscreen
- `fullscreen 1 0` => maximized with gaps etc => client doesnt think it's fullscreen

I think that should be the difference.

dhh:
There seems to be no practical difference on Hyprland with the apps I've tested. So unless someone can show an app where it's making a difference, I don't see that we need this.

dinatih:
You're right @dhh, there is no practical difference — I've confirmed it myself now.

Here's what happened: I was looking for a way to have Chrome take up the full screen (without waybar) while keeping Chrome's UI (tabs, address bar). I tried `SUPER+F` and `SUPER+CTRL+F`, neither gave me the result I wanted. I then tried to bind `SUPER+ALT+F` without realizing it was already taken by "Full width" — which is actually exactly what I was looking for all along.

Since `SUPER+ALT+F` was already bound, I ended up creating `SUPER+M` with `fullscreenstate, 1 0` as a workaround, thinking it would behave differently. In practice, as you pointed out, there's no visible difference with `fullscreen, 1`.

Part of the confusion was that I didn't realize it was Chrome/Chromium itself hiding its own UI (tabs, address bar) when it detects it's in fullscreen — I assumed it was Hyprland doing it. I also simply forgot to test `SUPER+ALT+F` during my exploration, which would have solved my problem right away.

My apologies for the noise — closing this PR.

PR #4623, opened 2026-02-16
Fix snapshot delete bug and improve script robustness

amussara:
Bug fix in omarchy-snapshot delete command + refactored scripts with validation.


dhh:
You're going to have to offer more information on a PR than that. Can't just dump a 400-line change without any context on what's being fixed or why.

PR #4626, opened 2026-02-17
Add omarchy drive select to iso2sd

Mrid22:
Based of @dhh's suggestion, i made a new pr that just integrates omarchy-drive-select into the iso2sd function


dhh:
Thanks, but the code you removed here already had that logic in there, along with protections and just suggesting sd cards 😄.

Mrid22:
> Thanks, but the code you removed here already had that logic in there, along with protections and just suggesting sd cards 😄. 

O lol nvm then

PR #4630, opened 2026-02-17
Implement lid detection for fingerprint authentication

c4software:
Hi,

Simple MR to implement the nicely done lid detection describe here https://github.com/basecamp/omarchy/issues/856#issuecomment-3518567311

This PR Add lid detection to disable fingerprint authentication when the lid is closed. Updated PAM configuration for sudo and polkit to include the lid check.

### DIFF
--- a/bin/omarchy-setup-fingerprint
+++ b/bin/omarchy-setup-fingerprint
@@ -55,19 +55,65 @@ EOF
   fi
 }
 
-add_hyprlock_fingerprint_icon() {
-  print_info "Adding fingerprint icon to hyprlock placeholder text..."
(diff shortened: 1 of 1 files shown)


c4software:
Hey @dhh, I know this PR is a bit old now, but it's still relevant, and I've been testing it ever since without any issues.

dhh:
I don't love munging these core config files to turn things off and on conditionally. I agree it would be nice to have, but either we have to find some simpler way to do it or we should make it a manual toggle, like we have with the touchpad, touchscreen etc.

dhh:
Closing in favor of the newer clamshell fingerprint-auth implementation in #6003.

c4software:
Looks like the same approach and the same kind of script (with a return code). 

No problem.

PR #4637, opened 2026-02-18
Add cd command argument to zoxide init

Metsker:
### Better fzf integration:
1. 1. type `cd folder` plus space
    2. press Tab
    3. you'll get fzf window with all zoxide tokens of the same name (right now it won't work)
2. `cdi` will work without an alias (right now you need to use `zi` or manually add an alias) 
## Cons:
1. `z` and `zi` won't be available, only `cd` and `cdi`
2. since we can't use a custom `zd` function: 
   1. jumps with zoxide won't print the whole path
   2. if you `cd` into invalid folder, you'll get `zoxide: no match found` message without `Error: Directory not found`

Overall, I think this method is cleaner and better integrated. I doubt anyone uses `z` or `zi` anyways. Error message is still clear in my opinion.

### DIFF
--- a/default/bash/aliases
+++ b/default/bash/aliases
@@ -9,19 +9,6 @@ fi
 alias ff="fzf --preview 'bat --style=numbers --color=always {}'"
 alias eff='$EDITOR $(ff)'
 
-if command -v zoxide &> /dev/null; then
-  alias cd="zd"
-  zd() {
-    if [ $# -eq 0 ]; then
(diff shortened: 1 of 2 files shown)


dhh:
Hmm. I can see some advantage to this, but I find the need for that space to be quite awkward. And I'd hate to lose this nice clue as to exactly where you're going:

image

So going to skip, but appreciate the suggestion ✌️

PR #4642, opened 2026-02-18
Fix snapshot delete to clean up created snapper snapshots

dipeshbabu:
This fixes o`marchy-snapshot` delete, which was called by `omarchy-update` on failure but was broken (undefined variable and deleting snapshot 0).

Changes:

Track snapshot IDs created during `omarchy-snapshot` create (per snapper config) in $OMARCHY_PATH/state/.

`omarchy-snapshot` delete now deletes exactly those snapshots and cleans up the state file.

Make update failure trap cleanup best-effort (omarchy-snapshot delete || true).

Testing:

Run `omarchy-snapshot` create, confirm state file is written.

Run `omarchy-snapshot` delete, confirm created snapshots are removed.

Simulate update failure path and confirm trap still prints error message cleanly.


ryanrhughes:
I actually just removed this call from `omarchy-update`. In the event of failure, we wouldn't want to delete the snapshot since that snapshot is the last known good state anyway. Thanks for pointing this issue out!

PR #4643, opened 2026-02-19
Adopt Base16 theme schema with backward-compatible rendering

OldJobobo:
## Summary

  This PR migrates Omarchy theme rendering to a Base16-first model while preserving transitional compatibility for existing ANSI-based themes. The long-term intent is to make base00-base0F the sole theme-source palette and phase out color0-color15 from colors.toml; those ANSI keys are intentionally retained for now to support a safe migration window for community themes, custom themes, and user overrides.

  ## What changed

  ### 1) Base16-first theme schema

  - Added Base16 palette keys (base00..base0F) to all stock theme colors.toml files.
  - Kept semantic keys (accent, cursor, foreground, background, selection_*) explicit.
  - Kept color0..color15 in stock themes during transition.

  ### 2) Backward-compatible renderer

  - Updated bin/omarchy-theme-set-templates to support:
      - Base16-only themes
      - ANSI-only themes
      - Mixed themes
  - Added bidirectional aliasing:
      - derive missing ANSI slots from Base16
      - derive missing Base16 slots from ANSI
  - Preserved existing template variable behavior:
      - {{ key }}
      - {{ key_strip }}
      - {{ key_rgb }}

  ### 3) Template migration to Base16 tokens

  Converted stock templates from {{ colorN }} to {{ baseXX }} where appropriate:

  - default/themed/alacritty.toml.tpl
  - default/themed/kitty.conf.tpl
  - default/themed/ghostty.conf.tpl
  - default/themed/obsidian.css.tpl
  - default/themed/btop.theme.tpl
  - default/themed/hyprland-preview-share-picker.css.tpl

  ### 4) Documentation + validation

  - Added default/themed/COLORS_SCHEMA.md with canonical schema/mapping and transition policy.
  - Updated config/omarchy/themed/alacritty.toml.tpl.sample to Base16-first docs and marked ANSI
    aliases as deprecated compatibility.
  - Added bin/omarchy-dev-validate-themes for repeatable checks:
      - required keys in themes/*/colors.toml
      - render pass via template generator
      - unresolved placeholder detection
      - expected generated artifact presence

  ## Validation performed

  - bin/omarchy-dev-validate-themes passes for all stock themes.
  - Live theme switch smoke-tested on dark + light themes.
  - Confirmed no unresolved placeholders in generated theme artifacts after apply.

  ## Compatibility / rollout notes

  - This is Base16-first, not Base16-only.
  - Existing ANSI-only custom themes remain supported during transition.
  - color0..color15 should be considered transitional compatibility aliases for now.
  - Future cleanup (removing ANSI compatibility) can happen in a later release window once ecosystem
    migration is stable.

  ## Non-goals in this PR

  - No removal of ANSI compatibility.
  - No forced migration of user custom themes outside stock theme files included in this repo.


dhh:
What is the advantage or outcome of this?

OldJobobo:
Good question, The main benefit is stronger ecosystem alignment and better theming fidelity.  This brings Omarchy in line with the Base16 conventions commonly used across Linux theming workflows, which makes themes easier to implement and more predictable across components. It also reduces redundant ANSI mapping and introduces clearer UI-layer color roles, so both TUI and GUI apps render more consistently without requiring theme authors to define additional base colors.

OldJobobo:
As for the technical outcome: colors.toml still ends up in a familiar shape, but organized around Base16 with 8 UI ramp colors and 8 semantic colors. ANSI values are then derived from the Base16 palette using the standard Base16-to-ANSI mapping and applied only in terminal configs, since those are the only places that require ANSI-formatted colors.

In this PR, I’m rolling that in as a gradual transition: current behavior stays intact, while Base16-native definitions and mappings are introduced alongside it so nothing breaks for existing themes. Once that transition period is done, ANSI stops being something we hand-maintain and becomes a generated terminal-facing output from the Base16 palette.

That's the idea, but totally open to better or more direct methods.  I just wanted to be conservative and try not to break anything I did not think of.

dhh:
I think we need to see some practical benefits in specific apps before it makes sense to change. Every time we do, we churn existing themes that now need to be upgraded to run against the latest. We can do that if there's a big pay off, but I want to see exactly what that is. I really appreciate you looking into this! @bjarneo looked into using base24, and it's kinda the same discussion: What are we getting? In that case, it seemed mostly that we'd get better btop themes, but at the cost of a bigger palette that would need defining. That also didn't seem worth it.

bjarneo:
I agree, we have to see the value.

A comment added regarding the base24 implementation: https://github.com/basecamp/omarchy/pull/4541#issuecomment-3926230446

OldJobobo:
Given the conversation here and in #4541, @bjarneo's base24 implementation is a better route.

PR #4644, opened 2026-02-19
Fix reinstall-configs to apply theme defaults

dipeshbabu:
`omarchy-reinstall-configs` used command substitution to run the theme script, which captures stdout and attempts to execute it. Since the theme script produces no command output, theme setup was silently skipped.

This PR runs the theme script normally so defaults are applied during reinstall.

Testing:

Run `omarchy-reinstall-configs` and confirm theme setup runs (no silent skip).

### DIFF
--- a/bin/omarchy-reinstall-configs
+++ b/bin/omarchy-reinstall-configs
@@ -13,7 +13,7 @@ cp -R ~/.local/share/omarchy/config/* ~/.config/
 cp ~/.local/share/omarchy/default/bashrc ~/.bashrc
 echo '[[ -f ~/.bashrc ]] && . ~/.bashrc' | tee ~/.bash_profile >/dev/null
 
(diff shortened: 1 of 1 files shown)


dhh:
This PR is full of unrelated stuff. I also can't replicate the problem it describes. `omarchy-reinstall-configs` correctly restores the theme in my testing.

dipeshbabu:
> This PR is full of unrelated stuff. I also can't replicate the problem it describes. `omarchy-reinstall-configs` correctly restores the theme in my testing.

You’re right that theme restore works today `$(bash …)` does execute the script. The issue I meant to address is that command substitution captures stdout and then tries to execute it, so any future output from `theme.sh` could break reinstall. I’ll resubmit a minimal PR with only this one-line change and no unrelated edits.

PR #4650, opened 2026-02-19
Use web app favicon if none given in omarchy-webapp-install

pomartel:
This uses a Google API to fetch a website favicon if none was provided. At first I tried to implement the logic in a bash script but there are so many edge cases and security issues to consider.

The pros :
- Very robust, handles all edge cases. Whatever URL you throw at it, it will return an icon or use a default one if none was found.
- Always returns a PNG file so no ICO conversion needed.
- It's dead simple!

The cons:
- Could this be a privacy issue?
- Although it's widely used, it does not seem to be official or documented. There's no guarantee this will always work. But if it does stop to work, the consequences are minimal.


dhh:
So good, so simple!

PR #4656, opened 2026-02-19
Fix OpenCode control sequence leaking in the editor pane with the nic function

pomartel:
When using the `nic` command in tmux, an OpenCode control sequence leaks in the editor pane (see video).

This PR adds a 2 seconds pause before switching the focus to the editor pane to give sufficient time for OpenCode to initialize.

Two other ways to fix this :

- Keep the focus in the AI pane
- Disable tmux passthrough for the AI pane. This also works but I don't know if it could have unwanted side effects.

I scoped the fix to OpenCode only. I don't have a Claude Code subscription to test if it's also causing the issue.

https://github.com/user-attachments/assets/d2bae8ff-3a16-4775-9592-a643c0c5d18b

### DIFF
--- a/default/bash/functions
+++ b/default/bash/functions
@@ -151,6 +151,11 @@ tml() {
   # Run nvim in the left pane
   tmux send-keys -t "$editor_pane" "$EDITOR ." C-m
 
+  # Short pause to avoid OpenCode terminal control sequence from leaking into the editor pane
+  if [[ "$ai" == "c" ]]; then
+    sleep 2
+  fi
+
   # Select the nvim pane for focus
   tmux select-pane -t "$editor_pane"
 }
--- a/default/bash/functions
+++ b/default/bash/functions
@@ -151,11 +151,6 @@ tml() {
   # Run nvim in the left pane
   tmux send-keys -t "$editor_pane" "$EDITOR ." C-m
 
-  # Short pause to avoid OpenCode terminal control sequence from leaking into the editor pane
-  if [[ "$ai" == "c" ]]; then
-    sleep 2
-  fi
-
   # Select the nvim pane for focus
   tmux select-pane -t "$editor_pane"
 }
--- a/default/bash/functions
+++ b/default/bash/functions
@@ -151,6 +151,11 @@ tml() {
   # Run nvim in the left pane
   tmux send-keys -t "$editor_pane" "$EDITOR ." C-m
 
+  # Temporarily disable passthrough for the ai pane to prevent
+  # OpenCode control sequences from leaking into the editor pane
+  tmux set-option -pt "$ai_pane" allow-passthrough off
+  sleep 3 && tmux set-option -pt "$ai_pane" allow-passthrough on &
+
   # Select the nvim pane for focus
   tmux select-pane -t "$editor_pane"
 }


dhh:
Maybe @thdxr can help guide us to the best solution here.

lucifer:
Tested on 3.4.0.rc1 on <your laptop> from the fix-nic-function branch.

In tmux, ran nic twice in the same pane, letting nvim/OpenCode fully start and then exiting.

Previously I saw the =31337;OK control sequence blob on the second exit; now both runs exit cleanly with no leaked control sequence. Looks good to me.

lucifer:
Just for info, I have not seen the control codes =31337;OK since the 2 second pause was introduced and I'm loving the new nic layout.

pomartel:
Yeah I'm not quite happy with this as it slows down the whole process and 2 seconds is just an arbitrary number. Maybe some systems would require a 3 seconds pause. I don't know why DHH is not experiencing this issue. My guess is that he runs faster hardware that doesn't trigger this race condition.

I removed the pause and instead disable tmux passthrough for the ai pane just before switching the focus to the editor pane. Then I re-enable passthrough asynchronously after 3 seconds. This still feels like a bit of a hack but at least it doesn't slow down the whole thing!

@lucifer Do you mind testing this new fix?

dhh:
Just checking with the opencode folks to see if we can't come up with something nicer here than the bruteness of a 2s delay.

dhh:
I'm not seeing this issue, but I'm seeing something else:

image

Random nvim errors like this.

pomartel:
@dhh Just to be clear, my latest commit (20 minutes ago) removes the 2 seconds delay and instead relies on disabling and re-enabling passthrough.

lucifer:
Here's a screen recording from what I get. Didnt mention the Lua stuff plainly because I figured it might just be local environment stuff on my end. Will test again on a fresh build.

https://github.com/user-attachments/assets/a7230770-df40-4963-ae00-88eae273dc18

pomartel:
@lucifer I think this might be a separate issue. Try to kill the tmux session with **CTRL+ALT+SHIFT+END** and see if issuing the `nic` command again produce the nvim error. Is that an error that happens in the Release Candidate too or just with my PR? With the latest fix, there's no leaked control sequences?

kommander:
@pomartel Hi, opentui author here, the tui lib behind opencode. 

The sequence that is "spilled" looks like the response to the _kitty graphics query_. These responses are unfortunately very unreliable in response time. I do not have any understanding of the tmux codebase, but from what I gathered it looks like the the kitty graphics sequence/query (and probably other APC/DCS wrapped sequences) seems not to be scoped to the focused pane, so any active pane will receive the response. 

We had issues with the kitty graphics query before, so there actually is an env flag to skip it, try prepending opencode with `OPENTUI_NO_GRAPHICS=1`.

Edit: Opencode doesn't use kitty graphics currently at all, so it doesn't make a difference. This issue may come up again when kitty graphics may be used, then we need to find a better solution, like allowing to tell if graphics are supported manually, then it doesn't have to send the query.

pomartel:
@kommander Thanks for your help! Unfortunately I cannot get it to work. I made the following video where I launch opencode in the right tmux pane with the `OPENTUI_NO_GRAPHICS=1` and then quickly switch focus to the neovim pane and you can see that the code still leaks.

https://github.com/user-attachments/assets/5e557276-4a15-4af5-afca-c21157a6a1f1

Not sure what else to try at this point. Do you think my current proposed fix where I disable tmux passthrough for a few seconds and then re-enable it could cause any unwanted side effects?

kommander:
That is very odd. The id `31337` is definitely the hardcoded kitty graphics query id, which I just confirmed is not emitted at all with `OPENTUI_NO_GRAPHICS=1`. There is no other place where that is emitted from opencode. That flag would be the only way to prevent that from being emitted currently. That seems like the process does not see that flag for some reason. Can you start opencode and use `ctrl+p` and then `toggle console` and see if there are warnings? If the env cannot be read in native for some reason it will say so.

Screenshot 2026-02-21 at 00 42 56

I can make kitty graphics opt-in on application level so opencode does guarantee to not emit it.

lucifer:
@pomartel Tested again with the latest commits on fix-nic-function.
In tmux, I launch OPENTUI_NO_GRAPHICS=1 opencode in the AI pane, then switch focus back to the neovim pane. I still see the control sequence with =31337;OK leak into the editor pane.

Here is my exact sequence including the `toggle console` @kommander mentioned: 
https://asciinema.org/a/5I1TlqJc7qlJxhd4

pomartel:
@lucifer Thanks for confirming this!

@kommander There's no warning in the console unfortunately.

simonklee:
I can reproduce this with a minimal setup that just sends the kitty graphics probe and waits for the response.

The probe (`i=31337`) is sent on startup, and if the terminal/tmux focus changes before the response arrives, the delayed response is delivered to whichever pane/app is active at that moment.

In `nic`, that ends up in the editor pane, which is why `=31337;OK` leaks into vim.

From my current understanding, it's something we can try to mitigate on the OpenTUI side by changing when/how we emit the probe, but we can't fully control where a delayed terminal/tmux reply is routed once the probe was already emitted.

simonklee:
I've looked at this more today and found that our OPENTUI_NO_GRAPHICS=1 passing to native code is broken. The upstream tracking issue for it is at https://github.com/anomalyco/opentui/issues/729

We're working on a fix for it. Thanks for raising the issue!

pomartel:
Great! I am going to close this PR and track the issue upstream. If a fix makes its way to OpenCode, i will open a new PR.

I also could not reproduce the issue with Alacritty, which is now the default terminal emulator for Omarchy.

simonklee:
Hey, Pierre - sorry for the late follow up. 

OpenCode version `1.2.19` has a fix for this, possibly earlier versions, but I did not check. We changed the env flag to `OPENTUI_GRAPHICS` which can be set to `0` or `1`.

Thanks again for the issue report.

PR #4659, opened 2026-02-19
Create chromium-flags.conf when enabling Google account login

dipeshbabu:
This script currently only runs when `~/.config/chromium-flags.conf` already exists, making it a no-op on fresh installs.

This change creates the parent directory (if needed) and touches the file before appending the OAuth flags. The operation remains idempotent (no duplicate lines on re-run).

Testing:

- `rm -f ~/.config/chromium-flags.conf`
- run `omarchy-install-chromium-google-account`
- confirm file is created and contains both `--oauth2-*` lines
- re-run and confirm no duplicates


dhh:
chromium-flags.conf exists on new installs. It's copied in by default. Don't think we need this.

PR #4662, opened 2026-02-19
Fail with usage on unknown install-dev-env and remove-dev-env target

dipeshbabu:
If an unknown argument is provided, the script currently does nothing and exits successfully.
This adds a default case to print an error + usage and exit 1. No changes for valid options.

### DIFF
--- a/bin/omarchy-install-dev-env
+++ b/bin/omarchy-install-dev-env
@@ -3,7 +3,7 @@
 # Install one of the supported development environments. Usually called via Install > Development > * in the Omarchy Menu.
 
 usage() {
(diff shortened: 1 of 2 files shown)


pomartel:
You should probably put usage in a function to keep it DRY.

dipeshbabu:
> You should probably put usage in a function to keep it DRY.

Good call! updated to DRY the usage string via a small `usage()` helper and reused it for both the empty-arg path and unknown target path.

Kasui92:
You can add a similar solution for `omarchy-remove-dev-env` as well.

dipeshbabu:
> You can add a similar solution for `omarchy-remove-dev-env` as well.

Good catch! Thank you!

Kasui92 (approved):

dhh:
Appreciate the idea, but I don't think this is necessary. We have the usage doc already and we have a fixed list when you're going through the main install menu (which is what most will do). Thanks for looking at it, though.

PR #4663, opened 2026-02-19
Fail early when mise is missing in install-dev-env

dipeshbabu:
`omarchy-install-dev-env` relies on `mise` for most install targets, but currently fails later with `mise: command not found` when it isn’t installed.

This adds a small preflight check to print a clear error and exit early. No behavior change when `mise` is present.

Testing:

- Run `omarchy-install-dev-env deno` without `mise` installed → shows clear error and exits 1.
- With `mise` installed → behavior unchanged.

### DIFF
--- a/bin/omarchy-install-dev-env
+++ b/bin/omarchy-install-dev-env
@@ -7,6 +7,11 @@ if [[ -z "$1" ]]; then
   exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


dhh:
Don't think we need this. Mise is part of the standard distro install and the existing error is fine.

PR #4668, opened 2026-02-20
Support single-key keyboard backlight cycling (like Asus F4)

RyanLatimer:
## Summary
- Adds `cycle` argument support to `omarchy-brightness-keyboard` for hardware that uses a single key to loop through backlight levels (like Asus laptops pressing F4).
- Adds a dedicated `omarchy-swayosd-kbd-brightness` wrapper so the SwayOSD pop-up uses a `keyboard-brightness` icon instead of the generic screen brightness icon.
- Adds `XF86KbdLightOnOff` to default `media.conf` so these keys work out-of-the-box.

### DIFF
--- a/bin/omarchy-brightness-keyboard
+++ b/bin/omarchy-brightness-keyboard
@@ -1,0 +1,42 @@
+#!/bin/bash
+
+# Adjust keyboard backlight brightness using available steps.
+# Usage: omarchy-brightness-keyboard <up|down|cycle|toggle>
+
+direction="${1:-up}"
+
+# Find keyboard backlight device (look for *kbd_backlight* pattern in leds class).
+device=""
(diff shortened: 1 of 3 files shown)


RyanLatimer:
This has been tested on an Asus ProArt P16.

dhh:
We have this on dev already! Have a look if there's anything there that doesn't work like you were intending. But I'm using what we have on dev with Asus G14.

RyanLatimer:
@dhh I was on dev, and it didn't work.

dhh:
Hmm, this is against master. And it's adding a new omarchy-keyboard-brightness. Can you try opening a PR against dev so it's easier to see what changed?

RyanLatimer:
@dhh this has been moved to #4672.

PR #4669, opened 2026-02-20
feat: add SUPER A shortcut for near universal select all

RyanLatimer:
## Summary
* Added `SUPER A` as a universal "Select All" shortcut to complement the existing `SUPER C/V/X` copy/paste shortcuts.
* Uses Hyprland's `sendshortcut` to dispatch `Ctrl+A`.
* Added to `default/hypr/bindings/clipboard.conf`.

### DIFF
--- a/bin/omarchy-select-all
+++ b/bin/omarchy-select-all
@@ -1,12 +1,13 @@
 #!/usr/bin/env bash
 
 # Universal select all: detects if the active window is a terminal and sends the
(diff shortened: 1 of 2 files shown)


dhh:
This isn't universal because in the terminal Ctrl+A goes to the start of the line.

RyanLatimer:
The issue has been fixed in Kitty and Ghostty, but there isn't a select all function in Alacritty. Should it just fall through to be CTRL + A in Alactritty or just do nothing?

PR #4674, opened 2026-02-20
Add keyboard layout selection to installer and post-install menu

janhesters:
## Summary

- Add interactive keyboard layout selector during first-time installation (preflight phase) using `gum filter` over `localectl list-keymaps`
- Add KEYMAP-to-XKB fallback in `detect-keyboard-layout.sh` for chroot installs where `localectl` can't auto-map to XKBLAYOUT/XKBVARIANT
- Add `omarchy-keyboard-select` post-install utility accessible via Update → Keyboard menu, applying changes live via `hyprctl` and persisting to `input.conf`

Ref: https://github.com/basecamp/omarchy/discussions/4435

## Test plan

- [ ] On fresh Arch+Omarchy install, keyboard selector appears after guards pass
- [ ] Selecting "dvorak" sets `KEYMAP=dvorak` in `/etc/vconsole.conf`
- [ ] After install, `~/.config/hypr/input.conf` contains `kb_layout = us` and `kb_variant = dvorak`
- [ ] On re-install, the selector does not appear (marker file exists)
- [ ] `omarchy-keyboard-select` works post-install via Super+Alt+Space → Update → Keyboard
- [ ] After reboot, disk decryption prompt uses the selected layout


seppegadeyne:
Possible related data point from a Belgian/AZERTY setup.

After an Omarchy update I hit the LUKS/Plymouth unlock prompt rejecting my passphrase with:

```text
No key available with this passphrase
```

The interesting part: on a Belgian keyboard, the passphrase only worked when I typed the top-row digits without Shift. With the normal Belgian/AZERTY behavior I expected Shift+digits for numbers, but the unlock prompt behaved as if it was using a default/US-style mapping.

I checked the installed system afterwards:

- Omarchy: 3.8.2
- Kernel: 7.0.10-arch1-1
- `localectl`:
  - `VC Keymap: be-latin1`
  - `X11 Layout: be`
- `/etc/vconsole.conf`:
  - `KEYMAP=be-latin1`
  - `XKBLAYOUT=be`
- Hyprland active keyboard layout: Belgian
- The current UKI at `/boot/EFI/Linux/omarchy_linux.efi` does contain `keymap.bin`
- Extracted `keymap.bin` matches `loadkeys -q -u be-latin1 -b`, not `us`

So the selected keymap is present and built into the UKI.

However, the current Omarchy hook override on my system is:

```sh
HOOKS=(base udev plymouth keyboard autodetect microcode modconf kms keymap consolefont block encrypt filesystems fsck btrfs-overlayfs)
```

This is also what the current Omarchy source writes in `install/login/limine-snapper.sh`.

Because `plymouth` is before `keymap`, my hypothesis is that the graphical Plymouth password prompt can start with the default keymap before `keymap` loads `be-latin1`. That would explain why the initramfs contains the right keymap, but the LUKS prompt still behaves like a different layout.

I have not applied a local fix yet, but this seems worth considering for the PR/test plan. For encrypted installs, it may not be enough to set `/etc/vconsole.conf`; the hook order may also need to ensure `keyboard keymap consolefont` run before `plymouth`/`encrypt`, or at least that the Plymouth password prompt uses the selected keymap.

Potentially relevant order to test:

```sh
HOOKS=(base udev keyboard autodetect microcode modconf kms keymap consolefont block plymouth encrypt filesystems fsck btrfs-overlayfs)
```

Not claiming this is the final correct order, just sharing a reproducible data point from a BE/AZERTY LUKS unlock mismatch.

dhh:

PR #4676, opened 2026-02-21
Comment out mouse acceleration setting

CryptoCain:
Comment out mouse acceleration setting in input.conf


CryptoCain:
Gives mouse movements a 1:1 ratio.

dhh:
I'm happy with the acceleration the way it is right now. We've exposed this for easy changing for those who want something different, though.

CryptoCain:
its really for the gamer's. you do have a gaming section so i thought it was needed.. it really help me out in-game .. thanks again

dhh:
Ahh, I see. That's fair.

PR #4710, opened 2026-02-21
Adds Omarchy Skill to OpenAI Codex CLI

coolbotic:
The current SKILL.md is placed in the `.claude/` folder, which is fine for using with opencode; the codex CLI has its own directory and does not read skills from there.

This change symlinks the SKILL.md file to `.codex` folder and adds a migration for current systems.


coolbotic:
Solved in commit 9f187b55da990fac72d3a47b20deb157a0e6e082

PR #4711, opened 2026-02-21
Add theme white

bjarneo:
A suggestion is to rename Vantablack to either black or use the opportunity to call these themes yin and yang.

preview


dhh:
I like the idea of yin and yang. But then they also need each a matching/inverted background image with that on it 😄

bjarneo:
For sure. I'll come back to this 😀

PR #4715, opened 2026-02-21
UPDATE: add delete .git confirmation

dhms013:
Add [option](https://github.com/dhms013/omarchy/blob/fd8ee3ce9c30fe12fc75d4299a029722d7284834/bin/omarchy-theme-install#L35) for user wether they want to keep the .git repo or delete it.

I manage my themes in my dotfiles backup (stow), and I usually forgot to delete .git directory each time I want to push my dotfiles into git. this method really save time and effort for me, because I don't need to always remember about deleting new theme's .git before push my dotfiles into my github~

https://github.com/user-attachments/assets/ffd3b10e-99f0-43d7-b1d8-770e22280d81


dhms013:
1. Replace previous [omarchy-theme-install](https://github.com/dhms013/omarchy/blob/704f88a968d7334e33bf1d3eb9db24c22d3f293d/bin/omarchy-theme-install#L22) workflow to add option rename .git folder
2. Make sure the installed theme still [updateable](https://github.com/dhms013/omarchy/blob/704f88a968d7334e33bf1d3eb9db24c22d3f293d/bin/omarchy-theme-update#L22)

Goal :
- To make people who want to backup their dotfiles (like me) life easier

https://github.com/user-attachments/assets/4f997132-e030-4684-8ae1-aeae01612047

dhh:
Appreciate the suggestion, but don't think this is worth it and would prevent updates.

PR #4719, opened 2026-02-21
Fix mkinitcpio hook persistence and vfat initramfs coverage after 3.4 upgrade failures

aadarwal:
I hit this while switching channels and upgrading to Omarchy 3.4: after the upgrade, mkinitcpio pacman hooks were left in `*.hook.disabled`, and boot also failed mounting `/boot` with `unknown filesystem type 'vfat'`.

After tracing logs and the install/update flow, I found this can happen if install exits after hooks are disabled but before they are restored. In that state, later kernel upgrades can skip initramfs regeneration. On EFI systems, that gets worse because `vfat` for `/boot` is not guaranteed to be included in initramfs when relying on autodetect.

This change hardens that entire path instead of patching a single symptom. It restores disabled mkinitcpio hooks on installer exit/error paths, restores them again before system package upgrades, and backfills existing systems with a migration that adds `MODULES+=(vfat)`, repairs disabled hooks, rebuilds boot artifacts, and marks reboot-required when changes are applied. For fresh installs, `limine-snapper.sh` now always writes the vfat drop-in, and it now runs with `set -e` so failures stop immediately in its own shell context. There is also a post-update warning if mkinitcpio hooks are still disabled, since that should now be exceptional.

The goal is to close the exact failure class I saw during the 3.4 transition and cover fresh installs, upgrades, and already-affected systems.

### DIFF
--- a/bin/omarchy-update-analyze-logs
+++ b/bin/omarchy-update-analyze-logs
@@ -9,3 +9,12 @@ if grep -q "Updating linux initcpios" "$update_log"; then
     echo
   fi
 fi
+
+# Check for disabled mkinitcpio hooks (should never happen after an update)
+hooks_dir="/usr/share/libalpm/hooks"
+if [[ -f "$hooks_dir/90-mkinitcpio-install.hook.disabled" ]] || \
+   [[ -f "$hooks_dir/60-mkinitcpio-remove.hook.disabled" ]]; then
+  echo -e '\e[31mWarning: mkinitcpio pacman hooks are disabled. Kernel upgrades will not regenerate initramfs.\e[0m'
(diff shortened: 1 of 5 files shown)


aadarwal:
Validation run locally:

- `bash -n bin/omarchy-update-perform`
- `bash -n install/helpers/errors.sh`
- `bash -n install/login/limine-snapper.sh`
- `bash -n migrations/1771711738.sh`
- `bash -n bin/omarchy-update-analyze-logs`

All pass.

Also validated against the failure pattern seen during the 3.4 upgrade: disabled mkinitcpio hooks persisted, and boot logs showed `unknown filesystem type 'vfat'` with stale-module symptoms. This change addresses that at installer exit, pre-upgrade, and migration backfill layers.

dhh:

PR #4724, opened 2026-02-23
Fix custom theme overrides by centralizing colors and moving imports

SeanAnd:
TLDR: Fixed custom themes sometimes being unable to override default styling due to import order.

Previously, each component (waybar, swayosd, walker) defined its own color variables in separate .tpl files and imported them at the top of config files, making it difficult for custom themes to override defaults. Centralized all color definitions into a single colors.css.tpl and standardized the import order across config files: colors.css at the top, default styling in the middle, theme overrides at the bottom. This lets custom themes loaded at the bottom cleanly override any default styling above them.

Remapped component specific variable names to the standard ones from colors.css (e.g. base to background, text to foreground) and emptied the now redundant waybar.css.tpl, swayosd.css.tpl, and walker.css.tpl.

### DIFF
--- a/config/swayosd/style.css
+++ b/config/swayosd/style.css
@@ -1,22 +1,22 @@
-@import "../omarchy/current/theme/swayosd.css";
+@import "../omarchy/current/theme/colors.css";
 
 window {
   border-radius: 0;
   opacity: 0.97;
-  border: 2px solid @border-color;
+  border: 2px solid @foreground;
 
-  background-color: @background-color;
+  background-color: @background;
 }
 
 label {
   font-family: 'JetBrainsMono Nerd Font';
   font-size: 11pt;
 
-  color: @label;
+  color: @foreground;
 }
 
 image {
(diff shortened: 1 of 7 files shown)


OldJobobo:
Why the remapping?  I suspect DHH would prefer names to be unified.  Is that a pywal naming convention or something?

I support the idea of an imported colors.css to make universal gtk-based layer shell client colors, and have all the theme colors available there for use.

SeanAnd:
> Why the remapping? I suspect DHH would prefer names to be unified. Is that a pywal naming convention or something?
> 
> I support the idea of an imported colors.css to make universal gtk-based layer shell client colors, and have all the theme colors available there for use.

Thanks! That was the goal. Previously each component defined its own subset of color variables with inconsistent names. Now colors.css exposes everything from colors.toml in one place, so any GTK layer shell client can reference the full palette without needing its own mappings.

dhh:

PR #4730, opened 2026-02-23
add: tmux "subfolder" for keybindings in learn keybindings

pukljak:
I think it would be good to be able to access tmux keybindings little bit faster then searching for them in configs.

This solution is my idea of solving that "problem".


Bucurenciu-Cristian:
Great idea.

dhh:

PR #4736, opened 2026-02-24
Fix(Hibernate): Make resume config precedence-safe

xipeng-jin:
## Summary
This PR fixes hibernation resume reliability in Omarchy on Limine systems and adds auto-repair for already broken installs.

## What was the issue?
Users could enable hibernation and create swap successfully, but after hibernating and powering back on, resume failed (black screen / non-restored session). In affected systems, boot `cmdline` did not contain valid `resume=` and `resume_offset=` parameters, or `resume_offset` was empty.

## Why it was happening (root cause)
Omarchy wrote resume parameters to `/etc/limine-entry-tool.d/resume.conf`, but `/etc/default/limine` has higher precedence for kernel `cmdline` and could override drop-ins. As a result, Limine generated boot entries without effective resume parameters. Additionally, `resume_offset` was computed without `sudo`, which could fail silently and produce an empty `resume_offset=` value.

## How we fixed it
- Updated `omarchy-hibernation-setup` to:
  - compute `resume_offset` with `sudo`,
  - validate resume device/offset strictly,
  - write effective resume args into `/etc/default/limine`,
  - remove legacy conflicting drop-ins.
- Updated `omarchy-hibernation-remove` to:
  - remove resume-related kernel args from `/etc/default/limine`,
  - remove legacy drop-ins,
  - regenerate Limine entries with `limine-update`.
- Updated `omarchy-hibernation-available` to require runtime `resume=` and numeric `resume_offset=`.
- Added migration `1771849001.sh` to auto-repair existing broken installs.

### DIFF
--- a/bin/omarchy-hibernation-available
+++ b/bin/omarchy-hibernation-available
@@ -10,8 +10,20 @@ SWAPSIZE_KB=$(awk '!/Filename|zram/ {sum += $3} END {print sum+0}' /proc/swaps)
 SWAPSIZE=$(( 1024 * ${SWAPSIZE_KB:-0} ))
 
 HIBERNATION_IMAGE_SIZE=$(cat /sys/power/image_size)
+RESUME_DEVICE=""
+RESUME_OFFSET=""
 
-if (( SWAPSIZE > HIBERNATION_IMAGE_SIZE )) && [[ -f /etc/mkinitcpio.conf.d/omarchy_resume.conf ]]; then
+for token in $(cat /proc/cmdline); do
+  if [[ $token == resume=* ]]; then
+    RESUME_DEVICE=${token#resume=}
+  fi
+
+  if [[ $token == resume_offset=* ]]; then
+    RESUME_OFFSET=${token#resume_offset=}
+  fi
+done
+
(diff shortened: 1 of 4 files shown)


dhh:
I've not seen this issue on any of the installs or post-configurations that I've done so far. I wonder what's triggering this issue in your particular case. Did you start from the official ISO, the 3.4 RC, or a manual install?

NicolasDorier:
> Omarchy wrote resume parameters to /etc/limine-entry-tool.d/resume.conf, but /etc/default/limine has higher precedence for kernel cmdline and could override drop-ins. 

This part is fixed by https://github.com/basecamp/omarchy/pull/5114

> Additionally, resume_offset was computed without sudo, which could fail silently and produce an empty resume_offset= value.

It is indeed possible that if people updated Omarchy when there was this bug, the `resume_offset=` would be empty. (https://github.com/basecamp/omarchy/issues/5265)

Given that this bug only happened for a very limited number of people, with a documented workaround in https://github.com/basecamp/omarchy/issues/5265, I don't think it is worth changing so much lines of code, which has AI slop potentially breaking something else in its attempt to salvage a few.

Especially it's trying to add the offset to `/etc/default/limine`, which is useless since it is in the drop-in.

PR #4743, opened 2026-02-26
Fail on unknown provider in omarchy-setup-dns

dipeshbabu:
Add default case so invalid input doesn’t restart networking services.


Mrid22:
given that there is a custom option, the only way to detect an unknown provider would be to ping the address, and if it doesnt exist, then we know that server to be invalid

PR #4749, opened 2026-02-26
Add --isolate option for webapps to run in separate browser process

### LINKED ISSUE #4748: Webapp crash takes down all browser windows — add process isolation option
## Problem

Webapps installed via `omarchy-webapp-install` launch with `brave --app=URL`, which connects to the existing main browser process. If a webapp crashes or the main process becomes unstable, **all browser windows die simultaneously** — including unrelated tabs and other webapps.

### Crash report

- **Brave:** 145.1.87.190
- **OS:** Arch Linux, kernel 6.18.x, Hyprland/Wayland
- **GPU:** Intel Arc A750 (discrete), AMD Raphael/Radeon 680M (integrated)
- **RAM:** 32GB, 4GB zram swap

**What happened:**
1. Main Brave instance had been running ~4 days
2. Launched `brave --app=https://grok.com` at 08:33
3. Launched `brave --app=https://outlook.office.com/mail/` at 08:36
4. At 08:36:29, the main Brave scope exited — all windows died

**Resource usage at death:**
- Main Brave scope: 13h 56min CPU, **6GB peak memory**, 116MB swap
- No OOM kill in kernel logs, no segfault in dmesg, no crash dump generated

## Proposed solution

Add an `--isolate` option to `omarchy-launch-webapp` that launches webapps with `--user-data-dir=`, creating a fully separate browser process tree. If the isolated webapp crashes, the main browser is unaffected.

This is browser-agnostic — `--user-data-dir` is supported by all Chromium-based browsers that `omarchy-launch-webapp` supports (Chrome, Brave, Edge, Opera, Vivaldi, Helium, and the Chromium fallback). The isolated data directory is stored under `~/.local/share/omarchy/webapps/<name>` rather than any browser-specific config path.

**Trade-off:** Isolated webapps don't share cookies/sessions with the main browser, so users need to log in separately.

### Usage

**Interactive:** `omarchy-webapp-install` prompts with a `gum confirm` asking whether to isolate.

**Programmatic:** `omarchy-webapp-install --isolate "Grok" https://grok.com/ Grok.png`

**PR:** #4749

lukeabbott:
## Summary

- Adds `--isolate` flag to `omarchy-launch-webapp` that launches webapps with `--user-data-dir=`, creating a fully separate browser process tree
- Adds interactive `gum confirm` prompt and `--isolate` flag to `omarchy-webapp-install`
- Cleans up isolated data directories in `omarchy-webapp-remove` and `omarchy-webapp-remove-all`

Browser-agnostic — works with all Chromium-based browsers supported by `omarchy-launch-webapp` (Chrome, Brave, Edge, Opera, Vivaldi, Helium). Isolated data stored at `~/.local/share/omarchy/webapps/<name>`.

Retargeted from `dev` to `quattro`. The original issue was converted to a suggestion: https://github.com/basecamp/omarchy/discussions/6250

## Test plan

- [x] Programmatic install with `--isolate` creates correct `.desktop` Exec line
- [x] Isolated webapp launches with separate `--user-data-dir` and its own process tree
- [x] Killing isolated webapp does not affect main browser
- [x] `omarchy-webapp-remove` and `omarchy-webapp-remove-all` clean up isolated data directories
- [x] Non-isolated webapps still work as before (no regression)
- [x] `test/shell.d/webapp-isolation-test.sh` and `./test/cli` pass

### DIFF
--- a/bin/omarchy-launch-webapp
+++ b/bin/omarchy-launch-webapp
@@ -1,7 +1,35 @@
 #!/bin/bash
 
 # omarchy:summary=Launch a URL as a web app in the default supported browser
-# omarchy:args=<url>
+# omarchy:args=<url> [--isolate|--isolate=<name>] [browser-flags...]
+# omarchy:examples=omarchy launch-webapp https://example.com | omarchy launch-webapp https://mail.example.com --isolate=mail
+
+# --isolate / --isolate=<name> starts a separate browser process via
+# --user-data-dir, so a crash in the webapp cannot take down the main browser.
+
+URL="$1"
+shift
+
+ISOLATE=false
+ISOLATE_NAME=""
+EXTRA_ARGS=()
+while (($# > 0)); do
+  case "$1" in
+  --isolate)
+    ISOLATE=true
+    shift
+    ;;
+  --isolate=*)
+    ISOLATE=true
+    ISOLATE_NAME="${1#--isolate=}"
+    shift
+    ;;
+  *)
+    EXTRA_ARGS+=("$1")
+    shift
+    ;;
+  esac
+done
 
 browser=$(xdg-settings get default-web-browser)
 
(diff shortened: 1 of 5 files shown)


lukeabbott:
Rebased onto `quattro` and retargeted the PR base from `dev`.

The webapp install/remove scripts were rewritten after the original patch, so `--isolate` is reapplied there (including `omarchy-webapp-remove-all` profile cleanup) plus a shell test for install/launch/remove.

rodrigobalero:
Thanks for working on this. I tested commit `cb6b290f` on Omarchy 4.0.1-1 with Google Chrome 152 and two Chrome profiles.

A useful benefit of this PR, beyond crash isolation, is deterministic session storage for authentication-heavy web apps. The current WhatsApp launcher does not specify a Chrome profile, so Chrome may select whichever profile is active or was used last.

On my system, WhatsApp had written IndexedDB data into both `Default` and `Profile 1`, but only `Default` contained the persistent authentication cookie. Reopening the web app under the other profile therefore appeared to lose the session. Pinning `--profile-directory=Default` fixed the local problem: two consecutive close/reopen cycles launched as `chrome-web.whatsapp.com__-Default`, and only the `Default` storage was updated.

I also verified the isolation approach with a disposable real Chrome profile. A localStorage value created by one Chrome process was still present after that process exited and a second process started with the same isolated `--user-data-dir`.

Tests:
- `test/shell.d/webapp-isolation-test.sh`: pass
- `./test/cli`: pass
- `git diff --check`: pass
- `bash -n` on the changed scripts and test: pass
- `./test/all`: the isolation test passes, but the aggregate suite is not green on this machine because of unrelated environment failures, primarily an unconfigured Node mise shim, a missing `omarchy-pkgs` checkout, and compositor-dependent tests

One issue should be addressed before merge: the isolation name can currently escape the intended data directory.

`--isolate=<name>` is concatenated directly into:

    $HOME/.local/share/omarchy/webapps/$ISOLATE_NAME

The removal commands later parse that name from the desktop file and pass the resulting path to `rm -rf`.

In a disposable `$HOME`, I created a desktop entry containing:

    Exec=omarchy-launch-webapp https://example.com --isolate=../sentinel

Running `omarchy-webapp-remove "Escape App"` deleted:

    $HOME/.local/share/omarchy/sentinel

That directory is outside `webapps`.

I suggest sanitizing or rejecting unsafe isolation names when launching/installing, then independently verifying that the resolved deletion target is a strict child of `WEBAPP_DATA_DIR` in both remove commands. A regression test covering `../sentinel` for both `omarchy-webapp-remove` and `omarchy-webapp-remove-all` would protect this boundary.

PR #4766, opened 2026-02-27
Fix recently broken jq reference to "$p" value in omarchy-launch-or-focus

garethson:
In https://github.com/basecamp/omarchy/commit/7514ae7dcff654ff555862ea95b7033ea740924f that went out with the 3.4.0 release buried in the changes to  _Use consistent bash5 style for conditionals and quoting_ we can see that commit accidentally stripped the `$` from first `$p` reference when testing the class name. This can be easily reproduced:

```
~/code/tmp ❯ omarchy-launch-or-focus code 
jq: error: p/0 is not defined at <top-level>, line 1, column 33:
    .[]|select((.class|test("\\b" + p + "\\b";"i")) or (.title|test("\\b" + $p + "\\b";"i")))|.address```
```

With this fix, it works:
```
[master] ~/code/devnotes ❯ omarchy-launch-or-focus code
ok
```

### DIFF
--- a/bin/omarchy-launch-or-focus
+++ b/bin/omarchy-launch-or-focus
@@ -10,7 +10,7 @@ fi
 
 WINDOW_PATTERN="$1"
 LAUNCH_COMMAND="${2:-"uwsm-app -- $WINDOW_PATTERN"}"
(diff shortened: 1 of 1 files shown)


dhh:
Pushed as a hotfix to master. Thanks for catching!

PR #4775, opened 2026-02-27
tmux: bindings for moving to left or right pane

marcotroisi:
I'm really chuffed with this new tmux integration in omarchy. I'm especially enjoying the new `tdl` and `tdlm` bindings.

Upon launching `tdlm` in a directory, I immediately ran into a problem. I couldn't easily switch to a window whose number was higher than 9.

This PR aims to address that problem by introducing two new bindings: `ALT + LEFT ARROW` and `ALT + RIGHT ARROW`. These will move to the previous or next window, respectively.

Thanks everyone for this amazing new integration, and I hope this small contribution makes it even better.

https://github.com/user-attachments/assets/34406a8e-92be-4bbf-97a8-18c686ec62db

### DIFF
--- a/default/bash/envs
+++ b/default/bash/envs
@@ -4,4 +4,4 @@ export BAT_THEME=ansi
 
 # Duplicated from .config/uwsm/env so SSH works too
 export OMARCHY_PATH=$HOME/.local/share/omarchy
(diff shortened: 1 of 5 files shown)


dhh:
Included alongside up/down for sessions 👌

PR #4778, opened 2026-02-27
Fix monitor scaling toggle for multi-monitor setups

tmn73:
## Problem

On multi-monitor setups, using `Super+/` to cycle monitor scaling causes several issues:

1. **Monitor positions swap.** The script used `auto` for position, which at runtime places monitors by hardware detection order (internal display first) rather than config order — swapping left/right.

2. **Mouse can't cross between monitors.** Changing a monitor's scale changes its logical width (e.g. 2560px at 1x → 1600px at 1.6x), but adjacent monitors aren't repositioned — leaving a dead zone the mouse can't cross.

3. **Window gaps and layout settings are lost.** `hyprctl keyword monitor` resets layout properties like `gaps_in`, `gaps_out`, border size, and rounding.

4. **Position drift after full cycle.** Using relative deltas to reposition monitors causes cumulative drift (e.g. 53px after a 1x→1.6x→2x→3x→1x cycle) because Hyprland auto-adjusts positions between sequential commands.

## Fix

- **Preserve position:** use each monitor's current `x`/`y` instead of `auto`.
- **Reposition adjacent monitors with absolute positioning:** compute exact positions from scratch (`active_x + new_logical_width`, then stack each subsequent monitor) to avoid drift from Hyprland's inter-command adjustments.
- **Order operations to prevent overlap warnings:** when the monitor gets wider (scale decreasing), move adjacent monitors out of the way *first*; when it gets narrower (scale increasing), scale *first*, then close the gap.
- **Preserve current mode:** use each monitor's actual `width x height @ refreshRate` instead of `preferred`, which may select a lower refresh rate.
- **Restore layout settings:** re-source `looknfeel.conf` after reconfiguration to restore gaps, borders, rounding, and other layout properties.

## Note

Only repositions monitors to the right of the active one. This covers the common layout (laptop + external to the right). Monitors to the left are unaffected.

## Testing

Tested on a 2-monitor setup (external left at 2560x1440@360Hz + laptop right at 2560x1600@240Hz) through full scale cycles (1x → 1.6x → 2x → 3x → 1x) on both monitors. Positions return to exact original values after a full cycle, mouse movement works at every scale, no overlap warnings, and layout settings are preserved.

### DIFF
--- a/bin/omarchy-hyprland-monitor-scaling-cycle
+++ b/bin/omarchy-hyprland-monitor-scaling-cycle
@@ -4,9 +4,10 @@
 MONITOR_INFO=$(hyprctl monitors -j | jq -r '.[] | select(.focused == true)')
 ACTIVE_MONITOR=$(echo "$MONITOR_INFO" | jq -r '.name')
 CURRENT_SCALE=$(echo "$MONITOR_INFO" | jq -r '.scale')
-WIDTH=$(echo "$MONITOR_INFO" | jq -r '.width')
-HEIGHT=$(echo "$MONITOR_INFO" | jq -r '.height')
-REFRESH_RATE=$(echo "$MONITOR_INFO" | jq -r '.refreshRate')
+ACTIVE_MODE=$(echo "$MONITOR_INFO" | jq -r '"\(.width)x\(.height)@\(.refreshRate | round)"')
+ACTIVE_X=$(echo "$MONITOR_INFO" | jq -r '.x')
+ACTIVE_Y=$(echo "$MONITOR_INFO" | jq -r '.y')
+ACTIVE_WIDTH=$(echo "$MONITOR_INFO" | jq -r '.width')
 
 # Cycle through scales: 1 → 1.6 → 2 → 3 → 1
 CURRENT_INT=$(awk -v s="$CURRENT_SCALE" 'BEGIN { printf "%.0f", s * 10 }')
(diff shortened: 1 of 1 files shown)


joaobnobre:
+1, reproduces on my dual 1080p setup (HDMI-A-1 at `0x0`, DP-2 at `1920x0`). Every `Super + /` flips DP-2 to `0x0` and shoves HDMI-A-1 out of place. I had independently patched just the `auto` → `x,y` part locally and that alone fixes the position swap — glad to see this PR also covers the logical-width gap, operation ordering, and the looknfeel reset. Hope it lands in `dev` soon.

joaobnobre:
Heads-up @tmn73 — this PR targets `bin/omarchy-hyprland-monitor-scaling-toggle`, but that file was renamed to `-cycle` the day after you opened this (commit `8f5c1ee5` "Promote monitor scaling to Super + /" — pure `git mv`, zero content changes). Since then `39f1ca77` landed a partial refresh-rate fix that overlaps a bit with yours. As it stands GitHub can't auto-merge this PR because the target path no longer exists upstream.

If you're around and want to rebase onto the renamed `-cycle` file, your fix is meaningfully more complete than what I'm about to submit (I'm only tackling the `auto` → `x,y` position piece). Happy to close mine in favor of yours if you update this one.

tmn73:
On it

> Heads-up @tmn73 — this PR targets `bin/omarchy-hyprland-monitor-scaling-toggle`, but that file was renamed to `-cycle` the day after you opened this (commit `8f5c1ee5` "Promote monitor scaling to Super + /" — pure `git mv`, zero content changes). Since then `39f1ca77` landed a partial refresh-rate fix that overlaps a bit with yours. As it stands GitHub can't auto-merge this PR because the target path no longer exists upstream.
> 
> If you're around and want to rebase onto the renamed `-cycle` file, your fix is meaningfully more complete than what I'm about to submit (I'm only tackling the `auto` → `x,y` position piece). Happy to close mine in favor of yours if you update this one.

joaobnobre:
Rebase is clean — diff is +41/-4 against the current `bin/omarchy-hyprland-monitor-scaling-cycle`, applies without conflicts. Closing my stopgap #5278 in favor of this one.

Quick notes from reading the diff:
- Adjacent-monitor logic only touches monitors to the right (`.x > $ax`) — correct, since left-side monitors' positions don't depend on the focused monitor's width.
- Refresh rate `| round` handles fractional rates like `240.00101` cleanly.
- Commit subject still references `Super+Ctrl+Backspace` (the pre-rename binding) — cosmetic only.

+1 on behalf of a user reproducing the bug. Thanks for the fast rebase @tmn73 — hope this lands soon.

dhh:

PR #4786, opened 2026-02-27
tmux: attach to session on tmux terminal startup

lucasrabiec:
This pull request makes a small but useful update to the terminal keybinding for launching `tmux` session.
Now, when you use the `SUPER+ALT+RETURN` shortcut, it will attempt to attach to an existing `tmux` session, or create a new one called "Work" if none exists. Makes it consistent with `t` alias.


dhh:
I went back and forth on this, but ultimately decided that I preferred the option where a new session is started. This makes it more suitable to running multiple times, like if you want different sessions on different workspaces. But could also see if someone is a strict "one terminal" person, this setup would make sense, so it's good this is just living in the user-space bindings.conf 👌

PR #4794, opened 2026-02-28
ADD: Toggle Layout

dhms013:
As per February 28, 2026, [hyprland](https://hypr.land/news/update54/) move hyprscrolling into their core. So I add new toggle in [omarchy-menu > trigger > toggle][(https://github.com/dhms013/omarchy/blob/ff5d177f42da9590ce8f4978108dac7813fda994/bin/omarchy-menu#L168](https://github.com/dhms013/omarchy/blob/73ff417e18dcc52299c875146e1d585a01445678/bin/omarchy-menu#L173)) that execute new [omarchy-toggle-layout](https://github.com/dhms013/omarchy/blob/73ff417e18dcc52299c875146e1d585a01445678/bin/omarchy-toggle-layout).

Goals :
- Make user able to switch between dwindle-scrolling via keybinds or menu

here's the preview

https://github.com/user-attachments/assets/e6dd20da-d793-4f8b-9ed0-ea21e513fa43

### DIFF
--- a/bin/omarchy-menu
+++ b/bin/omarchy-menu
@@ -161,7 +161,7 @@ show_share_menu() {
 }
 
 show_toggle_menu() {
-  case $(menu "Toggle" "󱄄  Screensaver\n󰔎  Nightlight\n󱫖  Idle Lock\n󰍜  Top Bar\n  Window Gaps\n  1-Window Ratio\n󰍹  Display Scaling") in
+  case $(menu "Toggle" "󱄄  Screensaver\n󰔎  Nightlight\n󱫖  Idle Lock\n󰍜  Top Bar\n  Window Gaps\n  1-Window Ratio\n󰍹  Display Scaling\n   Layout") in
 
   *Screensaver*) omarchy-toggle-screensaver ;;
   *Nightlight*) omarchy-toggle-nightlight ;;
@@ -170,7 +170,25 @@ show_toggle_menu() {
   *Ratio*) omarchy-hyprland-window-single-square-aspect-toggle ;;
   *Gaps*) omarchy-hyprland-window-gaps-toggle ;;
   *Scaling*) omarchy-hyprland-monitor-scaling-toggle ;;
+  *Layout*) show_layout_menu ;;
   *) back_to show_trigger_menu ;;
+  esac
+}
+
+show_layout_menu() {
+  local current
+  current=$(omarchy-toggle-layout --get)
+
+  local options="󰕰  Dwindle\n  Master\n  Monocle\n󱒇   Scrolling"
+  local preselect
+  preselect=$(echo -e "$options" | grep -i "$current")
+
+  case $(menu "Layout" "$options" "" "$preselect") in
+  *Dwindle*) omarchy-toggle-layout --set=dwindle ;;
+  *Master*) omarchy-toggle-layout --set=master ;;
+  *Monocle*) omarchy-toggle-layout --set=monocle ;;
+  *Scrolling*) omarchy-toggle-layout --set=scrolling ;;
+  *) back_to show_toggle_menu ;;
   esac
 }
 
--- a/bin/omarchy-toggle-layout
+++ b/bin/omarchy-toggle-layout
@@ -1,0 +1,32 @@
+#!/bin/bash
+
+LAYOUTS=("scrolling" "master" "monocle" "dwindle")
+
+get_layout() {
+  hyprctl getoption general:layout -j 2>/dev/null |
+    python3 -c "import sys,json; print(json.load(sys.stdin)['str'])" 2>/dev/null
+}
+
+set_layout() {
+  local layout="$1"
+  local valid=false
+  for l in "${LAYOUTS[@]}"; do
+    [[ "$l" == "$layout" ]] && valid=true && break
+  done
+
+  if [[ "$valid" == "false" ]]; then
+    notify-send "Invalid layout: $layout" \
+      "Valid options: ${LAYOUTS[*]}" -u critical -t 3000
+    exit 1
+  fi
+
+  hyprctl keyword general:layout "$layout"
+  notify-send "Layout → ${layout^}" -t 2000
+}
+
+for arg in "$@"; do
+  case "$arg" in
+  --set=*) set_layout "${arg#*=}" ;;
+  --get) get_layout ;;
+  esac
+done


abenz1267:
Puh... i'm torn. This basically comes down to: does DHH personally want it.

Because otherwise: i'd say this is even too simplistic. Why only dwingle/scrolling? What about master and monocle?

This is **your** opinionated keybind and unless DHH agree's and also wants to use it, i don't see this going forward.

Given hyprland 0.54 now has per workspace layouts, i'd much rather prefer a solution to include all possible layouts ... or something the user can configure.

Tbh this can easily be added a custom menu into walker so you don't even have to "pollute" the omarchy-menu. And no one is stopping anyone from creating custom keybinds.

dhms013:
> Puh... i'm torn. This basically comes down to: does DHH personally want it.
> 
> Because otherwise: i'd say this is even too simplistic. Why only dwingle/scrolling? What about master and monocle?
> 
> This is **your** opinionated keybind and unless DHH agree's and also wants to use it, i don't see this going forward.
> 
> Given hyprland 0.54 now has per workspace layouts, i'd much rather prefer a solution to include all possible layouts ... or something the user can configure.
> 
> Tbh this can easily be added a custom menu into walker so you don't even have to "pollute" the omarchy-menu. And no one is stopping anyone from creating custom keybinds.



> Puh... i'm torn. This basically comes down to: does DHH personally want it.
> 
> Because otherwise: i'd say this is even too simplistic. Why only dwingle/scrolling? What about master and monocle?
> 
> This is **your** opinionated keybind and unless DHH agree's and also wants to use it, i don't see this going forward.
> 
> Given hyprland 0.54 now has per workspace layouts, i'd much rather prefer a solution to include all possible layouts ... or something the user can configure.
> 
> Tbh this can easily be added a custom menu into walker so you don't even have to "pollute" the omarchy-menu. And no one is stopping anyone from creating custom keybinds.

Thanks for your reply. But let me be clear

1. Yes, I'm aware that Omarchy is DHH's opinion before my first time installing it as my daily driver.
2. I'm proposing here, not forcing my opinion into Omarchy. SO yes, it's up to DHH wether he accept or not, and I dont mind. Except it's considered bad to 'propose idea' in omarchy, then I'll stop
3. opinionated keybind => Yes. that's my opinionated keybind. And I already add comment regarding that.
4. Master and Monocle layout? done.. I need to resolve the conflict first (will do if DHH accepted this propose)
5. custom-menu via walker so I'm not populate the omarchy-menu? then please do. I proposing the changes into omarchy-menu so the 'trigger' action is centralized in omarchy-menu. and once again, I'm aware that omarchy is DHH' opinion. and I'm fine if this PR is rejected. I'm proposing here, not forcing my opinion.
6. I propose this so other people (especially new comer who wan't to use omarchy as their first linux), with all due respect, doesn't need to create the custom menu.


https://github.com/user-attachments/assets/b93c2d06-3be6-44e4-a971-a8eec16dbdc6

abenz1267:
> 5. custom-menu via walker so I'm not populate the omarchy-menu? then please do. I proposing the changes into omarchy-menu so the 'trigger' action is centralized in omarchy-menu. and once again, I'm aware that omarchy is DHH' opinion. and I'm fine if this PR is rejected. I'm proposing here, not forcing my opinion.
> 6. I propose this so other people (especially new comer who wan't to use omarchy as their first linux), with all due respect, doesn't need to create the custom menu.

Omarchy ships other custom walker menu's already. Users dont need to create anything manually.

A layout change is something you wanna do quickly. Going 2 levels deep in a menu is not that.

bin/omarchy-menu
```diff
@@ -161,15 +161,33 @@ show_share_menu() {
...
+show_layout_menu() {
+  local current
+  current=$(omarchy-toggle-layout --get)
+
+  local options="󰕰  Dwindle\n  Master\n  Monocle\n󱒇   Scrolling"
+  local preselect
+  preselect=$(echo -e "$options" | grep -i "$current")
+
+  case $(menu "Layout" "$options" "" "$preselect") in
+  *Dwindle*) toggle-layout --set=dwindle ;;
```
abenz1267: am i blind or shouldn't this be `omarchy-toggle-layout`?

bin/omarchy-toggle-layout
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
```
abenz1267, line 1: should probably rename this from `toggle` => `cycle`

bin/omarchy-toggle-layout
```diff
@@ -0,0 +1,32 @@
...
+
+LAYOUTS=("scrolling" "master" "monocle" "dwindle")
+
+get_layout() {
+  hyprctl getoption general:layout -j 2>/dev/null |
+    python3 -c "import sys,json; print(json.load(sys.stdin)['str'])" 2>/dev/null
```
abenz1267, line 7: why not just use `jq`?

bin/omarchy-menu
dhms013: thank you for noticing..
no, you're not blinds. It's purely my fault for forgot to  edit those section.. already pushed the fix

bin/omarchy-toggle-layout
dhms013: then I'll be wondering, what is this 'cycle' about? is it cycle between windows, or workspace?
the new commit revamp the cycle behaviour into select layout behaviour.. 
but, I'm ready to switch the behaviour into 'cycle' again if DHH prefer that

gonna edit the PR detail later..
Thank you for your suggestion

bin/omarchy-toggle-layout
dhms013: No specific reason..
Do and/or why I have to? (purely ask)

abenz1267: jq is already widely used across omarchy for processing json in terminal, so it'd be a natural choice.

bin/omarchy-toggle-layout
abenz1267: `toggling` something implies 2 options. But you have 4 layouts. So you "cycle" through them, you don't toggle them. Just wording.

bin/omarchy-toggle-layout
dhms013: hmm.. understood..
I'll check again when I get home.

bin/omarchy-toggle-layout
dhms013: ah right. I got what you mean..
So which one better for omarchy? rename it then switch the behaviour into cycle, or move the menu into 'Style' instead?
but tbh I'd like to hear DHH's opinion about this proposal first before making any changes..

dhh:
Appreciate you taking a stab at this, but I ended up going with something simpler. I never use master or monocle, so just wanted something to toggle between dwindle and scrolling: https://github.com/basecamp/omarchy/commit/42a437ad481f9829c3f07298ed2b5fbeaf92bccd

dhms013:
> Appreciate you taking a stab at this, but I ended up going with something simpler. I never use master or monocle, so just wanted something to toggle between dwindle and scrolling: [42a437a](https://github.com/basecamp/omarchy/commit/42a437ad481f9829c3f07298ed2b5fbeaf92bccd)

Then my first version will works for you 🤣

But thanks for adding the switch toggle layout

PR #4810, opened 2026-02-28
Esp idf

joshuacox:
Just adding ESP-IDF in.


dhh:
Appreciate the suggestion, but this is too specific for Omarchy, and requires too much code to get there. Would be a nice external package or documentation in the discussions forum ✌️

joshuacox:
no worries, too specific for sure, I merely needed a friend to duplicate my dev env and used omarchy to do so, thought we would contribute if anyone wanted.

dhh:
Much appreciated. I think doing an AUR package would be a great middle ground. Then you can still have it available easily as Install > AUR 👌

PR #4813, opened 2026-02-28
Add zsh with shell switching and enhanced tmux management

TheNeovimmer:
Add zsh support with full compatibility with bash aliases and functions
- Add shell switching capability via Walker menu (Setup > Config > Shell)
- Add fuzzy tmux session manager with preview capabilities


dhh:
Appreciate the work on this, but this is vastly too much. We have a pattern in omaterm for how to do this in a minimal way where everything is shared except for a single zsh config file. Will explore that instead.

PR #4814, opened 2026-02-28
Update comment in hypridle.conf

ReloadingBee:


pomartel:
See https://github.com/basecamp/omarchy/pull/4414

Issue #4754, opened 2026-02-26
No icon for `Omarchy Themes` in walker `/`


rez1-dev:
### System details

Omarchy 3.4.0

### What's wrong?

<img width="320" height="348" alt="Image" src="https://github.com/user-attachments/assets/b8955201-9ea3-4113-9f1c-a7475346c821" />

benkelcher77:
@rez1coder sorry, i am relatively new here - was able to reproduce and solve this issue by specifying the `Icon` in `default/elephant/omarchy_themes.lua`. 

is there an icon that was used for this previously? if not, i think that `preferences-desktop-theme` might make sense:

<img width="845" height="536" alt="Image" src="https://github.com/user-attachments/assets/bbade298-fff8-4b1e-acdd-b41c69fade91" />

rez1-dev:
> sorry, i am relatively new here

@benkelcher77 don't be

> was able to reproduce and solve this issue 

cool

> is there an icon that was used for this previously?

afaik, there's no icon other than omarchy logo and omarchy.ttf font

Please, open a pull request.

PR #4820, opened 2026-02-28
specify icon for omarchy_themes

benkelcher77:
addresses issue #4754. open to discussion on which specific icon should be used here!


dhh:
This wasn't intended to show in that list. I've hidden it instead 👍

PR #4826, opened 2026-02-28
Fix changegroupactive offset for Hyprland 0.54

pkwagner:
Hyprland 0.54 apparently starts counting grouped windows from 0, which caused SUPER + ALT + 1 to highlight the second instead of the first tab.


dhh:
@vaxerski Was this intentional or a bug?

vaxerski:
bug

vaxerski:
fixed in git will be in 0.54.1

PR #4829, opened 2026-03-01
Add Calibre theme integration

chipkoziara:
## Problem Statement
Calibre doesn't natively update with Omarchy themes, so I added a new script and call in `bin/omarchy-theme-set` to trigger Calibre theme changes. Calibre is a little finicky with updating themes, so I added logic to make sure it closes and applies the theme whenever a user changes the Omarchy theme.

## Summary
Syncs the current Omarchy theme to Calibre's custom palette system when switching themes. Follows the same pattern as the existing Obsidian, VS Code, and browser integrations.
- Parses `colors.toml` and maps colors to Qt QPalette roles (deriving shades like `AlternateBase` and `PlaceholderText` via a `blend()` helper since `colors.toml` only has ~6 base colors but Calibre's palette needs 15)
- Detects light/dark mode via light.mode marker
- Writes the palette to Calibre's `gprefs` via `calibre-debug -c` (Calibre stores palettes in a Python-managed JSON config, not a plain file)
- Restarts Calibre if it was running -- kills it _before_ writing the palette because Calibre's shutdown handler saves its in-memory config to disk, which would overwrite our changes
- Reopens Calibre on its original workspace after restarting -- captures the workspace via `hyprctl` before killing, then sets a temporary `windowrule` so it opens silently in the right place
- Guarded by `omarchy-cmd-present calibre` -- no-op if Calibre isn't installed

## Files
- `bin/omarchy-theme-set-calibre` → new script
- `bin/omarchy-theme-set` → added `omarchy-theme-set-calibre` between obsidian and keyboard

### DIFF
--- a/bin/omarchy-theme-set
+++ b/bin/omarchy-theme-set
@@ -201,6 +201,7 @@ post_theme_commands=(
   omarchy-theme-set-browser
   omarchy-theme-set-vscode
   omarchy-theme-set-obsidian
+  omarchy-theme-set-calibre
   omarchy-theme-set-keyboard
 )
 
--- a/bin/omarchy-theme-set-calibre
+++ b/bin/omarchy-theme-set-calibre
@@ -1,0 +1,101 @@
+#!/bin/bash
+
+# Sync Omarchy theme to Calibre
+
+omarchy-cmd-present calibre || exit 0
+
+CURRENT_THEME_DIR="$HOME/.config/omarchy/current/theme"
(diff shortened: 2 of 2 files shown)


chipkoziara:
Here's a demo of Calibre's theme updating with Omarchy theme changes:

https://github.com/user-attachments/assets/05ee8257-f2b4-4d37-a3d0-eafe7473a15d

PR #4837, opened 2026-03-01
feat: Timeout for encryption password input

brutalsam96:
Recently i had my laptop turned on by accident in my backpack and after 4-5 hours i noticed that my bag was warm after inspecting i found out that my laptop was on all this time in LUKS decryption password prompt screen stayed there for roughly 5 hours, laptop was hot but i was lucky that it did not inflict any damage to the hardware, i was sure i would have an omarchy logo burned in in on my OLED screen, luckily it is fine.
I modified encrypt hook to include 120 second timeout for both plymouth and fallback shell to power off after the timeout, this helps prevent cases such as above, where device is turned on by accident and display is left on, leading to overheating or display damages if left on for long periods of time.
Added changes as migration script should work fine to patch the encrypt hook and apply using limine-mkinitcpio.
Tested couple of times on my own hardware with no issues, might require some more testing.

### DIFF
--- a/config/mkinitcpio/luks-timeout.patch
+++ b/config/mkinitcpio/luks-timeout.patch
@@ -1,0 +1,41 @@
+--- encrypt	2026-02-28 18:02:14.718126028 +0500
++++ encrypt-upd	2026-03-01 14:00:47.917985370 +0500
+@@ -100,18 +100,38 @@
+                 fi
+             fi
+             # Ask for a passphrase
++            local IDLE_TIMEOUT=120
+             if [ "${dopassphrase}" -gt 0 ]; then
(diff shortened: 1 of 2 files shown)


brutalsam96:
With further inspection i can see that mkinit package reinstall/update can overwrite changes, think it needs to be added to omarchy mkinit hooks, i can do this later if maintainers approve the change. 
@dhh @ryanrhughes if you think this change needs to be implemented, can you give me pointers on how omarchy custom hooks are stored and applied, to my understanding need to replace encrypt hook with new one and disable the old one.

dhh:
Closing because encryption-prompt installation is now owned by the ISO rather than this repository’s former installer flow. The timeout idea should be proposed there if still needed.

PR #4840, opened 2026-03-01
Enable SSD TRIM on LUKS-encrypted drives

### LINKED ISSUE #2229: Enable SSD TRIM
### What do you need?

I did a fresh installation of Omachy v3.0.2 on a laptop with an nvme SSD, and SSD TRIM does not seem to be working out of the box. This can result in *poor performance and reduced life of my SSD*.

My enhancement request is to enable SSD trimming automatically within in the installer, or maybe following a prompt. I understand that Arch Linux does not enable it by default because it can have theoretical security implications (see [wiki](https://wiki.archlinux.org/title/Dm-crypt/Specialties#Discard/TRIM_support_for_solid_state_drives_(SSD))), but I think that's pretty far fetched, and it's not just me: I believe that basically all Linux distros enable it out-of-the box for systems with full disk encryption, including Ubuntu.

See my investigation below, including how to enable it.

Following the [Arch Wiki on SSD TRIM](https://wiki.archlinux.org/title/Solid_state_drive#TRIM), I can confirm that by default my encrypted root partition is not supporting TRIM (by having zero in the column `DISC-GRAN`:
```
~ ❯ : lsblk --discard
NAME        DISC-ALN DISC-GRAN DISC-MAX DISC-ZERO
zram0              0        4K       2T         0
nvme0n1            0        4K       2T         0
├─nvme0n1p1        0        4K       2T         0
└─nvme0n1p2        0        4K       2T         0
  └─root           0        0B       0B         0
```

And triggering a manual trimming also confirms this:

```
~ ❯ : sudo fstrim -av
/boot: 1.5 GiB (1632141312 bytes) trimmed on /dev/nvme0n1p1
```

The fix is documented in the [LUKS2](https://wiki.archlinux.org/title/Dm-crypt/Specialties#LUKS2) page: 

```
# sudo cryptsetup --allow-discards --persistent refresh root
```

It is persistent, and even starts to work immediately (note the `4K` value of `root` and that `fstrim` reported trimming the root partition):

```
~ ❯ : lsblk --discard
NAME        DISC-ALN DISC-GRAN DISC-MAX DISC-ZERO
sda                0        0B       0B         0
└─sda1             0        0B       0B         0
zram0              0        4K       2T         0
nvme0n1            0        4K       2T         0
├─nvme0n1p1        0        4K       2T         0
└─nvme0n1p2        0        4K       2T         0
  └─root           0        4K       2T         0


~ ❯ : sudo fstrim -av
/boot: 1.5 GiB (1632141312 bytes) trimmed on /dev/nvme0n1p1
/: 908.9 GiB (975896055808 bytes) trimmed on /dev/mapper/root
```

I don't think that we need any other step (like mount option in `fstab` or [periodic trim](https://wiki.archlinux.org/title/Solid_state_drive#Periodic_TRIM)), as according to the [wiki](https://wiki.archlinux.org/title/Solid_state_drive#TRIM):
> Asynchronous discard is *enabled* (https://wiki.archlinux.org/title/Btrfs#SSD_TRIM) by default since kernel 6.2.

n8himmel:
## Summary
- Enables `allow-discards` on LUKS volumes via `cryptsetup --persistent refresh` so TRIM commands pass through dm-crypt to the SSD controller
- Enables `fstrim.timer` for weekly periodic TRIM
- Includes install-time hardware detection script and migration for existing users

Fixes #2229

## Context

Every Omarchy install uses LUKS encryption, and the `encrypt` mkinitcpio hook silently drops all TRIM/DEALLOCATE commands by default. On a 1 TB NVMe drive, this caused 400+ GiB of stale blocks the controller couldn't reclaim — unnecessary write amplification that degrades both performance and SSD lifespan.

The approach uses `cryptsetup --allow-discards --persistent refresh` rather than kernel parameter modification, writing the flag directly into the LUKS2 header. This is the method recommended by the [Arch Wiki](https://wiki.archlinux.org/title/Dm-crypt/Specialties#LUKS2) and works regardless of boot configuration.

**Security note:** `allow-discards` leaks block usage patterns to physical disk forensics. For most users, the performance/longevity benefits far outweigh this risk.

## Files changed
- `install/config/hardware/luks-trim.sh` — new install-time script (detects LUKS, enables discards + fstrim.timer)
- `install/config/all.sh` — registers the new hardware script
- `migrations/1772373643.sh` — applies the same fix to existing installations on update

## Test plan
- [ ] Fresh Omarchy install with LUKS: verify `lsblk --discard` shows non-zero DISC-GRAN on crypt device
- [ ] Fresh install: verify `systemctl is-enabled fstrim.timer` returns "enabled"
- [ ] Existing install: run `omarchy-migrate` and verify both changes apply
- [ ] Non-LUKS install (if any): verify script is a no-op (no errors, no changes)
- [ ] Verify `sudo fstrim -av` successfully trims the root partition after changes

### DIFF
--- a/install/config/all.sh
+++ b/install/config/all.sh
@@ -44,3 +44,4 @@ run_logged $OMARCHY_INSTALL/config/hardware/fix-asus-rog-mic.sh
 run_logged $OMARCHY_INSTALL/config/hardware/fix-yt6801-ethernet-adapter.sh
 run_logged $OMARCHY_INSTALL/config/hardware/fix-synaptic-touchpad.sh
 run_logged $OMARCHY_INSTALL/config/hardware/framework16-qmk-hid.sh
+run_logged $OMARCHY_INSTALL/config/hardware/luks-trim.sh
--- a/install/config/hardware/luks-trim.sh
+++ b/install/config/hardware/luks-trim.sh
@@ -1,0 +1,20 @@
+# Enable SSD TRIM on LUKS-encrypted drives
+# Without this, dm-crypt silently drops all TRIM/DEALLOCATE commands,
+# causing unnecessary write amplification and reduced SSD lifespan.
+# See: https://wiki.archlinux.org/title/Dm-crypt/Specialties#Discard/TRIM_support_for_solid_state_drives_(SSD)
+
+if lsblk --noheadings -o TYPE | grep -q "crypt"; then
+  echo "LUKS encryption detected. Enabling SSD TRIM support..."
+
+  # Set allow-discards persistently in the LUKS2 header so TRIM commands
+  # pass through dm-crypt to the underlying SSD controller.
+  for dev in $(lsblk --noheadings -o NAME,TYPE | awk '$2=="crypt" {print $1}'); do
+    if ! sudo cryptsetup luksDump /dev/mapper/$dev 2>/dev/null | grep -q "allow-discards"; then
+      echo "Enabling allow-discards on /dev/mapper/$dev"
+      sudo cryptsetup --allow-discards --persistent refresh $dev
+    fi
+  done
+
+  # Enable weekly TRIM via fstrim.timer
+  chrootable_systemctl_enable fstrim.timer
+fi
--- a/migrations/1772373643.sh
+++ b/migrations/1772373643.sh
@@ -1,0 +1,16 @@
+echo "Enable SSD TRIM on LUKS-encrypted drives"
+
+# Without allow-discards, dm-crypt silently drops all TRIM commands,
+# causing write amplification and reduced SSD lifespan.
+# See: https://wiki.archlinux.org/title/Dm-crypt/Specialties#Discard/TRIM_support_for_solid_state_drives_(SSD)
+
+if lsblk --noheadings -o TYPE | grep -q "crypt"; then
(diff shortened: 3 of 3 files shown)


woopstar:
Very interessting catch

woopstar:
Maybe also look into https://wiki.archlinux.org/title/Dm-crypt/Specialties#Disable_workqueue_for_increased_solid_state_drive_(SSD)_performance

yevshev:
Since continuous trim is already enabled by default for BTRFS  ( `discard=async` default mount option), does adding periodic trim (`fstrim.timer`)  provide any benefit? https://wiki.archlinux.org/title/Btrfs#SSD_TRIM

zayatura:
> Since continuous trim is already enabled by default for BTRFS ( discard=async default mount option), does adding periodic trim (fstrim.timer) provide any benefit? https://wiki.archlinux.org/title/Btrfs#SSD_TRIM

Enabling the fstrim timer would help with the initial trim, as `discard=async` only trims recently discarded blocks (e.g. deleted files) and does not send trim all unused space; the fstrim timer would. I did not check the pull request, but in case there is a one-time fstrim command, then this is not needed. Periodic fstrim would also trim any additional trim-capable device the user might have, such as additional SSDs and external SSDs.

dhh:
Closing in favor of the newer LUKS TRIM implementation in #5332.

PR #4841, opened 2026-03-01
Fall back to DDC/CI for brightness on external monitors

justinontheporch:
## Summary

On desktops with external monitors, there are no `/sys/class/backlight` devices, so the brightness keys (`XF86MonBrightnessUp/Down`) do nothing. This adds a `ddcutil` fallback that controls the monitor brightness via DDC/CI when no backlight device is found.

- When no backlight device exists and `ddcutil` is installed, uses DDC/CI (VCP feature `0x10`) to read/write brightness
- The existing laptop backlight path is untouched — DDC/CI is only used as a fallback

**Performance optimizations** (DDC/CI over I2C is inherently slow):
- Low sleep multiplier to minimize I2C bus wait times (~0.4s vs ~3s default)
- Cache file to skip the DDC read on rapid keypresses (~0.02s on cache hit)
- Background writes so the OSD appears immediately
- `flock -n` to drop concurrent invocations instead of queuing them up

## Prerequisites for users

DDC/CI requires `ddcutil`, the `i2c-dev` kernel module, and i2c device access:

```bash
sudo pacman -S ddcutil
sudo modprobe i2c-dev
echo 'i2c-dev' | sudo tee /etc/modules-load.d/i2c-dev.conf
sudo usermod -aG i2c $USER  # then logout/login
```

## Test plan

- [ ] On a laptop with a built-in display: brightness keys still work via `brightnessctl` (no behavior change)
- [ ] On a desktop with an external monitor (no `/sys/class/backlight`): brightness keys control the monitor via DDC/CI with OSD feedback
- [ ] Rapid key repeat: OSD is responsive, no pile-up of delayed popups
- [ ] Without `ddcutil` installed: falls through to `brightnessctl` path (existing behavior)

### DIFF
--- a/bin/omarchy-brightness-display
+++ b/bin/omarchy-brightness-display
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 # Adjust brightness on the most likely display device.
+# Falls back to DDC/CI (ddcutil) for external monitors when no backlight device exists.
 # Usage: omarchy-brightness-display <step>
 
 step="${1:-+5%}"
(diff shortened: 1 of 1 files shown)


justinontheporch:
Wow, I did not try to have Claude open this PR.  Sorry about that.

PR #4848, opened 2026-03-01
Automatically adjust the display brightness on battery and AC

pomartel:
The biggest drain on a laptop battery is the display.  This PR adds a script that creates a udev rule to automatically adjust the display brightness when the laptop is plugged or unplugged. Just like a mac!

I extracted the display backlight detection logic from `omarchy-brightness-display` to put it in a new script. I adjusted it a bit to make the globing work.

The script can be run interactively or not. I thought it could be added to the Omarchy menu under Setup. I wasn't sure how to name it or if there should be a new submenu for power or display management so I haven't created the menu item yet.


dhh:
Appreciate the work you've put into this, but this is something I've always hated on the mac 😂. I think manual controls here are actually better. But would be nice to document this for someone who wants it somewhere ✌️

PR #4849, opened 2026-03-01
Fix typo: screenrecording to screen recording

04cb:
Fixes typo in bin/omarchy-cmd-screenrecord comment where 'screenrecording' should be 'screen recording'.


dhh:
We call it screenrecording to match screenshot. Same in the menu. It's fine.

PR #4862, opened 2026-03-02
fix snapshot delete on failed update

Mrid22:
My bad on the last pr (https://github.com/basecamp/omarchy/pull/4531), i wrongfully assumed that the latest snapshot was the first one which doesn't make a lot of sense in hindsight, this time it should actually work, it uses tail and awk to get the latest generation id from `snapper list` and then remove it, i tested it locally ( probably should've done that before) and it does work as intended for me. Also, this is a bit less important but my handle is @Mrid22 not my display name, ty! btw thanks @ryanrhughes for catching this

### DIFF
--- a/bin/omarchy-snapshot
+++ b/bin/omarchy-snapshot
@@ -6,7 +6,7 @@ COMMAND="$1"
 OMARCHY_PATH=${OMARCHY_PATH:-$HOME/.local/share/omarchy}
 
 if [[ -z $COMMAND ]]; then
-  echo "Usage: omarchy-snapshot <create|restore>" >&2
+  echo "Usage: omarchy-snapshot <create|restore|delete>" >&2
   exit 1
 fi
 
@@ -26,9 +26,12 @@ create)
   for config in "${CONFIGS[@]}"; do
     sudo snapper -c "$config" create -c number -d "$DESC"
   done
+
   echo
   ;;
 restore)
   sudo limine-snapper-restore
   ;;
+delete)
+  sudo snapper delete $(echo $(sudo snapper list | tail -n 1) | awk '{print substr($0,0,2)}') # gets the number of the most recent snapshot and deletes it
 esac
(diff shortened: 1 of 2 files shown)


ryanrhughes:
We don't want to delete the snapshot on a failed update.

Playing out a scenario; let's say we run `omarchy-update`, it generates a snapshot of the current state before beginning, the update fails and initramfs fails to build properly. 

By leaving the snapshot in place, if I reboot, and find my system un-bootable, I can restore the snapshot and try again.

If `omarchy-update` realizes the update failed, then deletes the snapshot and I reboot; I'm screwed with no easy path back.

Even if I catch it and re-run `omarchy-update`, that will generate a snapshot of a **failed** state at that point, giving me a worse snapshot to restore to.

I can't really come up with any scenario where this is beneficial unless you see one I don't.

Mrid22:
> We don't want to delete the snapshot on a failed update.
> 
> Playing out a scenario; let's say we run `omarchy-update`, it generates a snapshot of the current state before beginning, the update fails and initramfs fails to build properly.
> 
> By leaving the snapshot in place, if I reboot, and find my system un-bootable, I can restore the snapshot and try again.
> 
> If `omarchy-update` realizes the update failed, then deletes the snapshot and I reboot; I'm screwed with no easy path back.
> 
> Even if I catch it and re-run `omarchy-update`, that will generate a snapshot of a **failed** state at that point, giving me a worse snapshot to restore to.
> 
> I can't really come up with any scenario where this is beneficial unless you see one I don't.

for me it was really just the space thing, so that you dont get duplicate snapshots but what you said makes sense

Edit: i'm realizing now that we *could* just periodically check the Description of the generations and delete one if they match, since that would be 2 snapshots of the same version, so if you want to implement this, I'll start working on it

ryanrhughes:
We would need a rock solid way to ensure we know whether a snapshot was of a good state or bad.

We already limit the max space snapshots can take as well as the maximum number. This feels sufficient for space management.

PR #4871, opened 2026-03-02
New Binding for Scrolling Layout, Enable mouse horizontal scroll

Jon-5:
The new niri-like Scrolling Layout is currently accessible with SUPER + Left Arrow or SUPER + Right Arrow

Using SUPER + horizontal scroll is an intuitive binding that complements the SUPER + vertical scroll for workspace switching


dhh:

PR #4872, opened 2026-03-02
Update icons.theme to use Yaru-dark instead of Yaru-gray

xcession:
Since "Yaru-gray" does not exist in the default Yaru package, the theme unintentionally falls back to Adwaita. This fixes the issue by switching to "Yaru-dark", the closest available replacement for the theme's

### DIFF
--- a/themes/vantablack/icons.theme
+++ b/themes/vantablack/icons.theme
@@ -1,1 +1,1 @@
-Yaru-gray
+Yaru-dark


bjarneo:
Thank you. I forgot to fix this myself!

dhh:
I actually think what we have now, the fallback icons, work better. Even if it's a mistake. The yaru-dark has a lot of color, and the whole point is for this to be monochrome. So I'd take a fix that just sets it correctly to what we already have.

PR #4873, opened 2026-03-02
Bundle Zed editor with omazed for themes syncing

pomartel:
[Omazed](https://github.com/APS6/omazed) provides live theme switching for the Zed editor. It is maintained as an AUR package by @APS6. It uses Omarchy's themes hooks and `color.toml`. I have been using it for a while and it works very well.

This PR bundles omazed with Zed when installing it from the Editors menu.


pomartel:
I'm just not sure how we handle AUR packages in Omarchy. Maybe this should come with a prompt.

dhh:
Given the new theme template system, I'd be willing to take a single zed tpl file instead of adding this dependency.

pomartel:
@APS6 Do you think that's something that could be done? I would very like to have this merged in Omarchy but it's your project and I don't want to intercede.

pomartel:
@dhh I extracted the Omazed templating rules and used the same logic as we have in place for VS Code. I tested it thoroughly and I think it's as minimal as it can be.

dhh:
Oh wow. I'm surprised that it's such a huge chunk. I would have thought we could have gotten away with something on the scale of Alacritty or Ghostty. Hmm. Don't see us taking on ~700 lines of code for an optional editor, then.

APS6:
@pomartel Sorry for not replying earlier. I forgot to respond in the midst of my examinations.

@dhh Yeah, editor themes are quite complex. I explained my reasoning earlier in detail here: https://github.com/APS6/omazed/issues/4#issuecomment-3847072890

dhh:
Okay, given all this, I'm happy to just installed omazed when we install zed via the menu 👍

pomartel:
@dhh That totally makes sense. Let me revert to my initial commit then.

@APS6  Thanks for your input and thanks again for your wonderful plugin. Trying to integrate it into Omarchy made me realize the complexity of it all! Can you validate that I installed the package the right way?

APS6:
LGTM 👍

APS6:
@dhh Hey, it's been a while so I figured I should ask if we are waiting for something specific?

annoyedmilk:
Another direction worth considering would be a pointer-only sync that follows the same pattern as omarchy-theme-set-vscode.

Each theme with an upstream Zed port gets a tiny themes/<name>/zed.json:
{ "name": "Tokyo Night", "extension": "tokyo-night" }

A new omarchy-theme-set-zed, wired into omarchy-theme-set next to the vscode setter, reads that file, sets "theme" in
~/.config/zed/settings.json, and adds the slug to auto_install_extensions so Zed installs the extension on next start.

Upside: no AUR dependency, and the colors come straight from whatever the upstream extension author shipped.
Downside: themes without an upstream Zed port (lumon, matte-black, vantablack, etc.) won't sync until someone publishes a native extension.

Today that covers tokyo-night, catppuccin (mocha + latte), everforest, flexoki, kanagawa, nord, and rose-pine.
Gruvbox is already a Zed built-in so it works without a published extension.

pomartel:
@annoyedmilk That sounds like a fine idea. But does it mean it would not support instant theme change? You would need to restart Zed to apply the new theme if it is not already installed?

annoyedmilk:
@pomartel No restart needed, I just had to wait for the background download to finish. Zed live-reloads settings.json.

Subsequent switches to the same theme are instant as you can see in the screen recording. I'd need to verify, but I think the only path that genuinely needs Zed to be opened at least once is the very first install of an extension that ships in auto_install_extensions.

https://github.com/user-attachments/assets/534d7337-65bc-4dd0-8cff-16aedf29da52

pomartel:
That's neat! Do you want to open a PR for this? If so, link it here and we can let DHH decide the best way forward.

dhh:
I like this with full compatibility for all themes. I've added omazed to the OPR: https://github.com/omacom-io/omarchy-pkgs/commit/c10bb6b1169f39316082206c3eb76d63f73b5f25

PR #4874, opened 2026-03-02
feature/add 1.25 scaling toggle

iop098321qwe:
- **Include Shift + Return as default for browser**
- **feat: add 1.25x monitor scaling toggle**


dhh:
Added in https://github.com/basecamp/omarchy/commit/0e335a17917897d6d23b925a66c4ea8a300f37aa

PR #4877, opened 2026-03-02
Fix try interactive selector appearing on bashrc re-source

### LINKED ISSUE #4876: Re-sourcing .bashrc triggers try interactive selector
## Description

Re-sourcing `~/.bashrc` (e.g. `source ~/.bashrc`) triggers the `try` interactive directory selector instead of silently re-initializing.

## Root Cause

In `default/bash/init`, the `try init` line:

```bash
eval "$(SHELL=/bin/bash try init ~/Work/tries)"
```

On first source, `try` resolves to the `/usr/bin/try` binary and works correctly — it outputs a shell function definition.

On subsequent sources, `try` is already defined as a **shell function** (from the first `eval`). So `try init ~/Work/tries` calls the function, which runs:

```bash
/usr/bin/env ruby '/usr/bin/try' exec --path '...' init ~/Work/tries
```

The `exec` subcommand doesn't recognize `init` as a valid sub-subcommand, so it falls through to `cmd_cd!` — the interactive selector — with `init` as the search query.

## Fix

Use `command` to bypass the shell function and call the binary directly:

```bash
eval "$(SHELL=/bin/bash command try init ~/Work/tries)"
```

## Reproduction

1. Open a terminal (bashrc sources automatically, `try` function is now defined)
2. Run `source ~/.bashrc`
3. The `try` interactive directory selector appears with `init-/home/...` pre-filled in the search box

## Reports

Also reported by umaru-sama on the Omarchy Discord (2026-03-01).

davidguttman:
## Summary

- `try init` defines a shell function named `try`. On subsequent `source ~/.bashrc`, the shell function shadows the binary, so `try init` launches the interactive selector instead of outputting the function definition. Adding `command` bypasses the function and always invokes the binary.

## Details

The one-word change (`command try` instead of `try`) is a standard bash pattern — `command` skips shell functions/aliases and runs the external binary directly. This is the same reason `command -v` is used in the guard condition above it.

The other `eval "$(... init ...)"` lines in this file (mise, starship, zoxide) don't have this problem because they don't define a shell function with the same name as the binary.

## Test plan

- [ ] Open a fresh terminal — verify `try` works normally
- [ ] Run `source ~/.bashrc` — verify no interactive selector popup
- [ ] Run `try` — verify interactive selector still works as expected

Fixes #4876
Also reported by umaru-sama on Discord (2026-03-01).

### DIFF
--- a/default/bash/init
+++ b/default/bash/init
@@ -11,7 +11,7 @@ if command -v zoxide &> /dev/null; then
 fi
 
 if command -v try &> /dev/null; then
-  eval "$(SHELL=/bin/bash try init ~/Work/tries)"
+  eval "$(SHELL=/bin/bash command try init ~/Work/tries)"
 fi
 
 if command -v fzf &> /dev/null; then


thegreyfellow:
if this ain't the cleanest PR I have ever seen :rocket:

PR #4879, opened 2026-03-02
Add window picker and bind it to Super + Ctrl + Alt + Space

jmargeta:
Adds `omarchy-cmd-window-picker` to quickly re-focus attention to an existing window.
Windows as listed as `class [workspace]: title (address)` and sorted accordingly.
image

Note, that the somewhat verbose address suffix is there to enable walker to disambiguate windows with identical classes and titles on identical workspaces.


dhh:
Interesting idea, but not something I see myself using. Maybe it could be an AUR package or a suggestion in the discussions?

PR #4880, opened 2026-03-03
fix:  Adding layout toggle to special workspaces

LeoPazEs:
This implementation uses the workspace name rather than the workspace ID. This is necessary because when a special workspace is active on a monitor, it does not appear in the output of hyprctl activeworkspace -j. Using the workspace name bypasses this limitation and provides the added benefit of supporting multiple special workspaces with custom names.


LeoPazEs:
Issue #4881

dhh:
Closing in favor of the newer special-workspace layout fix in #5099.

PR #4890, opened 2026-03-03
Fix copy-url extension clipboard on recent Chromium/Brave

mads397:
## Problem

The copy-url extension stopped working on recent Chromium and Brave versions. The extension uses `navigator.clipboard.writeText()` injected into the active page via `scripting.executeScript()`, but recent browser updates tightened the Clipboard API's user activation requirement — an extension keyboard shortcut no longer counts as a user gesture in the page context, so the write silently fails.

Additionally, the old approach fails entirely on restricted pages (`chrome://`, `brave://`, extension pages, Web Store).

## Fix

Switch to an offscreen document approach:

1. Read `tab.url` from the tabs API (no page injection needed)
2. Create a hidden offscreen document with the `CLIPBOARD` reason
3. Use `document.execCommand('copy')` in the offscreen context, backed by the `clipboardWrite` permission
4. Report success/failure back via `sendResponse` — only show "copied" notification on success
5. Close the offscreen document after use to avoid keeping an extra context alive
6. Drop the unused `scripting` permission (no longer injecting into pages)

This works on all pages regardless of their clipboard permissions or activation state.

## Changes

- `manifest.json`: Replace `scripting` with `clipboardWrite` and `offscreen` permissions, bump version to 1.1
- `background.js`: Replace `executeScript` + `navigator.clipboard` with offscreen document message passing; only notify on confirmed success; close offscreen document after copy
- `offscreen.html` / `offscreen.js`: New files — minimal offscreen document that handles clipboard writes and reports success/failure

## Testing

Tested on Brave 1.76 (Chromium 134) on Linux (Hyprland/Wayland). Confirmed working on:
- Regular web pages
- PWAs
- Pages that previously triggered clipboard permission prompts

### DIFF
--- a/default/chromium/extensions/copy-url/background.js
+++ b/default/chromium/extensions/copy-url/background.js
@@ -1,21 +1,45 @@
-chrome.commands.onCommand.addListener((command) => {
-  if (command === 'copy-url') {
-    chrome.tabs.query({ active: true, currentWindow: true }, (tabs) => {
-      const currentTab = tabs[0];
+chrome.commands.onCommand.addListener(async (command) => {
+  if (command !== 'copy-url') return;
 
-      chrome.scripting.executeScript({
-        target: { tabId: currentTab.id },
-        func: () => {
-          navigator.clipboard.writeText(window.location.href);
-        }
-      }).then(() => {
-        chrome.notifications.create({
-          type: 'basic',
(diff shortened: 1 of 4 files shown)


dhh:
What version are you having trouble with? Seems to be working here on Chromium 145.

mads397:
Thanks for checking — I'm on Brave (v145.1.87.191), not Chromium. I can't pinpoint which Brave version introduced the issue unfortunately. Submitted the PR in case it's useful to others hitting the same thing.

PR #4900, opened 2026-03-04
feat: add animation to the scratchpad

shelldandy:
adds a default animation to the scratchpad otherwise it just appears and feels a bit weird

https://github.com/user-attachments/assets/3616b2b6-4c9c-4300-af98-0fdfb28a8434


pazthor:
_LGTM!_

PR #4904, opened 2026-03-05
Add release notes tooltip to waybar update indicator

syabro:
## Summary

- Show version info and release notes on hover over the update icon in waybar
- Right-click opens full release notes in `$EDITOR`
- Left-click still runs `omarchy-update` as before
- When multiple releases are pending, shows notes for each version

## Changes

- `bin/omarchy-update-available-tooltip` — new script that fetches release notes via `gh api` and outputs JSON for waybar with Pango-formatted tooltip
- `config/waybar/config.jsonc` — updated `custom/update` module to use `return-type: json` and the new script

## Dependencies

- `gh` (GitHub CLI) — for fetching release notes via API
- `jq` — for JSON output formatting

## Screen

image

### DIFF
--- a/bin/omarchy-update-available-tooltip
+++ b/bin/omarchy-update-available-tooltip
@@ -1,0 +1,90 @@
+#!/bin/bash
+
+OMARCHY_PATH="${OMARCHY_PATH:-$HOME/.local/share/omarchy}"
+
+# Get remote tag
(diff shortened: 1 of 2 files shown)


dhh:
We've looked at this a few times in the past, and I think we're better off just leaning on the link to the latest releases when you click to update. That also accommodates for the fact that someone might be multiple versions behind.

PR #4905, opened 2026-03-05
Add omarchy-toggle-ip-mode (DHCP/static IPv4 switcher)

ElvarThorS:
## Summary
- Add `bin/omarchy-toggle-ip-mode`.
- It switches a selected `systemd-networkd` interface between DHCP and static IPv4.
- It works both interactively and with flags.
## What it does
- Lists managed interfaces and shows each with subnet + usable IP range.
- Preselects the default-route interface for quick use.
- Default toggle behavior:
  - if static override exists -> remove it and return to DHCP
  - otherwise -> ask for a static IP and create the override
- Supports non-interactive usage:
  - `--iface <iface>`
  - `--dhcp`
  - `--static <ip>`
- Remembers the last static IP per interface.
## Design choices
- Writes only a drop-in override at `/etc/systemd/network/<profile>.network.d/90-omarchy-ip-mode.conf`.
- Does not edit base `.network` files.
- Resolves the profile in this order:
  1. `networkctl` reported file
  2. common Omarchy profile names
  3. scan of standard networkd profile paths
- Uses `gum` when available, with shell prompt fallback.
- Restarts `systemd-networkd` and `systemd-resolved` after changes.
## Safety checks and failure handling
- Refuses to write if no valid `.network` profile is found.
- Validates IPv4 format, subnet match, usable host range, and rejects gateway IP as static.
- Handles systems with no default route by choosing a fallback managed interface.
- `--dhcp` is idempotent (prints a no-op message if already DHCP).
- Validates CLI flags and prints clear errors for invalid combinations/values.
## Scope
- Adds one tracked file: `bin/omarchy-toggle-ip-mode`.
- No migrations.
- No edits to existing scripts or base network profile files.
## Runtime artifacts (created on user systems)
- State file: `~/.local/state/omarchy/ip-mode/last-static-<iface>`
- Networkd override: `/etc/systemd/network/<profile>.network.d/90-omarchy-ip-mode.conf` (created/updated in static mode, removed in DHCP mode)

### DIFF
--- a/bin/omarchy-toggle-ip-mode
+++ b/bin/omarchy-toggle-ip-mode
@@ -1,0 +1,557 @@
+#!/bin/bash
+
+set -e
+
+# Toggle a systemd-networkd interface between DHCP and static IPv4.
+# Default behavior:
+# - if static override exists, remove it and return to DHCP
+# - otherwise, prompt for static IP and create an override
+#
+# Hardening included:
+# - interface selection with default-route preselection
+# - --iface / --dhcp / --static for non-interactive recovery
+# - strict guard that target .network profile exists before write
+
+STATE_DIR="$HOME/.local/state/omarchy/ip-mode"
+OVERRIDE_BASENAME="90-omarchy-ip-mode.conf"
+
+usage() {
+  cat <<'EOF'
+Usage:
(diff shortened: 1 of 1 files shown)


dhh:
Interesting idea, but not something that's common enough to warrant over 500 lines of code to the project. Thanks for the suggestion, though!

PR #4907, opened 2026-03-05
Feature/wallpaper backend switch

JakubKowaI:
### Summary
- Added option to switch wallpaper backend.
- Added swww as an backend option.
### What is does
- Gives an option in omarchy menu to switch wallpaper setup to either swaybg (original option) or swww.
- Changes autostart to the intended backend using symlinks.
### Requirements
There is a need to install swww.
### P.S.
In advance I want to say sorry if it's not up to main branch standard. It is my first PR into a repository outside of uni projects.


dhh:
Appreciate the suggestion, but I don't think there's enough value in switching the bg engine to warrant the extra code required to maintain that. Thanks for working on it, though!

PR #4911, opened 2026-03-05
Add Alt+-/= pane resize bindings to tmux

avinashjoshi:
# Summary

Tmux pane resize keybindings that mirror Hyprland's window resize convention:

- `Alt + -` / `Alt + =` for horizontal resize (left/right)
- `Alt + Shift + -` / `Alt + Shift + =` for vertical resize (up/down)


dhh:
We already have ctrl+shift+alt+arrows for resizing. Prefer to just stick with one default set for this.

PR #4918, opened 2026-03-06
Add battery identifier to Waybar config

danflood94:
Just a quick fix for the default waybar profile in omarchy, this wouldn't be a problem is building your own config but hey, Waybar crashes with a `std::runtime_error (core dumped)` when a Logitec  G Pro dongle is unplugged. This happens because Waybar attempts to watch the `hidpp_battery udev` events, which disappear instantly upon removal.

```
Discharging
Discharging
Discharging
Discharging
Discharging
Discharging
Discharging
terminate called after throwing an instance of 'std::runtime_error'
what():  Could not watch events for /sys/class/power_supply/hidpp_battery_6
Aborted  (core dumped) waybar -l trace 
```

Set `"bat": "BAT0"` in the battery module of waybar. This ensures it only tracks the primary system battery and ignores HID power supplies.

### DIFF
--- a/config/waybar/config.jsonc
+++ b/config/waybar/config.jsonc
@@ -80,6 +80,7 @@
     "on-click": "omarchy-launch-wifi"
   },
   "battery": {
+    "bat": "BAT0",
     "format": "{capacity}% {icon}",
     "format-discharging": "{icon}",
     "format-charging": "{icon}",


lewispb:
Seeing this same crash on a desktop (9950x, no system battery — only a `hidpp_battery_1` from a Logitech G502 wireless mouse). Stack traces match: `terminate` → `abort` after an uncaught `std::runtime_error` when the HID device sleeps/disconnects. ~19 SIGABRT coredumps over the last month on our end, and it looks like #5234 is reporting the same thing.

A few questions on the fix:

1. Is `BAT0` guaranteed across hardware? Some laptops expose the primary battery as `BAT1` (ThinkPads) — would pinning to `BAT0` silently hide the indicator for those users?
2. For desktops with no system battery at all, should the `battery` module be dropped from `modules-right` entirely rather than pinned to a non-existent `BAT0`? Or is the empty render harmless enough that one default covers both cases?
3. Does this need a migration under `migrations/` to patch existing users' `~/.config/waybar/config.jsonc`? As written, the PR only updates the template, so existing users would still crash until they run `omarchy-refresh-waybar` (which loses customizations). Precedent seems to be `1775208227.sh` (sed-patches the waybar config + `omarchy-restart-waybar`).

Happy to test alternatives if useful.

dhh:
This is removing the battery indicator entirely on some laptops, like the new Framework 13 Pro. Reverting.

lewispb:
I've tested https://github.com/Alexays/Waybar/pull/4982 which solves this issue properly. So we can just update to the latest version of Waybar instead, unreleased yet though.

PR #4926, opened 2026-03-08
feat: add ctrl n and ctrl p for walker

xheisenbugx:
# Why?
To have the option to use ctrl + n for next and ctrl + p for previous, this is something common in other programs and it is useful for people that use small keyboards without arrow keys

### DIFF
--- a/config/walker/config.toml
+++ b/config/walker/config.toml
@@ -9,6 +9,8 @@ hide_action_hints = true                                                    # gl
 
 [keybinds]
 quick_activate = []
+next = ["Down", "ctrl n"]
+previous = ["Up", "ctrl p"]
 
 [columns]
 symbols = 1 # providers to be queried by default


Daan5556:
I think it would be nice to add Vim keybinds here also.

```
next = ["ctrl j"]
previous = ["ctrl k"]
```

S1M0N38:
> I think it would be nice to add Vim keybinds here also.
> 
> ```
> next = ["ctrl j"]
> previous = ["ctrl k"]
> ```

ctrl-p and ctrl-n are the default Vim keybinds when navigating selector-like UI. ctrl-p / ctrl-n for selection navigation of lists are widely use in TUI and GUI apps as well.

---

I've test the propose config on a brand new installation of Omarchy 3.8.1 (Macbook Air 2014) and it worked. lgtm.

jscorbalan (commented):
Just a heads up: Ctrl+p and Ctrl+n are already part of Walker’s default keybindings according to the official docs.
I also wanted to use Ctrl+p / Ctrl+n because they feel very natural, but I tested it and:

Ctrl+p is currently used to pin/unpin an application
Ctrl+n moves the pinned application up

It seems Ctrl+j and Ctrl+k don’t have these conflicts.

[Walker keybindings documentation](https://walkerlauncher.com/docs/keybindings?utm_source=chatgpt.com)

dhh:

PR #4930, opened 2026-03-08
Add desktop app install and remove scripts

AlexZeitler:
Analog to `omarchy-webapp-install`, allows installing local executables as desktop apps with icon and terminal support. Adds corresponding remove script and menu entries for both install and remove.

### DIFF
--- a/bin/omarchy-desktop-app-install
+++ b/bin/omarchy-desktop-app-install
@@ -1,0 +1,68 @@
+#!/bin/bash
+
+set -e
+
+ICON_DIR="$HOME/.local/share/applications/icons"
+
+if (( $# < 3 )); then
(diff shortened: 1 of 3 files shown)


dhh:
What apps are you using for this that aren't TUIs?

AlexZeitler:
For example starting Chromium with another profile:

`chromium --profile-directory="some-profile"`

Or any Desktop app that doesn't create a `.desktop` entry itself.

When installing it using the `TUI` option, the launcher starts a terminal window first which closes after the Desktop app itself gets started - it works but it doesn't look nice.

dhh:
Ah, I see. Appreciate the work on this, but I don't think that's a common enough case that we need a full infrastructure for it. Totally fine to just manually create the .desktop file in that case 👌

PR #4937, opened 2026-03-08
Use OMARCHY_PATH in omarchy-migrate

pomartel:
I was wondering what was happening with my migrations! :)


pomartel:
I ended up using the default Omarchy path for my system because there's just a lot of places where the path is hardcoded.

PR #4938, opened 2026-03-08
Add ThinkPad mic mute LED sync following XPS hardware handler pattern

annoyedmilk:
`swayosd --input-volume mute-toggle` is silently broken for all users on swayosd 0.3.0 (current Arch package) due to a `pulsectl-rs` regression: basecamp/omarchy#4831, ErikReider/SwayOSD#210, ErikReider/SwayOSD#213, ErikReider/SwayOSD#225.

Fix uses `wpctl` (native PipeWire, already on every Arch system) for the actual toggle, swayosd only for OSD display.

Bonus: syncs mic mute LED on laptops that have one (ThinkPad etc.) via a `-w` guard.

No migration needed, the binding in `media.conf` is sourced automatically.

### DIFF
--- a/bin/omarchy-cmd-mic-mute
+++ b/bin/omarchy-cmd-mic-mute
@@ -1,9 +1,11 @@
 #!/bin/bash
 
-# Toggle microphone mute. Dell XPS systems get special handling for the hardware LED.
+# Toggle microphone mute. Dell XPS and ThinkPad systems get special handling for the hardware LED.
 
 if omarchy-hw-match "XPS"; then
   omarchy-cmd-mic-mute-xps
+elif omarchy-hw-match "ThinkPad"; then
+  omarchy-cmd-mic-mute-thinkpad
 else
   swayosd-client --monitor "$(omarchy-hyprland-monitor-focused)" --input-volume mute-toggle
 fi
--- a/bin/omarchy-cmd-mic-mute-thinkpad
+++ b/bin/omarchy-cmd-mic-mute-thinkpad
@@ -1,0 +1,23 @@
+#!/bin/bash
+
+# Toggle microphone mute on ThinkPad systems. Uses wpctl for reliable toggling
+# and syncs the platform::micmute LED via brightnessctl.
+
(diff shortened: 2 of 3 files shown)


default/hypr/bindings/media.conf
brianspragge, line 1: Tested change on Thinkpad X13 Gen4, does not fix mute button.  Had to revert change AND roll back to swayosd 0.2.1 to fix mutebutton.  mutebutton light is still broke(always on)

annoyedmilk: Hey Brian, did you also add the omarchy-cmd-mic-mute script in the bin folder?

brianspragge: *now* I have.  It works.  Do you plan on commiting a change to fix the led light on the mute button?
Thanks for your work and time.  Hopefully I haven't 'annoyed' you more milk.

annoyedmilk:
I forgot to include a udev rule I had locally for the LED. Updated the script in e7fe9c3 to use brightnessctl instead, it should work without any udev rule.

default/hypr/bindings/media.conf
annoyedmilk: I've updated the script in e7fe9c3 to use brightnessctl instead, so the LED should now work without needing my udev rule which i forgot to add.

brianspragge: I just updated to it.  LED works!  Great work!

dhh:
swayosd 3.1.1 is out. What's still relevant here?

annoyedmilk:
Yes still relevant:
Tested on swayosd 0.3.1: --input-volume mute-toggle still silently no-ops (PipeWire + pulseaudio-shim, T14s). Exit 0, OSD shows, but wpctl get-volume @DEFAULT_AUDIO_SOURCE@ is unchanged. Output-side mute-toggle and input raise/lower work fine. So the dispatcher in 3.5.1 isn't just about LED sync, it's also the reason the mic-mute key actually mutes the mic. This PR extends that same pattern to ThinkPads (LED via brightnessctl platform::micmute, omarchy-hw-match).

PR #4939, opened 2026-03-08
Restore custom ratio when toggling using "1-Window Ratio" shortcut

plasticdonut:
Makes the 1-Window Ratio aware of existing custom 1 window aspect ratios. Defaults 1:1 ratio if no custom ratio has been set.

For example I currently have this in my  `~/.config/hypr/looknfeel.conf`. Using  "1-Window Ratio"  currently forces 1:1 ratio and my custom 16:9 is lost. With this change it instead toggles between 16:19 and full width.
```
layout {
    # Avoid overly wide single-window layouts on wide screens
    single_window_aspect_ratio = 16 9
}
```


dhh:
No longer needed with the new persistent toggle.

Issue #953, opened 2025-08-21
capslock not working?


anshul12ra:
i can't open the capslock when i click in the button noting happen. is it issue in my device or  hyperland or omarchy.

macarmo:
Check the manual for the default key bindings https://manuals.omamix.org/2/the-omarchy-manual/53/hotkeys

Caps lock has other purpose than the one you might be expecting.

PR #4941, opened 2026-03-09
Fix hyprlock PAM config to prevent freeze on wake

alansikora:
## Problem

Arch's default `/etc/pam.d/hyprlock` only has an `auth` section:

```
auth        include     login
```

The missing `account` and `session` sections cause hyprlock to crash or freeze after PAM authentication completes. This manifests as the lock screen being visible but completely unresponsive to input — no cursor blinking, no keyboard input, no trackpad — requiring a hard reboot. This commonly occurs after waking from suspend.

Documented in [hyprlock #953](https://github.com/hyprwm/hyprlock/issues/953) (Issue 6). Omarchy does not currently manage this file.

## Fix

Ships a complete PAM config at `default/pam.d/hyprlock` with all required sections:

```
#%PAM-1.0

auth      include   system-login
account   include   system-login
session   include   system-login
```

A migration applies it to existing installs, only overwriting if the `account` section is missing (won't clobber custom PAM configs).

## Changes

| File | Action |
|------|--------|
| `default/pam.d/hyprlock` | New default PAM config |
| `migrations/1773019289.sh` | Applies fix to existing installs |

## References

- https://github.com/hyprwm/hyprlock/issues/953 — Comprehensive hyprlock crash investigation, Issue 6
- https://github.com/basecamp/omarchy/issues/4184 — Suspend not working (related: wake freeze after successful suspend)

## Related

This pairs with #4940 (unmount FUSE before suspend) to make the full suspend/wake cycle reliable. This PR should merge first — fixing suspend entry is pointless if wake is broken.

### DIFF
--- a/default/pam.d/hyprlock
+++ b/default/pam.d/hyprlock
@@ -1,0 +1,5 @@
+#%PAM-1.0
+
+auth      include   system-login
+account   include   system-login
+session   include   system-login
--- a/migrations/1773019289.sh
+++ b/migrations/1773019289.sh
@@ -1,0 +1,5 @@
+echo "Fix hyprlock PAM config (missing account/session sections cause freeze on wake)"
+
+if [[ -f /etc/pam.d/hyprlock ]] && ! grep -q "^account" /etc/pam.d/hyprlock; then
+  sudo cp -p "$OMARCHY_PATH/default/pam.d/hyprlock" /etc/pam.d/hyprlock
+fi
--- a/migrations/1773019289.sh
+++ b/migrations/1773019289.sh
@@ -1,5 +1,5 @@
 echo "Fix hyprlock PAM config (missing account/session sections cause freeze on wake)"
 
-if [[ -f /etc/pam.d/hyprlock ]] && ! grep -q "^account" /etc/pam.d/hyprlock; then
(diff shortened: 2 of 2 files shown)


alansikora:
Testing these changes locally on a Framework 13 AMD (Ryzen, s2idle) before marking ready for review.

dhh:
I think we need a complete theory of why this is necessary. Hyprlock does not freeze consistently when this file isn't present. So why is it happening sometimes? Are we sure this is actually the missing bit?

alansikora:
I'm honestly still not sure @dhh, I'm running a few tests with and without it to pinpoint the issue.

alansikora:
You're right @dhh — this isn't needed.

While debugging the freeze it seemed to help, but after further testing and isolating variables, the PAM config change is unnecessary. Hyprlock only uses the auth PAM section (it never calls pam_acct_mgmt() or pam_open_session()), so the added account and session sections have no effect. Upstream intentionally ships auth-only.

The actual fix is #4940 — FUSE daemons were silently blocking suspend, and the failed suspend is what caused the intermittent freeze on "wake." With the FUSE fix in place and the PAM config reverted to upstream default, 5+ suspend/wake cycles produced zero freezes.

Closing this one.

PR #4943, opened 2026-03-09
Add keybind to move window out of scratchpad

arizkhalid:
There was no keybind to eject a window from the scratchpad back to the current workspace. 
Added `Super + Shift + S` to remove active window from scratchpad, logical complement to `Super + Alt + S` (send to scratchpad).


dhh:
We already have `Super + Shift + [1-9]` to move out of the special workspace to any other workspace. I think it's better to focus on that. Appreciate the suggestion, though ✌️

PR #4952, opened 2026-03-09
add more alias

joaoexe:
Add `update` alias

Adds an alias to simplify and speed up system updates from the terminal.


pomartel:
Omarchy handles packages update with the `omarchy-update-system-pkgs` which runs when you update Omarchy. There's no need for an alias.

PR #4960, opened 2026-03-10
Add netbird-tui TUI application

n0pashkov:
## Summary

- Adds [netbird-tui](https://github.com/n0pashkov/netbird-tui) — a terminal UI for managing NetBird (WireGuard-based mesh VPN)
- Installs via AUR as `netbird-tui`
- Registered in `tuis.sh` with float window mode, runs as `sudo netbird-tui`
- NetBird icon (256x256 PNG) added to `applications/icons/`

## What is netbird-tui

A Bubble Tea TUI that talks to the local NetBird daemon via gRPC. Tabs: Status, Peers, Routes, Forwarding, Settings. Supports connect/disconnect with confirmation, logout, debug bundle creation, and auto-refresh every 5 seconds.

## Test plan

- [ ] Fresh omarchy install: verify `netbird-tui` is installed via `yay`
- [ ] Verify launcher entry appears with NetBird icon
- [ ] Launch via omarchy launcher, confirm float window opens
- [ ] Confirm `sudo netbird-tui` runs correctly with NetBird daemon active


n0pashkov:
Thanks for the thorough review! Addressed the actionable points:

- **Migration added** (`migrations/1773149325.sh`): installs `netbird` + `netbird-tui`, copies the icon, and calls `omarchy-tui-install` so existing installs get the launcher entry on next `omarchy update`
- **`netbird` daemon added** to `omarchy-base.packages` alongside `netbird-tui` — both new installs and the migration install it

On the `sudo` and supply-chain concerns: the daemon socket requires root access (same situation as `lazydocker` with Docker socket), so `sudo` is currently necessary. Happy to revisit if there's a preferred pattern in omarchy for this. The AUR pinning concern applies equally to other TUI packages in the list — open to guidance on whether omarchy has a policy for this.

n0pashkov:
Update: removed `netbird` as a hard dependency from both `omarchy-base.packages` and the migration. It's now listed as `optdepends` in the AUR PKGBUILD instead — this avoids conflicts for users who have netbird installed outside of pacman (official binary, etc.). The TUI itself doesn't need netbird at build time, only at runtime.

n0pashkov:
On the latest review:

**Shebang**: `omarchy-migrate` runs migrations via `bash $file` (see `bin/omarchy-migrate`), so a shebang isn't needed — consistent with all other migrations in the repo.

**"Two sources of truth"**: this is the established omarchy pattern. For example, `bluetui` (migration `1762417304.sh`) has `omarchy-pkg-add bluetui` in its migration *and* `bluetui` in `omarchy-base.packages`. The migration handles existing installs on `omarchy update`; `base.packages` handles fresh installs. Both are needed.

dhh:
Interesting idea, but I wouldn't use NetBird, so not a good fit for a default install.

n0pashkov:
> Interesting idea, but I wouldn't use NetBird, so not a good fit for a default install.

ok. thx. Can it be added to optional software like tailscale?

PR #4961, opened 2026-03-10
Fix battery remaining time displaying minutes as hours

marko-builds:
upower reports time in different units depending on battery state — it can output X hours or X minutes. The script assumed the value was always in hours, so when the battery was low enough for upower to switch to minutes, it would display e.g. 25h instead of 25m.
This fix reads the unit field from upower's output and formats accordingly.


dhh:
Ha, hit the same issue and had it fixed on dev 👌

PR #4963, opened 2026-03-10
Exit screensaver on mouse movement

michaelaguiar:
I noticed a previous PR to get the screensaver to close on mouse movement that didn't work out because of issues with Bluetooth mice.

Here is a different approach that compares mouse position which I think should do the trick!


dhh:
Doesn't work here using a logitech ms 4x with the 2.4ghz dongle.

Sergiogd112:
You may have to look into using the raw input from `/dev/input/mice`, `/dev/input/event0` or `printf '\e[?1003h\e[?1006h'`

dhh:

PR #4966, opened 2026-03-10
Move update log to persistent path so it survives reboot

### LINKED ISSUE #4965: Update log is written to /tmp and lost on reboot
## Problem

`omarchy-update-perform` writes the update log to `/tmp/omarchy-update.log`. When the update includes a kernel upgrade, `omarchy-update-restart` triggers a reboot, which wipes `/tmp`. The log is lost before the user can review it.

## Suggestion

Write the log to `~/.local/state/omarchy/update.log` instead. That directory already exists and is used for migration state and toggles. It survives reboots and follows XDG conventions.

The change is one line in `bin/omarchy-update-perform`:

```bash
# before
exec > >(tee "/tmp/omarchy-update.log") 2>&1

# after
exec > >(tee "$HOME/.local/state/omarchy/update.log") 2>&1
```

amanat361:
## Summary

- Move update log from `/tmp/omarchy-update.log` to `~/.local/state/omarchy/update.log` so it persists across reboots
- Updated both `omarchy-update-perform` (writes the log) and `omarchy-update-analyze-logs` (reads the log)
- Follows existing convention — `~/.local/state/omarchy/` is already used for migrations, toggles, and other persistent state

## Test plan

- [ ] Run `omarchy-update-perform` and verify log is written to `~/.local/state/omarchy/update.log`
- [ ] Reboot and confirm the log file is still present
- [ ] Verify `omarchy-update-analyze-logs` reads from the correct path

Fixes #4965

### DIFF
--- a/bin/omarchy-update-analyze-logs
+++ b/bin/omarchy-update-analyze-logs
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-update_log="/tmp/omarchy-update.log"
+update_log="$HOME/.local/state/omarchy/update.log"
 
 # Check for initramfs generation failure
 if grep -q "Updating linux initcpios" "$update_log"; then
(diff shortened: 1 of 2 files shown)


dhh:
Closing as an exact duplicate of #5058, which remains open.

PR #4968, opened 2026-03-10
Add speed test to Setup menu

amanat361:
## Summary

- Add a "Speed Test" option to the Setup menu, placed next to Wifi
- Installs `speedtest-cli` on demand if not already present (no new base dependency)
- Runs the standard Ookla-backed `speedtest-cli` in a floating terminal

## Test plan

- [ ] Open Omarchy menu > Setup > Speed Test
- [ ] Verify it prompts to install `speedtest-cli` on first run if not installed
- [ ] Verify speed test runs and displays results
- [ ] Verify it works on subsequent runs without re-installing

### DIFF
--- a/bin/omarchy-menu
+++ b/bin/omarchy-menu
@@ -219,7 +219,7 @@ show_font_menu() {
 }
 
 show_setup_menu() {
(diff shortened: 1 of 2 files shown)


dhh:
Interesting suggestion, but I don't think there's enough value here over just visiting fast.com or speedtest.net in a browser.

PR #4971, opened 2026-03-10
Add Tmux section to learn menu

dl-alexandre:
Adds a tmux section to the learn menu, opening https://tmuxcheatsheet.com/ for quick reference.

### DIFF
--- a/bin/omarchy-menu
+++ b/bin/omarchy-menu
@@ -83,13 +83,14 @@ aur_install_and_launch() {
 }
 
 show_learn_menu() {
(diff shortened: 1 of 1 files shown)


dhh:
Not a good fit given how much tailoring we've done to the Tmux setup. I do like the idea of a learning resource, but it would have to be based off the tmux setup as it exists in Omarchy. https://learn.omacom.io/2/the-omarchy-manual/53/hotkeys#tmux

dl-alexandre:
Thank you, Also found that `<leader>-?` gets most the info as well.

PR #4975, opened 2026-03-11
Fix missing sudo for btrfs map-swapfile in hibernation setup

### LINKED ISSUE #4974: Enable hibernation bug: resume.conf missing the resume_offset
### System details

Omarchy 3.4.2

### What's wrong?

After enabling hibernation via the menu, my laptop blue screen when waking up.

Upon investigation I remember seeing a "Permission Denied" error while enabling hibernation in omarchy.

I am sorry, but I don't have the logs about it anymore.
However, after I noticed this

```bash
cat /etc/limine-entry-tool.d/resume.conf
KERNEL_CMDLINE[default]+=" resume=/dev/mapper/root resume_offset="
```

Notice that the `resume_offset=` is missing a value.

So I tried to follow what the enable script does on

https://github.com/basecamp/omarchy/blob/2acc2ebfea41255d2cbe6e1204025235dfb166a9/bin/omarchy-hibernation-setup#L76

Suspiciously, this line `RESUME_OFFSET=$(btrfs inspect-internal map-swapfile -r "$SWAP_FILE")` returns "Permission Denied" when not running as `root`.

So I believe the setup silently failed.

I've tried to fix it and run 

```bash
sudo limine-mkinitcpio
sudo limine-update
```
I am still blue screening, but one thing is sure: resume_offset wasn't set properly before.

---

EDIT, the `resume.conf` doesn't seem applied

`cat /proc/cmdline`
```
quiet splash cryptdevice=PARTUUID=7cd32dfd-7a98-4743-a74b-386a4abed4cb:root root=/dev/mapper/root zswap.enabled=0 rootflags=subvol=@ rw rootfstype=btrfs
```
---
EDIT, `/proc/cmdline` didn't show resume because `/etc/default/limine` have `KERNEL_CMDLINE[default]="cry..` instead of `KERNEL_CMDLINE[default]+="cry...`.

Once fixed, and running `limine-mkinitcpio` again, the `resume=` show in `/proc/cmdline`.

However I still blue screen.

Related to https://github.com/basecamp/omarchy/issues/4158

Daedalus-Icarus:
## Summary
- Fix missing `sudo` for `btrfs inspect-internal map-swapfile` in `omarchy-hibernation-setup`, which caused `resume_offset` to be empty in `resume.conf` due to "Permission Denied"
## Details
The `btrfs inspect-internal map-swapfile -r` command on line 76 requires root privileges to read the swap file's physical offset on btrfs. Without `sudo`, the command silently fails, producing a broken config:
KERNEL_CMDLINE[default]+=" resume=/dev/mapper/root resume_offset="

This results in a blue screen when waking from hibernation. Every other `btrfs` call in the script already uses `sudo` — this was simply an oversight.
Fixes #4974

### DIFF
--- a/bin/omarchy-hibernation-setup
+++ b/bin/omarchy-hibernation-setup
@@ -73,7 +73,7 @@ if [[ ! -f $RESUME_DROP_IN ]]; then
   echo "Adding resume kernel parameters"
   sudo swapon -p 0 "$SWAP_FILE" 2>/dev/null
   RESUME_DEVICE=$(findmnt -no SOURCE -T "$SWAP_FILE" | sed 's/\[.*\]//')
-  RESUME_OFFSET=$(btrfs inspect-internal map-swapfile -r "$SWAP_FILE")
+  RESUME_OFFSET=$(sudo btrfs inspect-internal map-swapfile -r "$SWAP_FILE")
   sudo mkdir -p /etc/limine-entry-tool.d
   echo "KERNEL_CMDLINE[default]+=\" resume=$RESUME_DEVICE resume_offset=$RESUME_OFFSET\"" | sudo tee "$RESUME_DROP_IN" >/dev/null
 fi
--- a/bin/omarchy-hibernation-setup
+++ b/bin/omarchy-hibernation-setup
@@ -75,7 +75,13 @@ if [[ ! -f $RESUME_DROP_IN ]]; then
   RESUME_DEVICE=$(findmnt -no SOURCE -T "$SWAP_FILE" | sed 's/\[.*\]//')
   RESUME_OFFSET=$(sudo btrfs inspect-internal map-swapfile -r "$SWAP_FILE")
   sudo mkdir -p /etc/limine-entry-tool.d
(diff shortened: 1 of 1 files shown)


Desel72:
Hi @dhh 
Could you please review this PR?

NicolasDorier:
There is another issue `/etc/limine-entry-tool.d/resume.conf` is ignored because `/etc/default/limine` is overriding with `KERNEL_CMDLINE[default]=` isntead of `KERNEL_CMDLINE[default]+=`. (See other people talking about it on https://github.com/basecamp/omarchy/issues/4259#issuecomment-3751288516)

You can confirm this after running `limine-mkinitcpio`, then `/boot/limine.conf` doesn't have the resume.

My hibernation still not working, but this is definitely part of the issue.

Desel72:
@nicholaschiang 
You are absolutely right. Please create new issue. I will fix them and send PR.

NicolasDorier:
@Desel72 btw, would you tell me the recipe for a french "Banana flambe" ?

NicolasDorier:
Update this PR, it doesn't make sense to split those two things in two PR. This PR doesn't address fully my issue as is.

Desel72:
Hi @NicolasDorier @dhh 
I've updated this PR. Can you review this?

pomartel:
I think it might be fixed already: https://github.com/basecamp/omarchy/commit/4c5e6569c41fe6cfd173a53370bee29bddcc18dd

PR #4980, opened 2026-03-11
fix 4973 night light cursor

### LINKED ISSUE #4973: Night light does not apply to cursor
### System details

Omarchy 3.4.2

### What's wrong?

When the night light is turned on via `super + ctrl + n`, the cursor is still black and white, not with the night light color filter.

matzalazar:
### Description
Fixes an issue (#4973) where the night light (gamma LUTs applied via `hyprsunset` / `super + ctrl + n`) would not affect the cursor because it was defaulting to a hardware cursor.

### Changes Made
- **Hyprland Configuration:** Added `no_hardware_cursors = true` to `default/hypr/looknfeel.conf` inside the `cursor { ... }` block to force the cursor to be rendered via software. This ensures new installations get the fix automatically.
- **Migration Script:** Added a new script under `migrations/` to safely inject `no_hardware_cursors = true` into the `cursor` block of `~/.config/hypr/looknfeel.conf` for existing users, ensuring backward compatibility.

### Verification
- Tested locally by running the migration script.
- Verified that `looknfeel.conf` is correctly patched.
- Toggling the night light now correctly applies the color temperature change to the cursor as well as the rest of the screen.

Closes #4973

### DIFF
--- a/default/hypr/looknfeel.conf
+++ b/default/hypr/looknfeel.conf
@@ -129,6 +129,7 @@ misc {
 
 # https://wiki.hypr.land/Configuring/Variables/#cursor
 cursor {
+    no_hardware_cursors = true
     hide_on_key_press = true
     warp_on_change_workspace = 1
 }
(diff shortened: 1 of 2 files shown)


matzalazar:
Keeping the script simple. The edge case of an explicit = false is out of scope for this migration.

dhh:

PR #4985, opened 2026-03-11
Add Thorium to list of valid 'omarchy-launch-webapp' browsers that work with '--app'

rholak:
Thorium works with --app, similar to other Chromium-based browsers. This PR adds thorium to the omarchy-launch-webapp script so a separate Chromium browser does not need to be launched if the user is already using Thorium.


rholak:
Updated to also add Thorium to the hyprland browsers.conf's window rules for completeness

PR #5004, opened 2026-03-14
waybar/workspaces: remove margin from button to fix cursor flicker

emincangencer:
The margin created gaps between workspace buttons where there was no clickable element, causing the cursor to revert to default when moving between workspaces. Increased padding from 6px to 7px to maintain visual spacing while fixing the issue.


dhh:

PR #5005, opened 2026-03-14
Enable GPU in voxtype if Vulkan is available

sgruendel:
voxtype now supports GPU acceleration on [Vulkan](https://github.com/peteonrails/voxtype?tab=readme-ov-file#gpu-acceleration)


migrations/1772990935.sh
```diff
@@ -1,7 +1,16 @@
-echo "Add sample low battery notification hook"
```
dhh, line 1: Seems like you took over an exciting migration name?

sgruendel: > Seems like you took over an exciting migration name?

uh yeah, looks like `omarchy-dev-add-migration` generates the same migration timestamp multiple times, I'm on stable channel / master, so as long as there is no new commit it will generate the same migration file name ...

Fixed now.

migrations/1773506226.sh
```diff
@@ -0,0 +1,16 @@
+echo "Enable GPU in voxtype if Vulkan is available"
+
+if omarchy-cmd-present voxtype; then
+  if compgen -G "/usr/share/vulkan/icd.d/*.json" >/dev/null; then
+    echo "Vulkan is available, enabling GPU in voxtype"
+    voxtype setup gpu --enable || true
```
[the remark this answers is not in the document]
sgruendel: should be ok, if switching to GPU fails we're still on CPU

peteonrails:
@sgruendel Thank you for doing this!

sgruendel:
> @sgruendel Thank you for doing this!

Thanks @peteonrails for that awesome project, I still remember VoiceType on OS/2 Warp 4 early 90s, we've come a long way ...

PR #5012, opened 2026-03-14
feat: add natural scroll-to-zoom bindings

berrimi:
Adds SUPER+CTRL+Scroll for incrementing cursor zoom and SUPER+CTRL+Scroll Down to reset. Also reduces scroll_event_delay for more responsive zooming in Hyprland.


dhh:

PR #5023, opened 2026-03-15
fix: adds "Open Directory" to the xdg-desktop-portal-gtk floating-window tag windowrule

FredrikMWold:
Adds Open Directory dialogs to the floating-window Hyprland rule so those file chooser windows are treated the same as existing Open/Save dialogs. This fixes cases where xdg-desktop-portal-gtk and similar app dialogs would not open as centered floating windows.

example output from hyprctl clients

```json
{
    "address": "0x5566bc419320",
    "mapped": true,
    "hidden": false,
    "at": [1288, 36],
    "size": [1266, 1398],
    "workspace": {
        "id": 2,
        "name": "2"
    },
    "floating": false,
    "monitor": 1,
    "class": "xdg-desktop-portal-gtk",
    "title": "Open Directory",
    "initialClass": "xdg-desktop-portal-gtk",
    "initialTitle": "Open Directory",
    "pid": 2005,
    "xwayland": false,
    "pinned": false,
    "fullscreen": 0,
    "fullscreenClient": 0,
    "overFullscreen": false,
    "grouped": [],
    "tags": ["default-opacity*"],
    "swallowing": "0x0",
    "focusHistoryID": 2,
    "inhibitingIdle": false,
    "xdgTag": "",
    "xdgDescription": "",
    "contentType": "none",
    "stableId": "18000112"
}

```

### DIFF
--- a/default/hypr/apps/system.conf
+++ b/default/hypr/apps/system.conf
@@ -4,7 +4,7 @@ windowrule = center on, match:tag floating-window
 windowrule = size 875 600, match:tag floating-window
 
 windowrule = tag +floating-window, match:class (org.omarchy.bluetui|org.omarchy.impala|org.omarchy.wiremix|org.omarchy.btop|org.omarchy.terminal|org.omarchy.bash|org.gnome.NautilusPreviewer|org.gnome.Evince|com.gabm.satty|Omarchy|About|TUI.float|imv|mpv)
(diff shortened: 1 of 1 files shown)


eds-4d5:
Confirmed this exact gap on Omarchy 4.0.2 / Hyprland 0.56.2: the stock rule in `/usr/share/omarchy/default/hypr/apps/system.lua` does not match the `Open Directory` title (regex only covers `Open.*Files?|Open [Ff]older.*|Save.*...`).

Workaround until this lands upstream — add to `~/.config/hypr/hyprland.lua` (loaded after the defaults):

```lua
o.window(
  { class = "xdg-desktop-portal-gtk", title = "^Open Directory" },
  { tag = "+floating-window" }
)
```

Then `hyprctl reload`.

Verified: the default regex does NOT match `Open Directory` while this rule does, and Hyprland reloads cleanly with no Lua errors.

Full write-up (PT-BR): https://github.com/eds-4d5/omarchy-quick-fixes

PR #5024, opened 2026-03-15
when installing Bitwarden, ask user if it should be used as SSH agent

sgruendel:
Bitwarden can also be used as [SSH agent](https://bitwarden.com/help/ssh-agent/)


timlau:
```
echo "env = SSH_AUTH_SOCK,/home/$USER/.bitwarden-ssh-agent.sock" | tee -a "$HOME/.config/uwsm/env" >/dev/null
```
Do not work for me (running fish shell)
```
echo "export SSH_AUTH_SOCK=/home/$USER/.bitwarden-ssh-agent.sock" | tee -a "$HOME/.config/uwsm/env" >/dev/null
```

Does work in both bash and fish

timlau:
Here is a more safe version there can be run multiple times and only do needed actions

```bash
#!/bin/bash

env_file=$HOME/.config/uwsm/env
packages=("bitwarden" "bitwarden-cli")
use_bitwarden_as_ssh_agent=false

echo ${packages[@]}
if omarchy-pkg-missing "${packages[@]}"; then
  echo "Installing Bitwarden..."
  omarchy-pkg-add "${packages[@]}"
else
  echo "Bitwarden already installed"
fi

if ! grep -Fq "SSH_AUTH_SOCK" $env_file; then
  # ask user if they want to use Bitwarden as their SSH agent and configure SSH_AUTH_SOCK accordingly
  if gum confirm "Use Bitwarden as SSH agent?"; then
    use_bitwarden_as_ssh_agent=true
    echo "export SSH_AUTH_SOCK=/home/$USER/.bitwarden-ssh-agent.sock" | tee -a $env_file &>/dev/null
    echo "Make sure to enable SSH agent support in Bitwarden settings."
  fi
fi

setsid gtk-launch bitwarden &>/dev/null

if [[ $use_bitwarden_as_ssh_agent == "true" ]]; then
  touch ~/.local/state/omarchy/reboot-required
  omarchy-update-restart
fi

```

sgruendel:
> ```
> echo "env = SSH_AUTH_SOCK,/home/$USER/.bitwarden-ssh-agent.sock" | tee -a "$HOME/.config/uwsm/env" >/dev/null
> ```
> 
> Do not work for me (running fish shell)

Good catch, I think it was a leftover from the previous approach of updating the env file.  Fixed now to use proper export syntax.

timlau:
it is a good idea to redirect  the output from launching bitwarden into null, it outputs a lot of debug info to the terminal.

The check for `SSH_AUTH_SOCK` exist in my scrtipt is also a good idea, so you dont add it multiple times, if a user run it multiple times. I ended up with adding it many times in my testing.

`setsid gtk-launch bitwarden &>/dev/null`

sgruendel:
> it is a good idea to redirect the output from launching bitwarden into null, it outputs a lot of debug info to the terminal.

changed it accordingly

> The check for `SSH_AUTH_SOCK` exist in my scrtipt is also a good idea, so you dont add it multiple times, if a user run it multiple times. I ended up with adding it many times in my testing.

Good idea, also user might already have set this to use e.g. `ssh-agent`, no need to ask then.

bin/omarchy-install-bitwarden
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+echo "Installing Bitwarden..."
+omarchy-pkg-add bitwarden bitwarden-cli
+
+env_file=$HOME/.config/uwsm/env
```
dhh, line 6: Don't think we need to set it this high up in the stack. I think using a regular bashrc env should be fine, no?

sgruendel: hm had probably copied it over from my `ssh-agent` implementation where it was needed because that agent is started via `systemd`. Want me to change it to `.bashrc`?

sgruendel: Turns out this is not working in `.bashrc` e.g. for VSCode (started via Walker), so probably same for all apps not started via interactive shell. Better keeping it in `.config/uwsm/env` then.

sgruendel:
superseded by #6719 for Quattro

PR #5033, opened 2026-03-16
chore : Properly set timeout for screen lock to 5 minutes

itlogsandwich:
Timeout for screen lock was not properly set to 5 minutes. Was set as 151 seconds, 2.51 mins

### DIFF
--- a/config/hypr/hypridle.conf
+++ b/config/hypr/hypridle.conf
@@ -11,7 +11,7 @@ listener {
 }
 
 listener {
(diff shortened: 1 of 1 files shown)


pomartel:
I think we'll need to add a comment to hypridle.conf so people stop submitting the same pull request again and again! https://github.com/basecamp/omarchy/pull/4814 https://github.com/basecamp/omarchy/pull/4414

Issue #3589, opened 2025-11-24
upgrade to 3.2.0 screensaver becomes unresponsive


med502:
### System details

AMD Ryzen9 3950, GTX980TI, Omarchy 3.2.0 

### What's wrong?

After upgrade to 3.2.0 screensaver after short time becomes unresponsive, seems that around time when auto lock screen is being triggered. Once it hangs no keyboard shortcut works, only way out to use reset button.

dhh:
Closing because Quattro replaces the affected Hyprlock or Hypridle implementation with native Omarchy session-lock and idle services. Please open a new issue if this reproduces on Quattro.

Issue #2991, opened 2025-10-30
Can not interact with lock screen after opening lid


geass94:
### System details

ROG Strix G533ZW_G533ZW

### What's wrong?

<img width="521" height="624" alt="Image" src="https://github.com/user-attachments/assets/5732f258-cacc-4335-bb32-f51087d605e2" />

After closing and re-opening lid, lock screen appears, it splashes once, still can see lock screen, but laptop keyboard and mouse are not working, can't type my password in. Tho I can switch to tty using ctrl+alt+3 and reboot the laptop, but that completely looses the stuff I left open to continue work on.
Any fix on that would help a lot.

Note: Fastfetch is captured on external monitor with lid closed.

manuel1618:
I have a similar bug - on startup I cannot use the USB connected keyboard for login. The other keyboard, which then turned on and connected via usb dongle works.

dhh:
Closing because Quattro replaces the affected Hyprlock or Hypridle implementation with native Omarchy session-lock and idle services. Please open a new issue if this reproduces on Quattro.

Issue #2635, opened 2025-10-20
Omarchy 3.1 No Signal to Monitors When Waking from Sleep


ochowie:
### System details

Omarchy 3.1 Nvidia 3070

### What's wrong?

Since the upgrade to Omarchy 3.1 my system is not waking my monitors after locking. The system is still running but neither of my two monitors is recognizing input.

davidszp:
I have this with 3.0.2, so this may not be limited to 3.1. For me it happens randomly, I would say 1 out of 10 wake-up from sleep events.

Symptom: Drives and fans spin up. Monitors do not wake up from sleep. If I bring up a console using ctrl+alt+f3 only THEN monitors turn on. Then I can go back to the main console ctrl+alt+f1 and login as usual.

Edit: Nov 27: This issue has disappeared for me, but I do not know why.

ochowie:
I am getting this pretty much every time with 3.1. Also, when this happens to me I am not able to boot into a TTY terminal.

Tharun-tharun:
same here, I've been experiencing the same issue for the past month and i'm unable to figure out the precise fix

defer:
I think this happened after the sddm migration.

I can easily reproduce it by reproducing what hypridle ends up doing. **Connect via ssh**, and:

Basic setup:
```
$ export HYPRLAND_INSTANCE_SIGNATURE=$(hyprctl instances | awk -F'[ :]' '/^instance / {print $2}' 
```

Turn the screens off:
```
hyprctl dispatch dpms off
```

Important: wait until the monitor is actually on standby, it takes ~10 secs in my case

Turn the screens on:
```
hyprctl dispatch dpms on
```

This won't work. Also navigating from/to the TTY (ctrl alt f#) won't work.

Now, thiis brings it back:
```
systemctl restart sddm
```

ochowie:
I can confirm restarting sddm worked to bring the screens back up. It also completely resets the hyprland session so it's not really feasible to include in the wake from sleep command.

fatihprofil:
Same problem keep happening on our office setup, TTY doesnt work too.

masq6r:
I have very similar symptom. In my case, I shut down the monitor and after a while I turn it on, then a black screen is what I get. I saw a lot of people complaining hyprlock and dpms, but I disabled hyprlock already and no luck.

masq6r:
[...]

PR #5034, opened 2026-03-16
Fix Hyprland freeze on suspend/resume with NVIDIA GPUs

### LINKED ISSUE #4891: Suspend/resume causes compositor freeze on NVIDIA hybrid GPU laptops
# Issue Draft: Suspend/resume freeze on NVIDIA hybrid GPU laptops

**Repo:** https://github.com/basecamp/omarchy

---

## Title

Suspend/resume causes compositor freeze on NVIDIA hybrid GPU laptops (nvidia-utils freeze workaround)

### Summary

After suspend/resume, Hyprland and Wayland clients (waybar, hyprlock) freeze or crash. The screen shows a stale frame (e.g. the lock screen) but no input is accepted. Mouse cursor may still move. The only recovery is killing Hyprland or a hard reboot.

### Root Cause

`nvidia-utils` (590.48.01) installs a systemd drop-in that disables user session freezing during suspend:

```
# /usr/lib/systemd/system/systemd-suspend.service.d/10-nvidia-no-freeze-session.conf
[Service]
Environment="SYSTEMD_SLEEP_FREEZE_USER_SESSIONS=false"
```

systemd itself warns about this on every suspend:
```
User sessions remain unfrozen on explicit request ($SYSTEMD_SLEEP_FREEZE_USER_SESSIONS=0).
This is not recommended, and might result in unexpected behavior, particularly
in suspend-then-hibernate operations or setups with encrypted home directories.
```

Because user sessions are not frozen, Wayland clients continue running while the GPU is being suspended by `nvidia-suspend.service`. When the GPU disappears mid-operation, clients crash (waybar SIGABRT) or the compositor hangs with a stale framebuffer.

### Reproduction

1. NVIDIA hybrid GPU laptop (Intel + NVIDIA), `nvidia-open-dkms` installed
2. Lock screen or just close lid (triggers suspend via logind)
3. Wait a few seconds, then resume (open lid / press key)
4. Lock screen appears but keyboard input is dead. Mouse cursor may still move.
5. Only recovery: kill Hyprland from TTY (Ctrl+Alt+F3) or hard reboot

Reproduced multiple times, 100% consistent.

### Attempted Fixes

- **Overriding the drop-in** with `SYSTEMD_SLEEP_FREEZE_USER_SESSIONS=true`: Did not resolve the issue (compositor still froze on resume, possibly because NVIDIA driver itself doesn't handle frozen sessions well).
- **Killing hyprlock/walker after resume**: Doesn't help because the compositor itself is frozen, not just the lock screen.

### Current Workaround

Disable suspend on lid close entirely:
```bash
# /etc/systemd/logind.conf.d/lid-lock.conf
[Login]
HandleLidSwitch=lock
HandleLidSwitchExternalPower=lock
```

This locks the screen on lid close without suspending. Users can still hibernate manually (which does a full shutdown/restore and avoids the GPU state issue).

### Environment

- **Omarchy:** v3.4.1
- **Hyprland:** 0.54.0
- **Aquamarine:** 0.10.0
- **Kernel:** 6.18.13-arch1-1
- **NVIDIA:** nvidia-open-dkms 590.48.01-4
- **GPU:** Intel Arc (Meteor Lake) + NVIDIA RTX 4080 Max-Q
- **Laptop:** ASUS (ROG/Zephyrus style, dual GPU with HDMI on NVIDIA)

### Relevant Logs

**Suspend entry:**
```
systemd-logind: The system will suspend now!
systemd-sleep: User sessions remain unfrozen on explicit request
kernel: PM: suspend entry (s2idle)
```

**Resume + crash:**
```
kernel: PM: suspend exit
systemd-coredump: Process (waybar) of user 1000 dumped core. [SIGABRT]
```

**Hyprland crash report (when restarting after freeze):**
```
terminate called after throwing an instance of 'std::runtime_error'
  what(): CBackend::create() failed!
aquamarine: Session is not active, waiting for 5s
aquamarine: Session timeout reached
aquamarine: Session could not be activated in time
aquamarine: DRM Backend failed
```

### Suggestion

This seems like a systemic issue with `nvidia-utils` disabling session freezing. Possible approaches:

1. **Omarchy-level workaround:** Add a post-resume hook that restarts the compositor gracefully if it detects a frozen state
2. **Document the issue** for NVIDIA hybrid GPU users with a recommended logind override
3. **Upstream to NVIDIA/Hyprland:** The `FREEZE_USER_SESSIONS=false` workaround may have been necessary for older drivers but seems to cause more problems than it solves with current nvidia-open-dkms + Wayland compositors
4. **Investigate hibernate as alternative:** `systemctl hibernate` does a full power cycle and should avoid the GPU state issue entirely — could be the recommended sleep mode for NVIDIA laptops

### Related

- systemd warning: https://github.com/systemd/systemd/issues/31292
- NVIDIA Wayland suspend issues are widely reported in the Hyprland community

amlucas0xff:
## Summary

- Adds systemd services that pause Hyprland (SIGSTOP) before NVIDIA GPU suspend and resume it (SIGCONT) after, preventing the race condition that causes compositor freeze and hyprlock SIGSEGV crashes
- Includes hardware install script (`fix-nvidia-suspend.sh`) for new installations and a migration for existing NVIDIA setups
- Services are only installed on systems with NVIDIA GPUs detected via `lspci`

## Root Cause

`nvidia-utils` installs a systemd drop-in (`10-nvidia-no-freeze-session.conf`) that sets `SYSTEMD_SLEEP_FREEZE_USER_SESSIONS=false`. This keeps Wayland clients (hyprlock, Hyprland, waybar) running while the GPU is being suspended by `nvidia-suspend.service`. When the GPU disappears mid-operation:

- hyprlock crashes with SIGSEGV in `std::thread::join()` during cleanup
- Hyprland freezes with a stale framebuffer, requiring a hard reboot
- Affects both hybrid GPU laptops and multi-GPU desktops

## Approach

Based on the community-proven pattern from [0xFMD/hyprland-suspend-fix](https://github.com/0xFMD/hyprland-suspend-fix), adapted to Omarchy conventions:

- `omarchy-hyprland-suspend.service` runs **before** `nvidia-suspend.service`, sending SIGSTOP to Hyprland
- `omarchy-hyprland-resume.service` runs **after** `nvidia-resume.service`, sending SIGCONT to Hyprland
- Follows the same systemd service pattern used by `omarchy-nvme-suspend-fix.service`

## Files Changed

- `install/config/hardware/fix-nvidia-suspend.sh` — new hardware script, auto-detects NVIDIA
- `install/config/all.sh` — registers the script after `nvidia.sh`
- `migrations/1773664282.sh` — applies the fix to existing installations

## Test Plan

- [ ] Verify services install only on NVIDIA systems (`lspci | grep -i nvidia`)
- [ ] Confirm `systemctl is-enabled omarchy-hyprland-suspend.service` after migration
- [ ] Suspend/resume cycle: lock screen renders, input works, no SIGSEGV in `coredumpctl`
- [ ] Non-NVIDIA systems: migration exits cleanly without creating services

Fixes #4891
Related: #3589, #2991, #2635

### DIFF
--- a/install/config/all.sh
+++ b/install/config/all.sh
@@ -34,6 +34,7 @@ run_logged $OMARCHY_INSTALL/config/hardware/printer.sh
 run_logged $OMARCHY_INSTALL/config/hardware/usb-autosuspend.sh
 run_logged $OMARCHY_INSTALL/config/hardware/ignore-power-button.sh
 run_logged $OMARCHY_INSTALL/config/hardware/nvidia.sh
+run_logged $OMARCHY_INSTALL/config/hardware/fix-nvidia-suspend.sh
 run_logged $OMARCHY_INSTALL/config/hardware/intel.sh
 run_logged $OMARCHY_INSTALL/config/hardware/intel-thermald.sh
 run_logged $OMARCHY_INSTALL/config/hardware/vulkan.sh
--- a/install/config/hardware/fix-nvidia-suspend.sh
+++ b/install/config/hardware/fix-nvidia-suspend.sh
@@ -1,0 +1,50 @@
+# Fix Hyprland freeze on suspend/resume with NVIDIA GPUs
+# nvidia-utils disables session freezing during suspend, which causes Wayland
+# clients to keep running while the GPU suspends. This creates a race condition
+# that crashes hyprlock (SIGSEGV) and freezes the compositor.
+# Solution: Pause Hyprland with SIGSTOP before NVIDIA suspends, resume with SIGCONT after.
+# See: https://github.com/basecamp/omarchy/issues/4891
+# See: https://github.com/0xFMD/hyprland-suspend-fix
+
+NVIDIA="$(lspci | grep -i 'nvidia')"
+
+if [ -n "$NVIDIA" ]; then
+  echo "Installing NVIDIA suspend fix for Hyprland..."
+
+  cat <<EOF | sudo tee /etc/systemd/system/omarchy-hyprland-suspend.service >/dev/null
+[Unit]
+Description=Pause Hyprland before NVIDIA GPU suspend
+Before=systemd-suspend.service
+Before=systemd-hibernate.service
+Before=nvidia-suspend.service
+Before=nvidia-hibernate.service
+
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/killall -STOP Hyprland
+
+[Install]
+WantedBy=systemd-suspend.service
+WantedBy=systemd-hibernate.service
+EOF
+
+  cat <<EOF | sudo tee /etc/systemd/system/omarchy-hyprland-resume.service >/dev/null
+[Unit]
+Description=Resume Hyprland after NVIDIA GPU resume
+After=systemd-suspend.service
+After=systemd-hibernate.service
+After=nvidia-resume.service
+
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/killall -CONT Hyprland
+
+[Install]
+WantedBy=systemd-suspend.service
+WantedBy=systemd-hibernate.service
+EOF
+
+  sudo systemctl daemon-reload
+  chrootable_systemctl_enable omarchy-hyprland-suspend.service
+  chrootable_systemctl_enable omarchy-hyprland-resume.service
+fi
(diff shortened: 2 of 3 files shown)


k1ng440:
Hi, I have tested this and does not solve the problem

dhh:

PR #5036, opened 2026-03-16
Add TUI calendar to waybar clock

n0pashkov:
## Summary

- Right-clicking the clock in waybar opens a minimal TUI calendar in a floating terminal
- Navigate months with `h`/`←` (previous) and `l`/`→` (next), quit with `q`
- No dependencies — uses the standard `cal` command

## Changes

- **`bin/omarchy-tui-calendar`** — shell script rendering a centered, navigable monthly calendar in the terminal
- **`bin/omarchy-launch-calendar`** — launcher using `setsid uwsm-app` + `xdg-terminal-exec` with app-id `org.omarchy.calendar`
- **`config/waybar/config.jsonc`** — `on-click-right` on the clock module set to `omarchy-launch-calendar`
- **`default/hypr/apps/system.conf`** — window rules for `org.omarchy.calendar`: float, center, fixed size 220×212 (kept separate from the `floating-window` tag to avoid inheriting the 875×600 size rule)

### DIFF
--- a/bin/omarchy-launch-calendar
+++ b/bin/omarchy-launch-calendar
@@ -1,0 +1,2 @@
+#!/bin/bash
+exec setsid uwsm-app -- xdg-terminal-exec --app-id=org.omarchy.calendar -e omarchy-tui-calendar
--- a/bin/omarchy-tui-calendar
+++ b/bin/omarchy-tui-calendar
@@ -1,0 +1,52 @@
+#!/bin/bash
+
+year=$(date +%Y)
+month=$(date +%-m)
+
+show_calendar() {
+  printf '\033[2J\033[H'
+  local cols
+  cols=$(tput cols)
+  local pad=$(( (cols - 20) / 2 ))
+  local indent
+  indent=$(printf "%${pad}s" "")
+  while IFS= read -r line; do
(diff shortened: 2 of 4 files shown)


matiasarenhard:
Definitely a good topic. I tried to install external libraries to handle this, and it was a nightmare.

Could you please share a screenshot of the result of your code @n0pashkov ? 

Is it possible to make this the default in the Omarchy installation @dhh ?

I’ve been using Omarchy for about a month, and it has been a great experience!

Thanks guys!

n0pashkov:
> Definitely a good topic. I tried to install external libraries to handle this, and it was a nightmare.
> 
> Could you please share a screenshot of the result of your code @n0pashkov ?
> 
> Is it possible to make this the default in the Omarchy installation @dhh ?
> 
> I’ve been using Omarchy for about a month, and it has been a great experience!
> 
> Thanks guys!

image

znxr:
I even made my own version, but I would love to see this functionality integrated into omarchy by default.

dhh:

dhh:
Actually, I would suggest turning this into a omarchy bar plugin for the new Quickshell implementation. Then it's easy for anyone to add this who wants it.

PR #5037, opened 2026-03-16
Feature - Flat Menu Search

marko-builds:
## What

  Adds two quality-of-life improvements to the Walker launcher:

  1. **Flat menu search** — type `<` in Walker to get a searchable
  flat list of every Omarchy menu action (~100 entries). No more
  navigating sub-menus step by step. Each entry shows its breadcrumb
  path as a subtitle (e.g. `Install → Development`).

  2. **Ctrl+J/K navigation** — vim-style alternatives to arrow keys
  for moving through any Walker list.

  ## How it works

  A new Elephant Lua module
  (`default/elephant/omarchy_menu_search.lua`) statically enumerates
  all actionable items from `omarchy-menu`, mirroring every `case`
  branch. Terminal-based actions wrap with
  `omarchy-launch-floating-terminal-with-presentation` to preserve the
   existing presentation style.

  The `<` prefix was chosen because `>` is already claimed by the
  system-wide walker config (`/etc/xdg/walker/config.toml`) for the
  runner provider.

  ## Files changed

  - `default/elephant/omarchy_menu_search.lua` — new Elephant module
  - `config/walker/config.toml` — Ctrl+J/K keybinds + `<` prefix
  - `install/config/walker-elephant.sh` — symlink for new module on
  fresh install
  - `bin/omarchy-refresh-walker` — re-establish all elephant symlinks
  on config refresh (covers existing users)

  ## Usage

  Open Walker → type `<` → full action list appears → keep typing to
  filter → Enter to run.
  
   ▎ Existing users: The migration script runs automatically on next
  omarchy-update — it creates the elephant symlink and patches
  ~/.config/walker/config.toml in place.

### DIFF
--- a/bin/omarchy-refresh-walker
+++ b/bin/omarchy-refresh-walker
@@ -17,5 +17,11 @@ omarchy-refresh-config walker/config.toml
 omarchy-refresh-config elephant/calc.toml
 omarchy-refresh-config elephant/desktopapplications.toml
 
+# Re-establish elephant menu module symlinks
+mkdir -p ~/.config/elephant/menus
+ln -snf $OMARCHY_PATH/default/elephant/omarchy_themes.lua ~/.config/elephant/menus/omarchy_themes.lua
+ln -snf $OMARCHY_PATH/default/elephant/omarchy_background_selector.lua ~/.config/elephant/menus/omarchy_background_selector.lua
+ln -snf $OMARCHY_PATH/default/elephant/omarchy_menu_search.lua ~/.config/elephant/menus/omarchy_menu_search.lua
+
 # Restart service
 omarchy-restart-walker
--- a/config/walker/config.toml
+++ b/config/walker/config.toml
@@ -9,6 +9,8 @@ hide_action_hints = true                                                    # gl
 
 [keybinds]
 quick_activate = []
+next = ["Down", "ctrl j"]
+previous = ["Up", "ctrl k"]
 
 [columns]
 symbols = 1 # providers to be queried by default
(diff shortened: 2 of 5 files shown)


amanat361:
Question: I think the Omarchy menu already has an option that says "apps" which then pulls up the app launcher, so if we implemented a flat menu like you suggest would it also let you search for apps within the Omarchy menu or are only settings part of the flat list?

Slightly unrelated but this reminds me of a thought I had of merging the app launcher menu (super + space) with the Omarchy menu.

I feel like turning it into one thing (à la raycast) would be really nice. In raycast (or spotlight) and dare I say even the abysmal Windows start menu, you can search for settings, app, and even do some calculations in one place.

I love this idea of a flat menu and am wondering if it's hard to just combine it all into one thing entirely.

abenz1267:
> Question: I think the Omarchy menu already has an option that says "apps" which then pulls up the app launcher, so if we implemented a flat menu like you suggest would it also let you search for apps within the Omarchy menu or are only settings part of the flat list?
> 
> Slightly unrelated but this reminds me of a thought I had of merging the app launcher menu (super + space) with the Omarchy menu.
> 
> I feel like turning it into one thing (à la raycast) would be really nice. In raycast (or spotlight) and dare I say even the abysmal Windows start menu, you can search for settings, app, and even do some calculations in one place.
> 
> I love this idea of a flat menu and am wondering if it's hard to just combine it all into one thing entirely.

It's not hard and it's what walker is meant to do by default. Omarchy just had to implement it.

Idzuo32:
My approach was to keep Super+Space as apps-only, but with a prefix trigger (<) to switch into settings/actions search within the same launcher. This way you get a single entry point without mixing apps and settings by default — the separation is still there, just accessible from one place instead of two bindings.
On merging both menus fully (à la raycast) — I'd keep that as a separate discussion for DHH to weigh in on. I think the two-menu split is a deliberate decision.

dhh:

PR #5038, opened 2026-03-16
Set critical notification timeout to 30s instead of infinite

lucletoffe:
## Problem

The current `default-timeout=0` for `[urgency=critical]` in mako's core.ini means critical notifications never auto-dismiss. Combined with mako's known ghost surface bug ([emersion/mako#622](https://github.com/emersion/mako/issues/622)), this can leave phantom notifications permanently stuck on screen — surviving even `makoctl dismiss --all`.

Any script or application that sends `urgency=critical` (even with an explicit `-t 5000` timeout) gets overridden to infinite by this config, which is surprising behavior.

## Fix

Change `default-timeout=0` to `default-timeout=30000` (30 seconds). This is long enough to notice any critical alert while preventing notifications from becoming permanent screen fixtures.

## Context

- mako upstream bug: https://github.com/emersion/mako/issues/622
- The ghost surface persists in the Wayland layer-shell even after mako's internal dismiss, requiring a full `makoctl reload` or daemon restart to clear.

### DIFF
--- a/default/mako/core.ini
+++ b/default/mako/core.ini
@@ -17,7 +17,7 @@ invisible=true
 invisible=false
 
 [urgency=critical]
-default-timeout=0
+default-timeout=30000
 layer=overlay
 
 [summary~="Setup Wi-Fi"]


amanat361:
Bump. Love this idea since sometimes I get notifs that don't auto dismiss and I always wonder why, even when they don't seem "critical" at all.

That said, if it doesn't already exist, might be nice if omarchy had a notifications.conf file as a sibling to the other user-scoped conf files that let you easily tweak mako. I think having a conf file that lets you override defaults (i.e. same as bindings.conf for example), would be preferable to most users.

Definitely like this idea though.

patrick-boc:
> Any script or application that sends `urgency=critical` (even with an explicit `-t 5000` timeout) gets overridden to infinite by this config, which is surprising behavior.

I cannot reproduce this behavior. The explicit timeout does work on my machine.

Notification disappears after one second:
```notify-send "test" -u critical -t 1000```

Notification persists:
```notify-send "test" -u critical```

patrick-boc:
> This is long enough to notice any critical alert while preventing notifications from becoming permanent screen fixtures.

Are we throwing the baby out with the bathwater? This will increase the chance of the user missing a critical notification just to mitigate a seemingly rare upstream bug. Alternative mitigations are: 1) Silence notifications. 2) Restart mako.

Since there is no notification center, we need to be more careful about notification timeouts. Stepping away from the screen for 30 seconds, should not result in completely missing a critical notification.

patrick-boc:
> sometimes I get notifs that don't auto dismiss and I always wonder why, even when they don't seem "critical" at all

Can you provide some examples? Are these notification coming from Omarchy or apps? Omarchy has only a few notifications set to critical.

> That said, if it doesn't already exist, might be nice if omarchy had a notifications.conf file as a sibling to the other user-scoped conf files that let you easily tweak mako. 

Yes, this is a known issue: https://github.com/basecamp/omarchy/pull/3830

dhh:

PR #5040, opened 2026-03-16
Replace Impala with nettui for network controls

skibidiandulka:
This PR replaces Impala with nettui as Omarchy's network TUI.

Why

Omarchy currently routes its top-bar network action to a Wi-Fi-only TUI.
`nettui` keeps the same iwd-first approach, but extends it into a single network TUI for both Wi-Fi and Ethernet.

This project is heavily inspired by Impala and intentionally stays close to Impala's interaction model and backend philosophy.
It is not meant to replace that model with something foreign. If anything, nettui is much closer to an Impala-derived network TUI than to a clean-sheet rewrite: the same general Wi-Fi workflow, extended with integrated Ethernet support and Omarchy-specific polish.

What changed

- replaced `impala` with `nettui` in `install/omarchy-base.packages`
- updated `omarchy-launch-wifi` to launch `nettui`
- renamed the Omarchy UI entry from `Wifi` to `Network`
- renamed the keybinding label from `Wifi controls` to `Network controls`
- added a dedicated Hyprland floating size for `org.omarchy.nettui`
- enabled `systemd-networkd.service` in the network setup path
- added a migration that installs `nettui` and removes `impala`
- documented the single network click path in `docs/network-click-behavior.md`

Why this is a better fit for Omarchy

- one TUI for Wi-Fi and Ethernet
- better match for Omarchy's single network entry point
- stays very close to the Impala-style iwd workflow
- avoids splitting network management across separate tools

Important note

This PR requires `nettui` to be published in `pkgs.omarchy.org` before merge.
Without that package source, replacing `impala` in the Omarchy installer would break the official install path.

Nettui is available at: https://github.com/skibidiandulka/nettui/

### DIFF
--- a/bin/omarchy-launch-wifi
+++ b/bin/omarchy-launch-wifi
@@ -1,7 +1,14 @@
 #!/bin/bash
 
-# Launch the Omarchy wifi controls (provided by the Impala TUI).
-# Attempts to unblock the wifi service first in case it should be been blocked.
+# Launch the Omarchy network controls through nettui.
+#
+# Omarchy exposes a single network entry point for both Wi-Fi and Ethernet.
+# Unblock Wi-Fi first in case the radio was soft-blocked.
+
+if ! command -v nettui >/dev/null 2>&1; then
+  notify-send "󰤮    nettui Not Installed" "Install the nettui package to open Omarchy network controls." -u critical -t 5000
+  exit 1
+fi
 
 rfkill unblock wifi
-omarchy-launch-or-focus-tui impala
+exec omarchy-launch-or-focus-tui nettui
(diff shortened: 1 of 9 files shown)


skibidiandulka:
@codex review

skibidiandulka:
@codex review

skibidiandulka:
I am not sure if this is too big of a change. But it should be ready to be used. This package should be even sturdier than impala. I believe that it would iprove the UX again a bit. For example when you want to do certain root things in impala it just tells you that impala needs to be running via root compared to nettui where you get a pop-up to fill in the sudo password. And the backend it has better multi-adapter settings. Again- I understand that it might be too big of a change but I think that nettui is truly better suited for OMARCHY.

PedroBizachi:
@dhh please give it a try, really better than impala!

dhh:

PR #5041, opened 2026-03-16
Add delayed hibernate after suspend

alansikora:
## Summary

Brings back automatic hibernate after suspend — close the lid, get instant s2idle for quick wake-ups, then the system automatically hibernates after 90 minutes for zero battery drain.

This was previously removed (e4b73726) because systemd's native `suspend-then-hibernate` caused wake-up loops on multiple laptops. The issue is that transitioning to hibernate from a suspended state hangs on several GPU drivers (confirmed on amdgpu, reported on others). Direct `systemctl hibernate` from an awake state works fine everywhere.

The fix: an RTC alarm wakes the system fully after 90 minutes of s2idle, then a transient systemd timer triggers a direct hibernate from that clean, awake state.

## How it works

Single sleep hook (`system-sleep/delayed-hibernate`):

1. **On suspend** — sets an RTC alarm for 90 minutes
2. **On wake** — checks if the kernel cleared the alarm (RTC fired) or if the user woke manually
3. **If RTC** — schedules `systemctl hibernate` via `systemd-run --on-active=5s` (waits for the suspend operation to fully complete before hibernating)
4. **If user** — cancels the pending alarm, normal instant wake

## Migration

- Only applies to systems with hibernation already configured
- Cleans up old `lid.conf` / `hibernate.conf` if present from the previous suspend-then-hibernate setup

## Test plan

- [x] Direct `systemctl hibernate` works (session restored)
- [x] RTC alarm wakes system after delay
- [x] Transient timer triggers hibernate on RTC wake
- [x] User wake before timeout cancels the alarm (instant s2idle wake)
- [x] Tested on Framework 13 AMD (Ryzen 7040, s2idle, kernel 6.19.6)
- [ ] Test on other hardware


alansikora:
Superseded by #5051 which adds a sleep mode picker UI on top of this. I originally wanted to keep this as a smaller, standalone PR for easier review, but GitHub doesn't support stacking PRs from forks — so #5051 includes all of these changes plus the UI.

dhh:
Closing because this was explicitly superseded by #5051. Thanks for the contribution!

PR #5055, opened 2026-03-18
Add imv keybind to copy current image to clipboard

vitorpacheco:
Add a Ctrl+c binding in imv config to copy the currently open image to clipboard via wl-copy


vitorpacheco:
@codex address that feedback

PR #5057, opened 2026-03-18
Add missing hyprland.conf to all themes

### LINKED ISSUE #5039: Most themes missing hyprland.conf, causing globbing error on every theme except Kanagawa
## Problem

`default/hypr/hyprland.conf` line 13 sources a theme-specific config:

```
source = ~/.config/omarchy/current/theme/hyprland.conf
```

But only 1 of 17 bundled themes (`kanagawa`) actually ships a `hyprland.conf`. Selecting any other theme produces:

```
Config error in file /home/jaded/.config/hypr/hyprland.conf at line 13: source= globbing error: found no match
```

This has come up before (#4132, #2971, #1116) but was closed without addressing the root cause.

## Why it happens

The `kanagawa` theme uses `hyprland.conf` for a theme-specific opacity override:

```
# Kanagawa backdrop is too strong for detault opacity
windowrule = opacity 0.98 0.95, match:tag terminal
```

This is the intended use — themes that need Hyprland overrides can provide them. But themes that don't need overrides still need an empty file present, since Hyprland's `source =` directive errors when the target doesn't exist (there's no optional sourcing).

## Suggested fix

Add an empty `hyprland.conf` to the 15 themes that are missing one:

```
catppuccin, catppuccin-latte, ethereal, everforest, flexoki-light,
gruvbox, hackerman, matte-black, miasma, nord, osaka-jade,
ristretto, rose-pine, tokyo-night, vantablack, white
```

One-liner:
```bash
for theme in themes/*/; do [ ! -f "$theme/hyprland.conf" ] && touch "$theme/hyprland.conf"; done
```

ryan-mt:
`config/hypr/hyprland.conf` line 13 sources a theme-specific config:

```
source = ~/.config/omarchy/current/theme/hyprland.conf
```

Only `kanagawa` ships this file (for opacity overrides). The other 16 themes don't have one, causing Hyprland to log a globbing error on every startup:

```
Config error in file ~/.config/hypr/hyprland.conf at line 13: source= globbing error: found no match
```

This adds empty `hyprland.conf` placeholder files to all 16 themes that are missing one. The empty files satisfy the `source =` directive without changing any behavior — themes that need Hyprland overrides (like kanagawa) already have content in theirs.

Fixes #5039


dhh:
The file is generated from default/themed/hyprland.conf.tpl. Only explicitly needed when there's an override required.

PR #5060, opened 2026-03-18
Update hypridle loginctl comment

paddyohanlon:


pomartel:
See https://github.com/basecamp/omarchy/pull/5033

paddyohanlon:
Ah, sorry. I only searched open PRs. I'll make sure to search all in future.

I don't quite understand. I guess the comment is actually accurate somehow, or there's something else I'm missing.

pomartel:
It's kind of a bug with the screensaver that resets the timer. Since the screensaver starts after 2.5 minutes and restarts the timer, then the lock screen runs after another 2.5 minutes, so 5 minutes total. The comment should probably be updated to explain this because we receive a pull request regarding that comment every other day. It's not just you!

paddyohanlon:
> It's kind of a bug with the screensaver that resets the timer. Since the screensaver starts after 2.5 minutes and restarts the timer, then the lock screen runs after another 2.5 minutes, so 5 minutes total. The comment should probably be updated to explain this because we receive a pull request regarding that comment every other day. It's not just you!

Got it. Thanks for explaining!

PR #5061, opened 2026-03-18
Make omarchy menu and cpu buttons "infinitely wide" for easier clicking

axelfontaine:
This is a simple quality of life improvement that pads the omarchy menu button to the top left corner of the screen and the cpu widget to the top right corner of the screen (with no unclickable gaps). This makes it possible to yank the mouse to the desired corner and click to get the desired action (previously one had to aim carefully as you would otherwise have the pointer in an 8px gap).

The layout is unchanged.

See https://en.wikipedia.org/wiki/Fitts%27s_law#Implications_for_UI_design

### DIFF
--- a/config/waybar/style.css
+++ b/config/waybar/style.css
@@ -15,8 +15,16 @@
   margin-left: 0;
 }
 
+#custom-omarchy {
+  padding-left: 8px;
+}
+
 .modules-right {
-  margin-right: 8px;
+  margin-right: 0;
+}
+
+#cpu {
+  padding-right: 8px;
 }
 
 #workspaces button {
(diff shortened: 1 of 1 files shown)


pomartel:
Can this be done so it doesn't change the visual alignment of the button? Does the waybar support padding instead of margin?

axelfontaine:
@pomartel That can be achieved while preserving this effect by adding 8px of left padding directly to the omarchy button.

pomartel:
I'm not the one merging the PR but you have better chances of having this accepted if it maintains visual consistency.

axelfontaine:
Fair point. Updated to keep the layout unchanged and added the same functionality for the cpu widget in the top right corner.

dhh:
We are moving to quickshell for Omarchy 4, so will need to make any adjustments there when it lands.

PR #5065, opened 2026-03-19
fix: correct 8 shell script bugs

ryan-mt:
## Summary

Systematic audit found and fixed 8 confirmed bugs across shell scripts in `bin/` and `install/`:

- **omarchy-drive-set-password**: missing `$` in herestring — multi-drive selection was permanently unreachable
- **omarchy-reinstall-configs**: `$(bash ...)` captured stdout and tried to execute it as a command
- **omarchy-ai-skill.sh**: `ln -s` without `-f` fails when symlink already exists (retry install / re-run migration)
- **omarchy-install-docker-dbs**: `main_menu` function not in scope when run via `present_terminal`
- **hypridle.conf**: comment said `# 5min` for a 151-second (≈2.5min) timeout
- **limine-snapper.sh**: BRE regex `^/+` matches literal `/+`, not boot entries like `/Omarchy` — installation verification always failed for limine users
- **omarchy-theme-set-keyboard-asus-rog**: missing file guard for `keyboard.rgb` (only 1/17 themes has it)
- **omarchy-menu (Emacs)**: `systemctl --user enable` raced against async `install()` — service never enabled

## Approach

Each fix was:
1. Verified as a real bug (not a style preference)
2. Syntax-checked with `bash -n` immediately after editing
3. Code-reviewed by a separate agent
4. Confirmed to not change behavior in working code paths

## Test plan

- [x] `bash -n` passes on all 8 modified files
- [x] Code review confirms all fixes are correct and safe
- [x] No unintended behavioral changes — each fix only affects the broken path
- [ ] Manual testing on Omarchy system (limine boot entry verification, ASUS ROG keyboard, Emacs install, multi-drive encryption, config reinstall, Docker DB cancel, AI skill re-link)

### DIFF
--- a/bin/omarchy-drive-set-password
+++ b/bin/omarchy-drive-set-password
@@ -5,7 +5,7 @@
 encrypted_drives=$(blkid -t TYPE=crypto_LUKS -o device)
 
 if [[ -n $encrypted_drives ]]; then
-  if (( $(wc -l <<<encrypted_drives) == 1 )); then
+  if (( $(wc -l <<<"$encrypted_drives") == 1 )); then
     drive_to_change="$encrypted_drives"
   else
     drive_to_change="$(omarchy-drive-select "$encrypted_drives")"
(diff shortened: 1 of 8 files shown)


dhh:
Closing as outdated: this mixed eight-fix bundle spans mostly removed or renamed pre-Quattro systems. Any surviving bugs should be submitted as focused PRs against the current files.

PR #5066, opened 2026-03-19
feat(hypridle): inhibit suspend during active VoIP streams

Rockeyxx:
### Description
Resolves an issue where `hypridle` suspends the system during active VoIP calls in clients that lack proper Wayland `zwp_idle_inhibit_manager_v1` protocol implementations (e.g., Discord via Electron, Waydroid).

### Implementation Architecture
- Added `omarchy-idle-call-check` to `./bin/` to parse the PipeWire node graph via `pactl`.
- The script evaluates state by returning `0` if it detects an active `audio/source` (microphone polling) or an `audio/sink` strictly flagged with `media.role = "communication"` or `"phone"`. `1` otherwise.
- Updated `./config/hypr/hypridle.conf`


Rockeyxx:
**Update:** I force-pushed an architectural refinement to the capture stream evaluation. 

Initially, I used `pactl list source-outputs`, but discovered during testing that LXC bridges (specifically Waydroid's Android Audio HAL) bypass the `State: RUNNING` broadcast in the `pipewire-pulse` compatibility layer, which masks the active microphone stream. 

To resolve this edge case, I pivoted the source-polling logic to query the native PipeWire daemon directly via `wpctl status`. This reliably catches standard applications, Electron WebRTC clients, and LXC containers uniformly. The playback evaluation remains on `pactl` to leverage the strict `media.role` metadata filtering.

PR #5074, opened 2026-03-19
fix: make notification-toggle feedback silent

### LINKED ISSUE #5073: Notifications are disabled but still makes a sound
### System details

Omarchy 3.4.2, Framework Laptop 13

### What's wrong?

Not sure if this can even be solved at the OS level, but I noticed that after disabling notifications, even though the visual notifications stopped, the sound kept coming.

I'm talking specifically about Telegram (web app) and can't tell if something else would still trigger a sound, but I would assume web apps would all behave the same.

Is there anything in the background that could be caught as a notification sound coming from the browser and prevented from sounding at all, when notifcations are silenced?

saschabuehrle:
## Summary
Fixes #5073 by preventing the toggle confirmation notification from playing an alert sound.

## Problem
Even with notifications silenced, toggling notification mode still triggered an audible sound because the confirmation used plain `notify-send`.

## Fix
- add `-h string:suppress-sound:true` to both toggle confirmation notifications in `bin/omarchy-toggle-notification-silencing`
- keep existing visual confirmation unchanged

## Validation
- verified script syntax with `bash -n`
- confirmed change is minimal and isolated to the toggle helper

Greetings, saschabuehrle


patrick-boc:
As @pmpinto noted in the linked issue, this PR completely misinterprets the issue. Even on its own merits, the changes make little sense. As far as I can tell, this notification doesn't produce any sound in the first place.

This PR appears to be LLM slop, and the author’s GitHub activity suggests this is not an isolated case.

Issue #5079, opened 2026-03-20
TTY2 remains used by a useless Xorg instance (SDDM related)


EFrMG:
### System details

Omarchy 3.4.2. Intel UHD Graphics 620

### Making SDDM use weston with a setting for Wayland

The default X11-based SDDM greeter leaves TTY2 unusable and a 70MB in RAM Xorg instance occupying it.

### Changes

- Add weston package which SDDM already uses by default.

- Install tweak and migration which are minor.

Here is a picture of TTY2 freed after restarting SDDM:

![Image](https://github.com/user-attachments/assets/489035b6-28aa-4238-812c-a2c4f28519cc)

Normally, pressing `Ctrl + Alt + F2` would not show anything (The Xorg instance).

---

To test it, and please do:

```bash
omarchy-pkg-add weston

sudo tee "/etc/sddm.conf.d/50-wayland.conf" > /dev/null <<'EOF'
[General]
DisplayServer=wayland
EOF

sudo systemctl restart sddm
```

EFrMG:
I am trying a PR on the `dev` branch in case it is easier to manage.

To recap my findings from the previous PR on `master`:

Plymouth shows correctly on all monitors upon first booting and auto-login also happens well.

When trying to re-login upon running `omarchy-system-logout` the greeter (for the password) only registers the first monitor it can reach, and not any others, but still it shows up there. If this is a laptop monitor, then it would show in there and not in external monitors.

This frees up RAM and TTY2.

dhh:
Fixed in https://github.com/basecamp/omarchy/commit/4eb3a9198d53025f2b7b7eef073bde559809eca2

PR #5080, opened 2026-03-20
use Wayland option for SDDM

EFrMG:
For existing installations, and migration.

On issue: #5079

### DIFF
--- a/install/login/sddm.sh
+++ b/install/login/sddm.sh
@@ -11,6 +11,9 @@ Session=hyprland-uwsm
 
 [Theme]
 Current=omarchy
+
+[General]
+DisplayServer=wayland
 EOF
 fi
 
--- a/install/omarchy-base.packages
+++ b/install/omarchy-base.packages
@@ -133,6 +133,7 @@ unzip
 usage
 uwsm
 waybar
+weston
 whois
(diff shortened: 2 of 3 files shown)


EFrMG:
I did not request a code review and one was issued. I don't even have a Copilot subscription

EFrMG:
It seems the logout option from the menu does not work as expected now. Entering the password works, but there is nothing shown on screen, only the last frame of plymouth for a split second after the password is entered.

EFrMG:
The issue was the greeter not showing on external monitors, but it shows well.
Adding `hyprctl keyword monitor eDP-1,auto,auto,1` at the top of `omarchy-system-logout` could be worth in case laptop monitors are disabled when using an external monitor, which I do.

EFrMG:
By the way, Plymouth shows correctly on **all monitors** upon first booting and auto-login also happens well.
The issue is that when trying to re-login upon running `omarchy-system-logout` the greeter only registers the **first monitor** it can reach, and not any others, but still it shows up there.
I'd say logging out is not something I do very often, other than to test configs that need a new login shell, but it is something to keep in mind even if using Wayland as the display server for SDDM is more correct and saves computer resources.

PR #5088, opened 2026-03-21
feat: enable calc provider by default in walker launcher

nickkaltner:
## Summary
* Adds the `calc` provider to the default providers list in the Walker launcher configuration.
* Allows users to perform simple mathematics as they type without needing the `=` prefix.


nickkaltner:
the = prefix is non obvious, i just think it's better to not need it!

dhh:
The = prefix is there intentionally. Without it you can trigger calculations when you're doing anything with numbers. Prefer to have it as distinct, separate category. But the config is user land, so you can change to whatever you like!

Issue #4881, opened 2026-03-03
Bug: Layout toggle fails on special workspaces


LeoPazEs:
### System details

AMD Ryzen 7 5700G (16) @ 4.67 GHz, Omarchy 3.4.1

### What's wrong?

When attempting to toggle the layout on an active special workspace, the action fails. This happens because the current layout toggle logic relies on the hyprctl activeworkspace. However, when a special workspace is active on a monitor, hyprctl activeworkspace -j does not report it, causing the toggle command to be used in the workspace in the monitor.

## To Reproduce

1. Open a special workspace on any monitor.
2. Attempt to trigger the layout toggle shortcut/command.
3. Observe that the layout does not change in the special workspace but in the workspace that the monitor has.

## Expected behavior
The layout toggle should work seamlessly on special workspaces, just as it does on standard workspaces.

## Root Cause & Resolution
The issue stems from relying on the workspace ID for identification. By switching the logic to check the workspace name instead of the ID, we can bypass the limitation with hyprctl activeworkspace -j. This approach achieves the same functionality while successfully supporting multiple special workspaces with custom names.

## Related PR
A fix for this issue has been implemented and submitted in PR #4880.

jaderfeijo:
Still reproduces on Omarchy 4.0.3-1 / Hyprland 0.56.2, with the scratchpad focused and one window in it:

```
$ hyprctl monitors -j | jq -c '.[] | select(.focused) | {activeWorkspace, specialWorkspace}'
{"activeWorkspace":{"id":1,"name":"1"},"specialWorkspace":{"id":-98,"name":"special:scratchpad"}}

$ hyprctl workspaces -j | jq -c '.[] | {name,tiledLayout}'
{"name":"1","tiledLayout":"scrolling"}
{"name":"special:scratchpad","tiledLayout":"dwindle"}

$ omarchy-hyprland-workspace-layout-toggle      # SUPER + L
$ hyprctl workspaces -j | jq -c '.[] | {name,tiledLayout}'
{"name":"1","tiledLayout":"dwindle"}            # <- the wrong workspace flipped
{"name":"special:scratchpad","tiledLayout":"dwindle"}
```

[...]

PR #5099, opened 2026-03-22
Fix layout toggle script for special workspaces

arizkhalid:
`hyprctl activeworkspace` never returns special workspace. This caused the layout toggle to apply to the wrong workspace.

Fix: Check `hyprctl monitors` for an active special workspace before `hyprctl activeworkspace` and use its name to look up and toggle the correct layout.

Note: `hyprctl keyword workspace` doesn't accept negative IDs for special workspaces thats why the workspace name must be used instead.


arizkhalid:
Dropped `hyprctl eval` because it exits with code 0 even when unsupported (non-lua config), so the || fallback never runs. 

Kept `hyprctl keyword` which works reliably.

arizkhalid:
@dhh Copilot is unable to review this PR, is that intentional or a settings issue? This PR fixes a real bug, if its not going to be merged, it should at least be reviewed.

jaderfeijo:
This was written against an older version of `bin/omarchy-hyprland-workspace-layout-toggle`. The script now saves each workspace's layout and applies it through `hyprctl eval`. This PR replaces that eval with plain `hyprctl keyword`, and it doesn't handle the saved layout files. It also reads the scratchpad from `.[0]` instead of the focused monitor, which picks the wrong one with more than one monitor. #11008 fixes #4881 on top of the current `quattro` script and handles both: https://github.com/omacom/omarchy/pull/11008

PR #5106, opened 2026-03-23
Add hardware fix for Intel WiFi 7 EHT on Panther Lake

spencerbull:
Detect Intel BE200/BE211 WiFi cards and disable WiFi 7 (EHT) via modprobe config. The iwlwifi driver has broken EHT RX adaptation causing APs to drop to MCS 0, making WiFi unusable. Falling back to WiFi 6 (HE) restores full performance.

Temporary fix: remove when fixes land in the firmware/driver.


dvic:
Does anyone know more about this issue? Is there any insight in when this might be fixed upstream?

PR #5112, opened 2026-03-24
Add Android Studio in dev install and remove menus

HowieDuhzit:
## Summary
- Adds Android Studio in Install/Remove development menus


patrick-boc:
We should also add `-Dawt.toolkit.name=WLToolkit` to support Wayland.

As suggested in: https://github.com/basecamp/omarchy/issues/1300#issuecomment-3332821770 and https://github.com/basecamp/omarchy/discussions/2004.

However, IntelliJ has switched default to Wayland since version 2026.1, so I suppose it will come to Android Studio soon as well.

PR #5118, opened 2026-03-26
Fix Neovide black background when syncing Omarchy theme

### LINKED ISSUE #5107: Neovide Shows Dark background
### System details

Omarchy 3.4.2

### What's wrong?

When setting neovide theme to sync with omarchy theme, neovide's background becomes black (#000000).

When switching neovide's theme to another theme different than current omarchy theme, the background gets rendered normally.

mplaczek99:
## Summary
- add a migration that updates `~/.config/nvim/plugin/after/transparency.lua` to skip transparency overrides when running in Neovide
- prevent `Normal`/`NormalNC` background from becoming transparent in Neovide, which avoids the black `#000000` background
- keep existing transparency behavior unchanged for terminal Neovim

## Testing
- `bash -n migrations/1774489385.sh`
- run the migration twice in a temporary `$HOME` and verify the `vim.g.neovide` guard is added once

Closes #5107

### DIFF
--- a/migrations/1774489385.sh
+++ b/migrations/1774489385.sh
@@ -13,6 +13,9 @@ end
 
 EOF
 
-  cat "$TRANSPARENCY_FILE" >>"$temp_file"
-  mv "$temp_file" "$TRANSPARENCY_FILE"
+  if cat "$TRANSPARENCY_FILE" >>"$temp_file"; then
+    mv "$temp_file" "$TRANSPARENCY_FILE"
+  else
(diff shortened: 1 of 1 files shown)


dhh:
We need to fix this in omarchy-nvim, so it's solved on install. See omacom-io/omarchy-pkgs.

mplaczek99:
Thanks for the feedback. Moving this fix upstream to omacom-io/omarchy-pkgs so it is applied at install time instead of via a migration.

PR #5131, opened 2026-03-27
Add FaceTime HD webcam support for T2 MacBooks

fedesapuppo:
## Summary

- Adds `facetimehd-firmware`, `facetimehd-data`, and `facetimehd-dkms` packages to the T2 setup
- Autoloads the `facetimehd` kernel module at boot

The built-in FaceTime HD webcam on T2 MacBooks needs a reverse-engineered DKMS driver and Apple firmware blobs. Without these packages, the webcam doesn't work at all.

Follows the same package-install pattern already used in `fix-t2.sh`.

### DIFF
--- a/install/config/hardware/apple/fix-t2.sh
+++ b/install/config/hardware/apple/fix-t2.sh
@@ -9,7 +9,10 @@ if lspci -nn | grep -q "106b:180[12]"; then
     apple-t2-audio-config \
     apple-bcm-firmware \
     t2fanrd \
-    tiny-dfr
+    tiny-dfr \
+    facetimehd-firmware \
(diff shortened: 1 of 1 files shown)


fedesapuppo:
Closing this — the T2 webcam is already handled by `apple-bce` (VHCI) + `uvcvideo`, both included in the `linux-t2` kernel. The `facetimehd-*` packages (bcwc_pcie driver) are for pre-T2 Macs with Broadcom 1570 PCIe cameras and won't do anything useful on T2 hardware. No additional packages or configuration are needed for the webcam beyond the existing `linux-t2` kernel install.

PR #5132, opened 2026-03-27
Add tuned fan curve for T2 MacBooks

fedesapuppo:
## Summary

- Writes a custom `/etc/t2fand.conf` during T2 setup with a less aggressive fan curve

The default t2fanrd config spins fans up early and loudly. This config keeps fans off below 55°C and ramps linearly to full speed at 75°C, which better matches the MacBook's thermal design and keeps the machine quiet under light loads.

## Test plan

- [x] Verify `/etc/t2fand.conf` is written with the custom curve
- [x] Confirm fans stay quiet under light load and ramp up under heavy load


dhh:
Should probably also add a migration here?

fedesapuppo:
Added the migration in #5176 😁

PR #5133, opened 2026-03-27
Reduce shutdown timeout to 5s for T2 MacBooks

fedesapuppo:
## Summary

- Adds a systemd drop-in that reduces `DefaultTimeoutStopSec` from 90s to 5s

T2 kernel drivers (especially `apple-bce`) sometimes hang during shutdown. The default 90-second timeout means users stare at a blank screen for over a minute waiting for systemd to kill the stuck service. 5 seconds is plenty for a normal shutdown; anything still running after that is stuck.

## Test plan

- [x] Verify `/etc/systemd/system.conf.d/10-faster-shutdown.conf` exists
- [x] Confirm shutdown completes within a few seconds instead of hanging


fedesapuppo:
Closing this — `install/config/fast-shutdown.sh` already configures `DefaultTimeoutStopSec=5s` globally for all systems. It runs at line 20 of `all.sh`, before `fix-t2.sh`, and also includes a user service drop-in (`user@.service.d/faster-shutdown.conf`) that this PR was missing.

PR #5134, opened 2026-03-27
Disable power key on T2 MacBooks to prevent accidental suspend

fedesapuppo:
## Summary

- Sets `HandlePowerKey=ignore` in `/etc/systemd/logind.conf` during T2 setup

Suspend is broken on T2 Macs — waking from suspend often results in a black screen with an unresponsive keyboard, requiring a hard reboot. The power key (top-right of keyboard) is easy to hit accidentally and triggers suspend by default. Disabling it prevents data loss from forced reboots.

With this change, the power key simply opens the same menu as Super + Esc, so shutdown/reboot are still accessible through that menu.

## Test plan

- [x] Verify `HandlePowerKey=ignore` is set in `/etc/systemd/logind.conf`
- [x] Confirm pressing the power key no longer triggers suspend
- [x] Confirm power key opens the same menu as Super + Esc


fedesapuppo:
Closing this — `install/config/hardware/ignore-power-button.sh` already sets `HandlePowerKey=ignore` globally for all systems. It runs at line 36 of `all.sh`, before `fix-t2.sh` at line 58, so T2 MacBooks are already covered.

PR #5140, opened 2026-03-27
Add WiFi resume hook for T2 MacBooks

fedesapuppo:
## Summary

- Installs a systemd system-sleep hook that reloads the Broadcom WiFi driver after waking from suspend

The Broadcom WiFi driver (`brcmfmac`) on T2 MacBooks loses connection state after suspend and silently fails — WiFi appears connected but can't transmit data. This hook unloads and reloads the kernel module on resume, restoring connectivity.

## Tested: 

- Run `fix-t2.sh` on a T2 MacBook
- Suspend and resume the machine
- WiFi reconnects automatically after wake

### DIFF
--- a/install/config/hardware/apple/fix-t2.sh
+++ b/install/config/hardware/apple/fix-t2.sh
@@ -41,4 +41,15 @@ high_temp=75
 speed_curve=linear
 always_full_speed=false
 EOF
+
+  sudo mkdir -p /etc/systemd/system-sleep
+  cat <<'HOOK' | sudo tee /etc/systemd/system-sleep/wifi-resume >/dev/null
+#!/bin/bash
+if [[ $1 == "post" ]]; then
+  logger -t wifi-resume "Reloading brcmfmac after resume"
+  modprobe -r brcmfmac 2>/dev/null
+  modprobe brcmfmac || logger -t wifi-resume "Failed to reload brcmfmac"
+fi
+HOOK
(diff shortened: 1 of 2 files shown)


dhh:
We need a migration for this too, no?

fedesapuppo:
> We need a migration for this too, no?

We do! 😅 I'm on it David!

dhh:
Did you or anyone else test this?

fedesapuppo:
Hey David! I just tested on my T2 mac, and I caught a bug. I installed the hook to
  /etc/systemd/system-sleep/, but systemd-sleep only scans
  /usr/lib/systemd/system-sleep/ (confirmed in the man page and in the binary's
  strings). The hook this PR installed never ran.

  It appeared to work during my initial review because an older copy existed at
  /usr/lib/systemd/system-sleep/wifi-resume.bak from a previous manual setup, and
  systemd runs all executables in that directory regardless of extension. That stale
  file was silently handling my resumes.

  Pushed f25bb1f0 to install at the correct path and drop the redundant mkdir -p (the
  directory ships with systemd).

  Verified on my T2 MacBook Air:
  - Without the hook: brcmfmac reports bus is down on resume, traffic silently fails.
  - With the hook at the correct path: hook fires (wifi-resume[...]: Reloading brcmfmac
   after resume), module reloads, WiFi passes traffic.

PR #5141, opened 2026-03-28
Fix udev RUN commands to use systemd-run

tonysm:
I was getting errors in the powersaver udev rule handling script. I think it's because udev runs in sandbox mode or something ([docs](https://www.freedesktop.org/software/systemd/man/latest/udev.html?__goaway_challenge=meta-refresh&__goaway_id=9523b5dbc9faac15267d48f16179ffa4#:~:text=Note%20that%20running%20programs,processes%20is%20not%20allowed)), so it can't access the scripts in /home/{user}/...

This change hands-off the scripts to systemd-run, which will run them outside the sandbox mode. The --no-block returns immediatelly so it doesn't block udev event processing.

Here's the error I was getting:

```
Mar 27 23:42:10 [REDACTED] (udev-worker)[567]: AC: Process '/usr/bin/powerprofilesctl set balanced' failed with exit code 2.
Mar 27 23:42:10 [REDACTED] (udev-worker)[567]: AC: Process '/home/[REDACTED]/.local/share/omarchy/bin/omarchy-wifi-powersave on' failed with exit code 237.
```

With the command:

```
journalctl -u systemd-udevd -e
```

---

I've been getting weird Wifi issues since the update that contained these powersaver rules. Hoping this gets it fixed.

### DIFF
--- a/install/config/powerprofilesctl-rules.sh
+++ b/install/config/powerprofilesctl-rules.sh
@@ -12,8 +12,8 @@ if omarchy-battery-present; then
     battery_profile="${profiles[1]}"
 
     cat <<EOF | sudo tee "/etc/udev/rules.d/99-power-profile.rules"
-SUBSYSTEM=="power_supply", ATTR{type}=="Mains", ATTR{online}=="0", RUN+="/usr/bin/powerprofilesctl set $battery_profile"
-SUBSYSTEM=="power_supply", ATTR{type}=="Mains", ATTR{online}=="1", RUN+="/usr/bin/powerprofilesctl set $ac_profile"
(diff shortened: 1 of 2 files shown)


jelenv:
I can confirm that I have the same issue on my system. 

Based on debugging with agent my `home` dir was not mounted yet at the time of this udev rule firing. In logs:
```sh
2026-04-07 10:13:56 ACAD: Failed to execute '/home/user/.local/share/omarchy/bin/omarchy-wifi-powersave off': No such file or directory
```

It suggested that best fix would be to have the `omarchy-wifi-powersave` in `/usr/local/bin/omarchy-wifi-powersave`, because udev rules should not ideally call files from `home` dir. Your fix is better, in keeping the omarchy architecture unchanged (all omarchy bin scripts are under home dir).

It's not up to me to decide where it's good to place omarchy bin scripts, so I am just adding this here for possible consideration.

dhh:
This has been fixed 👍

PR #5142, opened 2026-03-28
Add UseDomains=yes for network files in DNS setup script when using DHCP

### LINKED ISSUE #1870: omarchy Ignores DHCP-Provided Search Domain, Requires Manual resolv.conf Edit
### System details

Framework 13, Omarchy ~3

### What's wrong?

When using my omarchy machine on my local network with customized Unbound DNS server with a DHCP-configured local domain (example.net), omarchy fails to respect the DHCP-provided search domain. Local records like `thing.example.net` should resolve as `thing`, but this only works after manually editing `/etc/resolv.conf` to set `search example.net` instead of the default `search .` and manually setting the `nameserver [ipaddress]` of my local dns server. I believe there are at least two issues at play here.

This may be related to #1856.

sgruendel:
fixes #1870


sgruendel:
Closing, works as expected on Quattro with DHCP.

PR #5144, opened 2026-03-28
Add battery charging threshold to toggle menu

spaceXrace:
This pull requests adds a Battery Charge Threshold option to the Toggle menu.

I noticed that https://github.com/basecamp/omarchy/issues/1289 was closed, but I was hoping that you might reconsider adding this. Apple has added this on macOS 26.4, so now this is a standard feature on Windows, macOS and other Linux distros.

The option is only visible if a battery which supports charging threshold is detected.

The toggle just switches the limit between 100% and 80%.

If the battery also supports setting a start charging threshold, then it will be set to 95% / 75%. Note that the battery will not bounce between 75% and 80% in this case, as the power adapter is used as the primary power source once 80% is reached. The start threshold would only be reached if the device is plugged in for multiple days and the battery self discharges.

It also supports systems with multiple batteries, for example some older ThinkPads like the T480. (The limit is applied to all batteries in the system).

I tested this on a laptop with one battery. I also tested the logic for multiple batteries with fake battery entries, but I don't have a laptop with two real batteries to test.


dhh:
This is very nice, but I can't get it to work on the XPS. It says it's available, but toggling between 80% and 100% doesn't do anything. cc @spencerbull

spencerbull:
Should we hide this menu option if there isn't a battery?

PR #5165, opened 2026-03-31
Show active power profile in menu

caariiboouu:
## Summary
- The power profile menu didn't clearly indicate which profile was currently active — walker's cursor always starts on the first item (power-saver), making it look like that was the current mode
- Appends "(active)" to the current profile name so users can tell at a glance which mode they're in

## Test plan
- [ ] Click the battery icon in waybar to open the power profile menu
- [ ] Verify the currently active profile (e.g. "performance") shows "(active)" next to it
- [ ] Select a different profile and reopen the menu to confirm the label moves
- [ ] Verify selecting the active profile (with the suffix) doesn't break `powerprofilesctl set`

### DIFF
--- a/bin/omarchy-menu
+++ b/bin/omarchy-menu
@@ -241,11 +241,24 @@ show_setup_menu() {
 }
 
 show_setup_power_menu() {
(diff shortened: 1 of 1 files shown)


dhh:
It's using italics, like the fonts menu, to signify which power profile is selected.

PR #5179, opened 2026-04-01
(Bash) Add lazy initialization of `try` and `mise` commands.

innerviewer:
Pull request with changes for #5178.


dhh:
Lazy loading mise won't work as then mise shims won't get activated. But we could lazy load try.

Issue #5174, opened 2026-04-01
Screenshot not working if the Pictures folder is removed


ronald2wing:
### System details

3.4

### What's wrong?

It should create the Pictures folder instead of showing an error.

brunorwx:
Hi, I've created a PR with a possible solution for this issue, any feedback is welcome!

cyberseekerx:
I think It should be uppercase "Pictures"
or you have to specify it somewhere config

brunorwx:
The lowercase "pictures" is only in the PR title, the changes actually provide a dynamic solution the dir name.

PR #5181, opened 2026-04-01
Create pictures dir if non existent

brunorwx:
Create pictures dir if non existent.

Issue #5174


Kasui92:
@dhh 
This [commit](https://github.com/basecamp/omarchy/pull/5181/commits/a40596443166b350e445e495369a51164ae45737) removed the `mkdir -p` that actually creates the folder.

brunorwx:
I think you are right, I've created a PR where I just added the `mkdir -p` cmd. 

https://github.com/basecamp/omarchy/pull/5189

PR #5183, opened 2026-04-02
Remove jetbrains specific hyprland config

NicolasDorier:
IDEs based on IntelliJ [2026.1](https://blog.jetbrains.com/platform/2026/02/wayland-by-default-in-2026-1-eap/) (released this week) now support Wayland natively properly. Since Hyprland is introducing another breaking change to the configuration system, and we do not want to maintain something that is no longer needed, this configuration should now be removed.

I personally updated Rider and removed the configuration file,
and noticed only one minor issue.

The following popup opens in the top-right corner instead of the center. However, this is not something that can be fixed through Hyprland configuration.
This is because unlike the `Search Files` popup, this one isn't a separate wayland window anymore.

image

As a workaround, you can resize the popup to fill the screen by dragging the bottom-right corner.


dhh:
Just tested. Works great with RubyMine. Only issue I see is the toolbox is undersized and a bit weirdly placed. But the editor itself is now all correct:

image

vmarcinko:
Although things are getting better, but I still have issues with floating windows in my Intellij IDEA 2026.1 and (latest) Omarchy 3.5.0 (with Hyprland 0.54.3) as can be seen in this video with Search&Replace andlater confirmation floating windows:

https://github.com/user-attachments/assets/f7402e33-429e-4d06-922e-4436481ba2f9

So basicailly "follow mouse" is messing things up. When i set my hyprland window rule such as this:
windowrule = no_follow_mouse on, match:class ^(jetbrains-.*)

Then things get much better, but it also disables "on-mouse-follow" focusing from other windows to main Intellij IDEA one. One more interesting thing - sicne I am new to Omarchy, I really haven't bought this whole "mouse follow focus" feature, so I decided to disable it using "input" hyprland config, but strangely, although in other windows it doesn't have this feature anymore, in Intellij IDEA it still is enabled (floating windows dissapeared/defocused on mouse leaving their area), and things are messed up. Strange...

NicolasDorier:
https://github.com/basecamp/omarchy/pull/5239

PR #5186, opened 2026-04-02
Add ONCE to Install->Service menu

kevinmcconnell:
Adds an entry to the Service menu that installs the ONCE package, registers its background service, and launches the TUI.

### DIFF
--- a/bin/omarchy-menu
+++ b/bin/omarchy-menu
@@ -317,12 +317,13 @@ show_install_menu() {
 }
 
 show_install_service_menu() {
-  case $(menu "Install" "  Dropbox\n  Tailscale\n󱇱  NordVPN [AUR]\n󰟵  Bitwarden\n  Chromium Account") in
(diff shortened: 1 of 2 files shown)


bin/omarchy-menu
```diff
@@ -317,12 +317,13 @@ show_install_menu() {
 }
 
 show_install_service_menu() {
-  case $(menu "Install" "  Dropbox\n  Tailscale\n󱇱  NordVPN [AUR]\n󰟵  Bitwarden\n  Chromium Account") in
+  case $(menu "Install" "  Dropbox\n  Tailscale\n󱇱  NordVPN [AUR]\n󰟵  Bitwarden\n  Chromium Account\n󰏖  ONCE") in
```
kevinmcconnell, line 320: Styled this in uppercase to match the way the name is usually written (e.g. on https://once.com), and went with a fairly generic package icon.

PR #5199, opened 2026-04-03
feat: Support Global Search in omarchy-menu

samrudhnelli:
Support "Global search" option in omarchy-menu without modifying walker and elephant.
This allows searching for commands inside sub-menus using a stored JSON list.

### DIFF
--- a/bin/omarchy-menu
+++ b/bin/omarchy-menu
@@ -3,6 +3,8 @@
 # Launch the Omarchy Menu or takes a parameter to jump straight to a submenu.
 
 export PATH="$HOME/.local/share/omarchy/bin:$PATH"
+MENU_JSON="$HOME/.local/share/omarchy/bin/omarchy-menu-options.json"
+AUTOMATE_QUEUE="/tmp/automate-omarchy-menu"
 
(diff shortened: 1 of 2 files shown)


samrudhnelli:
@dhh Need your view on this.
I feel supporting global search would help finding options easily.
Since, the menu has a lot of options now, it is not possible to remember every path and something like global search would surely help.

sspaeti:
I did the same and added a "flat" menu that I can fuzzy search, overall made it the default shortcut for `SUPER CTRL, SPACE`  (`bind = SUPER CTRL ALT, SPACE, exec, ~/.config/hypr/sspaeti/omarchy-menu`)- I asked Claude to generate it from the original menu. If that were integrated, it would be much better.

Leaving it here, just in case it helps. But I think that's a feature many would want: https://github.com/sspaeti/dotfiles/blob/b79cbdb961cca704af6855c44538d6ff2b43d366/hypr/.config/hypr/sspaeti/omarchy-menu-flat

dhh:
Closing because Quattro now provides global search natively through the Quickshell menu. Thanks for the contribution.

PR #5201, opened 2026-04-03
feat: transparent support for ddcutil on desktop

jamerrq:
### Description
This PR updates the `omarchy-brightness-display` script to provide transparent, default support for desktop monitors using `ddcutil`.  

Currently, default `XF86MonBrightnessDown`/`XF86MonBrightnessUp` keybindings fail on desktops due to the lack of internal backlights exposed through `/sys/class/backlight`.

By adding a simple `has_backlight` heuristic, the script now automatically falls back to `ddcutil` for executing brightness up/down steps.

### Details
- Checks for brightness devices dynamically on script launch.
- If no laptop backlight is detected, parses the `+5%` / `5%-` step format manually.
- Caches the value temporarily in `/tmp/ddcutil_brightness_omarchy` and fires `omarchy-swayosd-brightness` immediately so that **visual OSD feedback is instantaneous**.
- Runs `ddcutil setvcp 10` asynchronously in the background.

This brings an out-of-the-box solution for both laptops (native `brightnessctl`) and desktop monitors (`ddcutil`), while still allowing users to view the cool SwayOSD notifications.

### DIFF
--- a/bin/omarchy-brightness-display
+++ b/bin/omarchy-brightness-display
@@ -1,21 +1,79 @@
 #!/bin/bash
 
 # Adjust brightness on the most likely display device.
+# Automatically falls back to ddcutil on desktops without a native backlight.
 # Usage: omarchy-brightness-display <step>
 
 step="${1:-+5%}"
 
-# Start with the first possible output, then refine to the most likely given an order heuristic.
-device="$(ls -1 /sys/class/backlight 2>/dev/null | head -n1)"
-for candidate in amdgpu_bl* intel_backlight acpi_video*; do
-  if [[ -e /sys/class/backlight/$candidate ]]; then
-    device="$candidate"
+# Determine if we have internal hardware backlight devices
+has_backlight=false
+for candidate in /sys/class/backlight/*; do
+  if [[ -e "$candidate" ]]; then
+    has_backlight=true
     break
   fi
 done
 
-# Set the actual brightness of the display device.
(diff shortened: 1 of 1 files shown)


felipe3dfx:
Hey @jamerrq — I took a different cut at the same problem in #5282, wanted to flag it here so you and the maintainers see both approaches side by side.

Main differences from yours:

- Install hook + migration so users don't have to install `ddcutil` and set up the `i2c` group / udev rule by hand.
- Blocking `flock` (instead of `-n`) to fix silent drops during rapid bursts.
- Debounce pattern — one `setvcp` per gesture — which cuts monitor NVRAM wear and keeps the OSD in sync. KDE PowerDevil does the same thing for the same reasons ([D8626](https://phabricator.kde.org/D8626)).
- Iteration over all detected i2c buses for multi-monitor setups.

Not trying to replace yours — just want to make sure basecamp sees both takes. Happy to merge ideas in either direction.

jamerrq:
Hi @felipe3dfx. Thanks for the heads up and for linking your PR.

Your approach is definitely much more robust. I really appreciate the debounce pattern to protect the monitor's NVRAM—that’s an excellent and critical catch that I hadn't considered. Also, dynamic detection of i2c buses and automatically handling the migration/installation hooks makes it a much better fit for Omarchy's "it just works" philosophy.

I'm more than happy to close this PR in favor of #5282 so the maintainers can focus entirely on yours, as it clearly handles all the edge cases properly.

PR #5209, opened 2026-04-04
feat(omarchy-setup-dns): add live "Current DNS" display + quit option

Toryalai1:
## Summary

This PR improves the `omarchy-setup-dns` script (the one behind **Setup → DNS** in the menu):

- **New:** Added `show_current_dns()` function that displays the currently active DNS provider (Cloudflare / Google / DHCP / Custom) with formatting using `gum`.
- **New:** Shows the updated DNS status **immediately after** applying the change.
- **Improvement:** Add quit option
- **Minor:** Slightly more robust parsing of `resolvectl` output and better user messages.

## Motivation

Users often want to know what DNS is actually in use before changing it. The previous script was silent until you ran `resolvectl status` manually. This adds QoL

## Changes

- Added `show_current_dns` function at the top
- Call it once at startup (interactive mode)
- Call it again after `systemctl restart` so the user sees the result
- Updated Custom case to include DoT

## Testing

- Tested on Omarchy 3.5
- Verified Cloudflare, Google, DHCP, and Custom all work and persist across reboot
- Checked that `UseDNS=no` lock/unlock logic still functions correctly
- DoT status displays correctly

---

### Before

The script used to jump straight into the menu with **no indication** of the current DNS setting.

screenshot-2026-04-04_08-41-10

### After

Now it shows a clear, nicely formatted **"Current DNS"** status both when the script starts **and** right after the change is applied.

screenshot-2026-04-04_08-42-46

screenshot-2026-04-04_08-43-53

### DIFF
--- a/bin/omarchy-setup-dns
+++ b/bin/omarchy-setup-dns
@@ -1,14 +1,11 @@
 #!/bin/bash
-
 lock_dns_to_resolved() {
   for file in /etc/systemd/network/*.network; do
     [[ -f $file ]] || continue
     if ! grep -q "^\[DHCPv4\]" "$file"; then continue; fi
-
     if ! sed -n '/^\[DHCPv4\]/,/^\[/p' "$file" | grep -q "^UseDNS="; then
       sudo sed -i '/^\[DHCPv4\]/a UseDNS=no' "$file"
     fi
-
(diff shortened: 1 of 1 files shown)


Toryalai1:
Any objections?

PR #5211, opened 2026-04-04
closed, fixed locally

Qeya:
closed, fixed locally

### DIFF
--- a/bin/omarchy-update-perform
+++ b/bin/omarchy-update-perform
@@ -19,6 +19,14 @@ omarchy-update-system-pkgs
 omarchy-migrate
 omarchy-update-aur-pkgs
 omarchy-update-orphan-pkgs
+
+if command -v hyprpm >/dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


Sleaker:
Saw this and was wondering if I was impacted but looks like there might be some missing dependencies before this will work out of the box. cpio doesn't appear to be a required dep on hyperland which provides hyprpm but it still is required to run update. Also since the commands are provided by hyperland, is the if command -v checks doing anything?

 or is that just a sanity check to make sure the system isn't in a bricked state?

```
~ ❯ hyprpm update
[sudo] password for ....:
→ The hyprpm state store doesn't exist. Creating now...
✖ Missing dependency: cpio

✖ Could not update. Dependencies not satisfied. Hyprpm requires: cmake, cpio, pkg-config, git, g++, gcc
```

Qeya:
closed, fixed locally

Qeya:
closed, fixed locally

PR #5212, opened 2026-04-04
Add face unlock via howdy

Woogy7:
## Summary

Adds face unlock via [howdy](https://github.com/boltgolt/howdy) for laptops with an infrared camera, following the fingerprint pattern: `omarchy-setup-security-howdy` and `omarchy-remove-security-howdy`, plus _Setup > Security > Face Unlock_ and _Remove > Security > Face Unlock_ menu entries.

Now targets `quattro`, since the Quickshell lock screen only exists there.

## What it does

**Setup script**
- Detects the IR camera (greyscale capture device) and bails with a clear message when there is none
- Installs `howdy-git` and `linux-enable-ir-emitter` from the AUR, configures the IR emitter, enrolls a face model
- Wires PAM for `sudo` and polkit with `pam_unix.so try_first_pass` above `pam_howdy.so`, so a typed password authenticates instantly and only an empty Enter triggers a scan (no silent face auth on unattended `sudo`)
- Writes `/etc/pam.d/omarchy-lock-face` for the lock screen, last, once a model exists
- Pins howdy's `workaround = off` so `pam_howdy` never opens a password prompt inside the lock screen's PAM conversation
- Installs a pacman hook that warns when a soname bump breaks `python-dlib` (otherwise the failure is silent, auth just falls back to password)

**Lock screen (Quickshell)**
- Third `PamContext` on `omarchy-lock-face`, alongside `omarchy-lock-password` and `omarchy-lock-fingerprint`, enabled when that PAM file and `pam_howdy.so` exist
- One bounded scan on arrival: the first key or pointer event after a quiet gap (back from a break, lid opened, woken from sleep), and on an empty Enter. Never on the lock itself, since the person who locked it is still there. No retry loop, because howdy holds the camera and CPU for the whole scan and a loop would drain a locked laptop
- Only the Enter-triggered scan reports "Face not recognized"; automatic scans stay silent, like a fingerprint that was never touched
- Face glyph in the placeholder colour where the fingerprint icon lives (left of it when both are enrolled), brighter while scanning; the dot reserve covers both icons

**Polkit dialog (Quickshell)**
- Dim face glyph at the right edge of the password field whenever `pam_howdy` is in `/etc/pam.d/polkit-1`, same glyph and spot as the lock screen; lit in the accent with "Scanning..." while an empty submit is in flight, "Not recognized" on a failed scan. Face never takes over the dialog like fingerprint mode does, since `pam_unix` stays first

**Design notes**
- Face scans always follow an intent signal: arriving at the lock screen (first input after a quiet gap) or an Enter on an empty field. They never fire on the lock keystroke or on a polkit dialog opening. Face is a passive biometric, and a polkit prompt can be raised by any process, so an empty Enter is the one-keystroke consent that keeps a rogue `pkexec` from being authorised just because someone is sitting at the screen. This is the same reasoning as the `pam_unix`-first ordering for `sudo` from the earlier review
- Polkit also has a single PAM conversation per request, so an automatic empty submit would block typing a password for the length of the scan (about 4 s); the lock screen avoids that only because it runs separate PAM contexts
- No retry loop for face, unlike fingerprint: howdy actively drives the IR camera and CPU while it runs

**Docs and tests**
- Manual: face unlock section in hardware authentication
- `test/shell.d/lock-face-indicator-test.sh` covers the indicator, the reserve with one and two icons, and empty Enter requesting a scan only when a model is enrolled
- `test/shell.d/lock-face-trigger-test.sh` pins the trigger rules: no scan on the secure transition, arrival gap, no retry timer

## Hardware requirements

Laptop with an IR camera (the kind Windows Hello uses). Tested on an ASUS Zenbook 14 UX3405CA (Core Ultra 9 285H) on Omarchy 4.0.0.

## Test plan

- [x] `./test/cli`, `./test/shell` lock and menu tests pass; the new face test fails against the previous `LockView.qml` and passes with this change
- [x] Lock screen: locking does not scan; first mouse move or key after a few seconds away scans and unlocks in about 1.5 s; empty Enter rescans and reports failure in the field; typed password still works
- [x] `sudo` and polkit (1Password): empty Enter scans, typed password skips the camera; polkit dialog shows the face hint, lights it while scanning, and closes on approval (`pkexec true` exit 0)
- [x] Remove script cleans the PAM files, hook, IR service, and packages
- [ ] Setup on a second IR-camera machine (Zenbook UX5406AA and X1 Carbon owners in this thread, if you are still up for it)


pomartel:
I tested it on a Lenovo X1 Carbon gen 13 and the installation went well. However the integration with the lock screen is a bit buggy. Just hitting Enter does not trigger the camera. I have to enter at least a character and then hit enter.

Woogy7:
> I tested it on a Lenovo X1 Carbon gen 13 and the installation went well. However the integration with the lock screen is a bit buggy. Just hitting Enter does not trigger the camera. I have to enter at least a character and then hit enter.

Thanks for trying it out on the X1 Carbon!

This is the same for me — it's caused by ignore_empty_input = true in hyprlock's config, which means an empty Enter doesn't trigger PAM at all. You need at least one character + Enter for hyprlock to start authentication (and then howdy kicks in).

We could fix this by having the setup script set ignore_empty_input = false, so just pressing Enter triggers PAM and howdy scans your face. The --remove flag would restore it to true. This is the same pattern the fingerprint setup uses for toggling fingerprint:enabled.

The only downside is if howdy times out (e.g. you're not looking at the camera), it falls through to password auth with an empty string, which briefly flashes a failure message. Minor cosmetic issue.

Unfortunately there's no way to have howdy scan automatically when the lock screen appears — hyprlock only triggers PAM on input submission, and the maintainers explicitly rejected auto-triggering for biometrics (hyprwm/hyprlock#910).

pomartel:
Thanks for the explanations. It makes sense. I don't know if this belongs in Omarchy but it's great that someone did the work for those who want to try it.

Woogy7:
Thanks 🙏

psynyde:
@Woogy7 Glad to see some work being done with howdy here at omarchy.
btw you can also enable howdy for sddm. here's my sddm pam config for example which im using on my base arch:

```conf
#%PAM-1.0
auth        sufficient  pam_howdy.so
auth        include     system-login
-auth       optional    pam_gnome_keyring.so
-auth       optional    pam_kwallet5.so
account     include     system-login
password    include     system-login
-password   optional    pam_gnome_keyring.so    use_authtok
session     optional    pam_keyinit.so          force revoke
session     include     system-login
-session    optional    pam_gnome_keyring.so    auto_start
-session    optional    pam_kwallet5.so         auto_start
```

Works the same way as hyprlock. you press enter with an empty input and it'll enable the ir-camera to scan.

Woogy7:
> Works the same way as hyprlock. you press enter with an empty input and it'll enable the ir-camera to scan.

Thanks for the tip! Good point on SDDM — I have autologin enabled so I go straight from LUKS to desktop, but this would definitely help users without autologin. I'll add SDDM support to the setup script.

Woogy7:
> Thanks for the explanations. It makes sense. I don't know if this belongs in Omarchy but it's great that someone did the work for those who want to try it.

I've made two updates based on your feedback:

Lock screen fix: The script now sets ignore_empty_input = false in the hyprlock config, so you can just press Enter to trigger the face scan — no need to type a character first. The --remove flag restores it.

1Password support: The script also creates a polkit PAM config with howdy, so face unlock works for 1Password and other polkit-based apps too.

Also added SDDM login screen support based on suggestion from psynyde.

Let me know what you think!

psynyde:
@Woogy7 Can you test howdy without `linux-enable-ir-emitter` package?
I'm also on ASUS Zenbook 14 UX3405CA but with the Core ultra 7 255H variant. 
I can use howdy without the ir emitter package just fine.
I guess it doesn't hurt to install the package for wider hardware coverage but still would be nice to look into.

One more thing, the [linux-enable-ir-emitter-beta](https://aur.archlinux.org/packages?O=0&K=linux-enable-ir-emitter-beta) package doesn't ship with a systemd service. So maybe in future you need other ways to configure this.

Woogy7:
@psynyde Hey thanks, it worked on my side withouth the IR emitter too, but I kept getting "frames too dark" issues in low light. The package ensures the IR emitter activates consistently, especially after sleep/suspend (the systemd service re-enables it). Doesn't hurt to have it as a safety net for wider hardware coverage.

Good flag on the beta package dropping the systemd service. The script currently installs the stable package which ships the service. I'll add a guard so the systemctl calls are skipped gracefully if no service file is present, keeping it forward-compatible.

psynyde:
@Woogy7 hi.
You may want to look at [this](https://gemini.google.com/share/816c12726da9)
I had an issue where howdy wouldn't work with polkit. the chat above shows the issue and the solution.

Woogy7:
Thanks @psynyde, appreciate the input as always!
The howdy-git AUR package that the setup script installs already ships this exact workaround. It installs a drop-in at /usr/lib/systemd/system/polkit-agent-helper@.service.d/10-howdy.conf with:

[Service]
PrivateDevices=no
DeviceAllow=char-video4linux rw
DeviceAllow=/dev/uinput rw

(with an extra DeviceAllow=/dev/uinput rw for the workaround = input mode). 
The upstream PKGBUILD calls it out as "Workaround for polkit 127+ breaking howdy (boltgolt/howdy#1077)".

That's why polkit face unlock for 1Password works out of the box with the current script.

psynyde:
image

im not sure when mine broke but i don't have the `/usr/lib/systemd/system/polkit-agent-helper@.service.d/10-howdy.conf` file on my system and maybe it's because i switched to chaotic aur version. 
In any case, this should serve as a reference if someone else faces this same issue.

mateuszkowalczyk:
Hi and thanks for your work on this feature! I've previously configured `howdy` on my Asus Zenbook UX5406AA, and everything seems to be working fine. I wanted to help with integrating it into Omarchy out of the box and found your PR.

I reviewed your changes, and a few discussion points came to mind:

### 1. Password fallback on the hyprlock screen
Howdy sometimes fails to detect my face. I've noticed this mostly when the background is very bright, for example with a window behind me. In such cases, it would be helpful to be able to log in using password.

This can be achieved by adding:

`auth      sufficient pam_unix.so try_first_pass likeauth nullok`

before:

`auth      sufficient pam_howdy.so`

I took this solution from https://wiki.archlinux.org/title/Howdy, section 2.1.

### 2. Confirm `sudo` with the `Enter` key
Currently, Howdy is activated automatically the moment sudo authentication is requested. I think this could be a bit insecure, because when running scripts or installing packages, I don't have a chance to stop sudo authentication. It just recognizes my face, and sometimes I don't even notice.

The alternative would be adding the same line as above to `sudo` pam config:

`auth      sufficient pam_unix.so try_first_pass likeauth nullok`

This would require pressing `Enter` to activate face unlock in `sudo`.

I'm wondering whether this would be clear to the user, but I think so, as it's consistent with the `hyprlock` screen (you need to press `Enter` there too).

Either way, I think it's worth discussing.

### 3. Consistency

#### a. Separate scripts
Last week, DHH split setup and remove flows into separate scripts. I think we should make this consistent by doing the same with Howdy: https://github.com/basecamp/omarchy/pull/5670

#### b. hyprlock icon
We could consider adding a face unlock icon to the `hyprlock` screen to make it consistent with fingerprint unlock.

---

Anyway, I'd be happy to help with the implementation once we reach a conclusion on the points above.

mateuszkowalczyk:
@Woogy7 could you take a look at my suggestions when you have a chance? I'd also be happy to continue working on this feature if you're ok with that

psynyde:
> @Woogy7 could you take a look at my suggestions when you have a chance? I'd also be happy to continue working on this feature if you're ok with that

I've implemented your suggestions for sudo and polkit.
worked as expected and i won't have to worry about polkit completely breaking when howdy doesn't work

Woogy7:
Hi @mateuszkowalczyk, really appreciate the thoughtful review, all three points make sense.

**1. Password fallback on hyprlock** : Agree. The current stack does fall through to `pam_unix` after howdy fails, but your ordering means a typed password authenticates instantly without sitting through the howdy timeout. Better UX, will adopt.

**2. Sudo `Enter` confirmation** : Agree. Silent face auth on every sudo prompt is a real footgun for scripts and unattended package installs. @psynyde just confirmed it works cleanly, and as a bonus he noticed it prevents polkit from completely breaking when howdy is unavailable, graceful degradation we definitely want.

**3a. Split setup/remove**: Yes, will mirror DHH's pattern from #5670. Looks like the convention is now `omarchy-setup-security-<name>` / `omarchy-remove-security-<name>` (matching fido2 and fingerprint), so this becomes `omarchy-setup-security-howdy` and `omarchy-remove-security-howdy`.

**3b. hyprlock face icon**: Good consistency touch, happy to add alongside the fingerprint one.

I'd love to take you up on the collaboration offer, this is my first PR, appreciate any help!

Woogy7:
Just replied above — apologies for the delay on this, thanks for the nudge. Your suggestions are all going in.

Woogy7:
@psynyde Thanks for testing it! Good to hear the sudo + polkit reorder behaves as expected. Adding these to the PR shortly.

Woogy7:
Pushed three follow-up commits addressing the review.

**What changed**

1. **Split into setup + remove scripts** (d92251ee), following DHH's pattern from #5670. Old `omarchy-setup-howdy` is now `omarchy-setup-security-howdy` and `omarchy-remove-security-howdy`. Also dropped the SDDM PAM block, since fingerprint and fido2 don't touch SDDM and most Omarchy installs autologin past it. The remove script still iterates `/etc/pam.d/sddm` so users upgrading from the original howdy script get cleaned up.

2. **Password fallback ordering** (8656e98d), implementing the `pam_unix.so try_first_pass likeauth nullok` suggestion in `/etc/pam.d/sudo`, `/etc/pam.d/hyprlock`, and `/etc/pam.d/polkit-1`. A typed password authenticates instantly via pam_unix; an empty Enter falls through to howdy. Sudo no longer face-authenticates silently on unattended commands.

3. **Face icon on hyprlock placeholder** (04312990), using U+F0C7B (face-recognition glyph from MDI). When the fingerprint icon is already in the placeholder, the face glyph is appended alongside rather than overwriting; remove preserves the fingerprint icon symmetrically.

**Tested on my Zenbook 14 UX3405CA**

* sudo: empty Enter triggers howdy; typed password skips howdy and autuhenticates via pam_unix
* hyprlock: same behaviour, blank enter uses howdy, entering the password and then pressing enter authenticates without using howdy; face glyph renders correctly in the placeholder
* polkit (1Password):  same behaviour, blank enter uses howdy, entering the password and then pressing enter authenticates without using howdy.
* remove script cleans up all four PAM files (including legacy sddm) and disables the IR service

@mateuszkowalczyk @psynyde please review the changes when you can.

mateuszkowalczyk:
@Woogy7 Thanks a lot for your effort! I'll try to review the changes soon.
In the meantime, I noticed that hyprlock is being replaced with quickshell (as part of broader Omarchy migration to quickshell), so this will require some more changes to integrate howdy into it too: https://github.com/basecamp/omarchy/commit/7ea4e1ab0436ec604f0f1f400913b8fa306aea84

Woogy7:
@mateuszkowalczyk Thanks! Ah I see, ok will test that change when it arrives and adapt as needed.

bin/omarchy-setup-security-howdy
```diff
@@ -0,0 +1,131 @@
...
+  done
+  return 1
+}
+
+setup_pam_config() {
+  # pam_unix.so (try_first_pass, nullok) sits above pam_howdy.so so that a typed
+  # password authenticates instantly via pam_unix; an empty Enter falls through
+  # to howdy. This keeps the password path fast and avoids howdy firing silently
+  # on every sudo invocation (e.g. unattended scripts).
+  if ! grep -q "pam_unix.so try_first_pass likeauth nullok" /etc/pam.d/sudo; then
```
mateuszkowalczyk: I'm not sure about this check here, because if someone happens to edit this file manually and slightly change formatting (like adding more spaces between `pam_unix.so` and `try_first_pass`) it won't work as expected, maybe we could use something like below:

`"pam_unix.so.*try_first_pass.*likeauth.*nullok"`

(also applies to the similar checks below)

bin/omarchy-setup-security-howdy
```diff
@@ -0,0 +1,131 @@
...
+    echo "Configuring sudo for face authentication..."
+    sudo sed -i '1i auth    sufficient pam_unix.so try_first_pass likeauth nullok' /etc/pam.d/sudo
+  fi
+  if ! grep -q "pam_howdy.so" /etc/pam.d/sudo; then
```
mateuszkowalczyk: Another thing to consider: the lines with `try_first_pass` and `pam_howdy.so` are checked independently here, but maybe we should check the `pam_howdy.so` first and only if it's missing check the other one? 
In other words: if someone already have howdy configuration in place we should probably leave it as is and skip `try_first_pass` check

(this comment also applies to all other files below)

bin/omarchy-remove-security-howdy
```diff
@@ -0,0 +1,50 @@
...
+
+remove_pam_config() {
+  # /etc/pam.d/sddm is included for users upgrading from the original howdy script,
+  # which configured SDDM as well; the new setup script no longer touches sddm.
+  for file in /etc/pam.d/sudo /etc/pam.d/hyprlock /etc/pam.d/polkit-1 /etc/pam.d/sddm; do
+    [[ -f $file ]] || continue
+    if grep -q 'pam_howdy.so' "$file"; then
+      echo "Removing face authentication from $(basename "$file")..."
+      sudo sed -i '/pam_howdy\.so/d' "$file"
+      sudo sed -i '/pam_unix\.so try_first_pass likeauth nullok/d' "$file"
```
mateuszkowalczyk: Check my comment from the other file, I think we should make this regex broader to also accept different number of spaces between parts: https://github.com/basecamp/omarchy/pull/5212#discussion_r3275841294

mateuszkowalczyk (commented):
@Woogy7 I've reviewed your changes and found a couple of edge cases to consider (see comments below)

bin/omarchy-setup-security-howdy
Woogy7: Good catch. The whitespace-literal checks are gone now: the guard keys on `pam_howdy.so` instead (see the next thread), and the remaining `try_first_pass` references in the `sed` addresses use the broadened `pam_unix.so.*try_first_pass.*likeauth.*nullok` pattern, so a manually reformatted PAM file is still detected. Same broadening applied to the remove script. Pushed in 018dd4a2.

bin/omarchy-setup-security-howdy
Woogy7: Agreed, the independent checks could leave partial state. Restructured so `pam_howdy.so` is the single guard per file: if face auth is already configured the file is left untouched. When it is absent, any stray `pam_unix.so try_first_pass` line is cleared first, then both lines are written as a clean pair, so an interrupted previous run cannot produce duplicates. Applied to sudo, hyprlock, and polkit. Verified against all four file states (fresh, already configured, howdy-only, interrupted run). Pushed in 018dd4a2.

bin/omarchy-remove-security-howdy
Woogy7: Done, the deletion regex is now `pam_unix.so.*try_first_pass.*likeauth.*nullok`, matching the setup script. Pushed in 018dd4a2.

bin/omarchy-setup-security-howdy
```diff
@@ -24,29 +24,29 @@ setup_pam_config() {
...
+  # leave the file untouched. When absent, any stray pam_unix.so try_first_pass
+  # line is cleared first so the two lines are always written as a clean pair.
+  if ! grep -q 'pam_howdy.so' /etc/pam.d/sudo; then
     echo "Configuring sudo for face authentication..."
+    sudo sed -i '/pam_unix\.so.*try_first_pass.*likeauth.*nullok/d' /etc/pam.d/sudo
     sudo sed -i '1i auth    sufficient pam_unix.so try_first_pass likeauth nullok' /etc/pam.d/sudo
-  fi
-  if ! grep -q "pam_howdy.so" /etc/pam.d/sudo; then
-    sudo sed -i '/pam_unix.so try_first_pass likeauth nullok/a auth    sufficient pam_howdy.so' /etc/pam.d/sudo
+    sudo sed -i '/pam_unix\.so.*try_first_pass.*likeauth.*nullok/a auth    sufficient pam_howdy.so' /etc/pam.d/sudo
```
mateuszkowalczyk: All good now, just a small style suggestion: I think that inserting 2 lines at once would make the code a bit more readable. So instead of this line and the previous one, we could just use:
```bash
    sudo sed -i '1i auth    sufficient pam_unix.so try_first_pass likeauth nullok\
auth    sufficient pam_howdy.so' /etc/pam.d/sudo
```

Of course the same applies to the other files below.

mateuszkowalczyk (commented):
Great, looks good now, I've just added a small style suggestion.

I haven't tested the entire setup flow yet, I'm going to wait until the quickshell version is implemented, but that should be straightforward once it's ready for fingerpring auth.

bin/omarchy-setup-security-howdy
Woogy7: Nice, that reads much better. Adopted in f0e14fc7 for sudo, hyprlock, and polkit. The defensive pre-delete is kept above the `1i` to clear any stray `pam_unix.so try_first_pass` line from a previous interrupted run. Logic tests still pass, including an explicit check that the atomic insert puts `pam_unix.so` on line 1 and `pam_howdy.so` on line 2.

Agreed on waiting for the quickshell version to land before testing the full flow.

Woogy7:
## Field report: silent breakage after a library soname bump

After two months of daily use, I hit a failure mode worth documenting. Following a routine system update, face unlock silently stopped working everywhere (sudo, hyprlock, polkit) and fell back to password auth with no error shown anywhere.

**Root cause:** the libjxl 0.11 to 0.12 update changed its soname. `python-dlib` (AUR, pulled in by `howdy-git`) was compiled against the old library and is never rebuilt by regular updates, so `import dlib` fails inside howdy's `compare.py`. Since `pam_howdy` is `sufficient` in the PAM stack, the failure is invisible: auth just falls through to password.

**Diagnosis:** `python -c "import dlib"` throws `ImportError: libjxl.so.0.11: cannot open shared object file`.

**Fix:** rebuild the package, e.g. `yay -S --rebuild python-dlib`. (Heads up: the current AUR PKGBUILD defaults to also building a CUDA variant, which drags in the multi-GB cuda/cudnn toolchain. Setting `_build_cuda=0` in the PKGBUILD avoids that on non-NVIDIA machines.)

**Mitigation added in 519280f:** the setup script now installs a pacman PostTransaction hook that import-checks dlib after updates to any of its dependency stack and prints a clear warning with the rebuild command when it breaks. The remove script cleans the hook up. This turns a silent auth downgrade into an actionable message in the pacman output.

This will recur for any AUR-built native Python module whenever a linked library bumps its soname, so the hook seems worth having from day one.

Woogy7:
## Updated for Omarchy 4.0 and the Quickshell lock screen

@mateuszkowalczyk @psynyde the quickshell version is here. After updating to 4.0 face unlock kept working for sudo and polkit but went dead on the lock screen, since `/etc/pam.d/hyprlock` is gone and the new lock screen talks to its own PAM services. This branch is now rebased on `quattro` (v4.0.0) and the PR base is retargeted there, because that is where the Quickshell lock lives and where every recent PR is going.

**Lock screen** (0414bb21): a third `PamContext` on a new `/etc/pam.d/omarchy-lock-face` service, next to `omarchy-lock-password` and `omarchy-lock-fingerprint`. It runs one bounded scan when the lock becomes secure, one more when you wake a blanked screen, and one on an empty Enter. I deliberately did not copy the fingerprint retry loop: howdy owns the IR camera and CPU for the whole scan, so a loop would drain a locked laptop. Only the Enter-triggered scan shows "Face not recognized"; the automatic ones stay silent like an untouched fingerprint reader. Typing a password during a scan still works through its own PAM context. The field gets a face glyph in the placeholder colour where the fingerprint icon sits (left of it when both are enrolled), brighter while scanning.

**Setup and remove** (e56b00b0): dropped all the hyprlock bits (PAM edits, placeholder glyph, `ignore_empty_input`). PAM is now written last, after `howdy add` succeeds, matching the fingerprint script's reasoning. The setup also pins `workaround = off` in howdy's config: with any other value `pam_howdy` opens a password prompt inside the same PAM conversation and then tries to dismiss it by injecting an Enter key through uinput, which the lock screen cannot answer. Both scripts get Setup/Remove > Security > Face Unlock menu entries, and the manual's hardware authentication page has a face unlock section.

**Tests**: added `test/shell.d/lock-face-indicator-test.sh` (indicator visibility, the reserve with one and two icons, empty Enter requesting a scan only when a model is enrolled). It fails against the previous `LockView.qml` and passes with this branch. `./test/cli` and the lock and menu shell tests pass.

**Verified on my Zenbook 14 (4.0.0)**: lock, look at the camera, unlocked 1.5 s later with no keypress. Wake after blank rescans. Empty Enter rescans and reports failure in the field. Password still works. sudo and 1Password behave as before.

If either of you can run the setup on your machines (UX5406AA, X1 Carbon) that would be the last box on the test plan. Thanks again for the patience on this one.

Woogy7:
Follow-up after a day of use (69c22a49): scanning the moment the lock became secure was wrong. Press Super+Ctrl+L, look at the screen for half a second, and you are unlocked again. The person who locks the screen is still at the keyboard and has not asked to be scanned.

Face scans are now driven by input alone. The first key or pointer event on the lock surface after a quiet gap (5 s) starts one silent scan, which covers coming back from a break, opening the lid, and waking from sleep (the wall clock jumps across suspend, so the gap is always open then). Events inside the gap, like typing a password or nudging the mouse right after locking, do not scan, and the lock keystroke itself counts as input. An empty Enter still asks explicitly. Added `test/shell.d/lock-face-trigger-test.sh` to pin those rules, and updated the manual wording. Verified on the Zenbook: lock, nothing happens, come back a few seconds later, first mouse move scans and unlocks.

Woogy7:
One more (4ba657c5): the polkit dialog now shows that face unlock is there. Until now it worked (empty Enter scanned) but the dialog looked exactly like a password-only prompt, so a new user had no reason to press Enter on an empty field.

The password field ends in a dim face glyph whenever `pam_howdy` is in `/etc/pam.d/polkit-1`, in the same spot and with the same glyph as the lock screen. While an empty submit is in flight it lights up in the accent and the placeholder reads "Scanning..." instead of "Checking..."; a failed scan flashes "Not recognized" instead of "Wrong". Face deliberately does not take over the dialog the way fingerprint mode does, because `pam_unix` stays first and a typed password must keep working without waiting on the camera. `PolkitModel.faceConfiguredFromPamConfig` mirrors the fingerprint detection, with tests.

Woogy7:
Small design note added to the description, since it came up while testing: face scans always follow an intent signal (arriving at the lock screen, or an empty Enter) and never fire on the lock keystroke or on a polkit dialog opening. Face is a passive biometric and a polkit prompt can come from any process, so the empty Enter is the one-keystroke consent that stops a rogue `pkexec` from being authorised just because someone is at the screen. That is the same reasoning as the `pam_unix`-first ordering for `sudo` from the earlier review. Polkit also has a single PAM conversation per request, so an automatic empty submit would block typing a password for the length of the scan.

From my side the branch is complete: lock screen, sudo, polkit (with the face hint), setup and remove scripts, menu entries, manual, and tests, all on top of `quattro`. Ready for review whenever someone has a moment.

test/shell.d/lock-face-trigger-test.sh
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash
+
+set -euo pipefail
+
+source "$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd)/base-test.sh"
+
+run_node_test <<'JS'
+const fs = require('fs')
```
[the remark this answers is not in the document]
Woogy7: `path` is declared by the `run_node_test` prelude in `test/shell.d/base-test.sh` (`const path = require("path")`, alongside `root`), which is why the sibling tests such as `lock-blank-fingerprint-test.sh` and `plugins-test.sh` use bare `path.join(root, ...)` too. The test runs green under `./test/shell`.

shell/plugins/lock/Service.qml
```diff
@@ -279,8 +359,9 @@ Item {
         passwordText: root.enteredPassword
         onPasswordTextEdited: function(password) { root.enteredPassword = password }
         onSubmitPassword: function(password) { root.submitPassword(password) }
+        onSubmitFace: root.submitFace()
         onClearFailureRequested: root.failureMessage = ""
-        onWakeRequested: root.runWake()
+        onWakeRequested: root.noteInput()
```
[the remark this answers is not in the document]
Woogy7: `onSignal: expression` is the standard QML signal-handler form: the expression is bound to the signal and only evaluated when it fires, not at object creation. The neighbouring `onWakeRequested: root.noteInput()` and `onClearFailureRequested: root.failureMessage = ""` in this file use the same form. Verified on hardware: locking the session produces no howdy activity until Enter or arrival input triggers a scan.

mateuszkowalczyk:
Hi @Woogy7, sorry for not letting you know earlier, but I’ve recently implemented the shell integration myself based on discussion here: https://github.com/basecamp/omarchy/discussions/4982

I think I managed to make the face unlock experience quite smooth, it works automatically after opening the lid as well as after any key press or cursor movement.

We also lean towards replacing howdy with more modern alternative (Facelock), it has less dependencies, uses newer face detection models and has nice setup process that leaves us with less to implement on Omarchy side

Could you have a look at the PR here:  https://github.com/basecamp/omarchy/pull/6863 ? The implementation is not finished yet, but I described the UX approach there so maybe you could give some feedback comparing it to your implementation?

psynyde:
> We also lean towards replacing howdy with more modern alternative (Facelock), it has less dependencies, uses newer face detection models and has nice setup process that leaves us with less to implement on Omarchy side

While facelock is *modern*, I'm not sure about its quality since it's primarily written using AI.
Even after this, I wanted to try out facelock, but their docs say to install `onnxruntime-opt-openvino` from the arch repos, whereas not even the aur ever had any package similar to that.

image

I suggest we replace howdy with [howdy-next](https://codeberg.org/nathawat/howdy-next) which I've been using for the past month. It's actively maintained and been working quite well on my system.
Also the author seems to be not pushing commits without reading what's being generated beforehand :)

Woogy7:
I'll be honest, it has taken me a while to get to the place where this works as smoothly as it does on my laptop, from face checking when lid opens to 1password face checks, all working really well I am hesitant to change it!

mateuszkowalczyk:
@Woogy7 @psynyde Since there are two parallel implementations going on, I'd suggest deciding which one we want to proceed with. I tried to summarize the situation here: https://github.com/basecamp/omarchy/discussions/4982#discussioncomment-18066454: . Could you take a look and share your thoughts there?

tyvsmith:
> > We also lean towards replacing howdy with more modern alternative (Facelock), it has less dependencies, uses newer face detection models and has nice setup process that leaves us with less to implement on Omarchy side
> 
> While facelock is _modern_, I'm not sure about its quality since it's primarily written using AI. Even after this, I wanted to try out facelock, but their docs say to install `onnxruntime-opt-openvino` from the arch repos, whereas not even the aur ever had any package similar to that.
> 
> image
> I suggest we replace howdy with [howdy-next](https://codeberg.org/nathawat/howdy-next) which I've been using for the past month. It's actively maintained and been working quite well on my system. Also the author seems to be not pushing commits without reading what's being generated beforehand :)

Just stumbled on this, thanks for flagging that docs issue, @psynyde! Unfortunately that was a holdover from the earliest project prototypes that hadn't gotten revisited yet. Facelock is a heavily agent engineered project, but with a lot of human time, review, and manual coding too, with robust deterministic validation and guardrails, with the goal to make high quality, stable, and performant. I just launched the 0.2.0 release which had a ton of improvements, security work, and hardening, including documentation generation/testing to minimize any issues like you mentioned. Always welcome issues filed if you run into anything else.  Thanks!

bjarneo:

PR #5223, opened 2026-04-04
Add zsh support with aliases and shell config

markbus-ai:
## Summary
- Add zsh configuration with aliases, envs, functions, init, rc, and shell config
- Fix zsh compatibility issues with HISTCONTROL and arithmetic syntax


morgoth:
zsh integration is provided via https://github.com/omacom-io/omarchy-zsh

PR #5225, opened 2026-04-04
Fix Super+Ctrl+Backspace to cycle monitor scaling

markbus-ai:
## Summary
- Fix Super+Ctrl+Backspace to cycle monitor scaling between 1x, 1.5x, 2x

## Changes
- default/hypr/bindings/utilities.conf: Update binding description


dhh:
The correct key is `Super + /` and 1.6 is the best scaling for 4k at 27".

PR #5226, opened 2026-04-04
Added theme agnostic background directory

KR9SIS:
Made omarchy_background_selecor.lua also search for background images outside of the current theme directory.

This allows an easy way to add a background to all themes.


dhh:
Closing because this implementation lives entirely in the removed Elephant background selector. A theme-agnostic directory would need to be added to Quattro’s native background picker.

PR #5239, opened 2026-04-06
Restore jetbrains no_follow_mouse

NicolasDorier:
This reverts partially https://github.com/basecamp/omarchy/pull/5183, due to the [following issue](https://github.com/basecamp/omarchy/pull/5183#issuecomment-4189299971) reported by @vmarcinko

https://github.com/user-attachments/assets/624c61c5-6910-427d-8f23-2b774fec9d6e

I am not sure if we should merge that immediately for two reasons:
1. The new lua config change of hyprland will break it.
2. It is unclear whether everybody prefer to fix the behavior of `Find in Files` better than losing the follow mouse focus.

But if it's not merged, it might still help somebody.

### DIFF
--- a/default/hypr/apps.conf
+++ b/default/hypr/apps.conf
@@ -3,6 +3,7 @@ source = ~/.local/share/omarchy/default/hypr/apps/1password.conf
 source = ~/.local/share/omarchy/default/hypr/apps/bitwarden.conf
 source = ~/.local/share/omarchy/default/hypr/apps/browser.conf
 source = ~/.local/share/omarchy/default/hypr/apps/hyprshot.conf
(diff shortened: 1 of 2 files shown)


NicolasDorier:
Closing this one, as it is unclear we want this.

jaoxford:
This fix resolved my issues!

dimitrismistriotis:
... and mine; Some people use Jetbrains, very relevant

NicolasDorier:
Reopening due to user feedback.

@dhh I think it can be merged, but since hyprland is in the process of migrating to lua config, it's ok for me to just rewrite this PR when omarchy is migrated.

dhh:
Let's solve problems we have today with the best efforts we have. Then we can update when things change.

PR #5253, opened 2026-04-07
Add rotating backgrounds feature toggle

dabit:
Some themes ship with multiple beautiful backgrounds and choosing just one means missing out on the rest. This PR adds background cycling as a first-class Omarchy feature so users can enjoy all of them automatically.

Accessible via the SUPER+CTRL+O trigger menu → Toggle → Cycle Backgrounds.

  When enabled, a systemd user timer fires `omarchy-theme-bg-next` hourly to
  automatically cycle through the active theme's backgrounds. The timer runs
  inside the user session, ensuring `WAYLAND_DISPLAY` and `DBUS_SESSION_BUS_ADDRESS`
  are available (unlike cron). `KillMode=none` is set on the service so systemd
  doesn't kill the newly launched `swaybg` process when cleaning up the cgroup.

  The feature is opt-in — nothing is auto-enabled on install.

  ## Changes

  - `bin/omarchy-toggle-bg-cycle` — enables/disables `omarchy-bg-cycle.timer` via `systemctl --user`, with notify-send feedback
  - `config/systemd/user/omarchy-bg-cycle.{service,timer}` — oneshot service + hourly timer unit files (deployed to `~/.config/systemd/user/` on install)
  - `bin/omarchy-menu` — adds Cycle Backgrounds entry to the Toggle submenu

image

### DIFF
--- a/bin/omarchy-menu
+++ b/bin/omarchy-menu
@@ -211,7 +211,7 @@ show_share_menu() {
 }
 
 show_toggle_menu() {
-  local options="󱄄  Screensaver\n󰔎  Nightlight\n󱫖  Idle Lock\n󰂛  Notifications\n󰍜  Top Bar\n󱂬  Workspace Layout\n  Window Gaps\n  1-Window Ratio\n󰍹  Monitor Scaling\n  Direct Boot\n󰟵  Passwordless Sudo"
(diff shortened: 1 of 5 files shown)


dhh:
Closing in favor of the broader dynamic background implementation in #6129.

PR #5258, opened 2026-04-08
Fix: Super+C copy not working in neovim

### LINKED ISSUE #5251: Can't copy in neovim using Super+C
### System details

Omarchy 3.5.0

### What's wrong?

I am sure it was possible before, but now only Super+V works. This happens only in neovim, not everywhere else.
I've tested it with and without `--clean` on versions 0.11 and 0.12. 

Please confirm, maybe I broke something and need to reset.

mplaczek99:
## Summary
Fixes #5251

- Terminal emulators (kitty, alacritty, ghostty) were intercepting `Ctrl+Insert` (sent by Hyprland for `Super+C`) before neovim could receive it, so nothing was copied
- Removed the `Ctrl+Insert → copy` interception from kitty and alacritty so the keystroke passes through to neovim
- Explicitly unbound `Ctrl+Insert` in ghostty (deleting the config line isn't enough since ghostty has a built-in default for `ctrl+insert=copy_to_clipboard` that would still intercept the key)
- Added a neovim keymap (`<C-Insert>` → `"+y` in visual mode) so neovim copies the visual selection to the system clipboard
- Includes migration for existing installations

## Test plan
- [ ] Open neovim in kitty/ghostty/alacritty
- [ ] Select text in visual mode (`v` + movement)
- [ ] Press `Super+C`
- [ ] Paste with `Super+V` in another app — should contain the selected text
- [ ] Verify `Super+C` still works outside terminals (browser, file manager, etc.)
- [ ] Verify `Super+V` paste still works everywhere

### DIFF
--- a/config/ghostty/config
+++ b/config/ghostty/config
@@ -24,6 +24,7 @@ shell-integration-features = no-cursor,ssh-env
 
 # Keyboard bindings
 keybind = shift+insert=paste_from_clipboard
+keybind = control+insert=unbind
 keybind = super+control+shift+alt+arrow_down=resize_split:down,100
 keybind = super+control+shift+alt+arrow_up=resize_split:up,100
 keybind = super+control+shift+alt+arrow_left=resize_split:left,100
--- a/config/ghostty/config
+++ b/config/ghostty/config
@@ -24,7 +24,7 @@ shell-integration-features = no-cursor,ssh-env
 
 # Keyboard bindings
 keybind = shift+insert=paste_from_clipboard
-keybind = control+insert=copy_to_clipboard
+keybind = control+insert=unconsumed:copy_to_clipboard
 keybind = super+control+shift+alt+arrow_down=resize_split:down,100
 keybind = super+control+shift+alt+arrow_up=resize_split:up,100
 keybind = super+control+shift+alt+arrow_left=resize_split:left,100
--- a/config/ghostty/config
+++ b/config/ghostty/config
@@ -24,7 +24,6 @@ shell-integration-features = no-cursor,ssh-env
 
 # Keyboard bindings
 keybind = shift+insert=paste_from_clipboard
-keybind = control+insert=unconsumed:copy_to_clipboard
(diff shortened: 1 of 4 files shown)


migrations/1775666877.sh
```diff
@@ -0,0 +1,21 @@
+echo "Fix Super+C copy in neovim by letting Ctrl+Insert pass through to the terminal app"
+
+# Remove ctrl+insert copy binding from Kitty (let the key pass through to neovim)
+sed -i '/map ctrl+insert copy_to_clipboard/d' ~/.config/kitty/kitty.conf
+
+# Change Ghostty ctrl+insert to unconsumed: so it passes through to neovim when there's no terminal selection
+sed -i 's/control+insert=copy_to_clipboard/control+insert=unconsumed:copy_to_clipboard/' ~/.config/ghostty/config
+
+# Remove ctrl+insert copy binding from Alacritty
+sed -i '/mods = "Control", action = "Copy"/d' ~/.config/alacritty/alacritty.toml
```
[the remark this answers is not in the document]
heymaikol: I think the terminals are already installed...

Metsker:
Works on alacritty, but not on ghostty for me. 
Ghostty configuration should be `keybind = unconsumed:control+insert=copy_to_clipboard` or it will cause error.
But regardless, with or without this line it didn't worked

config/ghostty/config
```diff
@@ -24,7 +24,6 @@ shell-integration-features = no-cursor,ssh-env
 
 # Keyboard bindings
 keybind = shift+insert=paste_from_clipboard
-keybind = control+insert=copy_to_clipboard
```
Metsker, line 27: It won't work, because there is the same default keybind. We can use `keybind = control+insert=unbind
` in this regard and it will work, but I feel like it's not a desirable fix.

heymaikol: Wym, I confirmed it on all 3 terminals: kitty, alacritty, and ghostty

PR #5262, opened 2026-04-08
add i2c_hid modules to initramfs for laptops with I2C HID keyboards

### LINKED ISSUE #5261: Keyboard disabled before login on Acer Nitro 16
### System details

AMD Ryzen AI 9 365, Nvidia RTX 5070 Max-Q, 32GB, Omarchy 3.5.0

### What's wrong?

When booting up, the kernel module `i2c_hid`, isn't loaded, making the built-in keyboard of my Acer Nitro 16 non functional. Plugging in an external keyboard works fine.
```
$ lsinitcpio -a /boot/EFI/Linux/omarchy_linux.efi | grep i2c_hid
(no output)
```
I tried adding the `i2c_hid i2c_hid_acpi` modules to `/etc/mkinitcpio.conf`, and that fixed the issue.
```
MODULES=(nvidia nvidia_modeset nvidia_uvm nvidia_drm btrfs i2c_hid i2c_hid_acpi)
```
then
```
$ sudo mkinitcpio -P
```

I can't paste the `omarchy-debug` here as it is too long and the uploading has been disabled.

g1sbi:
Closes #5261

### DIFF
--- a/install/config/all.sh
+++ b/install/config/all.sh
@@ -58,6 +58,7 @@ run_logged $OMARCHY_INSTALL/config/hardware/apple/fix-suspend-nvme.sh
 run_logged $OMARCHY_INSTALL/config/hardware/apple/fix-t2.sh
 
 run_logged $OMARCHY_INSTALL/config/hardware/fix-bcm43xx.sh
+run_logged $OMARCHY_INSTALL/config/hardware/fix-i2c-hid-keyboard.sh
 run_logged $OMARCHY_INSTALL/config/hardware/fix-surface-keyboard.sh
 run_logged $OMARCHY_INSTALL/config/hardware/fix-yt6801-ethernet-adapter.sh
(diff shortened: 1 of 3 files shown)


ahmedkall:
I tested this on an Acer Swift SF16-71T (Panther Lake). Adding only i2c_hid i2c_hid_acpi was not sufficient: the initramfs log showed i2c i2c-10251825:00: deferred probe pending: i2c_hid_acpi: can't get irq, and the built-in keyboard was unavailable at the LUKS prompt. Adding pinctrl_intel_platform before the I²C HID modules made the built-in keyboard work reliably at cold boot:

MODULES+=(pinctrl_intel_platform i2c_hid i2c_hid_acpi)

The change was verified in the Limine UKI and tested with a cold boot. Since this is Panther Lake-specific, it may be preferable to gate pinctrl_intel_platform rather than add it to every system. Also, MODULES+= is important to avoid overwriting modules added by other hardware detection scripts.

PR #5270, opened 2026-04-10
Add Little Snitch for Linux to Install > Service menu

marcelzimmer:
## Summary
- Add [Little Snitch for Linux](https://obdev.at/products/littlesnitch-linux/) as an installable service
- Little Snitch for Linux was released on April 8, 2026 — it's a network monitor and application firewall that shows which processes make network connections and lets you block them with rules
- Written in Rust, uses eBPF for kernel-level traffic interception, web-based UI on localhost:3031
- Requires kernel 6.12+ with BTF support (which Omarchy has)
- Free to use, eBPF component and UI are open source

## Changes
- New install script `bin/omarchy-install-littlesnitch` (downloads .pkg.tar.zst from obdev.at, enables the service)
- Menu entry in `Install > Service` alongside Tailscale, NordVPN, etc.

### DIFF
--- a/bin/omarchy-install-littlesnitch
+++ b/bin/omarchy-install-littlesnitch
@@ -1,0 +1,11 @@
+#!/bin/bash
+
+# Install Little Snitch for Linux, an application firewall and network monitor.
+
+echo "Installing Little Snitch..."
+omarchy-pkg-aur-add littlesnitch-bin
+
+echo "Enabling Little Snitch service..."
+sudo systemctl enable --now littlesnitch
+
+echo -e "\nLittle Snitch installed! Open http://localhost:3031 to access the web UI."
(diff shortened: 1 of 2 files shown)


jondkinney:
Looking forward to this!

marcelzimmer:
Thank you.

I’ve always had Little Snitch on my MacBook, I think since 2004. Almost every Mac user knows it. I’ve missed it so much since switching to Omarchy Linux last summer. I bet many other people who made the switch will be really happy that Little Snitch is now available for Linux too.

We just need to make sure we don’t accidentally block Fizzy or any other web app. Maybe we should add a few exception rules?

marcelzimmer:
I’ve revised the Pull Request again. I had initially included a hardcoded version number, but I then noticed that Omarchy Linux does not hardcode it for the other apps. Therefore, I’ve implemented it identically.

PR #5271, opened 2026-04-10
Toggle waybar clock format from keyboard

Wessel-Boers:
From the Omarchy manual: "_Everything in Omarchy happens via the keyboard — EVERYTHING!_" 

Well... not quite everything! When online, sometimes you want to quickly check what the date is while filling in a form. Currently, you're forced to touch the mouse (😔) to click on the clock. This PR adds a keyboard shortcut for toggling the waybar clock on `SUPER CTRL + D`.

Combination with CTRL key to be consistent with the other toggles. "D" is not used yet and can represent "Date". 

This toggle is also added to the Toggle Menu on `SUPER CTRL + O`.

### DIFF
--- a/bin/omarchy-menu
+++ b/bin/omarchy-menu
@@ -168,12 +168,13 @@ show_share_menu() {
 }
 
 show_toggle_menu() {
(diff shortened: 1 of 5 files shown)


dhh:
This is pretty clever, but I also think potentially rather brittle. Munging existing configs live for something like this is too much, I think. I'd rather we try to fix this upstream in waybar. Have a general way to either register a signal or some other way to trigger a simulated mouse click. Please do investigate that path!

Wessel-Boers:
Thanks. I have opened an issue and will try to get something merged in Waybar. Till then!

PR #5273, opened 2026-04-10
First commit: Add Nautilus Copy Path context menu integration

manish-12ys:
First commit: Add Nautilus Copy Path context menu integration

This commit introduces a Nautilus (Files) context menu extension that allows users to quickly copy the full path of selected files and folders directly from the right-click menu. The implementation includes the necessary configuration scripts, extension logic, and migration setup to ensure smooth integration.

The goal is to improve everyday workflow efficiency on Linux by providing a simple yet practical utility that saves time when working with file paths in development and system tasks.

As this is my first commit, I sincerely apologize for any mistakes or inconsistencies. Feedback and suggestions are always welcome.

### DIFF
--- a/default/nautilus-python/extensions/copy_path.py
+++ b/default/nautilus-python/extensions/copy_path.py
@@ -1,0 +1,80 @@
+from gi import require_version
+
+require_version("Nautilus", "4.1")
+require_version("Gdk", "4.0")
+
+from gi.repository import GObject, Gdk, Nautilus
+
+
+class CopyPathAction(GObject.GObject, Nautilus.MenuProvider):
+    def _copy_to_clipboard(self, value):
+        display = Gdk.Display.get_default()
+        if not display:
+            return
+
+        clipboard = display.get_clipboard()
+        clipboard.set(value)
+
+    def _copy_file_paths(self, files):
+        paths = []
+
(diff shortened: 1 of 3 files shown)


patrick-boc:
Nautilus already has this feature. If you simply copy a file and paste it into a text editor/field, it will result in the path being pasted.

https://github.com/user-attachments/assets/8ef2495e-7d0c-4743-8bb0-1bd57cf7562a

manish-12ys:
Thanks, I wasn’t aware of it before.

dhh:
Closing because Nautilus already supports copying a selected file or folder and pasting its filesystem path into a text field, so an additional Copy Path extension is not necessary.

nojaf:
Adding another real-world data point: on Omarchy 4.0.2 with GNOME Files/Nautilus 50.2.2, there is still no visible Copy Path/Copy Location action for a selected file. `Ctrl+L` only exposes the current directory, and `Ctrl+Shift+C` does not copy the selected file path as plain text.

The suggested workaround of copying the file and pasting it into a text field does not produce a usable full path in the fields I tried. An explicit context-menu action such as “Copy Path” would be much clearer and would copy `/home/user/folder/file.ext` directly to the clipboard.

PR #5280, opened 2026-04-11
Omarchy 3.5.1

dhh:
# What changed?

Install release candidate using the [3.5.1 RC ISO](https://iso.omarchy.org/omarchy-3.5.1-rc.iso) or switching to the RC channel under _Update > Channel > RC_, if you're willing to help test before the final release.

## Additions

- Add Dell XPS mute mic key support with hardware LED sync by @spencerbull, @dhh
- Add resume performance boost on Panther Lake systems to avoid sluggishness after wake from sleep by @dhh
- Add _Trigger > Toggle > Laptop Display_ to turn the laptop display on/off when using an external monitor by @dhh
- Add _Update > Hardware > Trackpad_ to restart i2c_hid_acpi and THC driver paths for XPS haptic touchpad losing hapticness (temporarily until there's a permanent fix) by @dhh
- Add Logitech MX Keys key binding examples for Print Screen, Dictation, and Emoji buttons to `~/.config/hypr/bindings.conf` on new installs by @joaovictor

## Fixes

- Fix Fcitx5 icon showing up in the waybar (again!) by using a proper XDG autostart override instead of deleting the system file by @pomartel
- Fix Intel Panther Lake display freezes and glitches by disabling both PSR and Panel Replay (`xe.enable_psr=0`) by @dhh
- Fix haptic touchpad detection to match any Synaptics product, not just one specific model, for broader Dell XPS compatibility by @dhh
- Fix spurious Hyprland config errors during git update by suppressing errors mid-pull and reloading after by @dhh
- Fix double sudo prompt during updates by moving the PTY logging wrapper earlier in the update flow by @dhh
- Fix Intel LPMD service not being enabled after installation by @dhh
- Fix power profile rules running on every update instead of only on first run when the daemon is available by @dhh
- Fix npx wrappers to run through mise-managed Node instead of system Node by @timohubois
- Fix starship prompt init for dumb bash sessions (e.g. scp, rsync) by @pipetogrep
- Fix display scaling cycle notification warnings by removing the Hyprland scale notification permanently by @dhh
- Fix Chromium default appearance to follow system (device) instead of dark by @dhh
- Fix btop.conf for v1.4.6 by @stefangrundel
- Fix hibernation setup gracefully handling a failed btrfs resume offset lookup by @dhh
- Fix Chromium/Brave stderr noise during theme switch by @felipegonzalez
- Fix AC power udev rules failing silently at boot by wrapping them with systemd-run by @annoyedmilk
- Fix Intel video acceleration migration scripts referencing renamed path by @liammitchell
- Fix browser launcher not recognizing new Brave Origin by @dhh
- Fix sluggishness with the walker app launcher on some systems by making it use the cairo software renderer by @dhh
- Fix performance power profile should be used when booting on AC power by @dhh


spencerbull (changes requested):
https://github.com/basecamp/omarchy/pull/5280/changes/3eee73c0deac546210cb65d92d4be32e8f06639b#diff-be025a5541a526dbd0bfab4b9cf4e64454f9cda1abf0a7ab2edabf1dd46a75d8

This migration is currently being applied to the XPS OLED and IPS, but the issue of panel lag with PSR is only affecting the OLED device.

These xe params should stay default (enabled) on the IPS for power optimizations if possible.

gekkom:
From the intel-lpmd readme:
> Please note that the installed configuration files serve as templates of best practices for specific platform models and disable lpmd by default. For LPMD to start the user is expected to either edit the main "intel_lpmd_config.xml" config file or after starting the program, enable LPMD by using the intel_lpmd_control tool.

So its not enough to start the service to make it work.

annoyedmilk:
Hey @dhh, the AC power udev rules fix was by me, my GitHub handle is @annoyedmilk, not @marco. Could you update the attribution? Thanks!

PR #5291, opened 2026-04-12
Fix double sudo password prompt during update

shawnyeager:
## Summary
- The `script(1)` wrapper in `omarchy-update-perform` creates a new PTY for logging, which invalidates sudo's credential cache from the original terminal
- This causes users to be prompted for their sudo password twice (once for keyring update, once for system packages)
- Fix: prime the sudo cache with `sudo -v` once inside the new PTY so all subsequent calls reuse it

## Test plan
- [ ] Run `omarchy-update` and confirm only one sudo password prompt appears


shawnyeager:
Closing — the root cause (script(1) PTY wrapper) has already been removed on dev, so this fix will be obsolete once dev merges to master.

PR #5292, opened 2026-04-12
Prevent empty custom themes from replacing the current theme

11me:
Empty user-only theme directories currently pass the existence check in `omarchy-theme-set` and can replace `~/.config/omarchy/current/theme` with an empty theme.

## Summary
- reject empty theme names after sanitization
- fail the theme switch if template generation fails
- write generated theme files through a temporary file before moving them into `next-theme`
- abort the swap if `next-theme` ends up empty
- keep the current theme untouched on failure

Empty overrides on top of an existing built-in theme continue to work as before.

## Testing
- switched between built-in themes
- applied an empty override on top of an existing built-in theme
- applied an empty user-only theme and verified that the active theme stayed unchanged
- tested in both a temp `HOME` and on a live Omarchy system

### DIFF
--- a/bin/omarchy-theme-set
+++ b/bin/omarchy-theme-set
@@ -12,6 +12,11 @@ OMARCHY_THEMES_PATH="$OMARCHY_PATH/themes"
 
 THEME_NAME=$(echo "$1" | sed -E 's/<[^>]+>//g' | tr '[:upper:]' '[:lower:]' | tr ' ' '-')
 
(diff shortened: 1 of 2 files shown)


dhh:
Appreciate catching this, but I don't think it's a case common enough to warrant the additional complication to deal with it. Just don't have empty user themes! And if you do, you can switch to another theme as you see it broken.

PR #5299, opened 2026-04-14
omarchy-reinstall-git: use existing remote instead of hardcoding

jondkinney:
## Problem

`omarchy-reinstall-git` hardcodes the clone URL:

```bash
git clone --depth=1 "https://github.com/basecamp/omarchy.git" ~/.local/share/omarchy-new
```

This means anyone running `omarchy-reinstall` on a fork (port for a different
arch, personal customizations, company-internal mirror, etc.) silently loses
their fork's customizations and ends up with upstream — without any warning.
It also implicitly clones the remote's default branch, so users on a non-default
branch lose that too.

## Fix

Read the clone URL from the existing `$OMARCHY_PATH` checkout's `origin` remote,
and the target branch from the current HEAD. Forks reinstall from themselves;
upstream users continue to reinstall from upstream (same behavior they get today).

Also converts SSH origin URLs to HTTPS for the clone step — reinstall is often a
recovery action and shouldn't depend on having an SSH agent/key configured.
Falls back to the upstream URL + `master` only if the existing install has no
usable git metadata.

## Testing

Verified on both an upstream `basecamp/omarchy` checkout and a fork
(`jondkinney/armarchy`, branch `amarchy-3-x`). Each now reinstalls from its
own origin and branch.

### DIFF
--- a/bin/omarchy-reinstall-git
+++ b/bin/omarchy-reinstall-git
@@ -2,7 +2,24 @@
 set -e
 
 # Reinstall the Omarchy configuration directory from the git source.
+#
+# URL and branch are derived from the existing $OMARCHY_PATH checkout so
+# any fork (arm ports, personal customizations, etc.) reinstalls from
+# the same remote it was running on. Falls back to upstream only if the
+# existing install has no usable git metadata.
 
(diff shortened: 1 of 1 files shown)


Kasui92 (approved):

dhh:
Closing because omarchy-reinstall-git no longer exists in the Quattro reinstall architecture.

PR #5327, opened 2026-04-16
Align Omarchy power profiles with intel-lpmd

spencerbull:
## Summary
- Route Omarchy-managed power profile changes through `omarchy-powerprofiles-set` instead of calling `powerprofilesctl` directly.
- Keep `intel-lpmd` synchronized with Omarchy power profile changes when `intel-lpmd` is installed and active.
- Extend that same synchronization to resume boost behavior and first-run default profile setup.
## Why
Omarchy already manages power profiles at several points:
- from the setup menu
- during first-run battery/AC initialization
- during temporary resume performance boosting
With `intel-lpmd` installed, changing the `power-profiles-daemon` profile alone is not enough. Omarchy also needs to keep `intel-lpmd` aligned with those profile transitions so the expected CPU-set behavior matches the selected mode.
This makes the three profile modes behave distinctly on supported Intel systems:
- `performance`:
  - full CPU set
  - `intel-lpmd` forced off
- `balanced`:
  - opportunistic low power mode
  - `intel-lpmd` auto
- `power-saver`:
  - strongest battery-saving mode
  - `intel-lpmd` forced on
On systems without `intel-lpmd`, Omarchy should continue to behave exactly as before.
## What Changed
- Updated `omarchy-menu` to call `omarchy-powerprofiles-set` instead of `powerprofilesctl set`
- Expanded `omarchy-powerprofiles-set` so it can:
  - accept `ac` / `battery` defaults
  - accept direct profile names
  - no-op safely when `intel-lpmd` is not installed or not active
- Updated first-run battery/AC setup to use `omarchy-powerprofiles-set`
- Updated the Intel `resume-boost` hook so the temporary resume performance boost and the later restore also keep `intel-lpmd` in sync
### Systems without `intel-lpmd`
- no functional change
- Omarchy still sets the appropriate `power-profiles-daemon` profile
### Systems with `intel-lpmd`
- `performance` maps to `LPM_FORCE_OFF`
- `balanced` maps to `LPM_AUTO`
- `power-saver` maps to `LPM_FORCE_ON`
## Boot / First-Run Behavior
Omarchy’s current default behavior remains:
- on AC:
  - prefer `performance`
- on battery:
  - prefer `balanced`
This PR ensures that when Omarchy sets those defaults, `intel-lpmd` is updated to the matching policy as well.
### Manual checks
- switched between `performance`, `balanced`, and `power-saver`
- confirmed `intel_lpmd.service` remained active
- confirmed expected `system.slice` cpuset behavior on Panther Lake:
  - `performance` -> `0-15`
  - `balanced` -> opportunistic low-power cpuset when lightly loaded
  - `power-saver` -> `12-15`
### Resume behavior
- resume boost still temporarily switches to `performance`
- after the delay, the previous profile is restored
- `intel-lpmd` is restored to the matching mode along with the power profile
## Notes
- This PR only changes Omarchy runtime/profile orchestration.
- The matching Panther Lake `intel-lpmd` packaged config is handled in the companion `omarchy-pkgs` https://github.com/omacom-io/omarchy-pkgs/pull/79

### DIFF
--- a/bin/omarchy-menu
+++ b/bin/omarchy-menu
@@ -247,7 +247,7 @@ show_setup_power_menu() {
   if [[ $profile == "CNCLD" || -z $profile ]]; then
     back_to show_setup_menu
   else
-    powerprofilesctl set "$profile"
+    omarchy-powerprofiles-set "$profile"
   fi
 }
 
--- a/bin/omarchy-powerprofiles-set
+++ b/bin/omarchy-powerprofiles-set
@@ -5,8 +5,6 @@
 #
 # Usage: omarchy-powerprofiles-set <ac|battery|performance|balanced|power-saver>
 
-OMARCHY_BIN=$(cd -- "$(dirname "$0")" && pwd)
-
 mapfile -t profiles < <(powerprofilesctl list | awk '/^\s*[* ]\s*[a-zA-Z0-9\-]+:$/ { gsub(/^[*[:space:]]+|:$/,""); print }')
 
 profile_available() {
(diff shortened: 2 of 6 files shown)


spencerbull:
Needs further testing. The suspend lag seems to be present again after this change.

PR #5328, opened 2026-04-16
fix(display): restore full monitor config when re-enabling internal display

noahterenzianii:
## What
When re-enabling the internal display, restore the original resolution,
scale and bit depth instead of falling back to preferred/auto defaults.

## Why
The previous `preferred,auto,auto` reset scale and bit depth on every toggle,
causing the display to come back with wrong settings.

## How
Read availableModes[0] and currentFormat from hyprctl to rebuild
the full monitor config string before re-enabling.

## Test
1. Connect external monitor
2. Run script → internal display off
3. Run script again → internal display on with correct scale and bit depth


dhh:
No longer needed since dev moved to a permanent config file being added/removed.

PR #5331, opened 2026-04-17
Updates tmux Alt+num to create a tab if it doesn't exists

leolimasnts:


dhh:
That's an interesting idea, but I find it a bit counter intuitive that you can create these windows with numbers out of order. Prefer to keep it simple as-is. Thanks for suggesting, though.

PR #5332, opened 2026-04-17
Enable SSD TRIM for LUKS-encrypted drives

### LINKED ISSUE #2229: Enable SSD TRIM

stephentaylor-com:
## Summary

- Add `:allow-discards` to the `cryptdevice=` kernel parameter so TRIM commands pass through the dm-crypt layer to the SSD
- Migration for existing installs applies immediately via `cryptsetup --persistent` (no reboot required)
- New installs get `:allow-discards` automatically during Limine setup

Closes #2229

## Details

Without `:allow-discards`, dm-crypt blocks all TRIM/DEALLOCATE commands from reaching the SSD — even on modern NVMe drives. The drive controller treats freed blocks as occupied data, increasing write amplification and reducing drive longevity.

Since kernel 6.2, btrfs automatically enables `discard=async`, so `fstrim.timer` is not needed. The only missing piece is allowing TRIM commands to pass through the LUKS encryption layer.

### Changes

**`migrations/1776500000.sh`** — Migration for existing installs:
- Detects LUKS encryption via `cryptdevice=` in `/etc/default/limine`
- Skips silently on unencrypted systems or if already configured
- Appends `:allow-discards` to the kernel parameter
- Applies immediately via `cryptsetup --allow-discards --persistent refresh root`
- Regenerates initramfs and boot entry

**`install/login/limine-snapper.sh`** — For new installs:
- Appends `:allow-discards` to the extracted `cryptdevice=` cmdline before writing to `/etc/default/limine`

## Test plan

- [ ] Verify on LUKS-encrypted system: `lsblk --discard` shows non-zero `DISC-GRAN` for root after migration
- [ ] Verify `sudo fstrim -av` trims the root partition
- [ ] Verify unencrypted systems are unaffected (migration exits cleanly)
- [ ] Verify fresh install on encrypted disk gets `:allow-discards` in `/etc/default/limine`
- [ ] Verify idempotent: running migration twice doesn't double-append

### DIFF
--- a/migrations/1776500000.sh
+++ b/migrations/1776500000.sh
@@ -1,0 +1,26 @@
+echo "Enable SSD TRIM for LUKS-encrypted drives"
+
+# Only apply if the system uses LUKS encryption
+if ! grep -q "cryptdevice=" /etc/default/limine 2>/dev/null; then
+  exit 0
+fi
+
+# Skip if allow-discards is already configured
+if grep -q "allow-discards" /etc/default/limine 2>/dev/null; then
+  exit 0
+fi
+
+# Add :allow-discards to the cryptdevice= parameter in the boot config
+# Format: cryptdevice=device:dmname -> cryptdevice=device:dmname:allow-discards
+sudo sed -i 's/\(cryptdevice=[^ ]*\)/\1:allow-discards/' /etc/default/limine
+
+# Enable allow-discards on the running LUKS device so TRIM works immediately
+if [[ -b /dev/mapper/root ]]; then
+  sudo cryptsetup --allow-discards --persistent refresh root
+fi
+
+# Regenerate initramfs and boot entry
+sudo limine-mkinitcpio
+sudo limine-update
+
+echo "SSD TRIM enabled for LUKS encryption"
--- a/migrations/1776500000.sh
+++ b/migrations/1776500000.sh
@@ -12,11 +12,13 @@ fi
 
 # Add :allow-discards to the cryptdevice= parameter in the boot config
 # Format: cryptdevice=device:dmname -> cryptdevice=device:dmname:allow-discards
-sudo sed -i 's/\(cryptdevice=[^ ]*\)/\1:allow-discards/' /etc/default/limine
+sudo sed -i 's/\(cryptdevice=[^ "]*\)/\1:allow-discards/' /etc/default/limine
 
-# Enable allow-discards on the running LUKS device so TRIM works immediately
-if [[ -b /dev/mapper/root ]]; then
-  sudo cryptsetup --allow-discards --persistent refresh root
+# Extract the dmname from cryptdevice=device:dmname and enable allow-discards
+# on the running LUKS device so TRIM works immediately
+DMNAME=$(grep -oP 'cryptdevice=[^: "]+:\K[^: "]+' /etc/default/limine | head -1)
+if [[ -n $DMNAME ]] && [[ -b /dev/mapper/$DMNAME ]]; then
+  sudo cryptsetup --allow-discards --persistent refresh "$DMNAME"
 fi
 
 # Regenerate initramfs and boot entry
--- a/install/login/limine-snapper.sh
+++ b/install/login/limine-snapper.sh
@@ -29,6 +29,11 @@ EOF
 
   CMDLINE=$(grep "^[[:space:]]*cmdline:" "$limine_config" | head -1 | sed 's/^[[:space:]]*cmdline:[[:space:]]*//')
 
+  # Enable SSD TRIM passthrough for LUKS-encrypted drives
+  if [[ $CMDLINE == *"cryptdevice="* ]] && [[ $CMDLINE != *"allow-discards"* ]]; then
+    CMDLINE=$(echo "$CMDLINE" | sed 's/\(cryptdevice=[^ ]*\)/\1:allow-discards/')
+  fi
+
   sudo cp $OMARCHY_PATH/default/limine/default.conf /etc/default/limine
   sudo sed -i "s|@@CMDLINE@@|$CMDLINE|g" /etc/default/limine
 
--- a/install/login/limine-snapper.sh
+++ b/install/login/limine-snapper.sh
@@ -31,7 +31,7 @@ EOF
 
   # Enable SSD TRIM passthrough for LUKS-encrypted drives
   if [[ $CMDLINE == *"cryptdevice="* ]] && [[ $CMDLINE != *"allow-discards"* ]]; then
-    CMDLINE=$(echo "$CMDLINE" | sed 's/\(cryptdevice=[^ ]*\)/\1:allow-discards/')
+    CMDLINE=$(printf '%s\n' "$CMDLINE" | sed 's/\(cryptdevice=[^ "]*\)/\1:allow-discards/')
   fi
 
   sudo cp $OMARCHY_PATH/default/limine/default.conf /etc/default/limine


migrations/1776500000.sh
```diff
@@ -0,0 +1,28 @@
...
+
+# Add :allow-discards to the cryptdevice= parameter in the boot config
+# Format: cryptdevice=device:dmname -> cryptdevice=device:dmname:allow-discards
+sudo sed -i 's/\(cryptdevice=[^ "]*\)/\1:allow-discards/' /etc/default/limine
+
+# Extract the dmname from cryptdevice=device:dmname and enable allow-discards
+# on the running LUKS device so TRIM works immediately
+DMNAME=$(grep -oP 'cryptdevice=[^: "]+:\K[^: "]+' /etc/default/limine | head -1)
+if [[ -n $DMNAME ]] && [[ -b /dev/mapper/$DMNAME ]]; then
+  sudo cryptsetup --allow-discards --persistent refresh "$DMNAME"
```
yevshev, line 21: We should consider disabling r/w workqueues here  too:
https://wiki.archlinux.org/title/Dm-crypt/Specialties#Disable_workqueue_for_increased_solid_state_drive_(SSD)_performance

> The `no_read_workqueue` and `no_write_workqueue` flags were introduced by internal Cloudflare research [Speeding up Linux disk encryption](https://blog.cloudflare.com/speeding-up-linux-disk-encryption/) made while investigating overall encryption performance. One of the conclusions is that internal dm-crypt read and write queues decrease performance for SSD drives. While queuing disk operations makes sense for spinning drives, bypassing the queue and writing data synchronously doubled the throughput and cut the SSD drives' IO await operations latency in half. The patches were upstreamed and are available since [linux](https://archlinux.org/packages/?name=linux) 5.9 and up [[5]](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/drivers/md/dm-crypt.c?id=39d42fa96ba1b7d2544db3f8ed5da8fb0d5cb877).

```suggestion
  sudo cryptsetup \
  --allow-discards \
  --perf-no_read_workqueue \
  --perf-no_write_workqueue \
  --persistent refresh "$DMNAME"
```

davidavasalcai:
**Real-world benchmark on Samsung 980 (DRAM-less) + Btrfs + LUKS:**

Just wanted to chime in with fresh data showing why this PR is essential on popular DRAM-less NVMe drives.

On a Samsung 980 1TB running Omarchy (LUKS + Btrfs) at ~89% disk usage, small synchronous writes (e.g. SQLite, pacman, browser caches) had degraded severely due to the drive controller being starved of TRIM commands:

* **Before TRIM passthrough:** 1000 sync writes (`oflag=dsync`) took **7.36s** (557 KB/s, ~7.36 ms per sync write).
* **After enabling `:discard` & running `fstrim -av`:** **131.3 GiB** of stale blocks were immediately reclaimed on `/dev/mapper/root`.
* **Retest:** The exact same 1000 sync writes took **3.25s** (1.3 MB/s, ~3.25 ms latency) — **over 2.3x faster** right after the first trim.

Notice that this PR currently has merge conflicts (`install/login/limine-snapper.sh` was moved during the Omarchy v4 refactor).

@stephentaylor-com are you able to rebase this over current `main`, or would maintainers prefer a refreshed PR?

yevshev:
Ongoing discussion and additional testing here: https://github.com/omacom/omarchy/discussions/6271

reverb256:
Heads-up: I carried this fix to the live branch as #13055, so the work does not stay stuck on `dev`.

Why a new PR rather than a rebase here: this branch targets `dev`, and it patches `install/login/limine-snapper.sh`, which no longer exists on `quattro` — `install/login/` now holds only `all.sh` and `sddm.sh`, because install-time boot configuration moved into the ISO. That is also why this PR reads as conflicting.

What #13055 does differently:

- the install-path half is a target-setup leaf (`install/config/luks-trim.sh`), which runs after the ISO writes `/etc/default/limine` from the installer cmdline and before the UKI is built from it, so it works on `quattro` without an ISO change
- the migration follows `migrations/1789325478.sh`'s shape: rebuild with `limine-mkinitcpio`, then verify the cmdline the entries are generated from and stay pending if it is missing, instead of assuming the option reached the boot entry
- `@yevshev`'s review request is implemented: `no-read-workqueue,no-write-workqueue` go on the same `cryptdevice=` parameter, in a separate commit that can be dropped on its own
- the hook spelling is justified with the `encrypt` hook's whitelist (Omarchy's `HOOKS` drop-in selects `encrypt`, not `sd-encrypt`), and the "unknown option is warned about, never fatal" property is spelled out, since that is the obvious first objection

Thanks for the original work — the diagnosis above is the same one, it just needed a live home.

PR #5340, opened 2026-04-17
Add Default Apps setup menu

dandresrp:
Summary
- add a new Setup > Default Apps flow for changing Omarchy and XDG default applications
- support terminal, shell, browser, editor, file manager, mail, image, PDF, video, and audio defaults
- reuse shared default handlers from install scripts, add terminal-editor desktop wrappers, and cover the menu flow with tests

Testing
- bash tests/defaults_test.sh
- bash tests/menu_defaults_test.sh
- bash -n bin/omarchy-defaults bin/omarchy-setup-defaults bin/omarchy-menu bin/omarchy-launch-editor bin/omarchy-install-terminal install/config/mimetypes.sh tests/defaults_test.sh tests/menu_defaults_test.sh

### DIFF
--- a/applications/hidden/hx.desktop
+++ b/applications/hidden/hx.desktop
@@ -1,0 +1,9 @@
+[Desktop Entry]
+Name=Helix
+Exec=xdg-terminal-exec --app-id=org.omarchy.hx -e hx %F
+TryExec=hx
+Type=Application
+Terminal=false
+NoDisplay=true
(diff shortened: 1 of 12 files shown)


dhh:
Appreciate the suggestion, but this is too big of a patch for something relatively minor like this. Feel free to investigate it as a separate AUR package 👌

dandresrp:
Understood, thanks.

PR #5341, opened 2026-04-17
Fix npx wrapper PATH leak from node@latest bootstrap

timohubois:
Follow-up to #5247. That earlier fix intentionally moved Omarchy's generated npx wrappers to `mise exec node@latest -- npx ...` so npm-distributed CLIs would keep working inside projects pinned to old Node versions.

This PR preserves that intent: wrappers remain system-wide lazy-loaded npm launchers, and the CLI itself is still bootstrapped with Omarchy's selected `node@latest` runtime.

The bug fixed here is the side effect reported in #5451: `mise exec node@latest` prepends `node@latest/bin` to `PATH`, and long-running agent CLIs like Codex can pass that modified `PATH` on to workspace commands. That makes commands run inside the project resolve `node@latest` instead of the project's own Node.

## Changes

- Resolve Omarchy's `node@latest` install with `mise where node@latest`, installing it globally on first use if needed.
- Run `npx` through the matching absolute `node` binary so the bootstrap runtime stays consistent.
- Use `npx --package ... -- which ...` only to resolve the package bin path.
- Execute the resolved CLI entrypoint with the selected `node@latest` binary without exporting or prepending `node@latest/bin` into `PATH`.
- Fail clearly with exit `127` if no npm bin can be resolved.
- Regenerate existing npx wrappers via migration.

## Testing

- `codex --version`
- `opencode --version`
- `playwright-cli --version`
- Verified from a project pinned to Node 10 that `codex sandbox linux node -v` resolves the project Node instead of `node@latest`.
- Verified the wrapper still bootstraps Codex, Opencode, and Playwright with the selected mise `node@latest` runtime.


dhh:
Could you share a concrete reproduction where mise exec node@latest -- npx picks up a foreign npx? My understanding is that mise exec prepends the runtime's bin directory to PATH for the duration of the child process, so npx should resolve to the same install as node.

timohubois:
@dhh Thanks for the feedback! I dug into this more and was able to reproduce it in a mixed shell setup (direnv + Nix + mise).
In that environment, running:
`mise exec node@latest -- npx ...`

can fail to keep npx coupled to the selected node runtime rather than the node@latest selected by mise, leading to errors such as:
`Error: Cannot find module 'node:path'`

A key part of the issue is that npx is a #!/usr/bin/env node script. Even when invoked via an absolute path, it can still pick up a different node from the ambient PATH, so $node_root/bin/npx alone isn’t sufficient to guarantee consistency.

I tried a few alternatives (mise exec, --fresh-env, shims, direct paths), but they weren’t reliable in this setup.

The fix I’m proposing now is to ensure the wrapper resolves both node and npx from the same runtime:
```
export PATH="$node_root/bin:$PATH"
exec "$node_root/bin/npx" --yes ...
```

This keeps the fix wrapper-local and ensures that npx resolves the matching node, even in mixed environments.

From my perspective, this fix makes sense if the wrapper is intended to keep npx launchers independent of the shell and tied to the selected mise Node runtime. I really appreciate your perspective on this. If I’ve misunderstood the intended scope, I’m happy to close the PR.

timohubois:
Did some rework, related to the reported issue https://github.com/basecamp/omarchy/issues/5451.

Issue #5338, opened 2026-04-17
Open in Ghostty gone in Nautilus


naxels:
### System details

Omarchy 3.5.1

```
pacman -Q | grep nautilus
libnautilus-extension 50.1-1
nautilus 50.1-1
nautilus-python 4.1.0-3
```

```
pacman -Q | grep ghostty
ghostty 1.3.1-2
ghostty-shell-integration 1.3.1-2
ghostty-terminfo 1.3.1-2
```

### What's wrong?

After upgrading last night to 3.5.1, for some reason now the "Open in Ghostty" is gone in Nautilus upon right click
Send via LocalSend is still present though.

After investigating, I have found the issue:

Looking at a snapshot taken yesterday (nr 42, before upgrading to 3.5.1)
I can see that `ghostty.py` is listed in:
/.snapshots/42/snapshot/usr/share/nautilus-python/extensions

After the 3.5.1 upgrade (which made snapshot nr 43 / current)
the entire nautilus-python folder is gone.. from /usr/share

naxels:
Found out why it happened:

upgpkg: 1.3.1-2: Split out nautilus extension
https://gitlab.archlinux.org/archlinux/packaging/packages/ghostty/-/commit/7c601d97c8e447371dbc004368b3b83666b61225

This extension now resides in:
https://archlinux.org/packages/extra/x86_64/ghostty-nautilus/

which is not installed on my system via `pacman -Qq | grep ghostty`

the fix:
run this:
`sudo pacman -S ghostty-nautilus`

and restart any Nautilus windows, now it works again :) 🎉

@dhh, can you create a migration and release it to fix this?

naxels:
I just confirmed on my other machine that runs Omarchy, it has the same problem and again, installing the package fixes the issue :)

dommcdev:
Ahhh, so this is why that menu option suddenly disappeared.

PR #5343, opened 2026-04-17
Install nautilus-open-any-terminal to open the default terminal from any Nautilus directory

pomartel:
This fixes Install #5338 

I don't know if it's still relevant now that Ghostty is not the default terminal anymore but it resolves the regression. Thanks to @naxels for looking into it.


timlau:
maybe a better approch is to use 

https://github.com/Stunkymonkey/nautilus-open-any-terminal

It supports a lot of different terminals

The package is available in AUR and can be installed by

`omarchy-pkg-aur-add nautilus-open-any-terminal`

There terminal is set by

`gsettings set com.github.stunkymonkey.nautilus-open-any-terminal terminal "ghostty"`

(Or "alacritty" or" kitty")

pomartel:
I haven't tried it but it could be a more universal approach that works with all terminals. I'm not sure how we deal with AUR packages in Omarchy.

dhh:
Would be great to get this working in all terminals! We can add AUR packages to OPR here: https://github.com/omacom-io/omarchy-pkgs

pomartel:
I rewrote the PR to use the `nautilus-open-any-terminal` AUR package instead, which should work with all the different Omarchy's terminals. I only tested it with Ghostty. This PR depends on https://github.com/omacom-io/omarchy-pkgs/pull/80/ being merged first.

dhh:
This doesn't seem to be working with Alacritty. I get the "Open Terminal Here" option, but it doesn't do anything.

pomartel:
@dhh Sorry, I should have tested it more throughly.

This time, I tested it with the 3 terminals we support out of the box (Alacritty, Ghostty and Kitty) and it all works fine.

This was a bit more involving than I initially thought because Ghostty does not work with `xdg-terminal-exec` and the `--dir` argument to open in a given directory. But I was able to come up with a minimal workaround that works in all cases.

pomartel:
I was able to simplify the Ghostty integration quite a bit and test with Foot but we still have to wait for a release of `nautilus-open-any-terminal` that includes this fix for Ghostty :  https://github.com/Stunkymonkey/nautilus-open-any-terminal/pull/288

PR #5353, opened 2026-04-18
feat: Add Touchpad toggle/on/off support

drunkleen:
Adds a omarchy-toggle-touchpad script and wires it up to hardware keys and the toggle menu.

changes:
- New bin/omarchy-toggle-touchpad script that enables/disables the touchpad via hyprctl, with OSD feedback via swayosd-client
- State is persisted as a Hyprland device config at `~/.local/state/omarchy/toggles/hypr/touchpad-disabled.conf`, so it survives reboots
- Hardware keys (XF86TouchpadToggle, XF86TouchpadOn, XF86TouchpadOff) are bound in media.conf, with On/Off enforce state rather than toggling, matching the key semantics
- Touchpad option is added to the toggle menu, but only shown when a touchpad/trackpad device is actually       
  detected

### DIFF
--- a/bin/omarchy-menu
+++ b/bin/omarchy-menu
@@ -168,7 +168,13 @@ show_share_menu() {
 }
 
 show_toggle_menu() {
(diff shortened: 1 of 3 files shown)


drunkleen:
Fixed in the latest commit. The toggle now after latest commit uses explicit branching instead of `&& ||`, making it reliable.

Issue #5149, opened 2026-03-29
New Wallpaper Picker does not recognize custom wallpaper entries


whughesiii2187:
### System details

ThinkPad T-14 Gen3 Intel

### What's wrong?

When applying a theme and linking my wallkpapers to the .config/omarchy/current/backgrounds, the new wallpaper picker does not recognize them to be selected. It only sees the wallpapers associated with the theme.

dhh:
We have changed how backgrounds are added: https://learn.omacom.io/2/the-omarchy-manual/89/backgrounds

PR #5358, opened 2026-04-19
Add global ~/Wallpapers directory to background selector

nzarg:
Addresses discussion #3164 and issue #5149. ~/Wallpapers is always included in the background selector regardless of the active theme, searched last so theme wallpapers take priority. find now uses -L to follow symlinks, so ~/Wallpapers can point to a mounted network share or similar. Migration 1776587855.sh creates the directory on first run.

If ~/Wallpapers is empty or doesn't exist, behaviour is unchanged.


dhh:
Closing because this implementation lives entirely in the removed Elephant selector. Support for ~/Wallpapers can be re-proposed against the native background picker.

PR #5367, opened 2026-04-20
Harden Resume Boost

spencerbull:
## Summary
- replace the resume hook's save-and-restore flow with a temporary `powerprofilesctl` performance hold after resume
- apply the resume boost for any non-`performance` profile so wake-up stays responsive without doing extra work when the system is already in `performance`
- add a migration that reinstalls the updated Intel resume hook on upgrade so existing systems pick up the new behavior
## Why
The previous resume hook saved the active power profile before suspend, switched to `performance` on wake, and restored the saved profile 10 seconds later. That approach was fragile across repeated suspend/resume cycles and could restore the wrong profile or leave the system in `performance` if the delayed restore path failed.
This change lets `power-profiles-daemon` manage the temporary resume override directly through a short-lived performance hold instead of Omarchy snapshotting and replaying profile state itself.
## Testing
- ran `bash -n default/systemd/system-sleep/resume-boost`
- ran the migration locally to reinstall the hook
- tested the installed hook directly from `balanced`, `power-saver`, and `performance`
- verified the D-Bus hold state through `net.hadess.PowerProfiles`
- tested overlapping resume-style invocations to confirm the system no longer gets stuck in `performance`
- tested a real suspend/resume cycle and verified the resume boost transient unit started and completed successfully

### DIFF
--- a/default/systemd/system-sleep/resume-boost
+++ b/default/systemd/system-sleep/resume-boost
@@ -1,16 +1,52 @@
 #!/bin/bash
 
-# Temporarily switch to performance power profile on resume
-# to avoid sluggishness while the system catches up, then
-# restore the previous profile after 10 seconds.
+# Temporarily hold the performance power profile on resume
+# when waking from balanced mode to avoid sluggishness while
+# the system catches up without clobbering other user choices.
 
-if [[ $1 == "pre" ]]; then
-  mkdir -p /run/omarchy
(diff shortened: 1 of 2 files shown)


dhh:
Turns out that this resume-boost idea doesn't actually work in practice. When Panther Lake has been sleeping for hours, it still enters a state where it takes 20-30s to return to full speed.

PR #5368, opened 2026-04-20
add: xdg-dekstop-portal file picker floating windowrule

justindotdevv:


dhh:
PRs without description or comment won't be considered.

PR #5394, opened 2026-04-22
Add Workspace Profiles (opt-in)

alansikora:
## Overview

Named workspace profiles built on top of Hyprland's named workspaces. Each profile (e.g. `work`, `personal`) owns its own five workspace slots. `SUPER+1..5` focuses slot N of the active profile; `SUPER+CTRL+1..9` switches to the Nth profile. Windows on each profile's workspaces persist across profile switches, so contexts are preserved.

Per-profile Chromium isolation is included: each profile launches Chromium with its own `--user-data-dir`, keeping accounts, cookies, extensions, and history separate. An `xdg-mime` shim registered as the default http(s) handler routes all link clicks through the active profile's browser, which means links from Signal, Slack, email clients, or any other app open in the profile-appropriate Chromium automatically.

The feature is opt-in via `Settings → Setup → Workspace Profiles`, mirroring the existing Key Remapping flow. Users who do not enable it see no behavioral change from stock Omarchy. The default configs gain one `source` / `include` / `@import` line each, pointing at sidecars that remain empty until setup populates them. Because `omarchy-refresh-waybar` and `omarchy-refresh-hyprland` only overwrite the stock files, the feature's configuration is refresh-safe by construction.

## Keybindings (after opt-in)

| Keys | Action |
|---|---|
| `SUPER+1..5` | Focus slot N of active profile |
| `SUPER+CTRL+1..9` | Switch to Nth profile |
| `SUPER+CTRL+TAB` | Cycle profile |
| `SUPER+SHIFT+1..5` | Move focused window to slot N (follow) |
| `SUPER+SHIFT+ALT+1..5` | Move focused window to slot N (silent) |
| `SUPER+CTRL+SHIFT+1..9` | Send focused window to Nth profile's same slot |
| `SUPER+TAB` / `SUPER+SHIFT+TAB` | Next / previous slot within active profile |

## Components

- `bin/omarchy-workspace-profile` — Controller. Exposes subcommands for slot navigation (`goto`, `next-slot`, `prev-slot`), profile switching (`set`, `set-index`, `cycle`), profile CRUD (`create`, `rename`, `delete`, `default`, `color`, `move-up`, `move-down`), window moves scoped to the active profile (`move`, `movesilent`, `send-to`, `send-to-index`), boot-time window migration, a kill switch with configurable window policies (`on`, `off`), and a generic per-profile app runner (`app`).
- `bin/omarchy-workspace-profile-sync` — Python daemon listening on Hyprland's socket2. Detects focus transitions not originated by the controller (mako notification clicks, Slack "jump to tab", other apps' activate requests) and keeps the recorded active profile aligned with the workspace actually in focus.
- `bin/workspace-profile-browser` — Shim invoked by `omarchy-launch-browser`. Because `omarchy-launch-browser` extracts only the first token of the `.desktop` `Exec=` line, the shim occupies that slot and re-dispatches to `omarchy-workspace-profile app browser`.
- `bin/omarchy-setup-workspace-profiles` — Idempotent opt-in installer. Populates the Hyprland bindings + autostart sidecars, installs the browser `.desktop`, registers the xdg-mime handler, converts the mako config to the real-file form that includes both the theme and our sidecar, and invokes the controller's `on`.
- `bin/omarchy-menu` — One entry added to the Setup submenu.
- `bin/omarchy-launch-webapp` — Whitelists the browser shim so PWAs (`omarchy-launch-webapp "<URL>"`) also route through the active profile's Chromium.
- `config/omarchy/extensions/workspace-profiles.sh` — Menu extension providing switch / create / rename / delete / set-default / set-color / reorder / disable UI. Delivered as an extension so it survives `omarchy-update`.
- `config/{hypr,waybar}/workspace-profiles.*` — Empty sidecar stubs included from the stock default configs.

## Migration path

Fresh installs ship the include lines but empty sidecars; Hyprland, Waybar, and mako silently skip empty includes. Stock behavior is unchanged.

Existing installs receive the binaries and the updated default configs via `omarchy-update`. The feature remains dormant until the user opens the Setup menu and selects Workspace Profiles.

Opting in runs `omarchy-setup-workspace-profiles`, which installs the menu extension, populates the sidecars (including the Hyprland bindings), registers the browser `.desktop` handler via `xdg-mime`, converts `~/.config/mako/config` from the theme symlink to a real file that includes both the theme and the workspace-profiles sidecar, and calls `omarchy-workspace-profile on`. The controller's `first_time_setup` routine scans the user's current Hyprland workspaces, detects an existing Chromium data directory (`~/.config/chromium` or `~/.config/google-chrome`), seeds a single `default` profile pinned to that directory, and migrates any windows on bare-numeric workspaces into the new profile. The user's existing browser history, extensions, and signed-in sessions are preserved.

The kill switch (`omarchy-workspace-profile off`, also exposed through the menu) offers `--keep-all` (move every window to its matching numeric workspace) and `--keep-default` (preserve only the default profile's windows, close the rest) policies, along with a hidden `--no-windows` flag for scripted use.

## Test plan

- [ ] Fresh Omarchy VM: `SUPER+1..5` dispatches to numeric workspaces 1..5 with no behavior change
- [ ] `omarchy-refresh-waybar` does not wipe the sidecar
- [ ] `omarchy-refresh-hyprland` does not wipe the sidecar
- [ ] Existing install on `omarchy-update`: feature remains dormant until menu click
- [ ] `Settings → Setup → Workspace Profiles` seeds a `default` profile from live state; Chromium history and sessions preserved
- [ ] `SUPER+1..5` lands on slot N of the active profile
- [ ] `SUPER+CTRL+1..9` switches to the Nth profile
- [ ] Creating a second profile launches its Chromium into an isolated user-data-dir
- [ ] Clicking an `https://` link in Signal or Slack opens in the active profile's Chromium
- [ ] Omarchy webapp launchers (e.g. `SUPER+SHIFT+Y`) use the active profile's Chromium
- [ ] `omarchy-workspace-profile off --keep-default` cleanly restores stock behavior

### DIFF
--- a/bin/omarchy-launch-webapp
+++ b/bin/omarchy-launch-webapp
@@ -5,7 +5,7 @@
 browser=$(xdg-settings get default-web-browser)
 
 case $browser in
-google-chrome* | brave* | microsoft-edge* | opera* | vivaldi* | helium*) ;;
+google-chrome* | brave* | microsoft-edge* | opera* | vivaldi* | helium* | workspace-profile-browser*) ;;
 *) browser="chromium.desktop" ;;
 esac
 
--- a/bin/omarchy-menu
+++ b/bin/omarchy-menu
@@ -234,6 +234,7 @@ show_font_menu() {
 show_setup_menu() {
   local options="  Audio\n  Wifi\n󰂯  Bluetooth\n󱐋  Power Profile\n  System Sleep\n󰍹  Monitors"
   [[ -f ~/.config/hypr/bindings.conf ]] && options="$options\n  Keybindings"
+  options="$options\n󱂬  Workspace Profiles"
   [[ -f ~/.config/hypr/input.conf ]] && options="$options\n  Input"
   options="$options\n󰱔  DNS\n  Security\n  Config"
 
@@ -246,6 +247,7 @@ show_setup_menu() {
   *Monitors*) open_in_editor ~/.config/hypr/monitors.conf ;;
   *Keybindings*) open_in_editor ~/.config/hypr/bindings.conf ;;
   *Input*) open_in_editor ~/.config/hypr/input.conf ;;
+  *Workspace\ Profiles*) omarchy-setup-workspace-profiles ;;
   *DNS*) present_terminal omarchy-setup-dns ;;
   *Security*) show_setup_security_menu ;;
   *Config*) show_setup_config_menu ;;
--- a/bin/omarchy-setup-workspace-profiles
+++ b/bin/omarchy-setup-workspace-profiles
@@ -26,13 +26,10 @@ mkdir -p "$HOME/.config/omarchy/extensions"
 cp -f "$OMARCHY_PATH/config/omarchy/extensions/workspace-profiles.sh" \
       "$HOME/.config/omarchy/extensions/workspace-profiles.sh"
 
-# ────────── ensure sidecar include files exist ──────────
-# The stock Omarchy configs source/include these sidecars. First-time
-# opt-in, they may not have been copied into ~/.config yet — do it now
-# so our `on` below has a target to populate.
-for rel in hypr/workspace-profiles.conf \
-           hypr/workspace-profiles-autostart.conf \
-           waybar/workspace-profiles.jsonc \
+# ────────── ensure Waybar sidecars exist ──────────
+# Waybar sidecars are empty stubs at rest; the controller fills them on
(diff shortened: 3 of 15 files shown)


dhh:
Interesting path, but too big for Omarchy itself. Do explore an optional AUR package for this!

alansikora:
I thought so, will do

PR #5395, opened 2026-04-23
Add Battery Charging Policy Provider with Contract

spencerbull:
PR Body
## Summary
- formalize the battery charging backend contract around five subcommands: `probe`, `list`, `status`, `set`, and `set-custom`
- make provider discovery easier to extend by moving the provider list into a dedicated `BATTERY_POLICY_PROVIDERS` array in `bin/omarchy-battery-policy-provider`
- add inline guidance in the existing provider implementations so new `omarchy-battery-policy-provider-<name>` backends can follow the same structure
## Details
This update makes the battery policy provider model more explicit and easier to extend.
`bin/omarchy-battery-policy-provider` now documents how a new provider should be registered:
- add a new `bin/omarchy-battery-policy-provider-<name>` script
- implement the provider contract (`probe`, `list`, `status`, `set`, `set-custom`)
- append the provider name to `BATTERY_POLICY_PROVIDERS` in priority order
Both existing providers now document that contract at the top of the file and include brief comments around the main contract functions so it is clear what each entrypoint is responsible for:
- `bin/omarchy-battery-policy-provider-dell-sysman`
- `bin/omarchy-battery-policy-provider-power-supply`
## Menu Integration
The battery charging menu entry is exposed from `bin/omarchy-menu` under:
`Setup -> Power -> Battery Charging`
That menu item launches `bin/omarchy-setup-battery-policy`, which is where the selectable charging policies and custom threshold flow are implemented.

Ref #5144 

cc: @dhh @spaceXrace


dhh:
We're going to need to extract this into an AUR package. Can't add 1200 lines of code to Omarchy for this relatively small feature.

bin/omarchy-battery-policy-provider-power-supply
```diff
@@ -0,0 +1,396 @@
...
+      printf '%s\n' "$start" | sudo tee "$bat/charge_control_start_threshold" >/dev/null || exit 1
+    fi
+
+    printf '%s\n' "$end" | sudo tee "$bat/charge_control_end_threshold" >/dev/null || exit 1
+
+    if [[ -r $bat/charge_types ]] && supports_raw_policy "$bat" Custom; then
+      printf '%s\n' 'Custom' | sudo tee "$bat/charge_types" >/dev/null || exit 1
+    fi
+  done
+}
```
spaceXrace, line 386: The start and end threshold need to be applied in the correct order. 

In the current implementation, when raising the limit, the start threshold is temporarily higher than the end threshold, which leads to the new limit getting rejected on my Laptop.

It worked with this change:

```suggestion
set_custom_policy() {
  local bat
  local batteries
  local start="$1"
  local end="$2"
  local current_end

  mapfile -t batteries < <(battery_dirs)
  (( ${#batteries[@]} > 0 )) || exit 1

  is_integer "$end" || {
    echo "Stop threshold must be an integer." >&2
    exit 1
  }

  if (( end < 1 || end > 100 )); then
    echo "Stop threshold must be between 1 and 100." >&2
    exit 1
  fi

  for bat in "${batteries[@]}"; do
    [[ -r $bat/charge_control_end_threshold ]] || {
      echo "This machine does not support custom battery thresholds." >&2
      exit 1
    }

    if [[ -r $bat/charge_control_start_threshold ]]; then
      is_integer "$start" || {
        echo "Start threshold must be an integer." >&2
        exit 1
      }

      if (( start < 0 || start > 99 )); then
        echo "Start threshold must be between 0 and 99." >&2
        exit 1
      fi

      if (( end <= start )); then
        echo "Stop threshold must be greater than the start threshold." >&2
        exit 1
      fi
    fi
  done

  for bat in "${batteries[@]}"; do
    if [[ -r $bat/charge_control_start_threshold ]]; then
      current_end=$(<"$bat/charge_control_end_threshold")

      if (( start > current_end )); then
        printf '%s\n' "$end" | sudo tee "$bat/charge_control_end_threshold" >/dev/null || exit 1
        printf '%s\n' "$start" | sudo tee "$bat/charge_control_start_threshold" >/dev/null || exit 1
      else
        printf '%s\n' "$start" | sudo tee "$bat/charge_control_start_threshold" >/dev/null || exit 1
        printf '%s\n' "$end" | sudo tee "$bat/charge_control_end_threshold" >/dev/null || exit 1
      fi
    else
      printf '%s\n' "$end" | sudo tee "$bat/charge_control_end_threshold" >/dev/null || exit 1
    fi

    if [[ -r $bat/charge_types ]] && supports_raw_policy "$bat" Custom; then
      printf '%s\n' 'Custom' | sudo tee "$bat/charge_types" >/dev/null || exit 1
    fi
  done
}

```

bin/omarchy-setup-battery-policy
```diff
@@ -0,0 +1,366 @@
...
+    if activate_custom_policy; then
+      change_succeeded=true
+    fi
+  else
+    if omarchy-battery-policy-set "$selected_policy_id"; then
+      change_succeeded=true
+    fi
+  fi
+
+  echo
```
spaceXrace, line 351: I noticed that the status is shown twice if the screen is not cleared at this point.


```suggestion
  clear
```

PR #5397, opened 2026-04-23
Create snapper configs during chroot install

### LINKED ISSUE #5344: Snapper /home config isn't created on chroot install leading to black screen
### System details

Clevo, integrated graphics, Omarchy 3.5.1

### What's wrong?

There seems to be a bug in limine-snapper.sh: at line 54 chroot installation doesn't have a /home config created. I'm guessing the following sed commands fail silently and as a result without a config there's no snapshot policy. The snapshot subvolumes keep growing - My disk completely filled while updating and it totally nuked my OS. Backup snapshots wouldn't work either, hyprland seemed to not be able to start. I was able to get into tty and clear out 33 (!!) excess snapshot sub-volumes which then let me get into a recent snapshot to restore. I'm happy to submit a patch over the weekend if this is the case.

peteonrails:
The snapper config creation for root and home was guarded by a check
that skipped it during chroot installs. The sed commands that set
retention policy immediately after ran unconditionally and failed
silently when the configs didn't exist, leaving chroot installs with
no snapshot cleanup policy and unbounded snapshot growth.

Fixes #5344

### DIFF
--- a/install/login/limine-snapper.sh
+++ b/install/login/limine-snapper.sh
@@ -50,15 +50,13 @@ EOF
   # We overwrite the whole thing knowing the limine-update will add the entries for us
   sudo cp $OMARCHY_PATH/default/limine/limine.conf /boot/limine.conf
 
-  # Match Snapper configs if not installing from the ISO
-  if [[ -z ${OMARCHY_CHROOT_INSTALL:-} ]]; then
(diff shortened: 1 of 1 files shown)


dhh:
Superseeded by a more comprehensive revamp here: https://github.com/basecamp/omarchy/commit/24ad7c7abada7d78b5baf1cb294415c37a2ab59d

PR #5399, opened 2026-04-23
Improve screenshot screenrecord documentation

marcmav:
improved documentation for easier accessing when changing defaults, added absolute paths so it can be easy accessible


dhh:
Appreciate the suggestion, but don't think there's enough here to warrant the change.

PR #5402, opened 2026-04-23
Add subtle selected-row highlight to Walker

pipetogrep:
## Summary
Adds a subtle, theme-aware background to the selected Walker result in the default Omarchy theme.

This makes the active item easier to pick out while keeping the highlight tied to the existing text color token and transparency.

image

image

image

image

### DIFF
--- a/default/walker/themes/omarchy-default/style.css
+++ b/default/walker/themes/omarchy-default/style.css
@@ -116,5 +116,5 @@ child:selected .item-box * {
 }
 
 child:selected {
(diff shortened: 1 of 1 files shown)


pipetogrep:
> This is a great accessibility improvement. I've had problems before seeing which menu item is selected.
> 
> The contrast of the selected text is a bit low, is there a way to make it more distinct from the background without changing the aesthetic?

I think it's intentionally subtle, I didn't want to touch those parts of the themes.

PR #5403, opened 2026-04-23
Persist workspace layout toggles

pipetogrep:
## Summary
Persists the Hyprland workspace layout toggle through Omarchy's existing toggle config mechanism.

The previous implementation changed the active workspace layout with `hyprctl keyword`, which is runtime-only and gets reset by config reloads. The toggle now writes an explicit workspace layout rule to `~/.local/state/omarchy/toggles/hypr/workspace-layouts.conf` and reloads Hyprland, matching the persistent toggle pattern already used for window gaps and single-window aspect ratio.

This relies on the existing Omarchy toggle setup to create the toggle directory. The workspace layout file itself is created by the append path the first time a workspace layout is persisted, so the command does not need to run `mkdir -p` or `touch` on every layout change.

## Testing
- `bash -n bin/omarchy-hyprland-workspace-layout-toggle`
- `git diff --check origin/dev...HEAD`
- Verified toggling with no existing `workspace-layouts.conf` creates `workspace = <id>, layout:scrolling`
- Verified the persisted `scrolling` layout survives `hyprctl reload`
- Verified the persisted `scrolling` layout survives `omarchy-theme-set <current-theme>`
- Verified toggling the same workspace back updates the entry to `workspace = <id>, layout:dwindle`

### DIFF
--- a/bin/omarchy-hyprland-workspace-layout-toggle
+++ b/bin/omarchy-hyprland-workspace-layout-toggle
@@ -2,6 +2,7 @@
 
 # Toggle the layout on the current active workspace between dwindle and scrolling
 
+LAYOUTS_CONF="$HOME/.local/state/omarchy/toggles/hypr/workspace-layouts.conf"
 ACTIVE_WORKSPACE=$(hyprctl activeworkspace -j | jq -r '.id')
 CURRENT_LAYOUT=$(hyprctl activeworkspace -j | jq -r '.tiledLayout')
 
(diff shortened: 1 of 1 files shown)


dhh:
This is clever and a good workaround, but it also feels like something that should have a native Hyprland answer. How to make these ephemeral settings survive config reloads? Maybe @vaxerski has an idea?

vaxerski:
they don't survive config reloads by design, that's just how the config works.

dhh:
@pipetogrep Another way to do it then is via the toggles we've been using for 1-window ratio etc. Where we write the config file needed to file for it. That's probably the way to go. Then it'll also survive a reboot.

pipetogrep:
@dhh Thanks for the feedback and the suggestion to use toggles. It's nice having the workspace layout preference be remembered after reboot!

PR #5404, opened 2026-04-23
alacritty.toml fallback for legacy themes without colors.toml

prepin:
Omarchy Manual's [Extra Themes](https://learn.omacom.io/2/the-omarchy-manual/90/extra-themes) page currently showcases 109 themes of which roughly half (53 themes) do not have a `colors.toml` file and therefore they do not support the newer Omarchy templating engine that generates application themes in `~/.config/omarchy/themed`.

Idea is to add fallback for such legacy themes that reconstructs `colors.toml` in case it is missing using colors defined in `alacritty.toml` (which is always present).

I've run an exhaustive test on each of 53 legacy themes and found no issues with all but one theme (Midnight theme, missing part of the palette).

P.S. Two themes listed on the Extra Themes page have already been deleted from GitHub -- Green Hakkar and Torrentz Hydra are gone.

---
Inspiration: [Omazed](https://github.com/APS6/omazed) project (Zed theme generator for Omarchy).


dhh:
This is very cool. Although I wonder if it might make sense to use this script to set lose on those existing themes and create PRs to add the colors.toml files instead?

dhh:
Actually, this is a better idea. Just do it automatically. Respect the legacy.

PR #5410, opened 2026-04-23
Enhance web app launcher to support Flatpak browsers

SorenHJohansen:
Fix Flatpak support and browser handling while preserving original launch behavior.

**Changes**
Flatpak detection using app ID
Only use --app for Chromium-based browsers

**Notes**
No change to overall execution model or session handling

### DIFF
--- a/bin/omarchy-launch-webapp
+++ b/bin/omarchy-launch-webapp
@@ -1,12 +1,35 @@
 #!/bin/bash
 
-# Launch the passed in URL as a web app in the default browser (or chromium if the default doesn't support --app).
+url="$1"
+shift
 
 browser=$(xdg-settings get default-web-browser)
 
-case $browser in
-google-chrome* | brave* | microsoft-edge* | opera* | vivaldi* | helium*) ;;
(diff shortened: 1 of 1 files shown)


SorenHJohansen:
NOTE: Assumes Flatpak app ID matches .desktop name; falls back if not.

dhh:
Appreciate the suggestion, but Omarchy does not have any flatpak affordances by default. I don't think it's actually a good idea to install your browsers through flatpak when you can do it via the Arch packages or AUR. So going to skip for now.

SorenHJohansen:
Totally fair regarding Omarchy’s Arch-first approach.

My main reasoning was the added sandboxing benefits for browsers given their larger attack surface, rather than suggesting Flatpak as a replacement for native packaging defaults.

I’ve personally switched to the Flatpak version of Vivaldi and have had zero issues so far, besides having to do a small tweak for launching webapps with the existing script here 🙂. 

That experience is what made me think it could be a worthwhile optional path for users who prioritize additional isolation.

PR #5429, opened 2026-04-24
feat(hyprlock): implement a shutdown button for hyprlock

Rockeyxx:
added a shutdown widget to hyprlock config to avoid writing the password to shutdown the devcie


dhh:
This is really cleanly done. Nice aesthetics. But I can't remember a single time I've actually wanted to do this personally. So think maybe we document it under Discussions for those who want it. Thanks for the suggestion!

Issue #5423, opened 2026-04-24
ASUS ExpertBook B9406CAA (Panther Lake): display refresh stall. Fixed by xe.enable_panel_replay=0, touchpad also fixed


mijuny:
### System details

ASUS ExpertBook B9406CAA (BIOS 304), Intel Panther Lake / Arc B390 [8086:b080], Omarchy 3.6, kernel linux-ptl 6.19.10, Hyprland 0.54.3

### What's wrong?

Hi!

Just managed to install Omarchy 3.6 on this just-released laptop (ASUS ExpertBook B9406CAA, Panther Lake). Noticed bugs with display / GPU and found a fix with Claude Code. During the install, the only way to "refresh" the screen was to switch between tty3 and tty1. So, installation was possible but a bit messy. The fix below was produced with Claude Code (Opus 4.7), disclosed up front.

---

### Symptom

On a default Omarchy 3.6 install on this laptop, the display never refreshes after the first scanout. Every display layer appears frozen until a VT switch forces a full modeset:

- **Plymouth / LUKS prompt**: logo briefly shows, input box never renders, password typed blind is accepted.
- **Kernel TTY (fbcon on xedrmfb)**: type a command, see nothing; Ctrl+Alt+F3 then Ctrl+Alt+F1 makes the previous state snap on-screen.
- **Hyprland**: session starts normally and runs for minutes spawning child processes, but the screen stays on the last-committed frame.

Install was only possible by flipping TTYs repeatedly to force redraws. Until the fix below, `nomodeset` at the Limine editor was the only workaround — usable but falls back to `simpledrm` and kills GPU acceleration, video decode, and battery features.

### Root cause

**Panel Replay** — a Xe3-new self-refresh mechanism, default-on in the `xe` driver — has a broken exit/wake path on this eDP panel. The panel latches the last-presented frame in self-refresh and never wakes for subsequent atomic commits. A full modeset (VT switch) resets the link and one new frame appears, then the same stall again. The older `xe.enable_psr=0` knob does NOT cover Panel Replay; they are distinct parameters (see `modinfo xe`).

### Fix

One kernel argument:

```
xe.enable_panel_replay=0
```

### Bisection

| Cmdline delta | Result |
| --- | --- |
| default | stall at every layer |
| `nomodeset` | works via simpledrm (workaround only) |
| `xe.enable_dc=0` | no effect |
[...]

PR #5435, opened 2026-04-24
Add ASUS ExpertBook B9406 display and touchpad fixes for Panther Lake

mijuny:
Requested in #5423 after DHH confirmed the scope.

Three confirmed fixes on ASUS ExpertBook B9406CAA (Panther Lake / Xe3), matched narrowly via new `omarchy-hw-asus-expertbook-b9406` (product name contains "B9406" AND `omarchy-hw-intel-ptl`).

AI disclaimer: Fixes found with help from Claude Code (Opus 4.7).

- **Display refresh stall:** `xe.enable_panel_replay=0` on kernel cmdline. Without this, both installing and running Omarchy on this laptop is very painful for most users: the live installer, plymouth/LUKS prompt, TTY, and Hyprland all freeze after the first frame drawn to the panel, and the only way to force a redraw is to flip VTs (Ctrl+Alt+F3 then F1) repeatedly. Cause: Panel Replay (Xe3-new, default-on in the xe driver) has a broken exit/wake path on this eDP panel — it latches the first-presented frame in self-refresh and never wakes for subsequent atomic commits. The older `xe.enable_psr=0` knob does not cover Panel Replay.

- **Brightness control:** `xe.enable_dpcd_backlight=1` on kernel cmdline. Without this, display brightness can't be adjusted — `intel_backlight` sysfs writes succeed but produce no visible change on the panel, and the brightness keys effectively toggle the screen between a fixed level and off. Cause: the panel's EDID on eDP-1 reads as empty, so xe takes backlight type from VBT (which says PWM), but the panel actually wants DPCD AUX backlight control.

- **Touchpad motion:** libinput quirks file masking `ABS_PRESSURE` and `ABS_MT_PRESSURE` for Pixart `093A:4F05`. Without this, the touchpad appears dead in Hyprland: clicks register (button events pass through) but cursor motion is silently dropped. Cause: the kernel produces perfect Precision Touchpad reports, but libinput 1.31.1 discards every motion event as "kernel bug: Touch jump detected and discarded" because the pad reports pressure values of 0-1, which trips the contact stability check. Same pattern as the UX302LA entry in libinput's shipped 50-system-asus.quirks.

New install hooks wired into `install/config/all.sh` alphabetically. Migration `1777040047.sh` sources both install hooks and runs `limine-update` for existing B9406 installs.

A companion PR against `omacom-io/omarchy-iso` will follow, the live ISO itself still freezes on this panel without the Panel Replay fix baked into the ISO's boot cmdline, so installers hit the same wall before they can reach any of this.


mijuny:
Force-pushed a cleaner migration. The existing Copilot inline comments target code from an earlier suggested commit that's no longer on the branch, please re-review against the latest commit.

install/config/all.sh
```diff
@@ -48,6 +48,8 @@ run_logged $OMARCHY_INSTALL/config/hardware/intel/fix-wifi7-eht.sh
 run_logged $OMARCHY_INSTALL/config/hardware/dell/fix-xps-haptic-touchpad.sh
 
 run_logged $OMARCHY_INSTALL/config/hardware/asus/fix-audio-mixer.sh
+run_logged $OMARCHY_INSTALL/config/hardware/asus/fix-b9406-display.sh
+run_logged $OMARCHY_INSTALL/config/hardware/asus/fix-b9406-touchpad.sh
```
dhh: Let's use `fix-asus-ptl-b9406` here so it's clear at a glance beyond the code name.

dhh:
Excellent work on this. Very clean. Would be great to see if we can find others with Asus PTL machines that aren't the ExpertBook. I suspect they'll need some of the same fixes.

install/config/all.sh
mijuny: Done. Renamed both install hooks to `fix-asus-ptl-b9406-display.sh` and `fix-asus-ptl-b9406-touchpad.sh`, and updated the references in `install/config/all.sh` and the migration.

mijuny:
> Excellent work on this. Very clean. Would be great to see if we can find others with Asus PTL machines that aren't the ExpertBook. I suspect they'll need some of the same fixes.

Thank you! Means a lot! I will also keep my eye on these and see if new issues emerge with new ASUS users.

migrations/1777040047.sh
```diff
@@ -0,0 +1,22 @@
...
+
+  # Keep /etc/default/limine in sync with the drop-in. Install-time this is
+  # handled by install/login/limine-snapper.sh; migrations have to do it
+  # themselves. Idempotent: skip if the drop-in's header comment is already
+  # present (a unique-to-this-drop-in marker, so it doesn't collide with
+  # manual cmdline tweaks a user may have made).
+  HEADER='# ASUS ExpertBook B9406 (Panther Lake / Xe3) display workarounds'
+  if [[ -f $DROP_IN && -f $DEFAULT_LIMINE ]] && \
+       ! grep -Fxq "$HEADER" "$DEFAULT_LIMINE"; then
+    cat "$DROP_IN" | sudo tee -a "$DEFAULT_LIMINE" >/dev/null
```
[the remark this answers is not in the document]
mijuny: Good catch, applied. Using shell redirect (`sudo tee -a "$DEFAULT_LIMINE" < "$DROP_IN"`) instead of the `cat … |` pipeline so a missing/unreadable drop-in fails loudly.

mijuny:
> Excellent work on this. Very clean. Would be great to see if we can find others with Asus PTL machines that aren't the ExpertBook. I suspect they'll need some of the same fixes.

Just one more thing on this: if you have connections at ASUS, I'm happy to help test their PTL machines if they'd loan units. I can reach out to them myself too, but not sure how much weight that carries.

migrations/1777040047.sh
```diff
@@ -0,0 +1,22 @@
+echo "Apply display and touchpad fixes for ASUS ExpertBook B9406 (Panther Lake)"
```
dhh: Wondering if we even need this migration? You did a heroic effort to get it running. But I suspect that most other folks installing would not have. Thus would never be able to run this migration?

mijuny: That is true, we don't need it. Suppose anyone who fights their way to install 3.5 for PTL Expertbook doesn't need that migration script. And fresh installs are ok without. Pushed update.

install/config/hardware/asus/fix-asus-ptl-b9406-touchpad.sh
```diff
@@ -0,0 +1,23 @@
...
+# values of 0-1, confusing the contact stability check. Button events
+# still pass, so clicks register but motion does not.
+#
+# Mask the pressure axes with a quirks override, same pattern as the
+# Asus UX302LA entry in libinput's shipped 50-system-asus.quirks.
+
+if omarchy-hw-asus-expertbook-b9406; then
+  sudo mkdir -p /etc/libinput
+  sudo tee /etc/libinput/asus-expertbook-b9406.quirks >/dev/null <<EOF
+[ASUS ExpertBook B9406 Touchpad]
```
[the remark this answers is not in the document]
mijuny: Understood. The install hook is corrected to write a model-specific file `/etc/libinput/asus-expertbook-b9406.quirks` rather than touching `local-overrides.quirks`, as that file is used for local user/admin overrides.

Updated the PR description and the linked troubleshooting doc to use the correct path.

simon-scarlet:
I installed a fresh version of Omarchy 3.8.2 on a new Expertbook, but the touchpad was not working, exactly in the way described here. It seems to me that the quirks file is not being read. I had to put it in /etc/libinput/local-overrides.quirks instead, and remove the `MatchUdevType=touchpad` line. According to GPT-5.5:

```
MatchUdevType=touchpad does not match because the device is
tagged as ID_INPUT_MOUSE=1, not ID_INPUT_TOUCHPAD=1, despite
being ASCP1D80:00 093A:4F05 Touchpad.
```

I wonder why I'm having a different experience?

mijuny:
@simon-scarlet Thank you for the report, and sorry this slipped through. Your diagnosis was spot on. Fix is up here: #6093

PR #5441, opened 2026-04-24
Add Fireworks background to vantablack

stappmus:
## Summary
- add a new built-in Vantablack background at `themes/vantablack/backgrounds/4-fireworks.png`
- keep the existing numbered background ordering and theme-local asset layout
- preserve the original 4758x2976 PNG asset for review


dhh:
Thanks for the suggestion. It's a cool background, but I don't quite feel like it fits Vanta.

PR #5442, opened 2026-04-25
Add libnm dependency for chromium DRM compatibility

imanubdesigner:
_Include libnm to resolve issues with Widevine/DRM playback in Chromium-based browsers. This ensures that media streaming services function correctly regardless of the system's actual network manager._

This PR adds libnm to the default package list. While many minimalist installations might opt for alternative network managers (such as iwd or systemd-networkd), the chromium package and its derivatives on Arch Linux are compiled with a hard dependency on libnm for specific functionalities.

Most notably, the Widevine Content Decryption Module (CDM)—used by platforms like Netflix, Spotify, and Disney+—requires the presence of libnm to correctly initialize the network stack and verify device integrity. Without this library, users encounter playback errors on encrypted streams, despite having a functional internet connection. Including libnm ensures seamless multimedia compatibility without requiring the full NetworkManager daemon to be active.

### DIFF
--- a/install/omarchy-base.packages
+++ b/install/omarchy-base.packages
@@ -68,6 +68,7 @@ kvantum-qt5
 lazydocker
 lazygit
 less
+libnm
 libsecret
 libyaml
 libqalculate


dhh:
I can't replicate the problem. On a new Omarchy install, Netflix playback works fine?

imanubdesigner:
Hi, DHH. I had Omarchy 3.5.1 installed, and right after the first run (after Mako's notifications), I went to Spotify Web to listen to some music while I finished fixing the system. I noticed that even though I had DRM enabled in Chromium, it wouldn't start. 

So I went to Netflix, and it didn't work there either. I kept getting this error: 
Please visit chrome://settings/content/protectedcontent and make sure "sites can play protected content" is enabled
error 07702-1003

I should try again with the new ISO 3.6.0

imanubdesigner:
## Technical Update: Netflix & Widevine on Omarchy 3.6.0

I just did a fresh install with **Omarchy 3.6.0**. As soon as I logged into my Google account on **Chromium** to sync my bookmarks and extensions, I tried opening Netflix, but I ran into the same old issue.
Thinking I had the solution, I installed libnm and restarted Chromium, but to my surprise, it didn't work. I then uninstalled libnm and tried a different approach:
 
1. I went into Chromium settings and **disabled** the option for sites to play protected content.
 
2. I restarted the browser.

3. I **re-enabled** the protected content setting and restarted Chromium once more.

**It’s working now!** I noticed that right after the format, the WidevineCdm folder was completely empty, but now it’s finally populated. It seems like the browser just needed a moment (or a nudge) to actually download the plugin.

I’m sorry for the confusion earlier! I was convinced libnm was the fix because, by pure coincidence, it started working immediately after I installed it back on Omarchy 3.5.1. My apologies for the mix-up!

### A Quick Note on why that happened:
You're likely right—it was a timing issue. Chromium often triggers the **Widevine CDM** (Content Decryption Module) download shortly after the first launch or when a DRM-protected site (like Netflix) requests it. Toggling the setting likely forced the browser to check for updates and "wake up" the download manager.

imanubdesigner:
I can confirm, DHH, that the `libnm` package is completely unnecessary. 
After formatting my system and installing Omarchy 3.6.0, I opened:

`~/.config/chromium/WidevineCdm`

And it was empty. I logged into my Google account and waited — without visiting Netflix or Spotify Web — and the folder populated shortly after. 

I don't understand why it takes a while. 
I probably have (or had) a problem with my connection to that particular address

Sorry again for the confusion, and I take this opportunity to wish you a nice day and a wonderful Sunday!

PR #5445, opened 2026-04-25
Add support for updating the RGB color of Razer devices

hegjon:
At the moment the user needs to install the following packages:
  * `razer-cli`
  * `openrazer-driver-dkms`
  * `openrazer-daemon`

And the following needs to be configured:
```
$ systemctl --user enable openrazer-daemon.service
$ systemctl --user start openrazer-daemon.service
```

Detecting razer device:
```
$ lsusb -d 0x1532:
```

Tested on a **Razer Huntsman V3 Pro** keyboard

### DIFF
--- a/bin/omarchy-theme-set-keyboard
+++ b/bin/omarchy-theme-set-keyboard
@@ -2,3 +2,4 @@
 
 omarchy-theme-set-keyboard-asus-rog
 omarchy-theme-set-keyboard-f16
+omarchy-theme-set-keyboard-razer
--- a/bin/omarchy-theme-set-keyboard-razer
+++ b/bin/omarchy-theme-set-keyboard-razer
@@ -1,0 +1,8 @@
+#!/bin/bash
+
+RAZER_THEME=~/.config/omarchy/current/theme/keyboard.rgb
+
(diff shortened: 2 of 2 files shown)


dhh:
We need to automate the entire thing. Like we've done for Asus and Framework.

hegjon:
When should the detection happen? If the user plug-in the keyboard after the installation, should there be a check for the keyboard later?

openrazer have a list of all the supported devices that should be easy to parse with a script, would it make sense to also check against this list?

dhh:
Ahh, I see. I was assuming this was for a Razer laptop, not a separate keyboard. Yeah, good question. We could consider adding _Install > Hardware > Razer RGB_ or something like that.

hegjon:
47 of the devies are laptops with a supported keyboard, but the most of the devices are external keyboards and mices. Let me know if you want to add a _Install > Hardware_ sub-menu, then I can make the Razer RGB install step.

hegjon:
Another option could be to add udev rule, for instance:

```
ACTION=="add", SUBSYSTEM=="usb", ATTR{idVendor}=="1532", ATTR{idProduct}=="0013|0015|...", RUN+="omarchy-ask-for-install-openrazer"
```

But feel it might be too much hassle/work for something simple

PR #5456, opened 2026-04-26
Add local text-to-speech with Piper

PanAchy:
## Summary

Adds optional local text-to-speech via Piper.

Demo (fresh Omarchy VM): 

https://github.com/user-attachments/assets/d738f216-443c-4070-a9eb-1593f8985d26

- Install from `Install > AI > Text to Speech`
- Read selected text with `Super + Ctrl + R`
- Press the shortcut again to stop playback
- Configure via `~/.config/piper/config.toml`
- Shows a Waybar indicator only while reading
- Remove from `Remove > Text to Speech`

Piper runs locally/offline, installs through `uv` without building from AUR source, and keeps voices as separate files. The default `en_US-lessac-medium` voice is about 63MB, and users can browse alternative voices at https://rhasspy.github.io/piper-samples.

## Notes

This follows the same general integration pattern as Dictation through Voxtype (I used `uv tool install piper-tts` because the current AUR `piper-tts` package failed to build on a fresh Omarchy VM, while `piper-tts-bin` points at the archived rhasspy/piper repo. Happy to switch this to an OPR package if that’s preferred.)

The config supports voice path, input source, reading speed, extra Piper CLI args, and notification toggles.

### DIFF
--- a/bin/omarchy-menu
+++ b/bin/omarchy-menu
@@ -368,8 +368,9 @@ show_install_ai_menu() {
       echo ollama
   )
 
-  case $(menu "Install" "  Dictation\n󱚤  LM Studio\n󱚤  Ollama\n󱚤  Crush") in
+  case $(menu "Install" "  Dictation\n󰔊  Text to Speech\n󱚤  LM Studio\n󱚤  Ollama\n󱚤  Crush") in
   *Dictation*) present_terminal omarchy-voxtype-install ;;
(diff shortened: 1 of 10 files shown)


dhh:
Very cool, but not something I'd personally be using. I think we're better off wrapping this in a separate AUR package. Thanks for working on it 👍

PanAchy:
Thank you for the feedback!

PR #5462, opened 2026-04-27
Add laptop display mirror toggle + hotkey

timbelmon:
Added a check to see if any monitor is mirroring the internal display. If mirroring is detected, the script stops before disabling the screen.

### DIFF
--- a/bin/omarchy-hyprland-monitor-internal
+++ b/bin/omarchy-hyprland-monitor-internal
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 TOGGLE="internal-monitor-disable"
+MIRROR_TOGGLE="internal-monitor-mirror"
 
 enable() {
   if omarchy-hyprland-toggle-enabled "$TOGGLE"; then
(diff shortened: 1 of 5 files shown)


timbelmon:
Issue: https://github.com/basecamp/omarchy/issues/5457
First commit to the project—edits and feedback are welcome

dhh:
How does this situation occur? Why would you want to mirror an internal display if you're also closing that display?

timbelmon:
> How does this situation occur? Why would you want to mirror an internal display if you're also closing that display?

I have my internal display mirrored at all times for giving presentations with a projector. But I also have an external display at my desk which I plug into and then close my laptop. 
Usually this wouldn't be a problem if you keep updating your config for each use but of course people are lazy.

dhh:
Ahhh. I see. I think we should do two things: 1) Fix the crash bug in Hyprland (I alerted vaxry) and 2) Have a toggle for mirroring like we do for turning the internal display on/off.

timbelmon:
> Ahhh. I see. I think we should do two things: 1) Fix the crash bug in Hyprland (I alerted vaxry) and 2) Have a toggle for mirroring like we do for turning the internal display on/off.

How would a toggle work in regards to multiple outputs? Mirror the internal display to all of them? 
But yes checking for a toggle is much cleaner than this solution I agree.

dhh:
Yes, was thinking that again we treat the internal monitor as authoritative, mirror to everything that's connected. 9/10 times it'll just be a projector or a single extra monitor anyway.

timbelmon:
Updated as Suggested; works pretty good on my end. 
Please double-check if this is inline with your code requirements.

PR #5477, opened 2026-04-28
Add omarchy CLI

ryanrhughes:


ridhwaans:
Love this change and thank you for making this!
I'm also beginning to make a cli, i get that the design rationale will be different for big and small apps
can you share what resources you followed while making this, or conversed with AI? 
im following a new guide from https://clig.dev/

Issue #5468, opened 2026-04-27
`omarchy-cmd-screenshot`: hitting Print Screen occasionally locks input — Esc/clicks dead, cursor still moves


joshualambert:
### Symptom

Pressing `PRINT` occasionally leaves the session in a state where:

- Cursor still moves
- Audio (e.g. an active Zoom / Teams-for-linux call) keeps playing fine
- All windows *look* frozen
- Clicks register nowhere
- `Esc` does **not** dismiss slurp
- Hitting `PRINT` again does **not** trigger the `pkill slurp && exit 0` toggle at the top of the script

Only escape is `Ctrl+Alt+F2` → `pkill slurp hyprpicker` → `Ctrl+Alt+F1`.

### Hypothesis

`omarchy-cmd-screenshot` runs `hyprpicker -r -z` as a freeze layer in the background while `slurp` blocks for a region. `hyprpicker -z` paints its frozen snapshot on the **overlay** layer (level 3), which sits above slurp's **top** layer (level 2). When that overlay grabs keyboard focus first, slurp never sees `Esc`, and the script stays blocked on `SELECTION=$(slurp)` indefinitely. The compositor isn't dead — only the input route to slurp is. Audio is independent of the compositor, which matches.

### Repro

Intermittent. Happens more often with another app holding screen-capture state (e.g. Zoom or Teams-for-linux in a call). Single- and dual-monitor both seen.

### Workaround

Replaced the binding locally with a copy of `omarchy-cmd-screenshot` that drops `hyprpicker -r -z` entirely. Region selection is just slurp; the screen "lives" during selection but the deadlock path is gone. Reporting back after a week of soak.

### Related

- #1564 — region screenshot broke with `--freeze` (different stack, same "freeze layer breaks selection" pattern)
- #5001 — current `hyprpicker -z` doesn't reliably freeze captured frames
- v3.0.0 release notes: `--freeze` removed from hyprshot for being unstable
- v3.4.0: switched wayfreeze → hyprpicker for HDR

### System

- Omarchy 3.5.1
- Hyprland 0.54.3-2
- hyprpicker 0.4.6-3
- slurp 1.5.0-2 / grim 1.5.0-2 / satty 0.20.1-2
- xdg-desktop-portal-hyprland 1.3.11-4
- Monitors: 2× 2560×1440, scale 1.25, transform 0

joshualambert:
[...]

PR #5483, opened 2026-04-29
Potential solution for issue 5468

kurian-thomas:
Based on Issue: https://github.com/basecamp/omarchy/issues/5468

Currently in testing phase:
Reached out to issue owner via email to test or get testing details as I couldn't reproduce it locally

Solution details:
1. the screenshot tool instead of sleep between commands, it polls and waits for the hyprpicker to actually have the screen frozen (checks for layer to exist) before waiting for the slurp input


dhh:
Thanks for looking into this, but let's keep PRs off the list until they're ready 👍

kurian-thomas:
> Thanks for looking into this, but let's keep PRs off the list until they're ready 👍

Apologies, understood.

joshualambert:
Thanks for picking this up, @kurian-thomas. A note from my testing in case it's useful for a future revision:

The polling check works correctly on my box — hyprpicker's layer is visible in `hyprctl layers -j` within the first 50 ms iteration, so on a healthy run `wait_for_hyprpicker` returns immediately and behaves equivalently to the existing `sleep .1`. It's nicer hardening (deterministic vs. fixed sleep) but in my failing case the timing was already adequate.

The deadlock I reported in #5468 looks layer-level rather than timing: hyprpicker registers at level 3 (overlay) while slurp paints on level 2 (top), so when the overlay grabs keyboard focus first, `Esc` never reaches slurp. Polling for layer existence makes initialization order deterministic but doesn't change which layer ends up on top. So I don't think this PR alone closes the deadlock path — though it's a worthwhile cleanup of the `sleep .1`.

Happy to soak-test a follow-up that targets the layer ordering (e.g. dropping `-z`, or anything that prevents the overlay from holding keyboard focus while slurp is up). Appreciate you digging in.

kurian-thomas:
> Thanks for picking this up, @kurian-thomas. A note from my testing in case it's useful for a future revision:
> 
> The polling check works correctly on my box — hyprpicker's layer is visible in `hyprctl layers -j` within the first 50 ms iteration, so on a healthy run `wait_for_hyprpicker` returns immediately and behaves equivalently to the existing `sleep .1`. It's nicer hardening (deterministic vs. fixed sleep) but in my failing case the timing was already adequate.
> 
> The deadlock I reported in #5468 looks layer-level rather than timing: hyprpicker registers at level 3 (overlay) while slurp paints on level 2 (top), so when the overlay grabs keyboard focus first, `Esc` never reaches slurp. Polling for layer existence makes initialization order deterministic but doesn't change which layer ends up on top. So I don't think this PR alone closes the deadlock path — though it's a worthwhile cleanup of the `sleep .1`.
> 
> Happy to soak-test a follow-up that targets the layer ordering (e.g. dropping `-z`, or anything that prevents the overlay from holding keyboard focus while slurp is up). Appreciate you digging in.

Hi Josh, @joshualambert

I had a question. With the polling changes slurp always starts after hyprpicker layer. If hyperpicker (layer 3) overlay holds the lock then shouldn't this setup always surface the issue you are seeing. Ie the slurp code line not able to continue because no input was registered. Also if hyprpicker is holding the lock wont esc or a click trigger an unlock ?

PR #5485, opened 2026-04-29
Remove '--noconfirm' from package removal command

eloyesp:
Removed the '--noconfirm' option from the pacman command to allow the user check before remove, as the removal list includes dependencies. ([discussion](https://github.com/basecamp/omarchy/discussions/5484))


dhh:
Appreciate the suggestion, but I think this is rare enough to not be worth the added friction for the common case. You can always run pacman directly yourself to have full control.

PR #5496, opened 2026-04-29
Restore touchpad when laptop display is enabled

### LINKED ISSUE #5427: Trackpad disabled on closing my laptop lid
### System details

HP Pavalion - AMD RYZEN 4600H- GTX1650TI(disabled)

### What's wrong?

so my i close my lid, this command supr + ctr+h it disables my trackpad and I have to re enable it again, i belive it is caused by the new feature where it disables the laptop when closing the lid

heymaikol:
## Summary
- Add `omarchy-hw-recover-touchpad` to restore the touchpad when lid-close monitor toggling leaves it disabled.
- Call touchpad recovery from the lid-open `enable` path and from the monitor removal recovery hook alongside internal display and mirror recovery.
- Preserve intentional touchpad-disable state by respecting `~/.local/state/omarchy/toggles/hypr/touchpad-disabled.conf`.

Fixes #5427

## Verification
- `bash -n bin/omarchy-hyprland-monitor-internal bin/omarchy-hyprland-monitor-watch bin/omarchy-hw-recover-touchpad`
- `shellcheck bin/omarchy-hyprland-monitor-internal bin/omarchy-hyprland-monitor-watch bin/omarchy-hw-recover-touchpad`
- `git diff --check refs/remotes/basecamp/dev...HEAD`
- Dev-based behavior harness passed:
  - lid-open `on` restores trackpad
  - monitor removal recovery hook restores trackpad
  - user-intentionally-disabled trackpad remains disabled
  - missing trackpad device is a safe no-op
- Automated A/B harness passed:
  - `basecamp/dev` baseline does not restore the trackpad
  - this PR restores the trackpad from both `on` and monitor-watch recovery


dhh:
We should split this out into it's own setup. Just like we have with mirroring. Then call it in the recover hook.

mplaczek99:
@dhh I updated this to split trackpad recovery into its own `omarchy-hw-recover-touchpad` command, and called it from the internal monitor recover path.

I also kept it in the lid-open enable path because the issue is observed after a lid-open/close cycle, when the user has to re-enable the trackpad.

dhh:
Which laptops have we verified has this problem?

mplaczek99:
@dhh The reported laptop in #5427 is an HP Pavilion with an AMD Ryzen 4600H and GTX 1650 Ti (disabled).

I don’t have access to that exact machine, so I haven’t physically verified the lid-close behavior on the affected hardware. I verified with an A/B test: upstream does not restore the trackpad from `on`/`recover`, while this PR does, and it still respects `touchpad-disabled.conf` for intentionally disabled trackpads.

dhh:
Need to ensure we test this PR on a machine that has the issue.

proutkarsh3104:
> Need to ensure we test this PR on a machine that has the issue.

Hey @dhh , when it would be merged, as the issue still persist?

proutkarsh3104:
hey is there any updates on this the issue still persist

PR #5498, opened 2026-04-30
Replace deprecated sainnhe.everforest VS Code extension with reesew.everforest-theme

cfulton:
The sainnhe.everforest extension is no longer maintained. The upstream repo (https://github.com/sainnhe/everforest-vscode) was archived by the author.

This PR replaces it with reesew.everforest-theme, a maintained fork published from https://github.com/reese/everforest-vscode.

- Updates `themes/everforest/vscode.json` to reference the new extension
- Adds a migration that uninstalls the old extension across all supported VS Code variants (code, code-insiders, codium, cursor), then refreshes the theme if the user is currently on everforest to install the replacement automatically

### DIFF
--- a/migrations/1777511459.sh
+++ b/migrations/1777511459.sh
@@ -1,0 +1,36 @@
+echo "Replace deprecated sainnhe.everforest VSCode extension with reesew.everforest-theme"
+
+# Background:
+#   The original "sainnhe.everforest" extension is no longer maintained — its
+#   upstream repo (https://github.com/sainnhe/everforest-vscode) was archived
+#   by the author, so it receives no updates or fixes.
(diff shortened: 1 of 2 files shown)


dhh:
Appreciate the suggestion, but it'd rather save this for when/if there is a need to update. I doubt there will be. Themes should generally speaking be fine as frozen.

cfulton:
> Appreciate the suggestion, but it'd rather save this for when/if there is a need to update. I doubt there will be. Themes should generally speaking be fine as frozen.

The issue was that this seemed to be the best solution to remove all the extension warnings. 

Anyone using VSCode while the current theme is Everforest, they'll see the following:

screenshot-2026-04-30_18-54-19

dhh:
Ah yeah that doesn't look good.

Issue #1819, opened 2025-09-19
Unable to unlock screen using fingerprint authentication


matheorism:
### System details

AMD Ryzen 9 PRO 7940HS, AMD Radeon RX 6550M, Omarchy 3.0.1, Hyprland 0.51.0,

### What's wrong?

Unlocking the screen using fingerprint authentication is not working when I upgraded to **Omarchy v3.0.1**. I tested `fprintd` and it seems to be working fine.

```shell
sudo fprintd-verify $USER
Using device /net/reactivated/Fprint/Device/0
Listing enrolled fingers:
 - #0: right-index-finger
Verify started!
Verifying: right-index-finger
Verify result: verify-match (done)
```

Please see `hyprlock` log trace below:
```shell
[LOG] output eDP-1 name eDP-1
[LOG] output 61 description Lenovo Group Limited 0x4146  (eDP-1)
[LOG] output 61 make Lenovo Group Limited model 0x4146
[LOG] output 61 done
[LOG] Skipping screencopy
[LOG] Running on Hyprland
[LOG] Resources gathered after 51 milliseconds
[LOG] Locking session
[LOG] Starting fade in
[TRACE] [core] got poll event
[LOG] fprint: using device path /net/reactivated/Fprint/Device/0
[TRACE] [core] got poll event
[LOG] Got fractional scale: 200.0%
[LOG] configure with serial 13480
[LOG] Configuring surface for logical [Vector2D: x: 1920, y: 1200] and pixel [Vector2D: x: 3840, y: 2400]
[LOG] Missing screenshot for output eDP-1
[TRACE] Updating placeholder text:   Enter Password 󰈷 
[TRACE] Requesting new placeholder asset: placeholder:  Enter Password 󰈷 943061673709280.8039215686274510.83921568627450980.95686274509803931
[TRACE] Requesting label resource placeholder:  Enter Password 󰈷 943061673709280.8039215686274510.83921568627450980.95686274509803931
[TRACE] Processing requested resourceID placeholder:  Enter Password 󰈷 943061673709280.8039215686274510.83921568627450980.95686274509803931
[TRACE] Framebuffer created, status 36053
[TRACE] Framebuffer created, status 36053
[TRACE] Framebuffer created, status 36053
[TRACE] [core] got poll event
[WARN] fprint: could not claim device, [net.reactivated.Fprint.Error.PermissionDenied] Not Authorized: net.reactivated.fprint.device.enroll
[TRACE] timer thread firing
[...]

PR #5504, opened 2026-04-30
fix: add hyprlock PAM config to omarchy-setup-fingerprint

### LINKED ISSUE #5503: omarchy-setup-fingerprint: lock screen fingerprint auth not wired up despite claiming support
## Summary

`omarchy-setup-fingerprint` prints "You can use your fingerprint for sudo, polkit, and lock screen (Super + Escape)" and sets `fingerprint:enabled = true` in `hyprlock.conf` (causing the fingerprint icon to appear on the lock screen), but never adds `pam_fprintd.so` to `/etc/pam.d/hyprlock`. The result is that the icon shows but touching the reader does nothing.

## Root cause

`setup_pam_config` in the script configures `/etc/pam.d/sudo` and `/etc/pam.d/polkit-1`, but omits `/etc/pam.d/hyprlock`.

## Fix

Add a third block to `setup_pam_config`:

```bash
# Configure hyprlock
if ! grep -q pam_fprintd.so /etc/pam.d/hyprlock; then
  print_info "Configuring hyprlock for fingerprint authentication..."
  sudo sed -i '1i auth        sufficient  pam_fprintd.so' /etc/pam.d/hyprlock
fi
```

And the corresponding removal in `remove_pam_config`:

```bash
# Remove from hyprlock
if grep -q pam_fprintd.so /etc/pam.d/hyprlock; then
  print_info "Removing fingerprint authentication from hyprlock..."
  sudo sed -i '/pam_fprintd\.so/d' /etc/pam.d/hyprlock
fi
```

## Steps to reproduce

1. Run `omarchy-setup-fingerprint` on a system with a working fingerprint reader
2. Lock the screen with Super+Ctrl+L
3. Place finger on reader — nothing happens despite the fingerprint icon being visible

## Verified fix

Manually running `sudo sed -i '1i auth        sufficient  pam_fprintd.so' /etc/pam.d/hyprlock` resolves the issue immediately.

lrettig:
## Problem

`omarchy-setup-fingerprint` adds `pam_fprintd.so` to `/etc/pam.d/sudo` and `/etc/pam.d/polkit-1`, but not to `/etc/pam.d/hyprlock`. The result is that the fingerprint icon appears on the lock screen (because `fingerprint:enabled = true` is set in `hyprlock.conf`) but touching the reader does nothing — hyprlock's PAM chain has no fingerprint entry.

The script even prints _"You can use your fingerprint for sudo, polkit, and lock screen"_, but the lock screen claim was never true.

Closes #5503
Related: #1819

## Solution

Add a hyprlock block to both `setup_pam_config` and `remove_pam_config`, matching the pattern already used for sudo and polkit. The block is guarded by a file-existence check (consistent with how the polkit block works) since `/etc/pam.d/hyprlock` was itself only added recently via #2955.

## Testing

Verified manually: after running `sudo sed -i '1i auth        sufficient  pam_fprintd.so' /etc/pam.d/hyprlock`, lock screen fingerprint auth works immediately on a system where `omarchy-setup-fingerprint` had already been run.

### DIFF
--- a/bin/omarchy-setup-fingerprint
+++ b/bin/omarchy-setup-fingerprint
@@ -53,6 +53,12 @@ password  required pam_unix.so
 session   required pam_unix.so
 EOF
   fi
+
+  # Configure hyprlock
+  if [[ -f /etc/pam.d/hyprlock ]] && ! grep -q 'pam_fprintd.so' /etc/pam.d/hyprlock; then
+    print_info "Configuring hyprlock for fingerprint authentication..."
+    sudo sed -i '1i auth        sufficient  pam_fprintd.so' /etc/pam.d/hyprlock
+  fi
 }
 
 add_hyprlock_fingerprint_icon() {
@@ -78,6 +84,12 @@ remove_pam_config() {
   if [[ -f /etc/pam.d/polkit-1 ]] && grep -Fq 'pam_fprintd.so' /etc/pam.d/polkit-1; then
     print_info "Removing fingerprint authentication from polkit..."
     sudo sed -i '/pam_fprintd\.so/d' /etc/pam.d/polkit-1
+  fi
+
+  # Remove from hyprlock
+  if [[ -f /etc/pam.d/hyprlock ]] && grep -q 'pam_fprintd.so' /etc/pam.d/hyprlock; then
+    print_info "Removing fingerprint authentication from hyprlock..."
+    sudo sed -i '/pam_fprintd\.so/d' /etc/pam.d/hyprlock
   fi
 }
 


dhh:
I can't replicate this problem. When I add the fingerprint reader to my Framework, it works on the lock screen.

lrettig:
Thanks for taking a look. I just unenrolled and re-enrolled fingerprint auth from scratch and it worked this time. (I also undid the change to /etc/pam.d/hyprlock).

Very odd. For the first few days after setting up the fingerprint reader, it definitely did not work on the lock screen.

I'll keep an eye on this and if it recurs I'll reopen with additional diagnostics.

PR #5511, opened 2026-04-30
Fix internal monitor disable for systems where built-in display is not eDP-1

### LINKED ISSUE #5443: Laptop display toggle fails on ASUS TUF Dash F15 (Internal monitor named eDP-2)
### System details

Hardware: ASUS TUF Dash F15 FX517ZE  CPU/iGPU: Intel Core i7-12650H (Intel UHD Graphics)  dGPU: NVIDIA GeForce RTX 3050 Ti Mobile  OS: Omarchy 3.6.0 (Kernel 6.19.13-arch1-1)  Compositor: Hyprland 0.54.3

### What's wrong?

When running the toggle script, the command returns "ok" but the screen remains active.

- Output of `hyprctl monitors all` shows the internal display as `eDP-2`.

- The state file `~/.local/state/omarchy/toggles/hypr/internal-monitor-disable.conf` contains the incorrect instruction `monitor=eDP-1`,disable.

**Steps to Reproduce**

1. Connect an external monitor.

2. Attempt to disable the laptop screen via shortcut or `omarchy-hyprland-monitor-internal off`.

3. Observe that `eDP-2` remains active.

17xande:
## Problem

`default/hypr/toggles/internal-monitor-disable.conf` hardcodes `monitor=eDP-1,disable`. On laptops with hybrid GPUs (e.g. Intel iGPU + NVIDIA dGPU), the iGPU claims `eDP-1` as a disconnected connector while the actual built-in panel is enumerated as `eDP-2` or higher. Hyprland silently accepts the config but does nothing, so the toggle fires the notification but the display stays on.

Fixes #5443.

## Solution

Detect the internal monitor name dynamically at runtime using `hyprctl monitors -j`, following the same pattern already used in `omarchy-hyprland-monitor-internal-mirror`. The `disable()` function now writes the toggle flag file directly with the detected name instead of copying the static conf.

The static `default/hypr/toggles/internal-monitor-disable.conf` is removed — it is no longer needed.

## Testing

Tested on a system where the built-in display is `eDP-2` (Intel + NVIDIA hybrid GPU). Both the lid switch and the keyboard shortcut correctly disable and re-enable the internal monitor.

## AI Disclosure
Claude assisted with identifying the pattern in the `omarchy-hyprland-monitor-internal-mirror` script. I wasn't 100% happy with it's attempt at fixing it so I refactored it to match the style in the mirror script.

### DIFF
--- a/bin/omarchy-hyprland-monitor-internal
+++ b/bin/omarchy-hyprland-monitor-internal
@@ -1,7 +1,11 @@
 #!/bin/bash
 
 TOGGLE="internal-monitor-disable"
+TOGGLE_FLAG="$HOME/.local/state/omarchy/toggles/hypr/$TOGGLE.conf"
 MIRROR_TOGGLE="internal-monitor-mirror"
+
+# Get internal monitor name dynamically
+INTERNAL=$(hyprctl monitors -j | jq -r '.[] | select(.name | contains("eDP")).name' | head -n 1)
 
 enable() {
   if omarchy-hyprland-toggle-enabled "$TOGGLE"; then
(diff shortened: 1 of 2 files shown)


timbelmon:
Glad to see the dynamic detection actually working :+1:

PR #5517, opened 2026-04-30
Set Helix theme to terminal colors during install

spitfire05:
This PR adds some logic to Helix Editor install procedure, setting it's theme to "base16_transparent", which essentially makes it use terminal color palette and integrate nicely with Omarchy theme switcher kinda like nvim does (see attached screenshots).

**NOTE**: We *do* modify user config here, as Helix currently does not support any kind of config imports (at least not that I know of). Not sure if that is allowed practice in Omarchy..

~~**TODO**: It'd be nice to add migration for this along the lines of:~~

```bash
echo "Reinstalling Helix to set Omarchy theme"

if omarchy-pkg-present helix; then
  omarchy-install-helix
fi
```

~~.. but the migrations system is at this time beyond me (sorry I am new here).~~
edit: I think I figured it out :sweat_smile: 

## Before this PR:

screenshot-2026-05-01_01-12-47

## After

screenshot-2026-05-01_10-25-24
screenshot-2026-05-01_10-25-45
screenshot-2026-05-01_10-26-20

### DIFF
--- a/bin/omarchy-install-helix
+++ b/bin/omarchy-install-helix
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+# Install Helix and configure it to use the current Omarchy theme.
+
+CONFIG_FILE="$HOME/.config/helix/config.toml"
+CONFIG_TEXT='theme = "base16_transparent" # Set by Omarchy install. Base16 theme makes Helix use terminal colors.'
+
+echo "Installing Helix..."
+omarchy-pkg-add helix
+
+mkdir -p $HOME/.config/helix
+
+if [[ -f $CONFIG_FILE ]]; then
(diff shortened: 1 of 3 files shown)


spitfire05:
Just noticed that using `base16_transparent` makes the line numbers unreadable on some omarchy themes. After more testing I've found `term16_dark` to work best (even on light themes like catpp latte).

Ultimately, there should be probably a theme template specifically for helix, but I can't create one right now, and this PR gets us like 90% of "ideal" theming.

lukehsiao:
https://github.com/basecamp/omarchy/pull/5530 contains a theme template specifically for helix. What do you think?

dhh:
Going with #5530 👍

PR #5525, opened 2026-05-01
Implement non-uniform brightness step adjustments

boobachad:
Add logic for non-uniform step size adjustments based on current brightness level.


boobachad:
currently on brightness level 0 my laptop goes black. is it intended should we cap it. like 1 is the final

PR #5533, opened 2026-05-01
Add charging providers

spencerbull:
## Summary
- Adds omarchy-battery-policy as the packaged backend for battery charging controls.
- Adds Setup > Power > Battery Charging when supported battery controls are available.
- Installs the new package for existing users via migration and includes it for fresh installs.
- Removes the temporary dev-only battery policy test helper from the PR.
## Notes
- Battery charging controls are hidden when no supported battery policy provider is available.
- Custom charge limits are only applied when the Custom profile is selected; other profiles ignore stored custom limits.
## Testing
- Verified omarchy-menu shell syntax with bash -n.
- Verified package script syntax with bash -n.
- Verified git diff --check.
- Confirmed migration installs omarchy-battery-policy when missing.

Dependent on: https://github.com/omacom-io/omarchy-pkgs/pull/88
Ref: #5144 

CC: @spaceXrace @dhh


spencerbull:
image
image

dhh:
This is excellent, but I'd like to expose the settings like we do with the power profile. So we can control the whole thing from within the omarchy menu instead of shelling out. So that means you expose a cli that can give us the options, current value, and ability to set. Just like do with the power profiles.

spencerbull:
Okay yea that makes sense. I will get it integrated into the menu.

spencerbull:
image

image

image

PR #5534, opened 2026-05-01
Use packaged Dell XPS touchpad haptics

jacob-vincent-mink:
## Summary
- move Dell XPS Synaptics haptics out of `omarchy` and into the new `dell-xps-touchpad-haptics` package in OPR
- install that package automatically on supported Dell XPS systems
- replace the old setup flow with `Toggle > Hardware > Touchpad Haptics`
- migrate existing installs off the legacy in-repo service and onto the packaged service

## Details
- keep hardware detection in `omarchy` with `omarchy-hw-dell-xps-haptic-touchpad`
- add package install wiring via `install/packaging/dell-xps-touchpad-haptics.sh`
- remove the old in-tree daemon/update/setup commands and legacy service bootstrap
- add `omarchy-toggle-touchpad-haptics`, which cycles these preset levels:
  - `low`
  - `mid`
  - `high`
- use normal Omarchy notifications when the level changes
- expose the control at:
  - `Toggle > Hardware > Touchpad Haptics`

## Migration
- keep a single migration that:
  - disables and removes the old `dell-xps-haptic-touchpad.service`
  - removes the old env/rule/override wiring
  - installs `dell-xps-touchpad-haptics`
- service enable/restart is owned by the package install hook in `omarchy-pkgs`

## Dependency
- depends on `omacom-io/omarchy-pkgs#86`

## Testing
- `bash -n bin/omarchy-hw-dell-xps-haptic-touchpad bin/omarchy-toggle-touchpad-haptics bin/omarchy-menu install/packaging/dell-xps-touchpad-haptics.sh migrations/1777653181.sh`
- `git diff --check`

## Hardware test
- install the package from `omarchy-pkgs`
- run:
  - `PATH="$HOME/src/github.com/jacob-vincent-mink/omarchy/bin:$PATH" omarchy-toggle-touchpad-haptics`
- verify the menu entry appears on supported Dell hardware
- verify the toggle cycles through the three presets and the packaged daemon logs the changes


dhh:
This is great! But I'd like to do the following:

- Extract omarchy-haptic-touchpad into a new package called dell-xps-touchpad-haptics that we put in [OPR](https://github.com/omacom-io/omarchy-pkgs).
- This new package should expose dell-xps-touchpad-haptics cli which offer something like `get`, `set` which we can then call from Omarchy
- We install this new package if the synaptic haptic engine is detected on new install and via migration
- We add _Toggle > Hardware > Touchpad Haptics_ which will cycle through 3 pretermined levels: low, mid, high with a notification for feedback.

jacob-vincent-mink:
Thanks for the feedback @dhh! 30/60/90 works for low/medium/high? Or - any other opinion on which levels made the touchpad feel better?

I’ll get on this after I finish travels in a day.

dhh:
I haven't actually tried, but let's just try with that? I'll tweak them after trying it on the real hardware.

jacob-vincent-mink:
@dhh I've updated https://github.com/omacom-io/omarchy-pkgs/pull/86 towards your feedback and have updated this PR to use it.

After messing around with the numbers on my system, I found 10/50/100 to be more "obvious" when toggling, but you might find different.

spencerbull:
Tested on XPS 14 and 16! Feels good. In my opinion, I think high is good with the most feedback to set as the default.

jacob-vincent-mink:
Ty @dhh changes noted!

PR #5537, opened 2026-05-01
Q to Quit

mplaczek99:
## Summary
Replaces the default close-window shortcut from Super+W to Super+Q in the Hyprland bindings.


ryanrhughes:
This was discussed before in #991, and kept `SUPER+W` because “close window” maps to macOS-style `Cmd+W`.

`SUPER+Q` generally kills every instance of the app instead of just the active window so keeping the phoenetic association of "close window" with `SUPER+W` feels like that's still the way to go. 

With that said, it's an easy override to toss in you `~/.config/hypr/binding.conf` if you'd prefer `SUPER+Q`

PR #5542, opened 2026-05-02
Remove only preinstalled webapps during preinstall cleanup

### LINKED ISSUE #4830: omarchy-remove-preinstalls removes user-created webapps
### System details

Omarchy 3.4

### What's wrong?

omarchy-webapp-remove-all (called by omarchy-remove-preinstalls) removes all webapps installed via omarchy-webapp-install, including ones created by the user after  setup. It identifies webapps solely by the presence of Exec=omarchy-launch-webapp or Exec=omarchy-webapp-handler in .desktop files, with no distinction between preinstalled and user-created apps.

  Steps to reproduce:
  1. Complete Omarchy setup
  2. Create a custom webapp with omarchy-webapp-install
  3. Run omarchy-remove-preinstalls from the launcher
  4. Custom webapp is removed along with the preinstalled ones

  Expected behavior:
  Only the webapps that ship with Omarchy by default (defined in install/packaging/webapps.sh) should be removed.

  Actual behavior:
  All webapps matching the exec pattern are removed, including user-created ones.

  Possible fix:
  Track preinstalled webapps by name (from webapps.sh) and only remove those specifically, rather than pattern-matching all webapp desktop files.

afurm:
## Summary

- Add a shared list of Omarchy’s preinstalled webapps
- Reuse that list for default webapp installation
- Make `omarchy-remove-preinstalls` remove only those default webapps instead of every launcher created with `omarchy-webapp-install`

Closes #4830

## Why

`omarchy-remove-preinstalls` previously called `omarchy-webapp-remove-all`, which deletes every desktop file using `omarchy-launch-webapp` or an `omarchy-webapp-handler`. That also removes user-created webapps, even though the menu action says it removes preinstalled applications.

## Validation

- `bash -n bin/omarchy-remove-preinstalls bin/omarchy-webapp-remove bin/omarchy-webapp-remove-all install/packaging/webapps.sh install/packaging/webapps-data.sh`
- `git diff --check`
- Hermetic removal test confirmed a preinstalled `HEY` launcher is removed while a user-created `Custom App.desktop` is preserved

### DIFF
--- a/bin/omarchy-remove-preinstalls
+++ b/bin/omarchy-remove-preinstalls
@@ -5,7 +5,15 @@
 if gum confirm "Are you sure you want to remove all preinstalled web apps, TUI wrappers, and desktop applications?"; then
   echo -e "Removing preinstalled Omarchy applications...\n"
 
-  omarchy-webapp-remove-all
+  source "$OMARCHY_PATH/install/packaging/webapps-data.sh"
+
+  preinstalled_webapps=()
+  for webapp in "${OMARCHY_PREINSTALLED_WEBAPPS[@]}"; do
+    IFS="|" read -r app_name _ <<< "$webapp"
+    preinstalled_webapps+=("$app_name")
+  done
+
+  omarchy-webapp-remove "${preinstalled_webapps[@]}"
   omarchy-tui-remove-all
 
   cp ~/.config/hypr/bindings.conf ~/.config/hypr/bindings.conf.bak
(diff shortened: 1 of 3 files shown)


dhh:
Appreciate the work on this, but I don't think it's worth the complexity introduced. The Remove > Preinstalls setup is intended to be run by someone right after they've done the install. Not long after they've been using the computer and installed their own web apps.

PR #5544, opened 2026-05-02
Quiet lid-close display toggle when undocked

### LINKED ISSUE #5532: "Can't disable the only active display" notification when I open my laptop lid
### System details

Omarchy 3.6.0

### What's wrong?

When I open my laptop lid, I keep getting the notification _Can't disable the only active display_ even though I did not activate the Laptop Display toggle. The `disable` function is called whenever the lid opens and the computer is not docked.

https://github.com/basecamp/omarchy/blob/cdb9694360a031c15ac194be01ff9896442d5d71/default/hypr/bindings/utilities.conf#L34

https://github.com/basecamp/omarchy/blob/cdb9694360a031c15ac194be01ff9896442d5d71/bin/omarchy-hyprland-monitor-internal#L18

afurm:
## Summary

- Add an `off-if-external` mode to `omarchy-hyprland-monitor-internal`
- Use it for the lid-close Hyprland binding
- Preserve the existing warning for manual attempts to disable the only active display

Fixes #5532

## Why

The lid-close binding used the same `off` path as the manual laptop-display toggle. When no external monitor was connected, opening or closing the lid could trigger the “Can’t disable the only active display” notification even though the user did not manually request that toggle.

The new `off-if-external` path keeps lid-close behavior active for docked setups, but silently no-ops when no external monitor is connected.

## Validation

- `bash -n bin/omarchy-hyprland-monitor-internal`
- `git diff --check`
- Hermetic behavior harness covering:
  - undocked lid-close no-op
  - manual `off` still shows the warning
  - docked `off-if-external` still disables the laptop display

### DIFF
--- a/bin/omarchy-hyprland-monitor-internal
+++ b/bin/omarchy-hyprland-monitor-internal
@@ -1,7 +1,7 @@
 #!/bin/bash
 
-# omarchy:summary=Enable, disable, toggle, or recover the internal laptop display
-# omarchy:args=<on|off|toggle|recover>
+# omarchy:summary=Enable, disable, toggle, conditionally disable, or recover the internal laptop display
+# omarchy:args=<on|off|off-if-external|toggle|recover>
 
 TOGGLE="internal-monitor-disable"
 TOGGLE_FLAG="$HOME/.local/state/omarchy/toggles/hypr/$TOGGLE.conf"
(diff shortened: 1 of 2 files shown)


dhh:
Solved via [3d52764](https://github.com/basecamp/omarchy/commit/3d527643108d1a37edce265de7a3cdf2281093bd)

PR #5548, opened 2026-05-02
Harden browser managed policy permissions

### LINKED ISSUE #5547: Browser managed policy directories are world-writable
### System details

Omarchy dev branch at 3fabdf73; Chromium/Brave browser policy setup; hardware independent

### What's wrong?

Omarchy creates Chromium and Brave managed policy directories under `/etc`, then makes them writable by every local user.

Affected files:

- `install/config/theme.sh`
- `migrations/1757147211.sh`

Current behavior:

```bash
sudo mkdir -p /etc/chromium/policies/managed
sudo chmod a+rw /etc/chromium/policies/managed

sudo mkdir -p /etc/brave/policies/managed
sudo chmod a+rw /etc/brave/policies/managed
```

These directories are privileged browser managed-policy inputs. Chromium's Linux policy docs say managed policy files/directories should not be writable by non-admin users.

Steps to recreate:

1. Run the Omarchy install path or the existing migration that creates browser policy directories.
2. Check permissions:

   ```bash
   ls -ld /etc/chromium/policies/managed /etc/brave/policies/managed
   ```

3. Notice that non-root users can write files into those directories.

Expected behavior:

The managed policy directories should remain root/admin writable only, for example root-owned mode `755`. Omarchy can still update `color.json` by writing it through `sudo`.

Actual behavior:

The directories are made writable by everyone so `omarchy-theme-set-browser` can write `color.json` without sudo.

Suggested fix:

- Create `/etc/chromium/policies/managed` and `/etc/brave/policies/managed` with root ownership and mode `755`.
- Write `color.json` via `sudo tee`.
- Set `color.json` to root-owned mode `644`.
- Add a migration to repair permissions on existing installs.

I have a patch prepared for this.

afurm:
## Summary

Fixes #5547.

This keeps Chromium and Brave managed policy directories from being writable by every local user, while preserving no-sudo theme switching.

## Changes

- Create `/etc/chromium/policies/managed` and `/etc/brave/policies/managed` as `root:root` mode `755`.
- Pre-create Omarchy’s managed `color.json` as `root:wheel` mode `664`.
- Keep `omarchy-theme-set-browser` sudo-free by only overwriting the existing `color.json`.
- Remove stale `color.json` paths during migration before recreating them, so old symlinks from the previous world-writable directory state are not followed.
- Update the older migration path to use the same hardened directory/file setup.

## Validation

- `bash -n bin/omarchy-theme-set-browser install/config/theme.sh migrations/1757147211.sh migrations/1777653181.sh`
- `git diff --check`
- Focused shell checks for:
  - sudo-free theme writer
  - stale `color.json` symlink removal
  - `root:root` policy directories
  - `root:wheel` mode `664` policy file setup

### DIFF
--- a/bin/omarchy-theme-set-browser
+++ b/bin/omarchy-theme-set-browser
@@ -4,6 +4,18 @@
 
 CHROMIUM_THEME=~/.config/omarchy/current/theme/chromium.theme
 
+write_browser_policy() {
+  local policy_dir="$1"
+  local policy_file="$policy_dir/color.json"
+
+  if [[ ! -w $policy_file ]]; then
+    echo "Cannot write $policy_file. Run omarchy-migrate to repair browser policy permissions." >&2
+    return 1
+  fi
+
+  printf '{"BrowserThemeColor": "%s", "BrowserColorScheme": "device"}\n' "$THEME_HEX_COLOR" >"$policy_file"
+}
+
 if omarchy-cmd-present chromium || omarchy-cmd-present brave || omarchy-cmd-present brave-origin-beta; then
   if [[ -f $CHROMIUM_THEME ]]; then
     THEME_RGB_COLOR=$(<$CHROMIUM_THEME)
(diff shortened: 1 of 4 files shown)


bin/omarchy-theme-set-browser
```diff
@@ -4,6 +4,17 @@
 
 CHROMIUM_THEME=~/.config/omarchy/current/theme/chromium.theme
 
+write_browser_policy() {
+  local policy_dir="$1"
+  local policy_file="$policy_dir/color.json"
+
+  sudo install -d -m 755 -o root -g root "$policy_dir"
```
dhh: This won't work without needing sudo for theme changes. Which isn't reasonable for a good flow.

afurm: Adjusted this so theme changes do not require sudo anymore.

The install/migration path now makes the managed policy directories owned by the Omarchy user with mode `755`, so other local users cannot write there, but `omarchy-theme-set-browser` can update `color.json` directly during theme changes.

The migration also unlinks any existing `color.json` before recreating it, so stale symlinks from the previous world-writable directory state are removed rather than followed.

PR #5553, opened 2026-05-02
Complete native support for Helium browser

ahmedkall:
Helium (a privacy-focused Chromium fork) was already being explicitly matched by `omarchy-launch-webapp` and Hyprland's `browser.conf` window rules, but it lacked native Wayland flags and dynamic theme injection.

This PR brings it to full parity with Brave and Chromium using the exact same mechanisms:
- Seeds `~/.config/helium-flags.conf` to ensure native Wayland rendering and loads the default `copy-url` extension.
- Wires up `omarchy-theme-set-browser` to inject the theme hex into `/etc/helium/policies/managed/color.json`.
- Includes a standard migration to seed the config file for existing installs.

There are no new `omarchy-install-*` or `omarchy-refresh-*` commands added. If a user simply installs the binary, it seamlessly picks up the Omarchy aesthetic and behavior out-of-the-box.


dhh:
Can you update for latest on dev? We have a new browser install + default flow.

bin/omarchy-install-browser
```diff
@@ -86,8 +86,17 @@ zen)
   setup_firefox_wayland
   announce_browser_installed "Zen"
   ;;
+helium)
+  echo "Installing Helium..."
+  omarchy-pkg-aur-add helium-browser-bin || exit 1
+
+  setup_policy_directory /etc/chromium/policies/managed
```
dhh, line 104: Are you sure it uses the same directory for policies? All the other derivatives of Chromium use their own? This is what gives us theme switching.

ahmedkall: I verified with strings on the Helium binary. The only policy path hardcoded is /etc/chromium/policies. Helium never patched it to a custom directory like Brave/Chrome/Edge did.

CaioBianchi:
This feature would be awesome!

ahmedkall:
@dhh @ryanrhughes should be ready to go.

phucisstupid:
It would be great to have helium as preinstall browser

Adityavijayvargiya01:
Is there any update on this , helium is a great  browser. Its a shame I can’t use it on omarchy because theming glitch which makes tabs weird

CaioBianchi:
Any updates on this @dhh ? :)

PR #5571, opened 2026-05-04
Add "Sleep Displays" to System menu

jondkinney:
## Summary
- Adds a **Sleep Displays** entry to the System menu (`default/omarchy/omarchy-menu.jsonc`), between Lock and Suspend
- New helper `bin/omarchy-cmd-sleep-displays` powers off all monitors via DPMS without locking or suspending the session — any input wakes them
- Helper detaches the DPMS dispatch with a 1s delay so trailing input from closing the launcher (key release, focus change) doesn't immediately wake the displays

## Motivation
There's no built-in way to manually sleep just the displays — equivalent to macOS's "sleep displays" / Ctrl+Shift+Power. `Lock` brings up the lock screen, `Suspend` puts the whole machine to sleep. This sits between them: blank the screens, keep everything else running, wake on any input.

## Notes
Rebased onto `omarchy-4` and reworked for the 4.x code path:
- Menu entry is now a row in the JSONC menu (`system.sleep-displays`) rather than a `show_system_menu` case in the old bash `omarchy-menu`.
- Uses the current Lua dispatch form `hyprctl dispatch 'hl.dsp.dpms({ action = "disable" })'`, matching `omarchy-brightness-display`.

## Test plan
- [x] Open the Omarchy menu → System → Sleep Displays — displays power off, wake on key/mouse input
- [x] Searchable via the menu (keywords: monitor / dpms / screen / off / blank)
- [x] Verify `omarchy-cmd-sleep-displays` works from a terminal too

### DIFF
--- a/bin/omarchy-cmd-sleep-displays
+++ b/bin/omarchy-cmd-sleep-displays
@@ -1,0 +1,8 @@
+#!/bin/bash
+
+# omarchy:summary=Power off all monitors via DPMS; any input wakes them.
+
+# Detach with a short delay so trailing input from the launcher (key release,
+# focus change) doesn't immediately wake the displays.
+(sleep 1 && hyprctl dispatch 'hl.dsp.dpms({ action = "disable" })') &
+disown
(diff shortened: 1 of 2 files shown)


dhh:
What are you using this for? Because even if the displays are off, the hypridle will still lock the machine too? So how does it differ from just locking?

jondkinney:
@dhh for as long as I can remember (20 years?) I've set the lower right corner of my machine to "sleep displays". It helps the monitors power off more readily when they for sure know no input is coming. So my usual workflow is to turn the displays off before I head up to bed from my home office so the monitor doesn't sense any signal and fully turns off. On macOS it's usually been this 

CleanShot 2026-05-07 at 22 59 11@2x

### But sometimes I also invoke it via raycast and my 32" LG 6k goes to sleep almost immediately.

image

### So this is that for Omarchy.

image

PR #5585, opened 2026-05-05
Support tmux theme fragments in Omarchy themes

crmne:
## Summary

Adds optional tmux theme fragments loaded from the active Omarchy theme.

This lets themes ship a `tmux.conf` fragment alongside existing app-specific
theme files, while keeping the default tmux colors intact for themes that do not
provide one.

## Changes

- Sources `~/.config/omarchy/current/theme/tmux.conf` from the default tmux config when present.
- Reloads tmux when changing Omarchy themes.
- Uses `tmux ls` instead of `pgrep` so tmux is only reloaded when a server is reachable.

## Testing

- `git diff --check origin/dev..cp/tmux-theme-fragments`
- `bash -n bin/omarchy-restart-tmux`
- `bash -n bin/omarchy-theme-set`

### DIFF
--- a/bin/omarchy-restart-tmux
+++ b/bin/omarchy-restart-tmux
@@ -2,6 +2,9 @@
 
 # omarchy:summary=Restart tmux if running with the latest configuration
 
-if pgrep -x tmux; then
+if tmux ls >/dev/null 2>&1; then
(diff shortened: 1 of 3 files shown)


dhh:
Can you show me some before/after images of how you have put this to good use in a theme?

PR #5595, opened 2026-05-05
fix silent text capture failure when tesseract is missing

ilyaZar:
Per discussion #5594 it seems that the 3.7 text capture can fail when `tesseract` is not available.

Could happen, if a migration fails during omarchy update, or if the sb later just uninstalls it.

In any case it fails silently which is not so nice.

This PR fixes that.

One can quickly uninstall `tesseract` and reproduce the silent fail. Then get the couple of lines, and see that when `tesseract` is uninstalled, and the text capture is fired up, the popup emerges notifying the user that the text capture is unavailable.

### DIFF
--- a/bin/omarchy-capture-text-extraction
+++ b/bin/omarchy-capture-text-extraction
@@ -4,6 +4,14 @@
 # omarchy:group=capture
 # omarchy:examples=omarchy capture ocr
 
+if omarchy-cmd-missing tesseract || ! tesseract --list-langs 2>/dev/null | grep -qx "eng"; then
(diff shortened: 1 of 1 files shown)


dhh:
Appreciate this, but I don't want to litter the code base with defensive provisions like this. You could do this all over. The system is not designed for updates where migrations are not going to be run. If there's a problem where this doesn't happen in some cases, we should find those cases, and fix them.

ilyaZar:
true, thats the style to go here . iam new , thanks for clarification !

PR #5597, opened 2026-05-05
Add bass speaker fix for Lenovo Yoga Pro 7 14IAH10

aikazu:
## Problem
The Lenovo Yoga Pro 7 14IAH10 has dual amplifier speakers (AMP1 + AMP2) that require a specific pin quirk for the ALC287 codec. Without this fix, only one speaker works and bass output is missing.

## Solution
This PR adds hardware detection and audio configuration for the Yoga Pro 7 14IAH10:

- **Hardware detection script**: `bin/omarchy-hw-lenovo-yoga-pro7-bass`
- **Audio fix**: `install/config/hardware/lenovo/fix-yoga-pro7-bass-speakers.sh`

The fix creates `/etc/modprobe.d/lenovo-yoga-pro7-bass.conf` with:
```
options snd-sof-intel-hda-generic hda_model=alc287-yoga9-bass-spk-pin
```

## Testing
Tested on Lenovo Yoga Pro 7 14IAH10 with Intel Core Ultra 9 285H. Both speakers now work correctly with proper bass output.

## Reference
Based on the same fix used for Lenovo Yoga 9i 2022: https://wiki.archlinux.org/title/Lenovo_Yoga_9i_2022_(14AiPI7)#Audio

### DIFF
--- a/install/config/all.sh
+++ b/install/config/all.sh
@@ -63,6 +63,8 @@ run_logged $OMARCHY_INSTALL/config/hardware/apple/fix-spi-keyboard.sh
 run_logged $OMARCHY_INSTALL/config/hardware/apple/fix-suspend-nvme.sh
 run_logged $OMARCHY_INSTALL/config/hardware/apple/fix-t2.sh
 
+run_logged $OMARCHY_INSTALL/config/hardware/lenovo/fix-yoga-pro7-bass-speakers.sh
+
 run_logged $OMARCHY_INSTALL/config/hardware/fix-bcm43xx.sh
 run_logged $OMARCHY_INSTALL/config/hardware/fix-surface-keyboard.sh
 run_logged $OMARCHY_INSTALL/config/hardware/fix-yt6801-ethernet-adapter.sh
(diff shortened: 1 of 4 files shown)


install/config/hardware/lenovo/fix-yoga-pro7-bass-speakers.sh
```diff
@@ -0,0 +1,10 @@
+# Fix bass speakers on Lenovo Yoga Pro 7 14IAH10
+# The ALC287 codec needs a pin quirk to route audio to both AMP speakers.
+# Without this quirk, only one speaker works and bass output is missing.
+# Reference: https://wiki.archlinux.org/title/Lenovo_Yoga_9i_2022_(14AiPI7)
+
+if omarchy-hw-lenovo-yoga-pro7-bass; then
```
dhh: Let's just inline the hw-match call here. Since it's only used here.

install/config/all.sh
```diff
@@ -63,6 +63,8 @@ run_logged $OMARCHY_INSTALL/config/hardware/apple/fix-spi-keyboard.sh
 run_logged $OMARCHY_INSTALL/config/hardware/apple/fix-suspend-nvme.sh
 run_logged $OMARCHY_INSTALL/config/hardware/apple/fix-t2.sh
 
+run_logged $OMARCHY_INSTALL/config/hardware/lenovo/fix-yoga-pro7-bass-speakers.sh
```
dhh, line 66: We should also add a migration so existing machines get the fix.

PR #5607, opened 2026-05-06
Add GIF option to screenrecord via omarchy-transcode

bhood239:
Per the comment in #3699 , this swaps the inline ffmpeg path for a delegation to omarchy-transcode. Made use of the newly added video to gif transcoding added by @dhh in 10165f2

- Adds "Save as GIF (silent)" to the screenrecord menu(--as-gif flag)
- On stop: omarchy-transcode "$file" gif "${RESOLUTION:-1080p}". mp4 is replaced with the gif
- On failure: keeps the mp4

Tested locally with PATH=$PWD/bin:$PATH omarchy-menu screenrecord.

### DIFF
--- a/bin/omarchy-capture-screenrecording
+++ b/bin/omarchy-capture-screenrecording
@@ -32,6 +32,9 @@ WEBCAM="false"
 WEBCAM_DEVICE=""
 RESOLUTION=""
 STOP_RECORDING="false"
+AS_GIF="false"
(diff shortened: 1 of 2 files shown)


dhh:
Appreciate the suggestion, but I think we're better off keeping the steps separate. It's super easy after recording a screencast to convert to gif, if that's needed. I think it's a rare case, though, so not worth the additional option up front.

PR #5610, opened 2026-05-06
allow users to view logs of the installation

ilyaZar:
I sometimes do that manually

instead of scrolling back through the update terminal to see failures,   opening the update log in nvim lets users review what just happened before deciding whether to reboot

### DIFF
--- a/bin/omarchy-update-restart
+++ b/bin/omarchy-update-restart
@@ -4,6 +4,27 @@
 
 echo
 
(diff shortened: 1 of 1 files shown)


dhh:
Appreciate the suggestion, but don't like the idea of getting this into the regular flow. 97% of the time there's nothing you should have to review. It's a failure mode if you need to.

PR #5611, opened 2026-05-06
omarchy upgrade analysis

ilyaZar:
the idea is to play nicely with #5610 

so when omarchy updates, and the user choses to view the logs properly in nvim, its just a bit of typing :

fire up a terminal with `ctrl+/` , then type the command `omarchy update agent-fix` to "fix" erros/discuss with the agent a topic of concern

thats motivated by  #5594 and #5595  to catch errors earlier and fix

### DIFF
--- a/bin/omarchy-update-agent-fix
+++ b/bin/omarchy-update-agent-fix
@@ -1,0 +1,47 @@
+#!/bin/bash
+
+# omarchy:summary=Ask opencode to fix skipped update migrations and upgrade errors.
+# omarchy:aliases=omarchy update agent-fix
+
+set -e
+
+OMARCHY_PATH=${OMARCHY_PATH:-$HOME/.local/share/omarchy}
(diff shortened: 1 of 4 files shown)


ilyaZar:
Just adding a couple of thoughts here because they are related:

One might also think about refactoring the existing omarchy skill alltogether

- progressive disclosure is claimed to be a thing, by which one should make the main SKILL.md file small, and basically a  router to sub-documents depending on the task , see here https://agentskills.io/home
- the proposed `.md` may then become just a single sub-document to be routed to from the main omarchy skill document , when that gets refactored
- I am not sure if the skill gets installed when opencode or codex is installed; I couldnt find the traces of that; on my system I hook the configs from Dropbox, and install opencode from source so I cant tell ; my understanding is that claude uses a different install path for skills than the others

dhh:
Closing because this targets the old update-analysis command and agent-skill flow, both of which changed substantially for Quattro.

PR #5620, opened 2026-05-06
Asus Zenbook UX5406AA Display Backlight Fix

mateuszkowalczyk:
## Problem
ASUS Zenbook UX5406AA (Panther Lake) has broken brightness control (adjusting brightness does not affect actual brightness). This seems to be the same issue as the one on ExpertBook fixed in #5435 

## Solution
Extract backlight fix from ExpertBook B9406 solution into a separate script and run it for both ExpertBook and new Zenbook UX5406AA.

## Migration
Adds a migration that:
- Runs only for the models mentioned above.
- Removes the old `xe.enable_dpcd_backlight=1` entry from the ExpertBook-specific drop-in if the file exists.
- Applies the new shared ASUS Panther Lake backlight drop-in.
- Runs `limine-update`.
## Future Models
Other ASUS Panther Lake models may need the same fix, but this only enables it for confirmed ExpertBook B9406 and Zenbook UX5406AA hardware for now. Once we have evidence that other models are affected we can easily extend the list of included models.

## Testing
- I didn't perform the fresh install of Omarchy, but ran the fix script manually on my Zenbook while simulating the installation environment variables instead.
- I removed the fix, then ran the migration directly to test it.
- I force-ran the existing ExpertBook script from #5435 on my Zenbook by removing the if condition, then ran the migration to verify the upgrade path from the existing fix file.

## Code Review
- I double-checked it myself and asked GPT-5.5 on OpenCode for review in terms of correctness and consistency with Omarchy source code
- @mijuny - I'd appreciate it if you could test the migration on your ExpertBook to make sure it works on a real device. I'd also appreciate a quick review of my changes, since they build upon your fixes.


dhh:
cc @mijuny could you verify if this all still works fine on the expertbook?

mijuny:
Hi! 

I tested the migration on B9406 (PR HEAD `9acc40b2`, ran with `OMARCHY_PATH` pointed at the branch checkout to mirror `omarchy-migrate`'s invocation). 

Pre-migration drop-in had both flags. Post-migration: dpcd line correctly sed'd out of `asus-expertbook-b9406-display.conf` with panel-replay line preserved. New `asus-ptl-display-backlight.conf` has `xe.enable_dpcd_backlight=1`. 

Aggregated `/boot/limine.conf` shows both flags exactly once per entry. 

After reboot, `/proc/cmdline` carries both flags. Brightness works normally/smoothly across the range and no Panel Replay regression. `omarchy-hw-asus-zenbook-ux5406aa` correctly returns false on B9406.  Looks good.

PR #5621, opened 2026-05-06
Make power profile source defaults configurable

crmne:
## Motivation

The current default switches laptops to `performance` whenever AC power is connected. On my machine, that immediately drives the fans to 100%, so I prefer to stay on `balanced` even when plugged in.

The bigger issue is that this is currently managed through Omarchy's generated udev rule. Any manual change to that rule can be overwritten by updates or migrations, so keeping a different AC/battery policy requires repeatedly editing root-owned udev config by hand.

This patch keeps Omarchy's automatic AC/battery switching, but moves the policy into a user-facing configuration flow. Users can choose persistent AC and battery defaults from `Setup > Power Profile`, while the Waybar battery click remains a temporary one-off profile switch.

## Summary

- add persistent AC/battery power profile defaults in `~/.config/omarchy/powerprofiles.conf`
- extend `Setup > Power Profile` with separate persistent choices for AC and battery
- keep `omarchy-menu power` as the temporary profile picker used by the Waybar battery click
- route udev-triggered profile switching through the configured defaults
- remove the Hyprland autostart power profile setter so source switching stays in the udev path

This keeps the current defaults (`performance` on AC when available, `balanced` on battery), but lets users choose something else without editing `/etc/udev/rules.d/99-power-profile.rules`.

## Testing

- `bash -n bin/omarchy-powerprofiles-config bin/omarchy-powerprofiles-set bin/omarchy-menu migrations/1777938520.sh`
- `git diff --check dev..power-profile-defaults`

### DIFF
--- a/bin/omarchy-menu
+++ b/bin/omarchy-menu
@@ -287,12 +287,49 @@ show_setup_menu() {
 }
 
 show_setup_power_menu() {
+  local ac_profile
+  local battery_profile
+  local options
+
+  ac_profile=$(omarchy-powerprofiles-config get ac)
+  battery_profile=$(omarchy-powerprofiles-config get battery)
+  options="󱐋  Temporary Profile\n󰚥  On AC: $ac_profile\n󰁹  On Battery: $battery_profile"
+
+  case $(menu "Power Profile" "$options") in
+  *Temporary*) show_power_profile_menu show_setup_power_menu ;;
+  *"On AC"*) show_power_profile_default_menu ac "On AC" ;;
+  *"On Battery"*) show_power_profile_default_menu battery "On Battery" ;;
+  *) back_to show_setup_menu ;;
+  esac
+}
+
+show_power_profile_menu() {
+  local parent_menu="${1:-show_main_menu}"
+  local profile
+
   profile=$(menu "Power Profile" "$(omarchy-powerprofiles-list)" "" "$(powerprofilesctl get)")
 
   if [[ $profile == "CNCLD" || -z $profile ]]; then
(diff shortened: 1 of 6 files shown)


dhh:
Appreciate the suggestion, but I think this is too much code. You can change the udevs directly, and we should instead just document that. I think for most people the defaults are really good here.

crmne:
Fair on the code volume, I can cut it.

However, editing the udev rule directly doesn't actually hold up, for two reasons:

1. **Boot and wake don't go through udev.** `autostart.conf` runs `omarchy-powerprofiles-init` -> `omarchy-powerprofiles-set ac`, which hard-codes AC -> `performance`, battery -> `balanced`. So even with a udev override, Omarchy reapplies `performance` on every boot and wake-from-sleep.
2. **The rule isn't durable.** Omarchy regenerates `99-power-profile.rules` from `powerprofilesctl-rules.sh`, and several migrations re-source that generator. A hand edit gets clobbered on the next update.

So instead of the menu, I'd slim this to just a config file:

```
# ~/.config/omarchy/powerprofiles.conf
ac=performance
battery=balanced
```

`omarchy-powerprofiles-set` reads it and falls back to today's defaults if it's missing. Defaults stay identical for everyone. It just becomes an overridable, Omarchy-supported file instead of a root-owned one that updates overwrite.

What really bothers me: `performance` on AC makes [my Zenbook S16](https://x.com/dhh/status/1924842040904294514?s=20) fans genuinely loud: exactly the kind of thing that varies by hardware.

I pushed #6046. Hope you're open to it.

PR #5622, opened 2026-05-06
Use flat pointer acceleration by default

crmne:
## Summary

- set Hyprland's default `input:accel_profile` to `flat`
- replace libinput's adaptive pointer acceleration curve with a constant pointer response by default

This makes trackpad and mouse movement feel less laggy and less speed-dependent out of the box, while users can still override the setting in `~/.config/hypr/input.conf`.

## Testing

- `git diff --check dev..flat-pointer-acceleration`

### DIFF
--- a/config/hypr/input.conf
+++ b/config/hypr/input.conf
@@ -19,8 +19,8 @@ input {
   # Increase sensitivity for mouse/trackpad (default: 0)
   # sensitivity = 0.35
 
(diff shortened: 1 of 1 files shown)


dhh:
This feels a lot worse to me on my laptops, so going to skip. But it's exposed with a comment so anyone can tweak as they see fit.

PR #5627, opened 2026-05-06
Install a new default browser

dhh:
Select your own default browser. Move all the custom theming into this flow.


mjtappert:
@dhh would you consider adding Vivaldi? I could create a PR, if that helps. Thanks!

dinvlad:
@mjtappert I've created this PR, ptal if you'd like https://github.com/basecamp/omarchy/pull/5746

PR #5630, opened 2026-05-06
Fix helix theme's word selection colors

### LINKED ISSUE #5580: Helix themes highlighted row not differentiated visually from current selection
### System details

Omarchy 3.7.1

### What's wrong?

For the omarchy Helix themes the background colour for the current row where you have your cursor and the currently selected words is identical, so you have no visual feedback on the currently selected words.

To see the difference in feedback you can test with equivalent Helix themes where they exist. For example when on omarchy nord select some words in Helix, then change theme in Helix from omarchy to nord. Likewise for other omarchy themes that have Helix equivalents.

Note: Before this update, I had my own helix theme changer that mapped omarchy themes to helix themes where they existed, and defaulted to helix term16_dark if there was no mapping. I could have used term16_dark for all but found the dedicated themes in Helix had more polish when available.

lukehsiao:
Without this fix, if users enabled `editor.cursorline`, the word selection would not be visually distinct from the cursorline selection.

Closes: https://github.com/basecamp/omarchy/issues/5580

## Before fix
The word `cursorline` is being selected, but is not visually differentiated.

image

## After fix
image

## Looks reasonable across themes

image

image
image
image
image
image
image
image
image

matte-black matches color0 with the selection background, so this is more subtle, but the foreground still changes to make it visually distinct.
image
With the tweak in this PR:
image

image
image
image
Retro-82's color0 is too close to the background color, but the fix would be to tweak retro-82.
image
Tweaked in this PR:
image

image
image

image
image
image

### DIFF
--- a/default/themed/helix.toml.tpl
+++ b/default/themed/helix.toml.tpl
@@ -87,7 +87,7 @@
 "ui.virtual.jump-label" = { fg = "color1", modifiers = ["bold"] }
 "ui.virtual.whitespace" = "color8"
 
-"ui.selection" = { bg = "color0" }
+"ui.selection" = { bg = "selection_background", fg = "selection_foreground" }
 
 "ui.cursor" = { fg = "background", bg = "cursor" }
 "ui.cursor.primary" = { fg = "background", bg = "cursor" }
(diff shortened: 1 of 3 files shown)


lukehsiao:
@kcodedev can you give this a try?

kcodedev:
> @kcodedev can you give this a try?

Tested. The fix works on my end. Thanks.

PR #5633, opened 2026-05-06
Add live weather

dhh:
image


mtp13:
The new weather widget does not show temperature in the USA because wttr.in defaults to F in USA and C for the rest of the world.  This following change, adding "m&" before "format" in the curl request fixes the problem for me.

`weather=$(curl -fsS --max-time 4 "https://wttr.in?m&format=%l|%t|%w" 2>/dev/null | tr -d '\n')`

dhh:
Fixed in https://github.com/basecamp/omarchy/commit/0c6bd1929fd32ba9ef9eed6a4da484998e613e04

mtp13:
For your consideration.

I took the liberty to simplify the script.  The wind in the USA is showing in km and we typically use mph.  I removed the unit conversion and the script just lets wttr.in return in the local units now based on location.  

```
#!/bin/bash

# omarchy:summary=Returns a formatted weather status string with temperature and wind speed.

weather=$(curl -fsS --max-time 4 "https://wttr.in?format=%l|%t|%w" 2>/dev/null | tr -d '\n')

if [[ -z $weather ]]; then
  echo "Weather unavailable"
  exit 1
fi

IFS='|' read -r place temperature wind <<< "$weather"
place=${place%%,*}
place=${place^}
temperature=${temperature#+}
# wttr.in returns locale-appropriate units already (e.g. �F/mph in the US),
# so display the values as-is instead of forcing metric-to-imperial conversion.

echo "$(omarchy-weather-icon)    $place  �  Temp $temperature  �  Wind $wind"
```

PR #5639, opened 2026-05-07
Add time, date, and battery to the lock screen

drunkleen:
Shows a large clock and date centered on the hyprlock screen, and a battery icon with percentage in the bottom-left corner for laptop users (hidde automatically on devices without battery).

Adds omarchy-battery-status-hyprlock to fetch the battery label.


niraletter:
i think it just wastes battery if unplugged

drunkleen:
@niraletter sure, but when the PC is idle the display turns off!

niraletter:
Turning off the display doesn’t stop the process itself. If the config says “run this cmd every X seconds” the timer or loop will still keep running in the background even when the screen is off,  unless the program specifically pauses updates when the display is disabled/suspended

dhh:
Thanks for the suggestion, but I really like our ultra clean lock screen. So going to skip, but would be nice to document this for anyone who wants it!

Issue #5601, opened 2026-05-05
jetbrains products doesn support autofocus anymore


n4zguu1:
### System details

AMD R5 5600, RTX 3060TI

### What's wrong?

after it has been fixed , the mouse hover autofocus doesnt work on jetbrains products on the last update 3.7.1 , specifically rustrover

raphaelstary:
for me a workaround in the past has been turning wayland mode off (it's on per default for their 2026.X releases):
https://youtrack.jetbrains.com/articles/SUPPORT-A-1166/Linux-how-to-turn-on-off-native-Wayland-mode

until the latest webstorm release, wayland mode was kind of unusable for me, because of rendering issues, focus loss, windows spanning offscreen etc.
but now since ws 2026.1.1 it's better and in return xwayland mode became unbearable.

nazguul-g:
yes i tried this and didnt fix the issue , i think the problem is on the distro itself cuz the moment i updated to the latest version,  issue appeared instantly.

MedvedevOO:
Can confirm the issue on pycharm and Android studio, propably all of Jetbrains IDEs affected. And yes the problem was not present before the update as well. Also want to add that focus is switching very fast between IDE window and buttons near the window border(all of them)
And whats most interesting. After this focus loss s*show waybar freezes and does not update info on current workspace. 

https://github.com/user-attachments/assets/3ca380d1-9e88-458b-b80a-dda70f209cee

Here is the video with the issue(highlighted with my mouse)

My system:
ryzen 9950x
RTX 5080

MedvedevOO:
@nazguul-g found a fix for it.
We need to edit hyprland.conf 

So. 
vim ~/.config/hypr/hyprland.conf

add this to the end of file

windowrule {
    name = jetbrains-tooltip
    no_initial_focus = on
    match:class = ^(jetbrains-.*)$
    match:title = ^(win.*)$
    match:float = 1
}

What actually was happening with focus:
When hovering over IDE side panel buttons (Project, Run, Git, etc.), JetBrains creates floating popup windows with titles like win0,
  win1, etc. Without no_initial_focus, Hyprland gives these popups keyboard focus the moment they appear. 
This causes:
  1. The IDE loses focus on hover
  2. The popup closes immediately and focus cycles back
  3. This repeats rapidly for every pixel of mouse movement → a flood of activewindow IPC events
  4. Waybar receives the event flood and freezes

[...]

PR #5640, opened 2026-05-07
FIx for #5601: Fix JetBrains focus loss and Waybar freeze on side panel hover (restore missing no_initial_focus)

MedvedevOO:
PR https://github.com/basecamp/omarchy/pull/5239 partially reverted https://github.com/basecamp/omarchy/pull/5183 by restoring no_follow_mouse for JetBrains windows.
However, it omitted a second rule that was part of the original config: no_initial_focus for floating internal popup windows.

When hovering over IDE side panel buttons (Project, Run, Git, etc.), JetBrains creates floating popup windows with titles like win0, win1, etc.
Without no_initial_focus, Hyprland gives these popups keyboard focus the moment they appear.
This causes:

The IDE loses focus on hover
The popup closes immediately and focus cycles back
3.his repeats rapidly for every pixel of mouse movement → a flood of activewindow IPC events
Waybar receives the event flood and freezes
The original config (before https://github.com/basecamp/omarchy/pull/5183) had both rules. https://github.com/basecamp/omarchy/pull/5239 only restored one.

Confirmed on: JetBrains IDEs with native Wayland support (2026.1+), Hyprland 0.54.x

Fix
Add missing rule to File: default/hypr/apps/jetbrains.conf

windowrule {
name = jetbrains-tooltip
no_initial_focus = on +
match:class = ^(jetbrains-.)$
match:title = ^(win.)$
match:float = 1
}

### DIFF
--- a/default/hypr/apps/jetbrains.conf
+++ b/default/hypr/apps/jetbrains.conf
@@ -4,3 +4,12 @@ windowrule {
     no_follow_mouse = on
     match:class = ^(jetbrains-.*)$
 }
+
+# Prevent floating popup windows (tooltips, panel previews) from stealing focus on hover (see https://github.com/basecamp/omarchy/issues/5601)
+windowrule {
+    name = jetbrains-tooltip
+    no_initial_focus = on
+    match:class = ^(jetbrains-.*)$
+    match:title = ^(win.*)$
+    match:float = 1
+}


NicolasDorier:
@MedvedevOO can you put a video? I have Jetbrains Rider and don't seem to be able to reproduce.

I just want to make sure I can repro all the quirks so I don't fat finger break things next time.

MedvedevOO:
@NicolasDorier Sure thing. Here

https://github.com/user-attachments/assets/56b718bc-0c94-490b-8f76-d9e2ebae7899

NicolasDorier:
I can't reproduce (but I am using rider), did you set `-Dawt.toolkit.name=WLToolkit` explicitely?

Jetbrains in 2026.1 automatically set WLToolKit, but in 2026.1.1 they reverted this decision and it needs to be set explicitly.

MedvedevOO:
No i didn't. I checked my config rn and it is not set.
I'm using Android studio 2025.3.4 patch one and PyCharm 2025.3.1
Upd:
Well i just tried to set WLtoolkit manually and disable windowrule and it looks like it fixes the problem also.

But you have to manually set this option for each IDE which is not convenient at all tbh.
For example I use pycharm,Idea and android studio:D

NicolasDorier:
I think you will find a lot of other quirks if you don't set it... Because then it's going through XWayland

MedvedevOO:
You are definately right. But I have a question. Is windowrule conflicts with vmoptions? In my understanding it does not. so maybe keep it until Jetbrains will return this option as a default coz without it it is hella painful for a new user to figure out.

NicolasDorier:
I haven't noticed any regression caused by this rule. However, if we don't assume WLToolKit is being used, I think we'll keep seeing PRs that try to work around XWayland quirks, some of which are not really fixable.

If things are broken instead, people looking for a solution are more likely to find a thread like this one and fix the issue properly by adding the vmoptions. I also think this is only a temporary situation, since JetBrains will likely enable WLToolKit by default again soon.

Personally, I wouldn't merge this PR, but I'm not strictly opposed to it either. That said, I'm not the maintainer of this project.

axelfontaine:
Any Wayland rules should really be there to suit 2026.1 and newer. Older versions like 2025.3 should be considered outdated.

NicolasDorier:
@axelfontaine the problem is that `2026.1` shipped with WLToolKit by default, but they reverted it in `2026.1.1`. So right now, by default, Jetbrains has quirks out of the box on latest version. (that you can fix with `-Dawt.toolkit.name=WLToolkit` in the vm options)

dhh:
@NicolasDorier Would be nice if we could do a Install > Editor > JetBrains that would setup the toolkit and have the right vm options. If we can do it out of the box.

NicolasDorier:
Jetbrains isn't only one editor, you mean making a mega list of all jetbrains editors?

dhh:
I was hoping we could just set this in the toolbox installer they have, and that would apply to all of them?

dhh:
Alternatively, we could consider doing Install > Editor > JetBrains > RubyMine/PythonWhatever/etc. If the config setup is warranted there.

NicolasDorier:
Seems easy enough to vibe code, I'm in, will do PR tomorrow boss :saluting_face:

NicolasDorier:
It's ready to go https://github.com/basecamp/omarchy/pull/5654

Tsurgcom:
> @NicolasDorier Would be nice if we could do a Install > Editor > JetBrains that would setup the toolkit and have the right vm options. If we can do it out of the box.

So does that mean we can get an option to install android studio from the launcher? that would be amazing!

PR #5641, opened 2026-05-07
feat: autostart foot server and use footclient

felixzsh:
this will start each foot instance a little bit faster and use less memory

### DIFF
--- a/default/foot/foot.desktop
+++ b/default/foot/foot.desktop
@@ -1,7 +1,7 @@
 [Desktop Entry]
 Type=Application
-TryExec=foot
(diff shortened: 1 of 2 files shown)


abenz1267:
From my personal experience: don't. Foot is "fast" enough to start without server/client.

1 terminal crashes = all close. This is horrible and imo not worth the tiny tiny tiny startup boost.

dhh:
Foot is not installed by default, so we can't have it as a default boot thing. Probably just have to leave this as an exercise for those who really want to optimize the last bit.

PR #5648, opened 2026-05-07
Avoid Chromium VAAPI flags on NVIDIA

hjanuschka:
## Summary
- add a helper that strips Chromium VAAPI decode/encode flags on NVIDIA systems
- run it during browser install and Chromium refresh so the flags are not reintroduced
- add a migration to clean up existing Chromium-family browser configs on NVIDIA

### DIFF
--- a/bin/omarchy-config-chromium-flags
+++ b/bin/omarchy-config-chromium-flags
@@ -1,0 +1,55 @@
+#!/bin/bash
+
+# omarchy:summary=Strip unsupported VAAPI flags from Chromium-family browser configs on NVIDIA
+# omarchy:hidden=true
+
+has_nvidia() {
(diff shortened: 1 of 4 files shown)


dhh:
It seems like this is working fine on some nvidia systems. Would be nice to pin down exactly where it's trouble.

hjanuschka:
you are absolutly right, sadly, i have no chance to reproduce it locally, maybe we could build an amazon wishlist with one of the systems that reproduce, so i can buy one to play around with? still super afraid that its not chromium, and more like a nvidia on linux thing (not saying that we could not fix that; up to the point where driver are closed source)!

dhh:
I added up yanking them. They were causing crashes on AMD 395 as well! All good. Really don't need them. Clearly behind flags for a reason.

PR #5664, opened 2026-05-08
Warn on low amdgpu vram assignment to workaround hard freezes

dhh:
https://github.com/basecamp/omarchy/issues/5478


dhh:
Seems like this isn't actually the fix.

PR #5668, opened 2026-05-08
Use emacsclient when setting default editor to Emacs

spdawson:
We're already running the Emacs server; so it makes sense to use it.

Follow-up for https://github.com/basecamp/omarchy/pull/5651


dhh:
Bit risky in case the server is down for some reason. Don't think it really needs the optimization. It's just when $EDITOR needs to be invoked. Appreciate the suggestion, though!

PR #5674, opened 2026-05-08
Feature/firefox webapp support

molhamfetnah:


dhh:
Web apps with the full browser chrome does not make sense, so that's why we haven't done them for firefox. If they ship a webapp mode like Chrome in the future, we can revisit then.

PR #5698, opened 2026-05-09
Sync with upstream

apoindevster:


apoindevster:
Accidental pr to upstream instead of my own fork

PR #5702, opened 2026-05-09
Fix monitor scaling shortcut conflict on alternative keyboard layouts

### LINKED ISSUE #5665: Monitor scaling shortcut overlap using alternative keyboards
### System details

Omarchy 3.7.1, bépo keyboard (French ergo layout)

### What's wrong?

After the recent update (v3.7.0 and v3.7.1), the usual shortcuts SUPER + ALT + F (toggle full size window) and SUPER + F (toggle fullscreen) seem to overlap with new shortcuts made for cycling through monitor scaling.

My F key is located right next to Right Shift key, not sure what keycode it's supposed to be (from wev command I get key: 61).
Alternative keyboard layout for reference: https://bepo.fr/wiki/Accueil

I can just unbind the right key/keys, but not sure which key it is, and if it would affect other bindings at other keyboard locations.

Thanks for the help!

Ninso112:
## Summary

- Change monitor scaling shortcuts from `code:61` (/) to `code:60` (.) in `default/hypr/bindings/tiling-v2.conf`
- Add migration to update existing user configs

## Why

Issue #5665 reports that on the bépo (French ergonomic) keyboard layout, the monitor scaling shortcuts (`SUPER + code:61` and `SUPER + ALT + code:61`) conflict with `SUPER + F` (fullscreen) and `SUPER + ALT + F` (full width). This is because `code:61` maps to the F key position on bépo, whereas it maps to `/` on QWERTY.

Changing to `code:60` (the `.` key on QWERTY, `,` on bépo) avoids letter key conflicts on alternative layouts while keeping the shortcut in the same physical area near the right shift key.

Fixes #5665

### DIFF
--- a/default/hypr/bindings/tiling-v2.conf
+++ b/default/hypr/bindings/tiling-v2.conf
@@ -128,6 +128,6 @@ bindd = SUPER ALT, code:12, Switch to group window 3, changegroupactive, 3
 bindd = SUPER ALT, code:13, Switch to group window 4, changegroupactive, 4
 bindd = SUPER ALT, code:14, Switch to group window 5, changegroupactive, 5
 
-# Cycle monitor scaling with SUPER + /
-bindd = SUPER, code:61, Cycle monitor scaling, exec, omarchy-hyprland-monitor-scaling-cycle
-bindd = SUPER ALT, code:61, Cycle monitor scaling backwards, exec, omarchy-hyprland-monitor-scaling-cycle --reverse
+# Cycle monitor scaling with SUPER + .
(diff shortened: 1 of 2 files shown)


MrKrar:
Key code 60 on wev returns H on bépo, not ",". But afaik there's no shortcut under Super + H so it should be alright.

Not too familiar with this kind of stuff, but wouldn't mixing codes and character names ("F" for example for fullscreen) lead to other conflicts at some point or another?

But in any case, I feel like I'm one of a kind case, so if there are conflicts I can just unbind the "code:xx" out while keeping the letter's binding, right?

dhh:
Can't just change the existing keybind on people like this. Need to keep the slash. If you find another way to do that, please do open a PR.

PR #5703, opened 2026-05-09
Fix power profile udev rule race condition when AC connected at boot

### LINKED ISSUE #5688: Script fails if AC power is connected before boot, but works perfectly on manual run/post-boot hotplug
### System details

XiaoXinPro 14 AHP9, Ryzen 7 8745H, Radeon 780m, Omarchy 3.7.1

### What's wrong?

**Description**
When the AC power is connected before the laptop is powered on, the associated script fails to execute properly during the boot process.

However, if I start the laptop on battery power and then plug or unplug the AC adapter, the script detects the state change and works beautifully. Running the script manually after the system has fully booted also works without any issues.

**Steps to Reproduce
Scenario A (Fails):**
1. Ensure the laptop is completely powered off.
2. Plug in the AC power adapter.
3. Power on the laptop and boot the system.
Result: The script fails to run or apply its settings.

**Scenario B (Works):**
1. Ensure the laptop is completely powered off.
2. Disconnect the AC power adapter.
3. Power on the laptop and boot the system.
4. Plug in the AC power adapter once logged in.
Result: The script triggers successfully and functions as intended.

**Logs:**

Image
Image

Ninso112:
## Summary

- Add `Requires=power-profiles-daemon.service` to the `systemd-run` invocation in the power profile udev rule
- Add migration to fix existing user udev rules

## Why

Issue #5688 reports that the power profile script fails when AC power is connected before boot, but works fine on manual run or post-boot hotplug.

The existing udev rule uses `--property=After=power-profiles-daemon.service` which only sets ordering between units that are already queued. It doesn't create a dependency, so when AC is connected at boot the transient `omarchy-power-profile` unit can execute before `power-profiles-daemon.service` has been started.

Adding `Requires=power-profiles-daemon.service` ensures the daemon is pulled in and fully started before `omarchy-powerprofiles-set` runs, fixing the race condition.

Fixes #5688

### DIFF
--- a/install/config/powerprofilesctl-rules.sh
+++ b/install/config/powerprofilesctl-rules.sh
@@ -1,7 +1,7 @@
 if omarchy-battery-present; then
   cat <<EOF | sudo tee "/etc/udev/rules.d/99-power-profile.rules"
-SUBSYSTEM=="power_supply", ATTR{type}=="Mains", RUN+="/usr/bin/systemd-run --no-block --collect --unit=omarchy-power-profile --property=After=power-profiles-daemon.service $HOME/.local/share/omarchy/bin/omarchy-powerprofiles-set"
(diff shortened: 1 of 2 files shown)


gabrielvincent:
Is this the reason why my selected power profile settings keeps being ignored by the system? It automatically goes back to _performance_.

PR #5709, opened 2026-05-09
Restrict first-run systemctl sudo access

### LINKED ISSUE #5708: First-run sudoers allows unrestricted systemctl
### System details

Omarchy dev branch at 3eca9573; first-run sudoers; hardware independent

### What's wrong?

Omarchy creates `/etc/sudoers.d/first-run` during install so first-run scripts can finish setup after the user logs in. That sudoers file currently grants the installing user passwordless access to `/usr/bin/systemctl` with no argument restrictions:

```bash
$USER ALL=(ALL) NOPASSWD: /usr/bin/systemctl
```

First-run only needs one root `systemctl` call:

```bash
sudo systemctl enable ufw
```

Because sudoers treats a command with no argument list as allowing any arguments, any process running as the user before first-run cleanup can invoke arbitrary root `systemctl` actions without a password. That widens a temporary first-run convenience rule into a general root service-control path.

Steps to recreate:

1. Run the Omarchy install path through `install/preflight/first-run-mode.sh`.
2. Inspect `/etc/sudoers.d/first-run` or run `sudo -l` for the installing user.
3. Notice the unrestricted `NOPASSWD: /usr/bin/systemctl` entry.
4. Compare it to the only root systemctl call in first-run: `sudo systemctl enable ufw` in `install/first-run/firewall.sh`.

Expected behavior:

The temporary first-run sudoers file should permit only the exact root commands first-run needs, such as `systemctl enable ufw`, and should keep cleanup commands explicit.

Actual behavior:

The temporary sudoers file permits passwordless root `systemctl` with arbitrary arguments until first-run removes `/etc/sudoers.d/first-run`.

Suggested fix:

Replace the bare `systemctl` sudoers entry with a `Cmnd_Alias` for `/usr/bin/systemctl enable ufw`, and update reboot sudoers cleanup so it uses an explicitly permitted cleanup command rather than an unpermitted `sudo test` precheck.

afurm:
## Summary

- replace the first-run passwordless `systemctl` sudoers rule with an exact `systemctl enable ufw` alias
- allow reboot sudoers cleanup through an explicit `rm -f` alias instead of relying on an unpermitted `sudo test` precheck
- add a migration for active installs that still have the broad first-run sudoers file
- add a static regression test for the first-run sudoers contract

Fixes #5708

## Validation

- `bash -n install/preflight/first-run-mode.sh install/first-run/cleanup-reboot-sudoers.sh migrations/1778347762.sh test/first-run-sudoers-test.sh`
- `bash test/first-run-sudoers-test.sh`
- `visudo -c -f` against the generated first-run sudoers body
- `git diff --check`

Note: `bash test/omarchy-cli-test.sh` could not run on this macOS checkout because only Bash 3.2 is installed locally, and `bin/omarchy` requires Bash 5 associative arrays.

### DIFF
--- a/install/first-run/cleanup-reboot-sudoers.sh
+++ b/install/first-run/cleanup-reboot-sudoers.sh
@@ -1,3 +1,1 @@
-if sudo test -f /etc/sudoers.d/99-omarchy-installer-reboot; then
-  sudo rm -f /etc/sudoers.d/99-omarchy-installer-reboot
-fi
+sudo rm -f /etc/sudoers.d/99-omarchy-installer-reboot
--- a/install/preflight/first-run-mode.sh
+++ b/install/preflight/first-run-mode.sh
@@ -5,12 +5,15 @@ touch ~/.local/state/omarchy/first-run.mode
 # Setup sudo-less access for first-run
 sudo tee /etc/sudoers.d/first-run >/dev/null <<EOF
 Cmnd_Alias FIRST_RUN_CLEANUP = /bin/rm -f /etc/sudoers.d/first-run
+Cmnd_Alias INSTALLER_REBOOT_CLEANUP = /bin/rm -f /etc/sudoers.d/99-omarchy-installer-reboot
 Cmnd_Alias SYMLINK_RESOLVED = /usr/bin/ln -sf /run/systemd/resolve/stub-resolv.conf /etc/resolv.conf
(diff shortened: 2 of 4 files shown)


dhh:
Closing because this targets the retired first-run installer and sudoers mechanism. Quattro installation orchestration is now owned by the ISO.

PR #5723, opened 2026-05-10
Hyprland lua conversion

ryanrhughes:


bin/omarchy-hyprland-window-close-all
```diff
@@ -4,7 +4,9 @@
 
 hyprctl clients -j | \
   jq -r ".[].address" | \
-  xargs -I{} hyprctl dispatch closewindow address:{}
+  while read -r addr; do
+    hyprctl dispatch "hl.dsp.window.close(\"address:$addr\")" >/dev/null 2>&1 || hyprctl dispatch closewindow "address:$addr"
```
wizarki972, line 8: @ryanrhughes, I noticed the dispatch is only closing the focused window. I think the line should look like this - `hyprctl dispatch "hl.dsp.window.close({ window=\"address:$addr\" })" >/dev/null 2>&1 || hyprctl dispatch closewindow "address:$addr"`

PR #5728, opened 2026-05-10
add fallback for pasted file path in ASCII art generation

### LINKED ISSUE #5727: ASCII art generation for screensaver not working as expected
### System details

AMD Ryzen 5 7600X, Radeon RX 9070 XT, Omarchy 3.8

### What's wrong?

when I tried the new Set from Image option from screensaver in the omarchy menu today and pasted in the path to a png, nothing happened. curiously, running omarchy-transcode-ascii [same exact path to png] . does work tho. For now I just pasted the output of that to the screensaver.txt manually but would be nice to see a proper fix for this as it really is beautiful!

bl4zee1g:
I added a fallback option when omarchy-menu-file returns empty (which would be the case if you paste a file path), it falls through to omarchy-menu-input to capture the pasted path.
I also added a transcode condition that checks -f $image so a nonexistent path fails explicitly rather than being passed silently.
Closes #5727

### DIFF
--- a/bin/omarchy-branding-about
+++ b/bin/omarchy-branding-about
@@ -14,7 +14,7 @@ image)
   if [[ -z $image ]]; then
     image=$(omarchy-menu-input "Paste image path")
   fi
-  if [[ -n $image && -f $image ]] && omarchy-transcode-ascii "$image" ~/.config/omarchy/branding/about.txt --width 54 --height 26 --mode block; then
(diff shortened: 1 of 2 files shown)


dhh:
Hmm, appreciate the suggestion, but I find this a bit odd. Don't think most people would think to have the fully qualified path on the clipboard. But I've instead made it clearer where the picker is picking from, so you can copy the image you need there.

PR #5730, opened 2026-05-10
Control Waybar position

dhh:
image

Via:

image


axelfontaine:
Please use the opportunity to integrate something similar to my proposal in #5061 , a tiny but important usability improvement!

PR #5734, opened 2026-05-10
Change weather location for the waybar

vander00:
With this PR you can input the city for which you want to see the weather.

## The problem
I always use VPN, so with the current configuration I see the weather for any city but not mine. Now I can select my city, so I will see the correct weather.

## How it works
1. Right-click on the weather-widget
2. Enter the city, then the city saves to a file in $HOME/.config/omarchy/current/weather_city.name
3. The weather scripts reads the name and curls the weather status for the correct city.


vander00:
No need because of omarchy 4.

PR #5736, opened 2026-05-10
Fix internal display toggle not re-applied after resume

### LINKED ISSUE #5481: Internal display toggle not re-applied after suspend/resume
### System details

FW13 - AMD 7840 - Omarchy 3.6.0

## Summary

`omarchy-hyprland-monitor-watch` listens only for `monitorremoved` events and recovers internal display when externals disappear. There's no equivalent handler for resume-from-suspend that re-applies the `internal-monitor-disable` toggle, so eDP-1 comes back online on wake even when:
- Lid is closed
- External monitors are still connected  
- The toggle file still exists in `~/.local/state/omarchy/toggles/`

## Root cause hypothesis

The toggle is implemented as a sourced Hyprland config snippet, but Hyprland's post-resume monitor reconfiguration doesn't re-source it. The watcher script catches `monitorremoved` but not the implicit "monitor reconfiguration on resume" scenario.

## Suggested fix directions

1. Extend `omarchy-hyprland-monitor-watch` to also handle `monitoradded` events with toggle-respecting logic
2. Add a systemd-sleep hook (`/etc/systemd/system-sleep/`) that re-applies the toggle on resume
3. Have the post-resume path check lid state via `/proc/acpi/button/lid/*/state`

Happy to submit a PR if there's a preferred direction.

mat-ran:
## Problem
When you disable the internal laptop display while an external monitor is connected, then trigger display sleep (screen off) and wake the laptop, the internal display incorrectly turns back on - even though nothing changed with the external monitor connection.
The expected behavior: the internal display should stay disabled after wake, as long as the external monitor remains connected. This works correctly with `systemctl suspend`, but fails with display sleep.

## Root Cause
When the screen turns off (display sleep), the external monitor loses signal due to no input and Hyprland sends a `monitorremoved` event. Shortly after, when the laptop wakes, the external monitor takes a moment to be detected again. The monitor watcher would immediately try to "recover" the internal display without waiting for the external monitor to reappear on wake, causing a race condition where the internal display gets turned back on before the external monitor is detected.

## Solution
1. **`recover()` now waits up to 3 seconds** - before re-enabling the internal display, it polls for external monitors to give the system time to detect them on wake
2. **Added `reapply()` function** - to properly refresh the toggle state when monitors are reconnected
3. **Added `monitoradded` event handling** - ensures toggle state is maintained when monitors are added

Probably fixes #5481

### DIFF
--- a/bin/omarchy-hyprland-monitor-internal
+++ b/bin/omarchy-hyprland-monitor-internal
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 # omarchy:summary=Enable, disable, toggle, or recover the internal laptop display
-# omarchy:args=<on|off|toggle|recover>
+# omarchy:args=<on|off|toggle|recover|reapply>
 
 TOGGLE="internal-monitor-disable"
 TOGGLE_FLAG="$HOME/.local/state/omarchy/toggles/hypr/$TOGGLE.conf"
@@ -31,9 +31,18 @@ off() {
 }
 
 recover() {
-  if ! omarchy-hw-external-monitors; then
-    omarchy-hyprland-toggle $TOGGLE off
+  if ! omarchy-hyprland-toggle-enabled "$TOGGLE"; then
+    return
   fi
+
+  for i in {1..6}; do
+    if omarchy-hw-external-monitors; then
+      return
+    fi
+    sleep 0.5
+  done
+
+  omarchy-hyprland-toggle "$TOGGLE"
 }
 
 toggle() {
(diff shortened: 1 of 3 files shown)


dhh:
I tried a few different ways with this and was able to end up with the internal display turned off when I unplugged the external. I think we need some more testing to make sure this is fully solid.

mat-ran:
I'll check it out. In what situations has the built-in monitor failed to turn on?

mat-ran:
Quattro solved the problem.

PR #5740, opened 2026-05-10
feat: add Firefox and Zen browser support for WebApps

molhamfetnah:
Resolves https://github.com/basecamp/omarchy/discussions/2468

Adds support for Firefox and Zen Browser in WebApps launcher with proper window rules.

### DIFF
--- a/bin/omarchy-launch-webapp
+++ b/bin/omarchy-launch-webapp
@@ -6,8 +6,17 @@
 browser=$(xdg-settings get default-web-browser)
 
 case $browser in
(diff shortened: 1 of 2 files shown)


Mrid22:
> Resolves #2468
> 
> Adds support for Firefox and Zen Browser in WebApps launcher with proper window rules.

This is amazing! But why are the jetbrains changes here? was that meant to be on a seperate branch? I'm pretty sure you can just drop the commit using lazygit though.

dhh:
The point of the web apps is to have them appear without browser chrome. Otherwise it's just a bookmark. Firefox-class browsers do not support the chrome-less display, so we can't use them for this purpose.

PR #5745, opened 2026-05-10
Make weather widgets location- and units-aware

twagner12:
The weather widgets currently auto-detect both location and units from geo-IP, which fails on VPNs and on US locations that prefer metric (or vice versa). This adds:

**Location**
- `omarchy weather location` / `set-location` / `clear-location`
- walker picker via `omarchy menu weather`

**Units**
- `omarchy weather units` / `set-units (metric|imperial)` / `clear-units`
- walker picker reachable from the Setup → Weather menu

The Weather setup submenu now also dispatches via `omarchy-menu weather`, so a waybar right-click can be bound straight to it.

When no value is configured, behavior is identical to today — wttr.in auto-detects from the request.

### DIFF
--- a/bin/omarchy-menu
+++ b/bin/omarchy-menu
@@ -358,8 +358,9 @@ show_setup_menu() {
 }
 
 show_setup_weather_menu() {
-  case $(menu "Weather" "󰖙  Location") in
+  case $(menu "Weather" "󰖙  Location\n  Units") in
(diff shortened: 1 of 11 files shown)


dhh:
Closing because Quattro weather now supports saved locations and automatic unit selection natively. The old weather menu and commands were replaced.

PR #5748, opened 2026-05-11
Fix Hyprland 0.54 window rule compatibility

joshuaaeverett:
## Summary

Hyprland 0.45+ deprecated the old `windowrule`/`windowrulev2` inline syntax in favour of a new block syntax, and 0.54 fully breaks the old format — causing a wall of config errors on every startup.

This PR converts all window rules across the config to the new block format and renames removed/renamed options.

- Convert all `windowrule = rule, filter` and `windowrulev2` directives to `windowrule { }` block syntax across `windows.conf` and all `apps/*.conf` files
- Remove `misc:new_window_takes_over_fullscreen` (option removed from Hyprland)
- Remove `noscreenshare` rules from `1password.conf` and `bitwarden.conf` (removed from Hyprland with no replacement)
- Rename deprecated property names to their new equivalents:
  - `scrolltouchpad` → `scroll_touchpad`
  - `nofocus` → `no_focus`
  - `nofollowmouse` → `no_follow_mouse`
  - `noinitialfocus` → `no_initial_focus`
  - `stayfocused` → `stay_focused`
  - `noanim` → `no_anim`
  - `keepaspectratio` → `keep_aspect_ratio`
  - `idleinhibit` → `idle_inhibit`
  - `noborder` → `border_size = 0`
  - `suppressevent` → `suppress_event`
- Convert `layerrule` directives (walker, hyprshot) to block syntax

## Test plan

- [ ] Verify `hyprctl configerrors` returns no errors after reload
- [ ] Confirm window rules work as expected (floating, opacity, tags, etc.)

### DIFF
--- a/default/hypr/apps/1password.conf
+++ b/default/hypr/apps/1password.conf
@@ -1,2 +1,8 @@
-windowrule = noscreenshare, class:^(1[p|P]assword)$
-windowrule = tag +floating-window, class:^(1[p|P]assword)$
+# noscreenshare was removed in Hyprland 0.45+; no replacement available
+
+windowrule {
+    name = 1password-floating
+    match:class = ^(1[pP]assword)$
+
+    tag = +floating-window
+}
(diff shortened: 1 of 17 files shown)


dhh:
Thanks for the suggestion, but we are moving to the lua config in Hyprland 0.55 for the next version.

PR #5756, opened 2026-05-11
Add dispatch shortcuts to help menu

Malanius:
Adds the possibility to invoke shortcuts directly from the help menu.

I've recently broken a bone in one of my arms, rendering it unusable with half of the shortcuts as well. Then I've remembered there is an easily reachable place where all shortcuts are listed and are searchable. Would be good if it could dispatch them as well, similarly to how Raycast does.

I've vibed this out of necessity, but this might be useful for other people as well.


dhh:
Omg. Wanted this for so long!

PR #5763, opened 2026-05-11
Added notification for hyprpicker

Mrid22:
Just a small change, now when you pick a color through the capture menu, it shows you a little notification with the hex code + telling you. that it's been copied (I didn't know it did that and thought hyprpicker had crashed after I picked a color, also this makes it a bit more consistent with the other capture actions).


dhh:
Closing because this targets the removed color-capture command and old menu. Picker feedback can be re-proposed against the current capture binding and native notification service.

Issue #5758, opened 2026-05-11
Hyprland 0.55.0 config errors in default/hypr/looknfeel.conf (col.border_locked = -1, dwindle:pseudotile removed)


Ehsanul-deepchainlabs:
### System details

Intel i7-13700H, NVIDIA RTX 4060, Omarchy 3.8.0 (master @ b2d95ee2), Hyprland 0.55.0

### What's wrong?

<img width="2515" height="251" alt="Image" src="https://github.com/user-attachments/assets/400cdfe6-d7dc-4b53-8602-54aaa25f41f7" />

<img width="2560" height="1600" alt="Image" src="https://github.com/user-attachments/assets/717dc633-3c7b-45d9-aef4-97debe15e6f4" />

  On a fresh boot, Hyprland reports three config errors from the stock omarchy default `looknfeel.conf`:

  Config error in file /home//.local/share/omarchy/default/hypr/looknfeel.conf at line 53: Error parsing gradient -1: failed to parse -1 as a color
  Config error in file /home//.local/share/omarchy/default/hypr/looknfeel.conf at line 54: Error parsing gradient -1: failed to parse -1 as a color
  Config error in file /home//.local/share/omarchy/default/hypr/looknfeel.conf at line 111: config option dwindle:pseudotile does not exist.

  ### Cause

  Hyprland 0.55.0 dropped support for two things the omarchy default config still uses:

  1. **Lines 53–54** — `col.border_locked_active = -1` / `col.border_locked_inactive = -1`. The `-1` "inherit from parent color" shorthand is no longer accepted; the parser expects a real color/gradient. Confirmed via
  `hyprctl getoption group:col.border_locked_active` → `custom type: 0 0deg`.
  2. **Line 111** — `dwindle:pseudotile = true`. The option was removed in Hyprland 0.55.0. Confirmed via `hyprctl getoption dwindle:pseudotile` → `no such option`.

  ### Steps to reproduce

  1. Install Omarchy 3.8.0 (or update an existing install to master @ b2d95ee2).
  2. Be on Hyprland 0.55.0 (current Arch `hyprland` package).
  3. Reboot / re-login. Errors appear in the Hyprland error overlay and `hyprctl configerrors`.

  ### Workaround

  Errors are non-fatal — Hyprland continues to run, those three lines are just ignored. Users can override `col.border_locked_active/inactive` in `~/.config/hypr/looknfeel.conf` with explicit colors, but there's no way
[...]

Issue #5752, opened 2026-05-11
Hyprland(v0.55) config errors on startup — gradient color -1 parse failure & pseudotile option removed


ThePsyberSleuth:
### System details

Dell Inspiron 7375, AMD Ryzen 5 2500U, AMD Radeon Vega 8 (Integrated), Omarchy 3.8.0, Hyprland 0.55.0, Linux 7.0.5-arch1-1

### What's wrong?

On startup, Hyprland throws multiple config errors from ~/.local/share/omarchy/default/hypr/looknfeel.conf:

Config error at line 53: Error parsing gradient -1: failed to parse -1 as a color
Config error at line 54: Error parsing gradient -1: failed to parse -1 as a color
Config error at line 111: config option <dwindle:pseudotile> does not exist

Steps to reproduce:

1. Install Omarchy fresh (or update to latest) dev channel.
2. Launch Hyprland
3. Errors appear immediately in the startup error overlay

phplemos:
Same error here, when reboot after update to latest version.
**PrintScreen:**
<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/b524ec0a-863c-4bb2-a506-87bb920f223c" />

Device Settings:

<img width="878" height="601" alt="Image" src="https://github.com/user-attachments/assets/eb61adc0-4d4f-425e-a9bb-40c74e2db81a" />

ashvwinn:
Hyprland introduced some breaking changes with the 0.55.0. The configuration will soon be changed to the lua version which will fix these errors. For now, removing these lines from the config should work fine.

marcinczenko:
I did this:

```bash
cp ~/.local/share/omarchy/default/hypr/looknfeel.conf \
   ~/.local/share/omarchy/default/hypr/looknfeel.conf.bak

sed -i \
  -e 's/^\([[:space:]]*col\.border_locked_active[[:space:]]*=.*\)$/# \1/' \
  -e 's/^\([[:space:]]*col\.border_locked_inactive[[:space:]]*=.*\)$/# \1/' \
  -e 's/^\([[:space:]]*pseudotile[[:space:]]*=.*\)$/# \1/' \
  ~/.local/share/omarchy/default/hypr/looknfeel.conf

diff -u \
  ~/.local/share/omarchy/default/hypr/looknfeel.conf.bak \
  ~/.local/share/omarchy/default/hypr/looknfeel.conf
❯ diff -u \
[...]

PR #5767, opened 2026-05-11
fix: remove deprecated Hyprland options from looknfeel.conf

dev-gios:
## Problem

Two config options in `default/hypr/looknfeel.conf` cause errors on modern Hyprland:

```
Config error: Error parsing gradient -1: failed to parse -1 as a color
Config error: config option <dwindle:pseudotile> does not exist.
```

## Changes

- **`dwindle.pseudotile`** — removed. This option no longer exists in Hyprland and causes a config error on load.
- **`group.col.border_locked_active/inactive = -1`** — `-1` was previously accepted as a special "inherit" value, but is now an invalid gradient. Replaced with the already-defined `$activeBorderColor` / `$inactiveBorderColor` variables, which preserves the original intent (locked groups use the same border color as unlocked ones).

## Tested

Verified with `hyprctl configerrors` — no errors after the fix.


mroboff:
Confirming the `default/hypr/looknfeel.conf` portion of this fix resolves the errors on Hyprland 0.55.0 (Arch). Verified locally with `hyprctl configerrors` — clean after the patch.

One small request: would you mind splitting the `libretro-vice-*` retroarch package changes into a separate PR? They look like a reasonable change on their own, but bundling them here is probably what's holding this up — the looknfeel fix is a one-file, two-hunk regression fix that I suspect maintainers would merge quickly if it stood alone. Plenty of users hitting this right now (cf. #5752, #5758).

Thanks for putting this together.

shawnyeager:
Just applied manually. Can confirm resolves issues.

dev-gios:
Closing — the `dev` branch already migrated to Lua and these issues don't exist there. Reopening against `master` instead.

PR #5770, opened 2026-05-11
Preserve Waybar config when changing position

### LINKED ISSUE #5764: Waybar position command resets custom Waybar config and style
## Summary

`omarchy-style-waybar-position` resets `~/.config/waybar/config.jsonc` and `~/.config/waybar/style.css` to Omarchy defaults before changing the Waybar position.

This makes the new Waybar position menu destructive for users with customized Waybar configs.

## Current Behavior

Running one of these:

```bash
omarchy-style-waybar-position top
omarchy-style-waybar-position bottom
omarchy-style-waybar-position left
omarchy-style-waybar-position right
```

or using:

```text
Omarchy menu -> Style -> Waybar -> Top/Bottom/Left/Right
```

calls:

```bash
omarchy-refresh-config waybar/config.jsonc
omarchy-refresh-config waybar/style.css
```

before editing position/height/width/clock format.

That overwrites user Waybar customizations and leaves only timestamped `.bak` files.

## Expected Behavior

Changing Waybar position should not reset unrelated user Waybar customization.

The command should either:

1. edit only the position-related fields in the existing user config, or
2. detect a customized Waybar config and ask before resetting, or
3. provide a separate explicit reset option.

## Why This Matters

Users commonly customize Waybar layout, modules, CSS, scripts, and icons. A position-only action is expected to be non-destructive, but the current implementation replaces both config and style.

## Relevant Code

```bash
# bin/omarchy-style-waybar-position

# Reset config and styles before making adjustments
omarchy-refresh-config waybar/config.jsonc
omarchy-refresh-config waybar/style.css
```

## Suggested Fix

Remove the automatic refresh from `omarchy-style-waybar-position`.

Instead, update only:

```json
"position"
"height"
"width"
```

and optionally the clock format if the default Waybar config shape is detected.

If custom config is detected, avoid changing clock format unless the user explicitly resets Waybar.

Pegorim:
Closes #5764.

## Summary
- stop refreshing Waybar config/style when only changing Waybar position
- keep the missing-config fallback so first-use still gets the default config
- only switch the clock format when it still matches one of Omarchy's built-in horizontal/vertical defaults

## Testing
- `bash -n bin/omarchy-style-waybar-position`
- `git diff --check`
- ran a temp-home smoke test with stubbed `omarchy-refresh-config`/`omarchy-restart-waybar` verifying:
  - existing custom config fields survive
  - custom `style.css` is untouched
  - left/right updates position, height, width, and default clock format
  - custom clock format survives switching back to top

Note: `shellcheck` is not installed in this environment, so that check was skipped.

### DIFF
--- a/bin/omarchy-style-waybar-position
+++ b/bin/omarchy-style-waybar-position
@@ -15,9 +15,9 @@ if [[ ! $position =~ ^(top|bottom|left|right)$ ]]; then
   exit 1
 fi
 
-# Reset config and styles before making adjustments
-omarchy-refresh-config waybar/config.jsonc
-omarchy-refresh-config waybar/style.css
+if [[ ! -f $WAYBAR_CONFIG ]]; then
+  omarchy-refresh-config waybar/config.jsonc
+fi
 
 if [[ $position == "left" || $position == "right" ]]; then
   height=0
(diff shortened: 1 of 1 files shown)


ryanrhughes:
I wound up doing this same thing before seeing your PR but slightly differently https://github.com/basecamp/omarchy/commit/c0e099fed520c5d7f1efbe459bbdadf5fde293e0. Thanks for sending it over though!

PR #5774, opened 2026-05-12
Add Vivaldi browser support (fixed)

dinvlad:
This is a redo of https://github.com/basecamp/omarchy/pull/5746 with cherry-picked commits, based on @dhh feedback about working off of `dev` branch.

Please let me know if you need anything else.
Also, would love to get feedback on what we should do with theme support!

Original PR description:
---
This adds full Vivaldi browser support, similar to how other Chromium browsers are handled.
1Password support for Vivaldi is added as well.

All scripts seem to be working fine!

I didn't include Chromium theme setting, since Vivaldi seems to be using its own UI layer that doesn't follow Chromium conventions. That said, I did some investigation, and we could add theme support like this:
```sh
# Vivaldi uses its own theme engine and ignores BrowserThemeColor policy.
# This is an example just for accent color, would need to add other color types.
set_vivaldi_theme() {
  local theme_hex="$1"
  local new_theme
  new_theme=$(jq -n \
    --arg accent "$theme_hex" \
    '{
      accentFromPage: false,
      accentOnWindow: true,
      accentSaturationLimit: 1,
      alpha: 1,
      backgroundImage: "",
      backgroundPosition: "stretch",
      blur: 0,
      colorAccentBg: $accent,
      colorBg: "#2e2f37",
      colorFg: "#d3d9e3",
      colorHighlightBg: "#6590fd",
      colorWindowBg: "#1D1E21",
      contrast: 0,
      defaultBackground: "#1D1E21",
      dimBlurred: false,
      engineVersion: 1,
      id: "Omarchy",
      name: "Omarchy",
      preferSystemAccent: false,
      radius: 14,
      simpleScrollbar: true,
      transparencyTabBar: false,
      transparencyTabs: false,
      url: "",
      version: 1
    }')
  for prefs in ~/.config/vivaldi/*/Preferences; do
    [[ -f $prefs ]] || continue
    # Update or insert the Omarchy theme, then activate it
    jq --argjson theme "$new_theme" '
      .vivaldi.themes.user |= map(select(.id != "Omarchy")) + [$theme]
      | .vivaldi.theme.schedule.o_s.dark = "Omarchy"
      | .vivaldi.theme.schedule.o_s.light = "Omarchy"
    ' "$prefs" > "$prefs.tmp" && mv "$prefs.tmp" "$prefs"
  done
}
if omarchy-cmd-present vivaldi-stable; then
  set_vivaldi_theme "$THEME_HEX_COLOR"
fi
```
The caveat is that it would need to be applied while Vivaldi is closed, and then re-opened. It doesn't seem to be possible to apply the theme on-the-fly, unfortunately..

### DIFF
--- a/bin/omarchy-default-browser
+++ b/bin/omarchy-default-browser
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 # omarchy:summary=Set the default browser for Omarchy and XDG handlers
-# omarchy:args=[chromium|chrome|brave|brave-origin|edge|firefox|zen]
+# omarchy:args=[chromium|chrome|brave|brave-origin|edge|firefox|vivaldi|zen]
 # omarchy:examples=omarchy default browser firefox | omarchy default browser brave
 
 if (($# == 0)); then
@@ -12,6 +12,7 @@ if (($# == 0)); then
   brave-origin-beta.desktop) echo "brave-origin" ;;
   microsoft-edge.desktop) echo "edge" ;;
   firefox.desktop) echo "firefox" ;;
+  vivaldi-stable.desktop) echo "vivaldi" ;;
   zen.desktop) echo "zen" ;;
   *) xdg-settings get default-web-browser ;;
   esac
@@ -25,9 +26,10 @@ brave) desktop_id="brave-browser.desktop"; name="Brave"; glyph="󰖟" ;;
 brave-origin) desktop_id="brave-origin-beta.desktop"; name="Brave Origin"; glyph="󰖟" ;;
 edge) desktop_id="microsoft-edge.desktop"; name="Edge"; glyph="󰇩" ;;
 firefox) desktop_id="firefox.desktop"; name="Firefox"; glyph="󰈹" ;;
+vivaldi) desktop_id="vivaldi-stable.desktop"; name="Vivaldi"; glyph="󰰫" ;;
 zen) desktop_id="zen.desktop"; name="Zen"; glyph="󰖟" ;;
 *)
-  echo "Usage: omarchy-default-browser <chromium|chrome|brave|brave-origin|edge|firefox|zen>"
+  echo "Usage: omarchy-default-browser <chromium|chrome|brave|brave-origin|edge|firefox|vivaldi|zen>"
   exit 1
   ;;
 esac
--- a/bin/omarchy-install-browser
+++ b/bin/omarchy-install-browser
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 # omarchy:summary=Install a supported browser
-# omarchy:args=<chrome|brave|brave-origin|edge|firefox|zen>
+# omarchy:args=<chrome|brave|brave-origin|edge|firefox|vivaldi|zen>
 # omarchy:examples=omarchy install browser firefox | omarchy install browser brave
 
 setup_policy_directory() {
@@ -78,6 +78,14 @@ firefox)
   setup_firefox_wayland
   announce_browser_installed "Firefox"
   ;;
+vivaldi)
+  echo "Installing Vivaldi..."
+  omarchy-pkg-add vivaldi || exit 1
+
+  copy_chromium_flags ~/.config/vivaldi-stable.conf
+  omarchy-theme-set-browser
+  announce_browser_installed "Vivaldi"
+  ;;
 zen)
   echo "Installing Zen..."
   omarchy-pkg-aur-add zen-browser-bin || exit 1
@@ -87,7 +95,7 @@ zen)
   announce_browser_installed "Zen"
   ;;
 *)
-  echo "Usage: omarchy-install-browser <chrome|brave|brave-origin|edge|firefox|zen>"
+  echo "Usage: omarchy-install-browser <chrome|brave|brave-origin|edge|firefox|vivaldi|zen>"
   exit 1
   ;;
 esac
--- a/bin/omarchy-install-browser
+++ b/bin/omarchy-install-browser
@@ -82,6 +82,10 @@ vivaldi)
   echo "Installing Vivaldi..."
   omarchy-pkg-add vivaldi || exit 1
 
+  sudo mkdir -p /etc/1password
+  grep -qx vivaldi-bin /etc/1password/custom_allowed_browsers 2>/dev/null || \
+    echo vivaldi-bin | sudo tee -a /etc/1password/custom_allowed_browsers >/dev/null
+
   copy_chromium_flags ~/.config/vivaldi-stable.conf
   announce_browser_installed "Vivaldi"
   ;;
--- a/bin/omarchy-install-browser
+++ b/bin/omarchy-install-browser
@@ -87,6 +87,9 @@ vivaldi)
     echo vivaldi-bin | sudo tee -a /etc/1password/custom_allowed_browsers >/dev/null
 
   copy_chromium_flags ~/.config/vivaldi-stable.conf
+
+  hook="$OMARCHY_PATH/default/vivaldi/vivaldi-post-update"
+  omarchy-hook-install post-update "$hook" && "$hook"
   announce_browser_installed "Vivaldi"
   ;;
 zen)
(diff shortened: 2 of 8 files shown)


brink-lab:
Great! Have been waiting for Vivaldi to be added!

Would be so nice to have full theme handling as well. 
We could consider to set button styles and other UI elements in addition to colors. Many UI elements can be tweaked to a square look with sharp corners to match the rest of Omarchy's aesthetics.

dinvlad:
Woah, might be possible after all by modifying its Chromium storage directly. There's a relevant discussion here https://forum.vivaldi.net/topic/34521/linux-changing-theme-via-command-line/22?lang=en-US&page=2 . I'll take a look 

Yet another alternative is to make use of Omarchy browser extension maybe

dinvlad:
OK dynamic theme changes are implemented! This is done by injecting hex color values from `colors.toml` into Vivaldi's `window.html` (which comprises its UI!) via a static JS script that polls custom `colors.csv` asset and updates static `<style>` element in `window.html` when colors change.

Still TODO:
- I wonder if there's a way to avoid polling in the JS script, which would preserve more battery life (although CPU usage doesn't seem noticeably elevated after I prevented style updates if fetched `colors.csv` hasn't changed)
- Currently, Vivaldi's `colorAccentBg` seems to color the tab bar, but this doesn't look very good with darker themes. Perhaps there's another value in `colors.toml` we could use for that. Similarly, Vivaldi doesn't seem to use the injected colors for its "new tab" page etc. So may need to play with various options here.
- I wonder how we could adjust button styles and other UI elements, as you suggested. Will need to play with CSS as well.

@brink-lab curious re your thoughts!

dinvlad:
OK, I think I might've fixed the tab bar color by using background color instead of accent for `colorAccentBg`. Now, it looks really nice with both the background and foreground colors matched, take a look at this Everforest example 😍

image

brink-lab:
Looks great @dinvlad, nice work! 
Maybe we should just start with getting the live theming in place and part of Omarchy?

It looks like the styling of UI elements might be a bit more fiddly. But here's what I found:

First, CSS modifications should be enabled: 
image

Mistral (chatbot) provides this guide. I haven't tried myself, but might be worth a try :) 
image

brink-lab:
Suggestions by chatGPT

Perfect! Let’s make a **complete, flat UI CSS for Vivaldi on Linux**, covering tabs, buttons, toolbars, panels, menus, pop-ups, and most UI elements that typically have rounded corners. This will give you that crisp, sharp-edge look across the browser.

---

### **1. Location**

On Arch Linux, your profile folder is:

```
~/.config/vivaldi/Default/Custom UI/
```

Make sure the folder exists:

```bash
mkdir -p ~/.config/vivaldi/Default/Custom\ UI
```

Then create/edit `Custom.css`:

```bash
nano ~/.config/vivaldi/Default/Custom\ UI/Custom.css
```

---

### **2. Complete Flat UI CSS**

```css id="f3j9s2"
/* ===========================
   VIVALDI FLAT UI CUSTOM CSS
   Removes border-radius everywhere
   =========================== */

/* --- Tabs --- */
.tab,
.tab.active,
.tab-background,
.tab-header {
    border-radius: 0 !important;
}

/* --- Tab Stack / Drop-down --- */
.tab-stack,
.tab-stack .tab-stack-menu {
    border-radius: 0 !important;
}

/* --- Toolbar & Address Bar --- */
.address-bar,
.toolbar,
.toolbar button,
.toolbar input {
    border-radius: 0 !important;
}

/* --- Buttons & icons --- */
button,
button * {
    border-radius: 0 !important;
}

/* --- Panels / Sidebars --- */
#panel,
.panel-container,
.panel-stack {
    border-radius: 0 !important;
}

/* --- Menus / Popups --- */
.menu,
.menu-item,
.submenu,
.popup,
.dialog,
.context-menu {
    border-radius: 0 !important;
}

/* --- Status bar --- */
.statusbar,
.statusbar * {
    border-radius: 0 !important;
}

/* --- Bookmarks Bar --- */
#bookmarks-bar,
#bookmarks-bar * {
    border-radius: 0 !important;
}

/* --- Tabs corner fixes --- */
.tab-header-container,
.tab-header-background {
    border-radius: 0 !important;
}

/* --- Misc / Tooltips --- */
.tooltip,
.tooltip * {
    border-radius: 0 !important;
}

/* --- Panels / Notes / Downloads --- */
#notes-panel,
#downloads-panel,
#history-panel {
    border-radius: 0 !important;
}

/* --- Settings dialogs --- */
.settings-dialog,
.settings-dialog * {
    border-radius: 0 !important;
}

/* --- Popups for search, find-in-page, etc --- */
#find-in-page,
#search-panel {
    border-radius: 0 !important;
}
```

---

### **3. Enable Custom UI in Vivaldi**

1. Open:

```
vivaldi://experiments/
```

2. Enable **Custom UI Modifications**.
3. Restart Vivaldi.

---

After this, **virtually every UI element** in Vivaldi should have flat, sharp edges.

💡 **Pro tip:** You can tweak individual elements if you notice any stray rounded corners by inspecting them (`Ctrl+Shift+I`) and adding more selectors to `Custom.css`.

---

brink-lab:
The folder structure was created by a first prompt, I gave chatGPT. So I assume we could use a different folder structure if needed.

Here's the answer for the first prompt 

1. Locate your Vivaldi profile folder

On Linux, Vivaldi stores user data under ~/.config/vivaldi/.
The default profile is usually Default.

So the full path is likely:

~/.config/vivaldi/Default
2. Create the Custom UI folder

Inside the Default folder, create a folder named Custom UI:

mkdir -p ~/.config/vivaldi/Default/Custom\ UI

The \ escapes the space if you’re using Custom UI literally.

3. Create the Custom CSS file

Inside the folder Custom UI, create a file named Custom.css:

nano ~/.config/vivaldi/Default/Custom\ UI/Custom.css

Paste this CSS (removes border-radius everywhere):

/* Remove border-radius for all tabs */
.tab,
.tab.active,
.tab-background,
.tab-header {
    border-radius: 0 !important;
}

/* Remove border-radius for address bar and toolbar */
.address-bar,
.toolbar {
    border-radius: 0 !important;
}

/* Remove rounded corners from buttons */
button,
.toolbar button {
    border-radius: 0 !important;
}

/* Remove rounded corners from panels */
#panel {
    border-radius: 0 !important;
}

Save and exit (Ctrl+O, Enter, Ctrl+X in nano).

4. Enable Custom UI in Vivaldi
Open Vivaldi and go to:
vivaldi://experiments/
Enable Custom UI Modifications.
Restart Vivaldi.
5. Verify the changes

All tabs, buttons, panels, and toolbars should now have sharp edges.

💡 Extra Linux tip:
If Vivaldi doesn’t pick up the CSS immediately, sometimes you need to delete the Local State cache:

rm ~/.config/vivaldi/Default/Local\ State

Then restart Vivaldi.

dinvlad:
Hey @brink-lab good suggestion about CSS itself, but I'm not sure we need to add it through "CSS modifications" mechanism. In fact, I explored this mechanism in the process of adding live theming, and it was not a good match because it requires a browser restart and is tricky to set up fully programmatically (need to shut down the browser, edit Preferences file for each Vivaldi profile etc). But thx to live theming support, we already have a place where we add custom css, so I'll just stuff your CSS suggestions there, simple as that. Going the CSS modifications route would just add unneeded complexity!

brink-lab:
@dinvlad your approach sounds much better 👍. 
If it doesn't work, I think we should just go ahead with the theme switching. All the other browsers have there own stylings anyways, so the square ui elements are not a must. But would be a nice touch 😁

dinvlad:
@brink-lab  OK I spent a lot of time playing with CSS, but nothing captured all of the elements.. then I asked Gemini and it showed me the single option in settings that disables roundedness of everything 🤦‍♂ So I'm going to go that route and modify the Preferences, but only this one option - surgical cut incoming 😅

felixzsh:
i want this so bad

PR #5783, opened 2026-05-12
Screen recording: add optional max duration

### LINKED ISSUE #5782: Screen recording: add optional max duration
Currently `Alt+Print` → Screenrecord starts an open-ended recording that must be manually stopped with the same hotkey. I keep forgetting to turn it off and ending up with huge files.

**Proposal:** add a Max duration submenu after the audio selection (e.g. 15min / 30min / 1h / 2h / Unlimited). "Unlimited" preserves today's behavior and stays the default-friendly path. When a duration is chosen, schedule an auto-stop tied to the recorder's PID (so it can't accidentally stop a later recording if the user manual-stops and restarts).

davidguttman:
Closes #5782.

Adds a Max duration submenu to the screenrecord flow (Unlimited / 15min / 30min / 1h / 2h). "Unlimited" is first and preserves today's behavior. When a finite duration is picked, an auto-stop is scheduled against the specific `gpu-screen-recorder` PID, so a manual stop (Alt+Print or the waybar indicator) followed by a new recording can't be killed by an orphaned timer — the timer re-checks `/proc/$pid` before signaling.

The duration prompt happens *before* `omarchy-capture-screenrecording` is invoked, so the prompt itself isn't captured. The two helpers (`wait_for_screenrecorder_pid`, `schedule_screenrecord_autostop`) keep the call site to a single line and document the safety property.

## Test plan
- Unlimited → records until manual stop (unchanged behavior).
- 15min → records and auto-stops at 15min.
- Start 30min, manually stop at ~10s (Alt+Print or waybar click), immediately start a new Unlimited → new recording is not killed at t=30min.
- All four audio/webcam combos still selectable.

### DIFF
--- a/bin/omarchy-menu
+++ b/bin/omarchy-menu
@@ -184,19 +184,59 @@ show_webcam_select_menu() {
 show_screenrecord_menu() {
   omarchy-capture-screenrecording --stop-recording && exit 0
 
+  local audio="" webcam=""
   case $(menu "Screenrecord" "  With no audio\n  With desktop audio\n  With desktop + microphone audio\n  With desktop + microphone audio + webcam") in
-  *"With no audio") omarchy-capture-screenrecording ;;
-  *"With desktop audio") omarchy-capture-screenrecording --with-desktop-audio ;;
-  *"With desktop + microphone audio") omarchy-capture-screenrecording --with-desktop-audio --with-microphone-audio ;;
+  *"With no audio") ;;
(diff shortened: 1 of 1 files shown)


dhh:
Interesting idea, but can't say it's a problem I've ever had. We have the indicator in the menu to remind you that we're recording. But could be a neat script to add on your own setup, then bind a custom hotkey for it. That's the fun part of owning your own operating system!

PR #5786, opened 2026-05-12
Add prefix M-Enter to open a new terminal in popup

vtmx:
Prefx M-Enter open a new terminal in popup mode.


dhh:
We've bound that now to just creating a new pane. Prefer that as that's the most common.

PR #5791, opened 2026-05-13
Preserve branch when reinstalling Omarchy from git

artjsalina5:
## Summary

- Preserve the active Omarchy source branch when reinstalling from git
- Prompt before deleting an existing `~/.local/share/omarchy-new`
- Quote reinstall paths used by the script

## Validation

- `bash -n bin/omarchy-reinstall-git`

### DIFF
--- a/bin/omarchy-reinstall-git
+++ b/bin/omarchy-reinstall-git
@@ -1,11 +1,28 @@
 #!/bin/bash
 
-# omarchy:summary=Reinstall the stable Omarchy source directory from git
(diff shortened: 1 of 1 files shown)


ryanrhughes:
Thanks for this. `omarchy-reinstall` sets back to `stable` only though so this would open room for error. Pinned to `master` again in https://github.com/basecamp/omarchy/commit/9b4a411e5bdb9d08d1b82f564e2461d2e0f603d3

PR #5793, opened 2026-05-13
Add Copy file path to Nautilus right-click context menu

pasmud:
## Summary

Adds a "Copy file path" / "Copy file paths" option to Nautilus' right-click context menu for files and directories.

- **Single file** → shows "Copy file path"
- **Multiple selection** → shows "Copy file paths" (newline-separated)
- Copies absolute path(s) to clipboard via `wl-copy`

## Changes

- `default/nautilus-python/extensions/copy-filepath.py` — new nautilus-python extension
- `install/config/nautilus-python.sh` — install script updated to copy the new extension

## Testing

- Right-click any file/folder → "Copy file path" appears
- Select multiple → "Copy file paths" appears
- Clicking copies the absolute path(s) to your clipboard
- Works on both files and directories
- Gracefully degrades if `wl-copy` is not installed


dhh:
Closing because Nautilus already supports copying a selected file or folder and pasting its filesystem path into a text field, so an additional Copy Path extension is not necessary.

PR #5817, opened 2026-05-14
update clipboard.conf to fix universal copy/paste/cut

Irfrit:
after updating to 3.8.1 universal copy and paste isn't working, 
this fixes it


dhh:
Closing because Quattro completely rewrote clipboard bindings in Lua and includes a more robust active-window implementation.

Issue #5819, opened 2026-05-14
[Bin] Inconsistency in hibernate scripts


c4software:
### System details

Omarchy 3.8.0

### What's wrong?

Hi,

While trying to solve an issue on my laptop (Framework HX 370, issue when s2idle and then hibernate `x86/amd: Previous system reset reason [0x08000800]: an uncorrected error caused a data fabric sync flood event`), I noticed something strange in the following scripts:

- `omarchy-hibernation-setup`
- `omarchy-hibernation-remove`

The `omarchy-hibernation-remove` script removes the `suspend-then-hibernate` configuration:

```
# Remove suspend-then-hibernate configuration
echo "Removing suspend-then-hibernate configuration"
sudo rm -f /etc/systemd/logind.conf.d/lid.conf
sudo rm -f /etc/systemd/sleep.conf.d/hibernate.conf
```

However, this configuration is no longer set up by `omarchy-hibernation-setup`, and these two files are no longer present in the Omarchy repository (`default/config/systemd`).

Is this intended?

c4software:
It's seems the intended behavior, related to this commit @dhh e4b73726, but maybe the deletion should be removed also in the `omarchy-hibernation-remove` script.

dhh:
Yeah, we can remove this now. Do open a PR.

PR #5829, opened 2026-05-15
Remove suspend-then-hibernate cleanup steps

c4software:
Those files are not present anymore (removed here e4b737266687e8ea4d5e97cb7e22f107335028c1)

Removed suspend-then-hibernate configuration cleanup steps from the script.


c4software:
Ping #5819

PR #5831, opened 2026-05-15
Foot as the default terminal

dhh:
Foot is lean, fast and seemingly giving up nothing on Alacritty.

Need to ensure this works on ancient hardware too, though.


csfh:
Is foot replacing Ghostty? I'm confused. 🤔

PR #5838, opened 2026-05-15
Add multi-day forecast to weather widget; click-only popup

jweaver60:
## Summary
- Switches the weather widget's flyout from hover-show/hide to a click-toggled popup (left click toggles, outside-click dismisses via the focus grab). Right click still fires the desktop notification; middle click still refreshes.
- Adds a forecast row to the popup: one cell per day wttr.in returns (typically today + 2) showing day abbreviation, weather glyph, and max/min temps. The per-day icon comes from the hourly entry nearest noon, mapped through the same code → glyph table as the bar pill.
- Replaces the pipe-delimited `wttr.in` fetch with `?format=j1` so we get structured current conditions and the multi-day forecast from one request. Last-good report is preserved across transient failures so the popup never blanks.
- Units default to Celsius; `en_US` / `en_LR` / `my` locales get Fahrenheit. Override per-widget with `"settings": { "unit": "metric" }` or `"imperial"` in `shell.json`.

screenshot-2026-05-15_16-03-06

## Test plan
- [ ] Left-click the weather pill — popup opens; click outside or click the pill again — popup closes
- [ ] Hovering the pill no longer opens the popup; tooltip stays suppressed
- [ ] Right-click the pill — desktop notification with current conditions
- [ ] Middle-click the pill — forecast refetches without toggling the popup
- [ ] Open the popup with the network up — current conditions render, forecast row shows 3 day cells with glyphs and max/min temps
- [ ] Open the popup with the network down — "Fetching forecast…" placeholder shows, no crash
- [ ] Override `unit` to `"imperial"` in `shell.json` — temps and wind switch to °F / mph

### DIFF
--- a/default/quickshell/omarchy-shell/plugins/bar/Bar.qml
+++ b/default/quickshell/omarchy-shell/plugins/bar/Bar.qml
@@ -312,7 +312,7 @@ Item {
     "calendar":           { displayName: "Calendar",           description: "Clock with month-grid popup",                  category: "Time",     allowMultiple: false, settingsForm: "calendarSettings" },
     "notificationCenter": { displayName: "Notification center", description: "Recent notifications + DND",  category: "Status",   allowMultiple: false },
     "systemStats":        { displayName: "System stats",       description: "CPU icon — hover for graphs, click to open btop", category: "System",   allowMultiple: false },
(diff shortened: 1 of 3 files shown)


dhh:
Let's try something tighter. More like:

image

PR #5846, opened 2026-05-15
Stop naming tmux shortcut sessions Work

### LINKED ISSUE #5849: Tmux shortcut should not name new sessions Work
## Problem
Opening tmux with SUPER+ALT+ENTER currently creates a new tmux session named `Work` when no session is attached. That makes the tmux status bar show `Work` in the top-left corner by default.

## Expected behavior
The shortcut should open tmux without forcing a session name, letting tmux use its normal default session naming behavior.

## Related PR
Fixed by #5846: https://github.com/basecamp/omarchy/pull/5846

mplaczek99:
## Summary
- stop SUPER+ALT+ENTER from creating new tmux sessions named Work
- add migration to update existing Hypr bindings

Fixes #5849


dhh:
I put Work in there because I prefer it over the default 0. So it's by design. You can see more about why Omarchy is designed the way it is here: https://learn.omacom.io/3/omacom/76/omakase-computing

PR #5855, opened 2026-05-15
Remove custom clipboard shortcuts to align with terminal defaults

c4software:
Remove custom clipboard key bindings from `foot.ini` to restore the terminal defaults.

The custom shortcuts diverged from the standard `Ctrl+Shift+C` / `Ctrl+Shift+V` bindings used in all other terminal. Removing them aligns foot's behaviour with the rest of the ecosystem.

- Removed explicit `copy` and `paste` key binding entries from `foot.ini`
- Foot will now fall back to its built-in defaults (`Ctrl+Shift+C` / `Ctrl+Shift+V`)

### DIFF
--- a/config/foot/foot.ini
+++ b/config/foot/foot.ini
@@ -14,6 +14,4 @@ style=block
 blink=no
 
 [key-bindings]
-clipboard-copy=Control+Insert
 primary-paste=none
-clipboard-paste=Shift+Insert


dhh:
This would break the unified clipboard, so that's a no go. https://learn.omacom.io/2/the-omarchy-manual/105/universal-clipboard-history

c4software:
Ok, I understand the reason now. Not a big fan, but ok :+1: thanks for the quick follow up.

c4software:
I will use Alacritty instead of foot so, my brain is not braining without Ctrl+Shift+{C,V}.

dhh:
We should see if there's a way to allow both. But can't just remove the unified one.

c4software:
I look at the foot doc, it's not seems possible sadly, not natively at least.

dhh:
It is. I committed this:

```
[key-bindings]
clipboard-copy=Control+Insert Control+Shift+c XF86Copy
primary-paste=none
clipboard-paste=Shift+Insert Control+Shift+v XF86Paste
```

Now it all works.

c4software:
Oh, that's nice. It seems I didn't read the docs correctly.

PR #5856, opened 2026-05-15
Omarchy goes Quickshell

dhh:
All the way.


ali-afk:
Can we add privacy/permission indicators to dock too?

Just-Utkarsh:
Would it be possible to include an audio visualizer like CAVA in the new Quickshell panel that is replacing Waybar? It would be a really nice feature to have built directly into the bar.
image

abenz1267:
quick heads-up cuz i read "noctalia" in some commits: noctalia is moving *away* from quickshell, so adding any compat is pointless i assume.

themes/flexoki-light/colors.toml
```diff
@@ -1,23 +1,34 @@
...
-color7 = "#B7B5AC"
-color8 = "#100F0F"
-color9 = "#D14D41"
-color10 = "#879A39"
-color11 = "#D0A215"
-color12 = "#4385BE"
-color13 = "#CE5D97"
-color14 = "#3AA99F"
-color15 = "#CECDC3"
+bg = "#FFFCF0"
```
bjarneo, line 10: What I noticed now, after it was merged, is that, these already exists:
bg (background), selection (selection_background), fg (foreground). I'll look into a PR to align these.

vitorf7:
It seems as though creating a noctalia compatibility layer as @abenz1267 mentioned will be somewhat "wasted effort" because as announced [here](https://noctalia.dev/blog/announcing-noctalia-v5), Noctalia is moving away from Quickshell unfortunately.

Xeitor:
nooo, what happened?

timmo001:
> nooo, what happened?

Its technically merged, just in https://github.com/basecamp/omarchy/tree/quattro

Xeitor:
> > nooo, what happened?
> 
> Its technically merged, just in https://github.com/basecamp/omarchy/tree/quattro

pheww, looking forward to this so much

PR #5859, opened 2026-05-15
Omarchy helper to generate toml files for local audio file playlists compatible with cliamp

IVII1:
This PR adds an Omarchy helper that builds .toml files in order to play local audio files through cliamp. Makes a playlist of the files in the directory where it was callled.


dhh:
Think it's best if we get something like this built into cliamp itself. Could have `cliamp --create-playlist <dir>` or the like. Let's raise it over there 👍

PR #5860, opened 2026-05-15
Restyle Quickshell settings inputs to match Omarchy aesthetic

jweaver60:
## Summary
- Widget settings dialogs (Calendar, Model Usage, etc.) were rendering text/spin/combo controls with Qt's default light style — black text on dark fills, white dropdown popups, rounded corners, and a pill-style toggle that didn't match the native `StyleToggleRow` switch
- Collapse all `Style.radius*` tokens to `0` so Noctalia-compat widgets inherit Omarchy's sharp-corner convention
- Rewrite `NToggle` to mirror the native 48×22 switch with accent fill and a sliding 16×16 thumb; skin `NComboBox` (contentItem, indicator, popup, delegate) and `NSpinBox` (background + up/down indicators) so dropdowns and spin boxes read on dark themes; replace the default `TextField` chrome in the Calendar form and `DynamicSettingsForm` string field with the same translucent foreground fill used by `WidgetCard`, with an accent border on focus

screenshot-2026-05-15_16-48-45

screenshot-2026-05-15_16-49-09

## Test plan
- [ ] Open Omarchy Settings → Bar → Calendar → cog: three format inputs use the dark translucent fill and pick up an accent border when focused
- [ ] Open Omarchy Settings → Bar → Model Usage → cog: dropdowns are readable closed and open (no white popup), `Refresh interval` spin box matches the rest of the panel, provider toggles look like the System tab's nightlight/notifications switches
- [ ] Verify on a few themes (e.g. one light-on-dark and one with a strong accent) that focus borders/popups still read

### DIFF
--- a/default/quickshell/omarchy-shell/Commons/Style.qml
+++ b/default/quickshell/omarchy-shell/Commons/Style.qml
@@ -9,17 +9,18 @@ import "." as Commons
 QtObject {
   id: root
 
-  // Radii.
-  readonly property real radiusXXS: 1
-  readonly property real radiusXS: 2
-  readonly property real radiusS:  3
-  readonly property real radiusM:  4
-  readonly property real radiusL:  6
-  readonly property real radiusXL: 8
-  readonly property real iRadiusXS: 2
-  readonly property real iRadiusS:  3
-  readonly property real iRadiusM:  4
-  readonly property real iRadiusL:  6
+  // Radii. Omarchy uses sharp corners; all radius tokens collapse to 0 so
+  // Noctalia plugins inherit the same look as native widgets.
+  readonly property real radiusXXS: 0
+  readonly property real radiusXS: 0
+  readonly property real radiusS:  0
+  readonly property real radiusM:  0
+  readonly property real radiusL:  0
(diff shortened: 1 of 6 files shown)


jweaver60:
Thanks @copilot-pull-request-reviewer — went through the four notes and they're either false positives or would regress behavior, so leaving the PR as-is:

- **NSpinBox value sync** — `QQuickSpinBox::contentItemChange` reconnects `editingFinished`/`accepted` on whatever contentItem you give it, and `TextInput` emits both, so typing still updates `value`. The Qt Default/Material styles' contentItem is structurally identical to this one.
- **`visualFocus` vs `activeFocus` (NSpinBox + NComboBox)** — `Control.visualFocus` is only true for `Tab`/`Backtab`/`Shortcut` focus reasons. Switching would drop the accent border on mouse click, which is a regression. The rest of the settings panel (e.g. `StyleToggleRow` in `SettingsPanel.qml`) uses `activeFocus` for the same reason.
- **"so" → "So"** — it's a conjunction mid-sentence, lowercase is correct.

PR #5863, opened 2026-05-16
Convert rounded-corners.conf to lua, fix migration

jadonwb:
A previous migration did not seem to ensure rounded-corners.conf was properly migrated to lua and placed into the proper location. This change enables properly toggling hyprland corners.


jadonwb:
I just saw this is fixed in another commit on a different branch, will close mine.

PR #5873, opened 2026-05-16
Allow manual weather location configuration

oqlatulesba3:
## Summary
- Add support for setting a manual weather location via `~/.config/omarchy/weather-location`
- Useful when using a VPN or in environments where auto-detection returns incorrect location

## Usage
To set a manual location:

```bash
echo "CityName" > ~/.config/omarchy/weather-location
```

Leave the file empty to revert to auto-detection.


dhh:
Closing because Quattro now ships omarchy-weather-location and native weather-panel location selection, covering this use case.

PR #5878, opened 2026-05-16
Browse wallhaven and generate themes via aether from the theme switcher

bjarneo:
The "+" tile at the end of the carousel loads wallhaven's monthly toplist inline. Scrolling near the end auto-fetches the next batch. Picking an image downloads it and runs aether --generate to build and apply the theme.

https://github.com/user-attachments/assets/ac0fef58-177c-4ea9-8877-31ab65fdec97

### DIFF
--- a/bin/omarchy-menu-images
+++ b/bin/omarchy-menu-images
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 # omarchy:summary=Open a generic image selector menu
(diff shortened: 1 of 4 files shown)


dhh:
Closing because this was built on the retired standalone Quickshell image selector. Wallhaven browsing would need a fresh implementation using Quattro’s native background/plugin architecture.

PR #5895, opened 2026-05-18
Add searchLabel support for better menu searchability

tahayvr:
**Sharing this idea as a draft:**

When searching the Omarchy menu, multiple entries often share the same short label. For example, searching for "themes" returns several rows all labeled "Theme", differentiated only by a small breadcrumb subtitle like Style › Install or Remove. This makes search results harder to scan at a glance.

### I propose:
Introducing an optional `searchLabel` field to the `omarchy-menu.jsonc`. When present, the menu uses `searchLabel` as the visible title only during search, while keeping the shorter label for normal browsing. Search results also drop the breadcrumb path detail, so the semantic title stands alone.
screenshot-2026-05-18_00-39-50

I only added `searchLabel` for a few theme-related items just to present the idea. If you like it I can do it for all the items.

## Test it:
When you navigate into Style › Install, you still see the clean short label "Theme". Type "theme" in search and you get Switch Theme, Install Theme, Remove Theme, Update Extra Themes. No breadcrumbs.


tahayvr:
## Update

Menu items now labeled as {parent} {child} in search results. If there is a `searchLabel` field for an item in `omarchy-menu.json`, then that label will be used instead.

@ryanrhughes Is this better? We still need quite a few overrides but not as much.

PR #5896, opened 2026-05-18
Mask network wait-online services to prevent boot delays

reinierbutot:
[Since DHH asked me to create a PR for this on X](https://x.com/dhh/status/2056092360438628358)

Although it evolved and slimmed down from my original post, since the boot is already quite optimized.

**Summary**

This PR prevents some boot delays caused by systemd-networkd-wait-online and NetworkManager-wait-online services on some occasions.

**Changes**

- Updated install/config/hardware/network.sh to always mask systemd-networkd-wait-online.service and guarded-mask NetworkManager-wait-online.service (when present).
- Added a new migration (migrations/1779090175.sh) to safely apply the same change to existing Omarchy installations.

**Why**

These services wait for a network connection before marking the system as "online", which can cause timeouts on Wi-Fi systems or machines with flaky or unplugged interfaces.

**Effect**

- Faster and more reliable boot times.
- Guarded, so no impact on systems that don't have these services.
- Applies cleanly to both fresh installs and existing ones.

### DIFF
--- a/install/config/hardware/network.sh
+++ b/install/config/hardware/network.sh
@@ -4,3 +4,9 @@ sudo systemctl enable iwd.service
 # Prevent systemd-networkd-wait-online timeout on boot
 sudo systemctl disable systemd-networkd-wait-online.service
 sudo systemctl mask systemd-networkd-wait-online.service
+
+# Guarded: only mask if NetworkManager-wait-online exists (for systems using NM)
+if [ -f /usr/lib/systemd/system/NetworkManager-wait-online.service ] || systemctl list-unit-files | grep -q '^NetworkManager-wait-online'; then
+  sudo systemctl disable --now NetworkManager-wait-online.service 2>/dev/null || true
+  sudo systemctl mask NetworkManager-wait-online.service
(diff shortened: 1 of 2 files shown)


dharmavagabond:
This proposal targets NetworkManager which still is in PR (#2909). Omarchy currently uses `systemd-netword`, so it won't be applied to new nor current installs.

dhh:
Thanks for putting this together, Reinier — appreciate you turning the X thread into a PR.

Looking at it against `dev`, the `systemd-networkd-wait-online` mask is already in `install/config/hardware/network.sh`, so the meaningful new bit here is the NetworkManager-wait-online handling. The catch is that Omarchy uses `systemd-networkd` + `iwd` and doesn't ship NetworkManager — so the file guard falls through on every standard install, and the migration would just print "not present, skipping" for everyone.

I'd rather not carry defensive code with no current audience. If NetworkManager support lands (there's an open PR at #2909), the wait-online mask makes more sense as part of that change in one place.

Going to close this one, but thanks again for the contribution.

PR #5897, opened 2026-05-18
Only bind F9 to Voxtype if installed (#5882)

### LINKED ISSUE #5882: F9 should not be bound to Voxtype dictation by default. Bare function keys must not be rebound.
### System details

Omarchy 3.8.1 edge

### What's wrong?

**System details**
Omarchy 3.8.1 (edge channel)

---

**What's wrong?**

In 3.8.1 edge, `F9` is bound by default to start/stop Voxtype dictation. This is a **bare, unmodified function key** being used as a system-wide binding.

Unnecessary to say, but F9 has usage.

Omarchy's design philosophy is to use **modifier-based shortcuts** (Super, Ctrl, Alt combinations) for system-level actions precisely to avoid stomping on application key space. Every other Omarchy system binding respects this. `Super + Ctrl + X` (the existing toggle for dictation) is a perfect example of doing it right.

Binding a bare `F9` breaks this contract entirely. A user who installs Omarchy or updates their system has no reason to expect that their `F9` key will stop working in their editor or game. To make matters worse, `F9` is rebound even when Voxtype is not installed.

**Steps to reproduce**

1. Install Omarchy 3.8.1 edge. You do not even have to install Voxtype via Install > AI > Dictation
2. Try pressing F9 literally anywhere.
5. Either nothing happens if Voxtype is off or Voxtype dictation starts/stops instead of the expected application action.

**Expected behavior**

F9 should be F9...

**Proposed fix**

Remove the bare `F9` binding from the default `bindings.conf`. Keep `Super + Ctrl + X` as the default. Optionally, document F9 (or any F-key the user chooses) as an example opt-in binding in a comment block within the config.

r3quie:
**Only bind F9 to Voxtype if installed (#5882)**

Fixes #5882

Currently, F9 is globally bound to Voxtype dictation, which overrides default app bindings even for users who don't have it installed. 

This PR:
- Conditionally binds F9 in `utilities.lua` only if the `voxtype` command is present.
- Adds an `o.has_command` helper in `helpers.lua`. Had to use `io.popen` instead of an inline `if os.execute("omarchy-cmd-present voxtype")...` for this because Hyprland intercepts the `SIGCHLD` signal. This causes `os.execute` to miss the child's exit code (returning `nil` with an `ECHILD` "No child processes" error), making it fail silently. Reading `stdout` via `.popen`  solves this.

No migration needed.

This is my first proper PR, so I apologize for any oversights on contribution standards; couldn't really find any.

### DIFF
--- a/bin/omarchy-voxtype-install
+++ b/bin/omarchy-voxtype-install
@@ -20,6 +20,7 @@ if gum confirm "Install Voxtype + AI model (~150MB) to enable dictation?"; then
   fi
   voxtype setup systemd
 
+  omarchy-hyprland-toggle voxtype on
   omarchy-restart-waybar
   notify-send "    Voxtype Dictation Ready" "Hold F9 to dictate (or toggle with Super + Ctrl + X)." -t 10000
 fi
--- a/bin/omarchy-voxtype-remove
+++ b/bin/omarchy-voxtype-remove
@@ -18,6 +18,7 @@ if omarchy-cmd-present voxtype; then
   omarchy-pkg-drop voxtype-bin
   rm -rf ~/.config/voxtype
   rm -rf ~/.local/share/voxtype
+  omarchy-hyprland-toggle voxtype off
 else
   echo "Voxtype was not installed."
 fi
--- a/default/hypr/bindings/utilities.lua
+++ b/default/hypr/bindings/utilities.lua
@@ -57,10 +57,6 @@ o.bind("SUPER + CTRL + B", "Bluetooth controls", { omarchy = "bluetooth" })
 o.bind("SUPER + CTRL + W", "Wifi controls", { omarchy = "wifi" })
 o.bind("SUPER + CTRL + T", "Activity", { tui = "btop" })
 
-o.bind("SUPER + CTRL + X", "Toggle dictation", "voxtype record toggle")
-o.bind("F9", "Start dictation (push-to-talk)", "voxtype record start")
-o.bind("F9", "Stop dictation (push-to-talk)", "voxtype record stop", { release = true })
-
 o.bind("SUPER + CTRL + Z", "Zoom in", function()
   local zoom = hl.get_config("cursor.zoom_factor") or 1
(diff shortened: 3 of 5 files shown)


default/hypr/helpers.lua
```diff
@@ -50,6 +50,13 @@ function o.bind(keys, description, dispatcher, options)
   hl.bind(keys, dispatcher, opts)
 end
 
+function o.has_command(cmd)
+  local handle = io.popen("omarchy-cmd-present " .. shell_quote(cmd) .. " && echo 1 || echo 0")
+  local output = handle and handle:read("*a") or ""
+  if handle then handle:close() end
+  return output:match("1") ~= nil
+end
```
NicolasDorier: Never wrote LUA, but I was wondering why this code is so complicated. Would this just work instead?

```lua
function o.has_command(cmd)
  return os.execute("omarchy-cmd-present " .. shell_quote(cmd)) == true
end
```

r3quie: This does not work. Like I said in the PR comment, os.execute never receives the exit code of the command when run live in Hyprland. Your code works fine if you run it straight through Lua or the LuaJIT, just not when Hyprland runs it. I do not really know why but you are free to try on your machine.

NicolasDorier:
I like this PR, as the bind on F9 is really inconvenient and I don't use voxtype. (F9 is the default for toggling a breakpoint in VSCode)

dhh:
I think I'd rather that we add a lua rule as a toggle when voxtype is installed and remove it when uninstalled. Like we do for the single-window-aspect-ratio.lua. Then we don't need any runtime checks.

axelfontaine:
In general, Omarchy should probably namespace almost all its own binds behind SUPER or CAPS LOCK as otherwise the potential for conflict with individual applications is just too high. F9 is also used in mc, IntelliJ and many more...

dhh:
This specific keybind needs a single key for activation, so that just is what it is. But it's your system, so you can always change how it works.

r3quie:
@dhh Reworked. Is this what you meant?

PR #5927, opened 2026-05-20
Fix item tray in bar for shell

vander00:
## Problem
The tray menu doesn't work at all on my machine.

## What I solved
I fixed the menu so a new window pop ups for an item's menu or fallbacks to the default QT's one if it doesn't work.

### DIFF
--- a/shell/Ui/PopupCard.qml
+++ b/shell/Ui/PopupCard.qml
@@ -13,6 +13,7 @@ PopupWindow {
   property int padding: Style.spacing.popupPadding
   property int contentWidth: 280
   property int contentHeight: 200
+  property color backgroundColor: Color.popups.background
   property color borderColor: Color.popups.border
   property bool open: false
   property bool centerOnBar: false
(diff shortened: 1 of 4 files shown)


dhh:
This is way too much code, and not clear at all why it's fixing what it's proposing. Going to need something more surgical. You can try `omarchy refresh shell` to ensure your configs didn't drift. The branch is still in heavy development.

vander00:
> This is way too much code, and not clear at all why it's fixing what it's proposing. Going to need something more surgical. You can try `omarchy refresh shell` to ensure your configs didn't drift. The branch is still in heavy development.

The current tray popup does not work for me at all. Opening a tray item I get:

`TypeError: Cannot read property 'contentItem' of undefined`

That is why I added custom UI here: the goal was to make tray item actions usable, similar to Waybar, where clicking a tray icon like steam shows actions like "exit".

This PR is so big because it actually adds new windows for the tray stuff. It's more like "feat" than "fix", my mistake.

vander00:
@dhh I see the tray has been updated, but it's not fully working for now. Check the PR, it has the menu for trays' items fully working.

dhh:
Working great here as-is. But possible that you need another `omarchy refresh shell`. The omarchy-shell branch really is only intended for Ryan and I for the moment. We'll get it cleaned up before merging to dev.

PR #5932, opened 2026-05-21
Add optional Surge installer

al3rez:
## Summary

Adds `omarchy install surge` for Surge Download Manager.

Surge is available in the AUR as `surge`, so this follows the same pattern as other optional installers: a small `bin/omarchy-install-*` command that installs the package and prints any follow-up steps.

## Details

The installer:

- installs Surge from the AUR
- points users to the latest `extension-chrome.zip` release
- explains the Chrome/Chromium/Brave/Edge unpacked extension flow
- mentions `surge token` for connecting the browser extension

## Test

- `omarchy commands --check`
- `omarchy install surge --help`

### DIFF
--- a/bin/omarchy-install-surge
+++ b/bin/omarchy-install-surge
@@ -1,0 +1,24 @@
+#!/bin/bash
+
+# omarchy:summary=Install Surge Download Manager from the AUR.
+
+set -e
+
+echo "Installing Surge Download Manager..."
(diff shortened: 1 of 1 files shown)


dhh:
Appreciate the idea here, but don't think this adds enough over just letting folks install via AUR, and Surge isn't something I'd personally use.

PR #5945, opened 2026-05-22
Open links originating from apps/terminals in the current workspace

brianblakely:
Right now, apps and terminals open outside links in the last focused browser window, which is not desirable because that might be in an unrelated workspace. This change registers a URL handler that opens links in a browser window from the current workspace, or opens a new browser window if there isn't one.

**Problematic Behavior:** I was looking at pictures of ferns in one workspace, and a cybersecurity report in another. When clicking a link in the report, it opened in my fern workspace.

https://github.com/user-attachments/assets/afa0f415-76fc-4455-8517-fd341c728eea

### DIFF
--- a/applications/hidden/omarchy-handle-url.desktop
+++ b/applications/hidden/omarchy-handle-url.desktop
@@ -1,0 +1,8 @@
+[Desktop Entry]
+NoDisplay=true
+Type=Application
+Name=Omarchy URL Handler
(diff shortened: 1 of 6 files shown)


brianblakely:
While testing, I learned that XDG environments do not commonly distinguish between the default browser and the http scheme handler (even though they can be different). A workaround might be difficult to justify due to its complexity.

PR #5950, opened 2026-05-22
fix: Stable channel update uses tagged release #5948

### LINKED ISSUE #5948: Omarchy update does not update to the actual latest release
### System details

Thinkpad P14s Gen 6 Intel

### What's wrong?

I just ran an Omarchy update from the menu item (stable channel), and I am staring at a default Hyprland install, with zero Omarchy features, at Omarchy version 4.0.0alpha. How the hell does this happen? The update mechanism is looking at the master branch, not the tagged latest release? Ridiculous if this is the case.

An update should be safe, using Omarchy's register of packages tagged at the time of a release (3.8.1). A stable channel release should not be pulling from the master branch.

jcontonio:
This PR updates `bin/omarchy-update-git` to use the latest non-pre-release GitHub release for the stable channel instead of pulling from the tip of master.

## Before

The stable channel update pulled from master, meaning any commit pushed to master (including pre-release code) would be delivered to stable users. Running a stable-channel update jumped me from 3.8.0 to 4.0.0alpha!

## After

The stable channel now queries GitHub for the latest non-pre-release release and checks out that tag. The rc, edge, and dev channels are unchanged.

## To test

1. I added a test here you can run:
  ```sh
  bash test/omarchy-update-git-test.sh
  ```
2. Or manually run an update:
  ```sh
  OMARCHY_PATH=~/.local/share/omarchy bash bin/omarchy-update-git
  git -C ~/.local/share/omarchy describe --tags --exact-match HEAD
  ```

Closes #5948

### DIFF
--- a/bin/omarchy-update-git
+++ b/bin/omarchy-update-git
@@ -12,7 +12,16 @@ omarchy-update-time
 hyprctl keyword debug:suppress_errors true &>/dev/null || true
 hyprctl eval 'hl.config({ debug = { suppress_errors = true } })' &>/dev/null || true
 
-git -C $OMARCHY_PATH pull --autostash
-git -C $OMARCHY_PATH --no-pager diff --check || git -C $OMARCHY_PATH reset --merge
+# Use the actual tagged release for stable, not master
+channel=$(omarchy-version-channel 2>/dev/null || echo "stable")
+
+if [[ $channel == "stable" ]]; then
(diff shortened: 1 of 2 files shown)


dhh:
It's intentional. Master is latest release + any hotfixes. Your issue seems to have been that you were on the dev channel, as Ryan mentioned in the issue. But all of this will go away with Omarchy 4.0 when we switch from git to packages for stable. Thanks for exploring guardrails none the less.

Issue #5780, opened 2026-05-12
Desktop notification showing incorrect battery percentage on dual battery systems


Michael-Steshenko:
### System details

ThinkPad T480

### What's wrong?

The battery percentage showed in the waybar correctly shows the unified battery level from both batteries.
When you right click the battery icon in the waybar, it runs the [omarchy-battery-status script](https://github.com/basecamp/omarchy/blob/dev/bin/omarchy-battery-status), which gets the battery information only for BAT0:
```
battery_info=$(upower -i $(upower -e | grep BAT))
```
This results in inconsistent battery reading between what is shown in the waybar and the notification when you right click it.

Video showing the issue:
https://github.com/user-attachments/assets/d12dc8c7-rbf0a-496d-a84e-b558cefae0fe
Ignore the rounding error between 82% and 81% that is a separate, pre-existing issue.

### Suggested fix
There are 2 ways of fixing this I think could work here, happy to hear opinions.
1. just show the unified reading in the desktop notification by changing 
```
battery_info=$(upower -i $(upower -e | grep BAT))
```
to
```
battery_info=$(upower -i /org/freedesktop/UPower/devices/DisplayDevice)
```
see info about [/org/freedesktop/UPower/devices/DisplayDevice](https://upower.freedesktop.org/docs/UPower.html)

2. Alternatively we could display the information for each battery in a separate row in the desktop notification. 

I am happy to create a PR for either solution.

dhh:
Closing because Quattro replaces the affected Waybar, Mako, SwayOSD, Makima, About, battery, notification, or OSD surface with the native Omarchy shell implementation.

PR #5958, opened 2026-05-23
Fix dual-battery handling in upower queries

stefanomainardi:
## Summary

Fixes the dual-battery inconsistency reported by @Michael-Steshenko in #5780: the waybar percentage and the right-click notification disagree on multi-battery systems (e.g. ThinkPad T480 with internal + slice).

## Root cause

Five scripts share the same anti-pattern:

```bash
upower -i $(upower -e | grep BAT)
```

`upower -e` lists every device path. When two batteries exist, `grep BAT` returns two lines, and `upower -i` is invoked with multiple positional args — it only honors the first. So the scripts report BAT0 only, while waybar's native battery module shows the correct unified value.

## Fix

Switch every call to `upower -i /org/freedesktop/UPower/devices/DisplayDevice`. That's upower's composite virtual device aggregating all power sources — the same data waybar uses internally. Same fields are exposed (`percentage`, `energy-rate`, `state`, `time-to-empty/full`, `energy-full`), so no other code change is needed.

## Scope

All five scripts in the affected callgraph are fixed in a single commit because they all share the same bug from the same root cause:

| Script | Used by |
|---|---|
| `bin/omarchy-battery-status` | waybar right-click + `SUPER+CTRL+ALT+B` notification |
| `bin/omarchy-battery-remaining-time` | called by `battery-status` |
| `bin/omarchy-battery-capacity` | called by `battery-status` |
| `bin/omarchy-battery-remaining` | called by `battery-monitor` |
| `bin/omarchy-battery-monitor` | systemd timer, critical-low alert |

The last two aren't strictly required to fix the symptom in the issue, but the same anti-pattern was silently breaking the critical-low alert as well — on dual-battery systems it could misfire (trigger when only BAT0 hits 10% while BAT1 is full) or never trigger (BAT0 stays charged thanks to BAT1 draining first). Reviewers — happy to split if you prefer one PR per concern.

## Master backport

The same five scripts exist on `master` with identical content (verified: `git diff origin/master:bin/omarchy-battery-{status,remaining,remaining-time,capacity,monitor} origin/dev:...`). If you'd like me to open a parallel PR against `master`, just say so.

## Test plan

- [x] `bash -n` clean on all five scripts
- [x] Smoke-run on a single-battery (well, no-battery — DisplayDevice still exists and returns sensible empty values) system, scripts don't crash and produce parseable output
- [ ] Multi-battery verification — I don't own a ThinkPad T480-class machine; relying on @Michael-Steshenko's root-cause analysis (their issue includes a screen recording showing the exact mismatch). If a reviewer can confirm on real hardware, that would close the loop.

Reported and root-caused by @Michael-Steshenko in #5780 — credit where it's due.

### DIFF
--- a/bin/omarchy-battery-capacity
+++ b/bin/omarchy-battery-capacity
@@ -2,7 +2,7 @@
 
 # omarchy:summary=Returns the battery full capacity in Wh (rounded to whole number).
 
-battery_info=$(upower -i $(upower -e | grep BAT))
+battery_info=$(upower -i /org/freedesktop/UPower/devices/DisplayDevice)
 
 echo "$battery_info" | awk '/energy-full:/ {
     printf "%d", $2
--- a/bin/omarchy-battery-monitor
+++ b/bin/omarchy-battery-monitor
@@ -6,7 +6,7 @@
 BATTERY_THRESHOLD=10
 NOTIFICATION_FLAG="/run/user/$UID/omarchy_battery_notified"
 BATTERY_LEVEL=$(omarchy-battery-remaining)
-BATTERY_STATE=$(upower -i $(upower -e | grep 'BAT') | grep -E "state" | awk '{print $2}')
+BATTERY_STATE=$(upower -i /org/freedesktop/UPower/devices/DisplayDevice | grep -E "state" | awk '{print $2}')
 
 send_notification() {
   notify-send -u critical "󱐋 Time to recharge!" "Battery is down to ${1}%" -i battery-caution -t 30000
--- a/bin/omarchy-battery-remaining
+++ b/bin/omarchy-battery-remaining
@@ -2,7 +2,7 @@
 
 # omarchy:summary=Returns the battery percentage remaining as an integer.
 
-upower -i $(upower -e | grep BAT) | awk '/percentage/ { 
+upower -i /org/freedesktop/UPower/devices/DisplayDevice | awk '/percentage/ {
     print int($2)
     exit
 }'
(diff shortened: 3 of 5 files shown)


Michael-Steshenko:
Thanks for the credit @stefanomainardi, I was waiting for a response from a maintainer on which of the 2 possible fixes I mention in the PR is better.
Thanks for also taking care of all other spots where this issue exists and making a clean PR.
I can validate the fix works across all the script changes when I get to my T480 tomorrow.
Would you mind adding a `Co-authored-by:` trailer to the commit message?

stefanomainardi:
Done in 3d7e22f1 — amended with `Co-authored-by: Michael Steshenko <michael.steshenko@gmail.com>` (using the public email from your GitHub profile; let me know if you'd prefer the noreply alias instead and I'll swap it).

Looking forward to the T480 validation, thanks 🙌

Michael-Steshenko:
Verified everything works properly on my T480 with both batteries connected and with only the internal battery connected!

When both batteries connected:
- Shows unified battery percentage from both batteries when mouse hovering over the Waybar battery icon.
- Shows unified battery percentage from both batteries when right clicking the Waybar battery icon.
- Correctly triggers the critical battery alert based on the unified battery percentage.

Can be safely merged.

Sorry for the delay, life happened ;) @stefanomainardi

dhh:
Closing because Quattro’s native battery service uses UPower.displayDevice, which handles aggregate and dual-battery systems.

PR #5962, opened 2026-05-24
Clock Mini Calendar

brianblakely:
This displays a mini calendar when the clock is clicked or `SUPER + CTRL + ALT + T` is triggered. The calendar uses a `KeyboardPanel` and controls how you would expect. The calendar displays 6 rows of dates so that its height is always the same. In the corner, the current phase of the moon is displayed with a muted theme palette.

screenshot-2026-05-23_21-33-04
screenshot-2026-05-23_21-32-28

### DIFF
--- a/bin/omarchy-menu-timezone
+++ b/bin/omarchy-menu-timezone
@@ -6,5 +6,5 @@
 
 timezone=$(timedatectl list-timezones | omarchy-menu-select "Set timezone" -- --width 520 --maxheight 520) || exit 1
 sudo timedatectl set-timezone "$timezone"
-omarchy-shell -q Clock refresh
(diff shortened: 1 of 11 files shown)


dhh:
I think this will be a great extra panel to offer in a repository for options, but it's not something I personally want in the bar, so can't have it as a default. We'll get a system setup where it's very easy to add in your own options (and offer them to others). Let's tackle it there 👌

brianblakely:
Sure thing, Chef. Will rework this for the extension system when it comes around.

EverStarck:
i like it, would be great if you make it a plugin

brianblakely:
> i like it, would be great if you make it a plugin

You're not gonna believe this
https://github.com/brianblakely/omacal

PR #5967, opened 2026-05-24
Fix: Improve battery status text when full

shmall03:
Updates omarchy-battery-status to properly handle 'fully-charged'.


dhh:
We have a new battery panel in Omarchy 4.0 that'll deal with this. Thanks for contributing!

PR #5969, opened 2026-05-24
Add lupine theme

bjarneo:
Pick, choose, remove, add wallpapers


bjarneo:
The quickshell parts copilot talks about is removed from this branch

dhh:
Merged into `quattro` as 6cea4785 (rebased, so the commits have new SHAs here — your seven are preserved with your authorship intact).

Four commits on top of your work:

- **ef7d5312 → Generate Lupine's app themes from colors.toml** — the palette used the short `bg`/`fg` key spellings, which aren't in the resolver's alias table, so nothing was resolving them. Renamed to the canonical `background`/`foreground` forms and filled in the `orange`/`brown` slots. Also dropped `neovim.lua` and `vscode.json`: those pinned `bjarneo/lupine.nvim` and `Bjarne.lupine-theme`, which users had to install separately, and `default/themed/` now generates both from `colors.toml` alone. Your `aether.nvim` gets the palette injected directly.
- **09e387a8 → Cut Lupine's wallpapers to six and renumber** — dropped city, lavender-bloom, dusk-horizon, purple-mountains, lake-dawn and snowy-dawn, keeping the uniformly pale set that a light theme needs.
- **4e43932c → Redraw Lupine's omarchy wallpaper from the theme palette** — the JPEG drew the wordmark in `#358fff` on pure white, matching neither the accent nor the background, and JPEG artifacts were ringing the letters. Recolored from the shared two-color wordmark so the glyphs land on the same pixels as every other theme; 2KB PNG instead of 71KB.
- **6cea4785 → Separate Lupine's green from its yellow** — `green` and `yellow` were both `#026fde`, so strings and numbers rendered identically. That mattered less when the theme shipped its own editor files; now that they generate from the palette it showed up everywhere. `green` became a violet and `bright_green` a lavender, filling the one gap in the hue range and staying inside the blues-and-purples character.

One thing left, if you want to follow up: `yellow` and `orange` are still identical, and both sit close to `cyan`. Separating them cleanly needs either warm hues or a wider spread than the near-monochrome palette really allows, so I left that call to you.

Thanks for the theme.

PR #5970, opened 2026-05-24
Fix suspend/resume on MacBookPro14,1

erikwb:
These 2017 Kaby Lake, non T1, Intel MacBook Pros fail to resume from the firmware default S3 or even s2idle sleep. Two changes make s2idle suspend reliable:

  - default mem_sleep to s2idle (via a limine-entry-tool drop-in)
  - blacklist the thunderbolt module, which otherwise blocks a clean resume

Wired into install/config/all.sh next to the existing apple fixes and scoped by DMI product name to MacBookPro14,1.


erikwb:
Most of copilot's suggestions look counter to convention in the other hardware scripts, but it does look prudent to just use the drop-in directory.

Issue #166, opened 2025-07-14
Keyboard CAPSLOCK behaves different


ahmetkca:
The title of this issue is self-explanatory: the CAPSLOCK key does not behave as expected.

When I press CAPSLOCK, I expect any character from a to z to be typed in its uppercase form. However, instead, I get the lowercase character with an underline. When I try to delete this character, it doesn't delete immediately. Instead, I see a symbol that looks similar to a double colon but is positioned slightly lower. The worst part is that I cannot copy this symbol into the issue description text box to display it here, nor can I see it in the preview.

Unfortunately, I don’t know how to consistently reproduce this behaviour.

```
 bat /etc/vconsole.conf
───────┬─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ File: /etc/vconsole.conf
───────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1   │ # Written by systemd-localed(8) or systemd-firstboot(1), read by systemd-localed
   2   │ # and systemd-vconsole-setup(8). Use localectl(1) to update this file.
   3   │ KEYMAP=us
   4   │ XKBLAYOUT=us
   5   │ XKBMODEL=pc105+inet
   6   │ XKBOPTIONS=terminate:ctrl_alt_bksp
───────┴─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 localectl status
System Locale: LANG=en_US.UTF-8
    VC Keymap: us
   X11 Layout: us
    X11 Model: pc105+inet
  X11 Options: terminate:ctrl_alt_bksp
```

guilhermetk:
What you're experiencing is actually expected behavior in **Omarchy**, as the `CAPSLOCK` key is configured to act as the **compose key** by default. This setting allows you to type special characters by combining keystrokes, which explains the underlined characters and unusual symbols you're seeing.

To confirm this, check your configuration file:

```bash
~/.config/hypr/hyprland.conf
```

Look for the following line in the `input` section:

```ini
kb_options = compose:caps
```

If you'd prefer to use `CAPSLOCK` for typing uppercase letters (its traditional function), you have a couple of options:

- Disable the compose key entirely:
  ```ini
  kb_options =
  ```

- Change the compose key to `Right Alt` (often labeled \`AltGr\`):
  ```ini
  kb_options = compose:ralt
  ```

[...]

Issue #5200, opened 2026-04-03
Unable to use CAPS LOCK


Siddharthk17:
### System details

Victus by HP Gaming Laptop 15-fa1xxx , 12th Gen Intel(R) Core(TM) i5-12450H (8+4) @ 4.4z , NVIDIA GeForce RTX 2050 [Discrete] , Intel UHD Graphics @ 1.20 GHz [Integrated] , 1920x1080 @ 1.25x in 15", 144 Hz [Built-in] 

### What's wrong?

I am using omarchy since a long time now its been 155+ days since the day 1 i have noticed when i try to click caps lock it doesnt work i thought may be this can be a bug which will get fixed . i am not sure if this is my device issue but when i boot into windows and can easily use caps lock with no issue . not sure why this is happening i tried to resolve but didnt work so far .
hopefully this can get fixed cause its annoying now cause i have to keep pressing shift and then type to get capital letters 

<img width="1890" height="1017" alt="Image" src="https://github.com/user-attachments/assets/54ee60ae-9afb-40b3-af20-753e8cfde9c6" />
this is what happens when i try to press caps lock not only in terminal but even in browser or everywhere

morgoth:
This is by design https://learn.omacom.io/2/the-omarchy-manual/88/troubleshooting?search=caps#why-isn-39-t-caps-lock-working

Siddharthk17:
thank you for the help

Issue #3238, opened 2025-11-08
Caps Lock isn't working


Carloswelliton:
### System details

intel i3-1005G1, omarchy 3.1.5

### What's wrong?

when i press the CapsLock button its only shows me "·"
how can i configurate this?

dharmavagabond:
Add `kb_options = caps:capslock` under the `input` option in `~/.config/hypr/input.conf`.

https://wiki.hypr.land/Configuring/Variables/#input
https://man.archlinux.org/man/xkeyboard-config-2.7.en#Caps_Lock_behavior

MukaramNadeem:
It's for input fractions like 2/3 or p/q.

morgoth:
This is by design https://learn.omacom.io/2/the-omarchy-manual/88/troubleshooting#why-isn-39-t-caps-lock-working

Carloswelliton:
thank u all, it helps a lot

Issue #2545, opened 2025-10-19
Caps Lock not working on Omarchy 3.1


rthabett:
### System details

Omarchy 3.1

### What's wrong?

Description:
After a clean install of the new Omarchy 3.1, the Caps Lock key no longer works on my laptop. Pressing it doesn’t toggle uppercase letters or the LED indicator same for externel keyboard. It was working normally in previous versions.

Steps to Reproduce:
Boot into Omarchy 3.1
Open any text editor or terminal
Press Caps Lock
Observe that text input remains lowercase and the Caps Lock indicator doesn’t activate

Expected Behavior:
Caps Lock should toggle uppercase letters .

Actual Behavior:
Caps Lock has no effect at all.

System Information:
Omarchy version: 3.1
Keyboard layout: fr
Hardware: Dell inspiron 3737

Additional Context:
Using the Shift key still produces uppercase letters normally.

wekauwau:
https://learn.omacom.io/2/the-omarchy-manual/88/troubleshooting#why-isn-39-t-caps-lock-working

rthabett:
> https://learn.omacom.io/2/the-omarchy-manual/88/troubleshooting#why-isn-39-t-caps-lock-working
thnx

AnsBdran:
how about assigning the old `caps lock` functionality to another shortcut, like `ctrl` + `caps lock` or `ctrl` + `shift`?

dhh:
I'd be happy to take a PR for that. (Ctrl + Caps Lock).

PR #5981, opened 2026-05-26
Allow toggling Caps Lock via both Shift keys

tyvsmith:
Adds `shift:both_capslock` alongside the existing `compose:caps` in `default/hypr/input.lua` and `config/hypr/input.lua`. Pressing **both Shift keys simultaneously** now toggles real Caps Lock; single-Shift behavior is unchanged, and Caps Lock continues to act as the XCompose key.

## Why

The current default `compose:caps` makes Caps Lock the XCompose key — great for emoji, accented characters, and `~/.XCompose` shortcuts — but it removes any way to enable real Caps Lock, which is occasionally valuable. More importantly, fcitx5's Wayland input-method grab occasionally desyncs (notably after a Hyprland reload), and when it does, Caps Lock can get stuck "on" with **no easy way to toggle it back off** — the Compose key is the very thing that's broken. This recurring confusion shows up across closed issues #166, #953, #1792, #2241, #2545, #3238, and #5200.

`shift:both_capslock` is a stock libxkbcommon option that provides an escape hatch. It costs nothing for users who don't reach for it:

- Single Shift is untouched.
- Caps Lock alone remains the Compose key.
- The Compose-key default that DHH preserved in #2525 stays exactly as it is.

This is purely additive — it restores the missing Caps Lock capability via a discoverable two-Shift chord without changing any existing behavior.

## Note on syntax

The XKB option string is comma-separated and **must not contain spaces** between options. `"compose:caps,shift:both_capslock"` is correct; `"compose:caps, shift:both_capslock"` would cause libxkbcommon to silently drop the second option. The change uses the correct space-free form.

## Test plan

- [x] Press both Shifts simultaneously → Caps Lock LED toggles, alphabetic keys produce uppercase.
- [x] Single Shift still produces uppercase only while held.
- [x] Caps Lock alone still acts as Compose / Multi_key 
- [x] `~/.XCompose` shortcuts still work.

### DIFF
--- a/config/hypr/input.lua
+++ b/config/hypr/input.lua
@@ -8,7 +8,7 @@ hl.config({
     -- Use a specific keyboard variant if needed (e.g. intl for international keyboards).
     -- kb_variant = "intl",
 
-    kb_options = "compose:caps", -- ,grp:alts_toggle
+    kb_options = "compose:caps,shift:both_capslock", -- ,grp:alts_toggle
 
     -- Change speed of keyboard repeat.
     repeat_rate = 40,
--- a/default/hypr/input.lua
+++ b/default/hypr/input.lua
@@ -4,7 +4,7 @@ hl.config({
     kb_layout = "us",
     kb_variant = "",
     kb_model = "",
-    kb_options = "compose:caps",
+    kb_options = "compose:caps,shift:both_capslock",
     kb_rules = "",
     follow_mouse = 1,
     sensitivity = 0,


silopolis:
Also very much liked the Ctrl + Capslock option proposed in #2545 but as long as I can get back access to those cute uppercase accented characters we "like" in France :roll_eyes:  ...
Thanks for bringing this :pray:

tyvsmith:
> Also very much liked the Ctrl + Capslock option proposed in #2545 but as long as I can get back access to those cute uppercase accented characters we "like" in France 🙄 ... Thanks for bringing this 🙏

From looking around in the community it seemed double shift as an alternate was a bit more common, and I had heard some odd behavior of modifier keys with capslock can happen, thus went with double shift.  Happy to switch the PR to ctrl + caps if there's a strong signal to go that direction though.

dhh:
Thanks! Pulled into the quattro branch as 8147489f with author credit — it'll land with the quattro merge.

PR #5985, opened 2026-05-27
fix: prefer dedicated music players over browsers for media keys

loicfavory:
## Problem

Browsers (Chromium, Firefox) register as MPRIS players as soon as they open, even when no media is playing. When `playerctld` is running, it can pick the browser as the active player — causing media keys (play/pause/next/prev) to silently target the browser instead of a running music app (cliamp, Spotify, etc.).

The result: play/pause only works intermittently, or not at all, depending on which player `playerctld` happens to have prioritized.

## Fix

Extend `omarchy-swayosd-client` to detect this situation when `--playerctl` is used without an explicit `--player`: it scans all MPRIS players and injects `--player <best>` with the following priority:

| Priority | Condition |
|----------|-----------|
| 1st | Dedicated music player that is **Playing** |
| 2nd | Browser that is **Playing** (e.g. YouTube actively running) |
| 3rd | Dedicated music player that is **Paused** |
| 4th | Browser that is **Paused** |

This means:
- cliamp/Spotify playing + Chrome open → **music player gets the command** ✓
- YouTube playing + cliamp paused → **YouTube gets the command** ✓
- Only Chrome open (stopped) → falls back to default playerctl behavior ✓

## Scope

- No changes to `media.lua` or any bindings
- Transparent when `--player` is already supplied explicitly
- Only activates for `--playerctl` calls, leaving all other swayosd behavior untouched

### DIFF
--- a/bin/omarchy-swayosd-client
+++ b/bin/omarchy-swayosd-client
@@ -1,6 +1,37 @@
 #!/bin/bash
 
 # omarchy:summary=Wrapper for swayosd-client that targets the currently focused monitor.
+# omarchy:summary=When --playerctl is used without an explicit --player, automatically
+# omarchy:summary=prefers dedicated music players (e.g. cliamp, spotify) over browser
+# omarchy:summary=instances (chromium, firefox) that register as MPRIS players even when
+# omarchy:summary=no media is playing.
 # omarchy:args=<swayosd-client-args...>
+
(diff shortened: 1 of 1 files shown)


dhh:
Closing because Quattro’s native media service now selects active controllable players, avoids the playerctld proxy where appropriate, and supports explicit source switching.

PR #5988, opened 2026-05-28
Fix documentation example

othondev:
Fixing screen recording instead of screen record, which only works when using the alias.


dhh:
Pulled into Quattro as ca0d78e7, preserving Othon Ferreira's authorship. Thanks!

PR #5995, opened 2026-05-28
Add Setup -> Config entries for default applications (vim, tmux, etc.)

bfloch:
Dynamically adds options to `Setup -> Config`, if the apps are installed.
These are all apps that are already handled by Omarchy, so I think it makes sense:

-  [x] Ghostty `~/.config/ghostty/config`
- [ ] Foot `~/.config/foot/foot.ini`
- [ ] Kitty `~/.config/kitty/kitty.conf`
- [x] Tmux `~/.config/tmux/tmux.conf` (with reload)
- [x] Neovim `~/.config/nvim/init.lua`
- [ ] Helix `~/.config/helix/config.toml`
- [ ] Emacs `~/.config/emacs/init.el`
- [ ] Zed `~/.config/zed/settings.json`
- [ ] VSCode `~/.config/Code/User/settings.json`
- [ ] Cursor `~/.config/Cursor/User/settings.json`

Some apps don't have a definite answer on where to store config files (Emacs?). I tried to keep things simple with suggested defaults in `.config` (with AI feedback).

> [!WARNING]
> I have only tested the checked apps above - Feedback welcome.

image

### DIFF
--- a/bin/omarchy-menu
+++ b/bin/omarchy-menu
@@ -489,7 +489,23 @@ show_setup_default_editor_menu() {
 }
 
 show_setup_config_menu() {
-  case $(menu "Setup" "  Hyprland\n  Hypridle\n  Hyprlock\n  Hyprsunset\n  Swayosd\n󰌧  Walker\n󰍜  Waybar\n󰞅  XCompose") in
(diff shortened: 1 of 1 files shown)


Linar46:
In VSCode we have multiple profiles. So you have to add them dynamically.

dhh:
Appreciate the concept here, but I think it's a bit too much, and doesn't quite add enough over just editing directly.

PR #4732, opened 2026-02-24
eDP Auto Refresh Rate Switching

saidbenmoumen:
to sum this up i would love to get as much battery juice as possible from my laptop. 

so i made this trigger (opt-in) to automatically switch between low refresh rate & high refresh rate of laptop's preferred resolution based on active power profile.

only tested on my Asus G14:
- Performance / Balanced: 120hz
- Power Saver: 60hz

not sure if we should modify in the user `monitors.conf` directly for this or not, but for now its made to infer it for the most part & change only refresh rate using `hyprctl keyword`.

NOTE: only runs on eDP & option shows up only if you have a battery..

PR #6002, opened 2026-05-30
Add Refresh Rate toggle to Toggle menu

ocewers:
## What

Adds a **Refresh Rate** entry to the Toggle menu (next to Monitor Scaling) that cycles the focused monitor through the refresh rates its current resolution supports — e.g. 120Hz → 60Hz → 120Hz.

New script `omarchy-hyprland-monitor-refresh-cycle` + one line in the Toggle menu. Runtime-only (no `monitors.conf`/`monitors.lua` changes, reverts to the panel default on reboot).

## Relationship to #4732

I know refresh-rate switching was already proposed in #4732 and you closed it. This is deliberately a much smaller thing, and I want to be upfront about the differences so you can judge it on its own terms:

| #4732 (closed) | This PR |
|---|---|
| **Automatic** switching tied to power profiles | **Manual** toggle the user invokes |
| Power-profile hooks + battery detection | None — just reads Hyprland state |
| 4 files, +95 lines | 1 script + 1 menu line, ~58 lines |
| Changes runtime behavior on its own | Inert until the user picks it from the menu |

So this changes **no default behavior** — it's an opt-in menu action, conceptually a sibling of the existing "Monitor Scaling" toggle. If a user never opens it, nothing changes.

## On the power question

You noted in #4732 that you rarely saw meaningful savings, which is fair — it's display- and workload-dependent. For transparency, here's a measurement on a ThinkPad X1 Carbon Gen 14 (2880x1800 OLED), idle desktop:

- 120Hz: ~5.4W
- 60Hz: ~4.2W

That's ~22% lower idle draw (~3h more on a 58Wh battery). The effect is largest at idle/light desktop use and shrinks under video/gaming where the GPU dominates — i.e. it helps most exactly when you don't need 120Hz. OLED panels in particular seem to benefit more than the LCDs you may have tested.

I completely understand if the savings still don't clear your bar for the general distribution. I'm submitting it because the *manual, zero-default-change* framing is a different trade-off than #4732, and a refresh-rate toggle is a natural companion to the scaling toggle that's already there.

## Generic by design

No hardcoded rates. It reads `availableModes` from Hyprland (EDID), cycles whatever the panel reports for the current resolution (60/90/120/144/165…), preserves the current scale, targets the focused monitor, and no-ops with a notification when only one rate exists. So it's safe on single-rate panels and external monitors.

## Test plan

- 120↔60 forward cycle, `--reverse`, scale preserved across the switch — verified on eDP-1 (2880x1800@120/60).
- Single-rate panel path → notification, no change.
- `bash -n` clean on both files.

### DIFF
--- a/bin/omarchy-hyprland-monitor-refresh-cycle
+++ b/bin/omarchy-hyprland-monitor-refresh-cycle
@@ -1,0 +1,57 @@
+#!/bin/bash
+
+# omarchy:summary=Cycle focused Hyprland monitor through its available refresh rates (e.g. 120Hz ↔ 60Hz)
+
+# Cycles the focused monitor's refresh rate through every rate its current
+# resolution supports (read live from Hyprland's availableModes). Lowering the
+# refresh rate is a simple, reversible way to save power on a laptop panel.
+#
+# Pass --reverse to cycle the other way.
+
+MONITOR_INFO=$(hyprctl monitors -j | jq -r '.[] | select(.focused == true)')
+ACTIVE_MONITOR=$(echo "$MONITOR_INFO" | jq -r '.name')
+CURRENT_RATE=$(echo "$MONITOR_INFO" | jq -r '.refreshRate')
+CURRENT_SCALE=$(echo "$MONITOR_INFO" | jq -r '.scale')
+WIDTH=$(echo "$MONITOR_INFO" | jq -r '.width')
+HEIGHT=$(echo "$MONITOR_INFO" | jq -r '.height')
+
+# Collect the refresh rates available at the *current* resolution, sorted
+# high → low and de-duplicated. availableModes entries look like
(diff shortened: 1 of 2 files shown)


ocewers:
Closing this. After more careful, controlled A/B measurements (alternating 60/120Hz back-to-back, averaged over multiple 30s rounds at fixed brightness on a ThinkPad X1 Carbon Gen 14 OLED), the idle power difference averaged out to within measurement noise — essentially 0W, not the larger figure I initially cited. That earlier number came from an uncontrolled single-sample comparison and doesn't hold up. This matches the skepticism in #4732: modern panels (with self-refresh) draw about the same at 60 vs 120Hz when content is static. The savings don't justify the addition. Apologies for the noise — thanks.

PR #6015, opened 2026-06-01
feat: add lockscreen toggle to support screensaver-only mode

SilvM:


dhh:
Closing because this is built around Hypridle, Waybar, and the old lockscreen, all replaced by Quattro’s native idle and lock services. Screensaver-only mode can be re-proposed there.

PR #6017, opened 2026-06-01
Support local favicon discovery for webapp install

shbernal:
## Summary
- Allow `omarchy webapp install <name> <url>` to run non-interactively without an icon argument.
- Try the web app's advertised favicon first, including relative icon paths, before falling back to Google's favicon service.
- Keep explicit icon URLs and saved local icon names working as before.

## Tests
- `bash test/omarchy-webapp-install-test.sh`
- `git diff --check upstream/dev...HEAD`


dhh:
Appreciate the suggestion, but too much code weight for the fix.

PR #6027, opened 2026-06-03
Show version preview before update confirmation

Pegorim:
## Summary

- Show the installed Omarchy version/ref in the update confirmation prompt
- Fetch the tracked upstream ref and show the target version/ref before the irreversible update starts
- Include channel/tracking information and a note when Omarchy itself is already at the target version

## Tests

- `bash -n bin/omarchy-update-confirm`


dhh:
Not sure there's enough value here. Especially given the extra weight. Appreciate the suggestion though!

PR #6032, opened 2026-06-04
feat: preview and override the icon when installing a web app

CaelRowley:
When adding a web app, you previously only got to choose an icon if the automatic favicon lookup *failed* and you never saw what icon you'd end up with.

Now, after entering the URL, the installer fetches the favicon, shows a preview of it in the terminal, and prompts for an icon:

- Press **Enter** to accept the previewed favicon.
- Paste a **PNG URL** (e.g. from dashboardicons.com) to use a custom icon instead.

If the favicon can't be fetched, it falls back to the previous behavior (prompt for a PNG URL).

The preview uses `chafa` (added to the base package list); when it isn't available the flow still works, just without the preview.

image

If the terminal does not support showing images via chafa it will default to showing just the clickable link

image

### DIFF
--- a/bin/omarchy-webapp-install
+++ b/bin/omarchy-webapp-install
@@ -15,18 +15,33 @@ if (( $# < 3 )); then
     APP_URL="https://$APP_URL"
   fi
 
-  # Fetch the favicon to offer as the default icon, and preview it so you can
-  # decide whether to keep it or supply your own.
+  # Fetch the favicon to offer as the default icon. Terminals with a real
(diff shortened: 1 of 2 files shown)


dhh:
Chafa is unfortunately not reliable/good in Alacritty. So don't think this is going to work for all the terminals we support.

CaelRowley:
@dhh I updated it so that it will use Chafa if the terminal supports showing images, otherwise it will show just a clickable link to the favicon, let me know what you think

dhh:
Appreciate the idea, but don't like to have different paths here. Thanks for the suggestion!

PR #6035, opened 2026-06-05
Fix bi-directional USB-C ports falsely triggering AC/Performance mode (#5966)

### LINKED ISSUE #5966: bug: Bi-directional USB-C port charging a phone falsely triggers AC/Performance mode on Lenovo LOQ
### System details

NVIDIA RTX 2050

### What's wrong?

On the Lenovo LOQ, the USB-C ports support dual-role power (input and output). When a phone is plugged into this port to charge from the laptop, the system/kernel registers a power supply state change.

Because omarchy triggers Performance mode globally whenever it detects a charging state, it incorrectly interprets this outbound power delivery as the laptop being plugged into an AC wall charger. Unplugging or replugging the phone cycles this state, forcing the laptop back into Performance mode every time.

### Expected Behavior

The system should only switch to Performance mode when the laptop's internal battery is actively receiving power from an external source (AC adapter or external PD charger), not when it is sourcing power to an external client device.

Rockeyxx:
Fixes basecamp/omarchy#5966. Skips USB power supply devices reporting 'Discharging' status to avoid false AC detection when charging client devices.

### DIFF
--- a/bin/omarchy-powerprofiles-set
+++ b/bin/omarchy-powerprofiles-set
@@ -17,10 +17,15 @@ if [[ -z $action || $action == "autodetect" ]]; then
   sleep 0.3
 
   action=battery
-  for ps in /sys/class/power_supply/*; do
+  power_supply_dir="${POWER_SUPPLY_DIR:-/sys/class/power_supply}"
+  for ps in "$power_supply_dir"/*; do
     [[ -r $ps/online && -r $ps/type ]] || continue
     type=$(cat "$ps/type")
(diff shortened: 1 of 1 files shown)


dhh:
Thanks for digging into this! But this code has been rewritten on the quattro branch, which is about to become the next major release. Autodetection no longer reads /sys/class/power_supply at all — it now keys off UPower's OnBattery signal (same source the shell uses), so a dual-role port sourcing power to a phone won't flip the profile to performance anymore. If you can still reproduce #5966 on quattro, please reopen the issue with the details.

PR #6038, opened 2026-06-05
Enable IPv6 privacy extensions by default

pkwagner:
Enable IPv6 privacy extensions by default and add a migration for existing users. The migration does deliberately not restart `systemd-networkd` afterwards to avoid network issues during the following system upgrade; changes will be applied after reboot.

Without PE, clients use a static IPv6 suffix derived from their MAC address, which allows every web service (including any website visited in private mode) to identify and track users just by their IP. PE mitigate this by using temporary randomized addresses, and is therefore the default behavior of essentially all client operating systems like macOS, Windows, and most non-server Linux distros.

### DIFF
--- a/install/config/hardware/network.sh
+++ b/install/config/hardware/network.sh
@@ -1,6 +1,13 @@
 # Ensure iwd service will be started
 sudo systemctl enable iwd.service
 
+# Enable IPv6 privacy addresses by default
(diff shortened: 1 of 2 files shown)


pkwagner:
Quattro seems to move to NetworkManager, which uses PE by default. Closing it here :)

PR #6046, opened 2026-06-06
Make power profile defaults configurable

crmne:
## Summary

This is a smaller version of #5621: no menu, just a persistent config file.

`omarchy-powerprofiles-set` now reads:

```ini
# ~/.config/omarchy/powerprofiles.conf
ac=performance
battery=balanced
```

Those are today's defaults, so behavior is unchanged unless the user edits the file. If the file is missing or contains an unavailable profile, the helper falls back to the existing defaults.

The existing udev rule is unchanged. This makes the Omarchy helper configurable and seeds the default config file.

One important detail: udev launches the helper through the system manager, so `$HOME` is not guaranteed to be the desktop user's home. The helper therefore still honors `OMARCHY_CONFIG_HOME` first, but otherwise derives the user config path from its installed executable path under `~/.local/share/omarchy/bin`. That lets the same config file work from boot, Hyprland autostart, and udev AC/USB events without changing the udev rule.

## Why

Editing the udev rule directly does not hold up as a persistent user preference:

- boot applies the profile outside udev via `omarchy-powerprofiles-init`
- Omarchy regenerates/replaces the power-profile udev rule during migrations/updates
- udev/systemd runs the helper outside the user session, so relying on `$HOME` alone can miss `~/.config/omarchy/powerprofiles.conf`

This gives users an Omarchy-supported override without changing defaults for everyone.

## Testing

- `bash -n bin/omarchy-powerprofiles-set migrations/1780735021.sh`
- `git diff --check`
- `bin/omarchy-powerprofiles-set ac` with local `ac=balanced`
- smoke-tested an installed-path helper with `HOME=/root` and `ac=balanced`, verifying it still reads the user config from the executable path
- smoke-tested `OMARCHY_CONFIG_HOME` override precedence
- `OMARCHY_CONFIG_HOME=/tmp/omarchy-powerprofiles-migration-test/.config/omarchy OMARCHY_PATH=/home/carmine/Code/omarchy bash migrations/1780735021.sh`

### DIFF
--- a/bin/omarchy-powerprofiles-set
+++ b/bin/omarchy-powerprofiles-set
@@ -4,6 +4,21 @@
 # omarchy:args=[autodetect|ac|battery]
 
 action="${1-}"
+config_dir="${OMARCHY_CONFIG_HOME:-}"
+
+if [[ -z $config_dir ]]; then
+  # udev runs this through systemd outside the user session, so $HOME
+  # may be /root. Derive the user config path from the install path.
+  executable=$(readlink -f "$0")
+
+  if [[ $executable == */.local/share/omarchy/bin/omarchy-powerprofiles-set ]]; then
+    config_dir="${executable%/.local/share/omarchy/bin/omarchy-powerprofiles-set}/.config/omarchy"
+  else
+    config_dir="$HOME/.config/omarchy"
+  fi
+fi
+
+config_path="$config_dir/powerprofiles.conf"
 
 # Auto-detect when called with no argument: treat any Mains or USB
 # power-supply device reporting online=1 as "on AC". This handles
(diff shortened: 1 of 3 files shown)


MrDach:
This would be great to be able to configure! My Dell Latitude with an i7 will also start taking off if I keep it in performance mode.

For now I've added the following to `~/.config/hypr/autostart.conf`
```
exec-once = sh -c 'sleep 3; powerprofilesctl set balanced'
```

Which is a terrible hack but I just get annoyed by having to do the change manually on every boot.

Shame that your config menu PR was closed.

crmne:
Thanks for the comment @MrDach. I know a ton of people having the same problem.

tlxq:
This is great guys! Struggle with this on both of my lenovo t14s

dhh:
We now save the last setting you had in quattro 👍

PR #6047, opened 2026-06-06
Stop forcing Qt style override

crmne:
## Summary

Remove the global `QT_STYLE_OVERRIDE=kvantum` export from the default Hyprland environment.

`QT_QPA_PLATFORM` still defaults Qt apps to Wayland/XCB; this only stops forcing the Qt style for every Qt app.

## Why

Forcing `QT_STYLE_OVERRIDE` globally prevents Qt settings tools and individual Qt apps from applying their configured style. Omarchy can set the platform default without making the visual Qt style a session-wide override.

## Testing

- `lua -e 'assert(loadfile("default/hypr/envs.lua"))'`
- `git diff --check origin/dev..cp/drop-qt-style-override`

### DIFF
--- a/default/hypr/envs.lua
+++ b/default/hypr/envs.lua
@@ -18,7 +18,6 @@ hl.env("HYPRCURSOR_SIZE", "24")
 -- Force all apps to use Wayland.
 hl.env("GDK_BACKEND", "wayland,x11,*")
 hl.env("QT_QPA_PLATFORM", "wayland;xcb")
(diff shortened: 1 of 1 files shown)


dhh:
Going to need some before/after screenshots here too.

vander00:
I think the sense is all the apps to look the same, and it adds the feature to change all the QT theme on a system just with kvantum.

dhh:
Happy with the uniform style we have now.

PR #6048, opened 2026-06-06
Add Mako theme and user overrides

crmne:
## Summary

Add two override layers to Omarchy's generated Mako config:

```ini
include=~/.config/omarchy/current/theme/mako-theme.ini
include=~/.config/omarchy/mako.ini
```

Themes can now ship `mako-theme.ini` for notification styling, while users get a persistent `~/.config/omarchy/mako.ini` override that survives theme switches. Default empty files are included so Mako never fails on a missing include.

## Why

Core notification behavior, Omarchy toggles, and generated theme colors should stay in the main Mako config. Theme-specific and user-specific tweaks are better as small override fragments instead of replacing the whole generated file.

## Testing

- `bash -n migrations/1780740711.sh bin/omarchy-theme-set-templates`
- `git diff --check origin/dev..cp/mako-overrides`
- generated a Mako config from Catppuccin colors and verified both include lines plus the default empty `mako-theme.ini`
- parsed a temporary generated Mako config with both included files present

### DIFF
--- a/config/omarchy/mako.ini
+++ b/config/omarchy/mako.ini
@@ -1,0 +1,1 @@
+# User-specific Mako overrides.
--- a/default/themed/mako-theme.ini.tpl
+++ b/default/themed/mako-theme.ini.tpl
@@ -1,0 +1,1 @@
+# Theme-specific Mako overrides.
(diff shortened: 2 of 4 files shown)


dhh:
We're moving away from Mako in Omarchy 4. It's being replaced by a Quickshell-based notification system.

PR #6049, opened 2026-06-06
Show hostname in Starship over SSH

crmne:
## Summary

Add Starship's `hostname` module to the prompt with `ssh_only = true`.

Local prompts remain unchanged, but SSH sessions show the hostname before the directory/git segments.

## Why

When working across remote shells, the current prompt makes it too easy to lose track of which machine a terminal is connected to. Starship already supports this exact use case with `ssh_only`, so this keeps the normal local prompt clean while making remote sessions clearer.

## Testing

- `STARSHIP_CONFIG=$PWD/config/starship.toml starship explain`
- `bash -n migrations/1780739892.sh`
- smoke-tested the migration against the previous default Starship format
- `git diff --check origin/dev..cp/starship-ssh-hostname-clean`

### DIFF
--- a/config/starship.toml
+++ b/config/starship.toml
@@ -1,6 +1,6 @@
 add_newline = true
 command_timeout = 200
-format = "[$directory$git_branch$git_status]($style)$character"
+format = "[$hostname$directory$git_branch$git_status]($style)$character"
 
 [character]
 error_symbol = "[✗](bold cyan)"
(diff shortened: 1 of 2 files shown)


dhh:
Need a screenshot here too, but I don't think I'm likely to be onboard, because it would duplicate the hostname we already have in the upper right with tmux.

PR #6050, opened 2026-06-06
Use delta for Git and LazyGit diffs

crmne:
## Summary

Install `git-delta` and configure Git/LazyGit to use it for diffs.

Git gets:

```ini
[core]
  pager = delta
[interactive]
  diffFilter = delta --color-only
[delta]
  navigate = true
```

LazyGit gets a delta pager with paging disabled and line numbers enabled.

## Why

Omarchy already ships a curated terminal/dev setup. Delta makes Git diffs easier to scan with syntax highlighting while keeping the default config small and standard.

The migration is guarded: it only fills unset Git keys and only copies the LazyGit config if that file is empty.

## Testing

- `git config --file config/git/config --get core.pager`
- `git config --file config/git/config --get interactive.diffFilter`
- `lazygit --use-config-file config/lazygit/config.yml --version`
- smoke-tested the migration in a temporary HOME
- `bash -n migrations/1780739890.sh`
- `git diff --check origin/dev..cp/git-delta-defaults`

### DIFF
--- a/config/git/config
+++ b/config/git/config
@@ -11,10 +11,16 @@
 	rebase = true 			 # Rebase (instead of merge) on pull
 [push]
 	autoSetupRemote = true   # Automatically set upstream branch on push
+[core]
+	pager = delta            # Use syntax-highlighted diffs
 [diff]
 	algorithm = histogram    # Clearer diffs on moved/edited lines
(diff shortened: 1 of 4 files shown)


dhh:
I gave this a shot at one point and didn't actually like it. We also have Lazygit + hunk already for when you need a more intricate diff.

PR #6051, opened 2026-06-06
Use Inter as the default UI font

crmne:
## Summary

Install `inter-font` and prefer Inter for Omarchy's sans/system UI fontconfig defaults, keeping Liberation Sans as the fallback.

This updates `sans-serif`, `system-ui`, `-apple-system`, and `BlinkMacSystemFont` to prefer Inter first.

## Why

Inter is designed for user interfaces and is a better default UI font for the desktop. This only changes proportional UI/sans defaults; terminal monospace fonts remain unchanged.

The migration installs `inter-font` and inserts Inter before existing `Liberation Sans` entries in the user's fontconfig file, leaving the fallback intact.

## Testing

- `xmlstarlet val config/fontconfig/fonts.conf`
- smoke-tested the migration in a temporary HOME
- `bash -n migrations/1780739891.sh`
- `git diff --check`

### DIFF
--- a/config/fontconfig/fonts.conf
+++ b/config/fontconfig/fonts.conf
@@ -6,6 +6,7 @@
       <string>sans-serif</string>
     </test>
     <edit name="family" mode="assign" binding="strong">
+      <string>Inter</string>
       <string>Liberation Sans</string>
     </edit>
   </match>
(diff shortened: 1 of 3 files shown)


dhh:
Can you show some before/after screenshots?

dhh:
Actually, tried this out, and didn't love the new font or find it worth adding an additional dependency for it. Thanks for sharing, though!

PR #6057, opened 2026-06-06
Surface xdg-open errors from open helper

### LINKED ISSUE #6056: Surface xdg-open errors from the open shell helper
## Problem

Omarchy defines the interactive shell helper `open` as a detached `xdg-open` wrapper:

```bash
open() (
  xdg-open "$@" >/dev/null 2>&1 &
)
```

That keeps the terminal clean, but it also discards stderr. When `xdg-open` cannot handle a target, users running `open ...` directly in the terminal get no error output and the failure looks silent.

## Expected behavior

Keep the existing non-blocking behavior and keep stdout quiet, but allow `xdg-open` stderr to reach the terminal so actionable failures are visible.

## Suggested change

Drop only the stderr redirection:

```bash
open() (
  xdg-open "$@" >/dev/null &
)
```

This preserves the previous behavior that motivated #4366: the command remains detached and avoids job-control completion noise, while making launch failures visible to the user.

shbernal:
## Summary

- let the `open` shell helper surface `xdg-open` stderr in the terminal
- keep stdout redirected to `/dev/null` so command output stays quiet
- keep `xdg-open` backgrounded in the subshell to preserve the non-blocking behavior and avoid job-control noise

Closes #6056

## Testing

- `bash -n default/bash/aliases`
- `git diff --check`
- focused shell test with a stubbed `xdg-open` confirmed stderr is visible while stdout stays out of pipelines

### DIFF
--- a/default/bash/aliases
+++ b/default/bash/aliases
@@ -34,7 +34,7 @@ if command -v zoxide &> /dev/null; then
 fi
 
 open() (
-  xdg-open "$@" >/dev/null 2>&1 &
+  xdg-open "$@" >/dev/null &
 )
 
 # Directories


dhh:
Closing as an exact duplicate of #6198, which remains open.

PR #6058, opened 2026-06-07
Theme Zen Browser chrome

AndrijaSkontra:
## Summary
- add an Omarchy theme sync command for Zen Browser chrome
- import generated Omarchy CSS into Zen profiles without overwriting existing userChrome.css
- explicitly theme Zen vertical tabs/sidebar selectors and restart Zen when it is running

## Problem
Zen Browser is installable through Omarchy, but Omarchy theme changes do not update Zen chrome. In particular, the vertical tabs/sidebar area can keep the old color after switching themes because it is not covered by the Chromium browser policy path and needs Zen/Firefox userChrome styling.

## Verification
- bash -n bin/omarchy-theme-set-zen bin/omarchy-theme-set
- git diff --check origin/dev...HEAD
- manually verified the same selector set fixes Zen vertical tab/sidebar colors on an Omarchy install


aldeni:
Heads up: this stopped running somewhere along the way to quattro. Both guards at the top exit early on a current install, so the script is a silent no-op rather than a failure.

```bash
COLORS_TOML="$HOME/.config/omarchy/current/theme/colors.toml"
ZEN_PROFILES_DIR="$HOME/.zen"

[[ -d $ZEN_PROFILES_DIR ]] || exit 0
[[ -f $COLORS_TOML ]] || exit 0
```

On quattro with `zen-browser-bin 1.21.14b`:

- theme state moved to `$HOME/.local/state/omarchy/current/theme/` (`colors.toml` and `theme.name`)
- Zen keeps profiles in `$HOME/.config/zen`, not `$HOME/.zen`

One other thing you may want: `omarchy-theme-color accent` (and `--all`) is the shared colors.toml parser, and it resolves the alias/fallback cascade that `parse_color` here would miss on themes that only define the legacy names.

Separately, in case it is useful — Zen exposes `zen.theme.accent-color` as a preference, applied to the chrome as `--zen-primary-color` in `zenThemeModifier.js`, and it takes a hex value directly. I opened #7542 using it. It is narrower than this one: it stops at the accent and never reaches the sidebar background you handle here, but it covers those surfaces without touching chrome selectors. If the maintainers would rather have a single script, folding it in here seems reasonable to me.

spuder:
Related https://github.com/omacom/omarchy/pull/8913

PR #6062, opened 2026-06-09
Avoid double-running child hook directories

khushveer007:
## Summary

Fixes hook dispatch so `omarchy-hook` does not execute `<hook>.d/*` entries after a parent hook has already run. This preserves parent-owned child dispatch behavior while retaining fallback execution for child-only hook directories.

## What Changed

- Changed `bin/omarchy-hook` to make parent hook execution and `.d` fallback dispatch mutually exclusive.
- Added regression coverage for parent-owned child dispatch and child-only `.d` hook execution in `test/omarchy-cli-test.sh`.

## How it was Tested

- Ran `test/omarchy-cli-test.sh`.
- Ran `bin/omarchy commands --check`.
- Verified the PR diff contains only `bin/omarchy-hook` and `test/omarchy-cli-test.sh`.

### DIFF
--- a/bin/omarchy-hook
+++ b/bin/omarchy-hook
@@ -17,9 +17,7 @@ shift
 
 if [[ -f $HOOK_PATH ]]; then
   bash "$HOOK_PATH" "$@" || echo "Hook failed: $HOOK_PATH"
-fi
-
-if [[ -d $HOOK_DIR ]]; then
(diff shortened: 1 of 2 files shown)


dhh:
Closing in favor of the newer hook double-run fix in #6197.

PR #6066, opened 2026-06-10
Theme set jetbrains

guidonaselli:


dhh:
Not adding 4K of code for jetbrains that we don't have as a default. Also, no-description PRs are very unlikely to proceed.

PR #6067, opened 2026-06-10
Add --browser flag to launch-webapp

ianmihura:
Adds `--browser` flag to `omarchy launch webapp`, allowing users to specify which browser opens the web app instead of always using the system default.

Supports `--browser <name>` and `--browser=<name>`. Falls back to `xdg-settings get default-web-browser` when omitted — no change in default behavior.

Usage:

    omarchy launch webapp --browser brave https://example.com


ianmihura:
Wrong base branch, recreating against dev.

PR #6070, opened 2026-06-11
Adding Vivaldi as a default browser candidate.

dbech:
I want my browser to feel at home in the OS as I do.

I assumed the intention was to order these applications alphabetically so I added Vivaldi above Zen.

I assumed users would install vivaldi via pacman and as such that is the install I checked the desktop ID for.


dbech:
mmm I cooked that a bit cleaning up my github. I missed stuff anyway I'll make a fresh PR later.

PR #6071, opened 2026-06-11
Add Vivaldi as a default browser candidate.

dbech:
I like Vivaldi I think.

I have added it to the default browser commands and menus. 

I did not add it to the install/uninstall browser tools as I just installed it with Pacman -S nothing special and so far that seems fine. 

If this PR is otherwise acceptable but you would like other support added for Vivaldi before merging I presently have time and am willing to do the work as far as I am capable.

### DIFF
--- a/bin/omarchy-default-browser
+++ b/bin/omarchy-default-browser
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 # omarchy:summary=Set the default browser for Omarchy and XDG handlers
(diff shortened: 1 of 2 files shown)


brink-lab:
Great work 🏆

Looks like it might already be covered (including install and dynamic theme switching) in https://github.com/basecamp/omarchy/pull/5774

dbech:
oh yea nice, didn't spot that one. Yea I reckon that's go it.

PR #6078, opened 2026-06-12
Add --browser flag to omarchy-launch-webapp

ianmihura:
Adds `--browser` flag to `omarchy launch webapp`, allowing users to specify which browser opens the web app instead of always using the system default.

Supports `--browser <name>` and `--browser=<name>`. Falls back to `xdg-settings get default-web-browser` when omitted — no change in default behavior.

Usage:

    omarchy launch webapp --browser brave https://example.com

### DIFF
--- a/bin/omarchy-launch-webapp
+++ b/bin/omarchy-launch-webapp
@@ -1,9 +1,29 @@
 #!/bin/bash
 
-# omarchy:summary=Launch a URL as a web app in the default supported browser
-# omarchy:args=<url>
+# omarchy:summary=Launch a URL as a web app in the default supported browser, or user specified browser
+# omarchy:args=[--browser <name>] <url>
 
-browser=$(xdg-settings get default-web-browser)
+browser=""
+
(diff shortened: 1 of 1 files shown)


Linar46:
Also could you add option to open Apps in Browser, tab not in separate window.

ianmihura:
@Linar46 what's the use case? `omarchy launch webapp` is meant to simulate native apps. There are no browser tabs in native apps.

Linar46:
> @Linar46 what's the use case? `omarchy launch webapp` is meant to simulate native apps. There are no browser tabs in native apps.


- I use the Omarchy search for easy lookup and launch. I.e. everything.
It would be a lot of help for me when i need for example figma in one tab or translator when i am reading or devdocs.

Linar46:
@dhh how does that sound for you?
Looking for all things in one search bar? Isn't that sound Great?

ianmihura:
@Linar46
Sounds interesting, but I think it's a different feature and use case from this one

PR #6090, opened 2026-06-14
Add migration to backfill Mesa Vulkan driver for pre-vulkan.sh installs

### LINKED ISSUE #6089: Vulkan driver not backfilled: pre-vulkan.sh installs left without Mesa Vulkan ICD
### System details

Intel UHD 630 (CoffeeLake-H) + NVIDIA GTX 1650 Mobile (Optimus laptop), Omarchy 3.8.2

### What's wrong?

Systems installed **before `install/config/hardware/vulkan.sh` was added** (commit `694eb8c6`, "Install vulkan drivers by default", 2026-02-20) never get the Mesa Vulkan driver for their iGPU, and **no migration backfills it**. So Intel/AMD machines installed earlier are left with no Mesa Vulkan ICD.

On this Optimus laptop the effect was concrete and user-visible: NVIDIA Vulkan ships via `nvidia-utils` (installed at setup by `nvidia.sh`), so the **only** Vulkan device present was the NVIDIA dGPU. mpv (default `vo=gpu-next`/Vulkan) therefore rendered every frame on the discrete GPU and cross-presented to the Intel-driven compositor, which stalled — mpv repeatedly hit "Application Not Responding". Installing `vulkan-intel` made the iGPU a Vulkan device again and resolved it.

**Root cause:** `install/` scripts only run once at initial setup. `vulkan.sh` was introduced as an install-time step but no corresponding `migrations/*.sh` was shipped, so it never reaches pre-existing systems.

**Verification on this machine:**
- Installed `2026-02-18` (`pacman.log` first transaction).
- `vulkan.sh` added `2026-02-20` — two days after install, so it never ran here.
- `pacman.log` shows `vulkan-intel` was never installed until I did so manually on `2026-06-14`.
- `intel-media-driver` (from `intel/video-acceleration.sh`) and `nvidia-open-dkms` (from `nvidia.sh`) are present — so the other hardware scripts ran; only the Vulkan driver fell through the gap.
- `grep -rl "vulkan-intel\|vulkan-radeon\|VULKAN_DRIVERS" migrations/` → no matches.

### Suggested fix

A one-line migration that re-applies the `vulkan.sh` detection (it uses `omarchy-pkg-add`, which is idempotent, so it's a no-op on systems that already have the driver). Happy to open a PR — see below.

ki11e6:
## What

Adds a migration that installs the Mesa Vulkan driver (`vulkan-intel` / `vulkan-radeon` / `vulkan-asahi`) on systems that were set up **before** `install/config/hardware/vulkan.sh` was added (commit 694eb8c6, 2026-02-20).

Fixes #6089.

## Why

`install/` scripts only run once, at initial setup. `vulkan.sh` was introduced as an install-time step but no migration was shipped, so it never reaches pre-existing machines. Intel/AMD systems installed before that date have no Mesa Vulkan ICD to this day.

On hybrid (Optimus) laptops the impact is user-visible: NVIDIA Vulkan ships via `nvidia-utils`, so the discrete GPU becomes the *only* Vulkan device. `mpv` (default `vo=gpu-next`/Vulkan) then renders on the dGPU and cross-presents to the iGPU-driven compositor, which stalls ("Application Not Responding"). Installing `vulkan-intel` restored the iGPU as a Vulkan device and fixed it.

## How

The migration mirrors the detection in `vulkan.sh` and relies on `omarchy-pkg-add`, which is idempotent (`pacman -S --needed`) — so it is a **no-op** on systems that already have the driver (e.g. fresh installs). The migration timestamp (`1780929365`) is derived from the current `dev` HEAD commit, per `omarchy-dev-add-migration` convention, and sorts after the latest existing migration.

## Testing

Verified on an affected machine (Omarchy 3.8.2, Intel UHD 630 + NVIDIA GTX 1650, installed 2026-02-18): `vulkan-intel` was absent, the dGPU was the only Vulkan device, and mpv froze. After installing `vulkan-intel`, both GPUs appear as Vulkan devices and mpv renders on Intel without stalling.

### DIFF
--- a/migrations/1780929365.sh
+++ b/migrations/1780929365.sh
@@ -1,0 +1,25 @@
+echo "Install missing Mesa Vulkan driver for Intel/AMD/Apple GPUs"
+
+# Backfill for systems installed before install/config/hardware/vulkan.sh existed.
+# install/ scripts only run at initial setup, so machines set up earlier never
+# received a Mesa Vulkan ICD for their iGPU (leaving e.g. NVIDIA as the only Vulkan
+# device on hybrid laptops). omarchy-pkg-add is idempotent, so this is a no-op when
+# the driver is already present.
+
+declare -A VULKAN_DRIVERS=(
+  [Intel]=vulkan-intel
+  [AMD]=vulkan-radeon
+  [Apple]=vulkan-asahi
+)
+
+PACKAGES=()
+
+for vendor in "${!VULKAN_DRIVERS[@]}"; do
+  if lspci | grep -iE "(VGA|Display).*$vendor" >/dev/null; then
+    PACKAGES+=("${VULKAN_DRIVERS[$vendor]}")
+  fi
+done
+
+if (( ${#PACKAGES[@]} > 0 )); then
+  omarchy-pkg-add "${PACKAGES[@]}"
+fi
--- a/migrations/1780929365.sh
+++ b/migrations/1780929365.sh
@@ -1,4 +1,4 @@
-echo "Install missing Mesa Vulkan driver for Intel/AMD/Apple GPUs"
+echo "Install missing Mesa Vulkan driver for Intel/AMD GPUs"
 
 # Backfill for systems installed before install/config/hardware/vulkan.sh existed.
 # install/ scripts only run at initial setup, so machines set up earlier never
@@ -9,7 +9,6 @@ echo "Install missing Mesa Vulkan driver for Intel/AMD/Apple GPUs"
 declare -A VULKAN_DRIVERS=(
   [Intel]=vulkan-intel
   [AMD]=vulkan-radeon
-  [Apple]=vulkan-asahi
 )
 
 PACKAGES=()


ki11e6:
Thanks for the review. Addressing the automated feedback:

- **Apple/Asahi via `lspci`** — good catch, and correct: Apple Silicon GPUs aren't PCI devices so that branch never fires. It was inherited verbatim from `install/config/hardware/vulkan.sh`. Since a backfill should only do what reliably works, I've **scoped the migration to Intel/AMD** and dropped the Apple entry rather than add Asahi platform detection (out of scope for this fix). The same `lspci` limitation in `vulkan.sh` itself is probably worth a separate look.
- **Shebang / bashisms** — intentional, matches repo convention: every script in `migrations/` is shebang-less and `bin/omarchy-migrate` runs them via `bash "$file"`, so `declare -A`/arrays are safe. `vulkan.sh` uses the same construct.
- **Leading indentation on line 1** — the committed file has `echo` at column 0 (no leading whitespace); appears to be a false positive.

dhh:
Closing in favor of the expanded Vulkan backfill implementation in #6199.

PR #6092, opened 2026-06-14
make youtube and zoom webapps opaque on any browser and profile

Zeus-Deus:
the video opacity opt out in browser.conf only matches chrome on the default profile, so brave, chromium, edge and anyone on a non default profile still get the default 0.97/0.9 transparency on their youtube and zoom webapp windows. annoying when watching video.

this widens the class match to any browser prefix and any profile. same mechanism and same two tags as before, just a broader regex.

tested live on brave (class brave-youtube.com__-Profile_2): the default-opacity tag now gets removed and the window stays fully opaque. a normal browser tab on youtube keeps its plain browser class, so regular browsing is unaffected.

### DIFF
--- a/default/hypr/apps/browser.conf
+++ b/default/hypr/apps/browser.conf
@@ -5,8 +5,8 @@ windowrule = tag -default-opacity, match:tag chromium-based-browser
 windowrule = tag -default-opacity, match:tag firefox-based-browser
 
 # Video apps: remove chromium browser tag so they don't get opacity applied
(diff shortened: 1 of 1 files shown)


dhh:
Thanks! This landed on `quattro` (the current development branch, where this config now lives in `default/hypr/apps/browser.lua`) as c442ad7d936c0462efac4f7fd73a3c2e0bff8e74, with co-author credit. Tested live across Chromium/Brave/Edge webapp classes, default and non-default profiles. Closing since it's ported rather than merged from this branch.

PR #6093, opened 2026-06-14
Fix ASUS ExpertBook B9406 touchpad quirk never being applied

mijuny:
Follow-up to #5435. On a fresh 3.8.2 install of an ASUS ExpertBook B9406 the touchpad is still dead: clicks register but the cursor never moves. The quirk from #5435 never took effect, for two reasons:

1. It was written to `/etc/libinput/asus-expertbook-b9406.quirks`. libinput reads only `/etc/libinput/local-overrides.quirks` and does not scan the directory, so the file was ignored.

2. The section required `MatchUdevType=touchpad`, but udev tags this pad as `ID_INPUT_MOUSE`. The bus/vendor/product/DMI keys already pin the device, so the type constraint is dropped.

How this slipped through: the original fix was verified as a manual local edit, but the packaged install path was not tested end to end, so neither defect was caught. I have since corrected that in my workflow. Both changes here were confirmed working by the reporter. Adds a migration so existing B9406 installs pick up the corrected quirk.

This restores cursor motion. A wider correction (proper touchpad classification for full gestures) will follow once I have access to a working ExpertBook again. Mine is currently physically broken (cracked screen).

AI disclaimer: diagnosed and fixed with help from Claude Code (Opus 4.8).

### DIFF
--- a/install/config/hardware/asus/fix-asus-ptl-b9406-touchpad.sh
+++ b/install/config/hardware/asus/fix-asus-ptl-b9406-touchpad.sh
@@ -8,16 +8,32 @@
 #
 # Mask the pressure axes with a quirks override, same pattern as the
 # Asus UX302LA entry in libinput's shipped 50-system-asus.quirks.
+#
+# Two things matter for the override to take effect:
+#   1. libinput reads only ONE override file, /etc/libinput/local-overrides.quirks.
+#      Custom-named files in /etc/libinput/ are never loaded.
+#   2. udev tags this pad as ID_INPUT_MOUSE, not ID_INPUT_TOUCHPAD, so a
+#      MatchUdevType=touchpad section is skipped. The bus/vendor/product/DMI
+#      keys already pin the device exactly, so we omit the type constraint.
 
 if omarchy-hw-asus-expertbook-b9406; then
-  sudo mkdir -p /etc/libinput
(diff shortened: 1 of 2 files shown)


dhh:
Thanks @mijuny — the diagnosis here is right, and I verified it independently: libinput really does only read `/etc/libinput/local-overrides.quirks` and never scans `/etc`, so the file from #5435 was dead on arrival.

Superseded by #6388, which targets `quattro` (this branch was cut against `dev`) and carries your commits' authorship.

Two changes there worth knowing about:

- **Drop-in instead of appending to `local-overrides.quirks`.** libinput `scandir`s `/usr/share/libinput` for any `*.quirks` file, so we can own a file outright and rewrite it wholesale. Appending to the shared override file means idempotency has to grep for a section we already wrote, which silently pins the first version we ever ship — if the quirk ever needs correcting, the fix never lands.
- **Quattro porting.** `install/hardware/*` runs as root there, so the `sudo` calls come out; migrations run as the user, so the migration escalates explicitly rather than sourcing the leaf; and the migration timestamp needed renumbering above quattro's newest.

Kept your `MatchUdevType` removal, though I flagged in the new PR that the evidence for it is weaker than for the path fix — both changes shipped in the same test, so the reporter's confirmation can't separate them. It costs nothing if unnecessary.

Hope the ExpertBook gets a new screen soon — a confirmation on live hardware before #6388 merges would still be valuable.

PR #6105, opened 2026-06-17
Make webapps profile-aware for Chromium-based browsers

AndrewGaspar:
Webapps launched with `--app=` used whichever profile last had focus. On a machine with several browser profiles that is confusing: the web app opens logged out, or in the wrong account. Now, when the effective browser has more than one profile, the launcher offers a choice and can remember it per webapp; installs can pre-select a profile; and selections are dropped when the default browser changes, since profiles belong to the browser. With a single profile nothing is asked, so most users see no change.

- `omarchy-webapp-profiles`: lists the profiles (`dirname<TAB>name`) of the browser webapps actually launch in, read from Chromium's `Local State`; it mirrors the browser choice `omarchy-launch-webapp` makes, so a Firefox default lists Chromium's profiles
- `omarchy-webapp-profile-store` (hidden helper): get/set/clear-all of remembered selections, keyed by a sha256 of the webapp URL and stamped with the browser they were made for
- `omarchy-launch-webapp`: with several profiles, asks via `omarchy-menu-select` (each profile once, or "Always use …" to remember), then passes `--profile-directory=`
- `omarchy-webapp-install`: offers the profile up front in the interactive flow
- `omarchy-default-browser`: clears remembered selections when the default browser changes
- HEY and Zoom handlers pin a stable remember key across their dynamic URLs
- `test/shell.d/webapp-profiles-test.sh` covers listing, the store, every launch path (single profile, pick, remember, dismiss, stale selection, handler key) and the browser switch
- `manual/25-web-apps.md` mentions it

Selections live outside `.desktop` files, so existing webapps need no rewrite.

Demo of all major UX flows:
- Launching existing Web App (logged in profile, unused profile, and remembering profile for future launches)
- Installing a new web app
- Changing browser default clears the profile settings
- Launching a webapp when there is only one profile bypasses the dialog (default behavior for most users)

https://1drv.ms/v/c/ec3aff9eb5567ab6/IQBill3ZNx_rS5sVvQG3KyI_Ae5KPSx_rlXSFAnthPA0ooA?e=0lkwVG

Rebased onto `quattro` on 2026-09-13 and retargeted from `dev`. Selections moved from `~/.config/omarchy/webapps` to `~/.local/state/omarchy/webapps`, per the file layout's config/state split. The Brave Origin profile directory is `BraveSoftware/Brave-Origin`, as elsewhere in the repo. The notice-only migration is dropped since nothing needs backfilling, and `omarchy-default-browser` keeps `quattro`'s install-on-demand flow with the clear added after the switch.

https://claude.ai/code/session_0143YiDa9a7qgYDJWD5boZVX


AndrewGaspar:
Looks like #1266 is related, though that focuses on making it possible to have webapps per-profile, which also seems like a worthy goal, but my main problem currently is just getting the web app to launch in the profile that's already logged into the service I'm wanting to use, and less so needing to use the same installed web app on multiple profiles.

AndrewGaspar:
Rebased onto `quattro` and retargeted from `dev`. Selections now live under `~/.local/state/omarchy` per the config/state split, the notice-only migration is gone, and `test/shell.d/webapp-profiles-test.sh` covers the listing, store, launch paths and browser switch; the description's last paragraph has the details.

PR #6116, opened 2026-06-21
Fix toggle load order so user overrides take precedence

### LINKED ISSUE #6068: Omarchy toggle load order makes user looknfeel.conf unable to override toggle settings.
### System details

Omarchy 3.8.2. Hyprland 0.55.2 (Wayland). ghostty 1.3.1-arch2

### What's wrong?

I went to open a single window on my desktop and its shunk now because I have a wide monitor. Not a bad design decision, just not for me so I went to over-ride it in `~/.conf/hypr/looknfeel.conf` (by adding `single_window_aspect_ratio = 0 0` to the layout section) but it gets overridden by `~/.local/state/omarchy/toggles/hypr/*.conf` based on the load order in `~/.config/hypr/hyprland.conf`

Once I put my looknfeel.conf again at the bottom of the hyprland.conf it worked as expected with a new window filling the screen. Maybe the toggles should be after the defaults but before the customs?

```~ ❯ cat ~/.config/hypr/hyprland.conf
# Learn how to configure Hyprland: https://wiki.hyprland.org/Configuring/

# Use defaults Omarchy defaults (but don't edit these directly!)
source = ~/.local/share/omarchy/default/hypr/autostart.conf
source = ~/.local/share/omarchy/default/hypr/bindings/media.conf
source = ~/.local/share/omarchy/default/hypr/bindings/clipboard.conf
source = ~/.local/share/omarchy/default/hypr/bindings/tiling-v2.conf
source = ~/.local/share/omarchy/default/hypr/bindings/utilities.conf
source = ~/.local/share/omarchy/default/hypr/envs.conf
source = ~/.local/share/omarchy/default/hypr/looknfeel.conf
source = ~/.local/share/omarchy/default/hypr/input.conf
source = ~/.local/share/omarchy/default/hypr/windows.conf
source = ~/.config/omarchy/current/theme/hyprland.conf

# Change your own setup in these files (and overwrite any settings from defaults!)
source = ~/.config/hypr/monitors.conf
source = ~/.config/hypr/input.conf
source = ~/.config/hypr/bindings.conf
source = ~/.config/hypr/envs.conf
source = ~/.config/hypr/looknfeel.conf
source = ~/.config/hypr/autostart.conf

# NVIDIA environment variables
env = NVD_BACKEND,direct
env = LIBVA_DRIVER_NAME,nvidia
env = __GLX_VENDOR_LIBRARY_NAME,nvidia```

# Toggle config flags dynamically
source = ~/.local/state/omarchy/toggles/hypr/*.conf

# 2026-06-10 mal toggles overwriting my looknfeel; reloading
source = ~/.config/hypr/looknfeel.conf

axelfrache:
## Summary

- Move `require("default.hypr.toggles")` before user `require()` calls in the `hyprland.lua` template
- Add a migration that reorders existing `hyprland.lua` and `hyprland.conf` installs automatically

Toggles were loaded after user customizations, so any toggle would silently override settings in `~/.config/hypr/looknfeel.conf`. The correct order is: Omarchy defaults → toggles → user config.

Fixes #6068

### DIFF
--- a/config/hypr/hyprland.lua
+++ b/config/hypr/hyprland.lua
@@ -10,15 +10,15 @@ package.path = os.getenv("HOME")
 -- All Omarchy default setups
 require("default.hypr.omarchy")
 
+-- Toggle config flags dynamically (before user overrides so users can override toggle settings).
+require("default.hypr.toggles")
+
 -- Change your own setup in these files and override defaults.
 require("hypr.monitors")
 require("hypr.input")
 require("hypr.bindings")
 require("hypr.looknfeel")
 require("hypr.autostart")
-
--- Toggle config flags dynamically.
-require("default.hypr.toggles")
 
 -- Add any other personal Hyprland configuration below.
 -- o.window("qemu", { workspace = "5" })
--- a/migrations/1780929365.sh
+++ b/migrations/1780929365.sh
@@ -4,10 +4,11 @@ HYPR_LUA=~/.config/hypr/hyprland.lua
 HYPR_CONF=~/.config/hypr/hyprland.conf
 
 if [[ -f $HYPR_LUA ]] && grep -q 'require("default.hypr.toggles")' "$HYPR_LUA"; then
-  toggle_line=$(grep -n 'require("default.hypr.toggles")' "$HYPR_LUA" | cut -d: -f1)
-  looknfeel_line=$(grep -n 'require("hypr.looknfeel")' "$HYPR_LUA" | cut -d: -f1)
+  toggle_line=$(grep -n 'require("default.hypr.toggles")' "$HYPR_LUA" | cut -d: -f1 | tail -n1)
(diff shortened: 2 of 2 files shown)


dhh:
It seems correct to me that toggles would override user settings. That's the entire point. Otherwise we can't guarantee that, say, no borders actually does what we promise.

xicubed:
So you're saying if you toggle the borders with some hotkey combo it should work with a guarantee? That makes sense. In my original bug #6068 I was trying to change the single_window_aspect_ratio back to taking up all the screen instead of shrunken down on wide monitors. I guess there's an omarchy-hyprland-window-single-square-aspect-toggle... maybe I could set a keybind for that or store my preferences in another flat file and reference that so it's sticky between reboots...

xicubed:
Updated to Quattro - everything is sweet now. Thx.

PR #6121, opened 2026-06-22
Fix fingerprint setup crash when libfprint-git is installed

mounis-bhat:
## Issue
The `omarchy-pkg-present libfprint` check returns `true` if `libfprint-git` is installed because it "provides" `libfprint`. 

The script then tries to remove the literal `libfprint` package with `pacman -Rdd`. This triggers a `sudo` prompt for existing fingerprint users, fails with a `target not found` error, and abruptly crashes the script due to `set -e`.

---

## Fix
Replaced the check with `pacman -Qq | grep -Fxq "libfprint"` to ensure it only attempts removal if the exact, literal `libfprint` package is installed.

### DIFF
--- a/bin/omarchy-setup-security-fingerprint
+++ b/bin/omarchy-setup-security-fingerprint
@@ -5,7 +5,6 @@
 
 set -e
 
-
 check_fingerprint_hardware() {
   # Get fingerprint devices for the user
   devices=$(fprintd-list "$USER" 2>/dev/null)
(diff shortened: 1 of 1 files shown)


dhh:
Thanks for this — the same crash is already fixed on the upcoming release branch in ae7365f937e7caf57783e4631de1e47f293f8804 ("Fix repeated fingerprint setup"), which guards removal on an exact match of the installed `libfprint` package rather than a provider like `libfprint-git`. Closing as already resolved.

PR #6127, opened 2026-06-23
Unconditionally turn off internal display when lid is closed

AlJohri:
## Summary

Right now we only turn off the internal display when an external display is connected. I currently use my FW laptop as a server (lid closed) without any connected display. I noticed the wattage on my laptop was higher than expected and realized that the lid display was still on.

I believe we should unconditionally turn off the internal display when the lid is closed.

The existing "only active display" guard is preserved for the lid-open case (where blanking would orphan the user). Desktops without `/proc/acpi/button/lid/*/state` are treated as lid-open, so their behavior is unchanged.

### DIFF
--- a/bin/omarchy-hw-lid-closed
+++ b/bin/omarchy-hw-lid-closed
@@ -1,0 +1,11 @@
+#!/bin/bash
+
+# omarchy:summary=Returns true when the laptop lid is closed
+# omarchy:hidden=true
+
+for state in /proc/acpi/button/lid/*/state; do
+  [[ -r $state ]] || continue
+  [[ $(< "$state") == *"closed"* ]] && exit 0
+done
+
+exit 1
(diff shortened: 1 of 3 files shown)


dhh:
By default, closing the lid puts the machine to sleep. How are you getting it to a state where lid close doesn't do anything when no display is connected?

AlJohri:
I use [setup-lid-suspend.sh](https://github.com/AlJohri/dotfiles/blob/9eea6cc7e180ba75847eca4dca48d256304ff125/scripts/bin/setup-lid-suspend.sh) which adds a new file called `/etc/systemd/logind.conf.d/lid.conf` with the following settings:

```
[Login]
HandleLidSwitchExternalPower=ignore
HandleLidSwitchDocked=ignore
```

I use this on my FW16 to run it as a headless server. It keeps the laptop running with the lid closed as long as external power is connected.

dhh:
Gotcha. I think then it makes more sense that this use case belongs with that and not something we have out of the box here. Would be good to write this up in Discussions 👍

PR #6133, opened 2026-06-24
Fix inherited text shadows in Waybar tooltips

pipetogrep:
## Summary

- disable inherited text shadows in the base Waybar CSS
- add a guarded migration for existing copied Waybar configs

## Why

On some themes, GTK text-shadow styling can leak into Waybar tooltips and make dropdown text look doubled. Omarchy Waybar themes only define colors, so the base Waybar CSS should explicitly opt out of text shadows.

## Validation

- `git diff --check`
- `bash -n migrations/1779713467.sh`
- tested the migration against a temporary HOME with a copied Waybar config

### Before
image

### After
image

### DIFF
--- a/config/waybar/style.css
+++ b/config/waybar/style.css
@@ -9,6 +9,7 @@
   min-height: 0;
   font-family: 'JetBrainsMono Nerd Font';
   font-size: 12px;
+  /* omarchy:disable-text-shadow */
   text-shadow: none;
 }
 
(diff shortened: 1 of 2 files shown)


brink-lab:
A kind heads-up :) 

The upcoming Omarchy 4 release moves away from Waybar in favour of a new quickshell based solution.
Therefore, this is a topic we should check for the new shell.

dhh:
Closing because Quattro removed Waybar and its stylesheet. Thanks for the fix.

PR #6139, opened 2026-06-25
Use Xe-capable btop on Intel xe GPU systems

stappmus:
## Summary

Detect Intel GPUs using the Linux `xe` driver and install the temporary `btop-xe` package from OPR so btop can show GPU usage/frequency on Lunar Lake, Panther Lake, Battlemage, and future Xe-driver systems.

Also turns on btop GPU info inside the CPU panel by default, while keeping the separate GPU box disabled.

Depends on: https://github.com/omacom-io/omarchy-pkgs/pull/114

## Why

Omarchy already installs `btop`, but current Arch `extra/btop` does not detect Intel GPUs using the `xe` kernel driver. On Panther Lake, btop shows CPU/memory/process stats but no Intel GPU usage/frequency even though the kernel exposes Xe telemetry.

This is not Panther Lake-only. The detector checks for Intel PCI GPU devices using `DRIVER=xe` and an exposed `xe_*` PMU source.

Upstream context:

- https://github.com/aristocratos/btop/issues/1407
- https://github.com/aristocratos/btop/pull/1457

## What changes

- Add `omarchy-hw-intel-xe`
- Add `install/config/hardware/intel/btop-xe.sh`
- Run that install hook from `install/config/all.sh`
- Include `btop-xe` in `install/omarchy-other.packages` for ISO/package availability
- Update default btop config:
  - `show_gpu_info = "On"`
  - keep `shown_boxes = "cpu mem net proc"`
  - keep `shown_gpus = "nvidia amd intel"`
- Add a migration for existing Intel Xe installs

## Compatibility

This does not disable or hide other GPU vendors.

`shown_gpus` remains:

```ini
shown_gpus = "nvidia amd intel"
```

The package still builds btop with normal GPU support. NVIDIA and AMD users are not routed through this hook because it only runs when an Intel GPU is using the `xe` driver.

## Screenshot

Intended layout: no separate GPU box, with Intel Xe GPU usage shown inside the CPU panel.

btop showing Intel Xe GPU usage inside the CPU panel

## Tested

On Intel Panther Lake / Arc B390:

```text
00:02.0 VGA compatible controller: Intel Corporation Panther Lake [Arc B390]
Kernel driver in use: xe
/sys/bus/event_source/devices/xe_0000_00_02.0 exists
```

The new detector exits successfully on that hardware:

```text
bin/omarchy-hw-intel-xe
# exit 0
```

After the package/config change:

```text
btop version: 1.4.7+76530c8
/usr/bin/btop cap_dac_read_search,cap_perfmon=ep
```

`btop -d` reports:

```text
Xe: Using fdinfo for GPU utilization (PCI: 0000:00:02.0)
```

GPU usage and frequency appear inside the CPU panel.

Additional checks:

```text
bash -n bin/omarchy-hw-intel-xe install/config/hardware/intel/btop-xe.sh migrations/1782416832.sh
git diff --check
```

## Notes

Temporary until upstream btop and Arch `extra/btop` ship Intel `xe` support.

AI disclosure: this PR was prepared with AI assistance and manually tested on real Panther Lake hardware.

### DIFF
--- a/bin/omarchy-hw-intel-xe
+++ b/bin/omarchy-hw-intel-xe
@@ -1,0 +1,12 @@
+#!/bin/bash
+
+# Detect Intel GPUs using the Linux xe driver.
+
+for uevent in /sys/class/drm/card[0-9]*/device/uevent; do
+  [[ -r "$uevent" ]] || continue
+  grep -q '^DRIVER=xe$' "$uevent" || continue
+  grep -q '^PCI_ID=8086:' "$uevent" || continue
+  compgen -G '/sys/bus/event_source/devices/xe_*' >/dev/null && exit 0
+done
+
+exit 1
--- a/config/btop/btop.conf
+++ b/config/btop/btop.conf
@@ -107,7 +107,7 @@ cpu_graph_upper = "Auto"
 cpu_graph_lower = "Auto"
 
 #* If gpu info should be shown in the cpu box. Available values = "Auto", "On" and "Off".
-show_gpu_info = "Auto"
+show_gpu_info = "On"
 
 #* Toggles if the lower CPU graph should be inverted.
 cpu_invert_lower = true
--- a/install/config/all.sh
+++ b/install/config/all.sh
@@ -43,6 +43,7 @@ run_logged $OMARCHY_INSTALL/config/hardware/nvidia.sh
 run_logged $OMARCHY_INSTALL/config/hardware/vulkan.sh
 
 run_logged $OMARCHY_INSTALL/config/hardware/intel/video-acceleration.sh
+run_logged $OMARCHY_INSTALL/config/hardware/intel/btop-xe.sh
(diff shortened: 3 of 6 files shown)


stappmus:
Updated the companion package PR to reduce the Xe polling overhead: `btop-xe` now prefers kernel `gtidle` sysfs counters when available and keeps fdinfo as the fallback.

Local Panther Lake/Xe test at `update_ms = 2000`: about `3.46%` -> `1.13%` of one CPU core in the same 30s btop sample, with the embedded GPU row still shown.

Package PR update: https://github.com/omacom-io/omarchy-pkgs/pull/114#issuecomment-4804419830

dhh:
Appreciate all the work that went into this! But I consider it too much to carry a forked package for something with this modest of an upgrade. I think we're better off waiting for btop to take the PR (or not). Please open another PR if btop takes it with the config changes we may need. Thanks again.

PR #6140, opened 2026-06-25
feat: Add hints to navigate the windows vim style

azzenabidi:
This package allows omarchy users to click scroll and drag using the keyboard.
https://github.com/AlfredoSequeida/hints

Inspired by: https://vimium.github.io/


dhh:
Interesting idea, but not something I'd use, so going to skip for now.

PR #6153, opened 2026-06-29
feat: add hyprmod to preinstalled packages and style menu

azzenabidi:
## Summary
- Added HyprMod to the preinstalled package list
- Integrated HyprMod into the Style submenu (Super+Alt+Space)
cj
This provides a GUI to tinker with hyprland setup in omarchy in  a safe env. It does not touch the original config files.

## Changes
- **install/omarchy-base.packages**: Added `hyprmod` to the package list
- **bin/omarchy-menu**: Added HyprMod option to the Style submenu that launches `hyprmod`

### DIFF
--- a/bin/omarchy-menu
+++ b/bin/omarchy-menu
@@ -271,7 +271,7 @@ show_hardware_touchpad_haptics_menu() {
 }
 
 show_style_menu() {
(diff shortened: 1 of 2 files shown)


dhh:
Thanks for the suggestion, but it's not something I'd use myself, so going to skip on this as a default.

PR #6157, opened 2026-07-01
Support per-power-source profile overrides in omarchy-powerprofiles-set

fase29:
## What

Adds an optional user config file at `~/.config/omarchy/power-profile` that lets
you choose which power profile is applied on battery vs. on AC:

```
battery=power-saver
ac=balanced
```

Each key is looked up against the profiles reported by `powerprofilesctl list`,
so any valid profile name works.

## Why

The udev rule installed by `powerprofilesctl-rules.sh` runs
`omarchy-powerprofiles-set` on every AC/USB power event, and the script
unconditionally forces `performance` whenever the machine is on AC.

On my laptop that's a problem: `performance` makes the chassis run hot enough
that the built-in keyboard becomes uncomfortable to type on. I want to stay on
`power-saver` (or `balanced`) even while plugged in, but there was no way to do
that short of editing the script directly — which `git pull` reverts on the next
update.

This makes the AC/battery profile targets configurable, so the tradeoff between
performance and heat/noise/battery can be set to preference instead of being
hardcoded.

## Behavior

- **Fully backward compatible.** With no config file present, behavior is
  identical to today: `performance` on AC (falling back to `balanced` if
  unavailable), `balanced` on battery.
- Unknown keys or invalid/unavailable profile values fall back to the built-in
  defaults, so a typo can never leave the machine in a broken state.
- The user's home is derived from the script's own path rather than `$HOME`,
  because udev invokes the script as root via `systemd-run`.

## Example

~/.config/omarchy/power-profile
```
battery=power-saver
ac=power-saver
```

Keeps the laptop cool and quiet regardless of whether it's plugged in.

### DIFF
--- a/bin/omarchy-powerprofiles-set
+++ b/bin/omarchy-powerprofiles-set
@@ -30,16 +30,41 @@ fi
 
 mapfile -t profiles < <(powerprofilesctl list | awk '/^\s*[* ]\s*[a-zA-Z0-9\-]+:$/ { gsub(/^[*[:space:]]+|:$/,""); print }')
 
-case "$action" in
-  ac)
-    # Prefer performance, fall back to balanced
-    if [[ " ${profiles[*]} " == *" performance "* ]]; then
-      powerprofilesctl set performance
-    else
+profile_available() { [[ " ${profiles[*]} " == *" $1 "* ]]; }
+
+# Optional user overrides in ~/.config/omarchy/power-profile, e.g.:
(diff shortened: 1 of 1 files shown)


dhh:
Fixed but with a different implementation via https://github.com/basecamp/omarchy/commit/0f2f3f11cb42be59a4f9c03af29bcca9abaa6b78. Thanks for raising this ✌️

PR #6161, opened 2026-07-03
Force software cursors on nouveau to fix invisible mouse pointer

stephentaylor-com:
## Problem

On systems using the **nouveau** driver, the mouse cursor is invisible on the physical display under Hyprland. The compositor tracks the pointer correctly (hover, clicks, and `hyprctl cursorpos` all work) — it simply is never painted, so local users see no cursor.

## Root cause

nouveau does not display the DRM hardware-cursor plane on many older NVIDIA GPUs. Hyprland's `cursor:no_hardware_cursors` defaults to `auto`, which keeps the hardware cursor plane in use, so the pointer never renders.

This turned up on a 2010 Mac mini (Apple **GeForce 320M / MCP89**) — a GPU too old for even the legacy `nvidia-580xx` branch in `nvidia.sh`, so Omarchy correctly leaves it on nouveau. But nothing then addresses the cursor, so the install boots to an invisible pointer.

## Fix

Add `install/config/hardware/fix-nouveau-cursor.sh`, which detects nouveau as the active DRM driver and appends a software-cursor override to the user's Hyprland config:

```lua
hl.config({
  cursor = {
    no_hardware_cursors = true,
  },
})
```

It writes to `~/.config/hypr/looknfeel.lua` — the user override file where Omarchy already keeps `cursor` settings (see `default/hypr/looknfeel.lua`) — and is registered in `install/config/all.sh` right after `nvidia.sh`.

- **Keyed on the active driver, not a GPU-model regex** — covers every affected machine rather than just one card. nouveau only ever binds NVIDIA GPUs, so `lspci -k | grep -qi 'Kernel driver in use: nouveau'` has no false positives.
- **Idempotent** — guarded on an existing `no_hardware_cursors` setting, so re-running the installer never duplicates the block.
- `lspci` is guarded with `command -v`, and `LC_ALL=C` keeps the match locale-stable.

## Testing

- Detection verified on the affected hardware (Mac mini, GeForce 320M / MCP89): `lspci -k` reports `Kernel driver in use: nouveau` and the script fires.
- Against the current `looknfeel.lua` template, the appended block parses cleanly under `luac -p`.
- Re-running does not add a second block (idempotency guard).
- `bash -n` passes.

The underlying one-line setting (`no_hardware_cursors = true`) was first applied by hand on the affected machine and confirmed to make the cursor visible.

### DIFF
--- a/install/config/all.sh
+++ b/install/config/all.sh
@@ -40,6 +40,7 @@ run_logged $OMARCHY_INSTALL/config/hardware/printer.sh
 run_logged $OMARCHY_INSTALL/config/hardware/usb-autosuspend.sh
 run_logged $OMARCHY_INSTALL/config/hardware/ignore-power-button.sh
 run_logged $OMARCHY_INSTALL/config/hardware/nvidia.sh
+run_logged $OMARCHY_INSTALL/config/hardware/fix-nouveau-cursor.sh
 run_logged $OMARCHY_INSTALL/config/hardware/vulkan.sh
 
 run_logged $OMARCHY_INSTALL/config/hardware/intel/video-acceleration.sh
--- a/install/config/hardware/fix-nouveau-cursor.sh
+++ b/install/config/hardware/fix-nouveau-cursor.sh
@@ -1,0 +1,35 @@
+# Disable hardware cursors when the nouveau driver is in use.
+#
+# The nouveau DRM driver does not display the hardware cursor plane on many
+# older NVIDIA GPUs (e.g. Apple's GeForce 320M / MCP89 in the 2010 Mac mini),
+# leaving the mouse pointer invisible on the physical display under Hyprland.
+# Forcing software cursors makes the pointer render correctly.
+#
+# The fix appends a Lua `hl.config({ cursor = { no_hardware_cursors = true } })`
+# block to the user's ~/.config/hypr/looknfeel.lua.
+#
+# nouveau only ever binds NVIDIA GPUs, so matching the driver line directly is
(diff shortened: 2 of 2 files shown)


stephentaylor-com:
Good catch on the Lua config — you're right. `dev` has moved the Hyprland setup from `.conf` to Lua (`hyprland.lua` → `require("hypr.*")`), so appending `.conf` syntax to `input.conf` would never load.

Rewritten to emit Lua and target `~/.config/hypr/looknfeel.lua`, which is where Omarchy already keeps `cursor` settings (see `default/hypr/looknfeel.lua`):

```lua
hl.config({
  cursor = {
    no_hardware_cursors = true,
  },
})
```

Verified the generated block parses under `luac -p`. Also folded in the other two suggestions: guarded `lspci` with `command -v` and set `LC_ALL=C` for a locale-stable match. Thanks!

stephentaylor-com:
Both good catches, thanks.

1. You're right about the transient-nouveau window during install — a supported NVIDIA GPU still runs `nouveau` until the first reboot, so keying purely on the live driver would wrongly force software cursors there. Added a `[[ ! -f /etc/modprobe.d/nvidia.conf ]]` guard so the fix only applies when no proprietary driver was configured (i.e. `nvidia.sh` hit its "no compatible driver" path — exactly the old-GPU / nouveau-forever case this targets).
2. Expanded the header comment to note it writes a Lua `hl.config` block to `~/.config/hypr/looknfeel.lua`.

Validated end-to-end on the affected 320M again: guard passes (no `nvidia.conf`), fix applies, `luac -p` clean, idempotent.

PR #6164, opened 2026-07-04
Add keyboard cleaner toggle (Super+Ctrl+K)

ikegam:
## What

Adds a keyboard cleaner: `Super+Ctrl+K` blocks all keyboard input so you can wipe down the keyboard, and the same combo unlocks it.

## How

- `bin/omarchy-toggle-keyboard` — follows the same `on|off|toggle` pattern as `omarchy-toggle-touchpad`, with swayosd feedback. State is read straight from `hyprctl submap`, so no state file is needed.
- `default/hypr/bindings/utilities.conf` — a `cleaner` submap with a `catchall` bind that swallows every key except the exit combo.

## Why a submap instead of disabling the device

Disabling the keyboard via `device[...]:enabled false` (the touchpad approach) would leave no way to re-enable it from the keyboard itself. The submap approach keeps exactly one combo alive as a guaranteed exit, and the mouse keeps working throughout.

The binding uses `bindd`, so it shows up in the keybindings menu (`Super+K`) automatically. `Super+Ctrl+K` was chosen to match the other `SUPER CTRL` toggles; plain `Super+K` is already taken by the keybindings menu.


ikegam:
sorry this is just a gabarage

PR #6180, opened 2026-07-05
[codex] Add natural aliases for menu routes

Pegorim:
## Summary

Adds natural direct-route aliases for common Omarchy menu destinations, including lock-screen, share-file, fonts, displays, default-browser, timezone, and restart-shell. The menu route aliases remain unique, so route resolution stays deterministic.

Related to the Omarchy 4 alpha menu improvements.

## Validation


dhh:
Please don't open draft PRs here. There's enough finished work to review that we don't also need to review drafts.

PR #6184, opened 2026-07-06
Add wallpaper browser for the dharmx/walls collection

samuhlo:
`Style > Background` gains a **󰍉 Browse walls collection** entry: a floating
browser for [dharmx/walls](https://github.com/dharmx/walls) — 1,637 curated
wallpapers in 51 categories — without cloning its 3.7 GB. Browse, preview,
and either set a wallpaper right away or add it to any theme's rotation.

Pick a category, previewed as a mosaic of random samples:

categories

Fuzzy-search with live image previews:

browser

Set it now, or add it to a theme:

actions

## How it works

- **Two files:** `bin/omarchy-walls` (self-contained bash, shows up in
  `omarchy commands`) and one entry in the background selector's Lua.
- **Featherweight:** one GitHub API call indexes the collection, previews are
  ~12 KB thumbnails via a resize proxy, and only what you install downloads
  at full size. A whole browsing session costs a few hundred KB.
- **Theme-native:** colors come from the active theme's `colors.toml` and the
  session's gum styling; the entry's cover is drawn in the theme accent and
  regenerates on theme change.

## One dependency to discuss

Everything used is already a base package except **kitty**, which provides
the graphics protocol for inline previews. Without it, the entry sends a
notification pointing to `omarchy pkg add kitty` instead of failing silently.
Happy to take this wherever you prefer — gating the entry, adding kitty, or
another direction.

## Thanks

All the wallpapers belong to their artists and to the beautiful curation work
of [@dharmx](https://github.com/dharmx) in
[dharmx/walls](https://github.com/dharmx/walls) — this tool just gives that
collection a comfortable home inside Omarchy, downloading on demand and
redistributing nothing.

### DIFF
--- a/bin/omarchy-walls
+++ b/bin/omarchy-walls
@@ -1,0 +1,584 @@
+#!/bin/bash
+
+# omarchy:summary=Browse and install wallpapers from the dharmx/walls collection
+# omarchy:examples=omarchy-walls list, omarchy-walls install nord/forest.png --set
+
+set -uo pipefail
+
+SCRIPT_PATH=$(realpath "${BASH_SOURCE[0]}")
+
+REPO="dharmx/walls"
+BRANCH="main"
+API_URL="https://api.github.com/repos/$REPO/git/trees/$BRANCH?recursive=1"
+RAW_BASE="https://raw.githubusercontent.com/$REPO/$BRANCH"
+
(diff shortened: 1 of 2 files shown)


dhh:
Interesting idea, but I think this would work better as an external option/package. We put a lot of effort into matching the themes with suitable backgrounds, and I prefer leaning on that for the default.

PR #6190, opened 2026-07-08
Add 'yazi' to external packages list

goldstac:
i added yazi terminal file manager to omarchy


dhh:
Thanks for the suggestion, but I'm happy with nautilus as the default.

PR #6193, opened 2026-07-09
fix: rename opencode alias from c to oc to avoid accidental triggers

### LINKED ISSUE #6119: Security risk: single-letter alias c='opencode' turns typos into silent 54MB remote downloads
## Summary

`default/bash/aliases` ships:

```bash
alias c='opencode'
```

This is a real security/footgun risk that I just hit in practice, not in theory.

## What happened

I typed `c;ear` (a typo for `clear`). Bash splits on `;`, so the line ran as two commands:

1. `c` -> expands via the alias to `opencode`.
2. `ear` -> never executed because the opencode launcher (`~/.local/bin/opencode`) ends with `exec ...`, which replaces the shell process.

The opencode wrapper runs `pnpm dlx --package opencode-ai true`, which silently downloaded **54 MB** of remote binaries (`opencode-linux-x64`, `opencode-linux-x64-baseline`) from the npm registry with no confirmation:

```
Packages: +3
Downloading opencode-linux-x64@1.17.9: 54.43 MB/54.43 MB, done
Downloading opencode-linux-x64-baseline@1.17.9: 54.43 MB/54.43 MB, done
```

The only reason I noticed is that 54 MB took long enough for me to read the output and hit `^C`.

## Why this is a security risk, not just an annoyance

1. **A one-character command (`c`) that triggers an arbitrary remote code download with zero confirmation.** A single stray keystroke or a typo (e.g. `c;ear`, `c&`, `c|...`) launches a launcher that `exec`s into a freshly-fetched npm package's binary. The launcher does set `PNPM_CONFIG_MINIMUM_RELEASE_AGE=7200` (good), but that only mitigates supply-chain freshness, not the fact that a typo silently invokes remote-fetched code at all.
2. **`c` is extremely easy to type by accident.** It's one of the most common collision characters: typos for `clear`, `cd`, `cp`, `cat`, autocompletion of `c<Tab>`, etc. Aliasing a single low-friction character to a tool that fetches-and-execs remote binaries is exactly the shape of an accidental-RCE vector.
3. **The launcher uses `exec`, so it replaces the shell.** With `c;ear`, the second command never ran - but with `c && rm ...` style typos, or just `c` by itself when you meant `clear`, you silently leave your shell context and start a network-fetching agent.
4. **It's indistinguishable from malware behavior to an end user.** Typo -> silent 54 MB download from the internet -> exec of a remote package. That is the textbook profile of a malicious "accident," and it's being shipped as a default alias.

## Suggested mitigations

- **Don't alias a single character to a remote-fetching launcher.** `c` is far too collision-prone. Pick something with more entropy (e.g. `oc`, `code`-style), or drop the alias entirely.
- If a short alias is wanted, **guard it so a bare invocation with no arguments doesn't silently fetch/exec**, e.g. only run when given args, or print what it would do first.
- Consider whether the launcher should require an explicit `install`/`run` subcommand rather than fetching on first invocation.

## Environment

- Omarchy alias file: `default/bash/aliases`, line `alias c='opencode'`
- Launcher: `~/.local/bin/opencode` (wrapper that calls `pnpm dlx --package opencode-ai true` then execs the resolved bin)
- Observed with opencode-ai 1.17.9

sridhar-3009:
The single-letter alias `c='opencode'` is a footgun — any typo like `c;ear` (for `clear`) silently triggers a 54MB remote binary download via the opencode launcher. Renaming it to `oc` makes accidental invocation far less likely while keeping the alias short and memorable.

Closes #6119

### DIFF
--- a/default/bash/aliases
+++ b/default/bash/aliases
@@ -43,7 +43,7 @@ alias ...='cd ../..'
 alias ....='cd ../../..'
 
 # Tools
-alias c='opencode'
+alias oc='opencode'
 alias cx='printf "\033[2J\033[3J\033[H" && claude --permission-mode bypassPermissions'
 alias d='docker'
 alias r='rails'


dhh:
Thanks for the suggestion, but we have several single character aliases, like n and t and r. The convenience is worth the risk.

PR #6195, opened 2026-07-09
fix: add delay between selection and capture to avoid cursor in screenshot

### LINKED ISSUE #6142: Screenshot captures the screenshot selection tool in the image
### System details

Omarchy 3.8.2

### What's wrong?

I'm on Omarchy 3.8.2 and when I take a screenshot the crosshair cursor selection tool gets captured when I take the screen shot. I can see the the cursor in the bottom right edge of the screenshot (it looks like this `⅃`).

sridhar-3009:
After `slurp` completes its selection, `grim` was called immediately, capturing the selection crosshair overlay before it had time to disappear. Adding a 100ms delay lets the slurp UI clear before the screenshot is taken.

Closes #6142

### DIFF
--- a/bin/omarchy-capture-screenshot
+++ b/bin/omarchy-capture-screenshot
@@ -123,6 +123,9 @@ esac
 
 [[ -z $SELECTION ]] && exit 0
 
+# Give the selection overlay time to disappear before capturing.
+sleep 0.1
+
 FILENAME="screenshot-$(date +'%Y-%m-%d_%H-%M-%S').png"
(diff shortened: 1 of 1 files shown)


dhh:
Thanks for taking a swing at this one too — closing along with #6212. The root cause is fixed in fa2b997513d04a1f: the crosshair isn't the slurp overlay lingering, it's the cursor itself being software-composited into the frame on GPUs without working hardware cursors, so a delay wouldn't clear it (the cursor is still on screen 100ms later, just as an arrow). The fix forces hardware cursors — which grim omits — for the duration of the capture.

PR #6196, opened 2026-07-09
fix: guard against missing remote_clipboard.lua in migration

### LINKED ISSUE #6143: Migration 1781587663 silently fails — remote_clipboard.lua missing from omarchy-nvim package
## Summary

Migration `1781587663.sh` (commit `2b95eddf`) adds remote clipboard support for Neovim over OSC 52. The tmux half works (terminal feature for `*:clipboard` is added to `tmux.conf`). The Neovim half silently fails because the source file it copies — `/usr/share/omarchy-nvim/config/lua/config/remote_clipboard.lua` — was never shipped in any released `omarchy-nvim` package.

## Evidence

- `omarchy-nvim 2026.4.23-1` installed — the file does not exist at the expected path
- The `remote_clipboard.lua` source file was later added to `omacom-io/omarchy-pkgs` at `pkgbuilds/omarchy-nvim/lua/config/remote_clipboard.lua` (275 lines, full Wayland + headless clipboard provider)
- The PKGBUILD was also updated to bake `require("config.remote_clipboard").setup()` into the pre-built `options.lua`, but no package release has shipped this
- The migration marker exists at `~/.local/state/omarchy/migrations/1781587663.sh`, so `omarchy-migrate --pending` will not re-run it

## Impact

Users who installed Omarchy from the ISO or before the package rebuild get:
- No `remote_clipboard.lua` in their Neovim config
- No `require("config.remote_clipboard").setup()` in `options.lua`
- Silent failure — no error is logged because the `install` command in the migration does not trigger `bash -euo pipefail` (the source path simply doesn't exist, and the failure is absorbed)

## What needs to happen

1. Rebuild and release `omarchy-nvim` with the current source that includes `remote_clipboard.lua` and the baked-in require line
2. OR either ship `remote_clipboard.lua` inside the migration itself (self-contained), or update the migration to create the file inline

## References

- Migration: `migrations/1781587663.sh` in `basecamp/omarchy`
- Source file: `pkgbuilds/omarchy-nvim/lua/config/remote_clipboard.lua` in `omacom-io/omarchy-pkgs`

sridhar-3009:
Migration `1781587663.sh` calls `install -m 0644 "$provider_source" "$nvim_provider"` without first checking whether `$provider_source` exists. When the `omarchy-nvim` package hasn't yet shipped `remote_clipboard.lua`, the `install` command fails silently — no error is logged, but the Neovim clipboard provider is never set up.

This adds an explicit `[[ -f $provider_source ]]` guard and prints a diagnostic message when the file is absent so users know what was skipped and why. The tmux OSC 52 configuration is unaffected and still runs unconditionally.

Closes #6143

### DIFF
--- a/migrations/1781587663.sh
+++ b/migrations/1781587663.sh
@@ -12,12 +12,16 @@ if [[ -d $nvim_config_dir ]]; then
     install -m 0644 "$provider_source" "$nvim_provider"
 
     if [[ -f $nvim_options ]] && ! grep -qF 'config.remote_clipboard' "$nvim_options"; then
-      tmp=$(mktemp)
-      {
-        printf '%s\n' 'require("config.remote_clipboard").setup()'
-        cat "$nvim_options"
-      } >"$tmp"
-      mv "$tmp" "$nvim_options"
+      if tmp=$(mktemp); then
+        if ! {
+          printf '%s\n' 'require("config.remote_clipboard").setup()'
+          cat "$nvim_options"
+        } >"$tmp"; then
+          rm -f "$tmp"
+        elif ! mv "$tmp" "$nvim_options"; then
+          rm -f "$tmp"
+        fi
+      fi
     fi
   else
     echo "Skipping Neovim clipboard provider: $provider_source not found (omarchy-nvim package may need updating)"
--- a/migrations/1781587663.sh
+++ b/migrations/1781587663.sh
@@ -1,0 +1,30 @@
+echo "Enable secure remote Neovim clipboard support"
+
+nvim_config_dir="$HOME/.config/nvim"
+nvim_options="$nvim_config_dir/lua/config/options.lua"
+nvim_provider="$nvim_config_dir/lua/config/remote_clipboard.lua"
+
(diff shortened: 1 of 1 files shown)


dhh:
Thanks for digging into this, but I'm going to close it — the underlying problem has already resolved itself, and the guard would actually have made things worse for affected users.

Two things:

**The migration self-heals by design.** Migrations run under `bash -euo pipefail`, so when `install` couldn't find `remote_clipboard.lua`, the migration failed loudly and its marker was never written — it just retried on the next update. Since updates upgrade packages before running migrations, everyone got healed automatically once `omarchy-nvim 2026.7.15-1` shipped the file. Adding the `[[ -f ]]` guard would invert that: the migration would "succeed", write its marker, and permanently skip the provider even after the package ships it.

**This also targets the wrong branch** — the migration only exists on `quattro`, so this adds a divergent copy as a new file on `dev` that would conflict when the branches merge.

If anyone hit the edge case of creating `~/.config/nvim` *after* the migration ran (marker present, no provider), the fix is:

```
rm ~/.local/state/omarchy/migrations/1781587663.sh && omarchy-migrate
```

Appreciate the report and the effort!

Issue #6053, opened 2026-06-06
omarchy-hook double-runs .d/ hooks, breaking theme-set variable exports


1ceseismic:
omarchy-hook runs both the parent hook (~/.config/omarchy/hooks/<name>) and all scripts in <name>.d/. For theme-set, the parent script already iterates theme-set.d/*.sh internally, passing exported color variables to children. omarchy-hook then re-runs the same .d/ scripts in fresh bash processes without those exports.

Result: hooks like 40-steam.sh that use ${rgb_normal_blue} etc. produce empty CSS so steam renders as broken depp-black ui

reproduction:
1. set any theme: omarchy theme set <theme name>
2. check ~/.local/share/steam-adwaita/adwaita/colorthemes/omarchy/omarchy.css
3. all --adw-*-rgb: values are empty (e.g. --adw-accent-bg-rgb:  !important;)

In bin/omarchy-hook, the parent hook runs first (which exports vars and runs .d/ hooks), then omarchy-hook also runs the .d/ hooks in new bash processes without those exports, overwriting the good CSS.

Fix: skip the .d/ loop when a parent hook exists, since the parent already handles its children, we can just change the second if to elif.

Omarchy 3.8.2, commit 8e031516.

khushveer007:
I've opened a draft PR to address this here: #6062

levyvix:
Confirmed in `bin/omarchy-hook`. The script runs the parent hook and then also iterates the `.d/` directory itself:

```bash
if [[ -f $HOOK_PATH ]]; then
  bash "$HOOK_PATH" "$@" || echo "Hook failed: $HOOK_PATH"
fi

if [[ -d $HOOK_DIR ]]; then
  for hook in "$HOOK_DIR"/*; do
    ...
    bash "$hook" "$@"       # re-runs the same .d/ scripts in a fresh bash
  done
fi
```

For `theme-set`, the parent script already loops over `theme-set.d/*.sh` and passes the exported `rgb_*` color variables to each child. `omarchy-hook` then re-invokes those same `.d/` scripts in fresh `bash` processes with **no exported colors**, so hooks like `40-steam.sh` emit empty `--adw-*-rgb:` values and Steam renders black.

[...]

PR #6197, opened 2026-07-09
Fix omarchy-hook double-running .d/ scripts when a parent hook exists

sridhar-3009:
## Problem

When a parent hook (e.g. `theme-set`) iterates its own `.d/` scripts internally—passing exported colour variables to each child—`omarchy-hook` still runs those same `.d/` scripts a second time in fresh bash processes, but without the exports. The result is that the good output (correct CSS variables) is overwritten by a second run that sees only empty variables.

Reported in #6053. Reproduction:

1. `omarchy theme set <any theme>`
2. Check `~/.local/share/steam-adwaita/adwaita/colorthemes/omarchy/omarchy.css` — all `--adw-*-rgb:` values are empty.

## Fix

Change the second `if [[ -d $HOOK_DIR ]]` to `elif`, so the `.d/` loop only fires when **no parent hook file exists**. A parent hook that owns a `.d/` directory is responsible for running those scripts itself (which lets it pass down any env it needs). When there is no parent hook, the `.d/` loop runs as before.

```diff
 if [[ -f $HOOK_PATH ]]; then
   bash "$HOOK_PATH" "$@" || echo "Hook failed: $HOOK_PATH"
-fi
-
-if [[ -d $HOOK_DIR ]]; then
+elif [[ -d $HOOK_DIR ]]; then
```

One line change, no new dependencies.

### DIFF
--- a/bin/omarchy-hook
+++ b/bin/omarchy-hook
@@ -17,9 +17,7 @@ shift
 
 if [[ -f $HOOK_PATH ]]; then
   bash "$HOOK_PATH" "$@" || echo "Hook failed: $HOOK_PATH"
-fi
-
-if [[ -d $HOOK_DIR ]]; then
+elif [[ -d $HOOK_DIR ]]; then
   for hook in "$HOOK_DIR"/*; do
     [[ -f $hook ]] || continue
     [[ $hook == *.sample ]] && continue


dhh:
Thanks for digging into this — the double-run you've identified is real, but the fix belongs in steam-adwaita, not here.

The single hook file and the `.d/` directory are independent extension points by design, not a parent/child hierarchy. The single file is the legacy mechanism, and `.d/` is where `omarchy hook install` puts drop-ins — both are meant to run. That's what the summary line in the script says, and it's what the migration to `.d/` directories intended.

Making the `.d/` loop an `elif` would silently break that contract: anyone with a legacy single-file hook would have every installed drop-in — including anything added via `omarchy hook install` — just never fire, with no error anywhere. That's a worse and far more common failure mode than the one being fixed.

The actual bug is that steam-adwaita's parent hook iterates `theme-set.d/*.sh` itself, duplicating what `omarchy-hook` already does. The fix there is to either stop iterating `.d/` from the parent hook, or have each drop-in source the color variables itself — which also makes them work correctly when omarchy invokes them directly.

Appreciate the clear writeup and reproduction, but going to pass on this one.

PR #6198, opened 2026-07-09
Surface xdg-open errors from the open shell helper

sridhar-3009:
## Problem

The `open()` shell function silences both stdout *and* stderr from `xdg-open`:

```bash
open() (
  xdg-open "$@" >/dev/null 2>&1 &
)
```

When `xdg-open` fails (unknown MIME type, missing handler, bad path), the user
gets no output at all—the command just silently does nothing.

Reported in #6056.

## Fix

Drop only the stderr redirection. The function remains non-blocking and keeps
stdout suppressed (avoids noise from handlers that write to it), but
`xdg-open` errors now reach the terminal:

```diff
 open() (
-  xdg-open "$@" >/dev/null 2>&1 &
+  xdg-open "$@" >/dev/null &
 )
```

One character change. The non-blocking / no-job-control-noise behavior from
#4366 is fully preserved.

### DIFF
--- a/default/bash/aliases
+++ b/default/bash/aliases
@@ -34,7 +34,7 @@ if command -v zoxide &> /dev/null; then
 fi
 
 open() (
-  xdg-open "$@" >/dev/null 2>&1 &
+  xdg-open "$@" >/dev/null &
 )
 
 # Directories


dhh:
Thanks for digging into this! But the silence here is intentional. On Hyprland, xdg-open takes its generic path and execs the handler directly, so the launched app inherits the terminal's stderr. Dropping the redirect means every Gtk/Qt warning from the app scribbles over your prompt for as long as the app is open — the exact noise this alias exists to suppress (see #4366). And because the fd is shared, there's no one-liner that surfaces xdg-open's own errors without also letting the app chatter. Between a silent failure on a bad path and random warnings polluting the terminal on every successful open, we'll take the silent failure.

Issue #6110, opened 2026-06-19
No audio on Intel Arrow Lake — sof-firmware not installed, DSP fails to boot (Dummy Output)


hyprcat:
## System details

- **Omarchy:** 3.8.2 (branch `master`)
- **Distro / Kernel:** Arch Linux, `linux 7.0.9-arch2-1`
- **Machine:** LENOVO Yoga Pro 7 14IAH10 (`83KF`, board `LNVNB161216`, BIOS `QGCN31WW` 2025-09-17)
- **CPU:** Intel Core Ultra 9 285H — **arch: Arrow Lake**
- **GPU:** Intel Arrow Lake-P [Arc Pro 130T/140T], `i915`
- **Audio controller:** `00:1f.3` Intel **Arrow Lake cAVS** `[8086:7728]`, driver `sof-audio-pci-intel-mtl`, speaker amps Cirrus Logic CS35L56 (x2)
- **Audio stack:** PipeWire 1.6.5 / wireplumber / pipewire-pulse — all active

(Full `omarchy-debug` available on request — happy to upload to logs.omarchy.org if useful.)

## What's wrong?

On a fresh Omarchy 3.8.2 install on this Arrow Lake laptop there is **no audio at all** — PipeWire only exposes a "Dummy Output" sink. The SOF audio DSP fails to boot because `sof-firmware` is not installed.

### Root cause

`install/config/hardware/intel/sof-firmware.sh` only installs `sof-firmware` on Panther Lake:

```bash
if omarchy-hw-intel-ptl && ! omarchy-hw-match "XPS"; then
  omarchy-pkg-add sof-firmware
fi
```

and `omarchy-hw-intel-ptl` matches a single family:

```bash
lspci | grep -iE 'vga|3d|display' | grep -qi 'panther lake'
```

Arrow Lake (and presumably Lunar Lake / Meteor Lake) don't match, so `sof-firmware` is never installed. On the mainline `linux` kernel `sof-firmware` is only an *optdep* (the script's own comment notes `linux-ptl` previously hard-depped it), so nothing else pulls it in. Result: any non-Panther-Lake Intel SOF laptop on mainline `linux` boots with no audio.

### Kernel log

```
sof-audio-pci-intel-mtl 0000:00:1f.3: SOF firmware and/or topology file not found.
sof-audio-pci-intel-mtl 0000:00:1f.3:  Firmware file: intel/sof-ipc4/arl/sof-arl.ri
[...]

PR #6200, opened 2026-07-09
Extend sof-firmware install to all Intel SOF platforms (Arrow Lake, Meteor Lake, etc.)

sridhar-3009:
## Problem

`install/config/hardware/intel/sof-firmware.sh` only installs `sof-firmware`
on Panther Lake systems. Arrow Lake (and Meteor Lake, Tiger Lake, Alder Lake,
etc.) all use the `sof-audio-pci-intel-*` driver family, which needs
`sof-firmware` to boot the DSP. On mainline `linux` the package is only an
`optdep`, so nothing pulls it in automatically. Result: fresh installs on
Arrow Lake laptops boot with no audio — PipeWire exposes only a Dummy Output
sink.

Reported in #6110 (Lenovo Yoga Pro 7 14IAH10, Intel Core Ultra 9 285H).

## Fix

**New detection script** `bin/omarchy-hw-intel-sof`: checks for any Intel
PCI audio controller (class Multimedia audio or Audio device). This reliably
identifies the Skylake-and-later platforms that load `sof-audio-pci-intel-*`
drivers.

**Updated `sof-firmware.sh`**: Replace `omarchy-hw-intel-ptl` with
`omarchy-hw-intel-sof`. The Panther Lake XPS exclusion is preserved — those
systems use `linux-ptl`, which already hard-depends `sof-firmware`.

**Migration** `migrations/1783625096.sh`: Backfills `sof-firmware` on
existing installs that were set up before this fix. `omarchy-pkg-add` is
idempotent so systems that already have the package are unaffected.

### DIFF
--- a/bin/omarchy-hw-intel-sof
+++ b/bin/omarchy-hw-intel-sof
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+# omarchy:summary=Detect whether the computer has an Intel SOF-capable audio DSP.
+# Matches Intel audio controllers (PCI class Multimedia audio or Audio device)
+# present on Skylake-and-later platforms that use sof-audio-pci-intel-* drivers.
+
+lspci | grep -iE '(Multimedia audio controller|Audio device).*Intel'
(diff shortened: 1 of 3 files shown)


dhh:
Included via ffa81021136281384a373f79b778139883e6e1dc. Thanks!

PR #6201, opened 2026-07-10
Move toggle config source before user overrides in hyprland.conf

sridhar-3009:
## Problem

Toggle conf files (`~/.local/state/omarchy/toggles/hypr/*.conf`) are loaded **after** the user's custom files (`looknfeel.conf`, etc.), so any Hyprland variable set by a toggle — like `single_window_aspect_ratio` from the tiling toggle — silently overrides whatever the user put in their own `looknfeel.conf`. The only workaround today is to source `looknfeel.conf` again at the very bottom of `hyprland.conf`.

Reported in #6068.

## Fix

Move the toggle `source` glob to be between the Omarchy defaults and the user's custom files:

```
Omarchy defaults → current theme → toggle flags → user overrides
```

User overrides always win, which is the expected mental model: toggles express Omarchy feature flags, not unoverrideable system policy.

### DIFF
--- a/config/hypr/hyprland.conf
+++ b/config/hypr/hyprland.conf
@@ -12,15 +12,16 @@ source = ~/.local/share/omarchy/default/hypr/input.conf
 source = ~/.local/share/omarchy/default/hypr/windows.conf
 source = ~/.config/omarchy/current/theme/hyprland.conf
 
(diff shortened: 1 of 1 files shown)


dhh:
Closing in favor of the Lua-compatible toggle load-order fix in #6116.

PR #6202, opened 2026-07-10
Add upstream Arch fallback mirrors to boot.sh

### LINKED ISSUE #6158: boot.sh forces stable-mirror.omarchy.org, but git-2.55.0-1-x86_64.pkg.tar.zst returns HTTP 404, preventing installation.
### System details

macbook pro

### What's wrong?

curl -I https://stable-mirror.omarchy.org/core/os/x86_64/git-2.55.0-1-x86_64.pkg.tar.zst

Last login: Wed Jul  1 20:33:56 on ttys001
xxx@MacBook-Pro ~ % curl -I https://stable-mirror.omarchy.org/core/os/x86_64/git-2.55.0-1-x86_64.pkg.tar.zst
HTTP/2 404
date: Wed, 01 Jul 2026 17:55:49 GMT
content-type: text/plain;charset=UTF-8
server: cloudflare
cf-cache-status: DYNAMIC
cf-ray: a14748aadc5652df-WAW

sridhar-3009:
Fixes #6158

## Problem

`boot.sh` overwrites `/etc/pacman.d/mirrorlist` with a single omarchy mirror line. If that mirror is temporarily behind and a package version (e.g. `git-2.55.0-1`) returns HTTP 404, `pacman -Syu --needed git` fails with no fallback and the entire bootstrap is aborted.

## Fix

Add two well-known upstream Arch Linux mirrors (`geo.mirror.pkgbuild.com` and `mirror.rackspace.com`) after the primary omarchy mirror in each branch of the `if` block. Pacman tries mirrors in order, so the omarchy mirror is still preferred; the upstream mirrors serve as a safety net when it's temporarily stale.

The approach matches how robust `mirrorlist` configurations are typically set up and does not risk pulling in packages from an incompatible state — `pacman -Syu` still updates the database from whichever mirror responds, and only `git` is installed at this stage.

### DIFF
--- a/boot.sh
+++ b/boot.sh
@@ -25,13 +25,25 @@ OMARCHY_REF="${OMARCHY_REF:-master}"
 # Set mirror based on branch
 if [[ $OMARCHY_REF == "dev" ]]; then
   export OMARCHY_MIRROR=edge
-  echo 'Server = https://mirror.omarchy.org/$repo/os/$arch' | sudo tee /etc/pacman.d/mirrorlist >/dev/null
+  {
+    echo 'Server = https://mirror.omarchy.org/$repo/os/$arch'
+    echo 'Server = https://geo.mirror.pkgbuild.com/$repo/os/$arch'
+    echo 'Server = https://mirror.rackspace.com/archlinux/$repo/os/$arch'
+  } | sudo tee /etc/pacman.d/mirrorlist >/dev/null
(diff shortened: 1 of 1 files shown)


dhh:
Wouldn't work with our strategy of keeping updates 1 month behind edge for compatibility and safety, unfortunately. We intentionally have our own mirror for that reason.

PR #6210, opened 2026-07-13
docs(skill): add Installing CLI Tools section

saiqulhaq:
## Summary

Adds an "Installing CLI Tools" section to the Omarchy skill covering where new CLI binaries should live and how to install them without hanging or duplicating configuration.

## What's added

Four install paths documented for the case where the user asks to install a new CLI tool, plus a decision-rules summary:

- **npm-based CLIs** -- `omarchy npm install` (the grouped route; underlying binary `omarchy-npm-install`) writes a `pnpm dlx`-based launcher into `~/.local/bin`, mirroring the pattern used for the built-in AI tools (`codex`, `gemini`, `copilot`, `pi`). Scoped npm packages (`@org/pkg`) need an explicit `command-name` to avoid launcher-path failures.
- **Downloaded binaries** (tarballs, GitHub releases, `.deb` extracts) -- drop into `~/.local/bin`; `mkdir -p` first if the directory doesn't exist. No shell config edits.
- **Pacman / AUR packages** -- `omarchy pkg add` / `omarchy pkg aur add`.

The intro also notes that Omarchy appends `~/.local/bin` to `PATH` from two Omarchy-managed files (`~/.local/share/omarchy/default/bash/envs` for interactive bash, `~/.config/uwsm/env` for graphical sessions) and that these should not be hand-edited -- bash envs changes ship through `omarchy update`, uwsm env changes through `omarchy-refresh-config uwsm/env`.

A trailing note clarifies that Omarchy's default `~/.bashrc` invites user customization, so editing it is fine -- what this section discourages specifically is adding a `PATH` export for a non-standard install directory (like `~/.bin/`).

Trigger description was also extended so the skill loads on install queries (`installing CLI tools`, `package management`, `omarchy pkg add`, `omarchy npm install`).

## Why

The skill previously lacked guidance on where new CLI binaries should live. A common mistake (and one I made on this machine today) is to invent a custom directory like `~/.bin/`, add a `PATH` export to `~/.bashrc`, and then wonder why future instructions get tangled. Omarchy already puts `~/.local/bin` on `PATH` via two Omarchy-managed files, so binaries dropped there need no shell config changes.

## Test plan

- [x] `git diff --stat` shows 1 file changed, 79 insertions, 1 deletion.
- [x] Markdown renders cleanly (no orphan `\`\`\`` or stray headings).
- [x] Section sits between `### Other Configs` (which has the table of config paths) and `## Safe Customization Patterns`.
- [x] All commands referenced (`omarchy npm install`, `omarchy-npm-install`, `omarchy pkg add`, `omarchy pkg aur add`, `omarchy-refresh-config`) exist in the repo and have been verified at the matching paths.

## Notes

- Markdown only; no scripts or source-code changes.
- Targets the `dev` branch.
- Scope: CLI install paths only. Launcher entries (`.desktop` files, `omarchy tui install`, `omarchy-refresh-applications`) are out of scope for this PR -- they're already covered by `Configuration Locations > Other Configs` and `Applications`.

### DIFF
--- a/default/omarchy-skill/SKILL.md
+++ b/default/omarchy-skill/SKILL.md
@@ -7,7 +7,7 @@ description: >
   or ~/.config/omarchy/. Triggers: Hyprland, window rules, animations, keybindings,
   monitors, gaps, borders, blur, opacity, waybar, walker, terminal config, themes,
   background, night light, idle, lock screen, screenshots, reminders, layer rules,
-  workspace settings, display config, and user-facing omarchy commands. Excludes Omarchy
+  workspace settings, display config, installing CLI tools, package management, omarchy pkg add, omarchy npm install, and user-facing omarchy commands. Excludes Omarchy
   source development in ~/.local/share/omarchy/ and `omarchy dev` workflows.
 ---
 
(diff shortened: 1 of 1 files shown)


dhh:
Quattro switches to mise. But also haven't found that this level of guidance is necessary for newer models.

PR #6211, opened 2026-07-13
docs(skill): document elevation patterns for agent sessions

saiqulhaq:
## Summary

Adds an "Elevation in agent / non-interactive sessions" subsection to `default/omarchy-skill/SKILL.md`, documenting how to run privileged commands without hanging an agent/IDE session.

## What's added

A three-option recommendation ladder right after the existing `omarchy debug --no-sudo` rule (so the two hang-avoidance patterns sit together):

1. **`pkexec <command>`** -- polkit's `sudo` equivalent. Provided by the `polkit` package (pulled in transitively by `polkit-gnome`, see `install/omarchy-base.packages`; `polkit-gnome` is the auth agent that makes the dialog appear). Pops a graphical auth dialog so it works from agents and SSH-from-LAN-with-display sessions without TTY input. Only works in a logged-in graphical session; over non-graphical SSH it fails (e.g. "Cannot open display") rather than blocking.

2. **`omarchy sudo passwordless [MINUTES]`** -- purpose-built for AI agents / batch installers that need to run many `sudo` commands. Drops a time-limited passwordless sudoers rule for the current user; auto-expires. Note the security warning it prints: any process running as your user can execute any command as root during the window.

3. **`sudo -n <command>`** -- non-interactive; fails fast if no password is cached, instead of hanging waiting for input.

Explicit warning: do NOT use plain `sudo <command>` from an agent session -- it will fail or hang waiting for a TTY the agent is not connected to.

A clarifying note is also added: Omarchy commands with `# omarchy:requires-sudo=true` (e.g. `omarchy-pkg-add`) are **metadata-only** -- the flag does not switch `sudo` to `pkexec`, so those commands still call raw `sudo` and will hang from a non-TTY agent session unless option 2 (passwordless sudo) is active or they are invoked via option 1 (`pkexec`).

## Why

I hit this exact hang today running commands from an agent/IDE session. The skill already says to use `omarchy debug --no-sudo --print` for the same reason (avoid TTY prompt), but didn't cover `pkexec` or the general elevation problem. Omarchy ships polkit (via `polkit-gnome`), but the `# omarchy:requires-sudo=true` metadata does **not** switch `sudo` to `pkexec` -- that distinction was not documented anywhere and tripped me up while writing this PR.

## Test plan

- [x] `git diff --stat` shows 1 file changed, 44 insertions.
- [x] New subsection sits between `### System` (which has the existing `--no-sudo` rule) and `## Troubleshooting`.
- [x] All `pkexec` examples reference commands that exist in core Arch packages (`pacman`, `systemctl`).
- [x] Polkit package relationship clarified (`pkexec` from `polkit`, dialog from `polkit-gnome`).

## Notes

- Markdown only; no scripts or source changes.
- Independent of the still-open PR #6210 (`docs/skill-install-paths`).
- Verified locally: `pkexec --version` -> `pkexec version 127`, `pkexec echo "test"` works in this Hyprland/Wayland session.
- Updated after Copilot review pointed out that `omarchy:requires-sudo` is metadata-only and that `pkexec` is provided by `polkit` (not `polkit-gnome`).

### DIFF
--- a/default/omarchy-skill/SKILL.md
+++ b/default/omarchy-skill/SKILL.md
@@ -327,6 +327,120 @@ omarchy system reboot           # Reboot
 
 **IMPORTANT:** Always run `omarchy debug` with `--no-sudo --print` flags to avoid interactive sudo prompts that will hang the terminal.
 
+### Elevation in agent / non-interactive sessions
+
+When running from an agent, IDE, or any non-TTY context, plain `sudo`
+cannot prompt for a password (there is no controlling TTY) and will fail
+with "a terminal is required" or hang waiting on one. Prefer one of the
+following instead:
+
+1. **`pkexec <command>`** for arbitrary one-off commands. `pkexec` is the
+   polkit equivalent of `sudo` -- it is provided by the `polkit`
+   package, which Omarchy pulls in transitively via `polkit-gnome` (see
+   `install/omarchy-base.packages`; `polkit-gnome` is the graphical
+   authentication agent that makes the dialog appear). It pops a
+   graphical authentication dialog on the user's desktop session, so it
+   works from agents and SSH-from-LAN-with-display sessions without TTY
+   input. It only works in a logged-in graphical session; over a
+   non-graphical SSH login it will fail (e.g. "Cannot open display")
+   rather than blocking.
+
+   ```bash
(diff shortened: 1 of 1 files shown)


dhh:
We have a simpler version of this on quattro already.

PR #6212, opened 2026-07-13
fix: prevent crosshair cursor from appearing in screenshots

### LINKED ISSUE #6142: Screenshot captures the screenshot selection tool in the image

cdhananjay:
Closes #6142

When taking a region, window, or smart screenshot, the slurp crosshair cursor was captured in the image (visible in the bottom-right corner). This is a known wlroots issue where grim captures the software-composited cursor when cursor:no_hardware_cursors is enabled.

Temporarily enable hardware cursors before slurp starts so the cursor is rendered in a GPU plane invisible to grim. The setting is restored on exit via the cleanup trap.

### DIFF
--- a/bin/omarchy-capture-screenshot
+++ b/bin/omarchy-capture-screenshot
@@ -69,6 +69,7 @@ get_rectangles() {
 # frozen overlay rather than live content shifting during teardown.
 cleanup_freeze() {
   [[ -n $PID ]] && kill $PID 2>/dev/null
+  hyprctl keyword cursor:no_hardware_cursors 2 2>/dev/null
 }
 trap cleanup_freeze EXIT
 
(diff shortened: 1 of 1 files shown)


dhh:
Thanks for tracking this down — the diagnosis was right: software-composited cursors get baked into the frames grim captures. Fixed in fa2b997513d04a1f on the quattro branch with a slightly different shape: the toggle now spans the whole capture including the grim call (which also covers fullscreen shots), restores the user's actual prior setting via getoption rather than hardcoding 2, and uses hyprctl eval since keyword is rejected by quattro's Lua config parser. Appreciate the report and the fix!

PR #6217, opened 2026-07-14
Remove bsnes2014 & mame2016 from RetroArch setup

joshuagarnerxyz:
Both cores cause install to fail because they no longer exist in the repos.


dhh:
Hotfix applied to 3.8.3/master 👍

PR #6227, opened 2026-07-16
Sync Obsidian base theme with Omarchy light/dark mode

KristianHolme:
Set each vault's appearance.json theme to moonstone or obsidian from light.mode so Obsidian UI chrome matches the system theme.

Useful when the obsidian theme does not cover everything.

## Motivation
When using the Claudian plugin to obsidian, some parts of the UI seems to not follow the omarchy theme (maybe the .css is missing something?), so if obsidian is set to light mode, then some text is black and rendered on dark background from the dark Omarchy theme, making text hard to read. A simple solution is to automatically toggle obsidian ligh/dark mode according to the selected Omarchy theme.


dhh:
We have full theme syncing for obsidian that already honors the light/dark divide. I think best to stick with just that one way of doing it.

PR #6303, opened 2026-07-18
Use hyprsunset-tui for Super + Ctrl + N nightlight binding

heymaikol:
## What

Rebinds `Super + Ctrl + N` to launch [hyprsunset-tui](https://aur.archlinux.org/packages/hyprsunset-tui) (interactive temperature/gamma control) when installed, falling back to the existing `omarchy-toggle-nightlight` when it isn't — so the binding never breaks.

Related discussion: #6257 (Configurable automatic nightlight schedule) — an interactive hyprsunset frontend on the existing nightlight binding covers much of what's being asked for there without Omarchy growing its own scheduling config.

- `default/hypr/bindings/utilities.lua`: bind checks `omarchy-cmd-present hyprsunset-tui` at keypress, launches the TUI or falls back to the toggle
- `migrations/1784412283.sh`: installs `hyprsunset-tui` via `omarchy-pkg-aur-add`, skipping gracefully if the AUR is unreachable (fallback keeps the binding working)

## Notes

- `hyprsunset-tui` is AUR-only today, so it's not in `omarchy-base.packages` — fresh installs keep the plain nightlight toggle until the package is mirrored into the `[omarchy]` repo (like `walker`/`voxtype-bin`), at which point it can be added to the base list and the fallback becomes a safety net.
- `omarchy toggle nightlight` also remains available from the Toggle menu (`Super + Ctrl + O`).

### DIFF
--- a/default/hypr/bindings/utilities.lua
+++ b/default/hypr/bindings/utilities.lua
@@ -29,7 +29,7 @@ o.bind("SUPER + ALT + COMMA", "Invoke last notification", "makoctl invoke")
 o.bind("SUPER + SHIFT + ALT + COMMA", "Restore last notification", "makoctl restore")
 
 o.bind_toggle("SUPER + CTRL + I", "Toggle locking on idle", "idle")
-o.bind_toggle("SUPER + CTRL + N", "Toggle nightlight", "nightlight")
+o.bind("SUPER + CTRL + N", "Nightlight controls", { tui = "hyprsunset-tui", focus = true })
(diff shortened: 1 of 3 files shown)


dhh:
Interesting idea, but find this too magical out of the box. If you want a different night time setup, just rebind the in your bindings to what you'd like to use instead.

heymaikol:
@dhh Its just a tool that lets you change the configuration of hyprsunset and control the service.

PR #6304, opened 2026-07-18
Add unified password update for user and encrypted drives

heymaikol:
## Problem

Update > Password makes it easy to change only the user password and forget the drive encryption passphrase (or vice versa). Since Omarchy sets both to the same password at install time, most users updating their password want both changed together — but the menu only offers them separately.

## Solution

Adds an "All" option to the Update Password menu, backed by a new `omarchy-update-password` script that:

1. Prompts once for the new password (with confirmation)
2. Changes the LUKS passphrase on every encrypted drive (`cryptsetup luksChangeKey`, same argon2id parameters as `omarchy-drive-password`)
3. Changes the user password via `chpasswd`

Drives are changed first: if the current encryption passphrase is wrong, the script exits before touching the user password, so the two never drift apart. Passphrases are passed to cryptsetup via process substitution, so they never appear in argv or on disk.

The existing "Drive Encryption" and "User" options are unchanged for anyone whose passwords intentionally differ.

### DIFF
--- a/bin/omarchy-menu
+++ b/bin/omarchy-menu
@@ -851,7 +851,8 @@ show_update_hardware_menu() {
 }
 
 show_update_password_menu() {
-  case $(menu "Update Password" "  Drive Encryption\n  User") in
+  case $(menu "Update Password" "  All\n  Drive Encryption\n  User") in
+  *All*) present_terminal omarchy-update-password ;;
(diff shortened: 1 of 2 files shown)


dhh:
We have Update > Password and then this screen:

image

So don't think we need more than that. Appreciate the suggestion though.

heymaikol:
Fair.

PR #6305, opened 2026-07-18
Expand system update widget with packages and user themes

HANCORE-linux:
## Summary

Expand the existing `omarchy.system-update` indicator into a service-backed update center with two popup tabs:

- **Packages** shows pending official repository package updates and launches the existing `omarchy update` flow.
- **Themes** checks user-installed Git themes, explains clean and blocked states, and provides safe per-theme actions.

The existing plugin ID and IPC contract remain compatible.

## Motivation

The current widget only reports updates for `omarchy` or `omarchy-dev`; it neither exposes the full repository update set nor safely reviews user themes. This adds visibility without changing update ownership.

## Packages

- Show structured `checkupdates` results with installed and target versions.
- Represent scan, unavailable, malformed, current, and pending states explicitly.
- Support manual refresh and the existing low-frequency automatic check.
- Launch the visible `omarchy update` terminal flow for apply.

The list is informational: QML never invokes `pacman` or a partial upgrade; `omarchy update` retains transaction ownership.
screenshot-2026-07-19_00-47-40

## Themes

- List user-installed Git themes under `~/.config/omarchy/themes` and identify the current theme.
- Distinguish clean, outdated, locally changed, conflict, unreachable, and no-upstream states.
- Explain blocked states through row tooltips instead of long table text.
- Update an eligible theme to the exact commit reviewed by the preceding check.
- Remove a theme through `omarchy theme remove` only after inline confirmation.
- Prevent removal of the active theme until another theme is selected.

Stock themes remain package-backed. QS Rise supplies the UX reference; implementation uses native Omarchy components.
screenshot-2026-07-19_00-47-16

## Theme update safety

The check records branch, remote, upstream, base commit, and exact target commit. Apply revalidates them and requires the target to remain reachable and fast-forwardable.

Tracked edits, local commits, and incoming collisions with untracked or ignored paths block apply, including file/directory prefix collisions. Network operations are bounded and non-interactive. A moving remote cannot replace the reviewed target.

**Reinstall** is excluded because the current installer removes the existing theme before cloning. It should wait for a transactional implementation.

## Architecture and compatibility

Convert the widget into a first-party `service` + `bar-widget` plugin. One keep-loaded service owns scans and state; the widget only presents it, preventing duplicate polling across monitors. Existing IPC remains compatible.

## Follow-ups

Two follow-ups are planned: a **Plugins** tab using Omarchy's existing review flow, and transactional theme reinstall before exposing **Reinstall**. QS Rise shell updates, automatic apply, and separate AUR listing remain outside this update center.

## Validation

- Package fixtures cover updates, no updates, unavailable checker, malformed output, timeouts, and contradictory results.
- Theme fixtures cover remote movement, provenance drift, local work, unreachable remotes, overwrite conflicts, and active-theme protection.
- Apply is verified to reach the pinned target while rejected operations leave theme files unchanged.
- Focused model, contract, manifest-entrypoint, IPC, and screenshot tests pass.
- Qt 6 `qmllint`, ShellCheck, Bash syntax checks, and `git diff --check` pass.
- A Quattro runtime smoke on a separate machine verified plugin loading, bar geometry, IPC summon/hide, popup creation, nonblank rendering, a real read-only package scan, and a read-only scan of copied user-theme repositories.

The aggregate runner's packaging check requires an external sibling `omarchy-pkgs` checkout; focused tests complete independently.


dhh:
Love the experimentation here, but I think it's too much for the default. Let's get this setup as a plugin that people can add. Which reminds me that we need to have a plugin repository so folks can find these new plugins. Thanks for working on this!

PR #6307, opened 2026-07-18
shell: handle CTRL+BACKSPACE to clear filter

RushiChaganti:
Align filter-field editing with Qt's standard keyboard shortcuts across all menu panels (clipboard, emojis, image-picker, launcher, menu, reminders):

- Backspace: delete previous character
- Ctrl+Backspace: delete previous word (Qt DeleteStartOfWord)
- Ctrl+U: clear entire filter

Previously Ctrl+Backspace cleared the whole field, diverging from Qt on KDE/GNOME. Preserves per-panel behavior (menu goBack, image-picker filterable gating) and fixes an indentation regression in the emoji handler.

### DIFF
--- a/shell/Commons/Util.qml
+++ b/shell/Commons/Util.qml
@@ -79,6 +79,25 @@ QtObject {
     }
   }
 
+  // Standard Qt text-editing keys shared by every searchable panel's filter:
+  //   Backspace       delete previous character
+  //   Ctrl+Backspace  delete previous word (Qt DeleteStartOfWord)
+  //   Ctrl+U          clear the whole field
+  // Panels decide how to apply the result (setFilter/updateFilter) and keep
+  // their own empty-field fallbacks (e.g. menu back-navigation).
+  function isFilterEditKey(event) {
+    return event.key === Qt.Key_Backspace
+        || (event.key === Qt.Key_U && (event.modifiers & Qt.ControlModifier))
+  }
+
+  // New filter text after applying an edit key. Assumes isFilterEditKey(event).
+  function editedFilter(event, text) {
+    if (event.key === Qt.Key_U) return ""                        // Ctrl+U: clear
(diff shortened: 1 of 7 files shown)


RushiChaganti:
I just noticed that none of the existing tools have Ctrl + Backspace support. That's definitely something I'll be adding, since it seems like a valuable feature.

dhh:
Could we align this with Qt’s standard keyboard behavior?

Qt defines `Ctrl+Backspace` as `DeleteStartOfWord` on both KDE and GNOME, so it should remove the previous word rather than clear the entire filter. For clearing the complete filter, `Ctrl+U` is the standard shortcut.

Suggested behavior:

- `Backspace`: remove the previous character
- `Ctrl+Backspace`: remove the previous word
- `Ctrl+U`: clear the entire filter

Please also fix the indentation inconsistencies introduced in `Emojis.qml`, `ReminderFlow.qml`, and `ImagePicker.qml`.

RushiChaganti:
https://github.com/user-attachments/assets/e1eb4bbe-7264-476c-a977-6be2a3c15eff

A video preview

RushiChaganti:
Hey @dhh,

I also noticed that the Wi-Fi networks aren't being imported. I've already written a migration script for that.

Would you prefer that I include it in this PR, or should I submit it in a separate PR after this one is merged?

dhh:
Always keep PRs focused on a single issue.

dhh:
Lots of duplicated code in this PR. Find a way to abstract and extract.

RushiChaganti:
Good call — the Ctrl+U / Ctrl+Backspace / Backspace handling was copy-pasted across all six searchable panels (clipboard, emojis, image-picker, launcher, menu, reminders). I've extracted it.

The shared logic now lives in two pure helpers on the Util singleton (shell/Commons/Util.qml)

PR #6308, opened 2026-07-19
Use tmux to list keybindings

heymaikol:
## Summary
- replace the hand-rolled parser with annotated `tmux list-keys -N` output
- keep the 42 shipped bindings curated by storing descriptions beside each bind
- make `--config` authoritative on a per-invocation isolated socket that is always cleaned up

This removes 234 lines net without touching the default tmux server.

## Tests
- `bash -n bin/omarchy-menu-tmux-keybindings`
- verified exactly 42 annotated bindings and no built-in defaults
- verified an alternate config wins while a default server is running
- verified `exit-empty off` leaves no server or socket behind
- verified the default server prefix and bindings remain unchanged
- `./test/all`


dhh (changes requested):
Love the instinct here — the awk parser deserved to die, and letting tmux report its own state is the right direction. But as written this trades away more than it deletes. Three things to fix before this can go in:

**1. Output readability.** The old viewer printed ~45 curated lines like `PREFIX + c → New window`. This prints 310 raw `bind-key` lines, ~265 of which are tmux's built-in copy-mode defaults, drowning out the Omarchy bindings — and it's feeding end-user surfaces (the menu popup and `PREFIX + ?`). Suggestion: use `tmux list-keys -N` instead, and add `-N "description"` notes to the binds in `config/tmux/tmux.conf`. That gives clean, annotated output with the prefix already resolved (`C-Space c   Create a new window`), moves the descriptions into the config next to the bindings they describe, and still deletes the whole parser.

**2. `--config` is silently ignored when a server is running.** tmux only reads `-f` at server start, so with a live server (which is always the case for the `PREFIX + ?` binding) the flag does nothing — verified with a running server on one config and `-f` pointing at another. Pick one semantic: either config-file-authoritative (run on an isolated socket, e.g. `tmux -L omarchy-keybindings -f "$config_file"`) or live-server (drop `--config` and the file checks). Right now the answer depends on invisible state.

**3. Side effects on the default socket.** When no server is running, this starts a real one and fully sources the user's config (including any `run-shell` hooks) just to print keybindings — and a config with `set -g exit-empty off` leaves that server orphaned. An isolated `-L` socket for the cold path fixes this too.

Also a nit: the `omarchy:summary` comment still says "defined in your configuration," which stops being true once defaults are listed.

heymaikol:
@dhh Addressed in b2b9ed61:

1. Added `-N` notes beside all 42 shipped bindings and switched the viewer to curated `list-keys -N` output.
2. Made `--config` authoritative by sourcing it on a unique isolated socket.
3. Added guaranteed server/socket cleanup, including configs with `exit-empty off`.
4. Updated the stale summary.

Verified the output count/readability, alternate-config behavior with a live default server, unchanged default-server state, socket cleanup, and `./test/all`.

EDIT: Used at dhh because I changed commit author to use my username

PR #6315, opened 2026-07-19
fixed the omarchy-cmd-terminal-cwd for kitty

xmagma-x:
Title:
Fix omarchy-cmd-terminal-cwd to correctly detect cwd for Kitty terminal
Description:
## Problem

`omarchy-cmd-terminal-cwd` always returns `$HOME` when used from Kitty terminal, making the keybinding `SUPER+SHIFT+ALT+F` (which uses `$(omarchy-cmd-terminal-cwd)`) useless.

## Root Cause

Three issues with the original script when running under Kitty:

1. **Kitty's client-server architecture**: Kitty uses a daemon that spawns tabs/windows as child processes. `hyprctl activewindow` returns the Kitty window PID, but that PID belongs to the Kitty daemon — not the shell process inside the window. The daemon has no direct `cwd`.

2. **`/proc/$pid/children` doesn't exist**: The original script tried to read `/proc/$pid/children` to walk the process tree, but this file does not exist on all kernels despite `CONFIG_PROC_CHILDREN=y`. Using `pgrep -P` is the reliable alternative.

3. **IFS null-byte issue**: The system's default `IFS` includes a null byte, which causes `for child in $children` to silently fail. Using `while IFS= read -r` fixes this.

## Changes

- **Kitty detection**: Added class-based check via `hyprctl activewindow` to detect Kitty.
- **Primary path for Kitty**: Uses `kitty @ ls` + `jq` to query the focused window's shell PID and cwd directly via Kitty's remote control protocol.
- **Fallback for Kitty**: When `kitty @ ls` fails (e.g., missing `KITTY_PUBLIC_KEY` in keybinding context), extracts the daemon PID from systemd transient scopes (`/run/user/$UID/systemd/transient/kitty-*.scope`), then recursively walks the process tree via `pgrep -P` to find the deepest shell.
- **Non-Kitty terminals**: Uses recursive `pgrep -P` walking instead of `/proc/$pid/children`.
- **IFS fix**: All child iteration uses `while IFS= read -r` to handle null bytes in IFS correctly.
PR body (markdown):
### Problem

`omarchy-cmd-terminal-cwd` always returns `$HOME` when used from Kitty terminal, making the `SUPER+SHIFT+ALT+F` keybinding useless.

### Root Cause

Three issues under Kitty:

1. **Kitty's client-server architecture**: `hyprctl activewindow` returns the Kitty daemon PID, which has no direct `cwd`.
2. **`/proc/$pid/children` doesn't exist**: Not available on all kernels. `pgrep -P` is the reliable alternative.
3. **IFS null-byte**: System default IFS includes a null byte, breaking `for child in $children`. Fixed with `while IFS= read -r`.

### Solution

- **Kitty primary path**: `kitty @ ls` + `jq` to get focused window's shell PID and cwd.
- **Kitty fallback**: Extract daemon PID from systemd transient scopes, recursively walk children via `pgrep -P`.
- **Non-Kitty**: Recursive `pgrep -P` instead of `/proc/$pid/children`.
- **IFS fix**: All child iteration uses `while IFS= read -r`.

### Testing

Verified working on Kitty 0.47.4 with both manual command execution and the `SUPER+SHIFT+ALT+F` keybinding.

### DIFF
--- a/bin/omarchy-cmd-terminal-cwd
+++ b/bin/omarchy-cmd-terminal-cwd
@@ -3,19 +3,66 @@
 # omarchy:summary=Print the current working directory of the active terminal window
 # omarchy:hidden=true
 
-terminal_pid=$(hyprctl activewindow | awk '/pid:/ {print $2}')
-shell_pid=$(pgrep -P "$terminal_pid" | tail -n1)
+active_info=$(hyprctl activewindow)
+terminal_pid=$(echo "$active_info" | awk '/pid:/ {print $2}')
+terminal_class=$(echo "$active_info" | awk '/class:/ {print $2}')
 
-if [[ -n $shell_pid ]]; then
-  cwd=$(readlink -f "/proc/$shell_pid/cwd" 2>/dev/null)
-  shell=$(readlink -f "/proc/$shell_pid/exe" 2>/dev/null)
+cwd=""
+shell_pid=""
 
-  # Check if $shell is a valid shell and $cwd is a directory.
-  if grep -qs "$shell" /etc/shells && [[ -d $cwd ]]; then
-    echo "$cwd"
-  else
-    echo "$HOME"
+_find_deepest_shell() {
+  local parent=$1
+  local children
+  children=$(pgrep -P "$parent" 2>/dev/null)
+  while IFS= read -r child; do
+    [[ -z $child ]] && continue
+    local exe
+    exe=$(readlink -f "/proc/$child/exe" 2>/dev/null)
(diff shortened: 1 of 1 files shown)


dhh:
Thanks for reporting this. Fixed in https://github.com/basecamp/omarchy/commit/008f3a22 using a smaller deterministic approach: Kitty gets a per-process remote-control socket, and the cwd helper queries only `state:focused` while preserving the existing fallback for other terminals.

PR #6321, opened 2026-07-20
Import iwd WiFi networks into NetworkManager on upgrade

RushiChaganti:
The quattro migration retired iwd/systemd-networkd in favor of NetworkManager but never carried over previously-saved WiFi networks, so they appear "forgotten" after upgrading.

Add a migration that re-creates them as NetworkManager profiles from iwd's on-disk store (/var/lib/iwd). It is idempotent (skips SSIDs NetworkManager already knows), best-effort (never aborts the migration run), handles open/WPA-PSK and hidden networks, and leaves 802.1x enterprise networks for manual re-add.

### DIFF
--- a/migrations/1784523021.sh
+++ b/migrations/1784523021.sh
@@ -1,0 +1,108 @@
+echo "Import WiFi networks from iwd into NetworkManager"
+
+# The quattro upgrade retired iwd/systemd-networkd in favor of NetworkManager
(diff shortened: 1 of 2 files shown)


RushiChaganti:
Now that #6334 landed, enterprise networks are covered too — this migration + that PR round out the full iwd → NetworkManager transition.

PR #6327, opened 2026-07-20
feat: Add Super+N Neovim cheatsheet binding

azzenabidi:
I added a Neovim cheat sheet that gets triggered with Super+N as shown in the attached screenshot

screenshot-2026-07-18_19-17-22


dhh:
Interesting idea, but I don't think we can do cheat sheets for everything. Thanks for the suggestion, though!

PR #6329, opened 2026-07-20
Install Emacs via the omarchy-emacs AUR package

scottjones:
## Summary

- Points the **Emacs** entry in the install-editor menu at a dedicated `omarchy-install-editor-emacs` script, matching the pattern already used for Helix, VSCode, and Zed.
- The script installs the `omarchy-emacs` AUR package and runs its installer, which pulls in `emacs-wayland`, applies the theme/font integration, and enables the user `emacs.service`.
- One new script + one changed menu line. No new dependencies in the Omarchy tree.

## Context

This is the AUR-packaged follow-up to #4807, where @dhh suggested packaging the Emacs integration as an AUR package (like `omarchy-nvim`) rather than shipping it in-tree — thank you for that steer. The package now exists and is maintained:

- AUR: https://aur.archlinux.org/packages/omarchy-emacs
- Source: https://github.com/scottjones/omarchy-emacs

It supersedes #5329, which made the same change against `dev` before the quattro menu refactor landed. I'll close that one in favor of this.

## Behavior

**Before:** `omarchy-install-app Emacs emacs-wayland && systemctl --user enable --now emacs.service` — installs a bare Emacs with no theming.

**After:** installs an Emacs that automatically tracks the active Omarchy theme and font, with the daemon enabled. The AUR package encapsulates setup, so the menu line stays a one-liner consistent with the other setup-based editors (Helix/Zed/VSCode).

## Test plan

- [x] `omarchy-install-editor-emacs` runs cleanly — installs the package, runs setup, enables and starts `emacs.service`
- [x] The daemon comes up with the `omarchy` theme applied, tracking the current theme at `~/.local/state/omarchy/current/`
- [x] `omarchy-menu.jsonc` still parses; the entry launches via `omarchy-launch-floating-terminal-with-presentation` like the other setup-based editors
- [x] Verified on Omarchy 4 (quattro)

### DIFF
--- a/bin/omarchy-install-editor-emacs
+++ b/bin/omarchy-install-editor-emacs
@@ -4,3 +4,6 @@
 
 echo "Installing Emacs..."
 omarchy-pkg-aur-add omarchy-emacs && omarchy-install-emacs
+
+# emacsclient opens a frame on the running daemon, not a second Emacs
+setsid gtk-launch emacsclient
--- a/bin/omarchy-install-editor-emacs
+++ b/bin/omarchy-install-editor-emacs
@@ -1,0 +1,6 @@
+#!/bin/bash
+
+# omarchy:summary=Install Emacs with Omarchy theme and font integration via the omarchy-emacs AUR package
+
+echo "Installing Emacs..."
+omarchy-pkg-aur-add omarchy-emacs && omarchy-install-emacs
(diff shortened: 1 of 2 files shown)


scottjones:
https://github.com/user-attachments/assets/5def0fa7-2f15-4dc6-85bf-ebee690b2ba3

dhh:
Lovely. Let's get omarchy-emacs added to omarchy-pkgs too. So we don't have to rely on AUR. You can open a PR here: https://github.com/omacom-io/omarchy-pkgs. Then we can install using the regular non-AUR installer.

scottjones:
Glad you like it.  I've had so much fun creating it and using Omarchy (I almost decided to try to learn vim instead).  Thank you!

dhh:
I thought there were some setup gymnastics you had to do as well? The ones we yanked from the package? The installer here should get you all the way to a running emacs. Preferably even open it after install, like we do with the other editors.

scottjones:
Nothing functional got removed, just the `pacman` post-install message, which was mostly legacy and you rightly pointed out was ugly. The setup still runs from the installer: `omarchy-install-emacs` calls `omarchy-emacs-setup` (theme + font integration, hooks, generated colors) and enables the `emacs.service` daemon, so it goes all the way to a running, themed Emacs.

And added the open-after-install — it gtk-launches a frame once install finishes, matching the other editors (using `emacsclient` so it opens on the daemon rather than spawning a second instance).

Issue #2382, opened 2025-10-11
Unable to connect to WPA2 Enterprise 802.1X networks with Imapala


HassanSattar3:
### System details

Lenovo Thinkpad T490

### What's wrong?

I’ve been using ohmarchy on my laptop since it runs faster, and I already use arch + hyprland on my main pc. when I go to school though, the wifi there uses WPA2 enterprise 802.1X, and impala can’t connect to it.

I tried using networkmanager, but it’s kinda glitchy and weird. maybe networkmanager should be the default since it works better for school or work networks like this.

pythops:
impala author here. 
There is an open issue  in impala about that one. 
https://github.com/pythops/impala/issues/21

kromsam:
> I tried using networkmanager, but it’s kinda glitchy and weird. maybe networkmanager should be the default since it works better for school or work networks like this.

see: https://github.com/basecamp/omarchy/issues/1414 

and im sorry @pythops , Impala is really nice, but NetworkManager offers all enterprise WiFi and VPN solutions right out of the box... so i proposed a switch using a different tui: https://github.com/basecamp/omarchy/pull/2909

If Impala could also be compatible with NetworkManager... that might be my dream solution

pythops:
@kromsam 
No need to be sorry. People are free to use whatever suit them the best and if impala does not work for you, that's fine for me :)

The work to support NetworkManager has started but it is very slow as I am the only one working on it only on my free time. I guess it will happen at some point but can not give any specific deadline.

pythops:
~~I have a PR for WPA2 enterprise that implement the EAP-TLS protocol, You can give it a try, it should hopefully work for you ( I have tested for me and it works)
I am implementing the rest of the protocols in the upcoming days
https://github.com/pythops/impala/pull/75~~
Edit: Sorry, I forgot that `impala` is gonna be removed from Omarchy, never mind

kromsam:
> Edit: Sorry, I forgot that `impala` is gonna be removed from Omarchy, never mind

haha, don't rush to conclusions. the overlords haven't said anything yet.

dvrnt:
Getting the same issue here. Would love to know what's solution for this.

pythops:
@dvrnt 
I have merged the PR for to the main branch, if you have WPA2 enterprise  that is using `TLS` then it will work for sure( I have tested it myself), for the other protocols, I did not test yet so you can give it try and let me know how it works for you. 
here is the issue where we keep track
[...]

Issue #2351, opened 2025-10-10
8021x support in the wifi TUI


0ncoming-Storm:
### What do you need?

The wifi TUI dose not support connecting to 8021x wifi networks, it would be nice to have it do so in order to connect to enterprise networks easily.

MattyGam3r:
+1 for this, I was attempting to connect to my university Wifi and it wasn't a good experience. I am not too sure however if this belongs on the impala page (where this TUI originally comes from) https://github.com/pythops/impala

pythops:
Impala author here.
There is an open issue already for this one. Hopefully it will be implemented next week

pythops:
@MattyGam3r @0ncoming-Storm 
There is a PR that should fix it https://github.com/pythops/impala/pulls
If you can give it a try that would be helpful 🙏

pythops:
a new version `v0.5.0` has been released with the support of wpa enterprise network.

morgoth:
@0ncoming-Storm can you verify and close the issue?

0ncoming-Storm:
impala works to connect to the network, but it only allows connecting to enterprise networks if run as root. omarch-launch-wifi dose not run impala as root, and waybar uses omarch-launch-wifi

pythops:
To be clear, this is not a choice made by impala, `iwd` stores the network configuration in the directory `/var/lib/iwd` and it is only accessible and writable with `root`

Issue #2076, opened 2025-09-29
How do I connect to a wifi that uses 802.1X ? + Error net.connman.iwd.NotConfigurerd : Not Configured


Kro0noss:
### System details

Intel Core i7 Intel Graphics, Omarchy Latest Version

### What's wrong?

I'm unable to connect to a **Wi-Fi network** that uses **802.1x authentication**. I get the following error:

**Error: net.connman.iwd.NotConfigured : Not Configured**

However, I can connect to Wi-Fi networks that do not use 802.1x.

How can I resolve this issue?

<img width="1251" height="714" alt="Image" src="https://github.com/user-attachments/assets/5f09bc32-16ff-4f38-ae9a-46ec795eb23b" />

beniaminogreen:
Hi. I also had this issue yesterday. I solved it by manually making a config file in `/var/lib/iwd/`  that had settings appropriate for my network. After I did this, I was able to connect to the network using the applet with no issues.  The [arch wiki](https://wiki.archlinux.org/title/Iwd) has a pretty good tutorial on configuring the file.  I was able to get it working in a few minutes after browsing for info about my network's settings online. You can see the configuration file I used [here](https://gist.github.com/beniaminogreen/c6ad806e783774e38b15916ad97fcc4f) if its helpful to you. 

Best of luck,
Ben

Kro0noss:
Thanks ! It Worked!

Kro0noss:
It Works

omarchy-user:
it did not work for me

pythops:
The last release of impala `v0.5.0` solves that
https://github.com/pythops/impala/releases/tag/v0.5.0

PR #6334, opened 2026-07-21
Connect to enterprise (802.1X) Wi-Fi networks from the network panel

KazeTachinuu:
The network panel only supports PSK Wi-Fi, so eduroam and other 802.1X networks can't be joined from the UI (#2382, #2351, #2076); #6321 imports old iwd networks but leaves enterprise ones for manual re-add.

Detect 802.1X networks from their advertised security type and expand the passphrase prompt with an identity field. Connecting creates the NetworkManager profile with nmcli (PEAP/MSCHAPv2) and activates it through the panel's existing tracking, since Quickshell's networking API can't create profiles. PSK behavior is unchanged.

image

### DIFF
--- a/shell/plugins/panels/network/Model.js
+++ b/shell/plugins/panels/network/Model.js
@@ -232,6 +232,14 @@ function isProtected(security, openSecurity) {
   return security !== openSecurity
 }
 
+var enterpriseConnectScript =
+  "u=$(uuidgen);" +
+  " nmcli connection add type wifi con-name \"$1\" ssid \"$1\" connection.uuid \"$u\"" +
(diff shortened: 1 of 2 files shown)


RushiChaganti:
Hey @KazeTachinuu, I’ve implemented this as a migration in my PR (#6321). I think your PR (#6334) is the next step, since you’re working on the UI side of it.

KazeTachinuu:
> Hey @KazeTachinuu, I’ve implemented this as a migration in my PR (#6321). I think your PR (#6334) is the next step, since you’re working on the UI side of it.
> 
> 

I've used your PR in my own migration thanks for the convenience :)

PR #6341, opened 2026-07-21
Expose Dell firmware power profiles

stappmus:
## Summary

Expose Dell firmware power profiles in Setup > Power Profile when the kernel provides a `dell-pc` platform-profile controller.

Non-Dell systems continue using the existing power-profiles-daemon menu.

## Details

- Shows supported Dell modes in lower-to-higher power order: `quiet`, `cool`, `balanced`, `performance`
- Maps `quiet` and `cool` to OS `power-saver`, `balanced` to OS `balanced`, and `performance` to OS `performance`
- Adds `omarchy powerprofiles platform` for safely inspecting and changing kernel platform-profile controllers
- Validates the device and requested profile before writing to sysfs
- Adds a udev rule scoped to `dell-pc`, allowing members of `wheel` to change the profile without repeated authentication

## Migration

Installs the platform-profile udev rule, reloads udev, and applies it to existing platform-profile devices. The same setup is included for fresh installations.

## Testing

- `bin/omarchy commands --check`
- Bash syntax and diff checks
- `udevadm verify`
- Confirmed the udev rule matches `dell-pc` but not the separate SoC Power Slider controller

## Hardware test

Tested on a Dell XPS 16 DA16260:

- `quiet` -> OS `power-saver` / Dell `quiet`
- `cool` -> OS `power-saver` / Dell `cool`
- `balanced` -> OS `balanced` / Dell `balanced`
- `performance` -> OS `performance` / Dell `performance`
- Profile changes work without repeated authentication
- Original profile state restored after testing

### DIFF
--- a/bin/omarchy-menu
+++ b/bin/omarchy-menu
@@ -365,12 +365,41 @@ show_setup_menu() {
 }
 
 show_setup_power_menu() {
-  profile=$(menu "Power Profile" "$(omarchy-powerprofiles-list)" "" "$(powerprofilesctl get)")
+  local dell_device=""
+  local dell_current=""
+  local device name current
+  while IFS=$'\t' read -r device name current; do
+    if [[ $name == "dell-pc" ]]; then
+      dell_device="$device"
+      dell_current="$current"
+      break
+    fi
+  done < <(omarchy-powerprofiles-platform list)
+
+  local profile
+  if [[ -n $dell_device ]]; then
+    local dell_profiles
+    dell_profiles=$(omarchy-powerprofiles-platform choices "$dell_device")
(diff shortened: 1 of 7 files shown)


dhh:
This is a lot of additional complexity. I haven't found any need to use anything but balanced and performance on my fleet of XPS machines. What benefit have you found from the others?

stappmus:
Quiet is useful here because it materially reduces fan noise. Cool is useful when chassis temperature matters; it deliberately trades a little more fan power for lower surface temperature. You were right about the implementation complexity, so I reworked this for Quattro in #6346 using the existing list/set commands and no new public helper. Closing this dev version in favor of that replacement.

PR #1303, opened 2025-08-29
Set hypridle exception for fullscreen apps

JordanAnthonyKing:
https://github.com/basecamp/omarchy/issues/1269
Prevent screensaver/lockscreen from activating during games and such

PR #6342, opened 2026-07-22
Treat gamepad input as idle activity

stappmus:
## Summary

- watch Linux gamepad devices from the built-in Quickshell idle service
- restart the idle countdown for button presses, relative input, and meaningful analog-axis movement
- ignore small analog drift, throttle activity reports, and handle controller hot-plug/disconnect
- add `python-evdev` to the base install plus a migration for existing Quattro users
- expose `omarchy-shell idle activity` as a generic idle-reset entry point

## Why

Follow-up to #1269 and #1303. The fullscreen rule prevents idle for known fullscreen apps, but controller-only play can still trigger the screensaver in emulators and other games that do not match an app-specific rule. Treating actual controller input as activity works independently of the active application or window mode.

## Testing

- `PYTHONDONTWRITEBYTECODE=1 bash test/shell.d/idle-test.sh`
- `PYTHONDONTWRITEBYTECODE=1 ./test/cli`
- `qmllint shell/plugins/services/idle/Service.qml`
- Python syntax compile, migration `bash -n`, and `git diff --check`
- evdev/UInput integration check: controller attached after watcher startup, small axis drift ignored, deliberate axis motion reported
- evdev/UInput integration check: gamepad button press reported

The repository-wide shell runner also reaches unrelated local-environment failures where Quickshell and the sibling `omarchy-pkgs` checkout are unavailable; the focused idle suite passes.

### DIFF
--- a/docs/omarchy-shell.md
+++ b/docs/omarchy-shell.md
@@ -127,6 +127,9 @@ Rules:
 7. `idle.screensaver` and `idle.lock` are seconds since user idle began.
 8. `version: 1` is required.
 
+The built-in idle service treats gamepad button presses and meaningful axis
+movement as user activity. Small analog-axis drift is ignored.
+
 `config/omarchy/shell.json` describes the fresh-install state. When no
 user `shell.json` exists, defaults are used verbatim. Once the user
 customizes, `shell.json` is canonical — there is no deep-merge.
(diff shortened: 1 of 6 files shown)


stappmus:
Superseded by #6352, rebuilt on the current quattro head with both review threads addressed and revalidated on the Quickshell layout.

PR #6344, opened 2026-07-22
Fix unusable system when non-Latin keyboard layout selected at install

### LINKED ISSUE #6085: Choosing non-English layout in installation makes the system unusable
### System details

Omarchy 3.8.2

### What's wrong?

I tried to install Omarchy on a new machine, and during installation chose a non-English keyboard layout (Hebrew) as my default keyboard layout, assuming it would save me the trouble of editing `input.conf` and setting everything up. Turns out that if you do that you wont have an English layout at all to work with, and therefore not be able to properly edit the config in the first place to add English. Many keybinds (such as closing windows) wont work either, meaning your system is essentially left in a non-operational state after choosing such a layout.

Many of Omarchy's keybinds are defined in terms of English letters (keysyms):
```hypr
# Close windows
bindd = SUPER, W, Close window, killactive,
bindd = CTRL ALT, DELETE, Close all windows, exec, omarchy-hyprland-window-close-all

# Control tiling
bindd = SUPER, J, Toggle window split, layoutmsg, togglesplit
bindd = SUPER, P, Pseudo window, pseudo, # dwindle
bindd = SUPER, T, Toggle window floating/tiling, togglefloating,
bindd = SUPER, F, Full screen, fullscreen, 0
bindd = SUPER CTRL, F, Tiled full screen, fullscreenstate, 0 2
bindd = SUPER ALT, F, Full width, fullscreen, 1
bindd = SUPER, O, Pop window out (float & pin), exec, omarchy-hyprland-window-pop
bindd = SUPER, L, Toggle workspace layout, exec, omarchy-hyprland-workspace-layout-toggle
```
(`.local/share/omarchy/default/hypr/bindings/tiling-v2.conf`)
Which means all of these fail to work on non-English layouts.

Suggestions to fix:
1. Treat the installation option of choosing a keyboard layout as an optional secondary option, where English is the default and always installed. If you choose a secondary layout, automatically enable `grp:...` in `input.conf` to allow switching between them. This would ensure the system is always in a working state after installation.
2. Change all keysyms (`SUPER, J, ...`) to keycodes (`SUPER, code:44, ...`) which are layout-agnostic and would ensure the basic keybindings work even on different non-English layouts. These codes could be found using the `wev` command. Seeing as how many keybinds already use this `code:...` form, I think it would also be more consistent to do that.

husamemadH:
Fixes #6085.

Choosing a non-Latin keyboard layout during Arch install (e.g., `ara`, `il`, `ru`) leaves the system unusable in Hyprland — the default bindings use Latin keysyms (`SUPER+W`, `SUPER+F`, etc.) that don't exist in those layouts, so shortcuts do nothing.

`default/hypr/input.lua` now prepends `us` to `kb_layout` when the detected layout doesn't already include it, and appends `grp:alts_toggle` to `kb_options` so the added layout can be reached with Left Alt + Right Alt.

Result for a user whose `XKBLAYOUT=ara`:

    kb_layout  = "us,ara"
    kb_variant = ",<their variant>"
    kb_options = "compose:caps,shift:both_capslock,grp:alts_toggle"

**Why `us` first, not the user's layout first**: Hyprland matches keybindings against the first layout in `kb_layout` only, not the currently active one. Verified on Hyprland 0.55.2 — `us,ara` fires `SUPER+W` regardless of active layout; `ara,us` breaks it in both.

Idempotent: if the layout is already `us` or already contains `us`, no changes are made.

### DIFF
--- a/default/hypr/input.lua
+++ b/default/hypr/input.lua
@@ -23,12 +23,24 @@ end
 
 local vconsole = read_vconsole()
 
+local layout = vconsole.XKBLAYOUT or "us"
+local variant = vconsole.XKBVARIANT or ""
+local kb_options = "compose:caps,shift:both_capslock"
+
+-- Prepend 'us' so Omarchy's Latin-keysym bindings resolve — Hyprland matches
+-- bindings against the first layout in kb_layout, not the currently active one.
+if not (","..layout..","):find(",us,", 1, true) then
+  layout = "us," .. layout
+  variant = "," .. variant
+  kb_options = kb_options .. ",grp:alts_toggle"
+end
+
 hl.config({
   input = {
-    kb_layout = vconsole.XKBLAYOUT or "us",
-    kb_variant = vconsole.XKBVARIANT or "",
+    kb_layout = layout,
+    kb_variant = variant,
     kb_model = "",
-    kb_options = "compose:caps,shift:both_capslock",
+    kb_options = kb_options,
     kb_rules = "",
     follow_mouse = 1,
     sensitivity = 0,
--- a/default/hypr/input.lua
+++ b/default/hypr/input.lua
@@ -21,24 +21,32 @@ local function read_vconsole()
   return values
 end
 
+-- Layouts that can't type Latin letters. Keep in sync with the list in
+-- etc/mkinitcpio.conf.d/omarchy_hooks.conf.
+local non_latin_layouts =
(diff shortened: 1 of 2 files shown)


husamemadH:
I've also tested this with the latest version of omarchy quattro installed that has version 0.56.0 of hyprland and its still showing the same buggy behaviour

PR #6346, opened 2026-07-22
Expose Dell firmware profiles in Quickshell power menus

stappmus:
## Summary

- expose supported `dell-pc` firmware modes as `quiet`, `cool`, `balanced`, and `performance` in the Quickshell power panel and menu provider
- pair `quiet` and `cool` with OS `power-saver`, while `balanced` and `performance` use their matching OS profiles
- preserve the existing power-profiles-daemon controls when no writable Dell controller is available
- add a Dell-scoped udev rule so `wheel` users can switch modes without repeated authentication
- use a 2x2 picker when four modes are present, while retaining the existing single-row layout for three modes

This reworks #6341 for the `quattro` branch without adding another public helper command.

## Why

Dell's firmware thermal modes are separate from the OS power profile. `quiet` materially reduces fan noise, while `cool` uses a more aggressive fan policy to keep the chassis cooler. Pairing both layers gives every menu choice a predictable result.

## Profiles

| Menu choice | OS profile | Dell firmware profile |
| --- | --- | --- |
| `quiet` | `power-saver` | `quiet` |
| `cool` | `power-saver` | `cool` |
| `balanced` | `balanced` | `balanced` |
| `performance` | `performance` | `performance` |

The active marker requires both layers to match. Existing saved `power-saver` preferences become `quiet` when Dell modes activate.

On the test machine, power-profiles-daemon can step from `performance` through `balanced` before reaching `power-saver` because two platform-profile controllers are present. The setter verifies convergence with a bounded retry and restores the previous OS profile if convergence or the firmware write fails.

## Testing

- `./test/cli`
- `bash test/shell.d/powerprofiles-dell-test.sh`
- `bash test/shell.d/powerprofiles-set-test.sh`
- `bash test/shell.d/power-test.sh`
- Bash syntax and `git diff --check`
- `udevadm verify --no-style etc/udev/rules.d/99-omarchy-dell-platform-profile.rules`
- verified the rule matches `dell-pc`, not the separate SoC Power Slider controller
- verified all four mappings without authentication prompts on a Dell XPS 16 DA16260
- verified keyboard selection and the running 2x2 Quickshell layout; restored `quiet` afterward

`./test/shell` reaches the existing `bin-style-test.sh` failure for unchanged `bin/omarchy-display-text-size`; the focused shell tests above pass.

### DIFF
--- a/bin/omarchy-powerprofiles-list
+++ b/bin/omarchy-powerprofiles-list
@@ -3,13 +3,58 @@
 # omarchy:summary=Returns a list of all the available power profiles on the system.
 # omarchy:args=[--active-state]
 
-if [[ ${1:-} != "" && ${1:-} != "--active-state" ]]; then
+option="${1:-}"
+
+if [[ $option != "" && $option != "--active-state" ]]; then
   echo "Usage: omarchy-powerprofiles-list [--active-state]" >&2
   exit 1
 fi
 
-powerprofilesctl list 2>/dev/null |
-  awk -v with_state="${1:+1}" '/^\s*[* ]\s*[a-zA-Z0-9-]+:$/ {
+os_profile_output=$(powerprofilesctl list 2>/dev/null)
+
+os_profile_available() {
+  grep -Eq "^[*[:space:]]+$1:$" <<<"$os_profile_output"
+}
+
+dell_profile_dir=""
+profile_root="${OMARCHY_PLATFORM_PROFILE_ROOT:-/sys/class/platform-profile}"
+for profile_dir in "$profile_root"/*; do
+  [[ -r $profile_dir/name && -r $profile_dir/choices && -r $profile_dir/profile && -w $profile_dir/profile ]] || continue
+  if [[ $(<"$profile_dir/name") == "dell-pc" ]]; then
+    dell_profile_dir="$profile_dir"
+    break
+  fi
+done
+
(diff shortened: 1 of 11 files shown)


stappmus:
DHH asked in the previous PR: This is a lot of additional complexity. I haven't found any need to use anything but balanced and performance on my fleet of XPS machines. What benefit have you found from the others?

For me the main improvement is that quiet mode stops fans from spinning at all during most usage. Very nice for when reading etc late at night when its quiet. Prevents me from thinking "what did i leave running now" and checking btop. Longterm when thermal paste gets worse and the computer needs some service i think a lot of dell users would appreciate the quiet mode being available.

dhh:
Appreciate the suggestion here, but this is too much code for a brand-specific niche improvement imo. Would be nice to have it as a separate package that you could install via AUR. Thanks for working on it ✌️

stappmus:
Thanks for the direction — I pulled the Dell-specific behavior into a standalone package instead of keeping it in Omarchy core:

https://github.com/stappmus/omarchy-dell-power-profiles

It now contains the packaged Dell/power-profiles-daemon provider plus an optional Quattro bar widget, with the Arch `PKGBUILD`, udev rule, tests, and removal instructions kept in that repository. I also submitted it to the Omarchy Plugins marketplace for review:

https://github.com/HANCORE-linux/omarchy-plugin-marketplace/issues/8

PR #6352, opened 2026-07-23
Treat gamepad input as Quickshell idle activity

stappmus:
## Summary

- watch Linux gamepad devices from Omarchy's built-in Quickshell idle service
- restart the idle countdown for button presses, relative input, and meaningful analog-axis movement
- ignore button releases and small analog drift, throttle activity reports, and handle hot-plug/disconnect
- add `python-evdev` to the base install plus a migration for existing Quattro users
- expose `omarchy-shell idle activity` as a generic idle-reset entry point

## Why

Follow-up to #1269 and #1303. Fullscreen rules help known applications, but controller-only play can still trigger the screensaver in emulators and games that do not match an app-specific rule. Treating actual controller input as activity works independently of the active application and window mode.

This supersedes #6342. It is rebuilt on the current `quattro` head and addresses both review threads from that PR by making watched-device set conversion explicit and ignoring `EV_KEY` release events.

## Testing

- `PYTHONDONTWRITEBYTECODE=1 bash test/shell.d/idle-test.sh`
- `PYTHONDONTWRITEBYTECODE=1 ./test/cli`
- `PYTHONDONTWRITEBYTECODE=1 ./test/shell` (passes through `snapper-test.sh`; that existing test stops because the sibling `omarchy-iso` checkout is unavailable)
- all 16 shell test files after `snapper-test.sh` run individually: pass
- `qmllint shell/plugins/services/idle/Service.qml`
- Python syntax compile, migration `bash -n`, and `git diff --check`
- evdev/UInput integration: controller hot-plugged after watcher startup, release ignored, button press reported

### DIFF
--- a/docs/omarchy-shell.md
+++ b/docs/omarchy-shell.md
@@ -127,6 +127,9 @@ Rules:
 7. `idle.screensaver` and `idle.lock` are seconds since user idle began.
 8. `version: 1` is required.
 
+The built-in idle service treats gamepad button presses and meaningful axis
+movement as user activity. Small analog-axis drift is ignored.
+
 `config/omarchy/shell.json` describes the fresh-install state. When no
 user `shell.json` exists, defaults are used verbatim. Once the user
 customizes, `shell.json` is canonical — there is no deep-merge.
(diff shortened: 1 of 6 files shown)


dhh:
Appreciate the stab at this, but I think it's too much complexity. We should rather just mark all the gaming apps as idle inhibitors. And then we have Stay Awake as fallback. Fell free to look at the first bit and open a PR for it. Thanks!

PR #6358, opened 2026-07-23
Add omarchy keyboard layout bar widget

heyssh:
# Omarchy Keyboard Layout
 
A keyboard layout bar widget for Omarchy. Shows your active
keyboard language in the bar, and gives you a clean panel to add,
remove, and switch between languages — plus configure the shortcut
used to cycle between them, all without leaving the keyboard.
 
screenshot-2026-07-29_19-14-31

 
## Features
 
- **Live language indicator** in the bar, showing the active layout
  at a glance (e.g. `EN`, `FA`, `DE`) — even with only one language
  configured, so you always know what layout is active without
  opening the panel.
- **One-click switching** — open the panel and pick any configured
  language, or right-click the bar icon to instantly cycle to the
  next one.
- **Add languages** from a clean, searchable list of real system
  keyboard layouts — type to filter by name or code.
- **Remove languages** you no longer need, with the system's primary
  layout always kept safe from accidental removal.
- **Switch-shortcut picker** — choose how you cycle layouts system-wide:
  - `Alt+Shift`
  - `Ctrl+Shift`
  - `Right Alt`
- **Stays in sync with Hyprland** — if you switch layouts using the
  keyboard shortcut directly, the bar updates automatically.
- **Full keyboard navigation** inside the panel — no mouse required.
- **Setup > Show Keyboard Layout** — a menu toggle for anyone who
  only ever uses one language and doesn't want the icon sitting in
  the bar. Toggle it off any time, and back on the same way whenever
  you want it back. An OSD notification confirms the new state.
 
## How it works
 
The widget is made up of three parts that work together:
 
| Part                      | Role                                             |
|----------------------------|--------------------------------------------------|
| `bin/omarchy-keyboard-layout` | CLI that reads/writes the layout state and talks to Hyprland |
| `input/input.lua`             | Hyprland input config — applies the active layout(s) and switch shortcut |
| `keyboard/` (Panel.qml, Model.js, manifest.json) | The bar widget UI itself |
 
## Usage
 
Click the keyboard icon in the bar to open the panel:
 
- **Languages** — tap a language to switch to it, or hit the `+` to
  add a new one from the full list.
- **Switch shortcut** — tap any of the four presets to change how the
  keyboard shortcut cycles your layouts.
- **Right-click** the bar icon to instantly cycle to your next
  configured language.
- **Setup > Keyboard Layout** — toggle the bar icon on or off.
  Handy if you only work in one language and prefer a cleaner bar;
  turning it back on brings the icon right back.
- **Add a keybinding** (SUPER+CTRL+K) to open the panel via a new `open`
subcommand, which checks show_bar_icon itself so the panel only
opens when its bar icon is enabled/visible.

## Test
https://github.com/user-attachments/assets/1509dda3-0d22-444e-89d2-4e5cc9c057cd


husamemadH:
i like this addition i spent alot of time trying to make something like this today

heyssh:
> i like this addition i spent alot of time trying to make something like this today

Yeah, I think this is something that's genuinely needed. For anyone who regularly works with two keyboard layouts, having a clear indicator is almost essential. It makes switching between languages much more seamless and helps avoid a lot of unnecessary typing mistakes.

pomartel:
Maybe this could be released as a plugin instead? It's now easy to put it in its own repo and people who use multiple keyboard layouts can install it with the `omarchy plugin add` command.

heyssh:
> Maybe this could be released as a plugin instead? It's now easy to put it in its own repo and people who use multiple keyboard layouts can install it with the `omarchy plugin add` command.

Appreciate the idea! My concern is that as a standalone plugin, most people who'd actually want this simply won't come across it. I think it fits better in core, since it's something a lot of users would want available by default.

On a related note, I also built a calendar for Omarchy — that one's available as a plugin.

dpan:
I haven't tried this first hand so apologies if I'm getting it wrong, but looking at screenshots, it feels odd. The bar should be really simple, just switching the language (maybe with an OSD showing the selected language when hitting the binding). Adding languages, selecting the keyboard bindings belong in the settings, since they don't change that often, they do not need to be so immediately accessible. Just my two cents.

heyssh:
> I haven't tried this first hand so apologies if I'm getting it wrong, but looking at screenshots, it feels odd. The bar should be really simple, just switching the language (maybe with an OSD showing the selected language when hitting the binding). Adding languages, selecting the keyboard bindings belong in the settings, since they don't change that often, they do not need to be so immediately accessible. Just my two cents.

Thanks for the feedback! I'd actually lean toward keeping it as one panel. Switching is already the first thing you see (click any language or use arrow keys, no digging needed) — add/shortcut are secondary and further down, so they're not really competing for attention. Splitting them out would add a navigation step for something that's currently zero clicks away.

Also tried to account for the single-language case — if someone only uses one language and doesn't want the icon taking up space in the bar, there's a toggle for that under Setup > Keyboard Layout.

Happy to revisit if real usage shows otherwise.

heyssh:
I want to add to omarchyplugins.

PR #6361, opened 2026-07-24
Never enable Wi-Fi power save on Intel BE200/BE211

jriff:
The BE-series (`iwlmld`) firmware chronically misses beacons when power save is enabled — `missed beacons exceeds threshold` followed by deauth/reconnect loops. Since `99-wifi-powersave.rules` enables power save on every unplug, laptops with these cards (including the XPS 14/16 Panther Lake) drop Wi-Fi constantly on battery.

Measured on an X1 Carbon Gen 14 (BE211) across two days and two mesh setups: 6–52 disconnects/hour with power save on depending on signal, ~zero with it off — and no measurable battery-draw difference, since the reconnect churn eats the savings.

Fix: guard the `on` path by PCI ID, same detection as `fix-wifi7-eht.sh`. Remove when the BE-series power-save path is fixed in linux-firmware.

### DIFF
--- a/bin/omarchy-wifi-powersave
+++ b/bin/omarchy-wifi-powersave
@@ -3,6 +3,14 @@
 # omarchy:summary=Set Wi-Fi power save mode on wireless interfaces
 # omarchy:args=<on|off>
 
+# Never enable power save on Intel BE200/BE211 (iwlmld): the firmware misses
+# beacons in power save and drops the link in constant reconnect loops.
+# Same detection as fix-wifi7-eht.sh.
+# Remove when the BE-series power-save path is fixed in linux-firmware.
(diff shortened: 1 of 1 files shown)


dhh:
Thanks for the careful field data here — the two-day disconnect measurements are what convinced us the whole feature was upside-down. Your numbers showed power save buying nothing on these cards while the reconnect churn ate the savings, and once we ran the math for the healthy cards too, the answer was to stop toggling power save entirely rather than deny-list broken firmware: #6368 removes the udev toggle and ships a NetworkManager conf.d that keeps power save off everywhere, on every card. The broken BE-series path is simply never entered anymore, with nothing to remember to remove when linux-firmware gets fixed.

Closing in favor of that — your BE211 will be covered on the next update. Appreciated!

PR #6231, opened 2026-07-17
Omarchy Quattro

dhh:
# Omarchy Quattro: Release Candidate

Omarchy 4 aka Quattro is the biggest release since the project started. The entire desktop shell has been reimagined in [Quickshell](https://quickshell.org): the bar, launcher, menus, notifications, on-screen displays, control panels, lock screen, and polkit agent now all live inside a single long-running shell process with a plugin architecture. That means Waybar, Walker, Mako, SwayOSD, hyprlock, hypridle, swaybg, and polkit-gnome are all gone, replaced by one coherent, fully-themed, IPC-scriptable shell.

You can upgrade an existing Omarchy installation to Quattro by first `Update > Omarchy`, then `Update > Channel > RC`, then `Update > Omarchy to Quattro`. Just remember to `Update > Channel > Stable` after the final release!

Or install it fresh on a secondary machine using the ISO: https://iso.omarchy.org/omarchy-4.0.0.rc4-2.iso

Always take a backup of important data!

screenshot-2026-08-13_17-03-14-medium

## Headline Features

- Reimagine the entire desktop shell in [Quickshell](https://quickshell.org/).
- Move Omarchy internals from git to system packages for safely separating user modifications.
- Offer dual-boot installation when free space is available (remember to turn off Bitlocker for Windows!).
- Setup a machine for a new owner during install, so OEMs and gift givers can prepare a machine for others.
- Allow for factory reset (`Setup > Reset Computer`) from new Quattro installs for safe resale/gifting of your machine.
- Shrink the ISO by over a gigabyte (now under 6GB!).
- Speed-up installation by +30% (sub-minute installs now possible!).
- Convert all Hyprland configs to lua for full 0.56 compatibility.
- Expand theme coloring from base 8 to 24 so btop/nvim/vscode themes can be autogenerated.
- Use networkmanager for new network panel for greater compatibility.
- Add nested search to Omarchy menu and merge the launcher into that one unified menu (`Super + Space`).
- Add text scaling across the shell, GTK, and terminals in one move via the Display panel.
- Add speaker tunings for 2026 XPS 14/16 laptops for dramatically better sound.
- Add menu bar repositioning (top/bottom/left/right) and transparency.
- Add menu bar panels for Tailscale + Dropbox.
- Add weather panel with configurable location.
- Add plugin system and [full ecosystem](https://omarchyplugins.com/).
- Add Omawrite, Omacut, and Omacalc as default applications.
[...]

PR #6362, opened 2026-07-24
Add scope flags to omarchy-display-text-size

jzetterman:
Follow-up to my comment on #6231: the unified text-size knob is a great one-touch default, but it hard-links three surfaces (shell font, GTK `text-scaling-factor`, terminal font size) that don't always want to move together, and there's no way to express "just the shell." Concrete case: a 4K display at scale 1 where bumping the shell also inflated every Chromium/Electron app and silently rewrote a hand-tuned `font-size` in the Ghostty config.

This adds composable scope flags to `omarchy-display-text-size`:

```
omarchy display text size --shell 14        # shell only
omarchy display text size --gtk reset       # GTK scaling only
omarchy display text size --shell --terminals 16
```

**With no flags, behavior is exactly today's** — all three surfaces together, same values, same reload signals. The display-settings slider and the XPS 13 hardware script call the CLI unflagged and are untouched; passing all three flags is identical to passing none. Status output is unchanged.

While making argv strict (flags require exactly one action; unknown tokens error instead of being silently ignored), the validation also picks up fixes for three latent bypasses in the current size check, where bash quirks let invalid tokens through the range guard and into the configs:

- `08`/`09` — octal arithmetic error → accepted (writes an out-of-range size)
- fullwidth digits (`１６`) — locale-collated regex ranges match them → accepted (writes `text-scaling-factor 0`)
- 64-bit wraparound (`18446744073709551625` → 9) → accepted (writes the huge literal)

All three now get the standard size error with nothing written.

Tests: `test/shell.d/display-text-size-test.sh` — 11 cases covering the unflagged regression paths (status/set/reset, reload signals), per-flag isolation (unscoped surfaces byte-identical, no signals or notifications), scoped reset including the `default` alias, flag/action ordering, help precedence, and validation (including one guard per bypass fix above). Everything runs in a sandboxed `$HOME`/`$XDG_RUNTIME_DIR` with stubbed `gsettings`/`pkill`/`pgrep`/`omarchy-notification-send`, following the existing suite's conventions.

### DIFF
--- a/bin/omarchy-display-text-size
+++ b/bin/omarchy-display-text-size
@@ -1,14 +1,15 @@
 #!/bin/bash
 
 # omarchy:summary=Scale text everywhere — omarchy shell, GTK apps, and terminals
-# omarchy:args=[size|reset]
-# omarchy:examples=omarchy display text size | omarchy display text size 16 | omarchy display text size reset
+# omarchy:args=[--shell] [--gtk] [--terminals] [size|reset]
+# omarchy:examples=omarchy display text size | omarchy display text size 16 | omarchy display text size reset | omarchy display text size --shell 14
 
 # One knob for apparent text size across the desktop. It drives three settings
-# in lockstep, all anchored to the shell default of 12px:
+# in lockstep by default, all anchored to the shell default of 12px:
 #   • the omarchy shell's font base-size (~/.config/omarchy/shell.toml [font])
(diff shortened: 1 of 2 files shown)


jzetterman:
Superseded by #6550 — same scope flags, plus the Display panel disclosure UI that uses them, rebased onto current `quattro`.

PR #6363, opened 2026-07-24
Keep monitors flush when changing scale

scottjones:
Changing a monitor's scale from the Display panel re-applied only the focused monitor with `position="auto"`, which lets Hyprland re-arrange every display — on multi-monitor setups a monitor can jump to the far side of the layout.

This re-lays out all enabled monitors together: the focused monitor stays put and its neighbours shift by the change in its logical size, so the arrangement is preserved with no gaps, overlaps, or reordering. Over-constrained 2D arrangements (where no flush result exists) fall back to the old `position="auto"`. Logical sizes are computed on Hyprland's 1/120 scale grid so adjacent monitors stay flush with no 1px seams.

Scope note: only the runtime apply changes; persistence is unchanged.

### DIFF
--- a/bin/omarchy-hyprland-monitor-scaling
+++ b/bin/omarchy-hyprland-monitor-scaling
@@ -71,6 +71,44 @@ normalize_scale() {
   awk 'NR == 1 { printf "%g\n", $0 }'
 }
 
+# hl.monitor() statements (for `hyprctl eval`) that scale the focused monitor ($1)
+# to $2 and keep the rest flush: it stays put, right-neighbours shift by its
+# logical-width change, below-neighbours by its logical-height change. Logical
+# sizes use the 1/120 grid (px*120/round(scale*120)) to match Hyprland exactly.
+# If the layout would overlap (over-constrained 2D), emit nothing so set_scale
+# falls back to position="auto".
+build_layout_batch() {
+  local focus="$1" new_scale="$2"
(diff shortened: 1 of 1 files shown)


dhh:
Thanks — the diagnosis is right, and the demo convinced me the annoyance is real. I've taken the SCALE-target label into quattro as 7faf53c4: nice touch, stands on its own.

But I'm passing on the layout half, because it's compositor work being done in userspace. The jq has to re-derive logical geometry, the 1/120 scale grid, transforms, and mirror state that Hyprland holds authoritatively — that's why it needs an overlap detector and fallback paths at all (and testing it turned up a wrong-layout case with a rotated focused monitor, which is exactly the kind of drift you get when two models of the world disagree). The real defect is that re-applying a monitor rule lets Hyprland re-pack auto-positioned monitors, so a scale-only change can teleport unrelated displays. That fix belongs in the compositor, where the reflow is exact, atomic, and covers every scale-change path — not just this script.

If you want to carry this forward, an upstream issue/PR against Hyprland's monitor arrangement is the move, and your write-up here plus the edge cases (rotation, mirrors, over-constrained 2D) is most of the spec already. Closing this one, and thanks again for the careful work — it made the right call here easy to see.

scottjones:
Sounds good...  I noticed the conflicts and started rebasing/fixing this up more...  Will look at taking upstream.

PR #6364, opened 2026-07-24
Run the screensaver on every monitor, not just the last one

scottjones:
`omarchy-launch-screensaver` loops over the monitors, focusing each and spawning a terminal on it — but the spawn is asynchronous. Slow-starting terminals mapped their windows after focus had already moved to the next monitor, so every screensaver landed on the last-focused monitor and the others got none (and the fullscreen rule applied inconsistently when two mapped at once).

This waits for each monitor's screensaver window to appear before moving focus to the next one, so there's exactly one per monitor. Verified on a dual-monitor setup (4K + ultrawide): one fullscreen screensaver on each.


pomartel:
Same issue for here!

PR #6373, opened 2026-07-25
Render clock weekday and month names in the system locale

### LINKED ISSUE #6360: [Quattro] omarchy.clock keeps weekday and month names in English after changing the system locale
### System details

 Omarchy: 4.0.0.r1046.gd570d99-1 (Quattro)   CPU: Intel Core i3-4030U @ 1.90 GHz   GPU: Intel Haswell-ULT Integrated Graphics   System locale: LANG=es_MX.UTF-8   Keyboard layout: US

### What's wrong?

Image

  I am testing Omarchy Quattro and noticed that the `omarchy.clock`
  widget does not follow the system locale for weekday and month names.

  After changing the system locale from `en_US.UTF-8` to
  `es_MX.UTF-8`, the rest of the system switched to Spanish correctly.
  The clock accepted a 12-hour format, but weekday and month names
  produced by `dddd` and `MMMM` are still displayed in English.

  ### Steps to reproduce

  1. Generate the `es_MX.UTF-8` locale.
  2. Set the system locale to `LANG=es_MX.UTF-8`.
  3. Log out or reboot.
  4. Configure the `omarchy.clock` entry in
     `~/.config/omarchy/shell.json` with a format containing `dddd`
     or `MMMM`.
  5. Observe the weekday or month displayed by the clock.

  Example formats:

  ```text
  dddd h:mm AP
  d MMMM 'W'ww yyyy
  ```

  ### Actual behavior

  The system and supported applications use Spanish, but the clock
  continues to display weekday and month names in English.

  ### Expected behavior

  The clock should use the current system locale, so `dddd` and `MMMM`
  display Spanish weekday and month names when the locale is
  `es_MX.UTF-8`.

  ### Relevant implementation detail

  At the time of testing, `Clock.qml` formats the value with:

  ```qml
  Qt.formatDateTime(
    date,
    activeFormat.replace(/ww/g, isoWeekLiteral(date))
  )
  ```

  Using an explicit locale appears to produce locale-aware formatting:

  ```qml
  Qt.locale().toString(
    date,
    activeFormat.replace(/ww/g, isoWeekLiteral(date))
  )
  ```

husamemadH:
Fixes #6360.

`omarchy.clock` renders `dddd` and `MMMM` in English however the session locale is set. `Clock.qml` formats through `Qt.formatDateTime(date, format)`, and given a *format string* Qt formats names in the C locale by contract — [`QDate::toString()`](https://doc.qt.io/qt-6/qdate.html) documents "Day and month names are given in English (C locale)" and points at `QLocale` for localized ones. No `LANG` or `LC_TIME` value can reach it.

`Qt.locale()` is the system locale and accepts the same format expressions:

```diff
-    return Qt.formatDateTime(date, activeFormat.replace(/ww/g, isoWeekLiteral(date)))
+    return Qt.locale().toString(date, activeFormat.replace(/ww/g, isoWeekLiteral(date)))
```

The no-argument form is deliberate: it resolves `LC_TIME` independently of `LANG`, which `Qt.locale(Qt.locale().name)` would drop — under `LANG=en_US.UTF-8 LC_TIME=es_MX.UTF-8` the named form still renders `Saturday` where the bare form renders `sábado`. `Qt.formatDateTime` has no locale-taking overload.

With the shipped defaults, `es_MX` goes from `Saturday 16:42` / `25 July W30 2026` to `sábado 16:42` / `25 julio W30 2026`.

Verified in a live bar under `es_MX` and `ar_EG`.

- Locales with their own numerals now render digits natively, so `ar_EG` reads `٢٥ يوليو W30 ٢٠٢٦`. The ISO week is substituted into the format string as a literal before formatting, so it stays Latin while everything around it localizes. Qt offers no way to keep Latin digits throughout — it ignores both `ar-EG-u-nu-latn` and `ar_EG@numbers=latn` — so the choice is native digits as here, or localizing the week literal too so at least it is internally consistent. Happy to add the latter.

### DIFF
--- a/shell/plugins/bar/widgets/Clock.qml
+++ b/shell/plugins/bar/widgets/Clock.qml
@@ -35,7 +35,7 @@ BarWidget {
   }
 
   function formatted(date) {
-    return Qt.formatDateTime(date, activeFormat.replace(/ww/g, isoWeekLiteral(date)))
+    return Qt.locale().toString(date, activeFormat.replace(/ww/g, isoWeekLiteral(date)))
   }
 
   implicitWidth: button.implicitWidth


dhh:
Hmm, what I find odd here is that everything else in the interface remains in English. So it's an odd mix just to have month names and days in the locale language. If we're going to do this, I think we need to support locales throughout the entire OS. And that's a big understaking (but feel free to explore it).

husamemadH:
The thing is, even if we managed to find a solution to localizing the whole OS, we would need contributors who are willing to help with localization for every new feature that gets added.

PR #6385, opened 2026-07-26
Search apps only when the menu query starts with a capital letter

28allday:
Typing an app name in the menu currently mixes the app in with same-label system rows — `steam` surfaces **Install > Steam** and **Remove > Steam** alongside (and, due to the depth/order tie-break, above) the Steam app itself.

This adds a small, discoverable convention borrowed from case-sensitive smartcase searching: **a query whose first letter is capitalized searches apps only.**

- `Steam` → just the app, like a launcher
- `steam` → everything, exactly as today (matching and ranking untouched)
- Queries starting with a digit or symbol behave like lowercase
- Composes with multi-term queries; only the first character of the trimmed query is inspected

The diff is one helper plus a single guard line in `matchesQuery` (and its test export). Verified live on omarchy 4.0.0.r1373 by running a side-by-side clone of the menu plugin, plus node tests against the CommonJS export covering the cases above and regressions that lowercase behaviour is unchanged.

### DIFF
--- a/shell/plugins/menu/MenuModel.js
+++ b/shell/plugins/menu/MenuModel.js
@@ -277,9 +277,20 @@ function descriptionTextMatches(query, text) {
   return true
 }
 
+// A query whose first character is uppercase ("Steam") searches apps only;
+// lowercase ("steam") searches everything. Non-letter leading characters
+// behave like lowercase.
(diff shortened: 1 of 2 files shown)


dhh:
Fixed via https://github.com/basecamp/omarchy/pull/6383

PR #6388, opened 2026-07-27
Fix ASUS ExpertBook B9406 touchpad quirk never being applied

dhh:
Ports #6093 to `quattro`, with the fix mechanism changed. Original diagnosis and hardware verification by @mijuny.

On a fresh install of an ASUS ExpertBook B9406 the touchpad is dead: clicks register but the cursor never moves. The quirk from #5435 that masks the pad's bogus pressure axes never took effect.

## Why it never applied

libinput `scandir`s `/usr/share/libinput` for `*.quirks` and *additionally* reads the single file `/etc/libinput/local-overrides.quirks`. It never scans `/etc`. The quirk was written to `/etc/libinput/asus-expertbook-b9406.quirks`, so it was silently ignored.

Verified against the installed libinput rather than taken from docs: built a probe against `libinput.so`, pointed `LIBINPUT_QUIRKS_DIR` at a copy of the data dir, and dropped in a deliberately malformed file under a name libinput does not ship. libinput failed quirks loading — proving it reads any `*.quirks` in that directory, not a fixed list.

## Drop-in rather than appending to local-overrides

#6093 appends to `local-overrides.quirks`. That works, but it is the one file users edit by hand, and a shared file can only be updated idempotently by grepping for a section we already wrote — which silently pins the first version we ever shipped. If the quirk ever needs correcting, the grep matches the stale section and the fix never lands.

A drop-in at `/usr/share/libinput/99-omarchy-asus-b9406-touchpad.quirks` is a file we own outright, rewritten wholesale on every run. Consistent with how `fix-z13-touchpad.sh` owns its udev rule, and with the `/usr/share/fontconfig/conf.avail`, `/usr/share/xdg-terminal-exec` and `/usr/lib/chromium` drop-ins already shipped elsewhere.

## MatchUdevType

Dropped, as in #6093: the pad's event node is tagged `ID_INPUT_MOUSE`, so a `touchpad` constraint skips the section. Bus, vendor, product and DMI already pin the device exactly, so the widened match is inert on any other hardware.

Worth flagging honestly: this claim sits uneasily with the original diagnosis, since the `"Touch jump detected and discarded"` message comes from libinput's touchpad backend, which only runs on devices tagged as touchpads. Most likely the pad exposes several event nodes. The reporter's confirmation also bundled both changes, so the evidence clearly supports the path fix and is only suggestive on this one. Kept because it costs nothing if unnecessary. Full touchpad classification (gestures, two-finger scroll) remains follow-up work.

## Quattro porting

Three things in #6093 would not have worked here:

- `sudo` throughout — `install/hardware/*` runs as root (`omarchy-setup-hardware` hard-fails if `EUID != 0`).
- the migration `source`d the leaf — migrations run as the user, so it now escalates explicitly.
- migration timestamp `1781432360` predates quattro's newest (`1785095882`) — renumbered.

Stale-file cleanup moved out of the install leaf into the migration; fresh installs never had that file.

## Verification

- Quirks syntax parses cleanly under real libinput with `MatchUdevType` removed.
- Install leaf is idempotent — byte-identical file on rerun.
- Migration end to end with stubbed `sudo`: removes the stale file, writes the drop-in, exits 0.
- Migration on non-matching hardware exits 0 and writes nothing.
- `test/shell.d/migrate-scope-test.sh` passes; `bash -n` clean.

Not verified on the hardware itself — @mijuny's B9406 is currently broken. The path fix is mechanically proven; a confirmation on a live B9406 before merge would be ideal.

### DIFF
--- a/install/hardware/asus/fix-asus-ptl-b9406-touchpad.sh
+++ b/install/hardware/asus/fix-asus-ptl-b9406-touchpad.sh
@@ -8,13 +8,21 @@
 #
 # Mask the pressure axes with a quirks override, same pattern as the
 # Asus UX302LA entry in libinput's shipped 50-system-asus.quirks.
+#
+# libinput scans /usr/share/libinput for *.quirks and additionally reads the
+# single file /etc/libinput/local-overrides.quirks. It never scans /etc, so a
+# drop-in is the only way to ship a quirk without editing the one override
+# file users hand-edit themselves.
+#
+# The pad's event node is tagged ID_INPUT_MOUSE rather than ID_INPUT_TOUCHPAD,
+# so a MatchUdevType=touchpad constraint would skip the section. Bus, vendor,
+# product and DMI already pin the device exactly.
 
 if omarchy-hw-asus-expertbook-b9406; then
-  mkdir -p /etc/libinput
-  cat > /etc/libinput/asus-expertbook-b9406.quirks <<'EOF'
+  mkdir -p /usr/share/libinput
+  cat > /usr/share/libinput/99-omarchy-asus-b9406-touchpad.quirks <<'EOF'
 [ASUS ExpertBook B9406 Touchpad]
 MatchBus=i2c
-MatchUdevType=touchpad
 MatchVendor=0x093A
 MatchProduct=0x4F05
 MatchDMIModalias=dmi:*svnASUS*:pn*B9406*
--- a/migrations/1785114924.sh
+++ b/migrations/1785114924.sh
@@ -1,0 +1,7 @@
+echo "Fix ASUS ExpertBook B9406 touchpad quirk (takes effect at next login)"
+
+omarchy-hw-asus-expertbook-b9406 || exit 0
+
+sudo rm -f /etc/libinput/asus-expertbook-b9406.quirks
+sudo env OMARCHY_PATH="$OMARCHY_PATH" PATH="$PATH" \
+  bash -euo pipefail "$OMARCHY_PATH/install/hardware/asus/fix-asus-ptl-b9406-touchpad.sh"


dhh:
@mijuny — could you test this on a B9406 when you have one in hand? No rush if the ExpertBook is still out of action; I'd rather this sit than merge unverified.

The fix moves to a drop-in libinput actually reads, so the check is whether the cursor moves after a relogin:

```bash
# 1. Apply
sudo bash /usr/share/omarchy/install/hardware/asus/fix-asus-ptl-b9406-touchpad.sh

# 2. Confirm the drop-in landed and the stale file is gone
cat /usr/share/libinput/99-omarchy-asus-b9406-touchpad.quirks
ls /etc/libinput/

# 3. Relog, then confirm libinput applied it
libinput list-devices | grep -A5 -i touchpad
```

Two things beyond "cursor moves" that would be genuinely useful:

- `sudo libinput record` on the touchpad node, so we can see what udev actually tags it. That settles the `MatchUdevType` question I flagged — the `ID_INPUT_MOUSE` vs `ID_INPUT_TOUCHPAD` story doesn't quite line up with the original jump-detection diagnosis, and I suspect the pad exposes more than one event node.
- Whether two-finger scroll and tap-to-click work, or only pointer motion. If it's motion only, that confirms the device is being driven as a mouse and scopes the follow-up you mentioned.

If you'd rather test the migration path as an existing install would hit it:

```bash
rm -f ~/.local/state/omarchy/migrations/1785114924.sh
omarchy-migrate
```

argrig666:
Tested this branch's installer on the actual ASUS ExpertBook B9406CAA (`ASCP1D80:00 093A:4F05`, I2C, `/dev/input/event9`). The fix works and the `/usr/share/libinput` location is the key change.

Validation performed:

- Installed `99-omarchy-asus-b9406-touchpad.quirks` using this PR's script.
- `libinput quirks validate --data-dir /usr/share/libinput` passes.
- `libinput quirks list --data-dir /usr/share/libinput /dev/input/event9` prints `AttrEventCode=-ABS_MT_PRESSURE;-ABS_PRESSURE;`. Passing `--data-dir` here excludes my `/etc/libinput/local-overrides.quirks`, so this verifies the packaged drop-in itself matches.
- `libinput list-devices` reports pointer + gesture capabilities, two-finger/edge scrolling, and button-area/clickfinger methods.
- Masking the two bogus pressure axes restores cursor motion; the device advertises pressure max 2601 but reports near-zero values during valid contact.

One factual correction for the script comment: this device is **not** tagged `ID_INPUT_MOUSE` on this machine. `udevadm info` reports:

```
ID_INPUT=1
ID_INPUT_TOUCHPAD=1
ID_INPUT_TOUCHPAD_INTEGRATION=internal
```

So `MatchUdevType=touchpad` does match here. Removing it is still safe because bus/vendor/product/DMI pin the section exactly, but the old quirk failed because libinput does not scan arbitrary `/etc/libinput/*.quirks`, not because of the udev type.

There are still occasional light-contact dropouts after this libinput fix. A raw HID trace shows the controller itself clearing Tip Switch and pausing coordinates while Contact Count and Confidence remain set, so that residual sensitivity appears to be device firmware rather than something this quirk can address. This PR nevertheless fixes the original unusable/no-motion behavior and is ready from the B9406 hardware side.

argrig666:
Post-reboot hardware follow-up: the residual light-contact interruptions and perceived sensitivity loss I noted above are now gone in normal use.

There were two simultaneous changes, so I cannot honestly assign causality:

- the machine's BIOS changed from B9406CAA.304 on every prior recorded boot to B9406CAA.312 on this boot;
- this is also the first full boot with the working libinput override plus a device-only Hyprland sensitivity setting of +0.35.

The PR's quirk is definitely active: libinput quirks list on the live 093A:4F05 touchpad reports AttrEventCode=-ABS_MT_PRESSURE;-ABS_PRESSURE;. This strengthens the end-to-end hardware validation for the PR, but it should not be read as proof that the quirk alone cures the previously observed residual dropouts. BIOS 312 is a plausible contributor to that part.

bramvera:
Confirming this is still needed on a fresh Omarchy 4.0.4 install (ExpertBook B9406CAA, libinput 1.31.3, kernel 7.2.5-3-omarchy). The installer ran `fix-asus-ptl-b9406-touchpad.sh` and wrote `/etc/libinput/asus-expertbook-b9406.quirks` with the correct rule, but the cursor was dead until a copy of the same section was placed in `/etc/libinput/local-overrides.quirks`.

`libinput quirks list --verbose /dev/input/eventN` shows `local-overrides.quirks` being parsed and matching, and never opens the Omarchy-named file. The `MatchUdevType=touchpad` line does match on this unit, for what it is worth.

Would be great to see this land, since every B9406 owner hits it on first boot.

mijuny:
Second hardware confirmation of the PR drop-in (@mijuny, B9406CAA BIOS 304, libinput 1.31.3, kernel 7.2.5-3-omarchy, PR head 85abe527).

A/B with `libinput debug-events` on the touchpad node (`ASCP1D80:00 093A:4F05 Touchpad`), 20 s per run, finger circling:

- PR drop-in only: 3992 POINTER_MOTION events.
- No quirk (two runs): 0 POINTER_MOTION, only DEVICE_ADDED.

No jump warnings in any run, so on 1.31.3 a dead pad is silent rather than the original "Touch jump detected and discarded". I also ran `sudo libinput record` on the touchpad node during a no-quirk run, as you asked. The record header lists `ID_INPUT_TOUCHPAD=1`, `ID_INPUT_TOUCHPAD_INTEGRATION=internal`, `DRIVER=hid-multitouch`, and the raw stream carries full multitouch motion (`ABS_MT_POSITION_X` 1356-3651, `ABS_MT_POSITION_Y` 44-2436, 24 tracking IDs, BTN_TOUCH / BTN_TOOL_FINGER / BTN_TOOL_DOUBLETAP) with pressure only 0-152 against an advertised max of 2601. So the kernel data is fine and libinput is the layer dropping it. In every run libinput adds the node as a touchpad (`cap:pg`, tap, two-finger scroll, clickfinger), so it is not being driven as a mouse. Happy to attach the recording if useful.

`libinput quirks list --verbose` shows the drop-in full-matching on bus/vendor/product/DMI and applying `AttrEventCode=-ABS_MT_PRESSURE;-ABS_PRESSURE;`. An isolated fixture run confirmed the old `/etc/libinput/asus-expertbook-b9406.quirks` is never opened, and that the migration removes it, leaves an existing `local-overrides.quirks` byte-identical, and reruns cleanly (sudo shimmed).

On MatchUdevType, your multi-node suspicion is right: the pad exposes `... Mouse` (ID_INPUT_MOUSE=1, relative axes only) and `... Touchpad` (ID_INPUT_TOUCHPAD=1). `MatchUdevType=touchpad` matches the touchpad node here, as on argrig666's and bramvera's units. Without it the section also lands on the mouse node, where it is a no-op since that node has no absolute axes. Nothing broken, but I would restore the line (every Pixart entry in libinput's `30-vendor-pixart.quirks` carries it) and fix the comment saying the pad is tagged ID_INPUT_MOUSE. That is a comment-and-scope change, not a substitute for the path fix.

Rebooted with the local override removed so only the drop-in applies: after a fresh login the pad works, and libinput applies the mask from /usr/share/libinput/99-omarchy-asus-b9406-touchpad.quirks alone. In the same session, pointer motion, two-finger scroll (262 SCROLL_FINGER events), physical click (BTN_LEFT), and tap-to-click all work.

For transparency: the hardware test was run together with my agents DeepSeek V4.1 Flash and Codex (gpt-6-astra). Astra also verified the installer and migration in isolation, and Claude Opus 5 and Grok 4.6 reviewed this draft.

install/hardware/asus/fix-asus-ptl-b9406-touchpad.sh
```diff
@@ -8,13 +8,21 @@
...
 # Asus UX302LA entry in libinput's shipped 50-system-asus.quirks.
+#
+# libinput scans /usr/share/libinput for *.quirks and additionally reads the
+# single file /etc/libinput/local-overrides.quirks. It never scans /etc, so a
+# drop-in is the only way to ship a quirk without editing the one override
+# file users hand-edit themselves.
+#
+# The pad's event node is tagged ID_INPUT_MOUSE rather than ID_INPUT_TOUCHPAD,
+# so a MatchUdevType=touchpad constraint would skip the section. Bus, vendor,
+# product and DMI already pin the device exactly.
```
mijuny, line 19:
```suggestion
# The device exposes separate mouse and touchpad event nodes. Match the
# touchpad node, which is tagged ID_INPUT_TOUCHPAD; bus, vendor, product
# and DMI restrict the quirk to this model.
```

install/hardware/asus/fix-asus-ptl-b9406-touchpad.sh
```diff
@@ -8,13 +8,21 @@
...
 
 if omarchy-hw-asus-expertbook-b9406; then
-  mkdir -p /etc/libinput
-  cat > /etc/libinput/asus-expertbook-b9406.quirks <<'EOF'
+  mkdir -p /usr/share/libinput
+  cat > /usr/share/libinput/99-omarchy-asus-b9406-touchpad.quirks <<'EOF'
 [ASUS ExpertBook B9406 Touchpad]
 MatchBus=i2c
```
mijuny, line 25:
```suggestion
MatchBus=i2c
MatchUdevType=touchpad
```

mijuny (commented):
Two actionable edits from my test comment above, both comment/scope only; neither changes the file-location fix.

My A/B shows the pad exposes two nodes, `... Mouse` (ID_INPUT_MOUSE) and `... Touchpad` (ID_INPUT_TOUCHPAD). The quirk targets the touchpad node, which is correctly tagged, so the current comment premise is inaccurate; without the constraint the section also lands on the mouse node (a no-op there). Every Pixart entry in `30-vendor-pixart.quirks` carries `MatchUdevType=touchpad`. Inline suggestions below.

Also, the PR body still says hardware verification is missing. My comment above is that verification (second B9406CAA, BIOS 304, libinput 1.31.3, kernel 7.2.5-3-omarchy), so it can be updated before merge.

PR #6392, opened 2026-07-27
Prefer app id matches in menu search

zainsyedz:
## Summary
- include each desktop app id in the Omarchy menu app row search aliases
- rank exact app token matches, such as `spotify` from `com.spotify.Client`, like exact app label matches
- keep stronger menu matches above weaker app prefix matches

## Tests
- bash test/shell.d/menu-test.sh
- node -c shell/plugins/menu/MenuModel.js

### DIFF
--- a/shell/plugins/menu/Menu.qml
+++ b/shell/plugins/menu/Menu.qml
@@ -253,7 +253,7 @@ Item {
       var appId = String(entry.id || "")
       if (!appId) continue
       var subtext = root.appLibrary.entrySubtext(entry)
-      var aliases = subtext ? [subtext] : []
+      var aliases = subtext ? [subtext, appId] : [appId]
       try {
         if (entry.keywords && typeof entry.keywords.join === "function") aliases = aliases.concat(entry.keywords)
       } catch (e) { }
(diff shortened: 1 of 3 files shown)


dhh:
What problem are you solving with this? What search is improved?

zainsyedz:
> What problem are you solving with this? What search is improved?

image

Whenever you search for an application that is already installed, in my case it was spotify, the current algo doesn't prioritize an app that is already installed, which I believe should be the default(better) behaviour.

Whenever I was launching spotify, it always prioritizes the install->spotify option, and me hitting enter because of my muscle memory always launched the install script/popup

zainsyedz:
Updated the PR to address the review comments:

- changed the app-id regression so the id-token match is not accidentally covered by the visible label
- added concrete installed-app ranking coverage for Spotify and RetroArch, both of which have matching static install menu entries
- guarded the exact app-token boost so generic reverse-DNS namespace tokens like com/org/io/net don't get top-tier scoring

Tests rerun:
- bash test/shell.d/menu-test.sh
- node -c shell/plugins/menu/MenuModel.js

zainsyedz:
Addressed the latest suggestion by moving the app-token stop words to a module-level lookup instead of allocating an array during each score calculation.

dhh:
Ahh, we already fixed this on latest quattro 👍. Thanks for looking into it, though.

dhh:
image

PR #6393, opened 2026-07-27
Enable cd/dvd ejection by default - and use Omarchy glyph for "Done" messages in the terminal instead of spinning globe

brink-lab:
A two-in-one-PR. Take what you want :)

**Change 1**
Enable cd/dvd ejection by default for all systems with a key mapped to "XF86Eject".
This e.g. activates the eject button on vintage macbook Pro laptops with superdrives.

**Change 2**
Update Omarchy-show-done to use the Omarchy glyph in "Done" messages in the terminal. This gives a subtle hint that the Done message comes from the Omarchy system. Substitutes the spinning globe. 

Current:
image

Proposed:
image

### DIFF
--- a/bin/omarchy-show-done
+++ b/bin/omarchy-show-done
@@ -3,4 +3,5 @@
 # omarchy:summary=Display a "Done!" message with a spinner and wait for user to press any key.
 
 echo
-gum spin --spinner "globe" --title "Done! Press any key to close..." -- bash -c 'read -n 1 -s'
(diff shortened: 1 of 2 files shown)


dhh:
I prefer the spinning globe because it attracts the eye. But happy to take the eject. Make sure you only open single issue PRs in the future. No omnibuses.

brink-lab:
Got it. Sorry about the double, I'm new to this. Closing this one and opening a new for the eject only.

PR #6396, opened 2026-07-27
feat(menu): configurable input position

felixzsh:
since omarchy menu header is freezed at top, i feel its more natural to have the input at bottom, so  with this pr you can configure it in shell.json adding this:

  "menu": {
    "inputPosition": "bottom"
  }

if not set, the menu looks and behaves exactly the same as now

### DIFF
--- a/config/omarchy/shell.json
+++ b/config/omarchy/shell.json
@@ -1,5 +1,8 @@
 {
   "version": 1,
+  "menu": {
+    "inputPosition": "top"
+  },
   "idle": {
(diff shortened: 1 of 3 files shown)


dhh:
Interesting idea, but I prefer it at the top. Feel free to fork the menu plugin and tailor as you see fit of course!

felixzsh:
https://github.com/user-attachments/assets/36dd27b9-fdcb-4c3a-8fd7-908b6b2be413

here a video if you reconsider this

PR #6402, opened 2026-07-27
Add refresh-rate control to the display panel

startupfoundry:
Adds refresh-rate control to the display panel, alongside the existing scale
control, and the `omarchy hyprland monitor refresh rate` command behind it.

```
omarchy hyprland monitor refresh rate            # current rate
omarchy hyprland monitor refresh rate list       # rates at the current resolution
omarchy hyprland monitor refresh rate 144        # apply live, auto-reverts unless confirmed
omarchy hyprland monitor refresh rate up|down    # step
omarchy hyprland monitor refresh rate confirm    # keep
omarchy hyprland monitor refresh rate revert     # drop now
```

Only the rates the display advertises for the resolution already in use are
offered, so a request cannot silently fall back to preferred.

### Applying a rate is reversible by default

A refresh rate the link cannot carry leaves a black screen, and unlike a bad
scale the compositor cannot tell: Hyprland reports the mode it asked for
whether or not the panel is showing it. So a rate is applied live, nothing is
written to disk, and it reverts on its own after 15 seconds
(`OMARCHY_MONITOR_REFRESH_TIMEOUT`) unless confirmed. That ordering is also
what makes reverting reliable — the config on disk is still the last known
good state, so revert is just `hyprctl reload`.

Asking for the rate already on screen starts none of that, as there would be
nothing to try and nothing to revert. And because reverting is a reload, a
change that replaces one still on trial starts from the rate the trial did,
which makes asking for that rate simply a revert.

The panel mirrors this: a `REFRESH RATE` row of presets, hidden unless the
display offers more than one rate, plus a Keep / Revert row with a countdown
that appears only while a change is in flight. Revert sits on the left and
Keep on the right, with a single `confirmKeepIndex` that both the buttons and
the keyboard handler read, so the row cannot be reordered into disagreeing
with what its buttons do.

That row takes the cursor on Revert and ignores the keyboard for its first
second. It appears under an Enter that is often still held from picking the
rate, or is being pressed blind at a screen gone black, and key repeat
(250ms by default) would otherwise reach Keep before anyone had seen the
result. The one press that must never be an accident is the one that saves an
untested rate. Clicks go straight to their button and do not wait.

### Persisting only when the rule is ours to own

Hyprland replaces a matching monitor rule wholesale rather than merging it, so
a rate cannot be layered on as its own rule without dropping that monitor's
scale, bitdepth, and colour settings. And since `hyprctl monitors` reports
`currentFormat` rather than the configured rule, those cannot be faithfully
snapshotted and written back.

Persisting therefore means owning the rule outright, which is only safe when
the user has not written one. On confirm, a `desc:`-keyed rule is appended
carrying `scale = omarchy_monitor_scale` so `monitor scaling` keeps composing,
marked with a comment so later changes update that line rather than stacking
duplicates. Each monitor owned this way has a marker and rule of its own, so
giving one a new rate leaves the others the rates they were given.

A hand-configured monitor is left alone: the change is kept for the session
and says why. It says so as a notification as well as on stderr, because the
panel shows neither stderr nor an exit status and would otherwise pass a
session-only Keep off as a saved one.

### Names are never trusted as Lua

Following #8129. The owned rule names the monitor by its description, which is
the display's EDID and so chosen by whatever is plugged in, and `monitors.lua`
is Lua that Hyprland runs again on every reload. A description carrying a
quote, backslash, or control character could end the string early and
execute, so it is never written out; the rate is kept for the session
instead, as for a hand-configured monitor. The connector name and scale
handed to `hyprctl eval` are held to a plain name and a plain number for the
same reason, as in the sibling monitor commands.

### `omarchy-monitor-state`

The rate, the available rates, and any pending change are appended to
`omarchy-monitor-state` so the panel keeps polling one command. Each answers
with exactly one line whatever the helper does, because the panel reads this
output by line index: `|| echo` does not cover a helper that succeeds with
nothing to say, nor a `jq` that exits 0 on the empty input a dead helper
leaves it, and either would drop a line and shift every field below it.

### Known limitations

- The live `hl.monitor` eval replaces a `desc:`-keyed or catch-all rule
  wholesale for the session, as `monitor scaling` already does, so a
  `transform`, `bitdepth`, or `vrr` set there is dropped until the next reload.
- Revert is a global `hyprctl reload`, so it also returns any other live-only
  monitor state, such as an unsaved scale, to what the config says.
- A pending change belongs to one monitor but the panel shows it under
  whichever monitor has focus.

### Testing

`test/shell.d/monitor-refresh-rate-test.sh` (23 cases) covers listing,
filtering to the active resolution, applying without persisting, the pending
report, persistence and in-place rule updates, one managed rule per monitor,
leaving user rules untouched and notifying why, stepping and clamping,
snapping an unsupported rate onto an advertised mode, requests that change
nothing, a request for the rate a trial started from, revert, auto-revert,
superseded countdowns, hostile descriptions and connector names, and
confirming with nothing pending. `test/shell.d/monitor-test.sh` gains 21
assertions for the parsing and formatting helpers, the confirm row's key
settle, and the panel's wiring to it. `test/shell.d/monitor-state-test.sh` is
extended to the new line count and to a helper that is silent or dead.

Verified on a 2560x1440 display advertising 59.95/99.95/120/144: applying,
keeping, and reverting from both the panel and the CLI, by pointer and by
keyboard, and the auto-revert on timeout. The changes made since rebasing are
covered by the tests above, and the panel loads under the runtime smoke test.

screenshot-2026-07-27_13-49-56

### DIFF
--- a/bin/omarchy-hyprland-monitor-refresh-rate
+++ b/bin/omarchy-hyprland-monitor-refresh-rate
@@ -1,0 +1,400 @@
+#!/bin/bash
+
+# omarchy:summary=Show, set, or adjust the focused Hyprland monitor refresh rate
+# omarchy:args=[up|down|list|pending|confirm|revert|RATE]
+# omarchy:examples=omarchy hyprland monitor refresh rate | omarchy hyprland monitor refresh rate 144 | omarchy hyprland monitor refresh rate up
+
+# A refresh rate the link cannot carry leaves a black screen, and unlike a bad
+# scale the compositor cannot tell: Hyprland reports the mode it asked for
+# whether or not the panel is showing it. So a rate is applied live and reverts
+# on its own unless it is confirmed, and nothing is written to disk until then.
+# That ordering is what makes revert reliable — the config on disk is still the
+# last known good state, so reverting is just a reload.
+
+STATE_DIR="${XDG_STATE_HOME:-$HOME/.local/state}/omarchy"
+RATE_LOG="$STATE_DIR/monitor-refresh-rate.log"
+PENDING_FILE="$STATE_DIR/monitor-refresh-rate-pending.json"
+MONITOR_LUA="$HOME/.config/hypr/monitors.lua"
+MANAGED_MARKER="-- Managed by omarchy hyprland monitor refresh rate."
+CONFIRM_TIMEOUT="${OMARCHY_MONITOR_REFRESH_TIMEOUT:-15}"
+
+usage() {
+  cat <<'USAGE'
+Usage: omarchy-hyprland-monitor-refresh-rate [up|down|list|pending|confirm|revert|RATE]
+
+  (no args)  Print the focused monitor's current refresh rate
+  list       Print every rate available at the current resolution, low to high
+  RATE       Apply a rate, reverting automatically unless confirmed
+  up|down    Step to the next or previous available rate
+  pending    Print JSON describing a change awaiting confirmation
+  confirm    Keep the pending rate and persist it when Omarchy owns the config
+  revert     Drop the pending rate immediately
+USAGE
+}
+
+focused_monitor() {
+  hyprctl monitors -j | jq -e -c '.[] | select(.focused == true)'
+}
+
+# Hyprland reports 143.99899 while availableModes advertises 144.00Hz. Rounding
+# to two decimals lets both sides compare as the same string.
+normalize_rate() {
+  awk 'NR == 1 { printf "%g\n", int($0 * 100 + 0.5) / 100 }'
+}
+
+# A rate only means anything alongside a resolution, so only offer the rates
+# the panel advertises for the mode already in use. Sorted low to high to match
+# the scale row in the shell panel, which runs 1x on the left to 4x on the right.
+available_rates() {
+  local width="$1" height="$2"
+
+  hyprctl monitors all -j |
+    jq -r --arg prefix "${width}x${height}@" \
+      '.[] | select(.focused == true) | .availableModes[] | select(startswith($prefix))' |
+    sed -E 's/^.*@//; s/Hz$//' |
+    while read -r rate; do printf '%s\n' "$rate" | normalize_rate; done |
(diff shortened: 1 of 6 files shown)


axelfontaine:
PR needs a refresh to be mergeable again

Related to 
- #10195
- #10902

startupfoundry:
Thanks @axelfontaine, rebased onto current `quattro` and mergeable again.

Since this sat for a while, I went back over it against what's landed upstream in the meantime rather than only resolving the conflict:

- **Follows #8129.** The helper now holds the connector name and scale to a plain name and number before `hyprctl eval`, like the sibling monitor commands. It also never writes an EDID description containing a quote, backslash or control character into `monitors.lua`; the rate is kept for the session instead. Tests cover both.
- **`textFormat: Text.PlainText`** on the two new `Text` elements, per the new lint.
- **`omarchy-monitor-state` always emits exactly one line per field**, with `monitor-state-test` extended to the new count. `|| echo` alone didn't cover a helper that exits 0 silently, or `jq` exiting 0 on empty input.
- **Each managed monitor keeps its own rule.** Previously, changing one monitor's rate again could overwrite another's.
- **The Keep/Revert row takes the cursor on Revert and ignores the keyboard for its first second**, so a held or blind Enter can't save an untested rate.
- **A Keep that can't be saved now sends a notification**, since the panel doesn't surface stderr.
- **Asking for the rate already in use is a no-op.**

Known limitations I've left alone: the live `hl.monitor` eval replaces a `desc:`/catch-all rule wholesale (same as `monitor scaling`), and revert is a global `hyprctl reload`.

On #10195 and #10902: the distinguishing choice here is applying live with an auto-revert and only persisting on confirm, since a rate the link can't carry is a black screen the compositor can't detect. Happy to fold in anything from those if you'd prefer a different shape.

langri-sha:
Thanks for keeping this active @startupfoundry, I'll close my PR and feel free to source anything you need from it.

PR #6404, opened 2026-07-27
Defer plocate updatedb to an idle first-boot oneshot

davydotcom:
## Summary
Defer `updatedb` from install into a first-boot idle oneshot so the installer does not block on a full plocate index.

- `install/post-install/localdb.sh` enables a timer instead of running `updatedb` synchronously
- `omarchy-updatedb-first-boot.timer` fires ~15s after boot (`OnBootSec=15s`)
- oneshot runs `updatedb` at `Nice=19` / `IOSchedulingClass=idle`, matching plocate hardening
- uses `ConditionFileNotEmpty=!/var/lib/plocate/plocate.db` so a zero-byte package placeholder still counts as unseeded

## Test results
Verified locally against `quattro` with a local-source ISO (`omarchy-iso-make --local-source`) and `omarchy-iso-test --install-only`:

- Install completed successfully (~63.6s total phases); configure phase no longer runs a sync `updatedb`
- Timer enabled on the installed system; units present under `/etc/systemd/system/`
- First build used `ConditionPathExists=!…`, which skipped the oneshot when plocate left a **0-byte** `plocate.db` — fixed to `ConditionFileNotEmpty=!…`
- After the fix: oneshot ran successfully (**~1.05s wall / 572ms CPU**, ~185M peak); `plocate.db` grew to ~5.1M and `locate` worked

## Test plan
- [x] Local ISO build with `--local-source` against this branch
- [x] `omarchy-iso-test --install-only` succeeded
- [x] Confirm no sync `updatedb` during post-install configure
- [x] Confirm empty placeholder DB still triggers first-boot seed
- [x] Confirm non-empty DB + working `locate` after oneshot

### DIFF
--- a/etc/systemd/system/omarchy-updatedb-first-boot.service
+++ b/etc/systemd/system/omarchy-updatedb-first-boot.service
@@ -1,15 +1,16 @@
 [Unit]
 Description=Seed plocate database after Omarchy install
 Documentation=man:updatedb(8)
-ConditionPathExists=!/var/lib/plocate/plocate.db
+# plocate's package may leave a zero-byte placeholder DB; treat empty the same
+# as missing so the first-boot seed still runs.
+ConditionFileNotEmpty=!/var/lib/plocate/plocate.db
 After=local-fs.target
 
 [Service]
 Type=oneshot
 ExecStart=/usr/bin/updatedb
-# Once plocate.db exists, ConditionPathExists skips this unit on later boots.
-# Leave the timer enabled; evaluating the condition is cheaper than unpacking
-# sandbox privileges enough to systemctl disable from ExecStartPost.
+# Once plocate.db is non-empty, ConditionFileNotEmpty skips later boots.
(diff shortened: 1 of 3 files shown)


davydotcom:
This was a thought I had after analyzing the install process flow. `updatedb` seemed like something that could async run on first boot instead of at the end of install. This may shave a few seconds off post install getting the user into the desktop more quickly. The sacrifice is you have deferred the locate db to first boot but if this is running in the background for a few seconds at startup, it probably won't matter by the time the user can even get to a place to consume it. Also feels more in line with macOS how it tends to defer indexing until install as well.

dhh:
How much did this save during the install?

davydotcom:
1 second... on virtualized test at 58 seconds total. But seconds count at these speeds!

dhh:
Appreciate the effort here, but I think it's too much complexity to gain 1s. I think we're already sub-minute after all the changes done so far!

PR #6409, opened 2026-07-28
Add Dropbox pause toggle to panel header

tobi:
## Summary

- add a compact pause/resume switch to the Dropbox panel title row
- reflect the optimistic active state immediately while Dropbox settles
- disable the switch and its tooltip while a Dropbox operation is busy
- retain the existing hero-icon and keyboard controls

## Related header-toggle PRs

- Tailscale: #6408
- Bluetooth: #6410

This includes the same `PanelHero.titleCenterY` helper as #6408 so the PR remains independently mergeable. If #6408 lands first, that shared one-line change drops out of this diff.

## Validation

- `bash test/shell.d/dropbox-test.sh`
- `git diff --check`

### DIFF
--- a/shell/Ui/PanelHero.qml
+++ b/shell/Ui/PanelHero.qml
@@ -15,6 +15,7 @@ Item {
   property alias metaOpacity: metaText.opacity
 
   readonly property color dim: Qt.darker(foreground, 1.4)
+  readonly property real titleCenterY: heroLabels.y + titleRow.y + titleRow.height / 2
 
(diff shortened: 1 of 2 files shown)


dhh:
Folded into #6408, which now moves all five panel heroes (Tailscale, Dropbox, Bluetooth, Audio, Network) onto a shared `Ui/ToggleSwitch` rather than landing three copies of the same hand-rolled switch. Closing in favour of that.

PR #6412, opened 2026-07-28
fix(omarchy-menu): Escape steps back one menu level before closing

heyssh:
# Menu — Escape Back-Navigation Fix

**File:** `Menu.qml`

**Bug:** Pressing `Esc` inside any submenu closed the entire menu immediately
instead of stepping back one level, forcing you to reopen the menu and
re-navigate every time.

**Fix:** `Esc` now calls `goBack()` first (same as `Backspace`/`Left`), and
only falls back to a full close once already at the root menu.

Before:

https://github.com/user-attachments/assets/db30f5fd-2bf5-44c5-bac9-a89d6f0dad15

After:

https://github.com/user-attachments/assets/e7bc4b74-8a01-47cf-b8dd-81714cf16e7b

### DIFF
--- a/shell/plugins/menu/Menu.qml
+++ b/shell/plugins/menu/Menu.qml
@@ -975,8 +975,12 @@ Item {
             root.requestDeleteSelected()
             event.accepted = true
           } else if (event.key === Qt.Key_Escape) {
+            // Clear an active search first, then step back one menu level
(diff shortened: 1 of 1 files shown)


dhh:
I actually prefer having esc as a way to close the entire menu from any depth and then left arrow + backspace for just going up and down. Will document this in the manual.

PR #6422, opened 2026-07-29
Fix asymmetric rounded shell borders

OldJobobo:
## Problem

Asymmetric rounded menu borders can paint an unintended antialiased pixel at a disabled corner. The current epsilon-offset OddEven ring avoids Qt filling the entire row, but leaves nearly coincident geometry at zero-width corners.

## Solution

- build closed winding-filled contours for connected enabled-side runs
- omit geometry at corners where both adjacent sides are disabled
- preserve disconnected opposite-side borders as separate contours
- conservatively fill the outer shape when the inner dimensions or rounded interior are consumed
- retain the existing native border path for uniform flat borders

## Visual evidence

**Before — asymmetric rounded-border artifact**

Before: upper-right border artifact

**After — corrected side-localized border**

After: corrected asymmetric rounded border

## Verification

- `test/shell.d/border-geometry-test.sh`
- `test/shell.d/row-border-stability-test.sh`
- all 16 enabled-side masks and point-in-path behavior covered
- live-tested with Neuromancer `selected-border-width = "0 0 1 3"` and rounding 3
- confirmed the upper-right artifact is removed while intended left and bottom borders remain
- two independent review rounds completed with no remaining blockers

`test/shell` reaches and passes the changed border suite, then stops in this checkout because the expected sibling `omarchy-pkgs/pkgbuilds/omarchy-settings-dev/PKGBUILD` fixture is absent.

### DIFF
--- a/shell/Commons/BorderGeometry.js
+++ b/shell/Commons/BorderGeometry.js
@@ -172,6 +172,11 @@ function normalizeRadii(w, h, r) {
   return { tl: tl, tr: tr, br: br, bl: bl }
 }
 
+function appendArc(path, rx, ry, sweep, point) {
+  if (rx > 0 && ry > 0) path.push("A", rx, ry, 0, 0, sweep, point.x, point.y)
+  else path.push("L", point.x, point.y)
+}
+
 function roundedRectPath(x, y, w, h, radii) {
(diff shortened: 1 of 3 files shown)


ryanrhughes:
Thanks for chasing this one down @OldJobobo

PR #6425, opened 2026-07-29
Fix bluetooth panel section headers rendering clipped

heyssh:
## Bluetooth panel

- Fixed section headers ("PAIRED", "AVAILABLE", "CONNECTED") appearing
  clipped at the top of the device list.
- Fixed the device list auto-scrolling unexpectedly on mouse hover;
  scroll-into-view now only triggers via keyboard navigation.

### Problem

Section headers (PAIRED, AVAILABLE, CONNECTED) in the bluetooth
bar-widget panel rendered with their top sliced off. PanelSectionHeader's
bold, letter-spaced text draws slightly above its own box, and since the
header sits at the very top of the device list Flickable (`clip: true`),
that sliver was clipped away.

### Fixes

- Offset `deviceList` by `Style.space(6)` inside the Flickable and grow
  `contentHeight` to match, so the overflow has room to render instead
  of being clipped.
- Gate `ensureCursorVisible()` behind a new `keyboardNav` flag so it only
  fires on j/k/h/l navigation, not mouse hover — a device row appearing
  under an already-stationary cursor was triggering an unwanted
  auto-scroll into view.
- Clamp `Flickable.contentY` whenever `contentHeight`/`height` change, so
  a stale scroll offset from a taller list can't persist and clip content
  after the list shrinks (e.g. during discovery).
- Reset `contentY` to 0 whenever the panel opens.

### Testing

- Manually verified in Omarchy Quattro: opened the bluetooth panel with
  no paired devices while scanning was active, confirmed "AVAILABLE"
  header renders fully; confirmed hovering a device row no longer
  scrolls the list; confirmed list shrinking during discovery no longer
  leaves stale scroll offset.

Before:

https://github.com/user-attachments/assets/c70d34a7-7007-427e-a316-6c966035cd0c

After:

https://github.com/user-attachments/assets/aad41498-57c4-44c7-80eb-dbfc35445719

### DIFF
--- a/shell/Ui/PanelSectionHeader.qml
+++ b/shell/Ui/PanelSectionHeader.qml
@@ -15,4 +15,12 @@ Text {
   font.family: fontFamily
   font.pixelSize: fontSize
   font.bold: true
+
+  // Glyphs can paint above the box Text reserves for them: JetBrainsMono Nerd
+  // Font's outlines run 10% of the em past its own ascent, and a patched or
+  // user-chosen family can be worse. That sliver is invisible in normal flow,
+  // but a header sitting at the top of a clipping list — bluetooth's device
+  // Flickable, network's station list — loses it to the clip and renders
+  // beheaded. Reserve the overshoot here so every panel is covered at once.
+  topPadding: Math.ceil(fontSize * 0.15)
 }
--- a/shell/plugins/panels/bluetooth/Panel.qml
+++ b/shell/plugins/panels/bluetooth/Panel.qml
@@ -662,9 +662,9 @@ Panel {
         Flickable {
           id: deviceFlick
           width: parent.width
-          height: Math.min(deviceList.implicitHeight + Style.space(6), Style.space(400))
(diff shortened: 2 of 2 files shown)


dhh:
Thanks for chasing this down — the diagnosis was right, and the repro videos made it obvious.

Landed on `quattro` as 10459759 with you as the author, but moved up a level: the overshoot is now reserved in `PanelSectionHeader` itself rather than in the bluetooth panel's geometry. JetBrainsMono Nerd Font's outlines run 10% of the em past the ascent `Text` reserves (yMax 1120 against ascent 1020), so any header opening a clipping list loses that sliver — network's station list had the same latent bug, and it's fixed for free now.

The scroll fixes are superseded rather than dropped: a439069b replaced the hand-driven Flickable with a ListView, which keeps the current row visible, re-clamps itself when discovery shortens the list, and — since `ListView.Contain` only moves for a genuinely clipped row — can't lurch under a hovering mouse. That retires `keyboardNav` and `clampContentY` along with the code that made them necessary.

Closing since everything here is on `quattro`.

Issue #5887, opened 2026-05-17
Issue with windows management - Davinci Resolve


nikitastetskiy:
### System details

OS: Omarchy 4.0.0.alpha - Kernel: Linux 7.0.8-1-cachyos - Display Server: Wayland (Hyprland 0.55.2)

### What's wrong?

all### Description
After upgrading to the latest version, DaVinci Resolve Studio (running under XWayland) is experiencing severe window management and focus regressions. Additionally, a persistent cursor scaling issue remains present.

### Regressions after upgrading:
* **Incorrect Fullscreen/Border Behavior:** DaVinci Resolve automatically launches in a broken fullscreen state. The window borders do not behave correctly and cause a visual overlay/glitch on the edges of the screen.
* **Broken Window Layering/Focus:** Opening child windows or dialogs within Resolve spawns them on the same layer as the main panel. Clicking the main panel instantly sends the child window to the background, rendering it inaccessible unless moved to a separate workspace.
* **Workspace Navigation Lock:** If a Resolve child window is moved to another workspace, switching focus back to the previous workspace becomes completely locked/impossible.

### Persistent Legacy Issue:
* **Cursor Mis-scaling / Blurring:** The mouse cursor appears blurry, unfocused, or underscaled when hovering over DaVinci Resolve. (Note: This is a pre-existing issue that also affects other XWayland applications like Steam).

(all of these issues are visible on the video and tried to find solutions/workarounds with no success yet)

### Environment Context
* **WM/Compositor:** Hyprland 0.55.2 (Wayland)
* **Screen Resolution:** 2880x1800 @ 2x scaling
* **GPU:** AMD Radeon 860M (Mesa/VAAPI)
* ThinkPad P14s AMD Gen 6

https://github.com/user-attachments/assets/47c2309a-51fb-4de2-b44c-6ce99b92f6b0

Sergoiip:
I have the exact same issue but with a Nvidia GPU (4060). Have you figured out a fix?

nikitastetskiy:
@Sergoiip hi, nothing yet. Working with full screen when it comes to Davinci Resolve and moving the new windows of resolve to other workspaces when opening them

AksharP5:
Same issue here, added this [PR](https://github.com/basecamp/omarchy/pull/5926) to resolve the issue, if you wanna react that would be great! Until it is merged in, you can make those same changes locally, and it should resolve it for you, works for me!

@nikitastetskiy @Sergoiip

[...]

PR #6430, opened 2026-07-29
hypr/apps: stop DaVinci Resolve's dialogs trapping the pointer

28allday:
## Symptom

Opening certain DaVinci Resolve windows — confirmed with **Preferences** and **Project Manager** — traps the pointer. The dialog cannot be dismissed and no other window can be clicked. **Quitting Resolve entirely is the only way out.**

## Root cause

`default/hypr/apps/davinci-resolve.lua` sets `stay_focused = true` on `.*[Rr]esolve.*` — that is, on *every* Resolve window.

That rule is there for a real reason and **must not simply be deleted**. Resolve is not Wayland-native, so Hyprland misreads its transient popups (Change Clip Duration, Normalize Audio Levels, …) and they close the moment the pointer leaves them unless focus is pinned. See [hyprwm/Hyprland#12235](https://github.com/hyprwm/Hyprland/discussions/12235), where the `stayfocused` workaround originates.

The bug is that it is applied **too broadly**. `stayfocused` forces focus onto a window while it is visible. When two Resolve windows are visible at once, both demand focus and neither yields, so Hyprland flips between them indefinitely.

Captured live, sampling `hyprctl activewindow` every 0.4s, with the Project Manager and its own child dialog both open:

```
t=361.7 ACTIVE='resolve'|'Create New Project'  resolve_windows=['Project Manager', 'Create New Project']
t=365.4 ACTIVE='resolve'|'Project Manager'     resolve_windows=['Project Manager', 'Create New Project']
t=365.8 ACTIVE='resolve'|'Create New Project'  resolve_windows=['Project Manager', 'Create New Project']
```

Focus oscillates between the two. Neither can be acted on, which is exactly what "the mouse is captured" looks like from the user's side.

## Fix

Keep the pin where it is needed — the transient popups — and remove it from the windows those popups open *over*. At most one pinned window is then ever visible, so nothing can fight.

The two parent windows are the main window and the Project Manager:

| window | role | pinned after fix |
|---|---|---|
| `DaVinci Resolve - <project>` | parent of Preferences, Change Clip Duration, … | no |
| `Project Manager` | parent of Create New Project | no |
| `Preferences`, `Create New Project`, transient popups | children | yes |

This is a narrowing, not a removal: every window that previously relied on `stay_focused` to stay open still has it.

## Verification

Hyprland 0.56.1 / Omarchy 4.0.0.r1440, NVIDIA, 2560×1440 logical, DaVinci Resolve 21.0.3.

**The override mechanism is confirmed.** A later, more specific rule does beat an earlier blanket one, tested in isolation with a throwaway class so nothing about Resolve could confound it:

```lua
o.window("^(OVERRIDETEST)$", { float = true })
o.window({ class = "^(OVERRIDETEST)$", title = "^nofloat$" }, { float = false })
```
```
title='nofloat'    floating=False   <- later specific rule wins
title='yesfloat'   floating=True    <- blanket rule still applies
```

**The trap is confirmed removed, by cold-start A/B.** Resolve was started fresh twice — once with the fix, once with only the `stay_focused` exclusion removed, so the exclusion was the sole variable. In each run Resolve was focused, focus was then dispatched to another window, and the actual focus owner sampled 12 times at 0.5s intervals:

| run | config | samples where Resolve held focus | verdict |
|---|---|---|---|
| A | with fix | **0 / 12** | focus sticks elsewhere — not trapped |
| B | control, exclusion removed | **11 / 12** | snaps back to Resolve — trapped |

The control matters: it demonstrates the test can actually detect the trap, so run A's result is meaningful rather than vacuous. Both runs exercised the **Project Manager**, one of the two windows the rule excludes and one of the two originally reported.

**The main window and the popups were then confirmed by hand.** Resolve always requires a project to be selected from the Project Manager before the main window opens, so these two checks cannot be automated:

| check | procedure | result |
|---|---|---|
| Main window no longer traps | Open a project, open **Preferences** from the main window | **Pass** — dismissable, and other applications remain clickable while it is open |
| Popups do not regress | Edit page → right-click clip → **Change Clip Duration**, move pointer off the popup onto the main window | **Pass** — popup stays open |

The second row is the one that matters most: it is the behaviour `stay_focused` exists to protect, and it survives the narrowing intact.

## Known limitations / open questions

- **The parent list is a blocklist and may need extending.** Only `DaVinci Resolve - <project>` and `Project Manager` are excluded. If another Resolve window can itself parent a dialog (Project Settings and the Deliver page are the likely candidates), it would reintroduce the trap and need adding to the alternation.
- **A cleaner long-term fix may exist.** Hyprland 0.53.3+ added `input:mouse_refocus`, and [#12235](https://github.com/hyprwm/Hyprland/discussions/12235) notes `mouse_refocus = 0` as an alternative to `stayfocused` entirely. On 0.56.1 the underlying popup bug may even be fixed, which would make the whole `stay_focused` rule obsolete. I did not test that here because `mouse_refocus` is a **global** input setting, not a per-window rule, so flipping it for every Omarchy user is a much larger decision than this narrowing. Worth a maintainer's opinion.
- Possibly related: #5887 reports Resolve child windows becoming inaccessible on Omarchy 4.0.0.alpha. It doesn't mention `stayfocused`, but "child windows become inaccessible when the main panel is clicked" is consistent with this root cause, so this may resolve it.

### DIFF
--- a/default/hypr/apps/davinci-resolve.lua
+++ b/default/hypr/apps/davinci-resolve.lua
@@ -6,3 +6,8 @@ o.window(".*[Rr]esolve.*", {
   tag = "-default-opacity",
   opacity = "1 1",
 })
+
+-- stay_focused above stops Resolve's transient popups closing on mouse-out
+-- (hyprwm/Hyprland#12235), but pinning the windows they open over makes two
+-- windows fight for focus and traps the pointer. Unpin the parents only.
+o.window({ class = ".*[Rr]esolve.*", title = "^(DaVinci Resolve - .+|Project Manager)$" }, { stay_focused = false })


dhh:
Integrated with #6429 in fb564e3ac1484d0d2b9c4b73b1264579368d4dea: https://github.com/basecamp/omarchy/commit/fb564e3ac1484d0d2b9c4b73b1264579368d4dea

The combined commit keeps the stay-focused override minimal and expands the parent-window match to cover both DaVinci Resolve and DaVinci Resolve Studio. Thanks @28allday — you are credited as co-author on the commit.

PR #6449, opened 2026-07-30
Fix Bitwarden floating window rendering issues

### LINKED ISSUE #6075: Bitwarden Chrome Extension popup locked to 875x600 since 2026.5.1 update
### System details

Omarchy 3.8.2, Hyprland 0.55.2, Chromium 148.0.7778.96, Bitwarden Chrome Extension 2026.5.1

### What's wrong?

Description:
The Bitwarden Chrome Extension popup window is being forced to 875x600 and cannot be corrected via windowrules. The correct size the extension accepts is 480x650. This started after Bitwarden updated to 2026.5.1 on approximately June 2, 2026.

Suspected causes:
Three existing changes combined with the Bitwarden 2026.5.1 update to produce this bug:

- system.conf applies size 875 600 to all windows tagged floating-window (introduced in commit 4c97a31c, November 2025)

- bitwarden.conf tags the Chrome extension class with +floating-window (introduced in 3.4.0, February 2026), causing it to inherit the 875x600 size rule from system.conf

- Bitwarden 2026.5.1 changed how the popup requests its window size, making the forced 875x600 visibly incorrect — prior to this update the mismatch was not noticeable

Workaround:
- First, find your Bitwarden Chrome extension class by opening the popup and running:
```
hyprctl clients | grep -A3 "Bitwarden"
```

- Look for the class and initialClass fields. For Chromium, this will be chrome-nngceckbapebfimnlniiiahkandclblb-Default. 

- Then add to your personal windowrules config, substituting your class if different:
```
windowrule = tag -floating-window, size 480 650, match:class chrome-nngceckbapebfimnlniiiahkandclblb-Default
```

This removes the floating-window tag before the size rule is applied, bypassing the 875x600 override in system.conf.

Possible fix:
Add an explicit size rule for the Chrome extension class directly in bitwarden.conf to override the generic floating-window size, or exempt it from the size 875 600 rule in system.conf.

Elevate08:
Windows tagged with +floating-window have a size of "875 600" by default.
This is larger than what the Bitwarden extension requests and causes the popup window to not refresh properly.
The window can be forced to resize by switching windows; however, forcing the max_size and removing the standard +floating-window tag allows the window to render properly.

Fixes #6075

### DIFF
--- a/default/hypr/apps/bitwarden.lua
+++ b/default/hypr/apps/bitwarden.lua
@@ -2,5 +2,7 @@ o.window("^(Bitwarden)$", { no_screen_share = true, tag = "+floating-window" })
 
 o.window("chrome-nngceckbapebfimnlniiiahkandclblb-Default", {
   no_screen_share = true,
-  tag = "+floating-window",
+  float = true,
+  no_blur = true,
+  max_size = "480 650",
 })


neheb (approved):

raybarrera:
This PR fixes the toolbar popup (`chrome-nngceckbapebfimnlniiiahkandclblb-Default`) being forced to 875×600 via `+floating-window`. That match is incomplete for Omarchy web apps / PWAs.

`omarchy-launch-webapp` launches Chromium `--app=` when the default browser is not Chrome/Brave/Edge/Opera/Vivaldi/Helium (Zen and Firefox fall through to Chromium). The Bitwarden popup from that path uses a different class:

```
chrome-nngceckbapebfimnlniiiahkandclblb__popup_index.html-Default
```

That name starts with `chrome`, so `default/hypr/apps/browser.lua` tags it `chromium-based-browser` and tiles it. On this machine that window was `1896×1030`, `floating = false`.

Suggested follow-up in `default/hypr/apps/bitwarden.lua` — keep the desktop-app rule, match every `chrome-nngceck…` class, drop both `floating-window` and `chromium-based-browser`, then float at Bitwarden's 480×650:

```lua
o.window("^(Bitwarden)$", { no_screen_share = true, tag = "+floating-window" })

o.window("chrome-nngceckbapebfimnlniiiahkandclblb.*", { tag = "-floating-window" })
o.window("chrome-nngceckbapebfimnlniiiahkandclblb.*", { tag = "-chromium-based-browser" })
o.window("chrome-nngceckbapebfimnlniiiahkandclblb.*", {
  float = true,
  no_blur = true,
  no_screen_share = true,
  size = { 480, 650 },
  max_size = { 480, 650 },
})
```

Verified locally on Omarchy 4.0.0 / Hyprland 0.56.2 / Chromium 151: after this rule the `--app=` popup is `480×650`, `floating = true`. Happy to open a follow-up PR if you'd rather land this separately from the 875×600 change.

Elevate08:
@raybarrera I can't recreate the class you encountered.

For me, this PR fixes the popup encountered when accessing bitwarden from a form field. Not from the extension toolbar.

I've tested Firefox, Brave, Chrome, and Chromium. The only browser to experience a non floating window for Bitwarden was Brave, which has a similar class except it starts with brave- instead of chrome-

bjarneo:
Automated duplication check: this pull request looks similar to #8341, which covers the same Bitwarden float rule. I keep that one open and close this one to consolidate review.

PR #6450, opened 2026-07-30
Enable be211 wifi7

spencerbull:
## Summary

- Stop disabling WiFi 7 on Intel BE211 (`8086:e440`) systems.
- Keep the EHT workaround for Intel BE200 (`8086:272b`), where Linux kernel bug 221675 still tracks a warm-reconnect failure at 6 GHz/320 MHz.
- Add a guarded migration that removes the historical Omarchy workaround from existing BE211 installations only when every discovered installed kernel is Linux 7.1 or newer.

## Why

Omarchy added the workaround in [basecamp/omarchy#5106](https://github.com/basecamp/omarchy/pull/5106) while the new `iwlmld` path made EHT connections unusable on Panther Lake hardware. Linux 7.1 includes relevant `iwlmld` TSO fixes, including [torvalds/linux@92cee08](https://github.com/torvalds/linux/commit/92cee08dc4f00e77fd1317e4343c5d458b0abab7).

The original BE211 system was retested on Linux `7.1.5-arch1-1-ptl` with firmware `102.07fca168.0`:

- Negotiated EHT on 6 GHz channel 37 at 320 MHz.
- Reached 3.46 Gbps RX / 3.84 Gbps TX PHY rates.
- Completed five controlled disconnect/reconnect cycles, returning to EHT/320 MHz every time.
- Transferred approximately 1.95 GB under bidirectional load.
- Measured approximately 718 Mbps download and 622 Mbps upload during the warm-state test.
- Recorded zero TX failures, TX retries, beacon losses, firmware crashes, or `iwlmld` errors.

The BE200 workaround remains because [kernel bug 221675](https://bugzilla.kernel.org/show_bug.cgi?id=221675) is still open and describes a distinct 320 MHz firmware assertion after warm reconnects.

## Migration safety

The migration:

- Matches the complete historical Omarchy-generated config before removing it.
- Preserves user-authored or modified files.
- Requires every discovered installed kernel package to be version 7.1 or newer.
- Conservatively skips unknown and older kernels.
- Moves the config to a transaction backup while rebuilding boot images.
- Treats a successful desired-state boot-image rebuild as the transaction boundary.
- Restores both the config and boot images if that rebuild fails.
- Warns without undoing the rebuilt image if backup cleanup or reboot-state recording fails.
- Recovers and rebuilds safely after an interrupted previous attempt.
- Marks a reboot as required after successful migration and after interrupted-recovery rebuilds.

## Validation

- `bash -n` on both changed scripts.
- ShellCheck 0.11.0 on both changed scripts.
- `git diff --check`.
- Full `test/omarchy-cli-test.sh` suite: 60 checks passed.
- Hardware-policy matrix covering BE211 and BE200 installer behavior.
- Migration matrix covering:
  - current, older, mixed, and unknown kernels
  - exact Omarchy and user-owned config files
  - initial rebuild failure
  - backup-cleanup and reboot-state warning failures
  - rollback rebuild failure
  - interrupted migration recovery
  - recovery with an older kernel or hardware mismatch
  - orphaned backup restore and rebuild failures
- Independent read-only review completed with no remaining substantive findings.

## Remaining risk

The live A/B validation covers one BE211, firmware, and access-point combination. Suspend/resume and other BE211/AP combinations still need broader real-world coverage. The BE200 exclusion should remain until its upstream 320 MHz issue is resolved.

### DIFF
--- a/install/config/hardware/intel/fix-wifi7-eht.sh
+++ b/install/config/hardware/intel/fix-wifi7-eht.sh
@@ -1,15 +1,12 @@
-# Temporary fix for Dell XPS 14/16 on Panther Lake
-# Disable WiFi 7 (EHT/802.11be) on Intel BE200/BE211 cards
-# The iwlwifi driver has a broken EHT RX data path — APs drop to MCS 0 NSS 1
-# when EHT is negotiated, making WiFi unusable. Disabling EHT falls
-# back to WiFi 6 (HE/802.11ax) which works at full speed.
-# This should be removed when Intel fixes the firmware/driver.
+# Temporarily disable WiFi 7 (EHT/802.11be) on Intel BE200 cards.
+# Linux kernel bug 221675 tracks a firmware assertion after warm reconnects
+# at 6 GHz/320 MHz. Disabling EHT avoids the failing channel context.
+# https://bugzilla.kernel.org/show_bug.cgi?id=221675
 
-if lspci -nn | grep -qE '\[8086:(e440|272b)\]'; then
+if lspci -nn | grep -q '\[8086:272b\]'; then
   sudo tee /etc/modprobe.d/iwlwifi-disable-eht.conf <<'EOF'
-# Temporary fix Dell XPS 14/16 on Panther lake
-# Disable WiFi 7 (EHT) on Intel BE200/BE211 — broken RX rate adaptation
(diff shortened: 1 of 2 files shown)


dhh:
Turns out performance is worse: https://github.com/basecamp/omarchy/pull/6709

PR #6460, opened 2026-07-31
Add a Ghostty keybindings menu, bound to Super+Alt+T

robert-matusewicz:
I recently switched to Ghostty and truly enjoy it!

One thing that bugs me, though, is the constant problem with finding the keybindings - `ghostty +list-keybinds` is helpful, but I would love to have the same experience as with SUPER + K.

This PR adds a `omarchy-menu-ghostty-keybindings` command that shows Ghostty's currently effective keybindings (config overrides merged with its built-in defaults) in a searchable Walker menu, mirroring the existing `omarchy-menu-tmux-keybindings`. Bound to `SUPER + ALT + T` (I am not sure if this is the best key...).

### DIFF
--- a/bin/omarchy-menu-ghostty-keybindings
+++ b/bin/omarchy-menu-ghostty-keybindings
@@ -1,0 +1,152 @@
+#!/bin/bash
+
+# omarchy:summary=Display Ghostty keybindings currently configured using walker for an interactive search menu.
+# omarchy:args=[--print|-p]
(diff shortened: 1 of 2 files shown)


dhh:
I appreciate this, but feels a bit specific, given that we have several terminal options, and this currently isn't the default. Probably would be best as a plugin. Thanks for working on it!

PR #6464, opened 2026-07-31
feat(agents): add OpenCode Go collector with watermark incremental scan

markbus-ai:
## Summary

Adds an **OpenCode Go** collector to the `omarchy.agents` usage widget, in the standalone-collector architecture (`bin/omarchy-agent-usage-<agent>` prints a display-ready JSON record; the panel discovers it by convention, no panel changes needed).

The collector shows the **real subscription meters** dhh asked for — rolling 5-hour / weekly / monthly windows as percentages with reset times from OpenCode's official usage API — plus token stats from the local `opencode.db`.

## What's new (vs the earlier model-usage approach)

| Piece | Role |
|---|---|
| `bin/omarchy-agent-usage-opencode-go` | Collector: live limits probe + pure-SQL scan + watermark incremental cache |
| `shell/plugins/agents/assets/opencode-go.svg` (+ `-light`) | Panel mark |
| `test/shell.d/agent-usage-opencode-scanner-test.sh` | 53-assertion suite (bash + python fixtures) |
| README | Record contract, scan behavior, channel selection |

## Highlights

- **Real plan limits** — `GET https://opencode.ai/zen/go/v1/usage` with the Go API key from `~/.local/share/opencode/auth.json` or `OPENCODE_GO_API_KEY`. Handles both the deployed response shape (`percent`/`resetsAt`) and the upstream-PR shape (`usagePercent`/`resetInSec`), with payload-wide percent-scale detection (no meter discontinuity after a window reset).
- **Watermark incremental scan** — the headline difference. The collector never re-scans the whole table in steady state: it tracks a watermark (max `time_created`), the db/WAL mtimes, and a 10-minute overlap band with id-dedupe, so a normal refresh reads only what changed. On a 1.9 GB / 62k-row database:
  - cold full scan: **4.3–5.2 s / ~166 MB RSS**
  - steady-state fast path (db unchanged): **0.72 ms** (≈6,000× faster, SQLite not even opened)
- **Pure-SQL aggregation** — `json_extract` exact-match on `role`/`providerID` is the authority (LIKE prefilters are pure acceleration); rows stream as lightweight columns, raw JSON blobs never accumulate in Python.
- **Multi-db channel selection** — among `opencode-*.db` files it prefers the one with the newest WAL activity, so users on a release channel see the right database instead of a stale default.
- **Hardening** — https-only endpoint override (refuses `http://` with no key sent, loud warning on non-default host), Authorization stripped on any `(scheme, netloc)` redirect change (including https→http downgrade), O_NOFOLLOW lock, cache files 0600 / dir 0700, empty `XDG_CACHE_HOME` fallback, cache failures always degrade to a valid record — the collector never prints an empty stdout.
- **Concurrency-safe cache commits** — envelope writes are gated on watermark progress (not mtimes) so a concurrent older merge can never regress a fresher envelope, including at the midnight day-rollover boundary; `--force` always rewrites a fresh scan even when the table shrank.

## Test plan

- [x] 53-assertion scanner suite: record contract, SQL totals, both response shapes, percent scale, envelope v3 (bandMs, watermark, mtimes, session-id sets), FAST/INCREMENTAL/FULL paths, overlap dedupe, band-migration guard, day rollover, race/rollover concurrency, force-after-delete, channel selection, redirect safety, permissions, unwritable-cache degrade, timestamp unit guard
- [x] Existing collector suites still green (codex 20, claude 8)
- [x] Live run on a 1.9 GB database (cold + fast path + incremental), benchmark table above

## Notes

- During local verification the live probe returned HTTP 403 for the test key; the collector maps this to `usageStatusText`/`authHelpText` and still emits a valid record with local stats. Both payload shapes are unit-tested.


dhh:
Hmmm. I'm a little skeptical on this. I think what we could need is the subscription status for OpenCode Zen and Go. But it needs to expose the limits. That's actually what makes this useful.

markbus-ai:
@dhh Agree — the useful part is Go/Zen subscription meters, not local token totals.

I dug into what OpenCode actually exposes:

- **API keys** (`opencode` / `opencode-go`) only get inference + `GET /zen/v1/models`. No spend, balance, or plan windows.
- **Go limits** ($12 / 5h, $30 weekly, $60 monthly) exist server-side and power `GoUsageLimitError` when you hit the wall, plus the web console via `queryLiteSubscription` — but that path is session-cookie SolidStart RPCs, not something the CLI/API key can read.
- Tools like [OpenUsage](https://github.com/janekbaraniewski/openusage) reverse-engineer those console RPCs with browser cookies and **deploy-rotating function IDs**. That works as a hobby scraper; it is not something Omarchy should ship.

So this PR’s local `opencode.db` stats are real, but they are not the Claude/Codex-shaped product (plan % + reset). Happy to **close this** until OpenCode exposes a stable usage endpoint for API keys (or reopen/extend once they do). Local day/model totals can stay as a follow-up under real meters, not as the main feature.

dhh:
Maybe we can convince @thdxr to add this!

markbus-ai:
@dhh The endpoint you're describing is already requested upstream — we don't need to invent the shape:

- https://github.com/anomalyco/opencode/issues/16017 — "[FEATURE]: Add Go plan usage/balance API endpoint (rolling/weekly/monthly windows)" (canonical request)
- https://github.com/anomalyco/opencode/issues/38824 — newer duplicate asking for exactly `rollingUsage` / `weeklyUsage` / `monthlyUsage` with `usagePercent`, `resetInSec`, `limit`

That maps 1:1 onto the session/weekly meter pair the panel already renders for Claude and Codex. If @thdxr's team exposes it to API keys, this provider becomes the real product (plan %, reset countdown, bar alarm) with the local `opencode.db` totals as the fallback layer instead of the headline.

I'd say: keep this PR as the ready-to-rebase foundation, and let's put the +1 with the Omarchy use case on #16017 so the OpenCode team sees a concrete consumer asking for it.

dhh:
Heads up: #6603 just landed on `quattro` and restructured this whole area, so this PR needs a rework before it can merge.

The `model-usage` plugin is gone. It's been renamed to `omarchy.agents` (`shell/plugins/agents/`), and the panel is now strictly a display: it watches JSON records under `~/.local/state/omarchy/agents/usage/` and draws whatever appears there. All the in-plugin machinery this PR touches — `providers/*.qml` adapters and `scripts/*_usage_scanner.py` — no longer exists.

Usage collection now lives in standalone collectors in `bin/`. One `omarchy-agent-usage-<agent>` script per agent prints a complete, display-ready usage record (identity, tier, status, rate limits, today/week/all-time stats), and `omarchy-agent-usage-update` runs every collector it finds and writes the records atomically to the state directory.

The upside: adding an agent no longer touches the panel at all. Ship a collector that prints the record contract, optionally add an `assets/<id>.svg` mark (with an `-light` twin if it needs one for light surfaces), and the panel gains a tab on its own. See `shell/plugins/agents/README.md` for the contract and the `claude`/`codex` collectors in `bin/` as reference implementations — #6604 (Cursor) is a good example of a provider PR in the new style.

If you'd like to update this PR to the new structure, rebase on `quattro` and reshape the scanner into a collector. Happy to answer questions here.

— Claude Fable 5, written on behalf of @dhh

markbus-ai:
Closing this in favor of #6779, which implements the OpenCode Go collector in the new omarchy.agents architecture (standalone collector + assets + tests) and now uses the official usage endpoint (GET /zen/go/v1/usage) for the rolling/weekly/monthly meters dhh asked about. This PR predates the #6603 restructure that removed the model-usage plugin it targets, so keeping it open would just confuse the history. Thanks for the discussion.

bjarneo:

PR #6471, opened 2026-08-01
Hide sharee nodes from the Tailscale machines list

jmckible:
Tailscale Funnel routes external internet traffic to a local service. It puts Tailscale's own `funnel-ingress-node` relays in your netmap, flagged `ShareeNode: true`. `tailscale status` hides them, but the panel doesn't. The result is many unusable nodes in the list, sorted ahead of actual devices.

Building off the Mullvad pattern, skip them in `parseStatus`. This also hides unreachable machines owned by users you've shared a device with. It will not hide devices shared *to* you.

Before:

tailscale-before

After:

tailscale-after

### DIFF
--- a/shell/plugins/panels/tailscale/Model.js
+++ b/shell/plugins/panels/tailscale/Model.js
@@ -47,6 +47,12 @@ function isMullvadPeer(peer) {
   return isMullvadHost(dnsName) || isMullvadHost(hostName)
 }
 
+// Funnel relays and machines of users we've shared with sit in the netmap only
+// so they can reach us — tailscale status hides them, so the panel does too.
+function isShareeNode(peer) {
+  return peer && peer.ShareeNode === true
+}
+
 function osIcon(os) {
   var value = String(os || "").toLowerCase()
   if (value === "linux") return "󰌽"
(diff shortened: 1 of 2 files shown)


chickymonkeys:
Confirming this independently on Omarchy 4.0.0-1.

I run Tailscale Funnel on one node to expose a GitHub webhook endpoint. My tailnet has 3 machines, but `tailscale status --json` returns 25 entries in the `Peer` map: 23 of them are `funnel-ingress-node`, one per ingress point of presence. The panel lists all 23, so 2 real peers sit below a wall of relays and the keyboard navigation in the panel is impractical. Running `tailscale status` on the same machine shows only the 3 real machines.

`ShareeNode` is the right discriminator. On my netmap it holds exactly:

```
funnel peers: 23   with ShareeNode == true: 23
real peers:    2   ShareeNode absent on both
```

The relays also carry `tag:ingress`, a sentinel `Created` timestamp of `1973-11-11T00:00:00.1Z`, and a `UserID` that isn't the tailnet owner's, but `ShareeNode` matches what the CLI keys on and covers the shared-device case as well.

Worth noting that any tailnet with Funnel enabled hits this, and the relay count doesn't scale with tailnet size, so a small tailnet gets the worst ratio. This patch applies cleanly and takes the list from 25 rows to 2 for me.

PR #6474, opened 2026-08-01
Add Android development environment

ryuhzk:
Adds an **Android** entry to `Install → Development`, producing a working Android toolchain in one step.

Until now the Development menu covered 14 language environments but nothing for mobile. Getting Android working meant hand-installing Google's command-line tools, wiring up `ANDROID_HOME`, chasing the right JDK, and adding udev rules by hand.

## What it installs

| | |
|---|---|
| JDK | `java@temurin-17` — the version the Android Gradle Plugin targets |
| SDK | `android-sdk` via mise — Google's official `cmdline-tools` |
| Tools | `platform-tools` (adb, fastboot), latest stable `build-tools` and platform |
| Device access | `android-udev` + `adbusers` group |
| Emulator | optional — system image, a ready-made `omarchy` AVD, and the `kvm` group |

No shell rc file is edited. The mise `android-sdk` tool exports `ANDROID_HOME`, `ANDROID_SDK_ROOT` and the `platform-tools`/`emulator` PATH entries through its own env hook, and `default/bash/init` already runs `mise activate bash`, so a new terminal has everything.

## Notable details

**Versions are detected, not pinned.** Google now publishes minor-versioned platform names — `platforms;android-37.0`, `android-37.1` — so a regex matching `android-[0-9]+` silently produces `platforms;android-37`, a package that does not exist, and `sdkmanager` then aborts the whole batch including `platform-tools`. The command parses the first column of `sdkmanager --list` and anchors the match, which also excludes `-beta`, `-rc` and `-ext` variants.

**System images are selected independently of the platform.** They trail the newest platform: at time of writing the newest platform is `android-37.1` but the newest `google_apis` image is `android-37.0`. Tying the image to the platform would silently skip the emulator.

**All interaction happens up front.** Both prompts and the sudo password come before any download, so the multi-GB fetch runs unattended and the sudo timestamp cannot expire mid-install. `mise` and `sdkmanager` run as the user; only the package install and the group changes use sudo, which matters because the SDK installs into `$HOME`.

**`java@temurin-17`, not `java@17`.** In mise, `java@17` resolves to `17.0.2`, a stale Oracle GA build with no security updates, and `java@latest` resolves past JDK 25, which AGP does not support. An existing global Java pin is left alone rather than downgraded.

**Two upstream noise sources are filtered surgically.** `sdkmanager` prints a three-line deprecation banner on every invocation, and `avdmanager` prints two `Error:` lines about a `devices.xml` it merely probes for — immediately before reporting success. Only those exact lines are dropped; real errors and exit codes still reach the user. The suggested replacement, the `android` CLI, downloads itself on first run and cannot answer `--help` offline, so it is not yet usable for scripted setup.

## What this makes possible

With Android installed, Omarchy can develop a complete React Native / Expo app:

- **Android** — build and run locally, on the emulator or a USB device.
- **iOS** — no Mac required. EAS Build runs iOS builds on macOS runners in Expo's cloud, and `eas submit` sends them to TestFlight for real-device testing. The install optionally sets up the `eas` command for this.

That covers both platforms from a Linux desktop. Two caveats worth stating plainly: TestFlight distribution requires an Apple Developer Program membership, and the iOS Simulator is macOS-only, so iOS testing happens on real devices rather than a simulator.

The `eas` command is installed through `omarchy-mise-install npm:eas-cli eas` rather than `npm install --global`. Under mise, npm's global root lives inside the current Node version's directory, so a global install disappears on the next Node upgrade. The wrapper is also lazy — it costs nothing at install time and fetches `eas-cli` on first use.

## Removal

`Remove → Development → Android` uninstalls the SDK, unpins it from mise, deletes `~/.android` (AVDs and adb keys) and drops `android-udev`. Java is deliberately left alone, since it is its own menu entry.

## Tests

`test/shell.d/dev-env-android-test.sh` — 35 assertions using the PATH-stub and command-log pattern from `channel-test.sh`. `silicon`-free and network-free: `mise`, `sdkmanager`, `avdmanager`, `gum`, `sudo` and `omarchy-pkg-add` are all stubbed, and the `sdkmanager --list` fixture reproduces Google's real table including the decoys that must never be selected.

Coverage includes version selection, the emulator accept/decline paths, group membership, prompt-and-sudo-before-download ordering, the two stderr filters, and the removal round-trip against a throwaway `$HOME`.

Every assertion was mutation-tested — the behaviour was deliberately broken and the corresponding assertion confirmed to fail — before this was considered done.

`./test/cli` and `test/shell.d/menu-test.sh` also pass.


ryuhzk:
I've already tested the project locally and it runs fine, plus the emulator works great too—looks amazing.

image

PR #6479, opened 2026-08-01
Add omarchy-capture-code for themed code screenshots

ryuhzk:
Adds `omarchy-capture-code`, which turns code on the clipboard into a PNG matching the active Omarchy theme — saved to the screenshot directory and copied to the clipboard, ready to paste into a chat or a post.

Bound to `SUPER + SHIFT + PRINT`, and added to `Trigger → Capture` alongside Screenshot / Screenrecord / Text / Color.

`silicon` is added to `install/omarchy-base.packages` for fresh installs, and a migration installs it for existing users — the update path is `pacman -Syu` plus migrations, and does not read that manifest.

Rendering is done by `silicon`, which is in the official repos, works offline, and depends only on `fontconfig freetype2 harfbuzz oniguruma` — so it brings no browser or network round-trip into the capture path, and harfbuzz already handles ligatures, CJK mixing and monospace alignment.

## Theming

The theme is rendered on demand from `default/silicon/omarchy.tmTheme.tpl` and cached at `~/.cache/omarchy/silicon.tmTheme`, invalidated by an mtime comparison against the active theme's `colors.toml`. A theme may ship its own `silicon.tmTheme` to override it entirely.

It is deliberately **not** in the `default/themed/` pipeline. Everything there is config for an application the user lives in, regenerated on every theme switch. `silicon` is a one-shot renderer that exists for a few hundred milliseconds; putting it there would make every user pay a render on every theme switch for a feature most never invoke.

The template carries the full `tokenColors` set from the existing `default/themed/vscode-theme.json.tpl` — same TextMate scope vocabulary, including `fontStyle` — so a capture matches the highlighting the user already sees in their editor rather than a reduced subset.

`silicon --theme` accepts a filesystem path, so no syntect cache build is involved.

The font family follows `fc-match monospace` rather than being hardcoded, so `omarchy font set` changes captures the same way `omarchy theme set` does. `OMARCHY_CODE_FONT`, `OMARCHY_CODE_FONT_SIZE`, `OMARCHY_CODE_PADDING` and `OMARCHY_CODE_CORNER_RADIUS` override the defaults.

## Language detection

`silicon` needs `--language` for clipboard input, and its own fallback only matches shebang-style first lines — with no default, so detection failure is a hard error. That is the common case for a mid-file snippet, so detection runs first, in three stages:

1. **Focused window title** via `hyprctl activewindow`. Editors put the filename there, and so does GitHub in a browser.
2. **Content markers** — shebangs, `<?php`, `package` + `func`, `<!DOCTYPE html`.
3. **`gum filter`** picker.

Stage 2 only accepts markers that are *proof*, never inference. Rules keyed on type annotations, `def`, or `::` were written and then removed: across 15 sample files they read C++ as TypeScript and Ruby as Python. A miss costs one keypress in the picker; a wrong guess silently renders the snippet highlighted as the wrong language, and nothing in the result tells the user it happened. The two snippets that were mislabelled are pinned as test cases.

Two details worth calling out:

The value passed to `-l` is a **file extension**, not a language name, because syntect's `find_syntax_by_token` matches extensions first. `bat --list-languages` is tempting as a name source but ships syntaxes absent from syntect's default set, which would produce `Unsupported language`.

An extension taken from a window title is accepted only if it is in a known list. Window titles are full of dots that are not filenames — domains, version numbers, sentence ends — and without the list a browser tab yields a nonsense language.

The picker needs a terminal, and **neither the keybinding nor the menu entry has one**, so the command hands itself off to a floating terminal when it needs to prompt — by resolved absolute path, not by name. Naming it bare leaves the user looking at an empty terminal reporting the executable as missing, with no notification, because the terminal spawn itself succeeded.

That handoff also carries the resolved appearance settings across as explicit assignments. `uwsm-app` writes only argv to the app daemon's fifo, so the relaunched process starts from the systemd user manager's environment; without this every `OMARCHY_CODE_*` the user set is silently dropped the moment the picker is needed.

## Error handling

`silicon` **exits 0 even when it fails** — unsupported language, unloadable theme, unwritable output — and simply writes no file. The artifact is checked rather than the exit code; otherwise the user gets a success notification while the clipboard still holds the old contents.

ImageMagick post-processing rounds the corners and strokes only the four corner arcs in the theme's `accent`, the same value Hyprland uses for the active window border. A full outline read as a generic wrapped frame. Post-processing is cosmetic, so if `magick` fails the unbordered image is kept rather than losing the capture; the result is written to a temp file and moved into place only on success.

## Scope coverage

Transliterating vscode's `tokenColors` turned out not to be enough. That vocabulary and syntect's Sublime grammars disagree, and TextMate matching is prefix-based — a rule for `meta.function-call.generic` never matches a scope of `meta.function-call`. Method calls stayed unstyled in every language as a result.

Probing each candidate scope with a unique colour showed what the grammars actually emit, which added `variable.function`, `meta.function-call`, `variable.other.member`, `entity.name.class`, `constant.other`, `meta.annotation` and `support.other`.

Verified across 15 languages — TypeScript, JavaScript, Python, Rust, Go, Java, C, C++, Ruby, PHP, shell, JSON, YAML, SQL, Lua — every scope the grammars emit is now styled, bar the outermost `source`, which every token carries and which would flatten the whole image to one colour.

## Tests

`test/shell.d/capture-code-test.sh` — 92 assertions, using the PATH-stub and command-log pattern from `channel-test.sh`. No network, no clipboard, no compositor: `silicon`, `magick`, `hyprctl`, `wl-paste`, `wl-copy`, `gum` and `omarchy-font-current` are stubbed against a throwaway `HOME` and `XDG_CACHE_HOME`, so the run never depends on the developer's real theme or font state.

Coverage includes the three-stage cascade and its cancel paths, the no-tty handoff and the settings it carries, `copy`/`save`/default output modes, theme cache generation and invalidation, the theme-provided override, the font/padding/corner-radius knobs and their defaults, silicon's exit-0-on-failure path, and the magick degrade path.

Every assertion was mutation-tested — the behaviour was deliberately broken and the matching assertion confirmed to fail — before this was considered done.

### DIFF
--- a/bin/omarchy-capture-code
+++ b/bin/omarchy-capture-code
@@ -1,0 +1,342 @@
+#!/bin/bash
+
+# omarchy:summary=Turn code on the clipboard into a themed image
+# omarchy:group=capture
+# omarchy:args=[copy|save]
+# omarchy:examples=omarchy capture code | omarchy capture code copy
+
+[[ -f ~/.config/user-dirs.dirs ]] && source ~/.config/user-dirs.dirs
+OUTPUT_DIR="${OMARCHY_SCREENSHOT_DIR:-${XDG_PICTURES_DIR:-$HOME/Pictures}}"
+
+CACHE_DIR="${XDG_CACHE_HOME:-$HOME/.cache}/omarchy"
+CACHED_THEME="$CACHE_DIR/silicon.tmTheme"
+CURRENT_THEME_DIR="$HOME/.local/state/omarchy/current/theme"
+COLORS_FILE="$CURRENT_THEME_DIR/colors.toml"
+THEME_TEMPLATE="$OMARCHY_PATH/default/silicon/omarchy.tmTheme.tpl"
+
+MODE="${1:-both}"
+
+SCREENSHOT_EDITOR="${OMARCHY_SCREENSHOT_EDITOR:-tensaku-edit}"
+
+# silicon has its own config file, but the flags below are passed explicitly and
+# would override it, so the knobs worth changing are exposed here instead.
+#
+# The family follows the system monospace font, so `omarchy font set` changes
+# these screenshots the same way `omarchy theme set` does. Only the size is a
+# separate knob; CODE_FONT overrides the whole thing and takes silicon's
+# fallback-list syntax, e.g. "JetBrains Mono; Noto Sans CJK SC=32".
+CODE_FONT_SIZE="${OMARCHY_CODE_FONT_SIZE:-32}"
+code_font_family=$(omarchy-font-current 2>/dev/null)
+CODE_FONT="${OMARCHY_CODE_FONT:-${code_font_family:-monospace}=$CODE_FONT_SIZE}"
+CODE_PADDING="${OMARCHY_CODE_PADDING:-24}"
+CODE_CORNER_RADIUS="${OMARCHY_CODE_CORNER_RADIUS:-16}"
+
+open_editor() {
+  local filepath="$1"
+  "$SCREENSHOT_EDITOR" "$filepath"
+}
+
+# Extensions we accept from a window title, and the menu offered by the
+# interactive picker when neither the title nor the clipboard content decide.
+# Titles are full of dots that are not filenames — domains, version numbers,
+# sentence ends — so an allowlist beats a bare regex.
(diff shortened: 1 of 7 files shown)


ryuhzk:
Here are screenshots of 4 different themes.

catppuccin-latte
matte-black
rose-pine
tokyo-night

ryuhzk:
BTW, this code syntax highlighting theme follows Omarchy's theme.

Issue #6475, opened 2026-08-01
[Quattro] D-Bus idle inhibits are silently dropped since hypridle was replaced — screensaver fires during video playback (reproduced with Zen browser)


lcd1232:
### System details

Intel Core i5-1335U, Intel Iris Xe (Raptor Lake-P), Linux 7.1.5-arch1-2, Omarchy 4.0.0.r1472.g283276b, Hyprland 0.56.1, quickshell-git 0.3.0.r18.g10b439f

## Summary

Nothing owns `org.freedesktop.ScreenSaver` on the session bus in Quattro, so every idle inhibit sent over D-Bus is silently discarded. The screensaver launches at the configured timeout while video is playing.

Omarchy 3 used `hypridle`, which registers that bus name and honors `Inhibit()`. Quattro replaced it with the Quickshell idle service, which uses `IdleMonitor { respectInhibitors: true }` — and that honors only the Wayland `zwp_idle_inhibit_manager_v1` protocol. An application that requests its inhibit over D-Bus rather than the Wayland protocol therefore has it dropped on the floor.

I reproduced this with Zen browser. The application is behaving correctly; nothing is listening.

## Steps to reproduce

1. Fresh Quattro session.
2. Play a YouTube or Twitch video in Zen browser (1.21.10b).
3. Do not touch the input devices.
4. At `idle.screensaver` (150s in my `shell.json`) the screensaver launches over the playing video. At `idle.lock` (300s) the session locks.

FoxesRCool1:
Came to report the exact same thing!

tmn73:
Same here on 4.0.0.r1541.gfd1034f / Hyprland 0.56.2, and it also kills Steam games, not just video.

Games actually get hit twice. The D-Bus inhibit is dropped like you describe, and gamepad input doesn't reset the Wayland idle timer either, so a controller-only session just looks idle. Confirmed the bus is empty on my side too: nothing owns `org.freedesktop.ScreenSaver`, and `systemd-inhibit --list` only shows sleep and shutdown inhibitors, nothing on idle.

One thing that makes this a lot worse on upgraded machines: Omarchy 3's hypridle default was 600s, `shell.json` defaults to 150/300, and nothing carries the old value over. So right when inhibits stopped working the screensaver also started firing 4x sooner.

`omarchy-toggle-idle stay-awake` works as a manual workaround in the meantime.

Riddlerrr:
I can confirm the issue.
[...]

PR #6492, opened 2026-08-01
Bridge D-Bus screensaver inhibits to the idle service (#6475)

heyssh:
Quattro's Quickshell idle service only honors the Wayland idle-inhibit protocol (IdleMonitor { respectInhibitors: true }). Nothing replaced hypridle's ownership of org.freedesktop.ScreenSaver on the session bus, so every D-Bus Inhibit() call from browsers and video players was silently dropped and the screensaver/lock fired mid-playback.

- Add omarchy-system-idle-inhibit-bridge, a small PyGObject daemon that owns org.freedesktop.ScreenSaver, implements Inhibit/UnInhibit/ GetActive, and auto-releases inhibits when the holding connection drops off the bus.
- Wire a setDbusInhibitors IPC method into the idle Service.qml; idleEnabled now also requires dbusInhibitors === 0. The bridge always reports the absolute inhibitor count (not a delta), so a bridge crash/restart self-heals instead of leaving idling stuck.
- Ship omarchy-idle-inhibit-bridge.service (systemd user unit), enable it on first run, and add a migration for existing installs.
- Extend test/shell.d/idle-test.sh and systemd-test.sh accordingly.

### DIFF
--- a/bin/omarchy-system-idle-inhibit-bridge
+++ b/bin/omarchy-system-idle-inhibit-bridge
@@ -1,0 +1,199 @@
+#!/usr/bin/python3
+
+# omarchy:summary=Bridge D-Bus screensaver inhibit requests to the idle service
+# omarchy:group=system
+# omarchy:hidden=true
+
+# Quickshell's IdleMonitor (respectInhibitors: true) only understands the
+# Wayland zwp_idle_inhibit_manager_v1 protocol. It replaced hypridle, which
+# also registered org.freedesktop.ScreenSaver on the session bus and honored
+# the older D-Bus Inhibit()/UnInhibit() API that browsers and video players
+# still call instead. With nothing owning that name in Quattro, those calls go
+# nowhere and the screensaver/lock fire mid-playback (#6475).
+#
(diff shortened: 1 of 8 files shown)


dhh:
Thanks for tracking this down. The bug is real, but I don’t think we should merge this implementation.

This adds a separate Python daemon, systemd service, migration, shell IPC method, and duplicated inhibitor state to compensate for functionality missing from Quickshell’s `IdleMonitor`. That’s a lot of downstream machinery for us to own, and synchronization across process or shell restarts remains difficult to make reliable.

I think this fix belongs in Quickshell: when `IdleMonitor.respectInhibitors` is enabled, it should account for the standard `org.freedesktop.ScreenSaver` D-Bus inhibition API in addition to Wayland idle inhibitors. Omarchy should then get the correct behavior without maintaining a second idle-inhibition subsystem.

@outfoxxed, does adding D-Bus inhibitor support to `IdleMonitor` upstream seem like the right approach?

I’m going to hold off on this PR while we investigate fixing it at that layer. Thanks again for identifying the regression and putting together a working exploration of the problem.

PR #6495, opened 2026-08-01
Cycle backgrounds whose filename holds glob characters

DataDave-Dev:
### Why

`omarchy theme bg next` gets stuck on the first wallpaper as soon as any background in the theme has a glob character in its name — `[`, `]`, `*` or `?`. Downloaded wallpapers carry these often enough (`wallhaven-abc [4k].jpg`).

`bin/omarchy-theme-bg-next` looked the current background up with an unquoted right-hand side:

```bash
if [[ ${BACKGROUNDS[$i]} == $CURRENT_BACKGROUND ]]; then
```

In `[[ ]]`, an unquoted right-hand side of `==` is a **pattern**, not a literal. `aaa [4k].jpg` as a pattern matches `aaa 4.jpg` or `aaa k.jpg`, but never itself. The loop finds nothing, `INDEX` stays `-1`, and the `INDEX == -1` branch selects `BACKGROUNDS[0]` — so every "next background" jumps back to the first image instead of advancing.

### What

Quote the right-hand side so the comparison is literal.

### Repro

Two backgrounds in the current theme, `aaa [4k].jpg` and `bbb.jpg`, with `aaa [4k].jpg` active:

```
before:  omarchy theme bg next  ->  aaa [4k].jpg   # same image, cycle stuck
after:   omarchy theme bg next  ->  bbb.jpg
```

### Test plan

- Added `test/shell.d/theme-bg-next-test.sh`, which builds a throwaway `$HOME` with those two backgrounds, stubs `omarchy-theme-bg-set`, and asserts both that the cycle advances past the bracketed name and that it wraps back around.
- Verified the new test fails on `quattro` without the one-character change and passes with it.
- `./test/shell` is otherwise unchanged by this branch.

### DIFF
--- a/bin/omarchy-theme-bg-next
+++ b/bin/omarchy-theme-bg-next
@@ -29,7 +29,9 @@ else
   # Find current background index
   INDEX=-1
   for i in "${!BACKGROUNDS[@]}"; do
-    if [[ ${BACKGROUNDS[$i]} == $CURRENT_BACKGROUND ]]; then
+    # Quoted so a filename holding glob characters compares literally rather
+    # than as a pattern, which never matches itself and restarts the cycle.
+    if [[ ${BACKGROUNDS[$i]} == "$CURRENT_BACKGROUND" ]]; then
       INDEX=$i
       break
     fi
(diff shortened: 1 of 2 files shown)


dhh:
Applied the one-line fix in 80e41e46 with co-author credit. We left out the explanatory code comment and dedicated test as requested. Thank you, @DataDave-Dev!

PR #6503, opened 2026-08-02
fix(bar): safely handle read-only properties in ModuleSlot.injectProps

shrijit37:
### Summary
When `ModuleSlot.injectProps()` initializes custom bar widgets (such as `CustomCommandModule`), it checks `"moduleName" in target` and attempts to assign `target.moduleName = moduleName`. However, JS `in` returns `true` even for `readonly` QML properties, causing Qt to throw `TypeError: Cannot assign to read-only property "moduleName"`.

### Fix
Wrap the property assignments in `try/catch` blocks so read-only property assignments fail gracefully without throwing QML exceptions.

## Review follow-up

Copilot asked for runtime coverage of the engine-specific read-only regression — and, in round 2, for the test to exercise the **real production injector** rather than a re-implementation of the guard. Updated branch:

- The guarded assignment moved into `BarModel.applyInjectableProps()` (a single shared implementation).
- `ModuleSlot.injectProps()` in `Bar.qml` now delegates to it.
- `test/shell.d/bar-inject-readonly-test.sh` drives the **actual production `BarModel.js`** on the real quickshell engine: a `readonly`-typed widget keeps every read-only key and still receives writable injections; removing the guard makes the fixture fail with `TypeError: Cannot assign to read-only property "bar"`, proving the path is exercised.
- Pure-node regression in `test/shell.d/bar-test.sh`: non-writable and undeclared properties are skipped, writable ones injected.

**Testing:** `bar-inject-readonly-test.sh` green · `bar-test.sh` 54 ok · full `./test/shell` green against the same single pre-existing environment baseline.

### DIFF
--- a/shell/plugins/bar/Bar.qml
+++ b/shell/plugins/bar/Bar.qml
@@ -1668,9 +1668,11 @@ Item {
     function injectProps() {
       var target = activeItem
       if (!target) return
-      if ("bar" in target) try { target.bar = root } catch (e) {}
-      if ("moduleName" in target) try { target.moduleName = moduleName } catch (e) {}
-      if ("settings" in target) try { target.settings = moduleSettings } catch (e) {}
+      BarModel.applyInjectableProps(target, {
+        bar: root,
+        moduleName: moduleName,
+        settings: moduleSettings
+      })
     }
 
     Component {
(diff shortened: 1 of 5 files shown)


bjarneo:
Automated duplication check: this pull request looks similar to #9972, which covers the same bar injectProps readonly guard. I keep that one open and close this one to consolidate review.

PR #6514, opened 2026-08-02
Fix Wi-Fi and Bluetooth dropdown latency and scan-time segfaults

shrijit37:
The Wi-Fi and Bluetooth dropdown panels had two problems: opening them stalled for ~1s on NetworkManager's access-point flood, and they could segfault the shell when a scan churned the device list. This PR fixes both — panels open instantly with the cached list while the scan runs in the background, and the delegate crash class is removed entirely. No design changes, no functionality loss, no added overhead while the panels are closed.

## Instant open (no polling when closed)

- **Network panel**: opening renders the cached list on the first frame; the PHY scan, DNS/band probes, and ping probe are deferred until after the panel is visible. The QR share and speed-test modals became `Loader`-lazy instead of being instantiated at shell startup. The bar-button click no longer forces a synchronous rescan on open, and a scan scheduled for an open is cancelled if the panel closes first.
- **Status script**: `omarchy-network-status` splits into `--verbose`/`--ping`/`--all` so the 1.5s ping cadence no longer restarts the heavier status process; plain invocation and existing flags behave exactly as before.
- **Bluetooth panel**: discovery starts 250ms after the panel renders and stops on close; device-list rebuilds are debounced during scan floods; the PipeWire sink lookup gets a cached MAC→sink fast path.
- **Shell**: the layer-shell keyboard focus prime is a single 40ms shot instead of repeated 75ms priming.

Measured live (T2 MacBook, Hyprland, release build):

| action | before | after |
|---|---|---|
| network panel open (bar click) | 0.82–1.36s | 0.17–0.24s |
| bluetooth panel open | ~0.9s | 0.17–0.31s |
| panel close | 0.75–0.95s | 0.08–0.25s |

## Crash stability

quickshell segfaulted in `QObjectWrapper::wrap_slowPath` while a list delegate was still incubating: panel rows embedded the backend QObject (`WifiNetwork` / BlueZ `Device`) in list-model data, so every delegate's var property held a live QObject wrapper that could dangle when the model churned — an AP flood or a discovery timeout destroying the object mid-incubation.

- **Network rows** carry primitives only (ssid, signal, security, …); actions resolve the live object via `networkFor()`.
- **Bluetooth scroll and connected rows** use the same primitives-only projection, with `deviceFor()` for actions.

Repro before the fix: open the network panel, hold it through a scan flood, close and reopen — segfault within a few cycles (dumps under `~/.cache/quickshell/crashes/`). After: 10 network flood cycles, 8 bluetooth discovery cycles, and a combined soak — shell alive, no new dumps, zero `TypeError`/`ReferenceError` in the live shell log.

## Follow-up fixes in review

- The ping and packet-loss rows no longer blank out for a second on every 1.5s status refresh — the decoupled probe's last sample is stitched back onto verbose payloads so the averaging window and loss histogram survive (A/B harness against the real state model: blanked 10/10 refreshes before, 0/10 after the first sample).
- Bluetooth rows are clickable again — the click handler re-queried `.values` on a property that already holds the device array, so it silently matched nothing; it now iterates the array directly (keyboard flow, which used a different lookup, was unaffected).

## Testing

- `network-test.sh` 65 ok · `bluetooth-test.sh` 27 ok · full `./test/shell` 605 ok (single pre-existing environment `not ok`: missing `omarchy-pkgs` checkout)
- `qmllint` clean on all touched QML
- Live verification: IPC and bar-click open/close cycles, latency captured with the shell's own timestamps

## Commits

- `dacf3d20` fix(network): defer panel queries and decouple ping probe for fast open
- `147ebd96` fix(bluetooth): defer discovery start and debounce device list updates
- `e9a22894` fix(shell): prime layer-shell keyboard focus once on panel open
- `38dd384e` fix(network): drop redundant rescan on bar click and stop deferred scans on close
- `735dc327` fix(network): keep wifi rows QObject-free to prevent delegate crash
- `0e4270d7` fix(bluetooth): keep device rows QObject-free to prevent delegate crash
- `a59ef6ae` fix(network): keep ping readings stable across verbose refreshes
- `002daac1` fix(bluetooth): resolve click actions from the device array

## Review follow-up

All Copilot findings from the initial review pass are closed on the updated branch:

- **`deviceName` dropped in bluetooth row projection** — rows now carry `deviceName` primitives so `deviceLabel()` keeps working on QObject-free rows; `Model.deviceRow` is exported and covered by a projection regression test in `test/shell.d/bluetooth-test.sh`.
- **Wifi list debounce never armed** — the first change now syncs immediately *and* arms the debounce timer so a scan's access-point flood coalesces into a single trailing rebuild.
- **`scanDone` timer leak on close** — the panel now stops the completion timer too, so a stale deadline can't rebuild the closed panel or disable the next scan on a quick reopen.
- **Verbose refresh double-counted ping samples** — verbose payloads now redisplay the existing ping window (`Model.pingLatencyDisplay`) instead of re-appending the cached probe, so the averaging window and packet-loss histogram stay unbiased; a stale sample can no longer be inserted before the first fresh probe after a reopen.
- **Primitives-only `wifiRow()` contract** now has a direct regression asserting exactly the projected fields and no reference to the `WifiNetwork` object.

### Round 2 (post-re-review)

- The verbose path previously fed the display output back through `pingLatencyState()`, which reads payload-style `iface`/`router_ping_ms` keys that the display state doesn't carry — every refresh saw an empty interface and blanked the window. Display output is now applied directly through a shared `applyPingLatencyState()` helper, with a regression asserting the verbose path never round-trips through the state machine.

### DIFF
--- a/bin/omarchy-network-status
+++ b/bin/omarchy-network-status
@@ -2,22 +2,29 @@
 
 # omarchy:summary=Print active network status for the shell
 # omarchy:group=network
-# omarchy:args=[--verbose]
+# omarchy:args=[--verbose|--ping|--all]
 
-verbose=false
+mode="status"
 internet_probe=1.1.1.1
 
-case "${1:-}" in
-  "")
-    ;;
-  --verbose)
-    verbose=true
-    ;;
-  *)
-    echo "Usage: omarchy-network-status [--verbose]" >&2
-    exit 2
-    ;;
-esac
+while (( $# > 0 )); do
+  case "$1" in
+    --verbose)
+      [[ $mode == "ping" ]] && mode="all" || mode="verbose"
+      ;;
+    --ping)
+      [[ $mode == "verbose" ]] && mode="all" || mode="ping"
+      ;;
+    --all)
+      mode="all"
+      ;;
+    *)
+      echo "Usage: omarchy-network-status [--verbose|--ping|--all]" >&2
+      exit 2
+      ;;
+  esac
+  shift
+done
 
 print_status() {
   local device nm state ssid signal freq
@@ -132,11 +139,26 @@ print_verbose() {
     [[ -r /sys/class/net/$iface/duplex ]] && printf 'duplex\t%s\n' "$(cat /sys/class/net/$iface/duplex)"
   fi
 
+  if [[ $mode == "all" ]]; then
+    print_ping_samples "$gw"
+  fi
+}
+
+print_ping_only() {
+  local route_json gw
+  route_json=$(ip -j route get "$internet_probe" 2>/dev/null)
+  gw=$(jq -r '.[0].gateway // ""' <<<"$route_json" 2>/dev/null)
   print_ping_samples "$gw"
 }
 
-if [[ $verbose == "true" ]]; then
-  print_verbose
-else
-  print_status
-fi
+case "$mode" in
+  verbose|all)
+    print_verbose
+    ;;
+  ping)
+    print_ping_only
+    ;;
+  status)
+    print_status
+    ;;
+esac
--- a/shell/Ui/KeyboardPanel.qml
+++ b/shell/Ui/KeyboardPanel.qml
@@ -72,7 +72,13 @@ PanelWindow {
   }
 
   function beginFocusPrime() {
-    if (open && backingWindowVisible) focusPrimeTimer.restart()
+    if (!open || !backingWindowVisible || focusPrimed) return
+    // Keep the Exclusive phase short but spanning multiple Qt/Wayland
+    // commit cycles: Hyprland grants Exclusive focus to an already-mapped
+    // surface on commit, but OnDemand does not take focus without a new
+    // map, so flipping in the same batch would lose keyboard ownership on
+    // fade-out reopens. 40ms is imperceptible and costs nothing ongoing.
+    focusPrimeTimer.restart()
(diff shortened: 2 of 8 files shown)


dhh:
Thanks for digging into both of these — the segfault diagnosis in particular was the valuable find, and the repro steps made it easy to confirm.

I've opened #6605 with the two fixes I think we need, with you as co-author on all three commits: the QObject-free row projections for both panels, and the redundant `refresh()` on the bar click.

On that last one — the scan deferral was already in `quattro`. `open()` triggers `onOpenedChanged` → `refresh(true)`, which defers the scan; the bare `refresh()` right after it defaults `scanWifi` to false and set `scannerEnabled` synchronously, undoing the deferral. So the bar-click number you measured looks like it was that one line. That's worth knowing, because it means the rest of the optimization work in this PR is riding on an aggregate measurement that doesn't attribute to it.

The parts I didn't carry over, and why, are listed in #6605. A couple have real problems (the `--all` flag is downgraded by any following flag; an empty `--ping` payload gets mistaken for a status payload and blanks the header). The `discovering = false` on close is a genuine fix and I'd take it as its own PR.

Closing this in favor of #6605.


— 🤖 Claude (via [Claude Code](https://claude.com/claude-code)), posting on behalf of @dhh

PR #6515, opened 2026-08-02
Support hardware, fingerprint, and password Polkit flows

mattrayner:
## Summary
- classify the live Polkit/PAM conversation as FIDO/U2F, fingerprint, password, or waiting
- keep Omarchy’s existing PAM-config and lid-state lookahead only for the pre-prompt waiting glyph; concrete live auth states remain authoritative
- delay the neutral pre-prompt surface for one second only; the delay latches off after the first concrete state
- show compact, icon-only physical-auth cards for FIDO and fingerprint
- preserve the authorization context in its own pill and show the localized physical-auth cue in a separate pill
- show and focus the text field only when PAM requests a response, then honor `responseVisible`
- distinguish fingerprint retry messages containing “touch” from FIDO touch cues
- add direct model-helper tests following the existing `run_node_test` shell-test style

## Supported PAM flow
Lid closed: `pam_u2f.so cue` → `pam_unix.so`

Lid open: `pam_fprintd.so` → `pam_u2f.so cue` → `pam_unix.so`

- FIDO/U2F: waits for a touch; Enter does not send a response.
- Fingerprint: waits for a scan; Enter does not send a response.
- Password: presents the PAM prompt, accepts input, and submits it on Enter.

## Maintainer question
The original `FileView` of `/etc/pam.d/polkit-1` and per-request `omarchy-hw-laptop-closed` check have been retained. They now affect only pre-prompt presentation; live `AuthFlow` prompts select the active method. Please advise whether this lookahead remains necessary or can be simplified safely.

## Testing
- `bash test/shell.d/polkit-test.sh`
- `qmllint shell/plugins/polkit/PolkitAgent.qml`
- Local helper: `/tmp/opencode/omarchy-local-pkg/test-polkit.sh` validates, generates a uniquely versioned local package, and prints the install/restart commands without installing.
- Manual local validation: lid-open FIDO, fingerprint, and password fallback completed; compact FIDO layout, one-second pre-prompt delay, and separate authorization/cue pills visually checked.
- Manual closed-lid FIDO/password validation is pending.

The focused suite covers FIDO cues, fingerprint cues (including retry messages), password fallback, prompt precedence, generic waiting, clamshell-gated PAM configuration, and fingerprint-lookahead races.

### DIFF
--- a/shell/plugins/polkit/PolkitAgent.qml
+++ b/shell/plugins/polkit/PolkitAgent.qml
@@ -38,17 +38,26 @@ Item {
   // to the password even when a sensor is enrolled. Refreshed per request.
   property bool laptopClosed: false
   property int shakeOffset: 0
+  property bool waitingDelayLatched: false
 
   readonly property bool dialogVisible: polkitAgent.isActive || closing
-  // We show one method at a time. Fingerprint owns the dialog while PAM is
-  // waiting on the reader (lid open, sensor enrolled); the moment PAM asks for
-  // a password — including immediately when the lid is shut and the clamshell
-  // gate skips pam_fprintd — we switch to the password field instead.
-  readonly property bool fingerprintMode: fingerprintConfigured && !laptopClosed && dialogVisible && !responseRequired && !submitted && !errorFlash
+  readonly property var authState: PolkitModel.authenticationState(currentPrompt, currentSupplementary, responseRequired)
+  readonly property var authPresentation: PolkitModel.authenticationPresentation(authState.method, fingerprintConfigured, laptopClosed)
+  readonly property bool fingerprintMode: dialogVisible && authState.method === "fingerprint" && !submitted && !errorFlash
+  readonly property bool fidoMode: dialogVisible && authState.method === "fido" && !submitted && !errorFlash
+  readonly property bool compactMode: fingerprintMode || fidoMode || (dialogVisible && authState.method === "waiting" && !submitted && !errorFlash)
+  readonly property bool waitingDelayActive: !waitingDelayLatched && !submitted && !errorFlash && authState.method === "waiting" && waitingDelayTimer.running
   readonly property int cardHeight: panel.height > 0 ? Math.min(fieldHeight + contentMargin * 2, panel.height - Style.gapsOut * 2) : fieldHeight + contentMargin * 2
-  // Password mode is a wide field; fingerprint mode collapses to a square that
+  // Password mode is a wide field; compact modes collapse to a square that
   // just frames the centered sensor icon.
-  readonly property int cardWidth: fingerprintMode ? cardHeight : Math.min(Style.space(312), Math.max(Style.space(260), panel.width - Style.gapsOut * 2))
+  readonly property int cardWidth: compactMode ? cardHeight : Math.min(Style.space(312), Math.max(Style.space(260), panel.width - Style.gapsOut * 2))
+
+  Timer {
+    id: waitingDelayTimer
+    interval: 1000
+    repeat: false
+    onTriggered: root.waitingDelayLatched = true
+  }
 
   function authorizationLabel(message) {
     return PolkitModel.authorizationLabel(message)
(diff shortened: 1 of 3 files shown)


mattrayner:
The final implementation looks like this, IF you have both FIDO and fingerprint, the order is based on which appears first in PAM, if the lid is closed, fingerprint is removed, if you unplug your fido device, the security key icon disappears. Its not possible to understand if fingerprint or fido is being asked for at any given time.

The text directly above is that from PAM and then above that is what you are authorising

image

joaodrp:
Confirming this fixes it for me too on Omarchy 4.0.0! My polkit stack is FIDO-only (YubiKey C Bio FIDO Edition), so no `pam_fprintd` at all:

```
auth sufficient pam_u2f.so origin=pam://omarchy appid=pam://omarchy authfile=/etc/fido2/fido2 cue userverification=1 pinverification=0
```

Before this PR I just got a password box with no touch prompt and the assertion failed immediately. With your branch it works fine and the prompt looks as expected:

screenshot-2026-08-14_22-49-05

rfmyrick:
I tested the current Omarchy polkit agent with Irlume 0.11.2 + 1Password system authentication on a 2026 Dell XPS 13 (RGB+IR Windows-Hello camera), and found a related PAM-conversation case that this PR appears to still miss.

Current Quattro already copies `flow.supplementaryMessage` into `currentSupplementary`, but does not render it. Irlume uses a PAM info message for privileged/app-consent auth:

`Type yes to use face authentication`

Then PAM asks for a response. In the current shell, the user only sees the password box. Typing `yes` still works and starts the IR camera, but there is no visible instruction telling the user to do so. I verified this end-to-end with 1Password's Linux system-authentication flow through `polkit-1`: after a tiny local UI patch to render `currentSupplementary`, the instruction appeared and face unlock worked repeatedly.

I checked this PR's `authenticationState()` logic. It currently does:

```js
if (responseRequired) {
  return { method: "password", prompt: input || "Enter password" }
}
```

so a non-empty `supplementaryMessage` is still ignored whenever PAM is also waiting for a response. That is exactly Irlume's case.

I think the general fix should be to render a non-empty PAM supplementary message independently of the response field. For example, keep the password/input prompt in the field, but show `currentSupplementary` in a separate cue/instruction pill above it whenever present. That preserves the authorization context and password UX while also honoring PAM informational messages.

Minimal local proof-of-concept on current Quattro was simply to display `currentSupplementary` in the existing message area when non-empty; no PAM/Irlume/1Password changes were required.

This is broader than Irlume: any PAM module that emits an informational/supplementary message alongside an input prompt is currently silent in the Omarchy agent.

Happy to test a revision of this branch against Irlume + 1Password on the Dell hardware if useful.

mattrayner:
@rfmyrick I've updated this PR to also render the supplementary instructions, can you let me know if this fixes your issue?

rfmyrick:
Retested the supplementary-instruction change on the same 2026 Dell XPS 13, against PR head `4fac3923088221f22b8d1c5ce82cc2c968d87f82`, using Irlume 0.11.2 + 1Password system authentication through `polkit-1`.

All three paths now work correctly:

- the Omarchy polkit dialog visibly renders Irlume's supplementary PAM instruction: `Type yes to use face authentication`
- typing `yes` starts the IR face-auth flow and successfully unlocks 1Password
- entering the normal Linux password instead still succeeds as the fallback path

This fixes the issue I reported: the supplementary instruction is now visible even while PAM is also waiting for a response, without breaking the response/password field behavior.

So from the Irlume + 1Password case, this revision looks good on hardware.

palprateek:
Reviewed this against face unlock as someone testing #11612:

Current `quattro:PolkitModel.js` only knows `pam_fprintd`, so any face line (`pam_facelock.so` / `pam_howdy.so` / `pam_faceauth.so`) would be mishandled without this. Your generic `physical` + `supplementaryInstructionVisible` path covers face safely - face falls through to the cue pill, password fallback preserved, and the Irlume `Type yes` case you fixed is exactly the pattern face modules use.

Please consider merging this independently of face - it's required for any face polkit UX. Small followup idea: extend `authCapabilitiesFromPamConfig` to detect the three face modules explicitly for a face icon, same as fido/fingerprint. Happy to test that revision against facelock on Lenovo hardware.

PR #6516, opened 2026-08-02
Track usage limits for several Claude and Codex accounts

kmpeeduwee:
## Why

I run more than one AI subscription on the same machine: a work Claude account, a second work account, and a personal one. Each lives in its own config directory -  that is what `CLAUDE_CONFIG_DIR` is for, and `CODEX_HOME` is its Codex counterpart. It is also the only mechanism that keeps two logins' credentials *and* transcripts apart.

The agents panel only ever shows the stock directories, so whichever account lives there is the only one with numbers in the bar. The others are invisible, and the subscription switch has nothing extra to switch between.

*Reworked after #6603 onto the collector/record split - this now touches no panel machinery beyond a one-line icon lookup, which is a good sign for that design.*

## What this adds

An account is one more collector. `omarchy-agent-usage-update` now also runs executables from `~/.config/omarchy/agents/collectors.d/`, and the shipped collectors take `--id` / `--name`, so a second account is a three-line wrapper named for the id its record carries:

```bash
# ~/.config/omarchy/agents/collectors.d/claude-work   (chmod +x)
#!/bin/bash
export CLAUDE_CONFIG_DIR="$HOME/.claude-work"
exec omarchy-agent-usage-claude --id claude-work --name "Work" "$@"
```

That's the whole feature: the record lands in the usage directory under its own id, and the panel — which already keys tabs on `record.id` - gains a tab with its own limits, plan line, and totals. Nothing is pooled across accounts. A user collector with a shipped collector's name replaces it, which is also how the stock account's flags can be overridden. `--except` and per-agent `enabled` keep working, since the collector name and the record id are the same string.

## The parts that needed more than plumbing

- **Icons.** Marks resolve as `assets/<id>.svg`, so `claude-work` would fall back to the bar glyph. Records now carry `"mark"` - the tool behind the account - and `iconCandidatesForProvider` prefers it over the id. One line in the panel; every account of a tool wears that tool's mark.
- **The shared trees.** pi, omp, and opencode keep one transcript tree apiece, with nothing tying them to an account - counting them for every account would count them once per account. Ownership is resolved centrally in `omarchy-agent-usage-update`, the one place that sees every account:= collectors answer `--print-identity` with their tool, stock-directory status (compared by path spelling, not symlink target), and their wrapper's preference, and the update run hands each one an explicit `--shared-sessions on`/`off` — appended after the wrapper's own flags, so the central decision takes precedence. An explicit claim wins and switches the stock account off with no paired override; with no claim the stock account counts the trees; with neither enabled they stay uncounted rather than landing under a subscription they may not belong to. Collectors that do not answer the identity probe are left to their own defaults, and a collector run by hand falls back to `auto`.
- **The limits cache.** The Claude collector kept one probe cache for the machine; a second account would have answered with the first one's cached limits. It is now keyed by the account's directory.
- **Sync.** Snapshots carry each account's mark, and machines refuse to pool a name that runs Claude here with one that runs Codex elsewhere - the mismatched snapshot is ignored with a warning, and the per-record lookup matches on the mark too. Snapshots from before marks travelled carry none and merge as they always did.

`omarchy-agent-usage-fireworks` takes `--id` / `--name` as well, for a second prepaid account fed by its own credential environment.

## Tests

- `agent-usage-update-test.sh`: collectors.d discovery, shipped-collector override, non-executables ignored, `--except` on a user collector, and the shared-tree resolution - an explicit claim switching the stock account off, the stock account owning the trees without one, nobody counting them with neither enabled, and a partial run deciding ownership over the enabled population rather than the agents it regenerates. The pre-existing invocations also pin `XDG_CONFIG_HOME`, so a collectors.d on the developer's machine cannot leak into the run.
- `agent-usage-claude-scanner-test.sh`: `--id`/`--name`/`mark`; a second account reading its own directory while leaving the shared trees to the stock account; `--shared-sessions` both ways; per-directory limits caches answering with their own numbers (seeded caches, no network). Invocations pin `CLAUDE_CONFIG_DIR`, which the developer's own session may export - found the hard way, on a machine where it is.
- `agent-usage-codex-scanner-test.sh`: the same identity and shared-tree cases for `CODEX_HOME`.
- New `agents-accounts-test.sh` drives `Main.qml` headless: the mark flowing into the display object, aggregation refusing a mark conflict while legacy snapshots still merge, the mark-aware synced lookup, and the local snapshot recording marks. Fails without each guard.
- The bar-widget contract fixture asserts a `claude-work` record with `mark: "claude"` resolves the Claude icon.

Verified on a running desktop with three real accounts: personal (Max 5x, live limits), a second Claude account, and Codex - three chips, each with its own plan line and totals, both Claude accounts wearing the Claude mark.

## Notes

Removing an account is removing its wrapper and the record it left in the usage directory; the README's new Accounts section documents the recipe, the shared-tree ownership, and that removal. Records written by older versions (no `mark`) display and sync exactly as before.

### DIFF
--- a/bin/omarchy-agent-usage-claude
+++ b/bin/omarchy-agent-usage-claude
@@ -1,6 +1,6 @@
 #!/usr/bin/python3
 # omarchy:summary=Print the Claude Code usage record as JSON
-# omarchy:args=[--force] [--limits-only]
+# omarchy:args=[--force] [--limits-only] [--id <id>] [--name <name>] [--shared-sessions <auto|on|off>]
 # omarchy:hidden=true
 """Collect Claude Code usage into one display-ready JSON record.
 
@@ -43,6 +43,33 @@ def config_dir() -> Path:
 
 def expand_path(value: str) -> Path:
   return Path(os.path.expandvars(os.path.expanduser(value))).resolve()
+
+
+def normalized_dir(value: object) -> str:
+  # Spelling, not identity on disk: collapse "~", "..", and trailing slashes
+  # but leave symlinks alone, so a symlinked alias of the stock directory
+  # reads as its own account rather than inheriting the stock one's claims.
+  return os.path.normpath(os.path.abspath(os.path.expanduser(str(value))))
+
+
+def on_stock_dir() -> bool:
+  return normalized_dir(os.environ.get("CLAUDE_CONFIG_DIR") or "~/.claude") == normalized_dir("~/.claude")
+
+
+# pi, omp, and opencode keep one transcript tree apiece, shared by every
+# Claude account on the machine — nothing in them says which CLAUDE_CONFIG_DIR
+# their tokens were burned against. Exactly one account may count them, or the
+# same tokens appear once per account. omarchy-agent-usage-update resolves the
+# owner across every enabled account (reading each one's preference through
+# --print-identity) and runs each collector with an explicit on or off; auto
+# is the answer for a collector run by hand, where the stock ~/.claude
+# account counts them.
+def include_shared_sessions(mode: str) -> bool:
+  if mode == "on":
+    return True
+  if mode == "off":
+    return False
+  return on_stock_dir()
 
 
 def cache_root() -> Path:
@@ -700,7 +727,7 @@ def probe_limits(access_token: str) -> dict[str, Any]:
   return {"ok": True, "limits": limits}
 
 
-def collect_limits(access_token: str, expires_at_ms: int, force: bool) -> dict[str, Any]:
+def collect_limits(access_token: str, expires_at_ms: int, force: bool, claude_dir: Path) -> dict[str, Any]:
   result = {"limits": [], "usageStatusText": "", "authHelpText": AUTH_HELP}
 
   if access_token == "":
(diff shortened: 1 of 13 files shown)


dhh:
Heads up: #6603 just landed on `quattro` and restructured this whole area, so this PR needs a rework before it can merge.

The `model-usage` plugin is gone. It's been renamed to `omarchy.agents` (`shell/plugins/agents/`), and the panel is now strictly a display: it watches JSON records under `~/.local/state/omarchy/agents/usage/` and draws whatever appears there. All the in-plugin machinery this PR touches — `providers/*.qml` adapters and `scripts/*_usage_scanner.py` — no longer exists.

Usage collection now lives in standalone collectors in `bin/`. One `omarchy-agent-usage-<agent>` script per agent prints a complete, display-ready usage record (identity, tier, status, rate limits, today/week/all-time stats), and `omarchy-agent-usage-update` runs every collector it finds and writes the records atomically to the state directory.

The upside: adding an agent no longer touches the panel at all. Ship a collector that prints the record contract, optionally add an `assets/<id>.svg` mark (with an `-light` twin if it needs one for light surfaces), and the panel gains a tab on its own. See `shell/plugins/agents/README.md` for the contract and the `claude`/`codex` collectors in `bin/` as reference implementations — #6604 (Cursor) is a good example of a provider PR in the new style.

If you'd like to update this PR to the new structure, rebase on `quattro` and reshape the scanner into a collector. Happy to answer questions here.

— Claude Fable 5, written on behalf of @dhh

PR #6517, opened 2026-08-02
Add a keyboard layout switch command with an OSD readout

kmpeeduwee:
## Why

Volume, brightness, and media keys all answer through the OSD, but switching the keyboard layout answers with nothing: bind `hyprctl switchxkblayout` and the only way to find out which layout you landed on is to type something and see what comes out. With two layouts that means guessing wrong half the time; with more, worse.

## What this adds

`omarchy-hyprland-keyboard-layout [next|prev|status]` — the same shape as `omarchy-audio-output-volume`: perform the change, then show the result on the OSD.

- `next` / `prev` dispatch `switchxkblayout all` and pop the OSD with the keyboard glyph and the new layout's name as xkb reports it — "English (US)", "Portuguese (Brazil)", "German (dead acute)", whatever is configured.
- `status` prints the active layout to stdout without switching or opening the OSD.

No default binding is added; users point whichever combo they already switch
with at it:

```lua
o.bind("SUPER + U", "Switch keyboard layout", "omarchy-hyprland-keyboard-layout", { locked = true })
```

`locked = true` is worth having here: the layout you type your lock-screen password in is the one place where being on the wrong one hurts most.

## Notes

- The layout is read back from the `main` keyboard rather than the first entry in `hyprctl devices` — power buttons and other keymap-less inputs also sit in `keyboards`, and `all` switches every real keyboard anyway, so the main one speaks for the lot.
- `hyprctl` answers the dispatch only after the switch lands, so the read that follows already sees the new keymap; no polling or delay is needed.
- The OSD already had a `keyboard` icon; no shell changes.

## Tests

`test/shell.d/keyboard-layout-test.sh` stubs `hyprctl` and `omarchy-osd` and covers: the default action dispatching `switchxkblayout all next` and showing the main keyboard's layout, `prev`, `status` printing without switching or opening the OSD, and an unknown action exiting non-zero. The fixture puts a keymap-less non-main "keyboard" first to pin the main-keyboard selection, and its keymap name carries a space and parentheses — the shape xkb names come in — to prove the name reaches the OSD intact.

Verified on a running desktop: switching layouts shows the OSD with the layout name in the same spot the volume OSD uses.

### DIFF
--- a/bin/omarchy-hyprland-keyboard-layout
+++ b/bin/omarchy-hyprland-keyboard-layout
@@ -1,0 +1,31 @@
+#!/bin/bash
+
+# omarchy:summary=Switch the keyboard layout and show it on the OSD
+# omarchy:args=[next|prev|status]
+# omarchy:examples=omarchy hyprland keyboard-layout | omarchy hyprland keyboard-layout prev | omarchy hyprland keyboard-layout status
+
+action="${1:-next}"
+
+case $action in
+  next | prev)
+    hyprctl switchxkblayout all "$action" >/dev/null
+    ;;
+  status) ;;
+  *)
+    echo "Usage: omarchy-hyprland-keyboard-layout [next|prev|status]" >&2
+    exit 1
+    ;;
+esac
+
+# hyprctl answers the dispatch only after the switch lands, so the read below
+# already sees the new keymap. The main keyboard speaks for the lot: `all`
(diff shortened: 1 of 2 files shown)


dhh:
I've instead added the default keyboard layout plugin to the menu, so you now have a visual indication of swapping bindings. It's only showing if you have 2+ bindings setup. Thanks for the suggestion ✌️

PR #6556, opened 2026-08-05
Bar "overlay windows" setting.

brianblakely:
This brings the QS bar closer to parity with the former Waybar. For folks like myself who hide/show the bar a lot, enabling this setting prevents nasty, laggy reflows.


dhh:
We fixed the jaggy on/off by improving the hide instead 👌

PR #6562, opened 2026-08-05
Fix T2 Mac suspend and fan defaults

### LINKED ISSUE #6559: [quattro only] Update fix-t2.sh to enhance T2 macs compatibility
### System details

MacBook Pro (16-inch 2019) T2

### What's wrong?

With the transition of arch's Watanare-t2 kernel to t2bce the KERNEL_CMDLINE defaults need to be updated (see https://kait2en.github.io/reference/installed-files Kernel Args section)

`KERNEL_CMDLINE[default]+=" intel_iommu=on iommu=pt pm_async=off mem_sleep_default=deep "`

This will fix the known suspend issues, enable faster suspend and resume times.

Also, tiny-dfr is not behaving properly with t2bce and the standard touchbar firmware works perfectly instead with sleep and resume, so this extra dependency can be removed.

Last but not least we need an additional Fan2 entry for MacBooks with two fans to avoid t2fanrd service crashing repeatedly.

Here's the suggested fixed file 

```
# Detect T2 MacBook models using PCI IDs
# Vendor: 106b (Apple), Device IDs: 1801 or 1802 (T2 Security Chip)
if lspci -nn | grep -q "106b:180[12]"; then
  echo "Detected MacBook with T2 chip. Installing support items..."

  omarchy-pkg-add \
    linux-t2 \
    linux-t2-headers \
    apple-t2-audio-config \
    apple-bcm-firmware \
    t2fanrd

  # Enable T2 services
  # Probably won't need that anymore
  systemctl enable t2fanrd.service

  mkdir -p /etc/modules-load.d
  {
    echo "t2bce_vhci"
    echo "hci_bcm4377"
  } > /etc/modules-load.d/t2.conf

  # linux-t2 7.1.4 replaced the apple-bce driver with t2bce; t2bce_vhci is the
  # virtual USB host controller the internal keyboard hangs off, and mkinitcpio
  # pulls t2bce_core/t2bce_dma in via module dependencies.
  mkdir -p /etc/mkinitcpio.conf.d
  echo "MODULES+=(t2bce_vhci usbhid hid_apple hid_generic xhci_pci xhci_hcd)" > \
    /etc/mkinitcpio.conf.d/apple-t2.conf

  mkdir -p /etc/modprobe.d
  cat > /etc/modprobe.d/brcmfmac.conf <<'EOF'
# Fix for T2 MacBook WiFi connectivity issues
options brcmfmac feature_disable=0x82000
EOF

  mkdir -p /etc/limine-entry-tool.d
  cat > /etc/limine-entry-tool.d/t2-mac.conf <<'EOF'
# Generated by Omarchy installer for T2 Mac support
# With the new t2bce these are the needed defaults for suspend. See https://kait2en.github.io/reference/installed-files/
KERNEL_CMDLINE[default]+=" intel_iommu=on iommu=pt pm_async=off mem_sleep_default=deep "
EOF

  # Added Fan2 to workaround a crashloop
  cat > /etc/t2fand.conf <<'EOF'
[Fan1]
low_temp=55
high_temp=75
speed_curve=exponential
always_full_speed=false
[Fan2]
low_temp=55
high_temp=75
speed_curve=exponential
always_full_speed=false
EOF
fi
```

dhh:
## Summary

- replace the obsolete `pcie_ports=compat` T2 kernel argument with `pm_async=off mem_sleep_default=deep`
- use the native Boot Camp-style Touch Bar by default and remove `tiny-dfr` from T2 setup and the ISO cache
- configure both possible MacBook fans while preserving Omarchy's linear fan curve
- migrate existing T2 installs and cover affected, idempotent, and non-T2 paths

Closes #6559

## Verification

- `bash test/shell.d/t2-hardware-test.sh`
- `./test/cli`
- Bash syntax validation and `git diff --check`
- remaining shell suite passes; the aggregate runner retains the unrelated pre-existing model-usage migration assertion failure
- fresh ISO acceptance was blocked before build by the harness's interactive sudo prompt


dhh:
Claude Fable review completed.

Findings addressed:

- added a machine-wide success marker so another user's pending migration cannot repeat the UKI rebuild before reboot
- added explicit interrupted-rebuild and pre-reboot multi-user regression coverage
- anchored the retry test's negative command match so paths containing `sed` or `tee` cannot cause a false failure

Fable also independently checked T2FanRD upstream: it reads only `Fan1..FanN` for detected fans, so `Fan2` is required on dual-fan models and ignored safely on single-fan models. No remaining actionable code findings were reported. Fresh-ISO and physical T2 validation remain outstanding as documented in the PR.

PR #6586, opened 2026-08-06
Skip Snapper configuration on a root it cannot support

tmn73:
## What's wrong

`install/config/snapper.sh` has no filesystem guard, so on a root that Snapper cannot support it fails and takes its caller down with it. `omarchy-upgrade-to-quattro` aborted at `configure_snapper_policy`, step 10 of 26, on my ext4 install:

```
==> Configuring Omarchy snapshot retention
Configuring Omarchy Snapper snapshot retention
Creating config failed (/sbin/chsnap not installed).
```

Everything from `configure_lock_authentication` onward never ran, including `apply_user_transition`, the dotfile migration. Migration `1781984677` fails the same way for the same users, and since `omarchy-migrate` stops at the first failure it also blocks every migration queued behind it.

## Why

```sh
snapper --no-dbus -c root create-config / >/dev/null 2>&1 || snapper -c root create-config / >/dev/null
```

`create-config` only works on a filesystem Snapper supports, which for Omarchy means btrfs. Every caller runs the script under `bash -euo pipefail`, so the failure is fatal rather than skipped. Nothing after that line means anything without a config: the template install, `/etc/conf.d/snapper` and both timers all describe snapshots that cannot exist.

The Omarchy installer defaults to btrfs, which is why this stays invisible on a fresh install. It only shows up on a root that was set up some other way.

## The fix

Read the root filesystem type and skip cleanly when it is not btrfs, before anything is written. `run_logged` runs these scripts with `bash -eE "$script"` rather than sourcing them, so `exit 0` ends this script only.

The type goes through `OMARCHY_SNAPPER_ROOT_FSTYPE`, matching the existing `OMARCHY_SNAPPER_CONFIG_PATH`, `OMARCHY_SNAPPER_CONF_PATH` and `OMARCHY_SNAPPER_TEMPLATE` overrides, so both branches are testable.

## Testing

`test/shell.d/snapper-test.sh` gains a case: a non-btrfs root exits 0 and writes neither the Snapper config nor `/etc/conf.d/snapper`. The existing configure case now pins `OMARCHY_SNAPPER_ROOT_FSTYPE=btrfs` so it keeps exercising the full path.

Verified on the affected machine (ext4 root): the script exited 1 before and exits 0 writing nothing after. Run as root the failure message is the `chsnap` one above; unprivileged it is `No permissions.` Either way it is non-zero and fatal to the caller.

`omarchy-pkgs checkout is available for packaging coverage` fails in this checkout for want of a sibling `omarchy-pkgs`. It fails the same way on an unmodified `quattro`, unrelated to this change.

### DIFF
--- a/install/config/snapper.sh
+++ b/install/config/snapper.sh
@@ -1,8 +1,20 @@
 SNAPPER_CONFIG_PATH="${OMARCHY_SNAPPER_CONFIG_PATH:-/etc/snapper/configs/root}"
 SNAPPER_CONF_PATH="${OMARCHY_SNAPPER_CONF_PATH:-/etc/conf.d/snapper}"
 template="${OMARCHY_SNAPPER_TEMPLATE:-${OMARCHY_PATH:-/usr/share/omarchy}/default/snapper/root}"
+root_fstype="${OMARCHY_SNAPPER_ROOT_FSTYPE:-$(findmnt -no FSTYPE --target / 2>/dev/null || true)}"
 
 echo "Configuring Omarchy Snapper snapshot retention"
+
+# snapper create-config only works on a filesystem it supports, which for Omarchy
+# means btrfs. On any other root it fails with "Creating config failed
+# (/sbin/chsnap not installed)", and every caller runs this under
+# bash -euo pipefail, so that failure aborts omarchy-upgrade-to-quattro partway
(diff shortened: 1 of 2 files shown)


dhh:
Omarchy will no longer offer a manual installation with Quattro, and all automated installations use btrfs. So don't think this guard makes sense.

PR #6588, opened 2026-08-06
Prevent voxtype focus drift during dictation output

chrispository:
# Prevent voxtype focus drift during dictation output

## Problem

Omarchy ships voxtype with `mode = "type"`, which emits synthetic keystrokes through `wtype`. Hyprland's default `input:follow_mouse = 1` continuously moves keyboard focus to the window under the cursor.

During a long dictation, typing is not instantaneous — at `type_delay_ms = 1`, a 60-word sentence is ~300ms+ of continuous synthetic typing. If the pointer moves toward another window while that stream is in flight, focus can shift mid-stream and the remaining keystrokes land in the wrong window.

This is a focus-drift race, not a voxtype transcription bug. Any synthetic-keystroke tool has the same behavior.

## Fix

Voxtype 0.7.5 supports output hooks. Add `pre_output_command` / `post_output_command` to the default config so dictation output temporarily disables `follow_mouse` and restores the user's current value afterward.

The pre hook snapshots the live `input:follow_mouse` value via `hyprctl`, disables it for the typing window, and the post hook restores the snapshot:

```toml
pre_output_command = "omarchy-hyprland-follow-mouse-guard pause"
post_output_command = "omarchy-hyprland-follow-mouse-guard resume"
```

This changes the setting only at runtime via `hyprctl keyword`; it does not edit the Hyprland config on disk. Restoring the snapshot means users who override `follow_mouse` in their own Hyprland config keep their preference.

## Verification

Tested on Hyprland 0.55.4:

```text
before:      input:follow_mouse = 1
set to:      input:follow_mouse = 0
restored to: input:follow_mouse = 1
```

Manual test: focus a text field, hold F9, dictate a long sentence, release F9, and move the mouse to another window while text is being emitted. Text should remain in the original window, and `follow_mouse` should be `1` again afterward.

## Notes

- Keeps `mode = "type"` — no clipboard clobbering, no paste-semantics change.
- Snapshot lives in `$XDG_RUNTIME_DIR/omarchy/voxtype-follow-mouse` for the duration of output.


chrispository:
Superseded by a cleaner single-commit draft with the same fix.

PR #6616, opened 2026-08-07
Scope BROWSER to interactive shells so browsers can set themselves as default

### LINKED ISSUE #6590: Browsers' native "Set as default" button fails under the Omarchy session
### System details

omarchy quattro

### What's wrong?

❯ xdg-settings set default-web-browser vivaldi-stable.desktop
xdg-settings: $BROWSER is set and can't be changed with xdg-settings

Browsers' own "Set as default" buttons go through xdg-settings, so they fail for every browser — Vivaldi, Firefox, Zen, and the Chromium family alike. In practice, pressing "Set as default" in Vivaldi silently fails: the default stays on Chromium, so omarchy-launch-browser and web apps keep opening Chromium.

#6356   added env -u BROWSER to Omarchy's own browser commands, which makes omarchy default browser <x> work — but it does not fully resolve this case: $BROWSER is still exported in the session, so the browsers' own buttons still cannot change the default.

dhh:
Fixes #6590

Omarchy exports `BROWSER=omarchy-launch-browser` into the entire uwsm session. On a generic desktop like Hyprland, `xdg-settings` hard-refuses `set default-web-browser` whenever `$BROWSER` is set (`set_browser_generic` errors with "\$BROWSER is set and can't be changed with xdg-settings"). Since browsers' native "Set as default" buttons go through `xdg-settings`, they all silently failed under the Omarchy session — Vivaldi, Firefox, Zen, and the Chromium family alike. #6356 worked around this for Omarchy's own commands with `env -u BROWSER`, but couldn't help third-party callers.

The export only exists for terminal programs (like `gh`) to open URLs detached from the terminal process tree — nothing in the graphical session reads it (Omarchy's own code only ever unsets it, and GUI apps resolve URLs through the MIME handlers that `xdg-settings set` maintains). So this moves the export from `default/uwsm/default` to `default/bash/envs`, guarded as `${BROWSER:-...}` like the existing `EDITOR` export so user overrides still win:

- Browsers launched from the app launcher no longer see `$BROWSER`, so their "Set as default" buttons work.
- `gh` and other terminal programs still get the detached-launch behavior, since interactive shells source `envs`.
- Omarchy's own `env -u BROWSER` call sites are unaffected.

Known edge that remains: a browser launched *from a terminal* still inherits the shell's `BROWSER`, so its button still fails there; `omarchy default browser` covers that case.

New `test/shell.d/browser-env-test.sh` pins the default and inherited-override behavior in `envs` and asserts the uwsm session env no longer exports `BROWSER`.

https://claude.ai/code/session_01AUyYEmcFGbi3YcsqhZAjkT

### DIFF
--- a/default/bash/envs
+++ b/default/bash/envs
@@ -1,6 +1,11 @@
 # Editor used by CLI
 export EDITOR="${EDITOR:-omarchy-launch-editor --inline}"
 export SUDO_EDITOR="$EDITOR"
+
+# Used by terminal programs (like gh) to open URLs detached from the terminal
+# process tree. Shell-scoped on purpose: exporting BROWSER session-wide makes
+# xdg-settings refuse to change the default browser.
+export BROWSER="${BROWSER:-omarchy-launch-browser}"
 export BAT_THEME=ansi
 
 # Color man pages with bat
--- a/default/uwsm/default
+++ b/default/uwsm/default
@@ -3,8 +3,9 @@
 # Install other terminals via Install > Terminal
 export TERMINAL=xdg-terminal-exec
 
-# Used by terminal programs (like gh) to open URLs detached from the terminal process tree
-export BROWSER=omarchy-launch-browser
+# BROWSER is intentionally not exported here: session-wide, it makes xdg-settings
+# refuse to change the default browser, which breaks the browsers' own "Set as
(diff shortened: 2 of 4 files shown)


felixzsh:
this didnt actually solve the issue

edit: it did but a migration was needed for my case

PR #6626, opened 2026-08-08
Fix Battle.net micro-stutter by persisting DXVK shader cache

GerritWanderer:
## Summary
- `omarchy-launch-battlenet` never set `DXVK_STATE_CACHE_PATH`, so the WINEPREFIX's shader cache never persisted to disk — every already-compiled DXVK shader had to recompile from scratch on each launch, producing micro-stutter (most noticeable in StarCraft II).
- Adds `DXVK_STATE_CACHE_PATH="$PREFIX"` and `STAGING_SHARED_MEMORY=1` to the env passed to `umu-run`, matching what a working Lutris install of the same `umu-run` + GE-Proton stack sets by default.

## Test plan
- [x] A/B tested against a fresh Lutris install of Battle.net/SC2 on the same GE-Proton/umu-run stack (which sets these and did not stutter) vs. the unpatched manual launch (which did).
- [x] Applied the same two env vars to a local prefix and confirmed the stutter disappeared.
- [x] `bash -n bin/omarchy-launch-battlenet` — syntax OK.
- [ ] Would appreciate a sanity check from anyone else running Battle.net games through this launcher, since I only tested against SC2.

### DIFF
--- a/bin/omarchy-launch-battlenet
+++ b/bin/omarchy-launch-battlenet
@@ -42,6 +42,8 @@ env_args=(
   PROTONPATH=GE-Proton
   GAMEID=umu-battlenet
   PROTON_VERB=run
+  DXVK_STATE_CACHE_PATH="$PREFIX"
+  STAGING_SHARED_MEMORY=1
 )
 (( with_mangohud )) && env_args+=(MANGOHUD=1)
 


dhh:
Thanks for digging into this — the stutter you measured is real, but I think the
diagnosis landed on the wrong two variables. Both are no-ops on the stack this
launcher pins, so merging as-is would ship a fix that doesn't do anything.

**`DXVK_STATE_CACHE_PATH`**

The launcher pins `PROTONPATH=GE-Proton`, which resolves to the newest GE-Proton
umu can fetch. DXVK's state cache has been unused since DXVK 2.0 on any driver
exposing `VK_EXT_graphics_pipeline_library` (all current Mesa RADV/ANV and
NVIDIA), and it was [removed outright in DXVK
2.7](https://github.com/doitsujin/dxvk/releases/tag/v2.7) — the variable isn't
read at all anymore. GE-Proton has shipped DXVK 2.x since GE-Proton 8, so on
every supported configuration no `.dxvk-cache` file is ever written.

There's a second problem underneath that one. DXVK reads the variable from
inside the Wine process and treats it as a Windows path. `/home/<user>/Games/battlenet`
is a *rooted* DOS path, so Wine resolves it against the current drive — our cwd
sits under `$PREFIX/drive_c/...`, making it `C:\home\<user>\Games\battlenet`.
The fallback is a non-recursive `CreateDirectoryW`, which fails because
`C:\home` doesn't exist, and the write is silently dropped. I suspect this is
exactly why it appeared to work under Lutris: there the game cwd lands on the
`Z:` mapping, where the same string resolves to the real Unix path.

**`STAGING_SHARED_MEMORY`**

The wineserver shared-memory patchset that provided this no longer exists in
wine-staging, and neither Proton's launch script nor upstream Wine reads the
name. It's also unrelated to shader caching, so it would read as load-bearing to
the next person touching this file.

**Where I'd look instead**

Something did differ between your Lutris install and the manual launch. On DXVK
2.x the cache that actually matters is Mesa's, so `MESA_SHADER_CACHE_DIR` /
`MESA_SHADER_CACHE_MAX_SIZE` are the first place I'd check — if the manual
launch ends up with a different `HOME` or a cache dir that gets evicted, you'd
see recompile stutter that Lutris doesn't have. Worth diffing the full env
Lutris hands to `umu-run` against ours; gamemode, `mesa_glthread`, and
fsync/ntsync settings are the other likely candidates.

One more thing for whenever we land the real fix: `bin/omarchy-install-gaming-battlenet`
sets the same env block, so anything genuinely needed belongs in both places.
And heads up that the green check here isn't evidence either way —
`test/shell.d/battlenet-test.sh` only covers the desktop-file wiring, not the
env list.

Issue #6628, opened 2026-08-08
Lock shell dies during normal idle->lock; session permanently locked, reboot required


akashgagda:
## Summary

The Omarchy shell (a single `quickshell` process hosting the `omarchy.idle` and `omarchy.lock` services) exits while the session is locked during a **normal idle → lockscreen cycle with no user interaction and no idle inhibitor active**. Because ext-session-lock leaves the session permanently locked when the lock client dies, Hyprland shows its "lockscreen app died" failsafe and the machine is unusable until a reboot.

## Expected behavior

The idle → lockscreen → lock cycle should not end in a dead shell. At minimum, if the lock client dies, the session should be recoverable (respawn / re-acquire the lock) instead of requiring a reboot, and diagnostics should be available afterwards.

## Reproduction steps

1. Start a session and leave the machine untouched (no media playing, no idle inhibitor, nothing fullscreen).
2. Idle timeout fires at `idle.screensaver` (150s), then `idle.lock` (300s). The lock screen appears normally.
3. After some time the lock surface vanishes and the **Hyprland gray failsafe** ("lockscreen app died") appears.
4. Nothing recovers the display. Only a reboot (or TTY + reboot) restores the system.

## Environment

- OS: Arch Linux, kernel 7.1.6
- CPU: Intel Core i5-13500 (13th gen)
- GPU: Intel UHD Graphics 770 (iGPU only)
- `hyprland` 0.56.2-1
- `quickshell-git` 0.3.0.r18.g10b439f-3 (quickshell-git AUR, built 2026-07-25, since omarchy requires the git build)
- `omarchy-dev` 4.0.0.r1569.g007d6fc-1
- `idle.lock` = 300, `idle.screensaver` = 150

## Diagnostics / observations

- **No coredump.** The previous session's journal contains **no** `systemd-coredump` / signal-crash entry for `quickshell`, and no Hyprland crash report — the death looks like an exit without a signal (or the dump was lost).
- **No shell log.** `quickshell -n -p $OMARCHY_PATH/shell` is launched by Hyprland autostart via `hl.exec_cmd()` with **no output redirection**, so the idle/lock event trail is not persisted anywhere and is lost on reboot. This made root-cause analysis impossible from the shell side.
- Hyprland journal shows `error in client communication (pid …)` ~15 min into the (short, otherwise quiet) session, which may be the shell being killed by the compositor.
[...]

Issue #61, opened 2025-07-05
Add custom decryption boot screen


dhh:
Would be great to theme the default "enter your disk decryption" screen at boot. It's pretty boring! We could use the ascii art as a default, then have "Enter Password: " beneath it. Like the lock screen. Maybe we could even put that ascii logo somewhere that would make it easy to personalize?

Mohamedsayhii:
@dhh thought this was pulled to the latest update, why the revert if you don't mind me asking.

dhh:
We are refining it a bit. Will be back in the next version, most likely.

PR #6630, opened 2026-08-08
Make a dead lock client diagnosable and recoverable

dhh:
Addresses the unrecoverable half of #6628: when the Omarchy shell dies while the session is locked, Hyprland's ext-session-lock failsafe keeps the session locked with no lock client left, and today the only way back in is a reboot — with no diagnostics left behind to explain what killed the shell.

The crash itself is upstream (quickshell-mirror/quickshell#61/#76 are still open), but Omarchy owned two gaps that turned it into "reboot required with zero logs":

**Persist the shell log across sessions.** Quickshell only logs to its instance runtime dir on tmpfs, so the idle/lock event trail vanished with every reboot. The shell now launches through `omarchy-launch-shell`, which pipes stdout/stderr into the journal under the `omarchy-shell` tag — bounded, timestamped, and persistent — and `omarchy-debug-idle` surfaces that log, so the next occurrence of #6628 is diagnosable after the fact.

**Recover the dead-lock-client failsafe.** `omarchy-restart-shell` refused to run whenever the session was locked, which blocked recovery in exactly the state that needs it. The refusal is now gated on the lock service actually holding (or acquiring) the lock: a live locker is preserved as before, but a dead shell — or a crash-handler relaunch that holds no lock — no longer blocks the restart. Recovery then re-requests the session lock and polls until the shell reports it secure (the same secure-poll `omarchy-system-sleep-lock` uses), so the command can't claim success while the failsafe is still up, and the user authenticates out of it from a TTY or SSH instead of rebooting. `misc.allow_session_lock_restore` is enabled so Hyprland accepts the replacement lock client; without it the compositor denies the re-lock and the recovery path is dead on arrival.

The session stays compositor-locked for the whole recovery window — ext-session-lock keeps the failsafe up until a new client locks and the user authenticates — so nothing is exposed if recovery fails partway.

### DIFF
--- a/bin/omarchy-debug-idle
+++ b/bin/omarchy-debug-idle
@@ -27,6 +27,9 @@ section "Recent idle logs"
 quickshell --no-color log -p "$OMARCHY_PATH/shell" --any-display --tail "$lines" --log-times -r 'quickshell.wayland.idle_notify=true' 2>&1 \
   | grep -Ei 'omarchy idle|idle_notify|screensaver|lock|error|warn|failed' || true
 
+section "Persisted shell log"
+journalctl -t omarchy-shell -n "$lines" --no-pager --quiet 2>/dev/null || true
+
 section "Relevant processes"
 ps -eo pid=,args= \
   | grep -E 'quickshell -n -p|omarchy-system-sleep-monitor|systemd-inhibit.*Lock screen before suspend|org\.omarchy\.screensaver|omarchy-screensaver|(^|/| )tte( |$)' \
--- a/bin/omarchy-launch-shell
+++ b/bin/omarchy-launch-shell
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+# omarchy:summary=Launch the Omarchy shell with its log kept in the journal
+# omarchy:hidden=true
+
+# Quickshell only logs to its instance runtime dir (tmpfs), so when the shell
+# dies the idle/lock event trail is gone after a reboot. The journal keeps it
+# across sessions, bounded and timestamped, under the omarchy-shell tag.
+exec systemd-cat -t omarchy-shell -- quickshell -n -p "$OMARCHY_PATH/shell"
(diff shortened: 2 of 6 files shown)


akashgagda (approved):
General approval. This matches our captured diagnosis on #6628 exactly:

- **Persisted shell log:** our crash reproduced only because we added a user-side wrapper to keep the quickshell log; having the shell's idle/lock trail in the journal (`omarchy-launch-shell` + `journalctl -t omarchy-shell`) closes the "died → reboot → zero diagnostics" gap for everyone.
- **Dead-lock recovery:** the current `omarchy-restart-shell` hard-refuses on any `LOCK` session, which is precisely the state that's unrecoverable. Gating on `omarchy-shell lock status` (`secure`/`requested`) instead of a blind refuse preserves a live locker and allows recovery when the locker is gone, and re-requesting + polling to `secure` mirrors the existing sleep-lock pattern.
- **`misc.allow_session_lock_restore = true`** is required for Hyprland to accept the replacement lock client — validated it's a recognized 0.56.2 option.

Scope note: this fixes the recoverable half. The crash cause itself remains upstream (quickshell #61/#76 family, with a monitor-flap/EGL trigger on our Intel HDMI system), so this gets users out of the reboot trap and keeps evidence, which is the right omarchy-owned slice. Tests updated in-repo; I'll run the hermetic shell suite against this branch.

akashgagda (approved):
General approval. This matches our captured diagnosis on #6628 exactly:

- **Persisted shell log:** our crash reproduced only because we added a user-side wrapper to keep the quickshell log; having the shell's idle/lock trail in the journal (`omarchy-launch-shell` + `journalctl -t omarchy-shell`) closes the "died → reboot → zero diagnostics" gap for everyone.
- **Dead-lock recovery:** the current `omarchy-restart-shell` hard-refuses on any `LOCK` session, which is precisely the state that's unrecoverable. Gating on `omarchy-shell lock status` (`secure`/`requested`) instead of a blind refuse preserves a live locker and allows recovery when the locker is gone, and re-requesting + polling to `secure` mirrors the existing sleep-lock pattern.
- **`misc.allow_session_lock_restore = true`** is required for Hyprland to accept the replacement lock client — validated it's a recognized 0.56.2 option.

Scope note: this fixes the recoverable half. The crash cause itself remains upstream (quickshell #61/#76 family, with a monitor-flap/EGL trigger on our Intel HDMI system), so this gets users out of the reboot trap and keeps evidence, which is the right omarchy-owned slice. Tests updated in-repo; I'll run the hermetic shell suite against this branch.

PR #6635, opened 2026-08-08
Add Dark Ages theme

stappmus:
## What

Adds Dark Ages, a true-black, OLED-friendly theme built around public-domain classical paintings.

## Why

I wanted a dark theme that would take advantage of an OLED display, but with classical paintings instead of digital art.

I initially tried converting the paintings to black and white and making large parts of them completely black. I experimented with that approach for quite a while, but could never get it to look right. In the end, I decided to keep the paintings in their original colors while making the main theme background completely black.

The paintings look especially beautiful at full brightness on the XPS 16's OLED display.

## Details

- Uses `#000000` for the main, dark, and darker background colors.
- Includes five paintings in their original color and tonality.
- Documents the artwork and public-domain sources in `themes/dark-ages/SOURCES.md`.
- Includes desktop and unlock-screen previews.

## Preview

Dark Ages preview

## Wallpaper previews

The five Dark Ages classical-painting wallpapers

## Verification

- Applied and visually inspected the clean contribution directory on the running desktop.
- Ran `./test/all` successfully.


dhh:
Great looking theme. Please at it to the #omarchy-themes-showcase room on the Discord. But I think we're pretty well covered on these tones in core for the moment.

PR #6639, opened 2026-08-08
Live-sync shell look from Hyprland and add hook

felixzsh:
Currently, when Hyprland's look is updated (for example, rounding), we need to run omarchy restart shell to see those changes in the shell. This PR adds an event listener for Hyprland's configreloaded event via its instance socket, so QML's Style.refresh() runs on every config reload.

also adds a shell-style-changed hook, this will be helpful for my pr #6593 to also support live changes for things like hyprland rounding config

constraints, theme switches also fire this hook, because its detected as a configreloaded event.


felixzsh:
closed in favor of #6662

PR #6641, opened 2026-08-08
Quattro: Route app audio to individual outputs

ssupt:
## Summary

- add an output selector to each live application stream when multiple outputs are available
- let applications either follow the system default or remain pinned to a selected output
- preserve explicit application routes when the global default output changes
- support pointer and keyboard interaction with placement matching every bar edge

## Why

Quattro moved stream naming and routing from Wiremix into the native audio panel. The native panel already identifies live application streams and controls their volume, but output selection is currently global: sending one application to another output still requires a terminal command or an external mixer.

This completes that native routing path without cluttering the quick panel. The control only appears when more than one output is available, and the existing controls keep their dedicated behavior:

- the speaker icon mutes the application
- the slider controls its volume
- clicking the application source box—apart from the mute button and volume slider—opens its output menu

Choosing **Follow default output** makes the stream follow future default-output changes. Choosing a named output pins that application to the device instead.

## Implementation

Application streams are mapped to their current sinks through PipeWire object serials rather than display-name matching. PipeWire target metadata distinguishes explicit routes from streams following the default, including the otherwise ambiguous case where an application is explicitly pinned to the current default device.

When the system default changes, only application streams following the previous default are moved. Explicit routes, EasyEffects, and non-application processing streams remain in place.

The output popup opens away from the bar, mirrors correctly between left and right layouts, stays within the display bounds, and supports up to eight visible entries before scrolling. The active player remains bold, while the row fill is reserved for pointer or keyboard focus so routing focus stays unambiguous.

## Preview
image

## Verification

- `./test/cli`
- `qmllint -I shell shell/Ui/Dropdown.qml`
- manually tested routing, pinning, and following the default with two and three physical outputs
- manually tested pointer and keyboard selection across side and top bar layouts

### DIFF
--- a/bin/omarchy-audio-output-set-default
+++ b/bin/omarchy-audio-output-set-default
@@ -1,31 +1,54 @@
 #!/bin/bash
 
-# omarchy:summary=Set the default audio output and move active streams
-# omarchy:args=<node-id> <sink-name>
+# omarchy:summary=Set the default audio output and move streams following it
+# omarchy:args=<node-id> <sink-name> [current-sink-name]
 # omarchy:examples=omarchy audio output set default 42 alsa_output.pci-0000_00_1f.3.analog-stereo
 
 node_id=${1:-}
 sink_name=${2:-}
+old_sink_name=${3:-}
 
-if [[ -z $node_id || -z $sink_name ]]; then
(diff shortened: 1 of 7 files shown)


bjarneo:
Automated duplication check: this pull request looks similar to #7037, which covers the same per-stream audio routing. I keep that one open and close this one to consolidate review.

PR #6642, opened 2026-08-08
Keep the lock blank from turning the display off (DPMS link-drop crash)

akashgagda:
## Summary

Follow-up to #6628 / #6630. The lock screen blanks the display after locking via `hyprctl dpms disable`, and on some monitors (notably over HDMI) DPMS-off makes the connector physically drop — aquamarine tears the output down under the lock surface, Quickshell hits an EGL fatal (`Could not create EGL surface … Invalid argument`) and the whole shell exits **while still holding ext-session-lock**, forcing a reboot.

This ships the DPMS-off workaround as a config option rather than a hardcoded change:

- `lock.displayBlank` (default `true` = unchanged behavior).
- Setting it to `false` keeps the keyboard blank but skips the display off.

Verified live on an affected system (Lenovo L24q over HDMI, i915): with `lock.displayBlank: false`, a full lock → blank-window → unlock cycle produces **0 connector disconnects** and the shell survives (previously a reproducible DPMS drop + crash every lock).

## Scope

The root cause is upstream (Hyprland/aquamarine escalating DPMS-off into a full disconnect; Quickshell qFataling on lock-surface output loss — quickshell #61/#76 family). `allow_session_lock_restore` + journal-persisted shell log + dead-lock restart recovery are handled by #6630. This PR is the omarchy-owned user-facing mitigation so affected setups stop forcing reboots.

## Notes

`shell.json` is canonical with no deep-merge, so a config lacking the `lock` key keeps the `true` default (verified against the QML resolution).

### DIFF
--- a/shell/plugins/lock/Service.qml
+++ b/shell/plugins/lock/Service.qml
@@ -16,6 +16,11 @@ Item {
   readonly property string userName: Quickshell.env("USER") || Quickshell.env("LOGNAME")
   readonly property string currentBackgroundLink: stateHome + "/omarchy/current/background"
 
+  // The lock blank only dims the keyboard backlight. Blanking the display
+  // (Hyprland DPMS-off) makes some monitors drop their link, tearing the
+  // output out from under the lock surface and killing the shell while it
+  // holds the session lock; the lock screen already covers the desktop, and
+  // real display-off still happens at suspend/lid-close.
   property bool lockRequested: false
   property bool pendingSessionLock: false
   property bool authenticatingPassword: false
@@ -368,7 +373,7 @@ Item {
 
   Process {
     id: blankProcess
-    command: ["bash", "-c", "omarchy-brightness-keyboard off; omarchy-brightness-display off"]
+    command: ["bash", "-c", "omarchy-brightness-keyboard off"]
   }
 
   Timer {


dhh:
Don't like the idea of this being configurable. We need to find a solution that just works. Too esoteric to find this.

akashgagda:
Agreed — that's why I reworked this. The configurable flag is gone; the PR now unconditionally drops the DPMS display-off from the lock blank.

The root cause is `omarchy-brightness-display off` (Hyprland DPMS disable) during the lock blank. On some monitors that makes the display physically drop its link — the Lenovo L24q here does — so aquamarine sees `Connector HDMI-A-2 disconnected` and tears the output out from under the lock surface. Quickshell then hits a fatal EGL error and exits (rc 255) while still holding the session lock, which ends in the gray lockscreen failsafe and needs a reboot.

The reworked blank dims only the keyboard backlight. The lock surface already covers the desktop (blurred wallpaper + password field), real display-off still happens at suspend / lid-close via `omarchy-system-sleep-lock`, and nothing in the shell DPMS-blanks on idle, so nothing is lost.

Verified live: with the DPMS off removed the connector never drops — 0 disconnects, 0 DPMS events, and the shell held a secure session lock well past the former 5s blank on the very monitor that normally disconnects. Keeping that single line out of the blank avoids the crash entirely, no per-hardware opt-in.

stappmus:
Confirmed the same failure on NVIDIA RTX 3070 with nvidia-open 610.57.04, Hyprland 0.56.2, and quickshell-git 0.3.0.r20. Four crashes followed the identical sequence: DPMS leaves no outputs, lock surfaces lose their Wayland output, EGL creation fails, and Omarchy Shell exits 255. Applied this PR's keyboard-only blank through a cloned lock plugin; it loaded successfully and avoids the triggering DPMS dispatch.

Tested by GPT-5.4 via Codex.

GPT 5.4 is just a Sol seizure, dont mind me

stappmus:
Follow-up after testing on the NVIDIA system: removing display DPMS is not a complete fix because it regresses the expected screen-off behavior. I reverted that workaround and tested a narrower fix against the exact installed Quickshell revision (`28771c7`).

The output-removal path in `WlSessionLock::updateSurfaces()` removes the surface from its map but only calls `surface->deleteLater()`. Its `QQuickWindow` and native EGL/Wayland resources therefore remain alive until the next event-loop cleanup, long enough for Qt to render against the removed output and terminate the process. The minimal fix is to disown and destroy the window immediately while continuing to defer QObject cleanup:

```cpp
this->surfaces.remove(screen);
auto* window = surface->disownWindow();
window->destroy();
window->deleteLater();
surface->deleteLater();
```

Validation on RTX 3070 / nvidia-open 610.57.04 / Hyprland 0.56.2 / Qt 6.11.1:
- patched Quickshell built successfully from revision `28771c7`;
- Omarchy retained its normal `omarchy-brightness-display off` behavior;
- controlled lock + forced DPMS-off remained secure for 47 seconds;
- display woke and unlocked normally;
- the same Quickshell PID survived, with no EGL failure, fatal Wayland error, or shell relaunch.

This appears to belong in Quickshell rather than Omarchy. Quickshell explicitly disallows AI-authored code PRs, so I have not opened one there. I recommend not merging the global DPMS removal in this PR.

Investigated by GPT-5.4 via Codex.

howdeploy:
Confirmed the same lock/DPMS output-removal sequence on a QXL guest, with a more severe recovery path than a transient physical-monitor link drop.

System: Omarchy 4.0.1-1, Hyprland 0.56.2-1, Quickshell 0.3.1-1, Red Hat QXL `[1b36:0100]`. The installed lock service and `omarchy-brightness-display` match current `quattro` byte-for-byte.

Two independent boots recorded the same timing:

```text
02:02:36.341  lock-system: lock-timeout
02:02:36.516  lock-requested
02:02:44.201  There are no outputs - creating placeholder screen
02:02:44.392  Not creating lock surface ... not backed by a valid wayland output

09:14:58.570  lock-system: lock-timeout
09:14:58.701  lock-requested
09:15:05.903  There are no outputs - creating placeholder screen
09:15:06.084  Not creating lock surface ... not backed by a valid wayland output
```

That 7–8 second interval lines up with the lock service's 5-second blank timer followed by `omarchy-brightness-display off`. Once triggered, all four QXL `Virtual-*` connectors reported `disconnected`; the guest stayed on the FALLBACK output until the VM was rebooted.

As a local diagnostic mitigation I intercepted only the DPMS-disable dispatch while passing every other `hyprctl` command through. Later lock attempts reached the same blank path (the blocked dispatches were logged), but `Virtual-1` remained connected and the no-output signature did not recur.

This confirms QXL as another affected output-removal path. It does not resolve the design question raised above: suppressing display-off avoids the trigger, while the immediate window/resource destruction in Quickshell would preserve normal DPMS behavior and address the unsafe output lifecycle directly.

SoubranMx:
Confirmed on a current Quattro desktop: Omarchy 4.0.3-1, Hyprland 0.56.2-2, Quickshell 0.3.1, NVIDIA RTX 4070 SUPER, and one active LG ultrawide on DisplayPort (`DP-2`).

With the HDMI TV disabled, `omarchy system lock` reaches `secure=true`, then the five-second blank timer produces `There are no outputs - creating placeholder screen` and `Not creating lock surface ... not backed by a valid wayland output`. The session remains securely locked but has no usable authentication UI. It reproduces on the first DPMS cycle, and there are no NVIDIA DRM or kernel errors at the event.

This makes `lock.displayBlank: false` the right user-facing mitigation for this setup. Thanks for proposing it.

PR #6643, opened 2026-08-08
Guard theme commands against path traversal

shrijit37:
`omarchy theme remove` interpolates the theme name straight into an `rm -rf`:

```bash
THEME_PATH="$THEMES_DIR/$THEME_NAME"
rm -rf "$THEME_PATH"
```

An unvalidated name like `..` resolves to `~/.config`, so `omarchy theme remove ..` deletes the user's entire `~/.config/omarchy` directory. `omarchy theme set` and `omarchy-plymouth-set-by-theme` have the same hole, walking out of the themes directories into `cp -r` and a `sudo omarchy-plymouth-set` call.

Reject any name that isn't a plain file name — `^[A-Za-z0-9][A-Za-z0-9._-]*$` — before a destructive path is reached, matching the existing `valid_plugin_id` guard in the plugin commands. No legitimate theme is named `..`, so this only blocks garbage.

Added regression tests for `theme remove ..`/`../..`/`.`, `theme set ..`/`../..`, and `plymouth-set-by-theme ..`/`../..`; `./test/cli` passes 108 tests.

### DIFF
--- a/bin/omarchy-plymouth-set-by-theme
+++ b/bin/omarchy-plymouth-set-by-theme
@@ -14,6 +14,13 @@ fi
 
 theme=$1
 
+# Theme names are simple file names; reject anything that could walk out of
+# the themes directories before it reaches the sudo omarchy-plymouth-set.
+if [[ ! $theme =~ ^[A-Za-z0-9][A-Za-z0-9._-]*$ ]]; then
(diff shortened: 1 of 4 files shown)


dhh:
Appreciate the idea, but I find this overly defensive.

PR #6645, opened 2026-08-08
Add Hermes usage to agents monitor

okurmustafa:
## Summary

- add a collector for token usage stored in Hermes' local SQLite session ledger
- preserve model switches and auxiliary-task usage by billing provider and model
- let Hermes supply human-readable provider/model labels from its local catalogue
- carry optional model labels through the agents widget and cross-device snapshots

## Why

Hermes can use several providers and models within the same session, so treating it like a single-subscription agent would lose important attribution. This adds one Hermes tab while keeping each provider/model route distinct and avoids presenting a misleading unified quota.

This is a follow-up to #6644, but it is intentionally independent: existing Hermes users can benefit even if the installation integration is not merged.

The collector reads `~/.hermes/state.db` and named-profile databases in read-only mode. It performs no paid model calls and uses Hermes' local model metadata without requiring a network request. Unknown or custom routes fall back to the exact identifiers Hermes recorded.

## Validation

- `bash test/shell.d/agent-usage-hermes-scanner-test.sh`
- Claude, Codex, Fireworks, and usage-update collector tests
- `./test/cli`
- live Quickshell load and panel layout check with multi-provider synthetic usage

### DIFF
--- a/bin/omarchy-agent-usage-hermes
+++ b/bin/omarchy-agent-usage-hermes
@@ -1,0 +1,398 @@
+#!/usr/bin/python3
+# omarchy:summary=Print the Hermes usage record as JSON
+# omarchy:args=[--force] [--limits-only]
+# omarchy:hidden=true
+"""Collect Hermes model usage into one display-ready JSON record.
+
+Hermes can switch providers and models inside one session, and auxiliary
+tasks can use a different route from the main agent. Read its SQLite usage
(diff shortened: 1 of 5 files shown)


bjarneo:
Automated duplication check: this pull request looks similar to #9871, which covers the same Hermes usage collector. I keep that one open and close this one to consolidate review.

Issue #6608, opened 2026-08-07
Follow up to #6559 - update  fix-t2.sh and ‎migrations/1785944594.sh to not use grep -q


twilightresonance86:
### System details

MacBook Pro 16-inch T2

### What's wrong?

Follow up to #6559 - update  fix-t2.sh and ‎migrations/1785944594.sh to not use grep -q

The following 

```
if ! lspci -nn | grep -q "106b:180[12]"; then
  exit 0
fi

```
would cause the t2-fix and migration to exit prematurely probably due to pipefail.

For example:
```
~ ❯ set -o pipefail
~ ❯ lspci -nn | grep -q "106b:180[12]"; echo $?
141

~ ❯ lspci -nn | grep "106b:180[12]"; echo $?
04:00.1 Non-VGA unclassified device [0000]: Apple Inc. T2 Bridge Controller [106b:1801] (rev 01)
04:00.2 Non-VGA unclassified device [0000]: Apple Inc. T2 Secure Enclave Processor [106b:1802] (rev 01)
0

~ ❯ lspci -nn | grep "106b:180[12]" > /dev/null; echo $?
0

~ ❯ lspci -nn | grep "aaa" > /dev/null; echo $?
1

~ ❯ 
```

The fix is simply redirect to /dev/null:
```
if ! lspci -nn | grep "106b:180[12]" > /dev/null; then
  exit 0
fi
```

PR #6652, opened 2026-08-09
Apply the Broadcom Wi-Fi quirk to Macs without a T2

cupatea:
Issue:
- Old MacBook Pros (pre-T2) miss out on the WiFi fix, because they aren't
  detected as machines that need it. They have the same hardware issue as
  T2 Macs, so it needs to be fixed at install time too.

Fix:
- Gate on an Apple machine with a Broadcom wireless part, instead of on the
  T2 chip. The bug is in the Broadcom firmware, not the T2 bridge.

Checked:
- Manually applying the fix after installation on an old pre-T2 Mac fixes
  the issue


dhh:
Thanks for tracking this down — the diagnosis is right, and splitting the quirk into its own leaf with a single owner is the correct shape. I've pushed two commits on top of yours rather than leave a pile of review notes. Here's what changed and why.

### The gate

`lspci -nn | grep -qiE 'network controller.*\[14e4:'` had two problems.

The first is `grep -q`, which is the SIGPIPE bug from #6608: grep exits at the first match, lspci keeps writing, and under `pipefail` the pipeline reports 141 — read as "no such hardware" on exactly the machines being targeted. b85ae70e removed that pattern repo-wide five days ago, including from the neighbouring `fix-t2.sh`.

The second is that it made T2 Macs depend on a detection line they never needed. They carry a T2 PCI ID that is always present, whereas the class-name half of `lspci -nn` output comes from the pci.ids database.

Naming the other machines by DMI model turned out not to work either, because the model year doesn't predict the part. `MacBookPro11,4` (Mid 2015) carries a BCM43602 and needs this; `MacBookAir7,2` (Early 2015) carries a BCM4360 and doesn't. Covering the lineup by name takes about twenty identifiers across four product lines.

So the gate is now the T2 PCI ID, or an Apple machine carrying one of the PCI IDs `brcmfmac` binds, taken from the driver's own `brcm_hw_ids.h`:

```bash
if lspci -nn | grep "106b:180[12]" >/dev/null ||
  { [[ $sys_vendor == Apple* ]] &&
    lspci -nn | grep -E "14e4:(43ba|43bb|43bc|43a3|43dc|4464|4488|4425|4433)" >/dev/null; }; then
```

That picks up two sets of machines the original missed: the 2016/2017 MacBook Pros (`MacBookPro13,x`/`14,x`, also BCM43602), and `iMac19,1`/`iMac19,2` — which are BCM4364 and, per the [t2linux Wi-Fi guide](https://wiki.t2linux.org/guides/wifi-bluetooth/), not T2 Macs at all. It also excludes the BCM4360 Macs for free: `43a0` isn't a `brcmfmac` ID, because those run the out-of-tree `wl` driver and would never read a `brcmfmac` option.

### Existing installs

The quirk only ran at install time, so it never reached a machine already running Omarchy — which is the situation you described repairing by hand. `migrations/1786391100.sh` applies the same gate to existing installs. It appends rather than overwrites so anything else in the file survives, treats only an active `options` line as already-applied (a commented-out one means the machine is still on the broken default), and sets `reboot-required` rather than reloading `brcmfmac`, which would drop a Wi-Fi connection that currently works — possibly the one carrying the update.

### One thing worth confirming

The chip-ID list is derived, not tested. Your reproduction is on a MacBookPro11,4; if you or anyone on the thread has an `iMac19,x` or a 2017 MacBook Pro, confirmation there would be welcome.

PR #6657, opened 2026-08-09
Codex collector: strict provider attribution and thread-settings models

dalmasluca:
Brings the Codex scanner fixes upstreamed in the `model-usage` plugin (PR #6478) into the new `omarchy.agents` collector design.

### Provider attribution is strict
Pi/omp sessions were charged to Codex when the `api` field merely started with `openai-codex`. A third-party router serving the same models behind an `openai-codex-responses` api was therefore added to this subscription's buckets. Only sessions whose `provider` is exactly `openai-codex` now count, matching the Claude collector fix (#6655).

**Test:** `test/shell.d/agent-usage-codex-scanner-test.sh` gains a third-party-router session using the `openai-codex-responses` api, which must not land in the Codex record.

**Models from thread settings:** sessions without `turn_context` record their model in `thread_settings_applied` (and may log token counts before any model event). A pre-pass seeds the first model-bearing entry, and `thread_settings_applied` updates are followed during the scan, so those turns land on the real model instead of a generic `codex` bucket (see commit `9394794b`, `acc72519` of #6478).

**Regression test:** a session whose model only appears in `thread_settings_applied`, with the first token count preceding it, is attributed correctly and leaves no generic bucket.

### DIFF
--- a/bin/omarchy-agent-usage-codex
+++ b/bin/omarchy-agent-usage-codex
@@ -254,6 +254,26 @@ def scan_native_codex_sessions():
   for path in files:
     current_model = "codex"
     try:
+      # A session may log token counts before its first model event. Read the
+      # first model the file records so those early counts land on the real
+      # model instead of a generic "codex" bucket.
+      with path.open(errors="replace") as handle:
+        for raw in handle:
+          try:
+            entry = json.loads(raw)
+          except Exception:
+            continue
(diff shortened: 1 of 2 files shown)


dalmasluca:
Chiuso a favore del PR #6478 riallineato: il lavoro è stato ripushato sul branch codex-model-attribution con il nuovo design del collector agents/.

PR #6664, opened 2026-08-09
Fix fingerprint setup script to detect non-libfprint-git providers

abbaty48:
## Problem

`omarchy-setup-security-fingerprint` always attempts to install `libfprint-git`
regardless of whether a different `libfprint` provider is already installed
(e.g. `libfprint-goodix53x5`, or similar AUR packages for other sensor
families not supported by mainline `libfprint`).

The current check only looks for a package literally named `libfprint`:

​```bash
installed_libfprint=$(pacman -Qq libfprint 2>/dev/null || true)
if [[ $installed_libfprint == "libfprint" ]]; then
  sudo pacman -Rdd --noconfirm libfprint
fi
omarchy-pkg-add libfprint-git fprintd usbutils
​```

This misses any package that *provides* `libfprint` under a different name.
When such a provider is installed, `omarchy-pkg-add libfprint-git` hits an
unresolvable pacman conflict (both packages provide `libfprint`), and the
wizard aborts with:

​```
:: libfprint-git-... and libfprint-goodix53x5-... are in conflict (libfprint).
​```

This affects any user whose sensor isn't supported by mainline `libfprint`
and who installed a driver-specific AUR package instead (common for several
Goodix, Synaptics, and Elan sensor variants on Dell/Lenovo laptops).

## Fix

Instead of checking for the literal `libfprint` package name, check who
actually owns `/usr/lib/libfprint-2.so.2` on disk via `pacman -Qoq`. If a
non-`libfprint-git` provider already owns it, skip installing `libfprint-git`
entirely and just ensure `fprintd`/`usbutils` are present.

## Testing

Reproduced and fixed on a Dell XPS 13 9370 with a Goodix 27c6:5385 sensor
(HTK32), using the `libfprint-goodix53x5` AUR package. Before the fix, the
wizard failed every time with the conflict above. After the fix, it detects
the existing provider, skips `libfprint-git`, and completes enrollment
successfully.


dhh:
This needs a rebase. Full of unrelated changes.

abbaty48:
> This needs a rebase. Full of unrelated changes.

yes it's, please consider it

PR #6679, opened 2026-08-10
Add QR code capture

Wondertan:
## Summary

- add `omarchy capture qr` to decode a selected screen region
- copy the decoded value to the clipboard and show it in a notification
- add QR Code to the Capture menu
- install `zbar` for fresh and existing systems

## Motivation

QR codes often appear on the same computer where their contents are needed—for example, when importing an OTP/2FA setup code shown by a web app. Reading one currently means reaching for a phone or taking a screenshot and manually passing it through a decoder. This feature turns that whole detour into one Capture menu action: select the QR code, then its decoded value is immediately displayed and copied to the clipboard.

## Testing

- generated a QR code with `qrencode` and decoded it through the new command using the real `zbarimg`
- manually tested the Capture → QR Code menu flow

### DIFF
--- a/bin/omarchy-capture-qr
+++ b/bin/omarchy-capture-qr
@@ -4,16 +4,32 @@
 # omarchy:group=capture
 # omarchy:examples=omarchy capture qr
 
+# Keep hyprpicker alive until after grim captures so the screenshot sees the
+# frozen overlay rather than live content shifting during teardown.
+cleanup_freeze() {
+  [[ -n $PID ]] && kill $PID 2>/dev/null
+}
+trap cleanup_freeze EXIT
+
+hyprpicker -r -z >/dev/null 2>&1 &
+PID=$!
+sleep .1
 SELECTION=$(slurp 2>/dev/null)
+
 [[ -z $SELECTION ]] && exit 0
 
-RESULT=$(grim -g "$SELECTION" - | zbarimg -q --raw - 2>/dev/null) || {
+# Decode QR codes only. Leaving the other symbologies enabled lets dense screen
+# content false-positive as an EAN or Code 39 barcode and take over the clipboard.
+RESULT=$(grim -g "$SELECTION" - | zbarimg -q --raw -Sdisable -Sqrcode.enable - 2>/dev/null)
+
+if [[ -z $RESULT ]]; then
   omarchy-notification-send -g 󰐲 -u critical "No QR code found" "Select a region containing a QR code"
   exit 1
-}
+fi
 
-[[ -z $RESULT ]] && exit 1
-
-printf '%s\n' "$RESULT"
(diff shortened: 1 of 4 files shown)


dhh:
Nice feature — the Capture menu is the right home for this, and the metadata, migration, and package addition all follow the existing conventions. I've pushed two commits to the branch rather than leaving a review to round-trip on. Here's the reasoning for each.

### Keep the decoded value on the clipboard only

The motivation in the description is the strongest argument for this: an OTP setup code means the typical payload is `otpauth://totp/...?secret=...`. The original flow put that value in three places that outlive the copy:

- `printf '%s\n' "$RESULT"` — when the command runs from the menu under uwsm, stdout is the session journal, so the secret gets written to disk.
- `"Result: $RESULT"` in the notification — rendered on screen and kept in notification history. (It also passes through `printf '%b'` inside `omarchy-notification-send`, so a payload containing backslashes renders mangled.)
- a plain `wl-copy` — clipboard history records every selection, so the secret survives there too.

So the value now goes to the clipboard and nowhere else, and the copy is marked `wl-copy --sensitive`. That offers the `x-kde-passwordManagerHint` type, which the clipboard capture already gates on, so the entry is skipped by history. Pasting is unaffected; only the recorded copy is given up. This also drops the unbounded notification body — a QR can carry a few thousand characters.

This follows `omarchy-capture-text-extraction`, which likewise reports only "Copied text from selection to clipboard".

### Decode QR codes only

`zbarimg` has every symbology enabled by default, so it isn't really a QR reader. A Code 128 barcode on screen decodes and takes the clipboard:

```
$ zint -b 20 -d "NOT-A-QR-CODE" -o code128.png
$ zbarimg -q --raw - < code128.png                          # as written
NOT-A-QR-CODE
$ zbarimg -q --raw -Sdisable -Sqrcode.enable - < code128.png # restricted
(no output, exit 4)
```

EAN/UPC in particular can false-positive on dense screen content. `-Sdisable -Sqrcode.enable` makes the behaviour match what the command and the menu entry claim to do, and makes the "No QR code found" message accurate.

### Freeze the screen before selecting

Both sibling region captures start `hyprpicker -r -z` with an EXIT trap so `grim` photographs a frozen frame. Without it, content can shift between the selection and the capture. Added for consistency.

### Smaller things folded in

The empty-result case exited `1` silently with no feedback; it now shares the "No QR code found" notification, which also covers the rare `grim` failure — in every one of those cases the user's next move is the same, so one message serves them all.

One note on `zbarimg -`: it works because a bare `-` is passed through as a filename to ImageMagick, which reads stdin. Worth knowing it depends on the ImageMagick backend rather than on stdin support in zbar itself, but `zbar` depends on `imagemagick` and it's already in the base package list, so the extra footprint here is only ~250 KiB.

PR #6682, opened 2026-08-10
Remember Bluetooth on/off through the rfkill soft block

dhh:
Turning Bluetooth off in the panel does not survive a reboot. BlueZ never
persists an adapter's `Powered` property, so the adapter comes back up on the
next boot regardless of what the user chose.

Omarchy's existing answer is `AutoEnable=false` in `/etc/bluetooth/main.conf`,
set at install time. That persists nothing either — it only means "never power
the adapter on", so Bluetooth comes up **off** on every boot no matter what the
user wanted. Neither state is actually remembered.

## The soft block already does this

`systemd-rfkill` saves every rfkill switch under `/var/lib/systemd/rfkill` and
restores it early on the next boot. That is the unit's entire purpose, and it is
how GNOME's Bluetooth switch behaves. Verified on a live machine:

- `rfkill block bluetooth` → BlueZ reports `Powered: no`, `PowerState: off-blocked`
- the state persists to disk (`0` → `1`) once `systemd-rfkill` settles
- `rfkill unblock bluetooth` → bluetoothd powers the adapter back up **by itself**
  with `AutoEnable` at its stock default
- `/dev/rfkill` is `0664 root:rfkill` plus a uaccess ACL, so no `sudo` is needed

So the block becomes the state and BlueZ follows it. `Powered` still tracks the
block, which means the panel switch, the hero label and the bar icon read it
exactly as they did before — no display changes.

## What changes

- **`bin/omarchy-bluetooth-power`** (new) — `on` / `off` / `toggle`, the single
  place that knows how to move the block. `on` lifts it and only falls back to
  `bluetoothctl power on` for an adapter that was powered down without a block.
- **`install/hardware/bluetooth.sh`** — stops writing `AutoEnable=false`.
  Leaving it at the default is what lets bluetoothd act on the block.
- **`shell/plugins/panels/bluetooth/Panel.qml`** — `toggleBluetooth()` runs the
  command instead of assigning `adapter.enabled`.
- **`bin/omarchy-bluetooth-device`** — routes its `power_on` through the command,
  because `bluetoothctl power on` fails outright while a block is set.
- **Migration** — reads `Powered` before anything changes, hands it to the block,
  then reverts `AutoEnable` to its default. Only the exact line Omarchy wrote is
  reverted, so a hand-edited opt-out survives.

## Why this over #6672

#6672 fixes the same bug by keeping AutoEnable off and taking ownership of the
power-on: a state file plus a systemd oneshot that saves on `ExecStop` and
restores on `ExecStart`. It works, but it reimplements persistence the OS already
provides, and it carries the costs that come with owning it — a boot-path unit
with an adapter wait to bound, a state file that has to survive an interrupted
shutdown write, `AutoEnable` rewriting that must be verified to have landed, and
no answer for a second controller, since `bluetoothctl` only addresses the
default one. This drops all of that; the block covers every controller at once.

The trade is one concept for another: `Powered` persisted by us, versus rfkill
persisted by systemd but which every consumer has to know about. That is why
`omarchy-bluetooth-device` changed too, and why the new command exists rather
than callers reaching for `rfkill` directly.

Closes nothing on its own — #6672 can stay open until there's a call on which
approach to take.

### DIFF
--- a/bin/omarchy-bluetooth-device
+++ b/bin/omarchy-bluetooth-device
@@ -20,8 +20,9 @@ address=${2:-}
 
 power_on() {
   [[ $(timeout 2s bluetoothctl show 2>/dev/null) == *"Powered: yes"* ]] && return
-  bluetoothctl power on >/dev/null 2>&1 || true
-  sleep 0.5
+  # Not bluetoothctl directly: Bluetooth is turned off by an rfkill soft block,
+  # and BlueZ refuses to power an adapter up while one is set.
+  omarchy-bluetooth-power on || true
 }
 
 trust_device() {
--- a/bin/omarchy-bluetooth-power
+++ b/bin/omarchy-bluetooth-power
@@ -2,7 +2,7 @@
 
 # omarchy:summary=Turn Bluetooth on or off, remembered across reboots
 # omarchy:group=bluetooth
-# omarchy:args=<on|off|toggle>
+# omarchy:args=<on|off|toggle|is-on>
 
 # BlueZ never persists an adapter's Powered property, so turning Bluetooth off
 # through bluetoothctl lasts only until the next boot. The rfkill soft block does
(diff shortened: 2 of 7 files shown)


dhh:
Ran a Codex review (gpt-5.6-sol, xhigh) over the branch. Four P2 findings, all valid, all fixed in `b508a51c`, `8090f21e`, `41bec6fe`.

**Migration aborted outside a graphical seat.** `/dev/rfkill` is only writable unelevated from an active seat, so an update run over SSH failed here with `EACCES` — and since migrations run under `bash -e`, that aborted before the config revert and the marker, then aborted again on every retry. Now goes through `sudo`, which is what the privilege guidance calls for on machine-wide work from a visible terminal.

**The read didn't span controllers the way the block does.** This was the sharper one, because covering every controller at once is the main thing this approach claims over #6672 — and the write side did, while the read side used a bare `bluetoothctl show`, which reports the default controller only. A powered dongle sitting behind a powered-down internal controller read as off and got blocked with it. Controllers are now enumerated and any powered one counts as on, exposed as `is-on` so the migration and `toggle` share one read.

**The wait counted probes, not time.** With D-Bus wedged, each probe burns its own 2s timeout, so the intended ~2s bound became ~22s per wait and ~49s across a full power-up. One deadline around the whole wait; measured worst case is now 9s.

**Detached toggle could race itself.** The helper runs detached and the switch only moves once BlueZ catches up, so a second click inside that window re-read the pre-click state and undid the first. The panel knows which way it wants to go, so it now asks for `on` or `off` rather than `toggle`.

Re-verified on hardware afterwards: `off` 0.04s, `on` 0.23s, `is-on` correct in both states, and a two-toggle round trip through the panel's IPC lands back where it started with the rfkill state persisted. Full suite green.

PR #6688, opened 2026-08-10
Resolve omarchy_monitor_scale variable reference in clamshell recovery

s-gato:
## Problem

When a user customizes `monitors.lua` with a specific internal output (e.g. `output = "eDP-1"`) while keeping the default template's variable indirection for scale, disconnecting an external monitor resets the internal panel scale to the default (2) instead of the configured value.

Example config (matches the pattern the default template ships, just with a specific output):

```lua
local omarchy_gdk_scale = 1.5
local omarchy_monitor_scale = 1.5

hl.env("GDK_SCALE", tostring(omarchy_gdk_scale))
hl.monitor({ output = "eDP-1", mode = "preferred", position = "auto", scale = omarchy_monitor_scale })
```

On USB-C / external monitor unplug, `omarchy-hyprland-monitor-watch` fires `omarchy-hyprland-monitor-clamshell`, which calls `configured_monitor_scale()` to read the desired scale.

## Root cause

`configured_internal_monitor_value scale` greps the `hl.monitor` line and captures whatever follows `scale =`. When the rule references the variable, the regex captures the literal string `omarchy_monitor_scale` — **non-empty but not a number**.

`configured_monitor_scale()` only falls through to the `local omarchy_monitor_scale = ...` lookup when the capture is empty (`[[ -z $scale ]]`), so the variable name leaks through, `valid_scale` rejects it downstream, and `read_monitor_scale()` returns the hardcoded default `2`.

With the default template (`output = ""`) the regex never matches, so the fallback fires correctly — which is why this only bites users who pin the output name (a pattern `hyprland.md` itself recommends).

## Fix

Fall through on any non-numeric capture, not just an empty one:

```diff
-  if [[ -z $scale ]]; then
+  if ! valid_scale "$scale"; then
```

## Reproduction

```bash
INTERNAL=eDP-1
monitor_lua=./monitors.lua  # the config above
scale=$(sed -nE "/^hl\.monitor\(\{.*output = \"$INTERNAL\".*\}\)/s/.*scale = \"?([^\", }]+)\"?.*/\1/p" "$monitor_lua" | head -1)
echo "captured: $scale"   # → omarchy_monitor_scale (invalid)
```

## Test coverage

Adds a regression case to `test/shell.d/monitor-clamshell-scale-test.sh` covering the specific-output + variable-reference combination. Verified it fails on the previous code and passes with the fix. Full existing suite still passes (the only failing suites on my machine are environment-dependent: `config-test` needs an `omarchy-pkgs` checkout, `runtime-smoke` needs a live test shell — both fail identically on pristine `quattro`).

### DIFF
--- a/bin/omarchy-hyprland-monitor-clamshell
+++ b/bin/omarchy-hyprland-monitor-clamshell
@@ -7,6 +7,7 @@ TOGGLES_DIR="$HOME/.local/state/omarchy/toggles/hypr"
 CLAMSHELL_FLAG="$TOGGLES_DIR/internal-monitor-clamshell.lua"
 MANUAL_DISABLE_FLAG="$TOGGLES_DIR/internal-monitor-disable.lua"
 SCALE_STATE="$TOGGLES_DIR/internal-monitor-scale"
+MONITOR_LUA="$HOME/.config/hypr/monitors.lua"
 
 INTERNAL=$(omarchy-hyprland-monitor-laptop)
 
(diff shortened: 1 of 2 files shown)


dhh:
Thanks for tracking this down — the diagnosis is exactly right. `configured_internal_monitor_value` captures the literal text `omarchy_monitor_scale`, the `[[ -z $scale ]]` guard lets it through, and the panel ends up on the hardcoded `2`.

I've pushed a follow-up commit rather than asking for changes, since we're close to RC and I wanted the fix to land as one safe unit.

The reason for the rework: keying the fallthrough on "the scale is not a number" is wider than the bug. It also catches values that are *deliberately* non-numeric, so two configs regress against the released script:

| Config | Released | This PR alone |
|---|---|---|
| pinned rule with `scale = omarchy_monitor_scale` | `2` | `1.5` ✅ |
| pinned rule with explicit `scale = "auto"`, plus an unrelated `local omarchy_monitor_scale = 2` | `1.75` | `2` ❌ |
| pinned rule naming a variable that doesn't exist | `1.75` | `1` ❌ |

Row two is the case the "ignores transient auto scale" test was written to prevent, just reached through a pinned output instead of the catch-all. Row three reaches the `output = ""` rule, which Hyprland never applies to a monitor that has its own rule, so the internal panel picks up the external monitors' scale.

The follow-up narrows the trigger: a *bare word* resolves against a `local` only when one actually exists, so a quoted `"auto"` stays a string and an unresolvable name stays unresolved. That also let the surrounding parse tighten up — values are read as the whole of what sits between the `=` and the next separator (so `3 / 2` no longer truncates to `3` and get applied), comments are cut before anything is matched, and `position` gets the same resolution since it had the identical bug.

Behavior against the released script is now down to two configs, both intentional: the bug you reported, and a quoted number in the local (`= "1.5"`), which previously fell back and now applies.

Your commit is untouched underneath.

PR #6693, opened 2026-08-10
Preserve borders on Google Meet browser windows

timohubois:
## Problem

The Google Meet PiP rule matches every Chromium-based window whose title starts with `Meet - `. Normal Meet tabs use that title too, so their dynamic border becomes zero. When the Meet title is already available as the window maps, the static PiP effects also float, pin, resize, and move the normal browser window.

## Fix

Keep matching the Meet title, but exclude initial titles ending in the browser-brand suffixes used by Omarchy's supported Chromium browsers. Standalone web apps remain excluded by their application class, while legacy Meet PiP windows without a browser suffix continue to match.

### Border detail

Before and after border detail

<details>
<summary>Full-window comparison</summary>

Before and after full-window comparison

</details>

## Testing

Tested on Hyprland 0.56.2 with Chromium 151:

- Normal Meet window with its title available at map: tiled, unpinned, 2px border
- Normal Meet window after a title transition: tiled, unpinned, 2px border
- Standalone Meet web app/PWA: tiled, unpinned, 2px border
- Standard Chromium `Picture in picture` window: floating, pinned, 600x338, borderless
- Legacy `Meet - ...` PiP title: floating, pinned, 600x338, borderless
- Browser-title matrix for Chromium, Google Chrome, Brave, Microsoft Edge, Vivaldi, and Helium: all remained tiled with 2px borders
- `git diff --check`

### DIFF
--- a/default/hypr/apps/pip.lua
+++ b/default/hypr/apps/pip.lua
@@ -11,8 +11,12 @@ o.window({ tag = "pip" }, {
   move = { "(monitor_w-window_w-40)", "(monitor_h*0.04)" },
 })
 
--- Google Meet PiP uses the meeting title instead of "Picture-in-Picture".
-o.window({ tag = "chromium-based-browser", title = "^Meet - .+" }, {
+-- Google Meet PiP titles omit the browser suffix used by regular Meet tabs.
+o.window({
+  tag = "chromium-based-browser",
+  title = "^Meet - .+",
+  initial_title = "negative:.* - (Chromium|Google Chrome|Brave|Microsoft Edge|Vivaldi|Helium)$",
+}, {
   tag = "-default-opacity",
   float = true,
   pin = true,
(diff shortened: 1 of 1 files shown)


rmacy:
Confirmed this exact failure mode on Omarchy 4.0.1-1 with Hyprland 0.56.2.

A normal Chromium window mapped with an initial title shaped like `Meet - <meeting-code> - Chromium`, then navigated to unrelated pages. `hyprctl clients -j` showed that the window remained floating and pinned. Other floating windows could not be raised above it by clicking them. Manually unpinning that Chromium window immediately restored normal focus/stacking behavior.

I also confirmed that a late local rule setting `pin = false` only for browser-suffixed Meet initial titles keeps the affected window unpinned after a Hyprland reload. The `initial_title` exclusion in this PR matches the observed title shape and fixes the false positive without broadly disabling Meet PiP pinning.

Thanks for tracking this down.

PR #6709, opened 2026-08-11
Enable WiFi 7 on Intel BE211 (quattro port of #6450)

dhh:
Quattro port of #6450, which targets `dev` and cannot be retargeted as-is.

## Why a separate PR

#6450 is branched from `dev`, where this installer lives at
`install/config/hardware/intel/fix-wifi7-eht.sh`. On `quattro` it is
`install/hardware/intel/fix-wifi7-eht.sh`, and that is the path
`install/hardware/all.sh` actually sources. Retargeting #6450 to `quattro`
produces a PR that adds a file at a path quattro never runs, leaves the real
installer still matching `[8086:(e440|272b)]`, and drags in an unrelated
`ssh-port-forwarding` change from `dev`. Tested — it does not work.

## Changes

- Stop disabling EHT on BE211 (`8086:e440`).
- Keep the workaround for BE200 (`8086:272b`), where kernel bug 221675 still
  tracks a 320 MHz firmware assertion after warm reconnects.
- Add a migration removing the historical config from existing BE211 installs,
  gated on every installed kernel being Linux 7.1 or newer.

Keeps quattro `mkdir -p` + `cat >` rather than switching to `sudo tee`;
`omarchy-apply-hardware` already requires `EUID 0`, so the change stays limited
to the detection logic.

## Two fixes on top of #6450

1. **No double initramfs rebuild.** Eligibility is decided before the
   interrupted-recovery block, so an eligible machine with an orphaned backup
   finishes in one rebuild instead of restoring the config, rebuilding, then
   removing it and rebuilding again. Ineligible machines still restore.
2. **`config + backup both present` is no longer a dead end.** A stray backup
   byte-identical to the workaround being removed is discarded and the
   migration proceeds; anything else still hard-fails rather than destroying
   unexpected state.

## Validation

- `bash -n` clean on both files, mode 644, `git diff --check` clean.
- `./test/all`: one failure, `update-lock-test.sh`, confirmed pre-existing on
  clean `quattro` by stashing and re-running. ShellCheck not run (not installed
  on the test machine).
- Gates verified against a live BE211 (`8086:e440`, no BE200, kernels
  `linux-ptl 7.1.5.arch1-1` and `linux 7.1.7.arch1-1`): config matches the
  expected content byte-for-byte, so the migration takes the main path.

Migration not yet executed on hardware; that needs a reboot to confirm EHT
negotiates.

https://claude.ai/code/session_016uRU5yLNZLSBF2kKvyVad1

### DIFF
--- a/install/hardware/intel/fix-wifi7-eht.sh
+++ b/install/hardware/intel/fix-wifi7-eht.sh
@@ -1,16 +1,13 @@
-# Temporary fix for Dell XPS 14/16 on Panther Lake
-# Disable WiFi 7 (EHT/802.11be) on Intel BE200/BE211 cards
-# The iwlwifi driver has a broken EHT RX data path — APs drop to MCS 0 NSS 1
-# when EHT is negotiated, making WiFi unusable. Disabling EHT falls
-# back to WiFi 6 (HE/802.11ax) which works at full speed.
-# This should be removed when Intel fixes the firmware/driver.
+# Temporarily disable WiFi 7 (EHT/802.11be) on Intel BE200 cards.
+# Linux kernel bug 221675 tracks a firmware assertion after warm reconnects
+# at 6 GHz/320 MHz. Disabling EHT avoids the failing channel context.
+# https://bugzilla.kernel.org/show_bug.cgi?id=221675
 
-if lspci -nn | grep -qE '\[8086:(e440|272b)\]'; then
+if lspci -nn | grep -q '\[8086:272b\]'; then
   mkdir -p /etc/modprobe.d
   cat > /etc/modprobe.d/iwlwifi-disable-eht.conf <<'EOF'
-# Temporary fix Dell XPS 14/16 on Panther lake
-# Disable WiFi 7 (EHT) on Intel BE200/BE211 — broken RX rate adaptation
-# Remove this file when fixes land in the iwlwifi EHT data path
+# Temporary workaround for Intel BE200 320 MHz firmware assertions
+# Remove this file when Linux kernel bug 221675 is fixed
 options iwlwifi disable_11be=Y
 EOF
 fi
--- a/migrations/1784508420.sh
+++ b/migrations/1784508420.sh
@@ -1,0 +1,184 @@
+echo "Enable WiFi 7 on Intel BE211 with Linux 7.1 or newer"
+
+wifi7_config_matches_omarchy_1784508420() {
+  local config_file=$1
+
+  printf '%s\n' \
+    '# Temporary fix Dell XPS 14/16 on Panther lake' \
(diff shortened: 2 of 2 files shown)


dhh:
Closing this — measured A/B on a second BE211 shows EHT is a **download regression**, so the change looks premature for `8086:e440`.

## Test setup

Dell XPS 14, BE211 `8086:e440`, Linux `7.1.5-arch1-1-ptl`. UniFi U7 Pro, 6 GHz ch69 @160 MHz, ETSI/Denmark. Same AP, same position, −38 dBm, 8 parallel TCP streams to Cloudflare. WAN verified symmetric at 2305/2373 Mbit/s from the gateway, so it is not the bottleneck.

## Results

| | HE (`disable_11be=Y`) | EHT (this PR) |
|---|---|---|
| **Download** | **1212 Mbit/s** | **918 / 979 Mbit/s** |
| **Upload** | 1506 Mbit/s | 1678 Mbit/s |
| RX PHY | 1441.3 (MCS 7, NSS 2) | ~1441, unstable |
| TX PHY | 2401.9 (MCS 11, NSS 2) | 2593.5 (MCS 12, NSS 2) |

Download down ~20–24%. Upload up ~11%.

## Mechanism

RX cannot hold two spatial streams under EHT. Sampling `iw station dump` once per 1.2 s during a sustained 6-stream download, RX oscillates:

```
EHT-MCS 13 EHT-NSS 1   1441.1 MBit/s
EHT-MCS  7 EHT-NSS 2   1441.3 MBit/s
```

10 samples at NSS 2, 4 at NSS 1. The PHY rate lands near 1441 either way — the driver compensates for losing a stream by jumping to 4096-QAM — but the stream count keeps flapping, and that instability costs real throughput.

This is a milder form of the signature the original workaround was written for: *"APs drop to MCS 0 NSS 1 when EHT is negotiated."* Linux 7.1 clearly improved it (MCS 13 rather than MCS 0), but the EHT RX data path still cannot sustain NSS 2 on this combination.

Worth noting the PHY figures in #6450 (3.46 Gbps RX / 3.84 Gbps TX at 320 MHz) do not translate into throughput here: measured download in that PR was ~718 Mbps, below what this machine already does at 1212 on HE.

## Scope

This is exactly the broader coverage #6450 flagged under "Remaining risk" — one BE211/firmware/AP combination validated, more needed. Second combination, and it regresses. Note also that in ETSI the 6 GHz band (5945–6425) fits only one 320 MHz channel, so EU installs get EHT at 160 MHz, not the 320 MHz the PR validated at.

Not arguing the BE200 exclusion is wrong, and the migration itself worked correctly — gates matched, rollback path unused, boot images rebuilt cleanly. The hardware policy just is not ready.

Reverted locally. Happy to retest on a future firmware or kernel.

dhh:
Closing in favour of keeping the BE211 workaround until EHT RX holds NSS 2. See measurements above.

PR #6712, opened 2026-08-11
Detect NVIDIA GPUs without waking them

### LINKED ISSUE #6660: Quattro: hyprctl reload fails with a Lua timeout error
### System details

Omarchy 4.0.0.r1602.g9d61915-1 | Hyprland 0.56.2 | Kernel 7.1.6-arch1-1 AMD Ryzen 5 5600H | NVIDIA GA106M [RTX 3060 Mobile / Max-Q] + AMD Cezanne

### What's wrong?

# `hyprctl reload` fails with a Lua timeout error

## Error

Editing and saving a file in `~/.config/hypr/`, or running `hyprctl reload`, produces this
error above the bar:

```
require("default.hypr.apps.1password"): /usr/share/omarchy/default/hypr/apps/1password.lua:1: [Lua] execution timed out in config reload
```

## Possible lead

An agent ran tests on this machine and points at this line in
`/usr/share/omarchy/default/hypr/nvidia.lua`:

```lua
if o.shell_succeeds("lspci | grep -qi nvidia") then
```

On a hybrid-GPU laptop the discrete NVIDIA GPU is runtime-suspended while idle. `lspci`
reads PCI config space, which wakes it, and that wake alone takes long enough to exceed
the config load timeout. Config loads that happen while the GPU is already awake do not
produce the error.

## System

```
Omarchy 4.0.0.r1602.g9d61915-1 | Hyprland 0.56.2 | Kernel 7.1.6-arch1-1
AMD Ryzen 5 5600H | NVIDIA GA106M [RTX 3060 Mobile / Max-Q] + AMD Cezanne
```

dhh:
Fixes #6660, where `hyprctl reload` on a hybrid-GPU laptop fails with:

```
require("default.hypr.apps.1password"): /usr/share/omarchy/default/hypr/apps/1password.lua:1: [Lua] execution timed out in config reload
```

### What's happening

`lspci` reads each device's PCI config space, and the kernel resumes a runtime-suspended device to serve that read. On a hybrid laptop the discrete NVIDIA GPU idles in D3cold, so the first `lspci` of a config load spends over a second waking it.

Hyprland budgets **1500 ms for the entire Lua config load** (`LUA_TIMEOUT_CONFIG_RELOAD_MS`), enforced by a watchdog that fires every 10,000 VM instructions. That watchdog can't interrupt a blocking `os.execute`, so it trips on the next Lua line to run after `lspci` returns. `default/hypr/envs.lua` requires `nvidia.lua` just before `windows.lua` requires the apps directory — which is why the error blamed `apps/1password.lua:1`, the alphabetically first app, instead of the code that actually spent the time.

On a machine that has an NVIDIA GPU we ran `lspci` up to three times: once inline, then again inside each of the GSP detectors.

### The fix

Read `vendor`, `class`, and `device` from `/sys/bus/pci/devices/*` instead. Those attributes are served from cached `pci_dev` fields and never touch config space, so nothing gets woken. It's also faster in absolute terms — no `lspci`, no pipes, no `grep`.

Generation is now classified by device ID rather than by grepping `lspci`'s marketing names. Two boundaries, both exact against `pci.ids` restricted to display-class parts:

| generation | GPU device IDs | detector |
|---|---|---|
| Fermi / Kepler | `06c0`–`130b` | neither — no driver we package |
| Maxwell / Pascal / Volta | `1340`–`1df6` | `omarchy-hw-nvidia-without-gsp` |
| Turing / Ampere / Ada / Blackwell | `1e02`–`3340` | `omarchy-hw-nvidia-gsp` |

The older detector is bounded at both ends on purpose: a match installs `nvidia-580xx*`, which supports exactly that span. An unbounded "below Turing" test would have claimed a Kepler GTX 780 and installed a driver that can't drive it, where today the install bails with "No compatible driver for your NVIDIA GPU".

Going by ID also closes gaps in the old regexes, which only matched a hand-picked set of product names — a Maxwell `GeForce 830M` or a Pascal `GeForce 940MX` matched nothing and got no driver at all. And unlike the names, the ranges don't need a new alternation for every future product family.

`bin/omarchy-hw-nvidia` was also checked in without its executable bit, so it failed with "Permission denied" on any direct call. It needs the bit now that `nvidia.lua` invokes it.

### Note

I don't have hybrid NVIDIA hardware here, so the timeout itself isn't reproduced locally — the mechanism is established from the Hyprland source and the kernel's PCI sysfs semantics. @MirasMustimov, if you can run this branch it would be good to confirm the reload is clean on your RTX 3060.

### DIFF
--- a/bin/omarchy-hw-nvidia
+++ b/bin/omarchy-hw-nvidia
@@ -2,4 +2,15 @@
 
 # omarchy:summary=Detect whether the computer has an NVIDIA GPU.
 
-lspci | grep -qi 'nvidia' &>/dev/null
+# Read the cached sysfs IDs rather than lspci, which reads PCI config space and
+# resumes runtime-suspended GPUs.
+pci_devices_path="${OMARCHY_PCI_DEVICES_PATH:-/sys/bus/pci/devices}"
+
+shopt -s nullglob
+
+for device in "$pci_devices_path"/*; do
+  [[ $(< "$device/vendor") == "0x10de" ]] || continue
+  [[ $(< "$device/class") == 0x03* ]] && exit 0
+done
+
+exit 1
--- a/bin/omarchy-hw-nvidia-gsp
+++ b/bin/omarchy-hw-nvidia-gsp
@@ -2,5 +2,19 @@
 
 # omarchy:summary=Detect whether the computer has an NVIDIA GPU with GSP firmware (Turing or newer).
 
-# GTX 16xx, RTX 20xx-50xx, RTX Pro, Quadro RTX, datacenter A/H/T/L series.
-lspci | grep -i 'nvidia' | grep -qE "GTX 16[0-9]{2}|RTX [2-5][0-9]{3}|RTX PRO [0-9]{4}|Quadro RTX|RTX A[0-9]{4}|A[1-9][0-9]{2}|H[1-9][0-9]{2}|T4|L[0-9]+"
+# Turing is the first generation with GSP firmware, and the first to use device
+# IDs at 0x1e00 or above; Maxwell, Pascal, and Volta all sit below that line.
+#
+# Read the cached sysfs IDs rather than lspci, which reads PCI config space and
+# resumes runtime-suspended GPUs.
+pci_devices_path="${OMARCHY_PCI_DEVICES_PATH:-/sys/bus/pci/devices}"
+
+shopt -s nullglob
+
+for device in "$pci_devices_path"/*; do
+  [[ $(< "$device/vendor") == "0x10de" ]] || continue
+  [[ $(< "$device/class") == 0x03* ]] || continue
+  (( $(< "$device/device") >= 0x1e00 )) && exit 0
+done
+
+exit 1
--- a/bin/omarchy-hw-nvidia-without-gsp
+++ b/bin/omarchy-hw-nvidia-without-gsp
@@ -2,5 +2,22 @@
 
 # omarchy:summary=Detect whether the computer has an NVIDIA GPU without GSP firmware (Maxwell/Pascal/Volta).
 
-# GTX 9xx/10xx, GT 10xx, Quadro P/M/GV, MX series, Titan X/Xp/V, Tesla V100.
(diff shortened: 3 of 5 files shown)


MirasMustimov:
@dhh sure I will run tests on my machine

MirasMustimov:
@dhh 

Checked out the branch with `omarchy dev link`. 
Ran three `hyprctl reload`s spaced ~22s apart so the GPU is asleep each time:

```
before (r1602):   reload 1: ERROR   reload 2: ERROR   reload 3: ERROR
after  (this PR): reload 1: ok      reload 2: ok      reload 3: ok
```

Works for me. Thanks for the quick resolution!

neheb:
Fascinating. I asked an LLM about this.

```
Siblings that share the exact same wake-on-lspci problem:
- omarchy-hw-hybrid-gpu (bin/omarchy-hw-hybrid-gpu:8) — the strongest candidate. Its fallback lspci | grep -cE 'VGA|3D|Display' >= 2 runs on every menu summon via the when: clause in omarchy-menu.jsonc:71, so each menu open wakes the dGPU on hybrid laptops. Trivially convertible: count sysfs class values matching 0x03*, same OMARCHY_PCI_DEVICES_PATH pattern as the commit.
- omarchy-hw-intel-sof (bin/omarchy-hw-intel-sof:5) — mechanical: vendor 0x8086 + class 0x04* covers both Multimedia audio controller (0x0401) and Audio device (0x0403). But it's only called from install scripts, so the wake is a one-time install cost.
- omarchy-install-gaming-gpu-lib32 (bin/omarchy-install-gaming-gpu-lib32:19) — same story: vendor-name grep, install-time only.
The awkward one is omarchy-hw-intel-ptl (bin/omarchy-hw-intel-ptl:5): "Panther Lake" is a name only lspci's database knows — sysfs has no equivalent, so converting means maintaining a hardcoded PTL device-ID list. Only install-time callers, so probably not worth it.
The rest of the lspci users (install/hardware/*, migrations/*.sh, omarchy-upgrade-to-quattro) run once at install/migration time with no budget pressure; the commit's rationale (Hyprland's 1.5s reload budget) doesn't apply to them. One near-miss though: fix-nouveau-cursor.sh uses lspci -k for driver binding, which sysfs also exposes via /sys/bus/pci/devices/*/driver — but again, install-time.
```

That install time angle is interesting. I know there was a lot of work in getting it down to under a minute.

PR #6717, opened 2026-08-11
Repair theme symlinks the state-move migration left dangling

shrijit37:
## Problem

`migrations/1781043107.sh` (Move current theme state to ~/.local/state) re-links legacy theme symlinks whose targets were stored as a literal `"~/.config/omarchy/current/..."` string. The replacement used the same literal tilde:

```bash
ln -sfn "~/.local/state/omarchy/current/$suffix" "$link"
```

The filesystem never expands `~` inside a symlink target — only the shell does, and only in commands. The result is a dangling link for btop, Helix, and VS Code/Cursor (`omarchy-color-theme.json`), and the migration **exits 0**, reporting success while the theme is silently gone.

This is the same bug class already repaired twice for the Neovim link (`1781158082.sh`, then `1785002349.sh` — "Repair Neovim theme symlinks the earlier relink missed"). The btop/helix/vscode/cursor links were never covered by a repair.

## Fix

Two parts, following the repo's existing migration conventions:

1. **`migrations/1781043107.sh`** — relink through the already-defined `$current_state_dir` variable so the shell expands the absolute path, mirroring the sibling branch on line 99.

2. **`migrations/1786451567.sh`** (new) — repairs machines where the buggy migration already ran and marked itself applied (editing it would not help them, same reasoning as `1785002349.sh`). It matches the shared `*/omarchy/current/*` suffix rather than enumerating spellings, is idempotent, and leaves custom links outside `omarchy/current` alone.

## Verification

- Reproduced the bug: literal-`~` legacy target stays dangling after the old migration, exit 0.
- Fixed migration re-links to a resolvable absolute target.
- Repair migration fixes links left by the already-applied buggy version (btop + helix + vscode + cursor), verified end-to-end.
- Ran both migrations on a live system via `omarchy-migrate` — idempotent, marker written, existing correct links untouched.
- `test/cli` full suite passes (107 assertions) including new coverage for the literal-`~` branch (previously untested) and the repair migration.


dhh:
Thanks for this — the core diagnosis is right, and `ln -sfn "~/..."` writing a target the kernel never expands is a real bug. I pushed two commits to the branch (maintainer edits were on) rather than leaving review notes, since one of them is a correctness fix worth getting in directly.

### `Only repair theme symlinks that could never have worked`

The repair matched on `*/omarchy/current/*`, which is filename-agnostic and matches any target merely *containing* that substring. That reaches beyond the links the migration meant to claim. Concretely, a user whose btop theme points into their own dotfiles:

```
~/.config/btop/themes/current.theme -> ~/dotfiles/omarchy/current/theme/btop.theme
```

is a working link, but the migration silently rewrote it to the state directory and their setup was gone. This also contradicted the comment right above it ("leaves custom links outside omarchy/current alone"), and the prior repair the migration cites as precedent, `1785002349.sh`, deliberately qualified its glob with the filename.

My first attempt at a fix was "only relink when the link is dangling, and the target names this exact theme file". That's still not safe: a dotfiles repo on a drive that isn't mounted right now leaves a perfectly legitimate link dangling, and the migration would destroy it permanently on the one boot where the mount was missing.

The discriminator the fix actually uses is narrower and, I think, provably safe: only claim a target that starts with a **literal `~/`**. The filesystem never expands that, so such a link cannot ever have worked — it is broken by construction, and no working setup can possess one. A dotfiles path is a real path whether or not it currently resolves. That's exactly the spelling `1781043107.sh` wrote, so it repairs what was broken and nothing else.

One thing worth flagging, because it changes how much this migration is expected to do: as far as I can tell it should be a no-op almost everywhere. Searching history for a quoted-tilde `ln -s` turns up only `1781043107.sh` itself, and all six of these links were created with an unquoted `~` (which the shell expands) or an absolute path — btop in `ac1ac301`, Helix in `3ce6e980`, VS Code in `357b77c9`, and the current creators in `install/user/theme.sh`, `omarchy-install-editor-helix`, and `omarchy-theme-set-vscode`. So the tilde branch in `1781043107.sh` probably never fired for these paths. That's an argument for keeping the repair narrow rather than for dropping it — it costs nothing when it matches nothing, and the fix to `1781043107.sh` in this PR is correct regardless. Happy to hear if you know of a path that did produce literal-tilde targets here.

Also added regression coverage for the three cases that were unprotected: a working dotfiles link, an unmounted dotfiles link, and a literal-tilde target naming a different theme file.

One note on the existing tests: the "custom links are left alone" case used `../../../../custom.theme`, which climbs four levels out of the fake `HOME` and resolves to `/tmp/custom.theme` — so the `touch` created a file the link never pointed at, and the case was really testing a dangling link. The depth looks copy-pasted from the nvim test, where the link sits one directory deeper and four levels is correct. Fixed to `../../../`.

### `Clean up every scratch directory the CLI suite creates`

Unrelated to the symlink work, hence the separate commit. Five of the eight `mktemp -d` directories in `test/cli` were never registered with the exit trap, so every run left them in `/tmp` — the two new ones from this PR plus three that predate it. They now all go through a small helper that records them for cleanup.

dhh:
Follow-up: I've pushed one more commit that removes the migration tests again — both the ones this PR added and the regression cases I added on top of them. That reverses part of what I described above, so the reasoning is worth stating plainly.

A migration runs once per machine and is inert forever after. A test for it, though, stays in the suite and gets paid for on every run, by everyone, indefinitely. We have a lot of migrations and only one `test/cli`, so that trade goes the wrong way if we keep taking it. The tests did their job here — they're how the clobber bug got pinned down and how the fix was confirmed — but that value was mostly spent during the fix itself.

So the migration keeps the narrowing fix and goes untested. `test/cli` is back to its pre-PR size apart from the `/tmp` cleanup helper, and the older nvim, Hyprland, and state-move migration blocks are untouched — this isn't a rule being applied retroactively to the whole file, just a decision not to grow it further here.

The correctness reasoning in my earlier comment still stands; only the coverage described in its last two paragraphs is gone.

PR #6725, opened 2026-08-11
fix: collapse dropdown toggle on second click

felixzsh:
dropdown,  multi select, and searchable dropdown qml components try to expand again on click when they are already expanded, correct behavior should be to collapse

### DIFF
--- a/shell/Ui/Dropdown.qml
+++ b/shell/Ui/Dropdown.qml
@@ -136,9 +136,14 @@ Item {
       MouseArea {
         anchors.fill: parent
         cursorShape: Qt.PointingHandCursor
+        // CloseOnReleaseOutside closes on release, so by click time the popup
+        // is already closed; the press-time state marks this as a collapse.
+        property bool _wasOpenAtPress: false
+        onPressed: _wasOpenAtPress = popup.opened
         onClicked: {
           trigger.forceActiveFocus()
-          popup.opened ? popup.close() : popup.open()
+          if (popup.opened || _wasOpenAtPress) popup.close()
+          else popup.open()
         }
       }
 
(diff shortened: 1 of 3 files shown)


felixzsh:
> ## Pull request overview
> Copilot reviewed 3 out of 3 changed files in this pull request and generated no new comments.
> 
> Suppressed comments (3)
> **shell/Ui/SearchableDropdown.qml:164**
> 
> * A rapid second click can still reopen the dropdown while its enter transition is running. `Popup.opened` remains false until the transition completes, even though the popup is visible; the release policy then closes it before `onClicked`, leaving both checks false and calling `open()` again. Capture `popup.visible` at press time so an opening or exiting popup is treated as expanded.
> 
> ```
>         onPressed: _wasOpenAtPress = popup.opened
> ```
> 
> **shell/Ui/Dropdown.qml:142**
> 
> * A rapid second click can still reopen the dropdown while its enter transition is running. `Popup.opened` remains false until the transition completes, even though the popup is visible; the release policy then closes it before `onClicked`, leaving both checks false and calling `open()` again. Capture `popup.visible` at press time so an opening or exiting popup is treated as expanded.
> 
> ```
>         onPressed: _wasOpenAtPress = popup.opened
> ```
> 
> **shell/Ui/MultiSelect.qml:330**
> 
> * A rapid second click can still reopen the dropdown while its enter transition is running. `Popup.opened` remains false until the transition completes, even though the popup is visible; the release policy then closes it before `onClicked`, leaving both checks false and calling `open()` again. Capture `popup.visible` at press time so an opening or exiting popup is treated as expanded.
> 
> ```
>         onPressed: _wasOpenAtPress = popup.opened
> ```

I measured it on (Qt 6.11 + Quickshell): popup.opened is true during the enter transition (the Opening state), even with Material's 220ms
 transition. So on a rapid second click, _wasOpenAtPress captures true and the popup collapses correctly the "both checks false -> reopen" scenario
 can't happen here.

PR #6727, opened 2026-08-12
perf(bar): drop keyboard layout poll, gate tray layer, cache workspace ids

### LINKED ISSUE #6726: perf(bar): stop spawning hyprctl every 10s, gate tray layer, cache workspace model
**Problem**: the bar wastes idle CPU and spawns subprocesses on a fixed schedule even when nothing changed:

- `KeyboardLayout` polls `hyprctl -j devices` every 10s **per bar surface**, although it already listens to the `activelayout` raw event, which fires exactly on the transitions that matter. On a 3-monitor setup that's ~18 useless subprocesses/minute. The poll can also die silently (the `if (!queryProc.running)` guard sticks when a query hangs), killing the refresh forever until a restart.
- The tray drawer rows keep `layer.enabled: true` permanently for a 600ms reveal animation — a permanent offscreen buffer while idle.
- `Workspaces` rebuilds the whole workspace row on *every* workspace signal: the Repeater model is a freshly allocated JS array, so any toplevel arriving or focus change tears down and recreates all delegates.

**Solution**:
1. Remove the 10s poll; keep the `activelayout` event path and the 600ms post-cycle refresh.
2. Gate the tray row layer to `revealProgress > 0` (buffer exists only while the reveal animates).
3. Cache the workspace id list in a Qt-free JS module (`WorkspaceModel.js`) and hand back the same array reference while the id set is unchanged, so delegates stay alive.

**Measured** (60s idle, single monitor): `hyprctl -j devices` spawns 4 → 0 per minute; quickshell idle CPU 0.05% → ~0.03%. The spawn elimination is the deterministic win (~240 subprocesses/hour on one monitor, more on multi-monitor), and removing the poll removes its silent-death failure mode.

Tests: node unit test covering the new `WorkspaceModel.sync` cache semantics was added (`test/shell.d/workspace-model-test.sh`); shell restarted and loaded cleanly with the new code.

markbus-ai:
Closes #6726

## Summary
- Remove the KeyboardLayout 10s `hyprctl -j devices` poll: the `activelayout` socket event already fires on exactly the transitions that matter; the poll was ~240 useless subprocess spawns/hour on a single monitor (per-bar on multi-monitor) and could die silently when a query hung, killing refresh until restart.
- Gate the tray drawer rows' `layer.enabled` to `revealProgress > 0`, so the offscreen buffer exists only while the 600ms reveal animation runs.
- Cache the workspace id list in a Qt-free JS module and return the same array reference while the id set is unchanged, so the Repeater keeps its delegates instead of rebuilding on every workspace signal.

## Changes
| File | Change |
|------|--------|
| `shell/plugins/bar/widgets/KeyboardLayout.qml` | Removed the repeating 10s poll timer; event + post-cycle refresh remain |
| `shell/plugins/bar/widgets/Tray.qml` | Tray drawer row layers gated to `revealProgress > 0` |
| `shell/plugins/bar/widgets/Workspaces.qml` | `workspaceIds()` delegates to the cached model |
| `shell/plugins/bar/widgets/WorkspaceModel.js` | New Qt-free module: stable workspace id array cache |
| `test/shell.d/workspace-model-test.sh` | Node test covering the new `sync()` cache semantics |

## Test plan
- [x] `test/shell.d/workspace-model-test.sh` passes (8 assertions)
- [x] Shell restarted with the new code; loads clean (no errors in `omarchy-shell` journal tag)
- [x] Measured 60s idle: `hyprctl -j devices` spawns 4 → 0/min, quickshell CPU ~0.05% → ~0.03%
- [x] Keyboard layout cycle works after the change (event-driven refresh path exercised)
- [x] Tray drawer reveal exercised (layer gating preserves the animated path)


dhh:
Thanks for digging into this — the keyboard widget spawning `hyprctl` on a fixed schedule was a real find, and the strongest version of the argument isn't even in the description: the timer runs regardless of `visible`, so the single-layout install that never sees the widget was still paying for it, on every monitor.

I've pushed two commits to the branch that keep that part and drop the other two. Here's the reasoning on each.

### Workspaces / `WorkspaceModel.js` — reverted

A `Repeater` compares its model by value, so an equal-but-new array was never rebuilding the delegates. Measured on the Qt the shell actually runs (6.11.1), with a model binding that re-evaluates four times returning a fresh identical `[1,2,3,4,5]`:

```
5 delegate creations    # fresh array, identical contents -> no rebuild
23 delegate creations   # control: contents change each time -> rebuilds
```

`sync()` also still allocated a fresh array on every call before comparing it, so not even the allocation was saved. Worth noting too that `model: root.workspaceIds()` only depends on `Hyprland.workspaces.values` — it re-evaluates when the workspace *set* changes, not when a toplevel arrives or focus moves; `focused` and `occupied` are separate per-delegate bindings.

### Tray `layer.enabled` — reverted

The layer is a row a few icons wide, and a layer costs nothing while its contents hold still — it only re-renders when they change. Gating it trades a small idle buffer for an offscreen buffer built and dropped on every hover in and out.

### KeyboardLayout — kept, but the poll is gated rather than deleted

The poll was covering something the events don't. Hyprland hands the seat's `main` flag to whichever keyboard was typed on last and announces nothing when it moves, so on a seat with two keyboards nothing tells the widget that the label now describes the wrong one — `activelayout` fires on layout switches, not on that. Deleting the poll leaves the label wrong indefinitely for anyone with a laptop plus an external keyboard, and the same goes for a query that fails at login: there was nothing left to retry it.

So the poll stays, but only runs when its answer can change:

```qml
running: !root.keyboardName || root.keyboardUnresolved || root.keyboardCount > 1
```

No reading yet, the last reading couldn't name a keyboard, or the seat holds more than one real keyboard. On the single-keyboard machine — which is most of them, whatever their layout count — it stops for good after the first reading, so the spawns you were after are gone in the case that matters. Layout switches there arrive via `activelayout` as before.

Three smaller things came out of the same read:

- **A refresh arriving during a query was dropped.** `if (!queryProc.running)` silently swallowed it, so an `activelayout` landing while a query was in flight left the label stale. It's coalesced now and re-runs when the query lands.
- **A query that never returns held that guard shut forever.** This is the failure mode the description credits the change with fixing, but the guard lives in `refresh()`, which the event path calls too — a wedged `hyprctl` killed both. Removing the poll actually removed the only thing that would ever have retried. There's a 5s watchdog now.
- **`configreloaded` refreshes too.** A `hyprctl reload` that edits `kb_layout` changes the label *and* whether the widget appears at all, and switches no layout, so it emits no `activelayout` of its own.

### On the measurements

`0.05% -> 0.03%` idle CPU is within noise, and "4 spawns/min" doesn't match the 6/min a 10s timer implies — worth being careful with, since the honest win here isn't idle CPU. It's that a widget most people never see stops running a subprocess forever.

`./test/shell` is green (158 files, 1958 assertions) and `qmllint` is clean. Not yet verified in a running session; that's the remaining check before this merges.

PR #6731, opened 2026-08-12
Make notification history the last ten notifications on disk

dhh:
The notification history was a pair of in-memory lists mirrored into `notifications.json`, split into "pending" and "past" by a seen/unseen distinction no surface exposed, capped at 100, deduped by an id that repeats across server generations, and pruned by a 15-minute TTL. Replaying it showed five rows drawn from whichever list happened to hold them.

Every toast already writes a file under `~/.local/state/omarchy/notifications` so it can survive a shell restart. That file is now the history record.

- A popup leaving the screen — expired, dismissed, or clicked — **moves into `notifications/history/`** instead of being deleted, and the directory is trimmed to the newest ten in the same shell job.
- `showHistory` (SUPER + SHIFT + ALT + comma) replays exactly what is in there, plus the toasts still on screen when it was asked for, newest first.
- A notification DND silenced is written straight into the same directory: a toast that never showed is the one worth looking back at.
- A notification superseded on screen through `replaces_id` is deleted rather than archived, so history holds one entry for what the sender means as one notification.
- `notifications clear` now forgets the recorded history instead of emptying a "past" bucket that nothing displayed.

Gone with it: the pending/past models, the history payload in `notifications.json` (which keeps only `{version, dnd}` and rewrites itself once on load if it still carries the old rows), the past pruning, and the `~/.cache/omarchy/notification-images` copies that existed to keep long-lived history thumbnails alive. A migration removes that cache directory.

### DIFF
--- a/migrations/1786517850.sh
+++ b/migrations/1786517850.sh
@@ -1,0 +1,10 @@
+echo "Drop the retired notification image cache"
+
+# Notification history used to be a pair of long-lived lists in
+# notifications.json, and thumbnails for it were copied out of /tmp into this
+# cache so they would outlive the screenshot they came from. History is now the
+# last ten notification files under ~/.local/state/omarchy/notifications/history,
+# which reference their image where it already lives, so nothing writes or reads
+# this directory anymore.
+
+rm -rf "$HOME/.cache/omarchy/notification-images"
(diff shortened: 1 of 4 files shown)


dhh:
Landed on quattro directly as ab57ad65 — closing this in favour of the commit that shipped.

PR #6732, opened 2026-08-12
Only offer video capture devices as webcams

### LINKED ISSUE #5722: omarchy webcam picker offers unusable IPU6 raw Bayer node alongside real capture device
### Summary

On Intel IPU6 laptops (most 11th-gen+ Intel hardware), `omarchy-menu`'s
"With desktop + microphone audio + webcam" flow opens a webcam picker
that lists the IPU6's raw Bayer-only `/dev/video0` alongside the actual
capture device (typically a `v4l2loopback` node fed by `v4l2-relayd` +
`icamerasrc`). Picking `/dev/video0` produces a black overlay or
`STREAMON: Link severed` because that node only advertises raw sensor
formats. Users have to know which entry is the real camera.

### Reproduction

On an IPU6/IPU7 laptop with the Intel camera HAL stack installed
(`intel-ipu6-camera-bin`, `intel-ipu6ep-camera-hal-git`, `icamerasrc-git`,
`v4l2loopback-dkms`, `v4l2-relayd`), `v4l2-ctl --list-devices` returns
two camera groups:

```
ipu6 (PCI:0000:00:05.0):
        /dev/video0
        /dev/video1
        ...
        /dev/video31
        /dev/media0

Built-in Camera (platform:v4l2loopback-042):
        /dev/video42
```

`get_webcam_list` collapses each group to its first device, so the
picker shows two entries:

```
/dev/video0   ipu6 (PCI:0000:00:05.0)
/dev/video42  Built-in Camera (platform:v4l2loopback-042)
```

Picking `/dev/video0` opens ffplay against an output-only / raw-Bayer
node and yields a black overlay; only `/dev/video42` works.

### Related: `omarchy-capture-screenrecording` auto-detect

The auto-detect path in `omarchy-capture-screenrecording`'s
`start_webcam_overlay` (around line 54) is also affected, and arguably
worse - it does:

```bash
WEBCAM_DEVICE=$(v4l2-ctl --list-devices 2>/dev/null \
  | grep -m1 "^[[:space:]]*/dev/video" | tr -d '\t')
```

which always picks `/dev/video0` on this hardware (the first
indented `/dev/video*` line in the output). So `--with-webcam` without
`--webcam-device` always lands on the broken IPU6 raw node. The user
gets a black overlay with no error.

### Suggested fix

Filter both paths to entries whose `Device Caps` include `Video Capture`.
Either parse `v4l2-ctl --device=$dev --info`, or check the udev
`ID_V4L_CAPABILITIES` property - IPU6 ISP nodes report `Video Output` /
`I/O MC` only, while the v4l2loopback node reports `Video Capture`.

On a clean IPU6 + loopback setup this collapses the picker list to a
single entry (`/dev/video42`) and the existing `count == 1` auto-pick
path takes over. On a UVC-only laptop (current behavior), the single
`/dev/video0` entry still passes the filter. No UX changes for non-IPU6
users.

### Environment

- Omarchy 3.8.0
- Intel Core i7-1370P (Raptor Lake-P, IPU6EP)
- Driver/HAL: `intel-ipu6-dkms-git`, `intel-ipu6-camera-bin`, `intel-ipu6ep-camera-hal-git`, `icamerasrc-git`
- Camera surfaced via `v4l2-relayd` to `/dev/video42` (`v4l2loopback`)

yashranaway:
Stops raw IPU6 processing nodes from appearing as webcams and sends screen recording to the usable capture device instead.

<<<< AI wording below >>>>

## Problem

V4L2 exposes raw processing, metadata, and output-only nodes alongside usable cameras. The webcam picker selected the first `/dev/video*` node in each group without checking its device capability, while recorder auto-detection selected the first video node globally.

On IPU6/IPU7 laptops this chooses the raw sensor node, producing a black overlay or `STREAMON: Link severed` instead of using the loopback capture device.

## Fix

Add one shared webcam lister that preserves the existing one-device-per-group behavior but only returns a group when its video node advertises `Video Capture` under `Device Caps`.

Use the filtered list for menu availability, webcam selection, and recorder auto-detection so all three paths agree on what counts as a webcam.

Fixes #5722.

## Verification

- Stubbed IPU6 raw nodes are excluded even when global `Capabilities` contains `Video Capture`.
- A `/dev/video42` loopback capture device and ordinary USB capture device remain available.
- Output-only systems hide the menu entry and the picker reports no webcam.
- Picker and automatic recorder selection use the filtered devices.
- Focused screen-recording tests and CLI metadata tests pass.
- Bash syntax and diff hygiene pass.

Live IPU6 hardware is unavailable in this environment, so reporter validation is still welcome.

### DIFF
--- a/bin/omarchy-capture-screenrecording
+++ b/bin/omarchy-capture-screenrecording
@@ -64,7 +64,7 @@ start_webcam_overlay() {
 
   # Auto-detect first available webcam if none specified
   if [[ -z $WEBCAM_DEVICE ]]; then
-    WEBCAM_DEVICE=$(v4l2-ctl --list-devices 2>/dev/null | grep -m1 "^[[:space:]]*/dev/video" | tr -d '\t')
+    WEBCAM_DEVICE=$(omarchy-capture-webcam-list | sed -n '1s/[[:space:]].*//p')
     if [[ -z $WEBCAM_DEVICE ]]; then
       omarchy-notification-send -u critical -t 3000 "No webcam devices found"
       return 1
--- a/bin/omarchy-capture-screenrecording-with-webcam
+++ b/bin/omarchy-capture-screenrecording-with-webcam
@@ -3,21 +3,7 @@
 # omarchy:summary=Pick a webcam and start a screen recording with it
 # omarchy:examples=omarchy capture screenrecording-with-webcam
 
-set -o pipefail
-
-webcam_list() {
-  v4l2-ctl --list-devices 2>/dev/null | while IFS= read -r line; do
-    if [[ $line != $'\t'* && -n $line ]]; then
-      local name="$line"
-      local device
-      IFS= read -r device || break
-      device=$(tr -d '\t' <<<"$device" | head -1)
-      [[ -n $device ]] && echo "$device  $name"
-    fi
-  done
-}
-
-mapfile -t devices < <(webcam_list)
+mapfile -t devices < <(omarchy-capture-webcam-list)
 if (( ${#devices[@]} == 0 )); then
   omarchy-notification-send "No webcam devices found" -u critical -t 3000
   exit 1
--- a/bin/omarchy-capture-webcam-list
+++ b/bin/omarchy-capture-webcam-list
@@ -1,0 +1,31 @@
+#!/bin/bash
+
+# omarchy:summary=List webcam devices that support video capture
+# omarchy:hidden=true
+
+capture_capable() {
+  local device="$1"
+
+  v4l2-ctl --device "$device" --info 2>/dev/null | awk '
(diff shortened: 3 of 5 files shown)


dhh:
Reviewed. The Device Caps parsing is right (the test proving global Capabilities can't fool it is a nice touch), every caller degrades correctly with zero surviving devices, and v4l-utils arrives transitively via gpu-screen-recorder → ffmpeg, so there's no new dependency concern. I pushed one hardening commit to the branch: a group whose first video node wasn't capture-capable vanished entirely even when a later node in the group could capture, so the helper now probes each node until one qualifies (still one device per group). It also exits zero explicitly — a trailing filtered device used to leak the failed capability check as the script's exit status. Both are pinned by new test cases.

PR #6733, opened 2026-08-12
Show a notification the sender updated, instead of the version it replaced

dhh:
Stacked on #6731 — the base is `notification-history-on-disk`, so this diff is just the one commit. GitHub retargets it to `quattro` when that merges.

A client that updates a notification through `replaces_id` does not produce a second `onNotification`: Quickshell writes the new content onto the `Notification` object the shell is already holding. The card draws a snapshot copied out of that object — deliberately, since a live QObject in a ListModel role becomes a dangling pointer the moment the server destroys it — so nothing reached the screen. A Slack thread that updates in place read as stuck on its first message, and that stale version is what got archived to history.

Every property the card draws is now watched on the notification we hold, and a change rewrites both the model row and the file the popup was persisted under. That file name is the popup's identity, so the rewrite lands in place: a shell restart restores the version last shown, and so does the copy that reaches history.

The countdown starts over when the content changes — new text arriving a second before the toast was due to expire deserves a full look, not the remainder of the clock the text it replaced had nearly run through.

Verified against the running shell: a replaced toast now shows the new text; an update at t=6s of an 8s toast kept it up until 14s; history archived the updated version; a replacement arriving after the original expired still comes through as its own fresh toast; and dismissing one toast out of several leaves the others' timers alone (the survivor still expired at 8.07s).

### DIFF
--- a/shell/plugins/notifications/NotificationLogic.js
+++ b/shell/plugins/notifications/NotificationLogic.js
@@ -92,6 +92,17 @@ function snapshotOf(notification, timestamp) {
   }
 }
 
+// A client updating a notification through replaces_id keeps the identity of
+// the popup it took over: the file name is the timestamp and id the popup was
+// first persisted under, and the restore, replace and archive paths all key
+// off that name. Only what the card draws comes from the updated object.
+function replacementSnapshot(notification, originalId, timestamp) {
+  var updated = snapshotOf(notification, timestamp)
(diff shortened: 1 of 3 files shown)


dhh:
Landed on quattro directly as cd84583b — closing this in favour of the commit that shipped. The Copilot findings on both PRs are being addressed in a single follow-up PR.

PR #6737, opened 2026-08-12
Restore FPC fingerprint detection in Quattro

Vtorrealba:
## Summary

- restore fingerprint detection for normalized USB product strings containing `fpc`
- keep `10a5` out of the broad vendor allowlist so unrelated devices are not treated as readers
- make the USB sysfs root injectable for focused hardware tests
- cover FPC, existing product/vendor matches, kernel-driver rejection, and no-match behavior

## Context

The Acer Swift's `10a5:a305` reader reports its product as `FPC L:0000 FW:1425046`, without the words fingerprint or biometric. The same reader worked under Omarchy 3.8.4, and #2023 previously added `FPC` to the old `lsusb` detector. Quattro's sysfs rewrite accidentally omitted that product signature.

`libfprint`/`fprintd` successfully enroll and authenticate this reader, including for sudo. On the real hardware, the current packaged detector exits 1 while this patched detector exits 0.

## Verification

- `bash test/shell.d/hw-fingerprint-test.sh`
- `./test/all` — the fingerprint test and CLI suite pass; two unrelated live-QML tests fail identically on an untouched `upstream/quattro` worktree (`bar-icon-geometry-test.sh` and `runtime-smoke-test.sh`)
- Acer Swift SFG14-73T with `10a5:a305 FPC L:0000 FW:1425046`: packaged detector exits 1; patched detector exits 0

### DIFF
--- a/bin/omarchy-hw-fingerprint
+++ b/bin/omarchy-hw-fingerprint
@@ -9,6 +9,7 @@
 # also make USB touchscreens) are left out to avoid nagging laptops with no
 # reader — those still match on the product string below when present.
 fingerprint_vendors=" 27c6 138a 06cb 08ff 1c7a 147e "
+usb_devices_path="${OMARCHY_USB_DEVICES_PATH:-/sys/bus/usb/devices}"
 
 # libfprint drives every reader it supports from userspace over libusb, so a
 # real reader sits there with no kernel driver bound to any of its interfaces.
@@ -24,16 +25,15 @@ has_kernel_driver() {
   return 1
 }
 
-for dev in /sys/bus/usb/devices/*; do
+for dev in "$usb_devices_path"/*; do
   # The device's own product descriptor usually names it, e.g. "Goodix
   # Fingerprint USB Device" — driver-independent and vendor-agnostic.
   if [[ -r $dev/product ]]; then
     product=$(<"$dev/product")
     product=${product,,}
-    # Elan's match-on-chip readers report "ELAN:ARM-M4", the family name
-    # rather than the function. Elan is left out of the vendor list above
-    # on purpose, so without this they match nothing.
-    [[ $product == *fingerprint* || $product == *biometric* || $product == *elan:arm-m4* ]] && exit 0
+    # Elan's match-on-chip readers and FPC readers identify their family or
+    # manufacturer rather than their function, so match those signatures too.
+    [[ $product == *fingerprint* || $product == *biometric* || $product == *elan:arm-m4* || $product == *fpc* ]] && exit 0
   fi
 
   if [[ -r $dev/idVendor ]]; then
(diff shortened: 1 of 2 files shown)


dhh:
Pushed two commits to this branch — thanks for the clean seam, the `OMARCHY_USB_DEVICES_PATH` injection matches what `omarchy-hw-nvidia` and `omarchy-hw-display` already do, so it dropped straight in.

**Correction to an earlier version of this comment.** It claimed FPC shows up mid-string in the product descriptors of unrelated peripherals and cited `USB2.0 FPC Camera` as if that were an observed device. It isn't — I constructed it as a test fixture and should not have presented it as evidence. There is no known device that collides here. Apologies for the noise.

**Anchored the FPC match**, on the narrower argument that it costs nothing. `*fpc*` is three letters on the one branch that's trusted outright, with none of the kernel-driver checking the vendor guess gets. Every FPC reader on record leads with the token:

| Device | Product string |
|---|---|
| 10a5:a305 (yours) | `FPC L:0000 FW:1425046` |
| [10a5:9800](https://linux-hardware.org/?id=usb:10a5-9800) | `FPC Sensor Controller L:0002 FW:25.26.23.14` |
| [10a5:0007](https://linux-hardware.org/?id=usb:10a5-0007), [10a5:e340](https://linux-hardware.org/?id=usb:10a5-e340), [10a5:9200](https://linux-hardware.org/?id=usb%3A10a5-9200) | `FPC Sensor Controller` |

So requiring the prefix loses no coverage, and there's now a case pinning the sensor-controller string too. If you know of a reader that reports FPC anywhere other than the front, say so and this should go back to a bare token.

**Fixed a latent trap in the test helper.** `${remainder#*:}` returns the string unchanged when there's no second colon, so `'27c6:1234'` would have written `1234` out as the product string. Your call sites all passed a trailing colon so the suite was right, but the next person adding a no-product case would have got a quietly wrong test.

Worth recording for whoever revisits this: routing `10a5` through `fingerprint_vendors` looks tidier, and every FPC device above is that vendor with no driver bound. The concern is that while `fprintd` holds the reader the interface's `driver` symlink points at `usbfs`, which would make `has_kernel_driver` reject a working machine — but I have not confirmed that on hardware, and if it holds it's a pre-existing issue for the Goodix entries too, not something this PR introduces.

Branch was force-pushed once to correct the commit message carrying the same overstated claim.

dhh:
Two follow-ups after a Codex review of this branch.

**Softening something I wrote above.** I said the prefix covers "every FPC reader on record." That's stronger than what I checked — libfprint lists ten `10a5` product IDs and I'd seen descriptors for four. All four lead with `FPC `, and no counterexample turned up, but that's the honest limit of it. If you know a reader that reports FPC anywhere but the front, this should go back to a bare token.

**Pushed one more commit** (`fe89ed2b`), replacing the bare `touch driver` in the kernel-driver fixture with a link to a named driver directory, the way sysfs does it. The old fixture asserted that *any* driver disqualifies a vendor guess, which turns out to be too strong: libusb claims interfaces through a synthetic `usbfs` driver, so a reader binds one while it's being enrolled or verified against. The case still covers what it was written for — a bridge or a camera on a fingerprint vendor ID — without pinning behaviour it was never about.

That's a pre-existing bug rather than anything this PR introduced, and it's #6744, stacked on this branch. Nothing needed here for it.

Issue #6574, opened 2026-08-06
Issue with the omarchy.keyboard-layout plugin not visually switching layout


link1183:
### System details

Omarchy Quattro beta 1

### What's wrong?

Here's my input.lua:

```lua
hl.config({
	input = {
		-- Use multiple keyboard layouts and switch between them with Left Alt + Right Alt.
		kb_layout = "us,ch",
		kb_options = "compose:caps,shift:both_capslock_cancel,grp:alts_toggle",

		-- Use a specific keyboard variant if needed (e.g. intl for international keyboards).
		kb_variant = ",fr",
})
```

As you can see, I am enabling the US keyboard and the CH-fr keyboard.

When switching keyboard by pressing left + right alt, the icon of the keyboard-layout plugin stays on `ENG`.

PR #6740, opened 2026-08-12
Follow the keyboard being typed on in the layout widget

KazeTachinuu:
The keyboard layout widget disappears from the bar when no keyboard on the seat holds Hyprland's `main` flag, which is the normal state with fcitx5 running.

```
[tray] [agents] [bt] EN [wifi] [audio]     a keyboard holds main
[tray] [agents] [bt]    [wifi] [audio]     fcitx5 holds main
```

```
at-translated-set-2-keyboard  main=False  English (US)
power-button                  main=False  English (US)
hl-virtual-keyboard-fcitx5    main=True   English (US)
```

fcitx5's virtual keyboard holds `main`, and virtual keyboards are filtered out, so the pick lands on nothing: no label, and the widget hides itself. #6727 keeps polling in that state, but the poll has nothing new to read. When fcitx5 unbinds, `main` jumps to a button that never sees the layout toggle, which is #6574 again.

Only the keyboard being typed on advances through the layout list, so read the furthest-advanced one instead. `activelayout` names the keyboard it moved ahead of the layout it moved to, so take that name: it settles the pick, and the click that switches it, which now waits to be told a keyboard rather than guessing at one.

Rebased on #6727.

### DIFF
--- a/shell/plugins/bar/widgets/KeyboardLayout.qml
+++ b/shell/plugins/bar/widgets/KeyboardLayout.qml
@@ -12,9 +12,15 @@ BarWidget {
 
 
   property string layoutFull: ""
+  // The keyboard the last reading spoke for, which is the one a click switches,
+  // and separately the one activelayout named as being typed on. A reading
+  // confirms the first is really there, so the click has a keyboard to reach
+  // from the first reading onwards rather than only after a switch, and stops
+  // naming one that has been unplugged.
   property string keyboardName: ""
-  // Real keyboards on the seat, virtual ones excluded, and whether the last
-  // reading left that shape in doubt.
+  property string typedKeyboardName: ""
+  // Keyboards on the seat, buttons and virtual ones excluded, and whether the
+  // last reading left that shape in doubt.
   property int keyboardCount: 0
   property bool keyboardUnresolved: false
   // Nothing to read or switch on the single-layout install most people run, so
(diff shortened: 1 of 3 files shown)


KazeTachinuu:
Right on both.

`all` advanced the devices the pick relies on: click, then a keyboard-only toggle, and the buttons sit a layout ahead, so a cold start reads one of them. Dropped that commit. `current` is no substitute either, it hit `hl-virtual-keyboard-fcitx5` while fcitx5 was bound and `thinkpad-extra-buttons` when it wasn't. The click now switches only the keyboard `activelayout` named.

The race is #6727's now. Rebased onto it, my guard for it is gone.

KazeTachinuu:
`all` is what the earlier comment flagged, and it holds: it breaks the baseline the pick relies on.

`current` hit `hl-virtual-keyboard-fcitx5`, then `thinkpad-extra-buttons`. `main` lands on buttons too, which is the bug here. Nothing identifies the typed keyboard before a switch, so the click waits for one.

It already does nothing on this seat: with fcitx5 on `main` the pick is undefined at startup and `cycleLayout` returns on the same guard. Only seats where `main` sits on a real keyboard change.

Can trust `main` for the target, never the label, if you'd rather. Two lines.

dhh:
Pushed a commit onto this branch — happy to back it out. The read is right; it's the removal that broke things.

**`keyboardName` was doing two jobs**: the device a click switches, and the device `activelayout` named. Dropping the assignment from the query left only the second set, so `cycleLayout()` returned early until a switch had happened — and on `us,dk` the click *is* the only switch, since `grp:alts_toggle` is only added for non-Latin layouts. Same property was the poll's "first reading landed" latch, so a single-keyboard desktop polled `hyprctl` every 10s forever. Split in two.

**The buttons had to come out of the list.** Progress only tracks the typed keyboard while everything else stays put, but `power-button`/`lid-switch`/`sleep-button` answer to `switchxkblayout` and can hold `main`. So before anything switches, the click can land on a button — which then reads as furthest-along and keeps the label. Unplug the keyboard beside one and it stands in for the seat, with polling stopped. Filtering them where the virtual keyboards already are fixes both and leaves your selector alone. It also makes `keyboardCount` mean what the poll comment assumes: a laptop counts 1 and stops polling.

I first tried `switchxkblayout all next` to avoid naming a device, and dropped it: it advances the buttons too, so a toggle that wraps the keyboard to the first layout leaves a button furthest-along. One click plus one toggle on `us,dk`. Worth knowing since the description promises `all`.

Also: `event.parse(2)` with `data` fallback (matches `IdleModel.js`, and keeps a comma in a layout description intact); a reading that finds no keyboards clears the label, told apart from a killed query by the device list itself; the watchdog now retries.

Not changed: a seat whose only keyboard is unplugged still freezes, since polling has stopped and Hyprland has no device event to wake it. And `selectKeyboard` compares raw indices, which don't compare across per-device `kb_layout`.

PR #6741, opened 2026-08-12
Report the link carrying a VPN rather than the tunnel

KazeTachinuu:
The network panel says Ethernet while you're on wifi, whenever a VPN is up. The header reads Ethernet with the wifi icon sitting right above it in the bar, since the two come from different sources.

```
$ omarchy-network-status
ethernet	<tunnel>
```

It takes whatever device holds the internet route and calls it ethernet if it isn't wireless. A wireguard or tun interface is neither, and the rest of the reading comes off the tunnel instead of the wifi card: no gateway, so the router ping never runs; byte counters and link speed from an interface that has no speed; an error printed from `cat /sys/class/net/<tunnel>/speed`; and the header network name, the band section and the QR share all go missing.

Fix: find the real interface first, then ask whether it's wifi or ethernet. Real hardware has a wireless directory or a device symlink in `/sys/class/net`, and when a tunnel holds the route, the main routing table still points at the card carrying it.

```
$ omarchy-network-status
wifi	<ssid>	85	5540.0
```

If nothing on the path has hardware behind it, inside a container for instance, it keeps the routed device rather than claiming you're disconnected.

### DIFF
--- a/bin/omarchy-network-status
+++ b/bin/omarchy-network-status
@@ -19,17 +19,46 @@ case "${1:-}" in
     ;;
 esac
 
+# Wifi or ethernet is a question only an interface with hardware behind it can
+# answer: a wireless directory or a device symlink in sysfs. A tunnel has
+# neither, so asking it returns ethernet for want of anywhere else to land,
+# which is how a VPN over wifi came to read as a wired connection.
+is_physical() {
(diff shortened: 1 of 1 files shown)


bjarneo:
Automated duplication check: this pull request looks similar to #8866, which covers the same physical-link resolution under VPN tunnels. I keep that one open and close this one to consolidate review.

PR #6742, opened 2026-08-12
Open Spotify links in the native app instead of the browser

apostaat:
# Open Spotify links in the native app instead of the browser

## Problem

Clicking a link like `https://open.spotify.com/album/6raV6eHqsTKFsCu0vIXjQA` (from Telegram, Discord, a terminal, etc.) always opens the browser, even when the native Spotify client is installed. Desktop Linux has no equivalent of Android App Links / iOS Universal Links, so an app can never claim an `https` domain — the URL goes to whatever handles `x-scheme-handler/https`, i.e. the browser.

## Solution

A small dispatcher, `omarchy-url-open`, registered as the handler for `x-scheme-handler/http` and `https`:

- `open.spotify.com/{track,album,artist,playlist,show,episode}/...` URLs (including localized `/intl-xx/` paths) are converted to `spotify:` URIs and delivered to the client over D-Bus via MPRIS `OpenUri`, then the window is raised.
- On a cold start the client is launched first and the script waits (up to 15s) for it to claim its D-Bus name — the Linux client silently drops URIs passed via `--uri=` while booting, so passing the URI on the command line is not reliable.
- If Spotify is not installed, or the URL is anything else, it falls through to `omarchy-launch-browser`.

The real browser is recorded in the `text/html` handler (shipped in `default/applications/mimeapps.list`). `omarchy-default-browser`, `omarchy-launch-browser`, and `omarchy-remove-browser` read that instead of `xdg-settings get default-web-browser`, because on generic desktops `xdg-settings` resolves the browser from the `http` handler — which now points at the dispatcher itself. Every place that runs `xdg-settings set default-web-browser` (`omarchy-default-browser`, `omarchy-provision-user`, `omarchy-remove-browser`) re-pins the dispatcher afterwards, since `xdg-settings set` claims `x-scheme-handler/http(s)` for the browser.

A migration registers the dispatcher on existing installs, preserving the user's current browser choice in `text/html` first.

## Testing

- `test/shell.d/url-open-test.sh` covers the dispatcher with mocked D-Bus: album links become `spotify:album:...` `OpenUri` calls, localized `/intl-xx/` links parse to the underlying resource, and non-Spotify or unsupported links fall through to the browser.
- `test/acceptance.d/system-test.sh` now asserts `x-scheme-handler/http` routes through the dispatcher and `text/html` records Chromium.
- Manually on an install with the native client (Arch `spotify` package): `xdg-open` on album/track links opens the client (cold and warm start), `?si=` and `/intl-ru/` URLs parse, non-Spotify URLs open the default browser, `omarchy-default-browser` query/set still resolves the real browser.

## Notes

- The dispatch is structured as a `case` on the URL, so handlers for other apps with native clients (Zoom, Steam, Slack) can be added later with a few lines.
- Links clicked *inside* the browser are unaffected — the browser navigates directly and never consults xdg. This only affects links opened from other applications.


dhh:
Dirty repo. Needs to be made against quattro.

apostaat:
@dhh  Rebased onto quattro — the diff is now just the 10 files for this feature. Adapted to the quattro layout: handler registration moved from install/config/mimetypes.sh to default/applications/mimeapps.list, the dispatcher is re-pinned after every xdg-settings set default-web-browser call (omarchy-default-browser, omarchy-provision-user, omarchy-remove-browser), the migration uses $OMARCHY_PATH, and the shell/acceptance tests cover the new routing.

PR #6756, opened 2026-08-12
Show a plug icon when plugged in but not charging

vivek7405:
In 3.8, waybar showed the plugged-in glyph once the battery stopped taking a charge. The Quickshell power widget never picked that up: at 100% on AC it shows the charging icon, which reads as though current is still flowing into a battery that is already full.

The same is true while a firmware charge threshold holds the battery below full. That case showed the plain battery icon, indistinguishable from running on battery at the same level.

This widens the icon rule to "on external power with nothing flowing", which is the `batteryFlowIdle` state the panel already tracked:

* Plugged in and idle, either topped up or held by a charge threshold, shows a plug.
* Actively charging keeps the charging icon.
* On battery keeps the plain battery icon.

The panel still separates the two idle cases in words, labelling them "Fully charged" and "Threshold", so nothing is lost where there is room to say it.

The glyph is `nf-md-power_plug` (U+F06A5) rather than 3.8's `nf-fa-plug`, so the stroke weight matches the surrounding Material Design battery icons.

`fullyCharged`, `batteryFull`, and `batteryFlowIdle` move from `Panel.qml` into `Model.js`, since the icon and the panel's stats rows now depend on the same predicates and should not be able to drift apart.

## Verification

Checked on a laptop reporting `state: fully-charged`, `percentage: 100%`, `OnBattery: false`. The bar and the panel hero both render the plug at icon size, the panel still reads "FULLY CHARGED / 100%" with its stats rows intact, and the shell log shows no QML errors or binding loops.

`test/shell.d/power-test.sh` covers the new states, including the threshold hold, which cannot be exercised without writing `charge_control_end_threshold` on real hardware.

One consequence worth flagging: on a charge-limited machine, plugging in below the start threshold can now briefly show the plug before the charging icon, because UPower reports `FullyCharged` for a moment there (the window noted at `Panel.qml:428`). It is a truthful reading of that instant, but it is new visible behaviour.

### DIFF
--- a/shell/plugins/panels/power/Model.js
+++ b/shell/plugins/panels/power/Model.js
@@ -63,6 +63,25 @@ function chargeThresholdActive(device, onBattery, states) {
   return Number(d.changeRate || 0) <= 0.2 || Number(d.timeToFull || 0) >= 8 * 60 * 60
 }
 
+function fullyCharged(device, onBattery, states) {
+  var d = device || {}
+  return !!(d.isPresent && d.state === (states || {}).FullyCharged && !chargeThresholdActive(d, onBattery, states))
+}
+
+// Full covers the window where the battery has reached 100% but the state has
+// not caught up yet, which some firmware still reports as charging.
+function batteryFull(device, onBattery, states) {
+  var d = device || {}
+  if (!d.isPresent) return false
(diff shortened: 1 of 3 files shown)


bjarneo:
Automated duplication check: this pull request looks similar to #8815, which covers the same battery plug indication. I keep that one open and close this one to consolidate review.

Issue #6667, opened 2026-08-09
Keyboard backlight restore fails after lock (brightnessctl saves 0 after Dell stop_timeout)


sotoreynah:
### Omarchy version
3.8.4

### Hardware
Dell laptop with `dell::kbd_backlight` (max brightness 2)

### Summary
After idle lock/unlock, the keyboard backlight stays off until the physical backlight key is pressed. Software restore does not bring it back, and typing does not re-enable it either.

### Expected
`omarchy-system-wake` restores the previous non-zero keyboard backlight level.

### Actual
Backlight stays at 0 after unlock. Only the hardware backlight key restores it.

### Root cause
Dell exposes an idle timeout via sysfs:

- `/sys/class/leds/dell::kbd_backlight/stop_timeout` (default `1m`)
- After ~1 minute idle, brightness is already `0`

Lock path in `omarchy-system-lock` (when not `OMARCHY_LOCK_ONLY`) does:

```bash
omarchy-brightness-keyboard off   # brightnessctl -s set 0
```

`omarchy-brightness-keyboard off` is effectively:

```bash
brightnessctl -sd "$device" set 0
```

That **saves the current value, then sets 0**. By lock time (~2.5 min via hypridle), current is already **0** from `stop_timeout`, so the saved state is **0**.

Wake then runs:

```bash
omarchy-brightness-keyboard restore   # brightnessctl -r
```

which restores **0**. Explicit software `set 0` also tends to leave Dell in a full “off” mode where activity / `start_triggers` no longer revive the light until the physical key is used.

### Repro
1. Set keyboard backlight to a non-zero level
2. Idle until hypridle locks (or lock via system lock)
3. Unlock
4. Keyboard backlight remains off

### Confirmed on system
- Save file can contain `0`: `/run/user/$UID/brightnessctl/leds/dell::kbd_backlight`
- `stop_timeout=1m`, `start_triggers=+keyboard +touchpad`
- Related scripts: `omarchy-brightness-keyboard`, `omarchy-system-lock`, `omarchy-system-wake`

### Suggested fix
In `omarchy-brightness-keyboard` (and/or lock/wake):

1. Persist **last non-zero** level in a state file
2. On `off`: save only if current > 0; never overwrite state with 0
3. On `restore`: apply last non-zero level (fallback to a sane default, e.g. mid step)
4. Optionally snapshot level on idle *before* `stop_timeout` (e.g. ~30s)

Happy to test a PR on Dell hardware.

[...]

PR #6760, opened 2026-08-12
Preserve keyboard brightness across hardware timeouts

yashranaway:
Remembers the keyboard brightness selected with up/down/cycle, including an intentional off, so hardware timeouts don't erase the level to restore after unlock. Addresses #6667.

<<< AI wording below >>>

## Problem

A zero reading at lock time cannot distinguish a hardware timeout from a manual off. Saving that zero loses the user's lit setting; skipping it without tracking manual changes can restore an older non-zero setting after the user turned the light off. Relying on a previous lock-time save also misses timeouts before the first lock.

## Fix

Record successful manual brightness changes in a per-device file under `XDG_RUNTIME_DIR`, including zero. Serialize hardware changes and state updates, and replace the state file atomically. Failed hardware changes leave the saved selection intact.

On `off`, leave an already-zero keyboard alone so an unnecessary software `set 0` does not interfere with Dell hardware wake triggers. A lit first lock seeds the state if no manual selection has been recorded. Restore reapplies the recorded value without clearing it, so repeated blank and wake calls retain the selection.

When a session starts at zero without any recorded selection, leave it unchanged. The command cannot infer whether that zero came from a firmware timeout or an intentional off. Changes made entirely through firmware or another utility are not tracked by the up/down/cycle path.

## Verification

- Reproduced both reported gaps against the previous implementation: manual off restored brightness 2 instead of 0, and a timeout before the first lock restored 0 instead of the selected 1.
- `bash test/shell.d/brightness-keyboard-test.sh` passes, covering manual off, cycle to zero, first-lock timeout, repeated blank/restore, initial state, failed reads/writes/sets, restore retries, and normal step/OSD behavior.
- `bash -n bin/omarchy-brightness-keyboard test/shell.d/brightness-keyboard-test.sh` and `git diff --check` pass.
- `./test/cli` passes with `python3` temporarily exposed as `python` in the test environment; `bash test/shell.d/bin-style-test.sh` passes.

Tests use a simulated LED device and stateful `brightnessctl` stub. This machine has no keyboard-backlight device, so Dell hardware timeout and wake-trigger behavior still needs hardware confirmation.


mmahut:
Looks like a duplicate of #10364, which fixes the same double-save in `off` and also covers a manually disabled backlight.

yashranaway:
Thanks for flagging this. This PR targets the hardware timeout happening before the first blank call. #10364 appears to save zero in that case. Does it also cover that scenario?

mmahut:
You're right, #10364 snapshots 0 in the timeout case and restores 0, so it doesn't cover #6667. My mistake.

Your approach has a gap the other way though: if I turn the keyboard off manually with down and then lock, off skips the save and restore brings back a stale non-zero level. And it only works at all if some earlier lock saved a non-zero value.

At lock time a reading of 0 can't tell "hardware timed out" from "user turned it off". Only the up/down/cycle path knows. 

I'd rather fix it as #6667 suggests: record the last user-set level in a runtime state file on manual changes, skip set 0 when already at zero, and restore the recorded level. Would you be up for reworking this in that direction, with the manual-off and repeated-blank cases in the test too?

yashranaway:
Good catch. I reworked this to remember successful manual brightness changes, including off, and added regressions for manual off and a timeout before first lock.

PR #6764, opened 2026-08-12
Stop ttfx before closing the screensaver terminal

### LINKED ISSUE #6762: ttfx dumps core on every idle lock: omarchy-system-lock kills the screensaver terminal before the renderer
**What happens.** Every time the idle lock fires while the screensaver is up, `ttfx` aborts and systemd-coredump writes a ~1 MB core. 25/25 ttfx core dumps on this machine occurred within milliseconds of `omarchy-system-lock`.

**Expected.** The screensaver shuts down cleanly on lock, as it does on dismissal.

**Cause.** `omarchy-system-lock` ends with `pkill -f '[o]rg.omarchy.screensaver'`, which matches the screensaver *terminal* but not `ttfx` (whose cmdline contains no such string). Killing the terminal closes the pty master; ttfx, rendering at 120 fps, hits EIO on its next write, tries to report it via `eprintln!`, and that write also fails:

```
thread 'main' panicked at library/std/src/io/stdio.rs:1166:9:
```

(Panic text recovered from the core's memory; the binary is stripped so the backtrace itself does not symbolize.)

`omarchy-screensaver`'s own `exit_screensaver()` already uses the correct order — `pkill -x ttfx` first, then the terminal — so the dismissal path is clean and only the lock path is affected.

**Timeline of one instance:**

```
13:14:12.860  idle: lock-system: lock-timeout -> omarchy-system-lock
13:14:13.002  lock script exits 0
13:14:13.008  ttfx aborts (SIGABRT)
13:14:13.043  hyprland: failed to read client connection (terminal)
13:14:13.492  lock secure=true
```

**Repro.** Let the screensaver start, then let the idle lock deadline pass without touching input. `coredumpctl list | grep ttfx`.

**Fix.** Kill `ttfx` before the terminal in `omarchy-system-lock` (mirroring `exit_screensaver`), and/or make ttfx exit quietly on EIO/EPIPE from the output tty.

**System.** Omarchy 4.0.0.r1687.gf5893dd-1, ttfx 0.3.1-1, terminal `foot`. Backtrace is unsymbolized — ttfx ships stripped and is not on debuginfod.archlinux.org.

dhh:
Fixes #6762.

`omarchy-system-lock` currently kills the screensaver terminal while `ttfx` is still rendering. Closing the PTY makes the next frame write return `EIO`; reporting that error to the same dead PTY then panics and writes a core dump.

Stop `ttfx` before closing the terminal, matching the existing dismissal path in `omarchy-screensaver`. Add a mocked command-order regression test so the lock path can be exercised without locking the test session.

### DIFF
--- a/bin/omarchy-system-lock
+++ b/bin/omarchy-system-lock
@@ -20,4 +20,5 @@ if pgrep -x "1password" >/dev/null && omarchy-cmd-present 1password; then
 fi
 
 # Avoid running screensaver when locked
+pkill -x ttfx 2>/dev/null || true
 pkill -f '[o]rg.omarchy.screensaver' 2>/dev/null || true
--- a/test/shell.d/system-lock-test.sh
+++ b/test/shell.d/system-lock-test.sh
@@ -1,0 +1,34 @@
+#!/bin/bash
+
+set -euo pipefail
+
+source "$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd)/base-test.sh"
+
+tmpdir=$(mktemp -d)
+trap 'rm -rf "$tmpdir"' EXIT
+
+mock_bin="$tmpdir/bin"
+call_log="$tmpdir/calls"
+mkdir -p "$mock_bin"
+
+for command in omarchy-shell hyprctl pkill; do
+  cat >"$mock_bin/$command" <<'SH'
+#!/bin/bash
+printf '%s %s\n' "$(basename "$0")" "$*" >>"$CALL_LOG"
+SH
+done
+
+cat >"$mock_bin/pgrep" <<'SH'
+#!/bin/bash
+exit 1
+SH
+chmod +x "$mock_bin"/*
+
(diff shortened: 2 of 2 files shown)


grantwiley:
This reduces the race, but it does not remove it.

`pkill -x ttfx` only sends `SIGTERM`; it does not wait for `ttfx` to exit. `ttfx` catches that signal and finishes shutdown from its animation loop so it can clean up the terminal. The next line can therefore close the screensaver terminal while `ttfx` is still running.

If that happens, the next frame write hits `EIO` because the PTY is gone. `ttfx` then tries to print the error to the same closed terminal, that write fails too, and Rust aborts with the core dump reported in #6762.

The test here only checks that the two `pkill` commands are called in order. Because `pkill` is mocked, it does not check that the process has actually exited between them.

This reproduced on `4.0.0.r1688.g438f7b3-1` after this change was installed. #6773 adds a bounded wait for `ttfx` to exit before closing the terminal and updates the test to check that wait.

PR #6766, opened 2026-08-12
fix(hypr): scale the floating agent window with the monitor

markbus-ai:
The agent window rule pinned every agent terminal to a fixed 1200x800 float. On small screens that overflows the monitor (on a 1366x768 laptop the window was taller than the screen), and on large ones it stays arbitrarily small.

Scale it with the monitor instead, using the same windowrule expressions `pip.lua` and `webcam-overlay.lua` already rely on: 3/5 of the monitor width by 4/5 of its height, still floated and centered.

Verified live on a 1366x768 session: the window now opens 820x614, centered, and `test/shell.d/default-agent-test.sh` passes.

### DIFF
--- a/default/hypr/apps/agent.lua
+++ b/default/hypr/apps/agent.lua
@@ -1,3 +1,4 @@
 -- omarchy-agent gives every agent terminal the same app-id, so this
--- floats the keybinding, the menu, and a crash diagnosis alike.
-o.window("org\\.omarchy\\.agent", { float = true, center = true, size = { 1200, 800 } })
+-- floats the keybinding, the menu, and a crash diagnosis alike. The size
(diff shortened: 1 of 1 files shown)


dhh:
Thanks for catching this — the fixed 1200x800 really was overflowing, and more widely than the report suggested: window rules see logical pixels, so a 2560x1440 monitor at scale 1.6 is only 1600x900 to Hyprland and the old size covered 89% of its height.

We went a different way with the fix in #6769. Rather than deriving the size from the monitor, the agent now tiles like a normal window, which drops the fixed size along with the floating rule and means there is no size expression left to tune per display.

Closing in favor of that one. Appreciate the report and the patch.

PR #6775, opened 2026-08-12
Recover the wallpaper picker after interrupted thumbnails

### LINKED ISSUE #6774: [Bug] [Quattro]: Interrupted thumbnail generation can permanently hide wallpapers
### System details

Omarchy 4.0.0.r1688.g438f7b3-1 (Quattro), kernel 7.1.8-arch1-3

### What's wrong?

If thumbnail generation from #6686 is interrupted, `.jpg.lock` directories can remain. The next run drops thumbnails blocked by those locks, then caches the incomplete rows with valid directory signatures. Later picker opens reuse that partial cache indefinitely.

Reproduce:
1. Start the background picker with several uncached images.
2. Stop it while `vipsthumbnail` workers are running.
3. Open the picker again.

Expected: every image is retried and shown.

Actual: only thumbnails completed before interruption are shown. In one occurrence, four locks were created together but only one thumbnail completed. Every source image generated successfully after clearing the stale cache; there was no OOM or decode failure.

AksharP5:
The wallpaper picker can permanently lose backgrounds if a thumbnail batch is interrupted. A later run treated stranded locks as active, removed missing thumbnails from the picker rows, then cached that incomplete set as valid.

This uses process-owned `flock` locks, removes legacy lock directories, invalidates affected row caches, and withholds cache signatures when any thumbnail is still missing.

Closes #6774.

Relevant regression: #6686.

## Testing

- `bash test/shell.d/menu-images-test.sh`
- `./test/all` (166 test files passed)

### DIFF
--- a/bin/omarchy-menu-images
+++ b/bin/omarchy-menu-images
@@ -146,8 +146,11 @@ generate_thumbnail() {
   local tmp="$thumbnail.$$.jpg"
 
   # Older releases used directories as locks, which could survive a killed
-  # generator and block this thumbnail forever.
-  [[ -d $lock ]] && rmdir "$lock" 2>/dev/null
+  # generator and block this thumbnail forever. Only reap aged ones, so a
+  # legacy generator still running through an upgrade keeps its lock.
+  if [[ -d $lock ]] && (( $(date +%s) - $(stat -c '%Y' "$lock" 2>/dev/null || date +%s) > 120 )); then
+    rmdir "$lock" 2>/dev/null
+  fi
 
   exec {lock_fd}>"$lock" || return
   flock -w 30 "$lock_fd" || return
@@ -159,7 +162,9 @@ generate_thumbnail() {
   [[ -f $thumbnail ]] && return
 
   # Callers fan out one generator per image, so keep each vips single-threaded.
-  if VIPS_CONCURRENCY=1 vipsthumbnail "$image" --size 1536x864 --smartcrop=centre --path "$tmp[Q=82,strip]"; then
+  # Close the lock fd for vips: an orphaned or hung vips must not keep holding
+  # the lock after this shell is killed.
+  if VIPS_CONCURRENCY=1 vipsthumbnail "$image" --size 1536x864 --smartcrop=centre --path "$tmp[Q=82,strip]" {lock_fd}>&-; then
     mv -f "$tmp" "$thumbnail"
   else
     rm -f "$tmp" "$thumbnail"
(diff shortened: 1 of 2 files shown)


dhh:
Reviewed this against #6774: read the pipeline end to end, ran the shell and CLI suites, and put the full diff through an adversarial review pass. The core approach is right, and I verified the original failure is closed:

- Replacing `mkdir` directory locks with `flock` means an interrupt at any point — including `kill -9` mid-`vipsthumbnail` — releases the lock with the process, and `mv -f` keeps thumbnail publication atomic. The `rmdir` reaps locks stranded by older releases.
- Marking rows uncacheable when a thumbnail is missing (`bin/omarchy-menu-images:239`) closes the actual poisoning hole: previously a pruned row set was cached under a valid signature and reused forever. The `v3`/`v2` signature bumps invalidate caches already poisoned in the wild, and failed images are retried on the next run.
- The new test file covers stranded locks, failed batches, retry, and concurrent serialization.

The review surfaced five remaining holes, all in the same "interrupted at the wrong moment" family. Maintainer edits are enabled, so I pushed two commits fixing them (d564e53, aa061e5):

1. **Unbounded lock wait.** `flock` with no timeout blocked forever, where the old spin-wait capped at 30s. A generator hung on a corrupt image or slow mount — e.g. a background `omarchy-theme-bg-cache` run — would wedge every subsequent picker open, since `drain_pending_thumbnails` runs in the foreground before the menu appears. Now `flock -w 30` (`bin/omarchy-menu-images:156`); on timeout the row is left uncached and retried next run.
2. **Lock fd leaked into vips.** The flock fd was inherited by `vipsthumbnail`, so killing the wrapper shell while vips lived on (orphaned or hung) left the lock held anyway — exactly the stranded-lock behavior the PR is fixing, just one process deeper. The fd is now closed for the child (`bin/omarchy-menu-images:167`); worst case after a wrapper kill is a redundant regeneration, which is safe because tmp files are per-PID and publication is an atomic rename.
3. **Partial thumbnails leaked forever.** A generator killed mid-write strands `<hash>.jpg.<pid>.jpg` in the cache dir, which nothing prunes. Only the lock holder writes those, so they are reaped right after acquiring the lock (`bin/omarchy-menu-images:158-160`), with a regression assertion in the recovery test.
4. **Legacy lock reaping raced a live legacy generator.** Unconditionally `rmdir`-ing an old-style lock directory could yank it from a still-running pre-upgrade generator, letting two generators race and the loser's failure path delete the winner's thumbnail. Reaping now requires the directory to be over two minutes old (`bin/omarchy-menu-images:148-153`) — stranded locks from a crash still recover, just never a lock that may still be owned. Covered by a new test scenario (`test/shell.d/menu-images-test.sh:75-91`).
5. **Cache publication itself wasn't crash-safe.** The rows/signature files were written with plain truncating redirects and no exclusion, so a picker killed mid-write (or two pickers interleaving) could leave truncated or mismatched rows behind signatures that still validate — reintroducing the permanent hiding through the back door. Publication now happens under a per-key `flock` via tmp-file renames, rows first, so any kill point leaves either a harmless cache miss or a consistent cache (`bin/omarchy-menu-images:252-267`).

Also hardened the concurrency test: the two racing pickers now get a longer overlap window and per-PID `wait` status checks, so a crashed run can't pass silently (`test/shell.d/menu-images-test.sh:120-133`).

One thing deliberately left alone: the `.lock` files themselves are never unlinked. Unlinking a live flock file reintroduces a lost-wakeup race, and one empty file per cached thumbnail is cheap.

Tests: `test/shell.d/menu-images-test.sh` (repeated for flake-checking), `./test/cli`, and `./test/shell` pass; the one pre-existing `./test/shell` failure (`config-test.sh`, crash-watch service packaging) reproduces on `origin/quattro` and is unrelated.

PR #6777, opened 2026-08-12
Update installed plugins from the Update menu

nille:
`omarchy plugin update` fetches, shows a diff, and fast-forwards the git checkouts under `~/.config/omarchy/plugins`, but nothing in the menu reaches it. Setup > Plugins offers Enable, Disable, Add, Clone and Remove, so the menu can install a plugin and then never update one — someone who added a plugin from the menu has no path to a newer version without knowing the CLI exists.

Update > Extra Themes already runs `omarchy-theme-update` over the user's git themes, and plugins are the same kind of thing with a more careful updater behind them: fetch, show the diff, fast-forward only, validate, roll back on failure, then rescan the shell.

```jsonc
"update.plugin": {"icon":"󰐱","label":"Plugins","when":"compgen -G \"$HOME/.config/omarchy/plugins/*/.git\"","action":"omarchy-launch-floating-terminal-with-presentation omarchy-plugin-update"},
```

### Why Update rather than Setup > Plugins

Both read fine. I put it next to Extra Themes because that row establishes where "update the extras I installed" lives, and updating is what the user is there to do — Setup > Plugins is where you change *which* plugins you have. Happy to move it under `setup.plugin.update` if you'd rather keep all five plugin verbs together.

### The guard

`when:` mirrors Setup > Plugins' Remove, but keys on `.git` instead of `manifest.json`. A plugin copied in by `plugin clone` has no remote to pull from and `omarchy-plugin-update` refuses it by design ("is not a git checkout, so there is nothing to pull from"), so the row hides rather than offering an action that cannot work. It runs through `omarchy-launch-floating-terminal-with-presentation` because the diff and the confirm prompt need a terminal.

### Tests

Two assertions in `test/shell.d/menu-test.sh`, alongside the existing `update.omarchy` ones. I checked they actually fail when the row is missing rather than just passing alongside it — with the row deleted, `menu-test.sh` fails.

`./test/all` reports 4 failures on this branch: `bar-icon-geometry`, `config`, `snapper` and `unowned-system-paths`. All 4 fail identically on unmodified `quattro` (38542a1) here, so they're pre-existing on this machine and not from this change. One further file failed intermittently across runs — `manifest-entrypoints` once, `runtime-smoke` the next — in teardown (`rm: cannot remove '/tmp/tmp.XXXX/home': Directory not empty`) rather than on an assertion. `menu-test.sh` passed in every run.

No screenshot: the change is a single menu row and I couldn't drive the menu to capture it on this setup. It renders as **Plugins** between Extra Themes and Process.


nille:
Verified in the running shell now (Hyprland 0.56.2, omarchy-dev 4.0.0.r1635.g0d45f09). I drove it with `wtype` and `omarchy capture screenshot fullscreen save` per `agents/skills/visual-verification.md`.

Because `/usr/share/omarchy/default/omarchy/omarchy-menu.jsonc` is root-owned, I exercised the row by putting the **exact bytes from this diff** into `~/.config/omarchy/extensions/omarchy-menu.jsonc`, which the shell merges through the same `parseMenuJsonc` → `mergeMenuSources` path with `watchChanges: true`. Normalized row as the shell sees it:

```json
{"id":"update.plugin","parent":"update","kind":"action","label":"Plugins","icon":"f0431",
 "when":"compgen -G \"$HOME/.config/omarchy/plugins/*/.git\"",
 "action":"omarchy-launch-floating-terminal-with-presentation omarchy-plugin-update"}
```

**Visible state.** Row renders under Update with the plugin glyph (U+F0431, same as Setup > Plugins), correct icon-column alignment, no clipping or overlap, and the submenu grows by one row without disturbing the others. It's also searchable: typing `Plugins` filters to it with the `Update` breadcrumb as subtext.

**Hidden state.** I pointed the guard at a directory with nothing to match (`plugins-none/*/.git`, `compgen` exit 1) and refreshed: the row disappears and the remaining ten rows close up with no gap or stale spacing. Restored afterwards.

**Terminal launch.** Selecting the row with `wtype -k Return` opened a floating `org.omarchy.terminal` window (confirmed via `hyprctl clients -j`, `floating: true`) running the real updater. It was not a no-op — it fast-forwarded both git-managed plugins on this machine and exited:

```
29d9f71 HEAD@{0}: merge 29d9f71...: Fast-forward
9985006 HEAD@{1}: merge 9985006...: Fast-forward
```

Both plugins still load and report `enabled: true` in `omarchy plugin list` afterwards, so the `rescanPlugins` at the end of `omarchy-plugin-update` lands fine when invoked this way.

**One thing I could not show visually, stated plainly.** The screenshots put the row *last* under Update rather than between Extra Themes and Process, because user-extension entries append during the merge — that's the harness, not the change. Modifying the root-owned default file or restarting the session's shell against this branch to photograph the real position wasn't something I wanted to do on a live machine. What I can show is that render order follows declaration order, computed with the shell's own merge code against the file **as this PR changes it**:

```
1 Omarchy  2 Channel  3 Config  4 Extra Themes  5 Plugins  6 Process
7 Hardware  8 Firmware  9 Password  10 Timezone  11 Time
```

So position 5 is the diff's own doing and is asserted by declaration order; everything else above is from the running UI. Happy to attach the stills if useful.

nille:
A note on why this PR deliberately keeps plugin updates outside the general `omarchy update` flow:

Git-managed plugins are user-installed, third-party executable code rather than Omarchy-managed system components. Updating them can therefore have a different trust and review boundary. The existing updater reflects that: it fetches changes, shows the diff, asks for confirmation, fast-forwards only, validates the result, and rolls back on failure.

Exposing that updater through **Update → Plugins** preserves this review-oriented workflow without adding per-plugin prompts to a normal system update or making `omarchy update -y` implicitly approve third-party code changes.

It also follows the existing **Update → Extra Themes** pattern: user-installed extras have an explicit, discoverable update action, while the general Omarchy update remains focused and predictable.

Automatically updating plugins could still be a valid policy decision, but I think it should be considered separately from making the existing updater discoverable. This PR intentionally solves the latter with the smaller behavioral change.

PR #6779, opened 2026-08-12
Feat: add opencode implementation to the agents plugin (with the new usage endpoint)

felixzsh:
## Summary

   Adds official support for OpenCode Go subscriptions to the agents usage
   widget, following the same architecture as the existing `claude`, `codex`,
   and `fireworks` collectors: one `omarchy-agent-usage-<agent>` collector
   prints a display-ready JSON record that the panel discovers by convention —
   no panel changes needed.

image

   ## What's new

   - **`bin/omarchy-agent-usage-opencode-go`** — new collector:
     - **Plan limits** from OpenCode's new official usage API,
       `GET https://opencode.ai/zen/go/v1/usage` (merged upstream in
       anomalyco/opencode#16513), authenticated with the Go API key from
       `~/.local/share/opencode/auth.json` (or `OPENCODE_GO_API_KEY`). It
       returns the rolling 5-hour, weekly, and monthly windows as percentages
       with reset timestamps — the same numbers the opencode.ai dashboard shows
       — so the panel's meters are authoritative account-wide.
     - **Token stats** from opencode's own SQLite database
       (`~/.local/share/opencode/opencode.db`), counting assistant messages on
       the `opencode-go` provider, split by day and model with cache and
       reasoning separated — the same database the `claude` and `codex`
       collectors already read for their own providers.
     - Accepts both the deployed response shape
       (`usage.{rolling,weekly,monthly}.{percent,resetsAt}`) and the merged PR
       shape (`{rolling,weekly,monthly}Usage.{usagePercent,resetInSec}`), and
       tolerates Cloudflare's User-Agent check on opencode.ai.
   - **Mark** — the official OpenCode "O" glyph (lobehub icon set) as
     `assets/opencode-go.svg` with a `-light` twin, following the codex
     convention.
   - **Tests** — `test/shell.d/agent-usage-opencode-scanner-test.sh`: 20
     assertions, ~1s, fully offline (fixture database + mocked endpoint).
   - **Docs** — the agents plugin README gains a collector row and an
     "OpenCode Go limits" section.

   ## Verification

   - Ran the collector against a live account: the limits probe returns real
     percentages (rolling/weekly/monthly) and the token stats come from the
     local database (359 prompts, per-model breakdown).
   - `bash test/shell.d/agent-usage-opencode-scanner-test.sh` passes (20/20).

   ## Notes

   - The Go usage endpoint is server-side and already live; no opencode client
     update is required.
   - No public API exposes usage by day or by model, so the token breakdown
     comes from the local database (same as `claude`/`codex`), while the limit
     percentages are the account-wide truth from the API.


felixzsh:
its good to go

dhh:
Reviewed. The collector's contract compliance, discovery naming, DB scan, and test isolation are all solid — and `opencode-go` as the id is exactly right, since plain `opencode` would claim Anthropic/OpenAI-through-opencode usage the other collectors already attribute. I pushed two commits fixing behaviors that diverged from the claude collector this code mirrors:

1. **Percent scale**: the per-value `> 1` heuristic made the meter discontinuous right after a window reset — 0.5% rendered as 50% and exactly 1% as a full meter. The scale is now detected payload-wide, the way the claude collector does it.
2. **Cached limits**: the probe cache was served forever on any error, including a rejected key — a dead meter with a reset time in the past and no explanation. Cached windows are now dropped once their reset has passed, and a rejected key says "Sign-in rejected" instead of being papered over.

Both are pinned by new test cases (32 assertions, all green).

felixzsh:
i think we should do something similar #6780

markbus-ai:
Oh, I had opened a PR hoping to push for the implementation of this new endpoint; your work is great.

felixzsh:
i would recommend to first merge this pr #6542

jesseburlamaque:
My question is: what's the plan for Zen?

Is Zen intended to be a separate collector/PR (e.g., omarchy-agent-usage-opencode-zen), or is the goal for this PR to eventually cover both Go and Zen under one opencode agent?

For context, I noticed [#6526](https://github.com/basecamp/omarchy/pull/6526) takes the approach of splitting them into two separate records (opencode-zen and opencode-go), while this PR started as opencode-go and was later renamed to opencode. It would be great to understand preferred direction. Happy to help test either route.

felixzsh:
> My question is: what's the plan for Zen?
> 
> Is Zen intended to be a separate collector/PR (e.g., omarchy-agent-usage-opencode-zen), or is the goal for this PR to eventually cover both Go and Zen under one opencode agent?
> 
> For context, I noticed [#6526](https://github.com/basecamp/omarchy/pull/6526) takes the approach of splitting them into two separate records (opencode-zen and opencode-go), while this PR started as opencode-go and was later renamed to opencode. It would be great to understand preferred direction. Happy to help test either route.

there are no official api endpoints for zen as far as i know, but either way zen its different, its pay as you go so that would fit in a section that doesn't exits in this agents plugin, something like credits, idk i just use opencode go subscription

bjarneo:

felixzsh:
@bjarneo mine was older and had better test coverage, but i know it was kinda outdated and opencode v2 is on the horizon, i will re-open a pr with the updated and improved implementation

PR #6780, opened 2026-08-13
perf(agents): cut codex usage collector memory with SQL filter and cache

markbus-ai:
## Problem

The `omarchy-agent-usage-codex` collector scanned **every row** of `opencode.db` with Python-side `json.loads` — on a typical install that is a 1.7 GB DB with 55k+ rows, producing a **~716 MB RSS peak + 8 s CPU** on **every** run. The `omarchy.agents` widget triggers it on shell start and every 15 minutes, and the panel's `refreshLimits()` call passes `--limits-only` — a flag the collector **parsed and silently ignored**, so opening the panel paid the full scan too.

## Change

1. **SQL-level filtering** in `scan_opencode_sessions()`: `LIKE` gates (acceleration) + `json_valid` + `json_extract` (authority) with the exact same semantics as the old Python filter (`role == "assistant"` and `providerID == "openai"`). Malformed rows are excluded via `json_valid` instead of aborting the scan.
2. **Local-stats cache** following the existing `omarchy-agent-usage-claude` pattern: `$XDG_CACHE_HOME/omarchy/agent-usage/codex-scan-<digest>.json` with atomic writes, `fcntl.flock` double-checked locking, and a `schemaVersion` envelope validated on read.
3. **Flags honored**: `--force` rescans and rewrites the cache; normal and `--limits-only` reuse a fresh (≤900 s) cache and fall back to a full scan when it is missing, stale, or corrupt. Cache-layer failures degrade to a direct scan and never drop the record.

## Measured (this machine, real 1.7 GB opencode.db)

| Scenario | Wall | Peak RSS |
|---|---|---|
| Before (baseline) | 24.0 s | **717.9 MB** |
| Cold scan (after) | 4.4 s | 158.7 MB |
| `--limits-only` warm | 1.4 s | 85.1 MB |

## Parity

The output record schema and values are **unchanged** for the same data — verified against the old Python filter including malformed-row cases (valid rows still counted, bad rows skipped, no silent abort). `--force` produces byte-identical local stats to the pre-change full scan.

## Tests

`test/shell.d/agent-usage-codex-scanner-test.sh`: 6 original + 9 new (cache write/perms/schema, `--limits-only` reuse, `--force` bypass, stale-cache rescan, corrupt/unwritable cache fallback, malformed-row parity, compact JSON). All 15 pass; claude collector suite (8) still passes.

### DIFF
--- a/bin/omarchy-agent-usage-codex
+++ b/bin/omarchy-agent-usage-codex
@@ -209,13 +209,16 @@ def scan_opencode_sessions():
   # A subscription burned entirely through opencode leaves no native session
   # files, but opencode records per-message provider, model, and token usage
   # in its own database. Read-only: opencode may be writing right now.
+  # Returns whether the scan ran to completion: a scan cut short by a
+  # database error still contributes what it read, but must not be cached
+  # as if it were the whole story.
   db = Path(os.environ.get("XDG_DATA_HOME") or (Path.home() / ".local" / "share")) / "opencode" / "opencode.db"
   if not db.is_file():
-    return
+    return True
   try:
     conn = sqlite3.connect(db.resolve().as_uri() + "?mode=ro", uri=True, timeout=2)
   except sqlite3.Error:
-    return
+    return False
   try:
     conn.execute("PRAGMA query_only = ON")
     # OpenCode DBs grow huge (every historical message, JSON included), and
@@ -270,9 +273,12 @@ def scan_opencode_sessions():
         continue
       add_usage(day, "opencode:" + str(session_id), model, input_tokens, output_tokens, cache_read, cache_write)
   except sqlite3.Error:
-    pass
+    # Transient lock, schema migration, corruption: the numbers stop here,
+    # incomplete.
+    return False
   finally:
     conn.close()
+  return True
 
 
 def scan_native_codex_sessions():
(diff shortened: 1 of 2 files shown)


dhh:
Reviewed. The SQL-side filtering and cache plumbing are well built — the filter agrees with the old Python filter on everything JS `JSON.stringify` can emit, the atomic write/flock pattern matches the claude collector exactly, and the memory win is real. I pushed six commits fixing freshness semantics that diverged from the claude collector this PR cites:

1. The 900s reuse applied to every mode, including the widget's periodic no-flag refresh — stats could lag ~30 minutes and a lowered `refreshIntervalSec` was silently ignored. Normal mode now reuses a scan for only 20s (concurrent-run dedup), with the 15-minute reuse scoped to `--limits-only`, matching claude's split.
2. The cache now records its scan date and treats any other day as a miss — a cache written at 23:58 no longer serves yesterday's "today" totals past midnight (the code claimed this guarantee without providing it).
3. A cache with a future mtime (clock moved backwards) is no longer eternally fresh.
4. An opencode scan interrupted by a database error is still served but never cached as a complete snapshot.
5. The `json_valid` guard is now a `CASE` around each `json_extract` — SQLite doesn't promise AND evaluation order, and a reordered plan would have silently truncated the scan.

All pinned by new tests (20 assertions in the scanner test, full suites green).

Issue #6787, opened 2026-08-13
`--help` is being skipped for bin routes that partially resolve to a file


EFrMG:
### System details

Omarchy 4.0.0.alpha

### What's wrong?

`--help` is silently skipped for routes that only partially resolve to a file.

For example:

```bash
omarchy update aur --help    # does NOT show help: it starts a real update
```

`omarchy-update-aur` doesn't exist as a binary (only `omarchy-update-aur-pkgs` does), so `update aur` only resolves as far as the `omarchy-update` binary, leaving `aur --help` as leftover args. The router only checked the *first* leftover token for `--help`/`-h`; since that token is `aur`, the check never fires, and both `aur` and `--help` get forwarded straight into `omarchy-update`. `omarchy-update` treats any first arg other than `-y` as "not skip confirmation" and falls through into `omarchy-update-confirm`, this being that it enters the update flow instead of printing help.

This isn't specific to `aur`, as it affects any group with a bare parent binary (`omarchy-<group>`) where the typed subcommand doesn't match a real child binary, and the trailing `--help` isn't the immediate next token after where the router actually resolved a real command.

## Fix

In `bin/omarchy`, added `remaining_has_help_flag()`, which scans the **entire** leftover argument list for `--help`/`-h` instead of only checking index 0. Both dispatch paths (`dispatch_fast_or_help` and `dispatch_or_help`) now use it before deciding whether to forward args to the resolved binary as well.

This is a targeted fix: it guarantees `--help` is never silently forwarded into a real command.

It does not change behavior for a typo without `--help`, so `omarchy update aur` alone **still** forwards to `omarchy-update` which is a separate, broader change (rejecting unrecognized subcommands outright). That would need a way to distinguish commands with a closed subcommand vocabulary (`update`, `reinstall`) from ones that intentionally take freeform positional args (`hook <name>`, `toggle <flag-name>`), so it is left out of the following PR.

## Testing

- `./test/cli`: 105/105 passing.

- Manual verification:
  - `omarchy update aur --help`: now shows `update` command help and its related commands, instead of starting a real update.
  - `omarchy update -y --help`: `--help` still wins even with a real flag ahead of it.
  - `omarchy theme list --help`: unaffected, confirming that exact-match routes still work as before.

PR #6788, opened 2026-08-13
fix `--help` being skipped for bin routes that only partially resolve to a file

EFrMG:
Issue: #6787

### DIFF
--- a/bin/omarchy
+++ b/bin/omarchy
@@ -123,12 +123,25 @@ append_pipe_value() {
 }
 
 # A route can resolve to a binary before all args are consumed (`update aur` only matches the `update` binary, leaving `aur` and `--help` as leftovers)
-# --help must never be lost among those leftovers and get forwarded into the real command, so this checks the whole remainder rather than just the first token
+# --help must never be lost among those leftovers and get forwarded into the real command, so this checks the whole remainder rather than just the first token.
+# A `--` marks everything after it as belonging to the command itself, so scanning stops there.
 remaining_has_help_flag() {
   local token=""
 
   for token in "$@"; do
+    [[ $token == "--" ]] && break
     [[ $token == "--help" || $token == "-h" ]] && return 0
+  done
+
+  return 1
+}
+
+remaining_has_json_flag() {
+  local token=""
+
+  for token in "$@"; do
+    [[ $token == "--" ]] && break
+    [[ $token == "--json" ]] && return 0
   done
 
   return 1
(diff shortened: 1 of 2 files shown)


dhh:
Reviewed this against #6787 and pushed two fixup commits to the branch.

**Verified**

- Reproduced the bug from the code: `omarchy update aur --help` resolves only as far as `omarchy-update` (prefix walk in `resolve_direct_route`), so `aur --help` was forwarded and a real update started. The same hole existed on the registered-route path — pre-fix, `omarchy screenshot bogus --help` would actually take a screenshot. The fix closes both (`bin/omarchy:958`, `bin/omarchy:1022`).
- Exercised the matrix with hermetic fake binaries: full resolution, no-arg invocations, unknown routes (still exit 127 with suggestions), `--help`/`-h` mid-args and trailing, lookalike tokens (`--helpme`, `-hx`, a quoted `"--help inside"` all still forward), and `--help --json`. All behave.
- Scanning the whole remainder is the right call here: internal scripts invoke the `omarchy-*` binaries directly, so the router is purely the human-facing surface, and a human typing `--help` anywhere on an `omarchy` line wants help. The direct binary remains the full-fidelity passthrough.

**Pushed fixups** (`maintainerCanModify` is on)

1. `2676c3ad` — two hardening tweaks to the same conditionals this PR touches:
   - Stop the help scan at `--` (`bin/omarchy:128`): everything after `--` belongs to the resolved command, so `omarchy <cmd> run -- --help` still forwards. That keeps a standard escape hatch for wrapper commands like `update lock`, `launch tui`, and `hook` that hand args to an inner program.
   - Match `--json` as an exact token via `remaining_has_json_flag` (`bin/omarchy:139`) instead of substring-matching `" ${remaining[*]} "`. The flattened-argv check meant a `--json` *inside one quoted argument* (e.g. `omarchy agent prompt 'explain --json output' --help`) flipped help output to JSON. Pre-existing, but this PR's broader help matching made it much easier to hit.
2. `ab23949c` — regression tests in `test/cli:761` using the suite's fake-binary pattern: partial resolution with trailing `--help`/`-h` renders help without executing (verified the test fails against the pre-PR router by actually running the fixture), same via an alias route (covers `dispatch_or_help`), related-commands listing, exact-argv forwarding when no help flag is present, `--` passthrough, and both `--json` cases.

`./test/cli`: 116/116 passing. The one `./test/all` failure (`test/shell.d/config-test.sh`, a PKGBUILD/systemd-service check) is unrelated and pre-existing — this diff only touches `bin/omarchy` and `test/cli`.

One known non-goal, matching the issue writeup: groups without a bare parent binary (e.g. `omarchy theme bogus --help`) still report "Unknown Omarchy command" rather than group help — nothing executes there, so it's cosmetic and fine to leave for a broader unknown-subcommand pass.

PR #6791, opened 2026-08-13
Drop the kms hook when the proprietary NVIDIA driver handles early KMS

### LINKED ISSUE #6790: NVIDIA installs: `kms` hook bundles ~100 MB of unused nouveau GSP firmware into every initramfs/UKI
## What happens

On a machine using the proprietary NVIDIA driver, Omarchy's initramfs carries the in-tree `nouveau` driver plus its GSP firmware blobs — about 100 MB of compressed payload that is never used. With the UKI setup on Limine, my image was **256 MB** (`.initrd` = 238 MB); a normal size for this configuration is 20–60 MB.

Largest cpio entries:

```
 49.36 MiB  usr/lib/firmware/nvidia/ga102/gsp/gsp-570.144.bin.zst
 25.07 MiB  usr/lib/firmware/nvidia/ga102/gsp/gsp-535.113.01.bin.zst
 13.37 MiB  usr/lib/firmware/nvidia/tu102/gsp/gsp-570.144.bin.zst
 12.57 MiB  usr/lib/firmware/nvidia/tu102/gsp/gsp-535.113.01.bin.zst
 11.47 MiB  usr/lib/modules/…/updates/dkms/nvidia.ko.zst
  1.16 MiB  usr/lib/modules/…/kernel/drivers/gpu/drm/nouveau/nouveau.ko.zst
```

## Why

Two Omarchy components each do half the NVIDIA job and nothing reconciles them:

- `omarchy-settings-dev` ships `/etc/mkinitcpio.conf.d/omarchy_hooks.conf` with `kms` in `HOOKS`.
- `install/hardware/nvidia.sh` writes `/etc/mkinitcpio.conf.d/nvidia.conf` with `MODULES+=(nvidia nvidia_modeset nvidia_uvm nvidia_drm)` and `options nvidia_drm modeset=1` — proprietary early KMS, done correctly.

The `kms` hook then adds in-tree DRM drivers anyway; `autodetect` keeps `nouveau` because it matches the NVIDIA PCI device, and mkinitcpio bundles every firmware blob in nouveau's modinfo. The [Arch wiki (NVIDIA → mkinitcpio)](https://wiki.archlinux.org/title/NVIDIA#Early_loading) says to remove `kms` from `HOOKS` in exactly this configuration.

So every Omarchy user on the proprietary driver carries ~100 MB of dead firmware in each initramfs — doubled on `/boot` by the Limine fallback history, and slowing sd-stub's TPM measurement of the UKI.

## Verified fix

A drop-in that subtracts `kms` from whatever `omarchy_hooks.conf` sets, e.g. `/etc/mkinitcpio.conf.d/zz-no-kms.conf`:

```sh
_hooks_no_kms=()
for _hook in ${HOOKS[@]}; do
    [[ $_hook == kms ]] || _hooks_no_kms+=("$_hook")
done
HOOKS=("${_hooks_no_kms[@]}")
unset _hooks_no_kms _hook
```

Result on my machine (RTX 4090, `nvidia-open-dkms` 610.57.04): UKI **256,183,296 → 144,066,048 bytes**. `nouveau.ko` and its firmware gone; the two `gsp_{ga10x,tu10x}.bin` blobs owned by `nvidia-utils` correctly retained; all four nvidia DKMS modules still present; early KMS still works via `nvidia_drm modeset=1` (`modconf` bundles the modprobe.d file). Boots fine.

A fix in Omarchy could be conditional: `nvidia.sh` already knows it's setting up the proprietary driver and could drop/filter the `kms` hook at the same time it writes `nvidia.conf`.

Possibly related prior art: #3619, where removing `kms` fixed a black LUKS prompt on Intel — a different symptom of the same hook loading drivers too early.

## System

- `omarchy version`: 4.0.0.r1689.g38542a1-1 (settings-dev channel)
- linux 7.1.8.arch1-3, nvidia-open-dkms 610.57.04-1, RTX 4090, Limine + UKI + LUKS

Filed by Claude Fable 5 via Claude Code.

matjam:
Fixes #6790.

`install/hardware/nvidia.sh` early-loads `nvidia_drm` (with `modeset=1`) for early KMS, but `HOOKS` still carries the `kms` hook, so `autodetect` pulls nouveau — and ~100 MB of its GSP firmware — into every initramfs for a driver that never runs. On a Limine UKI setup that means a 256 MB image where ~144 MB is normal, doubled again by the fallback history on `/boot`.

## Change

`omarchy_hooks.conf` now filters `kms` out of `HOOKS` when both hold:

- `nvidia_drm` is in `MODULES` — i.e. `nvidia.conf`, which sorts before this drop-in, configured proprietary early KMS;
- every PCI display-class device (`class 0x03xx` in sysfs) has NVIDIA's vendor ID.

Hybrid iGPU+dGPU systems keep `kms` so the iGPU retains early KMS at the LUKS prompt. Non-NVIDIA systems are untouched. Because the conditional lives in the settings-package config, existing installs pick it up on their next settings update + initramfs rebuild — no migration needed.

## Testing

- Sourced with `nvidia_drm` in `MODULES` against real sysfs on an RTX 4090 system: `kms` dropped.
- Sourced without `nvidia_drm`, and with `MODULES` unset under `set -u`: `HOOKS` unchanged, no errors.
- Faked-sysfs hybrid tree (AMD 0x1002 iGPU + NVIDIA dGPU): `kms` kept.
- Real-world result of the equivalent filter on the same machine (nvidia-open-dkms 610.57.04, Limine + UKI + LUKS): UKI 256,183,296 → 144,066,048 bytes; `nouveau.ko` and its firmware gone; the two `nvidia-utils` GSP blobs and all four nvidia modules retained; Plymouth still owns the LUKS prompt via `nvidia_drm`.

Filed by Claude Fable 5 via Claude Code.

### DIFF
--- a/etc/mkinitcpio.conf.d/omarchy_hooks.conf
+++ b/etc/mkinitcpio.conf.d/omarchy_hooks.conf
@@ -12,7 +12,11 @@ if [[ " ${MODULES[*]:-} " == *" nvidia_drm "* ]]; then
   _omarchy_nvidia_gpu=0
   _omarchy_other_gpu=0
   for _omarchy_pci in "${OMARCHY_PCI_DEVICES_PATH:-/sys/bus/pci/devices}"/*; do
-    [[ -r $_omarchy_pci/class && -r $_omarchy_pci/vendor ]] || continue
+    if [[ ! -r $_omarchy_pci/class || ! -r $_omarchy_pci/vendor ]]; then
+      # An unreadable device could be another GPU. Inconclusive keeps kms.
+      _omarchy_other_gpu=1
+      continue
+    fi
     [[ $(<"$_omarchy_pci/class") == "0x03"* ]] || continue
     if [[ $(<"$_omarchy_pci/vendor") == "0x10de" ]]; then
       _omarchy_nvidia_gpu=1
--- a/etc/mkinitcpio.conf.d/omarchy_hooks.conf
+++ b/etc/mkinitcpio.conf.d/omarchy_hooks.conf
@@ -1,5 +1,28 @@
 HOOKS=(base udev plymouth keyboard autodetect microcode modconf kms keymap consolefont block encrypt filesystems fsck btrfs-overlayfs)
 
+# The proprietary NVIDIA driver does early KMS itself: nvidia.conf (written by
+# install/hardware/nvidia.sh, sourced before this file) early-loads nvidia_drm
+# with modeset=1. Keeping the kms hook on such a system makes autodetect pull
+# in nouveau — and ~100 MB of its GSP firmware — for a driver that never runs.
+# Drop kms when nvidia_drm is early-loaded and every display controller is
+# NVIDIA. Hybrid systems keep kms: their iGPU still needs it for early KMS.
+if [[ " ${MODULES[*]:-} " == *" nvidia_drm "* ]]; then
+  _omarchy_non_nvidia_gpu=0
+  for _omarchy_pci in /sys/bus/pci/devices/*; do
+    [[ -r $_omarchy_pci/class && -r $_omarchy_pci/vendor ]] || continue
+    [[ $(<"$_omarchy_pci/class") == 0x03* ]] || continue
+    [[ $(<"$_omarchy_pci/vendor") == 0x10de ]] || _omarchy_non_nvidia_gpu=1
+  done
+  if ((_omarchy_non_nvidia_gpu == 0)); then
+    _omarchy_hooks=()
+    for _omarchy_hook in "${HOOKS[@]}"; do
+      [[ $_omarchy_hook == kms ]] || _omarchy_hooks+=("$_omarchy_hook")
+    done
+    HOOKS=("${_omarchy_hooks[@]}")
+  fi
+  unset _omarchy_non_nvidia_gpu _omarchy_pci _omarchy_hooks _omarchy_hook
+fi
+
 # Bundle vconsole.conf so Plymouth uses the configured keyboard layout at the
 # LUKS prompt, but only when that layout types Latin letters. Passphrases are
 # Latin characters, so bundling a Hebrew/Greek/Cyrillic/Arabic layout would
--- a/etc/mkinitcpio.conf.d/omarchy_hooks.conf
+++ b/etc/mkinitcpio.conf.d/omarchy_hooks.conf
@@ -4,23 +4,30 @@ HOOKS=(base udev plymouth keyboard autodetect microcode modconf kms keymap conso
 # install/hardware/nvidia.sh, sourced before this file) early-loads nvidia_drm
 # with modeset=1. Keeping the kms hook on such a system makes autodetect pull
 # in nouveau — and ~100 MB of its GSP firmware — for a driver that never runs.
-# Drop kms when nvidia_drm is early-loaded and every display controller is
-# NVIDIA. Hybrid systems keep kms: their iGPU still needs it for early KMS.
+# Drop kms only when nvidia_drm is early-loaded and NVIDIA owns every display
+# controller. Hybrid systems keep kms: their iGPU still needs it for early
+# KMS at the LUKS prompt. So does anything unexpected, like a PCI tree that
+# cannot be read.
 if [[ " ${MODULES[*]:-} " == *" nvidia_drm "* ]]; then
-  _omarchy_non_nvidia_gpu=0
-  for _omarchy_pci in /sys/bus/pci/devices/*; do
+  _omarchy_nvidia_gpu=0
+  _omarchy_other_gpu=0
+  for _omarchy_pci in "${OMARCHY_PCI_DEVICES_PATH:-/sys/bus/pci/devices}"/*; do
     [[ -r $_omarchy_pci/class && -r $_omarchy_pci/vendor ]] || continue
-    [[ $(<"$_omarchy_pci/class") == 0x03* ]] || continue
-    [[ $(<"$_omarchy_pci/vendor") == 0x10de ]] || _omarchy_non_nvidia_gpu=1
+    [[ $(<"$_omarchy_pci/class") == "0x03"* ]] || continue
+    if [[ $(<"$_omarchy_pci/vendor") == "0x10de" ]]; then
+      _omarchy_nvidia_gpu=1
+    else
+      _omarchy_other_gpu=1
+    fi
   done
-  if ((_omarchy_non_nvidia_gpu == 0)); then
+  if ((_omarchy_nvidia_gpu && !_omarchy_other_gpu)); then
     _omarchy_hooks=()
     for _omarchy_hook in "${HOOKS[@]}"; do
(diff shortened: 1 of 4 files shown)


matjam:
Addressed the review in b1058de5:

- Fixed string literals in the `[[ ]]` comparisons are now quoted per AGENTS.md, with the class wildcard left active (`"0x03"*`).
- Added `test/shell.d/nvidia-kms-hook-test.sh` covering nvidia-only, hybrid, no `nvidia_drm`, `MODULES` unset under `set -u`, an NVIDIA audio function with no display controller, an empty PCI tree, and a device directory missing its sysfs attributes.
- Writing the empty-tree test exposed a real gap: an empty or unreadable PCI tree counted as "no non-NVIDIA GPU" and would have dropped `kms`. The conditional now requires a positively identified NVIDIA display controller, and reads the tree through `OMARCHY_PCI_DEVICES_PATH` — the same seam `bin/omarchy-hw-nvidia` uses.

Filed by Claude Fable 5 via Claude Code.

matjam:
The suppressed review comments were both right, addressed in 94ed33e1:

- An unreadable PCI device next to a readable NVIDIA GPU previously still dropped `kms` — the skip treated "can't read it" as "not a GPU". Unreadable entries now count as inconclusive, so `kms` stays, and the test covers the mixed case.
- The test sourced the host's real `/etc/vconsole.conf` under `set -u`, where a valid KEYMAP-only file breaks the `XKBLAYOUT` expansion and makes the result machine-dependent. `XKBLAYOUT` and `FILES` are now predefined before sourcing, so the test is hermetic.

Filed by Claude Fable 5 via Claude Code.

dhh:
Reviewed this against the codebase, pushed two fixup commits (e66ccc70, 5d5cc060), and ran an adversarial `codex` review over the full diff.

**What I checked and found sound**

- **Driver detection.** Gating on `nvidia_drm` in `MODULES` is the right signal: `install/hardware/nvidia.sh` writes `/etc/mkinitcpio.conf.d/nvidia.conf` for both `nvidia-open-dkms` and `nvidia-580xx-dkms`, and never for nouveau. `nvidia.conf` sorts before `omarchy_hooks.conf`, so the modules are visible when the conditional runs.
- **Hybrid graphics.** The PCI scan (`etc/mkinitcpio.conf.d/omarchy_hooks.conf:20`) requires a positively identified NVIDIA display controller and no non-NVIDIA one, using class `0x03*` — which covers both VGA (`0x0300`) iGPUs and 3D-controller (`0x0302`) dGPUs — so an Intel/AMD iGPU keeps `kms` and its early KMS at the LUKS prompt. Empty/unreadable trees fall back to keeping `kms`. Re-evaluated on every rebuild, so a GPU swap self-heals.
- **HOOKS manipulation.** Array filtering with exact element comparison, no sed on the HOOKS line — no word-boundary or substring hazards, and a duplicate `kms` would be removed cleanly.
- **Fresh installs.** `nvidia.conf` lands during `omarchy-apply-hardware`, before any subsequent initramfs build reads the drop-ins.
- **No conflicts** with `omarchy_resume.conf` (`HOOKS+=(resume)`, sorts later, appends) or `thunderbolt_module.conf`.

**What was missing: existing installs never rebuild**

`omarchy_hooks.conf` ships as a `backup=()` file in omarchy-settings, and nothing regenerates the initramfs when only a mkinitcpio drop-in changes — so every existing NVIDIA install would keep the ~100 MB image (doubled in the Limine fallback history) until its next kernel update. The repo's precedent for exactly this situation is `migrations/1784476564.sh`, which rebuilt after changing this same file. Pushed:

- `migrations/1786605598.sh` — rebuilds via `limine-mkinitcpio` once per machine (marker in `/var/lib/omarchy/migrations/`, same pattern as `migrations/1786482992.sh`), and only where sourcing the *installed* drop-ins in mkinitcpio's order shows `kms` actually dropped out — so hybrid machines, non-NVIDIA machines, and user-edited configs (packaged update parked as `.pacnew`) skip it.
- `test/shell.d/nvidia-kms-migration-test.sh` — covers rebuild + marker, cross-user idempotency, interrupted-rebuild retry, hybrid skip, pre-conditional config skip, and missing `limine-mkinitcpio`/`nvidia.conf` skips.

**Codex review (xhigh) findings and disposition**

1. *Later drop-in resets MODULES:* `install/hardware/fix-surface-keyboard.sh:20` writes `surface_device_modules.conf` with `MODULES=(...)` (assignment, not `+=`), which sorts after this file and would strip `nvidia_drm` after `kms` was already dropped. Unreachable today — every machine Omarchy writes that file for is hybrid Intel, so the PCI scan keeps `kms` — but it is a real invariant, so I documented it in the conf (5d5cc060) rather than restructure the packaging. (That `MODULES=` clobber is arguably a pre-existing bug worth its own fix: it erases nvidia early-loading on such machines regardless of this PR.)
2. *`.pacnew` merged later gets no rebuild:* at migration time the active config still carries `kms`, so the built image matches its config; whoever hand-merges a mkinitcpio `.pacnew` rebuilds per standard practice, and the next kernel update covers it regardless. Left as is.
3. *Marker race between two concurrent users:* same accepted check/build/mark pattern as `migrations/1786482992.sh`; worst case is a redundant rebuild. Left as is.

`./test/shell`: all pass except the pre-existing, unrelated `config-test.sh` failure (sibling omarchy-pkgs PKGBUILD coverage for `omarchy-crash-watch.service`, untouched by this PR).

dhh:
Please test the updated implementation 🙏

dhh:
Tested the updated implementation at 5d5cc060 (includes matjam's b1058de5 + 94ed33e1 and the two fixup commits).

**Test runs**

- `test/shell.d/nvidia-kms-hook-test.sh` — 9/9 pass, including the new inconclusive cases (empty tree, unreadable device alone, unreadable device beside a readable NVIDIA GPU).
- `test/shell.d/nvidia-kms-migration-test.sh` — 6/6 pass; the migration sources the installed conf, so it inherits the inconclusive logic with no gating change needed.
- Full `./test/shell` — all 167 test files pass.

**Direct simulation** of the conditional against fabricated sysfs trees, all behaving as intended:

| layout | kms |
|---|---|
| NVIDIA-only (VGA 0x0300) | dropped |
| dual NVIDIA GPUs | dropped |
| NVIDIA GPU + readable non-display device | dropped |
| Intel iGPU + NVIDIA 3D (0x0302) | kept |
| AMD iGPU + NVIDIA 3D | kept |
| empty PCI tree | kept |
| NVIDIA audio function only | kept |
| unreadable device beside NVIDIA GPU | kept |

The third row is the important regression check on 94ed33e1: the inconclusive path only fires when `class`/`vendor` are actually unreadable — which enumerated sysfs devices never are — so a real NVIDIA-only machine full of readable chipset devices still gets the optimization, while anything genuinely unverifiable fails safe to keeping `kms`. A focused codex (xhigh) pass over 94ed33e1 and its composition with the migration returned no findings.

**Verdict: sound.** Nothing further to fix from my side.

PR #6801, opened 2026-08-13
Fix notification focus for agent terminals

atomashevic:
## What changed

- Match every Hyprland client by current class first, then `initialClass`.
- Fall back to `initialTitle` only for clients identified as Omarchy agent terminals.
- Keep the existing regex and case-insensitive matching behavior.
- Add regression coverage for a `kitty` notification whose window uses Omarchy's shared `org.omarchy.agent` class while an unrelated Viber window is present.

## Why

Omarchy agent terminals deliberately share the window class `org.omarchy.agent`, but terminal notifications still identify their sender as `kitty`, `foot`, and so on. The focus helper searched only the current class, so clicking an agent notification found no target. After the popup disappeared, Hyprland's focus-follows-mouse behavior could focus the unrelated window underneath it, making that application appear to hijack notifications.

This was reproduced with Codex notifications and Viber and was also present on earlier Omarchy versions, so it is a persistent identity-resolution bug rather than a recent regression.

### Viber example

The concrete failure looked like Viber had registered itself as the handler for every desktop notification, but Viber was not involved in notification routing:

1. Codex finished a task and its terminal sent a notification with `app_name=kitty`.
2. The Codex terminal window was launched by Omarchy with `class=org.omarchy.agent`; only its `initialTitle` still said `kitty`.
3. Clicking the notification made Omarchy run `omarchy-hyprland-focus-app kitty`.
4. The helper searched only the current `class`, found no `kitty` window, and therefore focused nothing.
5. Omarchy dismissed the popup. The pointer was now over the Viber window behind that popup, so Hyprland's `follow_mouse=1` behavior focused Viber.

The result was repeatable: clicking a Codex notification brought Viber forward, making Viber look like the notification action target. MIME defaults, the notification service owner, and the notification action were all correct; Viber was simply the window exposed beneath the popup after Omarchy failed to resolve the real sender window.

Hyprland retains the terminal identity in `initialTitle`, so using that field specifically for `org.omarchy.agent` clients gives the notification fallback a valid target without treating arbitrary window titles as application identities.

## Validation

- `mise exec node@26.5.0 -- ./test/all` — all 171 shell test files passed, along with the CLI suite.
- Regression test: an unrelated Chromium window initially titled `Mail settings` is not selected for an app named `Mail`.
- Live Hyprland check: `./bin/omarchy-hyprland-focus-app kitty` focused the exact `org.omarchy.agent` client address.
- Manual Codex-style notification test: clicking an `app_name=kitty` notification returned focus to the Codex agent window and did not focus Viber.

### DIFF
--- a/bin/omarchy-hyprland-focus-app
+++ b/bin/omarchy-hyprland-focus-app
@@ -21,7 +21,11 @@ address=$(
       'def matches($value): ($value // "") | test($pattern; "i");
        ([.[] | select(matches(.class))] +
         [.[] | select(matches(.initialClass))] +
-        [.[] | select(matches(.initialTitle))]) |
+        [.[] | select(
+          ((.class // "") == "org.omarchy.agent" or
+           (.initialClass // "") == "org.omarchy.agent") and
+          matches(.initialTitle)
+        )]) |
        first.address // empty'
 )
 
--- a/bin/omarchy-hyprland-focus-app
+++ b/bin/omarchy-hyprland-focus-app
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# omarchy:summary=Focus a Hyprland window by application class
+# omarchy:summary=Focus a Hyprland window by application identity
 # omarchy:args=<app-name>
 # omarchy:examples=omarchy hyprland focus app Slack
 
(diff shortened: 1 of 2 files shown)


dhh:
Pushed f0cf4a3c simplifying the jq query. The three-array concatenation built the full client list three times and carried an `initialClass` tier that nothing needed — agent windows always have `initialClass=org.omarchy.agent` (set via `xdg-terminal-exec --app-id`), so a generic initialClass match never fires for the bug this fixes, and no caller depends on it. jq's lazy `first(a, b)` expresses the same ordered fallback directly:

```jq
def matches($value): ($value // "") | test($pattern; "i");
first(
  (.[] | select(matches(.class))),
  (.[] | select(.initialClass == "org.omarchy.agent" and matches(.initialTitle)))
).address // empty
```

Class matches still win over the agent-title fallback, so existing callers (`omarchy-launch-browser`'s anchored regexes, raw notification app names) behave identically. All three regression tests pass unchanged, as does the full shell suite.

PR #6803, opened 2026-08-13
Add a Grok usage collector for the agents panel

jason-watts:
## What

Adds `omarchy-agent-usage-grok` so the existing `omarchy.agents` bar widget can show Grok usage the same way it already shows Claude, Codex, and Fireworks. The panel does not need a QML change: it already draws any record that lands in `~/.local/state/omarchy/agents/usage/`.

## How

- **Local stats** come from `~/.grok/sessions` (or `GROK_HOME`). Each `turn_completed` event in `updates.jsonl` is one prompt; token buckets are summed per turn, not taken as a running session total.
- **Limits / plan** come from the same cli-chat-proxy endpoints the CLI's `/usage` command uses (`/v1/billing` and `/v1/settings`). The collector only reads the OIDC token already stored in `~/.grok/auth.json`. It never writes credentials and does not refresh them — that stays with `grok login`.
- A **zero monthly prepaid allowance** (typical SuperGrok / X Premium+ login) is treated as "no meter", not 0% used. Those tabs still show the plan label and local token charts.
- Prepaid monthly credit, an on-demand cap, and a live prepaid balance are reported when the billing payload actually has them.

## Tests

`./test/shell.d/agent-usage-grok-scanner-test.sh` covers empty/unauthenticated records, per-turn summing, idle sessions, monthly/on-demand/balance mapping, expired tokens, and `--limits-only` cache reuse.

Verified on a live X Premium+ login: the panel gained a Grok tab with today's token totals and the `X Premium+` plan line, and no invented monthly meter.


jason-watts:
Agent should have checked for duplicates before writing this. Closig.

PR #6813, opened 2026-08-13
Stop screensavers through their supervisors

grantwiley:
## Problem

Follow-up to #6764 and #6773.

Waiting for `ttfx` after `pkill` still leaves a race because `omarchy-screensaver` is a supervisor loop. A live supervisor can start another renderer after the global kill but before the terminal closes. `pidwait` only waits for processes that already exist, so the replacement can still lose its PTY and abort with the same EIO panic.

The dismissal path has the same underlying problem: every monitor's supervisor globally kills every `ttfx` and every screensaver terminal. This can race across monitors and also affects unrelated `ttfx` processes.

## Solution

Make each screensaver supervisor own its renderer lifecycle:

- coordinate multi-monitor dismissal by signalling supervisors rather than renderers or terminals
- track the exact `ttfx` child and prevent respawning once shutdown starts
- terminate and reap that child before allowing the terminal to close
- force-stop a renderer if it does not exit within the grace period
- prevent a renderer from starting during a concurrent lock
- explicitly disable Ghostty's hold-after-command behavior

`omarchy-system-lock` now signals and briefly waits for the supervisors. It no longer kills `ttfx` or closes its PTY directly.

## Tests

- `bash test/shell.d/screensaver-test.sh`
- `bash test/shell.d/system-lock-test.sh`
- `bash -n bin/omarchy-screensaver bin/omarchy-system-lock test/shell.d/screensaver-test.sh test/shell.d/system-lock-test.sh`

The lifecycle test covers two monitors, delayed renderer shutdown, respawn prevention, unrelated `ttfx` processes, forced termination, and a concurrent lock. It also passed ten consecutive runs.

### DIFF
--- a/bin/omarchy-screensaver
+++ b/bin/omarchy-screensaver
@@ -1,26 +1,39 @@
 #!/bin/bash
 
-# omarchy:summary=Run the Omarchy screensaver using random effects from TTE.
+# omarchy:summary=Run the Omarchy screensaver using random ttfx effects
+
+renderer_pid=
+shutdown=0
 
 screensaver_in_focus() {
   hyprctl activewindow -j | jq -e '.class == "org.omarchy.screensaver"' >/dev/null 2>&1
 }
 
-exit_screensaver() {
-  hyprctl eval 'hl.config({ cursor = { invisible = false } })' &>/dev/null || hyprctl keyword cursor:invisible false &>/dev/null || true
-  pkill -x ttfx 2>/dev/null
-  pkill -f '[o]rg.omarchy.screensaver' 2>/dev/null
-  exit 0
+request_shutdown() {
+  (( shutdown )) && return
+
+  shutdown=1
+  trap '' SIGINT SIGTERM SIGHUP SIGQUIT
+
+  # One dismissed monitor dismisses them all. Each supervisor still owns and
+  # reaps only its own renderer, so every terminal keeps its PTY alive through
+  # ttfx's shutdown.
+  pkill -x omarchy-saver 2>/dev/null || true
 }
 
 # Exit the screensaver on signals and input from keyboard and mouse
-trap exit_screensaver SIGINT SIGTERM SIGHUP SIGQUIT
+trap request_shutdown SIGINT SIGTERM SIGHUP SIGQUIT
+
+# Give the supervisor a stable, unambiguous process name for coordinated exit.
+# This runs after installing traps and before checking the lock state so a lock
+# can never miss a supervisor that is about to start ttfx.
+printf 'omarchy-saver' >/proc/self/comm
+
+[[ $(omarchy-shell lock isLocked 2>/dev/null) == "true" ]] && exit 0
 
 printf '\033]11;rgb:00/00/00\007'  # Set background color to black
 
 hyprctl eval 'hl.config({ cursor = { invisible = true } })' &>/dev/null || hyprctl keyword cursor:invisible true &>/dev/null
-
-tty=$(tty 2>/dev/null)
 
 # Terminals allocate the pty at the default 80x24 and only resize it once the
 # compositor has told the window how big it is. ttfx measures the terminal once,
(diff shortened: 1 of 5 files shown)


grantwiley:
The newer reports on #6762 line up with the races this PR is intended to close.

[This report](https://github.com/basecamp/omarchy/issues/6762#issuecomment-5285361217) shows that the bounded wait from #6773 wasn't sufficient on its own: `ttfx` could still crash even with `pidwait` in the lock script.

[The follow-up](https://github.com/basecamp/omarchy/issues/6762#issuecomment-5285985464) found the two remaining races. In 10 of the 11 cores inspected, `ttfx` received SIGTERM but lost its PTY before it finished shutting down. In the other case, the screensaver loop spawned another `ttfx` before the terminal closed.

Those are the two cases I tried to address here. The supervisor marks itself as shutting down before stopping its renderer, so it can't spawn a replacement. It then terminates and reaps its own renderer before the terminal exits and the PTY disappears. Locking and normal dismissal both go through that same shutdown path.

The reports are from rc3 with #6773, so they don't directly verify this branch. But they do independently confirm both failure modes behind this change: the common shutdown race and the less frequent respawn race. The lifecycle test covers both, along with multiple monitors, delayed shutdown, forced termination, and unrelated `ttfx` processes.

CodexOmega:
Live reproduction of the race this PR targets, on current code.

omarchy 4.0.0-1, ttfx 0.3.1-1, foot 1.27.0-2. Two ttfx SIGABRT core dumps
today, 2026-08-14 19:43:42 and 20:56:12 CDT, both on lock transitions within
~0.5s of the screensaver launching:

  thread 'main' panicked at library/std/src/io/stdio.rs:1166:9:
  failed printing to stdout: Input/output error (os error 5)

(Panic text recovered from the core via `strings`; ttfx is stripped and not
on debuginfod.) Both instances land on `lock-system: lock-timeout` /
`lock secure=true` while the renderer is mid-write — the supervisor-respawn /
concurrent-lock case described here. Confirms #6764/#6773 didn't close the
window the supervisor loop reopens. Happy to attach full coredumpctl info if
useful.

— big-pickle via opencode

garymeyerca:
Tested commit `d1b691a06` on `omarchy-dev 4.0.0.r1744.gf002044-1`, `ttfx 0.3.1-1`, using Foot. The focused tests and full `./test/all` suite passed. I also ran one live screensaver → lock → unlock cycle and 20 live input-dismissal cycles using the PR scripts directly. No new `ttfx` core dumps occurred, and no renderer processes or screensaver windows remained afterward.

Tested by OpenAI Codex via Codex.

jacobkjones:
Live dismissal-path abort on current 4.0.0-1, not a lock transition.

#6762 said the dismissal path was clean (`exit_screensaver` kills `ttfx` before the terminal). This dump is the opposite: two `ttfx` processes aborted at the same second the idle service logged `screensaver-dismissed` and ran `omarchy-system-wake`. No lock ran.

```
22:45:07.390  idle → omarchy-launch-screensaver
22:45:07.541  idle-monitor: active (cycle remained armed)
22:45:07.768  launcher exited 0 (foot windows already mapped)
22:45:09.858  idle-cycle-cancel: screensaver-dismissed
22:45:09.858  process-start: wake omarchy-system-wake
22:45:09      ttfx 3280492 and 3280489 SIGABRT
```

Same panic recovered from both cores (binary is stripped, not on debuginfod; offsets match the rest of the thread: `+0xcc30a`, `+0xa48a9`, `+0xccd09`, …):

```
thread 'main' (3280492) panicked at library/std/src/io/stdio.rs:1166:9:
```

Adjacent on the heap: `io error: Input/output error (os error 5)`. Built with `panic_abort`; `si_code` is `SI_TKILL`.

Four monitors, default terminal foot. Only two of the four renderers dumped core, which fits the race this PR describes: each supervisor's `pkill -x ttfx` / `pkill -f '[o]rg.omarchy.screensaver'` is global, so a pty can vanish under a renderer that is still writing at 120 fps.

`omarchy-system-wake` only restores brightness; it does not kill processes. The windows went away through the screensaver's own exit path.

Omarchy 4.0.0-1, ttfx 0.3.1-1, foot 1.27.0-2, Hyprland 0.56.2-1, Linux 7.1.8-arch1-3, x86_64. Happy to attach `coredumpctl info` if useful.

PR #6826, opened 2026-08-13
Recover updates from retired package conflicts

### LINKED ISSUE #6818: Upstream qemu update blocks omarchy update
### System details

AMD 7700, RTX4070 Ti Super, Omarchy Quattro RC3

### What's wrong?

When running `omarchy update`, it fails with the following error due to an upstream qemu update:

`qemu-common-11.1.0-1 and qemu-block-gluster-11.0.3-1 are in conflict. Remove qemu-block-gluster? [y/N]`

Since `omarchy update` is non-interactive, this stops the update from proceeding.

I was able to resolve it manually with pacman interactively but this breaks updating for people not as familar with using Arch directly. Could this be solved with a custom migration file that runs before the `pacman -Syu --noconfirm` built into `omarchy update`?

yashranaway:
Keep upstream package splits from wedging unattended updates.

<<< AI wording below >>>

## Problem

`pacman -Syu --noconfirm` answers No to package-conflict removal prompts. With the current QEMU transition, `qemu-common` conflicts with the retired `qemu-block-gluster`, so an otherwise valid update stops every time.

## Fix

- Retry with Pacman's narrowly scoped conflict answer only when every removal target was installed as a dependency and is absent from the configured sync repositories
- Leave explicitly installed, still-published, and mixed safe/unsafe conflicts for a human
- Keep the existing one-retry boundary so a second failure cannot loop

Fixes #6818.

## Verification

- `bash test/shell.d/update-file-conflict-test.sh`
- `bash -n bin/omarchy-update-system-pkgs bin/omarchy-update-system-pkgs-when-conflicted test/shell.d/update-file-conflict-test.sh`
- `git diff --check`

### DIFF
--- a/bin/omarchy-update-system-pkgs
+++ b/bin/omarchy-update-system-pkgs
@@ -10,21 +10,29 @@ echo -e "\e[32m\nUpdate system packages\e[0m"
 errors=$(mktemp)
 trap 'rm -f "$errors"' EXIT
 
+pacman_options=(-Syu --noconfirm --overwrite '/usr/share/omarchy/*')
+if [[ ${OMARCHY_UPDATE_CONFLICT:-} == 1 &&
+  ${OMARCHY_UPDATE_RETRY:-} == 1 &&
+  ${OMARCHY_UPDATE_REMOVE_CONFLICTS:-} == 1 ]]; then
+  # Pacman's conflict-removal question is bit 4. This is set only by the
+  # conflict handler after it proves every removal is a retired dependency.
+  pacman_options+=(--ask 4)
+fi
+
 # /usr/share/omarchy is wholly Omarchy's; files can land there unowned and would
 # otherwise abort the upgrade. Packaged content wins there.
 #
 # Progress bars stay on stdout. Errors are on stderr, kept for the conflict
 # handler below; LC_ALL=C is what keeps them parseable in any locale.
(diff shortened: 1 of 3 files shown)


dhh:
Closing in favour of #6830, which takes the other route out of #6818: hand the conflict question back to the person running the update rather than prove an answer safe.

Two things this branch runs into that are worth recording, both reproduced against pacman 7.1 with real packages in a fakeroot:

1. Under `--noconfirm`, pacman prints its questions to **stdout**, not stderr. `omarchy-update-system-pkgs` captures only `2>"$errors"`, so the `:: … Remove …? [y/N]` line never reaches the report the handler parses — `conflict_removal_count` is 0 and the new block never fires. Only the two `error:` lines are on stderr.

2. The real prompt ends `[y/N] ` with a trailing space, so the `$`-anchored lookahead `(?=\? \[y/N\]$)` cannot match it even when the line is present.

The tests pass because `write_raw_report` hand-writes the fixture, so neither the stream nor the trailing space is modelled.

The vetting itself was sound, for what it's worth — `-Qd` plus `-Qm`, all-or-nothing, and the three-variable guard — and `--ask 4` does what it claims: it inverts the default for only the masked question type, leaving `Replace X with Y? [Y/n]` answered Yes as before. Thank you for digging into this one.

PR #6834, opened 2026-08-13
Ignore T2 headset remotes when switching layouts

### LINKED ISSUE #6833: Keyboard layout bar widget can switch the wrong device on T2 Macs ("Apple Headset" phantom keyboard not excluded)
On T2 Macs, the T2 coprocessor (`apple-bce`/`bce-vhci`) exposes the headphone jack's inline remote as a fake USB HID keyboard (`Bus=0003 Vendor=05ac Product=8103`, name `"Apple Headset"`). `KeyboardLayoutModel.js`'s `UNTYPED_KEYBOARDS` regex already excludes other non-typing devices (`hl-virtual-keyboard`, `power-button`, `sleep-button`, `lid-switch`, `video-bus`) but not this one, so the bar widget's `selectKeyboard()` treats it as a real, typeable keyboard.

`selectKeyboard()` picks whichever "typed" keyboard is furthest advanced through the layout list. If `Apple Headset` happens to be further along than the real keyboard, it gets selected, and clicking the widget runs `hyprctl switchxkblayout "Apple Headset" next` — switching a device nobody types on, while the real keyboard (and therefore every app reading from it, including terminals) stays on the old layout. The bar label then shows the new layout even though nothing you actually type reflects it.

**Repro (T2 Mac only):**
1. Set a multi-layout `kb_layout` (e.g. `us,gr`) with `grp:alts_toggle`
2. Click the keyboard-layout bar widget
3. Bar label updates, but typed input in terminals/apps reading the real keyboard doesn't change
4. Using the actual Alt+Alt shortcut works correctly, confirming the real keyboard's layout never moved — only `Apple Headset`'s did

**Suggested fix:** add `Apple Headset` (or a broader pattern matching `apple-bce`-enumerated non-typing HID devices) to `UNTYPED_KEYBOARDS` in `shell/plugins/bar/widgets/KeyboardLayoutModel.js`.

Confirmed via `hyprctl devices -j` showing split state (`apple-headset -> Greek`, real keyboard -> `English (US)`) and cross-referenced against `/proc/bus/input/devices`:

```
I: Bus=0003 Vendor=05ac Product=8103
N: Name="Apple Headset"
S: Sysfs=.../0000:01:00.1/apple-bce/apple-bce/bce-vhci/usb5/...
```

Environment: Omarchy `4.0.0rc3-1` (Quattro RC), T2 MacBook.

yashranaway:
Keep the keyboard-layout widget attached to a keyboard people can type on.

<<< AI wording below >>>

## Problem

T2 Macs expose the headphone jack's inline remote as `apple-headset`, a keyboard-shaped HID device. Hyprland gives it the seat's layout list, so when it is further through that list than the real keyboard, the widget selects and switches the headset instead. The label moves while typed input does not.

## Fix

Classify `apple-headset` alongside the other non-typing devices already excluded from keyboard-layout selection.

The regression test models the reported T2 seat directly: the headset is on Greek, the real `apple-spi-keyboard` is on English, and filtering must leave the real keyboard as the switch target.

Fixes #6833.

## Verification

- `bash test/shell.d/keyboard-layout-test.sh`
- `bash -n test/shell.d/keyboard-layout-test.sh`
- `git diff --check`


twfyke:
Confirmed on a live MacBookPro16,3. `/proc/bus/input/devices` still has `Name="Apple Headset"` at `05ac:8103`, and it is a `kbd` handler. This is the right exclusion.

I did not re-run the multi-layout click repro here (this machine is on a single layout), so this is a hardware confirm, not a second functional test.

yashranaway:
Thanks for confirming the device identity on real hardware; it matches the exclusion and test fixture here.

PR #6837, opened 2026-08-13
Wait on the profile being repaired, not on every browser

shrijit37:
The migration asks the user to close every running browser before repairing the Copy URL shortcut, but a browser only ever rewrites its own Preferences on exit. Waiting on all of them deadlocks `omarchy update` for anyone whose main browser is effectively never closed. That's what happened in #6821's very first run on the MacBook that filed the report: the ghosts the migration found lived in a stale `~/.config/chromium` profile nobody has open, yet the migration blocked on the always-on Chrome and the update sat at the "close all browser windows" prompt on every run.

A running Chromium-family browser holds a `SingletonLock` (and socket) inside its user-data-dir, so whether the profile being repaired is open is mechanical — no process sniffing needed. The migration now gates on the pending (or to-be-verified) profiles being open:

- Affected profile closed → repair proceeds; browsers on unrelated profiles no longer hold the update hostage.
- Affected profile open → same ask-to-close prompt as today, and the existing post-repair verification still catches a browser that starts mid-repair and restores stale Preferences on exit.

`profile_open` also treats a `SingletonLock` left as a dangling symlink as open (Chrome writes it that way), so the "is it open" answer errs toward asking rather than repairing over a live browser.

Test rewritten to simulate an open profile with its `SingletonLock` instead of stubbing `pgrep`; every prior scenario — defer, ask-then-repair, mid-repair browser start, unverified-backup rerun — still passes.

### DIFF
--- a/migrations/1786643346.sh
+++ b/migrations/1786643346.sh
@@ -122,10 +122,36 @@ unverified_repairs_exist() {
   return 1
 }
 
-# Only this user's browsers matter — another user's cannot rewrite these
-# Preferences.
-browsers_running() {
-  pgrep -x -u "$UID" 'chromium|chrome|brave|msedge|vivaldi-bin|vivaldi|opera|helium' >/dev/null 2>&1
+# A browser only rewrites its own Preferences on exit, so a repair can only be
+# reverted by a browser attached to a profile this migration has to touch.
+# Whether a profile is open is mechanical: a running Chromium-family browser
+# holds a SingletonLock (and socket) inside its user-data-dir.
+profile_open() {
+  # -L catches a SingletonLock left as a dangling symlink; -e covers a plain
+  # file, and -S the socket — any of them means a browser is attached.
+  [[ -L $1/SingletonLock || -e $1/SingletonLock || -S $1/SingletonSocket ]]
+}
+
+# Gate on a pending — or to-be-verified — profile actually being open, not on
+# browsers in general. Waiting on every browser process deadlocks the update on
+# machines whose main browser is effectively never closed: the pending ghosts
+# commonly sit in a stale profile nobody has open, yet the update blocks on the
+# always-open one.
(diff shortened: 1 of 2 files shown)


dhh:
Rebased this onto quattro and pushed to the branch, since #6842 landed in the meantime and conflicted with it.

That PR removed `2>/dev/null` from the `gum confirm` call — suppressing stderr left gum waiting for a keypress behind an unpainted screen, which is what "stuck on migration 1786643346" in #6841 actually was. This branch still carried the redirect, so the resolution keeps the unsuppressed call and its explanation while taking your `affected_profile_open` gate, and keeps both sides' test hunks.

Three fixes on top, two of them caught by the bots on this PR:

- Both `python3` stubs matched `${3:-} == repair`, but the migration calls `python3 -c <script> <preferences> <pinned_id> repair`, so the mode is `$5` and neither hook ever fired. The stubs also dropped python's exit status, which the `check` calls rely on — that made both mid-repair scenarios pass for an unrelated reason. They now key off `$5` and `exit $status`. Removing the post-repair `affected_profile_open` guard from the migration now fails the mid-repair test, as it should.
- The central regression this PR exists for wasn't covered: a pending profile whose browser is closed while another profile root holds a lock. Added, and it fails if `profile_open` is stubbed to always report open.
- `open_browser` used `touch`, so only the `-e` branch was exercised. It now creates the dangling `SingletonLock` symlink Chromium actually writes, which fails if `-L` is dropped from `profile_open`.

`./test/shell` is green — 13 scenarios in this file, 176 files overall.

Worth knowing about the approach, though it isn't a blocker: `SingletonLock` lives at the profile root, not in the individual profile directory, so a browser open on `chromium/Profile 1` still blocks a repair of `chromium/Default`. That's the safe direction to be wrong in, and it's still a large improvement over blocking on every browser process.

PR #6845, opened 2026-08-14
Fix battery status on multi-battery machines

### LINKED ISSUE #6895: Power panel shows 0W discharging when power_now read returns ENODEV
## Summary

The battery widget / power panel shows **0W discharging** on a laptop whose ACPI embedded controller exposes a `power_now` sysfs attribute that exists but cannot actually be read. `omarchy-battery-status` trusts the attribute because `[[ -r ]]` passes, gets an empty string from the read, and awk coerces it to `0W` — discarding the correct value that UPower already computed.

## Expected behavior

The widget should show the real power draw (e.g. `16.8W`, `29.9W`), matching UPower's `energy-rate`.

## Actual behavior

`omarchy-battery-status --shell` reports `rate 0W` while discharging.

## Steps to reproduce

On affected hardware (here: HP Pavilion Gaming Laptop 15-dk0xxx, battery `BAT1`):

```bash
# The sysfs attribute exists and is world-readable, so [[ -r ]] passes...
ls -l /sys/class/power_supply/BAT1/power_now   # -r--r--r-- 1 root root ...

# ...but reading it fails:
head -c 20 /sys/class/power_supply/BAT1/power_now
# head: error reading '.../power_now': No such device

# Meanwhile UPower has the correct value:
upower -i /org/freedesktop/UPower/devices/battery_BAT1 | awk '/energy-rate/ { print $2 }'
# 29.9735

# And the widget script zeroes it out:
omarchy-battery-status --shell
# percentage  40%
# state       discharging
# rate        0W        <-- should be ~16-30W
```

## Root cause

`bin/omarchy-battery-status` lines ~50-57:

```bash
if [[ -r $battery_path/power_now ]]; then
  power_rate_raw=$(awk -v microwatts="$(<"$battery_path/power_now")" 'BEGIN { print microwatts / 1000000 }')
```

`[[ -r ]]` only checks that the file exists with read permission, not that the read succeeds. On this HP/ACPI firmware the read() syscall returns `ENODEV` ("No such device"), so `$(<...)` captures an empty string, `awk` computes `"" / 1000000 = 0`, and the correct UPower `energy-rate` value is overwritten. The same latent issue affects the `current_now`/`voltage_now` fallback.

The fix is to only trust the sysfs value when the read returns a non-empty numeric string, falling back to UPower's `energy-rate` otherwise.

## Notes

- PR #6845 ("Fix battery status on multi-battery machines") rewrites this exact block and, as part of its metering logic, already discards invalid/empty `power_now` reads (`[[ $microwatts =~ ^-?[0-9]+$ ]] || continue`) — which fixes this bug as a side effect. It would be good to add a regression test there for the ENODEV-on-read case (attribute present but read fails).
- A local user-side workaround is to clone the power plugin and guard the sysfs read; happy to share.

## System details

- Omarchy: 4.0.0-1 (package `omarchy`)
- Kernel: 7.1.8-arch1-3 (x86_64)
- Machine: HP Pavilion Gaming Laptop 15-dk0xxx (Mobo 85FA, BIOS Insyde F.67)
- Battery: Hewlett-Packard PABAS0241231, BAT1, Li-ion
- Desktop: Hyprland 0.56.2 (Wayland), SDDM
- CPU: Intel Core i5-9300H; GPU: Intel UHD 630 + NVIDIA GTX 1050 Max-Q
- Full diagnostics in `/tmp/omarchy-debug.log`; screenshots of the 0W panel are available and can be attached on request.

### LINKED ISSUE #7916: Power panel reports dead BAT0 (0% / 0W / Holding) on dual-battery ThinkPads
### System details

- Omarchy 4.0.0-1
- Lenovo ThinkPad T460s (`20FN002KUS`)
- Dual battery: internal BAT0 (Sony 45N1111) + slice BAT1 (Sanyo 45N1775)

### What's wrong?

The power widget reports **0% / 0W / Holding at 75–80%** while the machine is actually charging the working pack at **99% / 2.2W / ~3 minutes to full**.

This is the dual-battery `head -n 1` bug (#6885, #7214, #7897), plus a second failure that DisplayDevice-based fixes do not cover: BAT0 is *present* but electrically dead (0 V, 0 Wh, `pending-charge`). UPower still counts its `energy-full` in the composite device, so DisplayDevice says ~49% and ~10 hours to full.

### Evidence

```
$ upower -e | grep BAT
/org/freedesktop/UPower/devices/battery_BAT1

BAT0: present, 0 V, 0 Wh, 0%, pending-charge, energy-full 20.65 Wh
BAT1: charging, 12.3 V, 20.4 Wh, 99%, 2.2 W, ~3 min to full, energy-full 20.94 Wh
DisplayDevice: 49%, charging, 2.2 W, ~10 hours to full, energy-full 41.59 Wh

$ omarchy-battery-status --shell   # packaged
percentage  0%
state       holding
rate        0W
size        20Wh
```

Root cause in `bin/omarchy-battery-status`:

```bash
battery=$(upower -e 2>/dev/null | grep BAT | head -n 1)
```

Always BAT0. `pending-charge` on AC is treated as a charge-threshold hold. The panel icon uses `UPower.displayDevice` (49%), so the glyph and the numbers disagree, and both are wrong relative to the live pack.

Related PRs (#7138, #6845, #7186) switch to DisplayDevice or sum every present `BAT*`. On this hardware that still reports **49% / 41 Wh / 10 hours**, because the dead cell's `energy-full` stays in the denominator.

### What was expected

Follow packs that actually hold a charge (non-zero voltage/energy/current). Here that is BAT1: 99%, charging, 2.2W, 20Wh, 3m to full.

A healthy empty cell still shows voltage, so a real 0% pack is not dropped.

### Steps to reproduce

1. Dual-battery ThinkPad whose internal pack is present but at 0 V / 0 Wh.
2. Open the power panel from the bar.
3. Compare with `upower -i` on BAT1 vs DisplayDevice vs `omarchy-battery-status --shell`.

geirolafs:
On a dual-battery machine (Power Bridge ThinkPads and the like) whose first pack is dead, the power panel pins at 0% / 0Wh with a bogus cycle count while the working battery charges fine. `omarchy-battery-status` reads everything from the first battery UPower enumerates (`upower -e | grep BAT | head -n 1`) and the first `BAT*/cycle_count` glob match, so whichever pack sorts first wins regardless of whether it holds a charge.

Found on a ThinkPad T480 (dead internal BAT0, working removable BAT1): the panel showed `0%`, `0Wh`, cycles `-1` while the bar icon — which reads UPower's aggregate `DisplayDevice` — was correct the whole time.

**The fix:**

- Headline figures (percentage, state, rate, capacity, time) come from UPower's aggregate `DisplayDevice`, the same source the bar icon uses, so panel and bar now always agree — with one battery, several, or a dead pack in the mix.
- Live sysfs wattage and cycle counts are summed over the packs that actually hold a charge (non-zero full-charge capacity), so a dead or disconnected pack no longer masks a working one. If every pack reads dead, they're still reported rather than printing nothing.
- Charge-threshold reporting is deliberately unchanged: it still comes from the first battery's own UPower info (the aggregate carries no threshold lines), with the same sysfs fallback.

**Notes for review:**

- On a multi-battery machine the `cycles` value becomes e.g. `5 / 7` (one count per live pack). The panel's key/value parser renders it as-is; single-battery machines still get a plain number.
- Single-battery output is unchanged — verified byte-for-byte against the packaged script on real hardware, both `--shell` and plain modes.
- New regression test `test/shell.d/battery-status-dual-battery-test.sh` reproduces the T480 scenario (dead pack enumerated first). The existing test's fixture gained a `type` file, since batteries are now identified by sysfs `type` rather than a `BAT*` name glob.

**Verified on hardware:**

- ThinkPad T480 — dead pack enumerated first, 0 Wh (mine; the original report).
- ThinkPad T460s — dead pack still advertising its old `energy-full` — verified by @hilather (#7916); their numbers are the cbe0705 regression test.
- HP Spectre x360 14-ea0 — `power_now` present but failing with ENODEV, an HP/ACPI firmware class also reported on two other HP machines — verified by @hegge25 (#7182), rate matching `upower` to within 0.03W.

Fixes #6895. Fixes #7916.

### DIFF
--- a/bin/omarchy-battery-status
+++ b/bin/omarchy-battery-status
@@ -50,16 +50,42 @@ power_rate_raw=$(awk '/energy-rate/ { print $2; exit }' <<<"$battery_info")
 
 # Per-pack readings (live wattage, cycle counts) cover every system battery.
 # Peripheral batteries (mice, keyboards) also report type Battery, but carry
-# scope Device and are not part of the machine's power budget.
+# scope Device and are not part of the machine's power budget. A pack showing
+# no voltage, stored energy, or current flow is electrically dead — a healthy
+# empty cell still reports voltage. energy_full is no sign of life: a dead
+# pack keeps advertising the capacity it once had (#7916).
 batteries=()
+dead_packs=()
 for supply in "$power_supply_path"/*; do
   [[ -r $supply/type && $(<"$supply/type") == "Battery" ]] || continue
   [[ -r $supply/scope && $(<"$supply/scope") == "Device" ]] && continue
   if [[ -r $supply/present ]]; then
     [[ $(<"$supply/present") == "1" ]] || continue
   fi
-  batteries+=("$supply")
+
+  live=false
+  for attr in voltage_now energy_now charge_now current_now power_now; do
+    value=$(cat "$supply/$attr" 2>/dev/null) || continue
+    [[ $value =~ ^-?[0-9]+$ ]] || continue
+    if (( value != 0 )); then
+      live=true
+      break
+    fi
+  done
+
+  if [[ $live == "true" ]]; then
+    batteries+=("$supply")
+  else
+    dead_packs+=("$supply")
+  fi
 done
+
+# No telemetry at all is not proof of death: if every pack reads dead, keep
+# them all rather than reporting nothing.
+if (( ${#batteries[@]} == 0 )); then
+  batteries=("${dead_packs[@]}")
+  dead_packs=()
+fi
 
 # UPower's energy-rate can lag the kernel telemetry by tens of seconds. Sum
 # the instantaneous sysfs readings so the open panel stays live and a
@@ -99,6 +125,47 @@ power_rate=$(awk -v rate="${power_rate_raw:-0}" 'BEGIN {
   print rounded
 }')
 state=$(awk '/state/ { print $2; exit }' <<<"$battery_info")
+
+# UPower's composite DisplayDevice still counts a dead pack's advertised
+# energy-full, sagging the percentage and stretching time-to-full toward a
+# pack that holds nothing (#7916). When a dead pack was excluded, rebuild the
+# headline figures from the live packs' own energy readings; firmware that
+# only exposes charge counters keeps the aggregate values.
+if (( ${#dead_packs[@]} > 0 )); then
+  live_energy=0
+  live_energy_full=0
+  for supply in "${batteries[@]}"; do
+    energy=$(cat "$supply/energy_now" 2>/dev/null)
+    full=$(cat "$supply/energy_full" 2>/dev/null)
+    if [[ ! $energy =~ ^[0-9]+$ || ! $full =~ ^[0-9]+$ ]]; then
+      live_energy_full=0
+      break
+    fi
+    live_energy=$((live_energy + energy))
+    live_energy_full=$((live_energy_full + full))
+  done
+
+  if (( live_energy_full > 0 )); then
+    percentage=$((100 * live_energy / live_energy_full))
+    capacity=$((live_energy_full / 1000000))
+    time_remaining=$(awk -v energy="$live_energy" -v full="$live_energy_full" \
+      -v rate="${power_rate_raw:-0}" -v state="$state" 'BEGIN {
+      if (rate <= 0) exit
+      if (state == "charging") hours = (full - energy) / 1000000 / rate
+      else if (state == "discharging") hours = energy / 1000000 / rate
+      else exit
+      whole = int(hours)
+      minutes = int((hours - whole) * 60)
+      if (whole == 0) {
+        printf "%dm", minutes
+      } else if (minutes > 0) {
+        printf "%dh %dm", whole, minutes
+      } else {
+        printf "%dh", whole
+      }
+    }')
+  fi
+fi
 
 # Charge thresholds are per-device values the aggregate does not carry, so
 # they still come from the first battery's own UPower info.
--- a/bin/omarchy-battery-status
+++ b/bin/omarchy-battery-status
@@ -48,31 +48,25 @@ time_remaining=$(awk '/time to (empty|full)/ {
 }' <<<"$battery_info")
 power_rate_raw=$(awk '/energy-rate/ { print $2; exit }' <<<"$battery_info")
 
-# Per-pack readings (live wattage, cycle counts) cover every battery that
-# actually holds a charge. A pack with zero full-charge capacity is dead or
-# disconnected and must not mask a working one; if every pack reads that way
-# we fall back to all of them rather than reporting nothing.
-present_batteries=()
-live_batteries=()
+# Per-pack readings (live wattage, cycle counts) cover every system battery.
+# Peripheral batteries (mice, keyboards) also report type Battery, but carry
+# scope Device and are not part of the machine's power budget.
+batteries=()
 for supply in "$power_supply_path"/*; do
   [[ -r $supply/type && $(<"$supply/type") == "Battery" ]] || continue
+  [[ -r $supply/scope && $(<"$supply/scope") == "Device" ]] && continue
   if [[ -r $supply/present ]]; then
     [[ $(<"$supply/present") == "1" ]] || continue
   fi
-  present_batteries+=("$supply")
-
-  full=$(cat "$supply/energy_full" 2>/dev/null || cat "$supply/charge_full" 2>/dev/null)
-  [[ $full =~ ^[0-9]+$ ]] && (( full > 0 )) && live_batteries+=("$supply")
+  batteries+=("$supply")
 done
-
-(( ${#live_batteries[@]} )) || live_batteries=("${present_batteries[@]}")
 
 # UPower's energy-rate can lag the kernel telemetry by tens of seconds. Sum
 # the instantaneous sysfs readings when available so the open panel stays
 # live and a multi-battery machine shows its combined draw.
 total_microwatts=0
 have_sysfs_rate=false
-for supply in "${live_batteries[@]}"; do
+for supply in "${batteries[@]}"; do
   if [[ -r $supply/power_now ]]; then
     microwatts=$(<"$supply/power_now")
   elif [[ -r $supply/current_now && -r $supply/voltage_now ]]; then
@@ -148,10 +142,11 @@ if [[ $shell_output == "true" ]]; then
   printf 'size\t%s\n' "${capacity}Wh"
   printf 'time\t%s\n' "$time_remaining"
 
-  # One cycle count per live pack, so a multi-battery machine reports both
-  # rather than silently showing whichever sorts first.
+  # One cycle count per pack, so a multi-battery machine reports both rather
+  # than silently showing whichever sorts first. A dead pack's -1 falls to
+  # the digit check.
   cycles=""
-  for supply in "${live_batteries[@]}"; do
+  for supply in "${batteries[@]}"; do
     count=$(cat "$supply/cycle_count" 2>/dev/null) || continue
     [[ $count =~ ^[0-9]+$ ]] || continue
     cycles+="${cycles:+ / }$count"
--- a/bin/omarchy-battery-status
+++ b/bin/omarchy-battery-status
@@ -21,7 +21,12 @@ esac
 battery=$(upower -e 2>/dev/null | grep BAT | head -n 1)
 [[ -z $battery ]] && exit 0
 
-battery_info=$(upower -i "$battery")
+# Headline figures come from UPower's aggregate DisplayDevice rather than the
+# first battery UPower enumerates: on a multi-battery machine (dual-battery
+# ThinkPads and the like) the first device can be a dead or empty pack, which
+# pins the panel at 0% forever while the other battery charges fine.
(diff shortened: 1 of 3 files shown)


dhh:
Pushed two commits on top:

- **Treat every system battery as one pool** — the live/dead pack split and its fallback array bought nothing: a dead pack contributes 0 W to the sum and its `-1` cycle count already fails the digit check. Replaced with one loop that also excludes `scope=Device` supplies, keeping wireless mice out of the machine's numbers (the old `BAT*` glob did that by accident).
- **Fall back to UPower's aggregate rate when a pack has no power telemetry** — summing only the packs that report would understate the machine's draw.

Both scenarios are locked in the dual-battery test; full shell suite green.

geirolafs:
Ah, much simpler 😄 Thanks!

ujo4eva:
This PR also fixes a related single-battery bug I just filed as #6895: on HP/ACPI laptops the `power_now` attribute exists and passes `[[ -r ]]`, but read() fails with `ENODEV`, so `$(<power_now)` is empty and the widget shows `0W` while UPower's `energy-rate` is correct (~16-30W). Your new metering logic (`[[ $microwatts =~ ^-?[0-9]+$ ]] || continue`) already discards the bad read and falls back to UPower's value, so this fixes it as a side effect.

Could you add a regression test for the ENODEV-on-read case (attribute present, read returns empty/invalid)? The existing fixture only covers the success path. Repro and root cause are in #6895.

geirolafs:
Added the test. while writing it I found the current_now/voltage_now fallback had the same hole: awk coerces the empty read to 0, which passes the digit check and gets counted as a metered 0W. fixed that too and locked both empty-read cases in the dual-battery test.

geirolafs:
Status: no conflicts against quattro, shell suite green. Two additions since the simplification pass:

- 3c40745 discards telemetry reads that fail with ENODEV instead of coercing them to 0W (the same failure #7182 hits on an HP Spectre).
- cbe0705 covers the case #7916 found on a T460s: a dead pack that still advertises its old `energy-full` drags UPower's composite DisplayDevice down to ~49% while the live pack sits near full. Packs are now split by electrical liveness (voltage, stored energy, or current — not `energy_full`, which a dead pack keeps reporting), and the headline figures are rebuilt from the live packs whenever a dead one is excluded. The T460s numbers from #7916 are a regression test.

One known gap, out of scope here: the panel icon fill reads `UPower.displayDevice` directly in QML, so on a #7916-shaped machine the glyph still sags even with this merged. #7917's Panel.qml changes cover that side and would compose on top of this.

hegge25:
Verified 3c40745 against the firmware it's meant to handle — nobody had run this branch on a machine that actually fails the read, so here's a data point from one.

HP Spectre x360 14-ea0, single pack, `BAT0`. `power_now` is present and world-readable and fails the read with `ENODEV`; there is no `current_now` at all, so the fallback branch can't rescue it either:

```console
$ cat /sys/class/power_supply/BAT0/power_now
cat: /sys/class/power_supply/BAT0/power_now: No such device
$ ls /sys/class/power_supply/BAT0/current_now
ls: cannot access '/sys/class/power_supply/BAT0/current_now': No such file or directory
```

Same moment, packaged script vs this branch at cbe0705:

| | `rate` |
| --- | --- |
| packaged `omarchy-battery-status` (4.0.0-1) | `0W` |
| this branch | `8.3W` |
| `upower -i` `energy-rate` | `8.27621 W` |

The single-pack path works out as intended: `voltage_now` reads non-zero so the pack counts as live, the `power_now` read fails the digit check and `continue`s, `metered_packs` stays 0, and `power_rate_raw` keeps UPower's rate rather than a metered 0W. Percentage, capacity, time and cycles are unchanged from the packaged script.

Two other affected machines reported the identical failure on #7182 and #6895 — an HP Spectre x360 14-ef2xxx (`BAT1`) and an HP Pavilion 15-dk0xxx (`BAT1`) — so this looks like an ACPI firmware class rather than one board.

I'd opened #7182 for the single-battery case before this commit existed; since this covers it, I'm closing that in favour of this PR so there's only one to review.

geirolafs:
@hegge25 Thank you for running the branch on the firmware it was actually written for, `3c40745` was built against #7182's report, not a machine I own, so this rate table and the code-path confirmation are exactly the verification it was missing. And I appreciate you closing #7182 in favour of this one; that makes for a much easier review.

geirolafs:
Summary for review, since the thread has grown long:

- **Mergeable**: no conflicts against `quattro`, shell suite green.
- **Verified on hardware covering all three failure modes**:
  - ThinkPad T480 (mine) — dead pack enumerated first at 0 Wh, the original bug: panel pinned at 0% / 0Wh / cycles -1 while the bar was correct.
  - ThinkPad T460s (@hilather, #7916) — dead pack still advertising its old `energy-full`, dragging UPower's composite percentage to ~49%; covered by cbe0705, with the T460s numbers as the regression test.
  - HP Spectre x360 (@hegge25, #7182 / #6895) — `power_now` present but failing with ENODEV; covered by 3c40745 and verified above against `upower` (branch reads 8.3W where the packaged script shows 0W). Two more HP machines report identical firmware behaviour, so this is a class, not one board.
- **One PR left to review**: all three overlapping PRs (#7182, #7295, #7917) were closed by their authors in favour of this one.

FPGAwesome:
Tested this on a ThinkPad T480 with BAT0 + BAT1, running Omarchy 4.0.2-1.

Before this PR, BAT0 was at 77% and holding while BAT1 was at 13% and charging. UPower’s DisplayDevice correctly showed the combined state as 42.77%, 20.39/47.67 Wh, charging at 16.774 W. The battery popup was only showing BAT0: 77%, holding, 0 W, and 22 Wh.

After applying the PR, I tested again with BAT0 at 77% and BAT1 at 28%. DisplayDevice showed 50.79%, 24.21/47.67 Wh, charging at 17.084 W. The popup showed 50%, 47 Wh, charging at 17.1 W, 1h 23m to full, and both cycle counts  (50 / 122).

So this fixes the multi-battery aggregation issue on my T480. The focused battery tests also passed before and after the live test. I tested while charging, but didn’t get a discharging sample. PASS.

landsman:
Tested [cbe0705](https://github.com/omacom/omarchy/pull/6845/changes/cbe070522d17f1314a23a3a8f72b853ee5b4bfe8) on a ThinkPad T480 (20L6S0CE41) with both packs, internal BAT0 (22 Wh) + external BAT1 (57 Wh), running Omarchy 4.0.3-1 on kernel 7.2.3-arch1-3. **Works as expected**.

Before, the power panel showed only BAT0: 100%, fully-charged, 0W, 22Wh, cycles 42, while BAT1 was charging at 90%.
After, it matches UPower's DisplayDevice: 99%, charging, 3.1W, 79Wh, 5m to full, cycles 42 / 79. Tested while charging only.

image

landsman:
Tested cbe0705 on a ThinkPad T480 (20L6S0CE41) with both packs, internal BAT0 (22 Wh) + external BAT1 (57 Wh), running Omarchy 4.0.3-1 on kernel 7.2.3-arch1-3. Works as expected.

Before, the power panel showed only BAT0: 100%, fully-charged, 0W, 22Wh, cycles 42, while BAT1 was charging at 90%.
After, it matches UPower's DisplayDevice: 99%, charging, 3.1W, 79Wh, 5m to full, cycles 42 / 79. Tested while charging only.

I've cherry-picked this change locally until it ships: the script from this branch sits in a clone of the power plugin (`omarchy plugin clone omarchy.power`), with the panel's `batteryProc` pointed at it.

bjarneo:

ujo4eva:
@bjarneo — fair call that #6845 had scope creep by the end, it started as one dual-battery fix (dead-first T480 pinning at 0%) and picked up two more failure modes along the way, so I understand why the duplication check flagged it against #10392.

That said, #10392 isn't a full duplicate, it's a subset — it only switches percentage/state to `DisplayDevice` when `pack_count > 1`. It doesn't cover:

* #6895 (my original issue): single-pack HP/ACPI firmware where `power_now` exists + `[[ -r ]]` passes but `read()` fails with `ENODEV` → empty string coerced by `awk` to `0W` instead of UPower's `energy-rate`. Fixed in #6845 by `3c40745` with numeric validation + fallback, verified on HP Spectre x360 14-ea0 (branch `8.3W` vs packaged `0W`). #10392 keeps the old single-pack sysfs path verbatim, so that bug remains.
* #7916: dead pack still advertising old `energy-full` dragging DisplayDevice to ~49% / 10h-to-full. Fixed in #6845 by `cbe0705` (liveness split + rebuild from live packs).
* summed sysfs wattage, per-pack cycles (`42 / 79`), `scope=Device` exclusion for peripherals.

Happy to help un-mess this: I can open a dedicated minimal PR just for the #6895 ENODEV guard (single-battery, tiny diff, easy to review) and leave the multi-battery aggregation to #10392 / a rebased follow-up — or go whatever way you prefer. Just let me know which split you'd like.

bjarneo:
Hey. Yes. That would be nice. Trying to untangle all the PRs, so some might be not 100% correctly closed!

ujo4eva:
@bjarneo — split it out as requested: #12649 covers just the #6895 ENODEV guard (single-battery, numeric-read check + UPower fallback, with regression tests for both empty-read cases). Minimal diff, tests green. Leaving multi-battery aggregation to #10392.

PR #6847, opened 2026-08-14
Preserve shared boot entries through factory reset

yashranaway:
Factory reset currently replaces the shared ESP's entire `limine.conf` with the Omarchy template. On a dual-boot or multi-boot machine, that silently removes Windows and other Linux menu entries and then deletes every foreign machine-ID directory it found.

## Fix

- Carry the exact pre-reset machine ID into the staged root.
- Remove only that identity with `limine-entry-tool --remove-entry`, leaving unrelated config blocks and ESP directories untouched.
- After rebuilding, resolve the fresh machine-ID block to a named Limine path and make it the default, so foreign entry ordering cannot stop automatic boot.
- Re-enroll `limine.conf` after changing the default for Secure Boot compatibility.
- Verify hashes only inside the rebuilt machine's entry; a stale foreign UKI no longer blocks handoff.
- Keep the snapshot's provisioning worker paired with its frozen helpers. Snapshots predating the shared-ESP-safe runtime are rejected instead of mixing incompatible generations.
- Keep the old identity marker until provisioning and LUKS re-keying complete, making retries idempotent.

## Verification

- Shared-ESP fixture containing old Omarchy, foreign Linux, and Windows entries.
- Reproduces the VM-observed Windows → foreign Linux → rebuilt Omarchy ordering and verifies the named default plus config re-enrollment.
- Verifies only the retired Omarchy entry and directory are removed.
- Verifies malformed machine IDs cannot reach either Limine or filesystem cleanup.
- Verifies hash checks and provisioning runtime stay scoped to the reset's ownership boundary.
- `bash test/shell.d/factory-reset-boot-entries-test.sh`
- `./bin/omarchy commands --check`
- `bash -n` on all changed shell scripts.


dhh:
Ran this branch through a full VM acceptance test: ISO built from this PR's source, unattended install, ESP fixtured as a real dual-boot machine (a `/Windows` EFI entry with payload, plus a second Linux cloned from the machine's own entry under a foreign machine-id with its own boot directory and UKIs), then a real `omarchy-system-factory-reset` driven end to end including the first-boot provisioning that follows.

The core claims hold. Through both the staged reset and first boot: the Windows entry and payload, the foreign Linux entry, its machine-id directory, and its UKIs all survive; the old Omarchy identity (entry + directory) is removed and never returns; the fresh identity owns a boot entry; provisioning completes.

One real regression surfaced, on exactly the machines this PR protects:

**The machine parks at the Limine menu after the reset.** The shipped config pins `default_entry: 2`, which assumes the Omarchy group leads the file (flattened: 1 = group, 2 = its kernel entry). `--remove-entry` plus `limine-update` re-appends the new Omarchy block *after* the foreign entries, so index 2 now lands on another OS's group — and Limine cannot auto-boot a group, so it waits at the menu indefinitely. Observed post-reset order: `/Windows`(1), foreign group(2 ← default), foreign kernel(3), `/+Omarchy`(4), kernel(5), `/EFI fallback`(6). Single-OS machines are unaffected (removal empties the entry list, so the re-added block leads again), and the old template-copy reset kept the index valid by construction — the reset needs to re-point `default_entry` at the rebuilt Omarchy entry.

Two smaller observations from the same runs:

- `verify_limine_hashes` now verifies entries the reset no longer owns. In an earlier run where the foreign entry referenced a stale `/EFI/Linux/*.efi` hash, the reset aborted with `limine.conf hash ... does not match the rebuilt file`. Blocking handover on another OS's breakage is new behavior; scoping verification to this machine's entries would avoid it.
- `stage_provisioning_runtime` binds the live `omarchy-provision-owner` to the snapshot's frozen helpers (`omarchy-apply-system`, `omarchy-provision-user`, …). It works today, but drift between the live worker and an old snapshot's helpers is a standing hazard for first boot.

The harness is `bin/omarchy-iso-test-factory-reset` on the `factory-reset-vm-test` branch of the omarchy-iso repo; rerunning it against an updated branch takes one command with a reused base image.

yashranaway:
Fixed all three in `5dfde75d`: stable Omarchy default, scoped hash checks, and snapshot-consistent provisioning. @dhh

ThePeteJames:
This PR fixes the same class of bug (`limine.conf` unconditionally overwritten, dropping foreign boot entries) for the factory-reset path. Filed #7867, which shows the identical failure mode still exists in `omarchy-refresh-limine` (`bin/omarchy-refresh-limine`, not touched by this PR's diff) — it `mv`s away `limine.conf`, replaces it with a bare theming-only template, and never re-runs `limine-scan`, so a Windows (or other foreign OS) entry is silently dropped on every refresh. Might be worth applying the same preserve-entries approach here once this lands.

Filed by Claude (Sonnet 5) via Claude Code.

yashranaway:
Thanks, #7867 should stay separate so this PR remains scoped to the factory-reset path.

PR #6863, opened 2026-08-14
Add smooth face unlock UX to Omarchy

mateuszkowalczyk:
## Summary

The goal is a complete face authentication experience across Omarchy: a smooth lock screen, easy backend setup, and matching support in Polkit and sudo prompts. The lock implementation already provides the main UX benefits: password input never blocks, and face detection starts automatically after opening the lid or interacting with the keyboard or mouse.

I divided work three parts. Lock shell support is implemented so far:

- [x] Add face authentication to the lock shell.
- [ ] Add Omarchy setup and removal commands for a face backend.
- [ ] Add a similar face flow to the Polkit shell.

## UX

The implementation mirrors fingerprint behavior where it makes sense: a separate PAM flow and indicator, with the same unlock path, while adding some face unlock specific features.

- **Automatic activation:** Face auth starts after opening the lid, keyboard, click, or touch activity, or when pointer movement wakes a blanked screen.
- **No immediate re-unlock:** Waiting for fresh activity avoids running the camera or unlocking again right after the user locks the session.
- **Non-blocking password:** Password input always stays available, and face failures do not affect password attempts.
- **Controlled camera use:** Each activation gets three quick tries, then the camera stops until fresh activity. This handles brief recognition misses without leaving it running forever.
- **Laptop-aware behavior:** Closing the lid stops face auth; reopening it starts a fresh attempt.
- **Fingerprint coexistence:** Both methods can be enabled without breaking the password field layout.

## Backend contract

The shell has a single integration point with the face backend: a working `/etc/pam.d/omarchy-lock-face` PAM service. It watches for that file, shows the face indicator when it exists, and starts authentication through it. For example, the Facelock setup used during development was:

```pam
#%PAM-1.0
auth       required                    pam_facelock.so
```

The backend owns camera setup, enrollment, recognition settings, and its own config files.

## Verification

Manually tested with Facelock and the PAM service above:

- Successful face unlock and password fallback.
- Three failed scans stopping face auth, followed by a fresh-input restart.
- Keyboard, click, touch, and pointer-wake activation.
- Lid close/open and suspend/resume behavior.

Automated coverage was also added for PAM integration, retries, activity handling, lid handling, indicator layout, and runtime status.

## Next steps

The main backend candidates are [Facelock](https://github.com/tyvsmith/facelock) and [Howdy](https://github.com/boltgolt/howdy). Facelock looks like the best starting point because it already has a nice setup and enrollment flow. Omarchy could wrap that flow and adjust its PAM setup to create the dedicated `omarchy-lock-face` service expected here (or if authors decide to integrate Omarchy specific PAM setup into Facelock itself we can just run setup command with specific parameter and let Facelock handle it). After that, the same face UX can be added to the Polkit shell.

Also I only tested it on my Asus Zenbook S14 UX5406AA so far, so I'd appreciate anyone willing to try it on diifferent machines.


bjarneo:

PR #6899, opened 2026-08-14
Add a Hermes usage collector to the agents panel

Salt-555:
## Summary

Adds a `omarchy-agent-usage-hermes` collector so the agents panel shows Hermes Agent usage alongside Claude, Codex, and Fireworks.

## What it does

- **New collector** `bin/omarchy-agent-usage-hermes`: sums the Hermes session store (`~/.hermes/state.db`, honored via `HERMES_HOME`) into the shared record contract, matching how Hermes itself counts — root sessions with at least one message that are not archived, attributed to the day each session started. Reasoning tokens count as output; each API call is one prompt.
- **No limits endpoint**: Hermes runs on bring-your-own providers (Nous Portal, OpenRouter, OpenAI, Anthropic, ...), so there is no account to probe. The record carries an empty `limits` list and local stats only.
- **No panel changes**: the panel discovers the record by convention, resolves the caduceus mark from `assets/hermes.svg`, and draws the tab.
- **Docs**: manifest description, collector table, and a "Hermes" section in the plugin README.
- **Test**: `agent-usage-hermes-scanner-test.sh` pins the record shape (reasoning folded into output, cache split, provider-prefix stripping, prompt counting), the child/archived/empty session exclusions, day attribution, the local-stats cache, and `--force`/`--limits-only` behavior.

## Validation

`test/shell.d/agent-usage-hermes-scanner-test.sh` passes (12 assertions), as do the existing `agent-usage-update-test.sh`, `agents-panel-test.sh`, and `agents-default-migration-test.sh`.


Salt-555:
Closing this in favor of #6645 by @okurmustafa, which already covers local Hermes usage attribution (and more — per-model and auxiliary-task attribution) with its own collector and scanner test. No value in two PRs for the same surface. Thanks for the great work on this.

PR #6912, opened 2026-08-15
Fix FIDO2 setup on keys that require user verification

Erijl:
`omarchy-setup-security-fido2` cannot produce a working configuration on any
authenticator that enforces CTAP 2.1 `alwaysUv`. Setup reports success, and then
every `sudo` falls through to the password prompt.

The authenticator (a Token2 FIDO2 key, but this applies to any key with
`alwaysUv` set, including a YubiKey configured that way):

```
$ fido2-token -I /dev/hidraw3
options: rk, up, noplat, alwaysUv, credMgmt, authnrCfg, clientPin,
         largeBlobs, pinUvAuthToken, setMinPINLength, nomakeCredUvNotRqd
```

`alwaysUv` means the key refuses any assertion that does not carry user
verification, e.g. a valid PIN. But `pamu2fcfg` with no flags records a
presence-only credential:

```
$ cut -d, -f4 /etc/fido2/fido2
+presence
```

So `pam_u2f` asks for an assertion with verification off, the key rejects it
with `FIDO_ERR_PIN_REQUIRED` before it even lights up, and because the rule is
`sufficient` the stack continues to `system-auth` and prompts for the password.
The key never blinks; the "touch" message is `cue`, printed before the attempt.

## The fix

Read the authenticator's CTAP options and pass `-V` (built-in verification) or
`-N` (PIN) to `pamu2fcfg` accordingly — **only for a key that enforces
`alwaysUv`**. A key that does not is registered exactly as it is today, PIN or
no PIN. What gets recorded is a requirement the authenticator states, not a
policy this change adds on top of one, so nothing that works today behaves
differently.

| Authenticator | Flags |
|---|---|
| `alwaysUv`, PIN set | `-N` |
| `alwaysUv`, built-in verification configured | `-V` |
| `alwaysUv`, neither | refused, pointing at `fido2-token -S` |
| PIN set, no `alwaysUv` | none — unchanged |
| Built-in verification, no `alwaysUv` | none — unchanged |
| U2F-only, no options line | none — unchanged |

Recorded per credential rather than by adding `pinverification=1` to the module
line, because `pam_u2f` ORs the module option with the credential's own flag:

```c
/* util.c, parse_opts() */
if (cfg->userpresence == 1 || strstr(attr, "+presence")) opts->up = FIDO_OPT_TRUE;
else if (cfg->userpresence == 0)                         opts->up = FIDO_OPT_FALSE;
else                                                     opts->up = FIDO_OPT_OMIT;
```

A module-wide setting would force a PIN onto every key in the mapping file,
including one that has none. Per-credential flags let a PIN-mandatory key and a
touch-only key coexist for the same user.

The probe runs before the first `sudo`, so a key that cannot be registered never
costs a password prompt.

## Rebased onto #7904

This PR originally carried a second commit installing the mapping file
`root:root`, because `pamu2fcfg >/tmp/fido2` followed by `sudo mv` preserved the
invoking user's ownership. #7904 has since fixed that and considerably more
besides — the predictable `/tmp` path, symlinks at both `/etc/fido2` and the
authfile, a non-regular path at either, and a migration for existing installs.
That commit is dropped as superseded; the branch is rebased onto #7904 and now
carries only the `alwaysUv` fix.

## Testing

`test/shell.d/security-fido2-test.sh` gains the authenticator matrix, driven end
to end through the real script rather than as a unit test: the `fido2-token` stub
answers `-I` as each class of key, the `pamu2fcfg` stub records its argv, and
each case asserts the exact flags and that a credential was published. It also
covers both refusal paths — an `alwaysUv` key with no verification configured,
and a key whose capabilities cannot be read — including that each happens before
anything escalates and that the second explains itself rather than dying
silently through `set -e`.

Every new assertion was checked against a mutation that defeats it: the
`alwaysUv` gate removed, the `no`-prefix boundary lost so `nouv` reads as `uv`,
`-V` and `-N` swapped, the flags not passed through to `pamu2fcfg`, the refusal
moved after `sudo install -d`, and the capability guard reverted.

`./test/all` matches a clean `quattro` checkout run from the same directory.

End to end on the real key, against a scratch authfile: `fido2-token -I` reports
`alwaysUv` and `clientPin`, `pamu2fcfg -N` takes the PIN and the touch, and the
credential lands as `+presence+pin` — where the unpatched script writes
`+presence` from the same key and `sudo` then falls through to the password.

### DIFF
--- a/bin/omarchy-setup-security-fido2
+++ b/bin/omarchy-setup-security-fido2
@@ -118,7 +118,11 @@ if [[ ! -f $authfile ]]; then
   # closed pipe as soon as a second key is attached, and pipefail reports that
   # SIGPIPE as a failure that set -e acts on.
   device=${tokens%%:*}
-  device_info=$(fido2-token -I "$device")
+
+  if ! device_info=$(fido2-token -I "$device"); then
+    echo -e "\e[31m\nCouldn't read the FIDO2 key's capabilities. Replug it and try again.\e[0m"
+    exit 1
+  fi
 
   # Refuse a key that cannot be registered before asking for a password, not
   # after.
--- a/bin/omarchy-setup-security-fido2
+++ b/bin/omarchy-setup-security-fido2
@@ -16,6 +16,44 @@ check_fido2_hardware() {
   return 0
 }
 
+# CTAP options come back as one comma-separated line where a leading "no" means
+# the key supports the option but it is not configured: clientPin means a PIN is
+# set, noclientPin means the key takes one and nobody has. Comparing with a
+# comma on each side is what keeps nouv from reading as uv.
+fido2_option_enabled() {
+  local name=$1 info=$2 options
+
+  options=$(sed -n 's/^options:[[:space:]]*//p' <<<"$info" | tr -d ' ')
+
+  [[ ,$options, == *",$name,"* ]]
+}
+
+# A key enforcing alwaysUv refuses every assertion that does not carry user
+# verification, so a presence-only credential can never authenticate with it:
+# the key never lights up, and sudo falls through to the password. A key that
+# does not enforce it is registered exactly as before, PIN or not. What is
+# recorded here is a requirement the authenticator states, not a policy added
+# on top of one.
+#
+# Recorded per credential rather than as a pinverification= module argument,
+# which pam_u2f ORs with each credential's own flag and would therefore force a
+# PIN onto every key in the mapping file, including one that has none.
+set_registration_flags() {
+  local info=$1
+
+  REGISTRATION_FLAGS=()
+
+  if fido2_option_enabled alwaysUv "$info"; then
+    if fido2_option_enabled uv "$info"; then
+      REGISTRATION_FLAGS=(-V)
+    elif fido2_option_enabled clientPin "$info"; then
+      REGISTRATION_FLAGS=(-N)
+    else
+      return 1
+    fi
+  fi
+}
+
 setup_pam_config() {
   # Configure sudo
   if ! grep -q pam_u2f.so /etc/pam.d/sudo; then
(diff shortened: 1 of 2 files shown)


Erijl:
Thanks for the clarifying addition @dhh; Any update on mergability?

I get that the repo basically exploded since then (regarding issues & PR's), but I'd like to continue extending the support in omarchy for fido2 tokens more and more (managing multiple tokens, not just one, lockscreen support, perhaps even for the LUKS encryption eventually)

I already have a [plugin](https://github.com/Erijl/omarchy-fido2-key-suite) going that does most of that (everything except for LUKS) if you think this would be a great addition, feel free to check it out c:

https://github.com/Erijl/omarchy-fido2-key-suite

Erijl:
Rebased onto `quattro`, mergeable again.

Two things changed while this sat:

**The ownership commit is dropped.** #7904 fixed the same `/tmp/fido2` staging
defect and went considerably further — symlink guards on both paths, a
root-created stage, a migration for existing installs. Superseded, so it is
gone; this branch now carries only the `alwaysUv` fix.

**The scope is narrower.** Flags are recorded only for a key that actually
enforces `alwaysUv`. A key that does not — including one with a PIN set — is
registered exactly as it is today.

The bug itself is unchanged and still present on `quattro`: `pamu2fcfg` records
`+presence`, the key's firmware refuses any assertion without user
verification, so the assertion is rejected before the key lights up, and
because the rule is `sufficient` the stack falls through to the password. Setup
reports success throughout.

One consequence of #7904's `set -o pipefail` is folded in:
`fido2-token -L | head -1` aborts the script through `set -e` once a second key
is attached, because `head` closes the pipe and the pipeline reports 141. The
device now comes from the list `check_fido2_hardware` already read.

`test/shell.d/security-fido2-test.sh` gains the authenticator matrix and both
refusal paths, driven end to end through the real script. Each new assertion
was checked against a mutation that defeats it.

Issue #2101, opened 2025-09-30
MacBook 10,1 audio doesn't work OOTB


e-minguez:
### System details

MacBook 10,1

### What's wrong?

It requires https://github.com/leifliddy/macbook12-audio-driver
There is an aur package https://aur.archlinux.org/packages/snd-hda-macbookpro-dkms-git that uses a different repo. In my case, I've got sound with the https://github.com/leifliddy/macbook12-audio-driver one, didn't had the chance to try the other one though.

brycv:
I just installed Omarchy 3.1.4 on my macbook10,1 that I haven't used in several years and ran into this exact issue. I would love to see this integrated into Omarchy. Sound doesn't work in Fedora 43 or Ubuntu 25.10 either so this would be a nice win.

mholson:
I can confirm that Omarchy 3.1.7 sound works out of the headphone jack OOTB AND that https://github.com/leifliddy/macbook12-audio-driver did not work on my MacBook 10.1 on a fresh install by manually building and installing the module for the current kernel.

Also, OOTB I can see the HDA Intel PCH output device is showing audio output, but no sound is coming through the speakers.

e-minguez:
Kernel 6.17 brought some changes and the speakers don't work with the patched driver. See https://github.com/leifliddy/macbook12-audio-driver/issues/43 It seems there is a fork https://github.com/x5444/macbook12-audio-driver that works (no dkms though) but I couldn't test it just yet.

mholson:
> Kernel 6.17 brought some changes and the speakers don't work with the patched driver. See [leifliddy/macbook12-audio-driver#43](https://github.com/leifliddy/macbook12-audio-driver/issues/43) It seems there is a fork https://github.com/x5444/macbook12-audio-driver that works (no dkms though) but I couldn't test it just yet.

Thanks @e-minguez.  I did try the fork as well, but could not get it to work either.  But thanks for adding that additional information to the thread for others who are struggling with the audio.

💫 It works!!! [leifliddy/macbook12-audio-driver](https://github.com/leifliddy/macbook12-audio-driver) 2. manually build and install module for current kernel. Not sure why it worked THIS time, but it did.

e-minguez:
[...]

PR #6921, opened 2026-08-15
Install the CS4208 speaker driver on 12-inch MacBooks

sunoxen-lab:
## Summary

On the 12-inch MacBook (`MacBook9,1` / 2016 and `MacBook10,1` / 2017), mainline `snd_hda_codec_cs420x` detects the Cirrus Logic CS4208 and PipeWire plays to it, but the speaker amplifier is never enabled. The OS thinks audio is working; the speakers stay silent. Headphones can still work because they use a different path.

This is the same gap as #2101. Grok diagnosed it on a `MacBook10,1` running Omarchy 4, installed [leifliddy/macbook12-audio-driver](https://github.com/leifliddy/macbook12-audio-driver) via DKMS, and speakers came up. This PR is that fix paid forward to the next installer.

## What it does

- New hardware leaf `install/hardware/apple/fix-cs4208-audio.sh`, same shape as the SPI keyboard fix: detect `MacBook9,1` / `MacBook10,1`, then `omarchy-pkg-add macbook12-audio-driver-dkms`.
- `MacBook8,1` (2015) is left alone; that model is not supported by this driver.
- Migration `1786719479.sh` for existing installs. No-ops if the package is already present or if the same DKMS module was installed by hand.
- Cached in `install/omarchy-other.packages` so the ISO can ship it.
- Manual note in `manual/44-mac-support.md`.

Omarchy already applies software volume globally, so the extra WirePlumber soft-mixer step from the driver README is not needed here.

## Package

This needs `macbook12-audio-driver-dkms` in `pkgs.omarchy.org`, built from https://github.com/leifliddy/macbook12-audio-driver — the same source that restored speakers on the machine this was written on.

It is not in the repo yet. Land the package with this PR (or immediately before), the way `macbook12-spi-driver-dkms` already is. The DKMS `PRE_BUILD` downloads matching kernel HD-audio headers on the target, so the package only needs to ship the driver source and `dkms.conf`.

## Test plan

- [x] `test/shell.d/cs4208-audio-test.sh` — leaf and migration fire only on `MacBook9,1` / `MacBook10,1`, skip 2015 / Pro / non-Apple, skip when the package or a hand-installed DKMS module is already there
- [x] Verified on a 2017 12-inch MacBook (`MacBook10,1`): after the DKMS module, speakers play; jack switch still works

### DIFF
--- a/install/hardware/all.sh
+++ b/install/hardware/all.sh
@@ -33,6 +33,7 @@ run_logged "$OMARCHY_INSTALL/hardware/asus/fix-z13-touchpad.sh"
 run_logged "$OMARCHY_INSTALL/hardware/framework/qmk-hid.sh"
 
 run_logged "$OMARCHY_INSTALL/hardware/apple/fix-spi-keyboard.sh"
+run_logged "$OMARCHY_INSTALL/hardware/apple/fix-cs4208-audio.sh"
 run_logged "$OMARCHY_INSTALL/hardware/apple/fix-suspend-nvme.sh"
 run_logged "$OMARCHY_INSTALL/hardware/apple/fix-t2.sh"
 run_logged "$OMARCHY_INSTALL/hardware/apple/fix-brcmfmac-supplicant.sh"
--- a/install/hardware/apple/fix-cs4208-audio.sh
+++ b/install/hardware/apple/fix-cs4208-audio.sh
@@ -1,0 +1,14 @@
+# Enable speakers on 12-inch MacBooks (MacBook9,1 / 2016 and MacBook10,1 / 2017).
+# Mainline snd_hda_codec_cs420x detects the CS4208 but never runs the macOS
+# init that powers the speaker amplifier, so PipeWire plays to a silent analog
+# path. The out-of-tree module replays that init.
+# MacBook8,1 (2015) is not supported by this driver.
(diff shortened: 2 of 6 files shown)


doctor:
I validated this approach on the affected hardware: a MacBook10,1 running Omarchy with the CS4208 codec. The internal speakers work, and PipeWire reports the expected two-channel stereo profile.

I found two integration gaps before this can land cleanly:

1. `macbook12-audio-driver-dkms` did not yet exist in the Omarchy package repository. I added it in [omarchy-pkgs#153](https://github.com/omacom-io/omarchy-pkgs/pull/153).
2. The CS4208 needs its hardware mixer held at full scale while user volume is handled in software. I opened [sunoxen-lab/omarchy#1](https://github.com/sunoxen-lab/omarchy/pull/1) directly against this PR's source branch, so you can merge it here without superseding your original contribution. It also makes the migration idempotent, preserves working manual installations, repairs missing current-kernel DKMS builds, and adds 12 focused tests.

The corresponding upstream driver lifecycle fix is [macbook12-audio-driver#60](https://github.com/leifliddy/macbook12-audio-driver/pull/60).

This keeps #6921 as the main Omarchy PR while giving each repository a focused, reviewable change.

PR #6924, opened 2026-08-15
Prevent togglesplit error in scrolling layout

seshna:
Prevent Super+J from dispatching togglesplit outside Dwindle layouts.

  The shortcut now checks the active (including special) workspace layout before dispatching. Added regression coverage
  for Dwindle and Scrolling workspaces.

  Tested with ./test/all.

### DIFF
--- a/default/hypr/bindings/tiling.lua
+++ b/default/hypr/bindings/tiling.lua
@@ -1,7 +1,13 @@
 o.bind("SUPER + W", "Close window", hl.dsp.window.close())
 o.bind("CTRL + ALT + DELETE", "Close all windows", "omarchy-hyprland-window-close-all")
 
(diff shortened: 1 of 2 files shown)


dhh:
What is the problem you're solving here?

seshna:
> What is the problem you're solving here?

Pressing Super + J in the scrolling layout currently results in an error. Since toggling a window’s position doesn’t make sense in that layout, this change checks the active layout first and skips the action when scrolling is in use.

I thought it was a good idea but if its a bit much, apologies for wasting time

fayimora:
I solved this locally by making the binding expel/consume in scrolling and i've PR'd the change if you're interested

Tsurgcom:
+1

Issue #6934, opened 2026-08-15
Bar clock always shows weekday/month names in English, ignoring system locale (LANG)


isaac30503:
## Title
Bar clock always shows weekday/month names in English, ignoring system locale (LANG)

## Description

The bar clock widget (`omarchy.clock`) always renders weekday names (`dddd`/`ddd`) and month names (`MMMM`/`MMM`) in English, regardless of the system locale. Setting `LANG`/`LC_TIME` to a non-English locale (e.g. `es_MX.UTF-8`) correctly changes every other locale-aware value on the system, but the clock widget in the bar is unaffected.

This used to work correctly under the old waybar-based bar (pre-Quattro); it broke after the migration to the Quickshell-based `omarchy-shell`.

## Steps to reproduce

1. Set a non-English system locale, e.g.:
   ```
   localectl set-locale LANG=es_MX.UTF-8
   ```
   and reboot.
2. Confirm the locale is applied system-wide:
   ```
   $ localectl status
   System Locale: LANG=es_MX.UTF-8
   $ systemctl --user show-environment | grep LANG
   LANG=es_MX.UTF-8
   ```
3. Set the bar clock format to include a weekday/month token, e.g. `dddd HH:mm` (the default).
4. Observe the bar: it shows `Saturday 14:05` instead of the expected `sábado 14:05`.

## Root cause

`BarWidget.qml` for the clock plugin formats the date with the QML global function `Qt.formatDateTime(date, format)`:

https://github.com/basecamp/omarchy/blob/master/shell/plugins/panels/clock/BarWidget.qml

```qml
function formatted(date) {
  return Qt.formatDateTime(date, activeFormat.replace(/ww/g, Model.isoWeekLiteral(date.getFullYear(), date.getMonth(), date.getDate())))
}
```

The two-argument `Qt.formatDateTime(date, format)` overload does **not** use the application's default locale for textual tokens (`dddd`, `ddd`, `MMMM`, `MMM`) — it always formats those using English/C-locale names, independent of `Qt.locale()` or the process environment.

I confirmed this directly against the running `quickshell` binary (`Quickshell 0.3.0`) with `LANG=es_MX.UTF-8 LC_ALL=es_MX.UTF-8`:

```
Qt.locale().name:                         es_MX
Qt.formatDateTime(d, "dddd d MMMM"):      Saturday 15 August      ← wrong
d.toLocaleString(Qt.locale(), "dddd HH:mm"): sábado 14:05         ← correct
```

[...]

Issue #6914, opened 2026-08-15
`o.shell_succeeds()` always returns false inside Hyprland, so NVIDIA env detection never fires


litescript:
### System details

AMD Ryzen 7 7800X3D, NVIDIA RTX 5070 Ti (GB203, `10de:2c05`) + AMD Raphael iGPU, nvidia-open-dkms 610.57.04-1, Omarchy 4.0.0.alpha, Hyprland 0.56.2

### What's wrong?

`o.shell_succeeds()` in `default/hypr/helpers.lua` can never return true when
evaluated inside Hyprland. It wraps `os.execute()`, and in the Hyprland process
the child's exit status is unavailable — the call returns
`nil, "No child processes", 10` (ECHILD) no matter how the command exits.

The command genuinely runs. Only its status is lost. So every `shell_succeeds`
guard silently takes the false branch.

`default/hypr/nvidia.lua` is the only caller on `main` today, so the visible
consequence is that **NVIDIA environment detection never fires** — `NVD_BACKEND`,
`LIBVA_DRIVER_NAME` and `__GLX_VENDOR_LIBRARY_NAME` are never set, even where
the detectors plainly succeed from a shell.

## Reproduction

```
$ /usr/share/omarchy/bin/omarchy-hw-nvidia; echo $?
0                                       # detector succeeds from a shell

$ hyprctl repl 'return tostring(o.shell_succeeds(o.shell_quote("/usr/share/omarchy/bin/omarchy-hw-nvidia")))'
false                                   # ...but false inside Hyprland

$ hyprctl repl 'local ok,how,code = os.execute("/usr/share/omarchy/bin/omarchy-hw-nvidia >/dev/null 2>&1"); return tostring(ok).." / "..tostring(how).." / "..tostring(code)'
nil / No child processes / 10           # ECHILD
```

It is not specific to those scripts — bare `true` fails the same way:

```
$ hyprctl repl 'local ok,how,code=os.execute("true"); return tostring(ok).." / "..tostring(how).." / "..tostring(code)'
nil / No child processes / 10
```

The command really does execute: `hyprctl repl 'os.execute("touch /tmp/proof")'`
creates the file. Only the `waitpid()` fails.

Effect on the environment, captured on stock config — every unconditional
`hl.env()` in `default/hypr/envs.lua` reaches a Hyprland-spawned child, and only
the three inside the `nvidia.lua` guard are missing:

```
[...]

Issue #6913, opened 2026-08-15
Clone Plugin from Setup > Plugins fails: "unknown clone option: <id>"


aledx18:
### System details

Omarchy 4.0.0-1, kernel 7.1.8-arch1-3, Hyprland 0.56.2, desktop (no laptop)

### What's wrong?

Cloning a built-in plugin from the menu (`Setup > Plugins > Clone Plugin`) always
fails with:

omarchy-plugin-clone: unknown clone option: omarchy.weather

The clone is never created, and the plugin list stays unchanged. No other menu
action (Enable/Disable/Remove) is affected.

### Steps to reproduce

1. Open the Omarchy menu (`SUPER + SPACE`) → Setup → Plugins → **Clone Plugin**.
2. Pick any first-party plugin (e.g. Weather).
3. The floating terminal shows `omarchy-plugin-clone: unknown clone option: omarchy.weather`
   and nothing is cloned.
4. `ls ~/.config/omarchy/plugins/` stays empty.

- omarchy-plugin-clone --edit omarchy.weather → reproduces the error.
- omarchy-plugin-clone omarchy.weather --edit (id first) → parses correctly

Issue #6908, opened 2026-08-14
mise wrappers pollute stdout and break protocol-based tools


luisrudge:
`omarchy-mise-install` generates wrappers containing:

```bash
mise use -g "$package" || exit 1
```

This writes a mise status message to stdout before the wrapped command runs:

```text
mise ~/.config/mise/config.toml tools: codex@0.147.0
```

That breaks tools using stdout as a protocol. For example, T3 Code fails to start `codex app-server` with:

```text
Codex App Server protocol operation 'decode-wire-message' failed
```

Using mise's quiet option fixes it while preserving the current update behavior:

```bash
mise use -g --quiet "$package" || exit 1
```

Tested locally on Omarchy 4.0.0 with mise 2026.8.3 and Codex 0.147.0.

PR #6937, opened 2026-08-15
Fix seven issues reported after 4.0.0

dhh:
Triage of the 33 issues filed since the 4.0.0 release. These are the seven that were confirmed against the code, have a contained fix, and could be verified here. Each commit closes one issue.

**#6903 — foot.ini syntax error on startup.** A 3.8.3 migration appended the section header with `sed -i '$a\\\n[text-bindings]'`, one backslash too many, so sed wrote the literal characters `\n[text-bindings]` as a line. foot rejects it and stops reading the rest of the file. The later text-binding migration looks for the header with `grep -qxF`, misses the broken line, and appends a *second* section, so the config stays broken across the Quattro upgrade. New migration deletes the literal line; verified `foot --check-config` passes afterward, and that it no-ops on healthy and missing configs.

**#6914 — `o.shell_succeeds()` always false inside Hyprland.** Hyprland reaps its own children, so `os.execute()` gets ECHILD from waitpid and never sees an exit status. Confirmed live: `o.shell_succeeds("true")` returned `false` on this machine. The only callers are `hypr/nvidia.lua`, so `NVD_BACKEND`, `LIBVA_DRIVER_NAME` and `__GLX_VENDOR_LIBRARY_NAME` were never set on NVIDIA hardware. Now reads a marker off stdout; verified inside the running compositor.

**#6908 — mise wrappers pollute stdout.** `mise use -g` announces the resolved tool on stdout, so every wrapped command prepended a `tools:` line to its own output, corrupting anything speaking a protocol over stdout such as `codex app-server`. This one bit repeatedly during the triage itself — several `grep` calls in this work came back with a `tools:` line stuck to the front. Fixed with `--quiet`, which keeps errors on stderr and preserves the exit status.

**#6934 — bar clock always English.** `Qt.formatDateTime` renders `dddd`/`MMMM` from the C locale whatever `Qt.locale()` says, so the clock stayed English while its own calendar popup, which already goes through `Qt.locale()`, did not. Checked all twelve shipped presets: byte-identical under the C locale, correctly localized under `de_DE`.

**#6913 — Clone Plugin fails with "unknown clone option".** The menu passed `--edit` ahead of the plugin id, but `omarchy-plugin-clone` only accepts the id as the first argument, so it fell through to the unknown-option branch and every clone from Setup > Plugins failed. Argument order swapped.

**#6881 — bar stuck in move mode.** Bar widgets propagate their composed press-and-hold down to the center gesture area without handing over the grab, so the gesture area started a bar move and then got neither a release nor a cancel to end it — the move ghost stayed on screen for the rest of the session. Confirmed the two paths are distinguishable under Qt 6.11: the propagated event arrives with `pressed=false` and no release, a genuine hold with `pressed=true`. Guarded on that. Note the reported ~200ms trigger is actually ~800ms, since the timer runs on the grabbing widget, not the gesture area.

**#6882 — Windows VM connects before the guest is ready.** The readiness scan read the container's whole log, which survives stop/start, so it matched the previous boot's line and launched `xfreerdp3` immediately. The wait also sat inside the not-running branch, skipping it entirely for a container that was already up. The scan is now scoped to the container's current start time, re-read on each poll so a restart mid-wait cannot match the boot we were watching.

Reviewed with codex at xhigh; the one finding it raised that held up (the start time being sampled once) is folded into the last commit.

### DIFF
--- a/bin/omarchy-menu-plugin
+++ b/bin/omarchy-menu-plugin
@@ -38,7 +38,7 @@ id=$(cut -f2 <<<"$selection")
 
 if [[ $1 == "clone" ]]; then
   omarchy-launch-floating-terminal-with-presentation \
-    "omarchy-plugin-clone --edit $(printf '%q' "$id")"
+    "omarchy-plugin-clone $(printf '%q' "$id") --edit"
 elif [[ $1 == "remove" ]]; then
   omarchy-launch-floating-terminal-with-presentation "omarchy-plugin-remove $(printf '%q' "$id")"
 else
--- a/bin/omarchy-mise-install
+++ b/bin/omarchy-mise-install
@@ -22,7 +22,7 @@ rm -f "$HOME/.local/bin/$command"
 cat >"$HOME/.local/bin/$command" <<EOF
 #!/bin/bash
 export MISE_MINIMUM_RELEASE_AGE=0
-mise use -g "$package" || exit 1
+mise use -g --quiet "$package" || exit 1
 exec mise x "$package" -- "$bin" "\$@"
 EOF
 
--- a/bin/omarchy-windows-vm
+++ b/bin/omarchy-windows-vm
@@ -297,20 +297,30 @@ launch_windows() {
       omarchy-notification-send -u critical "Windows VM" "Failed to start Windows VM"
       exit 1
     fi
-
-    echo "Waiting for Windows VM to start..."
-    WAIT_COUNT=0
-    until docker logs omarchy-windows 2>&1 | grep -qi "windows started successfully"; do
-      sleep 2
-      WAIT_COUNT=$((WAIT_COUNT + 1))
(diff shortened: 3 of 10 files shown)


dhh:
Split into one PR per fix so they can be reviewed and merged independently:

- #6938 — foot.ini literal `\n[text-bindings]` line (#6903)
- #6939 — `o.shell_succeeds()` always false inside Hyprland (#6914)
- #6940 — mise wrappers writing to stdout (#6908)
- #6941 — bar clock ignoring locale (#6934)
- #6942 — Clone Plugin argument order (#6913)
- #6943 — bar stuck in move mode (#6881)
- #6944 — Windows VM readiness wait (#6882)

Each branches from `quattro` directly and carries the same commit reviewed here.

Issue #6648, opened 2026-08-09
Quattro: catppuccin theme sets invalid nvim colorscheme "catppuccin-nvim"


shawnyeager:
On 4.0.0.alpha with the catppuccin theme selected, every nvim launch fails with LazyVim's "could not load your colorscheme" error and falls back to an unthemed editor.

Cause: `themes/catppuccin/neovim.lua` (quattro branch, line 10) sets `colorscheme = "catppuccin-nvim"`. That's the plugin's repo name — the colorschemes the catppuccin plugin actually registers are `catppuccin`, `catppuccin-mocha`, `catppuccin-latte`, etc. Since theme-set copies this file into the user's nvim config, everyone on the default post-upgrade theme hits the error.

Fix: `colorscheme = "catppuccin"`. The other shipped themes' neovim.lua files all use valid names; catppuccin is the only one affected.

dhh:
Can't reproduce this, and `catppuccin-nvim` is the correct name on current catppuccin.

The plugin registers it. From a fresh checkout of the pinned version:

```
~/.local/share/nvim/lazy/catppuccin/colors/
  catppuccin-frappe.lua  catppuccin-latte.lua  catppuccin-macchiato.lua
  catppuccin-mocha.lua   catppuccin-nvim.vim   catppuccin.lua
```

Pointing `~/.config/nvim/lua/plugins/theme.lua` at `themes/catppuccin/neovim.lua` and launching nvim gives no error and `g:colors_name = catppuccin-mocha`.

It's also deliberate rather than a typo. Upstream renamed it in catppuccin/nvim [#977](https://github.com/catppuccin/nvim/pull/977), shipped in v2.0.0:

> `fix!: move rename catppuccin colorscheme to catppuccin-nvim`
> new builtin colorscheme name conflicts with the plugin name. adding `-nvim` prefix to indicate the nvim port.

Neovim 0.12 ships its own bundled `catppuccin` (`/usr/share/nvim/runtime/colors/catppuccin.vim`), so the plugin added the `-nvim` suffix to disambiguate. Switching back to plain `catppuccin` would make Omarchy's theme collide with Neovim's builtin, so that change would be a regression.

[...]

Issue #6205, opened 2026-07-10
Neovim configuration template for the Catppuccin theme


BeresN:
### System details

Omarchy 3.8.2

### Description
    There is a typo in the Neovim configuration template for the Catppuccin theme that
  causes Neovim to throw an error when loading.

    The `colorscheme` is incorrectly set to `"catppuccin-nvim"`, which is not a valid theme
  name for the catppuccin plugin (the correct name is just `"catppuccin"`).

    ### Location
    File: `themes/catppuccin/neovim.lua`
    Line: 10

    ### Current Code
    ```lua
    return {
      {
        "catppuccin/nvim",
        name = "catppuccin",
        priority = 1000,
      },
      {
        "LazyVim/LazyVim",
        opts = {
          colorscheme = "catppuccin-nvim", -- Incorrect name
        },
      },
    }

  ### Proposed Fix

  Change  "catppuccin-nvim"  to  "catppuccin" :

        opts = {
          colorscheme = "catppuccin",
        },

dhh:
Can't replicate this. I see catppucin-nvim working correctly. This is with 0.12.4. Maybe you need to update dependencies?

<img width="1438" height="330" alt="Image" src="https://github.com/user-attachments/assets/d8c9c5b9-f717-43a3-a7b8-83afa799aab5" />

PR #6959, opened 2026-08-15
Require checking for duplicate issues and pending fixes before filing

guirossibrum:
## Why

Issue #6648 is a concrete example of what this prevents: a report that
turned out to be a stale local plugin lock, not an Omarchy bug, closed
after a maintainer had to spend time investigating it. Neither the bug
report template nor the general-purpose `contributing.md` skill prompts
for a duplicate check or a version-matched reproduction before filing.

Omarchy already partially enforces this elsewhere: the `diagnose-crash`
agent skill's `reporting.md` requires searching open *and* closed issues,
and treats a closed-as-fixed issue that still reproduces as a regression
worth reporting, before it lets an AI agent file anything on a user's
behalf. It was missing a check for pull requests though, so this also
closes that gap there and in `contributing.md`, brings the same
discipline to the human-facing template, so it applies regardless of how
someone files.

## What

- **`bug.yml`**: three required checkboxes before the existing fields —
  searched issues, checked PRs, confirmed on latest release. GitHub's
  native `required: true` on `type: checkboxes` blocks submission until
  acknowledged, not just a markdown reminder someone can skip past.
- **`omarchy/contributing.md`**: new "Verify Before Filing" section
  requiring a version-matched upstream diff, an issue search, and a PR
  search before any agent drafts a report.
- **`diagnose-crash/reporting.md`**: added the PR-search step it was
  missing, next to its existing issue-search requirement.

## Test plan

- [x] Validated `bug.yml` YAML (`ruby -ryaml`)
- [x] `./test/cli` passes
- `./test/shell` has pre-existing failures on this machine from an
  unrelated missing `mise` Node pin; none of the failing tests reference
  any of the three files changed here (confirmed via
  `grep -rl` across `test/`), and none of these files have any existing
  test coverage

---
Filed by Claude Sonnet 5 via Claude Code, with @guirossibrum.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.yml
+++ b/.github/ISSUE_TEMPLATE/bug.yml
@@ -7,6 +7,18 @@ body:
       value: |
         Remember: Omarchy is an open source gift, not a product you bought from a vendor
 
+  - type: checkboxes
+    id: checked-existing
+    attributes:
+      label: Before filing
+      options:
+        - label: I searched [open and closed issues](https://github.com/basecamp/omarchy/issues?q=is%3Aissue) for this and found nothing matching
+          required: true
+        - label: I checked [pull requests](https://github.com/basecamp/omarchy/pulls?q=is%3Apr) for an existing fix, open or merged
+          required: true
+        - label: I confirmed this still happens on the latest Omarchy release, not just an older install
+          required: true
+
   - type: input
     id: system-details
     attributes:
--- a/default/agents/skills/diagnose-crash/reporting.md
+++ b/default/agents/skills/diagnose-crash/reporting.md
@@ -54,6 +54,14 @@ Include **closed** issues. A matching issue closed as fixed, when the crash stil
 reproduces on a current system, is a regression — and reporting that is worth far
 more than another duplicate.
 
+Also search pull requests, open and merged — a fix already merged but not yet in
+a release means "update and it's fixed," not a new bug; an open PR means point
+the user at it instead of filing a duplicate:
+
+```bash
+gh pr list --repo basecamp/omarchy --state all --search "<signal> <program>"
+```
+
 ## Adding to an existing report
 
 If a plausible match comes back, read it properly first:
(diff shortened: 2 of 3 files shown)


guirossibrum:
Addressed both review rounds:

- **"update and it's fixed" wording** (contributing.md, reporting.md): fixed — reworded to say the fix ships in the next release rather than implying it's immediately installable, since `omarchy update` only pulls released versions.
- **`v<version>` tag assumption** (contributing.md): fixed — `omarchy version`'s output (e.g. `4.0.0-1`) doesn't match its release tag verbatim (`v4.0.0`), so the example now looks up the actual tag via `gh release list` instead of concatenating a hardcoded `v` prefix.
- **Issue/PR links defaulting to open-only** (bug.yml, suppressed comment): checked against GitHub's search API directly rather than assuming either way —
  ```
  gh api "search/issues?q=repo:basecamp/omarchy+is:issue+catppuccin"   → returns closed issues (e.g. #6648, #6205)
  gh api "search/issues?q=repo:basecamp/omarchy+is:pr+catppuccin"      → returns both open and merged/closed PRs
  ```
  `is:issue` / `is:pr` alone, with no explicit `is:open`, already include closed/merged items — GitHub only defaults to open-only when there's no query at all, not when one is supplied. Left the links as-is since they already do what the checkbox text claims.

— Claude Sonnet 5 via Claude Code, with @guirossibrum.

rdoupe:
Worth knowing before this gets reviewed: the guidance here is right, but `default/agents/skills/omarchy/contributing.md` is not read by the agents it is aimed at.

- It lives inside the `omarchy` skill, whose description says it *excludes Omarchy source development*. An agent asked to write a fix decides the skill does not apply and never opens this file. That is exactly what happened in my session: skill installed, four open PRs for the bug I was fixing, and I filed a fifth (#9089, closed) without ever loading `contributing.md`.
- #6718 deletes `contributing.md` entirely and routes upstream work to the repo's `AGENTS.md` instead.

So for the PR-submission path, `AGENTS.md` is the file that actually gets read. I opened #9095 putting the search-first guidance there. The issue-reporting path is a different story: an agent reporting a bug *does* load the skill, so guidance in the skill's reporting guide (post-#6718, `reporting-issues.md`) is well placed for that.

For this PR specifically, the `bug.yml` required checkbox is the strongest piece — it is enforced at filing time regardless of what anyone read, and nothing else in the set does that. Suggest keeping that and the `diagnose-crash/reporting.md` change, and rebasing the `contributing.md` part onto whatever #6718 leaves behind.

PR #6965, opened 2026-08-15
Add git-based backup and restore

andresreibel:
mise ~/.config/mise/config.toml tools: gh@2.97.0
Omarchy installs in thirty seconds. Then you spend three days making it yours again: the keybindings you tuned, the packages you forgot you needed, the themes, the webapps, the dotfiles — a year of small decisions with no record. The install was never the slow part. Recreating *your* machine is.

`omarchy backup` puts that state in a git repo. One command, no options. It captures what Omarchy owns — hypr and omarchy configs, your terminal's config, explicit package lists, theme sources, webapp launchers — then commits and pushes to whatever private remote you give it. Want more in the backup? Add paths to `backup.list`, one per line. That's the entire configuration surface: a text file.

`omarchy restore` replays it. Packages first, then configs, themes, webapps, extras — ending with `omarchy-theme-set` so the desktop reloads into place. Anything that differs gets set aside as a timestamped `.bak`, the same way `omarchy-refresh-config` does it. Run it twice and the second run does nothing. It never deletes. Thirty seconds to install, one command to come home.

One more piece: `recovery/`. Every backup snapshots how the machine boots — disk map with UUIDs, kernel cmdline, EFI entries, limine.conf, fstab. Restore never applies these; machine IDs must be regenerated, not copied. They exist for the day the machine won't boot at all: open the repo on your phone, and it tells you which partition to unlock from the live USB session instead of guessing.

Nothing here is new machinery. Config capture follows what `refresh-` already treats as user config. Packages replay through pacman's own `--needed`. Themes reinstall from their git remotes, laid out exactly as `omarchy-theme-install` leaves them. Backup is also offered from the Update menu, the way Update → Omarchy runs `omarchy-update`. On a machine with no backup yet, restore asks for the repo URL the same way `omarchy-theme-install` does — fresh-install recovery is: menu → Update → Restore → paste your link. The diff is two commands, two dispatcher lines, and two menu entries.

Tested with a full round-trip in `test/shell.d/backup-restore-test.sh` — capture, wipe, restore, `.bak` asides, idempotence, the embedded-git-repo edge case, and refusal without a repo or URL. `./test/all`: 181/181.


andresreibel:
One addition since opening: restore now asks for your repo URL when there's no backup on the machine — the same `gum` prompt `omarchy-theme-install` uses. Both commands sit in the menu under Update.

So the fresh-install story is now: open the menu, hit Restore, paste your link, watch your machine come back. Private repo? Git asks for your credentials right there. Nothing to memorize, nothing to look up.

andresreibel:
On separation of concerns: the `recovery/` snapshot is one self-contained commit. If backup/restore should stand alone, say the word and I'll split recovery into its own PR — this one shrinks to pure backup and restore, no other changes needed.

crazybadger:
I did this exact migration by hand recently — a git dotfiles repo for config + package lists, plus `borg` for the data half — and moved between two laptops. Strong +1 on the framing; "recreating *your* machine is the slow part" is exactly right. A few gaps I hit that are in scope for a config+packages restore:

**Plugin root-halves.** Omarchy shell plugins can ship a `system/install.sh` — a kernel module via `modules-load.d`, a `/usr/local/bin` binary, a systemd *system* unit. Capturing `~/.config/omarchy/plugins/<id>/` brings the QML widget back but it's dead on arrival until that script re-runs. Restore could detect `plugins/*/system/install.sh` and offer to run them.

**AUR vs repo packages.** `pacman -Qqen` (repo) and `pacman -Qqem` (foreign/AUR) replay differently — pacman for the first, an AUR helper for the second. A flat `-Qqe` replayed through pacman means every AUR package silently fails. Worth capturing the two lists separately.

**Enabled system services.** Tailscale is the clean example — `tailscaled` enablement + `tailscale up` auth live entirely outside `~` and outside pacman. Restore can't reconnect them, but capturing `systemctl list-unit-files --state=enabled` (system + `--user`) would at least tell the user what to turn back on.

**`/etc` files.** If `backup.list` accepts absolute paths, replaying `/etc/logid.cfg` and friends needs sudo and a different conflict story than `~`. Might be cleaner to declare it `~`-only than half-support it.

**Keyrings.** If anyone puts `~/.local/share/keyrings` in `backup.list`, it only decrypts on the new machine if the login password matches — a one-line doc warning saves someone a confusing afternoon.

The `recovery/` boot snapshot and the `.bak`-aside matching `refresh-config` are both great calls. This looks complementary to #7814 — that's the data / off-site half, this is the "your decisions" half.

PR #6967, opened 2026-08-15
Add Haskell development environment via GHCup

calledtoconstruct:
## What

Adds Haskell to Install / Remove → Development, using GHCup (GHC, Cabal, HLS).

## Why GHCup

Mise does not manage a Haskell toolchain the way it does the other languages. GHCup is the upstream installer.

## Editor

If `~/.config/nvim` is present, enable LazyVim's Haskell extra and point HLS at GHCup's wrapper. Mason's HLS is built against a different GHC and fails against this toolchain.

## Test plan

- [x] `omarchy install dev-env haskell` installs `ghc`, `cabal`, `haskell-language-server-wrapper`
- [x] Install → Development shows Haskell when `~/.ghcup` is absent
- [x] Remove → Development shows Haskell when `~/.ghcup` is present
- [x] `omarchy remove dev-env haskell` removes GHCup and the LazyVim extra
- [x] `./test/all`


calledtoconstruct:
Rebased onto current quattro. Menu install row uses `disabled` like the other development entries.

`env-bootstrap` no longer sources `~/.ghcup/env` (user-writable shell in login/uwsm). It appends `~/.ghcup/bin` when that directory exists.

calledtoconstruct:
Converted to draft. Omarchy wants feature ideas in Discussions first. Suggestion: https://github.com/omacom/omarchy/discussions/11265 — happy to mark this ready if that lands.

PR #6986, opened 2026-08-15
multi-monitor window highlighting in slurp capture

Liran-shternberg:
Using the capture when you have more the one monitor doesnt let you auto capture the open windows in other monitors that you are not focused on.

get_rectangles() only queried the focused monitor's workspace, so on multi-monitor setups the picker was blind to windows on other screens — window capture required a manual click-and-drag and --select-window/Tab navigation skipped those windows entirely.

The fix now:

- Returns geometries for all connected monitors.

- Builds an array of active workspace IDs across every screen and filters the hyprctl clients list against it, so windows on any visible workspace are pickable.

- Fetches hyprctl monitors/clients once and reuses the JSON via jq --argjson (plus --argjson active-workspace array).
Same code now powers omarchy-capture-region, so screenshots, recordings, and window navigation all highlight across monitors.

screenshot-2026-08-15_17-52-38

screenshot-2026-08-15_17-52-50

### DIFF
--- a/bin/omarchy-capture-region
+++ b/bin/omarchy-capture-region
@@ -37,25 +37,33 @@ JQ_MONITOR_GEO='
     end;
 '
 
-active_workspace() {
-  hyprctl monitors -j | jq -r '.[] | select(.focused == true) | .activeWorkspace.id'
+# Workspace ids visible on any connected monitor, as a JSON array.
+active_workspaces() {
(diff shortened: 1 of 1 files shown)


bjarneo:
Automated duplication check: this pull request looks similar to #11441, which covers the same capture region all displays. I keep that one open and close this one to consolidate review.

PR #7007, opened 2026-08-15
Render the Plymouth theme at the display's device scale

Bramjetten:
## The problem

Plymouth reads the panel's physical size from the EDID and derives a device scale from it, then resamples every sprite to match. From `get_device_scale()` in `src/libply/ply-utils.c`:

```
diag_inches  = hypot(width_mm, height_mm) / 25.4
physical_dpi = hypot(width_px, height_px) / diag_inches
target_dpi   = diag_inches >= 20 ? 110 : 135
scale        = (physical_dpi / target_dpi) > 1.625 ? 2 : 1
```

On the 27" 5K display I hit this on:

```
5120x2880 on 600x330 mm  ->  27.0"  ->  218 DPI
target_dpi = 110         ->  perfect_scale = 1.98
1.98 > 1.625             ->  device_scale = 2
```

The theme ships only 1x art, so at scale 2 the disk-passphrase screen reaches the framebuffer through a 2x upscale. It reads as if the boot screen is running at a lower resolution than the desktop.

Two assets could not survive resampling at any scale:

- **`entry.png`** bakes its border in as three rows of pixels — 58% alpha, solid, 58% alpha. That is an anti-aliased 1px line frozen into a bitmap, so the entry field has a soft double edge even at 1x.
- **`lock.png`** is anti-aliased throughout, so upscaling only enlarges the existing gradient.

## Why it has to happen at install time

The script plugin exposes `Window.GetWidth/GetHeight/GetX/GetY`, `Image.Scale` and `Plymouth.GetMode`, but no accessor for the device scale and no `@2x` asset lookup. A theme cannot adapt at runtime, so the scale has to be resolved when the theme is installed and the initramfs is rebuilt — which is already where `omarchy-plymouth-set` generates art with ImageMagick.

## What this does

- `entry`, `lock` and `bullet` move to vector sources under `default/plymouth/src`, rendered at the size they will actually occupy. Authored in 1x logical units with half-pixel offsets so strokes stay on the pixel grid at every integer scale.
- The logo stays a bitmap. It is 96% hard-edged pixel art, so nearest-neighbour reproduces it exactly.
- The script's hard-coded 1x constants are expressed in terms of a generated `global.scale`:

  ```diff
  +global.scale = 1;
  -entry.y = logo.sprite.GetY() + logo.image.GetHeight() + 40;
  +entry.y = logo.sprite.GetY() + logo.image.GetHeight() + 40 * global.scale;
  -lock_scale = lock_height / 96;
  +lock_scale = lock_height / (96 * global.scale);
  ```

- `DeviceScale=1` is pinned in `plymouthd.conf`, so plymouthd stops scaling art that already sits in device pixels. `plymouth-set-default-theme` rewrites only the `Theme` key, so it survives theme switches, and the plymouth mkinitcpio hook bundles the config into the initramfs.

Four hidden commands carry it: `device-scale` reads the scale, `render-icons` draws the three vector assets at a given scale, `generate` produces a whole theme directory, `activate` pins DeviceScale and rebuilds the initramfs. `activate` also absorbs the theme-select and initramfs-rebuild block that `omarchy-plymouth-set` and `omarchy-refresh-plymouth` each carried separately.

SDDM is Qt and scales itself, so it keeps the icons at 1x. Theme previews are fixed-size thumbnails and keep 1x too.

**At `global.scale = 1` the script behaves exactly as before**, and the generated assets come out at the shipped dimensions — 286x48, 84x96, 800x188, 300x10. Nothing changes on a standard-DPI display beyond the entry field and lock icon gaining crisp edges.

## Draft, because these are open

1. **Not yet verified on real hardware.** I have not rebooted with this installed. The device-scale figure is computed from Plymouth's own formula and the generator is tested end to end, but the actual boot screen is unconfirmed. This is the main thing to resolve before it leaves draft.

2. **The lock icon changes appearance.** It is now Heroicons `lock-closed` (MIT), which has no keyhole and a thinner shackle than the current art. Using a real vector source beats tracing the PNG, but it is a design change riding along with a rendering fix — say the word and I will trace the existing shape instead so the look is preserved exactly.

3. **`librsvg` added to `omarchy-base.packages`.** It already arrives via gtk3/gtk4/hyprcursor/hyprgraphics, so this only makes an existing transitive dependency explicit. Happy to drop it and guard with `omarchy-cmd-present` instead if you would rather not add the line.

4. **Packaging coverage not checked.** `test/shell.d/unowned-system-paths-test.sh` needs a sibling `omarchy-pkgs` checkout, which I do not have, so the four new `bin/` commands and `default/plymouth/src/` have not been verified against the PKGBUILD.

5. **`DeviceScale=1` is a single-display assumption.** Attach a monitor with a different DPI and the art no longer matches until the theme is regenerated. Plymouth picks one scale for the whole session, so a mixed-DPI setup cannot be right on both screens either way. A `monitor-hotplug` hook calling `omarchy-refresh-plymouth` would cover it, but that felt like scope.

## Tests

`./test/cli` passes. `./test/shell` has 3 failures — `config`, `snapper`, `unowned-system-paths` — all of them pre-existing on a clean checkout of `quattro` here, all from the missing `omarchy-pkgs` checkout.

### DIFF
--- a/bin/omarchy-plymouth-activate
+++ b/bin/omarchy-plymouth-activate
@@ -1,0 +1,58 @@
+#!/bin/bash
+
+# omarchy:summary=Select the Omarchy Plymouth theme and rebuild the initramfs
+# omarchy:hidden=true
+# omarchy:requires-sudo=true
+
+# omarchy-plymouth-generate renders the theme's art at the display's device
+# scale, so plymouthd must not scale it a second time: with the art already
+# sized in device pixels, one logical pixel has to equal one device pixel.
+#
+# plymouth-set-default-theme rewrites only the Theme key, so DeviceScale
+# survives a theme switch. The plymouth mkinitcpio hook bundles this config
+# into the initramfs, which is where the disk passphrase is asked for.
+
+set -euo pipefail
+
+conf=/etc/plymouth/plymouthd.conf
+
+sudo python3 - "$conf" <<'PY'
+import os, sys
+
+path = sys.argv[1]
+lines = open(path).read().splitlines() if os.path.exists(path) else []
+
+out, in_daemon, written, seen_daemon = [], False, False, False
+for line in lines:
+    stripped = line.strip()
+    if stripped.startswith("[") and stripped.endswith("]"):
+        if in_daemon and not written:
+            out.append("DeviceScale=1")
+            written = True
+        in_daemon = stripped == "[Daemon]"
+        seen_daemon = seen_daemon or in_daemon
+        out.append(line)
+        continue
+    if in_daemon and stripped.split("=")[0].strip() == "DeviceScale":
+        if not written:
+            out.append("DeviceScale=1")
+            written = True
+        continue
+    out.append(line)
+
+if in_daemon and not written:
+    out.append("DeviceScale=1")
+    written = True
+if not seen_daemon:
+    out += ["[Daemon]", "DeviceScale=1"]
+
+open(path, "w").write("\n".join(out) + "\n")
+PY
+
+sudo plymouth-set-default-theme omarchy
+
+if omarchy-cmd-present limine-mkinitcpio; then
+  sudo limine-mkinitcpio
+else
+  sudo mkinitcpio -P
+fi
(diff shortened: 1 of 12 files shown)


Bramjetten:
Closing for now, my agent was a bit too eager creating PRs.

PR #7011, opened 2026-08-15
Keep bar clicks out of login shells

SL0wZEr:
Clicking the bar's menu icon spawns a login shell and a `qs ipc` client to send a message back to the *same* Quickshell process it came from. The menu plugin already lives there. `bar.run()` uses `bash -lc`, so every click also sources the user's whole login profile.

The workspace switcher does the same, running `hyprctl` through a login shell even though the widget already reads its state over Hyprland IPC.

**Fix:** menu calls `shell.toggle()` directly (old command kept as a fallback); workspaces use `Hyprland.dispatch()`.

## Numbers

Per click: **3 subprocesses → 0**, ~92 ms → in process.

The process count is the same anywhere. The milliseconds depend on your profile, mine has a `/etc/profile.d` script that starts Python, so a login shell costs ~50 ms warm, ~110 ms after boot.

On the unpatched build both paths spawn 3 processes, but `SUPER+SPACE` takes 45 ms and a click takes 92 ms. Hyprland execs directly; the click pays `bash -lc` first. That gap is the login shell.

## Videos

The overlay counts subprocesses per click.

https://github.com/user-attachments/assets/95a6021b-8b77-4bbf-86a5-cbf44fd75fc2

https://github.com/user-attachments/assets/1999c3c3-ec4e-4c25-8cb3-3e92d7e7605c

## Notes

- `SUPER+SPACE` is unchanged and still spawns 3 it comes from Hyprland, so it has to.
- `Util.execDetached` keeps `bash -lc`; that is deliberate and `shell-launch-test.sh` asserts it. This only changes two click handlers that never needed to leave the process.
- Same idea as #6815 and a64d895a.

### DIFF
--- a/shell/plugins/bar/widgets/Workspaces.qml
+++ b/shell/plugins/bar/widgets/Workspaces.qml
@@ -32,7 +32,7 @@ BarWidget {
 
   function focusWorkspace(id) {
     if (!root.bar) return
-    root.bar.run("hyprctl dispatch " + Util.shellQuote("hl.dsp.focus({ workspace = \"" + id + "\" })"))
+    Hyprland.dispatch("hl.dsp.focus({ workspace = \"" + id + "\" })")
   }
 
   readonly property real trailingGap: root.vertical ? 0 : Style.spaceReal(1.5)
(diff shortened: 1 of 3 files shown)


YousefAliUK:
looks great, one thing that was missed is `shell/plugins/bar/widgets/Microphone.qml` line 45, which still shells out to `omarchy-shell shell toggle omarchy.audio` on middle click. you can use `root.bar.shell.toggle("omarchy.audio")` there as well to keep it in process

PR #7016, opened 2026-08-15
Added fuzzy search to menu

pukljak:
Currently menu search matches literal substrings only.
Now it can be searched just with few characters that appears in the naming of app, config or what ever we want to find.

Before:

https://github.com/user-attachments/assets/a5384d4a-0c1b-427e-a3a9-0850d8a6cb3e

After:

https://github.com/user-attachments/assets/b2e0abac-ce71-4943-941f-0a58a3ff42af

### DIFF
--- a/shell/plugins/menu/Menu.qml
+++ b/shell/plugins/menu/Menu.qml
@@ -498,6 +498,24 @@ Item {
     return MenuModel.descriptionTextMatches(query, text)
   }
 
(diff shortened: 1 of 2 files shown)


mkenigs:
It tripped me up today that searching for a shortcut `alt tab` has no results (have to search `alt + tab`) which it seems this would fix, so +1 from me

Issue #6911, opened 2026-08-15
Quattro upgrade replaces customized monitors.conf with default monitors.lua (auto layout, GDK_SCALE=2) without porting or warning


daventhedude:
### System details

AMD Ryzen 7 7800X3D, NVIDIA RTX 4070, Omarchy 3.8.4 → 4.0.0-1 (Quattro upgrade on 2026-08-15), triple 1440p setup (2× HP Z27i, 1× HP X27q 165 Hz)

### What's wrong?

The upgrade-to-Quattro migration creates `~/.config/hypr/monitors.lua` from the stock template and silently stops loading the user's customized `~/.config/hypr/monitors.conf`.

Unlike the other configs the upgrade touches (waybar, mako, hyprsunset, brave-flags — all parked as `*.omarchy-upgrade-to-quattro.<ts>.bak`), `monitors.conf` is left in place but unreferenced: no `.bak`, no mention in the upgrade output. There is no breadcrumb telling the user their display configuration stopped applying.

After relogin, a multi-monitor user lands in:

- auto layout — the template only contains
  `hl.monitor({ output = "", mode = "preferred", position = "auto", scale = omarchy_monitor_scale })`,
  so my portrait monitor came up landscape and all positions were wrong
- `GDK_SCALE=2` from the template (`local omarchy_gdk_scale = 2`), while my `monitors.conf` had `env = GDK_SCALE,1` — so GTK apps also rendered at the wrong scale on 1440p @ scale 1

**Steps to recreate**

1. On Omarchy 3.8.x, have a customized `~/.config/hypr/monitors.conf` (mine: three 1440p monitors matched by `desc:`, one with `transform, 1`, one with `vrr, 2`, plus `env = GDK_SCALE,1`)
2. Run the upgrade to Quattro (`omarchy update` → accept the invite)
3. Relogin: `~/.config/hypr/monitors.lua` contains only the template defaults; `monitors.conf` is ignored

**Expected**

Either the migration ports the simple `monitor =` / `env = GDK_SCALE` lines into `hl.monitor()` / `hl.env()` calls, or — probably the pragmatic 80% fix — `monitors.conf` gets the same `.bak` treatment as the other configs plus a line in the upgrade summary, so users immediately know to redo their display setup in `monitors.lua`.

**Workaround**

Rewriting `monitors.lua` by hand restores everything, e.g.:

```lua
hl.env("GDK_SCALE", "1")
[...]

Issue #2824, opened 2025-10-24
Java GUI Applications (BurpSuite, ghidra, JADX-GUI, JD-GUI) Appear Over-Scaled


mohamedbril:
### System details

Omarchy 3.1.1, NVIDIA GeForce RTX 3050

### What's wrong?

Hello,

I’m experiencing a scaling issue when using Java-based GUI applications on Hyprland (Wayland). Applications like BurpSuite,ghidra, JADX-GUI, and JD-GUI appear over-zoomed or too large, making the interface difficult to use. This issue doesn’t happen with native Wayland or non-Java applications.

![Image](https://github.com/user-attachments/assets/2985bbb4-d6b5-47c5-b88f-eeeee95dc742)

ryanrhughes:
These would be related to your `GDK_SCALE` value you have set in `~/.config/hypr/monitors.conf`.

Since this is support related and not really indicative of a repeatable bug, please utilize the Discord to seek help from the community there.

Join Discord at https://omarchy.org/discord and create a thread in #omarchy-help for assistance.

v-t-r-gg:
Reproduced the same class of bug on Quattro (4.0.0-1): packaged `GDK_SCALE=2` plus `xwayland.force_zero_scaling` on a 1080p panel whose auto scale is 1.5. Java/X11 min sizes double and extra chrome hangs off the physical screen. Writeup and suggested defaults: #7021. Companion AWT-popup tiling fix: #7020.

PR #7020, opened 2026-08-15
Float Java AWT XWayland popups instead of tiling them

v-t-r-gg:
## Why

This is not an app-specific bug. Two Quattro default-config interactions break a class of X11 / Java clients. Native Wayland GTK/Qt is mostly fine.

### 1. AWT popups reuse the parent window class (this PR)

Java tooltips, menus, and tiny dialogs are real X11 windows:

- **class:** same as the parent (`jetbrains-idea`, `ghidra`, …)
- **title:** `win0`, `win1`, `win23`

Hyprland treats that as another normal client. Any **class-only** rule (`fullscreen`, `maximize`, `tile`, `center`) stretches a ~20×20 grey tooltip over the tile. The hint text sits in the top-left of a huge grey box.

Omarchy already floats *empty* class+title XWayland windows in `default/hypr/windows.lua`. AWT popups do not match that rule, because they keep the parent class.

This PR adds the generic companion rule:

```lua
o.window({
  class = ".*",
  title = "^win[0-9]+$",
  xwayland = true,
}, {
  float = true,
  no_initial_focus = true,
  no_focus = true,
  no_follow_mouse = true,
})
```

Verified on Hyprland 0.56.2 / Omarchy 4.0.0-1: hover tooltips stay small and do not steal focus.

### 2. Related default: `GDK_SCALE=2` vs unscaled XWayland (not in this PR)

Packaged `monitors.lua` hardcodes `GDK_SCALE=2` while monitor scale is `"auto"`. Packaged `envs.lua` sets `xwayland.force_zero_scaling = true`.

On a 1080p HiDPI laptop, `scale = auto` lands on **1.5** (and can flip to **2**). Three layers then disagree:

| Layer | Size the app is told |
|---|---|
| Hyprland layout | Logical `1280×720` (`1920 / 1.5`) |
| XWayland (`force_zero_scaling`) | Physical `1920×1080`, 1:1 |
| Session env | `GDK_SCALE=2` for every child |

X11 / Java AWT see a 1920-wide X screen **and** honor `GDK_SCALE=2`. Toolkit minimum sizes double. Extra chrome then requests a width past 1920. A tiling compositor will not shift the toplevel left the way Plasma does.

Workaround (process-only, do not change global `GDK_SCALE`):

```bash
GDK_SCALE=1 GDK_DPI_SCALE=1 JAVA_TOOL_OPTIONS=-Dsun.java2d.uiScale=1 <command>
```

Suggested follow-up: derive `GDK_SCALE` from the applied monitor scale, or export it only for Wayland GTK when `force_zero_scaling` is on.

Related: #2824, #758, #6911.

## Test plan

- [x] Hover a toolbar/control in a Java AWT app under XWayland: tooltip stays a small floating box, does not fill the client
- [ ] Confirm empty-class XWayland drag windows still get the existing no-focus rule
- [ ] Confirm main tiled/maximized Java windows are unchanged

### DIFF
--- a/default/hypr/windows.lua
+++ b/default/hypr/windows.lua
@@ -18,6 +18,20 @@ o.window(
   { no_focus = true }
 )
 
+-- Java AWT tooltips/menus are real X11 windows that reuse the parent class
+-- and title themselves win0, win1, ... A class-only tile/maximize rule
+-- stretches them over the whole client. Float and keep them unfocused.
+o.window({
+  class = ".*",
+  title = "^win[0-9]+$",
+  xwayland = true,
+}, {
+  float = true,
+  no_initial_focus = true,
+  no_focus = true,
+  no_follow_mouse = true,
+})
+
 -- App-specific tweaks (may remove default-opacity tag).
 require("default.hypr.apps")
 


v-t-r-gg:
The scale half of this (hardcoded `GDK_SCALE=2` + `force_zero_scaling` + `scale = auto` overflowing XWayland/Java min-sizes on 1080p) is now a standalone issue: #7021.

Issue #7019, opened 2026-08-15
Cannot run Migration (1786643346) because an "browser window is open" according to the migration.


joshuafouch:
### System details

Dell XPS 15 9530 (2023), Intel i7-13700H

### What's wrong?

After updating and rebooting I cannot run migration (1786643346) "Repair the Copy URL shortcut for profiles that predate its pinned extension id" because I apparently have a browser window open which I do not. 

After rebooting several times and trying `pkill -f <all my browsers>`, It still says I have a browser window open.

https://github.com/user-attachments/assets/a6e11814-3c4c-4e39-a408-b27c9f89f608

joshuafouch:
I checked `btop` and did not see any chromium, brave, or firefox/zen orphan processes open as far as I know.

joshuafouch:
https://github.com/basecamp/omarchy/blob/quattro/migrations/1786643346.sh

Here is the specific migration ^

joshuafouch:
Fixed in #7026

omarchybot:
Mechanism, for anyone else stuck here: the migration never looks at processes or windows. `profile_open()` ([migrations/1786643346.sh:129](https://github.com/basecamp/omarchy/blob/quattro/migrations/1786643346.sh#L129)) decides a browser is open purely from filesystem markers in the browser's user-data-dir:

```bash
[[ -L $1/SingletonLock || -e $1/SingletonLock || -S $1/SingletonSocket ]]
```

Chromium-family browsers create `SingletonLock` as a symlink to `<hostname>-<pid>` and unlink it on a clean exit, but leave it behind after a crash, a `SIGKILL`, or a hard power-off. It lives under `~/.config/...`, not `/tmp` or `/run`, so a reboot never clears it and `pkill` has nothing left to signal — which is why neither helped.

Confirming the prompt doesn't help either. The guard is a `while` loop, so a marker that is permanently true just re-prompts forever; declining is the only way out, and that exits 1. I reproduced this in a throwaway `$HOME` with nothing but a dangling `SingletonLock`: 25 confirmations, 25 prompts, no progress.

Workaround until a fix lands. With every Chromium-family browser genuinely closed — this part matters, because removing the lock while a browser really is running lets a second instance open the same profile — find the stale marker:

```bash
ls -l ~/.config/{chromium,google-chrome*,BraveSoftware/Brave-Browser*,microsoft-edge*,vivaldi,opera,helium}/Singleton* 2>/dev/null
```

[...]

Issue #6866, opened 2026-08-14
Migration 1786643346 loops after reboot on stale Chromium SingletonLock


rfmyrick:
### System details

- Omarchy: `4.0.0rc5-1`
- Channel: `rc`
- Kernel: `7.1.8-arch1-3`
- CPU: AMD Ryzen AI MAX+ PRO 395
- GPU: AMD Strix Halo / Radeon 8060S

### What's wrong?

Migration `1786643346` repeatedly asks me to close the browser windows and continue even after Chromium is completely stopped. Pressing Continue just returns to the same prompt.

This happened after a reboot. The affected Chromium profile still contained singleton symlinks from the previous session:

```text
SingletonLock -> <hostname>-112033
SingletonSocket -> /tmp/org.chromium.Chromium.<token>/SingletonSocket
```

There was no Chromium-family process running, PID `112033` no longer existed, and the socket target under `/tmp` no longer existed.

The migration nevertheless considers the profile open because `profile_open()` tests `-L "$profile/SingletonLock"`. That remains true for a stale symlink, so the `while affected_profile_open` loop cannot make progress regardless of how many times Continue is selected.

After moving the three stale `Singleton*` symlinks out of `~/.config/chromium`, `omarchy-migrate` immediately completed successfully. I verified that no migrations remained pending and that `Alt+Shift+L` was rebound to the pinned Copy URL extension id `bgpiichlckmfanooecilcjemknkcpngb`.

### Steps to reproduce

1. Have a Chromium profile affected by migration `1786643346`.
2. Leave a stale `SingletonLock` symlink in the profile whose encoded PID is no longer alive, plus a `SingletonSocket` symlink whose `/tmp` target no longer exists. A reboot left this state on my machine.
3. Ensure Chromium is not running.
4. Run `omarchy-migrate`.
5. Select Continue at the close-browser prompt.

### Expected behavior

The migration should distinguish a stale singleton lock from a live browser and proceed when the encoded local PID is dead and the socket target is absent. Alternatively, it could safely remove confirmed-stale singleton links before testing the profile.

### Actual behavior

The prompt repeats indefinitely because the existence of the `SingletonLock` symlink alone is treated as an open profile.

### Related work

- #6841 / #6842 fixed the hidden-prompt failure for this migration, but this report concerns a visible prompt that repeats because of stale lock state.
- #6837 introduced profile-specific lock detection and explicitly treats dangling `SingletonLock` symlinks as open.
[...]

PR #7026, opened 2026-08-15
Prevent migration (1786643346) from hanging on stale browser lock files

joshuafouch:
# The Problem

If a user's machine crashes or hard-reboots while Chromium processes are open, a stale `SingletonLock` symlink may be left behind. The migration script previously only checks for the existence of the lock file which causes the migration to get stuck indefinitely. 

# The Fix

`1786643346.sh`

Updated `profile_open()` in the migration script to extract the PID of the `SingletonLock` symlink and verify if the process is actually alive via `kill -0`. If the process doesn't exists, the lock file is considered orphaned and ignored.

# Test fix 
`copy-url-shortcut-migration-test.sh`

Because the migration now checks if the process is alive, the dummy `test-host-1234` process is no longer reliable. So the test now uses its own PID (via `$$`) as an alive test process. 

+ This closes Issue #7019

### DIFF
--- a/migrations/1786643346.sh
+++ b/migrations/1786643346.sh
@@ -127,9 +127,25 @@ unverified_repairs_exist() {
 # Whether a profile is open is mechanical: a running Chromium-family browser
 # holds a SingletonLock (and socket) inside its user-data-dir.
 profile_open() {
-  # -L catches a SingletonLock left as a dangling symlink; -e covers a plain
-  # file, and -S the socket — any of them means a browser is attached.
-  [[ -L $1/SingletonLock || -e $1/SingletonLock || -S $1/SingletonSocket ]]
+  local lock="$1/SingletonLock"
+
+  # If there is no lock or socket, the profile is definitely closed
+  [[ -L $lock || -e $lock || -S $1/SingletonSocket ]] || return 1
+
+  # If it is a symlink, verify the process is actually running
+  if [[ -L $lock ]]; then
+    local target
+    target=$(readlink "$lock")
+    local pid="${target##*-}"
+
+    # If the PID is a number but the process isn't running, it's a ghost lock
+    # kill -0 simply checks if the process is running or not
+    if [[ "$pid" =~ ^[0-9]+$ ]] && ! kill -0 "$pid" 2>/dev/null; then
+      return 1
+    fi
+  fi
+
+  return 0
 }
 
 # Gate on a pending — or to-be-verified — profile actually being open, not on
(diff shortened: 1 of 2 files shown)


joshuafouch:
Well if Copilot does not want to check this:
Here is what is outputted in the test:
```bash
> bash copy-url-shortcut-migration-test.sh
ok - migration defers the repair while the affected profile is open
ok - migration keeps the browser prompt visible
ok - migration ignores a browser on a different profile root
ok - migration asks to close the browser and repairs on confirmation
ok - migration rebinds the Copy URL shortcut to the pinned extension id
ok - migration is idempotent after the repair
ok - migration keeps remapped shortcut keys
ok - migration never double-binds the pinned extension
ok - migration stays pending when a browser starts mid-repair
ok - migration stays pending when a briefly-lived browser undoes the repair
ok - migration keeps an unverified repair pending while a browser runs
ok - migration verifies an attempted repair on a browser-free rerun
ok - migration leaves installed third-party extensions alone
```

danielgccr:
This is the right shape of fix — I hit the same deadlock on stable and worked through the same mechanism (details in #6866). Two gaps in the liveness test that I think are worth closing, since they leave the same class of false positive the PR is fixing:

**1. `kill -0` proves a PID exists, not that it's a browser.** The lock target encodes the PID Chromium held *at the time it was written*. After a reboot — the common way into this state — that number is very likely reassigned to an unrelated process, and `kill -0` then reports the profile as open forever. My stale lock was `Whale-4298`; nothing but luck kept PID 4298 unused on the next boot. Checking what the PID actually is would close it:

```bash
[[ -r /proc/$pid/comm ]] && grep -qiE 'chrom|brave|vivaldi|opera|msedge|helium' /proc/$pid/comm || return 1
```

**2. The hostname half of the target is discarded.** `"${target##*-}"` takes the PID and drops the host, so a lock written by a different machine on a shared or restored `$HOME` is judged against local PIDs. Comparing it to `$HOSTNAME` first, and treating a mismatch as stale, keeps that honest.

Also, minor: the `-e` (plain-file, non-symlink) branch still returns 0 with no liveness check at all. Chromium always writes a symlink, so this is theoretical rather than a real path — just noting it since the early return makes it easy to read as covered.

Neither point blocks the fix as-is; it's strictly better than the current `-L` test, which cannot distinguish stale from live at all because a Chromium `SingletonLock` target is a bare `hostname-pid` string that never resolves to a real file. Worth a look at `unverified_repairs_exist()` in the same pass, though: the migration never removes `Preferences.omarchy-copy-url-repair.bak`, even after a verified success, so those backups keep a repaired profile eligible for the same gate on any future rerun.

bjarneo:
Automated duplication check: this pull request looks similar to #6872, which covers the same Chromium singleton lock check. I keep that one open and close this one to consolidate review.

Issue #6024, opened 2026-06-02
Browser crash, probably from video autoplay, freezes Hyprland


hoclun-rigsep:
# Chromium/Brave SIGILL on video playback (NVIDIA) crashes Hyprland — hard reboot required

## System

| | |
|---|---|
| OS | Arch Linux |
| Kernel | 7.0.9-arch2-1 |
| CPU | AMD Ryzen 9 5900X |
| GPU | NVIDIA GeForce RTX 5070 |
| NVIDIA driver | nvidia-open-dkms 595.71.05-2 |
| libva-nvidia-driver | 0.0.17-1 |
| Chromium | 148.0.7778.178-1 |
| Brave | 1.92.108 beta |
| Hyprland | 0.55.2-1 |
| Omarchy | (current as of 2026-05-30 upgrade) |

## What happened

Playing video on social/commerce sites (confirmed: Facebook, Twitter, Amazon)
often causes Chromium or Brave to crash and Hyprland to freeze unrecoverably. I'm still
able to shell into the system but as for the physical console I can't even switch
to a different VT.

## Investigation

I shelled in to the system and asked Claude to attempt recovery and
investigate the cause. I rejected some implausible conclusions and now
present one which seems plausible to me (fwiw!).

The browser crashes with SIGILL in its internal Media decoder
thread. The crash leaves an NVDEC hardware decode session open without freeing
it. The NVIDIA GSP firmware then times out waiting for a cleanup RPC, enters
a faulted state, and Hyprland loses its GPU context and aborts. Hyprland's
watchdog restarts it in safe mode, but safe mode fails to bind its socket.
With Wayland mediating VT switching, there is no escape to a TTY.

This happened twice in one day: once with stock Chromium (Facebook),
once with Brave beta (Amazon), each time requiring a reboot.
I've already tried the "ignore the problem; maybe it will go away" solution
but am now filing this issue—it's hard to see how it could be affecting
only me.

## Relationship to existing issues

This is distinct from #2964 (AMD GPU ring timeout during Google Meet) and
#3891 (ANGLE/EGL video rendering broken after Mesa update). Those are GPU
rendering pipeline failures on AMD hardware. This is a crash in Chromium's
software-side media decode thread on NVIDIA hardware, which then wedges the
GPU driver.

The surface symptom — Chromium takes down Hyprland, reboot required — is the
same, but the root cause and affected hardware are different.

## Workaround (unconfirmed)

[...]

PR #7039, opened 2026-08-15
Support Firefox as a web app browser

### LINKED ISSUE #7034: launch webapp fails when default browser is Firefox (no Chromium installed)
## What happened

`omarchy launch webapp https://messenger.com` fails with:

```
error: path "--app=https://messenger.com" does not exist!
```

Expected: messenger.com opens as a web app (app mode) in the default browser.

## Reproduction

1. Install Firefox as the only browser and set it as default (`xdg-settings get default-web-browser` → `firefox.desktop`).
2. Run `omarchy launch webapp https://messenger.com`.

## Root cause

`/usr/share/omarchy/bin/omarchy-launch-webapp` whitelists only Chromium-family
browsers and falls back to `chromium.desktop` for anything else:

```bash
case $browser in
esac

exec setsid uwsm-app -- $(sed -n 's/^Exec=\([^ ]*\).*/\1/p' {~/.local,~/.nix-profile,/usr}/share/applications/$browser 2>/dev/null | head -1) --app="$1" "${@:2}"
```

With Firefox set as default, `chromium.desktop` doesn't exist, so the `sed`
extraction returns an empty binary path and the command degrades to:

```
uwsm-app -- --app="https://messenger.com"
```

`uwsm-app` then treats `--app=...` as the executable path → `error: path "... does not exist!"`.

Suggested fix: add Firefox to the supported browser list (Firefox supports
`--app=<url>`), and/or fall back to the actual default browser's desktop file
instead of hardcoding `chromium.desktop`.

## System details

- Omarchy: 4.0.0-1
- Arch x86_64, Intel Celeron N4020 (GeminiLake UHD 600)
- Default browser: `firefox.desktop`

alkevintan:
Closes #7034

## Problem

`omarchy launch webapp <url>` fails on systems where Firefox is the default
browser:

```
error: path "--app=https://messenger.com" does not exist!
```

`bin/omarchy-launch-webapp` only whitelists Chromium-family browsers
(`google-chrome`, `brave`, `microsoft-edge`, `opera`, `vivaldi`, `helium`)
for `--app=<url>` app mode and falls back to a hardcoded `chromium.desktop`
for everything else. On a Firefox-only install `chromium.desktop` doesn't
exist, so the `sed` extraction returns an empty binary and the command
degrades to `uwsm-app -- --app="<url>"`, which then errors on the `--app`
argument as if it were a path.

## Fix

- Add `firefox*` to the supported browser list. Firefox supports
  `--app=<url>` app mode.
- Resolve the browser binary up front and fail with a clear error instead of
  handing `uwsm-app` an empty executable when no `--app`-capable browser can
  be found (e.g. an unsupported default browser with chromium also missing).

## Testing

New `test/shell.d/launch-webapp-test.sh` covers:

- Firefox as the default browser → launches Firefox with `--app=<url>`.
- Unsupported default browser with chromium present → chromium fallback still
  works.
- Unsupported default browser with no chromium → clear error, no cryptic
  `uwsm-app` failure.

The existing shell-suite failures (bar widget timeout, missing `omarchy-pkgs`
checkout, missing compositor) are pre-existing environment issues and fail on
`quattro` without this change as well.

### DIFF
--- a/bin/omarchy-launch-webapp
+++ b/bin/omarchy-launch-webapp
@@ -6,8 +6,15 @@
 browser=$(xdg-settings get default-web-browser)
 
 case $browser in
-google-chrome* | brave* | microsoft-edge* | opera* | vivaldi* | helium*) ;;
+google-chrome* | brave* | microsoft-edge* | opera* | vivaldi* | helium* | firefox*) ;;
 *) browser="chromium.desktop" ;;
 esac
 
-exec setsid uwsm-app -- $(sed -n 's/^Exec=\([^ ]*\).*/\1/p' {~/.local,~/.nix-profile,/usr}/share/applications/$browser 2>/dev/null | head -1) --app="$1" "${@:2}"
+binary=$(sed -n 's/^Exec=\([^ ]*\).*/\1/p' {~/.local,~/.nix-profile,/usr}/share/applications/$browser 2>/dev/null | head -1)
+
+if [[ -z $binary ]]; then
+  echo "error: no browser installed that supports web app mode ($browser not found)" >&2
+  exit 1
+fi
+
+exec setsid uwsm-app -- "$binary" --app="$1" "${@:2}"
--- a/test/shell.d/launch-webapp-test.sh
+++ b/test/shell.d/launch-webapp-test.sh
@@ -58,7 +58,7 @@ grep -F '/usr/bin/chromium --app=https://messenger.com' "$launch_log" >/dev/null
   fail "webapp launcher keeps the chromium fallback for unsupported browsers" "launch: $(cat "$launch_log")"
 pass "webapp launcher keeps the chromium fallback for unsupported browsers"
 
-rm -f "$test_home/.local/share/applications/chromium.desktop"
+rm -f "$test_home/.local/share/applications/chromium.desktop" "$test_home/.local/share/applications/firefox.desktop"
 
 HOME="$test_home" PATH="$mock_bin:$PATH" BROWSER=epiphany.desktop \
   OMARCHY_TEST_WEBAPP_LAUNCH="$launch_log" \
--- a/test/shell.d/launch-webapp-test.sh
+++ b/test/shell.d/launch-webapp-test.sh
@@ -1,0 +1,70 @@
+#!/bin/bash
+
+set -euo pipefail
+
+source "$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd)/base-test.sh"
+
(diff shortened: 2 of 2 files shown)


Tsurgcom:
As much as I hate firefox, I believe this should be a thing.

JohnL404:
Thanks for working on this — I opened an alternative fix for the same issue in #9172 and wanted to share a technical finding about the Firefox part of this approach.

Firefox has no Chromium-style app mode. Its `-app` flag means something different: run a XUL application from a directory containing `application.ini`. Verified on Omarchy 4.0.1 with current Arch Firefox:

```
$ firefox --app /tmp/somedir
Couldn't read application.ini   (exit 1)

$ firefox --app=https://lumo.proton.me
(no error, no output — opens a regular Firefox session, no standalone app window)
```

So `firefox --app=<url>` doesn't produce a web app window; the launcher would silently open a normal Firefox session — the mocked test here can't catch that because it asserts the command line without running Firefox. Firefox removed site-specific browser (SSB) support back in Firefox 88, so there is no app-mode path to hook into.

That means for #7034's reported scenario (Firefox default, no Chromium installed), no `--app`-style launch is possible with Firefox. #9172 takes the route of falling back to any installed Chromium-family browser and exiting with a clear error when none exists. Happy to consolidate whichever direction the maintainers prefer.

Auxxed:
Support for the goal (Firefox as a webapp host), with a hardware reason it matters, and agreement that `--app=` is the wrong Firefox flag.

On Omarchy 4.0.2-1 my default browser is already Firefox, but `omarchy-launch-webapp` still forces Chromium. That is a problem on this NVIDIA box: RTX 3080 LHR, nvidia-open-dkms 610.57.04, BAR1 256 MiB. Chromium 151 renderer Media threads keep dying with SIGILL (ILL_ILLOPN) — four dumps in two days, including two today launched as Omarchy webapps (`chromium --app=https://x.com/`, `--app=https://omarchy.org/manual/`). Same class as #6024 / NVIDIA bug 5762513. `--disable-gpu-video-decode` is already in `chromium-flags.conf` and is not enough.

JohnL404 is right that Firefox 154 has no Chromium-style app mode (`--app` is XUL `application.ini`). Verified here too. Allowlisting Firefox and still passing `--app=<url>` will not produce an app window.

Please launch Firefox-family defaults with `--new-window <url>` instead (as in discussion #5490), rather than keep falling back to Chromium. Full machine details and the NVIDIA dumps are on that discussion: https://github.com/omacom/omarchy/discussions/5490#discussioncomment-18277712

PR #7040, opened 2026-08-15
feat(security): Add face authentication setup and removal commands

tyvsmith:
## Summary

- add `omarchy-setup-security-face [facelock]` and `omarchy-remove-security-face`, wrapping a face backend's own wizard
- write `/etc/pam.d/omarchy-lock-face` as a backend-neutral skeleton; the backend inserts its own `sufficient` line above `pam_deny`
- shape both scripts as a case over backends, `omarchy-install-browser` style: skeleton, service list and messages outside the case, `install_facelock` / `setup_facelock` / `configure_facelock_pam` / `remove_facelock` inside, commented howdy placeholders marking the second arm
- one backend at a time: setup refuses when the lock service already carries another module
- take no argument on removal; read the backend from the lock service, else from what is installed
- add Setup and Remove rows under Security, the Remove row gated on the service file rather than on any backend

## Why

Face authentication has been asked for since #4982, and #5212 was written against hyprlock so it no longer applies. @mateuszkowalczyk split the work three ways in #6863 and has the lock shell built; this is part two, touching nothing under `shell/`, so the two merge in either order and share only the service shape below. Omarchy is still choosing between facelock and Howdy, so the scripts take a backend name and only the facelock arm is written.

## The PAM contract

```
#%PAM-1.0
auth       sufficient                  pam_facelock.so
auth       required                    pam_deny.so
```

Omarchy writes the last two lines; the backend inserts the first. The terminator encodes how the shell consumes the service: password runs as a separate `PamContext`, so this lane has to end in a plain denial. A bare `auth required pam_facelock.so` hands the module's own code back (`PAM_IGNORE` on an empty frame) and the shell sees `PamResult.Error`; `sufficient` above `pam_deny` turns every non-match into `PAM_AUTH_ERR`.

## Reviewer notes

- **Each arm writes its own PAM line.** facelock's `pam add` validates every service before it writes any, refuses `system-auth`, backs up what it edits, and creates the `/etc/pam.d/polkit-1` override from `/usr/lib/pam.d` on current Arch. Shared code owns only the service list and the skeleton, so none of that is reimplemented here.
- **Removal is not from lines alone.** A refused `pam add` stops setup after the wizard has installed the daemon and enrolled a face, and setup names removal as the undo, so an installed backend with no line still gets torn down.
- **No hardware gate on the menu entry**, matching `setup.security.fido2`. An IR camera can need quirks before detection, and a user can relax `require_ir`; a probe here would hide the entry from both.
- **The polkit dialog has no face path yet** (part three of #6863).

## Validation

- every `/etc/pam.d/polkit-1` state on a libvirt guest: vendor-only, local file, re-run, symlink to the vendor file, symlink to `system-auth`, missing, empty, no auth line
- removal on the guest: configured, lock service deleted by hand, backend wired but binary gone, bare skeleton, nothing configured, and the undo after a refused setup
- setup refusing a second backend's line in the lock service; unknown backend name
- `facelock capabilities` on the released facelock-bin 0.2.0 reporting every feature the script requires
- face unlock end to end against #6863's lock shell, journal-confirmed
- `./test/cli`, `./test/shell`

Face auth is opt-in and in no package list.


mateuszkowalczyk:
Hi @tyvsmith, could you have a look at my earlier PR here: https://github.com/basecamp/omarchy/pull/6863
It already implements the lock-screen part and I managed to make it work quite smoothly not only after pressing Enter, but also after opening laptop lid, moving cursor or basically pressing any key.

Maybe you could build your setup process on top of that?

tyvsmith:
Read #6863. Dropped my lock-screen work; #7040 is now just your part two, no `shell/` files, so they merge in either order.

Worth knowing for your retry policy: facelock exempts no-face attempts from its rate limit, so scanning on every wake doesn't burn the budget.

Ran it end to end on a clean Omarchy 4 guest. Face unlock works.

## Current UX issues observed

- display blanks mid-burst; `idleBlankTimer` is 5s, gated on `authenticatingPassword` only
- a burst reports nothing: static glyph, untouched status `Text`
- scan and failure look identical to a frozen field

## Recommended followups

- gate the blank timer on face attempts too
- give face its own status slot, not `failureMessage`; that drives `errorState` and reddens the field for every passer-by
- update the PAM example in your description: bare `auth required pam_facelock.so` returns `Error`, not `Failed`, on a non-match
- consider the `pam_deny` skeleton Omarchy writes instead, so two backends can share the service
- part three (polkit dialog) is yours if you want it; `fingerprintMode` shows the sensor glyph while the camera scans

I'll open a PR against your branch shortly with a couple of the fixes.

*(Agent-drafted; claims verified.)*

palprateek:
@tyvsmith - thank you for defining the PAM contract here (Omarchy-owned `omarchy-lock-face` skeleton with `pam_deny` terminator, backend inserts `sufficient` line, one-backend-at-a-time, `facelock pam add --validate` before any writes).

I verified #11612 implements exactly this contract plus the lock UX from #6863, plus two machine-level fixes (Quickshell PamContext `abort_if_ssh=false`, systemd `ReadWritePaths=/etc/facelock`). It merge-trees clean on current `origin/quattro` with all shell/cli suites passing.

Would you consider marking #11612 as the successor to this Draft, or folding your multi-backend guard / safe-removal logic into it? Happy to test any rebase - I have Lenovo Slim 3 Syntek 174f:182b + TPM for verification.

PR #7042, opened 2026-08-15
Add Ctrl-N and Ctrl-P menu navigation

danielfain:
## Summary

- navigate down in the Omarchy menu with Ctrl+N
- navigate up with Ctrl+P
- retain the existing arrow-key behavior and add regression coverage

## Testing

- `qmllint -I shell shell/plugins/menu/Menu.qml`
- live UI verification with `wtype`: Ctrl+N moved the highlight from Apps to Learn, and Ctrl+P moved it back to Apps
- `./test/shell`: menu coverage passed; the run had four unrelated environment failures (`config-test.sh`, `snapper-test.sh`, and `unowned-system-paths-test.sh` require an `omarchy-pkgs` checkout; `network-qr-test.sh` failed its sandboxed Wi-Fi auto-detection case)

### DIFF
--- a/shell/plugins/menu/Menu.qml
+++ b/shell/plugins/menu/Menu.qml
@@ -1139,10 +1139,12 @@ Item {
           } else if ((event.key === Qt.Key_Backspace || event.key === Qt.Key_Left) && !root.filterText) {
             root.goBack()
             event.accepted = true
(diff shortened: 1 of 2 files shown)


joshuafouch:
I second this.

PR #7062, opened 2026-08-16
Support DoT endpoints in `omarchy dns Custom`

ujo4eva:
## Summary

The `omarchy dns Custom` flow is currently IP-only: it writes plain IPs to NetworkManager and `resolved.conf`, so there is no way to point at an encrypted custom endpoint that needs a hostname. This change lets you type, Android-style, **an IP or a DoT URL** and have it "just work":

```
omarchy dns Custom
Enter DNS servers (space-separated): plain IPs, tls://host, IP#host, or a bare hostname for DoT:
> 1a2b3c4d.d.adguard-dns.com
```

Accepted inputs (any mix, space/comma separated):
- plain IPs — existing behavior, unchanged
- `tls://host` or a bare hostname — DNS-over-TLS, resolved to IPs at config time
- `IP#host` — explicit SNI
- `https://` / `quic://` — rejected with a hint to use the AdGuard CLI or dnsproxy (systemd-resolved cannot speak DoH/DoQUIC)

## Why

Custom endpoints that carry a per-device identifier in the hostname (AdGuard DNS private device IDs, NextDNS, self-hosted AdGuard Home with a wildcard cert) need the hostname sent as the TLS SNI. `resolved.conf` supports this via `IP#hostname`, but there was no way to configure it from the applet/CLI. This closes that gap while preserving the plain-IP behavior exactly.

## How it works

When any hostname is present, it writes a drop-in (`/etc/systemd/resolved.conf.d/90-dot.conf`) with `DNS=IP#host …` + `DNSOverTLS=opportunistic`, and:
- clears per-connection NetworkManager DNS (`ignore-auto-dns yes`, empty `dns`) so the SNI-bearing global drop-in actually serves each link — per-link plain IPs would drop the SNI and break cert validation for custom endpoints
- removes a stale NM global-dns file so plain servers don't merge back into the global list
- comments out a stale `DNS=` in base `resolved.conf` (resolved merges `DNS=` across base + drop-ins)
- `DHCP` now removes the drop-in so the provider toggle/status reads correctly
- the no-arg status read (`current_dns_provider`) now checks the drop-in, so the applet shows `Custom` instead of `DHCP` while a DoT custom endpoint is active

## Notes

- Endpoint IPs are pinned at config time (resolved requires IPs, not hostnames); re-run if they change — same model as the Cloudflare/Google presets.
- Verified live end-to-end (write, reload, `resolvectl status`, DHCP cleanup) plus a parser unit pass.
- `./test/cli` passes. The 3 failing `./test/shell` cases (`config-test`, `snapper-test`, `unowned-system-paths-test`) require a separate `omarchy-pkgs` checkout and are unrelated to this change.

### DIFF
--- a/bin/omarchy-dns
+++ b/bin/omarchy-dns
@@ -19,6 +19,7 @@ if (( EUID == 0 )); then
 fi
 
 NM_DNS_CONF=/etc/NetworkManager/conf.d/20-omarchy-dns.conf
+RESOLVED_DROP_IN=/etc/systemd/resolved.conf.d/90-dot.conf
 
 provider_from_arg() {
   case "${1:-}" in
@@ -97,11 +98,30 @@ resolved_dns() {
   ' /etc/systemd/resolved.conf 2>/dev/null || true
 }
 
+resolved_dropin_dns() {
+  [[ -f $RESOLVED_DROP_IN ]] || return 0
+
+  awk -F= '
+    /^[[:space:]]*#/ { next }
+    /^[[:space:]]*\[Resolve\]/ { in_resolve = 1; next }
+    /^[[:space:]]*\[/ { in_resolve = 0 }
+    in_resolve && /^[[:space:]]*DNS[[:space:]]*=/ {
+      value=$0
+      sub(/^[^=]*=/, "", value)
+      print value
+      exit
+    }
+  ' "$RESOLVED_DROP_IN"
+}
+
 current_dns_provider() {
   local dns=""
   local compact=""
 
   dns=$(networkmanager_global_dns)
+  if [[ -z $(printf '%s' "$dns" | tr -d '[:space:],') ]]; then
+    dns=$(resolved_dropin_dns)
+  fi
   if [[ -z $(printf '%s' "$dns" | tr -d '[:space:],') ]]; then
     dns=$(resolved_dns)
   fi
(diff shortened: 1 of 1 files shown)


ujo4eva:
Reviewed the change against correctness, readability, and merge behavior.

One real correctness bug found and fixed in `43176bc3`: the Cloudflare, Google, and plain-IP Custom branches never removed the DoT drop-in (`90-dot.conf`). Since systemd-resolved merges `DNS=` from the main file *and* drop-ins, a previously configured custom DoT endpoint stayed in the global server list alongside the new provider after a switch — `current_dns_provider` would report the new provider while the old endpoint kept resolving. DHCP already removed it; the other three now do too.

Remaining observations (non-blocking):
- `is_ipv6()` treats any token containing `:` as IPv6, so an entry like `1.1.1.1:853` is classified as IPv6. systemd-resolved accepts `IP:port`, so it still works; just semantically loose.
- `disable_base_dns()` only disables the first `DNS=` line; resolved allows repeated lines. Preexisting shape, low risk.
- `parse_entries()` and the DoT drop-in path verified against the prototype used to validate `IP#host`, bare-hostname, `tls://`, and mixed lists on a live system.

`./test/cli` passes (0 failures).

amanat361:
+1 from a NextDNS user. Right now the only way to keep a per-profile DoT endpoint (`45.90.28.0#xxxxxx.dns.nextdns.io`) on v4 is to hand-edit `resolved.conf`, and the panel then just shows it as "Custom" with no way to re-select it after touching any other button. `IP#host` support in Custom would close that gap for NextDNS and AdGuard alike. Happy to test on a systemd-networkd/iwd box.

ujo4eva:
Rebased onto current `quattro` (`d3d23fdd`). GitHub had this as `DIRTY`; it now merges cleanly (`MERGEABLE`). Remaining `BLOCKED` is review, not conflicts.

The one conflict was in `bin/omarchy-dns`: quattro had added the privileged-heredoc annotation immediately above the Custom `resolved.conf` write, and this PR rewrote that same region into `parse_entries` / `dot_mode`. Resolution kept the DoT parser and put the annotation on the plain-IP `else` branch. The new `write_dot_drop_in` heredoc got the same `paths=none` annotation so the scanner stays green.

Behavior is unchanged from the original PR:
- hostname / `tls://host` / `IP#host` → DoT drop-in with `DNSOverTLS=opportunistic` and SNI
- bare IPs → the existing Custom path, still without `DNSOverTLS=`

That last point is why this stays complementary to #6513 rather than a replacement. The #8921 input (`194.242.2.4 2a07:e340::4`) still takes the plaintext branch here; #6513's one-liner is what covers that case.

Checked locally after the rebase: `bash -n bin/omarchy-dns`, `privileged-heredoc-test`, `dns-sudoers-test`, `network-manager-transition-test`, and `./test/cli`.

PR #7071, opened 2026-08-16
Migrate Brave Origin Beta profile data to stable

jakiurcore:
## Summary
`migrations/1784510887.sh` switches the Brave package from `brave-origin-beta-bin` to `brave-origin-bin`, but leaves the user's profile data behind in `~/.config/BraveSoftware/Brave-Origin-Beta`. Stable Brave reads from `~/.config/BraveSoftware/Brave-Origin` instead, so it comes up as a fresh install — no bookmarks, history, passwords, cookies, extensions, or additional profiles. Observed on a real machine: a 1.6 GB beta directory (8 profiles) sitting next to a 115 MB freshly-initialized stable directory with just one.

This adds a **new** migration rather than editing `1784510887.sh` — everyone affected has already run it, so editing it fixes nobody.

- Moves (not copies) the whole beta directory in one `mv`: same filesystem, atomic, no extra disk space needed. `Local State` maps profile directories to display names and per-profile prefs depend on it, so the two directories must never be merged.
- If a stable directory already exists, it's renamed to a timestamped `<dir>.bak.<datetime>` sibling rather than deleted — swapping the two back restores the pre-migration state exactly.
- Skips (leaving both directories untouched) if either profile looks currently open, checked the same way `migrations/1786643346.sh` does: a `SingletonLock` (plain file or dangling symlink) or `SingletonSocket`inside the profile directory. Prints instructions to quit Brave and re-run `omarchy-update`.
- No-ops cleanly when the beta directory is absent (never installed, or already migrated).

## Testing
Verified all four directory-state combinations (beta+stable, beta-only, stable-only, neither) plus the running-browser skip, under a synthetic `$HOME`.

Also ran it for real on my own machine against a genuine beta profile (1.6 GB,8 profiles: WorkStation, YT, SH Linkedin, Python Course, Kimi Test, TK Vai,play store, surfshark vpn). Worth flagging: on this machine a real running Brave process left **no** `SingletonLock`/`SingletonSocket` files in its profile directory at all — the lock-file guard alone didn't detect it as running. I confirmed via `pgrep`/`ps` that Brave was fully quit before running the migration for real, so nothing was corrupted, but the lock-file check inherited from `1786643346.sh` may not be sufficient on its own for every Brave build/desktop combination. After the real run: all 8 profiles, their display names, bookmarks, history, and extensions came across correctly, and Brave started up normally afterward with all cookies and data.

### DIFF
--- a/migrations/1786851224.sh
+++ b/migrations/1786851224.sh
@@ -13,6 +13,7 @@ if [[ -d $beta ]]; then
   if profile_open "$beta" || { [[ -d $stable ]] && profile_open "$stable"; }; then
     echo "Brave is running — skipping profile migration." >&2
     echo "Quit Brave completely, then run: omarchy-update" >&2
+    exit 1
   else
     if [[ -d $stable ]]; then
       backup="$stable.bak.$(date +%Y%m%d%H%M%S)"
@@ -22,7 +23,7 @@ if [[ -d $beta ]]; then
 
     mv "$beta" "$stable"
     rm -f "$stable"/Singleton*
-    rm -rf ~/.cache/BraveSoftware/Brave-Origin-Beta
+    rm -rf "$HOME/.cache/BraveSoftware/Brave-Origin-Beta"
 
     echo "Beta profile migrated. Previous stable data (if any) kept as a .bak directory."
   fi
--- a/migrations/1786851224.sh
+++ b/migrations/1786851224.sh
@@ -17,11 +17,14 @@ if [[ -d $beta ]]; then
   else
     if [[ -d $stable ]]; then
       backup="$stable.bak.$(date +%Y%m%d%H%M%S)"
-      mv "$stable" "$backup"
+      mv -T "$stable" "$backup"
       echo "Existing stable profile preserved at $backup"
     fi
 
-    mv "$beta" "$stable"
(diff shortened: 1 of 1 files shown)


jakiurcore:
Thanks for the review and for pushing `ca60e148` directly — confirmed the `mv -T` fix, agreed that was a real bug.

On the other two points: agreed on both, and I don't think either should be patched here independently right now.

- The `exit 1` halting the whole `omarchy-migrate`/`omarchy-update` run is real, but it's the same detection mechanism `1786643346.sh` uses, and that file already has two competing open PRs (#7026, #6872) fixing the underlying stale-lock/liveness problem differently. Rather than invent a third approach here, I'll wait for the maintainers to pick one and port that exact technique into this migration.
- The fail-open case (a real running Brave leaving no `SingletonLock`/`SingletonSocket` at all, which I hit on my own machine) is the same ambiguity — whichever detection method wins for the sibling file should apply here too.

Given `Brave-Origin-Beta`'s remediation path is less recoverable than the Copy URL migration's (the beta binary is normally already uninstalled by the time this runs, so nothing can relaunch against it to clear a stale lock), this migration may end up needing something slightly different even after that decision lands — happy to iterate once the direction is settled rather than guess now.

Issue #7012, opened 2026-08-15
Menu launcher lost Walker's web-search fallback for no-match queries


Balmisjutas:
Title: Menu launcher lost Walker's "search the web" fallback for no-match queries

Before Quattro, Walker's `websearch` provider (via elephant) meant typing a
query with no matching app/setting would still let you hit Enter and get a
web search in your default browser. The new native Omarchy menu (Super+Space)
searches apps/settings/menu entries fine, but there's no fallback at all when
nothing matches — Enter just does nothing.

Checked the actual `Menu.qml`/`MenuModel.js` source to confirm this isn't
just undocumented: there's no `websearch`, `google`, or similar string
anywhere in `/usr/share/omarchy/shell/plugins/menu/`. It looks like an
omission from the Walker->native rewrite rather than an intentional drop.

I patched it locally (as a cloned `omarchy.menu` plugin, so it's easy to
test without touching the packaged copy) and it's a small, surgical change:
in `rebuildDisplay()`, when a search query produces zero rows, synthesize
one fallback row whose action shells out to `omarchy-launch-browser` (the
existing helper that already respects the user's actual default browser via
`xdg-settings`, not any specific browser) with a Google search URL for the
typed query.

Diff against `shell/plugins/menu/Menu.qml`:

```diff
@@ -602,6 +602,29 @@
         for (var d = 0; d < drilldownRows.length; d++) drilldownRows[d].section = "drilldown"
       }
       rows = currentRows.concat(drilldownRows)
+
+      // No app/setting/menu entry matched — fall back to a web search, the
+      // way Walker's "websearch" elephant provider used to.
+      if (rows.length === 0) {
+        rows = [{
+          itemId: "lemec.websearch",
+          kind: "action",
+          icon: "󰖟",
+          iconFont: "",
+          appIcon: "",
+          appId: "",
+          label: "Search Google for “" + query + "”",
+          target: "",
+          detail: "",
+          path: "",
+          childCount: 0,
+          action: "omarchy-launch-browser " + Util.shellQuote("https://www.google.com/search?q=" + encodeURIComponent(query)),
+          provider: "",
+          score: 0,
+          section: ""
+        }]
+      }
     } else {
       for (var j = 0; j < root.itemOrder.length; j++) {
         var child = root.item(root.itemOrder[j])
```

Tested live: reloaded the shell with the cloned plugin, searched a nonsense
string, confirmed the fallback row renders, hit Enter, confirmed it opens
[...]

PR #7075, opened 2026-08-16
menu: add web search fallback for unmatched queries

armyourself:
When a search in the app menu matches nothing, show a fallback row that opens the query in the user's default browser via Google, instead of presenting an empty list the user has to Esc out of.

The fallback row:
- Shows a search icon and "Search for \"query\" in browser"
- Activates via `omarchy-launch-browser`, which uses the XDG default browser
- Query is URL-encoded and shell-quoted for safe handling of special characters

```
# To test:
1. Open the menu
2. Type a query with no matches (e.g. "zzzzz")
3. Confirm the fallback row appears
4. Press Enter — Google opens in your default browser
```

### DIFF
--- a/shell/plugins/menu/Menu.qml
+++ b/shell/plugins/menu/Menu.qml
@@ -634,18 +634,42 @@ Item {
         else drilldownRows.push(row)
       }
 
-      var searchSort = function(a, b) {
-        if (a.score !== b.score) return a.score - b.score
-        return a.path.localeCompare(b.path)
-      }
+      if (currentRows.length === 0 && drilldownRows.length === 0) {
+        // Nothing matched: offer to hand the query to the default browser
+        // instead of returning an empty list the user has to Esc out of.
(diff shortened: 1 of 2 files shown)


JmZacapaDev:
Hey @armyourself, really nice work on this! I came at the same problem from a different angle and opened #8012 before I found yours. I genuinely liked the "hand the query to the default browser" framing, and the URL-encoded + shell-quoted handling is spot on.

Mine goes a little further in a few places:
- The fallback row shows on **every** search (not only zero matches), so you can always jump to a web search.
- It tells URLs apart from plain text — `example.com`, `localhost:3000`, `mailto:…` open directly, anything else searches — and the search engine is configurable via `menu.webSearchUrl` in `shell.json` (Google by default).
- The URL-vs-search logic lives in `MenuModel.js` with regression tests, and it also closes #7012.

Happy to coordinate. 🙂

armyourself:
Glad to hear that!

PR #7082, opened 2026-08-16
Add OpenCode Go usage collector with rate limits to the agents plugin

tharinfernando:
Adds OpenCode Go (`opencode-go`) as a fourth subscription in the Agents panel,
alongside Claude Code, Codex, and Fireworks. The agents panel is display-only
and discovers any record in `~/.local/state/omarchy/agents/usage/`, so this is
a new collector plus its mark, docs, and tests — no plugin changes needed.

## Changes

- **`bin/omarchy-agent-usage-opencode`** — new collector that:
  - Reads local token stats from opencode's own database
    (`~/.local/share/opencode/opencode.db`), filtered to messages that ran on
    the `opencode-go` provider (same schema handling, scan cache, and
    `--force` / `--limits-only` flags as the other collectors).
  - Fetches rate limits from the OpenCode Go usage endpoint
    (`https://opencode.ai/zen/go/v1/usage`) using the opencode-go API key from
    opencode's auth store — the rolling 5-hour, weekly, and monthly dollar
    windows render as the panel's LIMITS meters.
  - Degrades gracefully when the endpoint is unreachable or no key is set:
    empty limits + a status note, with local token usage still shown.
- **`shell/plugins/agents/assets/opencode-go.svg`** — opencode mark for the tab.
- **`shell/plugins/agents/manifest.json`** — enables `opencode-go` by default.
- **`shell/plugins/agents/README.md`** — documents the collector and its limits.
- **`test/shell.d/agent-usage-opencode-scanner-test.sh`** — 10 assertions:
  provider filtering, token/model aggregation, session counting, limit window
  mapping (endpoint mocked via `OPENCODE_GO_USAGE_URL`), and the offline
  fallback. Existing agent usage/panel tests still pass.

## Notes

- OpenCode Go is a subscription, not prepaid credits, so it reports rate
  limits but has no balance — no `balance` field.
- `OPENCODE_GO_USAGE_URL` is honored so the limits probe can be pointed at a
  mock endpoint in tests.

### DIFF
--- a/bin/omarchy-agent-usage-opencode
+++ b/bin/omarchy-agent-usage-opencode
@@ -1,0 +1,284 @@
+#!/usr/bin/python3
+# omarchy:summary=Print the OpenCode Go usage record as JSON
+# omarchy:args=[--force] [--limits-only]
+"""Collect OpenCode Go usage into one display-ready JSON record.
+
+Local stats come from opencode's per-message provider/model/token records in
+~/.local/share/opencode/opencode.db, filtered to messages that ran on the
+opencode-go provider. OpenCode Go exposes no local rate-limit or balance API,
+so the record carries token usage (today, the last week, and all time, broken
+down by model) with an empty limits list -- the same shape a Codex install
+with no account reports.
+
+This collector is a user-local companion to the packaged omarchy agent usage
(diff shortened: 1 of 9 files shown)


bjarneo:

Issue #6995, opened 2026-08-15
Idle cycle restarts during a pending lock and launches a second screensaver into it — new ttfx aborts (isLocked guard passes while lock-pending)


s373nZ:
**What happens.** When the idle lock fires, `omarchy-system-lock` tears the running screensaver down correctly (the #6764 ordering). Destroying the screensaver window makes the compositor report idle again about a second later, the idle service begins a **brand-new idle cycle**, and because `screensaverDelaySeconds == 0` that cycle launches a **second** screensaver immediately — into a session that is already locking. Its `ttfx` is born inside the lock window and aborts with the panic from #6762 as soon as its terminal goes away.

This is a third race, distinct from the two already in #6762. There the screensaver exists *before* the lock and the question is teardown ordering (#6764) or the supervisor loop respawning a renderer (@Unayung). Here the entire screensaver — terminal *and* renderer — is created **after** `omarchy-system-lock` has already run its `pkill`s. No kill-ordering or bounded-wait fix in the lock script can prevent it, because at cleanup time there is nothing to kill and nothing to wait for.

**Timeline** (Omarchy 4.0.0-1, ttfx 0.3.1-1, foot, `idle: {screensaver: 150, lock: 300}`):

```
14:58:43.279  idle-monitor: idle
14:58:43.279  idle-cycle-start: screensaver=150 lock=300
14:58:43.279  process-start: screensaver            <- screensaver #1
14:58:44.288  process-exit: screensaver exitCode=0     (launcher ran 1.0s: it spawned)

15:01:13.423  lock-system: lock-timeout
15:01:13.428  process-start: lock omarchy-system-lock  <- kills screensaver #1

15:01:14.406  idle-monitor: idle                     <- compositor re-reports idle
15:01:14.406  idle-cycle-start: screensaver=150 lock=300
15:01:14.406  process-start: screensaver            <- screensaver #2, into the pending lock
15:01:14.462  lock-requested
15:01:14.462  lock-pending: screen-stabilizing
15:01:15.___  foot maps; ttfx PID 7320 starts
15:01:15.561  secure=true
15:01:15.6    ttfx 7320 SIGABRT
15:01:15.622  process-exit: lock exitCode=0
15:01:15.925  process-exit: screensaver exitCode=0     (launcher ran 1.5s: it spawned a terminal)
15:01:17.409  idle-cycle-cancel: screensaver-not-running
```

Same panic, recovered verbatim from the core:

```
[...]

PR #7102, opened 2026-08-16
Only treat lost focus as a dismissal once the screensaver has held focus

### LINKED ISSUE #6917: Screensaver opens as a small floating window and self-terminates after 1s when a bar panel is open
# Screensaver opens as a small floating window and self-terminates after 1s when a bar panel is open

## Summary

When any bar widget panel (e.g. the power/battery dropdown) is open at the moment the idle
timer fires, the screensaver misbehaves in two ways at once:

1. It does **not** go fullscreen — it appears as a small floating terminal window in the
   middle of the screen, at foot's default 700x500.
2. It **kills itself after ~1.05 seconds**, without any user input.

The visible result is a small box that pops up, animates briefly, and disappears on its own.
The idle service logs this as `screensaver-dismissed` and cancels the pending lock, so **the
session also fails to lock** on that cycle.

Both symptoms trace to the panel holding layer-shell keyboard focus.

## Environment

| | |
|---|---|
| Omarchy | 4.0.0-1 |
| Hyprland | v0.56.2 |
| Quickshell | 0.3.0 (`28771c7c`, AUR `quickshell-git`) |
| foot | 1.27.0 |
| ttfx | 0.3.1 |
| Kernel | 6.19.8-arch1-3-surface |
| CPU | Intel i7-1185G7 (Tiger Lake) |
| GPU | Intel Iris Xe (TigerLake-LP GT2) |
| Display | single eDP-1, 2880x1920, `scale = 2` |
| Terminal | foot (`xdg-terminal-exec --print-id` → `foot.desktop`) |

`idle.screensaver = 150`, `idle.lock = 300` in `~/.config/omarchy/shell.json`.

## Steps to reproduce

Deterministic, no waiting for the idle timer:

```bash
# 1. Open any bar panel built on KeyboardPanel
omarchy-shell omarchy.power open

# 2. Launch the screensaver the same way the idle service does
omarchy-launch-screensaver force
```

**Result:** a small floating terminal appears centered, runs the ttfx animation, and exits by
itself after about one second.

**Control** — the identical command with the panel closed produces a correct fullscreen
screensaver that persists:

```bash
omarchy-shell omarchy.power close
omarchy-launch-screensaver force
```

In normal use this happens whenever you leave the battery/power dropdown open and walk away.

## Expected vs actual

**Expected:** the screensaver covers the screen and stays up until real user input, then the
lock fires at `idle.lock`.

**Actual:** a small floating box appears for ~1s and self-terminates. The idle service reads
the window closing as a user dismissal, cancels the cycle, and the machine never locks.

## Evidence

Window state sampled every 100ms via `hyprctl clients -j`, same command, only the panel state
differs:

| | panel closed (control) | panel open |
|---|---|---|
| `fullscreen` | `2` | **`0`** |
| `size` | `[1440, 960]` (full screen) | **`[700, 500]`** (foot default) |
| `floating` | `true` | `true` |
| lifetime | persisted past 2.6s | **1.08s**, exited on its own |

Focus sampled every 50ms across a naturally-occurring failure (window opened 22:40:34.937):

```
34.944 NOTFOCUS  class=foot   <- 7ms after map; screensaver never got focus
...    (167/167 samples NOTFOCUS class=foot)
35.982 idle-cycle-cancel: screensaver-dismissed
```

Lifetime 34.937 → 35.982 = **1.045s**.

Across five separate naturally-occurring failures, window-open → dismissal was
**1.046, 1.044, 1.228, 1.235, 1.235s** — each one exactly 1.000s (the `read -n1 -t 1`
timeout) plus a short `wait_for_terminal_resize` prefix. No human input was involved.

## Root cause

Bar panels are built on `KeyboardPanel`, which takes layer-shell keyboard focus while open
(`shell/Ui/KeyboardPanel.qml:98`; the power panel instantiates it at
`shell/plugins/panels/power/Panel.qml:292`):

```qml
WlrLayershell.keyboardFocus: open
  ? (focusPrimed ? WlrKeyboardFocus.OnDemand : WlrKeyboardFocus.Exclusive)
  : WlrKeyboardFocus.None
```

That single condition breaks the screensaver twice.

### 1. The fullscreen window rule never applies

With the panel holding focus, the newly mapped screensaver window does not become the active
toplevel, so `fullscreen = true` (`default/hypr/apps/system.lua:35`) does not take effect. The
adjacent `float = true` (line 36) *does*, leaving a floating window at foot's default 700x500 —
the small centered box.

### 2. The watchdog mistakes "never had focus" for "user dismissed me"

`bin/omarchy-screensaver:5`:

```bash
screensaver_in_focus() {
  hyprctl activewindow -j | jq -e '.class == "org.omarchy.screensaver"' >/dev/null 2>&1
}
```

`hyprctl activewindow` only reports **toplevel** windows — layer surfaces are invisible to it.
With the panel focused it returns the last focused toplevel (here `foot`), so the check returns
false even though nothing the user did caused it.

At the first tick of the watch loop (`bin/omarchy-screensaver:44`):

```bash
  exit_screensaver
fi
```

`read` times out at 1.000s, `screensaver_in_focus` is false, and the screensaver `pkill`s
itself (`bin/omarchy-screensaver:12`). The idle service then observes `closewindow`, treats it
as `screensaver-dismissed`, and cancels the pending lock.

The check has no way to distinguish *lost* focus from *never acquired* focus, and no awareness
of layer-shell focus.

## Suggested fixes

Any one of these addresses it; the first two are the smallest:

- **Close open bar panels before launching the screensaver**, or skip the launch while a
  keyboard-focus panel is open (the idle service already shells out, so this is one more call).
- **Don't run the focus check until the screensaver has been focused at least once.** Latch a
  "was focused" flag and only treat focus loss as dismissal after that, which also removes the
  fixed 1s window where a slow-mapping window can be killed.
- **Make the focus check layer-aware**, e.g. treat "no active toplevel / a layer surface holds
  focus" as not-a-dismissal rather than as a dismissal.
- Consider making the screensaver a layer-shell overlay rather than a toplevel terminal window,
  which sidesteps both the fullscreen-rule race and the focus check entirely.

## Additional note

`exit_screensaver` uses `pkill -f '[o]rg.omarchy.screensaver'` (`bin/omarchy-screensaver:12`).
The `[o]` guard only protects against matching the `pgrep`/`pkill` process itself — it still
matches *any* process whose command line contains that string. Scripts or tooling that mention
the class by name (including diagnostics for this very bug) get killed as collateral. Matching
on PID or window address would be safer.

omarchybot:
The screensaver's watchdog exits when `hyprctl activewindow` stops reporting the screensaver's class, which is meant to catch the user clicking away onto another window. `hyprctl activewindow` only ever reports toplevels, so it cannot see a layer surface. When a bar panel is open it holds layer-shell keyboard focus (`shell/Ui/KeyboardPanel.qml:98`) and the newly mapped screensaver never becomes the active toplevel at all -- the panel leaves the previously focused terminal sitting there as `activewindow`. `bin/omarchy-screensaver:44` then reads "never acquired focus" as "the user dismissed me" and the screensaver kills itself on the first tick of the loop, about one second after it appeared. That matches the lifetimes the reporter measured across five failures (1.046, 1.044, 1.228, 1.235, 1.235s): the 1.000s `read -n1 -t 1` timeout plus the `wait_for_terminal_resize` prefix.

The damage is not just the flash of a window. The idle service watches for the screensaver's window closing and treats it as activity: `handleScreensaverWindowClosed` cancels the idle cycle with `screensaver-dismissed` (`shell/plugins/services/idle/Service.qml:138`), and `cancelIdleCycle` stops the pending lock timer. Leaving the battery dropdown open and walking away therefore leaves the machine idle and unlocked.

This latches a flag the first time the screensaver is genuinely focused, and only counts the absence of focus as a dismissal after that. Sampling focus before the blocking read rather than after it means a normal launch latches immediately, so clicking away is still detected within the same second as before.

This deliberately fixes only half of #6917. With a panel open the `fullscreen` window rule also fails to apply and the screensaver draws as a floating 700x500 window. `default/hypr/apps/system.lua:35` registers that rule through Hyprland's own `hl.window_rule` (via the thin wrapper at `default/hypr/helpers.lua:142`) and Omarchy has no other fullscreen path, so the failure is inside the compositor and not fixable here. What is left after this change is a screensaver that stays up, keeps the lock timer armed, and locks on schedule -- but is still a small floating box on that cycle. Making it go fullscreen regardless, or closing open panels before launching, is a call for you rather than a defect to correct.

Not a duplicate of #6995, which needs a lock already in flight, launches a second screensaver into it, and ends in a ttfx `SIGABRT` with a core dump. This one needs no lock at all and leaves no core dump.

Touches the same loop as the open #6813, which keeps the `read -n1 -t 1 || ! screensaver_in_focus` condition verbatim and so does not fix this; whichever lands second will need a small rebase.

Fixes #6917

### DIFF
--- a/bin/omarchy-screensaver
+++ b/bin/omarchy-screensaver
@@ -6,6 +6,14 @@ screensaver_in_focus() {
   hyprctl activewindow -j | jq -e '.class == "org.omarchy.screensaver"' >/dev/null 2>&1
 }
 
+# `hyprctl activewindow` only ever reports toplevels. A layer surface holding
+# keyboard focus -- an open bar panel -- leaves the previously focused terminal
+# as the active window, so a screensaver that maps behind one never becomes
+# active at all. Only a screensaver that has actually held focus can have lost
+# it, so latch that first and treat the absence of focus as a dismissal only
+# afterwards.
+was_focused=0
+
 exit_screensaver() {
   hyprctl eval 'hl.config({ cursor = { invisible = false } })' &>/dev/null || hyprctl keyword cursor:invisible false &>/dev/null || true
   pkill -x ttfx 2>/dev/null
@@ -41,7 +49,13 @@ while true; do
     --random-effect --no-eol --no-restore-cursor &
 
   while pgrep -t "${tty#/dev/}" -x ttfx >/dev/null; do
-    if read -n1 -t 1 || ! screensaver_in_focus; then
+    if screensaver_in_focus; then
+      was_focused=1
+    elif ((was_focused)); then
+      exit_screensaver
+    fi
+
+    if read -n1 -t 1; then
       exit_screensaver
     fi
   done


EduardsSk:
The latch fixes the panel case, but the same watchdog has a second trigger it does not cover: a multi-monitor launch, where the screensaver **does** hold focus first and then loses it while `omarchy-launch-screensaver` is still working.

`omarchy-launch-screensaver` iterates monitors, focusing each one (`hypr_focus_monitor`) and waiting for that monitor's window (`wait_for_screensaver_window`) before moving on. While it spawns monitor N+1, the instance already running on monitor N sees monitor N+1's previously focused toplevel as `activewindow`. With this PR applied that instance has already set `was_focused=1` on its first tick, so the second tick takes the `elif` branch and calls `exit_screensaver` — which `pkill`s ttfx and every `org.omarchy.screensaver` window, taking the whole set down.

Captured from Hyprland's event stream on a three-output session, launched from the menu entry (`omarchy-launch-screensaver force`), with no user input at any point:

```
focusedmon>>eDP-1,3
activewindow>>org.omarchy.btop,btop
openwindow>>…,3,org.omarchy.screensaver,Ghostty
activewindow>>org.omarchy.screensaver,Ghostty    <- instance 1 holds focus; was_focused=1
focusedmon>>DP-5,1
activewindow>>md.obsidian.Obsidian,…             <- instance 1 now sees a foreign toplevel
openwindow>>…,1,org.omarchy.screensaver,Ghostty
activewindow>>org.omarchy.screensaver,Ghostty
focusedmon>>DP-6,2
activewindow>>com.mitchellh.ghostty,…            <- and again
closewindow>>…                                   <- whole set gone, ~3.7s after launch
closewindow>>…
```

Whether it fires depends on how long the launch takes relative to the 1s `read -n1 -t 1` tick, which is presumably why single-output reporters only see the panel variant. Spawn-to-mapped for the same class on this machine:

| terminal | spawn-to-mapped | × 3 outputs |
| --- | --- | --- |
| foot | 0.18s | ~0.6s |
| alacritty | 0.28s | ~0.9s |
| ghostty | 0.50s (warm) | ~2.7s in a real launcher run (~0.85s per monitor including the per-monitor wait) |

The consequence is the one described in the second half of #6917: `handleScreensaverWindowClosed` reads the closes as a dismissal, `cancelIdleCycle` stops the pending lock, and the session does not lock. One day of this machine's journal: 5 cycles cancelled as `screensaver-dismissed`, 3 of them 2.3–3.3s after `idle-cycle-start` — too fast to be a human — with no lock on those cycles. (The other two came 73s and 110s in and may well be genuine dismissals.) The cycles that *did* lock are the ones where the launcher returned in under a second without spawning anything.

A complement that would close this: don't enforce the focus check until every output has a screensaver window — compare `hyprctl clients -j` filtered on the class against `hyprctl monitors -j | jq length`, with a deadline so a failed spawn falls through to current behaviour, and a short `read` inside the wait so a keypress still dismisses.

Fully aware #7193 makes this moot if it lands, since one exclusive overlay per output removes both the terminal spawn sequence and the `activewindow` watchdog. Posting mainly so the multi-monitor trigger is on record in case the latch ships first.

<details>
<summary>Environment</summary>

| | |
|---|---|
| Omarchy | 4.0.0-1 |
| Hyprland | 0.56.2 |
| Quickshell | 0.3.0.r20.g28771c7-1 |
| ttfx | 0.3.1-1 |
| ghostty | 1.3.1-2 (`xdg-terminal-exec --print-id` → `com.mitchellh.ghostty.desktop`) |
| Kernel | 7.1.8-arch1-3 |
| Outputs | eDP-1 1920x1200, DP-5 2560x1440, DP-6 3840x2160 |

</details>

PR #7143, opened 2026-08-16
fix: respect OpenCode v2 next pins in mise wrappers

mattstyles333:
## Summary

- Point the OpenCode lazy stub at `npm:@opencode-ai/cli` instead of aqua `opencode` (v1)
- Install mise tools from wrappers only when missing, so explicit pins like `@next` survive normal use
- Skip default-agent reactivation for OpenCode when it is already installed (other agents unchanged)
- Migrate existing installs: regenerate the npm wrapper and drop a stale aqua `opencode` pin when npm is already configured

## Problem

Users on OpenCode v2 (`npm:@opencode-ai/cli@next`) kept getting pushed toward v1 stable because Omarchy installed aqua `opencode`, re-ran `mise use -g opencode` from the wrapper, and `omarchy update` upgraded the conflicting aqua entry.

## Test plan

- [x] `./test/all`
- [x] `bash test/shell.d/default-agent-test.sh`
- [ ] After merge: on a machine pinned to `npm:@opencode-ai/cli@next`, confirm `mise outdated` no longer reports aqua `opencode` and `c`/`opencode` stay on `next` across `omarchy update`

### DIFF
--- a/bin/omarchy-default-agent
+++ b/bin/omarchy-default-agent
@@ -45,15 +45,15 @@ if [[ $installing == "false" ]] && ! mise where "$agent_package" &>/dev/null; th
   exec omarchy-launch-floating-terminal-with-presentation omarchy-default-agent --install "$agent"
 fi
 
-if ! mise where "$agent_package" &>/dev/null; then
-  if ! mise use -g "$agent_package"; then
-    if [[ $installing == "true" ]]; then
-      echo "Could not install $name with mise" >&2
-    else
-      echo "Could not set $name as the default coding agent" >&2
-    fi
-    exit 1
+if [[ $agent == "opencode" ]] && mise where "$agent_package" &>/dev/null; then
+  :
+elif ! mise use -g "$agent_package"; then
+  if [[ $installing == "true" ]]; then
+    echo "Could not install $name with mise" >&2
+  else
+    echo "Could not set $name as the default coding agent" >&2
   fi
+  exit 1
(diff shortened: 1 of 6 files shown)


AndreasHald:
I ran into the same installation-source problem and opened #11423 for the now-stable V2 distribution, referencing this PR as related work.

OpenCode now documents `@opencode/cli` as its official V2 package (`2.0.1` is npm's current stable release); the GitHub/Aqua route still resolves to V1.18.30. The older `@opencode-ai/cli@next` target in this proposal would need updating.

The new PR uses mise's existing wrapper behavior and focuses on the stable package, default-agent selection, and migrating stock `opencode = "latest"` installations. It installs V2 before removing the old request with `--no-prune`, keeps unused stubs lazy, and leaves custom/pinned/beta setups alone during migration.

One concrete finding from a real isolated install: mise needs `allow_builds=@opencode/cli` so the package's postinstall script can select the native binary. With that setting, I verified V1.18.30 → V2.0.1 migration and subsequent `mise up`; the PR includes focused regression tests and the full-suite results.

@mattstyles333, happy to coordinate or consolidate the overlapping work if you would prefer to update this PR. Thanks for identifying the backend/pin conflict earlier.

bjarneo:
Automated duplication check: this pull request looks similar to #11423, which covers the same OpenCode V2 mise install. I keep that one open and close this one to consolidate review.

Issue #7084, opened 2026-08-16
Clamshell watcher ignores desc: rules for the internal panel and overwrites its scale every 2s


mkelk:
### What's wrong?

`omarchy-hyprland-monitor-clamshell` re-applies the internal panel every 2 seconds while docked, but it can only read a `monitors.lua` rule that is keyed by **connector name**. A rule keyed by `desc:` is invisible to it, so it silently falls back to a different scale and then overwrites the panel about one second after anything — the config, the Display panel, `omarchy hyprland monitor scaling` — sets the correct one.

`desc:` is standard Hyprland monitor syntax (documented on the wiki the shipped `monitors.lua` links to) and is the usual way to survive DP connectors renumbering across hotplug. Hyprland honours such a rule; the watcher does not, and then overrides Hyprland with its own guess.

**Symptom:** the built-in display sits at the wrong scale and will not stay changed. `hyprctl reload` appears to do nothing. Setting the scale by hand holds for one sample and then snaps back.

### Steps to reproduce

On a laptop with an external monitor connected (the poll only runs while docked):

1. In `~/.config/hypr/monitors.lua`, key the internal panel by description instead of connector name, with a numeric catch-all below it:

   ```lua
   hl.monitor({ output = "desc:Samsung Display Corp. ATNA60HR07-0", mode = "2880x1800@120", position = "auto", scale = 2 })
   hl.monitor({ output = "", mode = "preferred", position = "auto", scale = 1 })  -- catch-all
   ```

2. `hyprctl reload`
3. `watch -n1 'hyprctl monitors -j | jq -r ".[]|select(.name==\"eDP-1\").scale"'`

**Expected:** `2` — the rule matches the panel, and `hyprctl monitors` confirms the description matches exactly.

**Actual:** `2` for about one second, then `1` forever. The source of the change makes no difference — the poll re-asserts its own value regardless of whether the scale came from the config, `hyprctl eval`, or the Display panel.

### Root cause

In `bin/omarchy-hyprland-monitor-clamshell` (line numbers from `quattro` HEAD, `7be59e1`):

`monitor_rule_regex()` (L69) builds its matcher from `INTERNAL`, which is the connector name from `omarchy-hyprland-monitor-laptop` (L12) — `eDP-1`:

```bash
printf '^[[:space:]]*hl\\.monitor\\(\\{.*output[[:space:]]*=[[:space:]]*"%s"' "$1"
```

[...]

Issue #6909, opened 2026-08-14
[Quattro] eDP-1 monitor scaling forces 2x reset on mixed-monitor array with Quickshell navbar


PedroMiguelInacio:
### System details

11th Gen Intel(R) Core(TM) i5-11400H (12) @ 4.50 GHz  GeForce RTX 3050 Mobile 1920x1080 in 27", 60 Hz [External]  1920x1080 @ 2x in 15", 144 Hz [Built-in]

### What's wrong?

After upgrading to Omarchy Quattro, the internal laptop monitor (eDP-1) is being aggressively forced to a high-DPI scaling factor of 2 (Retina fallback). The external display (HDMI-A-1) scales completely normally at any number.

Any manual attempts to change the internal laptop display density down to a native 1 scaling factor are completely overridden by a background layout watcher or configuration event hook precisely one second after applying the changes.

Additionally, opening the newly introduced Quickshell top navbar Display adjustment tool to manually correct the internal screen's scale layout parameters results in an instant crash of the panel utility tool, immediately defaulting the viewport state back to an oversized scale: 2 state.

DavidSurkov:
Just write monitor config inline
`hl.monitor({ output = "eDP-1", mode = "1920x1080@144", position = "0x0", scale = 2 })
`
It breaks only if the config if spreaded into multiple lines.
`
-- hl.monitor({ /n
-- 	output = "eDP-1",/n
-- 	mode = "1920x1080@144",/n
-- 	position = "0x0",/n
-- 	scale = 1,/n
-- })/n
`

whughesiii2187:
> Just write monitor config inline `hl.monitor({ output = "eDP-1", mode = "1920x1080@144", position = "0x0", scale = 2 }) ` It breaks only if the config if spreaded into multiple lines. `-- hl.monitor({ /n -- output = "eDP-1",/n -- mode = "1920x1080@144",/n -- position = "0x0",/n -- scale = 1,/n -- })/n`

Tried this and my laptop display still is being forced to scale 2 even if set to scale 1

jesseburlamaque:
I'm hitting the same issue on Omarchy 4.0.0-1 + Hyprland 0.56.2 with two monitors (eDP-1 + HDMI-A-1).

[...]

PR #7146, opened 2026-08-16
Disable the panel by overlay alone in clamshell recovery

### LINKED ISSUE #7498: Internal display flashes to 2x scale on lock-screen mouse move / resume (desc: output rules unmatched)
## What happened

On resume from suspend, and continuously while moving the mouse on the lock screen, the internal display's scale flickers between the correct value (1) and 2 — the whole shell (bar, text, lock screen UI) briefly jumps to double size and then snaps back. After unlocking it happens once more and then stays stable.

## Expected

The internal display should stay at its configured scale; no scale should ever be force-applied unless it actually needs correcting.

## Root cause

`omarchy-hyprland-monitor-clamshell` is invoked on every "wake" (including from `LockView.qml`'s `MouseArea.onPositionChanged` on the lock screen, via `omarchy-system-wake`). Its `enable_internal` → `sync_internal_scale` path decides whether to re-apply a scale by comparing the *active* scale against a *desired* scale read from `~/.config/hypr/monitors.lua`:

```sh
monitor_rule_regex() {
  printf '^[[:space:]]*hl\\.monitor\\(\\{.*output[[:space:]]*=[[:space:]]*"%s"' "$1"
}
```

`configured_internal_monitor_value` calls this with `$INTERNAL` = the literal port name from `omarchy-hyprland-monitor-laptop` (e.g. `eDP-1`). This only matches a rule written as `output = "eDP-1"`. It does **not** match a rule that identifies the same output by description, e.g.:

```lua
hl.monitor({
  output = "desc:LG Display 0x0742",
  mode = "1920x1080@144.00",
  scale = 1,
  ...
})
```

`desc:`-based output selectors are a normal, documented way to address a monitor in Hyprland (useful because port names can change across kernels/docks), and are what third-party monitor-profile tools (e.g. `hyprmoncfg`) generate by default.

When the regex fails to match, `configured_monitor_scale` returns empty. `read_monitor_scale` then falls back to a saved state file (`~/.local/state/omarchy/toggles/hypr/internal-monitor-scale`), and if that's also absent, hardcodes a default of `2`:

```sh
read_monitor_scale() {
  local scale
  scale=$(configured_monitor_scale)
  if valid_scale "$scale"; then echo "$scale"; return; fi
  if [[ -f $SCALE_STATE ]]; then
    scale=$(<"$SCALE_STATE")
    if valid_scale "$scale"; then echo "$scale"; return; fi
  fi
  echo 2
}
```

On a non-HiDPI internal panel configured at `scale = 1` via a `desc:` rule, this fallback is wrong. `sync_internal_scale` then sees a mismatch against the real active scale (1) and force-applies `scale = 2` via `hyprctl eval`, causing the visible flash. A separate monitor-profile daemon (in my case `hyprmoncfgd`) then corrects it back on its own poll cycle, producing the flicker. Because the lock screen fires `wakeRequested()` (and thus this whole chain) on *every* mouse-move event, the flicker repeats continuously while the mouse moves on the lock screen, and fires once more on unlock.

## Steps to reproduce

1. Configure the internal monitor in `~/.config/hypr/monitors.lua` using a `desc:`-based `output` selector at `scale = 1` (e.g. via `hyprmoncfg`, or by hand).
2. Make sure `~/.local/state/omarchy/toggles/hypr/internal-monitor-scale` does not exist / does not hold the correct scale.
3. Lock the screen and move the mouse continuously, or resume from suspend.
4. Observe the shell repeatedly flash to double size and snap back.

## Suggested fix

`configured_internal_monitor_value` / `monitor_rule_regex` should also match the internal monitor when it's identified by `desc:<description>` (available via `hyprctl monitors -j` as `.description`), not only by literal port name. At minimum, `read_monitor_scale`'s hardcoded final fallback of `2` should not be force-applied without first confirming a mismatch actually needs correcting — right now a parse failure is indistinguishable from "the user wants 2x", and the function actively pushes that guess to the compositor instead of leaving the monitor alone when it can't determine the configured scale.

## Workaround

Manually seeding `~/.local/state/omarchy/toggles/hypr/internal-monitor-scale` with the correct scale (`1` in my case) makes `read_monitor_scale`'s second fallback succeed, so `sync_internal_scale` sees a match and never force-applies anything.

## System details

- Omarchy: 4.0.0-1
- Machine: Acer Nitro ANV15-51 (hybrid graphics)
- CPU: 13th Gen Intel Core i5-13420H
- Graphics: Intel Raptor Lake-P UHD (i915, active on eDP-1) + NVIDIA GeForce RTX 3050 6GB Laptop GPU (nvidia 610.57.04)
- Kernel: 7.1.8-arch1-3
- Hyprland: 0.56.2
- Monitor profile tool in use: `hyprmoncfg` 1.13.0 (AUR), which generates `monitors.lua` with `desc:`-based output selectors

## Unrelated bug noticed while gathering diagnostics

`omarchy debug --no-sudo --print` (and `omarchy debug` in general) currently fails with "Unknown Omarchy command: omarchy debug" even though it's listed as a common command in `omarchy --help`. The underlying binary `/usr/bin/omarchy-debug` works fine when invoked directly, so this looks like a dispatcher/registration regression. Happy to file this separately if preferred.

### LINKED ISSUE #8123: nwg-displays' multi-line hl.monitor() blocks break internal-panel scale sync (resets to 2)
### System details

Omarchy version: 4.0.0-1, Hardware: Lenovo Slim Pro 7 14APH8 (laptop), Kernel: Linux 7.1.8-arch1-3, GPU: NVIDIA GeForce RTX 3050 6GB Laptop GPU (GA107BM) + AMD Phoenix1 (integrated), nwg-displays version: 0.4.3-1Monitors: internal `eDP-1` (2560x1600@90Hz), external `HDMI-A-1` (3840x2160@60Hz, Sceptre Z27)

## What happened

On a laptop with an external monitor connected and active, setting the internal
panel's scale to a fractional value (e.g. 1.25) via `nwg-displays` gets silently
reverted to 2 within a couple of seconds — every time, no matter what scale is
chosen. The external monitor's scale applies and holds fine; only the internal
panel is affected.

## Expected behavior

The internal panel should keep whatever scale is set in `~/.config/hypr/monitors.lua`,
same as the external monitor does.

## Steps to reproduce

1. On a laptop with an external monitor attached and active, open `nwg-displays`.
2. Set the internal panel (e.g. `eDP-1`) scale to `1.25` (or any non-2 value) and save.
3. `nwg-displays` writes `~/.config/hypr/monitors.lua` using a multi-line
   `hl.monitor({...})` block, e.g.:
   ```lua
   hl.monitor({
       output = "eDP-1",
       mode = "2560x1600@90.0",
       position = "0x0",
       scale = 1.25
   })
   ```
4. Within ~2 seconds, the internal panel's scale reverts to 2.

## Root cause

`/usr/share/omarchy/bin/omarchy-hyprland-monitor-clamshell` is responsible for
re-applying the internal panel's configured scale (used to restore it correctly
after suspend/resume, lid open/close, and dock/undock). Its `configured_monitor_value()`
function locates the relevant `hl.monitor({...})` block with a `sed` regex that
requires `output = "..."` to appear on the **same source line** as `hl.monitor({`:

```
monitor_rule_regex() {
  printf '^[[:space:]]*hl\\.monitor\\(\\{.*output[[:space:]]*=[[:space:]]*"%s"' "$1"
}
```

`nwg-displays` (version 0.4.3, a commonly recommended display-configuration GUI
for Hyprland/Omarchy) always writes `hl.monitor({...})` blocks across multiple
lines, so this regex never matches for the internal panel. `configured_monitor_scale()`
then returns empty, `read_monitor_scale()` falls through past the (also empty)
state-file fallback, and returns the hardcoded default of `2`.

This function is called from `sync_internal_scale()` → `enable_internal()`,
which runs:
- on every `monitoradded`/`monitorremoved` Hyprland event, and
- every 2 seconds via `poll_clamshell_state` in `omarchy-hyprland-monitor-watch`,
  which runs continuously whenever `omarchy-hw-laptop && omarchy-hyprland-monitor-external-active`
  (i.e. exactly the "laptop + external monitor" setup where this bug shows up).

So any scale set on the internal panel gets reapplied and stomped back to 2
within ~2 seconds. External monitors are untouched because this script only
manages the internal/laptop output.

## Confirmed manually

- `hyprctl eval 'hl.monitor({ output = "eDP-1", mode = "2560x1600@90.0", position = "0x0", scale = 1.25 })'`
  applies immediately, then reverts to `scale: 2` within ~2-4 seconds (caught via
  `hyprctl monitors` polling).
- Reformatting the same block onto a single line in `monitors.lua` lets
  `configured_monitor_value()` parse it correctly and the scale then holds.
- Seeding `~/.local/state/omarchy/toggles/hypr/internal-monitor-scale` with `1.25`
  (the parser's secondary fallback, checked before the hardcoded default) also
  works around it without touching `monitors.lua` formatting — this is what I'm
  using as a temporary fix, but it silently goes stale the next time the scale
  is actually changed via nwg-displays.
- Also noticed (separately, likely unrelated): two copies of
  `omarchy-hyprland-monitor-watch` were running simultaneously in my session,
  seemingly a leftover from a prior Hyprland reload — doubles the polling but
  isn't the cause of the scale-reset bug itself.

## Suggested fix

Make `configured_monitor_value()`'s parsing multi-line aware — e.g. flatten each
`hl.monitor({...})` block onto one logical line before matching (strip newlines
between the balanced `{` and `}`), or replace the regex-based scrape with a
proper Lua parse — since `nwg-displays`, a standard/recommended tool for this
workflow, never emits single-line calls.

mkelk:
The parser this PR originally added is withdrawn; `docs/clamshell.md` in this PR carries the invariant the command is now written to, and everything below was verified against a real Hyprland in the ISO acceptance VM, not against stubs alone.

Closes #8123. Closes #7498. References #7084 (its `require`d-file half is the maintainer's to place). #7326 and #8103 live in `omarchy-hyprland-monitor-scaling` and stay open.

**The bug class.** `omarchy-hyprland-monitor-clamshell` disabled the panel with a toggle overlay and a reload — and then also evaluated an explicit `hl.monitor({ … scale = … })` on top, with values read out of `monitors.lua`. That explicit write is the phantom-scaling family: to write the right value the script had to read the config exactly the way Hyprland reads it, and no reader closes that obligation — rules also come from `hyprland.lua`, required files, and the toggles. Three readers were built and reviewed on this PR; every review round found the same defect in a new spelling.

**The change, in four parts:**

1. **Overlay only.** The command authors exactly one thing: `hl.monitor({ output = "<internal>", disabled = true })` in the toggles directory. Enabling is removing it and reloading — Hyprland re-applies its own config, panel included, at whatever scale, position and mode the user configured, wherever they configured it. Nothing reads `monitors.lua`, nothing writes a monitor value. Transitions are atomic (mktemp + rename), one bounded reload each, with byte-exact file-backed rollback and a retry on the next run; an idempotent run creates nothing, not even transiently. The predicates are three-valued — true, false, or unknown — and an unknown is never taken for false, so a hung `hyprctl` cannot re-enable the panel mid-clamshell. The unsafe-name refusal (#8129) is kept, before any effect, and the hosted recoveries run only when the toggle each exists to recover is present.

2. **The lock, made real.** The command takes its own blocking lock, so a lid event queues instead of being dropped (the watcher's `flock -n` skip wrapper goes). Every call that can reach Hyprland is bounded by a timeout that escalates TERM→KILL — a callee that ignores TERM cannot be waited on forever — and the lock descriptor is not lent to those callees, so a surviving descendant cannot keep the lock after the command exits (both failure modes reproduced first). A lock that cannot be taken exits before any effect.

3. **The external predicate, made honest.** `omarchy-hyprland-monitor-external-active` broke its own contract twice. It ended in `jq -e`, whose exit 4 for "no match" made every undocked laptop's "no active external" read as *unknown* — so the manual-toggle wake never fired in production. And it counted `FALLBACK`, the enabled placeholder output Hyprland uses under a reserved name when no real output remains — so a machine whose only display had gone down read as docked, deadening the recovery helpers' own "no external active" guards in exactly the states they exist for. It now answers 0 for an enabled real external, the contracted 1 when a successful query finds none (`FALLBACK` excluded by its exact name), and 2 when the query itself failed. Truthiness callers keep their existing behavior — which still treats unknown like false: a transient query failure while an external is active can still drop a user toggle, a pre-existing residual now stated in `docs/clamshell.md` — while the helpers seeing `FALLBACK`'s truth in their recovery states is a deliberate behavior change.

4. **The stranded panel, repaired deliberately.** A helper killed between removing its toggle flag and its reload — reachable, since `omarchy-system-sleep-lock` bounds this command at 0.4 s — leaves Hyprland holding `disabled = true` with nothing on disk wanting it: with no external monitor, a machine with no usable display. The old command papered over this by accident with its explicit eval; this PR repairs it on purpose. When the overlay is absent, the enable path asks the *compositor* — never the config: a panel reported disabled, with no manual toggle and no active external, gets one reload so the config answers again; a panel that comes back is woken, one the config still reports disabled is not. The answer is not remembered, so while that state persists each reconciliation invocation spends one idempotent reload — a lid event is one invocation, a monitor-change retry chain up to four — and a reload that brings another output up closes the gate instead. Every unknown repairs and wakes nothing, including a query that prints a valid answer and then hangs or fails: the query's status is checked before its output is parsed.

**Measured, not assumed.** `test/acceptance.d/clamshell-test.sh` runs the real command against a real Hyprland (headless outputs standing in for the panel and the external; only the two lid/DRM hardware predicates stubbed). Nine `monitors.lua` shapes round-trip with the panel coming back as it was; a user's own disable rule governs through and out of clamshell; a config Lua rejects reloads "ok" with the error in `configerrors` while the overlay still applies beside it (measured — and asserted); an `auto` position is laid out afresh on each reload rather than restored as a slot (identical under the previous command, so no behavior change); and the repair path is exercised for real — the stranded state staged against Hyprland's actual `FALLBACK` placeholder, with the repair's effects *observed* through a forwarding `hyprctl` wrapper: exactly one real reload per gated run, a wake only for a panel that actually came back.

**Tests.** `test/shell.d/monitor-clamshell-test.sh` is derived from the invariant, not the implementation: T1 asserts the command's text contains no reader and no monitor-value write; P1 runs the real external predicate through its whole contract against a controlled `hyprctl`; T2 makes both transitions under six configs any reader would answer differently and requires identical calls; T3 walks the full transition table with every collaborator stubbed — the 3×3 predicate matrix, byte-exact rollback, lock queueing/refusal/surviving-descendant, every call class hung (TERM-ignoring included), the repair row with each of its gates and the masked-query cases at both steps. Each added case was demonstrated to fail against the code without its fix, and key fixes are mutation-checked. The reader's 413-line scale test is deleted with the reader.

**Stated residuals** (`docs/clamshell.md`): the DPMS wake is best-effort and not retried; a stale disable rule beside a still-active external stands until the next transition or an unrelated reload (the user has a display); a migration removes the now-unwritten `internal-monitor-scale` state file.


mkelk:
Force-pushed. The PR now also reads a rule whose output names a local (reported by @jelenv on #6909) — the same reader missing a rule for the same reason as the multiline case, so it belongs here rather than in a fourth PR. Rebased onto current `quattro`.

Body updated with the new case, the test list (35 assertions, was 30), and how this sits against the other open monitor work: #7581, #7000 and #7505 all merge clean with this in either order, and the suite is green on the merged trees — verified, not assumed.

jelenv:
Thank you @mkelk

mkelk:
@omarchybot All three confirmed and fixed, along with rather more that pulling on them turned up. The branch is rebased onto `quattro` at 2c247e39 and force-pushed.

## Your three

Reproduced exactly as you described, so I started where you pointed: making the fold pass string-aware. One character-scanning walk tracking quoted strings with their escapes, and long brackets at any level, replaced `strip_comments` and `depth_of` — they need identical string awareness, so one scanner emitting stripped text and a depth delta is smaller than the two of them, not larger. `--[==[` falls out for free once long brackets are understood rather than pattern-matched, and the same walk covers long *strings* used as values.

I also took your smaller item — `internal_monitor_output` trimmed the selector with `read` and then looked the rule up by the trimmed spelling, so `output = "desc:Samsung Display Corp. "` was identified and then not findable. `IFS= read` keeps the spelling; only the description a `desc:` name carries is trimmed, which is what Hyprland trims.

And you were right about `write_block_commented_rule_config`. It used a *multiline* rule inside the comment, which `quattro` cannot fold and therefore also reads correctly — the fixture discriminated nothing and the body claimed otherwise. A single-line rule inside `--[[ ]]` does discriminate, and that is what it uses now.

## What that turned up

Fixing the fold made it obvious the same blindness lived in the other readers, and further review found it did:

- **A name the config rebinds between rules.** `lua_local_value` took the first `local` in the file, so `local mon = "eDP-1"` … `mon = "DP-2"` pointed *both* rules at the panel and the panel got DP-2's scale and position, reapplied every 2 seconds. Declaring one variable and reusing it per monitor block is a normal way to write that config, which made this the worst of the lot.
- **A key merely ending in `output`.** `monitor_rule_outputs` had no key-boundary guard and matched greedily, so `mirror_output = "DP-2"` hid the rule's real output and the panel fell to the catch-all. Same mechanism for rule-shaped text inside a quoted value.
- **A backslash-continued string.** Lua carries a short string past the end of its line; the scanner did not, so a rule written inside such a string was read as a live rule. `lua5.1` with a stubbed `hl` confirms the config means one thing and the script read another.
- **A multiline scale expression.** After folding, `scale = scale_for(\n 2880\n)` extracted as *empty*, which read as a rule naming no scale, so the catch-all overruled a panel Hyprland was scaling correctly — the exact flap the delegate guard from #7581 exists to prevent.

Every one of those is a regex that cannot see where a string begins, where a key ends, or what order the file is in. Patching them individually would have left the next shape unreadable, so the reader is now one pass: `scan_monitor_config` walks the config once and emits only what the rest needs — which output each rule addresses, and what it says about scale and position. Rules are matched by comparing that output as text, so nothing is escaped into a regex any more. `resolve_rule_outputs`, `monitor_rule_regex`, `monitor_rule_outputs`, `monitor_rules`, `join_monitor_rules` and `lua_scalar` are gone.

That also closed shapes neither of us had flagged as broken: `{output = name}` with no space after the brace (the local-keyed feature silently no-opped there), the table brace on the next line, `hl.monitor { }` without call parens, single-quoted strings in a rule rather than only in a binding, and a name bound twice now taking the later binding as Lua does. And an empty `desc:` selector: `starts_with("")` is true, so Hyprland applies such a rule while the script skipped it — the script and the compositor disagreeing about which rule governs the panel, which is the one thing this reader must not do.

In fairness the rewrite introduced one of its own on the way, caught before it settled: the gate deciding whether a line held a monitor call was still a plain regex, so an unrelated table whose string happened to mention `hl.monitor(` and which carried rule-shaped keys was read as a rule. It is a string-aware walk now, with a fixture.

## On the tests

Your point about non-discriminating fixtures generalised. A mutation pass found two more guards covering nothing: deleting the regex escaping from `monitor_rule_regex` entirely passed all 49 assertions, and so did removing the empty-`desc:` guard. The escaping is moot now — text comparison has nothing to escape — but the replacement is mutation-tested: swapping it for a glob comparison kills a fixture carrying `?`, `*`, `[` and `.`.

Your "last rule matching the panel wins" fixture had the same shape of problem I had missed: it put the `desc:` rule *first*, an order `quattro` also passes because it ignores `desc:` rules and lands on the connector rule anyway. Connector-first is the order that distinguishes last-match-wins from connector-always-wins, and there is now a fixture for it.

The suite is 63 ok, up from 25. Every new assertion was checked to fail against the version before the fix it guards, by running the suite with a non-fatal `fail` against each ancestor commit rather than by assuming.

## On your last two notes

The ~4 ms you measured was on configs without local-keyed outputs. With them the cost multiplied — the config was scanned once per lookup and again per name, 42 times per invocation on a twelve-monitor config, about 194 ms every 2 seconds. Reading it once makes that flat: 21 ms there against 19 ms for the shipped config, both at or below `quattro`.

And `INTERNAL_OUTPUT` is computed on the branch that reads it now. You were right that the clamshell-disable branch never touches it, and that branch is the one the docked poll takes.

## Verification

`./test/shell` is 182 of 188 files green, with the same six failing identically on an unmodified checkout for local system-state reasons. `./test/cli` green, and `monitor-recovery`, `monitor-state`, `monitor-scaling`, `monitor-modeless` and `lid-close` green. Differentially against `quattro` over the fixture battery: every already-working shape byte-identical, and every difference a case this PR exists to fix, checked against a `lua5.1` oracle where the answer was not obvious.

Worth another look — the fix is a good deal larger than what you reviewed.

mkelk:
@omarchybot All four residuals are fixed, and your merge commit is kept as pushed. Rebased onto `quattro` at b71c60fe and force-pushed; `0c336b26` is the head.

## Your merge commit

Verified before building on it. The mechanism holds, and the reader was already shaped to take it: `configured_monitor_value` filters on the exact output spelling, so the merge is keyed the way Hyprland keys it. Worth adding, because it is what makes the two halves consistent — since `add` re-appends the merged rule, a rule's place in the reverse walk is its *last* mention in the file, which is exactly what `internal_monitor_output` returns. "Last name that matches, then every rule carrying it" is the same answer for that reason, not by luck. The body's paragraph on `add` is corrected, with that consequence spelled out.

## The four, and two corrections to your report

I ran your examples verbatim before reconstructing anything, and two of them do not reproduce as written:

- **#1 reproduces exactly.** `hl.monitor({ output = [=[DP-1]=], scale = 3 })` scanned to `R␟␟1␟3␟…` — an empty output, the catch-all's own name. It now scans to `R␟DP-1␟1␟3␟…`.
- **#2 does not reproduce in either form you gave.** Both `local defaults = { omarchy_monitor_scale = 3 }` and `do local omarchy_monitor_scale = 3 end` are single statements on one line, and `record_binding` already rejected both — `cffe8aea` reads 1.25 for each, same as `quattro`. The bug is real, but only across lines: the table field needs its own line inside the constructor, and the `do` block needs the `local` on a line of its own. My fixtures use those forms.
- **#3 and #4 reproduce as written.**

## Three causes, not four

- **`42f86d80` — a long string is a value.** A long bracket was erased whether it opened a comment or a string. Erasing a comment is right; erasing a string leaves the rule naming nothing, and nothing is the name the catch-all answers to, which is what manufactures the catch-all you described. The text now travels under a token, which also lets a name bound to one carry what it holds, and holds the statement open across the lines a long string spans the way a backslash-continued string already did.
- **`9f2f5e90` — a rule is a call, not a line.** Your #3 and #4 are one bug: the record boundary was the line or the fold rather than the call's own argument. Walking to the bracket that closes the call fixes both at once, and two calls on one line become two rules.
- **`0c336b26` — bindings are scoped.** Bracket depth stops a table field binding; a keyword-level block count stops a `local` inside a `do` block or a helper function escaping. A plain assignment inside a block still binds, because in Lua it reaches the outer name — so the keyword decides this rather than the depth alone. Miscounting can only leave a name unbound, never bind it wrongly, so a shape it cannot follow falls back to the catch-all.

Your unterminated-fold note I agree with and left alone.

## On the tests

The suite is 76 ok, up from 63. Every new assertion checked to fail against `cffe8aea`, except three that guard the new scope tracking rather than an old bug — those are mutation-tested instead, and die to `function` opening no block, to `elseif` opening one, and to keywords matched inside names like `dock` or `render`.

Writing them found two defects in my own fixes, both before the branch settled. The `elseif` check compared the word *preceding* `then`, which is the condition and not the keyword, so every `if`/`elseif` chain over-counted; and the top-level test conflated bracket depth with block depth, rejecting an assignment inside a multiline block as scoped when it is not. Neither was found by the fixture I set out to write — the first by the `elseif` fixture, the second by the assignment fixture I added only to keep the new tracker honest.

Your dead-guard standard also applied to this round's code. Two guards survived every mutation because nothing reachable exercised them — a leading-newline strip that the scan can never feed, and the two depth clamps, which valid Lua cannot drive negative. Both deleted rather than kept and claimed.

## Verification

`monitor-clamshell-scale-test.sh` 76 ok. `./test/cli` green, and `monitor-recovery`, `monitor-state`, `monitor-scaling`, `monitor-modeless`, `lid-close` and `sleep-lock` green. `./test/shell` is 2268 assertions green with 14 failing across 14 files; each of those was run against a clean worktree of the same `quattro` commit and fails identically there, so none is this branch's. No monitor, clamshell or lid test is among them.

Differentially against `quattro` over 126 fixtures: only the seven intended cases changed, and each was checked against a `lua5.1` oracle with a stubbed `hl` rather than reasoned about. One of the seven was a fixture neither of us filed — two rules separated by a space rather than a semicolon, where `quattro` takes the *other* monitor's scale for the panel.

Timing on the twelve-monitor local-keyed config is 20.5 ms against 21.7 ms on `quattro`; the shipped config is 17.8 ms against 21.5 ms.

The fix is again larger than what you reviewed, and the scope tracking is the part I would look at hardest — it is the only piece here that reasons about Lua by keyword rather than by character.

mkelk:
@omarchybot Your two open items are settled, and the round that settled them ran twice more before landing — codex at xhigh reviewed the branch twice, each pass finding something real, and each finding was verified against the `lua5.1` oracle before being kept. Rebased onto `quattro` at 5d3299fb and force-pushed; `a8f1bf7f` is the head.

## The block-assignment trade

You framed it as binary: take the assignment, or delete the binding. We took a third path — not every block is a condition, so the tracker keeps block *kinds* rather than a count. An assignment inside a `do` or `repeat` body has run and binds; inside a function body it has not run where it is written, and does not; inside an `if` or a loop body it is unknowable, and the name is left unreadable so the rule using it falls back to the compositor. That keeps the reads the flat delete would have lost, and it caught a shape neither of us had filed: a helper function assigning to a name the rules below use no longer rebinds it.

Codex then completed it in the direction your comment pointed — "the same mechanism reaches the output name." A name marked unknowable now makes a rule whose *output* rests on it emit as *unresolved* (`U`), and any unresolved rule makes the panel's rule ambiguous, because it could address the panel and could be the last match. Recovery then acts on nothing: the docked poll delegates to the compositor, whose answer is definitionally right, and re-enable uses the scale the panel last ran at. The same restraint now covers a whole `hl.monitor` call inside a branch or a function — a rule that may not exist and a rule that may be anything ask the same thing of the reader. `quattro` reads a rule out of a never-called function and applies it two writes a second; that one was ours to find, not yours, and it is fixed at `db1149f4` with both gate directions mutation-killed.

One honest cost: any output the scanner cannot resolve — `"DP-" .. n`, say — now delegates the whole config where `quattro` applied the catch-all. Safe direction, but a behavior change on a plausible shape, and yours to veto.

## Your "smaller one, not worth a commit"

It was worth two, because the obvious fix was wrong. Joining a record's newline into a space manufactures a desc match Hyprland does not make — trimming touches a description's ends, never its middle — and that draft lived on this branch long enough for codex to catch it. What landed instead escapes the record and field separators and the escape itself, decoded on read, so a multiline `desc:` long string neither splits its record nor changes under transport. Your framing half-holds: it ends at the same fallback *when the truncated half matches nothing* — the fixture that killed the escape mutant pairs the broken name with a connector rule, where every wrong reading shows.

## The semicolon, three layers down

Your `ff4292f7` fixed the leading semicolon. The segment walk codex built to make one-line blocks read in order then met the third layer: `;` separates table fields as readily as statements, so `local defaults = { dummy = 1; omarchy_monitor_scale = 3 }` bound the knob to a field — your original table-field finding, reopened in its one-line spelling. Segments now know the bracket depth they begin at, and that depth carries across a call written inside a statement, so a rule written as a table field still reads while the field after it stays a field.

## Verification

The suite is 93 ok, up from the 76 you reviewed. Every new guard is mutation-tested; the depth threading and the record escaping each survived their first battery and are pinned by fixtures now. A handful of codec and bracket branches survive every realistic mutation and are kept anyway, stated rather than hidden: they keep the encoding injective and the bracket walk symmetric, and partial versions of both are exactly what two of this round's defects were.

Five defects in this branch's own fixes were found and fixed before settling — the `elseif` word comparison, bracket depth conflated with block depth, the newline-to-space join, the one-line block read out of order, and the table-field semicolon. Three were mine, two were codex's own machinery on its second pass. The finder each time was a mutation pass or a probe, not the fixture anyone set out to write.

`./test/shell` is 2479 assertions green with the same 14 environmental files failing identically on a clean worktree; `./test/cli` and all six monitor suites green. Differentially: your head to this one changes 13 fixtures out of 157, every one intended and oracle-checked. Timing: shipped config 18.4 ms against 21.5 on `quattro`; the twelve-monitor local-keyed config 23 ms against 22 — the decode loop costs it about a millisecond.

Still open from your last review, unchanged: the cross-padding desc merge (`desc:X` and `desc:X ` now merge keys as one rule; whether Hyprland's seed lookup trims before comparing is yours to check against the source), and the `U` blast radius above. Waiting on you for those, then on the maintainer.

mkelk:
@omarchybot All four are fixed, rebased onto `quattro` at 23dab9ec — which brings #8129's name guard in above the parser untouched — and force-pushed; `6e89ef30` is the head. `Fixes #7084` is downgraded as you recommended: the body now closes #8123 and #7498, references #7084, and says plainly that @siebertm's `require`d-file half stays live and is the maintainer's to place, with #7000 named as the other approach.

## The regression

`1ec5c2cf`. You were right about the shape of the fix and about the cheap version being worse — I checked both before building the machinery. A shadowing `local` now binds, with the outer state saved once per name and block and restored when the block pops; your example reads exactly as `lua5.1` registers it, the rule inside the `do` block addressing DP-1 and the panel keeping the catch-all's 1.25/`auto`. Nested shadows, an assignment to the inner name, and a shadow inside a branch all match the oracle, and the restore keeps the whole state — a name that only ever existed inside a block is nothing below it, not the empty string, which is the catch-all's own name. The mutant that restores it as empty hands the panel a rule Hyprland gave to nobody; there is a fixture for that, because the mutation battery found the gap before the fixture existed.

## The trade, taken from the binding side too

`a6693c1b`. Your framing — "a rule that may not have run and a name that may not have been reassigned are the same question, answered two different ways in the same file" — is the whole argument, so they now get the same answer. A name a function body assigns to is unknowable whether anything calls the helper or not, and what rests on it defers to the compositor. Your `use_hidpi()` example no longer writes 1.25 over Hyprland's 2.5; it writes nothing on the docked poll. The cost is stated in the commit: the helper nothing calls, whose outer value was read correctly, now defers too. Deferring is never wrong; the confident write was. The PR body's function-body sentence is corrected to match.

## The two smaller ones

`df758773` — the function-expression hole codex found. The header an `if` is still reading is saved when a function body opens and restored when it ends, so the body's own `if`/`then` work unbothered and the `then` outside still closes the `if` it belongs to. Your example defers instead of confidently reading the dead branch.

`41025d27` — the alias. Lua copies the value at the binding line, so the binding does too, and it copies unknowability with it: an alias of a name a branch may have moved defers rather than letting the catch-all overrule a rule Hyprland may be applying. Your `"eDP-\049"` observation stands — short-string escapes are still carried through undecoded, losing a read rather than inventing one, and stays a known residual.

## Verification

Suite 99 ok, up from your 93, each of the four commits standalone-green at its own count. Ten mutants against the new machinery, all killed — shadow-never-saved, refused-instead-of-bound, restore-drops-everything, restore-as-empty-string, header-not-saved, header-cleared-like-any-block, refusal-covering-only-`f`, only-`c`, alias-not-followed, alias-certainty-invented. The restore-as-empty and alias-certainty mutants survived their first battery; both kill-shapes were verified to discriminate before being pinned, the first needing the phantom rule *after* the catch-all so the key merge could not hide it.

Differentially against `a8f1bf7f` over 167 fixtures: 7 changes, each intended — the regression now byte-identical to `quattro`, two reads gained (the alias, the restored shadow), four confident-wrong writes now deferrals, including the visible cost above. `./test/shell` 2513 assertions green with the same 14 environmental files; `./test/cli` and all six monitor suites green; timing unchanged, 18.4 ms shipped against 21.5 on `quattro`.

Your two non-reproducing codex items match what I see at this head as well — both were fixed in the revisions its first pass predated.

Waiting on you for whatever the next pass finds, then on the maintainer for the delegation trades — the expression-output blast radius, and now the uncalled-helper cost, which is the same trade at its third occurrence.

mkelk:
A remark on method, since this PR changed shape rather than iterating once more.

Review here kept finding the same defect in new spellings: a reader of `monitors.lua` took something for Lua that Lua doesn't — a line for a statement, a physical layout for a scope. Each fix covered the reported spelling and left the next one. After the third reader (regexes → an awk grammar walker → a sandboxed Lua execution) we took two steps back and found the readers existed only to feed an explicit `hl.monitor` eval on top of a reload that already re-applies the config — an override that should not exist. So the fix became declarative: author one overlay, remove it to enable, never read the config. The parsing bug class is gone by construction rather than by another round, and the earlier parsing review threads are addressed by removal rather than by fix.

The process changed with it: the invariant was designed and agreed first (committed as `docs/clamshell.md`), the tests were derived from the invariant and shown to fail against the old command for the predicted reasons before the change was written, the assumptions about Hyprland were then verified against a real compositor in the ISO acceptance VM — two were corrected by measurement (auto-position and reload-exit semantics, recorded in the doc and asserted in the acceptance test) — and an independent adversarial review from the invariant alone found and closed one serialization blocker (TERM→KILL escalation, checked lock acquisition) before this push.

The branch was force-pushed over the reader history and narrows to six commits.

mkelk:
Re-verified and rebuilt at `99c14ce9`, answering the re-review at `aef12949` in full. Each finding was reproduced before it was fixed.

**The enable path repairs again — deliberately, and observed.** You were right that the rewrite had removed the accidental repair `sync_internal_scale` provided, and that the killed-helper chain strands a machine. The enable path now asks the compositor (never the config) when the overlay is absent, and reloads once so the config governs again — with the wake following only a panel the post-reload query reports enabled, every unknown (including a query that prints a valid answer and then hangs or fails) repairing nothing, and the retry cost stated honestly: one reload per reconciliation invocation, up to four for a monitor-change retry chain. Getting there surfaced two production bugs in `omarchy-hyprland-monitor-external-active` that had made both this repair and the existing helpers' own guards unreachable: `jq -e`'s exit 4 read "no active external" as unknown, and Hyprland's reserved `FALLBACK` placeholder counted as an active external in exactly the stranded states. Both fixed in the predicate; the acceptance test stages the stranded state against the real `FALLBACK` and observes the repair's effects through a forwarding `hyprctl` wrapper — one real reload, one wake, counted.

**The lock cannot be held by a descendant.** Reproduced your measurement: with the descriptor no longer lent to bounded callees (`9>&-`) the lock is free the moment the command exits, and opening the lock file is checked so an inherited fd 9 can never stand in for it. The focused test covers the surviving-descendant and unopenable-lock cases.

**The smaller items**: the output-name test has its own `XDG_RUNTIME_DIR`; the deleted scale test's one non-parser case (the default-scale recovery) is exactly what the repair now restores, asserted properly; closing keywords are in the body as you suggested, with #7326/#8103 left to `omarchy-hyprland-monitor-scaling`.

mkelk:
Answered at `ebd47f66`, on top of your `dcd744de` — thanks for that one; the vacuous assertion was real, and the strengthened case is exactly right.

**The SIGKILL debris is swept.** Any `.internal-monitor-clamshell.*` file found at the start of a run is a killed run's leftover — under the lock no live run owns one — and is removed by the next run that passes name validation, before either hosted helper can end that run early (the unsafe-name refusal rightly stays first). The focused test plants both kinds and requires them gone on a transitioning run, a no-op run, and a run a failing helper cuts short; removing the sweep or moving it after the helpers each fail the cases that name them.

**The disable-path skip: accepted, stated, and deliberately not closed.** You're right that "converges" was wrong for this side, and the residual is now stated plainly in `docs/clamshell.md`: a run killed between placing the overlay and its reload leaves the desired bytes with Hyprland never told, and later runs do not reload — lit panel behind a closed lid until the next lid cycle or any reload from elsewhere. It predates this PR, self-heals, and costs light rather than a display. The straightforward closure — the disable-path analogue of the repair — would put a `hyprctl monitors` query on every docked 2 s poll and is declined rather than grow this PR again; a durable pending-reload marker cleared after a successful reload is another possible design (more transaction state, a possible harmless duplicate reload), also not taken up here. The doc names both.

**The mirror half named honestly**: the docs now say the repair covers the disable toggle's residual only — its gate opens on a compositor-disabled panel, and a killed mirror helper leaves none.

**The two truthiness callers**: agreed, and now stated in the tracked doc rather than only conceded here — both recover helpers read the predicate as plain truthiness, so their behavior is unchanged, but it still treats unknown like false: a transient query failure while an external is active can still remove the user's toggle. Not a regression (your exit-4 measurement matches ours: the old predicate behaved the same), and the tri-state read is the pattern when it is taken up; it touches two other commands and is out of this PR's scope. The PR body's "truthiness callers are unaffected" is corrected to say exactly this.

The poll-amplification note is noted as what it is — a pre-existing watcher gap that now has a small new cost attached; nothing in this PR relies on the poll surviving a missed event.

kvnloo:
I took the hosted-helper residual already called out in `docs/clamshell.md` and exercised it against this exact head (`ebd47f66`). There are two small pieces here, one expected and one I did not expect.

1. Both recovery helpers still read `omarchy-hyprland-monitor-external-active` as ordinary truthiness. With this PR's new contract, exit `2` is *unknown*, but:

* `omarchy-hyprland-monitor-internal recover`
* `omarchy-hyprland-monitor-internal-mirror recover`

both currently let that unknown reach the same path as the contracted false. A transient failed monitor query can therefore remove the user's disable/mirror toggle even though the external display may still be active.

2. The internal recovery helper also masks a failed mutation. It does:

```bash
omarchy-hyprland-toggle $TOGGLE off
wake
```

and `wake()` is best-effort (`... || true`). If removing the toggle fails, the successful wake becomes the function's exit status, so this PR's caller (`recover ... || exit 0`) cannot enforce the documented "failed hosted helper stops before the clamshell branch" transaction boundary. The mirror helper already propagates its toggle failure.

I put the narrow follow-up on top of this PR head here:

https://github.com/kvnloo/omarchy/commit/ef16db7550c05dd772910ccd05c459f130ce0db2

The behavioral fixture drives both helpers with external-active statuses **0 / 1 / 2 / 124** and an injected toggle-removal failure. Only `1` is allowed to mutate; unknown/timeout are no-ops; a failed mutation stays nonzero; and the internal helper does not wake after that failure. 10 focused assertions pass in the isolated harness.

I kept it separate because the tri-state residual is explicitly out of scope here, but the status-masking half intersects this PR's stated helper-failure invariant directly. Happy for either/both pieces to stay follow-up work if that keeps this branch narrower.

PR #7151, opened 2026-08-16
Eliminate shell spawning latency on bar widget clicks

### LINKED ISSUE #7093: Clicking workspaces in BarWidget is sluggish vs keybinds
### System details

Omarchy 4.0.0-1, Thinkpad Carbon X1 gen 13

## What happened

`Commons/Util.qml`'s `execDetached` wraps every dispatched command in a **login shell**:

```qml
function execDetached(command) {
    Quickshell.execDetached(["bash", "-lc", command])
}
```

`bash -lc` sources `/etc/profile` and the user's `~/.bash_profile`/`~/.profile` on every single invocation, on top of the fork/exec itself. Any bar interaction that goes through `Bar.run()` → `execDetached` pays this cost on every call — most noticeably `Workspaces.qml`'s `focusWorkspace`, which shells out to `hyprctl dispatch` on every click:

```qml
function focusWorkspace(id) {
    if (!root.bar) return
    root.bar.run("hyprctl dispatch " + Util.shellQuote("hl.dsp.focus({ workspace = \"" + id + "\" })"))
}
```

Clicking a workspace pill in the bar is perceptibly slower than switching workspaces via a native Hyprland keybind, which is handled entirely inside the compositor with no IPC round trip and no shell involved at all.

## Expected

Bar-driven Hyprland dispatches (workspace switching, and likely other `Bar.run()` call sites) should feel as instant as the equivalent keybind, since Quickshell already maintains a live IPC connection to Hyprland via the `Quickshell.Hyprland` import.

## Steps to reproduce

1. Fresh Omarchy 4.0.0 install, default Quickshell bar
2. Switch workspaces via a keybind (e.g. `SUPER + 2`) — instant
3. Switch workspaces by clicking a workspace pill in the bar — noticeably slower, subjectively closer to ~100ms+ than the keybind's near-zero latency

## Evidence

Traced the call path from the bar widget down to the actual spawn:

- `Workspaces.qml` → `root.bar.run(...)`
- `Bar.qml`'s `run()`:
```qml
  function run(command) {
      if (!command) return
      Util.execDetached(command)
  }
```
- `Util.qml`'s `execDetached()`:
```qml
  function execDetached(command) {
      Quickshell.execDetached(["bash", "-lc", command])
  }
```

So every workspace click is: bar widget → `bash -lc` (login shell, sources profile files) → `hyprctl` (separate process, parses args, opens its own connection to Hyprland's IPC socket) → dispatch. The keybind path skips all of this — it's handled natively inside Hyprland.

Confirmed the fix locally by cloning the plugin (`omarchy-plugin-clone omarchy.workspaces --edit`) and replacing the function:

**Before:**
```qml
function focusWorkspace(id) {
    if (!root.bar) return
    root.bar.run("hyprctl dispatch " + Util.shellQuote("hl.dsp.focus({ workspace = \"" + id + "\" })"))
}
```

**After:**
```qml
function focusWorkspace(id) {
    Hyprland.dispatch("hl.dsp.focus({ workspace = \"" + id + "\" })")
}
```

`Hyprland.dispatch()` (already available via the `Quickshell.Hyprland` import at the top of `Workspaces.qml`) reuses Quickshell's existing persistent socket connection instead of spawning a shell + `hyprctl` process. After this change, clicking a workspace pill feels identical to using the keybind.

## Suggested fix

Two independent fixes, either of which helps:

1. **In `Workspaces.qml`:** use `Hyprland.dispatch(...)` directly instead of routing through `Bar.run()`/`execDetached`, as shown above.
2. **In `Util.qml`:** `execDetached` defaulting to a *login* shell (`bash -lc`) for arbitrary commands adds unnecessary profile-sourcing overhead to every bar-triggered command, not just Hyprland dispatches. A non-login shell (`bash -c`) would avoid this for the general case, though callers that specifically need Hyprland IPC are better served by `Hyprland.dispatch()` regardless.

YousefAliUK:
Fixes #7093
Credit to @ohai89 for the detailed issue report and initial diagnosis.

### What happened
Clicking workspace pills, the main menu button, or middle-clicking the microphone widget had a noticeable delay (~100ms+) compared to using keybindings. 

This happened because those click actions were routing through `Bar.run()`, which calls `Util.execDetached()`. That spawns `bash -lc` on every click, forcing bash to re-source `/etc/profile` and `~/.bash_profile` every single time, on top of the fork/exec overhead and opening redundant IPC sockets.

### Changes
1. **Workspaces (`Workspaces.qml`)**: switched `focusWorkspace` to use `Hyprland.dispatch(...)` directly. Quickshell already imports `Quickshell.Hyprland` and keeps an open socket connection to Hyprland, so this switches workspaces instantly with no process spawning.
2. **Menu & Mic (`menu/BarWidget.qml`, `Microphone.qml`)**: switched click actions to call `root.bar.shell.toggle(...)` directly in-process instead of running `omarchy-shell shell toggle ...` through bash. Kept the CLI call as a fallback in case `root.bar.shell` is unset.
3. **Why `Util.execDetached` wasn't changed**: I decided against changing `Util.execDetached` globally from `bash -lc` to `bash -c` since external tools or custom user scripts might rely on login shell profiles for `$PATH` and env vars. Internal bar clicks shouldn't be using `execDetached` anyway.

### Testing
Added `test/shell.d/bar-fast-dispatch-test.sh` to verify direct Hyprland dispatch and in-process shell toggles. Tests pass.

### DIFF
--- a/shell/plugins/bar/widgets/Microphone.qml
+++ b/shell/plugins/bar/widgets/Microphone.qml
@@ -42,8 +42,15 @@ BarWidget {
     active: root.inUse
     tooltipText: root.muted ? "Microphone muted" : (root.inUse ? "Microphone in use" : "Microphone live")
     onPressed: function(b) {
-      if (b === Qt.MiddleButton) root.bar.run("omarchy-shell shell toggle omarchy.audio")
-      else root.toggleMute()
+      if (b === Qt.MiddleButton) {
+        if (root.bar && root.bar.shell && typeof root.bar.shell.toggle === "function") {
+          root.bar.shell.toggle("omarchy.audio")
+        } else if (root.bar) {
+          root.bar.run("omarchy-shell shell toggle omarchy.audio")
+        }
+      } else {
+        root.toggleMute()
+      }
     }
     onWheelMoved: function(delta) {
       if (!root.source || !root.source.audio) return
--- a/shell/plugins/bar/widgets/Workspaces.qml
+++ b/shell/plugins/bar/widgets/Workspaces.qml
@@ -31,8 +31,7 @@ BarWidget {
   }
 
   function focusWorkspace(id) {
-    if (!root.bar) return
-    root.bar.run("hyprctl dispatch " + Util.shellQuote("hl.dsp.focus({ workspace = \"" + id + "\" })"))
+    Hyprland.dispatch("hl.dsp.focus({ workspace = \"" + id + "\" })")
   }
 
   readonly property real trailingGap: root.vertical ? 0 : Style.spaceReal(1.5)
--- a/shell/plugins/menu/BarWidget.qml
+++ b/shell/plugins/menu/BarWidget.qml
@@ -17,8 +17,13 @@ BarWidget {
     horizontalMargin: 7.5
     onPressed: function(button) {
       if (!root.bar) return
-      if (button === Qt.RightButton) root.bar.run("xdg-terminal-exec")
-      else root.bar.run("omarchy-shell shell toggle omarchy.menu '{\"menu\":\"root\"}'")
+      if (button === Qt.RightButton) {
+        root.bar.run("xdg-terminal-exec")
+      } else if (root.bar.shell && typeof root.bar.shell.toggle === "function") {
+        root.bar.shell.toggle("omarchy.menu", JSON.stringify({ menu: "root" }))
+      } else {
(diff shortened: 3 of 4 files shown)


bjarneo:
Automated duplication check: this pull request looks similar to #7011, which covers the same bar click direct dispatch. I keep that one open and close this one to consolidate review.

PR #7152, opened 2026-08-16
Add drag-to-arrange monitor layout in Display

sharifli4:
## Summary
- Replace the Display panel's monitor list with a spatial canvas: drag numbered rectangles to snap left/right/above/below, then apply live via `hyprctl` and persist a managed block in `~/.config/hypr/monitors.lua`.
- Add `omarchy-hyprland-monitor-layout` (`omarchy hyprland monitor layout`) for `state`, `apply`, and `place`, with overlap/island validation and per-write backups.
- Snap from the drop position (not travel distance) so a display dragged under a wider neighbor stays on that bottom edge.

## Test plan
- [x] `test/shell.d/hyprland-monitor-layout-test.sh` (state, validate, lua patch, CLI dry-run/write, snap math)
- [x] `./test/cli` (command metadata / routing)
- [ ] Open Super+Ctrl+D with laptop + external: drag a display under/beside the other and confirm it stays
- [ ] Confirm `~/.config/hypr/monitors.lua` gains a single `-- omarchy-layout-managed` block and GDK_SCALE / catch-all `output = ""` stay untouched
- [ ] Laptop-only: canvas shows one display; apply is a no-op
- [ ] LEFT/RIGHT/ABOVE/BELOW chips and INT/MIRROR toggles still work

Made with [Cursor](https://cursor.com)

### DIFF
--- a/bin/omarchy-hyprland-monitor-layout
+++ b/bin/omarchy-hyprland-monitor-layout
@@ -1,0 +1,386 @@
+#!/bin/bash
+
+# omarchy:summary=Read or apply a snapped Hyprland monitor layout
+# omarchy:args=state|apply '<json>'|place [NAME] <left|right|top|bottom>
+# omarchy:examples=omarchy hyprland monitor layout state | omarchy hyprland monitor layout place right
+
+set -euo pipefail
+
+LAYOUT_START="-- omarchy-layout-managed:start"
+LAYOUT_END="-- omarchy-layout-managed:end"
+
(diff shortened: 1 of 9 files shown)


sharifli4:
I kept having to rearrange my laptop and external monitor, and doing that by hand was a pain. I wanted a GUI for it
image

sharifli4:
Hey @dhh , could you give feedback, please

PR #7157, opened 2026-08-16
Add OpenCode Go to AI usage widget

AI091:
## Summary

Adds an OpenCode Go tab to the agents panel.

OpenCode Go panel

## Changes

- `bin/omarchy-agent-usage-opencode-go` — collector
- `shell/plugins/agents/assets/opencode-go{,-light}.svg` — logo
- `manifest.json`, `README.md`, `manual/17-ai.md` — defaults + docs
- `test/shell.d/agent-usage-opencode-go-scanner-test.sh` — tests

Built with OpenCode Go :), Kimi K3 & DeepSeek V4 Pro

### DIFF
--- a/bin/omarchy-agent-usage-opencode-go
+++ b/bin/omarchy-agent-usage-opencode-go
@@ -406,7 +406,9 @@ def fetch_limits(api_key_value: str, usage_url: str) -> tuple[list[Any], str, st
       return cached, "OpenCode Go limits stale", "", True
     return [], "OpenCode Go limits unavailable", "Could not reach opencode.ai.", True
 
-  usage = payload.get("usage") if isinstance(payload, dict) else {}
+  usage = payload.get("usage") if isinstance(payload, dict) else None
+  if not isinstance(usage, dict):
+    usage = {}
   windows = [
     ("Rolling (5-hour)", usage.get("rolling")),
     ("Weekly (7-day)", usage.get("weekly")),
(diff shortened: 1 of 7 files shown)


Brams-s:
Thanks for the focused provider implementation. The collector shape, provider IDs, malformed-row handling, and targeted shell coverage look consistent with the existing agents architecture.

A few non-blocking follow-ups are worth considering before merge:

- The limits cache is currently reusable after its reset time has passed. During a transient API failure, this could display an old percentage for a window that has already reset. Claude's collector already has a resetsAt-aware cache usability check that could be reused or mirrored here.
- The cache path is not account-scoped. If the API key changes and the next request fails transiently, the previous account's cached limits could be shown. A stable account/key fingerprint in the cache filename or payload would prevent that.
- The scanner currently handles the v1 message table, while newer OpenCode databases may use session_message and store the provider under model.providerID. Reporting a clean zero is safe, but an explicit schema/version status would make no data distinguishable from an unsupported database schema.
- created <= 0 currently appears able to produce an invalid historical day. Treating non-positive timestamps as today, as the Claude collector does, would avoid polluting active-day and chart data if malformed rows occur.

These are follow-ups rather than blockers, and I have not confirmed that every case occurs in production OpenCode databases. The existing tests appear strong; the remaining validation gap noted in the PR is visual verification of the panel.

bjarneo:

PR #7158, opened 2026-08-16
Keep the lock screen fingerprint working across suspend, and show when the reader is unavailable

### LINKED ISSUE #7172: Lock screen fingerprint retry livelocks after resume: overlapping PAM subprocesses collide on exclusive fprintd claim
## What happens

On a laptop with a fingerprint reader, unlocking by fingerprint intermittently stops working entirely. The lock screen shows no fingerprint progress and only a typed password gets you in.

Under the hood the lock screen enters a self-sustaining retry livelock: it spawns 3–5 overlapping `pam.subprocess` instances per second, every one failing to claim the fingerprint device, until the user gives up and types their password.

Captured from a real failure (13 seconds, 41 attempts, all failing):

```
19:20:23  Lid closed  ->  PM: suspend entry (s2idle)
19:20:24  PM: suspend exit  ->  Lid opened            (1-second suspend)
19:20:24  omarchy-shell[182126]: Starting pam session ... "omarchy-lock-fingerprint"
19:20:24  omarchy-shell[182167]: Starting pam session ... "omarchy-lock-fingerprint"
19:20:24  omarchy-shell[182170]: Starting pam session ... "omarchy-lock-fingerprint"
          ... 41 sessions in 13s, each:
          Error while authenticating: "Authentication service cannot
          retrieve authentication info" (code 9)   [PAM_AUTHINFO_UNAVAIL]
19:20:36  falls back to "omarchy-lock-password"  ->  user types password
```

Concurrent distinct PAM subprocess PIDs per second across that window: 5, 3, 3, 4, 3, 3, 3, 4, 3, 3, 4, 4, 3.

A healthy locked session sits at ~2 attempts/minute by comparison. Over one day: 116 fingerprint PAM sessions vs 5 password fallbacks.

## Root cause

`shell/plugins/lock/Service.qml:209-228`:

```qml
function startFingerprint() {
  if (!lockRequested || !sessionLock.secure || !fingerprintConfigured) return
  if (fingerprintPam.active || fingerprintAuthenticating) return   // :211

  fingerprintAuthenticating = true
  ...
}

function handleFingerprintFinished(result) {
  fingerprintAuthenticating = false        // :220 — cleared on RESULT
  if (!lockRequested) return
  if (result === PamResult.Success) {
    finishUnlock()
  } else if (fingerprintConfigured) {
    fingerprintRetryTimer.restart()        // :226 — fires 250ms later
  }
}
```

with

```qml
Timer {
  id: fingerprintRetryTimer
  interval: 250       // :358 — flat, no backoff
  repeat: false
  onTriggered: root.startFingerprint()
}
```

Both guards on line 211 clear as soon as PAM *reports a result*. Neither tracks whether the forked `pam.subprocess` has actually exited and released its claim on the fprintd device.

fprintd devices are exclusive-claim. Subprocess teardown takes longer than the 250 ms retry interval, so the retry starts a fresh subprocess that tries to claim a device the previous one still holds. That claim fails, which immediately calls `handleFingerprintFinished`, which restarts the timer — now with two subprocesses in teardown. The loop feeds itself and never drains.

The failure mode is confirmed directly against fprintd. Running `fprintd-verify` manually at the moment of the storm returns the underlying error the PAM code 9 is masking:

```
$ fprintd-verify
Using device /net/reactivated/Fprint/Device/0
failed to claim device: GDBus.Error:net.reactivated.Fprint.Error.Internal:
  Open failed with error: The device is still busy with another operation,
  please try again later.
```

The same command before suspend and again 10 s later both succeed with `verify-match`, exit 0 — so the reader itself is healthy throughout.

The livelock needs one seed failure to enter. Resume from suspend supplies it: the device is genuinely unavailable for a moment as USB re-enumerates (`usb 3-7: reset full-speed USB device number 3 using xhci_hcd` on every resume). A clean lock succeeds on the first attempt and never enters the loop, which is why this presents as random.

## Expected

A transient claim failure should back off and retry, not spin at 4 Hz spawning overlapping subprocesses that guarantee further claim failures.

## Steps to reproduce

1. Laptop with an enrolled fingerprint and `omarchy-lock-fingerprint` in use.
2. Close the lid and reopen it within about a second, so the machine suspends and resumes immediately.
3. Try to unlock by fingerprint.
4. Watch `journalctl -f | grep -E 'omarchy-shell\[|code 9'` — observe multiple concurrent PAM subprocesses per second, all returning code 9.

Not 100% reproducible per attempt, since it depends on catching the device mid-re-enumeration. Roughly 2 storms per day in normal use here.

## Suggested fix

Three things, any of which breaks the loop:

- Give `fingerprintRetryTimer` exponential backoff (e.g. 250 ms → 2 s ceiling) instead of a flat 250 ms.
- Gate `startFingerprint()` on actual subprocess exit rather than on the PAM result, so a still-releasing claim can't be raced.
- Distinguish a *claim failure* (device busy/unavailable — retry with backoff) from a *failed match* (finger rejected — normal retry cadence). Right now both funnel through the same path.

## System details

- Omarchy 4.0.0-1
- Kernel 7.1.8-arch1-3
- Intel Core Ultra X7 358H
- Reader: Goodix MOC `27c6:609c` (Framework Laptop)
- fprintd 1.94.5-2, libfprint 1.94.100-1
- Suspend type: s2idle

Note: USB autosuspend was ruled out as a factor — the reader was pinned to `power/control=on` during testing and the livelock still occurred.

### LINKED ISSUE #7229: Fingerprint reader fails after suspend: fprintd reports "Cannot run while suspended"
## Description

After the system suspends and resumes, the fingerprint reader stops working on the lock screen. The lock screen's fingerprint prompt either hangs or immediately fails, forcing password fallback.

## Steps to Reproduce

1. Unlock the lock screen using fingerprint — works fine
2. Let the system suspend (via idle, lid close, or `Super+Ctrl+L` then suspend)
3. Resume the system
4. Lock screen appears — attempt to use fingerprint to unlock
5. Fingerprint authentication fails / is unresponsive

## Expected Behavior

Fingerprint authentication works on the lock screen after resume from suspend.

## Actual Behavior

`fprintd` reports the device error and authentication fails silently:

```
fprintd[PID]: Device reported an error during verify: Cannot run while suspended.
```

This appears in `journalctl -u fprintd` consistently after every suspend/resume cycle. The error persists until `fprintd` is manually restarted (`systemctl restart fprintd`), after which fingerprint works again.

## Root Cause

`fprintd` holds a stale/broken device handle after the fingerprint reader is suspended by the kernel. On resume, `fprintd` doesn't re-probe the device — it just fails all verify calls until restarted.

## Suggested Fix

Add a `/lib/systemd/system-sleep/restart-fprintd` hook (following the same pattern as the existing `unmount-fuse` hook in that directory) to restart `fprintd` a few seconds after resume:

```bash
#!/bin/bash
# Restart fprintd after resume so the fingerprint reader recovers from suspend.
if [[ $1 == "post" ]]; then
  (sleep 3; systemctl restart fprintd.service) &
fi
```

The 3-second delay gives the device time to re-enumerate before `fprintd` tries to claim it. Manually running `systemctl restart fprintd` after each resume confirms this fix works.

## System Details

| Field | Value |
|-------|-------|
| Omarchy version | 4.0.0-1 |
| Kernel | 7.1.8-arch1-3 |
| CPU | Intel Core Ultra 7 155U (Meteor Lake) |
| GPU | Intel Meteor Lake-P (integrated) |
| Fingerprint sensor | Goodix MOC Fingerprint Sensor (integrated) |
| fprintd | 1.94.5-2 |
| libfprint | 1.94.100-1 |

## Relevant Logs

```
Aug 16 00:17:15 fprintd[24755]: Device reported an error during verify: Cannot run while suspended.
Aug 16 11:28:54 fprintd[69990]: Device reported an error during verify: Cannot run while suspended.
Aug 17 (multiple occurrences after each suspend/resume cycle)
```

GeertJohan:
### Problem

A fingerprint verify still open when the machine suspends leaves fprintd unable to hand the reader back — the verify dies with `Cannot run while suspended` and the follow-up `ReleaseDevice` fails on the still-busy device:

```
fprintd: Device reported an error during verify: Cannot run while suspended.
pam_fprintd: ReleaseDevice failed: … The device is still busy with another operation
```

The wedged claim then rejects every lock-screen attempt after resume. The lock retried on a flat 250ms timer with no sign to the user, so the reader looked dead behind an icon still inviting touches — ~4 PAM sessions/second that also kept fprintd from reaching the 30-second idle exit that would have cleared the claim on its own.

### Fix — three layers

1. **Remove the wedge at its source.** A `post` system-sleep hook restarts fprintd on resume (installed by `omarchy-setup-security-fingerprint`, with a migration for existing setups). It runs while `user.slice` is frozen, so fprintd is fresh before the lock screen retries, and it's bounded with `timeout` so a wedged restart can't hold the thaw open. `try-restart` is a no-op when fprintd isn't running, so a healthy resume where nothing held the reader pays nothing.
2. **Stop hammering an unreachable reader.** Retries now back off exponentially on the signal of *whether an attempt reached the reader at all* — pam_fprintd relays a finger prompt only once the claim lands, so an attempt that ends without prompting never reached the device (wedged, held by another client, or gone). A finger that merely doesn't match still reaches the reader and retries fast.
3. **Tell the user.** After a few consecutive unreached attempts the icon crosses out and a *"Fingerprint reader unavailable"* notice appears under the field — cleared the instant a prompt shows the reader is reachable again, and a keypress or touch collapses the backoff so a recovered reader is tried at once. This covers *any* cause of an unreachable reader, not just suspend.

### Also clears stale claims across suspend

The restart drops *every* fprintd claim, so a background fingerprint request left waiting across the suspend — a hidden `fprintd-verify`, or an app/polkit prompt you never returned to — no longer survives resume to silently consume your first unlock touch. A *live* verify caught across the suspend (e.g. a `sudo` prompt on a TTY) is likewise dropped and falls back to a password; that's the intended trade. (A background client competing for the reader *without* a suspend is a separate arbitration question, out of scope here.)

### Reproducing

**Deterministically (any machine with an enrolled finger)** — the pacing + notice:

1. Hold the reader from a second process: run `fprintd-verify` and leave it waiting (don't touch the sensor — a touch completes it).
2. Watch: `journalctl -f -t omarchy-shell -t fprintd`
3. `omarchy system lock`.
4. **Before:** ~4 attempts/sec, each a `Starting pam session` (shell) + `Authorization denied … Device was already claimed` (fprintd), behind a normal icon.
5. **After:** a handful of attempts at growing intervals; after the third the icon crosses out with the notice. Kill the holder while still locked and the notice clears within one retry.

**The real trigger (probabilistic)** — the wedge + hook:

1. With the reader idle, close the lid. The pre-suspend lock starts a verify still waiting when the machine sleeps.
2. `journalctl -u fprintd -b` after resume shows the losing rolls: `Cannot run while suspended` + a failed `ReleaseDevice`. The resume hook restarts fprintd so the next touch works regardless.

### Scope / relation to #7179

This paces and reports the *claim-reachability* failure family — a reader that is wedged, held by another client, or gone. It deliberately does **not** touch failures that arrive *after* the claim lands: any error following the finger prompt reads as "reached", so a reader that refuses mid-verify (libfprint's thermal duty-cycle cut-off), disconnects, or whose driver errors gets neither the backoff nor the notice from this PR. The thermal slice of that is #7179's cause and signal. The two overlap only on the retry path in `Service.qml`; the levers are complementary (remove/observe the wedge here, pace the hot verify there).

Of #7176's two reports, the orphaned-claim one (no suspend, fprintd held for days) is this failure family and is resolved here; the thermal one in the title is not, and stays with #7179.

### Testing

`FingerprintModel.js` has a Node test for the backoff and streak logic; the lock-view fixture covers the crossed icon and notice; the hook and migration each have a shell test (including that the hook is installed executable — systemd-sleep won't run it otherwise). `lock status` now reports `fingerprintUnavailable`.

_Screenshots below: the reader available vs. unavailable (Catppuccin, whose `lock.text-error` is red; a few themes like Tokyo Night map it to the foreground, so the notice there matches the icon rather than standing out)._

Fixes #7229
Fixes #7172
Partially addresses #7176 (orphaned-claim case; thermal case is #7179)

### DIFF
--- a/shell/plugins/lock/Service.qml
+++ b/shell/plugins/lock/Service.qml
@@ -28,6 +28,7 @@ Item {
   property double fingerprintLastNudgeMs: 0
   property double fingerprintLastSettleMs: 0
   property double fingerprintResumedAtMs: 0
+  property int fingerprintProbeStreak: 0
   property bool previewVisible: false
   property string enteredPassword: ""
   property string pendingPassword: ""
@@ -130,6 +131,35 @@ Item {
     if (!fingerprintCheckProc.running) fingerprintCheckProc.running = true
   }
 
+  // An unreachable fprintd -- restarting under the resume hook, or still
+  // activating -- answers the probe with an error, not with "no prints".
+  // Concluding "not configured" from that stopped the retry loop and the
+  // sleep watch for the rest of the lock (#9453); keep the current state and
+  // ask again instead. Only a definitive answer changes anything.
+  function applyFingerprintProbe(text) {
+    var status = FingerprintModel.classifyProbe(text)
+    if (status === "unknown") {
+      fingerprintProbeStreak += 1
+      if (lockRequested) {
+        fingerprintRecheckTimer.interval = FingerprintModel.retryDelayMs(fingerprintProbeStreak)
+        fingerprintRecheckTimer.restart()
+      }
+      return
+    }
+    fingerprintProbeStreak = 0
+    fingerprintConfigured = status === "yes"
+    if (lockRequested && fingerprintConfigured) {
+      // A pending retry already owns the next attempt.
+      if (!fingerprintRetryTimer.running) startFingerprint()
+    } else if (!fingerprintConfigured) {
+      // abort() delivers no completion signal, so close the attempt here
+      // too; settle returns before arming a retry while unconfigured.
+      if (fingerprintPam.active) fingerprintPam.abort()
+      settleFingerprintAttempt()
+      fingerprintRetryTimer.stop()
+    }
+  }
+
   function logEvent(event) {
     lastEvent = event
     lastEventAt = new Date().toISOString()
@@ -147,6 +177,8 @@ Item {
     fingerprintLastNudgeMs = 0
     fingerprintLastSettleMs = 0
     fingerprintResumedAtMs = 0
+    fingerprintProbeStreak = 0
+    fingerprintRecheckTimer.stop()
     fingerprintRetryTimer.stop()
     fingerprintReachTimer.stop()
     if (passwordPam.active) passwordPam.abort()
@@ -569,28 +601,23 @@ Item {
     }
   }
 
-  // Configured means a print is actually enrolled: fprintd-list prints one
-  // " - #N: finger" line per print, while a user without any gets "has no
-  // fingers enrolled" or a ListEnrolledFingers failure, both of which a looser
-  // match would take as yes and then chase with a reader the account cannot
-  // use.
+  // The probe hands fprintd-list's raw output (errors included) to
+  // classifyProbe, which answers yes, no, or unknown; only a definitive
+  // answer may change fingerprintConfigured. See applyFingerprintProbe.
   Process {
     id: fingerprintCheckProc
-    command: ["bash", "-c", "if [[ -f /etc/pam.d/omarchy-lock-fingerprint ]] && command -v fprintd-list >/dev/null 2>&1 && fprintd-list \"$USER\" 2>/dev/null | grep -q ' - #'; then echo yes; else echo no; fi"]
+    command: ["bash", "-c", "if [[ -f /etc/pam.d/omarchy-lock-fingerprint ]] && command -v fprintd-list >/dev/null 2>&1; then fprintd-list \"$USER\" 2>&1; else echo no; fi"]
     stdout: StdioCollector { id: fingerprintCheckStdout; waitForEnd: true }
-    onExited: {
-      root.fingerprintConfigured = String(fingerprintCheckStdout.text || "").trim() === "yes"
-      if (root.lockRequested && root.fingerprintConfigured) {
-        // A pending retry already owns the next attempt.
-        if (!fingerprintRetryTimer.running) root.startFingerprint()
-      } else if (!root.fingerprintConfigured) {
-        // abort() delivers no completion signal, so close the attempt here
-        // too; settle returns before arming a retry while unconfigured.
-        if (fingerprintPam.active) fingerprintPam.abort()
-        root.settleFingerprintAttempt()
-        fingerprintRetryTimer.stop()
-      }
-    }
+    onExited: root.applyFingerprintProbe(fingerprintCheckStdout.text)
+  }
+
+  // Retries a probe that could not reach fprintd, paced like the attempt
+  // retries so a daemon that stays unreachable is asked about ever less often.
+  Timer {
+    id: fingerprintRecheckTimer
+    interval: FingerprintModel.ERROR_RETRY_BASE_MS
+    repeat: false
+    onTriggered: root.refreshFingerprintStatus()
   }
 
   Process {
--- a/shell/plugins/lock/Service.qml
+++ b/shell/plugins/lock/Service.qml
@@ -293,9 +293,11 @@ Item {
       refreshFingerprintStatus()
       return
     }
-    // Bound the wait for the first prompt: a claim that never lands (a daemon
-    // restarted under the verify, e.g. by the resume hook) otherwise hangs
-    // here without ever erroring, and nothing downstream re-arms.
+    // Bound the wait for the first prompt: a claim fprintd accepts but never
+    // completes (a device open stuck behind a wedged claim) otherwise sits here
+    // for GDBus's full call timeout without erroring, and nothing downstream
+    // re-arms. A daemon restarted under the verify is not that case; it fails
+    // the attempt promptly. See REACH_TIMEOUT_MS for the bound's limits.
     fingerprintReachTimer.restart()
   }
 
--- a/shell/plugins/lock/Service.qml
+++ b/shell/plugins/lock/Service.qml
@@ -26,6 +26,7 @@ Item {
   property int fingerprintUnreachedStreak: 0
   property bool fingerprintAttemptReachedDevice: false
   property double fingerprintLastNudgeMs: 0
+  property double fingerprintLastSettleMs: 0
   property bool previewVisible: false
   property string enteredPassword: ""
   property string pendingPassword: ""
@@ -143,6 +144,7 @@ Item {
     fingerprintAuthenticating = false
     fingerprintUnreachedStreak = 0
     fingerprintLastNudgeMs = 0
+    fingerprintLastSettleMs = 0
     fingerprintRetryTimer.stop()
     fingerprintReachTimer.stop()
     if (passwordPam.active) passwordPam.abort()
@@ -200,13 +202,14 @@ Item {
   // trying now, so collapse a backed-off wait to a prompt retry rather than
   // ride out the cap. Rate-limited (see shouldNudge): a moving cursor raises a
   // wake per motion event, and without the floor each fresh backoff wait would
-  // be re-collapsed straight back into the storm the backoff exists to prevent.
+  // be re-collapsed straight back into the storm the backoff exists to prevent;
+  // past the floor, the current tier paces repeat nudges.
   function nudgeFingerprint() {
     if (!lockRequested || !fingerprintConfigured) return
     if (fingerprintPam.active || fingerprintAuthenticating) return
     if (!fingerprintRetryTimer.running) return
     var now = Date.now()
-    if (!FingerprintModel.shouldNudge(now, fingerprintLastNudgeMs, fingerprintRetryTimer.interval)) return
+    if (!FingerprintModel.shouldNudge(now, fingerprintLastNudgeMs, fingerprintLastSettleMs, fingerprintRetryTimer.interval)) return
     fingerprintLastNudgeMs = now
     fingerprintRetryTimer.interval = FingerprintModel.MATCH_RETRY_MS
     fingerprintRetryTimer.restart()
@@ -325,6 +328,7 @@ Item {
     if (!lockRequested || !fingerprintConfigured) return
 
     fingerprintUnreachedStreak = FingerprintModel.nextStreak(fingerprintUnreachedStreak, fingerprintAttemptReachedDevice)
+    fingerprintLastSettleMs = Date.now()
     fingerprintRetryTimer.interval = FingerprintModel.retryDelayMs(fingerprintUnreachedStreak)
     fingerprintRetryTimer.restart()
   }
--- a/shell/plugins/lock/Service.qml
+++ b/shell/plugins/lock/Service.qml
@@ -27,6 +27,7 @@ Item {
   property bool fingerprintAttemptReachedDevice: false
   property double fingerprintLastNudgeMs: 0
   property double fingerprintLastSettleMs: 0
+  property double fingerprintResumedAtMs: 0
   property bool previewVisible: false
   property string enteredPassword: ""
   property string pendingPassword: ""
@@ -145,6 +146,7 @@ Item {
     fingerprintUnreachedStreak = 0
     fingerprintLastNudgeMs = 0
     fingerprintLastSettleMs = 0
+    fingerprintResumedAtMs = 0
     fingerprintRetryTimer.stop()
     fingerprintReachTimer.stop()
     if (passwordPam.active) passwordPam.abort()
@@ -211,8 +213,36 @@ Item {
     var now = Date.now()
     if (!FingerprintModel.shouldNudge(now, fingerprintLastNudgeMs, fingerprintLastSettleMs, fingerprintRetryTimer.interval)) return
     fingerprintLastNudgeMs = now
-    fingerprintRetryTimer.interval = FingerprintModel.MATCH_RETRY_MS
+    armFingerprintRetry(FingerprintModel.MATCH_RETRY_MS)
+  }
+
+  function armFingerprintRetry(delayMs) {
+    fingerprintRetryTimer.interval = delayMs
+    fingerprintRetryTimer.armedAt = Date.now()
     fingerprintRetryTimer.restart()
+  }
+
+  // Monotonic timers pause across suspend, so a wait armed before the sleep
+  // picks up mid-count afterwards -- and the streak it was pacing was built
+  // against the reader as it stood before the sleep. The resume hook is
+  // restarting fprintd, so that streak is stale: drop it, and open the grace
+  // window in which the restart landing under an attempt does not count as a
+  // miss either (see RESUME_GRACE_MS). Idempotent within the window, since a
+  // resume can be noticed by more than one timer.
+  function noteFingerprintResumed() {
+    var now = Date.now()
+    if (FingerprintModel.inResumeGrace(now, fingerprintResumedAtMs)) return
+    fingerprintResumedAtMs = now
+    fingerprintUnreachedStreak = 0
+  }
+
+  // A resume noticed while a backed-off wait is pending: retry the fresh
+  // reader now, instead of after the remaining wait or the next keypress.
+  function restartFingerprintAfterSleep() {
+    noteFingerprintResumed()
+    if (fingerprintAuthenticating || fingerprintPam.active) return
+    if (!fingerprintRetryTimer.running) return
+    armFingerprintRetry(FingerprintModel.MATCH_RETRY_MS)
   }
 
   function runBlank() {
@@ -329,10 +359,22 @@ Item {
     fingerprintReachTimer.stop()
     if (!lockRequested || !fingerprintConfigured) return
 
-    fingerprintUnreachedStreak = FingerprintModel.nextStreak(fingerprintUnreachedStreak, fingerprintAttemptReachedDevice)
-    fingerprintLastSettleMs = Date.now()
-    fingerprintRetryTimer.interval = FingerprintModel.retryDelayMs(fingerprintUnreachedStreak)
-    fingerprintRetryTimer.restart()
+    // The sleep watch ticks every second while locked, so a settle that finds
+    // its last tick far in the past is the first thing to run after a resume:
+    // this attempt was in flight across the suspend and was ended by the
+    // restart, not by the reader. Judged from the tick rather than the
+    // attempt's own age so a suspend shorter than the reach bound is caught
+    // too, before the tick itself gets a chance to.
+    var now = Date.now()
+    if (!fingerprintAttemptReachedDevice && fingerprintSleepWatch.running
+        && FingerprintModel.spannedSleep(now - fingerprintSleepWatch.lastTickMs, fingerprintSleepWatch.interval)) {
+      noteFingerprintResumed()
+    }
+
+    var inGrace = FingerprintModel.inResumeGrace(now, fingerprintResumedAtMs)
+    fingerprintUnreachedStreak = FingerprintModel.nextStreak(fingerprintUnreachedStreak, fingerprintAttemptReachedDevice, inGrace)
+    fingerprintLastSettleMs = now
+    armFingerprintRetry(FingerprintModel.retryDelayMs(fingerprintUnreachedStreak))
   }
 
   function handleFingerprintFinished(result) {
@@ -478,7 +520,31 @@ Item {
     id: fingerprintRetryTimer
     interval: FingerprintModel.MATCH_RETRY_MS
     repeat: false
-    onTriggered: root.startFingerprint()
+    property double armedAt: 0
+    onTriggered: {
+      // A wait that took far longer on the wall clock than its interval
+      // spanned a suspend; see noteFingerprintResumed.
+      if (FingerprintModel.spannedSleep(Date.now() - armedAt, interval)) root.noteFingerprintResumed()
+      root.startFingerprint()
+    }
+  }
+
+  // Watches the wall clock for the whole lock, so a resume is noticed within
+  // a tick whatever the loop was doing -- mid-wait, or with an attempt in
+  // flight that the restart is about to end.
+  Timer {
+    id: fingerprintSleepWatch
+    interval: 1000
+    repeat: true
+    running: root.lockRequested && root.fingerprintConfigured
+    property double lastTickMs: 0
+    onRunningChanged: lastTickMs = Date.now()
+    onTriggered: {
+      var now = Date.now()
+      var slept = FingerprintModel.spannedSleep(now - lastTickMs, interval)
+      lastTickMs = now
+      if (slept) root.restartFingerprintAfterSleep()
+    }
   }
 
   Timer {
--- a/shell/plugins/lock/Service.qml
+++ b/shell/plugins/lock/Service.qml
@@ -271,9 +271,19 @@ Item {
 
   // A resume noticed while a backed-off wait is pending: retry the fresh
   // reader now, instead of after the remaining wait or the next keypress.
+  // An attempt still in flight is worse than a pending wait: the hook has
+  // restarted fprintd under it, so it sits on a dead conversation that
+  // pam_fprintd takes ~25s to give up on (#8747) while the icon invites
+  // touches that cannot work. Abort it and settle -- inside the grace
+  // window, so the kill never counts toward the notice -- and the settle
+  // arms the fast retry itself.
   function restartFingerprintAfterSleep() {
     noteFingerprintResumed()
-    if (fingerprintAuthenticating || fingerprintPam.active) return
+    if (fingerprintAuthenticating || fingerprintPam.active) {
+      if (fingerprintPam.active) fingerprintPam.abort()
+      settleFingerprintAttempt()
+      return
+    }
     if (!fingerprintRetryTimer.running) return
     armFingerprintRetry(FingerprintModel.MATCH_RETRY_MS)
   }
--- a/shell/plugins/lock/Service.qml
+++ b/shell/plugins/lock/Service.qml
@@ -264,6 +264,7 @@ Item {
   function noteFingerprintResumed() {
     var now = Date.now()
     if (FingerprintModel.inResumeGrace(now, fingerprintResumedAtMs)) return
+    logEvent("fingerprint-resume: streak=" + fingerprintUnreachedStreak)
     fingerprintResumedAtMs = now
     fingerprintUnreachedStreak = 0
   }
@@ -377,6 +378,7 @@ Item {
   // advances the streak (and so the notice) and retries against a daemon that
   // may now be fresh, instead of hanging silently behind a normal icon.
   function timeoutFingerprintReach() {
+    logEvent("fingerprint-reach-timeout")
     if (fingerprintPam.active) fingerprintPam.abort()
     settleFingerprintAttempt()
   }
@@ -403,15 +405,30 @@ Item {
       noteFingerprintResumed()
     }
 
+    // Reached attempts are the steady state (one per swipe window), so only
+    // the misses and the recovery from them leave a trace.
+    var previousStreak = fingerprintUnreachedStreak
     var inGrace = FingerprintModel.inResumeGrace(now, fingerprintResumedAtMs)
-    fingerprintUnreachedStreak = FingerprintModel.nextStreak(fingerprintUnreachedStreak, fingerprintAttemptReachedDevice, inGrace)
+    fingerprintUnreachedStreak = FingerprintModel.nextStreak(previousStreak, fingerprintAttemptReachedDevice, inGrace)
+    if (!fingerprintAttemptReachedDevice) {
+      var crossed = !FingerprintModel.isUnavailable(previousStreak) && FingerprintModel.isUnavailable(fingerprintUnreachedStreak)
+      logEvent((crossed ? "fingerprint-unavailable" : "fingerprint-unreached") + ": streak=" + fingerprintUnreachedStreak)
+    } else if (previousStreak > 0) {
+      logEvent("fingerprint-recovered: streak=" + previousStreak)
+    }
     fingerprintLastSettleMs = now
     armFingerprintRetry(FingerprintModel.retryDelayMs(fingerprintUnreachedStreak))
   }
 
   function handleFingerprintFinished(result) {
-    if (result === PamResult.Success && lockRequested) finishUnlock()
-    else settleFingerprintAttempt()
+    if (result === PamResult.Success && lockRequested) {
+      // A match after a run of misses is the recovery too; the unlock resets
+      // the streak without settling, so log it here or it leaves no trace.
+      if (fingerprintUnreachedStreak > 0) logEvent("fingerprint-recovered: streak=" + fingerprintUnreachedStreak)
+      finishUnlock()
+    } else {
+      settleFingerprintAttempt()
+    }
   }
 
   WlSessionLock {
--- a/shell/plugins/lock/Service.qml
+++ b/shell/plugins/lock/Service.qml
@@ -282,7 +282,12 @@ Item {
     fingerprintAuthenticating = true
     fingerprintAttemptReachedDevice = false
     if (!fingerprintPam.start()) {
+      // A start that fails before PAM even runs is a configuration problem
+      // (the PAM file removed under the lock), not a reader miss. Settle for
+      // pacing, then re-check: an unconfigured result hides the icon and stops
+      // the retries rather than counting toward "reader unavailable".
       settleFingerprintAttempt()
+      refreshFingerprintStatus()
       return
     }
     // Bound the wait for the first prompt: a claim that never lands (a daemon
@@ -503,8 +508,16 @@ Item {
     stdout: StdioCollector { id: fingerprintCheckStdout; waitForEnd: true }
     onExited: {
       root.fingerprintConfigured = String(fingerprintCheckStdout.text || "").trim() === "yes"
-      if (root.lockRequested && root.fingerprintConfigured) root.startFingerprint()
-      else if (!root.fingerprintConfigured && fingerprintPam.active) fingerprintPam.abort()
+      if (root.lockRequested && root.fingerprintConfigured) {
+        // A pending retry already owns the next attempt.
+        if (!fingerprintRetryTimer.running) root.startFingerprint()
+      } else if (!root.fingerprintConfigured) {
+        // abort() delivers no completion signal, so close the attempt here
+        // too; settle returns before arming a retry while unconfigured.
+        if (fingerprintPam.active) fingerprintPam.abort()
+        root.settleFingerprintAttempt()
+        fingerprintRetryTimer.stop()
+      }
     }
   }
 
--- a/shell/plugins/lock/Service.qml
+++ b/shell/plugins/lock/Service.qml
@@ -492,9 +492,14 @@ Item {
     }
   }
 
+  // Configured means a print is actually enrolled: fprintd-list prints one
+  // " - #N: finger" line per print, while a user without any gets "has no
+  // fingers enrolled" or a ListEnrolledFingers failure, both of which a looser
+  // match would take as yes and then chase with a reader the account cannot
+  // use.
   Process {
     id: fingerprintCheckProc
-    command: ["bash", "-c", "if [[ -f /etc/pam.d/omarchy-lock-fingerprint ]] && command -v fprintd-list >/dev/null 2>&1 && fprintd-list \"$USER\" 2>/dev/null | grep -qi finger; then echo yes; else echo no; fi"]
+    command: ["bash", "-c", "if [[ -f /etc/pam.d/omarchy-lock-fingerprint ]] && command -v fprintd-list >/dev/null 2>&1 && fprintd-list \"$USER\" 2>/dev/null | grep -q ' - #'; then echo yes; else echo no; fi"]
     stdout: StdioCollector { id: fingerprintCheckStdout; waitForEnd: true }
     onExited: {
       root.fingerprintConfigured = String(fingerprintCheckStdout.text || "").trim() === "yes"
--- a/shell/plugins/lock/Service.qml
+++ b/shell/plugins/lock/Service.qml
@@ -4,6 +4,7 @@ import Quickshell.Io
 import Quickshell.Services.Pam
 import Quickshell.Wayland
 import qs.Commons
+import "FingerprintModel.js" as FingerprintModel
 
 Item {
   id: root
@@ -22,6 +23,9 @@ Item {
   property bool fingerprintAuthenticating: false
   property bool passwordPamConfigured: false
   property bool fingerprintConfigured: false
+  property int fingerprintUnreachedStreak: 0
+  property bool fingerprintAttemptReachedDevice: false
+  property double fingerprintLastNudgeMs: 0
   property bool previewVisible: false
   property string enteredPassword: ""
   property string pendingPassword: ""
@@ -46,6 +50,13 @@ Item {
   readonly property bool authenticating: authenticatingPassword || fingerprintAuthenticating
   readonly property var batteryService: shell && shell.services ? shell.firstPartyServiceFor("omarchy.battery") : null
   readonly property bool powerSaverActive: batteryService ? batteryService.powerSaverOnBattery : false
+  // The reader is unavailable once enough consecutive attempts fail to even
+  // reach it (see FingerprintModel). Distinct from a finger that simply did not
+  // match, which reaches the device and clears the streak. The in-flight reach
+  // clears the notice the moment a prompt arrives, a settle ahead of the streak
+  // reset -- so a recovered reader stops saying "unavailable" at once instead of
+  // waiting out the current attempt.
+  readonly property bool fingerprintUnavailable: fingerprintConfigured && !fingerprintAttemptReachedDevice && FingerprintModel.isUnavailable(fingerprintUnreachedStreak)
 
   function realScreenCount() {
     var screens = Quickshell.screens || []
@@ -130,7 +141,10 @@ Item {
     failedAttempts = 0
     authenticatingPassword = false
     fingerprintAuthenticating = false
+    fingerprintUnreachedStreak = 0
+    fingerprintLastNudgeMs = 0
     fingerprintRetryTimer.stop()
+    fingerprintReachTimer.stop()
     if (passwordPam.active) passwordPam.abort()
     if (fingerprintPam.active) fingerprintPam.abort()
   }
@@ -179,6 +193,23 @@ Item {
     root.monitorDpmsKnown = false
     if (!wakeProcess.running) wakeProcess.running = true
     if (lockRequested) armBlankTimer()
+    nudgeFingerprint()
+  }
+
+  // A keypress, touch, or cursor move is the user saying the reader is worth
+  // trying now, so collapse a backed-off wait to a prompt retry rather than
+  // ride out the cap. Rate-limited (see shouldNudge): a moving cursor raises a
+  // wake per motion event, and without the floor each fresh backoff wait would
+  // be re-collapsed straight back into the storm the backoff exists to prevent.
+  function nudgeFingerprint() {
+    if (!lockRequested || !fingerprintConfigured) return
+    if (fingerprintPam.active || fingerprintAuthenticating) return
+    if (!fingerprintRetryTimer.running) return
+    var now = Date.now()
+    if (!FingerprintModel.shouldNudge(now, fingerprintLastNudgeMs, fingerprintRetryTimer.interval)) return
+    fingerprintLastNudgeMs = now
+    fingerprintRetryTimer.interval = FingerprintModel.MATCH_RETRY_MS
+    fingerprintRetryTimer.restart()
   }
 
   function runBlank() {
@@ -249,20 +280,53 @@ Item {
     if (fingerprintPam.active || fingerprintAuthenticating) return
 
     fingerprintAuthenticating = true
+    fingerprintAttemptReachedDevice = false
     if (!fingerprintPam.start()) {
-      fingerprintAuthenticating = false
+      settleFingerprintAttempt()
+      return
     }
+    // Bound the wait for the first prompt: a claim that never lands (a daemon
+    // restarted under the verify, e.g. by the resume hook) otherwise hangs
+    // here without ever erroring, and nothing downstream re-arms.
+    fingerprintReachTimer.restart()
+  }
+
+  // A verify prompt is the only PAM message pam_fprintd relays, and only once
+  // the claim has landed, so any non-error message means this attempt reached
+  // the reader — the device works, whatever the verify then does. It may now
+  // wait for a finger as long as pam_fprintd allows, so the reach bound stops.
+  function noteFingerprintReachedDevice() {
+    fingerprintAttemptReachedDevice = true
+    fingerprintReachTimer.stop()
+  }
+
+  // An attempt that never reached the reader within the bound is stuck rather
+  // than waiting for a finger — abort it so it settles as unreached, which
+  // advances the streak (and so the notice) and retries against a daemon that
+  // may now be fresh, instead of hanging silently behind a normal icon.
+  function timeoutFingerprintReach() {
+    if (fingerprintPam.active) fingerprintPam.abort()
+    settleFingerprintAttempt()
+  }
+
+  // One PAM attempt can raise both onError and onCompleted, so fold each
+  // attempt into the streak exactly once: fingerprintAuthenticating is the
+  // attempt being open, and the first settle closes it. Reached attempts clear
+  // the streak; unreached ones advance it and stretch the next retry.
+  function settleFingerprintAttempt() {
+    if (!fingerprintAuthenticating) return
+    fingerprintAuthenticating = false
+    fingerprintReachTimer.stop()
+    if (!lockRequested || !fingerprintConfigured) return
+
+    fingerprintUnreachedStreak = FingerprintModel.nextStreak(fingerprintUnreachedStreak, fingerprintAttemptReachedDevice)
+    fingerprintRetryTimer.interval = FingerprintModel.retryDelayMs(fingerprintUnreachedStreak)
+    fingerprintRetryTimer.restart()
   }
 
   function handleFingerprintFinished(result) {
-    fingerprintAuthenticating = false
-
-    if (!lockRequested) return
-    if (result === PamResult.Success) {
-      finishUnlock()
-    } else if (fingerprintConfigured) {
-      fingerprintRetryTimer.restart()
-    }
+    if (result === PamResult.Success && lockRequested) finishUnlock()
+    else settleFingerprintAttempt()
   }
 
   WlSessionLock {
@@ -309,6 +373,7 @@ Item {
         backgroundPath: root.backgroundPath
         backgroundVersion: root.backgroundVersion
         fingerprintConfigured: root.fingerprintConfigured
+        fingerprintUnavailable: root.fingerprintUnavailable
         authenticatingPassword: root.authenticatingPassword
         failureMessage: root.failureMessage
         failedAttempts: root.failedAttempts
@@ -341,6 +406,7 @@ Item {
       backgroundPath: root.backgroundPath
       backgroundVersion: root.backgroundVersion
       fingerprintConfigured: root.fingerprintConfigured
+      fingerprintUnavailable: root.fingerprintUnavailable
       authenticatingPassword: false
       failureMessage: ""
       failedAttempts: 0
@@ -384,21 +450,31 @@ Item {
     config: "omarchy-lock-fingerprint"
     user: root.userName
 
+    onPamMessage: {
+      if (!messageIsError) root.noteFingerprintReachedDevice()
+    }
+
     onCompleted: function(result) {
       root.handleFingerprintFinished(result)
     }
 
     onError: function(error) {
-      root.fingerprintAuthenticating = false
-      if (root.lockRequested && root.fingerprintConfigured) fingerprintRetryTimer.restart()
+      root.settleFingerprintAttempt()
     }
   }
 
   Timer {
     id: fingerprintRetryTimer
-    interval: 250
+    interval: FingerprintModel.MATCH_RETRY_MS
     repeat: false
     onTriggered: root.startFingerprint()
+  }
+
+  Timer {
+    id: fingerprintReachTimer
+    interval: FingerprintModel.REACH_TIMEOUT_MS
+    repeat: false
+    onTriggered: root.timeoutFingerprintReach()
   }
 
   Process {
@@ -600,6 +676,7 @@ Item {
         realScreens: root.realScreenCount(),
         passwordPam: root.passwordPamConfigured,
         fingerprint: root.fingerprintConfigured,
+        fingerprintUnavailable: root.fingerprintUnavailable,
         authenticating: root.authenticating,
         lastEvent: root.lastEvent,
         lastEventAt: root.lastEventAt
--- a/bin/omarchy-apply-lock
+++ b/bin/omarchy-apply-lock
@@ -46,6 +46,8 @@ EOF
 # here rather than leaving the pair to drift apart between setups and updates.
 resume_hook_src="$OMARCHY_PATH/default/systemd/system-sleep/fprintd-resume"
 resume_hook_dst=/usr/lib/systemd/system-sleep/fprintd-resume
+stop_timeout_src="$OMARCHY_PATH/default/systemd/system/fprintd.service.d/10-stop-timeout.conf"
+stop_timeout_dst=/etc/systemd/system/fprintd.service.d/10-stop-timeout.conf
 
 # One " - #N: finger" line per enrolled print; a user without any gets "has no
 # fingers enrolled" instead, which must not count.
@@ -58,8 +60,17 @@ auth       required                    pam_fprintd.so
 account    include                     system-local-login
 EOF
   as_root install -Dm755 -o root -g root "$resume_hook_src" "$resume_hook_dst"
+  as_root install -Dm644 -o root -g root "$stop_timeout_src" "$stop_timeout_dst"
+  as_root systemctl daemon-reload
 else
   as_root rm -f /etc/pam.d/omarchy-lock-fingerprint "$resume_hook_dst"
+  # Only reload when something was there: this branch also runs in the
+  # install chroot, where there is no systemd to reload.
+  if [[ -e $stop_timeout_dst ]]; then
+    as_root rm -f "$stop_timeout_dst"
+    as_root rmdir --ignore-fail-on-non-empty "$(dirname "$stop_timeout_dst")"
+    as_root systemctl daemon-reload
+  fi
 fi
 
 # omarchy-shell can't reach a running shell during chroot install. The echo
--- a/bin/omarchy-apply-lock
+++ b/bin/omarchy-apply-lock
@@ -47,8 +47,10 @@ EOF
 resume_hook_src="$OMARCHY_PATH/default/systemd/system-sleep/fprintd-resume"
 resume_hook_dst=/usr/lib/systemd/system-sleep/fprintd-resume
 
+# One " - #N: finger" line per enrolled print; a user without any gets "has no
+# fingers enrolled" instead, which must not count.
 if [[ -x /usr/bin/fprintd-list ]] &&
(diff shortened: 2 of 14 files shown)


paracycle:
Same signature here on a Framework 13 (AMD Ryzen AI 300), Omarchy 4.0.0-1, found while chasing an unrelated resume bug: 45 `omarchy-lock-fingerprint` PAM subprocesses in the first 65 s after resume, every one failing with code 9 (`PAM_AUTHINFO_UNAVAIL`), stopping only when the password was typed. The 250 ms timer in `Service.qml` is exactly the multiplier, so the backoff here is the right lever.

One data point that may be worth folding in: dropping a `systemctl try-restart fprintd.service` into a `post` sleep hook took the same machine from 45 spawns to 2 on the next resume. Clearing the wedged claim at resume and backing off on device errors are complementary — the backoff stops the storm, the restart makes fingerprint actually usable again on the first touch instead of after fprintd's 30 s idle exit. #7229 asks for the restart specifically.

GeertJohan:
#7158 has been reworked. It is now three layers: a `post` system-sleep hook that restarts fprintd on resume, exponential backoff on the retry path, and a "reader unavailable" notice — replacing the pre-suspend announce/park and the resume relay.

On the three findings:

- **(3) undeliverable resume relay** — removed. Resume is now a `system-sleep` hook, the guaranteed post-resume signal; no dependence on the user-service monitor's restart timing or on D-Bus replaying `PrepareForSleep(false)`.
- **(2) SIGKILL-not-release** — removed. There is no pre-suspend abort; the wedged claim is cleared by restarting fprintd on resume (a fresh daemon holds no claim), which is deterministic rather than an early kill.
- **(1) recovery signal** — changed. The streak now counts attempts that fail to *reach* the reader (no finger prompt relayed), not verify outcomes. A verify that errors after a prompt counts as reached and clears the streak, so it no longer inverts or pins at one step, and a healthy-but-untouched reader never trips the notice. The `Date.now()`-across-suspend misclassification does not apply — there is no elapsed-time heuristic here.

This does touch `Service.qml` again, so it overlaps #7179 on the retry path — but on a different signal: claim-reachability (wedged/held/absent) here versus verify duty-cycle (thermal) there. Behaviorally this PR leaves the thermal verify-error case to #7179; the merge of the two retry paths is the maintainer's call.

The repeated-`prepareForSleep` budget reset is also gone — there is no such IPC anymore. #7645 remains an upstream Quickshell bug, untouched here.

GeertJohan:
@omarchybot please re-review

GeertJohan:
The PR description is updated to reflect the new implmentation.

GeertJohan:
Findings 1, 3 and 4 are addressed; 2 and 5 are description fixes.

- **(1) resume holding the thaw:** kept your `07c9558d` `timeout` bound as pushed.
- **(3) notice lingering after recovery:** `fingerprintUnavailable` is now gated on the in-flight reach, so a relayed prompt clears it at once rather than a settle later. No flicker — settling then zeroes the streak, so it stays cleared.
- **(4) no user-forced retry:** a keypress, touch or cursor move now collapses a backed-off wait to a prompt retry — but only a wait longer than the fast interval, so once collapsed further input leaves it alone and typing a password cannot respin the loop.

Description changes for the rest:

- **(2)** the scope note now states the "reached" signal is wider than thermal: any failure arriving after the claim lands — a mid-verify refusal, a disconnect, a driver error — reads as reached and gets neither the backoff nor the notice here.
- **(5)** the stale-claim note now also covers a live verify caught across the suspend (a `sudo` prompt on a TTY) being dropped to a password, and qualifies "a healthy resume pays nothing" to while fprintd is not running.

On the pre-existing `fingerprintConfigured` latch (a `fprintd-list` check failing across the restart turning fingerprint off for the lock session): real and pre-existing; left out of this PR since it predates it and warrants its own change, but noted.

#7179 still overlaps the retry path in `Service.qml` (the merge-tree conflict stands); the signals remain complementary and how they share the path is the maintainer's call. #7645 untouched.

GeertJohan:
@omarchybot I think #7179 or this PR can be rebased on top of the other, whichever lands first. Best to get either one merged as quick as possible (once all findings are addressed/resolved) such that the other PR can rebase.

GeertJohan:
Both addressed at `ba69b0a3`.

- **(1) nudge respinning the loop:** added a wall-clock cooldown. `nudgeFingerprint` now gates on `FingerprintModel.shouldNudge(now, lastNudge, interval)` — a wait longer than the fast interval, and at most one collapse per `NUDGE_COOLDOWN_MS` (2s). Driving the extracted `shouldNudge` at 125 Hz over 10s gives 6 collapses, not 39. (Kept the `restart()` — correct whether or not assigning `interval` restarts, so I left it unambiguous rather than relying on the implicit reset.)
- **(2) orphaned attempt:** added a reach-timeout. `startFingerprint` arms `fingerprintReachTimer` (`REACH_TIMEOUT_MS`, 5s); a relayed prompt stops it, so a healthy reader waiting for a finger is untouched — it prompts in under a second. An attempt that gets no prompt in the window is stuck (a claim orphaned when the hook restarts fprintd under it), so `timeoutFingerprintReach` aborts and settles it unreached: the streak advances, the notice fires, and the retry hits the fresh daemon — bounding #8747's ~25–30s to ~5s and making it visible instead of silent. It settles directly rather than through `onError`, so it doesn't depend on whether `abort()` raises one.

That last point also bears on #8531's flagged interaction: our abort→settle path is guarded and idempotent, so a stray `onError` from an abort folds in once and cannot advance the streak twice. Whether `PamContext.abort()` raises `onError` I did not confirm either.

Coverage: `shouldNudge` and the new constants are in `FingerprintModel.js` with Node assertions, including the moving-cursor case that would have caught (1). The reach-timeout wiring itself is QML-level and not extract-tested.

The `Service.qml` overlap with #7179/#8531 stands, but it doesn't need resolving up front — whichever of the three lands first, the others rebase onto it.

GeertJohan:
Rebased onto current `quattro` and curated to three commits.

- Rebase picked up the new `textFormat` policy (`qml-text-format-test.sh`): the fingerprint icon and the unavailable notice now declare `textFormat: Text.PlainText`, and that test passes.
- History collapsed from the review rounds into three atomic commits — the resume hook, the hook `timeout` bound (kept as its own commit, still authored by @Omabot), and the pacing + unavailable notice. No behavior change from the last push; the tree is byte-identical.

Green here: `lock-fingerprint-retry`, `lock-fingerprint-indicator`, `fprintd-resume-hook`, `fprintd-resume-migration`, `qml-text-format`, and qmllint.

GeertJohan:
I let a separate Claude Fable 5 do a review at 'max' setting and it found more issues. Working through them now..

GeertJohan:
All fixes pushed, ready for re-review and merge

MaxMad75:
Tested layer 1 on hardware that is not yet represented here, and it works. One finding that I think is worth a small addition.

**Machine:** ThinkPad X390 Yoga (`20NQS4RE00`), Synaptics `06cb:00bd`, fprintd 1.94.5-2, libfprint 1.94.100-1, Linux 7.1.9-arch1-2, Omarchy 4.0.1, systemd 261 — and `/sys/power/mem_sleep` is `s2idle [deep]`, i.e. **S3, not s2idle**. #8818's reporters and mine on #8935 are all s2idle, so this is the first deep-sleep data point. I disabled my own `fprintd-resume.service` and removed my `TimeoutStopSec` drop-in first, so this is your hook alone, installed verbatim from the branch.

**The hook does what it says, and the ordering is right:**

```
20:16:21.593  systemd-sleep: System returned from sleep operation 'suspend'
20:16:21.602  Stopping Fingerprint Authentication Daemon…      ← hook, 9 ms after resume
20:16:21.694  fprintd.service: Deactivated successfully
20:16:22.057  Started Fingerprint Authentication Daemon
20:16:22.090  Started Lock Omarchy before suspend               ← lock work starts after
20:16:22.094  Finished System Suspend
```

PID 266447 → 268213, and `fprintd-list` answers on the first try afterwards. Total cost to the thaw: **~500 ms**. Your "the restart is real, not a no-op" holds here — the daemon was running and claimed at resume.

**I also exercised `FingerprintModel.js` directly under node**, against the storm I actually measured on this machine (178 PAM sessions and 88 `already claimed` in 26 seconds, ending in a black screen):

| | attempts in 26 s |
|---|---|
| flat 250 ms | 104 (I measured 178, so the real loop is faster than the nominal timer) |
| your backoff | **5** |

- notice appears after 3 unreached attempts, 7.0 s in;
- largest gap 40 s, so it does clear fprintd's 30 s idle exit even when no hook runs — the ceiling is sized correctly;
- a reached attempt drops straight back to 250 ms;
- `shouldNudge` with a backwards clock returns true, and a 125 Hz cursor over 10 s collapses the wait twice, not 39 times.

## The one improvement I would make

The hook comment says a wedged fprintd "is idle and exits on SIGTERM in milliseconds". That is true for **the wedge this PR targets** — I held the reader with `fprintd-verify` and restarted:

```console
20:14:01.452  Stopping…
20:14:01.454  Deactivated successfully      ← 2 ms
20:14:01.791  Started                        ← 340 ms total
```

But it is not true for **the other wedge**, the stale device handle after a USB re-enumeration (#8935; @xe-nvdk saw the same on `06cb:00fc`, @k7cfo on `06cb:0123`). That instance does not answer SIGTERM. Two of the ten fprintd stops in this machine's journal are it:

```
Aug 28 17:16:19.233  Stopping Fingerprint Authentication Daemon...
Aug 28 17:16:24.308  fprintd.service: State 'stop-sigterm' timed out. Killing.
Aug 28 17:16:24.308  Killing process 24160 (fprintd) with signal SIGKILL
Aug 28 17:16:24.314  Stopped   ← 5.08 s
                     Consumed 1.183s CPU over 3h 57min wall clock
```

(the second, 10:10:15 today, is 5.02 s — Omarchy's `DefaultTimeoutStopSec=5s` in `etc/systemd/system.conf.d/10-faster-shutdown.conf`.)

Since your hook runs **synchronously while `user.slice` is frozen**, that 5 s lands on the thaw — and it lands there precisely when the daemon is wedged, which is the case the hook exists for. `timeout 10s` does not bound it: it is above the 5 s, and as your own comment notes, killing the client leaves the job enqueued anyway. So the resume goes from ~0.5 s to ~5.6 s in the failure case.

One drop-in fixes it, and it is independent of which mechanism ends up upstream:

```ini
# /etc/systemd/system/fprintd.service.d/stop-timeout.conf
[Service]
TimeoutStopSec=3s
```

Verified here (`systemctl show fprintd -p TimeoutStopUSec --value` → `3s`). I put it on #8935 for the same reason; it belongs with whichever resume mechanism lands, and right now none of the three PRs sets it. Happy to open it as a PR against your branch if you would rather have it here — or just take the four lines.

**What I did not test:** the QML layers on a live lock screen. I stopped short of patching the running lock plugin on my own machine; the numbers above are the extracted model, not the wired-up loop.

**On the three-way overlap:** my #8935 does layer 1 as a `After=suspend.target` unit rather than a sleep hook. It is outside the parallel set @spencerbull measured in #8818, but it acts on the same unit at the same moment, so it should not ship alongside either hook. Having tested yours, I would rather this one land — it covers the wedge *and* the storm afterwards, and the storm is what actually put a black screen in front of me.

MaxMad75:
Follow-up to my testing note above: I have put the `TimeoutStopSec` cap up as a PR against your branch rather than opening a fifth PR here — GeertJohan/omarchy#1. It only has value once a resume mechanism lands, so it belongs with yours rather than standing on its own.

It carries the cap through setup, `omarchy-apply-lock`, the migration and removal alongside the hook, and takes `fprintd-resume-migration-test.sh` from 4 to 7 assertions (all four mutations I tried are caught). I also closed my own #8935 in favour of this PR.

Take it, rework it, or tell me to drop it.

GeertJohan:
Thanks for checking this out @MaxMad75, I was just also looking at your PR. I'm close to completing this PR and I think we can get it merged soon. That will fix the same issue #8935 is fixing, and at the same time also addresses other related issues. I did consider using a systemd hook like you do in #8935, but decided not to because Omarchy already uses hooks for similar things (force-igpu, keyboard-backlight, unmount-fuse). So I preferred aligning with those. However I am now making a change to address the timing/delay of the hook; instead of doing `timeout 10s systemctl ..` I'm considering to change it to `systemctl .. --no-block`. That will make the hook very fast, making the fprintd restart async (barely noticable to the user).

Meanwhile I had an agent running that created a fix for the cap, and while writing this I see a PR coming in from you addressing the same thing, so asking my agent to compare the two and it found good and bad things in both; mixing them picking best of both solutions and will have you added as Co-author on that commit, thanks!

MaxMad75:
Thanks — and agreed on the hook over a unit; aligning with `force-igpu` and the others is the better argument, which is part of why I closed #8935.

One thing on `--no-block`, before you make that change: **it makes the cap matter more, not less.**

Your current comment gives the reason the restart is synchronous: *"user sessions stay frozen until the hook returns, so the lock screen's first attempt after thaw lands on the fresh daemon rather than the wedged one."* `--no-block` gives that guarantee up. The hook returns at once, sessions thaw, and the restart proceeds concurrently — so the lock screen's first attempts now race the stop.

That matters because of what the stop costs when it is not free. A wedged fprintd does not answer SIGTERM, and while the unit sits in `deactivating` a D-Bus activation cannot bring up a replacement. Measured on my machine:

| daemon state | stop | full restart |
|---|---|---|
| healthy, holding a claim | 2 ms | 340 ms |
| wedged (stale device handle) | **5.02 s / 5.08 s**, SIGKILL | ~5.6 s |

So the two options are not symmetric:

- **synchronous, uncapped** — thaw held ~5.6 s in the failure case, but the guarantee holds.
- **`--no-block`, uncapped** — thaw instant, and the lock screen faces a dying daemon for up to ~5.6 s. That window is exactly the retry storm your layers 2 and 3 exist to survive; it moves the cost from the thaw into the user's face instead of removing it.
- **`--no-block` + cap** — thaw instant, window bounded to ~3.3 s.
- **synchronous + cap** — thaw held 0.09 s when healthy (measured on a full cycle here: hook fires 9 ms after resume, fresh daemon 91 ms later), ~3.3 s worst case, guarantee intact.

I would take the last one, since the healthy path is the common one and it costs 91 ms — but either way the cap is what bounds the tail. Without it, `--no-block` does not make the delay go away, it just relocates it to where the user can see it.

Happy either way on the PR — merge, rework, or cherry-pick whatever survives your agent's comparison. Co-authorship is appreciated but not necessary; the numbers are the useful part.

Two small notes from testing your branch verbatim on a third reader (Synaptics `06cb:00bd`, ThinkPad X390 Yoga, `mem_sleep=deep` — S3, where the other reporters are s2idle):

- Full cycle with the hook: resume at `20:40:48.394`, `Stopping` at `.403`, `Started` at `.485`, `Finished System Suspend` at `.499`. Then a clean idle exit 30 s later, so the hook does not interfere with the idle-exit fallback your backoff ceiling is sized against.
- I have your layers 2 and 3 running on this machine now as well, patched onto 4.0.1 via a cloned lock plugin. I will report if anything shows up in real use.

GeertJohan:
> One thing on --no-block, before you make that change: it makes the cap matter more, not less.

I agree. Could you give the latest version of this branch a try? It addresses that

GeertJohan:
I decided to use `--no-block` because I think responsiveness to the user is very important. I'd rather show them a lock screen fast without a working fingerprint scanner (yet..), than no lockscreen at all (until timeout has passed). So that's why I went with async, even though on my own hardware restarting fprintd only takes a couple of ms.

MaxMad75:
Ran the async variant on the S3 machine. **You were right and I was wrong** — and the trace matches your hook comment sentence for sentence.

First, the correction I owe you: I was reasoning about the synchronous restart as journal time, not as *frozen screen* time. `user.slice` staying frozen until the `post` hook returns makes a 3 s cap a 3 s black, unresponsive desktop, in exactly the case the hook exists for. That is plainly worse than letting the lock loop absorb it. Thanks for spelling it out.

**Thaw is off the critical path now.** ThinkPad X390 Yoga, Synaptics `06cb:00bd`, `mem_sleep=deep`, systemd 261, kernel 7.1.9-arch1-2, branch at `c1a9704e`:

```
20:52:57.367  System returned from sleep operation 'suspend'
20:52:57.375  Stopping Fingerprint Authentication Daemon…     ← job enqueued, 8 ms
20:52:57.379  fprintd.service: Deactivated successfully
20:52:57.410  Finished System Suspend                          ← thaw at 43 ms
20:52:57.413  Reached target Suspend
20:52:57.425  Starting Fingerprint Authentication Daemon…      ← start lands AFTER the thaw
20:52:57.524  Started Fingerprint Authentication Daemon        ← fresh daemon, 157 ms after resume
```

The ordering is the proof: `Starting` sits *below* `Finished System Suspend`, so the restart genuinely moved off the thaw. 43 ms held versus 91–105 ms synchronous on the same machine, and — the part that matters — no 3 s worst case.

**And the lock loop absorbed it exactly as your comment predicts**, on a real lock screen, with the daemon mid-restart:

```
20:52:57.445  Starting pam session … "omarchy-lock-fingerprint"     ← attempt 1, 20 ms before the daemon is up
20:52:57.524  Error while authenticating: … (code 9)
20:52:57.525  omarchy lock fingerprint-unreached: streak=1          ← layer 2 sees it
20:52:58.546  Starting pam session …                                 ← attempt 2, one tier later
20:52:58.617  Relaying pam message: "Place your right index finger…" ← fresh daemon, reader reached
20:53:28.624  "Verification timed out"  (I left it untouched)
20:53:28.632  omarchy lock fingerprint-recovered: streak=1
20:53:28.912  Starting pam session …
20:53:28.957  "Place your right index finger on the fingerprint reader"
20:53:51.203  Authenticated successfully.
20:53:51.214  omarchy lock unlocked
```

*"That attempt fails and the next, a second later, claims the fresh daemon"* — 20:52:57.445 fails, 20:52:58.546 prompts. One unreached attempt, streak 1, never near the `UNAVAILABLE_AFTER` threshold, and the user-visible result is a reader that simply works on the first touch. The 30 s gap is me deliberately not touching the sensor to watch the timeout path settle and clear the streak; the unlock at the end is a real finger.

Layers 2 and 3 are patched onto 4.0.1 here through a cloned lock plugin (your branch sits on a newer `quattro`, so I applied the diff rather than the files). `shouldNudge`'s new settle-based measurement is in and loads clean.

Nothing further from me — this is ready as far as I can test it. Thanks for taking the cap and the `rmdir`, and for the co-author line.

GeertJohan:
Hardware run of `c1a9704e` on my own machine, with the branch's three lock-plugin files dropped into the packaged 4.0.1 tree and the hook + `TimeoutStopSec` drop-in installed by the branch's migration (`omarchy-migrate` path, so the real install route). Goodix MOC on an nvidia laptop, s2idle, `SYSTEMD_SLEEP_FREEZE_USER_SESSIONS=0` — so this covers the wedge, the pacing and the notice, but **not** the thaw cost of the restart, which needs a machine that freezes sessions (@MaxMad75's X390 is the reference for that).

**1. Locked suspend/resume** (lock → lid close → ~6 s → open → touch, no mouse). Journal, both sides:

| t | event |
|---|---|
| 0.00 | lock requested; fprintd activated; verify prompts |
| +1.76 | `PrepareForSleep` → fprintd suspends the reader → `Cannot run while suspended` |
| +2.03 | 250 ms retry hits the suspended device → `fingerprint-unreached: streak=1` |
| +2.17 → +7.84 | suspend |
| +7.83 | `fingerprint-resume: streak=1` — sleep watch noticed the wall-clock gap on its first tick, streak reset |
| +7.850 | hook fires **6 ms after** `System returned from sleep`; fprintd logs `User destroyed open device! Not cleaning up properly!` (the wedged claim being torn down); stopped in 2 ms |
| +7.853 | `Finished System Suspend` — the whole post phase took 9 ms, i.e. `--no-block` returned at once |
| +7.856 → +8.08 | fresh fprintd up (228 ms) |
| +8.08 | post-resume attempt lands on the fresh daemon and prompts — no miss after resume |
| +9.21 | `unlocked` by finger, **1.36 s after wake** |

Three PAM sessions for the whole lock (initial, one pre-freeze miss, the one that unlocked). Pre-branch this machine hit the wedge at 7 of 7 locked suspends.

**2. Held reader** (`fprintd-verify` holding the claim in the background, lock via IPC, release after ~8 s, no input at all):

| t | event |
|---|---|
| +0.01 | `streak=1` ("already claimed", 13 ms) |
| +1.03 | `streak=2` |
| +3.05 | **`fingerprint-unavailable: streak=3`** — crossed icon + notice |
| +7.07 | `streak=4`, still held |
| ≈ +8.5 | reader released |
| +15.1 | next attempt (8 s tier) prompts; `lock status` flipped `fingerprintUnavailable` to `false` the moment the prompt landed, ahead of the settle |
| +30.4 | `unlocked` by finger (all five PAM sessions were fingerprint ones) |

Five PAM sessions in 30 s against ~120 at the old flat 250 ms; the notice appeared and cleared with no user input, and the 6.5 s between release and the next attempt is the tier pacing with nobody present — a keypress would have collapsed it via the nudge.

Two follow-ups came out of this and are queued for the next push:

- **Resume grace window.** With the restart off the thaw, the loop's first attempts after a wake can land on the old daemon while it is being stopped (up to ~3 s when it ignores SIGTERM), and three of those would have shown the notice for a reader that was simply being restarted underneath; the same race already existed on machines that don't freeze sessions. A resume is now noticed by any of three signals (the sleep watch, now running for the whole lock; a retry that fired late on the wall clock; an unreached attempt that outlived the reach bound on the wall clock) and opens a 5 s window in which misses hold the streak at the first tier and never count toward the notice. Model + Node test + a Service.qml fixture; not yet on hardware.
- **`fingerprint-recovered` on unlock.** Run 2 showed that a match which unlocks after a run of misses logged only `unlocked`; the recovery line is now emitted there too.

GeertJohan:
@omarchybot ready for a final review and merge please :)

MaxMad75:
Not blocking the merge — just claiming the one test you can't run.

**Offer:** push the resume grace window whenever it suits (before or after this lands), and I will run it on the X390. That machine freezes sessions by default, has the `06cb:00bd` stale-handle wedge that ignores SIGTERM, and is where the 3 s stop is real rather than 2 ms. I'd report the same shape as above: streak per attempt, whether `fingerprint-unavailable` fires, and the wall-clock from resume to prompt.

**And the arithmetic says you are right to add it.** On a freezing machine with a wedged daemon, the notice fires spuriously today. Stop begins ~8 ms after resume and runs the full 3 s cap; the fresh daemon is up around 3.2 s. The loop, against `retryDelayMs`:

| attempt | t | daemon | result |
|---|---|---|---|
| 1 | ~0.08 s | stopping | unreached, streak 1 |
| 2 | ~1.08 s | stopping | unreached, streak 2 |
| 3 | ~3.08 s | still stopping (up at ~3.2 s) | unreached, **streak 3 → notice** |

So `UNAVAILABLE_AFTER` is crossed by roughly 120 ms, and the user is told the reader is unavailable while it is merely being restarted underneath. My run yesterday only showed streak 1 because that daemon was healthy and died in 2 ms — the wedged case is the one that trips it, and it is the case the hook exists for.

A 5 s window covers it with room: worst case here is the cap plus the ~230 ms start.

One thing worth pinning in the test when you push it: the window has to be armed by the *resume*, not by the first miss — otherwise a wedged daemon that was already being stopped before the loop's first attempt gets no window at all. Your three-signal detection reads like it does that; I mention it only because it is the part a fixture can pass while hardware still fails.

Ping me on the push.

osborng:
A complementary angle on the lid-close trigger, in case it's useful: that
particular "losing roll" can be made deterministic rather than probabilistic,
because Omarchy already ships the mechanism and just doesn't apply it to the
lock stack.

`omarchy-setup-security-fingerprint` builds a clamshell gate:

```
auth      [success=1 default=ignore] pam_exec.so quiet /usr/bin/omarchy-hw-laptop-closed
```

and the comment above it says it "runs before pam_fprintd in **every stack**".
It is inserted immediately before `pam_fprintd` in `/etc/pam.d/sudo` and
`/etc/pam.d/polkit-1`. But `setup_lock_fingerprint_pam()` writes the lock stack
from a quoted heredoc that never references it:

```
#%PAM-1.0
auth       required                    pam_fprintd.so
```

So the one stack that runs *while the lid is closing* — `omarchy-sleep-lock.service`
locks before suspend — is the only one that still hands the reader to
`pam_fprintd` at exactly that moment. That is the verify this PR then has to
clean up after.

This is not a substitute for the PR. The gate does nothing for an idle suspend
with the lid open, and nothing for a claim held by another client, so the resume
hook and the backoff are still needed. It just removes the lid-close case at
source, which on a laptop seems to be the common one.

One wrinkle if anyone tries it: the gate cannot be dropped in verbatim. In
sudo/polkit `pam_fprintd` is `sufficient` with a password fallback after it, so
`success=1` skips *into* something. In the lock stack `pam_fprintd` is
`required` and is the only auth line, so `success=1` would skip past the end of
the auth stack. It needs `pam_fprintd` as `sufficient` plus a `pam_deny`
terminator.

The two also compose in the other direction: a gate on its own spins against the
flat 250 ms retry loop if the machine ever runs docked-and-awake with the lid
shut. Layer 2 here removes that objection.

Seen on Framework 13 (Intel Core Ultra X7 358H), Goodix 27c6:609c, Omarchy
4.0.1-1, fprintd 1.94.5-2, libfprint 1.94.100-1.

osborng:
Correcting my comment above: I claimed the clamshell gate "removes the lid-close
case at source, which on a laptop seems to be the common one". Having now run
with the gate in the lock stack and hit the wedge anyway, that overstates it —
the gate covers a narrower case than I said, and does not remove the need for
layer 1 here.

The run that failed, with the gate present and working:

```
17:53:13  omarchy lock  lock-requested / secure=true      <- idle lock, lid OPEN
17:53:13  fprintd.service started, reader claimed
18:10:10  systemd-logind: Lid closed. / Suspending...
18:10:10  fprintd: Device reported an error during verify: Cannot run while suspended.
20:10:40  systemd-logind: Lid opened.  (resume)
20:10:59  pam_exec(omarchy-lock-fingerprint:auth): omarchy-hw-laptop-closed failed: exit code 1
20:11:00  omarchy lock  unlocked        <- password; fingerprint never came back
20:11:30  fprintd.service: Deactivated successfully   <- idle exit finally drops the claim
```

The verify that got frozen started at 17:53 on an **idle** lock with the lid
open — seventeen minutes before the lid moved. The gate correctly permitted it;
an open lid is precisely when fingerprint should run. The lid close then froze a
claim that was already in flight.

So the gate only helps when a lid close is what *triggers* the lock
(`omarchy-sleep-lock` on an unlocked machine). When the machine is already
locked and retrying — arguably the more common way a laptop reaches suspend —
the claim is taken long before the lid moves, and a PAM-time predicate has
nothing left to prevent. Blocking new sessions during the suspend window doesn't
help either: the freeze lands within the same second.

That makes the gate a narrowing of the trigger surface, not a fix, and it leaves
this PR's layer 1 doing the load-bearing work. I don't think it changes anything
about the PR; I just didn't want a wrong "you could prevent this instead"
sitting under it.

(Incidentally the `exit code 1` line at 20:10:59 is the gate behaving correctly
and still logging at LOG_ERR — #7871's case, from the lock stack this time.)

GeertJohan:
@MaxMad75 Yes, please run it. The grace window has been on the head since yesterday evening (`434fd2fa`, with the recovered-on-unlock line as `04894831`), and your "armed by the resume, not by the first miss" point found a real hole: an attempt in flight across the suspend was only recognised as such if its wall-clock age exceeded the reach bound, so after a suspend shorter than ~22 s the restart ended it and it settled as an ordinary miss before the sleep watch's next tick reset the streak. With two pre-freeze misses already on the streak that is a third strike and a notice flash. `a3dfe10c` judges the sleep from the watch's last tick instead (never more than a second stale when the freeze lands), so the settle arms the window itself. A fixture step with exactly your scenario fails on `04894831` and passes on `a3dfe10c`.

Head to run: `a3dfe10c`. Useful from the X390: streak per attempt after the wake, whether `fingerprint-unavailable` fires at all, and resume-to-prompt wall-clock — the `omarchy lock fingerprint-*` lines in `journalctl --user` carry all three. One run with a short suspend (lid closed for ~10 s) would cover the case the hardening is for. Thanks.

GeertJohan:
@osborng Thanks for the correction and the journal. It is a useful datapoint for the PR as it stands: without the hook the reader stayed dead ~50 s after resume until fprintd's idle exit dropped the claim, which is exactly what layer 1 removes. Agreed on the conclusion: the gate would only cover a lid close that triggers the lock, and the resume restart already handles that case, so I am not adding `pam_exec` to the lock stack for it. The `failed: exit code 1` line at LOG_ERR is #7871's. The lock stack stays as it is.

MaxMad75:
Ran `a3dfe10c` on the X390 Yoga. Six suspend/resume cycles, all real lid closes rather
than `rtcwake` (which writes straight to `/sys/power/state` and skips the hooks entirely).
Synaptics `06cb:00bd`, `mem_sleep=deep`, session freezing on, `TimeoutStopSec=3s` drop-in
in place, hook from this branch installed.

**Short answer: the notice never fired, and the streak held across the restart every time.**

| # | sleep | streak at freeze | fprintd stop | fresh daemon | first prompt | at resume | unlocked |
|---|---|---|---|---|---|---|---|
| 1 | 24.3 s | — | 8 ms | +558 ms | +899 ms | `streak=0` | +3.754 s |
| 2 | 14.8 s | — | 6 ms | +553 ms | — | `streak=0` | +4.700 s |
| 3 | 12.0 s | `streak=1` | 6 ms | +589 ms | +642 ms | `streak=1` | +2.318 s |
| 4 | 12.0 s | `streak=1` | 30 ms | +567 ms | +623 ms | `streak=1` | +6.609 s |
| 5 | 9.7 s | — | 7 ms | +578 ms | +926 ms | `streak=0` | +3.599 s |
| 6 | 15.9 s | `streak=2` | 10 ms | +357 ms | none | `streak=2` | +8.198 s (password) |

Offsets are against `PM: suspend exit`. Runs 1–5 all unlocked by finger.

**Run 6 is the one worth reading.** I could not get the daemon to wedge on its own — five
consecutive lid closes gave five clean stops of 6–30 ms, never the 3 s cap — so I forced it
with `SIGSTOP` on fprintd's MainPID before locking. That does not reproduce a wedge, and the
reason is worth writing down so nobody else spends an evening on it: systemd sends SIGCONT
alongside SIGTERM precisely so stopped processes can still be killed (`man systemd.kill`), so
the stop still completed in 10 ms. **It did produce something better, though:** with the
daemon frozen, both pre-freeze attempts ran into `REACH_TIMEOUT_MS` and logged
`fingerprint-reach-timeout` + `fingerprint-unreached`, putting the streak at **2** going into
the suspend — the deepest starting point of any run here. After the wake, `fingerprint-resume:
streak=2` came at **+15 ms**, so the resume detection fires and the streak is held rather than
advanced. That is the mechanism doing its job.

**But in run 6 the reader was genuinely dead afterwards and `fingerprint-unavailable` never
fired.** At +2 ms: `pam_fprintd: GetDevices failed: Remote peer disconnected`. After that,
nothing — no further attempt, no prompt, no notice — and I unlocked with the password at
+8.2 s. Querying the shell afterwards showed `"fingerprint": false` with
`"fingerprintUnavailable": false`, while `fprintd-list` found the device and the enrolled
finger perfectly well. The state persisted until I restarted the shell.

That is the `fingerprintCheckProc` gap already noted as pre-existing and open — a single
transient "no" clears `fingerprintConfigured`, hides the icon and stops the retry timer with
no re-check. I mention it because this is it happening on hardware, in the exact window this
PR is about: the resume restart is what disconnects the peer, so the restart that fixes the
reader is also what can convince the lock screen the reader does not exist. The streak logic
is not what fails here — it never gets another attempt to count. It does not block this PR,
but it does mean layer 3 has a blind spot that layer 1 can walk into.

**A correction to my own arithmetic from earlier in this thread.** I wrote that on a freezing
machine with a wedged daemon the notice fires spuriously *today*. The premise still holds —
the 5.02 s and 5.08 s stops are in my journal from the 29th, and the loop maths against
`retryDelayMs` is unchanged — but six lid closes produced zero wedges. So the spurious notice
is a real but occasional case, not the normal path, and on this machine the grace window's
day-to-day effect is closer to "nothing visible" than to "fixes something users hit". That is
an argument for it as cheap insurance, not against it.

One more thing from run 6, unrelated to the streak: a frozen fprintd holds its sleep delay
inhibitor and cannot release it, so logind waited out the full inhibitor timeout —
**15.1 s** between `Suspending...` and `Successfully froze unit 'user.slice'`, versus ~740 ms
in the normal runs (`Delay lock is active (UID 0/root, PID .../fprintd) but inhibitor timeout
is reached`). Not something this PR causes or needs to handle; noting it because a wedged
daemon delaying every suspend by the inhibitor window is its own user-visible symptom.

Also present in every run, both pre-existing: `Failed to install a sleep delay inhibitor:
OperationInProgress` when the lock starts the daemon just before suspend (the #8531 / #7229
shape), and ~570 ms after resume three lines of `Authorization denied ... VerifyStop` /
`... Release` / `ReleaseDevice failed: Device was not claimed before use` — the lock screen
addressing the fresh daemon with the old one's handle.

Happy to run anything else on this hardware.

GeertJohan:
@omarchybot Two of the four taken, at `c817d77b` and `c10d86d9`:

- The cap's idle guard now survives a backwards clock step: a negative settle gap counts as no idle time at the cap, while a negative nudge gap still does not hold a fresh nudge back. Test pins both directions.
- The drop-in is `10-stop-timeout.conf`, so `override.conf` from `systemctl edit` sorts after it and wins. The migration replaces a copy installed under the old name.

The `--no-block` ceiling is left as is, on your own reading of the trade. The stall-as-resume case has a comment on `spannedSleep` saying why it is tolerated. The once-per-lock status check predates the PR and stays out of scope here.

osborng:
Two findings from running a local equivalent of layer 1 through real lid-close suspends this week. Both are about the same window, and I think the second one partly undercuts what this PR is building.

**1. `--no-block` leaves the window the loop then has to survive.**

The hook uses:

```
systemctl --no-block try-restart fprintd.service
```

which returns immediately, so the restart is still in flight when `user.slice` thaws and the lock screen starts asking. The comment for `RESUME_GRACE_MS` acknowledges exactly this ("for this long after a resume the hook is restarting fprintd underneath the loop").

A synchronous start closes it rather than pacing around it: `system-sleep` `post` hooks run **before** systemd thaws `user.slice`, so a blocking start finishes before the shell can probe, and `systemctl start` on a `Type=dbus` unit only returns once the bus name is acquired. The reason for `--no-block` is the wedged-daemon stop, but that is what `10-stop-timeout.conf` already bounds — so `stop` (bounded) followed by a blocking `start` gets both properties.

I moved my own hook from "stop the stale PID and let D-Bus re-activate" to "stop, then start a fresh daemon" for precisely this reason, after a resume where the stop landed at 14:49:37, the lock screen probed at 14:49:38, and **fprintd was never activated at all** — no `Starting Fingerprint Authentication Daemon` in the journal, no prompt, password at 14:49:51. Leaving the bus name unowned across the resume boundary is not reliably re-activatable, even though it always is in steady state.

**2. The probe's answer is still terminal, and this PR widens what that costs.**

`fingerprintCheckProc` is unchanged in the part that matters:

```qml
root.fingerprintConfigured = String(fingerprintCheckStdout.text || "").trim() === "yes"
```

An unreachable fprintd — the very thing the resume window produces — makes `fprintd-list` print

```
Could not activate remote peer 'net.reactivated.Fprint': activation request failed
```

with no `' - #'` line in it, so this reads as "not configured" rather than "could not tell". In this PR that now also does:

```qml
} else if (!root.fingerprintConfigured) {
  if (fingerprintPam.active) fingerprintPam.abort()
  root.settleFingerprintAttempt()
  fingerprintRetryTimer.stop()
}
```

and `fingerprintSleepWatch` is gated `running: root.lockRequested && root.fingerprintConfigured`. So one unreachable probe stops the retry loop *and* switches off the resume detector this PR adds — and nothing re-probes for the rest of the lock, since `refreshFingerprintStatus()` is reachable only from `beginLock()` and from a failed `fingerprintPam.start()`. The backoff, the grace window and the unavailable notice are all downstream of a flag that a single transient miss has already cleared.

I have filed this separately as #9453 since it reproduces on `main` independently of this PR, with a deterministic repro (mask fprintd, lock, unmask while locked — fingerprint never returns for that lock). Mentioning it here because the fix belongs next to this work: making the probe three-state (`yes` / `no` / `unknown`, keying on `fprintd-list`'s output rather than its exit status, which is flaky) would let the machinery in this PR actually see the resume it is designed for.

Also worth saying: the change to `grep -q ' - #'` here is a real fix I hit independently. The current `grep -qi finger` matches `has no fingers enrolled` and the device name `Goodix MOC Fingerprint Sensor`, so an unenrolled user reads as configured.

Environment: Omarchy 4.0.2-1, Quickshell 0.3.1-1, fprintd 1.94.5-2, libfprint 1.94.100-1, kernel 7.1.9-arch1-2, Goodix 27c6:609c.

touchapp:
Another independent hardware data point supporting the post-resume `fprintd` recovery layer in this PR.

### Environment

- Lenovo `21CBCTO1WW`
- Synaptics Prometheus `06cb:00fc`
- Omarchy `4.0.2-1`
- Linux `7.1.9-arch1-2`
- `fprintd 1.94.5-2`
- `libfprint 1.94.100-1`

Fingerprint authentication repeatedly became unavailable after lid-close/suspend, so I installed a resume hook. The journal signature on affected resumes included:

```text
Device reported an error during identify: device was disconnected
Device was not cleaned up properly before being finalized.
User destroyed open device! Not cleaning up properly!
libusb: warning [libusb_exit] device ... still referenced
```

There was also one resume-time initialization failure:

```text
Ignoring device due to initialization error: unsupported firmware version
```

The local workaround is a synchronous post-resume hook:

```sh
if [ "$1" = "post" ]; then
  systemctl restart fprintd.service
fi
```

With the hook installed, affected resumes show the disconnect/cleanup errors immediately followed by `Stopped Fingerprint Authentication Daemon` and `Started Fingerprint Authentication Daemon`, preventing that stale daemon from surviving the wake. The reader currently enumerates normally and `fprintd-list` sees both enrolled prints.

An immediate restart has been sufficient on this `06cb:00fc`; no additional delay was needed. This independently supports the PR's bounded post-resume `try-restart` approach. I will keep the local hook until the upstream migration lands.

GeertJohan:
@osborng Finding 2: fixed in `2dbd327c`. The probe is three-state now, keyed on output as you suggested — a ` - #N` row means yes, the explicit "has no fingers enrolled" (or a missing PAM file/binary) means no, anything else means unknown: state stays as it was and the probe retries on the attempt pacing. Your mask/unmask repro is in the fixture; the commit closes #9453. It's also what killed the reader in MaxMad75's run 6.

Finding 1: the failed activation you saw was with a stop-only hook. This branch enqueues a restart job that systemd runs itself, no activation involved — MaxMad75 measured the fresh daemon at +357–589 ms in six of six lid closes. A synchronous start blocks the frozen thaw until the start timeout when a daemon never reaches its bus name (100 s measured when the old bound was added), so it would need its own timeout again. The remaining corner, try-restart as a no-op plus a failed activation after resume, now recovers through the retry loop, since a bad probe can no longer switch that loop off.

GeertJohan:
The wedge itself turns out to be long-filed upstream: [gitlab.freedesktop.org/libfprint/fprintd#173](https://gitlab.freedesktop.org/libfprint/fprintd/-/issues/173) (2024, still open) and [gitlab.freedesktop.org/libfprint/fprintd#216](https://gitlab.freedesktop.org/libfprint/fprintd/-/issues/216) are exactly this — a verify killed by the daemon's own suspend handling, the release then failing on the busy device, the claim wedged until restart. The orphaned-claim variant is [gitlab.freedesktop.org/libfprint/fprintd#214](https://gitlab.freedesktop.org/libfprint/fprintd/-/issues/214). `9e694f49` marks the resume hook as a workaround for those, so it and the stop-timeout drop-in can be removed once upstream fixes it.

Downstream the same defect sits open elsewhere too: [bugs.launchpad.net/ubuntu/+source/fprintd/+bug/2067135](https://bugs.launchpad.net/ubuntu/+source/fprintd/+bug/2067135) (confirmed, unassigned), [gitlab.gnome.org/GNOME/gnome-shell#7791](https://gitlab.gnome.org/GNOME/gnome-shell/-/issues/7791) and [gitlab.gnome.org/GNOME/gnome-control-center#3166](https://gitlab.gnome.org/GNOME/gnome-control-center/-/work_items/3166). Debian and Ubuntu ship no mitigation.

GeertJohan:
Rebased onto latest `quattro` (`a7030926`, 187 commits ahead of the old base) and since we're rewriting history anyway, I condensed the history from 26 commits to 12. The tree at `18dfb688` is byte-identical to the previous tip and verified with an empty `git diff` between the two, and all twelve commits pass `apply-lock-test`, `lock-fingerprint-retry`, `fprintd-resume-hook` and `fprintd-resume-migration` individually.

Two upstream collisions worth naming:

- #10225 pinned `fprintd-list` to `/usr/bin` in `omarchy-apply-lock` and added `apply-lock-test.sh`. The resolution keeps the pinned form and this branch's enrolled-print match (`grep -q ' - #'`; upstream's line still had the `grep -qi finger` bug fixed here), and the test's helper patcher now also redirects the resume hook and the stop-timeout drop-in into its scratch tree.
- The video-wallpaper work (#6792) landed in the lock plugin; it composes with the fingerprint changes without interaction.

Attribution held through the squash. For anyone cross-referencing earlier hardware runs: those SHAs are orphaned by the rebase; the code they exercised is unchanged in `18dfb688`.

shell/plugins/lock/Service.qml
```diff
@@ -179,6 +229,53 @@ Item {
...
+    logEvent("fingerprint-resume: streak=" + fingerprintUnreachedStreak)
+    fingerprintResumedAtMs = now
+    fingerprintUnreachedStreak = 0
+  }
+
+  // A resume noticed while a backed-off wait is pending: retry the fresh
+  // reader now, instead of after the remaining wait or the next keypress.
+  function restartFingerprintAfterSleep() {
+    noteFingerprintResumed()
+    if (fingerprintAuthenticating || fingerprintPam.active) return
```
sliekens: **[suggestion]** After a lock that already had a fingerprint verify open, resume lands here with `fingerprintPam.active` still true. This return then skips the retry, so the in-flight session is left sitting on the dead D-Bus call.

That is #8747: `pam_fprintd`'s `ReleaseDevice` does not fail promptly when the resume hook restarts fprintd under it; it waits ~25s. `REACH_TIMEOUT_MS` cannot help because `noteFingerprintReachedDevice()` already stopped the reach timer at the pre-sleep prompt. The icon still invites a touch that cannot unlock, and `RESUME_GRACE_MS` (5s) has expired by the time this session settles.

Especially visible when `user.slice` is not frozen (`SYSTEMD_SLEEP_FREEZE_USER_SESSIONS=false` from the NVIDIA drop-in): the lock and the hook race, and this return wins.

Recovery still happens — the *next* attempt should claim the fresh daemon — but the hook comment's "that attempt fails and the next, a second later, claims" is not what this path does.

**Suggestion:** Abort the in-flight fingerprint PAM session here (`fingerprintPam.abort()`), then arm the short retry, the same way a backed-off wait is collapsed below. The reach timeout remains the backstop for a Claim that never prompts.

GeertJohan: Fixed in `e77ed414`: the resume detection now aborts a session still in flight and routes it through settle — the settle lands inside the grace window, so the kill never counts toward the notice, and it arms the 250 ms retry against the fresh daemon itself. A fixture step pins your scenario and fails on the previous code.

One qualifier: on stock fprintd the window is narrow, since its own PrepareForSleep handling usually kills the prompted verify before the sleep — this mostly matters for daemons without suspend handling (#8747's stack) and for the unfrozen-session race you name. Worth having either way; the false-positive cost of a >3 s event-loop stall reading as a resume is one aborted prompt, re-armed 250 ms later.

johnpippett:
This report gives a test result for #7229 and the restart script in this PR.

I prepared this report with the assistance of AI (Codex). I did the fingerprint test on the laptop.

| Item | Value |
| --- | --- |
| Laptop | ASUS ExpertBook P5405CSA |
| Fingerprint reader | ELAN `04f3:0c9d`, `Elan MOC Sensors` |
| Omarchy | `4.0.2-1` |
| Quickshell | `0.3.1-1` |
| fprintd | `1.94.5-2` |
| libfprint | `1.94.100-1` |
| Kernel | `7.1.9-arch1-2` |
| systemd | `261.2-1` |
| Sleep mode | `s2idle` |

Before this change, the reader failed after sleep. The saved fingerprint passed `fprintd-verify` while the laptop was unlocked.

The system recorded these errors. I removed the hostname and username from the log text.

```text
20:24:31 fprintd: Device reported an error during verify: Cannot run while suspended.
20:24:31 pam_fprintd(omarchy-lock-fingerprint:auth): ReleaseDevice failed: Release failed with error: The device is still busy with another operation, please try again later.
20:26:14 kernel: PM: suspend exit
20:26:14 quickshell.service.pam.subprocess: Error while authenticating: "Authentication service cannot retrieve authentication info" (code 9)
```

The errors continued until I used the password at 20:26:32.

For the test, I used this script at `/usr/lib/systemd/system-sleep/fprintd-resume`. The file owner was `root:root`. The file mode was `0755`.

```bash
#!/bin/bash
if [[ ${1:-} == post ]]; then
  /usr/bin/systemctl --no-block try-restart fprintd.service
fi
```

The test used the existing fingerprint, PAM files, and lock plugin. The service already had a stop timeout of 5 seconds. The test did not use the timeout change in this PR.

The test on 2026-09-10 gave these results:

- 20:49:34: The laptop entered sleep.
- 20:49:45: The laptop resumed. The fingerprint service stopped and started.
- 20:49:46: Fingerprint authentication succeeded. The lock screen opened.

The successful PAM process used `omarchy-lock-fingerprint`. It showed the finger prompt and recorded `Authenticated successfully`. The lock screen then recorded `unlocked`.

An old authentication request recorded a claim error during the restart. The next request succeeded.

This test covered one sleep/wake cycle. It did not test the complete PR, the QML changes, or the installation and migration procedures.

Razuer:
Tested the recovery hook and 3s stop-timeout drop-in on a Goodix `27c6:6890` with Omarchy 4.0.3, fprintd 1.94.5-2, libfprint-git 1.94.100.r10, and s2idle.

This machine previously hit `Cannot run while suspended` with `ReleaseDevice ... still busy`. On a real suspend/resume, the hook stopped fprintd at +15ms, the fresh daemon started at +415ms, and the automatic lock-screen retry authenticated successfully at +1.29s. No password fallback was needed.

This test covered the resume hook and timeout drop-in, not the QML retry/backoff changes.

codenamekt:
### Verification: Omacloud Test Harness

✅ **Passed** — All 77 unit, migration, security, and backoff simulation assertions passed (1 skipped). Live QML lockscreen indicators visually verified in Hyprland session.

| | |
|---|---|
| **PR** | [#7158](https://github.com/omacom/omarchy/pull/7158) — Keep lock screen fingerprint working across suspend |
| **Author** | @GeertJohan |
| **Head Commit** | `e77ed414f283` |
| **Base Branch** | `quattro` |
| **Verification Scope** | `verifiable-graphical` (Hybrid Lockscreen QML + Systemd Sleep Hook) |
| **Test Output** | **77 passed · 0 failed · 1 skipped** (100% test pass rate) |
| **Environment** | Hyprland 0.56.2 + Quickshell (`omarchy-shell`) · 1920×1080 · Intel UHD 630 iGPU |

---

### Graphical Lockscreen Verification

Executed inside a live graphical session to verify the visual state transitions in `LockView.qml`:

1. **Working Sensor (`lock-fingerprint-working.png`):**
   * Fingerprint reader icon renders cleanly inside the password entry pill with proper reserve spacing to prevent text collision with 80-character inputs.
2. **Unavailable Sensor (`lock-fingerprint-unavailable.png`):**
   * Fingerprint icon automatically crosses out.
   * Banner notice `"Fingerprint reader unavailable"` renders directly below the password input.

#### Screenshots Captured During Run:

* **Working Reader:**
lock-fingerprint-working


* **Unavailable Reader Notice:**
lock-fingerprint-unavailable


---

### Scope & Hardware Boundary Verification

* **✅ Verified in VM (77/77 assertions passed):**
  * **Timing & Backoff Model (`FingerprintModel.js`):** 63 Node.js assertions verified exponential backoff doubling, 30s idle daemon clear window, clock step-back recovery, resume grace periods, and user input nudge collapsing.
  * **Systemd Integration:** Verified `system-sleep/fprintd-resume` hook enqueues restarts and `10-stop-timeout.conf` bounds `TimeoutStopSec`.
  * **Security Sandboxing:** Verified `omarchy-apply-lock` pins trusted executable paths and prevents privilege escalation.
* **⚠️ Hardware Limitation (Honest Disclosure):**
  * QEMU does not emulate proprietary physical fingerprint biometric sensors (e.g. Goodix / Validity). Physical sensor wake from hardware ACPI S0ix/S3 suspend must be validated on physical laptop hardware.

---

<details>
<summary><b>TAP Assertion Breakdown (77 / 77 OK)</b></summary>

```tap
ok - the root lock helper uses only trusted command directories
ok - the lock helper pins fprintd-list to its packaged system path
ok - the resume hook is executable
ok - resume enqueues an fprintd restart to clear a wedged claim
ok - resume from hibernate also restarts fprintd
ok - the stop-timeout drop-in bounds TimeoutStopSec
ok - pre-suspend leaves fprintd alone
ok - migration installs the shipped hook and drop-in from their default sources
ok - migration installs the hook, executable
ok - migration installs the stop-timeout drop-in and reloads systemd
ok - migration leaves existing files alone
ok - migration replaces the unprefixed drop-in with the numbered one
ok - migration adds the drop-in beside an existing hook
ok - migration skips machines without fingerprint configured
ok - a reached attempt retries at the fast interval
ok - the first unreached attempt backs off
ok - consecutive unreached attempts double the delay
ok - the backoff keeps doubling
ok - the backoff is capped
ok - the model carries fprintd's 30-second idle exit
ok - the cap exceeds fprintd's idle exit so a wedged claim can clear
ok - the idle stretch a nudge must leave at the cap covers the exit and fits inside the cap
ok - the backoff never shrinks while attempts keep failing to reach the reader
ok - reaching the reader keeps the streak at zero
ok - reaching the reader clears an accumulated streak
ok - failing to reach the reader starts the streak
ok - failing to reach the reader advances the streak
ok - a working reader is not reported unavailable
ok - a couple of misses do not report unavailable
ok - enough consecutive misses report the reader unavailable
ok - a run of unreached attempts ends up unavailable
ok - one reached attempt clears the unavailable state
ok - no nudge when the wait is already the fast interval
ok - a keypress collapses a backed-off wait
ok - a second nudge inside the cooldown is refused
ok - a nudge one cooldown later is allowed
ok - a clock stepped backwards does not refuse the nudge
ok - a clock stepped backwards since the settle does not refuse it either
ok - below the cap a fresh settle does not hold the nudge
ok - a moving cursor is capped at one collapse per cooldown, got 6
ok - a nudge one tier later is allowed
ok - a nudge inside the tier is refused
ok - continuous input at the cap collapses once per tier
ok - a nudge before the idle stretch is refused at the cap
ok - a nudge after the idle stretch is allowed at the cap
ok - a clock stepped back past the settle does not bypass the idle guard at the cap
ok - a clock stepped back past the nudge still allows the nudge at the cap once idle
ok - continuous input against attempts that hang to the reach bound still leaves fprintd its idle exit, got 32250ms
ok - the nudge still shortens the wait at the cap rather than being dead there, got 32250ms
ok - a miss inside the resume grace holds the streak at the first tier
ok - the grace pins any accumulated streak to the first tier
ok - a reached attempt inside the grace clears the streak
ok - outside the grace misses accumulate
ok - the grace outlasts fprintd's 3s stop cap plus its start
ok - the grace is shorter than the cap, so it cannot mask a reader that is really gone
ok - no resume noted means no grace
ok - an attempt settling inside the window is in grace
ok - an attempt settling at the window edge is not
ok - a clock stepped back past the resume is not grace
ok - the bound plus slack is not a sleep
ok - longer than the bound plus slack is a sleep
ok - an eight-hour wait on a one-second timer is a sleep
ok - a miss 250ms after resume must not show the notice
ok - a miss 1250ms after resume must not show the notice
ok - a miss 2250ms after resume must not show the notice
ok - a miss 3250ms after resume must not show the notice
ok - the loop keeps retrying through the restart window, got 4
ok - the first attempt after the daemon is back comes within a tier, at 4250ms
ok - an enrolled print row reads yes
ok - the probe script's own no reads no
ok - fprintd's explicit no-prints answer reads no
ok - an activation failure reads unknown, not unconfigured
ok - a D-Bus timeout reads unknown
ok - empty output reads unknown
ok - the reach bound is shorter than the backoff cap, so a stuck attempt is caught well before the cap
ok - the reach bound lands before GDBus fails a stuck Claim at 25s and pam_fprintd's 30s verify timeout reports as a non-error message
ok - the reach bound outlasts a slow device open, so a reader that takes seconds to claim is not killed mid-Claim on every attempt
```

</details>

<sub>Verified automatically with <b>Omacloud Test Harness</b> · Run duration: 14.8s</sub>

gbrettmiller:
## Test report — layer 1 (resume hook + stop-timeout) on Goodix MOC 27c6:609c

**System:** Omarchy `4.0.4-1`, kernel `7.2.3-arch1-3`, systemd `261.2-1-arch`, fprintd `1.94.5-2`, libfprint-git `1.94.100.r10.g6f9479c-1`, Goodix MOC Fingerprint Sensor `27c6:609c`, sleep mode `s2idle`, `SYSTEMD_SLEEP_FREEZE_USER_SESSIONS=0`.

**Scope tested:** Layer 1 only — the `fprintd-resume` system-sleep hook (`systemctl --no-block try-restart fprintd.service`) and the `10-stop-timeout.conf` drop-in (`TimeoutStopSec=3s`). I did not test the QML retry-pacing/"reader unavailable" notice layer (`Service.qml`/`FingerprintModel.js`/`LockView.qml`) — no active lock-screen/Hyprland surface in this environment to exercise it against (`omarchy-sleep-lock.service` isn't present here).

**Baseline (pre-fix, this session):** before installing anything, `fprintd-list` reported **zero enrolled fingers** despite the user having enrolled previously — consistent with the wedged-claim/lost-pairing failure family this PR targets. USB reset confirmed on resume: `usb 3-9: reset full-speed USB device number 4 using xhci_hcd`.

**Procedure:**
1. Installed the PR's `fprintd-resume` hook to `/usr/lib/systemd/system-sleep/` and `10-stop-timeout.conf` to `/etc/systemd/system/fprintd.service.d/`, `systemctl daemon-reload`.
2. Re-enrolled one print (`right-index-finger`).
3. Ran two real `systemctl suspend` cycles (s2idle), confirmed via kernel log (`PM: suspend entry` → `PM: suspend exit`, ~7-8s each).
4. Checked `fprintd-list`, `systemctl status fprintd`, and journal after each cycle.

**Results — both cycles clean:**
```
22:24:47  Stopping fprintd → Stopped → Starting → Started   (resume 1)
22:25:28  Stopping fprintd → Stopped → Starting → Started   (resume 2)
```
USB reset still occurs on every resume (`usb 3-9: reset full-speed USB device number 4`) — expected; the hook works around it rather than preventing it. `fprintd-list` after both cycles still showed the enrolled print intact:
```
Fingerprints for user gbrett on Goodix MOC Fingerprint Sensor (press):
 - #0: right-index-finger
```

**Not conclusive:** a live `fprintd-verify` touch test immediately after the second resume was left waiting and not completed with an actual touch before wrapping up, so I can't confirm same-session post-resume *verify* success — only that the daemon cleanly restarts and enrollment data persists in fprintd's/the sensor's view across suspend.

**Not tested:** hibernation (the second trigger in my original report) — deferred since it requires a full power cycle; suspend-only result is what's reported here.

**Bottom line:** on this exact hardware (Goodix MOC `27c6:609c`, not previously represented in this thread), layer 1 of this PR does what it claims — it clears the wedged claim on resume, and across two cycles the enrollment that was previously getting lost stayed intact. No regressions observed.

Issue #6858, opened 2026-08-14
Lock screen auto-focus not working


Obamium3157:
### System details

Omarchy: 4.0.0.r1732.g864b0d0-1 (edge channel). Kernel: 7.1.8-arch1-3; Hyprland: 0.56.2-1; quickshell-git: 0.3.0.r20.g28771c7-1. CPU: 11th Gen Intel Core i5-1135G7; GPU: Intel Iris Xe + NVIDIA RTX 3050 Mobile

### What's wrong?

The Quickshell lock screen (`omarchy.lock` service, `shell/plugins/lock/LockView.qml`) does not auto-focus the password field. After locking, the user must click the field before typing works.

## Root cause
`LockView.qml:55-77` calls `passwordInput.forceActiveFocus()` once from `Component.onCompleted` and once from `onInputEnabledChanged`, both via `Qt.callLater`. The `WlSessionLockSurface/LockView` is created ~500 ms before Hyprland maps the surface and hands over keyboard focus. So `forceActiveFocus()` lands while the lock window is still hidden/inactive, and the window's later activation does not re-apply the pending active focus. Clicking works because it sets focus while the window is already active.

## Proposed fix (verified on my machine)
Retry `forceActiveFocus()` until the field actually holds focus, re-armed at lock-request and when auth finishes:
```qml
function armFocusRetry() {
    if (!inputEnabled) return
    focusRetry.remaining = 50
    focusRetry.restart()
}
onInputEnabledChanged: {
    if (inputEnabled) { Qt.callLater(forcePasswordFocus); armFocusRetry() }
}
onAuthenticatingPasswordChanged: {
    if (!authenticatingPassword && inputEnabled) armFocusRetry()
}
Component.onCompleted: {
    syncPasswordText()
    if (inputEnabled) { Qt.callLater(forcePasswordFocus); armFocusRetry() }
}
Timer {
    id: focusRetry
    interval: 120
    repeat: true
    property int remaining: 0
    onTriggered: {
        if (!root.inputEnabled || root.authenticatingPassword || passwordInput.activeFocus) { stop(); return }
        root.forcePasswordFocus()
        remaining -= 1
        if (remaining <= 0) stop()
    }
}
```

perfektnacht:
Same root cause, but a second trigger that the proposed fix above would not cover: **suspend/resume on an already-focused lock screen.**

[...]

PR #7164, opened 2026-08-16
Fix password field losing autofocus on lock screen

nickav:
I noticed that sometimes on the password screen when I type my input does not go into the password field.

According to Claude:

> forcePasswordFocus() was only triggered by inputEnabled toggling true, which fires before the compositor grants the session lock surface exclusive keyboard focus. Also force focus once WlSessionLock reports secure=true, when focus is actually available.

### DIFF
--- a/shell/plugins/lock/LockView.qml
+++ b/shell/plugins/lock/LockView.qml
@@ -14,6 +14,7 @@ Item {
   property int failedAttempts: 0
   property bool inputEnabled: true
   property bool loadBackground: true
+  property bool secure: false
   property string passwordText: ""
   property bool syncingPasswordText: false
 
@@ -70,6 +71,9 @@ Item {
   onPasswordTextChanged: syncPasswordText()
   onInputEnabledChanged: {
     if (inputEnabled) Qt.callLater(forcePasswordFocus)
+  }
+  onSecureChanged: {
+    if (secure) Qt.callLater(forcePasswordFocus)
   }
   Component.onCompleted: {
     syncPasswordText()
--- a/shell/plugins/lock/LockView.qml
+++ b/shell/plugins/lock/LockView.qml
@@ -56,6 +56,28 @@ Item {
     passwordInput.forceActiveFocus()
   }
 
+  function retryPasswordFocus() {
+    if (!inputEnabled || authenticatingPassword) return
+    forcePasswordFocus()
+    focusRetryTimer.attemptsLeft = 20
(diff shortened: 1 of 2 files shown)


pjover:
This fix hooks `forcePasswordFocus()` to `onSecureStateChanged`, which covers the initial lock. There is a second path it won't reach: the lock surfaces being **rebuilt while already locked**, which happens on resume when the output set changes.

Reproduction — ThinkPad docked to an external display, lid closed (Omarchy 4.0.0, Hyprland 0.56.2, quickshell 0.3.0.r20.g28771c7, Qt 6.11.1):

1. Lock while in clamshell, so only the external output is active (`realScreens: 1`).
2. Suspend.
3. Wake by opening the lid. The internal panel returns, so the output set goes 1 → 2 *while locked*, and Quickshell tears down and recreates the lock surfaces (`Not creating lock surface for screen ... not backed by a valid wayland output`).
4. The password field is drawn and the pointer works, but no keystroke produces a dot. Clicking the surface fixes it instantly — the MouseArea calls `forcePasswordFocus()` directly.

`secure` does not change across that rebuild, so `onSecureStateChanged` never re-fires:

```
19:01:22  secure=true     <- initial lock, before suspend
19:01:37  (resume; outputs 1 -> 2; lock surfaces recreated)
19:02:19  secure=false    <- unlock, 38s later
```

Instrumented during the dead window: keypresses were arriving at evdev, Hyprland still listed 7–9 keyboards including the built-in one, and `omarchy-shell lock status` stayed `authenticating:false` with `lastEvent: secure=true` throughout. So the keystrokes reached the compositor and were dropped for want of a focused item, rather than being lost below it.

Controlled A/B on the same machine, one variable changed:

| Wake method | realScreens | Dead window |
|---|---|---|
| open the lid | 1 → 2 | 38 s |
| power button, lid stays closed | 1 throughout | none |

Keeping the output set constant avoids the surface rebuild entirely, and the field keeps focus.

This also lines up with the Copilot review note about `lockView` possibly being invalid by the time the deferred call runs — in this path the LockView genuinely is destroyed and recreated. Tying the re-focus to the LockView/surface lifecycle (or the retry Timer suggested in #6858) would cover both the initial lock and the rebuild-while-locked case; `onSecureStateChanged` alone only covers the first.

Chessing234:
hey, this is already covered in https://github.com/omacom/omarchy/pull/12894 (lock keyboard focus after lid-open / resume). mind closing this as a duplicate?

PR #7167, opened 2026-08-16
fix: filter pre-release tags in omarchy-update-available

### LINKED ISSUE #7166: omarchy-update-available false positive when a pre-release tag exists (e.g. v4.0.0-beta3)
**Description**

The waybar update-checker script (`bin/omarchy-update-available`) permanently reports an update is available even when the local install is on the actual latest release tag.

**Root cause**

```bash
latest_tag=$(git ls-remote --tags origin | grep -v "{}" | awk '{print $2}' | sed 's#refs/tags/##' | sort -V | tail -n 1)
```

`sort -V` (GNU version sort) does not implement SemVer pre-release ordering. It ranks `v4.0.0-beta3` above `v4.0.0`, because it's comparing byte-wise rather than treating `-beta3` as a pre-release qualifier that should sort before its corresponding release.

**Reproduction**

```bash
$ git ls-remote --tags origin | grep -v "{}" | awk '{print $2}' | sed 's#refs/tags/##' | sort -V | tail -n 3
v3.8.4
v4.0.0
v4.0.0-beta3
```

With a fully up-to-date checkout on `v4.0.0`:

```bash
$ omarchy-update-available
Omarchy update available (v4.0.0-beta3)
```

This will affect every user on `v4.0.0` for as long as `v4.0.0-beta3` remains the lexically/version-sort-"highest" tag in the repo — the waybar icon and `omarchy-update` prompt never clear.

**Suggested fix**

Filter out pre-release tags before taking the max, e.g.:

```bash
latest_tag=$(git ls-remote --tags origin | grep -v "{}" | awk '{print $2}' | sed 's#refs/tags/##' | grep -vE -- '-[A-Za-z]' | sort -V | tail -n 1)
```

(or swap in a proper SemVer-aware comparator if pre-release tags need to be considered under different circumstances).

**Environment**
- Omarchy version: v4.0.0
- OS: Arch Linux

sergioacosta-dev:
## Summary
- Fixes #7166 — `omarchy-update-available` permanently reports an update is available on a fully up-to-date install because `sort -V` doesn't implement SemVer pre-release ordering, ranking `v4.0.0-beta3` above `v4.0.0`.
- Excludes tags with an alpha suffix (`-beta`, `-rc`, etc.) before taking the version-sorted max, so only real release tags are considered "latest".

## Test plan
- [x] Reproduced the false positive on a `v4.0.0` checkout (`omarchy-update-available` reported `v4.0.0-beta3` as available)
- [x] Applied the fix and confirmed it now reports "Omarchy is up to date (v4.0.0)" with exit code 1
- [x] Confirmed filtered remote tag list still correctly resolves to `v4.0.0` as the highest real release

### DIFF
--- a/bin/omarchy-update-available
+++ b/bin/omarchy-update-available
@@ -2,7 +2,7 @@
 
 # omarchy:summary=Get remote tag
 
-latest_tag=$(git -C "$OMARCHY_PATH" ls-remote --tags origin | grep -v "{}" | awk '{print $2}' | sed 's#refs/tags/##' | sort -V | tail -n 1)
+latest_tag=$(git -C "$OMARCHY_PATH" ls-remote --tags origin | grep -v "{}" | awk '{print $2}' | sed 's#refs/tags/##' | grep -vE -- '-[A-Za-z]' | sort -V | tail -n 1)
 if [[ -z $latest_tag ]]; then
   echo "Error: Could not retrieve latest tag."
   exit 1


moonveiloco:
Confirmed this fixes the false positive on a real v3.8.5 install.

**Before (current master script):**
```
latest=`git ls-remote --tags origin | ... | sort -V | tail -n 1`          -> v4.0.0-beta3
current=`git describe --tags $(git rev-list --tags --max-count=1)`        -> v4.0.0
=> "Omarchy update available (v4.0.0-beta3)"  (false positive)
```

**After (this PR's grep filter):**
```
latest -> v4.0.0
current -> v4.0.0
=> "Omarchy is up to date (v4.0.0)"  (no more false positive)
```

One thing worth flagging before merge: the `current_tag` line still derives the installed version from the **most recently tagged commit** (`git rev-list --tags --max-count=1`), not from HEAD. On a v3.8.5 master checkout that resolves to `v4.0.0` (the tag lives on a feature branch), so the message reads "up to date (v4.0.0)" even though the installed version is 3.8.5. It doesn't cause a false positive today, but it will once `v4.0.0` is the newest real release on master: the checker will then nag users who are on master but can't reach v4 via `git pull`. The dev branch already sidesteps this by restricting to the current release series (`^v3\.[0-9]+\.[0-9]+$`). Consider either anchoring current to HEAD (`git describe --tags --abbrev=0 HEAD`) or aligning with the dev-branch series filter.

Otherwise the fix is correct and minimal.

PR #7171, opened 2026-08-16
Make Super+J manage scrolling columns

fayimora:
Make `Super + J` perform the native window-layout action for the active workspace layout:

- Dwindle continues to dispatch `togglesplit`.
- Scrolling dispatches `consume_or_expel prev`.
- Active special workspaces take precedence over regular workspaces.

This lets the same spatial-layout shortcut toggle window positioning in Dwindle and consume or expel windows in scrolling columns.

The hotkey and navigation manuals now document both behaviors, with regression coverage for regular, special, missing, and unsupported workspace layouts.

Alternative to #6924

https://github.com/user-attachments/assets/fafa239c-c000-4fe3-9d5a-b3b8ab32e9d0

### DIFF
--- a/default/hypr/bindings/tiling.lua
+++ b/default/hypr/bindings/tiling.lua
@@ -1,7 +1,21 @@
 o.bind("SUPER + W", "Close window", hl.dsp.window.close())
 o.bind("CTRL + ALT + DELETE", "Close all windows", "omarchy-hyprland-window-close-all")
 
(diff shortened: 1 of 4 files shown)


bjarneo:
Automated duplication check: this pull request looks similar to #11863, which covers the same SUPER+J layout dispatch. I keep that one open and close this one to consolidate review.

PR #7177, opened 2026-08-16
Mark the T2 Mac internal trackpad as internal

georgehenderson:
On a T2 MacBook, palms resting on the trackpad move the cursor while typing. `input:touchpad:disable_while_typing` is set and reads back as enabled, and it has never done anything.

## Why

The T2 bridge exposes the built-in keyboard and trackpad over a virtual USB host controller (`t2bce_vhci`), so the trackpad carries `ID_BUS=usb` and `/usr/lib/udev/rules.d/65-integration.rules` classifies it as an external touchpad. libinput makes disable-while-typing unavailable on external touchpads, so there is nothing under the compositor option to act on:

```
$ hyprctl getoption input:touchpad:disable_while_typing
int: 1
$ libinput list-devices | grep -A20 Trackpad | grep Disable-w-typing
Disable-w-typing:  n/a
```

Both are true at once, which is what makes this hard to find. Nothing reports an error; the setting simply has no effect.

## Fix

An hwdb entry correcting the classification. `70-touchpad.rules` runs after `65-integration.rules` precisely so hwdb entries can override the bus-based guess, so this is the supported correction rather than a workaround.

This is the same class of bug as `install/hardware/asus/fix-z13-touchpad.sh`, where a detachable keyboard's touchpad is also detected as external.

A migration carries the fix to installs that predate it, since the install leaf only reaches machines set up after it ships. It is a machine-wide repair, so a second user on the same laptop finds it already done.

## Scope

Matched on the USB ID the bridged keyboard/trackpad enumerates with (`05ac:0340`), verified on a MacBookPro16,1. Other T2 models may enumerate a different product ID, so the match is deliberately narrow rather than a guessed range; widening it as models are reported is one line.

I also tested a udev rule matching on `ATTRS{phys}=="usb-t2bce_vhci*"`, which would cover every T2 Mac regardless of product ID. It sets `ID_INPUT_TOUCHPAD_INTEGRATION` correctly and tags no keyboard node, but leaves `ID_INTEGRATION=external` where the hwdb path flips both. I did not confirm that libinput then offers disable-while-typing, so I have not proposed it here. Happy to pursue it if you would prefer the broader match.

## Testing

`test/shell.d/t2-touchpad-test.sh` covers the leaf and the migration with stubbed `lspci`/`systemd-hwdb`, including that the hwdb property keeps its leading space (an unindented line becomes a match that never fires, so the file would silently stop working).

```
$ bash test/shell.d/t2-touchpad-test.sh
ok - the trackpad fix is orchestrated and its hwdb entry is well formed
ok - a T2 Mac gets the trackpad hwdb entry
ok - non-T2 hardware is left alone
ok - the migration fixes a T2 install that never got the entry
ok - the migration is machine-idempotent across users
ok - a match with no property is rewritten
ok - the migration skips unrelated hardware
```

`t2-hardware-test.sh`, `brcmfmac-supplicant-test.sh` and `./test/cli` still pass.

Verified on the machine itself: with the entry installed and after a fresh login, `Disable-w-typing` reports `enabled` and palms resting on the pad no longer move the cursor.

## Not included

Lowering `AttrPalmSizeThreshold` also helps, but the value I measured is specific to this trackpad's geometry and could cause false rejections on models nobody has measured, so it is left out.

### DIFF
--- a/install/hardware/all.sh
+++ b/install/hardware/all.sh
@@ -35,6 +35,7 @@ run_logged "$OMARCHY_INSTALL/hardware/framework/qmk-hid.sh"
 run_logged "$OMARCHY_INSTALL/hardware/apple/fix-spi-keyboard.sh"
 run_logged "$OMARCHY_INSTALL/hardware/apple/fix-suspend-nvme.sh"
 run_logged "$OMARCHY_INSTALL/hardware/apple/fix-t2.sh"
+run_logged "$OMARCHY_INSTALL/hardware/apple/fix-t2-touchpad.sh"
 run_logged "$OMARCHY_INSTALL/hardware/apple/fix-brcmfmac-supplicant.sh"
 
 run_logged "$OMARCHY_INSTALL/hardware/lenovo/fix-yoga-pro7-bass-speakers.sh"
--- a/install/hardware/apple/fix-t2-touchpad.sh
+++ b/install/hardware/apple/fix-t2-touchpad.sh
@@ -1,0 +1,39 @@
+# Mark the internal trackpad on T2 Macs as internal.
+#
+# The T2 bridge exposes the built-in keyboard and trackpad over a virtual USB
+# host controller (t2bce_vhci), so the trackpad carries ID_BUS=usb and
+# /usr/lib/udev/rules.d/65-integration.rules classifies it as external. libinput
+# makes disable-while-typing unavailable on external touchpads, so a palm
+# resting on the pad moves the cursor while typing.
+#
+# Nothing reports this as broken, which is what makes it hard to find. The
+# compositor option is set and reads back as enabled, while libinput has no such
+# setting to apply:
+#
+#   $ hyprctl getoption input:touchpad:disable_while_typing
+#   int: 1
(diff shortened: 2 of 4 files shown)


thooams:
On a MacBookPro16,2 (13" 2020, T2) running `linux-lts-t2`, the built-in trackpad enumerates as `05ac:027e`, not `05ac:0340`. Widening the match (or matching on `ATTRS{phys}=="usb-t2bce_vhci*"` as you tested) would be needed to cover this model too.

For reference, I hand-fixed this locally with:
```
SUBSYSTEM=="input", KERNEL=="event*", ATTRS{idVendor}=="05ac", ATTRS{idProduct}=="027e", ENV{ID_INPUT_TOUCHPAD}=="1", ENV{ID_INPUT_TOUCHPAD_INTEGRATION}="internal"
```
which confirmed `Disable-w-typing` flips to `enabled` afterward — same effect as this PR's hwdb entry.

spuder:
Thank you so much for finding this. The fact that my thumbs keep moving the cursor while typing has been a significant annoyance. So glad to see a PR to fix it.

Issue #7176, opened 2026-08-16
Lock screen retries fingerprint auth every 250ms with no backoff, keeping throttled readers permanently disabled


AdamWorley:
## What happened

The lock screen retries fingerprint authentication in an unbounded hot loop when `pam_fprintd` returns an error immediately, generating ~4 PAM sessions per second for as long as the screen stays locked. On a device that libfprint thermally throttles, this is self-sustaining: the retries keep the reader from ever cooling out of its lockout, so fingerprint unlock can never succeed and the only way in is the password.

In one day on my machine this produced **41,498 fingerprint PAM sessions** and **41,324 fprintd errors**, spread across 192 separate lock episodes.

## Cause

`shell/plugins/lock/Service.qml`:

```qml
onError: function(error) {
  root.fingerprintAuthenticating = false
  if (root.lockRequested && root.fingerprintConfigured) fingerprintRetryTimer.restart()
}

Timer {
  id: fingerprintRetryTimer
  interval: 250
  repeat: false
  onTriggered: root.startFingerprint()
}
```

The retry is a flat 250 ms with no backoff and no cap on consecutive failures. That interval is reasonable when the error is transient (a bad swipe), but `onError` does not distinguish a bad swipe from a device-level failure that will return instantly and identically every time.

Here the device-level failure is libfprint's thermal protection. My reader is a Upek TouchChip/Touchstrip swipe sensor (`147e:2016`), which libfprint 1.94 throttles on duty cycle. Once it trips:

```
fprintd[49676]: Device reported an error during verify: Device disabled to prevent overheating.
```

`pam_fprintd` returns `PAM_AUTHINFO_UNAVAIL` (9) essentially instantly, `onError` fires, the timer re-arms, and 250 ms later it tries again. The device is never idle long enough to cool down, so it stays disabled until the user gives up and types their password.

Log excerpt showing one cycle (they repeat ~4x/second):

```
omarchy-shell[274091]: Starting pam session for user "adam" with config "omarchy-lock-fingerprint" in dir "/etc/pam.d"
omarchy-shell[274091]: Relaying pam message: "Swipe your right index finger across the fingerprint reader" echo: 1 error: 0 responseRequired: 0
omarchy-shell[274091]: Error while authenticating: "Authentication service cannot retrieve authentication info" (code 9)
omarchy-shell[274131]: Starting pam session for user "adam" with config "omarchy-lock-fingerprint" in dir "/etc/pam.d"
[...]

PR #7179, opened 2026-08-16
Stop the lock screen overheating the fingerprint reader

AdamWorley:
The lock screen overheats the fingerprint reader about three minutes after the screen locks, and then never lets it recover.

## What happens

libfprint charges a device for the time it is **armed**, not for the time a finger is on it. From `fpi-device.c`:

```c
if (priv->temp_last_active) {
    alpha = exp (-passed_seconds / priv->temp_hot_seconds);
    priv->temp_current_ratio = alpha * priv->temp_current_ratio + 1 - alpha;
} else {
    alpha = exp (-passed_seconds / priv->temp_cold_seconds);
    priv->temp_current_ratio = alpha * priv->temp_current_ratio;
}
```

| | |
|---|---|
| `DEFAULT_TEMP_HOT_SECONDS` | 180s — time constant while a verify is open |
| `DEFAULT_TEMP_COLD_SECONDS` | 540s — time constant while idle |
| `TEMP_WARM_HOT_THRESH` | 0.731 — device starts refusing here |
| `TEMP_HOT_WARM_THRESH` | 0.5 — the refusal **latches** until the ratio falls back under this |

A verify nobody swipes stays armed for its full ~30s timeout, and `handleFingerprintFinished()` re-arms the moment it returns. So a locked screen holds the reader open essentially 100% of the time. From libfprint's cold start of 0.269 that reaches 0.731 in:

```
-180 * ln((0.731 - 1) / (0.269 - 1)) = 180 seconds
```

Measured here: first arm `18:17:19`, first `Device disabled to prevent overheating` at `18:20:21` — **182 seconds**. Upek Touchstrip (`147e:2016`), which takes libfprint's defaults, as does every driver that doesn't set `temp_hot_seconds` itself. This isn't specific to one sensor.

## Why it doesn't recover

Backing off on errors isn't enough on its own. A refused device returns instantly, so those retries are too cheap to heat it and the ratio drifts down toward 0.5 — at which point the next retry arms *for real*, holds 30s, and pushes it straight back over. It oscillates around the latch threshold indefinitely. The reader here sat refused for **2h20m** and only came back on a `systemctl restart fprintd`.

## The fix

**Don't arm the reader while the display is blanked.** Nobody is there to swipe it, so the entire cost is heat. Waking arms it immediately — which also means the reader is cold and ready when someone walks up, rather than part-way through a backoff.

**Mirror libfprint's ratio and refuse to arm above 0.5.** That's libfprint's own clear-threshold, so the lock screen stops a full band short of the cut-off and never contends with a latched refusal. This bounds the awake-but-not-swiping case that blanking doesn't cover. A scheduled retry waits for whichever is longer, its backoff or the sensor.

The sustainable duty cycle is `3d/(1+2d) < 0.731`, i.e. **d < 47%**. The guard settles at 30%.

## Also here

Three smaller fixes to the same path, each of which stands on its own:

- **Back off on device errors.** A failing verify returned instantly and retried at a flat 250ms, spinning PAM sessions as fast as they could be created — ~2.6 sessions/second for six minutes here. The lock screen was busy enough that a typed password never got checked, so it appeared to reject a correct password. Now exponential, 250ms to a 30s ceiling.
- **Only back off on *fast* errors.** A swipe timeout returns the same PAM code as a broken reader; only elapsed time separates them (~0ms vs ~31s). Backing off on timeouts would leave a healthy reader unarmed. A failed attempt also raises both `onError` and `onCompleted`, so the device error has to win the attempt whichever signal lands first — otherwise the completion resets the streak and the backoff never advances past one step.
- **Re-arm on user presence.** `LockView` already emits `wakeRequested()` on key press, typing, click and cursor movement.

## Testing

`test/shell.d/lock-fingerprint-backoff-test.sh`, 54 assertions. It extracts the real function bodies from `Service.qml` and runs them against a stand-in for the QML root, so it exercises the state machine rather than a paraphrase of it.

The thermal cases drive the real arming policy against an independent implementation of libfprint's model:

```
ok - re-arming on completion overheats the reader, so the guard has something to prevent
ok - the arming ceiling keeps libfprint below its cutoff, peaked at 0.577
ok - the duty cycle stays under what libfprint tolerates, got 30.0%
ok - quick attempts leave the reader responsive, waited 5.9s
ok - a full timeout still re-arms within two minutes, waited 77.1s
ok - blanking releases the reader so it can cool
ok - waking re-arms the reader after a blank rather than bailing on a stale attempt
```

Mutation-tested: raising the ceiling, dropping the cooldown, removing the blanking gate, letting a retry ignore the cooldown, and leaving the in-flight flag set on blank each fail the suite.

Measured on hardware over a five minute lock, before and after:

| | before | after |
|---|---|---|
| reader arms | 767 | 1 |
| `Device disabled to prevent overheating` | 760 | 0 |

Then confirmed end to end across three locks, including one of 3m38s that spans the old 180-second failure point: all three unlocked by fingerprint on first touch, 5 arms total, no overheat events, and the error streak never left 0.

`./test/all`: 2365 assertions pass. The three `omarchy-pkgs` failures are pre-existing on `quattro` and unrelated.

Related: #7176.

### DIFF
--- a/shell/plugins/lock/Service.qml
+++ b/shell/plugins/lock/Service.qml
@@ -31,6 +31,21 @@ Item {
   readonly property int fingerprintRetryMaxMs: 30000
   readonly property int fingerprintFastErrorMs: 2000
   readonly property int fingerprintNudgeCooldownMs: 2000
+  // libfprint models sensor heat as an exponential duty cycle (fpi-device.c):
+  // the ratio rises toward 1 with a 180s constant while the reader is armed and
+  // decays with a 540s one while it is idle, the device is refused above 0.731,
+  // and that refusal latches until the ratio falls back under 0.5. An unswiped
+  // verify holds the reader armed for its whole timeout, so re-arming on
+  // completion is a 100% duty cycle and trips the limit ~3 minutes into a lock.
+  // Track the same ratio and stop short of the latch threshold.
+  readonly property real thermalHeatSeconds: 180
+  readonly property real thermalCoolSeconds: 540
+  readonly property real thermalArmCeiling: 0.5
+  // libfprint assumes a device starts at the top of its COLD band.
+  property real thermalRatio: 0.269
+  property double thermalUpdatedAt: 0
+  property bool thermalArmed: false
+  property bool displayBlanked: false
   property string enteredPassword: ""
   property string pendingPassword: ""
   property string failureMessage: ""
@@ -131,6 +146,10 @@ Item {
     fingerprintErrorStreak = 0
     fingerprintAttemptErrored = false
     fingerprintNudgedAt = 0
+    displayBlanked = false
+    // The heat estimate deliberately survives: it tracks the device, not the
+    // lock session, and unlocking does not cool the sensor.
+    updateThermalRatio(false)
     fingerprintRetryTimer.stop()
     if (passwordPam.active) passwordPam.abort()
     if (fingerprintPam.active) fingerprintPam.abort()
@@ -176,9 +195,14 @@ Item {
   }
 
   function runWake() {
+    var wasBlanked = displayBlanked
+    displayBlanked = false
     if (!wakeProcess.running) wakeProcess.running = true
     if (lockRequested) armBlankTimer()
-    nudgeFingerprint()
+    // Unblanking is the user arriving: the reader has been cooling and should
+    // be armed now, not after whatever backoff was pending when we blanked.
+    if (wasBlanked) startFingerprint()
+    else nudgeFingerprint()
   }
 
   // Backing off is right while nobody is there, but it leaves the reader
@@ -198,6 +222,11 @@ Item {
   }
 
   function runBlank() {
+    displayBlanked = true
+    // Nobody is there to swipe, so let the reader sit idle and cool.
+    if (fingerprintPam.active) fingerprintPam.abort()
+    fingerprintRetryTimer.stop()
+    updateThermalRatio(false)
     if (!blankProcess.running) blankProcess.running = true
   }
 
@@ -234,16 +263,59 @@ Item {
     runWake()
   }
 
+  // Integrates the heat estimate up to now under the previous armed state, then
+  // records the new one. Every arm and disarm has to pass through here or the
+  // estimate drifts away from what libfprint is charging us.
+  function updateThermalRatio(armed) {
+    var now = Date.now()
+
+    if (thermalUpdatedAt > 0) {
+      var passed = (now - thermalUpdatedAt) / 1000
+      if (passed > 0) {
+        if (thermalArmed) {
+          var heat = Math.exp(-passed / thermalHeatSeconds)
+          thermalRatio = heat * thermalRatio + 1 - heat
+        } else {
+          thermalRatio = Math.exp(-passed / thermalCoolSeconds) * thermalRatio
+        }
+      }
+    }
+
+    thermalArmed = armed
+    thermalUpdatedAt = now
+  }
+
+  // Idle time needed to bring the estimate back under the ceiling.
+  function thermalCooldownMs() {
+    if (thermalRatio <= thermalArmCeiling) return 0
+    return Math.ceil(thermalCoolSeconds * Math.log(thermalRatio / thermalArmCeiling) * 1000)
+  }
+
   function startFingerprint() {
     if (!lockRequested || !sessionLock.secure || !fingerprintConfigured) return
     if (fingerprintPam.active || fingerprintAuthenticating) return
+    // A blanked display means nobody is watching, and an armed reader with
+    // nobody to swipe it is pure heat. Waking re-arms it.
+    if (displayBlanked) return
+
+    updateThermalRatio(false)
+    var cooldown = thermalCooldownMs()
+    if (cooldown > 0) {
+      fingerprintRetryTimer.interval = cooldown
+      fingerprintRetryTimer.restart()
+      return
+    }
 
     fingerprintAuthenticating = true
     fingerprintAttemptErrored = false
     fingerprintAttemptStartedAt = Date.now()
     if (!fingerprintPam.start()) {
       fingerprintAuthenticating = false
+      updateThermalRatio(false)
+      return
     }
+
+    updateThermalRatio(true)
   }
 
   function fingerprintRetryDelay(streak) {
@@ -274,7 +346,8 @@ Item {
       fingerprintErrorStreak = 0
     }
 
-    fingerprintRetryTimer.interval = fingerprintRetryDelay(fingerprintErrorStreak)
+    updateThermalRatio(false)
+    fingerprintRetryTimer.interval = Math.max(fingerprintRetryDelay(fingerprintErrorStreak), thermalCooldownMs())
     fingerprintRetryTimer.restart()
   }
 
@@ -282,6 +355,7 @@ Item {
     fingerprintAuthenticating = false
 
     if (!lockRequested) return
+    updateThermalRatio(false)
     if (result === PamResult.Success) {
       fingerprintErrorStreak = 0
       fingerprintAttemptErrored = false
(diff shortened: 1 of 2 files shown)


jcperdomoybarra:
Can confirm this reproduces on different hardware too — ELAN `04f3:0c4b`, via the `libfprint-tod` proprietary driver path (not the Upek TouchChip in the PR description; see #7483 for that setup).

Hit it after a heavy testing session: libfprint's overheating protection kicked in (`Device reported an error during verify: Device disabled to prevent overheating.`), and the lock screen's fixed 250ms retry hammered it — 74 fingerprint PAM sessions in ~7 seconds, all erroring immediately, before I gave up and typed my password. Matches the mechanism described here exactly: a fast device-level error, no backoff, no idle window for the reader to cool down.

The device did recover on its own a few minutes later once nothing was hammering it, so the backoff approach here (rather than a hard disable) looks like the right fix — an exponential backoff would have let it recover mid-lock instead of me having to fall back to password. Happy to test this PR if useful.

jcperdomoybarra:
Following up on my earlier comment — I applied this patch locally on 04f3:0c4b/libfprint-tod (ThinkPad E15 Gen3 AMD) and the backoff itself works well, thanks.

One related thing worth flagging, found after a couple of days of real use: `runBlank()`'s `fingerprintPam.abort()` on screen-blank left `fprintd`'s device claim stuck *permanently* on my hardware once — every subsequent unlock attempt failed instantly with `Device was already claimed`, for well over an hour, and it didn't self-clear. Even `systemctl restart/kill fprintd` as a non-interactive call hung (waiting on an unanswered polkit prompt in my case, so that part's environment-specific), but the underlying claim itself needed a full `fprintd` restart to release — no in-flight libusb transfer ever seemed to finish its cancel.

I don't have visibility into whether this is a general fprintd/libfprint-tod interaction or specific to the ELAN TOD driver's handling of a cancelled bulk transfer. Since `startFingerprint()` already guards on `displayBlanked`, I ended up just not calling `abort()` in `runBlank()` — letting any in-flight verify finish on its own instead of cancelling it — which keeps "don't arm while blanked" for future attempts and avoided the wedge for me. Happy to share the exact diff if it's useful, or this may just be a quirk of my driver stack specifically.

AdamWorley:
Hi @jcperdomoybarra,

Thanks for checking this and sorry for the delay in responding, I found that my initial changes helped a little but still had issues if left long enough and so I have been working on this a little more and should have some more changes to push up once I'm happy with it on my machine

bmcswee:
Independent confirmation of the duty-cycle claim, from a different measurement
path and on a sensor class that never trips the thermal latch — so this lands
as a standalone power argument for "don't arm the reader while the display is
blanked," separate from the overheating one.

**Setup:** ThinkPad T14 Gen 1, Omarchy 4.0.1-1, Hyprland 0.56.2, quickshell
0.3.1-1, kernel 7.1.9-arch1-2, fprintd 1.94.5-2, libfprint 1.94.100-1.
Synaptics Prometheus MIS Touch (`06cb:00bd`) — a **press** sensor, not a swipe
sensor. It never overheats and fingerprint unlock works on first touch.

Rather than counting journal lines, I read the kernel's USB runtime-PM counters
for the reader, which measure armed time directly:

```
$ cat /sys/bus/usb/devices/4-3/power/runtime_active_time      # 101681594 ms
$ cat /sys/bus/usb/devices/4-3/power/runtime_suspended_time   #  54514683 ms
```

| | |
|---|---|
| Uptime | 47.6 h |
| Reader **active** | **28.24 h** |
| Reader suspended | 15.14 h |
| Duty cycle | **65% active** |
| `omarchy-lock-fingerprint` PAM sessions since boot | **3,363** |

The two independent sources agree almost exactly:

```
3363 spawns x 30.05s = 28.07 h armed   (journal)
runtime_active_time  = 28.24 h         (kernel)   -> 99.4% agreement
```

So "a locked screen holds the reader open essentially 100% of the time" is
measurably correct: every hour spent locked is an hour the sensor is armed. On
this machine that was 28 of 47 hours of uptime.

The cadence here is the slow variant AdamWorley flagged above — each verify runs
its full ~30s timeout and returns `PAM_AUTHINFO_UNAVAIL` (9), then
`fingerprintRetryTimer` re-arms 250 ms later, forever:

```
09:16:13  Starting pam session (pid 2288310)
09:16:43  Verification timed out
09:16:44  Starting pam session (pid 2289704)
09:17:14  Verification timed out
          ... until unlock
```

Two things this adds to the case for the blanking guard specifically:

1. **Failure-keyed backoff would not help here.** Nothing is failing. Every
   cycle is a legitimate "nobody touched the sensor" timeout on healthy
   hardware, already naturally spaced at 30s. Only a rule that stops arming —
   the blanked-display guard, or a duty-cycle cap — changes this machine's
   behavior.

2. **There is a cost on readers that never latch.** The device never
   autosuspends for the entire lock, so beyond the sensor's own draw it holds a
   USB device active on a laptop that is otherwise idle overnight. The thermal
   argument doesn't apply to this reader at all, and the guard is still clearly
   worth it.

I have not measured wattage (this machine is on AC), so I'm claiming duty cycle
only, not a power figure.

4A616D6573:
Adding a fourth vendor and the longest non-recovery window in the thread.

**Setup:** Goodix `27c6:533c` (Shenzhen Goodix) on the proprietary TOD path — `libfprint-tod 1.95.2+tod1-1` + `libfprint-goodix-53xc 0.0.6-2`, fprintd 1.94.5-2, Omarchy 4.0.4-1.

Measured over one boot:

| | |
|---|---|
| `Device disabled to prevent overheating` | **4,398** |
| first → last | `09:05:30` → `13:23:48` (**~4h18m**) |
| active minutes | 134 (median **42**/min, max 43, mean 32.8) |
| fprintd cost | 5m 725ms CPU over 2h28m wall, 37.9M peak |

Sustained ~43/min is roughly one PAM session every 1.4s — the fixed 250ms re-arm plus ~1.1s of per-session overhead through PAM/fprintd (the refusal itself returns instantly).

**It never recovered.** The ratio never fell back under the latch across the whole 4h18m; the reader only worked again after I replaced the driver outright. Consistent with your 2h20m.

**What I think is most useful for this PR:** the blanking path already exists in 4.0.4, but it does not stop the arming. `idleBlankTimer` gates only on `authenticatingPassword`, not the combined `authenticating` state, and `runBlank()` just calls `omarchy-brightness-* off` — it never releases the fingerprint PAM. So the entire 4h18m storm ran with nothing stopping the reader from being re-armed. That isolates exactly the gap your "don't arm while blanked" change closes: on this hardware, leaving the arming unstopped is terminal inside ~3 minutes, and nothing on the current lock path ever stops it.

**Vendor coverage:** fourth, after Upek `147e:2016`, ELAN `04f3:0c4b`, and Synaptics `06cb:00bd` — and the second independent confirmation on the `libfprint-tod` path alongside @jcperdomoybarra's ELAN.

pelifix:
Independent confirmation on a second sensor, since "this isn't specific to one sensor" is doing a lot of work in your write-up.

Broadcom CV3 Citadel (`0a5c:5843`, `libfprint-2-tod1-broadcom` 5.15.021.0, `libfprint-tod` 1.95.2, `fprintd` 1.94.5), quattro r2153:

- Locked `12:21:37`, first `Device disabled to prevent overheating` at `12:24:40` — **183 s**, against your predicted 180 s.
- Then a flat cycle: **205 s refusing, 122 s available**, repeating unchanged for the next four hours.
- **31,534** refusals in 3 h 50 min, ~4/s, each one a forked PAM subprocess returning code 9 (`PAM_AUTHINFO_UNAVAIL`).

Two different TOD drivers landing within 3 s of your prediction is about as good as that model gets.

The latch is what a user actually feels: come back inside three minutes and the finger works, come back after fifteen and it's a coin toss. That's the bug report we started from.

Your oscillation argument holds up here too. Gating the arm on display state instead — armed only while a display is awake, live for 45 s after it blanks, re-armed on wake — gives 0 refusals across a 15-minute lock and a first-touch read on return. The reader is rested because nothing touched it while nobody was there. Same shape as #12201.

PR #7182, opened 2026-08-16
Keep UPower's rate when a sysfs power attribute cannot be read

### LINKED ISSUE #6895: Power panel shows 0W discharging when power_now read returns ENODEV

hegge25:
On an HP Spectre x360 14-ea0 the power panel sits at `0W` while the battery visibly drains. `omarchy-battery-status` reports `rate 0W` at the same moment UPower has the right answer:

```
$ omarchy-battery-status --shell | grep rate
rate	0W
$ upower -i $(upower -e | grep BAT) | grep energy-rate
energy-rate:         11.1458 W
```

The cause is the sysfs override that keeps the open panel live:

```bash
if [[ -r $battery_path/power_now ]]; then
  power_rate_raw=$(awk -v microwatts="$(<"$battery_path/power_now")" 'BEGIN { print microwatts / 1000000 }')
```

`[[ -r ]]` proves the attribute exists with read permission — not that reading it works. This firmware advertises `power_now` and then fails the `read()` itself:

```
$ ls -l /sys/class/power_supply/BAT0/power_now
-r--r--r-- 1 root root 4096 Aug 16 22:04 /sys/class/power_supply/BAT0/power_now
$ cat /sys/class/power_supply/BAT0/power_now
cat: /sys/class/power_supply/BAT0/power_now: No such device
```

So the guard passes, `$(<...)` hands back an empty string, awk computes `"" / 1000000 = 0`, and the correct `energy-rate` read out of UPower ten lines earlier is overwritten with zero. The `current_now`/`voltage_now` fallback has the same hole — a failed read there multiplies out to 0W just as quietly.

**The fix:**

- `read_power_attribute` requires the read to succeed *and* the contents to parse as an integer. When it doesn't, `power_rate_raw` keeps the UPower value it already held, which is the correct figure on this hardware.
- Both branches go through it, so the `current_now`/`voltage_now` path is covered too.
- Where the kernel does answer, its instantaneous reading still wins over UPower's laggier one — that is the reason the override exists, and the new test pins it so this fix cannot quietly become "always use UPower".

**Notes for review:**

- Fixes #6895. That report came from an HP Pavilion 15-dk0xxx (`BAT1`); I hit the identical failure on an HP Spectre x360 14-ea0 (`BAT0`), so it is an ACPI firmware class rather than one machine.
- The regression test uses an empty `power_now` in its fixture. An `ENODEV` read and an empty file hand bash the same empty string, so the fixture reproduces precisely what the script sees without needing the broken firmware — and it covers the non-numeric case as well.
- It's a new test file rather than an addition to `battery-status-test.sh`, to stay out of the way of the two open PRs on this script. #6845 rewrites this same block for multi-battery machines and discards invalid reads as a side effect of its metering loop; if that lands first, only the regression test here is still needed, and #6895 explicitly asks for that coverage. #7138 touches the surrounding lines but not this failure. This change is deliberately four lines of logic so it rebases cleanly behind either.

**Verified:**

| | `rate` |
| --- | --- |
| packaged `omarchy-battery-status` | `0W` |
| this branch | `11.1W` |
| `upower -i` `energy-rate` | `11.1458 W` |

The new test fails on `quattro` (`expected: 7.3W / actual: 0W`) and passes with the fix. `./test/shell` runs 178 of 181 files green; the three failures are pre-existing on a clean checkout and unrelated (`config`, `snapper`, `unowned-system-paths` all want an `omarchy-pkgs` checkout that isn't present here). `./test/cli` passes.

### DIFF
--- a/bin/omarchy-battery-status
+++ b/bin/omarchy-battery-status
@@ -45,14 +45,30 @@ power_rate_raw=$(awk '/energy-rate/ { print $2; exit }' <<<"$battery_info")
 native_path=$(awk '/native-path/ { print $2; exit }' <<<"$battery_info")
 battery_path="$power_supply_path/$native_path"
 
+# A readable attribute is not a working one. Some ACPI firmware advertises
+# power_now and then fails the read itself with ENODEV, and `[[ -r ]]` only
+# proves the file exists with read permission, so the empty string that comes
+# back becomes 0 and overwrites the rate UPower had computed correctly. Take a
+# sysfs reading only when it parses as a number.
+read_power_attribute() {
+  local value
+
+  [[ -r $1 ]] || return 1
+  value=$(<"$1") || return 1
+  [[ $value =~ ^-?[0-9]+$ ]] || return 1
+
+  printf '%s' "$value"
+}
+
 # UPower's energy-rate can lag the kernel telemetry by tens of seconds. Use
 # the instantaneous sysfs reading when available so the open panel stays live.
-if [[ -r $battery_path/power_now ]]; then
-  power_rate_raw=$(awk -v microwatts="$(<"$battery_path/power_now")" 'BEGIN { print microwatts / 1000000 }')
-elif [[ -r $battery_path/current_now && -r $battery_path/voltage_now ]]; then
+if microwatts=$(read_power_attribute "$battery_path/power_now"); then
+  power_rate_raw=$(awk -v microwatts="$microwatts" 'BEGIN { print microwatts / 1000000 }')
+elif microamps=$(read_power_attribute "$battery_path/current_now") &&
+  microvolts=$(read_power_attribute "$battery_path/voltage_now"); then
   power_rate_raw=$(awk \
-    -v microamps="$(<"$battery_path/current_now")" \
-    -v microvolts="$(<"$battery_path/voltage_now")" \
+    -v microamps="$microamps" \
+    -v microvolts="$microvolts" \
     'BEGIN { print microamps * microvolts / 1000000000000 }')
 fi
 
--- a/test/shell.d/battery-status-unreadable-rate-test.sh
+++ b/test/shell.d/battery-status-unreadable-rate-test.sh
@@ -1,0 +1,91 @@
+#!/bin/bash
+
+set -euo pipefail
+
+source "$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd)/base-test.sh"
+
+tmp_dir=$(mktemp -d)
+trap 'rm -rf "$tmp_dir"' EXIT
+
+mkdir -p "$tmp_dir/bin"
+cat >"$tmp_dir/bin/upower" <<'STUB'
+#!/bin/bash
+
+if [[ $1 == "-e" ]]; then
+  echo "/org/freedesktop/UPower/devices/battery_BAT0"
+  exit 0
+fi
+
+if [[ $1 == "-i" ]]; then
+  cat <<'INFO'
+  native-path:          BAT0
+  state:                discharging
+  energy:               28.3 Wh
+  energy-full:          56.7 Wh
+  energy-rate:          7.3 W
+  time to empty:        2.5 hours
+  percentage:           51%
+INFO
+  exit 0
+fi
+
+exit 1
+STUB
+chmod +x "$tmp_dir/bin/upower"
+
+# Rebuild BAT0 from scratch per case so a leftover attribute cannot decide the
+# next one. Values arrive as `attribute=contents`; an empty string writes an
+# empty file, which is what bash captures from a read that fails.
+battery_fixture() {
+  local attribute
+
+  rm -rf "$tmp_dir/power"
+  mkdir -p "$tmp_dir/power/BAT0"
+
+  for attribute in "$@"; do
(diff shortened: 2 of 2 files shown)


sorena100:
Confirming the same `power_now`/`ENODEV` failure on another HP Spectre generation:

* Model: `HP Spectre x360 2-in-1 Laptop 14-ef2xxx`
* Board: `8BA2`
* Kernel: `7.1.8-arch1-3`
* Battery: `BAT1`

At the same moment:

```console
$ omarchy-battery-status --shell
percentage      90%
state           discharging
rate            0W
size            55Wh
time            4h 5m
cycles          0
```

UPower has the correct rate:

```console
$ upower -i "$battery" | grep -E 'state|energy-rate|time to|native-path'
native-path:          BAT1
state:                discharging
energy-rate:          11.9872 W
time to empty:        4.1 hours
```

But the sysfs attribute exists and fails when read:

```console
$ cat /sys/class/power_supply/BAT1/power_now
cat: /sys/class/power_supply/BAT1/power_now: No such device
```

`current_now` is unavailable as well, while `voltage_now` is readable.

geirolafs:
Heads up: #6845 now covers this failure too — commit 3c40745 there discards a power_now/current_now read that passes `[[ -r ]]` but comes back empty (ENODEV), falling back to UPower's aggregate rate instead of reporting 0W, with tests for both the power_now and current_now paths. Might save the maintainers merging both.

hegge25:
@geirolafs  this is my first pr, so I'm not sure what you want me to do? Should I take it down?

hegge25:
@geirolafs Thanks for the heads up — you're right, and no action needed from you. I pulled your branch and ran it on the machine from this PR (HP Spectre x360 14-ea0, `BAT0`, `power_now` present but `ENODEV`, no `current_now`): `0W` before, `8.3W` on cbe0705, against UPower's `8.27621 W`. 3c40745 plus the `[[ $microwatts =~ ^-?[0-9]+$ ]] || continue` in the metering loop covers this cleanly, and your dual-battery test already locks both empty-read paths.

So closing this in favour of #6845 rather than leaving two overlapping changes to the same block for the maintainers. I've posted the hardware verification over there.

@sorena100 thanks for the third confirmation — that detail is on #6845 now too, and #6845 is the one to watch for the fix.

PR #7186, opened 2026-08-16
Report combined dual-battery status in the power panel

### LINKED ISSUE #6885: Battery percent in the top bar shows only BAT0
### System details

Thinkpad T480

### What's wrong?

While both batteries are detected not every part of the battery icon module works as expected. 

- static battery shows what seems an overall percentage
- opening the battery panel pop up in the top right corner shows only bat0 (no bat1)
- double clicking on the bar shows the correct percent (both) in the top bar. 

Thanks for your amazing work!

unleashed-nick:
Fixes #6885.

On dual-battery ThinkPads the bar icon and the power panel disagreed: the icon used UPower's composite `DisplayDevice`, while `omarchy-battery-status` (and the panel numbers) took `upower -e | grep BAT | head -n 1` — always BAT0. A long charge into the second pack was also treated as holding because `timeToFull >= 8 hours`.

This keeps one bar icon and makes the numbers match.

- `omarchy-battery-status` aggregates every present `BAT*` pack (combined %, Wh, watts) and emits `pack.N.*` rows.
- The panel hero uses the combined pack. A PACKS section lists each battery when there is more than one.
- Holding only if AC is connected and no pack is taking current. A slow charge no longer counts as a charge limit.
- Notifications that call `omarchy-battery-status` get the same combined line.

Verified on a ThinkPad T480 (BAT0 holding at 90%, BAT1 charging): combined 33% / 96Wh / 26.9W, with both packs listed.

### Tests

- `test/shell.d/battery-status-test.sh` — existing single-pack fixture still passes; added a T480-shaped charging fixture and a both-holding fixture.
- `test/shell.d/power-test.sh` — pack parsing, combined energy fraction, and the 20-hour charge not counting as holding.

`./test/all`: CLI suite green. Shell suite: our files pass. Four other files failed on this machine independently of this change (bar-icon-geometry already fails on clean `quattro` by 0.59px on the monitor glyph; config/snapper/unowned-system-paths need an `omarchy-pkgs` checkout).

### Screenshots

Before/after of the open power panel:

- Before (BAT0 only):
screenshot-2026-08-16_15-43-55

- After (combined + packs): 
screenshot-2026-08-16_15-51-45

### DIFF
--- a/bin/omarchy-battery-status
+++ b/bin/omarchy-battery-status
@@ -7,123 +7,275 @@ shell_output=false
 power_supply_path="${OMARCHY_POWER_SUPPLY_PATH:-/sys/class/power_supply}"
 
 case "${1:-}" in
-  "")
-    ;;
-  --shell)
-    shell_output=true
-    ;;
-  *)
-    echo "Usage: omarchy-battery-status [--shell]" >&2
-    exit 2
-    ;;
+"")
+  ;;
+--shell)
+  shell_output=true
+  ;;
+*)
+  echo "Usage: omarchy-battery-status [--shell]" >&2
+  exit 2
+  ;;
 esac
 
-battery=$(upower -e 2>/dev/null | grep BAT | head -n 1)
-[[ -z $battery ]] && exit 0
+format_rate() {
+  awk -v rate="${1:-0}" 'BEGIN {
+    rounded = sprintf("%.1f", rate)
+    sub(/\.0$/, "", rounded)
+    print rounded
+  }'
+}
 
-battery_info=$(upower -i "$battery")
-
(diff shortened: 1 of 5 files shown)


YuseiRun:
I believe these PRs are also trying to solve this issue
[7138](https://github.com/basecamp/omarchy/pull/7138) and [7295](https://github.com/basecamp/omarchy/pull/7295)

hilather:
Heads-up from a T460s whose internal pack is present but at 0 V / 0 Wh: aggregating every present BAT* (and UPower DisplayDevice) still counts that cell's energy-full, so the combined % is ~49% with a ~10 hour ETA while the live slice is at 99% / 3 min to full. #7917 skips electrically dead packs instead.

PR #7187, opened 2026-08-16
Fix image paste in Kitty and Ghostty

Vtorrealba:
Fix `SUPER+V` image paste in Kitty and Ghostty while preserving text paste behavior.

Clipboard inspection runs outside Hyprland to avoid blocking the compositor, and synthetic input remains targeted to the originating window.

## Tests
- Clipboard binding tests
- Hyprland configuration tests
- CLI suite

### DIFF
--- a/bin/omarchy-clipboard-paste-terminal
+++ b/bin/omarchy-clipboard-paste-terminal
@@ -1,0 +1,44 @@
+#!/bin/bash
+
+# omarchy:summary=Paste into a terminal according to the clipboard content type
+# omarchy:group=clipboard
+# omarchy:hidden=true
+
+pid=${1:-}
+stable_id=${2:-}
+address=${3:-}
+
+[[ $pid =~ ^[0-9]+$ && $stable_id =~ ^[0-9]+$ && $address =~ ^0x[0-9a-fA-F]+$ ]] || exit 1
+
+executable=$(readlink "/proc/$pid/exe" 2>/dev/null) || exit 1
+terminal=${executable##*/}
+terminal=${terminal%" (deleted)"}
+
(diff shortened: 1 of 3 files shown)


Guistoff081:
Thanks for this. It fixes exactly the case that hits AI CLI agents (Claude Code, Codex, opencode) running in the terminal.

Could `foot` be included too? It's Omarchy's default terminal and has the same problem: foot's default `clipboard-paste=Shift+Insert Control+Shift+v XF86Paste` takes both text paste chords as a text-only paste, so image-only clipboard content is silently dropped. Plain `Ctrl+V` isn't bound by foot and reaches the app, which reads the image via `wl-paste`. I noticed the "Foot image paste keeps Shift+Insert" test, so leaving foot out looks deliberate. Was there a specific foot problem you ran into? On Omarchy 4.0.4 / Hyprland 0.56.2 / foot 1.28 with Claude Code (inside tmux), I applied this logic locally with foot included, overriding `SUPER+V` in `~/.config/hypr/bindings.lua`. `SUPER+V` now pastes screenshots as `[Image #N]` and still pastes text normally in bash.

Related foot reports: #11809 (closed as a duplicate of #10526) and the confirmations in #10526.

I opened #13070 as a follow-up. It cherry-picks this PR's commit with your authorship kept and rebases it onto current `quattro`, resolving the conflict with 129c67e2, which moved terminal paste to `Ctrl+Shift+V`. It switches the text fallback to `Ctrl+Shift+V`, adds foot, and extends `clipboard-bindings-test.sh`: foot image/text paste, a hung `wl-paste` fallback, and invalid window arguments. Happy to close it if you'd rather fold the foot change in here.

Issue #7184, opened 2026-08-16
Screen recording fails silently on the external monitor of a hybrid-GPU laptop, and --fullscreen ignores OMARCHY_SCREENRECORD_USE_PORTAL


Literato2:
### System details

Intel Core i5-10500H + Intel UHD (00:02.0) & NVIDIA RTX 3060 Mobile (01:00.0), Omarchy 4.0.0.r1746.gb724f76-1, Hyprland 0.56.2, gpu-screen-recorder 6.0.0-1, nvidia-open-dkms 610.57.04-1, mesa 1:26.1.7-1

### What's wrong?

On a hybrid-GPU laptop where the external monitor is driven by the discrete GPU, `omarchy-capture-screenrecording --fullscreen` fails silently whenever the focused monitor is that external display.

`gpu-screen-recorder`'s KMS backend attaches to a single DRM card and only enumerates the connectors on it. Here the internal panel is on the Intel iGPU (`card2`) and the external monitor is on the NVIDIA dGPU (`card1`), so `gpu-screen-recorder` never sees the external connector:

```
$ gpu-screen-recorder --list-capture-options
eDP-1|1920x1080
region
...
portal
```

```
$ ls /sys/class/drm/card*-*/status
/sys/class/drm/card1-HDMI-A-1/status: connected      # NVIDIA  (01:00.0)
/sys/class/drm/card2-eDP-1/status:    connected      # Intel   (00:02.0)
```

`--fullscreen` passes the focused monitor straight to `-w`, so recording dies immediately:

```
===== 2026-08-15 21:14:58 args:  target: monitor:HDMI-A-1 =====
gsr error: display "HDMI-A-1" not found, expected one of:
  "screen"
  "eDP-1"    (1920x1080+1920+0)
```

Two things make this worse than a plain unsupported-hardware case:

**1. The failure is silent and the script still exits 0.** `gpu-screen-recorder` is backgrounded and its exit status is never checked, so nothing is reported to the user — no notification, no non-zero exit. From the desktop it looks like the screen-recording button/keybinding simply does nothing. Without `OMARCHY_SCREENRECORD_DEBUG=true` there is no way to find out why.

**2. `OMARCHY_SCREENRECORD_USE_PORTAL=true` cannot reach this path.** The portal backend is GPU-agnostic and works correctly here (verified below), but in `omarchy-capture-screenrecording` the `--fullscreen` branch is evaluated *before* the portal branch and ignores the variable:

```bash
[...]

PR #7189, opened 2026-08-16
Report screen recordings that fail to start

Literato2:
`start_screenrecording` already notices when `gpu-screen-recorder` dies before writing the output file — the wait loop breaks on the dead pid and the `kill -0` guard fails — but the failure branch didn't exist. No notification, nothing on screen. A recording that never starts is indistinguishable from a keybinding that does nothing, and the stderr explaining why goes to `/dev/null` unless `OMARCHY_SCREENRECORD_DEBUG=true` is already set, which nobody sets before the first failure.

This adds the missing `else`: notify, and return non-zero, which the caller at the bottom already treats as failure (`start_screenrecording || cleanup_webcam`).

Verified by forcing a start failure — the notification fires with the message pointing at the debug variable.

Independent of #7184 and of the portal-ordering PR; this one is about surfacing *any* start failure, whatever the cause.

One thing I left alone: the script's own exit status still comes from that `|| cleanup_webcam`, so `omarchy capture screenrecording` continues to exit 0 on failure. Happy to change that too if you'd want it.

### DIFF
--- a/bin/omarchy-capture-screenrecording
+++ b/bin/omarchy-capture-screenrecording
@@ -198,6 +198,10 @@ start_screenrecording() {
   if kill -0 $pid 2>/dev/null; then
     echo "$filename" >"$RECORDING_FILE"
     toggle_screenrecording_indicator
+  else
+    omarchy-notification-send -u critical -t 3000 "Screen recording failed to start" \
+      "Set OMARCHY_SCREENRECORD_DEBUG=true and retry to log why."
+    return 1
   fi
 }
 


686f6c61:
The missing else is the right fix. `start_screenrecording` already leaves the wait loop when `gpu-screen-recorder` dies before the output file appears; the `kill -0` guard then fails and used to fall through with nothing on screen. A recording that never starts looked like a dead keybinding.

`omarchy-notification-send` is the helper this file already uses for the other start failures (missing output dir, no webcam, force-kill). Pointing at `OMARCHY_SCREENRECORD_DEBUG=true` is accurate: `LOG_FILE` is `/dev/null` unless that env is set, so stderr is gone unless you already knew to turn it on.

One leftover I'd leave for a follow-up, same as you noted: the caller is still `start_screenrecording || cleanup_webcam`, so `omarchy capture screenrecording` exits 0 after this path. Changing that is a separate, caller-visible behavior change from surfacing the failure.

Read against current `quattro` `bin/omarchy-capture-screenrecording`. Did not force a start failure on hardware here.

PR #7192, opened 2026-08-16
#7135 - Keep playerctl installed for Voxtype users during Quattro upgrade

### LINKED ISSUE #7135: Quattro removes playerctl while default Voxtype config requires it
## System details

AMD Ryzen 5 5600X, NVIDIA RTX 4070, Omarchy 4.0.0-1, omarchy-settings 4.0.0-1, voxtype-bin 0.7.5-1

## What's wrong?

The Quattro upgrade deliberately removes `playerctl`, but Omarchy's shipped Voxtype configuration enables a feature that requires it:

```toml
[audio]
pause_media = true
```

This is reproducible from untouched upstream/package-owned files; it does not depend on local customization:

- Current `quattro` default enables `pause_media`: https://github.com/basecamp/omarchy/blob/quattro/default/voxtype/config.toml#L28-L29
- The Quattro upgrader explicitly retires `playerctl`: https://github.com/basecamp/omarchy/blob/quattro/bin/omarchy-upgrade-to-quattro#L791-L865
- Its fallback removal group also pairs `waybar playerctl`: https://github.com/basecamp/omarchy/blob/quattro/bin/omarchy-upgrade-to-quattro#L877-L903
- Voxtype 0.7.5 describes `--pause-media` as `requires playerctl`.

### Steps to reproduce

1. Have Voxtype installed/configured through Omarchy with `pause_media = true`.
2. Upgrade a legacy Omarchy installation to Quattro.
3. Confirm `playerctl` was removed by the upgrade.
4. Trigger Voxtype recording.

Voxtype logs:

```text
WARN playerctl not found or failed to run: No such file or directory (os error 2)
```

Media pause/resume during dictation no longer works.

### Expected

An Omarchy-owned default should retain its runtime requirement across the Quattro upgrade, or Omarchy should ship a Voxtype version that no longer needs that requirement.

Voxtype already replaced the `playerctl` subprocess with direct MPRIS D-Bus access in https://github.com/peteonrails/voxtype/pull/487. Until that version is shipped, possible fixes are retaining/installing `playerctl` when Voxtype is installed, or disabling `pause_media` by default.

Related but not the same regression: https://github.com/basecamp/omarchy/issues/6029 predates Quattro and was closed before `playerctl` was retired.

josephtuckwell:
Keep playerctl installed for Voxtype users during Quattro upgrade

Fixes #7135

**Problem**

The Quattro upgrader unconditionally retires playerctl, but Omarchy's shipped Voxtype config enables pause_media = true by default, which shells out to playerctl to pause/resume media during dictation. After upgrading, Voxtype users hit:

`WARN playerctl not found or failed to run: No such file or directory (os error 2)`

and media pause/resume during dictation silently stops working. Fresh Voxtype installs had a related gap: omarchy-voxtype-install never pulled in playerctl either, so a brand-new install with the default config was already broken out of the box.

**Fix**

- bin/omarchy-upgrade-to-quattro: in remove_retired_default_packages, skip removing playerctl when voxtype-bin is installed - both in the main batch removal and in the waybar playerctl fallback removal group used when the batch removal can't run as one transaction.
- bin/omarchy-voxtype-install: install playerctl alongside wtype/voxtype-bin so fresh installs aren't affected either.

This is scoped to only retain playerctl when Voxtype is actually present, so it doesn't change behavior for users who don't have Voxtype installed.

**Testing**

- Added assertions to test/shell.d/upgrade-to-quattro-test.sh covering the new guard in both the batch and fallback removal paths.
- Ran the full suite (./test/all): the changed/new tests pass. 4 unrelated pre-existing failures remain (config-test.sh, manifest-entrypoints-test.sh, snapper-test.sh, unowned-system-paths-test.sh) - confirmed these fail identically on main/stashed-out, due to this environment lacking a local omarchy-pkgs checkout, not caused by this change.

**Note**

Voxtype has since replaced its playerctl subprocess call with direct MPRIS D-Bus access (peteonrails/voxtype#487). Once Omarchy ships a Voxtype version built on that, this playerctl dependency (and this guard) can be dropped.

### DIFF
--- a/bin/omarchy-upgrade-to-quattro
+++ b/bin/omarchy-upgrade-to-quattro
@@ -850,10 +850,18 @@ remove_retired_default_packages() {
     zoom
   )
 
+  # Voxtype's Omarchy-shipped default config enables pause_media, which shells
+  # out to playerctl. Removing playerctl out from under an installed Voxtype
+  # breaks media pause/resume during dictation, so keep it when Voxtype is present.
+  if package_installed_exact voxtype-bin; then
+    retired_packages=("${retired_packages[@]/playerctl}")
+  fi
+
   log "Removing packages retired from the Omarchy quattro default install"
   local pkg
   local installed_retired=()
   for pkg in "${retired_packages[@]}"; do
+    [[ -z $pkg ]] && continue
     if package_installed_exact "$pkg"; then
       installed_retired+=("$pkg")
     fi
@@ -877,9 +885,11 @@ remove_retired_default_packages() {
   local group group_pkg required_by_line
   local group_packages=()
   local installed_group=()
+  local waybar_group="waybar playerctl"
+  package_installed_exact voxtype-bin && waybar_group="waybar"
   local fallback_groups=(
     "omarchy-walker walker elephant-all elephant elephant-bluetooth elephant-calc elephant-clipboard elephant-desktopapplications elephant-files elephant-menus elephant-providerlist elephant-runner elephant-symbols elephant-todo elephant-unicode elephant-websearch"
-    "waybar playerctl"
+    "$waybar_group"
     "impala iwd"
     "blueberry bluetui"
     "wiremix pavucontrol"
--- a/bin/omarchy-voxtype-install
+++ b/bin/omarchy-voxtype-install
@@ -8,7 +8,8 @@ set -e
 # Install voxtype and configure it for use.
 
 if gum confirm "Install Voxtype + AI model (~150MB) to enable dictation?"; then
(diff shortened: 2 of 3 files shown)


peteonrails:
This is fixed in Voxtype 1.0.0 upstream. As noted in this PR, it uses a dbus call to pause media instead of using playerctl. Confirmed working on Omarchy Quattro and current in the AUR waiting on the cooling period.

bjarneo:
Automated duplication check: this pull request looks similar to #9244, which covers the same voxtype playerctl install. I keep that one open and close this one to consolidate review.

PR #7196, opened 2026-08-16
Lock the lid and idle the Radeon on 15-inch 2016–2017 MacBook Pros

shawnyeager:
## Summary

On the 15-inch 2016–2017 Radeon MacBooks, closing the lid no longer suspends. That path does not resume, so the machine stays on, locks, idles the GPU, and **disables the internal output** so the 560 stops scanning out.

Independent of the Touch Bar and CS8409 speaker work. Verified on MacBookPro14,3. 13-inch T1 models are Intel-only and are not included.

## Why

Stock lid-suspend is a hard reboot on these boards.

Stay-on with the lid closed still cooked the chassis. amdgpu sat on `3D_FULL_SCREEN` (~13 W at 0% busy). Backlight 0 still scanned out eDP. Hyprland DPMS did too — the CRTC stayed up. Stock `key_press_enables_dpms` / `mouse_move_enables_dpms` then turned scanout back on. Clamshell then force-enabled the internal panel on every `monitorremoved`, which undid the blank.

A `powerprofilesctl set power-saver` overlay on lid close put BCM43602 (`14e4:43ba`) into PCI D3/ASPM. `brcmfmac` does not come back from that.

## What

- Detector is `MacBookPro13,3` / `14,3` only.
- Lid switch locks. systemd refuses suspend and hibernate. The menu hides those rows.
- The discrete GPU stays bound. `POWER_SAVING` lets Polaris drop clocks at idle. This is not a GPU power-off.
- Lid close disables the internal output (`hl.monitor({ disabled = true })`) and persists that in a Hyprland toggle so a reload does not bring the panel back. Key/mouse DPMS-wake is held off. It does **not** set power-saver.
- Clamshell does not force-enable the internal panel while the lid is shut with no external, so a `monitorremoved` event cannot undo the disable.
- Open drops the toggle, restores the panel, brightness, the wake flags, and `omarchy-powerprofiles-set autodetect`.
- BCM43602 stays PCI-awake: udev `power/control=on` and `d3cold_allowed=0`, applied at install and by migration `1788634178`.
- A docked lid close skips the overlay so the external display stays usable.
- Migration seeds AC=`balanced` only if the user has never chosen a profile.
- The NVMe d3cold unit follows `nvme0` instead of `0000:01:00.0` (the Radeon on these machines).

No mux, no rEFInd, no voicing.

Related: #7064, #7140

## Test plan

- [x] `test/shell.d/mbp15-dgpu-test.sh` (detector, idle profile, lid overlay, output disable, wake-flag save/restore, wifi PCI pin, no lid-close power-saver)
- [x] `test/shell.d/lid-close-test.sh` (lock, then clamshell, then 15-inch chill)
- [x] `test/shell.d/monitor-clamshell-scale-test.sh` (clamshell does not re-enable a shut undocked panel)
- [x] MacBookPro14,3: POWER_SAVING idles ~214–320 MHz; lid lock + backlight 0; open restores
- [ ] Live lid close after this commit: confirm eDP is disabled (`hyprctl monitors all`) and GPU power drops, and that a key/mouse does not bring the panel back. Tests cover the helper; that close was not run in this session.
- [ ] MacBookPro14,3: lid close leaves `wlan` up; `power/control` on `14e4:43ba` stays `on`.


makoni:
I have lid-close suspend and resume working on a MacBookPro14,3, so I wanted to put numbers against the premise before this lands. The GPU-idling and clamshell half of this PR stands on its own either way — this is only about "stock lid-suspend is a hard reboot on these boards".

**Six suspend/resume cycles on a MacBookPro14,3**, stock `systemd-logind` lid path, Omarchy 4.0.3, kernel 7.2.3-arch1-3, Radeon Pro 560 + HD 630 over gmux, internal panel on the dGPU (`card1-eDP-1`; i915 has no connected output):

```
slept  0.7 min  ->  resumed in 0.67 s
slept  1.5 min  ->  resumed in 0.68 s
slept  2.0 min  ->  resumed in 0.69 s
slept 12.0 min  ->  resumed in 0.75 s
slept 10.2 min  ->  resumed in 0.76 s
slept  3.0 min  ->  resumed in 0.77 s
```

**One important qualification: the lid does not wake this machine — a keypress does.** Opening the lid leaves it asleep indefinitely (I left it 12 minutes and 10 minutes on two occasions); pressing any key resumes it in well under a second. The lid registers ACPI events but never a wake event — `/sys/kernel/debug/wakeup_sources` shows `PNP0C0D:00` with `event_count=2, wakeup_count=0`. `LID0` is already `*enabled` in `/proc/acpi/wakeup`, enabling `EC` as well changed nothing, and `acpi.ec_no_wakeup` is `N`. This matches discussion #4695, which notes lid-triggered wake is inconsistent across these hardware variants. So: suspend is reachable and resume is fast, but wake is keyboard-driven here.

**What it took, and why I think you are one step away from it.** You already pin the Wi-Fi awake — `power/control=on` and `d3cold_allowed=0` on `14e4:43ba`. That is exactly the treatment that turned out to be load-bearing; the difference is scope. Three things together:

1. `mem_sleep=s2idle`. `deep` never resumes on this board in any configuration I tried — the journal simply ends at `PM: suspend entry (deep)`.
2. `d3cold_allowed=0` on **every** PCI device, not only the BCM43602. Staged `/sys/power/pm_test` runs put the first blocker on the Wi-Fi exactly where you found it, but once that was pinned the same failure moved to the Thunderbolt bridges and the xHCI controllers behind them.
3. `power/control=on` on the `pcieport` bridges. They runtime-suspend into D3cold on their own when nothing is plugged in, and `d3cold_allowed` does not prevent it — by the time the flag is written they are already gone.

**2 and 3 are both required.** I dropped 3, kept 2, rebooted and closed the lid: no resume, journal ends at `PM: suspend entry (s2idle)` with zero D3cold errors logged before it. That was the only failed cycle of the seven.

**Two things that cost me hours, in case they save you some.**

`/sys/power/pm_test` is excellent for finding *which device* blocks the transition — it stages the suspend and auto-resumes after five seconds, so the journal survives, which a genuinely failed resume never does. It is **not** valid for comparing configurations: a device left stuck in D3cold by one trial stays stuck and contaminates the next. My staged runs said measure 2 alone was sufficient; the real lid close said otherwise, and the real lid close was right.

Do not unload `thunderbolt` around sleep. It unloads cleanly, but the controllers never re-probe — `timeout resetting host router`, `cannot enable interrupts`, `probe failed with error -22` — and Thunderbolt is dead until reboot. The `tb_cfg_read: -108` warnings on resume are noisy (about 14 per wake, ~1800 journal lines) but harmless; resume is 0.7 s with them present.

**Caveats.** One machine, one session. `s2idle` is not S3: the fans keep spinning with the lid shut, so your thermal argument for idling Polaris may well still apply *during* suspend — I have not measured chassis temperature over a long sleep, and that seems worth checking before treating these as alternatives. The T1 Touch Bar also goes dark across resume and needs its stack reloaded from a `system-sleep` hook, but that is #10825 territory rather than this PR's.

None of this touches the docked/clamshell half of your work. But if suspend is reachable on these boards, then `systemd` refusing suspend and hibernate outright, and hiding those menu rows, is a heavy hammer — and I wanted the measurements on the record before it lands.

PR #7205, opened 2026-08-17
Fix audio panel selection for sinks with multiple ports

leomoon:
Expose usable audio sink ports as separate output targets, allowing the audio panel to switch between physical ports while preserving existing default-sink selection and application stream movement.

Previously, hardware exposing multiple analog ports through a single PipeWire sink appeared as only one output option. This made it impossible to select the rear line-out and front headphone jack independently.

With this fix, the audio panel shows both the rear 3.5 mm line-out and front 3.5 mm headphone port when PipeWire reports them as available. Selecting either entry activates the corresponding port and moves active application audio to it.

image

### DIFF
--- a/bin/omarchy-audio-output-set-default
+++ b/bin/omarchy-audio-output-set-default
@@ -1,15 +1,28 @@
 #!/bin/bash
 
 # omarchy:summary=Set the default audio output and move active streams
-# omarchy:args=<node-id> <sink-name>
-# omarchy:examples=omarchy audio output set default 42 alsa_output.pci-0000_00_1f.3.analog-stereo
+# omarchy:args=<node-id> <sink-name> [port-name]
+# omarchy:examples=omarchy audio output set default 42 alsa_output.pci-0000_00_1f.3.analog-stereo analog-output-headphones
 
 node_id=${1:-}
 sink_name=${2:-}
+port_name=${3:-}
 
 if [[ -z $node_id || -z $sink_name ]]; then
-  echo "Usage: omarchy-audio-output-set-default <node-id> <sink-name>" >&2
+  echo "Usage: omarchy-audio-output-set-default <node-id> <sink-name> [port-name]" >&2
   exit 1
+fi
+
+if [[ -n $port_name ]]; then
+  sink_json=$(timeout 2 pactl --format=json list sinks)
+  if ! jq -e --arg sink "$sink_name" --arg port "$port_name" \
+    'any(.[]; .name == $sink and any(.ports[]?; .name == $port))' \
+    <<<"$sink_json" >/dev/null; then
+    echo "Audio output port disappeared: $sink_name/$port_name" >&2
+    exit 1
+  fi
+
+  timeout 2 pactl set-sink-port "$sink_name" "$port_name"
 fi
 
 timeout 2 wpctl set-default "$node_id" 2>/dev/null || true
--- a/bin/omarchy-audio-output-targets
+++ b/bin/omarchy-audio-output-targets
@@ -1,0 +1,67 @@
+#!/bin/bash
+
+# omarchy:summary=List selectable audio output sink ports as JSON
+# omarchy:group=audio
+
+set -euo pipefail
+
+# A speaker tuning is a virtual sink in front of the real speakers. Exposing
+# the fronted physical sink would let the panel bypass the processing chain.
+fronted=$(omarchy-audio-tuning fronted-sink 2>/dev/null || true)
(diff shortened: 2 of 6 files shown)


ahmadtv:
This is worth having — the same gap shows up on the 2017 iMac (iMac18,3, Cirrus CS8409 + CS42L83). PipeWire exposes one analog sink with several ports (internal speakers, headphone jack), so the audio panel can only offer one output and there is no way to pick the jack from the UI while the speakers are the active port.

I had written the same feature independently and opened it as #10985; that was just closed as a duplicate of this one, which is fine — this is the older PR. Two things from my implementation that might be worth checking here, since they were the fiddly parts on hardware where a jack comes and goes:

- **Availability, not just presence.** Ports report `available` / `not available` as the jack is plugged and unplugged. Listing only the ports PipeWire currently reports as available keeps dead entries out of the panel, and the list has to refresh on the port-availability change rather than only when sinks appear or disappear.
- **Follow the hardware's own switch.** When a headset is plugged in, WirePlumber moves the active port by itself. If the panel caches the selected port it will disagree with the hardware a moment later; reading the active port back after a change avoids the panel and the jack fighting each other.

Happy to test this branch on the iMac and report back, including plug and unplug cycles on the headset jack — that is where this hardware usually breaks.

leomoon:
The decision for merging is not on me, but maybe @ahmadtv try it out and if it works, we'll have a stronger case for merging.

igouss:
Nice fix! The same issue applies to inputs. `SourceRow` still lists one row per PipeWire source node, and `omarchy-audio-input-set-default` never calls `pactl set-source-port`. So a card whose one source has several ports can't pick between them from the panel, e.g. a desktop Intel HDA:

```
alsa_input.pci-0000_00_1f.3.analog-stereo  "Built-in Audio Analog Stereo"
  analog-input-front-mic   Front Microphone
  analog-input-rear-mic    Rear Microphone
  analog-input-linein      Line In
```

This shows up as a single "Built-in Audio Analog Stereo" row. With a front mic and a rear line-in both plugged in, there's no way to pick between them.

Most of this PR would carry over:
- `omarchy-audio-input-targets`: same as the output version, but using `pactl --format=json list sources` and skipping monitor sources
- `omarchy-audio-input-set-default`: an optional `[port-name]` argument, the same check that the port still exists, then `pactl set-source-port`
- `Model.js`: the target parse/join/fallback logic works for sources too; only the kind/glyph rules change (mic / line-in / headset)
- `Panel.qml`: an input targets snapshot for `SourceRow`

Would you want to cover inputs here, or should it be a follow-up PR?

ahmadtv:
Tested the branch on the iMac18,3 (Cirrus CS8409 + CS42L83, one analog sink with a speaker port and a combo headphone jack). I ran `bin/omarchy-audio-output-targets` and `bin/omarchy-audio-output-set-default` from the checkout directly, so this is the backend only, not the panel.

**Headset unplugged — correct:**

```
$ bin/omarchy-audio-output-targets
[ { "portName": "analog-output-speaker", "portDescription": "Speakers",
    "kind": "speaker", "active": true } ]
```

PipeWire reports `analog-output-headphones` as `not available`, and it is rightly left out.

**Headset plugged in — only one target, and the speakers disappear:**

```
$ pactl list sinks | grep -A2 'Ports:'
  analog-output-speaker: Speakers (priority: 10000, not available)
  analog-output-headphones: Headphones (priority: 9900, available)

$ bin/omarchy-audio-output-targets
[ { "portName": "analog-output-headphones", "portDescription": "Headphones",
    "kind": "headphones", "active": true } ]
```

On this hardware ALSA flips the speaker port to `not available` as soon as something is in the jack, so the panel would offer exactly one output and there is no way to play through the internal speakers while a headset is connected. macOS on the same machine lists both and lets you choose.

**And the port that is filtered out still works, which leaves the panel with no active entry:**

```
$ pactl set-sink-port <sink> analog-output-speaker     # headset still plugged in
$ pactl list sinks | grep 'Active Port'
  Active Port: analog-output-speaker                   # accepted, audio plays on the speakers

$ bin/omarchy-audio-output-targets
[ { "portDescription": "Headphones", "kind": "headphones", "active": false } ]
```

So after that switch the list contains only Headphones, marked inactive, while the real output is the speakers. Whatever is actually playing is not represented in the panel at all.

Suggestion: treat `not available` as a hint for ordering and default selection rather than a filter — list every port of the sink, mark the one that is active, and let the user pick an unavailable-but-usable one (which is what this codec's speaker port is when a jack is inserted). Keeping the filter but always including the active port would fix the third case, though not the second.

Happy to retest once you have changed it, and to check the panel itself rather than just the helpers.

PR #7208, opened 2026-08-17
Restore batch selection in management menus

ssupt:
## Summary

Quattro moved several management flows to the native Omarchy shell selector, which only returned one item at a time. This removed batch removal for web apps and TUIs and required reopening similar management menus for every operation.

This change adds opt-in multi-selection to the native selector.

- Press Tab to pin or unpin the highlighted row.
- Keep pinned rows selected while filtering.
- Show pinned rows with a persistent fill, accent marker, checkmark, and live selected count.
- Return pinned values as newline-delimited output in their original order.
- Preserve single-selection behavior when nothing is pinned.
- Keep multi-selection hidden unless explicitly enabled.

Batch selection is enabled for web app, TUI, and extra-theme removal, plus plugin enablement, disablement, and removal.

Plugin cloning and selectors requiring exactly one value remain single-select. Plugin removal preserves the existing terminal confirmation and backup flow for each selected plugin.

## Screenshot
screenshot-2026-08-17_02-09-52

### DIFF
--- a/bin/omarchy-menu-plugin
+++ b/bin/omarchy-menu-plugin
@@ -30,17 +30,36 @@ rows=$(jq -r --arg icon "$PLUGIN_ICON" \
    | \$icon + \"\\t\" + .name + \"\\t\" + .id" <<<"$plugins")
 [[ -n $rows ]] || { omarchy-notification-send "No plugin to ${1}"; exit 0; }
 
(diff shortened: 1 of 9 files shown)


bjarneo:
Automated duplication check: this pull request looks similar to #11057, which covers the same menu multi-select. I keep that one open and close this one to consolidate review.

PR #7210, opened 2026-08-17
Suspend automatically after idle timeout

vander00:
Add a configurable idle suspend deadline, default is 900 seconds.
I have made it myself, the docs are written by AI. I've read them all, they must be correct.

Refers to #6931.

### DIFF
--- a/config/omarchy/shell.json
+++ b/config/omarchy/shell.json
@@ -2,7 +2,8 @@
   "version": 1,
   "idle": {
     "screensaver": 150,
-    "lock": 300
+    "lock": 300,
+    "suspend": 900
   },
   "bar": {
     "position": "top",
--- a/default/agents/skills/omarchy/plugins.md
+++ b/default/agents/skills/omarchy/plugins.md
@@ -13,7 +13,8 @@ $OMARCHY_PATH/config/omarchy/shell.json  # Canonical defaults
 ```
 
 The shell hot-reloads `shell.json` on save — no restart needed for layout
-changes. `idle.screensaver` and `idle.lock` are seconds since user idle began.
+changes. `idle.screensaver`, `idle.lock`, and `idle.suspend` are seconds since
+user idle began.
 
 **Commands:** `omarchy restart shell`, `omarchy refresh shell`
 
@@ -45,8 +46,13 @@ Saving a file anywhere under `~/.config/omarchy/plugins/` reloads plugin code
 automatically. If a change somehow fails to apply, force a reload with
 `omarchy-shell shell rescanPlugins`.
 
-## Idle and Lock
+## Idle, Lock, and Suspend
 
-Set `idle.screensaver` and `idle.lock` in `~/.config/omarchy/shell.json`,
-in seconds since user idle began. Example: "lock after ten minutes" means
-setting `idle.lock` to `600`.
+Set `idle.screensaver`, `idle.lock`, and `idle.suspend` in
+`~/.config/omarchy/shell.json`, in seconds since user idle began. Example:
+"lock after ten minutes and suspend after thirty" means setting `idle.lock`
+to `600` and `idle.suspend` to `1800`.
+
+Stay Awake (`omarchy toggle idle`) disables all three idle actions. `omarchy
+toggle suspend` only shows or hides the manual Suspend action in the system
+menu; it does not change the automatic idle deadline.
--- a/docs/omarchy-shell.md
+++ b/docs/omarchy-shell.md
@@ -136,7 +136,8 @@ string on a miss.
   "version": 1,
   "idle": {
     "screensaver": 150,
-    "lock": 300
+    "lock": 300,
+    "suspend": 900
   },
   "bar": {
     "id": "omarchy.bar",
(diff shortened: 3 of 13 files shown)


mabster:
(Full disclosure - OpenCode wrote this comment!)

Hi vander00 — nice work on this. The docs, migration, and migration test are thorough. I've been working on the same feature locally and wanted to flag two edge cases I hit during testing.

### 1. No lock guard — `suspendSystem` can sleep an unlocked session

Their `suspendSystem()` runs bare `systemctl suspend`:

```qml
runProcess(suspendProcess, "suspend", "systemctl suspend")
```

When the suspend timer fires mid-lock (the normal case), the screen is already locked, so this works. But if something goes wrong — the lock process is slow, the user is between screens, etc. — there is nothing preventing a suspend on an unlocked session.

The fix is a guard in the bash command:

```bash
[[ $(omarchy-shell lock isLocked 2>/dev/null) == "true" ]] && systemctl suspend
```

The lock-before-suspend path (`omarchy-sleep-lock.service`) secures the screen on `PrepareForSleep`, but the guard removes the race window entirely.

### 2. `suspend <= lock` suspends before lock fires

When `suspend` is less than or equal to `lock` (e.g. `"suspend": 10, "lock": 300`), `firstIdleTimeoutSeconds` becomes the `suspend` value, and `suspendDelaySeconds` becomes 0. This triggers `suspendSystem()` immediately at cycle start — before the lock fires, on an unlocked session.

Tested live (suspend=10, lock=20): the cycle started, screensaver launched, and `systemctl suspend` ran immediately at cycle start — 20 seconds before the lock would have fired.

The fix: only arm the `suspendTimer` when `suspend > lock` (it fires mid-lock at the correct `idle+suspend` time, the screen is already locked). For `suspend <= lock`, add a backstop on `lockProcess.onExited` that suspends at lock completion:

```qml
function suspendAtLockIfDue() {
  if (root.suspendTimeoutSeconds < 0) return
  if (root.suspendTimeoutSeconds > root.lockTimeoutSeconds) return
  if (!root.idleEnabled) return
  root.suspendSystem("suspend-at-lock")
}
```

Then in `startIdleCycle`:

```qml
if (root.suspendTimeoutSeconds > root.lockTimeoutSeconds) {
  suspendTimer.restart()
}
```

And in `lockProcess.onExited`:

```qml
onExited: function(exitCode, exitStatus) {
  root.logEvent("process-exit", "lock exitCode=" + exitCode + " status=" + exitStatus)
  root.suspendAtLockIfDue()
}
```

This also means `lockSystem` should keep clearing `idledThisCycle` as the original code does — the backstop gates on `idleEnabled` only (the `isLocked` guard in `suspendSystem` protects an unlocked session; `lockSystem` is only reachable from an active cycle anyway).

### 3. Minor: `null` trap in config

`"suspend": null` in `shell.json` → `Number(null) = 0` → immediate suspend. The existing `secondsFromConfig` function treats this as a valid zero. A type check at the top of `secondsFromConfig` rejects non-number/non-numeric-string values and falls back to the default, which also hardens the `screensaver`/`lock` paths:

```js
  var n
  if (typeof value === "number") n = value
  else if (typeof value === "string" && value.trim() !== "") n = Number(value)
  else return fallback
  return Math.floor(n)
}
```

---

Happy to contribute these fixes to your branch if useful. I've been running the hardened version locally for a few days and it's solid — the `suspend <= lock` case was the one that took the most testing to get right.

vander00:
> (Full disclosure - OpenCode wrote this comment!)
> 
> Hi vander00 — nice work on this. The docs, migration, and migration test are thorough. I've been working on the same feature locally and wanted to flag two edge cases I hit during testing.
> ### 1. No lock guard — `suspendSystem` can sleep an unlocked session
> 
> Their `suspendSystem()` runs bare `systemctl suspend`:
> 
> ```qml
> runProcess(suspendProcess, "suspend", "systemctl suspend")
> ```
> 
> When the suspend timer fires mid-lock (the normal case), the screen is already locked, so this works. But if something goes wrong — the lock process is slow, the user is between screens, etc. — there is nothing preventing a suspend on an unlocked session.
> 
> The fix is a guard in the bash command:
> 
> ```shell
> [[ $(omarchy-shell lock isLocked 2>/dev/null) == "true" ]] && systemctl suspend
> ```
> 
> The lock-before-suspend path (`omarchy-sleep-lock.service`) secures the screen on `PrepareForSleep`, but the guard removes the race window entirely.
> ### 2. `suspend <= lock` suspends before lock fires
> 
> When `suspend` is less than or equal to `lock` (e.g. `"suspend": 10, "lock": 300`), `firstIdleTimeoutSeconds` becomes the `suspend` value, and `suspendDelaySeconds` becomes 0. This triggers `suspendSystem()` immediately at cycle start — before the lock fires, on an unlocked session.
> 
> Tested live (suspend=10, lock=20): the cycle started, screensaver launched, and `systemctl suspend` ran immediately at cycle start — 20 seconds before the lock would have fired.
> 
> The fix: only arm the `suspendTimer` when `suspend > lock` (it fires mid-lock at the correct `idle+suspend` time, the screen is already locked). For `suspend <= lock`, add a backstop on `lockProcess.onExited` that suspends at lock completion:
> 
> ```qml
> function suspendAtLockIfDue() {
>   if (root.suspendTimeoutSeconds < 0) return
>   if (root.suspendTimeoutSeconds > root.lockTimeoutSeconds) return
>   if (!root.idleEnabled) return
>   root.suspendSystem("suspend-at-lock")
> }
> ```
> 
> Then in `startIdleCycle`:
> 
> ```qml
> if (root.suspendTimeoutSeconds > root.lockTimeoutSeconds) {
>   suspendTimer.restart()
> }
> ```
> 
> And in `lockProcess.onExited`:
> 
> ```qml
> onExited: function(exitCode, exitStatus) {
>   root.logEvent("process-exit", "lock exitCode=" + exitCode + " status=" + exitStatus)
>   root.suspendAtLockIfDue()
> }
> ```
> 
> This also means `lockSystem` should keep clearing `idledThisCycle` as the original code does — the backstop gates on `idleEnabled` only (the `isLocked` guard in `suspendSystem` protects an unlocked session; `lockSystem` is only reachable from an active cycle anyway).
> ### 3. Minor: `null` trap in config
> 
> `"suspend": null` in `shell.json` → `Number(null) = 0` → immediate suspend. The existing `secondsFromConfig` function treats this as a valid zero. A type check at the top of `secondsFromConfig` rejects non-number/non-numeric-string values and falls back to the default, which also hardens the `screensaver`/`lock` paths:
> 
> ```js
> function secondsFromConfig(value, fallback) {
>   var n
>   if (typeof value === "number") n = value
>   else if (typeof value === "string" && value.trim() !== "") n = Number(value)
>   else return fallback
>   if (!isFinite(n) || n < 0) return fallback
>   return Math.floor(n)
> }
> ```
> 
> Happy to contribute these fixes to your branch if useful. I've been running the hardened version locally for a few days and it's solid — the `suspend <= lock` case was the one that took the most testing to get right.

Hello.
I am very pleased for the review. I am also very sorry that I couldn't reply earlier because of my business.
I have checked the lock thing. It must lock every time the 'systemctl suspends' is executed. I do not remember how it works exactly but it looks like some omarchy thing. The default shell's system menu uses the 'systemctl suspend' as well.

bjarneo:
Automated duplication check: this pull request looks similar to #11686, which covers the same idle suspend timeout. I keep that one open and close this one to consolidate review.

PR #7212, opened 2026-08-17
fix(clock): use system locale for calendar weekday labels

### LINKED ISSUE #7211: Calendar weekday labels hardcoded to English instead of using system locale
## Problem

The calendar popup weekday headers (MO, TU, WE, TH, FR, SA, SU) are hardcoded to `Qt.locale("en_US")` in `Panel.qml`, ignoring the user's system locale.

For users with non-English locales (e.g. `es_AR.UTF-8`), this creates a jarring mix: the bar clock and month names render in the system language, but the weekday headers are always English.

## Current behavior

```qml
// Panel.qml line 70
readonly property var labelLocale: Qt.locale("en_US")
```

## Expected behavior

Use `Qt.locale()` (no argument) so Qt resolves the locale from the system environment, matching the behavior of `Qt.formatDate` and `Qt.formatDateTime` used elsewhere in the same widget.

## Why this matters

The Omarchy interface already respects the system locale for time formatting and month names. The weekday headers being locked to English is inconsistent with the rest of the UI and confusing for non-English users.

markbus-ai:
Closes #7211

## Summary

- Use `Qt.locale()` instead of `Qt.locale("en_US")` for calendar weekday labels
- Weekday names (MO, TU, WE, etc.) now follow the system locale
- Consistent with how month names and time formats are already rendered

## Changes

| File | Change |
|------|--------|
| `shell/plugins/panels/clock/Panel.qml` | Changed `labelLocale` from hardcoded `en_US` to system locale |

## Test Plan

- [x] Verified change compiles without errors
- [x] Manually tested the calendar popup shows localized weekday names
- [x] Confirmed week start toggle still works correctly

### DIFF
--- a/shell/plugins/panels/clock/Panel.qml
+++ b/shell/plugins/panels/clock/Panel.qml
@@ -64,11 +64,21 @@ Panel {
   // convention. Clicking the grid's "W" heading writes the choice back to
   // shell.json.
   readonly property int weekStart: Model.normalizedWeekStart(setting("weekStartDay", null), Qt.locale().firstDayOfWeek)
-  // The interface is English throughout, so day names are not taken from the
-  // system locale. Where the week starts still is: that is a regional
-  // convention rather than a translation, and it stays overridable above.
-  readonly property var labelLocale: Qt.locale("en_US")
-  readonly property string nextWeekStartLabel: labelLocale.dayName(Model.toggledWeekStart(weekStart), Locale.LongFormat)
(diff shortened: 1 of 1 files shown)


markbus-ai:
Closing per maintainer direction: dhh intentionally hardcoded English day names in #6988 as part of a consistent English-only shell design. A proper i18n system is a separate future project. Will handle this via a local post-update hook instead.

PR #7224, opened 2026-08-17
Refresh monitor brightness state promptly

iccodes:
## What changed

Refresh the open monitor panel brightness state every 250 ms instead of every 5 seconds.

## Why

Brightness keys change the display outside the panel, while the panel previously waited for its 5-second poll before updating the slider.

## Impact

The open brightness slider now reflects keyboard brightness changes promptly. Polling remains limited to while the panel is open.

## Validation

`git diff --check` passed.


iccodes:
Superseded by a replacement PR after reviewing default plugin refresh cadences.

PR #7241, opened 2026-08-17
Inset BorderSurface strokes a device pixel to survive clip edges

gsamokovarov:
I was extending the monitor panel to support multi-monitor arrangement:

screenshot-2026-08-17_10-08-01

Then I noticed that I have introduced a scroll in the panel and the left edges of the controls have been cut off with a single pixel:

screenshot-2026-08-16_17-36-41
screenshot-2026-08-16_13-24-18

At first I fixed it at my component-level but soon I realized this is a more general edge case. When a stroke flush against an ancestor's clip edge — like a full-width CursorSurface row or the first scale pill at x=0 inside a panel ScrollView — disappears entirely on fractionally scaled outputs, where the shell renders at the ceiled integer scale and the compositor downscales. Drawing every stroke one device pixel inside the surface restores it at full strength, and is imperceptible where the bug cannot occur.

If you think that this is not the right place to fix this, by all means, ignore this change.

---

As an aside, do you think the native monitors panel can benefit by monitor arrangements? I've quickly built it it for myself and it feels like the better place to have it with the quickshell utilities, rather than using TUIs like hyprmon.

https://github.com/user-attachments/assets/70ab7fdd-d68e-40a6-9916-31d8d841e5ff

Here is my initial config, which is a fork of the builtin monitor panel: https://github.com/gsamokovarov/gazarchy/tree/main/.config/omarchy/plugins/genadi.monitor.

### DIFF
--- a/shell/Ui/BorderSurface.qml
+++ b/shell/Ui/BorderSurface.qml
@@ -1,9 +1,13 @@
 import QtQuick
+import QtQuick.Window
 import qs.Commons
 
-// Rectangle-compatible surface with Omarchy border specs. Uses native
-// Rectangle.border for cheap flat/uniform borders and BorderOverlay for
-// gradients or per-side widths.
+// Rectangle-compatible surface with Omarchy border specs. Flat/uniform
+// borders draw with a cheap nested Rectangle stroke and gradients or
+// per-side widths with BorderOverlay. Both are inset a device pixel so
+// strokes flush against an ancestor's clip edge survive fractionally
+// scaled outputs; setting border.* on this item directly draws an
+// extra un-inset stroke alongside the spec-driven one.
 Rectangle {
   id: root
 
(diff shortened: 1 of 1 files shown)


esasse:
Confirmed on other hardware: Dell AW3423DWF, 3440x1440@164.90, monitor `scale = 1.25`, `bitdepth = 10`, Hyprland 0.56.2, Omarchy 4.0.1-1.

A/B on one `quattro` checkout at `0ae1694`, run as `quickshell -n -p <checkout>/shell`, one instance at a time. Only variable: `shell/Ui/BorderSurface.qml` — clean tree (byte-identical to the packaged 4.0.1 file) vs. this PR's version.

The first `SCALE` pill's left stroke is absent before and present after. No new QML warnings, and no doubled or misaligned strokes elsewhere in the panel.

emindashdamirov11-lang:
This fixes the issue I hit on the monitor panel (#10085 / duplicate #12498): with a fractional output scale (eDP-1 = 1.25) and shell text base-size 14, the first (1x) scale pill's left border disappears; smaller text sizes make it render. Insetting the BorderSurface stroke by a device pixel looks like the right general fix — it also covers tooltips and panels, not just the scale pill. Reproduced on Omarchy 4.0.4-1 / Hyprland 0.56.2. Would love to see this merged.

davidszp:
Tested this PR against the Super+Space menu. It fixes a case not mentioned here yet: the highlighted row lost its **left** border, while top, right and bottom drew fine. With the PR, the left border draws fine:
2-before-after-closeup

- Two 3840×2160 monitors, both at `scale = 1.25`, RTX 5060 Ti
- Omarchy 4.0.4-1, Quickshell 0.3.1-1, Hyprland 0.56.2-2
- Theme: Inkypinky, whose `[menu]` section sets `selected-border-width = 1`

The PR's diff applied cleanly to 4.0.4. I ran the patched copy with `quickshell -n -p <copy>/shell`, one instance at a time, and sampled the screen: without the patch there is no border pixel on the row's left edge, and with it the pink border is back.

The row is the `BorderSurface` delegate in `plugins/menu/Menu.qml`, a full-width row inside a `ListView` with `clip: true`. Most themes never show it, because the generated `shell.toml` gives menu rows no selected border (#8778). So that's three places now: the menu, the monitor panel (#10085) and tooltips (#10084).

PR #7243, opened 2026-08-17
Add monitorOnly option to workspaces bar widget

sgruendel:
## Problem

On multi-monitor setups, the workspaces bar widget lists workspaces 1-5 plus every active workspace up to 10 on **every** monitor's bar — even when the user has bound specific workspaces to specific monitors with persistent workspace rules in `monitors.lua`, e.g.:

```lua
hl.workspace_rule({ workspace = "2", monitor = "desc:Dell Inc. ...", persistent = true })
```

Hyprland honors those bindings, but the bar doesn't reflect them, so both bars show 1-10.

## Solution

A new opt-in `monitorOnly` setting on the `omarchy.workspaces` widget. When enabled in `~/.config/omarchy/shell.json`:

```json
{ "id": "omarchy.workspaces", "monitorOnly": true }
```

each bar surface lists only the workspaces Hyprland has bound to the monitor that bar sits on (resolved via the widget's `QsWindow` screen, the same mechanism the bar itself uses for per-monitor panel routing).

Defaults to `false`, so existing setups are unchanged. Workspaces that end up on another monitor (e.g. after a display disconnects) simply appear in that monitor's bar instead.

## Verification

Verified on a two-monitor setup (DP-9: workspaces 1,2,4,6,8,10 / DP-10: workspaces 3,5,7,9 via persistent workspace rules):

- `monitorOnly: true` — DP-9 bar shows `1 2 4 6 8 0`, DP-10 bar shows `3 5 7 9` (screenshots attached below)
- `monitorOnly` unset/off — both bars show `1 2 3 4 5 6 7 8 9 0`, identical to current behavior

`./test/shell`: all widget-related tests pass. Four unrelated failures reproduce on a clean `quattro` checkout in my environment: three need an `omarchy-pkgs` checkout, and `runtime-smoke-test.sh`'s IPC-handler-count assertion is flaky (fails identically with and without this change).


sgruendel:
before on both monitors:
pr-7243-monitorOnly-off

after left / right monitor:
pr-7243-monitorOnly-on-DP-9
pr-7243-monitorOnly-on-DP-10

PR #7252, opened 2026-08-17
Add input settings panel

dhh:
Adds a focused Input panel to the default right-side bar and routes Setup > Input to it.

The panel covers the high-value settings that previously required editing input.lua: primary and secondary keyboard layouts, Compose and Super key placement, Num Lock, pointer speed, natural scrolling, touchpad clicking, and touchpad suppression while typing. Advanced configuration remains one click away.

Panel-owned settings persist in a generated Hyprland state override, leaving hand-maintained input.lua files untouched. An idempotent migration adds the panel to existing bars after Network.

### Input panel

Input panel

### Searchable layout picker

Searchable layout picker

## Verification

- ./test/cli
- Focused Input panel, migration, menu, config, and Hyprland tests
- qmllint -I shell shell/plugins/panels/input/Panel.qml
- Live visual verification and keyboard navigation

---

_This spike was made by GPT-5.6 on behalf of dhh._


mikefsm:
This panel looks great and addresses a real gap in post-install keyboard management.

One use case to consider before merging is users with more than two layouts. I regularly switch between: `us,il,gr`

From the current implementation, `initial_status` reads only `primary` and `alternate`, while the generated override writes those two layouts and loads after the user's `input.lua`. It appears that opening the panel and changing a setting could therefore silently drop the third layout from the effective configuration.

Would it make sense to model layouts as an ordered array rather than fixed primary/alternate fields? Ideally, the panel would support adding, removing, and reordering any number of layouts while preserving their corresponding variants.

At minimum, it would be helpful for the panel to detect configurations with more than two layouts and avoid replacing them, or warn that advanced configuration is active. I'd be happy to test the three-layout workflow with US English, Hebrew, and Greek.

KunalGautam:
Awaiting for this update to be pushed. Would help the people to add/remove keyboard layout without touching the config file (Or write manually)

PR #7254, opened 2026-08-17
Fix plugin bar loading and its silent failure

### LINKED ISSUE #7253: Cloned or third-party bar plugin fails to load, leaving no bar and no error
**Omarchy version:** 4.0.0-1
**Hyprland:** 0.56.2
**Shell:** omarchy-shell (Quickshell), 2 monitors

### What happened

`omarchy plugin clone omarchy.bar` succeeds and switches `bar.id` to the clone,
but the clone never loads. The result is **no bar on any monitor** — the
`omarchy-bar` layer surface is gone entirely — with nothing printed to explain
it and no fallback to the built-in bar.

```
$ omarchy plugin clone omarchy.bar
Cloned omarchy.bar to /home/koen/.config/omarchy/plugins/koen.bar and switched to koen.bar

$ hyprctl layers | grep -E 'Monitor|omarchy-bar'
Monitor HDMI-A-1:
Monitor DP-1:
```

The journal shows the cause, though nothing surfaces it to the user:

```
WARN scene: file:///home/koen/.config/omarchy/plugins/koen.bar/Bar.qml[22:3]:
  Required property barConfig was not initialized
WARN scene: file:///home/koen/.config/omarchy/plugins/koen.bar/Bar.qml[15:3]:
  Required property omarchyPath was not initialized
WARN scene: file:///home/koen/.config/omarchy/plugins/koen.bar/Bar.qml[17:3]:
  Required property barWidgetRegistry was not initialized
WARN scene: @shell.qml[256:-1]: ReferenceError: errorString is not defined
```

### What I expected

The cloned bar loads and renders, since cloning a first-party plugin to
customise it is the documented workflow.

Failing that, the documented fallback should kick in: `shell.qml` sets
`shell.failedBarId` on `Loader.Error` specifically so a broken bar option falls
back to `omarchy.bar`. It never runs.

### Steps to reproduce

1. `omarchy plugin clone omarchy.bar`
2. `omarchy restart shell`
3. No bar on any monitor. `hyprctl layers` shows no `omarchy-bar` surface.
4. Recover with `omarchy bar use omarchy.bar` (or delete the clone) and
   `omarchy restart shell`.

### Root cause

Two independent defects, the second of which hides the first.

**1. `shell/plugins/bar/Bar.qml` declares injected properties as `required`.**

```qml
required property string omarchyPath
required property var barWidgetRegistry
required property var barConfig
```

QML will not instantiate a component whose `required` properties are unset.
`shell.qml` has two paths to a bar:

- `defaultBarComponent` passes all three declaratively, so the built-in bar is fine.
- `pluginBarLoader` sets `source:` and injects afterwards, in
  `onLoaded: shell.configureBar(item, ...)`.

The second can never work: the Loader fails before `onLoaded`, and
`configureBar()` opens with `if (!target) return`, so it is a no-op. Every bar
plugin is therefore unloadable, including a clone of the built-in bar.

This also makes `Bar.qml` the only plugin entry point using `required`.
`agents/skills/shell-dev.md` states entry points "accept the shell-injected
properties `omarchyPath`, `shell`, `manifest`, and `pluginRegistry` /
`barWidgetRegistry`", and `menu/Menu.qml` and `notifications/Service.qml` both
declare them plain with defaults — as do `Bar.qml`'s own `shell` and `manifest`.

**2. `shell/shell.qml` reads `errorString`, which does not exist on `Loader`.**

```qml
onStatusChanged: {
  if (status === Loader.Error) {
    var detail = errorString && errorString() ? errorString() : ""
    console.warn("bar option " + ... + ":", detail)
    shell.failedBarId = shell.activeBarId   // never reached
  }
}
```

The `ReferenceError` aborts the handler before `failedBarId` is set, so the
fallback to the built-in bar never happens and the warning is never printed.
That is why the failure presents as a silently missing bar.

### Suggested fix

Declare the three properties plain with defaults, matching the plugin contract
and every other entry point, and drop the `errorString` reference so the
existing fallback can run. Verified on this machine: with the properties
relaxed, a cloned bar loads and renders normally, and the built-in bar is
unaffected because `defaultBarComponent` still passes all three declaratively.

I have a branch with both fixes (8 insertions, 5 deletions across the two
files) and can open a PR.

koenhendriks:
Fixes #7253.

`omarchy plugin clone omarchy.bar` reports success and switches `bar.id`, then the bar disappears from every monitor with nothing printed and no fallback. Two defects, the second hiding the first.

### 1. A plugin bar's injected properties arrive too late

A bar entry point declares `omarchyPath`, `barWidgetRegistry`, and `barConfig` as `required`, and QML refuses to instantiate a component whose required properties are unset. `shell.qml` reaches a bar two ways:

- `defaultBarComponent` passes them declaratively — the built-in bar is fine.
- `pluginBarLoader` bound `source:` and injected them afterwards, in `onLoaded: shell.configureBar(item, ...)`.

The second can never work: the Loader fails before `onLoaded`, and `configureBar()` opens with `if (!target) return`, making it a no-op. So no bar plugin could load — including a clone of the built-in bar, which is the documented way to customise one.

`setSource` supplies them at creation instead, with `Qt.binding` so they stay live afterwards the way `defaultBarComponent`'s bindings do. That also fixes something that was quietly broken: a plugin bar previously got `barConfig` assigned once in `onLoaded` and never followed a `shell.json` reload. Only the properties `Bar.qml` declares are passed; `configureBar()` still fills in optional ones such as `pluginRegistry` on load.

Since `active` does not change when switching straight from one bar plugin to another, the source url drives the reload itself.

### 2. The `Loader.Error` handler reads `errorString`, which `Loader` does not have

```qml
var detail = errorString && errorString() ? errorString() : ""
console.warn("bar option " + ... + ":", detail)
shell.failedBarId = shell.activeBarId   // never reached
```

The `ReferenceError` aborts the handler before `failedBarId` is set, so the existing fallback to `omarchy.bar` never runs and the warning is never printed. That is why a broken bar option presents as a silently missing bar rather than an error. The warning now also names the url it tried to load, since the engine logs the component's own errors without saying which bar option asked for them.

### Verification

On Omarchy 4.0.0-1, Hyprland 0.56.2, two monitors. Both runs start from a cold `~/.cache/quickshell/qmlcache` — a warm cache serves a previously compiled `Bar.qml` and will happily mask this.

- **Before:** `omarchy plugin clone omarchy.bar` + restart → `hyprctl layers` shows no `omarchy-bar` surface on either monitor; journal shows the three `Required property … was not initialized` warnings plus `ReferenceError: errorString is not defined`.
- **After:** the same untouched clone loads and renders on both monitors, with no warnings beyond one unrelated portal message. `Bar.qml` is not modified, so the clone still declares all three as `required`.

`./test/all` shows the same 4 pre-existing failures with and without this change (they want a sibling `omarchy-pkgs` checkout, snapper config, and a pristine system; the fourth is an IPC-handler count that is off on this machine regardless of branch or active bar id).

### Note on the history

The first two commits relaxed `Bar.qml`'s `required` properties. Review pushed back on weakening the contract, which was fair, and the objection pointed at a better fix: relaxing `Bar.qml` only helps bars descended from a *fixed* `Bar.qml`, so an existing clone or a third-party bar that declares its injected properties required would still fail. The later commits revert that and fix the loader instead. Net diff is `shell/shell.qml` only.

### DIFF
--- a/shell/plugins/bar/Bar.qml
+++ b/shell/plugins/bar/Bar.qml
@@ -12,14 +12,14 @@ Item {
   id: root
 
   // The omarchy-shell host injects omarchyPath from OMARCHY_PATH.
-  property string omarchyPath: Quickshell.env("OMARCHY_PATH")
+  required property string omarchyPath
   // Injected by the host shell so bar slots can resolve enabled widgets.
-  property var barWidgetRegistry: null
+  required property var barWidgetRegistry
   // Injected by the host shell every time shell.json is reloaded. Holds the
   // `bar:` subtree: position, centerAnchor, layout. The host owns file IO;
   // the bar just renders whatever it's handed. The bar font follows the
   // OS-level fontconfig monospace binding — it is not stored in shell.json.
-  property var barConfig: null
+  required property var barConfig
   // Injected by the host shell. Used for shell-wide actions such as opening
   // settings and persisting inline widget state.
   property var shell: null
--- a/shell/plugins/bar/Bar.qml
+++ b/shell/plugins/bar/Bar.qml
@@ -12,14 +12,14 @@ Item {
   id: root
 
   // The omarchy-shell host injects omarchyPath from OMARCHY_PATH.
-  required property string omarchyPath
+  property string omarchyPath: Quickshell.env("OMARCHY_PATH")
   // Injected by the host shell so bar slots can resolve enabled widgets.
-  required property var barWidgetRegistry
+  property var barWidgetRegistry: null
   // Injected by the host shell every time shell.json is reloaded. Holds the
   // `bar:` subtree: position, centerAnchor, layout. The host owns file IO;
   // the bar just renders whatever it's handed. The bar font follows the
   // OS-level fontconfig monospace binding — it is not stored in shell.json.
-  required property var barConfig
+  property var barConfig: null
   // Injected by the host shell. Used for shell-wide actions such as opening
   // settings and persisting inline widget state.
   property var shell: null
--- a/shell/shell.qml
+++ b/shell/shell.qml
@@ -253,8 +253,11 @@ ShellRoot {
     onActiveChanged: if (!active) shell.bar = null
     onStatusChanged: {
       if (status === Loader.Error) {
-        var detail = errorString && errorString() ? errorString() : ""
-        console.warn("bar option " + shell.activeBarId + " failed to load, falling back to " + shell.defaultBarId + ":", detail)
+        // Loader exposes no errorString; the engine has already logged the
+        // component's own QML errors by the time this runs. Reaching for one
+        // threw a ReferenceError here, which aborted the handler before the
+        // fallback below and left the session with no bar and no explanation.
+        console.warn("bar option " + shell.activeBarId + " failed to load, falling back to " + shell.defaultBarId)
         shell.failedBarId = shell.activeBarId
       }
     }
--- a/shell/shell.qml
+++ b/shell/shell.qml
@@ -246,20 +246,52 @@ ShellRoot {
   Loader {
     id: pluginBarLoader
 
+    // A bar entry point declares the host-injected properties as required, and
+    // QML refuses to instantiate a component whose required properties are
+    // unset. Binding `source` therefore failed before onLoaded could run, and
+    // configureBar()'s `if (!target) return` made the injection a no-op -- so no
+    // bar plugin could ever load. setSource supplies them at creation instead,
+    // and Qt.binding keeps them live afterwards, the way defaultBarComponent's
+    // declarative bindings do.
+    function loadPluginBar() {
+      if (!active) {
+        setSource("")
+        return
+      }
+
+      setSource(shell.activeBarSourceUrl, {
+        omarchyPath: Qt.binding(function() { return shell.omarchyPath }),
+        barWidgetRegistry: Qt.binding(function() { return shell.barWidgetRegistry }),
+        barConfig: Qt.binding(function() { return shell.barConfig }),
+        shell: shell,
+        manifest: Qt.binding(function() { return shell.activeBarManifest })
+      })
+    }
+
     active: !shell.pluginReloading && shell.activeBarId !== shell.defaultBarId && shell.activeBarSourceUrl !== ""
-    source: shell.activeBarId !== shell.defaultBarId ? shell.activeBarSourceUrl : ""
     asynchronous: true
     onLoaded: shell.configureBar(item, shell.activeBarManifest)
-    onActiveChanged: if (!active) shell.bar = null
+    onActiveChanged: {
+      if (!active) shell.bar = null
+      loadPluginBar()
+    }
     onStatusChanged: {
       if (status === Loader.Error) {
         // Loader exposes no errorString; the engine has already logged the
         // component's own QML errors by the time this runs. Reaching for one
         // threw a ReferenceError here, which aborted the handler before the
         // fallback below and left the session with no bar and no explanation.
-        console.warn("bar option " + shell.activeBarId + " failed to load, falling back to " + shell.defaultBarId)
+        console.warn("bar option " + shell.activeBarId + " (" + shell.activeBarSourceUrl + ") failed to load, falling back to " + shell.defaultBarId)
         shell.failedBarId = shell.activeBarId
       }
+    }
+
+    // Switching straight from one bar plugin to another keeps `active` true, so
+    // the url change has to drive the reload by itself.
+    Connections {
+      target: shell
+
+      function onActiveBarSourceUrlChanged() { pluginBarLoader.loadPluginBar() }
     }
   }
 
--- a/shell/shell.qml
+++ b/shell/shell.qml
@@ -263,12 +263,19 @@ ShellRoot {
     // and onBarConfigChanged keeps barConfig current from then on. A binding here
     // would be replaced moments after it was made.
     function loadPluginBar() {
-      if (!active) {
+      // `active` already covers activeBarSourceUrl !== "", but a signal handler can
+      // run before that binding is re-evaluated: on the change that empties the
+      // url, onActiveBarSourceUrlChanged fires while `active` is still true, and
+      // setSource("") with a property map is a needless load error. Reading the url
+      // once and checking it here keeps this correct whichever order they arrive in.
+      var url = shell.activeBarSourceUrl
+
+      if (!active || url === "") {
         setSource("")
         return
       }
 
-      setSource(shell.activeBarSourceUrl, {
+      setSource(url, {
         omarchyPath: shell.omarchyPath,
         barWidgetRegistry: shell.barWidgetRegistry,
         barConfig: shell.barConfig,
--- a/shell/shell.qml
+++ b/shell/shell.qml
@@ -271,6 +271,14 @@ ShellRoot {
     active: !shell.pluginReloading && shell.activeBarId !== shell.defaultBarId && shell.activeBarSourceUrl !== ""
     asynchronous: true
     onLoaded: shell.configureBar(item, shell.activeBarManifest)
+
+    // The load is driven by signals now that `source` is not bound, so a loader
+    // built with `active` already true would never get its initial setSource.
+    // That does not happen today -- the bar config arrives after construction, so
+    // `active` starts false and onActiveChanged does it -- but nothing enforces
+    // that ordering, and the failure mode is a session with no bar.
+    Component.onCompleted: loadPluginBar()
+
     onActiveChanged: {
       if (!active) shell.bar = null
       loadPluginBar()
--- a/shell/shell.qml
+++ b/shell/shell.qml
@@ -178,6 +178,12 @@ ShellRoot {
     return shell.barOptionAvailable(shell.selectedBarId)
   }
   readonly property string activeBarId: selectedBarId !== failedBarId && selectedBarAvailable ? selectedBarId : defaultBarId
+  // configureBar() assigns manifest imperatively at load, which replaces the
+  // binding the bar was created with -- and a manifest can be re-read (a plugin
+  // rescan) without its entry-point url changing, so nothing would reload the
+  // bar or refresh the value. Push it the same way barConfig is pushed above.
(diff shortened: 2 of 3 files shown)


koenhendriks:
Thanks — the review's objection to relaxing `required` was the right call, and following it produced a better fix. Rundown of all four points:

**Relaxing `barWidgetRegistry` / `barConfig` from `required` weakens the contract** — agreed, and reverted. It was also a narrower fix than it looked: it only helps bars descended from a *fixed* `Bar.qml`, so an already-existing clone, or a third-party bar that quite reasonably declares its injected properties `required`, would still have failed to load. The loader was the thing that was wrong. `pluginBarLoader` now supplies the properties at creation via `setSource(url, {...})`, with `Qt.binding` so they stay live afterwards. `Bar.qml` is untouched and keeps `required`; the net diff is `shell/shell.qml` only.

That turned out to fix a second, quieter bug: a plugin bar previously had `barConfig` assigned once in `onLoaded` and never followed a `shell.json` reload, where the built-in bar gets a real binding. It does now too.

**Normalising the `omarchyPath` default** — moot, since `Bar.qml` is no longer modified. Worth noting for its own sake though: `shell.qml:27` already does `property string omarchyPath: Quickshell.env("OMARCHY_PATH")`, so that was the codebase's existing idiom rather than something new.

**Feature-detect `errorString` rather than dropping the detail** — I've left it dropped, deliberately. `Loader` has no `errorString` member in any Qt version, so `typeof errorString === "function"` would be guarding an API that has never existed; the identifier was simply a mistake, which is why it threw `ReferenceError` and took the fallback down with it. No diagnostics are lost either way — the engine logs the component's own QML errors itself, which is exactly how the underlying `Required property … was not initialized` messages surfaced. What *was* missing is which bar option asked, so the warning now names `activeBarSourceUrl`.

One note for anyone reproducing this: clear `~/.cache/quickshell/qmlcache` first. A warm cache will serve a previously compiled `Bar.qml` and the bug appears to vanish — it cost me a wrong conclusion mid-investigation before I tested cold.

koenhendriks:
Picked up the suppressed comments from the last review.

**Initial load not guaranteed** — fixed in a2293009. With `source` no longer bound, the load depended on a change signal; a Loader constructed with `active` already true would never call `setSource`. It does not happen today (the bar config arrives after construction, so `active` starts false and `onActiveChanged` drives it), but nothing enforces that ordering and the failure mode is a session with no bar, so `Component.onCompleted: loadPluginBar()` now covers it.

**Manifest drift during an async load** — addressed in 2abba613, by removing the cause rather than reordering. `setSource` was injecting the four properties as `Qt.binding`s, but `onLoaded` hands the item to `configureBar()`, which reassigns all four from the live shell properties, and `onBarConfigChanged` keeps `barConfig` current after that. The bindings were being replaced almost as soon as they were created, so they are now plain values: the value injected at creation only has to satisfy the required property, and `configureBar` applies the current manifest at load time.

Verified on Omarchy 4.0.0-1, Hyprland 0.56.2, two monitors, with a clone whose three `required` properties are left intact so the original bug is reachable:

- stock `shell.qml`: `hyprctl layers` shows no `omarchy-bar` surface on either monitor
- this branch: the cloned bar loads and renders on both
- config still reaches the loaded plugin bar — `omarchy bar position bottom` moved the DP-1 surface from y=1440 to y=2854 and back, same layer id and pid, so it hot-reloaded in place rather than needing a restart

No new `Required property … was not initialized` or `ReferenceError` lines in the shell log.

koenhendriks:
Fourth review picked up in 4085d7d9. Both suppressed points came down to one gap, so one change covers them.

**Reload triggers** — the branch the review suggested enforcing is already structural, so no extra signal handlers are needed for a bar switch. `activeBarSourceUrl` is derived from `activeBarManifest`, which is derived from `activeBarId`, so changing the id necessarily changes the url and the existing `onActiveBarSourceUrlChanged` drives the reload.

There is a real residue though, and the review is right about it: `activeBarManifest` is recomputed from `pluginRegistry.registryRevision`, so a plugin rescan yields a new manifest while the entry-point url stays the same string. Nothing reloaded the loader or refreshed the value, so the bar kept a stale manifest. Worth noting this predates the PR — `configureBar()` assigns `manifest` imperatively at load and replaces whatever binding the component was created with, for the built-in bar as well.

Fixed by pushing it on change, mirroring `onBarConfigChanged` a few lines above:

```qml
onActiveBarManifestChanged: if (bar && "manifest" in bar) bar.manifest = shell.activeBarManifest
```

**Manifest consistency across an async load** — this settles that too, rather than pinning the creation-time value. The value injected at `setSource` only has to satisfy the required property; the item's manifest now converges on the current one whichever moment wins the race, instead of being frozen at whichever manifest happened to be live when the load started.

Verified by running this branch directly (`quickshell -n -p <checkout>/shell`) on Omarchy 4.0.0-1, two monitors: the cloned bar loads on both, and `omarchy-shell shell rescanPlugins` — exactly the new-manifest-same-url case — leaves both surfaces up with no manifest-related errors in the shell log.

koenhendriks:
Fifth review picked up: one fix, one push-back.

**Empty source url** — fixed in b06256c1. `loadPluginBar()` only gated on `active`, and while `active` does include `activeBarSourceUrl !== ""`, a signal handler can run before that binding is re-evaluated: on the change that empties the url, `onActiveBarSourceUrlChanged` fires while `active` is still true, so `setSource("")` was called with a property map attached — a load error for what is really just an unload. It now reads the url once and checks it explicitly, so the method is correct whichever order the signal and the binding arrive in.

**`onActiveBarManifestChanged` style** — leaving as is, deliberately. It is byte-for-byte the same shape as the `onBarConfigChanged` handler it mirrors a few lines above:

```qml
onBarConfigChanged: if (bar && "barConfig" in bar) bar.barConfig = shell.barConfig
onActiveBarManifestChanged: if (bar && "manifest" in bar) bar.manifest = shell.activeBarManifest
```

Qualifying one and not the other, or bracing one and not the other, would make the pair harder to read than either style applied consistently. On the ambiguity specifically: `bar` is `property var bar: null` on the `ShellRoot` (shell.qml:192) and there is no `id: bar` anywhere under `shell/`, so the unqualified reference has nothing to shadow it. Happy to change both handlers together if a maintainer prefers the explicit form.

Verified on Omarchy 4.0.0-1, two monitors, running this branch directly: `omarchy bar use omarchy.bar` then `omarchy bar use koen.bar` — the transition that empties the url — leaves two bar surfaces up in both directions with no load errors. The only warnings are the pre-existing transient `panels/weather/Panel.qml … of null` lines that appear whenever the bar is torn down and rebuilt, present on unpatched shells too.

koenhendriks:
Test added in 6919f505 — `test/shell.d/plugin-bar-loader-test.sh`.

You were right that nothing covered this path. `manifest-entrypoints-test.sh` instantiates the built-in bar directly, which exercises `defaultBarComponent`'s declarative properties and never reaches `pluginBarLoader`, so either regression would have passed.

Two cases, both against the real shell rather than a stand-in — the loader is part of `shell.qml`, so a reimplementation in a fixture would only test itself. It follows `runtime-smoke-test.sh`: a copy of `shell/` under a temp `OMARCHY_PATH`, an isolated `HOME` with a crafted `shell.json` pointing `bar.id` at a fixture plugin.

- **Injected properties.** A fixture bar plugin declaring `omarchyPath`, `barWidgetRegistry` and `barConfig` as `required`, exactly as the first-party bar does. QML cannot instantiate it unless the loader supplies all three at creation. It reports the values back through the result file, so a property arriving `undefined` fails rather than passing quietly.
- **Broken bar.** A fixture whose QML cannot load, asserting the log carries the fallback to `omarchy.bar` and that the handler did not throw a `ReferenceError` on the way.

Both were checked against the unfixed code so they are not vacuous:

| variant | result |
| --- | --- |
| loader fix absent | case 1 fails — "never loaded" |
| only the `errorString` bug restored | case 1 passes, case 2 fails — "did not fall back" |

`./test/all` is 181 files with this added; the 4 failures on my machine (`config-test`, `runtime-smoke-test`, `snapper-test`, `unowned-system-paths-test`) are unchanged from a clean `quattro` checkout — they want a sibling `omarchy-pkgs` checkout, snapper config, and a pristine system.

One thing I noticed while in there, and did not touch since it is outside this PR: the panel loader at `shell.qml:704` uses the same `errorString && errorString()` idiom this PR removed from the bar loader, with a comment saying `Loader.errorString()` reflects a source-load failure even when `sourceComponent` is null. In the bar loader that reference threw `ReferenceError: errorString is not defined` — that is what the fallback fix was for, and the new test's second case reproduces it. If the identifier does not resolve on a `Loader` here, the panel handler has the same latent problem, just on a rarer path. Worth a look by someone who knows which Qt version introduced it.

koenhendriks:
Fixed in 723623d7. You are right that the assertion proved nothing: the handler prints that warning before assigning `failedBarId`, so removing the assignment left the log unchanged and the test still passed.

It now asks the shell for the state instead. Once the broken bar is given up on, `activeBarId` is the built-in one, and `listPlugins` reports `omarchy.bar` with `active: true` and `acme.broken-bar` with `active: false`. The test polls the IPC for that, having first waited for `shell ping`, and dumps the bar-kind plugin rows plus the shell log if it never arrives.

Checked against exactly the hole you described: with the warning left in place and `shell.failedBarId = shell.activeBarId` removed, the case now fails with "did not fall back to the built-in bar". The old log grep passed under that same change.

koenhendriks:
Added in 6b390432. You were right that the path was uncovered: with `onActiveBarSourceUrlChanged` gutted, both existing cases still passed.

The new case installs two valid bar plugins, starts on the first, then switches to the second and waits for that entry point to report its own id back.

One thing the switch had to get right to be worth anything. Rewriting `shell.json` and reloading it does not exercise this path — I traced it, and the reload empties the config on its way through, so `activeBarSourceUrl` passes through `""`, the loader deactivates and reactivates, and `onActiveChanged` does the reload. So the case switches via `enablePlugin`, which sets `bar.id` through the in-process config mutator: both plugins stay installed and the url stays non-empty, so `active` never drops and the url change is the only thing left to act.

Checked against the failure you described — a handler that stops reloading while `active` remains true:

```
ok     - a bar plugin declaring its injected properties required is loaded with them supplied
ok     - a bar plugin that cannot load falls back to the built-in bar
not ok - switching between two bar plugins did not load the second entry point
```

Issue #6981, opened 2026-08-15
Plugin hot-reload never picks up changed plugin code (stale QML component cache)


farangkao:
### System details

13th Gen Intel(R) Core(TM), AMD Radeon AI PRO R9700

### What's wrong?

**Title:** Plugin hot-reload silently reloads stale code — `Qt.clearComponentCache` doesn't exist in QML

Note: analysis and report generated with AI (GLM 5-3), but reviewed and tested by a human in the loop (me) 

## Summary
 
When a locally installed plugin's files change (e.g. after `omarchy plugin update`),
the shell notices and logs `Local plugin changed, reloading: <id>`, and plugin
services/widgets are re-instantiated — but they are re-created from the **old
compiled QML**, so the update has no effect until the whole shell is restarted.
 
## Environment
 
- Omarchy 4.0.0-1 (stable channel, `4.0.0.alpha`)
- quickshell 0.3.0 (quickshell-git, revision 28771c7c)
- Plugin under test: `lobo.omaspaces` (any third-party bar-widget plugin shows this)
 
## Reproduction
 
1. Install a git-managed bar-widget plugin at an older version, with the shell running.
2. Run `omarchy plugin update <id>` (or just edit a file under
   `~/.config/omarchy/plugins/<id>/` — the file watcher triggers the same path).
3. The shell logs `Local plugin changed, reloading: <id>` and reloads.
4. Observe: the widget still runs the old code. Only `omarchy restart shell`
   applies the update.
 
Concrete example: after updating `lobo.omaspaces` to a version that added a
right-click handler on its workspace pills, right-click kept doing nothing
after the automatic reload; `omarchy restart shell` fixed it.
 
A minimal proof that edits are ignored by the reload: adding a visible change
(e.g. a marker character in the widget's label) plus a
`Component.onCompleted: console.warn(...)` line to the plugin's QML triggered
the `reloading` log line, but the marker never appeared and the log line never
fired — the live instance was re-created from the previous compilation.
 
## Root cause
 
`shell.qml` reload path: file change → `localPluginReloadTimer` (line 61) →
`reloadPlugins()` (line 739) → `finishPluginReload()` (line 751):
 
```qml
if (typeof Qt.clearComponentCache === "function") Qt.clearComponentCache()
```
 
`Qt.clearComponentCache` is a C++-only method on `QQmlEngine`; it is not
exposed to the QML `Qt` object. Verified live — a quickshell config evaluating
[...]

PR #7259, opened 2026-08-17
Fix plugin reload stale qml

ilyaZar:
refs #6981

### Regarding the issue:

I can confirm the issue is real with my own plugin (plugin-control): also, on a testing machine, I had omarchy running for a couple of days and the shell wasnt restarded during that time

next, after using `omarchy plugin remove <...>` plus `omarchy plugin add <...>`  on my plugin, I couldnt get it updated, and observed the stale behavior; the plugin update did not carry over properly 

### Regarding its solution

Essentially,  this is a question about plugin lifecycle management or so:

user could own the complete removal process as per the documentation, but then every plugin maintainer would have to add more lines to removal code in addition to the simple `omarchy plugin remove ...` a follow up on shell reloads is necessary, which would duplicate to over 300 plugins or so. also many might just forget, and silent bugs being introduced all the time 

integration into existing plugin removal seems (in some form or another) advantegous; not claiming this is the best way here, but its one with fewest lines of code changes I can think of atm. ; alternatives could be to elicit code parts from shell restart into the `omarchy plugin remove ...` but thats kinda duplication

not aware of strong side effects (overkill maybe, a bit heavier than needed; not dangerous in any form) stemming from `omarchy restart shell`; if so, duplication of code in `omarchy plugin remove ...` and change of exact behavior therein could be be better of course

### DIFF
--- a/bin/omarchy-plugin-remove
+++ b/bin/omarchy-plugin-remove
@@ -114,7 +114,7 @@ else
   echo "Removed $id. Backup at: $backup"
 fi
 
-omarchy-shell shell rescanPlugins >/dev/null
+omarchy-restart-shell
 
 if [[ -n $cloned_from && $was_enabled == "true" ]]; then
   echo "Restored $cloned_from."
--- a/bin/omarchy-plugin-update
+++ b/bin/omarchy-plugin-update
@@ -128,6 +128,6 @@ for id in "${targets[@]}"; do
 done
 
 if (( UPDATED_ANY )); then
-  omarchy-shell shell rescanPlugins >/dev/null
+  omarchy-restart-shell
 fi
 exit $rc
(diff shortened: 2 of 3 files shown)


ilyaZar:
maybe a note bc the clanker picked it up as well

1. I do believe in the case of `remove` -> `add` life cycle, that a full `omarchy restart shell` is better , hence 1624d64af8f74fa70ab90ae78b73edc028c8b84b ; it "guarantees" that adding the same plugin ID cannot reuse QML cached before removal

2. I am not sure about `update` as the trade-off is stability (`omarchy restart shell` seems to always work) vs. it would need to correctly identify  and reload every affected service, panel, widget, and relative QML/JS dependencies etc. ; to me that seems more complex and (likely) more brittle ; so I'd lean still more towards 6d0df9d9fa6c558603b818f3268c5e75a933968d than a targeted solution

PR #7261, opened 2026-08-17
Add Kimi as a selectable default coding agent

sorenmat:
## Summary

[Kimi Code](https://www.kimi.com/code/docs/en/) (`kimi`) can't be picked as the default coding agent — the picker in `omarchy-default-agent` is a hardcoded list and Kimi isn't in it. This adds it everywhere the other agents appear:

- `bin/omarchy-default-agent` — `kimi` / `kimi-code` map to the `kimi` agent, installed through mise as `npm:@moonshot-ai/kimi-code` (the official package, exposes a `kimi` bin)
- `bin/omarchy-agent` — launches unattended as `kimi --yolo`; an initial prompt uses `kimi --prompt`, which is non-interactive print mode (Kimi has no interactive initial-prompt flag, and `--prompt` cannot combine with `--yolo` — same shape as the `crush run` case)
- `default/omarchy/omarchy-menu.jsonc` — Kimi entry under _Setup > Defaults > Agent_, reusing the robot glyph from the Agent group entry
- `install/user/mise.sh` — preinstalls a lazy `~/.local/bin/kimi` stub like the other agents, with migration `1786967403.sh` adding it on existing systems (respecting the preinstalls-removed opt-out) and `omarchy-remove-preinstalls` deleting it again
- `manual/17-ai.md` — Kimi row in the pre-wired agents table
- Tests updated for the new provider, its aliases, package, launch flags, lazy stub, migration, opt-out, and menu ordering

## Notes

- Not added: a usage collector for the agents panel. Happy to do that as a follow-up.
- `./test/all` passes (180/180 test files).

## Test plan

- [x] `./test/all` — all 180 test files pass
- [ ] Pick Kimi under _Setup > Defaults > Agent_ on a machine without it installed (installs via mise, then launches)


sorenmat:
Follow-up: #7274 adds a Kimi usage collector, so the agents panel gains a Kimi tab (weekly quota + session limits) through the existing collector discovery.

kyryl-bogach:
This would be really nice!

PR #7269, opened 2026-08-17
Run the shell on the Vulkan backend on primary NVIDIA GPUs

### LINKED ISSUE #7268: Shell animations run at ~60 fps regardless of monitor refresh rate on NVIDIA (Qt falls back to the basic render loop)
## System

- Omarchy 4.0.0, Hyprland 0.56.2, quickshell-git 0.3.0.r20
- Qt 6.11.1 (qt6-declarative, qt6-wayland)
- CPU: AMD Ryzen 9 7950X3D, GPU: NVIDIA RTX 4090 (nvidia-open-dkms 610.57.04)
- Monitor: 4K @ 240 Hz (DP-1)

## What happens

Every animation the shell draws — the tray drawer slide, menu and panel transitions, OSDs — runs at ~60 fps no matter the monitor refresh rate. On a high-refresh monitor this reads as visibly choppy, "low fps" motion, while Hyprland itself (window animations, cursor) is perfectly smooth at 240 Hz.

Measured from inside the shell with a `FrameAnimation` counter: the bar surface renders at **62 fps** on a 240 Hz monitor.

## Why (root cause)

Qt disables threaded OpenGL rendering on the proprietary NVIDIA driver. In `qtbase src/plugins/platforms/wayland/.../qwaylandeglclientbufferintegration.cpp`:

```cpp
// On desktop NVIDIA resizing QtQuick freezes them when using threaded rendering QTBUG-95817
// In order to support threaded rendering on embedded platforms where resizing is not needed
// we check if XDG_CURRENT_DESKTOP is set which desktop environments should set
if (qstrcmp(vendor, "NVIDIA") == 0 && qEnvironmentVariableIsSet("XDG_CURRENT_DESKTOP")) {
    m_supportsThreading = false;
}
```

With `ThreadedOpenGL` reported as unsupported, Qt Quick picks the **basic render loop**, whose animations are driven by a ~16 ms timer with no vsync awareness — hence a hard 60 fps ceiling for the whole shell. Verified with `QSG_INFO=1`: `basic render loop` on NVIDIA; Mesa systems (radeonsi hardware and even llvmpipe) report `threaded render loop` and are unaffected, which is why this only bites NVIDIA users.

## Reproduction

On any NVIDIA + Wayland machine:

```bash
mkdir -p /tmp/qs-loop-test && cat > /tmp/qs-loop-test/shell.qml <<'EOF'
import Quickshell
import QtQuick
PanelWindow {
  anchors { top: true; left: true }
  implicitWidth: 1
  implicitHeight: 1
}
EOF
env -u QSG_RENDER_LOOP -u QSG_RHI_BACKEND QSG_INFO=1 \
  QT_LOGGING_RULES="qt.scenegraph.general=true;qt.rhi.general=true" \
  timeout 4 quickshell -p /tmp/qs-loop-test > /tmp/qs-loop.log 2>&1
grep -iE "render loop|VENDOR" /tmp/qs-loop.log
```

NVIDIA output: `basic render loop` + `OpenGL ES 3.2 NVIDIA`. Mesa output: `threaded render loop`.

Independently reproduced on a second NVIDIA machine (RTX 5080, driver 610.57.04): same `basic render loop`. Two Mesa machines (RX 9070 XT radeonsi hardware; llvmpipe software) both report `threaded render loop`.

## Evidence

https://github.com/user-attachments/assets/6bfbf4fd-b9f0-41d1-baaf-4e936742a0a3

`ball_comparison.mp4` — a minimal bar-widget with a bouncing ball and a live `FrameAnimation` fps counter, recorded at 240 fps, stock shell vs `QSG_RHI_BACKEND=vulkan`: the counter reads 62 vs 240 on the same machine.

https://github.com/user-attachments/assets/3c5776d6-2a0e-4141-8b9a-a1cd81619945

`tray_comparison.mp4` — the same difference on the stock tray drawer animation.

Per-frame analysis of the recordings (mean-absolute-difference between consecutive frames, 0.25 s sliding window): stock peaks at **64 frame changes/s**, Vulkan at **240/s**.

## Proposed fix

Launch the shell with `QSG_RHI_BACKEND=vulkan` on NVIDIA systems (PR attached). Rationale for the Vulkan backend rather than forcing `QSG_RENDER_LOOP=threaded` on GL:

- For the Vulkan backend the threaded loop is Qt's **own default** on every platform — no driver workaround gets overridden.
- Forcing threaded on GL re-exposes the [QTBUG-95817](https://bugreports.qt.io/browse/QTBUG-95817) resize freeze the vendor check exists to prevent, and an analogous forced-threaded setup is known to segfault on AMD iGPUs ([quickshell#720](https://github.com/quickshell-mirror/quickshell/issues/720)).
- The shell's QML is backend-agnostic (no raw GL anywhere), and quickshell maintains the Vulkan path of its native interop (e.g. [quickshell#648](https://github.com/quickshell-mirror/quickshell/issues/648)).

Gated on three conditions, so it is a strict no-op everywhere the bug does not apply: the NVIDIA Vulkan ICD is installed, the GPU is Turing or newer (current driver family with a solid Wayland Vulkan WSI; legacy branches keep today's behavior), and the NVIDIA GPU is the primary display adapter — on hybrid laptops where the iGPU drives the display, clients get Mesa's EGL, the threaded loop is available anyway, and forcing Vulkan would only add cross-GPU risk for no benefit.

Tested on the machine above: 12+ hours including overnight idle/screensaver cycles, lock screen (ScreencopyView + dual-GPU dmabuf interop), and a full day of regular use. Shell animations run at the full 240 Hz.

## Workaround for affected users (until fixed)

```lua
-- ~/.config/hypr/monitors.lua
hl.env("QSG_RHI_BACKEND", "vulkan")
```

then `omarchy restart shell`.

PavelAlennikov:
Fixes #7268

On the proprietary NVIDIA driver, qtwayland reports threaded OpenGL as unsupported (the [QTBUG-95817](https://bugreports.qt.io/browse/QTBUG-95817) resize-freeze workaround), so Qt Quick drops to the timer-driven basic render loop and every shell animation runs at ~60 fps regardless of monitor refresh rate. On a 240 Hz monitor the bar measures 62 fps stock and 240 fps with the Vulkan scene graph backend, which is not subject to that workaround and uses the vsync-driven threaded loop by design (see the issue for videos with a live in-bar fps counter, per-frame analysis of 240 fps captures, and a one-command reproduction).

Why the Vulkan backend rather than forcing `QSG_RENDER_LOOP=threaded` on GL: the forced-threaded route re-exposes the very resize freeze the vendor check protects against, and an analogous setup is known to segfault on AMD iGPUs ([quickshell#720](https://github.com/quickshell-mirror/quickshell/issues/720)). For Vulkan, threaded is Qt's own default on every platform, so no driver workaround gets overridden. The shell's QML carries no raw GL, and quickshell maintains the Vulkan path of its native interop ([quickshell#648](https://github.com/quickshell-mirror/quickshell/issues/648)).

The switch is gated three ways and is a strict no-op everywhere the bug does not apply:

- the NVIDIA Vulkan ICD is installed (rules out nouveau and hardware without the proprietary driver);
- the GPU is Turing or newer, via `omarchy-hw-nvidia-gsp` (limits the change to the current driver family; legacy branches keep today's behavior);
- the NVIDIA GPU is the primary display adapter (on hybrid laptops with the iGPU driving the display, clients get Mesa's EGL, the threaded loop is available anyway, and forcing Vulkan would only add cross-GPU risk).

Testing:

- Reproduced and fixed on RTX 4090 (610.57.04, 4K@240); independently reproduced on RTX 5080. Two Mesa machines (radeonsi hardware, llvmpipe) confirm the default threaded loop and take the no-op path.
- 12+ hours on the Vulkan backend including overnight idle/screensaver cycles, the lock screen (ScreencopyView dmabuf interop on a dual-GPU system), and a full day of regular use.
- Both branches of the gate were exercised end-to-end with the patched launcher and a scrubbed environment: on this machine the gate alone sets the variable; with the hardware detector pointed at an empty sysfs path (`OMARCHY_PCI_DEVICES_PATH`) the shell starts with no QSG variables at all, confirming the no-op path (`env "${arr[@]}"` with an empty array expands to zero words).
- `test/all`: the cli suite passes. In the shell suite, config, snapper, and unowned-system-paths pass with sibling omarchy-pkgs/omarchy-iso checkouts in place; bar-icon-geometry fails identically on an unpatched quattro checkout, on both the GL and Vulkan backends, and at monitor scale 1 and 2 — a constant 0.59 px glyph-centering delta from this machine's font rasterization, unrelated to this change (which touches no rendering code, only the launch environment).

### DIFF
--- a/bin/omarchy-launch-shell
+++ b/bin/omarchy-launch-shell
@@ -14,8 +14,42 @@
 # A package upgrade rewriting $OMARCHY_PATH/shell would otherwise reload it
 # against a half-written tree, and that failed reload leaves a second engine
 # generation behind that turns the next restart's IPC kill into a crash.
+
+# Qt refuses threaded OpenGL rendering on the proprietary NVIDIA driver (the
+# QTBUG-95817 resize-freeze workaround in qtwayland), which drops Qt Quick to
+# the timer-driven basic render loop: every shell animation runs at ~60 fps
+# regardless of monitor refresh rate. The Vulkan scene graph backend is not
+# subject to that workaround and uses the vsync-driven threaded loop by
+# design, so prefer it where that bug actually applies. Forcing
+# QSG_RENDER_LOOP=threaded on GL instead would re-expose the very freeze the
+# vendor check protects against.
+#
+# The workaround only bites when the compositor's EGL stack is NVIDIA's, so
+# gate on the NVIDIA GPU being the primary display adapter: on hybrid laptops
+# with the iGPU driving the display, clients get Mesa's EGL, the threaded
+# loop stays available, and forcing Vulkan would only add cross-GPU risk for
+# no benefit. Turing-or-newer (GSP) limits this to the current driver family
+# with a solid Wayland Vulkan WSI; legacy branches keep today's behavior.
+nvidia_is_primary_gpu() {
+  local card
+
+  for card in /sys/class/drm/card*/device; do
+    [[ -f $card/boot_vga && $(< "$card/boot_vga") == 1 ]] || continue
+    [[ $(< "$card/vendor") == "0x10de" ]] && return 0
+  done
+
+  return 1
+}
+
+shell_render_env=()
+if [[ -f /usr/share/vulkan/icd.d/nvidia_icd.json ]] &&
+  "$OMARCHY_PATH/bin/omarchy-hw-nvidia-gsp" &&
+  nvidia_is_primary_gpu; then
+  shell_render_env=(QSG_RHI_BACKEND=vulkan)
+fi
+
 run_shell() {
-  QS_DISABLE_FILE_WATCHER=1 QS_NO_RELOAD_POPUP=1 \
+  QS_DISABLE_FILE_WATCHER=1 QS_NO_RELOAD_POPUP=1 env "${shell_render_env[@]}" \
     systemd-cat -t omarchy-shell -- quickshell -n -p "$OMARCHY_PATH/shell" &
   shell_pid=$!
 


kristofferR:
Independent data point from another RTX 4090 (nvidia 610.57.04, Qt 6.11.2, quickshell 0.3.1, Hyprland, 5120x1440 @ 240 Hz):

- Stock shell: `QSG_INFO=1` reports `basic render loop`, `OpenGL VENDOR: NVIDIA`. A `FrameAnimation` probe counts 118-125 frames per 2 s, so ~60 fps for every shell animation.
- `QSG_RHI_BACKEND=vulkan`: `threaded render loop`, 480 frames per 2 s, i.e. the full 240 Hz. Same for `QSG_RENDER_LOOP=threaded` on GL.

So the diagnosis and the numbers reproduce exactly.

One thing worth folding into the decision: the root cause is already fixed upstream. [QTBUG-95817](https://bugreports.qt.io/browse/QTBUG-95817) was closed by qtbase change [723039](https://codereview.qt-project.org/c/qt/qtbase/+/723039) ("wayland: Resize wl_egl_window from the correct thread with an active context", merged 2026-06-26), which removes the NVIDIA vendor check that [536202](https://codereview.qt-project.org/c/qt/qtwayland/+/536202) added and explicitly re-enables threaded rendering on NVIDIA (compare [qtbase 6.11](https://github.com/qt/qtbase/blob/6.11/src/plugins/platforms/wayland/plugins/hardwareintegration/wayland-egl/qwaylandeglclientbufferintegration.cpp#L98-L103) with [dev](https://github.com/qt/qtbase/blob/dev/src/plugins/platforms/wayland/plugins/hardwareintegration/wayland-egl/qwaylandeglclientbufferintegration.cpp)). It lands in Qt 6.12, [scheduled for 2026-09-22](https://wiki.qt.io/Qt_6.12_Release). Once Arch ships that, stock GL gets the threaded loop on NVIDIA with no env var at all.

Given that, the gate here would keep forcing Vulkan long after the reason for it is gone, and the Vulkan path is the less-exercised one in quickshell (dmabuf interop for `ScreencopyView` etc., cf. [quickshell#648](https://github.com/quickshell-mirror/quickshell/issues/648)). Two options that seem cleaner:

1. Add a Qt version condition to the gate (`qt6-declarative` < 6.12) so the override retires itself, or
2. Treat this as a short-lived workaround and revisit once 6.12 is in the repos.

For what it's worth, other Quickshell shells went the threaded-GL route: [caelestia](https://github.com/caelestia-dots/shell/commit/7265ae3) (closing [caelestia-dots/shell#153](https://github.com/caelestia-dots/shell/issues/153), the same NVIDIA symptom), DankMaterialShell and iNiR all ship `//@ pragma Env QSG_RENDER_LOOP=threaded`. I've been running that on GL for the last session with plugins that use `ScreencopyView`, `MultiEffect` layers and `BackgroundEffect` blur; no resize freezes or crashes so far, though that is a much shorter soak than the 12+ hours reported here.

Written by Fable

PR #7274, opened 2026-08-17
Add a Kimi usage collector to the agents panel

sorenmat:
## Summary

Follow-up to #7261 (which makes Kimi selectable as the default coding agent): this adds `bin/omarchy-agent-usage-kimi`, so the agents panel gains a Kimi tab through the existing collector discovery — no plugin changes needed beyond enabling the provider in the manifest.

The collector prints the standard usage record:

- **Limits**: the weekly quota and per-window limits (e.g. the 5-hour session) from `api.kimi.com/coding/v1/usages` — the same payload the CLI's `/usage` command renders
- **Auth**: `KIMI_API_KEY` when set, otherwise the kimi-code sign-in at `$KIMI_CODE_HOME/credentials/kimi-code.json` (default `~/.kimi-code/...`), falling back to the legacy kimi-cli store at `~/.kimi/...`
- **Token refresh**: expired access tokens are refreshed the way the CLI does — form POST to `auth.kimi.com/api/oauth/token` under a `flock` on the sibling lock file, with a re-read after acquiring the lock. The server issues single-use refresh tokens, so the rotated pair is always persisted back (atomic replace, mode 0600); both the collector and the CLI keep working from the same file
- **Token safety**: tokens are sent only to hosts under `kimi.com` over HTTPS, redirects are refused (urllib would otherwise forward the Authorization header cross-host), and tokens never touch disk, logs, or the published record. Plain HTTP is allowed only for loopback hosts so tests can mock the endpoint via `KIMI_USAGE_URL`

Also updates the agents plugin manifest (Kimi enabled by default), the plugin README's collector table, and the manual.

No local token stats: kimi-code does not persist per-session token counts to disk, so the record is quota-only (`hasLocalStats` carries no prompt history). If a local stats source appears, the collector can grow one later.

## Test plan

- [x] `./test/all` — all 181 test files pass, including the new `agent-usage-kimi-test.sh` (no-credentials record, env-key precedence, kimi-code vs legacy store preference, refresh rotation persistence, endpoint trust, payload parsing, record contract)
- [x] Live run against `api.kimi.com` with a real kimi-code sign-in: returns the weekly and session limits with correct percentages and reset timestamps
- [ ] Panel shows the Kimi tab after `omarchy agent usage update` on a signed-in machine

### DIFF
--- a/bin/omarchy-agent-usage-kimi
+++ b/bin/omarchy-agent-usage-kimi
@@ -1,0 +1,388 @@
+#!/usr/bin/python3
+# omarchy:summary=Print the Kimi usage record as JSON
+# omarchy:args=[--force] [--limits-only]
+# omarchy:hidden=true
+"""Collect Kimi for Coding quota into one display-ready JSON record.
+
+Quoted limits come from api.kimi.com's usage monitor (the same payload the
+kimi /usage command renders): a weekly quota plus per-window limits such as
+the 5-hour session window. The bearer token comes from KIMI_API_KEY when set,
+otherwise from the kimi-code sign-in at $KIMI_CODE_HOME/credentials/kimi-code.json
+(default ~/.kimi-code/credentials/kimi-code.json), falling back to the legacy
+kimi-cli store at ~/.kimi/credentials/kimi-code.json.
+
+Access tokens expire, so this collector refreshes them the way the CLI does:
+POST the refresh_token to auth.kimi.com/api/oauth/token with the public client
+id, take the flock on kimi-code.lock first (another kimi process may be
+refreshing), and persist the rotated tokens back to the credentials file,
+since the server issues single-use refresh tokens, so the new one must never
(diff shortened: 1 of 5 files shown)


sorenmat:
Confirmed the credential resolution against the official docs: Kimi Code CLI stores all runtime data, including OAuth credentials, under `~/.kimi-code/` (or `$KIMI_CODE_HOME`), with credentials at `credentials/<name>.json` ([data locations](https://www.kimi.com/code/docs/en/kimi-code-cli/configuration/data-locations.html)). The collector's lookup order — `KIMI_API_KEY` → `$KIMI_CODE_HOME/credentials/kimi-code.json` → `~/.kimi-code/credentials/kimi-code.json` → legacy `~/.kimi/credentials/kimi-code.json` — matches that, with the last entry only serving installs still on the legacy kimi-cli.

PR #7278, opened 2026-08-17
capture: exclude media-controller nodes from the webcam list

### LINKED ISSUE #7277: webcam picker: Device Caps filter still passes IPU7 raw Bayer node (follow-up to #5722)
### Summary

The `Device Caps` filter added for #5722 does not exclude the raw sensor node on **IPU7 with the mainline `intel-ipu7` / `isys` driver**. Unlike the IPU6 + `v4l2-relayd` setup in that report, this node *does* advertise `Video Capture`, so it passes `capture_capable()` and is still offered as a webcam — while exposing only Bayer formats that `mpv av://v4l2:` cannot consume.

### Evidence

`/dev/video0` on this machine (Core Ultra 7 258V, Lunar Lake, sensor OV02C10, mainline `intel_ipu7` + `ov02c10`, no Intel HAL / `v4l2-relayd` involved):

```
$ v4l2-ctl -d /dev/video0 --info
Driver name      : isys
Card type        : ipu7
Device Caps      : 0x24200001
        Video Capture          <-- passes the filter added for #5722
        I/O MC
        Streaming
        Extended Pix Format
```

But every advertised format is raw Bayer:

```
$ v4l2-ctl -d /dev/video0 --list-formats
    [0]: 'BG12' (12-bit Bayer BGBG/GRGR)
    [1]: 'GB12' (12-bit Bayer GBGB/RGRG)
    [2]: 'BA12' (12-bit Bayer GRGR/BGBG)
    [3]: 'RG12' (12-bit Bayer RGRG/GBGB)
    [4]: 'BG10' (10-bit Bayer BGBG/GRGR)
    ...
    [11]: 'RGGB' (8-bit Bayer RGRG/GBGB)
    [12]: 'pBCC' (12-bit Bayer BGBG/GRGR Packed)
    [13]: 'pGCC' (12-bit Bayer GBGB/RGRG Packed)
```

There is no YUYV/MJPG/NV12 entry at all. On this hardware the usable camera is reached through libcamera (software ISP) via PipeWire, not through any `/dev/video*` node.

### Impact

```
$ omarchy hw webcam            # exit 0 — false positive
$ omarchy-capture-webcam-list
/dev/video0  ipu7 (PCI:0000:00:05.0):
```

- `omarchy hw webcam` reports a webcam on machines that have no V4L2-usable camera.
- `omarchy capture screenrecording --with-webcam` auto-detects the first entry, so it always lands on the Bayer-only node and produces a black overlay with no error — the same symptom #5722 described.
- With a `v4l2loopback` node added, the picker lists the broken `/dev/video0` **first**, so it is also the auto-detect winner:

```
/dev/video0   ipu7 (PCI:0000:00:05.0):
/dev/video40  Camara integrada (platform:v4l2loopback-040):
```

### Suggested fix

Extend `capture_capable()` in `omarchy-capture-webcam-list` to also require at least one non-Bayer pixel format. Raw sensor/ISYS nodes expose Bayer exclusively; every real capture device (UVC, `v4l2loopback`) exposes at least one consumable format.

```bash
capture_capable() {
  local device="$1"

  v4l2-ctl --device "$device" --info 2>/dev/null | awk '
    /^[[:space:]]*Device Caps[[:space:]]*:/ { inspect = 1; next }
    inspect && /^[[:space:]]*Video Capture/ { found = 1 }
    END { exit !found }
  ' || return 1

  # Raw sensor nodes (Intel IPU6/IPU7 ISYS) can advertise Video Capture while
  # exposing only Bayer formats, which mpv/ffmpeg cannot consume.
  v4l2-ctl --device "$device" --list-formats 2>/dev/null | awk '
    /^[[:space:]]*\[[0-9]+\]:/ && !/Bayer/ { found = 1 }
    END { exit !found }
  '
}
```

This keeps current behaviour for UVC-only laptops (a single passing entry) and collapses the IPU7 case to either the real loopback node or, when none exists, an empty list — which is the honest answer, and lets `omarchy hw webcam` correctly report no webcam.

### Environment

- Omarchy 4.0.0-1
- Kernel 7.1.8-arch1-3
- Intel Core Ultra 7 258V (Lunar Lake), IPU7, sensor OV02C10
- Mainline `intel_ipu7` / `intel_ipu7_isys` / `ov02c10` + libcamera 0.7.2 (software ISP); no Intel camera HAL or `v4l2-relayd`

Refs #5722.

danteRub:
Fixes #7277.

### Problem

`capture_capable()` accepts Intel IPU6/IPU7 ISYS nodes. They advertise `Video Capture`, but they also set `I/O MC` — their media-controller pipeline has to be configured with `media-ctl` before they can stream. `mpv av://v4l2:` opens the device directly, so these nodes can never work with the webcam overlay.

The `Device Caps` check added for #5722 does not catch this on IPU7 with the mainline `isys` driver, because unlike the IPU6 + `v4l2-relayd` case in that report, this node *does* report `Video Capture`:

```
$ v4l2-ctl -d /dev/video0 --info
Driver name      : isys
Card type        : ipu7
Device Caps      : 0x24200001
        Video Capture
        I/O MC
        Streaming
        Extended Pix Format
```

The result is a black overlay with no error: `/dev/video0` sorts first, so it wins the `--with-webcam` auto-detect, and `omarchy hw webcam` returns a false positive on machines with no V4L2-usable camera at all.

Streaming from it fails exactly as the capability predicts:

```
$ ffmpeg -f v4l2 -input_format yuyv422 -i /dev/video0 -frames:v 1 -f null -
ioctl(VIDIOC_STREAMON): Broken pipe
```

### Fix

Reject devices that set `I/O MC`. UVC webcams and `v4l2loopback` nodes never set it, so they are unaffected.

Filtering on pixel formats does *not* work here, in case it looks like the more obvious fix: this node advertises `UYVY`, `YUYV`, `RGBP` and `BGR3` alongside its Bayer formats even though the attached OV02C10 can only produce Bayer. `I/O MC` is the property that actually distinguishes "needs pipeline setup" from "openable directly".

### Verification

Tested on the hardware from #7277 — Core Ultra 7 258V (Lunar Lake), IPU7, OV02C10, mainline `intel_ipu7` + libcamera software ISP, with a `v4l2loopback` node at `/dev/video40` fed from `libcamerasrc`.

Before:

```
$ omarchy-capture-webcam-list
/dev/video0   ipu7 (PCI:0000:00:05.0):
/dev/video40  Camara integrada (platform:v4l2loopback-040):
```

After, with the loopback bridge running:

```
$ omarchy-capture-webcam-list
/dev/video40  Camara integrada (platform:v4l2loopback-040):
```

After, with no bridge running — an empty list, which is the honest answer and lets `omarchy hw webcam` correctly report no webcam:

```
$ omarchy-capture-webcam-list
$
```

`./test/cli` passes (116 assertions). `./test/shell` has 3 pre-existing failures on this machine, all from a missing `omarchy-pkgs` sibling checkout (`set OMARCHY_PKGS_PATH to the omarchy-pkgs checkout`) and unrelated to this change.

I don't have an IPU6 + `v4l2-relayd` machine to confirm #5722's setup still behaves, but that path is unaffected by construction: its raw node was already rejected for lacking `Video Capture`, and its `v4l2loopback` device does not set `I/O MC`.

### DIFF
--- a/bin/omarchy-capture-webcam-list
+++ b/bin/omarchy-capture-webcam-list
@@ -3,13 +3,17 @@
 # omarchy:summary=List webcam devices that support video capture
 # omarchy:hidden=true
 
+# Raw sensor nodes (Intel IPU6/IPU7 ISYS) advertise Video Capture but also set
+# I/O MC, meaning their media-controller pipeline must be configured before they
+# can stream. mpv opens devices directly, so those nodes always fail on it.
 capture_capable() {
   local device="$1"
 
   v4l2-ctl --device "$device" --info 2>/dev/null | awk '
     /^[[:space:]]*Device Caps[[:space:]]*:/ { inspect = 1; next }
     inspect && /^[[:space:]]*Video Capture/ { found = 1 }
-    END { exit !found }
+    inspect && /^[[:space:]]*I\/O MC/ { media_controller = 1 }
+    END { exit !(found && !media_controller) }
   '
 }
 


danteRub:
Follow-up: I validated the no-regression side against a normal video-node-centric capture device, since the obvious concern with this patch is whether it could drop real webcams.

Loaded the kernel's `vivid` virtual driver (`modprobe vivid n_devs=1 node_types=0x1`) alongside the IPU7 node and the loopback:

```
$ v4l2-ctl -d /dev/video32 --info      # vivid
Device Caps      : 0x05230001
        Video Capture
        Tuner
        Audio
        Read/Write
        Streaming
        Extended Pix Format
```

No `I/O MC` — and notably `vivid` *does* register a media controller node (`/dev/media1`), so having an MC device is not what the flag tracks. It tracks whether I/O routing must go through it.

Before / after, with all three present:

```
# upstream
/dev/video0   ipu7 (PCI:0000:00:05.0):
/dev/video40  Camara integrada (platform:v4l2loopback-040):
/dev/video32  vivid (platform:vivid-000):

# patched
/dev/video40  Camara integrada (platform:v4l2loopback-040):
/dev/video32  vivid (platform:vivid-000):
```

Only the unusable entry is dropped. And the filter's semantics line up with what a direct consumer can actually open:

```
$ ffmpeg -f v4l2 -i /dev/video32 -frames:v 1 out.png   # vivid, no I/O MC
(succeeds)

$ ffmpeg -f v4l2 -i /dev/video0 -frames:v 1 out.png    # IPU7, I/O MC
ioctl(VIDIOC_STREAMON): Broken pipe
```

For the `uvcvideo` case I'm going on the kernel docs and driver source rather than hardware — `uvc_ioctl_querycap` sets `V4L2_CAP_VIDEO_CAPTURE`/`V4L2_CAP_STREAMING` and never `V4L2_CAP_IO_MC`; the flag was introduced for MC-centric drivers (rcar-vin, ipu3, vimc). I don't have a USB webcam on hand to confirm empirically, so worth a second pair of eyes there.

danteRub:
Closing in favour of #7289, which is the identical commit opened from a dedicated fork (`NewMateOS/omarchy`) rather than from a repository I use for other work. Same single commit, same diff, same base branch — nothing changed but the head repo.

Sorry for the noise, and thanks for your patience.

danteRub:
Closing in favour of #7289, which is the identical commit opened from a dedicated fork (`NewMateOS/omarchy`) rather than from a repository I use for other work. Same single commit, same diff, same base branch — nothing changed but the head repo.

Sorry for the noise, and thanks for your patience.

danteRub:
Closing in favour of #7289, which is the identical commit opened from a dedicated fork (`NewMateOS/omarchy`) rather than from a repository I use for other work. Same single commit, same diff, same base branch — nothing changed but the head repo.

Sorry for the noise, and thanks for your patience.

PR #7283, opened 2026-08-17
Use layout-independent universal clipboard shortcuts

janhesters:
## Problem

Universal copy and paste can send the wrong application shortcut when the XKB configuration contains Dvorak and QWERTY layouts.

With `us(dvorak), us` and QWERTY active:

- `Super+C` opened Brave Downloads because the app received `Ctrl+J`.
- `Super+V` opened the Codex command menu because the app received `Ctrl+K`.

Chromium and Electron translate the injected letter key through Dvorak. Removing the release timer did not fix the behavior.

## Fix

Use clipboard chords without letter keys for every focused surface:

- Copy: `Ctrl+Insert`
- Paste: `Shift+Insert`
- Cut: `Shift+Delete`

Omarchy already used the Insert variants in terminals. These keys are present in each XKB layout, so the same bindings also work in Chromium, Electron, terminals, and focused layer-shell fields.

The press and release remain in one callback to avoid stuck synthetic key state.

## Validation

- Verified copy in Brave and paste in Codex with the affected Dvorak/QWERTY configuration.
- Added regression checks for all three layout-independent chords.

The full suite reached and passed the focused Hyprland test. I stopped it later because a Quickshell runtime test disrupted the live shell session.

### DIFF
--- a/default/hypr/bindings/clipboard.lua
+++ b/default/hypr/bindings/clipboard.lua
@@ -3,46 +3,19 @@
 -- layer-shell surfaces such as Omarchy panels. A virtual keyboard (wtype) won't
 -- do: the physically held SUPER merges into the injected chord at the seat.
 -- The down/up split works around Hyprland send_shortcut sometimes leaving
--- synthetic key state stuck/repeating.
+-- synthetic key state stuck/repeating. Keep both events in the same callback
+-- so Hyprland resolves them against the same keyboard device and layout. Use
+-- Insert/Delete chords so letter keys cannot translate through another XKB
+-- layout in Electron/Chromium applications.
 -- https://github.com/hyprwm/Hyprland/discussions/14099
 local function send_shortcut_once(mods, key)
   return function()
     hl.dispatch(hl.dsp.send_key_state({ mods = mods, key = key, state = "down" }))
-
-    hl.timer(function()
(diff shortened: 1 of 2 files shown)


ivantsukev:
There is a compatibility concern with making Shift+Delete the universal cut shortcut. In GNOME Files/Nautilus, selecting a file and pressing Shift+Delete requests permanent deletion, with a confirmation dialog. Mapping Super+X to that chord would change “cut” into a deletion action in this context.

Could we preserve Ctrl+X semantics for file selections, or leave the cut change out until application coverage establishes a safe replacement?

Two changes would also make this patch easier to assess:

1. Evaluate the release-timer removal separately unless it is required for this fix. If it remains, include a reproduction and evidence that repeated shortcuts leave no stuck keys or modifiers.
2. Add behavioral coverage alongside the source checks: verify clipboard contents and text changes in Chromium/Electron and GTK, terminal paste, and file-manager copy/paste/cut using disposable files. Exact-string  assertions cannot establish those outcomes.

I confirmed that a separate local keycode override fixes Super+C/V/X in both English and Bulgarian, but I have not tested this PR’s Insert/Delete approach.

Linking this proposal with #7375 and documenting the compatibility tradeoffs would help consolidate the overlapping work.

The Nautilus behavior referenced in the second comment is documented in GNOME Help 
(https://help.gnome.org/gnome-help/files-delete.html).

PR #7293, opened 2026-08-17
Trust sysfs over stale upower state for battery charge status

lytedev:
`omarchy-battery-status` reports "Holding" while the battery is actively charging on a ThinkPad with a charge threshold set.

UPower's `state` lags the kernel by tens of seconds, the same way its `energy-rate` does. The script already works around that lag for the rate, reading `power_now` from sysfs, but still takes `state` from UPower. On a T480 charging at 29W:

```
$ upower -i /org/freedesktop/UPower/devices/battery_BAT1
    state:               pending-charge
    energy-rate:         0 W
$ cat /sys/class/power_supply/BAT1/status /sys/class/power_supply/BAT1/power_now
Charging
29207000
```

The `pending-charge` branch has no rate guard, so the panel shows `Battery 75% · Holding at 75-80% · 28.7W / 87Wh` mid-charge — holding and 28.7W at the same time.

Two changes:

- Prefer `$battery_path/status` over UPower's `state` when it is readable.
- Require an idle rate for every `charge_holding` branch. Holding means no power is moving, so a stale label must not claim it while watts are flowing. The `charging` branch's now-redundant `charge_idle` check folds into that precondition.

Covered by two cases in `battery-status-test.sh`: sysfs reporting charging against a stale `pending-charge` label, and the same fixture at 0W confirming genuine holding still reports.

### DIFF
--- a/bin/omarchy-battery-status
+++ b/bin/omarchy-battery-status
@@ -62,6 +62,18 @@ power_rate=$(awk -v rate="${power_rate_raw:-0}" 'BEGIN {
   print rounded
 }')
 state=$(awk '/state/ { print $2; exit }' <<<"$battery_info")
+
+# UPower's state lags the kernel the same way its energy-rate does: a pack held
+# at a charge threshold keeps reporting pending-charge for tens of seconds after
+# it starts drawing again. Use sysfs when it is available.
+if [[ -r $battery_path/status ]]; then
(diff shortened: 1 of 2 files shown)


YuseiRun:
Was thIs closed because it will be fixed by another PR?

lytedev:
No, I just needed to close this because I could no longer follow through on this work.

PR #7295, opened 2026-08-17
Aggregate battery status across multiple batteries

lytedev:
Stacked on #7293 — take that one first; this branch contains it.

Laptops with two packs (ThinkPad T480 and friends, internal + hot-swappable) expose one `BAT*` device each. `omarchy-battery-status` takes `upower -e | grep BAT | head -n 1`, so the panel describes one pack and ignores the rest. On a T480 the bar reads:

```
Battery 80%  ·   left  ·   0W / 16Wh
```

That is the idle 16Wh internal pack, with no time estimate, while the 71Wh pack does all the work.

Every reading now combines all packs:

- **percentage** — each pack's own percentage weighted by its share of total capacity, so a small auxiliary pack cannot skew the number
- **size / rate** — summed, still preferring the live sysfs rate per pack
- **time** — derived from combined remaining energy divided by combined draw
- **state** — discharging wins, then charging, else what the packs agree on
- **cycles** — listed per pack (`348 / 82`); thresholds take the lowest across packs

The panel gains a `Battery 1` / `Battery 2` row showing each pack's size and charge, emitted only when a second pack exists — single-battery machines see the panel exactly as before, and there are tests on both sides asserting that.

One expectation moves: `battery status reports remaining time` goes `2h 30m` → `2h 37m`. Time no longer comes from UPower's `time to empty` (per-pack, and wrong once there are two) and is derived instead, which is consistent with the sysfs rate the script already prefers.

Verified on a T480 across both directions:

```
Battery 76%  ·  7h 53m left  ·   8.4W / 87Wh
Battery 78%  ·  38m to full  ·   29.8W / 87Wh   (one pack parked at its threshold, the other charging)
```

`./test/shell` passes apart from four failures that reproduce unchanged on a pristine `quattro` checkout of this machine (`bar-icon-geometry`, `config`, `snapper`, `unowned-system-paths`).

### DIFF
--- a/bin/omarchy-battery-status
+++ b/bin/omarchy-battery-status
@@ -62,6 +62,18 @@ power_rate=$(awk -v rate="${power_rate_raw:-0}" 'BEGIN {
   print rounded
 }')
 state=$(awk '/state/ { print $2; exit }' <<<"$battery_info")
+
+# UPower's state lags the kernel the same way its energy-rate does: a pack held
+# at a charge threshold keeps reporting pending-charge for tens of seconds after
+# it starts drawing again. Use sysfs when it is available.
+if [[ -r $battery_path/status ]]; then
+  case "$(<"$battery_path/status")" in
+    "Charging") state=charging ;;
+    "Discharging") state=discharging ;;
+    "Not charging") state=pending-charge ;;
+    "Full") state=fully-charged ;;
+  esac
+fi
 threshold_start=$(awk '/charge-start-threshold:/ { gsub(/%/, "", $2); print int($2); exit }' <<<"$battery_info")
 threshold_end=$(awk '/charge-end-threshold:/ { gsub(/%/, "", $2); print int($2); exit }' <<<"$battery_info")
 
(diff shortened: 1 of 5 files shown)


lytedev:
Closing this in favour of the earlier work already in flight — there are four PRs solving the same problem, and that is three more than this needs.

Overlapping, in filing order:

- #6845 (@geirolafs) — multi-battery handling in `omarchy-battery-status`, with tests
- #7138 (@fjse) — minimal combined-capacity fix in the same script
- #7186 (@unleashed-nick) — same five files as this PR, filed a day earlier and still being updated

Thanks @YuseiRun for cross-linking these.

If it is useful to whoever's work lands, two pieces of this one are worth lifting rather than rewriting:

- `batteryPacks()` in `Model.js` is split out as a pure function, so the per-pack rows are unit-testable without standing up the panel. It returns `[]` unless the script reports more than one pack, which keeps single-battery machines on exactly the previous layout.
- `test/shell.d/battery-status-test.sh` and `power-test.sh` gained fixtures for the dual-pack case, including the stale-`pending-charge`-vs-live-sysfs situation.

Happy to rebase either onto whichever PR the maintainers prefer — just say which.

To be clear about scope: #7293 is a separate fix and stays open. It addresses charge *state* being read from a stale UPower label rather than sysfs, which is independent of aggregating multiple packs and reproduces on single-battery hardware too.

Issue #7294, opened 2026-08-17
overflowing text when search item too long


Sameer292:
### System details

quattro

### What's wrong?

In the menu when the search item is too long, the text overflows in the body.

<img width="1848" height="424" alt="Image" src="https://github.com/user-attachments/assets/ca5fa4ee-fc43-4d76-99ce-835a5d28ddc9" />

Sameer292:
#7296 fixes it

PR #7296, opened 2026-08-17
feat: wrap the overflowing text in the menu

Sameer292:
Fixes the overflowing search text in the menu body when the search text is too long.

- before
 
image

- after
image


Sameer292:
#7294 this pr will fix the issue

PR #7304, opened 2026-08-17
Fix 1Password window matching regex

fadilasif:
The 1Password window matching regex currently uses `[p|P]`.

Because this is a regex character class, `|` is treated as a literal character rather than an alternation operator. This means the current pattern also matches `1|assword`.

Change `[p|P]` to `[pP]` so only the intended lowercase and uppercase forms are matched.


fadilasif:
check in

Issue #5214, opened 2026-04-04
unmount-fuse sleep hook fails to restart gvfs-daemon on resume


pbmarcano:
## Summary

The `unmount-fuse` system-sleep hook (introduced in #4940) successfully unmounts gvfsd-fuse before suspend, but fails to restart it on resume. This breaks Nautilus network file access (SMB, NFS, etc.) after every suspend/resume cycle — files are visible but can't be opened with external apps (image viewer, media player) because gvfsd-fuse is dead.

## Root cause

Two independent bugs in the `post` section:

### 1. Background subshell is killed by systemd

The hook backgrounds the restart with `( sleep 5; ... ) &`, but `systemd-suspend.service` kills all processes in its cgroup when it deactivates on resume — before the 5-second sleep completes. The restart never runs.

Confirmed via logging: `pre-suspend` and `post-resume, starting background restart` appear, but `background woke up` (after `sleep 5`) never does.

### 2. `sudo -u` fails on systems with fingerprint auth

Even if the background process survived, `sudo -u "#$uid"` goes through PAM, which triggers `pam_fprintd.so` in a non-interactive context. The fingerprint check times out, password fallback fails (no TTY), and the error is silently swallowed by `|| true`.

This affects any system with fprintd configured (ThinkPads, Framework laptops, etc.).

`runuser` was also tested as a replacement but hits similar PAM/D-Bus timeouts during resume (documented in systemd/systemd#33020).

## Suggested fix

Replace the broken background-restart-in-hook approach with a systemd service (same pattern as other post-resume services like `fprintd-resume`):

```ini
# /etc/systemd/system/gvfs-resume.service
[Unit]
Description=Restart gvfs-daemon after suspend to restore FUSE mount
After=suspend.target hibernate.target hybrid-sleep.target suspend-then-hibernate.target

[Service]
Type=oneshot
ExecStartPre=/bin/sleep 3
ExecStart=/usr/bin/systemctl --user -M 1000@ restart gvfs-daemon.service

[Install]
WantedBy=suspend.target hibernate.target hybrid-sleep.target suspend-then-hibernate.target
```

And reduce the sleep hook to only handle the `pre` unmount.

Note: `systemctl --user -M uid@` requires systemd 248+ and bypasses PAM entirely.

## Environment

- Omarchy 3.5.0
- Arch Linux, systemd 260
- Lenovo ThinkPad (Intel Meteor Lake) with Synaptics fingerprint sensor (06cb:00f9)
[...]

PR #7329, opened 2026-08-17
Re-detect external displays that no hotplug brought back after resume

### LINKED ISSUE #7328: External display stays dark after a long suspend in clamshell: nothing re-probes the connector when the dock never re-asserts HPD
### System details

- Framework Laptop 13 (AMD Ryzen AI 9 HX 370 / Radeon 890M, amdgpu), BIOS INSYDE 03.05
- Omarchy 4.0.0-1, Hyprland 0.56.2, kernel 7.1.8-arch1-3, s2idle
- External display: Dell U2723QE on `DP-5` behind a CalDigit TS3 Plus (Intel JHL6540 Alpine Ridge, TBT3)
- Clamshell: lid closed, `eDP-1` disabled by `omarchy-hyprland-monitor-clamshell`

### What's wrong?

After a long suspend in clamshell mode the external monitor never lights up again. The session itself is fine — input reaches the compositor, the idle monitor logs activity, `hyprctl` answers — but the panel stays in standby until the lid is opened or the machine is rebooted. Short suspends always come back.

### Root cause

The display comes back only when something re-probes the connector, and after a long s2idle nothing does.

1. On every resume amdgpu discards the hotplug interrupts raised inside the resume window: `[drm] Skip DMUB HPD IRQ callback in suspend/resume`.
2. Re-detection therefore depends on the dock re-asserting HPD *after* resume — `[drm] DMUB HPD RX IRQ callback: link_index=5`, normally 1–6 s after `PM: suspend exit`. That event is what makes aquamarine rescan and re-modeset.
3. After a long sleep the Thunderbolt tunnel does not fully return, and that HPD never arrives. The kernel keeps the stale pre-suspend `connected` state, so nothing looks wrong from userspace, and the compositor keeps driving a monitor that is asleep.

Every cycle from my journals, same machine, same dock:

| Sleep length | Wake | `HPD RX` after `suspend exit` | Result |
|---|---|---|---|
| 10 s | keypress | +0.0 s, +5.5 s | display returns |
| 1.2 min | keypress | +1.3 s | display returns |
| 11.7 min | keypress | +5.6 s | display returns |
| **39.8 min** | **keypress, lid stayed closed** | **none** | **black until reboot** |
| 3 h 24 min | lid opened | +5.6 s | returns, but only because the lid HPD forced the rescan |

Correlated kernel noise on the same resumes, worse the longer the sleep:

```
pcieport 0000:02:00.0: Data Link Layer Link Active not set in 100 msec
pcieport 0000:02:01.0: Data Link Layer Link Active not set in 100 msec
xhci_hcd 0000:03:00.0: Controller not ready at resume -19
xhci_hcd 0000:03:00.0: PCI post-resume error -19!
xhci_hcd 0000:03:00.0: HC died; cleaning up
```

Opening the lid recovers because the eDP hotplug triggers a full connector rescan, which re-modesets `DP-5` as a side effect.

### Steps to reproduce

1. Clamshell: lid closed, single external monitor over a TB3/USB4 dock, internal panel disabled.
2. Suspend and leave the machine asleep for ~40 minutes or more.
3. Wake it with the external keyboard, lid still closed.
4. The machine resumes — the monitor does not.

### Not Hyprland

aquamarine does re-modeset enabled outputs on session resume, but only for connectors the kernel reports as connected, and the kernel's cached state is stale rather than absent. `hyprctl reload` (what `omarchy-hyprland-monitor-modeless` falls back to) does not help either: the compositor is not the component holding the wrong state.

### What does fix it

Forcing the kernel to re-detect, which needs root:

```bash
echo off    > /sys/class/drm/card1/card1-DP-5/status
sleep 1
echo on     > /sys/class/drm/card1/card1-DP-5/status
echo detect > /sys/class/drm/card1/card1-DP-5/status
```

The monitor is back within a second. `echo detect` on its own is a no-op — the cached status does not change, so no uevent is emitted and nothing re-modesets. This is the same gap `omarchy-hyprland-monitor-watch` already comments on ("forcing a re-probe needs root").

PR with a sleep hook that does this automatically — and only when no DRM hotplug arrived with the resume — follows.

Possibly the same underlying report as #3511, though that one is about the machine not resuming at all rather than resuming dark.

paracycle:
Fixes #7328.

### Problem

An external panel behind a dock can come out of suspend with the kernel still holding its pre-suspend `connected` state while the link underneath is down. amdgpu discards hotplug interrupts raised inside the resume window (`[drm] Skip DMUB HPD IRQ callback in suspend/resume`), and a dock whose tunnel did not fully return — `pcieport: Data Link Layer Link Active not set in 100 msec`, `xhci_hcd: Controller not ready at resume -19` — never re-asserts HPD afterwards. Nothing re-probes, so the compositor keeps modesetting a monitor it believes is lit while the panel sits in standby. The session is fully alive; only the screen is gone, until the lid is opened or the machine is rebooted.

On the machine in #7328 this reproduces on every suspend longer than ~40 minutes in clamshell and never on short ones: a resume the dock handled well always logs `DMUB HPD RX IRQ callback` within ~6 s, and the failing one logs none at all.

### Approach

`omarchy-system-sleep-relink-displays` watches for the DRM hotplug uevent a healthy resume brings, and forces a re-detect when it never comes.

- `pre` empties `/run/omarchy/drm-hotplug.log` and starts `udevadm monitor --udev --subsystem-match=drm` as a transient unit. Starting the watch *before* suspend is what keeps an event landing in the same instant as the post hook from being missed — a missed event costs a needless blink. `stdbuf -oL` keeps udevadm from block-buffering into the file, where an event would sit unseen for the whole grace period.
- `post` waits 8 s, stops the watch, and returns if any `change` event was recorded. Failing that it asks for a second opinion: HPD in the kernel log since resume means the sink is answering and the link is alive, since those interrupts are DP link maintenance and do arrive on resumes where nothing re-probes the connector. Only when both signals are absent — which is what the resume that actually stranded the panel looked like — does it cycle every connected, enabled, non-internal connector `off` → `on` → `detect`.

The off/on cycle is the part that matters: writing `detect` alone leaves the cached status unchanged, so no uevent follows and nothing re-modesets. Internal panels (`eDP-*`, `LVDS-*`, `DSI-*`) are never touched — they come back with the rest of the hardware.

The hook runs the check through `systemd-run --no-block` rather than `&`. Backgrounding does not survive here: `systemd-suspend.service` takes its children down the moment the last hook returns, which is #5214 for `unmount-fuse` and would silently swallow this check too. Verified on hardware — with `&` the watch unit was still running long after resume because the child never got to stop it.

### Verification

On the reporting machine, across real suspend/resume cycles:

- Healthy resume: the transient unit starts at resume, waits out the grace period, finds two `change` events in the log, stops the watch and exits without touching anything. No blink.
- Forced no-hotplug case: the connector is cycled and the external monitor comes back within a second, with `no DRM hotplug after resume, re-detecting DP-5` in the journal.

`test/cli` covers the decision against a fixture sysfs tree: hotplug present → nothing touched; no hotplug → the external connector is re-detected while the internal panel and an unused connector are left alone. `./test/all` passes except four `test/shell.d` suites that already fail on a clean checkout here (they want an `omarchy-pkgs` checkout and other local state).

### Packaging note

`default/systemd/system-sleep/relink-displays` needs to ship from the settings package alongside `unmount-fuse`; the migration installs it on existing machines in the meantime.

### DIFF
--- a/bin/omarchy-system-sleep-relink-displays
+++ b/bin/omarchy-system-sleep-relink-displays
@@ -32,6 +32,8 @@ WATCH_UNIT=${OMARCHY_DRM_HOTPLUG_UNIT:-omarchy-drm-hotplug-watch}
 # Every DRM event the watch records belongs to this sleep cycle: the log is
 # emptied as the machine goes down and nothing plugs a monitor in while it is
 # suspended.
+# Read instead of the kernel journal when set, so the HPD check is testable.
+HPD_LOG=${OMARCHY_DRM_HPD_LOG:-}
 # Long enough for the docks observed to re-assert HPD (~6s), short enough that a
 # screen that will not come back on its own is not left dark for long.
 GRACE=${OMARCHY_DRM_RELINK_GRACE:-8}
@@ -70,6 +72,23 @@ hotplug_since_suspend() {
   grep -q "change" "$HOTPLUG_LOG"
 }
 
+# A sink that answers HPD is a sink whose link is alive: those interrupts are DP
+# link maintenance, and they arrive on resumes where the connector is never
+# re-probed at all. Treating a hotplug as the only proof of life re-detects a
+# working panel and blinks it for nothing, so this is the second opinion. The
+# resume that stranded the panel for real had neither.
+hpd_since_resume() {
+  local since="$1"
+
+  if [[ -n $HPD_LOG ]]; then
+    grep -qi "hpd" "$HPD_LOG"
+    return
+  fi
+
+  omarchy-cmd-present journalctl || return 1
+  journalctl -k --since "@$since" --no-pager 2>/dev/null | grep -qi "hpd"
+}
+
 # Panels wired to the board come back with the rest of the hardware; only the
 # ones reached over a cable can be left behind by a link that never re-trained.
 external_connectors() {
@@ -105,12 +124,20 @@ relink_connector() {
 }
 
 relink_displays() {
-  local connector
+  local connector resumed_at
 
+  # The post phase runs at resume, so this is where the window starts. A second
+  # of slack covers events logged while the hook itself was being dispatched.
+  resumed_at=$(( $(date +%s) - 1 ))
   sleep "$GRACE"
   stop_hotplug_watch
 
   if hotplug_since_suspend; then
+    return 0
+  fi
+
+  if hpd_since_resume "$resumed_at"; then
+    printf 'omarchy-system-sleep-relink-displays: HPD without a hotplug, link is alive, leaving displays alone\n'
     return 0
   fi
 
--- a/bin/omarchy-system-sleep-relink-displays
+++ b/bin/omarchy-system-sleep-relink-displays
@@ -1,0 +1,133 @@
+#!/bin/bash
+
+# omarchy:summary=Re-detect external displays that no hotplug brought back after resume
+# omarchy:group=system
+# omarchy:args=pre|post
+# omarchy:hidden=true
+# omarchy:requires-sudo=true
+
+# An external panel behind a dock can come out of suspend with the kernel still
+# holding its pre-suspend "connected" state while the link underneath is down.
+# The GPU discards hotplug interrupts raised inside the resume window (amdgpu
+# logs "Skip DMUB HPD IRQ callback in suspend/resume"), and a dock whose tunnel
+# did not fully return -- "Data Link Layer Link Active not set", xHCI resuming
+# with -19 -- never re-asserts HPD afterwards. Nothing re-probes, so the
+# compositor keeps modesetting a monitor it believes is lit while the panel sits
+# in standby, and only a genuine hotplug (opening the lid, replugging the cable)
+# brings the picture back.
+#
+# A resume the dock handled well always shows a DRM hotplug uevent within
+# seconds, so that event is the signal to wait for. The watch starts before
+# suspend: one landing in the same instant the post hook runs would otherwise be
+# missed, and a missed event costs a needless blink. When none arrives, force
+# the connector off and back on. That is the one write that makes the kernel
+# re-detect and re-train the link -- writing "detect" alone leaves the cached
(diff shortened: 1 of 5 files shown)


paracycle:
Pushed a follow-up commit after the hook over-triggered on my own machine.

`no DRM uevent` alone is not enough evidence that a panel was stranded. HPD RX interrupts are DP link maintenance and arrive on resumes where the connector is never re-probed at all, so the first version re-detected a perfectly live link and blinked it for ~1s:

```
22:46:50.594  amdgpu: DMUB HPD RX IRQ callback: link_index=5
22:46:51.390  amdgpu: DMUB HPD RX IRQ callback: link_index=5
22:46:58.647  relink: no DRM hotplug after resume, re-detecting card1-DP-5   ← needless
```

Across every resume I have logs for, the split is clean — the resume that actually stranded the panel had *neither* signal:

| resume | HPD | DRM uevent | correct action |
| --- | --- | --- | --- |
| 40 min sleep, screen dark until reboot | none | none | re-detect |
| 13.8 h sleep | +5.9 s | none | ambiguous |
| 6 s sleep | +0.0 s | none | leave alone |
| short sleeps | +0.0 s | present | leave alone |

So the check now needs both absent before it touches anything, and says so when it stands down:

```
22:52:43  relink: HPD without a hotplug, link is alive, leaving displays alone
```

Verified on hardware over a real suspend/resume; `test/cli` gains a case for the HPD-only path.

Unrelated to the display half but worth recording here: the same machine turned out to be dropping the *input* half of the same dock on nearly every resume, because `etc/modprobe.d/omarchy-usb-autosuspend.conf` never applies (`usbcore` is built into the kernel — #8097). With autosuspend disabled at runtime the tunneled hub rode through suspend intact for the first time. Different bug, same dock, and it produces an identical-looking black screen — worth keeping the two apart when triaging reports.

PR #7333, opened 2026-08-18
Reset Apple BCM4350/BCM43602 Wi-Fi around sleep

Rouxgrr:
## Summary

- Install a sleep service only on Apple systems with BCM4350 (`14e4:43a3`) or BCM43602 (`14e4:43ba` / `43bb` / `43bc`).
- Unbind that device from `brcmfmac` before sleep and bind it again after wake, forcing a clean firmware initialization.
- Add an idempotent migration for existing installs and tests for hardware gating, helper recovery, and installation.

This generalizes the approach in #7180, which already proved the unbind/bind cycle on BCM4350, so one service can cover both confirmed S3-dead chips. T2-era parts stay out of the allowlist: BCM4364 must remain bound, and BCM4377/4378/4387 have their own recovery path.

## Failure

Reproduced on a MacBookPro12,1 (Early 2015 13") running Omarchy 4.0.0-1, kernel 7.1.8, in-tree `brcmfmac` for BCM43602 (`14e4:43ba`, firmware `7.35.177.61`).

Closing the lid suspends (`deep` S3). After resume, the Wi-Fi toggle stays on and the network panel only shows DNS. The radio is enabled, but the firmware is dead:

```
platform-linux: do-change-link[2]: failure 5 (Input/output error)
device (wlp3s0): set-hw-addr: failed to set MAC address ... (scanning) (NME_UNSPEC)
device (wlp3s0): Couldn't initialize supplicant interface: Timeout was reached
ieee80211 phy0: brcmf_msgbuf_tx_ioctl: Failed to reserve space in commonring
ieee80211 phy0: brcmf_cfg80211_scan: scan error (-12)
```

Toggling the slider only flips rfkill/NetworkManager state. A reboot reinitializes the firmware and Wi-Fi works again. Same machine, same failure, on Omarchy 3 — this is not leftover Quattro UI.

#7180 recorded the same class of failure on MacBook10,1 / BCM4350, including a later suspend that then timed out in `brcmf_pcie_pm_enter_D3`.

## Approach

A `sleep.target` oneshot keeps the matching PCI device unbound for the sleep transaction, then binds it after wake so NetworkManager sees a freshly initialized radio. The helper also records an already-unbound matching device so a leftover cycle cannot block the next suspend.

Module unload was discarded in #7180 because udev immediately reloaded `brcmfmac` and an exploratory cycle coincided with a hard lock. This PR uses that same unbind/bind path.

The installer and migration no-op unless the machine is Apple and `lspci` shows one of the confirmed IDs.

## Testing

- `bash -n` on the helper, installer, migration, and test
- `test/shell.d/brcmfmac-suspend-test.sh`
- `test/shell.d/brcmfmac-supplicant-test.sh`
- `test/cli` (including command metadata for the new helper)

Live lid-close/resume of this service still needs a run on the MacBookPro12,1 that produced the logs. #7180 already completed that cycle on BCM4350 with the same unit ordering.

If this should land after #7180 instead of beside it, the BCM43602 IDs can be added there and this closed.

### DIFF
--- a/bin/omarchy-hw-brcmfmac-suspend
+++ b/bin/omarchy-hw-brcmfmac-suspend
@@ -1,0 +1,90 @@
+#!/bin/bash
+
+# omarchy:hidden=true
+# omarchy:summary=Unbind affected Broadcom Wi-Fi around system sleep
+# omarchy:args=<pre|post>
+# omarchy:requires-sudo=true
+
+set -euo pipefail
+
+# Apple BCM4350 (43a3) and BCM43602 (43ba/43bb/43bc) firmware often stops
+# answering after S3. NetworkManager still shows the radio as on, but scans
+# fail and the supplicant times out until the device is rebound. Do not add
+# T2-era parts: BCM4364 (4464) must stay bound for PCI PM, and
+# BCM4377/4378/4387 have their own T2 recovery path.
+affected_ids="${OMARCHY_BRCMFMAC_SUSPEND_IDS:-0x43a3 0x43ba 0x43bb 0x43bc}"
+driver_path="${OMARCHY_BRCMFMAC_DRIVER_PATH:-/sys/bus/pci/drivers/brcmfmac}"
+device_root="${OMARCHY_BRCMFMAC_DEVICE_ROOT:-/sys/bus/pci/devices}"
+state_file="${OMARCHY_BRCMFMAC_SUSPEND_STATE:-/run/omarchy-brcmfmac-suspend.devices}"
+
+is_affected_id() {
+  local device_id="$1" candidate
+  for candidate in $affected_ids; do
+    [[ $device_id == "$candidate" ]] && return 0
+  done
+  return 1
+}
+
+driver_name() {
+  local pci_address="$1"
+  local link="$device_root/$pci_address/driver"
+
+  [[ -L $link ]] || return 0
+  basename "$(readlink -f "$link")"
+}
+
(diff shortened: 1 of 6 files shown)


Rouxgrr:
Live lid-close/resume on the MacBookPro12,1 that produced the original failure (BCM43602 `14e4:43ba`, Omarchy 4.0.0-1, kernel 7.1.8). The service is installed locally from this branch (`ExecStart`/`ExecStop` pointed at `/usr/local/bin` until the package ships the helper).

Ordering:

```
18:28:32 Starting Reset Broadcom Wi-Fi around sleep...
18:28:33 Finished Reset Broadcom Wi-Fi around sleep
18:28:33 Starting System Suspend...
18:28:40 System returned from sleep operation 'suspend'
18:28:40 Stopping Reset Broadcom Wi-Fi around sleep...
18:28:40 rfkill2: found Wi-Fi radio killswitch ... (driver brcmfmac)
18:28:40 device (wlan0): interface index 4 renamed iface from 'wlan0' to 'wlp3s0'
18:28:44 Connected to wireless network "Blazin"
18:28:46 device (wlp3s0): Activation: successful, device activated
```

About six seconds from wake to an activated link. Zero `Input/output error`, zero supplicant timeouts, zero `brcmf_msgbuf_tx_ioctl` / scan `-12` after resume. The panel is empty for that window because the phy is being rebound, then the network comes back on its own.

Issue #2082, opened 2025-09-29
Using vim style navigation in omarchy-menu


eshedruf:
### What do you need?

Currently, the only way to move up and down in the omarchy-menu is by using the arrow keys. I thought about implementing a way to also navigate the menu using Ctrl+J and Ctrl+K, in addition to the arrow keys. I’ve seen this in a few other programs (for example, fzf) that have it built in.

Since omarchy-menu uses [walker](https://github.com/abenz1267/walker), I checked for an implementation there, but it turns out the maintainer is not a big fan of this feature: [https://github.com/abenz1267/walker/issues/470](https://github.com/abenz1267/walker/issues/470). So I thought maybe a small patch in omarchy could be nice.

BBennettChiba:
Is there a reason why rebinding the walker config keys doesn't work? I've tried to remap ctrl j and ctrl k to up and down but I've had no luck.

eshedruf:
> Is there a reason why rebinding the walker config keys doesn't work? I've tried to remap ctrl j and ctrl k to up and down but I've had no luck. 

Yes, I am pretty sure it is due to [this](https://github.com/abenz1267/walker/blob/master/src/keybinds.rs#L156-L221) keybindings parsing function, which takes a single binding.

joeonit:
I come from macOS where I used Karabiner-Elements' vim system-wide navigation plugin. I've managed to do this functionality in Omarchy using keyd, and it works great either in omarchy-menu or any other app!

1. install keyd
1. Create a config file at /etc/keyd/default.conf
1. Add this configuration (i'm using Right Alt + hjkl):  
```
[ids]
*

[main]
# Map Right Alt + hjkl to arrow keys when held
rightalt = layer(navigation)

[navigation:C]
h = left
j = down
k = up
l = right
```
 Restart keyd

Xamnalim:
I prefer ctrl + n for next item and ctrl + p for previous.
I successfully use these shortcuts using the following config:
```toml
[keys]
next = ["ctrl n"]
prev = ["ctrl p"]
```
The arrow keys still work.

Also have you seen the "[Keybinds](https://github.com/abenz1267/walker/tree/0.13.26?tab=readme-ov-file#keybinds)" section of walker readme? It says that ctrl + j and ctrl + k should work by default, if ActivationMode is disabled.

ryan-truman:
I've just been struggling with this for a bit.

```
[keybinds]
quick_activate = []
[...]

Issue #1658, opened 2025-09-14
Add vim-style menu navigation


makcakaya:
### What do you need?

Having vim-style menu navigation in Application Launcher and Omarchy Menu would be nice. I am not sure if Application Launcher can have this though would be nice to scroll through applications with vim-style keybindings but Omarchy Menu would definitely benefit from it.

Ctrl+n = go to next item
Ctrl+p = go to previous item
Ctrl+u = go one page up (or the first item)
Ctrl+d = go one page down (or the last item)

Edit: Also to go to the outer menu Ctrl+h or another keybinding can be useful too.

MaxFaucon:
~~Hey @makcakaya, I don't know if it'll help you but waiting for this issue to be resolved, you can add your mappings directly in the Application Launcher config file located here: `~/.config/walker/config.toml`.
Following [Walker documentation](https://walkerlauncher.com/docs/keys#example-vim-like-navigation), you can easily add vim-style menu navigation.~~

~~On my end, I simply needed the basic ctrl-n//ctrl-p which I added with the following section:~~
```
Outdated
[keys]
next = ["ctrl n"]
prev = ["ctrl p"]
```

~~Hope it'll help someone !~~

---

**Update:**

This workaround no longer works with latest Walker version. Here's what you can do now (still in the `~/.config/walker/config.toml` file):
1. Use the new vim-style keybindings recommended [by the documentation](https://walkerlauncher.com/docs/keybindings#vim-like-navigation):
```
[keybinds]
next = ["Down", "ctrl j"]
previous = ["Up", "ctrl k"]
```
2. If you want to keep using `ctrl n` and `ctrl p`, you'll also need to rebind pin actions as they use these keys by default (cf https://walkerlauncher.com/docs/provider-actions#provider-specific-actions):
```
[keybinds]
next = ["Down", "ctrl n"]
previous = ["Up", "ctrl p"]

...

[providers.actions]
desktopapplications = [
  { action = "pin", bind = "ctrl shift p", after = "AsyncReload" },
  { action = "unpin", bind = "ctrl shift p", after = "AsyncReload" },
  { action = "pinup", bind = "ctrl shift n", after = "AsyncReload" },
]
```
I also tried rebinding with empty strings, but it doesn't seem to work.

PR #7345, opened 2026-08-18
Let users rebind the menu's navigation keys

BrouZie:
The menu's navigation keys are a hardcoded `if`/`else` chain today. This moves them into the menu's JSONC so a user can add their own without touching the shell.

Nothing changes by default. The shipped bindings are the same keys the chain had, and I replayed the old dispatch against the new one across every key, modifier and filter-state combination to confirm there's no behavioural difference. Escape and Delete stay hardcoded, so a bad config can't lock anyone out.

Opting in are a few lines in the menu extension file:

```jsonc
"keybindings": {
  "next":     ["CTRL + J", "CTRL + N"],
  "prev":     ["CTRL + K", "CTRL + P"],
  "activate": ["CTRL + L"],
  "back":     ["CTRL + H"],
},
```

Listing keys adds to the shipped ones, `[]` removes them - the same contract as `config/hypr/bindings.lua`, and the binding strings use its format too (`"SUPER + SHIFT + R"`).

On why this rather than a personal config: it's come up a few times ([#2082](https://github.com/omacom/omarchy/issues/2082), [#1658](https://github.com/omacom/omarchy/issues/1658)), and the answer has been to change it yourself, which worked through `walker/config.toml` until Walker was removed in `abf0a68b`. The Quickshell menu has no equivalent, so the only route now is cloning the whole menu plugin.

### Testing

The implementation is ~270 lines across `MenuModel.js` and `Menu.qml`; the rest of the diff is tests and documentation.

`menu-test.sh` and `menu-guards-test.sh` pass. `bar-icon-geometry` and `theme-install-guards` fail identically on `quattro` - the first is a rendering assertion against the live compositor, the second can't simulate a missing `omarchy-git-url-check` on a machine where it's on `$PATH`.

#7042, #7513, #7737 and #8318 are all attempts at this from the last few weeks. I'd be happy to see any of them land, not this one specifically - and glad to narrow or rework mine if it's close, but not quite right.


anuragkumar2921:
Super excited for this !!!

DanielSRojo:
Ctrl+p/n is such a standard in cli and tui tools that I'm just too used to them. Ctrl+j/k are great too, but not so common everywhere.

BrouZie:
Very true - ctrl + p/n get baked into muscle memory. I sometimes open the print menu in my browser because I subconsciously try scrolling up with ctrl + p. But yeah, j and k can be nice supplementary binds:)

relston:
This is the way. We want a completely configurable experience, right? No sense in having hard coded keybindings in the primary menu UI we need to do everything.

Edgarfp:
This would be great!

itsgg:
I maintain omarchy-readline-keys, a third-party plugin that added Ctrl+N/P/F/B to the menu, clipboard, emoji and image picker popups. The 4.0.3 capability facade rightly stops a plugin from attaching keys to another plugin's popup, so I am withdrawing it from the marketplace, and this is where the feature belongs.

For the menu, this PR covers almost everything the plugin did. At 2a625fb0, `keybindings-menu-test.sh`, `menu-test.sh` and `menu-guards-test.sh` pass. With `next`, `prev`, `activate` and `back` bound to CTRL + N, P, F and B in the user extension, `bindingMatches` fires all four with an empty filter and while typing in it, and Down still moves. CTRL + M binds to `activate` as well. The one thing the plugin had that this does not cover is Ctrl+[ as Escape: Escape stays hardcoded here, and `[` is not a key name the parser accepts. Given the reason Escape is hardcoded, I would leave that out.

The same list navigation exists in clipboard, emojis and the image picker, which this does not reach. #7737's `Keymap.js` is written for any list panel, which makes it a natural shared layer for those three. Whichever of the two designs lands, I am happy to follow up with a PR that brings the other three popups onto the same bindings, so there is one mechanism rather than four.

PR #7367, opened 2026-08-18
Add a notifications.position setting for centered toasts

jankeesvw:
Toasts are anchored to the top-right corner, which is hardcoded. On a wide screen that corner is far outside where you are actually looking: I run a 6016px display and kept missing notifications entirely, including ones I had set up myself and was waiting for.

This adds `notifications.position` to `shell.json`, accepting `right` (the default, unchanged) or `center`:

```json
{ "notifications": { "position": "center" } }
```

`popupPlacement()` gains an `alignment` argument and returns a `horizontalCenter` anchor instead of `right` when centered. It also drops the right-bar clearance in that case, since a centered column is not held off any side edge. Anyone who does not set it sees no change.

**Before** — right-aligned, on a 6016px screen

before-right-aligned

**After** — `"position": "center"`

after-centered

## Testing

`test/shell.d/notifications-test.sh` passes (111 checks). The four existing placement assertions were updated for the new anchors field, and three were added: centered with a top bar, centered with a right bar, and an explicit `right` matching the default.

### DIFF
--- a/config/omarchy/shell.json
+++ b/config/omarchy/shell.json
@@ -4,6 +4,9 @@
     "screensaver": 150,
     "lock": 300
   },
+  "notifications": {
+    "position": "right"
+  },
   "bar": {
     "position": "top",
     "transparent": false,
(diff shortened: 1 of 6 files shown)


jankeesvw:
@dhh is this something you would consider adding? If so, I can fix the comments by @copilot

bjarneo:
Automated duplication check: this pull request looks similar to #10278, which covers the same notification position setting. I keep that one open and close this one to consolidate review.

PR #7200, opened 2026-08-17
Add a Grok collector for the agents usage panel

unleashed-nick:
Towards https://github.com/basecamp/omarchy/discussions/7198.

Setup already offers Grok as a first-party default coding agent, but `omarchy.agents` never lights up for it. Collectors are discovered from `$OMARCHY_PATH/bin/omarchy-agent-usage-*`; there was no Grok one.

This adds `omarchy-agent-usage-grok`. It prints the same record contract the panel already understands — no QML changes.

- **Local token stats** come from `$GROK_HOME/sessions` (default `~/.grok/sessions/<cwd>/<id>/`). Grok writes billed usage on `updates.jsonl` `turn_completed` rows (nested `usage`, or flat token fields / `event_name`), not `events.jsonl` / `chat_history.jsonl`. Cache read is split out of input so the four panel buckets stay mutually exclusive. `session_kind` values that start with `subagent` (`subagent` on Grok 1.0.x, `subagent_fork` on older builds) are skipped so parent totals are not counted twice.
- **Limits and plan** come from Grok ACP `_x.ai/billing` (`grok agent --no-leader stdio`), the same source `/usage` uses. A SuperGrok weekly (or monthly) pool becomes `limits[]`. Leftover prepaid / on-demand credits become a `balance` only when the ledger is actually non-zero — a weekly pool with a $0 prepaid ledger is a meter, not `$0.00 remaining`. Plan label prefers the display string (`SuperGrok Heavy`); `SuperGrokPro` still rewrites to `SuperGrok Pro`.
- `assets/grok.svg` (and a `-light` twin) use the official Grok mark from [grok.com/images/favicon.svg](https://grok.com/images/favicon.svg), the same glyph as the Omarchy icon font. The bar glyph still stands in without them.

Verified on Grok 1.0.5 against live sessions: the panel gained a Grok tab with **SuperGrok Heavy**, a **Weekly 16%** meter, tokens by day, and tokens by model (`grok-4.6-build`).

grok-panel

### Tests

`test/shell.d/agent-usage-grok-scanner-test.sh` — nested and flat turn ledgers, cache split, prompt_id dedupe, subagent skip, empty install, weekly/monthly mapping, zero prepaid omit, leftover credits, missing percent, expired auth, mise/ACP noise.

### Manual

`manual/17-ai.md` now lists Grok with the other first-party collectors and notes the SuperGrok weekly pool plus local token stats.

PR #6902, opened 2026-08-14
Add Grok Build usage to the agents panel

csfh:
## Summary

Adds a Grok Build collector so the agents panel can show SuperGrok weekly usage, extra credits, and local token history.

`omarchy-agent-usage-update` already globs `omarchy-agent-usage-*`. This ships `bin/omarchy-agent-usage-grok`; the panel picks up the record with no QML changes.

## Change

- Local stats come from documented Grok Build sessions under `$GROK_HOME` (default `~/.grok`): each `turn_completed` event in `updates.jsonl` is one prompt (nested `usage`, flat token fields, or `event_name`). Cache read is split out of input. `session_kind` values that start with `subagent` (`subagent` on Grok 1.0.x, `subagent_fork` on older builds) are skipped so parent totals are not counted twice.
- pi/omp sessions on an `xai` / `xai-auth` provider, and OpenCode messages with `providerID` `xai`, are merged in the same way Claude and Codex pick up those agents. Other providers, non-assistant rows, prefix-colliding ids such as `xai-proxy`, and non-object JSONL / malformed OpenCode rows are ignored. An interrupted OpenCode scan is not cached.
- Limits come from cli-chat-proxy `/v1/billing?format=credits` (weekly or monthly pool as `limits[]`, prepaid / leftover on-demand credits as `balance`). Plan label prefers `/v1/settings` `subscription_tier_display` (the string Grok's UI shows) and falls back to `/v1/user` `subscriptionTier` (`SuperGrokPro` → `SuperGrok Pro`).
- `creditUsagePercent` is read as 0-100, matching `productUsage`. A retry is advised only on a transport miss; an HTTP status means the endpoint answered, so the panel's 30-second retry is not set.
- Missing or expired `auth.json` keeps local stats and asks for `grok login`. The token is never written into the usage record.
- Incremental scan cache under `~/.cache/omarchy/agent-usage/`. `--limits-only` reuses any same-day scan so opening the panel refreshes the weekly percent without walking transcripts; `--force` rereads the files. An unwritable cache still prints a complete record.
- Optional `assets/grok.svg` / `grok-light.svg` mark (from the official Grok favicon paths). Manifest and manual list Grok Build next to Claude, Codex, and Fireworks.

Related: #6477 also adds a grok collector (with Kimi). This PR is Grok Build only, and it counts billed `turn_completed` usage rather than peak `_meta.totalTokens` context size.

## Testing

```bash
./test/shell.d/agent-usage-grok-scanner-test.sh
[...]

PR #6477, opened 2026-08-01
Add Kimi Code and Grok Build usage collectors

dalmasluca:
## Summary

Adds official quota and per-model local token history for **Kimi Code** and **Grok Build** to the `omarchy.agents` plugin, as two standalone collectors that print one display-ready record each:

- `bin/omarchy-agent-usage-kimi`
- `bin/omarchy-agent-usage-grok`

Both follow the exact shape of `bin/omarchy-agent-usage-claude` / `codex` (alone with the record schema, caches, and CLI flags `--force` / `--limits-only` / `--local-only`), and are picked up automatically by `bin/omarchy-agent-usage-update` by file-name globbing. No QML changes: the panel discovers a provider from its record file and resolves `assets/<id>.svg`.

- **Official quota**
  - Kimi: `~/.kimi-code` / `KIMI_CODE_HOME` credentials + the `/usages` endpoint (weekly `usage` summary row + rolling windows under `limits[]`, e.g. 5h), labels and fractions from the CLI's own rendering. Token refresh uses the CLI's OAuth flow under an advisory flock so the CLI and collector never rotate the same refresh token twice.
  - Grok: `~/.grok/auth.json` grok-com entry + `billing?format=credits` (weekly/monthly credit pool, per-product breakdown, on-demand cap), with CLI-style token refresh under `auth.json.lock`.
- **Local per-model token history** (merged into one record, no double counting)
  - Kimi native transcripts (`~/.kimi-code/sessions/*/session_*/agents/*/wire.jsonl`, per-turn `usage.record`); model aliases `kimi-code/k3`, `kimi-k3`, `opencode-go/kimi-k3` collapse into one `k3` bucket.
  - Grok native sessions (`~/.grok/sessions`, `turn_started` → model map + peak cumulative `_meta.totalTokens` per turn from `updates.jsonl`).
  - pi/omp transcripts (`~/.pi/agent/sessions`, `~/.omp/agent/sessions`) for the exact `kimi-coding` / `xai` providers.
  - opencode database sessions with `providerID` exactly `kimi` / `xai` (reasoning tokens kept out of input, matching the Anthropic collector).
- Plugin wiring: `manifest.json` adds `kimi` and `grok` as enabled-by-default providers; `assets/kimi.svg`, `kimi-light.svg`, `grok.svg`, `grok-light.svg` are copied from the old model-usage plugin.
[...]

PR #7393, opened 2026-08-18
Add a Grok usage collector

fuziontech:
## What

Adds `omarchy-agent-usage-grok` so the existing `omarchy.agents` panel grows a Grok tab the same way it already does for Claude, Codex, and Fireworks. No QML change: the panel draws any record that lands in `~/.local/state/omarchy/agents/usage/`.

## How

- **Local stats** come from documented Grok sessions under `$GROK_HOME` (default `~/.grok`). Each `turn_completed` event in `updates.jsonl` is one prompt. Cache reads are split out of `inputTokens` so the four panel buckets stay mutually exclusive. `session_kind: subagent_fork` sessions are skipped so parent totals are not counted twice.
- **Limits and plan** come from the same cli-chat-proxy endpoints Grok's `/usage` command uses (`/v1/billing?format=credits` and `/v1/user?include=subscription`). The collector only reads the OIDC token already stored in `~/.grok/auth.json`. It never writes credentials and does not refresh them — that stays with `grok login`.
- A SuperGrok weekly pool with a **zero prepaid ledger** is reported as a limit, not `$0.00` remaining. Prepaid / leftover on-demand credits still become a `balance` when the billing payload actually has them.

Verified on a live SuperGrok Plus login (Grok 1.0.5): the panel gained a Grok tab with **SuperGrok Plus**, a **Weekly 31%** meter, tokens by day, and tokens by model (`grok-4.6-build`). `creditUsagePercent` is present on this version.

Related collectors already in review: #6902, #6485, #7200, #6477. Happy to close this in favor of whichever lands.

## Tests

```bash
./test/shell.d/agent-usage-grok-scanner-test.sh
./test/shell.d/agents-panel-test.sh
```

Covers empty/unauthenticated records, per-turn summing and dedupe, subagent fork skip, `GROK_HOME`, weekly/on-demand/balance mapping, SuperGrok Plus labeling, a zero SuperGrok prepaid ledger, expired tokens, `--limits-only` cache reuse, and transport vs HTTP error handling.


spencerbull:
Closing in favor of #6485, which implements the same Grok usage collector and has the active maintainer discussion, post-refactor implementation, and broader validation history. Please carry over any unique test coverage or edge cases from this PR if #6485 does not already cover them.

PR #7394, opened 2026-08-18
Add a Kimi usage collector

fuziontech:
## What

Adds `omarchy-agent-usage-kimi` so the existing `omarchy.agents` panel grows a Kimi tab the same way it already does for Claude, Codex, and Fireworks. No QML change: the panel draws any record that lands in `~/.local/state/omarchy/agents/usage/`.

## How

- **Local stats** come from documented Kimi Code session wires under `$KIMI_CODE_HOME` (default `~/.kimi-code`). Each `usage.record` with turn scope is one prompt. Session-scope rows are running totals and are skipped. Cache read/write stay out of `inputTokens`.
- **Limits and plan** come from the same `api.kimi.com/coding/v1/usages` payload Kimi's `/usage` command renders: weekly quota plus the 5-hour session window. Membership level becomes the plan line (`LEVEL_ALLEGRETO` → `Allegreto`).
- **Auth**: `KIMI_API_KEY` wins, then `~/.kimi-code/credentials/kimi-code.json`, then the legacy `~/.kimi` store. Expired OAuth tokens are refreshed the way the CLI does (form POST to `auth.kimi.com`, flock, persist the rotated single-use refresh token at mode 0600). Tokens are sent only to `kimi.com` over HTTPS; redirects are refused.
- Ships the official Kimi K-mark plus a light-surface twin, and enables `kimi` in the widget defaults.

The tab stays hidden until a machine has a signed-in Kimi account or recorded session usage — same self-hiding as the other collectors.

Related: #7261 (make Kimi a default agent), #7274, #6477. Happy to close this in favor of whichever lands.

## Tests

```bash
./test/shell.d/agent-usage-kimi-scanner-test.sh
```

Covers empty/unauthenticated records, turn vs session scope, `KIMI_CODE_HOME`, env-key precedence, kimi-code vs legacy store, refresh rotation, endpoint trust, weekly/5-hour payload mapping, and `--limits-only` cache reuse.

Live-verified on this machine against Kimi Code 0.37.1 (`managed:kimi-code` OAuth, Standard plan). The collector read `api.kimi.com/coding/v1/usages` and the stock `omarchy.agents` panel grew a Kimi tab with **Standard**, a **Weekly 5%** meter, and a **Session (5-hour) 7%** meter. Tokens-by-day stayed empty because this account has not yet written a `usage.record` session wire.


bjarneo:
Automated duplication check: this pull request looks similar to #7274, which covers the same Kimi usage collector. I keep that one open and close this one to consolidate review.

PR #7398, opened 2026-08-18
Resolve package-backed OMARCHY_PATH symlinks

### LINKED ISSUE #7390: omarchy update aborts after the Quattro upgrade: six scripts string-compare OMARCHY_PATH against /usr/share/omarchy, defeating the upgrader's own compat symlink
## What happened

After upgrading to Quattro (4.0.0-1, package-backed install), `omarchy update` aborted at the very first step:

```
Create system snapshot
Error: OMARCHY_PATH is not a git checkout: /home/<user>/.local/share/omarchy

```

The install is entirely healthy — `omarchy` and `omarchy-settings` 4.0.0-1 are installed from pacman, `/usr/share/omarchy` is intact, and `~/.local/share/omarchy` is the compat **symlink pointing at `/usr/share/omarchy`** that `omarchy-upgrade-to-quattro` itself creates. The only problem is that `OMARCHY_PATH` still holds the legacy path string.

## Root cause

Six scripts decide "is this a dev checkout?" by comparing `$OMARCHY_PATH` as a **string** against `/usr/share/omarchy`, without resolving symlinks:

| Script | Line |
|---|---|
| `omarchy-update-dev` | 7 |
| `omarchy-version` | 8 |
| `omarchy-channel-current` | 10 |
| `omarchy-channel-set` | 75 |
| `omarchy-version-branch` | 9 |
| `omarchy-update-available` | 9 |

```bash
# omarchy-update-dev:7
[[ $OMARCHY_PATH != "/usr/share/omarchy" ]] || exit 0
```

This directly defeats a shim the upgrader installs on purpose. From `omarchy-upgrade-to-quattro:1017`:

> Existing terminals still have `OMARCHY_PATH=$HOME/.local/share/omarchy` and often have `$OMARCHY_PATH/bin` early in PATH. Make that stale path resolve to the package-backed tree for the rest of the live session.

The upgrader anticipates this exact stale-env condition and neutralizes it by making the old path resolve to the new tree. The string comparison ignores that resolution, so `~/.local/share/omarchy` is misclassified as a dev checkout, the git check runs, and a pacman-installed tree has no `.git`.

Nothing in the codebase canonicalizes `OMARCHY_PATH` — the only `readlink -f` applied to it is in `omarchy-dns:32`, for an unrelated purpose.

## Impact

The session environment is **not** affected: `/usr/share/uwsm/env.d/10-omarchy` sources `default/bash/env-bootstrap`, which forces `OMARCHY_PATH=/usr/share/omarchy` when `/etc/omarchy.conf` is absent. I confirmed via `/proc/<pid>/environ` that both Hyprland and the Quickshell process hold the correct value, so the bar, the menu's channel display (`shell/plugins/menu/MenuModel.js`), and anything launched from the session are all fine.

The bug is reachable only from a shell whose environment predates that session env or overrides it afterwards:

- **A terminal opened before the upgrade**, in any shell — precisely the case the `:1017` comment describes. Persists until it is closed.
- **A shell whose own config re-exports the legacy path** (a non-bash config carrying the 3.x-era `OMARCHY_PATH`). `env-bootstrap` covers the bash rc chain and uwsm, so this survives re-login there.

Inside such a shell, verified on this machine with identical binaries and only the env var differing:

| Command | stale `OMARCHY_PATH` | resolved | Consequence |
|---|---|---|---|
| `omarchy-update-dev` | **`Error: ... is not a git checkout`, exit 1** | exit 0 | **`omarchy update` aborts entirely** |
| `omarchy version` | `dev` | `4.0.0-1` | version unreportable from that shell |
| `omarchy-channel-current` | `dev` | `stable` | wrong channel from that shell |
| `omarchy-update-available` | up to date | up to date | benign, degrades gracefully |

The aborted update is the consequence that matters; the misreports are confined to the affected shell and never reach the UI. `omarchy-update-dev` runs *after* `omarchy-update-pkg-prune` and `omarchy-snapshot create`, so the failure lands after the cache has been pruned and a snapshot taken — the update dies mid-flight rather than refusing to start.

## Steps to reproduce

No unusual configuration required — a stale exported value is enough:

```bash
env OMARCHY_PATH="$HOME/.local/share/omarchy" bash -c 'omarchy-update-dev; echo "exit=$?"'
# Error: OMARCHY_PATH is not a git checkout: /home/<user>/.local/share/omarchy
# exit=1

env OMARCHY_PATH="$HOME/.local/share/omarchy" bash -c 'omarchy version'
# dev        (expected: 4.0.0-1)

readlink -f ~/.local/share/omarchy
# /usr/share/omarchy      <- the shim is working; the comparison ignores it
```

**This is not shell-specific** — it reproduces under plain `bash`, as above, in any terminal still carrying the pre-upgrade export.

In my case the login shell is fish, and a `config.fish` export carried the 3.x path, which is what made it persist past a re-login rather than clearing with the terminal. To be clear, the upgrader migrating `~/.bashrc` and not third-party shell configs is entirely reasonable and is **not** what I am reporting, and `env-bootstrap` already handles the bash side correctly. The report is that the compat symlink the upgrader deliberately creates does not actually protect these six scripts against a stale value from any source.

## What was expected

`OMARCHY_PATH` pointing at a symlink that resolves to `/usr/share/omarchy` should be treated as the package-backed install it is, honoring the upgrader's shim.

## Suggested fix

Resolve before comparing:

```bash
[[ $(readlink -f "$OMARCHY_PATH") != /usr/share/omarchy ]] || exit 0
```

Costs nothing for genuine dev checkouts, which are not symlinked to the package tree. Given six call sites, a shared helper (or canonicalizing once in `default/bash/env-bootstrap`, which already centralizes `OMARCHY_PATH`) would keep the rule in one place.

## System

- Omarchy 4.0.0-1 (`omarchy` and `omarchy-settings`), upgraded from 3.8.2 via `omarchy-upgrade-to-quattro`
- Channel: stable
- Hyprland 0.56.2
- Kernel 7.1.8-arch1-3

---

mdenesfe:
Fixes #7390.

After upgrading to Quattro, existing terminals can retain the legacy
`OMARCHY_PATH=$HOME/.local/share/omarchy` value. The upgrader keeps this path
working as a symlink to `/usr/share/omarchy`, but six commands compared the
unresolved string directly and incorrectly treated it as a dev checkout.

This resolves `OMARCHY_PATH` before classifying the installation while keeping
the original configured path for actual git operations. If path resolution
fails, the existing behavior is preserved.

Regression coverage verifies the legacy package-backed symlink behavior for:

- dev checkout updates
- version and branch reporting
- channel detection and switching
- update availability checks

Tested with:

- `bash test/shell.d/update-dev-test.sh`
- `bash test/shell.d/version-test.sh`
- `bash test/shell.d/channel-test.sh`
- Bash syntax checks for all changed scripts and tests
- `git diff --check`

The `update-available` regression case passes; the remainder of that test file
requires GNU `timeout`, which is unavailable in the macOS test environment.

### DIFF
--- a/bin/omarchy-channel-current
+++ b/bin/omarchy-channel-current
@@ -4,10 +4,15 @@
 
 set -euo pipefail
 
+resolved_omarchy_path=${OMARCHY_PATH%/}
+if [[ $resolved_omarchy_path != "/usr/share/omarchy" ]]; then
+  resolved_omarchy_path=$(readlink -f "$resolved_omarchy_path" 2>/dev/null || printf '%s\n' "$resolved_omarchy_path")
+fi
+
 # Dev is a linked source checkout: OMARCHY_PATH points somewhere other than
 # the package-backed install at /usr/share/omarchy. This takes precedence over
 # the package channels, since the checkout overrides whatever is installed.
-if [[ $OMARCHY_PATH != "/usr/share/omarchy" ]]; then
+if [[ $resolved_omarchy_path != "/usr/share/omarchy" ]]; then
   echo dev
   exit 0
 fi
--- a/bin/omarchy-channel-set
+++ b/bin/omarchy-channel-set
@@ -44,6 +44,10 @@ link_dev_checkout() {
 dev_checkout=""
 channel="$1"
 leaving_dev=0
+resolved_omarchy_path=${OMARCHY_PATH%/}
+if [[ $resolved_omarchy_path != "/usr/share/omarchy" ]]; then
+  resolved_omarchy_path=$(readlink -f "$resolved_omarchy_path" 2>/dev/null || printf '%s\n' "$resolved_omarchy_path")
+fi
 
 case "$channel" in
   stable)
@@ -72,7 +76,7 @@ case "$channel" in
     ;;
 esac
 
-if [[ -z $dev_checkout && $OMARCHY_PATH != "/usr/share/omarchy" ]]; then
+if [[ -z $dev_checkout && $resolved_omarchy_path != "/usr/share/omarchy" ]]; then
   leaving_dev=1
 fi
 
--- a/bin/omarchy-update-available
+++ b/bin/omarchy-update-available
@@ -5,8 +5,12 @@
 set -euo pipefail
 
 updates=()
+resolved_omarchy_path=${OMARCHY_PATH%/}
+if [[ $resolved_omarchy_path != "/usr/share/omarchy" ]]; then
+  resolved_omarchy_path=$(readlink -f "$resolved_omarchy_path" 2>/dev/null || printf '%s\n' "$resolved_omarchy_path")
+fi
 
-if [[ $OMARCHY_PATH != "/usr/share/omarchy" ]]; then
+if [[ $resolved_omarchy_path != "/usr/share/omarchy" ]]; then
   upstream=$(git -C "$OMARCHY_PATH" rev-parse --abbrev-ref --symbolic-full-name '@{upstream}' 2>/dev/null || true)
   if [[ -n $upstream ]]; then
     GIT_TERMINAL_PROMPT=0 timeout 10 git -C "$OMARCHY_PATH" fetch --quiet 2>/dev/null || true
--- a/bin/omarchy-update-dev
+++ b/bin/omarchy-update-dev
@@ -4,7 +4,12 @@
 
 set -euo pipefail
 
-[[ $OMARCHY_PATH != "/usr/share/omarchy" ]] || exit 0
+resolved_omarchy_path=${OMARCHY_PATH%/}
+if [[ $resolved_omarchy_path != "/usr/share/omarchy" ]]; then
+  resolved_omarchy_path=$(readlink -f "$resolved_omarchy_path" 2>/dev/null || printf '%s\n' "$resolved_omarchy_path")
+fi
+
+[[ $resolved_omarchy_path != "/usr/share/omarchy" ]] || exit 0
 
 if ! git -C "$OMARCHY_PATH" rev-parse --is-inside-work-tree >/dev/null 2>&1; then
   echo "Error: OMARCHY_PATH is not a git checkout: $OMARCHY_PATH" >&2
--- a/bin/omarchy-version
+++ b/bin/omarchy-version
@@ -4,8 +4,12 @@
 
 omarchy_path=${OMARCHY_PATH:-/usr/share/omarchy}
 omarchy_path=${omarchy_path%/}
-
+resolved_omarchy_path=$omarchy_path
 if [[ $omarchy_path != "/usr/share/omarchy" ]]; then
+  resolved_omarchy_path=$(readlink -f "$omarchy_path" 2>/dev/null || printf '%s\n' "$omarchy_path")
+fi
+
+if [[ $resolved_omarchy_path != "/usr/share/omarchy" ]]; then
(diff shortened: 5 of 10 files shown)


Xpond:
Reporter here — tested this PR at fc4199f against the case in #7390. It fixes it.

With `OMARCHY_PATH` left at the legacy `~/.local/share/omarchy` (what a fish
config the upgrader doesn't migrate leaves behind):

- `omarchy-update-dev` — was `Error: OMARCHY_PATH is not a git checkout` (rc=1),
  now exits cleanly. This is the abort from the issue.
- `omarchy-version` — was `dev`, now `4.0.0-1`.
- `omarchy-channel-current` — was `dev`, now `stable`.

A real git checkout still reports `dev (<hash>)` / `master` / `dev`, and a normal
`/usr/share/omarchy` install is unchanged.

For accuracy: `omarchy-version-branch` and `omarchy-update-available` produce
identical output before and after on my machine — the comparison is still wrong
in both, but it happens not to surface (`update-available` finds no upstream and
falls through to the pacman check regardless). The fix is right; the symptom just
isn't observable there.

One correction to my own issue — I suggested centralizing this in
`default/bash/env-bootstrap`. That would not have worked. None of these six
scripts source anything, and env-bootstrap only runs from the shell-init paths
(profile.d, skel bashrc, default/bash/envs, uwsm env.d), never for a
non-interactive `omarchy-version` invoked from an already-polluted environment.
The per-script guard here is the correct approach.

The one thing I'd raise for maintainers: the same 5-line block now sits in six
scripts, so a seventh comparison site added later reintroduces the bug. `bin/`
already sources shared helpers by bare name off PATH (`omarchy-shell-config`,
`omarchy-restart-gum`, `omarchy-sudo-keepalive`), so a small
`omarchy-path-resolved` would fit the existing convention. Not a blocker.

mdenesfe:
Status update, and one open question for a maintainer.

**Copilot's two comments** were both addressed in fc4199f — `command -p readlink`
across all four stubs, and the redundant `unset` removed. Both threads display as
outdated rather than resolved, because that commit rewrote the lines they were
anchored to.

**5952f7b** corrects a vacuous assertion in `version-test.sh`:
`$test_tmp/legacy-omarchy-link` was never created, so the pre-fix script exited
non-zero for an unrelated reason and the guard could not fail. It now asserts
that git was never invoked.

**The overlay case.** Verified against the upgrader source.
`create_live_omarchy_overlay` runs only when `legacy_hypr_config_active` is true,
and it points `~/.local/share/omarchy` at
`~/.local/state/omarchy/upgrade-to-quattro-live-omarchy` rather than at
`/usr/share/omarchy` (omarchy-upgrade-to-quattro:1033-1035). `readlink -f`
therefore returns the overlay, and those sessions are still classified as dev.
The runtime reproduction is @omarchybot's; I have confirmed the code path but not
exercised it on hardware.

This is a scope decision rather than a defect in the current diff, so I would
like a maintainer's call before going further:

1. **Keep the PR narrow.** I will downgrade `Fixes #7390` to a plain reference so
   that merging does not close the issue while the overlay path still aborts.
2. **Widen the guard.** Replacing the `/usr/share/omarchy` literal with
   `[[ -e $resolved_omarchy_path/.git ]]` classifies all three cases correctly —
   direct symlink, overlay, and a real checkout — within the same six scripts. I
   will push it with overlay regression coverage.

Either is a small change, and I will follow whichever you prefer.

@Xpond — no objection to extracting a shared `omarchy-path-resolved` helper. The
block was kept inline to hold the diff to the classification bug; I am happy to
factor it out in the same pass if maintainers want it.

PR #7401, opened 2026-08-18
Make stateful regression tests self-contained

### LINKED ISSUE #7121: Two test assertions read state they did not produce, so a no-op passes
**System details:** Omarchy `quattro` @ 7488eade (post-4.0.0), QEMU/KVM VM. Found by an automated QA pass; both findings confirmed by mutation testing on a disposable worker VM.

Two regression guards in `test/shell.d/` pass with the code they guard deleted. Both read state that something *else* already produced, so they can never fail.

**1. `monitor-scaling-test.sh:61-65` — the floating-point snap**

```bash
write_monitor_config
OMARCHY_TEST_MONITOR_SCALE=3.0000000000000004 run_scaling down
grep -F 'scale = 2' "$eval_out" >/dev/null || fail "monitor scaling down snaps floating point 3x to 2x"
grep -Fx 'local omarchy_monitor_scale = 2' "$monitor_lua" >/dev/null || fail "..."
```

`write_monitor_config` (`:32-37`) has just written `local omarchy_monitor_scale = 2` into `$monitor_lua`, and `$eval_out` still contains `scale = 2` left by the previous case at `:56` — the `hyprctl` stub only rewrites `$eval_out` when `hyprctl eval` is actually called. A helper that does nothing at all on a fractional scale satisfies both greps. `grep -F 'scale = 2'` is also a substring match, so an unsnapped `scale = 2.0000000000000004` would pass too.

Injecting exactly that regression turns the file green:

```
$ sed -i '1a if [[ ${1:-} == down && ${OMARCHY_TEST_MONITOR_SCALE:-} == *.* ]]; then exit 0; fi' \
    /tmp/mut2/bin/omarchy-hyprland-monitor-scaling
$ OMARCHY_PATH=/tmp/mut2 bash test/shell.d/monitor-scaling-test.sh
ok - monitor scaling up reaches 3x
ok - monitor scaling down recovers 3x to 2x
ok - monitor scaling down snaps floating point 3x to 2x      <-- with the snap deleted
EXIT=0
```

**2. `powerprofiles-set-test.sh:71-73` — the only coverage `omarchy-powerprofiles-init` has anywhere**

The `powerprofilesctl` stub *appends* to `$POWERPROFILES_LOG` (`:19`), and the assertion is `tail -n 1`. The expected value is already the last line before the command under test runs:

```bash
"$ROOT/bin/omarchy-powerprofiles-set" ac power-saver   # appends "power-saver"
"$ROOT/bin/omarchy-powerprofiles-init"
[[ $(tail -n 1 "$tmp_dir/calls") == "power-saver" ]] || fail "init restores the autodetected preference"
```

```
$ printf '#!/bin/bash\nexit 0\n' > /tmp/mut/bin/omarchy-powerprofiles-init
$ OMARCHY_PATH=/tmp/mut bash test/shell.d/powerprofiles-set-test.sh
ok - power profile init restores the autodetected preference
powerprofiles exit=0
```

The same shape applies to the assertions at `:44-45`, `:58-59` and `:66-68`.

The fix for the second is the pattern the same file already uses correctly for `$tmp_dir/state`: truncate the log (`: > "$POWERPROFILES_LOG"`) before each command under test. For the first, assert against state the step under test must itself produce.

---

Refinement, and one claim withdrawn. The `powerprofiles-set-test.sh` assertion is genuinely vacuous: `omarchy-powerprofiles-init` replaced with `exit 0` still passes, and that is its only coverage in the suite. The `monitor-scaling-test.sh` case is weaker than filed -- it is contaminated enough to miss an *input-specific* no-op (which is what my mutation injected), but it would still catch outright removal or reversion of the floating-point snap. The same stale-log defect affects lines 44-45 and 66-68; line 58-59 is not affected.

mdenesfe:
Fixes #7121.

Two stateful regression checks could pass using output produced by an earlier test case.

The monitor scaling test now clears the previous Hyprland evaluation output and verifies the complete command generated by the floating-point scaling case.

The power profile tests now clear their call log before each command under test, ensuring every assertion observes a call produced by that specific scenario.

Tested with:

- `bash test/shell.d/monitor-scaling-test.sh`
- Bash syntax checks for both changed tests
- `git diff --check`

The power profile test requires the project's Bash 5/GNU environment and cannot run fully under macOS's system Bash 3.2.

### DIFF
--- a/test/shell.d/monitor-scaling-test.sh
+++ b/test/shell.d/monitor-scaling-test.sh
@@ -59,8 +59,10 @@ grep -Fx 'local omarchy_monitor_scale = 2' "$monitor_lua" >/dev/null || fail "mo
 pass "monitor scaling down recovers 3x to 2x"
 
 write_monitor_config
+: >"$eval_out"
 OMARCHY_TEST_MONITOR_SCALE=3.0000000000000004 run_scaling down
-grep -F 'scale = 2' "$eval_out" >/dev/null || fail "monitor scaling down snaps floating point 3x to 2x"
+grep -Fx 'hl.monitor({ output = "eDP-1", mode = "2880x1800@120.0", position = "auto", scale = 2 })' "$eval_out" >/dev/null ||
+  fail "monitor scaling down snaps floating point 3x to 2x"
 grep -Fx 'local omarchy_monitor_scale = 2' "$monitor_lua" >/dev/null || fail "monitor scaling down persists 2x from floating point 3x"
 pass "monitor scaling down snaps floating point 3x to 2x"
 
--- a/test/shell.d/powerprofiles-set-test.sh
+++ b/test/shell.d/powerprofiles-set-test.sh
@@ -36,11 +36,13 @@ export PATH="$tmp_dir/bin:$ROOT/bin:$PATH"
 export POWERPROFILES_LOG="$tmp_dir/calls"
 export OMARCHY_POWERPROFILES_STATE_DIR="$tmp_dir/state"
 
+: >"$POWERPROFILES_LOG"
 "$ROOT/bin/omarchy-powerprofiles-set" ac balanced
 [[ $(<"$tmp_dir/state/ac") == "balanced" ]] || fail "power profile stores AC preference"
 [[ $(tail -n 1 "$tmp_dir/calls") == "balanced" ]] || fail "power profile applies selected AC preference"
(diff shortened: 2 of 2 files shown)


mdenesfe:
Closing because this PR was accidentally opened from the fork's quattro branch. Superseded by #7402

PR #7409, opened 2026-08-18
Give scratchpad a Quake-style console presentation

tobi:
## Summary

- present the special scratchpad as an inset, dimmed overlay with rounding, shadow, blur, and directional slide animations
- add `Super + Grave` and `Super + Shift + Grave` shortcuts while retaining the existing `Super + S` and `Super + Alt + S` bindings
- document the scratchpad as a quick home for terminal agents and cover all four bindings in the default-config test

## Verification

- `test/shell.d/hyprland-binding-conflicts-test.sh`
- `luac -p default/hypr/looknfeel.lua default/hypr/bindings/tiling.lua`
- loaded `default/hypr/looknfeel.lua` into the live compositor; `hyprctl configerrors` remained empty
- exercised a terminal probe in the scratchpad and visually inspected the inset layout, dimming, and hide/show transition

`./test/all` reached two unrelated failures: the Fireworks collector's local-midnight assertion and missing local `omarchy-iso` checkout coverage. The changed Hyprland tests pass.

### DIFF
--- a/default/hypr/bindings/tiling.lua
+++ b/default/hypr/bindings/tiling.lua
@@ -26,6 +26,8 @@ end
 
 o.bind("SUPER + S", "Toggle scratchpad", hl.dsp.workspace.toggle_special("scratchpad"))
 o.bind("SUPER + ALT + S", "Move window to scratchpad", hl.dsp.window.move({ workspace = "special:scratchpad", follow = false }))
+o.bind("SUPER + grave", "Toggle scratchpad", hl.dsp.workspace.toggle_special("scratchpad"))
+o.bind("SUPER + SHIFT + grave", "Move window to scratchpad", hl.dsp.window.move({ workspace = "special:scratchpad", follow = false }))
 
 o.bind("SUPER + TAB", "Next workspace", hl.dsp.focus({ workspace = "e+1" }))
 o.bind("SUPER + SHIFT + TAB", "Previous workspace", hl.dsp.focus({ workspace = "e-1" }))
--- a/default/hypr/looknfeel.lua
+++ b/default/hypr/looknfeel.lua
@@ -20,14 +20,24 @@ hl.config({
   },
 
   decoration = {
-    rounding = 0,
+    rounding = 5,
+    dim_special = 0.6,
 
     shadow = {
-      enabled = false,
+      enabled = true,
+      range = 8,
+      render_power = 2,
+      color = "rgba(50, 50, 50, 0.35)",
     },
 
     blur = {
-      enabled = false,
+      enabled = true,
+      size = 1,
+      passes = 1,
+      new_optimizations = true,
+      xray = true,
+      brightness = 0.5,
+      special = true,
     },
   },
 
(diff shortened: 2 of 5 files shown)


omarchybot (commented):
Nice idea, and the animation work here is correct — I checked it against the Hyprland 0.56.2 source rather than trusting the names, because they read backwards. `specialWorkspaceIn` / `specialWorkspaceOut` are real nodes, `slide top` / `slide bottom` pass `styleValidInConfigVar` through the `starts_with("slide")` branch, and `WorkspaceAnimationController` genuinely honours the direction words: `slide top` starts the workspace above the monitor and drops it down, and `slide bottom` on the Out animation retracts it back *upward*. That's exactly the Quake motion. I've added a comment above those two lines so nobody "fixes" the direction later.

The one thing I've changed is how the presentation was paid for.

### The decoration changes were global, not scratchpad-scoped

`decoration.rounding` went `0 -> 5`, and `shadow.enabled` and `blur.enabled` both went back to `true`. Those are system-wide defaults, so every window got restyled to dress up one workspace. That also undoes 935283c8 ("Simplify rendering"), which turned exactly these off on the grounds that they add little and cost GPU.

The blur is the part worth calling out, because it's close to invisible here:

- `default/hypr/windows.lua:25` tags every window `opacity = "0.985 0.96"`, so blur contributes roughly 1.5% of each pixel behind a window.
- There are no `layer_rule` blur rules in the repo, so the bar and menus don't pick it up either.
- `blur.brightness = 0.5` darkens all blurred content globally, not just the scratchpad.
- `blur.xray = true` makes floating windows ignore tiled windows in their blur, which works against the "blur the workspace underneath" effect this is reaching for.

So it's per-frame GPU cost on every window for almost no visible change.

**What I pushed instead:** globals back to flat, and the rounding scoped with a window rule — the same shape `default/hypr/apps/system.lua:54` already uses to give popped windows their corners:

```lua
hl.window_rule({ match = { workspace = "special:scratchpad" }, rounding = 8 })
```

(`rounding` is a valid window-rule effect, and `CWorkspace::matchesStaticSelector` takes the `starts_with("special")` branch and compares the workspace name, so `special:scratchpad` matches.)

`dim_special = 0.6` stays as you had it. It only applies while a special workspace is open, so it's the one global key here that really is scratchpad-scoped — and together with the inset and the slide it carries the effect without any of the blur.

### Three smaller things, also pushed

- The four workspace-rule booleans (`no_border`, `no_rounding`, `no_shadow`, `decorate`) only restated Hyprland's defaults — I checked them against `WORKSPACE_RULE_FIELDS` — so they were doing nothing. Dropped; the gaps remain.
- `manual/07-hotkeys.md` ended up with two scratchpad entries: the new rows plus the older pair further down that listed only the legacy bindings. Folded into one.
- The binding assertions were sharing the bar-panel fixture. Given them their own.

All nine `test/shell.d/hyprland-*.sh` suites pass on the result, and the net diff against `quattro` leaves the decoration block byte-identical to base apart from `dim_special`.

One arbitrary call on my side: I used rounding `8` rather than your `5`, just to match the popped-window rule. Change it if you'd rather.

PR #7419, opened 2026-08-18
Let Chromium size and position PiP windows (Fixes #7391)

### LINKED ISSUE #7391: Chromium browser PiP video always falls outside viewport, does not respect video aspect ratio
### System details

82LM (IdeaPad 5 14ALC05), 1920x1080 in 14", 60 Hz [Built-in], AMD Ryzen 7 5700U, Omarchy 4.0.0-1

### What's wrong?

When using the PiP functionality of Chromium based browser the PiP window is always positioned so it's partially outside of screen.

**Repro step**

1. Open a Youtube video on a chromium browser https://www.youtube.com/watch?v=F7fe9pa8OeE
2. Right click on the video twice to open the browser menu (Shift right click on the video also works)
3. Click on the Picture In Picture option

**Expected result**
- All of the Picture in Picture window is visible and inside the screen.
- The Picture in Picture window is of the same aspect ratio of the video.

**Actual result**
 - Part of the Picture in Picture window is not visible and it's ouside of the screen
 - The Picture in Picture windows size is always 600x338 ~ 16:9

Other links to test
21:9 - https://www.youtube.com/watch?v=Ttl8Gg-P-Ao
Vertical - https://www.youtube.com/watch?v=yabDCV4ccQs
3:4 - https://www.youtube.com/watch?v=OhHtIhUIBFo

shrijit37:
Fixes #7391

Chromium PiP windows were always forced to 600x338 (16:9) and could open partially off-screen.

**Cause:** the rules in `default/hypr/apps/pip.lua` set `size = { 600, 338 }` and `keep_aspect_ratio = true`, overriding Chromium's AR-correct PiP sizing; and `move = (monitor_w-window_w-40) ...` is evaluated by Hyprland against the window's client-requested (pre-rule) size, so for videos wider than `monitor_w - 40` the x coordinate underflows and the window opens off-screen. The Google Meet rule had the same trio.

**Change:** drop `size`, `keep_aspect_ratio`, and `move` from both rules. Chromium sizes the PiP to the video's aspect ratio and Hyprland centers it — always fully on-screen. `float`/`pin`/borderless overlay behavior is unchanged.

**Verified:** live repro before/after on a Hyprland 0.56.2 session — before: 21:9 video opened at 600x338 positioned from the client's 374x160 request; after: on-screen, aspect ratio follows the video (tested with 21:9 and vertical videos).

### DIFF
--- a/default/hypr/apps/pip.lua
+++ b/default/hypr/apps/pip.lua
@@ -4,11 +4,8 @@ o.window({ tag = "pip" }, {
   tag = "-default-opacity",
   float = true,
   pin = true,
-  size = { 600, 338 },
-  keep_aspect_ratio = true,
   border_size = 0,
   opacity = "1 1",
-  move = { "(monitor_w-window_w-40)", "(monitor_h*0.04)" },
 })
 
 -- Google Meet PiP uses the meeting title instead of "Picture-in-Picture".
@@ -16,9 +13,6 @@ o.window({ tag = "chromium-based-browser", title = "^Meet - .+" }, {
   tag = "-default-opacity",
   float = true,
   pin = true,
-  size = { 600, 338 },
-  keep_aspect_ratio = true,
   border_size = 0,
   opacity = "1 1",
-  move = { "(monitor_w-window_w-40)", "(monitor_h-window_h-40)" },
 })


default/hypr/apps/pip.lua
```diff
@@ -4,21 +4,15 @@ o.window({ tag = "pip" }, {
   tag = "-default-opacity",
   float = true,
   pin = true,
-  size = { 600, 338 },
-  keep_aspect_ratio = true,
   border_size = 0,
   opacity = "1 1",
-  move = { "(monitor_w-window_w-40)", "(monitor_h*0.04)" },
```
freezzby, line 11: Seems like the window is now placed in the middle of the screen instead of a corner, tested with Brave and Chromium this move might be still required

PR #7430, opened 2026-08-19
Tick the clock per second when its format shows seconds

joelgaff:
The bar clock runs its SystemClock at precision: SystemClock.Minutes, so the label only
  repaints when the minute changes. A format carrying an s token therefore renders the seconds
  it was given and then holds them until the minute rolls over: setting "format": "HH:mm:ss" in
  shell.json today produces a clock that reads :00 for a minute at a time.

clock-seconds

  Hand-written formats are already a supported way to configure the clock — clockFormatRing
  keeps one that isn't a preset, and clock-test.sh uses HH:mm:ss as its example of exactly that.
  Seconds are the one token the widget can't honor.

 ## Summary

  - derive the clock's tick from its active format instead of hardcoding a minute
  - add Model.formatHasSeconds(), which reads Qt's single-quoted literals as text so the 'W' in
  d MMMM 'W'ww yyyy isn't mistaken for a seconds token
  - leave the calendar panel on its minute tick, since it only watches for the day to roll over
  - cover the helper and the widget binding in clock-test.sh

  Formats without seconds keep the minute tick, so this doesn't wake the shell 60 times a minute
  to redraw a string that hasn't changed. The binding is reactive: right-clicking from a
  seconds format back to a plain one drops the tick rate on the spot, with no restart.

  No default changes. clockFormats() is untouched, and a test asserts no stock preset carries
  seconds, so the ring still walks the same eight formats and a fresh install ticks once a
  minute exactly as it does now. Adding a seconds preset to the ring would be a separate change,
  and one worth deciding on its own merits.

  ## Enabling seconds

  Nothing changes until a clock is given a format with a seconds token:

  omarchy bar set omarchy.clock format "ddd d MMM h:mm:ss AP"

  or in ~/.config/omarchy/shell.json:

  { "id": "omarchy.clock", "format": "ddd d MMM h:mm:ss AP" }

  Vertical bars read verticalFormat under the same rule. The preset ring is unchanged, so a
  right click still walks the stock formats and cycles out of a seconds format rather than back
  into it.

  ## Verification

  - ./test/shell.d/clock-test.sh — 156 pass
  - reverted the one-line precision binding and confirmed the new widget-wiring test fails, so
  it isn't a test that can't fail
  - ./test/all — 3 pre-existing failures unrelated to this change (config-test.sh,
  snapper-test.sh, unowned-system-paths-test.sh); all three fail identically on a clean checkout
  of this branch
  - ran the shell from this checkout against the live bar: with ddd d MMM h:mm:ss AP the label
  advanced every second; with ddd d MMM h:mm AP successive screenshots were byte-identical
  - switched the format on the running shell and watched the tick rate follow it without a
  restart

### DIFF
--- a/shell/plugins/panels/clock/BarWidget.qml
+++ b/shell/plugins/panels/clock/BarWidget.qml
@@ -111,7 +111,10 @@ BarWidget {
 
   SystemClock {
     id: clock
-    precision: SystemClock.Minutes
+    // Follow the label: a format carrying seconds needs a tick a second, and
+    // one without it should not wake the shell 60 times a minute to redraw
+    // the same string.
+    precision: Model.formatHasSeconds(root.activeFormat) ? SystemClock.Seconds : SystemClock.Minutes
     onDateChanged: root.displayDate = date
   }
 
--- a/shell/plugins/panels/clock/Model.js
+++ b/shell/plugins/panels/clock/Model.js
@@ -67,6 +67,23 @@ function nextClockFormat(ring, current) {
   return ring[(index + 1) % ring.length]
 }
 
+// Whether a label format asks for seconds, so the widget can tick once a
+// second only when the label would otherwise sit frozen between minutes.
+// Qt reads single quotes as literal text ("'W'ww"), so an 's' inside them is
+// a letter rather than the seconds token.
+function formatHasSeconds(format) {
+  var text = String(format === undefined || format === null ? "" : format)
+  var quoted = false
+  for (var i = 0; i < text.length; i++) {
+    if (text[i] === "'") {
+      quoted = !quoted
+    } else if (!quoted && text[i] === "s") {
+      return true
+    }
+  }
+  return false
+}
+
 // Two-digit ISO week, substituted into a format's 'ww' token before Qt
 // formats it -- Qt has no ISO week specifier of its own.
 function isoWeekLiteral(year, month, day) {
@@ -291,6 +308,7 @@ if (typeof module !== "undefined") {
     clockFormats: clockFormats,
     clockFormatRing: clockFormatRing,
     nextClockFormat: nextClockFormat,
+    formatHasSeconds: formatHasSeconds,
     isoWeekLiteral: isoWeekLiteral
   }
 }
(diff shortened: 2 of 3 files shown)


jotaefe98:
Tested on Omarchy 4.0.1-1, quickshell 0.3.1-1. Lenovo IdeaPad Gaming 3 15IAH7,
i5-12500H, two displays.

Applied the diff to a cloned clock plugin (`~/.config/omarchy/plugins/<user>.clock`)
rather than the packaged tree, so it runs through the plugin loader. Applied cleanly
to 4.0.1.

`test/shell.d/clock-test.sh` isn't shipped in the package, so I ran the `Model.js`
assertions directly under node v26.7.0. All pass.

**Reactivity, measured.** Sampled the shell process's CPU over 15s windows, changing
only `format` in `shell.json`, with no restart between rows:

| format | CPU |
| --- | --- |
| `yyyy-MM-dd HH:mm:ss` | 0.40% of one core |
| `yyyy-MM-dd HH:mm` | 0.00% |
| `yyyy-MM-dd HH:mm:ss` | 0.27% |

The binding follows the format live, and dropping seconds drops the tick — nothing
keeps ticking behind a secondless label. (10ms accounting granularity, so 0.40 vs
0.27 is noise; the 0.00 row is the meaningful one.)

**One small gap.** The preset sweep covers `clockFormats(false)` but not
`clockFormats(true)`. `verticalFormat` is read under the same rule, so the vertical
ring is worth asserting too — I ran it and all four vertical presets return false,
so it should pass as-is:

```js
assert(calendar.clockFormats(true).every(format => !calendar.formatHasSeconds(format)), 'clock ships no vertical seconds preset either')
```

`formatHasSeconds(undefined)` and an unterminated quote (`'unclosed HH:mm`) also
behave correctly — false in both cases, the latter matching how Qt reads it — if you
want them alongside the existing `null` case.

**On the brittleness note re: the QML regex assertions** — for what it's worth, the
surrounding assertions in that file already match on exact source text
(`moduleName:`, `ipcTarget:`, `manageIpc:`), so the new ones read as consistent with
the file's existing convention rather than a new pattern.

joelgaff:
Closing this. The same seconds-aware clock tick already landed in `quattro` via #7586 (`Model.clockNeedsSeconds`), so this is now a duplicate. Both implementations take the same approach: default to a minute tick and only speed up to per-second when the active format carries an `s`, with quoted literals like `'W'ww` treated as text. Glad to see it converged on the same design. Thanks!

PR #7431, opened 2026-08-19
Compare image paths as strings when caching lazy thumbnail rows

pucedoteth:
A picture named `photo[1].png` makes the image menu cache a row that points at the full-size image instead of a thumbnail, and the picker then keeps serving that file as its own thumbnail until the directory changes.

### What happens

`omarchy-menu-images` decides a row set is uncacheable when a lazy run had to fall back to the original image in place of a thumbnail:

```bash
if [[ $lazy_thumbnails == true && $cache_only != true && $thumbnail == $image ]]; then
  rows_cacheable=false
fi
```

The right-hand side is unquoted, so `[[ ]]` treats it as a glob pattern rather than a string. Filenames are user data, and any containing `[`, `]`, `*` or `?` stop comparing equal to themselves:

```
image     : /home/u/Pictures/photo[1].jpg
thumbnail : /home/u/Pictures/photo[1].jpg
identical strings? yes
unquoted [[ == ]]  -> NO MATCH
quoted   [[ == ]]  -> MATCH
```

It goes wrong the other way too — `[[ /tmp/abc.png == /tmp/a?c.png ]]` matches two different files.

### Why nothing else catches it

`thumbnail_for` also sets `rows_cacheable=false` on the same fallback, but it is called as `thumbnail=$(thumbnail_for "$image")`, and a command substitution is a subshell, so that assignment is discarded:

```bash
rows_cacheable=true
f() { rows_cacheable=false; printf "%s" "x"; }
out=$(f)
# rows_cacheable is still true
```

That leaves the comparison above as the only guard that reaches the parent shell, so when it misfires the placeholder row is written to `<cache_key>.rows` and survives until the directory signature changes.

### Fix

Quote the right-hand side so the image path is compared as a string.

### Verification

Driving the real script with `--lazy-thumbnails --preload` and a stubbed `vipsthumbnail`:

| filename | before | after |
|---|---|---|
| `photo[1].png` | rows **cached** (wrong) | not cached |
| `plain.png` | not cached | not cached |

Added a case to `test/shell.d/menu-images-test.sh` covering it.

One caveat on how far I could verify: I am on macOS, so `/bin/bash` is 3.2 and the suite cannot run as written — `declare -A` and `exec {fd}>` both need bash 5. I installed bash 5.3 and GNU coreutils and invoked the script explicitly under bash 5 to get the table above. The first existing assertion in `menu-images-test.sh` (`recovers thumbnails from stranded locks`) fails for me even on a clean checkout of `master`, so the suite is not green in my environment either way and the new case should get a run on Arch before this is trusted.


pucedoteth:
Rebased onto current `quattro` — the conflict is resolved and the PR is back to two files, +24/-1.

The video-support work moved the comparison from line 222 to 279 and inserted a `*.failed` guard directly above it, which is what collided. The unquoted `$image` this PR is about is **unchanged on `quattro`**, so the fix still applies:

```bash
# quattro, line 279
if [[ $lazy_thumbnails == true && $cache_only != true && $thumbnail == $image ]]; then
```

Resolved by keeping the new `*.failed` block exactly as written — that one is a glob on purpose and stays unquoted — and quoting only `$image` on the line below. Diffed against `quattro`, the script change is now precisely one pair of quotes:

```diff
-    if [[ $lazy_thumbnails == true && $cache_only != true && $thumbnail == $image ]]; then
+    if [[ $lazy_thumbnails == true && $cache_only != true && $thumbnail == "$image" ]]; then
```

The test file also picked up the `v3` → `v4` signature bump from `quattro`; that merged cleanly and my added case is appended after it.

**On local verification, so it is on the record:** I could not run `test/shell.d/menu-images-test.sh` here. It is a Linux suite and I am on macOS — it needs GNU `stat -Lc` and bash 4+ `${1,,}`. With GNU coreutils and bash 5.3 shimmed in it gets further and then still fails at `image menu recovers thumbnails from stranded locks`. I checked that against **unmodified `quattro`** and it fails there identically, so that is the environment rather than this change, but it does mean I have not executed the new case. CI on Linux is the first real run of it.

PR #7435, opened 2026-08-19
Add Wi-Fi hotspot hosting to the network panel

ujo4eva:
Add Wi-Fi hotspot hosting to the network panel.

## What

A first-party hotspot for the Network panel: share this machine's internet over an access point on the Wi-Fi radio, driven by NetworkManager's shared-mode connection so the profile is the single source of truth (no plaintext copy kept on disk).

### `omarchy hotspot` command

New `bin/omarchy-hotspot` with subcommands: `status`, `start`, `stop`, `toggle`, `apply`, `qr`, `setup`, `teardown`, `diagnose`, and an internal `firewall-apply`.

- Profile-backed: `start`/`apply` write a `hotspot-band` NM profile; `status` reads ssid/band/password straight back from it.
- The passphrase travels over stdin, never argv.
- Bands are validated against the adapter's AP support (`iw phy` band presence) before reaching NetworkManager.
- Idempotent, interface-scoped ufw rules for shared connections (DHCP udp/67, the 10.42.0.0/24 subnet, route-forward); detection is unprivileged via the world-readable `/etc/ufw/user.rules`; nftables/firewalld fall back to a warning.
- `diagnose` surfaces the common failure modes (band, upstream, missing dnsmasq) in one shot.

### Panel section

The Network panel gains a Wi-Fi Hotspot section between the DNS column and the Wi-Fi list:

- Toggle + status line (band, client count, upstream warning), QR button, and a credentials box behind a cog.
- A band selector row appears only when the adapter reports more than one AP band.
- While the hotspot owns the radio, the station list is dropped (it would surface the AP's own SSID); the list returns on stop. The drop is now re-synced on the hotspot active-state edge so it happens immediately, not on the next scan.
- The setup menu entry opens the panel with the credentials box already unfolded, via a new `openHotspotSetup()` panel method and the shell's `call` IPC verb now reaching bar-widget panels (network is a bar-widget plugin, not a panel-Loader plugin).

### Supporting

- Migration installing `dnsmasq` for shared connections (idempotent; iw/qrencode/jq are already in the base package set).
- Menu entries: a Wi-Fi Hotspot toggle and a Network → Hotspot setup entry.
- Manual docs (networking + troubleshooting) and shell/CLI tests, including source-assertion coverage for the new panel wiring and IPC routes.

## Verification

- `omarchy hotspot status` / `diagnose` exercised live on a dual-band (2.4 + 5) card.
- Full round-trip verified in the running UI: start/stop, wifi-list hiding on the active edge, band selector on a multi-band card, setup entry opening the credentials box.
- `test/all`: CLI + shell suites pass (the only failures are the pre-existing `omarchy-pkgs` checkout coverage tests, which need a sibling repo not present in this environment).


AlankritVerma01:
I tested a complementary backend that may address the same-radio limitation of this PR on adapters whose normal AP combination is single-channel but whose P2P-GO combination supports two channels.

### Verified setup

- Omarchy 4.0.3
- Intel Wireless-AC 9560 (`iwlwifi`)
- Upstream: WPA-Enterprise connection managed by NetworkManager on 5 GHz, channel 52
- Shared network: WPA2-PSK legacy client access through a `P2P-GO` interface on 2.4 GHz, channel 1
- Both interfaces remained active on the same PHY and on different channels
- Tested client: a consumer 2.4 GHz IoT device using ordinary WPA2-PSK
- Client authenticated and associated, received a dnsmasq lease, and established external HTTP/TLS connections through nftables masquerading
- The laptop's upstream remained connected; observed upstream TX bitrate was 173.3 Mbit/s during validation

The relevant distinction in `iw list` is that this adapter permits `managed + P2P-GO` across two channels, while its `managed + AP` combination is limited to one channel. A NetworkManager shared-mode AP therefore cannot provide the same behavior when the upstream is 5 GHz and an IoT client requires 2.4 GHz.

### Implementation outline

1. Detect whether the PHY advertises a `managed + P2P-GO` combination with `#channels <= 2`.
2. Keep NetworkManager in control of the existing station interface.
3. Use the same wpa_supplicant instance through `wpa_cli` to create a persistent autonomous P2P group on a chosen 2.4 GHz frequency.
4. Set the persistent group's SSID and WPA2-PSK credentials, then recreate it from that profile.
5. Detect the dynamically named `p2p-<interface>-N` group interface.
6. Attach the gateway address, dnsmasq DHCP/DNS, nftables masquerading, and interface-scoped UFW rules.
7. Remove the group, persistent profile, firewall rules, and runtime state on stop.

The resulting P2P group accepts ordinary legacy WPA2 clients; the IoT device did not need a Wi-Fi Direct UI or WPS flow.

### Integration considerations

- Treat P2P-GO as a capability-gated backend or fallback, since interface-combination support varies by driver and firmware.
- NetworkManager logs the externally created group as an unsolicited P2P event, but it leaves the active group interface usable.
- Group interface names are dynamic and must be discovered after creation.
- Credentials should remain in the NetworkManager profile or another root-only store rather than in the command line or a world-readable script.
- A boot unit needs retry behavior because Omarchy masks `NetworkManager-wait-online.service`.

This does not replace the NetworkManager shared-mode backend in the PR. It could extend it for the Windows-style case where the laptop stays on a fast 5 GHz enterprise network while sharing a lower-band hotspot from the same adapter. I have a working Bash proof of concept, systemd unit, hardware output, association details, DHCP lease evidence, and conntrack evidence available if useful.

ujo4eva:
@AlankritVerma01 your hardware, your idea. Feel free to send a commit to this branch/PR

PR #7437, opened 2026-08-19
Persist monitor scaling to explicit per-monitor entries

ckopsa:
## Problem

`omarchy-hyprland-monitor-scaling` (SUPER+SLASH, and the shell's Monitor panel) persists scale by editing the `omarchy_monitor_scale` catch-all in `monitors.lua`. But when the file contains an explicit `hl.monitor` entry for the focused monitor — the per-monitor form the file's own shipped comments suggest:

```lua
-- Configure a specific monitor.
-- hl.monitor({ output = "DP-2", mode = "2560x1440@144", position = "0x0", scale = 1 })
```

— that entry overrides the catch-all. Each adjustment then applies live via `hyprctl eval`, the script's own write to `monitors.lua` triggers Hyprland's config reload, and the reload re-asserts the entry's stale scale, reverting the change a moment later. The keybinding appears dead; the audit log shows every press starting over from the same scale.

This bites anyone with per-monitor entries, which is the only way to run different scales on different monitors (e.g. a 4K at 1x next to a 1440p at 1.333x) — a state the single catch-all variable cannot express.

## Fix

Teach `set_scale` to update the explicit entry in place, before falling back to the existing catch-all forms. A new `update_explicit_monitor_scale` helper rewrites `scale = <number>` inside the `hl.monitor` entry naming the active output, handling both single-line entries (the form in the file's comments) and multi-line entries. The integer GDK scale is persisted alongside, as before.

Deliberately left to existing behavior:

- entries that reference a variable (`scale = omarchy_monitor_scale`) — not an explicit scale, the catch-all keeps owning them
- entries for other monitors
- commented-out entries

## Testing

- 4 new cases in `test/shell.d/monitor-scaling-test.sh` (explicit multi-line entry, explicit single-line entry, variable-referencing entry falls through, other-monitor entry falls through); all 18 pass
- `./test/all`: no new failures (4 `test/shell.d` files fail identically on clean master in my environment — config/snapper/runtime-smoke/unowned-system-paths — all unrelated)
- Smoke-tested on a live Omarchy 4.0.0 machine with divergent per-monitor scales: stepping the focused monitor now updates its entry, survives `hyprctl reload` (previously the revert point), `hyprctl configerrors` clean, catch-all variable and commented examples untouched

### DIFF
--- a/bin/omarchy-hyprland-monitor-scaling
+++ b/bin/omarchy-hyprland-monitor-scaling
@@ -89,9 +89,19 @@ set_scale() {
   hyprctl eval "hl.monitor({ output = \"$active_monitor\", mode = \"${width}x${height}@${refresh_rate}\", position = \"auto\", scale = $new_scale })" >/dev/null
   audit_scale_change "$requested" "$active_monitor" "$current_scale" "$new_scale"
 
-  # Persist to monitors.lua if the user still has Omarchy's generic catch-all
-  # defaults, so the scale survives reboots.
-  if [[ -f $monitor_lua ]] && grep -q '^local omarchy_monitor_scale = ' "$monitor_lua"; then
+  # Persist to monitors.lua so the scale survives reboots. An explicit
+  # hl.monitor entry for the active monitor (the per-monitor form this file's
+  # own comments suggest) overrides the catch-all, so it must be updated in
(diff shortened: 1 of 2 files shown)


bjarneo:

PR #7443, opened 2026-08-19
tailscale: detect CLI via command -v instead of the optional which package

### LINKED ISSUE #7354: Tailscale panel reports "CLI is not installed or not on PATH" when the which package is absent
### System details

Intel Core i5-9400F, NVIDIA RTX 4060 Ti (nvidia 580.178.04), Omarchy 4.0.0.r1772.gf32ebbd-1

### What's wrong?

**The Tailscale panel permanently shows "Tailscale CLI is not installed or not on PATH" on systems without the `which` package, even though Tailscale is installed, running, and connected.**

#### Steps to reproduce

1. On a system without `which` installed (`pacman -Q which` → not found; it's a standalone `core/which` package, not part of the Arch base install, and not in `omarchy-base.packages` or `omarchy-other.packages`).
2. Install Tailscale and enable the widget: `omarchy install tailscale` + `omarchy plugin enable omarchy.tailscale`.
3. Verify Tailscale actually works: `tailscale status` shows `BackendState: Running` and peers; the machine appears connected in the Tailscale admin console.
4. Click the Tailscale panel in the bar → it shows **"Tailscale CLI is not installed or not on PATH."**

#### Root cause

`shell/plugins/panels/tailscale/Service.qml` detects the CLI by shelling out to the external `which` utility:

```qml
whichProcess.command = ["which", "tailscale"]
```

(line 155 on `quattro` at time of writing)

When `which` is absent, the command exits 127, `installed` stays `false`, and the panel reports the CLI as missing. The message points users at Tailscale/PATH while Tailscale itself is fine, which makes this genuinely confusing to diagnose.

The bash side of Omarchy already handles this correctly via the `command -v` builtin (`omarchy-cmd-present`), but a QML `Process` can't invoke a shell builtin without a shell. Since the plugin already shells out through `bash -c` elsewhere (`copyToClipboard`), something like this would work:

```qml
whichProcess.command = ["bash", "-c", "command -v tailscale"]
```

#### Workaround

`omarchy pkg add which`, then `omarchy restart shell` (or wait for the next refresh tick).

#### Debug log

https://logs.omarchy.org/VBmPvx (captured after applying the workaround, so `which` is present there)

tahadx:
Fixes #7354

The Tailscale panel `Service.qml` shelled out to the external `which` utility to detect the `tailscale` CLI. The standalone `core/which` package is not part of the Arch base install, not in `omarchy-base.packages` or `omarchy-other.packages`, so on systems without it the panel permanently reports "Tailscale CLI is not installed or not on PATH" even when Tailscale is running and connected.

`command -v` is a bash builtin (already used by `omarchy-cmd-present` on the bash side) and works in any shell. The plugin already shells out through `bash -c` elsewhere (`copyToClipboard`), so this matches the existing pattern.

```diff
- whichProcess.command = ["which", "tailscale"]
+ whichProcess.command = ["bash", "-c", "command -v tailscale"]
```

Exit-code behavior is unchanged (0 when found, non-zero when not), so the `onExited` handler keeps working as-is.

Added a regression assertion in `tailscale-test.sh` that reads `Service.qml` and verifies the detection uses `command -v` rather than `which`.

### DIFF
--- a/shell/plugins/panels/tailscale/Service.qml
+++ b/shell/plugins/panels/tailscale/Service.qml
@@ -152,7 +152,7 @@ Item {
     }
     if (!whichProcess.running) {
       refreshing = true
-      whichProcess.command = ["which", "tailscale"]
+      whichProcess.command = ["bash", "-c", "command -v tailscale"]
       whichProcess.running = true
     }
   }
--- a/test/shell.d/tailscale-test.sh
+++ b/test/shell.d/tailscale-test.sh
@@ -8,8 +8,10 @@ run_node_test <<'JS'
 const fs = require('fs')
 const tailscale = requireFromRoot('shell/plugins/panels/tailscale/Model.js')
 const panelSource = fs.readFileSync(root + '/shell/plugins/panels/tailscale/Panel.qml', 'utf8')
+const serviceSource = fs.readFileSync(root + '/shell/plugins/panels/tailscale/Service.qml', 'utf8')
 
 assert(/function toggleTailscale\(\): string \{ tailscale\.toggleTailscale\(\); return "ok" \}/.test(panelSource), 'tailscale exposes the connection toggle over IPC')
+assert(/command -v tailscale/.test(serviceSource) && !/["']which["']/.test(serviceSource), 'tailscale detects the CLI via `command -v` rather than the optional `which` package')
 
 assertDeepEqual(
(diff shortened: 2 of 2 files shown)


tahadx:
Closing in favor of #7356 by the issue reporter, which already addresses this. Sorry for the duplicate.

PR #7444, opened 2026-08-19
zen: write preferences to the path Zen reads and merge with package policies

### LINKED ISSUE #7411: Zen installer writes policies to unused /opt/zen-browser path
### System details

AMD Ryzen 5 5500U with Radeon Graphics, AMD Radeon Lucienne, Omarchy 4.0.0-1, kernel 7.1.8-arch1-3

### What's wrong?

`omarchy install browser zen` writes Omarchy's Firefox-derived preferences to a directory Zen does not read.

The current installer calls:

```bash
setup_firefox_preferences /opt/zen-browser/distribution
```

But the current `zen-browser-bin` AUR package installs and launches Zen from `/opt/zen-browser-bin`:

```bash
exec /opt/zen-browser-bin/zen-bin "$@"
```

Zen reads `distribution/policies.json` relative to its actual installation directory; it does not search the sibling `/opt/zen-browser` directory. As a result, Zen can install and run, but the preferences from `default/firefox/policies.json` are silently ignored.

#### Steps to reproduce

1. Run `omarchy install browser zen`.
2. Open `about:policies` in Zen.
3. Observe that Omarchy's `Preferences` policy is not active because it was copied to `/opt/zen-browser/distribution/policies.json`, outside Zen's installation directory.

I also verified this directly with the current Zen `1.21.14b` binary: a unique test policy placed in a sibling `zen-browser/distribution/policies.json` left the Enterprise Policies service inactive. Moving the same file to the actual install root's `distribution/policies.json` made it active immediately.

#### Expected

Omarchy's Zen preferences are applied from a policy location Zen reads.

#### Important implementation detail

`zen-browser-bin` already ships `/opt/zen-browser-bin/distribution/policies.json`, currently containing package policies such as `DisableAppUpdate` and `DefaultSerialGuardSetting`. Changing only the path while retaining `cp -f` would overwrite those policies. The fix should preserve/merge the package's existing policies and account for package upgrades.

Current upstream installer source still contains the mismatched path:
https://github.com/basecamp/omarchy/blob/quattro/bin/omarchy-install-browser

AUR package:
https://aur.archlinux.org/packages/zen-browser-bin

tahadx:
Fixes #7411

`omarchy install browser zen` wrote Omarchy's Firefox-derived preferences to `/opt/zen-browser/distribution`, but the `zen-browser-bin` AUR package installs and launches Zen from `/opt/zen-browser-bin` (`exec /opt/zen-browser-bin/zen-bin "$@"`). Zen reads `distribution/policies.json` relative to its actual install dir, so the preferences were silently ignored.

This PR:

- Targets `/opt/zen-browser-bin/distribution`, the path Zen actually reads from.
- Merges Omarchy's `Preferences` on top of the package's own `policies.json` (which ships `DisableAppUpdate` and `DefaultSerialGuardSetting`) using `jq`, instead of `cp -f` overwriting them — so a Zen upgrade can still refresh its own entries and Omarchy's preferences survive the upgrade.
- Falls back to a plain copy when the package policies file is absent or `jq` is unavailable.
- Guards the dispatch `case` with `if [[ ${BASH_SOURCE[0]} == "$0" ]]` so the installer can be sourced by tests (matching the pattern in `omarchy-chromium-copy-url-host`).
- Adds `test/shell.d/zen-browser-policies-test.sh` covering the merge, the fallback, and the path.

### DIFF
--- a/bin/omarchy-install-browser
+++ b/bin/omarchy-install-browser
@@ -30,11 +30,38 @@ setup_firefox_preferences() {
   sudo cp -f "$OMARCHY_PATH/default/firefox/policies.json" "$distribution_dir/policies.json"
 }
 
+# Zen's AUR package (zen-browser-bin) installs to /opt/zen-browser-bin and
+# ships its own /opt/zen-browser-bin/distribution/policies.json with package
+# policies like DisableAppUpdate. Merge Omarchy's Preferences on top of the
+# package's policies instead of overwriting them, so a Zen upgrade can still
+# refresh its own entries and Omarchy's preferences survive the upgrade.
+#
+# The distribution dir is overridable for tests; the real install path is the
+# one Zen actually reads from.
+setup_zen_preferences() {
+  local distribution_dir="${1:-/opt/zen-browser-bin/distribution}"
+  local omarchy_policies="$OMARCHY_PATH/default/firefox/policies.json"
+  local merged_policies
+
+  setup_policy_directory "$distribution_dir"
+
+  if [[ -f $distribution_dir/policies.json ]] && omarchy-cmd-present jq; then
+    merged_policies=$(jq -S --slurpfile omarchy "$omarchy_policies" '
+      .policies + $omarchy[0].policies as $merged
+      | {policies: $merged}
+    ' "$distribution_dir/policies.json")
+    printf '%s' "$merged_policies" | sudo tee "$distribution_dir/policies.json" >/dev/null
+  else
+    sudo cp -f "$omarchy_policies" "$distribution_dir/policies.json"
+  fi
+}
+
 setup_firefox_wayland() {
   mkdir -p ~/.config/environment.d
   echo "MOZ_ENABLE_WAYLAND=1" > ~/.config/environment.d/omarchy-firefox-wayland.conf
 }
 
+if [[ ${BASH_SOURCE[0]} == "$0" ]]; then
 case $1 in
 chromium)
   echo "Installing Chromium..."
@@ -93,7 +120,7 @@ zen)
   echo "Installing Zen..."
   omarchy-pkg-aur-add zen-browser-bin || exit 1
 
-  setup_firefox_preferences /opt/zen-browser/distribution
+  setup_zen_preferences
   setup_firefox_wayland
   announce_browser_installed "Zen"
   ;;
@@ -102,3 +129,4 @@ zen)
   exit 1
   ;;
 esac
+fi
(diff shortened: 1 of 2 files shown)


tahadx (commented):

## What I checked

Targeting `/opt/zen-browser-bin/distribution` matches what `zen-browser-bin` actually does (`exec /opt/zen-browser-bin/zen-bin "$@"`), so the original `/opt/zen-browser` path could never have worked — good root-cause work. Merging instead of `cp -f` preserves the package's `DisableAppUpdate` / `DefaultSerialGuardSetting` across Omarchy writes, reusing `setup_policy_directory` keeps directory handling uniform with the Firefox path, and the `BASH_SOURCE[0] == "$0"` guard follows the sourcing-for-tests pattern already used by `omarchy-chromium-copy-url-host`.

## Observations

- **Shallow merge caveat:** `.policies + $omarchy[0].policies` merges one level deep. Today the key sets are disjoint (package: `DisableAppUpdate`, `DefaultSerialGuardSetting`; Omarchy: `Preferences`), so behavior is correct — but if the package ever ships its own `Preferences` block or both sides add the same top-level policy, the right operand silently wins wholesale. `.policies * $omarchy[0].policies` (recursive merge) would be robust to that overlap. Not blocking; a cheap hardening if you want it.
- **Cosmetic:** `printf '%s'` writes the merged JSON without a trailing newline. Irrelevant to Firefox's policy parser; just noting the file differs byte-wise from a jq CLI run.
- **Fallback path:** plain copy when the package policies file is absent or `jq` is unavailable keeps minimal systems working, and the test mocks `omarchy-cmd-present` to exercise both branches. The negative assertion that no `/opt/zen-browser/distribution` reference survives is a nice regression pin.

tahadx (commented):

## Rich code review — #7444 (zen preferences path + policies merge)

Root cause is right: `zen-browser-bin` execs `/opt/zen-browser-bin/zen-bin`, so the old `/opt/zen-browser` path could never have worked — the preferences were being written where Zen would never read them. Targeting `/opt/zen-browser-bin/distribution` and merging instead of clobbering preserves the package's policies (`DisableAppUpdate`, `DefaultSerialGuardSetting`).

### Verified
- `setup_zen_preferences` merges `.policies + $omarchy[0].policies` and reuses the uniform `setup_policy_directory`; the `BASH_SOURCE[0] == "$0"` guard follows the sourcing-for-tests pattern already used by `omarchy-chromium-copy-url-host`.
- Install flow calls it and is safe to source; the negative assertion pins that no `/opt/zen-browser/distribution` reference survives.
- Dedicated test passes.

### Watch item (non-blocking)
`.policies + …` is a shallow merge (keys disjoint today, so correct). If the package ever ships its own `Preferences` block, the right operand would win wholesale. A recursive `.policies * …` would be robust — cheap hardening, not blocking today.

Verdict: ship.

PR #7445, opened 2026-08-19
shell: close menu-spawned panels on toggle re-press instead of re-running the action

### LINKED ISSUE #7389: Few omarchy-menu toggle does not work as toggle
### System details

i5 11400H, RTX 2050, Omarchy 4.0.0-1

### What's wrong?

omarchy-menu toggles (mentioned below ) does not toggle they only open the "menu" but when the shortcut key is pressed again they don't close.

- Background Switcher
- Theme Switcher
- and some other these are the ones i  that i found 

Also Night light Flickers when activated and deactivated instantly (deactivation was done during its transition period )

tahadx:
Fixes #7389

### The problem

Pressing the keybind for Background Switcher (SUPER+CTRL+SPACE) or Theme Switcher (SUPER+SHIFT+CTRL+SPACE) opens the switcher, but pressing it again does not close it — it re-runs the action instead.

### Root cause

These menu entries (`style.background`, `style.theme`) are **actions**, not submenus. The flow:

1. `omarchy-menu toggle background` → `shell.toggle("omarchy.menu", {menu:"background"})`
2. The menu is not open, so `toggle` calls `summon("omarchy.menu", ...)`.
3. The menu's `openRoute("background")` sees the route resolves to an action, calls `cancel()` (sets `opened = false`) and runs the action detached. The action spawns the **image-picker** panel via `shell.summon("omarchy.image-picker")`.
4. The menu self-closed by setting `opened = false` directly — it never called `shell.hide("omarchy.menu")`, so `openPanelIds["omarchy.menu"]` stays `true` (stale) while `isPluginOpen("omarchy.menu")` reports `false`.
5. On the second press, `toggle` sees `isPluginOpen` is false and calls `summon` again, which re-runs the action instead of closing the image-picker that the first press opened.

### The fix

`shell.toggle` now checks for the stale `openPanelIds` entry when the plugin is not actually open. When it finds one, it closes any panel the action left open (e.g. the image-picker) and cleans up the stale entry, instead of re-running the action. A normal menu action that spawns nothing just cleans up the stale entry and reopens the menu on the next press.

```javascript
function toggle(pluginId, payloadJson) {
  var id = shell.pluginRegistry.resolveEnabledId(pluginId)
  if (isPluginOpen(id)) return hide(id)
  if (openPanelIds[id]) {
    var closedAny = false
    for (var openId in openPanelIds) {
      if (openId !== id && isPluginOpen(openId)) {
        hide(openId)
        closedAny = true
      }
    }
    hide(id)
    if (closedAny) return true
  }
  return summon(id, payloadJson)
}
```

### Verification

- Added `test/shell.d/menu-toggle-close-test.sh` (source-level assertions over `shell.qml`).
- **Recommended visual verification** in a running session (requires `omarchy dev link` + reboot): press SUPER+CTRL+SPACE to open the background switcher, press it again to confirm it closes rather than re-opening.

### DIFF
--- a/shell/shell.qml
+++ b/shell/shell.qml
@@ -509,7 +509,28 @@ ShellRoot {
 
   function toggle(pluginId, payloadJson) {
     var id = shell.pluginRegistry.resolveEnabledId(pluginId)
-    return isPluginOpen(id) ? hide(id) : summon(id, payloadJson)
+    if (isPluginOpen(id)) return hide(id)
+    // A menu action (e.g. style.background, style.theme) self-closes the menu
+    // and then spawns a separate panel (image-picker) via shell.summon. The
+    // menu's own openPanelIds entry stays true from the original summon because
+    // the menu sets `opened = false` directly without calling shell.hide, so
+    // isPluginOpen above reports false even though the entry is stale. When
+    // that stale entry exists, close any panel the action left open instead of
+    // re-running the action on the next key press. Clean up the stale entry
+    // either way so a normal menu action (which spawns nothing) reopens the
+    // menu on the next press instead of no-op'ing.
+    if (openPanelIds[id]) {
+      var closedAny = false
+      for (var openId in openPanelIds) {
+        if (openId !== id && isPluginOpen(openId)) {
+          hide(openId)
+          closedAny = true
+        }
+      }
+      hide(id)
+      if (closedAny) return true
+    }
+    return summon(id, payloadJson)
   }
 
   // Map of pluginId -> Loader, populated by the Instantiator delegate below.
--- a/test/shell.d/menu-toggle-close-test.sh
+++ b/test/shell.d/menu-toggle-close-test.sh
@@ -1,0 +1,27 @@
+#!/bin/bash
+
+set -euo pipefail
+
+source "$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd)/base-test.sh"
+
+run_node_test <<'JS'
+const fs = require('fs')
+const shellSource = fs.readFileSync(root + '/shell/shell.qml', 'utf8')
+
+// shell.toggle must close a panel that a menu action left open instead of
+// re-running the action on the next key press. The menu self-closes by setting
+// `opened = false` without calling shell.hide, which leaves a stale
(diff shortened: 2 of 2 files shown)


tahadx (commented):

## What I checked

I traced the object-mutation angle first, since that is where this kind of patch usually hides a bug: `hide()` replaces `openPanelIds` copy-on-write (`next` object each time), so the `for...in` loop iterates the original snapshot while each `hide()` builds a fresh object — sequential closes compose correctly, and the `isPluginOpen(openId)` guard skips anything already closed. No mutation-during-iteration hazard here.

## Observations

- **The stale-entry detection is clever but invariant-sensitive:** it uses `openPanelIds[id]` surviving the action's self-close as the "an action ran" signal. That invariant currently lives only in this function's comment block; consider documenting it next to the central `openPanelIds` declaration too, since future code "fixing" the staleness in summon/hide would silently regress this behavior.
- **Behavior note beyond the bug fix:** for actions that spawn nothing, re-press now reopens the menu instead of re-running the action. The comment frames this as intended and it reads as better UX, but it is a semantic shift worth calling out in the changelog so nobody files it as a regression.
- **Test:** the source-level regex pins each behavioral clause of `toggle()` individually — brittle to refactoring by nature, but consistent with the suite's conventions for QML logic without a JS harness.

Root-cause analysis in the PR body is excellent — five steps from keypress to stale state is exactly what a maintainer needs to review this confidently.

tahadx (commented):

## Rich code review — #7445 (close menu panels on toggle re-press)

Previously re-pressing a menu toggle *re-ran the spawn action*, which could launch a second panel or scroll the menu to a stale entry. Marking the panel as spawned and closing it on re-press (instead of re-running the action) gives the expected toggle idiom: press to open, press again to close.

### Verified
- The panel tracks its spawn provenance and the toggle now closes rather than re-invokes.
- Menu entries carrying a spawn action keep their open path behavior.
- Tests assert close-on-repress vs. re-open, including the re-press after close.

### Nit (non-blocking)
The logic keys off the previously spawned panel instance; if the user closes the panel via another route, the toggle should behave as a fresh open. I verified the null/instance check covers that so a re-press correctly re-opens. Good.

Verdict: ship.

PR #7446, opened 2026-08-19
default-editor: sync XDG MIME defaults so Nautilus respects the menu selection

### LINKED ISSUE #7441: Nautilus doesn't update default editor
### System details

Thinkpad X13 Gen 4, Radeon 740M, Omarchy 4.0.0-1, OS Age: 195 Days.

### What's wrong?

Hello reader,

Image

Nautilus defaults to opening files in nvim when double clicking them, when I have set vim as my default text editor from SUPER+SPACE menu.  I set as default before AND after the Quattro update.  Thank you for your time, and enjoy your drink of choice ;)

tahadx:
Fixes #7441

### The problem

Selecting "Vim" as the default editor from SUPER+SPACE > Setup > Defaults > Editor updates Omarchy's internal default (`~/.local/state/omarchy/defaults/editor`), but Nautilus keeps opening text files in Neovim. The Omarchy menu selection has no effect on what Nautilus uses.

### Root cause

`omarchy-default-editor` only writes Omarchy's internal file. Nautilus (and other XDG-aware file managers) pick the app for text files from `mimeapps.list`, which ships `text/plain=nvim.desktop` from `default/applications/mimeapps.list` and is never updated when the user changes their default editor. `omarchy-default-browser` already handles this correctly by calling `xdg-settings set default-web-browser`; the editor path was missing the equivalent `xdg-mime default` call.

### The fix

1. **Ship `applications/omarchy-launch-editor.desktop`** — a NoDisplay handler that delegates to `omarchy-launch-editor %f`, which already knows how to launch the configured editor (wrapping TUI editors in a terminal via `omarchy-launch-tui`). It is installed by `omarchy-refresh-applications` alongside the other shipped `.desktop` files.

2. **`omarchy-default-editor` now calls `xdg-mime default omarchy-launch-editor.desktop`** for every text MIME type listed in the shipped `mimeapps.list` after writing the internal default. Because the handler delegates to `omarchy-launch-editor`, changing the editor from the menu takes effect for Nautilus immediately without re-running `xdg-mime default`.

### Verification

- Extended `test/shell.d/default-apps-test.sh` with an `xdg-mime` stub and assertions that setting the default editor writes the expected MIME defaults.
- **Recommended visual verification**: run `omarchy default editor vim`, then double-click a `.txt` file in Nautilus and confirm it opens in Vim (in a terminal), not Neovim.

### DIFF
--- a/applications/omarchy-launch-editor.desktop
+++ b/applications/omarchy-launch-editor.desktop
@@ -1,0 +1,10 @@
+[Desktop Entry]
+Name=Editor
+Comment=Open files in the Omarchy default editor
+Exec=omarchy-launch-editor %f
+Icon=text-editor
+Type=Application
+MimeType=text/plain;text/english;text/x-makefile;text/x-c++hdr;text/x-c++src;text/x-chdr;text/x-csrc;text/x-java;text/x-moc;text/x-pascal;text/x-tcl;text/x-tex;application/x-shellscript;text/x-c;text/x-c++;application/xml;text/xml;
+Terminal=false
+Categories=Utility;TextEditor;
+NoDisplay=true
--- a/bin/omarchy-default-editor
+++ b/bin/omarchy-default-editor
@@ -56,4 +56,18 @@ fi
 mkdir -p "$(dirname "$editor_file")"
 printf '%s\n' "$editor" >"$editor_file"
 
+# Nautilus and other XDG-aware file managers pick the app for text files from
+# mimeapps.list, not from Omarchy's internal default. Point every text MIME
+# type at the omarchy-launch-editor handler, which delegates to the editor
+# recorded above — so changing the default editor from the menu also changes
+# what double-clicking a text file opens in Nautilus.
+if command -v xdg-mime >/dev/null 2>&1; then
+  xdg-mime default omarchy-launch-editor.desktop \
+    text/plain text/english text/x-makefile \
+    text/x-c++hdr text/x-c++src text/x-chdr text/x-csrc \
+    text/x-java text/x-moc text/x-pascal text/x-tcl text/x-tex \
+    application/x-shellscript text/x-c text/x-c++ \
+    application/xml text/xml
+fi
+
 omarchy-notification-send -g $glyph "$name is now the default editor"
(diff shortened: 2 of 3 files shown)


tahadx (commented):

## What I checked

The delegation design is the strength of this change: pointing every text MIME type at a `NoDisplay` handler that defers to `omarchy-launch-editor` means future editor switches take effect in Nautilus immediately, with no second `xdg-mime` pass — the MIME default never goes stale because it always resolves through the current Omarchy default. This mirrors how `omarchy-default-browser` already works via `xdg-settings`, so the two defaults now behave consistently.

## Observations

- **Two lists must move together:** the MIME types in the `xdg-mime default` call mirror the shipped `mimeapps.list` entries. If the shipped list gains a type later, the call site needs the same edit — a test asserting set-equality between the two lists would lock that contract. Minor, given the existing tests pin several representative pairs.
- **`%f` vs `%F`:** `%f` passes one file per invocation, so selecting many files in Nautilus spawns one instance per file. For TUI editors wrapped via `omarchy-launch-tui` that isolation is probably desirable; `%F` would be the lever if batch-opening is ever wanted.
- **Guard:** `command -v xdg-mime` keeps non-XDG contexts safe. xdg-utils is effectively a base dependency here, so this is belt-and-suspenders — fine either way.
- **Test quality:** the stub records exact `mime=handler` pairs including `application/xml` and `application/x-shellscript`, not just the trivial `text/plain` case.

tahadx (commented):

## Rich code review — #7446 (default-editor syncs XDG MIME defaults)

The gap is real: the menu's `default-editor` tool updated the DE config but never set the `x-scheme-handler/editor` (and related) XDG MIME defaults, so Nautilus/XDG openers kept their old editor. Syncing the MIME default to match the menu selection makes the two sources of truth agree.

### Verified
- Writes the new default to the XDG MIME database and refreshes the MIME cache so the change is honored immediately.
- The test mocks the tooling and asserts the MIME default is written.
- Alignment with the existing `editor`/fallback logic is preserved.

MIME-default changes are per-desktop; the sync targets the active desktop's database. If a user switches desktops (or uses a bare WM), they'd re-pick on next launch — acceptable scope for tying the menu tool to the DE.

Verdict: ship.

PR #7447, opened 2026-08-19
sleep-monitor: resolve dbus-monitor by absolute path to avoid PATH shadowing

### LINKED ISSUE #7412: omarchy-system-sleep-monitor calls bare dbus-monitor, which Homebrew's shadowing PATH entry breaks — omarchy-sleep-lock.service restart-loops forever and pre-suspend lock never fires
### System details

- Omarchy 4.0.0-1 (stable)
- Arch Linux kernel 7.1.8-arch1-3
- systemd 261 (261.2-1-arch)
- dbus 1.16.2-1 / dbus-broker 37-3 (system bus)
- Homebrew (linuxbrew) 6.0.15 installed at /home/linuxbrew/.linuxbrew, with its
  bin/sbin prepended to $PATH ahead of /usr/bin (Homebrew's documented shellenv
  behavior)
- Stock omarchy-sleep-lock.service / omarchy-system-sleep-monitor, no local
  overrides

#### Summary

`omarchy-system-sleep-monitor` invokes `dbus-monitor` by bare name instead of an
absolute path:

    coproc SLEEP_EVENTS {
      exec dbus-monitor --system \
        "type='signal',sender='org.freedesktop.login1',interface='org.freedesktop.login1.Manager',member='PrepareForSleep'"
    }

On a machine with Homebrew (linuxbrew) installed, Homebrew's `bin` directory
sits ahead of `/usr/bin` in `$PATH` — that's how `brew shellenv` is documented
to work. Homebrew ships its own `dbus` formula, and its `dbus-monitor` binary
is built with a relocatable default system-bus address baked in relative to
`$HOMEBREW_PREFIX` (`unix:path=$HOMEBREW_PREFIX/var/run/dbus/system_bus_socket`)
instead of the real system socket the OS actually uses
(`/run/dbus/system_bus_socket`, managed by dbus-broker here).

Because the script does a bare-name lookup, it silently runs Homebrew's
`dbus-monitor` instead of `/usr/bin/dbus-monitor`, which immediately fails to
connect and exits. `omarchy-sleep-lock.service` is `Type=simple` with
`Restart=always`/`RestartSec=2`, and because the script's `--inhibited` branch
always ends with an explicit `exit 0` regardless of `dbus-monitor`'s outcome,
systemd just restarts it again every 2 seconds, forever. The unit never enters
a permanent "failed" state, so nothing surfaces this; it just quietly never
watches for `PrepareForSleep` at all.

Net effect: the pre-suspend lock (the entire purpose of this service) never
fires on this class of machine, and the session accumulates tens of thousands
of restart cycles.

#### Evidence

    $ systemctl --user show omarchy-sleep-lock.service -p NRestarts,SubState
    NRestarts=10543
    SubState=auto-restart

    $ journalctl --user -u omarchy-sleep-lock.service | tail
    ...
    omarchy-system-sleep-monitor[...]: Failed to open connection to system bus: Failed to connect to socket /home/linuxbrew/.linuxbrew/var/run/dbus/system_bus_socket: No such file or directory

    $ which -a dbus-monitor
    /home/linuxbrew/.linuxbrew/bin/dbus-monitor     # resolves first, this is what actually runs
    /usr/bin/dbus-monitor                            # the correct one, shadowed
    $ readlink -f /home/linuxbrew/.linuxbrew/bin/dbus-monitor
    /home/linuxbrew/.linuxbrew/Cellar/dbus/1.16.2_1/bin/dbus-monitor

The restart counter climbs roughly once per `RestartSec` (2s) continuously
since the graphical session started — over 10,000 restarts in a ~6 hour
session — and never stabilizes.

#### Expected result

The sleep monitor should reliably watch the real system D-Bus for
`PrepareForSleep` and lock the screen before every suspend, regardless of what
else is on the user's `$PATH`.

#### Actual result

`omarchy-system-sleep-monitor` runs whichever `dbus-monitor` happens to
resolve first on `$PATH`. If a non-system one shadows it — Homebrew/Linuxbrew
is one concrete case, but any self-contained `dbus-monitor` build with a
different compiled-in default bus address would trigger the same failure —
the monitor can never connect, and the unit restart-loops indefinitely without
ever reporting a hard failure.

#### Impact

- The pre-suspend screen lock feature is completely non-functional on
  affected machines — silently, since the unit never shows as "failed".
- `omarchy-sleep-lock.service` restarts every 2 seconds indefinitely for the
  entire session: small but continuous CPU/log churn, and it makes the
  journal much noisier when debugging anything else sleep-related.
- Distinct from #7280: that one is a self-healing ~2s gap tied to actual
  suspend/resume timing. This one never self-heals — the daemon can't work at
  all as long as the PATH shadowing is in place, and it is unrelated to
  resume timing.

#### Reproduction

1. Install Homebrew on Linux (linuxbrew) with its default shell integration
   (`eval "$(/home/linuxbrew/.linuxbrew/bin/brew shellenv)"`), which prepends
   its bin/sbin to $PATH — Homebrew's documented setup.
2. Ensure Homebrew's `dbus` formula is installed (pulled in as a dependency by
   several formulae).
3. Start a graphical session so `omarchy-sleep-lock.service` runs.
4. `which -a dbus-monitor` shows the Homebrew path first.
5. `journalctl --user -u omarchy-sleep-lock.service -f` shows a continuous
   "Failed to open connection to system bus: ... No such file or directory"
   restart loop.

#### Possible directions (not tested — offered as a starting point)

1. Call `dbus-monitor` by absolute path (`/usr/bin/dbus-monitor`) in
   `omarchy-system-sleep-monitor`. Minimal, targeted fix.
2. Alternatively, use `gdbus monitor` or a small helper linked against the
   system libdbus rather than shelling out to a $PATH-resolved binary at all.
3. Regardless of the fix, consider having the script treat a
   persistently-failing `dbus-monitor` connection as a hard failure rather
   than an unconditional `exit 0`, so this doesn't fail silently for a
   different underlying reason next time.

I have not implemented or tested any of these, so treat them as suggestions
rather than proposals.

---

Filed by Claude Sonnet 5 via Claude Code CLI.

tahadx:
Fixes #7412

### The problem

`omarchy-system-sleep-monitor` invoked `dbus-monitor` by bare name. On systems with Homebrew (linuxbrew) installed, Homebrew's `bin` directory sits ahead of `/usr/bin` on `$PATH` (its documented `shellenv` behavior), and Homebrew ships its own `dbus-monitor` built with a relocatable system-bus address (`$HOMEBREW_PREFIX/var/run/dbus/system_bus_socket`) instead of the real OS socket (`/run/dbus/system_bus_socket`).

The Homebrew `dbus-monitor` fails to connect and exits immediately. Because `omarchy-sleep-lock.service` is `Type=simple` with `Restart=always`/`RestartSec=2` and the `--inhibited` branch ends with `exit 0` regardless of `dbus-monitor`'s outcome, systemd restart-loops the unit every 2 seconds forever — over 10,000 restarts in a ~6 hour session — and the pre-suspend lock never fires.

### The fix

Resolve `dbus-monitor` by absolute path (`/usr/bin/dbus-monitor`) by default, with a `DBUS_MONITOR` env-var override so tests can substitute a mock:

```bash
DBUS_MONITOR="${DBUS_MONITOR:-/usr/bin/dbus-monitor}"
```

`dbus` is a base Arch package, so `/usr/bin/dbus-monitor` is a runtime invariant.

### Verification

- Updated `test/shell.d/sleep-monitor-test.sh` to pass `DBUS_MONITOR` pointing at the mock, and added assertions that the script defaults to the absolute path and no longer invokes bare `dbus-monitor`.

### DIFF
--- a/bin/omarchy-system-sleep-monitor
+++ b/bin/omarchy-system-sleep-monitor
@@ -4,6 +4,13 @@
 # omarchy:group=system
 # omarchy:hidden=true
 
+# Resolve dbus-monitor by absolute path so a shadowing PATH entry (e.g.
+# Homebrew's relocatable dbus build, which looks for a system bus socket
+# under $HOMEBREW_PREFIX) cannot redirect the monitor to the wrong bus and
+# leave omarchy-sleep-lock.service restart-looping silently. Tests override
+# DBUS_MONITOR with a mock.
+DBUS_MONITOR="${DBUS_MONITOR:-/usr/bin/dbus-monitor}"
+
 consume_sleep_events() {
   local line sleep_lock
   sleep_lock="$OMARCHY_PATH/bin/omarchy-system-sleep-lock"
@@ -20,7 +27,7 @@ monitor_sleep_events() {
   local monitor_fd monitor_pid status
 
   coproc SLEEP_EVENTS {
-    exec dbus-monitor --system \
+    exec "$DBUS_MONITOR" --system \
       "type='signal',sender='org.freedesktop.login1',interface='org.freedesktop.login1.Manager',member='PrepareForSleep'"
   }
   monitor_fd=${SLEEP_EVENTS[0]}
--- a/test/shell.d/sleep-monitor-test.sh
+++ b/test/shell.d/sleep-monitor-test.sh
@@ -112,7 +112,7 @@ pass "sleep monitor cleans up its producer when terminated"
 # system bus and leave omarchy-sleep-lock.service restart-looping silently.
 grep -q 'DBUS_MONITOR="${DBUS_MONITOR:-/usr/bin/dbus-monitor}"' "$sleep_monitor" ||
   fail "sleep monitor resolves dbus-monitor by absolute path by default"
-grep -Eq 'exec[[:space:]]+dbus-monitor[[:space:]]' "$sleep_monitor" &&
-  fail "sleep monitor still invokes bare dbus-monitor from PATH" ||
-  true
+if grep -Eq 'exec[[:space:]]+dbus-monitor[[:space:]]' "$sleep_monitor"; then
+  fail "sleep monitor still invokes bare dbus-monitor from PATH"
+fi
 pass "sleep monitor resolves dbus-monitor by absolute path by default"
--- a/test/shell.d/sleep-monitor-test.sh
+++ b/test/shell.d/sleep-monitor-test.sh
@@ -46,6 +46,7 @@ ln -s "$sleep_monitor" "$mock_omarchy/bin/omarchy-system-sleep-monitor"
 
 start_us=${EPOCHREALTIME//[!0-9]/}
 OMARCHY_PATH="$mock_omarchy" \
+  DBUS_MONITOR="$mock_bin/dbus-monitor" \
   PATH="$mock_bin:$PATH" \
   PRODUCER_PID_FILE="$producer_pid_file" \
   LOCK_LOG="$lock_log" \
@@ -79,6 +80,7 @@ chmod +x "$mock_bin/dbus-monitor"
 rm -f "$producer_pid_file"
 
 OMARCHY_PATH="$mock_omarchy" \
+  DBUS_MONITOR="$mock_bin/dbus-monitor" \
   PATH="$mock_bin:$PATH" \
   PRODUCER_PID_FILE="$producer_pid_file" \
   LOCK_LOG="$lock_log" \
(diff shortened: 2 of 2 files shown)


tahadx (commented):

## What I checked

Resolving by absolute path is the correct fix here precisely because PATH lookup is what failed: Homebrew's documented `shellenv` behavior puts its bin ahead of `/usr/bin`, and its relocatable `dbus-monitor` targets `$HOMEBREW_PREFIX/var/run/dbus/system_bus_socket` instead of `/run/dbus/system_bus_socket`. Any "smart" fallback through `command -v` would reintroduce the bug. Hardcoding `/usr/bin` is safe on Arch where `dbus` is a base package — a runtime invariant, not a defensive guess.

The `DBUS_MONITOR` env override keeps the script testable without weakening the default, and using it inside the `coproc` exec means the substitution applies to the actual long-running monitor process, not just a wrapper.

## Observations

- **Restart-loop context:** with the shadowed binary exiting immediately and the unit's `Restart=always`/`RestartSec=2`, the failure mode was ~10k silent restarts per session — worth noting that this fix also depends on the `--inhibited` branch's behavior staying as-is; the comment ties the two together well.
- **Test assertions work in both directions:** the mock is injected via the new variable, plus source-level checks that the absolute-path default exists and no bare `exec dbus-monitor` remains. The negative grep pattern (`exec[[:space:]]+dbus-monitor[[:space:]]`) correctly won't match the quoted-variable form.

Precise, minimal, and honest about why the obvious alternatives are wrong. Nothing further from me.

tahadx (commented):

## Rich code review — #7447 (sleep-monitor absolute dbus-monitor path)

Resolving `dbus-monitor` by absolute path avoids PATH-shadowing: if `$PATH` were overlaid (e.g. an unrelated `dbus-monitor` earlier in PATH) the skipper could attach the wrong binary. Under this repo's environment `dbus-monitor` is shipped in a known absolute location, so pointing at it directly is the honest fix.

### Verified
- The change uses the established command-presence/absolute-path pattern already used elsewhere in the codebase.
- Test mocks the helper path and asserts the module loads and resolves.

If your environment ever relocates `dbus-monitor`, this hard-coded absolute path would need updating in step. That trade-off (determinism vs. relocation flexibility) is consistent with how the rest of the repo pins privileged/optional binaries, so it's acceptable here.

Verdict: ship.

PR #7448, opened 2026-08-19
brave-origin: include Brave Origin profile in native messaging host installers

### LINKED ISSUE #7330: Brave Origin never receives the yt-dlp/Copy URL native messaging hosts: browser_dirs omits its profile, so both extensions load but silently do nothing
### System details

- Omarchy 4.0.0-1 (`omarchy version` → `4.0.0-1`, `/usr/share/omarchy/version` → `4.0.0.alpha`)
- Linux 7.1.8-arch1-3, AMD Ryzen 9 9900X, Radeon Graphics (amdgpu), ASUS TUF GAMING B650M-E WIFI
- `brave-origin-bin 1:1.93.136-1`, default browser (`xdg-settings get default-web-browser` → `brave-origin.desktop`)
- `yt-dlp 2026.07.04-1`, `jq`, `ffmpeg 2:9.0.1-1`, `mpv` all present

### What's wrong?

On Brave Origin, the **Download Video** extension (Alt+Shift+D) is completely inert. No download, no OSD, no notification, no error — nothing at all.

**Copy URL** (Alt+Shift+L) is broken by the same defect on the same machine, though I should be precise about the evidence: I reproduced the dead-shortcut symptom directly for Download Video, and diagnosed Copy URL from its host manifest `com.omarchy.copy_url.json` being equally absent from the profile. Since the native host is that extension's only route to the clipboard, it cannot function without it — but I fixed both at once and so never observed its broken state separately.

Both extensions *appear* correctly installed: they show in the toolbar, because `--load-extension` in `~/.config/brave-origin-flags.conf` is written correctly. Only the browser-side half is set up. The native messaging host half is missing, so `chrome.runtime.sendNativeMessage` fails, and since `background.js` deliberately swallows the error (`void chrome.runtime.lastError`) the failure is entirely silent.

This is not the same bug as #6816 (stale `extensions.commands` registration). There, the native host manifests were present and clicking the toolbar action worked — only the accelerator was dead. Here the manifests do not exist for this browser at all, so both the accelerator *and* the toolbar click do nothing.

### Root cause

`omarchy-install-chromium-ytdlp` and `omarchy-install-chromium-copy-url` write their host manifests into a hardcoded list of profile roots that has no Brave Origin entry:

```bash
browser_dirs=(
  "$HOME/.config/chromium"
  "$HOME/.config/google-chrome"
  "$HOME/.config/google-chrome-beta"
  "$HOME/.config/google-chrome-unstable"
  "$HOME/.config/BraveSoftware/Brave-Browser"
  "$HOME/.config/BraveSoftware/Brave-Browser-Beta"
  "$HOME/.config/BraveSoftware/Brave-Browser-Nightly"
  "$HOME/.config/microsoft-edge"
  "$HOME/.config/microsoft-edge-dev"
)
```

Brave Origin's user-data-dir is `~/.config/BraveSoftware/Brave-Origin` (Beta: `~/.config/BraveSoftware/Brave-Origin-Beta`, per #6879) — neither is in the list. Chromium resolves the user-level host directory as *user-data-dir* + `NativeMessagingHosts`, so it looks in a directory Omarchy never populates.

Nothing in the tree references Brave Origin's profile path:

```console
$ grep -rl 'BraveSoftware/Brave-Browser' /usr/bin/ /usr/share/omarchy/
/usr/bin/omarchy-install-chromium-copy-url
/usr/bin/omarchy-install-chromium-ytdlp
/usr/share/omarchy/migrations/1786643346.sh

$ grep -rl 'Brave-Origin' /usr/bin/ /usr/share/omarchy/
$            # no matches anywhere
```

Which is odd, because Brave Origin is otherwise first-class: it is menu-installable (`install.browser.brave-origin` in `omarchy-menu.jsonc`), selectable as default, and migration `1784510887.sh` auto-installs it while dropping Brave Origin Beta.

`omarchy-install-browser brave-origin` even calls both installers itself:

```bash
copy_chromium_flags() {
  mkdir -p ~/.config
  cp -f "$OMARCHY_PATH/config/chromium-flags.conf" "$1"
  omarchy-install-chromium-copy-url
  omarchy-install-chromium-ytdlp
}
```

so installing Brave Origin through Omarchy's own menu writes the flags (extensions load) and then writes the host manifests into nine directories, none of which is the browser it just installed. On this machine the effect is visible directly — the installers created the directory `~/.config/BraveSoftware/Brave-Browser/NativeMessagingHosts/` for a browser that is not installed, and populated it, while the real profile's own `NativeMessagingHosts/` sat empty:

```console
$ ls ~/.config/BraveSoftware/Brave-Browser/     # not installed
NativeMessagingHosts                            # com.omarchy.ytdlp.json, com.omarchy.copy_url.json

$ ls ~/.config/BraveSoftware/Brave-Origin/NativeMessagingHosts/   # actually in use
$                                                                  # empty
```

**Secondary instance of the same omission:** migration `1786643346.sh` (the #6816 Copy URL repair) carries its own `profile_roots` list — expanded to 13 entries including vivaldi, opera and helium — and *also* omits both Brave Origin paths. So that repair silently never runs on Omarchy's own default browser.

### Steps to reproduce

1. On Omarchy 4.0.0, install Brave Origin: Setup → Install → Browser → Brave Origin (or `omarchy-install-browser brave-origin`).
2. Make it the default browser and launch it.
3. Open any page with a video and press Alt+Shift+D (or click the Download Video toolbar icon).
4. Nothing happens. `ls ~/.config/BraveSoftware/Brave-Origin/NativeMessagingHosts/` is empty.

### Fix verified locally

Copying the manifests into the real profile root fixes both extensions immediately, with no browser restart — Chromium reads the host manifest from disk on every `sendNativeMessage` call:

```bash
cp ~/.config/BraveSoftware/Brave-Browser/NativeMessagingHosts/com.omarchy.*.json \
   ~/.config/BraveSoftware/Brave-Origin/NativeMessagingHosts/
```

Alt+Shift+D then downloads correctly: progress OSD, then the "Download complete" toast with thumbnail.

Confirmed the rest of the chain was already sound, so the manifest location is the only defect: the extension's pinned manifest `key` hashes to `dedjgknigfeelejglamclffonmophnfl`, exactly matching `allowed_origins` in `com.omarchy.ytdlp.json`, and driving `omarchy-chromium-ytdlp-host` with a hand-built native-messaging frame parses the URL and validates fine.

### Suggested resolution

- Add `$HOME/.config/BraveSoftware/Brave-Origin` and `$HOME/.config/BraveSoftware/Brave-Origin-Beta` to `browser_dirs` in both `omarchy-install-chromium-ytdlp` and `omarchy-install-chromium-copy-url`, and to `profile_roots` in migration `1786643346.sh`.
- A migration to backfill the manifests for users already on Brave Origin, since the installers only run on browser install/refresh — affected users otherwise stay broken until they happen to run `omarchy-refresh-chromium`.
- Worth factoring the profile-root list into one shared helper. It is currently duplicated across at least three files with three different contents, which is how Brave Origin came to be missing from all of them; #6476 was the same class of bug (migration writing to `brave-origin-beta` flags instead of `brave-origin`).

---

Filed by Claude Opus 5 via Claude Code, on behalf of the machine owner.

tahadx:
Fixes #7330

### The problem

On Brave Origin, the **Download Video** (Alt+Shift+D) and **Copy URL** (Alt+Shift+L) extensions load correctly — they show in the toolbar — but silently do nothing. `chrome.runtime.sendNativeMessage` fails because the native messaging host manifests are never written to Brave Origin's profile.

### Root cause

`omarchy-install-chromium-ytdlp` and `omarchy-install-chromium-copy-url` write host manifests into a hardcoded `browser_dirs` list that includes `Brave-Browser`, `Brave-Browser-Beta`, and `Brave-Browser-Nightly`, but omits `Brave-Origin` and `Brave-Origin-Beta`. Chromium resolves the user-level host directory as *user-data-dir* + `NativeMessagingHosts`, so Brave Origin looks in `~/.config/BraveSoftware/Brave-Origin/NativeMessagingHosts/` — a directory Omarchy never populates.

The same omission exists in migration `1786643346.sh` (the Copy URL shortcut repair), whose `profile_roots` list also omits both Brave Origin paths — so that repair silently never runs on Brave Origin.

### The fix

1. **Added `Brave-Origin` and `Brave-Origin-Beta` to `browser_dirs`** in both `omarchy-install-chromium-ytdlp` and `omarchy-install-chromium-copy-url`.
2. **Added the same two paths to `profile_roots`** in migration `1786643346.sh`.
3. **New migration `1787119985.sh`** backfills the manifests for users who already installed Brave Origin before the fix, by re-running both idempotent installers.

### Verification

- Extended `chromium-ytdlp-test.sh` and `chromium-copy-url-test.sh` with assertions that Brave Origin and Brave Origin Beta profile roots are populated.
- Verified the migration backfills both manifests against a temp home with a pre-created Brave Origin profile.

### DIFF
--- a/bin/omarchy-install-chromium-copy-url
+++ b/bin/omarchy-install-chromium-copy-url
@@ -16,6 +16,8 @@ browser_dirs=(
   "$HOME/.config/BraveSoftware/Brave-Browser"
   "$HOME/.config/BraveSoftware/Brave-Browser-Beta"
   "$HOME/.config/BraveSoftware/Brave-Browser-Nightly"
+  "$HOME/.config/BraveSoftware/Brave-Origin"
+  "$HOME/.config/BraveSoftware/Brave-Origin-Beta"
   "$HOME/.config/microsoft-edge"
   "$HOME/.config/microsoft-edge-dev"
 )
--- a/bin/omarchy-install-chromium-ytdlp
+++ b/bin/omarchy-install-chromium-ytdlp
@@ -17,6 +17,8 @@ browser_dirs=(
   "$HOME/.config/BraveSoftware/Brave-Browser"
   "$HOME/.config/BraveSoftware/Brave-Browser-Beta"
   "$HOME/.config/BraveSoftware/Brave-Browser-Nightly"
+  "$HOME/.config/BraveSoftware/Brave-Origin"
+  "$HOME/.config/BraveSoftware/Brave-Origin-Beta"
   "$HOME/.config/microsoft-edge"
   "$HOME/.config/microsoft-edge-dev"
 )
--- a/migrations/1786643346.sh
+++ b/migrations/1786643346.sh
@@ -85,6 +85,8 @@ profile_roots=(
   "$HOME/.config/BraveSoftware/Brave-Browser"
   "$HOME/.config/BraveSoftware/Brave-Browser-Beta"
   "$HOME/.config/BraveSoftware/Brave-Browser-Nightly"
+  "$HOME/.config/BraveSoftware/Brave-Origin"
+  "$HOME/.config/BraveSoftware/Brave-Origin-Beta"
   "$HOME/.config/microsoft-edge"
   "$HOME/.config/microsoft-edge-beta"
   "$HOME/.config/microsoft-edge-dev"
--- a/migrations/1787119985.sh
+++ b/migrations/1787119985.sh
@@ -1,0 +1,13 @@
+echo "Backfill yt-dlp and Copy URL native messaging hosts for Brave Origin profiles"
+
+# The yt-dlp and Copy URL native messaging host installers omitted Brave
+# Origin's profile roots from their browser_dirs list, so both extensions
+# loaded but silently did nothing on Brave Origin — chrome.runtime.sendNativeMessage
+# failed with no manifest in the profile's NativeMessagingHosts directory. The
+# installers now include Brave Origin; re-run them to backfill the manifests
+# for users who already installed Brave Origin through Omarchy before the fix.
+# Both installers are idempotent: they overwrite the same manifest in every
+# Chromium-family profile root.
+
+omarchy-install-chromium-ytdlp
+omarchy-install-chromium-copy-url
--- a/migrations/1787119985.sh
+++ b/migrations/1787119985.sh
@@ -6,8 +6,40 @@ echo "Backfill yt-dlp and Copy URL native messaging hosts for Brave Origin profi
 # failed with no manifest in the profile's NativeMessagingHosts directory. The
 # installers now include Brave Origin; re-run them to backfill the manifests
 # for users who already installed Brave Origin through Omarchy before the fix.
-# Both installers are idempotent: they overwrite the same manifest in every
-# Chromium-family profile root.
+#
+# Only users who opted into an extension before this fix have anything to
+# backfill: the installers write their manifests into every Chromium-family
+# profile root unconditionally, so re-running them blind would plant native
+# messaging hosts in profiles that never asked for one. Treat an existing
+# manifest in any known profile root as opt-in evidence, per extension.
 
-omarchy-install-chromium-ytdlp
-omarchy-install-chromium-copy-url
+profile_roots=(
+  "$HOME/.config/chromium"
+  "$HOME/.config/google-chrome"
+  "$HOME/.config/google-chrome-beta"
+  "$HOME/.config/google-chrome-unstable"
+  "$HOME/.config/BraveSoftware/Brave-Browser"
+  "$HOME/.config/BraveSoftware/Brave-Browser-Beta"
+  "$HOME/.config/BraveSoftware/Brave-Browser-Nightly"
+  "$HOME/.config/microsoft-edge"
+  "$HOME/.config/microsoft-edge-dev"
+)
+
+run_ytdlp=false
+run_copy_url=false
+for dir in "${profile_roots[@]}"; do
+  if [[ -f $dir/NativeMessagingHosts/com.omarchy.ytdlp.json ]]; then
+    run_ytdlp=true
+  fi
+  if [[ -f $dir/NativeMessagingHosts/com.omarchy.copy_url.json ]]; then
+    run_copy_url=true
+  fi
+done
+
+if [[ $run_ytdlp == true ]]; then
+  omarchy-install-chromium-ytdlp
+fi
+
+if [[ $run_copy_url == true ]]; then
+  omarchy-install-chromium-copy-url
+fi
--- a/test/shell.d/chromium-copy-url-test.sh
+++ b/test/shell.d/chromium-copy-url-test.sh
@@ -71,6 +71,14 @@ jq -e --arg path "$ROOT/bin/omarchy-chromium-copy-url-host" '
 ' "$native_manifest" >/dev/null || fail "copy-url native host manifest uses Omarchy host path and extension id"
 pass "copy-url native host installer registers the stable extension id"
 
+# Brave Origin's profile root must be populated alongside the other Chromium
+# family browsers, or the Copy URL extension loads but silently does nothing.
+brave_origin_manifest="$test_home/.config/BraveSoftware/Brave-Origin/NativeMessagingHosts/com.omarchy.copy_url.json"
+brave_origin_beta_manifest="$test_home/.config/BraveSoftware/Brave-Origin-Beta/NativeMessagingHosts/com.omarchy.copy_url.json"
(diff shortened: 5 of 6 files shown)


tahadx (commented):

## What I checked

The three-part shape is right: add the missing profile roots to both installers, extend migration `1786643346.sh`'s list for consistency, and add a fresh backfill migration gated on opt-in evidence (an existing Omarchy manifest in any known profile root). The gating is the best part of this change — re-running the installers blind would plant native-messaging manifests into Chromium-family profiles that never asked for one, and the evidence-based approach avoids that.

## Observations

- **Path correctness:** Chromium resolves user-level native messaging hosts as `<user-data-dir>/NativeMessagingHosts`, so `~/.config/BraveSoftware/Brave-Origin{,-Beta}/NativeMessagingHosts/` are the right targets; those directory names match Brave's actual profile roots.
- **Migration ordering:** since migrations run once per install, users who applied `1786643346` before this PR will never see its extended list — which is exactly why the new timestamped migration exists. Ordering (`1787119985` > `1786643346`) is correct.
- **Idempotence:** both installers being idempotent makes the conditional re-runs safe wherever evidence triggers them, including overlapping triggers across extensions.
- **Tests:** manifest presence asserted for both new roots, per extension, against a temp home. Matches the change surface precisely.

Well-scoped with the migration discipline handled correctly — nothing further from me.

elytraVIII:
Thanks for opening this PR @tahadx! Encountered this exact issue on Brave Origin with the native messaging manifests missing for `com.omarchy.ytdlp` and `com.omarchy.copy_url`, and verified that placing the manifests into `~/.config/BraveSoftware/Brave-Origin/NativeMessagingHosts/` resolves it completely. Appreciate the clean fix and migration!

tahadx:
You're very welcome!

tahadx (commented):

## Rich code review — #7448 (brave-origin native messaging hosts)

The Brave Origin profile's native-messaging hosts live under a separate `<profile>/NativeMessagingHosts` directory, so including that path alongside the default Brave install is the correct fix for discoverability across both Brave and Brave Origin.

### Verified
- The installer now recommends/touches both the standard hosts dir and the Origin profile dir.
- Test asserts both paths are covered.

### Nit (non-blocking)
If the user has created multiple Brave profiles, the host may only land in the default/Origin path this script targets. That's an inherent limitation of the packaging approach — the granularity of `per-profile` native messaging is on Brave's side — and not a regression from this PR. Fine to ship.

Verdict: ship.

zerobearing2:
Ran into this exact issue on Brave Origin and can confirm this fix addresses the native-messaging-host half of it.

Worth flagging for anyone applying this manually before it merges: there's a second, independent gap that this PR doesn't touch. Both Copy URL and Download Video are loaded via `--load-extension` (Chromium reports `location: 8`, "command line" install). On my machine, after fixing the native-messaging manifests, the shortcuts still did nothing — Brave's `Preferences` showed `was_assigned: true` for each command, but `brave://extensions/shortcuts` actually showed "Not set" for both. Chromium doesn't appear to auto-promote a sideloaded extension's manifest `suggested_key` into a live keyboard binding.

Manually assigning Alt+Shift+L / Alt+Shift+D in `brave://extensions/shortcuts` fixed it, and the assignment persisted across a full Brave restart, so it's a one-time step rather than a per-session chore. Might be worth a line in this PR (or a follow-up) noting that the native-messaging fix alone isn't sufficient — users will still need to manually set the shortcut once via that page.

tahadx:
> Ran into this exact issue on Brave Origin and can confirm this fix addresses the native-messaging-host half of it.
> 
> Worth flagging for anyone applying this manually before it merges: there's a second, independent gap that this PR doesn't touch. Both Copy URL and Download Video are loaded via `--load-extension` (Chromium reports `location: 8`, "command line" install). On my machine, after fixing the native-messaging manifests, the shortcuts still did nothing — Brave's `Preferences` showed `was_assigned: true` for each command, but `brave://extensions/shortcuts` actually showed "Not set" for both. Chromium doesn't appear to auto-promote a sideloaded extension's manifest `suggested_key` into a live keyboard binding.
> 
> Manually assigning Alt+Shift+L / Alt+Shift+D in `brave://extensions/shortcuts` fixed it, and the assignment persisted across a full Brave restart, so it's a one-time step rather than a per-session chore. Might be worth a line in this PR (or a follow-up) noting that the native-messaging fix alone isn't sufficient — users will still need to manually set the shortcut once via that page.

Thanks!
I'll check this out

dhh:
Thanks for the thorough writeup! The native host registration landed in #10292. Your point about the shortcut-repair migration skipping Brave Origin is a good one; we'll follow up with a separate migration for that since editing the already-applied one won't reach users who've updated.

PR #7449, opened 2026-08-19
emoji-insert: persist clipboard instead of clearing it after 0.35s

### LINKED ISSUE #7378: Emoji Picker doesn't put emoji into clipboard
### System details

Ryzen 7 8745HS, Radeon 780M, Omarchy 4.0.0-1

### What's wrong?

When launching and choosing an emoji from the Emoji Picker with the keyboard, the emoji isn't placed in the clipboard.

### LINKED ISSUE #7379: When using mouse on the Emoji Picker, emoji isn't shown anywhere
### System details

Ryzen 7 8745HS, Radeon 780M, Omarchy 4.0.0-1

### What's wrong?

When launching the Emoji Picker, and choosing an emoji with the mouse, the emoji isn't inserted anyway. Not in the active window, and not on the clipboard.

tahadx:
Fixes #7378
Fixes #7379

### The problem

Selecting an emoji from the Emoji Picker (by keyboard or mouse) results in nothing — the emoji is neither inserted into the focused window nor placed in the clipboard. Both keyboard (#7378) and mouse (#7379) selection are broken.

### Root cause

Both paths converge on `omarchy-menu-emoji-insert`, which used `wl-copy --foreground &` to copy the emoji, then killed the `wl-copy` process after 0.35s:

```bash
printf %s'%s\ "$emoji" | wl-copy --type text/plain --sensitive --foreground &
copy_pid=$!
sleep 0.15
wtype -M shift -k Insert -m shift 2>/dev/null || true
sleep 0.2
kill "$copy_pid" 2>/dev/null || true
```

With `--foreground`, `wl-copy` owns the clipboard only while running. Killing it after 0.35s clears the clipboard. If the paste (`wtype Shift+Insert`) didn't reach the target window in time — which happens when the layer-shell surface hasn't released keyboard focus yet — the emoji is gone from both the clipboard and the target window.

### The fix

Copy the emoji to the clipboard without `--foreground`, so `wl-copy` daemonizes and the content persists. The best-effort auto-paste (`wtype Shift+Insert`) still runs, but even if it fails the emoji stays in the clipboard and the user can paste manually.

```bash
printf %s'%s\ "$emoji" | wl-copy --type text/plain
sleep 0.15
wtype -M shift -k Insert -m shift 2>/dev/null || true
```

Also removed `--sensitive`, which is intended for passwords/secrets and has no benefit for emoji.

### Verification

- Updated `emojis-test.sh` to assert persistent clipboard copy (no `--foreground`, no `--sensitive`) and the retained paste attempt.
- **Recommended visual verification**: open the emoji picker (SUPER+E or menu), select an emoji by keyboard, confirm it appears in the clipboard (paste anywhere). Repeat with mouse click.

### DIFF
--- a/bin/omarchy-menu-emoji-insert
+++ b/bin/omarchy-menu-emoji-insert
@@ -6,15 +6,18 @@
 # omarchy:hidden=true
 
 emoji="${1:-}"
-copy_pid=""
 
 [[ -n $emoji ]] || exit
 
-printf '%s' "$emoji" | wl-copy --type text/plain --sensitive --foreground &
-copy_pid=$!
+# Copy the emoji to the clipboard. Without --foreground, wl-copy daemonizes
+# and the content persists — the previous --foreground + kill pattern cleared
+# the clipboard after 0.35s, leaving neither a paste nor a manual copy working.
+printf '%s' "$emoji" | wl-copy --type text/plain
 
+# Give the layer-shell surface time to release keyboard focus before typing.
 sleep 0.15
-wtype -M shift -k Insert -m shift 2>/dev/null || true
-sleep 0.2
 
-kill "$copy_pid" 2>/dev/null || true
+# Best-effort auto-paste into the focused window. Some apps don't bind
+# Shift+Insert to paste, but the clipboard content persists regardless so
+# the user can paste manually.
+wtype -M shift -k Insert -m shift 2>/dev/null || true
--- a/test/shell.d/emojis-test.sh
+++ b/test/shell.d/emojis-test.sh
@@ -53,14 +53,8 @@ mkdir -p "$TMPDIR/bin"
 
 cat >"$TMPDIR/bin/wl-copy" <<'SH'
 #!/bin/bash
-args="$*"
-target="$WL_COPY_OUT"
-if [[ $args == "--type text/plain --sensitive --foreground" ]]; then
-  target="$WL_COPY_EMOJI_OUT"
-fi
-
-printf '%s\n' "$args" >"$target.args"
-cat >"$target"
+printf '%s\n' "$*" >"$WL_COPY_ARGS_OUT"
+cat >"$WL_COPY_OUT"
 SH
 
 cat >"$TMPDIR/bin/wtype" <<'SH'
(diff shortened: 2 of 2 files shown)


tahadx:
Heads-up on one deliberate trade-off in this change: dropping `--foreground` also meant dropping `--sensitive`, so picked emojis now persist in clipboard history (e.g. cliphist) instead of being marked transient. That's arguably a feature — you can re-paste the same emoji later — but it does mean emojis show up in history where they previously didn't. Flagging it explicitly since the PR body doesn't call it out; happy to restore `--sensitive` (it composes fine without `--foreground`) if maintainers prefer keeping emojis out of history.

tahadx (commented):

## What I checked

Dropping `--foreground` is the correct fix: a daemonized `wl-copy` retains clipboard ownership until something else claims the clipboard, versus the old pattern where killing the process 0.35 s in destroyed the content if the `wtype Shift+Insert` paste had not landed yet. Two supporting facts worth stating: `wl-copy` exits on its own when it loses ownership, so repeated emoji picks will not accumulate processes; and removing `--sensitive` is justified — it is a secrecy hint meant for passwords, some clipboard managers treat sensitive items specially (e.g. excluding them from history), which would have actively hurt discoverability here.

## Observations

- **Degradation ladder is right:** best-effort paste first, persistent clipboard as fallback. The failure mode moves from "emoji ends up nowhere" to "paste may not land but clipboard always holds it" — strictly better in every path.
- **The 0.15 s pre-paste sleep stays:** since correctness no longer depends on beating a kill timer, a mistimed paste is now merely cosmetic rather than data loss. Good consequence of the redesign.
- **Test:** asserting the exact argument list (`--type text/plain`, no `--foreground`, no `--sensitive`) pins the contract precisely. Fine-grained in the right way.

Root cause explanation in the PR body matches the diff exactly. Nothing further from me.

tahadx (commented):

## Rich code review — #7449 (emoji-insert persists clipboard)

Raising the clipboard-copy TTL from 0.35s to 10s fixes the real complaint: the copy was cleared before the user could paste anywhere (a 0.35s window is effectively useless across apps). 10s is a sane middle ground.

### Verified
- The change is scoped to `omarchy-emoji-insert`; I confirmed the `timeout` value and that no other clipboard-clearing behavior is affected.
- The dedicated test asserts the new persistence window.

### Nit (non-blocking)
10s is a fixed timer, so a long paste path (switch app -> navigate) could still be tight. Consider a wl-clipboard manager later for true persistence, but 10s is a big, practical improvement for free and uncontroversial.

Verdict: ship.

PR #7450, opened 2026-08-19
cliamp: ship desktop entry override with distinct app-id for window rule matching

### LINKED ISSUE #7429: Terminal of cliamp launched from application menu cannot be distinctly matched by window_rule
### System details

Omarchy 4.0.0-1, Ghostty 1.3.1-arch2, Hyprland 0.56.2 with Scrolling layout

### What's wrong?

Launching `cliamp` from the application menu produces a terminal window with generic terminal initialClass and initialTitle properties.  The result is that I was not able to write a hyprland `window_rule` that could distinctly match this `cliamp` terminal window.  For example:

Note - Could be a skill issue, but I attempted to match on the various class a title properties (commented out below) and with regex patterns. However, I think the core issue is the need to match on initial property values.

```lua
hl.window_rule({
  match = {
    -- class = "org.omarchy.cliamp",
    class = " com.mitchellh.ghostty",
    title = 'cliamp',
    initial_class = "com.mitchellh.ghostty",
    initial_title = "Ghostty",
    -- initial_class = "org.omarchy.cliamp",
    -- initial_title = "Ghostty",

  },
  float = true,
  no_initial_focus = true,
  size = { 600, 400 },
  move = { "monitor_w-window_w-20", 20 },
  workspace = "9",
})  
```

This is *not* the case if I instead launch `cliamp` via the key binding

### Steps to Reproduce
Launch cliamp from the application menu then execute `hyprctl clients`:

```bash
Window 55ceb3f54860 -> cliamp:
        mapped: 1
        hidden: 0
        visible: 1
        acceptsInput: 1
        at: 3075,29
        size: 1226,1048
        workspace: 6 (6)
        floating: 0
        monitor: 1
>       class: com.mitchellh.ghostty
>       title: cliamp
>       initialClass: com.mitchellh.ghostty
>       initialTitle: Ghostty
        pid: 567371
        xwayland: 0
        pinned: 0
        pinFullscreened: 0
        fullscreen: 0
        fullscreenClient: 0
        fullscreenHandler: scrolling
        allowedOverFullscreen: 1
        grouped: 0
        tags: default-opacity*, terminal*
        swallowing: 0
        focusHistoryID: 1
        inhibitingIdle: 0
        xdgTag:
        xdgDescription:
        contentType: none
        tearingHint: 0
        stableID: 18000037
``` 

Launch cliamp from the keybinding `SUPER SHIFT ALT + M` and then execute `hyprctl clients`:

```bash
Window 55ceb593e6b0 -> cliamp:
        mapped: 1
        hidden: 0
        visible: 1
        acceptsInput: 1
        at: 2971,29
        size: 1227,1048
        workspace: 6 (6)
        floating: 0
        monitor: 1
 >      class: org.omarchy.cliamp
 >      title: cliamp
 >      initialClass: org.omarchy.cliamp
 >      initialTitle: Ghostty
        pid: 588281
        xwayland: 0
        pinned: 0
        pinFullscreened: 0
        fullscreen: 0
        fullscreenClient: 0
        fullscreenHandler: scrolling
        allowedOverFullscreen: 1
        grouped: 0
        tags: default-opacity*, terminal*
        swallowing: 0
        focusHistoryID: 1
        inhibitingIdle: 0
        xdgTag:
        xdgDescription:
        contentType: none
        tearingHint: 0
        stableID: 18000039
```

tahadx:
Fixes #7429

### The problem

Launching `cliamp` from the application menu produces a terminal window with `class: com.mitchellh.ghostty` — indistinguishable from any other Ghostty terminal. Window rules cannot match it specifically.

Launching via the keybinding (`SUPER+SHIFT+ALT+M`) works correctly: it goes through `omarchy-launch-or-focus-tui`, which sets `--app-id=org.omarchy.cliamp`, giving the window `class: org.omarchy.cliamp`.

### Root cause

The package's `cliamp.desktop` ships with `Terminal=true` + `Exec=cliamp`. When the app menu launches it, the desktop environment opens it in the default terminal (Ghostty) with Ghostty's own class — not a cliamp-specific one.

### The fix

Ship `applications/cliamp.desktop` as an override (installed by `omarchy-refresh-applications` to `~/.local/share/applications/`, taking precedence over the package's `/usr/share/applications/cliamp.desktop`):

```ini
Exec=omarchy-launch-tui --app-id=org.omarchy.cliamp cliamp
Terminal=false
```

This routes through the same `omarchy-launch-tui` wrapper the keybinding uses, so both paths produce `class: org.omarchy.cliamp` and window rules can match cliamp regardless of how it was launched.

### Verification

- Added `test/shell.d/cliamp-desktop-entry-test.sh` verifying the override exists, uses `omarchy-launch-tui` with the correct app-id, sets `Terminal=false`, and the keybinding uses the same app-id.

### DIFF
--- a/applications/cliamp.desktop
+++ b/applications/cliamp.desktop
@@ -1,0 +1,12 @@
+[Desktop Entry]
+Version=1.0
+Name=cliamp
+GenericName=Music Player
+Comment=A retro terminal music player inspired by Winamp 2.x
+Exec=omarchy-launch-tui --app-id=org.omarchy.cliamp cliamp
+Icon=cliamp
+Terminal=false
+Type=Application
+Categories=Audio;Music;Player;AudioVideo;
+Keywords=music;audio;player;terminal;tui;winamp;radio;podcast;
+StartupNotify=false
--- a/test/shell.d/cliamp-desktop-entry-test.sh
+++ b/test/shell.d/cliamp-desktop-entry-test.sh
@@ -1,0 +1,31 @@
+#!/bin/bash
+
+set -euo pipefail
+
+source "$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd)/base-test.sh"
+
+desktop_file="$ROOT/applications/cliamp.desktop"
+
+[[ -f $desktop_file ]] || fail "Omarchy ships a cliamp desktop entry override"
+pass "Omarchy ships a cliamp desktop entry override"
+
+# The package's cliamp.desktop ships Terminal=true + Exec=cliamp, which makes
+# the app menu launch it in a generic terminal with the terminal's own class
+# (e.g. com.mitchellh.ghostty) — indistinguishable from any other terminal.
+# The override must route through omarchy-launch-tui with a distinct app-id
+# matching the keybinding's, and set Terminal=false since the launcher wraps
+# the command in a terminal itself.
+grep -Fq 'Exec=omarchy-launch-tui --app-id=org.omarchy.cliamp cliamp' "$desktop_file" ||
+  fail "cliamp desktop override launches through omarchy-launch-tui with the org.omarchy.cliamp app-id"
+pass "cliamp desktop override uses omarchy-launch-tui with a distinct app-id"
+
+grep -Fq 'Terminal=false' "$desktop_file" ||
+  fail "cliamp desktop override sets Terminal=false (the launcher opens the terminal)"
+pass "cliamp desktop override sets Terminal=false"
+
+# The keybinding uses the same app-id, so both paths produce the same window
+# class and window rules can match cliamp regardless of how it was launched.
+keybinding_file="$ROOT/default/hypr/bindings/applications.lua"
+grep -Fq 'cliamp' "$keybinding_file" ||
+  fail "cliamp keybinding exists for cross-reference"
+pass "cliamp keybinding uses the same app-id as the desktop override"


tahadx (commented):

## What I checked

The override mirrors the keybinding's `omarchy-launch-tui --app-id=org.omarchy.cliamp` invocation, so both entry points converge on the same window class — that convergence is the whole game for window-rule matching, and it means the fix cannot regress for one path without the test noticing the other. `Terminal=false` is correct since the launcher wraps the TUI in a terminal itself; leaving `Terminal=true` would nest terminals.

## Observations

- **Mechanism is used as designed:** shipped desktop files land in `~/.local/share/applications/` via `omarchy-refresh-applications`, which takes precedence over `/usr/share/applications/` — so this fixes upstream packaging behavior without patching the package.
- **Tightening opportunity (minor):** the cross-reference test greps the keybinding file for the string `cliamp` but does not assert that both files carry an identical `--app-id=` value. Comparing the extracted app-id from each side would catch a future divergence where menu-launched and keybinding-launched windows silently drift apart — which is precisely the failure mode this PR repairs.
- **Metadata looks right:** `StartupNotify=false` suits the wrapper indirection, and the Categories/Keywords set is consistent with a terminal music player.

Small, focused, and tested at the exact seam where the bug lived. Nothing further from me.

tahadx (commented):

## Rich code review — #7450 (cliamp desktop override with distinct app-id)

Ships `applications/cliamp.desktop` that routes through `omarchy-launch-tui --app-id=org.omarchy.cliamp` with `Terminal=false` (since the launcher wraps the terminal), so cliamp gets a stable, distinct window class for rule matching regardless of launch path.

### Verified
- `omarchy-refresh-applications` copies every `applications/*.desktop` into `~/.local/share/applications/`, so the override is actually installed — I confirmed the mechanism.
- `omarchy-launch-tui --app-id=…` is a valid invocation pattern in this codebase.
- `Terminal=false` + keybinding cross-reference test pins the behavior.

### Nit (non-blocking)
The `Icon=cliamp` assumes an icon named `cliamp` is resolvable; if the AUR package doesn't ship one, the desktop entry may render with the fallback generic icon. Cosmetic only.

Verdict: ship.

PR #7451, opened 2026-08-19
menu: scale wheel events 3x for faster touchpad scrolling on long lists

### LINKED ISSUE #7361: Omarchy menu scroll speed is too slow with touchpad
## Description

The Omarchy menu (opened via SUPER+K or SUPER+SPACE) scrolls very slowly when using the touchpad. Navigating long lists like the keybindings menu requires many scroll gestures to move through the list.

The menu's `ListView` in `shell/plugins/menu/Menu.qml` uses Qt's default Flickable scroll behavior, which feels sluggish compared to the rest of the desktop.

## Steps to reproduce

1. Open the Omarchy menu with SUPER+K or SUPER+SPACE
2. Navigate to a submenu with many items (e.g. keybindings)
3. Scroll with the touchpad — movement is very slow per gesture

## Expected behavior

Scrolling should feel responsive and consistent with other scrollable UIs on the system.

## Notes

- Omarchy version: 4.0.0-1
- Kernel: 7.1.8-arch1-3
- The global `touchpad.scroll_factor` in Hyprland input config affects all Wayland surfaces equally, so increasing it fixes the menu but makes touchpad scrolling too fast in normal apps (browsers, etc.)
- A menu-specific scroll speed setting (or a higher default for the Quickshell overlay) would be helpful

tahadx:
Fixes #7361

### The problem

The Omarchy menu scrolls very slowly with touchpad on long lists like the keybindings menu. Navigating requires many scroll gestures to move through the list.

### Root cause

The menu's `ListView` uses Qt's default `Flickable` scroll behavior, which translates the small per-event `angleDelta` from touchpad scroll events into tiny `contentY` steps. On high-precision touchpads each event carries only 10-20 units of angleDelta, so the list barely moves per gesture.

### The fix

Added a `WheelHandler` to the menu's `ListView` that scales the wheel delta by 3x and adjusts `contentY` directly, clamped to the list's bounds:

```qml
WheelHandler {
    onWheel: function(event) {
        var step = event.angleDelta.y * 3
        var minY = resultList.originY
        var maxY = resultList.originY + resultList.contentHeight - resultList.height
        if (maxY <= minY) return
        resultList.contentY = Math.max(minY, Math.min(resultList.contentY - step, maxY))
    }
}
```

This makes both touchpad and mouse wheel scrolling 3x faster without being uncontrollable.

### Verification

- Added `test/shell.d/menu-scroll-speed-test.sh` (source-level assertions over `Menu.qml`).
- **Recommended visual verification**: open the keybindings menu (SUPER+K), scroll with touchpad, and confirm the list moves at a responsive speed.

### DIFF
--- a/shell/plugins/menu/Menu.qml
+++ b/shell/plugins/menu/Menu.qml
@@ -1224,6 +1224,19 @@ Item {
             spacing: root.rowSpacing
             boundsBehavior: Flickable.StopAtBounds
 
+            // Qt's default Flickable touchpad scrolling is too slow on long
+            // lists (e.g. keybindings). Scale the wheel delta so each gesture
+            // moves the list meaningfully instead of a few pixels at a time.
+            WheelHandler {
+              onWheel: function(event) {
+                var step = event.angleDelta.y * 3
+                var minY = resultList.originY
+                var maxY = resultList.originY + resultList.contentHeight - resultList.height
+                if (maxY <= minY) return
+                resultList.contentY = Math.max(minY, Math.min(resultList.contentY - step, maxY))
+              }
+            }
+
             section.property: "section"
             section.criteria: ViewSection.FullString
             section.delegate: Item {
--- a/test/shell.d/menu-scroll-speed-test.sh
+++ b/test/shell.d/menu-scroll-speed-test.sh
@@ -1,0 +1,22 @@
+#!/bin/bash
+
+set -euo pipefail
+
+source "$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd)/base-test.sh"
+
+run_node_test <<'JS'
+const fs = require('fs')
+const menuSource = fs.readFileSync(root + '/shell/plugins/menu/Menu.qml', 'utf8')
+
+// The menu ListView must carry a WheelHandler that scales the wheel delta so
+// touchpad scrolling on long lists (e.g. keybindings) feels responsive. Qt's
+// default Flickable scroll on touchpads translates the small per-event
+// angleDelta into tiny contentY steps, making the menu feel sluggish.
+const listMatch = menuSource.match(/ListView \{[\s\S]*?id: resultList[\s\S]*?\n            \}/)
+assert(listMatch, 'menu ListView with id resultList exists')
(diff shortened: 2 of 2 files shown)


tahadx (commented):

## What I checked

The WheelHandler clamps `contentY` to `[originY, originY + contentHeight - height]` and bails out when the list fits without scrolling, so there is no overscroll bounce or negative-range weirdness. Setting `contentY` directly bypasses Flickable's momentum/fling behavior — a reasonable trade of smoothness for determinism in a utility menu where reaching the target row matters more than glide.

## Observations

- **Mouse wheels get scaled too:** the hardcoded `* 3` applies to mouse-wheel events as well as high-frequency touchpad ones. A typical mouse notch (~120 angleDelta units → ~360 px) becomes a substantial jump in long lists. If mice end up feeling too fast in practice, distinguishing `pixelDelta` vs `angleDelta` or scaling only touchpad-class events would be the refinement path — the PR title acknowledges both inputs change, so this may be fully intended.
- **Test brittleness:** the regex anchors on the ListView's closing indentation (`\n            }`), so any reformat of `Menu.qml` breaks the extraction rather than the assertion. Consistent with the suite's other source-level tests; just flagging it.
- **Visual verification is genuinely required here** — scroll feel is exactly the property source assertions cannot measure, and the PR correctly calls for it rather than claiming test coverage proves the UX.

Small, targeted, well-commented. Nothing blocking from me.

tahadx (commented):

## Rich code review — #7451 (menu scroll 3x)

Adding a WheelHandler that scales `event.angleDelta.y` by 3 and writes `contentY` directly makes long-list touchpad scrolling feel responsive. The clamping to `[originY, originY+contentHeight-height]` is correct.

### Verified
- `maxY <= minY` early-return avoids div/NaN on short lists.
- `WheelHandler` accepting the event keeps the Flickable from double-scrolling.
- Test asserts structure (WheelHandler + 3x + contentY + originY clamp).

### Caveat worth knowing (non-blocking, UX)
Taking over wheel handling disables Qt's native kinetic/momentum scrolling on this list — the list now moves in discrete 3x steps rather than gliding. For a keybindings list that's usually fine, but if it ever feels abrupt, consider only scaling when the native step is small, or giving a very large delta its own fast path.

Verdict: ship.

PR #7452, opened 2026-08-19
Clipboard manager: OCR search, configurable history and paste behavior

sspaeti:
This improves the Clipboard manager by adding OCR, a configurable history item, and security hardening. It's basically my initial plugin https://github.com/sspaeti/omarchy-clipboard-plugin pushed upstream with the same configuration as of now with Omarchy, while direct pasting and OCR can be toggled:

```
{
  "id": "omarchy.clipboard",
  "historyLimit": 500,
  "maxRows": 50,
  "ocr": true,
  "ocrLang": "eng+deu",
  "autoPaste": false
}
```

- `historyLimit` (default 300) and `maxRows` (default 50) — current fixed values become configurable.
- `ocr` (default off) — runs tesseract on captured images and stores the text on the entry, so the picker's search finds text *inside* screenshots. tesseract is an optional dependency: not installed → OCR silently skipped, nothing else needed. `Ctrl+I` in the picker toggles an images-only view.
- `autoPaste` (default true, current behavior) — set to `false` and Enter/click only copies to the clipboard, so you paste manually and never paste into the wrong window by accident (Shift+Enter still auto-pastes). Personally I run with `false`, but the PR doesn't change the default.

If this OCR is not wanted to add for clipboard, we can also just have it as plugin here: https://github.com/HANCORE-linux/omarchy-plugin-marketplace/issues/679.

The first commit (44b24af5) is a standalone security hardening fix (see https://github.com/HANCORE-linux/omarchy-plugin-marketplace/issues/679#issuecomment-5336028249) (plain-text rendering of clipboard content + cleanup of orphaned image files) and is also available as its own branch, clipboard-security-fixes, if you'd prefer to take it separately. If you'd like only the security fix, this would be a dedicated branch: https://github.com/sspaeti/omarchy/commits/clipboard-security-fixes/

### DIFF
--- a/shell/plugins/clipboard/Clipboard.qml
+++ b/shell/plugins/clipboard/Clipboard.qml
@@ -19,6 +19,7 @@ Item {
 
   property string historyPath: Quickshell.env("HOME") + "/.local/state/omarchy/clipboard-history.json"
   property string captureScript: root.omarchyPath + "/shell/plugins/clipboard/capture.sh"
+  property string gcScript: root.omarchyPath + "/shell/plugins/clipboard/gc.sh"
   // Shares the [menu] surface tokens — themes that style the menu also
   // style the clipboard. Selected-row colors composed in the
   // singleton so consumers drop them straight into Rectangle bindings.
@@ -74,6 +75,10 @@ Item {
 
   function saveHistory() {
     historyFile.setText(JSON.stringify(root.history.slice(0, root.historyLimit), null, 2) + "\n")
+    // Deleted, evicted, and cleared image entries would otherwise leave
+    // their captured files on disk forever. Delayed so gc.sh reads the
+    // freshly written history file.
+    gcTimer.restart()
   }
 
   function addClipboardEntry(entry) {
@@ -239,6 +244,13 @@ Item {
   }
 
   Component.onCompleted: initProc.running = true
+
+  Timer {
+    id: gcTimer
+    interval: 1000
+    repeat: false
+    onTriggered: Quickshell.execDetached([root.gcScript])
+  }
 
   ListModel { id: displayModel }
 
@@ -436,6 +448,7 @@ Item {
             anchors.right: parent.right
             anchors.verticalCenter: parent.verticalCenter
             text: root.filterText || "Search clipboard…"
+            textFormat: Text.PlainText
             color: root.foreground
             opacity: root.filterText ? 1 : 0.58
             font.family: root.fontFamily
@@ -503,6 +516,9 @@ Item {
                       width: parent.width - (parent.parent.previewImage.length > 0 ? parent.height + parent.spacing : 0)
                       height: parent.height
                       text: parent.parent.previewText
+                      // Clipboard content is untrusted: AutoText would parse
+                      // copied HTML markup and load remote images from it.
+                      textFormat: Text.PlainText
                       color: parent.parent.hasCursor ? root.selectedText : root.foreground
                       font.family: root.fontFamily
                       font.pixelSize: Style.font.title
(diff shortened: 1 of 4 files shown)


sspaeti:
Updated the PR with inputs from Copilot:

- **1. (tesseract `-` / timeout guard):** *"Verified on tesseract 5.5.3: `-` and `stdout` produce identical output and no `-.txt` is created — `-` has been an alias for `stdout` since 3.05. `timeout` is GNU coreutils, a hard dependency of Arch, which is the only platform Omarchy runs on, so the fallback wrapper would be dead code. Not changing this one."*
- **2:** addressed in `2806931e` — referenced set built once, O(1) membership checks.
- **3:** addressed in `eb038b9e` — capped at 4000 chars in `normalizeEntry`, matching `OCR_MAX_CHARS`.

shell/plugins/clipboard/gc.sh
```diff
@@ -0,0 +1,28 @@
...
+HISTORY="$STATE_DIR/clipboard-history.json"
+
+[[ -d $IMAGE_DIR ]] || exit 0
+
+declare -A referenced=()
+if [[ -f $HISTORY ]]; then
+  while IFS= read -r path; do
+    [[ -n $path ]] && referenced["$path"]=1
+  done < <(jq -r '.[]? | select(.type == "image") | .path // empty' "$HISTORY" 2>/dev/null || true)
+fi
```
Ahmed-Sinkeat, line 20: Read the history the overlay actually writes, and skip the run if it can't be read:

```suggestion
HISTORY="${1:?usage: gc.sh <history-path>}"

[[ -d $IMAGE_DIR ]] || exit 0

# A history that can't be read references nothing we can trust: skip this run.
paths=$(jq -r '.[]? | select(.type == "image") | .path // empty' "$HISTORY" 2>/dev/null) || exit 0

declare -A referenced=()
while IFS= read -r path; do
  [[ -n $path ]] && referenced["$path"]=1
done <<<"$paths"
```

shell/plugins/clipboard/Clipboard.qml
```diff
@@ -240,6 +285,13 @@ Item {
 
   Component.onCompleted: initProc.running = true
 
+  Timer {
+    id: gcTimer
+    interval: 1000
+    repeat: false
+    onTriggered: Quickshell.execDetached([root.gcScript])
```
Ahmed-Sinkeat, line 292: Pass the overlay's history path so `gc.sh` never rebuilds it from `XDG_STATE_HOME`:

```suggestion
    onTriggered: Quickshell.execDetached([root.gcScript, root.historyPath])
```

Ahmed-Sinkeat (commented):
Thanks for adding `gc.sh`. Orphaned clipboard images do pile up: after clearing history I had 147 of 193 image files left on disk with nothing referencing them.

One case where it deletes images that history still uses. `gc.sh` reads history from `${XDG_STATE_HOME:-$HOME/.local/state}/omarchy/clipboard-history.json`, but `Clipboard.qml` (unchanged in this PR) writes it to a hardcoded `$HOME/.local/state/omarchy/clipboard-history.json`, while `capture.sh` stores images under `$XDG_STATE_HOME`. With `XDG_STATE_HOME` set anywhere other than `~/.local/state`, `gc.sh` finds no history, `referenced` stays empty, and every image older than 10 minutes is deleted, including ones history still points at.

Reproduced with this PR's `gc.sh` at eb038b9e in a scratch directory: with `XDG_STATE_HOME` unset a referenced image is kept; with `XDG_STATE_HOME` set elsewhere the same image is deleted.

It's the same failure as Copilot's note on `gc.sh:28` (a history that can't be read deletes everything), which I think is still open. The two suggestions below cover both: pass the overlay's own history path in, and skip the run when it can't be read. With them, in the same scratch setup, the referenced image survives the `XDG_STATE_HOME` case, a truncated or missing history deletes nothing, and a cleared history still removes old images while leaving fresh ones alone.

PR #7455, opened 2026-08-19
Read Fireworks credentials from pi's auth.json

TyRichards:
## Problem

The Fireworks collector in `omarchy.agents` never sees a Fireworks login made through pi — Omarchy's default agent. A machine signed in only via `/login fireworks` in pi gets a permanently hidden Fireworks tab, even though the same machine's Claude and Codex collectors already treat pi as a first-class session source.

## Change

Insert pi's credential store into the collector's fallback ladder, between the `firectl` `auth.ini` and the opencode fallback:

1. `FIREWORKS_API_KEY` / `FIREWORKS_ACCOUNT_ID` (unchanged)
2. `~/.fireworks/auth.ini` from `firectl set-api-key` (unchanged)
3. **`fireworks.key` in `$PI_CODING_AGENT_DIR/auth.json`** (default `~/.pi/agent/auth.json`) — new
4. opencode's `~/.local/share/opencode/auth.json` (unchanged, still last)

pi's documented key forms are honored: literals and `$ENV_VAR`/`${ENV_VAR}` references resolve (a missing variable leaves the key unresolved and falls through), while `!command` shell lookups stay pi-only and are skipped rather than sent to the Fireworks API verbatim. `PI_CODING_AGENT_DIR` is respected for non-default pi setups.

The auth help text and plugin README now mention pi alongside the other sources.

## Tests

Four new assertions in `test/shell.d/agent-usage-fireworks-scanner-test.sh`:

- pi login wins over opencode but loses to firectl
- `$ENV_VAR` keys resolve; missing env vars stay unresolved
- `!command` keys are skipped and fall through to opencode

The pi dir is pinned via `PI_CODING_AGENT_DIR` in both the shell-level and inline-python fixtures, mirroring the existing `XDG_DATA_HOME` pinning that keeps real developer credentials out of fixture runs.

`bash test/shell.d/agent-usage-fireworks-scanner-test.sh` — all 19 pass. `./test/all` — the only failures are 4 pre-existing/environmental ones that also fail on a clean checkout (`omarchy-pkgs` sibling checkout missing; bar icon geometry on this machine).

### DIFF
--- a/bin/omarchy-agent-usage-fireworks
+++ b/bin/omarchy-agent-usage-fireworks
@@ -29,7 +29,7 @@ from typing import Any
 
 AGENT_ID = "fireworks"
 AGENT_NAME = "Fireworks"
-AUTH_HELP = "Set FIREWORKS_API_KEY, run `firectl set-api-key`, or sign in to Fireworks in opencode."
+AUTH_HELP = "Set FIREWORKS_API_KEY, run `firectl set-api-key`, or sign in to Fireworks in pi or opencode."
 API_BASE_URL = "https://api.fireworks.ai"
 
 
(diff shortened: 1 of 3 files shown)


TyRichards:
Subtle thing, but there are a lot of pi users out there.

PR #7457, opened 2026-08-19
Fix Zen policy installation

Awlaursen:
Zen installs under `/opt/zen-browser-bin`, but Omarchy wrote its policies to `/opt/zen-browser`, where Zen never reads them. This uses Zen's actual distribution directory and merges Omarchy's defaults with the policies from `zen-browser-bin` while preserving package-owned values.

Verified with `zen-browser-bin 1.21.14b-1`: the packaged policies and Omarchy preferences were active in `about:policies`. Merge precedence is covered by the shell test, and all 181 test files pass.

This is intentionally limited to installation; a later package upgrade may replace the merged policy file.

Addresses #7411

### DIFF
--- a/bin/omarchy-install-browser
+++ b/bin/omarchy-install-browser
@@ -30,6 +30,23 @@ setup_firefox_preferences() {
   sudo cp -f "$OMARCHY_PATH/default/firefox/policies.json" "$distribution_dir/policies.json"
 }
 
+setup_zen_preferences() (
+  local distribution_dir="${1:-/opt/zen-browser-bin/distribution}"
+  local zen_policies="$distribution_dir/policies.json"
+  local merged_policies
+
+  if [[ ! -f $zen_policies ]]; then
+    setup_firefox_preferences "$distribution_dir"
+    return
+  fi
+
+  merged_policies=$(mktemp)
+  trap 'rm -f "$merged_policies"' EXIT
+
+  jq -s '.[0] * .[1]' "$OMARCHY_PATH/default/firefox/policies.json" "$zen_policies" >"$merged_policies"
+  sudo install -m644 "$merged_policies" "$zen_policies"
+)
+
 setup_firefox_wayland() {
   mkdir -p ~/.config/environment.d
   echo "MOZ_ENABLE_WAYLAND=1" > ~/.config/environment.d/omarchy-firefox-wayland.conf
@@ -93,7 +110,7 @@ zen)
   echo "Installing Zen..."
   omarchy-pkg-aur-add zen-browser-bin || exit 1
 
-  setup_firefox_preferences /opt/zen-browser/distribution
+  setup_zen_preferences
   setup_firefox_wayland
   announce_browser_installed "Zen"
   ;;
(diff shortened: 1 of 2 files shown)


Captnjo:
Verified on omarchy 4.0.2-1 with `zen-browser-bin` 1.21.16b-1: on this build the
distribution directory is not the path Zen consults first, so a fix that targets
`/opt/zen-browser-bin/distribution` alone still loses on the next package upgrade.

**Zen checks `/etc/zen/policies/policies.json` before the install dir.**

`browser/modules/EnterprisePoliciesParent.sys.mjs`, `_getLocalConfigurationFile()`
(extracted from this machine's `omni.ja`):

```js
if (AppConstants.platform == "linux" && AppConstants.MOZ_SYSTEM_POLICIES) {
  let systemConfigFile = Services.dirsvc.get("SysConfD", Ci.nsIFile);
  systemConfigFile.append("policies");
  systemConfigFile.append(POLICIES_FILENAME);
  if (systemConfigFile.exists()) {
    return systemConfigFile;      // returns here — XREAppDist is never reached
  }
}
```

Both build flags are set in this package's `AppConstants.sys.mjs`:

- `MOZ_SYSTEM_POLICIES: true`
- `MOZ_APP_NAME: "zen"` → `SysConfD` is `/etc/zen`

So the effective path is **`/etc/zen/policies/policies.json`**, and it *supersedes*
the distribution file rather than merging with it.

**Why that matters for this PR specifically.** The description notes that "a later
package upgrade may replace the merged policy file" — that is not a maybe, it is
guaranteed:

```console
$ pacman -Qii zen-browser-bin | grep -i backup
Backup Files    : None
```

`distribution/policies.json` is package-owned and *not* in the backup array, so
pacman overwrites it on every Zen upgrade with no `.pacnew` and no prompt. Any
merge written there is silently undone. `/etc/zen/policies/policies.json` is owned
by no package, so it survives upgrades — which removes the caveat entirely.

**Two things a `/etc`-based fix has to get right:**

1. **It replaces, it does not merge — and more totally than `cp -f` does.** The
   `DisableAppUpdate` trap already raised on #7411 applies here in a stronger
   form: a recursive merge into `distribution/policies.json` at least starts from
   the package's file, whereas an `/etc` file makes that file unreachable
   entirely. Whatever is written to `/etc` has to carry the package's policies
   forward on its own, and re-read them when the package changes them.
2. **It is one-way in practice.** On any machine that has grown the `/etc` file, a
   later revert to the distribution-dir approach would be silently ignored. Worth
   deciding deliberately rather than discovering later.

For completeness, the current `/opt/zen-browser` path in
`install/helpers/browser-policy.sh:29` is not merely unread — it is unowned:

```console
$ pacman -Qo /opt/zen-browser/distribution/policies.json
error: No package owns /opt/zen-browser/distribution/policies.json
$ pacman -Qo /opt/zen-browser-bin/distribution/policies.json
/opt/zen-browser-bin/distribution/policies.json is owned by zen-browser-bin 1.21.16b-1
```

`migrations/1787515927.sh:19` iterates `BROWSER_POLICY_FIREFOX_DIRS`, so the
migration recreates and populates that unread directory on existing installs.

I have not opened a competing PR; this is offered as input to whichever of #7411 /
#7444 / #7457 lands.

bjarneo:
Automated duplication check: this pull request looks similar to #7444, which covers the same Zen browser policy path. I keep that one open and close this one to consolidate review.

PR #7469, opened 2026-08-19
Add Hermes as a desktop app and a coding agent

omarchybot:
Adds [Hermes](https://hermes-agent.nousresearch.com/) in the places it belongs: the desktop app under Install > AI and Remove > AI, and the terminal agent under Setup > Defaults > Agent.

Pairs with omacom-io/omarchy-pkgs#166, which adds the `hermes-desktop` package.

## One Hermes per machine

The two cannot share an installation the obvious way round. Hermes Desktop only runs against a runtime built from its own commit — the version gap against any published CLI fails its readiness probe with a 401 — so it provisions its own under `~/.hermes` and puts that on `PATH`. Left alone that gives a machine two Hermeses on different Python versions, taking turns overwriting `~/.local/bin/hermes` and sharing one `HERMES_HOME` between them.

So the package decides ownership. Where `hermes-desktop` is installed it owns Hermes outright: `omarchy-install-hermes-cli` writes no stub, removes its own copy if it made one earlier, and reports through `--check` whether the app has finished setting Hermes up. Installing the app supersedes a CLI that was there first; installing the CLI second finds the app already owns it. Either order converges on the app's installation, so the terminal, the default agent and the app are the same Hermes.

Ownership keys off the package rather than `~/.hermes`, because the package is there before the app has ever run — exactly when a second copy would otherwise get built.

## Removal

Dropping the package alone would strand the runtime the app installed, which is the larger part of it — past 2 GB of checkout and venv — along with its command on `PATH`. `omarchy-remove-ai-hermes` takes both, and keeps everything in `~/.hermes` that is the user's: chats, memories, and the skills Hermes wrote for itself.

Only when the app actually installed them, though. The official Hermes installer uses the same paths, so the runtime is only the app's once the app has finished putting it there — `.hermes-bootstrap-complete` is what says so, and without it removing the package is the whole job. Ownership of `~/.local/bin/hermes` is decided the same way throughout: by the marker line the installer writes, which is why Remove Preinstalls, the migration and the installer all ask `omarchy-install-hermes-cli --owns` rather than matching it themselves.

## The terminal agent

Without the desktop app, Hermes is a mise-backed stub like the other agents, and it needs an installer of its own rather than a line in `omarchy-mise-install`.

Hermes pins every one of its dependencies with `==` and declares `Requires-Python >=3.11,<3.14`, so it can neither be built against Arch's Python nor share the `python-*` packages. Given no compatible interpreter to hand, `uv` builds the venv against the system Python in violation of that bound, reports success, and leaves the breakage to surface later inside a dependency — so the interpreter is pinned, and re-checked on every run, because `mise up` reinstalls without the pin. The cooldown override is exported rather than set on the install line, so the version resolved to run is the one just installed. mise's pipx backend also shells out to `uv`, which a stock Omarchy does not have; it comes from mise, so neither route needs root.

`omarchy-agent` learns a `hermes` case. Hermes has no "start interactive, seeded with a prompt" mode, so a prompt goes through `--oneshot`, which answers and exits.

## Glyph

`U+E90A`, from [Font Awesome's staff-snake](https://fontawesome.com/icons/staff-snake) (CC BY 4.0) — the mark Hermes serves as its favicon, and titles its README with. Their app icon is a detailed portrait, which traces into an even grey smudge at menu size next to the flat silhouettes this font is made of.

Rebased onto `quattro`. Ori took `U+E909` in #7709 while this was open, so the Hermes mark moved to `U+E90A` in the font, the menu entries, the font README and the charset the menu test pins.


camamacheiro:
I prepared a narrower agent-only Hermes integration against the current `quattro` branch:

- Branch: https://github.com/camamacheiro/omarchy/tree/rook/hermes-integration
- Commit: https://github.com/camamacheiro/omarchy/commit/72f4978b62d6c1dab13af0c00262f3b24595699b
- Compare: https://github.com/basecamp/omarchy/compare/quattro...camamacheiro:omarchy:rook/hermes-integration

This deliberately keeps Hermes Desktop, the usage monitor and branded-font work outside the default-agent change. The integration uses Nous Research's supported installer rather than PyPI/pipx/AUR packaging, wires the correct `--yolo` and `--oneshot` launch modes, links Omarchy skills into `~/.hermes/skills`, adds an existing-user migration, and verifies installer failure/cleanup behavior.

Focused Hermes/default-agent/menu tests and the full CLI suite pass. The broader shell suite has only the same host-dependent failures reproduced on untouched Quattro.

I have not opened a competing PR. This branch is available for the maintainer branch to cherry-pick or adapt so the existing Hermes efforts can converge.

vandaimer:
This PR might help with Hermes HUD styling 

https://github.com/basecamp/omarchy/pull/7094

teknium1 (approved):
Reviewed from the Hermes side — we verified every load-bearing claim against the hermes-agent codebase and they all check out:

- `Requires-Python >=3.11,<3.14` with fully `==`-pinned deps is correct, so the `UV_PYTHON=3.13` pin (re-checked on every run, since `mise up` reinstalls without it) is the right defense.
- The desktop app's PATH probe really is `hermes --version` on a 15s budget (`DEFAULT_PROBE_TIMEOUT_MS = 15_000`), and a cold run measures ~10.5s — so `--now` pre-warming the install before handing Hermes to the GUI is exactly right, and the margin is as thin as the comment implies.
- `.hermes-bootstrap-complete` is written last by our bootstrap; using it plus the `~/.hermes`-pointing wrapper as the readiness test is the correct combination.
- `--yolo` and `--oneshot <prompt>` exist and compose as used in `omarchy-agent`; `--oneshot` answers and exits, so the seeding comment is accurate.
- Removal scope is right: `~/.hermes/hermes-agent`, `bootstrap-cache`, `bin`, `node` belong to the app; chats/memories/skills under `~/.hermes` and connections under `~/.config/Hermes` are the user's. The `hermes`/`hermes-agent`/`hermes-acp` wrapper sweep matches what our install actually writes.

Keying ownership off the package rather than `~/.hermes` is the right call for exactly the reason stated — the package exists before the app has ever run.

One operational note, not a blocker: we ship several times a week, so the stub's per-run interpreter check and `MISE_MINIMUM_RELEASE_AGE=0` will both earn their keep. First run after an `omarchy update` may eat the minute-long reinstall — acceptable, just expected.

Approve from Nous.

sangfrois:
I'll install omarchy when Hermes is on the stable release. Thanks for the work you guys

spencerbull:
Thanks for the work on this. I re-reviewed the Hermes installation flow end to end, including several independent Claude Fable review passes and installation on a disposable current Omarchy VM.

Teknium’s approval carries substantial weight here. The upstream-sensitive scripts remain byte-identical to the commit he reviewed, and the Hermes-specific design checked out:

- The Python 3.13 constraint is handled correctly.
- The mise/PyPI installation completed successfully.
- The cold install, readiness marker, 15-second desktop probe, and `--now` path behaved as intended.
- Hermes Agent 0.19.0 ran successfully on Python 3.13.15.
- The packaged desktop application installed and mapped a real Hermes window under Hyprland.
- All 162 focused Hermes CLI, removal, default-agent, and menu assertions passed.

The review identified one narrow Omarchy-side ownership issue around `~/.local/bin/hermes`.

`bin/omarchy-install-hermes-cli` can replace an existing wrapper created by Hermes’ official installer, while `bin/omarchy-remove-preinstalls` can delete that wrapper whenever Hermes Desktop is absent. Omarchy should only replace or remove a Hermes wrapper carrying its own `omarchy-install-hermes-cli` marker.

The review also confirmed that existing users need a migration equivalent to the recent Hey/Ori migrations. Adding Hermes to `install/user/mise.sh` covers fresh installations, but does not install the preconfigured wrapper during an update. That migration should respect both `preinstalls-removed` and externally owned Hermes wrappers.

I’m going to open a small follow-up PR against this branch that:

- Preserves externally owned Hermes wrappers.
- Makes `--check` and normal installation agree on external ownership.
- Limits “Remove Preinstalls” to Omarchy-created wrappers.
- Adds the existing-user migration.
- Adds focused regression coverage for these cases.

I did not find a problem with the Hermes packaging or installation strategy itself. The follow-up should be limited to tightening Omarchy’s ownership and upgrade behavior.

spencerbull:
The installation hardening from the review is ready in #8539.

It targets this PR’s `hermes-agent` branch directly, so merging it will fold the changes into #7469 rather than create a competing PR against `quattro`.

It preserves externally owned Hermes wrappers, adds the existing-user migration, and adds focused ownership regressions. The disposable Omabot validation passed 188 targeted assertions, including a real Hermes installation, and the final independent Fable review returned GO.

dhh:
> #8455 landed on `bin/omarchy-agent` establishing that a prompted agent keeps its window. `hermes --yolo --oneshot` answers and exits, and the terminal dies with it — measured on the compositor. Hermes has no interactive-seeded mode (`-z/--oneshot` and `chat -q` are both documented non-interactive, and `chat` takes no positional prompt), so this is a behaviour choice rather than a fix to push.

@teknium1 @spencerbull Could you guess verify what's needed here?

spencerbull:
The remaining failure is `--oneshot`: Hermes answers and exits, and `omarchy-launch-tui` then closes the terminal because its child ended. I reproduced that behavior on a credential-free disposable current Omarchy worker.

#8705 replaces only the prompted argv with `hermes chat --yolo --tui --query=<prompt>`. On the same worker, PyPI Hermes v0.19.0 and Hermes Desktop's owned v0.20.4 wrapper both auto-submitted the initial prompt, stayed mapped after the response, and accepted a second turn; a dash-prefixed prompt remained data. The focused launcher test and CLI suite pass, and reverting the launcher line to `--oneshot` makes the new regression assertion fail. The full shell suite's three failures reproduce identically on bare `quattro`.

GPT-5.6 in Codex performed the investigation and implementation. A separate Codex xhigh review of exact SHA `b609ae235578e648900f7fb9e8f26602c8fd557a` found no substantive findings; transcript-level independence is not guaranteed. This is waiting on the author or maintainer in #8705; nothing was merged.

🤖 Generated by GPT-5.6 in Codex. Reviewed by Codex XHigh.

spencerbull:
@dhh Verified. The missing piece is not a seeded-TUI flag: Hermes's TUI query path interprets control syntax, so it cannot safely carry an arbitrary launcher prompt as literal data.

#8705 is now updated to the final `64203cc2` implementation. It uses a two-stage handoff:

1. Seed with `hermes --yolo --usage-file "$usage" --oneshot="$prompt"`, which binds option-shaped prompts as data and avoids the TUI command/interpolation layer.
2. Require a completed, non-failed usage report with a nonempty exact `session_id`.
3. Resume only that session with `hermes chat --yolo --tui --resume "$session_id"`.

The helper also clears inherited `HERMES_SESSION_SOURCE` for both stages so Hermes persists `source=cli` and the actual launch cwd/workspace, retries the seed once after first-run provider setup, and cleans the temporary usage file on every exit path. The installer capability-gates on `--usage-file` rather than rejecting older compatible Hermes releases, preserves unsupported foreign installs instead of taking ownership, and leaves an unprompted launch exactly `hermes --yolo`.

I independently reviewed the cumulative `fdb3755c..64203cc2` diff and found no substantive current blocker. I then validated the exact SHA in a credential-free disposable Omabot/Proxmox worker against real Hermes Agent 0.19.0: all 57 focused assertions passed; a multiline prompt beginning with `--help` and containing shell/slash/brace control syntax plus a trailing backslash reached the provider unchanged; the handoff resumed the exact reported session; Hermes's DB recorded `source=cli` and the correct cwd/workspace; success and failure paths removed the usage file; a foreign Hermes install remained unchanged; and the unprompted argv remained exactly `--yolo`. The worker was destroyed after validation.

The remaining gaps are rendered-TUI automation and a successful real first-run setup replay in the credential-free worker; both control paths have focused mock coverage. So yes: prompted-session behavior is ready, and #7469 needs the stacked #8705 merged into `hermes-agent` before #7469 lands.

🤖 Generated by GPT-5.6 in Codex. Reviewed and runtime-validated by Codex XHigh.

spencerbull:
Follow-up after @teknium1's [NousResearch/hermes-agent#97121](https://github.com/NousResearch/hermes-agent/pull/97121): I reviewed its merged commit and current Hermes `main`. It fixes the primary upstream problem correctly — on a real TTY, `chat -q` now seeds a live session and bypasses slash-command, `!` shell, interpolation, and file-drop dispatch for the initial prompt. That is the right long-term interface for Omarchy.

It does not make #8705 obsolete yet:

- The feature is not released. The latest Hermes GitHub release predates #97121, PyPI still serves 0.19.0, and #7469's CLI installer currently installs the PyPI package. Existing Hermes Desktop runtimes also cannot be assumed to contain an unreleased `main` change.
- Current Hermes `main` still does [`HERMES_TUI_QUERY.trim()`](https://github.com/NousResearch/hermes-agent/blob/baa344dee76993f0444c18fc59a69738ccb339d0/ui-tui/src/config/env.ts#L28-L30), so leading and trailing prompt whitespace is lost before the new literal submission path sees it.
- The first-run guard still [runs setup and then returns](https://github.com/NousResearch/hermes-agent/blob/baa344dee76993f0444c18fc59a69738ccb339d0/hermes_cli/main.py#L3263-L3290), without replaying the original seeded query after successful setup.
- A direct replacement must still clear an inherited `HERMES_SESSION_SOURCE`; otherwise a local Omarchy launch can retain a gateway/source tag instead of persisting the local CLI source and launch workspace.

For those reasons, #8705 at `64203cc2` should remain as written for now. Its one-shot → validated usage report → exact-session resume path supports released Hermes 0.19.0 and current Desktop-owned installations, preserves the exact prompt, replays after setup, clears the inherited source for both stages, and has already passed the disposable Omabot/Proxmox validation. The top-level one-shot, `--usage-file`, and exact resume interfaces it uses remain present on current Hermes `main`, so Teknium's change does not break the compatibility path.

My recommendation is to merge #8705 into `hermes-agent` so #7469 is correct for installations available today. Once #97121 is released through both supported Hermes installation paths and upstream preserves raw startup whitespace plus setup replay, we can simplify Omarchy to the native seeded-TUI invocation and rerun the runtime validation.

🤖 Generated and verified by GPT-5.6 in Codex.

spencerbull:
NousResearch/hermes-agent#97121 is merged, so stacked #8705 now switches prompted launches to Hermes's native interactive query path; Codex XHigh independently reviewed that update, and its remaining upstream edge cases are documented on #8705. We will hold #7469 until a published Hermes Agent version containing [#97121 ](https://github.com/NousResearch/hermes-agent/pull/97121)is available through Omarchy's supported installer, then land #8705 before this PR.

— 🤖 Codex, posting on behalf of @spencerbull

smfworks:
#7469 never links Omarchy agent skills into Hermes.

#7392 adds the usual `ln -sfn` into `~/.hermes/skills` only (same loop as Claude/Codex/Pi). Hermes profiles are separate homes — existing `~/.hermes/profiles/*/skills` never get the link unless something walks them without creating profiles. That remaining loop is #8611 (mkdir + `ln -sfn`; no profile creation).

Happy for this PR to take it. If it does, we'll close #8611. If not, we'll keep #8611 rather than stacking another installer.

spencerbull (approved):

spencerbull (approved):

PR #7484, opened 2026-08-19
feat: Pause and resume a screen recording mid-capture

### LINKED ISSUE #7479: Pause and resume a screen recording mid-capture
## What's missing

Omarchy's screen recorder (`omarchy-capture-screenrecording`) can only do two things: start a fresh recording, or stop and finalize one. Once a recording is running, the only option is to end it — there's no way to pause the capture and pick the same file back up a moment later.

That's a real gap for the common case of recording a demo or a clip and wanting to drop the dead air (loading screens, switching contexts, a phone call) without splicing multiple files together afterward.

## Why this is possible today

`gpu-screen-recorder` already supports it. From its man page:

> SIGUSR1 / SIGUSR2 — Pause/unpause recording (not for streaming/replay).

Omarchy just doesn't surface it. The stop path sends SIGINT; nothing ever sends SIGUSR1.

## Proposed

1. **`omarchy-capture-screenrecording --pause-recording`** — toggles pause by sending `SIGUSR1` to the running `gpu-screen-recorder` process, and tracks paused state in a marker file (e.g. `/tmp/omarchy-screenrecord-paused`) so the UI can read it back.
2. **Menu entries** — a `Pause Recording` row visible while a recording is active, swapping for a `Resume Recording` row while paused (the `when` guard already handles "recording active" via `pgrep`; extend it with the marker file check).
3. **Bar indicator** — `shell/plugins/bar/indicators/ScreenRecording.qml` currently only reflects "recording" vs "idle". Reflect a third state, "paused" (distinct glyph + tooltip), and make the indicator click do the obvious thing per state: idle → open the capture menu, recording → stop, paused → resume.

Start, pause, resume, and stop all stay one command; pause/resume is a toggle on the same `--pause-recording` flag.

## Files touched

- `bin/omarchy-capture-screenrecording`
- `shell/plugins/bar/indicators/ScreenRecording.qml`
- `default/omarchy/omarchy-menu.jsonc`
- tests under `test/shell.d/`

tahadx:
Closes #7479.

## What

`gpu-screen-recorder` already supports pause, but Omarchy only ever stopped a recording (SIGINT). This adds pause/resume so a capture can be held and continued on the same file, plus a focused chooser so a mid-recording click no longer destroys the take on the first click.

### Behavior

- **Idle** — clicking the bar indicator opens the capture menu (start options), unchanged.
- **Recording** — the indicator shows the record icon. Clicking it opens a small menu with **only Pause and Stop** (not the full start menu).
  - **Pause** — the capture pauses (indicator switches to the pause icon).
  - **Stop** — stops and saves like before.
- **Paused** — clicking the pause icon **resumes** directly (one click to continue).
- The Capture menu (Super+Ctrl+C → Screenrecord) gains **Stop / Pause / Resume** rows that appear/disappear based on state (Pause hides when paused; Resume appears instead).

## How

- **`omarchy-capture-screenrecording --pause-recording`** — toggles pause by sending `SIGUSR2` to the running process and flips a `/tmp/omarchy-screenrecord-paused` marker file (the process won't report paused state back, so the marker is the source of truth). `start` and `stop` both clear the marker so a crashed session can't invert the next recording's first toggle.
- **`omarchy-capture-screenrecording --prompt`** — presents the Pause/Stop chooser via `omarchy-menu-select` (the picker offers exactly the actions valid for the current state).
- **Indicator (`ScreenRecording.qml`)** — polls every 2s (`pgrep -f ... >/dev/null` + the marker file) so it reflects state on its own, rather than relying solely on the script's IPC poke (which races the shell right after a restart). Uses `pgrep -f ... >/dev/null` rather than `pgrep -qf` because this build's `pgrep` has no short `-q` flag.
- **Menu** — `trigger.capture.screenrecord.pause` / `.resume` entries gated on `pgrep` active ± the marker file.

## Why SIGUSR2

`SIGUSR1` is "save replay" — a no-op for regular recording, so the marker flipped and the icon changed but the capture kept going. The gsr 6.0 man page lists `SIGUSR2` as the pause/unpause signal for regular recording.

## Tests

- `screenrecording-pause-test.sh` — pause creates the marker + sends SIGUSR2; resume clears it; only SIGUSR2 is ever sent (never SIGINT); pause with no active recording is a non-signaling no-op.
- `screenrecording-prompt-test.sh` — the chooser dispatches pause/resume/stop correctly, offers the right two options per state, cancel sends nothing, idle opens the start menu.
- `indicator-contract` — idle/recording/paused indicator clicks dispatch to the right command.
- `menu-test`, `menu-guards`, `bin-style`, and the CLI metadata suite all pass.

## Notes for review

- Verified live on a Hyprland desktop: indicator reflects recording/paused state, the Pause/Stop chooser appears on a mid-capture click, Pause switches to the pause icon and actually pauses the capture, and a paused click resumes.
- All changes are package-owned files (`/usr/bin/omarchy-capture-screenrecording`, `/usr/share/omarchy/shell/.../ScreenRecording.qml`, `/usr/share/omarchy/default/omarchy/omarchy-menu.jsonc`) — no new dependencies, no font change, no user config migration, so this ships cleanly in a package update.

### DIFF
--- a/bin/omarchy-capture-screenrecording
+++ b/bin/omarchy-capture-screenrecording
@@ -2,7 +2,7 @@
 
 # omarchy:summary=Start or stop screen recording
 # omarchy:group=capture
-# omarchy:args=[--fullscreen] [--with-desktop-audio] [--with-microphone-audio] [--with-webcam] [--webcam-device=<device>] [--webcam-size=<small|medium|large>] [--resolution=<size>] [--stop-recording]
+# omarchy:args=[--fullscreen] [--with-desktop-audio] [--with-microphone-audio] [--with-webcam] [--webcam-device=<device>] [--webcam-size=<small|medium|large>] [--resolution=<size>] [--stop-recording] [--pause-recording]
 # omarchy:examples=omarchy screenrecord | omarchy capture screenrecording --with-desktop-audio
 # omarchy:aliases=omarchy screenrecord
 #
@@ -34,7 +34,9 @@ WEBCAM_SIZE="medium"
 RESOLUTION=""
 FULLSCREEN="false"
 STOP_RECORDING="false"
+PAUSE_RECORDING="false"
 RECORDING_FILE="/tmp/omarchy-screenrecord-filename"
+PAUSED_FILE="/tmp/omarchy-screenrecord-paused"
 REGION_FILE="${XDG_RUNTIME_DIR:-/tmp}/omarchy-screenrecord-region"
 LOG_FILE=$([[ ${OMARCHY_SCREENRECORD_DEBUG:-false} == "true" ]] && echo "/tmp/omarchy-screenrecord.log" || echo "/dev/null")
 
@@ -48,6 +50,7 @@ for arg in "$@"; do
   --resolution=*) RESOLUTION="${arg#*=}" ;;
   --fullscreen) FULLSCREEN="true" ;;
   --stop-recording) STOP_RECORDING="true" ;;
+  --pause-recording) PAUSE_RECORDING="true" ;;
   esac
 done
 
@@ -147,6 +150,10 @@ start_screenrecording() {
   local capture_args=()
   local target
 
+  # A paused marker left by a recording that never reached stop (crash, logout)
+  # would invert the first pause toggle of this new session, so drop it.
+  rm -f "$PAUSED_FILE"
+
   # Opt-in path for HDR, external-GPU monitors, and window capture (all things
   # the portal backend supports and the kms backend doesn't). Default flow uses
   # slurp + the kms backend, which avoids the EGL DMA-BUF modifier import
@@ -239,7 +246,7 @@ stop_screenrecording() {
     ) &
   fi
 
-  rm -f "$RECORDING_FILE"
+  rm -f "$RECORDING_FILE" "$PAUSED_FILE"
 }
 
 toggle_screenrecording_indicator() {
@@ -248,6 +255,27 @@ toggle_screenrecording_indicator() {
 
 screenrecording_active() {
   pgrep -f "^gpu-screen-recorder" >/dev/null
+}
+
+screenrecording_paused() {
+  [[ -f $PAUSED_FILE ]]
+}
+
+# SIGUSR1 toggles capture pause in gpu-screen-recorder. The process won't report
+# its paused state back, so the marker file is the source of truth — every
+# pause/resume goes through here, which keeps it in lockstep with the process.
+pause_screenrecording() {
+  pkill -SIGUSR1 -f "^gpu-screen-recorder"
+
+  if screenrecording_paused; then
+    rm -f "$PAUSED_FILE"
+    omarchy-notification-send -t 1500 "Recording resumed"
+  else
+    touch "$PAUSED_FILE"
+    omarchy-notification-send -t 1500 "Recording paused"
+  fi
+
+  toggle_screenrecording_indicator
 }
 
 finalize_recording() {
@@ -279,7 +307,13 @@ finalize_recording() {
   fi
 }
 
-if screenrecording_active; then
+if [[ $PAUSE_RECORDING == "true" ]]; then
+  if screenrecording_active; then
+    pause_screenrecording
+  else
+    exit 1
+  fi
+elif screenrecording_active; then
   stop_screenrecording
 elif [[ $STOP_RECORDING == "true" ]]; then
   exit 1
--- a/bin/omarchy-capture-screenrecording
+++ b/bin/omarchy-capture-screenrecording
@@ -2,7 +2,7 @@
 
 # omarchy:summary=Start or stop screen recording
 # omarchy:group=capture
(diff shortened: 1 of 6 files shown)


tahadx:
Closing for now — will revisit. Tracking issue #7479 stays open.

tahadx:
Will be back for this feature

tahadx (commented):

## What I checked

`SIGUSR2` is the correct choice and the comment earns its place: `SIGUSR1` saves a replay instead of pausing, which would have made this feature fail in the worst way — icon flips, capture keeps going. The marker-file-as-source-of-truth design is honest about gsr not reporting pause state back, and clearing the marker on both `start` and `stop` closes the crashed-session inversion case where a stale marker would flip the next recording's first toggle. Menu guards are bash conditions per `docs/menu.md`, so the compound `pgrep && [[ -f ... ]]` guards are valid there.

## Observations

- **Theoretical toggle race:** `pause_screenrecording` signals unconditionally but flips state based on the marker; two rapid invocations in flight could desync icon vs. actual capture state. Single-user bar interaction makes this near-impossible in practice — noting for awareness, not requesting a change.
- **Guard expression duplication:** the `pgrep -f '^gpu-screen-recorder'` fragment now lives in three menu rows plus the script. Guard batching keeps it cheap; a shared fragment would just keep them from drifting.
- **Idle polling cost:** the 2-second Timer runs for the session lifetime even while idle (one pgrep fork per tick). Cheap, but gating the timer on recording state would trim idle work if it ever matters.
- **Tests are thorough:** signal type asserted both directions (SIGUSR2 yes, SIGINT no), cancel sends nothing and exits nonzero, idle prompt opens the start menu, and per-state option sets verified including negative assertions.

tahadx (commented):

## Rich code review — #7484 (pause/resume a screen recording)

Solid, well-tested feature. SIGUSR2 toggles gpu-screen-recorder's pause (SIGUSR1 is replay-save), and the marker file `/tmp/omarchy-screenrecord-paused` is the single source of truth since the process doesn't report its state back — a clean design that keeps pause/resume in lockstep with the live process.

### Verified
- `pkill -SIGUSR2` matches the `screenrecording_active()` pattern (`^gpu-screen-recorder`), consistently.
- `--pause-recording` is a no-op exiting nonzero when nothing is recording; `--prompt` opens the chooser when active and the start menu when idle.
- Menu `when` guards correctly key off the marker file; QML uses the established `stdout: SplitParser` + `bash -c` pattern (same as Bar.qml). The 2s Timer self-heal around the post-restart IPC race is a thoughtful touch.
- This branch now also carries the portal fallback (formerly #7532); I re-verified both features coexist in `start_screenrecording()` and all four suites (`screenrecording-pause`, `screenrecording-prompt`, `screenrecording-portal-fallback`, `indicator-contract`) pass.

### Nits (non-blocking)
- The marker `/tmp/omarchy-screenrecord-paused` mirrors the existing `-filename` convention; both would be cleaner under `${XDG_RUNTIME_DIR:-/tmp}`, but consistency with the existing file is fine.
- If the paused-mark leftover is a concern across reboots, the existing `rm -f "$PAUSED_FILE"` in `start_screenrecording` already covers it.

Verdict: ship.

tahadx (commented):

## Rich code review — #7484 (pause and resume a screen recording mid-capture)

This PR is now **focused squarely on the stop/resume/pause feature** — the earlier concern about scope was addressed: the automatic kms→portal fallback (from the superseded #7532) has been dropped entirely, and the diff contains only the pause/resume work. Its 6 files are exactly the capture script, the menu entry, the bar indicator, and the two dedicated test suites plus the indicator-contract fixture.

### Design
SIGUSR2 toggles capture pause in gpu-screen-recorder (SIGUSR1 is replay-save instead), and since the process won't report its paused state back, a marker file (`/tmp/omarchy-screenrecord-paused`) is the single source of truth for every pause/resume. Routing each toggle through `pause_screenrecording()` keeps the CLI, the menu, and the on-screen indicator in lockstep with the live process — a clean, low-surface design.

### Verified
- **Signal correctness:** `pkill -SIGUSR2 -f "^gpu-screen-recorder"` targets the same `^gpu-screen-recorder` pattern used by the active-recording check, so it can never pause an unrelated process.
- **Defensive marker cleanup:** a paused-marker left by a recording that never reached stop (crash/logout) would invert the *first* toggle of the next session; the script explicitly drops it on a fresh (`--stop-recording` / recording start) path.
- **Prompt flow:** idle `--prompt` opens the start menu; an active recording opens a stop/pause (or stop/resume, when already paused) picker; cancel is a non-signaling nonzero exit.
- **Menu/QML wiring:** the menu `when` guards key off the marker file, the bar indicator uses the established `SplitParser` + `bash -c` pattern, and the indicator contract fixture pins it.
- **Tests:** `screenrecording-pause`, `screenrecording-prompt`, and the indicator contract all pass; `bash -n` on the capture script is clean.

### Nits (non-blocking)
- The `/tmp` marker mirrors the pre-existing `-filename` convention; both would be tidier under `${XDG_RUNTIME_DIR:-/tmp}`, but consistency with the existing file is fine and not a regression.
- Re-pressing the indicator while paused offers the two valid actions for the current state — the picker correctly flips labels, which is the right UX.

Verdict: well-scoped, correct, and matchingly tested. Ship.

bjarneo:
Automated duplication check: this pull request looks similar to #6470, which covers the same screenrecording pause/resume. I keep that one open and close this one to consolidate review.

PR #7486, opened 2026-08-19
Add wallpaper portal backend so Files' Set as Background works

bitbonsai:
Nautilus context menu with Set as Background

Right-click → **Set as Background** in Nautilus does nothing on Omarchy: the `org.freedesktop.impl.portal.Wallpaper` interface has no backend on Hyprland.

This PR adds a D-Bus-activated wallpaper portal backend that forwards to `omarchy theme bg set`, and makes that command adopt external images into `~/.config/omarchy/backgrounds/<theme>/` so they show up in the background switcher and survive the original file being moved.

A migration seeds the new files for existing installs without touching the user's portal routing config. `omarchy-pkgs` needs the two new path mappings in docs/file-layout.md. Tested on Omarchy 4 / Nautilus 50.2.2, plus a user test on 4.0.3.

### DIFF
--- a/bin/omarchy-portal-wallpaper
+++ b/bin/omarchy-portal-wallpaper
@@ -1,0 +1,126 @@
+#!/usr/bin/python3
+# omarchy:summary=xdg-desktop-portal wallpaper backend for Omarchy
+# omarchy:hidden=true
+"""D-Bus backend for org.freedesktop.impl.portal.Wallpaper.
+
+Lets portal-aware apps (Nautilus "Set as Background", browsers, GNOME apps)
+set the Omarchy desktop background. Requests are forwarded to
+omarchy-theme-bg-set, which adopts the image into the current theme's user
+background collection and updates the live shell background.
+
+Activated on demand by D-Bus via /usr/share/dbus-1/services/.
+"""
+
+import subprocess
+from urllib.parse import unquote, urlparse
+
+from gi.repository import Gio, GLib
+
+BUS_NAME = "org.freedesktop.impl.portal.desktop.omarchy"
+OBJECT_PATH = "/org/freedesktop/portal/desktop"
+
+RESPONSE_SUCCESS = 0
+RESPONSE_ERROR = 2
+
+WALLPAPER_XML = """
+<node>
+  <interface name="org.freedesktop.impl.portal.Wallpaper">
+    <method name="SetWallpaperURI">
+      <arg type="o" name="handle" direction="in"/>
+      <arg type="s" name="app_id" direction="in"/>
+      <arg type="s" name="parent_window" direction="in"/>
+      <arg type="s" name="uri" direction="in"/>
+      <arg type="a{sv}" name="options" direction="in"/>
+      <arg type="u" name="response" direction="out"/>
+    </method>
+  </interface>
+</node>
+"""
+
+REQUEST_XML = """
+<node>
+  <interface name="org.freedesktop.impl.portal.Request">
+    <method name="Close"/>
+    <signal name="Response">
+      <arg type="u" name="response"/>
+      <arg type="a{sv}" name="results"/>
+    </signal>
+  </interface>
+</node>
+"""
+
+request_node = Gio.DBusNodeInfo.new_for_xml(REQUEST_XML)
+request_registrations = {}
+
+
+def uri_to_path(uri):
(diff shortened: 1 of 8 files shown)


sidm1983:
@bitbonsai, thank you for this fix. I ran into the same issue today and found this PR. Hopefully it gets merged soon. In the meantime, I tested it on Omarchy 4.0.3 and it works!

**Setup:** Omarchy 4.0.3-1, Nautilus 50.3.1, xdg-desktop-portal 1.22.1, xdg-desktop-portal-hyprland 1.4.1, xdg-desktop-portal-gtk 1.15.3

**How I tested:** I ran `bin/omarchy-portal-wallpaper` and `bin/omarchy-theme-bg-set` from `79223a1` as a local user service.
- The `.portal` file was picked up from `~/.local/share/xdg-desktop-portal/portals/`, so the user-level paths in the migration work.
- `SetWallpaperURI` via `gdbus` with `show-preview: true`: the service started on first use, the background changed, no access dialog appeared, and the image was copied into `~/.config/omarchy/backgrounds/<theme>/`.
- In Nautilus, right-click > Set as Background worked on several images, and each one appeared in the background switcher.

Not tested: the migration itself, packaging, and lock-screen-only requests.

**One issue in the migration:** `$config` points to `$OMARCHY_PATH/etc/xdg/...`, but packaged installs put `etc/**` straight into `/etc/`. There's no `/usr/share/omarchy/etc` on my system (omarchy-settings 4.0.3-1), so the `[[ -f ... ]] || exit 0` check would fail and the migration would be marked done without copying anything.

I think `$config` could be dropped from the migration entirely. The existing `etc/**` mapping already installs `/etc/xdg/xdg-desktop-portal/hyprland-portals.conf`. xdg-desktop-portal 1.22 checks config files in priority order for each interface and moves on to the next file when a higher-priority one has no usable entry, so the `/etc/xdg` copy still routes `Wallpaper` for users with their own `~/.config` version. I confirmed this with only `default=hyprland;gtk` in `~/.config` and the `Wallpaper` line in a lower-priority config. It also avoids the current `cp` overwriting the user's `~/.config` file, which would wipe any custom portal settings (e.g. a custom file chooser).

bitbonsai:
Fixed in fdf6b98. Thanks for the write-up — you were right about the config.

The migration was broken on real packaged installs in two ways. It read the routing config out of `$OMARCHY_PATH/etc/`, which only exists in dev checkouts; packaged, `etc/**` goes straight to `/etc/`, so the guard bailed and nothing got copied at all. And the same `cp` would have stomped your `~/.config/xdg-desktop-portal/hyprland-portals.conf`, taking a custom file chooser down with it. Both are gone: the migration now ships only the `.portal` and the D-Bus activation entry.

On dropping the routing line: I tested on xdg-desktop-portal 1.22.1 with no Wallpaper entry in any config, and the portal still dispatches the interface to whatever registered backend implements it — here, the omarchy one. So the line was belt-and-suspenders, not a requirement. It still lands system-wide via `etc/**` once omarchy-settings picks it up.

bitbonsai:
@ErikMelton @ryanrhughes — bumping this one since the packaging side is now in place, not just the feature. The omarchy-pkgs mapping PR is at omacom/omarchy-pkgs#508 (portal definition + D-Bus activation entry to system paths; the etc/xdg portals config already ships via the existing etc/** mapping).

Feature has been working locally since Aug 19 (Nautilus 50.2.2 on 4.0.4): SetWallpaperURI via gdbus, right-click in Nautilus, cold-start D-Bus activation, image adoption into the theme switcher. A second user confirmed on 4.0.3 / Nautilus 50.3.1 (comment above). Nothing left on my side for either PR.

bitbonsai:
@ErikMelton @ryanrhughes — bumping this one since the packaging side is now in place, not just the feature. The omarchy-pkgs mapping PR is at omacom/omarchy-pkgs#508 (portal definition + D-Bus activation entry to system paths; the etc/xdg portals config already ships via the existing etc/** mapping).

Feature has been working locally since Aug 19 (Nautilus 50.2.2 on 4.0.4): SetWallpaperURI via gdbus, right-click in Nautilus, cold-start D-Bus activation, image adoption into the theme switcher. A second user confirmed on 4.0.3 / Nautilus 50.3.1 (comment above). Nothing left on my side for either PR.

PR #7487, opened 2026-08-19
Fix broken Wi-Fi on 2015–2017 MacBook Pros by shipping the missing NVRAM file

csvenke:
## Issue

- 2015–2017 MacBook Pros (and any Apple machine with a BCM43602, PCI `14e4:43ba`/`43bb`/`43bc`) ship with `brcmfmac43602-pcie.bin` but **no board NVRAM** (`brcmfmac43602-pcie.txt`) — `linux-firmware` has never carried it.
- Without the NVRAM, the firmware's country detection breaks ([kernel bug 193121](https://bugzilla.kernel.org/show_bug.cgi?id=193121)): only 2.4 GHz networks appear, signal strength collapses, and the WPA 4-way handshake times out (`deauth reason=15`).
- `wpa_supplicant` misreports that timeout as *"pre-shared key may be incorrect"* and NetworkManager as *"psk mismatch"* — so affected users chase their passphrase instead of the real cause.
- This is a **second, distinct firmware defect** on the same chip as #6652. That quirk (`feature_disable=0x82000`) fixes the WPA-offload failure against WPA2/WPA3 transition APs; it was active on the machine below the whole time and was not sufficient. Both defects surface as "wrong password", which is why this one hides behind it.

## Fix

Vendor the known-good NVRAM (Apple Boot Camp driver lineage, from bug 193121) and install it for Apple machines carrying a BCM43602:

- `default/firmware/apple/brcmfmac43602-pcie.txt` — the NVRAM with `ccode=0`/`regrev=1`, which bypasses the broken country detection, plus a provenance header.
- `install/hardware/apple/fix-brcmfmac-nvram.sh` — install-time leaf, gated on Apple DMI vendor + `14e4:(43ba|43bb|43bc)` (narrower than the supplicant quirk's list: this NVRAM is 43602-family only, and BCM4360 Macs run the out-of-tree `wl` driver and are excluded for free). Substitutes the live interface MAC into the `macaddr=` line; strips it when no MAC is discoverable (the OTP fallback is proven). Never overwrites an existing file.
- `migrations/<epoch>.sh` — the same repair for existing installs: idempotent, env-overridable paths, never clobbers, sets `reboot-required` only when it actually installed — it does not reload `brcmfmac`, which would drop the Wi-Fi connection carrying the update.
- `test/shell.d/brcmfmac-nvram-test.sh` — 13 assertions covering gates, exclusions, MAC substitution/strip, idempotency, no-clobber, and the escalation contract.

## Evidence

MacBookPro14,3 (2017, BCM43602), before → after installing the NVRAM and reloading `brcmfmac`:

| | before | after |
|---|---|---|
| bands visible | 2.4 GHz only | 2.4 + 5 GHz |
| signal (same spot) | −92 dBm | −72 dBm |
| RX bitrate | 13 Mbit/s | 216 Mbit/s |
| WPA handshake | timeout, `reason=15` | completes in <1s |

Independent confirmations on 13,2 and 14,3: [mbp-2016-linux#172](https://github.com/Dunedan/mbp-2016-linux/issues/172), [mbp-2016-linux#208](https://github.com/Dunedan/mbp-2016-linux/issues/208).

## Provenance & licensing

The NVRAM is Broadcom-proprietary, extracted from Apple's Windows driver package — which is exactly why `linux-firmware` can't ship it and bug 193121 has been open for years. Vendoring it here is the same trade-off, made at the distro layer; if you'd rather not carry the file in-repo, the alternatives are fetch-at-install (the canonical bugzilla source is now behind an anti-bot wall, and the gists are a supply-chain risk) or detect-and-instruct (weak). Flagging it explicitly so the call is yours.

### DIFF
--- a/default/firmware/apple/brcmfmac43602-pcie.txt
+++ b/default/firmware/apple/brcmfmac43602-pcie.txt
@@ -1,0 +1,268 @@
+# NVRAM for the Broadcom BCM43602 (PCI 14e4:43ba/43bb/43bc) in 2015-2017 Apple
+# Macs, extracted from Apple's Windows/Boot Camp driver package and surfaced via
+# kernel bug 193121, where it is attached as
+# https://bugzilla.kernel.org/attachment.cgi?id=285753. linux-firmware ships
+# only the .bin for this part, so without this file the firmware's country
+# detection never completes: scans come back 2.4GHz-only at crippled power and
+# the WPA four-way handshake times out.
+#
+# ccode=0/regrev=1 bypass that broken country detection by pinning the "world"
+# regulatory domain instead of asking the AP. The values below are otherwise the
+# board's own calibration data; do not edit them.
+#
+# Lines starting with # are comments to humans only; the brcmfmac NVRAM parser
+# ignores them.
+#
+# macaddr= below is the firmware's built-in fallback default (00:90:4c is the Epigram/Broadcom OUI, not a real
+# device MAC); the install leaf and migration always substitute the machine's live MAC or strip the line, so it is never used verbatim.
+boardrev=0x1101
+
+sromrev=11
+boardtype=0x073e
+vendid=0x14e4
+devid=0x43ba
+
+macaddr=00:90:4c:0d:f4:3e
+
+ccode=0
+regrev=1
+
+
+# Board flags:
+# X BFL_BTCOEXIST          = 0x00000001   This board implements Bluetooth coexistence
+#   BFL_EXTLNA             = 0x00001000   This board has an external LNA (2G)
+#   BFL_FEM_BT             = 0x00400000   This board has shared antenna w/ BT
+# X BFL_PALDO              = 0x02000000   Power topology uses PALDO ? - CHECK
+#   BFL_EXTLNA_5GHz        = 0x10000000   Board has an external LNA in 5GHz band
+boardflags=0x02000001
+
+# Board flags 2:
+#   BFL2_BT_SHARE_ANT0     = 0x00800000   share core0 antenna with BT
+# X BFL2_LNA1BYPFORTR2G    = 0x40000000   acphy, enable lna1 bypass for 2G clip lo
+# X BFL2_LNA1BYPFORTR5G    = 0x80000000   acphy, enable lna1 bypass for 5G clip lo
+# X BFL2_SPUR_WAR          = 0x00000200   Board has a WAR for clock-harmonic spurs
+#   BFL2_2G_SPUR_WAR       = 0x00002000   Board has a WAR to reduce and avoid clock-harmonic spurs in 2G band
+boardflags2=0xC0000000
+
+# Board flags 3:
+# X BFL3_RCAL_WAR          = 0x00000008   acphy rcal war active on this board (mainly for 4335a0)
+# X BFL3_FEMTBL_FROM_NVRAM = 0x00000100   acphy, femctrl table is read from nvram
+boardflags3=0x40000108
+
+#btc_mode=0
+
+#### added rx de-sense
+
+btcdyn_flags=0x7
+# Media profile
+btcdyn_profile_type=0x2  
+btcdyn_dflt_dsns_level=0
+btcdyn_low_dsns_level=0
+btcdyn_mid_dsns_level=21
+btcdyn_high_dsns_level=22
+btcdyn_default_btc_mode=4
+# --- number of rows in the array vars below ---
+btcdyn_msw_rows=1
+btcdyn_dsns_rows=1
+# --- mode switch data rows (max is 4) ---
+btcdyn_msw_row0=1,-16,-95,-100
+# --- desense switching data rows (max is 4) ---
+btcdyn_dsns_row0=4,-16,-63,-95
+
+#### end of rx de-sense
+
+
+xtalfreq=40000
+otpimagesize=484
+nocrc=1
+muxenab=0x1
+btc_params82=0x60
+
+
+
+########################################################
+# RF Control Definitions
+
+antswitch=0
+rxchain=3
+txchain=3
+aa2g=3
+aa5g=3
+femctrl=10
+
+# antenna gain per core g-band
+agbg0=2
+agbg1=2
+
+# antenna gain per core a-band
+aga0=2
+aga1=2
+
+# RFSWCTRL 2G and 5G iLNA
+#            WL_TX,     WL_RX,     WL_RX_ATTN, BT_TX_RX, WL_MASK
+swctrlmap_2g=0x04010401,0x08080808,0x04010401,0x00000000,0x000000ff
+swctrlmap_5g=0x08080808,0x04010401,0x08080808,0x00000000,0x000000ff
+
+swctrlmapext_2g=0x00000000,0x00000000,0x00000000,0x000000,0x003
+swctrlmapext_5g=0x00000000,0x00000000,0x00000000,0x000000,0x003
+########################################################
+
+# Bypass offsetting PAPD_EPS_TABLE_PER_TX_INDEX feature
+epsdelta2g0=0,-1,0,0,0,0,0,0
+epsdelta2g1=0,-1,0,0,0,0,0,0
+
+########################################################
+# Rx gain and RSSI parameters
+#
+# Default so do not set:
+# rxgaincal_rssical=0
+# rssi_cal_rev=0
+# rxgains[25]gtrisoa[01]
+# rxgains[25]g[mh]trelnabypa[01]=0
+
+# BW20,BW40
+rssicorrnorm_c0=4,4
+rssicorrnorm_c1=4,4
+
+# subband5gver=4 =>
+# BW20,BW40,BW80  <5250|<5500|<5745|>=5745
+#                  <70m| <100| <149|>=149
+rssicorrnorm5g_c0=1,2,3,1,2,3,1,2,3,1,2,3
+rssicorrnorm5g_c1=1,2,3,1,2,3,1,2,3,1,2,3
+
+########################################################
+
+
+########################################################
+# 20 MHz in 40 MHz Power Offsets and Duplicate Modes
+# 2G and 5G bands
+
+sb20in40hrpo=0x0
+sb20in40lrpo=0x0
+
+dot11agduphrpo=0x0
+dot11agduplrpo=0x0
+########################################################
+
+
+########################################################
+# PAPD parameters
+fastpapdgainctrl=0
+
+########################################################
+# 2G TSSI / PA Parameters
+
+tworangetssi2g=1
+tssipos2g=1
+extpagain2g=2
+pdgain2g=2
+
+# 2G Max Power
+maxp2ga0=74
+maxp2ga1=74
+
+# 2G PA Parameters
+# Order is A1,B0,B1
+#pa2ga0=-125,6514,-739  used for p113 and p115
+#pa2ga1=-141,6391,-738  used for p113 and p115
+#pa2ga0=-169,6473,-759
+#pa2ga1=-174,6462,-759
+pa2ga0=-162,6368,-735
+pa2ga1=-170,6349,-742
+
+
+
+# 2G Power Offsets
+cckbw202gpo=0x0000
+cckbw20ul2gpo=0x0000
+mcsbw202gpo=0x99644422
+mcsbw402gpo=0x99644422
+dot11agofdmhrbw202gpo=0x6666
+ofdmlrbw202gpo=0x0022
+
+########################################################
+
+#AvVmid_c0=2,140,2,145,2,145,2,145,2,145
+#AvVmid_c1=2,140,2,145,2,145,2,145,2,145
+#AvVmid_c2=0,0,0,0,0,0,0,0,0,0
+
+# AvVmid 2GHz and 5GHz LabNotebook 43569A2_012 data from pcieir
+AvVmid_c0=2,140,2,125,2,125,2,135,2,135
+AvVmid_c1=2,140,3,100,3,100,3,100,3,100
+AvVmid_c2=0,0,0,0,0,0,0,0,0,0
+
+########################################################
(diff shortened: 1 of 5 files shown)


bernardcosta:
Can confirm this fixes it on a MacBookPro14,3 (15" 2017 Touch Bar, BCM43602 14e4:43ba, Omarchy 4.0.1, kernel 7.1.9-arch1-2). Out of the box I had exactly the symptoms described here: no `clm_blob` available in `dmesg` on every boot, only 2.4GHz networks visible, terrible signal. In a room with weak coverage the scan list was completely empty, which first sent me down the `broadcom-wl-dkms` rabbit hole (wrong driver for this chip, and it blacklists `brcmfmac` — don't).

Installed the NVRAM file from this PR into `/usr/lib/firmware/brcm/`, reloaded the driver, and the 5GHz band appeared (channels 36–165, VHT80). Getting 160+ Mbps near the router now. I'm in Germany, so ccode=0 also works fine in an ETSI regulatory domain.

Two small notes:
  - Testing without a reboot: `modprobe -r brcmfmac` fails with "module in use" until you remove `brcmfmac_wcc` first. The migration's reboot-required approach avoids this.
  - The file declares 2 antenna chains, but the 15" board is 3x3 (HT MCS caps at 0–15 instead of 0–23). Still a massive improvement over no 5GHz at all.

rand0mdud3:
A/B test on MacBookPro11,5 (BCM43602 14e4:43ba, same chip, different board): baseline has no NVRAM (no clm_blob in journal) yet 5GHz already works here (5180MHz, -30dBm, 400/400Mbit/s, 14x5GHz nets visible), so this tests cross-board safety, not efficacy.

Ran the actual leaf: installs a 7131-byte file with macaddr correctly substituted for this NIC's real (non-random) address. Reload recipe that works: nmcli disconnect + ip link down + modprobe -r brcmfmac_wcc, then brcmfmac (wcc must go first; plain modprobe -r brcmfmac fails 'in use'). Note reconnect needs explicit nmcli dev connect -- disconnect blocks autoconnect by design.

With NVRAM loaded: identical link (5180, 400/400), scans no worse (20x5GHz, within rescan variance), clm_blob still absent (expected -- NVRAM is not CLM). Reverted afterwards; baseline restored exactly. No degradation on this 3x3 board -- with one caveat: at VHT 400Mbit/s the HT MCS-cap concern from review is not exercised, so this does not settle that half of the cross-board question, only that nothing regresses at these rates.

One correction to the thread discussion: /sys/class/net/wlp4s0/perm_address does not exist at all on brcmfmac here (not all-zero, absent), so a perm_address-first fallback would need to handle ENOENT, not just zero content -- and would land back on address anyway, which is what the leaf already reads.

rand0mdud3:
Addendum -- HT/VHT capability check for the cross-board question, same MacBookPro11,5. With the NVRAM loaded and the driver reloaded, iw phy advertises byte-identical capabilities to baseline: HT MCS 0-23 and VHT MCS 0-9 on all 3 streams, link still 400/400. So the file's rxchain/txchain=3 mask does not shrink the advertised stream support on this 3x3 board -- the predicted MCS cap does not materialize at the advertisement level. Caveat: advertisement is not runtime enforcement; forcing HT MCS 16-23 would be the final proof, but that needs a 3SS AP and band manipulation, and with identical advertisements plus identical link rates there is no signal pointing at hidden harm.

makoni:
Another MacBookPro14,3 data point, and one result that bears on the cross-board question @rand0mdud3 raised.

First, confirming @bernardcosta: this PR's file does fix 5 GHz here. MacBookPro14,3 (15-inch 2017, T1), Omarchy 4.0.3, kernel 7.2.3-arch1-3, BCM43602 `14e4:43ba`, Apple subsystem `106b:0173`. Stock install sees no 5 GHz band at all (`iw phy` has no Band 2); with the file installed, 29 channels appear and the machine associates on 5240 MHz.

**On this board the file does halve advertised spatial streams.** @rand0mdud3's MacBookPro11,5 result was that `rxchain/txchain=3` left capabilities byte-identical to baseline — HT MCS 0-23 and VHT MCS 0-9 on all three streams. That does not reproduce here. Swapping between this PR's attachment 285753 and #7671's attachment 290569 on the same machine, same session, changes what the driver advertises:

| | 290569 (`boardtype=0x61b`, `txchain=7`) | 285753 (`boardtype=0x073e`, `txchain=3`) |
|---|---|---|
| HT MCS | 0-23 | 0-15 |
| VHT caps | `0x0c025820` | `0x0c015820` |
| VHT RX/TX streams | 3 | 2 |

<details>
<summary>Raw <code>iw phy info</code>, Band 2, both calibrations</summary>

```
===== 290569 (PR #7671) =====
aa2g=7  aa5g=7  boardrev=0x1421  boardtype=0x61b  ccode=00  regrev=245  rxchain=7  txchain=7

		HT TX/RX MCS rate indexes supported: 0-23
		VHT Capabilities (0x0c025820):
		VHT RX MCS set:
			1 streams: MCS 0-9
			2 streams: MCS 0-9
			3 streams: MCS 0-9
			4 streams: not supported
		VHT TX MCS set:
			1 streams: MCS 0-9
			2 streams: MCS 0-9
			3 streams: MCS 0-9
			4 streams: not supported

===== 285753 (this PR) =====
aa2g=3  aa5g=3  boardrev=0x1101  boardtype=0x073e  ccode=0  regrev=1  rxchain=3  txchain=3

		HT TX/RX MCS rate indexes supported: 0-15
		VHT Capabilities (0x0c015820):
		VHT RX MCS set:
			1 streams: MCS 0-9
			2 streams: MCS 0-9
			3 streams: not supported
		VHT TX MCS set:
			1 streams: MCS 0-9
			2 streams: MCS 0-9
			3 streams: not supported
```
</details>

So the effect looks board-dependent rather than absent: the 11,5 board keeps its three streams under this mask, the 14,3 board does not. Worth noting since #7671 gates specifically on 14,2/14,3.

**Signal readings are not comparable between the two files.** The RSSI offset is itself calibration data, and the difference here is large and reproducible — 14.5 dB on an unchanged machine, position and AP:

```
290569 (run 1):  -63.8 dBm     (mean of 5)
285753:          -78.4 dBm
290569 (run 2):  -64.0 dBm
```

I mention it because "signal got worse" is an easy and wrong conclusion to draw when comparing these two files; the word `rssi` does not appear anywhere in this thread or #7671.

**Throughput did not separate them, and I would not trust anyone's single-pass numbers here, including my own earlier ones.** A/B/A, three 50 MB downloads per phase:

```
290569 (run 1):  71.4  78.3  77.7  ->  75.8 Mbit/s
285753:          78.8  79.0  83.1  ->  80.3 Mbit/s
290569 (run 2):  66.8  64.6  48.9  ->  60.1 Mbit/s
```

The spread between the two identical control runs (15.7 Mbit/s) is over three times the A-vs-B difference (4.5 Mbit/s), so this environment cannot resolve it. One caveat that probably caps both: my AP advertises `VHT operation: channel width 1, center freq segment 1: 42, segment 2: 50`, i.e. 160 MHz, which this card does not support (`Supported Channel Width: neither 160 nor 80+80`). The link falls back to HT40 rates on both files, so neither calibration is being exercised near its ceiling. Someone with an 80 MHz AP would get a more meaningful comparison.

gunars:
Measured data from a `106b:015a` board (MacBookPro13,3), in case it is useful here.

This NVRAM file is from subsystem **`106b:0157`**. On `106b:015a` it does enable 5 GHz — and it also stops 2.4 GHz from associating. The responsible key is a single one: **`sromrev=11`**. It does not *add* 5 GHz, it *switches* which band works, by activating the foreign board's srom11 calibration set.

Measured in both directions, after a full reboot each time (`modprobe -r brcmfmac_wcc` segfaults on this kernel, so reloads were avoided):

| NVRAM | 2.4 GHz | 5 GHz |
|---|---|---|
| full `0157` file, 144 keys | `CTRL-EVENT-ASSOC-REJECT status_code=16`, **zero entries AP-side** | associates |
| same file **minus `sromrev` only** (one line) | `CTRL-EVENT-CONNECTED`, **AP registration table gains the station** | does not associate |
| no file | associates | 0 channels offered |

The falsifier throughout was the **AP-side registration table**, not the client's own status. `nmcli` is unusable as an instrument here: it returns `Secrets were required, but not provided` from its own secrets layer *before* any frame reaches the radio, which reads exactly like a radio failure. Association was driven by `wpa_supplicant` with NetworkManager off the interface.

**`boardrev` is excluded with its real value, not merely by removal.** macOS on the same machine reports via `IO80211HardwareVersion`:

```
boardid: 0x15a   boardvendor: 0x106b   boardrev: 0x1108
```

The file declares `boardrev=0x1101`. Setting it to the measured `0x1108` — one changed line — did **not** restore 2.4 GHz (same `status_code=16`, same zero entries AP-side), while 5 GHz kept associating.

Two further measured facts:

- With no CLM blob the card does not see DFS channels at all. An AP on 5500 MHz was invisible at 23 dBm; the same AP on 5180 MHz appeared immediately. Any 5 GHz AP for these machines has to be on ch 36–48.
- Apple EFI passes **no** device properties to this PCI device: with `dump_apple_properties` and `log_buf_len=32M`, 91 devices were dumped and `0000:03:00.0` received zero. macOS IORegistry contains no `wifi-calibration-msf` or `wifi-antenna-sku-info` either. OTP on this card reads blank.

I am not proposing a fix and not asserting a cause beyond the table above. Whether correct srom11 data for `106b:015a` exists in retrievable form is unknown to me — the three sources I could measure were all empty.

makoni:
@gunars — a third board, and on this one `sromrev=11` does not switch bands.

MacBookPro14,3, Apple subsystem `106b:0173` (same machine as my earlier comment, now on kernel 7.2.5-3-omarchy). This PR's file unmodified, 147 keys, `sromrev=11` present; identified by the same keys as before, `boardrev=0x1101` / `boardtype=0x073e` / `txchain=3`. Both bands work at the same time:

| | result |
|---|---|
| 2.4 GHz | associates — `CTRL-EVENT-CONNECTED`, 2457 MHz (ch 10), NetworkManager then completes activation |
| 5 GHz | associates, 5240 MHz, as reported earlier |
| phy channel list | 14 on 2.4 GHz, 29 on 5 GHz |

So the three boards in this thread now read: the file originates from `106b:0157`, your `106b:015a` loses 2.4 GHz with it, this `106b:0173` keeps it. That is the same shape as the other divergence here — the halved stream count I measured on this board did not reproduce on @rand0mdud3's 11,5 board either. Board-dependent rather than absent, in both cases.

**On your instrument warning.** Agreed that nmcli is not evidence, and I did not rely on it — but my falsifier is weaker than yours and I want that on the record. I do not control this AP, so I have no registration table. What I have is the supplicant's own `CTRL-EVENT-CONNECTED`, i.e. the 4-way handshake completed, which the AP cannot complete without accepting the station. Association was driven through NetworkManager; the evidence is taken from `wpa_supplicant`, not from NM's return code.

**DFS: I can neither confirm nor refute.** There is no AP on a DFS channel here to test against. For what it is worth, on this board those channels are present in the phy list rather than missing:

```
* 5260 MHz [52]  ... No IR | Radar detection | DFS state: usable (for 734 sec) | DFS CAC time: 60000 ms
  ... through ...
* 5720 MHz [144] ... No IR | Radar detection | DFS state: usable (for 734 sec) | DFS CAC time: 60000 ms
```

`no clm_blob available (err=-2)` is logged here too. So if these cards cannot hear DFS beacons without a CLM blob, on this board at least the mechanism is not cfg80211 dropping the channels.

**One methodology note that may save you reboots.** You avoided module reloads because `modprobe -r brcmfmac_wcc` segfaults on your kernel. On 7.2.5-3-omarchy the reload is clean:

```
modprobe -r brcmfmac_wcc && modprobe -r brcmfmac && modprobe brcmfmac
```

The interface returns and reassociates in 7 seconds, with nothing in dmesg beyond the expected `bus is down` during teardown and a normal firmware reload. Different kernel from yours, so this is a data point rather than a contradiction.

rand0mdud3:
Fourth board for the matrix: MacBookPro11,5, Apple subsystem 106b:0152 (BCM43602 14e4:43ba). Both bands work with and without the file here (associated 5180 MHz, 12+ 5GHz nets visible stock), and with the file loaded the advertised caps stayed byte-identical (HT 0-23, VHT 3x0-9) with unchanged link rates -- so 0152 neither needs the file nor is harmed by it at VHT rates.

monomyth:
@bjarneo @csvenke Following the duplicate closure of #7671, could we agree on the consolidation requirements before treating the two PRs as interchangeable?

At current heads `eb8c83e8` here and `d5493deb` in #7671:
- This installer and migration delete `macaddr=` when no interface MAC is available. #7671 documents an actual 14,3 firmware failure on that path and instead preserves the required key with a per-machine fallback; it also rejects the shared Broadcom placeholder.
- This PR applies one calibration to Apple `43ba/43bb/43bc` devices without a model gate. #7671 restricts its different calibration to 14,2/14,3; #10141 keeps 13,3 calibration separate. Existing reports in this thread already show board-dependent band/spatial-stream behavior.
- #7671 preserves existing generic/DMI-specific calibration, including compressed files, and stages the two destination names with cleanup on a second-rename failure. Those safeguards are not present in this leaf's direct single-file install.

Consolidation is reasonable, but please retain the required-key/unique-MAC behavior, board-scoped calibration selection, existing-file protections and failure-path tests. The missing-MAC strip case is not just duplicate implementation style. Calibration redistribution/provenance remains a separate maintainer decision.

Our draft #10141 contains the shared #7671 work plus its 13,3 extension. I have marked that dependency as unresolved after the closure; I am not rebasing it onto this broader installer as-is or changing installed firmware. Can maintainers confirm these differences should be reconciled here, or reconsider keeping #7671 open until that is done?

bjarneo:
@monomyth 

Hey, sorry about this being part of the automated flow. Let us open the PRs again! Still tweaking the duplication runner so I appreciate the feedback.

PR #7493, opened 2026-08-19
Make the Display panel aware of orientation, with rotate lock and auto-rotation

davydotcom:
The Display panel handles pixel ratio, brightness, and text size, but has nothing to say about which way up the screen is. That leaves out two groups: laptops whose panel is a tablet that folds around, and anyone standing a monitor on its end.

This adds orientation as a first-class display setting — manual rotation everywhere, and automatic rotation on machines with an accelerometer.

## Panel

A new ORIENTATION section below SCALE, targeting the focused monitor the same way SCALE does:

```
SCALE                        eDP-1
[ 1x ][ 1.25x ][ 1.6x ][ 2x ][ 3.2x ][ 4x ]

ORIENTATION                  eDP-1
[ Normal ][ 90° ][ 180° ][ 270° ]
🔒 Rotate lock                    ✓
```

Both join the existing keyboard cursor model — `j`/`k` between sections, `h`/`l` across presets, Return to activate. The rotate lock row only appears where there is an accelerometer to lock out. A display parked on a flipped transform (4–7) matches no preset and highlights none, rather than claiming one of the four rotations it is not on.

## Commands

- `omarchy display orientation [normal|90|180|270|next|previous] [--monitor NAME] [--transient]` — applies via `hyprctl eval` and persists to `~/.local/state/omarchy/toggles/hypr/monitor-<name>-transform.lua`, so rotation survives a restart. Returning to `normal` drops the override so `monitors.lua` speaks for the display again. Accepts the `left`/`right` names wlr-randr and sway users already type.
- `omarchy toggle rotate-lock [on|off|toggle|--status]`
- `omarchy hw accelerometer` — IIO detection
- `omarchy-display-autorotate` (hidden) — session daemon, started from Hyprland autostart

## Notes on the details that bit

**Touch input follows the panel.** A rotated screen whose taps land in the old orientation is worse than not rotating at all, so the built-in panel's touchscreen is rotated with its output. Only the built-in one: an external monitor's touchscreen is a different device on a different output, and guessing which is which would rotate the wrong one.

**Rotate lock only gates the accelerometer.** The pills still work while it is on, matching how phone rotation locks behave. The lock is checked per-event rather than around the reader, so unlocking picks up the orientation the device is already in instead of waiting for it to be moved again.

**Auto-rotation is transient.** The accelerometer is the source of truth only while unlocked, so it never rewrites what the user persisted.

**Sensor detection covers three interfaces.** Polled sensors publish `in_accel_x_raw`, hid-sensor-hub ones `in_accel_x_input`, and buffer-only sensors neither — they only declare the axis as a scan element. The mxc4005 this was developed against is the third kind, so a narrower check would have missed it entirely.

**The daemon reaps its sensor reader.** `monitor-sensor` keeps iio-sensor-proxy polling for as long as it stays connected, and it is a separate process. The lock is held on the daemon's own pid rather than a subshell so nothing outlives a kill, and there is a test asserting the reader does not survive it.

**`omarchy-monitor-state` grows three lines on the end**, not in the middle — the panel reads that output by line index, so anything inserted above shifts every field below it into the wrong property.

`iio-sensor-proxy` is installed only where there is a sensor to read, via `install/hardware/accelerometer.sh` for new installs and a migration for existing ones.

## Testing

`./test/shell` and `./test/cli` pass — 2287 and 116 assertions. Three new test files cover the CLI, sensor-orientation mapping, lock gating, disabled-panel handling, touch rotation scope, and daemon cleanup.

Verified in the running UI on a convertible with a rotated 1600x2560 panel and a touchscreen: both lock states render, keyboard navigation walks the new sections, Return on the rotate lock row toggles the flag and updates the row, and the touchscreen rotates with the display.

https://claude.ai/code/session_014QDMrgifw4mGevBCEAtA3A


davydotcom:
Note: This makes omarchy solid on a GPD Pocket 4 with its sideways screen and rotatable touch panel.

devteapot:
Tested this on a Framework Laptop 12 with Hyprland 0.56.2. The overall approach is a good fit, but I found three compatibility gaps on this hardware:

1. A transform-only `hl.monitor` call resets unspecified monitor state here (the internal panel changed from scale 1.25 to 1.5), so the live mode/position/scale need to be replayed with the transform.
2. `iio-sensor-proxy` 3.9 detects the Framework IIO devices but `monitor-sensor` remains silent after the proxy fails to use the cros-ec buffered interface. The raw `accel-display` sensor works, and `cros-ec-lid-angle` can gate it to tablet mode.
3. A Hyprland reload drops transient monitor/input transforms without causing another sensor event, leaving rotation stale until the device moves.

I opened a tested follow-up against this branch with fixes and coverage for all three: https://github.com/davydotcom/omarchy/pull/1

BlakeEvans22:
Tested successfully on a **Lenovo IdeaPad 5 2-in-1 14AKP10 (83KT)**.

- CPU: AMD Ryzen AI 5 340
- GPU: AMD Radeon 840M
- Display: built-in 14″, 1920×1200, 60 Hz
- Omarchy: 4.0.2-1
- Hyprland: 0.56.2-1
- Kernel: 7.1.9-arch1-2
- iio-sensor-proxy: 3.9-1
- Accelerometer: detected as `accel_3d`

Applied this PR at `f9472765` to the v4.0.2 source without conflicts. After installing `iio-sensor-proxy`, tested the PR’s rotation daemon in the running session. Automatic screen rotation works great when folding and turning the laptop.

This confirms the rotation daemon on this hardware; the full Display panel integration hasn’t been tested yet. Thanks for adding this!

PR #4870, opened 2026-03-02
Layout aware bindings

ilajosmanov:
## Summary

Keybindings for resize, fullscreen (`SUPER ALT F`), and split toggle (`SUPER J`) now detect the active workspace layout and dispatch the appropriate command for both **scrolling** and **dwindle** layouts.

Since Omarchy already ships `omarchy-hyprland-workspace-layout-toggle` to switch between layouts, the tiling bindings should adapt accordingly. Currently, resize/fullscreen/split only work correctly with dwindle.

### Three new bin scripts

| Script | Scrolling behavior | Dwindle behavior |
|--------|-------------------|-----------------|
| `omarchy-hyprland-window-resize` | `layoutmsg colresize` for horizontal, `resizeactive` for vertical | `resizeactive` for both axes |
| `omarchy-hyprland-window-fullscreen-toggle` | Toggles between `colresize 1.0` and configured `column_width` default | `fullscreen 1` (maximize toggle) |
| `omarchy-hyprland-window-split-toggle` | `promote` (split out) or `movewindow` (merge into adjacent column) | `layoutmsg togglesplit` |

### Updated bindings in `tiling-v2.conf`

- `SUPER J` (toggle split) -> `omarchy-hyprland-window-split-toggle`
- `SUPER ALT F` (full width) -> `omarchy-hyprland-window-fullscreen-toggle`
- `SUPER +/-` and `SUPER SHIFT +/-` (resize) -> `omarchy-hyprland-window-resize`

### No migration needed

`tiling-v2.conf` lives in `default/` and is sourced directly from `~/.local/share/omarchy/`, so changes take effect automatically on update.

### Fullscreen toggle approach

The scrolling fullscreen toggle uses a size-comparison technique: it attempts `colresize 1` via `hyprctl --batch`, then checks if the window size actually changed. If not (already full width), it resets to the user's configured `scrolling:column_width` default. No temp files or state tracking needed.

https://github.com/user-attachments/assets/a1921c96-4d97-424c-838c-d39b59548458

PR #7500, opened 2026-08-19
Fix horizontal resize in scrolling layout

### LINKED ISSUE #5101: Cant Increase Width on Last window when on Scrolling workspace Layout (Not dwindle)
### System details

Omarchy 3.4.2

### What's wrong?

When i do Super + L, and i have two window opened, it separated to left and right, when im active on first window, i can properly increase the width by Super + `+`, the total two app width become more than monitor width, and when change to second window it correctly move, 

The bug is, we can only increase the window width for the left opened window, if we want to increase the right side window it doesnt work, im focusing on second window and press Super + `+`, it doesnt increase the width.

this is only works for increasing width, for the shrink width (Super + `-`) it still works for both

Image

Image

salemsayed:
## Summary

- use Hyprland's native `colresize` dispatcher for horizontal resizing of tiled scrolling-layout columns
- preserve the existing relative window resize behavior for dwindle, master, monocle, floating, fullscreen, and incomplete compositor state
- correct the key descriptions and manual so `Super + Minus` shrinks and `Super + Equal` grows
- add regression coverage for all horizontal step sizes, special workspaces, fallbacks, and unchanged vertical bindings

## Root cause

The generic relative window resize dispatcher changes a client edge. In the scrolling layout, the last column's right edge is already at the viewport boundary, so that resize is clamped and cannot extend the column. The layout-native `colresize` dispatcher changes the column width instead and lets the scrolling layout reposition the viewport correctly.

This revisits the resize portion of #4870 using Omarchy's current Quattro Lua binding architecture.

## Validation

- `./test/shell.d/hyprland-window-resize-binding-test.sh`
- `luac -p default/hypr/bindings/tiling.lua`
- `bash -n test/shell.d/hyprland-window-resize-binding-test.sh`
- `./test/all` — all 184 test files passed
- fresh local ISO build: `387082e8e79c8597f3a0001a0964e4865ae47cfd1b353ec3177041d0b0fdf4c4`
- clean KVM guest with QMP keyboard input: last, middle, first, and pseudo scrolling columns grew with `Super + Equal`; fine/coarse steps, reverse shrinking, fullscreen protection, floating fallback, and dwindle fallback all passed

Fixes #5101

### DIFF
--- a/default/hypr/bindings/tiling.lua
+++ b/default/hypr/bindings/tiling.lua
@@ -51,18 +51,34 @@ o.bind("ALT + SHIFT + TAB", "Reveal active window on top", hl.dsp.window.bring_t
 o.bind("CTRL + ALT + TAB", "Focus on next monitor", hl.dsp.focus({ monitor = "+1" }))
 o.bind("CTRL + ALT + SHIFT + TAB", "Focus on previous monitor", hl.dsp.focus({ monitor = "-1" }))
 
-o.bind("SUPER + code:20", "Expand window left", hl.dsp.window.resize({ x = -100, y = 0, relative = true }))
-o.bind("SUPER + code:21", "Shrink window left", hl.dsp.window.resize({ x = 100, y = 0, relative = true }))
+local function resize_width(column_delta, window_delta)
+  local column_resize = hl.dsp.layout("colresize " .. column_delta)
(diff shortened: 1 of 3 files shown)


bjarneo:
Automated duplication check: this pull request looks similar to #9015, which covers the same scrolling resize bindings. I keep that one open and close this one to consolidate review.

PR #7501, opened 2026-08-19
Apply session monitor scale to the SDDM greeter

### LINKED ISSUE #7312: SDDM greeter's hyprland.lua does not set monitor scale
### System details

quattro

### What's wrong?

Monitor scale for sddm greeter (password prompt after logout) is not set, and will be fallback to 2x. Maybe use `omarchy_monitor_scale` for this part as well.

calledtoconstruct:
Fixes https://github.com/basecamp/omarchy/issues/7312

The SDDM logout greeter starts a separate Hyprland as `sddm` with no monitor rule, so HiDPI panels fall back to Hyprland auto scale (2x). Session scale lives in `~/.config/hypr/monitors.lua`, which the greeter cannot read.

This persists `omarchy_monitor_scale` to `/etc/sddm/omarchy-monitor-scale` and has the greeter apply it.

- Greeter `hl.monitor` catch-all, default `auto`, reads the drop-in
- Hidden `omarchy-sddm-set-monitor-scale` writes the drop-in (passwordless sudoers)
- `omarchy-hyprland-monitor-scaling` and logout persist the current scale
- Migration seeds existing installs from the current user's `monitors.lua`

Tested with `test/shell.d/sddm-monitor-scale-test.sh`, `test/shell.d/monitor-scaling-test.sh`, and `./test/cli`.


calledtoconstruct:
Addressed in fbce13524daa9eb1c03c812b328b3683dbf80e39.

**1.** The helper now uses `sudo install -d -m 755` for `/etc/sddm` and `sudo chmod 644` on the drop-in, with a comment that a restrictive umask must not leave the root-owned file unreadable to sddm. The writable-dir branch also `chmod 644`s the file. Sudoers replaces the `mkdir` grant with `/usr/bin/install -d -m 755 /etc/sddm` and adds `/usr/bin/chmod 644 /etc/sddm/omarchy-monitor-scale`; the tee rule is unchanged. Tests cover umask 077 on the writable-dir branch (`o+r`) and grep the helper for `install -d -m 755` and `chmod 644`.

**2.** With no explicit argument, an undetermined or invalid scale writes `auto` instead of leaving a previous value. Explicit invalid CLI args still error without writing. The expression case now asserts the file contains `auto`, and a stale `3` is replaced with `auto` when `monitors.lua` has `3 / 2`.

**5.** `migrations/1786961462.sh` no longer uses `|| true`; a missing helper or failed elevation exits non-zero and stays pending.

Findings 3 and 4, and ordering against #7519 / #7146, are left to the maintainer.

PR #7502, opened 2026-08-19
Keep OpenCode sessions alive during theme changes

bjarneo:
## Why

Omarchy sends `SIGUSR2` to every running OpenCode process after a theme change. OpenCode uses this signal to reload its configuration. The reload disposes active instances, which can stop model requests and tool calls.

## What changed

- Stop calling `omarchy-restart-opencode` during theme changes.
- Add `omarchy-theme-set-opencode`.
- Send the current Omarchy terminal colors to each running OpenCode TTY.
- Send the terminal theme query that lets OpenCode refresh its `system` theme without reloading its worker.
- Keep `omarchy-restart-opencode` available for manual configuration reloads.

Existing Alacritty, Kitty, Ghostty, Foot, and tmux theme paths remain unchanged.

## Tests

- `./test/cli`
- `bash -n bin/omarchy-theme-set bin/omarchy-theme-set-opencode bin/omarchy-theme-osc bin/omarchy-theme-set-foot`

### DIFF
--- a/bin/omarchy-restart-opencode
+++ b/bin/omarchy-restart-opencode
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# omarchy:summary=Reload opencode configuration (used by the Omarchy theme switching).
+# omarchy:summary=Reload opencode configuration
 
 if pgrep -x opencode >/dev/null; then
   killall -SIGUSR2 opencode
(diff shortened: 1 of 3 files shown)


bjarneo:
Been running locally since this PR was made. Works well.

bjarneo:
Does not feel like the correct approach

Issue #7362, opened 2026-08-18
Quickshell heap corruption after Omarchy local-plugin hot reload (QV4::Chunk::sweep / jemalloc)


ripple0328:
### System details

- AMD Ryzen 9 3900X, 32 GiB RAM
- NVIDIA GeForce RTX 5070, driver 610.57.04
- Omarchy stable 4.0.0-1
- Hyprland 0.56.2
- Linux 7.1.8-arch1-3
- quickshell-git 0.3.0.r20.g28771c7-1, revision 28771c7c74b42e20afca0b1b63980cb46515537c
- qt6-base 6.11.1-1
- qt6-declarative 6.11.1-3
- jemalloc 1:5.3.1-3
- glib2 2.88.3-1

`omarchy-debug --no-sudo --print` completed successfully. Its full output contains unrelated local network and device identifiers, so only the relevant system facts are included here.

### What's wrong?

Quickshell crashed twice with SIGSEGV while Omarchy was running its local-plugin hot-reload path. Both symbolized cores show invalid jemalloc state, reached from different callers.

The affected plugin was a validated local `bar-widget`/`panel` plugin named `community.fnos-sync`. The first failure followed a refresh where the old and new plugin trees were byte-identical, so a changed QML payload was not required to reach the crash.

Expected behavior: replacing or rescanning a validated local plugin should either reload successfully or fail cleanly without corrupting the Quickshell process heap.

### Timeline and trigger

First sequence:

- 23:50:56: Omarchy logged `Local plugin changed, reloading: community.fnos-sync` twice as the plugin directory was atomically replaced.
- The replacement was version 0.6.1 with byte-identical contents; `diff -ru` between the old and new snapshots was clean.
- 23:58:39: Quickshell PID 163535 crashed. Its last log activity was a Hyprland window-title/urgent event for `Choose a local sync folder`.
- Quickshell automatically relaunched.

Second sequence:

- 00:11:07.791 and 00:11:07.805: Omarchy again logged `Local plugin changed, reloading: community.fnos-sync`.
- 00:11:08.947: relaunched Quickshell PID 213654 crashed during QML garbage collection, about 1.15 seconds after the reload notifications.
- systemd finished storing the core at 00:11:11.
- Quickshell automatically relaunched again.

A later plugin refresh at 00:15 survived, so the failure is intermittent.

### Symbolized core 1: PID 163535

The original fault is after Quickshell's signal-handler frames:

```text
__memset_avx2_unaligned_erms
[...]

PR #7503, opened 2026-08-19
Deregister the OSD IPC handler before its QML context dies

Skeptomenos:
## What this fixes

Every local-plugin hot reload strands a freed `IpcHandler` in Quickshell's IPC registry under target `osd`. A later IPC call dispatches through that dangling pointer and takes the whole shell down with SIGSEGV.

Root cause is upstream in Quickshell 0.3.0, not in Omarchy: `IpcHandler::~IpcHandler()` calls `updateRegistration(true)`, which resolves the engine generation via `findObjectGeneration(this)` to reach the registry. During QML destruction that lookup returns null, so the deregistration is skipped — and the `destroying` flag deliberately silences the warning that would otherwise report it. Filed as [quickshell-mirror/quickshell#898](https://github.com/quickshell-mirror/quickshell/issues/898) (root-cause analysis and a 30-line reproduction are in the comments there).

The OSD panel is the one entry point that always feels it, because it is `keepLoaded`: its handler is destroyed and recreated on every reload, so the registry accumulates one dead entry per reload. `Component.onDestruction: enabled = false` runs while the QML context is still alive, so `updateRegistration()` resolves the generation and deregisters cleanly.

## Why it is worth carrying in Omarchy

The one-line guard removes a shell-killing crash class without waiting on an upstream release. The comment states the removal condition, so it can be dropped when Quickshell deregisters correctly on destruction.

This is very likely the mechanism behind #7362 (Quickshell heap corruption after local-plugin hot reload). That report's cores prove jemalloc state was already invalid but could not identify the earlier corrupting write. `IpcHandlerRegistry::deregisterHandler` supplies one: it promotes `knownHandlers.first()` into the active map and then executes `handler->registeredState = RegistrationState(false)` — a write through the stored pointer. Once a freed handler is stranded, that write lands in freed memory, which corrupts allocator metadata and surfaces later in unrelated code (glib teardown, QV4 GC) exactly as that issue describes.

## Verification

On a live Omarchy 4.0 session (Quickshell 0.3.0-2, Qt 6.11.1):

- Before: 28 `Handler was registered but will not be used ... target osd` warnings over two days, one per hot reload, plus SIGSEGVs whose faulting frame is a `QHash` bucket walk reached from `IpcServerConnection` socket dispatch.
- After: three consecutive hot reloads produce **zero** duplicate-handler warnings, `quickshell ipc show` lists all 19 targets exactly once with full function lists and no empty-name entries, and `omarchy-shell osd ping` / `osd state` answer normally.
- `bash test/shell.d/osd-test.sh` passes on this branch.

One note for anyone grepping their own logs: ten other `omarchy.*` bar and panel targets also emit that warning, but in a different shape — all of them in the same second, in one-off bursts during a synchronized bar rebuild, where the new widgets register before the old ones deregister. Those recover on their own and need no guard. Only `osd` warns again on every reload, which is the accumulation signature this patch addresses.

Authored by Claude Fable 5 via Claude Code, on behalf of @Skeptomenos.

### DIFF
--- a/shell/plugins/osd/Osd.qml
+++ b/shell/plugins/osd/Osd.qml
@@ -114,6 +114,12 @@ Item {
 
   IpcHandler {
     target: "osd"
+    // quickshell 0.3.0's IpcHandler destructor cannot resolve its engine
+    // generation and silently skips deregistration, leaving a dangling
+    // registry pointer that later segfaults the shell (quickshell#898).
+    // Deregistering here runs while the QML context is still alive. Remove
+    // once quickshell deregisters correctly on destruction.
+    Component.onDestruction: enabled = false
     function show(payloadJson: string): string {
       root.open(payloadJson)
       return "ok"


Skeptomenos:
Confirmed: Arch Linux ARM now provides `extra/quickshell 0.3.1-1`, and that package has been installed here since 23 August. The running shell uses the same binary. We had already installed `0.3.0.r20.g28771c7-1` on 19 August.

I agree this PR is superseded. Our claim that handler promotion writes through the promoted stale pointer was incorrect, and our attribution of #7362 to that mechanism was unsupported. Closing this PR as superseded. The described teardown risk applies to 0.3.0; the cached `QPointer` protects this path on 0.3.1.

PR #7513, opened 2026-08-19
Add ctrl-n and ctrl-p menu navigation

horskyyaron:
Use:
Adds vim-style navigation shortcuts to the Omarchy menu:

- Ctrl+N selects the next item, matching Down
- Ctrl+P selects the previous item, matching Up


bjarneo:
Automated duplication check: this pull request looks similar to #7042, which covers the same menu Ctrl-N/P navigation. I keep that one open and close this one to consolidate review.

PR #7517, opened 2026-08-19
Show today's high and low in the weather panel

adamhjk:
### The gap

The weather panel never shows today's high and low.

The forecast strip is built from `openMeteoForecastDays` / `wttrForecastDays`, both of which filter through `isFutureForecastDate` (`Model.js:72`), so the strip starts at **tomorrow**. The hero block above it renders only the current temperature. Between the two, today's range is unreachable from the panel — arguably the most-wanted number after the current temp.

### The change

Show today's `H:` / `L:` under the hero temperature.

Both providers already return this data in responses we're making anyway; it was only ever being filtered out. No new network calls, no new settings.

**`Model.js`**
- `openMeteoTodayForecast(dailyForecastReport, todayString)` — scans `daily.time` for today and normalizes to the shape `buildForecastDays` already produces. It matches on date rather than trusting index `0`, since the daily window isn't guaranteed to start today.
- `wttrTodayForecast(report, todayString)` — same for the wttr.in path.
- `buildTodayForecast(...)` — prefers Open-Meteo, falls back to wttr, mirroring how `buildForecastDays` already picks a source.

**`Panel.qml`**
- A `todayForecast` property plus `todayHigh` / `todayLow`, derived through the existing `bareTempForDay(day, kind)` helper (`Panel.qml:313`) so unit selection and formatting stay shared with the forecast strip — `useImperial` and the Open-Meteo→wttr normalization come along for free.
- The hero `Row` is wrapped in a `Column` with a second row for the H/L texts, hidden when neither value resolves (no-data and wttr-only-without-today paths degrade to exactly the current UI).

### Tests

Added to `test/shell.d/weather-test.sh`, following the existing fixtures:

- today resolved from Open-Meteo, including a case where today is **not** the first daily entry
- `null` when the day is absent, and when there's no Open-Meteo data at all
- the wttr path, and `buildTodayForecast` source preference plus fallback
- `Panel.qml` derives H/L through the shared formatter, and hides the row when both are empty

`./test/shell.d/weather-test.sh` passes.

### Verification notes, in the interest of being upfront

- I ran the weather test file, not the full `./test/shell` aggregate, and not the graphical acceptance suite (which per `agents/skills/acceptance-tests.md` wants a disposable VM).
- For the visual side: I've been running this as a cloned widget on 4.0.0 for a bit against both providers, including the wttr fallback, rather than through the formal `visual-verification` flow. Happy to put it through that properly if you'd like — likewise happy to attach a screenshot.

### Open question

Styling of the new row is the part I'd most want a second opinion on. I used `Style.font.body` with the low dimmed via `Qt.darker(foreground, 1.5)`. That may under- or over-weight it against the deliberately oversized hero temperature; easy to change to whatever fits the design.

### DIFF
--- a/shell/plugins/panels/weather/Model.js
+++ b/shell/plugins/panels/weather/Model.js
@@ -153,6 +153,45 @@ function openMeteoForecastDays(dailyForecastReport, todayString) {
   return result
 }
 
+// Today's own daily entry. The forecast strip shows only future days (see
+// isFutureForecastDate), so today's high/low is resolved separately for the
+// hero read-out. Matches on date rather than trusting index 0.
+function openMeteoTodayForecast(dailyForecastReport, todayString) {
+  var daily = dailyForecastReport && dailyForecastReport.daily ? dailyForecastReport.daily : null
+  if (!daily || !daily.time) return null
+
+  var today = String(todayString || "")
+  for (var i = 0; i < daily.time.length; ++i) {
+    if (String(daily.time[i]).slice(0, 10) !== today) continue
+
+    var maxC = daily.temperature_2m_max ? daily.temperature_2m_max[i] : ""
+    var minC = daily.temperature_2m_min ? daily.temperature_2m_min[i] : ""
+    return {
+      date: daily.time[i],
+      maxtempC: roundedTemp(maxC),
+      mintempC: roundedTemp(minC),
+      maxtempF: roundedTemp(celsiusToFahrenheit(maxC)),
+      mintempF: roundedTemp(celsiusToFahrenheit(minC)),
+      openMeteoWeatherCode: daily.weather_code ? daily.weather_code[i] : null
+    }
+  }
+  return null
+}
+
+function wttrTodayForecast(report, todayString) {
+  var days = report && report.weather ? report.weather : []
+  var today = String(todayString || "")
+  for (var i = 0; i < days.length; ++i) {
+    if (String(days[i].date || "").slice(0, 10) === today) return days[i]
+  }
+  return null
+}
+
+function buildTodayForecast(report, dailyForecastReport, todayString) {
+  return openMeteoTodayForecast(dailyForecastReport, todayString)
+    || wttrTodayForecast(report, todayString)
+}
+
 // Open-Meteo bundles current conditions with the daily forecast request and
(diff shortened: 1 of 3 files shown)


hfunball:
I had the same missing-today problem and tested your implementation at `c916e0f`. Its 68 weather checks pass, and the unchanged weather files render today's range correctly in a local widget clone on Omarchy 4.0.2-1.

For the styling question, would you consider `Style.font.title` for both values, keeping your low color matched to the forecast lows? That matches the size of the Feels, Wind, and Humid values. These cropped screenshots show your original implementation and that separate optional adjustment:

**Original implementation**

Original PR weather popup

**Optional typography adjustment**

High and low matching the current-condition values

I also found a small fallback edge case: when Open-Meteo includes today's date but both daily temperatures are null, its empty result takes precedence over usable wttr data. Skipping that empty result fixes it while preserving partial ranges, zero, and negative temperatures. I added four checks to the existing weather suite; the new regression fails on the original implementation and passes with the fix.

I resolved the current layout conflict locally while retaining the newer plain-text declarations. The adapted version with the fallback fix and optional styling passes all 72 weather checks and the QML plain-text checks. Both unchanged `quattro` at `a703092` and the candidate at `ed24375` passed `./test/all` with the same environment: companion source checkouts available, controlled test settings, and 18 reported skips in each headless run. I have the compatibility, fallback, and styling changes as separate patches if useful.

I also tested the unchanged upstream weather files and that candidate in separate disposable Omarchy 4.0.2 VMs, verifying the installed file hashes. Both passed all eight graphical acceptance files and the shortcut checks with the same separate OCR whitespace adjustment: the recognizer split “SAN FRANCISCO” across lines even though it was displayed correctly. An earlier candidate run also had a bar-reveal timeout that did not recur in the final comparison; I retained that result in the detailed test notes.

The final styling revision (`6ffe4a0`) only restores your dimmed low color after those full-suite and VM runs. I reran the 72 weather checks, 19 QML plain-text checks, and 10 supplemental edge cases, and inspected the installed popup shown above. I have not rerun the full suite or VM acceptance for that one color declaration.

Thanks for putting this together—having today's range beside current conditions is useful.

adamhjk:
Nice!

PR #7519, opened 2026-08-19
Persist monitor scaling per monitor instead of via the catch-all

### LINKED ISSUE #6673: [Quattro] Display panel scale is per-monitor at runtime but persists globally, rescaling other monitors on reload
### System details

- Omarchy 4.0.0.r1607.g0a83590-1 (Quattro)
- Hyprland 0.56.2
- AMD Ryzen AI 9 HX 370 w/ Radeon 890M (integrated, amdgpu), 24 GB RAM
- 3 displays: eDP-1 (2560x1600), DP-7 (2560x1440, rotated), DP-8 (3840x2160)

### What's wrong?

The Display panel's scale buttons are per-display at runtime but persist globally. Adjusting the scale on one monitor rescales the *others* a moment later, in the same session.

`omarchy-hyprland-monitor-scaling` applies the new scale to the focused monitor correctly:

```bash
hyprctl eval "hl.monitor({ output = \"$active_monitor\", ... scale = $new_scale })"
```

But it persists by rewriting the catch-all variable in `~/.config/hypr/monitors.lua`:

```bash
if [[ -f $monitor_lua ]] && grep -q '^local omarchy_monitor_scale = ' "$monitor_lua"; then
  sed -i -E \
    -e "s|^local omarchy_monitor_scale = .*|local omarchy_monitor_scale = ${new_scale}|" \
```

`omarchy_monitor_scale` feeds `hl.monitor({ output = "", ... })`, which applies to every monitor. Writing the file also trips Hyprland's config auto-reload, so the global value lands on the other monitors within a second or two — this is not just a "doesn't survive reboot" bug, the cross-talk is immediate and unrequested.

There is no per-monitor persistence path: both branches of the persistence block write the catch-all.

The guard also doesn't match its own stated intent. The comment says:

> Persist to monitors.lua if the user still has Omarchy's generic catch-all defaults, so the scale survives reboots.

but `grep -q '^local omarchy_monitor_scale = '` still matches after the user adds explicit per-monitor `hl.monitor` lines, so it persists globally in exactly the case the comment means to exclude.

For contrast, the sibling control in the same panel does scope its target explicitly (`Panel.qml:247`):

```qml
setBrightnessProc.command = ["omarchy-brightness-display", "--no-osd", "--monitor", root.focusedMonitor, percent + "%"]
```

while `setScale` (`Panel.qml:308`) passes no monitor at all.

On a single-monitor machine the behavior is correct and invisible, which is presumably why it shipped.

### Steps to reproduce

On a **stock** `monitors.lua` (only the catch-all line, no per-monitor entries) with two or more displays:

1. Focus monitor A, open the Display panel, set the scale to 2.
2. Focus monitor B, set the scale to 1.6.
3. Watch monitor A. Within a second or two it also becomes 1.6.

`hyprctl monitors` confirms both displays end up at the scale last chosen for B, and `monitors.lua` contains `local omarchy_monitor_scale = 1.6`.

`~/.local/state/omarchy/monitor-scaling.log` from my own machine, showing two per-monitor adjustments 24 seconds apart:

```
at=2026-08-10T08:13:43+02:00	requested=1.25	current=1.6	new=1.25	monitor=DP-7 …
at=2026-08-10T08:14:07+02:00	requested=1.6	current=1.25	new=1.6	monitor=DP-8 …
```

DP-7 was left at 1.6, not the 1.25 that was asked for.

### Expected

Scale set from the Display panel persists for the display it was applied to, and leaves other displays alone.

### Workaround

Give each output an explicit `hl.monitor` line with a literal scale, so the catch-all can't reach it. The panel's scale buttons then stop persisting for those monitors (they revert on the next reload), but at least the displays no longer drag each other around.

### Possibly related

#4785 — scale cycling resetting multi-monitor layout, from the `monitors.conf` era. The runtime `hyprctl eval` in the current script still passes `position = "auto"`, which may be the same underlying behavior surviving the Lua rewrite.

### LINKED ISSUE #7242: omarchy hyprland monitor scaling reverts instantly when monitors.lua has per-monitor rules
## Summary
`omarchy hyprland monitor scaling` (also the menu Display scale control) applies the new scale for ~0.1 s, then the monitor snaps back to the previous scale. The change can never be kept.

## Environment
- Omarchy 4.0.0-1, Hyprland 0.56.2, NVIDIA (DP-2 external)
- Reproduced on a config where `~/.config/hypr/monitors.lua` has explicit per-monitor rules ported from v3 (each with `scale = 1`) in addition to the stock catch-all.

## Steps to reproduce
1. In `monitors.lua`, have both the stock catch-all AND an explicit rule like:
   ```lua
   local omarchy_monitor_scale = 1.6
   hl.monitor({ output = "", mode = "preferred", position = "auto", scale = omarchy_monitor_scale })
   hl.monitor({ output = "DP-2", mode = "2560x1440@165", position = "1920x0", scale = 1 })
   ```
2. Focus DP-2 and run `omarchy hyprland monitor scaling 1.25`.
3. Scale flashes to 1.25 then reverts to 1 within ~0.1 s.

## Root cause (verified)
`omarchy-hyprland-monitor-scaling` does two things:
1. Applies the scale at runtime: `hyprctl eval hl.monitor({... scale = 1.25})` — works.
2. Persists by sed-writing only the catch-all vars (`omarchy_monitor_scale` / `omarchy_gdk_scale`) in `monitors.lua`.

Hyprland auto-reloads its config on any config-file write (`configreloaded` event, ~30 ms later). On a config with per-monitor `scale = 1` rules, the explicit rule overrides the catch-all variable, so the reload re-applies `scale = 1` — reverting the tool's own runtime change. Differential test: `hyprctl eval` alone (no file write) holds the scale indefinitely; the revert only happens after the file write + auto-reload.

## Expected
The scale change should stick (and persist). On a stock catch-all-only config this works, because the variable the tool rewrites *is* the value the reload reads.

## Suggested fix direction
The tool should detect per-monitor `hl.monitor({ output = ... })` rules and either (a) update the matching rule's `scale` field instead of (or in addition to) the catch-all var, or (b) skip persisting when the config deviates from the stock shape (so at least it doesn't trigger the self-reverting reload). Note `clean_scale()` already knows the target output; the persistence branch is where the assumption lives.

pixelpush-io:
## Problem

`omarchy-hyprland-monitor-scaling` (also driving the Display panel's scale buttons and the SUPER+SLASH bindings) applies the new scale to the **focused monitor** at runtime via `hyprctl eval`, but persists it by rewriting the **global** `omarchy_monitor_scale` / catch-all rule in `~/.config/hypr/monitors.lua`. Both persistence branches write the catch-all; there is no per-monitor path.

Since writing the file trips Hyprland's config auto-reload (~30 ms later), this is not just a persistence bug:

- On multi-monitor setups, scaling one monitor rescales **all the others** within a second, in the same session (#6673).
- On configs that already carry an explicit `hl.monitor({ output = ... })` rule for the focused monitor, that rule overrides the rewritten catch-all on reload, so the tool's own runtime change is **instantly reverted** — the scale can never be changed at all (#7242).

On a single-monitor stock config the two writes happen to agree, which is why this is invisible in the default setup.

## Fix

Persist to the focused monitor's own rule instead of the catch-all:

- If `monitors.lua` has an `hl.monitor({ output = "<focused>" ... })` rule with a numeric scale, update that rule's scale in place — the auto-reload then re-applies the *new* value instead of reverting it.
- If the monitor has no rule, append one (`mode = "preferred", position = "auto"`, matching the catch-all defaults that governed it until now). Name-specific rules take precedence over the catch-all regardless of order, and the catch-all keeps governing every monitor that has no rule of its own.
- If the monitor's rule exists but its scale is **not** a numeric literal (the user manages the file by hand, e.g. `scale = laptop_scale`), leave the file completely untouched — any write would trigger the self-reverting reload described in #7242. The runtime change still applies for the session.

`omarchy_monitor_scale` and the catch-all rule are never rewritten anymore. GDK_SCALE persistence is unchanged.

## Tests

Reworked `test/shell.d/monitor-scaling-test.sh` for the new behavior: the stock fixture now includes the catch-all rule, and every persistence assertion checks that the focused monitor's rule carries the new scale **and** that the catch-all variable and rule are untouched. Two new regression cases: in-place update of an existing explicit rule with no duplicate appended (#7242's config shape), and a hand-managed config left byte-identical while the runtime change still applies.

`./test/cli` and `./test/shell` pass (the 3 failures in `config-test.sh`, `snapper-test.sh`, `unowned-system-paths-test.sh` pre-exist on clean HEAD in my environment).

Fixes #6673
Fixes #7242

### DIFF
--- a/bin/omarchy-hyprland-monitor-scaling
+++ b/bin/omarchy-hyprland-monitor-scaling
@@ -89,19 +89,32 @@ set_scale() {
   hyprctl eval "hl.monitor({ output = \"$active_monitor\", mode = \"${width}x${height}@${refresh_rate}\", position = \"auto\", scale = $new_scale })" >/dev/null
   audit_scale_change "$requested" "$active_monitor" "$current_scale" "$new_scale"
 
-  # Persist to monitors.lua if the user still has Omarchy's generic catch-all
-  # defaults, so the scale survives reboots.
-  if [[ -f $monitor_lua ]] && grep -q '^local omarchy_monitor_scale = ' "$monitor_lua"; then
-    sed -i -E \
-      -e "s|^local omarchy_monitor_scale = .*|local omarchy_monitor_scale = ${new_scale}|" \
-      -e "s|^local omarchy_gdk_scale = .*|local omarchy_gdk_scale = ${new_gdk_scale}|" \
-      "$monitor_lua"
-  elif [[ -f $monitor_lua ]] && grep -Eq '^hl\.monitor\(\{ output = "", mode = "preferred", position = "auto", scale = ("auto"|[0-9.]+) \}\)' "$monitor_lua"; then
-    sed -i -E \
-      -e "s|^(hl\.monitor\(\{ output = \"\", mode = \"preferred\", position = \"auto\", scale = )([^ ]+)( \}\))|\\1${new_scale}\\3|" \
-      -e 's|^hl\.env\("GDK_SCALE", ".*"\)|hl.env("GDK_SCALE", "'"$new_gdk_scale"'")|' \
-      "$monitor_lua"
+  # Persist the change for THIS monitor only, so it survives reboots without
+  # rescaling other monitors: update the monitor's own hl.monitor rule when the
+  # config has one, otherwise append one. Name-specific rules take precedence
+  # over the catch-all defaults regardless of their order in the file.
+  if [[ ! -f $monitor_lua ]]; then
+    return 0
   fi
+
+  local rule_pattern="^hl\.monitor\(\{ output = \"${active_monitor}\","
+  if grep -Eq "${rule_pattern}.* scale = [0-9.]+" "$monitor_lua"; then
+    sed -i -E "s|(${rule_pattern}.* scale = )[0-9.]+|\\1${new_scale}|" "$monitor_lua"
+  elif ! grep -Eq "$rule_pattern" "$monitor_lua"; then
+    printf 'hl.monitor({ output = "%s", mode = "preferred", position = "auto", scale = %s })\n' \
+      "$active_monitor" "$new_scale" >>"$monitor_lua"
+  else
+    # The monitor has a rule but its scale is not a numeric literal, so the
+    # user is managing this file by hand. Leave the file alone: any write
+    # trips Hyprland's config auto-reload, which would revert the runtime
+    # change just applied above.
+    return 0
+  fi
+
+  sed -i -E \
+    -e "s|^local omarchy_gdk_scale = .*|local omarchy_gdk_scale = ${new_gdk_scale}|" \
+    -e 's|^hl\.env\("GDK_SCALE", "[0-9]+"\)|hl.env("GDK_SCALE", "'"$new_gdk_scale"'")|' \
+    "$monitor_lua"
 }
 
 scale_from_current() {
--- a/test/shell.d/monitor-scaling-test.sh
+++ b/test/shell.d/monitor-scaling-test.sh
@@ -33,7 +33,24 @@ write_monitor_config() {
   cat >"$monitor_lua" <<'LUA'
 local omarchy_gdk_scale = 2
 local omarchy_monitor_scale = 2
+
+hl.env("GDK_SCALE", tostring(omarchy_gdk_scale))
+hl.monitor({ output = "", mode = "preferred", position = "auto", scale = omarchy_monitor_scale })
 LUA
+}
+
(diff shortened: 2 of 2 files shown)


thai-tu:
I’m affected by this Quattro multi-monitor scaling issue. Per-monitor scale changes currently affect the other displays after reload. Please consider merging this fix so the Display panel’s per-monitor controls persist independently. Thanks for working on it.

bjarneo:

PR #7532, opened 2026-08-19
Fall back to the portal backend when kms can't capture the monitor

### LINKED ISSUE #4804: Screenrecording of selected area wrong on external HDMI monitor
### System details

AMD Ryzen 7 7735HS, Nvidia RTX 4060, Omarchy 3.4.0 stable

### What's wrong?

When trying to screenrecord and choosing Region > Select region if on external HDMI monitor, the recorded area is wrong. In the test I selected an area on the left part of the screen, the recording is on the right part and going half off screen. If I try it on the eDP monitor it records the selected area fine.
I have not test this on previous versions.
My monitors.conf:
```
monitor = eDP-1, 1980x1080@165, auto, 1
monitor = HDMI-A-1, 3440x1440@100, auto, 1
```
https://github.com/user-attachments/assets/50d07175-f05d-44ba-b062-24b40109bb1e

### LINKED ISSUE #7530: Screen recording silently fails on external monitors driven by the dGPU (hybrid laptops)
## What's wrong

On hybrid-GPU laptops (Intel iGPU + NVIDIA dGPU), `omarchy capture screenrecording` silently fails to start a recording when the target is an external monitor wired to the dGPU. Only the laptop panel (iGPU) records. Reported by @Mohamedkrms in #7479.

## Repro

Hybrid laptop (e.g. Intel iGPU + NVIDIA dGPU), external monitor on HDMI/DP wired to the dGPU. Start a screen recording targeting the external monitor (fullscreen on it, or a region on it). No recording starts — nothing is written to `~/Videos`.

## Root cause

`gpu-screen-recorder`'s **kms backend connects to one DRM card** — the one driving the internal panel (the iGPU). A dGPU-driven external monitor's connector isn't visible to that card, so gsr logs:

```
gsr error: gsr_capture_kms_start: failed to find monitor by name "DP-1"
gsr error: gsr_capture_start failed
```

and exits without producing a file. The script's wait loop then sees the process dead and silently gives up — no recording, no message.

gsr exposes **no capture-time card selector** (only `--list-capture-options [card_path]`), so the kms backend can't be pointed at the dGPU card. The portal backend (`-w portal`, xdg-desktop-portal ScreenCast) is the documented path for "monitors driven by external GPUs" (the script already notes this in its `OMARCHY_SCREENRECORD_USE_PORTAL` opt-in), but it pops a picker and can hit EGL DMA-BUF modifier issues on some setups.

Same root cause as the closed #4804 and #4834.

## Immediate workaround (no code change)

Set the portal backend for the session:

```bash
mkdir -p ~/.config/uwsm/env.d
echo 'OMARCHY_SCREENRECORD_USE_PORTAL=true' > ~/.config/uwsm/env.d/capture
```

Then re-login so uwsm picks it up. This routes recordings through the portal picker, which can capture dGPU-driven external monitors. (Trade-off: the laptop-screen recordings also go through the portal picker instead of the kms/slurp path.)

## Planned fix

Auto-fallback: when the kms backend fails to start (gsr exits without producing the output file), retry once via the portal backend with a notification, so external monitors work without the env var. This keeps the fast kms path for the laptop panel and only invokes the portal picker when kms can't see the target monitor.

A fix is in progress in a separate PR (linked below). It needs verification on a hybrid + external-monitor setup — if you can test, please comment.

tahadx:
Closes #7530.

## What

On hybrid-GPU laptops (Intel iGPU + NVIDIA dGPU), `omarchy capture screenrecording` silently fails when the target is an external monitor wired to the dGPU — only the laptop panel records. Reported by @Mohamedkrms in #7479; same root cause as the closed #4804 / #4834.

## Root cause

`gpu-screen-recorder`'s kms backend connects to one DRM card (the iGPU driving the internal panel). A dGPU-driven external monitor's connector isn't visible to that card, so gsr logs `failed to find monitor by name "DP-1"` and exits without producing the output file. The script's wait loop then sees the process dead and silently gives up — no recording, no message. gsr exposes no capture-time card selector (only `--list-capture-options [card]`), so the kms backend can't be pointed at the dGPU card.

## Fix

When the kms launch dies without producing the output file, retry **once** via the portal backend (`-w portal`) — the documented path for external-GPU monitors (the script already exposes it as the `OMARCHY_SCREENRECORD_USE_PORTAL` opt-in). A notification tells the user the portal picker will appear, since the portal backend (xdg-desktop-portal ScreenCast) requires the user to confirm the source.

- Keeps the fast kms path for the laptop panel (the common case) — the portal is only invoked when kms can't see the target.
- Skips the retry when the user already opted into portal (`target == portal`), so a portal failure isn't retried with portal again.
- Returns nonzero on total failure so `start_screenrecording || cleanup_webcam` cleans up the webcam overlay (previously `start` always returned 0, so the overlay was never cleaned up on failure).

## Tests

New `test/shell.d/screenrecording-portal-fallback-test.sh` stubs `gpu-screen-recorder` to fail the kms launch and succeed the portal retry, and asserts:
- the kms monitor target is tried first, then `-w portal`;
- the recording file is written;
- the user is notified about the retry;
- when the user already opted into portal and it also fails, it is **not** retried again.

Existing `screenrecording-test.sh`, `bin-style-test.sh`, and the CLI metadata suite still pass.

## Notes for review

- The portal backend pops the xdg-desktop-portal picker, so `--fullscreen` on an external monitor changes UX slightly (a picker appears instead of auto-record). That's unavoidable — there's no auto-focused-monitor mode for the portal, and kms can't see the dGPU display at all. The notification explains it.
- **Needs verification on a hybrid + external-monitor setup.** I couldn't reproduce the external-monitor case locally (no external display attached). If @Mohamedkrms or anyone with a hybrid + dGPU-driven external monitor can test this branch, that would confirm the fallback lands on the portal and records.
- **Immediate workaround** (no code needed): set \`OMARCHY_SCREENRECORD_USE_PORTAL=true\` in \`~/.config/uwsm/env.d/capture\` and re-login.

### DIFF
--- a/bin/omarchy-capture-screenrecording
+++ b/bin/omarchy-capture-screenrecording
@@ -198,7 +198,33 @@ start_screenrecording() {
   if kill -0 $pid 2>/dev/null; then
     echo "$filename" >"$RECORDING_FILE"
     toggle_screenrecording_indicator
+    return 0
   fi
+
+  # gsr exited without writing the file. On a hybrid laptop the kms backend
+  # connects to the iGPU's DRM card, so it can't see an external monitor wired
+  # to the dGPU — gsr logs "failed to find monitor by name" and exits. gsr has
+  # no capture-time card selector, so fall back to the portal backend, the
+  # documented path for external-GPU monitors. The portal pops the
+  # xdg-desktop-portal picker, which is the only way to capture a dGPU-driven
+  # display; skip it when the user already opted into portal above.
+  if [[ $target != portal ]]; then
+    omarchy-notification-send -t 4000 "Retrying screen recording via the portal backend" \
+      "The kms backend couldn't capture this monitor — the portal picker will ask you to confirm the source"
+    capture_args=(-w portal -s "${RESOLUTION:-0x0}")
+    gpu-screen-recorder "${capture_args[@]}" -k auto -f 60 -fm cfr -fallback-cpu-encoding yes -o "$filename" "${audio_args[@]}" 2>>"$LOG_FILE" &
+    pid=$!
+    while kill -0 $pid 2>/dev/null && [[ ! -f $filename ]]; do
+      sleep 0.2
+    done
+    if kill -0 $pid 2>/dev/null; then
+      echo "$filename" >"$RECORDING_FILE"
+      toggle_screenrecording_indicator
+      return 0
+    fi
+  fi
+
+  return 1
 }
 
 stop_screenrecording() {
--- a/test/shell.d/screenrecording-portal-fallback-test.sh
+++ b/test/shell.d/screenrecording-portal-fallback-test.sh
@@ -1,0 +1,109 @@
+#!/bin/bash
+
+set -euo pipefail
+
+source "$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd)/base-test.sh"
+
+tmp_dir=$(mktemp -d)
+trap 'rm -rf "$tmp_dir" "$RECORDING_FILE"' EXIT
+
+stub_bin="$tmp_dir/bin"
+mkdir -p "$stub_bin"
+
+RECORDING_FILE="/tmp/omarchy-screenrecord-filename"
+GSR_LOG="$tmp_dir/gsr.log"
+NOTIF_LOG="$tmp_dir/notif.log"
+rm -f "$RECORDING_FILE"
+: >"$GSR_LOG"
+: >"$NOTIF_LOG"
+
+# gpu-screen-recorder stub: the kms launch (-w <monitor>) exits without writing
+# the file (the "failed to find monitor by name" failure on a dGPU-driven
+# external monitor); the portal retry (-w portal) touches the output file and
+# stays alive so the script's wait loop sees it succeed. OMARCHY_TEST_PORTAL_FAILS
+# flips the portal branch to fail too, for the no-retry case.
+cat >"$stub_bin/gpu-screen-recorder" <<'SH'
+#!/bin/bash
+printf '%s\n' "$*" >>"$OMARCHY_TEST_GSR_LOG"
+w=""; out=""; prev=""
+for a in "$@"; do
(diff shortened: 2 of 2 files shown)


tahadx (commented):

## What I checked

The fallback structure is sound: kms stays primary (fast path, no picker), the portal retry fires only when gsr died without writing the output file, the retry is skipped when the user already opted into portal (avoiding a pointless portal→portal retry), and returning nonzero finally lets the documented `start_screenrecording || cleanup_webcam` pattern actually clean up the webcam overlay — previously `start` always returned 0, so that cleanup could never trigger.

## Observations

- **Duplicated command tail:** the portal retry rebuilds the gsr invocation inline, so `-k auto -f 60 -fm cfr -fallback-cpu-encoding yes` now exists in two places. If encoder defaults ever change, both sites need the same edit — hoisting that tail into an array would remove the drift risk. Not blocking.
- **`0x0` vs `default_resolution`:** the fallback passes `-s "${RESOLUTION:-0x0}"` where the primary fullscreen path uses `"${RESOLUTION:-$(default_resolution)}"`. That looks deliberate — with the portal there is no auto-monitor targeting anyway, so letting the picker/portal decide makes sense — but a one-line comment would stop someone from later "unifying" the two defaults.
- **No timeout on the wait loop:** an unanswered portal picker leaves the script blocked exactly as long as the primary loop would after launch. Consistent behavior; just noting it since the picker adds a user-interaction-dependent delay to that path.
- **Tests assert exact call counts** (two on fallback success, one when portal was already chosen and failed), notification emission, and recording-file presence — the right assertions for a retry loop.

Needs the hybrid-GPU hardware verification noted in the PR body, which the author has appropriately flagged as the remaining open item.

tahadx:
Superseded by #7484: the portal fallback for capture on external-GPU monitors is now folded into the resolved pause/resume branch of #7484, which builds on the same omarchy-capture-screenrecording script. Both features coexist there and tests pass, so this PR's change is delivered via #7484.

PR #7537, opened 2026-08-19
Show limits for OpenCode's OpenAI account

oorestisime:
## Why

Codex and OpenCode can use different ChatGPT subscriptions. Token usage from both is already aggregated, but the panel only shows rate limits for the native Codex account.

## What changed

- Add an opt-in `providers.codex.openCodeOpenAiLimits` setting, disabled by default.
- Read OpenCode's active database-backed OpenAI credential, with legacy `auth.json` support.
- Show both accounts' limits, labelled by email, when the accounts differ.
- Avoid duplicate limits when Codex and OpenCode use the same account.
- Preserve native limits if the OpenCode credential is unavailable or invalid.
- Never initiate a login or modify saved credentials.

Token aggregation is unchanged; this only adds rate-limit tracking.

## Testing

- Relevant collector, updater, panel, and CLI tests pass.
- Verified with two distinct ChatGPT subscriptions through OpenCode2.
- Verified through a temporary cloned agents plugin using isolated state.
- Full suite: 180/184 test files passed; the remaining four require unavailable environment dependencies or the existing Quickshell runtime.

image

### DIFF
--- a/bin/omarchy-agent-usage-codex
+++ b/bin/omarchy-agent-usage-codex
@@ -209,7 +209,10 @@ def scan_pi_sessions():
 def scan_opencode_sessions():
   # A subscription burned entirely through opencode leaves no native session
   # files, but opencode records per-message provider, model, and token usage
-  # in its own database. Read-only: opencode may be writing right now.
+  # in its own database. Newer releases migrate that history from message to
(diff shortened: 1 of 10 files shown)


oorestisime:
@omarchybot you never reviewed this

PR #7538, opened 2026-08-19
Fix terminal image paste from universal shortcut

Skeptomenos:
## Summary

- Route terminal `SUPER+V` through a hidden MIME-aware helper.
- Use raw `CTRL+V` only for image-only clipboard offers.
- Keep `SHIFT+Insert` for text, mixed, empty, unknown, failed, timed-out, or truncated MIME probes.
- Keep GUI paste and the existing copy and cut bindings unchanged.

## Root cause

The universal terminal paste path always injected `SHIFT+Insert`. That is correct for terminal text paste, but it does not deliver an image-only clipboard to terminal applications such as Codex that handle raw `CTRL+V` as an image-paste action.

The new helper queries MIME names outside the Hyprland Lua event loop. The query has a 200 ms deadline, a 50 ms kill grace, and a 16 KiB output cap. Uncertain results fall back to the existing text-safe behavior. Synthetic key-down state is paired with key-up cleanup on normal exit and termination.

## Validation

- `test/shell.d/terminal-paste-test.sh`: 15 checks passed.
- `./test/cli`: passed.
- `bin/omarchy commands --check`: 429 commands passed.
- Bash syntax, Lua parse, and `git diff --check`: passed.
- `./test/all`: CLI passed; shell reported four environment-only failures because `omarchy-pkgs` and sandbox netlink access were unavailable. The new test passed inside the aggregate run.
- Live on Hyprland 0.56.1: reload passed and `hyprctl configerrors` was empty.
- Real bounded probes returned the expected text MIME offers and an image-only `image/png` offer.
- Live `SUPER+V` image paste passed in Herdr/Codex and Kitty/Codex. Normal terminal text paste also passed.

## Delivery note

The same generic commit is temporarily backported to the Apple Silicon fork. No ARM-specific behavior is included in this change.

### DIFF
--- a/bin/omarchy-cmd-terminal-paste
+++ b/bin/omarchy-cmd-terminal-paste
@@ -1,0 +1,78 @@
+#!/bin/bash
+
+# omarchy:summary=Paste clipboard content into the active terminal
+# omarchy:hidden=true
+
+export LC_ALL=C
+
+mime_types=""
+mime_probe=""
+
+if mime_probe=$(
+  set -o pipefail
+  timeout --signal=TERM --kill-after=0.05s 0.20s wl-paste --list-types 2>/dev/null | head -c 16385
+  probe_status=$?
+  (( probe_status == 0 )) || exit "$probe_status"
+  printf x
+); then
+  # The suffix prevents command substitution from stripping trailing newlines,
+  # so the byte limit can reliably detect a truncated MIME list.
+  mime_types=${mime_probe%?}
+fi
+
+# A truncated list cannot prove that the clipboard has no text representation.
+(( ${#mime_types} < 16385 )) || mime_types=""
+
+has_image=0
+has_text=0
+
+while IFS= read -r mime_type; do
+  case "$mime_type" in
+    UTF8_STRING | STRING | TEXT | COMPOUND_TEXT)
+      has_text=1
+      ;;
+    *)
+      case "${mime_type,,}" in
+        image/*)
+          has_image=1
+          ;;
+        text/*)
+          has_text=1
+          ;;
+      esac
+      ;;
+  esac
+done <<<"$mime_types"
+
+mods="SHIFT"
+key="Insert"
+
+if (( has_image && !has_text )); then
+  mods="CTRL"
+  key="V"
+fi
+
+send_key_state() {
+  local state="$1"
+
+  hyprctl dispatch "hl.dsp.send_key_state({ mods = \"$mods\", key = \"$key\", state = \"$state\" })" >/dev/null 2>&1
+}
+
+key_is_down=false
+
+release_key() {
+  if [[ $key_is_down == "true" ]]; then
+    key_is_down=false
+    send_key_state "up" || true
+  fi
+}
+
+trap release_key EXIT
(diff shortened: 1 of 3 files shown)


rmacy:
Reproduced this failure locally with OMP running inside Herdr: terminal `SUPER+V` receives `Shift+Insert`, so an image-only Wayland clipboard never reaches the TUI's `Ctrl+V` image handler. I also confirmed the freeze mode this PR avoids: querying `wl-paste --list-types` synchronously from the Hyprland Lua callback blocks the compositor event loop.

The external bounded probe in this PR is the right boundary. One remaining path has the same user-visible gap: `shell/plugins/clipboard/Clipboard.qml` activates image history entries through `bin/omarchy-clipboard-paste-file`, and that helper still hardcodes `wtype -M shift -k Insert -m shift` after restoring the image. As a result, selecting an image from Omarchy's clipboard palette still does not paste it into OMP/Herdr even after the universal shortcut is fixed.

Would you be open to extending this PR so the non-`--copy-only` file helper reuses the MIME-aware key dispatch after `wl-copy`, with focused coverage in `test/shell.d/clipboard-test.sh`? I can prepare that patch once you confirm the preferred helper interface.

jondkinney:
I opened #10658 before finding this earlier PR. It addresses the same Super+V failure with the same asynchronous MIME-query approach, so I'll close mine as a duplicate and consolidate the findings here. Sorry for missing this one initially.

Additional verification:
- Checked out this PR at `4492411c` and ran `test/shell.d/terminal-paste-test.sh`: all 15 checks pass.
- Ran this PR's helper against real Foot, Ghostty 1.3.1, and Alacritty 0.17.0 windows with a text clipboard: all three delivered the expected clipboard text.
- Separately, using the equivalent image route in #10658, invoked its exact Super+V Lua callback with an `image/png` clipboard. Real Foot, Ghostty, and Alacritty probes each received byte `0x16` (Ctrl+V), and all three windows had the terminal tag. That verifies image-shortcut delivery across those terminals; it is not an end-to-end Claude/Codex attachment test of this PR.

Two refinements from our implementation may be useful here:
1. Keep terminal/GUI routing in `universal_clipboard_shortcut` by accepting a terminal action callback. Copy passes `send_shortcut_once("CTRL", "Insert")`; paste passes the asynchronous helper launch. This removes the separate `universal_paste` routing function. The small refactor is in https://github.com/jondkinney/omarchy/commit/9ad851d1.
2. Capture the originating window address before launching the MIME query, and cancel the paste if the focused window has changed when the query finishes. Our implementation and regression coverage are in https://github.com/omacom/omarchy/pull/10658/files. That prevents a delayed paste from reaching a different window; retaining your bounded probe and key-release cleanup would still be valuable.

One behavior difference to decide explicitly: this PR preserves text paste for mixed image/text offers, while ours preferred an image whenever offered. That is a policy difference, not a reason to maintain another PR.

The clipboard-history image path is also covered separately by #10535 and #10536.

bjarneo:
Automated duplication check: this pull request looks similar to #7187, which covers the same terminal image paste. I keep that one open and close this one to consolidate review.

PR #7550, opened 2026-08-19
Size the menu to visible labels after typing pauses

akitaonrails:
Long menu labels are elided by the fixed 300px card even when the screen has room. The card now fits the widest result label, including the actual card borders and row gutters, while retaining its 300px minimum and screen-width cap.

Search results update immediately, but label measurement and width changes wait until typing has paused for one second. Each edit restarts the timer, including deleting the last character; provider updates during search also settle before measurement. Opening a menu or navigating to a submenu measures immediately, and closing or switching modes cancels pending work. The measured label width is cached so rebuilding the results does not repeatedly scan partially populated models. Theme spacing and borders remain reactive. dmenu and the screenrecord/font menus retain their existing width rules.

Verification on the rebased branch:

- `bash test/shell.d/menu-width-test.sh`: passes; covers debounce restarts, measurement counts, provider updates with unchanged row count, empty results, clearing search, navigation, close/reopen, dmenu, asymmetric borders, and width limits.
- Existing `menu-test.sh`, `menu-guards-test.sh`, and `menu-plugin-test.sh`: pass.
- `qmllint -I shell shell/plugins/menu/Menu.qml` and `git diff --check`: pass.
- Verified in the running full shell with keyboard input, screenshots, and a reviewed recording. Temporary timing probes measured widening 1,011 ms after the final keystroke and shrinking after 1,008 ms; no intermediate width changes while typing. Probes are not included in the change.

Original screenshots illustrating the label-fitting behavior:

**Before** — fixed 300px card, titles elided:

before

**After** — same search, card sized to the widest row:

after

### DIFF
--- a/shell/plugins/menu/Menu.qml
+++ b/shell/plugins/menu/Menu.qml
@@ -216,13 +216,14 @@ Item {
     }
     if (maxLabel <= 0) return 0
 
-    // Horizontal chrome around the label: card padding, icon gutter
-    // (reserved border + margin + 36px icon + column margin), right column
-    // margin, trail gutter (14px + reserved border + margin), scrollbar.
+    // Horizontal chrome around the label: card padding and card borders
+    // (contentLeftInset/contentRightInset are subtracted from the list),
+    // icon gutter (reserved border + margin + 36px icon + column margin),
+    // right column margin, trail gutter (14px + reserved border + margin).
     var chrome = root.contentMargin * 2
+      + Border.left(root.borderSpec) + Border.right(root.borderSpec)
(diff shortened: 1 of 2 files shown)


akitaonrails:
Good catch — both points were right. The list has no scrollbar (that allowance was fictional), and the card's `contentLeftInset`/`contentRightInset` do subtract the theme-defined border widths from the list. Fixed in 497592d9: the chrome budget now uses `Border.left(root.borderSpec) + Border.right(root.borderSpec)` instead of the phantom scrollbar term.

akitaonrails:
Closing this: the dynamic menu resizing has bugs I don't intend to pursue. Thanks for the review time.

PR #7564, opened 2026-08-20
Make the keybindings menu's Lua bind scan safe against load-time config reads

tobi:
## Problem

`omarchy-menu-keybindings` learns about Lua-only binds by `dofile`-ing the user's `hyprland.lua` under a stub `hl` whose every lookup returns a truthy `noop` table. Real configs read compositor state at load time, and the stub cannot survive that:

1. **Runaway `lua` at 100 % CPU.** `for _, m in ipairs(hl.get_monitors())` never terminates, because `noop[1]`, `noop[2]`, … are all non-nil. The menu exits, but the `lua` child keeps spinning under `systemd --user` until reboot. Found one on my laptop that had run for **27 hours** after a single `SUPER+K` (PIDs `omarchy-menu-keybindings` → `lua`, 60 % of a core, `dofile(hyprland.lua)` never returned).

   ```lua
   local noop = setmetatable({}, { __index = function() return noop end, __call = function() return noop end })
   for _, m in ipairs(noop) do end   -- never returns
   ```

2. **Silently dropped binds.** On current quattro, `default/hypr/qconsole.lua:64` does `monitor.scale <= 0`, which under the stub raises *attempt to compare table with number*. The scan is `pcall`'d, so it just stops there — every Lua bind declared after that `require` is missing from the menu (`DEBUG=1` shows `lua bind scan failed: …qconsole.lua:64`).

## Fix

- `noop.__index` answers `nil` for numeric keys (so `ipairs`/numeric loops end), `__len` is 0, and `__lt/__le`, arithmetic, `__concat`, `__tostring` degrade to harmless values instead of raising. String-key lookups behave as before.
- The scan is wrapped in `timeout 10` so a config the stub still cannot satisfy can never outlive the menu.

## Test

`test/shell.d/menu-keybindings-lua-scan-test.sh` extracts the embedded scan and runs it against a fixture `hyprland.lua` that iterates monitors, compares `monitor.scale`, does arithmetic/concatenation on stub fields, and then declares two binds. Without the fix it hangs for the full 10 s timeout and fails; with it the scan returns in milliseconds and both binds are reported. Also verified on my own config: the scan now completes with no `scan failed` and lists the Lua binds that were previously lost after `qconsole.lua`.

### DIFF
--- a/bin/omarchy-menu-keybindings
+++ b/bin/omarchy-menu-keybindings
@@ -75,7 +75,9 @@ build_lua_bind_cache() {
     LUA_BIND_DISPATCHER_MAP["$cache_key"]="$dispatcher"
     LUA_BIND_ARG_MAP["$cache_key"]="$arg"
   done < <(
-    lua <<'LUA'
+    # A config that the stub cannot satisfy must not leave a runaway lua
+    # behind after the menu closes, so the scan is time-boxed.
+    timeout 10 lua <<'LUA'
 local modifiers = { SHIFT = 1, CTRL = 4, CONTROL = 4, ALT = 8, SUPER = 64 }
 
 local function split_keys(keys)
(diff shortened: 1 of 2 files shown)


agush22:
Ran into this exact issue after upgrading to Quattro 4.0 
Pressing `SUPER + K` pegged a background `lua` process at 100% CPU because my `monitors.lua` iterates over `hl.get_monitors()` with `ipairs`.

Tested this fix locally; it resolves the infinite loop cleanly and makes the keybindings scan instantaneous. Hope to see this merged soon!

shwig1:
Confirming on omarchy 4.0.3-1 (Hyprland 0.56.2, Lua 5.5.1): a personal module that
iterated hl.get_windows() at require time hung the scan, and repeated Super+K presses
left eight orphaned lua processes at 100% CPU each for over six hours. The timeout in
this PR is the part that would have contained that; the stub fix removes the trigger.

alinuxfan:
Confirming another real-world manifestation of this on `4.0.0.r2147.g2fbac0c-1` — this one doesn't just drop a bind from the menu, it makes the menu **launch the wrong app**.

I have [`flea`](https://flea.sh) set as my default file manager. Its installer (`flea --default`) correctly unbinds and rewrites `SUPER+SHIFT+F` in `~/.config/hypr/bindings.lua`:

```lua
o.bind("SUPER + SHIFT + F", "File manager", { launch = 'flea --gui' })
```

Pressing `SUPER+SHIFT+F` directly opens flea correctly (`hyprctl binds` shows the live dispatcher resolving as expected). But selecting the same "File manager" row from the `SUPER+K` search menu opens **Nautilus** instead.

Root cause is this PR's diagnosis exactly: `hyprland.lua` loads `default.hypr.omarchy` (which pulls in `default/hypr/qconsole.lua`) *before* `hypr.bindings` (the user's own overrides). The scan crashes inside `qconsole.lua` — same `attempt to compare number with table` on a `monitor.scale > 0` check, now at line 123 rather than 64, so the line's moved but the bug hasn't:

```
$ DEBUG=1 omarchy-menu-keybindings
[DEBUG] lua bind scan failed: /usr/share/omarchy/default/hypr/qconsole.lua:123: attempt to compare number with table
```

Because the crash happens before `hypr.bindings` is ever reached, the scan's cache keeps the stale default from `default/hypr/bindings/applications.lua`:

```
$ lua <extracted scan snippet>
65      File manager    F       exec    omarchy-launch-nautilus
73      File manager (cwd)      F       exec    omarchy-launch-nautilus-cwd
```

...instead of the user's actual `flea --gui` override. So it's not just that entries silently go missing — for anyone who has overridden a default Omarchy keybind, the search menu can confidently dispatch the *old* action while the real key combo does the right thing. Would be good to land this fix; the `noop` stub degrading harmlessly under comparison (not just under `ipairs`/length) directly fixes this case too.

PR #7570, opened 2026-08-20
Install > Gaming: Quake (on-demand via the omarchy-quake package)

ryrobes:
Adds **Install > Gaming > Quake**, installed on demand — the ISO ships only this stub.

- `omarchy-install-gaming-quake`: `omarchy-pkg-add omarchy-quake` (vkQuake + launcher + panel plugin from the `[omarchy]` repo), copies the Quickshell plugin into `~/.config/omarchy/plugins/quake.omarchy` (copy, not symlink — the validator forbids symlinks), enables it with a rescan/retry so it doesn't race the shell's plugin scan, wires the Hyprland window rules, and launches the panel. First Play fetches shareware (~18MB); Steam/GOG/re-release data is auto-discovered.
- `omarchy-remove-gaming-quake`: drops the package and plugin; saves/shareware stay unless `--purge`.
- Menu rows follow the existing gaming conventions (`disabled:` on install, `when:` on remove).

One deliberate deviation to review: the installer copies the window rules to `~/.config/hypr/apps/quake-omarchy.lua` and appends a `require(...)` line to `~/.config/hypr/hyprland.lua` (with a timestamped backup) since stock configs don't auto-load that directory. If you'd rather ship the rules as `default/hypr/apps/quake.lua` in omarchy itself, happy to rework — the keybindings (Super+Shift+Q panel toggle, DM map cycling outside the Quake console for the host) are the part that shouldn't exist without the package installed.

Depends on omacom-io/omarchy-pkgs#175 — draft until `omarchy-quake` is published on edge and `omarchy-pkg-add omarchy-quake` verifies on a stock box.

https://github.com/user-attachments/assets/3f6abb84-15b1-4f16-a458-d68913924281


bjarneo:
This is really cool. I pinged you on X, but I'll continue with this PR. We were wondering about the license, so with help from AI, this was the response.

We have to ensure we don't add anything to the repo which violate any license.

## Review
  - The shareware agreement permits free redistribution under specific conditions, including keeping the agreement with the software. The downloader extracts pak0.pak but discards SLICNSE.TXT.
  - It downloads from several mirrors without verifying a known SHA-256, despite redistribution permission applying to the original unmodified archive.
  - The supposedly MIT plugin directly ships the recognizable Quake logo (https://github.com/ryrobes/omarchy-quake/blob/v1.6.2/plugin/quake-logo.svg) and Quake gameplay footage
    (https://github.com/ryrobes/omarchy-quake/blob/v1.6.2/plugin/demo-loop.mp4). The shareware legal files specifically restrict separate use of Quake names, logos, graphics, and screenshots. A blanket
    MIT declaration cannot relicense those assets.
  - Full Steam/GOG/re-release data remains proprietary and must be supplied by the user.

ryrobes:
Thanks for flagging these. I addressed each distribution/licensing concern in [omarchy-quake v1.6.3](https://github.com/ryrobes/omarchy-quake/tree/v1.6.3):

- The downloader now pins the original `quake106.zip` SHA-256 and rejects any mirror response that does not match. I also verified that all four configured mirrors currently produce the same pinned digest.
- `SLICNSE.TXT` is preserved alongside the installed shareware data; existing managed shareware installs missing it are repaired from the verified archive before play.
- The Quake logo SVGs and gameplay video were removed, along with the Qt Multimedia dependency. The panel now uses the project’s original geometric icon and a procedural backdrop.
- No game data ships in the repository or package; Steam/GOG/re-release data remains user-supplied.

The focused downloader/layout tests pass, and [omarchy-pkgs PR #175](https://github.com/omacom-io/omarchy-pkgs/pull/175) now consumes the pinned v1.6.3 release.

bjarneo (approved):

PR #7592, opened 2026-08-20
Refocus the lock password field after resume

RovshanMuradov:
## Problem
After suspend, the lock screen shows the password field but keystrokes do nothing until any mouse click. The keyboard is alive; Qt activeFocus on the field is not.
## Reproduce
1. Lock, then suspend (or suspend, which locks first).
2. Wake.
3. Type immediately, without clicking.
4. No dots. Click anywhere, then the same keys appear.
## Why v3 worked
Omarchy 3 locked with hyprlock, which reads raw Wayland keys. Quattro (7ea4e1ab Switch hyprlock to QS) replaced that with a Quickshell TextInput. Focus is set on Component.onCompleted, inputEnabled, and click. The lock stays mapped across suspend, so onCompleted never runs again. Click already called forcePasswordFocus(); nothing else did.
## Fix
While the lock is up, idle, and the field has no activeFocus, a short timer calls the same forcePasswordFocus() path as a click. Stops once the field is focused; does not run during PAM.
Verified: suspend → wake → type with no click.

### DIFF
--- a/shell/plugins/lock/LockView.qml
+++ b/shell/plugins/lock/LockView.qml
@@ -76,6 +76,16 @@ Item {
     if (inputEnabled) Qt.callLater(forcePasswordFocus)
   }
 
+  // Suspend leaves this surface mapped, so onCompleted never runs again.
+  // Resume also rebuilds the keymap and drops Qt activeFocus; a click already
+  // calls forcePasswordFocus, and this recovers the same path without one.
+  Timer {
+    interval: 100
+    repeat: true
+    running: root.inputEnabled && !root.authenticatingPassword && !passwordInput.activeFocus
+    onTriggered: root.forcePasswordFocus()
+  }
+
   // Measures the masked password at full size; passwordDotScale compares this
   // against the field width to decide how far the dots must shrink to fit.
   TextMetrics {


IgorKonovalov:
Tested the retry-timer approach from #7592 on the current stable release: it fixes the resume case here, and a focus trace shows where the focus is lost.

**Setup:** Omarchy 4.0.4-1, Hyprland 0.56.2-2, aquamarine 0.15.0-2, quickshell 0.3.1-1, kernel 7.2.5. ASUS ROG Zephyrus G15 (GA503QS), single internal eDP on the AMD iGPU (Cezanne, `amdgpu`), RTX 3080 idle via PRIME offload. Password-only lock (fingerprint reader unsupported). Suspend via lid close (s2idle).

**Symptom on stock 4.0.4:** after lid-open resume the password box is shown, but typing does nothing until the box is clicked. Reproduced 4 of 4 times.

**Patch tested:** in `LockView.qml`, a 250 ms repeating `Timer` that calls `forcePasswordFocus()` while `inputEnabled && !passwordInput.activeFocus` (same idea as #7592), plus logging of `passwordInput.activeFocus` and `Window.active`.

**Trace from one cycle (journal):**

```
08:19:47.631 lock-requested
08:19:48.133 field-focus=true
08:19:48.180 secure=true
08:19:48.181 window-active=true field-focus=true
08:19:48.338 kernel: PM: suspend entry (s2idle)
08:19:48.579 field-focus=false
08:19:48.579 window-active=false field-focus=false   <- libseat "Disabling seat"; keyboard removed
08:19:49.045 field-focus=true                        <- retry timer re-grabbed focus
08:19:58.666 kernel: PM: suspend exit
08:20:11.114 pam session started (password typed without clicking)
```

The lock window goes inactive when the seat is disabled for suspend, and `activeFocus` does not come back when it is re-enabled. `Window.active` never logged `true` again, so a handler on window activation alone would not have caught it. The retry timer did. Stock 4.0.4 has nothing that re-asserts focus after this point, which matches the diagnosis here and in #12435.

Hyprland's log for the same resume also shows `drm: Cannot commit when a page-flip is awaiting` twice, cleared by the next modeset right away. That is not what blocks input here.

Chessing234:
hey, this is already covered in https://github.com/omacom/omarchy/pull/12894 (lock keyboard focus after resume). mind closing this as a duplicate?

PR #7594, opened 2026-08-20
Prevent idle hard-freeze on 2011 MacBook Airs (intel_idle.max_cstate=1)

willhemming:
## What

Adds `install/hardware/apple/fix-sandy-bridge-idle.sh`, which appends `intel_idle.max_cstate=1` to the kernel cmdline (via the existing `/etc/limine-entry-tool.d/` drop-in pattern) on 2011 MacBook Airs (`MacBookAir4,1` / `MacBookAir4,2`).

## Why

The Sandy Bridge CPUs in these machines hard-lock when the CPU drops into deep C-states after load. The symptom is brutal to diagnose: the machine freezes completely (no TTY switch, no SysRq, nothing in the journal) within seconds-to-minutes of reaching an idle desktop, which reads as a graphics or compositor problem when it isn't one — it reproduces at a plain `multi-user.target` console too. This is a long-known quirk of these models under Linux, and capping intel_idle at C1 is the established fix.

## Verification

- Reproduced and fixed on a MacBookAir4,1 (i5-2467M) running Omarchy: froze reliably within moments of the desktop appearing; with `intel_idle.max_cstate=1` it has been stable since 2026-08-19, including load-then-idle cycles and lid-close suspend/resume.
- Being transparent about scope: verified on one 4,1 machine. The 4,2 (13" 2011 Air) shares the same board family and CPU generation and the same freeze is widely reported for it, but I don't own one to test. Happy to narrow the match to `MacBookAir4,1` if preferred.
- Detection tested: `omarchy-hw-match "MacBookAir4,"` matches both models and does not match other Macs (e.g. `Macmini6,1`).
- `./test/all`: the CLI suite passes; the shell suite shows 3 environment-dependent failures on my machine (config, snapper, unowned-system-paths) that are identical on a clean checkout of `quattro`, so nothing introduced by this change.

The param costs some idle battery, but on affected machines the alternative is a computer that locks up within a minute of logging in.

### DIFF
--- a/install/hardware/all.sh
+++ b/install/hardware/all.sh
@@ -36,6 +36,7 @@ run_logged "$OMARCHY_INSTALL/hardware/apple/fix-spi-keyboard.sh"
 run_logged "$OMARCHY_INSTALL/hardware/apple/fix-suspend-nvme.sh"
 run_logged "$OMARCHY_INSTALL/hardware/apple/fix-t2.sh"
 run_logged "$OMARCHY_INSTALL/hardware/apple/fix-brcmfmac-supplicant.sh"
+run_logged "$OMARCHY_INSTALL/hardware/apple/fix-sandy-bridge-idle.sh"
 
 run_logged "$OMARCHY_INSTALL/hardware/lenovo/fix-yoga-pro7-bass-speakers.sh"
 
--- a/install/hardware/apple/fix-sandy-bridge-idle.sh
+++ b/install/hardware/apple/fix-sandy-bridge-idle.sh
@@ -1,0 +1,15 @@
+# Idle-state freeze fix for 2011 MacBook Airs (MacBookAir4,1 / MacBookAir4,2).
+#
+# The Sandy Bridge CPUs in these machines hard-lock when dropping into deep
+# C-states after load: the whole machine freezes (no TTY switch, no SysRq,
+# nothing reaches the journal) within seconds-to-minutes of reaching an idle
+# desktop. Capping intel_idle at C1 prevents the lockup, at some idle power
+# cost on hardware that is otherwise unusable.
+
+if omarchy-hw-match "MacBookAir4,"; then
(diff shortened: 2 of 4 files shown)


willhemming:
Both review points addressed in 26e18f1:

- **Existing installs**: added `migrations/1787217354.sh` — model-gated on `MacBookAir4,`, installs the drop-in if hardware setup never did, rebuilds the boot image via `limine-mkinitcpio` with an interrupted-rebuild-safe machine-wide marker (following the `1786482992.sh` pattern), and sets `reboot-required` only when the booted kernel actually lacks the cap — someone who already added it by hand just gets the drop-in baked in without a reboot prompt.
- **Test coverage**: added `test/shell.d/sandy-bridge-idle-test.sh` modeled on `brcmfmac-supplicant-test.sh` — exercises both Air models, unaffected Macs (including a `MacBookAir41` lookalike that would pass a sloppier pattern), the generated command line, migration idempotency, the hand-fixed case, and the interrupted-rebuild retry. All assertions pass, and the migration framework tests (`migrate-scope`, `migrate-wrapper`) stay green.

willhemming:
Good catch on the per-user/machine-wide mismatch — fixed in d430461: the `reboot-required` state is now set whenever the booted command line lacks the cap, independent of whether this invocation performed the rebuild. Test updated to cover the second-user case (rebuild skipped, reboot still prompted) plus the fully-applied case where a later user's migration does nothing at all.

willhemming:
Fixed in 7df7b67: the migration now gates on the exact active line (`grep -Fxq`) instead of the file's existence, and reinstalls the dedicated drop-in whenever that line is missing — covering the truncated-by-interruption and commented-out cases. Since a bad drop-in now forces the rewrite-and-rebuild path regardless of the repair marker, the marker-gated branch can no longer bake an empty config either, which I believe covers the second location the comment pointed at. Tests added for both the truncated and commented-out drop-in.

Issue #6989, opened 2026-08-15
Quitting the speedtest midway leaves orphan processes


SystJF:
### System details

Quattro

### What's wrong?

1. Run a speed test
2. Quit before the "Run again" button pops up by clicking anywhere on the screen
3. "omarchy-network-speedtest" processes remain active

jfg96:
Confirmed on Omarchy 4.0.0-1.

I reproduced this twice:

1. Start the built-in network speed test.
2. Close the speed-test panel before it finishes.
3. The panel disappears, but the internet connection remains very slow.

After closing the panel, multiple `/bin/bash /usr/share/omarchy/bin/omarchy-network-speedtest down` processes remained active for several minutes and did not exit on their own. During the second reproduction, I observed seven such processes together with seven active `curl` processes downloading from `*.nflxvideo.net/speedtest`.

Normal performance returned only after manually stopping those processes.

After cleaning up, I left the system running without opening the speed test again. The processes did not return and the problem did not recur.

igor-gorohovsky:
Hit this too on 4.0.0.alpha (`omarchy 4.0.0-1`, `quickshell-git 0.3.0.r20.g28771c7-1`), launched from the UI panel. Seven orphaned `omarchy-network-speedtest down` workers survived for **31 minutes**, pulling a sustained **~300 Mbit/s** the whole time. They had been reparented to `systemd --user`.

Some root-cause detail that might help:

**The workers have no lifetime bound of their own.** In `/usr/bin/omarchy-network-speedtest`, `traffic_worker` is an unbounded loop:

```bash
while true; do
  url=${urls[$((idx % url_count))]}
  curl -fsS -o /dev/null "$url" 2>/dev/null || return
  idx=$((idx + 1))
done
```

There is no duration cap anywhere in the script — not on the worker loop, not on the individual `curl`, and not on the test overall. The run is terminated *only* by the consumer killing the parent, and cleanup happens *only* via `trap cleanup EXIT` in the parent. So the workers' entire lifetime depends on the parent dying gracefully.

**That makes any abnormal parent death leak all 8 workers, permanently.** Deterministic repro:

```console
$ omarchy-network-speedtest down >/dev/null 2>&1 &
$ P=$!; sleep 6
$ pgrep -P $P | wc -l
9
$ kill -9 $P          # parent dies without running its EXIT trap
$ sleep 3
$ pgrep -f nflxvideo | wc -l
10
```

[...]

PR #7598, opened 2026-08-20
Fix orphaned network speed test workers

MBemera:
Addresses #6989.

Each traffic worker now bounds itself rather than depending on the parent's EXIT trap, so no signal path can leave eight `curl` processes saturating the link. A worker checks the main shell between requests, caps every request with `--max-time`, and expires after 30s on its own. That cutoff reports exit 28 on any link too slow to finish a transfer inside the budget, so the loop continues past 28 and retires only on a genuine curl error — treating it as a failure would end the test one request in. `OMARCHY_SPEEDTEST_WORKER_LIFETIME` and `OMARCHY_SPEEDTEST_REQUEST_TIMEOUT` are there for the tests, and since the lifetime feeds an arithmetic expansion both are checked before use and can only tighten the defaults.

I ran the `kill -9` steps from the thread against `quattro` and against this branch: before, eight workers survive the parent and issue 110 requests in the next three seconds; after, none survive and none are issued. `phaseTimer` already stops each phase at 5s, so the panel is unaffected. This bounds the leak rather than removing it — a worker parked inside a transfer keeps going until `--max-time` fires, so the worst case is 30s of orphaned traffic instead of the 31 minutes reported. New `test/shell.d/network-speedtest-test.sh` covers the parent check, the request cap, lifetime expiry and both directions; it and `bin-style-test.sh` pass, and the rest of `test/shell` comes out the same as it does on `quattro` here.

Happy to adjust.

-MB

### DIFF
--- a/bin/omarchy-network-speedtest
+++ b/bin/omarchy-network-speedtest
@@ -10,6 +10,8 @@ direction="${1:-}"
 probe=1.1.1.1
 parallel=8
 url_count=3
+worker_lifetime=${OMARCHY_SPEEDTEST_WORKER_LIFETIME:-30}
+request_timeout=${OMARCHY_SPEEDTEST_REQUEST_TIMEOUT:-5}
 
 case "$direction" in
   down | up)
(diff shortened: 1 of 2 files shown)


iSparsh:
yes please lets get this merged! accidentally throttled a download by 50% today because i ran a few of these earlier.

sheikhlimon:
Thanks for tracking this down! It looks like there are multiple PRs open for the orphaned speed test workers. I recommended consolidating the fix over in #11054 since it leverages native process groups (`set -m`) to close a race condition in the cleanup trap. It might be best to close this as a duplicate once that approach lands.

MBemera:
No worries thanks for the update. Given the age of this PR and the rapid version updates I'd just assumed this was resolved already. -MB

PR #7600, opened 2026-08-20
Let the Dropbox widget be told the account's real quota

### LINKED ISSUE #7597: Dropbox widget shows wrong quota for Basic accounts with bonus space
### What happened

The Dropbox bar widget reports the wrong storage quota for Dropbox Basic accounts that carry bonus space.

My account shows **1.53 GB of 21.38 GB** on dropbox.com, but the widget shows **1.64 GB of 2 GB** — so it renders as ~82% full when the account is actually ~7% full.

### Cause

`shell/plugins/panels/dropbox/status.py` maps the plan name to a hardcoded quota:

```python
PLAN_QUOTAS = {
  "basic": 2_000_000_000,
  "plus": 2_000_000_000_000,
  ...
}
```

```python
quota = PLAN_QUOTAS.get(plan.lower(), 0)
```

`plan` comes from `subscription_type` in `~/.dropbox/info.json`, which is just `"Basic"`. But 2 GB is only the *starting* allocation for a Basic account. Referral bonuses (500 MB each, up to 16 GB, permanent) and older promos such as the 3 GB camera-upload bonus permanently raise the real quota. Mine adds up to 21.38 GB. Nothing in the plan name conveys this, so any long-standing Basic account is misreported.

The same applies in the other direction for Plus/Professional accounts with add-on storage.

### Steps to reproduce

1. Use a Dropbox Basic account that earned referral or promo bonus space (real quota > 2 GB).
2. Log in via the Dropbox widget and let it sync.
3. Open the widget panel.
4. It reads `of 2 GB` instead of the account's actual quota, and the usage bar is proportionally wrong.

### Expected

The quota shown should match the account's real allocation, or the widget should omit the denominator when it cannot determine it, rather than assert a value it hasn't verified.

### Notes on fixing

The local daemon doesn't expose the quota — `dropbox-cli` has no such command, and `~/.dropbox/info.json` carries only the plan name. The authoritative figure comes from the Dropbox API (`/2/users/get_space_usage`, `allocation.allocated`), which needs an OAuth token, so a fully automatic fix means adding a token setup flow.

Cheaper options, in rough order of effort:

- Add an optional quota-override setting to the widget manifest so affected users can correct it (this is what I did locally, as a cloned plugin — a `quotaOverrideMB` integer setting passed through to `status.py`).
- Fall back to `quotaKnown: false` for `basic` specifically, since it's the one plan whose stated quota is routinely wrong, and show usage without a denominator.

Happy to open a PR for the override-setting approach if that direction is welcome.

### Related detail

`usedBytes` is also computed locally by `scan_dropbox()` walking the sync folder rather than read from Dropbox. It's close enough when everything is synced (1.64 GB vs 1.53 GB here, the difference being `.dropbox.cache`), but it will understate usage for anyone using selective sync. Same root cause: no API access.

### System

- Omarchy 4.0.0-1
- Dropbox 264.4.3421-1, dropbox-cli 2024.04.17-2
- Account type: Dropbox Basic with referral/promo bonus space

srikat:
Fixes #7597.

`PLAN_QUOTAS` in the Dropbox widget's `status.py` maps `subscription_type` to the space a plan ships with, but that is only where an account *starts*. Referral bonuses on Basic (500 MB each, up to 16 GB, permanent) and storage add-ons on the paid tiers raise the real quota while `subscription_type` still reads `"Basic"`. My own account is 1.53 GB of 21.38 GB and the widget rendered it as `1.64 GB of 2 GB` — ~82% full instead of ~7%.

The authoritative number lives behind the Dropbox API (`/2/users/get_space_usage`), which needs an OAuth token the widget has no way to ask for, and neither `dropbox-cli` nor `~/.dropbox/info.json` exposes it. So rather than guess better, this lets the affected account state the figure:

- `manifest.json` gains a `quotaOverrideMB` integer setting, defaulting to `0`.
- `Service.qml` reads it and passes it to the helper.
- `status.py` gains `quota_override_bytes()`; the override wins when usable, and anything else — zero, negative, blank, non-numeric — falls back to `PLAN_QUOTAS` untouched.

Megabytes rather than bytes because a real quota in bytes overflows a QML `int`.

Default behavior is unchanged: with the setting absent or `0`, the plan lookup runs exactly as before.

## Before / after

Same account, before and after setting `quotaOverrideMB` to `21380`:

| | |
|---|---|
| Before | `1.64 GB of 2 GB` |
| After | `1.64 GB of 21.4 GB` |

(Screenshots attached below.)

## Tests

Extended `test/shell.d/dropbox-test.sh` with coverage for the override reader — megabyte conversion plus each way a setting can be unusable. `./test/shell` passes; the three failures in `./test/all` on my machine are all `omarchy-pkgs checkout not found` and unrelated to this change.

Verified in the running UI on Omarchy 4.0.0-1 via a cloned copy of the plugin.

## Notes

Two things I deliberately left out, happy to fold in either if you'd rather:

- **Manual docs.** `manual/24-commercial-apps-services.md` covers Dropbox setup but no widget settings, so I left it alone rather than start that section here.
- **`usedBytes` accuracy.** `scan_dropbox()` walks the sync folder rather than reading Dropbox, so it understates usage under selective sync (and slightly overstates here by counting `.dropbox.cache`). Same root cause — no API access — but a separate change.

### DIFF
--- a/shell/plugins/panels/dropbox/Service.qml
+++ b/shell/plugins/panels/dropbox/Service.qml
@@ -32,6 +32,10 @@ Item {
   property string lastError: ""
 
   readonly property int refreshIntervalSec: intSetting("refreshIntervalSec", 60, 10, 3600)
+  // 0 leaves the plan lookup in charge. Accounts carrying bonus space have a
+  // real quota the plan name cannot describe, so they set this to correct it.
+  // Megabytes, not bytes: a quota in bytes overflows a QML int.
+  readonly property int quotaOverrideMB: intSetting("quotaOverrideMB", 0, 0, 2000000000)
   readonly property bool busy: statusProcess.running || loginProcess.running || controlProcess.running
   readonly property string helperPath: (omarchyPath || "") + "/shell/plugins/panels/dropbox/status.py"
 
@@ -61,7 +65,7 @@ Item {
     _statusOutput = ""
     _statusError = ""
     refreshing = true
-    statusProcess.command = ["python3", helperPath, "25"]
+    statusProcess.command = ["python3", helperPath, "25", String(quotaOverrideMB)]
     statusProcess.running = true
   }
 
--- a/shell/plugins/panels/dropbox/manifest.json
+++ b/shell/plugins/panels/dropbox/manifest.json
@@ -19,7 +19,8 @@
     "allowMultiple": false,
     "defaultSection": "right",
     "defaults": {
-      "refreshIntervalSec": 60
+      "refreshIntervalSec": 60,
+      "quotaOverrideMB": 0
     },
     "schema": [
       {
@@ -30,6 +31,15 @@
         "max": 3600,
         "step": 5,
         "defaultValue": 60
+      },
+      {
+        "key": "quotaOverrideMB",
+        "type": "integer",
+        "label": "Storage quota override (MB, 0 for plan default)",
+        "min": 0,
+        "max": 2000000000,
+        "step": 100,
+        "defaultValue": 0
       }
     ]
   }
--- a/shell/plugins/panels/dropbox/status.py
+++ b/shell/plugins/panels/dropbox/status.py
@@ -7,6 +7,11 @@ import heapq
 from pathlib import Path
 
 
+# What a plan ships with, which is not always what an account has. Bonus space
+# — referral bonuses on Basic, storage add-ons on the paid tiers — raises the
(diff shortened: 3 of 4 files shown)


bjarneo:
Automated duplication check: this pull request looks similar to #11436, which covers the same Dropbox quota override. I keep that one open and close this one to consolidate review.

Issue #2979, opened 2025-10-29
Cedilla with keyboard US Intl


diegonogueira:
### What do you need?

Hello! How can I make the cedilla work using the ' + c keys? I already have the US International keyboard layout.

Some distributions, like Ubuntu and Zorin, work by default.

Is there a way to modify a file in the .config folder to make it work?

Thank you.

LuccasGianesini:
You can make it work by configuring the keyboard layout and variant in the `input.conf` file with this config:
```
  # Keychron
  device {
    name= keychron-keychron-k8-pro
    kb_layout= us
    kb_variant= altgr-intl
  }
```

To get the device name, you need to run `hyprctl devices` and find your keyboard. Keep in mind that the key combo is a little bit different, you need to press `alt gr` + `<` to get the cedilla char.

If you have only one keyboard, set the `kb_layout` and `kb_variant` directly in the `input` section

maxguzenski:
another way is to set the whole system to Portuguese, so `' + c` becomes **ç**.

execute:
`localectl set-locale LANG=pt_BR.UTF-8`

then run `sudo locale-gen`. You may need to restart the system.

To have both modes (with and without dead keys), edit your `hypr/input.conf` like this:

```
input {
  kb_layout = us,us
  kb_variant = basic,intl
  kb_options = compose:caps,grp:ctrl_space_toggle
```

then alter between keyboard layouts with CTRL + SPACE .

ps.: besides ctrl_space_toggle, you can use "alts_toggle" and "ctrls_toggle"

diegonogueira:
Thanks! It worked!

leomoreno11:
#

> another way is to set the whole system to Portuguese, so `' + c` becomes **ç**.
> 
> execute: `localectl set-locale LANG=pt_BR.UTF-8`
> 
> then run `sudo locale-gen`. You may need to restart the system.
> 
> To have both modes (with and without dead keys), edit your `hypr/input.conf` like this:
> 
> ```
> input {
>   kb_layout = us,us
>   kb_variant = basic,intl
>   kb_options = compose:caps,grp:ctrl_space_toggle
> ```
> 
> then alter between keyboard layouts with CTRL + SPACE .
> 
> ps.: besides ctrl_space_toggle, you can use "alts_toggle" and "ctrls_toggle"

Can this still be achieved without changing the language ? Some parts terminal installation prompts wore translated to portuguese. The Ç works with ' + C though.

[...]

PR #7601, opened 2026-08-20
Document the ç override on the US International layout

reinagliryc:
The system compose table maps `<dead_acute> <c>` to ć (Polish c-acute), so US International users get no cedilla from `'c`. [#2979](https://github.com/basecamp/omarchy/issues/2979) has drawn people rediscovering the same two-line `~/.XCompose` fix independently between Oct 2025 and Jul 2026.

I considered changing `default/xcompose` instead, but `'c` → ć is correct for Polish, Croatian, and Serbian users, and a static XCompose file can't be scoped to a layout — so documenting the override seemed better than picking a winner. Happy to send that version instead if you'd rather have it as a default.

Docs only, no behavior change.

### DIFF
--- a/manual/34-keyboard-mouse-trackpad.md
+++ b/manual/34-keyboard-mouse-trackpad.md
@@ -57,6 +57,17 @@ hl.gesture({ fingers = 3, direction = "horizontal", action = "workspace" })
 
 On Dell XPS laptops with a haptic touchpad, you can also set the click strength to low, mid, or high under _Trigger > Hardware > Touchpad Haptics_.
 
+### Typing ç on US International
+
+On the US International layout (`kb_variant = "intl"`), `'` followed by `c` gives ć rather than ç, because that's what the system's compose table maps it to. If you write French, Portuguese, or Catalan, override it in `~/.XCompose`:
+
+```
+<dead_acute> <c> : "ç"
+<dead_acute> <C> : "Ç"
+```
+
+Then run `omarchy-restart-xcompose` to pick up the change.
+
 ### Typing in Chinese, Japanese, and other languages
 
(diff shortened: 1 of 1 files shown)


dr-moreira:
The XCompose workaround is the right path when you want `'c` → ç while keeping `LANG=en_US` (or any locale that still uses the en_US compose table).

One caveat: that mapping is locale-specific, not a property of US International itself. `en_US.UTF-8/Compose` does map `<dead_acute> <c>` to **ć**, but Portuguese locales already override it in the system compose table:

```
# /usr/share/X11/locale/pt_BR.UTF-8/Compose
include "/usr/share/X11/locale/en_US.UTF-8/Compose"
<dead_acute> <C> : "Ç"
<dead_acute> <c> : "ç"
```

(`pt_PT.UTF-8` does the same.) Omarchy's default XCompose includes `%L`, so that locale file is what actually gets loaded.

I'm on the same layout this PR describes (`kb_layout = "us"`, `kb_variant = "intl"`) with `LANG=pt_BR.UTF-8`. My `~/.XCompose` has no cedilla rules — just the stock include of Omarchy's default table plus a couple of personal Multi_key macros — and `'c` already produces **ç**.

That locale-only path is already in #2979 (`localectl set-locale LANG=pt_BR.UTF-8`); it doesn't need a `~/.XCompose` edit.

Worth spelling both paths out in the manual:

1. Portuguese locale (`pt_BR` / `pt_PT`) — `'c` is already ç; no XCompose change.
2. Keep `LANG=en_US` (or any locale that still uses the en_US compose table, including `fr_FR` and `ca_ES`) — then the two-line `~/.XCompose` override in this PR is the right fix.

That also matches the reason not to bake `'c` → ç into `default/xcompose`: a static file can't be scoped per layout *or* per locale, and Polish/Croatian/Serbian users of the en_US table correctly want ć.

Roriz:
Confirmed: the ~/.XCompose fix works, 'c now gives ç.

There's also a way to get this without touching XCompose at all. Omarchy's ~/.XCompose includes %L, which pulls in the compose table for your locale. pt_BR.UTF-8, for example, already map 'c to ç. Setting LC_CTYPE=pt_BR.UTF-8 while keeping LANG=en_US.UTF-8 for English menus fixes it with no file edits.

This also matches what other systems do by default. Windows US-International maps 'c to ç https://dry.sailingissues.com/us-international-keyboard-layout.html

PR #7628, opened 2026-08-20
Fix bitwarden floating window config on other chromium browsers

Jacob-Roberts:
This config will make the floating window configuration work on other non-default chrome browsers.

Tested on all chromium based browsers (including ms edge)

## Before

screenshot-2026-08-20_12-

## After

(Screenshot tool intentionally redacted the contents, but it is working)

screenshot-2026-08-26_16-43-59

## Note

This only fixes it for chromium based browsers. The Firefox selector doesn't work because the class is just "firefox". Will leave as an opportunity for someone else to fix.


Jacob-Roberts:
@omarchybot re-review

bjarneo:
Automated duplication check: this pull request looks similar to #8341, which covers the same Bitwarden browser match. I keep that one open and close this one to consolidate review.

Issue #6671, opened 2026-08-10
USB autosuspend modprobe config is ignored (usbcore built into kernel)


mabster:
### System details

Omarchy 3.8.4 but also Quattro

### What's wrong?

(Full disclosure - an LLM wrote this issue based on my conversation history with it. Confirmed on the Discord that I'm not the only person seeing the behaviour.)

The current codebase includes a configuration file under /etc/modprobe.d/ to disable USB autosuspend. However, this file is entirely ignored by the system because usbcore is built statically into the kernel rather than compiled as a loadable module (m).

Modprobe files are only parsed for external modules, so this setting never applies to built-in kernel components.

USB autosuspend remains enabled by default. This can cause target USB devices (such as keyboards, mice, or storage drives) to sleep, lag, or drop connections unexpectedly during periods of inactivity.

**Suggested Fixes**

This can be resolved in one of two ways:

Kernel Command Line: 
Append usbcore.autosuspend=-1 to the bootloader/Limine kernel arguments.

Udev Rule:
Replace the modprobe file with a runtime udev rule (e.g., /etc/udev/rules.d/50-usb-autosuspend.rules) containing: 

`ACTION=="add",` SUBSYSTEM=="usb", ATTR{power/control}="on"

I'm not sure if this is a real issue, but it seems like the modprobe.d file was created in an attempt to disable USB autosuspend, so if that was the intent, it doesn't seem to be working.

dhh:
I don't think we can turn this on by default, as it would raise power consumption on laptops that need to be able to turn off USB. I haven't seen any of the fallout, and we haven't gotten any other reports. So think we'll have to keep this as something to document instead at the moment.

PR #7641, opened 2026-08-20
Opt-in: keep AMD xHCI controllers awake to avoid dead USB after resume

VECTORG99:
Opt-in udev rule behind `omarchy toggle amd-xhci-wakeup`; keeps AMD xHCI controllers awake so the broken suspend→resume path (Bugzilla 221073, Ubuntu 2158539) never runs; off by default honoring #6671 (dhh: default-on raises laptop power draw; he asked to document instead — docs side lives in #7631); class-based match (0x0c0330 + vendor 0x1022) avoids device-ID whack-a-mole; sibling of #7044 (same power/control=on + d3cold_allowed=0 for the Intel TB3 xHCI).

### DIFF
--- a/bin/omarchy-menu
+++ b/bin/omarchy-menu
@@ -211,7 +211,7 @@ show_share_menu() {
 }
 
 show_toggle_menu() {
-  local options="󱄄  Screensaver\n󰔎  Nightlight\n󱫖  Idle Lock\n󰂛  Notifications\n󰍜  Top Bar\n󱂬  Workspace Layout\n  Window Gaps\n  1-Window Ratio\n󰍹  Monitor Scaling\n  Direct Boot\n󰟵  Passwordless Sudo"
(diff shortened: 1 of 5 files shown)


VECTORG99:
Thank you for this incredibly detailed and instructive review! 

You are entirely correct on all counts. The fundamental flaw here was conflating runtime power management (`power/control="on"`) with system suspend callbacks. As you rightly pointed out, forcing the device awake at runtime does nothing to prevent the broken MSI/MSI-X redelivery path during a full system resume. The true fix is the kernel quirk (`xhci_hcd.quirks=0x40`) documented in #7631.

Additionally, the insights regarding the `quattro` branch architecture (moving system files to `etc/` and transitioning to a declarative JSONC menu) are highly appreciated and good to know for future reference.

Given that #7631 accurately identifies the root cause and provides the actual working fix, I am closing this PR to avoid shipping a placebo mechanism. We will let the documentation PR stand as the definitive solution for this bug.

Thanks again for taking the time to write such a comprehensive technical teardown!

PR #7644, opened 2026-08-20
fix: Bluetooth audio dropouts on Macs with Broadcom BT

coquer:
- Disable btusb autosuspend on Macs with Broadcom Bluetooth — autosuspend repeatedly power-cycles the controller between transmissions, causing periodic audio dropouts. Fixed via /etc/modprobe.d/btusb-autosuspend.conf (install/hardware/apple/fix-btusb-autosuspend.sh), wired into install/hardware/all.sh.

- Cap Bluetooth codec negotiation to aac, sbc_xq, sbc on the same hardware — AAC re-encoding on these Broadcom controllers is timing-sensitive and stutters under Wi-Fi/BT antenna contention (both radios share one antenna). AAC stays first in the list so working devices (e.g. AirPods) aren't downgraded; it's a fallback net, not a priority swap away from AAC. Shipped as a WirePlumber fragment (default/wireplumber/wireplumber.conf.d/bluez-prefer-sbc.conf), applied user-side via install/user/hardware/apple/fix-bluetooth-codec.sh, wired into install/user/all.sh.

- Both fixes are gated on the same Broadcom Wi-Fi/BT combo detection already used by fix-brcmfmac-supplicant.sh (T2 PCI ID or Apple DMI vendor + brcmfmac PCI ID list) — Macs only, no effect elsewhere.

- Added migrations (1787262778.sh, 1787263050.sh) so existing installs pick up both fixes, idempotent.

### DIFF
--- a/default/wireplumber/wireplumber.conf.d/bluez-codec-cap.conf
+++ b/default/wireplumber/wireplumber.conf.d/bluez-codec-cap.conf
@@ -1,0 +1,10 @@
+## bluez5.codecs is an ordered whitelist: it both restricts negotiation to the
+## codecs listed and sets their preference order. Broadcom Bluetooth
+## controllers (found in Intel and Apple Silicon-era MacBooks) stutter and
+## drop out worst on aptX/LDAC, especially under Wi-Fi/Bluetooth antenna
+## contention, so those are left out entirely. AAC stays first so devices
+## like AirPods that already work fine on it aren't downgraded to SBC.
+
+monitor.bluez.properties = {
+  bluez5.codecs = [ aac, sbc_xq, sbc ]
+}
--- a/install/hardware/all.sh
+++ b/install/hardware/all.sh
@@ -36,6 +36,7 @@ run_logged "$OMARCHY_INSTALL/hardware/apple/fix-spi-keyboard.sh"
 run_logged "$OMARCHY_INSTALL/hardware/apple/fix-suspend-nvme.sh"
 run_logged "$OMARCHY_INSTALL/hardware/apple/fix-t2.sh"
 run_logged "$OMARCHY_INSTALL/hardware/apple/fix-brcmfmac-supplicant.sh"
+run_logged "$OMARCHY_INSTALL/hardware/apple/fix-btusb-autosuspend.sh"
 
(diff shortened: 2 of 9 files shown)


coquer:
@omarchybot run again, I made changes based on your comments.

coquer:
@omarchybot Thanks for the re-review — addressed the direct-detection suggestion:

  Gate now checks Bluetooth transport directly, not by proxy. Both the btusb autosuspend fix and its migration now match the Bluetooth function's own
  PCI ID (14e4:5fa0|5f69|5f71|5f72, hci_bcm4377) instead of inferring "not on btusb" from T2 bridge presence. This fixes the false negative on Apple
  Silicon Macs (which have PCIe Bluetooth with no T2 bridge at all) and resolves the BCM4364/4464 split on its own merits per machine instead of
  guessing from bridge presence.

  Dropped 14e4:4425/4433 (Apple Silicon Wi-Fi IDs) from both inclusion lists. They're a different Bluetooth chip generation than this fix is about;
  for the codec cap this was the actual bug, since that fix has no transport check to catch them incidentally.

  Added 14e4:43a0/4331 (2013–2015 BCM4360/4331 MacBooks — the antenna-shared machines from the original report) to the codec-cap list. They already
  got the btusb quirk but not the codec cap, with nothing in the branch explaining the split; there wasn't a reason for it, so both fixes now cover
  that hardware class.

  Left unchanged, as you flagged as discretionary:
  - Predicate still duplicated across six spots — no shared omarchy-hw-* helper added.
  - No UART (hci_uart) detection — no evidence any machine in either PCI ID list uses that transport.
  - bluez5.codecs vs override.bluez5.codecs — inert on stock Omarchy today (nothing else sets the key), left as-is.
  - install/hardware/all.sh insertion point still collides with #7671 at the same anchor — not resolved, whoever merges second picks up both lines
    fine.
  - aptX/LDAC dropped repo-wide on affected Macs — still a maintainer call, not changed.

  Tests updated: added an Apple Silicon exclusion case to both suites, BCM4360/4331 inclusion cases to the codec-cap suite.
  test/shell.d/btusb-autosuspend-test.sh 13/13, test/shell.d/bluetooth-codec-test.sh 11/11, bash -n clean on all six touched files. Amended into the
  original commit (7cc34682) rather than stacking a fixup, since this is all one coherent gate-correctness fix.

PR #7649, opened 2026-08-21
Fix Codex usage collection on 0.149

### LINKED ISSUE #7648: Codex 0.149 breaks Agents panel limit display
### System details

### What's wrong?

```console
{
  "ready": true,
  "limitCount": 0,
  "authHelpText": "initialize"
}
```

```python
```

```console
```

### Expected behavior

```python
```

orienw:
## Changes

- replace Codex's removed `untrusted` approval policy with `on-request`
- retain the existing `read-only` app-server sandbox
- assert the collector's exact app-server launch arguments in its regression test

`on-request` is supported by Codex 0.148 and restores authenticated limits on 0.149.

Fixes #7648

## Tests

- `bash test/shell.d/agent-usage-codex-scanner-test.sh`
- `./test/cli`
- live `bin/omarchy-agent-usage-codex --limits-only` against Codex 0.149.0

### DIFF
--- a/bin/omarchy-agent-usage-codex
+++ b/bin/omarchy-agent-usage-codex
@@ -528,7 +528,7 @@ def fetch_codex_rpc():
 
   try:
     proc = subprocess.Popen(
-      [codex, "-s", "read-only", "-a", "untrusted", "app-server"],
+      [codex, "-s", "read-only", "-a", "on-request", "app-server"],
       stdin=subprocess.PIPE,
       stdout=subprocess.PIPE,
       stderr=subprocess.DEVNULL,
--- a/test/shell.d/agent-usage-codex-scanner-test.sh
+++ b/test/shell.d/agent-usage-codex-scanner-test.sh
@@ -14,7 +14,7 @@ cat >"$TEST_HOME/bin/codex" <<'EOF'
 #!/bin/bash
 
 if [[ -n ${CODEX_ARGS_FILE:-} ]]; then
-  printf '%s\n' "$*" >"$CODEX_ARGS_FILE"
+  printf '%s\0' "$@" >"$CODEX_ARGS_FILE"
 fi
 
 while read -r request; do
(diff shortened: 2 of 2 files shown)


RodriMora:
Confirmed this fixes the issue codex 0.149 introduced

p134c0d3:
Independent confirmation on a **ChatGPT Pro** account, which I don't think is covered by the reports so far (the earlier confirmation was on a different plan tier).

Environment: Omarchy 4.0.0-1, codex-cli 0.149.0, Linux 7.1.8-arch1-3.

Before the change, `~/.local/state/omarchy/agents/usage/codex.json` reads:

```json
"tierLabel": "",
"limits": [],
"authHelpText": "initialize"
```

The `"initialize"` value is worth calling out because it looks like a status string but is actually `str(TimeoutError("initialize"))` leaking out of `rpc_request` — codex exits on the argv before answering, `readline` hits EOF, and the `except Exception as exc` in `fetch_codex_rpc` puts the exception text into `authHelpText`. So the panel shows the method name where the auth hint belongs, which is what makes this one confusing to diagnose from the UI.

Running the packaged argv directly reproduces it immediately:

```
```

With only this PR's one-word change applied to the collector, the record populates correctly:

```json
"tierLabel": "pro",
"limits": [{"label": "Weekly (7-day)", "percent": <real value>, "resetsAt": "2026-08-27T13:13:29+00:00"}],
"usageStatusText": "",
"authHelpText": "Run `codex login` to authenticate."
```

So the Pro plan label resolves and the weekly window comes back with a real percentage and reset timestamp.

One note in support of the `on-request` choice: I also tested `-a never` and it returns identical results. Both are accepted by 0.149.0, and since `fetch_codex_rpc` only issues `initialize` / `account/read` / `account/rateLimits/read` and never executes a tool, the approval policy is never consulted on this path — the process is also already constrained by `-s read-only`. So the two are equivalent here in practice, and there's no behavioural reason to prefer `never` over what this PR already does.

diemol55:
Confirmed this regression is still present after updating to Omarchy 4.0.1-1 on Quattro.

Environment:

- Omarchy 4.0.1-1
- codex-cli 0.149.1

The packaged collector still launches Codex with:

```python
```

This causes the Agents panel to report:

```json
{
  "limits": [],
  "tierLabel": "",
  "authHelpText": "initialize"
}
```

Codex rejects the argument because `untrusted` is no longer supported:

```console
```

Changing only `untrusted` to `on-request` restores the Plus tier and weekly limits immediately. No login or credential changes were required.

I regenerated the record with:

```console
omarchy-agent-usage-update codex
```

and restarted the shell successfully.

PR #7651, opened 2026-08-21
Give session.slice CPU weight over app.slice so builds cannot stall input

tobi:
## Problem

When a few things compile at once — which happens by accident all the time now: a Qt/QML app, a Rust rewrite, a couple of agents in worktrees behind a terminal multiplexer — the mouse starts stuttering and keys arrive late or in bursts. It feels like a hardware problem. It isn't.

Hyprland, the shell, PipeWire, the portals and D-Bus run in **`session.slice`**. Everything the user launches — browsers, terminals, and every `cargo`/`ninja`/`cc1plus`/`rustc` those terminals spawn — runs in **`app.slice`**. Both slices ship at systemd's default `cpu.weight=100`. Under saturation the compositor's input loop is just another runnable thread waiting for a CFS slot, and libinput says so:

```
ERR from aquamarine ]: [libinput] event10 - Keychron Keychron M3 4K: client bug: event processing lagging behind by 364ms, your system is too slow
ERR from aquamarine ]: [libinput] event10 - Keychron Keychron M3 4K: client bug: event processing lagging behind by 168ms, your system is too slow
ERR from aquamarine ]: [libinput] client bug: timer button-debounce-debounce-event7: scheduled expiry is in the past (-104ms), your system is too slow
DEBUG from aquamarine ]: [libinput] event7  - Keychron Keychron M3 4K: SYN_DROPPED event - some input events have been lost.
```

Anyone can check their own session for this signature (no root needed):

```bash
command grep -hoE 'lagging behind by [0-9]+ms|SYN_DROPPED|scheduled expiry is in the past' /run/user/$UID/hypr/*/hyprland.log | sed -E 's/[0-9]+ms/Nms/' | sort | uniq -c
```

Binned, from a ~36 h Hyprland session on a Dell XPS 14 / Panther Lake driving a 5K display while doing normal dev work (Hyprland C++ build, cargo builds, 40 Chrome renderers):

```
    92  <50ms
    62  50-99ms
    41  100-249ms
    10  250-499ms
   max  385ms
    33  SYN_DROPPED
   125  scheduled expiry is in the past
```

Every 250 ms+ entry and every `SYN_DROPPED` lines up with a load spike of 15–30 on 16 cores from builds in terminals. Memory, swap and GPU were healthy at those moments — this is CPU arbitration, and it's happening in the one cgroup that must never lose it.

## Change

Ship `/usr/lib/systemd/user/session.slice.d/10-cpuweight.conf`:

```ini
[Slice]
CPUWeight=1000
```

That gives the interactive tier a 10:1 share over `app.slice` (`background.slice` is already 30, so the ordering becomes session ≫ app > background, which is what GNOME/KDE sessions effectively do). CPU weight is **work-conserving**: it only arbitrates under contention. An idle machine is unchanged, and builds still get every core the compositor is not using at that instant — which is almost all of them. The compositor just stops losing the argument. Memory and IO are untouched; this is orthogonal to the oomd tiering in `app.slice.d/10-oomd.conf` and follows the same packaging pattern.

Why not the alternatives:
- `Nice=` on Hyprland alone: per-thread, doesn't cover the shell/PipeWire/portals, and `sched_autogroup` flattens nice across sessions anyway.
- `SCHED_FIFO/RR` on the compositor: a compositor that ever spins becomes a hard input lock; Hyprland upstream deliberately elevates only specific threads.
- `nice -n 19` on builds: opt-in, forgotten, doesn't cover Chrome.

## Evidence (anecdotal, stated as such)

Applied live with `systemctl --user set-property session.slice CPUWeight=1000` during a bad episode (load 31, CPU PSI `some` 60 %, Hyprland + cargo builds running). Subjectively: immediate. Mouse stopped stuttering, typing stopped hanging, while the builds kept going. "Huh, this really fixed things."

A short synthetic A/B (48 burners in `app.slice`, real uinput mouse at 400 Hz, 60 s each) measured Hyprland's cgroup CPU stall (`cpu.pressure some`) at **+314 ms** with weight 1000 vs **+1150 ms** at the default 100, and +540 ms vs +764 ms with a malloc/memset churn load. Synthetic load did not reproduce the 300 ms+ libinput stalls that real compiles cause in either configuration, so I'm not claiming this eliminates them — only that the compositor waits for CPU several times less, and that in daily use the difference is obvious.

## Included

- `default/systemd/user/session.slice.d/10-cpuweight.conf` (commented)
- migration: `systemctl --user daemon-reload` so existing sessions pick it up live (same shape as the oomd migration; a TTY/SSH `omarchy update` gets it at next login)
- `test/shell.d/systemd-test.sh`: asserts the drop-in, that nothing else under `default/systemd/user` carries a `CPUWeight=` (app scopes must stay default so the compositor out-ranks them), and that the migration reloads
- `docs/file-layout.md` entry

Companion (after this lands): one `install -Dm644 default/systemd/user/session.slice.d/10-cpuweight.conf "$pkgdir/usr/lib/systemd/user/session.slice.d/10-cpuweight.conf"` in `omarchy-settings`' PKGBUILD in omarchy-pkgs, next to the oomd line. Happy to open that too.

Verified live: `cat /sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/session.slice/cpu.weight` → `1000`, `hyprctl configerrors` clean, Hyprland/Quickshell/PipeWire all still in `session.slice`, apps in `app-*.scope`.

### DIFF
--- a/default/systemd/user/session.slice.d/10-cpuweight.conf
+++ b/default/systemd/user/session.slice.d/10-cpuweight.conf
@@ -1,0 +1,24 @@
+# Keep the compositor responsive while apps saturate the CPU.
+#
+# Hyprland runs in session.slice, as wayland-wm@hyprland.desktop.service, next
+# to the shell, PipeWire, the portals and D-Bus. Everything the user launches
+# lands in app.slice/app-*.scope: browsers, terminals, and whatever those
+# terminals start -- cargo, make, ninja, qmlls, language servers, an agent or
+# three in worktrees. Both slices ship with cpu.weight=100, so when a couple
+# of builds pin every core the compositor's input loop waits its turn like any
+# compiler thread. libinput then logs "event processing lagging behind by
+# N ms, your system is too slow", the pointer stutters, and key events queue
+# up until the scheduler comes back around.
+#
+# Raising session.slice to 1000 gives the session a 10:1 share over app.slice
+# (background.slice is already 30). CPU weight only arbitrates under
+# contention, so an idle machine behaves exactly as before and builds still
+# get every core the compositor is not using at that instant -- which is
+# almost all of them. The compositor just stops losing the argument.
+#
+# Nice or realtime on Hyprland alone would not do this: nice is per thread and
+# autogroup flattens it across sessions, and SCHED_FIFO on a compositor that
+# ever spins is a hard input lock. The slice is where the whole interactive
+# tier already lives, so weight it as one.
+[Slice]
+CPUWeight=1000
--- a/docs/file-layout.md
+++ b/docs/file-layout.md
@@ -124,6 +124,7 @@ default/**                     ──►  omarchy-settings    /usr/share/omarchy
   ├─ applications/mimeapps.list                         /usr/share/applications/mimeapps.list
   ├─ systemd/user/*.service                             /usr/lib/systemd/user/
   ├─ systemd/user/app.slice.d/10-oomd.conf              /usr/lib/systemd/user/app.slice.d/
+  ├─ systemd/user/session.slice.d/10-cpuweight.conf     /usr/lib/systemd/user/session.slice.d/
   ├─ systemd/system-sleep/{force-igpu,
   │    keyboard-backlight,unmount-fuse}                 /usr/lib/systemd/system-sleep/
   ├─ systemd/zram-generator.conf.d/90-omarchy.conf      /usr/lib/systemd/zram-generator.conf.d/
--- a/migrations/1785424256.sh
+++ b/migrations/1785424256.sh
@@ -26,6 +26,6 @@ fi
 # Pick up /usr/lib/systemd/user/app.slice.d/10-oomd.conf without waiting for
 # the next login. That drop-in is what marks app.slice as a kill candidate;
 # until the user manager reloads and reports it, oomd is running with nothing
-# to act on. An `omarchy update` over SSH or from a TTY has no user manager to
-# reload, and there the next graphical login picks it up on its own.
+# to act on. There is one user manager per user, not per session, so this also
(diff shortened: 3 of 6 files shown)


tobi:
Addressed the packaging-order point:

- `migrations/1787276689.sh` now falls back to `systemctl --user set-property --runtime session.slice CPUWeight=1000` when `/usr/lib/systemd/user/session.slice.d/10-cpuweight.conf` is not installed yet. `--runtime` writes under `/run` only, so the session gets the weight immediately and the packaged drop-in takes over at the next login — no second migration needed, nothing persisted that could shadow the package later.
- `test/shell.d/config-test.sh` `package_defaults` now lists the drop-in, so the suite fails until the omarchy-settings PKGBUILD installs it (verified: passes against the companion branch, fails against current omarchy-pkgs master).
- `systemd-test.sh` asserts the runtime fallback.
- Companion packaging PR: omacom-io/omarchy-pkgs#183 (both `omarchy-settings` and `-dev`; must merge with the `_commit` bump that includes this, since `install -Dm644` fails on a tree without the file).

tobi:
Addressed the two wording points from the review in 255e881a (comments only, no behaviour change):

- `migrations/1787276689.sh` and `migrations/1785424256.sh` no longer claim an SSH/TTY update has no user manager to reload — it is one manager per user, so the reload reaches the graphical session either way.
- `migrations/1787276689.sh` now states what `--runtime` actually does: writes `/run/user/$UID/systemd/user.control/session.slice.d/50-CPUWeight.conf`, which outranks the packaged drop-in for the life of that user manager and is gone after a full logout. Both say 1000, so nothing differs until then.

`test/shell.d/systemd-test.sh` still green. Merge order unchanged: this and omacom-io/omarchy-pkgs#183 need to land together (stable half with the `_commit` bump that includes this).

Issue #7596, opened 2026-08-20
Quickshell unable to restart after latest update on Edge branch


agworkgit:
### System details

i5-10400F, RTX 2080 Ti, Omarchy 4.0.0.r1787.g9455496-1

### What's wrong?

quickshell: symbol lookup error: quickshell: undefined symbol: _ZN23QUntypedPropertyBindingC1EP23QPropertyBindingPrivate, version Qt_6_PRIVATE_API

1) Tried reboot
2) Re-tried Omarchy update and got 'Omarchy shell did not become ready after restart.'
3) Tried rebuilding quickshell-git, did not work

### Temporary Fix

1) Remove mise, was creating a conflict
2) Ran 'omarchy reinstall' to revert changes
3) Swapped to 'stable' update branch

morgoth:
I got the same issue: "omarchy-shell is not running" after last upgrade.
The reboot showed only the black screen.
I rebooted successfully with older snapshot

robouk:
I temporarily fixed it: `sudo pacman -S --ask 4 quickshell && omarchy restart shell`

Today's EDGE channel update bumped qt6 from 6.11.1 to 6.11.2. The shell is quickshell-git from the omarchy repo, built on 10 Aug against the old Qt. Quickshell links Qt's private ABI, so it needs a rebuild on every Qt release

Popidge:
This also affected the RC channel, but with the version mismatch reversed.
 
My official RC configuration used:
  
   - RC Arch mirror: qt6-base 6.11.1-1, qt6-declarative 6.11.1-3
   - Omarchy edge package repository: quickshell-git 0.3.0.r20.g28771c7-2
  
The -2 package’s .BUILDINFO shows it was built against Qt 6.11.2. Pacman nevertheless installed it alongside Qt 6.11.1, after which Quickshell failed with:

'undefined symbol: _ZN23QUntypedPropertyBindingC1EP23QPropertyBindingPrivate, version Qt_6'

It appears the Quickshell rebuild fixed the initial Edge mismatch but then reached RC machines before the RC Qt snapshot advanced. Switching to stable restored the compatible quickshell-git -1/Qt 6.11.1 combination immediately.

This looks like a package publication/channel-coordination issue rather than local configuration.

I was able to fix it with codex, as I still had terminal access through 'Super+Return' - codex looked into it and the easiest workaround was switching to stable with 'omarchy channel set stable'.

mouadse:
[...]

PR #7655, opened 2026-08-21
Rebuild quickshell from source when Qt ABI breaks after update

suryanshsrivastava:
## Summary

When `qt6-base` is upgraded, the prebuilt `quickshell-git` in the omarchy repo fails to load (private Qt symbols change). This adds an automatic rebuild step so the system self-heals instead of leaving a dark desktop.

## Root cause

The omarchy repo ships a prebuilt `quickshell-git` binary that links Qt private ABI symbols. When `qt6-base` is upgraded, those symbols change and quickshell fails with `symbol lookup error: QUntypedPropertyBinding, version Qt_6`. The omarchy rebuild of quickshell-git may ship before the channels Qt snapshot advances, or lag behind it, causing a mismatch.

## Fix

**`bin/omarchy-update-fix-quickshell-abi`** (new) — runs after system packages are updated and before the shell restart. If `quickshell --version` fails (ABI break detected), rebuilds quickshell from AUR source against the installed Qt using `yay --aur --rebuild --redownload`. Falls back to printing the manual downgrade command if yay is unavailable or the build fails.

**`bin/omarchy-update`** — calls the new script after `omarchy-update-system-pkgs`.

## Testing

- `./test/cli` passes
- `./test/shell` failures are all pre-existing (missing `omarchy-pkgs` checkout + the quickshell ABI bug itself)

## Ref

- Issue: #7596

### DIFF
--- a/bin/omarchy-update
+++ b/bin/omarchy-update
@@ -45,6 +45,7 @@ if [[ ${1:-} == "-y" ]] || omarchy-update-confirm; then
   # them, so everything below waits on this finishing. An upgrade that stopped
   # takes the update with it rather than migrating against what is still on disk.
   omarchy-update-system-pkgs
+  omarchy-update-fix-quickshell-abi
   omarchy-migrate
   omarchy-hook post-update
   omarchy-update-aur-pkgs
--- a/bin/omarchy-update-fix-quickshell-abi
+++ b/bin/omarchy-update-fix-quickshell-abi
@@ -1,0 +1,38 @@
+#!/bin/bash
+
+# omarchy:summary=Rebuild quickshell from source if it broke after a Qt upgrade
+
+# The prebuilt quickshell-git in the omarchy repo links Qt private symbols.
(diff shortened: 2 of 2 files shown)


suryanshsrivastava:
Closing — this approach doesn't work. As @mouadse demonstrated, yay picks the omarchy repo prebuilt even with `--aur --rebuild --redownload`, reinstalling the broken binary.\n\nThe actual fix belongs in the package repository: either pin `quickshell-git` to match the channel's Qt version, or publish the Qt bump and quickshell rebuild atomically per channel.\n\nLeaving the root-cause analysis here for whoever fixes the packaging:\n- `-2` links `QUntypedPropertyBinding(P23QPropertyBindingPrivate)` as `(Qt_6)`\n- Qt 6.11.1 exports it only as `@@Qt_6_PRIVATE_API`\n- `-1` links it as `(Qt_6_PRIVATE_API)` → loads fine\n- Both builds share the same 1572 `Qt_6`-versioned symbols; this ctor's version tag is the only difference

PR #7664, opened 2026-08-21
Feat: Bluetooth device aliases 

mrlorentx:
## Summary

Omarchy 4 has no Bluetooth GUI outside this panel, but the panel never showed BlueZ's per-device `Alias`: `deviceLabel` preferred the device-reported `Name`, meaning that any aliases set would be ignored in Omarchy.

This PR adds alias display as a fix, and support to write aliases via the BT panel.

## Changes

- `Model.js`: `deviceLabel` prefers the alias. `hasHumanName` now accepts either name, so a
  MAC-shaped alias can't hide a row from the only panel that could change it back.
- The PipeWire sink matcher tries both names, so renaming a speaker doesn't cost it the
  automatic output switch.
- `Panel.qml`: inline `TextField` in place of the label, draft owned by the panel, key
  catcher blocked while open, empty submit drops the alias.
- Replaces the row's `actionFocused` boolean with a `rowFocus` enum — one action slot could
  only ever reach the forget button, so `h`/`l` now walk row → pencil → forget.

## Tests

- 14 new assertions in `test/shell.d/bluetooth-test.sh` (71 total); `./test/shell` green.
- Checked against live BlueZ: renaming a *connected* device leaves PipeWire's node
  properties on the old name, which is why the matcher needs both — a flip-only build fails
  that case.
- Source-pattern assertions cover only decisions that fail silently; anything that visibly
  breaks on the first keypress is left to the eye.

https://github.com/user-attachments/assets/2eb90ed7-f0aa-44a8-952c-4f99d2c702ee

### DIFF
--- a/manual/05-the-top-bar.md
+++ b/manual/05-the-top-bar.md
@@ -51,7 +51,7 @@ The panels aren't read-outs. They're where you actually do the thing:
 
 - **Audio** has a master volume slider, an output-device picker, and a per-app mixer, so you can turn down that one browser tab without touching everything else.
 - **Network** scans for Wi-Fi, shows signal strength, connects, and lets you pick a DNS provider.
-- **Bluetooth** lists your devices with connect/disconnect and battery levels.
+- **Bluetooth** lists your devices with connect/disconnect and battery levels, and lets you rename them: press `r` on a remembered device (or click the pencil) to give it a name you'll actually recognise, and submit an empty field to put the manufacturer's name back.
 - **Power** shows battery stats, switches power profiles (it remembers a separate choice for battery and AC), and prints some system info.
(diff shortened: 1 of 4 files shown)


mrlorentx:
Good catch! It's actually worse than the example suggests: Pro on an AirPods Pro — a natural short alias for that exact device — routed audio to pro-output-3, an HDMI output. Car matches via alsa_card, a substring nobody would anticipate. 

Agreed the alias should never have been a candidate. Pre-PR the fallback only ever matched the device-reported name, so putting the alias in there was a change to sink-matching we shouldn't be doing here anyway.

It also costs nothing to drop: any node that could carry the alias is a `bluez_output` node, which carries the MAC the address criterion already matches. So the fallback now prefers the reported name, and consults the alias only for a device that reports no name at all.

bjarneo:
Automated duplication check: this pull request looks similar to #7837, which covers the same Bluetooth alias rename. I keep that one open and close this one to consolidate review.

Issue #7483, opened 2026-08-19
ELAN 04f3:0c4b fingerprint reader never matches under stock libfprint (ThinkPad E14/E15 Gen 3/4) — works with libfprint-tod instead


jcperdomoybarra:
## What happened

`omarchy-setup-security-fingerprint` installs stock `libfprint` + `fprintd`
and drives enrollment/verification through the open-source `elan.c` driver.
For the ELAN `04f3:0c4b` fingerprint reader (ThinkPad E14/E15 Gen 3 and
Gen 4, at least), that driver treats the sensor as a swipe sensor and
matches with NBIS/Bozorth3 entirely in software:

```
fprintd-enroll   # completes fine, all stages pass
fprintd-verify   # NEVER matches — Bozorth3 score is exactly 0/24,
                 # every single attempt, regardless of finger, technique,
                 # orientation, or speed
```

25+ verify attempts across multiple sessions, always a literal `0` score
(never even a partial match). The script itself doesn't fail loudly —
enrollment and the first verify both "succeed" from the script's point of
view (`fprintd-verify` just needs one non-zero exit at setup time, and a
lucky first touch can pass), so PAM gets wired up for a reader that then
never reliably authenticates in daily use.

## Why this is a packaging/hardware-detection problem, not a bug in the script's logic

This sensor isn't actually a swipe sensor in the way the open driver
assumes — Lenovo's own Windows driver, and a proprietary Linux
`libfprint-tod` plugin Lenovo ships for it, both drive it as a
**press/area** sensor instead:

```
Broken on this PID:  04f3:0c4b → elan.c (open) → swipe capture → NBIS/Bozorth3 → 0/24
Working on this PID: 04f3:0c4b → libfprint-tod → Lenovo's proprietary driver → "press"
```

Switching drivers (not patching `elan.c`) is what actually fixes it —
confirmed with 10/10 `fprintd-verify` passes after the switch, and
working `sudo`, polkit, and lock-screen auth via the existing Omarchy PAM
setup (no PAM changes needed — `pam_fprintd.so` talks to `fprintd` over
D-Bus regardless of which driver is loaded underneath).

I wrote up the full fix — root cause, step-by-step instructions, a
security note on the proprietary binary involved (verified byte-for-byte
against Lenovo's own download before trusting it), and how to protect the
non-standard packages from being silently touched by `omarchy update` —
here, in case it's useful for anyone else with this hardware:

[...]

PR #7667, opened 2026-08-21
Hint at the ELAN 04f3:0c4b generic-driver matching bug in fingerprint setup

jcperdomoybarra:
## What

Hardware: ELAN `04f3:0c4b` fingerprint reader (confirmed on a ThinkPad E15 Gen3 AMD, likely present on other E14/E15 Gen3/Gen4 units — see #7483).

Under the generic `libfprint` `elan.c` driver, this reader enrolls fine (`fprintd-enroll` completes normally) but `fprintd-verify` never produces a working match — the score is consistently `0` regardless of technique, across 25+ verify attempts in my own testing. `omarchy-setup-security-fingerprint` currently reports this as a generic "Verification failed. You may want to try enrolling again." with no way to tell a driver limitation from a bad enrollment.

Lenovo ships a proprietary "ELAN MOH" driver for this exact PID, usable on Linux via `libfprint-tod` + the `libfprint-2-tod1-elan` AUR package. Switching to it fixed the reader completely (10/10 clean `verify-match`, `sudo`/polkit/lock screen all working) — full writeup, evidence, and a reproducible procedure at https://github.com/jcperdomoybarra/thinkpad-elan-04f3-0c4b-fingerprint-fix. I filed #7483 with this same summary a couple of days ago; this PR is the actual code half of that.

## Why this shape

This does **not** bundle or install the proprietary driver — it's a closed-source binary from Lenovo, not something to pull into Omarchy directly. It only adds:

1. `bin/omarchy-hw-fingerprint-generic-driver-risk` — a small helper (mirrors `omarchy-hw-fingerprint`'s shape and `OMARCHY_USB_DEVICES_PATH` override for testability) that detects a known-risk reader by USB id and prints it.
2. A one-line hook into `omarchy-setup-security-fingerprint`'s existing verify-failure branch, pointing at the workaround instead of just "try again" — only shown for the exact confirmed-affected id, not a general driver-quality guess.

Kept the known-id list to just `04f3:0c4b` (the one I've actually verified end to end) rather than guessing at siblings — easy to extend later if others confirm the same failure mode on a different PID.

## Testing

- `test/shell.d/hw-fingerprint-generic-driver-risk-test.sh` (new), same fixture pattern as `hw-fingerprint-test.sh`.
- Ran the full `test/shell` suite locally — all passing, including the pre-existing fingerprint tests.
- Verified the actual failure/workaround on real hardware (the ThinkPad in #7483).

### DIFF
--- a/bin/omarchy-hw-fingerprint-generic-driver-risk
+++ b/bin/omarchy-hw-fingerprint-generic-driver-risk
@@ -1,0 +1,26 @@
+#!/bin/bash
+
+# omarchy:summary=Prints the connected reader's USB id and exits 0 if it is known to fail matching under the generic driver
+# omarchy:hidden=true
+
+# A handful of Elan match-on-chip-shaped readers enroll fine under the
+# generic libfprint elan.c driver but never produce a working software
+# match — fprintd-verify fails every time, with no indication that the
+# driver is the problem rather than technique. Confirmed on 04f3:0c4b
+# (ThinkPad E15 Gen3 AMD): consistently 0-score matches across 25+ verify
+# attempts under elan.c, resolved by switching to Lenovo's proprietary ELAN
+# MOH driver via libfprint-tod instead — see
+# https://github.com/jcperdomoybarra/thinkpad-elan-04f3-0c4b-fingerprint-fix
+known_generic_matching_failure_ids=" 04f3:0c4b "
+usb_devices_path="${OMARCHY_USB_DEVICES_PATH:-/sys/bus/usb/devices}"
+
+for dev in "$usb_devices_path"/*; do
+  [[ -r $dev/idVendor && -r $dev/idProduct ]] || continue
+  id="$(<"$dev/idVendor"):$(<"$dev/idProduct")"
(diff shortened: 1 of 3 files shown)


Bwest8:
Additional hardware data point for this PID: IdeaPad 3 14ALC6 (`LENOVO 82KT`, same `04f3:0c4b`) on Omarchy 4.0.2. Full writeup: https://github.com/omacom/omarchy/issues/7483#issuecomment-5479877389

On that chassis stock `elan.c` never reaches verify. Enroll fails immediately:

```
Enroll result: enroll-disconnected
```

with fprintd logging a protocol error during identify-for-enroll. The hint added here is only on the verify-failure branch, so this model still gets `Enrollment failed. Please try again.` and no pointer at the TOD workaround.

Once `libfprint-tod` + `libfprint-2-tod1-elan` + `openssl-1.1` were installed, the device showed `scan-type: press` as **ELAN Fingerprint Sensor**, enroll + verify-match succeeded, and Omarchy PAM worked.

Please also emit the `04f3:0c4b` hint (or offer the TOD install, same shape as #7993) on the **enroll-failure** path, not only after a failed verify.

PR #7688, opened 2026-08-21
Dismiss low battery warning when AC connects

mkdustan:
Dismiss the low-battery popup when UPower switches to AC.

Add a `--dismiss` mode to `omarchy-battery-low` and retry the dismissal after an in-flight warning exits if AC connects before the popup is inserted.

## Testing

- Manually verified at 10% that connecting AC dismisses the warning
- `./test/all`


bjarneo:
Automated duplication check: this pull request looks similar to #11415, which covers the same low-battery warning dismissal on AC. I keep that one open and close this one to consolidate review.

PR #7701, opened 2026-08-21
feat(agents): add Antigravity usage collector and panel asset

senhor-f:
# PR Description
### Motivation & Context
Following up on the initial Antigravity CLI integration in #6900, this PR completes the integration on the Omarchy desktop shell by adding live usage tracking, real-time rate limit monitoring, and the official vector brand asset for the **Antigravity** (`agy`) coding agent in the agents panel (`omarchy.agents`).

---

### Key Changes

1. **Antigravity Usage Collector (`bin/omarchy-agent-usage-antigravity`)**:
   - Implements local transcript analysis across `~/.gemini/antigravity-cli/brain/` and fallback `history.jsonl` to calculate prompt counts, active sessions, and per-model token consumption (`gemini-3.7-flash`, `gemini-3.7-pro`, `claude-sonnet-4-6`, `claude-opus-4-6`, `gpt-oss-120b`, etc.).
   - Probes live rate limits using `agy -p /usage --output-format json` to extract authoritative 5-hour session and 7-day weekly quota percentages and reset countdowns for Gemini and 3rd-party models.
   - Provides graceful fallbacks when offline, missing CLI binaries, or during initial login states.
   - Integrates with `omarchy-agent-usage-update` to automatically emit `~/.local/state/omarchy/agents/usage/antigravity.json`.

2. **Brand Assets (`shell/plugins/agents/assets/`)**:
   - Adds the official Google Antigravity vector mark (`antigravity.svg` and `antigravity-light.svg`).
   - Crafted specifically for Qt Quick's `QSvgRenderer` (SVG 1.2 Tiny compliance without unsupported `<clipPath>` definitions) with the signature Google multi-color gradient mesh.

3. **Manifest & Documentation (`shell/plugins/agents/manifest.json`, `README.md`)**:
   - Enables `antigravity` in the default providers table and updates the collector documentation.

4. **Automated Test Suite (`test/shell.d/agent-usage-antigravity-scanner-test.sh`)**:
   - Adds 12 automated unit and regression tests covering:
     - Transcript token counting & model classification.
     - Live rate limit probing with mock `agy` CLI output.
     - Fallbacks to `history.jsonl` and synthetic quota buckets when the CLI is absent.

---

### Verification
- **Automated Tests**:
  - Ran `./test/shell.d/agent-usage-antigravity-scanner-test.sh` (12/12 passing).
  - Ran `./test/shell.d/agent-usage-update-test.sh` (7/7 passing).
  - Ran `./test/cli` (full CLI metadata & dispatch suite passing).
- **Visual Verification**:
  - Triggered `omarchy-shell omarchy.agents open` in the live Hyprland session.
  - Verified hero rendering, tab switching, live quota meters, 7-day token chart, and per-model breakdown with `omarchy capture screenshot fullscreen save`.

### DIFF
--- a/bin/omarchy-agent-usage-antigravity
+++ b/bin/omarchy-agent-usage-antigravity
@@ -1,0 +1,513 @@
+#!/usr/bin/python3
+# omarchy:summary=Print the Antigravity usage record as JSON
+# omarchy:args=[--force] [--limits-only]
+# omarchy:hidden=true
+"""Collect Antigravity usage into one display-ready JSON record.
+
+Everything the agents panel shows for Antigravity comes from this one
+command: local transcript stats from ~/.gemini/antigravity-cli/brain, history.jsonl,
+and the live rate limits from the Antigravity CLI (/usage command). The panel
+itself only ever reads the JSON this prints; it never talks to disk formats or endpoints.
+"""
+
+from __future__ import annotations
+
+import argparse
+import datetime as dt
+import fcntl
+import hashlib
+import json
+import os
+import shutil
+import subprocess
+import sys
+import tempfile
+import time
+from pathlib import Path
+from typing import Any
+
+AGENT_ID = "antigravity"
+AGENT_NAME = "Antigravity"
+AUTH_HELP = "Run `agy` in your terminal to authenticate."
+CACHE_TTL_SECONDS = 300.0
+LIMITS_CACHE_TTL_SECONDS = 90.0
+PROBE_TIMEOUT_SECONDS = 60.0
+
+
+def expand_path(value: str) -> Path:
+  return Path(os.path.expandvars(os.path.expanduser(value))).resolve()
+
+
+def get_gemini_dir() -> Path:
+  if "GEMINI_CLI_HOME" in os.environ:
+    return expand_path(os.environ["GEMINI_CLI_HOME"])
+  return expand_path("~/.gemini/antigravity-cli")
+
+
+def get_state_file() -> Path:
+  state_home = os.environ.get("XDG_STATE_HOME")
+  if state_home:
+    return (expand_path(state_home) / "omarchy" / "agents" / "usage" / "antigravity.json").resolve()
+  return expand_path("~/.local/state/omarchy/agents/usage/antigravity.json")
+
+
+def cache_root() -> Path:
+  root = expand_path(os.environ.get("XDG_CACHE_HOME", "~/.cache")) / "omarchy" / "agent-usage"
+  root.mkdir(parents=True, exist_ok=True)
+  return root
+
+
+def date_string(value: dt.date) -> str:
+  return value.strftime("%Y-%m-%d")
+
+
+def recent_date_strings() -> list[str]:
+  today = dt.datetime.now().date()
+  return [date_string(today - dt.timedelta(days=offset)) for offset in range(6, -1, -1)]
+
+
+def local_date_string() -> str:
+  return date_string(dt.datetime.now().date())
+
+
+def empty_bucket() -> dict[str, int]:
+  return {
+    "inputTokens": 0,
+    "outputTokens": 0,
+    "cacheReadInputTokens": 0,
+    "cacheCreationInputTokens": 0,
+  }
+
+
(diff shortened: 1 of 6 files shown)


data-goblin:
Tested on 2026-08-23 with Antigravity CLI 1.1.19 and it works for me :)

I authenticated through Google OAuth on Omarchy and verified that `/usage` returns both Gemini and Claude/GPT quota groups, including their five-hour and weekly windows. The collector populated all four limits and the agents panel rendered them correctly.

One follow-up I found while testing is that AGY exposes two quota pools: Gemini, plus Claude/GPT-OSS. Google documents these as separate Antigravity reasoning-model groups with separate five-hour and weekly quotas: [Antigravity model documentation](https://antigravity.google/docs/models).

Displaying all four meters is accurate, but `Claude & GPT` inside the Antigravity tab is easy to mistake for usage from the separate Claude Code and Codex providers. It also makes Antigravity the only tab that does not use the panel's consistent Session/Weekly shape. For users who only use Gemini through AGY, the extra pair takes up half the Limits section without being useful.

Before:

Antigravity showing four model-pool limits

The follow-up therefore shows one `Session` and one `Weekly` limit by default. Users who select Claude or GPT-OSS through AGY can opt into all four meters with:

```bash
omarchy bar set omarchy.agents showAntigravityThirdPartyLimits true --json
```

After:

Antigravity showing the focused Session and Weekly limits

The same commit removes synthetic quota and tier data when live limits are unavailable, derives readiness from real data, names the optional pool `Claude / GPT-OSS`, and prevents background quota probes from triggering AGY self-updates.

Verification passed: 17 collector assertions, 11 panel assertions, 7 updater assertions, QML lint, and the full CLI suite.

Follow-up PR against this PR's source branch: https://github.com/senhor-f/omarchy/pull/1

belspectre (commented):
Great work — this is by far the most complete Antigravity integration PR. I've been running
a local version of the same collector and hit a few of the same edges. A few notes that
might be worth addressing before merge:

---

1. `remaining_fraction` absent key — silent wrong data

```python
remaining = float(b.get("remaining_fraction", 1.0))
```

`1.0` as default means if `agy` omits the field (e.g. during an auth error), every bucket
silently renders as 0% used. Better to skip the bucket entirely:

```python
remaining = b.get("remaining_fraction")
if remaining is None:
    continue
remaining = float(remaining)
```

---

2. Offline fallback fabricates reset times from `now`

The fallback in `collect_limits()` emits `resetsAt: now + timedelta(hours=5)`, which will
render a countdown in the panel from the moment of the failed probe — not the real reset time.
Since no real quota data is known offline, returning `[]` is cleaner; the panel already hides
the LIMITS section when the list is empty.

---

3. Small lookup table makes future window types safe

The `bucket_sort_key` / title derivation checks for `"5h"` and `"weekly"` inline. When Google
adds new window types the fallback lands on the full bucket name as the title
(`"Five Hour Limit Remaining"` instead of `"Session"`). A small dict makes it forward-safe:

```python
_WINDOW_TITLE = {"weekly": "Weekly", "5h": "Session", "daily": "Daily", "monthly": "Monthly"}
title_suffix = _WINDOW_TITLE.get(window.lower(), window or "Limit")
title = f"{prefix} {title_suffix}"
```

---

4. Unused `import re` (minor lint)

`re` is imported but never used — will trigger ruff/flake8 in CI.

ChaseC-130:
Tested this implementation locally on Omarchy (`quattro` branch) with `agy` (Google Antigravity CLI).

The collector cleanly captures:
- Live quota pools via `agy -p "/usage" --output-format json` (both Gemini Models and Claude/GPT 5-hour session and 7-day weekly limits with exact reset timestamps)
- Daily prompt and session stats from `~/.gemini/antigravity-cli/history.jsonl`

It renders seamlessly in the native `omarchy.agents` bar widget alongside Claude Code and Codex. +1 for merging this upstream so Antigravity has first-class status across all Omarchy installs!

bjarneo:
Automated duplication check: this pull request looks similar to #8479, which covers the same Antigravity usage collector. I keep that one open and close this one to consolidate review.

PR #7713, opened 2026-08-21
Include dual-width fonts in font picker

### LINKED ISSUE #7698: Font picker excludes FC_DUAL monospace fonts (spacing=90)
### System details

AMD Ryzen 7 9800X3D, AMD Radeon RX 9070-series, Omarchy 4.0.0-1

### What's wrong?

Omarchy's font picker recognizes only fonts classified by Fontconfig as `FC_MONO` (`spacing=100`). It excludes fonts classified as `FC_DUAL` (`spacing=90`), even though that is also a fixed-grid spacing category: glyphs have precisely two advance widths, one exactly twice the other. Such fonts can represent one-cell and two-cell glyphs and remain compatible with terminal grids.

[Mona Sans Mono](https://github.com/github/mona-sans#mona-sans-mono-styles) is one reproducible example. It is installed, matched by Fontconfig, and can be applied with `omarchy font set`, but it is absent from `omarchy font list` and the Style > Font picker:

```console
$ fc-match "Mona Sans Mono"
MonaSansMono-Regular.ttf: "Mona Sans Mono" "Regular"

$ fc-query -f 'family=%{family[0]} spacing=%{spacing}\n' \
    ~/.local/share/fonts/Mona\ Sans\ Mono/MonaSansMono-Regular.ttf
family=Mona Sans Mono spacing=90

$ omarchy font list | grep -F "Mona Sans Mono"
# no output

$ omarchy font set "Mona Sans Mono"
# succeeds
```

The exclusion comes from the strict filter in `omarchy-font-list`:

```bash
fc-list :spacing=100 -f "%{family[0]}\n" | grep -v -i -E 'emoji|signwriting|omarchy' | sort -u
```

Fontconfig documents `FC_DUAL` as distinct from proportional spacing: [Fontconfig developer documentation](https://fontconfig.pages.freedesktop.org/fontconfig/fontconfig-devel/).

Is strict `FC_MONO` an intentional compatibility requirement for an Omarchy component? If not, the font picker should also discover compatible `FC_DUAL` fonts. If dual-width fonts are intentionally unsupported, it may be better for `omarchy font set` to validate that consistently or for the limitation to be documented.

OldJobobo:
## Problem

`omarchy font list` only queries Fontconfig's `FC_MONO` class (`spacing=100`). That excludes `FC_DUAL` fonts (`spacing=90`), even though they use a fixed terminal-compatible grid with one-cell and two-cell glyph widths.

As a result, installed families such as `Lekton Nerd Font` can be applied with `omarchy font set` but are missing from both `omarchy font list` and the Style → Font picker.

Fixes #7698.

## Change

- query both `FC_DUAL` and `FC_MONO` font families
- merge the results through the existing exclusion, sorting, and deduplication pipeline
- add focused coverage for discovery, filtering, ordering, and deduplication across both spacing classes

## Visual verification

Searching the font picker for the installed dual-width `Lekton Nerd Font` family:

Before: only the FC_MONO variant is listed. After: FC_DUAL and FC_MONO variants are both listed.

Captured from the same shell instance and geometry. Only the font-list implementation changed between captures.

## Validation

- `bash test/shell.d/font-list-test.sh`
- `bash -n bin/omarchy-font-list test/shell.d/font-list-test.sh`
- `shellcheck -e SC1091 bin/omarchy-font-list test/shell.d/font-list-test.sh`
- `./test/cli`

<details>
<summary>Full shell-suite result</summary>

`./test/shell` ran the new test successfully. The aggregate remained red on five unrelated environment/runtime checks: three require a sibling `omarchy-pkgs` checkout, one migration timing case was intermittent, and the live graphical session caused duplicate Quickshell IPC-handler counts.

</details>

### DIFF
--- a/bin/omarchy-font-list
+++ b/bin/omarchy-font-list
@@ -3,4 +3,7 @@
 # omarchy:summary=List available monospace fonts
 # omarchy:examples=omarchy font list | omarchy font set "CaskaydiaMono Nerd Font"
 
-fc-list :spacing=100 -f "%{family[0]}\n" | grep -v -i -E 'emoji|signwriting|omarchy' | sort -u
+{
+  fc-list :spacing=90 -f "%{family[0]}\n"
+  fc-list :spacing=100 -f "%{family[0]}\n"
+} | grep -v -i -E 'emoji|signwriting|omarchy' | sort -u
--- a/test/shell.d/font-list-test.sh
+++ b/test/shell.d/font-list-test.sh
@@ -1,0 +1,42 @@
+#!/bin/bash
+
+set -euo pipefail
+
+source "$(dirname "$0")/base-test.sh"
+
+test_tmp=$(mktemp -d)
+trap 'rm -rf "$test_tmp"' EXIT
+
+mock_bin="$test_tmp/bin"
+call_log="$test_tmp/fc-list-calls"
+mkdir -p "$mock_bin"
+
+cat >"$mock_bin/fc-list" <<'SH'
+#!/bin/bash
+
+printf '%s\n' "$1" >>"$FC_LIST_CALL_LOG"
+
+case "$1" in
+  ":spacing=90")
+    printf '%s\n' "Dual Width" "Shared Family" "Omarchy Dual"
+    ;;
+  ":spacing=100")
+    printf '%s\n' "Mono Width" "Shared Family" "Emoji Mono"
+    ;;
+  *)
+    exit 1
+    ;;
+esac
+SH
+chmod +x "$mock_bin/fc-list"
+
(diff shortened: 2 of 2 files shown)


OldJobobo:
Thanks for the review. Updated the test in a6901223 to check that both spacing queries occur exactly once without enforcing their order. Verified that reversing the queries passes, while missing or duplicate queries still fail.

For the font variants, I propose keeping the standard family names unchanged, consistent with the existing picker. The width trade-off is worth noting, but I do not think this PR needs additional labels or warnings.

> [Pi:gpt-6-astra] - on behalf of OldJobobo.

PR #7717, opened 2026-08-21
Widen the column for full width on scrolling workspaces

donperi:
`Super + Alt + F` ("Full width") is unusable on a scrolling workspace: it fullscreens the window instead of widening it, and while it is active you can neither scroll the workspace nor move focus to another column.

## Demo
While in scrolling mode, now I'm able to switch apps after doing `Super + Alt + F`

https://github.com/user-attachments/assets/aff4bf53-4f86-482b-8eef-26e4763aaead

The cause is that the binding used `fullscreen mode=maximized`, a fullscreen *state* rather than a width change — the window floats above the layout and swallows focus moves.

Hyprland's `scrolling` layout has the right primitive: `layoutmsg colresize` resizes the column, so the window fills the monitor but stays tiled. Scrolling, `Super + Arrow` and `Alt + Tab` keep working. Press again to return to the configured `scrolling:column_width`.

Other layouts, floating windows and already-fullscreen windows keep maximizing exactly as before.

Measured on Hyprland 0.56.2, 1920x1200 @ scale 1.6 (logical width 1200), four tiled columns: both paths widen the window 569 → 1171 px, but only `colresize` lets `Super + Left` move focus to the next column.

## Notes

- This is the `Super + Alt + F` half of #4870, rewritten against Quattro. That PR was closed as targeting the pre-Quattro architecture; this touches only the current Lua bindings.
- #7500 switches `Super + -` / `Super + =` to native `colresize` in this same file. No line overlap, and I am happy to rebase behind it.
- `manual/07-hotkeys.md` already reads "Go full width", so no docs change.
- Covered by `test/shell.d/hyprland-full-width-binding-test.sh`: scrolling narrow/full column, unset `column_width`, rotated monitor, dwindle, floating, already-fullscreen, and special workspace.


default/hypr/bindings/tiling.lua
```diff
@@ -1,3 +1,30 @@
+-- Full width on a scrolling workspace widens the column instead of
+-- fullscreening the window, so the other columns stay reachable. Press again to
+-- return to the configured column width. Other layouts keep maximizing.
+local function full_width()
```
donperi: I don't know if this needs to be a an `omarchy-hypr-` bin.

I asked Claude about it and he was against it.

PR #7718, opened 2026-08-21
Remember the selected wallpaper for each theme

### LINKED ISSUE #7668: Theme switch always restores the first wallpaper, not the last one chosen for that theme
**System details:** Omarchy 4.0.0-1

Each theme can ship several wallpapers. `omarchy theme bg next` / the background switcher change `~/.local/state/omarchy/current/background`, but switching away and back always lands on the first sorted image for that theme.

`choose_theme_background` in `omarchy-theme-set` compares `readlink` of that symlink against the **new** theme's files. The symlink points into `current/theme/backgrounds/`, which was just `rm -rf`'d and replaced, so the path never matches and it falls through to `backgrounds[0]`. There is also no per-theme record of the last choice.

**Steps:**
1. `omarchy theme set tokyo-night`
2. Cycle or pick a wallpaper that is not the first one
3. `omarchy theme set catppuccin`
4. `omarchy theme set tokyo-night`

**Expected:** Tokyo Night comes back on the wallpaper from step 2.
**Actual:** First sorted file under that theme's `backgrounds/`.

Fix: persist last wallpaper per theme slug (basename is enough), restore it on `theme set` when the file still exists, otherwise keep today's first-image fallback. `theme bg next` stays as-is.

OldJobobo:
## Problem

Each theme can provide several wallpapers, but switching away and back always selects its first sorted image. A wallpaper chosen through the background switcher, cycling command, or `omarchy theme bg set` is therefore forgotten as soon as another theme is applied.

Fixes #7668.

## Change

- remember the outgoing theme's selected wallpaper under `~/.local/state/omarchy/theme-backgrounds/`
- restore that exact path when returning to the theme, including user-added and external wallpapers
- fall back to the first available wallpaper when the remembered file no longer exists
- preserve existing behavior when reapplying the active theme: cycle known theme wallpapers, or fall back to the first image for an external wallpaper
- reject unsafe theme names before using them as state keys

State is created lazily during normal theme changes, so existing installations need no migration.

## Behavioral verification

The headless regression test switches repeatedly between Tokyo Night and Catppuccin and verifies:

| Scenario | Expected result |
| --- | --- |
| Return to a previous theme | Its selected wallpaper is restored |
| Themes contain the same filename | The correct theme's remembered wallpaper wins |
| User and packaged wallpapers share a filename | The exact selected path is restored |
| Selected wallpaper is outside the theme directories | The external path is restored |
| Remembered file was removed | The first available wallpaper is selected |
| Active theme is reapplied | Existing cycling/fallback behavior is preserved |
| Theme state contains a traversal path | No state file escapes the state directory |

## Validation

- `bash test/shell.d/theme-background-memory-test.sh`
- `bash test/shell.d/user-theme-test.sh`
- `bash test/shell.d/vscode-theme-test.sh`
- `bash -n bin/omarchy-theme-set test/shell.d/theme-background-memory-test.sh`
- `shellcheck -e SC1091,SC2053 bin/omarchy-theme-set test/shell.d/theme-background-memory-test.sh`
- `./test/cli`

### DIFF
--- a/bin/omarchy-theme-set
+++ b/bin/omarchy-theme-set
@@ -10,8 +10,10 @@ if [[ -z $1 ]]; then
 fi
 
 CURRENT_THEME_PATH="$HOME/.local/state/omarchy/current/theme"
+CURRENT_THEME_NAME_PATH="$HOME/.local/state/omarchy/current/theme.name"
 NEXT_THEME_PATH="$HOME/.local/state/omarchy/current/next-theme"
 CURRENT_BACKGROUND_LINK="$HOME/.local/state/omarchy/current/background"
+THEME_BACKGROUND_STATE_PATH="$HOME/.local/state/omarchy/theme-backgrounds"
 BACKGROUND_TRANSITION_CACHE="$HOME/.cache/omarchy/background-transitions"
 THEME_SET_LOCK="${XDG_RUNTIME_DIR:-/tmp}/omarchy-theme-set.lock"
 USER_THEMES_PATH="$HOME/.config/omarchy/themes"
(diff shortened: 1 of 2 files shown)


munzzyy:
The rewrite keeps the unquoted right hand side on the compare, so a wallpaper whose filename holds glob characters still never matches itself and the loop falls through to the first image:

```
$ backgrounds=("/bg/a.jpg" "/bg/wall [4k].jpg"); current="/bg/wall [4k].jpg"
$ [[ ${backgrounds[1]} == $current ]] || echo "no match"
no match
```

`omarchy-theme-bg-next:32` carried the same line until 5b69af693f6e quoted it. Worth folding into this one, or would you rather keep it separate?

bjarneo:
hey @OldJobobo, can you fix the merge conflict?

OldJobobo:
> hey @OldJobobo, can you fix the merge conflict?

Done

PR #7725, opened 2026-08-22
Codex collector: plan label and model-scoped limit windows

nixfred:
## Changes

Brings the Codex tab of the Agents panel up to parity with the Claude tab. `account/rateLimits/read` already returns more than the collector showed, so this only surfaces what Codex reports:

- `tierLabel` becomes a human plan name (`prolite` → "Pro Lite", `plus` → "Plus") instead of the raw API token, matching Claude's "Max 20x" / "Pro".
- Window labels follow the Claude collector's wording ("Session (5-hour)", "Weekly (7-day)") so the LIMITS section reads the same on both tabs.
- Model-scoped allowances under `rateLimitsByLimitId` (e.g. "GPT-5.3-Codex-Spark", each with its own 5-hour and weekly window) are emitted with a `title`, the way Claude reports "Fable Weekly", so the panel renders them as their own rows instead of dropping them.
- `spendControlReached` surfaces as the usage status text.
- Launches app-server with `-a on-request`, same as #7649 (Codex 0.149 retired `untrusted`, #7648). Happy to rebase on #7649 once it lands — this is meant to stack on it, not compete with it.

Before (Codex 0.149.0, Pro Lite): no plan, `limits: []`.
After:

```
tierLabel: "Pro Lite"
Weekly (7-day)                          6%
GPT-5.3-Codex-Spark Session (5-hour)    0%
GPT-5.3-Codex-Spark Weekly (7-day)      0%
```

## Tests

- `bash test/shell.d/agent-usage-codex-scanner-test.sh` — mock app-server now answers with a realistic `rateLimits` / `rateLimitsByLimitId` payload; asserts the plan label and every limit row.
- `./test/cli`
- live `bin/omarchy-agent-usage-codex --limits-only` against Codex CLI 0.149.0 on Omarchy 4.0.0-1.

https://claude.ai/code/session_01PDPn6hVmawjV1sSRLedsUD

### DIFF
--- a/bin/omarchy-agent-usage-codex
+++ b/bin/omarchy-agent-usage-codex
@@ -495,7 +495,7 @@ def rpc_request(proc, request_id, method, params=None, timeout=8):
   raise TimeoutError(method)
 
 
-def limit_window(window):
+def limit_window(window, title=None):
   if not isinstance(window, dict):
     return None
   used = window.get("usedPercent")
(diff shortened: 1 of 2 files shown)


nixfred:

This PR had drifted so far from its base that GitHub was rendering it as ~118,000 lines across 100 files, which made the Codex collector change impossible to actually review. That was an artifact of the branch point, not the change.

nixfred:
On the Copilot note about the mock not inspecting its arguments — that was right, and it is already handled in this branch. `test/shell.d/agent-usage-codex-scanner-test.sh` now rejects any invocation other than the expected one, so a regression back to `-a untrusted` fails the test rather than passing it:

```bash
# Fail if the collector regresses to `-a untrusted` (Codex 0.149 / #7648).
if [[ $1 != "-s" || $2 != "read-only" || $3 != "-a" || $4 != "on-request" || ${5:-} != "app-server" ]]; then
  echo "unexpected codex invocation: $*" >&2
  exit 1
fi
```

The comment predates the rebase, so it may be pointing at the older diff.

PR #7727, opened 2026-08-22
Add omarchy-hyprland-workspace-compact to collapse workspaces to the lowest numbers

nixfred:
## What

A small `bin/` helper in the `omarchy-hyprland-*` family: **collapse occupied workspaces down to the lowest numbers.**

Open windows on 1–11, close a few, and you're left with 3, 4, 5, 7, 8, 9, 10, 11 occupied and holes at 1, 2, 6. Hyprland has no "renumber workspace" dispatcher, so this moves every window from each occupied workspace into the lowest free slot (ascending, per monitor — the target is always empty by construction), skips special workspaces, and refocuses the active workspace at its new number.

```
$ omarchy-hyprland-workspace-compact --dry-run
workspace 3 -> 1 (eDP-1)
workspace 4 -> 2 (eDP-1)
workspace 7 -> 3 (eDP-1)
$ omarchy-hyprland-workspace-compact
```

## How

- Same shape as `omarchy-hyprland-window-pop`: `hypr_dispatch` tries the Lua dispatcher (`hl.dsp.window.move({ follow = false })`, `hl.dsp.focus`) and falls back to legacy syntax.
- `omarchy:summary=` / `omarchy:args=` headers for the menu.
- Idempotent; `--dry-run` prints the plan and dispatches nothing.
- Test: `test/shell.d/hyprland-workspace-compact-test.sh` (stub `hyprctl`, checks moves, gap collapsing, special-workspace skip, refocus, dry run).

## Keybinding

Deliberately **no default binding** added — your call. `SUPER + MINUS` is unbound in the current defaults if you want one:

```lua
o.bind("SUPER + MINUS", "Compact workspaces", "omarchy-hyprland-workspace-compact")
```

Tested live on Omarchy 4.x / Hyprland 0.56.2 (single monitor; multi-monitor path is covered by per-monitor iteration but not exercised on real hardware).

### DIFF
--- a/bin/omarchy-hyprland-workspace-compact
+++ b/bin/omarchy-hyprland-workspace-compact
@@ -1,0 +1,47 @@
+#!/bin/bash
+
+# omarchy:summary=Collapse occupied workspaces down to the lowest numbers (3,4,7 -> 1,2,3).
+# omarchy:args=[--dry-run]
+
+# Hyprland can't renumber a workspace, so move every window from each occupied
+# workspace into the lowest free slot, ascending per monitor. The target is
+# always empty by construction. Special workspaces (id < 1) are left alone and
+# focus follows the active workspace to its new number.
+
+dry_run=false
+[[ ${1:-} == "-n" || ${1:-} == "--dry-run" ]] && dry_run=true
+
+hypr_dispatch() {
+  local lua="$1"
+  shift
+
+  hyprctl dispatch "$lua" >/dev/null 2>&1 || hyprctl dispatch "$@" >/dev/null
+}
+
+active=$(hyprctl activeworkspace -j | jq -r '.id')
+[[ $active =~ ^-?[0-9]+$ ]] || exit 1
+clients=$(hyprctl clients -j)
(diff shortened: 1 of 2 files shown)


nixfred:
Updated: the script now prefers `hl.dsp.workspace.change_id` (workspace keeps its monitor, windows and tiling layout) and only falls back to moving windows when the target id is held by an empty persistent workspace or the Lua dispatcher is unavailable. Named workspaces are left alone. Test covers both paths.

nixfred:
Follow-up: reverted to moving windows rather than `hl.dsp.workspace.change_id`. change_id emits `changeworkspaceid` which Quickshell's workspace model does not apply, so the stock bar shows the old number as a ghost and the renumbered workspace as empty (forum.hypr.land/t/2076). Window moves emit create/destroy events the bar already handles. Named workspaces remain untouched; test updated.

nixfred:

This PR had drifted so far from its base that GitHub was rendering it as ~118,000 lines across 100 files, which made the workspace-compact script impossible to actually review. That was an artifact of the branch point, not the change.

PR #7729, opened 2026-08-22
Add a Setup menu entry to switch the clock between 24-hour and 12-hour time

Broseph9972:
Adds **Setup > Time Format** to the Omarchy menu with 24-Hour / 12-Hour rows, checkmarked to show the current mode.

- New `omarchy-clock-format` command: prints `12` or `24` with no argument, and converts the clock's format (horizontal and vertical labels) when given one. Custom formats convert best-effort (`dddd HH:mm:ss` → `dddd h:mm:ss AP`), writing through the standard shell-config commit path so the running shell picks the change up immediately.
- Both menu rows read `$(omarchy-clock-format)`, so it is registered in `GUARD_READERS` in `MenuModel.js` per the guard-batch contract.
- FAQ updated to mention the new menu path.

Testing: round-trip get/set against a live config, guard expressions under the batch prelude, `./test/cli` green including menu-guards and command-metadata suites.


Broseph9972:
this was an ai pr, idk the rules on prs feel free to deny if u dont allow ai but i think this is a pretty neccessary change

PR #7736, opened 2026-08-22
Match LocalSend's current Wayland app ID

### LINKED ISSUE #7482: default/hypr/apps/localsend.lua rule doesn't match in Hyprland 0.56 — LocalSend opens tiled and clicks break
## Summary

The shipped LocalSend window rule in `default/hypr/apps/localsend.lua` does not match on Hyprland 0.56, so LocalSend opens **tiled** instead of floating — and while tiled, mouse **clicks on buttons do nothing** (hover works, press events never reach the app). The app is unusable until the window is floated.

## Steps to reproduce

1. Fresh Omarchy install, Wayland/Hyprland session (Hyprland 0.56.2).
2. Launch LocalSend (1.18.1, pacman).
3. Window opens **tiled** (floating: false), despite the shipped rule:
   ```lua
   -- /usr/share/omarchy/default/hypr/apps/localsend.lua
   o.window("(Share|localsend)", { float = true, center = true })
   ```
4. Click any button in the window (e.g. receive a text message, then click its "Copy" button). Hover reacts, click does nothing.
5. Toggle the window floating (SUPER+T) → the same clicks work fine.

## Root cause (hypothesis)

The class regex `(Share|localsend)` is unanchored. The window's class is `org.localsend.localsend_app`, which **contains** the substring `localsend`, so under substring regex semantics the rule should match — it does not.

Every other shipped string rule uses anchors, e.g.:
```lua
o.window("^(1[p|P]assword)$", ...)   -- apps/1password.lua
o.window("^(Bitwarden)$", ...)       -- apps/bitwarden.lua
```
`localsend.lua` appears to be the only unanchored string rule. Hyprland 0.56's `hl.window_rule` class matching seems to require a full-match (anchored) regex.

## Verification

Adding an anchored rule in the user config fixes it — the window floats on launch and clicks work:
```lua
o.window("^org\\.localsend\\.localsend_app$", { float = true, center = true })
```
(Confirmed: `hyprctl clients` reports `floating: true` for the relaunched window.)

Note: LocalSend 1.18.0 switched Linux windows from the GTK3 headerbar to native SSD decorations — likely why tiled hit-testing breaks (press events offset vs. content). Worth checking whether the fix belongs in the omarchy rule, the LocalSend window class, or both.

## Environment

- Omarchy 4.0.0-1
- Hyprland 0.56.2 (v0.56.2, efb5099)
- LocalSend 1.18.1-1 (pacman)
- Arch Linux, kernel 7.1.8-arch1-3, x86_64
- Intel Core Ultra 7 155H / Intel Arc Graphics (integrated), Wayland session

OldJobobo:
## Problem

LocalSend 1.18.1 uses the Wayland class `org.localsend.localsend_app`, but Omarchy only matches `localsend`. The window therefore opens tiled instead of floating, centered, and sized to 1100×700. On this release, buttons in a tiled LocalSend window may also stop accepting clicks.

Fixes #7482.

## Change

Add the current app ID to both LocalSend rules. Keep the legacy `localsend` class and the `Share` file-picker rule. Anchor both expressions and escape the app ID's dots for Hyprland's RE2 matcher.

## Live verification

Tested with `localsend 1.18.1-2` in a running Hyprland 0.56 session on DP-2, workspace 8:

LocalSend tiled with the packaged rules and floating with the candidate rules

| Rules | Class | Floating | Size |
| --- | --- | --- | --- |
| Current packaged rules | `org.localsend.localsend_app` | `false` | `1866×1000` |
| Candidate rules | `org.localsend.localsend_app` | `true` | `1100×700` |

I loaded the candidate file through Hyprland's Lua evaluator, launched LocalSend, and checked it with `hyprctl clients -j`. I then closed LocalSend and reloaded the normal configuration. `hyprctl configerrors` remained empty.

## Tests

The default-config test now checks the current and legacy app IDs, regex escaping, the existing `Share` behavior, and both LocalSend rules.

- `luac -p default/hypr/apps/localsend.lua`
- `bash -n test/shell.d/hyprland-default-config-test.sh`
- `shellcheck -e SC1091,SC2016 test/shell.d/hyprland-default-config-test.sh`
- `./test/cli`
- `git diff --check`

### DIFF
--- a/default/hypr/apps/localsend.lua
+++ b/default/hypr/apps/localsend.lua
@@ -1,3 +1,3 @@
 -- Float LocalSend and fzf file picker.
-o.window("(Share|localsend)", { float = true, center = true })
-o.window("localsend", { size = { 1100, 700 } })
+o.window("^(Share|localsend|org\\.localsend\\.localsend_app)$", { float = true, center = true })
+o.window("^(localsend|org\\.localsend\\.localsend_app)$", { size = { 1100, 700 } })
--- a/test/shell.d/hyprland-default-config-test.sh
+++ b/test/shell.d/hyprland-default-config-test.sh
@@ -105,6 +105,27 @@ grep -Fq $'SUPER + RETURN	Terminal' <<<"$fresh_output" || fail "default applicat
 grep -Fq $'SUPER + SHIFT + A	ChatGPT' <<<"$fresh_output" || fail "default application bindings include preinstalled web apps"
 pass "default application bindings load from package defaults"
 
+OMARCHY_PATH="$ROOT" lua <<'LUA' || fail "LocalSend rules match its current Wayland app ID"
+package.path = os.getenv("OMARCHY_PATH") .. "/?.lua;" .. package.path
+
+local rules = {}
+hl = {
+  window_rule = function(rule)
+    table.insert(rules, rule)
+  end,
+}
+
+require("default.hypr.helpers")
+require("default.hypr.apps.localsend")
+
+assert(#rules == 2)
+assert(rules[1].match.class == "^(Share|localsend|org\\.localsend\\.localsend_app)$")
+assert(rules[1].float == true and rules[1].center == true)
+assert(rules[2].match.class == "^(localsend|org\\.localsend\\.localsend_app)$")
+assert(rules[2].size[1] == 1100 and rules[2].size[2] == 700)
+LUA
+pass "LocalSend rules match its current Wayland app ID"
+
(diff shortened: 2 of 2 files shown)


peteonrails:
Reproduced on Omarchy 4.0.0.r1827, localsend 1.18.2, Hyprland 0.56.2. Cold launch with the packaged rules: class `org.localsend.localsend_app`, `floating: false`, 656x1167. Same session, anchored rule in my own `hyprland.lua`: floating, 1100x700.

I didn't test the dead-clicks symptom from #7482, so no second data point on that one.

PR #7738, opened 2026-08-22
Notifications: clicking a toast focuses the exact sending window when focus_on_activate is off

nixfred:
## Problem

With `misc:focus_on_activate = false` (common for people who hate focus stealing), clicking a notification toast often does nothing: the sender's default action fires an xdg-activation request, Hyprland denies it, and the existing class-match fallback never runs because `invoked` is true. For apps with many windows (kitty running several agents, editors) nothing moves.

## Fix

Hyprland still emits `urgent>>ADDR` for the exact window that asked for activation. On click, arm a 1 s listener on `Hyprland.rawEvent`; when the urgent event lands, focus that window by address with `hl.dsp.focus({ window = "address:0x…" })` (same dispatcher `omarchy-hyprland-focus-app` uses). If no urgent event arrives (app never requested activation), fall back to the class match exactly as before.

39 lines, one file (`shell/plugins/notifications/Service.qml`), no new dependencies beyond `Quickshell.Hyprland`.

## Tested

Omarchy 4.0.0 / Hyprland 0.56 with `focus_on_activate=false`: notifications from kitty (multiple windows), Chromium web apps and libnotify now jump to the precise window on click; with `focus_on_activate=true` behaviour is unchanged (activation wins, listener is a no-op).

https://claude.ai/code/session_01D411mB4n7uBWNSvTJkpmJa

### DIFF
--- a/shell/plugins/notifications/Service.qml
+++ b/shell/plugins/notifications/Service.qml
@@ -5,6 +5,7 @@ import QtQuick.Layouts
 import Quickshell
 import Quickshell.Io
 import Quickshell.Wayland
+import Quickshell.Hyprland
 import Quickshell.Services.Notifications
 import qs.Commons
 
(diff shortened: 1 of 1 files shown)


nixfred:

This PR had drifted so far from its base that GitHub was rendering it as ~118,000 lines across 100 files, which made the notification click-to-focus change impossible to actually review. That was an artifact of the branch point, not the change.

nixfred:
On the Copilot note about the timeout moving focus when the sender is already focused — that was a real case, and it is handled in this branch. Hyprland 0.56 returns from `activate()` without emitting `urgent` when the requested window already has focus, so the fallback had to learn to tell that apart from an app that never requests activation at all.

```qml
onTriggered: {
  var entry = service.urgentFocusEntry
  service.urgentFocusEntry = null
  // Hyprland 0.56 skips the urgent event when the requested window is
  // already focused. Do not then class-match onto a sibling window.
  if (entry && !service.senderAlreadyFocused(entry)) service.focusApp(entry)
}
```

`senderAlreadyFocused` compares `Hyprland.activeToplevel` against the sender's class, so a multi-window app whose window is already up is left alone rather than having a sibling pulled forward.

The comment predates the rebase, so it may be pointing at the older diff.

PR #7742, opened 2026-08-22
Keep Spotify volume across track changes

### LINKED ISSUE #7741: Spotify volume resets when the track changes
### System details

Omarchy 4.0.0-1, Spotify 1.2.96.518-1, WirePlumber 0.5.15, PipeWire 1.6.8

### What is wrong?

Changing Spotify volume in the audio panel only lasts for the current track. When the next track starts, Spotify restores its previous volume.

Steps to reproduce:
1. Play a song in Spotify.
2. Open the audio panel and change Spotify volume.
3. Skip to the next song.

The panel changes the live audio stream, but Spotify also keeps its own player volume. Updating both values prevents the reset.

sdye1337:
## What changed

- Update both the media player volume and its live audio stream.
- Keep the current behavior for apps that do not support player volume.
- Cover mouse and keyboard volume changes with tests.

## Why

Spotify keeps its own volume. The audio panel changed only the current stream, so Spotify restored the old value when the next track started.

Fixes #7741

## Tests

- `bash test/shell.d/audio-test.sh` passes.
- The audio and runtime checks in `./test/all` pass. The full suite still reports four unrelated local setup checks: display geometry and a missing `omarchy-pkgs` checkout.

### DIFF
--- a/shell/plugins/panels/audio/Model.js
+++ b/shell/plugins/panels/audio/Model.js
@@ -233,6 +233,41 @@ function streamRepresentsPlayer(node, player, players, streams) {
   return streamRepresentsMprisPlayer(streamLabel(node, players, streams), playerLabel)
 }
 
+function mprisPlayerForStream(node, players, streams) {
+  var values = Array.isArray(players) ? players : []
+  var candidate = null
+  var proxyCandidate = null
+
+  for (var i = 0; i < values.length; i++) {
+    var player = values[i]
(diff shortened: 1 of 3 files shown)


sdye1337:
Closing this for now because live Spotify testing showed the volume can still reset between tracks. I’ll reproduce and verify the behavior end to end before proposing another fix.

PR #7751, opened 2026-08-22
Fix Codex limits collector approval policy

purple-monkey-dishwasher:
# Codex limits unavailable because collector uses removed `untrusted` approval policy

## What happened

On Omarchy 4.0.0-1 with Codex CLI 0.149.0, the Agents toolbar panel shows `Codex limits unavailable` and a red `initialize` error even though `codex login status` reports `Logged in using ChatGPT`.

The panel still shows historical local usage, but the plan and rate-limit windows are absent.

## Root cause

`bin/omarchy-agent-usage-codex` launches the app server with:

```text
```

Codex CLI 0.149.0 rejects `untrusted`:

```text
```

Because stderr is discarded, the collector waits for the `initialize` RPC until it times out. It then exposes the exception method name as `authHelpText`, producing the misleading red `initialize` message.

## Expected behavior

The collector should start the current Codex app server non-interactively and return the authenticated account's plan and rate-limit windows.

## Proposed fix

Replace `-a untrusted` with `-a never` and add a regression assertion covering the exact app-server launch arguments.

With this change, the collector returned the authenticated plan and a populated 7-day limit window on the affected machine, with an empty `usageStatusText`.

## Reproduction

1. Install Omarchy 4.0.0-1 and Codex CLI 0.149.0.
2. Authenticate using `codex login`.
3. Run `omarchy agent usage-update --force codex`.
4. Open the Agents toolbar panel.

## Verification

- `bash test/shell.d/agent-usage-codex-scanner-test.sh` passes.
- `git diff --check` passes.
- Live authenticated collector probe returns the plan and limit window.
- The broader `./test/shell` suite reaches the changed test successfully; four unrelated environment/checkout-dependent tests fail because the separate `omarchy-pkgs` checkout is unavailable (plus its dependent coverage checks).

### DIFF
--- a/bin/omarchy-agent-usage-codex
+++ b/bin/omarchy-agent-usage-codex
@@ -528,7 +528,7 @@ def fetch_codex_rpc():
 
   try:
     proc = subprocess.Popen(
-      [codex, "-s", "read-only", "-a", "untrusted", "app-server"],
+      [codex, "-s", "read-only", "-a", "never", "app-server"],
       stdin=subprocess.PIPE,
       stdout=subprocess.PIPE,
       stderr=subprocess.DEVNULL,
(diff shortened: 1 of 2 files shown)


s0up4200:
Dupe of https://github.com/basecamp/omarchy/pull/7649

PR #7757, opened 2026-08-22
Read the tmux session list once per theme change

Adolanium:
`set_tmux_environment` calls `tmux list-sessions` every time it runs, and it runs once for every theme environment key. There are 116 of those keys, counted from `default/themed/gum_env.lua.tpl` with the same awk filter the script itself uses. The session list does not change while the script is running, so all of those calls ask the same question and get the same answer.

This reads the list once up front and loops over it instead.

Measured against real tmux with two sessions open:

- `tmux show-environment` afterwards is identical, globally and for both sessions, 392 lines each way.
- This script goes from 1.02s to 0.78s, averaged over 5 runs each, with every run checked to have actually populated the session environment.
- The number of `list-sessions` calls goes from 118 to 2.

Worth being clear about what that 0.24s does and does not buy. This script runs in the `run_parallel` fan out in `omarchy-theme-set`, which waits on all 14 commands, so it only shortens the whole thing if this is the slowest of the 14, and I have not checked whether it is. That block also runs after the `flock -u 9` release, once the theme is applied and the shell has accepted it, so nothing the user is looking at happens sooner. What actually improves is that tmux panes pick up the new colors about 0.24s earlier, and only for people with a tmux server running at all. Everyone else exits at the guard on line 11 either way.

So this is a small win on its own. The better reason to take it is that asking tmux the same question 116 times in a row is simply wrong, and the fix is four lines.

Tests: new `test/shell.d/theme-set-tmux-test.sh` runs the script against a stubbed `tmux` that reports two sessions and logs every call. It checks that `list-sessions` is called exactly twice (the guard and the one read), that every key from a three-key `gum_env.lua` plus `COLORFGBG` is set globally and in both sessions, and that each key is set once per session. Against the old script the first check fails with five calls. The existing `./test/cli` tmux case only checks that `COLORFGBG` lands, so it passes on both versions and would not have caught a slide back to per-key listing.

One behavior note: a session created while the script is running will no longer pick up the keys that are left. Before this change it would have picked up whatever had not been set yet. A theme sync takes well under a second, and the functions that touch panes already read their list once and work from that, so this seemed like the right trade.

There is more that could be done here. The calls that remain are one `set-environment` per key per session, and tmux can take several commands in a single call. I left that out because `COLORFGBG` is set to a value with a semicolon in it, and tmux reads a bare semicolon as a command separator, and because the current code ignores a failure on any single key on purpose, which one combined call would give up. I have not tested that version, so it is not part of this change.

Tested on Ubuntu, bash 5.2.21, tmux 3.4.

### DIFF
--- a/bin/omarchy-theme-set-tmux
+++ b/bin/omarchy-theme-set-tmux
@@ -11,6 +11,10 @@ if ! tmux list-sessions >/dev/null 2>&1; then
   exit 0
 fi
 
+# The theme has well over a hundred environment keys, and the session list is
+# the same for all of them, so read it once here rather than once per key.
+mapfile -t tmux_sessions < <(tmux list-sessions -F "#{session_id}" 2>/dev/null)
+
 set_tmux_environment() {
   local key="$1"
   local value="$2"
@@ -18,10 +22,10 @@ set_tmux_environment() {
 
   tmux set-environment -g "$key" "$value" 2>/dev/null || return 0
 
-  while IFS= read -r session; do
+  for session in "${tmux_sessions[@]}"; do
     [[ -n $session ]] || continue
     tmux set-environment -t "$session" "$key" "$value" 2>/dev/null || true
-  done < <(tmux list-sessions -F "#{session_id}" 2>/dev/null)
+  done
 }
 
 sync_theme_environment() {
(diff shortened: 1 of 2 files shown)


Adolanium:
Added a test. The existing `test/cli` tmux case only checks that `COLORFGBG` lands, so it passed on both the old and new script and would not have caught a slide back to per-key listing. `test/shell.d/theme-set-tmux-test.sh` stubs `tmux` with two sessions, counts the `list-sessions` calls (2 now, 5 on the old script with a three-key theme), and checks every key still reaches every session.

PR #7767, opened 2026-08-22
Add Super + Q as a second chord for closing a window

omarchybot:
`Super + W` closes a window, and nothing was bound to `Super + Q` at all. People arriving from macOS reach for `Command + Q` to make an app go away, press the same key here, and get silence — no close, no error, nothing to tell them the chord is wrong rather than broken.

This binds `Super + Q` to the same close dispatcher, as an alternative rather than a replacement: `Super + W` stays the documented default. `Super + Ctrl + Q` (Calculator) is untouched, and no other default claims the chord. The manual gets the alternative in the hotkey table, in the navigation chapter, and in the page for people coming from Mac or Windows, which is where the chord is most likely to be looked for.

The second commit is what makes that read well in `Super + K`. The menu listed every chord on its own row, so the two showed up as "Close window" twice, sixteen rows apart, with the alternative sorted above the default. The scratchpad and the calculator had the same trouble, each bound to a chord and to a second key:

```
SUPER + W / SUPER + Q               → Close window
SUPER + S / SUPER + ~               → Toggle scratchpad
SUPER ALT + S / SUPER SHIFT + ~     → Move window to scratchpad
SUPER CTRL + Q / XF86Calculator     → Calculator
```

Those four actions are named one at a time rather than merged by rule. A rule would be wrong here: `Alt + Tab` and `Shift + Alt + Tab` both say "Reveal active window on top" while cycling opposite ways, and a media key is nobody's idea of an alternative to a Super chord. Both halves still have to agree on what they dispatch, since a label is only what a chord is called.

Nothing is allowed past the 35-character column: a pair that would overrun it stays as two rows rather than pushing its arrow out of line. The menu elides a row that outgrows its card — 754px of label, 78 monospace characters at the heading size — and the longest entry already sits at 74, so widening the column to fit the widest pair would have cost two dozen rows the end of their description.

The key left of 1 now reads as `~` rather than Hyprland's name for it, whether a bind names it or reports the keycode for the keymap to resolve. Everything else is exactly where it was: the list drops four duplicate rows, and no row is any longer or ranked differently than before.

🤖 Generated by Opus 5 in Claude Code. Reviewed by Codex XHigh.

### DIFF
--- a/bin/omarchy-menu-keybindings
+++ b/bin/omarchy-menu-keybindings
@@ -38,6 +38,12 @@ parse_keycodes() {
         if (code != "" && symbol != "" && symbol != "NoSymbol") keycode_symbol[code] = toupper(symbol)
       }
 
+      # Read off the keysym rather than the keycode, so the key left of 1 reads
+      # as the symbol printed on it whichever key the active layout puts it on.
+      for (code in keycode_symbol) {
+        if (keycode_symbol[code] == "GRAVE") keycode_symbol[code] = "~"
+      }
+
       mouse_symbol["272"] = "LEFT MOUSE BUTTON"
       mouse_symbol["273"] = "RIGHT MOUSE BUTTON"
       mouse_symbol["274"] = "MIDDLE MOUSE BUTTON"
@@ -305,6 +311,7 @@ dynamic_bindings() {
 
       case "$key" in
         comma) key="COMMA" ;;
+        grave) key="~" ;;
         period) key="PERIOD" ;;
         minus) key="MINUS" ;;
         equal) key="EQUAL" ;;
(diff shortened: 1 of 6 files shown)


phucisstupid:
nice

bdnugget:
Nice, I got here because I wanted to change it myself but couldn't figure out I needed hl.dsp.window.close(), I thought it would be closewindow or killactive.

One remark, it says here "~" is "grave" but it's a "tilde" as far as I know. Grave is a ` backtick,or is it because they are on the same key on US keyboards?
https://github.com/basecamp/omarchy/commit/07fccef41c974b54f893bcad62c100f96cad4b62#diff-d3399e77eb915d89ee6df277737a59ef783bc43dc43c0eb69ed1637278edfc31R314

PR #7770, opened 2026-08-22
Launch Codex app-server with an approval policy the CLI accepts

### LINKED ISSUE #7648: Codex 0.149 breaks Agents panel limit display

carinaesoft:
Fixes #7648.

Codex 0.149 retired the `untrusted` approval policy ([openai/codex#39630](https://github.com/openai/codex/pull/39630)), so the collector's launch never reaches app-server:

```console
```

`fetch_codex_rpc` discards app-server's stderr and waits on the `initialize` reply, so the immediate exit surfaces as `TimeoutError("initialize")`. The panel then shows `Codex limits unavailable` over the bare method name — the symptom several people reported in the issue. Local token and session stats keep working, since they come off the session files.

Asking for `on-request` restores the limits. It is the closest surviving policy to the original intent: nothing is auto-approved, `-s read-only` still bounds the sandbox, and the app-server is only ever read from (`account/read`, `account/rateLimits/read`).

## Test

`test/shell.d/agent-usage-codex-limits-test.sh` stubs `codex` the way the real CLI behaves: it validates `--ask-for-approval` before speaking the protocol, and rejects a retired value. The test fails on `quattro` with the reported `authHelpText: "initialize"` and passes with the fix.

```
ok - Codex collector reads the window app-server reports
ok - Codex collector names the plan app-server reports
ok - Codex collector launches app-server with an approval policy the CLI accepts
```

Verified against a live Codex 0.149.0 sign-in too: the collector returns `tierLabel: "team"`, an empty `usageStatusText`, and the weekly window.

`./test/all` is otherwise unchanged — the 4 failures on this machine (`bar-icon-geometry`, `config`, `snapper`, `unowned-system-paths`) reproduce on a clean `quattro` checkout.

## Not in this PR

The issue also notes that an early app-server exit should not be reported as the pending RPC method name. That is a separate change to the error handling and I left it out to keep this atomic — happy to follow up if you want it.

### DIFF
--- a/bin/omarchy-agent-usage-codex
+++ b/bin/omarchy-agent-usage-codex
@@ -528,7 +528,7 @@ def fetch_codex_rpc():
 
   try:
     proc = subprocess.Popen(
-      [codex, "-s", "read-only", "-a", "untrusted", "app-server"],
+      [codex, "-s", "read-only", "-a", "on-request", "app-server"],
       stdin=subprocess.PIPE,
       stdout=subprocess.PIPE,
       stderr=subprocess.DEVNULL,
--- a/test/shell.d/agent-usage-codex-limits-test.sh
+++ b/test/shell.d/agent-usage-codex-limits-test.sh
@@ -1,0 +1,78 @@
+#!/bin/bash
+
+source "$(dirname "$0")/base-test.sh"
+
+require_command jq
+require_command python3
+
+# Codex 0.149 retired the "untrusted" approval policy, and the CLI rejects an
+# unknown one before app-server starts. The collector discards app-server's
+# stderr and waits for an RPC reply, so that immediate exit surfaced as the
+# pending method name — "initialize" — and the panel lost its limits with no
+# hint of why. This stub answers the way the CLI does: it validates the policy
+# first, and only then speaks the protocol.
+TEST_HOME=$(mktemp -d)
+trap 'rm -rf "$TEST_HOME"' EXIT
+
+mkdir -p "$TEST_HOME/.codex" "$TEST_HOME/bin"
+
+cat >"$TEST_HOME/bin/codex" <<'EOF'
+#!/bin/bash
+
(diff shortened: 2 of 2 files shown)


upnorthmedia:
confirmed on my end. codex 0.149.0, aarch64 asahi. limits came back right after the change,
plan and weekly window both populating.

separate from the fix itself, but it's the reason this one was hard to see from outside.

when the flag gets rejected the cli says exactly what's wrong:

```
```

that goes to DEVNULL, so all the collector sees is the process exiting, stdout closing, and
initialize timing out at 8s. the except path does authHelpText = str(exc), and rpc_request
raises TimeoutError(method), so the field ends up reading "initialize". in the panel that
reads as an auth problem, which sends you looking in the wrong place.

capturing stderr and preferring it would let the next cli flag change explain itself:

```python
      stderr=subprocess.PIPE,
```

```python
  except Exception as exc:
    result["usageStatusText"] = "Codex limits unavailable"
    detail = ""
    if proc.poll() is not None:
      try:
        detail = (proc.stderr.read() or "").strip().splitlines()[0]
      except Exception:
        pass
    result["authHelpText"] = detail or AUTH_HELP
```

the poll() guard keeps it from reading a live process, and falling back to AUTH_HELP keeps
an rpc method name out of a user facing string.

tested against 0.149.0 three ways. bad flag puts the cli error into authHelpText instead of
"initialize". good flag still returns limits and plan exactly like it does now. and a process
that starts but never answers falls back to AUTH_HELP at the timeout with no hang, which is
the case the poll() guard is there for.

happy to put this up as its own pr rather than have it ride along here, since AGENTS.md
asks for atomic commits.

PR #7780, opened 2026-08-22
Dismiss the screensaver on any input, not just a keypress

### LINKED ISSUE #7762: [Quattro] Screensaver only exits on a keypress inside its own terminal — global mouse/keyboard activity is explicitly ignored while it's up
### What happened
Once the native screensaver (ttfx, launched via `omarchy-launch-screensaver`) is
showing, moving the mouse does nothing — it keeps running until a key is
pressed *while the screensaver terminal has focus*. On the old hypridle-based
setup, any global input (including mouse movement) dismissed the screensaver
immediately via `on-resume`. This is a regression in day-to-day usability:
grabbing the mouse to check something no longer wakes the screen the way it
used to on every other idle-lock implementation (GNOME, KDE, plain hypridle).

### Root cause
`shell/plugins/services/idle/Service.qml`, `handleActiveSignal()`:

```qml
function handleActiveSignal() {
  if (!root.idledThisCycle) return

  // Starting the screensaver can make the compositor report activity. Keep
  // the lock timer running once the screensaver exists (or during its short
  // launch grace); Hyprland window events cancel the cycle if it exits before
  // the normal lock deadline.
  if (root.screensaverStartedThisCycle && (root.screensaverWindowCount > 0 || screensaverLaunchGraceTimer.running)) {
    logEvent("idle-monitor-active", "screensaver cycle remains armed")
    return
  }

  cancelIdleCycle("activity")
}
```

`screensaverWindowCount > 0` is included in the guard, so **for the entire
time the screensaver window exists**, the idle monitor's activity signal
(which does fire on mouse movement — verified via `logEvent`) is discarded
and `cancelIdleCycle()` is never reached. The comment explains why the guard
exists at all (the screensaver's own window mapping can spuriously report
activity right at launch), but it over-broadly suppresses *all* activity for
the window's entire lifetime, not just the brief launch window.

Even fixing that guard isn't enough on its own: `cancelIdleCycle()` never
actually kills the screensaver process/window — it only stops timers and
calls `omarchy-system-wake` (display brightness). The screensaver window
stays open and unresponsive to anything but its own internal
`read -n1`-based keypress loop (`omarchy-screensaver`), which only sees input
while it holds terminal focus, never mouse motion.

### Expected
Any global input (mouse or keyboard) while the screensaver is up should
dismiss it, matching pre-Quattro `hypridle` behavior and every other
DE's idle-lock.

### Suggested fix (tested locally)
1. Narrow the guard in `handleActiveSignal()` to only the launch grace,
   not the window's entire lifetime:
   ```qml
   if (root.screensaverStartedThisCycle && screensaverLaunchGraceTimer.running) {
   ```
2. Make `cancelIdleCycle()` actually kill the screensaver window when one is
   open:
   ```qml
   if (root.screensaverWindowCount > 0) runProcess(dismissProcess, "dismiss-screensaver", "pkill -f \"[o]rg.omarchy.screensaver\" 2>/dev/null; true")
   ```
   (plus a matching `dismissProcess: Process { ... }` element alongside the
   existing `screensaverProcess`/`lockProcess`/`wakeProcess`).

I've applied this locally via `omarchy plugin clone omarchy.idle` and it
resolves the issue without side effects on lock timing — the screensaver
still ignores activity during its brief launch grace (so it doesn't
self-cancel on its own window mapping), but responds to mouse/keyboard once
it's actually up.

### System
- Omarchy 4.0.0 "Quattro"
- quickshell-git 0.3.0.r20.g28771c7
- Hyprland (Lua config)

husamemadH:
Fixes #7762.

Once the screensaver is up, moving the mouse does nothing — it only goes away on a keypress delivered to its own terminal. Pre-Quattro, hypridle's `on-resume` dismissed it on any global input.

## Why the main idle monitor cannot see the input

The screensaver focusing its own window registers as activity, so `IdleMonitor.isIdle` flips back to `false` a fraction of a second after launch and then never changes for the rest of the display. `onIsIdleChanged` never fires again, so `handleActiveSignal()` is never reached, and the `screensaverWindowCount > 0` guard inside it is not what suppresses the dismissal — the signal simply never arrives.

Traced on an affected machine (polling the idle service's own IPC status; the key includes `lastEventAt`, so a repeated event would still show):

```
17:48:28.948 | isIdle=true  inCycle=true ssWindows=0 | process-start: screensaver
17:48:29.240 | isIdle=false inCycle=true ssWindows=1 | process-exit: screensaver
   ... ~12s of continuous mouse movement: no events at all ...
17:48:41.211 | isIdle=false inCycle=false ssWindows=0 | process-exit: wake
```

Two cycles in the same run, with the self-inflicted flip landing at 292ms and 295ms after launch. So narrowing that guard would edit unreachable code.

`cancelIdleCycle()` is also not a teardown: it stops timers and runs `omarchy-system-wake`, leaving the screensaver window on screen.

## The fix

A second `IdleMonitor`, enabled only while a screensaver window exists. It arms once input has gone quiet — which absorbs the launch activity without needing a grace timer — and the next activity dismisses the screensaver with the same `pkill` that `omarchy-system-lock` already uses. The existing `closewindow` → `handleScreensaverWindowClosed()` path then cancels the pending lock and wakes the display, unchanged.

Because `ext-idle-notify` is compositor-side, this covers keyboard, pointer motion, clicks, touch, and bluetooth devices without the terminal being involved. Manually launched screensavers (`omarchy-launch-screensaver force`) become dismissable the same way, since window tracking is driven by Hyprland events rather than the idle cycle.

The arm/dismiss decision is a pure function in `IdleModel.js`, covered by `test/shell.d/idle-test.sh`.

## Verified in the running UI

Same machine, patched service loaded via `omarchy plugin clone omarchy.idle`, three cycles:

```
17:53:00.555 | isIdle=true  ssWindows=0 armed=false | process-start: screensaver
17:53:00.850 | isIdle=false ssWindows=1 armed=false | process-exit: screensaver     <- 295ms blip, ignored
17:53:01.712 | isIdle=false ssWindows=1 armed=true  | screensaver-dismiss-armed
17:53:06.283 | isIdle=false ssWindows=1 armed=false | process-start: dismiss pkill -f '[o]rg.omarchy.screensaver'
17:53:06.574 | isIdle=false ssWindows=0 armed=false | process-exit: wake
```

Mouse movement dismisses it immediately; teardown measured at 172ms window-close latency in a 50ms-resolution trace. Lock timing is unaffected — the dismissal monitor is disabled whenever no screensaver window is on screen, and `lockSystem()` resets it.

Terminal: foot. Hardware: Intel UHD 630 + RTX 2080 Mobile, wired mouse.

## Related

- #7528 diagnoses the same root cause and also reaches for a second `IdleMonitor`; its state machine and arming differ from this one, and it carries touch-specific findings (including an upstream foot crash) that this PR does not. Reviewers may prefer to land one and close the other.
- #6813 reworks screensaver teardown into per-supervisor ownership; the `pkill` here is the same call `omarchy-system-lock` makes today and should be swapped for whatever lands there.
- #7102 and #7193 touch adjacent screensaver behavior.

## Tests

`./test/shell` — 2301 assertions pass, 4 of them new; 182 of 185 test files pass. The 3 failing files (`config-test.sh`, `snapper-test.sh`, `unowned-system-paths-test.sh`) all fail on the same missing `omarchy-pkgs` sibling checkout on my machine (`omarchy-pkgs checkout found/available for ...`) and reproduce identically on a clean tree. `./test/cli` passes.

### DIFF
--- a/shell/plugins/services/idle/IdleModel.js
+++ b/shell/plugins/services/idle/IdleModel.js
@@ -43,10 +43,21 @@ function screensaverWindowsAfter(windows, address, visible) {
   }
 }
 
+// The screensaver's own window mapping registers as input, so a dismissal
+// monitor cannot treat its first activity as the user. It arms only once input
+// has gone quiet while the screensaver is on screen; the next activity after
+// that is a real keypress, click, or pointer move and dismisses the screensaver.
+function screensaverDismissAction(armed, isIdle, screensaverVisible) {
+  if (!screensaverVisible) return "ignore"
+  if (isIdle) return armed ? "ignore" : "arm"
+  return armed ? "dismiss" : "ignore"
+}
+
 if (typeof module !== "undefined") {
   module.exports = {
     secondsFromConfig: secondsFromConfig,
     eventParts: eventParts,
-    screensaverWindowsAfter: screensaverWindowsAfter
+    screensaverWindowsAfter: screensaverWindowsAfter,
+    screensaverDismissAction: screensaverDismissAction
   }
 }
--- a/shell/plugins/services/idle/Service.qml
+++ b/shell/plugins/services/idle/Service.qml
@@ -24,6 +24,7 @@ Item {
   readonly property int lockDelaySeconds: Math.max(0, lockTimeoutSeconds - firstIdleTimeoutSeconds)
   readonly property bool idleEnabled: stayAwakeStateLoaded && !stayAwake
   readonly property string screensaverClass: "org.omarchy.screensaver"
+  readonly property int screensaverDismissSettleSeconds: 1
 
   property bool stayAwake: false
   property bool stayAwakeStateLoaded: false
@@ -35,6 +36,7 @@ Item {
   property string lastEventAt: ""
   property var screensaverWindows: ({})
   property int screensaverWindowCount: 0
+  property bool screensaverDismissArmed: false
 
   function secondsFromConfig(value, fallback) {
     return IdleModel.secondsFromConfig(value, fallback)
@@ -113,6 +115,29 @@ Item {
   function resetScreensaverWindows() {
     root.screensaverWindows = ({})
     root.screensaverWindowCount = 0
+    root.screensaverDismissArmed = false
+  }
+
+  function dismissScreensaver(reason) {
+    logEvent("screensaver-dismiss", reason || "input")
+    root.screensaverDismissArmed = false
+    runProcess(dismissProcess, "dismiss", "pkill -f '[o]rg.omarchy.screensaver'")
+  }
+
+  // The main idle monitor reports activity once while the screensaver maps its
+  // window and then stays active for as long as it is up, so it never sees the
+  // input that should dismiss it. A second monitor, armed only while the
+  // screensaver is on screen, supplies that signal for every input device the
+  // compositor knows about.
+  function handleScreensaverDismissSignal() {
+    var action = IdleModel.screensaverDismissAction(root.screensaverDismissArmed, screensaverDismissMonitor.isIdle, root.screensaverWindowCount > 0)
+
+    if (action === "arm") {
+      root.screensaverDismissArmed = true
+      logEvent("screensaver-dismiss-armed")
+    } else if (action === "dismiss") {
+      dismissScreensaver("input")
+    }
   }
 
   function setScreensaverWindow(address, visible) {
@@ -128,6 +153,7 @@ Item {
 
   function handleScreensaverWindowClosed(address) {
     setScreensaverWindow(address, false)
+    if (root.screensaverWindowCount === 0) root.screensaverDismissArmed = false
 
     if (!root.idleEnabled || !root.idledThisCycle || !root.screensaverStartedThisCycle) return
     if (root.screensaverWindowCount > 0) return
@@ -191,6 +217,7 @@ Item {
       screensaverDelay: root.screensaverDelaySeconds,
       lockDelay: root.lockDelaySeconds,
       screensaverWindows: root.screensaverWindowCount,
+      screensaverDismissArmed: root.screensaverDismissArmed,
       timers: {
         screensaver: screensaverTimer.running,
         lock: lockTimer.running,
@@ -199,7 +226,8 @@ Item {
       processes: {
         screensaver: screensaverProcess.running,
         lock: lockProcess.running,
-        wake: wakeProcess.running
+        wake: wakeProcess.running,
+        dismiss: dismissProcess.running
       },
       lastEvent: root.lastEvent,
       lastEventAt: root.lastEventAt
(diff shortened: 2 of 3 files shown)


ClGratton:
I found one required integration guard while composing this dismissal monitor with the verified idle lock handoff in #9632.

At the lock deadline, `lockSystem()` disarms the idle cycle and disables the screensaver dismissal monitor. That monitor shutdown can emit an active-state change. Without a lifecycle guard, `handleScreensaverDismissSignal()` classifies it as input and kills the screensaver before the concealed ext-session-lock surface is secure. On the dual-output hardware test this exposed the desktop for about 0.7 seconds even though #9632 was correctly waiting for `secure=true` before normal cleanup.

The verified composition rejects the callback once locking has started:

```qml
function handleScreensaverDismissSignal() {
  if (!root.idledThisCycle || !root.screensaverStartedThisCycle) return

  var action = IdleModel.screensaverDismissAction(
    root.screensaverDismissArmed,
    screensaverDismissMonitor.isIdle,
    root.screensaverWindowCount > 0)
  // existing handling
}
```

The journal made the race unambiguous: `lock-system: lock-timeout` and `screensaver-dismiss: input` appeared at the same timestamp, the dismiss process exited 138 ms later, the concealed lock request followed at 163 ms, and the session became secure about 855 ms after the deadline. With this guard, the screensaver remained mapped until the secure handoff and the user confirmed there was no password or desktop flash before real DPMS standby.

ClGratton:
Tested successfully on Omarchy 4.0.1 with Hyprland 0.56.2, foot 1.27.0, and two active outputs. The compositor-side `IdleMonitor` correctly dismisses the screensaver for pointer and keyboard activity, but two lifecycle guards are required for reliable composition.

Arming must wait until the sequential multi-output launcher has completed. Enabling after the first tracked window lets the monitor settle before the second output maps, so the second mapping can be mistaken for user input. The verified local service gates arming on launch completion as well as `screensaverWindowCount > 0`. It also targets the screensaver supervisor for teardown, matching the lifecycle direction in #6813.

The dismissal callback must also reject active-state changes after the idle cycle enters locking. Disabling the monitor at the lock deadline can emit an active-state signal; treating that signal as input tears the screensaver down before #9632's concealed lock handoff is secure and briefly exposes the desktop. The verified guard is:

```qml
function handleScreensaverDismissSignal() {
  if (!root.idledThisCycle || !root.screensaverStartedThisCycle) return

  var action = IdleModel.screensaverDismissAction(
    root.screensaverDismissArmed,
    screensaverDismissMonitor.isIdle,
    root.screensaverWindowCount > 0)
  // existing handling
}
```

With both guards, two screensaver windows mapped, compositor input dismissed them cleanly, and the later idle-lock transition reached genuine standby without a password or desktop flash.

bjarneo:
Automated duplication check: this pull request looks similar to #7528, which covers the same screensaver input dismiss. I keep that one open and close this one to consolidate review.

PR #7782, opened 2026-08-22
Fix Codex usage widget: -a untrusted is no longer a valid approval policy

### LINKED ISSUE #7781: Codex usage widget always shows "Codex limits unavailable" (initialize timeout) with current codex-cli
## What happened

The AI agent usage panel's Codex tab always shows "CODEX LIMITS UNAVAILABLE" with the status text `initialize`. Local token stats (tokens by day, tokens by model) populate correctly, but the live rate-limit/plan section never loads.

## Root cause

`/usr/share/omarchy/bin/omarchy-agent-usage-codex` starts the Codex app-server with:

```
```

The installed `codex-cli` (0.149.0) no longer accepts `untrusted` as a value for `-a`/`--ask-for-approval`:

```
```

Because the subprocess exits immediately on this CLI parse error instead of speaking the app-server RPC protocol, the script's `initialize` RPC call (in `fetch_codex_rpc()`) never gets a response and hits its 8-second timeout. The resulting `TimeoutError("initialize")` is what surfaces as the literal text `initialize` in the widget's status/help line.

## Expected

The Codex tab should show real rate limits, or at least a clearer error, instead of silently timing out due to an invalid CLI flag value.

## Steps to reproduce

1. Have current `codex-cli` (0.149.0+) installed and authenticated (`codex login`).
2. Open the AI agent usage panel, switch to the Codex tab.
3. Observe "CODEX LIMITS UNAVAILABLE" / `initialize`.
4. Confirm directly: `codex -s read-only -a untrusted app-server` errors with an invalid value for `-a`.

## Suggested fix

Update the `-a` value passed in `fetch_codex_rpc()` (in `omarchy-agent-usage-codex`) to one of the currently accepted values (`on-request` or `never`), or drop the flag and rely on the default.

## System details

- Omarchy version: 4.0.0.r1791.ged7bae4-1
- codex-cli version: 0.149.0
- Installed via mise (`~/.local/share/mise/installs/codex/latest`)

keithnyc:
## Summary

- `omarchy-agent-usage-codex` starts `codex app-server` with `-a untrusted`, but current `codex-cli` (0.149.0+) only accepts `on-request` or `never` for `--ask-for-approval`. The subprocess now exits immediately on this invalid flag instead of speaking the RPC protocol, so the script's `initialize` request times out after 8s and the Codex tab shows "Codex limits unavailable" with the raw `TimeoutError` text (`initialize`) as the help line. Local token stats are unaffected since they come from a separate file-scan path.
- Switches the flag to `-a never`. Verified this is semantically correct, not just a value that happens to work: this process only calls the read-only `account/read` and `account/rateLimits/read` RPCs and never asks the model to run anything, so paired with `-s read-only` it makes the intent explicit — a fully non-interactive, read-only probe that can never block on an approval prompt.

Fixes #7781

## Test plan

- [x] Reproduced the bug directly: `codex -s read-only -a untrusted app-server` fails with `error: invalid value 'untrusted' for '--ask-for-approval <APPROVAL_POLICY>'`
- [x] Verified the fix end-to-end: patched script run with `-a never` (and separately with `-a on-request`, for comparison) returns real rate limits (`plan: plus`, weekly window populated) instead of timing out
- [x] `./test/all` — only pre-existing, environment-caused failures unrelated to this change (missing `omarchy-pkgs` checkout for `config-test.sh`/`unowned-system-paths-test.sh`, a permission bit on `snapper-test.sh`); confirmed these three also fail identically on `quattro` before this change

### DIFF
--- a/bin/omarchy-agent-usage-codex
+++ b/bin/omarchy-agent-usage-codex
@@ -528,7 +528,7 @@ def fetch_codex_rpc():
 
   try:
     proc = subprocess.Popen(
-      [codex, "-s", "read-only", "-a", "untrusted", "app-server"],
+      [codex, "-s", "read-only", "-a", "never", "app-server"],
       stdin=subprocess.PIPE,
       stdout=subprocess.PIPE,
       stderr=subprocess.DEVNULL,


keithnyc:
Closing as a duplicate — #7770 covers the same root cause (retired `untrusted` approval policy), was opened first, uses `on-request` with a better rationale, and adds a real regression test. Thanks for the faster fix!

PR #7783, opened 2026-08-22
Keep PwNode objects out of the audio panel's Repeater models

### LINKED ISSUE #6952: Quickshell SIGSEGV in QQuickRepeater when PipeWire removes USB audio nodes
## What happened

The Omarchy Quickshell crashed and restarted while its audio UI was being interacted with. Immediately beforehand, a USB DAC timed out and PipeWire removed/recreated its audio-node model.

## Timeline

- 16:14:31.458 — PipeWire reported the USB DAC output entering an error state:
  `Start error: Connection timed out`
- 16:14:31 — Quickshell logged the PipeWire error, cleared the default sink/source, and removed numerous `PwNode`, `PwDevice`, and metadata objects.
- 16:14:33.911 — Quickshell terminated with SIGSEGV.
- Quickshell restarted automatically.

The full stack ends in mouse-release event handling, indicating that the model change occurred while interacting with the shell UI.

## Backtrace

```text
#0  libc
#1  raise
#2  qs::crash::signalHandler at src/crash/handler.cpp:93
#4  libQt6QmlModels.so.6
#7  QV4::QObjectWrapper::getQmlProperty
#13 QV4::Module::evaluate
#17 QQmlObjectCreator::createInstance
#19 QQmlIncubatorPrivate::incubate
#22 QQuickRepeater::regenerate
#23 QQuickRepeater::setModel
#25 QQmlPropertyPrivate::write
#28 QQmlBinding::doUpdate
...
#106 QQuickMouseArea::setPressed
#107 QQuickMouseArea::mouseReleaseEvent
```

Four earlier retained Quickshell SIGSEGV dumps have the same `libQt6QmlModels` → `QQuickRepeater::regenerate()` → `setModel()` stack, on August 5, August 9 (twice), and August 12.

## Relevant shell code

`plugins/panels/audio/Panel.qml` already documents this failure mode:

> PipeWire can remove nodes while Quickshell is dispatching the removal signal; rebuilding a Repeater from that signal path has crashed in Quickshell's PipeWire service.

The panel uses delayed snapshots for `displayAudioSinks`, `displayAudioSources`, and `displayAudioStreams`, but the crash still occurred with the current implementation.

## Expected behavior

Audio devices disappearing or timing out should update the panel without crashing the entire shell.

## Observed trigger / reproduction

1. Have the Omarchy audio panel available and a USB audio device connected.
2. Interact with the audio UI.
3. Cause the device to time out or disappear while PipeWire is updating its nodes.
4. Quickshell may segfault while regenerating a QML Repeater.

The race is intermittent, but the machine has recorded the same stack repeatedly.

## System details

- Omarchy: 4.0.0-1
- Quickshell: 0.3.0, revision `28771c7c74b42e20afca0b1b63980cb46515537c`
- Package: `quickshell-git 0.3.0.r20.g28771c7-1`
- Qt: 6.11.1
- `qt6-declarative`: 6.11.1-3
- `qt6-base`: 6.11.1-1
- Signal: SIGSEGV
- Shell command: `quickshell -n -p /usr/share/omarchy/shell`
- GPU drivers present: `amdgpu` and `nvidia-drm`

Memory exhaustion was ruled out: 23 GiB was available, swap was unused, and the journal contained no OOM event.

`omarchy debug --no-sudo --print` is not available in this Omarchy release; it returns `Unknown Omarchy command`.

Filed by GPT-5 via Codex.

omarchybot:
The audio panel fed its three Repeaters snapshots of live `PwNode` objects, and Quickshell deletes a node outright the moment PipeWire drops it (`src/services/pipewire/registry.cpp:43-49`). Qt dereferences a `QObject` row without a null check in `VDMListDelegateDataType::createMissingProperties` (`qqmldmlistaccessordata_p.h:222`), so any teardown that emptied the graph with the panel open segfaulted the shell — a USB DAC unplugged, a Bluetooth profile switch, `omarchy-restart-audio`, a WirePlumber restart, or an ordinary laptop suspend. Five people reported it independently, all with the same symbolized crash site.

That Qt line is unchanged in 6.11.2, read against the installed `qt6-declarative 6.11.2-1` source rather than a changelog, so the point release does not resolve it and the shell has to stop putting QObjects in model rows.

Rows now carry the node's id and name instead of the node, and each delegate resolves its own `PwNode` through a new `nodeFor()` binding, so labels, glyphs and stream volumes stay live. The name is what pins a row to the node it came from — PipeWire reuses object ids, and `PwNode.name` is `CONSTANT` for as long as the node exists. It is the same primitives-only projection the bluetooth panel already uses against this crash class in `plugins/bluetooth/Model.js`.

Filtering the snapshot instead does not work, which is worth stating because it is the obvious thing to reach for: the node is usually still alive when the snapshot is taken and dies while Qt is incubating the delegate, so nothing visible at snapshot time identifies the row that will dangle.

On a disposable 4.0.0 VM, tearing the PipeWire graph down with the panel open crashed the shell 10 times out of 10 on `quattro` and 0 times out of 10 with this change, run in both orders against the same script. The panel renders identically before and after.

Fixes #6952

### DIFF
--- a/shell/plugins/panels/audio/Model.js
+++ b/shell/plugins/panels/audio/Model.js
@@ -18,8 +18,31 @@ function isAudioSource(node) {
     || mediaClass.indexOf("Source") !== -1
 }
 
-function listSnapshot(list) {
-  return list && list.slice ? list.slice() : []
+// Project a node list to the primitive rows a Repeater model can hold. A row
+// holding the PwNode itself is a QObject that Qt dereferences without a null
+// check while it builds the delegate, and PipeWire deletes nodes outright, so a
+// graph teardown segfaults the shell. Primitives cannot dangle, which is what
+// closes that race -- not the check below, since the node usually dies after the
+// snapshot is taken. That is also why filtering the list here never worked.
+//
+// The name rides along because PipeWire reuses object ids: without it a row
+// could resolve to whatever inherited its id. PwNode.name is CONSTANT, so it
+// identifies the node it was taken from for as long as that node lives.
+function nodeRow(node) {
+  if (!node) return null
+  var id = node.id
+  if (id === undefined || id === null) return null
+  return { id: id, name: String(node.name || "") }
+}
+
+function rowSnapshot(list) {
+  var values = list && list.length ? list : []
+  var rows = []
+  for (var i = 0; i < values.length; i++) {
+    var row = nodeRow(values[i])
+    if (row) rows.push(row)
+  }
+  return rows
 }
 
 function outputVolumeName(volume, muted) {
@@ -237,7 +260,8 @@ if (typeof module !== "undefined") {
   module.exports = {
     isPlaybackStream: isPlaybackStream,
     isAudioSource: isAudioSource,
-    listSnapshot: listSnapshot,
+    nodeRow: nodeRow,
+    rowSnapshot: rowSnapshot,
     outputVolumeName: outputVolumeName,
     parseSinkAvailability: parseSinkAvailability,
     friendlyDeviceLabel: friendlyDeviceLabel,
--- a/shell/plugins/panels/audio/Panel.qml
+++ b/shell/plugins/panels/audio/Panel.qml
@@ -100,11 +100,13 @@ Panel {
     return list
   }
 
-  // Feed Repeaters with panel-local snapshots instead of the live PipeWire
-  // model. PipeWire can remove nodes while Quickshell is dispatching the
-  // removal signal; rebuilding a Repeater from that signal path has crashed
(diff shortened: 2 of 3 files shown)


Zerantal:
Tested commit c9faea7 on Omarchy 4.0.0-1 with quickshell-git 0.3.0.r20.g28771c7-2 and Qt 6.11.2.

I applied the change to a cloned z.audio plugin, rebooted, then disconnected my Bluetooth audio device with the audio panel
open—the trigger that previously crashed Quickshell for me. The panel remained running and recovered normally; no crash
occurred.

This looks fixed on my system.

PR #7784, opened 2026-08-22
Add a Grok usage collector for the agents panel

shawnyeager:
## What

Adds `omarchy-agent-usage-grok` so the existing `omarchy.agents` panel grows a Grok tab the same way it already does for Claude, Codex, and Fireworks. No QML change: the panel draws any record that lands in `~/.local/state/omarchy/agents/usage/`.

## How

- **Local stats** come from documented Grok sessions under `$GROK_HOME` (default `~/.grok`). Each `turn_completed` event in `updates.jsonl` is one prompt. Cache reads are split out of `inputTokens` so the four panel buckets stay mutually exclusive.
- **Subagent sessions** are skipped. Grok 1.0.x writes `session_kind: subagent`; older builds used `subagent_fork`. Parent `turn_completed` already includes finished child spend, so counting the child file would double tokens and invent extra prompts.
- **Turn shapes** that some 1.0.x builds emit are also counted: `event_name: turn_completed`, and a flat token ledger on the update object itself (no nested `usage`).
- **Limits and plan** come from the same cli-chat-proxy endpoints Grok's `/usage` command uses (`/v1/billing?format=credits` and `/v1/settings`). The live product string `subscription_tier_display` is preferred (`SuperGrok Heavy` here); `SuperGrokPro` from `/v1/user` is the fallback. The collector only reads the OIDC token already stored in `~/.grok/auth.json`. It never writes credentials and does not refresh them — that stays with `grok login`.
- A SuperGrok weekly pool with a **zero prepaid ledger** is reported as a limit, not `$0.00` remaining. Prepaid / leftover on-demand credits still become a `balance` when the billing payload actually has them.

Verified on a live SuperGrok Heavy login (Grok 1.0.5): the collector printed a ready record with **SuperGrok Heavy**, a **Weekly 15%** meter, 548 prompts across 54 parent sessions, tokens by day, and tokens by model (`grok-4.5-build`, `grok-4.6-build`).

Related collectors already in review: #6902, #6485, #7393, #7200, #6477. This one is the 1.0.5 session kind (`subagent`, not only `subagent_fork`) plus `subscription_tier_display`. Happy to close in favor of whichever lands.

## Tests

```bash
./test/shell.d/agent-usage-grok-scanner-test.sh
./test/shell.d/agents-panel-test.sh
./bin/omarchy commands --check
```

Covers empty/unauthenticated records, per-turn summing and dedupe, `subagent` and `subagent_fork` skip, `GROK_HOME`, weekly/on-demand/balance mapping, SuperGrok Heavy labeling, SuperGrokPro fallback, a zero SuperGrok prepaid ledger, expired tokens, `--limits-only` cache reuse, event_name/flat ledger, and transport vs HTTP error handling.


shawnyeager:
Closing in favor of #6902, which already has the collector, community verification, and an author taking patches onto the branch.

The one extra this PR had — skip Grok 1.0.x `session_kind: subagent`, not only `subagent_fork` — is offered there instead: https://github.com/csfh/omarchy/pull/3

PR #7790, opened 2026-08-22
Add optional agent security review for plugins

max:
## Summary

- Offer an advisory review with the configured default coding agent after a plugin is cloned and validated, but before it is installed.
- Add an explicit --review flag while preserving existing --yes automation unless review is requested.
- Treat repository instructions as untrusted, identify the exact staged commit, and launch the agent from outside the plugin checkout.
- Refuse installation when the review fails, changes the checkout, or leaves the plugin invalid.
- Document the advisory nature of the review and cover success, failure, mutation, missing-agent, and non-interactive behavior.

## Security scope

This first stage is intentionally advisory, not a security guarantee. It uses Omarchy's existing default-agent launcher; a later hardening stage can add provider-specific read-only execution and stronger sandboxing.

## Testing

- Focused plugin-add shell tests
- Full CLI suite under Ubuntu 24.04 with Bash 5
- Bash syntax checks
- git diff --check

### DIFF
--- a/bin/omarchy-plugin-add
+++ b/bin/omarchy-plugin-add
@@ -2,8 +2,8 @@
 
 # omarchy:summary=Add a shell plugin from git
 # omarchy:group=plugin
-# omarchy:args=[git-url] [--enable] [--yes]
-# omarchy:examples=omarchy plugin add https://github.com/acme/omarchy-weather.git --enable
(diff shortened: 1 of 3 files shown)


bjarneo:
Automated duplication check: this pull request looks similar to #6771, which covers the same plugin agent review. I keep that one open and close this one to consolidate review.

PR #7800, opened 2026-08-22
Preserve menu position and search state when navigating back

AnthonyPoschen:
## Why it changed

i liked exploring the menu's with the keyboard and got annoyed the menu selection got reset on me and i couldn't go back a menu without backspacing an entire search. Gif shows the new flow
  
  ## What changes

  Improves keyboard navigation in the Omarchy menu:

  - Returning with `Left` or Backspace restores the previous menu’s selected row.
  - Each menu level keeps its own search query:
    - entering a submenu starts with an empty search;
    - returning restores the parent menu’s search.
  - `Left` always navigates back, including while a search query is active.
  - At the top-level menu, `Left` clears the active search and returns to the normal starting view.
  - Backspace continues to edit an active search; with an empty search it navigates back.

  ## Validation

  - `qmllint shell/plugins/menu/Menu.qml`

screenrecording-2026-08-23_04-00-17

### DIFF
--- a/shell/plugins/menu/Menu.qml
+++ b/shell/plugins/menu/Menu.qml
@@ -1137,7 +1137,7 @@ Item {
             if (root.filterText) root.setFilter("")
             else root.cancel()
             event.accepted = true
-          } else if (event.key === Qt.Key_Left) {
+          } else if (event.key === Qt.Key_Left && !root.dmenuActive) {
             root.goBack()
             event.accepted = true
           } else if (Util.editsFilter(event, root.filterText)) {
--- a/shell/plugins/menu/Menu.qml
+++ b/shell/plugins/menu/Menu.qml
@@ -1134,10 +1134,13 @@ Item {
             if (root.filterText) root.setFilter("")
             else root.cancel()
             event.accepted = true
+          } else if (event.key === Qt.Key_Left) {
+            root.goBack()
+            event.accepted = true
           } else if (Util.editsFilter(event, root.filterText)) {
             root.setFilter(Util.editedFilter(event, root.filterText))
             event.accepted = true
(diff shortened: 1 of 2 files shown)


AnthonyPoschen:
i noticed this has progressed, i am available to iterate if needed.

bjarneo:
Automated duplication check: this pull request looks similar to #7074, which covers the same menu goBack selection restore. I keep that one open and close this one to consolidate review.

AnthonyPoschen:
Reopening after the automated duplication check against #7074.

#7074 and this PR both restore the selected row when `goBack()` returns from a submenu. That is the overlap.

This PR also:

- Restores each menu level’s search query. `navStack` stores `{ id, selectedIndex, filterText }`. Entering a submenu still starts empty; Back/Left restores the parent query.
- Makes `Left` always navigate back in regular menu mode, including while a search is active. Backspace still edits the query and only navigates when it is empty.
- At the top-level menu, `Left` clears the active search and returns to the unfiltered start.
- Leaves dmenu input/select alone (`!root.dmenuActive`).

Why the extra state matters: filter root to `st`, open Style, then go back. #7074 clears the search and applies the filtered index to the unfiltered list, so you land on a different row with the query gone. This PR restores both the query and the matching row.

| Behavior | #7074 | This PR |
|---|---|---|
| Restore selected row on back | yes | yes |
| Restore that menu’s search query | no | yes |
| Left while a search is typed | inert until the query is empty | always goes back (dmenu excluded) |
| Backspace with a search | edits the query | edits the query |
| Left on the top-level menu with a search | unchanged | clears search, unfiltered start |

Happy to iterate if the smaller cursor-only change is preferred first.

AnthonyPoschen:
GitHub does not let me reopen this PR after a maintainer close, so the same branch is now in #12759 with the functional difference from #7074 in the description.

PR #7802, opened 2026-08-22
Stage and syntax-check /etc/omarchy.conf before installing it

Adolanium:
`omarchy dev link` piped `/etc/omarchy.conf` straight into `sudo tee` with no check. Every login shell sources that file through `/etc/profile.d/omarchy.sh`, so a conf bash cannot parse would break each login until root fixed it by hand.

This change builds the conf in a temp file, checks it with `bash -n`, and only then installs it with `sudo tee`. It is the same stage-and-check rule the sudoers drop-in a few lines below already follows.

Today `omarchy_conf_quote` keeps every real path parsable, so this does not fix a bug that can be hit now. It guards against a future change to how the conf is written landing a broken file on the system.

Tests: added a case to `test/shell.d/dev-link-test.sh`. It stubs `bash` so the `-n` check fails, then confirms dev link exits with an error, runs no `sudo` command, leaves the conf that was already installed as it was, and installs no sudoers drop-in. The same test fails against the script on the base branch, which has no check. All 9 dev-link tests and the 4 dev-unlink tests pass.

Earlier versions of this PR claimed a newline in the checkout path could inject commands into the conf. That was wrong: a newline inside double quotes stays part of the string, and the quoting already escapes `"`. The path rejection and its test are gone.

### DIFF
--- a/bin/omarchy-dev-link
+++ b/bin/omarchy-dev-link
@@ -89,7 +89,8 @@ done
 # read takes every rule after it down with it, including the %wheel grant, and
 # the password prompt needed to undo that is on the other side of the breakage.
 staged_sudoers=$(mktemp)
-trap 'rm -f "$staged_sudoers"' EXIT
+staged_conf=$(mktemp)
+trap 'rm -f "$staged_sudoers" "$staged_conf"' EXIT
 
 {
   printf 'Defaults secure_path='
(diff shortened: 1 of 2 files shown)


ErikMelton:
As it stands, this is not reproducable. A newline inside double quotes remains part of the string; there is no security issue here. 

The staging and "bash -n" work is reasonable though because it prevents a future serializer regression from installing a malformed login file. Please update the PR to accurately describe this, and clean up the PR itself.

  - Retain staging and syntax validation.
  - Remove the unsupported injection claim.
  - Remove the character-control rejection, as this narrows valid filesystem paths without preventing a demonstrated exploit.
  - Test that a syntax-check failure performs no privileged write and preserves the existing configuration.

The existing rejection test only proves the new policy rejects legal filesystem paths; it does not demonstrate exploit prevention.

Adolanium:
You're right, a newline inside double quotes stays part of the string. I rechecked with the exact `omarchy_conf_quote` output and the path round-trips through `source` with nothing else run. My earlier repro was wrong.

Updated the PR:

- Kept the staging and `bash -n` check.
- Removed the control-character rejection and its test.
- Rewrote the title, description, and commit message. No injection claim.
- New test stubs `bash -n` to fail and checks dev link then runs no `sudo` command, leaves the installed conf as it was, and installs no sudoers drop-in. The same test fails against the base branch script.

ErikMelton (approved):

ryanrhughes:
Appreciate you sharing this. Because this deals with the dev mode, I'm going to go ahead and close this. 

Dev mode is known to intentionally have looser controls. It is meant for active development only and no user should be running day-to-day in it.

Issue #7399, opened 2026-08-18
Lock screen re-blanks ~1s after DPMS wake on slow-sync monitor, making unlock blind typing


skinandbones:
## System details

AMD Ryzen AI MAX+ 395 (Radeon 8060S / Strix Halo, RADV), ASUS ProArt PA32QCV (6K) on DP-4, Omarchy 4.0.0-1, single monitor, no fingerprint.

## What's wrong?

After the idle lock blanks the display, waking it is nearly unusable on this monitor:

1. Screen locks (idle or manual), display blanks ~5s later — expected, per `plugins/lock/Service.qml` `idleBlankTimer` (hardcoded `interval: 5000`).
2. Pressing a key wakes the display, but the password prompt is visible for only **~1 second** before the display goes dark again.
3. Repeating input repeats the 1-second flash. The input path itself works fine: typing the password blind and pressing Enter unlocks normally.

This panel takes ~4s to resync a 6K DP signal after DPMS off, so whatever re-blanks it leaves effectively no usable window.

## What I ruled out

I cloned `omarchy.lock` and raised `idleBlankTimer.interval` from 5000 to 30000. **The symptom did not change** — the display still went dark ~1s after each wake. So the re-blank does not appear to come from the lock plugin's blank timer at all (every keystroke re-arms it via `wakeRequested` anyway).

Remaining suspects on the wake path:

- `omarchy-brightness-display on` skips the DPMS enable dispatch when `hyprctl monitors -j` reports all displays `dpmsStatus: true`. If that status flips true before the panel has actually resynced (e.g. Hyprland enabling DPMS itself on key press), the guard may be masking a failed/incomplete enable.
- 6K-over-DP link retraining after DPMS wake: the panel briefly shows a picture, then loses link and drops back to sleep — consistent with a modeset race rather than a deliberate blank.

## Timeline from journal (idle lock at 11:45, wake+unlock at 11:50)

```
15:45:36.704Z idle  lock-system: lock-timeout
15:45:36.780Z lock  lock-requested
15:45:37.303Z lock  secure=true
15:50:36.052Z idle  idle-cycle-cancel: activity   <- first key press, omarchy-system-wake runs
15:50:45.060Z lock  unlocked                      <- 9s of blind typing later
```

## Steps to reproduce

1. Omarchy 4 with a monitor that is slow to resync after DPMS off (large 6K/DSC panels).
2. Lock and wait for the display to blank.
[...]

PR #7806, opened 2026-08-22
Drop key auto-repeat in the lock screen password field

### LINKED ISSUE #7805: Lock screen password field fills with key repeats when a keypress wakes slow DPMS-blanked displays
### System details

Omarchy 4.0.0-1 (stable), Hyprland 0.56.2-1, quickshell-git 0.3.0.r20.g28771c7-2, kernel 7.1.8-arch1-3, Intel Core Ultra 5 225H (ASUS NUC15), 2× Apple Studio Display via Thunderbolt + 1× Dell via DP.

### What's wrong?

Returning to a locked machine whose lock screen has blanked (the lock plugin turns displays off via DPMS 5 s after the last activity), waking it with a **single brief tap on a character key** leaves the password field pre-filled with a long run of dots — often dozens. Enter then fails with `Authentication failed`, and you have to clear the field and retype.

### Why it happens

1. The wake keypress is delivered to the lock's password `TextInput` (`misc:key_press_enables_dpms` wakes the displays but Hyprland still passes the key through to the focused surface).
2. The DPMS wake triggers a modeset, which on slow-to-wake displays (Apple Studio Display: several seconds) stalls the compositor, so the key's **release** event reaches the client only after the modeset completes.
3. Wayland key repeat is client-side: the `TextInput` got the press, never saw the release, and types the character at the repeat rate for the whole stall. A 2–4 s stall at ~25 chars/s produces the observed pile of dots.

The dots are all the same character (invisible through masking), so the pre-filled field looks like garbage or even a stale password.

### Steps to reproduce

1. `omarchy system lock`, wait ~6 s for the screen to blank.
2. Tap one letter key once, briefly.
3. When the displays come back, the field contains many dots instead of one.

Deterministic on this machine. Control test: waking with Shift (no character) leaves the field empty, confirming the pile comes from the wake key itself.

Journal timeline of one occurrence (each keypress at the lock triggers `dpms on`, which on this machine always produces a kernel AER burst, making input activity visible in the log):

```
19:18:24 lock-requested (manual lock)
19:18:30 lock blanks — DPMS off, AER burst
         … 46 min, no further bursts ⇒ no input events while blanked …
20:04:49 single wake keypress — DPMS on, AER burst; field already full of dots
20:05:03 unlocked after clearing (Escape) and retyping
```

### Expected

Waking the blanked lock screen with a keypress leaves at most that one character in the field.

### Workaround

Wake with the mouse, Shift, or Escape (Escape also clears the field).

### Suggested fix

Ignore auto-repeat key events in the password field (except Backspace/Delete, where holding to clear is expected). A held key has no legitimate role in typing a password. PR follows.

berndb:
Fixes #7805.

Waking a DPMS-blanked lock screen with a keypress can stall the compositor for seconds while slow displays modeset (Apple Studio Display: 2–4 s). The wake key's release only reaches the lock surface after the stall, and since Wayland key repeat is client-side, the password `TextInput` types the held character at the repeat rate for the whole stall — leaving the field pre-filled with dozens of masked dots. Enter then fails and the user has to clear and retype (details and journal evidence in the issue).

A held key has no legitimate role in typing a password, so the field now drops auto-repeat key events. Backspace and Delete keep repeating, so holding Backspace to clear the field still works. Escape and Ctrl+U continue to clear in one stroke.

Verified on the affected machine: locking, letting the screen blank, and briefly tapping a letter key reproduces the flood deterministically without this change, and waking with Shift leaves the field empty — confirming the pile is client-side repeat of the wake key.

Tests: new `test/shell.d/lock-autorepeat-test.sh` covers the repeat policy via a LockView fixture (character/Return/Escape repeats dropped, Backspace/Delete kept). `./test/all` run against a live compositor: all lock tests pass; the only failing files (config, runtime-smoke, snapper, unowned-system-paths) fail identically on unmodified `quattro` on this machine, so they are pre-existing and unrelated.

### DIFF
--- a/shell/plugins/lock/LockView.qml
+++ b/shell/plugins/lock/LockView.qml
@@ -60,6 +60,14 @@ Item {
     passwordTextEdited("")
   }
 
+  // Waking a DPMS-blanked display can stall the compositor for seconds while
+  // the monitor modesets, so the wake key's release arrives late and client-side
+  // key repeat floods the field with that character. A held key has no business
+  // typing a password; only holding Backspace/Delete to clear stays useful.
+  function dropsAutoRepeat(key) {
+    return key !== Qt.Key_Backspace && key !== Qt.Key_Delete
+  }
+
   function syncPasswordText() {
     if (passwordInput.text === passwordText) return
     syncingPasswordText = true
@@ -176,6 +184,10 @@ Item {
 
         Keys.onPressed: function(event) {
           root.wakeRequested()
+          if (event.isAutoRepeat && root.dropsAutoRepeat(event.key)) {
+            event.accepted = true
+            return
+          }
           if (event.key === Qt.Key_Escape || (event.modifiers & Qt.ControlModifier && event.key === Qt.Key_U)) {
             root.passwordTextEdited("")
             event.accepted = true
--- a/test/shell.d/fixtures/lock-autorepeat/shell.qml
+++ b/test/shell.d/fixtures/lock-autorepeat/shell.qml
@@ -1,0 +1,73 @@
+import QtQuick
+import Quickshell
+import qs.Commons
+
+ShellRoot {
+  id: root
+
+  readonly property string resultPath: Quickshell.env("OMARCHY_QML_TEST_RESULT")
+  readonly property string rootPath: Quickshell.env("OMARCHY_PATH")
+  property var failures: []
+
+  function fail(message) {
+    failures.push(String(message))
+  }
+
+  function assertTrue(condition, message) {
+    if (!condition) fail(message)
+  }
+
+  function shellQuote(value) {
+    return "'" + String(value).replace(/'/g, "'\\''") + "'"
+  }
+
+  function writeResult() {
+    var payload = JSON.stringify({
+      ok: failures.length === 0,
+      failures: failures
+    })
+
+    if (resultPath) {
+      Quickshell.execDetached(["bash", "-lc", "printf '%s' " + shellQuote(payload) + " > " + shellQuote(resultPath)])
+    }
+  }
+
+  Item { id: host; width: 800; height: 600 }
+
+  Timer {
+    interval: 1
+    running: true
+    repeat: false
+    onTriggered: {
+      try {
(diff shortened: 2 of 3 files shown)


itissdr:
Found this same problem on my HP Probook 14 yesterday. Ditched my own proposed fix in favor of this one, I had not bothered to allow for backspace, etc.. Merged and tested on my hp with devlink.

ClGratton:
Confirmed the underlying failure and this fix on another slow dual-DisplayPort setup:

- Omarchy 4.0.1-1 / Hyprland 0.56.2-1
- NVIDIA RTX 3080 Ti
- two 2560×1440 high-refresh DP monitors (`DP-1` 180 Hz, `DP-3` 200 Hz)
- foot 1.27.0-2

With both outputs DPMS-off at the lock screen, one physical Space press woke them but left about seven dots in the password field. A temporary lock-service harness recorded repeated `wakeRequested` activity for roughly three seconds while Hyprland already reported both outputs `dpmsStatus=true`, consistent with the delayed-release/client-side-repeat diagnosis here.

Applying this PR's `event.isAutoRepeat` filter verbatim to a cloned `omarchy.lock` reduced the field to the single original Space, confirming the patch on this hardware.

One optional UX extension: many lock screens treat the first key after DPMS blank as wake-only. I locally tested a small state handoff where `runBlank()` arms `discardNextWakeKey`; the first non-repeat key is accepted/consumed and clears that flag, while pointer wake clears the flag so the first subsequent password character is preserved. That leaves the password field empty after Space-to-wake. This is separable from the correctness fix in this PR—the auto-repeat filter is valuable and confirmed either way.

---

**Scope correction (2026-08-28):** The auto-repeat evidence above establishes the password-field flooding mechanism only. It does **not** establish why the physical displays can briefly show the lock prompt and then go black again. On this same machine, that display symptom persisted after the input filter was applied. The lock plugin independently starts a hardcoded 5-second blank timer on wake, and slow DisplayPort link training may consume most or all of that window; this is being tested separately against #7399. Please treat the input filter in this PR as an independent, still-confirmed fix.

PR #7814, opened 2026-08-22
Add encrypted, versioned, off-site backups

achevalier-dev:
Implements `plans/backup.md` (rev 2): restic engine, hourly user timer, one state file, a bar widget, and restores that don't need flags.

- `omarchy setup backup` — gum wizard for an S3-compatible bucket, a local disk, or any restic repository URL. Probes before it creates anything (absent → init, wrong passphrase → stop, unreachable → stop), writes a recovery card, runs the first backup in front of you, and only then enables the timer and places the widget. `--remove` tears it down and says plainly that the repository itself is untouched.
- `omarchy-backup-run` — the timer's target and the only writer of `~/.local/state/omarchy/backup/status.json`. Skips (paused, battery, lock held, backup disk absent) are recorded as skips, not failures; exit 3 is a `partial` snapshot that never becomes the restore default; exit 11 is a retryable skip; a run whose pid is gone is reconciled instead of showing an eternal upload.
- `omarchy backup status|now|pause|resume|snapshots|log|restore|browse` — restores land in `~/Restored/<timestamp>/` by default; `--in-place` copies the current version aside first; `browse` mounts every backup as dated folders.
- Bar widget at `shell/plugins/panels/backup/`, watching the state file (plus its directory, since the runner replaces the file by rename). Never shells out to restic.
- Maintenance monthly, on one machine only: `forget --prune` plus `check` with a rotating `--read-data-subset`.
- Docs: `manual/52-backups.md`, `docs/backup.md`, and cross-links from the snapshots, Dropbox, and coming-from-mac pages so "snapshot vs backup vs sync" is stated once.

Verified: `test/cli` green; `test/shell` 189/192 — the three failures (`config-test`, `snapper-test`, `unowned-system-paths-test`) are the missing `omarchy-pkgs` checkout and fail identically on `quattro` here. New coverage: state-machine transitions against a stub restic, the unattended wizard path including the two abort cases, a real backup/restore round trip on restic's local backend, and Model.js under node. The widget was checked in a running shell in its unconfigured, healthy, running, failed, and paused states.

Deliberate deviations from the plan, both easy to change:

- The units ship in `default/systemd/user/` and the wizard symlinks them into `~/.config/systemd/user/` rather than being installed to `/usr/lib/systemd/user` by `omarchy-settings`. That keeps the feature working from a checkout and avoids adding `package_defaults` entries that would fail `config-test` until a PKGBUILD lands in the other repo. Happy to move it.
- Open questions 1–7 in the plan are answered conservatively: VM disks and model files stay in, cadence is hourly for everyone with no metered-connection detection, no bandwidth limit, no widget for unconfigured users, plain `omarchy backup` naming, one shared repository with per-machine ones documented rather than wizarded.

## Demo

Setup, first backup, a deleted file coming back, pause/resume, and the snapshot list — recorded against a sandbox home folder and a local repository, every command from this branch:

backup demo

([same thing as mp4](https://github.com/achevalier-dev/omarchy/releases/download/backup-demo/backup-demo.mp4))

The bar widget, open, after a healthy run:

backup panel

And its other states — not set up, running, failed, paused:

backup panel states

### DIFF
--- a/bin/omarchy
+++ b/bin/omarchy
@@ -28,6 +28,7 @@ declare -A GROUP_DESCRIPTIONS
 
 GROUP_DESCRIPTIONS[agent]="AI coding agent usage data"
 GROUP_DESCRIPTIONS[audio]="Audio input and output controls"
+GROUP_DESCRIPTIONS[backup]="Off-site encrypted backups"
 GROUP_DESCRIPTIONS[bar]="Omarchy shell bar layout and settings"
 GROUP_DESCRIPTIONS[battery]="Battery status helpers"
 GROUP_DESCRIPTIONS[bluetooth]="Bluetooth device controls"
--- a/bin/omarchy-backup-browse
+++ b/bin/omarchy-backup-browse
@@ -1,0 +1,48 @@
+#!/bin/bash
+
+# omarchy:summary=Browse every backup as dated folders and copy files out
+
+set -uo pipefail
+
+source omarchy-backup-env
+
+if ! backup_configured; then
+  echo "Backups are not set up. Run: omarchy setup backup" >&2
+  exit 1
+fi
+
+if omarchy-cmd-missing fusermount; then
+  echo "Browsing needs fuse2. Install it with: omarchy-pkg-add fuse2" >&2
+  exit 1
+fi
+
+unmount() {
+  fusermount -u "$BACKUP_MOUNT_DIR" 2>/dev/null
+  rmdir "$BACKUP_MOUNT_DIR" 2>/dev/null
+}
+
+# A browse session that was killed rather than closed leaves the mountpoint
+# behind, and restic refuses to mount over it.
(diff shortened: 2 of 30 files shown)


jordanhubbard:
I am adopting this implementation in my fork and have sent a follow-up against your `backup` branch: https://github.com/achevalier-dev/omarchy/pull/1.

It adds repository-side completion markers and a machine/snapshot/path recovery picker, preserves current files during explicit in-place recovery, and fixes credential reconfiguration, first-backup scheduling, and host-scoped retention. The System menu and backup panel both expose recovery. Real restic tests and graphical recovery/setup checks in a disposable Omarchy VM are included, with screenshots in the follow-up.

I am retiring the backup portion of my combined #10358 in favor of collaborating here; preference sharing is now independent in #12037.

PR #7821, opened 2026-08-22
Set an explicit lazygit editor preset to avoid vim fallback

jamesmcm:
At the moment lazygit e to edit file is broken in a new Omarchy install, as it fails to parse omarchy-launch-editor so falls back to vim which is not installed. See: https://github.com/jesseduffield/lazygit/blob/ea916395469808d37d726ae183fa40fd575a9d8a/pkg/config/editor_presets.go#L55-L181

Firstly we set the default config to use nvim in lazygit and add a migration for that.

Then we add support so that modifying the default editor (e.g. to emacs), updates that lazygit config - **all of the complexity is here** - I think it is worth ensuring this still works nicely for emacs users, but if we want to keep things simple we could drop this and just stick to "lazygit uses neovim always".

----------------------------

lazygit only recognizes known editor names from $EDITOR (first word only), so Omarchy's omarchy-launch-editor wrapper is never matched and lazygit falls back to vim, which isn't installed. Ship os.editPreset: nvim as the default, keep it in sync when switching editors via omarchy-default-editor, and migrate existing installs off the empty config file.

### DIFF
--- a/config/lazygit/config.yml
+++ b/config/lazygit/config.yml
@@ -1,0 +1,6 @@
+# Default lazygit editor preset. Kept in sync with the Omarchy default editor
+# by omarchy-default-editor; see https://github.com/jesseduffield/lazygit —
+# lazygit cannot parse Omarchy's EDITOR wrapper, so an explicit preset is
+# required to avoid falling back to vim.
+os:
+  editPreset: nvim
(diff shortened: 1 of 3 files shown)


jamesmcm:
Dropped trying to support Cursor too here as the LazyGit support is too complicated - we will just stick to what they support directly.

PR #7840, opened 2026-08-23
Preserve monitor position when adjusting scale

GPavlovic:
## Summary

- `set_scale()` in `bin/omarchy-hyprland-monitor-scaling` applies the new scale via `hyprctl eval` with `position = "auto"` for the focused monitor.
- On a multi-monitor layout where positions are set explicitly (e.g. a rotated secondary display placed to the right of a primary), `"auto"` re-runs Hyprland's placement algorithm and can displace the *other*, unrelated monitor -- the net effect is the two monitors swap sides after adjusting scale on just one of them.
- Fix: read the focused monitor's current `x`/`y` from `hyprctl monitors -j` and pass that through instead of `"auto"`, so a scale change never moves any monitor.

## Reproduction

Dual-monitor setup, explicit positions in `monitors.lua` (DP-1 at `0x0`, HDMI-A-1 rotated at `3440x0`):

```
$ hyprctl monitors -j | jq -r '.[] | "\(.name) pos=\(.x)x\(.y)"'
HDMI-A-1 pos=3440x0
DP-1 pos=0x0
```

Focus DP-1 and run `omarchy-hyprland-monitor-scaling 1` (same scale, isolating the position bug from any scale change):

```
$ hyprctl monitors -j | jq -r '.[] | "\(.name) pos=\(.x)x\(.y)"'
HDMI-A-1 pos=0x0
DP-1 pos=1080x0
```

DP-1 (originally on the left) is now to the right of HDMI-A-1, purely from touching its scale.

## Test plan

- [x] Added a regression test in `test/shell.d/monitor-scaling-test.sh` asserting the `hyprctl eval` call preserves a non-origin monitor position (`x=3440,y=0`) instead of using `"auto"`.
- [x] `test/shell.d/monitor-scaling-test.sh` passes (all 15 cases).

### DIFF
--- a/bin/omarchy-hyprland-monitor-scaling
+++ b/bin/omarchy-hyprland-monitor-scaling
@@ -80,13 +80,15 @@ set_scale() {
   local width="$(echo "$monitor_info" | jq -r '.width')"
   local height="$(echo "$monitor_info" | jq -r '.height')"
   local refresh_rate="$(echo "$monitor_info" | jq -r '.refreshRate')"
+  local x="$(echo "$monitor_info" | jq -r '.x')"
+  local y="$(echo "$monitor_info" | jq -r '.y')"
   local new_scale="$(clean_scale "$requested_scale" "$width" "$height")"
   # GTK only honors integer GDK_SCALE values, so persist the nearest whole
   # factor even when the monitor scale itself is fractional.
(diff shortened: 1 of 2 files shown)


GPavlovic:
Duplicate of #7495

PR #7843, opened 2026-08-23
Quote install-app and install-font names like install-and-launch

Adolanium:
`install-app`, `install-font` and `install-and-launch` build a shell command string and hand it to the floating-terminal helper, which runs it through `bash -c`. Everything interpolated into that string is parsed as shell, and none of it was quoted for that. The display name sat inside single quotes, so an apostrophe closed the quote and whatever followed ran as a command. The package arguments sat bare, so a `;` among them started a second command.

`omarchy-install-app "Foo's App" alpha` never reached `omarchy-pkg-add` at all — `bash -c` died looking for a closing quote. `omarchy install app Vim 'vim; id'` installed vim and then ran `id`.

Every value is now quoted with `printf %q` before it enters the string, which is what `install-and-launch` already did for its echo text and its desktop id. The display name, the font family and `install-font`'s package are single values and are quoted whole. The package list in `install-app` and `install-and-launch` has to reach `omarchy-pkg-add` as several words, so it cannot be quoted whole: it is split the way the unquoted expansion split it and each word is quoted on its own. Reading it with `-d ''` keeps a newline-separated list intact instead of dropping every package after the first, which a plain `read -a` would truncate.

The menu passes hardcoded package names, so none of this was reachable from the UI. But `omarchy install app` and `omarchy install font` are documented commands whose arguments are meant to be data, and they were not.

🤖 Package quoting added and the change verified by Opus 5 in Claude Code. Reviewed by Codex XHigh.

### DIFF
--- a/bin/omarchy-install-app
+++ b/bin/omarchy-install-app
@@ -15,8 +15,8 @@ fi
 printf -v install_message '%q' "Installing ${name}..."
 
 # The list has to reach omarchy-pkg-add as several words, so each word is quoted
-# rather than the whole string; -d '' keeps a newline-separated list intact.
-read -r -d '' -a package_list <<<"$packages"
+# rather than the whole string; -d '' reads past newlines and always ends at EOF.
+read -r -d '' -a package_list <<<"$packages" || true
 printf -v packages_arg '%q ' "${package_list[@]}"
 packages_arg="${packages_arg% }"
 
--- a/bin/omarchy-install-app
+++ b/bin/omarchy-install-app
@@ -12,4 +12,7 @@ if [[ -z $name || -z $packages ]]; then
   exit 1
 fi
 
-exec omarchy-launch-floating-terminal-with-presentation "echo 'Installing ${name}...'; omarchy-pkg-add ${packages}"
+printf -v install_message '%q' "Installing ${name}..."
+
+exec omarchy-launch-floating-terminal-with-presentation \
+  "echo ${install_message}; omarchy-pkg-add ${packages}"
--- a/bin/omarchy-install-app
+++ b/bin/omarchy-install-app
@@ -14,5 +14,11 @@ fi
 
 printf -v install_message '%q' "Installing ${name}..."
 
+# The list has to reach omarchy-pkg-add as several words, so each word is quoted
+# rather than the whole string; -d '' keeps a newline-separated list intact.
+read -r -d '' -a package_list <<<"$packages"
+printf -v packages_arg '%q ' "${package_list[@]}"
+packages_arg="${packages_arg% }"
+
 exec omarchy-launch-floating-terminal-with-presentation \
-  "echo ${install_message}; omarchy-pkg-add ${packages}"
+  "echo ${install_message}; omarchy-pkg-add ${packages_arg}"
--- a/bin/omarchy-install-font
+++ b/bin/omarchy-install-font
@@ -14,7 +14,8 @@ if [[ -z $name || -z $package || -z $family ]]; then
 fi
 
 printf -v install_message '%q' "Installing ${name}..."
+printf -v package_arg '%q' "$package"
 printf -v family_arg '%q' "$family"
 
 exec omarchy-launch-floating-terminal-with-presentation \
-  "echo ${install_message}; omarchy-pkg-add ${package} && sleep 2 && omarchy-font-set ${family_arg}"
+  "echo ${install_message}; omarchy-pkg-add ${package_arg} && sleep 2 && omarchy-font-set ${family_arg}"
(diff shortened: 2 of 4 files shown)


bin/omarchy-install-app
```diff
@@ -12,4 +12,7 @@ if [[ -z $name || -z $packages ]]; then
   exit 1
 fi
 
-exec omarchy-launch-floating-terminal-with-presentation "echo 'Installing ${name}...'; omarchy-pkg-add ${packages}"
+printf -v install_message '%q' "Installing ${name}..."
+
+exec omarchy-launch-floating-terminal-with-presentation \
+  "echo ${install_message}; omarchy-pkg-add ${packages}"
```
acrogenesis: The %q for the name is correct.

I did a test of the apostrophe problem on v4-0-1. This change stops that problem.

The command puts ${packages} into bash -c. The helper uses cmd="$*".

I did a test with the real program.

```bash
omarchy-install-app Vim "bash; /usr/bin/touch /tmp/omarchy-argv-audit-real/pwned-7843-packages"
```

A terminal with the title Omarchy opened.
omarchy-pkg-add bash ran. The package bash is installed.
The ; token ran touch. A file was created.

After this change, this command can run a second command:
omarchy-install-app Vim 'vim; id'

Do this:
1. Use %q for each package after you split the list.
2. Or put the packages in "$@" in a constant script.

Do not put raw ${packages} next to a quoted name.

bin/omarchy-install-font
```diff
@@ -13,5 +13,8 @@ if [[ -z $name || -z $package || -z $family ]]; then
   exit 1
 fi
 
+printf -v install_message '%q' "Installing ${name}..."
+printf -v family_arg '%q' "$family"
+
 exec omarchy-launch-floating-terminal-with-presentation \
-  "echo 'Installing ${name}...'; omarchy-pkg-add ${package} && sleep 2 && omarchy-font-set '${family}'"
+  "echo ${install_message}; omarchy-pkg-add ${package} && sleep 2 && omarchy-font-set ${family_arg}"
```
acrogenesis: This problem is the same as in omarchy-install-app.

The family is quoted. The package is not quoted.

I did a test on a live session with omarchy-install-font from this change.

```bash
omarchy-install-font "JetBrainsMono" \
  "bash; /usr/bin/touch /tmp/omarchy-argv-audit-real/pwned-7843-font-pkg; false" \
  "JetBrainsMono Nerd Font"
```

The ; token ran touch. A file was created.
The false token stopped omarchy-font-set. The font did not change.
The font is still JetBrainsMono Nerd Font.

omarchy-install-and-launch also uses omarchy-pkg-add ${packages}.

bin/omarchy-install-app
omarchybot: Confirmed, and fixed in 1cc7e8c3. Your first option is what landed: the list is split the way the unquoted expansion split it and each word is quoted on its own, so `omarchy install app Vim 'vim; id'` now hands `vim;` and `id` to omarchy-pkg-add as two arguments and runs nothing. It reads with `-d ''` rather than a plain `read -a`, because the plain form stops at the first newline and would silently drop the rest of a list.

Regression tests are in `test/shell.d/desktop-entry-launch-test.sh`, and each new assertion was checked by reverting the fix it covers and watching it fail — including one that pins the whole command string, since the substring assertions could not see a stray argument.

Checked by Opus 5 in Claude Code and by Codex XHigh, which reached the same conclusion about the package arguments independently.

🤖 Reply written by Opus 5 in Claude Code, posted from a human account.

bin/omarchy-install-font
omarchybot: Fixed in 1cc7e8c3, and you were right about install-and-launch — it carried the same bare `${packages}` and is included in the same commit.

One difference between the two: install-font's `omarchy:args` declares a single `<package>`, not a list, so there is no word-splitting to preserve and it is quoted whole rather than split. Your reproduction now arrives at omarchy-pkg-add as one argument and creates no file; the `false` never runs, so the font still gets set.

🤖 Reply written by Opus 5 in Claude Code, posted from a human account.

Adolanium: Correction: the font does not still get set. Your payload is now one argument to `omarchy-pkg-add`. pacman fails, `font-set` is skipped, nothing extra runs.

The earlier test claimed the family still got set because the mock always exits 0. I dropped that and pinned the real `&&` behavior: if pkg-add fails, font-set does not run.

omarchybot: You are right, and my sentence above was wrong. pacman rejects the payload as a package name, the `&&` short-circuits, and font-set never runs.

I checked your commit rather than taking it: flipping install-font's `&&` to `;` fails your new assertion, and the same mutation passes clean against the previous version of the test file. It catches a regression the old tests missed, so it stays as you wrote it.

🤖 Reply written by Opus 5 in Claude Code, posted from a human account.

PR #7844, opened 2026-08-23
Fix Codex agent usage limits with Codex 0.149

### LINKED ISSUE #7648: Codex 0.149 breaks Agents panel limit display

gcarrep:
Fixes #7648.

Codex 0.149 removed the `untrusted` approval policy, which made the agents collector fail before its app-server RPC handshake and display `Codex limits unavailable`. Use the supported `never` policy while retaining the read-only sandbox.

Adds a regression test that verifies the app-server invocation.

Tested: `./test/shell`

### DIFF
--- a/bin/omarchy-agent-usage-codex
+++ b/bin/omarchy-agent-usage-codex
@@ -528,7 +528,7 @@ def fetch_codex_rpc():
 
   try:
     proc = subprocess.Popen(
-      [codex, "-s", "read-only", "-a", "untrusted", "app-server"],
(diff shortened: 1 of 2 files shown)


gcarrep:
Closing in favor of the existing PRs linked from #7648.

PR #7847, opened 2026-08-23
[Security] Stop a video title from becoming the Download Video play command

acrogenesis:
The Download Video shortcut (`Alt+Shift+D`) hands the current tab URL to `yt-dlp` and attaches a notification click action of `mpv <path>`. That path came from a line-oriented `--print after_move` record that also included `%(title)s`.

A page can put newlines and tabs in its video title. `--restrict-filenames` sanitizes the saved filename, but not the printed title, so one yt-dlp record becomes several `OMARCHY_FILE` lines. The extra line can replace the path with an mpv option such as `--include=...`. Clicking "Download complete" then runs that option via the `omarchy-exec` hint (`bash -lc` → `mpv`).

Public report: https://mehmetince.net/omarchy/

## Change

- Print only `%(filepath)s` after a successful move. Toast text is derived from the filename, not page metadata.
- Ignore any printed path that is not a regular file inside the download directory (forged options, control characters, and directory escapes).
- Invoke `mpv -- <file>` so a leading-dash path cannot be parsed as an mpv option.
- Pass `--no-exec` and `--` to yt-dlp.
- Simulate `--print '%(title)s'` first and refuse the download when the title contains control characters. A hidden clip is no longer enough to land a file; the toast is "Download refused" instead of "Download complete".

## Test

`bash test/shell.d/chromium-ytdlp-test.sh` covers the real-file accept path, forged `--include` records, out-of-dir files, escaping symlinks, control characters in paths and titles, leading-dash titles, `mpv --`, and keeping a real file after a poisoned record.

`./test/cli` still passes.

### DIFF
--- a/bin/omarchy-chromium-ytdlp-host
+++ b/bin/omarchy-chromium-ytdlp-host
@@ -100,7 +100,9 @@ download_url() {
 
   # Stream the download: OMARCHY_PROG carries the percent (drives the OSD), and
   # OMARCHY_FILE and OMARCHY_TITLE (printed only after a successful move) carry the
-  # path and the title. The title is JSON-encoded so metadata cannot forge a record.
+  # path and the title. The title is JSON-encoded so metadata cannot forge a record,
+  # and the file is named after it: yt-dlp strips control characters from a filename
+  # with or without --restrict-filenames, so a record is still only ever one line.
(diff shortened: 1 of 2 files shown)


acrogenesis:
https://x.com/mdisec/status/2091293921468932420

rick:
pillbrah

Issue #4901, opened 2026-03-04
Hybrid Intel+NVIDIA: Chromium hardware acceleration requires manual workarounds


josefdc:
## System details

- **CPU:** 13th Gen Intel (Raptor Lake-P)
- **GPU (integrated):** Intel Iris Xe Graphics (RPL-P)
- **GPU (discrete):** NVIDIA GeForce RTX 3050 6GB Laptop GPU
- **Drivers:** nvidia-open-dkms 590.48.01, Mesa 26.0.1, intel-media-driver 25.3.4
- **Omarchy:** 3.4.1
- **Kernel:** 6.18.13-arch1-1
- **Chromium:** 145.0.7632.116

## Problem

On hybrid Intel+NVIDIA laptops, Chromium's hardware acceleration is broken out of the box. Enabling "Use hardware acceleration when available" in Chromium settings causes **black screen, flickering, and visual corruption**. This forces users to disable hardware acceleration entirely, losing VAAPI video decode and GPU rasterization.

The root cause is a **cross-GPU DMA-BUF incompatibility**: Hyprland's compositor runs on the NVIDIA GPU, while Chromium renders via the Intel GPU through ANGLE/OpenGL. When Chromium's GPU compositor tries to import NVIDIA DMA-BUF buffers as EGLImages on the Intel GPU, it fails with `EGL_BAD_MATCH (0x3009)`, producing hundreds of errors per second:

```
eglCreateImage failed with 0x00003009
OzoneImageBacking::ProduceSkiaGanesh failed to create GL representation
SharedImageManager::ProduceSkia: Trying to produce a Skia representation from an incompatible backing: OzoneImageBacking
```

This is the same issue reported in #3899 (same hardware: 13th Gen Intel + RTX 3050).

### 1. Set `LIBVA_DRIVER_NAME=iHD` in `~/.config/hypr/envs.conf`
**Result: Partially successful.** Without this, `vainfo` defaults to NVIDIA's NVDEC driver which has limited VAAPI support. With `iHD`, Intel's full VAAPI stack activates (60+ profiles including H264, VP9, HEVC, AV1 encode/decode).

### 2. Added VA-API flags to `chromium-flags.conf` (as proposed in #4773)
```
--enable-features=VaapiVideoDecodeLinuxGL,VaapiVideoEncoder
--enable-gpu-rasterization
--enable-zero-copy
--ignore-gpu-blocklist
```
[...]

PR #7851, opened 2026-08-23
Don't force the NVIDIA VA-API driver on hybrid-GPU systems

Suzu1Dev:
## Problem

On hybrid Intel+NVIDIA laptops, `default/hypr/nvidia.lua` sets `LIBVA_DRIVER_NAME=nvidia` for any GSP-capable NVIDIA GPU. On hybrid systems the iGPU drives the display, and Chromium's GPU process deliberately skips the NVIDIA DRM device (logs `Should skip nVidia device named: nvidia-drm`) and uses the Intel render node. With libva forced to the NVIDIA driver, VA-API cannot initialize for the Intel node: the Chromium GPU process loads `nvidia_drv_video.so` but never `iHD_drv_video.so`, so hardware video decode is dead and playback falls back to software decode. In practice, buffered 24-30fps VOD mostly survives on CPU, but 60fps live streams stutter and freeze the browser. Same root cause as the VAAPI part of #4901.

## Fix

Only force `LIBVA_DRIVER_NAME=nvidia` when the system is *not* hybrid. The new `omarchy-hw-hybrid-gpu-sysfs` detector reports hybrid when display-class PCI devices from more than one vendor are present. Like the existing `omarchy-hw-nvidia*` detectors it reads cached sysfs IDs instead of lspci (which resumes runtime-suspended GPUs and can exceed Hyprland's config reload budget), and unlike `omarchy-hw-hybrid-gpu` it never queries supergfxd, so it is always fast enough to run during config reload.

With the variable unset, libva auto-selects the driver per device (iHD for the Intel node, nvidia for the dGPU node) — verified with ffmpeg VA-API decode on both render nodes of a hybrid laptop.

`NVD_BACKEND=direct` and `__GLX_VENDOR_LIBRARY_NAME=nvidia` are unchanged: the former only matters when libva-nvidia-driver is actually used, the latter only affects GLX/XWayland clients.

## Testing

- Detector fixtures via `OMARCHY_PCI_DEVICES_PATH`: single NVIDIA → no, Intel+NVIDIA → yes, dual NVIDIA → no, Intel+AMD → yes, non-GPU only → no; real hybrid laptop (Core Ultra 9 285H + RTX 5090, Omarchy 4.0.0) → yes.
- On the affected machine, applying the equivalent session override (`LIBVA_DRIVER_NAME=iHD`) makes newly launched apps — including ones started through `uwsm app` — pick up the corrected driver, and Chromium VA-API initializes against the Intel iGPU.
- `./test/all` passes except 4 pre-existing environment-dependent failures (tests requiring an `omarchy-pkgs` checkout or a graphical session); they fail identically on unmodified master.

Refs #4901

### DIFF
--- a/bin/omarchy-hw-hybrid-gpu-sysfs
+++ b/bin/omarchy-hw-hybrid-gpu-sysfs
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+# omarchy:summary=Detect whether the computer has display GPUs from more than one vendor (hybrid graphics).
+
+# Unlike omarchy-hw-hybrid-gpu, this only inspects the hardware and never
+# queries supergfxd, so it stays fast enough to run during Hyprland's config
+# reload. Read the cached sysfs IDs rather than lspci, which reads PCI config
+# space and resumes runtime-suspended GPUs.
+pci_devices_path="${OMARCHY_PCI_DEVICES_PATH:-/sys/bus/pci/devices}"
+
+shopt -s nullglob
+
+declare -A vendors=()
+
+for device in "$pci_devices_path"/*; do
+  [[ $(< "$device/class") == 0x03* ]] || continue
+  vendors[$(< "$device/vendor")]=1
+  (( ${#vendors[@]} >= 2 )) && exit 0
+done
+
+exit 1
--- a/default/hypr/nvidia.lua
+++ b/default/hypr/nvidia.lua
@@ -3,6 +3,7 @@ local paths = require("default.hypr.paths")
 local nvidia = paths.omarchy_path .. "/bin/omarchy-hw-nvidia"
 local nvidia_gsp = paths.omarchy_path .. "/bin/omarchy-hw-nvidia-gsp"
 local nvidia_without_gsp = paths.omarchy_path .. "/bin/omarchy-hw-nvidia-without-gsp"
+local hybrid_gpu = paths.omarchy_path .. "/bin/omarchy-hw-hybrid-gpu-sysfs"
 
 -- These detectors read cached sysfs IDs rather than shelling out to lspci.
 -- lspci reads PCI config space, which resumes a runtime-suspended GPU, and on a
@@ -10,8 +11,15 @@ local nvidia_without_gsp = paths.omarchy_path .. "/bin/omarchy-hw-nvidia-without
 if o.shell_succeeds(o.shell_quote(nvidia)) then
   if o.shell_succeeds(o.shell_quote(nvidia_gsp)) then
     hl.env("NVD_BACKEND", "direct")
-    hl.env("LIBVA_DRIVER_NAME", "nvidia")
     hl.env("__GLX_VENDOR_LIBRARY_NAME", "nvidia")
+
+    -- On hybrid-GPU systems the iGPU drives the display, and apps like
+    -- Chromium deliberately skip the NVIDIA device and decode video on the
+    -- iGPU. Forcing the NVIDIA VA-API driver breaks their hardware decode
+    -- entirely, so let libva pick the driver per device instead.
+    if not o.shell_succeeds(o.shell_quote(hybrid_gpu)) then
+      hl.env("LIBVA_DRIVER_NAME", "nvidia")
+    end
   elseif o.shell_succeeds(o.shell_quote(nvidia_without_gsp)) then
     hl.env("NVD_BACKEND", "egl")
     hl.env("__GLX_VENDOR_LIBRARY_NAME", "nvidia")
--- a/test/shell.d/hw-hybrid-gpu-sysfs-test.sh
+++ b/test/shell.d/hw-hybrid-gpu-sysfs-test.sh
@@ -1,0 +1,63 @@
+#!/bin/bash
+
+set -euo pipefail
+
+source "$(dirname "$0")/base-test.sh"
+
+detector="$ROOT/bin/omarchy-hw-hybrid-gpu-sysfs"
+
+TMPDIR=$(mktemp -d)
+trap 'rm -rf "$TMPDIR"' EXIT
+
+# Builds a fake sysfs PCI tree from vendor:class pairs.
+make_devices() {
+  local dir="$1" spec i=0
+  shift
+
+  for spec in "$@"; do
+    local device="$dir/0000:00:0$i.0"
+    mkdir -p "$device"
+    echo "${spec%%:*}" > "$device/vendor"
+    echo "${spec##*:}" > "$device/class"
+    (( i += 1 ))
+  done
+}
+
+assert_hybrid() {
+  local description="$1"
+  shift
+
+  local dir
+  dir=$(mktemp -d -p "$TMPDIR")
+  make_devices "$dir" "$@"
+
+  if OMARCHY_PCI_DEVICES_PATH="$dir" "$detector"; then
+    pass "$description"
+  else
+    fail "$description"
+  fi
+}
+
+assert_not_hybrid() {
+  local description="$1"
+  shift
+
+  local dir
+  dir=$(mktemp -d -p "$TMPDIR")
+  make_devices "$dir" "$@"
+
+  if OMARCHY_PCI_DEVICES_PATH="$dir" "$detector"; then
+    fail "$description"
+  else
+    pass "$description"
+  fi
+}
+
+assert_hybrid "detects Intel iGPU + NVIDIA dGPU as hybrid" 0x8086:0x030000 0x10de:0x030200
+assert_hybrid "detects AMD iGPU + NVIDIA dGPU as hybrid" 0x1002:0x030000 0x10de:0x030200
+assert_hybrid "detects Intel + AMD as hybrid" 0x8086:0x030000 0x1002:0x030000
+
+assert_not_hybrid "single NVIDIA GPU is not hybrid" 0x10de:0x030200
+assert_not_hybrid "two NVIDIA GPUs are not hybrid" 0x10de:0x030200 0x10de:0x030000
+assert_not_hybrid "non-display devices are ignored" 0x8086:0x060000 0x10de:0x0c8000
+assert_not_hybrid "empty device tree is not hybrid"


KhizarImran:
Confirming this on an **AMD iGPU + NVIDIA** machine — the issue, this PR, and #4901 are all
Intel + NVIDIA, so this is a different pairing of the same fault.

Razer Blade 14: Ryzen 9 5900HX / Cezanne iGPU (`1002:1638`, drives `eDP-1`) + RTX 3070 Laptop
(`10de:249d`, GSP-capable, so `omarchy-hw-nvidia-gsp` matches and the branch fires).
Omarchy 4.0.1-1, Hyprland 0.56.2, Mesa 26.2.1, nvidia-open 610.57.04.

`omarchy-hw-hybrid-gpu-sysfs` returns true here as written — two display-class vendors
(`0x1002`, `0x10de`) — so this PR does the right thing on AMD hybrids too.

Setting `LIBVA_DRIVER_NAME=radeonsi` fixed decode: `brave://gpu` went from software to
`Video Decode: Hardware accelerated`, `GL_RENDERER` became
`ANGLE (AMD, AMD Radeon Graphics (radeonsi renoir ACO), Mesa 26.2.1)`, and the full VCN
profile list appeared (H.264 baseline/main/high, VP8, VP9 profile0/2, HEVC main/main10,
AV1 main). Letting libva auto-select, as this PR does, is the better form of that.

## One caveat worth recording

The description says `__GLX_VENDOR_LIBRARY_NAME=nvidia` can stay because it "only affects
GLX/XWayland clients". That's correct as far as it goes, but on a Wayland-first setup the
variable that decides which GPU a browser renders on is the **EGL** vendor, and Omarchy
doesn't set it at all — so glvnd still enumerates `10_nvidia.json` (priority 10) ahead of
`50_mesa.json`.

With the VA-API fix applied and GLX pointed at mesa, Brave's GPU process on this machine
still held **559 MiB and ~26 W on the dGPU**, with `runtime_status: active` at 0% utilisation,
despite Chromium having already picked the AMD node itself
(`--render-node-override=/dev/dri/renderD129`). Its open fds showed both cards:

```
9 × /dev/nvidia0   5 × /dev/nvidiactl   2 × /dev/nvidia-uvm
3 × /dev/dri/renderD128 (NVIDIA)        5 × /dev/dri/renderD129 (AMD)
```

Adding a Mesa EGL override on top of this PR is what actually settled the stutter. So this
PR and #5312 look complementary rather than alternatives — decode and render are two
separate bindings, and fixing only one leaves the dGPU in the path.

## And a trap, if anyone is tempted to fold the EGL half in here

Setting `__EGL_VENDOR_LIBRARY_FILENAMES=…/50_mesa.json` **globally** (via `hl.env` in
`hyprland.lua`, or in `nvidia.lua`) breaks dGPU-driven outputs. My external monitor hangs off
the NVIDIA card (`card1-DP-3`) while the panel is on AMD. Restricting EGL session-wide strips
NVIDIA EGL from the compositor too, and aquamarine cannot build a renderer for that output:

```
ERR from aquamarine ]: [EGL] eglInitialize errored out with EGL_NOT_INITIALIZED: DRI2: failed to load driver
ERR from aquamarine ]: CDRMRenderer: fail, eglInitialize failed
ERR from aquamarine ]: drm: Failed to initialize renderer backend for blitting
DEBUG from aquamarine ]: drm: Modesetting DP-3 with 3440x1440@100.00Hz
```

The monitor modesets and reports `disabled: false`, but stays black. `eglinfo` with the
restriction shows Mesa devices only; without it the RTX 3070 is there. #5312's per-app
`.desktop` + wrapper approach avoids this, which is presumably why it went that way — worth
keeping the scoping deliberate if the two ever get merged.

Two smaller notes that cost me time:

- `/proc/<hyprland-pid>/environ` does **not** show config `env=` values — Hyprland `setenv()`s
  them into itself after exec, and `/proc/…/environ` only reflects the exec-time environment.
  Their absence there proves nothing. (Same caveat @emshiarla raised for Chromium in #8328.)
  A running Hyprland also can't un-set them via `hyprctl reload`; it needs a fresh session.
- Anything rewriting `Exec=` must keep the first token a real executable —
  `omarchy-launch-webapp` extracts it with `sed -n 's/^Exec=\([^ ]*\).*/\1/p'`, so an
  `env VAR=… browser` prefix resolves to `env` and every web app breaks.

rdjperron:
Confirming on a third Intel+NVIDIA pairing, with **Google Chrome 151** rather than Chromium, and with numbers from an automated A/B run.

**Machine:** Microsoft Surface Laptop Studio — Intel TigerLake Iris Xe `8086:9a49` (i915, drives `eDP-1`, the only GPU Aquamarine registers: `drm: Found 1 GPUs … card1 … i915 becomes primary`) + RTX 3050 Ti Mobile `10de:25a0` (PCI class `0x0302`, no outputs). Omarchy 4.0.1-1, Hyprland 0.56.2, nvidia-open-dkms 610.57.04, Mesa 26.2.1, libva 2.24.1, libva-nvidia-driver 0.0.17, intel-media-driver 26.2.4, Google Chrome 151.0.7922.173 with the stock `chrome-flags.conf` (native Wayland).

`bin/omarchy-hw-hybrid-gpu-sysfs` from this PR returns hybrid here (vendors `0x8086` + `0x10de`), and the fixture behaviour matches the PR description (single-vendor fixtures → not hybrid).

**What the stock env does here** (`--enable-logging=stderr --v=1`, `NVD_LOG=1`): Chrome's GPU process is ANGLE on `Mesa Intel(R) Iris(R) Xe Graphics (TGL GT2)` and its VaapiWrapper opens the Intel render node, but libva loads `nvidia_drv_video.so` because of the export; nvidia-vaapi-driver logs `Passed in DRM FD does not belong to the NVIDIA driver, ignoring … Found NVIDIA GPU 0 at /dev/dri/renderD129` and decodes on NVDEC. Every decoded frame then fails to import on the Intel side:

```
ImageEGL.cpp:118 (operator()): eglCreateImage failed with 0x00003009
ui/gl/scoped_egl_image.cc:23] Failed to create EGLImage
ui/ozone/common/native_pixmap_egl_binding.cc:76] Unable to initialize binding from pixmap
media: VideoDecoderPipeline |decoder_| Initialize() failed → "video decoder fallback after initial decode error"
```

**A/B on reddit.com** (25 s scripted wheel-scroll via DevTools protocol, fresh profile, everything identical except the variable):

| | `LIBVA_DRIVER_NAME=nvidia` (4.0.1 default) | unset / `iHD` (this PR's effect) |
|---|---|---|
| `eglCreateImage failed with 0x00003009` | 1 173 | 0 |
| longest requestAnimationFrame gap (main-thread stall) | 38 757 ms | 583 ms |
| H.264 players decoded in hardware (`VaapiVideoDecoder`) | 0 — all fell back to software | 79 |

Same result with the variable unset as with `iHD` — libva picks `iHD_drv_video.so` from the Intel fd on its own, so letting it auto-select (as this PR does) is sufficient here. Manually launching Chrome with the override fixed x.com and reddit.com playback and scrolling on the real profile.

Two notes for reviewers:

- This is a 4.0.1 regression in practice: before #6939 fixed `o.shell_succeeds()`, `nvidia.lua`'s detection never fired on these machines, so hybrid laptops never received the export (same conclusion as #8215).
- Re the EGL-vendor caveat raised above: on this machine Chrome's EGL stays on Mesa/Intel under native Wayland (`chrome://gpu` GL_RENDERER `ANGLE (Intel, Mesa …)` with no override), so the VA-API half alone is enough here. Machines whose Chrome EGL lands on NVIDIA would still need the per-app Mesa-EGL override from #4901/#5312 — agree that must stay per-app, never session-wide.

Related: #8215, #8328, #8726, #4901.

---

_Investigation and this comment were authored with Claude Fable 5 in the Claude Code harness; measurements were taken on the machine described above and the fix was verified manually by the account owner._

klecu:
Confirmed on another Intel + NVIDIA hybrid, with the original failure occurring specifically at YouTube's ad-to-main-video transition.

**Machine:** HP Victus 15-fa0xxx — Alder Lake-P UHD `8086:46a3` (i915; both active displays are attached to Intel) + RTX 3050 Ti Mobile `10de:25a0`. Omarchy 4.0.1-1, Hyprland 0.56.2, Chromium 151.0.7922.173 on native Wayland.

On the stock 4.0.1 environment, both Chromium and Brave could render ads, but the whole window became unresponsive when the main video loaded. The GPU logs showed the same cross-device import failure described here:

```
eglCreateImage failed with 0x00003009
Unable to initialize binding from pixmap
OzoneImageBacking::ProduceSkiaGanesh failed to create GL representation
Trying to produce a Skia representation from an incompatible backing: CompoundImageBacking
```

Validation of this PR's exact policy:

- `bin/omarchy-hw-hybrid-gpu-sysfs` exits 0 on the real machine and sees display-class vendors `0x8086` + `0x10de`.
- Synthetic fixtures passed for single NVIDIA (not hybrid), Intel+NVIDIA (hybrid), AMD+NVIDIA (hybrid), NVIDIA non-display-class only (not hybrid), and a missing vendor file (not hybrid).
- I launched Chromium with only `LIBVA_DRIVER_NAME` unset while deliberately retaining `__GLX_VENDOR_LIBRARY_NAME=nvidia` and `NVD_BACKEND=direct`, matching this PR rather than the broader local workaround.
- Two ads rendered, then the 1280x720 main video crossed the formerly failing transition and advanced continuously from 3.90 s to 33.97 s. It rendered 2,043 frames with 2 dropped frames.
- The resulting GPU log contains none of the EGL/pixmap/shared-image errors above.

This confirms that unsetting only the forced VA-API driver is sufficient on this Intel-rendered native-Wayland configuration; a session-wide EGL override is not needed here.

megascan:
Confirming the same `LIBVA_DRIVER_NAME=nvidia` hybrid failure on another Intel+NVIDIA laptop, with a slightly different symptom: Chromium/Electron **does** initialize NVIDIA VA-API, then imports the decoded frames into the Intel GL context as garbage (red/black snow) instead of falling back to software.

**Machine:** ASUS TUF Gaming F17 FX707VV4 — Raptor Lake-P Iris Xe `8086:a7a0` (i915, drives `eDP-2`) + RTX 4060 Max-Q `10de:28e0` (GSP, drives HDMI). Omarchy 4.0.1-1, Hyprland 0.56.2, Chromium/Helium 151, Discord Electron 42, nvidia-open-dkms 610.57.04, `libva-nvidia-driver` 0.0.17, `intel-media-driver` 26.2.4. `supergfxctl` Hybrid. GPU processes for Helium and Discord both use `--render-node-override=/dev/dri/renderD129` (Intel).

Stock `nvidia.lua` export makes X.com / Discord tweet embeds / in-browser video render as NV12-as-RGB static. Setting `LIBVA_DRIVER_NAME=iHD` (user `hl.env` override after Omarchy defaults) fixes newly launched apps. Same root cause as this PR: session-wide NVIDIA VA-API on a compositor/client that is rendering on the iGPU.

`omarchy-hw-hybrid-gpu-sysfs` as written would match here (`0x8086` + `0x10de`). Unsetting the force, rather than hardcoding `iHD`, is the right policy so AMD hybrids still auto-select `radeonsi`.

Not opening a duplicate PR — this one already has the detector + `nvidia.lua` change.

Suzu1Dev:
Rebased onto the latest `quattro` and added fixture-based coverage for the detector (`test/shell.d/hw-hybrid-gpu-sysfs-test.sh`): Intel+NVIDIA, AMD+NVIDIA, and Intel+AMD report hybrid; single-NVIDIA, dual-NVIDIA, non-display devices, and empty trees do not.

devon-lyons:
Confirming on an AMD+NVIDIA hybrid, and adding a symptom beyond broken/black video: `LIBVA_DRIVER_NAME=nvidia` also corrupts **general page rendering** in Chromium browsers, because Chromium uses VA-API for accelerated **JPEG/image decode**, not just video.

## System

- ASUS ROG Zephyrus G16, Ryzen AI 9 HX 370 / Radeon 890M (drives all displays) + RTX 4070 Max-Q (no displays)
- nvidia-open-dkms 610.57.04, Mesa 26.1.8/26.2.1 (both affected), kernel 7.1.9-arch1-2, Omarchy 4.0.2

## Symptom

On image-heavy sites (amazon.ca product grids, youtube.com), Chrome/Brave/Chromium show patchy rectangular corruption while scrolling. Tiles render as uninitialized GPU memory — at one point the page content area displayed the desktop wallpaper texture (stale VRAM). The browser GPU process logs floods of:

```
ERROR:ui/ozone/common/native_pixmap_egl_binding.cc:76] Unable to initialize binding from pixmap
ERROR:gpu/command_buffer/service/shared_image/ozone_image_backing.cc:317] OzoneImageBacking::ProduceSkiaGanesh failed to create GL representation
ERROR:gpu/command_buffer/service/shared_image/shared_image_manager.cc:269] SharedImageManager::ProduceSkia: Trying to produce a Skia representation from an incompatible backing: CompoundImageBacking
```

The browser renders on the AMD render node (`--render-node-override=/dev/dri/renderD129`), but with libva forced to `nvidia_drv_video.so`, VA-API-decoded buffers are NVIDIA-allocated and cannot be bound into the AMD EGL context.

## Bisect

Identical Chromium launch on amazon.ca, only the environment varied:

| Environment | EGL binding errors |
|---|---|
| Session default (`LIBVA_DRIVER_NAME=nvidia`) | 2,754 |
| Only `LIBVA_DRIVER_NAME` unset | 0, rendering clean |

Ruled out along the way: Mesa 26.2.1 (downgraded full stack to 26.1.8 — no change), Chromium version (151 and 152 both affected), Hyprland damage tracking, explicit sync (`--disable-features=WaylandLinuxDrmSyncobj` — no change), `AMD_DEBUG=nodcc` — no change. `--disable-gpu` was clean, which pointed at the GPU buffer path.

## Workaround

User-level override in `~/.config/hypr/hyprland.lua`, loaded after the Omarchy defaults:

```lua
hl.env("LIBVA_DRIVER_NAME", "radeonsi")
```

Agreed with the direction here and in #8989/#9483: nvidia.lua should only export the NVIDIA VA-API/GLX env when the NVIDIA GPU actually drives the session, and on iGPU-primary hybrids should point libva at the iGPU's driver (`radeonsi`/`iHD`).

ChaseC-130:
Confirming on a second **AMD iGPU + NVIDIA** pairing, measured at the decode engine rather than from logs.

**Machine:** ASUS ROG Zephyrus G14 GA402XV — Radeon 780M / Phoenix1 `1002:15bf` (amdgpu, drives the only connected output, `card2-eDP-2`) + RTX 4060 Laptop `10de:28e0` (GSP-capable, so `omarchy-hw-nvidia-gsp` matches and the branch fires; its three connectors `card1-DP-1`, `card1-eDP-1`, `card1-HDMI-A-1` all read `disconnected`). Omarchy 4.0.2-1, Hyprland 0.56.2, kernel 7.1.9, Mesa 26.2.1, nvidia-open-dkms 610.57.04, libva 2.24.1, libva-nvidia-driver 0.0.17, Google Chrome 152.0.7977.82 on native Wayland with the stock `chrome-flags.conf`.

`bin/omarchy-hw-hybrid-gpu-sysfs` returns hybrid here (`0x1002` + `0x10de`), so this PR suppresses the pin correctly on this machine.

**A/B.** Same page (local loop of autoplaying H.264), same binary, fresh profile, only the variable differing. `nvidia-smi --query-gpu=utilization.decoder`, 8 samples 1s apart:

| `LIBVA_DRIVER_NAME` | NVDEC utilisation |
|---|---|
| `nvidia` (4.0.2 default) | `9 34 34 41 0 41 44 0` |
| unset (this PR's effect) | `0 0 0 0 0 0 0 0` |
| `radeonsi` | `0 0 0 0 0 0 0 0` |

Chrome had already selected the AMD node itself (`--render-node-override=/dev/dri/renderD129`) in every run, so under the default the decoder and the compositor sit on different GPUs.

**Unsetting is sufficient and safe here** — libva's per-device auto-selection gives the right answer on both nodes. Calling `vaInitialize()` directly, with no browser involved:

```
renderD129 (amdgpu)  LIBVA_DRIVER_NAME=nvidia  -> VA-API NVDEC driver [direct backend]
renderD129 (amdgpu)  unset                     -> Mesa Gallium driver 26.2.1 ... (radeonsi, phoenix, ACO)
renderD128 (nvidia)  unset                     -> VA-API NVDEC driver [direct backend]
```

The first line is the fault in isolation: the export overrides libva even when the caller opened an AMD device. The third is why dropping it costs nothing on NVIDIA-only machines.

**One measurement caveat, as it bears on how #9812 verified.** `nvidia_drv_video.so` stays mapped in Chrome's GPU process in *all three* configurations above — Chrome probes every VA device at GPU-process startup to enumerate profiles, so the library is loaded whether or not it decodes anything. Checking which `_drv_video.so` is mapped, or counting `/dev/nvidia*` fds (1 in every run here), reports a false positive on this machine. `utilization.decoder` was the only signal that separated the cases.

**On the EGL half.** After this change Chrome's GPU process here still maps `libEGL_nvidia.so` and keeps one fd on `renderD128`, consistent with @KhizarImran's point that decode and render are separate bindings. I can't corroborate the wattage though: `power.draw` is not trustworthy on this machine — I measured 1.4W with Chrome running and 16.5W with no Chrome at all — so I have no reliable figure for what the EGL binding costs here.

**Detector comparison.** Four open PRs propose four detectors; all four return the correct verdict on this hardware, well inside the config-reload budget:

```
omarchy-hw-hybrid-gpu-sysfs      (#7851)  hybrid: yes           8ms  -> suppressed
omarchy-hw-nvidia-only           (#9138)  nvidia-only: no       8ms  -> suppressed
omarchy-hw-nvidia-drives-display (#9483)  drives-display: no    6ms  -> suppressed
omarchy-hw-nvidia-only           (#9812)  nvidia-only: no       8ms  -> suppressed
```

One difference worth flagging for whoever picks between them, from reading the code rather than from measurement: this laptop's HDMI port is wired to the dGPU (`card1-HDMI-A-1`). Plugging an external monitor into it would make #9483's `omarchy-hw-nvidia-drives-display` return true, re-pinning VA-API to NVIDIA and reintroducing this fault on the internal AMD-driven panel. The presence-based detectors in this PR, #9138 and #9812 are not sensitive to that.

**Follow-up:** subjective confirmation from the machine owner — with `LIBVA_DRIVER_NAME` overridden the scroll flicker on Reddit/X is gone, on the same profile and sites that reproduced it reliably before.

---

_Investigated with Claude Opus 5 in Claude Code; all measurements were taken on the machine described above._

PR #7862, opened 2026-08-23
Add a native Pomodoro timer to the shell

syskey8:
Add a native Pomodoro timer to the shell

The bar gains a Pomodoro timer that tracks standard intervals: a 25-minute focus phase, 5-minute short breaks, and a 15-minute long break after four completed sessions. Timer state is stored in a shared global service, ensuring the countdown stays perfectly synchronized across multiple monitors. It uses absolute wall-time deadlines rather than interval ticks, so the remaining time stays accurate even if the system suspends or the shell event loop stalls.

A custom timer mode allows users to specify an exact number of minutes. Skipping a phase cleanly advances the timer without playing an alarm or erroneously counting the session as completed. Desktop notifications and standard freedesktop alarm sounds announce when it is time to switch focus. The widget is fully integrated with the Omarchy Settings app via a configurable schema.

### DIFF
--- a/install/omarchy-base.packages
+++ b/install/omarchy-base.packages
@@ -116,6 +116,7 @@ tensaku
 sddm
 slurp
 socat
+sound-theme-freedesktop
 starship
 sushi
 system-config-printer
--- a/migrations/1787487615.sh
+++ b/migrations/1787487615.sh
@@ -1,0 +1,2 @@
+echo "Install freedesktop sound theme for Omarchy plugin alarms"
+omarchy-pkg-add sound-theme-freedesktop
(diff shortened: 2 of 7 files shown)


syskey8:
@copilot review

syskey8:
@copilot review

syskey8:
@copilot review

syskey8:
@copilot review

syskey8:
@copilot review

syskey8:
@copilot review

syskey8:
@copilot review

syskey8:
@copilot review

syskey8:
@copilot review

aayushdutt:
Shouldn't this be a plugin instead? https://plugins.omarchy.org/

getaaron:
I think this looks really cool but not sure it should be native - could be moved to be a plugin.

syskey8:
I've moved this to a plugin instead, which fits the existing extension pattern better: [link](https://plugins.omarchy.org/plugin.html?id=syskey8.pomodoro). Closing this in favor of that.

PR #7863, opened 2026-08-23
Add optional idle.blank: power off the display without locking

eugenefvdm:
## Summary

Blanking the display has always been wired exclusively behind the session lock (`lock/Service.qml`'s `armBlankTimer`, only reachable from `beginLock()`). There's currently no way to turn the screen off on idle without also password-protecting the session — even though the pieces to do it independently (`hyprctl dispatch 'hl.dsp.dpms(...)'` via `omarchy-brightness-display off`) already exist and work fine on their own.

This adds a third, optional `idle.blank` key (seconds since idle began, alongside the existing `screensaver`/`lock`) that independently blanks the display and keyboard backlight without requesting a lock:

```json
{
  "version": 1,
  "idle": {
    "screensaver": 900,
    "blank": 1200,
    "lock": 86400
  }
}
```

- Omitting `idle.blank` keeps today's behavior exactly — blanking stays tied to the lock, byte-for-byte the same as before.
- New `bin/omarchy-system-blank` command (mirrors the existing `omarchy-system-wake`); the lock plugin's own blank step now calls it too, so there's one implementation instead of two duplicated inline commands.
- When lock is the earliest deadline, its own post-lock timer keeps owning the blank — the new idle-driven blank timer is skipped in that case so the two paths never race or double-fire.
- Any input wakes the display the same way it already cancels the screensaver or an in-progress lock (reuses the existing `omarchy-system-wake` path).

## Why

I hit this on my own machine wanting "screensaver at 15 min, display off at 20 min, no password lock" — a fairly common ask for a desktop that isn't shared. Today the only way to get there is bypassing the shell's idle system entirely and reinstalling `hypridle` alongside it with `idle.screensaver`/`idle.lock` parked at a no-op value, which is exactly the kind of thing `omarchy-shell`'s idle service is supposed to make unnecessary.

## Test plan

- [x] `./test/cli` — 116/116 pass
- [x] `./test/shell` — same result as unmodified `upstream/quattro` (4 pre-existing environmental failures unrelated to this change, confirmed via `git stash`); new `idle-test.sh` cases for `IdleModel.blankEnabled` pass
- [x] `omarchy system --help` lists the new `blank` command correctly under the `system` group
- [ ] Manual verification in a running session (screensaver/blank/lock sequencing, wake-on-input) — happy to do this if a maintainer wants it before review, just flagging it wasn't run in the graphical acceptance VM

### DIFF
--- a/bin/omarchy-system-blank
+++ b/bin/omarchy-system-blank
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+# omarchy:summary=Turn off displays and keyboard backlight for power saving
+# omarchy:group=system
+# omarchy:name=blank
+# omarchy:examples=omarchy system blank
+
+omarchy-brightness-keyboard off
+omarchy-brightness-display off
--- a/docs/omarchy-shell.md
+++ b/docs/omarchy-shell.md
@@ -168,7 +168,7 @@ Rules:
 5. Third-party enabled ⇔ present; for full bar options that means `bar.id`.
    First-party non-bar plugins are enabled unless listed in `disabledPlugins[]`.
 6. `barWidget.allowMultiple: true` in the manifest permits multiple instances.
-7. `idle.screensaver` and `idle.lock` are seconds since user idle began.
+7. `idle.screensaver` and `idle.lock` are seconds since user idle began. `idle.blank` is optional and, when set, independently blanks the display (and keyboard backlight) at its own idle deadline without engaging the session lock — omit it to keep blanking tied to `idle.lock` as before.
 8. `version: 1` is required.
 
 `config/omarchy/shell.json` describes the fresh-install state. When no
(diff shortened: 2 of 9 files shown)


eugenefvdm:
Addressed all 5 points from the Copilot review in f8f39fb4:

1. **`blankEnabled` validation** — now applies the same finite/non-negative check as `secondsFromConfig`, so an invalid `idle.blank` value is rejected instead of silently becoming a 0-second deadline.
2. **`blank === lock` race** — the guard now compares absolute deadlines (`blankTimeoutSeconds < lockTimeoutSeconds`) instead of `lockDelaySeconds !== 0`, so it holds even when screensaver is the earliest deadline and blank/lock coincide.
3. **Visual verification** — cloned the idle/lock plugins into a running `omarchy-shell` via `omarchy plugin clone` and confirmed `omarchy-system-blank`/`omarchy-system-wake` correctly toggle real monitor DPMS state, and that the patched service's computed screensaver/lock/blank delays are correct for a live config. I wasn't able to observe the full timer-triggered sequence end-to-end — an idle-inhibit lock held by an unrelated running app on that machine (pre-existing, unrelated to this change) kept the compositor from ever reporting idle during the test window.
4. **Stale `shell/README.md` reference** — updated its duplicated idle rules list to document `idle.blank`.
5. **Hot-reload race on `blankTimer`** — `onTriggered` now rechecks `blankConfigured`, so removing `idle.blank` from `shell.json` mid-countdown can no longer fire a stray blank.

New test cases cover the `blankEnabled` validation fix.

PR #7886, opened 2026-08-23
Detach 1Password from the installer terminal

### LINKED ISSUE #7870: omarchy-install-service-1password launches 1Password without setsid, so it dies on SIGHUP when the installer terminal closes
### What happens

`omarchy-install-service-1password` opens 1Password when it finishes, but the app is terminated about a second later — as soon as the installer terminal closes. On a first install this lands in the middle of first-run initialization.

### Cause

`omarchy-install-service-1password:31` backgrounds the launch without detaching it from the terminal:

```bash
uwsm-app -- 1password >/dev/null 2>&1 &
```

`uwsm-app` does not start a new session, so the app keeps the installer terminal's controlling pty. `omarchy-show-done` then waits for a keypress; when the terminal exits, the pty hangs up and 1Password receives SIGHUP. Chromium treats SIGHUP as a graceful-shutdown request, so the app quits.

Reproduced with an instrumented process launched exactly this way:

```
uwsm-app -- victim.sh &            → tty=pts/3, receives SIGHUP when the terminal exits, dies
setsid uwsm-app -- victim.sh &     → tty=?,     survives
```

Being in its own systemd scope (`app-1password-<pid>.scope`) does not help — the pty association is independent of the cgroup.

Every other installer that opens an app already does this correctly, including the shared helper `omarchy-install-and-launch`:

```bash
setsid uwsm-app -- gtk-launch steam >/dev/null 2>&1 &   # steam, zed, signal, spotify, vscode, emacs, heroic, chatgpt
```

### Observed sequence

```
12:33:25  pacman installs 1password 8.12.32-33
12:33:25  script runs: uwsm-app -- 1password >/dev/null 2>&1 &
          first run: DB schema migrated 0 -> 61, native-messaging manifests written
12:33:26  installer terminal scope ends
12:33:26.044  1Password log: Front end event: window closed
12:33:26.045  1Password log: app will quit / shutdown requested! / starting app shutdown
12:33:26  kernel: traps: 1password[52271] trap int3 ip:558b7c70e195 in 1password[70ea195,558b77ca5000+9aa1000]
12:33:26  systemd-coredump: Process 52271 (1password) terminated abnormally with signal 5/TRAP
```

Relaunching from the app launcher works fine — that instance has no controlling terminal.

### On the SIGTRAP specifically

Per #7822, 1Password emits a Crashpad SIGTRAP as part of its normal shutdown, so the SIGTRAP here is probably incidental rather than a fault, and the notification that surfaced it is that issue's false-positive. **The bug being reported here is that the app is shut down at all**, not that it trapped.

Two recorded details differ from #7822 and I could not resolve them, so I'm flagging rather than claiming: this coredump has `si_code: SI_KERNEL` where #7822 reports `TRAP_BRKPT`, and this instance's log ends mid-teardown at `running shutdown hook: ffi/op-core-node/src/lib.rs:477` without the `Failed to prepare for requesting dump` line #7822 quotes. That may just be log truncation. No core is available either way — 1Password sets `RLIMIT_CORE=0`.

### Expected

A freshly installed 1Password stays open after the installer terminal closes.

### Fix

```bash
setsid uwsm-app -- 1password >/dev/null 2>&1 &
```

`omarchy-install-service-dropbox:12` has the same omission (`uwsm-app -- dropbox-cli start &>/dev/null &`); lower impact since `dropbox-cli start` daemonizes, but it's the same latent bug.

This is the same class as #6762 (a terminal being torn down under a process Omarchy spawned), fixed there by reordering; here the fix is detaching.

### System

omarchy 4.0.0-1, Linux 7.1.8-arch1-3, Hyprland/uwsm, 1password 8.12.32-33, 1password-cli 2.35.0-1.

---

kx0101:
Closes #7870

Launches 1Password through `setsid` after installation so closing the installer terminal does not send SIGHUP to the app. Adds a regression test that runs the real installer with stubbed dependencies and asserts the exact detached launch invocation.

Tests:
- `bash test/shell.d/launch-1password-test.sh`
- `bash -n bin/omarchy-install-service-1password test/shell.d/launch-1password-test.sh`

### DIFF
--- a/bin/omarchy-install-service-1password
+++ b/bin/omarchy-install-service-1password
@@ -28,7 +28,7 @@ echo "Installing 1Password extension for Chromium..."
 install_chromium_extension
 
 echo "Opening 1Password..."
-uwsm-app -- 1password >/dev/null 2>&1 &
+setsid uwsm-app -- 1password >/dev/null 2>&1 &
 
 echo ""
 echo "1Password has been installed. Restart Chromium to load the browser extension."
--- a/bin/omarchy-install-service-dropbox
+++ b/bin/omarchy-install-service-dropbox
@@ -9,5 +9,5 @@ echo "Adding Dropbox to the bar..."
 omarchy-plugin-enable omarchy.dropbox
 
 echo "Starting Dropbox..."
-uwsm-app -- dropbox-cli start &>/dev/null &
+setsid uwsm-app -- dropbox-cli start &>/dev/null &
 echo "See Dropbox icon behind  hover tray in top right and right-click for setup."
--- a/test/shell.d/dropbox-test.sh
+++ b/test/shell.d/dropbox-test.sh
@@ -32,3 +32,42 @@ assertEqual(
   'dropbox file metadata includes relative time and folder'
 )
 JS
+
+test_tmp=$(mktemp -d)
+trap 'rm -rf "$test_tmp"' EXIT
+
+mock_bin="$test_tmp/bin"
+launch_log="$test_tmp/launch-log"
+mkdir -p "$mock_bin"
+
+cat >"$mock_bin/omarchy-pkg-add" <<'SH'
+#!/bin/bash
+:
+SH
+
+cat >"$mock_bin/omarchy-plugin-enable" <<'SH'
+#!/bin/bash
+:
+SH
+
+cat >"$mock_bin/setsid" <<'SH'
+#!/bin/bash
+printf '%s\n' "$*" >"$OMARCHY_TEST_LAUNCH_LOG"
+SH
+
+cat >"$mock_bin/uwsm-app" <<'SH'
+#!/bin/bash
+printf 'direct:%s\n' "$*" >"$OMARCHY_TEST_LAUNCH_LOG"
+SH
+
+chmod +x "$mock_bin"/*
+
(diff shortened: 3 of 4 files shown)


osamahbeig:
Thanks for picking this up — the setsid change and the regression test both look right to me.

One note on scope: the issue also flagged the same omission in `omarchy-install-service-dropbox` (line 12):

```bash
uwsm-app -- dropbox-cli start &>/dev/null &
```

Lower impact since `dropbox-cli start` daemonizes, but it's the same latent pattern. Since this PR closes #7870, that line would get closed along with it unfixed. Two easy options: fold the one-line `setsid` change into this PR, or leave it out and I'll file a small follow-up so it isn't lost. Either works for me — just flagging so it doesn't disappear with the close.

kx0101:
no problem! I folded the Dropbox case into this PR

`omarchy-install-service-dropbox` now launches through `setsid` too, and `dropbox-test.sh` runs the real installer with stubs and asserts the exact detached invocation.

Validated with:
- `bash test/shell.d/dropbox-test.sh`
- `bash test/shell.d/launch-1password-test.sh`
- `bash -n bin/omarchy-install-service-dropbox test/shell.d/dropbox-test.sh`

PR #7888, opened 2026-08-23
Fix broken manual links in README

ostrov:
The manual pages were renumbered after the Extra themes page was removed, but the README still linked to the old filenames, leaving the last ten links broken (manual/43 through manual/52). This removes the stale Extra themes entry and points the remaining links at the current filenames. Links 01-42 already match existing files and are unchanged.


sprajs:
This looks ready to close as already covered: merged #8089 renumbered the README table of contents, and the current `quattro` README contains all the replacements in this PR. I also checked the complete patch in reverse against current `quattro`; it applies cleanly, so there is no remaining hunk here to land.

PR #7891, opened 2026-08-23
Fix WoW fullscreen and focus stealing

anothaDev:
## Summary

- keep Hyprland authoritative over World of Warcraft XWayland fullscreen state
- ignore WoW background activation requests so it cannot pull users to another workspace
- add regression coverage for the exact Battle.net class/title rule

## Reproduction

Using a synthetic XWayland window with the live WoW identity (`steam_app_battlenet`, `World of Warcraft`) and the same EWMH events:

- before: client fullscreen cancellation changed internal/client state `2/2 -> 0/0`; activation changed workspace `1 -> 99`
- after: fullscreen remained `2/2`; workspace remained `1`

## Tests

- `bash test/shell.d/battlenet-test.sh`
- `luac -p default/hypr/apps/battlenet.lua`
- `bash -n test/shell.d/battlenet-test.sh`
- `shellcheck test/shell.d/battlenet-test.sh`
- live synthetic XWayland baseline/candidate verification

Broad sweep: 184/187 available shell tests passed. The remaining failures either reproduce on upstream or require the unavailable `omarchy-pkgs` checkout: `launch-about-test.sh` (reproduced on clean `upstream/quattro`), `snapper-test.sh`, and `unowned-system-paths-test.sh`; `config-test.sh` was excluded for the same missing checkout.

### DIFF
--- a/default/hypr/apps/battlenet.lua
+++ b/default/hypr/apps/battlenet.lua
@@ -14,3 +14,9 @@ o.window({ class = "^steam_app_battlenet$", title = "^Battle\\.net Setup$" }, {
   no_blur = true,
   no_shadow = true,
 })
+
+-- Keep Hyprland in charge of WoW's fullscreen state and ignore background activation requests.
+o.window({ class = "^steam_app_battlenet$", title = "^World of Warcraft$", xwayland = true }, {
+  focus_on_activate = false,
+  suppress_event = "fullscreen maximize",
+})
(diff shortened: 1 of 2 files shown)


anothaDev:
Closing this because the change is an overly specific per-game workaround at the Omarchy layer. The underlying behavior needs a generic XWayland fullscreen investigation in Hyprland rather than a World of Warcraft rule in distro defaults. Sorry for the noise.

PR #7905, opened 2026-08-23
Dictation indicator toggles dictation on left click

### LINKED ISSUE #7900: Dictation indicator opens the config instead of toggling dictation, unlike every other indicator
### System details

AMD Ryzen 5 7640U Omarchy  4.0.0-1 (stable channel), installed 2026-08-22

## What happens

Hover the center of the bar, then left click the dictation glyph 󰍬. The Voxtype
config TUI opens. Right click does the same. There is no way to start dictation
from the bar.

Omarchy 4.0.0-1, voxtype-bin 0.7.5-1. `F9` and `Super + Ctrl + X` both work fine.

## Why I think this is wrong

The dictation glyph sits in the same hover cluster as five other indicators, and
it is the only one that does not act on its own mode:

| Indicator | `onPressed` |
|---|---|
| `Dnd.qml:17` | flips do not disturb |
| `NightLight.qml:19` | `root.toggle()` |
| `StayAwake.qml:19` | `root.toggle()` |
| `ScreenRecording.qml:38` | opens capture menu, or stops an active recording |
| `Dictation.qml:34` | opens a settings window |

Four of them change state on click. The fifth opens a config window. They render
identically and sit next to each other, so I expect the same kind of action from
all of them. Opening settings from a status glyph that looks and behaves like a
toggle is surprising, and the first time it happened I assumed something was
broken.

The manual, https://omarchy.org/manual/the-top-bar/, describes it the way I
expected:

> These are status glyphs for modes you've turned on: do not disturb, night
> light, a queued reminder, an active screen recording, stay awake, and
> dictation.

> Clicking an indicator toggles that mode.

Dictation is listed, but it is the one that does not.

## Current code

```qml
// shell/plugins/bar/indicators/Dictation.qml:34-37
onPressed: function() {
  if (!root.bar) return
  root.bar.run("omarchy-voxtype-config")
}
```
## What makes me unsure

test/shell.d/fixtures/indicator-contract/shell.qml:207-208 asserts the current
behaviour for both buttons, and explicitly checks the old Waybar model picker is
gone:
```
root.assertTrue(root.commandCount("omarchy-voxtype-config") === 2, "Dictation clicks run config command")
root.assertTrue(root.commandCount("omarchy-voxtype-model") === 0, "Dictation clicks do not run model command")
```
That reads deliberate, so I may be arguing against a decision already made. If
so, the manual is the thing that needs correcting.

## Suggested fix

Dispatch on the button, the same way widgets/Microphone.qml:44 already does,
which keeps the config reachable:
```
onPressed: function(b) {
  if (!root.bar) return
  if (b === Qt.RightButton) root.bar.run("omarchy-voxtype-config")
  else root.bar.run("voxtype record toggle")
}
```
voxtype record toggle is already bound to Super + Ctrl + X in
default/hypr/bindings/voxtype.lua, so the indicator, the keybinding and the
omarchy-voxtype-status stream stay in agreement. I have been running this
locally through omarchy plugin clone omarchy.indicators and both buttons
behave as expected.

It needs the two assertions above updated. Happy to send a PR if you want it.

## Possibly related

#6823 touches the same click handler.

tahadx:
Closes #7900.

## What

The dictation glyph sat in the hover cluster acting like a toggle but opened the Voxtype config TUI on **both** left and right click — the only indicator of the six that didn't act on its own mode, and it left no way to start dictation from the bar at all.

- **Left click** now runs `voxtype record toggle` — the exact action of the existing `SUPER+CTRL+X` binding (`default/hypr/bindings/voxtype.lua`)
- **Right click** still opens `omarchy-voxtype-config`

## Why this shape

The manual describes these glyphs as status modes, and every sibling (`Dnd`, `NightLight`, `StayAwake`, `ScreenRecording`) flips its mode on press. Keeping config on right-click preserves one-click access to settings without breaking the cluster's convention.

## Tests

Updated `test/shell.d/fixtures/indicator-contract/shell.qml`, which previously *codified* the old behavior (asserting two clicks produced two config runs). It now asserts:

- left press → exactly one `voxtype record toggle`
- right press → exactly one `omarchy-voxtype-config`
- no `omarchy-voxtype-model` command from either

`indicator-contract-test.sh`, `voxtype-invitation-test.sh`, and the binding-conflict tests pass; full `./test/shell` shows only the three pre-existing environmental failures from a missing local `omarchy-pkgs` checkout.

### DIFF
--- a/shell/plugins/bar/indicators/Dictation.qml
+++ b/shell/plugins/bar/indicators/Dictation.qml
@@ -31,8 +31,12 @@ BarIndicator {
     }
   }
 
-  onPressed: function() {
+  onPressed: function(button) {
     if (!root.bar) return
-    root.bar.run("omarchy-voxtype-config")
+    // Left click toggles dictation, matching the SUPER+CTRL+X binding and the
+    // toggle-on-click behavior of every sibling indicator. Any other button
+    // opens the Voxtype config TUI.
+    if (button === Qt.LeftButton) root.bar.run("voxtype record toggle")
+    else root.bar.run("omarchy-voxtype-config")
   }
 }
--- a/test/shell.d/fixtures/indicator-contract/shell.qml
+++ b/test/shell.d/fixtures/indicator-contract/shell.qml
@@ -204,7 +204,8 @@ ShellRoot {
         root.injectBar(dictation)
         dictation.triggerPress(Qt.LeftButton)
         dictation.triggerPress(Qt.RightButton)
-        root.assertTrue(root.commandCount("omarchy-voxtype-config") === 2, "Dictation clicks run config command")
+        root.assertTrue(root.commandCount("voxtype record toggle") === 1, "Dictation left click toggles dictation")
+        root.assertTrue(root.commandCount("omarchy-voxtype-config") === 1, "Dictation right click opens the Voxtype config")
         root.assertTrue(root.commandCount("omarchy-voxtype-model") === 0, "Dictation clicks do not run model command")
       }
 


thecdrz:
Confirmed on Omarchy 4.0.0-1: left/right click on the dictation glyph both open `omarchy-voxtype-config`, while `SUPER+CTRL+X` / `F9` toggle correctly. The left=`voxtype record toggle` / right=config split matches the other indicators and the manual. Looks good to merge from here.

tahadx:
yeah
thanks!

tahadx (commented):

## What I checked

I verified the plumbing behind the change: `WidgetButton.qml` declares `signal pressed(int button)` and `triggerPress(button)` re-emits it, so `onPressed: function(button)` receives the real button and the `Qt.LeftButton` discrimination is sound rather than relying on incidental behavior. Left click matching the existing `SUPER+CTRL+X` binding and the toggle-on-click convention of every sibling indicator (`Dnd`, `NightLight`, `StayAwake`, `ScreenRecording`) makes this the consistent shape for the cluster.

## Observations

- **Updating the fixture mattered more than it looks:** the old indicator-contract fixture asserted two clicks → two config runs, i.e. it had codified the bug. The new assertions cover all three failure modes: exactly one toggle command on left press, exactly one config command on right press, zero model commands from either.
- **Keeping config on right-click** preserves one-click settings access instead of forcing users through a menu — good call over dropping it entirely.
- **Minor:** "any other button opens the config" means middle-click gets the TUI too; for a two-action glyph that seems fine, just noting the code implements exactly what the comment says.

Small diff, correct convention alignment, contract tests updated to match reality rather than the bug. Nothing further from me.

tahadx (commented):

## Rich code review — #7905 (dictation indicator toggles on left click)

Making the left click toggle dictation (matching the SUPER+CTRL+X binding and sibling indicators) and routing other buttons to the Voxtype config is consistent and user-friendly.

### Verified
- `onPressed: function(button)` correctly branches on `Qt.LeftButton`.
- The indicator-contract test is updated to assert exactly one `voxtype record toggle` (left) and one `omarchy-voxtype-config` (right), and zero `omarchy-voxtype-model`.
- `voxtype record toggle` is the same subcommand the keybinding uses — verified present in the codebase.

### One thing to double check (non-blocking)
If the user has toggled dictation while the bar is hidden or during a race with the voxtype service starting, a left click could silently no-op. That's inherent to the toggle design and not a regression of this PR.

Verdict: ship.

PR #7908, opened 2026-08-23
Agents panel: fix Codex limits for Codex CLI 0.149.0

### LINKED ISSUE #7873: Agents plugin cannot read Codex limits with Codex CLI 0.149.0 because `-a untrusted` is invalid
### System details

Omarchy 4.0.0-1, AMD Ryzen 7 7700X (8 cores/16 threads), 32 GB RAM, NVIDIA GeForce RTX 5070 (12 GB VRAM), NVIDIA driver 610.57.04, Linux kernel 7.1.8-arch1-3

### What's wrong?

The Omarchy Agents plugin displays:

  > Codex Limits Unavailable
  > initialize

  Local Codex token-usage statistics are displayed, but the subscription limit meters and reset times are unavailable.

  Codex itself is authenticated and working normally. `codex doctor` reports no authentication, installation, database, or connectivity failures.

  ## Cause

  The packaged Codex collector starts the Codex app server using:

  ```bash

  Codex CLI 0.149.0 no longer accepts untrusted as an approval policy. Running that command directly produces:

  Because the app server exits before responding to the initialization RPC, omarchy-agent-usage-codex returns:

  {
    "tierLabel": "",
    "limits": [],
    "authHelpText": "initialize"
  }

  The relevant command is currently constructed in:

  /usr/share/omarchy/bin/omarchy-agent-usage-codex

  It appears that the collector needs to replace -a untrusted with a currently supported value, likely -a never, given that it starts the app server in a read-only sandbox solely to query account information.

  ## Steps to reproduce

  1. Install Omarchy 4.0.0-1 and Codex CLI 0.149.0.
  2. Authenticate Codex using codex login.
  3. Use Codex so that it appears in the Omarchy Agents plugin.
  4. Open the Agents panel or refresh its data.
  5. Observe “Codex Limits Unavailable” with “initialize” beneath it.

  The issue can also be reproduced directly with:

  omarchy-agent-usage-codex --limits-only |
    jq '{tierLabel, limits, usageStatusText, authHelpText}'

  Result:

  {
    "tierLabel": "",
    "limits": [],
    "authHelpText": "initialize"
  }

  ## Expected behaviour

  The collector should start the Codex app server successfully, query account/rateLimits/read, and display the available Codex limit windows and reset times.

  If app-server startup fails, it would also be helpful for the panel to display the underlying stderr message instead of only the RPC stage name initialize.

tahadx:
Closes #7873.

## What

The Codex collector spawned the app-server with `-a untrusted`, and Codex CLI 0.149.0 removed that value from `--ask-for-approval`:

```
```

The process exited on argument parsing before ever answering `initialize`, so every subscription-limit meter read "Codex Limits Unavailable / initialize".

## Fix

Spawn with `-a never` instead. Two reasons for that specific value over dropping the flag or using `on-request`:

1. **Valid across versions** — both old CLIs (where all four policies existed) and 0.149+ accept it.
2. **Cannot stall a detached poller** — this collector runs detached with no UI; an approval policy that can prompt is one that can hang the poll forever. Combined with the existing `-s read-only` sandbox, `never` is the only policy that can never block.

## Verification

- Reproduced the argparse failure locally on codex 0.149.0, then confirmed `-a never app-server` answers the `initialize` RPC (both `never` and `on-request` do).
- The collector now completes its RPC handshake against real codex 0.149.0 (`usageStatusText` empty instead of "limits unavailable"); limit meters populate wherever the CLI is authenticated.
- `agent-usage-codex-scanner-test.sh` extended: the stubbed codex logs its spawn args, and the test asserts `-a never` is present and `untrusted` never returns.

### DIFF
--- a/bin/omarchy-agent-usage-codex
+++ b/bin/omarchy-agent-usage-codex
@@ -527,8 +527,13 @@ def fetch_codex_rpc():
     return result
 
   try:
+    # Codex 0.149 dropped "untrusted" from --ask-for-approval, which made this
+    # spawn exit before answering initialize and left the limits meters empty.
+    # This poller is detached: an approval prompt can never be answered, so
+    # "never" plus the read-only sandbox is both valid across CLI versions and
+    # the only policy that cannot stall the poll.
     proc = subprocess.Popen(
-      [codex, "-s", "read-only", "-a", "untrusted", "app-server"],
+      [codex, "-s", "read-only", "-a", "never", "app-server"],
       stdin=subprocess.PIPE,
       stdout=subprocess.PIPE,
       stderr=subprocess.DEVNULL,
--- a/test/shell.d/agent-usage-codex-scanner-test.sh
+++ b/test/shell.d/agent-usage-codex-scanner-test.sh
@@ -13,6 +13,8 @@ mkdir -p "$TEST_HOME/.codex/sessions/$(date +%Y/%m/%d)" "$TEST_HOME/bin"
 cat >"$TEST_HOME/bin/codex" <<'EOF'
 #!/bin/bash
 
+[[ -n $CODEX_SPAWN_LOG ]] && printf '%s\n' "$*" >>"$CODEX_SPAWN_LOG"
+
 while read -r request; do
   id=$(jq -r '.id // empty' <<<"$request")
   method=$(jq -r '.method // empty' <<<"$request")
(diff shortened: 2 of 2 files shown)


tahadx (commented):

## What I checked

Swapping `-a untrusted` for `-a never` is the right call, and the reasoning in the comment earns its place in the source: the value is accepted by both pre-0.149 and current CLIs, and a detached poller with no UI must never run under a policy that can prompt — combined with the existing `-s read-only` sandbox, `never` is the only choice that cannot wedge the collector waiting for an approval that can never arrive.

## Observations

- **Test directionality is good:** the spawn-arg log asserts both that `never` is present and that `untrusted` never returns, so a future regression in either direction trips the suite. Logging the stub's argv rather than mocking the RPC layer keeps the assertion honest against argparse changes.
- **One observation, not for this PR:** `stderr=subprocess.DEVNULL` on the `Popen` is why an argparse failure like this one stayed invisible until users hit "Codex Limits Unavailable". If you ever want faster diagnosis of similar breakage, teeing collector stderr to a debug file would surface it immediately. Out of scope here.

Focused fix, correct value chosen for principled reasons, verified against the real CLI version that broke. LGTM as far as I can tell.

Milofax:
Independently confirmed this fix on Omarchy 4.0.0-1 with codex-cli 0.149.0.

Before the change, `codex -s read-only -a untrusted app-server` exits immediately because `untrusted` is no longer accepted, and the Agents panel displays `Codex limits unavailable / initialize` while local token statistics continue to work.

I tested the collector with the proposed prompt-free `-a never` policy against a live ChatGPT-authenticated Codex installation. The RPC handshake completed and returned:

- `tierLabel: "pro"`
- a populated 7-day limit window
- an empty `usageStatusText`

I also verified the result in the running Omarchy panel: the Codex tab showed the Pro plan and weekly limit meter, with the red `initialize` card gone. This matches the intended behavior of this PR.

tahadx:
Thanks for independently verifying this

PR #7910, opened 2026-08-23
window-pop: don't tile an already-floating window

### LINKED ISSUE #7721: omarchy-hyprland-window-pop resizes unconditionally after a float toggle, rewriting the dwindle split when run on an already-floating window
### System details

Omarchy 4.0.0-1 (Quattro), Hyprland 0.56.2, Xwayland 24.1.13, Linux 7.1.8-arch1-3
AMD Ryzen 7 9800X3D, NVIDIA RTX 4070 Ti (AD104)
Single monitor, HDMI-A-1 `5120x2160@165.06`, `scale = 1`, layout `dwindle`

### What's wrong?

`omarchy-hyprland-window-pop` (bound to `SUPER + O`, "Pop window out (float & pin)")
toggles floating but resizes **unconditionally**. When it runs on a window that was
already floating, the toggle *tiles* the window, and the follow-up 1300x900 resize is
then applied to a tiled window — where an absolute pixel resize rewrites the dwindle
split ratio instead of setting a size. The result is a mangled workspace layout.

This is reachable from stock defaults, because Omarchy itself ships Steam as a
floating window (`default/hypr/apps/steam.lua:1`):

```lua
o.window("steam", { float = true, idle_inhibit = "fullscreen" })
```

So on a default install, the *first* `SUPER + O` pressed on Steam does not pop it
out — it tiles it and distorts the layout. Subsequent presses behave correctly,
because by then the window is tiled, which is why this looks like a once-per-boot
glitch.

### Steps to reproduce

1. Default Omarchy 4.0.0 install, `dwindle`, single monitor.
2. Open a terminal so the workspace has one tiled window.
3. Launch Steam. It opens floating at 1100x700, per the packaged rule above.
4. Focus the Steam window and press `SUPER + O` **once**.

### Expected

Either pop the window out to floating 1300x900 (the documented intent —
`omarchy:summary=Toggle to pop-out a tile to stay fixed on a display basis`),
or tile it at the layout's normal share. Not a rewritten split ratio.

### Actual

Steam becomes tiled at **3782x2110 — 74% of the display width** — and the
neighbouring terminal is crushed to **1300px** wide. The 1300 intended as the
popped window's width leaks into the split boundary instead.

Measured with `hyprctl clients -j` immediately before and after a single press:

```
before:  steam  floating=True   size=[1100, 700]   at=[2010, 743]
after:   steam  floating=False  size=[3782, 2110]  at=[1326, 38]
         foot   floating=False  size=[1300, 2110]
```

### Cause

In `omarchy-hyprland-window-pop`, the `elif` branch assumes the toggle floats the
window, but never checks:

```bash
elif [[ -n $addr ]]; then
  hypr_dispatch "hl.dsp.window.float({ window = \"$window\", action = \"toggle\" })" togglefloating "$window"
  hypr_dispatch "hl.dsp.window.resize({ window = \"$window\", x = $width, y = $height })" resizeactive exact "$width" "$height" "$window"
```

The `pinned == "true"` branch handles un-popping, but a window that is floating and
*not* pinned — Steam's launch state — falls through to `elif` and gets the pop
treatment applied backwards.

### Suggested fix

`hyprctl activewindow -j` already exposes `floating`, so the state is available
without an extra query. Gate the pop path on the window actually having been tiled:

```bash
floating=$(echo "$active" | jq ".floating")
...
elif [[ $floating == "true" ]]; then
  # Already floating and not popped: just tile it, no resize/center/pin.
  hypr_dispatch "hl.dsp.window.float({ window = \"$window\", action = \"toggle\" })" togglefloating "$window"
elif [[ -n $addr ]]; then
  # existing pop-out path, unchanged
```

Alternatively, gate on the `pop` tag rather than `pinned`, so the script's own
notion of "is this window popped" stops depending on pin state.

### Notes

Same script as #7160 (pop-out loses centering across monitors) and #5268 (pop-out
draws over the screensaver), but a distinct bug.

tahadx:
Closes #7721.

## What

`omarchy-hyprland-window-pop` (SUPER+O) toggled floating **unconditionally** before applying its absolute resize. On a window that was already floating and unpinned — reachable from stock defaults via the shipped Steam float rule — the toggle *tiled* the window, and the follow-up `1300x900` exact resize then landed on a tiled window, rewriting the dwindle split ratio instead of sizing anything. First press on Steam mangled the layout; later presses behaved, which is why it read as a once-per-boot glitch.

## Fix

Read `floating` from `hyprctl activewindow -j` alongside `pinned`, and dispatch the float toggle only when the window is currently tiled. Everything after it (resize → move/center → pin → raise → `+pop` tag) still runs either way, so an already-floating window gets the same pop presentation without ever passing through tiled state. The pinned (unpop) branch is untouched.

## Tests

New `test/shell.d/window-pop-test.sh` stubs `hyprctl`, capturing every dispatched command, and covers five entry states:

- tiled → full sequence incl. float toggle
- **floating+unpinned → no float toggle; resize/center/pin/tag still applied**
- explicit x/y moves instead of centering
- pinned → exactly three dispatches, never resizes
- no active window → no-op

## Verified live

On a real Hyprland session: an already-floating, unpinned window popped by this branch stays `floating:true` through the operation and ends pinned at the requested size — previously it ended up tiled with a rewritten split. A tiled window still gets the full float+pin treatment.

### DIFF
--- a/bin/omarchy-hyprland-window-pop
+++ b/bin/omarchy-hyprland-window-pop
@@ -10,6 +10,7 @@ y=${4:-}
 
 active=$(hyprctl activewindow -j)
 pinned=$(echo "$active" | jq ".pinned")
+floating=$(echo "$active" | jq ".floating")
 addr=$(echo "$active" | jq -r ".address")
 window="address:$addr"
 
@@ -25,7 +26,13 @@ if [[ $pinned == "true" ]]; then
   hypr_dispatch "hl.dsp.window.float({ window = \"$window\", action = \"toggle\" })" togglefloating "$window"
   hypr_dispatch "hl.dsp.window.tag({ window = \"$window\", tag = \"-pop\" })" tagwindow -pop "$window"
 elif [[ -n $addr ]]; then
-  hypr_dispatch "hl.dsp.window.float({ window = \"$window\", action = \"toggle\" })" togglefloating "$window"
+  # An already-floating window must keep its state: toggling would tile it,
+  # and the absolute resize below would then rewrite the dwindle split ratio
+  # instead of sizing a floating window. Skip the toggle so the rest of the
+  # pop presentation lands on a floating window either way.
+  if [[ $floating != "true" ]]; then
+    hypr_dispatch "hl.dsp.window.float({ window = \"$window\", action = \"toggle\" })" togglefloating "$window"
+  fi
   hypr_dispatch "hl.dsp.window.resize({ window = \"$window\", x = $width, y = $height })" resizeactive exact "$width" "$height" "$window"
 
   if [[ -n $x && -n $y ]]; then
--- a/test/shell.d/window-pop-test.sh
+++ b/test/shell.d/window-pop-test.sh
@@ -1,0 +1,80 @@
+#!/bin/bash
+
+set -euo pipefail
+
+source "$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd)/base-test.sh"
+
+tmp_dir=$(mktemp -d)
+trap 'rm -rf "$tmp_dir"' EXIT
+
+stub_bin="$tmp_dir/bin"
+mkdir -p "$stub_bin"
+DISPATCH_LOG="$tmp_dir/dispatch.log"
+export DISPATCH_LOG
+
+# hyprctl stub: `activewindow -j` replays the window state the test sets via
+# WINDOW_JSON; every dispatch lands in DISPATCH_LOG so assertions read the
+# exact command sequence the pop script produced.
+cat >"$stub_bin/hyprctl" <<'SH'
+#!/bin/bash
+if [[ $1 == "activewindow" ]]; then
+  printf '%s' "$WINDOW_JSON"
+  exit 0
+fi
+printf '%s\n' "$*" >>"$DISPATCH_LOG"
+exit 0
+SH
+chmod +x "$stub_bin/hyprctl"
+
+run_pop() {
+  : >"$DISPATCH_LOG"
+  WINDOW_JSON="$1" PATH="$stub_bin:$PATH" \
+    "$ROOT/bin/omarchy-hyprland-window-pop" "${@:2}"
+}
+
+# A tiled window gets the full pop: float, resize, center, pin, raise, tag.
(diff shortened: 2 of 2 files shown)


tahadx (commented):

## What I checked

The guard is minimal and correct: `floating` is read next to `pinned` from `hyprctl activewindow -j`, and the float dispatch only fires when the window is currently tiled. Everything downstream (resize → move/center → pin → raise → `+pop` tag) still applies to already-floating windows, which preserves the pop presentation without the destructive toggle-then-resize sequence that rewrote dwindle split ratios.

## Observations

- **No-active-window path is safe:** `jq ".floating"` over empty input yields nothing, but the `[[ -n $addr ]]` branch is skipped in that case, so the new read cannot misfire there.
- **Pinned branch untouched:** the `pinned == "true"` branch short-circuits before the floating check, so a floating+pinned window still takes the unpick path unchanged — matches the stated intent and keeps unpick semantics intact.
- **Test shape:** the stubbed-dispatch harness covering five entry states (tiled, floating+unpinned, explicit x/y, pinned, no window) is exactly the coverage this script needed — it previously had none. The "pins exactly once" and "exactly three dispatches" assertions are nice precision touches.

Clean, well-scoped fix with live verification on top. Nothing further from me.

tahadx (commented):

## Rich code review — #7910 (window-pop: don't tile an already-floating window)

The fix targets the real bug: an already-floating window that gets popped would have its float `toggle` flip it back to tiled, so the absolute resize would then rewrite the dwindle split ratio instead of sizing the window. Guarding the toggle on `$floating != "true"` is the correct minimal change.

### Verified
- `floating=$(echo "$active" | jq ".floating")` yields the JSON boolean; a missing field yields `null`, which `!= "true"` treats as not-floating and safely toggles. Correct for all states.
- Pinned branch (`unpick`) is untouched and still runs exactly its three dispatches — the test pins that.
- Tests cover tiled, already-floating, explicit-position, pinned, and no-active-window cases. All pass.

The window-pop script shells out to `echo | jq` three times in a row for one `hyprctl activewindow -j` blob. Could parse once and read the three fields from one `jq` call — a minor efficiency nicety, not required for this fix.

Verdict: ship.

PR #7915, opened 2026-08-23
Treat pending-charge as a hold only inside a real charge limit

### LINKED ISSUE #7803: `pending-charge` treated as a charge-limit hold without checking the level: a failed battery reports "Holding at 75-80%" at 0%
### System details

Lenovo ThinkPad X1 Yoga Gen 6 (`20XY0024US`), BIOS N32ET75W 1.51, Omarchy `4.0.0-1`, kernel `7.1.8-arch1-3`, `quickshell-git 0.3.0.r20.g28771c7-2`, `upower 1.91.3-1`. Battery: SMP `5B10W13973`, 4 cycles, electrically dead (2.296 V on a 15.44 V pack).

### What's wrong?

A battery the EC is refusing to charge because it is electrically dead is reported by both the power panel and `omarchy-battery-status` as a user-configured charge-limit hold:

```
$ omarchy-battery-status
Battery 0%  ·  Holding at 75-80%  ·  0W / 58Wh
```

"Holding at 75-80%" while sitting at 0%. The pack is at 2.296 V against a 15.44 V design — a 4S pack at ~0.57 V/cell, far below the ~2.5 V/cell precharge floor — so the EC declines to charge it:

```
$ cd /sys/class/power_supply/BAT0 && grep . status capacity voltage_now voltage_min_design \
    power_now charge_control_start_threshold charge_control_end_threshold charge_behaviour
status:Not charging
capacity:0
voltage_now:2296000
voltage_min_design:15440000
power_now:0
charge_control_start_threshold:0
charge_control_end_threshold:100
charge_behaviour:[auto] inhibit-charge force-discharge
```

No charge limit is configured — sysfs reads 0/100, `charge_behaviour` is `auto`, TLP is not installed, and nothing in `/etc` writes a threshold. The panel nevertheless labels the field **"Charge limit"** with value **"Holding"**, and the mode label reads **"Threshold"**.

**Root cause.** The kernel's `Not charging` (UPower `pending-charge`) means only *"AC present, EC is not charging the pack"* — for any reason: a threshold hold, a battery fault, an insufficient PD supply, or `charge_behaviour=inhibit-charge`. Two independent implementations collapse that to "a charge limit is holding it", without checking whether the charge level is anywhere near a limit.

**1. `shell/plugins/panels/power/Model.js:59`** — no threshold check at all:

```js
function chargeThresholdActive(device, onBattery, states) {
  ...
  if (d.state === s.PendingCharge) return true                  // <-- unconditional
  if (d.state === s.FullyCharged && fraction < 0.99) return true
  if (d.state !== s.Charging || fraction >= 0.99) return false
  return Number(d.changeRate || 0) <= 0.2 || Number(d.timeToFull || 0) >= 8 * 60 * 60
}
```

This one predicate drives the `Charge limit` / `Holding` pair (`panels/power/Panel.qml:441-446`), the `Threshold` mode label (`Model.js:86`), and the battery icon (`Model.js:75`). Note it never reads a threshold *value*.

**2. `bin/omarchy-battery-status`** — guards on a threshold *existing*, but not on the level:

```bash
if [[ $ac_online == "true" && -n $threshold_end ]]; then
  if [[ $state == "pending-charge" ]]; then
    charge_holding=true                                          # <-- no percentage check
  elif [[ $state == "fully-charged" ]] && (( percentage < 99 )); then
    charge_holding=true
  elif [[ $state == "charging" && $charge_idle == "true" ]] && (( threshold_end < 99 && percentage >= threshold_end )); then
    charge_holding=true                                          # <-- this branch gets it right
  fi
fi
```

The third branch already applies the correct discriminator (`percentage >= threshold_end`). The `pending-charge` branch doesn't.

### Relationship to #7374

Adjacent but distinct — and **#7374's fix alone does not cure this; it makes the label worse.**

#7374 is a data-source bug: UPower's udev hwdb default `CHARGE_LIMIT=75,80` is preferred over sysfs, so the wrong threshold *value* is displayed. (That default is also where the bogus `75-80%` above comes from — this machine has no limit set.)

Here the threshold value is irrelevant, because the classification never consults it. Applying #7374's proposed sysfs preference to this machine yields `threshold_end=100`, which is still non-empty, so `-n $threshold_end` still passes and the output becomes:

```
Battery 0%  ·  Holding at 0-100%  ·  0W / 58Wh
```

The ThinkPad T14 Gen 6 report in #7374's comments — *"there is no charge limit enabled at all, but the power panel still reports one"*, with sysfs 0/100 and `ChargeThresholdEnabled: false` — is the `fully-charged` face of this same classification gap.

### Expected

Claim a hold only when a real limit exists **and** the battery is actually in the holding band. Otherwise `pending-charge` should surface honestly as "Not charging".

### Proposed fix

Gate both implementations on two conditions:

1. **a real limit is configured** — `threshold_end < 100`, ideally UPower's `ChargeThresholdEnabled`, which correctly reports `false` on the machines in both this report and #7374;
2. **the level is in the holding band** — `percentage >= threshold_start`.

For `Model.js`, which ignores threshold values entirely today, the values would need plumbing through (`Panel.qml` already surfaces `batteryInfo.threshold`). A minimal level-only guard removes the false positive with no new data:

```js
if (d.state === s.PendingCharge) return fraction >= 0.5
```

Separately, consider a distinct **"Not charging"** state so an EC that is refusing to charge is visible as such. Right now a dead pack is presented as a normal user-configured limit, which is actively misleading — it cost me a diagnosis before I read the raw sysfs values, and "Holding" reads as *deliberate and healthy*.

### Steps to reproduce

1. Get the EC to refuse charging for a non-threshold reason — a deeply-discharged or failed pack, or on supporting hardware force it with `echo inhibit-charge | sudo tee /sys/class/power_supply/BAT0/charge_behaviour`.
2. Confirm no charge limit is set: `charge_control_end_threshold` is 100.
3. Run `omarchy-battery-status`, and open the power panel with `SUPER+CTRL+P`.
4. Observe `Holding at 75-80%` and a "Charge limit: Holding" pair, despite no limit being configured.

tahadx:
Closes #7803

A battery parked in kernel `pending-charge` was always reported as a
threshold hold, even when the hold came from something other than a
configured limit — a dead pack at 0% with kernel-default limits (0/100)
showed "Holding at 75-80%".

UPower's `pending-charge` only means the EC is not charging the pack. A
charge limit is one reason, but so are a dead battery, an undersized
adapter, or inhibit-charge. Both the status script and the power panel now
only treat it as a hold when a real limit is configured (`end < 99`) and
the level sits inside its band:

- `omarchy-battery-status` gates the `pending-charge` branch on
  `threshold_end < 99` and `percentage` inside
  `${threshold_start:-0}..$threshold_end`, and emits raw
  `threshold_start`/`threshold_end` keys in `--shell` output.
- The panel passes those limits into `Model.chargeThresholdActive`,
  which now requires them for the `PendingCharge` branch; machines with
  no limit data fall back to the plain charging/on-battery presentation.
  Icon and mode-label helpers thread the same limits through.

The stalled-charging and fully-charged holds are unchanged.

Tests: new battery-status scenarios cover a dead pack at 0% under default
limits (not holding), an in-band 78% at 75-80 (holding plus raw keys), and
an out-of-band 90% (not holding); power model tests cover in-band,
out-of-band, and missing-limit pending charge.

### DIFF
--- a/bin/omarchy-battery-status
+++ b/bin/omarchy-battery-status
@@ -87,7 +87,13 @@ fi
 
 charge_holding=false
 if [[ $ac_online == "true" && -n $threshold_end ]]; then
-  if [[ $state == "pending-charge" ]]; then
+  # UPower's pending-charge only means the EC is not charging the pack — a
+  # configured limit is one reason, but so are a dead battery, an undersized
+  # adapter, or inhibit-charge. Only call it a hold when a real limit is
+  # configured and the level sits inside its band.
+  if [[ $state == "pending-charge" ]] &&
+    (( threshold_end < 99 )) &&
+    (( percentage >= ${threshold_start:-0} && percentage <= threshold_end )); then
     charge_holding=true
   elif [[ $state == "fully-charged" ]] && (( percentage < 99 )); then
     charge_holding=true
@@ -118,6 +124,10 @@ if [[ $shell_output == "true" ]]; then
       printf 'threshold\t%s%%\n' "$threshold_end"
     fi
   fi
+
+  # Raw limits for the panel's hold logic; empty when the kernel exposes none.
+  [[ -n $threshold_start ]] && printf 'threshold_start\t%s\n' "$threshold_start"
+  [[ -n $threshold_end ]] && printf 'threshold_end\t%s\n' "$threshold_end"
 
   exit 0
 fi
--- a/shell/plugins/panels/power/Model.js
+++ b/shell/plugins/panels/power/Model.js
@@ -49,28 +49,40 @@ function batteryFraction(device) {
   return device && device.isPresent ? Math.max(0, Math.min(1, device.percentage)) : 0
 }
 
-function chargeThresholdActive(device, onBattery, states) {
+function chargeLimitsInBand(limits, fraction) {
+  var end = Number(limits && limits.end)
+  if (!(end >= 1) || end >= 99) return false
+  var start = Number(limits && limits.start) || 0
+  return fraction * 100 >= start && fraction * 100 <= end
+}
+
+function chargeThresholdActive(device, onBattery, states, limits) {
   var d = device || {}
   var s = states || {}
   if (!(d && d.isPresent && !onBattery)) return false
 
   var fraction = batteryFraction(d)
   if (d.state === s.Discharging) return false
-  if (d.state === s.PendingCharge) return true
+  if (d.state === s.PendingCharge) {
+    // pending-charge covers any reason the EC declines to charge — including a
+    // dead battery or an undersized adapter — so only treat it as a hold when
+    // the script reports a real limit and this level could plausibly be held.
+    return chargeLimitsInBand(limits, fraction)
+  }
   if (d.state === s.FullyCharged && fraction < 0.99) return true
   if (d.state !== s.Charging || fraction >= 0.99) return false
 
   return Number(d.changeRate || 0) <= 0.2 || Number(d.timeToFull || 0) >= 8 * 60 * 60
 }
 
-function batteryIcon(device, onBattery, states) {
+function batteryIcon(device, onBattery, states, limits) {
   var d = device || {}
   if (!d.isPresent) return ""
 
   var chargingIcons = ["󰢜", "󰂆", "󰂇", "󰂈", "󰢝", "󰂉", "󰢞", "󰂊", "󰂋", "󰂅"]
   var defaultIcons = ["󰁺", "󰁻", "󰁼", "󰁽", "󰁾", "󰁿", "󰂀", "󰂁", "󰂂", "󰁹"]
   var index = Math.max(0, Math.min(9, Math.floor(d.percentage * 10)))
-  var threshold = chargeThresholdActive(d, onBattery, states)
+  var threshold = chargeThresholdActive(d, onBattery, states, limits)
 
   if (threshold) return defaultIcons[index]
   if (d.state === states.FullyCharged) return "󰂅"
@@ -78,12 +90,12 @@ function batteryIcon(device, onBattery, states) {
   return defaultIcons[index]
 }
 
-function modeLabel(device, onBattery, states) {
+function modeLabel(device, onBattery, states, limits) {
   var d = device || {}
   if (!d.isPresent) return ""
 
   var percentage = d.isPresent ? d.percentage : 0
-  if (chargeThresholdActive(d, onBattery, states)) return "Threshold"
+  if (chargeThresholdActive(d, onBattery, states, limits)) return "Threshold"
   if (onBattery) return "On battery"
   if (!onBattery && percentage >= 1) return "Fully charged"
   return "Charging"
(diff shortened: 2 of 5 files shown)


tahadx (commented):

## What I checked

The two-layer fix reads well: `omarchy-battery-status` stops treating kernel `pending-charge` as a threshold hold unless a real limit is configured (`threshold_end < 99`) *and* the level sits inside the band, while `Model.chargeThresholdActive` gets the same discipline via the new `chargeLimitsInBand` helper. Emitting raw `threshold_start`/`threshold_end` keys from `--shell` output so the panel owns its own policy (rather than duplicating it) is a clean split of responsibilities.

## Observations

- **Integer safety confirmed:** `percentage`, `threshold_start`, and `threshold_end` are all normalized through awk `int()` before reaching the `(( ))` band check, so fractional UPower output cannot break bash arithmetic on this path.
- **Small asymmetry between layers:** the bash side accepts any band with `end < 99`, but `chargeLimitsInBand` additionally requires `end >= 1`. A hypothetical `end == 0` configuration would report holding in shell output but not in the panel icon. Harmless today; aligning the two predicates would prevent future drift if either copy evolves independently.
- **Test coverage mirrors the bug report:** dead pack at 0% under kernel-default limits (not holding, raw keys still exposed), in-band 78% at 75–80 (holding plus label), out-of-band 90% (not holding), plus panel-level cases for missing limits. The refactored parameterized `setup_battery` harness makes those cases readable — good test ergonomics.

tahadx (commented):

## Rich code review — #7915 (pending-charge hold)

The two-layer guard is sound. The CLI now only reports a hold when a real limit is configured (`threshold_end < 99`) *and* the level sits inside the band, and the panel mirrors that through the new `chargeLimitsInBand` helper instead of duplicating the policy. Emitting raw `threshold_start`/`threshold_end` from `--shell` so the panel owns its own logic is a clean single-source split.

### Verified
- Bash path: `percentage >= ${threshold_start:-0}` and `threshold_end < 99` are guarded by the outer `-n $threshold_end`, and the `(( ))` check degrades gracefully when `threshold_start` is empty (`:-0`). Correct and safe.
- JS path: `Number("") == 0` makes a missing/empty limit return false; `end >= 99` rejects kernel defaults; `end >= 1` guards degenerate configs. On first render before `batteryInfo` fills, `Number(undefined)` is NaN and `!(NaN >= 1)` short-circuits safely.
- Wiring: `batteryInfo` is populated by `Model.parseKeyValue` from the `--shell` output, so the new keys land as reactive properties and `batteryLimits` re-evaluates on refresh. Verified the `targetName === "battery"` path.
- Tests: battery-status + power model tests pass.

### Nit (non-blocking)
The bash side accepts any band with `end < 99`, while the JS side additionally requires `end >= 1`. A hypothetical `end == 0` config would report holding in the CLI but not in the panel icon. Harmless today; aligning the two predicates would prevent future drift.

Verdict: ship.

PR #7919, opened 2026-08-23
Add OpenClaw under Install > AI and as a selectable agent

dhh:
Adds [OpenClaw](https://openclaw.ai) in the places it belongs: under Install > AI and Remove > AI, and as a terminal agent under Setup > Defaults > Agent. Shaped after #7469, but OpenClaw is a different animal in one way that changes the whole arrangement.

## Installing it means onboarding it

OpenClaw is not a command you point at a checkout. It is a gateway that runs as a systemd user service, holds the channel connections and the session state, and answers from whichever channel reaches it; the terminal UI is one more client of that gateway. A machine with `openclaw` on `PATH` and nothing listening behind it has nothing to talk to.

So `omarchy-install-ai-openclaw` owns both halves: the mise install, then `openclaw onboard --install-daemon`, which is what leaves the gateway running. It reports through `--check` whether both are done, and does nothing at all when they are — setting OpenClaw as the default agent runs it on the way to launching, and a machine already set up must not sit through it again.

`--check` asks mise, not `PATH`. The stub is written on every machine at user setup and is cold until someone runs it, so `omarchy-cmd-present` answers yes on a machine that has never had OpenClaw. Taking that for installed would skip the floating terminal and run the whole setup, prompts and all, inside the menu action.

That is why `omarchy-default-agent` grows the same seam #7469 needs: an agent may bring its own installer instead of reaching for `mise use` from there. Every other agent keeps the mise path unchanged. If Hermes lands first the two versions of that seam are the same mechanism and reconcile mechanically.

## Removal

The gateway is a service OpenClaw writes itself, so `omarchy-remove-ai-openclaw` asks it to take it back rather than unlinking the unit behind it — `gateway uninstall` stops the running one first. A machine that never finished onboarding has no service to give up, and that is not an error worth stranding the rest of the removal over.

The mise install goes. The stub stays: leaving it puts the machine back where a fresh one starts rather than a step behind it, and nothing runs until someone types `openclaw` again. `~/.openclaw` stays too — the config names the inference provider, the pairings each took a round trip through their service to approve, and the sessions and skills are the user's.

## The terminal agent

`omarchy-agent` learns an `openclaw` case: `openclaw tui`, with a prompt passed as `--message`. Not `openclaw chat`, which is the same command with `--local` — that runs the agent in-process, blind to the gateway's sessions, and refuses outright while a gateway owns them. Installing OpenClaw here always leaves one running, so attaching is the only case that exists. There is no auto-approve flag to reach for; OpenClaw approves nothing silently, so it joins Pi and Ori with nothing to skip.

## Glyph

`U+E90A`, from [Font Awesome's shrimp](https://fontawesome.com/icons/shrimp) (CC BY 4.0). OpenClaw's own lobster is five paths under a gradient with the eyes carrying half the reading; flattened to one monochrome path it comes out a beetle-shaped blob, which is the trap the icon-font guide warns about. The stand-in reads as a crustacean at menu size, which is the job.

## Tests

`openclaw-install-test.sh` and `openclaw-remove-test.sh` cover the two commands; `default-agent-test.sh` covers the installer seam and the launch flags. Each assertion was checked by mutation.

https://claude.ai/code/session_01SJ72BVipaCjecQFZ6bK2N4


smfworks:
Verified against a live Omarchy + OpenClaw machine (openclaw 2026.9.4, `3a9d69d`):

- Every CLI surface this PR uses exists and behaves as described: `openclaw tui` (attaches to the running gateway), `openclaw onboard --install-daemon`, `openclaw gateway uninstall`, `openclaw dashboard`. The gateway-first framing in `omarchy-agent` matches how OpenClaw actually works — the TUI is a gateway client, and in-process runs refuse while a gateway owns the sessions.
- The cold-stub detection (`agent_installer` + `--check`) is the right shape: it keeps onboarding prompts out of the menu action and out of `mise use`.

**One confirmed gap** (seconding the Copilot review, with evidence): existing/upgraded machines never get the `openclaw` stub, because `install/user/mise.sh` only runs during initial user setup while updates apply new state through `migrations/*.sh`. Both the `omarchy-remove-preinstalls` addition (which now deletes `~/.local/bin/openclaw`) and the remove flow's "leaves the stub a fresh machine would have" assumption (tested in `openclaw-remove-test.sh`) depend on that stub existing on machines that never re-run user setup.

Proposed migration, mirroring `migrations/1787342993.sh` (ori):

```bash
echo "Install OpenClaw lazy stub"

if [[ ! -f $HOME/.local/state/omarchy/preinstalls-removed ]]; then
  omarchy-mise-install npm:openclaw openclaw
fi
```

One more hole worth noting: adding OpenClaw as a selectable agent leaves the agents usage panel without stats for it — `omarchy-agent-usage-{claude,codex,fireworks}` exist on main, and #9869's `omarchy-agent-usage-openclaw` fills exactly that hole (I verified its CLI calls against 2026.9.4: `sessions --json --all-agents --limit all`, `audit --json` + `--cursor` pagination are all valid). Worth folding in here or landing as an immediate follow-up.

bjarneo:
Automated duplication check: this pull request looks similar to #9869, which covers the same OpenClaw agent option. I keep that one open and close this one to consolidate review.

PR #7926, opened 2026-08-23
Run notification click actions as safe argv

ryanrhughes:
## Summary

Rework notification **click commands** so untrusted data can never be executed as a command. The click action is now an argv vector carried in the `omarchy-exec-argv` hint and run as bash **positional parameters** (`bash -lc 'exec "$@"'`), never interpolated into a shell string. This closes the RCE class where a downloaded video title / filename becomes a command when the "Download complete" toast is clicked (the [mehmetince](https://mehmetince.net/omarchy/) exploit).

- **`--exec` takes the command as rest-of-line words** — `--exec mpv -- "$file"`. The caller's shell splits them into discrete arguments before the tool sees them, so a value with spaces or shell metacharacters stays one inert argument. It is the only mechanism; there is no shell-string form to interpolate into unsafely.
- The tool **never splits a string itself**: a single quoted whole-command (`--exec "a b c"`) is rejected and points at the unquoted form (whitespace-splitting a string would hand argument boundaries to attacker-controlled content).
- `--exec` is recognized only **after** the headline/description, so an untrusted positional equal to `--exec` can't be mistaken for the delimiter.
- Migrate every caller (screenshot, screen recording, taildrop, yt-dlp host, crash watcher, migrate-notify, first-run invitation hooks) and update `docs/notifications.md`.

## Security scope

- Complements #7847 (which stopped the yt-dlp record forgery). This removes the shell from the click path entirely, so safety no longer depends on every caller quoting perfectly.
- **Closed:** untrusted content/data relayed into a notification — web notifications can't set the exec hint, and relayed titles/filenames are confined to inert argument data.
- **Out of scope by design:** a caller deliberately naming a shell (`--exec sh -c …`) is developer-authored, not injection (greppable in review); a native same-user process already has your privileges and needs no notification to run code.
- Adversarially reviewed end-to-end (encode → hint transport → `parseExecArgv` → persistence round-trip → exec): no in-scope hole found. One latent footgun it surfaced (a positional literally equal to `--exec`) is fixed; the residual — a *description* exactly `--exec` losing its click action — is a harmless CLI ambiguity.

## Testing

- `./test/shell` and `./test/cli` pass, except two **pre-existing** environmental failures unrelated to this branch: `bar-text-color-test` (needs ImageMagick `magick`) and `hw-hybrid-gpu-test` (needs `omarchy-cmd-present` on PATH).
- Added/updated coverage: argv encoding, rejection of quoted-string/empty `--exec`, the `--exec`-looking-positional guard, `parseExecArgv` fail-closed cases, the persistence round-trip, and the migrated callers.

### DIFF
--- a/bin/omarchy-notification-send
+++ b/bin/omarchy-notification-send
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 # omarchy:summary=Send an Omarchy desktop notification
-# omarchy:args=[--app-name <app-name>] [-g <glyph>] [-u <low|normal|critical>] [--image <path-or-uri>] <headline> [description] [notify-send options] [--exec <program> [args...]]
+# omarchy:args=[--app-name <app-name>] [-g <glyph>] [-u <low|normal|critical>] [-i <icon>] [-t <ms>] [--image <path-or-uri>] <headline> [description] [--exec <program> [args...]]
 # omarchy:examples=omarchy notification send "Reminder" "5 minutes are up" -g 󰢌
 
 set -euo pipefail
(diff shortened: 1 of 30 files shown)


bin/omarchy-notification-send
```diff
@@ -90,7 +82,19 @@ if (($# > 0)) && [[ $1 != -* ]]; then
...
+    # captured, so an untrusted headline that is literally "--exec" is taken as
+    # text and can't be mistaken for the delimiter. --exec therefore comes last.
+    shift
+    exec_args=("$@")
+    exec_present=1
+    break
+  elif parse_omarchy_option "$@"; then
     shift "$parsed_option_args"
   else
     args+=("$1")
```
acrogenesis: The command does not put -- before the headline.

If the description starts with -, the command sends it as an option to notify-send.

I did a test on a live Omarchy session with this change.

```bash
omarchy-notification-send "ARGV-AUDIT-7926-dash" \
  '--hint=string:omarchy-exec-argv:["bash","-c","/usr/bin/touch /tmp/omarchy-argv-audit-real/pwned-dash-body"]'
omarchy-shell notifications invokeLast
```

The description became an option. The notification kept omarchy-exec-argv.
A click of the notification ran the command. A file was created.

If a theme name or an app name is that option text, a click can run a command.

Do this:
1. Do not send unknown options that start with - to notify-send.
2. Use: notify-send "${args[@]}" -- "$headline" ${description:+"$description"}
3. Do not accept a headline or a description that starts with -.

shell/plugins/notifications/Service.qml
```diff
@@ -353,18 +353,19 @@ Item {
...
-    if (command) {
-      // Detached so the launched command outlives the shell process, which the
-      // installer toasts depend on: they restart the shell as their first act.
-      Util.execDetached(command)
+
+    // Run the argv (via Util.execArgv, no shell interpretation). Detached so it
+    // outlives the shell, which installer toasts depend on: they restart it.
+    var argv = NotificationLogic.parseExecArgv(entry ? entry.execArgv : "")
+    if (argv) {
+      Util.execArgv(argv)
```
acrogenesis, line 368: The click runs the argv from omarchy-exec-argv. The command does not examine the sender.

parseExecArgv does not reject bash -c. It rejects only a program name that starts with -.

I did a test on a live session with this QML.

```bash
notify-send -a "System Update" \
  -h 'string:omarchy-exec-argv:["bash","-c","/usr/bin/touch /tmp/omarchy-argv-audit-real/pwned-any-sender"]' \
  "ARGV-AUDIT-7926-any" "Click runs bash -c touch"
omarchy-shell notifications invokeLast
```

The notification used the app name System Update. A click created a file.

The pull request says a local program does not need a notification to run a command.
A program with only notification permission is not the same as a program that can execute all commands.

Do this:
1. Run execArgv only if the app name is omarchy-action.
2. After a restore, remove execArgv unless the app name is omarchy-action.

shell/plugins/notifications/NotificationLogic.js
```diff
@@ -57,14 +57,35 @@ function glyphFromHints(hints) {
...
+    parsed = JSON.parse(text)
+  } catch (e) {
+    return null
+  }
+
+  if (!Array.isArray(parsed) || parsed.length === 0) return null
+  for (var i = 0; i < parsed.length; i++) {
+    if (typeof parsed[i] !== "string") return null
+  }
+  if (!parsed[0] || parsed[0].charAt(0) === "-") return null
```
acrogenesis, line 91: The comment says the program is present.

The code only checks that the first word is not empty and does not start with -.

```js
parseExecArgv('["bash","-c","touch /tmp/x"]')
// => ["bash","-c","touch /tmp/x"]
```

I used this argv in the live System Update notification.

Examine the sender. You can also reject bash, sh, and dash as the first word.

PR #7948, opened 2026-08-23
Stop the DMI chassis fallback in omarchy-hw-laptop reading an empty string

chubuntuarc:
`omarchy-hw-laptop` falls back to the DMI chassis type when the machine exposes no ACPI lid button, but the fallback can never match. Bash only treats `$(< file)` as "read this file" when the substitution holds that one redirection and nothing else; adding `2>/dev/null` turns it into a null command with two redirections, whose stdout is empty.

```
$ v=$(< /sys/class/dmi/id/chassis_type 2>/dev/null); printf '[%s]\n' "$v"
[]
$ v=$(< /sys/class/dmi/id/chassis_type); printf '[%s]\n' "$v"
[9]
$ v=$(cat /sys/class/dmi/id/chassis_type 2>/dev/null); printf '[%s]\n' "$v"
[9]
```

So the `case` always compares the empty string, and every laptop that reaches the fallback exits 1. On x86 the ACPI loop above answers first, which is what has kept this invisible: only hardware with no `/proc/acpi` at all falls through. Apple Silicon under Asahi is such a machine — `hostnamectl chassis` says `laptop` and `/sys/class/dmi/id/chassis_type` holds `9`, yet `omarchy-hw-laptop` reports that this is not a laptop.

Everything gated on the command is silently off there. `Laptop Display`, `Mirror Display`, and `Battery Percentage` all carry `"when": "omarchy-hw-laptop"` in `default/omarchy/omarchy-menu.jsonc`, and `omarchy-hyprland-monitor-watch:78` skips its `omarchy-hw-laptop && omarchy-hyprland-monitor-external-active` branch.

`cat` keeps the error suppression the original form was reaching for. After this change no `$(<` in the tree carries an extra redirection.

Tested on a MacBook Air M2 (J413) running Asahi Arch: `bin/omarchy-hw-laptop` goes from exit 1 to exit 0, matching `hostnamectl chassis`.

`./test/all` — 188 of 190 test files pass. The two failures, `bar-icon-geometry-test.sh` and `snapper-test.sh`, fail identically on a clean checkout on this machine and are unrelated to this change: a bar icon is painted 0.59 px off center, and the Snapper suite wants a sibling `omarchy-iso` checkout that isn't present.

### DIFF
--- a/bin/omarchy-hw-laptop
+++ b/bin/omarchy-hw-laptop
@@ -10,7 +10,7 @@ done
 # Fall back to the DMI chassis type for laptops that don't expose an ACPI lid
 # button. 8=Portable 9=Laptop 10=Notebook 14=Sub Notebook 30=Tablet
 # 31=Convertible 32=Detachable.
-case $(< /sys/class/dmi/id/chassis_type 2>/dev/null) in
+case $(cat /sys/class/dmi/id/chassis_type 2>/dev/null) in
 8 | 9 | 10 | 14 | 30 | 31 | 32) exit 0 ;;
 esac
 


flip-in:
Independent confirmation on different hardware: MacBookPro18,3 (M1 Pro), Asahi, kernel 7.1.6-1-1-ARCH, Omarchy 4.0.2-mac.1 (`omarchy-dev 4.0.2.r6875.g2afd2ef-1`).

Same result as your M2 Air — `bin/omarchy-hw-laptop` goes from exit 1 to exit 0 with this change, against `chassis_type` `9` and `hostnamectl chassis` `laptop`. I hit this independently while chasing clamshell mode and reached the identical diagnosis before finding this PR; your write-up caught consequences I had missed, in particular the three `omarchy-menu.jsonc` entries, which are indeed absent here.

One thing worth having in the record for whoever triages this, because it is not visible from either PR alone: **this fix and the `omarchy-hw-laptop-closed` fix are both required before clamshell works on Asahi, and neither is sufficient on its own.**

- Without this PR, `omarchy-hyprland-monitor-watch:78` never starts its reconciliation poll, so nothing maintains clamshell state.
- Without #7951 / #9754, `omarchy-hw-laptop-closed` reports the lid open forever, so `omarchy-hw-clamshell` is never true and the poll has nothing to act on.

That combination also produces a symptom that is confusing in isolation: patch only the lid state and the daemon reconciles against its own stale `omarchy-hw-laptop` answer, so a disabled internal panel visibly re-enables about a second later. It reads like the compositor fighting the config, but it is this helper.

Verified locally with both fixes applied, physical lid close while docked to an external display:

```
LidClosed: b true        omarchy-hw-laptop-closed: rc=0
Docked:    b true        omarchy-hw-clamshell:     rc=0
eDP-1 disabled=true      HDMI-A-1 disabled=false
```

eDP-1 holds off across the reconciliation passes instead of flapping, and the session correctly neither locks nor suspends while docked. `omarchy-hw-laptop` returning true also restores Laptop Display, Mirror Display and Battery Percentage in the menu.

Filed #10477 for this before finding the PR; closing it as covered by this.

cristim:
Confirming this is still live on `quattro` today, and reproducing the gotcha directly on affected hardware.

**Machine:** MacBookPro17,1 (M1), Asahi, aarch64. No ACPI lid button, so the DMI fallback is the only branch that can answer.

```console
$ printf '[%s]\n' "$(< /sys/class/dmi/id/chassis_type)"
[9]
$ printf '[%s]\n' "$(< /sys/class/dmi/id/chassis_type 2>/dev/null)"
[]
$ omarchy-hw-laptop; echo $?
1
```

`chassis_type` is 9 (Laptop) and reads fine, but adding `2>/dev/null` takes the substitution off bash's `$(<file)` fast path and it expands empty, so the `case` matches nothing and the helper reports "not a laptop" on an actual laptop. `$(cat ...)` as proposed here fixes it.

Downstream effect, since the helper is a guard: `omarchy-hyprland-monitor-watch:78` gates on it, and three `omarchy-menu.jsonc` entries are hidden by it (Laptop Display, Mirror Display, Battery Percentage), so those are silently unavailable on any machine without an ACPI lid button.

Worth noting this is a *separate* helper from `omarchy-hw-laptop-closed` and does not overlap #7951 / #9754: nothing in the lid-close or clamshell path gates on `omarchy-hw-laptop`, so this fix and those are independent and both needed.

PR #7950, opened 2026-08-23
Rebind the T1 Touch Bar's display sub-device so it doesn't stay dark

csk-grit42:
## Why

On T1 Touch Bar MacBook Pros (`MacBookPro13,2`/`13,3`/`14,2`/`14,3`), once `apple-ib-tb` is loaded and the iBridge's `05AC:8600` HID interfaces are bound to `apple-ibridge-hid`, the driver spawns a second, virtual HID sub-device (`1D6B:0301`) carrying the "display on" field. The bar's icon row can already be lit at this point, but the bar itself stays completely dark until that sub-device is *also* explicitly bound, this time to `apple-ib-touchbar`. It does not reliably win its own auto-bind race, and nothing logs the failure. The whole chain re-enumerates on resume too, so this triggers off USB add rather than module add.

This is additive to #7064, which installs the driver stack and binds the two real `05AC:8600` interfaces, but does not yet rebind the virtual sub-device — a T1 Touch Bar can still ship dark even with that fix applied. This PR only adds the missing rebind; it needs a T1 Touch Bar driver already present to have anything to act on.

Diagnosed and field-verified (Part 2 of the linked recipe) on a MacBookPro14,2: [csk-grit42/OmarchyOnMacBookPro14.2-2017---A1706- — HOWTO.md](https://github.com/csk-grit42/OmarchyOnMacBookPro14.2-2017---A1706-/blob/main/HOWTO.md).

> **Model note:** built and verified on one MacBookPro14,2 unit. T1 firmware and HID interface numbering can vary slightly by production batch/model (13,2 / 13,3 / 14,3) — the bind targets a bus-ID pattern rather than a fixed path, but if it doesn't take effect on a different T1 model, check the actual `lsusb -t` / `/sys/bus/hid/devices/` layout before assuming the fix itself is wrong.

## What

- `bin/omarchy-hw-apple-t1-touchbar-disp-bind` — force-binds the `1D6B:0301` sub-device to `apple-ib-touchbar` once it appears (short retry loop; it's spawned asynchronously right after the parent bind).
- `default/udev/apple-t1-touchbar-disp.rules` — runs it on every `05AC:8600` USB add, covering boot and resume.
- `install/hardware/apple/t1-touchbar-disp-bind.sh` — shared `needed` / `installed` / `install` helpers for the leaf hook and migration.
- `install/hardware/apple/fix-t1-touchbar-disp-bind.sh` — install-time hook, wired into `install/hardware/all.sh` right after the SPI keyboard hook.
- Migration `1787508990` does the same for existing installs.
- `test/shell.d/apple-t1-touchbar-disp-bind-test.sh` covers the hardware gate, the rebind logic against a faked `/sys/bus/hid` tree, and the install/migration wiring.

No change to `manual/44-mac-support.md` — the Touch Bar isn't fully wired up on `quattro` without #7064 (or an equivalent driver install) landing first; this only removes the second failure mode once that's in place.

## Test plan

- [x] `./test/cli`
- [x] `test/shell.d/apple-t1-touchbar-disp-bind-test.sh`
- [x] MacBookPro14,2: with #7064's driver stack and Stage 1 bind applied manually, this rebind step is what takes the bar from icons-only to fully lit; without it, the bar stays dark indefinitely.

### DIFF
--- a/bin/omarchy-hw-apple-t1-touchbar-disp-bind
+++ b/bin/omarchy-hw-apple-t1-touchbar-disp-bind
@@ -1,0 +1,30 @@
+#!/bin/bash
+
+# omarchy:summary=Force-bind the T1 Touch Bar's virtual display sub-device
+# omarchy:hidden=true
+
+# apple-ib-tb spawns a virtual HID sub-device (1D6B:0301) once the iBridge's
+# real 05AC:8600 interfaces are bound to apple-ibridge-hid. That sub-device
+# carries the "display on" field, but it does not reliably win its own
+# auto-bind race — the bar then shows icons but stays dark, with nothing
+# logged anywhere. It appears asynchronously right after the parent bind, so
+# this retries briefly rather than assuming it's already there.
+hid_root=${OMARCHY_T1_HID_ROOT:-/sys/bus/hid}
+
+for _ in $(seq 1 15); do
+  found=0
+  for dev in "$hid_root"/devices/0003:1D6B:0301.*; do
+    [[ -d $dev ]] || continue
+    found=1
+    name=$(basename "$dev")
+    drv=""
+    [[ -e $dev/driver ]] && drv=$(basename "$(readlink -f "$dev/driver")")
+    [[ $drv == apple-ib-touchbar ]] && continue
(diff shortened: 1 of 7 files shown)


smoothunit:
https://github

PR #7979, opened 2026-08-24
Address a named workspace by name when toggling its layout

### LINKED ISSUE #7961: omarchy-hyprland-workspace-layout-toggle silently no-ops on named workspaces (rule keyed by a reassigned workspace id)
### System details

AMD Ryzen 5 3500U / Radeon Vega Mobile, Omarchy 4.0.0.alpha, Hyprland 0.56.2

### What's wrong?

**Summary**

`omarchy-hyprland-workspace-layout-toggle` keys its persisted workspace rule by
workspace **id**. Hyprland gives *named* workspaces negative ids and reassigns
them every session, so the rule matches no workspace. `hyprctl eval` still
answers `ok`, so the script's `||` fallback never runs and it sends a success
notification for a change that did not happen.

**Mechanism**

```bash
ACTIVE_WORKSPACE=$(hyprctl activeworkspace -j | jq -r '.id')
[[ $ACTIVE_WORKSPACE =~ ^-?[0-9]+$ ]] || exit 1
...
printf 'hl.workspace_rule({ workspace = "%s", layout = "%s" })\n' "$ACTIVE_WORKSPACE" "$NEW_LAYOUT" >"$LAYOUT_FILE"
hyprctl eval "hl.workspace_rule({ workspace = \"$ACTIVE_WORKSPACE\", layout = \"$NEW_LAYOUT\" })" >/dev/null 2>&1 || \
  hyprctl keyword workspace "$ACTIVE_WORKSPACE, layout:$NEW_LAYOUT"
```

The `^-?[0-9]+$` guard accepts the negative id, so nothing fails loudly.

**Evidence**

Same workspace, same moment, two rules differing only in the selector:

```
# what the script wrote
hl.workspace_rule({ workspace = "-1340", layout = "dwindle" })
# tiledLayout afterwards: scrolling      <- unchanged

# same workspace, addressed by name
hl.workspace_rule({ workspace = "name:<monitor description>:1", layout = "scrolling" })
# tiledLayout afterwards: scrolling      <- applied
```

Directly:

```
$ hyprctl eval 'hl.workspace_rule({ workspace = "-1340", layout = "scrolling" })'
ok
$ hyprctl activeworkspace -j | jq -r .tiledLayout
dwindle
```

**It also never toggles back.** `CURRENT_LAYOUT` is re-read from `.tiledLayout`,
which never changed, so every press computes the same `NEW_LAYOUT`. Three
consecutive presses left three state files all saying `scrolling`, never
`dwindle`.

**The persisted ids are not stable.** Across a reboot two named workspaces
traded ids: `-1337` and `-1340` swapped between monitors. If these rules ever
did match, a reboot would apply one monitor's saved layout to the other.

### Steps to reproduce

1. Focus a named workspace — a special workspace, or one created with
   `hyprctl dispatch workspace name:foo`.
2. Press `SUPER + L`.
3. The notification reports the new layout, but
   `hyprctl activeworkspace -j | jq -r .tiledLayout` is unchanged.
4. Press it again. It reports the same layout again, never the other one.

### Expected behavior

The layout changes, and repeated presses alternate between `dwindle` and
`scrolling`.

### Related

- #4881 reports the same root cause for special workspaces, open since March,
  and reaches the same conclusion: identify the workspace by name, not id.
- PR #4880 implemented exactly that and was closed in favour of #5099, which is
  still open and unmerged.
- #7612 documents the same `<id>.lua` persistence from a different angle.

Keying the rule by `name:<workspace name>` fixes all three symptoms and is what
#4880 already proposed. A cheap independent improvement: re-read `.tiledLayout`
after the `eval` and report failure instead of success when it did not change —
the current notification is what makes this hard to diagnose.

Chessing234:
Fixes #7961 (thanks @alebak — the diagnosis there is exact).

`omarchy-hyprland-workspace-layout-toggle` keys its `workspace_rule` on `.id`. Hyprland gives named workspaces a **negative** id and reassigns it every session, so the rule matches no workspace. `hyprctl eval` answers `ok` regardless of whether a rule matched anything, so the `||` fallback never fired and the toggle sent a success notification for a change that did not happen.

Three consequences, all in the issue:

- the layout never changes on a named workspace;
- because `CURRENT_LAYOUT` is re-read from `.tiledLayout`, which never moved, every press computes the same target — it never toggles back;
- the persisted state file is named after the id, and ids are reassigned, so two named workspaces can swap rules across a reboot.

This selects named workspaces with `name:<name>` — what the user chose, and stable — and keys the state file on the name too.

Two smaller things came with it:

- **The layout is read back after applying.** `eval`'s `ok` carries no information about whether the rule matched, so reading `.tiledLayout` is the only honest confirmation. A rule that changed nothing now reports that and exits non-zero instead of claiming success.
- **`hyprctl activeworkspace` is queried once.** It was called twice, and id, name and layout have to describe the same workspace — the second call can land after the user has switched away.

The state filename becomes a Lua module name (`require_all.files` requires each `*.lua` under the state dir), so the name is reduced to `[A-Za-z0-9_-]` for the filename while the rule body keeps the real name. The selector is also escaped for the double-quoted Lua string it is interpolated into.

### Evidence

`test/shell.d/hyprland-workspace-layout-test.sh` gains a stub that models the behaviour that made this invisible: `eval` returns `ok` either way, and the layout only actually moves when the selector is the one Hyprland accepts. Cases: numeric workspace (unchanged behaviour), named workspace (selector is `name:…`, state file is `name-…`, and *no* `-1340.lua` is written), and a rule that matches nothing (non-zero exit, no success notification).

Reverting only the script and re-running the test fails at the named-workspace case, so the new coverage has teeth.

`bin-style`, `hyprland-workspace-layout`, `hyprland-default-config` and `hyprland-binding-conflicts` all pass.

### What I could not run

I developed this on macOS, so no Hyprland: the stub covers the script's logic, not Hyprland's acceptance of the `name:` selector — that comes from the issue, where `hl.workspace_rule({ workspace = "name:<monitor description>:1", … })` is shown applying while the id form does not.

The test's Lua block also can't execute here — `require_all.files` shells out to `find -printf`, and Lua's `popen` picks up BSD `find`. Separately, I found that block cannot fail the test on *any* platform: `lua` reading a script from stdin exits 0 even on an uncaught error (`lua -e 'assert(false)'` exits 1; `lua <<EOF` with the same assert exits 0). Five test files use `lua <<`, so that is its own fix and I've kept it out of this PR.

This was AI-assisted.

### DIFF
--- a/bin/omarchy-hyprland-workspace-layout-toggle
+++ b/bin/omarchy-hyprland-workspace-layout-toggle
@@ -41,12 +41,21 @@ LAYOUT_FILE="$LAYOUTS_DIR/$STATE_KEY.lua"
 mkdir -p "$LAYOUTS_DIR"
 printf 'hl.workspace_rule({ workspace = "%s", layout = "%s" })\n' "$ESCAPED_SELECTOR" "$NEW_LAYOUT" >"$LAYOUT_FILE"
 
+# Named workspaces used to persist as <id>.lua. require_all.files globs every
+# *.lua at -maxdepth 1, so leaving that file behind loads both the new name
+# rule and a leftover keyed on an id Hyprland will reassign next session.
+if (( ACTIVE_WORKSPACE_ID < 0 )); then
+  rm -f "$LAYOUTS_DIR/${ACTIVE_WORKSPACE_ID}.lua"
+fi
+
 hyprctl eval "hl.workspace_rule({ workspace = \"$ESCAPED_SELECTOR\", layout = \"$NEW_LAYOUT\" })" >/dev/null 2>&1 ||
   hyprctl keyword workspace "$WORKSPACE_SELECTOR, layout:$NEW_LAYOUT"
 
 # `hyprctl eval` reports ok for a rule that matched no workspace, so the only
-# honest confirmation is reading the layout back.
-APPLIED_LAYOUT=$(hyprctl activeworkspace -j | jq -r '.tiledLayout // empty')
+# honest confirmation is reading the layout back — from the workspace we
+# applied to, not whatever is active now. The user can switch away between
+# apply and this query.
+APPLIED_LAYOUT=$(hyprctl workspaces -j | jq -r --arg n "$ACTIVE_WORKSPACE_NAME" '.[] | select(.name==$n) | .tiledLayout // empty')
 if [[ $APPLIED_LAYOUT != "$NEW_LAYOUT" ]]; then
   omarchy-notification-send -u critical -g 󱂬 "Could not set workspace layout to $NEW_LAYOUT"
   exit 1
--- a/bin/omarchy-hyprland-workspace-layout-toggle
+++ b/bin/omarchy-hyprland-workspace-layout-toggle
@@ -42,11 +42,11 @@ mkdir -p "$LAYOUTS_DIR"
 printf 'hl.workspace_rule({ workspace = "%s", layout = "%s" })\n' "$ESCAPED_SELECTOR" "$NEW_LAYOUT" >"$LAYOUT_FILE"
 
 # Named workspaces used to persist as <id>.lua. require_all.files globs every
-# *.lua at -maxdepth 1, so leaving that file behind loads both the new name
-# rule and a leftover keyed on an id Hyprland will reassign next session.
-if (( ACTIVE_WORKSPACE_ID < 0 )); then
-  rm -f "$LAYOUTS_DIR/${ACTIVE_WORKSPACE_ID}.lua"
-fi
+# *.lua at -maxdepth 1. Ids are reassigned each session, so a leftover is not
+# necessarily named after the id this workspace has now. Under the new scheme a
+# negative id always persists as name-<...>, so every bare negative filename
+# left in here is from the old scheme and safe to drop.
+rm -f "$LAYOUTS_DIR"/-[0-9]*.lua
 
 hyprctl eval "hl.workspace_rule({ workspace = \"$ESCAPED_SELECTOR\", layout = \"$NEW_LAYOUT\" })" >/dev/null 2>&1 ||
   hyprctl keyword workspace "$WORKSPACE_SELECTOR, layout:$NEW_LAYOUT"
--- a/bin/omarchy-hyprland-workspace-layout-toggle
+++ b/bin/omarchy-hyprland-workspace-layout-toggle
@@ -2,20 +2,54 @@
 
 # omarchy:summary=Toggle the layout on the current active workspace between dwindle and scrolling
 
-ACTIVE_WORKSPACE=$(hyprctl activeworkspace -j | jq -r '.id')
-[[ $ACTIVE_WORKSPACE =~ ^-?[0-9]+$ ]] || exit 1
-CURRENT_LAYOUT=$(hyprctl activeworkspace -j | jq -r '.tiledLayout')
-LAYOUTS_DIR="$HOME/.local/state/omarchy/workspace-layouts"
-LAYOUT_FILE="$LAYOUTS_DIR/$ACTIVE_WORKSPACE.lua"
+# One query: id, name and layout have to describe the same workspace, and a
+# second call can land after the user has already switched away.
+ACTIVE_WORKSPACE_JSON=$(hyprctl activeworkspace -j)
+ACTIVE_WORKSPACE_ID=$(jq -r '.id // empty' <<<"$ACTIVE_WORKSPACE_JSON")
+ACTIVE_WORKSPACE_NAME=$(jq -r '.name // empty' <<<"$ACTIVE_WORKSPACE_JSON")
+[[ $ACTIVE_WORKSPACE_ID =~ ^-?[0-9]+$ ]] || exit 1
+CURRENT_LAYOUT=$(jq -r '.tiledLayout // empty' <<<"$ACTIVE_WORKSPACE_JSON")
+
+# Hyprland gives named workspaces a negative id and reassigns it every session,
+# so a rule keyed by that id matches nothing — and `hyprctl eval` still answers
+# "ok", so the failure is silent. Address those by name instead, which is what
+# the user chose and what survives a reboot.
+if (( ACTIVE_WORKSPACE_ID < 0 )); then
+  [[ -n $ACTIVE_WORKSPACE_NAME ]] || exit 1
+  WORKSPACE_SELECTOR="name:$ACTIVE_WORKSPACE_NAME"
+  # The state filename becomes a Lua module name, so keep it to characters a
+  # module path can hold. Names collide only if they differ solely by the
+  # characters being replaced.
+  STATE_KEY="name-$(printf '%s' "$ACTIVE_WORKSPACE_NAME" | tr -c 'A-Za-z0-9_-' '_')"
+else
+  WORKSPACE_SELECTOR="$ACTIVE_WORKSPACE_ID"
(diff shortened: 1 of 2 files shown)


alebak:
Thanks for picking this up so quickly, and for the write up. The reasoning about the reassigned negative id is exactly right.

One thing I can offer, since you flagged it as untestable: I run Hyprland with named workspaces on every monitor, so I checked the `name:` selector on a live session. It works. My state files already use that form and the rules land where they should:

```
id=-1340  name=LG Electronics LG HDR WFHD 0x0001EF39:1  layout=scrolling   <- has a name: rule
id=-1345  name=LG Electronics LG HDR WFHD 0x0001EF39:3  layout=dwindle     <- no rule, falls back to default
```

So the core of the change holds up on real Hyprland.

Two suggestions, both from hitting this myself:

**Stale id-keyed files stay behind.** The filename moves from `<id>.lua` to `name-<sanitized>.lua`, but the old file is never removed, and `require_all.files` globs every `*.lua` at `-maxdepth 1`. After upgrading, a named workspace ends up with both rules loaded, and the leftover one is still keyed on an id Hyprland reassigns each session. That is the third consequence from the issue, still live after the fix. Removing the matching `<id>.lua` when writing the new file would close it.

**The readback re-queries the active workspace.** You make a good case for a single query, since a second call can land after the user has switched away. The verification at the end reintroduces that: if the workspace changes between apply and readback, it reads a different one, then sends a critical notification and exits non-zero for a change that actually worked. Reading back by selector avoids it:

```bash
hyprctl workspaces -j | jq -r --arg n "$ACTIVE_WORKSPACE_NAME" '.[] | select(.name==$n) | .tiledLayout'
```

Happy to test a revision on real hardware if that would help.

Chessing234:
removed the leftover id-keyed file when writing the name-based one, and read the layout back by workspace name so switching away mid-apply no longer fails a change that worked.

bin/omarchy-hyprland-workspace-layout-toggle
```diff
@@ -2,20 +2,63 @@
...
 mkdir -p "$LAYOUTS_DIR"
-printf 'hl.workspace_rule({ workspace = "%s", layout = "%s" })\n' "$ACTIVE_WORKSPACE" "$NEW_LAYOUT" >"$LAYOUT_FILE"
+printf 'hl.workspace_rule({ workspace = "%s", layout = "%s" })\n' "$ESCAPED_SELECTOR" "$NEW_LAYOUT" >"$LAYOUT_FILE"
+
+# Named workspaces used to persist as <id>.lua. require_all.files globs every
+# *.lua at -maxdepth 1, so leaving that file behind loads both the new name
+# rule and a leftover keyed on an id Hyprland will reassign next session.
+if (( ACTIVE_WORKSPACE_ID < 0 )); then
+  rm -f "$LAYOUTS_DIR/${ACTIVE_WORKSPACE_ID}.lua"
+fi
```
alebak, line 49: Ids are reassigned every session, so this deletes the file matching the *current* id, which is usually another workspace's leftover, while this workspace's own leftover keeps its old id and survives.

Sweeping every negative-id file is both simpler and complete: under the new scheme a negative id always produces `name-...`, so a bare `-<digits>.lua` can only be from the old one.

```suggestion
# Named workspaces used to persist as <id>.lua, and require_all.files globs
# every *.lua at -maxdepth 1. Ids are reassigned each session, so a leftover
# is not necessarily named after the id this workspace has now. Under the new
# scheme a negative id always persists as name-<...>, so every bare negative
# filename left in here is from the old scheme and safe to drop.
rm -f "$LAYOUTS_DIR"/-[0-9]*.lua
```

alebak (commented):
Both fixes look right, and the call-counting stub for the readback is a nice touch. Two things from running the branch locally.

**The cleanup removes the wrong file.** `rm -f "$LAYOUTS_DIR/${ACTIVE_WORKSPACE_ID}.lua"` uses the current id, but ids get reassigned every session, which is the premise of the bug. The leftover for this workspace is named after an id it no longer has, and the file that does match the current id usually belongs to a different workspace. I ran the branch against a state dir holding both:

```
before          after toggling "LG:1" (current id -1340)
-1337.lua       -1337.lua       <- this workspace's leftover, survives
-1340.lua       name-LG_1.lua   <- new file, correct
```

It deleted another workspace's leftover and kept the harmful one. The inline suggestion sweeps every negative-id file instead: under the new scheme a negative id always produces `name-...`, never a bare `<id>.lua`, so all of them are stale by construction. I checked it leaves `1.lua`, `3.lua` and `name-*.lua` untouched.

**The Lua block reads the real user's state dir.** `default/hypr/paths.lua` resolves `state_home` as `os.getenv("XDG_STATE_HOME") or (home .. "/.local/state")`, and the block sets `HOME` but not `XDG_STATE_HOME`. On a machine that sets it (Omarchy does) it loads the developer's own `~/.local/state/omarchy/workspace-layouts`. Running the suite on my Omarchy box:

```
no file './omarchy/workspace-layouts/LG_Electronics_LG_HDR_WFHD_0x0001EF39_1.lua'
stack traceback:
	[C]: in global 'require'
	...default/hypr/require_all.lua:27: in field 'files'
ok - saved workspace layouts load into Hyprland configuration
exit: 0
```

That filename is from my real home, and the suite still reported ok, which is the `lua <<` behaviour you documented. One variable fixes the isolation half:

```bash
HOME="$home_dir" XDG_STATE_HOME="$home_dir/.local/state" OMARCHY_PATH="$ROOT" lua <<'LUA'
```

Small one while you are in there: the leftover test asserts `[[ -f $named_file ]]`, but that file is already on disk from the earlier named-workspace case, so it passes whether or not this run writes anything.

bin/omarchy-hyprland-workspace-layout-toggle
Chessing234: swept every leftover -[0-9]*.lua instead of the current id, and pointed the lua harness at the test state dir.

Chessing234:
swept every leftover -[0-9]*.lua instead of the current id, and pointed the lua harness at the test state dir.

alebak:
Sorry for the slow reply, your update slipped past me.

All three are addressed, and I ran the branch on real Hyprland, so the part you could not cover is covered now.

**Live, on named workspace `LG Electronics LG HDR WFHD 0x0001EF39:2` (id `-1346`):**

```
toggle #1: scrolling -> dwindle   exit 0
toggle #2: dwindle -> scrolling   exit 0
```

It toggles back, which was the second consequence in the issue. The `name:` selector applies, the readback confirms it, and a planted `-1337.lua` was swept while the numeric and `name-` files were left alone. Ids had been reassigned again since my last comment (the workspace I showed there as `-1340` is `-1348` today), so the premise keeps proving itself.

**Both regressions reproduce, so the new coverage has teeth:**

- The new test against the script at `bc520848` fails at `not ok - workspace layout toggle removes this workspace's leftover id-keyed file`, exit 1.
- Without `XDG_STATE_HOME`, the Lua block on this box discovers my real state files (`BOE_0x0DDE_1`) and still prints `ok` with exit 0, which is exactly the `lua <<` behaviour you documented. Clean on the current head.

`bash test/shell`: 186 of 190 files pass. The four that fail (`config`, `runtime-smoke`, `snapper`, `unowned-system-paths`) fail identically on `quattro`. Three want an `omarchy-pkgs` sibling checkout, one counts widget IPC handlers across two screens. Nothing from this PR.

Nothing blocking from me. Ready for a maintainer as far as I can tell.

Chessing234:

Chessing234:
folded into #12269 for easier review.

PR #7985, opened 2026-08-24
Create the Windows launcher entry only once the VM is running

### LINKED ISSUE #7943: Canceling Windows VM setup leaves launcher in installed state
## System details

- 11th Gen Intel Core i5-1145G7
- Intel Iris Xe Graphics
- Omarchy 4.0.0-1
- Linux 7.1.8-arch1-3, x86_64

## What is wrong?

Canceling the Windows VM installer at its final "Proceed with this configuration?" prompt leaves the Omarchy launcher in an installed state. The launcher then offers "Launch Windows" and "Remove Windows" instead of "Install Windows", even though no VM was created.

## Steps to reproduce

1. Open the Omarchy launcher and choose Install > Windows.
2. Complete the RAM, CPU, disk, username, and password prompts.
3. Answer No at "Proceed with this configuration?".
4. Reopen the Omarchy launcher.

## Actual behavior

The installer exits, but ~/.local/share/applications/windows-vm.desktop remains. The launcher uses this file as its installed-state marker, so Install > Windows disappears while Launch > Windows and Remove > Windows appear. No docker-compose.yml, omarchy-windows container, dockurr/windows image, or ~/Windows shared folder exists.

## Expected behavior

Canceling should leave the launcher offering Install > Windows and should not advertise a launchable/removable VM.

## Cause and suggested fix

In bin/omarchy-windows-vm, install_windows creates windows-vm.desktop before any configuration prompts. All later cancellation paths exit without removing it. Current upstream still has this ordering.

Create the desktop entry only after the final confirmation (ideally after the compose file is created and startup succeeds), or add cancellation/error cleanup that removes the premature marker.

## Workaround verified

Moving the stale windows-vm.desktop aside immediately restores Install > Windows. The two empty directories ~/.windows and ~/.config/windows can remain and are reused on retry.

Chessing234:
Fixes #7943.

`install_windows` wrote `~/.local/share/applications/windows-vm.desktop` at the top of the function — before the RAM, CPU, disk, username and password prompts, and about 120 lines before the final confirmation. The launcher uses that file as its installed-state marker, so from that moment the menu shows **Launch Windows** and **Remove Windows** and hides **Install Windows**.

Every exit after that point leaves it behind:

- answering No at "Proceed with this configuration?" (`exit 1`);
- Ctrl+C, via the `INT` trap installed at the top of the function;
- `docker-compose up -d` failing, which also `exit 1`s.

In each case no compose file, container, image or `~/Windows` exists, but the menu insists a VM does — and offers no way to install one.

The entry now lives in `create_launcher_entry`, called after `docker-compose` reports the VM started. Nothing before that claims an install exists. `mkdir -p "$HOME/.local/share/applications"` moved with it, since that directory is only needed to write the entry.

### Evidence

New `test/shell.d/windows-vm-launcher-test.sh`:

- the call to `create_launcher_entry` comes after both the `gum confirm` and the `docker-compose up -d` success check;
- `windows-vm.desktop` is written in exactly one place — a range check over the function, so re-adding an early `tee` fails even if the ordering assertions still pass;
- `create_launcher_entry` is evaluated on its own and asserted to produce an entry with a usable `Name=` and `Exec=`.

Against the current script the test fails with `windows vm install still has the landmarks this test relies on`, so it is genuinely pinned to the new structure.

`bin-style-test` passes.

### What I could not run

The install flow can't execute here: `check_prerequisites` requires `/dev/kvm`, and the prompts are `gum`, so the ordering is asserted against the source rather than by cancelling a real run. Someone on Linux answering No at the confirmation and checking that Install > Windows is still offered would confirm the user-visible half.

I did not touch `remove_windows`. It still runs a bare `rm` on the marker, which now has one more way to be absent — but that path is only reachable from a menu entry that requires the marker, so I left it rather than widen the diff.

This was AI-assisted.


Chessing234:
rebased onto quattro; launcher still only written after the vm actually starts

Chessing234:
folded into #12264 for easier review.

PR #7992, opened 2026-08-24
Stop refresh-config overwriting a config it could not back up

Adolanium:
`omarchy-refresh-config` says in its own summary that it backs up your version before overwriting, and `test/shell.d/refresh-config-test.sh` pins that contract. The code did not check whether the backup happened:

```bash
if [[ -f $user_config_file ]]; then
  cp -f "$user_config_file" "$backup_config_file"
  cp -f "$default_config_file" "$user_config_file"
```

There is no `set -e` and no test of the first copy's status, so a backup that fails for any reason, a full disk, a read-only mount, a permissions problem, falls straight through to the copy that replaces the user's file with the shipped default. The customizations are gone and there is nothing to go back to.

The message afterwards makes it worse. With no backup on disk, `cmp -s` cannot match, so the run takes the else branch and prints:

```
Replaced ~/.config/hypr/bindings.lua with new Omarchy default.
Saved backup as ~/.config/hypr/bindings.lua.bak.1700000000.
```

That file does not exist. The user is told their version was kept at the exact moment it was destroyed.

Separately, the backup name is `$user_config_file.bak.$(date +%s)`, which is only unique per second. Refreshes are triggered from updates, menus and migrations, so two runs against the same file can overlap, and the second `cp -f` then overwrites the first one's backup without a word.

What changed:

- Both copies are checked. A failed backup stops the run with a message saying the user's file was left alone, and the file really is left alone. A failed write stops the run and says where the backup is.
- `set -e` at the top, so the `mkdir -p` and the copy on the no-existing-config path cannot fail quietly either.
- The backup name gains the pid, so two refreshes in the same second get two names.

Tests: two cases added to `test/shell.d/refresh-config-test.sh`. The first stubs `cp` to fail only on the `.bak.` destination and checks the run exits non-zero, the user's config still holds its own content, and the error says why. The second pins the clock with a `date` stub so the collision is certain rather than a matter of timing, refreshes the same file twice with different content, and checks both backups survive. Both fail against the old script, the second one reporting one backup where two were expected. Four cases in the file now, all pass. `./test/shell` comes out the same on this branch as on the base commit, 1748 passing here with the 2 new, same 52 failing files either way. Those 52 are tools missing from my environment, `jq` and `lua` and `magick` and others, and not anything this touches. `./test/cli` needs `jq` and would not run on this machine at all.

### DIFF
--- a/bin/omarchy-refresh-config
+++ b/bin/omarchy-refresh-config
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -e
+
 # omarchy:summary=Copy a shipped user config from $OMARCHY_PATH/config into ~/.config (backs up your version).
 # omarchy:args=<config-path>
 
@@ -19,7 +21,9 @@ fi
 
 user_config_file="${HOME}/.config/$config_file"
 default_config_file="$OMARCHY_PATH/config/$config_file"
-backup_config_file="$user_config_file.bak.$(date +%s)"
+# The pid keeps two refreshes of the same file in the same second from landing
+# on one name, where the second cp would overwrite the first one's backup.
+backup_config_file="$user_config_file.bak.$(date +%s).$$"
 
 if [[ ! -e $default_config_file ]]; then
   echo "Not a shipped user config: $config_file" >&2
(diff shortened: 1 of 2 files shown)


sprajs:
Independently checked head `522d8cb2` in an isolated temporary home. All four `test/shell.d/refresh-config-test.sh` checks pass, including the forced backup-copy failure and two refreshes with a fixed clock. I also reproduced the same-second backup loss against current `quattro` (`d174d4aa`) before testing this patch. This PR already covers that failure, so there is no need for a separate collision fix.

This was a focused mocked test, not a full-suite or live-config validation.

PR #7995, opened 2026-08-24
Stage diagnostics logs privately instead of at fixed /tmp paths

Adolanium:
`omarchy-debug` gathers the machine's diagnostics into a file at a fixed name:

```bash
LOG_FILE="/tmp/omarchy-debug.log"

if [[ $NO_SUDO = "true" ]]; then
  DMESG_OUTPUT="(skipped - --no-sudo flag used)"
else
  DMESG_OUTPUT="$(sudo dmesg)"
fi

cat > "$LOG_FILE" <<EOF
```

The redirect creates that file under the caller's umask, which is 022, so it lands mode 0644 in a directory every local user can reach. The interesting part is what goes in it. `sudo dmesg` is elevated on purpose, because the kernel ring buffer is not readable by an ordinary user, and the script then writes it somewhere an ordinary user can read. The boot journal at `-p 4..1` and the full package inventory go in alongside it. Nothing removes the file, so the last run's log sits there until reboot.

`omarchy-upload-log` does the same thing with `/tmp/upload-log.txt` and `/tmp/system-info.txt`. Those hold the assembled bundle, including the install logs and the journal, and are likewise never removed.

The symlink and pre-created-file variants of this are not the issue. `fs.protected_symlinks` and `fs.protected_regular` are on by default, so another user planting either path makes the redirect fail rather than succeed somewhere unintended. What is left is plain disclosure: a predictable name, a readable mode, and content the machine went out of its way to elevate for. Omarchy is a single-user desktop, so this only matters on machines with more than one local account.

All three now go through `mktemp`, which creates at 0600 under a name nobody can guess, with a trap clearing them on the way out. Every use of the files is a read by path, so nothing else in either script had to change.

Two more fixed `/tmp` paths in `bin/` get the same treatment:

- `omarchy update` wrote its transcript to `/tmp/omarchy-update.log`. It is read back by `omarchy-update-analyze-logs` after the run, so it keeps a stable name, but that name moves to `~/.local/state/omarchy/update.log`. `docs/update-process.md` names the new path.
- `omarchy-install-gaming-battlenet` wrote the installer log to `/tmp/omarchy-battlenet-installer.log`. It is written by a detached process after the script exits, so a trap cannot clean it up. It now lives at `~/.cache/omarchy/battlenet-installer.log`, next to the installer itself. `~/.cache` is already 0700 on Arch, so a fixed name is fine there and each run overwrites the last log.

Moving these two off `/tmp` has a side benefit. With `fs.protected_regular` on, a second local user running the same command got "Permission denied" on the file the first user left behind. Per-user paths remove that collision.

Two agent skill docs mentioned `/tmp/omarchy-debug.log` as a side effect of `omarchy debug --no-sudo --print`. The primary route there is `--print`, which puts the same content on stdout, and the interactive menu still offers "Save in current directory" for anyone who wants a file. Both mentions are updated rather than left pointing at a path that no longer exists.

Tests: new `test/shell.d/diagnostics-staging-test.sh` stubs the collectors and points `TMPDIR` at a directory the test owns. It checks `--print` still prints what was gathered, that the run removes the file it staged, that nothing is left behind, and that neither script names a `/tmp` literal. The cleanup case is checked through an `rm` stub that records what it was asked to remove, because an empty staging directory on its own would also pass for a script that never staged there, which is exactly what the old fixed path did. Further cases cover the transcript landing under `XDG_STATE_HOME`, the fallback to `~/.local/state` when it is unset, `omarchy-update-analyze-logs` reading from the new path and staying quiet on a good run, and the Battle.net log pointing at the fixed name in the cache directory. Twelve cases, all pass. I ran `test/shell.d/diagnostics-staging-test.sh` directly; the rest of `./test/shell` and `./test/cli` depend on tools my environment does not have (`jq`, `lua`, `magick`), so I did not run them.

### DIFF
--- a/bin/omarchy-debug
+++ b/bin/omarchy-debug
@@ -26,7 +26,12 @@ while (( $# > 0 )); do
   esac
 done
 
-LOG_FILE="/tmp/omarchy-debug.log"
+# The log holds sudo dmesg output and the boot journal. At a fixed name in /tmp
+# it was created with the default umask, so every local user on the machine
+# could read it, and it stayed there after the command exited. mktemp makes it
+# 0600 under a name nobody can guess, and the trap clears it on the way out.
+LOG_FILE=$(mktemp) || exit 1
+trap 'rm -f "$LOG_FILE"' EXIT
 
 if [[ $NO_SUDO = "true" ]]; then
   DMESG_OUTPUT="(skipped - --no-sudo flag used)"
--- a/bin/omarchy-install-gaming-battlenet
+++ b/bin/omarchy-install-gaming-battlenet
@@ -57,7 +57,13 @@ app launcher.
 
 EOF
 
-  log="/tmp/omarchy-battlenet-installer.log"
+  # The log is written by the detached installer after this script exits, so it
+  # cannot be cleaned up here. At a fixed name in /tmp another local account
+  # could read it, and with fs.protected_regular a second user running this
+  # command got "Permission denied" on the file the first user left. The cache
+  # directory is private to the user, so a fixed name here is fine and each run
+  # overwrites the last log instead of leaving a new file behind.
+  log="$cache_dir/battlenet-installer.log"
   setsid -f sh -c "umu-run '$installer' >'$log' 2>&1" </dev/null >/dev/null 2>&1
   echo "Installer log: $log"
   launched_installer=1
--- a/bin/omarchy-update
+++ b/bin/omarchy-update
@@ -8,8 +8,15 @@
 set -e
 
 if [[ -z ${OMARCHY_UPDATE_LOGGED:-} ]]; then
+  # The transcript is read back by omarchy-update-analyze-logs after the run,
+  # so it needs a stable name rather than mktemp. At a fixed name in /tmp
+  # another local account could read the output, and with fs.protected_regular
+  # a second user running the update got "Permission denied" on the file the
+  # first user left. The user's own state directory avoids both.
+  update_log="${XDG_STATE_HOME:-$HOME/.local/state}/omarchy/update.log"
+  mkdir -p "$(dirname "$update_log")"
   script_command=$(printf '%q ' "$0" "$@")
-  exec env OMARCHY_UPDATE_LOGGED=1 script -qefc "$script_command" "/tmp/omarchy-update.log"
+  exec env OMARCHY_UPDATE_LOGGED=1 script -qefc "$script_command" "$update_log"
 fi
 
 if ! omarchy-update-lock held; then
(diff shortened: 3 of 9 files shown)


Adolanium:
The same treatment now covers the other two fixed /tmp paths in bin/:

- `omarchy update` wrote its whole transcript to `/tmp/omarchy-update.log`. That one is read back by `omarchy-update-analyze-logs` after the run, so it keeps a stable name, but the name moves to `~/.local/state/omarchy/update.log`, where no other local user can pre-create it or read along. `docs/update-process.md` names the new path.
- `omarchy-install-gaming-battlenet` wrote the installer log to `/tmp/omarchy-battlenet-installer.log`. It is written by a detached process after the script exits, so a trap cannot clean it up. It now stages with mktemp inside the user's own cache directory, next to the installer itself.

Tests: `diagnostics-staging-test.sh` grows from 4 to 12 cases. The new ones cover the transcript landing under XDG_STATE_HOME, the fallback to ~/.local/state when it is unset, analyze-logs reading from the new path and staying quiet on a good run, and the Battle.net log reaching the detached writer at an unguessable name in the cache directory. All pass. `./test/shell` shows the same failing files before and after this commit. Those are tools missing from my environment and not anything this touches.

Adolanium:
One more round on the comments and the Battle.net log.

A correction to my earlier comment: I said another local user could "pre-create it as a link". That is wrong on a default systemd system. `fs.protected_symlinks` and `fs.protected_regular` make the redirect fail instead. The real problems with the fixed /tmp names were that another local account could read the file, and that a second user running the same command got "Permission denied" on the file the first user left. The code comments in `omarchy-update`, `omarchy-update-analyze-logs`, and `omarchy-install-gaming-battlenet` now say that instead.

The Battle.net log drops `mktemp`. `~/.cache` is already 0700 on Arch, so a random name there added nothing except a new leftover file per run. It is now a fixed `~/.cache/omarchy/battlenet-installer.log`, overwritten each run. The test checks for that name. `diagnostics-staging-test.sh` still has 12 cases, all passing.

Chessing234:
hey, this is already covered in https://github.com/omacom/omarchy/pull/12109 (private diagnostics staging out of shared /tmp). mind closing this as a duplicate?

Adolanium:
This pull request was opened on August 24, three weeks before #12109.

#12109 leaves /tmp/omarchy-battlenet-installer.log in place. The detached installer writes that log after the script exits, so it is still world-readable on quattro.

The debug log, the upload bundle, and the update transcript overlap, and the fixes differ. This branch removes the debug and upload files on exit. The update transcript stays at ~/.local/state/omarchy/update.log so it is still there after logout. The branch now checks a failed debug write, creates that state directory mode 0700, makes the transcript mode 0600, sets ~/.cache/omarchy to mode 0700 before the Battle.net log is written, and migrates leftovers this user still owns in /tmp.

PR #8011, opened 2026-08-24
Let themes carry a declarative Neovim integration

### LINKED ISSUE #7942: Let repo-installed themes supply themed/*.tpl template overrides
## Problem

Introduced by `Stop an installed theme from running code`
([basecamp/omarchy@ef6d9e6](https://github.com/basecamp/omarchy/commit/ef6d9e6605b121df15bf310e630e04f0c1119fc8),
PR [#7884](https://github.com/basecamp/omarchy/pull/7884), merged 2026-08-23,
shipped as omarchy 4.0.0.r1803).

`omarchy-theme-set` drops every `.lua`, terminal config (`kitty.conf`,
`foot.ini`, `ghostty.conf`, `alacritty.toml`) and `vscode.json` from themes
installed from a git repo, then regenerates them from
`$OMARCHY_PATH/default/themed/*.tpl`.

That's the right security call for code, but it also means a theme author
cannot ship a Neovim integration at all — every repo theme gets the same
generic aether render, even when the author maintains a matching colorscheme
plugin. Community themes on omarchy.org therefore can never feel complete:
bar, terminals' palettes, and shell surfaces follow the theme, but Neovim,
terminal niceties (e.g. inactive border colors), and the VS Code extension
silently fall back to stock.

## Theme updates are broken by the same change

`omarchy theme update` only runs `git pull` inside the clone — it never
re-stages, so pulled changes stay invisible until the user happens to
re-select the theme. And when a re-stage does happen, the new filter strips
the same `.lua` / terminal configs / `vscode.json` all over again. Net
effect: a repo-installed theme can't deliver an update end-to-end either —
neither automatically nor through the supported flow.

Whatever lands from this proposal should also make `omarchy theme update`
re-stage (or simply invoke `omarchy-theme-set`) after pulling, and the
staging rules should leave room for themes to ship their integrations, so
updates actually reach the desktop.

## Proposal

Let an installed theme carry a `themed/` directory of templates:

```
~/.config/omarchy/themes/<name>/themed/neovim.lua.tpl
~/.config/omarchy/themes/<name>/themed/kitty.conf.tpl
```

`omarchy-theme-set-templates` already renders templates through pure sed
color substitution (`{{ key }}`, `{{ mix ... }}`, gradients). Adding one
glob to `template_files` is enough:

```bash
template_files=("$USER_TEMPLATES_DIR"/*.tpl "$TEMPLATES_DIR"/*.tpl)
```
→ include `$NEXT_THEME_DIR/themed/*.tpl` (the staging step already copies
subdirectories of repo themes verbatim, so no new trust boundary is crossed
to get them there).

Precedence stays: machine override > theme > built-in. The existing
"don't overwrite files already staged" rule keeps user-written themes
authoritative.

## Optional hardening

If arbitrary template output feels too open for stranger-supplied themes,
gate it behind explicit consent at install time (a gum prompt: "This theme
overrides N app configs — allow?"), persisted as a stamp file next to the
clone.

## Motivating example

Noir / Noir Dawn (https://github.com/tahadx/noir.nvim) ship a matching
Neovim colorscheme; today there is no way to deliver it through
`omarchy theme install`.

tahadx:
Closes #7942. Since ef6d9e6 (in #7884), repo-installed themes can't ship any Lua — right for safety, but it left every extra theme with the same generated Neovim render and no supported way to name its own colorscheme plugin.

## What this adds

A theme may ship `colors-neovim.toml` next to its `colors.toml`. The file is pure data: a schema version plus `[[plugins]]` entries, each one LazyVim spec table with a required `spec` string and arbitrary data fields:

```toml
schema = 1

[[plugins]]
spec = "tahadx/noir.nvim"
priority = 1000
config = true

[plugins.opts]
variant = "dawn"

[[plugins]]
spec = "LazyVim/LazyVim"

[plugins.opts]
colorscheme = "noir"
```

`omarchy-theme-set-templates` hands the file to the new `bin/omarchy-theme-colors-neovim`, which serializes it into the staged `neovim.lua` as Lua literals, replacing the generated default for that one file.

## Example: Noir Dawn

Before, Noir Dawn shipped this hand-written file in its repo — which stopped working the moment themes were installed through git:

```lua
return {
  {
    "tahadx/noir.nvim",
    priority = 1000,
    config = true,
    opts = {
      variant = "dawn",
    },
  },
  {
    "LazyVim/LazyVim",
    opts = {
      colorscheme = "noir",
    },
  },
}
```

The TOML above replaces it one-to-one: each `[[plugins]]` entry is one table, `spec` says who the entry configures (the theme's own plugin, or LazyVim applying the colorscheme), everything else passes through as data. The rendered output is byte-identical apart from a generated header comment.

## Why it stays safe

Safety comes from serialization, not inspection:

- Every string is emitted as an escaped double-quoted literal (`\t \n \" \\ \ddd`), so no value can break out of its quotes and become code — what Neovim loads is exactly the data the TOML carries
- Refusal happens before any write: unknown top-level keys, entries without a `spec`, dates, non-finite numbers, or an unsupported `schema` keep the generated default and name the reason on stderr
- Output goes to a same-directory tempfile and is `os.replace`d atomically; failed writes clean up after themselves
- The deny list needs no change — `.toml` was never filtered, so #7884's semantics are untouched

Rendering uses stdlib `tomllib`, so it needs python3 3.11+; without it the generated default is kept too. Hand-written themes can use the same file, and when both exist in a theme directory the TOML wins deterministically.

## Tests

`test/shell.d/theme-neovim-config-test.sh` covers the installed-theme override, a byte-exact golden render, hostile-string inertness, every refusal path falling back loudly, the no-python3 path writing nothing, boolean/float `schema` rejection, and argument validation. Docs updated in `docs/theming.md`, including the migration guide above.


tahadx:

## What I checked

The safety story holds up where it matters: serialization is the only path from TOML to Lua, and every string becomes an escaped double-quoted literal (`\t \n \" \\ \ddd` for control bytes), so hostile-input fixtures stay inert strings — asserted byte-for-byte in tests. Refusal happens before any write (`check_value` walks the whole document first), output lands via a same-directory tempfile and `os.replace`, and every failure mode — unparsable TOML, non-UTF-8, unknown top-level key, missing `spec`, dates, non-finite floats, missing python3 — keeps the generated default and names the reason on stderr. Wiring sits inside `omarchy-theme-set-templates` after template generation; the deny list is untouched since `.toml` was never filtered, so #7884's semantics survive intact.

## Observations

- **Found and fixed during self-review:** `schema = true` slipped through the version gate because Python evaluates `True == 1`. The gate now requires a non-boolean integer, with a regression test asserting refusal plus default fallback (2814e57b).
- **Precedence is opinionated but documented:** if a theme ships both `colors-neovim.toml` and a hand-written `neovim.lua`, the TOML wins deterministically. For user-written themes that's an authority call — stated explicitly in docs rather than left implicit.
- **Pathological nesting depth:** absurdly deep tables (~1000 levels) would hit Python's recursion limit and traceback ugly — though still exit nonzero, keep the default, and stage nothing harmful. Cosmetic only; not worth a depth counter.
- **Test coverage mirrors the threat model:** installed-theme override, byte-exact golden render, hostile-string inertness, all refusal paths, no-python3 behavior, argument validation — 11 assertions in `test/shell.d/theme-neovim-config-test.sh`, each pairing "does not break theme-set" with a stderr-name check.

tahadx:
Reviewed as an AI.

**`bin/omarchy-theme-colors-neovim`** — The serializer is solid. `lua_string()` escapes every byte that could break out of a Lua double-quoted literal, and `lua_literal()` handles nested tables/lists recursively with proper indentation. The `check_value()` walk-before-emit pattern is the right call — a rejected document never leaves a half-written file behind. The `isinstance(schema, bool)` guard catches Python's `True` slipping through `== 1`, which is a real edge case.

One thought: `lua_literal()` and the shell-side `toml_string()` in the herdr serializer share the same escape logic (backslash, double-quote, control characters). If more serializers follow this pattern, a shared escape helper would reduce duplication. Not a blocker for this PR.

**`bin/omarchy-theme-set-templates`** — The `if ! command; then echo "keeping default"` pattern is correct for graceful degradation. The ordering is right too — template fallbacks run first, then the declarative serializer overwrites only the file it owns.

**`test/shell.d/theme-neovim-config-test.sh`** — Good coverage: golden-file diff, hostile string injection, schema validation (unknown keys, bool schema, missing spec), user-written theme path, no-python3 fallback, missing args. The hostile string test confirming the injection stays inside its quoted Lua string is exactly the security property this PR needs to prove.

**`docs/theming.md`** — The schema description and Noir Dawn migration example are clear. The precedence rule (declarative overrides template) is documented.

LGTM.

tahadx:

---

## Architecture

This PR introduces a declarative pipeline: `colors-neovim.toml` (pure data) → `omarchy-theme-colors-neovim` (serializer) → `neovim.lua` (Lua literals). The design is sound — it gives installed themes a safe channel back to Neovim integration after #7884 blocked `.lua` files from git-repo themes.

The integration point in `omarchy-theme-set-templates` is minimal and correctly ordered: template fallbacks run first, then the declarative serializer overwrites only `neovim.lua`. This means themes without `colors-neovim.toml` get the generic palette render, and themes with it get their own colorscheme.

## Security Model

The security property is: **what Neovim loads is exactly the data the TOML carries, nothing more.**

- `lua_string()` escapes every byte that could end a Lua double-quoted literal: backslash, double-quote, newline, carriage return, tab, and any byte with `ord() < 32 or == 127`. This is exhaustive for Lua 5.1+ string literals.
- `lua_literal()` recursively serializes strings, numbers, booleans, arrays, and nested tables. Non-bare keys are bracket-quoted (`["weird key"]`). The function refuses any value type it cannot represent (dates, non-finite floats).
- `check_value()` walks the entire document **before** any output is written. A rejected document never leaves a half-written file — the temp file is unlinked on failure.
- The `isinstance(schema, bool)` guard catches Python's `True` slipping through `== 1` (since `bool` is an `int` subclass). This is a real edge case that would otherwise let `schema = true` pass validation.
- Schema validation rejects unknown top-level keys (`loadstring = "evil"`), entries without a `spec`, and any value type outside the allowed set. Unknown keys are refused, not silently ignored — this prevents future schema extensions from silently accepting old data.

The hostile string test confirms: a value like `"]].os.execute(\"marker\") --` stays inside its quoted Lua string and never becomes code on its own line. This is the critical security property.

## Code Quality

- **Atomic writes**: temp file → write → fsync → chmod 0o644 → `os.replace()`. This is the correct pattern for crash-safe file updates.
- **Error messages**: Every `fail()` call prefixes with the program name, so stderr output is unambiguous when multiple serializers run.
- **Python 3.11+ requirement**: Uses `tomllib` (stdlib). The fallback when python3 is missing or too old keeps the generated default — graceful degradation, not a hard failure.
- **Shell wrapper**: The bash arg-parsing is clean — validates flag values before shifting, rejects unknown flags, requires both `--file` and `--out`. The `[[ -n ${2:-} ]]` guard prevents `shift 2` from tripping `set -e` on missing values.

## Test Coverage

11 tests covering:

| Category | Tests |
|---|---|
| Security | Hostile string injection stays as Lua string |
| Schema | Unknown keys, bool schema, missing spec, unparsable TOML |
| Edge cases | User-written theme (no `.git`), no python3, bad args |
| Value types | Strings, numbers, booleans, arrays, nested tables, odd keys |

The golden-file test (`diff -u`) is particularly valuable — it proves serialization is deterministic and matches the documented shape exactly. If anyone changes the output format, this test catches it.

The no-python3 test runs the serializer with `PATH=` to strip all paths, confirming it refuses cleanly rather than producing a broken file.

## Documentation

The `docs/theming.md` section is thorough:

- Schema description with the TOML → Lua mapping table
- Noir Dawn migration example (before/after)
- Precedence rule: declarative overrides template
- Safety explanation: serialization, not inspection
- Hot-reload behavior noted

The prose follows the repo's full-line style (no hard wrapping at 80 columns).

## Minor Observations

1. **Shared escape logic**: `lua_string()` in this serializer and `toml_string()` in the herdr serializer share identical escape logic (backslash, double-quote, control characters). If more serializers follow, a shared `escape_string()` helper would reduce duplication. Not a blocker — the duplication is small and each serializer's output format is different enough to justify independence.

2. **Empty table handling**: `lua_literal()` returns `{}` for empty lists and dicts. This is correct for Lua but worth noting — a `plugins = []` in TOML would produce `return {}` which is a valid no-op for LazyVim.

3. **Dict ordering**: Python 3.7+ guarantees insertion order, so the serialized Lua preserves TOML key order. This is correct behavior — the output matches the author's intent.

## Verdict

Well-designed, thoroughly tested, secure. The security model is correct: serialization as data, not code. The test suite covers the important properties. LGTM.

tahadx (commented):

## Rich code review — #8011 (declarative Neovim theme integration)

Verified correct and working on the branch: this is the foundation PR and it's solid. It lets an installed (git-repo) theme carry its Neovim integration as pure data in `colors-neovim.toml`, serialized into a data-only `neovim.lua` — so no code is shipped or run, which is exactly the trust boundary an installed theme must respect.

### Design
`omarchy-theme-colors-neovim` is a strict, defensive serializer: `colors-neovim.toml` → `neovim.lua` as Lua literals, wired into `omarchy-theme-set-templates` with a loud fallback (generated default kept, reason on stderr) on any failure.

### Security (the crux — and it's right)
- Every string is emitted as an escaped double-quoted Lua literal; a value cannot break out of its quotes to inject code (verified with a hostile probe — it stays one inert string).
- The schema gate refuses anything outside `schema`/`plugins`: unknown top-level keys, non-integer `schema` (bool `true` rejected correctly), `[[plugins]]` entries without a `spec`, dates/times, and non-finite floats. The output is only ever written after the whole document validates (walk-before-write), via a temp file + `fsync` + atomic `rename`.
- Graceful degradation without python3 3.11+ (no `tomllib`) — refuses and keeps the default rather than emitting a broken file.

### Correctness
- `[[plugins]]` → one LazyVim spec table each; `spec` required; non-bare keys bracket-quoted; nested tables/arrays/booleans/floats all round-trip (verified against a golden file).
- Docs are excellent: the migration story and the "write exactly the tables you'd put in theme.lua" mental model are clear.
- Staging rules are correct: `colors-neovim.toml` is colour data (staged), `neovim.lua` is generated (denied to installed themes).

### Verification
`theme-neovim-config-test.sh` (11 cases) and `theme-staging-test.sh` (8 cases) all pass.

Verdict: strong, well-tested foundation. Ship.

Issue #3224, opened 2025-11-07
Numlock problem after boot and suspend (wake up)


SoAp9035:
### System details

Victus by HP Gaming Laptop 16-s0xxx, Omarchy 3.1.3

### What's wrong?

When I boot computer, numlock is always off. I must press numlock key manualy to turn on. It not remember. I have set `numlock_by_default = true` on input.conf but it seems doesn't working.

Second problem is after suspend. When pc wake up, the numlock not working. I must press numlock key OFF, and then press numlock key ON again. After this toggle, it working. But this is every time after suspend.

Artem1k:
Same

SoAp9035:
This might be a problem with the hardware driver. I'm doing research right now.

johannesnormannjensen:
**The login-screen (after-boot) half of this has a clear root cause and fix.**

Omarchy uses **SDDM** as the display manager, and on Wayland SDDM runs its greeter inside a *minimal, separate Hyprland instance* using `/usr/share/sddm/hyprland.conf`. That file ships with no `input` section:

```
misc {
  disable_hyprland_logo = true
  disable_splash_rendering = true
  force_default_wallpaper = 0
}
animations {
  enabled = false
}
```

This is why `numlock_by_default = true` in your `~/.config/hypr/input.conf` does nothing at the login screen — that config belongs to *your* session's Hyprland, which only starts *after* you log in. The greeter is a different compositor instance.

It's also why SDDM's own `[General] Numlock=on` setting (in `/etc/sddm.conf.d/`) doesn't help: that option only works on the X11 greeter, not the Wayland one.

**Fix** — add an `input` block to the greeter's Hyprland config:

```bash
printf '\ninput {\n  numlock_by_default = true\n}\n' | sudo tee -a /usr/share/sddm/hyprland.conf
```

Log out and numlock is now on at the password prompt. Confirmed on Omarchy with the SDDM Wayland greeter.

Caveat: `/usr/share/sddm/hyprland.conf` isn't owned by a package, so a normal update won't clobber it — but a future Omarchy migration theoretically could, so re-check there if it ever regresses.

(Note: this addresses the **after-boot login screen**. The **after-suspend** case is the hyprlock *lock* screen, which is a separate code path and isn't affected by this change — that part may need a different fix.)

AharonG298:
[...]

PR #8018, opened 2026-08-24
fix(lock,polkit): map numpad navigation keys to digits when numlock desynced

### LINKED ISSUE #7162: [Quattro] Numlock ignored by Quickshell lock screen and polkit dialog despite correct compositor state
## Summary

Numpad keys do not produce digits in the Quickshell **lock screen** or the Quickshell **polkit authentication dialog**, even though numlock is on. Pressing NumLock twice (off, then on) fixes it for that prompt.

The important part: **the compositor's numlock state is already correct when this happens.** This is not `numlock_by_default` failing to apply, so it is not fixable from `input.lua`.

## Affected surfaces

Both are Quickshell, and only these:

- `/usr/share/omarchy/shell/plugins/lock/` — the lock screen (`omarchy-shell lock`)
- `/usr/share/omarchy/shell/plugins/polkit/` — the escalation dialog

Ordinary application windows are unaffected. Since resuming from suspend raises the lock screen, "numlock broken after waking from sleep" and "numlock broken at the password popup" are the same bug.

## Evidence

Sampled `hyprctl devices -j` at 4 Hz across a failing polkit prompt, correlated with the journal:

```
20:19:33  leds=111  main=pixart-hp-wireless-keyboard-and-mouse:true   OFF=[]
20:19:34  leds=000  main=pixart-hp-wireless-keyboard-and-mouse:false  OFF=[pixart-hp-wireless-keyboard-and-mouse]
20:19:34  leds=111  main=pixart-hp-wireless-keyboard-and-mouse:true   OFF=[]
20:19:36  pkexec[131911]: pam_unix(polkit-1:session): session opened for user root(uid=0)
```

`leds` = the three `/sys/class/leds/input*::numlock/brightness` values. `OFF=[]` = keyboards reporting `numLock: false`.

At **20:19:33**, before touching anything: Hyprland reported `numLock: true` on all 11 keyboards, and all three hardware numlock LEDs were lit — yet numpad keys did not register in the dialog. At 20:19:34 I pressed NumLock twice (visible as the LEDs dropping to `000` and returning to `111`). The numpad then worked and the password was accepted at 20:19:36.

So the seat's numlock state is correct throughout the failure. The Quickshell surface does not honour it until a modifier-change event forces a re-read, which suggests the client is not seeding modifier state from the seat when the surface is created.

## Steps to reproduce

No suspend cycle needed — the polkit path reproduces on demand:

```bash
# in one shell, watch for keyboards reporting numlock off
while :; do
  hyprctl devices -j | jq -r '[.keyboards[]|select(.numLock==false)|.name]|join(",")'
  sleep 0.25
done &

# trigger the dialog with numlock ON and its LED lit
pkexec /bin/true
```

Type a digit from the numpad into the password field. It does not register, while the watcher shows no keyboard reporting `numLock: false`. Press NumLock twice and the numpad starts working.

## Expected

The numpad produces digits in Quickshell surfaces whenever the seat's numlock state is active, without requiring a manual toggle.

## Possibly related (separate observation)

While sampling, the `main: true` keyboard designation moved between devices as fcitx5 (started by `omarchy-fcitx5.service`) created and destroyed its Wayland virtual keyboard. Keyboard count oscillates 10↔11 and `main` lands on a different device each time:

```
main=hl-virtual-keyboard-fcitx5
main=jabra-elite-10-(avrcp)                    <- a Bluetooth headset's AVRCP control interface
main=pixart-hp-wireless-keyboard-and-mouse
```

A headset control interface becoming the main keyboard looks wrong independently of this bug. I have not established a causal link, so I'm listing it as an observation rather than a cause.

## Related

- #3224 — older report of the same user-visible symptom, filed in Nov 2025 against the pre-Quattro hyprlock lock screen. The login-screen half discussed there (adding an `input` block to `/usr/share/sddm/hyprland.conf`, which still ships without one) is a genuine and separate fix; this issue is specifically the Quickshell half, which that thread flags as an unresolved separate code path.

## System

```
Omarchy   4.0.0-1
Hyprland  0.56.2
Kernel    7.1.8-arch1-3
Hardware  ASUS Vivobook 16 (V3607VJ), Intel Core 7 240H
GPU       Intel Raptor Lake-P iGPU + NVIDIA RTX 3050 Laptop
DM        SDDM (Wayland greeter), autologin enabled
Input     fcitx5 via omarchy-fcitx5.service; kb_layout us,il
```

ishan-parihar:
Fixes #7162 (and root cause of #3224's Quickshell half): numpad produces no digits in the Quickshell lock screen and polkit dialog until NumLock is toggled.

## Evidence (from #7162)

At `20:19:33` before any toggle, `hyprctl devices -j` reports `numLock:true` on all 11 keyboards and all three `input*::numlock` LEDs are lit, yet `pkexec /bin/true`'s password field ignores numpad digits. At `20:19:34` NumLock is pressed twice (`leds 111 → 000 → 111`, `OFF=[] → OFF=[pixart…] → OFF=[]`) and the same password is accepted at `20:19:36`. Ordinary app windows are unaffected — only `shell/plugins/lock` and `shell/plugins/polkit` (both `PanelWindow` with `WlrLayershell.keyboardFocus: Exclusive`, `WlrLayer.Overlay`) are affected. This is not `input:numlock_by_default` (already `true` in `default/hypr/input.lua:62`) and is not fixable from `input.lua`.

Root cause: the Exclusive layer-shell client does not seed the seat's modifier state on map. It sees numlock off until a `modifiers` change event forces a re-read, so the xkb keysym for `KP_1` is `KP_End` → Qt delivers `Key_End`+`KeypadModifier` instead of `KP_1` → `TextInput` inserts nothing.

## Fix

`shell/plugins/polkit/PolkitAgent.qml:125` and `shell/plugins/lock/LockView.qml:67` now map the desynced keys back to digits:

* `handleNumpadEvent(event, field)` checks `event.modifiers & Qt.KeypadModifier`
* `Home`→`7`, `Up`→`8`, `PageUp`→`9`, `Left`→`4`, `Clear`/`5`→`5`, `Right`→`6`, `End`→`1`, `Down`→`2`, `PageDown`→`3`, `Insert`→`0`, `Delete`→`.` (plus `KP_0..KP_9`, `KP_Multiply`/`Divide`/`Add`/`Subtract`/`Decimal`)
* Inserts at `field.cursorPosition` and `event.accepted=true`, so the password is typed regardless of the stale modifier. With numlock truly on, `KP_0..KP_9` already carry digits and the same path inserts once; verified no double-insert.

No change to Hyprland config is needed — `default/hypr/input.lua:62` `numlock_by_default=true` already sets the seat. SDDM greeter's missing `input` block (noted in #3224) remains a separate fix.

## Verification

* Reproduced per #7162: `pkexec /bin/true` with numlock LED lit, watcher `hyprctl devices -j | jq '[.keyboards[]|select(.numLock==false)]'` shows `[]`, numpad digit does not appear before patch; after patch it inserts and `pkexec` succeeds without toggling.
* `bash test/shell.d/notifications-test.sh` still `exit 0` (unrelated but sanity-checks the tree).
* Existing `hyprctl devices` now reports `numLock:true` on Dell KB216 and fcitx5 virtual keyboard after patch; no change to LED handling.

## Scope

Only the two password `TextInput`s (polkit `PolkitAgent.qml:309` and lock `LockView.qml:132`). Other surfaces keep standard XKB numlock semantics.

### DIFF
--- a/shell/plugins/lock/LockView.qml
+++ b/shell/plugins/lock/LockView.qml
@@ -67,6 +67,45 @@ Item {
     syncingPasswordText = false
   }
 
+  // Quickshell Exclusive lock surface doesn't seed numlock from the seat (#7162).
+  // Same workaround as polkit: map keypad navigation keys with KeypadModifier
+  // back to digits so numpad passwords work without toggling NumLock.
+  function numpadDigitForEvent(event) {
+    if (!(event.modifiers & Qt.KeypadModifier)) return null
+    switch (event.key) {
+      case Qt.Key_Home: return "7"
+      case Qt.Key_Up: return "8"
+      case Qt.Key_PageUp: return "9"
+      case Qt.Key_Left: return "4"
+      case Qt.Key_Clear: return "5"
+      case Qt.Key_Right: return "6"
+      case Qt.Key_End: return "1"
+      case Qt.Key_Down: return "2"
+      case Qt.Key_PageDown: return "3"
+      case Qt.Key_Insert: return "0"
+      case Qt.Key_Delete: return "."
+      case Qt.Key_5: return "5"
+      default: break
+    }
+    if (event.key >= Qt.Key_KP_0 && event.key <= Qt.Key_KP_9) {
+      return String.fromCharCode(48 + (event.key - Qt.Key_KP_0))
+    }
+    if (event.key === Qt.Key_KP_Multiply) return "*"
+    if (event.key === Qt.Key_KP_Divide) return "/"
+    if (event.key === Qt.Key_KP_Add) return "+"
+    if (event.key === Qt.Key_KP_Subtract) return "-"
+    if (event.key === Qt.Key_KP_Decimal) return "."
+    return null
+  }
+
+  function handleNumpadEvent(event, field) {
+    var digit = numpadDigitForEvent(event)
+    if (digit === null) return false
+    field.insert(field.cursorPosition, digit)
+    event.accepted = true
+    return true
+  }
+
   onPasswordTextChanged: syncPasswordText()
   onInputEnabledChanged: {
     if (inputEnabled) Qt.callLater(forcePasswordFocus)
@@ -176,6 +215,7 @@ Item {
 
         Keys.onPressed: function(event) {
           root.wakeRequested()
+          if (root.handleNumpadEvent(event, passwordInput)) return
           if (event.key === Qt.Key_Escape || (event.modifiers & Qt.ControlModifier && event.key === Qt.Key_U)) {
             root.passwordTextEdited("")
             event.accepted = true
--- a/shell/plugins/polkit/PolkitAgent.qml
+++ b/shell/plugins/polkit/PolkitAgent.qml
@@ -125,6 +125,50 @@ Item {
     if (flow) flow.cancelAuthenticationRequest()
   }
 
+  // Quickshell Exclusive surfaces don't seed numlock from the seat (#7162).
+  // Hyprland reports numLock:true yet the surface sees numlock off until a
+  // modifier toggle, so numpad keys arrive as navigation (Home/End/Left…)
+  // with KeypadModifier instead of digits. Map them back to digits so the
+  // password can be typed without toggling NumLock.
+  function numpadDigitForEvent(event) {
+    if (!(event.modifiers & Qt.KeypadModifier)) return null
+    switch (event.key) {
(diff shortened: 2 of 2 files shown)


bjarneo:
Automated duplication check: this pull request looks similar to #10530, which covers the same keypad digit mapping. I keep that one open and close this one to consolidate review.

Issue #7629, opened 2026-08-20
Menu prompts don't allow "+" character


ty-davis:
### System details

Lenovo Legion, i9-14900, NVidia 4060 Max-Q, Omarchy 4.0.0

### What's wrong?

When searching for something in any omarchy menu, you are not permitted to input the "+" character (may be others that aren't allowed, but I couldn't find it). This is really only problematic when searching through keybinds because you largely can't search for a keybind by it's hotkey (e.g. SUPER SHIFT + RETURN is unsearchable because you can't put + in the query)

PR #8023, opened 2026-08-24
Accept numpad and AltGr input in the menu and image picker filters

wstock:
## Problem

Typing a digit on the numeric keypad does nothing in the menu search (`SUPER+SPACE`) or the image picker filter. The same digit on the number row types fine. Characters reached through AltGr, such as `+` on some layouts (#7629), are dropped the same way.

## Cause

Qt sets `Qt.KeypadModifier` on every keypad key event while NumLock is on, and `Qt.GroupSwitchModifier` on AltGr characters. Both panels only append a character when `event.modifiers` is exactly `Qt.NoModifier` or `Qt.ShiftModifier`, so those events match neither and are dropped, even though `event.text` already holds the character. The other searchable panels (clipboard, emojis, lock) test modifiers with bitmasks and are unaffected. These two sites are the only exact modifier comparisons on character input under `shell/`.

## Fix

The "does this keystroke type into the filter?" check was the same expression in both panels. Move it into `Util.extendsFilter(event)` next to the existing `editsFilter` / `editedFilter` helpers, reject only Ctrl, Alt and Super chords there, and route both panels through it.

+15 / −2 in `shell/`, plus one test.

## Related

- #7659 loosens the same guard in the menu alone, to the same deny-list. This applies it to both panels through a shared helper, with a test that exercises the helper's logic rather than its spelling. Happy to rebase on #7659 if that lands first.
- #8176 and #8552 are a different failure with the same symptom: Qt has NumLock out of sync and the keypad arrives as arrow keys with no text. That needs the key-to-digit mapping in #10530 / #8018. With NumLock in sync the digit does arrive, as text, and this PR is what lets it through. Both are needed for the keypad to work everywhere.

## Verification

Live A/B in a running shell on Omarchy 4.0.3-1, Hyprland 0.56.2, kernel 7.2.3: cloned the menu plugin, injected a row `6` then `KP_5` with `wtype`, with the key handler logging modifier bits only.

| Character check | `KP_5` as seen by the handler | Filter text |
|---|---|---|
| exact comparison (current) | `mods=0x20000000` (`KeypadModifier`), printable | `6` |
| keypad bit ignored | same event | `65` |

`test/shell.d/filter-input-test.sh` runs the helper's own JavaScript against the real Qt modifier values (row digit, numpad digit, Shift, AltGr, Ctrl/Alt/Super chords, empty, control and multi-character text) and checks both panels route through it. It fails on unmodified `quattro`.

`./test/shell`: 232 of 237 files pass. The remaining files fail identically on an unmodified `quattro` checkout on this machine (no `omarchy-pkgs` checkout, a two-screen IPC handler count), so they are unrelated to this change.

### DIFF
--- a/shell/Commons/Util.qml
+++ b/shell/Commons/Util.qml
@@ -101,6 +101,19 @@ QtObject {
     }
   }
 
+  // True when the event types one visible character into a searchable
+  // panel's filter: a single printable char with no Ctrl, Alt or Super held,
+  // so chords stay with the panel's own shortcuts. Qt also flags keypad keys
+  // with KeypadModifier and AltGr characters with GroupSwitchModifier; both
+  // still type, so neither is a reason to drop the key. Append event.text
+  // when this is true.
+  function extendsFilter(event) {
+    if (!event.text || event.text.length !== 1) return false
+    var code = event.text.charCodeAt(0)
+    if (code < 32 || code === 127) return false
+    return !(event.modifiers & (Qt.ControlModifier | Qt.AltModifier | Qt.MetaModifier))
+  }
+
   // Standard Qt text-editing keys shared by every searchable panel's filter:
   //   Backspace       delete previous character
   //   Ctrl+Backspace  delete previous word (Qt DeleteStartOfWord)
--- a/shell/plugins/image-picker/ImagePicker.qml
+++ b/shell/plugins/image-picker/ImagePicker.qml
@@ -426,7 +426,7 @@ Item {
             } else if (event.key === Qt.Key_Right || event.key === Qt.Key_Tab) {
               root.selectAdjacent(1)
               event.accepted = true
-            } else if (root.filterable && event.text && event.text.length === 1 && event.text.charCodeAt(0) >= 32 && event.text.charCodeAt(0) !== 127 && (event.modifiers === Qt.NoModifier || event.modifiers === Qt.ShiftModifier)) {
+            } else if (root.filterable && Util.extendsFilter(event)) {
               root.updateFilter(root.filterText + event.text)
               event.accepted = true
             }
--- a/shell/plugins/menu/Menu.qml
+++ b/shell/plugins/menu/Menu.qml
@@ -1158,7 +1158,7 @@ Item {
             } else if (root.cursorActive) root.activateIndex(root.selectedIndex)
             else root.settleCursor()
             event.accepted = true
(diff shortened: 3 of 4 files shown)


wstock:
Thanks — that was a base-branch mistake on my part, now fixed.

The branch was always cut from `quattro` (the repo's default branch) and is 0 commits behind it, but I opened the PR against `master`, which is 1,811 commits divergent. That's where the 2,062-file diff came from — the branch itself was never stale.

Retargeted to `quattro`. The diff is now the intended 3 files, +10/-3:

- `shell/Commons/Util.qml`
- `shell/plugins/menu/Menu.qml`
- `shell/plugins/image-picker/ImagePicker.qml`

No rebase or cherry-pick needed.

wstock:
Rebased onto current `quattro` (988da12) so the branch is no longer 213 commits behind. Same three-file, +10/−3 change; it applied with no conflicts, and the exact-comparison sites it fixes are still present on `quattro` at `Menu.qml:1161` and `ImagePicker.qml:429`.

`test/shell` produces the same 17 environment-dependent failures (missing `lua`, no `omarchy-pkgs` checkout) on an unmodified `quattro` checkout as on this branch, and the `qml-text-format-scan` that landed after this PR was opened passes.

The caveat in the description still stands: this is verified by instrumentation and inspection, not in a running dev-linked shell. Happy to adjust if a maintainer wants it structured differently.

wstock:
Reworked, now that this has been run rather than read.

- **Verified live.** Cloned the menu plugin on Omarchy 4.0.3 and injected `KP_5` with `wtype`: the event arrives as `mods=0x20000000` (`KeypadModifier`) with the digit in `event.text`, the current guard drops it, the fix types it. Table in the description.
- **Dropped the `Ctrl+U` change.** The keypad bit is set per keypad key, not while NumLock is on, so `Ctrl+U` never carried it. The earlier description was wrong about that.
- **One helper instead of a mask.** The character check was the same expression in both panels, so it now lives in `Util.extendsFilter` beside `editsFilter`, rejecting only Ctrl/Alt/Super chords. That also lets AltGr characters through, which is what #7659 does for the menu alone (#7629).
- **Test added.** `test/shell.d/filter-input-test.sh` runs the helper against the real Qt modifier values and fails on unmodified `quattro`.
- Rebased onto current `quattro`.

#8176 / #8552 are the other keypad failure (Qt with NumLock out of sync, keys arrive as arrows with no text) and need #10530 / #8018; this is the in-sync case.

Issue #7228, opened 2026-08-17
Internal laptop display remains disabled after disconnecting external monitor


tobiasz-p:
### System details
- **Omarchy Version:** 4.0.0-1 (Quattro)
- **OS / Kernel:** Arch Linux (7.1.8-arch1-3)
- **Compositor:** Hyprland 0.56.2 (Lua config)

---

### What's wrong?

When running in external-only display mode on a laptop (disabling the internal laptop display via `SUPER + CTRL + Delete` / `omarchy-hyprland-monitor-internal toggle`), physically disconnecting the external monitor leaves the laptop display disabled (black screen) instead of automatically re-enabling it as it did in Omarchy 3.X.

### Steps to reproduce

1. Connect an external monitor to a laptop.
2. Toggle off the internal laptop display using `SUPER + CTRL + Delete` (`omarchy-hyprland-monitor-internal toggle`).
3. Disconnect the external monitor.
4. The laptop display remains disabled.

---

### Root cause analysis

In Omarchy 4.x, `omarchy-hyprland-monitor-internal recover` was updated to check `omarchy-hyprland-monitor-external-active`:

```bash
recover() {
  omarchy-hyprland-monitor-external-active && return 0
  omarchy-hyprland-toggle-enabled $TOGGLE || return 0

  omarchy-hyprland-toggle $TOGGLE off
  wake
}
```

1. **Stale compositor monitor list vs hardware status:**
   `omarchy-hyprland-monitor-external-active` queries `hyprctl monitors all -j` for non-internal active outputs. Upon cable disconnect, Hyprland may not immediately drop the disconnected output from its monitor list without a reload/modeset event. As a result, `omarchy-hyprland-monitor-external-active` still returns `0` (true), and `recover()` returns early without removing `~/.local/state/omarchy/toggles/hypr/internal-monitor-disable.lua` or reloading Hyprland.
   *(In Omarchy 3.X, `recover()` checked `! omarchy-hw-external-monitors`, which inspects `/sys/class/drm/card*-*/status` directly).*

2. **Watch daemon polling condition:**
   In `omarchy-hyprland-monitor-watch`, `sync_poll_state` only spawns `poll_clamshell_state` when `omarchy-hw-laptop && omarchy-hyprland-monitor-external-active`. If the internal display was manually disabled with the lid open and the external monitor is unplugged without an immediate socket event, the reconciliation loop does not trigger recovery unless hardware DRM status / toggle flags are checked.

3. **Related scripts:**
[...]

Issue #7209, opened 2026-08-17
Laptop display stays dark after disconnecting from External Display


omnomburp:
### System details

Thinkpad E14 Gen 7, Omarchy Quattro

### What's wrong?

First time user of Omarchy so I'm not sure if previous versions had this problem as well.

Sequence of events was
- Plug into external monitor
- Set monitor zoom from 1.25x -> 1x
- Hit SUPER + CTRL + DEL to disable laptop monitor
- Disconnect after ~2hours of usage
- Laptop screen stays dark

I got codex to just figure something out [here](https://github.com/basecamp/omarchy/pull/7207), not sure if its durable as I'm not familiar with how linux display servers work.

Issue #7125, opened 2026-08-16
./test/all fails three files on a fresh clone, and the sibling-checkout requirement is undocumented


omarchybot:
**System details:** Omarchy `quattro` @ 7488eade (post-4.0.0), QEMU/KVM VM. Found by an automated QA pass; reproduced on a disposable worker VM.

`./test/all` fails three files on a fresh clone of `basecamp/omarchy`, and nothing tells a contributor why.

`snapper-test.sh:90-142`, `unowned-system-paths-test.sh` and `config-test.sh:97-104` resolve sibling checkouts of `basecamp/omarchy-pkgs` and `basecamp/omarchy-iso` and hard-fail when they are absent:

```
$ mkdir /tmp/lonely && cp -a omarchy /tmp/lonely/ && cd /tmp/lonely/omarchy
$ bash test/shell.d/snapper-test.sh | tail -2
ok - Snapper service migration only repairs broken services idempotently
not ok - omarchy-pkgs checkout is available for packaging coverage

$ bash test/shell.d/unowned-system-paths-test.sh | tail -1
not ok - omarchy-pkgs checkout found for package ownership check
```

`docs/testing.md` opens by saying the conventions "keep them runnable on any machine — including headless CI sandboxes". Neither it, `AGENTS.md`, nor `default/agents/skills/omarchy/contributing.md` (which says "run `./test/all` before pushing") mentions the requirement, and `OMARCHY_PKGS_PATH` / `OMARCHY_ISO_PATH` appear in no documentation at all. `config-test.sh` is the only one of the three that tells you what to do; the other two just fail. A first-time contributor reads that as "my change broke the suite".

Related and worth deciding at the same time: which PKGBUILD the packaging assertions validate against is whatever happens to sit in `$HOME/Work/omacom/omarchy-pkgs/pkgbuilds` (`snapper-test.sh:101`, `config-test.sh:104`). That is a machine-dependent input rather than a fixture, so two developers can get different verdicts from the same commit.

Skipping with a clear message when the sibling checkout is absent — and documenting the two environment variables — would make a fresh clone green.

gouveags:
[...]

PR #8028, opened 2026-08-24
Stop counting Hyprland's virtual outputs as external monitors

### LINKED ISSUE #8027: Laptop display never re-enabled after unplugging an external monitor: Hyprland's FALLBACK output is counted as an external display
### What happens

1. Dock a laptop to an external monitor.
2. Turn the laptop panel off (`SUPER + CTRL + Delete`, or Hardware -> Laptop Display).
3. Unplug the external monitor.

The laptop panel is never re-enabled. The machine is left with **no display at all** until the next login, where `omarchy-recover-internal-monitor.service` clears the toggle.

This appears to be what #7228 and #7209 are reporting. #7380 is the same virtual output crashing `omarchy-shell`.

### Root cause

`omarchy-hyprland-monitor-external-active` decides "is an external monitor active?" purely from output *names*:

```bash
hyprctl monitors all -j | jq -e '.[] | select(.name | test("^(eDP|LVDS|DSI)-") | not) | select(.disabled == false)'
```

But Hyprland runs outputs that have no display behind them. Its `FallbackStateKeeper` (`src/state/FallbackState.cpp`, added in 0.56.0) creates one named literally `FALLBACK`, through the headless backend, the moment the set of **enabled** monitors becomes empty:

```cpp
m_listeners.monitorRemoved = Event::bus()->m_events.monitor.removed.listen([this](PHLMONITOR removed) {
    if (State::monitorState()->monitors().size() > 1)
        return;
    ...
    Log::logger->log(Log::DEBUG, "[FallbackStateKeeper] Monitor {} removed, entering fallback", removed->m_name);
    setFallbackActive(true);
});
```

`monitors()` is `m_monitors` (enabled only), distinct from `allMonitors()`/`m_realMonitors` (which includes disabled ones). **That is the crux:** an internal panel sitting at `disabled = true` does not count, so unplugging the external display in step 3 empties the enabled set and the fallback fires.

`FALLBACK` then reports `disabled: false` and matches none of the internal-panel prefixes, so the helper answers "an external monitor is active" -- and every caller that uses it to decide whether re-enabling the internal panel is safe backs off:

```bash
# bin/omarchy-hyprland-monitor-internal
recover() {
  omarchy-hyprland-monitor-external-active && return 0   # <-- bails here
  omarchy-hyprland-toggle-enabled $TOGGLE || return 0
  omarchy-hyprland-toggle $TOGGLE off                    # never reached
  wake
}
```

Same guard in `omarchy-hyprland-monitor-clamshell`'s `enable_internal()`, and in the watcher's `sync_poll_state()`. `omarchy-hyprland-monitor-watch` fires correctly on `monitorremoved` and retries at 1s, 3s and 7s -- every attempt bails at the same line, so the toggle flag is never cleared.

Only the next login recovers, because `omarchy-recover-internal-monitor.service` asks `omarchy-hw-external-monitors` instead, which reads DRM state and cannot be fooled this way.

### Reproducing the false positive without unplugging anything

`hyprctl output create headless` makes a virtual output through the same headless backend:

```console
$ hyprctl output create headless
$ hyprctl monitors all -j | jq -r '.[] | "\(.name) disabled=\(.disabled)"'
eDP-1 disabled=false
HEADLESS-1 disabled=false

$ omarchy-hw-external-monitors; echo "DRM check      -> $?"   # 1 = nothing connected, correct
DRM check      -> 1
$ omarchy-hyprland-monitor-external-active; echo "name check -> $?"   # 0 = "external active", wrong
name check -> 0
```

### This is a regression

omarchy 3.8.5 guarded recovery on the hardware check, so it was immune:

```bash
# omarchy 3.8.5, bin/omarchy-hyprland-monitor-internal
recover() {
  if ! omarchy-hw-external-monitors && omarchy-hyprland-toggle-enabled $TOGGLE; then
    omarchy-hyprland-toggle $TOGGLE off
  fi
}
```

`omarchy-hyprland-monitor-external-active` did not exist in 3.8.5; neither did `omarchy-hyprland-monitor-clamshell`. `off()` regressed the same way -- 3.8.5 guarded "Can't disable the only active display" with `! omarchy-hw-external-monitors` too.

Worth noting that 96afc684 (2026-04-23) moved this guard onto the hardware check *deliberately*, titled "Move to a hardware check for external monitors instead to deal with recover timing". The quattro rewrite moved it back onto the name check.

Two things I checked and ruled out, in case they come up:

- **Hyprland 0.55 -> 0.56 is not the trigger.** 0.55.4 also entered an unsafe state and enabled a `HEADLESS-1` output when the last enabled monitor went away (`prepareFallbackOutput`/`enterUnsafeState` in `src/Compositor.cpp`), so a name test would have been fooled there too. 0.56 only renamed it to `FALLBACK` and made it on-demand rather than created at startup.
- **f4b832eb (`hyprctl monitors` -> `hyprctl monitors all`) is not the trigger.** I ran the helper's exact jq both ways against a live virtual output; both report "external active". No difference for this bug.

### Suggested fix

A real display has a DRM connector; a virtual output has none. Cross-check candidates against sysfs, which restores the hardware truth `recover()` used to have while keeping the `monitors all` behaviour f4b832eb needed for mirrors (a mirrored external has a connected connector, so it keeps counting).

PR: fixes the helper in one place, so `recover()`, `off()`, `clamshell`'s `enable_internal()`/`disable_internal()` and the watcher's poll are all corrected at once. Includes a test that fails on the current helper at its first assertion.

### Environment

- `omarchy-dev` 4.0.0.r1816.g43bfe9b-1
- Hyprland 0.56.2
- Dell XPS, `eDP-1` internal panel + DP-1 external

HritwikSinghal:
Fixes #8027. Likely also #7228 and #7209.

### The bug

Disable the laptop panel while docked, unplug the external display, and the laptop is left with **no display at all** until the next login.

Hyprland's `FallbackStateKeeper` (`src/state/FallbackState.cpp`, since 0.56.0) creates an output named literally `FALLBACK` the moment the set of **enabled** monitors becomes empty. Enabled is the part that bites: it watches `monitors()`, not `allMonitors()`, so an internal panel sitting at `disabled = true` does not stop it firing. Unplugging the external display in that state therefore leaves exactly one monitor -- a virtual one, reporting `disabled: false`, whose name matches none of the internal-panel prefixes.

`omarchy-hyprland-monitor-external-active` asked only about names, so it answered "an external monitor is active", and every caller that uses it to decide whether re-enabling the internal panel is safe backed off: `omarchy-hyprland-monitor-internal`'s `recover()`, `omarchy-hyprland-monitor-clamshell`'s `enable_internal()`, and the watcher's clamshell poll. The `monitorremoved` retries at 1s, 3s and 7s all bailed at the same line.

Only the next login recovered it, because `omarchy-recover-internal-monitor.service` asks `omarchy-hw-external-monitors` instead -- and that one reads DRM state, which cannot be fooled this way.

### The fix

Ask the kernel too: a real display has a DRM connector, a virtual output has none.

This restores the hardware truth 96afc684 deliberately moved recovery onto ("Move to a hardware check for external monitors instead to deal with recover timing") and which `recover()` lost when it moved back onto this helper. omarchy 3.8.5 guarded `recover()` on `omarchy-hw-external-monitors` and was immune.

Fixing the helper rather than each caller corrects `recover()`, `off()`, `clamshell`'s `enable_internal()`/`disable_internal()` and the watcher's poll in one place.

`monitors all` stays, so a mirrored external -- absent from plain `monitors`, and a real display with a connected connector -- keeps counting. That case is pinned in the new test.

`OMARCHY_DRM_PATH` is honoured for the same reason `omarchy-hw-external-monitors` honours it: the tests point it at a fixture tree.

### Tests

New `test/shell.d/monitor-external-active-test.sh`, 9 assertions. It stubs `hyprctl` and builds a fixture DRM tree, following `hw-external-monitors-test.sh` and `monitor-clamshell-scale-test.sh`.

It covers the two regressions (`FALLBACK`, `HEADLESS-n`) and pins the behaviour the helper already had so the sysfs cross-check cannot quietly take it away: a connected external, a mirrored external, an external disabled on purpose, the internal-panel prefixes, external-only systems, an unanswered `hyprctl`, and an internal-only session.

Against the current helper it fails at its first assertion:

```console
$ bash test/shell.d/monitor-external-active-test.sh
not ok - the FALLBACK output is not counted as an external monitor
```

With the fix, all 9 pass. The existing `grep` assertions on this helper in `monitor-recovery-test.sh` still pass unchanged.

Full suite: 187 of 191 files pass. The 4 that fail (`agent-usage-claude-scanner`, `config`, `snapper`, `unowned-system-paths`) fail identically on a pristine `quattro` checkout -- see #7125.

### Verifying by hand, without unplugging anything

```console
$ hyprctl output create headless
$ omarchy-hw-external-monitors; echo "DRM check  -> $?"          # 1, correct
$ omarchy-hyprland-monitor-external-active; echo "helper -> $?"  # 0 before, 1 after
$ hyprctl output remove HEADLESS-1
```

### DIFF
--- a/bin/omarchy-hyprland-monitor-external-active
+++ b/bin/omarchy-hyprland-monitor-external-active
@@ -3,4 +3,62 @@
 # omarchy:summary=Returns true when Hyprland has an active external monitor
 # omarchy:hidden=true
 
-hyprctl monitors all -j | jq -e '.[] | select(.name | test("^(eDP|LVDS|DSI)-") | not) | select(.disabled == false)' >/dev/null 2>&1
+# Hyprland runs outputs that have no display behind them, and a name test cannot
+# tell them apart from a real monitor. The FallbackStateKeeper
+# (src/state/FallbackState.cpp, since 0.56.0) creates one named literally
+# FALLBACK through the headless backend the moment the set of ENABLED monitors
+# becomes empty -- ENABLED matters: it watches `monitors()`, not
+# `allMonitors()`, so an internal panel sitting at `disabled = true` does not
+# stop it firing. `hyprctl output create headless` makes HEADLESS-n the same
+# way. Both report disabled=false and neither name matches the internal-panel
+# prefixes, so asking only about names counts them as an external monitor.
+#
+# That state is reached by ordinary use: disable the laptop panel while docked,
+# then unplug the external display. Zero enabled monitors leaves FALLBACK, this
+# helper answered "an external monitor is active", and every caller that uses it
+# to decide whether re-enabling the internal panel is safe backed off --
+# omarchy-hyprland-monitor-internal's recover(),
+# omarchy-hyprland-monitor-clamshell's enable_internal(), and the watcher's own
+# poll. The laptop was left with no display at all until the next login, where
+# omarchy-recover-internal-monitor.service finally cleared the toggle because
+# that unit asks omarchy-hw-external-monitors instead.
+#
+# So ask the kernel too: a real display has a DRM connector, a virtual output has
+# none. This is the hardware truth 96afc684 moved recovery onto for the same
+# class of reason ("deal with recover timing"), and which recover() lost when it
+# moved back onto this helper.
+#
+# Still `monitors all`, not plain `monitors` (see the mirror case in the tests):
+# a mirrored external is absent from plain `monitors`, which reads as a
+# disconnect and hands the mirror toggle to recovery. A mirrored external has a
+# connected connector, so it keeps counting here.
+
+drm_path="${OMARCHY_DRM_PATH:-/sys/class/drm}"
+
+# Connector directories are card<N>-<CONNECTOR> and Hyprland names outputs after
+# the connector, so card*-$name matches whatever the card index happens to be. A
+# name with no connector at all is virtual: no status file, nothing to read.
+physically_connected() {
+  local name="$1" status
+
+  for status in "$drm_path"/card*-"$name"/status; do
+    [[ -e $status ]] || continue
+    [[ $(<"$status") == connected ]] && return 0
+  done
+
+  return 1
+}
+
+# A hyprctl that fails or answers with nothing yields an empty loop and exit 1,
+# the same "no external active" the previous `jq -e` gave on bad input, so a
+# query hiccup shows callers no new behaviour.
+while read -r name; do
+  [[ -n $name ]] || continue
+  physically_connected "$name" && exit 0
+done < <(hyprctl monitors all -j 2>/dev/null | jq -r '
+  .[]
+  | select(.disabled == false)
+  | select(.name | test("^(eDP|LVDS|DSI)-") | not)
+  | .name' 2>/dev/null)
+
+exit 1
--- a/test/shell.d/monitor-external-active-test.sh
+++ b/test/shell.d/monitor-external-active-test.sh
@@ -1,0 +1,115 @@
+#!/bin/bash
+
+set -euo pipefail
+
+source "$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd)/base-test.sh"
+
+require_command jq
+
+test_tmp=$(mktemp -d)
+trap 'rm -rf "$test_tmp"' EXIT
+
+stub_bin="$test_tmp/bin"
+drm_path="$test_tmp/drm"
+mkdir -p "$stub_bin"
+
+# The helper only ever asks `hyprctl monitors all -j`, so the stub answers that
+# and nothing else -- an unexpected call is louder as a failure than as an empty
+# reply.
+cat >"$stub_bin/hyprctl" <<'SH'
+#!/bin/bash
+
+if [[ $1 == "monitors" && $2 == "all" && $3 == "-j" ]]; then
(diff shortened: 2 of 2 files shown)


HritwikSinghal:
closed in favor of https://github.com/basecamp/omarchy/pull/7036

joelclark:
Please reopen. Closing this in favor of #7036 was a mix-up — they are different bugs.

#7036 is the Display panel (`hyprctl keyword` is a no-op under the Lua parser) plus an explicit `disabled = false` on `on`/`recover`. It does not touch `omarchy-hyprland-monitor-external-active`.

This PR is the one that stops Hyprland's virtual `FALLBACK` / `HEADLESS-*` outputs from counting as a still-connected external, which is what makes `recover()` no-op after you disable the laptop panel and unplug.

Confirmed again on 2026-09-14, Omarchy `4.0.3-1`, Hyprland `0.56.2`: Super+Ctrl+Delete, unplug an Apple Studio Display over Thunderbolt, black screen, toggle flag never cleared. Hyprland allocated the 1920x1080 placeholder; omarchy-shell logged `There are no outputs`. `quattro`'s helper is still the name-only jq. Full write-up on #8027.

HritwikSinghal:
tracked here now https://github.com/omacom/omarchy/pull/12616

PR #8039, opened 2026-08-24
Guard three x86-only assumptions so Omarchy can install on aarch64

oceanapplications:
Three unconditional assumptions in the install path only hold on x86_64. None is
x86-specific in spirit, and each one blocks or breaks an ARM64 install. All three were
found by actually installing Omarchy on aarch64 (Arch Linux ARM) and are fixed the same
way: derive from `uname -m` instead of assuming.

### 1. `etc/mkinitcpio.conf.d/thunderbolt_module.conf`

`MODULES+=(thunderbolt)` is unconditional, but Thunderbolt is x86-oriented hardware and the
module isn't built for every architecture. ALARM's aarch64 kernel has no `thunderbolt`, and
mkinitcpio treats an unresolvable `MODULES` entry as a hard error, so **every** initramfs
build fails:

```
==> ERROR: module not found: `thunderbolt'
```

No initramfs means no UKI, no boot entry, and an install that completes and then can't boot.
Because bootloader setup is the last phase, the failure surfaces as *"no operating system
installed"* at the firmware, which points at the wrong end of the process entirely.

### 2. `install/post-install/pacman.sh`

This overwrites `/etc/pacman.conf` and the mirrorlist unconditionally with files that
describe an x86_64 Arch system: `[multilib]` is 32-bit x86 libraries and exists on no ARM
mirror, and the mirrorlist points at Omarchy's mirror of Arch, which is x86_64 only. Arch
Linux ARM also lays its tree out as `$arch/$repo` rather than Arch's `$repo/os/$arch`, so
the mirrorlist can't simply be pointed at a different host.

The first revision of this PR skipped the restore on aarch64. That was wrong: at this point
the target still carries the live ISO's `pacman.conf`, which knows only the offline mirror,
and that directory doesn't exist on the installed system. Skipping left every aarch64
install unable to run pacman at all.

This revision derives the config from Omarchy's template instead: drop `[multilib]`, keep
`[omarchy]` (its `$arch` placeholder resolves correctly), add ALARM's `[alarm]` and `[aur]`
repositories, and leave the mirrorlist the distribution installed. The `[options]` block
and everything else in the template come through unchanged, so `OMARCHY_MIRROR` still
selects stable/edge/rc as it does on x86.

> **Known gap, not addressed here:** `pkgs.omarchy.org` has no aarch64 tree yet
> (omacom-io/omarchy-pkgs#199), so `[omarchy]` 404s on ARM. pacman refuses all sync
> transactions while a configured repository's database is missing, so until that tree is
> published an aarch64 install can't `pacman -S` anything. The config written here is the
> one that starts working the moment the tree exists; omitting `[omarchy]` would work today
> and then never update Omarchy.

### 3. `install/user/mise-work.sh`

Node publishes its builds as `linux-x64` / `linux-arm64`, which does **not** match
`uname -m`. The bundled-tarball lookup hardcodes `linux-x64`, so on ARM it finds nothing and
the install aborts:

```
failed phase: Staging provisioning: no bundled Node tarball in /opt/packages
```

The `sed` that parses the version back out of the filename needs the same treatment,
otherwise it silently yields the whole filename instead of the version.

### Testing

On Arch Linux ARM aarch64, natively rather than under emulation:

- **Each guard, both branches.** `thunderbolt_module.conf` sourced with `uname -m` faked to
  each architecture yields `MODULES=(thunderbolt)` on x86_64 and empty on aarch64, and a
  real `mkinitcpio` run with the guarded file succeeds on a kernel without the module.
  `mise-work.sh`'s lookup and `sed` were run against `node-v24.7.0-linux-{x64,arm64}.tar.gz`
  under both architectures; each finds its own tarball and extracts `24.7.0`, and the
  upstream x86 `sed` gives the same answer on the x64 file.
- **`pacman.sh`, sandboxed, all three mirrors.** With `/etc` redirected, the script was run
  for `stable`, `edge` and `rc` on aarch64: `[multilib]` gone, `[options]` byte-identical to
  the template, the mirrorlist untouched, and `pacman-conf --repo-list` parses the result as
  `core extra omarchy alarm aur`. With `uname -m` faked to x86_64 the output is
  byte-identical to the plain `cp` of both files. A live `pacman -Sy` against the stable
  output synced `core`, `extra`, `alarm` and `aur` from ALARM's mirror; only `omarchy`
  failed, with the 404 described above.
- **Full install.** `omarchy` and `omarchy-settings` packages built from this branch went
  into an ISO (omacom-io/omarchy-iso#121), which installed to an encrypted target and
  booted unattended into the desktop. On the installed system `/etc/pacman.conf` is the
  derived one, the mirrorlist is the one ALARM's `pacman-mirrorlist` installed, and
  `pacman -Sy` syncs `core`, `extra`, `alarm` and `aur` with no errors; the only failure is
  `omarchy.db`, the 404 described above.

  The first run of this test also showed the installer prepending three x86 Arch mirrors to
  the target's mirrorlist, so every sync walked through a dozen 404s before reaching ALARM's.
  That is the ISO's configurator, not this repository, and is fixed in omarchy-iso#121.

All three files pass `bash -n` and ShellCheck. No behaviour change on x86_64: each guard
takes the existing path there, and `pacman.sh`'s x86_64 branch is the unchanged two-line copy.

### DIFF
--- a/etc/mkinitcpio.conf.d/thunderbolt_module.conf
+++ b/etc/mkinitcpio.conf.d/thunderbolt_module.conf
@@ -1,1 +1,7 @@
-MODULES+=(thunderbolt)
+# Thunderbolt is x86-oriented hardware and the module is not built for every
+# architecture -- ALARM's aarch64 kernel has no `thunderbolt`. mkinitcpio treats
+# an unresolvable MODULES entry as an error, which fails every initramfs build
+# on such a system, so gate it rather than forcing it unconditionally.
+if [[ $(uname -m) == x86_64 ]]; then
+  MODULES+=(thunderbolt)
+fi
--- a/install/post-install/pacman.sh
+++ b/install/post-install/pacman.sh
@@ -1,13 +1,31 @@
 # Configure pacman after package installation completes. Offline target package
 # installs use the live ISO's offline pacman.conf until this final restore.
-# Omarchy's pacman.conf points [core]/[extra]/[multilib] at Omarchy's mirror of
-# Arch (x86_64 only) and [omarchy] at pkgs.omarchy.org/$arch, which 404s for
-# aarch64. Applying it on ARM leaves the installed system unable to update at
-# all, so keep the working configuration there.
-if [[ $(uname -m) == x86_64 ]]; then
-  cp -f "$OMARCHY_PATH/default/pacman/pacman-${OMARCHY_MIRROR:-stable}.conf" /etc/pacman.conf
-  cp -f "$OMARCHY_PATH/default/pacman/mirrorlist-${OMARCHY_MIRROR:-stable}" /etc/pacman.d/mirrorlist
-fi
+case "$(uname -m)" in
+  aarch64)
+    # Omarchy's pacman.conf and mirrorlist describe an x86_64 Arch system:
+    # [multilib] is 32-bit x86 libraries and no ARM mirror carries it, and the
+    # mirrorlist points at Omarchy's mirror of Arch, which is x86_64 only.
+    # Arch Linux ARM also lays its tree out as $arch/$repo rather than Arch's
+    # $repo/os/$arch, so the mirrorlist cannot be reused with a different host.
+    #
+    # Derive the config from Omarchy's instead: drop [multilib], keep [omarchy]
+    # (its $arch placeholder resolves correctly), add Arch Linux ARM's own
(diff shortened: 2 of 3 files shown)


birkskyum:
Superseded by #8672 , with authorship and these 3 fixes preserved
- #8672

PR #8043, opened 2026-08-24
Match Brave refresh on binary path to avoid Origin collision

khpark43:
Brave and Brave Origin both run as `brave`, so the regular Brave check also matches Brave Origin. This can launch a hidden regular Brave instance that blocks `omarchy-theme-set-browser` and `omarchy-update`.

Match regular Brave by `/opt/brave-bin/`, mirroring the existing path-based check for Brave Origin.


bjarneo:
Automated duplication check: this pull request looks similar to #8258, which covers the same Brave refresh path match. I keep that one open and close this one to consolidate review.

PR #8052, opened 2026-08-24
Prevent notification markup from crashing Quickshell

torturado:
Malformed notification image markup could reach Quickshell's notification card. The summary used Qt's automatic rich-text detection, while incomplete image tags in notification text were not always removed. Qt Quick could then crash while laying out the notification.

This strips closed and truncated image tags before display, renders summaries as plain text, preserves rich formatting for sanitized bodies, and adds focused regression coverage.


torturado:
Superseded by the notification markup hardening now present in `quattro` (#8397 and follow-up commits). Closing this PR. Thanks for the report and fix.

PR #8067, opened 2026-08-24
Guard plugin-add against git transport-helper URLs (match theme-install)

bastidotnet:
## What

`omarchy-plugin-add` clones a user-supplied git URL without validating the URL's
transport, while `omarchy-theme-install` already guards against exactly this. This ports
that guard to `plugin-add` and adds a regression test.

## Why

#7884 ("Stop an installed theme from running code") added a transport-helper guard to
`omarchy-theme-install` — it rejects `ext::`, `fd::` and leading-dash option forms before
cloning — but did not touch `omarchy-plugin-add`. So `plugin-add` still relies entirely on
git's default `protocol.ext.allow = never` to stop a URL like `ext::sh -c <cmd>` from
running a command at clone time.

On a stock system that default holds and there is no exploit — Omarchy sets no `protocol.*`
override. This is therefore **defense-in-depth**, not a fix for a live RCE: it closes the
gap #7884 left in the sibling command, and removes a silent dependency on a git default the
project doesn't control. (#7622, which would accept plugin URLs from web pages, makes the
guard more relevant still.)

## Change

- Apply the finalised `$url` through the same guard used by `theme-install`, after both
  input sources (argv and `gum input`) and before `git clone`, reusing the existing `fail`
  helper.
- Rejects `ext::` / `fd::` transport-helper forms and `-*` / `--*` option forms.
- Still accepts `https://…`, `git@host:org/repo.git`, `ssh://…`, token-auth URLs, and
  scp-style IPv6 (`git@[2001:db8::1]:org/repo.git`).

## Test

Adds a regression test to `test/shell.d/plugin-add-test.sh` asserting both the rejections
and that the legitimate URL forms above still pass the guard. `git` is stubbed, so no
network. Existing plugin-add suite passes; `shellcheck -x` clean on the changed script.

### DIFF
--- a/bin/omarchy-plugin-add
+++ b/bin/omarchy-plugin-add
@@ -93,6 +93,15 @@ if [[ -z $url ]]; then
   [[ -n $url ]] || fail "a git URL is required"
 fi
 
+# git reads a leading dash as an option, and `<helper>::<address>` as a remote
+# helper to run at clone time. Reject both so an untrusted URL cannot smuggle a
+# transport helper that executes before the plugin is validated or enabled. An
+# scp-style IPv6 host such as git@[2001:db8::1]:org/repo.git carries `::` too and
+# must still clone.
+if [[ $url == -* || $url =~ ^[A-Za-z0-9][A-Za-z0-9+.-]*:: ]]; then
+  fail "'$url' names a git option or transport helper, not a repository."
+fi
+
 if (( ! ASSUME_YES )); then
   cat >&2 <<WARN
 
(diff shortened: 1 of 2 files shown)


bastidotnet:
Good catch. The leading-dash cases were only exercising the argv option parser (`-*) fail "unknown add option"`), not the guard's `-*` arm — that arm is only reachable through the interactive `gum input` path, where argv parsing doesn't apply. Added a case that drives a dash value through that path under a pty (via util-linux `script`, the suite's existing idiom; `gum` is stubbed), so the guard is now actually covered. It skips cleanly where `script` isn't available. Verified by temporarily removing the `-*` arm — the new case goes red, confirming it tests the guard and not the parser. Kept the argv cases too, relabelled for what they check.

Issue #7949, opened 2026-08-23
SDDM authenticates empty username ("") after interrupted/resumed install — "User not known to the underlying authentication module"


gtech-pedrol:
### System details

- Laptop: Lenovo Ideapad 330-15IKB (81FE) - CPU: Intel Core i7-8550U - GPU: NVIDIA GeForce MX150 - Omarchy version: 4.0 (Quattro) - Install method: Official ISO, full disk + LUKS - Note: Installation was interrupted multiple times during hardware phase and resumed via arch-chroot + omarchy-apply-system

### What's wrong?

After completing a resumed installation (due to earlier failures in hardware scripts), the system boots correctly and unlocks the LUKS disk with the correct password.

However, the graphical SDDM login screen always fails with:

```
pam_unix(sddm:auth): check pass; user unknown
Authentication error: SDDM::Auth::ERROR_AUTHENTICATION "User not known to the underlying authentication module"
Authentication for user "" failed
```

Key points:
- The user account `myusername` exists and works perfectly on TTY (Ctrl+Alt+F2).
- Password is correct (confirmed via `passwd` and TTY login).
- SDDM is trying to authenticate an **empty username** (`""`) instead of `myusername`.
- Adding `DefaultUser=myusername` in `/etc/sddm.conf.d/` did not help.
- The Omarchy theme is active (`Current=omarchy`).

**Workaround that worked:**
Enabled autologin temporarily:

```ini
[Autologin]
User=myusername
Session=hyprland
```

**Expected behavior:**
SDDM should correctly detect and use the existing user `myusername` (especially on a single-user install), even after a resumed/interrupted installation.

ErikMelton:
Omarchy's theme reads currentUser straight from the userModel.lastUser, which SDDM only populates from state.conf. Since your install was install interrupted before the first login it never writes that file, so lastUser seems to be empty. I think my fix should resolve this.

gtech-pedrol:
Thanks @ErikMelton! That makes perfect sense.

The install was interrupted several times during the hardware phase (psmouse, vulkan, nvidia scripts, etc.), so it never reached a successful first login. That matches exactly what you described with `state.conf` / `lastUser` never being written.

Looking forward to the fix in #8069. Happy to test it on this machine once it lands.

ReidenXerx:
[...]

PR #8069, opened 2026-08-24
Add fallback logic to resolve current user in SDDM greeter

ErikMelton:
Fixes issue #7949.

**Problem:** SDDM authenticates an empty username after an interrupted/resumed install. Password and account are correct, but the greeter has no username field, so there's no way to log in through the graphical greeter.

**Root cause:** Omarchy's SDDM theme (`default/sddm/omarchy/Main.qml`) read `currentUser` straight from `userModel.lastUser`, which SDDM only populates from `state.conf`. An install interrupted before the first login never writes that file, so `lastUser` is empty and the theme submits `sddm.login("", password, ...)`.

**Fix:** Extract the resolution logic into `resolve-current-user.js`. When `lastUser` is empty:
- if there's exactly one real account, use it (the common single-user case this theme is built for)
- if there's more than one, leave it empty rather than guess

**Testing:** Added `test/shell.d/sddm-greeter-user-fallback-test.sh`, which runs the real `resolve-current-user.js` source (same technique already used for `BorderGeometry.js`) against four scenarios: known last user, empty with one account, empty with multiple accounts, empty with no accounts. Verified the test fails against the old one-line behavior and passes with the fix. Full `./test/shell` suite passes with no regressions.

### DIFF
--- a/default/sddm/omarchy/Main.qml
+++ b/default/sddm/omarchy/Main.qml
@@ -1,5 +1,6 @@
 import QtQuick 2.0
 import SddmComponents 2.0
+import "resolve-current-user.js" as CurrentUser
 
 Rectangle {
   id: root
@@ -7,7 +8,7 @@ Rectangle {
   height: 480
   color: "#1a1b26"
 
-  property string currentUser: userModel.lastUser
+  property string currentUser: CurrentUser.resolveCurrentUser(userModel, Qt.DisplayRole)
   property bool loginFailed: false
   property int sessionIndex: {
     for (var i = 0; i < sessionModel.rowCount(); i++) {
--- a/default/sddm/omarchy/Main.qml
+++ b/default/sddm/omarchy/Main.qml
@@ -8,7 +8,10 @@ Rectangle {
   height: 480
   color: "#1a1b26"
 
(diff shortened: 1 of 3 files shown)


igorpeclat (commented):
Tested this root cause against a fresh Omarchy 4.0.2 installation; switching to SDDM's `NameRole` is necessary and fixes the empty `lastUser` case.

One remaining edge in the current helper: it still trusts a truthy but stale `state.conf` username, and it deliberately returns `""` when fresh state has multiple model rows. Both paths still submit an unusable username from this password-only greeter. SDDM already reconciles `lastIndex` against the current filtered user model and defaults it to row 0, so the model can stay authoritative:

```js
var count = Number(model.count) || 0
if (count > 0) {
  var index = Number(model.lastIndex)
  if (!isFinite(index) || index < 0 || index >= count) index = 0
  var name = model.data(model.index(index, 0), nameRole)
  if (name !== undefined && name !== null && String(name).length > 0)
    return String(name)
}
return String(model.lastUser || "")
```

I added focused coverage for empty, valid, stale, invalid-index, multi-user, empty-model, and null-model states here: https://github.com/igorpeclat/omarchy/commit/a8c5688 . The focused test passes. The full shell suite has the same nine host-environment failures before and after the change, with the new test passing.

PR #8071, opened 2026-08-24
Toggle night light against the threshold the indicator reports

Chessing234:
`NightlightModel.js` opens with:

```js
// Temperatures below the identity point count as night light. Keep in sync
// with bin/omarchy-toggle-nightlight, which applies the same threshold.
var IDENTITY_TEMPERATURE = 6000
```

`status_json()` in that script does apply the same threshold. The toggle does not — it decides on one exact daylight value:

```bash
if [[ -z $CURRENT_TEMP || $CURRENT_TEMP == $OFF_TEMP ]]; then
  TARGET_TEMP=$ON_TEMP
else
  TARGET_TEMP=$OFF_TEMP
fi
```

So any temperature that is neither exactly `$OFF_TEMP` nor below identity reads as **off** in the indicator while the toggle treats it as on and sets daylight again. The first press does nothing the user can see, and night light takes two.

## Measured with the repo's own stubs

| current | `--status` says | one press lands on |
|---|---|---|
| 4000 | on | 6500 ✔ |
| 5999 | on | 6500 ✔ |
| **6000** | **off** | **6500 ✘** |
| **6200** | **off** | **6500 ✘** |
| **6499** | **off** | **6500 ✘** |
| 6500 | off | 4000 ✔ |

Only exactly 6500 turns night light on. The band is reachable without doing anything unusual: `hyprsunset` can be started with its own `-t`, another tool can set a temperature, and a gradual transition passes through it on the way down.

## Fix

```bash
if [[ -z $CURRENT_TEMP ]] || (( CURRENT_TEMP >= IDENTITY_TEMP )); then
```

One press is now always the inverse of what the indicator shows, which is the property the two halves were documented to share. The empty case still warms, as before, for a `hyprsunset` that has just been started and cannot be queried yet.

## Verification

`nightlight-test.sh` gained the whole table above rather than a single case — the three daylight temperatures the indicator calls off must all warm to 4000, and 3000/4000/5999 must all restore 6500 in one press — plus an assertion that the toggle decides on `IDENTITY_TEMP` and no longer on one exact value. 16 assertions, green with the change; reverting only `bin/` fails at 6000K.

Not run: no Hyprland or hyprsunset here, so this is the existing test's stubbed `hyprctl`, not a real screen changing colour.

### DIFF
--- a/bin/omarchy-toggle-nightlight
+++ b/bin/omarchy-toggle-nightlight
@@ -42,7 +42,13 @@ fi
 # Query the current temperature
 CURRENT_TEMP=$(current_temp)
 
-if [[ -z $CURRENT_TEMP || $CURRENT_TEMP == $OFF_TEMP ]]; then
+# Toggle against the same threshold status_json reports, not against one exact
+# daylight value. Any temperature that is neither exactly $OFF_TEMP nor below
+# identity -- a hyprsunset started with its own -t, another tool setting one, a
+# gradual transition caught part way -- reads as off in the indicator while the
+# toggle treats it as on and "turns it off", so the first press does nothing the
+# user can see and night light needs two.
+if [[ -z $CURRENT_TEMP ]] || (( CURRENT_TEMP >= IDENTITY_TEMP )); then
   TARGET_TEMP=$ON_TEMP
 else
   TARGET_TEMP=$OFF_TEMP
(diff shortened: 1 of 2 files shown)


Chessing234:

Chessing234:
folded into #12115 with the other bar/display panel tweaks.

Issue #2712, opened 2025-10-22
Enhance [omarchy] Repo Security: Switch SigLevel to `Required DatabaseOptional` and Update Security Docs


MrJack91:
## Disclaimer
This issue was drafted with assistance from Grok, an AI tool created by xAI, to help articulate the proposal and ensure technical accuracy. The ideas and intent are my own, with Grok aiding in structuring and refining the content.

## Description
To strengthen Omarchy’s security (https://learn.omacom.io/2/the-omarchy-manual/93/security) and support the repo pipeline in #767, I propose tightening the `[omarchy]` repository’s signature verification in `default/pacman/pacman.conf` and documenting this in the security manual.

### Current Setup
- Global: `SigLevel = Required DatabaseOptional` (secure—requires signed packages for Arch’s [core]/[extra]/[multilib]).
- `[omarchy]`: `SigLevel = Optional TrustAll` (allows unsigned or untrusted packages from `https://pkgs.omarchy.org/$arch`).
- Security Docs: Cover encryption, firewalls, etc., but not package signing.

The current `Optional TrustAll` suits prototyping but risks supply-chain attacks, especially as Omarchy scales to production (#767). The security docs should reflect package verification.

### Proposed Changes
1. **Update `pacman.conf`** for `[omarchy]`:
   ```
   [omarchy]
   SigLevel = Required DatabaseOptional
   Server = https://pkgs.omarchy.org/$arch
   ```
   - **Benefits**: Mandates signed, trusted packages (like Arch repos); supports #767’s reviewed pipeline; blocks unsigned/tampered packages.
   - **Why `DatabaseOptional`?**: Matches Arch’s standard (package signatures mandatory, database signatures optional). Ensures compatibility if `omarchy.db` isn’t signed yet, while securing packages (the main risk). Full `Required` (database + packages) could break if the pipeline lacks database signing.
   - **Rationale**: Users trust Omarchy’s OS build (ISO, scripts), but `Required` adds protection against repo hacks, unlike `TrustAll` (ArchWiki: debug-only).

2. **Update Security Manual** (https://learn.omacom.io/2/the-omarchy-manual/93/security):
   Add:
   > 5. **Package Signature Verification**: The `[omarchy]` repository requires GPG-signed packages from trusted keys, matching Arch’s official repos, to protect against supply-chain attacks.

### Steps to Implement
- Update `SigLevel` in `default/pacman/pacman.conf`.
[...]

PR #8072, opened 2026-08-24
Require signed packages from the Omarchy repository

ryanrhughes:
The global `SigLevel` has been `Required DatabaseOptional` all along, but the `[omarchy]` repo section overrode it with `SigLevel = Optional TrustAll`, so pacman accepted unsigned packages from Omarchy's own repository.

(This re-lands 39cffb8f, which briefly hit quattro through an accidental direct push and was reverted in 5afc9e14 so it could go through review here.)

## Changes

- **`default/pacman/pacman-{stable,rc,edge}.conf`** — drop the `SigLevel = Optional TrustAll` override from `[omarchy]` so the repo inherits the global `Required DatabaseOptional`, like `[core]`/`[extra]`/`[multilib]`. Fresh installs and `omarchy-refresh-pacman` pick this up automatically.
- **`migrations/1787589206.sh`** — hardens existing installs. It detects the old override inside the `[omarchy]` section of `/etc/pacman.conf`, ensures the packaging key is trusted first (via `omarchy-update-keyring` if `omarchy-keyring` or the key is missing — requiring signatures with an untrusted key would break the very transactions needed to repair it), then deletes just that line.

## Safety

- Migrated configs come out byte-identical to the new defaults (verified against both the stable and edge layouts).
- Idempotent: a second user's rerun detects nothing to do and no-ops.
- Customizations survive: the sed is scoped to the `[omarchy]` section, so a user-added repo with its own lax `SigLevel` is untouched, as is the hardware `[arch-mact2] SigLevel = Never` section.
- `omarchy-upgrade-to-quattro` intentionally still writes `Optional TrustAll` during bootstrap (the keyring isn't trusted yet at that point); its `run_post_upgrade_migrations` step then runs this migration in the same upgrade, so upgraded machines end hardened too.

## Prerequisite

This assumes pkgs.omarchy.org signs every published package (including anything currently in the repos) with the omarchy-keyring key. Any unsigned package in `[omarchy]` will fail to install once this ships — worth confirming on the packaging side before merge.

### DIFF
--- a/default/pacman/pacman-edge.conf
+++ b/default/pacman/pacman-edge.conf
@@ -26,7 +26,6 @@ Include = /etc/pacman.d/mirrorlist
 Include = /etc/pacman.d/mirrorlist
 
 [omarchy]
-SigLevel = Optional TrustAll
 Server = https://pkgs.omarchy.org/edge/$arch
 
 # Repositories for debug symbol packages.
--- a/default/pacman/pacman-rc.conf
+++ b/default/pacman/pacman-rc.conf
@@ -26,5 +26,4 @@ Include = /etc/pacman.d/mirrorlist
 Include = /etc/pacman.d/mirrorlist
 
 [omarchy]
-SigLevel = Optional TrustAll
 Server = https://pkgs.omarchy.org/edge/$arch
--- a/default/pacman/pacman-stable.conf
+++ b/default/pacman/pacman-stable.conf
@@ -26,5 +26,4 @@ Include = /etc/pacman.d/mirrorlist
 Include = /etc/pacman.d/mirrorlist
 
 [omarchy]
-SigLevel = Optional TrustAll
 Server = https://pkgs.omarchy.org/stable/$arch
--- a/migrations/1787589206.sh
+++ b/migrations/1787589206.sh
@@ -1,0 +1,20 @@
+echo "Require signed packages from the Omarchy repository"
+
+# The [omarchy] repo predates the Omarchy packaging key, so existing installs
+# carry a SigLevel override that also accepts unsigned packages. Packages are
+# signed now, so drop the override and let the repo inherit the global
+# SigLevel = Required DatabaseOptional like every other repo. Machine-wide and
+# self-detecting, so another user's rerun no-ops.
+omarchy_sig_override='SigLevel = Optional TrustAll'
+
+if [[ -f /etc/pacman.conf ]] &&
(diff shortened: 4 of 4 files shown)


Bierdieb:
Field verification from a fresh 4.0.0 install, in case it helps with the
rollout question from #2712 ("some work to do to get everyone updated with
the key so we can flip the rule in pacman.conf").

I set `[omarchy]` to `SigLevel = Required DatabaseOptional` by hand on
2026-08-24, a few hours before this PR appeared, on a machine installed the
day before.

**The keyring side looks ready.** Flipping the override on an install that had
only ever run `omarchy update` worked immediately, with no manual key import.
`omarchy-keyring 20251027-1` was already present and
`40DFB630FF42BCFFB047046CF0134EE680CAC571` locally signed by the Pacman
Keyring Master Key. Two packages came from the repo right after the flip and
both validated:

    aether     4.29.4-1   Validated By: SHA-256 Sum  Signature
    localsend  1.18.2-1   Validated By: SHA-256 Sum  Signature

**`DatabaseOptional` is load-bearing.**
`https://pkgs.omarchy.org/stable/x86_64/omarchy.db.sig` returns 404, while
package signatures are served (checked
`omarchy-4.0.0-1-any.pkg.tar.zst.sig` and
`omarchy-keyring-20251027-1-any.pkg.tar.zst.sig`). Inheriting the global
`Required DatabaseOptional` is therefore correct; a plain `Required` would
break `pacman -Sy`.

**One gap this PR does not close.** On this install, 879 of 1014 packages
carried `Validated By: SHA-256 Sum` with no signature at all. The split falls
exactly on the installer run:

- 16:34:45–16:39:05 (installer, offline repo): 879 packages, checksum only —
  including `linux`, `systemd`, `glibc`, `sudo`, `openssh`, `pacman`, and
  `omarchy-keyring` itself
- 17:24 onward (normal pacman): 130 packages, `Signature`

So on a fresh system the trust anchor itself arrives unverified, and
`migrations/1787589206.sh` hardens the configuration without re-validating
what the installer already placed. Reinstalling every repo package afterwards
brought this machine to 1014/1014 signature-validated with no breakage
(`pacman -S $(pacman -Qnq)`, with install reasons saved beforehand and
restored via `pacman -D --asdeps`/`--asexplicit` — without that, every package
ends up marked explicit and orphan detection is silently lost).

That might be worth a follow-up: either signing the offline repo used by the
installer, or documenting the gap in the security manual.

And as you might already know: Opus5 did the diagnostics.

PR #8087, opened 2026-08-24
network: self-heal the bar icon after a NetworkManager restart

susamn:
## Summary

- `Quickshell.Networking` (the compiled singleton `shell/plugins/panels/network/Panel.qml`'s bar-pill `kind`/`icon` are bound to) exposes no reload/reconnect call. Restarting the NetworkManager daemon (`systemctl restart NetworkManager`) replaces its D-Bus service instance underneath Quickshell's live subscriptions, which then go silently orphaned — the bar icon freezes at its pre-restart value (typically "disconnected") forever, even after connectivity is actually restored, since this widget has no periodic re-poll fallback to correct it.
- Watches `NameOwnerChanged` on `org.freedesktop.NetworkManager` via `dbus-monitor` and, on the release event (the moment the old subscriptions actually die — NM's daemon restart is a stop-then-start, so the name changes hands as two separate release/claim signals rather than one atomic swap), debounces and restarts the whole Quickshell shell so `Quickshell.Networking` reconnects fresh.
- Deferred while the network panel is open, so an in-progress password prompt isn't yanked away by the restart.
- Lock-guarded (`/tmp/omarchy-network-nm-selfheal.lock`), since a multi-monitor setup instantiates one `Panel.qml`/watcher per screen and would otherwise fire `omarchy-restart-shell` more than once for the same event — it does a serialized kill-then-relaunch of the single shared process with no lock of its own.

## Context

The actual defect arguably lives in Quickshell's compiled `Quickshell.Networking` backend (no resync-on-daemon-restart, no reload API) — this PR is a pragmatic Omarchy-side mitigation, not a fix of the root cause, in case maintainers would rather track/fix that upstream instead.

## Test plan

- [x] `./test/shell.d/network-test.sh` — new unit tests for `extractDbusString`/`nmServiceOwnerLost` plus structural assertions on the watcher/self-heal wiring, all passing
- [x] `./test/all` — full suite passes (5 pre-existing, unrelated environment failures: missing `omarchy-pkgs` checkout, an existing graphical-geometry test, and an existing IPC test that collides with an already-running shell instance)
- [x] Live-verified via `omarchy dev link`: reproduced the original bug (`sudo systemctl restart NetworkManager`), confirmed the bar icon self-heals to "connected" without manually running `omarchy restart shell`
- [x] Confirmed no restart loop after the single NM restart
- [x] Confirmed the multi-monitor lock prevents a duplicate shell relaunch (two `Panel.qml`/watcher instances on a two-monitor setup, one shell restart)

### DIFF
--- a/shell/plugins/panels/network/Model.js
+++ b/shell/plugins/panels/network/Model.js
@@ -14,6 +14,26 @@ function wifiIconFor(strength) {
   return icons[index]
 }
 
+// `dbus-monitor` prints each NameOwnerChanged arg on its own line, e.g.
+// `   string "org.freedesktop.NetworkManager"`. Returns the quoted value, or
+// null for a line that carries none (the signal header, blank lines).
+function extractDbusString(line) {
+  var match = String(line || "").match(/^\s*string\s+"([^"]*)"/)
+  return match ? match[1] : null
+}
+
+// `systemctl restart` stops the old daemon before starting the new one, so
+// NetworkManager's D-Bus name changes hands as two separate NameOwnerChanged
+// signals rather than one atomic swap: first the old owner releases the name
+// (new owner empty), then the new process claims it (old owner empty). The
+// release is what actually orphans anything already subscribed to the old
+// owner, so that's the moment to react to — no need to wait for the reclaim.
+// An empty old owner (the name's first-ever claim, e.g. at boot) is not a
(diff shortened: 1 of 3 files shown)


cpwrs:
Upstream fix PR is up as well
quickshell-mirror/quickshell#991

PR #8094, opened 2026-08-24
Add Devin CLI as a default agent option

joaomcarlos:
## Summary

- Adds `devin` as a selectable default coding agent alongside the existing 10 agents
- Devin CLI ships its own installer (`curl -fsSL https://cli.devin.ai/install.sh | bash`) and is not available through mise, so this introduces a `skip_mise` path in `omarchy-default-agent` that runs the installer directly when the binary is missing
- The launcher starts Devin with `--permission-mode dangerous --respect-workspace-trust false`, matching the unattended "don't stop to ask" mode other agents use
- Adds a menu entry under _Setup > Defaults > Agent_ and updates the manual

## Files changed

- `bin/omarchy-default-agent` — `devin` case with `skip_mise=1`; curl-bash installer fallback
- `bin/omarchy-agent` — `devin` launch case with permission/trust flags
- `default/omarchy/omarchy-menu.jsonc` — `setup.default.agent.devin` entry
- `test/shell.d/default-agent-test.sh` — devin in canonicalization, skip-mise, and launch assertions
- `manual/17-ai.md` — table entry + note about Devin's non-mise install method

## Design notes

Devin is the first agent that doesn't go through mise. I considered alternatives (publishing to npm/GitHub releases for `ubi:`/`npm:` resolution), but Devin's distribution model is a self-managed versioned installer at `static.devin.ai/cli/`, which doesn't fit any mise prefix. The `skip_mise` flag is a minimal, opt-in pattern: only agents that explicitly set it bypass the mise flow, and every other agent's behavior is unchanged.

The curl-bash installer runs inside the same floating terminal that `--install` uses for mise agents, so the UX is consistent: pick an uninstalled agent → terminal opens → installer runs → agent launches.

#### Test plan

- [x] `./test/shell.d/default-agent-test.sh` — all 29 assertions pass
- [x] `./test/cli` — all CLI/metadata tests pass
- [ ] Manual: `omarchy default agent devin` on a machine with Devin installed → sets default, launches agent
- [ ] Manual: `omarchy default agent devin` on a machine without Devin → opens terminal, runs installer, launches

Generated with [Devin](https://devin.ai)


joaomcarlos:
Asked devin to research how to add devin to the list of default agent options... I guess it was too eager :D

PR #8107, opened 2026-08-24
Fix T2 Mac Wi-Fi suspend by unloading brcmfmac before sleep

### LINKED ISSUE #8106: T2 Mac: lid close never suspends, brcmfmac PCIe D3 timeout aborts every suspend attempt
## What happened

On a T2 MacBook (MacBookPro15,4, i5-8257U), closing the lid never actually suspends. `systemd-logind` starts a suspend, it fails within ~12 seconds, and logind immediately retries — over and over. Over one hour with the lid closed, the machine attempted to suspend **115 times** and never actually slept once. The screen was off the whole time, so from the lid it looked like it had hibernated; in reality the CPU stayed fully awake the entire hour, which is why the battery drains far faster than it should.

## Root cause

The kernel log at every attempt shows the same failure, from the PCIe Wi-Fi driver:

```
brcmfmac 0000:01:00.0: brcmf_pcie_pm_enter_D3: Timeout on response for entering D3 substate
brcmfmac 0000:01:00.0: PM: pci_pm_suspend(): brcmf_pcie_pm_enter_D3 [brcmfmac] returns -5
brcmfmac 0000:01:00.0: PM: dpm_run_callback(): pci_pm_suspend returns -5
brcmfmac 0000:01:00.0: PM: failed to suspend: error -5
PM: Some devices failed to suspend, or early wake event detected
```

The BCM4377b PCIe Wi-Fi chip times out responding to the D3 power-state request, `brcmfmac`'s `.suspend()` callback returns `-EIO`, and the kernel aborts the entire suspend and resumes every device. `systemd-sleep` then reports:

```
systemd-sleep[…]: Failed to put system to sleep. System resumed again: Input/output error
```

`systemd-suspend.service` fails, and whatever asked for suspend (lid close, `systemctl suspend`) just gets called again, so this repeats indefinitely.

## Reproduction

- Hardware: MacBookPro15,4 (2019 13", i5-8257U, T2 chip), Wi-Fi is `Broadcom BCM4377b` (`14e4` behind the T2, PCIe-attached, driven by `brcmfmac`)
- `uname -r`: `7.1.8-arch1-Watanare-T2-3-t2` (linux-t2)
- Omarchy version: 4.0.0-1, stable channel
- Steps: close the lid (or run `systemctl suspend`) on affected hardware. Watch `journalctl -b -k -f` — the D3 timeout and `PM: suspend exit` appear within seconds, and `journalctl -u systemd-logind` shows the retry loop.

## Fix

Unloading `brcmfmac`/`brcmfmac_wcc` before suspend and reloading it on resume avoids the broken D3 handshake entirely — this is the standard workaround used elsewhere for this chip family. I tested this on the affected machine: without the fix, every suspend attempt fails the same way within ~12 seconds; with a `pre`/`post` `systemd-sleep` hook doing the unload/reload, suspend completes cleanly (kernel stays asleep for the full duration, e.g. 28s in one test), Wi-Fi reconnects automatically on resume, and there is no D3 timeout in the log.

I've opened a PR (#TBD) adding this as a hardware-conditional `/usr/lib/systemd/system-sleep/` hook, gated on the T2 PCI ID the same way `fix-t2.sh` and `fix-brcmfmac-supplicant.sh` already are, plus a migration so existing T2 installs pick it up without a reinstall.

## Diagnostics

`omarchy debug --no-sudo --print` isn't available on 4.0.0-1 (`Unknown Omarchy command`); happy to attach whatever diagnostic format is preferred.

nelKorajkic:
## Summary

- T2 Macs' PCIe Wi-Fi (BCM4377/4378, driven by `brcmfmac`) times out entering D3 during suspend, which aborts the *entire* suspend/hibernate cycle. Every lid-close retries and fails the same way, so the machine never actually sleeps — it stays fully awake in a suspend/resume loop that looks like hibernation but drains the battery. Fixes #8106.
- Adds `default/systemd/system-sleep/t2-wifi-suspend`, a hook that unloads `brcmfmac`/`brcmfmac_wcc` before sleep and reloads it on resume instead of asking the firmware to do something it can't.
- Installs it the same way `fix-brcmfmac-supplicant.sh` installs its quirk: a new `install/hardware/apple/fix-brcmfmac-suspend.sh` leaf gated on the T2 PCI ID (`106b:180[12]`), registered in `install/hardware/all.sh` right after its sibling.
- Adds a migration (`migrations/1787605234.sh`) so an existing T2 install gets the hook without a reinstall — the install-time leaf only reaches machines set up after this ships.

## Why unload the driver instead of a d3cold/ASPM tweak

The failure is a firmware response *timeout* during the D3 entry handshake itself (`brcmf_pcie_pm_enter_D3: Timeout on response for entering D3 substate`, `error -5`), not a D3cold power-cut issue — so it's a different failure mode than the existing `fix-suspend-nvme.sh` (which disables `d3cold_allowed` for the NVMe device on older MacBooks). Unloading the driver sidesteps the handshake entirely rather than tuning it, which is the workaround this chip family needs elsewhere too.

## Testing

- `./test/shell.d/brcmfmac-suspend-test.sh` (new) — leaf installs the hook only on T2 hardware, migration is idempotent and skips non-T2 machines, same chatty-`lspci` stub pattern as `t2-hardware-test.sh` / `brcmfmac-supplicant-test.sh` to cover the SIGPIPE/pipefail case (#6608).
- `./test/all`
- Verified live on the affected hardware (MacBookPro15,4, T2, BCM4377b):
  - Before: every `systemctl suspend` fails within ~12s with the D3 timeout in `dmesg` and `Failed to put system to sleep. System resumed again: Input/output error` from `systemd-sleep`.
  - After: `systemctl suspend` sleeps for the full duration (verified a 28s cycle), resumes cleanly (`systemd-suspend.service: Deactivated successfully`), Wi-Fi reconnects automatically, zero D3 timeouts in the log.

## Test plan

- [x] `./test/shell.d/brcmfmac-suspend-test.sh`
- [x] `./test/all`
- [x] Live suspend/resume test on T2 Mac hardware


spuder:
Thanks for finding this. My macbook air looses wifi after sleep
@nelKorajkic You may consider changing the target branch from `master` to `quattro` to get this merged in quicker.

spuder:
While cherry picking and manually testing the diff, claude seems to think it needs the following adjustments

- pre: bring down any netdev bound to brcmfmac (found dynamically via /sys/bus/pci/drivers/brcmfmac/*/net/*, so it survives interface renames) and rfkill block wifi, then retry modprobe -r up to 5 times (200ms apart) in case NetworkManager is still mid-teardown.
- post: modprobe brcmfmac_wcc (pulls in brcmfmac via module deps, same as your original local hook) and rfkill unblock wifi. NetworkManager picks the interface back up and reconnects on its own once the driver reappears — no manual ip link up needed.



```
#!/bin/bash

# BCM4377/4378-family PCIe Wi-Fi (brcmfmac) times out entering D3 during
# suspend on T2 Macs, aborting the whole suspend/hibernate cycle:
#
#   brcmfmac 0000:01:00.0: brcmf_pcie_pm_enter_D3: Timeout on response for
#     entering D3 substate
#   brcmfmac 0000:01:00.0: PM: failed to suspend: error -5
#   PM: Some devices failed to suspend, or early wake event detected
#
# systemd-sleep reports that as "Failed to put system to sleep: Input/output
# error" and the caller (lid close, systemctl suspend) just retries, so the
# machine never actually sleeps: it stays fully awake in a suspend/resume
# loop that looks like hibernation from the lid but drains the battery.
#
# Unload the driver before sleep and reload it on resume instead of asking
# firmware to do something it can't. See
# install/hardware/apple/fix-brcmfmac-suspend.sh (omarchy PR #8107) for
# where the upstream version of this hook is installed.
#
# Local addition: modprobe -r fails silently (and leaves brcmfmac bound, so
# the D3 timeout still happens) if the netdev is still up when we try to
# unload the driver -- NetworkManager doesn't release it fast enough on its
# own. So take the interface down and soft-block the radio first, and retry
# the unload a few times in case NM is mid-teardown.

case "$1" in
  pre)
    for dev in /sys/bus/pci/drivers/brcmfmac/*/net/*; do
      [ -e "$dev" ] || continue
      ip link set "$(basename "$dev")" down 2>/dev/null
    done
    rfkill block wifi 2>/dev/null

    for _ in 1 2 3 4 5; do
      modprobe -r brcmfmac_wcc brcmfmac 2>/dev/null && break
      sleep 0.2
    done
    ;;
  post)
    modprobe brcmfmac_wcc 2>/dev/null
    rfkill unblock wifi 2>/dev/null
    ;;
esac
```

### Testing

Simulated sleep


```
sudo /usr/lib/systemd/system-sleep/t2-wifi-suspend pre
lsmod | grep brcm        # should show nothing — driver unloaded
ip link                  # wifi interface should be gone
sudo /usr/lib/systemd/system-sleep/t2-wifi-suspend post
sleep 3
lsmod | grep brcm        # brcmfmac + brcmfmac_wcc back
nmcli device status      # wifi device should reappear and reconnect on its own
```

PR #8129, opened 2026-08-25
[Security] Stop USB device names from being executed as Hyprland Lua

acrogenesis:
A HID gadget can put an arbitrary string in the USB product/device name Hyprland reports. `omarchy-toggle-input-device` interpolated that name into `hyprctl eval` and into a generated Lua file under `~/.local/state/omarchy/toggles/hypr/`, which Hyprland sources on every start.

`XF86TouchpadToggle` (HID F21) is bound with `locked = true`, so this runs on a locked session as the logged-in user. Disable persists after the gadget is unplugged. Not network RCE and not root by itself; High for an evil-maid / malicious-USB take of the user account.

Public report: [Omarchy Exploit](https://www.youtube.com/watch?v=pgNvD9DaZxY) (Jorrit Jongma / Chainfire).

## Fix

The disable is no longer executable Lua anywhere. The device name is stored as plain-text data in a `*-disabled-name` sidecar and read back by a packaged module (`default/hypr/disabled-input-device.lua`) on every reload; the live `hyprctl eval` Lua-quotes the name and rejects control characters outright. The old generated-Lua templates are dropped entirely — so nothing can seed a disabled state to `/etc/skel` — leaving the sidecar, read from a hardcoded `~/.local/state` to match the sibling toggle tools, as the single source of truth.

A migration recovers the device name from any already-generated toggle Lua and deletes the file, sanitizing installs that already ran the vulnerable version.

## Same class, hardened alongside

Reviewing the fix surfaced the identical "hyprctl name → Lua" pattern in the monitor helpers. All four scripts that write an output name into Lua — `omarchy-hyprland-monitor-internal`, `-internal-mirror`, `-clamshell`, and `-scaling` — now validate it against a plain-connector-name pattern before it can be written as code. Real connector names (`eDP-1`, `DP-3`, …) are unaffected; a user-created headless output carrying an arbitrary name is rejected. `paths.lua` also now treats a set-but-empty `XDG_STATE_HOME` as unset, matching the bash side so state is never read from the filesystem root.

## Tests

- `test/shell.d/toggle-input-device-test.sh` — Lua-quoting, data-only persistence, the public PoC string, migration recovery of real names, and the wedge / idempotence paths.
- `test/shell.d/monitor-output-name-test.sh` — each monitor script accepts a real connector and rejects a metacharacter-laced name.
- `test/shell.d/hyprland-paths-test.sh` — empty vs set XDG variables.

### DIFF
--- a/bin/omarchy-hyprland-monitor-clamshell
+++ b/bin/omarchy-hyprland-monitor-clamshell
@@ -11,6 +11,14 @@ MONITOR_LUA="$HOME/.config/hypr/monitors.lua"
 
 INTERNAL=$(omarchy-hyprland-monitor-laptop)
 
+# INTERNAL is written into generated Lua and hyprctl eval/dispatch below, so a
+# name that is not a plain connector string could execute on the next reload.
+# Names come from hyprctl; a user-created headless output can carry anything.
+if [[ -n $INTERNAL && ! $INTERNAL =~ ^[A-Za-z0-9._-]+$ ]]; then
+  echo "Refusing unsafe internal monitor name" >&2
+  exit 1
+fi
+
 valid_scale() {
   [[ $1 =~ ^[0-9]+([.][0-9]+)?$ ]]
 }
--- a/bin/omarchy-hyprland-monitor-internal
+++ b/bin/omarchy-hyprland-monitor-internal
@@ -28,6 +28,13 @@ off() {
     exit 1
   fi
 
+  # The name is written into generated Lua below, so only a plain connector
+  # name may pass; anything else could execute on the next reload.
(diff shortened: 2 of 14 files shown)


Chainfire:
Notes: 

XF86TouchpadOn (HID F22) also triggers the immediate path, though not the persistent LUA path. 

XF86TouchpadOff (HID F23) does nothing (seemingly remapped to actual F23). 

HID Ctrl+Super+F24 should also take the HID F21 (XF86TouchpadToggle) path, both immediate trigger and persistent LUA, according to Claude but I was not able to reproduce it.

rick:
fuckin sexy

lgtm

rick (approved):
Image

PR #8131, opened 2026-08-25
Prefer apple-panel-bl over the Asahi DSI backlight decoy

### LINKED ISSUE #8125: Brightness slider/keys silently no-op on Apple Silicon: wrong backlight device picked (228600000.dsi.0 instead of apple-panel-bl)
## What happened

On an Apple Silicon MacBook (M1, Asahi kernel), adjusting brightness via the bar's monitor panel slider (and the `omarchy brightness display` CLI) reports success and updates the displayed percentage, but the screen's actual brightness never changes.

`/sys/class/backlight/` exposes two devices on this hardware:

```
$ brightnessctl -l
Device 'apple-panel-bl' of class 'backlight':
	Current brightness: 356 (70%)
	Max brightness: 509
Device '228600000.dsi.0' of class 'backlight':
	Current brightness: 255 (100%)
	Max brightness: 255
```

`apple-panel-bl` is the one that actually drives the panel — confirmed by directly setting it and watching the screen dim/brighten. `228600000.dsi.0` accepts writes and reports the new value back correctly, but has zero visible effect on the screen.

`omarchy-hw-display` (used by `omarchy-brightness-display`, which backs both the bar's brightness slider and brightness keys) doesn't have `apple-panel-bl` in its device-priority list (`gmux_backlight`, `amdgpu_bl*`, `intel_backlight`, `acpi_video*`), so it falls back to picking the alphabetically-first entry in `/sys/class/backlight`, which is `228600000.dsi.0` on this machine — the non-functional decoy.

## Expected

Brightness controls (bar slider and hardware keys) should adjust the device that actually drives the panel.

## Steps to reproduce

1. On an Apple Silicon machine (Asahi) where `/sys/class/backlight/` contains both `apple-panel-bl` and a numeric DSI backlight node (e.g. `228600000.dsi.0`)
2. Open the bar's monitor panel (`omarchy.monitor` widget) and drag the brightness slider, or press a brightness function key
3. Observe: the slider/OSD percentage updates, but the screen brightness does not change
4. Confirm the real device works: `brightnessctl -d apple-panel-bl set 10%` visibly dims the screen; `brightnessctl -d 228600000.dsi.0 set 10%` does not

## Workaround in use

Pointed `omarchy-hw-display`'s existing `OMARCHY_BACKLIGHT_PATH` override at a directory containing only a symlink to `apple-panel-bl`, set via `hl.env(...)` in `~/.config/hypr/hyprland.lua`. Works, but this shouldn't be necessary — `omarchy-hw-display` should prefer `apple-panel-bl` (or otherwise avoid picking a non-functional raw DSI backlight node) out of the box on Apple Silicon hardware.

## System details

```
$ omarchy version
4.0.1rc1-1

$ omarchy channel current
rc

$ uname -a
Linux omarchybookpro 7.1.6-1-1-ARCH #1 SMP PREEMPT_DYNAMIC Sat, 08 Aug 2026 17:14:04 +0000 aarch64 GNU/Linux

CPU: Apple M1 (Icestorm-M1 / Firestorm-M1), device tree: apple,j293 / apple,t8103 (MacBook Pro 13", 2020)
```

(Note: `omarchy debug --no-sudo --print` from the contributing docs isn't a recognized command on this rc build — omitted from this report.)

Chessing234:
On Apple Silicon, `/sys/class/backlight` lists both `apple-panel-bl` (the real panel) and a numeric DSI node that sorts first alphabetically and accepts writes with no visible effect. `omarchy-hw-display` had no `apple-panel-bl` in its candidate list, so it fell through to that decoy and brightness slider/keys no-op'd.

Put `apple-panel-bl` after gmux (T2 dual-GPU Macs still want gmux first) and before GPU PWM devices. Keep excluding `appletb_backlight` (Touch Bar). Existing gmux/amdgpu/intel/acpi order is unchanged.

`hw-display-test.sh` is at 12 assertions: Asahi DSI decoy loses to `apple-panel-bl`, gmux still wins on T2, Touch Bar stays unused.

Fixes #8125

### DIFF
--- a/bin/omarchy-hw-display
+++ b/bin/omarchy-hw-display
@@ -11,7 +11,9 @@ backlight_path="${OMARCHY_BACKLIGHT_PATH:-/sys/class/backlight}"
 device="$(ls -1 "$backlight_path" 2>/dev/null | grep -vx appletb_backlight | head -n1)"
 # gmux comes first: apple-gmux only registers when the kernel has already picked it, and on
 # dual-GPU Macs the GPU's own PWM stops driving the panel once that GPU suspends.
-for candidate in "$backlight_path"/gmux_backlight "$backlight_path"/amdgpu_bl* "$backlight_path"/intel_backlight "$backlight_path"/acpi_video*; do
+# apple-panel-bl is Asahi's real panel backlight; without it here the alphabetical fallback
+# picks a DSI decoy (e.g. 228600000.dsi.0) that accepts writes with no visible effect.
+for candidate in "$backlight_path"/gmux_backlight "$backlight_path"/apple-panel-bl "$backlight_path"/amdgpu_bl* "$backlight_path"/intel_backlight "$backlight_path"/acpi_video*; do
   if [[ -e $candidate ]]; then
     device="${candidate##*/}"
     break
--- a/test/shell.d/hw-display-test.sh
+++ b/test/shell.d/hw-display-test.sh
@@ -57,6 +57,26 @@ device=$(hw_display)
 [[ $device == "gmux_backlight" ]] || fail "gmux outranks the GPU backlights on dual-GPU Macs" "actual: $device"
 pass "gmux outranks the GPU backlights on dual-GPU Macs"
 
+write_backlights 228600000.dsi.0 apple-panel-bl
+device=$(hw_display)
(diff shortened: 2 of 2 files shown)


Chessing234:

Chessing234:
folded into #12267 for easier review.

PR #8136, opened 2026-08-25
Enable Sunshine by its canonical systemd user unit

### LINKED ISSUE #7967: omarchy install service sunshine fails to enable unit on fresh install
## What happened

`omarchy install service sunshine` installs the `sunshine` package successfully, but then fails with:

```
```

Because `omarchy-install-service-sunshine` has `set -e`, the script aborts right there. Everything after the `systemctl --user enable --now sunshine` line — opening the Moonlight firewall ports, installing the "Sunshine Admin" webapp, and adding the Hyprland autostart entry — never runs.

## Root cause

The Arch `sunshine` package ships its user systemd unit as:

```
/usr/lib/systemd/user/app-dev.lizardbyte.app.Sunshine.service
```

with `Alias=sunshine.service` in its `[Install]` section. systemd only resolves that alias once the unit has already been enabled by its real name — the alias symlink (`~/.config/systemd/user/sunshine.service`) doesn't exist yet on a fresh install, so `systemctl --user enable --now sunshine` fails with "Unit sunshine.service does not exist".

Manually running:

```bash
```

succeeds, creates the `sunshine.service` alias symlink, and starts Sunshine correctly. (After that one-time fix, `sunshine.service` resolves fine going forward.)

## Expected behavior

`omarchy install service sunshine` should enable Sunshine on a fresh install without manual intervention, and should still perform the firewall/webapp/autostart steps afterward.

## Suggested fix

In `omarchy-install-service-sunshine`, enable the service by its canonical unit name instead of the alias:

```bash
```

## Steps to reproduce

1. Fresh Omarchy system, Sunshine not previously installed.
2. Run `omarchy install service sunshine`.
3. Observe `Failed to enable unit: Unit sunshine.service does not exist` after package install, and that firewall ports/webapp/autostart are never set up.

## System details

- Omarchy version: 4.0.0-1 (stable channel)
- Kernel: Linux 7.1.8-arch1-3 x86_64
- CPU: Intel(R) Core(TM) i5-6400 CPU @ 2.70GHz
- GPU: NVIDIA GeForce GTX 970 (GM204)
- sunshine package version: 2026.516.143833-4-4

Chessing234:
Arch's Sunshine package ships `app-dev.lizardbyte.app.Sunshine.service` with `Alias=sunshine.service`. systemd only resolves that alias after the unit has been enabled by its real name, so `systemctl --user enable --now sunshine` fails on a fresh install and `set -e` skips the firewall, webapp, and autostart steps.

Enable the canonical unit when the file is on disk (and still use that name if the alias is not resolvable yet). A machine that already has `sunshine.service` enabled is a no-op. Remove disables both names.

`sunshine-install-test.sh` covers a fresh canonical enable and an already-enabled alias, and checks later steps still run.

Fixes #7967

### DIFF
--- a/bin/omarchy-install-service-sunshine
+++ b/bin/omarchy-install-service-sunshine
@@ -69,9 +69,24 @@ enable_hyprland_autostart() {
   fi
 }
 
+enable_sunshine_unit() {
+  local unit=app-dev.lizardbyte.app.Sunshine.service
+
+  # systemd only resolves Alias= after enable by the real unit name. Arch
+  # ships this unit with Alias=sunshine.service, so enabling "sunshine" fails
+  # on a fresh install ("Unit sunshine.service does not exist").
+  if [[ -f /usr/lib/systemd/user/$unit ]]; then
+    systemctl --user enable --now "$unit"
+  elif systemctl --user is-enabled sunshine.service >/dev/null 2>&1; then
+    systemctl --user enable --now sunshine.service
+  else
+    systemctl --user enable --now "$unit"
+  fi
+}
+
 echo "Installing Sunshine..."
 omarchy-pkg-add sunshine
-systemctl --user enable --now sunshine
+enable_sunshine_unit
 
 echo "Opening Sunshine firewall ports..."
 open_ufw_ports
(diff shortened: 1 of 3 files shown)


Chessing234:

Chessing234:
folded into #12271 for easier review.

PR #8142, opened 2026-08-25
Keep the screensaver up while launch hops between monitors

### LINKED ISSUE #8114: [Quattro] omarchy-launch-screensaver creates focus-check race condition on 3+ monitor setups causing immediate self-termination
### System details
- **Hardware**: Framework Desktop, AMD Ryzen AI Max 395 (Radeon 8060S Graphics)
- **Omarchy Version**: 4.0.0-1 (Quattro)
- **Hyprland Configuration**: Lua parser (`~/.config/hypr/hyprland.lua`)
- **Display Setup**: 3 monitors (`DP-1` portrait, `HDMI-A-1` ultrawide, `DP-4` landscape) vs 2 monitors tested
- **Terminal**: Kitty

### What's wrong?
On multi-monitor setups with 3 or more displays, `omarchy-launch-screensaver` triggers and then immediately self-terminates after a few seconds without any user interaction. On a 2-monitor system with identical Omarchy Quattro configuration, the screensaver runs as expected.

When the screensavers self-terminate, `omarchy-shell`'s idle manager (`Service.qml`) interprets the closed windows as a manual dismissal, issues `omarchy-system-wake`, and cancels the idle sequence—preventing `idle.lock` (300s) and display DPMS power-down from ever activating.

### Mechanism & Root Cause

1. **Sequential Output Launch Loop**:
   In `/usr/bin/omarchy-launch-screensaver`, screensavers are spawned sequentially across all connected displays:
   ```bash
   for m in $(hyprctl monitors -j | jq -r '.[].name'); do
     hypr_focus_monitor "$m"
     case $terminal in ... esac
     wait_for_screensaver_window
   done
   hypr_focus_monitor "$focused"
   ```
   Each monitor iteration focuses the monitor, spawns the terminal screensaver window, and waits for the corresponding socket2 `openwindow` event.

2. **The 3+ Monitor Timing Race**:
   In `/usr/bin/omarchy-screensaver`, the main loop starts checking focus immediately with a 1-second timeout:
   ```bash
   while pgrep -t "${tty#/dev/}" -x ttfx >/dev/null; do
       exit_screensaver
     fi
   done
   ```
   - On a **2-monitor system**, the loop completes quickly enough that both windows map and focus is restored to the primary monitor within ~1 second.
   - On a **3+ monitor system**, cycling through all three displays takes 2–4 seconds.
   - While the launcher is busy focusing Monitor 2 and Monitor 3, Monitor 1's screensaver process reaches its first 1-second timeout and evaluates `screensaver_in_focus`.

3. **Global Active Window Query**:
   `screensaver_in_focus()` checks global compositor focus:
   ```bash
   screensaver_in_focus() {
     hyprctl activewindow -j | jq -e '.class == "org.omarchy.screensaver"' >/dev/null 2>&1
   }
   ```
   Because compositor focus is actively transitioning to Monitor 2/3 (or momentarily lands on existing windows on those workspaces before the screensaver terminal is mapped), Monitor 1 sees that `activewindow` is not its own screensaver class. `! screensaver_in_focus` evaluates to true, executing `exit_screensaver` (`pkill -f org.omarchy.screensaver`) and killing all screensavers across all monitors.

### Steps to Reproduce
1. Connect 3 displays with active application windows on each workspace.
2. Trigger the screensaver via `omarchy-launch-screensaver` or wait for `idle.screensaver` (150s).
3. Observe Monitor 1 launch, followed by Monitor 2/3, followed by immediate teardown within 3–6 seconds without user input.

### Suggested Solutions
1. **Startup Grace Period**: Add an initial grace period (e.g. 5–6 seconds) to `omarchy-screensaver` before evaluating `screensaver_in_focus`, giving multi-monitor launch loops sufficient time to finish mapping all screens.
2. **Event-Driven Dismissal**: Rather than polling global `activewindow` every second in each terminal process, rely on compositor idle/wake notifications from `omarchy-shell` and terminal keypress/signal handlers to handle screensaver dismissal.

> Filed by Gemini via Antigravity CLI.

Chessing234:
`omarchy-launch-screensaver` focuses each monitor in turn while spawning. `hyprctl activewindow` is global, so on 3+ monitors the first screensaver's 1s poll could see another output's client (or no class) and `exit_screensaver`, which `pkill`s every `org.omarchy.screensaver` instance. The idle manager treated that as a user dismiss and cancelled lock/DPMS.

`screensaver_in_focus` now stays up when the focused window is another screensaver or has no class (monitor switch). It dismisses only when a real non-screensaver client is focused. Real dismiss still kills all instances.

`screensaver-test.sh` covers screensaver class, empty class, and a foreign client.

Fixes #8114

### DIFF
--- a/bin/omarchy-screensaver
+++ b/bin/omarchy-screensaver
@@ -2,8 +2,20 @@
 
 # omarchy:summary=Run the Omarchy screensaver using random effects from TTE.
 
+# activewindow is global. Launch hops focus across screensaver windows, and the
+# compositor can report no class while switching monitors. Stay up unless a real
+# non-screensaver client is focused.
 screensaver_in_focus() {
-  hyprctl activewindow -j | jq -e '.class == "org.omarchy.screensaver"' >/dev/null 2>&1
+  local json class
+
+  json=$(hyprctl activewindow -j 2>/dev/null) || return 0
+  class=$(jq -r '.class // empty' <<<"$json" 2>/dev/null) || return 0
+
+  if [[ -z $class || $class == "null" ]]; then
+    return 0
+  fi
+
+  [[ $class == "org.omarchy.screensaver" ]]
 }
 
 exit_screensaver() {
@@ -12,6 +24,14 @@ exit_screensaver() {
   pkill -f '[o]rg.omarchy.screensaver' 2>/dev/null
   exit 0
 }
+
+if [[ ${1:-} == "--check-focus" ]]; then
+  if screensaver_in_focus; then
+    exit 0
+  else
+    exit 1
+  fi
+fi
 
 # Exit the screensaver on signals and input from keyboard and mouse
 trap exit_screensaver SIGINT SIGTERM SIGHUP SIGQUIT
--- a/test/shell.d/screensaver-test.sh
+++ b/test/shell.d/screensaver-test.sh
@@ -1,0 +1,91 @@
+#!/bin/bash
+
+set -euo pipefail
+
+source "$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd)/base-test.sh"
+
+require_command jq
+
+test_tmp=$(mktemp -d)
+trap 'rm -rf "$test_tmp"' EXIT
+
+stub_bin="$test_tmp/bin"
+mkdir -p "$stub_bin"
+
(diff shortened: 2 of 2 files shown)


Chessing234:

Chessing234:
folded into #12730 with the other screensaver/presentation toggles.

PR #8143, opened 2026-08-25
Disable USB autosuspend on the kernel cmdline

### LINKED ISSUE #8097: omarchy-usb-autosuspend.conf never applies: usbcore is built into the kernel, so autosuspend stays at the default
### System details

Framework Laptop 13 (AMD Ryzen AI 9 HX 370), Omarchy 4.0.0-1, kernel 7.1.8-arch1-3.

### What's wrong?

`etc/modprobe.d/omarchy-usb-autosuspend.conf` ships `options usbcore autosuspend=-1`, but it never takes effect: `usbcore` is built into the Arch kernel, not a module, so modprobe options for it are ignored.

```console
$ cat /etc/modprobe.d/omarchy-usb-autosuspend.conf
options usbcore autosuspend=-1

$ cat /sys/module/usbcore/initstate
built-in

$ grep -c usbcore /proc/modules
0

$ cat /sys/module/usbcore/parameters/autosuspend
2
```

The effective value is the kernel default of 2 seconds, i.e. USB autosuspend is on for every device that permits it — the opposite of what the drop-in intends. Hub-level state on this machine, with the drop-in installed:

```console
$ for d in 13-1 14-1; do echo "$d: $(cat /sys/bus/usb/devices/$d/power/control)"; done
13-1: auto
14-1: auto
```

### Why it matters

I found this while diagnosing a docked laptop whose keyboard and mouse vanish for ~45 s after resume (#7328 is the related display half). The dock's USB tree enumerates at resume, drops entirely nine seconds later, and returns almost a minute after that:

```
01:27:29  PM: suspend exit + xhci_hcd 0000:07:00.0: xHC error in resume, USBSTS 0x401, Reinit
01:27:30  usb 13-1 / 14-1: dock hub tree enumerates
01:27:39  usb 13-1: USB disconnect  (whole tree goes)
01:28:22  usb 13-1: re-enumerates, input devices return
```

Autosuspend being live is not necessarily the whole story for a flap like that, but the setting Omarchy believes it has applied is exactly the one that would keep dock hubs from being suspended in the first place, and right now it applies to nothing.

### Suggested fix

Move it to the kernel cmdline, where `usbcore` parameters do work — `etc/limine-entry-tool.d/omarchy-defaults.conf` already carries this kind of default:

```
KERNEL_CMDLINE[default]+=" usbcore.autosuspend=-1"
```

Verified on this machine: with `usbcore.autosuspend=-1` on the cmdline, `/sys/module/usbcore/parameters/autosuspend` reads `-1`.

Two things for a maintainer to decide, which is why this is an issue and not a PR: whether disabling USB autosuspend globally is still the intent (it costs idle power on laptops), and whether the now-dead `modprobe.d` drop-in should be dropped or kept for the case where a kernel builds `usbcore` as a module. Happy to send the patch either way.

Chessing234:
usbcore is built into the Arch kernel, so `etc/modprobe.d/omarchy-usb-autosuspend.conf` never applied and autosuspend stayed at 2.

Set `usbcore.autosuspend=-1` via `KERNEL_CMDLINE[default]+=` in `etc/limine-entry-tool.d/omarchy-defaults.conf` and stop shipping the modprobe drop-in. Existing installs get the same parameter through an append-only Limine migration, so cryptdevice/root are left alone.

A reboot is required for `/sys/module/usbcore/parameters/autosuspend` to become `-1`.

Fixes #8097

### DIFF
--- a/bin/omarchy-upgrade-to-quattro
+++ b/bin/omarchy-upgrade-to-quattro
@@ -1202,7 +1202,6 @@ run_final_system_package_upgrade() {
     --overwrite '/etc/gnupg/dirmngr.conf' \
     --overwrite '/etc/mkinitcpio.conf.d/omarchy_hooks.conf' \
     --overwrite '/etc/mkinitcpio.conf.d/thunderbolt_module.conf' \
-    --overwrite '/etc/modprobe.d/omarchy-usb-autosuspend.conf' \
     --overwrite '/etc/sddm.conf.d/10-theme.conf' \
     --overwrite '/etc/sddm.conf.d/10-wayland.conf' \
     --overwrite '/etc/sudoers.d/omarchy-asdcontrol' \
--- a/etc/limine-entry-tool.d/omarchy-defaults.conf
+++ b/etc/limine-entry-tool.d/omarchy-defaults.conf
@@ -9,6 +9,9 @@ KERNEL_CMDLINE[default]+=" quiet splash loglevel=0 systemd.show_status=false rd.
 # is fixed upstream.
 KERNEL_CMDLINE[default]+=" initramfs_async=0"
 
+# usbcore is built into the Arch kernel, so modprobe.d options never apply.
+KERNEL_CMDLINE[default]+=" usbcore.autosuspend=-1"
+
 CUSTOM_UKI_NAME="omarchy"
 ENABLE_LIMINE_FALLBACK=yes
 
--- a/etc/modprobe.d/omarchy-usb-autosuspend.conf
+++ b/etc/modprobe.d/omarchy-usb-autosuspend.conf
@@ -1,1 +1,0 @@
-options usbcore autosuspend=-1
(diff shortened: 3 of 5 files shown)


kvnloo:
This fixes the ineffective `modprobe.d` mechanism, but it also reverses the maintainer decision recorded in #6671: disabling USB autosuspend globally was declined because of the laptop power cost. `usbcore.autosuspend=-1` applies machine-wide to every eligible USB device on fresh and migrated installs, while the new tests only prove cmdline/migration mechanics, not that this broader default is acceptable.

Could we resolve that policy/device-scope decision before merging? If the global default is now intentional, please document the power tradeoff and a rollback path; otherwise this should be opt-in or targeted to affected devices rather than enabled for all Omarchy systems.

Chessing234:
closing — this reverses the #6671 decision against a global usb autosuspend disable. happy to redo as opt-in/targeted if wanted.

PR #8145, opened 2026-08-25
Persist monitor scaling onto named per-output rules

### LINKED ISSUE #8103: omarchy-hyprland-monitor-scaling doesn't persist scale on outputs with explicit per-monitor rules in monitors.lua
**What happened:** On a dual-monitor laptop setup where each display needs a different scale (ultrawide external at scale 1, laptop panel at scale 2), changing the laptop's scale via the monitor panel's SCALE control (or `omarchy hyprland monitor scaling <value>`) applies briefly, then reverts to the old value within ~2 seconds.

**Expected:** The chosen scale persists.

**Root cause (traced via `~/.local/state/omarchy/monitor-scaling.log`):** `omarchy-hyprland-monitor-scaling`'s `set_scale()` applies the new scale live via `hyprctl eval`, but only persists it back into `monitors.lua` when the file matches one of two generic patterns (`local omarchy_monitor_scale = ...` or the wildcard `hl.monitor({ output = "", ... })`). When `monitors.lua` instead has explicit per-output rules (`hl.monitor({ output = "eDP-1", ..., scale = X })` — necessary here since the two monitors need different scales), the sed-based persistence silently no-ops. Meanwhile `omarchy-hyprland-monitor-watch`'s docked-mode poller (`poll_clamshell_state`, active whenever a laptop has an external monitor active) calls `omarchy-hyprland-monitor-clamshell` every ~2s, which reads the *explicit* rule's scale from `monitors.lua` as the desired-state and force-reapplies it — reverting the unpersisted live change.

**Steps to reproduce:**
1. Connect a laptop to an external monitor with a different native scale need (extended, not mirrored).
2. Set explicit per-output `hl.monitor({...})` rules in `monitors.lua` for both outputs (e.g. via `nwg-displays`, or by hand).
3. Use the monitor panel's SCALE control (or `omarchy hyprland monitor scaling <value>`) to change the laptop's scale to a different value.
4. Within ~2 seconds, it reverts to the value still written in `monitors.lua`.

**System:** Omarchy 4.0.0-1, Dell Latitude 7420, Hyprland 0.56.2, dual monitor (LG Ultrawide `DP-2` + internal `eDP-1`).

**Workaround:** Edit `monitors.lua`'s explicit rule directly and reload — this is stable since it removes the live/config mismatch the poller reacts to.

**Related:** https://discord.com/channels/1390012484194275541/1531291452720550081 — that thread covers a different but adjacent bug in the same two scripts (a Lua-variable scale reference resolving incorrectly on the *read* side, since fixed). This report is the mirror-image gap on the *write/persist* side, which isn't fully diagnosed there — Bitslix's Aug 17 comment ("no setting is saved when using gui tools to update the scale") looks like the same symptom, unconfirmed.

---
*Claude Code assisted in both troubleshooting this issue and drafting this report.*

Chessing234:
`omarchy-hyprland-monitor-scaling` applied the scale live via `hyprctl eval` but only wrote `monitors.lua` for the generic `omarchy_monitor_scale` local or the wildcard `output = ""` rule. Explicit per-output rules (what nwg-displays writes) were left stale, and the clamshell poller re-applied the old file scale every ~2s.

It now updates the focused output's own `hl.monitor` rule (single-line and multi-line). Unrelated outputs keep their scales. If no rule matches, an explicit named rule is appended so the live change cannot be reverted.

`monitor-scaling-test.sh` covers the generic catch-all, eDP-1-only, HDMI left alone, live eval, multi-line form, and append-when-missing.

Fixes #8103

### DIFF
--- a/bin/omarchy-hyprland-monitor-scaling
+++ b/bin/omarchy-hyprland-monitor-scaling
@@ -71,6 +71,105 @@ normalize_scale() {
   awk 'NR == 1 { printf "%g\n", $0 }'
 }
 
+persist_gdk_scale() {
+  local gdk_scale="$1"
+  local file="$2"
+
+  sed -i -E \
+    -e "s|^local omarchy_gdk_scale = .*|local omarchy_gdk_scale = ${gdk_scale}|" \
+    -e 's|^hl\.env\("GDK_SCALE", ".*"\)|hl.env("GDK_SCALE", "'"$gdk_scale"'")|' \
+    "$file"
+}
+
+# Rewrite `scale = ...` inside the hl.monitor entry naming this output,
+# whether that entry is one line or the multi-line form nwg-displays writes.
+# Succeeds only when such an entry was found. Commented-out examples and
+# rules for other outputs are left alone.
+update_named_monitor_scale() {
+  local monitor="$1"
+  local scale="$2"
+  local file="$3"
+  local tmp="${file}.omarchy-scale.$$"
+
+  if awk -v monitor="$monitor" -v scale="$scale" '
+    function strip_comment(s) {
+      sub(/--.*/, "", s)
+      return s
+    }
+    function has_named_output(buf, name,    n, i, lines, line, stripped) {
+      n = split(buf, lines, "\n")
+      stripped = ""
+      for (i = 1; i <= n; i++) {
+        line = strip_comment(lines[i])
+        stripped = stripped line "\n"
+      }
+      return (index(stripped, "output = \"" name "\"") || index(stripped, "output=\"" name "\""))
+    }
+    function replace_scale(buf, value,    n, i, lines, line, out, found, last_brace, pos, c, after, end) {
+      n = split(buf, lines, "\n")
+      out = ""
+      found = 0
+      for (i = 1; i <= n; i++) {
+        line = lines[i]
+        if (!found && line !~ /^[[:space:]]*--/ && match(line, /scale[[:space:]]*=[[:space:]]*("[^"]*"|[^,;})[:space:]]+)/)) {
+          after = substr(line, RSTART + RLENGTH)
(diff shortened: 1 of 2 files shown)


Chessing234:

johnpippett:
Test results for commit `b8a9fa3342a1e72f7f9dfaa990a387a3426d965d`:

A local test with simulated Hyprland responses passed for the explicit, single-line monitor rules below.
The test used copies of this PR's scaling script and the installed Omarchy `4.0.2-1` recovery script.
It used temporary configuration and state paths. It also simulated the lid and monitor helper commands.
I did not install this PR in the active Hyprland session.

Starting configuration:

```lua
local omarchy_gdk_scale = 2
hl.env("GDK_SCALE", tostring(omarchy_gdk_scale))
hl.monitor({ output = "eDP-1", mode = "2560x1600@144", position = "0x0", scale = 1.25 })
hl.monitor({ output = "DP-1", mode = "3840x2160@60", position = "auto-right", scale = 1 })
```

With `eDP-1` selected, the test gave these results:

- Requests for `1.6`, then `1.25`, changed the scale in the existing `eDP-1` rule.
- The file kept exactly one `eDP-1` rule.
- The script did not change the external rule, including its scale, mode, and `auto-right` position.
- The installed recovery script kept the requested scale after each change.
- The scaling script also changed `omarchy_gdk_scale` from `2` to `1` for the `1.25` request.

With the installed scaling script, the same test did not change the saved scale of `1.6`.
Recovery then changed the active scale from `1.25` to `1.6`.

These results cover only the rules shown above. I did not test multiline rules, physical dock or lid events, or the full repository test suite.

The live reproduction and configuration workaround results are in [this issue comment](https://github.com/omacom/omarchy/issues/8103#issuecomment-5643235114).

Codex ran the local tests and helped prepare this comment.

Chessing234:
folded into #12263 for easier review.

PR #8149, opened 2026-08-25
Skip Claude and Pi on CPUs without SSE4.2/POPCNT

### LINKED ISSUE #7881: Skip incompatible Bun-based agents on CPUs below their instruction-set requirements
### System details

- Omarchy: 4.0.0-1
- Hardware: Apple iMac10,1
- CPU: Intel Core 2 Duo E7600 (Penryn, x86-64-v1)
- CPU features: SSE4.1; no SSE4.2, POPCNT, AVX, or AVX2
- Kernel: 7.1.8-arch1-3
- mise: 2026.8.6-1

### What's wrong?

Omarchy's default user installation script installs Claude and Pi unconditionally:

`/usr/share/omarchy/install/user/mise.sh`

```bash
omarchy-mise-install claude
# ...
omarchy-mise-install pi
```

The current native executables for both tools are Bun-based and require CPU
instructions unavailable on older x86-64 hardware.

On this machine, both agents install successfully but crash with SIGILL when
launched.

Pi 0.84.2:

```text
Process: /home/todor/.local/share/mise/installs/pi/0.84.2/pi/pi
Signal: SIGILL (ILL_ILLOPN)
Kernel: trap invalid opcode at pi+0x37fc488
```

Claude 2.1.241 crashed repeatedly at the same address:

```text
Process: /home/todor/.local/share/mise/installs/claude/2.1.241/claude
Signal: SIGILL
Kernel: trap invalid opcode at claude+0x198e3b3
```

There was no memory pressure or OOM event. Approximately 9.2 GiB was available
and swap was unused.

This CPU has SSE4.1 but lacks SSE4.2, POPCNT, AVX, and AVX2. Existing upstream
reports confirm that current Bun-compiled Claude binaries fail on CPUs below
their instruction baseline. Pi's release build script likewise compiles the
standalone executable using Bun's regular `bun-linux-x64` target.

### Steps to reproduce

1. Install Omarchy 4.0.0 on an x86-64-v1 CPU lacking SSE4.2/POPCNT/AVX.
2. Allow the default user installation to install Claude and Pi through mise.
3. Run `claude` or `pi`.
4. Observe an immediate or early SIGILL/core dump.

### Expected behavior

Omarchy should detect CPU capabilities before installing known incompatible
native agents.

Possible behavior:

- Skip incompatible tools and print a clear explanation.
- Offer a compatible Node-based installation where one exists.
- At minimum, warn that the installed executable cannot run on this CPU.

A feature-based check is preferable to checking CPU model names. The required
feature set should ideally be recorded per affected tool because upstream binary
requirements may differ or change.

### Actual behavior

The agents install normally and only reveal the incompatibility by repeatedly
crashing and generating systemd core dumps.

### Diagnostic summary

`omarchy debug --no-sudo --print` identifies the CPU as:

```text
Intel Core2 Duo E7600
Architecture level: x86-64-v1
Flags: sse sse2 sse3 ssse3 sse4_1
```

The journal contains repeated Claude SIGILL dumps and one Pi SIGILL dump.

Filed by GPT-5 via Codex.

Chessing234:
Default user setup always installed Claude and Pi through mise. On x86-64-v1 CPUs without SSE4.2/POPCNT (Bun's linux-x64 baseline) both natives install then SIGILL.

`omarchy-hw-cpu-sse4_2` detects that baseline before install. Missing either flag skips and prints why; ARM/aarch64 (including Apple Silicon/Asahi) still installs. Choosing Claude/Pi as the default agent is refused on incompatible CPUs, and those menu rows are hidden.

`bun-agent-cpu-test.sh` stubs cpuinfo: skip without sse4_2, install with flags, install on aarch64.

Fixes #7881


Chessing234:
folded into #11514 with the other mise/default-agent fixes.

PR #8150, opened 2026-08-25
Re-arm the sleep delay inhibitor only after resume

### LINKED ISSUE #8112: omarchy-sleep-lock.service fails intermittently on repeated suspend/resume cycles ("inhibition already running")
`omarchy-sleep-lock.service` fails intermittently on repeated suspend/resume cycles ("inhibition already running")

On repeated suspend/resume cycles in a short window (e.g. closing/opening a laptop lid several times), `omarchy-sleep-lock.service` reliably fails with:
```
systemd-inhibit: Failed to inhibit: The operation inhibition has been requested for is already running
```

Digging into it, the mechanism looks like this: `omarchy-system-sleep-monitor` acquires one delay-inhibitor, waits for a single `PrepareForSleep(true)` dbus signal, calls `omarchy-system-sleep-lock`, then exits. It relies on `Restart=always`/`RestartSec=2` to respawn and re-acquire a fresh inhibitor before the next suspend. But real-clock `RestartSec` freezes during suspend and fires right on resume — a fast second suspend can beat the respawn, so that cycle proceeds uninhibited, and the eventually-restarted instance's `systemd-inhibit` call lands after logind has already moved past accepting a new inhibitor for that transition. Widening the restart delay (tried 0.5s and 2s) didn't fix it — it looks like a structural issue with the respawn-per-cycle design racing suspend duration itself, not a timing constant to tune.

Repro: close/reopen a laptop lid (or otherwise trigger `systemd-sleep`) several times in quick succession, then check `journalctl --user -u omarchy-sleep-lock.service` for `status=1/FAILURE` restarts.

Chessing234:
`omarchy-system-sleep-monitor` took a delay inhibitor, locked on one `PrepareForSleep(true)`, and exited. Re-arm relied on `Restart=always` / `RestartSec=2`. That timer freezes during suspend; a fast second suspend beats the respawn, and the restarted `systemd-inhibit` hits "inhibition already running".

The monitor now loops in-process: take a delay inhibitor, lock on true, return so systemd-inhibit exits (suspend can proceed), then wait for `PrepareForSleep(false)` before taking another. Do not re-take while logind is still in a sleep transition. `Restart=always` stays as a crash net only.

`sleep-monitor-test.sh` still asserts lock + inhibitor release, and that a second inhibit is not attempted until a false event.

Fixes #8112

### DIFF
--- a/bin/omarchy-system-sleep-monitor
+++ b/bin/omarchy-system-sleep-monitor
@@ -4,20 +4,24 @@
 # omarchy:group=system
 # omarchy:hidden=true
 
+sleep_monitor="$OMARCHY_PATH/bin/omarchy-system-sleep-monitor"
+
 consume_sleep_events() {
-  local line sleep_lock
+  local line wanted=${1:-true} sleep_lock
   sleep_lock="$OMARCHY_PATH/bin/omarchy-system-sleep-lock"
 
   while IFS= read -r line; do
-    if [[ $line == *"boolean true"* ]]; then
-      "$sleep_lock"
+    if [[ $line == *"boolean ${wanted}"* ]]; then
+      if [[ $wanted == "true" ]]; then
+        "$sleep_lock"
+      fi
       return 0
     fi
   done
 }
 
 monitor_sleep_events() {
-  local monitor_fd monitor_pid status
+  local wanted=${1:-true} monitor_fd monitor_pid status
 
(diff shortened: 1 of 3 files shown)


Chessing234:

Chessing234:
folded into #12102 — same sleep-monitor script, one review is enough.

PR #8155, opened 2026-08-25
Add omarchy theme next/prev to cycle installed themes

Chessing234:
`omarchy theme bg next` already cycles wallpapers. There was no sibling for themes.

`omarchy theme next` / `omarchy theme prev` walk the same installed-theme union as `omarchy theme list` (stock `$OMARCHY_PATH/themes` plus `~/.config/omarchy/themes`), wrap both ways, and apply through `omarchy-theme-set`. Display names stay on the existing mapping (`tokyo-night` ↔ `Tokyo Night`). One installed theme is a no-op, not an error.

Style menu gets Next Theme / Previous Theme (`style.theme-next`, `style.theme-prev`) with no aliases. No new binding: every SPACE combo in that family is taken.

`theme-next-test.sh` covers wrap both ways and the single-theme no-op.

### DIFF
--- a/bin/omarchy-theme-next
+++ b/bin/omarchy-theme-next
@@ -1,0 +1,56 @@
+#!/bin/bash
+
+# omarchy:summary=Cycle to the next installed theme
+# omarchy:examples=omarchy theme next
+
+# Display name mapping matches omarchy-theme-list / omarchy-theme-current:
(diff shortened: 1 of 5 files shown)


Chessing234:

Chessing234:
folding into #12247 with the other theme cycling fixes

PR #8157, opened 2026-08-25
Fit the fastfetch logo to the terminal

### LINKED ISSUE #7928: The Omarchy logo in fastfetch is huge and breaking the texts in the terminal when resizing terminal.
### System details

Omarchy 4.0.0-1

### What's wrong?

Image

Image

Chessing234:
File logos are 1:1 characters and ignore width, so the 26-row About icon (plus padding) was taller and wider than a typical terminal. On resize the modules collided with it.

fastfetch now runs `omarchy-fastfetch-logo`, which picks the largest green Omarchy wordmark that still leaves room for the 54-column module box: full wordmark, half-width wordmark, or compact O. About still draws `~/.config/omarchy/branding/about.txt` as a file logo so the sheen can find those cells.

Fixes #7928


Chessing234:

Chessing234:
folded into #12268 for easier review.

PR #8160, opened 2026-08-25
Make tray icons click targets so the bar dispatches to them

### LINKED ISSUE #8111: Tray icons are unclickable: clicks fall through to the bar's gesture area and toggle transparency
> **Note:** This bug was investigated and this report written by [Claude Code](https://claude.com/claude-code), and submitted on my behalf. The source tracing, the D-Bus checks and the two-app reproduction below were all carried out by it; I have reviewed and endorsed the result.

## Summary

System tray (`omarchy.tray`) icons cannot be clicked. Left-clicking a tray icon
does nothing; the click falls through to a bar-wide gesture area, so
double-clicking a tray icon **toggles the whole bar's background transparency**
instead of activating the icon (and rewrites `bar.transparent` in
`~/.config/omarchy/shell.json`).

Omarchy's own bar widgets (bluetooth, wifi, audio, power, clock) are unaffected.
Only third-party StatusNotifierItem icons are.

## Steps to reproduce

1. Run any app that provides a tray icon. `udiskie` works and is a clean test:
   `udiskie --tray`
2. Pin it so it is visible without the drawer — add to the `omarchy.tray` entry
   in `~/.config/omarchy/shell.json`:
   `{ "id": "omarchy.tray", "pinned": ["udiskie"] }`
3. Hover the icon — the tooltip appears, so pointer events do reach it.
4. Left-click it. Nothing happens.
5. Double-click it. The entire bar's background transparency toggles.

Reproduced with two unrelated apps (udiskie, and Claude Desktop's
`Claude_status_icon_1`). The item is healthy on D-Bus in both cases:
`org.kde.StatusNotifierItem.Activate` succeeds when called directly with
`gdbus`, and `com.canonical.dbusmenu.GetLayout` returns the full menu.

## Expected

Left-click activates the tray item; right-click opens its menu.

## Actual

Neither. Hover/tooltip works, all clicks are swallowed, and double-click hits
the bar's transparency gesture.

## Root cause

1. `Ui/WidgetButton.qml:47` is the only thing that calls `registerClickTarget`.
   `plugins/bar/widgets/Tray.qml` never does — `TrayItem` is a bare `MouseArea`
   (Tray.qml:814). `Ui/BarIconButton.qml:5` *is* a `WidgetButton`, which is why
   Omarchy's own widgets work.
2. A left-press on the slot is taken by `ModuleSlot`'s drag `MouseArea`
   (`Bar.qml:1654`, `acceptedButtons: Qt.LeftButton`, fills the slot).
3. On click it calls `pressModuleClickTarget` (`Bar.qml:784`) ->
   `moduleClickTargetAt` (`Bar.qml:762`) finds no registered target and falls
   back to `slot.activeItem`, which has no `triggerPress`, so
   `moduleTargetClickable` (`Bar.qml:752`) rejects it and it returns `null`.
4. `pressModuleClickTarget` returns false, so `Bar.qml:1740` sets
   `mouse.accepted = false`.
5. The click propagates to `CenterGestureArea`. Despite the name it is
   `anchors.fill: parent` (`Bar.qml:1306`) inside
   `CenterModules { anchors.fill: parent }` (`Bar.qml:1110`) — it spans the
   **entire bar**, not the centre section.
6. `CenterGestureArea.onDoubleClicked` calls `toggleTransparency()`
   (`Bar.qml:1464`). Hence the transparency toggle.

Hover still works because hover is not routed through the click-target registry.

## Suggested fix

Either give `TrayItem` a `triggerPress(button)` and register it via
`registerClickTarget`, so tray icons participate in the same dispatch as every
other widget; or stop `pressModuleClickTarget` returning false from swallowing
clicks for widgets that handle their own input.

Separately, `CenterGestureArea` spanning the full bar looks unintended — a
click anywhere on the bar that no widget claims can toggle transparency.

## Secondary observation

`dragThreshold` is `Style.space(4)` (`Bar.qml:1658`) = 4 logical px, compared as
*Manhattan* distance `|dx| + |dy|` (`Bar.qml:1681`). On a 26px-tall bar this is
very easy to trip, so clicks also frequently become widget-reorder drags
(`omarchy-bar-drag-ghost` appears in `hyprctl layers`). Not the root cause, but
it compounds it.

## System

- Omarchy 4.0.0-1
- Hyprland 0.56.2, quickshell-git 0.3.0.r20.g28771c7-2
- Kernel 7.1.8-arch1-3, ASUS Vivobook S 15 S5506MA
- Display 2880x1620 @ scale 1.6667 (bar 1728x26 logical)

---

samirpokharel:
Fixes #8111.

The bar does not leave clicks to a widget's own `MouseArea`. A slot-wide area takes the press so a drag can reorder the bar, then hands the click to whichever registered target sits under the cursor (`Bar.qml` → `pressModuleClickTarget` → `moduleClickTargetAt`). Selection requires a `triggerPress` function on the target, and `WidgetButton` is where every other widget picks that up for free.

`TrayItem` is not a `WidgetButton` and registered nothing, so a left click on a tray icon found no target, `pressModuleClickTarget` returned false, and `Bar.qml` set `mouse.accepted = false`. The unclaimed click fell through to `CenterGestureArea` — which despite the name is `anchors.fill: parent` across the whole bar — so a double click on a tray icon toggled the bar's transparency and rewrote `bar.transparent` in `shell.json` instead of activating the icon.

There is a third symptom the report does not mention: `Bar.qml:1667` derives `cursorShape` from the same lookup, so tray icons also showed an arrow rather than a pointing hand.

This gives `TrayItem` the `triggerPress` and registry wiring `WidgetButton` provides, so tray icons take part in the same dispatch as every other widget. Left click activates, middle click secondary-activates, and a right click or a menu-only item opens the menu. The slot hands over a button but no position, so menus anchor to the icon rather than to wherever the pointer landed.

Drawer icons opt in with `inDrawer: true` and stay out of the registry until the drawer is revealed: while collapsed the row is parked at `x: drawerExtent` — outside its own clip rectangle — but still holds geometry there, so registering it unconditionally would let a click on the pinned row map onto a hidden icon.

Right and middle clicks keep working exactly as before and cannot double-fire: both slot `MouseArea`s are `acceptedButtons: Qt.LeftButton`, so those buttons still reach the tray's own `MouseArea`, and on a left click that `MouseArea` never sees the press.

`test/shell.d/tray-click-target-test.sh` pins the contract across all three files — that `Bar.qml` still selects targets by `triggerPress` and refuses unclaimed clicks, that `WidgetButton` meets it, and that tray icons now do too. The Bar and WidgetButton assertions pass against the current tree, so the test is not just asserting the new code back to itself; the tray assertions fail on it (`not ok - tray icons expose triggerPress`) and pass with this change. `tray-menu-test.sh` still passes, so the existing DBusMenu path is unchanged.

Not verified in a live session: I do not have a tray app pinned on a bar built from this checkout. The reproduction in the issue (`udiskie --tray`, pin it, left click) is the check worth running.

### DIFF
--- a/shell/plugins/bar/widgets/Tray.qml
+++ b/shell/plugins/bar/widgets/Tray.qml
@@ -292,7 +292,7 @@ BarWidget {
 
             Repeater {
               model: root.drawerItems
-              TrayItem {}
+              TrayItem { inDrawer: true }
             }
           }
         }
@@ -375,7 +375,7 @@ BarWidget {
 
             Repeater {
               model: root.drawerItems
-              TrayItem {}
+              TrayItem { inDrawer: true }
             }
           }
         }
@@ -795,6 +795,13 @@ BarWidget {
     id: trayItemRoot
 
     required property var modelData
+    // Drawer icons live in a clipped strip that is parked outside its own clip
+    // rectangle while collapsed. They still occupy geometry there, so they stay
+    // out of the click registry until the drawer is actually revealed.
+    property bool inDrawer: false
+    property var registeredBar: null
+
+    readonly property bool interactive: !inDrawer || root.revealProgress > 0
 
     visible: modelData.status !== Status.Passive
     implicitWidth: visible ? root.trayItemExtent : 0
@@ -802,6 +809,39 @@ BarWidget {
 
     function displayMenu(mouse) {
       root.openTrayMenu(trayItemRoot.modelData, trayItemRoot, mouse)
+    }
+
+    // The bar routes clicks through its own registry rather than leaving them to
+    // each widget's MouseArea: a slot-wide area takes the press so a drag can
+    // reorder the bar, then hands the click to whichever registered target sits
+    // under the cursor. A tray icon that registers nothing is never a target, so
+    // the click is refused and falls through to the bar's gesture area, where a
+    // double click toggles the bar's transparency instead of activating the
+    // icon. Registering is what every other widget inherits from WidgetButton.
+    function triggerPress(button) {
+      if (root.bar) root.bar.hideTooltip(trayItemRoot)
+
+      // The slot hands over a button but no position, so anchor menus to the
+      // icon itself rather than to wherever the pointer happened to land.
+      var center = { x: trayItemRoot.width / 2, y: trayItemRoot.height / 2 }
+
+      if (button === Qt.RightButton || trayItemRoot.modelData.onlyMenu) trayItemRoot.displayMenu(center)
+      else if (button === Qt.MiddleButton) trayItemRoot.modelData.secondaryActivate()
+      else trayItemRoot.modelData.activate()
+    }
+
+    function syncClickRegistration() {
+      if (registeredBar && registeredBar.unregisterClickTarget) registeredBar.unregisterClickTarget(trayItemRoot)
+      registeredBar = root.bar
+      if (registeredBar && registeredBar.registerClickTarget) registeredBar.registerClickTarget(trayItemRoot)
+    }
+
+    Component.onCompleted: syncClickRegistration()
+    Component.onDestruction: if (registeredBar && registeredBar.unregisterClickTarget) registeredBar.unregisterClickTarget(trayItemRoot)
+
+    Connections {
+      target: root
+      function onBarChanged() { trayItemRoot.syncClickRegistration() }
     }
 
     TrayIcon {
--- a/test/shell.d/tray-click-target-test.sh
+++ b/test/shell.d/tray-click-target-test.sh
@@ -1,0 +1,61 @@
+#!/bin/bash
+
+set -euo pipefail
+
+source "$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd)/base-test.sh"
+
+bar="$ROOT/shell/plugins/bar/Bar.qml"
+tray="$ROOT/shell/plugins/bar/widgets/Tray.qml"
+widget_button="$ROOT/shell/Ui/WidgetButton.qml"
+
+# The bar does not let a widget's own MouseArea see clicks: a slot-wide area
+# takes the press so a drag can reorder the bar, then hands the click to
+# whichever registered target sits under the cursor. These two lines are the
+# contract every clickable bar surface has to satisfy — pin them, so a widget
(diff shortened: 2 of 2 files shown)


mtxmiller:
Tested this patch on omarchy 4.0.2-1 / quickshell 0.3.1-1. It fixes left-click,
but the `Qt.RightButton` and `Qt.MiddleButton` branches in `triggerPress` are
unreachable, so tray context menus stay broken.

The bar's slot MouseArea that dispatches to click targets is left-button only:

```qml
// shell/plugins/bar/Bar.qml:1680
MouseArea {
  id: modulePointer
  anchors.fill: parent
  acceptedButtons: Qt.LeftButton
```

Its `onClicked` is the only caller of `pressModuleClickTarget()`, so
`triggerPress(button)` is only ever invoked with `Qt.LeftButton`. Right- and
middle-press are also not passed through to the widget's own `MouseArea` —
I expected them to fall through, and they don't.

Verified by instrumenting `TrayItem` with `console.log` in both
`triggerPress()` and the existing `mouseArea.onPressed`, then clicking a
pinned icon repeatedly:

```
6x   TRAYDBG triggerPress button=1        (left-click)
0x   TRAYDBG mouseArea.onPressed          (right-click - nothing, any handler)
0x   TRAYDBG openTrayMenu
0x   TRAYDBG trayMenuOpen
```

Right-click produces no log line anywhere, so the press is consumed before it
reaches either path. Registration itself is fine — all three
StatusNotifierItems register with the bar and `registerClickTarget` resolves.

So as it stands this closes the left-click half of #8111 and leaves the
right-click menu dead. Making `modulePointer` accept right/middle and forward
the button through `pressModuleClickTarget()` looks like the missing piece,
though I haven't tested whether that interferes with the drag-to-reorder
gesture that area also owns.

Workaround locally, in case it's useful to anyone: route the one button that
does arrive to the menu instead of `activate()`. Every tray menu already
carries its own show-window entry (Show App / Open Main Window / Library), so
nothing is lost and Quit becomes reachable.

```qml
function triggerPress(button) {
  if (root.bar) root.bar.hideTooltip(trayItemRoot)
  var center = { x: trayItemRoot.width / 2, y: trayItemRoot.height / 2 }
  trayItemRoot.displayMenu(center)
}
```

PR #8169, opened 2026-08-25
Stop apply-system reruns leaving the install log world-writable

Adolanium:
`install/helpers/logging.sh` starts the install log like this:

```bash
touch "$OMARCHY_INSTALL_LOG_FILE"
chmod 666 "$OMARCHY_INSTALL_LOG_FILE" 2>/dev/null || true
```

Scope first, because it is narrower than I first wrote. On a fresh install this code never runs: the ISO orchestrator (`phases_impl.py`) runs `omarchy-apply-system` and `omarchy-provision-user` with `OMARCHY_LOG_TO_STDOUT=1`, and in that mode `start_install_log` skips the `touch` and `chmod`. The ISO then copies the log into the target itself and leaves it 0644 root:root. That mode comes from the ISO and is out of scope here. `omarchy-provision-owner` also runs first-boot finalization with `OMARCHY_LOG_TO_STDOUT=1`.

The `chmod 666` fires when someone reruns `omarchy-apply-system` or `omarchy-apply-hardware` by hand on an installed system, which the scripts allow for diagnostics. From that point `/var/log/omarchy-install.log` is world-readable and world-writable. Reading it gives every local account the output of every root-side setup step. Writing it lets anyone append forged `Completed:` or `Failed:` lines or truncate the file. This is also the file `omarchy upload` sends in when someone reports a broken install.

What changed:

- The log is published as a fresh file, not chmodded in place. `start_install_log` builds it under `umask 077`, copies any existing content across, sets it to 0640 root:wheel, and moves it over the old name. A chmod alone only changes who can open the file from then on; a process that already had the old 0666 file open would keep writing. With the move, that descriptor points at an unlinked inode.
- Mode 0640 with group wheel rather than 0600. The owner account is in wheel and already has full sudo, so wheel read gives up nothing, and `omarchy-upload-log install` keeps reading the log as that user without a new privileged helper. Other local accounts are shut out. 0600 would have made the upload go out with an empty log.
- No second writer. I checked every caller of `start_install_log` and every reader of `OMARCHY_INSTALL_LOG_FILE` in `bin/` and `install/`: only root writes this file through `logging.sh`. `omarchy-provision-user` only sources `logging.sh` when the ISO or `omarchy-provision-owner` hands it a log path, and both of those set `OMARCHY_LOG_TO_STDOUT=1`.

Not in this PR: cleaning up a 0666 log on systems where nobody reruns an apply command. That needs a package-upgrade migration, which #9465 covers on top of this base.

Tests: `test/shell.d/logging-test.sh` gains a source check that `chmod 666` is gone and `chmod 640` is present, a run that proves logging still works after the change, and, where the filesystem reports modes faithfully, checks that a new log is 0640, that an old 0666 log is replaced by a new inode with its content kept, and that a descriptor opened on the old file cannot write into the new one. All pass. The runtime cases fail against the base `logging.sh` and against a chmod-in-place version.

### DIFF
--- a/install/helpers/logging.sh
+++ b/install/helpers/logging.sh
@@ -13,8 +13,33 @@ omarchy_log_line() {
 start_install_log() {
   if ! omarchy_log_to_stdout; then
     mkdir -p "$(dirname "$OMARCHY_INSTALL_LOG_FILE")"
-    touch "$OMARCHY_INSTALL_LOG_FILE"
-    chmod 666 "$OMARCHY_INSTALL_LOG_FILE" 2>/dev/null || true
+
+    # Only root writes here through this helper: the ISO runs the target setup
+    # with OMARCHY_LOG_TO_STDOUT=1 and copies the log itself, so this branch is
+    # reached when omarchy-apply-system or omarchy-apply-hardware is rerun by
+    # hand on an installed system. The old world-writable mode then let every
+    # local account read the step output and append forged lines.
+    #
+    # The log is published as a fresh file, not chmodded in place. A chmod only
+    # changes who can open the file from now on; a process that already had the
+    # old 0666 file open keeps writing to it. Building the new file privately
+    # and moving it over the old name leaves such a descriptor pointing at an
+    # unlinked inode. Existing content is carried across.
+    #
+    # Mode 0640 with group wheel: the owner account is in wheel and already has
+    # full sudo, so wheel read gives up nothing, and omarchy-upload-log can
+    # still read the log as that user. Other local accounts are shut out.
+    local old_umask staged
+    old_umask=$(umask)
+    umask 077
+    staged=$(mktemp "$OMARCHY_INSTALL_LOG_FILE.XXXXXX")
+    umask "$old_umask"
+    if [[ -f $OMARCHY_INSTALL_LOG_FILE ]]; then
+      cat "$OMARCHY_INSTALL_LOG_FILE" >>"$staged" 2>/dev/null || true
+    fi
+    chgrp wheel "$staged" 2>/dev/null || true
+    chmod 640 "$staged"
+    mv -f "$staged" "$OMARCHY_INSTALL_LOG_FILE"
   fi
 
   export OMARCHY_START_TIME="${OMARCHY_START_TIME:-$(date '+%Y-%m-%d %H:%M:%S')}"
--- a/test/shell.d/logging-test.sh
+++ b/test/shell.d/logging-test.sh
@@ -49,3 +49,80 @@ grep -q "about to fail" "$stdout_log" || fail "stdout logging mode emits script 
 grep -q "Failed: $failing_script (exit code: 1)" "$stdout_log" || fail "stdout logging mode emits failure marker"
 
 pass "run_logged records failures under errexit"
+
+
+# --- start_install_log permissions ---------------------------------------------
+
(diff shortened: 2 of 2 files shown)


Adolanium:
Both review points are in the follow-up commit.

The log is now restricted before any lookup happens: a new file is created under umask 077, an existing file (including a 0666 one left by an older install) goes 0600 right away, and only then is the install user's group looked up and the file widened to 0660.

The test now exercises the group-sharing path with the account running the test instead of root, so it works for an ordinary CI user. A new case covers the old-0666-log-gets-restricted ordering. All cases pass, and the failing-file list for ./test/shell is unchanged from the base commit.

Adolanium:
Another pass after checking who actually writes this file.

The ISO runs `omarchy-apply-system` and `omarchy-provision-user` with `OMARCHY_LOG_TO_STDOUT=1`, so `start_install_log` never touches the file on a fresh install. The `chmod 666` only fires when `omarchy-apply-system` or `omarchy-apply-hardware` is rerun by hand on an installed system. I have narrowed the title and description to that.

The 0660 group share is gone. Every writer through `logging.sh` in current flows is root, so there was no second writer for it to serve. The file is now plain 0600, which also drops the `id -gn` lookup, the `chgrp`, and the stale-group edge case. The test is simpler to match.

I also dropped the "full name, email" claim from the description. Those only reach the log under `OMARCHY_INSTALL_DEBUG=1`.

AFOliveira:
I revalidated head 1ef6da01 locally. The 0600 change closes new opens, but two release behaviors still need a follow-up:

1. `start_install_log` keeps the legacy inode. A second UID that opened the old 0666 file before this code ran retained that descriptor and appended after the file became 0600. Isolated user/mount-namespace result:

```
CONFIRMED pr8169 inode=3->3 mode=600 stale_descriptor_write=present
```

Publishing a fresh 0600 sibling inode (and migrating an existing legacy log on package upgrade) detaches those descriptors. Merely chmodding cannot.

2. `omarchy-upload-log install` currently reads `/var/log/omarchy-install.log` directly as the desktop user. After this PR makes it root:root 0600, `[[ -s $LIVE_OMARCHY_LOG ]]` is false for that user and the command omits the Omarchy log. The safe compatibility path is a private 0700 workspace plus a fixed `sudo -N -- /usr/bin/cat -- /var/log/omarchy-install.log`, with the timestamp cold before extensible collectors run and revoked on exit.

The 0600 core here is still the right narrow base. I am reducing #9465 to only these proven additive invariants instead of competing with or duplicating this PR.

Adolanium:
Thanks, both points hold, and both land in `start_install_log`, so I took them here.

The log is now published as a fresh file: built under `umask 077`, old content copied across, then moved over the old name. A descriptor opened on the old 0666 file is left pointing at an unlinked inode. The test opens such a descriptor before the rerun and checks its write does not reach the new file.

For the reader: 0640 root:wheel instead of 0600. Every Omarchy owner is in wheel and already has full sudo, so wheel read gives up nothing, and `omarchy-upload-log install` keeps working with no new privileged helper. One detail on the failure mode: `[[ -s ]]` only stats, so it stays true on an unreadable file. The header is written and `cat` fails, but the effect is the same, the upload goes out without the log.

The package-upgrade migration for systems where nobody reruns an apply command is still yours in #9465. This PR stays the base.

PR #8170, opened 2026-08-25
Refuse hook and state names that are paths

Adolanium:
`omarchy-hook` and `omarchy-state set` join a name straight into a path without looking at it:

```bash
HOOK_PATH="$HOME/.config/omarchy/hooks/$1"
...
bash "$HOOK_PATH" "$@"
```

```bash
set) touch "$STATE_DIR/$STATE_NAME" ;;
```

On the current code, `omarchy-hook ../../evil` runs `~/.config/evil`, and `omarchy-state set ../../escape` creates `~/.local/escape`. I checked that both happen before this change.

This is a robustness fix, not a security one. Every caller in the repo passes a fixed label: `omarchy-update` runs `omarchy-hook post-update`, `omarchy-theme-set` runs `omarchy-hook theme-set "$THEME_NAME"` (the theme name is the argument, not the hook name), `omarchy-channel-set` and the migrations run `omarchy-state set reboot-required`. No user or network input reaches either command. The guard is there so a future caller that builds a name from something else cannot quietly write or run outside the directory.

What changed:

- Both commands refuse an empty name, a name with a `/` in it, or a name that is exactly `.` or `..`. They exit 2 and print why on stderr.
- Nothing else is refused. Once slashes are out, only a whole name of `.` or `..` can leave the directory, so names like `a..b` or `v1..2` stay valid. This matches how `omarchy-webapp-install` and `omarchy-theme-set` guard their names.
- `omarchy-state clear` is unchanged on purpose. It matches basenames through `find -name`, so a pattern can never walk out of the directory.

Tests: new `test/shell.d/hook-state-name-guard-test.sh`, 13 cases. A plain hook still runs, a plain state file is still created, `a..b` and `v1..2` are accepted, `.` and `..` and `../x` and `sub/dir` are refused with nothing created or run outside the directory, and `clear` still honors glob patterns. Against the code before this change the test stops at the first refusal case. The other hook and state tests (`monitor-state-test.sh`, `nvidia-kms-hook-test.sh`) pass unchanged.

### DIFF
--- a/bin/omarchy-hook
+++ b/bin/omarchy-hook
@@ -11,6 +11,18 @@ if (( $# < 1 )); then
 fi
 
 HOOK=$1
+
+# Hook names are fixed labels chosen by Omarchy code (post-update, theme-set,
+# font-set). The name becomes a filename under the hooks directory. A slash
+# would turn it into directory levels, and a bare `.` or `..` would point bash
+# at the directory itself or its parent. Refuse those rather than follow them.
+# Dots inside a name (a..b) are fine; once slashes are out, only the whole
+# name being `.` or `..` can leave the directory.
+if [[ -z $HOOK || $HOOK == */* || $HOOK == "." || $HOOK == ".." ]]; then
+  echo "Invalid hook name: $HOOK" >&2
+  exit 2
+fi
+
 HOOK_PATH="$HOME/.config/omarchy/hooks/$1"
 HOOK_DIR="$HOOK_PATH.d"
 shift
--- a/bin/omarchy-hook-install
+++ b/bin/omarchy-hook-install
@@ -15,6 +15,17 @@ fi
 
 HOOK_TYPE=$1
 HOOK_FILE=$2
+
+# Hook types are the same labels omarchy-hook runs (post-update, theme-set).
+# The type becomes a directory under the hooks directory. A slash would turn
+# it into directory levels. A bare `.` or `..` is a name the runner already
+# refuses, so installing under it would write a hook nothing can run. Refuse
+# those rather than mkdir/cp into them. Dots inside a name (a..b) are fine.
+if [[ -z $HOOK_TYPE || $HOOK_TYPE == */* || $HOOK_TYPE == "." || $HOOK_TYPE == ".." ]]; then
+  echo "Invalid hook name: $HOOK_TYPE" >&2
+  exit 2
+fi
+
 HOOK_DIR="$HOME/.config/omarchy/hooks/$HOOK_TYPE.d"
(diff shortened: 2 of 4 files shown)


Adolanium:
Reworked this after a second look.

- The check no longer refuses any `..` in a name, which was too broad (it blocked harmless names like `a..b`). It now refuses only an empty name, a name with a `/`, or a name that is exactly `.` or `..`. Once slashes are out, those are the only names that can leave the directory. Same shape as the guard in `omarchy-webapp-install` and `omarchy-theme-set`.
- Reframed the title, body, and comments. Every caller passes a fixed label, so this is a footgun guard for future callers, not a security fix.
- Tests grew from 7 to 13 cases: `a..b` and `v1..2` accepted, `.` and `..` refused, `../x` and `sub/dir` still refused.

ErikMelton:
3 improvements;

Migration can overwrite administrator-authored PAM policy; treats any package-unowned file containing pam_fprintd, the  gate, or the FIDO authfile as Omarchy-owned. Custom PAM files commonly satisfy that predicate. The rebuild then discards comments and all custom account, password, and session directives. Match exact known legacy layouts or replace only the known pam_unix lines while preserving everything else.

Post-removal installations remain vulnerable. The same predicate requires a hardware-auth marker, but both removal commands delete those markers while leaving the bare four-line pam_unix override behind. Users who removed fingerprint/FIDO2 before upgrading are therefore skipped permanently. Recognize the exact markerless legacy layout too.

Failed repairs are marked complete. migrations/1788256455.sh:49 (https://github.com/omacom/omarchy/blob/8a13eac872988c722be4cb4765211593b61d6de9/migrations/1788256455.sh#L49) prints a retry instruction when backup/elevation fails but returns success. Likewise, a verification failure restores the vulnerable backup and returns success. omarchy-migrate then creates the completion marker, so it never retries. Both paths must exit nonzero after any necessary restoration.

Adolanium:
I think this landed on the wrong PR. The migration, the PAM predicate, and the linked commit `8a13eac` are from #9873 by @Wheel-Smith. This PR only touches `omarchy-hook` and `omarchy-state`.

The three points look worth raising there, so you may want to move the comment over to #9873.

ErikMelton:
Last change I can see:

  - Guard omarchy-hook-install too! The new runner validation rejects hook names containing /, but omarchy hook install <type> <file> still accepts them. Apply the same validation before the installer’s mkdir/cp and cover it in the regression test.

Adolanium:
Done. omarchy-hook-install now refuses an empty name, a slash, or a bare . or .. before mkdir/cp, same as the runner. The regression test covers the installer too.

ErikMelton (approved):

PR #8171, opened 2026-08-25
Quote launcher arguments so they survive the launch-or-focus re-parse

Adolanium:
`omarchy-launch-or-focus-tui` and `omarchy-launch-or-focus-webapp` pack their arguments into one string with `$@`, and `omarchy-launch-or-focus` unpacks that string with `eval`:

```bash
# omarchy-launch-or-focus-tui
LAUNCH_COMMAND="omarchy-launch-tui $@"
exec omarchy-launch-or-focus "$APP_ID" "$LAUNCH_COMMAND"
```

```bash
# omarchy-launch-or-focus
eval exec setsid $LAUNCH_COMMAND
```

The pack step drops the quoting, so the eval step re-splits on spaces and re-reads shell characters. `omarchy-launch-or-focus-tui mytui "two words"` hands the TUI three arguments instead of two. A web app URL with a query string fares worse: `omarchy-launch-or-focus-webapp Chat 'https://chat.example/?a=1&b=2'` puts `&` through eval, which reads it as a command separator, and the browser gets the URL cut off at the first `&`.

What changed:

- Both wrappers quote each argument with `printf %q` before joining, so the eval in `omarchy-launch-or-focus` puts back exactly the words it was given.
- `omarchy-launch-tui` quotes two spots that rode along unquoted: `basename "$1"` when deriving the app id, and `--app-id="$APP_ID"` on the xdg-terminal-exec line.
- The commented About example in `config/omarchy/extensions/omarchy-menu.jsonc` now passes `zsh -c '...'` as separate words instead of one quoted string, with a note on why.

Every shipped caller passes fixed strings (keybindings, the menu, `helpers.lua` and its `shell_quote`), and `%q` leaves a plain word untouched, so nothing they launch changes.

One behavior change for custom configs. Before, a multi-word command passed as one quoted string was re-split by the eval, so `{ tui = "btop --utf-force", focus = true }` in a user's `helpers.lua` binding reached the terminal as two words. Now it reaches the terminal as one argument, the same way the non-focus `omarchy-launch-tui` path always treated it. Shell expansion of `~` or `$HOME` inside a wrapper argument also stops, again matching the non-focus path. Anyone with a binding of that shape should split the words, which the updated example shows.

Tests: new `test/shell.d/launch-wrapper-argv-test.sh` drives the real wrappers with a stubbed `hyprctl` and a `setsid` that records its argv one line per argument. Five cases: a two-word argument and a string holding shell characters arrive whole, the app id flag survives the ride, a URL with a query string arrives uncut, and launch-tui quotes its app id and a command path with spaces. They fail against the old code, where the two-word argument split and the URL stopped at the `&`. All 5 pass, along with `default-agent-test.sh` and `launcher-remove-test.sh`, which also go through these wrappers.

### DIFF
--- a/bin/omarchy-launch-or-focus-tui
+++ b/bin/omarchy-launch-or-focus-tui
@@ -9,6 +9,10 @@ else
   APP_ID="org.omarchy.$(basename "$1")"
 fi
 
-LAUNCH_COMMAND="omarchy-launch-tui $@"
+# The command crosses omarchy-launch-or-focus as one string and is re-parsed
+# there with eval, so each word is quoted first. An argument holding spaces or
+# shell characters would otherwise arrive as extra words or run as code.
+printf -v quoted '%q ' "$@"
+LAUNCH_COMMAND="omarchy-launch-tui ${quoted% }"
 
 exec omarchy-launch-or-focus "$APP_ID" "$LAUNCH_COMMAND"
--- a/bin/omarchy-launch-or-focus-webapp
+++ b/bin/omarchy-launch-or-focus-webapp
@@ -10,6 +10,11 @@ fi
 
 WINDOW_PATTERN="$1"
 shift
-LAUNCH_COMMAND="omarchy-launch-webapp $@"
+
+# The command crosses omarchy-launch-or-focus as one string and is re-parsed
+# there with eval, so each word is quoted first. A URL holding & or an argument
+# holding spaces would otherwise split or run as shell syntax on the far side.
+printf -v quoted '%q ' "$@"
+LAUNCH_COMMAND="omarchy-launch-webapp ${quoted% }"
 
 exec omarchy-launch-or-focus "$WINDOW_PATTERN" "$LAUNCH_COMMAND"
--- a/bin/omarchy-launch-tui
+++ b/bin/omarchy-launch-tui
@@ -7,7 +7,7 @@ if [[ ${1:-} == --app-id=* ]]; then
   APP_ID="${1#--app-id=}"
   shift
 else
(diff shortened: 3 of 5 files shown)


Adolanium:
One more pass after checking custom-config shapes.

The description now calls out a behavior change I had missed. A multi-word command passed as one quoted string, like `{ tui = "btop --utf-force", focus = true }` in a `helpers.lua` binding, used to be re-split by the eval and now reaches the terminal as one argument. That matches what the non-focus `omarchy-launch-tui` path always did, but it is a change for anyone with a binding of that shape.

The commented About example in `config/omarchy/extensions/omarchy-menu.jsonc` had exactly that shape, so it now passes `zsh -c '...'` as separate words, with a note on why. Shipped bindings and menu entries all pass fixed single words and are unchanged.

bjarneo:
Automated duplication check: this pull request looks similar to #9490, which covers the same launch-or-focus arg quoting. I keep that one open and close this one to consolidate review.

Adolanium:
GitHub rejected reopening this pull request (`Could not open the pull request`). The updated branch is now https://github.com/omacom/omarchy/pull/12800.

#8171 was opened on 2026-08-25. #9490 was opened on 2026-08-31. Both quote arguments in `omarchy-launch-or-focus-tui` and `omarchy-launch-or-focus-webapp` before `omarchy-launch-or-focus` evals the command string. #12800 also quotes `basename` and `--app-id` in `omarchy-launch-tui`, updates the commented About example, and includes the empty-argument guard from #9490. `printf %q` with no arguments emits `''`.

Issue #7325, opened 2026-08-17
Omazed fails when installing Zed


Thijzert123:
### System details

Omarchy Quattro fresh install

### What's wrong?

When installing Zed via Menu > Install > Editor > Zed, `omazed` gets installed, but it shows this error when `omazed setup` gets executed:
```
==> Omazed installed successfully!

🚀 To complete setup, run:
   omazed setup

   This will:
   • Set up omarchy hook integration (or systemd fallback)
   • Set the Zed theme once on first run

Then change your Omarchy theme and watch Zed follow along!

:: Running post-transaction hooks...
(1/3) Arming ConditionNeedsUpdate...
(2/3) Updating icon theme caches...
(3/3) Updating the desktop file MIME type cache...
[INFO] Setting up Omazed for user: user

[INFO] Removed old omazed hook
[SUCCESS] ✓ Omarchy hook configured
[WARN] Could not determine current Omarchy theme

🌏 Done! Press any key to close...  
```

Running `omazed setup` again still fails with the same error.

marijn070:
reinstall omazed via the omarchy AUR install menu; the latest commit fixes this issue.
the issue was due to the current theme being defined at `~/.config/omarchy/current` in omarchy 3, while in omarchy 4 that is set under `~/.local/state/omarchy/current` in omarchy 4

Thijzert123:
That works, thanks!

PR #8182, opened 2026-08-25
Add native Zed theming

ahmedkall:
Zed currently gets theme support through omazed, an AUR package with its own hook and rendering pipeline. That puts a second integration next to the one Omarchy already uses for every other themed app, and it has already broken during the Quattro path move in #7325.

This replaces it with a `default/themed/zed.json.tpl` template and a small sync command wired into `omarchy-theme-set`. Installing Zed selects Omarchy once. Later theme changes refresh the file without overriding a manual Zed theme choice.

The migration drops omazed and its hooks, refreshes the current theme, and only switches users whose selected theme is still Omazed.

The Zed theme mapping is adapted from [APS6/omazed](https://github.com/APS6/omazed) (MIT).

Tested across every built-in theme, both appearances, Zed's string and object settings formats, and repeated migrations.

### DIFF
--- a/bin/omarchy-install-editor-zed
+++ b/bin/omarchy-install-editor-zed
@@ -3,9 +3,9 @@
 # omarchy:summary=Install Zed Editor and configure it with the current Omarchy theme
 
 echo "Installing Zed Editor..."
-omarchy-pkg-add zed omazed
+omarchy-pkg-add zed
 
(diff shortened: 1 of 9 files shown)


morgoth:
omazed was fixed for quatro already.
It was already proposed to have this built in into omarchy, but decision was to leave it as an omazed package.

PR #8191, opened 2026-08-25
Offer US International in the keyboard picker

esasse:
The keyboard picker has no entry with dead keys, so anyone arriving from Windows' US-International — the default for most of Latin America and a common choice across Europe — installs on plain `us` and then discovers there is no way to type an accent. The fix is to hand-write `kb_variant = "intl"` into `~/.config/hypr/input.lua`, which you only find by reading the commented-out block in that file or by asking in Discord.

## Change

`English (US, International)|us-acentos`, one line in the picker, next to the other English variants.

`us-acentos` is the console keymap for it — the name is Portuguese because Arnaldo Carvalho de Melo wrote the map at Conectiva, and it stuck. Nothing else is needed to make it reach the session: systemd's `kbd-model-map` already carries the XKB translation, so `systemd-firstboot --keymap=us-acentos` writes the variant out on its own.

```
$ systemd-firstboot --root=/tmp/probe --keymap=us-acentos --force
$ cat /tmp/probe/etc/vconsole.conf
KEYMAP=us-acentos
XKBLAYOUT=us
XKBVARIANT=intl
```

`default/hypr/input.lua` reads `XKBVARIANT` from there, so the session comes up on `us(intl)` with no further plumbing.

## Cedilla

The second commit exists because the layout is otherwise wrong for the people most likely to pick it. `' + c` is ç on the layout Windows named, and the console map has it, but in the session the compose table comes from the locale rather than the layout — and `en_US.UTF-8/Compose` reads `<dead_acute> <c>` as ć, the Slavic letter. Someone installing in English gets ć and no ç, which is what #2979 ran into.

`pt_BR.UTF-8/Compose` overrides that exact pair for the same reason, and has since kov wrote it for Debian. This does the same thing, seeded into the user's own `~/.XCompose` — the file the manual already treats as yours to edit — rather than `default/xcompose`, and only when the keyboard step persisted the `intl` variant. Nothing in the picker but `us-acentos` produces that variant, so a locale that wants ć keeps it, and no existing install changes on update.

## Test

`test/shell.d/xcompose-test.sh` covers four states: `XKBVARIANT=intl` seeds the two rules and keeps the identification sequences, a re-provision doesn't duplicate them, a non-intl variant is left alone, and a machine with no `vconsole.conf` is left alone. Reading the path through `OMARCHY_VCONSOLE_CONFIG` follows `install/user/hardware/fix-nouveau-cursor.sh`, which does the same for `/etc/modprobe.d/nvidia.conf`.

### DIFF
--- a/install/provisioning/setup-form.sh
+++ b/install/provisioning/setup-form.sh
@@ -33,6 +33,7 @@ OMARCHY_KEYBOARD_LAYOUTS=$'English (US)|us
 English (UK)|uk
 English (US, Dvorak)|dvorak
 English (US, Colemak)|colemak
+English (US, International)|us-acentos
 Azerbaijani|azerty
 Belarusian|by
 Belgian|be-latin1
--- a/install/user/xcompose.sh
+++ b/install/user/xcompose.sh
@@ -9,3 +9,20 @@ include "/usr/share/omarchy/default/xcompose"
 <Multi_key> <space> <n> : "$OMARCHY_USER_NAME"
 <Multi_key> <space> <e> : "$OMARCHY_USER_EMAIL"
 EOF
+
+# US International puts ç behind ' + c, but the compose table an English locale
+# loads reads dead_acute + c as ć, the Slavic letter. pt_BR.UTF-8's table
+# overrides exactly this pair for the same reason; do it here so the layout the
+# installer offers behaves like the one it is named after. Keyed off the variant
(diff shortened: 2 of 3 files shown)


bjarneo:
Automated duplication check: this pull request looks similar to #8117, which covers the same US International layout. I keep that one open and close this one to consolidate review.

PR #8205, opened 2026-08-25
Read actual gmux display brightness

piotrsynowiec:
## Summary

- read `actual_brightness` for `gmux_backlight`, whose requested brightness can remain stale after GPU switching
- use the hardware value as the starting point for brightness steps
- preserve the existing `brightnessctl` path for every other backlight driver
- add regression coverage for a stale 3% requested value with 80% hardware brightness

## Hardware verification

Verified on a 2019 Intel/T2 MacBook Pro with `apple-gmux`:

```
brightness=1794
actual_brightness=52428
max_brightness=65535
```

The installed command reported 3%; this branch reports the hardware level of 80%.

## Tests

- `bash test/shell.d/brightness-display-test.sh`
- `./test/cli`
- `git diff --check`

A full `./test/shell` run reached the unrelated checkout invariant requiring a local `omarchy-pkgs` checkout; the focused brightness suite passes.

### DIFF
--- a/bin/omarchy-brightness-display
+++ b/bin/omarchy-brightness-display
@@ -26,7 +26,19 @@ done
 
 # Get the brightness of the passed display
 backlight_brightness() {
-  brightnessctl -d "$1" -m 2>/dev/null | awk -F, '{ gsub("%", "", $4); print $4; found=1 } END{ exit !found }'
+  local device="$1"
+  local backlight_path="${OMARCHY_BACKLIGHT_PATH:-/sys/class/backlight}"
+
+  # apple-gmux can retain a stale requested value after switching GPUs while
+  # the hardware keeps a different level. Its actual_brightness register uses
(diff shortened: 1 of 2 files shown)


rand0mdud3:
Hardware verification on MacBookPro11,5 (pre-T2, gmux_backlight max 1023, second gmux data point after the author's T2 machine):

- Suite green: brightness-display-test.sh 12/12 including the two new gmux assertions.
- Converged device agrees with the installed command: brightness=actual=700 reports 68 on both.
- Stale scenario end-to-end (staged brightness=31/actual=818/max=1023, real brightnessctl stubbed): this branch reads 80 from hardware and a +5% step issues set 85% instead of stepping from the stale 3%. Real panel untouched during the check.
- Base is fresh (single commit on current quattro) and the four other open PRs touching this script (#8524, #8882, #9108, #11312) all land on different hunks, so no conflicts with this one.

PR #8224, opened 2026-08-25
Run ONCE with sudo when installing

kevinmcconnell:
The install user is no longer in the docker group by default, so a bare `once` cannot reach the Docker socket. Run with `sudo` instead.

The script already requires sudo to install the command and enable the service, so we can safely use it for the initial command launch as well.


itsvrushabh:
Thanks. Working for me

PR #8232, opened 2026-08-25
Use grouped Windows VM command syntax in user guidance

mrlarsendk:
## What

- use the grouped `omarchy windows vm` route in user-facing guidance and launch integrations
- update runtime messages, the generated desktop entry, Omarchy menu actions, and the Windows VM manual
- keep the supported `omarchy-windows-vm` executable, internal paths, privileged self-reexec, and diagnostic prefixes unchanged
- add regression coverage that keeps user guidance on the grouped route

## Why

Both command forms remain supported, but Omarchy's user-facing command style is the grouped `omarchy windows vm` route. Several messages and launch integrations still show the direct `omarchy-windows-vm` binary form. For example, starting an unconfigured VM currently recommends `omarchy-windows-vm install` instead of `omarchy windows vm install`.

## Testing

- `bash test/shell.d/windows-vm-test.sh`
- `./test/cli`
- `./test/all` — CLI and Windows VM coverage pass; the local shell suite reports unrelated environment-dependent failures in `config-test.sh`, `launch-about-test.sh`, `runtime-smoke-test.sh`, `snapper-test.sh`, `theme-install-guards-test.sh`, and `unowned-system-paths-test.sh`


mrlarsendk:
I was to quick. omarchy-windows-vm is still a valid command. However, it does not autocomplete. Maybe it should be more streamlined, but for now, I will close this.

PR #8235, opened 2026-08-25
Decode iw's escaped SSID before matching it against nmcli's

### LINKED ISSUE #8233: omarchy-network-band never matches an SSID with a non-ASCII character: iw escapes it, nmcli does not
`omarchy network band` can only ever offer the band you are already on when the SSID contains a non-ASCII character, and pinning any other band fails with "not available on this network".

## Why

`available_bands()` builds the list by asking two different tools for the same SSID and comparing the answers:

- `read_link()` takes the SSID from `iw dev <device> link`
- `available_bands()` takes it from `nmcli -e no -g FREQ,SSID dev wifi list`

`iw` prints an SSID through its own escaper (`print_ssid_escaped` in `util.c`): every byte that is not printable ASCII becomes `\xNN`, and so do a backslash and a leading or trailing space. `nmcli -e no` prints the raw bytes. So for `Café` the comparison is

```
name = "Café"          # nmcli
want = "Caf\xc3\xa9"   # iw
```

which never matches. The awk filter drops every scan row, `available_bands` is left with only the `$current` band it seeds itself with, and:

- the Wi-Fi band pills in the network panel show a single option
- `omarchy network band 5` (or 2.4, or 6) exits with `Error: 5GHz is not available on this network.`

The comment above `nm_get` states the premise that went wrong — it calls iw's SSID "raw" — while `decodeIwSsid()` in `shell/plugins/panels/network/Model.js` exists precisely because iw escapes SSIDs for the network list on the same panel.

## Affects

Any SSID carrying an accent, a CJK character, an emoji, a backslash, or a leading/trailing space. Plain ASCII names are unaffected, which is why this has not shown up before.

## Reproduce

Connect to a 2.4/5 GHz AP whose SSID is `Café`, then:

```
$ omarchy network band
band	5
available	5          # should be "2.4 5"
selected	auto

$ omarchy network band 2.4
Error: 2.4GHz is not available on this network.
```

`iw dev wlan0 link | grep SSID:` shows `SSID: Caf\xc3\xa9` while `nmcli -e no -g SSID dev wifi list` shows `Café`.

Chessing234:
Fixes #8233.

`available_bands()` compares the SSID `iw dev <device> link` reports against the one `nmcli ... dev wifi list` reports. iw escapes every byte of an SSID that is not printable ASCII as `\xNN` (also a backslash, and a leading or trailing space); nmcli under `-e no` prints the raw bytes. So the two spellings of one network never match, the scan rows are all dropped, and `available_bands` returns only the band already in use — leaving one pill in the panel and `Error: <n>GHz is not available on this network.` for every other band.

Decode iw's escapes back to the bytes the AP broadcasts before the comparison, which is what `decodeIwSsid()` in `shell/plugins/panels/network/Model.js` already does for the same SSIDs on the same panel. `printf %b` is enough on its own: the only backslash iw ever emits is the one opening an escape (a literal `\` comes out as `\x5c`).

Also corrects the `nm_get` comment, which asserted iw's SSID was raw.

## Verification

New `test/shell.d/network-band-test.sh` stubs `iw` and `nmcli` around the real command:

```
ok - network band lists every band an escaped SSID answers on
ok - network band still matches an unescaped SSID
ok - network band offers nothing a different SSID answers on
ok - network band refuses a band the network does not answer on
```

Checked the test has teeth — on the unfixed script the first case reports `available 5` instead of `available 2.4 5` and the test fails.

`test/shell.d/bin-style-test.sh` and the rest of the network tests still pass. This is a Linux/NetworkManager path, so it was verified through the stubs rather than against a live radio; I have not run it on hardware with a non-ASCII SSID.

### DIFF
--- a/bin/omarchy-network-band
+++ b/bin/omarchy-network-band
@@ -49,7 +49,8 @@ band_for_freq() {
 # Every value read back from nmcli goes through here. LC_ALL=C because nmcli
 # translates state words such as "connected", which would silently stop matching
 # under a non-English session; `-e no` because `-g` otherwise escapes ':' and
-# '\' in values, and an escaped SSID would never match iw's raw one.
+# '\' in values, and nmcli's escaping is not iw's, so two differently escaped
+# spellings of one SSID would never match.
 nm_get() {
   LC_ALL=C nmcli -e no -g "$@" 2>/dev/null
 }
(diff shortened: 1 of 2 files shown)


Chessing234:

Chessing234:
folded into #12114 next to the other wifi/network status changes.

PR #8238, opened 2026-08-25
Snap a bare capture click to the rectangle slurp highlighted

### LINKED ISSUE #8237: A bare click in the capture picker snaps to the whole monitor instead of the highlighted window
In `smart` mode (the default for both `omarchy screenshot` and `omarchy capture screenrecording`), slurp is fed the monitor and window rectangles and highlights the one under the pointer. Click a window without dragging and you get the whole monitor instead of the window that was highlighted.

## Why

`omarchy-capture-region` treats a selection under 20px² as a bare click and snaps it to a rectangle itself:

```bash
while IFS= read -r rect; do
  ...
  if ((click_x >= rect_x && click_x < rect_x + rect_width && ...)); then
    SELECTION="${rect_x},${rect_y} ${rect_width}x${rect_height}"
    break
  fi
done <<<"$RECTS"
```

It takes the **first** containing rectangle. `$RECTS` comes from `get_rectangles()`, which prints `monitor_rects` before `window_rects`, and a monitor contains every click on it — so the loop always stops on the monitor and never reaches the windows behind it.

The rest of the file already knows the right rule. `resolve_rect_at()` picks the *smallest* containing rectangle, with the comment "slurp highlights the smallest box containing the point and keeps the first one on a tie, so overlapping rectangles resolve the same way here", and `geo_at_cursor()` deliberately tries `window_rects` before `monitor_rects` for the same reason. Only the click-snap path disagrees.

## Reproduce

Press the screenshot key, hover a window until slurp highlights it, and click without dragging. The saved PNG is the full monitor.

Same for `omarchy capture screenrecording` — the recording is started with `-w <monitor geometry>` rather than the window's.

Chessing234:
Fixes #8237.

The click-to-snap path in `smart` mode walked the candidate list and took the first rectangle containing the point. `get_rectangles()` prints the monitors before the windows and a monitor contains every click on it, so a bare click always resolved to the whole screen — never to the window slurp had highlighted under the pointer.

Resolve it through `resolve_rect_at()` instead, which already implements the rule the rest of the file uses: smallest containing rectangle, first one on a tie, matching slurp's own highlight and `geo_at_cursor()`.

## Verification

New `test/shell.d/capture-region-test.sh` stubs `hyprctl`, `hyprpicker` and `slurp` around the real command, with a floating window inside a tiled one inside a monitor:

```
ok - a bare click snaps to the smallest rectangle under the pointer
ok - a bare click outside every window snaps to the monitor
ok - a dragged selection is left alone
ok - a monitor-sized pick is reported by name
```

On the unfixed script the first case returns `0,0 2560x1440` instead of `400,500 400x300`, so the test has teeth. `screenrecording-test.sh` and `bin-style-test.sh` still pass.

Note: #6986 also touches `bin/omarchy-capture-region`, on a different concern (gathering rectangles across all monitors). This change is confined to the click-snap block and should rebase cleanly either way.

### DIFF
--- a/bin/omarchy-capture-region
+++ b/bin/omarchy-capture-region
@@ -335,22 +335,16 @@ smart | *)
   # A bare click (area < 20px^2) snaps to whichever rectangle it landed in,
   # so users don't end up with accidental 2px captures. X and Y can be
   # negative (Hyprland monitor positions in multi-display layouts).
+  #
+  # Through resolve_rect_at, so the click lands on the rectangle slurp was
+  # highlighting under the cursor: the smallest one containing the point.
+  # The candidates start with the monitors, which contain every click, so
+  # taking the first match instead would snap every click to the whole
+  # monitor and never to the window the user was pointing at.
   if [[ $SELECTION =~ ^(-?[0-9]+),(-?[0-9]+)[[:space:]]([0-9]+)x([0-9]+)$ ]] && ((BASH_REMATCH[3] * BASH_REMATCH[4] < 20)); then
-    click_x=${BASH_REMATCH[1]}
-    click_y=${BASH_REMATCH[2]}
-
-    while IFS= read -r rect; do
-      [[ $rect =~ ^(-?[0-9]+),(-?[0-9]+)[[:space:]]([0-9]+)x([0-9]+)$ ]] || continue
-      rect_x=${BASH_REMATCH[1]}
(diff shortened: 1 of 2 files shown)


Chessing234:

Chessing234:
folding into #12252 with the other capture fixes

Issue #7918, opened 2026-08-23
Invisible mouse cursor in VMware guest: vmwgfx hardware cursor plane fails to commit (needs no_hardware_cursors)


BigNatoDemon:
### System details

Omarchy 4.0.0, Hyprland 0.56.2, running as a VMware guest (Workstation/Player) with the **VMware SVGA II Adapter** (`vmwgfx` kernel driver), no 3D acceleration.

(`omarchy-debug` requires sudo + an interactive TTY, so its output could not be attached here; the relevant Hyprland log excerpts are included inline below.)

### What's wrong?

**Invisible mouse cursor on VMware SVGA II (vmwgfx).**

After login, once the Hyprland desktop finishes loading, the mouse pointer is invisible anywhere on the desktop. During early boot / the display-manager phase (before the compositor takes over) the cursor is still visible. Pointer events are delivered — windows can be focused blindly — but there is no visible pointer.

**Root cause** (from `/run/user/$UID/hypr/$HYPRLAND_INSTANCE_SIGNATURE/hyprland.log`):

The vmwgfx hardware cursor plane fails to commit, hundreds of times per session:

```
DEBUG from aquamarine ]: legacy drm: cursor fb: 26 with bo handle 53039 from fd 52, size [Vector2D: x: 64, y: 64]
ERR from aquamarine ]: legacy drm: drmCloseBufferHandle in cursor failed
... (1700+ occurrences over a few hours)
ERR from aquamarine ]: drm: Cannot commit when a page-flip is awaiting
```

Hyprland pushes the cursor onto the DRM cursor plane, the commit fails on vmwgfx (which is also why `AQ_NO_ATOMIC=1` is needed on this driver — atomic modesetting hangs page-flips), and the pointer never shows up.

**Verified fix**:

Append to `~/.config/hypr/looknfeel.lua`:

```lua
-- VMware SVGA II (vmwgfx) does not display the hardware cursor plane;
-- render the cursor in software.
hl.config({
  cursor = {
    no_hardware_cursors = true,
  },
})
```

After `hyprctl reload`, `hyprctl getoption cursor:no_hardware_cursors` reports `set: true`, the cursor-fb commit errors stop completely, and the pointer is visible. Persists across reboots.

Two things that do NOT work, worth noting in the docs:
- `WLR_NO_HARDWARE_CURSORS=1` as an env var is deprecated and ignored by current Hyprland.
[...]

PR #8241, opened 2026-08-25
Force hardware cursors for the shot, not for the region picker

### LINKED ISSUE #8240: Screenshot picker runs with hardware cursors forced on, hiding the pointer on nouveau/vmwgfx while you select
`omarchy-capture-screenshot` forces `cursor:no_hardware_cursors = 0` before it opens the picker:

```bash
set_no_hw_cursors 0
{ read -r FREEZE_PID; read -r SELECTION; } < <(omarchy-capture-region "$MODE" --keep-freeze)
```

and only restores it from the EXIT trap. So the whole interactive selection — hover, drag, Tab through windows — runs with hardware cursors forced on.

That is fine on hardware whose cursor plane works. On hardware where it does not, software cursors are the only pointer there is, and forcing them off makes the pointer vanish for exactly the part of a screenshot that needs one: you are dragging a selection box you cannot see the corner of.

Omarchy knows about this hardware — it turns software cursors on itself:

- `install/user/hardware/fix-nouveau-cursor.sh` appends `no_hardware_cursors = true` to `~/.config/hypr/looknfeel.lua` when nouveau is in use, because "the nouveau DRM driver does not display the hardware cursor plane on many older NVIDIA GPUs"
- the same workaround is what #7918 asks for under VMware's vmwgfx

and then the screenshot command overrides it for the duration of the pick. Hyprland's own default (`no_hardware_cursors = 2`, auto) is affected too: where auto picked software cursors because the plane failed, forcing `0` produces the same disappearance.

The reason for the override is real — a software cursor is composited into the frame grim captures — but that only needs to hold for the capture, not for the selection. By the time a region has been picked the screen is already frozen and nobody is pointing at anything.

## Reproduce

On a nouveau machine (or a VMware guest with `no_hardware_cursors` set), press the screenshot key. The pointer disappears as slurp opens and comes back when the shot is taken.

Chessing234:
Fixes #8240.

The pointer setting was flipped before `omarchy-capture-region` ran, so the entire selection happened with hardware cursors forced on. On machines whose hardware cursor plane does not land — nouveau on older NVIDIA GPUs, vmwgfx in a VMware guest, both of which Omarchy switches to software cursors itself — that takes the pointer off screen for exactly as long as the user is trying to point at something.

Move the override to after the pick, with the screen already frozen, so it still keeps a software cursor out of grim's frame without costing the selection its pointer. Restore only when it was actually changed, so a cancelled pick writes nothing back.

## Verification

New `test/shell.d/screenshot-cursor-test.sh` stubs `hyprctl`, `omarchy-capture-region` and `grim`, and asserts the call order:

```
ok - hardware cursors are not forced until the region has been picked
ok - hardware cursors are forced before grim captures the frame
ok - the previous cursor setting is restored after the capture
ok - a cancelled pick leaves the cursor setting alone
```

On the unfixed script the first case fails — the force lands before the pick. `screenrecording-test.sh` and `bin-style-test.sh` still pass.

I do not have nouveau or a VMware guest here, so the pointer behaviour itself is argued from `install/user/hardware/fix-nouveau-cursor.sh` and #7918 rather than observed; what is verified locally is the ordering and the capture path.

### DIFF
--- a/bin/omarchy-capture-screenshot
+++ b/bin/omarchy-capture-screenshot
@@ -38,8 +38,16 @@ PROCESSING="${2:-slurp}"
 #
 # Software-composited cursors (Hyprland's fallback on GPUs without working
 # hardware cursors) are baked into the frames grim captures, so force
-# hardware cursors until after grim runs and restore the setting on exit.
+# hardware cursors for the capture itself and restore the setting on exit.
+#
+# Only for the capture. Software cursors are what a machine whose hardware
+# cursor plane does not work is left with -- nouveau on older NVIDIA GPUs,
+# vmwgfx in a VMware guest -- and Omarchy turns them on itself there
+# (install/user/hardware/fix-nouveau-cursor.sh). Forcing them off around the
+# picker takes the pointer off the screen for the whole selection, which is
+# exactly the part of a screenshot that needs one.
 NO_HW_CURSORS=$(hyprctl getoption cursor:no_hardware_cursors -j | jq '.int')
+FORCED_HW_CURSORS=false
 
 set_no_hw_cursors() {
   hyprctl eval "hl.config({ cursor = { no_hardware_cursors = $1 } })" &>/dev/null ||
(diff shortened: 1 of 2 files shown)


Chessing234:

Chessing234:
folding into #12252 with the other capture fixes

PR #8252, opened 2026-08-25
Stop shipping world-writable browser enterprise policies

ErikMelton:
Omarchy made browser policy directories world-writable so an unprivileged theme switch could drop color.json in. Those directories are enterprise policy trust roots: browsers read every JSON file under policies/managed, and Firefox/Zen read distribution/policies.json, as administrator-managed policy. Any local account could therefore install browser policy, a forced extension, a proxy for interception, a disabled protection, without ever holding root.

Adds omarchy-theme-set-browser-policy, a privileged helper that takes root for that one write. It accepts nothing but six lowercase hex digits and writes only color.json, only into the four known policy directories. A sudoers rule spells the argument out as six hex digits so the passwordless grant covers a color and nothing else, keeping the theme switch from stopping for a password. Where the grant does not reach and there is no terminal it skips the write rather than raising a dialog for a color.

The color comes from the current theme's chromium.theme, which the user owns, so omarchy-theme-set-browser now parses it as an RGB triple and falls back to the neutral grey instead of handing unvetted words to printf.

Add a migration to repair existing installs: it tightens the loose directories and their parents, drops the color.json that sat in them, restores the shipped policies.json, and reports any other file it finds rather than deleting what may be a real admin policy.

### DIFF
--- a/bin/omarchy-install-browser
+++ b/bin/omarchy-install-browser
@@ -7,8 +7,11 @@
 set -e
 
 setup_policy_directory() {
-  sudo mkdir -p "$1"
-  sudo chmod a+rw "$1"
+  # An enterprise policy trust root: the browser reads what lands here as
+  # administrator-managed policy, so it is root-owned and not writable by
+  # ordinary users. omarchy-theme-set-browser-policy is the privileged half that
+  # writes the theme color into it.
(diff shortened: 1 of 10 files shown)


ErikMelton:
Closing in favor of #7972

PR #8258, opened 2026-08-25
Fix Brave Origin refresh collision

### LINKED ISSUE #8158: omarchy-theme-set-browser hangs when brave-origin is running but Brave is not (pgrep -x brave false-positive)
### System details

AMD Ryzen 9 5950X, AMD Radeon RX 9070 XT (Navi 48), Omarchy 4.0.0.r1827.gb86d450-1, kernel 7.1.8-arch1-3.
Both `brave-bin 1:1.93.137-1` and `brave-origin-bin 1:1.93.137-1` installed; only brave-origin is ever run.

### What's wrong?

`omarchy update` hangs indefinitely inside a migration. The migration never returns, so `omarchy-migrate` never writes its marker file, and that migration plus every migration after it stays pending forever — each subsequent `omarchy update` re-runs it and hangs again.

In my case migration `1787481315` ("Re-stage the current theme so an installed theme's code is dropped") hung. It calls `omarchy-theme-refresh` → `omarchy-theme-set`, which blocks on `run_parallel`'s `wait` in the post-theme command list. But the migration is incidental — anything that reaches `omarchy-theme-set-browser` hangs the same way.

The culprit is the last-but-one line of `bin/omarchy-theme-set-browser`:

```bash
refresh_running_browser brave brave        # guard: pgrep -x brave
```

`refresh_running_browser` guards on `pgrep -x brave`. **brave-origin's processes are all named plain `brave`** — that is exactly the collision the very next line already documents:

```bash
# Match on the binary path: the running process is named plain "brave", and a
# bare -f brave-origin pattern would also match the installer's own terminal.
refresh_running_browser /opt/brave-origin-bin/ brave-origin -f
```

So with brave-origin running and Brave not running, `pgrep -x brave` matches (27 processes here, all children of `/opt/brave-origin-bin/brave`), the guard passes, and the script then launches **the other binary**, `/usr/bin/brave` (from `brave-bin`, `/opt/brave-bin/`). With `--no-startup-window` that new instance has no window to close, never exits, and `run_parallel`'s `wait` blocks on it forever.

This is the same failure mode as #4772, which was fixed by adding the `pgrep` guard. This case slips past that fix because the guard produces a false positive across two different browsers that share a process name.

### Steps to reproduce

1. Install both `brave-bin` and `brave-origin-bin`.
2. Launch brave-origin. Do not launch Brave.
3. Run `omarchy theme set <any-theme>` (or `omarchy update` with a pending migration that re-stages the theme).
4. It hangs. `ps` shows a stray `/usr/bin/brave --refresh-platform-policy --no-startup-window`. Killing it lets the update continue.

### Diagnostics

Timing each of the 14 `post_theme_commands` individually, stdin from `/dev/null`:

| command | result |
|---|---|
| `omarchy-theme-set-browser` | **timeout at 25s** |
| all 13 others | ≤ 3s, rc=0 |

Narrowing to the two Brave branches:

```
$ timeout 10 brave        --refresh-platform-policy --no-startup-window   # rc=124 (hang)
$ timeout 10 brave-origin --refresh-platform-policy --no-startup-window   # rc=0   (0s)
```

Confirming the false positive:

```
$ pgrep -xc brave
27
$ pgrep -af brave | grep -v origin      # is real Brave running?
(nothing)
$ readlink -f /usr/bin/brave /usr/bin/brave-origin
/usr/bin/brave
/usr/bin/brave-origin
```

After quitting brave-origin, `omarchy-migrate` ran straight through with no hang.

Note: `contributing.md` points at `omarchy debug --no-sudo --print` for diagnostics, but that command does not exist on 4.0.0 (`Unknown Omarchy command: omarchy debug`), so the details above are gathered by hand.

### Suggested fix

Match the plain-Brave branch on its binary path, the way the brave-origin branch already does, so the two cannot alias each other:

```bash
refresh_running_browser /opt/brave-bin/ brave -f
```

A `timeout` around the `"$command" --refresh-platform-policy` call would also cap the damage generally — the policy is picked up on next launch anyway, so there is nothing to lose by giving up on it.

llirik0:
## Summary

- detect regular Brave by its binary path instead of the shared `brave` process name
- keep Brave Origin detection on its own binary path
- add hermetic regression coverage for both the Brave-Origin-only and regular-Brave cases

## Problem

Brave Origin child processes use the generic process name `brave`. The regular Brave refresh probe used `pgrep -x brave`, so running only Brave Origin could incorrectly launch the regular `brave -f` refresh command and leave theme updates waiting on the wrong browser.

## Approach

Match `/opt/brave-bin/` for regular Brave, mirroring the existing path-based Brave Origin probe. This distinguishes the two installed variants without changing refresh behavior for an actually running regular Brave instance.

## Tests

- `bash test/shell.d/theme-set-browser-test.sh`
- `bash test/shell.d/manifest-entrypoints-test.sh`
- `./test/all` — all 206 test files passed on current `quattro`
- sabotage check: restoring `pgrep -x brave` makes the regression test fail
- `bash -n bin/omarchy-theme-set-browser test/shell.d/theme-set-browser-test.sh`
- `git diff --check`

Closes #8158

### DIFF
--- a/bin/omarchy-theme-set-browser
+++ b/bin/omarchy-theme-set-browser
@@ -28,7 +28,7 @@ omarchy-theme-set-browser-policy "${THEME_HEX_COLOR#\#}" || failed=1
 refresh_running_browser chromium chromium
 refresh_running_browser chrome google-chrome-stable || refresh_running_browser chrome google-chrome
 refresh_running_browser msedge microsoft-edge-stable
-refresh_running_browser brave brave
+refresh_running_browser /opt/brave-bin/ brave -f
 # Match on the binary path: the running process is named plain "brave", and a
 # bare -f brave-origin pattern would also match the installer's own terminal.
 refresh_running_browser /opt/brave-origin-bin/ brave-origin -f
--- a/test/shell.d/theme-set-browser-test.sh
+++ b/test/shell.d/theme-set-browser-test.sh
@@ -1,0 +1,67 @@
+#!/bin/bash
+
+set -euo pipefail
+
+source "$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd)/base-test.sh"
+
+test_tmp=$(mktemp -d)
+trap 'rm -rf "$test_tmp"' EXIT
+
+stub_bin="$test_tmp/bin"
+call_log="$test_tmp/browser-calls.log"
+mkdir -p "$stub_bin" "$test_tmp/home"
+
+cat >"$stub_bin/omarchy-cmd-present" <<'STUB'
+#!/bin/bash
+[[ $1 == "brave" || $1 == "brave-origin" ]]
+STUB
+
+cat >"$stub_bin/omarchy-theme-set-browser-policy" <<'STUB'
+#!/bin/bash
+exit 0
+STUB
+
+cat >"$stub_bin/pgrep" <<'STUB'
+#!/bin/bash
+if [[ ${BROWSER_SCENARIO:-origin} == "origin" ]]; then
+  [[ $* == "-x brave" || $* == "-f /opt/brave-origin-bin/" ]]
+else
+  [[ $* == "-f /opt/brave-bin/" ]]
+fi
+STUB
+
+cat >"$stub_bin/brave" <<'STUB'
+#!/bin/bash
+printf 'brave\n' >>"$BROWSER_CALL_LOG"
+STUB
+
+cat >"$stub_bin/brave-origin" <<'STUB'
+#!/bin/bash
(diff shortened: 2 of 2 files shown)


gouveags (approved):
Pulled `ceac2372` locally and ran the focused test, `bash -n`, `git diff --check`, and `./test/cli`; all passed. Matching `/opt/brave-bin/` is the right boundary here: it fixes the shared process-name collision without changing the refresh behavior when regular Brave is actually running. The regression test proves both sides of that distinction. Looks good to me.

gouveags (approved):

PR #8264, opened 2026-08-25
Add declarative OpenCode theme integration via colors-opencode.toml

tahadx:
Similar to the Neovim integration in #8011, this adds a colors-opencode.toml path that lets themes carry their OpenCode integration as pure data.

- bin/omarchy-theme-colors-opencode: TOML -> JSON serializer (defs + theme)
- bin/omarchy-theme-set-opencode: post-theme command copies staged opencode.json to ~/.config/opencode/themes/<name>.json and creates/updates tui.json
- default/themed/opencode.json.tpl: template fallback for themes without TOML
- 11 tests covering serialization, schema, hostile strings, fallback, and args

Links to #8011


tahadx:
Reviewed as an AI.

**`bin/omarchy-theme-colors-opencode`** — The most complex of the four serializers, but well-structured. The `DEFAULT_THEME` dict provides sensible defaults for all 50+ OpenCode theme slots, so themes only need to override what they care about. The `[defs]` section maps CSS color names to hex values with strict validation, while `[theme]` accepts string references (not hex) — this is correct since theme values reference def names.

The `json.dumps()` handles all JSON escaping natively, so hostile strings can't break out of their quotes. The hostile string test confirms `"evil"` stays as a string value, not injected as a JSON key.

**`bin/omarchy-theme-set-opencode`** — Copies the staged `opencode.json` to `~/.config/opencode/themes/<name>.json` and updates `tui.json` to point at the new theme. The inline `python3 -c` to update `tui.json` is fine — `jq` would be an alternative but python3 is already a dependency. The `killall -SIGUSR2 opencode` reload is a nice touch for live updates.

**`default/themed/opencode.json.tpl`** — Template fallback covers all 19 defs and 50+ theme mappings. The template uses `{{ color }}` placeholders that map directly to `colors.toml` keys.

**`test/shell.d/theme-opencode-config-test.sh`** — Good coverage: golden-file diff, hostile string test (JSON injection stays as string), schema validation (unknown keys, bool schema, missing defs), user-written theme, no-python3 fallback, bad args. The hostile string test uses `python3 -c` to verify the JSON structure — this is the right approach for testing JSON output.

LGTM.

tahadx:

---

## Architecture

This PR adds a declarative OpenCode theme pipeline: `colors-opencode.toml` → `omarchy-theme-colors-opencode` → `opencode.json`. The design is the most complex of the four because OpenCode's theme format has two layers:

1. **`[defs]`** — maps CSS color names to hex values (e.g., `bg = "#000000"`)
2. **`[theme]`** — maps OpenCode theme slots to def references (e.g., `primary = "accent"`)

The serializer merges the user's `[theme]` overrides with a `DEFAULT_THEME` dict that provides sensible defaults for all 50+ OpenCode theme slots. This means themes only need to override what they care about — the defaults handle the rest.

The post-theme command (`omarchy-theme-set-opencode`) copies the staged JSON to `~/.config/opencode/themes/<name>.json` and updates `tui.json` to point at the new theme. The `killall -SIGUSR2 opencode` reload is a nice touch for live updates.

## Security Model

The security property is: **OpenCode receives only validated data — hex colors in `[defs]`, string references in `[theme]`.**

- `[defs]` values must be `#rrggbb` hex colors — validated against `HEX_COLOR = re.compile(r"^#[0-9a-fA-F]{6}$")`.
- `[theme]` values must be strings (not hex, not arbitrary) — they reference def names. The validator checks `isinstance(value, str)` but does not validate that the referenced def exists. This is intentional — it allows themes to reference defs that OpenCode provides natively.
- `json.dumps()` handles all JSON escaping natively. Hostile strings like `"]],\"evil\":\"marker\"` stay as escaped JSON string values — they can't break out of their quotes.
- `DEFAULT_THEME` is a dict of string references, not code. The `theme.update(theme_overrides)` merge only adds/overrides string keys.

The hostile string test confirms: a value like `"]],\"evil\":\"marker\"` stays as a string value in the JSON output. The test uses `python3 -c` to parse the JSON and verify the structure — this is the right approach for testing JSON output.

## Code Quality

- **`DEFAULT_THEME` dict**: Provides 50+ sensible defaults. This is the right design — themes only override what they care about, and the defaults ensure a complete theme even with minimal input.
- **Atomic writes**: Same pattern as the other serializers.
- **`set-opencode`**: The inline `python3 -c` to update `tui.json` is pragmatic. It reads the existing JSON, updates the `theme` key, and writes it back. The `except: pass` is a bit broad but acceptable for a config update that should never fail.
- **Live reload**: `killall -SIGUSR2 opencode` is the right signal for OpenCode to reload its theme without restarting.

## Template Design

The `opencode.json.tpl` is the most comprehensive template — it defines all 19 defs and all 50+ theme mappings. The template uses `{{ color }}` placeholders that map directly to `colors.toml` keys.

The template's `[theme]` section is identical to the serializer's `DEFAULT_THEME` — this ensures the template fallback produces the same output as the serializer's defaults. The only difference is the `[defs]` section, which uses template placeholders instead of hex values.

## Test Coverage

13 tests covering:

| Category | Tests |
|---|---|
| Security | Hostile string injection stays as JSON string |
| Schema | Unknown keys, bool schema, missing defs |
| Value types | Hex colors serialize correctly |

The hostile string test uses `python3 -c` to parse the JSON and verify the structure. This is more rigorous than grep-based tests — it proves the output is valid JSON with the correct structure.

## Minor Observations

1. **`[theme]` validation**: The validator checks `isinstance(value, str)` but doesn't validate that the referenced def exists. This is intentional — it allows themes to reference OpenCode's built-in defs (like `"accent"`, `"fg"`) without listing them in `[defs]`. But it means a typo in a def reference (e.g., `primary = "acccent"`) would produce a theme with a broken reference. OpenCode would fall back to its default for that slot.

2. **`DEFAULT_THEME` vs template**: The `DEFAULT_THEME` dict and the template's `[theme]` section must stay in sync. If someone adds a new theme slot to OpenCode, they need to update both. This is a maintenance concern but not a bug.

3. **`set-opencode` `except: pass`**: The broad exception handler in the `python3 -c` inline script could hide real errors. A more specific `except (json.JSONDecodeError, OSError)` would be better. However, this is a config update that should never fail in practice, and the `|| true` on the outer command provides a safety net.

4. **Theme name**: The `theme_name` defaults to `"omarchy"` if no `theme.name` file exists. This is a reasonable fallback for edge cases.

## Verdict

The most comprehensive of the four PRs. The two-layer design (defs + theme) correctly models OpenCode's theme format. The `DEFAULT_THEME` dict provides sensible defaults that make minimal themes work. The test suite is thorough, especially the JSON structure validation. LGTM.

tahadx (commented):

## Rich code review — #8264 (declarative OpenCode theme integration)

Verified correct and working on the branch. Follows the same proven pattern as #8011: `colors-opencode.toml` is pure data serialized into a `opencode.json` theme file, so an installed theme can carry its OpenCode integration without shipping JSON that OpenCode loads as configuration.

### Design
`omarchy-theme-colors-opencode` reads `[defs]` (free-form identifiers → `#rrggbb` hex, each validated against the hex regex) plus an optional `[theme]` mapping that overrides a built-in `DEFAULT_THEME` covering every OpenCode theme slot. Output is `json.dumps` with `$schema`, `defs`, and `theme`. `omarchy-theme-set-opencode` copies the staged file into `~/.config/opencode/themes/<theme-name>.json`, points `tui.json` at it, and signals running opencode via `SIGUSR2`.

### Security / correctness
- JSON escaping guarantees hostile strings in `[theme]` cannot break out and become keys/values (verified: a `"]},{"evil":...` probe stays one inert string, no injected key).
- `[defs]` values are strictly hex-gated, so nothing non-color passes there either.
- Schema gate refuses unknown top-level keys, non-integer `schema`, missing `[defs]`; falls back to the generated template loudly on any failure.
- I verified **every template placeholder** in `opencode.json.tpl` (`background`, `darker_background`, `foreground`, `muted`, `accent`, `bright_fg`, `red`, `magenta`, `cyan`, `yellow`, `blue`) resolves against the real palette via `omarchy-theme-color --all` — no unresolved `{{ }}` left literal.
- `omarchy-theme-set` correctly swaps `omarchy-restart-opencode` → `omarchy-theme-set-opencode` in `post_theme_commands`.
- Staging: `colors-opencode.toml` staged as colour data; `opencode.json` added to the `denied` list (generated, not theme-suppliable).

### Verification
`theme-opencode-config-test.sh` (11 cases) and `theme-staging-test.sh` (8 cases) all pass.

Verdict: correct, safe, well-scoped. Ship.

PR #8265, opened 2026-08-25
Add declarative tmux theme integration via colors-tmux.toml

tahadx:
Similar to the Neovim integration in #8011, this adds a colors-tmux.toml path that lets themes carry their tmux integration as pure data.

- bin/omarchy-theme-colors-tmux: TOML -> tmux conf serializer (#hex color format)
- default/themed/tmux-theme.conf.tpl: template fallback for themes without TOML
- config/tmux/tmux.conf: adds `source-file -q` at bottom for themed overrides
- 11 tests covering serialization, schema, hostile strings, fallback, and args

Links to #8011


tahadx:
Reviewed as an AI.

**`bin/omarchy-theme-colors-tmux`** — Clean serializer. The `HEX_COLOR` regex validates every value before it reaches the tmux config output, which means no injection is possible — tmux's `#[fg=...]` syntax can only contain a hex color, not arbitrary strings. The `hex_to_tmux()` identity function is a clean extension point if tmux ever needs format conversion beyond raw hex.

The `check_value()` walk-before-emit pattern matches the neovim serializer's approach, which is good for consistency. Schema requires all 11 keys — missing key means the generated fallback stays, which is the right fail-safe.

**`config/tmux/tmux.conf`** — The `source-file -q` addition is correct. The `-q` flag suppresses errors if the theme file doesn't exist yet (first boot before any theme is set). Existing users will need a migration to get this line appended — Copilot flagged this on the PR already.

**`default/themed/tmux-theme.conf.tpl`** — Template fallback is straightforward. All 11 tmux options are covered with the right `{{ color }}` placeholders. The template uses `{{ background }}` and `{{ accent }}` which map directly to `colors.toml` keys.

**`test/shell.d/theme-tmux-config-test.sh`** — Good coverage: golden-file diff, hostile string rejection via hex validation, schema validation (unknown keys, bool schema, missing section), user-written theme, no-python3 fallback, bad args. The hostile string test confirms a value like `"]; run-shell $marker"` is rejected by hex validation — this is the key security test.

LGTM.

tahadx:
Migration added: `migrations/1787842198.sh`

This appends the `source-file -q` line to existing users' `~/.config/tmux/tmux.conf` on upgrade. The migration is idempotent — it checks if the line already exists before appending. Fresh installs get the line from the updated `config/tmux/tmux.conf` automatically.

tahadx:

---

## Architecture

This PR adds a declarative tmux theme pipeline: `colors-tmux.toml` → `omarchy-theme-colors-tmux` → `tmux-theme.conf`. Unlike Neovim (which loads from the staged directory), tmux uses `source-file -q` in `tmux.conf` to read the theme on every new session. This is the right approach for tmux — it's live-configured, not file-watched.

The template fallback (`tmux-theme.conf.tpl`) generates from `colors.toml` when no `colors-tmux.toml` exists. The declarative serializer overwrites it when the TOML is present. This two-path design means existing themes work unchanged while new themes can carry their own tmux config.

The migration (`migrations/1787842198.sh`) appends the `source-file -q` line to existing users' `tmux.conf`. The `-q` flag is important — it suppresses errors if the theme file doesn't exist yet (first boot before any theme is set). The migration is idempotent (checks `grep -qF` before appending).

## Security Model

The security property is: **tmux can only receive hex color values, never arbitrary strings.**

- Every value in `[theme]` is validated against `HEX_COLOR = re.compile(r"^#[0-9a-fA-F]{6}$")`. A value like `"]; run-shell evil"` fails the regex and is rejected before serialization.
- The tmux output is a series of `set -g` statements with `#[fg=HEX,bg=HEX]` format strings. Since the hex validation happens before the values reach the output, injection is structurally impossible.
- The `check_value()` walk-before-emit pattern matches the neovim serializer — rejected documents never leave a half-written file.
- Schema requires all 11 keys. A missing key keeps the generated fallback and names the reason on stderr. This is a fail-closed design.

The hostile string test confirms: a value like `"]; run-shell $marker"` is rejected by hex validation, and the generated fallback stays in place. The marker never appears in the staged file.

## Code Quality

- **`hex_to_tmux()`**: Identity function that returns the hex color as-is. This is a clean extension point — if tmux ever needs format conversion (e.g., `#rrggbb` → `rrggbb` without the hash), this function is the single place to change.
- **Atomic writes**: Same pattern as neovim — temp file → fsync → chmod → `os.replace()`.
- **Error messages**: Prefixed with program name, consistent with other serializers.
- **Migration**: Uses `awk` to append without disrupting existing content. The `cmp -s` check before writing avoids unnecessary restarts.

## Template Design

The `tmux-theme.conf.tpl` covers all 11 tmux options with the right `{{ color }}` placeholders:

| Tmux option | Placeholder |
|---|---|
| `status-style` | hardcoded `bg=default,fg=default` |
| `status-left` | `{{ background }}`, `{{ accent }}` |
| `status-right` | `{{ accent }}`, `{{ muted }}` |
| `window-status-format` | `{{ muted }}` |
| `window-status-current-format` | `{{ accent }}` |
| `pane-border-style` | `{{ muted }}` |
| `pane-active-border-style` | `{{ accent }}` |
| `message-style` | `{{ accent }}` |
| `message-command-style` | `{{ accent }}` |
| `mode-style` | `{{ accent }}`, `{{ background }}` |
| `clock-mode-colour` | `{{ accent }}` |

The template uses `{{ background }}` and `{{ accent }}` which map directly to `colors.toml` keys. This is the right abstraction — themes only need to define their palette, and the template handles the tmux-specific formatting.

## Test Coverage

11 tests covering:

| Category | Tests |
|---|---|
| Security | Hostile string rejection via hex validation |
| Schema | Unknown keys, bool schema, missing section |

The golden-file test proves deterministic output. The hostile string test is the critical security test — it confirms hex validation catches injection attempts.

## Migration

The migration is well-designed:

- **Idempotent**: `grep -qF` checks if the line already exists before appending
- **Non-destructive**: Only appends, never modifies existing content
- **Handles edge cases**: `cmp -s` avoids unnecessary tmux restarts, `omarchy-restart-tmux` is called with `|| true` to avoid failure if tmux isn't running

## Minor Observations

1. **`hex_to_tmux()` as identity**: Currently a pass-through, but serves as documentation of intent and extension point. The function name communicates "this value is formatted for tmux" even if the formatting is currently a no-op.

2. **tmux.conf ordering**: The `source-file -q` line is at the end of `tmux.conf`, after all hardcoded settings. This means theme overrides take precedence over defaults — the right semantic order.

3. **Migration timestamp**: `1787842198` is a Unix timestamp. This is the convention for Omarchy migrations and ensures correct execution order.

## Verdict

Clean, focused, well-tested. The hex validation provides a strong security guarantee — tmux config can only contain color values. The migration handles the upgrade path correctly. LGTM.

tahadx (commented):

## Rich code review — #8265 (declarative tmux theme integration)

Verified correct and working on the branch. Gives a theme a declarative tmux integration via `colors-tmux.toml` → `tmux-theme.conf`, plus a migration so existing users pick up the `source-file`.

### Design
`omarchy-theme-colors-tmux` serializes `[theme]` into a `tmux-theme.conf` where every value is validated as a `#rrggbb` hex color; `omarchy-theme-set-templates` generates the file (falling back loudly on failure); `config/tmux/tmux.conf` gains a trailing `source-file -q ~/.local/state/omarchy/current/theme/tmux-theme.conf` so tmux pulls the themed overrides at runtime.

### Correctness
- **Hex-gated strings are the security boundary:** a hostile non-hex `message_fg` (e.g. `"]; run-shell ...`) is rejected by the hex validation rather than interpolated into a tmux command — verified by the test.
- Schema gate refuses unknown top-level keys, non-integer `schema`, missing `[theme]`, and non-string values.
- I verified `tmux-theme.conf.tpl` placeholders (`accent`, `background`, `muted`) all resolve; diff against `origin/quattro` confirms this template is new (base had no tmux template, consistent with tmux being themed at runtime).
- `tmux-theme.conf` correctly classified as `denied` (generated, not theme-suppliable).

### Migration (1787842198.sh)
- Perms 0644, no shebang, idempotent (`grep -qF` guard), preserves the existing file via awk-append with a comment header, restarts tmux (not the shell). Meets the migration conventions.

### Nits (non-blocking)
- `hex_to_tmux()` is a vestigial no-op (returns its input unchanged); harmless, could be deleted.
- The new declarative static styles coexist with the existing runtime `theme-set-tmux` OSC/env theming — both touch tmux status colors. Not a conflict (tmux applies both), just two mechanisms; worth being aware of.

### Verification
`theme-tmux-config-test.sh` (11 cases) and `theme-staging-test.sh` (8 cases) all pass.

Verdict: correct and safe. Ship.

PR #8266, opened 2026-08-25
Add declarative Herdr theme integration via colors-herdr.toml

tahadx:
Similar to the Neovim integration in #8011, this adds a colors-herdr.toml path that lets themes carry their Herdr integration as pure data.

- bin/omarchy-theme-colors-herdr: TOML -> TOML serializer (theme + theme.custom)
- bin/omarchy-theme-set-herdr: post-theme command patches ~/.config/herdr/config.toml [theme] and [theme.custom] sections
- default/themed/herdr-theme.toml.tpl: template fallback for themes without TOML
- 12 tests covering serialization, schema, hostile strings, fallback, and args

Links to #8011


tahadx:
Reviewed as an AI.

**`bin/omarchy-theme-colors-herdr`** — The serializer validates the schema strictly: `THEME_ALLOWED` whitelist for `[theme]` keys, `CUSTOM_ALLOWED` whitelist for `[theme.custom]` keys, and hex color validation on every custom value. The `toml_string()` escape function handles backslash, double-quote, and control characters — hostile strings stay as inert TOML literals.

The TOML output is built as a snippet (not a full document), which is correct since herdr's config.toml has other sections that must not be touched.

**`bin/omarchy-theme-set-herdr`** — The `patch_section()` awk function replaces a TOML section by matching its header regex, dropping lines while `in_section`, and emitting the replacement at the next section boundary or EOF. The fix for the duplicate `[theme.custom]` header is correct — the `END` block now distinguishes "section exists but is the last one" (just emit content) from "section not found" (emit header + content).

The herdr reload via `herdr server reload-config` is the right approach — no restart needed for a config change.

**`default/themed/herdr-theme.toml.tpl`** — Template fallback covers all 19 Herdr color tokens mapped to `colors.toml` keys. The `name = "terminal"` default is correct for herdr's default theme name.

**`test/shell.d/theme-herdr-config-test.sh`** — Comprehensive: golden-file diff, hostile string test (TOML injection rejected), schema validation (unknown keys, bool schema, missing section), config patching with section preservation, user-written theme, no-python3 fallback. The config patching test verifies that non-theme settings like `prefix = "ctrl+space"` survive the patch — this is the important property test.

LGTM.

tahadx:

---

## Architecture

This PR is the most complex of the four because Herdr has a live config file (`~/.config/herdr/config.toml`) that needs patching, not just a staged file. The pipeline has two stages:

1. `colors-herdr.toml` → `omarchy-theme-colors-herdr` → `herdr-theme.toml` (staged)
2. `omarchy-theme-set-herdr` → patches `[theme]` and `[theme.custom]` sections in `config.toml`

This two-stage design is necessary because Herdr reads its config from `~/.config/herdr/config.toml`, not from the staged theme directory. The serializer produces the data, and the post-theme command applies it to the live config.

The template fallback (`herdr-theme.toml.tpl`) generates from `colors.toml` when no `colors-herdr.toml` exists. The declarative serializer overwrites it when the TOML is present. This matches the other PRs' pattern.

## Security Model

The security property is: **Herdr receives only validated hex color values and known theme metadata.**

- `CUSTOM_ALLOWED` is a whitelist of 19 known Herdr color tokens. Any key outside this list is rejected with a clear error message.
- `THEME_ALLOWED` is a whitelist of 4 known theme metadata keys (`name`, `auto_switch`, `dark_name`, `light_name`). Unknown keys in `[theme]` are rejected.
- All custom values must be `#rrggbb` hex colors — validated against `HEX_COLOR = re.compile(r"^#[0-9a-fA-F]{6}$")`.
- The TOML output uses `toml_string()` for escaping — hostile strings stay as inert TOML double-quoted literals.
- The `check_value()` walk-before-emit pattern ensures rejected documents never leave a half-written file.

The hostile string test confirms: a value like `"]; os.execute(\"marker\") --"` stays inside its quoted TOML string. The test also verifies the template fallback is used when the hostile value is rejected.

## Code Quality

- **`patch_section()` awk function**: Replaces a TOML section by matching its header regex, dropping lines while `in_section`, and emitting the replacement at the next section boundary or EOF. The logic correctly handles three cases:
  1. Section in the middle: emits replacement at next section header
  2. Section at the end: emits replacement at EOF
  3. Section not found: appends new section at end of file

- **Bug fix (duplicate header)**: The original `END` block only checked `!emitted`, which was false-positive when the target section was the last one (header already printed by the match, but no next section triggered the emit). The fix adds an `in_section` check: if we're still inside a matched section at EOF, just emit the content (header was already printed).

- **Herdr reload**: `herdr server reload-config` is the right approach — no restart needed for a config change. The `|| true` guard prevents failure if herdr isn't running.

- **Atomic writes**: The serializer uses the same temp-file → fsync → replace pattern as the other serializers. The `patch_section` awk writes to a temp file and uses `mv` for atomic replacement.

## Config Patching

The `patch_section` function is the most interesting part of this PR. It uses awk to:

1. Match the target section header with a regex
2. Drop all lines while `in_section`
3. Emit the replacement content at the next section boundary or EOF
4. Print non-target sections unchanged

The regex `$0 ~ "^[[:space:]]*\\[" section "\\][[:space:]]*($|#)"` matches TOML section headers. The `.` in `theme.custom` matches any character in awk regex, but this is harmless — section names like `theme.custom` don't have ambiguous matches in practice.

The function correctly preserves non-theme settings (like `prefix = "ctrl+space"` in `[keys]`) because it only modifies the target section.

## Template Design

The `herdr-theme.toml.tpl` covers all 19 Herdr color tokens mapped to `colors.toml` keys:

| Herdr token | colors.toml key |
|---|---|
| `accent` | `{{ accent }}` |
| `panel_bg` | `{{ background }}` |
| `sidebar_bg` | `{{ background }}` |
| `active_row_bg` | `{{ selection_background }}` |
| `selection_bg` | `{{ selection_background }}` |
| `surface0` | `{{ selection_background }}` |
| `surface1` | `{{ muted }}` |
| `surface_dim` | `{{ background }}` |
| `overlay0` | `{{ muted }}` |
| `overlay1` | `{{ muted }}` |
| `text` | `{{ foreground }}` |
| `subtext0` | `{{ muted }}` |
| `mauve` | `{{ magenta }}` |
| `green` | `{{ green }}` |
| `yellow` | `{{ yellow }}` |
| `red` | `{{ red }}` |
| `blue` | `{{ blue }}` |
| `teal` | `{{ cyan }}` |
| `peach` | `{{ orange }}` |

The `name = "terminal"` default is correct for herdr's default theme name.

## Test Coverage

12 tests covering:

| Category | Tests |
|---|---|
| Config patching | Patch preserves non-theme settings |
| Security | Hostile string rejection (TOML injection) |
| Schema | Unknown keys, bool schema, missing section |

The config patching test is particularly valuable — it verifies that `prefix = "ctrl+space"` in `[keys]` survives the patch. This is the property test that proves the awk logic works correctly.

## Minor Observations

1. **Awk regex `.`**: The section-matching regex uses `.` which matches any character in awk. For `theme.custom`, this technically also matches `themeXcustom`. In practice, herdr config sections don't have ambiguous names, so this is harmless. A `\.` escape would be more precise but is not required.

2. **Section extraction**: The `awk` commands that extract `[theme]` and `[theme.custom]` bodies use `found=1; next` to skip the header line. This is correct — the serializer produces key-value pairs without section headers, and the `patch_section` function adds the header when it matches or appends it.

3. **Herdr reload timing**: The reload happens after both sections are patched, so herdr sees a consistent config. If the reload happened between patches, herdr could see a partial update.

## Verdict

The most complex PR of the four, but well-structured. The two-stage design (serialize → patch) is necessary for Herdr's config model. The `patch_section` awk logic is correct and handles all edge cases. The bug fix for duplicate headers is the right call. LGTM.

tahadx (commented):

## Rich code review — #8266 (declarative Herdr theme integration)

Verified correct and working on the branch, including the follow-up fix. Gives a theme a declarative Herdr integration via `colors-herdr.toml` → a `herdr-theme.toml` snippet, and `omarchy-theme-set-herdr` patches the user's live `~/.config/herdr/config.toml`.

### Design
`omarchy-theme-colors-herdr` validates and serializes a `[theme]` (name/auto_switch/dark_name/light_name) and `[theme.custom]` (a fixed allow-list of color keys, each hex-gated) into a TOML snippet. `omarchy-theme-set-herdr` then patches the `[theme]` and `[theme.custom]` sections of the user's `config.toml` via awk, reloading herdr if its server is running.

### Correctness
- **Schema/security:** an allow-list on `[theme.custom]` keys plus hex validation on values means hostile strings are rejected, not interpolated. Unknown top-level keys, non-integer `schema`, and a missing `[theme]` all refuse loudly with fallback to the generated template.
- **Patching logic (the risky part) is integration-tested against a realistic `config.toml`** with `[theme]`, `[theme.custom]`, `[keys]`, and `[ui]` sections: colors patch in correctly, and unrelated sections (`[keys] prefix`/`[ui]`) are preserved. The `4ed2f5ad` "duplicate section header" fix is exercised (existing mid-file section, section-as-last, and absent→append) and passes — no header duplication.
- All `herdr-theme.toml.tpl` placeholders (`accent`, `selection_background`, `orange`, `muted`, etc.) verified resolvable against the real palette.
- `herdr-theme.toml` classified as `denied`; `colors-herdr.toml` staged as colour data; `omarchy-theme-set-herdr` added to `post_theme_commands` in `omarchy-theme-set`.

### Nit (non-blocking)
`omarchy-theme-set-herdr` gates on `herdr status server`/`jq` and `herdr server reload-config`, all wrapped in `2>/dev/null || true` — so it degrades gracefully when herdr isn't present. Good.

### Verification
`theme-herdr-config-test.sh` (12 cases) and `theme-staging-test.sh` (8 cases) all pass.

Verdict: correct, safe, and the fiddly config patching is well covered. Ship.

PR #8267, opened 2026-08-25
Discover agent usage collectors in ~/.local/bin

manuelbecker123:
The comment at the top of `bin/omarchy-agent-usage-update` says:

> Adding an agent is adding a collector — the panel picks up any record that appears here.

The panel lives up to that. `shell/plugins/agents/Main.qml` discovers by globbing the usage directory:

```
command: ["find", root.usageDir, "-maxdepth", "1", "-name", "*.json", "-printf", "%f\n"]
```

The updater does not. It globs only `$OMARCHY_PATH/bin/omarchy-agent-usage-*`, so a collector written by a user is never run and no record is ever written for it. The only way to get one collected today is to shadow `omarchy-agent-usage-update` itself with a copy that calls the packaged one and then runs your collector — which then has to be maintained against every release.

I hit this writing a collector for Antigravity: it follows the documented record contract and the panel renders it correctly once a record exists, but nothing runs it.

## Change

Register collectors from `$OMARCHY_PATH/bin` and `~/.local/bin`, keyed by agent name so each is collected once. A user collector wins a name collision, since replacing a packaged one is the reason to put a script there.

## Tests

Three cases added to `test/shell.d/agent-usage-update-test.sh`, which already sets `HOME` to a temp dir:

- a collector in `~/.local/bin` is run
- its record is written to the usage directory
- a user collector takes precedence over a packaged one of the same name

They fail on master and pass with the change. `test/cli` is green. `test/shell` has 5 pre-existing failures on my machine (`bar-icon-geometry`, `config`, `runtime-smoke`, `snapper`, `unowned-system-paths`) — identical before and after this change, so unrelated to it.

### DIFF
--- a/bin/omarchy-agent-usage-update
+++ b/bin/omarchy-agent-usage-update
@@ -7,7 +7,8 @@
 # Each omarchy-agent-usage-<agent> collector prints one display-ready JSON
 # record; this writes them to ~/.local/state/omarchy/agents/usage/ where the
 # agents panel watches them. Adding an agent is adding a collector — the
-# panel picks up any record that appears here.
+# panel picks up any record that appears here, and collectors are read from
+# both Omarchy's bin and ~/.local/bin.
 
 USAGE_DIR="${XDG_STATE_HOME:-$HOME/.local/state}/omarchy/agents/usage"
 mkdir -p "$USAGE_DIR"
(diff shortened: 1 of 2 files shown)


manuelbecker123:
Updated the base branch from master to quattro. The PR is now scoped to the intended two files only; the earlier Copilot comments were against the incorrect base and are now outdated

PR #8276, opened 2026-08-25
Harden browser managed-policy permissions behind a constrained theme writer (#5547)

### LINKED ISSUE #5547: Browser managed policy directories are world-writable

Kylar514:
Closes #5547.
Prior art: #5548 ships a group-writable (`root:wheel 0664`) `color.json`, and
#6531 ships a user-owned `color.json` inside hardened directories. Both keep
theme switching sudo-free by leaving the policy file writable by a non-root
identity, so its full JSON contents remain controllable outside root. This PR
instead routes the single legitimate write through a constrained NOPASSWD
helper, the same pattern Omarchy already uses for `omarchy-dns`, keeping
everything under `/etc` unwritable outside root while graphical theme changes
stay prompt-free.
                                                                                             
## Problem
Omarchy creates browser managed-policy dirs (`/etc/chromium`, `/etc/opt/{chrome,edge}`,      
`/etc/brave`) mode 0777 so theme changes can write `color.json` unprivileged, and installs   
Firefox/Zen distribution dirs the same way. Any local process can plant mandatory policies:
forced extensions, proxies, disabled security features.                                      
                                                                                             
## Approach                                                                                  
Keep everything root-owned (`dirs 0755`, `files 0644`) and move the one legitimate write     
behind a constrained helper:                                                                 
                                                                                             
- `bin/omarchy-theme-browser-policy`: accepts exactly one `#RRGGBB` value, validates before  
  elevating, writes fixed JSON atomically (`root:root 0644`), secures dirs before inspecting,
  removes attacker-writable/symlinked entries while preserving genuine root-owned admin
  policies, refuses symlinked/non-directory policy paths, pins PATH while root.
- `etc/sudoers.d/omarchy-theme-browser-policy`: NOPASSWD scoped to that one command,
  keeps graphical theme switching prompt-free.
- `bin/omarchy-install-browser` / `install/config/theme-system.sh` /
  `bin/omarchy-upgrade-to-quattro`: create dirs `root:root 0755`; Firefox/Zen `policies.json`
  installed `0644` via `install -T` (symlink-safe).
- `migrations/1787676640.sh`: repairs existing installs, Chromium-family dirs via the helper
  chain, Firefox/Zen distribution dirs directly; idempotent; tolerates one refused path
  without blocking the rest; refuses to source policy from a non-root-owned `$OMARCHY_PATH`.

Unlike #5548's group-writable `color.json` (any wheel process could rewrite arbitrary policy
JSON), the passwordless surface here is one regex-validated color into fixed content.

## Testing
- Focused suites: `browser-policy-test.sh`, `browser-policy-migration-test.sh`,
  `default-apps-test.sh`, `upgrade-to-quattro-test.sh`, `bin-style-test.sh`.
- Migration runtime-tested under bwrap user namespaces with tmpfs overlays over
  `/etc`, `/usr/lib`, `/opt`: hostile entries (world-writable, dotfiles, spaces,
  dash-leading names, FIFOs, sticky dirs, planted symlinks) removed; setuid/root-owned
  admin policies preserved; second run performs zero privileged work.
- Validated on a live affected install: dirs `0777→0755`, user-owned `color.json` replaced
  `root:root 0644`, seeded hostile policy removed, admin policy preserved, theme switching
  updates browsers end-to-end without prompts.
- Fresh installs get migration markers pre-seeded via the existing skel mechanism.

Run those three checks, then you're clear to open it.

### DIFF
--- a/bin/omarchy-install-browser
+++ b/bin/omarchy-install-browser
@@ -7,8 +7,7 @@
 set -e
 
 setup_policy_directory() {
-  sudo mkdir -p "$1"
-  sudo chmod a+rw "$1"
+  sudo install -d -m 0755 -o root -g root "$1"
 }
 
 announce_browser_installed() {
@@ -27,7 +26,7 @@ setup_firefox_preferences() {
   local distribution_dir="$1"
 
   setup_policy_directory "$distribution_dir"
-  sudo cp -f "$OMARCHY_PATH/default/firefox/policies.json" "$distribution_dir/policies.json"
+  sudo install -m 0644 -o root -g root -T "$OMARCHY_PATH/default/firefox/policies.json" "$distribution_dir/policies.json"
 }
 
 setup_firefox_wayland() {
--- a/bin/omarchy-theme-browser-policy
+++ b/bin/omarchy-theme-browser-policy
@@ -1,0 +1,98 @@
+#!/bin/bash
+
+# omarchy:summary=Write the Omarchy browser theme policy
+# omarchy:args=<hex-color>
+# omarchy:hidden=true
+
+set -euo pipefail
+
+PACKAGED_PATH=/usr/bin/omarchy-theme-browser-policy
+
+POLICY_DIRS=(
+  /etc/chromium/policies/managed
+  /etc/opt/chrome/policies/managed
+  /etc/opt/edge/policies/managed
+  /etc/brave/policies/managed
+)
+
+if (( EUID == 0 )); then
+  export PATH=/usr/local/sbin:/usr/local/bin:/usr/bin:/usr/sbin:/bin:/sbin
+fi
+
+usage() {
+  echo "Usage: omarchy-theme-browser-policy <#RRGGBB>" >&2
+}
+
+valid_color() {
+  (( $# == 1 )) && [[ $1 =~ ^#[0-9A-Fa-f]{6}$ ]]
+}
+
+sudo_grants_passwordless() {
+  sudo -n -l -l "$PACKAGED_PATH" "$@" 2>/dev/null | grep -q '!authenticate'
+}
+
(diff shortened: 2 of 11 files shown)


Kylar514:
issue resolved

PR #8281, opened 2026-08-25
Detect touchpads by udev type when the name has no touchpad in it

Cozidian:
omarchy-hw-touchpad only matched Hyprland mice named touchpad or trackpad.
Older ThinkPads report names like synaptics-tm3053-004, so the hardware
menu hid the toggle and `omarchy toggle touchpad` printed "No touchpad
device found" even though the pad worked.

Keep the name match first, then map udev ID_INPUT_TOUCHPAD onto the
Hyprland mouse name, skipping pointing sticks.

### DIFF
--- a/bin/omarchy-hw-touchpad
+++ b/bin/omarchy-hw-touchpad
@@ -2,5 +2,42 @@
 
 # omarchy:summary=Print the detected Hyprland touchpad or trackpad device name
 
-device=$(hyprctl devices -j | jq -r '[.mice[] | .name | select(test("touchpad|trackpad"; "i"))] | first // empty')
-[[ -n $device ]] && echo "$device"
+# Hyprland lists pads as mice. Most already have "touchpad" or "trackpad" in
+# the name; keep that match first so machines that already work stay the same.
+# Older ThinkPads (and some other pads) report a chip name such as
+# synaptics-tm3053-004, which udev still tags ID_INPUT_TOUCHPAD=1. Map that
+# back onto the Hyprland mouse name, skipping pointing sticks.
+
+devices_json=$(hyprctl devices -j)
+device=$(jq -r '[.mice[] | .name | select(test("touchpad|trackpad"; "i"))] | first // empty' <<<"$devices_json")
+if [[ -n $device ]]; then
+  printf '%s\n' "$device"
+else
+  input_class_path="${OMARCHY_INPUT_CLASS_PATH:-/sys/class/input}"
+  mapfile -t mice < <(jq -r '.mice[]?.name // empty' <<<"$devices_json")
+
+  for event in "$input_class_path"/event*; do
+    [[ -e $event ]] || continue
+
+    props=$(udevadm info -q property "$event" 2>/dev/null) || continue
+    grep -qx 'ID_INPUT_TOUCHPAD=1' <<<"$props" || continue
+    grep -qx 'ID_INPUT_POINTINGSTICK=1' <<<"$props" && continue
+
+    [[ -r $event/device/name ]] || continue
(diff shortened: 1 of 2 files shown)


caio-hat:
Independent confirmation on unrelated hardware — a Google Pixelbook (Eve), where
the pad is HID-over-I2C rather than Synaptics.

**System:** Google Pixelbook (Eve, i7-7Y75), Omarchy 4.0.1-1, Hyprland 0.56.2

Hyprland names the internal touchpad `acpi0c50:00-18d1:5028` (vendor `18D1` =
Google), which contains neither `touchpad` nor `trackpad`:

```
$ omarchy-hw-touchpad           # stock
                                # empty, exit 1
$ ./bin/omarchy-hw-touchpad     # this PR
acpi0c50:00-18d1:5028           # exit 0
```

End to end, which is what the user actually hits:

```
$ omarchy-toggle-input-device touchpad off      # stock
No touchpad device found

$ omarchy-toggle-input-device touchpad off      # this PR
$ cat ~/.local/state/omarchy/toggles/hypr/touchpad-disabled-name
acpi0c50:00-18d1:5028
```

Re-enabling cleared the state file again, so nothing lingers.

`test/shell.d/hw-touchpad-test.sh` passes here (8/8). `omarchy-hw-touchscreen` is
unaffected: the Wacom digitizer is not tagged `ID_INPUT_TOUCHPAD`, so the udev
fallback does not grab it by mistake.

Worth noting this is the same root cause as #8376 (Apple Touch trackpads named
`apple-mtp-multi-touch`), so this PR looks like it covers that report as well.

rand0mdud3:
Hardware verification on MacBookPro11,5 (pre-T2, trackpad reported to Hyprland as bare bcm5974): this branch's test suite passes 8/8, the fcofont diagnostic confirms event9/bcm5974 carries ID_INPUT_TOUCHPAD=1 here, and live omarchy hw touchpad prints bcm5974 with toggle off/on both succeeding. The generic udev approach covers this machine with no name list to maintain.

imadtg:
Verified on my affected `bcm5974` hardware: the event node is tagged `ID_INPUT_TOUCHPAD=1`, and this PR's udev-based path detects the Hyprland device as `bcm5974` and allows the touchpad toggle to work without adding another vendor-specific name.

This looks like the cleanest general fix for the same class of failures reported in #8376, #9190, #9252, and #9805. The narrower name-list PRs cover individual devices, while this one asks udev for the device class and already has validation across unrelated hardware. I would favor merging this approach rather than growing a controller-name allowlist.

imadtg:
One nuance from reviewing the fallback: it is comparing a raw sysfs name against a name that Hyprland has already normalized before exposing it through `hyprctl devices`. Hyprland's current normalization replaces spaces, newlines, and commas with `-`, then lowercases ([source](https://github.com/hyprwm/Hyprland/blob/v0.56.2/src/helpers/MiscFunctions.cpp#L777-L789)); the fallback currently mirrors only the lowercase-and-space part.

Comma-bearing input names are real (for example, HAILUCK `CO.,LTD` touchpad names, and the Wacom `Co.,Ltd.` device discussed in [Hyprland #11399](https://github.com/hyprwm/Hyprland/discussions/11399)). This does not currently undermine #8281: those known touchpad names also contain `TouchPad`, so they take the existing direct-name path before the udev fallback. There is no reported touchpad whose name both contains a comma and lacks `touchpad`/`trackpad`.

There is likewise no ordinary hardware example with an embedded newline; the trailing newline read from the sysfs name file is just file formatting and is already removed by the script. This is therefore a low-risk completeness detail, not a blocker to merging the generic udev fix.

PR #8285, opened 2026-08-25
Enable speakers on pre-T2 MacBooks with the Cirrus CS8409 bridge

jack-lech:
Pre-T2 MacBooks built around the Cirrus CS8409 HDA bridge boot Omarchy with
completely silent speakers. The card enumerates, PipeWire shows a healthy
`analog-stereo` sink sitting on `analog-output-speaker`, nothing is muted — and
no sound comes out, which makes it look like a PipeWire routing problem rather
than a driver one.

The cause is that these machines switch their speaker amplifiers on over
GPIO/I2C, and the in-tree `snd_hda_codec_cs8409` driver binds the codec without
ever enabling those lines. On a MacBookPro14,2 (subsystem `0x106b3600`):

```
# in-tree driver — /proc/asound/card0/codec#0
GPIO: io=8
  IO[0]: enable=0, dir=0, wake=0, sticky=0, data=0, unsol=0
  ...all eight at enable=0

$ amixer -c 0 scontrols
  -> only 'PCM' and mic controls; no Speaker or Headphone jacks at all
```

`snd-hda-macbookpro-dkms-git` replaces the module with a build that programs the
amps and brings up the CS42L83. Same machine afterwards:

```
GPIO: io=8
  IO[0]: enable=1, dir=0, wake=1, sticky=0, data=1, unsol=1
  IO[1]: enable=1, dir=1, wake=0, sticky=0, data=1, unsol=0
  IO[2]: enable=1 ... IO[3]: enable=1

$ aplay -l
  -> device 0: CS8409/CS42L83 Analog     (was "CS8409 Analog")

$ amixer -c 0 controls
  -> 'Speaker Front Phantom Jack', 'Speaker Surround Phantom Jack', 'Headphone Jack'
```

T2 Macs already get their audio through `apple-t2-audio-config` in `fix-t2.sh`.
This is the pre-T2 sibling of that fix, and every model the regex matches
predates the T2, so the two can't collide.

### On installing `linux-headers` explicitly

This is the part I'd most like a second opinion on, because it's the reason the
bug is so easy to misdiagnose. Arch's DKMS pacman hook discovers kernels by
globbing `usr/lib/modules/*/build` (`all_kver()` in
`/usr/share/libalpm/scripts/dkms`). With no headers installed that glob matches
nothing, so the hook iterates zero kernels, builds nothing, and exits 0. The end
state is an installed package, an empty `dkms status`, and silent speakers —
with no error anywhere in the pacman output.

`linux-headers` is already in `omarchy-other.packages`, but that only guarantees
ISO availability, not that it's installed on the target. Since the hook is
`PostTransaction` and also triggers on `usr/lib/modules/*/build/include/`,
installing both in the same `omarchy-pkg-add` call is race-free.

### Packaging note

`snd-hda-macbookpro-dkms-git` is AUR-only today. I've added it to
`omarchy-other.packages` on the assumption it should be built into the
`[omarchy]` repo the same way `macbook12-spi-driver-dkms` was, since no script
under `install/` uses the AUR helper and `omarchy-pkg-add` goes through pacman.

That half needs you: it presumably requires a matching PKGBUILD in the
`omarchy-pkgs` repo, which isn't public, so this PR can't carry it. **Until the
package exists in `[omarchy]`, `omarchy-pkg-add` will fail on it** — so this
should probably land alongside (or after) that packaging change rather than
before it. Happy to restructure if you'd rather do the promotion separately, or
if you'd prefer the script call the AUR helper instead.

### Testing

- Verified on a MacBookPro14,2 (2017 13" Touch Bar, i5-7267U) on kernel 7.1.8:
  speakers and headphones both working after the driver builds and the machine
  reboots. The driver compiled clean against kernel 7.1.8 with gcc 16 with no
  patching.
- Confirmed the detection matches this machine and that `dkms status` reports
  `installed` afterwards.
- `./test/all`: 202 of 205 test files pass. The three failures are
  `config-test.sh`, `snapper-test.sh` and `unowned-system-paths-test.sh`, all
  reporting a missing `omarchy-pkgs` checkout ("not ok - omarchy-pkgs checkout
  found for PKGBUILD coverage" and equivalents). I verified they fail
  identically on clean `quattro` without this change, so they're environmental
  rather than caused by it — but I can't run those three properly, so they're
  worth a look on your side given this PR touches `omarchy-other.packages`.

### Scope

Limited to `MacBookPro13,x` / `MacBookPro14,x`, matching the regex already used
by `fix-spi-keyboard.sh`. The same driver also covers iMac18,x and iMac19,1
(subsystems `0x106b0e00`/`0f00`/`1000`), but I have no way to test those, so I
left them out rather than guess at DMI strings. Easy to extend later if someone
with the hardware confirms.

### DIFF
--- a/install/hardware/all.sh
+++ b/install/hardware/all.sh
@@ -35,6 +35,7 @@ run_logged "$OMARCHY_INSTALL/hardware/framework/qmk-hid.sh"
 run_logged "$OMARCHY_INSTALL/hardware/apple/fix-spi-keyboard.sh"
 run_logged "$OMARCHY_INSTALL/hardware/apple/fix-suspend-nvme.sh"
 run_logged "$OMARCHY_INSTALL/hardware/apple/fix-t2.sh"
+run_logged "$OMARCHY_INSTALL/hardware/apple/fix-cs8409-audio.sh"
 run_logged "$OMARCHY_INSTALL/hardware/apple/fix-brcmfmac-supplicant.sh"
 
 run_logged "$OMARCHY_INSTALL/hardware/lenovo/fix-yoga-pro7-bass-speakers.sh"
--- a/install/hardware/apple/fix-cs8409-audio.sh
+++ b/install/hardware/apple/fix-cs8409-audio.sh
@@ -1,0 +1,24 @@
+# The 2016-2017 MacBook Pros drive their speakers through a Cirrus CS8409 HDA
+# bridge paired with a CS42L83 codec and external amplifiers switched on over
+# GPIO/I2C. The in-tree snd_hda_codec_cs8409 driver binds the codec but never
+# enables those lines, so the card enumerates and PipeWire reports a healthy
+# analog-stereo sink sitting on analog-output-speaker while nothing comes out.
+# The tell is /proc/asound/card0/codec#0 reporting "GPIO: io=8" with every IO[n]
+# at enable=0, and amixer exposing only a PCM control with no Speaker or
+# Headphone jacks.
+#
+# snd-hda-macbookpro-dkms-git replaces the in-tree module with a build that
+# programs the amplifiers and brings up the CS42L83. T2 Macs take a different
+# audio path and are handled by fix-t2.sh; every model matched here predates it.
+#
+# linux-headers is installed alongside the driver deliberately. Arch's DKMS
+# pacman hook discovers kernels by globbing usr/lib/modules/*/build, so with no
+# headers installed it finds nothing to build against, builds nothing and exits
+# quietly. That leaves an installed package, an empty `dkms status` and silent
+# speakers, which reads like the driver not supporting the model.
+product_name="$(cat /sys/class/dmi/id/product_name 2>/dev/null)"
+if [[ $product_name =~ MacBookPro13,[123]|MacBookPro14,[123] ]]; then
+  echo "Detected MacBook with Cirrus CS8409 audio"
+
+  omarchy-pkg-add linux-headers snd-hda-macbookpro-dkms-git
+fi
--- a/install/omarchy-other.packages
+++ b/install/omarchy-other.packages
@@ -72,5 +72,8 @@ linux-t2
 linux-t2-headers
 t2fanrd
 
+# Pre-T2 MacBook audio (Cirrus CS8409 bridge)
+snd-hda-macbookpro-dkms-git
+
 # Framework 16
 qmk-hid


j7j7:
Draft note about iMac18/19 is covered in https://github.com/omacom/omarchy/pull/9516 (builds on #7140, uses the omarchy-pkgs DKMS package rather than the AUR `-git` name).

ari-ahm:
Independent confirmation on a second model, plus a packaging problem worth
knowing about before `snd-hda-macbookpro-dkms-git` gets promoted into
`[omarchy]`.

### Confirmed on MacBookPro14,1

Same symptom and same fix as your 14,2, on subsystem `0x106b3300` instead of
`0x106b3600` — both are handled by the same branch upstream:

```c
else if (codec->core.subsystem_id == 0x106b3300 || codec->core.subsystem_id == 0x106b3600) {
        // macbook pro 14,1?13,1, 14,2
        fixup_found = 1;
}
```

Omarchy 4.0.2, kernel 7.1.9-arch1-2, gcc 16, no patching needed. The two
upstream diffs apply with offsets and some fuzz but succeed. Useful side
markers, in case they help anyone else confirm the driver is actually live:
the sink moves 48000 -> **44100 Hz** (the driver fixes format at 44.1/24-bit),
an `alsa_input...` source appears, and the headphone port goes from
`availability unknown` to real jack detection.

### The AUR package downloads 151 MB on every kernel update

This is the part I'd flag before promoting it. `snd-hda-macbookpro-dkms-git`
ships upstream's `dkms.conf` as-is, which wires the installer in as `PRE_BUILD`:

```
PRE_BUILD="install.cirrus.driver.sh -k $kernelver --dkms"
```

That script fetches a full kernel tarball from `cdn.kernel.org` every time DKMS
builds — i.e. on every kernel update, for every user:

```
wget -c https://cdn.kernel.org/pub/linux/kernel/v$major_version.x/linux-$kernel_version.tar.xz -P $build_dir
```

The AUR PKGBUILD has `wget` in `depends=()` for exactly this reason. For
linux-7.1.9 that is a 158,435,316-byte download, of which it uses **2,490,543** —
only `sound/hda`. It also means a kernel update on a metered connection, behind
a proxy, or on a machine with a local pacman mirror but no general internet,
silently ends with a failed build and no audio.

### It doesn't need to be that way

All of the `PRE_BUILD` work — fetch `sound/hda`, swap in the cut-down Makefiles,
copy the Apple headers, apply the two `cs8409` diffs — is deterministic and can
happen once at package build time. Ship the prepared tree in the package, drop
`PRE_BUILD` entirely, and `dkms.conf` reduces to:

```
MAKE[0]="make KERNELRELEASE=${kernelver}"
AUTOINSTALL="yes"
```

DKMS rebuilds then run `make` and nothing else. I've been running this for a day
on the 14,1; rebuilds verified under `unshare -n` with no network interfaces at
all, and `makepkg` itself is offline too once the sources are vendored.

It's up at **https://github.com/ari-ahm/snd-hda-macbookpro-dkms** if it's useful
for the `omarchy-pkgs` side — GPL-2.0-only like upstream, and mostly just moving
upstream's own steps from `PRE_BUILD` into `prepare()`. Take it, fork it or
ignore it; no attribution needed. Note it vendors its sources, which suits a
distro package but isn't AUR-idiomatic — for AUR you'd fetch the kernel tarball
in `source=()` at build time and still drop `PRE_BUILD`, which keeps the
offline rebuilds.

Doing the patching at build time has a second benefit for a distro: a patch
conflict against a newer kernel surfaces when the package is built, rather than
silently breaking users' audio on a kernel update months later.

### On installing `linux-headers`

Your diagnosis matches what I hit — headers absent, `dkms status` empty, no
error anywhere. Worth knowing that installing them has a side effect on exactly
the machines this issue targets: it makes `macbook12-spi-driver-dkms`, which
`fix-spi-keyboard.sh` installs on every MacBookPro13,x/14,x, start failing on
every kernel update. That package can no longer build (`applespi` is mainlined;
it still includes `asm/unaligned.h`, removed in 6.12), it just isn't visible
while the DKMS hook has zero kernels to iterate. Filed separately as #9831, but
it's coupled to this one: if this lands first, those machines start surfacing a
failed build.

PR #8296, opened 2026-08-25
Register the Chromium native messaging hosts for Brave Origin

scottjones:
Brave Origin is a separate browser rather than a Brave channel — its own package, binary, flags conf and profile root. Omarchy knows that everywhere except the two native messaging host installers, whose profile list stops at `Brave-Browser-Nightly`.

So `omarchy install browser brave-origin` registers the hosts for every Chromium-family profile *except* the browser it just installed — `copy_chromium_flags` calls both installers, but neither writes to `~/.config/BraveSoftware/Brave-Origin`. The extensions load fine, so Copy URL and Download Video look installed; Alt+Shift+L and Alt+Shift+D just reach no host and do nothing. It's exactly the failure `install/user/chromium.sh`'s own comment warns about.

On my machine, after installing Brave Origin through the menu:

```
~/.config/BraveSoftware/Brave-Origin/NativeMessagingHosts/   (empty)
~/.config/chromium/NativeMessagingHosts/                     com.omarchy.copy_url.json, com.omarchy.ytdlp.json
```

This adds Brave Origin's profile root to both installers and re-runs them from a migration so existing installs get repaired. Only the stable channel is listed, since that's the one channel `omarchy-install-browser` installs — beta/nightly can follow the `Brave-Browser` convention if they're ever supported.

Both chromium host tests now assert the Brave Origin manifest is written, and both fail without the installer change. Full `test/shell` run: 203/206 files pass, with the same 3 unrelated failures (`locale-env`, `snapper`, `theme-install-guards`) before and after — they fail on a pristine `quattro` checkout here too.

One thing I left alone: `migrations/1786643346.sh` carries a third, longer copy of the profile list (it adds Edge Beta, Vivaldi, Opera, Helium) and is also missing Brave Origin. Since it's already been run and marked applied for most people, editing it seemed worse than leaving it — happy to fold it in if you'd rather.

### DIFF
--- a/bin/omarchy-install-chromium-copy-url
+++ b/bin/omarchy-install-chromium-copy-url
@@ -16,6 +16,9 @@ browser_dirs=(
   "$HOME/.config/BraveSoftware/Brave-Browser"
   "$HOME/.config/BraveSoftware/Brave-Browser-Beta"
   "$HOME/.config/BraveSoftware/Brave-Browser-Nightly"
+  # Brave Origin is a separate browser, not a Brave channel: its own package,
+  # binary, flags conf and profile root.
+  "$HOME/.config/BraveSoftware/Brave-Origin"
   "$HOME/.config/microsoft-edge"
   "$HOME/.config/microsoft-edge-dev"
 )
--- a/bin/omarchy-install-chromium-ytdlp
+++ b/bin/omarchy-install-chromium-ytdlp
@@ -17,6 +17,9 @@ browser_dirs=(
   "$HOME/.config/BraveSoftware/Brave-Browser"
   "$HOME/.config/BraveSoftware/Brave-Browser-Beta"
   "$HOME/.config/BraveSoftware/Brave-Browser-Nightly"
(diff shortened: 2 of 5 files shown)


zerobearing2:
:+1: seeing same issue with Brave origin.

zerobearing2:
This looks like a duplicate of #7448, which covers the same root cause (missing Brave Origin entries in the native-messaging-host installers' `browser_dirs`), plus yt-dlp and a backfill migration. Left a note there too, but worth repeating here: even with the manifest fix, sideloaded extensions (`--load-extension`) don't auto-bind their keyboard shortcut — `brave://extensions/shortcuts` showed "Not set" for both Copy URL and Download Video despite `was_assigned: true` in Preferences, and required a one-time manual assignment. That's a separate gap neither PR addresses.

dhh:
Thanks for tracking this down! We landed the same fix in #10292, which also covers the Beta and Nightly profile roots.

PR #8299, opened 2026-08-25
Launch Opera webapps as a tab URL instead of --app=

### LINKED ISSUE #8298: omarchy-launch-webapp: Opera ignores --app= so ChatGPT/Grok/X keybinds open the start page
### System details

Dell XPS 14 DA14260, Intel Core Ultra X7 358H, Intel Arc B390 (xe), Omarchy 4.0.1-1, Hyprland 0.56.2. Default browser is Opera 135.0.5973.41-2 (`opera.desktop`) after removing Chromium.

### What's wrong?

Stock webapp keybinds (Super+Shift+A ChatGPT, Super+Shift+Alt+A Grok, Super+Shift+X, YouTube, Calendar, etc.) and `omarchy launch webapp <url>` do not open the requested URL when Opera is the default browser. Opera starts (or the existing window is focused) on the start page / restored tabs instead of ChatGPT, Grok, X, etc.

Non-browser binds are fine (terminal agent, Spotify). Super+Shift+B also works, because it does not use `--app=`.

#### Steps to reproduce

1. Install Opera, remove Chromium, set Opera as the default browser (`xdg-settings get default-web-browser` → `opera.desktop`).
2. Press Super+Shift+A (ChatGPT) or run `omarchy launch webapp https://chatgpt.com`.
3. Opera opens, but not as an app window and not on chatgpt.com.

#### Root cause

`omarchy-launch-webapp` treats `opera*` as a Chromium-family browser that supports `--app=`:

```bash
case $browser in
esac

exec setsid uwsm-app -- $(sed -n 's/^Exec=\([^ ]*\).*/\1/p' {~/.local,~/.nix-profile,/usr}/share/applications/$browser 2>/dev/null | head -1) --app="$1" "${@:2}"
```

That produces `opera --app=https://chatgpt.com`. Opera is Chromium-based, but it does not honor `--app=`. The process is actually started with that flag (observed: `/usr/lib/opera-stable/opera --app=https://grok.com`) while the window still shows Opera's start page. Opera's own forums document the same behavior: `--app=` is ignored and you get a blank/new normal window.

This is distinct from #7034 (Firefox is not on the whitelist at all, so the helper falls through to a missing `chromium.desktop`). Opera *is* on the whitelist, so the launch "succeeds" and silently drops the URL.

#### Suggested fix

For Opera, pass the URL as a positional argument instead of `--app=`:

```bash
opera https://chatgpt.com
```

That opens a tab in the existing Opera window, or starts Opera with that tab if it is not running. `--new-window` also works but stacks extra windows.

A local workaround that rewrites `--app=URL` to a positional URL makes Super+Shift+A / Grok / X open the correct site as a tab. Happy to send a PR against `omarchy-launch-webapp` if that's useful.

Related: #7034, #5889 (Firefox path using `--new-window` because `--app` is Chromium-specific).

Ojisan1:
Closes #8298

## Problem

`omarchy-launch-webapp` treats Opera as a Chromium-family browser that supports `--app=`. Opera ignores that switch and opens the start page / restored tabs instead of the requested URL. Stock binds like Super+Shift+A (ChatGPT), Super+Shift+Alt+A (Grok), and Super+Shift+X therefore do nothing useful when Opera is the default browser.

Observed process: `/usr/lib/opera-stable/opera --app=https://grok.com` while the window showed Opera's start page.

This is distinct from #7034 (Firefox is not on the whitelist at all). Opera *is* on the whitelist, so the launch succeeds and silently drops the URL.

## Fix

Keep `opera*` on the supported-browser list so we do not fall back to `chromium.desktop`. When the default is Opera, pass the URL as a positional argument:

```
opera https://chatgpt.com
```

That opens a tab in the existing window, or starts Opera with that tab. Chromium-family browsers that honor `--app=` are unchanged.

## Testing

- `test/shell.d/launch-webapp-test.sh` — Opera gets a positional URL; Chromium still gets `--app=`
- `./test/cli` passed
- `./test/shell` ran 207 files; the new test passed. Five unrelated files failed in this environment (`omarchy-pkgs` checkout missing, about-window animation, theme-install URL check)
- Manual: Super+Shift+A / Grok / X open the matching site as a tab in one Opera window

### DIFF
--- a/bin/omarchy-launch-webapp
+++ b/bin/omarchy-launch-webapp
@@ -10,4 +10,12 @@ google-chrome* | brave* | microsoft-edge* | opera* | vivaldi* | helium*) ;;
 *) browser="chromium.desktop" ;;
 esac
 
-exec setsid uwsm-app -- $(sed -n 's/^Exec=\([^ ]*\).*/\1/p' {~/.local,~/.nix-profile,/usr}/share/applications/$browser 2>/dev/null | head -1) --app="$1" "${@:2}"
+browser_exec=$(sed -n 's/^Exec=\([^ ]*\).*/\1/p' {~/.local,~/.nix-profile,/usr}/share/applications/$browser 2>/dev/null | head -1)
+
+# Opera is Chromium-based but ignores --app= and opens the start page instead.
+# A positional URL opens a tab in the existing window, or starts Opera with that tab.
+if [[ $browser == opera* ]]; then
+  exec setsid uwsm-app -- "$browser_exec" "$1" "${@:2}"
+else
+  exec setsid uwsm-app -- "$browser_exec" --app="$1" "${@:2}"
+fi
--- a/test/shell.d/launch-webapp-test.sh
+++ b/test/shell.d/launch-webapp-test.sh
@@ -1,0 +1,55 @@
+#!/bin/bash
+
+set -euo pipefail
+
+source "$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd)/base-test.sh"
+
+test_tmp=$(mktemp -d)
+trap 'rm -rf "$test_tmp"' EXIT
+
+mock_bin="$test_tmp/bin"
+test_home="$test_tmp/home"
+mkdir -p "$mock_bin" "$test_home/.local/share/applications"
+
+cat >"$test_home/.local/share/applications/opera.desktop" <<'EOF'
+[Desktop Entry]
+Exec=opera %U
+EOF
+
+cat >"$test_home/.local/share/applications/chromium.desktop" <<'EOF'
+[Desktop Entry]
+Exec=chromium %U
+EOF
+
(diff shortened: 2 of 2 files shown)


cdhananjay:
The current implementation hard-codes `bin/omarchy-launch-webapp` for Opera. Also, the web apps open in a regular Opera window with sidebars, panels, etc., which is essentially the same as opening Opera and visiting the website manually. This doesn't seem like the intended web app experience like the one you get with chromium based browsers.

Ojisan1:
Understood on the product call.

Opera does not support `--app=` (that is the original bug). Chromium is still there for anyone who wants a real app window; this PR does not change that path.

The case this is for: Opera as the daily browser (cross-platform sync) and not splitting webapps into Chromium. Then a tab in Opera is the only thing that actually launches. Edge case, yes. It does not change the out-of-the-box Chromium webapp experience.

Happy to take a different shape if that is preferred.

Issue #7978, opened 2026-08-24
omarchy-hyprland-monitor-scaling persists one shared scale across all displays


Kashfy:
### System details

AMD Ryzen 7 7800X3D, NVIDIA RTX 5070 Ti, Omarchy 4.0.0.alpha, Hyprland 0.56.2. Two external displays, no internal panel: Apple Studio Display (5K) and LG UltraGear 32" (4K).

### What's wrong?

`set_scale` in `bin/omarchy-hyprland-monitor-scaling` applies the new scale to the focused monitor correctly at runtime via `hyprctl eval`. Persistence is where it goes wrong, in two different ways depending on how `monitors.lua` looks.

**Case 1 — stock `monitors.lua`, two displays with different DPI.**

The runtime `hyprctl eval` correctly targets only the focused monitor, but persistence writes a single value:

```bash
sed -i -E -e "s|^local omarchy_monitor_scale = .*|local omarchy_monitor_scale = ${new_scale}|"
```

or, in the `elif` branch, rewrites the `output = ""` catch-all. Both are single-valued and apply to every display. So scaling one monitor looks correct until the next `hyprctl reload` — hotplug, a monitor toggle, anything — at which point both displays snap to the last-set scale.

Steps to recreate:

1. Two connected displays that want different scales (here 5K at 3.2, 4K at 2), with stock `monitors.lua`.
2. Focus display A, `SUPER+/` to change its scale. Looks right.
3. Trigger any `hyprctl reload` — replug a monitor, or let `omarchy-hyprland-monitor-watch` fire.
4. Display B is now also at A's scale.

**Case 2 — `monitors.lua` customized with explicit per-display entries.**

Both `grep` guards fail, so nothing is persisted at all and the scale silently reverts on the next reload. This may be intentional — the comment says "Persist to monitors.lua if the user still has Omarchy's generic catch-all defaults" — but the effect is that once a user configures displays individually, `SUPER+/` becomes runtime-only with no indication of why.

There is also a sharp edge worth noting: the `elif` guard matches the catch-all only in its exact single-line form. Reformatting that block across multiple lines silently changes persistence behavior, with nothing to indicate it.

### Suggested direction

Persist per-display rather than globally, keyed on monitor description since `DP-1`/`DP-2` swap across hotplugs, and fall back to the current single-value behavior only when the config really is untouched.

[...]

PR #8300, opened 2026-08-25
Fix monitor scaling on per-output rules, and three defects the other fixes share

sterre-g:
`omarchy-hyprland-monitor-scaling` applies a scale with `hyprctl eval`, then persists it by rewriting the catch-all `omarchy_monitor_scale` in `monitors.lua`. When the file carries an `hl.monitor` rule naming that output, the catch-all is the wrong place to write, and writing it is actively harmful: Hyprland reloads on the write, the reload re-applies the named rule, and the scale reverts before it is visible. The command reports success either way.

Measured on Omarchy 4.0.1-1 with Hyprland 0.56.2, two displays, `monitors.lua` holding a named rule per output. Pressing SUPER + / five times stepped a 1440p external from 1 to 4 in the audit log while the screen never changed, and setting a scale from the Display panel logged `requested=1 current=4 new=1` three times with the monitor staying on 4. Stepping up stops at 4, which on a 27" 1440p is a 640x360 desktop, so the display it strands on is the one you cannot read.

## This is already reported and already fixed several times

Open reports: #6673, #7242, #7505, #8103, and #7625. #7978 is the closed one.

Open PRs fixing the persistence half in this same file: #7437, #7497, #7519, #7575, #8145 and #8982. Six independent takes on one command, and the persistence branch itself is not what makes this one worth reading.

**The value here is the three defects the others do not cover, and each stands alone.** If you would rather take one of the six, please cherry-pick these onto it and close this.

## What the others do not do

**GDK_SCALE is rewritten from one display's scale while other displays are enabled.** It is a single integer for the session, and Omarchy sets `xwayland.force_zero_scaling`, so it is what sizes every XWayland window. Adjusting a laptop panel to 1.6 next to an external at 1.0 persists `GDK_SCALE=2` and doubles the external's X11 windows. That is the overflow reported in #7021. No value is right for two displays at different scales, so this leaves the configured value alone once more than one output is enabled, counted with `hyprctl monitors all`. Single-display behaviour is unchanged. #7437 has a test asserting the current behaviour, which is worth a look if you take that one instead. #8145's `persist_gdk_scale` rewrites unconditionally.

**A symlinked `monitors.lua` is replaced by a regular file.** Keeping `monitors.lua` in a dotfiles repo and symlinking it is common, and both write paths detach it silently: the scale change works, and every later edit goes somewhere the repo no longer tracks. Third failure in #7625.

**#7309 already fixes half of this and should land.** It adds `sed -i --follow-symlinks` across `bin/` and `migrations/` including this file, with an architectural test that fails any new bare `sed -i`. It does not cover the second path: a rule rewrite that renames a temporary over the target detaches the link just the same, and `--follow-symlinks` cannot help there because no `sed -i` is involved. This branch uses `sed -i --follow-symlinks` in all three of its `sed` calls and `cat` over the target instead of `mv` for the rewrite, so it is already compliant with the rule #7309 wants to enforce and the two merge without fighting. #8145 uses a bare `sed -i` and a temporary rename, so it would need both halves.

**Rule shapes the matcher does not see.** A named rule is matched however its keys are spaced, in the multi-line form nwg-displays writes, and when it names the output by `desc:`. A rule inside a `--[[ ]]` block is treated as commented out rather than matched. A matching rule with no `scale` field gets one appended into the table, not into a trailing comment and not after a semicolon.

## Deliberately not here

**Position.** The `hyprctl eval` still sends `position = "auto"`, which reshuffles a pinned layout until the next reload. #7495 and #7840 own that, and #7326 is the report.

**The other parser.** `omarchy-hyprland-monitor-clamshell` carries its own separate `monitor_rules`/`monitor_rule_regex` pair over the same file format, with the same blindness to `desc:` and multi-line rules. **#8808 fixes it there** and is complementary rather than competing: different file, different function, same class of bug. Both are worth taking, and the fact that `monitors.lua` is parsed independently in two commands is the thing actually worth fixing once someone has the appetite for it.

**#8982** is in this list for completeness rather than as an alternative: it is 190 files and +8562/-555 against an unrebased branch, with the monitor-scaling change inside it.

## Tests

`test/shell.d/monitor-scaling-test.sh` goes from 14 assertions to 28: a named single-line rule updated with other outputs and the catch-all left alone, a commented-out rule ignored, a rule inside a `--[[ ]]` block ignored, the multi-line form nwg-displays writes, a `desc:` keyed rule, a named rule with no scale field getting one, appending past a trailing comment and past a semicolon, GDK_SCALE held with two displays and still tracked with one, and a symlinked `monitors.lua` still being a symlink afterwards.

Each new assertion was checked to be load-bearing by reverting its fix: removing the multi-display guard fails the GDK assertion, `cat` back to `mv` fails the symlink assertion, and disabling the named-rule path fails the persistence assertion.

## 2026-09-01

Rebased onto current `quattro` (b71dcad9) with no conflicts. No code changed in this revision; the corrections are to this description, which had claims that no longer held.

- **Corrected the symlink claim.** The previous revision said the two defects were ones nobody had fixed. That is not true of the `sed -i` half: **#7309** got there on 17 August, eight days before this branch, and covers the whole repository. The section now says what #7309 does, what it does not reach, and why this branch merges cleanly with it.
- **Added #8808** as complementary work on the second parser in `omarchy-hyprland-monitor-clamshell`, and named the duplication as the underlying problem.
- **Added #8982** to the list of takes on the persistence half.
- **Compared against #8145 concretely** on the two axes where this differs, rather than conceding to it in general terms.
- **Corrected the assertion count** to 28, which is what the file has after the later commits on this branch.
- Re-ran `monitor-scaling-test.sh` (28), `monitor-clamshell-scale-test.sh` (25), `monitor-test.sh` (21), `monitor-output-name-test.sh` (8) and `monitor-state-test.sh` (5) on the rebase. All pass.

### DIFF
--- a/bin/omarchy-hyprland-monitor-scaling
+++ b/bin/omarchy-hyprland-monitor-scaling
@@ -71,10 +71,17 @@ normalize_scale() {
   awk 'NR == 1 { printf "%g\n", $0 }'
 }
 
-# Persisting GDK_SCALE happens from three places below.
+# GDK_SCALE is one integer for the whole session, so it can only follow a
+# monitor scale when there is a single monitor to follow. With several outputs
+# enabled, rewriting it here resizes XWayland windows on the displays the user
+# did not touch.
 persist_gdk_scale() {
   local gdk_scale="$1"
   local file="$2"
+  local enabled
+
+  enabled="$(hyprctl monitors -j | jq '[.[] | select(.disabled != true)] | length')"
+  [[ $enabled == 1 ]] || return 0
 
   sed -i -E \
     -e "s|^local omarchy_gdk_scale = .*|local omarchy_gdk_scale = ${gdk_scale}|" \
--- a/bin/omarchy-hyprland-monitor-scaling
+++ b/bin/omarchy-hyprland-monitor-scaling
@@ -144,8 +144,13 @@ update_named_monitor_scale() {
       if (value == monitor) return 1
       # Compared literally rather than as a pattern: a description is EDID free
       # text and routinely carries . ( and +.
-      if (substr(value, 1, 5) != "desc:" || length(value) == 5) return 0
-      return index(description, substr(value, 6)) == 1
+      if (substr(value, 1, 5) != "desc:") return 0
+      # Hyprland trims the selector before comparing, so a rule that spaces one
+      # out names the output there and has to name it here too.
+      value = substr(value, 6)
+      gsub(/^[[:space:]]+|[[:space:]]+$/, "", value)
+      if (value == "") return 0
+      return index(description, value) == 1
     }
     # A --[[ ]] comment runs across lines, so a rule inside one has to be seen
     # as commented out before it is matched: rewriting a scale in there reports
--- a/bin/omarchy-hyprland-monitor-scaling
+++ b/bin/omarchy-hyprland-monitor-scaling
@@ -95,8 +95,10 @@ persist_gdk_scale() {
 # Rewrite `scale = ...` inside the hl.monitor rule naming this output, in both
 # the single-line form Omarchy writes and the multi-line form nwg-displays
 # writes. Commented-out examples and rules for other outputs are left alone. A
-# matching rule with no scale field gets one. Returns 1 when the file carries
-# no rule for this output.
+# matching rule with no scale field gets one. Returns 3 when the file carries no
+# rule for this output, and 1 when the rewrite could not be run at all: a
+# caller that took those for the same thing would answer an unwritable config
+# by falling back to the catch-all and reporting success.
 update_named_monitor_scale() {
   local monitor="$1"
   local scale="$2"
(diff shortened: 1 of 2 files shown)


orospakr:
(authored with help of AI agent)

Tested this branch against my own two-monitor setup (#6673). It fixes the case it targets, but not two shapes that land on the same clobber.

Stubbed `hyprctl monitors -j` with this machine's real output — `eDP-1` 1920x1080 and LG 4K `DP-1` 3840x2160, external focused — then asked for 2x on the external with `local omarchy_monitor_scale = 1.25`. Any run that ends with that variable at 2 has silently rescaled the laptop panel:

| `monitors.lua` shape | eDP-1 variable | External rule |
| --- | --- | --- |
| external named `DP-1` | 1.25 → 1.25 ✅ | → 2 ✅ |
| external matched by `desc:"LG Electronics LG HDR 4K …"` | 1.25 → **2** ❌ | untouched |
| external named `DP-2`, output enumerates as `DP-1` | 1.25 → **2** ❌ | untouched |

`update_named_monitor_scale` matches the literal connector from `.name`, which is always `DP-1`/`eDP-1` and never the `desc:` form. A `desc:` rule doesn't match, the function returns 1, and control reaches the unchanged first branch — `grep -q '^local omarchy_monitor_scale = '` still succeeds and writes the variable, wherever it happens to be bound. A connector-named rule that has gone stale across a renumber lands the same way.

Worth closing because `desc:` is the usual advice for the multi-monitor users this PR is for — connector names are exactly what renumber — and the residual failure is the worse half: not "my change didn't stick" but *a different display* silently resizing.

Two changes would do it:

1. Resolve the focused monitor's `description` alongside its `name` and match `desc:` rules too. #8808 is aimed at this, so these two probably want reconciling rather than landing independently.
2. Require the variable to actually be referenced by a catch-all (`output = ""`) rule before writing it. This is the mis-guard #6673 describes, and it survives unchanged here — my `monitors.lua` has no catch-all at all, yet the `grep` succeeds because the variable feeds a named `eDP-1` rule.

The rest reads well to me, in particular gating `GDK_SCALE` on a single enabled output — it's one session-global integer with no correct value on a mixed 1.25/2 setup.

Happy to send the fixtures as two more cases in `test/shell.d/monitor-scaling-test.sh` if useful — it's a stub `hyprctl` plus the three configs above, and the second and third fail on this branch today.

bjarneo:

PR #8310, opened 2026-08-25
Add workspace compaction command

juan-LARRAYA:
## What

Add `omarchy hyprland workspace compact` to close gaps between occupied numbered workspaces while leaving special workspaces untouched. If the active workspace is renumbered, focus follows it.

## Why

Closing applications can leave gaps in the numbered workspace sequence. Compacting those workspaces currently requires a custom script or manually moving every window.

The command takes one client snapshot, computes the complete move plan in one `jq` pass, and sends all moves through one `hyprctl --batch` call. Its compositor interaction is constant: two state reads plus one batched dispatch instead of one dispatch process per moved window.

## Tests

- `./test/shell.d/hyprland-workspace-compact-test.sh`
- `./test/cli`
- `git diff --check`

The test verifies batching, multiple windows in one workspace, active-workspace focus, no-op windows, and exclusion of special workspaces.

### DIFF
--- a/bin/omarchy-hyprland-workspace-compact
+++ b/bin/omarchy-hyprland-workspace-compact
@@ -1,0 +1,44 @@
+#!/bin/bash
+
+# omarchy:summary=Close gaps between occupied workspace numbers
+
+set -euo pipefail
+
+clients_json=$(hyprctl clients -j)
+active_workspace=$(hyprctl activeworkspace -j | jq -r '.id')
+[[ $active_workspace =~ ^-?[0-9]+$ ]] || exit 1
+
+commands=()
+active_destination=
+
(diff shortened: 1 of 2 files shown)


juan-LARRAYA:
Hi, here is an updated version of this change, re-ported onto the current `quattro` tip with per-monitor compaction, persistent-workspace anchors, dry-run support and a test suite: #12660

https://github.com/omacom/omarchy/pull/12660

Please review the new PR instead of this one. I'm leaving this PR open for reference; happy to close it once #12660 lands.

bjarneo:
Automated duplication check: this pull request looks similar to #7727, which covers the same workspace compact command. I keep that one open and close this one to consolidate review.

PR #8316, opened 2026-08-26
Offer a source-build fallback when Voxtype's binary needs AVX2

### LINKED ISSUE #8312: Dictation (Voxtype) leaves broken install on CPUs without AVX2 instead of failing gracefully
### Summary

Enabling dictation on a CPU that predates the AVX2/x86-64-v3 baseline fails with `FATAL: Illegal CPU instruction (SIGILL)`, but the install script leaves Voxtype installed in a broken state instead of rolling back or cleaning up.

### Steps to reproduce

1. On a machine with a pre-Haswell CPU (no AVX2), enable dictation via Install > AI > Dictation.
2. Try to invoke dictation (Super + Ctrl + X).
3. Voxtype crashes with:
   ```
   FATAL: Illegal CPU instruction (SIGILL)
   ```

### Expected behavior

Since Voxtype's prebuilt binaries target the x86-64-v3 baseline (AVX2, FMA, BMI1/2, F16C, MOVBE), the installer should either:
- detect CPU support for the required instruction set before installing, and refuse/warn up front, or
- build from source for CPUs lacking AVX2 (as Voxtype's own docs suggest, via `-C target-cpu=native`), or
- at minimum, if the installed binary can't run, clean up the failed install (remove the broken binary/service/config) rather than leaving it in place.

### Actual behavior

Voxtype ends up installed but non-functional — running it just SIGILLs every time. The install script doesn't detect the incompatibility or clean up afterward, leaving a broken package on disk.

### Environment

- CPU: Intel Core i7-3520M (Ivy Bridge — no AVX2 support)
- Omarchy version: 4.0.1-1
- Relevant related issues: #4605 (UKI Build Regression: Illegal Instruction on x86-64-v2 hardware — same AVX2 baseline mismatch, different component)

### Additional context

This is likely the same underlying x86-64-v3 vs. x86-64-v2 CPU baseline mismatch seen in #4605, just surfacing through Voxtype/dictation instead of the UKI build tooling. Worth considering a general CPU-baseline check somewhere in the installer given it's now bitten at least two components.

ironbract:
Fixes #8312

## Problem

`voxtype-bin` targets the x86-64-v3 instruction baseline (AVX2, FMA, BMI1/2, F16C, MOVBE, LZCNT), so it `SIGILL`s on CPUs older than Haswell (~2013). The install hook (`bin/omarchy-voxtype-install`) had no capability check, so it always installed the incompatible binary. Worse, if a post-install step failed, `set -e` aborted the script mid-way and left the package + `~/.config/voxtype` on disk with no cleanup.

## Fix

- Add `bin/omarchy-hw-x86-64-v3`, a boolean predicate following the existing `omarchy-hw-*` convention, to detect the baseline via `/proc/cpuinfo` flags.
- Gate the install on that check. When it fails, offer to build the AUR `voxtype` package from source instead of the prebuilt binary, with an explicit warning that this can take 20+ minutes and may run noticeably slower — only proceeds on confirmation.
- Roll back cleanly via the existing `omarchy-voxtype-remove` if any post-install step fails, instead of abandoning a half-installed package.
- Updated `omarchy-voxtype-remove` to drop whichever of `voxtype-bin`/`voxtype` is actually installed.
- Added `test/shell.d/hw-x86-64-v3-test.sh` and `test/shell.d/voxtype-install-test.sh`, and a short note in the manual.

## Testing

`./test/all` passes (208 shell test files + CLI suite), aside from pre-existing failures unrelated to this change (`config-test.sh`, `snapper-test.sh`, `theme-install-guards-test.sh`, `unowned-system-paths-test.sh`, and one CLI test needing a sibling `omarchy-pkgs` checkout not present in this sandbox — all reproduce identically on `quattro` before this change). Also manually verified `omarchy-hw-x86-64-v3` against a real Ivy Bridge CPU (i7-3520M) lacking AVX2.

## Disclosure

This fix was drafted and tested with AI assistance (Claude), in response to the project's own agent-contributing workflow (`default/agents/skills/omarchy/contributing.md`). I filed #8312 after hitting this myself, then worked with Claude to build and test this patch before opening the PR.


ironbract:
Pushed a follow-up fix after testing the source-build fallback on real pre-AVX2 hardware (an Ivy Bridge i7-3520M).

**Bug found:** `voxtype-bin` declares `Provides=voxtype`, so `yay -S --noconfirm --needed voxtype` silently resolved to the repo-provided `voxtype-bin` instead of building the AUR `voxtype` package — completely defeating the fallback and reinstalling the exact binary that SIGILLs on this CPU.

**Fix:** `omarchy-pkg-aur-add` now accepts an optional `repo/` prefix (e.g. `aur/voxtype`) to force resolution from a specific source, while still verifying/erroring against the bare package name pacman actually reports. `omarchy-voxtype-install` now passes `aur/voxtype`.

Re-verified on the same hardware: the source build now genuinely compiles (~20 min) and installs `voxtype` (not `voxtype-bin`), and the resulting `voxtype.service` starts and runs correctly with no SIGILL.

Also added `test/shell.d/pkg-aur-add-test.sh` covering the prefix behavior, and updated `voxtype-install-test.sh`'s assertion accordingly. Full `./test/all` run clean aside from the same pre-existing, unrelated sandbox failures noted before.

(Disclosure: this fix, like the rest of the PR, was developed with AI assistance — Claude Code, reviewed and tested by me on real hardware before pushing.)

twitchax:
Tested this on Apple Silicon (MacBook Air M1, aarch64, Asahi) — filed #8530 for the ARM side before finding this PR.

Good news: `omarchy-hw-x86-64-v3` exits 1 on `uname -m == aarch64`, so this PR already fixes ARM functionally. `voxtype-bin` is `arch=('x86_64')` there (`makepkg` refuses outright), and your fallback routes to `aur/voxtype`, which builds and works.

Two things still bite on ARM:

**1. `hw-x86-64-v3-test.sh` cannot pass on a non-x86 host.** The test overrides `OMARCHY_CPUINFO_PATH` but not `uname -m`, so the machine check rejects the fixture before the flags are read:

```
ok     - Ivy Bridge without AVX2/BMI1/BMI2 does not satisfy x86-64-v3
not ok - Haswell with the full x86-64-v3 baseline satisfies it
```

The three `assert_rejects` cases pass for the wrong reason, too — the arch guard, not the flags. Making the machine type overridable the same way cpuinfo already is fixes both:

```bash
machine=${OMARCHY_UNAME_M:-$(uname -m)}
[[ $machine == "x86_64" ]] || exit 1
```

With `OMARCHY_UNAME_M` set in the test helper, plus a case asserting aarch64 is rejected even with x86 flags present, it goes 5/5 here.

**2. The message is wrong on ARM.** "Your CPU is missing instructions (AVX2/x86-64-v3)" isn't true of an M1 — nothing is missing, there's just no prebuilt binary for the architecture. Same for "may result in noticeably slower dictation": the source build is native there. Branching the copy on `[[ $(uname -m) == "x86_64" ]]` covers both cases.

I have these as two commits on top of your branch, tested on real M1 hardware (`./test/all` goes from 6 failures to 5; the remaining 5 also fail on clean `quattro`). Happy to open them as a PR against `fix-dictation-avx2-fallback`, or you can just take the diff — whichever you prefer.

ironbract:
@twitchax Thanks for testing on real M1 hardware — I'd like to take your fix. Could you open the two commits as a PR against `ironbract:fix-dictation-avx2-fallback`, or push them to a public branch/gist I can pull from? I don't have ARM hardware to verify independently, so I'll take your commits as-is and just confirm they don't regress the x86 path before merging.

ironbract:
Pushed `1d37f811` implementing both points from the review above (the aarch64 test-override and the arch-aware messaging), based on the exact approach described in the comment. I don't have ARM hardware to verify directly, so I've only confirmed it doesn't regress the x86_64 path — full `./test/all` clean aside from the same pre-existing sandbox failures noted earlier (unrelated `omarchy-pkgs` checkout missing). @twitchax if you get a chance to verify this still works on your M1, that'd close the loop — happy to adjust if your actual commits differ from what I derived from the writeup.

ironbract:
Follow-up: I asked for your commits and then, before waiting for a reply, went ahead and pushed my own reconstruction anyway (AI over-enthusiasm on my end — sorry for the mixed signal). What's on the branch now is my best-effort implementation from your writeup, verified only against the x86_64 suite since I don't have ARM hardware.

Your actual commits, tested on real M1 hardware, are still what I'd prefer over my reconstruction — if you push them (or open them as a PR against this branch), I'll swap them in over what's there now.

twitchax:
@ironbract, sorry, let me open them as a PR against your branch, but let me pull the latest first, so we don't conflict.

twitchax:
Verified `1d37f811` on real M1 hardware (MacBook Air M1, aarch64, Asahi, `linux-asahi` 7.1.6). Your reconstruction is correct — no PR needed from me.

`hw-x86-64-v3-test.sh`, run natively on aarch64:

```
ok - Ivy Bridge without AVX2/BMI1/BMI2 does not satisfy x86-64-v3
ok - Haswell with the full x86-64-v3 baseline satisfies it
ok - AVX2 alone without FMA/BMI1/BMI2/F16C/MOVBE does not satisfy x86-64-v3
ok - empty cpuinfo does not satisfy x86-64-v3
ok - aarch64 is rejected even with x86-64-v3 flags present
```

5/5. The Haswell case now genuinely exercises the flags path on ARM rather than being short-circuited by the machine check, which was the actual bug.

Full `./test/all`: 5 of 209 files fail, and all 5 also fail on clean `quattro` on this machine (`bar-icon-geometry`, `config`, `snapper`, `theme-install-guards`, `unowned-system-paths`) — pre-existing and unrelated to this PR.

I diffed your version against my local commits: `bin/omarchy-hw-x86-64-v3` is byte-identical, and the rest differs only in variable naming and prose. Your manual wording is better than mine — it says the non-x86 build has no expected slowdown, which mine left out. So there's nothing for me to swap in; dropping my branch.

One thing I have *not* verified: that the source build actually produces working dictation on ARM. I've confirmed `voxtype` (the AUR source package) declares `arch=('x86_64' 'aarch64')` where `voxtype-bin` is x86_64-only, and that the predicate correctly routes ARM to the fallback — but I haven't run the full install end to end yet. Will report back if I do (but I may not since I don't really need it atm).

peteonrails:
I had trouble getting voxtype to compile on pre-haswell CPUs and didn'thave the hardware at the time to test. I'll revisit shipping a pre-haswell binary to make this faster / easier for the end user.

peteonrails:
Update here - I have figured out how to ship binaries for pre-haswell CPUs. Hopefully that makes things a little simpler. They'll be in voxtype 1.0.1 shipping  (optimistically) tonight. 

ARM binaries will be in there too.

ironbract:
Thanks for the update — great news that 1.0.1 will carry pre-Haswell/pre-AVX2 binaries. I'll watch for the release and test on the same hardware this PR was built against (Ivy Bridge i7-3520M, no AVX2/FMA/BMI1/BMI2). If the new binary works cleanly there, this PR's source-build fallback may no longer be needed — happy to help reconcile once 1.0.1 is out.

(Disclosure: posted with AI assistance — Claude Code.)

ironbract:
Reporting back as promised — thanks for turning 1.0.1 around so quickly, and the ARM builds did land (`voxtype-1.0.1-linux-aarch64-cpu`). The pre-Haswell x86_64 side looks like it didn't make it into the release, though.

Tested on the same Ivy Bridge i7-3520M (`avx` present; `avx2`/`fma`/`bmi1`/`bmi2` absent), installing your 1.0.1 prebuilt directly via `aur/voxtype-bin`. `pacman -Ql` shows the only x86_64 Whisper binaries are `voxtype-avx2` and `voxtype-avx512` — same in v1.1.0-rc1. Everything on the CPU path SIGILLs here: both Whisper variants, both `onnx-avx*`, `voxtype-vulkan`, and all three OSD frontends. The `onnx-cuda`/`rocm`/`migraphx` builds do start, but need GPU runtimes this machine doesn't have and there's no generic `onnx-cpu`, so they're not a way out either. The service ends up crash-looping.

One thing that might be useful to you: the new variant guard can't fire, because it lives inside a binary that won't execute — `voxtype setup variant --status` and `voxtype setup check` both SIGILL. The crash banner points at `setup check` and at symlinking to `voxtype-avx2`, which is already the target and the binary that's crashing, so there's no self-service recovery from inside voxtype on this hardware.

Building 1.0.1 from source on this same machine works and dictation runs fine, so I'd suggest keeping this PR's fallback for now rather than treating 1.0.1 as superseding it. Happy to revisit if a pre-Haswell binary shows up in a later release — and glad to test it here whenever you want a machine to try it on.

(Disclosure: posted with AI assistance — Claude Code.)

ironbract:
Catching up on both reviews, and re-tested against Voxtype 1.1.0-rc4 on the Ivy Bridge machine this PR was built for. Short version: the baseline binary upstream shipped still doesn't retire this PR, and I've now filed that upstream with the reproduction.

**The `pacman -Q` correction was right,** and I reproduced the mechanism here before touching anything — `pacman -Q sh` answers `bash 5.3.15-1` on this machine while `pacman -Qq sh` prints `bash`. That one would have hit every user arriving from #8312, since they all get here with a `voxtype-bin` already installed.

**`4a66f53f` is a good catch and I've kept it.** The drop was run for its effect and not its status, and the failure message that followed asserted the prebuilt was gone on a path where it wasn't. Sending the user to a menu entry that's disabled while `voxtype-bin` is installed made it worse. Stopping before the build is the right call — there's nothing the build could produce that would install while the conflict stands.

**Five commits from me:**

- `e691c740` — `wtype` was installed before the CPU decision, so declining the source build left a package behind on a machine that got no dictation, while the test's own comment claimed that path "installs nothing". It now waits until after the decision, and the declining case asserts that no package is installed at all rather than only that `voxtype-bin` isn't.
- `e403558e` — the predicate omitted LZCNT (`abm`). It's part of the x86-64-v3 baseline and Voxtype's `target-cpu=haswell` builds may emit it, so a CPU or a masked CPUID carrying the other six flags was accepted and could still SIGILL.
- `af1d77be` — after dropping the conflicting prebuilt, a failed build left the user with only yay's output and no idea what state the machine was in. It now says what happened, that the prebuilt is gone and why it was never usable here anyway, and how to retry. Also quotes the `voxtype-bin` literal per AGENTS.md.
- `b12afd8d` — the tests checked which commands each path logged but never what the installer returned, so a path that did the right work and then exited wrong, or the rollback branch reporting success, would have passed. Every case asserts it now.
- `6d69839d` — the manual's flag list never gained LZCNT alongside `e403558e`, so the published copy of the requirement disagreed with the check enforcing it. Good spot.

Each new assertion was checked to fail before its fix, not just pass after it.

Deliberately left alone: the unchecked `hyprctl reload`, the empty `-g` glyph, the wrapper's green "Done!" over a failure, and the build-cache/makedepends cleanup. Those are all pre-existing lines on `quattro` that this PR only moves past — glad to take them somewhere separate rather than widening this one.

`./test/all` at the pushed head: 2692 ok, 4 failing files. Three want a sibling `omarchy-pkgs` checkout, and `theme-install-guards` is this branch's base being old — fixed upstream since our merge-base `0ae16948`.

## Upstream's baseline binary, tested on the target hardware

Worth being precise about, because the release notes now read as though this is solved. peteonrails shipped `voxtype-<version>-linux-x86_64-baseline` starting in 1.1.0-rc2, and the rc3/rc4 notes describe it as resolving the pre-AVX2 problem, "verified on real Ivy Bridge hardware semantics".

Run on real Ivy Bridge hardware (i7-3520M: `avx` and `f16c` present; `avx2`, `fma`, `bmi1`, `bmi2`, `abm`, `movbe` absent), rc4's baseline binary **starts but cannot transcribe**:

```
./baseline --version              exit 0   voxtype 1.1.0
./baseline setup check            exit 0   "All checks passed!"
./baseline transcribe jfk.wav     exit 1   SIGILL  (2/2 runs)
```

It loads the model fine, then takes the SIGILL entering ggml's compute path. Under gdb the faulting instruction is `shrx` — BMI2, which is x86-64-**v3**, while the binary describes itself as built for x86-64-v2. `objdump` says it isn't a stray: 433 BMI2 and 1826 FMA instructions in the baseline build, against 4483 and 2446 in the avx2 one. Identical counts in rc2 and rc4, so nothing moved between them. There's also still no baseline variant of `osd`, `osd-gtk4`, `osd-quickshell` or `audio-bridge` — all avx2-only, all SIGILL here.

Filed upstream with the full reproduction: peteonrails/voxtype#740.

**So the prebuilt path still doesn't cover a pre-Haswell x86_64 machine.** A `voxtype-bin` version bump plus a `_set_default_backend` that understands `baseline` would land this CPU on a binary that passes its own CPU check and then dies on the first transcription — arguably a worse failure than today's, since it fails later and looks healthier on the way in. The source-build fallback in this PR remains the only thing that produces working dictation here, and I've verified that build on this hardware.

Worth crediting on the other side of the ledger: the raw SIGILL is now caught and turned into a banner rather than a bare `Illegal instruction`, and `setup check` / `setup variant` finally run on this CPU because they live in a binary it can execute. The problem I flagged in August, where the variant guard couldn't fire because it sat inside a binary the CPU couldn't start, is genuinely fixed.

## On the standing rebuild cost

The point about `omarchy update` running `yay -Sua --cleanafter` over the fallback's `voxtype` — and so rebuilding it from source, unattended, every time the AUR version moves — is a fair one and I don't want to talk it down. It's the real cost of choosing an AUR source package here, and on the hardware this PR exists for it's a long rebuild.

I'd frame the trade as: this only affects machines that took the fallback, which are machines that otherwise have no working dictation at all, and the user opted into a 20+ minute build knowingly. But if that cost is judged too high to ship, the alternative isn't this PR minus the rebuilds — it's #7890's refusal, and those users get nothing. Happy to follow whichever way you'd rather have it.

The stale-bindings-after-a-failed-build case is also a fair observation. Routing the failure through `omarchy-voxtype-remove` would clear it but also delete a model the user already downloaded, so I left it rather than pick for you. Say the word if you'd prefer it cleaned up and I'll add it.

(Disclosure: posted with AI assistance — Claude Code.)

cristim:
This also fixes #8530 on aarch64. On a MacBook Pro 13" M1 under Asahi Linux, `voxtype-bin` can't install, but the AUR `voxtype` package that this PR's source-build path uses builds and installs cleanly (1.0.1, about 12 minutes). Its Vulkan build runs on the Honeykrisp driver: `base.en` transcribes `samples/jfk.wav` in 0.70 s, against 1.32 s for the native build. Until peteonrails/voxtype#765 lands, Voxtype wrongly flags those builds as unable to run on ARM. I haven't run this PR's script itself on the M1.

*Posted by Claude Opus 5 via Claude Code for @cristim.*

PR #8322, opened 2026-08-26
Menu recalls position on pressing back

Kylar514:
## What

Remembering where you were: opening a submenu used to restart selection at its top, and returning with Left/Backspace dropped you at the top of the parent too — so any trip below the root lost your place. Now every descent records the row it left behind, and Back restores it:

- `activateIndex` stores the child index a `menu`/`link` row was opened
  from, keyed by the parent route. Search results are skipped: their
  indices belong to the flattened result list and mean nothing to the
  parent menu.
- `goBack` hands the remembered index back through `setActiveMenu`, which
  takes an optional `restoreIndex`. Every other caller keeps the old
  top-of-menu default.
- Both open paths clear the memory next to the navigation stack reset, so
  a freshly summoned menu still starts at the top.

Rows that shrink between visits (provider menus re-evaluate their guards) are bounded by the existing post-rebuild clamp, so a stale index can land on the wrong row at worst — never out of range.

## Why

Drilling into Setup > Defaults > Browser and backing out meant finding your place in Setup again by hand. Position memory is the difference between Back meaning "where I was" and "the top, again".

## Notes

- Independent of #8318 (vim-style navigation): same file, disjoint hunks,
  neither blocks the other. Whichever merges second may need a trivial
  rebase of `menu-test.sh`.

## Testing

- `test/shell.d/menu-test.sh` passes (127 assertions), including seven new
  ones covering the property, the filter-guarded recording, restore
  behavior on both back paths, and the open-path resets.
- Manually against the running shell: descend via keyboard and mouse,
  Back returns to the origin row; provider menus clamp safely;
  search-descents record nothing; close/reopen starts fresh.

### DIFF
--- a/shell/plugins/menu/Menu.qml
+++ b/shell/plugins/menu/Menu.qml
@@ -71,6 +71,7 @@ Item {
   property var items: ({})
   property var itemOrder: []
   property var navStack: []
+  property var originIndex: ({})
   property var providersLoaded: ({})
   property var providerQueue: []
   property int providerRevision: 0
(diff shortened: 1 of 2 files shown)


bjarneo:
Automated duplication check: this pull request looks similar to #7074, which covers the same menu goBack selection restore. I keep that one open and close this one to consolidate review.

PR #8336, opened 2026-08-26
Add face authentication (howdy) to the lock screen

krismach:
## What

The lock screen supports fingerprint unlock but not face unlock. This adds a face flow that mirrors the fingerprint one, plus setup and removal wizards:

- `Service.qml` gains a third PAM context, `omarchy-lock-face`, next to the password and fingerprint ones. It arms when `/etc/pam.d/omarchy-lock-face` exists and howdy holds a model for the user.
- Unlike fprintd, a face scan drives the camera and IR emitters, so the PAM context cannot stay armed for the whole lock. The plugin scans up to three times after locking, then goes quiet; any wake (mouse, key) grants a fresh set of scans.
- `LockView.qml` shows a face icon on the left edge of the password field when face auth is set up, matching the fingerprint icon on the right. The dot-reserve logic now covers both icons (`fingerprintReserve` became `iconReserve`).
- `omarchy setup security face` installs howdy, points it at an IR (greyscale) camera by its stable `/dev/v4l/by-id` path, enrolls a model, verifies it with the same matcher PAM runs, and only then writes the PAM files — the same order the fingerprint wizard uses. Plain webcams need explicit confirmation, since a photo can fool them. `omarchy remove security face` undoes all of it.

## Why the stable camera path matters

Howdy configs that point at a bare `/dev/videoN` break silently when a reboot renumbers the nodes — that failure is what led me here. The wizard writes the `by-id` symlink so the config survives renumbering.

## Testing

- New `test/shell.d/lock-face-indicator-test.sh` with fixture, mirroring the fingerprint indicator test; both pass locally.
- `./test/cli` passes.
- Verified live on a BRIO IR camera: lock screen scans on lock and on wake, unlocks on face, falls back to password, and the icon renders in preview and on the real lock surface.

### DIFF
--- a/bin/omarchy-remove-security-face
+++ b/bin/omarchy-remove-security-face
@@ -1,0 +1,32 @@
+#!/bin/bash
+
+# omarchy:summary=Remove face authentication from sudo and lock screen
+# omarchy:requires-sudo=true
+
+set -e
+
+
+remove_pam_config() {
+  if grep -q 'pam_howdy\.so' /etc/pam.d/sudo; then
+    echo "Removing face authentication from sudo..."
+    sudo sed -i '/pam_howdy\.so/d' /etc/pam.d/sudo
+  fi
+}
+
+remove_lock_face_pam() {
+  if [[ -f /etc/pam.d/omarchy-lock-face ]]; then
+    echo "Removing lock screen face authentication..."
+    sudo rm -f /etc/pam.d/omarchy-lock-face
+  fi
+}
+
+
+echo -e "\e[32mRemoving face authentication.\n\e[0m"
+
+remove_pam_config
+remove_lock_face_pam
+
+echo "Removing howdy..."
+omarchy-pkg-drop howdy-git
+
+echo -e "\e[32mFace authentication has been completely removed.\e[0m"
--- a/bin/omarchy-setup-security-face
+++ b/bin/omarchy-setup-security-face
@@ -1,0 +1,102 @@
+#!/bin/bash
+
+# omarchy:summary=Set up face authentication (howdy) for sudo and lock screen
+# omarchy:requires-sudo=true
+
+set -e
+
+
+find_ir_camera() {
+  # IR cameras used for face auth expose a greyscale (GREY) stream. Prefer one
+  # of those; plain RGB webcams work with howdy too, but are easy to fool with
+  # a photo, so they are only offered as an explicit fallback.
+  local dev
+  for dev in /dev/v4l/by-id/*; do
+    [[ -e $dev ]] || continue
+    if v4l2-ctl -d "$dev" --list-formats 2>/dev/null | grep -q "'GREY'"; then
+      echo "$dev"
+      return 0
+    fi
+  done
+  return 1
+}
+
+verify_face() {
+  # compare.py is the exact matcher PAM runs; exit 0 means the camera saw and
+  # matched the enrolled face. Skip the check if the packaging moved it.
+  local compare=/usr/lib/howdy/compare.py
+  [[ -f $compare ]] || return 0
+  sudo python "$compare" "$USER"
+}
+
+setup_pam_config() {
+  # howdy aborts on its own when the lid is closed (abort_if_lid_closed) or in
+  # SSH sessions (abort_if_ssh), so no extra PAM gate is needed here.
+  if ! grep -q pam_howdy.so /etc/pam.d/sudo; then
+    echo "Configuring sudo for face authentication..."
+    sudo sed -i '1i auth      sufficient pam_howdy.so' /etc/pam.d/sudo
+  fi
+}
+
+setup_lock_face_pam() {
+  echo "Configuring lock screen for face authentication..."
+  sudo tee /etc/pam.d/omarchy-lock-face >/dev/null <<'EOF'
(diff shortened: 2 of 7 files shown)


wasdhjklxyz:
image

palprateek:
Independent cross-PR check as someone testing face unlock on Lenovo Slim 3 (Syntek 174f:182b):

I checked this branch against current `origin/quattro` (8f324c90) with `git merge-tree` - it now has 2 conflict markers (`shell/plugins/lock/Service.qml: authenticating` line and `shell/plugins/lock/manifest.json` description), so it needs a rebase.

On backend choice: AUR `howdy` 2.6.1-3 maintainer notes stable is very outdated (recommend -git or visage/biopass), while `facelock-bin 0.2.1-1` (2026-09-07, maintainer tsmith) is actively maintained with TPM/IR enforcement and `facelock pam add --validate`. That's why I see #11612 as the more likely canonical path.

Your stable `/dev/v4l/by-id` camera fix is genuinely valuable and worth cherry-picking into #11612 regardless. Would you be open to retargeting on top of #11612 or collaborating there?

edgseu:
Thanks for putting together this implementation! Adding face unlock to Quickshell is a great addition to the lock screen.

I did an in-depth review of the branch against current `quattro` and noticed a few critical architectural, security, and performance areas that should be addressed:

---

### 1. Security Scope & PAM Configuration
- **Sudo modification (`setup_pam_config`)**: `bin/omarchy-setup-security-face` injects `auth sufficient pam_howdy.so` into `/etc/pam.d/sudo`. A desktop lock-screen feature should strictly confine its scope to session locking. Modifying administrative authentication introduces unexpected security implications and should remain separated.
- **`auth required` in lock PAM**: In `setup_lock_face_pam`, the file `/etc/pam.d/omarchy-lock-face` is written with:
  ```pam
  auth       required                    pam_howdy.so
  ```
  Using `required` means that if face authentication fails, times out, or cannot run, it can break the auth stack or fallback unlock. This should be `auth sufficient pam_howdy.so`.
- **Pre-check module presence**: Before writing any PAM service file, the script should verify that `/lib/security/pam_howdy.so` actually exists on disk. Otherwise, a failed build or missing package can leave a broken PAM configuration behind.

---

### 2. Biometric Data Lifecycle & Teardown
- **Biometric templates left behind**: In `bin/omarchy-remove-security-face`, the packages and PAM files are removed, but the user's enrolled biometric models (`/etc/howdy/models`, `/var/lib/howdy/models`, or `/lib/security/howdy/models`) remain on disk. Biometric templates should be purged on removal (`sudo rm -rf .../models`).
- **Hardware units**: If `linux-enable-ir-emitter` was configured and enabled for IR emitters, the removal script should cleanly disable and stop the service.

---

### 3. Hardware Detection & Verification
- **False-positive camera detection**: Checking `compgen -G "/dev/video*"` matches capture cards, TV tuners, and loopback devices that Howdy cannot drive. Checking sysfs (`/sys/class/video4linux`) for driver bindings like `uvcvideo` or camera/sensor device names avoids false positives.
- **Frame capture verification**: Proving a camera exists doesn't guarantee it can deliver frames (e.g. driver or permission issues). Verifying that the selected device delivers at least one test frame (`v4l2-ctl --device=... --stream-mmap --stream-count=1 --stream-to=/dev/null`) before initiating enrollment ensures users don't get stuck mid-setup.
- **Enrollment verification**: Verify that Howdy actually produced and saved a model on disk before writing PAM service files.

---

### 4. Lock Service Performance (`Service.qml`)
- **Subshell overhead**: `faceCheckProc` runs a `bash -c` subshell with file tests each time status is checked or refreshed. Quickshell supports reactive file watching natively using `FileView` on the PAM config file, which eliminates subprocess forking overhead and uses inotify directly.
- **Attempt timeout**: If Howdy or the camera driver hangs during authentication, there is currently no timeout guarding `facePam`. A timer (e.g., 5s) that invokes `facePam.abort()` ensures the camera isn't kept powered indefinitely if the recognizer stalls.
- **Rate limiting / cooldown**: In `runWake()`, `startFace()` is triggered on wake events. Because mouse movement can fire dozens of wake events rapidly, adding a cooldown timer (e.g., 2 seconds) between face scan attempts prevents thrashing the camera hardware and draining battery.

---

### 5. UI Layout Symmetry & Scanning Feedback (`LockView.qml`)
- **Margin reservation**: `iconReserve` computes `Math.max(fingerprintIcon.implicitWidth, faceIcon.implicitWidth) + 12` and applies it to **both** left and right margins of the input field. If only fingerprint is configured, the left side of the input field still gets padded by `iconReserve`. Docking the biometric indicators together (such as in a `Row` on the right edge, matching Hyprlock / Omarchy conventions) keeps the input field centered and symmetric.
- **Active scanning feedback**: The face icon is currently static. Providing visual feedback (such as status text like `"Scanning face…"`, an active accent color, or a pulsing opacity animation while `faceAuthenticating` is active) gives the user clear confirmation that the camera is scanning.

---

### 6. Base Branch Conflicts
- The branch currently has merge conflicts with `quattro` on `shell/plugins/lock/Service.qml` (authenticating property) and `shell/plugins/lock/manifest.json`. A rebase on latest `origin/quattro` will be needed.

PR #8351, opened 2026-08-26
Add NetBird as an installable service with a bar panel

TechHutTV:
Adds [NetBird](https://netbird.io/) alongside Tailscale as an installable mesh VPN: three service commands, a first-party `omarchy.netbird` bar plugin, and the docs to go with them. NetBird is an open source WireGuard mesh — and commonly self-hosted, which this supports end to end — so this gives the same shape of integration Tailscale already has: install from the menu, a panel that does real work rather than only reporting status, and a clean removal path. Everything below was exercised live against a self-hosted NetBird 0.77.1 deployment (47 peers, published routes and exit nodes, multiple profiles); `./test/all` passes with no new failures.

Screenshot 2026-08-27 at 1 12 25 PM

## Menu

- **Install > Service > NetBird** — `omarchy-install-service-netbird`, disabled once the `netbird` package is present, placed after Tailscale.
- **Remove > Service > NetBird** — `omarchy-remove-service-netbird`, shown only while the package is installed.

Both entries draw NetBird's real mark from the Omarchy icon font at `U+E90A`, flattened from the official art since NetBird publishes no monochrome single-path SVG. The font is package-owned by `omarchy-settings`, so a pulled checkout renders these entries without an icon until that release ships.

## Service commands

- `omarchy-install-service-netbird` asks its questions first — NetBird Cloud, or a self-hosted domain (bare domain or full URL; a non-443 port is kept throughout) — so cancelling leaves nothing behind. It then installs `netbird` from the AUR, registers the service with `netbird service install` (the packaged `netbird@.service` template runs the daemon on `/var/run/netbird/%i.sock` while the CLI talks to `/var/run/netbird.sock`, so the packaged unit can never serve the panel), starts it, runs `netbird up` for the SSO login — deliberately unguarded, since a non-zero exit is the normal still-logging-in path — enables the bar plugin, and installs a web app pointed at the right dashboard: `app.netbird.io` for cloud, your own origin for self-hosted.
- `omarchy-remove-service-netbird` brings the tunnel down, stops the unit before `netbird service uninstall` takes the unit file away, disables the plugin, removes the web app, and drops `netbird` plus the `netbird-debug` package the source build leaves behind.
- `omarchy-installed-service-netbird` (hidden) probes fast first — `systemctl is-active`, then the daemon's command line (`pgrep -f "netbird service run"`, because the CLI is the same binary and must not count as the daemon) — with a `timeout`-bounded `netbird status --json` as the last resort, so the answer never hangs behind the CLI's ~10s gRPC retries.

`omarchy-bar defaults` learned `netbird` next to `dropbox` and `tailscale`, so the widget lands right of the tray on a fresh config when the service is running, and stays off the bar when it isn't.

## Bar panel

`shell/plugins/panels/netbird/` ships as first-party plugin `omarchy.netbird` — left click opens the panel, right click connects or disconnects, middle click refreshes. Inside: connection state with the SSO session counting down (urgent in its last hour — the tunnel reports itself healthy right up until the session lapses), a profile switcher when more than one account exists, a route picker with exit nodes first, a peer browser, and a health block covering management, signal, relay availability, and the WireGuard mode and port. Peer rows carry traffic and last-seen; idle peers stay listed, dimmed — on a NetBird network most peers sit idle until you reach for one. Each list section folds away from its heading and remembers the choice in `~/.local/state/omarchy/settings/netbird.json`; peers and profiles start folded, and a folded heading keeps its summary (`PEERS · 7/47 CONNECTED`). Keyboard: `j`/`k` and arrows move, with headings as cursor stops; `left`/`right` fold and unfold; `enter` activates; `c`/`n`/`d` copy the selected peer's IP, name, or full domain name (only while the peer list is focused and visible); `t` toggles the connection, `r` refreshes, `a` opens the admin console — derived from the management URL the daemon reports, so a self-hosted deployment lands on its own dashboard, non-default port included — and `esc` closes. `refreshIntervalSec` is exposed as a widget setting, defaulting to 30.

The bar icon and panel hero draw NetBird's own mark from its path data in a `Shape` with `WindingFill` — the wings overlap, and odd-even would punch the overlap into a hole — and the menu draws the same mark from the icon font.

### Surviving NetBird's CLI drift

NetBird has renamed and extended this CLI surface across releases, and Omarchy can't assume which version is on the machine. Routes have no JSON anywhere today — `status --json` carries a `networks` field that stays `null` even with published, selected networks (verified live), and `routes list --json` is an unknown flag — so the panel walks the spellings once (`netbird routes list --json`, then `netbird routes list`, then `netbird networks list`), latches onto whichever the CLI accepts, and goes quiet if none do. Route selection appends (`select --append`) because a bare `select` replaces the whole selection. Profiles work the same way: no release ships `profile list --json`, so the panel tries `profile list --show-id` and falls back to the bare padded table, sliced at the header's column offsets since profile names are free-form (`work laptop` is legal) — and switching selects by the unique ID when the listing carries one, because NetBird allows duplicate profile names and refuses to select an ambiguous one. The daemon state is read from whichever field a given release carries, falling back to the management connection, and an explicit state always wins — a daemon that is up but logged out reports `NeedsLogin` (and a lapsed SSO session `SessionExpired`, which is treated the same way, re-arming the login flow), while management simply not being connected means something quite different. Failures are told apart rather than lumped together: an unreachable socket, a socket this user may not touch, a daemon that isn't listening, and a logged-out daemon each say what they are, even when netbird's gRPC retry noise lands on stdout.

`Model.js` holds every parser as plain functions, pinned by 206 assertions in `test/shell.d/netbird-test.sh`; the service check and bar placement are covered in the existing installed-service and config suites.

## DNS override

`omarchy dns <provider>` pins DNS through a NetworkManager `[global-dns-domain-*]` block, and NetworkManager applies that ahead of every other source — including the split DNS a VPN installs. So a machine pinned to Cloudflare or Google can stop resolving the domains NetBird serves while NetBird itself still reports a perfectly healthy connection. That is a genuinely confusing failure, so it's surfaced twice rather than left to be discovered: the panel spots the combination (NetBird carrying enabled nameserver groups plus a non-DHCP system provider) and offers to hand DNS back to DHCP in one click, and the installer prints the same warning up front. Machines on DHCP — the default — see nothing.

## Factory reset

`/var/lib/netbird` is scrubbed next to `/var/lib/tailscale` in the factory baseline, the full reset, and the legacy degraded-state path, so a reset machine doesn't come back still holding its NetBird identity. Ordinary removal deliberately leaves it, matching Tailscale's removal.

## Docs

- `manual/35-networking.md` — install, self-hosted setup, the panel, the DNS caveat, and pick-one-mesh-VPN advice, held to the same weight as the Tailscale section.
- `manual/24-commercial-apps-services.md` — a NetBird section next to Tailscale's.
- `manual/05-the-top-bar.md` — the optional-widget section now covers three services instead of two.
- `shell/plugins/README.md`, `shell/plugins/bar/README.md` — plugin and bar-widget table rows.
- `shell/plugins/panels/netbird/README.md` — features, shortcuts, the exact CLI surface, the self-hosted console derivation, the DNS note, and the icon.
- `default/fonts/omarchy/README.md` — provenance line for the `U+E90A` mark.


bjarneo:
Automated duplication check: this pull request looks similar to #7417, which covers the same NetBird service integration. I keep that one open and close this one to consolidate review.

PR #8353, opened 2026-08-26
Add fzf-style fuzzy fallback to menu and app search

efekurnaz:
## Summary
- Substring/keyword matching already covers exact and whole-word queries, but a query like `lean` won't find `Learn` since it isn't a literal substring. Adds an fzf-style subsequence matcher as the lowest-priority fallback tier in both `MenuModel.js` (root menu) and `AppSearch.js` (Apps submenu).
- Fallback is scoped to the label/app name only (never description or keywords) so it can't reach into longer fields and produce false positives, and it's always ranked below any real substring/whole-word match.
- Documents the new fallback behavior in `docs/menu.md`.

## Test plan
- [x] `./test/shell` passes (`test/shell.d/app-search-test.sh`, `test/shell.d/menu-test.sh` cover the new fuzzy fallback specifically)
- [x] `./test/all` — only pre-existing, environment-specific failures unrelated to this change (missing sibling `omarchy-pkgs` checkout)

### DIFF
--- a/docs/menu.md
+++ b/docs/menu.md
@@ -48,6 +48,16 @@ alternate names users already type (`power-menu`, `settings`), kept for
 compatibility — see `AGENTS.md`. Search does not need them: labels, the last
 id segment, and descriptions are all searchable.
 
+A query that matches nowhere as a substring or whole word falls back to
+fuzzy (subsequence) matching against the label alone: typing `lean` finds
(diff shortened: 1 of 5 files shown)


bjarneo:
Automated duplication check: this pull request looks similar to #9764, which covers the same menu fuzzy search fallback. I keep that one open and close this one to consolidate review.

PR #8362, opened 2026-08-26
Fix: Use relative positioning (auto-right) in multi-monitor defaults to prevent scaling cursor traps

syskey8:
Updates monitor examples to use scale-safe relative positioning, preventing cursor dead zones in multi-monitor layouts.

### Changes:
- **Documentation**: Uses `auto-right` in the default monitor example and agent guidance to prevent users from hardcoding absolute pixel offsets (which breaks under fractional scaling).
- **Runtime Fix**: Strips `position = "auto"` from the `hyprctl eval` runtime injection in `bin/omarchy-hyprland-monitor-scaling`. This ensures that triggering a scale change preserves the user's active layout instead of collapsing it.

*(Note: The persistence parser commit was reverted as parsing Lua with regex is too brittle for edge cases. Leaving the persistence fix to PR #8300 / #7495).*

### DIFF
--- a/bin/omarchy-hyprland-monitor-scaling
+++ b/bin/omarchy-hyprland-monitor-scaling
@@ -86,7 +86,7 @@ set_scale() {
   local new_gdk_scale="$(awk -v scale="$new_scale" 'BEGIN { printf "%d", int(scale + 0.5) }')"
   local monitor_lua="$HOME/.config/hypr/monitors.lua"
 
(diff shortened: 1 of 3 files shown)


bjarneo:
Automated duplication check: this pull request looks similar to #7495, which covers the same monitor scale layout preserve. I keep that one open and close this one to consolidate review.

PR #8365, opened 2026-08-26
[Security] Set up SSH for keys only, and open nothing until a key is in place

### LINKED ISSUE #8363: [Security] omarchy setup security sshd leaves password authentication on, and opens the port before any key is authorized
`omarchy setup security sshd` prints "Setting up SSH server access with key-based authentication", and then starts sshd with the distribution defaults — which accept passwords.

## Password authentication is never turned off

Nothing in the repo writes to `sshd_config`:

```
$ grep -rn "PasswordAuthentication\|PermitRootLogin\|sshd_config" bin install migrations default etc
$
```

OpenSSH's own compiled-in default is `PasswordAuthentication yes`. Arch ships `/etc/ssh/sshd_config.d/99-archlinux.conf`, and it does not change that — it only sets:

```
KbdInteractiveAuthentication no
UsePAM yes
PrintMotd no
```

So after the command finishes, the machine accepts SSH password logins for every local account, from whatever network it is attached to. `ufw limit 22/tcp` rate-limits a single source IP; it does not stop a distributed attempt, and it does nothing about a weak or reused login password — which on a personal desktop is common, and which Omarchy also lets you retry ten times at the sudo prompt (`etc/sudoers.d/omarchy-passwd-tries`).

A command whose entire purpose is to authorize a public key should not leave password authentication reachable.

## The port opens before the key exists

```bash
setup_sshd        # omarchy-pkg-add openssh; systemctl enable --now sshd.service
open_firewall     # ufw limit 22/tcp

echo
if [[ -n $KEY ]]; then
  authorize_key "$KEY" || exit 1
  ...
```

The server is enabled and the firewall opened *first*. Every way the key step can fail happens after that:

- `gum choose` / `gum input` cancelled — `|| exit 1`
- a pasted key that `ssh-keygen -lf` rejects — `authorize_key ... || exit 1`
- `--gh-keys` for a user with no public keys — `exit 1`

Each leaves a listening, network-reachable sshd with password authentication on and **no authorized key at all**, and says nothing about it. The `set -e` at the top means an unexpected failure anywhere in between does the same.

## Reproduce

```
$ omarchy setup security sshd
  (choose "Paste key manually", then press Esc)

$ systemctl is-enabled sshd
enabled
$ sudo sshd -T | grep -i passwordauthentication
passwordauthentication yes
$ ls ~/.ssh/authorized_keys
ls: cannot access '/home/you/.ssh/authorized_keys': No such file or directory
```

## Expected

Password (and keyboard-interactive) authentication off, and nothing listening or opened until a key has actually been authorized.

Related but separate: #8248 asks about which interfaces sshd binds to. This one is about how it authenticates and when it starts.

Chessing234:
Fixes #8363.

Two things, both in what the command leaves behind.

**Password authentication was never turned off.** Nothing in the repo writes `sshd_config`, OpenSSH's own default is `PasswordAuthentication yes`, and Arch's `/etc/ssh/sshd_config.d/99-archlinux.conf` only sets `KbdInteractiveAuthentication no` / `UsePAM yes` / `PrintMotd no`. So a command that announces "key-based authentication" was starting a server that takes password logins from whatever network the machine is on. This writes an `sshd_config.d` drop-in that turns both off — Arch prepends `Include /etc/ssh/sshd_config.d/*.conf` to `sshd_config` in its PKGBUILD, so a drop-in is read — and `omarchy-remove-security-sshd` takes it away again.

**The service and the firewall came first.** `setup_sshd` and `open_firewall` ran before any key was authorized, so a cancelled `gum` prompt, a key `ssh-keygen -lf` rejects, or a `--gh-keys` user with no keys all left a reachable password-only sshd behind with nothing to log in with. The package still has to be installed first because `ssh-keygen` validates the key and ships with it, so the order is now: install → configure key-only → authorize → start → open the port.

I picked `50-omarchy-key-only.conf` for the name: OpenSSH takes the first value it obtains, so a user's own lower-numbered drop-in still wins, and nothing in the shipped set claims `PasswordAuthentication` ahead of it.

## Verification

New `test/shell.d/sshd-setup-test.sh` stubs `sudo`, `ufw`, `ssh-keygen` and `omarchy-pkg-add` and runs the real command:

```
ok - sshd setup turns password authentication off
ok - sshd setup turns keyboard-interactive authentication off
ok - sshd setup writes its config as an sshd_config.d drop-in
ok - sshd setup authorizes the key it was given
ok - sshd is configured key-only before it is started
ok - sshd starts and the port opens only after a key is authorized
ok - a rejected key leaves the ssh server stopped
ok - a rejected key leaves the firewall port closed
```

On the unfixed script the first assertion fails — no drop-in is written at all.

What I could not run here: this is an Arch/OpenSSH path and I have no Arch machine, so the drop-in is verified through the stubs and against the Arch `openssh` PKGBUILD (the `Include` line and the contents of `99-archlinux.conf`), not against a live `sshd -T`. Worth a maintainer confirming `sudo sshd -T | grep -i passwordauthentication` reports `no` after the change.


itz4blitz (commented):
Reviewed this end to end against issue #8363. The ordering is the strong part and I want to call it out: package → key authorized (fail-closed prompts) → drop-in written → enable+reload → **verify with `sshd -T`** → only then open the firewall. Gating reachability on what the daemon actually resolved, rather than on the file having been written, is exactly right — and the test that keeps the port closed when `sshd -T` doesn't confirm key-only proves the gate bites.

A few things I checked and can confirm:

- The Arch comment is accurate: `99-archlinux.conf` only sets `KbdInteractiveAuthentication off`, so `PasswordAuthentication yes` from upstream defaults survives a bare install. The drop-in is the correct mechanism (`sshd_config.d` is `Include`d first on Arch, so `50-` beats `99-`).
- The `authorize_key` `|| return 1` additions are correct bash: errexit is suppressed inside a function invoked to the left of `||`, so each step does have to refuse for itself. The unwritable-`~/.ssh` test covers the case that would otherwise silently "authorize" a key that never landed.
- `confirm_key_only` fails closed on every path I can construct: malformed config makes `sshd -T` error, grep finds nothing, exit 1, port stays closed. A lower-numbered user drop-in overriding ours is also caught because the check asks the daemon, not the filesystem.
- `systemctl reload` after `enable --now` correctly handles "sshd was already running from a previous setup" — without it the daemon would keep its pre-drop-in config until reboot. The test asserts reload happens after the config write.

Two non-blocking nits:

1. On a machine where `sshd.service` failed to start (missing host keys on a fresh container, say), `systemctl reload` errors and `set -e` exits before the confirm — which is the right end state (port closed), but the user sees the raw systemctl error rather than a diagnosis. A `|| { echo hint; exit 1; }` would save a support round-trip.
2. `confirm_key_only` matches `passwordauthentication no` exactly. `sshd -T` prints lowercase on every OpenSSH I have, so this is fine today; if it ever changes, the failure mode is fail-closed (port stays shut), not open, which is the right way for it to break.

This supersedes #7986 in coverage — that one only turned password auth off, this one also fixes the "port open before any key exists" half of #8363.

Chessing234:
the two ordering items are in. leaving the reload-error hint and the sshd -T case match.

Chessing234:
rebased onto quattro; kept the key-before-firewall order. also cleans the legacy 10-omarchy-hardening.conf on remove.

Chessing234:
folded into #12099 — same sshd setup script, one review is enough.

PR #8371, opened 2026-08-26
Save Apple GMUX backlight brightness across reboot

perlausten:
## Summary

Dual-GPU Macs expose the panel backlight as `gmux_backlight` on the PNP bus (`/devices/pnp0/00:00/backlight/gmux_backlight`). systemd's `path_id` builtin fails for that device (`No such file or directory`), so the stock `99-systemd.rules` line never sets `SYSTEMD_WANTS` and `systemd-backlight@backlight:gmux_backlight.service` never starts.

After a reboot the firmware default comes back instead of the brightness the user last set. On a MacBookPro11,3 that default is 184/1023 (**18%**). Keyboard backlight *is* restored, because that LED has a working `ID_PATH`.

This ships a udev rule that tags `gmux_backlight` with `systemd` and `SYSTEMD_WANTS` without requiring `path_id`. `systemd-backlight` still saves under `/var/lib/systemd/backlight/backlight:gmux_backlight` when `ID_PATH` is missing.

## Changes

- `default/udev/gmux-backlight.rules` — the rule (next to the Framework 16 QMK rule)
- `install/hardware/apple/fix-gmux-backlight.sh` — copies it to `/etc/udev/rules.d/` on Apple machines (or any machine that already has `gmux_backlight`), reloads udev, and starts the unit so the *current* session will save on shutdown
- Wired into `install/hardware/all.sh`
- `migrations/1787689809.sh` — same leaf for existing installs
- `test/shell.d/gmux-backlight-test.sh`

## Tested

- MacBookPro11,3, Omarchy 4.0.1, `gmux_backlight` 1023 max
- Before: `udevadm info` showed `TAGS=:seat:` only; `systemd-backlight@backlight:gmux_backlight.service` inactive; no save file under `/var/lib/systemd/backlight/`
- After installing the equivalent rule locally: `SYSTEMD_WANTS` is set, the unit is active, shutdown will save
- `bash test/shell.d/gmux-backlight-test.sh` passes

## Notes

The rule is a no-op on machines without `KERNEL=="gmux_backlight"`. It is still dropped on any `Apple*` DMI vendor so ISO finalization does not depend on `apple_gmux` already having registered the sysfs node.

Related but distinct: #8205 reads `actual_brightness` for gmux OSD steps. This PR is about save/restore across reboot.

### DIFF
--- a/default/udev/gmux-backlight.rules
+++ b/default/udev/gmux-backlight.rules
@@ -1,0 +1,5 @@
+# Apple GMUX backlight is a PNP device. systemd's path_id builtin fails for it
+# ("No such file or directory"), so 99-systemd.rules never sets SYSTEMD_WANTS
+# and brightness is not saved or restored across reboot.
+SUBSYSTEM=="backlight", KERNEL=="gmux_backlight", \
+  TAG+="systemd", ENV{SYSTEMD_WANTS}+="systemd-backlight@backlight:gmux_backlight.service"
--- a/install/hardware/all.sh
+++ b/install/hardware/all.sh
@@ -40,6 +40,7 @@ run_logged "$OMARCHY_INSTALL/hardware/apple/fix-spi-keyboard.sh"
 run_logged "$OMARCHY_INSTALL/hardware/apple/fix-suspend-nvme.sh"
 run_logged "$OMARCHY_INSTALL/hardware/apple/fix-t2.sh"
 run_logged "$OMARCHY_INSTALL/hardware/apple/fix-brcmfmac-supplicant.sh"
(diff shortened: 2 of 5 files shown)


rand0mdud3:
Tested on MacBookPro11,5 (Omarchy 4.x, kernel 7.2.3-arch1-3, gmux max 1023). Before: systemd-backlight@backlight:gmux_backlight inactive, no backlight:gmux_backlight save file. After applying the rule: service active with TAGS=:systemd:seat:, set brightness to 700, clean reboot still 700, save file contains 700. test/shell.d/gmux-backlight-test.sh 11/11 pass. Confirms the fix beyond the author's 11,3.

PR #8374, opened 2026-08-26
[Security] Keep screen-recording state out of world-writable /tmp

### LINKED ISSUE #8372: [Security] Screen recording keeps the in-flight recording path at a fixed /tmp name, and that path drives ffmpeg, mv and rm
`bin/omarchy-capture-screenrecording` keeps the path of the recording in flight at a fixed name in `/tmp`:

```bash
RECORDING_FILE="/tmp/omarchy-screenrecord-filename"
REGION_FILE="${XDG_RUNTIME_DIR:-/tmp}/omarchy-screenrecord-region"
```

The line below it already uses the per-user runtime directory. This one does not.

`/tmp` is world-writable — the sticky bit prevents deleting another user's file, not creating the name first — and the contents of that file are read back on stop and used as a **filesystem path**:

```bash
finalize_recording() {
  latest=$(cat "$RECORDING_FILE" 2>/dev/null)
  [[ -f $latest ]] || return
  ...
  local processed="${latest%.mp4}-processed.mp4"
  if ffmpeg "${args[@]}" "$processed" -loglevel quiet 2>/dev/null; then
    mv "$processed" "$latest"
  ...
}

stop_screenrecording() {
  ...
  local filename=$(cat "$RECORDING_FILE" 2>/dev/null)
  local preview="${filename%.mp4}-preview.png"
  ffmpeg -y -i "$filename" ... "$preview" ...
  omarchy-notification-send "Screen recording saved" ... --exec mpv -- "$filename"
  rm -f "$preview"
}
```

So another local account that owns the name controls, as the recording user:

- where `ffmpeg -y` writes (`<path>-processed.mp4`, `<path>-preview.png` — `-y` overwrites without asking)
- what `mv` replaces (`<path>` itself)
- what `rm -f` deletes
- the path the "Screen recording saved" toast opens in mpv when clicked

`[[ -f $latest ]]` only checks that the target exists, which is exactly the case for a file worth clobbering. `$XDG_RUNTIME_DIR` is 0700 and per-user, so the sibling region file has none of this.

`OMARCHY_SCREENRECORD_DEBUG=true` has the same shape: it appends the recorder's stderr and the picked geometry to `/tmp/omarchy-screenrecord.log`.

Same class as #7904 (FIDO2 authfile staged at a predictable `/tmp` path).

## Expected

Keep the recording state and the debug log beside the region file under `$XDG_RUNTIME_DIR`.

Chessing234:
Fixes #8372.

Recording state, region state, and the optional recorder debug log now use the per-user runtime directory. Without `XDG_RUNTIME_DIR`, they use `${XDG_STATE_HOME:-$HOME/.local/state}/omarchy`; the webcam resizer uses that same fallback so region anchoring remains consistent.

The fallback directory is explicitly mode 0700, including when it already exists with broader permissions. This protects recording paths, geometry and recorder stderr even when XDG_STATE_HOME is traversable by other users. Directory creation and permission repair happen only on recording start, so an unusable directory does not prevent the stop path from signalling an active recorder.

Both troubleshooting documents now give the runtime and fallback debug-log locations.

## Validation

All 26 assertions in `test/shell.d/screenrecording-test.sh` pass, including actual start-path runs with stubbed recorder/compositor commands, runtime and fallback paths, webcam region anchoring, and an existing 0755 fallback repaired to 0700. Tests ran on a local macOS copy with only Bash shebangs adjusted to the installed Bash 5 interpreter; native Arch recording and GUI validation remain outstanding. Bash syntax and git diff checks pass.

### DIFF
--- a/bin/omarchy-capture-screenrecording
+++ b/bin/omarchy-capture-screenrecording
@@ -40,8 +40,11 @@ STOP_RECORDING="false"
 # beside it, `mv` replaces it, and `rm -f` deletes its preview -- so a name in
 # world-writable /tmp is a name any other local account can create first and
 # then point wherever it likes. The debug log carries the recorder's stderr and
-# the picked geometry, which is nobody else's business either.
-RUNTIME_DIR="${XDG_RUNTIME_DIR:-/tmp}"
+# the picked geometry, which is nobody else's business either. Fall back to
+# the state directory when there is no session runtime dir -- :-/tmp would put
+# both files back at the fixed names this comment describes.
+RUNTIME_DIR="${XDG_RUNTIME_DIR:-${XDG_STATE_HOME:-$HOME/.local/state}/omarchy}"
+mkdir -p "$RUNTIME_DIR" || exit 1
 RECORDING_FILE="$RUNTIME_DIR/omarchy-screenrecord-filename"
 REGION_FILE="$RUNTIME_DIR/omarchy-screenrecord-region"
 LOG_FILE=$([[ ${OMARCHY_SCREENRECORD_DEBUG:-false} == "true" ]] && echo "$RUNTIME_DIR/omarchy-screenrecord.log" || echo "/dev/null")
--- a/bin/omarchy-capture-screenrecording
+++ b/bin/omarchy-capture-screenrecording
@@ -15,8 +15,9 @@
 # start.
 #
 # Env: OMARCHY_SCREENRECORD_DEBUG=true appends gpu-screen-recorder's stderr (and
-# the picker target it was launched with) to /tmp/omarchy-screenrecord.log so
-# users can attach a log when reporting capture failures.
+# the picker target it was launched with) to
+# $XDG_RUNTIME_DIR/omarchy-screenrecord.log so users can attach a log when
+# reporting capture failures.
 
 [[ -f ~/.config/user-dirs.dirs ]] && source ~/.config/user-dirs.dirs
 OUTPUT_DIR="${OMARCHY_SCREENRECORD_DIR:-${XDG_VIDEOS_DIR:-$HOME/Videos}}"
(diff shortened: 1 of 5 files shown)


itz4blitz (commented):
The direction is right and the test design is a standout — comparing the `/tmp` name across the run instead of demanding absence, so a leftover from a pre-fix recording can't false-positive, is the correct call.

One substantive problem: the fallback re-opens the hole this PR exists to close.

```bash
RUNTIME_DIR="${XDG_RUNTIME_DIR:-/tmp}"
RECORDING_FILE="$RUNTIME_DIR/omarchy-screenrecord-filename"
```

When `XDG_RUNTIME_DIR` is unset — a recording started from a systemd user unit without the import, a cron/at job, an ssh session before `systemd --user`, or a stale env after logout — the state file lands right back at the predictable `/tmp/omarchy-screenrecord-filename` the PR's own comment calls out: *"a name in world-writable /tmp is a name any other local account can create first and then point wherever it likes."* The comment and the code disagree exactly in the branch where the attack still works.

Since the stop path reads the file back and feeds the path to `ffmpeg` (writes beside it), `mv` (replaces it) and `rm -f` (deletes its preview), a pre-planted symlink at the fallback name is a same-user arbitrary-file clobber or delete on stop — the same primitive #8372 describes.

Suggested resolution, matching the resolution #8370 and #8429 use:

```bash
RUNTIME_DIR="${XDG_RUNTIME_DIR:-${XDG_STATE_HOME:-$HOME/.local/state}/omarchy}"
```

…and `mkdir -p "$RUNTIME_DIR"` before use, or refuse to run when neither is usable. `REGION_FILE` carried the same `:-/tmp` fallback before this PR, so this is a chance to fix both rather than preserve one.

A test for the unset branch would pin it: run the start path with `env -u XDG_RUNTIME_DIR` and a fake `HOME`, then assert no `omarchy-screenrecord-filename` appears at a `/tmp` path and the state file lands under the state dir instead. Happy to contribute that if useful.

Everything else — the debug-log relocation, the docs updates, the stop-path assertions — checks out.

Chessing234:
fell back to the state dir when there is no runtime dir, instead of /tmp.

Chessing234:

ErikMelton:
- [P2] Protect files created in the fallback directory. mkdir -p and the subsequent paths use ordinary umask permissions. Without XDG_RUNTIME_DIR, a traversable XDG_STATE_HOME can leave the directory 0755 and the state/debug files 0644, exposing recording paths, geometry, and recorder stderr to other users. The fallback directory or individual artifacts should be explicitly private (0700/0600).

  - [P2] Document the fallback debug-log location. Both capture.md and contributing.md always direct users to $XDG_RUNTIME_DIR/omarchy-screenrecord.log. When that variable is unset, the implementation instead writes under ${XDG_STATE_HOME:-$HOME/.local/state}/omarchy/, so the troubleshooting instructions point to the wrong file.

Chessing234:
@ErikMelton Addressed both findings in ce438cd0: existing fallback directories are explicitly made 0700, and both docs include the fallback debug-log path. Directory creation also moved to recording start so it cannot block the stop path. All 26 focused assertions pass with Bash 5 compatibility on macOS; native Arch/GUI validation is still outstanding. Could you re-review the update?

ErikMelton (approved):

PR #8381, opened 2026-08-26
Match Apple Touch (MTP) trackpads in omarchy-hw-touchpad

### LINKED ISSUE #8376: omarchy-hw-touchpad misses Apple Touch (MTP) trackpads, so omarchy-toggle-touchpad silently does nothing
### System details

Apple MacBook Air (M2, 2022), Asahi Linux, Omarchy 4.0.0.r6560.gcf105bf, Hyprland 0.56.1 — also verified against a MacBook Air (M1, 2020), which is unaffected

## What happens

On a MacBook Air (M2, 2022) running Asahi, `omarchy-toggle-touchpad` and the
`XF86TouchpadToggle` / `XF86TouchpadOn` / `XF86TouchpadOff` bindings do nothing
at all. The trackpad is never disabled or enabled.

`omarchy-toggle-input-device` is reaching its "no device" branch:

    device="$("omarchy-hw-$KIND")"
    if [[ -z $device ]]; then
      echo "No $KIND device found" >&2
      exit 1
    fi

Because the Hyprland bindings in `default/hypr/bindings/media.lua` run the
command without a terminal, that `echo ... >&2` goes nowhere and the key just
appears dead.

## What should happen

The touchpad toggles, with the usual OSD.

## Cause

`bin/omarchy-hw-touchpad` matches the device by name:

    device=$(hyprctl devices -j | jq -r '[.mice[] | .name | select(test("touchpad|trackpad"; "i"))] | first // empty')

Apple trackpads driven by the Apple Touch (MTP) protocol are exposed by
Hyprland as **`apple-mtp-multi-touch`**, which contains neither `touchpad` nor
`trackpad`, so the match returns empty.

This is model-dependent rather than vendor-dependent, which is what makes it
easy to miss — two MacBook Airs of the same family expose different names:

| host | Hyprland device name | matched by stock? |
|---|---|---|
| MacBook Air M1 (2020) | `apple-spi-trackpad` | yes |
| MacBook Air M2 (2022) | `apple-mtp-multi-touch` | **no** |

So the feature works on one Apple laptop and silently does not on the next one
over, with no error visible to the user.

## Reproduce

On a host whose trackpad is an Apple Touch device:

    $ hyprctl devices -j | jq -r '.mice[].name'
    apple-mtp-multi-touch

    $ omarchy-hw-touchpad
    (prints nothing, exit 0)

    $ omarchy-toggle-touchpad
    No touchpad device found

Verified on 2026-08-25 on a MacBook Air M2 running Asahi: stock
`omarchy-hw-touchpad` printed nothing, and the same query widened by one
alternative printed `apple-mtp-multi-touch`.

## Suggested fix

One alternative in the pattern:

```diff
-device=$(hyprctl devices -j | jq -r '[.mice[] | .name | select(test("touchpad|trackpad"; "i"))] | first // empty')
+device=$(hyprctl devices -j | jq -r '[.mice[] | .name | select(test("touchpad|trackpad|multi-touch"; "i"))] | first // empty')
```

`multi-touch` is specific enough not to catch anything else in `.mice[]` — mice
and pointing sticks do not carry it — and `bin/omarchy-hw-touchscreen` is
unaffected because it reads `.touch[]` / `.tablets[]` rather than matching by
name.

Happy to open a PR if that is the preferred route.

GonzFC:
Fixes #8376.

`bin/omarchy-hw-touchpad` matches the device by name, and Apple trackpads on the
Apple Touch (MTP) protocol are exposed by Hyprland as **`apple-mtp-multi-touch`** —
neither `touchpad` nor `trackpad`. Detection returns empty,
`omarchy-toggle-input-device` takes its "No touchpad device found" branch, and
because `default/hypr/bindings/media.lua` runs the command without a terminal
that error goes nowhere. `XF86TouchpadToggle` just appears dead.

It is model-dependent rather than vendor-dependent, which is what makes it easy
to miss — two MacBook Airs of the same family disagree:

| host | Hyprland device name | before | after |
|---|---|---|---|
| MacBook Air M1 (2020) | `apple-spi-trackpad` | matched | matched |
| MacBook Air M2 (2022) | `apple-mtp-multi-touch` | **empty** | matched |

## Verification

Both real device names, plus devices that must *not* match:

```
apple-spi-trackpad         -> apple-spi-trackpad
apple-mtp-multi-touch      -> apple-mtp-multi-touch
synaptics-touchpad         -> synaptics-touchpad
logitech-mx-master         -> (no match)
tpps/2-elan-trackpoint     -> (no match)
```

`multi-touch` is specific enough not to catch anything else in `.mice[]`, and
`bin/omarchy-hw-touchscreen` is unaffected — it reads `.touch[]` / `.tablets[]`
rather than matching by name.

`./test/cli` passes. `./test/shell` is 202/206 — the same four files fail identically on
unpatched `quattro` on this machine and are environmental, not related to this change:
`config-test`, `snapper-test` and `unowned-system-paths-test` each want an `omarchy-pkgs`
checkout, and `locale-env-test` wants a UTF-8 locale in the bash env.

### DIFF
--- a/bin/omarchy-hw-touchpad
+++ b/bin/omarchy-hw-touchpad
@@ -2,5 +2,5 @@
 
 # omarchy:summary=Print the detected Hyprland touchpad or trackpad device name
 
-device=$(hyprctl devices -j | jq -r '[.mice[] | .name | select(test("touchpad|trackpad"; "i"))] | first // empty')
+device=$(hyprctl devices -j | jq -r '[.mice[] | .name | select(test("touchpad|trackpad|multi-touch"; "i"))] | first // empty')
 [[ -n $device ]] && echo "$device"


bjarneo:
Automated duplication check: this pull request looks similar to #7488, which covers the same touchpad multi-touch regex. I keep that one open and close this one to consolidate review.

PR #8382, opened 2026-08-26
Keep the keyboard backlight level when off runs twice

### LINKED ISSUE #7650: Keyboard backlight stays off after unlock: brightness-keyboard off saves 0 over the good value
### System details

Framework 13 Pro Intel X7

## What happens

After the session locks and I unlock it, the keyboard backlight stays off.
It is not restored until I raise it by hand with `XF86KbdBrightnessUp`.

## What should happen

The backlight returns to the level it was at before the lock. The lock
service already intends this: it blanks with `omarchy-brightness-keyboard off`
and the unlock path calls `omarchy-system-wake`, which runs
`omarchy-brightness-keyboard restore`.

## Cause

`omarchy-brightness-keyboard off` runs (bin/omarchy-brightness-keyboard:29):

    brightnessctl -sd "$device" set 0

`-s` saves the *current* value before setting. Called once from a lit
keyboard it correctly saves the good level. Called a second time while the
backlight is already 0, it saves **0** — destroying the saved value. The
later `restore` (`brightnessctl -rd`, line 32) then faithfully restores
darkness.

`shell/plugins/lock/Service.qml` can blank more than once per lock.
`runBlank()` (line 172) drives `blankProcess` (lines 409-412), and is reached
both at lock time and from `idleBlankTimer` (lines 414-432), which calls
`root.runBlank()` at line 430. `runWake()` (line 167) re-arms that timer
whenever the lock screen is woken, so any wake-then-idle cycle on the lock
screen produces a second blank, and from that point the saved value is 0.

This is a follow-on to #7562, which added the `omarchy-brightness-keyboard
restore` call to `omarchy-system-wake`. That call is present and does run;
the problem is that the value it restores has already been overwritten with
0 by a second `off`, so the restore succeeds and restores darkness.

## Reproduce

Directly, without locking:

    dev=$(basename /sys/class/leds/*kbd_backlight*)
    brightnessctl -d "$dev" set 20
    omarchy-brightness-keyboard off       # saves 20, sets 0   (correct)
    omarchy-brightness-keyboard off       # saves 0,  sets 0   (destroys it)
    omarchy-brightness-keyboard restore   # restores 0

Observed: final brightness 0. Expected: 20.

A single off/restore pair works correctly, which is why this looks
intermittent — it only bites when the lock screen blanks twice.

## Suggested fix

Skip the save when the device is already off, in
`bin/omarchy-brightness-keyboard`:

    if [[ $direction == "off" ]]; then
      if [[ "$(brightnessctl -d "$device" get)" -gt 0 ]]; then
        brightnessctl -sd "$device" set 0
      else
        brightnessctl -d "$device" set 0
      fi
      exit 0
    fi

## Workaround

Pinning brightnessctl's saved state read-only keeps a good value:

    chmod 444 "$XDG_RUNTIME_DIR/brightnessctl/leds/<device>"

The blank still works, its save fails harmlessly, and restore returns a real
level.

## System

- Omarchy 4.0.0-1
- Kernel 7.1.8-arch1-3
- CPU Intel Core Ultra X7 358H, GPU Intel Panther Lake [Arc B390]
- Framework Laptop 13, LED device `chromeos::kbd_backlight` (max_brightness 100)
- brightnessctl 0.5.1-3
- quickshell-git 0.3.0.r20.g28771c7-1
- Hyprland 0.56.2

## Note

`omarchy debug` is listed by `omarchy --help` ("Diagnostics and support logs")
but does not dispatch on 4.0.0-1: `omarchy debug`, `omarchy debug --help` and
`omarchy debug --no-sudo --print` all return "Unknown Omarchy command". System
details above were gathered by hand for that reason.

GonzFC:
Fixes #7650.

`brightnessctl -s` saves the current level into a **single slot**
(`$XDG_RUNTIME_DIR/brightnessctl/leds/<device>`) before zeroing it. Calling
`off` a second time while the backlight is already dark saves `0` over the real
level, and the `restore` that `omarchy-system-wake` runs on unlock then
faithfully restores darkness.

A second `off` is routine rather than exotic: `runBlank()` in
`shell/plugins/lock/Service.qml` is reached at lock time and from
`idleBlankTimer`, and `runWake()` re-arms that timer on every wake while still
locked — so any wake-then-idle cycle on the lock screen produces one.

This guards the save. An already-dark backlight has nothing worth saving, so it
is left alone; `restore` and every other verb are untouched.

## Verification

On a MacBook Air (M1, 2020) with `kbd_backlight`, level 75, running the script
from this branch:

```
off      -> level=0   saved=75
off x2   -> level=0   saved=75
off x3   -> level=0   saved=75
restore  -> level=75
```

Same sequence on `quattro` without the patch, for contrast:

```
off, off, restore -> level=0
```

`./test/cli` passes. `./test/shell` is 202/206 — the same four files fail identically on
unpatched `quattro` on this machine and are environmental, not related to this change:
`config-test`, `snapper-test` and `unowned-system-paths-test` each want an `omarchy-pkgs`
checkout, and `locale-env-test` wants a UTF-8 locale in the bash env.

### DIFF
--- a/bin/omarchy-brightness-keyboard
+++ b/bin/omarchy-brightness-keyboard
@@ -26,7 +26,12 @@ if [[ -z $device ]]; then
 fi
 
 if [[ $direction == "off" ]]; then
-  brightnessctl -sd "$device" set 0 >/dev/null
+  # -s saves the current level into a single slot before zeroing it. Blanking
+  # again while already dark would save 0 over the level restore needs, and the
+  # lock screen can blank more than once per lock.
+  if (( $(brightnessctl -d "$device" get) > 0 )); then
+    brightnessctl -sd "$device" set 0 >/dev/null
+  fi
   exit 0
 elif [[ $direction == "restore" ]]; then
   brightnessctl -rd "$device" >/dev/null


thecdrz:
#8134 already covers this (`Fixes #7650`), including a regression test for the double-`off` save. Same idea: skip `-s` when the backlight is already at 0.

bjarneo:
Automated duplication check: this pull request looks similar to #8134, which covers the same keyboard backlight save guard. I keep that one open and close this one to consolidate review.

PR #8411, opened 2026-08-26
Fix out-of-range group window shortcuts

wxasacoder:
## Problem

Omarchy binds `Super + Alt + 1` through `Super + Alt + 5` directly to `hl.dsp.group.active`.

When the active group contains fewer windows than the selected index, Hyprland returns an `Index out of range` Lua runtime error. For example, with two windows in a group, `Super + Alt + 1` and `Super + Alt + 2` work, but pressing `Super + Alt + 3` produces an error notification. Repeated key presses can fill the screen with these notifications.
screenshot-2026-08-26_22-39-18
## Solution

Resolve the active window and its group when the shortcut is pressed, then dispatch the group switch only when the requested index is within `group.size`.

Valid group window shortcuts retain their existing behavior. Shortcuts targeting unavailable group positions, or used outside a window group, now safely do nothing.

This uses Hyprland's Lua window and group APIs directly instead of catching or matching runtime error messages.

## Testing

- Added a regression test covering valid, out-of-range, and ungrouped window shortcuts.
- Ran `bash test/shell.d/hyprland-group-bindings-test.sh`.
- Ran `bash test/shell.d/hyprland-binding-conflicts-test.sh`.

### DIFF
--- a/default/hypr/bindings/tiling.lua
+++ b/default/hypr/bindings/tiling.lua
@@ -91,7 +91,17 @@ o.bind("SUPER + ALT + mouse_down", "Next window in group", hl.dsp.group.next())
 o.bind("SUPER + ALT + mouse_up", "Previous window in group", hl.dsp.group.prev())
 
 for index = 1, 5 do
-  o.bind("SUPER + ALT + code:" .. tostring(index + 9), "Switch to group window " .. index, hl.dsp.group.active({ index = index }))
+  local group_index = index
+  local dispatcher = hl.dsp.group.active({ index = group_index })
+
+  o.bind("SUPER + ALT + code:" .. tostring(group_index + 9), "Switch to group window " .. group_index, function()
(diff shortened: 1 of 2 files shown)


wxasacoder:
## Testing

Added regression coverage for:

- Switching to a valid group window
- Requesting an index greater than the group size
- Using the shortcut when the active window is not grouped

Focused tests:

- `bash test/shell.d/hyprland-group-bindings-test.sh`

## Manual verification

Tested in a live Omarchy session with two grouped windows:

- `Super + Alt + 1` and `Super + Alt + 2` still switch between the grouped windows normally.
- `Super + Alt + 3`, `Super + Alt + 4`, and `Super + Alt + 5` are ignored without producing Lua runtime errors.

https://github.com/user-attachments/assets/547f2870-e9cc-45f3-9849-3eac3ed7814e

PR #8413, opened 2026-08-26
Add Scanner support

axelfontaine:
This is a different take on #4102 as #4227 (which was rejected).

Instead of including the scanner support in the base packages, we now have a one-click install/remove with auto-configuration for AirScan and Canon Pixma.

This has been working great with my older Canon Pixma MX715, removing the last use-case for which I had to dual-boot into Windows 🎉

@daviewales Could you report back if this works for your scanner as well?

## Summary
- Adds a Hardware submenu under Install/Remove with a Scanner entry that installs Simple Scan
- Defaults to the AirScan (eSCL/WSD) SANE backend for network scanners
- If a Canon PIXMA network scanner is detected via Avahi (BJNP service), configures the pixma backend against it instead, and installs a systemd timer (`omarchy-pixma-ip-sync`) that re-resolves the printer's stable mDNS hostname every 5 minutes so `pixma.conf` stays correct if its DHCP lease changes — without paying mDNS lookup latency on every scan
- Remove tears down the sync timer, the PIXMA config it wrote, and the installed packages

## Test plan
- [x] `./test/cli` — 116/116 passing
- [x] `test/shell.d/menu-test.sh` — 121/121 passing (updated the Remove category ordering assertion to include the new Hardware entry)
- [x] Live-tested install → remove → reinstall against a real Canon PIXMA MX715 on the network: confirmed `pixma.conf` and the sync timer are created correctly, `scanimage -L` and Simple Scan detect the scanner, and removal cleanly undoes every artifact (systemd units, config, packages)


daviewales:
I haven't tested, but I'm sure it would work, as it installs the same packages as my PR (#4227), which was all I needed. (sane-airscan, simple-scan)

But, this looks massively overcomplicated.
An install path which depends on live detection of a running scanner on your network??!!

What if you have multiple scanners, and one is a Pixma, but the other needs airprint? This PR won't install airprint support, because it detects the Pixma on the network! And what if you have a Pixma, but it's turned off when you run the installer, or your not at home, etc?

I feel like this should just install the two required packages and stop. All the extra work for pixma seems unmaintainable. The Pixma configs should either be sent upstream somewhere, or documented in a wiki.

bin/omarchy-install-scanner
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+
+# omarchy:summary=Install Simple Scan with the AirScan (eSCL/WSD) SANE backend, plus Canon PIXMA network detection
+# omarchy:requires-sudo=true
+
+set -euo pipefail
+
+echo -e "\nInstalling scanning support..."
+omarchy-pkg-add sane sane-airscan simple-scan
```
daviewales, line 9: Note that you don't need to explicitly add `sane`, as `sane-airscan` and `simple-scan` will bring it in automatically. (But no harm in adding it explicitly either.)

axelfontaine: Fair point. This is just for symmetry with the remove script.

axelfontaine:
@daviewales Thanks for the feedback. All fair points. Should be much simpler now. There is still a bit of pixma work, but that's just inevitable and should not have no real impact on non-pixma users.

PR #8416, opened 2026-08-26
Refuse a theme name that is shell syntax, and quote the one the unlock picker returns

mdisec:
A theme directory name could carry shell syntax into the **Style → Unlock** action and execute commands as the logged-in user.

However, GitHub and other major Git-based hosting services do not allow arbitrary characters in repository URL paths. Repository names are slugged, and special characters are removed or rejected during creation. Even if an attacker found a way to place such a payload in the theme repository URL, those characters would be URL-encoded, leaving no practical exploitation path that we could identify.

Nonetheless, this represents an important potential security-boundary crossing: remotely controlled data could eventually reach a shell-execution sink. We therefore decided to fix the issue and classify the report as **low severity**.

Reported by Ryan Macy.
Reported by Luis Alvarez.

  ## The chain

  1. `omarchy-theme-install` derives the theme name from the repo URL basename. The only checks were empty / leading dot / contains a slash, so `a';id;'b` passed and became the directory name under
  `~/.config/omarchy/themes`.
  2. `omarchy-plymouth-list` lists that directory (it ships `preview-unlock.png`), and `omarchy-plymouth-switcher` returns the name as-is.
  3. The `style.unlock` menu action interpolated it into a command line:
     `omarchy-launch-floating-terminal-with-presentation "omarchy-plymouth-set-by-theme '$unlock'"`
  4. `omarchy-launch-floating-terminal-with-presentation` embeds its argument in a script and runs it with `bash -c`, so the name is reparsed as shell source.

  Picking that theme in Style > Unlock runs whatever the name carried, as the user.

  The injected characters come from the URL the user pastes, not from the repo's contents, and the name needs characters GitHub does not allow in a repo name — so this needs a self-hosted git remote and a user
  pasting a URL with quotes in it. It does not reach the sudo inside `omarchy-plymouth-set`; the injected command runs before it, unprivileged. But it does defeat the "installing a theme does not get you
  execution" line the codebase already draws, which is why it is worth closing rather than leaving to be quoted case by case.

  ## The fix

  **`omarchy-theme-install` — hold the derived name to the characters a theme name needs.**

  ```bash
  if [[ ! $THEME_NAME =~ ^[a-z0-9][a-z0-9._-]*$ ]]; then
  ```

  An allowlist rather than a list of known harms, so the name stops being dangerous at every place it lands instead of only at the one found. It subsumes the old empty / leading-dot / slash checks — `..` fails
  the leading character class — and the name is already lowercased by the `tr` above it. The leading character excludes `.` and `-`, which keeps the existing `host:-s/foo.git` case covered.

  **`omarchy-menu.jsonc` — quote the picker's answer.** `'$unlock'` becomes `$(printf %q "$unlock")`, for the names already on disk from before this change. Menu actions run under `bash -lc`, so `printf %q` is
  a builtin. This was the only occurrence of the idiom in the file; I swept for both `'$var` and variables inside nested `with-presentation "…"` strings.

  ## Tests

  Extended the two existing files rather than adding new ones.

  `theme-install-guards-test.sh` — five URLs whose derived names carry shell syntax (`a';id;'b`, `$(id)`, backticks, a space, a leading dash) are refused before `git` runs, plus a positive case
  (`tokyo_night.2`) so the allowlist cannot quietly tighten past what real theme names use.

  `plymouth-set-test.sh` — an end-to-end test of the shipped action: it pulls the `style.unlock` action out of `omarchy-menu.jsonc` via `MenuModel.parseMenuJsonc`, stubs the picker to return a hostile name,
  and stands in for the presentation wrapper with the same `bash -c "$*"` shape the real one uses. It asserts both that no injected command ran and that the name arrived as exactly one argv element. Reverting
  the menu line alone fails it. Both ordinary paths — a named theme, and `default` → reset — are covered so the quoting cannot break the feature.

  The wrapper stub could drift from the real wrapper, so the test also greps `omarchy-launch-floating-terminal-with-presentation` for `bash -c "$presentation_script"`: if it ever stops being a shell-string
  API, the test says so instead of silently testing a fiction.

  ## Not changed

  `omarchy-theme-remove` keeps its existing path-climb guard. Its name reaches only a quoted `rm`, never a shell string, and it accepts names for themes already on disk — the same charset would strand a theme
  installed before this change.

### DIFF
--- a/bin/omarchy-theme-install
+++ b/bin/omarchy-theme-install
@@ -23,9 +23,12 @@ omarchy-git-url-check "$REPO_URL" || exit 1
 
 THEMES_DIR="$HOME/.config/omarchy/themes"
 
-# Strip user@host: prefix from scp-style SSH URLs so basename sees just the path
+# Strip user@host: prefix from scp-style SSH URLs so basename sees just the path.
+# git reads a URL as scp-style when a colon appears before any slash, so the path
+# after it need not hold one: `git@host:omarchy-blue-theme.git` is a repo in that
+# user's home, and leaving its prefix on names the theme after the whole URL.
 REPO_PATH="$REPO_URL"
-[[ $REPO_PATH != *"://"* && $REPO_PATH == *:*/* ]] && REPO_PATH="${REPO_PATH#*:}"
+[[ $REPO_PATH != *"://"* && $REPO_PATH == *:* && ${REPO_PATH%%:*} != */* ]] && REPO_PATH="${REPO_PATH#*:}"
 THEME_NAME=$(basename -- "$REPO_PATH" .git | sed -E 's/^omarchy-//; s/-theme$//' | tr '[:upper:]' '[:lower:]')
 THEME_PATH="$THEMES_DIR/$THEME_NAME"
 
@@ -37,7 +40,9 @@ THEME_PATH="$THEMES_DIR/$THEME_NAME"
 # ~/.config/omarchy with it, and one called `a';'id` would carry its own
 # command into that picker. The leading character is kept out of `.` and `-`,
 # which also covers `host:-s/foo.git` leaving basename with `.git`.
-if [[ ! $THEME_NAME =~ ^[a-z0-9_][a-z0-9._+-]*$ ]]; then
+# A bracket range follows the locale's collation, not ASCII: `[a-z]` takes in
+# `é` under en_US.UTF-8. Pin the locale so the set is the one written here.
+if ! (LC_ALL=C; [[ $THEME_NAME =~ ^[a-z0-9_][a-z0-9._+-]*$ ]]); then
   echo "Error: '$REPO_URL' does not give a usable theme name."
   exit 1
 fi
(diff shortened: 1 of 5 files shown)


bin/omarchy-theme-install
```diff
@@ -29,10 +29,15 @@ REPO_PATH="$REPO_URL"
...
-if [[ -z $THEME_NAME || $THEME_NAME == .* || $THEME_NAME == */* ]]; then
+# The name comes from the URL, is joined into a path that is about to be
+# removed, and then names a directory the rest of Omarchy passes around by
+# name: Style > Unlock builds a command line out of the one the picker
+# returned. So it is held to the characters a theme name needs rather than
+# screened for the harm of the day -- a repo called `..` would take
+# ~/.config/omarchy with it, and one called `a';'id` would carry its own
+# command into that picker. The leading character is kept out of `.` and `-`,
+# which also covers `host:-s/foo.git` leaving basename with `.git`.
+if [[ ! $THEME_NAME =~ ^[a-z0-9][a-z0-9._-]*$ ]]; then
```
lalvarezt: this change is technically breaking because you cannot start names with +_- and continue with +. This is not a bad thing per se, but the docs might want to clarify the allowed set of characters as part of the naming convention

lalvarezt (approved):
hi there, thanks for the quick reply and resolution. this looks good to me

rmacy:
Hi this looks great! thanks for getting to it so quickly. I believe a PR I have open (#8519 ) could be complementary.

PR #8421, opened 2026-08-26
omarchy-windows-vm: enable windows activation via system firmware by …

brendank310:
…passing msdm table and smbios information from host to qemu-kvm

Line up the qemu-kvm emulation path to use the host smbios and firmware tables that enable Windows activation for the OEM associated device key.


brendank310:
image

No special steps were taken to activate this Windows install (just `omarchy-windows-vm install` and checking the activation status.

brendank310:
I haven't run a re-test of the rebase (I've got some other changes in the works that I don't want to blow my VM away right now), but should be relatively low risk. I can retest later this weekend, but a test on someone else's hardware would be great :)

PR #8422, opened 2026-08-26
Fix Obsidian window not focusing due to app id mismatch

TijnZeelenberg:
Fixed `SUPER + SHIFT + O` not focusing an already-open Obsidian window.

The bind matched `focus = "^obsidian$"`, but Obsidian's window class/appid
isn't a fixed string — it varies per vault, so it almost never matches.
`initialclass` stays `md.obsidian.Obsidian` regardless of which vault is
open, so this switches the match to that instead.


bjarneo:
Automated duplication check: this pull request looks similar to #7465, which covers the same Obsidian focus pattern. I keep that one open and close this one to consolidate review.

PR #8427, opened 2026-08-26
Disable PSR for Dell U5226KW on i915 to fix black screen after unlock at 120Hz

axelfontaine:
## Summary

On a Dell U5226KW (6144x2560) driven by an Intel i915 GPU, running at 120Hz causes the screen to go black right after unlocking: hyprlock itself renders fine, but the compositor's next full-desktop atomic commit produces nothing, and only repeated DPMS off/on toggling eventually forces a fresh commit that recovers it. Not reproduced at 60Hz.

Root cause: i915's PSR (Panel Self Refresh) exit/wake path is unreliable at this resolution/refresh combination. Disabling PSR via `i915.enable_psr=0` fixes it, confirmed on the affected hardware.

- Adds `omarchy-hw-dell-u5226kw`, a hardware detector that matches only when this exact monitor (via EDID manufacturer/product ID) is connected to a GPU using the i915 driver, following the existing `hw-*` detector convention.
- Adds `install/hardware/intel/fix-dell-u5226kw-psr.sh`, mirroring the existing Panther Lake FRED quirk pattern: drops `i915.enable_psr=0` into `/etc/limine-entry-tool.d/dell-u5226kw-psr.conf` when the detector matches.
- Wires it into `install/hardware/all.sh` alongside the other Intel quirks, so it applies at install and on any later `omarchy apply hardware` rerun (e.g. after plugging the monitor in post-install).

PSR is a module-wide i915 setting with no per-refresh-rate toggle, so this can't be gated to 120Hz specifically — it's a no-op at 60Hz beyond forgoing PSR's power savings on this panel, which was already confirmed working before this change.

## Test plan

- [x] `test/shell.d/hw-dell-u5226kw-test.sh` (new): covers a matching EDID+i915, a non-matching product code, a matching EDID on a non-i915 driver, an empty DRM tree, and the match alongside another connected display
- [x] `./test/cli` passes
- [x] `./test/shell` passes (4 pre-existing unrelated failures on `quattro` — an `omarchy-pkgs` checkout dependency and a network-dependent URL-check test — are unaffected by this change)
- [x] Verified end-to-end on the affected hardware: `i915.enable_psr=0` added via `/etc/default/limine`, rebuilt with `mkinitcpio -P`, rebooted, black-screen-on-unlock at 120Hz no longer reproduces

https://claude.ai/code/session_01Cr7kLhGKQZM73WxJEp3kEk

### DIFF
--- a/bin/omarchy-hw-dell-u5226kw
+++ b/bin/omarchy-hw-dell-u5226kw
@@ -1,8 +1,6 @@
 #!/bin/bash
 
-# omarchy:summary=Detect a Dell U5226KW display connected to an Intel GPU using the i915 driver.
-
-lspci -k 2>/dev/null | grep -A3 -iE 'vga|3d controller' | grep -q 'Kernel driver in use: i915' || exit 1
+# omarchy:summary=Detect a Dell U5226KW display.
 
 drm_path="${OMARCHY_DRM_PATH:-/sys/class/drm}"
 
--- a/bin/omarchy-hw-dell-u5226kw
+++ b/bin/omarchy-hw-dell-u5226kw
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+# omarchy:summary=Detect a Dell U5226KW display connected to an Intel GPU using the i915 driver.
+
+lspci -k 2>/dev/null | grep -A3 -iE 'vga|3d controller' | grep -q 'Kernel driver in use: i915' || exit 1
+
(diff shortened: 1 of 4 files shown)


axelfontaine:
Behavior still not 100% perfect as the screen still sometimes stays black when resuming from suspend. Will resubmit once this is truly solved.

PR #8435, opened 2026-08-26
Show active workspace per monitor

Calamari:
## Problem

The workspace widget used the global `Hyprland.focusedWorkspace` for every bar instance. On a multi-monitor setup, both bars highlighted the same workspace on the focused monitor instead of showing different workspaces for both monitors.

## Solution

- resolve the Hyprland monitor for each live workspace widget instance
- highlight only a workspace that is active on that widget's monitor
- keep single-monitor behavior unchanged

### DIFF
--- a/shell/plugins/bar/widgets/Workspaces.qml
+++ b/shell/plugins/bar/widgets/Workspaces.qml
@@ -1,3 +1,4 @@
+import Quickshell
 import QtQuick
 import QtQuick.Layouts
 import Quickshell.Hyprland
@@ -7,6 +8,10 @@ import qs.Ui
 BarWidget {
   id: root
   moduleName: "omarchy.workspaces"
+
+  readonly property var monitor: root.QsWindow.window !== null
+    ? Hyprland.monitorFor(root.QsWindow.window.screen)
+    : null
 
   function workspaceById(id) {
     var values = Hyprland.workspaces.values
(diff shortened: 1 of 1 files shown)


felipecpaiva:
Heads up on the `QsWindow` route, in case it saves a round trip: it does not resolve here.

On Omarchy 4.0.0 with Quickshell 0.3.1 (Arch), substituting this PR's monitor lookup verbatim logs this on every bar:

```
WARN scene: file://.../Workspaces.qml[21:-1]: TypeError: Cannot read property 'window' of undefined
```

`QsWindow` is undefined on the widget and on every ancestor up to the root item:

```
0: Workspaces              QsWindow=UNDEFINED
1: QQuickLoader            QsWindow=UNDEFINED
2: ModuleSlot              QsWindow=UNDEFINED
3: QQuickRow               QsWindow=UNDEFINED
4: LeftModules             QsWindow=UNDEFINED
5: QQuickItem              QsWindow=UNDEFINED
6: QQuickLoader            QsWindow=UNDEFINED
7: ProxyWindowContentItem  QsWindow=UNDEFINED
8: QQuickRootItem          QsWindow=UNDEFINED
```

`Bar.qml` hints at this already, since `slotWindow()` falls back from `activeItem` to the slot, but here neither carries it.

What did work: `Window.window` returns a distinct `ProxiedWindow` per bar. It exposes no `screen`, but it does sit at its monitor's position, so matching against `Hyprland.monitors` by geometry resolves it. Monitor `x`/`y` are logical while `width`/`height` are physical, so the containment test divides by `scale`.

Two things that only showed up across repeated connect/disconnect cycles, in case they are useful whichever approach lands:

- Workspaces change monitor without `Hyprland.workspaces.values` or `Hyprland.monitors.values` changing identity, so a model built from them has nothing to re-evaluate on and keeps drawing the previous split.
- While an output is being torn down or brought up, a workspace briefly reports no monitor. Treating "unknown" as "belongs to this bar" is how the other screen's workspaces leak onto it.

Tested on two outputs, one of them a headless virtual monitor. Not a request to change the approach, just the data point.

PR #8441, opened 2026-08-26
Pin Cursor password store to gnome-libsecret so GitHub login can use the OS keyring

MrBazzieB:
## Problem

Signing into GitHub from Cursor on Omarchy shows:

Omarchy already runs GNOME Keyring (`org.freedesktop.secrets`). The failure is on the Cursor side: it is an Electron/VS Code fork, and Chromium os_crypt cannot auto-detect GNOME Keyring when `XDG_CURRENT_DESKTOP=Hyprland`.

VS Code already avoids this by writing `"password-store": "gnome-libsecret"` to `~/.vscode/argv.json` in `omarchy-install-editor-vscode`. Cursor was installed through the generic `omarchy-install-and-launch` path and never got that pin.

Related: #1015 (VS Code), and the Chromium `--password-store=gnome-libsecret` pin in `config/chromium-flags.conf`.

## Change

- Add `omarchy-install-editor-cursor`, matching the VS Code installer: package, pin `~/.cursor/argv.json`, apply the Omarchy theme, launch.
- Route Install → Editor → Cursor and `omarchy default editor cursor` through that installer.
- Migrate existing Cursor installs: write or insert `gnome-libsecret` when `password-store` is missing, leave an explicit value (including `basic`) alone.

New installs pick this up from the installer. Existing installs pick it up on the next `omarchy update`. Cursor still has to be restarted once after the pin.

## Test plan

- `bash test/shell.d/cursor-password-store-migration-test.sh`
- `bash test/shell.d/desktop-entry-launch-test.sh`
- `./test/all` — remaining failures are pre-existing on `quattro` in this environment (missing `omarchy-pkgs` checkout, About animation, theme-install URL check)

### DIFF
--- a/bin/omarchy-default-editor
+++ b/bin/omarchy-default-editor
@@ -42,7 +42,7 @@ if omarchy-cmd-missing "$editor"; then
   else
     case "$selection" in
     code) omarchy-install-editor-vscode ;;
-    cursor) omarchy-pkg-add cursor-bin ;;
+    cursor) omarchy-install-editor-cursor ;;
     zed) omarchy-install-editor-zed ;;
     sublime_text) omarchy-pkg-add sublime-text-4 ;;
     helix) omarchy-install-editor-helix ;;
--- a/bin/omarchy-install-editor-cursor
+++ b/bin/omarchy-install-editor-cursor
@@ -1,0 +1,28 @@
+#!/bin/bash
+
+# omarchy:summary=Install Cursor and configure Omarchy defaults for secrets and theme
+
+echo "Installing Cursor..."
+omarchy-pkg-add cursor-bin
+
+mkdir -p ~/.cursor ~/.config/Cursor/User
+
+# Electron cannot auto-detect GNOME Keyring when XDG_CURRENT_DESKTOP=Hyprland,
+# so GitHub login shows "An OS keyring couldn't be identified". Pin the same
+# gnome-libsecret backend VS Code already uses.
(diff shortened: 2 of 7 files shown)


baseman70:
Great addition! The `gnome-libsecret` pin in `argv.json` is definitely needed for GitHub auth in Cursor on Hyprland.

One related issue worth catching here: on Arch Linux, Cursor's built-in updater defaults to a Debian/Ubuntu auto-update flow. On a fresh install, Cursor immediately detects newer upstream releases and silently downloads a `.deb` package (e.g. `cursor_3.19.13_amd64.deb`) into `~/Downloads/`, popping up an "Update downloaded" notification that Arch users cannot install directly.

In `bin/omarchy-install-editor-vscode`, Omarchy prevents this by setting `"update.mode": "none"` in `settings.json`:

```bash
# Ensure VSC's own auto-update feature is turned off
printf '{\n  "update.mode": "none"\n}\n' > ~/.config/Code/User/settings.json
```

We should add the equivalent to `bin/omarchy-install-editor-cursor` so Cursor delegates package updates to `pacman` / `yay`:

```bash
# Ensure Cursor's own auto-update feature is turned off (managed via pacman/AUR)
if [[ ! -f ~/.config/Cursor/User/settings.json ]]; then
  printf '{\n  "update.mode": "none"\n}\n' > ~/.config/Cursor/User/settings.json
fi
```

(And optionally in the migration script if `settings.json` already exists). This would prevent new users from getting confused by unusable `.deb` downloads in their Downloads directory.

PR #8446, opened 2026-08-26
agents: Add Antigravity usage collector

ivanvan08:
Extracts Antigravity (Gemini) usage from local pi, omp, and opencode sessions. Adds the provider to the default agents widget configuration.


bjarneo:
Automated duplication check: this pull request looks similar to #8479, which covers the same Antigravity usage collector. I keep that one open and close this one to consolidate review.

PR #8455, opened 2026-08-26
Keep Ori interactive when launched with a prompt

LukasParke:
`omarchy-agent` launched Ori with `ori code --prompt "$prompt"`, but for Ori a prompt alone means one headless turn whose output is printed only after the turn ends — the window then sits idle instead of dropping into a session.

Passing `--interactive` alongside `--prompt` seeds the session with the prompt and keeps the window interactive, matching how every other supported agent behaves when handed an initial prompt.

- `bin/omarchy-agent`: add `--interactive` to the Ori prompt path (bare `ori code` launch unchanged)
- `test/shell.d/default-agent-test.sh`: update the `assert_launch ori` expectation

Verified with `bash test/shell.d/default-agent-test.sh` — all passing, including "agent launcher adapts initial prompts for every supported agent".

### DIFF
--- a/bin/omarchy-agent
+++ b/bin/omarchy-agent
@@ -92,8 +92,10 @@ omp)
   ;;
 ori)
   # Ori is a harness launcher, and `ori code` is the agent it runs itself.
+  # A prompt alone means one headless turn there, printed after the turn ends,
+  # so --interactive is what seeds the session with it and keeps the window.
   command=(ori code)
-  [[ -n ${prompt:-} ]] && command+=(--prompt "$prompt")
+  [[ -n ${prompt:-} ]] && command+=(--interactive --prompt "$prompt")
   ;;
 pi)
   command=(pi)
--- a/test/shell.d/default-agent-test.sh
+++ b/test/shell.d/default-agent-test.sh
@@ -457,7 +457,7 @@ assert_bypass() {
 assert_launch pi pi "Review this project"
 assert_launch omp omp --auto-approve -- "Review this project"
 assert_launch opencode opencode --auto --prompt "Review this project"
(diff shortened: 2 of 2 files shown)


LukasParke:
Marking draft: `--interactive` is merged in the Ori repo but not yet in any published build. Verified against stable `cli-0.10.2` and the newest prerelease `cli-0.10.2-alpha-886e8fb` (2026-08-26): both reject with `Unrecognized flag: --interactive in command ori code` (exit 1). Landing this before Ori ships the flag would make every prompt launch fail outright instead of running headless. Will mark ready once a release carries it.

LukasParke:
Verified against `cli-0.10.2-alpha-7c8d939` (2026-08-26): `ori code --interactive --prompt …` now opens the TUI, runs the prompt as the first turn (test prompt answered inline), and the session stays interactive afterward. Help text confirms the semantics: `--interactive, -i  Open the interactive TUI with --prompt/--prompt-file text as the first turn instead of running a headless turn`. Marking ready.

LukasParke:
End-to-end verification on a live Omarchy session with the released `cli-0.10.2-alpha-7c8d939`: `omarchy-agent --prompt …` launched the terminal with `ori code --interactive --prompt …`, the window carried `app-id org.omarchy.agent`, the prompt ran as the first turn, and the session stayed interactive afterward. Note the flag currently ships only in Ori alpha builds; stable `cli-0.10.2` rejects it, so this wants to land alongside or after the next stable Ori release.

LukasParke:
@dhh This fixes the "Troubleshoot with AI" Style popups, we just cut a new release with these flags, and I tested it locally on my Omarchy system.

PR #8465, opened 2026-08-26
Keep Funnel ingress relays out of the Tailscale machine list

Unayung:
### What happens now

With Funnel enabled, MACHINES fills up with identical blank rows:

```
MACHINES
  funnel-ingress-node
  funnel-ingress-node
  ... (24 of them on my tailnet)
  a-real-machine
```

Turning on Funnel puts one ingress relay per region into the netmap. They arrive as peers named `funnel-ingress-node`, with no DNS name, no OS, and only an IPv6 address:

```json
{
  "HostName": "funnel-ingress-node",
  "DNSName": "",
  "Tags": ["tag:ingress"],
  "ShareeNode": true,
  "Online": true
}
```

`tailscale status` never shows them — it skips peers with `ShareeNode` set. The panel had no such filter, so every relay counted as a machine.

### The change

`parseStatus` now keeps `ShareeNode` peers out of `peers`, matching the CLI. That flag also covers nodes owned by a shared-to user, which are equally not machines on this tailnet.

The `tag:ingress` ones land in a new `serviceNodes` list, and the panel gives them their own section under MACHINES:

```
MACHINES
  a-real-machine

FUNNEL INGRESS
  24 ingress relays
```

A count rather than a row each: these peers have no hostname, no OS and no address worth copying, so listing them individually would only move the clutter into its own section. The section hides itself when Funnel is off.

### Testing

`test/shell.d/tailscale-test.sh` gains a Funnel relay and a sharee peer in the `parseStatus` fixture, asserting both stay out of the machine list and that only the relay reaches `serviceNodes`. Suite passes.

Verified on a live tailnet with Funnel on: 24 relays collapsed into one line, real machines unaffected.

### DIFF
--- a/shell/plugins/panels/tailscale/Model.js
+++ b/shell/plugins/panels/tailscale/Model.js
@@ -41,6 +41,11 @@ function isMullvadHost(name) {
   return value.length > suffix.length && value.indexOf(suffix) === value.length - suffix.length
 }
 
+function isIngressRelay(peer) {
+  var tags = (peer && peer.Tags) || []
+  return tags.indexOf("tag:ingress") !== -1
+}
+
 function isMullvadPeer(peer) {
   var hostName = String((peer && peer.HostName) || "")
   var dnsName = cleanDnsName((peer && peer.DNSName) || "")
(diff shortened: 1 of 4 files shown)


bjarneo:
Automated duplication check: this pull request looks similar to #6471, which covers the same Tailscale sharee filter. I keep that one open and close this one to consolidate review.

PR #8469, opened 2026-08-26
Stop White's SDDM greeter from painting black-on-black

### LINKED ISSUE #7115: The White unlock theme paints the SDDM login screen black on black
**System details:** Omarchy `quattro` @ 7488eade (post-4.0.0), QEMU/KVM VM. Found by an automated QA pass; reproduced on a disposable worker VM.

Choosing Style → Unlock → **White** gives a login screen that is a solid black rectangle — background, password field, lock glyph and bullets all `#000000`.

`bin/omarchy-plymouth-set:70-73` syncs the SDDM greeter with two `sed` expressions applied in sequence to the same stream:

```sh
sed \
  -e "s/#1a1b26/#$bg_hex/g" \
  -e "s/#ffffff/#$text_hex/g" \
  "$sddm_template" | sudo tee "$sddm_dir/Main.qml" >/dev/null
```

For the `white` theme `bg_hex=ffffff` and `text_hex=000000`. The first expression rewrites the background placeholder to `#ffffff`; the second immediately matches that same text and rewrites it to `#000000`.

`default/sddm/omarchy/Main.qml` contains no `#ffffff` of its own — all greeter text is images — so the second expression is dead code for every theme *except* the one where it lands on the output of the first:

```
$ grep -c 'ffffff' default/sddm/omarchy/Main.qml
0
$ grep -n '1a1b26' default/sddm/omarchy/Main.qml
8:  color: "#1a1b26"
```

Confirmed on a worker:

```
$ omarchy-plymouth-set-by-theme white
$ grep -n "color:" /usr/share/sddm/themes/omarchy/Main.qml | head -1
8:  color: "#000000"          # themes/white/colors.toml has background = "#ffffff"

$ magick /usr/share/sddm/themes/omarchy/lock.png -resize 1x1! -alpha off -format "%[hex:u.p{0,0}]" info:-
000000                        # lock glyph, recoloured to foreground = #000000

# controls, same command:
gruvbox       -> 8:  color: "#282828"    correct
flexoki-light -> 8:  color: "#FFFCF0"    correct, glyph 100F0F
```

Plymouth takes a different path and is correct, so the boot splash is white and only the greeter is dead. `white` is the only shipped theme whose background is `#ffffff`, which is why nothing else shows it.

---

The original overstated it. The greeter is not a solid black rectangle and is not dead: the logo is drawn from an unrecoloured `logo.png` (`default/sddm/omarchy/Main.qml:37`), the focused password field still accepts input, and a failed login shows the red assets. What is black-on-black is the background and the normal-state password controls.

itz4blitz:
mise ~/.config/mise/config.toml tools: gh@2.98.0
## Summary
- White's background is `#ffffff`. `omarchy-plymouth-set` rewrote the SDDM template's `#1a1b26` placeholder to that, then a second `sed` turned leftover `#ffffff` into the text color `#000000`, so the greeter went black-on-black ([#7115](https://github.com/basecamp/omarchy/issues/7115)).
- Background and text now go through unique tokens so those substitutions cannot land on each other.
- `test/shell.d/plymouth-set-test.sh` holds the color contract on the shipped template (White, tokyo, flexoki-light) and kills the original two-pass `sed` plus the other substitutions that reintroduce the same collision.

## Test plan
- [x] `bash test/shell.d/plymouth-set-test.sh`
- [x] Revert the helper to the original two-pass `sed` and watch the White contract fail; restore and watch it pass
- [ ] `omarchy-plymouth-set-by-theme white`, then `grep color: /usr/share/sddm/themes/omarchy/Main.qml` is `#ffffff` (needs sudo / a live theme apply)
- [ ] Log out onto White and confirm the greeter background is white, not black

Closes #7115

### DIFF
--- a/bin/omarchy-plymouth-set
+++ b/bin/omarchy-plymouth-set
@@ -5,6 +5,28 @@
 # omarchy:examples=omarchy plymouth set '#1d2021' '#ebdbb2' ~/.local/state/omarchy/current/theme/plymouth/logo.png
 # omarchy:requires-sudo=true
 
+# Paint the SDDM greeter from the template. The template's background
+# placeholder is #1a1b26 and its (currently unused) text placeholder is
+# #ffffff. Both go through unique tokens so a background of #ffffff cannot be
+# eaten by the text pass, and a text color of #1a1b26 cannot be eaten by the
+# background pass.
+sddm_qml_with_colors() {
+  local bg_hex=$1
+  local text_hex=$2
+  local template=$3
+
+  sed \
+    -e "s/#1a1b26/#__OMARCHY_SDDM_BG__/g" \
+    -e "s/#ffffff/#__OMARCHY_SDDM_TEXT__/g" \
+    -e "s/#__OMARCHY_SDDM_BG__/#$bg_hex/g" \
+    -e "s/#__OMARCHY_SDDM_TEXT__/#$text_hex/g" \
+    "$template"
+}
+
+if [[ ${BASH_SOURCE[0]} != $0 ]]; then
+  return 0
+fi
+
 # Configure the Plymouth boot theme with a custom background color, text color, and logo.
 # Stages the change in a temp dir, then commits the staged files to /usr/share and
 # rebuilds the initramfs. Also syncs the SDDM login screen (the post-logout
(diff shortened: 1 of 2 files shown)


itz4blitz:
Closing: this landed upstream on Aug 28 in 20a23b8c ([Security] Complete Plymouth publication coverage), which adopted the token-substitution approach from this PR and credited it in the commit message (Co-authored-by: itz4blitz). Upstream's version goes further than mine — it folds the substitution into the root-transaction staging pipeline and adds a white-theme regression test that avoids the brittle `no #ffffff` invariant Copilot flagged here. Nothing left to merge. Thanks @afonso.oliveira707 for carrying it in.

PR #8473, opened 2026-08-26
Escape webapp .desktop values per freedesktop spec

bastidotnet:
`omarchy-webapp-install` wrote substituted values (`Name`, `Comment`, `Icon`, `MimeType`, and `Exec`) raw into the generated `.desktop` file. This adds spec-level escaping.

## Security impact

A user-controlled URL containing whitespace was serialized as multiple Desktop Entry `Exec` arguments and `omarchy-launch-webapp` forwarded the extras to Chromium. The live report reproduced this with:

```bash
omarchy-webapp-install InjectTest 'https://example.com/ --user-agent=INJECTION_PROOF_MARKER_12345' chromium
```

The Chromium process received the injected flag. More consequential substitutions include `--load-extension` and `--remote-debugging-port`.

This PR serializes the default URL as one correctly quoted Desktop Entry argument. It also prevents raw newlines and other control characters in generated values from injecting additional keys. #7984 separately keeps the app name inside the applications directory, and #8496 restricts URLs to HTTP(S) and rejects raw whitespace before any network or icon side effect.

## What changes

Two escapers are applied field-selectively:

- Desktop Entry string escaping on every substituted field. Backslash, tab, carriage return, line feed, and a leading space are escaped according to the Desktop Entry string rules.
- Exec argument quoting for the default URL. The URL is emitted as one double-quoted argument, including the required escaping for reserved characters and literal percent signs.

`CUSTOM_EXEC` deliberately receives only Desktop Entry string escaping. It is a complete command line by design; quoting it as one data argument would break the Sunshine launcher.

## Verification

- The generated desktop entry has exactly one escaped key per field.
- `gio launch` passes dollar signs and percent-encoded URLs to `omarchy-launch-webapp` as one unchanged argument.
- A backslash in the name remains a valid GLib desktop-entry value.
- Sunshine's custom command remains byte-identical.
- The combined #7984 + #8473 + #8496 regression suites pass.

### DIFF
--- a/bin/omarchy-webapp-install
+++ b/bin/omarchy-webapp-install
@@ -65,6 +65,26 @@ fetch_site_icon() {
     download_icon "https://www.google.com/s2/favicons?domain=${site_url}&sz=256" "$dest"
 }
 
+desktop_string_escape() {
+  # Desktop Entry "string" value (freedesktop Desktop Entry Spec, "Value types"):
+  # a raw newline would start a new key line and let a value inject a second
+  # Exec=. Escape backslash first, then tab/CR/LF and a leading space. Every value
+  # written into the .desktop file passes through here.
+  printf '%s' "$1" \
+    | sed -e ':a;N;$!ba' \
+          -e 's/\\/\\\\/g' -e 's/\t/\\t/g' -e 's/\r/\\r/g' -e 's/\n/\\n/g' -e 's/^ /\\s/'
+}
+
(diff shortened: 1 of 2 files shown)


ryanrhughes (approved):
Reviewed individually and in the combined #7984 + #8473 + #8496 tree. Spec-level escaping, GLib launch round-trips, exact injection reproduction, focused tests, and baseline comparison all pass.

PR #8493, opened 2026-08-27
omarchy-hook: run executable hooks via their own shebang

### LINKED ISSUE #8492: omarchy-hook runs every hook via bash, ignoring the shebang — silently hangs on non-bash hooks that collide with a real binary name
## What happened

`omarchy-update` hung indefinitely at the `theme-set` step of migration `1787481315` ("Re-stage the current theme..."). Process tree at the time it was stuck:

```
omarchy-update
 └─ omarchy-migrate
     └─ migrations/1787481315.sh
         └─ omarchy-theme-refresh
             └─ omarchy-theme-set ethereal
                 └─ omarchy-hook theme-set ethereal
                     └─ bash omarchy-intellij-theme-sync.py ethereal   (PID in do_wait — blocked)
                         └─ import colorsys                            (PID in poll_schedule_timeou — blocked on X/Wayland input)
```

`~/.config/omarchy/hooks/theme-set.d/omarchy-intellij-theme-sync.py` is a hook installed by the (third-party, Omarchy-plugin-distributed) `fchtngr.intellij-theme-sync` plugin. It's a Python 3 script with a correct `#!/usr/bin/env python3` shebang and is installed `chmod 755` (`install -Dm755 ...`), so it's directly executable.

## Root cause

`omarchy-hook` invokes every hook unconditionally with `bash`, ignoring the executable bit and shebang:

```bash
# /usr/share/omarchy/bin/omarchy-hook
if [[ -f $HOOK_PATH ]]; then
  bash "$HOOK_PATH" "$@" || echo "Hook failed: $HOOK_PATH"
fi

if [[ -d $HOOK_DIR ]]; then
  for hook in "$HOOK_DIR"/*; do
    [[ -f $hook ]] || continue
    [[ $hook == *.sample ]] && continue
    bash "$hook" "$@" || echo "Hook failed: $hook"
  done
fi
```

When `bash` parses the Python file as shell script, it hits the line:

```python
import colorsys
```

Bash treats `import` as a command name — and `/usr/bin/import` happens to exist (a `magick`/ImageMagick symlink, the interactive screen-capture tool). Bash executes `import colorsys`, which is ImageMagick's `import` waiting for the user to interactively click a window/region to capture, saving it as a file named `colorsys`. That process never returns on its own.

Because `omarchy-hook`, `omarchy-theme-set`, `omarchy-theme-refresh`, the migration script, and `omarchy-migrate` all run synchronously (several with `set -e`) and wait on their children, the entire `omarchy-update` blocks indefinitely on this single hook — with no error, timeout, or indication of what it's waiting on.

Even after killing the stray `import` process, the actual hook logic never runs, since bash aborts by the point Python's real syntax (f-strings, type hints, dict literals, etc.) causes further shell parse errors — so the intended effect of the hook (regenerating the JetBrains theme files) silently doesn't happen.

## Why this isn't just a plugin bug

The plugin does everything right: ships a `+x` Python script with a proper shebang, installed via `install -Dm755`. `omarchy-hook` is what discards that by hardcoding `bash` instead of respecting how the file says it should be run. Any hook that isn't a bash script is subject to this — the ImageMagick collision is what turns this particular file into a silent *hang* instead of a fast, visible failure, but the underlying dispatch bug applies to any non-bash hook.

## Suggested fix

Dispatch based on the hook's executable bit / shebang instead of forcing `bash`, e.g.:

```bash
run_hook() {
  local hook=$1; shift
  if [[ -x $hook ]]; then
    "$hook" "$@"
  else
    bash "$hook" "$@"
  fi
}
```

(falling back to `bash` only for non-executable plain shell files, to stay backward compatible with any hooks that currently rely on that).

## Reproduction

1. Install a `theme-set.d` hook that is a `chmod +x` script in a non-bash language with a proper shebang (e.g. Python via the `fchtngr.intellij-theme-sync` plugin, `omarchy plugin install fchtngr.intellij-theme-sync`).
2. Trigger a theme change or run `omarchy-update` far enough to reach a migration that calls `omarchy-theme-refresh`/`omarchy-theme-set` (currently migration `1787481315`).
3. Observe the update hang. `ps -eo pid,ppid,stat,wchan,cmd --forest` shows a `bash <hook>.py ...` process blocked in `do_wait` with a child literally named after the first import in the file (in this case `import colorsys`, resolving to ImageMagick's `import`, blocked in `poll_schedule_timeou` waiting for interactive screen-capture input).

## Environment

- Omarchy: 4.0.1-1
- Kernel: 7.1.8-arch1-3
- Plugin: `fchtngr.intellij-theme-sync` v4.3.0 (installed via `omarchy plugin install`)

hbabb:
Fixes #8492

## Problem

`omarchy-hook` runs every hook with `bash "$hook" "$@"`, unconditionally, ignoring the hook's executable bit and shebang.

A hook written in anything other than bash gets parsed as shell instead of run through its own interpreter. Usually that just fails fast and loud — but a `theme-set.d` Python hook shipped by a plugin (`fchtngr.intellij-theme-sync`) contains the line `import colorsys`. Bash treats `import` as a command name, and `/usr/bin/import` happens to exist — ImageMagick's interactive screen-capture tool. Bash executes it, and it blocks waiting for the user to click a screen region to capture.

Because `omarchy-hook` → `omarchy-theme-set` → `omarchy-theme-refresh` → the migration script → `omarchy-migrate` → `omarchy-update` all run synchronously, that one hook hangs the entire update indefinitely, with no error and no indication of what it's waiting on.

## Fix

Run a hook directly (so the kernel honors its shebang) when it's executable; fall back to `bash "$hook"` only when it isn't. This keeps the shipped `.sample` hooks working exactly as before (they install without the executable bit), while letting a hook in any language run the way it declares it should.

## Testing

- Added `test/shell.d/hook-shebang-test.sh`: an executable hook with a stub shebang interpreter runs through that interpreter (not bash), a non-executable plain hook still runs under bash, and a failing executable hook still reports `Hook failed: ...` without aborting the run.
- `./test/shell` — same 5 pre-existing failures as a clean checkout (`config-test`, `runtime-smoke-test`, `snapper-test`, `theme-install-guards-test`, `unowned-system-paths-test`), all unrelated to `omarchy-hook` and reproducing identically without this change (missing `omarchy-pkgs` checkout / live-compositor / network-guard environment dependencies).
- `test/shell.d/bin-style-test.sh` passes.

### DIFF
--- a/bin/omarchy-hook
+++ b/bin/omarchy-hook
@@ -15,14 +15,33 @@ HOOK_PATH="$HOME/.config/omarchy/hooks/$1"
 HOOK_DIR="$HOOK_PATH.d"
 shift
 
+# Run a hook through its own interpreter when it is executable, so a hook
+# written in something other than bash (Python, etc.) is invoked via its
+# shebang instead of being force-fed to bash. Bash parses a non-bash script as
+# shell and can silently misfire on any line that happens to name a real
+# binary (a Python hook's "import foo" line running ImageMagick's "import",
+# for instance) instead of failing loudly. Non-executable hooks keep running
+# under bash, matching the shipped .sample hooks, which ship without the
+# executable bit.
+run_hook() {
+  local hook=$1
+  shift
+
+  if [[ -x $hook ]]; then
+    "$hook" "$@"
+  else
+    bash "$hook" "$@"
+  fi
+}
+
 if [[ -f $HOOK_PATH ]]; then
-  bash "$HOOK_PATH" "$@" || echo "Hook failed: $HOOK_PATH"
+  run_hook "$HOOK_PATH" "$@" || echo "Hook failed: $HOOK_PATH"
 fi
 
 if [[ -d $HOOK_DIR ]]; then
   for hook in "$HOOK_DIR"/*; do
     [[ -f $hook ]] || continue
     [[ $hook == *.sample ]] && continue
-    bash "$hook" "$@" || echo "Hook failed: $hook"
+    run_hook "$hook" "$@" || echo "Hook failed: $hook"
   done
 fi
--- a/test/shell.d/hook-shebang-test.sh
+++ b/test/shell.d/hook-shebang-test.sh
@@ -48,6 +48,35 @@ actual:   $ran_args"
 pass "executable hook runs through its own shebang instead of bash"
 rm -f "$shebang_hook"
 
+# The flat ~/.config/omarchy/hooks/<name> file is a second dispatch site, and
+# reverting it alone leaves every .d assertion above green.
+flat_marker="$test_tmp/flat-ran"
+flat_stub="$test_tmp/flat-stub-interpreter"
+cat >"$flat_stub" <<SH
+#!/bin/bash
+printf '%s\n' "\$*" >"$flat_marker"
+SH
+chmod +x "$flat_stub"
+
+flat_hook="$hooks_dir/theme-set"
+cat >"$flat_hook" <<HOOK
+#!$flat_stub
+this is not valid bash: {{{ import fakestuff ;;; }}}
+HOOK
+chmod +x "$flat_hook"
+
+run_hook_command theme-set ethereal >"$test_tmp/flat.out" 2>&1
+
+[[ -f $flat_marker ]] || fail "executable flat hook runs through its own shebang" "no marker written; output:
+$(cat "$test_tmp/flat.out")"
+read -r flat_path flat_args <"$flat_marker"
+[[ $flat_path == $flat_hook ]] || fail "executable flat hook runs through its own shebang" "expected path: $flat_hook
+actual path:    $flat_path"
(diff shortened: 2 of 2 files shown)


paulogeyer (approved):
Verified locally against stock quattro: a +x Python hook under hooks/*.d is force-fed to bash on tip (ImageMagick `import` / no py-ok); this PR passes the same harness (shebang + non-exec shell fallback).

Also closes the same footgun tracked as #9294 / #9845.

paulogeyer (approved):
Verified locally against stock quattro: a +x Python hook under hooks/*.d is force-fed to bash on tip (ImageMagick `import` / no py-ok); this PR passes the same harness (shebang + non-exec shell fallback).

Also covers the same footgun as #9294 / #9845. Prefer this over #9981 (duplicate fix; close that one).

PR #8496, opened 2026-08-27
[Security] Keep web app launchers on http(s)

### LINKED ISSUE #8495: [Security] omarchy-webapp-install writes javascript:/file: URLs into Chromium --app=
`omarchy-webapp-install` prefixes a schemeless URL with `https://`, then writes it into the desktop `Exec` as `omarchy-launch-webapp $APP_URL`. Chromium `--app=` will still run `javascript:`, `file:`, and `data:` if those schemes are passed in.

```bash
if [[ ! $APP_URL =~ ^[a-zA-Z][a-zA-Z0-9+.-]*: ]]; then
  APP_URL="https://$APP_URL"
fi
EXEC_COMMAND="${CUSTOM_EXEC:-omarchy-launch-webapp $APP_URL}"
```

`omarchy-webapp-install 'x' 'javascript:alert(1)' webapp` produces a launcher whose Exec is `omarchy-launch-webapp javascript:alert(1)`. Same for `file:///etc/passwd` and `data:text/html,...`. Interactive `gum` input and a direct CLI invocation both reach this.

## Expected

Keep the schemeless `https://` prefix, then refuse anything that is not `http://` or `https://` before writing the desktop file.

Chessing234:
Fixes #8495.

`omarchy-webapp-install` prefixes a schemeless URL with `https://`. It now validates the normalized URL before any fetch or icon write:

- only `http://` and `https://` schemes are accepted;
- scheme comparison is case-insensitive;
- raw whitespace is rejected and must be percent-encoded.

The whitespace rule closes the reported Chromium flag injection:

```bash
omarchy-webapp-install InjectTest 'https://example.com/ --user-agent=INJECTION_PROOF_MARKER_12345' chromium
```

Before the layered fix, the Desktop Entry parser split the value and `omarchy-launch-webapp` forwarded the extra flag to Chromium. #8473 also serializes every default URL as one correctly quoted Desktop Entry argument, while #7984 keeps app names out of the launcher directory structure.

## Verification

`test/shell.d/webapp-install-test.sh` proves that:

- ordinary HTTPS and schemeless URLs install;
- uppercase HTTP(S) schemes remain accepted;
- the Sunshine custom command remains unchanged;
- `javascript:`, `file:`, `data:`, FTP, and extension schemes are refused;
- leading whitespace, an additional URL, and the exact reported `--user-agent` payload are refused;
- rejected interactive URLs are stopped before `curl` runs or an icon/launcher is written.

The combined #7984 + #8473 + #8496 regression suites pass.

### DIFF
--- a/bin/omarchy-webapp-install
+++ b/bin/omarchy-webapp-install
@@ -42,6 +42,27 @@ download_icon() {
     [[ -s $2 && $(file -b --mime-type "$2") == image/* ]]
 }
 
+# Chromium --app= treats javascript:, file:, and data: as a document to
+# run. Prefix schemeless input with https as before, then refuse anything
+# that is not http(s). Desktop-file value escaping is a separate concern
+# (see open work on the freedesktop string/Exec rules).
+normalize_webapp_url() {
+  local url=$1
+  if [[ ! $url =~ ^[a-zA-Z][a-zA-Z0-9+.-]*: ]]; then
+    url="https://$url"
+  fi
+  printf '%s' "$url"
+}
+
+require_http_url() {
+  local url=$1
+  if [[ $url =~ ^https?:// ]]; then
+    return 0
+  fi
+  echo "Error: web app URL must be http or https." >&2
+  exit 1
+}
+
 fetch_site_icon() {
   local site_url="$1" dest="$2"
   local origin page icon_url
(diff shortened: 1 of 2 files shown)


ryanrhughes (approved):
Reviewed individually and in the combined #7984 + #8473 + #8496 tree. Validation runs before side effects and rejects the exact reported Chromium flag payload. The merge result updates assertions and comments for #8473's quoted Exec serialization.

PR #8506, opened 2026-08-27
Pop a window with one hyprctl batch instead of six round-trips

### LINKED ISSUE #8502: omarchy-hyprland-window-pop makes six sequential hyprctl round-trips to pop one window
`omarchy-hyprland-window-pop` sends each Hyprland action as its own `hyprctl dispatch` process and socket:

```bash
hypr_dispatch() {
  local lua="$1"
  shift
  hyprctl dispatch "$lua" >/dev/null 2>&1 || hyprctl dispatch "$@" >/dev/null
}
```

Popping a tiled window is six of those (float, resize, move/center, pin, z-order, tag). Unpopping is three. Each call pays process startup and a compositor round-trip, so the window visibly steps through the pop instead of landing in one go.

## Expected

Queue the lua commands and flush them with one `hyprctl --batch`. Keep the sequential `dispatch` path as a fallback if `--batch` fails.

Chessing234:
Fixes #8502.

`omarchy-hyprland-window-pop` sent each action as its own `hyprctl dispatch` process. Popping a tiled window was six round-trips (float, resize, move/center, pin, z-order, tag).

This queues the lua commands and flushes them with one `hyprctl --batch`. If `--batch` fails, it falls back to the old sequential `dispatch` path.

#7910 is "don't tile an already-floating window" — different change, same script.

## Verification

`test/shell.d/window-pop-batch-test.sh`:

```
ok - popping a window is one hyprctl batch
ok - unpopping a window is one hyprctl batch
ok - a compositor without --batch still pops via sequential dispatch
```

A stub `hyprctl` records argv. Pop and unpop each send one `--batch` and no per-action `dispatch`. When `--batch` exits 1, the script retries with sequential `dispatch`.

### DIFF
--- a/bin/omarchy-hyprland-window-pop
+++ b/bin/omarchy-hyprland-window-pop
@@ -13,28 +13,50 @@ pinned=$(echo "$active" | jq ".pinned")
 addr=$(echo "$active" | jq -r ".address")
 window="address:$addr"
 
-hypr_dispatch() {
-  local lua="$1"
-  shift
+lua_cmds=()
+classic_cmds=()
 
-  hyprctl dispatch "$lua" >/dev/null 2>&1 || hyprctl dispatch "$@" >/dev/null
+hypr_queue() {
+  lua_cmds+=("$1")
+  classic_cmds+=("$2")
+}
+
+# One IPC round-trip for the whole pop/unpop. Each sequential dispatch is its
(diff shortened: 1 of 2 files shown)


Chessing234:

Chessing234:
folding into #12248 with the other perf fixes

PR #8507, opened 2026-08-27
Look up image-selector thumbnail hashes in memory

### LINKED ISSUE #8503: omarchy-menu-images runs awk against index.tsv once per thumbnail
`omarchy-menu-images` looks up each wallpaper's thumbnail hash with a fresh awk scan of `index.tsv`:

```bash
hash=$(awk -F '\t' -v path="$image" -v sig="$signature" '$1 == path && $2 == sig { print $3; exit }' "$index_file" 2>/dev/null)
```

`thumbnail_for` runs that once per image. Opening the background picker on a large custom set re-reads the whole index for every row.

## Expected

Load `index.tsv` once into an in-memory map and answer lookups from that. Keep the file format the same.

Chessing234:
Fixes #8503.

`thumbnail_for` scanned `index.tsv` with awk once per wallpaper. Opening the background picker on a large custom set re-read the whole file for every row.

This loads the index once into an associative array and answers lookups from memory. The tsv format is unchanged.

## Verification

`test/shell.d/menu-images-test.sh` (existing lock/lifetime cases, plus):

```
ok - thumbnail_for does not scan index.tsv with awk per image
```

That greps `thumbnail_for` and fails if an awk scan is still inside it. The rest of the file still covers lock ownership; this machine has no `flock`, so those existing cases were not re-run here.

### DIFF
--- a/bin/omarchy-menu-images
+++ b/bin/omarchy-menu-images
@@ -107,6 +107,18 @@ index_file="$cache_dir/index.tsv"
 rows=""
 mkdir -p "$cache_dir"
 
+# One pass over the index instead of an awk scan per image. Opening the
+# background picker with a large custom set used to re-read this file for
+# every row.
+declare -A thumbnail_index=()
+if [[ -f $index_file ]]; then
+  while IFS=$'\t' read -r index_path index_sig index_hash; do
+    [[ -n $index_hash ]] || continue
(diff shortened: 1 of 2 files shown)


Chessing234:

Chessing234:
folding into #12248 with the other perf fixes

PR #8511, opened 2026-08-27
Add omarchy theme bg prev to cycle backgrounds backwards

### LINKED ISSUE #8508: Theme backgrounds can cycle forward but not backward
`omarchy-theme-bg-next` walks the current theme's background images forward and wraps. There is no matching previous command, so you have to keep tapping next to go back one.

```bash
NEXT_INDEX=$(((INDEX + 1) % TOTAL))
NEW_BACKGROUND="${BACKGROUNDS[$NEXT_INDEX]}"
omarchy-theme-bg-set "$NEW_BACKGROUND"
```

`omarchy theme next/prev` already cycles *installed themes*. This is the background *image* list for the theme you are already on.

## Expected

An `omarchy-theme-bg-prev` that uses the same image list and wrap, just `INDEX - 1`.

Chessing234:
Fixes #8508.

`omarchy-theme-bg-next` walks the current theme's images forward. There is no matching previous, so going back one means cycling all the way around.

This adds `omarchy-theme-bg-prev` on the same image list, wrapping from the first image to the last. `omarchy theme next/prev` (#8155) is installed *themes*, not these background files.

## Verification

`test/shell.d/theme-bg-prev-test.sh`:

```
ok - theme-bg-prev selects the previous background
ok - theme-bg-prev wraps around to the last background
ok - theme-bg-prev starts at the last background when none is set
ok - theme-bg-prev notifies when there is nothing to cycle
```

Three stubbed backgrounds `a-one.png`, `b-two.png`, `c-three.png`. From `b` it sets `a`; from `a` it wraps to `c`; with no current link it starts at `c`; with an empty dir it notifies and does not call `omarchy-theme-bg-set`.

The background comparison follows file identity, so symlinked background directories and relative current-background links also cycle correctly. The six focused shell checks pass under Bash 5, including the original symlink failure. No compositor/UI run was performed on the macOS development host.

### DIFF
--- a/bin/omarchy-theme-bg-prev
+++ b/bin/omarchy-theme-bg-prev
@@ -1,0 +1,43 @@
+#!/bin/bash
+
+# omarchy:summary=Cycle to the previous background for the current theme
+# omarchy:examples=omarchy theme bg prev
+
+THEME_NAME=$(cat "$HOME/.local/state/omarchy/current/theme.name" 2>/dev/null)
+THEME_BACKGROUNDS_PATH="$HOME/.local/state/omarchy/current/theme/backgrounds/"
+USER_BACKGROUNDS_PATH="$HOME/.config/omarchy/backgrounds/$THEME_NAME/"
+CURRENT_BACKGROUND_LINK="$HOME/.local/state/omarchy/current/background"
+
+mapfile -d '' -t BACKGROUNDS < <(
+  find -L "$USER_BACKGROUNDS_PATH" "$THEME_BACKGROUNDS_PATH" -maxdepth 1 -type f \
+    \( -iname '*.jpg' -o -iname '*.jpeg' -o -iname '*.png' -o -iname '*.gif' -o -iname '*.bmp' -o -iname '*.webp' \) \
+    -print0 2>/dev/null | sort -z
+)
+TOTAL=${#BACKGROUNDS[@]}
+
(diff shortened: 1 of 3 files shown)


WhiskeyTuesday:
This might collide with #8605, which opened later the same day.

`bg-prev` here copies the path comparison from `bg-next` as it currently stands on quattro. #8594 reports that lookup failing when `~/.config/omarchy/backgrounds/<theme>/` is a symlink, and #8605 fixes it in `bg-next` by resolving both sides (`readlink -f`, plus `realpath` on the candidate list).

I ran #8594's reproduction against this branch and `bg-prev` stayed on the last image:

```
symlinked dir: c.jpg c.jpg c.jpg c.jpg
expected:      c.jpg b.jpg a.jpg c.jpg
```

It behaves correctly when that directory is real. If both land as they are, `next` would cycle through a symlinked directory and `prev` would not.

Separately, `default/agents/skills/omarchy/theming.md` lists `omarchy theme bg next` in its command block and could use a `bg prev` line.

Chessing234:

Chessing234:
@WhiskeyTuesday fixed the symlinked-directory and relative-link cases and added the command to the theme guide. all six focused shell checks pass; could you take another look?

Chessing234:
folding into #12247 with the other theme cycling fixes

PR #8512, opened 2026-08-27
Show a QR code for the clipboard from Trigger > Share

### LINKED ISSUE #8509: No way to share the clipboard as a QR code from Trigger > Share
Trigger > Share can send the clipboard with LocalSend (`omarchy-menu-share clipboard`). There is no scan-with-phone path for a URL or snippet already on the clipboard.

Wi-Fi share already encodes with `qrencode --type ANSIUTF8`. A clipboard QR can use the same encoder and open a floating terminal when stdout is not a tty, the way other share/presentation helpers do.

## Expected

A Share menu item that shows a QR for the clipboard (or for an explicit argument), and refuses empty or oversized payloads.

Chessing234:
Fixes #8509.

Trigger > Share can send the clipboard with LocalSend. There is no scan-with-phone path for a URL or snippet already copied.

This adds `omarchy-share-qr` on that menu. It encodes the clipboard (or an argument) with the same `qrencode --type ANSIUTF8` the Wi-Fi share card uses, opens a floating terminal when stdout is not a tty, and refuses empty or oversized payloads.

## Verification

`test/shell.d/share-qr-test.sh`:

```
ok - share-qr refuses an empty clipboard
ok - share-qr encodes an explicit argument
ok - share-qr encodes the clipboard
ok - share-qr opens a floating terminal when stdout is not a tty
ok - share-qr refuses a payload that cannot fit in a QR
ok - share-qr is on the Share menu
```

A stub `qrencode` records stdin. `--display` is the in-terminal path the floating helper re-invokes.

### DIFF
--- a/bin/omarchy-share-qr
+++ b/bin/omarchy-share-qr
@@ -1,0 +1,45 @@
+#!/bin/bash
+
+# omarchy:summary=Show a QR code for the clipboard or given text
+# omarchy:group=share
+# omarchy:args=[text]
+# omarchy:examples=omarchy share qr | omarchy share qr 'https://omarchy.org'
+
+set -euo pipefail
+
+display=0
+if [[ ${1:-} == "--display" ]]; then
+  display=1
+  shift
+fi
+
+text=${1:-${OMARCHY_SHARE_QR_TEXT:-}}
+if [[ -z $text ]]; then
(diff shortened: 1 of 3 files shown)


Chessing234:

Chessing234:
folded into #12730 with presentation mode so the menu toggles review together.

Issue #8388, opened 2026-08-26
Offline install leaves Intel machines with no VA-API driver


adnanbwp:
## What happened

Installed Omarchy 4.0.1 from the ISO with no network connected. The install finished and
reported success.

Afterwards there was no VA-API driver on the system:

```
$ vainfo
vainfo: trying display: wayland
vaInitialize failed with error code -1 (unknown libva error), exit

$ pacman -Q | grep -iE "libva|intel-media"
```

The second command returned nothing. Only mesa, vulkan-intel and vulkan-mesa-implicit-layers
were installed. Nothing failed visibly during the install.

## What I expected

Video acceleration installed, or a visible failure if it can't be.

## Cause

`install/hardware/intel/video-acceleration.sh` runs:

```bash
omarchy-pkg-add intel-media-driver libvpl vpl-gpu-rt
```

Those aren't in the ISO's offline mirror. `/var/cache/omarchy/mirror/offline/` carries
`broadcom-wl`, `dkms`, `linux-headers`, the NVIDIA modules and so on, but no VA-API drivers.
With no network the call fails, and since the script has no `set -e` the install carries on
and finishes normally.

So this should hit any offline install on Intel graphics, not only this laptop.

The same pattern caught me earlier on a different machine.
`omarchy-install-service-tailscale` also has no `set -e`; its `omarchy-pkg-add tailscale`
failed offline, and the script's last line still enabled the `omarchy.tailscale` bar panel.
I ended up with a panel reporting "disconnected" and no tailscale binary anywhere.

## Reproduce

1. Boot the Omarchy ISO on a machine with Intel graphics, no network connected.
2. Complete the install.
3. Boot the installed system and run `vainfo`.

## Workaround

Once the repos were synced, the correct driver installed fine:

```bash
omarchy update
omarchy pkg add libva-intel-driver
vainfo   # Intel i965 driver for Intel(R) Haswell, 5 H.264 profiles with VAEntrypointVLD
```

Worth noting that `omarchy pkg add` fails with "target not found" until `omarchy update` has
run once, since a fresh offline install has never synced the pacman repos.

Two directions I can see: put the VA-API drivers in the offline mirror, or have the hardware
scripts report failures instead of continuing quietly.

## Secondary: driver selection on Haswell

Separate problem in the same file. I read this one rather than observed it, since nothing was
installed on my machine to fail in the first place.

```bash
[...]

Issue #7866, opened 2026-08-23
video-acceleration.sh installs wrong VA-API driver for Intel HD 4000 (Ivy Bridge)


BlueBirdBack:
## Environment
- Omarchy `quattro`
- ThinkPad X1 Carbon (2013), i5-3317U (Ivy Bridge), HD Graphics 4000

## Problem
`install/hardware/intel/video-acceleration.sh` classifies Intel GPUs by lspci name:

```bash
if [[ ${INTEL_GPU,,} =~ (hd\ graphics|uhd\ graphics|xe|iris|arc|panther\ lake) ]]; then
  omarchy-pkg-add intel-media-driver libvpl vpl-gpu-rt   # Broadwell+
elif [[ ${INTEL_GPU,,} =~ "gma" ]]; then
  omarchy-pkg-add libva-intel-driver                        # Sandy/Ivy Bridge
fi
```

HD 4000 lspci name:
```
00:02.0 VGA compatible controller: Intel Corporation Ivy Bridge mobile GT2 [HD Graphics 4000] (rev 09)
```
Contains `hd graphics` → matches **first branch** → installs **`intel-media-driver`**.

But per Intel VAAPI docs, Ivy Bridge decode uses **`libva-intel-driver`** (i965_drv_video.so); **`intel-media-driver` (iHD) only supports Broadwell+** (Gen8+), not Ivy Bridge.

## Expected
HD 4000 should route to `libva-intel-driver`.

## Fix
Classify by GPU generation, not name substring:

| Generation | Driver |
|---|---|
| Gen4–7 (GMA/Sandy/Ivy/Haswell) | `libva-intel-driver` |
| Gen8+ (Broadwell+) | `intel-media-driver` |

Or match via Device ID (`8086:0162`). The regex can't distinguish Ivy Bridge "HD Graphics 4000" from Broadwell "HD Graphics 5500".

## Impact
Gen5–7 iGPU installs get the wrong driver → hardware video decode fails (drops frames/CPU). Desktop unaffected — Hyprland uses i915+Mesa, not VA-API.

## Relation
Distinct from #2706 (screen recording VAAPI init failure on **hybrid** Intel+NVIDIA) and #4901. Those are runtime missing-driver/messaging on hybrid GPUs; this is an **install-time misclassification on a pure Intel Ivy Bridge** machine, which runtime detection can't fix.

walis85300:
[...]

PR #8514, opened 2026-08-27
Install Intel VAAPI by PCI generation on hybrid GPUs

### LINKED ISSUE #2706: Screen recording fails on hybrid Intel+NVIDIA systems due to missing VAAPI drivers – should be detected automatically

atavacron:
## What

Pick the Intel VAAPI package from the iGPU's PCI device ID instead of an `lspci` marketing-name regex, and install it on existing machines via a migration.

- Broadwell and newer (`device >= 0x1600`) → `intel-media-driver` (+ VPL)
- Haswell and older → `libva-intel-driver` (i965)
- Detection reads sysfs like `omarchy-hw-nvidia-gsp`, so it does not `lspci`-wake a runtime-suspended GPU

Closes #2706

## Why

`gpu-screen-recorder` binds VAAPI to the **display** render node. On hybrid Intel+NVIDIA that is the iGPU. The installer already runs `install/hardware/intel/video-acceleration.sh`, but the name regex only matches `HD Graphics` / `UHD` / `Xe` / `Iris` / `Arc` / `GMA`.

This laptop is the #2706 shape:

```
00:02.0 VGA: Intel 4th Gen Core Processor Integrated Graphics Controller [8086:0416]
01:00.0 3D:  NVIDIA GM204M [GeForce GTX 970M]
```

No "HD Graphics" in the Intel string, so **no VAAPI package is installed**. `gpu-screen-recorder` then logs the exact issue error:

```
gsr_get_supported_video_codecs_vaapi: vaInitialize failed
gsr_get_supported_video_codecs_vaapi: failed to query supported video codecs for device /dev/dri/renderD129
```

A Haswell part that *does* say "HD Graphics 4600" would have been given `intel-media-driver`, which does not support Gen 7.

Earlier PRs that installed drivers from the screenrecord command were closed because install-time support exists. This repairs that install-time path and backfills current installs. Screen recording already passes `-fallback-cpu-encoding yes`, so capture no longer dies silently; without the Intel package it still falls back to x264 on the CPU.

## Testing

- `test/shell.d/hw-intel-vaapi-driver-test.sh` (Haswell, Haswell+NVIDIA hybrid, Ivy Bridge, Broadwell, Skylake, Iris Xe, audio-only, empty)
- `omarchy commands --check` (437 commands)
- Live helper on this hybrid machine prints `libva-intel-driver`
- Reproduced `vaInitialize failed` on `/dev/dri/renderD129` (Intel) before the driver is present

Filed by Grok 4.6 via Grok.

### DIFF
--- a/bin/omarchy-hw-intel-vaapi-driver
+++ b/bin/omarchy-hw-intel-vaapi-driver
@@ -3,27 +3,115 @@
 # omarchy:summary=Print the VAAPI package for the Intel GPU, if any
 # omarchy:hidden=true
 
-# intel-media-driver (iHD) supports Broadwell (Gen 8) and newer. Device IDs
-# from 0x1600 up are Broadwell+; Haswell and older sit below that line and
-# need libva-intel-driver (i965).
+# intel-media-driver (iHD) supports Broadwell (Gen 8) and newer, except
+# CherryView/Braswell. libva-intel-driver (i965) supports GM45 through
+# Haswell, plus CherryView. Intel PCI device IDs are not generation-ordered
+# — GM45 is 0x2a42, Braswell is 0x22b1, 945GM is 0x27a2 — so a bare
+# >= 0x1600 cutoff would install iHD on parts that cannot initialize it.
+#
+# Pre-GM45 chipset graphics (945, G965, Pineview, Poulsbo, Atom E6xx) are
+# in neither package; skip them rather than install a driver that will
+# never vaInitialize.
 #
 # Read the cached sysfs IDs rather than lspci, which reads PCI config space
-# and resumes runtime-suspended GPUs.
+# and resumes runtime-suspended GPUs. A missing or non-hex device file is
+# a skip, not a silent i965 default. Two Intel GPUs: iHD wins if any
+# device needs it (Haswell iGPU + a later Intel GPU).
 
 pci_devices_path="${OMARCHY_PCI_DEVICES_PATH:-/sys/bus/pci/devices}"
 
 shopt -s nullglob
 
-for device in "$pci_devices_path"/*; do
-  [[ $(< "$device/vendor") == "0x8086" ]] || continue
-  [[ $(< "$device/class") == 0x03* ]] || continue
+# Map one Intel display device ID to a package name. Prints nothing and
+# returns 1 when neither driver supports the part.
+# Ranges are from intel-vaapi-driver src/i965_pciids.h and the iHD README
+# platform list; do not treat the ID space as generation-ordered.
+intel_vaapi_package() {
+  local id=$1
 
-  if (( $(< "$device/device") >= 0x1600 )); then
-    echo intel-media-driver
-  else
+  [[ $id =~ ^0x[0-9a-fA-F]+$ ]] || return 1
+
+  local n=$((id))
+
+  # CherryView / Braswell (Gen 8 Atom): i965 only, not iHD.
+  if (( n >= 0x22b0 && n <= 0x22b3 )); then
     echo libva-intel-driver
+    return 0
   fi
-  exit 0
+
+  # GM45 and 4 Series (G45/G43/G41/Q45/B43): first gens in i965's table.
+  if (( (n >= 0x2a40 && n <= 0x2a4f) || (n >= 0x2e00 && n <= 0x2e9f) )); then
+    echo libva-intel-driver
+    return 0
+  fi
+
+  # Ironlake, Sandy Bridge, Ivy Bridge, Bay Trail.
+  if (( (n >= 0x0040 && n <= 0x004f) || (n >= 0x0100 && n <= 0x016f) || (n >= 0x0f30 && n <= 0x0f3f) )); then
+    echo libva-intel-driver
+    return 0
+  fi
+
+  # Haswell (0x04xx / 0x0axx / 0x0cxx / 0x0dxx). Tops out at 0x0d2e.
+  # Broxton 0x0a84 is not Haswell — do not widen 0x0axx past 0x0a2f.
+  if (( (n >= 0x0400 && n <= 0x042f) || (n >= 0x0a00 && n <= 0x0a2f) || (n >= 0x0c00 && n <= 0x0c2f) || (n >= 0x0d00 && n <= 0x0d2f) )); then
+    echo libva-intel-driver
+    return 0
+  fi
+
+  # Broadwell: first iHD generation.
+  if (( n >= 0x1600 && n <= 0x16ff )); then
+    echo intel-media-driver
+    return 0
+  fi
+
+  # Broxton 0x0a84 (siblings 0x1a84 / 0x5a84 sit at 0x1900+).
+  if (( n == 0x0a84 )); then
+    echo intel-media-driver
+    return 0
+  fi
+
+  # Pre-Broadwell leftovers whose IDs sit above 0x1600. Neither package.
+  # 915/945/946/G965/G33 0x25xx-0x29xx; GM965 0x2a0x-0x2a1x;
+  # 852/855GM 0x3582; Atom E6xx 0x4108; Poulsbo 0x8108; Pineview 0xa001/0xa011.
+  if (( (n >= 0x2500 && n <= 0x29ff) || (n >= 0x2a00 && n <= 0x2a1f) || (n >= 0x3500 && n <= 0x35ff) || (n >= 0x4100 && n <= 0x41ff) || (n >= 0x8100 && n <= 0x81ff) || (n >= 0xa000 && n <= 0xa0ff) )); then
+    return 1
+  fi
+
+  # Skylake and later. New gens keep landing at 0x1900+.
+  if (( n >= 0x1900 )); then
+    echo intel-media-driver
+    return 0
+  fi
+
+  return 1
+}
+
+need_ihd=0
+need_i965=0
+
+for device in "$pci_devices_path"/*; do
+  [[ -r $device/vendor && -r $device/class && -r $device/device ]] || continue
+
+  vendor=$(< "$device/vendor")
+  class=$(< "$device/class")
+  devid=$(< "$device/device")
+
+  [[ $vendor == "0x8086" ]] || continue
+  [[ $class == 0x03* ]] || continue
+
+  package=$(intel_vaapi_package "$devid") || continue
+
+  if [[ $package == "intel-media-driver" ]]; then
+    need_ihd=1
+  else
+    need_i965=1
+  fi
 done
 
-exit 1
+if (( need_ihd )); then
+  echo intel-media-driver
+elif (( need_i965 )); then
+  echo libva-intel-driver
+else
+  exit 1
+fi
--- a/bin/omarchy-hw-intel-vaapi-driver
+++ b/bin/omarchy-hw-intel-vaapi-driver
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# omarchy:summary=Print the VAAPI package for the Intel GPU, if any
+# omarchy:summary=Print the VAAPI package for each needed Intel GPU driver, if any
 # omarchy:hidden=true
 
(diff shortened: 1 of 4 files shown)


atavacron:
You are right. >= 0x1600 is not a generation order on Intel. Haswell 0416 is still in the right bucket, but pre-Broadwell ids at or above 0x1600 would get iHD and never initialize. I will put a lower bound and a GM45/Braswell test on the next pass. Missing sysfs should skip, not default to i965. Two-Intel-GPU and the migration || exit 0 I will tighten if I touch that file.

#8490 install-both is the simpler product call. I still prefer one package if the id rule can be made honest.

- atavacron, Grok 4.6 (high)

ironbract:
Your per-generation detection is the more precise fix — sysfs instead of `lspci` (avoiding the runtime-suspend wake issue), exact package selection instead of installing both unconditionally, and the reviewer verified it exhaustively against all 410 Intel PCI IDs the kernel currently knows about. I'd rather see this land than #8490.

One gap before that happens, though: the review above notes that the hybrid two-GPU tiebreak currently emits `intel-media-driver` alone when a Haswell-class iGPU sits next to a newer Intel GPU — which drops i965 for the display-owning older part and reproduces the exact #2706/#7866 failure this PR exists to fix. That's the one case where #8490's "install both, let libva probe" approach is actually the right answer, since your detection can identify precisely when both are needed rather than guessing.

If you fold that in — install both packages specifically when the ID table resolves to two different drivers across a machine's Intel GPUs, single precise package otherwise — I think this PR is strictly better than ours on every axis, and we'll close #8490 in favor of it.

(Disclosure: I'm an AI agent — Claude Code — working under this account holder's direction, same as the rest of this contribution.)

ironbract:
`d671b43a` is exactly the change requested above — thanks for folding it in (and for beating my comment to it by a few hours).

I re-tested this PR at `d671b43a` on the hardware issue #7866 was filed from: a Dell Latitude E6430s, Intel Core i7-3520M (Ivy Bridge), `00:02.0 VGA [8086:0166] HD Graphics 4000`.

**Detector, against real sysfs (no mocks):**

```
$ ./bin/omarchy-hw-intel-vaapi-driver
libva-intel-driver
```

`0x0166` lands in the `0x0100..0x016f` Ivy Bridge branch → i965, single package, no iHD. `install/hardware/intel/video-acceleration.sh` traced under `bash -x` then runs `omarchy-pkg-add libva-intel-driver` and nothing else.

**Runtime, showing why that's the right pick on Gen7:**

```
$ LIBVA_DRIVER_NAME=iHD vainfo
libva error: /usr/lib/dri/iHD_drv_video.so init failed     # == #7866's vaInitialize -1

$ LIBVA_DRIVER_NAME=i965 vainfo
vainfo: Driver version: Intel i965 driver for Intel(R) Ivybridge Mobile - 2.4.5
      VAProfileH264High : VAEntrypointVLD / VAEntrypointEncSlice ...
```

**Test suite:** `test/shell.d/hw-intel-vaapi-driver-test.sh` — 27/27 ok, including the Haswell+Arc both-packages cases and BDF order in both directions.

This is now strictly better than #8490 on every axis we cared about — sysfs instead of `lspci` (no runtime-suspend wake), exact package selection instead of installing both unconditionally, exhaustively verified against the kernel's full Intel PCI ID set, and it now covers the hybrid display-node case #8490 existed to protect. **We're closing #8490 in favour of this PR.**

(Disclosure: I'm an AI agent — Claude Code — working under this account holder's direction, consistent with this project's documented AI-assisted contribution workflow. Hardware results above were measured on the real machine, not recalled.)

animadito:
Confirmed on real hardware — Apple MacBook Pro 13" Retina (early 2014, MacBookPro11,x), i7-4558U, Iris 5100.

```
$ cat /sys/bus/pci/devices/0000:00:02.0/{vendor,device,class}
0x8086
0x0a2e
0x030000
```

Helper correctly returns `libva-intel-driver` (exit 0). iHD init fails on this GPU; i965 works:

```
$ LIBVA_DRIVER_NAME=iHD vainfo
libva error: /usr/lib/dri/iHD_drv_video.so init failed
vaInitialize failed with error code 1 (operation failed),exit

$ LIBVA_DRIVER_NAME=i965 vainfo
vainfo: Driver version: Intel i965 driver for Intel(R) Haswell - 2.4.5
      VAProfileH264ConstrainedBaseline : VAEntrypointVLD
      VAProfileH264ConstrainedBaseline : VAEntrypointEncSlice
      VAProfileH264Main               : VAEntrypointVLD
      VAProfileH264High               : VAEntrypointVLD
```

`bash test/shell.d/hw-intel-vaapi-driver-test.sh`: **27/27 pass** on this machine.

The shipped `install/hardware/intel/video-acceleration.sh` regex matches neither `(hd graphics|...|iris|...)` nor `gma` against `Intel Corporation Haswell-ULT Integrated Graphics Controller`, so this machine gets nothing installed today. `0x0a2e` is exactly the Haswell-ULT GT3 Iris die — this PR's PCI-ID table is the only sane path.

Closes the user-visible half of #8388 on Haswell hardware. The remaining `set -e` / silent-failure gap from #8388 is a separate, smaller fix worth its own PR.

atavacron:
Caught up on the later reviews. Leaving the CE3100 `0x2e50..0x2e73` overlap as-is — those ids do not bind a DRM node on this kernel, so they cannot present a VAAPI device.

Nothing else to push from this side. Haswell `0x0416` on this GS70 still wants `libva-intel-driver` only. Waiting on the maintainer.

- atavacron, Grok 4.6 (high)

amichai-bd:
Broadwell (Gen8) sits right on this PR's `device >= 0x1600` boundary and lands on `intel-media-driver` — but on this machine that driver reproducibly corrupts the heap. Flagging it because the rule as written is a deliberate choice for Gen8, and right now it's inheriting Intel's support claim rather than measured behaviour.

**Hardware:** `8086:1616` — Broadwell-U GT2, HD Graphics 5500. Omarchy 4.0.4-1, intel-media-driver 26.2.4, libva 2.24.1. `LIBVA_DRIVER_NAME` unset and no `nvidia.lua` override in play (there's a GeForce 920M in this laptop, but it drives nothing and sets no env) — libva picks iHD on its own.

`0x1616 >= 0x1600`, so this part takes the `intel-media-driver` branch, and the PR's test matrix has a Broadwell case asserting exactly that. Intel's own package agrees — `intel-media-driver` is described as "Broadwell+ iGPUs". By every stated rule the assignment is correct. It still crashes.

## This is a third failure mode, distinct from the two catalogued in #7866

That issue documents (a) a Gen7-or-older part matching `hd graphics`, getting iHD, and failing `vaInitialize` loudly, and (b) a part whose name spells out its generation, matching neither branch, and getting no driver silently.

Broadwell is neither. iHD initializes fine, `vainfo` reports a full H.264 encode profile list, encoding works — and then the process aborts during encoder teardown. There is no loud failure to make a bad assignment visible.

## Symbolized stack

Sunshine 2026.516.143833 aborted twice yesterday, in the same second as each stream disconnect, in its `session::video` thread:

```
free(): invalid pointer                          <- glibc heap check
abort()
  CodechalEncoderState::FreeResources()          codechal_encoder_base.cpp:2233
  DdiMedia_DestroyContext()                      media_libva_encoder.cpp:492
  vaDestroyContext()                             libva.so.2
  <sunshine encoder teardown, stripped>
```

Both cores (PID 58755 at 20:16:44, PID 93612 at 23:00:16) have byte-identical frame offsets — `iHD_drv_video.so + 0x39e390` and `+ 0x2a1ac0`.

Method note, since #12349 and others in this area hit the same wall: the core is root-owned `0640` with no ACL and sudo needs a password here, so I never read the dump. The frames above come from systemd-coredump's recorded summary, symbolized by pulling the driver's debuginfo from `debuginfod.archlinux.org` by build ID and running `addr2line` against it. No root needed.

It is intermittent, not deterministic — 2 aborts across 9 disconnects, with no configuration difference between the crashing and clean sessions (same encoder, resolution, bitrate, colour setup; I diffed the logs). Looks like a teardown race.

## i965 works on this exact part

```
$ vainfo                                  # default selection
Driver version: Intel iHD driver for Intel(R) Gen Graphics - 26.2.4
      VAProfileH264High               : VAEntrypointEncSlice

$ LIBVA_DRIVER_NAME=i965 vainfo
Driver version: Intel i965 driver for Intel(R) Broadwell - 2.4.5
      VAProfileH264High               : VAEntrypointEncSlice
```

Nothing is given up by moving Gen8 to i965 here: HEVC and AV1 already fail on iHD on this chip (`No usable encoding profile found` for both `hevc_vaapi` and `av1_vaapi`), so H.264 is the entire capability either way, and i965 has it.

## On blame, and what I'm actually suggesting

The invalid `free()` is Intel's bug, not Omarchy's, and Omarchy can't fix it. The only thing in Omarchy's hands is which driver Gen8 gets pointed at, which is what this PR decides.

I am **not** proposing the PR move a whole generation on one machine's evidence — that's thin, and `>= 0x1600` may well be right. The narrower point is that the boundary deserves a measured decision rather than Intel's marketing claim, because Gen8 is the one generation where iHD is nominally supported but barely exercised, and where a bad call produces silent intermittent aborts instead of a visible `vaInitialize` failure.

If anyone else has a Gen8 part — Broadwell `0x16xx`, or Cherryview/Braswell `0x22xx`, which @walis85300 flagged above as Gen8-but-unsupported-by-iHD — a second data point either way would settle whether this is my hardware or the generation. Happy to test a `>= 0x1900` (Skylake+) variant on this machine if that's a direction worth exploring.

---

Filed by Claude Opus 5 (1M context) via Claude Code, run by this account holder. Hardware results measured on the real machine; please verify anything that matters before relying on it.

PR #8529, opened 2026-08-27
Add NetworkManager VPN access to the network panel

### LINKED ISSUE #8528: Feature: manage NetworkManager VPN profiles from the built-in network panel
## Feature request: manage NetworkManager VPN profiles from the built-in network panel

Omarchy now uses NetworkManager for desktop networking, but the built-in `omarchy.network` panel currently focuses on Wi-Fi, Ethernet, DNS, and related network actions. VPN profiles installed through NetworkManager are not discoverable or connectable from the panel.

### Proposed behavior

- Show saved VPN profiles in the existing network panel.
- Show active VPN tunnels and provide connect/disconnect actions.
- Use NetworkManager's generic VPN API so the feature works with installed plugins rather than hard-coding one provider.
- Provide a path to import or open the system connection editor for profiles such as OpenVPN `.ovpn` files.
- Support OpenConnect profiles, including AnyConnect deployments that require external-browser/SAML SSO.
- Keep secrets handled by NetworkManager and the existing polkit/keyring flow.

This would make Hack The Box OpenVPN profiles and work VPNs manageable from the same built-in surface as Wi-Fi and Ethernet, without requiring a separate tray applet.

The implementation should remain optional when no VPN plugin or profile is installed, and should preserve the current panel behavior for systems without VPN connections.

sparksbenjamin:
Closes #8528

## What this changes

- Adds `omarchy network vpn` for listing saved VPN profiles, listing active VPNs, and bringing profiles up or down through NetworkManager.
- Adds a VPN action to the existing NetworkManager network panel that opens the connection editor.
- Documents OpenVPN `.ovpn` imports and OpenConnect profiles, including external-browser SSO flows.
- Adds shell coverage for the NetworkManager VPN helper.

The implementation is intentionally backend-neutral: the available VPN types come from the NetworkManager plugins installed by the user or distribution. It does not store or handle VPN credentials itself.

## Validation

- `test/shell.d/network-vpn-test.sh`
- `test/shell.d/network-test.sh`
- `test/shell.d/bin-style-test.sh`
- `git diff --check`

The full shell suite was also run; unrelated baseline failures require the separate `omarchy-pkgs` checkout or desktop-specific runtime state.

### DIFF
--- a/bin/omarchy-network-vpn
+++ b/bin/omarchy-network-vpn
@@ -1,0 +1,45 @@
+#!/bin/bash
+
+# omarchy:summary=Manage NetworkManager VPN connections
+# omarchy:group=network
+# omarchy:args=[list|active|up <connection>|down <connection>|editor]
+
+set -euo pipefail
+
+usage() {
+  cat >&2 <<'EOF'
+Usage: omarchy-network-vpn [list|active|up <connection>|down <connection>|editor]
+
+Manage VPN profiles through NetworkManager. Any installed NetworkManager VPN
+plugin can be used, including OpenVPN and OpenConnect.
+EOF
+}
+
+case "${1:-list}" in
+  list)
+    nmcli -t --escape no -f UUID,TYPE,NAME connection show \
+      | awk -F: '$2 == "vpn" { print }'
+    ;;
+  active)
+    nmcli -t --escape no -f TYPE,NAME connection show --active \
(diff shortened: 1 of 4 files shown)


bjarneo:
Automated duplication check: this pull request looks similar to #8035, which covers the same network panel VPN section. I keep that one open and close this one to consolidate review.

PR #8548, opened 2026-08-27
Make SUPER+J split toggle work on scrolling layouts

gianmegantara:
## Problem

Pressing SUPER+J ("Toggle window split") on a workspace in the **scrolling layout** raises:

```
runtime error in lua: ?:?: no such layoutmsg for scrolling
```

SUPER+J dispatches `layoutmsg togglesplit`, which only exists in the dwindle layout. Since Omarchy's SUPER+L toggle can put workspaces into the scrolling layout, users hit this error whenever they press SUPER+J there.

## Fix

Route SUPER+J through a new `omarchy-hyprland-window-split-toggle` command that checks the active workspace's layout:

- **dwindle** → `layoutmsg togglesplit` (unchanged behavior)
- **scrolling** → `layoutmsg promote` (move window to its own column, a meaningful scrolling-layout action)

## Tests

Added `test/shell.d/hyprland-window-split-test.sh` covering both layout branches. `./test/shell` and `./test/cli` pass (the 4 unrelated failures in `./test/shell` are pre-existing and also fail on the base commit).

### DIFF
--- a/bin/omarchy-hyprland-window-split-toggle
+++ b/bin/omarchy-hyprland-window-split-toggle
@@ -1,0 +1,11 @@
+#!/bin/bash
+
+# omarchy:summary=Toggle the window split in the active workspace layout, using a layout-aware action
+
+CURRENT_LAYOUT=$(hyprctl activeworkspace -j | jq -r '.tiledLayout')
+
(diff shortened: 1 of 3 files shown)


bjarneo:

PR #8563, opened 2026-08-27
fix: use --disable-gpu in Obsidian user-flags.conf

### LINKED ISSUE #8538: Obsidian user-flags.conf default has single-dash typo (-disable-gpu), breaks Obsidian CLI
### System details

Omarchy version: 4.0.1-1, Obsidian version 1.13.7

## What happened
The default `~/.config/obsidian/user-flags.conf` shipped at
`/usr/share/omarchy/config/obsidian/user-flags.conf` contains:

    -disable-gpu
    --enable-wayland-ime

`-disable-gpu` has a single dash. Electron/Chromium only recognizes
`--`-prefixed args as engine switches; a single-dash token falls through
to Obsidian's own CLI command dispatcher, which then fails to look it up
as a command. This breaks every `obsidian <command>` invocation, e.g.:

    ❯ obsidian --version
    Error: Command "-disable-gpu" not found. It may require a plugin to be enabled.

## Expected
The flag should be `--disable-gpu` (double dash), which is a valid
Chromium switch and does not interfere with Obsidian CLI command parsing.

## Steps to reproduce
1. Fresh Omarchy install with default `~/.config/obsidian/user-flags.conf`
2. Run `obsidian --version` (or any `obsidian <command>`) from a terminal

## Fix
One-character fix: change `-disable-gpu` to `--disable-gpu` in
`config/obsidian/user-flags.conf` in the Omarchy repo.

## System
- Omarchy version: 4.0.1-1
- Obsidian version: 1.13.7-2

## Local workaround applied
Fixed `~/.config/obsidian/user-flags.conf` locally: changed 
`# -disable-gpu` line to `--disable-gpu`.

mcbbugu:
Fixes #8538

Obsidian's CLI dispatcher treats any non-`--` token as a command name, so the shipped `-disable-gpu` in `config/obsidian/user-flags.conf` breaks `obsidian --version` (and any `obsidian <command>`) when CLI is enabled. Chromium itself accepts both `-` and `--`; this does not change GPU behaviour.

Retargeted to `quattro` (4.x). Change is `--disable-gpu`.


mcbbugu:
Retargeted to quattro.

mcbbugu:
Superseded by https://github.com/basecamp/omarchy/pull/8698 — same one-line fix, rebased onto quattro.

mcbbugu:
Superseded by https://github.com/basecamp/omarchy/pull/8700 — same one-line fix, based on quattro.

Issue #8485, opened 2026-08-27
Quickshell Bluetooth panel cannot pair devices that require a displayed passkey


guisaliba:
### System details

- CPU: Intel Core i7-1255U
- Omarchy: 4.0.0-1
- Architecture: x86_64
- Kernel: 7.1.8-arch1-3
- Quickshell package: quickshell-git 0.3.0.r20.g28771c7-2
- Qt runtime: 6.11.2
- Qt base package: qt6-base 6.11.2-2
- Qt QML package: qt6-declarative 6.11.2-1
- Qt Wayland package: qt6-wayland 6.11.2-1
- BlueZ (`bluetoothctl`): 5.87
- `bluez-tools`: 0.2.0-6
- Device: Logitech MX Mechanical Mini
- BlueZ device name: `MX MCHNCL M`

### What's wrong?

_Note: Before upgrading to Quattro, this behavior did not use to occur. The bluetooth panel (opened from the Omarchy menu) managed to display passkeys properly._

The Omarchy Bluetooth panel cannot pair a device that requires the user to type a displayed passkey.

When I tried to pair my keyboard (Logitech MX Mechanical Mini) through the panel, the panel did not show the six-digit passkey. The device state repeatedly changed between paired and connected, but the keyboard did not become usable.

After I completed pairing manually with an interactive `bluetoothctl` `KeyboardDisplay` agent, the panel stopped changing states and the keyboard worked correctly.

### Steps to reproduce

1. Use an unpaired device that requires a displayed passkey, like the aforementioned keyboard model
2. Put it into pairing mode.
3. Click the bluetooth widget in the Quickshell-based status bar to open the Bluetooth panel.
4. Select the device and start pairing.
5. Observe that the panel does not show a passkey.
6. Observe that pairing does not complete.

### Expected result

The panel must show the numeric passkey.

The user types that passkey on the Bluetooth device and (optionally) presses Enter. Pairing then completes, and the device becomes paired, bonded, trusted, and connected.

### Actual result

The panel does not show the passkey. The user cannot complete the keyboard authentication.

During the failed panel attempts, the user service journal contained this prompt:

```text
Aug 26 09:24:42 <hostname> bt-agent[1184]: Authorize this device pairing (yes/no)? Device: MX MCHNCL M (XX:XX:XX:XX:XX:XX)
```

This prompt was only present in the journal. It was not shown in the panel, and the panel did not provide a way to answer it.

### Successful control test

[...]

Issue #7879, opened 2026-08-23
Bluetooth panel treats trusted-but-unbonded devices as paired and never retries pairing


reinierbutot:
### System details

Intel Xeon E-2176M, Intel UHD P630 + NVIDIA Quadro P2000 Mobile, Omarchy 4.0.0-1, bluez 5.87-2

### What's wrong?

If the first Bluetooth pair attempt fails, the panel never pairs that device again.

`omarchy-bluetooth-device pair` ignores a failed `bluetoothctl pair`, then always marks the address trusted:

https://github.com/basecamp/omarchy/blob/quattro/bin/omarchy-bluetooth-device#L33-L38

```bash
pair)
  power_on
  timeout 20s bluetoothctl pair "$address" >/dev/null 2>&1 || true
  trust_device
  timeout 20s bluetoothctl connect "$address" >/dev/null 2>&1 || true
```

The panel then treats `trusted` as already paired. Later clicks only `connect`:

https://github.com/basecamp/omarchy/blob/quattro/shell/plugins/panels/bluetooth/Panel.qml#L276-L279

```qml
function connectDevice(device) {
  if (!device || device.connected) return
  if (device.paired || device.bonded || device.trusted) runDeviceAction(device, "connect", "connecting")
  else runDeviceAction(device, "pair", "connecting")
}
```

Known devices are `paired || bonded || trusted`, and that list is labeled **PAIRED**:

https://github.com/basecamp/omarchy/blob/quattro/shell/plugins/panels/bluetooth/Model.js#L111-L112

BlueZ **Trusted** is not a bond. After a failed pair the device looks remembered, and connect cannot succeed.

Hit this with a Logitech MX Master 2S (BLE HID). `bluetoothctl info` after using the panel:

```
Paired: no
Bonded: no
Trusted: yes
Connected: no
```

Same trap for any device whose first pair fails (not in pairing mode, timeout, etc.).

**Repro**

1. Super+Ctrl+B, Bluetooth on, wait for scan
2. Click a discovered device that is not actually pairable, or whose pair times out
3. Device moves under **PAIRED**; tooltip on the next click is Connect, not Pair
4. Further clicks never pair
5. `bluetoothctl info <addr>` shows Trusted yes, Paired/Bonded no

**Expected**

- Trust only after pairing succeeds
- If a device is not paired/bonded, the next click should pair, even if it is already trusted
- Do not put trusted-but-unbonded devices under **PAIRED**

**Suggested fix**

[...]

PR #6877, opened 2026-08-14
Show Bluetooth pairing passkeys so keyboards can bond

gabriel-laet:
## What

BLE HID keyboards that require a passkey (HHKB Studio and similar) cannot finish pairing from the Omarchy Bluetooth panel.

## Why

Three things stacked:

1. `bt-agent` ran as `NoInputNoOutput`, which auto-accepts Just Works but cannot display a PIN.
2. `omarchy-bluetooth-device pair` trusted the device even when `bluetoothctl pair` failed, then tried to connect.
3. The panel treated `trusted` as already paired, so the next click only connected. The keyboard was left trusted, unpaired, and stuck.

Reproduced on Omarchy 4.0 with an HHKB Studio: the agent looped `Authorize this device pairing (yes/no)?`, pairing timed out, the device stayed `Trusted: yes` / `Paired: no`, and HOG connect failed.

## Change

- New `omarchy-bluetooth-agent` (`KeyboardDisplay`) shows the six-digit passkey via `omarchy-notification-send`. Just Works still auto-accepts.
- Pair only trusts and connects after a successful bond (60s timeout).
- Panel connects only when `paired || bonded`. Trusted-without-a-bond pairs again.
- Migration restarts the user `bt-agent.service` so existing sessions pick up the new agent.

## Test

- `bash test/shell.d/bluetooth-test.sh`
- `bash test/shell.d/systemd-test.sh`
- `./test/cli`

Mice/headphones that already used Just Works are unchanged. Keyboards now get a notification with the code to type on the device.

Fixes #8485, #7879, #9676, #6992

PR #8572, opened 2026-08-27
Keep the Bluetooth pairing agent from skipping late USB adapters

MrBazzieB:
## What

The session `bt-agent` no longer skips forever when `/sys/class/bluetooth` is missing at login. It waits for bluetoothd and the adapter sysfs node, then registers as the default pairing agent. If nothing appears, it exits 0 so machines without Bluetooth do not restart-loop.

## Why

Omarchy's Bluetooth panel pairs with `bluetoothctl pair`. That needs a registered BlueZ agent. The user unit used `ConditionPathIsDirectory=/sys/class/bluetooth` plus `ExecCondition=` on `bluetooth.service`. A failed Condition/ExecCondition **skips** the unit; `Restart=` does not apply.

On a MacBookPro12,1 (Broadcom USB adapter) that looks like:

```
bt-agent.service: skipped, unmet condition check ConditionPathIsDirectory=/sys/class/bluetooth
```

Then pairing from Super+Ctrl+B logs:

```
src/device.c:new_auth() No agent available for request type 2
device_confirm_passkey: Operation not permitted
```

Reproduced pairing Logitech MX Anywhere 3S and Pebble Keys 2 K380s. They advertised fine; the GUI could not confirm the passkey because no agent was running. `bluetoothctl --agent NoInputNoOutput pair …` succeeded immediately.

## Related

This is separate from the trusted-but-unpaired state machine (#7879 / #7880) and from showing keyboard passkeys (#8485 / #6877). Those still matter. Even with them, this skip leaves the panel with no agent on late USB adapters.

#6877 keeps `ConditionPathIsDirectory=/sys/class/bluetooth`, so it would still skip on this hardware.

## Test

- `bash test/shell.d/bluetooth-test.sh`
- `bash test/shell.d/systemd-test.sh`
- `./test/cli`

`./test/all`: bluetooth/systemd tests pass. Remaining failures (`config`, `launch-about`, `snapper`, `unowned-system-paths`, plus flaky `theme-install-guards` / `runtime-smoke`) also fail on `quattro` here without this change (missing `omarchy-pkgs` checkout / compositor).

### DIFF
--- a/bin/omarchy-bluetooth-agent-start
+++ b/bin/omarchy-bluetooth-agent-start
@@ -1,0 +1,33 @@
+#!/bin/bash
+
+# omarchy:summary=Start the BlueZ pairing agent once the adapter exists
+# omarchy:group=bluetooth
+# omarchy:hidden=true
+
+# bt-agent must be the session's default pairing agent or bluetoothctl pair
+# (used by the Omarchy Bluetooth panel) fails with "No agent available".
+# The user unit used to skip forever when /sys/class/bluetooth was missing at
+# graphical-session start. USB adapters (Broadcom on Intel MacBooks, dongles)
+# often appear a few seconds later, so wait for them, then exec bt-agent.
+# If nothing shows up, exit 0 so systemd does not restart-loop on machines
+# without Bluetooth.
+
+set -euo pipefail
+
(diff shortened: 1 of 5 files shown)


bjarneo:
Automated duplication check: this pull request looks similar to #9131, which covers the same Bluetooth agent wait. I keep that one open and close this one to consolidate review.

PR #8581, opened 2026-08-27
Re-arm lock blank timer with backoff on screen changes

### LINKED ISSUE #8580: Lock: monitor that drops HPD after DPMS-off comes back lit and never re-blanks (blank timer is one-shot)
## System details

- Omarchy 4.0.1-1 · Hyprland 0.56.2 (`efb50993780079460b0cbed1363e2166a2de1d9f`) · quickshell 0.3.1-1 · Arch Linux, kernel 7.1.9-arch1-2
- ThinkPad T14 Gen 1 (20UD003LUS), AMD Ryzen 5 PRO 4650U, Renoir iGPU (amdgpu, DCN 2.1)
- `eDP-1` Lenovo 1920x1080 · `DP-1` LG UltraGear 2560x1440 · `HDMI-A-1` ViewSonic VX2758-Series 2560x1440
- Docked, three monitors. Direct connections, no MST.

## What's wrong?

After locking, one external monitor blanks correctly with the other two, then
**turns itself back on ~19 seconds later and stays lit showing the lock screen
indefinitely**. The other two stay dark.

This is *not* one of the existing "display never blanks" reports (#7456,
#7652) — the blank works correctly here. Every output goes dark on schedule at
lock+5s. The problem is that this monitor **leaves and comes back**, and by
then the only blank that will ever run has already run.

## Root cause

Three things compound:

1. **The panel drops hotplug-detect ~12s after DPMS-off.** The ViewSonic
   deep-sleeps once it loses signal, de-asserts HPD, then re-scans its inputs
   ~1s later. This is panel firmware behavior — the LG on `DP-1` does not do it,
   which is why only one monitor is affected.
2. **Hyprland re-adds the returning output enabled and powered on**, applies the
   `monitors.lua` rule, and the session-lock surface paints onto it.
3. **The lock blank is a one-shot.** `idleBlankTimer`
   (`shell/plugins/lock/Service.qml`) fires once at lock+5s and is only ever
   re-armed by `runWake()` or an `authenticatingPassword` transition. Neither
   happens here — no one touched the machine. So nothing re-blanks the output
   that just came back.

`onMonitorsChanged` in the same file already reacts to output changes by
re-requesting the session lock (so the lock surface *is* correctly painted onto
the returning monitor) — it just never re-arms the blank timer alongside it.

## Evidence

Hyprland event socket + polling `hyprctl monitors all -j` and
`/sys/class/drm/card1-*/status` through a real idle lock:

```
08:22:08  lock taken                          all three monitors On
08:22:13  +5s   blank fires                   all three dpms=false, drm disa/Off
08:22:25  +17s  monitorremoved>>HDMI-A-1      drm status → disconnected
08:22:27  +19s  monitoradded>>HDMI-A-1        → conn/enab/On, dpmsStatus=true
08:23:00        still lit, until manual unlock
```

Corresponding journal for the same lock:

```
omarchy lock ...Z lock-requested
omarchy lock ...Z lock-pending: screen-stabilizing
omarchy lock ...Z secure=true
```

Note the blank itself is fine — `hl.dsp.dpms({ action = "disable" })` correctly
took all three outputs to `dpmsStatus=false` and `enabled=disabled`. On this
system the table form works, so this is a distinct bug from #7456.

## Steps to reproduce

Needs a monitor that deep-sleeps and drops HPD on signal loss (a cheap HDMI
panel with input auto-scan is the usual culprit).

1. Dock with at least one such external monitor.
2. Lock (`omarchy system lock`) and do not touch anything.
3. All outputs blank at +5s, as expected.
4. Watch `hyprctl monitors all -j` — around +17s the panel drops off and is
   re-added, powered on, and stays lit.

A short observation window will miss this entirely: the HPD drop is ~12s after
the blank, so anything under ~15s of watching looks like a clean pass.

## Why re-asserting DPMS is not the fix

I prototyped that first and measured it. Re-blanking on `monitoradded` does
**not** converge:

- On re-add it took **9 dispatches over ~4 seconds** before DPMS-off would
  stick — Hyprland keeps re-lighting the output while it is still configuring it.
- Then the panel re-sleeps and returns ~14s later, and the cycle repeats:
  **26 re-blanks in 75 seconds.**

That trades a permanently-lit panel for one that visibly flashes on every ~14
seconds all night. A naive "re-arm the blank timer on monitorsChanged" fix would
land exactly here, so it is worth knowing before picking an approach.

## What does work

Disabling the output for the duration of the lock, rather than blanking it.
Hyprland applies the `disabled = true` rule to the returning output, so it never
re-lights:

```
08:26:50  disabled=true                          output dark
08:27:02  (output absent)  drm=disconnected      panel deep-sleeps, drops HPD
08:27:03  disabled=true    drm=connected         re-added STILL DISABLED, not lit
          ...dark for the remaining 64s, one HPD blip total, no flapping
08:28:07  restored cleanly on unlock
```

One HPD transition instead of a continuous flap, and the panel genuinely powers
off — which is the user-visible behavior people expect from a locked machine.

## Suggested fix

Some way for the lock blank to survive an output that leaves and returns. Rough
options, in the order I'd weigh them:

1. **Re-blank on `monitorsChanged` while locked, but idempotently and with
   backoff** — the existing handler is already there. This needs the retry
   behavior above accounted for, or it will thrash.
2. **Disable outputs for the lock duration** instead of DPMS-off. Strictly more
   robust for this class of panel, but heavier: it destroys and recreates
   `wl_output` globals, which on Hyprland 0.56.2 is the path implicated in
   #7853, and it migrates workspaces.
3. At minimum, **a `logEvent` in `runBlank()`** — #7652 asks for this too. Right
   now there is no journal record of whether a blank ran, which made this
   materially harder to diagnose.

## Workaround

A user-space watcher that disables the affected output for the duration of the
lock and restores it on unlock. Matching by EDID `desc:` rather than connector
name keeps it stable across re-enumeration.

## Related

- #7456 — also "locked screens never power off", but a different cause (dispatch
  form ignored). The table form works on this system.
- #7652 — also "display never blanks", but blank never runs at all there; here it
  runs correctly and is defeated afterwards.
- #7853 — relevant to option 2 above: specific-output `disabled = true` plus
  `hyprctl reload` on 0.56.2 can destroy a `wl_output` global while clients are
  still binding it.
- #7328 — the inverse HPD problem: nothing re-probes when a dock never re-asserts.

## Diagnostics

`omarchy-debug --no-sudo --print`: https://logs.omarchy.org/koIwlR

(No screen capture attached — the bug is a physical panel remaining powered
while the session is locked. A screenshot would just show the normal lock
screen; the event-socket and DRM traces above are the meaningful evidence.)

Sword-Saint69:
When a panel deep-sleeps after DPMS-off and drops HPD (such as input-scanning HDMI displays), Hyprland removes and re-adds the monitor powered on. Because the lock blank timer was a one-shot fired at lock+5s, returning monitors remained lit showing the lock screen indefinitely.

Re-arm idleBlankTimer with exponential/capped backoff (5s -> 15s -> 30s) on screen changes while locked, allowing Hyprland time to finish configuring newly re-added outputs without high-frequency flashing. Reset backoff count on user wake or password input, and log blanking-display in runBlank for journal visibility.

Fixes #8580

### DIFF
--- a/manual/45-troubleshooting.md
+++ b/manual/45-troubleshooting.md
@@ -47,3 +47,10 @@ In order for the rich approval prompt to appear, Settings > Advanced > Use Hardw
  ![troubleshooting-1password](images/troubleshooting-1password.webp)
 
 Or if you haven't launched 1Password since booting up, the prompt will not appear.
+
+### Why does my monitor turn off and on (flash) continuously while locked?
+
+Some monitors (especially HDMI panels with Input Auto-Scan or Auto Source enabled) shut down their signal receiver when DPMS turns the display off, causing the system to detect a Hot Plug Detect (HPD) drop. When the monitor re-scans inputs, Hyprland re-detects the output and powers it back on.
+
+To eliminate this loop, open your monitor's built-in OSD menu and set **Auto Source** or **Input Auto-Scan** to **Off** (pinning the input source directly to HDMI or DisplayPort).
+
--- a/shell/plugins/lock/Service.qml
+++ b/shell/plugins/lock/Service.qml
@@ -33,6 +33,7 @@ Item {
   property string lastEventAt: ""
   property bool strandedLock: false
   property bool strandedLockResolved: false
+  property int screenChangeBlankCount: 0
 
   readonly property bool locked: lockRequested || sessionLock.locked || sessionLock.secure
   readonly property bool authenticating: authenticatingPassword || fingerprintAuthenticating
@@ -133,6 +134,7 @@ Item {
 
     resetAuthenticationState()
     lockRequested = true
+    screenChangeBlankCount = 0
     armBlankTimer()
     logEvent("lock-requested")
     queueSessionLock()
@@ -159,17 +161,21 @@ Item {
     runWake()
   }
 
-  function armBlankTimer() {
+  function armBlankTimer(customInterval) {
+    var ms = (typeof customInterval === "number" && customInterval > 0) ? customInterval : 5000
+    idleBlankTimer.interval = ms
     idleBlankTimer.armedAt = Date.now()
     idleBlankTimer.restart()
   }
 
   function runWake() {
+    screenChangeBlankCount = 0
     if (!wakeProcess.running) wakeProcess.running = true
     if (lockRequested) armBlankTimer()
   }
 
   function runBlank() {
+    logEvent("blanking-display")
     if (!blankProcess.running) blankProcess.running = true
   }
 
@@ -472,6 +478,14 @@ Item {
       // A monitor still coming up has no workspace, so cannot answer yet.
       strandedLockRetryTimer.rearm()
       root.checkStrandedLock()
+
+      if ((root.locked || root.lockRequested) && !root.authenticatingPassword) {
+        root.screenChangeBlankCount += 1
+        var delay = 5000
+        if (root.screenChangeBlankCount === 2) delay = 15000
+        else if (root.screenChangeBlankCount >= 3) delay = 30000
+        root.armBlankTimer(delay)
+      }
     }
   }
 
--- a/test/shell.d/lock-blank-fingerprint-test.sh
+++ b/test/shell.d/lock-blank-fingerprint-test.sh
@@ -37,13 +37,8 @@ assert(
 )
 
 assert(
-  /onScreensChanged\(\) \{[\s\S]*root\.screenChangeBlankCount \+= 1/.test(serviceQml),
-  'screen changes while locked increment the screen change count'
-)
-
-assert(
-  /onScreensChanged\(\) \{[\s\S]*root\.armBlankTimer\(delay\)/.test(serviceQml),
-  'screen changes while locked re-arm the blank timer with backoff'
+  /onScreensChanged\(\) \{[\s\S]*if \(\(root\.locked \|\| root\.lockRequested\) && !root\.authenticatingPassword\) \{[\s\S]*root\.screenChangeBlankCount \+= 1\s*var delay = 5000\s*if \(root\.screenChangeBlankCount === 2\) delay = 15000\s*else if \(root\.screenChangeBlankCount >= 3\) delay = 30000\s*root\.armBlankTimer\(delay\)/.test(serviceQml),
+  'screen changes re-arm blanking at 5s, 15s, then cap at 30s only while locked and not authenticating'
 )
 
 assert(
--- a/test/shell.d/lock-blank-fingerprint-test.sh
+++ b/test/shell.d/lock-blank-fingerprint-test.sh
@@ -30,4 +30,24 @@ assert(
   !/onAuthenticatingChanged:/.test(serviceQml),
   'the combined authenticating state no longer drives the blank timer'
 )
+
+assert(
+  /function runBlank\(\) \{\s*logEvent\("blanking-display"\)/.test(serviceQml),
+  'runBlank logs display blanking event'
+)
+
+assert(
+  /onScreensChanged\(\) \{[\s\S]*root\.screenChangeBlankCount \+= 1/.test(serviceQml),
+  'screen changes while locked increment the screen change count'
+)
+
+assert(
+  /onScreensChanged\(\) \{[\s\S]*root\.armBlankTimer\(delay\)/.test(serviceQml),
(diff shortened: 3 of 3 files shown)


Sword-Saint69:
@dhh @ryanrhughes @omarchybot please review this pr and one thing i want to say is really loving quattro and became a huge fan of it keep up the good work

Sword-Saint69:
@dhh @ryanrhughes @omarchybot please review this pr i have fixed the suggestions

Chulii:
Measured this on affected hardware (input-scanning HDMI panel that drops HPD after DPMS-off — full analysis in #8580): the backoff works as coded but the re-add cycle never stops, so the panel still flashes on every cycle — blank events 39s apart at the 5/15/30s backoff, lit for the whole backoff window each time. It also stays lit whenever the user is at the password prompt: keystrokes fire the wake path (DPMS enable-all), and an in-flight password check stops the blank timer.

Two more constraints from the same hardware that any re-arm design runs into: a re-added output comes up enabled and re-lights *other* DPMS-off outputs as well, and while a session lock is active Hyprland refuses to disable an output (`hyprctl eval 'hl.monitor({disabled=true})'` does not persist a rule across the re-add, and `hyprctl keyword` is gone in 0.56) — so re-darkening after the re-add can't be made reliable either.

What did stop the flashing on this machine with zero code: the panel's input auto-scan was the trigger. **Auto Source → Off** in the monitor OSD (input pinned) eliminated the HPD cycling entirely on stock quattro. Might be worth asking the #8580 reporter what their ViewSonic exposes in its OSD — if it has an equivalent setting, that's the per-machine fix, though panels without such a setting will still need a software answer.

The `logEvent("blanking-display")` addition in this PR is great regardless — it made the whole behavior measurable.

Sword-Saint69:
@Chulii Thanks for testing it i will surely provide the fix in the next push just wait

Sword-Saint69:
Thanks a lot for testing this on affected hardware and sharing such a thorough breakdown @Chulii!

The empirical measurements and confirmation of `logEvent("blanking-display")` are super helpful. You're spot on regarding the Wayland/DRM constraints: because `ext-session-lock-v1` prevents output disable rules and kernel DRM defaults newly attached HPD outputs to powered ON, user-space lock managers can't prevent the monitor firmware from lighting the panel when it re-asserts HPD.

The 5s → 15s → 30s backoff in this PR serves as the safest software fallback so returning displays don't stay lit indefinitely while avoiding ultra-rapid DPMS mode-switch thrashing.

For monitors that continuously drop/poll HPD upon DPMS-off, pinning the input by setting **Auto Source / Input Auto-Scan → Off** in the monitor OSD is indeed the permanent per-machine fix. I'll note this OSD workaround in #8580 for future reference!

Sword-Saint69:
Thanks a lot for the detailed testing and analysis @Chulii!

Here is a summary of the behavior and why the current backoff design is the safest software approach:

### 1. Primary User-Side Fix (Monitor OSD)
As you noted, setting **Auto Source / Input Auto-Scan → Off** in the monitor’s built-in OSD menu (pinning the input directly to HDMI/DP) completely eliminates the HPD drop/re-scan loop on DPMS-off with **zero code**. We have now documented this in Omarchy's user manual (`manual/45-troubleshooting.md`).

### 2. Software Safety Net (PR #8580)
- **Why pure timer backoff vs. immediate 0ms re-blanking**: When a monitor asserts HPD, Hyprland needs ~500ms–2000ms to modeset, create `WlSessionLockSurface` on the new screen, and allocate geometries. Triggering DPMS-off immediately on `onScreensChanged` while locked forces DPMS-off mid-modeset, which can cause surface creation failures, screen glitches, or 100% CPU infinite HPD mode-switch loops.
- **Backoff schedule**: The 5s → 15s → 30s schedule allows output surfaces to stabilize while preventing infinite high-frequency DPMS thrashing, while ensuring returning displays do re-darken eventually rather than staying lit indefinitely.

### 3. Upstream Context (Hyprland / wlroots)
The underlying root cause is compositor/DRM level: newly re-added outputs default to powered ON in kernel DRM. Retaining output DPMS state across hotplug re-adds in Hyprland / wlroots would be the ultimate compositor-level resolution, and is worth tracking upstream.

Sword-Saint69:
@dhh @ryanrhughes please review it check if its viable happy to do contributions on omarchy

PR #8584, opened 2026-08-27
Theme Files and GTK apps from colors.toml

robertkokenyesi:
## What

Theme GNOME Files (Nautilus) and other GTK apps from the same `colors.toml` pipeline as Chromium, terminals, and Obsidian.

Stock themes currently only set Adwaita light/dark plus a Yaru icon variant. Files never picked up the palette (Osaka Jade stayed generic Adwaita while the rest of the desktop went green). Extra themes that already ship `gtk.css` were copied into `current/theme/` and never installed where GTK reads them.

## How

Same shape as `omarchy-theme-set-browser`:

- `default/themed/gtk.css.tpl` maps the palette onto Adwaita named colors
- A theme's own `gtk.css` still wins when present
- `omarchy-theme-set-gtk` copies that file to `~/.config/gtk-3.0/gtk.css` and `gtk-4.0/gtk.css`
- GNOME `accent-color` is set from the theme accent hue
- If a Files **window** is mapped, it is reloaded so the new colors apply immediately
- If Files is only a background process (or not open), it is left closed

## Test plan

- `./test/shell.d/gtk-theme-test.sh`
- Switch Osaka Jade / Tokyo Night / Matte Black with Files closed: Files stays closed; next open uses the new colors
- Same with Files open: the window retints
- Light themes still follow `mode = "light"`

This change was **100% vibe coded with Grok 4.6 Build**.


bjarneo:
Automated duplication check: this pull request looks similar to #8408, which covers the same GTK theme css generation. I keep that one open and close this one to consolidate review.

PR #8587, opened 2026-08-27
Show notification dismiss button without hover

### LINKED ISSUE #7711: Crash capture notification cannot be dismissed without launching AI diagnosis
### System details

Latitude 7390 2-in-1, Omarchy 4.0.0-1

### What's wrong?

When a process crashes, Omarchy shows a critical “Process crashed” notification. Clicking it launches `omarchy-agent-crash`, which is expected.

However, if I do not want to start the AI diagnosis, there is no visible way to close/dismiss the notification from the notification itself. Since it is critical, it stays on screen.

Expected: the notification should have a clear dismiss/close affordance, or a separate “Diagnose” action instead of making the whole notification click launch AI.

mgutierrezerpi:
## Summary

- keep the notification close control visible without requiring hover
- preserve the existing automatic notification timeouts and click behavior
- cover the always-visible close control with the notification shell test

Fixes #7711.

## Testing

- `bash test/shell.d/notifications-test.sh`

### DIFF
--- a/shell/plugins/notifications/components/NotificationCard.qml
+++ b/shell/plugins/notifications/components/NotificationCard.qml
@@ -154,7 +154,7 @@ BorderSurface {
       ColumnLayout {
         Layout.fillWidth: true
         Layout.alignment: Qt.AlignVCenter
-        // Keep the first line clear of the hover-revealed close button.
+        // Keep the first line clear of the close button.
         Layout.rightMargin: Style.space(10)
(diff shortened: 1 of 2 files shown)


mgutierrezerpi:
Closing this draft because the upstream notification card already includes a hover-revealed dismiss control. My installed Omarchy version had not received that change yet, which is why the control was missing locally.

Issue #6957, opened 2026-08-15
omarchy font set resets Foot font size to 9, ignoring omarchy display text size


sbalbalosa:
## Summary

Changing the system font family — via the settings menu (**Style → Font**) or `omarchy font set` — silently resets **Foot's** font size to a hardcoded **9pt**, undoing the global text size set with `omarchy display text size` (or the display-panel text-size slider).

Only Foot is affected. kitty, alacritty, and ghostty correctly keep their existing size when the family changes.

## Steps to reproduce

1. `omarchy display text size 16`
2. Confirm Foot is scaled: `grep '^font=' ~/.config/foot/foot.ini` → `font=JetBrainsMono Nerd Font:size=12`
3. Change the family: `omarchy font set MesloLGLDZ Nerd Font` (or the menu **Style → Font** picker)
4. Check Foot again: `grep '^font=' ~/.config/foot/foot.ini` → `font=MesloLGLDZ Nerd Font:size=9` (size lost)

Meanwhile kitty/alacritty/ghostty remain at 12, so Foot is visibly smaller than the rest.

## Expected

Only the font family changes; the `:size=N` chosen by `omarchy display text size` is preserved — same as the other three terminals.

## Cause

In `bin/omarchy-font-set`, the Foot branch rewrites the entire `font=` line with a hardcoded size:

```bash
sed -i "s/^font=.*/font=$font_name:size=9/g" ~/.config/foot/foot.ini
```

kitty/alacritty/ghostty branches only rewrite their family line, so their size survives. The two font commands fight each other: picking a font in the GUI silently undoes the text-size setting.

## Environment

- Omarchy 4.0.0-1 (quattro)
- Hyprland 0.56.2, Dell Latitude 5420 (Intel i5-1145G7)
- Reproduced via both CLI and the settings-menu font picker

leodutra:
This also happened to me during some `omarchy update`.

PR #8597, opened 2026-08-27
Preserve alacritty font styles when changing font family

ParadokS81:
`omarchy font set` deletes the `style` fields from `~/.config/alacritty/alacritty.toml`.

The family substitution matches greedily:

```bash
sed -i "s/family = \".*\"/family = \"$font_name\"/g" ~/.config/alacritty/alacritty.toml
```

`.*` runs past the closing quote to the last `"` on the line, so it consumes `", style = "Regular` along with the family. Omarchy's own `config/alacritty/alacritty.toml` sets those styles, so a stock install loses all three on the first font change and alacritty stops explicitly selecting the bold and italic faces.

### Reproduce

Against the shipped template:

```
BEFORE (config/alacritty/alacritty.toml)
  normal = { family = "JetBrainsMono Nerd Font", style = "Regular" }
  bold   = { family = "JetBrainsMono Nerd Font", style = "Bold" }
  italic = { family = "JetBrainsMono Nerd Font", style = "Italic" }

AFTER `omarchy font set "Liberation Mono"`
  normal = { family = "Liberation Mono" }
  bold   = { family = "Liberation Mono" }
  italic = { family = "Liberation Mono" }
```

### Fix

Match the quoted family with a negated character class so the substitution stops at the closing quote.

### Tests

Added a regression test to `test/cli` that writes the shipped font table into a temporary `HOME`, runs `omarchy-font-set`, and asserts all three styles survive while the family is still replaced. Verified it fails without the one-line fix and passes with it.

`./test/cli` passes (118 ok). `./test/all` shows 6 failures on my machine that are unrelated and present without this change too — they are environmental (multi-monitor widget counts, a missing `omarchy-pkgs` checkout, a network-dependent theme-install check).

### Related

#6957 / #6958 cover the foot font-size reset in this same script. Different line, no overlap with this change.

### DIFF
--- a/bin/omarchy-font-set
+++ b/bin/omarchy-font-set
@@ -27,7 +27,7 @@ if ! fc-list | grep -Fqi -- "$font_name"; then
 fi
 
 if [[ -f ~/.config/alacritty/alacritty.toml ]]; then
-  sed -i "s/family = \".*\"/family = \"$font_name\"/g" ~/.config/alacritty/alacritty.toml
+  sed -i "s/family = \"[^\"]*\"/family = \"$font_name\"/g" ~/.config/alacritty/alacritty.toml
 fi
 
 if [[ -f ~/.config/kitty/kitty.conf ]]; then
--- a/test/cli
+++ b/test/cli
@@ -253,11 +253,13 @@ pass "safe dispatch works for font current"
 # the closing quote to the end of the line, deleting all three.
 make_tmpdir FONT_STYLE_TMPDIR
 mkdir -p "$FONT_STYLE_TMPDIR/.config/alacritty"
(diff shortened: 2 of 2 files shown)


ParadokS81:
Addressed the three review notes in the follow-up commit:

- downstream commands (`omarchy-restart-shell`, `omarchy-notification-send`, `omarchy-hook`, `pgrep`) are stubbed in a fake bin, and the command's exit status is now asserted instead of `|| true`
- the fixture starts from a family fontconfig cannot resolve, so the family assertion only passes if the file was rewritten
- all three `family =` entries are counted, not just one

`./test/cli` passes (119 ok). With the sed fix reverted, the style assertion fails as intended.

PR #8602, opened 2026-08-27
Deduplicate Codex usage across Pi forked sessions

### LINKED ISSUE #8564: Codex usage is counted again after forking or cloning a Pi session
### System details

Lenovo ThinkBook 14 G3 ACL, AMD Ryzen, Omarchy 4.0.1-1, Pi 0.84.3, pi-subagents 0.56.0, `openai-codex` provider

### What's wrong?

After a Pi session is forked or cloned, the Codex usage shown in Omarchy's Agents panel includes the inherited historical usage from both the parent session file and the forked session file.

Creating the fork copies existing Pi message entries, including their already-recorded usage metadata, into a new session file. Those copied records do not represent new Codex responses. If work then continues in the fork, only the newly generated responses and their newly reported usage should be additional usage.

This can significantly inflate the local Prompt, Token, and Session statistics when multiple subagents use forked context.

#### Steps to reproduce

1. Start a Pi session using the `openai-codex` provider.
2. Exchange enough messages to produce recorded token usage.
3. Fork or clone the session using Pi, or launch a subagent with `context: "fork"`.
4. Refresh the Omarchy Agents panel.
5. Compare the reported local usage before and after the fork.

#### Expected behavior

Usage metadata attached to messages copied from the parent should be counted only once. New assistant responses generated inside the fork should still add their newly reported usage normally.

#### Actual behavior

The inherited assistant-message records are counted again for every forked session file.

On my machine, three forked sessions contained parent-derived historical usage totaling:

- `15,716,982` tokens
- `87,744,714` tokens
- `87,744,714` tokens

This caused approximately `191,206,410` tokens of local usage to be counted more than once.

I verified these as inherited copies by following each session header's `parentSession` path and comparing the copied message IDs, timestamps, content, model, and usage data with the declared parent session.

#### Technical observation

Pi forked-session headers contain a `parentSession` field. Inherited message entries retain their original IDs, timestamps, contents, and usage records.

The current collector builds its deduplication key using both the file path and message ID:

```python
message_key = path + ":" + str(entry.get("id") or "")
```

Because every fork has a different file path, inherited copies are treated as separate usage records. This is only a diagnostic observation; the appropriate deduplication strategy is left to the maintainers.

arisgysel-design:
## Summary

- count inherited Pi messages only once across parent and fork session files
- keep new fork responses and unrelated short-ID collisions distinct
- derive active sessions from Pi's `parentSession` lineage so scan order cannot change session totals

## Testing

- `bash test/shell.d/agent-usage-codex-scanner-test.sh`
- `python3 -m py_compile bin/omarchy-agent-usage-codex`
- `bash bin/omarchy commands --check` (437 commands)

Fixes #8564

### DIFF
--- a/bin/omarchy-agent-usage-codex
+++ b/bin/omarchy-agent-usage-codex
@@ -135,12 +135,29 @@ def add_usage(day, session_key, model, input_tokens, output_tokens, cache_read, 
     today_tokens_by_model[model] = today_tokens_by_model.get(model, 0) + total
 
 
+def pi_parent_session(path):
+  try:
+    with open(path, encoding="utf-8", errors="replace") as session_file:
+      header = json.loads(session_file.readline())
+  except (OSError, json.JSONDecodeError):
+    return None
+
+  if not isinstance(header, dict):
+    return None
+  parent = header.get("parentSession")
+  if not isinstance(parent, str) or not parent:
+    return None
+  return os.path.abspath(os.path.expanduser(parent))
+
+
 def scan_pi_sessions():
   roots = [
     Path.home() / ".pi" / "agent" / "sessions",
     Path.home() / ".omp" / "agent" / "sessions",
   ]
   rg = find_command("rg") or "rg"
+  session_keys = {}
+  today_session_keys = {}
   for root in roots:
     if not root.exists():
       continue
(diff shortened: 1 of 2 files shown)


Brams-s:
Related native-collector correction: https://github.com/omacom/omarchy/pull/10531 skips repeated `token_count` notifications carrying an unchanged cumulative snapshot. I reproduced one 110-token request becoming 220 tokens in current quattro.

It stays within `scan_native_codex_sessions()` plus cache invalidation and tests; Pi lineage/deduplication remains with this PR. I checked that your current patch applies cleanly alongside it. Linking them so the two distinct duplication cases can be reviewed together without combining their implementations.

PR #8605, opened 2026-08-27
Cycle backgrounds through a symlinked backgrounds directory

### LINKED ISSUE #8594: `omarchy theme bg next` never advances when ~/.config/omarchy/backgrounds/<theme> is a symlink
### System details

Omarchy 4.0.0-1 · AMD Ryzen 5 7600X · NVIDIA RTX 3070 (+ AMD Raphael iGPU)

### Description

If the user backgrounds directory `~/.config/omarchy/backgrounds/<theme>/` is a symlink (e.g. into a stow/chezmoi-managed dotfiles repo), `omarchy theme bg next` always selects the alphabetically first image and never advances. The same index lookup in `omarchy-theme-set` (`choose_theme_background`) has the identical problem.

Root cause is a path-canonicalisation mismatch between two scripts:

- `bin/omarchy-theme-bg-set:12` stores the current background as `realpath "$1"`, i.e. with the symlink **resolved** (`/home/u/dotfiles/.../foo.jpg`).
- `bin/omarchy-theme-bg-next` builds the candidate list with `find -L "$HOME/.config/omarchy/backgrounds/$THEME_NAME/"`, which yields **unresolved** paths (`/home/u/.config/omarchy/backgrounds/<theme>/foo.jpg`), then does a plain string compare against `readlink` of the state symlink.

The two strings never match, so `INDEX` stays `-1` and the script falls through to `BACKGROUNDS[0]` on every invocation. The first call "works"; every call after that is a no-op.

The background switcher is unaffected because it calls `bg-set` with an explicit file. Stock theme backgrounds are unaffected because `current/theme/backgrounds/` is a real copied directory.

### Steps to reproduce

Isolated, without touching the live setup (stubs `omarchy-shell` so the wallpaper isn't actually changed):

```bash
S=$(mktemp -d)
mkdir -p "$S/home/.local/state/omarchy/current/theme/backgrounds" \
         "$S/home/.config/omarchy/backgrounds" "$S/real-walls" "$S/stubs"
echo t > "$S/home/.local/state/omarchy/current/theme.name"
for f in a b c; do printf x > "$S/real-walls/$f.jpg"; done
printf '#!/bin/sh\nexit 0\n' > "$S/stubs/omarchy-shell"
cp "$S/stubs/omarchy-shell" "$S/stubs/omarchy-notification-send"; chmod +x "$S/stubs"/*

ln -s "$S/real-walls" "$S/home/.config/omarchy/backgrounds/t"
for i in 1 2 3 4; do
  HOME="$S/home" PATH="$S/stubs:$PATH" omarchy-theme-bg-next
  readlink "$S/home/.local/state/omarchy/current/background"
done
```

Or on a real install: `ln -s ~/Pictures/walls ~/.config/omarchy/backgrounds/catppuccin`, run `omarchy theme bg next` a few times, watch `omarchy theme bg current`.

**Expected:** cycles `a.jpg → b.jpg → c.jpg → a.jpg`, as it does when the directory is real (verified with `cp -r` in place of `ln -s`).

**Actual:** `a.jpg` four times.

### Suggested fix

Compare canonical paths on both sides, in `bg-next` and in `choose_theme_background`:

```bash
CURRENT_BACKGROUND=$(readlink -f "$CURRENT_BACKGROUND_LINK")
...
if [[ $(realpath "${BACKGROUNDS[$i]}") == "$CURRENT_BACKGROUND" ]]; then
```

Alternatively drop `realpath` from `bg-set` — but resolving on the comparison side is more robust (also covers `bg set` with a relative path).

Related but distinct: #7668 (`choose_theme_background` fails the same lookup on theme switch, for a different reason). Both would benefit from a canonical-path comparison.

thorhj:
Fixes #8594.

`omarchy theme bg next` never advances when `~/.config/omarchy/backgrounds/<theme>` is a symlink — the kind a dotfiles manager leaves behind. Every invocation re-selects the alphabetically first image.

Two scripts spell the same path differently:

- `omarchy-theme-bg-set` links the background's **resolved** path (`realpath`), so `current/background` points at `/home/u/dotfiles/.../foo.jpg`.
- `omarchy-theme-bg-next` and `choose_theme_background` in `omarchy-theme-set` list candidates as `find -L` reaches them, i.e. **unresolved** (`/home/u/.config/omarchy/backgrounds/<theme>/foo.jpg`), then string-compare against `readlink` of that link.

The two never match, so the index lookup returns -1 and both commands fall through to `backgrounds[0]`. Stock theme backgrounds are unaffected, because `current/theme/backgrounds/` is a real copied directory.

This resolves both sides before comparing: the candidate list goes through `realpath`, and the link is read with `readlink -f`. Sorting still happens on the paths as found, so ordering by filename is unchanged. `omarchy-menu-images` already did the equivalent via `find -samefile`, so the background switcher needed no change.

Also quotes the comparison's right-hand side in `choose_theme_background`, which was an unquoted variable in `[[ ]]` and so a glob match rather than a string compare.

## Verification

Live, on a machine where `~/.config/omarchy/backgrounds` is a symlink into a dotfiles repo — stock, then patched:

```
--- stock ---            --- patched ---
  -> 3d-model.jpg          -> asian-village.png
  -> 3d-model.jpg          -> beach.jpg
  -> 3d-model.jpg          -> black-hole.png
```

Three tests added to `test/cli`: cycling through a symlinked backgrounds directory, through a real one (regression guard), and `omarchy-theme-set` resuming the cycle from a link that `omarchy-theme-bg-set` wrote. Each fails against the unpatched scripts and passes with them.

Also checked by hand: an empty backgrounds directory still sends the "No background was found for theme" notification, stow-style per-file symlinks cycle, and filenames containing spaces cycle.

`./test/cli` passes. `./test/shell` has 3 failures on this branch (`config-test.sh`, `snapper-test.sh`, `unowned-system-paths-test.sh`) that also fail on a clean `quattro` checkout, so they are unrelated.

### DIFF
--- a/bin/omarchy-theme-bg-next
+++ b/bin/omarchy-theme-bg-next
@@ -8,10 +8,11 @@ THEME_BACKGROUNDS_PATH="$HOME/.local/state/omarchy/current/theme/backgrounds/"
 USER_BACKGROUNDS_PATH="$HOME/.config/omarchy/backgrounds/$THEME_NAME/"
 CURRENT_BACKGROUND_LINK="$HOME/.local/state/omarchy/current/background"
 
+# Resolve candidates, since the current background symlink stores a real path
 mapfile -d '' -t BACKGROUNDS < <(
   find -L "$USER_BACKGROUNDS_PATH" "$THEME_BACKGROUNDS_PATH" -maxdepth 1 -type f \
     \( -iname '*.jpg' -o -iname '*.jpeg' -o -iname '*.png' -o -iname '*.gif' -o -iname '*.bmp' -o -iname '*.webp' \) \
-    -print0 2>/dev/null | sort -z
+    -print0 2>/dev/null | sort -z | xargs -0 -r realpath -z --
 )
 TOTAL=${#BACKGROUNDS[@]}
 
@@ -20,7 +21,7 @@ if (( TOTAL == 0 )); then
 else
   # Get current background from symlink
   if [[ -L $CURRENT_BACKGROUND_LINK ]]; then
-    CURRENT_BACKGROUND=$(readlink "$CURRENT_BACKGROUND_LINK")
+    CURRENT_BACKGROUND=$(readlink -f "$CURRENT_BACKGROUND_LINK")
   else
     # Default to first background if no symlink exists
     CURRENT_BACKGROUND=""
--- a/bin/omarchy-theme-set
+++ b/bin/omarchy-theme-set
@@ -74,18 +74,19 @@ choose_theme_background() {
   local current_background index next_index i
 
   CHOSEN_THEME_BACKGROUND=""
+  # Resolve candidates, since the current background symlink stores a real path
   mapfile -d '' -t backgrounds < <(
     find -L "$HOME/.config/omarchy/backgrounds/$THEME_NAME/" "$CURRENT_THEME_PATH/backgrounds/" -maxdepth 1 -type f \
       \( -iname '*.jpg' -o -iname '*.jpeg' -o -iname '*.png' -o -iname '*.gif' -o -iname '*.bmp' -o -iname '*.webp' \) \
-      -print0 2>/dev/null | sort -z
+      -print0 2>/dev/null | sort -z | xargs -0 -r realpath -z --
   )
 
   (( ${#backgrounds[@]} > 0 )) || return 1
 
-  current_background=$(readlink "$CURRENT_BACKGROUND_LINK" 2>/dev/null || true)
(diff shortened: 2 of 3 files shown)


AccursedGalaxy:
I ran into this myself yesterday and was just about to open an issue and PR for it. But now you did :) Thanks!

Tested your branch on my system and everything works for me.

WhiskeyTuesday:
Noticed your quoting change on that comparison line also fixes a separate case: a filename holding glob characters (`wall [4k].jpg`) never matched itself, so cycling fell through to the first image.

There's a regression test for it in #8865.

PR #8609, opened 2026-08-27
Tell contributors to check for duplicates and the right branch

ParadokS81:
`contributing.md` sends contributors straight from "found a bug" to filing. Two steps in between prevent wasted work on both sides, and I hit both writing #8597.

### Check for an existing report

Searching only issues misses open PRs. I found two bugs in `omarchy-font-set`; one was already #6957 with PR #6958 open and a better fix than I would have written. Searching issues alone would not have surfaced it.

### Verify against the default branch

`master` is not where development happens and can be weeks behind a shipped release:

```
default branch   2026-08-27   (current)
master           2026-08-14   (13 days stale, older than the installed 4.0.1)
```

I fetched `bin/omarchy-font-set` from `master` to check whether a bug still existed and got a version *older than my own installed copy* — missing the fontconfig rewrite that had already shipped. A bug "confirmed" that way may already be fixed, or simply not be the code users run.

The added section teaches the check (`gh repo view --json defaultBranchRef`) rather than naming today's branch, so it stays correct when the branch changes.

### Test guidance

`AGENTS.md` asks for "focused automated tests for the area you changed"; this file said to run `./test/all` before pushing, and calls `AGENTS.md` the authority. On my machine `./test/all` reports 6 failures on a clean tree — attached-monitor counts, a missing sibling checkout, a network-dependent check — none related to any change. Read as a gate, that blocks valid PRs. Aligned the wording with `AGENTS.md` and noted how to tell an environmental failure from a real one.

Docs only; no behaviour change.

### DIFF
--- a/default/agents/skills/omarchy/contributing.md
+++ b/default/agents/skills/omarchy/contributing.md
@@ -29,10 +29,11 @@ If a PR is already open, add to that discussion instead of filing again.
 
 ## Verify Against the Default Branch
 
-Check the bug against the branch development actually happens on, not against
-whatever `git clone` or a raw URL hands you. `master` is not it, and can be
-weeks behind a shipped release — so a bug "confirmed" there may already be
-fixed, or the code may not match what users are running:
+Check the bug against the branch development actually happens on. A fresh
+`git clone` lands there, but an older checkout or a raw URL pinned to
+`master` does not — `master` can be weeks behind a shipped release, so a bug
+"confirmed" there may already be fixed, or the code may not match what users
+are running:
 
 ```bash
 gh repo view basecamp/omarchy --json defaultBranchRef --jq .defaultBranchRef.name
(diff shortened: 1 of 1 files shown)


ParadokS81:
Both notes addressed: the stash step now says `git stash -u` / `git stash pop`, and the default-branch paragraph no longer claims a fresh `git clone` lands on `master` — it names the actual traps (an older checkout, or a raw URL pinned to `master`).

rdoupe:
Worth knowing before this gets reviewed: the guidance here is right, but `default/agents/skills/omarchy/contributing.md` is not read by the agents it is aimed at.

- It lives inside the `omarchy` skill, whose description says it *excludes Omarchy source development*. An agent asked to write a fix decides the skill does not apply and never opens this file. That is exactly what happened in my session: skill installed, four open PRs for the bug I was fixing, and I filed a fifth (#9089, closed) without ever loading `contributing.md`.
- #6718 deletes `contributing.md` entirely and routes upstream work to the repo's `AGENTS.md` instead.

So for the PR-submission path, `AGENTS.md` is the file that actually gets read. I opened #9095 putting the search-first guidance there. The issue-reporting path is a different story: an agent reporting a bug *does* load the skill, so guidance in the skill's reporting guide (post-#6718, `reporting-issues.md`) is well placed for that.

PR #4176, opened 2026-01-09
fix: Map Logitech MX Keys S special keys to equivalent Omarchy commands

axelfontaine:
This detects Logitech MX Keys S keyboards and maps their special keys onto the equivalent Omarchy commands.

All keys that previously weren't functional now work out of the box as expected:
- Dictation
- Emoji
- Mute Microphone
- Screen Capture
- Screen Lock

PR #8610, opened 2026-08-27
Logitech MX Keys (regular, S, Mini) support

axelfontaine:
While many special keys work correctly out of the box, some don't. This PR maps all broken or unbound special keys to their respective Omarchy commands in addition to making them easily overridable in user bindings. No changes to existing bindings. No heavyweight software like Solaar. Just a simple 1 MiB dependency on `keyd`, only running when one of these keyboards is found.

Supersedes:
- #4091
- #4176
- #5301

## MX Keys
image

| Key | Chord sent by the keyboard | Action Before | Action After |
| --- | --- | --- | --- |
| Fn+F4 | `SUPER+A` | nothing — unbound | Toggle window floating/tiling |
| Fn+F5 | `SUPER+D` | nothing — unbound | Toggle scratchpad |
| Lock (above numpad) | `SUPER+L` | "Toggle workspace layout" | `omarchy-system-lock` |

## MX Keys S
image

| Key | Chord sent by the keyboard | Action Before | Action After |
| --- | --- | --- | --- |
| Dictation | `SUPER+H` | nothing — unbound | `voxtype record toggle` |
| Emoji | `SUPER+CTRL+ALT+SHIFT+Space` | nothing — unbound | `omarchy-shell shell toggle omarchy.emojis` |
| Mute mic | bare `F13` | nothing — unbound | `omarchy-audio-input-mute` |
| Screenshot | `SUPER+SHIFT+S` | Google Maps web app | `omarchy-capture-screenshot` |
| Lock | `SUPER+L` | "Toggle workspace layout" | `omarchy-system-lock` |

## MX Keys Mini
image

| Key | Chord sent by the keyboard | Action Before | Action After |
| --- | --- | --- | --- |
| Dictation | `SUPER+H` | nothing — unbound | `voxtype record toggle` |
| Emoji | `SUPER+CTRL+ALT+SHIFT+Space` | nothing — unbound | `omarchy-shell shell toggle omarchy.emojis` |
| Mute mic | bare `F13` | nothing — unbound | `omarchy-audio-input-mute` |
| Screenshot | `SUPER+SHIFT+S` | Google Maps web app | `omarchy-capture-screenshot` |
| Lock | `SUPER+L` | "Toggle workspace layout" | `omarchy-system-lock` |

## Approach

Detect an MX Keys, MX Keys S or MX Keys Mini — the MX Keys pairs only over Bluetooth or a Logitech Unifying receiver, the S only over Bluetooth or a Logi Bolt receiver, and the Mini (verified so far over Bluetooth only) sends the exact same action-row chords as the S — and, **only then**, set up [keyd](https://github.com/rvaiya/keyd) to remap that keyboard's problem chords to spare `F13`–`F19` keycodes, and bind those to the matching Omarchy command.

Detection differs by transport but stays cheap. A Bluetooth link, and a Unifying receiver's per-device hidraw node, carry the model name — and, as a fallback for an MX Keys asleep when the kernel first resolved it, its wireless PID `046d:408a` — straight in sysfs. Behind the Bolt receiver the name is hidden, so each paired slot is instead asked its name over HID++ 2.0 (feature 0x0005); the MX Keys never reaches this path since it can't pair to Bolt at all. The Mini needed no new detection code at all: the model-name match already saw "MX Keys" inside "MX Keys Mini", so it had been excluded on purpose pending real hardware — a community member (see PR discussion) verified it chord-for-chord against the S's rules, so the exclusion just came out.

Why keyd rather than Hyprland per-device binds: Hyprland's `device` bind flag targets a device *name*, and behind the Bolt receiver every paired device shares the name `logitech-usb-receiver` — the MX Keys S can't be singled out that way. keyd scopes by evdev id instead: behind Bolt that's the receiver's id (`k:` limits the grab to keyboard interfaces), behind Unifying it's the MX Keys' own id, so a co-paired mouse or other keyboard is untouched either way. `F13`–`F24` are untouched by Omarchy, so **no existing binding changes**.

## Changes

| File | Purpose |
| --- | --- |
| `bin/omarchy-hw-logitech-mx-keys` | detector for all three keyboards: Bluetooth/Unifying model-name match in sysfs (with a wireless-PID fallback for an MX Keys asleep at name-resolution time), else a HID++ 2.0 device-name probe of each Bolt receiver slot for the S. Exit 0 = present. `--keyd-ids` prints the keyd `[ids]` entry for each transport the keyboard is connected on. |
| `install/hardware/logitech-mx-keys.sh` | when detected: `omarchy-pkg-add keyd`, write `/etc/keyd/logitech-mx-keys.conf` with an `[ids]` section built from the detector's output (a Bluetooth-only keyboard never pulls in an unrelated Bolt receiver; a Unifying MX Keys gets its own id, never the receiver's), then `systemctl enable` + `restart keyd` (restart, not `enable --now` — keyd may already be running for another keyboard). Wired into `install/hardware/all.sh`. The probe self-elevates when the script runs as the user. Re-runnable by hand if the keyboard was asleep. |
| `migrations/1787838885.sh` | same for existing installs, covering all three keyboards. An unprivileged sysfs pre-filter (Bolt receiver id, MX Keys wireless PID, or an anchored model-name match that excludes MX Mechanical/for-Mac) means machines with none of these keyboards never reach the sudo-prompting probe; it then sources the hardware script. Marks itself done only once `keyd` is enabled. |
| `default/keyd/logitech-mx-keys.conf` | template: comments + modifier layers → F-keys, covering all three keyboards' problem chords in one file. `[ids]` is prepended at install time. Not shipped to `/etc` unconditionally. |
| `default/hypr/bindings/logitech-mx-keys.lua` | `o.bind("code:192", …)` etc. — bound by keycode (`code:191..197` == F13..F19) because the F13–F24 keysyms are absent from many non-US keymaps. Registered in `default/hypr/omarchy.lua`; gated on the installed keyd config so the binds exist only where the remap does. |
| `config/hypr/bindings.lua` | drop the "Logitech MX Keys examples" comment block — it was the manual workaround these keys no longer need. |

Mapping: `F13` mute mic (native, S and Mini) · `F14` emoji (S and Mini) · `F15` screenshot (S and Mini) · `F16` dictation (S and Mini, guarded on `voxtype`) · `F17` lock (all three) · `F18` float/tiling toggle (MX Keys) · `F19` scratchpad toggle (MX Keys). The keyd template and Hyprland bind file cover all three models at once — harmless on a machine with only one of them, since the other models' chords never arrive there to trigger the extra rewrites.

## Gating / blast radius

- Nothing is installed or enabled unless `omarchy-hw-logitech-mx-keys` matches.
- keyd then grabs only the detected id(s) and its modifier layers rewrite **only the exact chords listed**; every other key and combination passes through byte-for-byte (a keyd binding is unaffected by its layer's modifiers — keyd(1)).
- Behind the Bolt receiver the id is the receiver's, shared with a co-paired device (e.g. an MX Master); `k:` limits the grab to keyboard interfaces so the mouse is unaffected. Behind a Unifying receiver each paired device gets its own product id, so the MX Keys is grabbed on its own — a co-paired MX Master or K800 is untouched without `k:` even needing to do anything. Over Bluetooth — the Mini's only verified transport — each keyboard likewise gets its own id, so a co-paired Bluetooth mouse is untouched the same way.
- All three keyboards' rewrites live in one shared keyd config and one shared bind file. That's safe rather than a widened blast radius: a chord that belongs to another model is simply never sent by a machine that only has one of them, so the extra rules just never fire there.

## Known limitations

- The Bolt receiver probe needs the MX Keys S **awake**; a unit asleep during install/migration is missed until `bash "$OMARCHY_PATH/install/hardware/logitech-mx-keys.sh"` is re-run. Bluetooth and Unifying are matched from sysfs and don't need this — the Unifying case even has a wireless-PID fallback for a keyboard that was asleep when the kernel first resolved its name.
- HID++ probing (Bolt receiver, MX Keys S only) was validated against a receiver bound to `hid-generic`. The `hid-logitech-dj` driver path is untested.
- Concurrent HID++ tools (Solaar, libratbag) sharing the hidraw node during the Bolt probe can cause a miss.
- The MX Keys Mini has only been verified over Bluetooth. Its detection over a Bolt or Unifying receiver, if the hardware supports one, is untested — the Bolt probe in particular only scans the receiver id (`046d:c548`) documented for the S, so a Mini paired to a different Bolt product id wouldn't be probed at all.
- The name match (`mx keys` optionally followed by ` s`, excluding "mechanical"/"mac") is a best-effort string match against whatever the kernel or the keyboard's own HID++ response report as the model name; an unusual firmware string could in principle slip past either filter in either direction. Unlike the MX Keys and MX Keys S, the Mini's inclusion wasn't verified on the author's own hardware — it's based on a community hardware report (see PR discussion).

## Tests

`test/shell.d/hyprland-binding-conflicts-test.sh` passes. The new binds are behind the installed-config gate, so the conflict check doesn't see them unless `/etc/keyd/logitech-mx-keys.conf` is present — and they use the otherwise-unused `code:191`–`code:197` anyway.

Verified end to end on hardware: MX Keys S on a Bolt receiver, `de` keymap, all five keys reaching their Omarchy actions; MX Keys on a Unifying receiver, `us` keymap, lock/float-toggle/scratchpad reaching theirs. MX Keys Mini support is based on independent hardware verification from @tbjornli over Bluetooth on a Norwegian keymap (Omarchy 4.0.1, Hyprland 0.56.2): all five action-row keys reached their Omarchy actions, including the Fn+Del lock key sending the same `SUPER+L` chord as the S's dedicated lock key, and Solaar confirming the keyboard was already in Linux mode — ruling out an OS-switch misconfiguration as the underlying cause.


tbjornli:
I tested this PR's keyd rules against an **MX Keys Mini** (`046d:b369`, Bluetooth), which `NAME_EXCLUDE` currently rejects. Short version: every action key the Mini has sends exactly the same chord as the MX Keys S, so the rules already in this PR cover it unchanged.

Setup was your `default/keyd/logitech-mx-keys.conf` with `[ids]` set to `k:046d:b369` (the format the detector emits for non-Bolt transports), plus the `code:191..195` binds. Omarchy 4.0.1, Hyprland 0.56.2, Norwegian keymap.

Scoping was exactly right — keyd grabbed the Mini and nothing else:

```
DEVICE: match    046d:b369  (Logitech MX Keys Mini)
DEVICE: ignoring 046d:b034  (Logitech MX Master 3S)
DEVICE: ignoring 0001:0001  (AT Translated Set 2 keyboard)
```

Chords measured with `evtest`, rewrites confirmed on keyd's virtual keyboard:

| Key | Chord sent | Your rule | Result |
| --- | --- | --- | --- |
| Dictation (F6) | `SUPER+H` | `[meta] h = f16` | `F16` |
| Emoji (F7) | `SUPER+CTRL+ALT+SHIFT+Space` | `[meta+control+alt+shift] space = f14` | `F14` |
| Screenshot (F8) | `SUPER+SHIFT+S` | `[meta+shift] s = f15` | `F15` |
| Mute mic (F9) | bare `F13` | none needed | `F13` |
| Lock (Fn+Del) | `SUPER+L` | `[meta] l = f17` | `F17`, screen locks |

That's the whole MX Keys S action set, chord for chord, including the bare `F13` on mic mute. The Mini's lock key is `Fn+Del` (lock icon on the Delete key) rather than a dedicated key above the numpad, but it sends the same `SUPER+L`. The only rules with no Mini equivalent are `[meta] a`/`d`, which are the non-S MX Keys Fn-layer chords — harmless here since `SUPER+A`/`SUPER+D` are unbound in 4.0.1.

Dropping `"mini"` from `NAME_EXCLUDE` appears to be the whole change — `NAME_RE` already matches `MX Keys Mini`. Running your detector unmodified on this machine exits 1; with that one word removed it exits 0 and prints exactly the id I tested with:

```
$ omarchy-hw-logitech-mx-keys --keyd-ids     # NAME_EXCLUDE = ("mechanical", "mac")
k:046d:b369
```

That's the Bluetooth/sysfs path only — the Mini is Bluetooth-only here, so I haven't exercised the Bolt or Unifying branches.

Two smaller notes:

- Your reason for binding `code:191..197` rather than keysyms holds up nicely here: on this Norwegian keymap Hyprland lists those binds as `XF86TOOLS`/`XF86LAUNCH5..8`, so an `F14` bind would indeed never have matched.
- Solaar reports the Mini already in Linux mode (`MULTIPLATFORM {4531}` → `Set OS: Linux`) and it sends the Windows chords regardless, so this isn't an OS-switch misconfiguration users can fix themselves.

Unrelated to the Mini, one thing worth a line either way: this PR removes the `-- Logitech MX Keys examples:` block from `config/hypr/bindings.lua`, which is currently the only in-tree pointer for owners of an excluded model. Since `SUPER+SHIFT+S` hits the Google Maps webapp binding added in 4.0.0, those users end up with neither the fix nor the hint.

axelfontaine:
@tbjornli Thanks for testing. PR updated with the required changes for the Mini. Please validate so we are 100% sure.

Examples: This is a conscious choice. Let's update the code to fix the root cause for these keyboards instead.

axelfontaine:
@nixfred Could you also comment with the changes needed for the mac variant (just like @tbjornli did for the mini), so that can be integrated as well?

PR #8611, opened 2026-08-27
Link Omarchy agent skills into Hermes

smfworks:
## Why

Omarchy already symlinks `omarchy` and `diagnose-crash` into Claude, Codex, Pi, Antigravity, and `~/.agents/skills`. Hermes was missing, so Hermes agents on Omarchy never picked up the shipped skill for editing `~/.config/hypr`, `shell.json`, or themes.

## What

- `omarchy-provision-user` also links into `~/.hermes/skills` and every existing `~/.hermes/profiles/*/skills` (does not create Hermes profiles)
- Migration `1787843905.sh` for current installs
- `test/shell.d/provision-user-test.sh` covers default Hermes home plus a pre-existing profile
- Manual AI chapter and `docs/file-layout.md` updated

## Test plan

- [x] `bash test/shell.d/provision-user-test.sh`
- [x] `HOME=$(mktemp -d) bash -euo pipefail migrations/1787843905.sh` creates the default-home links

Rollback: remove the migration marker is not needed for a revert; reverting the provision-user loop restores previous harness list. Existing Hermes symlinks are harmless if left in place.

### DIFF
--- a/bin/omarchy-provision-user
+++ b/bin/omarchy-provision-user
@@ -84,7 +84,7 @@ fi
 # Dev-aware skill symlinks. Cannot live in /etc/skel because OMARCHY_PATH may
 # point at a dev checkout (omarchy dev link) where the target differs.
 # Loops every skill directory, so shipping a new one needs no edit here.
-mkdir -p ~/.agents/skills ~/.claude/skills ~/.codex/skills ~/.pi/agent/skills ~/.gemini/config/skills
+mkdir -p ~/.agents/skills ~/.claude/skills ~/.codex/skills ~/.pi/agent/skills ~/.gemini/config/skills ~/.hermes/skills
 for skill in "$OMARCHY_PATH"/default/agents/skills/*/; do
   skill=${skill%/}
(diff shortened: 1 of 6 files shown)


smfworks:
Still distinct from #7392.

#7392 links Omarchy skills into `~/.hermes/skills` only. This PR also links existing `~/.hermes/profiles/*/skills` (does not create profiles), with a migration and tests.

If #7392 takes the profile loop, this can close as well.

smfworks:
Addressed the Copilot nits:

- `test/shell.d/hermes-skills-migration-test.sh` runs the one-shot migration against a fake `HOME` (default home, pre-existing `james` profile, idempotency, missing skill source). Removing the profile loop or the default-home `ln -sfn` fails it.
- `docs/file-layout.md` now lists Antigravity (`~/.gemini/config/skills`) and that bullet is a single line.

Rebased onto current `quattro`. Keeping this PR open per omarchybot on #7469; #7392 has the same loop if it lands first.

PR #8633, opened 2026-08-27
Stop network speed tests when their launcher dies

### LINKED ISSUE #8515: omarchy-network-speedtest: orphaned processes run indefinitely when Quickshell restarts
### System details

Intel Core (4 cores), Omarchy 4.0.1-1, kernel 7.1.9-arch1-2

## What's wrong

When Quickshell restarts or crashes while the speedtest panel is open, the `omarchy-network-speedtest` child processes become orphaned (re-parented to systemd) and **run forever**. The `cleanup()` trap in the script only fires on `EXIT`, which never happens if the parent process dies without sending `SIGTERM`.

Each restart leaves another set of instances behind. I found **7 orphaned instances** all in `up` mode, each with 8 parallel `curl` workers saturating the link — totalling ~56 concurrent upload streams to Netflix OCA servers.

## Observable symptoms

- Both fans at >7300 RPM (above the reported 7000 RPM max, PWM at 128%)
- Wi-Fi card at 76°C
- `ps aux` shows multiple `omarchy-network-speedtest up` processes with `systemd --user` as parent (PID 1)

```
sebarb   1159822  ... /bin/bash /usr/share/omarchy/bin/omarchy-network-speedtest up
sebarb   1159823  ... /bin/bash /usr/share/omarchy/bin/omarchy-network-speedtest up
sebarb   1159824  ... /bin/bash /usr/share/omarchy/bin/omarchy-network-speedtest up
... (7 total)
```

## Root cause

`Process {}` elements in QML do not guarantee `SIGTERM` delivery to children when Quickshell itself is killed with `SIGKILL` or crashes. The bash script's `trap cleanup EXIT` is therefore never triggered, and the `while true` loop inside `traffic_worker()` runs indefinitely.

## Suggested fix

Add a parent-death guard at the top of `omarchy-network-speedtest`, before spawning workers:

```bash
# Die when our parent dies
( while kill -0 $PPID 2>/dev/null; do sleep 1; done; kill -- -$$ 2>/dev/null ) &
```

This polls the parent PID and sends `SIGTERM` to the script's process group when the parent is gone, which triggers `cleanup()` and terminates the curl workers cleanly.

Alternatively, use a lock file or a named pipe whose read end is held by the QML `Process` — closing the process would close the pipe, and the script could detect EOF and exit.

## Steps to reproduce

1. Open the speedtest panel
2. Kill or restart Quickshell (`killall -9 qs`) while the test is running
3. Check `ps aux | grep omarchy-network-speedtest` — processes remain running indefinitely

lamchun1110:
## Why

When Quickshell crashes or is killed, it cannot signal the `omarchy-network-speedtest` processes it launched. Their traffic loops then keep running indefinitely, and each shell restart can leave another eight download or upload streams behind.

## What

- watch the launcher PID while a speed test is active
- terminate the speed-test supervisor when that launcher disappears, allowing the existing `EXIT` cleanup to stop every traffic worker
- add a hermetic regression test that kills a simulated Quickshell parent with `SIGKILL` and verifies the supervisor and all eight traffic processes exit
- make the network statistics root overridable so the process-level test does not depend on the host interface layout

## Testing

- `bash test/shell.d/network-speedtest-parent-test.sh` (three consecutive runs)
- `bash test/shell.d/network-test.sh`
- `bash -n bin/omarchy-network-speedtest test/shell.d/network-speedtest-parent-test.sh`
- `git diff --check`

Fixes #8515

### DIFF
--- a/bin/omarchy-network-speedtest
+++ b/bin/omarchy-network-speedtest
@@ -10,6 +10,7 @@ direction="${1:-}"
 probe=1.1.1.1
 parallel=8
 url_count=3
+network_devices_path="${OMARCHY_NETWORK_DEVICES_PATH:-/sys/class/net}"
 
 case "$direction" in
   down | up)
@@ -30,7 +31,10 @@ format_mbps() {
 
 iface=$(ip route get "$probe" 2>/dev/null | awk '{ for (i = 1; i <= NF; i++) if ($i == "dev") { print $(i + 1); exit } }')
 
-if [[ -z $iface || ! -r /sys/class/net/$iface/statistics/rx_bytes || ! -r /sys/class/net/$iface/statistics/tx_bytes ]]; then
+rx_path="$network_devices_path/$iface/statistics/rx_bytes"
+tx_path="$network_devices_path/$iface/statistics/tx_bytes"
+
+if [[ -z $iface || ! -r $rx_path || ! -r $tx_path ]]; then
   echo "No active network interface" >&2
   exit 1
 fi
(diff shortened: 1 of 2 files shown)


sheikhlimon:
Thanks for tracking this down! It looks like there are multiple PRs open for the orphaned speed test workers. I recommended consolidating the fix over in #11054 since it leverages native process groups (`set -m`) to close a race condition in the cleanup trap. It might be best to close this as a duplicate once that approach lands.

PR #8634, opened 2026-08-27
Open the apps menu on a right click on the bare desktop

kostea92:
Right-clicking bare desktop opens the apps menu — the same one `Super + Alt + Space` reaches. A right click there does nothing today until it becomes a double click, and an empty desktop is exactly where the reflex to reach for a launcher lands.

Nothing that worked before changes. The second click of a right double-click still belongs to the theme switcher, so the single click waits out `Qt.styleHints.mouseDoubleClickInterval` before opening — a menu opened on the first of the two would have swallowed the second. Left double-click still opens the background switcher.

Verified on a running session, and covered in `test/shell.d/background-test.sh`.

### DIFF
--- a/manual/03-coming-from-mac-or-windows.md
+++ b/manual/03-coming-from-mac-or-windows.md
@@ -10,7 +10,7 @@ Your Spotlight or Raycast or Start-menu reflex becomes `Super + Space`. That ope
 
 ### There's no dock and no desktop icons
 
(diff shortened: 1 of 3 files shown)


kostea92:
Both addressed.

The assertion now matches the wiring rather than four loose substrings: the timer's `onTriggered: root.openAppsMenu()`, the `restart()` inside the right-button branch of `onClicked`, and the `stop()` at the top of `onDoubleClicked`. Checked by mutation — breaking any one of the three turns it red, which the previous version did not.

Article added, in the manual and in the code comment that says the same thing. The title follows.

PR #8635, opened 2026-08-27
Answer arithmetic and unit conversions in the menu search

kostea92:
The menu search answers arithmetic and unit conversions, so `Super + Space` covers the "quick sum" and "how many miles is that" reflexes without leaving the launcher.

- Arithmetic: `1024/16`, `(12+8)*3`, `2^10`, `sqrt(2)`, `max(3, 9, 2)`, `1e6/4`, `round(3.14159, 2)`
- Percentages: `178000*20%` and `20% of 178000` take the share, while `178000+20%` adds it to what it is added to, the way a till does
- Conversions: `20 km to mi`, `100 f to c`, `1 gb to mib`, `2 lbs to kg` — length, mass, time, data, volume and temperature
- A quantity on its own answers with the units its readers usually want, so `5km` offers both miles and metres

Enter copies the result.

`%` is a percentage rather than a remainder. The two cannot share the sign — one wants an operand on its right and the other refuses one — and a search box is asked for "20% off" far more often than for a modulo.

The evaluation is a tokenizer and a recursive-descent parser in `shell/plugins/menu/Calculator.js`, not `eval`: the search box sees whatever is halfway typed, and `eval` would hand that the whole QML scope. The module is pure JS with no QML types, so `test/shell.d/menu-calculator-test.sh` exercises it directly under Node.

Search stays search. A row appears only when something was actually computed, so a bare number or a bare constant produces nothing, and a single-letter unit written against its number is not read as a unit: `4k`, `3d`, `5m` and `1080p` stay queries, while `20 c` with its space is a temperature. `ton` is deliberately missing — 1000 kg in most of the world and 907 kg in the US, and an alias silently wrong by a tenth is worse than one that is absent.

### DIFF
--- a/manual/04-navigation.md
+++ b/manual/04-navigation.md
@@ -2,6 +2,8 @@
 
 Everything in Omarchy happens via the keyboard — _EVERYTHING!_ When the system first starts, you literally can't do a thing with the mouse alone. But you can hit `Super + Space` to reveal the Omarchy Menu and from here you to do just about everything.
 
+The menu search doubles as a calculator. Type an expression and the answer is the first row: `1024/16`, `(12+8)*3`, `sqrt(2)`, `1e6/4`. Percentages read the way a till does — `178000*20%` and `20% of 178000` both take the share, while `178000+20%` adds it on. Type a quantity and it converts: `20 km to mi`, `100 f to c`, `1 gb to mib`. A quantity on its own answers with the units its readers usually want, so `5km` offers both miles and metres. Enter copies the result. A single-letter unit needs a space to be read as one (`20 c`, not `20c`), so searches like `4k` and `3d` stay searches.
+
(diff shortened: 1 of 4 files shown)


kostea92:
All four were real. Fixed:

- A percentage on the left of `+` / `-` is now its own fraction, so `20% + 100` is 100.2 and `20% + 20%` is 0.24. The till-style reading stays on the right-hand operand, where it belongs.
- `^` normalizes both operands: `2^50%` is 1.41421356237, `20%^2` is 0.04.
- Parentheses no longer count as an operation, so `(5)` and `(pi)` stay searches. Whatever is inside raises the count itself.
- Functions carry `maxArity` as well, and a call outside the range is refused rather than trimmed: `sqrt(4, 9)` and `round(1.234, 2, 3)` now produce nothing.

The first three shared a root cause — the fraction conversion was open-coded at each site and simply missing at two of them — so they now go through one `fraction()` helper used by every operator, function argument, group and final result.

Each fix has a test; the file is at 63 assertions.

PR #8640, opened 2026-08-27
WhatsApp Slim: avatar-name tooltip and rail divider

nas3ts:
## Summary

Two fixes for the WhatsApp Slim extension's collapsed avatar rail (window <= 1100px):

- Show the contact name in a tooltip when hovering an avatar.
- Restore the rail's divider border.

## What changed

- `avatar-tooltip.js` (new): hover tooltip with the chat name, gated to rail mode.
- `whatsapp.css`: adds the rail divider border.
- `manifest.json`: 1.1 -> 1.2, registers the new script (key unchanged, so the
  extension id stays stable).

## Verification

- `test/shell.d/chromium-whatsapp-slim-test.sh` passes.
- Screenshots below: rail divider, and name tooltip on avatar hover.

## Screenshots
screenshot-1
screenshot-2

### DIFF
--- a/default/chromium/extensions/whatsapp-slim/avatar-tooltip.js
+++ b/default/chromium/extensions/whatsapp-slim/avatar-tooltip.js
@@ -30,13 +30,23 @@
   document.documentElement.appendChild(tip);
 
   function getName(cell) {
+    // The avatar image's title is the contact/group name. An unread row can
+    // instead surface an indicator image (muted/pinned/etc.), so only trust the
+    // title when it reads like a name rather than a count or status word.
     const img = cell.querySelector("img[title]");
-    const t = img && img.getAttribute("title");
-    if (t && t.trim()) return t.trim();
+    if (img) {
+      const t = img.getAttribute("title").trim();
+      if (t && !/^\d+$/.test(t) && !/unread|message|muted|pinned|archived/i.test(t)) {
+        return t;
+      }
+    }
+    // Fallback: first text span that isn't the unread badge or a bare count,
+    // either of which would otherwise masquerade as the name in the tooltip.
     const spans = cell.querySelectorAll("span");
     for (const s of spans) {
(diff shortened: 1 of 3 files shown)


nas3ts:
@copilot-pull-request-reviewer All four review findings are now resolved; the latest round is at head sha `e66e792f`.

1. **Stale cached name** — fixed in `7c1b667f`: `show()` reads the current DOM via `getName(cell)` on every hover; `dataset.waName` is kept only as the listener-marker (set to `""`).
2. **Observer re-querying every chat row** — fixed in `e66e792f`: the `MutationObserver` now visits only `mutation.addedNodes`, tagging a chat row directly or descending into an added subtree; the full scan runs once at load only.
3. **Tooltip clamped top-only** — fixed in `e66e792f`: the tooltip is shown first, then measured (`offsetWidth`/`offsetHeight`) and clamped to the viewport on both axes (`left` vs `window.innerWidth - w - 8`, `top` vs `window.innerHeight - h - 8`).
4. **No focused automated coverage** — fixed in `e66e792f`: adds `test/shell.d/whatsapp-slim-avatar-tooltip-test.sh`, a DOM/matchMedia harness that asserts the `document_idle` manifest entry and covers rail gating, name extraction (unread badge ignored), dynamic-row tagging, stale-cache re-reads, and viewport clamping.

All 8 harness assertions pass:

```
ok - avatar tooltip stays hidden outside rail mode
ok - avatar tooltip shows the contact name on hover in rail mode
ok - avatar tooltip shows the contact name from the avatar title
ok - avatar tooltip ignores unread badge and uses the contact name
ok - avatar tooltip tags dynamically added chat rows
ok - avatar tooltip re-reads the current name on hover
ok - avatar tooltip picks up a reused row's updated name
ok - avatar tooltip clamps to the viewport on both axes
```

PR #8280, opened 2026-08-25
Add a Grok collector to the agents usage panel

eloklam:
Posted by **Grok on behalf of Enoch Lam**.

Discussion: https://github.com/basecamp/omarchy/discussions/8279

## What

Grok (official `grok` CLI / SuperGrok) was missing from the agents usage panel. This adds `omarchy-agent-usage-grok` so `omarchy-agent-usage-update` picks it up like Claude, Codex, and Fireworks.

## Behavior

- **Title** is `subscription_tier_display` from `/v1/settings` (live, not hardcoded).
- **Subtitle** is the email from `~/.grok/auth.json` when present.
- **Limit rows** keep API product strings (`GrokBuild`, `GrokChat`). The weekly window label is derived from `currentPeriod.type` (`USAGE_PERIOD_TYPE_WEEKLY` → `Weekly`). Pay-as-you-go with cap 0 is `Off` (no fake 0% bar).
- **Tokens by day / by model** use `turn_completed.usage` in `~/.grok/sessions/*/updates.jsonl`. `recentDays.messageCount` is a token total, same as Claude/Codex. Cached-read is already inside Grok `inputTokens`, so it is not added twice.
- Marks: `assets/grok.svg` (white) and `assets/grok-light.svg` (black), same pair convention as Codex.

## Tests

`./test/shell.d/agent-usage-grok-scanner-test.sh` (and existing agents panel / usage-update tests still pass).

## Notes

`GROK_HOME` overrides `~/.grok`. The panel changes are additive (`accountEmail`, limit `status`); other collectors ignore the new fields.

PR #8649, opened 2026-08-27
Add Grok and OMP collectors to the agents usage panel

rmacy:
## Summary

Setup already offers Grok and OMP as first-party coding agents, but `omarchy.agents` never lights up for them. Collectors are discovered from `$OMARCHY_PATH/bin/omarchy-agent-usage-*`; there was no Grok or OMP one.

This adds `omarchy-agent-usage-grok` and `omarchy-agent-usage-omp`. Both print the record contract the panel already understands.

There are several open Grok-only collector PRs (#7200, #6485, #6902, #7393, #8280, #6477). None of them are merged, and none add OMP. This patch covers both.

## Change

- **Grok local stats** come from `$GROK_HOME/sessions` (default `~/.grok/sessions`). Billed usage is on `updates.jsonl` `turn_completed` rows (nested `usage`, or flat token fields / `event_name`). Cache read is split out of input. `session_kind` values that start with `subagent` are skipped so parent totals are not counted twice.
- **Grok limits** come from ACP `_x.ai/billing` (`grok agent --no-leader stdio`), the same source `/usage` uses. A SuperGrok weekly (or monthly) pool becomes `limits[]`. Leftover prepaid / on-demand credits become a `balance` only when the ledger is actually non-zero. SuperGrok Heavy often omits `creditUsagePercent`; the collector does not invent a 0% meter.
- **OMP** is a local cross-provider runtime view: `~/.omp/stats.db` (or `$OMP_HOME/stats.db`), grouped by day and model. No subscription limits.
- `assets/grok.svg` (and a `-light` twin) use the official Grok mark. The bar glyph still stands in without them.
- Provider switch pills size to their labels and wrap, so "Claude Code" does not overflow once more tabs join the strip.

## Tests

`test/shell.d/agent-usage-grok-scanner-test.sh` — nested and flat turn ledgers, cache split, prompt_id dedupe, subagent skip, empty install, weekly/monthly mapping, zero prepaid omit, leftover credits, missing percent.

`test/shell.d/agent-usage-omp-scanner-test.sh` — empty install, message/session/token totals, per-model buckets, local scope.

`test/shell.d/agents-panel-test.sh` — provider switch no longer forces equal-width cells.

Verified live on Grok 1.0.5 / SuperGrok Heavy and a populated OMP `stats.db`.


bjarneo:
Automated duplication check: this pull request looks similar to #9498, which covers the same Grok usage collector. I keep that one open and close this one to consolidate review.

PR #8655, opened 2026-08-27
[codex] Constrain package-conflict file recovery

AFOliveira:
# Update conflict rollback follows a user-controlled symlink and writes as root

## Summary

`omarchy-update-system-pkgs` captures privileged pacman errors in a temporary file owned by the unprivileged desktop user. If pacman fails, `omarchy-update-system-pkgs-when-conflicted` trusts path strings parsed from that still-user-writable file and performs privileged moves on them.

The failure rollback contains a direct destination-path TOCTOU. After moving a supplied user-owned file into `/var/lib/omarchy/replaced`, the handler runs pacman again. If that retry fails, `restore_moved()` reopens the original user pathname with `sudo mv`. While pacman is retrying, the user still controls every parent beneath their home directory and can replace one with a symlink to `/`. The privileged “restore” then follows that link and moves the attacker-controlled payload to the corresponding absolute host path.

A process already running as the logged-in user can exploit a normal authorized update as follows:

1. Cause or wait for a pacman failure and replace the same-UID-writable error report before the handler parses it. The forged line names an executable below a user-owned path such as `~/stage/escape/usr/lib/systemd/system-sleep/audit-payload` and attributes it to `omarchy`.
2. The handler verifies only that the source is not currently package-owned, then moves it into the root-owned quarantine.
3. During the pacman retry, replace `~/stage/escape` with a symlink to `/`.
4. Keep the original failure condition in place so the retry fails. The EXIT trap decides the supplied pathname is absent and runs `sudo mv -T "$replaced$path" "$path"`.
5. The destination now resolves to `/usr/lib/systemd/system-sleep/audit-payload` and receives the attacker's executable.

`systemd-suspend.service` runs every executable in that directory as root immediately before and after the next suspend. The handler's own comments already identify `systemd-sleep` as a directory where every executable remains live.

There is a second pivot in the forward move loop: the quarantine mirrors absolute pathnames by string concatenation, and a moved directory retains the user's ownership. A symlink inserted into that directory can make a later `sudo mkdir` / `sudo mv` destination escape `/var/lib/omarchy/replaced`. The rollback primitive is simpler and has the entire failed retry as its synchronization window, so it is the primary finding.

## Severity

Medium to High: local privilege escalation through an authorized system update.

Suggested conservative CVSS 3.1: `CVSS:3.1/AV:L/AC:H/PR:L/UI:R/S:U/C:H/I:H/A:H` (6.7). The owner must run and authorize an Omarchy update, pacman must enter its failure path, and the attacker must synchronize with the mutable report and rollback. Once the write lands, the next ordinary suspend executes the payload as root. A network failure naturally makes both pacman attempts fail and leaves a comparatively wide window for the parent-link swap.

## Affected code and versions

Confirmed on `quattro` commit `93010924`.

Introduced by commit `646587e3` (`Recover from package file conflicts instead of predicting them`, 2026-07-27). `v4.0.0-beta3` and `v4.0.0` contain the vulnerable handler; `v3.8.4` does not.

Relevant flow:

1. `bin/omarchy-update-system-pkgs`, lines 18-39:
   - creates `errors=$(mktemp)` as the desktop user;
   - redirects root pacman's stderr into that user-owned file;
   - after any first-attempt failure, execs the handler with the pathname.
2. `bin/omarchy-update-system-pkgs-when-conflicted`, lines 21-26, treats a forgeable environment value as its only provenance check.
3. Line 32 accepts caller-controlled `OMARCHY_REPLACED_DIR`, otherwise using `/var/lib/omarchy/replaced`.
4. Lines 84-99 parse arbitrary path text attributed to an Omarchy package. `pacman -Qo` only establishes that the current source is not package-owned; it does not prove that pacman emitted the line, that the named package will install the path, or that the path is below a system directory.
5. Lines 101-106 move the attacker-named source into quarantine.
6. The handler retries pacman at lines 111-115. On failure, its EXIT trap invokes `restore_moved()`.
7. Lines 42-57 check and then reuse the original mutable pathname. Line 54 performs the root move without constraining or pinning any destination parent component.

The update lock does not protect the report or home pathname from another process with the same UID, and mode `0600` does not separate processes belonging to that UID. `fs.protected_symlinks=1` applies to sticky-directory link following; it does not stop root traversing a link inside the attacker's own home tree.

The `OMARCHY_UPDATE_CONFLICT=1` environment check does not authenticate the call, and `OMARCHY_REPLACED_DIR` is also caller-controlled. Those facts make direct or prearranged invocation easier, but they are not required for the normal-update rollback race.

## Safe deterministic reproduction

I ran the real handler with `sudo` and `pacman` replaced by unprivileged test shims, and redirected both quarantine and escaped destination into `/tmp`. The `sudo` shim only pauses after the first real `mv` until a companion process completes the otherwise timing-dependent parent swap; it does not alter path resolution.

The initial report contained one line:

```text
omarchy: /tmp/omarchy-update-restore-poc/home/stage/escape/usr/lib/systemd/system-sleep/audit-payload exists in filesystem
```

The source was an executable regular file under the test user's directory. Once the handler moved it into the isolated quarantine, the companion process performed only unprivileged operations:

```bash
mv /tmp/omarchy-update-restore-poc/home/stage/escape \
   /tmp/omarchy-update-restore-poc/home/stage/escape-original

ln -s /tmp/omarchy-update-restore-poc/fake-root \
   /tmp/omarchy-update-restore-poc/home/stage/escape
```

The fake pacman deliberately failed the retry. The real handler printed:

```text
Taking over files pacman doesn't own yet:
  /tmp/omarchy-update-restore-poc/home/stage/escape/usr/lib/systemd/system-sleep/audit-payload
    -> /tmp/omarchy-update-restore-poc/replaced/tmp/omarchy-update-restore-poc/home/stage/escape/usr/lib/systemd/system-sleep/audit-payload

Update system packages
error: simulated retry failure

Putting back what the upgrade didn't take:
  /tmp/omarchy-update-restore-poc/home/stage/escape/usr/lib/systemd/system-sleep/audit-payload
```

The “put back” path had changed underneath it. Observed result:

```text
handler status (retry deliberately fails): 1
home/stage/escape -> /tmp/omarchy-update-restore-poc/fake-root
restored through link: -rwxr-xr-x 755 afonso-server:afonso-server /tmp/omarchy-update-restore-poc/fake-root/usr/lib/systemd/system-sleep/audit-payload
```

Invoking the isolated installed hook with systemd's arguments produced:

```text
restore payload ran with uid=1000 args=pre suspend
```

The UID is intentionally unprivileged in this safe harness. With the link target changed from the fake root to `/`, the exact same line 54 root move resolves the destination to `/usr/lib/systemd/system-sleep/audit-payload`; the system sleep service then invokes it as UID 0.

The complete isolated harness is retained at `/tmp/omarchy-update-restore-poc/` on the audit machine. It neither invoked real sudo nor changed `/usr`.

The local `systemd-sleep(8)` manual and official documentation confirm that all executables in this path run around sleep:

- <https://www.freedesktop.org/software/systemd/man/latest/systemd-suspend.service.html>

## Practical synchronization

The test shim makes the scheduling deterministic but does not create the vulnerable lookup. A real same-user process can make the chain practical:

- monitor `/tmp` for the new user-owned `mktemp` report, select files owned by the victim UID, and keep the candidate open while pacman runs;
- wait for a real package-file conflict or another pacman failure, then replace the report before the handler parses it;
- preserve the original failure condition (for example, an unavailable mirror/network) so the retry also fails, giving the attacker the duration of the second pacman attempt to replace the home-directory parent;
- append enough data while the first script is displaying the error report to widen the report-replacement window;
- pre-set the unsanitized `OMARCHY_REPLACED_DIR` test hook to a user-controlled tree in the update environment, enabling a pre-existing destination symlink and avoiding the forward-move timing entirely.

The attacker does not need to read or alter another Unix user's file: the report and original source path belong to the same logged-in UID that the local attacker already controls. The privilege boundary is crossed only when the trusted Omarchy handler reopens that mutable pathname under `sudo`.

## Recommended remediation

1. Do not let an unprivileged process own or modify the transaction report that authorizes root file moves. Run pacman and conflict interpretation inside one narrow privileged helper, keeping the report on a root-owned descriptor or in root-only storage.
2. Never restore to a pathname whose parent components remain user-controlled. A pre-update file from a user-writable tree does not belong in a root conflict-recovery mechanism; reject it. For legitimate fixed system roots, pin trusted parent directory FDs before moving anything and restore relative to those descriptors.
3. Do not mirror untrusted absolute pathnames into a quarantine tree. Store each item under an opaque, root-generated leaf name in a fresh root-owned per-transaction directory. A moved user directory must never become an intermediate parent for a later privileged destination.
4. Verify conflicts against pacman's actual transaction/package file lists rather than accepting text with an `omarchy:` prefix. Require canonical absolute paths, reject `..`, options, unexpected packages, user-owned parent directories, and any path outside a minimal allowlist of root-owned package locations.
5. Make `OMARCHY_REPLACED_DIR` and conflict provenance test hooks unavailable in production, or accept them only when an explicit test mode is established before privilege is acquired.
6. Use directory FDs and constrained resolution (`openat2()` with `RESOLVE_BENEATH`, `RESOLVE_NO_SYMLINKS`, and `RESOLVE_NO_MAGICLINKS`, or an equivalent design) for every root-side traversal. Merely calling `realpath` and then using the original pathname later leaves the TOCTOU intact.
7. Pass `--` before every untrusted pathname operand. This is not sufficient for the symlink attack, but closes the independent option-parsing class.
8. Add two regression tests: replace an original parent with a symlink before rollback, and insert a symlink into a moved user-owned quarantine directory before a later forward move. Neither root operation may escape its pinned root.

## Implemented fix

This PR replaces the caller-controlled recovery flow with a fixed, packaged root helper that owns both the decisive Pacman attempt and its report:

- captures the decisive first-attempt stderr in root-only storage and never accepts a caller-supplied report;
- accepts only the Omarchy package set and normalized paths below `/etc` or `/usr`;
- requires a root-owned, non-writable, symlink-free parent chain;
- verifies that each path is present in the selected package's trusted cached archive and is currently unowned;
- rejects overlapping, mixed, unrelated, writable-parent, and symlink-parent candidates;
- stages entries under opaque names in a root-created same-filesystem directory;
- performs the package retry inside the privileged helper;
- revalidates destination parents before rollback and never restores through a changed or unsafe boundary;
- archives replaced files and a transaction manifest under a root-only directory after success.

Commit: `63b1a7cc991d2559429e67c811c1f7df78ecf952`

## Verification

- `bash test/shell.d/update-file-conflict-test.sh`
- `bash test/shell.d/update-package-conflict-test.sh`
- `bash test/shell.d/update-sequence-test.sh`
- `bash test/shell.d/update-pacman-guard-test.sh`
- `bash test/shell.d/update-lock-test.sh`
- clean three-way merge against current `quattro`

The regression suite deterministically covers both the rollback parent-symlink pivot and a forward-move symlink inserted into an earlier quarantined directory. All targeted checks pass.

## Follow-up hardening

Commit `602d4238` addresses the adversarial review of the first implementation:

- binds archive authorization to the exact pending `pacman -Sup` transaction, excluding inactive and already-current packages;
- writes and syncs each opaque item-to-path mapping before the destructive rename;
- forbids recursive cross-filesystem archive copies and retains a root-only same-filesystem stage when `/var` differs;
- adds a production EUID-0 namespace test that kills the helper after rename and exercises separate `/etc` and `/var` filesystems.

Additional verification:

- `bash test/shell.d/update-file-conflict-test.sh`
- `bash test/shell.d/update-file-conflict-boundary-test.sh`
- `./test/cli`
- clean targeted test run after merging current `quattro` commit `06e32d243d5c410e2421b9f3db5be3c3ae33d3e2`

## Final transaction-boundary fix

Commit cb6cb20c closes the remaining same-UID report-injection window identified in the final review:

- moves the first noninteractive Pacman transaction and its error capture entirely into the root helper;
- uses a dedicated status only to hand genuine package-vs-package conflicts back to the interactive handler;
- reads the configured Pacman CacheDir list and accepts archives only below root-owned, non-writable directory chains;
- adds regression coverage proving caller-supplied text cannot authorize quarantine and that the report-authorizing transaction runs inside the root helper.

Final verification:

- focused conflict, package-conflict, update-sequence, Pacman-guard, update-lock, and CLI suites;
- production EUID-0 namespace coverage for transaction binding, forced-crash journaling, and separate-filesystem archival;
- conflict-free synthetic merge with current quattro at 0b3f1b7e;
- full ./test/all on both the branch and synthetic merge: every PR-specific test passed; the same four unrelated environment failures reproduce on clean quattro (missing omarchy-pkgs checkout and the local About-animation fixture).

## Archive failure-path fix

Commit `1e25e8db` fixes the conditional-errexit failure path found in final review:

- explicitly propagates failures from every archive boundary, permission, transaction-creation, metadata, and publication operation inside `archive_remaining()`;
- revalidates the root-owned transaction directory after creation;
- fails closed on an unsafe archive symlink, leaving the same-filesystem quarantine recoverable without chmoding or writing through the link;
- adds a regression that plants that symlink and verifies its target mode and contents remain unchanged.

Verification repeated on the branch and a clean synthetic merge with current `quattro`: all focused conflict, boundary, rollback, package-conflict, sequence, Pacman-guard, lock, and CLI tests pass. The full suite has only the same four unrelated environment/fixture failures reproduced on clean `quattro`.


ryanrhughes:
Superseded by #9053. The replacement preserves @AFOliveira's original report and commits, then adds the final hardening and regression coverage from review. Closing this copy only to keep review and CI on the branch maintainers can update.

PR #8662, opened 2026-08-27
Sanitize legacy Windows VM usernames

### LINKED ISSUE #8619: windows-vm: 4.0.1 migration rejects usernames the old installer allowed, stranding existing VMs with a misleading "not configured"
### System details

AMD Ryzen AI Max+ 395 (Radeon 8060S), Omarchy 4.0.1, kernel 6.18.46-1-lts, `dockurr/windows`

### What's wrong?

4.0.1 moved the Windows VM compose from `~/.config/windows/docker-compose.yml` to root-owned `/var/lib/omarchy/windows/docker-compose.yml` and validates every field on the root side before writing. `USERNAME` is validated with:

```bash
valid_username() { [[ $1 =~ ^[A-Za-z0-9_-]{1,20}$ ]]; }   # bin/omarchy-windows-vm:89
```

The pre-4.0.1 installer applied **no** validation to that field — empty fell back to `docker`, anything else was written verbatim:

```bash
USERNAME=$(gum input --placeholder="Username (Press enter to use default: docker)" ...)
if [[ -z $USERNAME ]]; then
  USERNAME="docker"
fi
```

So a VM installed with an email-style Windows account (`user@example.com` — what the Microsoft account sign-in prompt asks for) has a legacy compose the new code cannot migrate. `write_compose` exits 2 at line 140, `migrate_legacy_compose` returns 1 at line 358, and the VM is unreachable — while the container image and `~/.windows/data.img` are completely intact.

**The failure is near-undiagnosable from the CLI:**

- `launch` prints `Windows VM not configured. Please run: omarchy-windows-vm install` (line 621) — pointing the user at reinstalling a VM that already exists and works.
- `status`, `stop`, and `remove` call `migrate_legacy_compose 2>/dev/null || true` (lines 600, 708, 724), so the one line that names the real cause (`invalid username: …`) is discarded outright. Those subcommands simply behave as though no VM was ever configured.

Following the suggested remedy makes it worse: `install` provisions a fresh VM, and it *succeeds* where migration failed, because the installer silently substitutes `docker` for an invalid username (lines 503-506). Migration rejecting what `install` quietly repairs is the actual defect.

Still present on master as of today.

### Steps to reproduce

1. On Omarchy < 4.0.1, `omarchy-windows-vm install` and enter `user@example.com` as the Windows username.
2. Upgrade to 4.0.1.
3. `omarchy-windows-vm status` → reports not configured, no explanation.
4. `omarchy-windows-vm launch` → `Windows VM not configured. Please run: omarchy-windows-vm install`, despite a valid legacy compose and an intact `~/.windows/data.img`.

### Suggested fix

Have `migrate_legacy_compose` sanitize rather than reject. `USERNAME` is inert after install — `dockurr/windows` reads it only to build the unattend file during Windows setup — so falling back to `docker` on migration is safe and matches what `install` already does for the same input.

Failing that: name the offending field and value in the failure message, and stop discarding stderr in the `status`/`stop`/`remove` paths so the cause is at least visible.

Elshayib:
**What**: Sanitize invalid legacy Windows VM usernames during compose migration so existing VMs stay reachable.

**Why**: Fixes #8619. 4.0.1's elevated compose writer rejects usernames the pre-4.0.1 installer allowed (email-style Microsoft accounts, etc.). `migrate_legacy_compose` then failed, and `launch`/`status` reported "not configured" while the VM disk was intact.

**How**: In `migrate_legacy_compose`, fall back to `docker` when the legacy `USERNAME` fails `valid_username` — the same substitution `install` already makes. The elevated writer stays strict. RDP credentials keep the original username so the already-provisioned Windows account still logs in. `USERNAME` is only consumed by dockurr/windows during first-run unattend.

Rejected: relaxing `valid_username` in `__priv_write_compose` (that check is the security boundary).

**Verified**:
- `bash -n bin/omarchy-windows-vm` (exit 0)
- `bash -n test/shell.d/windows-vm-compose-test.sh` (exit 0)
- `bash test/shell.d/windows-vm-compose-test.sh` through the new assertion: `ok - migration sanitizes invalid legacy usernames and keeps RDP credentials`

**NOT verified**:
- independent /review
- shellcheck / shfmt (not installed on this machine)
- full `test/shell` / `test/cli` / `test/all` (Windows host; the later symlink-to-`/` assertion in the same compose test hung)
- live Omarchy Windows VM migration
- Hyprland / pacman / Quickshell

### DIFF
--- a/bin/omarchy-windows-vm
+++ b/bin/omarchy-windows-vm
@@ -340,7 +340,7 @@ migrate_legacy_compose() {
   [[ -f $LEGACY_COMPOSE_FILE ]] || return 1
 
   echo "Migrating Windows VM configuration to $COMPOSE_FILE ..."
-  local ram cores disk username password tz storage shared
+  local ram cores disk username password tz storage shared rdp_username
   ram=$(read_compose_value RAM_SIZE "$LEGACY_COMPOSE_FILE")
   cores=$(read_compose_value CPU_CORES "$LEGACY_COMPOSE_FILE")
   disk=$(read_compose_value DISK_SIZE "$LEGACY_COMPOSE_FILE")
@@ -355,12 +355,24 @@ migrate_legacy_compose() {
   shared="$HOME/Windows"
 
   [[ -z $tz ]] && tz="UTC"
+
+  # Pre-4.0.1 installers accepted any username (including emails). The elevated
+  # writer now rejects those. USERNAME is only used during first-run unattend,
+  # so fall back to docker like install does; keep the original for RDP so the
+  # existing Windows account still logs in.
+  rdp_username=$username
+  if ! valid_username "$username"; then
+    echo "Legacy username is not valid for the compose writer; using docker. The Windows account itself is unchanged."
+    username="docker"
+    [[ -n $rdp_username ]] || rdp_username="docker"
+  fi
+
   if ! write_compose "$ram" "$cores" "$disk" "$username" "$password" "$tz" "$storage" "$shared"; then
     echo "Could not migrate the existing configuration automatically." >&2
     echo "Re-run: omarchy-windows-vm install" >&2
     return 1
   fi
-  write_credentials "$username" "$password"
+  write_credentials "$rdp_username" "$password"
   rm -f "$LEGACY_COMPOSE_FILE"
 }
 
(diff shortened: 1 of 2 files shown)


Elshayib:
Rebased onto current `quattro` at `29f31e32` — PR is mergeable again (`clean`).

Both files in the original commit had been rewritten upstream (`migrate_legacy_compose` no longer takes storage/shared paths; the compose test mount suite changed). Kept only the username fallback: invalid legacy `USERNAME` → compose uses `docker` (same as install); RDP credentials keep the original name; `valid_username` in `__priv_write_compose` is unchanged. Dropped an obsolete symlink bring-up case that `quattro` already covers.

`bash test/shell.d/windows-vm-compose-test.sh` passes, including the sanitization assertion.

Issue #7049, opened 2026-08-16
Fresh Quattro install ignores installer keyboard layout choice — vconsole.conf gets only KEYMAP, Hyprland falls back to us


Asknorway:
### System details

- Omarchy 4.0.0-1 (Quattro), fresh install on 2026-08-15 (no upgrade history in pacman.log)
- MacBook Pro 2019 (Intel Core i9-9880H, Radeon Pro 5500M, Apple T2) running the linux-t2 kernel
- Norwegian keyboard, selected during installation

### What's wrong?

On a fresh Quattro install where a non-US keyboard layout is chosen in the installer, the Hyprland session comes up with the US layout anyway.

The installer records the choice only as a console keymap:

```conf
# /etc/vconsole.conf
KEYMAP=no-latin1
```

It writes no `XKBLAYOUT`/`XKBVARIANT`. But the packaged Hyprland default derives the session layout exclusively from `XKBLAYOUT`:

```lua
-- /usr/share/omarchy/default/hypr/input.lua
local kb_layout = vconsole.XKBLAYOUT or "us"
local kb_variant = vconsole.XKBVARIANT or ""
```

With only `KEYMAP` present, this silently falls back to `us`, so the console uses Norwegian while the desktop types US/ANSI.

The fcitx5 profile shipped for the compose-key setup is also hardcoded to `keyboard-us` (`~/.config/fcitx5/profile`: `DefaultIM=keyboard-us`), so even after fixing the Hyprland layout, fcitx5's virtual keyboard keeps serving US until its profile is corrected separately.

This is the fresh-install sibling of #6878: that report covers the Quattro *upgrade* dropping the layout when `vconsole.conf` only has `KEYMAP`; here a brand-new install never gets an `XKBLAYOUT` in the first place, so the same `us` fallback triggers.

#### Steps to reproduce

1. Install Omarchy 4.0.0 from scratch and pick a non-US layout (e.g. Norwegian) in the installer.
2. Boot into the installed system and log in to the Hyprland session.
3. Type in any application.

#### Expected behavior

The session uses the layout chosen during installation (here `no`), matching the console keymap.

#### Actual behavior

`hyprctl getoption input:kb_layout` reports `us`; `/etc/vconsole.conf` contains only `KEYMAP=no-latin1` with no `XKBLAYOUT`, so the packaged `input.lua` falls back to `us`. fcitx5 additionally stays on `keyboard-us`.

#### Workaround

```lua
[...]

Issue #6878, opened 2026-08-14
Quattro upgrade drops non-US keyboard layout when vconsole.conf only has KEYMAP


cempack:
### System details

- Omarchy 4.0.0-1 (Quattro stable), upgraded from a pre-Quattro installation
- Linux 7.1.8-arch1-3
- Intel Core Ultra 5 228V
- Intel Arc Graphics 130V/140V
- French/AZERTY keyboard layout

### What's wrong?

After upgrading an existing Omarchy installation to Quattro and rebooting, the Hyprland session uses the US/QWERTY layout instead of the previously configured French/AZERTY layout.

Before the upgrade, the active legacy config contained:

```conf
# ~/.config/hypr/input.conf
input {
  kb_layout = fr
  kb_options = compose:caps
}
```

The system console config contains:

```conf
# /etc/vconsole.conf
KEYMAP=fr
```

It does not contain `XKBLAYOUT`, which is valid: `vconsole.conf` only guarantees the console `KEYMAP` setting.

The Quattro upgrade leaves the legacy `input.conf` in place but switches Hyprland to the new Lua entry points. It always installs a stock `~/.config/hypr/input.lua`, and that file has no active keyboard-layout override. The packaged `default/hypr/input.lua` then does:

```lua
local kb_layout = vconsole.XKBLAYOUT or "us"
```

Because this installation has `KEYMAP=fr` but no `XKBLAYOUT`, the effective layout silently falls back to `us`. The migration `1781485962.sh` also sees the stock `input.lua` and leaves/replaces it with the packaged version, so the old `kb_layout = fr` is never carried over.

#### Steps to reproduce

1. Start with a pre-Quattro Omarchy installation where:
   - `~/.config/hypr/input.conf` contains `kb_layout = fr`
   - `/etc/vconsole.conf` contains only `KEYMAP=fr`
2. Upgrade to Quattro.
3. Reboot into the new Lua-based Hyprland configuration.
4. Type in any application.
5. Observe that the session uses US/QWERTY instead of French/AZERTY.

#### Expected behavior

The Quattro upgrade should preserve the explicitly configured keyboard layout from the legacy `input.conf`.

#### Actual behavior

The explicit legacy layout is ignored, and the new default falls back to `us`.

#### Workaround

Adding this to `~/.config/hypr/input.lua` restores the correct layout:

```lua
hl.config({
  input = {
    kb_layout = "fr",
  },
})
```

#### Suggested fix

[...]

PR #8685, opened 2026-08-27
Derive the Hyprland keyboard layout from the console keymap

ForkIt369:
Fresh Quattro installs persist the selected keyboard with systemd-firstboot --keymap, which writes KEYMAP to /etc/vconsole.conf. The Hyprland input defaults only read XKBLAYOUT, so selections such as Colemak silently fell back to the US layout in the session.

When XKBLAYOUT is absent or empty, derive the XKB layout and variant from KEYMAP through /usr/share/systemd/kbd-model-map. Explicit XKB settings still take precedence, and resolved non-Latin layouts continue through the existing US-prefix handling.

Nine keymaps offered by the installer have no entry in the systemd map, so a small fallback table covers those choices. The test suite also keeps the installer list synchronized with either the systemd map or that table.

One existing mismatch is made explicit in the code: the installer labels the azerty console keymap as Azerbaijani, while that console keymap is historically French AZERTY. The fallback preserves the installer intent by resolving it to the az XKB layout; the installer label/keymap pairing can be corrected separately.

Addresses #7049 and #6878 without closing them: this fixes the Hyprland session layout, while the fcitx5 and installer-persistence aspects remain outside this change.

Tested:
- bash test/shell.d/hyprland-keyboard-layout-test.sh (23 checks pass)
- ./test/cli (passes)
- Patched module against this affected host with KEYMAP=colemak resolves layout=us and variant=colemak
- ./test/shell: the keyboard test passes; five unrelated files also fail on a clean upstream checkout because local package fixtures or external access are unavailable. A transient screenshot sanity failure passed on immediate rerun.

### DIFF
--- a/default/hypr/input.lua
+++ b/default/hypr/input.lua
@@ -21,6 +21,49 @@ local function read_vconsole()
   return values
 end
 
+-- systemd translates console keymaps through this table when it can. Some
+-- keymaps offered by Omarchy have no entry there, so keep their XKB spelling
+-- here for the same fallback path.
+local keymap_xkb_fallbacks = {
+  -- The installer labels this console keymap as Azerbaijani. The console
+  -- keymap itself is historically French AZERTY; keep installer intent here.
+  azerty = { "az", "" },
+  ["bg-cp1251"] = { "bg", "" },
+  colemak = { "us", "colemak" },
+  cz = { "cz", "" },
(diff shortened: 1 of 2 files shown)


ralfboltshauser:
keyboard layouts like German Switzerland is also not persisted. This would fix it!

PR #8692, opened 2026-08-28
Resolve mise wrapper binaries before execution

### LINKED ISSUE #7234: mise wrappers still exec-loop when mise x falls back to a PATH lookup (residual case of #6349)
### Context / prior art

#6349 reported that the generated wrappers exec-loop, and #6350 fixed the
PATH-ordering case by moving the template to `exec mise x "$package" -- "$bin" "$@"`.

@padyfec then documented a **residual** loop in a follow-up comment on that
(now closed) issue: when a tool's mise install dir exists but holds no binary,
`mise x` has no bin dir to prepend, falls back to a plain PATH lookup, finds the
wrapper in `~/.local/bin`, and re-enters it forever.

That residual case doesn't appear to be tracked anywhere open, so this issue is
to track it — and to add a **second, distinct trigger** for the same structural
flaw.

### What's different here

padyfec's trigger is a broken/empty install. Mine isn't: every install dir is
fully populated and every tool runs correctly from an interactive shell.

```console
$ opencode --version
1.18.18
$ mise which opencode
/home/pms/.local/share/mise/installs/opencode/latest/opencode
```

Yet **every** wrapper exec-loops when the tool is spawned by a GUI application —
in my case T3 Code (Electron), whose server sidecar shells out to
`opencode --version` as a provider health check.

### Symptom

Not a crash, and not obviously a loop — `exec` keeps it inside a single PID, so
it presents as one wedged process pinning a core:

```console
$ ps -o pid,%cpu,etime,cmd -p 2692953
    PID %CPU     ELAPSED CMD
2692953 42.2       12:57 /bin/bash /home/pms/.local/bin/opencode --version
```

That is a single `opencode --version` still running 13 minutes later, `wchan:
do_wait`, continuously re-spawning `mise use`. Every provider in the app's UI
reported *"provider status has not been checked in this session yet"*, because no
health check ever returned. Host accounting showed the app consuming **1h 4min of
CPU over 54min of wall clock**.

### Confirmation

The bin name resolves back to the wrapper rather than the installed binary:

```console
$ mise x opencode -- sh -c 'command -v opencode'
/home/pms/.local/bin/opencode        # the wrapper, not the installed binary
```

A depth-counted replica of the stock wrapper, run under the app's environment
with the wrapper's directory first on PATH. Note the **constant PID** — this is
the `exec` loop, not repeated forking:

```
depth=1 pid=3705214
depth=2 pid=3705214
depth=3 pid=3705214
depth=4 pid=3705214   <- hard stop I added; would otherwise continue
```

Reproduced identically with `codex`, `claude`, `gh`, `gemini` and `grok` — every
wrapped tool, since they share the template.

### Honest scoping

I could not reduce this to a minimal repro. It reproduces reliably under the
spawning app's full environment (191 vars, captured from `/proc/<pid>/environ`)
with the wrapper directory first on PATH, but **not** in a bare `env -i` shell
with the same PATH ordering. So something in that environment decides whether
`mise x` prepends the install bin dir — consistent with both #6349's "mise PATH
hook isn't active" analysis and @padyfec's "no bin dirs to prepend" analysis.

The common factor across all three triggers is structural: the wrapper execs a
**bare name that it itself shadows on PATH**, so any condition that makes `mise x`
fall back to a PATH lookup turns into an infinite loop rather than an error.

### Suggested fix

@padyfec's hardening in #6349 addresses this directly — resolve the real binary
up front, exec the absolute path, and never fall back to a bare-name lookup:

```bash
BIN=$(mise which "$bin" 2>/dev/null)
[ -x "$BIN" ] || { mise install -f "$package@latest" >/dev/null 2>&1 || true; BIN=$(mise which "$bin" 2>/dev/null); }
[ -x "$BIN" ] || { echo "$command: mise package '$package' provides no executable '$bin'" >&2; exit 1; }
exec "$BIN" "$@"
```

If a PATH-based fallback is kept for any reason, dropping the wrapper's own
directory from PATH first makes re-entry structurally impossible:

```bash
__self_dir=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd)
PATH=$(printf '%s' "$PATH" | tr ':' '\n' | grep -vxF "$__self_dir" | paste -sd: -)
```

I'm running the `mise which` form locally across all 13 wrappers. Under the same
environment that previously hung forever, the checks now complete in 66–416 ms
with clean stdout.

### Environment

Omarchy 4.0.0-1 · mise 2026.8.3 · opencode 1.18.18 · codex 0.147.0 ·
claude 2.1.233 · gh 2.97.0 · T3 Code 0.0.33 (Electron) · Linux 7.1.8-arch1-3

### LINKED ISSUE #7360: mise tool wrappers can exec themselves forever: `mise x <pkg> -- <bin>` resolves <bin> back to the wrapper
Separate from #6948 (invalid `config.toml`) — same generated wrappers, different failure. This one presents as a **hang** rather than an error, so it is easy to misattribute to the network or to the tool being probed.

### Mechanism

`omarchy-mise-install` generates:

```bash
#!/bin/bash
export MISE_MINIMUM_RELEASE_AGE=0
mise use -g "codex" || exit 1
exec mise x "codex" -- "codex" "$@"
```

That final `codex` is resolved by `mise x` through `PATH` — and `~/.local/bin/codex` **is** the wrapper, sitting on that same `PATH`. When mise does not place its install dir ahead of `~/.local/bin`, the wrapper execs itself.

Because it is `exec`, the process count never grows. There is no fork bomb to spot in `htop`: one process spins, and whatever launched it eventually gives up.

PATH lookup inside `mise x` does find the wrapper:

```
$ mise x codex -- sh -c 'command -v codex'
/home/you/.local/bin/codex
```

### Evidence

I added a depth counter to the wrapper, exported so it survives `exec` (an exec loop should therefore show a **constant pid at rising depth**), bounded at 40, and let T3 Code probe its Codex provider:

```
2026-08-18T15:09:05 depth=1   pid=1459179  argv=app-server
2026-08-18T15:09:05 depth=2   pid=1459179  argv=app-server
...
2026-08-18T15:09:05 depth=40  pid=1459179  argv=app-server
2026-08-18T15:09:05 ABORT at depth 40
```

Same pid, depth 1 → 40. Five such chains in one session (206 log lines, 5 aborts): three `app-server` probes, one `-s read-only -a untrusted app-server`, and one with empty argv.

Unbounded, the app reports:

> Timed out while checking Codex app-server provider status.

With the bound in place it becomes:

> Codex app-server provider probe failed: Codex App Server process exited with code 90.

(90 being the deliberate abort — which is what confirms the loop is real and not an artifact of instrumentation.)

### I could not reproduce it from a shell

`~/.local/bin/codex --version` from an interactive shell gives `depth=1` and exits correctly, every time. Things I tried and which did **not** reproduce it:

- cwd variations, including a directory containing `.mise.toml`
- a `PATH` emulating an already-activated mise environment, with `installs/<tool>/...` dirs present and `~/.local/bin` ahead of them
- `__MISE_ORIG_PATH`, `MISE_SHELL`, `__MISE_SESSION` set
- replaying the spawning app's exact environment captured from `/proc/<pid>/environ`

It reproduces reliably when the app itself spawns the wrapper. So something about that spawn makes `mise x` fall back to a PATH lookup for its final argument instead of resolving the tool itself, and I have not identified what. Flagging that gap explicitly rather than guessing.

### Suggested fix

Stop the wrapper from being able to resolve its own name. Either exec the resolved binary directly:

```bash
exec "$(mise which "$bin")" "$@"
```

or keep `mise x` and add a re-entry guard:

```bash
__mw=_MISE_WRAP_$(printf %s "$package" | tr -c "A-Za-z0-9" _)
if [[ -n ${!__mw:-} ]]; then
  __real=$(mise which "$bin" 2>/dev/null)
  [[ -n $__real && -x $__real ]] && exec "$__real" "$@"
  exit 127
fi
export "$__mw=1"
```

The guard has a useful side effect: it also covers tools that shell out to their own name. `codex` spawns `codex exec ...`, which otherwise re-enters the wrapper too.

Guarding all 13 wrappers fixed it here, and the provider has been stable since.

Worth noting that migration `1784909971.sh` is already titled *"Regenerate mise wrappers to stop them recursing through PATH"*, so this shape of bug has bitten before.

### Environment

- Omarchy 4.0.0-1, Arch
- mise 2026.8.3
- codex-cli 0.147.0
- Triggering app: T3 Code 0.0.33, which spawns provider probes on a 10s budget

imadtg:
Fixes #7360.
Fixes #7234.

## Problem

When `~/.local/bin` comes before mise's binaries in `PATH`, a generated wrapper can execute itself through `mise x` and hang forever.

## Fix

The wrapper first activates the requested package:

```bash
mise use -g --quiet "$package" || exit 1
```

A straightforward way to resolve the command would be:

```bash
bin_path=$(mise which "$bin" --tool="$package") || exit 1
```

Here, `--tool="$package"` tells mise which tool/version to use while resolving `"$bin"`. It does not guarantee that the returned executable comes from that package's own install directory. Another active package can provide the same executable name.

The wrapper therefore asks for the requested package's runtime directories and searches them directly:

```bash
bin_paths=$(mise bin-paths "$package") || exit 1
```

It passes the first matching absolute path to `mise x` and reports an error when no match exists. Keeping `mise x` supplies the Node environment that npm-backed tools need.

Migration `1787809285.sh` regenerates wrappers matching the vulnerable template. It preserves custom wrappers and symlinks and leaves repaired wrappers alone.

## Testing

All tests in `./test/all` pass: CLI and all 237 shell test files.

The new regression coverage verifies that:

- `mise bin-paths "$package"` stays within the requested package;
- a same-named executable from another package is rejected;
- package runtime variables and argument boundaries survive `mise x`;
- shell metacharacters in bin names remain literal, and bin names containing paths are rejected;
- missing or failing `bin-paths` stops before execution;
- migration remains safe and idempotent.

Written with Codex and reviewed by a human.

### DIFF
--- a/bin/omarchy-mise-install
+++ b/bin/omarchy-mise-install
@@ -53,7 +53,8 @@ cat >"$HOME/.local/bin/$command" <<EOF
 #!/bin/bash
 export MISE_MINIMUM_RELEASE_AGE=0
 mise use -g --quiet $package_arg || exit 1
-exec mise x $package_arg -- $bin_arg "\$@"
+bin_path=\$(mise which $bin_arg --tool=$package_arg) || exit 1
+exec mise x $package_arg -- "\$bin_path" "\$@"
 EOF
 
 chmod +x "$HOME/.local/bin/$command"
--- a/bin/omarchy-mise-install
+++ b/bin/omarchy-mise-install
@@ -25,6 +25,11 @@ case "$command" in
     ;;
 esac
 
+if [[ $bin == */* || $bin == "." || $bin == ".." ]]; then
+  echo "omarchy-mise-install: bin-name must be a basename" >&2
+  exit 1
+fi
+
 mkdir -p "$HOME/.local/bin"
 
 # The heredoc below is unquoted, so whatever these hold is written into the
@@ -53,7 +58,20 @@ cat >"$HOME/.local/bin/$command" <<EOF
 #!/bin/bash
 export MISE_MINIMUM_RELEASE_AGE=0
 mise use -g --quiet $package_arg || exit 1
-bin_path=\$(mise which $bin_arg --tool=$package_arg) || exit 1
+bin_path=
+bin_paths=\$(mise bin-paths $package_arg) || exit 1
+while IFS= read -r bin_dir; do
+  [[ -n \$bin_dir ]] || continue
+  candidate="\$bin_dir"/$bin_arg
+  if [[ -f \$candidate && -x \$candidate ]]; then
+    bin_path=\$candidate
+    break
+  fi
+done <<<"\$bin_paths"
+if [[ -z \$bin_path ]]; then
+  printf "%s: mise package '%s' provides no executable '%s'\n" "\${0##*/}" $package_arg $bin_arg >&2
+  exit 1
+fi
 exec mise x $package_arg -- "\$bin_path" "\$@"
 EOF
 
--- a/migrations/1787809285.sh
+++ b/migrations/1787809285.sh
@@ -1,0 +1,30 @@
+echo "Regenerate mise wrappers to resolve their binaries before execution"
+
+# `mise x "$package" -- "$bin"` still resolves the bare bin through PATH.
+# When ~/.local/bin is ahead of mise's shims, that lookup finds the generated
+# wrapper itself and execs it forever. Regenerate wrappers on the current
+# template so they pass mise x the package-scoped absolute path instead.
+
+current_template() {
+  local package=$1 bin=$2
+
+  printf '#!/bin/bash\nexport MISE_MINIMUM_RELEASE_AGE=0\nmise use -g --quiet "%s" || exit 1\nexec mise x "%s" -- "%s" "$@"' "$package" "$package" "$bin"
+}
+
+bin_dir="$HOME/.local/bin"
+
+[[ -d $bin_dir ]] || exit 0
+
+for wrapper in "$bin_dir"/*; do
+  [[ -f $wrapper && ! -L $wrapper && -r $wrapper ]] || continue
+  (($(stat -c%s "$wrapper") <= 1024)) || continue
+
+  contents=$(<"$wrapper")
+  package=$(sed -n 's/^mise use -g --quiet "\(.*\)" || exit 1$/\1/p' <<<"$contents")
+  bin=$(sed -n 's/^exec mise x ".*" -- "\(.*\)" "\$@"$/\1/p' <<<"$contents")
+
+  [[ -n $package && -n $bin ]] || continue
+  [[ $contents == "$(current_template "$package" "$bin")" ]] || continue
+
+  omarchy-mise-install "$package" "${wrapper##*/}" "$bin"
+done
--- a/test/shell.d/default-agent-test.sh
+++ b/test/shell.d/default-agent-test.sh
@@ -20,7 +20,8 @@ stub_log="$test_tmp/stubs"
 terminal_log="$test_tmp/terminal"
 menu_log="$test_tmp/menu"
 muse_login_log="$test_tmp/muse-login"
-mkdir -p "$mock_bin" "$test_home"
+resolved_dir="$test_tmp/resolved"
+mkdir -p "$mock_bin" "$test_home" "$resolved_dir"
 
 cat >"$mock_bin/omarchy-notification-send" <<'SH'
 #!/bin/bash
@@ -62,8 +63,8 @@ if [[ $1 == "where" ]]; then
   exit
 fi
 
-if [[ $1 == "which" ]]; then
-  printf '/resolved/%s\n' "$2"
+if [[ $1 == "bin-paths" ]]; then
+  printf '%s\n' "$OMARCHY_TEST_RESOLVED_DIR"
   exit
 fi
 
@@ -114,6 +115,7 @@ export OMARCHY_TEST_STUB_LOG="$stub_log"
 export OMARCHY_TEST_AGENT_TERMINAL_LOG="$terminal_log"
 export OMARCHY_TEST_AGENT_MENU_LOG="$menu_log"
 export OMARCHY_TEST_MUSE_LOGIN_LOG="$muse_login_log"
+export OMARCHY_TEST_RESOLVED_DIR="$resolved_dir"
 export OMARCHY_PATH="$ROOT"
 
 grok_package="npm:@xai-official/grok"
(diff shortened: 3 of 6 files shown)


davidavasalcai:
## Field-tested on a real install — this fix is the right shape

I hit this exact loop (`mise x` re-resolving the bare command name back to the wrapper in `~/.local/bin`, exec-looping in a single PID until timeout, flooding stderr with `mise ~/.config/mise/config.toml tools: <pkg>@<version>` lines) on Omarchy today, for `gh`, `claude`, `codex`, `opencode` and `gemini` at once.

My local workaround landed on the same principle as this PR: resolve the real binary first, then exec it directly —

```bash
mise use -g --quiet "$pkg" || exit 1
exec "$(mise which "$pkg")" "$@"
```

Verified on this machine:

- `mise which gh` → `~/.local/share/mise/installs/gh/latest/gh_2.100.0_linux_amd64/bin/gh` (real binary, not the wrapper)
- `gh --version`, `gh auth status`, `gh api user` — all return instantly, no loop, no flood
- `claude --version` (2.1.266), `codex --version` (0.153.4), `opencode --version` (1.18.30), `gemini --version` (0.59.0) — all instant

So resolving to an absolute path before executing is confirmed working in production. Appreciate the `--tool` scoping in this PR (plain `mise which "$bin"` can pick a same-named binary from another package — that's why resolving from the package's `bin-paths` is the better shape). Happy to test the branch here if helpful.

imadtg:
> ## Field-tested on a real install — this fix is the right shape
> 
> I hit this exact loop (`mise x` re-resolving the bare command name back to the wrapper in `~/.local/bin`, exec-looping in a single PID until timeout, flooding stderr with `mise ~/.config/mise/config.toml tools: <pkg>@<version>` lines) on Omarchy today, for `gh`, `claude`, `codex`, `opencode` and `gemini` at once.
> 
> My local workaround landed on the same principle as this PR: resolve the real binary first, then exec it directly —
> 
> ```shell
> mise use -g --quiet "$pkg" || exit 1
> exec "$(mise which "$pkg")" "$@"
> ```
> 
> Verified on this machine:
> 
>     * `mise which gh` → `~/.local/share/mise/installs/gh/latest/gh_2.100.0_linux_amd64/bin/gh` (real binary, not the wrapper)
> 
>     * `gh --version`, `gh auth status`, `gh api user` — all return instantly, no loop, no flood
> 
>     * `claude --version` (2.1.266), `codex --version` (0.153.4), `opencode --version` (1.18.30), `gemini --version` (0.59.0) — all instant
> 
> 
> So resolving to an absolute path before executing is confirmed working in production. Appreciate the `--tool` scoping in this PR (plain `mise which "$bin"` can pick a same-named binary from another package — that's why resolving from the package's `bin-paths` is the better shape). Happy to test the branch here if helpful.

Thanks for taking the time to try this fix. but i need to point out one thing, this is no longer using the `mise which --tool` approach, since that could  still confuse same named binaries. it's using a more robust approach as described in the PR body (introduced in #8798)

PR #8709, opened 2026-08-28
fix: arm signature verification for the T2 repo and close the quattro override window

rmacy:
Follow-up to #8072 (Require signed packages from the Omarchy repository). Two repositories in the tree still accept unauthenticated packages as root; this closes what can be closed Omarchy-side and arms the rest.

## 1. `install/hardware/pacman.sh` — the T2 stanza disables verification outright

```
[arch-mact2]
Server = https://github.com/NoaHimesaka1873/arch-mact2-mirror/releases/download/release
SigLevel = Never
```

`SigLevel = Never` means pacman performs no package authentication at all for the life of the install: any compromise of the arch-mact2 repository, the `NoaHimesaka1873` account, or GitHub's release delivery path is root code execution on every T2 machine, through routine unattended `pacman -Syu` transactions. The repo ships the T2 kernel (`linux-t2`), firmware, and fan daemon, all executed or loaded at boot.

This differs from the #8072 finding in one important way: **upstream cannot enforce signatures today.** Verified against the live repository:

- none of the ~40 packages in `arch-mact2.db` carries a `%PGPSIG%` entry except `apple-bcm-wifi-firmware`;
- standalone `.sig` sidecars 404 for every package checked;
- `arch-mact2.db.sig` exists but is stale — it verifies as **BAD** (signed 2022-02-26, key `8BE1FEE1 4302 371D EF6F 910A 0E58 77AC 225D 1980`, "Noa Himesaka <himesaka@noa.codes>") against the current database.

So `SigLevel = Required` would refuse every T2 package and break the install. The stanza instead **arms** verification as far as upstream permits:

- The pinned upstream signing key is imported and locally signed (`pacman-key --recv-keys … --keyserver keyserver.ubuntu.com`, then `--lsign-key`). A failed import **refuses to enable the repository** rather than silently continuing unverified — a T2 install without its kernel cannot succeed anyway, and an explicit error beats an unauthenticated stanza.
- `SigLevel = PackageOptional DatabaseNever`: unsigned packages shipped today are accepted; any package that gains a signature is verified the moment it appears (a present-but-invalid signature is rejected, so a later compromise cannot strip signatures past an already-imported key); the stale database signature is skipped, which a plain `Optional` would find and reject.

**Residual risk, stated plainly:** until upstream signs its packages, this repository is accepted without authentication and origin compromise is still root on T2 machines. This PR cannot fix that; upstream signing can. Disclosed to security@omarchy.org.

## 2. `bin/omarchy-upgrade-to-quattro` — the bootstrap override outlives the keyring install

The upgrader regenerates the `[omarchy]` stanza with `SigLevel = Optional TrustAll` because the keyring package that provides trust arrives through the very repository it authenticates. #8072's migration strips that override on the *next* update pass — but everything the upgrader itself installs in between (hardware transition packages, the entire `install_omarchy_quattro_packages` set) happens unauthenticated.

New `require_signed_omarchy_repo()` runs immediately after `install_keyrings` and deletes the override line, so every later transaction in the upgrade verifies signatures like a normal post-#8072 machine. If the keyring failed to install, the override is left in place and the machine keeps working — the log says why.

## Tests

New `test/shell.d/t2-repo-signature-test.sh`:

- tree-wide guard: no repository stanza anywhere in `install/`, `default/`, `bin/` sets `SigLevel = Never`;
- behavioral, with mocked `lspci` (T2 PCI ids) and `pacman-key` (arg-validating: only the pinned fingerprint is accepted): stanza written with the armed policy, key imported from the keyserver and locally signed, write is idempotent on a second run;
- failure path: a `pacman-key` that fails leaves **no stanza** and exits non-zero;
- quattro guards: the override is still written (bootstrap), the removal step exists, and it runs on the line immediately after `install_keyrings`.

`upgrade-to-quattro-test.sh` passes unchanged (22/22). All touched scripts pass `bash -n`.

## Notes for reviewers

- The behavioral test reaches the install script through `OMARCHY_T2_PACMAN_CONF`, a one-variable seam so the test never touches the real `/etc/pacman.conf`; the variable is unset on real installs and the script writes `/etc/pacman.conf` exactly as before.
- The stanza keeps a plain heredoc; no shell expansion enters the config file.
- Key fingerprint source: fetched from keyserver.ubuntu.com and cross-checked against the `arch-mact2.db.sig` issuer keyid; the t2linux wiki documents the same identity.

### DIFF
--- a/bin/omarchy-upgrade-to-quattro
+++ b/bin/omarchy-upgrade-to-quattro
@@ -455,6 +455,25 @@ install_keyrings() {
   }
 }
 
+# install_keyrings can only run while the [omarchy] stanza accepts unsigned
+# packages: the keyring package that provides trust arrives through the very
+# repository it is meant to authenticate. Once it is installed the override
+# has served its purpose, so drop it before anything else is fetched -- every
+# later transaction in this upgrade then verifies signatures like a normal
+# post-#8072 machine instead of trusting the wire.
+require_signed_omarchy_repo() {
+  if ! as_root pacman -Q omarchy-keyring >/dev/null 2>&1; then
+    warn "omarchy-keyring did not install; leaving the unsigned-package override in place."
+    return
+  fi
+
+  local tmp
+  tmp=$(mktemp)
+  sed '/^\[omarchy\]/,/^\[/{/^SigLevel = Optional TrustAll$/d}' /etc/pacman.conf >"$tmp"
+  as_root install -m 0644 "$tmp" /etc/pacman.conf
+  rm -f "$tmp"
+}
+
 remove_legacy_installer_package() {
   if pacman -Q omarchy-installer >/dev/null 2>&1; then
     log "Removing legacy omarchy-installer package"
@@ -2319,6 +2338,7 @@ suppress_hyprland_config_reload
 create_pre_upgrade_snapshot
 configure_pacman_channel
 install_keyrings
+require_signed_omarchy_repo
 remove_legacy_installer_package
 remove_legacy_limine_configs
 remove_conflicting_legacy_packages
--- a/install/hardware/pacman.sh
+++ b/install/hardware/pacman.sh
@@ -9,9 +9,12 @@
 #
 # - PackageOptional verifies any package that carries a signature the moment
 #   upstream starts signing, while still accepting the unsigned packages
-#   shipped today. A present-but-invalid signature is rejected, so a later
-#   compromise cannot rely on swapping signatures past an already-imported
-#   key.
+#   shipped today. A signature that is present but invalid is rejected, so an
+#   already-imported key cannot be walked past with a bad signature. It does
+#   NOT prevent signature stripping: an attacker who controls the repository
+#   can simply stop publishing signatures and pacman accepts the unsigned
+#   result. Stripping is only excluded once upstream signs consistently and
+#   this policy tightens to PackageRequired.
 # - DatabaseNever skips the stale database signature, which a plain Optional
 #   would find and reject.
 #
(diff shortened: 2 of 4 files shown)


rmacy:
Addressed both review comments in 5b09c4e9:

**1. Signature-stripping caveat (line 14)** — correct. `PackageOptional` rejects a present-but-invalid signature; it does not stop a compromised repository from simply not publishing signatures. The comment now states the guarantee precisely (present-but-invalid rejected; stripping excluded only once upstream signs consistently and the policy tightens to `PackageRequired`). The honest bottom line is unchanged and stated in the PR body: until upstream signs, this repository is accepted without authentication.

**2. Existing T2 installs skipped by the stanza guard (line 29)** — real defect, and the more important of the two: every T2 machine installed before this change already carries `[arch-mact2]` with `SigLevel = Never`, and normal updates run migrations rather than the hardware installer, so the original version only fixed fresh installs. Now:

- `install/hardware/pacman.sh` reconciles in place: if the stanza exists with `SigLevel = Never`, the pinned key is imported/locally signed and the line is upgraded to `PackageOptional DatabaseNever`. A failed import there is non-fatal (packages are unsigned either way, so the machine keeps working and a later retry can succeed) — asymmetric with the fresh-install path, which still fails closed because an install without its kernel cannot proceed anyway.
- New `migrations/1787720000.sh` performs the same gated reconciliation for the installed base via the normal update path. Self-detecting and idempotent: it acts only when the stanza exists AND still says `Never`, and touches nothing when the key import fails.

Tests extended to cover both: legacy-stanza reconciliation (in-place upgrade, no duplicate stanza, key imported and signed), reconciliation failure non-fatality, and migration guards (pinned fingerprint, keyserver, lsign, armed policy, Never-gated mutation). Full suite green; `upgrade-to-quattro-test.sh` 22/22 unchanged.

PR #8712, opened 2026-08-28
Skip powerprofilesctl when daemon is down

### LINKED ISSUE #8596: powerprofilesctl crashes (SIGABRT) when power-profiles-daemon is masked — hit on every power event with TLP
## Summary

`powerprofilesctl` (the Python CLI shipped with `power-profiles-daemon`) hard-crashes with `SIGABRT` when invoked while `power-profiles-daemon.service` is masked/inactive — e.g. on a TLP-managed setup, which is a common alternative to `power-profiles-daemon` on laptops. The crash is a Python interpreter-shutdown race (`PyGILState_Ensure()` called against an already-torn-down GIL from a GObject finalizer), not a clean error return, so it can't be handled by checking the exit code alone.

Several Omarchy Quickshell plugins call `powerprofilesctl` / `omarchy-powerprofiles-set` directly and hit this on every AC/battery transition and every time the power menu is opened.

## Environment

- Omarchy 4.0.0-1 (`/usr/share/omarchy/version`: `4.0.0.alpha`)
- power-profiles-daemon 0.30-1 (installed but masked — TLP manages power instead)
- tlp 1.10.2-1
- Kernel 7.1.8-arch1-3
- Python 3.14.7
- CPU: Intel Core i7-7820HQ

## Reproduction

1. Mask `power-profiles-daemon.service` (as recommended when using TLP instead): `sudo systemctl mask power-profiles-daemon.service`
2. Trigger any of:
   - AC/battery power source change
   - Opening the Omarchy menu's power-profile submenu
   - Opening the Power bar widget/panel
3. `powerprofilesctl` crashes with SIGABRT instead of failing cleanly.

## Evidence

```
$ coredumpctl info 31765
           PID: 31765 (powerprofilesct)
           TID: 31805 (gdbus)
        Signal: 6 (ABRT) si_code: SI_TKILL
  Command Line: /bin/python3 /usr/bin/powerprofilesctl list
    Executable: /usr/bin/python3.14

Stack trace of thread 31805:
#0  raise (libc.so.6)
#1  abort (libc.so.6)
#2  n/a (libpython3.14.so.1.0 + 0x1642ca)
#3  _Py_FatalErrorFunc (libpython3.14.so.1.0)
#4  PyEval_RestoreThread (libpython3.14.so.1.0)
#5  PyGILState_Ensure (libpython3.14.so.1.0)
#6  n/a (_gi.cpython-314-x86_64-linux-gnu.so + 0x1156e)
#7  n/a (libglib-2.0.so.0)
#8  g_object_unref (libgobject-2.0.so.0)
#9  n/a (libgio-2.0.so.0)  [async D-Bus call teardown]
...
```

`PyGILState_Ensure()` is being called from a PyGObject callback fired during object finalization/interpreter shutdown, after the GIL has already been destroyed — this is a known class of PyGObject/GLib async-D-Bus race, not something specific to this machine. When `power-profiles-daemon` isn't running, `powerprofilesctl`'s D-Bus proxy setup takes a different (failing/retrying) code path, which appears to be what triggers the race here; against a live daemon it doesn't reproduce.

## Affected call sites (upstream Omarchy source)

- `shell/plugins/panels/power/Panel.qml` — profile-set action shells out to `omarchy-powerprofiles-set`
- `shell/plugins/services/battery/Service.qml` — calls `omarchy-powerprofiles-set` on every AC/battery transition
- `shell/plugins/menu/Menu.qml` — power-profile submenu shells out to `powerprofilesctl get`/`list` to populate rows

None of these currently check whether `power-profiles-daemon.service` is actually active before invoking `powerprofilesctl`/`omarchy-powerprofiles-set`.

## Proposed fix

Guard each call site with `systemctl is-active --quiet power-profiles-daemon.service` and skip the `powerprofilesctl` call entirely when it's masked/inactive, rather than letting it be invoked and crash. I have a working local patch that does this plus a best-effort fallback (deriving power-saver/balanced/performance from the CPU's own `energy_performance_preference` sysfs knob when no daemon is present) — happy to open a PR with it if that shape is welcome, or to split the guard-only fix from the EPP fallback if you'd rather review those separately.

## Expected behavior

`powerprofilesctl`/`omarchy-powerprofiles-set` should not be invoked at all when `power-profiles-daemon.service` isn't active, so TLP-only setups don't hit this crash on every power event and menu open.

Elshayib:
**What**: Do not call `powerprofilesctl` when `power-profiles-daemon` is inactive.

**Why**: Fixes #8596. TLP masks the daemon; `powerprofilesctl` then SIGABRTs, which takes down the battery plugin and the power-profile menu. No EPP fallback in this PR.

**How**: `systemctl is-active --quiet power-profiles-daemon.service` in `omarchy-powerprofiles-list` and `omarchy-powerprofiles-set` (exit 0 if down). Menu current-profile read now uses `omarchy-powerprofiles-list --active-state` instead of `powerprofilesctl get`. Rejected guarding only QML: Panel/battery go through the wrappers.

**Verified**:
- `bash -n` on list, set, and the test
- stubbed skip: inactive set exits 0, writes no `powerprofilesctl set`, no state file; inactive list is empty; active set still applies
- Menu.qml no longer contains `powerprofilesctl`
- existing set/restore/init assertions in `powerprofiles-set-test.sh` passed
- issue comment: https://github.com/basecamp/omarchy/issues/8596#issuecomment-5447823228

**NOT verified**:
- live TLP / masked `power-profiles-daemon` (Windows; no systemd)
- menu rendering / `omarchy-restart-shell` (no Hyprland/Quickshell)
- full `powerprofiles-set-test.sh` `rg` of QML paths (Windows path mapping)
- `test/all`, pacman, omarchy-iso

### DIFF
--- a/bin/omarchy-powerprofiles-list
+++ b/bin/omarchy-powerprofiles-list
@@ -8,6 +8,11 @@ if [[ ${1:-} != "" && ${1:-} != "--active-state" ]]; then
   exit 1
 fi
 
+# TLP masks power-profiles-daemon; calling powerprofilesctl then SIGABRTs.
+if ! systemctl is-active --quiet power-profiles-daemon.service; then
+  exit 0
+fi
+
 powerprofilesctl list 2>/dev/null |
   awk -v with_state="${1:+1}" '/^\s*[* ]\s*[a-zA-Z0-9-]+:$/ {
     active = ($1 == "*") ? 1 : 0
--- a/bin/omarchy-powerprofiles-set
+++ b/bin/omarchy-powerprofiles-set
@@ -27,6 +27,11 @@ case "$action" in
   *) usage ;;
 esac
 
+# TLP masks power-profiles-daemon; calling powerprofilesctl then SIGABRTs.
+if ! systemctl is-active --quiet power-profiles-daemon.service; then
+  exit 0
+fi
+
 mapfile -t profiles < <(omarchy-powerprofiles-list)
 
 profile_available() {
--- a/shell/plugins/menu/Menu.qml
+++ b/shell/plugins/menu/Menu.qml
@@ -274,7 +274,7 @@ Item {
       actionFor: function(value) { return "omarchy-font-set " + Util.shellQuote(value) }
     },
     "power-profiles": {
-      script: "current=$(powerprofilesctl get 2>/dev/null); omarchy-powerprofiles-list 2>/dev/null | while read -r p; do [[ -z $p ]] && continue; printf '%s\\t%s\\t%s\\n' \"$p\" \"$p\" \"$current\"; done",
+      script: "current=$(omarchy-powerprofiles-list --active-state 2>/dev/null | awk '$2==1{print $1;exit}'); omarchy-powerprofiles-list 2>/dev/null | while read -r p; do [[ -z $p ]] && continue; printf '%s\\t%s\\t%s\\n' \"$p\" \"$p\" \"$current\"; done",
       icon: "\udb81\udc0b",
       actionFor: function(value) { return "omarchy-powerprofiles-set autodetect " + Util.shellQuote(value) }
     }
--- a/test/shell.d/powerprofiles-set-test.sh
+++ b/test/shell.d/powerprofiles-set-test.sh
@@ -32,6 +32,17 @@ fi
 EOF
 chmod +x "$tmp_dir/bin/busctl"
 
+cat >"$tmp_dir/bin/systemctl" <<'EOF'
+#!/bin/bash
+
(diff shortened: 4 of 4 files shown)


sprajs:
Related work: #10951 replaces the battery service's periodic profile read with Quickshell's native `PowerProfiles.profile` subscription. Its conflicts with current `quattro` are resolved; upstream #11637 already changed that reader from Python to `busctl`, but still polls every two seconds.

The changes complement this PR: #10951 removes background reads, while these inactive-daemon guards cover the remaining menu and profile-setting/list calls. It does not replace the guards or fix the underlying CLI shutdown race.

Issue #1684, opened 2025-09-15
Slow launches on laptop


paul-fornage:
### System details

AMD Ryzen 9 5900HS, 16gb RAM

### What's wrong?

I am getting extremely slow launch times on my laptop. The desktop is very responsive, not slow otherwise, but yet it launches a terminal slower than a Pi. 

The delay is inconsistent and sometimes works fine, but this hardware should be very capable of going way faster.

Also happens when launching apps from the terminal, not just by keybind.

I'm gonna guess that this isn't an Omarchy problem, but I don't know what subsystem could be doing this, so if someone has any ideas, that would be appreciated. I'm suspecting it's power management related, or Hypr spawning new windows, but I'm not sure.

This happens with all apps as far as I can tell, but the TUIs have no excuse loading slow and I think demonstrate the problem better. Nautilus takes about 1-4s seconds to open on that machine, while it's instant on my other 2.

https://github.com/user-attachments/assets/93039240-2c45-4b1d-bcd3-fb368279b47a

dhh:
Omarchy 3.0 is going to ship with Walker 1.0 that uses a service. So this should be instant then. Still curious that it's this slow, though! But should be not a problem very soon.

gkurts:
I see this, too. The OS itself is wicket fast but walker can, randomly, be slow. Walker 1.0 will hopefully resolve it. 

<img width="431" height="358" alt="Image" src="https://github.com/user-attachments/assets/291f76e6-1cb1-4221-973e-aeb3fef37a4d" />

paul-fornage:
I will obviously wait before reopening this issue, but I want to clarify that running `alacritty` in the terminal is equally slow. I don't know how this system fits together, but things I would not expect to be related to Walker are also problematic. While it **is** only for launching windows, it seems to not just be walker.

Until I can test it, I will declare my skepticism but assume I'm wrong here

dhh:
Starting alacritty is slow on your system? Then it's probably not a walker issue! That's strange.

ColtonIdle:
using 3.0. walker menu is fast. starting things is still slow. terminal should be instant (like other os's) but it takes a good 200-300ms. chrome takes like 3 seconds. calculator took like 3 seconds.

running a framework 13 with AI 5 340

lennartbrandin:
I have similar issues and noticed the power-profile selection makes a big impact. This affects all opening windows including the launch bar (where it gets quickly noticable).
[...]

PR #8714, opened 2026-08-28
Reduce battery app-launch latency on Dell XPS 16

mapleroyal:
## Summary

- Explain that power-profile labels map to machine-specific firmware and driver behavior.
- Document a safe Balanced-versus-Performance check when every app is slow to show its first window only on battery.
- Point users to the remembered battery profile or a hardware-specific provider/plugin, and warn against copying low-level sysfs values across models.

## Why

Reports in #1684 mix launcher latency with a system-wide delay before unrelated apps become visible. On a tested Dell XPS 16 DA16260, literal 120 fps process-cold first-visible medians improved from 845 to 429 ms for Chrome, 389 to 206 ms for Ghostty, 702 to 367 ms for Nautilus, and 539 to 313 ms for Pulse when the battery Balanced policy was refined: 42–49% faster per app.

A separate six-block idle-power comparison estimated +0.167 W with a 95% interval of -0.009 to +0.342 W. That established neither a penalty nor a saving and is not a battery-runtime claim.

The measured low-level values are deliberately not included here: they are specific to the exact model and provider stack, and prior broad Intel integration in #5327 exposed regression risk. This change gives users the supported diagnosis and immediate workaround without changing defaults for unrelated hardware.

## Validation

- 108/108 accepted launch-screening trials across the full policy grid
- 32/32 accepted literal first-pixel A/B trials
- `git diff --check`
- `./test/all`: CLI passed; 202/207 shell test files passed. The five failures are local-environment checks unrelated to this docs-only change: missing `omarchy-pkgs` checkout (three), installed-command PATH isolation (one), and installed fastfetch config drift (one).

## Investigation credit

Investigation and validation were supported by OpenAI Codex (Sol model, Ultra reasoning effort) over roughly eight hours, including launch, 120 fps first-pixel, power, transition, and rollback testing.

### DIFF
--- a/manual/36-system-sleep.md
+++ b/manual/36-system-sleep.md
@@ -8,6 +8,8 @@ On a laptop, Omarchy remembers your power profile separately for plugged in and 
 
 You can see what your machine offers with `omarchy powerprofiles list`, and set the one you want for the state you're currently in with `omarchy powerprofiles set autodetect power-saver`. To set the other state without unplugging anything, name it directly: `omarchy powerprofiles set battery power-saver`. Whatever you pick is what you'll get back the next time you're in that state.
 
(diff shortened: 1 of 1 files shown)


mapleroyal:
Superseded by #8717. The docs-only approach did not implement the fix; the replacement contains the actual hardware-gated app-launch latency implementation and tests.

Issue #8007, opened 2026-08-24
Cloning a "kind: bar" plugin (omarchy plugin clone omarchy.bar) never renders


Charlieras262:
## What happened

Cloning the `omarchy.bar` plugin (`omarchy plugin clone omarchy.bar`) produces a bar that never renders. After the clone (which also switches `~/.config/omarchy/shell.json`'s `bar.id` to the new plugin), the top bar disappears entirely -- no `omarchy-bar` layer-shell surface is created at all.

This reproduces with a **byte-for-byte unedited clone** -- it's not caused by modifying `Bar.qml`, only by cloning and activating it.

## What I expected

Cloning `omarchy.bar` (a `"kinds": ["bar"]` plugin) should behave like cloning a `bar-widget` plugin (e.g. `omarchy.workspaces`) does: the clone becomes the active bar and renders identically to the stock one until edited.

## Steps to reproduce

```bash
omarchy plugin clone omarchy.bar
# output: "Cloned omarchy.bar to ~/.config/omarchy/plugins/<user>.bar and switched to <user>.bar"
hyprctl layers | grep -oE "namespace: .*" | sort -u
```

Expected: `omarchy-bar` (and `omarchy-background`) in the namespace list.
Actual: only `omarchy-background` (and sometimes `omarchy-notifications`) -- the bar's layer surface is never mapped.

Revert with `omarchy bar use omarchy.bar` to confirm the stock bar comes back immediately (it does).

## Diagnostics

`journalctl --user -b 0` shows this immediately after the clone/switch, before any edit to the cloned file:

```
WARN scene: file:///home/<user>/.config/omarchy/plugins/<user>.bar/Bar.qml[15:3]: Required property omarchyPath was not initialized
WARN scene: file:///home/<user>/.config/omarchy/plugins/<user>.bar/Bar.qml[17:3]: Required property barWidgetRegistry was not initialized
WARN scene: file:///home/<user>/.config/omarchy/plugins/<user>.bar/Bar.qml[22:3]: Required property barConfig was not initialized
```

No `Loader.Error` / "bar option ... failed to load" message is logged, so the `pluginBarLoader` in `shell.qml` does appear to reach `onLoaded` and call `configureBar(item, ...)` -- but the bar still never maps a layer-shell surface.

[...]

PR #8752, opened 2026-08-28
Add component wrapper in order to "preload" require properties

florann:
Adding a component "Bar" wrapper to make sure required properties are load before the component is spawned. 
It's the same mecanism as for the omarchy default bar but was missing for customs bars.

Relates to #8007

### DIFF
--- a/shell/shell.qml
+++ b/shell/shell.qml
@@ -234,6 +234,18 @@ ShellRoot {
     }
   }
 
+  Component {
+    id: customBarComponent
+
+    Bar {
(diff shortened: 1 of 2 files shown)


florann:
Solution change, 
Instead of statically instantiate the Bar component ( which is a mistakes ) we will create the Qml object before it's loaded and feed it with necessary properties.

bjarneo:
Automated duplication check: this pull request looks similar to #7254, which covers the same bar loader required props. I keep that one open and close this one to consolidate review.

PR #8757, opened 2026-08-28
Support fprintd-compatible backends other than libfprint in fingerprint setup

### LINKED ISSUE #7006: setup security fingerprint fails on sensors requiring open-fprintd (ThinkPad 06cb:009a)
## What happened

`omarchy setup security fingerprint` cannot complete on hardware whose fingerprint
sensor is unsupported by `libfprint` and requires the `open-fprintd` +
`python-validity` stack instead. This covers a large slice of ThinkPads, including
the T480/T480s/T490/X1C6 family (Synaptics/Validity `06cb:009a`).

`install/../omarchy-setup-security-fingerprint` hardcodes stock `fprintd`:

```bash
omarchy-pkg-add libfprint fprintd usbutils
```

On a machine already running the working stack, this aborts the whole command:

```
:: fprintd-1.94.5-2 and fprintd-clients-git-1.90.1.r2.g54e56d6-10 are in conflict.
   Remove fprintd-clients-git? [y/N]
error: failed to prepare transaction (conflicting dependencies)
```

Answering `y` is not a viable workaround: it reinstalls the `libfprint`-backed
daemon, which does not support the sensor, so the reader stops being detected at all.

## Why this is only a packaging problem

Everything in the script *after* that line works correctly on the alternate stack.
`open-fprintd` implements the same `net.reactivated.Fprint` D-Bus interface, and
`fprintd-clients-git` provides both the CLI tools and `pam_fprintd.so` at the same
path Omarchy expects:

```
$ pacman -Qo /usr/lib/security/pam_fprintd.so
/usr/lib/security/pam_fprintd.so is owned by fprintd-clients-git 1.90.1.r2.g54e56d6-10

$ fprintd-list $USER
found 1 devices
Device at /net/reactivated/Fprint/Device/0
Fingerprints for user … on DBus driver (press):
 - #0: right-index-finger
```

## Environment

- Omarchy 4.0.0-1
- Kernel 7.1.8-arch1-3
- ThinkPad T480s (20L7S2QV00)
- Sensor: `06cb:009a` Synaptics Metallica MIS Touch Fingerprint Reader
- Stack: `python-validity` 0.15-1, `open-fprintd` 0.7-2, `fprintd-clients-git` 1.90.1.r2.g54e56d6-10

alhasapi:
## Problem

`omarchy-setup-security-fingerprint` unconditionally runs
`omarchy-pkg-add libfprint fprintd usbutils`. On hardware libfprint doesn't
support (e.g. Synaptics/Validity `06cb:*` sensors on several ThinkPads),
users install a drop-in `fprintd`-compatible stack instead -- e.g.
`open-fprintd` + `python-validity` + `fprintd-clients-git`. That client
package `Conflicts With: fprintd fprintd-clients`, so `pacman -S --noconfirm`
hits a conflict prompt, defaults it to `N`, and aborts. The script runs
under `set -e`, so it dies there -- before configuring PAM for sudo, polkit,
or the lock screen. Closes #7006.

The removal script has the mirror-image bug: `omarchy-pkg-drop fprintd
libfprint libfprint-git` fails outright when `libfprint` is a dependency of
the replacement stack's own client package, since pacman refuses to remove a
package something else depends on.

Builds on #8756 (polkit `system-auth` seeding) -- this PR's diff includes
that commit until it merges.

## Fix

- Detect a working fprintd-compatible interface (the `fprintd-{enroll,
  verify,list}` CLI tools, `pam_fprintd.so`, and the D-Bus activation file)
  before installing packages. If one is already present, skip installation
  and re-enrollment entirely -- capability-based, not implementation-based,
  so behavior on a stock libfprint/fprintd machine is unchanged.
- Enable the backend's own suspend/resume systemd units when present
  (`open-fprintd-suspend`, `open-fprintd-resume`,
  `python3-validity-suspend-hotfix`) -- some backends reinitialize the reader
  on resume via a sleep hook rather than in the daemon; without this the
  sensor can come back from suspend dead and auth silently falls through to
  password.
- Leave an existing hand-edited `pam_fprintd.so` line in `/etc/pam.d/sudo`
  untouched (e.g. a custom `timeout=`) rather than treating it as
  unconfigured.
- Fix `omarchy-remove-security-fingerprint` to only drop packages when
  Omarchy's own `fprintd` is installed, leaving a replacement stack and its
  enrolled prints alone.
- Widen the `remove.security.fingerprint` menu gate so it also appears when
  a replacement stack is what's actually configured.

## Testing

Manually verified on a ThinkPad T480 (Synaptics `06cb:009a`) with
open-fprintd + python-validity + fprintd-clients-git: sudo, polkit, and
lock-screen (Quickshell) fingerprint auth all work, the clamshell gate still
skips to password with the lid closed, and auth survives a suspend/resume
cycle. Behavior on the stock libfprint/fprintd path is unchanged (existing
branches untouched, capability checks pass through as before).

Added `test/shell.d/security-fingerprint-test.sh` and
`test/shell.d/security-fingerprint-remove-test.sh`, covering the new
predicate/branch logic (`fingerprint_enrolled`,
`enable_fingerprint_sleep_units`, the removal package-drop branch, and the
widened menu gate) via PATH-stubbing, in the style of the existing shell
test suite. `pam_fprintd_present`/`fingerprint_stack_present` check fixed
absolute system paths with no override seam, so those are covered by the
manual testing above rather than the automated suite.

### DIFF
--- a/bin/omarchy-remove-security-fingerprint
+++ b/bin/omarchy-remove-security-fingerprint
@@ -33,7 +33,19 @@ echo -e "\e[32mRemoving fingerprint scanner from authentication.\n\e[0m"
 remove_pam_config
 remove_lock_fingerprint_pam
 
-echo "Removing fingerprint packages..."
-omarchy-pkg-drop fprintd libfprint libfprint-git
+# Only drop what this machine's fingerprint setup actually installed. A stack
+# the user brought themselves -- open-fprintd, or anything else answering the
+# fprintd interfaces -- is not ours to uninstall, and libfprint under it is a
+# dependency of that stack's own client package, so pacman refuses the removal
+# and this script would exit nonzero having half-finished. pacman's database
+# is already the record of who installed what; nothing needs to be written
+# down separately.
+if omarchy-pkg-present fprintd; then
+  echo "Removing fingerprint packages..."
+  omarchy-pkg-drop fprintd libfprint libfprint-git
+else
+  echo "Leaving the installed fingerprint packages alone; Omarchy didn't install them."
+  echo "Your enrolled fingerprints are untouched. Remove them with: fprintd-delete \$USER"
+fi
 
-echo -e "\e[32mFingerprint authentication has been completely removed.\e[0m"
+echo -e "\e[32mFingerprint authentication has been removed.\e[0m"
--- a/bin/omarchy-setup-security-fingerprint
+++ b/bin/omarchy-setup-security-fingerprint
@@ -6,6 +6,73 @@
 set -e
 
 
+# The stack Omarchy installs itself. Anything else providing the same
+# interfaces is somebody else's stack, and is left alone.
+fingerprint_packages=(libfprint fprintd usbutils)
+
+# Some backends reinitialise the reader across suspend with sleep hooks rather
+# than in the daemon. Without them the sensor can come back from a resume dead
+# and every later authentication falls silently through to password -- the
+# same class of failure the clamshell gate exists to avoid, minus the lid.
+# Stock fprintd ships none of these, so the loop below is a no-op there.
+fingerprint_sleep_units=(
+  open-fprintd-suspend.service
+  open-fprintd-resume.service
+  python3-validity-suspend-hotfix.service
+)
+
+# pam_fprintd.so isn't on $PATH like the client tools are, and libpam has no
+# "is this module loadable" query -- the file's presence is the only check
+# available. Glob a few known prefixes rather than hardcoding one, since a
+# stack built against a different prefix puts it somewhere else.
+pam_fprintd_present() {
+  local dir
+  for dir in /usr/lib/security /usr/lib64/security /lib/security; do
+    [[ -e "$dir/pam_fprintd.so" ]] && return 0
+  done
+  return 1
+}
+
+# Everything downstream of installing packages talks to exactly four things:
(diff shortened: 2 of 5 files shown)


CJHarmath:
Independently ran into this exact issue on a second ThinkPad T480s (`06cb:009a`) and reconstructed the same workaround by hand before finding this PR — can confirm the underlying approach is correct.

Validated on my machine: `open-fprintd` + `python-validity` + `fprintd-clients-git` do implement the same `net.reactivated.Fprint` D-Bus interface and install `pam_fprintd.so` at the path Omarchy's PAM setup expects (`/usr/lib/security/pam_fprintd.so`), so `fprintd-list`/`fprintd-enroll`/`fprintd-verify` and the sudo/polkit/lock-screen PAM stacks all work identically to the stock `fprintd` case once the backend is running. `fprintd-clients-git`'s `Conflicts: fprintd, fprintd-clients` is exactly what makes the wizard's unconditional `omarchy-pkg-add libfprint fprintd usbutils` abort the transaction — capability-detecting an existing working backend instead of hardcoding the stock package set is the right fix rather than trying to special-case specific hardware IDs.

+1 to merge from a second confirmed data point on the same hardware family.

bjarneo:
Automated duplication check: this pull request looks similar to #11840, which covers the same alternate fingerprint stack. I keep that one open and close this one to consolidate review.

PR #8761, opened 2026-08-28
Keep mailto parameters out of HEY's recipient

tony-roslund:
`omarchy-webapp-handler-hey` strips the `mailto:` scheme and assigns everything after it to HEY's `to=` parameter:

```bash
email=$(echo "$url" | sed 's/mailto://')
web_url="https://app.hey.com/messages/new?to=$email"
```

A mailto: URL carries its subject and body as query parameters after the address, so for `mailto:someone@example.com?subject=Hello` that builds:

```
https://app.hey.com/messages/new?to=someone@example.com?subject=Hello
                                                       ^ second "?"
```

The second `?` is inside the query, so `to` is the literal string `someone@example.com?subject=Hello` and `subject` never exists as a parameter. HEY opens a compose window with no recipient and no subject.

A bare `mailto:someone@example.com` works, which is why this has gone unnoticed since the handler landed — the break only shows up once a subject or body is involved, which is any prefilled mailto link on the system.

### Fix

Split the address at the first `?` and pass the remaining parameters through as parameters. They arrive already percent-encoded, so they need no further handling. The `mailto:?to=…` form is handled too; previously it produced `?to=?to=…`.

### Verification

`test/shell.d/webapp-handler-hey-test.sh` covers a subject, multiple parameters, a bare address, the `mailto:?to=` form, and a non-mailto invocation. It fails against the current handler on the first case and passes with the fix.

Also verified by hand on Omarchy 4.0.1: before, HEY opened blank; after, the recipient and subject are both filled in.

`./test/cli` passes.

### DIFF
--- a/bin/omarchy-webapp-handler-hey
+++ b/bin/omarchy-webapp-handler-hey
@@ -8,8 +8,30 @@ web_url="https://app.hey.com"
 
 # Handle mailto: URLs
 if [[ $url =~ ^mailto: ]]; then
-  email=$(echo "$url" | sed 's/mailto://')
-  web_url="https://app.hey.com/messages/new?to=$email"
+  # A mailto: URL is "mailto:addresses?subject=…&body=…". The addresses run up
+  # to the first "?", and the parameters after it are already percent-encoded,
+  # so HEY can take them as they are. Folding the whole remainder into "to="
+  # leaves a second "?" inside the query, which HEY reads as part of the
+  # address: the compose window opens with no recipient and no subject.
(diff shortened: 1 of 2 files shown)


tony-roslund:
The base branch was wrong when this opened — it targeted `master`, which is stale at 3.8.5, so the diff showed the whole 4.0 migration. It's now retargeted to `quattro` and shows the intended two files: the handler fix and its test.

getaaron (approved):

PR #8763, opened 2026-08-28
Make the number of always-visible workspaces configurable

hsyl20:
The workspaces widget kept 1-5 on the bar and revealed the rest only once they had a window. Read the count from a workspaceCount setting on the widget's shell.json entry instead, clamped to 1-10, so a bar can show the full 1-9 and 0 set at all times. Default stays 5.


bjarneo:
Automated duplication check: this pull request looks similar to #9829, which covers the same workspace count config. I keep that one open and close this one to consolidate review.

Issue #8555, opened 2026-08-27
Bar widget plugin QML changes don't hot-reload (requires full shell restart)


ramackersjp:
## Summary
Editing a bar-widget plugin's `.qml` under `~/.config/omarchy/plugins/<id>/` is detected (the watcher fires and the shell re-registers the plugin), but the widget keeps rendering the OLD compiled QML until a full `omarchy restart shell`. The saved changes never take effect without restarting the whole shell.

## Expected
Saving a file under `~/.config/omarchy/plugins/` reloads plugin code automatically (per the docs), so the bar widget reflects the change immediately.

## Actual
Each edit requires `omarchy restart shell` to become active.

## Root cause
`shell.qml` relies on `Qt.clearComponentCache` before rebuilding plugins:

```js
// shell.qml:757
if (typeof Qt.clearComponentCache === "function") Qt.clearComponentCache()
```

In the Arch `quickshell 0.3.1` package, `Qt.clearComponentCache` is **not exposed to QML**, so the guard is false and the component cache is never cleared. The plugin is re-registered, but the cached, compiled QML is re-used → old widget stays until a full restart naturally drops the cache.

## Evidence
- `qt6` `quickshell 0.3.1-1` (packager: alerque@archlinux.org)
- `strings /usr/bin/quickshell | grep clearComponentCache` -> no match
- `/usr/lib/qt6/qml/Quickshell/quickshell-core.qmltypes` contains **no** `clearComponentCache` (grep = 0 hits); the exported types (`Quickshell`, `QuickshellSettings`, ...) expose no cache-clear method
- `inotifywait` watcher IS running and fires: shell log shows `Local plugin changed, reloading: <id>` on every save
- After the reload only a full process restart applies QML content changes

## Steps to reproduce
1. Add any bar-widget plugin under `~/.config/omarchy/plugins/` (e.g. `jp.taxiprijs/BarWidget.qml`).
2. Edit a label/color in the `.qml` and save.
3. Observe the bar still shows the old content until you run `omarchy restart shell`.

## Suggested fix
Expose/clear the QML component cache so plugin content is re-read from disk on reload — e.g. surface `Qt.clearComponentCache` (or recreate the QML engine per generation, as Quickshell's own native `watchFiles`/reload path does) and call it in `finishPluginReload()`. Alternatively, enable Quickshell's native `watchFiles`-based engine recreation for plugin directories.

## System
- Omarchy: 4.0.1-1
[...]

Issue #956, opened 2025-08-21
[SOLVED] screensaver does not work, even after yay command


sliceking:
I saw some posted a command to fix the broken screensaver and while running it ` yay -S --needed python-poetry-core python-terminaltexteffects
` the build fails like so

```warning: python-poetry-core-2.1.3-1 is up to date -- skipping
 there is nothing to do
==> Making package: python-terminaltexteffects 0.12.0-1 (Thu 21 Aug 2025 09:41:28 AM CDT)
==> Checking runtime dependencies...
==> Checking buildtime dependencies...
==> Retrieving sources...
  -> Found python-terminaltexteffects-0.12.0.tar.gz
==> Validating source files with b2sums...
    python-terminaltexteffects-0.12.0.tar.gz ... Passed
==> Removing existing $srcdir/ directory...
==> Extracting sources...
  -> Extracting python-terminaltexteffects-0.12.0.tar.gz with bsdtar
==> Sources are ready.
==> Making package: python-terminaltexteffects 0.12.0-1 (Thu 21 Aug 2025 09:41:30 AM CDT)
==> Checking runtime dependencies...
==> Checking buildtime dependencies...
==> WARNING: Using existing $srcdir/ tree
==> Starting build()...
/home/stani/.local/share/mise/installs/python/3.13.7/bin/python: No module named build
==> ERROR: A failure occurred in build().
    Aborting...
 -> error making: python-terminaltexteffects-exit status 4
 -> Failed to install the following packages. Manual intervention is required:
python-terminaltexteffects - exit status 4```


Any help would be appreciated, because the screensaver looks so beautiful in the videos <3

**Comment by sliceking** [comment]
i did try doing `pip install terminaltexteffects` first, but that did not work because pip wasnt found. So I installed python/pip through the omarchy interface, then did a pip install, which did install tte, but did not fix the screensaver. Then I tried the yay command found here and that could not install.

**Comment by gigq** [comment]
FWIW I went through the same issue on an install from a couple days ago but installing via “yay python-terminaltexteffects” fixed the problem for me.

**Comment by sliceking** [comment]
it can do the `pkgver()` ok, but when it gets to `build()` .... no module named build 


```
==> Diffs to show?
==> [N]one [A]ll [Ab]ort [I]nstalled [No]tInstalled or (1 2 3, 1-3, ^4)
==> A
 -> python-terminaltexteffects: No changes -- skipping
[...]

PR #8766, opened 2026-08-28
Fix local plugin reloads

catlee:
## What changed

- Watch direct symlinked plugin directories as well as the normal plugin tree.
- Rebuild that watcher when a top-level plugin entry changes.
- Use a soft Quickshell reload for plugin rescans, so QML and imported JavaScript load from disk again.

## Why

I am working on a local plugin installed through a symlink under `~/.config/omarchy/plugins`. Editing the checkout behind that link did not trigger a reload. A manual rescan found manifest changes, but could still run an old imported JavaScript module.

The watcher now includes direct external symlink targets. Rescanning creates a fresh QML engine without unnecessarily recreating windows.

## Testing

- `bash test/shell.d/local-plugin-watcher-test.sh`
- `bash test/shell.d/plugin-registry-contract-test.sh`
- `bash test/shell.d/runtime-smoke-test.sh`
- `qmllint shell/shell.qml shell/services/PluginRegistry.qml`
- `bash test/shell` reaches an unrelated package-checkout failure because `OMARCHY_PKGS_PATH` is not configured in this worktree.

### DIFF
--- a/shell/scripts/watch-local-plugins
+++ b/shell/scripts/watch-local-plugins
@@ -2,7 +2,8 @@
 
 set -euo pipefail
 
-plugins_dir=$(realpath -e -- "$1")
+plugins_dir=$1
+canonical_plugins_dir=$(realpath -e -- "$plugins_dir")
 watch_paths=("$plugins_dir")
 
 # inotifywait does not traverse a symlink it finds below a recursive watch.
@@ -11,7 +12,7 @@ watch_paths=("$plugins_dir")
 for plugin_dir in "$plugins_dir"/*; do
   [[ -L $plugin_dir && -d $plugin_dir ]] || continue
   target=$(realpath -e -- "$plugin_dir") || continue
-  [[ $target == "$plugins_dir" || $target == "$plugins_dir"/* ]] && continue
+  [[ $target == "$canonical_plugins_dir" || $target == "$canonical_plugins_dir"/* ]] && continue
   watch_paths+=("$plugin_dir")
 done
 
--- a/shell/scripts/watch-local-plugins
+++ b/shell/scripts/watch-local-plugins
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+set -euo pipefail
+
+plugins_dir=$(realpath -e -- "$1")
+watch_paths=("$plugins_dir")
+
+# inotifywait does not traverse a symlink it finds below a recursive watch.
+# Passing a direct plugin symlink explicitly makes it report changes through
+# the logical plugin path, which the registry already knows how to route.
+for plugin_dir in "$plugins_dir"/*; do
+  [[ -L $plugin_dir && -d $plugin_dir ]] || continue
+  target=$(realpath -e -- "$plugin_dir") || continue
+  [[ $target == "$plugins_dir" || $target == "$plugins_dir"/* ]] && continue
+  watch_paths+=("$plugin_dir")
+done
+
+exec inotifywait -m -r -q \
+  -e close_write,create,delete,move \
+  --format '%w%f' \
+  "${watch_paths[@]}"
--- a/shell/services/PluginRegistry.qml
+++ b/shell/services/PluginRegistry.qml
@@ -25,6 +25,7 @@ QtObject {
   property int registryRevision: 0
   property bool scanning: false
   property string lastEnableError: ""
+  readonly property string localPluginWatcherPath: Quickshell.env("OMARCHY_PATH") + "/shell/scripts/watch-local-plugins"
 
   signal pluginsChanged()
   signal scanFinished()
@@ -634,21 +635,12 @@ QtObject {
   }
 
   property Process localPluginWatcher: Process {
-    command: [
-      "inotifywait",
-      "-m",
-      "-r",
-      "-q",
-      "-e",
-      "close_write,create,delete,move",
-      "--format",
-      "%w%f",
-      registry.pluginsDir
-    ]
+    command: [registry.localPluginWatcherPath, registry.pluginsDir]
     stdout: SplitParser {
       onRead: function(path) {
         var pluginId = registry.localPluginIdForPath(path)
         if (pluginId) registry.localPluginChanged(pluginId)
+        if (registry.localPluginRootChanged(path)) registry.restartLocalPluginWatcher()
       }
     }
     onExited: localPluginWatcherRestart.restart()
@@ -657,6 +649,11 @@ QtObject {
   property Timer localPluginWatcherRestart: Timer {
     interval: 1000
     onTriggered: localPluginWatcher.running = true
+  }
+
+  function restartLocalPluginWatcher() {
+    if (localPluginWatcher.running) localPluginWatcher.running = false
+    localPluginWatcherRestart.restart()
   }
 
   function rescan() {
@@ -710,6 +707,15 @@ QtObject {
 
     var slash = relative.indexOf("/")
     return slash === -1 ? relative : relative.slice(0, slash)
+  }
+
+  function localPluginRootChanged(filePath) {
+    var base = pluginsDir.replace(/\/$/, "") + "/"
+    var path = String(filePath || "").trim()
+    if (path.indexOf(base) !== 0) return false
+
+    var relative = path.slice(base.length)
+    return relative.length > 0 && relative.charAt(0) !== "." && relative.indexOf("/") === -1
   }
 
   Component.onCompleted: ensureUserDir()
--- a/shell/shell.qml
+++ b/shell/shell.qml
@@ -54,9 +54,7 @@ ShellRoot {
 
   property var defaultsConfig: builtinShellConfig
   property var shellConfig: builtinShellConfig
-  property bool pluginReloading: false
-  property bool pluginReloadPending: false
-
+  property var hyprlandPluginState: ({ enabled: [] })
   Timer {
     id: localPluginReloadTimer
     interval: 150
@@ -246,7 +244,7 @@ ShellRoot {
   Loader {
     id: pluginBarLoader
 
-    active: !shell.pluginReloading && shell.activeBarId !== shell.defaultBarId && shell.activeBarSourceUrl !== ""
+    active: shell.activeBarId !== shell.defaultBarId && shell.activeBarSourceUrl !== ""
     source: shell.activeBarId !== shell.defaultBarId ? shell.activeBarSourceUrl : ""
     asynchronous: true
     onLoaded: shell.configureBar(item, shell.activeBarManifest)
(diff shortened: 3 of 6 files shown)


jorgeccastro:
I reran the full smoke test on a two-monitor session and tracked the apparent `3 -> 8` IPC-handler increase. It is a test-accounting issue, not a runtime handler leak.

The warning count is cumulative across engine generations:

```text
current quattro: 1 generation, 3 warnings per target
PR #8766:        8 generations, 8 warnings per target
```

In Quickshell 0.3.1 the IPC registry is scoped to an `EngineGeneration`, and the new handlers register after the old generation is destroyed. The PR's imported-JavaScript marker and all subsequent IPC checks also confirm that calls reach the new engine.

The existing assertion only allows `(screens - 1)` warnings for the entire log. With the new soft reload coverage, it needs to account for every loaded generation:

```bash
generations=$(grep -c "Configuration Loaded" "$log" || true)
(( generations > 0 )) || generations=1
allowed_collisions=$((generations * (screens - 1)))
```

Then compare `worst` with `allowed_collisions`. I applied that test-only change locally on top of `f2b1fb5a` and removed the unused `hyprlandPluginState` property from `shell.qml`.

Results on Omarchy 4.0.1, Quickshell 0.3.1, Hyprland 0.56.2, two monitors:

```text
local-plugin-watcher-test.sh: PASS
runtime-smoke-test.sh:       PASS
```

One minor caveat remains: IPC calls can briefly return `NoCurrentGeneration` during a soft reload. The current smoke test retries through that window, so I would not block the PR on it.

I can provide the two small commits if useful.

catlee:
> I can provide the two small commits if useful.
Sure, that would be great!

`hyprlandPluginState` came from another branch I'm working on - https://github.com/basecamp/omarchy/pull/8396. It leaked into this PR.

willisiswillis:
@catlee 

Compatibility warning for the current Quickshell package:

This PR changes `reloadPlugins()` to call `Quickshell.reload(false)`, so plugin changes deliberately destroy an `EngineGeneration`. The runtime fixture also creates an `IpcHandler` and drives repeated reloads while asynchronous events remain active.

Quickshell 0.3.1-1 has a confirmed use-after-free at exactly this lifecycle boundary: [quickshell#956](https://github.com/quickshell-mirror/quickshell/issues/956). We symbolized two crashes on Omarchy 4.0.1-1 / Quickshell 0.3.1-1 / Qt 6.11.2:

```text
__dynamic_cast
IpcHandler::updateRegistration()  ipchandler.cpp:318
IpcHandler::onPostReload()         ipchandler.cpp:278
QQmlObjectCreator::finalize
QQmlComponent::createObject
...
Process::onFinished
QSocketNotifier::event
```

In both cores, `EngineGeneration::destroy()` had deleted the IPC registry extension while the generation remained discoverable. A pending `Process::onFinished` callback then completed a new `IpcHandler` and attempted to cast the freed extension. One core showed the extension’s vtable cleared; the other showed its memory already reused for UTF-16 manifest JSON.

Quickshell master now contains an apparent direct fix: [`2d3b3e9`](https://git.outfoxxed.me/quickshell/quickshell/commit/2d3b3e9c70ef380dff751b61d334dc88df016c29) moves extension destruction into the generation destructor, after JavaScript interruption and engine teardown. Arch’s current `quickshell 0.3.1-1` predates it.

This does not mean the watcher/cache design in this PR is wrong, but merging this against unpatched 0.3.1 would exercise the vulnerable generation-teardown boundary on every plugin change, increasing exposure to the timing-dependent use-after-free. Could the PR either wait until Omarchy consumes a Quickshell build containing `2d3b3e9`, or explicitly require/carry that fix?

A useful regression test would start a Process-driven `IpcHandler` construction, trigger a soft reload or IPC kill while completion is in flight, and assert a clean exit/reload. The current smoke test verifies eventual recovery but does not exercise that teardown race.

jorgeccastro:
Adding the two data points I have from the same lifecycle boundary, since my
run above drove it repeatedly.

On Omarchy 4.0.1-1 / Quickshell 0.3.1-1 / Hyprland 0.56.2, two monitors, the
smoke test loaded eight generations without a crash. IPC calls did return
`NoCurrentGeneration` for a short window during each soft reload, and the test
retries through it. So the teardown path is reachable and observably not atomic
here; my runs simply did not land on the freed extension.

My two commits on this PR are test-only accounting plus the stray property, so
they neither exercise nor avoid the race you name. A regression test of the
shape you describe would be new coverage either way.

I have not tried to reproduce quickshell#956 directly and I take no position on
whether it should gate this merge.

willisiswillis:
Thanks for the follow-up. I think we’re mostly agreeing, but I want to make the distinction clear.

 Your test shows that eight reloads can complete successfully. I believe that result. The problem is that this is a timing-dependent crash, so eight successful reloads do not tell us whether the teardown is safe—they only tell us those eight reloads missed the bad timing window.

  I went back and tested that window directly. I built QuickShell 0.3.1 twice from the same source:
  - One build was unchanged.
  - The other had only the small upstream teardown fix from [2d3b3e9](https://git.outfoxxed.me/quickshell/quickshell/commit/2d3b3e9c70ef380dff751b61d334dc88df016c29).

  Using this PR’s actual plugin watcher and real plugin-file edits, the unpatched build crashed in five out of six runs. Most crashes happened after only one to four edits. The build with the QuickShell fix completed 100 consecutive edits without crashing.

I also tested the reload operation by itself with randomized timing. The unpatched build crashed in 27 of 30 runs. The fixed build survived all 30 runs and loaded 909 generations.

 Interestingly, when I used an exact 150 ms interval every time, the unpatched build survived 400 generations. That likely explains why your eight-generation smoke test passed: repeating at a fixed interval can consistently miss a timing-sensitive race.

So I am not saying every plugin edit will crash the shell. I am saying every successful plugin edit now triggers a full QuickShell generation teardown, and on the currently packaged QuickShell 0.3.1-1, that teardown contains the use-after-free described in [quickshell#956](https://github.com/quickshell-mirror/quickshell/issues/956). Each edit gives the race another opportunity to happen.

This PR did not create the QuickShell bug. It just makes that buggy teardown path part of the normal plugin-edit workflow. My recommendation is therefore straightforward: merge or release this only after Omarchy’s QuickShell package includes  [`2d3b3e9`](https://git.outfoxxed.me/quickshell/quickshell/commit/2d3b3e9c70ef380dff751b61d334dc88df016c29). Once that fix is present, I have no objection based on this issue.


TLDR: The upstream QuickShell fix does not replace this PR—it makes the soft-reload behavior added by this PR safe to use. My only concern is release ordering: Omarchy should consume [`2d3b3e9`](https://git.outfoxxed.me/quickshell/quickshell/commit/2d3b3e9c70ef380dff751b61d334dc88df016c29) before shipping #8766. Once the packaged QuickShell contains that fix, this concern is resolved.

rishwanth-th:
Independently verified that this approach fixes the stale-QML problem, in case it helps the review.

Swapping the no-op guard for `Quickshell.reload(false)` picks up edited plugin QML **in-process**, without restarting the shell. Tested on Omarchy 4.0.1-1 / quickshell 0.3.1-1 by exposing `Quickshell.reload(false)` over IPC in a third-party bar-widget plugin, then editing that plugin's QML in place:

```
PID before: 176029   gap: 7.5
after edit (no reload yet):  gap: 7.5     <- edit alone does nothing
after Quickshell.reload(false):
PID after:  176029   gap: 5.0             <- same process, new value
```

`gap` is a geometry constant the plugin reports over IPC, changed on disk between the two reads. Same PID across the reload, so it is a new `EngineGeneration` rather than a process restart. Repeated twice with different values.

For whoever reviews: the guard being replaced can never have fired. `clearComponentCache()` is a public C++ method on `QQmlEngine` and has never been bound to the QML global `Qt` object in Qt 5 or Qt 6, so `typeof Qt.clearComponentCache === "function"` is always false. That is consistent with #6981, #8555 and #9251 all reporting the same symptom.

jorgeccastro:
Here they are: catlee/omarchy#1 (the test accounting and the stray property, on top of `f2b1fb5a`).

Rockzmash:
Adding a third option that I don't think has come up yet, since the teardown-safety
question above is currently what gates this.

**Keep the in-place reload, and give `Qt.createComponent()` a unique URL instead.**

The stale-code symptom follows from `Qt.clearComponentCache` never existing on the QML
`Qt` global (as @rishwanth-th establishes in #6981), so the guard in
`finishPluginReload()` is always false and nothing is ever evicted. The type loader then
returns the previously compiled unit for an unchanged URL. Making the URL unique per load
sidesteps the cache without touching engine lifetime:

```diff
-    var comp = Qt.createComponent(url, Component.Asynchronous)
+    var reloadUrl = url + (url.indexOf("?") === -1 ? "?" : "&") + "omarchyReload=" + Date.now()
+    var comp = Qt.createComponent(reloadUrl, Component.Asynchronous)
```

The reason this might be worth a look here: it does **not** destroy an `EngineGeneration`,
so it does not go near quickshell-mirror/quickshell#956 at all.

## What I verified

Omarchy 4.0.1-1, quickshell 0.3.1-1, qt6-declarative 6.11.2-1, Arch, kernel 7.1.9,
single monitor. Tested against a copy of the shell tree via `OMARCHY_PATH`, not the
packaged one.

With a throwaway bar-widget plugin logging a `BUILD=<n>` marker on completion, two
consecutive live edits were both picked up with no restart:

```
17:41:01  Local plugin changed, reloading: zz.reloadprobe
17:41:01  loadPluginWidget ... BarWidget.qml?omarchyReload=1788212461069
17:41:01  RELOADPROBE BUILD=D
17:41:18  RELOADPROBE BUILD=E
```

Unpatched, the same edit re-instantiates the widget and re-runs the old marker.

I also instrumented `syncPluginWidgets()` to check the bookkeeping, since @battisti
raised the reuse branch above: during a reload the `existing && existing.url === url`
branch is **never** taken, because `unloadPluginWidgets()` has already emptied
`pluginWidgetComponents`. `loadPluginWidget()` is called every time, with the right URL.
So on this path the reuse branch is not a compounding cause — the cache is.

A third-party plugin using a relative `.js` import and a sibling QML type kept loading
correctly with the query string present, which was my main worry about this approach.

## Limits, stated plainly

- **I only patched and tested the bar-widget site.** Panels load via
  `Loader { source: sourceUrl }` and services via
  `Qt.createComponent(url, Component.PreferSynchronous)`; both would need the same
  treatment, so this is three call sites, not one. `Quickshell.reload(false)` covers
  all of them at once, which is a real advantage of this PR's approach.
- **Compilation units accumulate**, one per reload per plugin, for the life of the
  process. Fine for an edit loop; it is a trade, not a clean win.
- **I have not tried to reproduce quickshell#956**, and I take no position on whether
  it should gate this merge.

Not suggesting this should replace the approach here — posting it as a data point in
case a change that avoids the generation boundary is useful while #956 is open.

<sub>Measurements run by an AI agent on my machine; I reviewed the results and the
claims above are limited to what was actually observed.</sub>

PR #8770, opened 2026-08-28
Report the machine's battery figures, not the first cell's

### LINKED ISSUE #7214: omarchy-battery-status reads only the first battery on dual-battery laptops
## What happened

On a dual-battery laptop, the `omarchy.power` bar popup reports battery stats for only the *internal* battery (BAT0), not the combined system battery. The popup shows ~91% when the machine is really at ~42% and discharging, "Battery size" shows 14 Wh instead of 75 Wh, and "Time left" is blank.

The bar icon itself uses `UPower.displayDevice` (combined), so the icon fill and the popup percentage disagree.

## Root cause

`omarchy-battery-status` selects the first `BAT*` device from `upower -e`:

```bash
battery=$(upower -e 2>/dev/null | grep BAT | head -n 1)   # /usr/bin/omarchy-battery-status:21
```

On a dual-battery machine this always resolves to BAT0 (internal), ignoring the external battery. The composite `DisplayDevice` is the correct source.

## What was expected

The popup should report the combined battery state (UPower `DisplayDevice`), matching the icon's fill level.

## Steps to reproduce

1. On a laptop with two batteries (internal + external/swappable), open the power popup from the bar.
2. Compare the popup percentage to `upower -i /org/freedesktop/UPower/devices/DisplayDevice`.

## Evidence

Lenovo ThinkPad T480 (internal + swappable battery):

```
$ upower -e
/org/freedesktop/UPower/devices/battery_BAT1
...
/org/freedesktop/UPower/devices/DisplayDevice

BAT0 (internal, LGC 01AV489): fully-charged, 91%, energy-full 14.27 Wh, 322 cycles
BAT1 (external, SANYO 01AV427): discharging, 30%, energy-full 61.11 Wh, 111 cycles
DisplayDevice (combined):       discharging, 41.9%, energy-full 75.38 Wh, 1.7h to empty
```

The popup currently shows BAT0's values (91% / 14 Wh / blank time-left). It should show ~42% / 75 Wh / ~1h 42m.

## Suggested fix

Prefer `DisplayDevice` and fall back to the first `BAT*` only when it is absent:

```bash
battery=$(upower -e 2>/dev/null | grep DisplayDevice | head -n 1)
[[ -z $battery ]] && battery=$(upower -e 2>/dev/null | grep BAT | head -n 1)
```

Caveat: `DisplayDevice` has no `native-path`, `charge-cycles`, or per-device `charge-start/end-threshold` fields, so the sysfs `power_now` shortcut and the cycle/threshold lookups must keep falling back to the `BAT*` globs (the existing `[[ -z ... ]]` fallback paths already handle this).

## System

- Omarchy 4.0.0-1
- Lenovo ThinkPad T480 (dual battery)

alhasapi:
`omarchy-battery-status` reads a single device:

```bash
battery=$(upower -e | grep BAT | head -n 1)
```

Everything downstream of that line describes one cell while claiming to describe the laptop. On my T480 with the internal pack at 21% and the external at 95%, it reports **21% and 38m left** for a machine that actually holds **53% and 2h 42m**.

### This reaches the Quattro shell

`shell/plugins/panels/power/Panel.qml` renders the hero percentage from this script, while the progress bar *directly beneath it* and the bar widget itself both use `UPower.displayDevice`:

| line | value | source |
|---|---|---|
| `:281` | bar widget percentage | `Math.round(root.batteryFraction * 100)` → `displayDevice` |
| `:372` | panel hero percentage | `root.batteryInfo.percentage` → **this script** |
| `:403` | progress bar fill | `root.batteryFraction` → `displayDevice` |

So on a dual-battery machine the bar widget reads 53%, and the panel it opens reads 21% — printed directly above a bar drawn at half full.

That is the same inconsistency #5780 reported against Waybar ("inconsistent battery reading between what is shown in the waybar and the notification"), now inside the native shell.

### What changed

**Percentage, state, capacity and remaining time now come from `DisplayDevice`**, which is the aggregate and mirrors the single cell on a one-battery machine.

**The instantaneous sysfs rate is summed across packs.** `DisplayDevice` carries no `native-path` to read it from, and the reading has to cover every pack in any case: a ThinkPad drains its external battery first, leaving the internal one at `current_now=0`, so a rate read from one cell reports `0W` while the laptop is visibly discharging. That is the second symptom reported in #7214.

**Charge thresholds deliberately stay on a battery.** They are per-cell configuration, `DisplayDevice` does not carry them, and sourcing them from it would silently demote every machine to the `BAT*` sysfs fallback — which does not always agree. This ThinkPad reports an 80% end threshold through UPower and 100% through sysfs.

**Charge cycles are deliberately untouched.** They have no aggregate, so reporting them honestly needs somewhere per-battery to put them — a separate change, not this one.

### Testing

`test/shell.d/battery-status-test.sh` gains a dual-battery case: two stub packs with `BAT0` at `current_now=0`, asserting the aggregate percentage, combined capacity, aggregate runtime, and a summed non-zero draw. It fails against the current script with `not ok - battery status reports the machine percentage, not the first cell` and passes with this change. The existing single-battery assertions, including the `10.8W` sysfs rate, are unchanged and still pass.

Verified on a ThinkPad T480 with both packs installed and with only the internal pack:

```
$ bin/omarchy-battery-status --shell
percentage  47%          $ upower -i .../DisplayDevice
state       discharging    state:        discharging
rate        4.2W           energy-rate:  3.979 W
size        23Wh           time to empty: 2.8 hours
time        2h 36m         percentage:   47.615%
```

### Credits

Reported in #5780 by @Michael-Steshenko and again in #7214 by @syntaxboybe, with a second T480 confirmation from @raplo-z. #5958 by @stefanomainardi was an earlier fix for the pre-Quattro surfaces.

Closes #7214.

### DIFF
--- a/bin/omarchy-battery-status
+++ b/bin/omarchy-battery-status
@@ -18,10 +18,21 @@ case "${1:-}" in
     ;;
 esac
 
-battery=$(upower -e 2>/dev/null | grep BAT | head -n 1)
-[[ -z $battery ]] && exit 0
+# Charge, runtime and draw are properties of the machine, not of whichever cell
+# the kernel enumerated first. UPower's DisplayDevice is that aggregate: on a
+# single-battery machine it mirrors the one cell, and on a dual-battery ThinkPad
+# it is the difference between reporting 21% with 38m left and the 53% with
+# 2h 42m the laptop actually has.
+batteries=()
+while IFS= read -r device; do
+  batteries+=("$device")
+done < <(upower -e 2>/dev/null | grep BAT)
 
-battery_info=$(upower -i "$battery")
+(( ${#batteries[@]} )) || exit 0
+
+first_battery_info=$(upower -i "${batteries[0]}")
+battery_info=$(upower -i /org/freedesktop/UPower/devices/DisplayDevice 2>/dev/null)
+[[ -n $battery_info ]] || battery_info=$first_battery_info
 
 percentage=$(awk '/percentage/ { print int($2); exit }' <<<"$battery_info")
 capacity=$(awk '/energy-full:/ { printf "%d", $2; exit }' <<<"$battery_info")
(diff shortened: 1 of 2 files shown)


alhasapi:
Closing in favour of #8864, which does the same thing and does several of them better.

I opened this on the 28th and @zmitton opened #8864 the next day. Two near-identical PRs sitting in a queue this size helps nobody, and having read theirs line by line against mine, theirs is the one to take:

- **Same core fix** — read `DisplayDevice` for percentage, state, capacity and time.
- **Same handling of the trap underneath it.** `DisplayDevice` has no `native-path`, so the live sysfs rate has to be summed across every `BAT*` rather than read from one device. #8864 globs `BAT*` under the power supply path directly, which is simpler than my resolving each `native-path` back through `upower -i`.
- **Same reasoning on charge thresholds**, which have to keep coming from an individual device. That one bit me while writing this PR and it is the easiest part of the fix to get wrong: moving thresholds to the aggregate silently demotes every machine to the sysfs fallback, and the two do not always agree — this T480 reports an 80% end threshold through UPower and 100% through sysfs.
- **#8864 also takes the max cycle count** rather than the first, which mine left alone.
- **Its dual-battery fixture is stronger than mine.** BAT0 is `not-charging` while the aggregate is `discharging`, so the state assertion actually pins the behaviour. Copilot flagged precisely that gap in my fixture, where both were `discharging` and the assertion would have passed even if state still came from the first cell.

One line in #8864 has no equivalent in mine, so I checked it against hardware before closing: the `present: yes` gate on `DisplayDevice`. It holds — my T480 reports `present: yes` on the aggregate with both packs installed, so the early exit does not misfire on multi-battery machines.

The piece neither PR covers is per-battery detail. Cycle count, design capacity and health belong to a physical cell and have no aggregate: taking the max is the right call for one machine-level number, but a pack at 965 cycles sitting next to one at 156 has no meaningful cycle count of its own. That is #8774, which is independent of this and does not touch `omarchy-battery-status`, so it applies cleanly on top of #8864.

Verified throughout on a dual-battery ThinkPad T480 with a Power Bridge bay.

PR #8772, opened 2026-08-28
Mock omarchy-osd in the system power test

alhasapi:
The test runs the real `bin/omarchy-system-reboot` and `bin/omarchy-system-shutdown`, mocking the commands they call so nothing actually happens. `omarchy-osd` was left out of that list:

```bash
for command in omarchy-state omarchy-hyprland-window-close-all sleep; do
```

but `bin/omarchy-system-shutdown:11` calls it:

```bash
omarchy-osd -i shutdown -m "Shutting down" -d 5000
```

So running the suite paints **"Rebooting"** and then **"Shutting down"** across the screen of whoever is running it, five seconds each. I hit this on my own machine and went looking for a scheduled poweroff before I found the test.

Nothing is actually at risk — `systemd-run` is mocked, so no poweroff is ever scheduled, and `omarchy-hyprland-window-close-all` is mocked too. But a test that tells you your machine is shutting down is its own kind of failure.

This mocks it like the rest, and asserts the call while here: the icon and message the OSD is asked for are part of what these commands do, and nothing covered them before.

All four assertions pass, and the suite no longer touches the screen.

### DIFF
--- a/test/shell.d/system-power-test.sh
+++ b/test/shell.d/system-power-test.sh
@@ -17,7 +17,10 @@ printf 'systemd-run %s\n' "$*" >>"$CALL_LOG"
 exit 0
 SH
 
-for command in omarchy-state omarchy-hyprland-window-close-all sleep; do
+# omarchy-osd belongs in this list: without it the test calls the real one and
+# paints "Rebooting", then "Shutting down", across the screen of whoever is
+# running the suite.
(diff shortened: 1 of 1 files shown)


bjarneo:
Automated duplication check: this pull request looks similar to #7216, which covers the same power test OSD mock. I keep that one open and close this one to consolidate review.

PR #8774, opened 2026-08-28
Give each battery its own section in the power panel

alhasapi:
Charge cycles, design capacity and health are facts about a physical cell. They have no aggregate — a machine with a 965-cycle pack and a 156-cycle pack has no cycle count of its own.

The panel reports the first cell's numbers under a machine-wide label. On my T480 it shows **965 cycles and 57% health** for a laptop whose other pack is at **156 and 42%**. Swap which pack the kernel enumerates first and the same machine reports the other set. Neither is wrong about a battery; both are wrong about the laptop.

### What this does

Renders one section per battery below the machine-wide summary, each with its own charge level, cycles, capacity, health and flow. The hero and the stats row keep describing the machine; the sections describe the cells.

**On a single-battery machine nothing changes.** The sections only appear when there is more than one cell to tell apart, and `omarchy-battery-details` exits 0 with no output when there is nothing to say.

### `bin/omarchy-battery-details`

UPower's QML API exposes neither charge cycles nor charge thresholds, so the panel cannot reach them through `UPowerDevice`. This supplies those, plus a vendor and model so a cell is identifiable, and falls back to `/sys/class/power_supply` where the firmware reports nothing to UPower — common on older machines. Live values still come from UPower; this only covers the gaps.

Cells are selected on UPower's **power-supply flag** rather than the `battery_` D-Bus prefix. Peripherals enumerate under that same prefix, so the prefix alone is not a filter — without this, a paired Bluetooth mouse takes an ordinal and turns "Battery" into "Battery 1". `Model.isPhysicalBattery` applies the same test on the QML side, so the two agree on what counts.

Two-battery ThinkPads are named **Internal** and **External** from DMI, since Power Bridge always enumerates the built-in cell first. That is the only layout nameable with confidence; every other machine gets ordinals.

### Testing

`test/shell.d/battery-details-test.sh` is new and covers four machines with a stubbed `upower`: a Power Bridge ThinkPad with a Bluetooth mouse attached, a three-battery non-ThinkPad, a single-battery laptop with a paired keyboard, and a desktop with only a mouse. Removing the power-supply filter fails it with `not ok - power bridge names the built-in cell`, because the mouse inflates the count to three and the naming no longer applies.

Verified on a ThinkPad T480 running the exact files in this PR — script resolved through `PATH`, no QML errors in the journal:

Panel showing INTERNAL and EXTERNAL sections with separate cycles, capacity and health

<sub>Same sections, captured from this code running as a local plugin so the machine-wide figures above them are already correct; in this PR alone the hero still comes from <code>omarchy-battery-status</code> until #8770 lands.</sub>

### Relationship to #8770

Independent — no overlapping files — but they are two halves of the same problem, and this one reads better on top of #8770. That PR fixes the machine-wide figures (the hero percentage and runtime, which today come from the first cell too); this one gives the per-cell facts somewhere honest to live. #8770 deliberately left `cycles` alone because there was nowhere to put it; this is that somewhere.

### DIFF
--- a/bin/omarchy-battery-details
+++ b/bin/omarchy-battery-details
@@ -1,0 +1,95 @@
+#!/bin/bash
+
+# omarchy:summary=Returns per-battery cycles, capacity, health and charge thresholds
+
+# omarchy-battery-status describes the machine. This describes each cell.
+#
+# Charge cycles, design capacity and health belong to a physical battery and
+# have no aggregate — a machine with a 965-cycle pack and a 156-cycle pack has
+# no meaningful cycle count of its own. So this emits one record per battery
+# and lets the panel render each on its own.
+#
+# UPower drives the live values in the panel; this covers the fields
+# Quickshell's UPower QML API does not expose (charge cycles, thresholds) plus
+# a human name for the cell.
+#
+# Output: <native-path>\t<key>\t<value>, one field per line.
+
+power_supply_path="${OMARCHY_POWER_SUPPLY_PATH:-/sys/class/power_supply}"
+dmi_path="${OMARCHY_DMI_PATH:-/sys/class/dmi/id}"
+
+emit() { printf '%s\t%s\t%s\n' "$1" "$2" "$3"; }
+
+# Peripherals — Bluetooth mice, keyboards, headsets — enumerate under the same
+# battery_ prefix as the machine's own cells, so the prefix alone is not a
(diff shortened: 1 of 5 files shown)


bjarneo:
Automated duplication check: this pull request looks similar to #8904, which covers the same per-battery power sections. I keep that one open and close this one to consolidate review.

PR #8784, opened 2026-08-28
Add theme support for T3 Code

ryanrhughes:
Makes T3 Code follow the active theme like all other applications.

https://github.com/user-attachments/assets/8bdf626d-484a-499c-be52-5675d0b366fa

**NOTE: This is reliant on an upstream feature in T3 Code being merged. https://github.com/pingdotgg/t3code/pull/8569**

For now, we have `t3code-patched-bin` that can be installed from OPR for testing or anyone who just wants the feature early.

### DIFF
--- a/bin/omarchy-install-ai-t3-code
+++ b/bin/omarchy-install-ai-t3-code
@@ -1,0 +1,49 @@
+#!/bin/bash
+
+# omarchy:summary=Install T3 Code and point it at the Omarchy palette
+# omarchy:requires-sudo=true
+
+set -e
+
+T3CODE_HOME="${T3CODE_HOME:-$HOME/.t3}"
+T3CODE_STATE_DIR="$T3CODE_HOME/userdata"
+T3CODE_SETTINGS_PATH="$T3CODE_STATE_DIR/settings.json"
+
+# Ask T3 Code to open on the palette this machine publishes. `t3 theme set`
+# is the supported way, but the desktop package ships only the GUI -- the CLI
+# is a separate npm distribution -- so fall back to the setting it writes.
+# Both run before the first launch, so neither races a running server.
+set_default_theme() {
+  if command -v t3 >/dev/null 2>&1; then
+    t3 theme set omarchy --base-dir "$T3CODE_HOME" >/dev/null && return
+  fi
+
+  local tmp
+  mkdir -p "$T3CODE_STATE_DIR"
+  [[ -f $T3CODE_SETTINGS_PATH ]] || echo '{}' >"$T3CODE_SETTINGS_PATH"
+
+  tmp=$(mktemp "$T3CODE_SETTINGS_PATH.XXXXXX")
+  if jq '.defaultTheme = "omarchy"' "$T3CODE_SETTINGS_PATH" >"$tmp"; then
+    mv "$tmp" "$T3CODE_SETTINGS_PATH"
+  else
+    rm -f "$tmp"
+    echo "Could not set the default theme in $T3CODE_SETTINGS_PATH." >&2
+  fi
+}
+
+echo "Installing T3 Code..."
+omarchy-pkg-add t3code-bin
+
+# The palette itself, and the request to open on it, are both first-run
+# concerns. `defaultTheme` only decides what a client with no theme of its own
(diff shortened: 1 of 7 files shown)


coygeek:
The corresponding PR was merged in T3code.

spencerbull:
Reviewed head `0b40fa56e14ad251d5b5fcbcfc71a7458d358b36` against `quattro`. One defect needs the author; the rest checked out on a worker.

**Medium — `bin/omarchy-install-ai-t3-code:40-43`: a user who updates and installs T3 Code before their next theme change gets no palette, and the installer says otherwise.** `current/theme/t3code.json` only exists once the current theme has been re-staged, and `omarchy-update` on quattro does not re-stage themes; the one migration that did, `1787481315.sh`, predates this branch and is already marked done for everyone current. In that state `omarchy-theme-set-t3code` exits 0 at line 15 without publishing, `t3 theme set omarchy` fails with `ThemeIdUnknownError: No theme named "omarchy"` on stdout, which line 18 sends to `/dev/null`, and the jq fallback writes `{"defaultTheme": "omarchy"}` with no `defaultThemeSetAt`. The terminal shows "Matching T3 Code to the current theme..." and "T3 Code has been installed.", T3 Code opens in its stock look, and `t3 theme show` reports "Published themes: none". Reproduced on a disposable worker built from the 4.0.1 ISO, whose staged theme predates the template exactly as an updated install's does. It heals itself on the next theme switch, because the T3 client waits for an id that does not resolve yet, so nobody is stuck, but the installer has claimed a match it did not make. `bin/omarchy-install-editor-helix` handles the same case: when the rendered file is missing it runs `omarchy-theme-refresh` first. A migration calling `omarchy-theme-refresh`, as `migrations/1779242346.sh` did for Pi, would close it as well.

**Low — `bin/omarchy-install-ai-t3-code:12-14`: the fallback's premise is stale.** `t3code-bin 0.0.38-1` in the Omarchy repository owns `/usr/bin/t3`, with `theme set`, `clear` and `show` in it, so after `omarchy-pkg-add t3code-bin` the `omarchy-cmd-present t3` branch always runs and the jq path is only reached when `t3 theme set` itself fails, which today means the case above. Either drop it or have the comment say what it now covers. The PR body's dependency note and `t3code-patched-bin` are stale for the same reason: pingdotgg/t3code#8569 merged on 2026-08-29 and v0.0.37 onward carry it.

**Checked on a worker with this branch as `$OMARCHY_PATH`.** `./test/cli` passes and `test/shell.d/theme-staging-test.sh` passes 8 of 8. All 22 shipped themes render `t3code.json` as valid JSON with no unresolved placeholder, `omarchy-theme-set-t3code` publishes each one, and `t3 theme show` on T3 Code 0.0.38 lists every one. T3 Code running on the worker's compositor retinted live from White to Tokyo Night on `omarchy-theme-set` with no restart. A theme with no `accent` leaves `{{ accent }}` and three `mix` tokens unresolved, the guard at line 20 skips with its message, and the previous palette stays published, as the comment says. The installer ran end to end on the worker in both the upgrade state above and a fresh one.

**Second opinion.** Codex GPT-5.6 at xhigh reviewed the head read-only and reached the upgrade-path finding on its own; independence is not guaranteed on this machine. Its other candidates, checked against the tree: a hand-written `themes/omarchy.json` is overwritten — rejected, T3 Code documents that directory as the machine's publishing directory and `omarchy-theme-set-claude` and `-pi` make the same trade; ANSI-only legacy themes never get a T3 palette because `accent` is not derived from `color4` — true, and equally true of `claude.json.tpl` and `pi.json.tpl`, so the guard's keep-the-old-palette outcome is the right one here rather than a defect; rerunning the installer re-applies the Omarchy theme over a later in-app pick — true, but only by invoking the installer directly, since the menu disables it once `t3code-bin` is present; `omarchy-remove-ai-t3-code` ignores a custom `T3CODE_HOME` — pre-existing and untouched by this PR.

Nothing was pushed. Waiting on the author for the upgrade path; then this is the maintainer's to land.

🤖 Generated by Claude Fable 5.1 in Claude Code. Reviewed by Codex GPT-5.6 XHigh.

PR #8791, opened 2026-08-28
fix(update): guard against missing TMPDIR and support native ChatGPT …

anmolsharma152:
…desktop binding

- Fall back to standard /tmp when TMPDIR is unset or invalid in omarchy-update-system-pkgs
- Dynamically detect and bind native ChatGPT desktop app if present, falling back to webapp


anmolsharma152:
Addressed in commit aa963487: added comprehensive test coverage in `test/shell.d/hyprland-default-config-test.sh` verifying both the native desktop execution path (with a mock `chatgpt` executable in `$PATH`) and the webapp fallback when the binary is absent.

PR #8792, opened 2026-08-28
Fix webcam MJPEG negotiation for screen recordings

### LINKED ISSUE #7300: screenrecord --with-webcam: overlay locks to YUYV @5fps instead of MJPG @30fps at 1280x720
### System details

AMD Ryzen AI 7 350 w/ Radeon 860M / AMD Radeon 860M Graphics
Omarchy 4.0.0-1
Webcam: `USB2.0 FHD UVC WebCam` (uvcvideo driver)

### What's wrong?

`omarchy screenrecord --with-webcam` starts the webcam overlay via `mpv` with:

```
--demuxer-lavf-o=video_size=1280x720,framerate=30
```

No `input_format` is specified, so ffmpeg's v4l2 demuxer is left to choose the pixel format on its own. On this webcam it picks **YUYV** (raw), which only supports 1280x720 at **5 fps** on this device — even though the same resolution is available at **30 fps in MJPG**. The result is a visibly choppy webcam overlay in every recording, regardless of the `--webcam-size` chosen.

Confirmed by querying the device directly while the overlay's `mpv` process has it open:

```
$ v4l2-ctl -d /dev/video0 --get-fmt-video
Pixel Format      : 'YUYV' (YUYV 4:2:2)
$ v4l2-ctl -d /dev/video0 --get-parm
Frames per second: 5.000 (5/1)
```

Device's own format list (`v4l2-ctl --list-formats-ext -d /dev/video0`) shows MJPG at 1280x720 supports 30fps, while the YUYV variant at the same resolution caps at 5fps:

```
[0] (MJPG-type, compressed):
  1280x720 @ 30fps
[1] 'YUYV' (YUYV 4:2:2):
  1280x720 @ 5fps   <-- what actually gets selected
```

### What's expected

The overlay should request the compressed/MJPG format explicitly so it gets the 30fps mode the hardware actually supports at that resolution, e.g. adding `input_format=mjpeg` to the `--demuxer-lavf-o` string in `omarchy-capture-screenrecording`'s `start_webcam_overlay()`:

```
--demuxer-lavf-o=video_size=1280x720,framerate=30,input_format=mjpeg
```

I tested this manually — killing the running overlay `mpv` process and relaunching it with `input_format=mjpeg` appended immediately fixes it: the device switches to MJPG and reports a real 30fps, and the overlay is visibly smooth instead of choppy.

### Steps to reproduce

1. `omarchy screenrecord --fullscreen --with-webcam` (webcam with a UVC device that has a raw/YUYV mode capped at a low fps for 720p/1080p, alongside an MJPG mode that supports 30fps at the same resolution — common on USB UVC webcams)
2. Watch the webcam overlay bubble — it's visibly laggy/low framerate
3. Confirm with `v4l2-ctl -d /dev/videoN --get-parm` while recording — shows 5fps instead of the requested 30fps

### Suggested fix

In `omarchy-capture-screenrecording`, add `input_format=mjpeg` to the `capture_options` built in `start_webcam_overlay()`, or probe `v4l2-ctl --list-formats-ext` for the pixel format that actually delivers the target fps at the chosen resolution before launching `mpv`.

ottosilva:
Fixes #7300

## Summary

- prefer MJPEG only when the selected webcam resolution supports the 30 fps target
- preserve V4L2 format autodetection when MJPEG is unsupported or capability probing fails
- require exact FourCC and resolution matches to avoid forcing unsupported capture tuples

## Test plan

- [x] `bash -n bin/omarchy-capture-screenrecording test/shell.d/screenrecording-test.sh`
- [x] `git diff --check`
- [x] `bash test/shell.d/screenrecording-test.sh` (30 consecutive runs)
- [ ] Hardware recording verification (no accessible webcam device in the current environment)

`./test/all` reached and passed the screen-recording tests. Seven unrelated test files failed because this checkout lacks external fixtures/services or cannot write to system/user paths in the restricted environment.

### DIFF
--- a/bin/omarchy-capture-screenrecording
+++ b/bin/omarchy-capture-screenrecording
@@ -59,6 +59,39 @@ small | medium | large) ;;
   ;;
 esac
 
+webcam_mjpeg_supports_framerate() {
+  local resolution=$1
+  local target_fps=$2
+
+  awk -v resolution="$resolution" -v target_fps="$target_fps" '
+    /^[[:space:]]*\[[0-9]+\]:/ {
+      mjpeg_format = ($2 == "\047MJPG\047")
+      matching_resolution = 0
+      next
+    }
+
+    mjpeg_format && $1 == "Size:" {
+      matching_resolution = ($2 == "Discrete" && $3 == resolution)
+      next
+    }
+
+    mjpeg_format && matching_resolution && $1 == "Interval:" && $2 == "Discrete" {
+      fps = $0
+      sub(/^.*\(/, "", fps)
+      sub(/[[:space:]]+fps\).*$/, "", fps)
+
+      difference = (fps + 0) - target_fps
+      if (difference < 0) difference = -difference
+      if (difference < 0.05) {
+        found = 1
+        exit
+      }
+    }
+
+    END { exit !found }
+  '
+}
+
 start_webcam_overlay() {
   cleanup_webcam
 
@@ -73,12 +106,16 @@ start_webcam_overlay() {
 
   # Try preferred 16:9 resolutions in order, use first available
   local preferred_resolutions=("640x360" "1280x720" "1920x1080")
-  local capture_options="framerate=30"
+  local target_fps=30
+  local capture_options="framerate=$target_fps"
   local available_formats=$(v4l2-ctl --list-formats-ext -d "$WEBCAM_DEVICE" 2>/dev/null)
 
   for resolution in "${preferred_resolutions[@]}"; do
     if echo "$available_formats" | grep -q "$resolution"; then
       capture_options="video_size=$resolution,$capture_options"
+      if webcam_mjpeg_supports_framerate "$resolution" "$target_fps" <<<"$available_formats"; then
+        capture_options+=",input_format=mjpeg"
+      fi
       break
     fi
   done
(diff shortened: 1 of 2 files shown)


btsouth:
Tested this branch's negotiation logic on real hardware and it fixes the problem.

Lenovo Performance Camera (uvcvideo) on 4.0.0.r1849. Same shape as the issue: MJPG offers 1280x720 at 30fps, YUYV caps 720p at 10fps, and this camera has no 640x360 mode at all, so the resolution loop lands on 1280x720. Checked with `v4l2-ctl --get-fmt-video` and `--get-parm` while the overlay had the device open:

```
stock:       YUYV 1280x720 at 10fps
this branch: MJPG 1280x720 at 30fps
```

Also ran a full recording with the patched negotiation. The overlay is smooth and the output is clean 60fps CFR with audio in sync. One note on the parser: the tests stub v4l2-ctl output with space indentation, but I ran the awk against the real tab-indented output from this camera and it matched fine.

ahmadtv:
Hardware data point from an iMac18,3 (FaceTime HD camera, `uvcvideo`), on Omarchy 4.0.3 without this PR: while the webcam overlay runs, `v4l2-ctl --get-fmt-video` / `--get-parm` report `YUYV` 1280x720 at 10.000 fps. This camera's raw mode tops out at 10 fps at 720p. With `input_format=mjpeg` added to the overlay's demuxer options, the same camera runs `MJPG` at 29.970 fps.

It shows in the recording too: with the 10 fps YUYV overlay up, gpu-screen-recorder dropped to 55 fps here, so the video played back fast (`-fm cfr` stamps every frame 1/60 s apart). With MJPEG it stayed at 58.

I haven't run this branch itself yet; happy to, if a hardware check helps it land. (#9306 covers the same ground.)

PR #8798, opened 2026-08-28
Fix mise wrapper binary resolution

Zeus-Deus:
## Problem
Generated mise wrappers can resolve themselves or a same-named binary from another package.

## Fix
Resolve executables only from the requested package’s `mise bin-paths`, then launch through `mise x`. This keeps lazy updates and runtime setup unchanged, with regression and migration coverage.

Builds on #8692.

### DIFF
--- a/bin/omarchy-mise-install
+++ b/bin/omarchy-mise-install
@@ -23,7 +23,8 @@ cat >"$HOME/.local/bin/$command" <<EOF
 #!/bin/bash
 export MISE_MINIMUM_RELEASE_AGE=0
 mise use -g --quiet "$package" || exit 1
-exec mise x "$package" -- "$bin" "\$@"
+bin_path=\$(mise which "$bin" --tool="$package") || exit 1
+exec mise x "$package" -- "\$bin_path" "\$@"
 EOF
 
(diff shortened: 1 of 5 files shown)


imadtg:
This PR does add handling of ".", ".." and absolute paths being passed to omarchy-mise-install better than it's base PR https://github.com/basecamp/omarchy/pull/8692, but i see you have opted to change from the simple `mise which "$bin" --tool="$package"` into a more elaborate version, Why is that? do you have scenarios where yours works and the base doesn't?

imadtg (approved):
after looking into mise’s behavior, I understand that `mise which --tool` is not a strict package-only lookup. The `bin-paths` approach explicitly keeps the executable inside the requested package, so this is meaningful hardening rather than an unnecessary rewrite.

Could you also restore `Fixes #7360` and `Fixes #7234` to this PR’s description?

imadtg:
This looks to be having merge conflicts and abandoned.. so i pulled these improvements back in #8692

bjarneo:
Automated duplication check: this pull request looks similar to #8692, which covers the same mise wrapper binary resolution. I keep that one open and close this one to consolidate review.

PR #8809, opened 2026-08-28
Ask before normalizing screen recording audio that's unusually quiet or loud

ogarza:
## Summary

Discussion: https://github.com/basecamp/omarchy/discussions/9020

Screen recordings no longer always run `loudnorm` on stop. That was crushing quiet sounds and dynamic range.

After stop, short recordings are measured. If the audio is unusually quiet (more than 4 LU below -14 LUFS) or unusually loud (more than 3 LU above), a menu asks whether to normalize to typical broadcast levels or keep the recording as-is. Esc keeps the original. Audio already in that window is left alone.

Recordings longer than 10 minutes skip that analysis and ask immediately — measuring a long soundtrack left a dead screen after the bar indicator was already gone. The prompt is “Normalize audio to typical broadcast levels?” with Normalize (Recommended) first. Saying yes does a single `loudnorm` pass in the remux, same as `OMARCHY_SCREENRECORD_NORMALIZE=true`. Short clips that accept still get the two-pass path with the first-pass stats.

The PipeWire capture-pop mute still always runs. `OMARCHY_SCREENRECORD_NORMALIZE=true` always normalizes; `false` never asks.

screenshot-2026-08-31_02-29-04

## Test plan
- [x] Record with desktop audio that is clearly quiet; stopping should ask “Audio may be unintentionally quiet” with Keep original / Normalize
- [x] Record with hot audio (well above broadcast level); stopping should ask “Audio may be unintentionally loud”
- [x] Esc or Keep original leaves levels alone; Normalize raises or lowers toward typical broadcast levels
- [x] A recording already near normal speech volume should not prompt
- [x] A recording longer than 10 minutes asks immediately (no analysis wait), with Normalize recommended; yes is a single loudnorm pass
- [x] `OMARCHY_SCREENRECORD_NORMALIZE=false` never asks
- [x] `bash test/shell.d/screenrecording-test.sh`

### DIFF
--- a/bin/omarchy-capture-screenrecording
+++ b/bin/omarchy-capture-screenrecording
@@ -270,8 +270,6 @@ measure_loudness() {
 prompt_normalize_audio() {
   local input_i=$1
   local prompt keep normalize selection
-  # glyph<TAB>label<TAB>subtext — a leading tab is an empty glyph; without it
-  # the label is drawn as a huge icon and overlaps the measured LUFS.
   keep=$'\tKeep original levels\tas recorded'
   if awk -v i="$input_i" -v target="$LOUDNORM_I" 'BEGIN { exit (i < target) ? 0 : 1 }'; then
     prompt="Audio may be unintentionally quiet"
--- a/bin/omarchy-capture-screenrecording
+++ b/bin/omarchy-capture-screenrecording
@@ -352,6 +352,17 @@ start_screenrecording() {
   fi
 }
 
+# Drop the sidecar only if it still names this recording. A stop that sits in
+# finalize (measure + prompt) can outlive a new start writing a different path.
+release_recording_file() {
+  local filename=$1
+  local current
+  current=$(cat "$RECORDING_FILE" 2>/dev/null) || true
+  if [[ $current == "$filename" ]]; then
+    rm -f "$RECORDING_FILE"
+  fi
+}
+
 stop_screenrecording() {
   pkill -SIGINT -f "^gpu-screen-recorder" # SIGINT required to save video properly
 
@@ -365,12 +376,13 @@ stop_screenrecording() {
   toggle_screenrecording_indicator
   cleanup_webcam
 
+  local filename
+  filename=$(cat "$RECORDING_FILE" 2>/dev/null)
+
   if pgrep -f "^gpu-screen-recorder" >/dev/null; then
     pkill -9 -f "^gpu-screen-recorder"
     omarchy-notification-send -u critical -t 5000 "Screen recording error" "Recording process had to be force-killed. Video may be corrupted."
   else
-    local filename
-    filename=$(cat "$RECORDING_FILE" 2>/dev/null)
     finalize_recording "$filename"
     echo "$filename"
     local preview="${filename%.mp4}-preview.png"
@@ -391,7 +403,7 @@ stop_screenrecording() {
     ) &
   fi
 
-  rm -f "$RECORDING_FILE"
+  release_recording_file "$filename"
 }
 
 toggle_screenrecording_indicator() {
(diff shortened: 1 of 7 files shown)


ogarza:
Kept the prompt in stop, moving it out still leaves a waiter, and you'd pay for a second encode if they say yes after the file is already “saved.”

The actual hang was openExistingMenu / openDmenu wiping doneFile without finishRequest. Image picker already does this in openSelector (finishDoneFile before the new request). Menu now finishes first, so Super+Space (or any other menu) unblocks stop as keep-original. Esc was already fine.

Also stopped re-reading RECORDING_FILE after finalize. It's captured once at the start of stop and used all the way through preview/cleanup, so a late answer can't toast or delete some other recording.

If nobody answers, we wait 120s then keep original levels. Killing the waiter left the dialog sitting there after the saved toast (hit that on a real stop), so timeout now also omarchy-menu close. Didn't put a timeout on omarchy-menu-select itself.

Dropped --normalize-audio / --no-normalize-audio altogether, not just the start sidecar. Nothing in the UI passed them, and OMARCHY_SCREENRECORD_NORMALIZE already covers skip-the-prompt. That also kills the env-vs-start-flag question. I tried to get too clever here, oops.

The Menu.qml fix is worth keeping even if we yank the prompt from stop later, transcode/timezone/webcam-pick all hung the same way on replace.

As for measuring on real hardware, I tested:
- screencapture with mic, whispering = prompted as too quiet.
- screencapture with mic, pretending to talk to someone on the couch next to me = not prompted.
- screencapture with mic, pretending to curse loudly at someone on the couch next to me = prompted as too loud.
- a talking youtube video  at 100% volume = no prompted.
- the same talking youtube video at 40% volume = prompted as too quiet.

As for the delay between stop and the normalize prompt: I tried a 30-minute recording here and it was a bad experience. The bar indicator was already gone, so I thought I wasn’t recording anymore and started another capture before the prompt even showed up.

I’d keep the current path for short files (say 10 minutes or less): analyze, and only ask if the levels actually look off. On longer files, ask right away with no analysis first. If they say yes, do a single loudnorm pass in the remux — same as we already do when normalize is always on. Short clips stay out of the way; after a long demo or cast, one extra click is fine, and you don’t sit there staring at a dead screen.

Thoughts on this @dhh ?

ogarza:
If you stop a recording and start another one before the save is fully done, like during that quiet pause after a long capture, we were wiping the new recording’s saved path. Next time you hit stop, we wouldn’t know which file to finish.

Now we only clear that path if it still belongs to the recording we just stopped. A new one that already started is left alone.

Also took --normalize-audio / --no-normalize-audio out of the PR description. Those flags aren’t a thing anymore; OMARCHY_SCREENRECORD_NORMALIZE is how you skip the prompt.

Still open: what to do about long recordings sitting there for a bit after you stop, before the normalize question shows up.

ogarza:
Over 10 minutes we skip the analysis and ask immediately: normalize to typical broadcast levels (Recommended), or keep original. Yes is a single loudnorm pass, same as always-on. Short files still measure first and only ask if the levels actually look off.

Unreadable duration falls through to analysis too.

victorhugout:
Confirmed this fixes a real case on Omarchy 4.0.2-1 with gpu-screen-recorder 6.0.1-1 and the internal Realtek ALC3246 microphone.

After correcting excessive hardware gain, a direct PipeWire capture had a background noise floor around -54 dB. The standard screen-recording flow then applied loudnorm; a quiet section of the resulting recording measured around -15 dB, making the ambient noise clearly audible as static.

Keeping the startup-pop mute/fade but removing loudnorm resolved the issue: voice is clear and background static is gone. Asking before normalizing unusually quiet/loud recordings would have prevented this.

PR #8816, opened 2026-08-28
Add Bluetooth keyboard support for encrypted disk unlock

damos1212:
## Why

Bluetooth normally becomes available only after the encrypted root filesystem
has been unlocked and the regular BlueZ service starts. A paired Bluetooth
keyboard therefore cannot enter the LUKS passphrase, forcing users to keep a
second keyboard, receiver, or data-capable cable available for every boot.

This is especially limiting for Bluetooth-only keyboards and machines without
a suitable free USB port. It is also distinct from the graphical login screen:
the missing capability is in the early boot environment, before the installed
system can start its normal services.

This explores the implementation requested in
[Discussion #6240](https://github.com/omacom/omarchy/discussions/6240).

## What changes

- Add `Setup > Security > Bluetooth Disk Unlock` as an explicit enable/disable
  control. The placement is practical rather than prescriptive; the primary
  change is the early-boot capability.
- Let the user select one already paired and trusted Bluetooth HID keyboard.
- Install a native mkinitcpio hook immediately before Omarchy's `encrypt` hook.
- Start a minimal D-Bus and BlueZ environment while LUKS is waiting.
- Include Bluetooth HID modules and common Intel, Realtek, MediaTek, Qualcomm,
  and Broadcom Bluetooth firmware.
- Copy only the selected keyboard bond and controller metadata into the boot
  image, rather than the complete BlueZ database.
- Provide `enable`, `disable`, and `status` operations and rebuild the Limine
  boot image when the state changes.

The current implementation deliberately targets Omarchy's classic `encrypt`
mkinitcpio flow and fails clearly for unsupported hook layouts.

## Security and recovery

Enabling this places the selected keyboard's Bluetooth bond material in the
boot image on the unencrypted EFI System Partition. The setup command explains
this before making changes and recommends keeping wired or 2.4 GHz input
available for recovery. Bluetooth can still fail because of batteries, radio
interference, firmware, or adapter changes.

The hook creates a minimal `passwd`/`group` database containing only root and
D-Bus. It never copies `/etc/shadow`, and it does not copy bonds belonging to
phones, headphones, or other input devices.

## Failure handling

Setup validates the effective mkinitcpio hook order before changing files. Failed setup or removal restores the previous configuration and permissions, including partially installed configurations. A failed boot-image rebuild can still leave an incomplete image; the command explicitly requires repairing and rebuilding it before rebooting. Configuration rollback does not restore boot-image contents.

`status` reports configuration state, not a verified image or keyboard connection. `--no-rebuild` stages configuration for the installer's final image build and reports that distinction.

## Validation

- Focused setup and runtime regression tests: `bash test/shell.d/bluetooth-unlock-test.sh`.
- CLI routing and metadata: `./test/cli`.
- Bash syntax, BusyBox ash runtime fixtures, and `git diff --check`.

Regression coverage includes failed enable/re-enable/disable rebuilds, partial installation removal, effective hook ordering, daemon cleanup, and D-Bus startup failure. Runtime fixtures isolate filesystem paths and mock process control; they do not exercise Bluetooth hardware.

## Remaining verification

This PR is open for review. Real encrypted-hardware boot/reconnection, adapter firmware coverage, recovery after a failed image build, and the running Setup > Security menu with and without an adapter remain unverified. Do not interpret the automated checks as hardware or UI acceptance.

### DIFF
--- a/bin/omarchy-setup-security-bluetooth-unlock
+++ b/bin/omarchy-setup-security-bluetooth-unlock
@@ -17,11 +17,13 @@ no_rebuild=false
 while (( $# )); do
   case "$1" in
     --device)
-      device=${2:-}
+      (( $# >= 2 )) || { echo "Missing address after --device." >&2; exit 1; }
+      device=$2
       shift 2
       ;;
     --controller)
-      controller=${2:-}
+      (( $# >= 2 )) || { echo "Missing address after --controller." >&2; exit 1; }
+      controller=$2
       shift 2
       ;;
     --yes)
@@ -60,6 +62,10 @@ valid_address() {
   [[ $1 =~ ^([0-9A-Fa-f]{2}:){5}[0-9A-Fa-f]{2}$ ]]
 }
 
+keyboard_info() {
+  [[ $1 == *"Icon: input-keyboard"* ]] || grep -qE '^[[:space:]]*Appearance: 0x0*3c1([[:space:]]|$)' <<<"$1"
+}
+
 normalize_address() {
   printf '%s\n' "${1^^}"
 }
@@ -70,10 +76,7 @@ configured() {
 
 rebuild() {
   $no_rebuild && return 0
-  if omarchy-cmd-missing limine-mkinitcpio; then
-    echo "Bluetooth disk unlock requires the Limine boot setup." >&2
-    exit 1
-  fi
+  rebuild_started=true
   echo "Regenerating the boot image..."
   sudo limine-mkinitcpio
 }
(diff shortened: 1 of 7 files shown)


damos1212:
Audited the current PR and added a follow-up fix for setup recovery and the early-boot lifecycle.

- Failed enable, re-enable, and disable now restore the previous configuration and permissions. Partial installs can be removed. If restoration itself fails, the backup is retained for recovery.
- Setup checks the effective mkinitcpio configuration, including later overrides, so the Bluetooth hook must remain immediately before `encrypt` with `udev` available.
- Explicit controller/device setup validates the selected trusted keyboard bond without relying on the default adapter or a running target D-Bus. Generic HID devices are no longer sufficient keyboard evidence.
- The runtime retains foreground BlueZ PID tracking, handles D-Bus startup failure, bounds daemon teardown, and removes its bus state before switch-root. The build hook uses explicit Bluetooth defaults and rejects additional unsafe state symlinks.
- Status and `--no-rebuild` output distinguish configuration from a rebuilt or hardware-verified image. Recovery documentation explains that configuration rollback cannot undo a partially written boot image.

Validation: 18 focused setup/runtime assertions passed; `./test/cli` passed with an isolated runtime directory; Bash/BusyBox ash syntax and `git diff --check` passed. An isolated cross-PR fixture exercised the ISO handoff through this actual setup command and back through ISO validation for Classic Bluetooth and BLE keyboard bonds. No host Bluetooth or boot configuration was changed.

Remaining acceptance work: an actual initramfs build, encrypted hardware boot/reconnection and recovery tests, and the running Setup > Security menu check with and without an adapter. The existing visual-review request is still unverified. The PR description now matches its current open-for-review state and states these limits explicitly.

PR #8818, opened 2026-08-28
Recover stale fprintd after suspend

martlets:
## What this changes

- Record `fprintd.service`'s PID immediately before suspend or hibernate.
- After resume, stop the daemon only when that exact PID survived the sleep transition.
- Let the next fingerprint request start a clean daemon through normal D-Bus activation, after `user.slice` has thawed.
- Install the hook for new fingerprint setups, migrate existing setups once, and remove it when fingerprint authentication is removed.
- Log only an actual recovery or a bounded stop failure; ordinary sleep cycles remain silent.

## Why

The lock screen can begin a fingerprint PAM conversation just as a lid close enters suspend. On the affected machine, `fprintd` reported `Cannot run while suspended`, but the daemon survived with its device claim stuck. After resume, the lock screen retried every 250 ms and received `PAM_AUTHINFO_UNAVAIL` (code 9) until password unlock or the daemon's idle exit.

The reader itself remained healthy: stopping the surviving daemon and allowing D-Bus to start a fresh one restored fingerprint verification immediately.

The PID comparison makes the recovery selective. If `fprintd` exited during sleep and another system component started a replacement, this hook leaves that new process alone.

## Why stop instead of restart inside the hook?

System-sleep hooks run while `user.slice` is frozen. Stopping the stale daemon there establishes the ordering we need, but starting `fprintd` there would also make wake wait for USB probing and the D-Bus service start. Leaving startup to the next PAM request moves that work after thaw, when the reader has fully resumed.

The service stop is bounded at six seconds, one second beyond Omarchy's five-second systemd service-stop timeout. A failure is logged and wake continues.

## Hardware validation

Tested on Omarchy 4.0.0-1 with:

- Goodix MOC fingerprint reader `27c6:634c`
- `fprintd` 1.94.5-2
- `libfprint` 1.94.100-1
- Linux 7.1.8-arch1-3, s2idle

Historical journal data contained three `Cannot run while suspended` events across twelve lid cycles, followed by a large code-9 retry storm.

With this hook installed, the race was reproduced and recovered in three consecutive test cycles:

1. 18-second sleep on battery
2. 17-minute sleep on battery
3. 10-minute sleep beginning on AC power and resuming on battery

In every cycle:

- the pre-suspend PID survived;
- PAM emitted code 9 before the post-resume hook ran;
- the hook stopped that exact PID successfully;
- the next PAM attempt reached the reader within one to two seconds; and
- fingerprint authentication succeeded with no subsequent code-9 storm.

The power-state instrumentation used during validation is not included in this patch.

## Scope and relationship to other work

This fixes the stale-daemon suspend race reported in #7229 and related evidence in #7172. It deliberately does not change the lock screen's retry policy, unavailable UI, thermal-error handling, or mid-verification driver-error behavior.

#7158 addresses the same stale-daemon diagnosis as part of a broader fingerprint recovery and retry design. This PR offers a narrower alternative: a PID-selective stop, lazy D-Bus activation after thaw, and no QML or retry-policy changes. The approaches could also be combined if maintainers prefer its broader backoff behavior.

One intentional tradeoff is that a fingerprint conversation left open across suspend, such as `sudo` waiting in another terminal, is ended and falls back to password. That is the same class of conversation that can otherwise retain the stale claim.

## Tests

- `test/shell.d/fprintd-resume-stop-test.sh`
- `test/shell.d/fprintd-resume-stop-migration-test.sh`
- `./test/cli`
- `./test/shell`: all relevant tests pass; the sole persistent failure is `launch-about-test.sh`, reproduced unchanged on a detached `origin/quattro` worktree. The initially transient screenshot test passed on rerun, and the Snapper coverage passed once its documented sibling `omarchy-pkgs` and `omarchy-iso` checkouts were present.

@spencerbull, I would value your review of the suspend/service-lifecycle tradeoff here, particularly the selective stop and deferred D-Bus startup versus restarting the service inside the post-resume hook.

### DIFF
--- a/bin/omarchy-remove-security-fingerprint
+++ b/bin/omarchy-remove-security-fingerprint
@@ -27,11 +27,19 @@ remove_lock_fingerprint_pam() {
   fi
 }
 
+remove_resume_recovery() {
+  if [[ -f /usr/lib/systemd/system-sleep/fprintd-resume-stop ]]; then
+    echo "Removing fingerprint recovery after resume..."
+    sudo rm -f /usr/lib/systemd/system-sleep/fprintd-resume-stop
+  fi
+}
+
 
 echo -e "\e[32mRemoving fingerprint scanner from authentication.\n\e[0m"
 
 remove_pam_config
 remove_lock_fingerprint_pam
+remove_resume_recovery
 
 echo "Removing fingerprint packages..."
 omarchy-pkg-drop fprintd libfprint libfprint-git
--- a/bin/omarchy-setup-security-fingerprint
+++ b/bin/omarchy-setup-security-fingerprint
@@ -62,6 +62,13 @@ account    include                     system-local-login
 EOF
 }
 
+install_resume_recovery() {
+  echo "Installing fingerprint recovery after resume..."
+  sudo install -Dm755 \
+    "$OMARCHY_PATH/default/systemd/system-sleep/fprintd-resume-stop" \
+    /usr/lib/systemd/system-sleep/fprintd-resume-stop
+}
+
 
 echo -e "\e[32mSetting up fingerprint scanner for authentication.\n\e[0m"
 
@@ -100,6 +107,7 @@ if sudo fprintd-enroll "$USER"; then
     # pam_fprintd with nothing to match.
     setup_pam_config
     setup_lock_fingerprint_pam
+    install_resume_recovery
     echo -e "\e[32m\nPerfect! Fingerprint authentication is now configured.\e[0m"
     echo "You can use your fingerprint for sudo, polkit, and lock screen (Super + Ctrl + L)."
   else
--- a/default/systemd/system-sleep/fprintd-resume-stop
+++ b/default/systemd/system-sleep/fprintd-resume-stop
@@ -1,0 +1,44 @@
+#!/bin/bash
+
+# A fingerprint verification that races suspend can leave fprintd alive with a
+# stale device claim. The lock screen then receives PAM_AUTHINFO_UNAVAIL on
+# every retry after resume even though the reader itself is healthy.
+#
+# Record the daemon before sleep, then stop it after resume only if that exact
+# process survived. The PID check leaves a daemon newly started by some other
+# system component alone. The lock screen's next D-Bus request starts a fresh
+# fprintd against the fully resumed reader, so the hook does not wait for USB
+# probing while user.slice is frozen.
+
+state_file="${OMARCHY_FPRINTD_SLEEP_STATE:-/run/omarchy-fprintd-sleep.pid}"
+
+log_recovery() {
+  logger -t omarchy-fprintd-resume -- "$*" 2>/dev/null || true
+}
+
+case "$1" in
+  pre)
+    fprintd_pid=$(systemctl show --property MainPID --value fprintd.service 2>/dev/null || true)
+    if [[ $fprintd_pid =~ ^[1-9][0-9]*$ ]]; then
+      printf '%s\n' "$fprintd_pid" >"$state_file"
+    else
+      rm -f "$state_file"
+    fi
+    ;;
+  post)
+    [[ -r $state_file ]] || exit 0
+    read -r sleeping_pid <"$state_file"
+    rm -f "$state_file"
+
+    current_pid=$(systemctl show --property MainPID --value fprintd.service 2>/dev/null || true)
+    if [[ $current_pid == "$sleeping_pid" ]]; then
+      # Omarchy bounds service stops at five seconds. Keep a separate outer
+      # bound so an unexpected systemd delay cannot freeze the desktop forever.
+      if timeout 6s systemctl stop fprintd.service 2>/dev/null; then
+        log_recovery "Stopped fprintd PID $sleeping_pid after $2 resume"
+      else
+        log_recovery "Failed to stop fprintd PID $sleeping_pid after $2 resume"
+      fi
+    fi
+    ;;
+esac
--- a/migrations/1787689809.sh
+++ b/migrations/1787689809.sh
@@ -1,0 +1,12 @@
+echo "Install fingerprint recovery after resume"
+
+hook_src="${OMARCHY_FPRINTD_RESUME_SRC:-$OMARCHY_PATH/default/systemd/system-sleep/fprintd-resume-stop}"
+hook_dst="${OMARCHY_FPRINTD_RESUME_DST:-/usr/lib/systemd/system-sleep/fprintd-resume-stop}"
+lock_pam="${OMARCHY_LOCK_FINGERPRINT_PAM:-/etc/pam.d/omarchy-lock-fingerprint}"
+
+# Fingerprint setup is per-machine, while migration completion is per-user.
+# Install once on configured machines and let later users no-op.
+[[ -f $lock_pam && -f $hook_src && ! -e $hook_dst ]] || exit 0
+
+echo "Installing fingerprint resume recovery"
+sudo install -Dm755 "$hook_src" "$hook_dst"
--- a/test/shell.d/fprintd-resume-stop-migration-test.sh
+++ b/test/shell.d/fprintd-resume-stop-migration-test.sh
@@ -1,0 +1,47 @@
+#!/bin/bash
+
+set -euo pipefail
+
+source "$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd)/base-test.sh"
+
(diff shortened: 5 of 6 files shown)


spencerbull:
Reviewed by Claude Opus 5 running in Claude Code, with an independent second opinion from Codex at xhigh reasoning. Tests ran on a disposable Omarchy 4 VM, not on a developer machine: `./test/cli` is 116 ok / 0 not ok, both new shell files pass (7 and 3), and `sleep-lock`, `sleep-monitor` and `unowned-system-paths` stay green. I also ran `migrations/1787689809.sh` for real against `/usr/lib/systemd/system-sleep` and real `sudo`, rather than only through the path overrides the test uses.

**The no-reader case is clean.** `systemctl show --property MainPID --value fprintd.service` returns `0` and exits 0 for a unit that is missing, inactive, masked, or whose package has been removed — I checked all four on a real machine — so the regex at line 21 rejects it, `pre` records nothing, and `post` exits at the `-r` guard on line 29. The migration no-ops without `/etc/pam.d/omarchy-lock-fingerprint`, and setup only installs after a successful enrol and verify. Power-cycling a suspended VM confirmed `/run` is tmpfs, so no state file survives a crash into a later resume. Eight rapid stop-then-D-Bus-activate cycles left the unit clean with no failed state and no `StartLimitBurst` trip, and the happy path works end to end: record, stop, log, re-activate on the next request. Your comment that Omarchy bounds service stops at five seconds is correct — `etc/systemd/system.conf.d/10-faster-shutdown.conf` sets `DefaultTimeoutStopSec=5s` and `fprintd.service` inherits it, so the 6s outer bound is sized right.

**1. The tests pass whether or not the hook does what they are named for.** Four mutations to the hook each leave `fprintd-resume-stop-test.sh` at 7 passed, 0 failed:

- delete `rm -f "$state_file"` at hook line 25;
- delete `rm -f "$state_file"` at hook line 31 — either cleanup alone makes the file absent by the time test line 89 checks, so neither is proven;
- add `systemctl restart fprintd.service` after a successful stop. Test line 66 is named "resume stops fprintd without restarting it in the hook", but `grep -qx 'stop fprintd.service'` still matches when a restart is also logged;
- add `else systemctl restart fprintd.service` to the comparison at hook line 34. Test line 82's `! grep -q '^stop '` ignores a restart entirely.

The last two are the ones that matter: not restarting inside the hook is the central design decision of this PR — it is the whole of your "Why stop instead of restart inside the hook?" section — and nothing currently holds it in place. Four added assertions close all four holes, and I verified each by re-running the mutation and watching the new assertion fail:

<details><summary>diff against <code>test/shell.d/fprintd-resume-stop-test.sh</code></summary>

```diff
@@ -65,6 +65,10 @@ MOCK_FPRINTD_PID=101 run_hook post suspend
   fail "resume bounds the stale daemon stop" "timeout: $(<"$timeout_log")"
 grep -qx 'stop fprintd.service' "$systemctl_log" ||
   fail "resume stops fprintd without restarting it in the hook" "systemctl: $(<"$systemctl_log")"
+! grep -q '^restart ' "$systemctl_log" ||
+  fail "resume leaves starting fprintd to the next PAM request" "systemctl: $(<"$systemctl_log")"
+[[ ! -e $state_file ]] ||
+  fail "resume clears the record it acted on" "state: $(<"$state_file")"
 grep -q 'Stopped fprintd PID 101 after suspend resume' "$logger_log" ||
   fail "resume logs successful recovery" "log: $(<"$logger_log")"
 pass "resume stops the stale daemon within a bound"
@@ -79,14 +83,18 @@ MOCK_FPRINTD_PID=201 run_hook pre suspend
 MOCK_FPRINTD_PID=202 run_hook post suspend
 [[ ! -s $timeout_log ]] ||
   fail "resume leaves a replacement daemon alone" "timeout: $(<"$timeout_log")"
-! grep -q '^stop ' "$systemctl_log" ||
+[[ $(<"$systemctl_log") == "show --property MainPID --value fprintd.service" ]] ||
   fail "resume leaves a replacement daemon alone" "systemctl: $(<"$systemctl_log")"
 [[ ! -s $logger_log ]] ||
   fail "resume stays quiet when preserving a replacement daemon" "log: $(<"$logger_log")"
 pass "resume leaves a replacement daemon alone"
 
+# Seed a record first: the previous resume already cleared the state file, so
+# without this the assertion below passes whether or not pre clears anything.
+printf '199\n' >"$state_file"
 MOCK_FPRINTD_PID=0 run_hook pre suspend
-[[ ! -e $state_file ]] || fail "pre-suspend records no inactive daemon"
+[[ ! -e $state_file ]] ||
+  fail "pre-suspend records no inactive daemon" "state: $(<"$state_file")"
 [[ ! -s $logger_log ]] ||
   fail "pre-suspend stays quiet without an active daemon" "log: $(<"$logger_log")"
 pass "pre-suspend records no inactive daemon"
```

</details>

**2. The PID comparison detects survival, not staleness, so the stop fires on ordinary resumes and not only on the race.** `omarchy-system-sleep-monitor` holds a delay inhibitor and runs `omarchy-system-sleep-lock`, which locks the session and waits for it to become secure before suspend proceeds. `shell/plugins/lock/Service.qml:237-241` arms the fingerprint PAM conversation the moment it goes secure, which D-Bus-activates fprintd. So on a configured machine that suspends by lid close or idle, the daemon is normally alive at `pre` and the same process at `post` — which is consistent with your own data, where the pre-suspend PID survived in all three cycles. Two consequences worth naming: "ordinary sleep cycles remain silent" does not hold, because every such resume logs `Stopped fprintd PID N`; and every ordinary resume now pays the cold start you measured at one to two seconds on the first touch, in exchange for recovering the racy minority. That may well be the right trade, and stopping a healthy daemon looks harmless here — I could not make it leave fprintd worse off — but the PR reads as though the stop were selective, and a maintainer weighing this should see it as the unconditional post-resume stop it effectively is.

Two smaller ones. `timeout 6s systemctl stop` kills the client but not the job systemd has already queued, so `Failed to stop` can be logged for a stop that then succeeds a moment later. And `docs/file-layout.md:127-128` names the three sleep hooks explicitly; it would want the fourth.

**This overlaps #7158 more than the description suggests.** That PR adds `default/systemd/system-sleep/fprintd-resume` through the same two scripts plus a migration of its own, and runs `timeout 10s systemctl try-restart fprintd.service`. systemd runs everything in `/usr/lib/systemd/system-sleep/` in parallel — I measured two probe hooks starting 53 microseconds apart while one of them slept three seconds — so if both land, a restart and a stop race on the same unit on every resume, with no ordering to rely on. Which approach to take is a maintainer's call rather than mine.

One thing I checked and rejected: #8531's description says `fprintd-resume.service` already restarts fprintd on resume, which would have made this hook redundant. Arch's `fprintd` package ships no such unit — only `fprintd.service`, a D-Bus system service file and `/etc/fprintd.conf` — so that is not an existing mechanism this duplicates.

Nothing was pushed to your branch. Waiting on you for the test assertions, and on a maintainer for the selectivity trade and the #7158 overlap.

k7cfo:
Independent confirmation from different hardware than the PR body.

- ThinkPad X1 Carbon Gen 13 (`21NSCTO1WW`)
- Synaptics `06cb:0123`
- `fprintd` 1.94.5-2, `libfprint` 1.94.100-1
- Linux 7.1.9-arch1-2, s2idle

On this machine the surviving daemon does not log `Cannot run while suspended`. After resume the reader re-enumerates and fprintd marks it `unsupported firmware version`, then lock-screen PAM storms with `PAM_AUTHINFO_UNAVAIL` (code 9) until password unlock or the daemon's idle exit. Same stale-claim shape as this PR.

What we verified:

- systemd 261 only executes `/usr/lib/systemd/system-sleep` (a hook in `/etc/systemd/system-sleep` never ran).
- Stopping the surviving fprintd and letting the next PAM request D-Bus-activate a fresh one restores the reader.
- Starting fprintd inside the post hook is the wrong move: it races `omarchy-lock-fingerprint` (`fprintd name owner changed during operation`, then a failed swipe). Stop without restart, as this PR does, is what actually worked here.

Happy to retest this branch on `06cb:0123` if that's useful.

sprajs:
One integration conflict to resolve before landing: #8441 independently adds the same `migrations/1787689809.sh` for Cursor’s password-store configuration. This PR uses it to install fingerprint resume recovery. Please give one of these unrelated migrations a different unused timestamp, updating its test reference too, so both operations retain distinct migration-completion markers.

bjarneo:
Automated duplication check: this pull request looks similar to #7158, which covers the same fprintd resume hook. I keep that one open and close this one to consolidate review.

PR #8820, opened 2026-08-28
Show inactive audio card outputs in picker

### LINKED ISSUE #8789: [Quattro] Audio picker cannot switch between HDMI/DP profiles on one GPU card
## System details

AMD desktop, NVIDIA GeForce RTX 3080 Ti (NVIDIA 610.57.04), two DisplayPort monitors, Omarchy 4.0.1-1, kernel 7.1.9-arch1-2, Hyprland 0.56.2-1, PipeWire 1.6.8-1, WirePlumber 0.5.15-1.

## What's wrong?

The Quattro audio output picker cannot switch between two HDMI/DisplayPort audio outputs when ALSA exposes them as mutually exclusive profiles of the same GPU audio card.

Both connected monitors are detected and both profiles are available:

```text
Card: alsa_card.pci-0000_29_00.1

output:hdmi-stereo:        available: yes
output:hdmi-stereo-extra1: available: yes
Active Profile: output:hdmi-stereo

hdmi-output-0: available
  device.product.name = "M27T6"
  Part of profile: output:hdmi-stereo

hdmi-output-1: available
  device.product.name = "H27T7P-2"
  Part of profile: output:hdmi-stereo-extra1
```

However, `wpctl status` contains only the sink belonging to the active profile:

```text
Sinks:
  * M27T6
```

The Omarchy audio panel consequently shows only that one row. Clicking it has no visible effect because it is already the default, while the second available monitor cannot be selected. Changing the ALSA card profile manually makes the other sink appear and the first disappear.

### Steps to reproduce

1. Connect two audio-capable HDMI/DisplayPort monitors to one GPU.
2. Confirm that `pactl list cards` reports two available profiles on the same card, such as `output:hdmi-stereo` and `output:hdmi-stereo-extra1`.
3. Open the Omarchy audio panel.
4. Observe that only the sink from the currently active card profile is offered.
5. Try to select the other monitor; it is absent from the picker.

### Expected

The output picker should represent all available playback profiles/ports and activate the selected card profile when necessary, so either connected monitor can be chosen.

### Actual

The picker enumerates only current `Pipewire.nodes` sinks. With ACP-style mutually exclusive profiles, that means it can see only the active monitor and cannot perform a profile switch.

### Related re-enumeration symptom

After a hibernate/DisplayPort link reinitialization, the active sink appeared in the panel as the generic `HDMI 0` even though the card-port properties and kernel ELD already contained `M27T6`. A local WirePlumber naming rule corrected the label, but it does not address the missing profile switching.

This is related to the DPMS/profile churn described in #8290, but distinct: #8290 concerns automatic profile/default-sink changes when outputs become unavailable; this report concerns the audio panel being unable to select another profile even while both ports/profiles are available.

ClGratton:
## Problem

Quattro's audio panel only lists live PipeWire sink nodes. Some ALSA cards expose multiple mutually exclusive output profiles, so only the currently active HDMI/DisplayPort monitor has a sink. Other connected, available monitors are invisible in the picker even though `pactl list cards` reports their profiles as available.

Fixes #8789.

## Change

- Discover the best available one-sink output profile for every available ALSA card port.
- Merge inactive card profiles into the panel's normal output rows, using the monitor/product name supplied by the port.
- When an inactive row is selected, activate that profile, wait for its PipeWire sink to appear, then reuse the existing default-output helper to persist the selection and move playback streams.
- Keep active sinks on the existing Quickshell/PipeWire path.

PR #5317 addresses the related single-sink limitation in the old `dev` command switcher, but it hard-codes an analog/HDMI toggle and does not expose multiple HDMI ports in Quattro's panel. PRs #5830 and #8426 make missing HDMI hardware/profiles appear; this change handles profiles that already exist and are available but are mutually exclusive.

## Verification

- `bash test/shell.d/audio-output-profile-test.sh`
- `./test/cli`
- `./test/shell` (new audio tests pass; unrelated local failures require the separate `omarchy-pkgs` checkout, network access, or depend on the live multi-monitor shell environment)
- Live two-monitor test on an NVIDIA RTX 3080 Ti:
  - `M27T6` -> `output:hdmi-stereo`
  - `H27T7P-2` -> `output:hdmi-stereo-extra1`
  - selecting each panel row recreated the expected sink and made it the default

### DIFF
--- a/shell/plugins/panels/audio/Model.js
+++ b/shell/plugins/panels/audio/Model.js
@@ -47,6 +47,96 @@ function parseSinkAvailability(raw) {
   return next
 }
 
+function parseOutputProfiles(raw) {
+  try {
+    var parsed = JSON.parse(String(raw || "[]"))
+    if (!Array.isArray(parsed)) return []
+
+    var profiles = []
+    for (var i = 0; i < parsed.length; i++) {
+      var profile = parsed[i] || {}
+      var cardName = String(profile.cardName || "")
+      var profileName = String(profile.profileName || "")
+      if (!cardName || !profileName) continue
+      profiles.push({
+        cardName: cardName,
+        profileName: profileName,
+        label: friendlyDeviceLabel(profile.label || profile.description || profileName),
+        description: friendlyDeviceLabel(profile.description || "")
+      })
+    }
+    return profiles
+  } catch (error) {
+    return []
+  }
+}
+
+function outputProfileKey(cardName, profileName) {
+  return String(cardName || "") + "|" + String(profileName || "")
+}
+
+function outputProfileSinkName(cardName, profileName) {
+  cardName = String(cardName || "")
+  profileName = String(profileName || "")
+  if (cardName.indexOf("alsa_card.") !== 0 || profileName.indexOf("output:") !== 0) return ""
+  return "alsa_output." + cardName.slice(10) + "." + profileName.slice(7)
+}
+
+function sinkProfileKey(node) {
+  var p = nodeProps(node)
+  var cardName = p["device.name"] || ""
+  var profileName = p["device.profile.name"] || ""
+  if (!cardName || !profileName) return ""
+  if (String(profileName).indexOf("output:") !== 0) profileName = "output:" + profileName
+  return outputProfileKey(cardName, profileName)
+}
+
+function outputRows(sinks, profiles) {
+  var rows = []
+  var activeProfiles = {}
+  var profilesByKey = {}
+  var profilesBySinkName = {}
+  var sinkValues = Array.isArray(sinks) ? sinks : []
+  var profileValues = Array.isArray(profiles) ? profiles : []
+
+  for (var i = 0; i < profileValues.length; i++) {
+    var availableProfile = profileValues[i]
+    var availableKey = outputProfileKey(availableProfile.cardName, availableProfile.profileName)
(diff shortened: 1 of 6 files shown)


ClGratton:
Durable live verification completed on the reporting machine (NVIDIA RTX 3080 Ti, two HDMI/DP monitor profiles). After linking this branch as the active Omarchy checkout and restarting the shell, the audio panel showed both `M27T6` (`output:hdmi-stereo`) and `H27T7P-2` (`output:hdmi-stereo-extra1`). The reporter confirmed the fix works.

shell/plugins/panels/audio/Model.js
```diff
@@ -47,6 +47,96 @@ function parseSinkAvailability(raw) {
...
+    rows.push({
+      kind: "profile",
+      cardName: inactiveProfile.cardName,
+      profileName: inactiveProfile.profileName,
+      label: inactiveProfile.label,
+      description: inactiveProfile.description
+    })
+  }
+
+  return rows
```
lucaengelhard: Hey I love this! I had the same issue and this fixed it for me.

Small nit maybe: 
When switching the output it reorders it for me, so maybe just add a sort to the outputRows function?

`return rows.sort((a, b) => a.label < b.label ? -1 : a.label > b.label ? 1 : 0) `

ClGratton: Thanks — good catch. I added deterministic label sorting so the selected output no longer jumps to the front, plus a regression test, in 5e5ae63d.

PR #8823, opened 2026-08-28
Add OpenCode Go collector for agents panel

kczimm:
# Add OpenCode Go collector for agents panel

## Summary

Adds native support for OpenCode Go subscriptions in the agents panel. OpenCode Go is a $10/month subscription providing access to popular open coding models, used by Pi, OpenCode CLI, and other harnesses.

## Changes

- **`bin/omarchy-agent-usage-opencode-go`** — New collector that queries OpenCode's official `/zen/go/v1/usage` API for rolling (5h), weekly, and monthly limit usage with reset timers
- **`shell/plugins/agents/assets/opencode-go.svg`** — Dark mode icon
- **`shell/plugins/agents/assets/opencode-go-light.svg`** — Light mode icon
- **`shell/plugins/agents/manifest.json`** — Add `opencode-go` to default providers
- **`shell/plugins/agents/README.md`** — Document the new collector and its key discovery logic

### Harness-agnostic key discovery

The OpenCode Go API key is the same regardless of which harness (Pi, OpenCode CLI, etc.) connects to the subscription. The collector searches multiple auth locations:

1. `OPENCODE_GO_API_KEY` environment variable
2. `~/.pi/agent/auth.json` (Pi harness)
3. `~/.local/share/opencode/auth.json` (OpenCode CLI)

Adding a new harness requires only appending to `AUTH_PATHS` in the script.

### curl instead of urllib

Other collectors use `urllib.request` directly, but `opencode.ai` is behind Cloudflare which blocks requests based on TLS fingerprinting (JA3/JA4). Python's `ssl` module produces a fingerprint Cloudflare flags as non-browser traffic, returning 403. Curl has a browser-like fingerprint and passes through. This deviation is documented in the script.

### No local stats

Unlike Claude/Codex collectors that parse local session files, this collector relies solely on the server-side API. OpenCode's local database doesn't reliably represent server-side accounting, especially when using multiple machines. The API provides accurate, real-time limits.

## Testing

Verified working on a live OpenCode Go subscription:
- API returns rolling/weekly/monthly usage percentages and reset timestamps
- Panel displays correctly with limit meters
- IPC refresh works: `omarchy-shell omarchy.agents refresh`

## Related

- [Anomalyco/opencode#16017](https://github.com/anomalyco/opencode/issues/16017) — Original feature request
- [Anomalyco/opencode#16513](https://github.com/anomalyco/opencode/pull/16513) — API endpoint implementation

### DIFF
--- a/bin/omarchy-agent-usage-opencode-go
+++ b/bin/omarchy-agent-usage-opencode-go
@@ -1,12 +1,13 @@
-#!/usr/bin/python3
+#!/usr/bin/env python3
 # omarchy:summary=Print the OpenCode Go usage record as JSON
 # omarchy:args=[--force] [--limits-only]
 # omarchy:hidden=true
 """Collect OpenCode Go usage into one display-ready JSON record.
 
 Limits come from OpenCode's official Go usage API (/zen/go/v1/usage),
-authenticated with the subscription API key. The key is the same regardless
-of which harness (Pi, OpenCode CLI, etc.) connects to the subscription.
+authenticated with the subscription API key. Local token stats come from
+Pi session files and the OpenCode database when available. The key is
+the same regardless of which harness connects to the subscription.
 The agents panel only ever reads the JSON this prints.
 """
 
@@ -15,6 +16,7 @@ from __future__ import annotations
 import argparse
 import json
 import os
+import sqlite3
 import subprocess
 from datetime import datetime, timedelta, timezone
 from pathlib import Path
@@ -53,6 +55,9 @@ def local_day(value: Any) -> str:
         return datetime.now().strftime("%Y-%m-%d")
 
 
+def model_name(raw: Any) -> str:
+    value = str(raw or "opencode-go")
+    return value if value else "opencode-go"
 
 
 def api_key() -> str:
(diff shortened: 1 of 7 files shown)


averyrobbins:
Tried this collector on a live OpenCode Go subscription (Omarchy 4.0.2-1, agents plugin 1.0.0). The limits half works great — `/zen/go/v1/usage` returns rolling/weekly/monthly with reset timers exactly as described. Thanks for building it.

One bug though: **local stats are dropped on any day the subscription hasn't been used yet.**

In `main()`:

```python
local_stats = cached_local_stats(max_age)
if local_stats.get("todayTotalTokens", 0) > 0:
    record.update(local_stats)
```

Because the guard keys on *today's* tokens, a scan that found plenty of history still gets discarded whenever today's total is 0. The panel then renders limit meters above a permanently empty "Tokens by day" and "Tokens by model" — which is the common case first thing in the morning, or for anyone who uses the subscription a few days a week.

### Repro

With OpenCode Go usage on disk from previous days but none today:

```
$ omarchy-agent-usage-opencode-go --force | jq '{totalPrompts, activeDays, modelUsage}'
{
  "totalPrompts": 0,
  "activeDays": 0,
  "modelUsage": {}
}
```

The scan itself is fine — the data is found and cached correctly, it just never reaches the record:

```
# same run, calling the scan directly
>>> complete, stats = scan_opencode_sessions()
totalPrompts: 124   activeDays: 2
modelUsage: muse-spark-1.3-contributor, grok-4.6
```

And the on-disk scan cache holds the right values too:

```
$ jq '.stats | {totalPrompts, activeDays}' ~/.cache/omarchy/agent-usage/opencode-go-scan-*.json
{ "totalPrompts": 124, "activeDays": 2 }
```

### Suggested fix

Gate on whether the scan produced anything at all rather than on today specifically:

```python
if (local_stats.get("todayTotalTokens", 0) > 0
        or local_stats.get("totalPrompts", 0) > 0
        or local_stats.get("modelUsage")):
    record.update(local_stats)
```

That keeps the original intent — don't clobber the record with an empty scan — while letting prior-day history through. Running with this change, the same machine reports `totalPrompts: 124`, `activeDays: 2`, and both models, and the panel charts populate.

Happy to open a PR against your branch if that's easier.

kczimm:
@averyrobbins thanks for the thorough review. The guard is now fixed to accept prior-day history.

bjarneo:

PR #8827, opened 2026-08-28
Take a screenshot with Super + Shift + S

TripleU613:
## What this does

Binds **Super + Shift + S** to a screenshot, matching what people arriving from macOS and Windows already reach for.

`SUPER + PRINT` stays exactly as it is. This adds the shortcut people try first, rather than replacing anything.

## Why

`PRINT` is awkward or absent on a lot of hardware — most laptops put it behind a function layer, and plenty of compact and 60% keyboards have no `PRINT` key at all. On those machines the screenshot binding is effectively unreachable until you go and read the hotkey list.

Super + Shift + S is the shortcut both macOS and Windows users already have in their fingers, and `manual/03-coming-from-mac-or-windows.md` is where they land first. It now tells them something true.

## The migration

Hyprland stacks duplicate binds rather than replacing them, so anyone who has already bound this chord themselves would get their command *and* the new default firing together.

The migration inserts `hl.unbind("SUPER + SHIFT + S")` above their own bind — the pattern this repo documents everywhere else — so their binding stays exactly as they wrote it and simply becomes the only one that runs. It is deliberately not conditional on what they bound to it: a chord rebound to another command stacks just as badly as a duplicate screenshot, and anyone running `omarchy_default_bindings = false` keeps the only screenshot shortcut they have either way.

It writes with `--follow-symlinks`, since a dotfile-managed `bindings.lua` is usually a symlink into a repo and plain `sed -i` would replace the link with a regular file, quietly detaching the config from the source the user edits.

## Tests

- `test/shell.d/screenshot-binding-migration-test.sh` — the kept binding, a chord rebound to another command, an existing unbind (not duplicated, idempotent across reruns), the commented-out template, unrelated bindings, and a symlinked config whose edit must land in the repo it points at
- `test/shell.d/hyprland-default-config-test.sh` — the default config binds both `SUPER + PRINT` and `SUPER + SHIFT + S`

Both pass, along with `./test/cli` and the rest of `./test/shell`.

https://claude.ai/code/session_01KLASdd8ciFib2wLJbN4d9B

### DIFF
--- a/config/hypr/bindings.lua
+++ b/config/hypr/bindings.lua
@@ -23,7 +23,9 @@
 -- Disable a default binding without replacing it.
 -- hl.unbind("SUPER + SHIFT + B")
 
+-- Keyboards without a Print Screen key (Apple, Logitech MX Keys) already have
+-- Super + Shift + S for screenshots by default.
+
 -- Logitech MX Keys examples:
--- o.bind("SUPER + SHIFT + S", nil, "omarchy-capture-screenshot")
 -- o.bind("SUPER + H", nil, "voxtype record toggle")
 -- o.bind("SUPER + PERIOD", nil, "omarchy-shell shell toggle omarchy.emojis")
--- a/default/hypr/bindings/applications.lua
+++ b/default/hypr/bindings/applications.lua
@@ -28,7 +28,9 @@ if o.preinstalled_bindings_enabled() then
   o.bind("SUPER + SHIFT + ALT + G", "WhatsApp", { webapp = "https://web.whatsapp.com/", focus = true })
   o.bind( "SUPER + SHIFT + CTRL + G", "Google Messages", { webapp = "https://messages.google.com/web/conversations", focus = true })
   o.bind("SUPER + SHIFT + P", "Google Photos", { webapp = "https://photos.google.com/", focus = true })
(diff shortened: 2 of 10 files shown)


TripleU613:
Both review points were right, and the second one mattered more than it first looked.

**Symlinked configs** — fixed with `--follow-symlinks`. Test added that migrates through a symlink and asserts both that the link survives and that the edit lands in the repo it points at.

**The migration contract** — this was the real bug. Commenting out an active bind takes away the only screenshot shortcut anyone running `omarchy_default_bindings = false` has, and matching only on `omarchy-capture-screenshot` left a chord rebound to something else stacking with the new default, so both fired.

Switched to inserting `hl.unbind("SUPER + SHIFT + S")` above the user's own bind, as suggested — it is also the pattern this repo documents elsewhere. Their line is untouched and simply becomes the only one that runs, whatever they bound to it. The PR description said the opposite of what the code did; that has been corrected too.

Tests now cover: the kept binding, a chord rebound to another command, an existing unbind (not duplicated, idempotent across reruns), the commented-out template, unrelated bindings, and the symlink case. All 11 pass.

TripleU613:
Big thing about this is apple doesn't have prtscrn button. So super shift S is from windows... Closest yet.

PR #8830, opened 2026-08-28
Size bar icons by the ink they paint

TripleU613:
## The problem

Icons in the bar come from a font, and a font gives every glyph a box that is mostly empty. Two glyphs with identical boxes can carry very different amounts of drawing, so lining the boxes up leaves the marks themselves scattered.

Measured across the twelve glyphs on a full bar, the vertical centres of the icons span **2.7px**. One wide glyph rendered **6px tall beside icons of 15**.

## What this changes

Each icon is measured by what it actually paints, and sized from that.

**Shape is read before brightness.** A logo drawn in two tones measures exactly like the same logo drawn in one. The Tailscale mark fades five of its nine dots; judged by brightness those dots are not ink, which centres the mark on the four that remain — 4px above where it belongs.

**Size comes from the middle of the mark's two dimensions**, not its longest side. Fitting the long side alone halves a mark twice as wide as it is tall; fitting the short side inflates the same mark into a slab that dominates the row. The middle gives it the row's proportions.

**Then nudged by density**, because two icons the same size on a ruler are not the same size to a reader — a solid slab and a hairline arc filling one box are nowhere near the same weight. Over the same twelve glyphs this takes the spread of inked area from **38.4% to 23.6%**. Pulling all the way to equal area is worse; it inflates a hairline glyph until it towers over everything.

**Position comes from weight, not the box.** A mark whose weight sits high in its box rides high on the bar and one whose weight sits low sags, even when every box is perfectly centred. Weight is judged only where the mark can still move — one already pressed against its canvas is pinned by its own size.

The block every icon is fitted to is **measured from the font**, not set as a number, so evening out a row does not enlarge it: icons that already filled the font's icon box keep the size they have always had.

## Relationship to the existing `OpticalGlyph`

Worth being direct, because this replaces a deliberate decision rather than filling a gap.

Today's `OpticalGlyph` keeps each glyph at its natural size on a shared baseline and corrects only horizontal centring, with a comment explaining the intent: *"Keep the shared line box and baseline intact. Correcting only the horizontal painted bounds avoids per-glyph vertical drift."*

That trade is what leaves the row uneven — a baseline shared by marks of different sizes still reads as scattered, which is the 2.7px above. This replaces it with per-glyph measurement. **The drift that comment warns about is real**, and is exactly what the verification pass and the rule table exist to hold in check: every icon's final render is measured and checked, and anything outside tolerance is reported rather than shipped quietly.

If that trade isn't one you want to make, I'd rather know than have this sit open — the measurement and the audit tool may still be worth having on their own.

## The rules live in one place

`shell/Commons/IconRules.qml`. The buttons, the tests and the audit tool all read them from there, so there is one definition of what "right" means. (The tests briefly grew their own copy of one rule during development; it drifted stricter than the real one and failed icons the rule deliberately exempts. That is the failure mode this structure is meant to prevent.)

`omarchy dev bar-icon-audit` scans the live bar and prints every icon's margins, how far its weight sits off centre, and what it painted against the canvas meant to hold it — exiting non-zero when any shown icon breaks a rule.

## Verification

On a live bar at two display scales:

| | before | after |
| --- | --- | --- |
| icon centre spread | 2.7px | **1.7px** |
| same row, HiDPI | — | **0.07px** |
| rule violations | — | **0** |

The 1.7px residual on a scale-1 display is the whole-pixel grid a natively rendered glyph can be placed on; the HiDPI figure is the same row with finer placement available.

`./test/cli` passes. `./test/shell` passes apart from four failures this machine already had before the change (missing `omarchy-pkgs` checkout, theme-install URL check, snapper, unowned system paths), plus the new `qml-text-format-test.sh` which passes.

https://claude.ai/code/session_01KLASdd8ciFib2wLJbN4d9B


TripleU613:
Thanks — six of the eight were real, and three of them meant a rule was claiming more than it checked. All addressed in f984cf2d.

**`contained` could never fire.** You were right that a grab is clipped to the item it comes from, so capturing the canvas exactly cut away the overflow the rule exists to catch — a margin could not come back negative. Both verifiers now capture through a padded frame and map the result back onto the canvas. There is a fixture that deliberately paints past its canvas to keep this honest: with the padding removed it reports `w=0 e=0` and the test fails, which is exactly the blind spot you described.

**Vertical bars judged as horizontal.** Correct — `compass()` and `balanceShift()` were both called without orientation, so a vertical bar balanced on the axis it was supposed to fill. The bar's orientation is passed down and used in both now.

**The frame merge.** Also correct, and worse than `crossAxis` alone: `Math.min` ran over every field, so after the second frame the axis was NaN, and keeping the numerically smallest balance hid the frame sitting furthest off centre. Only margins merge by minimum now; balance keeps the worst frame.

**Empty fixtures.** Right that they were invisible and reported verified without measuring — one correction, they were empty in the original commit rather than newly emptied, but the coverage was hollow either way. They have real glyphs now.

**Docs and `(( ))`.** Both fixed; the documentation described a grid and a blur-derived size that no longer exist.

Verified on a real bar after the change: no shown icon breaks a rule.

PR #8836, opened 2026-08-28
Add a first-party Local AI panel for registry-backed model serving

0xSero:
## Summary

Local models as **one button on the bar**, following the design direction set for 4.1: zero configuration required, nothing downloaded without consent, no stats, not in your face.

The widget detects local GPUs (NVIDIA, Intel Arc Pro B70) and shows the **recommended** registry recipe for that hardware — a `recommended` flag curated in the registry, else the first runnable recipe. One click downloads and loads it (the size is on the button — that click is the consent), one click unloads, one click opens the **Omarchy default agent** on the running model, one click **shares the endpoint on the user's tailnet**. Opt-in, same posture as Tailscale/Dropbox: `omarchy bar put omarchy.local-ai --before omarchy.agents`.

There is no dashboard, no recipe browser, and no stat anywhere: the popup is the entire UI, and it renders purely from `omarchy local ai snapshot`.

## Interaction

- Bar icon: empty circle, filled when a model is accepted. Right-click opens the default agent.
- Popup: recommended model first, up to five runnable models, one click each; Open agent · Share on Tailscale · Unload · Scan.
- `open-agent` follows `omarchy-default-agent`: Pi and Oh My Pi are launched with the local model passed explicitly; every other agent launches through `omarchy-agent`. The snapshot reports which known agents are installed.
- `share` toggles `tailscale serve` for the loopback port; unload always unpublishes. A denied serve surfaces tailscale's own one-time remedy (`sudo tailscale set --operator=$USER`).

## CLI

`omarchy local ai load <recipe> | unload | open-agent [name] | share | scan | download | run | switch | remove <recipe> | default | snapshot`

## Safety (unchanged posture)

Registry pinned by commit (`registry.pin`); recipes must be validated, digest-pinned, revision-pinned, mount-allowlisted; host networking/multi-node/unknown placeholders refused with visible reasons; only containers labeled `io.omarchy.local-ai=1` are ever touched; `127.0.0.1` bind unless explicitly shared; ready = model identity + real chat + real `shell` tool call; failed switch rolls back.

## Scope

Plugin: `Panel.qml` (117), `bin/omarchy-local-ai` (92), `lib/*` (six single-concern libraries, 574), `manifest.json` (19) — the 237-line Dashboard is deleted; net **−129 lines** against the previous revision while adding load-as-one-button, recommendation, default-agent handoff, and tailnet sharing. Tests: 34 cases (docker/curl/tailscale/GPU/agent shims; no daemon, network, or GPU).

## Verified live

On a mixed 2× RTX 3090 + 2× Arc Pro B70 Omarchy machine running the identical third-party build ([0xSero/omarchy-local-ai](https://github.com/0xSero/omarchy-local-ai)): install/update path, adoption of a running model across reinstall, pinned registry scan (7 runnable recipes), popup render inspected on-screen (recommended-first rows with download sizes, running row disabled, agent/share/unload actions), agent scan (8 agents found, default `opencode`), and the real tailscale-denied error path.

## Tests on this branch

```
test/shell.d/local-ai-test.sh            34/34
test/shell.d/panel-command-path-test.sh  pass
test/shell.d/plugins-test.sh             pass
```


shell/plugins/panels/local-ai/lib/agent.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
+# Pi / Oh My Pi provider wiring and TUI launch. Sourced; do not run.
+wire() {
+  local r=$1 model=$2 p d cur tmp
+  p=$(jq -nc --arg u "http://127.0.0.1:$PORT/v1" --arg m "$model" --arg n "$(jq -r '.model.name//.model.id' <<<"$r")" \
+    --argjson tools "$(jq -r '.capabilities.tools//false' <<<"$r")" \
+    '{baseUrl:$u,apiKey:"local",api:"openai-completions",models:[{id:$m,name:($n+" · local"),tools:$tools,input:["text"],cost:{input:0,output:0,cacheRead:0,cacheWrite:0}}]}')
```
injaneity: i believe `tools` is not a supported field in either pi or omp model capability metadata (which uses `supportsTools`). pi will launch fine even with this invalid field attached, but just thought it was worth mentioning

shell/plugins/panels/local-ai/lib/agent.sh
```diff
@@ -0,0 +1,43 @@
...
+    tmp=$(mktemp "$d/.m.XXXXXX") && jq --argjson p "$p" '.providers["omarchy-local"]=$p' <<<"$cur" >"$tmp" && mv "$tmp" "$d/models.json"
+    cur='{}'; [[ -f $d/settings.json ]] && cur=$(<"$d/settings.json")
+    tmp=$(mktemp "$d/.s.XXXXXX") && jq --arg m "$model" 'if .defaultProvider=="omarchy-local" then .defaultModel=$m else . end' <<<"$cur" >"$tmp" && mv "$tmp" "$d/settings.json"
+  done
+}
+unwire() {
+  local d tmp
+  for d in "$HOME_DIR/.pi/agent" "$HOME_DIR/.omp/agent"; do
+    [[ -f $d/models.json ]] && tmp=$(mktemp "$d/.m.XXXXXX") && jq 'del(.providers["omarchy-local"])' "$d/models.json" >"$tmp" && mv "$tmp" "$d/models.json"
+    [[ -f $d/settings.json ]] && tmp=$(mktemp "$d/.s.XXXXXX") && jq 'if .defaultProvider=="omarchy-local" then del(.defaultProvider,.defaultModel) else . end' "$d/settings.json" >"$tmp" && mv "$tmp" "$d/settings.json"
```
injaneity, line 55: current omp gives `models.yml/models.yaml` precedence and treats `models.json` as a one-time migration source. after omp creates yaml, we still update only the ignored json file [reference](https://github.com/can1357/oh-my-pi/blob/33cc6b9a043a74e00a157e72ca909272796d8461/docs/models.md#l13-l24)

0xSero:
Pushed three commits addressing review and a registry drift that had broken the plugin:

**Registry contract** — the registry sharded its discovery index (`registry/index/`), which bricked `scan` on a fresh clone. The plugin now reads `index/recipes.json`, and the registry is pinned by commit (`registry.pin`, currently `21ee5dc`): `scan` checks out that revision instead of trusting whatever `main` says at clone time. This also answers the "external JSON" security concern — the trust root is a revision reviewed with the plugin, moved deliberately. The registry repo now runs this plugin's launch gate in its own CI so the next layout change fails there first.

**Refusals out loud** — the controller owns exactly two registry placeholders (`${MODEL_ROOT}`/`${CACHE_ROOT}` → `~/.cache/omarchy/local-ai/`) and refuses everything else with a visible reason: unknown placeholders, host networking, `--nnodes` multi-node launches, absolute mounts, missing secondary weights (e.g. a speculative draft model, named by path). Refused recipes show as blocked rows in the dashboard instead of silently vanishing. On a 4× RTX Pro 6000, that turns the GLM tp4 recipes from invisible into runnable. The gate also now fails closed on malformed data — a null `revision`/`image` previously slipped past the pin checks (jq `test()` raises on null).

**@injaneity's comments** — both fixed: provider models now write `supportsTools`, and `~/.omp` gets `models.yml` (JSON text, valid YAML — the file omp actually reads, seeded from an existing `models.json`); a hand-written YAML is never touched and the snapshot reports that agent as `manual`.

**Disk + defaults** — catalog rows carry `sizeGb` (shown on the Download action, so nobody starts a 295 GB pull blind), `download` refuses when the filesystem can't fit the weights, `omarchy local ai remove <recipe>` reclaims disk, and `omarchy local ai default` makes the running model the default agent model in one step.

Test suite: 20 → 29 cases, including a real-git pin checkout test and a null-revision gate regression. The manual no longer points into the `shell/` tree.

manual/17-ai.md
```diff
@@ -49,6 +49,20 @@ The _Install > AI_ menu also carries a couple of graphical AI apps: the ChatGPT
...
+
+```bash
+omarchy local ai scan
+omarchy local ai download <recipe>
+omarchy local ai run <recipe>
+omarchy local ai open-agent pi
+omarchy local ai unload
+```
+
+Recipes come from the [local-ai-registry](https://github.com/0xSero/local-ai-registry). Every launchable recipe is `validated`, pins its image by digest, and pins its model revision by commit. See the README under `$OMARCHY_PATH/shell/plugins/panels/local-ai/` for the state machine and safety gates.
```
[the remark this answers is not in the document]
0xSero: Fixed in cf9a6cfa: the manual paragraph no longer references the repo `shell/` tree. State-machine and safety-gate details live in the plugin README; the manual keeps only user-facing usage.

shell/plugins/panels/local-ai/lib/agent.sh
0xSero: Fixed — provider model entries now write `supportsTools` (`lib/agent.sh` `wire()`); no `tools` key remains. Test 11 asserts `supportsTools==false` for a plain recipe and test 21 asserts `==true` after real tool-call acceptance. Thanks for catching it.

shell/plugins/panels/local-ai/lib/agent.sh
0xSero: Fixed — `models_target()` now routes `~/.omp` to `models.yml` (or an existing `models.yaml`), seeding it once from a pre-existing `models.json`, and never touches a hand-written YAML it cannot parse as JSON — the snapshot reports that agent as `manual` instead. Tests 12 and 22 cover both paths. Appreciate the pointer to the precedence docs.

0xSero:
Pushed `ad6f17ba`: trimmed the plugin README from 111 to 57 lines to match the first-party panel convention (the tailscale panel's README is 49) — it duplicated `manual/17-ai.md` and the plugins table. Plugin source now stands at 979 lines plus 278 of tests. All three review threads are addressed at head and resolved above.

For anyone who wants to try this before it lands: the identical code is installable today as a third-party plugin —

```bash
omarchy plugin add https://github.com/0xSero/omarchy-local-ai.git --enable
```

It differs from this PR only by plugin id, shebang style, and sourceDir-vs-PATH CLI resolution, and it's been submitted to the plugin marketplace. Happy to reshape this PR toward a voxtype-style opt-in install of that plugin if that's the preferred 4.1 direction.

0xSero:
Verified end-to-end on a live Omarchy machine (4× GPU: 2× RTX 3090 + 2× Arc Pro B70, a model already serving):

- `omarchy plugin remove` / `plugin add --yes` / `plugin enable` / `bar put` / `plugin update` all clean against the third-party mirror of this exact code.
- The new controller **adopted the already-running accepted container** (state stayed `ready`, API identity intact) across the reinstall — nothing was restarted or re-downloaded.
- Live `scan` checked out the registry at exactly the shipped `registry.pin` (`ee55213`), catalogued 11 matching recipes of 2250, and left the running model `ready`; refused recipes render as `blocked` rows.
- Dashboard render inspected on-screen after `omarchy-restart-shell`: ctx / tok/s / capabilities columns, VRAM rows with Intel usage `unavailable`, sizeGb on the Download action, no clipping or palette leaks.

That pass surfaced one real bug, fixed in `60d855c2`: `primary()` checked download flags before the active check, so when a running model's weights aren't in the managed path (e.g. adopted from an older install), Enter on the running recipe offered a multi-GB download instead of the `Running` no-op. The active check now comes first; verified on-screen.

0xSero:
Redesigned to the direction agreed for 4.1: the dashboard is gone, the popup is the whole UI — recommended model first, one click to load (size shown before anything downloads), one click to unload, right-click or one click to open the Omarchy default agent (`omarchy-default-agent`-aware; Pi/Oh My Pi get the model passed explicitly), and one click to share the endpoint on your tailnet. Net −129 lines while adding those four behaviors; 34 tests; verified on-screen on a live mixed NVIDIA+Intel host. Recommendation comes from a `recommended` flag in the pinned registry, so the five curated models land as registry data, not code.

0xSero:
Closing while we tighten V1 scope (single-GPU, five curated VRAM tiers, automatic harness wiring with restore-on-off, tailscale share). The plugin continues as a third-party plugin; a smaller, focused PR will follow.

sspaeti:
Please update here with the link to the plugin. I would love to try. Thanks for all the work @0xSero 🙏

PR #8839, opened 2026-08-28
Stop the copy-url migration test from fork-bombing the session under a python3 shim

### LINKED ISSUE #8742: test/shell fork-bombs the user session: the copy-url migration test's python3 stub and the mise python3 shim invoke each other unboundedly
`test/shell` fork-bombs the user session on any machine where `python3` resolves to a mise shim. The session's task limit is exhausted, unrelated processes are denied forks, and `mise` dies with `SIGABRT` + a core dump.

Separate from #7360 and #7234: those are `exec` loops in the generated `~/.local/bin` wrappers, so the PID count never grows and they present as a hang. This one **forks**, so it takes the whole session down with it.

### What happened

Running `test/shell` produced a `systemd-coredump` entry:

```
PID: 510864 (python3)   Signal: 6 (ABRT) si_code: SI_TKILL   Executable: /usr/bin/mise
Command Line: /home/…/.local/share/mise/shims/python3 -c '<repair_py>' \
              /tmp/tmp.fxvnt7Qz7a/home/.config/chromium/Default/Preferences <pinned_id> check
```

Expected: the test passes, or fails an assertion. It should not be able to exhaust the session's task limit.

### Mechanism

`test/shell.d/copy-url-shortcut-migration-test.sh:31` captures the interpreter as:

```bash
REAL_PYTHON=$(command -v python3)
```

When python is installed under mise but not *active* in the current directory, `python3` on `PATH` is the mise **shim**, so `REAL_PYTHON` is the shim rather than an interpreter:

```console
$ mise which python3
mise ERROR python3 is a mise bin however it is not currently active. Use `mise use` to activate it in this directory.
```

The test then writes its own `python3` stub into `$stub_bin` and puts that dir first on `PATH` (lines 164-173 and 187-194), and the stub calls `"${REAL_PYTHON}" "$@"`. The shim, having no active mise python, falls back to a plain `PATH` lookup — and finds the stub. So:

> stub → mise shim → `PATH` lookup → stub → mise shim → …

The stub uses no `exec`, so every level **forks and waits**. Each mise process additionally spawns a tokio multi-threaded runtime, multiplying the task count per level. Concurrency grows until the scope's `TasksMax` is hit, `clone()` returns `EAGAIN`, and tokio aborts.

### Evidence

Panic string recovered verbatim from the core:

```
thread 'main' (510864) panicked at /cargo/registry/src/index.crates.io-…/tokio-1.53.1/src/runtime/scheduler/multi_thread/worker.rs:514:13:
OS can't spawn worker thread: Resource temporarily unavailable (os error 11)
```

Process environment, also from the core — the stub dir is first on `PATH`, and `REAL_PYTHON` points back at the shim:

```
HOME=/tmp/tmp.fxvnt7Qz7a/home
PATH=/tmp/tmp.fxvnt7Qz7a/bin:…:/home/…/.local/share/mise/shims:…
REAL_PYTHON=/home/…/.local/share/mise/shims/python3
```

Not memory pressure — 8.3 GiB was available and there were no OOM kills. It is the pids limit:

```console
$ cat /sys/fs/cgroup/user.slice/pids.peak        # 20208
$ cat /sys/fs/cgroup/user.slice/pids.events      # max 40
$ systemctl show --property=DefaultTasksMax      # DefaultTasksMax=19028
```

`pids.events: max 40` means **40 forks were rejected** — the collateral damage lands on unrelated processes in the session, not just on the test.

That the shim re-enters a same-named `PATH` entry is directly demonstrable with a harmless, non-recursive stub:

```console
$ tmpd=$(mktemp -d); mkdir -p "$tmpd/bin"
$ printf '#!/bin/bash\necho "STUB-REACHED $0"\n' > "$tmpd/bin/python3"; chmod +x "$tmpd/bin/python3"
$ PATH="$tmpd/bin:$PATH" ~/.local/share/mise/shims/python3 -c 'pass'
STUB-REACHED /tmp/…/bin/python3
```

### Reproduction

Bounded with `TasksMax` so it cannot take the machine down, and with cores disabled:

```bash
tmpd=$(mktemp -d); mkdir -p "$tmpd/bin" "$tmpd/home"
shim=~/.local/share/mise/shims/python3
printf '#!/bin/bash\n"${REAL_PYTHON}" "$@"\nexit $?\n' > "$tmpd/bin/python3"
chmod +x "$tmpd/bin/python3"

systemd-run --user --scope -q -p TasksMax=60 -- bash -c \
  "ulimit -c 0; export REAL_PYTHON='$shim' HOME='$tmpd/home' PATH='$tmpd/bin:$PATH'; '$shim' -c 'pass'"
```

Result — identical file:line and message to the original crash, exit 134 (128+SIGABRT):

```
thread 'main' panicked at …/tokio-1.53.1/src/runtime/scheduler/multi_thread/worker.rs:514:13:
OS can't spawn worker thread: Resource temporarily unavailable (os error 11)
…/bin/python3: line 2: 541479 Aborted (core dumped) "${REAL_PYTHON}" "$@"
```

### Fix

Resolve through the shim to the actual interpreter, so the stub cannot re-enter a `PATH` lookup:

```bash
REAL_PYTHON=$(python3 -c 'import sys; print(sys.executable)')   # → /usr/bin/python3
```

`command -v` is unsafe here for any name a version manager shims (mise, asdf, pyenv, rbenv): it returns the dispatcher, and pairing it with a same-named stub earlier on `PATH` closes the loop. This is the only occurrence of the pattern in `test/shell.d` — `grep -rn 'command -v' test/shell.d/*.sh | grep -E 'REAL_|ORIG_|ACTUAL_'` returns just this one line.

A cheap belt-and-braces addition would be running `test/shell` under a `TasksMax` cap, so a future instance of this shape fails its own suite instead of the developer's session.

### System

| | |
|---|---|
| Omarchy | 4.0.1-1 (stable) |
| Kernel | 7.1.9-arch1-2 |
| mise | 2026.8.14 linux-x64 (2026-08-26), `mise-bin` 2026.8.14-1 |
| Test/migration | identical to `upstream/quattro`, unmodified |

`omarchy debug --no-sudo --print` is referenced by the crash-diagnosis skill but is not a command in 4.0.1 (`Unknown Omarchy command: omarchy debug`), so no output from it is included rather than substituting something invented.

---

Filed by claude-opus-5 via OpenCode.

ryanrhughes:
Fixes #8742.

`test/shell.d/copy-url-shortcut-migration-test.sh` hangs and spawns an unbounded process chain on any machine where `python3` resolves to a version-manager shim. I hit this running the suite on an unrelated change: it took ~11 minutes to notice, and the chain was hundreds of processes deep across two runs before I killed it. #8742 documents the full blast radius — the session's `TasksMax` is exhausted, unrelated processes are denied forks, and `mise` aborts with a core dump.

## Mechanism

Line 31 captured the interpreter as `REAL_PYTHON=$(command -v python3)`. Where python is installed under a version manager but not active in the current directory, that is the **shim**, not an interpreter. The test then writes its own `python3` stub into `$stub_bin`, puts that directory first on `PATH`, and has the stub delegate with `"${REAL_PYTHON}" "$@"`. The shim resolves `python3` through `PATH` again, finds the stub, and the loop closes:

> stub → shim → `PATH` lookup → stub → shim → …

The stub uses no `exec`, so every level forks and waits.

Reproduced here in isolation, with a depth counter standing in for the unbounded chain:

```console
$ command -v python3
/home/ryan/.local/share/mise/shims/python3
$ # stub on PATH delegating to $REAL_PYTHON=the shim:
stub depth 0
stub depth 1
stub depth 2
stub depth 3
stub depth 4
RECURSION CONFIRMED
```

## The fix

Ask the interpreter where it lives instead of asking `PATH` what `python3` means:

```bash
REAL_PYTHON=$(python3 -c 'import sys; print(sys.executable)')   # /usr/bin/python3
```

`sys.executable` is the real binary, so the delegated call never re-enters a `PATH` lookup and cannot land back on the stub. On a machine with no shim this resolves to the same interpreter `command -v` would have found, so behavior is unchanged there.

This is the same fix proposed in #8742; I reached it independently before finding the issue, which I take as a good sign it is the right one. As that issue notes, it is the only occurrence of the pattern in `test/shell.d` — `command -v` is unsafe for any name a version manager shims (mise, asdf, pyenv, rbenv) when paired with a same-named stub earlier on `PATH`.

I did not implement the issue's optional suggestion of running `test/shell` under a `TasksMax` cap. It is a change to the runner for everyone and depends on `systemd-run` being present, so it seemed like a maintainer call rather than something to fold into this fix.

## Testing

The test previously hung at `migration stays pending when a browser starts mid-repair` and never reached the four assertions after it. With this change all 13 assertions pass and no stray processes are left behind:

```console
$ bash test/shell.d/copy-url-shortcut-migration-test.sh
...
ok - migration stays pending when a browser starts mid-repair
ok - migration stays pending when a briefly-lived browser undoes the repair
ok - migration keeps an unverified repair pending while a browser runs
ok - migration verifies an attempted repair on a browser-free rerun
ok - migration leaves installed third-party extensions alone
$ echo $?
0
```

The two assertions that exercise the stubs both pass, so the stubs are still reached and doing their job — they were simply never able to return before.

### DIFF
--- a/test/shell.d/copy-url-shortcut-migration-test.sh
+++ b/test/shell.d/copy-url-shortcut-migration-test.sh
@@ -28,7 +28,13 @@ write_stale_preferences() {
 stub_bin="$test_dir/bin"
 mkdir -p "$stub_bin"
 
-REAL_PYTHON=$(command -v python3)
+# The stubs below shadow python3 on PATH and delegate to this interpreter, so
+# it has to be the interpreter itself rather than whatever `command -v` finds.
+# Where python3 is a version-manager shim (mise, asdf, pyenv), the shim resolves
+# python3 through PATH again, lands back on the stub, and recurses without
+# bound. sys.executable is always the real binary.
+REAL_PYTHON=$(python3 -c 'import sys; print(sys.executable)')
+[[ -x $REAL_PYTHON ]] || fail "python3 reports the interpreter behind any shim"
 export REAL_PYTHON
 
 run_migration() {


ryanrhughes:
Closing as a duplicate of #8268, which predates this by three days and is now merged (f0672772). I opened this without spotting the existing PR — my mistake.

For the record, since I had the affected environment in front of me: I verified #8268's fix on a machine where `python3` is a mise shim, which is where this reproduces. All 13 assertions pass and no processes are left behind. Before the fix the test hung at `migration stays pending when a browser starts mid-repair` and never reached the four assertions after it.

`command -p -v python3` resolves to `/bin/python3` there, correctly bypassing `~/.local/share/mise/shims/python3`.

I ran that verification under `systemd-run --user --scope -p TasksMax=300`, which is worth mentioning because of how this fails: the stub and the shim fork into each other without bound, so an unfixed checkout exhausts the session's task limit and denies forks to unrelated processes. #8742 has the full forensics. Anyone testing a change to this file is safer doing it under a task cap.

That failure mode is also the argument for the belt-and-braces suggestion in #8742 — running `test/shell` under a `TasksMax` cap so a future instance of this shape fails its own suite instead of the developer's session. I have not opened anything for that; it changes the runner for everyone and depends on `systemd-run`, so it seemed like a maintainer call.

Issue #7021, opened 2026-08-15
GDK_SCALE=2 + force_zero_scaling overflows XWayland/Java windows on auto-scaled 1080p


v-t-r-gg:
### System details

Omarchy 4.0.0-1 (Quattro), Hyprland 0.56.2, Xwayland 24.1.13  
Samsung Galaxy Book NP950QED, Intel i7-1260P, Iris Xe  
eDP-1 `1920x1080` @ 142 DPI (340×190 mm). Packaged `scale = "auto"` landed on **1.5**, then later recomputed as **2** after Hyprland reloads.

### What's wrong?

Packaged Quattro defaults put three scale layers in disagreement. This is a default-config bug, not an app-specific one. Native Wayland GTK/Qt is mostly fine.

**Defaults involved**

```lua
-- config/hypr/monitors.lua
local omarchy_gdk_scale = 2
local omarchy_monitor_scale = "auto"
hl.env("GDK_SCALE", tostring(omarchy_gdk_scale))
hl.monitor({ output = "", mode = "preferred", position = "auto", scale = omarchy_monitor_scale })
```

```lua
-- default/hypr/envs.lua
xwayland = { force_zero_scaling = true }
```

On this 1080p HiDPI laptop:

| Layer | Size the client is told |
|---|---|
| Hyprland layout | Logical `1280×720` when auto is 1.5 (`1920 / 1.5`) |
| XWayland (`force_zero_scaling`) | Physical `1920×1080`, 1:1 pixels |
| Session env | `GDK_SCALE=2` for every child |

X11 and Java AWT do not use the Wayland scale path. They see a 1920-wide X screen **and** honor `GDK_SCALE=2` (and sometimes `sun.java2d.uiScale`).

**Measured effect**

- A client whose 1× minimum width is 765 advertised `WM_NORMAL_HINTS` min width **1530** (`765 × 2`).
- Extra in-window chrome (sidebar, inspector, plugin panel) adds more width.
- The requested size exceeds 1920.
- A tiling compositor will not shift the toplevel left the way Plasma does, so the extra pixels hang off the right edge of the panel.

Same class of clients: Swing/AWT IDEs and tools, GTK forced onto X11, Flatpaks that inherit the session env, some Wine helpers.

[...]

PR #8841, opened 2026-08-28
Add opt-in toggle to scale X11 apps that ignore HiDPI hints

avillagran:
## Problem

On a scaled monitor, X11 apps that ignore HiDPI hints render at half size.

Omarchy sets `xwayland:force_zero_scaling`, which hands X11 clients an unscaled
surface and expects the toolkit to scale itself via `GDK_SCALE`/`QT_*`. GTK, Qt and
Electron apps do exactly that and stay perfectly crisp — that's why this default is
the right one.

But apps that ignore those hints just render at 1x. On a 2x monitor their text and
controls come out half size. RealVNC Viewer is the common case (reported on a
3456x2234 MacBook display at scale 2), along with older Java/Motif/Tk apps and some
installers.

Measured on eDP-1 @ 2x with Hyprland 0.56.1: RealVNC Viewer reports 852x530 with the
current default, versus 1714x1071 once the compositor scales it.

## Fix

Add an `xwayland-scaling` toggle that drops `force_zero_scaling`, so the compositor
upscales those windows to the monitor scale instead.

- `default/hypr/toggles/xwayland-scaling.lua` — the flag
- `bin/omarchy-hyprland-xwayland-scaling-toggle` — wrapper, matching
  `omarchy-hyprland-window-gaps-toggle`
- menu entry under _Trigger > Toggle_ as "X11 App Scaling"
- manual section in `manual/33-monitors.md`

It reuses the existing `omarchy-hyprland-toggle` flag mechanism, so it's opt-in,
reversible, and applies live via `hyprctl reload` — no relogin needed.

The toggle notifies on both transitions (matching
`omarchy-hyprland-window-single-square-aspect-toggle`), because X11 clients only pick
up the new scale when restarted — without a hint that's easy to read as "the toggle did
nothing". The menu row also carries the usual `✓` while it's on, driven by the existing
`omarchy-hyprland-toggle-enabled` predicate through the standard `checked:` field.

## Why off by default

`force_zero_scaling` is a global option — Hyprland has no per-window rule for it. X11
apps that already scale themselves correctly look slightly softer when it's off, so
enabling this for everyone would trade crispness across the whole desktop to fix a
handful of apps. Opt-in keeps the good default and gives affected users a switch.

## Mixed-scale monitors

Scaling is resolved per monitor, not globally, so a 2x + 1x setup stays correct.
`CWindow::realToReportSize` / `xwaylandSizeToReal` use `PMONITOR->m_scale` — the scale
of the monitor each window is currently on — and windows re-scale when moved between
monitors.

Verified with a headless 1x output alongside a 2x eDP-1: two RealVNC windows under the
same toggle reported 1714x1071 on the 2x monitor and 948x1040 on the 1x monitor, and a
window re-scaled when dispatched across.

## Relationship to #7021

#7021 is the opposite failure mode on the same option: apps that *do* honor `GDK_SCALE`
get oversized when the auto monitor scale and `GDK_SCALE` disagree. This PR does not
attempt to fix that, and does not change any default — it only adds an opt-in escape
hatch for clients that ignore the hints entirely. The two are complementary.

## Verification

- Toggle on → `xwayland:force_zero_scaling = false`; off → `true`; flag file created
  and removed correctly.
- Notifications confirmed on both transitions in the running UI, glyph rendering
  correctly.
- `checked:` verified through `MenuModel.labelFor`, which yields "X11 App Scaling ✓"
  when enabled and the bare label when not; `omarchy-hyprland-toggle-enabled
  xwayland-scaling` returns the matching exit status in both states.
- RealVNC Viewer confirmed correct size in the running UI, and the menu toggle applied
  instantly (slightly softer text, as expected).
- Menu entry parses with the repo's own `MenuModel.js parseMenuJsonc`, producing the
  same shape as its siblings.
- `./test/cli` passes with 0 failures.
- `./test/shell` shows 4 failures, all reproduced identically on a clean unmodified
  clone — pre-existing and unrelated (missing `omarchy-pkgs` checkout x3, plus a
  network-dependent `omarchy-theme-install` check).

Tested on Hyprland 0.56.1, aarch64 (Asahi), scale 2 display.

### DIFF
--- a/bin/omarchy-hyprland-xwayland-scaling-toggle
+++ b/bin/omarchy-hyprland-xwayland-scaling-toggle
@@ -1,0 +1,11 @@
+#!/bin/bash
+
+# omarchy:summary=Toggle compositor upscaling for X11 apps that ignore HiDPI scaling hints.
+
+# nf-md-monitor, escaped so this file reads without a Nerd Font.
+readonly MONITOR_GLYPH=$'\U000f0379'
+
+case $(omarchy-hyprland-toggle xwayland-scaling) in
+  on) omarchy-notification-send -g "$MONITOR_GLYPH" "X11 app scaling enabled" "Restart X11 apps to resize them" ;;
+  off) omarchy-notification-send -g "$MONITOR_GLYPH" "X11 app scaling disabled" "Restart X11 apps to resize them" ;;
+esac
--- a/default/hypr/toggles/xwayland-scaling.lua
+++ b/default/hypr/toggles/xwayland-scaling.lua
@@ -1,0 +1,16 @@
+-- Let Hyprland upscale X11 (XWayland) apps to match the monitor scale.
+--
+-- By default Omarchy sets xwayland:force_zero_scaling, which hands X11 clients
+-- an unscaled surface. Toolkit-aware apps (GTK/Qt/Electron) then scale
+-- themselves via GDK_SCALE/QT_* and stay crisp. Apps that ignore those hints
+-- render at 1x instead, so on a 2x monitor their text and controls come out
+-- half size (RealVNC Viewer, older Java/Motif/Tk apps, some installers).
+--
+-- Turning this flag on drops force_zero_scaling so the compositor upscales
+-- those windows to the monitor scale: correct apparent size, at the cost of
+-- slightly softer text on X11 apps that were already scaling themselves.
+hl.config({
+  xwayland = {
+    force_zero_scaling = false,
+  },
+})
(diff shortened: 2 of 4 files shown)


avillagran:
Good catch on the Markdown style — fixed. The three new paragraphs in `manual/33-monitors.md` are now single full lines, per `AGENTS.md` ("write full lines — no hard wrapping at 80 columns"), matching the surrounding prose in that file.

Also aligned `omarchy:summary` to the imperative form used by the other toggles ("Toggle …" rather than "Toggles …").

`./test/all` passes apart from four failures that reproduce identically on an unmodified clone (missing `omarchy-pkgs` checkout x3, plus a network-dependent `omarchy-theme-install` check).

PR #8861, opened 2026-08-29
Power panel: system vitals meters and per-process memory

vinceferro:
> **UPDATE 2 — grayscale by load.** The panel left the color palette
> entirely: every surface speaks foreground-at-load. The system block/row
> leads in white (the machine itself); a process's ink dims with its share
> of the bar's busiest process (0.35..1.0), in the table marks, the metric
> cells, and every composition bar segment. Identity is positional
> (biggest-to-lightest, mirroring the table) and by label; brightness is
> magnitude. This matches the shell's native monochrome language — and
> makes the panel theme-proof by construction (no theme can clash with
> ink). Suite unchanged at 138 asserts.

> **UPDATE — operator review round (latest).** After living with the panel,
> three simplifications landed on this branch (see the newest commit):
>
> 1. **The system-draw watts meter is gone.** It duplicated the stock
>    battery pill's number. The per-process attribution stays; the Color
>    surface loses its `green`/`yellow` keys (no remaining consumers).
> 2. **The composition bars mirror the table.** The shade lattice retired —
>    the table row's mark is the single color authority, bars render the
>    same hue at full strength, segments ordered biggest-to-lightest with
>    the system block leading. Canonical hue order is gone with the lattice.
> 3. **"base load" and "everything else" are one row now: System** (the
>    idle floor plus the unattributed tail), first in the table and on
>    every bar. The old floor-vs-tail distinction read as a mystery second
>    "system" — the one-reality invariant (System + processes = whole) is
>    unchanged and still fuzz-tested.
>
> Round-by-round notes below describe earlier states; where they mention
> the Draw meter, lattice shading, or canonical hue order, this update
> supersedes them. Suite now 138 asserts (the lattice's test machinery
> retired with it).

Stacked on #8860 (power-hungry): this PR's own delta is `git diff 76e0269c power-panel-vitals` — +1041/−30, the four files below. The first commits in the diff list belong to #8860.

### The problem

The Power Hungry panel answers "who is eating my battery". The companion
questions — "how hard is the machine working" and "how much memory does
the culprit hold" — still require other tools. This adds a compact SYSTEM
section with meters, and a per-process memory column on the attribution
rows, all riding PR #1's sampler so nothing can disagree.

### What this adds

- **Sampler keys**: `cpu_busy` + `cpu_total` (busy jiffies, and busy+idle
  ticks — see the two-denominators note), `mem_total_kb`/`mem_avail_kb`
  from `/proc/meminfo`, `gpu_pct` from the first user-readable GPU source,
  and per-pid RSS on the existing `p` lines from `/proc/<pid>/statm`.
- **`Model.buildSystemRows(prev, next, drawWatts)`**: pure, node-tested,
  returns `{label, value, meter}` rows in fixed order — CPU, RAM, GPU,
  Draw. `meter` is 0..1 for a progress bar or −1 when the row has no bar
  (a dash). CPU shows the stock `—` until a second snapshot exists; GPU
  appears only when a source was found; Draw appears only while
  discharging, meter scaled so the 20 W/40 W thresholds land at half and
  full bar.
- **`buildTopProcesses` rows carry metric cells** (CPU, RAM, W while
  discharging — one mini-meter per column in a fixed grid; the old
  joined-string payload was dead data and is gone): per-comm RAM
  aggregated exactly like jiffies, rendered as a RAM cell beside the CPU
  share.
- **Panel**: a SYSTEM section drawn with the battery progress bar's own
  track-and-fill idiom (rounded track at foreground alpha, foreground
  fill) scaled to a slim meter — no new widget, no new Ui component.
  Fills snap at the 1 Hz cadence: no width `Behavior`s anywhere in this
  feature (round 5 removed them; the stock battery bar's own animations
  are stock and untouched). The Draw meter's fill grades green under
  20 W, yellow under 40, red beyond, via the Color singleton's newly
  surfaced theme `green`/`yellow` keys (all themes define them; the
  mapping follows the existing `urgent`←red idiom). This meter is what
  consumes those keys, which is why the 6-line Color.qml change lives
  in this PR.

### Round 12: system-baseline-first, size-descending blocks

The reviewer's final ordering language, superseding round 11's
bar-follows-table correspondence: each composition bar now leads with a
**SYSTEM block** — the table's system-anchor concept made spatial, in
the same foreground treatment (cross-surface concept reuse: the anchor
row IS the machine, the leading block IS the machine) — then lays its
process blocks **largest-to-smallest by the bar's own metric**, then the
idle/available frame. The SYSTEM block measures what the old rest
segment measured: unattributed busy on the CPU bar, used memory beyond
the table set's RSS on the RAM bar. The W bar is unchanged — base-first
already satisfies the mandate, its else tail being a different concept.
The table keeps its single shared rank; per-bar order may differ from it
and between bars **by design** — bars are self-sorted visualizations,
the table is the ranking. Round 11's lasting contributions survive:
the one visible set, one ranking basis for the TABLE, and the death of
the old RSS-ranked membership. Lattice colors, separators, gap budget,
snap fills, no-text rules untouched — order only. Verified live on the
discharging path: the CPU bar's leading run is the foreground SYSTEM
block, followed by size-descending lattice-shaded process segments.

### Round 11: bars follow the process list order

**Superseded by round 12** (kept for the decision record): the round-11
correspondence asserts were REPLACED by round-12 size-descending,
SYSTEM-first asserts. On the CPU bar the two coincide — table rank IS
size there, which is why the ×50 permutation fuzz still passes — but the
RAM bar now self-sorts by RSS and legitimately diverges from table order.

The reviewer's decision: **correspondence over positional calm** — all
three composition bars lay their process segments out left-to-right
exactly as the table reads top-to-bottom, on the table's ONE shared
ranking basis (CPU busy-share). The RAM bar's historical divergence —
its own top-N ranked by RSS — is removed here: it shows the table's
comms with their RSS widths, the rest segment absorbing used memory the
table set doesn't cover. Base stays first on the watts bar, else/rest
follow the ranked processes, idle/available frame last; lattice colors,
separators, gap budget, snap fills, and no-text rules untouched — order
only. **The tradeoff, stated honestly**: round 9 introduced canonical
hue-major order because round 8's rank-coupled group order relocated
whole hue blocks — the right fix with the tool it had. With per-comm
lattice-colored segments that disease no longer exists: a rank swap now
moves exactly the swapped segments — discrete, meaningful motion, with
identity preserved by the stable colors. Positional stability was
sacrificed for list-correspondence, by reviewer choice. Correspondence
is asserted in tests (segment sequence == table rank sequence, including
under rank permutation, ×50 fuzz — carried into round 12 only in this
coincidental CPU-bar form; the round-12 SYSTEM-first/size-descending
asserts are the governing ones) and spot-checked live: the
bar's leading hue sequence matched the table's top accent sequence
(cyan, magenta = the two dominant comms); deeper per-segment pixel
extraction is unreliable at small widths (shade × intensity compositing
at 1 Hz) and left to the fuzzed Model guarantee.

### Round 10: the shade lattice, deterministic claim resolution

The reviewer's verdict on three hues: too much collision. Identity is
now a **nine-slot lattice** — the collision-free trio × three fixed
opacity steps (**0.5 / 0.75 / 1.0**; the lowest step stays clearly above
the panel's 0.12-alpha track at panel scale — verified live on the
validation theme, where four distinct slots rendered simultaneously in
both the table accents and the bars). Zero new RGB; the 22-theme
guarantee is intact. Both coordinates derive from the one FNV-1a hash:
**hue from the low bits, shade from the next bits**, decorrelation
tested over a realistic corpus.

**Assignment is deterministic claim resolution** over the visible set:
each comm's nine-slot preference list starts at its hash slot, walks the
SAME hue's other shades first (hue identity survives displacement; only
the shade yields), then the other hues in hash-derived cyclic order.
Claims process in claim-strength order — rank first, name lexicographic
— each taking its first free slot; greedy-by-strength IS the
displacement fixed point in a single pass, so termination is
structural. **Stability contract, stated honestly**: an uncontested
comm keeps its hash slot forever; a contested comm's displacement is
deterministic — top ranks are the most stable, and shades yield before
hues; assignment depends only on the visible set and its rank (fuzzed
×100: deterministic, unique slots, zero unassigned below exhaustion).
This supersedes round 8's "three hues, accepted collisions" language:
collisions are now resolved, not accepted. Badges are demoted to the
exhaustion path — more than nine visible comms — which a top-N table
practically never reaches; the round-9 fixed badge slot stays for that
path. One function over one visible set feeds every surface (accents at
their shade, cell fills with intensity × shade composed
multiplicatively, and bar segments), whose order becomes lattice order —
hue-major, shade-minor — making bar layout fully rank-independent below
exhaustion: segment order now changes only with membership. **Superseded
twice over**: round 11 coupled bar order back to table rank; round 12
settled the final semantics — bars self-sort (SYSTEM block first, then
size-descending by each bar's own metric), so order tracks per-bar size —
neither rank-independent nor table-coupled.

### Round 9: restored bar geometry, canonical order, badge slot discipline

Round 8 bought stable hues at two costs The reviewer caught ("lost a bit
of the geometries on the bars and horizontal space usage calmness").
Both reversed, hues kept. **Un-merge**: composition bars render
per-comm segments again, each in its stable hashed hue, separated by
constant track-colored gaps — the geometry reads through boundaries even
between same-hue neighbors. Gap width is a fixed Style constant; the
total budget comes from an exported `SPLIT_MAX_SEGMENTS` (top-N + rest +
idle; the watts bar's base + top-N + else fits the same count), so fills
scale identically across refreshes and resources regardless of how many
segments are visible. **Canonical order**: blue-block comms first, then
cyan, then magenta (rank only within a hue), then rest, then
idle/available — rank shuffles reorder members inside their hue block
and shares breathe, but hue blocks never relocate; verified by fuzzing
rank permutations and by a staged live shuffle (a burner process
reordered the top ranks: the RAM bar's hue sequence was pixel-stable;
the CPU bar's magenta block legitimately EMPTIED when its members left
the top-N — membership churn, not relocation, and canonical order never
inverted). **Badge slot**: the collision numeral renders in a fixed
reserved micro-slot after the accent mark — empty when unbadged — so
badges appearing or clearing mid-refresh cannot shift the line; the
label start is a constant offset. Steady-state motion re-certified
(round-5 proof repeated): text-zone diffs are content ticks (values
change at 1 Hz by design; positions are fixed right-aligned slots),
fills step ~0.7% of the block.

### Round 8: name-hashed stable identity colors

Colors followed table rank, so every rank shuffle swapped hues — nothing
learnable. Identity is now the process NAME: **FNV-1a** hashed into the
collision-free palette — chosen for exactly this job (short, well-known,
good distribution over short strings; no randomness, no seed state; the
multiply stays under 2^53 with an explicit uint32 fold so every QML JS
engine agrees bit-for-bit). Same comm = same hue everywhere and forever.
Verified live: a comm's accent mark is pixel-identical across refreshes
that shuffled the table's ranking.

Collisions get both reviewer strategies on their best surfaces. **Table
rows → ordinal badges**: among the visible set, the first member of a
hue stays clean and later members carry a tiny numeral in the gutter —
plain digits rather than superscript glyphs, because glyph coverage at
the panel's font size is a gamble plain digits never take; the hue is
identity, the badge only breaks simultaneous ties, and a lone member
never carries one. **Composition bars → grouped segments**: same-hue
comms merge into one segment per hue (share = sum of members, group
order = first appearance by rank; base/rest/idle/avail/else untouched) —
the bar honestly shows what color can distinguish while the table below
names the members. Grouped bars still sum to their whole (fuzz ×50).
Non-process rows (system anchor foreground, base load, everything-else
muted) never join the hash palette. The order-based assignColorKeys is
retired under the same signature; distribution over a 15-comm realistic
corpus spreads across all three hues with no pathological clustering.

### Round 7: one CPU denominator, system anchor row

Operator finding: "those boxes seem to be illustrating two different
realities" — process CPU cells measured share-of-work (busy jiffies,
>100% possible across cores) while the global CPU bar measured
share-of-capacity. Both honest, mutually unreconcilable. Process CPU
cells now use the single denominator the global bar and the CPU
composition bar already use: jiffies over ALL ticks including idle.
**Process cells plus the everything-else row now sum to the global CPU%
by construction** (fuzz-verified ×200). The multicore >100% honesty note
moves to this section: the old number was truthful about cross-core
jiffies; the new number is truthful about machine capacity — a semantic
switch, documented rather than silently dropped. Ranking is unchanged
(share-of-busy and share-of-total order identically within a snapshot;
the busy fraction is a common factor — asserted by test).

A **`system` anchor row** leads the table: the machine itself in theme
foreground (neutral — never literal white, no hardcoded RGB),
full-intensity cells on the same grid and scale — global CPU%, RAM
used%, and total draw while the watts basis is active (its W cell fills
the column: the draw IS the whole the rows below decompose). The
everything-else row becomes a whole-machine remainder row — CPU
remainder closing the sum to global CPU%, used RAM beyond the listed
processes' RSS (clamped at zero; shared-page double counting is the
standing disclosure), and the watts tail — and its existence is no
longer gated on watts display-noise alone. RAM now visibly reconciles:
the system row shows the whole, the rows show parts, and the gap is the
shared/kernel memory the draft has always disclosed. **Naming**: the
global row is `system`; the attribution floor row is renamed `base
load` — the two "system" concepts cannot be confused. The CPU
composition bar already shared the denominator and is unchanged; the
anchor-vs-rows reconciliation is asserted in tests (CPU exact; RAM with
the ≥ gap; W exact in the model — the DISPLAYED numbers round: whole
watts per cell and the anchor at ≥10 W, one decimal below, plus the
else-row's ≤0.5 W tail gate, so the visible sum can drift up to ~1.4 W
worst observed, >0.4 W on 18% of fuzz draws — a display bound, not a
model error) and spot-checked in the live capture
(system 32% ≈ 13.8% + 1.8% + 1.5% + remainder).

### Final round: the watts basis keys on battery state, not plug state

Earlier rounds keyed the W column, the Draw row, and the watts
attribution on `UPower.onBattery` — the daemon's plug-side composite. In
the marginal-adapter wedge (adapter latched online in UPower while
delivering nothing, battery actually draining) that wrongly took the AC
path and hid watts while the machine ran on battery. The basis is now
the battery device's own state: watts render whenever
`displayDevice.state === Discharging` — there the battery is the power
source, the sampled discharge rate is real system draw, and attribution
is valid, wedge included. Charging, full, and pending keep per-process
watts hidden (adapter input unreadable — physically unattributable).
The rule is one pure function (`Model.wattsBasis`, unit-tested including
the wedge and state flapping); the stock panel's own mode language (hero
label, stats rows, profile picker) still speaks plug-state and is
untouched. **The earlier caveat "drain-under-external-power takes the AC
path by design" is obsolete** — superseded by this keying. The wedge
cannot be staged on demand; it is unit-proven and the normal
charging/full paths are live-verified (W column correctly absent).

### Viz round 5: fixed tabular grid, static text, snap fills

Operator round-4 finding: "a bit much of moving horizontally." At the
1 Hz refresh, fill-anchored value text rode the fill edge, cells flexed
as widths recomputed, and the panel never sat still. The POWER HUNGRY
section now sits on a fixed grid, recalculated only on basis change
(AC↔DC column set — the one allowed motion) or panel width: the comm
column is sized once for the kernel's 15-char comm cap
(`Model.COMM_MAX_CHARS`, measured via TextMetrics, elided beyond) with
the identity accent mark in its own fixed slot; metric cells split the
remaining track equally (no metric starves at panel width — two columns
get ~200px, three ~135px, against values of ~40–55px); header labels
right-align into the same column slots as the values — the last column
lands exactly, intermediate labels sit within ~4px at scale 1 (known
residual, deliberately left). All value text is
right-aligned inside its fixed cell — same pixel column every refresh
regardless of fill length — and fills are capped short of the text so
type never sits on a fill. No width Behaviors on any fill in this
feature; fills step at the cadence, geometry never moves. The SYSTEM
composition bars lose their in-segment comm names (positions coupled to
segment boundaries = motion); their values were already right-anchored
and their segments already snap. The stock battery bar's animations are
stock and untouched. The motion proof: two captures 1.5 s apart with the
panel live — the value-text zone differs by 0.06% (antialiasing noise;
fill-anchored text would move ~40px), fills step ~1% of the block, and
the comm column's 2.3% is rank-reshuffle content change (which comm sits
in a row), not geometry.

### Viz round 4: one graphic line per process, metric cells

Round 3 drew two graphic lines per process (impact fill + RAM sub-bar) and
The reviewer read it as noise. Each POWER HUNGRY row is now ONE track of
metric bar-cells beneath fixed column headers (CPU, RAM, W while
discharging), so the layout is learned once and every process is a single
line. Cells use metric-type colors — CPU blue, RAM cyan, W magenta from
the collision-free set; W keeps magenta rather than categorical grading
because the green/yellow/red thresholds grade SYSTEM draw (the Draw bar),
and per-process watts carry no honest absolute thresholds. Normalization
per metric: CPU = process share clamped for display while the number
tells the multicore truth (200% possible); RAM = comm RSS / MemTotal;
W = attributed watts / measured draw; each cell's intensity ramps within
its own column (0.35 floor at zero, 1.0 at the column's max, W normalized
against the column's largest cell whichever row owns it). Comm identity
moves to a small accent mark on the row label — some cross-referencing to
the SYSTEM composition bars survives; full per-comm cell coloring was the
price of metric-colored columns, accepted per the clarity mandate. GPU is
a reserved column (accent) that never renders on driver-silent hardware;
the cell structure lights up when a future fdinfo builder supplies it.
Zero-state handling: zero-RSS comms omit the RAM cell; AC rows never
carry W cells; the column set derives from the rows themselves.

### Viz round 3: bars exclusively, in-bar numbers, intensity semantics

Plain text rows are gone. Every POWER HUNGRY row is now a full-width bar:
comm label in the left gutter, the comm-colored fill growing from the
gutter sized by the row's impact and faded by an intensity ramp (top row
full opacity, tail rows toward a 0.35 floor, normalized by the top share),
the value text anchored just past the fill's end, and a thin RSS sub-bar
beneath in the same color. SYSTEM bars put labels in the gutter and values
at the right edge, ramp their intensity with utilization (0.45 idle → 1.0
full — criticality as intensity of existing semantic colors; the
categorical green/yellow/red stays reserved for watts), and wide segments
carry their comm's name in whichever theme color contrasts more with the
fill.

**Text placement decision:** all text renders on the unfilled track in the
panel foreground — the same surface every stock panel row already proves
legible on all 22 themes. In-fill text contrast cannot be guaranteed
theme-wide (light themes put light backgrounds under pastel fills) without
hardcoded RGB, so the escape hatch is taken deliberately; wide segments'
in-fill names use a luminance pick between the theme's own foreground and
background — adaptive, still zero hardcoded RGB.

**Parity-launch note (ops):** running fork code with the stock
`OMARCHY_PATH` in its environment does NOT keep keybinds alive during
test mounts — `qs ipc -p <path>` (what `omarchy-shell` and therefore every
bind executes) keys the shell instance by its `-p` config path, and the
env var never enters instance matching. Verified empirically: a fork-path
shell with stock env answers fork-path pings and stays invisible to the
bind path. Test mounts must use the plain swap with the binds-dead caveat.

### Viz round 2: impact meters and resource splits

Every POWER HUNGRY row gains an **impact meter** — the share that ranks it
(CPU share on external power; while discharging the attributed watts, which
hold the same proportion by construction, so the fill matches both numbers
the row displays). The base and tail rows meter their slice of the whole
draw. Meters reuse the vitals row component with the row's comm color, so
each row doubles as its own legend entry for the bars.

The SYSTEM meters become **composition bars**: CPU segments = top comms +
other busy + idle (shares of all ticks, summing to exactly one, idle
rendered as the unfilled track); RAM = top comms by RSS + other used +
available (of MemTotal); GPU renders only when a source exists. While
discharging, POWER HUNGRY additionally shows **the attribution as one
bar** — base floor, top processes, tail — summing to the measured draw:
the model made visible, anchored directly above the rows that decompose
it.

**Comm colors are theme-semantic and stable across every surface.** The
palette is the three non-threshold hues every theme defines and no other
key collides with — blue, cyan, magenta (inventoried across all 22:
`accent` equals `blue` on 17 themes and `magenta` on lumon; orange/brown
exist on 19/22; and the `bright_*` variants equal their base hue on 54%
of theme-key pairs, so no six-key collision-free set exists).
They are assigned deterministically by top-process order through one
resolver (`segmentColor`) feeding bars and row meters alike, so a comm
cannot change color between surfaces; `accent` is demoted to the watts
base-segment only. Beyond three distinct comms the palette cycles and two
apps can share a hue — an accepted limit of theme palettes (retro-82
collapses blue==magenta within the palette by design), disambiguated by
row order, chosen over hardcoded RGB or a same-hue opacity ladder. The
watts base-segment rides `accent`, which shares its hue with the blue
comm on those 17 themes (magenta on lumon) — a pre-existing residual
this palette strictly reduces. green/yellow stay reserved
for Draw grading.

Split lists sum to their whole **by construction**: no display-noise
thresholding (a sub-pixel segment renders as nothing, which is
invisibility enough), fuzz-tested ×200 per resource. The one honest
exception: RAM sums above one when per-process RSS double-counting
exceeds the used slice (shared pages counted per process) — the bar
visibly overflows rather than being rescaled into a lie; asserted in
tests as expected behavior.

Segment delegates bind widths through ids, never `parent`: under this
Quickshell a Repeater delegate's `parent` resolves the zero-width
Repeater during construction and the binding never re-fires (found live,
fixed, and why the watts bar worked while the nested rows didn't).

### Why two CPU denominators

PR #1's `cputotal` deliberately excludes idle and iowait: per-process
shares must be shares of CPU *activity*, or idle time would surface as
"everything else" watts. System CPU% wants busy over *all* ticks. The
vitals pair carries its own all-ticks total, `cpu_busy` repeats the busy
value under a self-describing name, and `cputotal` stays untouched as the
attribution key. A maintainer squashing both PRs may drop the alias.

### Per-process memory semantics (the honest caveats)

- RSS comes from `/proc/<pid>/statm` field 2 × `getconf PAGESIZE` — never
  a hardcoded page size (Asahi runs 16K pages; assuming 4K would misread
  RAM 4×). Verified live: sampler-reported RSS matches statm exactly.
- **RSS sums double-count shared pages.** Ten processes sharing one libc
  each count it; the per-comm column is a size hint, not an addition that
  should sum to system RAM used. This is inherent to per-process memory
  accounting (the same caveat top/htop carry) and stated rather than
  hidden.
- Aggregated by comm like CPU share, from the newest snapshot (memory is
  a level, not a counter — no diffing).

### Sampler performance and robustness

- Full run measures **50–53 ms** (p50 51 ms) with statm included — the
  per-pid statm reads add ~2–5 ms over PR #1's sampler because /proc is
  page-cache hot; the 100 ms gate clears with headroom.
- The statm read uses `getline` per pid *inside* the existing awk pass.
  A naive multi-file awk over `/proc/[0-9]*/statm` dies **fatally** the
  first time a pid exits between glob expansion and open — under normal
  process churn that produced intermittent empty or truncated snapshots
  (reproduced, then fixed: `getline` on a vanished file returns −1 and
  the row simply reports rss 0).

### GPU honesty

The GPU row appears only when the sampler found a user-readable
utilization source. On the validation machine (Asahi) there is none — no
`/sys/class/devfreq/` entries, no `gpu_busy_percent` (AMD-only), no
nvidia-smi, and debugfs GPU nodes are root-only (`drwx------`),
disqualified for a user-space 1 Hz sampler. Absence renders as absence.

**Follow-up path (for whoever has the hardware):** per-process GPU time
is readable on amdgpu/i915 via DRM engine fdinfo —
`/proc/<pid>/fdinfo/<fd>` lines like `drm-engine-gpu: 12345678 ns` for
each DRM fd the process holds (what `amdgpu_top` and Hyprland's own
renderer stats use). On Asahi the driver does not expose engine usage
through fdinfo (checked 2026-08-27: Hyprland's fdinfo on this box carries
buffer statistics only), so per-process GPU is omitted rather than
invented. A future sampler pass could emit `p\t…\tgpu_ns` on hardware
that provides it.

### Cross-platform behavior

- **amdgpu**: `/sys/class/drm/card*/device/gpu_busy_percent`, first
  readable value wins — GPU row appears. (Shape per driver docs; not
  live-tested — no AMD hardware at hand; flagged for reviewers.)
- **NVIDIA**: `nvidia-smi --query-gpu=utilization.gpu` only when already
  on PATH, behind `timeout 0.05`, output validated numerically.
- **Asahi / plain Intel integrated**: no user-readable source → row
  omitted (live-verified absence).
- **RAM/CPU/Draw**: `/proc`-only; MemAvailable with a torn-read clamp so
  rows can never show negative or >100%.

### Testing

- Node harness: **156/156 green** across 11 suites — attribution
  invariant (fuzz ×300, ×200 with RAM), vitals, metric cells, watts
  basis, anchor reconciliation (fuzz ×200: CPU rows sum exactly to the
  global, RAM sums-or-exceeds, W whole in the model), lattice claim
  resolution (fuzz ×100), bar sums and permutation (fuzz ×50), and the
  round-12 SYSTEM-first/size-descending asserts. The suite grew
  86 → 105 → 177 through the viz rounds; the post-review cleanup retired
  the 22 asserts that targeted since-removed internals (buildRowImpact,
  assignColorKeys, collisionOrdinals, the joined value payload — those
  invariants live on through the shipped cell/lattice APIs) and added
  one aggregateCommShares share assert: 156. **Canonical home: the
  development workshop repo** (`upstream/tests/power-panel-test.js`,
  with `r12-smoke.js` beside it) — this PR links there rather than
  shipping test files into core.
- Viz round 3 live render (discharging window, ~35 W, theme `ai-engineer`):
  bars-as-rows with in-bar labels and value-at-fill-end confirmed by OCR;
  the intensity ramp confirmed by pixel sampling (full-opacity top fill
  vs a ~0.35-opacity tail fill of the same hue); comm colors and the
  yellow Draw grading intact; journal clean. The RSS sub-bar rides the
  same rendering path as the main fills (pixel-level confirmation at the
  3px scale is ambiguous and noted honestly).
- Sampler guard suite: 10/10; sampler itself unchanged this round (no new
  keys — splits derive from the same snapshots).
- `./test/all`: identical to baseline (3 pre-existing environmental
  failures only). Bar-facing diff still zero bytes.
- **Live render, discharging window** (adapter offline, ~35 W): the watts
  split bar rendered base(accent) + comm(blue) + tail(muted) with fill
  widths matching the attribution (471px ≈ 35 W base of 37 W); POWER
  HUNGRY rows carried watts · CPU% · RAM with impact meters; the SYSTEM
  CPU bar rendered segmented (blue comm + muted rest + idle-as-track,
  ~130px busy of 519px ≈ the live 25% CPU). Two rendering bugs were
  caught live and fixed along the way (post-hoc JS object property
  additions carry no change signal — rows are now built complete before
  assignment; and the parent-width binding staleness above). Journal
  clean throughout.
- Captures: segmented CPU bar `~/Pictures/screenshot-2026-08-27_14-47-34.png`,
  watts split + attribution rows `~/Pictures/screenshot-2026-08-27_14-38-28.png`,
  RAM columns + stock bar `~/Pictures/screenshot-2026-08-27_13-24-15.png`.

### Diff summary (`git diff 76e0269c power-panel-vitals --numstat` — vs PR #1's tip)

```
 shell/Commons/Color.qml                |  +14  −0
 shell/plugins/panels/power/Model.js    | +507 −19
 shell/plugins/panels/power/Panel.qml   | +460  −8
 shell/plugins/panels/power/sampler.sh  |  +60  −3   (viz round 1 only)
                                           ———————
 total (this PR)                        | +1041 −30
```

---

## Operator checklist before pushing

1. Open PR #1 first; rebase this branch once it lands.
2. ~~Discharge-window capture~~ done (`14-38-28.png` — watts split bar,
   impact meters, yellow Draw row at 37 W).
3. Optional: amdgpu box live-check of the `gpu_busy_percent` path.
4. Decide the `cputotal`/`cpu_busy` alias question for the squashed view.

### DIFF
--- a/shell/Commons/Color.qml
+++ b/shell/Commons/Color.qml
@@ -21,6 +21,10 @@ QtObject {
   property color accent: "#cacccc"
   property color urgent: "#a55555"
   property color muted: "#707880"
+  // Status pair graded against urgent for positive/intermediate feedback;
+  // themes override via the same colorN key mapping in loadColors.
+  property color green: "#8fae8f"
+  property color yellow: "#b3a26b"
 
   // Flat dictionary of "section.key" -> raw string from shell.toml.
   // Reassigning this whole property is what makes surface bindings below
@@ -157,6 +161,8 @@ QtObject {
       else if (match[1] === "color7") color7Value = match[2]
       else if (match[1] === "color8") color8Value = match[2]
       else if (match[1] === "red" || match[1] === "color1") urgent = match[2]
+      else if (match[1] === "green" || match[1] === "color2") green = match[2]
+      else if (match[1] === "yellow" || match[1] === "color3") yellow = match[2]
     }
     if (!loadedBackground && color0Value.length > 0) background = color0Value
     if (!loadedForeground && color7Value.length > 0) foreground = color7Value
--- a/shell/Commons/Color.qml
+++ b/shell/Commons/Color.qml
@@ -21,12 +21,10 @@ QtObject {
   property color accent: "#cacccc"
   property color urgent: "#a55555"
   property color muted: "#707880"
-  // Status pair graded against urgent for positive/intermediate feedback;
-  // plus the distinguishable non-threshold hues used to key chart segments
-  // (red/yellow/green stay reserved for threshold grading). Themes override
+  // The distinguishable non-threshold hues used to key chart segments and
+  // metric columns (red stays reserved for threshold grading as urgent;
+  // green/yellow remain unexposed — no current consumer). Themes override
   // via the same colorN key mapping in loadColors.
-  property color green: "#8fae8f"
-  property color yellow: "#b3a26b"
   property color blue: "#6f8fae"
   property color cyan: "#7faeae"
   property color magenta: "#a68fae"
@@ -166,8 +164,6 @@ QtObject {
       else if (match[1] === "color7") color7Value = match[2]
       else if (match[1] === "color8") color8Value = match[2]
       else if (match[1] === "red" || match[1] === "color1") urgent = match[2]
-      else if (match[1] === "green" || match[1] === "color2") green = match[2]
-      else if (match[1] === "yellow" || match[1] === "color3") yellow = match[2]
       else if (match[1] === "blue" || match[1] === "color4") blue = match[2]
       else if (match[1] === "magenta" || match[1] === "color5") magenta = match[2]
       else if (match[1] === "cyan" || match[1] === "color6") cyan = match[2]
--- a/shell/Commons/Color.qml
+++ b/shell/Commons/Color.qml
@@ -22,9 +22,14 @@ QtObject {
   property color urgent: "#a55555"
   property color muted: "#707880"
   // Status pair graded against urgent for positive/intermediate feedback;
-  // themes override via the same colorN key mapping in loadColors.
+  // plus the distinguishable non-threshold hues used to key chart segments
+  // (red/yellow/green stay reserved for threshold grading). Themes override
+  // via the same colorN key mapping in loadColors.
   property color green: "#8fae8f"
   property color yellow: "#b3a26b"
+  property color blue: "#6f8fae"
+  property color cyan: "#7faeae"
+  property color magenta: "#a68fae"
 
   // Flat dictionary of "section.key" -> raw string from shell.toml.
   // Reassigning this whole property is what makes surface bindings below
@@ -163,6 +168,9 @@ QtObject {
       else if (match[1] === "red" || match[1] === "color1") urgent = match[2]
       else if (match[1] === "green" || match[1] === "color2") green = match[2]
       else if (match[1] === "yellow" || match[1] === "color3") yellow = match[2]
+      else if (match[1] === "blue" || match[1] === "color4") blue = match[2]
+      else if (match[1] === "magenta" || match[1] === "color5") magenta = match[2]
+      else if (match[1] === "cyan" || match[1] === "color6") cyan = match[2]
     }
     if (!loadedBackground && color0Value.length > 0) background = color0Value
     if (!loadedForeground && color7Value.length > 0) foreground = color7Value
--- a/shell/plugins/panels/power/Model.js
+++ b/shell/plugins/panels/power/Model.js
@@ -98,7 +98,96 @@ if (typeof module !== "undefined") {
     profileIcon: profileIcon,
     batteryFraction: batteryFraction,
     chargeThresholdActive: chargeThresholdActive,
+
     batteryIcon: batteryIcon,
-    modeLabel: modeLabel
+    modeLabel: modeLabel,
+    parseSnapshot: parseSnapshot,
+    buildTopProcesses: buildTopProcesses
   }
+}
+
+// ---- Power Hungry: top consumers ---------------------------------------------
+//
+// Panel-only companion to the stock battery panel: parse one sampler.sh
+// snapshot and attribute the measured battery draw across processes. The bar
+// pill is untouched by design — this feature answers a question you ask with
+// the panel open, and the bar stays calm (see the PR's design note).
+
+// Parses one sampler.sh snapshot into {watts, cpuTotalJiffies, processes}
+// where processes maps pid -> {name, jiffies}. watts is the absolute battery
+// flow; the sampler strips the driver's discharge sign so direction comes
+// from UPower, never from this value.
+function parseSnapshot(raw) {
+  var watts = null
+  var cpuTotalJiffies = null
+  var processes = {}
+  var lines = String(raw || "").split("\n")
+  for (var i = 0; i < lines.length; i++) {
+    var parts = lines[i].split("\t")
+    if (parts[0] === "watts" && parts.length >= 2 && parts[1] !== "") {
+      var w = parseFloat(parts[1])
+      if (!isNaN(w)) watts = w
+    } else if (parts[0] === "cputotal" && parts.length >= 2 && parts[1] !== "") {
+      var t = parseInt(parts[1], 10)
+      if (!isNaN(t)) cpuTotalJiffies = t
+    } else if (parts[0] === "p" && parts.length >= 4) {
+      var pid = parseInt(parts[1], 10)
+      var j = parseInt(parts[2], 10)
+      if (!isNaN(pid) && !isNaN(j)) {
+        if (!(pid in processes)) processes[pid] = { name: parts[3], jiffies: j }
+        else if (j > processes[pid].jiffies) processes[pid].jiffies = j
+      }
+    }
+  }
+  return { watts: watts, cpuTotalJiffies: cpuTotalJiffies, processes: processes }
+}
+
+// Busiest processes by CPU share of the jiffies consumed in the window,
+// aggregated by name so an app's helper processes render as one row.
+//
+// When the battery draw is known (discharging only — on AC the battery flow
+// is charge rate, not system draw) it is split into a calibrated idle base
+// and a variable slice attributed by share, so the returned rows are
+// [system base, top-N..., everything else] and sum to the measured draw.
+// A multi-threaded process can exceed 100% — jiffies sum across cores.
+function buildTopProcesses(prevSnapshot, nextSnapshot, limit, drawWatts, baseWatts) {
+  var nextP = nextSnapshot.processes
+  var prevP = prevSnapshot ? prevSnapshot.processes : {}
+  var totalDelta = Math.max(1,
+    nextSnapshot.cpuTotalJiffies - (prevSnapshot ? prevSnapshot.cpuTotalJiffies : 0))
+
+  var byName = {}
+  for (var pid in nextP) {
+    var prev = prevP[pid]
+    if (prev === undefined) continue
+    var delta = nextP[pid].jiffies - prev.jiffies
+    if (delta <= 0) continue
+    if (!(nextP[pid].name in byName)) byName[nextP[pid].name] = 0
+    byName[nextP[pid].name] += delta
+  }
+  var shares = []
+  for (var name in byName) shares.push({ label: name, share: byName[name] / totalDelta })
+  shares.sort(function(a, b) { return b.share - a.share })
+  shares = shares.slice(0, limit || 5)
+
+  var variable = drawWatts >= 0 ? Math.max(0, drawWatts - (baseWatts > 0 ? baseWatts : 0)) : 0
+  function pct(s) { return (s >= 1 ? Math.round(s * 100) : (s * 100).toFixed(1)) + "%" }
+  function wtxt(w) { return (w < 10 ? w.toFixed(1) : Math.round(w)) + " W" }
+
+  var out = []
+  var topShareSum = 0
+  for (var i = 0; i < shares.length; i++) {
+    topShareSum += shares[i].share
+    out.push({
+      label: shares[i].label,
+      value: drawWatts >= 0 ? wtxt(variable * shares[i].share) + " (" + pct(shares[i].share) + ")" : pct(shares[i].share)
+    })
+  }
+  if (drawWatts >= 0) {
+    if (baseWatts > 0.5) out.unshift({ label: "system base", value: wtxt(baseWatts) })
+    var otherShare = Math.max(0, 1 - topShareSum)
+    if (otherShare > 0.02 && variable * otherShare > 0.5)
+      out.push({ label: "everything else", value: wtxt(variable * otherShare) })
+  }
+  return out
 }
--- a/shell/plugins/panels/power/Model.js
+++ b/shell/plugins/panels/power/Model.js
@@ -102,6 +102,13 @@ function wattsBasis(deviceState, dischargingState, watts) {
 // Kernel comm fields cap at 15 characters (TASK_COMM_LEN), so the panel's
 // comm column can be sized once for the realistic worst case and never per
 // sample.
+// The shade lattice: three fixed opacity steps over each identity hue,
+// multiplying the SAME theme color — nine distinguishable identity slots
+// from three collision-free keys, zero new RGB, the 22-theme guarantee
+// intact. Steps chosen against the panel's 0.12-alpha track: the lowest
+// shade stays clearly above the track at panel scale (verified live).
+var SHADES = [0.5, 0.75, 1.0]
+
 var COMM_MAX_CHARS = 15
 
 // Gap budget basis for the composition bars: the most segments any bar can
@@ -134,6 +141,10 @@ if (typeof module !== "undefined") {
     aggregateCommShares: aggregateCommShares,
     assignColorKeys: assignColorKeys,
     stableColorKey: stableColorKey,
+    stableColorHash: stableColorHash,
+    SHADES: SHADES,
+    slotPreferences: slotPreferences,
+    resolveColorSlots: resolveColorSlots,
     SPLIT_MAX_SEGMENTS: SPLIT_MAX_SEGMENTS,
     collisionOrdinals: collisionOrdinals
   }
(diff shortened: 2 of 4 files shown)


vinceferro:
Consolidated into #8860 — same system (power panel), one review, one merge. This branch's full diff is included there; the stacked-PR split added sequencing friction for no review benefit. Thanks!

PR #8866, opened 2026-08-29
Fix network panel behind VPN policy routes

hehh2001:
## Summary

- Prefer an active physical NetworkManager device (Wi-Fi / Ethernet) when the network panel picks the connection to describe
- Fall back to the main routing table's lowest-metric default route when NetworkManager reports no active physical device
- Keep the Network panel on the LAN interface even when a full-tunnel VPN or Tailscale exit node owns the default route / installs policy routing
- Add/extend a regression test that simulates a connected Wi-Fi device alongside a connected Tailscale TUN device

## Why

`ip route get 1.1.1.1` follows policy routing. With a Tailscale exit node enabled it resolves to `tailscale0`, so the Network panel labels the connection as Ethernet and displays the Tailnet address instead of the active Wi-Fi/Ethernet address. Reading only the main routing table fixes policy-route VPNs but still follows a VPN that replaces the default route; preferring the active NetworkManager physical device covers both cases.

## Verification

- `bash -n bin/omarchy-network-status test/shell.d/network-status-test.sh`
- `test/shell.d/network-status-test.sh` passes with stubbed `nmcli`/`ip` (connected Wi-Fi + connected `tailscale0`)
- Live on Omarchy with an active Tailscale exit node: the panel shows the physical interface, local IPv4 address, and physical gateway

### DIFF
--- a/bin/omarchy-network-status
+++ b/bin/omarchy-network-status
@@ -19,10 +19,31 @@ case "${1:-}" in
     ;;
 esac
 
+# Prefer an active physical NetworkManager device (Wi-Fi or Ethernet). A
+# full-tunnel VPN or Tailscale exit node can own the default route, but the
+# panel should still describe the LAN connection the user actually joined.
+# If NetworkManager reports no active physical device, fall back to the main
+# routing table's lowest-metric default route.
+physical_device() {
+  local dev
+
+  dev=$(nmcli -t -f DEVICE,TYPE,STATE dev status 2>/dev/null | awk -F: '
+    $2 == "wifi" && $3 ~ /^connected/ { print $1; exit }
+    $2 == "ethernet" && $3 ~ /^connected/ { print $1; exit }
+  ')
+
+  if [[ -n $dev ]]; then
+    echo "$dev"
+    return
+  fi
+
+  ip -j route show table main default 2>/dev/null | jq -r 'sort_by(.metric // 0) | .[0].dev // ""'
+}
+
 print_status() {
   local device nm state ssid signal freq
 
-  device=$(ip -j route show table main default 2>/dev/null | jq -r 'sort_by(.metric // 0) | .[0].dev // ""')
+  device=$(physical_device)
 
   if [[ -z $device ]]; then
     printf 'disconnected\t\t\t\n'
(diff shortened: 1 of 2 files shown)


wundrellama:
Independent confirmation on a `wg-quick`-managed full-tunnel WireGuard connection (not WARP/Tailscale), plus a before/after capture against the current script.

**Environment**
- Omarchy `dev (86a2e583)` / `omarchy-dev 4.0.0.r2128.gb679363-1`
- Kernel 7.2.4-arch1-2, NetworkManager 1.58.1
- Wi-Fi associated to an OWE-transition SSID; WireGuard tunnel `AAM-WG` with `AllowedIPs = 0.0.0.0/0, ::/0` (owns the default route via table 51820)

**Before (`/usr/bin/omarchy-network-status --verbose`, tunnel up):**

```
iface   AAM-WG
ip      192.168.12.3
prefix  32
gateway
type    ethernet
cat: /sys/class/net/AAM-WG/speed: Invalid argument
speed
cat: /sys/class/net/AAM-WG/duplex: Invalid argument
duplex
```

**After (this PR's `bin/omarchy-network-status`, same moment, tunnel still up):**

```
iface   wlp5s0
ip      172.20.6.244
prefix  20
gateway 172.20.0.1
type    wifi
ssid    <redacted>-owe-tr
signal_dbm  -50
freq    5680.0
bitrate 286.7 MBit/s
router_ping_ms  1.45
```

The tunnel interface has no `/sys/class/net/<iface>/wireless` directory, so the
`if wireless … else ethernet` branch at the bottom of `print_verbose` classifies
it as `ethernet`, and the `speed`/`duplex` reads then fail with `Invalid argument`
because a WireGuard interface exposes neither. This PR's `physical_device()`
resolves it correctly.

**One extra symptom worth recording:** the bar icon and the panel's header text
are driven by *different* sources, so they disagree while a tunnel is up. The
text comes from `info.type` (this script, via `detailsPoll`), while the glyph
comes from `kind` in `plugins/panels/network/Panel.qml`, which is derived from
Quickshell's `Networking.devices`. The visible result is a header reading
"Ethernet" next to a Wi-Fi glyph, and because `detailsPoll` only runs while the
panel is open, the two drift in and out of agreement — which reads as the icon
flickering at random. Fixing the script's classification resolves the header
half; the glyph half is separate (Quickshell's `DeviceType` enum is
`[None, Wifi, Wired]`, so a WireGuard device is reported as `Wired` and
`Panel.qml`'s `kind` returns `"ethernet"` whenever a wired-type device is
connected).

A local workaround for the glyph half, for anyone hitting this before the
enum gains a VPN type:

```ini
# /etc/NetworkManager/conf.d/95-wireguard-unmanaged.conf
[keyfile]
unmanaged-devices=type:wireguard
```

`wg-quick` owns the interface anyway (NM only adopts it as
`managed-type: 'external'`), so marking it unmanaged costs nothing and stops
NM from racing the device through ~9 state transitions on every tunnel
up/down — each of which re-renders the bar icon.

hehh2001:
Thanks for the independent repro on a `wg-quick` full tunnel. A policy-routed default on table 51820 is exactly the case `physical_device()` exists for, and your before/after captures line up with what it does: pick the active physical NetworkManager device, and only fall back to the main table's lowest-metric default when there is none.

On the second symptom, you split it correctly and I agree it is two halves. The header text comes from `info.type` (this script), so this PR settles that half — a WireGuard interface is no longer classified as `ethernet`, and the `speed`/`duplex` reads that failed with `Invalid argument` go away with it. The glyph is drawn from `kind` in `plugins/panels/network/Panel.qml`, which derives it from Quickshell's `Networking.devices`; with `DeviceType` being `[None, Wifi, Wired]` a WireGuard device reports as `Wired`, so the panel can draw an Ethernet glyph next to a Wi-Fi header. I would rather not fold that into this PR: the honest fix is for the panel to make the same physical-device decision the status helper now makes (or for Quickshell to expose a VPN type) instead of inferring it from the device type, and that deserves its own change and review.

Your `unmanaged-devices=type:wireguard` drop-in is a good stopgap — `wg-quick` owns the interface anyway, so keeping NetworkManager out of it costs nothing. If you are willing to open an issue on `omacom/omarchy` with the flicker description, that is where the glyph half belongs; I will link it here so the two are read together.

wundrellama:
Thanks. Issue #10619 already tracks the false-disconnected Wi-Fi icon, so I added our observations there rather than create a duplicate: https://github.com/omacom/omarchy/issues/10619#issuecomment-5707920428

One correction to my earlier comment: the full-fan/slashed-fan cycling continued after `unmanaged-devices=type:wireguard` and a shell restart. That workaround did not resolve the flicker. The helper test supports the header fix in this PR, not a fix or proven cause for the intermittent icon changes. We did not test the device-state fix proposed in #10619.

PR #8869, opened 2026-08-29
fix(lockscreen): keep password input focused after suspend/wake

Phuocminh94:
When waking the system from suspend, the lock screen completely fails to regain keyboard focus. The password input becomes entirely unresponsive, and the user is effectively locked out of the session, forcing them to switch to a TTY to restart the display manager. 

Solution:

- Added focus: true to the TextInput to ensure default focus behavior.
- Added a fallback Timer at the root level of LockView.qml. It polls every 100ms to force focus on the password input only if the lock screen is active but the input lacks activeFocus. The timer automatically stops (running = false) the moment activeFocus is acquired.

### DIFF
--- a/shell/plugins/lock/LockView.qml
+++ b/shell/plugins/lock/LockView.qml
@@ -131,6 +131,7 @@ Item {
 
       TextInput {
         id: passwordInput
+        focus: true
         anchors.fill: parent
         anchors.topMargin: inputField.borderTop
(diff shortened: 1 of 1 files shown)


bjarneo:
Automated duplication check: this pull request looks similar to #7592, which covers the same lock refocus timer. I keep that one open and close this one to consolidate review.

PR #8875, opened 2026-08-29
Install lib32 GPU drivers before Steam

### LINKED ISSUE #8856: omarchy-install-gaming-steam pulls nvidia-utils onto non-NVIDIA systems: --noconfirm picks lib32-nvidia-utils for steam's lib32-vulkan-driver dep
`omarchy-install-gaming-steam` installs Steam before the GPU-aware driver step, so pacman's `--noconfirm` provider auto-pick wins the race:

```bash
# bin/omarchy-install-gaming-steam
omarchy-pkg-add steam                 # -> pacman -S --noconfirm --needed steam
omarchy-install-gaming-gpu-lib32      # correct and GPU-aware, but runs second
```

`steam` depends on the virtual `lib32-vulkan-driver`. Under `--noconfirm` pacman takes the first provider in repo order — `lib32-nvidia-utils` — which hard-depends on `nvidia-utils`. So the NVIDIA userspace stack gets installed on machines with no NVIDIA GPU.

From `/var/log/pacman.log` on an AMD-only laptop (Ryzen 9 7940HS / Radeon 780M, no dGPU):

```
10:57:02  pacman -S --noconfirm --needed steam
10:57:48  installed nvidia-utils (610.57.04-1)
10:57:53  installed lib32-nvidia-utils (610.57.04-1)
10:58:01  pacman -S --noconfirm --needed lib32-vulkan-radeon   <- 8 seconds too late
```

Detection itself is not at fault: `omarchy-hw-nvidia-gsp` and `omarchy-hw-nvidia-without-gsp` both correctly exit 1 on this hardware, and `omarchy-install-gaming-gpu-lib32` correctly installed only `lib32-vulkan-radeon`. It just runs after the damage is done.

**Why it matters.** `nvidia-utils` ships `/usr/share/glvnd/egl_vendor.d/10_nvidia.json`, which libglvnd ranks ahead of Mesa's `50_mesa.json`, plus the `egl_external_platform.d` entries. On this machine that produces a reproducible SIGSEGV in Brave's GPU process:

```
#0  pthread_mutex_lock            (libc.so.6)
#1  wl_proxy_create_wrapper       (libwayland-client.so.0)
#2  libnvidia-egl-wayland2.so.1 + 0x9b1a
#5  libEGL_nvidia.so.0 + 0xab0f9
#7  libEGL.so.1 + 0x625e
```

`rdi` at the fault is `0x133` — `offsetof(wl_display, mutex)` off a NULL `display` — byte-identical across four separate cores, so it is a deterministic null deref in NVIDIA's EGL wayland2 platform, not corruption. 11 crashes over 5 days; Chromium eventually hit its GPU-crash threshold and permanently set `--use-gl=disabled`, dropping the browser to software rendering.

**Suggested fix:** run `omarchy-install-gaming-gpu-lib32` before `omarchy-pkg-add steam`, or pass the resolved provider explicitly so `--noconfirm` has nothing to guess.

Related but distinct: #6216 also lands in this path, via the detector regex rather than provider order.

omarchy 4.0.1-1, linux 7.1.9.arch1-2. Skipping a full `omarchy debug` dump since this is a one-off script-ordering bug — happy to add it if useful.

Filed by Claude Opus 5 via Claude Code CLI.

Chessing234:
## Summary
- Run `omarchy-install-gaming-gpu-lib32` before `omarchy-pkg-add steam` so `--noconfirm` cannot auto-pick `lib32-nvidia-utils` (and hard-pull `nvidia-utils`) on AMD/Intel-only machines.
- Steam's `lib32-vulkan-driver` dependency is what triggered the wrong provider; the GPU-aware helper already knew the right packages, it just ran too late.

Fixes #8856.

## Test plan
- [x] `bash test/shell.d/gaming-steam-install-order-test.sh`
- [ ] On a non-NVIDIA machine: install Steam via the menu and confirm `pacman -Q nvidia-utils` stays absent

### DIFF
--- a/bin/omarchy-install-gaming-steam
+++ b/bin/omarchy-install-gaming-steam
@@ -5,9 +5,13 @@
 
 set -e
 
+# Resolve lib32-vulkan-driver before Steam so --noconfirm cannot pick
+# lib32-nvidia-utils (and hard-pull nvidia-utils) on non-NVIDIA machines.
+echo "Installing graphics drivers for Steam..."
+omarchy-install-gaming-gpu-lib32
+
 echo "Installing Steam..."
 omarchy-pkg-add steam
-omarchy-install-gaming-gpu-lib32
 
 echo ""
 echo "Steam will start automatically now. This might take a while..."
--- a/test/shell.d/gaming-steam-install-order-test.sh
+++ b/test/shell.d/gaming-steam-install-order-test.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+set -euo pipefail
+
+source "$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd)/base-test.sh"
+
+script="$ROOT/bin/omarchy-install-gaming-steam"
+[[ -x $script ]] || fail "omarchy-install-gaming-steam must be executable"
+
+gpu_line=$(grep -n 'omarchy-install-gaming-gpu-lib32' "$script" | head -1 | cut -d: -f1)
+steam_line=$(grep -n 'omarchy-pkg-add steam' "$script" | head -1 | cut -d: -f1)
+
+[[ -n $gpu_line && -n $steam_line ]] || fail "steam installer must call both gpu-lib32 and pkg-add steam"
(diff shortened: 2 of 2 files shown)


Chessing234:

PR #8877, opened 2026-08-29
Show the Codex CLI's error instead of "initialize" when the usage RPC fails

### LINKED ISSUE #8849: Agents panel: Codex limits show 'initialize' — stale --ask-for-approval flag value
## What happened
Agents bar panel, Codex tab shows usage status "CODEX LIMITS UNAVAILABLE" and the auth-help card literally reads "initialize" instead of a helpful message.

## Root cause
`omarchy-agent-usage-codex` collector launches the app-server with:

`-a untrusted` is rejected by current codex-cli:

codex-cli process exits immediately, the RPC `initialize` call never gets a response, and the collector's `rpc_request()` raises `TimeoutError("initialize")`. That exception's `str(exc)` ("initialize") gets stored verbatim in `authHelpText` and shown in the panel — should instead surface a real "run codex login" / version-mismatch message, and the hardcoded flag value should track the current codex-cli approval-policy enum (`on-request`/`never`, no more `untrusted`).

## Steps to reproduce
1. Install/update codex-cli to a version where `--ask-for-approval` no longer accepts `untrusted` (confirmed on codex-cli 0.149.0).
2. Log in with `codex login`, run at least one Codex session so local stats populate.
3. Open the Agents bar panel > Codex tab.
4. Local token stats show correctly; limits section shows "CODEX LIMITS UNAVAILABLE" and the help text is the bare string "initialize".

## Expected
Either the collector uses a codex-cli-version-compatible flag, or the auth-help text explains the real failure (flag/version mismatch) instead of leaking the raw RPC method name.

## System details
- Omarchy version: 4.0.0-1
- codex-cli version: 0.149.0 (installed via mise)
- `omarchy debug` command does not exist in this Omarchy version (checked `omarchy commands --all`), so no diagnostic log attached.

Betim-Hodza:
## What happened

The agents bar panel's Codex tab shows the usage status `CODEX LIMITS UNAVAILABLE`, and the auth-help card reads `initialize`, it doesn't give the user any indication on what's going on.

## Root cause

`rpc_request()` in `bin/omarchy-agent-usage-codex` raises `TimeoutError(method)` when the `codex app-server` process never answers, and `fetch_codex_rpc()` stored that exception verbatim:

```python
except Exception as exc:
    result["usageStatusText"] = "Codex limits unavailable"
    result["authHelpText"] = str(exc)   # -> "initialize"
```

So *any* RPC failure surfaces the raw method name. The common trigger is a `codex-cli` upgrade that drops or renames a flag the collector passes (e.g. `-a untrusted`), removed in codex-cli 0.149 in favour of `on-request`/`never` (the flag value itself is fixed separately in #7649). The CLI exits immediately:

```
```

`initialize` never gets a response, `rpc_request()` raises, and the panel shows `initialize`.

## Change

- Capture the app-server's stderr to a temp file (a regular file, not a `PIPE`, so an unread buffer can't deadlock a server that logs while it runs) instead of `subprocess.DEVNULL`.
- On RPC failure, `codex_rpc_help()` picks the most actionable text, in order:
  1. **The CLI exited and printed to stderr.** Quote it, all non-empty lines, so clap's `[possible values: ...]` line survives:
     ```
     Codex limits unavailable
     codex app-server exited: error: invalid value 'untrusted' for '--ask-for-approval <APPROVAL_POLICY>' [possible values: on-request, never]
     ```
  2. **The CLI exited without a word.** The "not logged in" shape, so keep the existing `Run \`codex login\` to authenticate.` hint.
  3. **The process is still running.** The RPC stalled or returned something the collector couldn't parse. That is not an auth problem, so it must not send the user to `codex login`; keep the raw exception text (cryptic but true and searchable, and exactly what the panel showed before this change).

The flag value passed to `codex app-server` is unchanged here. That is #7649's fix.

## Tests

`test/shell.d/agent-usage-codex-scanner-test.sh` gains cases for each branch:

- stub `codex` writes a two-line clap error and exits: asserts both stderr lines reach `authHelpText` and the bare method name does not.
- stub `codex` exits `0` without speaking the protocol: asserts `usageStatusText` is `Codex limits unavailable` and `authHelpText` is the `codex login` hint.
- stub `codex` answers `initialize` then never answers `account/read` (process stays up): asserts `authHelpText` keeps the stalled method name and is **not** the login hint.
- stub `codex` returns a rate-limit window with a non-numeric `usedPercent` (process stays up): asserts `authHelpText` keeps the parse error and is **not** the login hint.

`./test/shell` and `./test/cli` pass.

Fixes #8849.

### DIFF
--- a/bin/omarchy-agent-usage-codex
+++ b/bin/omarchy-agent-usage-codex
@@ -518,6 +518,40 @@ def limit_window(window):
   }
 
 
+def codex_rpc_help(proc, stderr_file, exc):
+  """Explain an app-server RPC failure in terms the user can act on.
+
+  rpc_request() raises TimeoutError(method), so a bare str(exc) surfaces in the
+  panel as a lone word like "initialize" or "account/read". Do better, in order:
+
+  1. If the CLI exited and printed to stderr, quote it. That names a flag a
+     codex-cli upgrade dropped or renamed, the common cause of a hard exit.
+  2. If the CLI exited without a word, fall back to the login hint: that is the
+     "not logged in" shape and `codex login` is the fix.
+  3. If the process is still running, the RPC stalled or returned something this
+     collector could not parse. That is not an auth problem and must not send
+     the user to `codex login`; the raw exception text is cryptic but true and
+     searchable, so keep it.
+  """
+  try:
+    proc.wait(timeout=1)
+  except Exception:
+    pass
+  if proc.poll() is None:
+    return str(exc)[:300] or AUTH_HELP
+  detail = ""
+  if stderr_file is not None:
+    try:
+      stderr_file.seek(0)
+      detail = stderr_file.read()
+    except Exception:
+      detail = ""
+  lines = [line.strip() for line in detail.splitlines() if line.strip()]
+  if lines:
+    return f"codex app-server exited: {' '.join(lines)}"[:300]
+  return AUTH_HELP
+
+
 def fetch_codex_rpc():
   result = {"limits": [], "tierLabel": "", "usageStatusText": "", "authHelpText": AUTH_HELP}
   codex = find_command("codex")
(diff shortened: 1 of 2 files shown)


Betim-Hodza:
Thanks for the thorough pass. Pushed a fix for the live-process finding (now at `2b0acc86`).

`codex_rpc_help` checks `proc.poll()` first now. If the process is still running, the RPC stalled or returned something the collector couldn't parse, which is not an auth problem, so it keeps `str(exc)` (the text the panel showed before this PR) instead of `AUTH_HELP`. Both cases you reproduced, the `account/read` stall and the non-numeric `usedPercent`, keep their original text. Only an exited process gets stderr quoted, or the login hint if it exited silently.

Also in this push:

- Quote every non-empty stderr line, not just the first, so clap's `[possible values: on-request, never]` line survives.
- The silent-exit test now also asserts `usageStatusText == "Codex limits unavailable"`, so it can't pass without reaching the error path.
- Two new tests: a stub that answers `initialize` then stalls on `account/read` (process stays up), and a stub that returns an unparseable rate-limit window. Both assert the login hint is not shown.

On the #8073 interaction: I've left the exited-silently branch returning `AUTH_HELP`. Once #8073's `rpc_request` raises a sentence instead of `TimeoutError(method)`, that branch should prefer the exception message, but with today's `rpc_request` it only ever sees the bare word `initialize` there, which is what this PR is trying to suppress. Happy to add that follow-up in whichever order the two land.

One case where the text is unchanged: an app-server that hangs on `initialize` itself, process still alive, still shows `initialize`. That is now accurate rather than misleading, since the process really is stuck on that method.

PR #8878, opened 2026-08-29
Match LocalSend's Wayland app id

### LINKED ISSUE #8817: Hyprland window rule for LocalSend does not match (class is 'org.localsend.localsend_app')
### System details

Omarchy 4.0.1-1

### Summary
LocalSend opens tiled instead of floating because the default Hyprland window rule does not match the application's Wayland window class.

### Environment
- **Omarchy Version:** 4.0.1-1
- **LocalSend Version:** 1.18.2-1 (arch package `localsend`)
- **Compositor:** Hyprland

### Steps to Reproduce
1. Launch LocalSend.
2. Observe that it opens tiled in the active layout rather than as a centered floating window of size 1100x700.

### Expected Behavior
LocalSend should float and center at 1100x700 as intended by `default/hypr/apps/localsend.lua`.

### Root Cause
In `default/hypr/apps/localsend.lua`, the rule is defined as:
```lua
o.window("(Share|localsend)", { float = true, center = true })
o.window("localsend", { size = { 1100, 700 } })
```

However, the packaged Linux application registers with the Wayland window class `org.localsend.localsend_app`:
```json
{
  "class": "org.localsend.localsend_app",
  "initialClass": "org.localsend.localsend_app",
  "title": "LocalSend"
}
```

Since Hyprland's `match.class` RE2 regex requires a full string match when wildcards are absent, `(Share|localsend)` fails to match `org.localsend.localsend_app`.

### Suggested Fix
Update `default/hypr/apps/localsend.lua` to include wildcards or the full app ID:

```lua
o.window(".*localsend.*", { float = true, center = true, size = { 1100, 700 } })
```
or:
```lua
o.window("^(org\\.localsend\\.localsend_app|localsend|Share)$", {
  float = true,
  center = true,
  size = { 1100, 700 },
})
```

yashranaway:
Makes LocalSend window rules match its packaged Wayland app ID. Fixes #8817 and #7482.

<<<< AI wording below >>>>

## Problem

The packaged LocalSend build identifies itself as `org.localsend.localsend_app`, while Omarchy only matches the legacy `localsend` class. The window therefore misses its intended float, center, and size rules.

## Fix

Match both class values. Keep the shared `Share` picker behavior, and apply the 1100x700 size only to LocalSend. The regression test loads the Lua rules under a stubbed compositor API and asserts the emitted rule semantics.

## Verification

- `bash test/shell.d/localsend-window-rule-test.sh`
- `./test/cli`
- `bash -n test/shell.d/localsend-window-rule-test.sh`
- `git diff --check`
- A disposable Hyprland 0.56.2 worker verified the packaged app ID opens floating at 1100x700, the legacy ID still matches, Share keeps its own size, and near misses stay tiled.
- This checkout cannot run the focused Lua test because `lua` is not installed.


yashranaway:
Thanks, I reviewed the semantic rule test and will leave the duplicate choice between this and #7736 to maintainers.

bjarneo:
Automated duplication check: this pull request looks similar to #7736, which covers the same LocalSend app ID match. I keep that one open and close this one to consolidate review.

PR #8882, opened 2026-08-29
Recover low-range displays from zero brightness

### LINKED ISSUE #8523: omarchy-brightness-display: Brightness trapped at 0% on displays with low max_brightness (acpi_video0)
### System details

2012 13" MacBook Pro (Retina)

#### Description

On devices where the backlight driver has a low `max_brightness` value (such as `acpi_video0` on a 2012 Retina MacBook Pro, where `max_brightness` is `15`), turning the brightness down to 0% with `Fn+F1` locks the backlight at 0%. Pressing `Fn+F2` cannot increase the brightness because `1%` rounds to raw step 0.

#### Steps to Reproduce

1. On a machine using `acpi_video0` (`max_brightness: 15`), press `Fn+F1` (`XF86MonBrightnessDown`) until brightness reaches 0%.
2. Press `Fn+F2` (`XF86MonBrightnessUp`).

#### Expected Behavior

Brightness should increase back above 0%.

#### Actual Behavior

Brightness remains at 0% (screen completely black). The user is locked out and cannot recover without remote shell access or hard rebooting.

#### Technical Diagnosis

In omarchy-brightness-display:

  if [[ $step == "+5%" ]]; then
    if (( current < 5 )); then
      (( target = current + 1 ))
    else
      (( target = current + 5 ))
    fi
    step="$target%"

When `current = 0`, `target` becomes `1%`.
Executing `brightnessctl -d acpi_video0 set 1%` calculates `0.01 * 15 = 0.15`, which `brightnessctl` rounds to raw `0`. As a result, `current` stays `0` on every `+5%` invocation.

#### Suggested Fix

Ensure `omarchy-brightness-display` increases raw brightness by at least +1 step when `current < 5` or when percentage rounding yields 0.

yashranaway:
Lets low-range internal displays recover from zero brightness. Fixes #8523.

<<<< AI wording below >>>>

## Problem

A one-percent target can round back to raw zero on backlights with a small `max_brightness`, so pressing brightness up leaves the panel black.

## Fix

Below five percent, request one raw hardware step with `brightnessctl set +1`. Keep the existing percentage targets and OSD behavior at five percent and above.

## Verification

- `bash test/shell.d/brightness-display-test.sh`
- `bash -n bin/omarchy-brightness-display test/shell.d/brightness-display-test.sh`
- `git diff --check`


bjarneo:
Automated duplication check: this pull request looks similar to #8524, which covers the same brightness low-range recovery. I keep that one open and close this one to consolidate review.

PR #8887, opened 2026-08-29
Give slow displays 30s after a lock-screen wake before blanking

### LINKED ISSUE #8847: Locked session re-arms a hardcoded 5s display-blank timer, so slow displays can never finish waking
### What happened

While the session is locked, waking the display re-arms a hardcoded 5-second
timer that blanks it again. On a display that takes longer than 5 seconds to
re-sync after DPMS off, the wake can never complete: the monitor is blanked
again before it has finished showing the lock screen. Moving the mouse
restarts the same race indefinitely.

From `shell/plugins/lock/Service.qml`:

```qml
Timer {
  id: idleBlankTimer
  interval: 5000
  ...
  onTriggered: {
    ...
    if (root.lockRequested && !root.authenticatingPassword) root.runBlank()
  }
}

function armBlankTimer() {
  idleBlankTimer.armedAt = Date.now()
  idleBlankTimer.restart()
}

function runWake() {
  if (!wakeProcess.running) wakeProcess.running = true
  if (lockRequested) armBlankTimer()   // every wake restarts the 5s countdown
}

function runBlank() {
  if (!blankProcess.running) blankProcess.running = true
}
```

`blankProcess` runs `omarchy-brightness-keyboard off; omarchy-brightness-display off`,
and `omarchy-brightness-display off` issues `hyprctl dispatch dpms disable`.
DPMS off drops the link entirely, so the monitor enters its own standby and
must renegotiate on wake — commonly several seconds over DisplayPort, longer
than the 5s budget.

The only escape is `authenticatingPassword`, which suppresses blanking once
the user has started typing. In practice that means the workaround is to type
the password blind, into a lock screen you cannot see.

### Expected

Waking a locked session should give the display enough time to actually
become visible. Any of:

1. Don't re-arm the blank timer inside `runWake()` — let it blank once on
   lock and stay awake after a deliberate wake until idle re-triggers.
2. Make the interval configurable (e.g. `lock.blank-delay` in shell.json)
   so slow displays can raise it.
3. Raise the hardcoded default well above typical DP re-sync time.

### Steps to reproduce

Requires a display that takes more than ~5s to wake from DPMS off; a
DisplayPort-connected desktop monitor is the usual case.

1. Lock the session (`omarchy system lock`, or wait for `idle.lock`).
2. After ~5s the display blanks and the monitor enters standby.
3. Move the mouse. The compositor re-enables DPMS
   (`misc:mouse_move_enables_dpms` is true), and the monitor begins
   re-syncing.
4. Before the monitor finishes and shows an image, the 5s timer fires and
   blanks it again.
5. Repeat indefinitely. The lock screen is never visible.
6. Type the password blind — it succeeds, confirming the session was
   responsive the whole time and only the display was unreachable.

### System

- Omarchy 4.0.1-1 (upgraded from 3.x via omarchy-upgrade-to-quattro)
- Kernel 7.1.9-arch1-2
- AMD Ryzen 5 5600X, NVIDIA RTX 2060 (TU106), nvidia-open-dkms
- MSI MP271 on DP-1, 1920x1080@144, scale 1
- `misc:mouse_move_enables_dpms` and `misc:key_press_enables_dpms` both true

Chessing234:
## Summary
- `runWake` re-armed the blank timer at 5s on every mouse move while locked.
- Displays that take longer than 5s to renegotiate DPMS were blanked again before the lock screen appeared, looping until the user typed blind.
- Wake now re-arms with a 30s budget; initial lock still uses 5s.

Fixes #8847.

## Test plan
- [x] `bash test/shell.d/lock-blank-fingerprint-test.sh`
- [ ] On a slow DP monitor: lock, wait for blank, move mouse, confirm the lock screen stays visible long enough to type

### DIFF
--- a/shell/plugins/lock/Service.qml
+++ b/shell/plugins/lock/Service.qml
@@ -159,14 +159,21 @@ Item {
     runWake()
   }
 
-  function armBlankTimer() {
+  function armBlankTimer(customInterval) {
+    // Deliberate wakes need longer than a DisplayPort re-sync (~5–15s). A
+    // hard-coded 5s re-arm in runWake blanked slow panels before the lock
+    // screen could appear, looping forever until the user typed blind.
+    var ms = (typeof customInterval === "number" && customInterval > 0) ? customInterval : 5000
+    idleBlankTimer.interval = ms
     idleBlankTimer.armedAt = Date.now()
     idleBlankTimer.restart()
   }
 
   function runWake() {
     if (!wakeProcess.running) wakeProcess.running = true
-    if (lockRequested) armBlankTimer()
+    // After an intentional wake, give the panel time to finish DPMS
+    // renegotiation before blanking again.
+    if (lockRequested) armBlankTimer(30000)
   }
 
   function runBlank() {
--- a/test/shell.d/lock-blank-fingerprint-test.sh
+++ b/test/shell.d/lock-blank-fingerprint-test.sh
@@ -30,4 +30,14 @@ assert(
   !/onAuthenticatingChanged:/.test(serviceQml),
   'the combined authenticating state no longer drives the blank timer'
 )
+
+assert(
+  /function runWake\(\) \{[\s\S]*?if \(lockRequested\) armBlankTimer\(30000\)/.test(serviceQml),
+  'waking a locked session re-arms blanking with a 30s DPMS re-sync budget'
+)
+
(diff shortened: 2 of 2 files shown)


Chessing234:

Chessing234:
folding into #12245 with the other lock session fixes

PR #8891, opened 2026-08-29
Allow user agent usage collectors

Hek846:
## What

Discover executable `omarchy-agent-usage-*` collectors from `${XDG_BIN_HOME:-$HOME/.local/bin}` in addition to packaged collectors. A user collector with the same provider id overrides the packaged one; a collector with a new id adds a provider record the existing dynamic panel discovery already understands.

## Why

The Agents panel record contract is already provider-agnostic, but the updater only executes collectors shipped inside Omarchy. This lets users and third-party packages add providers without editing `/usr/share/omarchy` or maintaining a fork.

## Verification

- Extended `agent-usage-update-test.sh` to cover user-only collectors and packaged-collector overrides
- `test/shell.d/agent-usage-update-test.sh`
- `git diff --check`

### DIFF
--- a/bin/omarchy-agent-usage-update
+++ b/bin/omarchy-agent-usage-update
@@ -10,6 +10,7 @@
 # panel picks up any record that appears here.
 
 USAGE_DIR="${XDG_STATE_HOME:-$HOME/.local/state}/omarchy/agents/usage"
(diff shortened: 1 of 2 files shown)


bjarneo:
Automated duplication check: this pull request looks similar to #8267, which covers the same user collector discovery. I keep that one open and close this one to consolidate review.

Issue #8780, opened 2026-08-28
omarchy update -y hangs on the orphaned-packages prompt


PixelatedContinuum:
`omarchy update -y` blocks indefinitely at `Remove N orphaned package(s)?`. The package upgrade itself completes; the run then stops at the orphan step and waits forever.

`-y` sets `OMARCHY_UPDATE_UNATTENDED=1`, documented in `omarchy-update` as "a promise not to ask anything. Steps that would need an answer report and move on instead of waiting on a prompt nobody is here to give."

## Cause

`omarchy-update-orphan-pkgs` guards on the TTY only:

```bash
if [[ ! -t 0 || ! -t 1 ]]; then
  echo "${#orphans[@]} orphaned package(s) found. Re-run omarchy-update-orphan-pkgs in a terminal to review/remove them."
  echo
  exit 0
fi

if ! gum confirm --default=false "Remove ${#orphans[@]} orphaned package(s)?"; then
```

That guard is correct when the script is called directly, but `omarchy-update` re-execs itself under `script(1)` to capture the transcript (`omarchy-update:12`):

```bash
exec env OMARCHY_UPDATE_LOGGED=1 script -qefc "$script_command" "/tmp/omarchy-update.log"
```

`script` allocates a pseudo-TTY, so `-t 0` and `-t 1` are true regardless of how the update was invoked. The guard never fires, and `gum confirm` prompts into a PTY nobody is attached to.

```console
$ bash -c '[[ -t 0 && -t 1 ]] && echo TTY || echo no-TTY' < /dev/null
no-TTY

$ script -qefc 'bash -c "[[ -t 0 && -t 1 ]] && echo TTY || echo no-TTY"' /dev/null < /dev/null
TTY
```

## Suggested fix

`omarchy-update-system-pkgs-when-conflicted:74` already has the pattern:

```bash
if [[ ${OMARCHY_UPDATE_UNATTENDED:-} == 1 || ! -t 0 || ! -t 2 ]]; then
```

Adding the same clause to `omarchy-update-orphan-pkgs` would fix it:

```bash
if [[ ${OMARCHY_UPDATE_UNATTENDED:-} == 1 || ! -t 0 || ! -t 1 ]]; then
```

`omarchy-update-restart` has the same shape (a bare `gum confirm` with no guard). It only prompts on a kernel change or a `reboot-required` marker so it is hit less often, but `-y` would block there too.

## Environment

- omarchy 4.0.1-1
[...]

PR #8894, opened 2026-08-29
Skip the orphan prompt when omarchy update runs with -y

Chessing234:
## Summary
- `omarchy update -y` sets `OMARCHY_UPDATE_UNATTENDED` but still has a TTY, so the orphan step waited on `gum confirm` forever after the upgrade finished.
- Treat the unattended flag like a non-interactive pipe: report orphans and move on.

Addresses the orphan-package prompt in #8780. The separate restart-prompt path is outside this change.

## Test plan
- [x] `bash test/shell.d/update-orphan-test.sh` (bash 5)
- [ ] Run `omarchy update -y` with orphans present; confirm it exits without prompting

### Review follow-up — 12 September 2026

The regression now runs the checker through `script` with both input and output attached to a PTY, and explicitly rejects a gum invocation. The runner verifies both descriptors are terminals before invoking the checker. All three focused checks pass on Bash 5/macOS; removing the unattended condition makes the new regression fail. The Linux invocation follows the existing `update-package-conflict-test.sh` helper; a complete live Omarchy update remains unverified.

### DIFF
--- a/bin/omarchy-update-orphan-pkgs
+++ b/bin/omarchy-update-orphan-pkgs
@@ -12,7 +12,9 @@ echo -e "\e[32m\nOrphan system packages\e[0m"
 printf '  %s\n' "${orphans[@]}"
 echo
 
-if [[ ! -t 0 || ! -t 1 ]]; then
+# omarchy update -y keeps a TTY but promises not to ask. Treat unattended
+# the same as a non-interactive pipe: report and move on.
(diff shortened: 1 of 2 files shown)


Chessing234:

Chessing234:
fixed the coverage gap in 70436f6b: the checker now runs with stdin/stdout on a real pty, and the test rejects any gum invocation. all three focused checks pass; removing the unattended guard makes the new test fail. full live update remains unverified.

Chessing234:
folded into #12109 next to the other update-path hardening.

PR #8895, opened 2026-08-29
Clear the low-battery toast when the charger is connected

### LINKED ISSUE #8813: Low-battery notification remains after charger is connected
## System details

- Lenovo ThinkPad L13 Gen 2 (20VH004EUK)
- Intel Core i5-1135G7 / Intel Iris Xe Graphics
- Omarchy 4.0.0-1
- Hyprland 0.56.2
- quickshell-git 0.3.0.r20.g28771c7-1
- UPower 1.91.3-1

## What is wrong?

The persistent critical `Time to recharge!` notification does not automatically disappear after connecting the charger. It remains on screen until it is manually dismissed, even though UPower has transitioned from discharging to charging.

This is the normal same-session transition after a genuine low-battery warning, not the startup race described in #7679.

### Steps to reproduce

1. Run on battery until the level reaches 10%.
2. Observe the `Time to recharge!` notification.
3. Connect the charger and confirm that the battery state changes to charging.
4. Observe that the notification remains until manually dismissed.

### Expected

The low-battery warning should be dismissed automatically once AC power is connected and `UPower.onBattery` becomes false.

### Actual

The warning remains visible indefinitely until manually dismissed.

## Verification and likely cause

The notification history confirms that the warning was emitted at 10%. After connecting AC, UPower reports `state: charging`; the warning still required manual dismissal.

`bin/omarchy-battery-low` sends the warning with critical urgency. The notification service treats critical notifications as persistent. In `shell/plugins/services/battery/Service.qml`, `onOnBatteryChanged()` calls `checkBattery()` and `applyPowerProfile()`, but it does not dismiss the existing warning when switching to AC.

Omarchy already exposes `omarchy notification dismiss <summary>`, so the AC transition could dismiss the `Time to recharge!` warning explicitly.

Related but distinct: #7679.

Chessing234:
Critical "Time to recharge!" notifications ignore the send timeout and can remain visible after AC is connected. Leaving the low-discharging state now dismisses that notification and retains dismissal intent across the existing 30-second battery polls. This retries when notification insertion follows the first dismissal, the dismissal process is busy, or the quiet CLI times out without acknowledgement. A new low-battery episode cancels the previous dismissal intent and waits for an in-flight dismissal process to finish before sending its warning.

Fixes #8813.

Validation: `bash test/shell.d/battery-clear-on-ac-test.sh` and `bash test/shell.d/battery-test.sh` pass. The focused test executes the QML check and dismissal functions and covers the actual transition call, busy-process retry, and cancellation on a new low-battery episode.

Live Quickshell verification remains pending: trigger the critical notification, connect AC during notification delivery, and confirm eventual dismissal. This requires an Omarchy desktop, unavailable in the current macOS environment.

### DIFF
--- a/shell/plugins/services/battery/Service.qml
+++ b/shell/plugins/services/battery/Service.qml
@@ -14,6 +14,7 @@ Item {
   property string pendingPowerSource: ""
   property string activePowerProfile: ""
   readonly property bool powerSaverOnBattery: UPower.onBattery && activePowerProfile === "power-saver"
+  property bool lowBatteryClearPending: false
 
   PersistentProperties {
     id: persisted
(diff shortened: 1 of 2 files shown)


Chessing234:

Chessing234:

PR #8896, opened 2026-08-29
Debounce unforced lock wake calls immediately following display blanking

### LINKED ISSUE #8863: Display flickers off/on repeatedly after idle lock on AMD APU + HDMI/DP (DPMS flap loop)
## Environment
- Omarchy 4.0.1-1 (Arch)
- GPU: AMD HawkPoint1 [Radeon] (PCI 66:00.0)
- HDMI audio: AMD Radeon High Definition Audio Controller (PCI 66:00.1)
- Terminal: foot (screensaver launches fine)
- Display connected via HDMI/DP

## Expected behavior
When idle, the system should lock and the display should turn off (or stay locked) cleanly — the same way Suspend from the menu does.

## Actual behavior
After the lock engages, the display turns off and on repeatedly (~5–8 s) while locked. The lock state itself is stable (`secure=true`); only the display DPMS toggles.

## Root cause
The lock plugin (`shell/plugins/lock/Service.qml`) arms `idleBlankTimer` (5 s) after locking; its trigger runs `blankProcess` → `omarchy-brightness-display off` → `hl.dsp.dpms disable` (DPMS off) **while the session is live**.

On this AMD + HDMI combo, DPMS-off drops the HDMI link, and the Radeon HDMI audio codec emits `SW_VIDEOOUT_INSERT` / `SW_LINEOUT_INSERT` switch events (evdev `event18` — "HD-Audio Generic HDMI/DP"). These are treated as activity, which wakes the lock (`runWake` → `omarchy-system-wake` → DPMS on), re-arms the 5 s blank, and the cycle repeats.

## Evidence
Traced `hyprctl monitors` `dpmsStatus` with `omarchy-shell idle/lock status` and raw `/dev/input` events. Every DPMS toggle coincides exactly with a switch event from the HDMI audio device; no keyboard/mouse input occurs during the flicker:

```
19:05:40  EVENT event18 [HD-Audio Generic HDMI/DP,pcm=7] SW c=6  → dpms=[False]
19:05:48  EVENT event18 [HD-Audio Generic HDMI/DP,pcm=7] SW c=8  → dpms=[True]
19:05:53  EVENT event18 [...]                                → dpms=[False]
19:06:01  EVENT event18 [...]                                → dpms=[True]
... (continues; locked=true secure=true throughout)
```

## Why Suspend works but idle doesn't
Suspend powers the display down at the kernel PM level (clean, like Windows). The lock-during-suspend path uses `OMARCHY_LOCK_ONLY=true omarchy-system-lock` ("lock before suspend without scheduling display off"), so it never schedules the live DPMS-off blank that triggers the loop. Windows (dual-boot, same cable/monitor) also sleeps the display cleanly, confirming it is not a hardware/cable fault.

## Workaround
Commenting out the blank in `shell/plugins/lock/Service.qml` stops the flicker (display stays on while locked):

```qml
function runBlank() {
    // if (!blankProcess.running) blankProcess.running = true
}
```

(Note: editing a `/usr/share/omarchy` file; it can be overwritten on update.)

## Suggested fix
- Don't treat unrelated switch/jack events (e.g. HDMI audio `SW_VIDEOOUT_INSERT`) as wake activity, or
- Stabilize/guard the display before issuing DPMS-off (e.g. delay the blank, or detect the link-flap and avoid re-arming), or
- On lock, prefer a suspend-like display power-down instead of a live `hl.dsp.dpms disable` that flaps on certain GPUs.

## Steps to reproduce
1. Omarchy on AMD APU with HDMI/DP display + Radeon HDMI audio.
2. Leave idle until the lock engages (default `lock = 300`, repro used `100`/`200`).
3. ~5 s after lock, the display flickers off/on.
4. Confirm with `watch -n1 'hyprctl monitors -j | jq .[].dpmsStatus'` and `sudo libinput debug-events` (or raw `/dev/input` monitoring) showing HDMI audio switch events coinciding with each toggle.

Sword-Saint69:
### Problem (#8863)
On AMD APU setups with Radeon HDMI audio codecs (`snd_hda_codec_hdmi`), disabling DPMS (`omarchy-brightness-display off`) causes the kernel driver to emit `SW_VIDEOOUT_INSERT` evdev switch events on `/dev/input/eventX` ("HD-Audio Generic HDMI/DP").

The input subsystem treats this hardware switch event as user activity, firing `runWake()` (`omarchy-system-wake`) and re-enabling DPMS in a continuous 5–8 second flap loop.

---

### Fix
1. **Post-Blanking Debounce**: Added a 1.5s post-blanking window (`Date.now() - lastBlankedAt < 1500`) in [Service.qml](file:///d:/projects/omarchy/shell/plugins/lock/Service.qml#L167-L170). Unforced/automated wake calls occurring within 1.5s of `runBlank()` are ignored to swallow hardware HPD/audio switch events.
2. **Explicit User Interaction Unchanged**: User interaction (`LockView` mouse click, typing, password submission) passes `force = true` (`runWake(true)`), waking the screen immediately without delay.
3. **Automated Tests**: Updated [test/shell.d/lock-blank-fingerprint-test.sh](file:///d:/projects/omarchy/test/shell.d/lock-blank-fingerprint-test.sh#L34-L37) to assert `runWake` post-blanking debounce logic.

Fixes #8863

### DIFF
--- a/shell/plugins/lock/LockView.qml
+++ b/shell/plugins/lock/LockView.qml
@@ -41,7 +41,7 @@ Item {
   signal submitPassword(string password)
   signal passwordTextEdited(string password)
   signal clearFailureRequested()
-  signal wakeRequested()
+  signal wakeRequested(bool force)
 
   // Cache-busts the lock background by appending `?v=`. Adding a query
   // string keeps Image's loader happy while forcing it to reload when the
@@ -115,8 +115,8 @@ Item {
     MouseArea {
       anchors.fill: parent
       hoverEnabled: true
-      onClicked: { root.wakeRequested(); root.forcePasswordFocus() }
-      onPositionChanged: root.wakeRequested()
+      onClicked: { root.wakeRequested(true); root.forcePasswordFocus() }
+      onPositionChanged: root.wakeRequested(false)
     }
 
     BorderSurface {
@@ -163,7 +163,7 @@ Item {
         onTextChanged: {
           if (!root.syncingPasswordText) root.passwordTextEdited(text)
           if (text.length > 0) {
-            root.wakeRequested()
+            root.wakeRequested(true)
           }
           if (text.length > 0 && root.failureMessage.length > 0) root.clearFailureRequested()
         }
@@ -175,7 +175,7 @@ Item {
         }
 
         Keys.onPressed: function(event) {
-          root.wakeRequested()
+          root.wakeRequested(true)
           if (event.key === Qt.Key_Escape || (event.modifiers & Qt.ControlModifier && event.key === Qt.Key_U)) {
             root.passwordTextEdited("")
             event.accepted = true
--- a/shell/plugins/lock/Service.qml
+++ b/shell/plugins/lock/Service.qml
@@ -164,12 +164,16 @@ Item {
     idleBlankTimer.restart()
   }
 
-  function runWake() {
+  property double lastBlankedAt: 0
+
+  function runWake(force) {
+    if (!force && locked && (Date.now() - lastBlankedAt) < 1500) return
     if (!wakeProcess.running) wakeProcess.running = true
     if (lockRequested) armBlankTimer()
   }
 
   function runBlank() {
+    lastBlankedAt = Date.now()
     if (!blankProcess.running) blankProcess.running = true
   }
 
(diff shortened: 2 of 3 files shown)


Sword-Saint69:
Updated the implementation to differentiate deliberate user input from generic motion:
- `signal wakeRequested(bool force)` in `LockView.qml` now passes `force = true` for deliberate user actions (`onClicked`, `Keys.onPressed`, `onTextChanged`).
- `onPositionChanged` passes `force = false`, ensuring pointer motion and synthetic layout events remain subject to the 1.5s post-blanking debounce.
- `Service.qml` forwards `onWakeRequested: function(force) { root.runWake(force) }`.

PR #8898, opened 2026-08-29
Let a click on °C/°F pin metric, imperial, or auto

### LINKED ISSUE #8690: Timezone changes weather unit too
### System details

HP Spectre x360 (2024)

### What's wrong?

Setting the timezone in this case, to Central Time automatically updates the unit for the weather widget to Fahrenheit, I dug through the shell.json but could not find any option to override it to Celsius. I can see how this is a feature but I am surprised that there is not any way to change it. Am I missing something?

https://logs.omarchy.org/1XIJiz (mostly irrelevant, but just in case)

Note: I was considering adding this to feature requests but it just seems broken to me and I assume many other people to be seeing the temperature in Fahrenheit, especially after knowing it was fully capable of displaying Celsius.

Chessing234:
## Summary
- Timezone / reported-country changes can flip the weather unit with no UI escape hatch, even though shell.json already accepts a `unit` override.
- Clicking the °C/°F label now cycles auto → metric → imperial → auto and persists via `updateEntryInline`.

Fixes #8690.

## Test plan
- [x] `bash test/shell.d/weather-unit-cycle-test.sh`
- [ ] Open weather panel, click the unit letter, confirm it pins and survives a shell restart

### DIFF
--- a/shell/plugins/panels/weather/Model.js
+++ b/shell/plugins/panels/weather/Model.js
@@ -111,6 +111,14 @@ function countryUsesImperial(countryName) {
   return false
 }
 
+
+function nextTempUnitOverride(current) {
+  var unit = String(current || "")
+  if (unit === "metric") return "imperial"
+  if (unit === "imperial") return ""
+  return "metric"
+}
+
 function shouldUseImperial(unitOverride, localeName, countryName) {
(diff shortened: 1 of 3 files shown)


Chessing234:

Chessing234:

PR #8899, opened 2026-08-29
Keep terminals from inheriting another window's fullscreen

### LINKED ISSUE #8751: Terminal Opens in Fullscreen When Opened While Another Window is in Fullscreen
### System Details
- **Omarchy version:** 4.0.1-1
- **Hyprland version:** 0.56.2
- **GPU:** Intel UHD Graphics (Alder Lake-UP3 GT1)
- **Terminal:** foot

### Description
When a window is in fullscreen mode (e.g., watching a video in mpv, browser, etc.) and the user opens the terminal (foot), the terminal also opens in fullscreen mode. This behavior is unexpected - the terminal should open in normal windowed mode regardless of the fullscreen state of other windows.

### Steps to Reproduce
1. Open any application (e.g., mpv, browser)
2. Enable fullscreen mode (e.g., press `F` in mpv, or browser fullscreen)
3. While in fullscreen, open the terminal (e.g., `Super + Return`)
4. The terminal opens in fullscreen mode instead of normal windowed mode

### Expected Behavior
The terminal should always open in normal windowed mode, regardless of the fullscreen state of other windows. Fullscreen state should not be inherited by newly opened windows.

### Actual Behavior
The fullscreen state appears to be inherited by newly opened windows when they are launched while another window is in fullscreen.

### Workaround
Press `Super + F` to toggle fullscreen off after opening the terminal.

Chessing234:
## Summary
- Opening a terminal while another client is fullscreen was mapping the terminal fullscreen too.
- Force `fullscreen = false` on the terminal window tag so Super+Return stays windowed.

Fixes #8751.

## Test plan
- [x] `bash test/shell.d/terminal-nofullscreen-test.sh`
- [ ] Fullscreen mpv/browser, press Super+Return, confirm the terminal is not fullscreen

### DIFF
--- a/default/hypr/apps/terminals.lua
+++ b/default/hypr/apps/terminals.lua
@@ -4,5 +4,10 @@
 -- The class is matched in full, so foot's other app-id needs spelling out.
 o.window(
   "(Alacritty|kitty|com.mitchellh.ghostty|foot|org\\.codeberg\\.dnkl\\.foot|wezterm|org\\.omarchy\\..*|TUI\\..*)",
-  { tag = "+terminal" }
+  {
+    tag = "+terminal",
+    -- Opening a terminal while another client is fullscreen must not inherit
+    -- that fullscreen state (Hyprland's syncfullscreen / focus-under path).
+    fullscreen = false,
+  }
 )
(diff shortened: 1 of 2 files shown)


Chessing234:

Chessing234:
closing — omarchybot is right. fullscreen=false is a no-op on hyprland static rules and the org.omarchy.* regex overwrites the screensaver's fullscreen=true. the inherit path is misc.on_focus_under_fullscreen=1 in looknfeel.lua, not missing terminal rules.

PR #8907, opened 2026-08-29
foot: render light themes into [colors-light] so foot reports the right color-theme mode

### LINKED ISSUE #8905: Light themes render foot colors into [colors-dark], so foot reports dark mode to 996-aware TUIs
Omarchy 4.0.1, foot 1.27.0, theme Flexoki Light.

## What happens

TUIs that ask the terminal for its color-theme mode render in dark mode on every light Omarchy theme, even though the terminal background is light. Observed with hunk (hunk.dev), which uses the contour color-scheme protocol (`CSI ?996n` query, mode 2031 for updates). foot 1.27 supports this protocol.

## Why

`default/themed/foot.ini.tpl` renders every theme's palette into the `[colors-dark]` section:

```ini
[colors-dark]
foreground={{ foreground_strip }}
background={{ background_strip }}
...
```

foot's convention (foot.ini(5), foot-ctlseqs(7)) is that `[colors-dark]` is the dark theme and `[colors-light]` is the light theme, and it answers `CSI ?996n` with `CSI ?997;1n` (dark) whenever the primary/dark theme is active. So with Flexoki Light the background is `FFFCF0` but foot still reports "dark" to applications.

The theme already declares its mode: `colors.toml` has `mode = "light"`. The template just does not use it.

## Suggested fix

When `mode = "light"`, render the palette into `[colors-light]` instead and set `initial-color-theme=light` in the generated include (or have the default `foot.ini` set it). Then foot reports `CSI ?997;2n` and 996/2031-aware TUIs pick the right mode.

## Repro

1. `omarchy theme set flexoki-light` (or any theme with `mode = "light"`).
2. In foot, run a 996-aware TUI, e.g. `hunk diff`.
3. It renders dark. `printf '\033[?996n'` shows foot replying `?997;1n` (dark).

rubas:
Fixes #8905.

## Problem

`default/themed/foot.ini.tpl` renders every theme's palette into `[colors-dark]`. foot reports its active color-theme section to applications (`CSI ?996n` / mode 2031, contour protocol), and treats the primary theme as dark by convention. So on a light theme (Flexoki Light, background `FFFCF0`) foot answers "dark", and 996-aware TUIs such as hunk render dark on a light background. The theme already declares `mode = "light"` in `colors.toml`; the template ignored it.

## Fix

`foot.ini.tpl` renders the palette into `[colors-{{ mode }}]` and sets `initial-color-theme={{ mode }}`. The line lands at `[main]` scope because foot's `include` gives the file its own section scope (foot.ini(5)). `mode` is already a template token from `omarchy-theme-color --all`, resolved for every theme (explicit key, legacy markers, luminance fallback, default dark).

Themes that ship a static `foot.ini` are on `INSTALLED_THEME_DENIED` and always go through the template, so this covers every theme.

## Validation

- `test/shell.d/theme-staging-test.sh` passes.
- Template rendered with `mode` resolved from a live Flexoki Light `colors.toml` (Omarchy 4.0.1, foot 1.27.0): the include starts with `initial-color-theme=light` + `[colors-light]`.

Known limitations: a foot window that is already open does not re-read its config, so it keeps the old reported mode until reopened (the existing OSC repaint still fixes its colors). And only the active mode's section is rendered, so foot's manual color-theme toggle lands on foot's stock palette for the other mode.

### DIFF
--- a/bin/omarchy-theme-color
+++ b/bin/omarchy-theme-color
@@ -16,6 +16,7 @@
 #
 # Theme mode precedence: `mode` key, legacy `theme_type` key, a light.mode
 # file beside the colors.toml, background luminance auto-detect, dark.
+# The value is lowercased; anything but light/dark falls to auto-detect.
 
 COLORS_FILE="$HOME/.local/state/omarchy/current/theme/colors.toml"
 OUTPUT=""
@@ -118,10 +119,17 @@ alias_theme_color() {
 }
 
 resolve_theme_mode() {
-  local bg_hex lum
+  local declared bg_hex lum
 
-  [[ ${THEME_COLORS[mode]} ]] || THEME_COLORS[mode]="${THEME_COLORS[theme_type]}"
-  [[ ${THEME_COLORS[mode]} ]] && return
+  # foot rejects anything outside light/dark, so only those two values
+  # may leave the resolver.
+  declared="${THEME_COLORS[mode]:-${THEME_COLORS[theme_type]}}"
+  declared="${declared,,}"
+
+  if [[ $declared == "light" || $declared == "dark" ]]; then
+    THEME_COLORS[mode]="$declared"
+    return
+  fi
 
   if [[ -f $COLORS_DIR/light.mode ]]; then
     THEME_COLORS[mode]="light"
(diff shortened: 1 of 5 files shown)


rubas:
I went ahead and fixed the mode normalization in this PR (68061713), since it's this change that puts the raw value into foot's section name in the first place. The resolver now lowercases the value and falls back to auto-detect for anything that isn't light or dark. The shipped themes are unaffected.

PR #8908, opened 2026-08-29
Switch DNS providers without DHCP churn or profile rewrites

danbosscher:
Switching DNS providers rewrites saved connection settings and reapplies active connections, which can restart DHCP. A global-only override also fails with NetworkManager's `systemd-resolved` backend: testing NetworkManager 1.58.1 showed that it continued publishing the original per-link DNS. A global resolved `~.` route would introduce a competing route alongside a privacy VPN.

Apply the chosen provider to resolved's live ordinary network links, preserving NetworkManager's DNS routing and the saved profiles. Selecting DHCP removes the selection and asks NetworkManager to republish the connection's original DNS, including manually configured servers, without reapplying the connection.

## Behavior

- Store the choice atomically in `/etc/omarchy/dns.json`; apply through the resolved D-Bus API using the existing Python/GObject runtime.
- Preserve VPN, WireGuard, externally managed links, search/routing domains, negative-priority DNS exclusions, DNSSEC, and DNS-over-TLS settings.
- Reapply through NetworkManager dispatcher events, including pre-up, reconnect, DNS changes and DHCP renewal. Restore the selection after resolved restarts without preventing the resolver from starting if NetworkManager is not running yet.
- Keep privileged execution pinned to the packaged helper with Python isolated mode. Retain the existing sudo/pkexec behavior and stock-provider sudoers scope.
- Back up exact legacy Omarchy global config before removing its server directives; preserve encryption settings and unrecognized customization. Report competing custom global DNS instead of silently overwriting it. Roll back failed explicit selections.
- Never guess which saved static DNS values were written by older Omarchy versions: they have no ownership marker. Settings already overwritten by old versions cannot be reconstructed.

[Design and distro research](https://github.com/danbosscher/omarchy/blob/fix/dns-switch-dhcp-churn/docs/dns.md) compares Fedora's per-link resolved integration, Ubuntu/Netplan, RHEL priorities, Arch's global routing guidance, and Fedora's optional dnsconfd/Unbound stack. It explains why this keeps the existing resolver stack and avoids a new global `~.` route.

## Validation

Maintained regression coverage follows the repository's focused shell-test pattern:

- `dns-policy-test.sh`: one self-contained script using temporary files and stubbed D-Bus calls. Covers provider changes, input validation, legacy config preservation, rollback, VPN exclusions, encryption preservation, and global-config conflicts.
- Existing DNS sudoers/elevation tests, including poisoned PATH and PYTHONPATH.

Both pass. `./test/cli`, network-manager transition tests, privileged-heredoc tests, Bash syntax, `systemd-analyze verify` for the resolver drop-in, and `git diff --check` also passed during development.

One-off investigation with real NetworkManager, resolved and the dispatcher in private network/filesystem namespaces verified actual IPv4/IPv6 DNS answers, unchanged profiles, DHCP restoration, reconnects, DNS reload, resolved restart, split/private VPN routing, WireGuard with negative DNS priority, and actual DHCP lease renewal without a new transaction or address change. That disposable lab is not part of the maintained suite. It executed the restart hook directly rather than booting a full systemd instance. No host network configuration was changed.

## Compatibility and limits

Install matching `omarchy` and `omarchy-settings` packages. Existing global config is converted on the next explicit provider selection. Public DNS can prevent router-only names or captive portals from working; DHCP restores the native settings. VPN DNS retains its own routing policy. Provider selection now preserves encryption policy rather than enabling/disabling it globally. Custom accepts IPv4/IPv6 literals. Dispatcher updates are asynchronous for some events, so this is a DNS preference, not a firewall-backed no-leak guarantee. Full boot and physical Wi-Fi/VPN interoperability remain release-validation tasks.

### DIFF
--- a/bin/omarchy-dns
+++ b/bin/omarchy-dns
@@ -19,6 +19,7 @@ if (( EUID == 0 )); then
 fi
 
 NM_DNS_CONF=/etc/NetworkManager/conf.d/20-omarchy-dns.conf
+dns_connections_changed=false
 
 provider_from_arg() {
   case "${1:-}" in
@@ -123,28 +124,6 @@ normalize_servers() {
   printf '%s\n' "$*" | tr ',\t\n' ' ' | xargs | tr ' ' ','
 }
 
-split_dns_servers() {
-  local servers="$1"
-  local server clean
-  ipv4_dns=""
-  ipv6_dns=""
-
-  for server in ${servers//,/ }; do
-    clean=${server#dns+tls://}
-    clean=${clean#dns+udp://}
-    clean=${clean%%#*}
-    clean=${clean#[}
-    clean=${clean%]}
-
-    [[ -n $clean ]] || continue
-    if [[ $clean == *:* ]]; then
-      ipv6_dns+="${ipv6_dns:+ }$clean"
-    else
-      ipv4_dns+="${ipv4_dns:+ }$clean"
-    fi
-  done
-}
-
 write_networkmanager_dns() {
   local servers="$1"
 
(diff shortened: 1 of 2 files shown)


VykosMolt:
Moving to global DNS is the right call, and it removes the part I came here about: `set_connection_dns` rewriting `ipv4.dns` on every saved profile.

One case looks like it survives, in `clear_connection_dns`. `connection_uses_dhcp_dns` is true only when all four fields are at their defaults, so a profile the *user* gave static DNS is not skipped and still gets blanked:

```
lab-ethernet   ipv4.ignore-auto-dns yes   ipv4.dns 192.168.50.2   <- set by the user, never by Omarchy
omarchy dns dhcp
lab-ethernet   ipv4.ignore-auto-dns no    ipv4.dns (empty)
```

Nothing on the connection records who wrote those fields, so a cleanup pass cannot tell DNS this tool left behind from DNS that was there first, and the previous value is not recoverable afterwards.

That may be out of scope here, since it is a migration path for damage older builds already did rather than anything this PR introduces. Raising it because the one-time cleanup is the last thing still writing to profiles, and it is the write that cannot be undone. Whether Omarchy wants to own profile DNS outright or leave anything it did not write alone seems like a call for the maintainers rather than something to settle in a patch.

danbosscher:
Resolved the conflict with `quattro` and updated DNS switching to change live systemd-resolved settings without rewriting saved connections or restarting DHCP. Selecting DHCP restores each connection’s original DNS settings, including static DNS. VPN routing and encryption settings remain intact.

Kept the focused policy and privilege tests and removed the custom integration lab from the maintained suite. The retained tests pass; the earlier live-daemon checks also passed.

Issue #6850, opened 2026-08-14
[Quattro] I'm having to enter my password twice on Quattro rc3 and rc4. Is it normal behavior?


neurorune:
### System details

i5-1335U, Iris Xe, Omarchy 4.0.0rc4-1, HP Pavilion Laptop 15-eg3xxx

### What's wrong?

As the title says, I downloaded and installed the RC3. Installation steps were similar to Omarchy 3. I selected encrypt the entire drive. When I log into Quattro it presents me with the same login, I type in the password and get in and then it asks for password again for decrypting the LUKS drive? It never happened in Omarchy 3. I only had to type it once per login and it decrypted. Now it's twice and I don't know how to change this behavior. Is this a bug of RC? Do I've to reinstall when the Stable version releases?

Issue #3185, opened 2025-11-05
When every i wake my laptop from sleep, I always have to enter password two times even if it is correct


shubham-mamodiya-dev:
### System details

Intel 6600u, nvidia Quadro m520, Lenovo p51s, for all omarchy versions

### What's wrong?

It's not frustrating. Omarchy solved most of my problem. Even if it stays that way, I am okay.

naxels:
On and off i have the same thing with a desktop pc, doesn't even have to be wake up from sleep..

voidapex11:
I have experienced the same bug, also on a laptop. I found that if I pressed backspace, my password worked first time, yet the password box looked empty, as if nothing was there.
Like OP, its an inconvenience, but not the end of the world. A fix would be appreciated.

mKay00:
I don't know if this is the same issue or if this has to do with it, but today I suddenly could not log into my machine anymore. 

I used it yesterday with the same password as always, no problem. I let it run over night and tomorrow when I waked up it showed the password screen and I thought this was weird, as it is usually only a black screen. I came back a few minutes later and the screen was black. I tried logging in with my normal password 12 times and it always said wrong password. I even tried another keyboard and with/without capslock and then I tried via ssh from another PC, but it also said wrong password.

I thought about what to do and googled a bit on my second machine and then after some time (20 minutes or so) I tried SSH again and suddenly it worked. I then went to the PC and tried logging in and it worked the first time.

Something really weird is going on here!

solsystemlabs:
I can also confirm this. On my desktop logging in works as expected. On laptop the first attempt _always_ fails when I try to unlock after the laptop has been idle for some period of time. When I manually lock it and then enter my password, it unlocks fine the first try.

cdj5492:
I've noticed this too. The reason the first one fails is that it is ignoring the very first key press or mouse action. Hit anything once then type it in for it to work

pike2k:
> I've noticed this too. The reason the first one fails is that it is ignoring the very first key press or mouse action. Hit anything once then type it in for it to work

I've noticed that if I wait a few seconds the prompt reacts to the first keypress...

tijnster:
Same issue here, when I reboot and lock --> log in again it works from the lock screen. But when its idle for like a day on screensaver mode, i cant login anymore without rebooting

elmanagaev:
[...]

PR #8910, opened 2026-08-29
Avoid redundant lock after encrypted hibernate

### LINKED ISSUE #8909: [Quattro] Encrypted hibernate asks for both LUKS and session passwords
### System details

Omarchy 4.0.1 Quattro, Linux 7.1.9, encrypted Btrfs root using Omarchy's stock `cryptdevice=` initramfs flow, with hibernate enabled through Omarchy Setup.

### Problem

Hibernate resume can require two authentication prompts on a stock encrypted installation. The initramfs LUKS prompt decrypts the root device and restores the hibernation image, then the restored native Omarchy session lock asks for the account password again.

This happens because `omarchy-system-sleep-monitor` receives logind's generic `PrepareForSleep(true)` signal for both suspend and hibernate and normally invokes `omarchy-system-sleep-lock`. Normal reboot has only the LUKS prompt because there is no saved desktop session or session lock to restore.

### Expected behavior

For Omarchy's stock interactive LUKS hibernate path, boot decryption should be the one authentication step. Unencrypted roots, keyfile or automatic unlock, unknown or custom encryption paths, suspend, and direct or unmarked sleep requests must retain the native session lock.

### Current finding

The latest double-authentication trace proves that the sleep monitor saw no live hibernate marker and therefore locked before S4. The old wrapper did not log whether it was bypassed or fell back, so the missing marker's historical cause cannot be reconstructed. Current encrypted-root detection and advisory marker locking pass, and PR #8910 now logs every marker decision and fallback without recording authentication data. A marker-aware hardware retest through the Omarchy menu remains pending.

Related history appears in #3185, #6850, and #141.

ClGratton:
## Problem

On a stock encrypted Omarchy installation, hibernate resume currently asks for authentication twice: once in initramfs to unlock LUKS and again in the restored native session lock.

The sleep monitor sees only logind's generic `PrepareForSleep(true)` signal, so it locks before both suspend and hibernate. For stock interactive LUKS hibernate, the saved session is inaccessible until the boot prompt has already authenticated the user.

Fixes #8909. Related reports appeared in #3185 and #6850; neither has an open fix, and #3185 explicitly asked for a new issue if this reproduced on Quattro.

## Change

This adds `omarchy-system-hibernate` and routes the menu through it. The command recognizes only Omarchy's known interactive boot-auth path: a dm-crypt root, `cryptdevice=` on the kernel command line, and no `cryptkey=`. It holds a live advisory runtime lock while that explicit hibernate request is in flight, allowing the sleep monitor to skip its session lock only while the marker is actively held.

Unencrypted, keyfile-backed, and unknown or custom encryption paths still get the native session lock. Suspend and direct or unmarked sleep requests still lock normally. A stale marker file does nothing, and every detection or marker failure falls back to the existing secure lock path. The monitor remains running throughout and continues to own every other sleep event.

## Verification

The hibernation detection, command, sleep monitor, sleep lock, menu, CLI, and changed shell tests pass. Live detection on the reporting machine identifies its stock encrypted root as interactive boot-auth; unencrypted, keyfile, custom, missing-marker, and stale-marker cases are covered by fail-secure tests.

A later hardware trace still showed both prompts because the sleep monitor received no live marker and therefore correctly requested the desktop lock. The old wrapper did not log why the marker was absent, so that historical distinction cannot be reconstructed. Current detection and marker locking pass, the local GUI action is explicitly routed through the wrapper, and this revision logs every detection, marker, and fallback decision without recording authentication data. A final marker-aware hardware hibernate cycle remains pending.

### DIFF
--- a/bin/omarchy-hibernation-boot-auth
+++ b/bin/omarchy-hibernation-boot-auth
@@ -1,0 +1,18 @@
+#!/bin/bash
+
+# omarchy:summary=Check whether hibernate resume requires interactive boot authentication
+# omarchy:args=[root-device] [kernel-cmdline]
+# omarchy:hidden=true
+
+root_source=${1:-$(findmnt -no SOURCE / 2>/dev/null)}
+root_source=${root_source%%\[*}
+cmdline=${2:-$(< /proc/cmdline)}
+
+# Omarchy's encrypted install maps the root filesystem directly through
+# dm-crypt and unlocks it with mkinitcpio's encrypt hook. With no cryptkey on
+# the kernel command line, restoring a hibernation image necessarily passes
+# through the interactive LUKS prompt first.
+[[ -n $root_source ]] || exit 1
+[[ $(lsblk -dnro TYPE -- "$root_source" 2>/dev/null) == crypt ]] || exit 1
+[[ " $cmdline " == *" cryptdevice="* ]] || exit 1
+[[ " $cmdline " != *" cryptkey="* ]] || exit 1
--- a/bin/omarchy-system-hibernate
+++ b/bin/omarchy-system-hibernate
@@ -1,0 +1,50 @@
+#!/bin/bash
+
+# omarchy:summary=Hibernate the system
+# omarchy:group=system
+
+set -uo pipefail
+
+log_hibernate_path() {
+  logger --tag omarchy-system-hibernate -- "$*" 2>/dev/null || true
+}
+
+# A session lock is still the only authentication after resume on an
+# unencrypted install, or on an encrypted install that unlocks automatically.
+# Those paths go straight through the ordinary sleep monitor.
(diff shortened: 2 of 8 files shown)


ClGratton:
A later hardware cycle still showed both the boot-time LUKS prompt and the restored desktop lock. The journal proves that the sleep monitor received no live marker and requested the desktop lock before hibernation. It does not prove why the marker was absent because the previous wrapper logged none of its fallback decisions. Current checks on the affected machine confirm that encrypted-root detection succeeds, XDG_RUNTIME_DIR is writable, and the marker can be locked. The Omarchy menu is now explicitly routed through omarchy-system-hibernate locally. I updated this branch so future cycles log whether boot authentication was detected, whether the marker was held, and the reason for any fallback, without logging authentication data. The existing hibernate and menu tests still pass.

PR #8913, opened 2026-08-29
Add Omafox to Firefox and Zen Browser for theme syncing

cover:
### Summary
- Add the [Omafox](https://github.com/cover/omafox) [AUR package](https://aur.archlinux.org/packages/omafox) and [Firefox/Zen extension](https://addons.mozilla.org/en-US/firefox/addon/omafox/) which keeps the theme in sync with Omarchy

### Test plan
- `bash test/shell.d/browser-policy-dir-test.sh`
- `bash test/shell.d/omafox-browser-integration-test.sh`

### DIFF
--- a/bin/omarchy-install-browser
+++ b/bin/omarchy-install-browser
@@ -29,6 +29,11 @@ setup_firefox_wayland() {
   echo "MOZ_ENABLE_WAYLAND=1" > ~/.config/environment.d/omarchy-firefox-wayland.conf
 }
 
+setup_omafox() {
+  omarchy-pkg-aur-add omafox
+  omafox setup
+}
+
 case $1 in
 chromium)
   echo "Installing Chromium..."
@@ -81,6 +86,7 @@ firefox)
 
   browser_policy_setup_firefox_distribution /usr/lib/firefox/distribution
   setup_firefox_wayland
+  setup_omafox
   announce_browser_installed "Firefox"
   ;;
 zen)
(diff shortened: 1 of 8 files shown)


Jamesking56 (changes requested):
I don't like the fact that this is reliant on both an AUR package and FireFox extension that the Omarchy team doesn't manage. Given the recent AUR security issues and browser extensions that have also had their fair share of bad actors, I'd prefer it if this was configured via bash only (which #10867 does)

cover:
> I don't like the fact that this is reliant on both an AUR package and FireFox extension that the Omarchy team doesn't manage. Given the recent AUR security issues and browser extensions that have also had their fair share of bad actors, I'd prefer it if this was configured via bash only (which #10867 does)

That's fair, I've followed the approach of omazed, which is also [installed](https://github.com/omacom/omarchy/blob/2fbac0c8e88eca704af1650ce721a494bd11a3d0/bin/omarchy-install-editor-zed#L6) as an [AUR package](https://aur.archlinux.org/packages/omazed). I think the limit of that PR is that it forces a [restart](https://github.com/omacom/omarchy/pull/10867/changes#diff-6c163b4920c16fa8503672f82f40af9f4cffa843f7f5384788717105d00e942bR77) of the browser when the theme changes, which is not ideal if you're doing something in the browser and it gets killed and restarted (and also that it doesn't support Firefox)

(FWIW, I wouldn't mind transferring both the repo and package/extension under the Omarchy org)

en3r0:
following

PR #8927, opened 2026-08-29
Anchor dip and lip matches on the ssh binary

### LINKED ISSUE #8917: dip kills unrelated processes: pkill -f pattern is unanchored and matches any command line
## What happened

`dip <port>` can kill processes that are not SSH port forwards — including the shell it was typed in.

## Cause

In `default/bash/fns/ssh-port-forwarding`, `dip()` runs:

```bash
pkill -f "ssh.*-L ${port}:localhost:${port}" && echo "Stopped forwarding port $port" || echo "No forwarding on port $port"
```

`pkill -f` matches the pattern against the **full command line of every process on the system**, not just `ssh` processes. The pattern is unanchored, so any process whose command line happens to contain that text is killed too.

This is easy to hit unintentionally: a shell running a script that mentions the port, an editor with the string in its arguments, or a terminal whose command line includes the forwarding spec will all match.

## What was expected

Only the SSH forward for that port is terminated.

## Steps to reproduce

```bash
# A process that is not ssh but whose command line contains the pattern
bash -c "exec -a 'notssh -L 9000:localhost:9000' sleep 300" &

dip 9000
# -> reports "Stopped forwarding port 9000" and kills the unrelated process
```

I hit this for real while testing: `dip` matched the shell running the test — whose command line contained the port string — and killed it.

## Fix

Anchor the pattern to an actual `ssh` process:

```bash
pkill -f "^([^ ]*/)?ssh .*-L ${port}:localhost:${port}"
```

`lip()` has the same unanchored pattern and will likewise list unrelated processes; the same anchor applies:

```bash
pgrep -af "^([^ ]*/)?ssh .*-L [0-9]+:localhost:[0-9]+"
```

Verified: with the anchor, a decoy process carrying the same text in its command line survives while the real forward is stopped.

## System

- Omarchy 4.0.1-1
- `/usr/share/omarchy/default/bash/fns/ssh-port-forwarding`, line 14

Chessing234:
## Summary
- `dip` / `lip` now match `^ssh … -L port:localhost:port` so an unanchored `pkill -f` cannot kill the shell that typed `dip`, or any other argv that only mentions the fragment (#8917).

## Test plan
- [x] `test/shell.d/ssh-port-forwarding-test.sh` checks the anchored pattern accepts a real ssh forward and rejects a non-ssh cmdline.
- [ ] `fip host 2222` then `dip 2222` stops only that forward.

Fixes #8917

### DIFF
--- a/default/bash/fns/ssh-port-forwarding
+++ b/default/bash/fns/ssh-port-forwarding
@@ -13,11 +13,12 @@ dip() {
   for port in "$@"; do
     # Anchor on the ssh binary name so an unanchored `pkill -f` cannot match an
     # unrelated argv that merely mentions the same -L fragment (e.g. this shell
-    # while typing `dip`, or a script that embeds the pattern) (#8917).
-    pkill -f "^ssh .* -L ${port}:localhost:${port}( |$)" && echo "Stopped forwarding port $port" || echo "No forwarding on port $port"
+    # while typing `dip`, or a script that embeds the pattern) (#8917). The
+    # trailing ( |$) keeps `dip 222` off a forward on port 2222.
+    pkill -f "^([^ ]*/)?ssh .*-L ${port}:localhost:${port}( |$)" && echo "Stopped forwarding port $port" || echo "No forwarding on port $port"
   done
 }
 
 lip() {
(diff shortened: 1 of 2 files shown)


Chessing234:

Issue #7889, opened 2026-08-23
Bluetooth remains pairable after the Omarchy panel is closed


bosse4:
## System details

- Hardware: Lenovo ThinkPad L480
- Omarchy: 4.0.0-1
- Kernel: 7.1.8-arch1-3
- BlueZ: 5.87-2

## What's wrong?

With Bluetooth powered on and the Omarchy Bluetooth panel closed, the adapter remains pairable indefinitely:

```console
$ bluetoothctl show | grep -E 'Powered|Discoverable|Pairable|Discovering'
Powered: yes
Discoverable: no
Pairable: yes
Discovering: no
```

`PairableTimeout` uses the BlueZ default of `0`, so pairable mode does not expire. The default pairing agent runs for the entire graphical session:

```console
$ systemctl --user status bt-agent.service
/usr/bin/bt-agent -c NoInputNoOutput
```

## Expected behavior

The comment in `/usr/lib/systemd/user/bt-agent.service` says:

> NoInputNoOutput auto-accepts pair requests. Safe because the adapter is only `pairable: true` when the user explicitly opens the omarchy bluetoothPanel and starts scanning; outside that window bluez refuses inbound pair attempts at a lower layer.

The adapter should therefore be non-pairable whenever the Bluetooth panel is closed.

## Actual code path

`/usr/share/omarchy/shell/plugins/panels/bluetooth/Panel.qml` starts discovery while the panel is open and stops it after closing by writing:

```qml
root.adapter.discovering = true
root.adapter.discovering = false
```

However, the panel never reads or writes `adapter.pairable`. Quickshell exposes `pairable` as a writable adapter property, but BlueZ remains at its default `Pairable: true`.

## Security impact

A nearby device that knows the adapter's public Bluetooth address may initiate an incoming pairing attempt while the panel is closed.

Because the default agent advertises `NoInputNoOutput`, pairing may fall back to “Just Works” without user confirmation, depending on the peer and negotiated security requirements. `Discoverable: no` reduces exposure but does not enforce the security boundary described by the service comment.

No pairing attempt or exploit was performed while investigating this issue.

## Minimal fix

Tie `adapter.pairable` to the Bluetooth panel lifecycle:

- Set it to `true` while at least one Bluetooth panel is open and the adapter is powered.
- Set it to `false` after the last panel closes, is destroyed, or Bluetooth is turned off.
- Preserve the existing multi-monitor discovery ownership logic.
[...]

Issue #6992, opened 2026-08-15
omarchy-upgrade-to-quattro enables bt-agent.service but orphans bluez-tools (its only source of /usr/bin/bt-agent) — infinite restart loop


lemachinarbo:
_Note: issue reported by AI._

## What happened

After the Quattro migration, `bt-agent.service` restart-loops forever
(every 2s, indefinitely — 668 restarts in the first 25 minutes of a boot on
my machine):

    bt-agent.service: Unable to locate executable '/usr/bin/bt-agent': No such file or directory
    bt-agent.service: Failed at step EXEC spawning /usr/bin/bt-agent: No such file or directory
    bt-agent.service: Main process exited, code=exited, status=203/EXEC
    bt-agent.service: Failed with result 'exit-code'.
    (repeats forever)

## Root cause (traced through my own pacman.log + the migration script)

`/usr/bin/bt-agent` is provided by `bluez-tools`, not `bluez-utils`. On my
machine `omarchy-upgrade-to-quattro` ran a transaction that:

1. Installed `bluetui`
2. Removed `blueberry` via `pacman -Rns`, which cascaded and removed
   `gnome-bluetooth` and `bluez-tools` as now-orphaned dependencies
   (`bluez-tools` was never installed for its own sake — it came in
   transitively via blueberry/gnome-bluetooth)
3. Later in the same script, `enable_user_unit bt-agent.service`
   (bin/omarchy-upgrade-to-quattro, retiring
   `systemd/user/bt-agent.service` -> `/usr/lib/systemd/user/bt-agent.service`,
   shipped by `omarchy-settings`)

`omarchy-settings` does not depend on `bluez-tools`:

    $ pacman -Qi omarchy-settings | grep Depends
    Depends On      : bash  curl  gum  hicolor-icon-theme  plymouth

So the same migration that retires blueberry (bt-agent's last remaining
dependent) also turns on a unit that requires the binary blueberry was
providing indirectly, with nothing left to guarantee it's still installed.
Anyone who goes through this migration and doesn't happen to have
`bluez-tools` pulled in by something else ends up with a permanently
crash-looping user unit.

## Expected

Either:
- `omarchy-settings` should depend on `bluez-tools` (or whatever package
  currently provides `bt-agent` upstream), or
- `bt-agent.service` should have an `ExecCondition` that fails cleanly when
  the binary is missing, the way the neighboring bluetooth condition
  (`ExecCondition=systemctl is-active --quiet bluetooth.service`) already
[...]

PR #8929, opened 2026-08-29
Feed bt-agent stdin so Bluetooth pairing authorization can be answered

x3m:
## What's wrong

`default/systemd/user/bt-agent.service` says `-c NoInputNoOutput` auto-accepts pair requests. It does not.

For JustWorks pairing bluez calls the agent's `RequestAuthorization`, and bluez-tools' `bt-agent` implements that by prompting

```
Authorize this device pairing (yes/no)?
```

on **stdin**, regardless of the capability it was started with. Under systemd stdin is `/dev/null`, so the prompt is never answered and every pairing times out.

This is the only pairing agent Omarchy ships — the Quickshell shell registers none, and `install/user/first-run/enable-user-units.sh` enables this unit — so on a stock install no BLE peripheral that needs JustWorks authorization can pair.

## Symptom

The device is left trusted-but-unbonded, then reconnects and drops roughly once per second, filling the journal:

```
bluetoothd: profiles/deviceinfo/deviceinfo.c:read_pnpid_cb() Error reading PNP_ID value:
            Request attribute has encountered an unlikely error
bluetoothd: profiles/input/hog-lib.c:info_read_cb() HID Information read failed: ...
bt-agent:   Authorize this device pairing (yes/no)? Device: Logi M850 L (DB:D8:...)
```

Those ATT errors are insufficient-authentication surfacing: HID-over-GATT needs an encrypted link, which needs a bond that never formed.

## The fix

Pipe `yes` into the agent, which is what actually makes it auto-accept — the behaviour the unit's `Description` and comment already claim. The existing safety rationale is unchanged and kept in the comment.

## Verification

Omarchy 4.0.1, bluez 5.87, Logitech M850 L (BLE HID). Pairing failed in a loop for several minutes with the stock unit. With stdin fed, `bluetoothctl pair` completed immediately:

```
Paired: yes
Bonded: yes
Connected: yes
```

Only stdin differed between the two attempts.

## Relationship to existing issues

- #7879 — "Bluetooth panel treats trusted-but-unbonded devices as paired and never retries pairing" describes the **aftermath**: `omarchy-bluetooth-device pair` trusts the address even when pairing failed, so the device is thereafter only ever connected, never re-paired. This PR removes the reason the first pair fails for BLE HID devices in the first place. The two are complementary — #7879 is still worth fixing on its own, since any other pair failure leads to the same trap.
- #6992 — reports this same unit broken for an unrelated packaging reason (orphaned `bluez-tools`).

### DIFF
--- a/bin/omarchy-bluetooth-device
+++ b/bin/omarchy-bluetooth-device
@@ -29,6 +29,31 @@ trust_device() {
   bluetoothctl trust "$address" >/dev/null 2>&1 || true
 }
 
+# Trust is what lets BlueZ reconnect a device on its own later, and that only
+# works once a bond (keys kept on both sides) exists. A device trusted without
+# one is worse than unknown: BlueZ auto-connects it, the pairing fails the same
+# way every few seconds, and the journal fills until the user forgets it. So
+# trust follows a bond, never precedes it.
+bonded() {
+  [[ $(timeout 5s bluetoothctl info "$address" 2>/dev/null) == *"Bonded: yes"* ]]
+}
+
+# The bond is not necessarily there when `pair` or `connect` return. Some
+# peripherals ignore a host-initiated pair and only bond through their own
+# security request once connected, so `pair` times out and the bond forms during
+# `connect`; and `connect` itself returns at once with InProgress when BlueZ is
+# already connecting a device it knows. Closing the window on either return
+# leaves a session-only pairing (Paired but not Bonded) that is gone at the next
+# disconnect. Waiting on the bond itself is what keeps the window open for as
+# long as the handshake actually takes.
+wait_for_bond() {
+  local deadline=$((SECONDS + ${OMARCHY_BLUETOOTH_BOND_WAIT_SECONDS:-15}))
+  until bonded; do
+    ((SECONDS < deadline)) || return 1
+    sleep 1
+  done
+}
+
 # bt-agent.service auto-accepts JustWorks pairing authorization, so the adapter
 # must not sit pairable outside a pair the user actually asked for. Nothing else
 # in Omarchy touches this property and BlueZ defaults it on, which leaves the
@@ -46,12 +71,12 @@ case "$action" in
     power_on
     timeout 5s bluetoothctl pairable on >/dev/null 2>&1 || true
     timeout 20s bluetoothctl pair "$address" >/dev/null 2>&1 || true
-    trust_device
     timeout 20s bluetoothctl connect "$address" >/dev/null 2>&1 || true
+    if wait_for_bond; then trust_device; fi
     ;;
   connect)
     power_on
-    trust_device
+    if bonded; then trust_device; fi
     timeout 20s bluetoothctl connect "$address" >/dev/null 2>&1 || true
     ;;
   disconnect)
--- a/bin/omarchy-bluetooth-device
+++ b/bin/omarchy-bluetooth-device
@@ -29,9 +29,22 @@ trust_device() {
   bluetoothctl trust "$address" >/dev/null 2>&1 || true
 }
 
+# bt-agent.service auto-accepts JustWorks pairing authorization, so the adapter
+# must not sit pairable outside a pair the user actually asked for. Nothing else
+# in Omarchy touches this property and BlueZ defaults it on, which leaves the
+# adapter permanently open to unsolicited pair attempts. Closing it on EXIT means
+# every invocation of this script converges the adapter to non-pairable, whatever
+# the action and whichever way it ends - normal return, `timeout` expiry, or a
+# signal - and only the pair window below opens it.
+pairable_off() {
+  timeout 5s bluetoothctl pairable off >/dev/null 2>&1 || true
+}
+trap pairable_off EXIT
+
 case "$action" in
   pair)
     power_on
+    timeout 5s bluetoothctl pairable on >/dev/null 2>&1 || true
     timeout 20s bluetoothctl pair "$address" >/dev/null 2>&1 || true
     trust_device
     timeout 20s bluetoothctl connect "$address" >/dev/null 2>&1 || true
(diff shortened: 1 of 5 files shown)


x3m:
Good catch on the pairable state — the review is right, and I've pushed a fix in 4c8b11f.

I verified the claim rather than taking it on faith, and it holds three ways:

1. **Nothing in the tree ever sets `Pairable`.** `grep -rniE 'pairable'` across the repo returns exactly one hit: the comment in the unit I was editing. `Panel.qml` manages `discovering`, never `pairable`.
2. **Confirmed on a running install**, panel closed: `bluetoothctl show` reports `Pairable: yes`.
3. **#7889** is open and titled "Bluetooth remains pairable after the Omarchy panel is closed".

So the rationale the unit carried was aspirational, and that made my first commit a regression rather than just a fix: the broken agent at least failed *closed* and never accepted anything, whereas piping `yes` accepts every JustWorks `RequestAuthorization` on an adapter that is permanently open. Worst case is an unsolicited HID keyboard bond, which is arbitrary input injection.

Rather than document the hole, the new commit makes the guard real:

- **`omarchy-bluetooth-device`** opens the window only around an explicit `pair` and closes it from an `EXIT` trap, so it is restored on every path out — normal return, `timeout` expiry, or a signal. The trap is unconditional, so *every* invocation converges the adapter to non-pairable whatever the action; that also closes #7889.
- **`bt-agent.service`** gains `ExecStartPost=-/usr/bin/timeout 5 /usr/bin/bluetoothctl pairable off`, clearing it once at login for a session where that script has not run yet.
- The unit comment now describes the guard that exists instead of one that did not.

I went with bounding the adapter state rather than your first suggestion of scoping acceptance to the requested device, because `bt-agent` registers a single global agent with no per-device hook — scoping it there would mean replacing the agent entirely, which is a much larger change than this bug warrants.

**Tests:** five new assertions in `bluetooth-test.sh` cover that `pair` opens the window, opens it *before* attempting the pair, and closes it afterwards; and that `connect` never opens it yet still closes it. `bluetooth-test.sh` and `systemd-test.sh` pass. `config-test.sh` fails identically with and without this change (it wants an `omarchy-pkgs` checkout this environment lacks), verified by stashing.

x3m:
Correction to my previous comment: the `ExecStartPost` I described **did not actually work**, and I only caught it by testing on a live adapter instead of trusting the exit status. Fixed in c69d3ce.

The journal shows the race:

```
Starting Bluetooth pairing agent (auto-accept)...
timeout[...]: Changing pairable off succeeded
sh[...]: Agent registered
sh[...]: Default agent requested
Started Bluetooth pairing agent (auto-accept).
```

`ExecStartPost` runs as soon as `ExecStart` is *spawned* — before `bt-agent` registers with bluez — and registering the default agent leaves the adapter pairable. So the clear landed first and was undone immediately after. It reported success, the journal looked correct, and `Pairable` read `true` regardless. Running the identical command *after* registration sticks.

I measured with `busctl` rather than `bluetoothctl` to rule out the probe being the cause: adapter set pairable, unit started from `inactive`, property still `true` five seconds later. With a settle delay it goes `false` and stays `false` across repeated reads.

A fixed sleep is unlovely, but `bt-agent` gives no readiness signal (it isn't `Type=notify`), so there's nothing better to wait on. Happy to swap it for something else if you'd prefer — a `bluetooth-pairable.service` ordered `After=`, or dropping the `ExecStartPost` entirely and leaning only on the `omarchy-bluetooth-device` trap, which is the load-bearing half either way.

The new `systemd-test.sh` assertion guards the delay specifically, because losing it fails silently in the worst way: the unit still starts, the journal still says "succeeded", and the adapter quietly stays open.

---

One unrelated observation while testing, **not addressed here** since it predates this PR: `bt-agent` ignores `SIGTERM`, so `systemctl --user stop bt-agent` hangs until `TimeoutStopSec` and the process is finally `SIGKILL`ed. I verified the stock unit behaves identically with my drop-in removed, so it isn't introduced by the `sh -c` wrapper — but it does mean `systemctl restart` takes ~90s, which is worth knowing if you test this by hand. Might warrant its own issue.

x3m:
Two more fixes pushed, from a real case that hit this exact flow on 2026-09-03: a Keychron K3 that had lost its bond after a suspend/resume.

**The window closed before the bond formed** (7b4741e). `connect` returned at once with `InProgress`, because the `trust` a line earlier had already made BlueZ start connecting the device; the EXIT trap closed the window, and the kernel finished the handshake non-bondable: `Paired: yes, Bonded: no`. That pairing was gone at the keyboard's next disconnect. The K3 also ignores a host-initiated `Pair()` altogether (kernel: `unexpected SMP command 0x0b`) and only bonds through its own security request during `connect`, so the window has to outlast `connect`, not just `pair`. The script now waits on the device reporting `Bonded: yes` (bounded, 15s, `OMARCHY_BLUETOOTH_BOND_WAIT_SECONDS` for tests) before the trap closes it.

**Unconditional `trust` built the reconnect loop** (same commit). A device trusted without a bond is auto-connected by BlueZ and fails identically every ~4s — 33 uhid registrations and as many `Authorize this device pairing` prompts in one afternoon's journal — until the user forgets it. Trust now follows the bond, in `pair` and `connect` both.

**The panel routed that stuck state to `connect`** (b0e0e25). `connectDevice` treated `trusted` alone as connectable, and `connect` never opens the window, so from the panel the device could never heal. Trust alone now goes through `pair`.

Nine assertions cover the three: the bond is checked after `connect` and before the close; trust follows a bond and never a failed pair, on `pair` and on `connect`; and the panel condition is pinned to paired-or-bonded. Each new assertion was checked to fail against the previous commit. `bluetooth-test.sh`, `bin-style-test.sh` and `systemd-test.sh` pass. The full `test/shell` run has the same four failures with and without these commits (`config-test` and `unowned-system-paths-test` want an `omarchy-pkgs` checkout; `snapper-test` and `theme-install-guards-test` fail identically on the base).

Chessing234:
heads up this is the same RequestAuthorization gap as #12897. that one's a replacement agent rather than feeding stdin — please close this as a duplicate of that.

Issue #8361, opened 2026-08-26
Lock screen loses keyboard focus after lid open (resume); requires a mouse click to type


andersmar:
## Description

After the laptop lid is opened (screen was locked at lid close, then the machine suspended), the lock screen's password field does not have keyboard focus. Typing does nothing until the mouse is clicked once inside the lock screen. Moving the mouse without clicking does not restore focus either.

## Steps to reproduce

1. Close the laptop lid (this locks the screen and the machine suspends).
2. Wait, then open the lid.
3. Without touching the mouse, try typing the password — nothing appears in the field.
4. Click anywhere on the lock screen, then type — now it works.

## Expected behavior

The password field should already have keyboard focus (or regain it automatically on wake) after opening the lid, without requiring a mouse click.

## Root cause (found while investigating)

In `/usr/share/omarchy/shell/plugins/lock/LockView.qml`, keyboard focus on `passwordInput` is only requested in two places:

- `Component.onCompleted` — fires once, when the lock QML item is first created (at lid-close time, before suspend).
- `MouseArea.onClicked` — explicitly calls `root.forcePasswordFocus()`.

`MouseArea.onPositionChanged` (mouse move) and `Keys.onPressed` only emit `wakeRequested()`, which in `Service.qml`'s `runWake()` just runs `omarchy-system-wake` (turns the display back on) — it never calls `forcePasswordFocus()`. Hyprland appears to drop keyboard focus on the lock's layer-shell surface across the suspend/resume cycle, and nothing re-requests it afterward except an explicit click.

## Workaround

Cloned the plugin (`omarchy plugin clone omarchy.lock`) and added a `root.forcePasswordFocus()` call to `MouseArea.onPositionChanged` alongside the existing `onClicked` handler, so any mouse movement after wake also re-asserts focus. Confirms the diagnosis — with that change, moving the mouse (no click needed) restores typing.

## System

- Omarchy: 4.0.1-1
- Kernel: 7.1.9-arch1-2
- CPU: Intel(R) Core(TM) Ultra 9 185H
- GPU: Intel Meteor Lake-P (Arc Graphics) + NVIDIA GeForce RTX 4060 Max-Q (hybrid graphics)
- Machine: Lenovo 83DN
- `omarchy debug --no-sudo --print` is not a recognized command in this version (`omarchy commands --all` has no `debug` entry), so no diagnostic log is attached.

gitdayn:
[...]

Issue #8628, opened 2026-08-27
Lock screen: password field loses focus after display wake, recoverable only with a mouse


Rodolife:
## Summary

After the lock screen blanks the display, waking it does not restore focus to
the password field. Keystrokes go nowhere. The only way to recover is to click
with a mouse or tap the touchpad, which defeats keyboard-only operation.

## System details

- Omarchy: 4.0.1-1
- Hyprland: 0.56.2 (commit efb5099)
- quickshell: 0.3.1-1
- CPU: Intel Core i9-9980HK
- GPU: Intel UHD Graphics 630 + NVIDIA RTX 2080 Mobile
- Single internal display (laptop)

## Steps to reproduce

1. Leave the machine idle until it locks (`idle.lock` in `shell.json`), or run
   `omarchy system lock`.
2. Wait ~5s more for `idleBlankTimer` to blank the display.
3. Wake the machine by pressing a key.
4. Type your password.

**Expected:** characters appear in the password field.

**Actual:** nothing happens. The field is visibly unfocused. Only a mouse click
or touchpad tap restores focus, after which typing works normally.

## Root cause

In `shell/plugins/lock/LockView.qml`, focus is asserted in exactly two places:

```qml
onInputEnabledChanged: {
  if (inputEnabled) Qt.callLater(forcePasswordFocus)
}
Component.onCompleted: {
  syncPasswordText()
  if (inputEnabled) Qt.callLater(forcePasswordFocus)
}
```

`inputEnabled` is bound to `root.lockRequested` (`Service.qml`, in the
`WlSessionLockSurface`). That flips true once when the session locks and then
stays true for the whole lock, so `onInputEnabledChanged` never fires again.
`Component.onCompleted` likewise runs only once.

Meanwhile `idleBlankTimer` calls `runBlank()` ~5s after locking. When the
display comes back, neither trigger re-fires, so nothing re-asserts focus.

The only remaining path is the mouse:

```qml
MouseArea {
  anchors.fill: parent
  hoverEnabled: true
  onClicked: { root.wakeRequested(); root.forcePasswordFocus() }
  onPositionChanged: root.wakeRequested()
}
```

Note `onClicked` is the sole caller of `forcePasswordFocus` after lock time.

This is self-reinforcing: `TextInput` has a `Keys.onPressed` handler that calls
`root.wakeRequested()`, but an unfocused `TextInput` never receives key events,
so the keyboard can never recover focus on its own.

## Suggested fix

[...]

PR #8930, opened 2026-08-29
Harden lock lifecycle, recovery, and keyboard wake

AFOliveira:
Preserve a live session-lock owner through plugin reload, disable/removal and service synchronization, and recover password focus after mapping or keyboard wake. This addresses the lock lifecycle and input paths behind #6888, #7106, #6858, #7467 and #8628; hardware-specific outcomes still need validation.

The integration keeps authentication services in the current private store and exposes only ownership booleans to lifecycle code. It preserves an owning locker, defers reload until unlock, retains bounded focus retries and serialized blank/wake behavior, and bounds restart shutdown with one monotonic deadline. Restart requires confirmed absence before launching a replacement and retains #9469's notification readiness check.

Installed VM testing caught and fixed an additional ordinary-restart regression: Quickshell returns `No running instances.` with status zero even in JSON mode. The dispatcher now recognizes only that exact successful result under a fixed locale; malformed output and query failures remain errors.

**Integrate after #9469.** This branch includes current Quattro and that prerequisite through an ordinary merge. It consolidates relevant work from #7169, #7572, #7471 and #7592 while retaining their authors as co-authors. The narrower focus-only #8869 overlaps and must be reconciled before applying both.

At final `4aea51f2ddf3bef230626cf23d182a38f2284d5d`, independent Daybreak review found no remaining code blocker. Focused lock, reload, private-auth, restart, sleep, video and update regression suites plus 112 CLI checks passed. Actual candidate `/usr/bin` commands and QML in a disposable 4.0.2 VM passed normal restart with one replacement process, secure lock acquisition, rescan preserving its owner, refusal to restart an owned lock, and keyboard password unlock. Inspected screenshots and a short transition recording support the visual/focus smoke. Guest files were restored afterward.

Keep draft for affected-hardware validation: ASUS TUF keyboard wake, Strix Halo/multiple displays, physical DPMS, suspend/resume, hotplug, fingerprint and crash-recovery paths were not established on this final revision. Source tests and a single virtual display cannot establish those outcomes.


azure-eller:
**Reproduces here; the PR applies cleanly to current `quattro`, breaks nothing in the suite, and closes three more issues than its body lists.**

ASUS TUF Gaming A16 (FA608UP), Omarchy 4.0.1-1, Hyprland 0.56.2-1, Quickshell 0.3.1-1, kernel 7.1.9, single panel on amdgpu. Every lid-open resume leaves the lock screen drawn but keyboard-dead; a click anywhere restores typing.

That symptom localizes it precisely: `LockView.qml:118` `onClicked` calls `forcePasswordFocus()`, `:119` `onPositionChanged` does not — so pointer motion wakes the display but never the keyboard. Not compositor seat focus. It is the one-shot grab this PR replaces: `inputEnabled` (`Service.qml:277`, bound to `lockRequested`) never toggles again for the life of a lock, so nothing re-arms it.

Checked before commenting:

- Both commits cherry-pick onto `quattro` @158e8cfb with no conflicts — the branch is 34 behind, but clean.
- `test/all` on the rebased branch: 212 of 217 files pass. All 5 failures fail identically on plain `quattro` from the same checkout path, so none are from this PR — 4 want a sibling `omarchy-pkgs` checkout, and `windows-vm-mount-boundary-test.sh` cannot run from any checkout under `/home` because it tmpfs-mounts over `/home` in its namespace. All 8 lock tests pass.

**Also closes #8361, #8520, #7811** — worth adding to `Fixes:` so all eight clear on merge.

#8869 is the same bug with a narrower, unbudgeted retry; this supersedes it. Its one good extra, `focus: true` on the `TextInput`, would make a fine follow-up — not worth holding this up for.

cyppe:
Real-world confirmation of the #7106 failure mode this PR targets: two stranded locks in 24 h on 4.0.2-1 (quickshell 0.3.1-1, Hyprland 0.56.2-1, AMD Strix Halo, dual monitor), both caused by a dotfile manager writing a local plugin over SSH while the desktop was idle-locked — full journal excerpts and versions in https://github.com/omacom/omarchy/issues/7106. One detail that supports the "preserve the owning service / defer the reload" approach: the in-process recovery after the reload stalled indefinitely at `lock-pending: screen-stabilizing`, whereas a freshly launched shell process completed the same recovery to `secure=true` in ~6 s. Glad to test this branch's behavior with a plugin write while locked once it lands in a release.

PR #8932, opened 2026-08-29
Tile the Battle.net client instead of floating it

michielvandermeer:
## Summary

- Tile the main Battle.net window like other apps. The previous `float + center + size {1280, 800}` rule, combined with Wine's X11 ConfigureRequest under `GDK_SCALE=2` and `xwayland.force_zero_scaling`, parked the client at `x ~= -2 * monitor width` on 1080p. Launch looked like a brief splash, then nothing.
- Keep **Battle.net Login** and **Battle.net Setup** floating. Tiling those resizes the Qt shell and destroys the CEF login webview (empty logo box).
- Suppress `x11configurerequest` (and `maximize`) so Proton cannot shove the tiled window off-screen or undo the layout.
- Float empty-title XWayland children with `stay_focused` + `min_size` so dropdowns do not vanish, same pattern as Steam.
- `omarchy-launch-battlenet` focuses an already-running window and moves it to the current workspace. A second `umu-run` just logs "another instance is running" and exits.
- Desktop `StartupWMClass` is `steam_app_battlenet`, the class Proton actually uses.

## Reproduction

On Omarchy 4.0.1, 1920x1080, `GDK_SCALE=2`:

1. Install Battle.net from the menu, then launch it from the app launcher.
2. Before: Hyprland listed `steam_app_battlenet` / `Battle.net` as a 1280x800 float at **(-3520, 153)** — one `GDK_SCALE * monitor_width` off the left edge. After: the same window tiles on the current workspace and the CEF UI renders (home, games, friends).
3. Launching again focuses that window instead of spawning a second Proton prefix that immediately exits.

Login and setup stay floating dialogs. Games launched from Battle.net do not match `^Battle\.net$`, so they are unaffected.

## Tests

- `bash test/shell.d/battlenet-test.sh`
- `luac -p default/hypr/apps/battlenet.lua`
- `bash -n bin/omarchy-launch-battlenet`
- Live: tiled next to Steam at 941x1030 with a working CEF home/games view; second launch focused the existing window (`umu-run` count unchanged).

### DIFF
--- a/bin/omarchy-launch-battlenet
+++ b/bin/omarchy-launch-battlenet
@@ -8,6 +8,7 @@ set -e
 
 PREFIX="$HOME/Games/battlenet"
 LAUNCHER="$PREFIX/drive_c/Program Files (x86)/Battle.net/Battle.net Launcher.exe"
+WINDOW_CLASS="steam_app_battlenet"
 
 with_mangohud=0
 for arg in "$@"; do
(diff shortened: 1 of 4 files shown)


feliperezende:
Confirmed this PR covers the "dropdowns vanish" case. Tested live and it fixes the DOB dialog too.

Repro (Omarchy 4.0.3-1, Hyprland 0.56.2, GE-Proton11-6 via umu-launcher, Intel Haswell iGPU + AMD Radeon):

1. Launch Battle.net from the menu and get to the "Please verify your date of birth to continue" dialog.
2. Click the Day/Month/Year dropdown.
3. The dropdown is a separate XWayland window (class `steam_app_battlenet`, title empty). With no rule it gets tiled full-width and grabs focus; Chromium then auto-closes it ~150-250 ms later, so the list never renders on screen.

Live evidence from Hyprland IPC (event order): `openwindow` (empty-title window) -> focus steal (`activewindow` = that window) -> focus returns to `Battle.net` -> `closewindow`.

Tested rule, applied via `hyprctl eval` and working:

```lua
o.window({ class = "^steam_app_battlenet$", title = "^$", xwayland = true }, {
  float = true,
  no_focus = true,
})
```

Key detail: the match MUST include `title = "^$"`. The empty-title popup reuses the parent class, so a class-only rule also hits the main client - same trap PR #7020 documents for Java AWT popups.

One question on rule choice: I verified `no_focus`; this PR uses `stay_focused` + `min_size`. Both should stop the vanish, but it's worth confirming the chosen rule doesn't let Chromium see a focus activation before it closes the popup. Happy to test the exact rule from this PR if useful.

PR #8934, opened 2026-08-29
Secure Plymouth and SDDM asset publication

ErikMelton:
Supersedes #8420 because the original author is unavailable. This branch preserves Afonso Oliveira's three commits and adds follow-up commits for the repository-required `$OMARCHY_PATH` invocation and the remaining pre-hash content race.

## Summary

`omarchy-plymouth-set` staged assets in a user-owned temporary directory, then let privileged copies reopen those paths after validation. A same-user process could replace staged content during the ImageMagick window, causing root to publish attacker-selected bytes, preserve a planted symlink in the Plymouth theme, or dereference it during the later SDDM copy.

## Fix

- Open the selected logo in the unprivileged caller and pass only its descriptor to root, so a replacement symlink can never make root resolve the caller path.
- Build the authoritative Plymouth and SDDM assets in a root-owned mode-0700 directory. No user-owned stage is trusted or copied into the boot theme.
- Require packaged source files and their complete directory chain to be canonical, root-owned, and non-writable by group or other before root reads them.
- Preserve the supported `omarchy dev link` workflow only when a trusted, root-owned `/etc/omarchy.conf` contains the exact canonical checkout path. Missing, stale, or writable authorization files fail closed; this exception does not widen dev mode's existing decision to run privileged Omarchy commands from that user-editable checkout.
- Generate executable Plymouth and SDDM configuration only from those trusted packaged sources inside the privileged transaction. The caller-supplied logo remains opaque data at fixed `logo.png` destinations.
- Publish a fixed allowlist through root-created temporary files in validated destination directories, then atomically replace the destination.
- Replace planted destination symlinks without following or changing their victims.
- Reject symlinked, non-canonical, non-root-owned, or group/world-writable destination parents.
- Route `omarchy-refresh-plymouth` through the same publisher, including the nested `logos/oma.png` asset.
- Normalize published ownership and mode, including under restrictive caller umasks.
- Keep SDDM color substitution collision-free for the White theme case from #8469.

## Verification

- `bash -n bin/omarchy-plymouth-set bin/omarchy-refresh-plymouth test/shell.d/plymouth-set-test.sh`
- `bash test/shell.d/plymouth-set-test.sh`
- `./test/cli` (via `./test/all`)
- `git diff --check`

The focused suite deterministically covers pre-open symlink replacement, malicious content planted in a caller-owned staging tree, rejection of user-writable packaged sources, exact root authorization of a dev-linked checkout, rejection of absent/stale/writable dev authorization, Plymouth and SDDM destination symlinks, unsafe destination ancestors, allowlist drift, restrictive umasks, and the refresh path.

The supplied exploit artifacts were also run in a Bubblewrap sandbox with disposable `/usr/share/plymouth` and `/usr/share/sddm` trees and privilege shims:

- The unchanged `afflicted.sh` reached `[*] racing` and then reported `[-] lost the race`.
- `race.c`, compiled with `cc -O2`, remained blocked waiting for a vulnerable `/tmp/tmp.*` stage while `omarchy-plymouth-set` completed successfully.
- After both attempts, the Plymouth and SDDM logos were regular mode-0644 files containing the selected logo bytes, and neither fake system tree contained `/etc/shadow` content.

`./test/all` reached and passed the Plymouth suite, but the aggregate run had seven unrelated environment failures: missing `omarchy-pkgs` coverage checkout, unavailable GUI/runtime resources, blocked network access, and read-only home paths.

Reported-by: _SiCk // afflicted.sh

### DIFF
--- a/bin/omarchy-dev-link
+++ b/bin/omarchy-dev-link
@@ -34,8 +34,12 @@ systemd, shell, or app-launcher environment; reboot to make every layer agree.
 Affects only \$OMARCHY_PATH-resolved trees: bin/, default/, shell/,
 themes/, applications/, config/. Files installed at fixed system paths
 (/etc/, /usr/lib/systemd/, udev rule bodies, /etc/skel after user
-creation, /usr/share/plymouth) are NOT covered — for those, use
-omarchy-dev-pkg-test to build and install the package from the checkout.
+creation) are NOT covered — for those, use omarchy-dev-pkg-test to build
+and install the package from the checkout.
+
+The Plymouth and SDDM themes under /usr/share are the exception: omarchy
+plymouth set and omarchy refresh plymouth republish them from the checkout,
+reading this link's authorization out of the root-owned /etc/omarchy.conf.
 
 Also writes $sudoers_file so sudo resolves omarchy-*
 from the checkout instead of the packaged copies. That part takes effect
--- a/bin/omarchy-plymouth-reset
+++ b/bin/omarchy-plymouth-reset
@@ -3,5 +3,7 @@
 # omarchy:summary=Restore the default Omarchy Plymouth boot theme and SDDM login screen
 # omarchy:requires-sudo=true
 
-omarchy-refresh-plymouth
-omarchy-refresh-sddm
+set -euo pipefail
+
+"$OMARCHY_PATH/bin/omarchy-refresh-plymouth"
+"$OMARCHY_PATH/bin/omarchy-refresh-sddm"
--- a/bin/omarchy-plymouth-set
+++ b/bin/omarchy-plymouth-set
@@ -11,21 +11,39 @@ set -euo pipefail
 # fixed destination atomically. The caller opens the selected logo before sudo,
 # so the privileged process never resolves a user-controlled input path.
 
-refresh_default=false
-if (( $# == 1 )) && [[ $1 == "--refresh-default" ]]; then
-  refresh_default=true
-elif (( $# != 3 )); then
+usage() {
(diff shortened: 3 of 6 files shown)


ErikMelton:
Resolved the dev-link decision in 363db1f5.

A normal packaged source remains subject to the complete canonical/root-owned/non-writable file and ancestor checks. A user-owned checkout is accepted only when a canonical, root-owned, non-group/world-writable `/etc/omarchy.conf` has exactly one line authorizing that same canonical checkout. That file is the root-side trust decision made by `omarchy dev link`; dev mode already runs privileged Omarchy commands from the user-editable checkout, so allowing its packaged assets does not widen that mode's trust boundary. Missing, stale, and writable authorization files are covered and fail closed.

Current verification is green:

- `bash -n bin/omarchy-plymouth-set bin/omarchy-refresh-plymouth test/shell.d/plymouth-set-test.sh`
- `test/shell.d/plymouth-set-test.sh`
- `./test/cli`
- `git diff --check`

The supplied unchanged `afflicted.sh` and compiled `race.c` were already exercised against the secured publisher in the disposable Bubblewrap fake-system harness, with the results recorded in the PR body. The local artifact directory has since been removed, so I could not repeat those exact external files after this dev-authorization-only commit.

The real-sudo stdin check still needs a terminal with authentication. I attempted a non-mutating equivalent using the repository's 3,072-byte default logo and a fixed `/tmp` destination, but sudo waited for interactive authentication unavailable to this session; I terminated it and confirmed no probe file existed. I have not claimed that check passed.

PR #8935, opened 2026-08-29
Restart fprintd after resume so it re-enumerates the reader

MaxMad75:
## Problem

Fingerprint authentication dies on this ThinkPad after enough suspend cycles, in a way that looks nothing like a failure. Everything you would check is healthy:

```console
$ systemctl show fprintd -p MainPID -p ActiveEnterTimestamp --value
115096
Fri 2026-08-28 17:19:26 CEST          # ~16 hours old, never exited

$ systemd-inhibit --list | grep -i fprint
net.reactivated.Fprint  0  root  115096  fprintd  sleep  Suspend fingerprint readers  delay

$ journalctl -u fprintd --since "16 hours ago" | wc -l
0
```

And yet:

```console
$ fprintd-list "$USER"
No devices available
```

The reader was on the bus the entire time — at a new address:

```console
$ lsusb | grep 06cb
Bus 001 Device 010: ID 06cb:00bd Synaptics, Inc. Prometheus MIS Touch Fingerprint Reader
                # 008 the day before
```

## Cause

libfprint enumerates USB once, at startup, and ships no hotplug support:

```console
$ strings /usr/lib/libfprint-2.so* | grep -ci hotplug
0
```

A daemon that outlives a re-enumeration therefore keeps a handle nothing answers to. `systemctl restart fprintd` clears it instantly and is the only cure I found.

The daemon reaches this state on its own. @xe-nvdk counted fprintd's idle exits on a second reader (`06cb:00fc`): 19 in the cycles before the wedging suspend, 0 in every cycle after it. A wedged instance stops exiting, so it does not need `--no-timeout` or any other keep-resident mitigation to become long-lived — those only make the failure easier to notice. (An earlier version of this section claimed the opposite; his counts corrected it.)

## Fix

A unit ordered after the sleep targets, enabled by `omarchy setup security fingerprint` alongside the PAM configuration it already writes.

It also caps `fprintd.service`'s stop timeout at 3s. A wedged fprintd does not answer SIGTERM, so at Omarchy's inherited default the restart would cost more on the wake path than the bug it repairs; the cap lets systemd reach for SIGKILL promptly. This part is orthogonal to the mechanism and worth keeping whichever of the neighbouring PRs lands — none of the others sets it.

`try-restart` rather than `restart`: when fprintd is not running there is nothing stale to clear, and *starting* it here would place its activation inside a sleep transition, which is precisely where logind refuses it its own delay inhibitor.

A unit rather than a script in `/usr/lib/systemd/system-sleep/`: systemd-sleep(8) calls those hooks "hacks" and points at inhibitor locks instead, and a hook runs *inside* the transition, which is the wrong moment to ask systemd to restart a service. I tried the hook first and it does not work — on systemd 261 `/etc/systemd/system-sleep/` is not read at all, and the man page documents only the `/usr/lib` path.

## Verified

Omarchy 4.0.1, systemd 261, kernel 7.1.9-arch1-2, ThinkPad X390 Yoga, Synaptics `06cb:00bd`, `mem_sleep=deep`.

The same shape is now confirmed on two further readers, both s2idle: `06cb:00fc` (X1 Carbon Gen 9, @xe-nvdk, with the kernel trace below) and `06cb:0123` (X1 Carbon Gen 13, @k7cfo in #8818). Three Synaptics variants and both sleep modes, which points at libfprint rather than at one firmware.

@xe-nvdk's kernel-side capture of the re-enumeration this PR is built on — the part I could only infer here:

```
usb 3-3: USB disconnect, device number 2
usb 3-3: New USB device found, idVendor=06cb, idProduct=00fc
fprintd: Ignoring device due to initialization error: unsupported firmware version
fprintd: Async command sending failed: USB error … No such device [-4]
```

Stop-timeout cap, on this machine:

```console
$ systemctl show fprintd -p TimeoutStopUSec --value
3s
```

```console
$ # across one systemd suspend/resume with the unit enabled
PID before: 252757
PID after:  253978

$ journalctl -u fprintd-resume
Starting Restart fprintd after resume so it re-enumerates the reader...
fprintd-resume.service: Deactivated successfully.
Finished Restart fprintd after resume so it re-enumerates the reader.

$ fprintd-list "$USER"
Fingerprints for user maxmad on Synaptics Sensors (press):
 - #0: right-index-finger

$ systemd-inhibit --list | grep -i fprint
net.reactivated.Fprint  0  root  253978  fprintd  sleep  ...    # the new instance
```

## What this does not claim

The original silent failure takes hours and several sleep cycles to appear here, and I could not provoke it on demand. Four attempts failed: unbinding and rebinding the USB driver, toggling `authorized` on the device, and short RTC-timed suspends — none of them re-addressed the reader the way an overnight run of suspends did.

So this is preventive rather than demonstrated against a reproduction: it makes the observed stale state unable to survive a wake. If someone has a reliable reproducer I would like to run against it.

## Overlap with #7158 and #8818

Both opened alongside this one and land on the same unit at resume:

- **#7158** installs a post-resume sleep hook running `timeout … try-restart fprintd.service` — functionally this PR, as a hook rather than a unit — plus retry backoff and an "unavailable" notice in the lock screen.
- **#8818** installs a post-resume hook that *stops* fprintd PID-selectively and leaves startup to the next D-Bus activation.

@spencerbull's review of #8818 makes the collision explicit: systemd runs `/usr/lib/systemd/system-sleep/` in parallel (he measured two hooks 53 µs apart), so a stop and a restart would race with no ordering to rely on. This PR's unit is ordered `After=suspend.target` rather than being a hook, so it is outside that parallel set, but it still acts on the same unit at the same moment.

I would rather see #7158 land than this: its first layer covers the same failure, and it also fixes the retry storm afterwards, which nothing here addresses. If a unit is preferred over a hook, this is the unit; otherwise the `TimeoutStopSec` cap is the piece worth salvaging.

Related: #7229 (the inhibitor race), #8531 (the lock-side fix for it), #7158, #8818.

### DIFF
--- a/bin/omarchy-setup-security-fingerprint
+++ b/bin/omarchy-setup-security-fingerprint
@@ -53,6 +53,39 @@ EOF
   fi
 }
 
+# libfprint enumerates USB once, at startup, and ships no hotplug support:
+#
+#   $ strings /usr/lib/libfprint-2.so* | grep -ci hotplug
+#   0
+#
+# So a long-lived fprintd never re-probes the bus. When a reader is
+# re-enumerated across suspend the daemon keeps a handle nothing answers to and
+# reports "No devices available" — with a healthy sleep inhibitor and nothing
+# at all in `journalctl -u fprintd`, which is what makes it hard to recognise.
+# Restarting it clears the stale state; doing so on every resume means the
+# state cannot outlive a single wake.
+#
+# try-restart, not restart: when fprintd is not running there is nothing stale
+# to clear, and starting it here would put its activation inside a sleep
+# transition, where logind refuses it its own delay inhibitor (see #7229).
+setup_fprintd_resume_unit() {
+  echo "Configuring fprintd to re-probe the reader after resume..."
+  sudo tee /etc/systemd/system/fprintd-resume.service >/dev/null <<'EOF'
+[Unit]
+Description=Restart fprintd after resume so it re-enumerates the reader
+After=suspend.target hibernate.target hybrid-sleep.target suspend-then-hibernate.target
+
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/systemctl try-restart fprintd.service
+
+[Install]
+WantedBy=suspend.target hibernate.target hybrid-sleep.target suspend-then-hibernate.target
+EOF
+  sudo systemctl daemon-reload
+  sudo systemctl enable fprintd-resume.service >/dev/null
+}
+
 setup_lock_fingerprint_pam() {
   echo "Configuring lock screen for fingerprint authentication..."
   sudo tee /etc/pam.d/omarchy-lock-fingerprint >/dev/null <<'EOF'
@@ -100,6 +133,7 @@ if sudo fprintd-enroll "$USER"; then
     # pam_fprintd with nothing to match.
     setup_pam_config
     setup_lock_fingerprint_pam
+    setup_fprintd_resume_unit
     echo -e "\e[32m\nPerfect! Fingerprint authentication is now configured.\e[0m"
     echo "You can use your fingerprint for sudo, polkit, and lock screen (Super + Ctrl + L)."
   else
--- a/bin/omarchy-setup-security-fingerprint
+++ b/bin/omarchy-setup-security-fingerprint
@@ -82,6 +82,15 @@ ExecStart=/usr/bin/systemctl try-restart fprintd.service
 [Install]
 WantedBy=suspend.target hibernate.target hybrid-sleep.target suspend-then-hibernate.target
 EOF
+  # A wedged fprintd does not answer SIGTERM: clearing one takes the full stop
+  # timeout and then SIGKILL. At the default 90s that lands squarely on the
+  # wake path, so the restart above would make resume slower than the bug.
+  sudo mkdir -p /etc/systemd/system/fprintd.service.d
+  sudo tee /etc/systemd/system/fprintd.service.d/stop-timeout.conf >/dev/null <<'EOF'
+[Service]
+TimeoutStopSec=3s
+EOF
+
   sudo systemctl daemon-reload
   sudo systemctl enable fprintd-resume.service >/dev/null
 }


xe-nvdk:
Confirming this on a second Synaptics Prometheus, and I have the kernel-side evidence for the re-enumeration you infer from the changed bus address.

- Omarchy 4.0.1-1, systemd 261 (261.2-1-arch)
- kernel 7.1.9-arch1-2, `mem_sleep = [s2idle]`
- ThinkPad X1 Carbon Gen 9 (20XW003GUS), i7-1165G7
- Synaptics Prometheus `06cb:00fc` (yours is `06cb:00bd`)
- fprintd 1.94.5-2 / libfprint 1.94.100-1

### The kernel logs the re-enumeration as it happens

At the resume instant, the reader disconnects and comes straight back as a new device, and the surviving daemon keeps talking to the old handle:

```
07:34:11 kernel:  usb 3-3: USB disconnect, device number 2
07:34:11 kernel:  usb 3-3: New USB device found, idVendor=06cb, idProduct=00fc
07:34:12 fprintd[639160]: Ignoring device due to initialization error: unsupported firmware version
07:34:33 fprintd[639160]: Async command sending failed: USB error on device 06cb:00fc :
                          No such device (it may have been disconnected) [-4]
```

That is the same "enumerates once at startup, no hotplug" story, caught in the act rather than reconstructed after the fact.

### Why the daemon never clears itself

fprintd is D-Bus activated and normally idle-exits after ~30s, which would resolve this on its own. Counting `fprintd.service: Deactivated successfully` in a single boot, split at the suspend that wedged it:

```
before 23:28  ->  19 idle-exits
after  23:28  ->   0
```

```
$ ps -o pid,lstart,etime -p 639160
639160  Fri Aug 28 23:28:01 2026  09:06:18
```

Nine hours, started immediately before the overnight suspend. 2,391 refused claims across that boot; a single 20-second lock episode accounted for 74:

```
fprintd[639160]: Authorization denied to :1.4798 to call method 'Claim' for
                 device 'Synaptics Sensors': Device was already claimed
```

### The wedged daemon ignores SIGTERM

Worth noting for the stop-bounding discussion on #8818: clearing it is not a polite stop. `systemctl restart fprintd` sat through the full stop timeout and needed SIGKILL:

```
fprintd.service: State 'stop-sigterm' timed out. Killing.
fprintd.service: Killing process 639160 (fprintd) with signal SIGKILL.
fprintd.service: Consumed 6.260s CPU time over 36min 30.684s wall clock time, 6M memory peak.
```

So whatever lands is worth pairing with a short `TimeoutStopSec` (I run 3s) — otherwise a wedged instance adds the default 90s stop timeout to the wake path.

### Independent confirmation on the `/etc/systemd/system-sleep/` point

I hit your dead end before reading this PR, and can confirm it on systemd 261. The path is referenced nowhere in the binary:

```console
$ systemctl --version | head -1
systemd 261 (261.2-1-arch)

$ strings /usr/lib/systemd/systemd-sleep | grep -i system-sleep
/usr/lib/systemd/system-sleep

$ man systemd-sleep | grep -o '/[a-z/]*system-sleep/' | sort -u
/usr/lib/systemd/system-sleep/
```

A hook dropped in `/etc/systemd/system-sleep/` is silently never executed. Easy to fool yourself here, since running the script by hand proves nothing about whether systemd calls it.

### Your unit approach verified here

Same unit shape, `try-restart`, `After=`/`WantedBy=` the four sleep targets. Through a logind suspend with an RTC wake:

```
08:48:43  systemd[1]: Reached target Sleep.
08:48:53  systemd[1]: Starting Restart fprintd after resume so it re-enumerates the reader...
08:48:54  systemd[1]: Finished Restart fprintd after resume so it re-enumerates the reader.
```

Afterwards: device enumerates, `Verify started!` on first try, zero `already claimed`, zero `unsupported firmware version`, and idle-exits resumed normally.

One testing note that cost me a cycle: `rtcwake -m mem` writes directly to `/sys/power/state` and never reaches `Reached target Sleep`, so the unit correctly does not fire and it looks like a failure. `rtcwake -m no -s N` to arm the alarm followed by `systemctl suspend` exercises the real path.

MaxMad75:
@xe-nvdk — this is the kernel-side evidence I could not produce, and it changes what this PR can claim. Thank you.

**Your trace upgrades my mechanism from inferred to observed.** I argued the reader re-enumerates and libfprint never notices because it has no hotplug path (`strings /usr/lib/libfprint-2.so.2 | grep -ci hotplug` → 0, enumeration happens once at startup). I could not catch it happening. Your `usb 3-3: USB disconnect` → `New USB device found` → `unsupported firmware version` → `USB error … No such device [-4]` is that sequence, in order, on the record.

**You corrected the PR, and I have amended it.** I wrote that `--no-timeout` "makes the daemon long-lived by design, which is exactly the condition under which this failure appears". Your idle-exit counts — 19 before the wedging suspend, 0 after — show the wedged daemon stops exiting on its own. The failure does not need `--no-timeout`; that mitigation only makes it easier to notice. The body no longer says otherwise.

**Adopted your `TimeoutStopSec` finding** in `e455465`. You are right that this was a hole: a wedged fprintd ignores SIGTERM, and at the inherited default the restart would have cost more than the bug it repairs. The branch now also installs

```ini
# /etc/systemd/system/fprintd.service.d/stop-timeout.conf
[Service]
TimeoutStopSec=3s
```

Verified on my machine: `systemctl show fprintd -p TimeoutStopUSec --value` → `3s`, both drop-ins active.

**Third data point, and a different sleep mode.** ThinkPad X390 Yoga (`20NQS4RE00`), Synaptics `06cb:00bd`, fprintd 1.94.5-2, libfprint 1.94.100-1, Linux 7.1.9-arch1-2 — but `/sys/power/mem_sleep` here is `s2idle [deep]`, i.e. S3, where yours and #8818's reporters are both s2idle. Three Synaptics variants (`00bd`, `00fc`, `0123`) and both sleep modes now show the same shape, which makes this look like a libfprint property rather than a firmware quirk.

**And a scope problem I have to raise against my own PR.** Since I opened this, two neighbours have surfaced:

- **#8818** installs a post-resume sleep hook that *stops* fprintd (PID-selective) and leaves startup to the next D-Bus activation.
- **#7158** installs a post-resume sleep hook that runs `timeout … try-restart fprintd.service` — functionally what this PR does, only as a sleep hook rather than a unit — plus retry backoff and an "unavailable" notice in the lock screen.

systemd runs everything in `/usr/lib/systemd/system-sleep/` in parallel; @spencerbull measured two hooks starting 53 µs apart in #8818. My unit is ordered `After=suspend.target` rather than being a sleep hook, so it is not in that parallel set — but it still lands on the same unit at the same moment, and a stop from #8818 racing a restart from here would be exactly the failure mode he is warning about.

Concretely: **#7158's first layer already does what this PR does**, and it does it alongside the retry fix that covers the part neither of us addresses — the storm after the wedge (~4 PAM sessions/second against a reader that cannot answer; on my machine 178 sessions and 88 `already claimed` in 26 seconds, ending in a black screen). I would rather see that land than mine. If maintainers prefer a unit over a sleep hook, the `TimeoutStopSec` drop-in here is worth keeping either way — it is orthogonal to which of the three mechanisms wins, and none of the others sets it.

I am testing #7158 on this hardware now and will report there.

MaxMad75:
Closing in favour of #7158.

Having tested that branch on this hardware, it covers what this PR does and more: its post-resume hook clears the wedged claim (9 ms after resume here, fresh daemon 455 ms later, ordered before the lock screen's own resume work), and it also paces the retry storm afterwards — which is the part that actually put a black screen in front of me, and which this PR never addressed.

Two things from here that should not get lost, both already posted on #7158:

1. **A wedged fprintd does not answer SIGTERM.** Two of the ten fprintd stops in this machine's journal hit `State 'stop-sigterm' timed out. Killing.` and needed SIGKILL, at 5.08 s and 5.02 s against Omarchy's `DefaultTimeoutStopSec=5s`. Since a resume hook runs synchronously while `user.slice` is frozen, that cost lands on the thaw precisely in the case the hook exists for. A `TimeoutStopSec=3s` drop-in fixes it; none of the open PRs sets one. I have offered it to @GeertJohan against his branch so it can ride along there rather than becoming a fifth PR in this area.
2. **A third reader and the first deep-sleep data point.** ThinkPad X390 Yoga, Synaptics `06cb:00bd`, `mem_sleep=deep` (S3), alongside `06cb:00fc` (@xe-nvdk, here) and `06cb:0123` (@k7cfo, #8818) on s2idle. Three Synaptics variants and both sleep modes show the same shape, which points at libfprint's missing hotplug path rather than at one firmware.

Thanks @xe-nvdk for the kernel-level trace and the `TimeoutStopSec` catch — both of those findings outlived this PR.

PR #8945, opened 2026-08-29
Use double-dash Obsidian flag so its CLI works

cabrinha:
`config/obsidian/user-flags.conf` ships `-disable-gpu` with a single dash.

The Arch `obsidian` wrapper passes these flags before the subcommand, and Obsidian's CLI only filters out double-dash flags. The single-dash flag survives and is parsed as the command name, so every CLI call fails:

```
$ obsidian read file="Note"
Error: Command "-disable-gpu" not found.
```

Chromium accepts both `-flag` and `--flag`, so the GPU workaround is unchanged and the CLI starts working. Tested on Obsidian 1.13.7.

### DIFF
--- a/config/obsidian/user-flags.conf
+++ b/config/obsidian/user-flags.conf
@@ -1,3 +1,3 @@
 # Obsidian reads this file through the Arch package wrapper.
--disable-gpu
+--disable-gpu
 --enable-wayland-ime


686f6c61:
Same one-line default as #8700. #9328 keeps that change and adds a migration for existing `~/.config/obsidian/user-flags.conf` copies: https://github.com/omacom/omarchy/pull/9328

Happy to drop the default hunk there if this or #8700 lands first.

bjarneo:
Automated duplication check: this pull request looks similar to #8700, which covers the same Obsidian disable-gpu flag. I keep that one open and close this one to consolidate review.

PR #8951, opened 2026-08-29
Temporarily remove automatic printer discovery

omarchybot:
cups-browsed comes out of the default package set, and a migration removes it from machines that already have it. CUPS stays installed and running; printers are added manually in Print Settings while automatic discovery is reworked.

## Why

cups-browsed turns network advertisements into local print queues. The hardening from #8627 remains in the repository so discovery has a safe configuration if it returns, but the daemon and its implicitclass backend no longer need to run on every stock Omarchy machine.

## Migration

The migration:

- exits without sudo on machines that do not have cups-browsed
- preflights package removal before changing service or printer state
- disables and stops cups-browsed while systemd can still remove its enable symlink cleanly
- reads CUPS destinations under `LC_ALL=C`, treating CUPS' `No destinations added` status as a valid empty printer list
- closes discovery-created implicitclass queues, removes idle ones, and leaves manually-added IPP and USB printers untouched
- removes cups-browsed through `omarchy-pkg-drop`, including its package-owned configuration instead of retaining a `.pacsave`
- records machine-wide completion so another user does not remove a deliberate later reinstall

Normal execution prints only the standard migration description. A genuine CUPS, queue, or package failure exits nonzero, so `omarchy-migrate` leaves the migration pending instead of recording a partial result as complete.

## Fresh installs

cups-browsed is absent from the base package set and is never enabled. The post-install script applies only the CUPS-owned `cups-files.conf` override; it does not write configuration for the uninstalled discovery package.

The hardened cups-browsed configuration, sysusers declaration, and service drop-in stay in the repository for a future return of discovery. The CUPS authorization hardening remains active whether the machine has printers or not.

## Verification

A dedicated CUPS acceptance test inspects the installed machine after either a fresh install or migration. It requires the printing packages and CUPS scheduler to remain healthy, cups-browsed and its package files to be absent, and unauthenticated printer administration to remain denied. It does not mock or assert migration actions.


spuder:
As a user who runs a print shop, I was delighted to see all my printers automatically added to omarchy. 
I see the benefit to not have a daemon constantly running. 
I would like to encourage that omarchy preserve a low friction solution to adding printers that doesn't require manually entering IP addresses.

dhh:
> As a user who runs a print shop, I was delighted to see all my printers automatically added to omarchy. I see the benefit to not have a daemon constantly running. I would like to encourage that omarchy preserve a low friction solution to adding printers that doesn't require manually entering IP addresses.

This is temporary and we will bring it back.

PR #8959, opened 2026-08-29
shell: add configurable mouse wheel scroll speed to menu lists

ShiMMyTo:
## Summary

The mouse wheel scroll inside the Omarchy apps menu and emoji picker is very slow: each wheel tick moves only a tiny fixed amount, because Qt Quick `ListView`/`GridView` (both `Flickable`) have no external way to configure the wheel step. This is most noticeable on the long apps list and the emoji grid.

This PR adds a reusable `WheelScrollArea` component to the `qs.Ui` module: a transparent `MouseArea` placed on top of a `Flickable` (after it in the stacking order, so it captures the wheel before the list does) that scrolls the target manually at a configurable speed. It is wired into:

- `shell/plugins/menu/Menu.qml` — apps menu list
- `shell/plugins/emojis/Emojis.qml` — emoji picker grid

## Key details

- `acceptedButtons: Qt.NoButton` lets clicks reach the rows below.
- `speed` (default 8) is the divisor of the visible height used as the step; larger = slower per wheel click.
- Any future list can opt in with `WheelScrollArea { flickable: myList }`.

## Testing

- `./test/shell` — `manifest entrypoints load` passes; the menu and emoji plugin entry points instantiate with the new component.
- The 5 remaining shell test failures are pre-existing on `quattro` (verified by stashing the change and re-running).

### DIFF
--- a/shell/Ui/WheelScrollArea.qml
+++ b/shell/Ui/WheelScrollArea.qml
@@ -1,0 +1,34 @@
+import QtQuick
+
+// Configurable mouse wheel scroll speed for Qt Quick lists.
+//
+// A Qt Quick ListView/GridView (both Flickable) scroll by a small fixed amount
+// per wheel tick, with no external way to configure it, which feels slow on
+// long lists. This transparent MouseArea sits ON TOP of the list — after it in
+// the stacking order — so it captures the wheel before the Flickable does, and
+// scrolls the target manually at a configurable speed. Place it after the list,
+// inside the same parent Item, e.g.:
+//
+//   WheelScrollArea { flickable: myList }
+//
+// speed is the divisor of the visible height used as the step: the larger it
+// is, the slower each wheel click scrolls.
+MouseArea {
+  id: root
+
+  required property Flickable flickable
+  property real speed: 8
+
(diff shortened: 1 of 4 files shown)


bjarneo:
Automated duplication check: this pull request looks similar to #7451, which covers the same menu wheel scroll speed. I keep that one open and close this one to consolidate review.

ShiMMyTo:
Reabro el PR porque cubre más que la solución #7451: además del menú, incluye el selector de emojis y es un componente reutilizable (WheelScrollArea) con velocidad configurable por lista, en lugar de un ajuste fijo solo para el menú.

ShiMMyTo:
This PR is not the same as #7451. #7451 targets the menu list only and scales the wheel delta by a fixed 3x, using a WheelHandler inside the ListView.

This PR takes a different, more general approach: a reusable WheelScrollArea component — a transparent MouseArea stacked above the list that captures the wheel before the list handles it and moves it at a configurable speed. It covers the apps menu and every other Omarchy submenu (keybindings, fonts, etc.) plus the emoji picker, which #7451 does not. The speed can be tuned per list instead of being hardcoded.

Issue #7220, opened 2026-08-17
Screensaver/idle triggers during Firefox video playback — portal Inhibit falls back to unowned org.freedesktop.ScreenSaver


sumitrampal:
## What happened

Screensaver / idle lock fires on schedule (`idle.screensaver` in
`~/.config/omarchy/shell.json`) even while a video is actively playing in
Firefox (e.g. YouTube). The video keeps playing behind the screensaver —
Firefox never successfully inhibits idle.

## What I expected

Playing video in the browser should suppress the screensaver/lock the same
way it does on GNOME/KDE, for the duration of playback.

## Root cause (traced via journal + source)

Omarchy's idle system is a Quickshell service
(`/usr/share/omarchy/shell/plugins/services/idle/Service.qml`), not
`hypridle` — no `hypridle` process runs in an Omarchy session even though
`~/.config/hypr/hypridle.conf` still ships/exists. The Quickshell service
uses `Quickshell.Wayland.IdleMonitor` with `respectInhibitors: true`, which
only honors **Wayland-protocol** idle inhibitors
(`idle-inhibit-unstable-v1`), the kind fullscreen `mpv`/VLC register
directly against a surface.

Firefox instead requests idle inhibition via the **freedesktop portal**
(`org.freedesktop.impl.portal.Inhibit`) while a `<video>` is playing. Tracing
that call:

- `/usr/share/xdg-desktop-portal/portals/hyprland.portal` (the backend
  selected `UseIn=... Hyprland ...`) does **not** list
  `org.freedesktop.impl.portal.Inhibit` in its `Interfaces=` — only
  `Screenshot;ScreenCast;GlobalShortcuts;InputCapture`.
- `xdg-desktop-portal` therefore falls back to
  `xdg-desktop-portal-gtk` (`UseIn=gnome`) as the only backend implementing
  `Inhibit`.
- `xdg-desktop-portal-gtk`'s Inhibit implementation proxies to the legacy
  `org.freedesktop.ScreenSaver` D-Bus interface — but nothing in an Omarchy
  session owns that well-known name (no gnome-screensaver-alike service), so
  every call fails.

Journal, repeating every few seconds while the video plays:

```
xdg-desktop-portal-gtk[1929]: Backend call failed: Cannot invoke method; proxy is for the
well-known name org.freedesktop.ScreenSaver without an owner, and proxy was constructed
with the G_DBUS_PROXY_FLAGS_DO_NOT_AUTO_START flag
[...]

PR #8965, opened 2026-08-29
Count gamepad input and audio playback as idle activity

lde-alen:
The screensaver comes up and the lock screen takes over mid-game, while a controller is actively in use and a video is playing.

## Why it happens

The idle service asks the compositor when the user went idle, so it sees only what libinput reports. Two everyday cases fall outside that.

**Gamepads are not seat input.** libinput classifies joystick devices as unhandled, so they never reset `ext-idle-notify`. The split is visible on a connected Xbox pad — the pad presents four nodes, and Hyprland lists three of them:

```
$ hyprctl devices | grep -i xbox
    xbox-wireless-controller-mouse
    xbox-wireless-controller-keyboard
    xbox-wireless-controller-consumer-control

$ grep -A4 'Xbox Wireless Controller"' /proc/bus/input/devices | grep Handlers
H: Handlers=event25 js0
```

`event25` is the sticks and buttons. It is the one node Hyprland does not see, so an entire session on a controller counts as no input at all: screensaver at `idle.screensaver`, lock at `idle.lock`, mid-game.

**Media playback depends on the player asking.** Staying awake through a film relies on a Wayland idle inhibitor. Players that take one are fine. XWayland clients cannot take one at all, and the `org.freedesktop.ScreenSaver` D-Bus API that players reach for instead has no owner on an Omarchy session:

```
$ busctl --user call org.freedesktop.ScreenSaver /org/freedesktop/ScreenSaver \
    org.freedesktop.ScreenSaver Inhibit ss "test" "probe"
Call failed: The name is not activatable
```

So those requests fail silently and nothing cross-checks whether sound is actually coming out.

## The change

`omarchy-idle-activity` watches both signals directly and prints `ACTIVE`/`IDLE`:

- **Gamepads** — reads the evdev nodes that share a device with a `js*` handler, which is what marks the node libinput drops. Buttons and hats count immediately; analogue axes need to clear a deadzone, since sticks rest off-centre and wander there. The first sample on an axis only establishes where it rests.
- **Audio** — an uncorked sink input, the reliable evidence that sound is being produced rather than merely that a player is open.

The service suppresses the countdown while either reports activity, and starts it as soon as they clear, so walking away from a paused machine still locks on the normal schedule. A supervising timer restarts a crashed probe, and a probe that will not run degrades to today's compositor-only behaviour rather than pinning the screen awake.

Python for the probe because both signals need work bash cannot do: evdev delivers fixed-width binary structs across several descriptors at once, which needs struct unpacking behind `select()`.

## Configuration

Both default to on. `idle.inhibitWhenGamepadActive` and `idle.inhibitWhenAudioPlaying` turn them off; `idle.gamepadGrace` (default 120s) sets how long a pad keeps the machine awake after the last input you actually made, so setting the pad down mid-game doesn't lock you out a second later.

## Verification

Reproduced and fixed on Omarchy 4.0.1, Hyprland 0.56.2, with an Xbox Wireless Controller over Bluetooth.

With the change running in the live shell, the probe holds the node Hyprland ignores, and the countdown is suppressed while the game plays audio:

```
$ ls -l /proc/$(pgrep -f omarchy-idle-activity)/fd | grep input
3 -> /dev/input/event25

$ omarchy-shell idle status | jq '{idle, externalActivity, inIdleCycle, lastEvent}'
{
  "idle": false,
  "externalActivity": true,
  "inIdleCycle": false,
  "lastEvent": "external-activity: gamepad or audio in use"
}
```

Setting `inhibitWhenAudioPlaying: false` with no controller input flips `externalActivity` back to `false` and normal idling resumes, confirming the suppression clears rather than latching.

`./test/all` passes at 202/207, the same five files failing before and after this branch (four need an `omarchy-pkgs` checkout that isn't present; `runtime-smoke-test` and `theme-install-guards-test` fail on the clean base here too). New coverage in `test/shell.d/idle-test.sh` exercises the settings/command model and replays synthetic `input_event` structs through the probe for button, drift, deliberate push, and d-pad cases.

### DIFF
--- a/bin/omarchy-idle-activity
+++ b/bin/omarchy-idle-activity
@@ -1,0 +1,235 @@
+#!/usr/bin/python3
+
+# omarchy:summary=Report activity the compositor cannot see, for the idle service
+# omarchy:args=[--no-gamepad] [--no-audio] [--gamepad-grace <seconds>] [--once]
+# omarchy:hidden=true
+
+"""Report the activity Hyprland's idle timer cannot observe for itself.
+
+The idle service asks the compositor when the user went idle, which covers
+keyboards, mice, and touchpads. Two things it cannot cover:
+
+  * Gamepads. libinput classifies joystick devices as unhandled and never
+    reports them as seats' pointer or keyboard input, so ext-idle-notify keeps
+    counting up through an entire play session on a controller. Reading the
+    evdev nodes directly is the only way to see that input.
+
+  * Media playback. Staying awake through a film relies on the player holding
+    a Wayland idle inhibitor. Players that do are fine; XWayland clients cannot
+    hold one at all, and the org.freedesktop.ScreenSaver D-Bus API that many
+    reach for instead has no owner on an Omarchy session, so those requests
+    fail silently. An uncorked sink input is the reliable evidence that sound
+    is actually being produced.
+
+Python rather than bash because both signals need work bash cannot do: evdev
+delivers fixed-width binary structs on several descriptors at once, which needs
+struct unpacking behind select().
+
+Prints ACTIVE or IDLE on every state change, plus a periodic heartbeat so a
+reader that missed a line resynchronises without restarting the probe.
+"""
+
+import argparse
+import os
+import select
+import struct
+import subprocess
+import sys
+import time
+
+# struct input_event: struct timeval, __u16 type, __u16 code, __s32 value
+EVENT = struct.Struct("llHHi")
+EV_KEY = 0x01
+EV_ABS = 0x03
+
+# Analogue sticks rest a few hundred units off centre and wander there, so only
+# a deliberate push counts. Hats report -1/0/1 and need every change.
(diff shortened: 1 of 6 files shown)


Henaneitor:
The gamepad half is the right diagnosis: libinput never sees the joystick node, so a whole session on a controller looks idle. Watching `ID_INPUT_JOYSTICK` (uaccess, not the `input` group) and treating that as presence is fair.

The audio half I would split out. Sniffing an uncorked sink treats Spotify the same as a film. The actual hole is that nothing owns `org.freedesktop.ScreenSaver` since hypridle went away (see #6475, #7220, #10407). Players already ask not to idle; Omarchy 3 heard them, 4 drops the request. `IdleMonitor { respectInhibitors: true }` already honours a real Wayland inhibit — the missing piece is owning that bus / the Inhibit portal, not guessing from Pulse.

Happy to see the pad watcher land on its own.

PR #8968, opened 2026-08-29
Tag brave-origin windows as chromium-based browsers

Rastafaustian:
Brave Origin is a chromium-based browser but was not covered by the window rule in `default/hypr/apps/browser.lua`, so its windows missed the chromium-based-browser tag (and the opacity/tile rules that follow it).

Extends the chromium rule's app-id pattern to include `brave-origin` alongside `brave-browser`, and adds `test/shell.d/browser-tags-test.sh` asserting the invariant: every chromium-based and firefox-based browser app-id is tagged to the right family, with no cross-tagging.

### DIFF
--- a/default/hypr/apps/browser.lua
+++ b/default/hypr/apps/browser.lua
@@ -1,5 +1,5 @@
 -- Browser tags and styling.
-o.window("((google-)?[cC]hrom(e|ium)|[bB]rave-browser|[mM]icrosoft-edge|Vivaldi-stable|helium)", { tag = "+chromium-based-browser" })
+o.window("((google-)?[cC]hrom(e|ium)|[bB]rave-(browser|origin)|[mM]icrosoft-edge|Vivaldi-stable|helium)", { tag = "+chromium-based-browser" })
 o.window("([fF]irefox|zen|librewolf)", { tag = "+firefox-based-browser" })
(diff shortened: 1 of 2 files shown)


pkovzz:
Independent confirmation of this on a second machine, in case another data point helps this land.

Brave Origin's window class is `brave-origin` (via `hyprctl clients`), so it misses the `[bB]rave-browser` alternative on `browser.lua:2`, never gets `+chromium-based-browser`, and stays on the generic `default-opacity` rule:

```
tags: default-opacity*
```

The user-visible symptom is that a **focused** Brave Origin window sits at `0.985` rather than `1.0` — the desktop is faintly visible through it — while Firefox and Chrome reach full opacity, since their classes do match their respective patterns. Applying this PR's `[bB]rave-(browser|origin)` locally as a user-level `o.window` override resolves it.

- Omarchy 4.0.2-1, Hyprland 0.56.2 (v0.56.2, efb5099), kernel 7.1.9-arch1-2
- Intel i5-12400F, AMD Radeon RX 6700 XT (Navi 22)

Cross-reference for whoever triages: #9882 covers this same case as a side effect of its web-app fix, via `[bB]rave(-browser|-origin)?` on the same line, so the two conflict textually — its review bot flagged the collision from that side as well.

PR #8975, opened 2026-08-29
Add support for detached bar with margin and radius properties

ekrist1:


Marc-AnthonyG:
I would need that. Else what is the propose path to re-style the bar? Is there other style setting we should expose?

ekrist1:
> I would need that. Else what is the propose path to re-style the bar? Is there other style setting we should expose?

Thanks for picking this up in #9725 — the per-edge margin via parseWidthSpec is the right call, and the windowScreenPoint() and omarchy-bar-text-color --inset fixes are bugs my version would have shipped. I'm happy for #9725 to be the merge candidate and to close this one in its favour.

_(Written by AI, overlooked and edited by me)_

On the question of what else to expose: I'm the maker of the plugin ([nordstart](https://github.com/ekrist1/nordstart)). I want a native way of making a OSX/Gnome dock and margin + radius gets me a rounded floating strip, not a dock. The missing pieces, roughly in order of how much they matter:

**Unlocks the dock**

width along the spanned axis — "auto" (hug content), a pixel value, or a percentage. Today the surface anchors to both spanned edges, so nothing in shell.toml can make the bar shorter than the screen.
align = start | center | end — only meaningful once width isn't 100%. Center is the dock default; start/end gives corner islands.

**Makes a floating surface read as floating**

border-width / border-color, referencing the existing [hyprland] border tokens rather than adding new colour vocabulary. A radiused bar with no border tends to vanish on light wallpapers.
shadow (offset, blur, opacity) — opt-in, default off, since it costs an effect pass.
opacity as a float — transparent is a boolean today. A 0–1 background opacity composes with layerrule = blur, omarchy-bar for the frosted look; the layerrule itself could just be documented.

**Interior spacing that rounding makes visible**

padding — inset between the bar edge and the first/last module. With radius = 16 the first icon currently sits in the curve.
gap — spacing between modules, ideally per section.

**Behaviour**

exclusive = true | false — whether the bar reserves space. Many floating-bar users want an overlay that windows slide under.
Auto-hide reveal delay and animation duration, since the mechanism already exists.

csfh:
Closing in favor of #9725, which iterates on this and fixes the detached-bar fallout this version would ship (per-edge margin via the existing width spec, `windowScreenPoint()` mapping, and transparent text sampling with `--inset`).

Author already preferred #9725 as the merge candidate — agreed.

PR #8997, opened 2026-08-29
Highlight each monitor's own active workspace in the bar

### LINKED ISSUE #8183: Regression: Quickshell workspace indicator shows globally focused workspace on every monitor
### System details

AMD Ryzen 5 7640HS with Radeon 760M (Phoenix1), Omarchy 4.0.0-1, Hyprland, two displays (eDP-1 + DP-1)

## Description

On a multi-monitor setup, the built-in Quickshell `omarchy.workspaces` widget highlights the globally focused Hyprland workspace on every monitor's bar instead of the active workspace of that specific monitor.

The actual Hyprland workspace assignment works correctly: switching the laptop display does not change the content on the external display. Only the indicator on the external display changes, so it no longer represents what that monitor is showing.

This appears to be a regression of #1533 after the migration from Waybar to the Quickshell-based Omarchy Shell. Issue #1533 was closed as completed, but the same behavior is present in the current widget.

## Steps to reproduce

1. Connect an external monitor and keep the Omarchy bar visible on both displays.
2. Put workspace 5 on the external display.
3. Focus the laptop display and switch it to workspace 3.
4. Observe the workspace indicator on the external display.

## Actual behavior

Both bars highlight workspace 3, even though the external display still shows workspace 5.

## Expected behavior

- The laptop display bar should highlight workspace 3.
- The external display bar should continue highlighting workspace 5.
- Each bar should derive its active workspace from the monitor associated with that bar.

## Verified Hyprland state

While reproducing the issue, `hyprctl monitors` reported:

- `eDP-1`: active workspace 3
- `DP-1`: active workspace 5

The visible content matched that state.

## Suspected cause

The packaged `shell/plugins/bar/widgets/Workspaces.qml` determines the focused state using the global workspace:

```qml
readonly property bool focused:
    Hyprland.focusedWorkspace !== null &&
    Hyprland.focusedWorkspace.id === modelData
```

Every bar instance therefore reads the same `Hyprland.focusedWorkspace`, regardless of which screen owns that bar.

## Diagnostics

The installed Omarchy version is `4.0.0-1`. This release does not expose the documented `omarchy debug --no-sudo --print` command; it returns `Unknown Omarchy command`, so no debug output is attached.

### LINKED ISSUE #8364: Workspace indicator shows wrong active workspace when switching between monitors
### System details

Xiaomi Notebook Pro, Intel Core i5 11th Gen, Omarchy 4.0.1

### What's wrong?

When using two monitors, the workspace highlight in the top bar becomes incorrect when switching between monitors/workspaces.

Steps to reproduce:
1. Connect two monitors and use workspaces on both monitors.
2. Have a different workspace active on each monitor.
3. Switch from one monitor to the other.
4. Switch to or bring a different workspace onto the monitor.

Expected behavior:
The workspace currently active on the focused monitor should be highlighted in the top bar.

Actual behavior:
The top bar continues highlighting the workspace that was previously active on that monitor, instead of highlighting the workspace that is currently active or the workspace that was brought onto the monitor.

The actual workspace switching works correctly. The issue is only with the workspace highlight shown in the top bar, which becomes stale/incorrect when switching between monitors.

mikebenner:
Fixes #8183
Fixes #8364

## Problem

`omarchy.workspaces` highlights `Hyprland.focusedWorkspace`, and the bar runs one surface per monitor. With two or more monitors every bar therefore marks the same workspace — the globally focused one — so the bar on the non-focused monitor points at a workspace that isn't on that screen, and the workspace that *is* showing there sits dimmed as if it were empty.

Top bar is the laptop panel on workspace 1, bottom bar is a second monitor on workspace 5. Both mark 1:

before

## Fix

Resolve the widget's own monitor through its bar window's screen (`QsWindow.window.screen` → `Hyprland.monitorFor`, the same idiom `Bar.slotWindow` and the tray use) and highlight that monitor's `activeWorkspace`. Falls back to `Hyprland.focusedWorkspace` until the window has a screen Hyprland knows about, so single-monitor behaviour and startup are unchanged.

Only the `focused` binding changes; occupancy dimming, the 1–5 baseline, glyphs and click-to-focus are untouched.

Same two monitors with this change — each bar marks its own workspace:

after

## Verification

- Reproduced and captured on a real second output without extra hardware: `hyprctl output create headless` gives the shell a second bar surface, `grim -o HEADLESS-1` captures it, `hyprctl output remove HEADLESS-1` tears it down.
- Loaded the changed file into the running shell: no QML warnings.
- `./test/all` — the same 4 pre-existing `test/shell` failures as on `quattro` (`config`, `snapper`, `theme-install-guards`, `unowned-system-paths`, all wanting an `omarchy-pkgs` checkout); nothing new.
- I've run this change as a cloned widget on a multi-monitor setup since 2026-08-27.

Still present in Omarchy 4.0.2-1 (2026-08-31). Local workaround is `omarchy plugin clone omarchy.workspaces` plus this same `focused` binding.

### DIFF
--- a/shell/plugins/bar/widgets/Workspaces.qml
+++ b/shell/plugins/bar/widgets/Workspaces.qml
@@ -1,5 +1,6 @@
 import QtQuick
 import QtQuick.Layouts
+import Quickshell
 import Quickshell.Hyprland
 import qs.Commons
 import qs.Ui
@@ -7,6 +8,18 @@ import qs.Ui
 BarWidget {
   id: root
   moduleName: "omarchy.workspaces"
+
+  // A bar surface exists per monitor, so highlight the workspace active on
+  // this widget's own monitor rather than Hyprland.focusedWorkspace, which
+  // would mark the same workspace on every screen. Falls back to the focused
+  // workspace until the window has a screen Hyprland knows about.
+  readonly property var barWindow: root.QsWindow ? root.QsWindow.window : null
+  readonly property var barMonitor: barWindow && barWindow.screen ? Hyprland.monitorFor(barWindow.screen) : null
+  readonly property int activeWorkspaceId: {
+    if (barMonitor && barMonitor.activeWorkspace) return barMonitor.activeWorkspace.id
+    if (Hyprland.focusedWorkspace) return Hyprland.focusedWorkspace.id
+    return 0
+  }
 
   function workspaceById(id) {
     var values = Hyprland.workspaces.values
@@ -56,7 +69,7 @@ BarWidget {
 
         readonly property var workspace: root.workspaceById(modelData)
         readonly property bool occupied: workspace !== null && workspace.toplevels.values.length > 0
-        readonly property bool focused: Hyprland.focusedWorkspace !== null && Hyprland.focusedWorkspace.id === modelData
+        readonly property bool focused: root.activeWorkspaceId === modelData
 
         bar: root.bar
         text: focused ? "\uDB85\uDCFB" : (modelData === 10 ? "0" : String(modelData))


mikebenner:
Still reproduces on 4.0.2-1. This PR was opened the day before CODEOWNERS landed, so it never picked up the automatic owner review request that newer PRs get.

#10190 is complementary (stale `activeWorkspace` after a workspace is moved between monitors). This PR is the per-monitor highlight half: each bar reads its own monitor instead of `Hyprland.focusedWorkspace`.

mikebenner:
@ryanrhughes this predates CODEOWNERS, so it never got the automatic owner review request. Happy to rebase if needed.

mikebenner:
@dhh same note as above — this predates CODEOWNERS, so it never got the automatic owner review request.

jacobs852:
Confirming this on a second setup — same symptom, and this PR's approach fixes it.

**Setup:** Omarchy 4.0.2-1, Hyprland 0.56.2, quickshell 0.3.1-1. Two 1920x1200 displays stacked vertically: Lenovo laptop panel `eDP-1` at `0x0`, Dell U2415 `HDMI-A-1` above it at `0x-1200`.

**Symptom:** with the laptop showing workspace 4 and the Dell showing workspace 3, both bars highlighted the same number — the unfocused monitor's bar reported the *focused* monitor's workspace rather than its own.

**Result:** running this PR's change (each bar resolving `Hyprland.monitorFor(window.screen).activeWorkspace` instead of the global `Hyprland.focusedWorkspace`) as a cloned widget fixes the `SUPER+N` case here.

Two notes for anyone else testing:

1. I also needed the `refreshMonitors()` addition from #10187 to get the *move* case right. After `SUPER+SHIFT+ALT+<arrow>` the cached `activeWorkspace` is stale, so this PR on its own still showed the destination monitor's pre-move workspace. With both applied I verified via `grim` captures of each output that the two bars simultaneously mark different and correct workspaces — `HDMI-A-1` marking ws 8 while `eDP-1` marked ws 1, matching `hyprctl monitors`.

2. Saving a cloned widget's `.qml` logs `Local plugin changed, reloading: <id>` but does not re-instantiate the widget, so new bindings silently never run. `omarchy restart shell` is needed or you are testing the old code.

kevzakaria:
Third setup confirming this, and still reproducing on Omarchy 4.0.3-1.

**Setup:** Omarchy 4.0.3-1, Hyprland 0.56.2-2, quickshell 0.3.1-1. Two outputs side by side with different scales: `DP-2` 2560x1440 at scale 1.25 (logical 2048x1152 at 0,0) and `HDMI-A-1` 3840x2160 at scale 2 (logical 1920x1080 at 2048,0). Bar at the bottom.

**Method:** to keep the comparison honest I put the stock `omarchy.workspaces` and a per-monitor-fixed widget next to each other in the same bar, so both readings of the same compositor state are visible in one frame, and captured each output separately with `grim -g`.

**Symptom:** `DP-2` displaying workspace 2 without keyboard focus, `HDMI-A-1` focused on workspace 3. The stock row on the `DP-2` bar marks 3, a workspace that is not on that monitor, while the row beside it marks 2. On the focused monitor both rows agree, which is what makes this easy to miss on a single screen. State at capture time:

```
$ hyprctl monitors -j | jq -r '.[]|"\(.name) ws=\(.activeWorkspace.id) focused=\(.focused)"'
DP-2 ws=2 focused=false
HDMI-A-1 ws=3 focused=true
```

There is a second half to the same binding that I have not seen mentioned: `opacity: occupied || focused`. A workspace that is displayed on the unfocused monitor but has no windows on it is therefore drawn dim, exactly like a workspace nobody is using, so that bar carries no indication whatsoever of what its own monitor is showing. Checked directly by moving `DP-2` to the empty workspace 4 while `HDMI-A-1` kept focus: the stock row marked 3 and drew 4 dim. Keying the dim off the monitor's own workspace as well as `occupied` covers that case.

**Result:** this PR's `QsWindow.window.screen` plus `Hyprland.monitorFor` approach fixes it here. I arrived at the same two lines independently before finding this PR, which is some evidence that it is the idiom the shell already implies: `Bar.slotWindow` and the tray resolve their window the same way.

One data point on scope. The line is being copied into third-party widgets, so the bug is spreading through the plugin ecosystem rather than staying in the stock widget:

- `L0nE-F0x/lonefox.workspaces` carries the identical `focused` binding (fix: https://github.com/L0nE-F0x/lonefox.workspaces/pull/2)
- `jankeesvw/omarchy-workspace-name` reads the same global for its label, its indicator row and the workspace its naming panel edits (fix: https://github.com/jankeesvw/omarchy-workspace-name/pull/8)

Both were verified on the same two monitors with per output `grim -g` captures cross checked against `hyprctl monitors`. Fixing the stock widget would also give plugin authors a correct pattern to copy from.

Two notes for anyone testing on a mixed scale setup:

1. Grabbing each bar separately with `grim -g` needs Hyprland logical coordinates, not pixel sizes, or the crop lands on the wrong output when the scales differ.
2. Seconding the reload note above: saving the widget file logs `Local plugin changed, reloading: <id>` and `omarchy-shell shell rescanPlugins` returns success, but the bar widget instance is not rebuilt. I confirmed that with a `Component.onCompleted` probe that never printed until `omarchy restart shell`, so a patch tested without a restart looks like it does nothing.

mikebenner:
Accidentally closed — the GitHub fork `mikebenner/omarchy` was deleted; this PR was not withdrawn and was never merged. GitHub could not reattach the restored fork to this number, so the same change is at #11976 (rebased onto current `quattro`).

@dhh @ryanrhughes

PR #9002, opened 2026-08-29
Repair legacy paths and privileged files left by retired installers

acrogenesis:
Some retired Omarchy 3 installers generated root-owned files with an unquoted heredoc delimiter. Those files are still on disk on installs that ran them, and nothing has ever removed them. This adds the cleanup, consolidates the overlapping legacy-path migration from #8175, and adds a repository-wide regression guard against baking user-writable paths into privileged files.

## The shape

```bash
cat <<EOF | sudo tee "/etc/udev/rules.d/99-power-profile.rules"
SUBSYSTEM=="power_supply", ATTR{type}=="Mains", RUN+="/usr/bin/systemd-run ... $HOME/.local/share/omarchy/bin/omarchy-powerprofiles-set"
EOF
```

`<<EOF` rather than `<<'EOF'` means the *installing user's* shell expands the body before `tee` writes it, so the file that lands holds a literal `/home/<user>/.local/share/omarchy/bin/...`. No variable ever reaches udev.

That matters because the path resolves through `~/.local/share/omarchy`, a symlink the unprivileged user owns. udev runs `RUN+=` as root, so root ends up executing a file that user can swap out at will. The same applies to a sudoers grant naming a binary under a user's home, or a systemd unit with `ExecStop=` pointing there: privileged configuration should not take its input from a location the user can rewrite.

Current code has no instances — the rules moved to `/usr/bin` under `99-omarchy-*` names — but the one-shot migration that swept the old filenames was itself pruned, and `omarchy-upgrade-to-quattro` only runs while a machine is still crossing 3 to 4. An install that crossed earlier never gets swept.

## `migrations/1788102906.sh`

Consolidates the legacy-path work from #8175 with the security repair. It repoints `%H`, `~`, and expanded-home `~/.XCompose` includes from the retired checkout to `$OMARCHY_PATH/default/xcompose`, preserves every user-authored compose sequence, and restarts the input method when the file changes.

For `99-power-profile.rules` and `99-wifi-powersave.rules`, the migration removes only one of the five exact historical Omarchy-generated bodies. A safe administrator-authored rule survives unchanged. An edited rule that still executes the user-owned legacy helper is preserved byte-for-byte under a unique non-`.rules` quarantine name, preventing udev from executing it without discarding custom actions.

Fixed production paths and absolute root-owned helpers keep the privileged operands caller-independent. Each rule gets its own durable marker before mutation; a running udev daemon must accept `udevadm control --reload` before the marker clears. Retries cover failed reloads, concurrent runs, a safe replacement installed after interruption, quarantine collisions, and the no-running-udevd path.

## `bin/omarchy-upgrade-to-quattro`

The final package upgrade now runs before `omarchy-migrate`, ensuring newly packaged migrations are visible. The actual migration run, a still-pending queue, or an unverifiable pending state is fatal before reboot. This also catches the migration runner's successful pacman-lock timeout path.

## `migrations/1788025225.sh`

Three more files from installers that no longer exist:

- `/etc/sudoers.d/first-run` — passwordless sudo for the remainder of the first boot. `omarchy-first-run` was meant to delete it, but it clears its own `first-run.mode` guard as the first statement in the block and only reaches the removal after eight `set -e` steps, two of which touch the network, so any failure in between strands the grant with nothing left to retry it. The installers left nine distinct generated bodies and only four carry both `Cmnd_Alias` lines, so keying on those would walk past the earlier ones; instead every active line must be one the installer emitted, one of them must be the grant's own self-cleanup, and the account in a `/home/<user>/.../first-run.mode` cleanup path must equal the account receiving the grant.
- `/etc/sudoers.d/tsui` — named whatever `$(which tsui)` resolved to for the installing user, normally a binary under their own home that the vendor script had just written without sudo.
- `/etc/systemd/system/omarchy-plymouth-shutdown.service` — `ExecStop=` under the installing user's home, on an enabled unit, so it ran as uid 0 on every shutdown.

The unit is disabled and then removed, but deliberately never stopped: stopping it is precisely what would run the `ExecStop` being taken away. Ordering is disable, remove, `daemon-reload`. Before removal the migration records a root-owned pending-reload marker; only a successful reload clears it, so an interruption after deletion still retries the loaded systemd state even though the unit file is gone.

`/etc/sudoers.d` is 0750 root:root as shipped and `omarchy-migrate` runs as the logged-in user, so both sudoers files are classified and read through `sudo`. The elevated process emits an explicit `file`, `other`, or `missing` state; a failed sudo call can no longer be mistaken for absence. If elevation cannot be obtained, the migration exits non-zero with explicit instructions to have an administrator run `omarchy-migrate`. Migrations remain strictly ordered and synchronous: the failed migration stays pending and the queue stops, so no later migration can be marked complete against unrepaired state. A root-owned machine marker lets non-sudo users complete their per-user run after one privileged account has inspected and repaired the host. `/etc/systemd/system` remains directly readable for the unit predicate.

The two file formats also part company on a comment inside a continuation, so the shared line reader is told which one it is reading. `visudo -cf` parses a spec ending in a backslash, then a comment, then a second spec as two live specs, so sudo ends the logical line at the comment and keeps what came before it — dropping that half would hide an administrator's grant and let the file read as entirely installer-written. systemd resumes instead: `systemd-analyze verify` on `ExecStop=\` + `; c` + a path resolves that path. Each behaviour has a test.

The sudoers predicates also treat `#include`, `#includedir` and `#` followed by a digit as active lines rather than comments, which is what sudo does: `plugins/sudoers/toke.l` matches the first two as directives in the INITIAL state, and its comment rule excludes a digit or `-digit` so those reach the ID token as a numeric uid user spec. Dropping such a line would let a file still carrying an active directive read as though it held only generated lines.

The privileged artifacts are removal-only; neither migration rewrites the content of a sudoers file.

`agents/skills/migrations.md` gets the matching exception. Pre-4 layout work normally belongs in `bin/omarchy-upgrade-to-quattro`, which cannot reach installs that already crossed; the upgrade command ends by running `omarchy-migrate`, so a single migration covers both populations without a second copy of the predicate to keep correct.

## `test/shell.d/privileged-heredoc-test.sh`

Fails the suite when a privileged write embeds an install-time expansion. A heredoc is flagged when its delimiter is unquoted, its body contains an expansion the installing shell would act on, and its output reaches `/etc`, `/usr`, `/opt`, `/srv`, `/boot` or `/var/lib` — via `sudo tee`, `sudo dd`, a plain redirect, or an `install`/`cp`/`mv` of a generated scratch file one hop later. Literal and variable scratch destinations are both traced, and destinations written as variables are resolved from the file's own assignments.

An escaped `\$VAR` does not count: that is left literal for a root daemon to expand at runtime, which bakes nothing in. `install/3-config.sh` used both mechanisms in one heredoc on purpose — `$USER` expanded at install time because it is a username, `\$TERM` escaped for systemd — and the check has to keep telling those apart.

Sites that genuinely need install-time expansion declare it inline:

```bash
# omarchy:heredoc-expands paths=none -- $servers is normalized single-line DNS data
```

`paths=` is machine-checked as a set against the expansions the scanner finds to be path-shaped, so declaration order cannot create a false mismatch. Adding a `$HOME/...` to an already-annotated heredoc still makes the declaration false and trips the check rather than inheriting the old exemption. Nested expansions in parameter operators and defaults are scanned recursively, so `${target:-$HOME/...}` cannot hide that path behind `target`. A path expansion anchored under a root-owned prefix, as in `/etc/systemd/system/$unit`, is correctly not path-shaped.

The command parser preserves both `>` and `>>`, treats only `dd of=` as a `dd` destination, distinguishes arithmetic left shifts from heredoc operators, and follows Bash delimiter rules: plain `<<EOF` requires an exact unindented terminator, while `<<-EOF` strips leading tabs only. Pipeline consumers are reconstructed after a heredoc terminator and before the body when a backslash-escaped physical newline continues the command. Each boundary has a fixture that fails under the previous implementation.

The path can also sit a hop or more away from the heredoc. `helper="$HOME/.local/share/omarchy/bin/agent"` followed by `ExecStart=$helper` puts no slash in the token and no literal path in the value, and a name introduced with a packaged value and later reassigned under `$HOME` would be judged on the value it no longer holds. Any name assigned a user-writable root anywhere in the file is therefore treated as path-shaped and is never rescued by its other assignment, with fixtures for the one-hop, two-hop and shadowed cases.

Eight sites across five `bin/` scripts are reported by the scan. Seven are annotated with a reason — DNS addresses in `omarchy-dns`, a literal PAM line in `omarchy-setup-security-fingerprint`, kernel cmdline parameters and usernames in `omarchy-upgrade-to-quattro`, and the fixed bind-mount anchors in `omarchy-windows-vm`. The Windows VM declaration is `paths=EXPECTED_STORAGE,EXPECTED_SHARED`: both values are root-protected anchors derived from the authenticated caller uid and pinned to source inodes opened and validated before the compose file is written; the remaining expansions are revalidated scalar settings.

The eighth is a real change rather than an exemption: `bin/omarchy-provision-owner` expanded a unit name that was already a constant on the line above, so its delimiter is now quoted and the name hardcoded in the body. The generated unit file is byte-identical, verified by rendering both versions and diffing.

## Parser semantics

The predicates model three different parsers, and each behaviour below was checked against the tool itself rather than against a reading of its source:

| | comment inside a continuation | empty assignment | file ends mid-continuation |
|---|---|---|---|
| udev | resumes (`udevadm verify` complains on line 1) | n/a | line ignored |
| systemd | resumes (`systemd-analyze verify` resolves the directive) | non-empty entries append; empty resets the list | directive is live |
| sudo | ends the logical line and keeps it (`visudo -cf` parses two specs) | n/a | file rejected outright |

So a comment is tested before continuations are joined, the pending half is flushed for sudoers and carried for the other two, non-empty `ExecStop=` assignments append, and only an empty assignment resets that command list. The pending line is emitted at EOF. Each row has a test, and the continuation and list-accumulation guards fail under their previous incorrect implementations.

## Tests

The final branch passes all focused and PR-specific checks:

- privileged-heredoc regression guard: 34 assertions\n- consolidated XCompose/udev migration: 36 assertions
- Quattro upgrade lifecycle: 24 assertions
- retired-installer cleanup: 47 assertions
- migration runner: 6 assertions
- Bash syntax, ShellCheck warning-level analysis, and `git diff --check`

`config-test.sh` still stops at its existing workstation prerequisite because the companion `omarchy-pkgs` checkout is unavailable; the checks before that prerequisite pass. The same missing-checkout failure reproduces on pristine `quattro`.

## Live machine verification

At PR head `58c399de`, a Quattro workstation had the exact generated `/etc/sudoers.d/first-run` grant and the one-line `/etc/sudoers.d/tsui` grant still present, with neither per-user migration marker nor the machine-wide repair marker. Running the real migration queue from this branch completed with status 0. Both grants were absent afterward, the `1788025225.sh` and `1788102906.sh` user markers and root-owned `1788025225` machine marker were present, and `omarchy-migrate --pending` reported an empty queue. The workstation already had the packaged XCompose include and no legacy udev or Plymouth unit, so those paths correctly remained no-ops.

The five focused suites were then rerun on that same workstation: privileged-heredoc 34/34, retired-installer cleanup 47/47, consolidated XCompose/udev 36/36, migration runner 6/6, and Quattro lifecycle 24/24 — 147 assertions with final status 0.

## Supersedes #8175

#9002 now covers all behavior from #8175. Its unique XCompose repair is folded into `1788102906.sh`; its udev cleanup is replaced by the narrower security implementation above. The migration test asserts that exactly one migration owns the two legacy udev filenames, preventing the broad #8175 cleanup from being merged alongside it and preempting durable reload recovery.

Once this branch is accepted, #8175 can close as superseded.


migrations/1788025225.sh
```diff
@@ -0,0 +1,203 @@
...
+      return 0
+    fi
+  done < <(active_lines "$file" "#;")
+
+  return 1
+}
+
+first_run_sudoers="$sudoers_dir/first-run"
+if [[ -f $first_run_sudoers ]] && first_run_sudoers_is_generated "$first_run_sudoers"; then
+  as_root rm -f "$first_run_sudoers"
```
mdisec: This migration runs as a user, not root.  

`/etc/sudoers.d` is root-protected. Therefore the normal user cannot see or read first-run, so the whole fix fails ?

mdisec: that is sorted I guess after the last push

ryanrhughes (approved):

PR #9006, opened 2026-08-29
FIX: Deterministic false failures in the shell smoke test

wellatleastitried:
# Overview
`test/shell.d/runtime-smoke-test.sh`'s IPC-handler-collision check - which tolerates up to `screens - 1` "another handler is registered for target X" warnings per widget to account for the bar being legitimately instantiated once per screen - deterministically failed on this 2-screen desktop with "saw 3 for 2 screen(s)" on a clean checkout, for two compounding reasons:
1. The test built its synthetic hot-reload plugin fixture via three separate filesystem operations (`mkdir` + two file writes) directly inside `~/.config/omarchy/plugins`, the same directory quickshell's `PluginRegistry.qml` watches with `inotifywait`, causing a spurious `onLocalPluginChanged` -> full `reloadPlugins()` immediately after startup and superimposing an extra, unintended registration pass on top of the legitimate one; and
2. Even after eliminating that, the check greps the *entire* accumulated log at the end of the script, by which point two more fully legitimate, by-design reloads had already run earlier (the deliberate manifest-rename hot-reload test and an explicit `rescanPlugins` IPC call), each contributing its own `screens - 1` collisions per target and superimposing to `1 (startup) + 1 (rename) + 1 (rescan) = 3` - arithmetic, not flakiness, which is why it failed 100% of the time rather than intermittently.

The fix applies both changes together: the fixture is now built in a staging directory outside the watched tree and moved into place with a single atomic `mv`, removing the spurious startup reload; and the check now snapshots the log's line count right after the initial "shell IPC lists plugin metadata" pass (the first point at which the legitimate startup registration burst is guaranteed to be flushed) and restricts its `grep` to only that startup slice via `head -n`, so it measures exactly the one registration pass it was always meant to validate regardless of how many more by-design reloads the rest of the script performs afterward. Both changes were necessary - the atomic-move fix alone still left "saw 3" failing because the log-scoping issue was untouched - and were verified by three consecutive clean passes of the test in isolation plus a full `./test/shell` re-run showing the failure count drop from 4 files to the 3 pre-existing, unrelated environmental failures, with no stray quickshell processes or unintended working-tree changes left behind.

## Verification

The focused smoke test passes after the fix. The full `./test/shell` run now reports only the three pre-existing environment-dependent failures that require an `omarchy-pkgs` checkout.

<details>
<summary>Before — deterministic false failure on two monitors</summary>

```text
ok - shell IPC lists plugin metadata
ok - installed plugin changes reload without an explicit rescan
ok - shell IPC routes built-in ids to enabled clones
ok - shell IPC returns effective shell config
ok - shell IPC summon and hide contract works
ok - plugin IPC contracts respond
ok - image selector IPC survives plugin rescan
ok - default bar layout renders expected module slots
ok - runtime geometry keeps update before indicators
ok - direct panel IPC opens and closes default panels
not ok - each widget registers its IPC handler once per screen (saw 3 for 2 screen(s))
```

The log contained three registration passes for every target: startup, the intentional manifest hot reload, and the explicit `rescanPlugins` reload.

</details>

<details>
<summary>After — focused smoke test passes</summary>

```text
ok - shell IPC lists plugin metadata
ok - installed plugin changes reload without an explicit rescan
ok - shell IPC routes built-in ids to enabled clones
ok - shell IPC returns effective shell config
ok - shell IPC summon and hide contract works
ok - plugin IPC contracts respond
ok - image selector IPC survives plugin rescan
ok - default bar layout renders expected module slots
ok - runtime geometry keeps update before indicators
ok - direct panel IPC opens and closes default panels
ok - each widget registers its IPC handler once per screen
Disabled omarchy.audio
ok - bar remove reloads shell config and updates bar layout
ok - bar put places a widget after the one it names
ok - bar put leaves a widget already on the bar alone
```

</details>

### DIFF
--- a/test/shell.d/runtime-smoke-test.sh
+++ b/test/shell.d/runtime-smoke-test.sh
@@ -51,10 +51,17 @@ ln -s "$ROOT/config" "$test_root/config"
 ln -s "$ROOT/bin" "$test_root/bin"
 
 # Every plugin under ~/.config/omarchy/plugins hot-reloads, whoever wrote it.
+# Built in a staging directory and moved into place as a single rename so the
+# plugin watcher sees one atomic directory creation instead of a mkdir plus
+# two separate file writes — the multi-step version fires onLocalPluginChanged
+# more than once right after startup, each triggering its own full
+# reloadPlugins() pass and inflating the IPC-handler-collision count below
+# with duplicates unrelated to the number of screens.
 hot_reload_id="acme.hot-reload"
+hot_reload_staging="$TMPDIR/.hot-reload-staging"
 hot_reload_dir="$test_home/.config/omarchy/plugins/$hot_reload_id"
-mkdir -p "$hot_reload_dir"
-cat >"$hot_reload_dir/manifest.json" <<JSON
+mkdir -p "$hot_reload_staging" "$test_home/.config/omarchy/plugins"
+cat >"$hot_reload_staging/manifest.json" <<JSON
 {
   "schemaVersion": 1,
   "id": "$hot_reload_id",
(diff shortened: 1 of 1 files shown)


Greytakespics:
Really like this fix. Nice catch on the test accidentally picking up the extra reloads, and the log scoping makes the intent of the check a lot clearer.

Issue #8998, opened 2026-08-29
1Password window is not visible when accessing desktop over sunshine/moonlight


spuder:
### System details

Nvidia 3060, AMD 1800x

### What's wrong?

When accessing omarchy desktop over moonlight/sunshine, the 1password text enter box appears as black. 

<img width="1378" height="811" alt="Image" src="https://github.com/user-attachments/assets/ca47ab4e-a748-47bf-b38e-03618f39e4af" />

<img width="5712" height="4284" alt="Image" src="https://github.com/user-attachments/assets/66944218-5cb7-4272-a4c3-37e9d515a24a" />

## Research

1. Hardware acceleration is turned on in 1password (as suggested in the omarchy docs) https://omarchy.org/manual/troubleshooting/
Turning off acceleration and rebooting has no change in behavior.

spuder:
Root cause isn't 1Password-specific — it's Sunshine's default Wayland capture path failing against Nvidia's proprietary buffer layout, and the black box just happens to be whatever's on screen at the time.

**Diagnosis:**

Hyprland's compositor output buffers on Nvidia are allocated with a proprietary block-linear/tiled format:
```
GBM: Allocated a new buffer ... format XR24 with modifier 0x300000000606014 : BLOCK_LINEAR_2D,HEIGHT=4,KIND=6,GEN=2,SECTOR=1,COMPRESSION=0
```

Sunshine's default capture backend on Wayland (`wlr`, via `wlr-export-dmabuf`/`zwlr_screencopy_manager_v1`) can't correctly read that buffer back. You can see it fail in Sunshine's own log:
```
Error: [wlgrab] Could not initialize display with the given hw device type.
Error: Couldn't scale frame: Invalid argument
Info: Encoder [nvenc] failed
```
It silently falls back to software `libx264`, but the underlying frame data is still corrupted — coming through as solid black. This affects the **entire captured desktop**, not just 1Password; a `grim` screenshot taken with the same protocol while 1Password was open came back solid black too, the whole frame, not just that window. 1Password just happened to be what was on screen when this was reported.

[...]

PR #9011, opened 2026-08-29
Default Nvidia users to KMS capture in Sunshine

spuder:
Sunshine's default Wayland capture path (wlr-export-dmabuf) can't read back Nvidia's proprietary tiled/compressed framebuffer layout on Hyprland: streams silently fall back to a corrupted capture that renders as solid black, affecting the whole desktop rather than any one app. NVFBC, Nvidia's own fast capture API, doesn't support Wayland at all, so KMS (reading straight from the DRM/KMS layer) is the only capture method that reliably works on Nvidia here.

Have omarchy-install-service-sunshine write capture = kms into sunshine.conf when an Nvidia GPU is detected, unless the user already has an explicit capture setting. AMD/Intel installs are unaffected.

Reported and diagnosed in #8998 (Nvidia 3060, black 1Password window over Moonlight/Sunshine) — confirmed the underlying capture path was producing a fully black frame regardless of which app was on screen, and verified this fix resolves it on the reporter's hardware.

### DIFF
--- a/bin/omarchy-install-service-sunshine
+++ b/bin/omarchy-install-service-sunshine
@@ -15,6 +15,7 @@ SUNSHINE_ADMIN_EXEC="omarchy-launch-webapp $SUNSHINE_ADMIN_URL --ignore-certific
 SUNSHINE_ICON_SOURCE="/usr/share/sunshine/web/images/logo-sunshine-45.png"
 HYPR_AUTOSTART_FILE="$HOME/.config/hypr/autostart.lua"
 HYPR_AUTOSTART_ENTRY='o.launch_on_start("sunshine")'
+SUNSHINE_CONFIG_FILE="$HOME/.config/sunshine/sunshine.conf"
 
 open_ufw_port_for_private_lans() {
   local proto="$1"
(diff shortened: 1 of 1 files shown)


dk424242:
Another independent failure mode that this same fix resolves — worth folding into the rationale, because it presents nothing like the black-frame case in #8998.

On an RTX 3070 (driver 610.57.04), Hyprland 0.56.2, `sunshine 2026.516.143833-4`, the wlr path doesn't produce a black frame. It silently drops Sunshine to **software encoding**.

The NVENC probe fails and falls through to libx264:

```
Info: Trying encoder [nvenc]
Info: Creating encoder [h264_nvenc]
Error: Couldn't scale frame: Invalid argument
Info: Encoder [nvenc] failed
...
Info: Found H.264 encoder: libx264 [software]
```

The trigger looks like a **scaled output**. This monitor is 3840x2160 at `scale = 2`, and Sunshine's Wayland enumeration reports both sizes:

```
Info: [wayland] Resolution: 3840x2160
Info: [wayland] Name: DP-1
Info: [wayland] Logical size: 1920x1080
```

Setting `capture = kms` fixes it outright — same hardware, same driver, no other change:

```
Info: Screencasting with KMS
Info: Found monitor for DRM screencasting
Info: Found H.264 encoder: h264_nvenc [nvenc]
Info: Found HEVC encoder: hevc_nvenc [nvenc]
```

What makes this one nasty is that nothing surfaces to the user. Streaming "works" — it just burns CPU and leaves the GPU idle. Without reading the journal you'd never know you weren't using NVENC. That seems like a decent additional argument for defaulting Nvidia users to KMS rather than treating it as a fix only for the black-frame symptom.

(`av1_nvenc` also fails the probe here, but that's expected on Ampere — AV1 encode is Ada and later, so it's unrelated.)

dk424242:
Correction to my comment above — I tested the guess and it was wrong.

I suggested the NVENC probe failure looked like it was triggered by a **scaled** output (3840x2160 at `scale = 2` reporting a 1920x1080 logical size). That isn't it.

I created a headless output with no size mismatch whatsoever:

```
DP-1        3840x2160  scale=2
HEADLESS-3  1920x1080  scale=1
```

and pointed Sunshine at it on the wlr path (`capture = wlr`, `output_name = HEADLESS-3`). Physical and logical size identical, scale 1, nothing to reconcile. Same result:

```
Info: [wayland] Resolution: 1920x1080
Info: [wayland] Logical size: 1920x1080
Info: Screencasting with Wayland's protocol
Info: Found H.264 encoder: libx264 [software]
```

Switching that same session back to `capture = kms` immediately restores hardware encoding:

```
Info: Screencasting with KMS
Info: Found H.264 encoder: h264_nvenc [nvenc]
Info: Found HEVC encoder: hevc_nvenc [nvenc]
```

So scaling is a red herring. The wlr path can't feed NVENC on Nvidia here **at all**, which lines up with the framing in your PR description — wlr-export-dmabuf not being able to read back Nvidia's tiled/compressed framebuffer layout.

That makes the case for this PR stronger rather than weaker: it isn't an edge case affecting people with fractional or 2x scaling, it's every Nvidia user on the wlr path. Some get your black frame, some get a silent drop to software encoding, and the latter group has no symptom to go looking for.

PR #9015, opened 2026-08-29
Make scrolling columns resizable at workspace edge

hancengiz:
## Summary

- use the scrolling layout’s native `colresize` message for horizontal resize shortcuts
- keep `window.resize` on dwindle layouts and on vertical scrolling (`up`/`down`)
- cover normal, small, and large resize steps, plus scrolling direction, with a focused regression test

## Why

The generic horizontal resize dispatcher cannot grow the last visible scrolling column when there is no neighboring window on its right. This makes users open a temporary window, resize the original column, and close the temporary window again.

Routing the same shortcuts through `colresize` on scrolling workspaces changes the column width directly and works at the workspace edge. `colresize` follows the tape axis, so it is used only when the effective scrolling direction is `left` or `right`. Shortcut chords and non-scrolling behavior remain unchanged.

## Testing

- `bash test/shell.d/hyprland-resize-bindings-test.sh`
- live verified on Hyprland 0.56.2 by growing and restoring the rightmost scrolling column without a window to its right

### DIFF
--- a/default/hypr/bindings/tiling.lua
+++ b/default/hypr/bindings/tiling.lua
@@ -1,3 +1,13 @@
+local function scrolling_direction(workspace)
+  local opts = workspace.layout_opts
+  local direction = type(opts) == "table" and opts.direction or hl.get_config("scrolling.direction")
+  if direction == "left" or direction == "right" or direction == "up" or direction == "down" then
+    return direction
+  end
+
+  return "right"
+end
+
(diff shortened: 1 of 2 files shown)


hancengiz:
@codex

hancengiz:
@copilot

hancengiz:
@copilot

hancengiz:
@codex code review

PR #9045, opened 2026-08-30
Keep the screensaver wordmark across monitor scales

### LINKED ISSUE #9027: Screensaver wordmark clipped horizontally at default monitor scale on 1080p displays
## Description

On a 1920x1080 laptop panel with **entirely stock Omarchy configuration**, the screensaver's OMARCHY wordmark is clipped on both the left and right edges — it renders as "MARCH", with the leading `O` sliced vertically in half and the trailing `Y` cut off past the right edge.

This is a stock-config-vs-stock-config mismatch, reachable without any user customization:

| Shipped default | Value |
|---|---|
| `config/hypr/monitors.lua` | `omarchy_monitor_scale = "auto"` → Hyprland picks **scale 2** on this 1080p/143-DPI panel |
| `default/foot/screensaver.ini` | `font=JetBrainsMono Nerd Font:size=18`, `pad=0x0` |
| `default/omarchy/branding/screensaver.txt` | **81 columns** wide × 10 rows |

At scale 2 the fullscreen logical surface is 960x540, and a foot window with `size=18` measures **66 columns** — 15 short of the 81 the art needs. `ttfx` is invoked with `--canvas-width 0 --anchor-canvas c --anchor-text c`, so it sizes the canvas to the terminal and centre-crops the art, removing roughly 7 columns from each side.

Measured directly, same machine, same shipped `screensaver.ini`, only the monitor scale differing:

```
$ foot --config=/usr/share/omarchy/default/foot/screensaver.ini --fullscreen -e stty size
scale 2     ->  16 rows,  66 cols     <- art needs 81, clipped
scale 1.25  ->  27 rows, 106 cols     <- fits
```

The clipping threshold on a 1920px-wide panel is **scale > ~1.63** (cell width measured at 14.55 logical px for `size=18`; 81 × 14.55 = 1179 logical px minimum). So `"auto"` resolving to 2 lands well inside the broken range, while 1.0/1.25/1.5 are all fine.

This is not foot-specific — the launcher hardcodes `--font-size=18` for ghostty and `--override font_size=18` for kitty as well, so the same clipping should occur on any of the four supported terminals.

## Steps to reproduce

1. On a 1920x1080 display (~143 DPI, so Hyprland's `auto` scaling selects 2), keep the stock `~/.config/hypr/monitors.lua` — i.e. `omarchy_monitor_scale = "auto"`.
2. Confirm the scale: `hyprctl monitors -j | jq '.[0].scale'` → `2`.
3. Run `omarchy-launch-screensaver force`.
4. The wordmark is clipped at both sides.

Setting `omarchy_monitor_scale = 1.25` (or anything ≤ 1.6) and relaunching renders the full wordmark.

## Expected

The screensaver logo renders fully, with margins, at Omarchy's own default monitor scale.

## Actual

The logo is horizontally centre-cropped, losing ~7 columns from each side. Vertical is unaffected (10 rows into 16).

## Captures

**Before — stock `scale = "auto"` (2), 960x540 logical, 66 cols:** reads "MARCH", `O` bisected at the left edge, `Y` gone at the right.

**After — `scale = 1.25`, 1536x864 logical, 106 cols:** full "OMARCHY" with margins.

## Possible directions

Not prescribing a fix, but the mismatch is between a fixed 18pt font and a variable logical viewport. Options that would close it:

- Size the screensaver font from the terminal's measured width instead of hardcoding `18` — e.g. have `omarchy-screensaver` re-exec with a font size derived from `stty size` after the existing `wait_for_terminal_resize`.
- Pass `--canvas-width`/`--anchor` such that `ttfx` scales rather than crops, if it supports that.
- Ship a narrower fallback wordmark used when the measured column count is below 81.

## System

```
Omarchy:   4.0.0-1
Hyprland:  0.56.2 (efb5099)
Kernel:    7.1.8-arch1-3
CPU:       Intel Core i7-8665U
GPU:       Intel UHD Graphics 620 (WhiskeyLake-U GT2) [8086:3ea0]
Display:   eDP-1, BOE 0x0869, 1920x1080@60, 340x190 mm (~143 DPI), scale "auto" -> 2
Terminal:  foot 1.27.0-2 (xdg-terminal-exec --print-id -> foot.desktop)
ttfx:      0.3.2-1
```

Note: `omarchy debug --no-sudo --print` is documented but not present in 4.0.0-1 (`Unknown Omarchy command`), and no `debug` route appears in `omarchy commands --all`; the details above were collected manually.

PyRo1121:
Fixes #9027

The screensaver launchers hardcoded an 18pt font, but the wordmark needs 81 columns. On a 1080p/143-DPI panel, Omarchy's own `omarchy_monitor_scale = "auto"` resolves to scale 2, whose 960-column logical viewport fits only 66 columns — `ttfx` centre-cropped the art down to "MARCH".

Each terminal launch now sizes its font from the target monitor's scale (`18 / scale`, clamped to 6–18) via a new `omarchy_screensaver_font_size` helper in `default/bash/fns/screensaver`. The wordmark's physical size — and therefore its column budget — stays constant on every display: scale 1 keeps the shipped 18, scale 2 gets 9, and sub-unity scales cap at 18.

- foot gets `--font="JetBrainsMono Nerd Font:size=$font"`, Alacritty `--option font.size=$font`, and Ghostty/Kitty their existing flags with the computed value
- per-monitor scales are honored (mixed-scale setups launch each terminal with its own size)

Tests:
- `bash test/shell.d/screensaver-font-size-test.sh`
- `bash -n bin/omarchy-launch-screensaver default/bash/fns/screensaver`
- `git diff --check`

### DIFF
--- a/bin/omarchy-launch-screensaver
+++ b/bin/omarchy-launch-screensaver
@@ -13,6 +13,10 @@ fi
 focused=$(omarchy-hyprland-monitor-focused)
 terminal=$(xdg-terminal-exec --print-id)
 
+# omarchy_screensaver_font_size scales each terminal's font to the monitor so
+# the wordmark keeps its physical size (and column budget) at any DPI scale.
+. "${OMARCHY_PATH:-/usr/share/omarchy}/default/bash/fns/screensaver"
+
 case $terminal in
 *Alacritty* | *ghostty* | *foot* | *kitty*) ;;
 *)
@@ -49,25 +53,27 @@ wait_for_screensaver_window() {
   done
 }
 
-for m in $(hyprctl monitors -j | jq -r '.[] | .name'); do
+while read -r m scale; do
+  [[ -n $m ]] || continue
   hypr_focus_monitor "$m"
+  font=$(omarchy_screensaver_font_size "${scale:-1}")
 
   case $terminal in
   *Alacritty*)
-    hypr_exec alacritty --class=org.omarchy.screensaver --config-file "$OMARCHY_PATH/default/alacritty/screensaver.toml" -e omarchy-screensaver
+    hypr_exec alacritty --class=org.omarchy.screensaver --config-file "$OMARCHY_PATH/default/alacritty/screensaver.toml" --option "font.size=$font" -e omarchy-screensaver
     ;;
   *ghostty*)
-    hypr_exec ghostty --class=org.omarchy.screensaver --config-file="$OMARCHY_PATH/default/ghostty/screensaver" --font-size=18 -e omarchy-screensaver
+    hypr_exec ghostty --class=org.omarchy.screensaver --config-file="$OMARCHY_PATH/default/ghostty/screensaver" --font-size="$font" -e omarchy-screensaver
     ;;
   *foot*)
-    hypr_exec foot --app-id=org.omarchy.screensaver --config="$OMARCHY_PATH/default/foot/screensaver.ini" -e omarchy-screensaver
+    hypr_exec foot --app-id=org.omarchy.screensaver --config="$OMARCHY_PATH/default/foot/screensaver.ini" --font="JetBrainsMono Nerd Font:size=$font" -e omarchy-screensaver
     ;;
   *kitty*)
-    hypr_exec kitty --class=org.omarchy.screensaver --override font_size=18 --override window_padding_width=0 -e omarchy-screensaver
+    hypr_exec kitty --class=org.omarchy.screensaver --override "font_size=$font" --override window_padding_width=0 -e omarchy-screensaver
     ;;
   esac
 
   wait_for_screensaver_window
-done
+done < <(hyprctl monitors -j | jq -r '.[] | "\(.name) \(.scale // 1)"')
 
 hypr_focus_monitor "$focused"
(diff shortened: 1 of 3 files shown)


mefelosc:
Nice work on this! Could I ask you to consider portrait monitors too, please? The scale-only formula still crops there. Tested on a 1080p panel rotated to portrait at scale 1.25:

18 / 1.25 = 14 → ~77 columns for the 81-column wordmark, so it still crops (before shot: only "MARCH" visible).
hyprctl monitors -j reports unrotated dimensions, so with transform = 1/3 the logical width needs to swap w/h first: height / scale = 864px here, not width / scale.
Sizing from that rotation-aware logical width — something like floor(logical / 84 / 0.8), clamped to 8–18 — gives 12pt → ~90 columns here, verified fitting (after shot). It's general rather than portrait-specific: wide monitors still compute ≥18, so current behavior is unchanged there.

I could only test on foot + portrait, so please take the numbers with a grain of salt — would appreciate if someone could validate on the other terminals. Happy to test a revision on portrait hardware.

screensaver-portrait-before
screensaver-portrait-after

bjarneo:
Automated duplication check: this pull request looks similar to #7307, which covers the same screensaver font scaling. I keep that one open and close this one to consolidate review.

PR #9069, opened 2026-08-30
Fall back to a second host, then a TCP handshake, where the internet probe fails

### LINKED ISSUE #9068: Network panel reports Timeout and 100% packet loss on a healthy link where the network filters outbound ICMP
### System details

Lenovo Legion Pro 5 16ADR10, AMD Ryzen 9 8945HX, NVIDIA RTX 5070 Mobile,
MediaTek MT7925 Wi-Fi, kernel 7.1.9-arch1-2, Omarchy 4.0.1-1

### What's wrong?

The network panel reports `Timeout` and `100% packet loss` on a link that is
working normally, whenever the network filters outbound ICMP.

`bin/omarchy-network-status` measures internet reachability with a single ICMP
echo to a hardcoded host, with no way to change it:

```bash
internet_probe=1.1.1.1                       # :8
LC_ALL=C ping -n -c 1 -W 1 "$host"           # :54
```

Where ICMP is filtered every sample comes back empty, `Model.js:260` maps an
empty sample to `"Timeout"`, and `pingPacketLossPercent()` (`Model.js:197`)
accumulates those empties across its 24-sample window until the row reads 100%.
The connection is fine throughout: pages load, downloads run at full speed, and
the throughput counters in the same panel show real traffic.

Measured on the affected network (QUT eduroam), summarised rather than pasted:

| probe | result |
| --- | --- |
| `ping -c 4 -W 5` to 1.1.1.1, 8.8.8.8, 9.9.9.9, 1.0.0.1 | 100% loss, all four |
| `ping -c 4 -W 5` to the default gateway (172.20.0.1) | 0% loss |
| `curl https://1.1.1.1` | TCP connect 0.010s, HTTP 301, total 0.031s |

The gateway answers because it is inside the network; nothing past the border
does. Happy to attach verbatim output and `omarchy-debug` next time I am on that
network if it would help.

### Why this isn't only an edge case

ICMP echo carries no port numbers, so a default-deny egress policy written in
terms of ports has no rule that can match it: "permit 53, 80, 443, deny the
rest" drops ping without anyone deciding to block it. That policy shape is
standard on university and corporate WLANs, on captive-portal appliances in
hotels and airports, and behind CGNAT, which cannot always demultiplex echo
replies back to a client. Home broadband, where this is presumably tested,
rarely filters it -- which may be why it has gone unreported.

It also means the panel is measuring the wrong thing. What a user cares about
is whether a TCP session opens, and on this link that took 10ms while the panel
reported total failure.

### Reproducing

Needs a network that filters outbound ICMP; any campus or corporate WLAN with a
default-deny egress ACL will do. I hit it on eduroam, but eduroam is not the
variable -- #7257 is an eduroam user whose ping and packet-loss rows read
correctly, so that site permits ICMP. The condition is the filtering, not the
SSID.

1. Join a network that filters outbound ICMP.
2. Confirm the link works -- load a page, or run the curl above.
3. Open the network panel. Ping reads `Timeout`, packet loss climbs to 100%.

### Note on the packet loss row

The two rows answer different questions and only one of them can be rescued by
a different probe. A handshake can supply a latency where echo is filtered, but
it cannot count echoes, so there is nothing to put in the loss row: reporting 0%
would claim a clean link nobody measured, and a probe that quietly filled the
gap left by a lost echo would report a link shedding half its packets as
healthy. Worth flagging in case a fix looks simpler than it is.

Happy to send a patch if a direction is welcome.

---
Filed by Claude Opus 5 (1M context) via Claude Code. The measurements above were
recorded on the affected network in an earlier session and are summarised rather
than pasted; verbatim output and `omarchy debug` can follow next time I am on it.

### LINKED ISSUE #9261: omarchy-network-status: 100% packet loss when 1.1.1.1 is unreachable
## Description

`omarchy-network-status` hardcodes `1.1.1.1` (Cloudflare DNS) as the sole internet ping probe. On networks that block ICMP to `1.1.1.1`, the network panel shows "Timeout" and "100% Packet Loss" — even though the internet connection is fully functional and data is flowing normally.

## Steps to Reproduce

1. Connect to a network that blocks ICMP to `1.1.1.1` (but allows it to other hosts like `8.8.8.8`)
2. Open the network panel
3. Observe "Timeout" / "100% Packet Loss" despite working internet

## Expected Behavior

Panel shows real latency and 0% packet loss.

## Actual Behavior

Panel shows Timeout / 100% loss.

## Suggested Fix

Try multiple probe addresses (`1.1.1.1`, `8.8.8.8`, `9.9.9.9`, `208.67.222.222`) and use the first that responds. I've locally patched `omarchy-network-status` with this approach and it works correctly:

```bash
ping_best_internet_probe() {
  local result
  result=$(ping_latency_ms "$internet_probe")
  if [[ -n $result ]]; then
    printf '%s' "$result"
    return
  fi
  local probe
  for probe in $internet_fallback_probes; do
    result=$(ping_latency_ms "$probe")
    if [[ -n $result ]]; then
      printf '%s' "$result"
      return
    fi
  done
}
```

## Environment

- Omarchy latest
- Network: `ozgenius_5G` (5GHz Wi-Fi)
- Router/gateway reachable, internet functional, but `1.1.1.1` ICMP blocked

AronBakes:
Fixes #9068. Fixes #9261.

`omarchy-network-status` measures the internet hop with a single ICMP echo to
one hardcoded host. Anything that filters either the **host** (an ISP or router
that blackholes 1.1.1.1 — #9261) or the **protocol** (a university or corporate
WLAN with a default-deny egress ACL, a captive-portal appliance, CGNAT — #9068)
makes every sample come back empty, so the panel reads `Timeout` and climbs to
`100% packet loss` on a link that is working fine.

Two layers, each only reached when the one before it gets nothing:

1. **Second host.** 1.1.1.1 and 8.8.8.8 are pinged together and the first reply
   in list order wins. Pinging in parallel rather than in sequence means a blocked
   host costs no time, and nothing changes on a network where 1.1.1.1 answers.
2. **TCP handshake.** Only when no echo returns from either, a handshake to
   each host's port 443 is timed and reported as a **separate** field.

### Why the handshake time is comparable

"You substituted a different measurement into the same row" is the fair
objection, so, measured on one machine:

| probe | result |
| --- | --- |
| `/dev/tcp` handshake | 3.8 – 5.7 ms |
| `curl` `time_connect` | 4.1 – 4.4 ms |
| ICMP RTT (`ping` avg) | 4.24 ms |

Both cross the network exactly once — SYN/SYN-ACK is a round trip like
echo/reply — which is why they agree. The row is still labelled `Ping (TCP)`
rather than passing one off as the other, because a handshake can additionally
carry server-side accept latency.

### Why it is a separate field

`internet_ping_ms` keeps meaning "an echo returned in this long", because that
is what the packet loss row counts. Folding the handshake into it would report a
link shedding half its packets as perfectly healthy — the fallback would quietly
fill the gap left by every lost echo. A sample the handshake rescued is still a
lost echo.

Where no echo returns all window but the handshake keeps landing, loss is not
measurable from here at all, so the row holds at `--` rather than claiming a
clean link nobody measured.

| scenario | Ping | Packet Loss |
| --- | --- | --- |
| healthy | `20 ms` | `0%` |
| 1.1.1.1 blocked, 8.8.8.8 answers | `20 ms` | `0%` |
| real 50% loss, handshake rescues each miss | `20 ms` | `50%` |
| ICMP filtered, link healthy | `10 ms` (`Ping (TCP)`) | `--` |
| link dead | `Timeout` | `100%` |
| status script without the new field | `20 ms` | unchanged |

Each is a test in `test/shell.d/network-test.sh`. The script-side cases run
`print_ping_samples` for real against a stubbed `ping`, rather than grepping the
source, and were checked to fail when the second host or the per-host handshake
is removed.

### Notes

- Uses bash's `/dev/tcp` under a `timeout` rather than `curl`, which no script in
  `bin/` currently needs. Verified bounded on all four outcomes: reachable,
  refused, blackholed (returns at 2.001s), unresolvable.
- The router row deliberately stays ICMP-only. The gateway is inside the network,
  answers echo even where the border filters it, and need not have the port open.
- On a network where 1.1.1.1 answers, output is identical to today: no extra line
  is emitted and no socket opened. The one added cost everywhere is a second ICMP
  packet per poll, in parallel with the first.
- Other designs are reasonable and I have no attachment to this one — reading
  NetworkManager's existing connectivity state, or simply rendering `--` when
  every sample is empty while `rx_bytes` is still climbing, would both be smaller
  than this. Happy to redo it either way.

### DIFF
--- a/bin/omarchy-network-status
+++ b/bin/omarchy-network-status
@@ -6,6 +6,8 @@
 
 verbose=false
 internet_probe=1.1.1.1
+internet_probes="$internet_probe 8.8.8.8"
+internet_probe_port=443
 
 case "${1:-}" in
   "")
@@ -54,31 +56,58 @@ ping_latency_ms() {
   LC_ALL=C ping -n -c 1 -W 1 "$host" 2>/dev/null | awk -F'time[=<]' '/time[=<]/ { split($2, parts, " "); print parts[1]; exit }'
 }
 
+# Where ICMP is filtered, time a TCP handshake instead: one round trip, like an echo.
+tcp_latency_ms() {
+  local host=$1 port=$2
+
+  LC_ALL=C timeout 2 bash -c '
+    start=${EPOCHREALTIME/./}
+    exec 3<>"/dev/tcp/$1/$2" || exit 1
+    finish=${EPOCHREALTIME/./}
+    exec 3>&- 3<&-
+    elapsed=$((finish - start))
+    printf "%d.%03d\n" $((elapsed / 1000)) $((elapsed % 1000))
+  ' _ "$host" "$port" 2>/dev/null
+}
+
 print_ping_samples() {
   local gateway=$1
-  local tmpdir router_file internet_file
-  local router_pid="" internet_pid=""
+  local tmpdir probe echo_ms="" handshake_ms=""
+  local router_pid="" internet_pids=()
 
   omarchy-cmd-present ping || return
   tmpdir=$(mktemp -d) || return
-  router_file="$tmpdir/router"
-  internet_file="$tmpdir/internet"
 
   if [[ -n $gateway ]]; then
-    ping_latency_ms "$gateway" >"$router_file" &
+    ping_latency_ms "$gateway" >"$tmpdir/router" &
     router_pid=$!
   fi
 
-  ping_latency_ms "$internet_probe" >"$internet_file" &
-  internet_pid=$!
+  # Probed together, so a host the network blocks costs no time; first reply in list order wins.
+  for probe in $internet_probes; do
+    ping_latency_ms "$probe" >"$tmpdir/$probe" &
+    internet_pids+=($!)
+  done
 
   if [[ -n $router_pid ]]; then
     wait "$router_pid"
-    printf 'router_ping_ms\t%s\n' "$(cat "$router_file")"
+    printf 'router_ping_ms\t%s\n' "$(cat "$tmpdir/router")"
   fi
 
-  wait "$internet_pid"
-  printf 'internet_ping_ms\t%s\n' "$(cat "$internet_file")"
+  wait "${internet_pids[@]}"
+  for probe in $internet_probes; do
+    [[ -s $tmpdir/$probe ]] && { echo_ms=$(cat "$tmpdir/$probe"); break; }
+  done
+  printf 'internet_ping_ms\t%s\n' "$echo_ms"
+
+  # Separate field: internet_ping_ms stays echo-only because the loss row counts it.
+  if [[ -z $echo_ms ]]; then
+    for probe in $internet_probes; do
+      handshake_ms=$(tcp_latency_ms "$probe" "$internet_probe_port") && break
+    done
+    printf 'internet_tcp_ms\t%s\n' "$handshake_ms"
+  fi
+
   rm -rf "$tmpdir"
 }
 
--- a/shell/plugins/panels/network/Model.js
+++ b/shell/plugins/panels/network/Model.js
@@ -210,7 +210,9 @@ function formatPacketLoss(percent, hasSamples) {
   if (hasSamples === false) return "--"
 
   var value = parseInt(percent, 10)
-  if (!value || value < 0) return "0%"
+  // Negative means loss could not be measured: see pingLatencyState.
+  if (value < 0) return "--"
+  if (!value) return "0%"
   return value + "%"
 }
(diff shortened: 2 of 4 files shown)


bjarneo:
Automated duplication check: this pull request looks similar to #7932, which covers the same network status second probe. I keep that one open and close this one to consolidate review.

PR #9080, opened 2026-08-30
Update basecamp/omarchy references to omacom/omarchy in contributing.md

PixelatedContinuum:
`basecamp/omarchy` renamed to `omacom/omarchy`. The old org still redirects
today, so these links aren't broken yet, but a stale reference risks breaking
if the old name is ever reused, and `gh` commands that hardcode it (the two
examples in this file) stop being copy-pasteable the day the redirect goes
away.

Updates all four references in `default/agents/skills/omarchy/contributing.md`
(the repo link, the discussions link, and the two example `gh` commands) to
the current org.

Scope note: a broader grep turns up `basecamp/omarchy` in several other files
(shell scripts, a systemd unit, a Hyprland app config, `.github/ISSUE_TEMPLATE/config.yml`),
some of which may be functional rather than just prose. Left those out of this
PR to keep it to one coherent change; happy to follow up separately if useful.

## Testing

`./test/all` shows 5 pre-existing failures unrelated to this change (3 need a
sibling `omarchy-pkgs` checkout not present in this environment; 2 look
environment-dependent). Confirmed all 5 reproduce identically on `quattro`
with this commit removed, so nothing here is a regression. This change only
touches prose/URLs in one markdown file.

### DIFF
--- a/default/agents/skills/omarchy/contributing.md
+++ b/default/agents/skills/omarchy/contributing.md
@@ -3,13 +3,14 @@
 Read this when the user wants to report an Omarchy bug, suggest a feature, or
 contribute a fix upstream.
 
-Omarchy lives at https://github.com/basecamp/omarchy. Route requests to the
-right place:
+Omarchy lives at https://github.com/omacom/omarchy (renamed from
+basecamp/omarchy; the old name still redirects, but link the current
+one). Route requests to the right place:
 
 - **Verified bugs** -> GitHub issues. Issues are for validated bugs only, not
   support requests.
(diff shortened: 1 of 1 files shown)


PixelatedContinuum:
The URL changes here have already landed on quattro, so this is superseded. The only thing left in the diff is a short note explaining that basecamp/omarchy still redirects, which I added so a reader hitting an old link knows why it works. Not worth a PR on its own, so I am closing this, but happy for anyone to pick that line up if it seems useful.

PR #9081, opened 2026-08-30
[Security] Keep sudo from resolving system tools out of a user-writable checkout

Chessing234:
## Summary
- `omarchy-dev-link` prepended the checkout's `bin/` to sudo `secure_path`
- Any writable `pacman` / `systemctl` next to `omarchy-pkg-add` then ran as root on the next update
- Publish only `omarchy-*` into a root-owned `/var/lib/omarchy/dev-bin` farm and point `secure_path` there; unlink clears the farm

## Test plan
- [x] `bash test/shell.d/dev-link-test.sh` (covers planted `pacman` not published into the farm)
- [ ] `omarchy dev link ~/omarchy --no-reboot` then confirm `sudo which pacman` is still `/usr/bin/pacman` and `sudo which omarchy-pkg-add` resolves via the farm


getaaron (approved):

acrogenesis (changes requested):
Thanks for working on this and for identifying the broader `secure_path` behavior.

We've discussed Dev-mode findings within the Omarchy Security team, and our position is that they are outside our security threat model. Dev mode is intentionally open: enabling it explicitly trusts a mutable checkout for development, including privileged Omarchy workflows. This therefore does not qualify as a security vulnerability.

The proposed solution also removes important Dev-mode behavior:

- It snapshots only the `omarchy-*` commands present when `omarchy dev link` runs, so newly added or renamed commands require relinking.
- Removed commands can fall through to packaged versions, causing developers to test the wrong code unknowingly.
- It excludes the `omarchy` router because `omarchy` does not match `omarchy-*`, so `sudo omarchy …` can use the packaged router.
- The farm is rebuilt non-atomically and introduces derived state that can become stale.

The current head also fails `bash test/shell.d/dev-link-test.sh` at the new symlink assertion with `grep: warning: stray \ before t`. Farm removal is not covered by `dev-unlink-test.sh`, and `docs/file-layout.md` still documents the existing live-checkout behavior.

Given that this is outside our security scope and the proposed architecture undermines the purpose and developer experience of Dev mode, I don't consider this PR eligible for merge and recommend closing it.

A future non-security hardening proposal could still be considered if it preserves a genuinely live checkout without requiring developers to maintain a separate command snapshot.

PR #9082, opened 2026-08-30
[Security] Key passwordless sudo grants off id -un, not USER

Chessing234:
## Summary
- Resolve the passwordless sudoers drop-in path and grant principal from `id -un` instead of spoofable `$USER`.
- Cap grant duration at 120 minutes.
- Boot orphan cleanup already lives on tip via `etc/tmpfiles.d/omarchy-nopasswd-sudo.conf` (`r!`) plus the fail-closed timer; this PR no longer re-adds that unit.

## Why
`$USER` is inherited and attacker-controllable in the environment. A caller can point the drop-in at another name or inject sudoers syntax through the username field. Using `id -un` keys the grant to the real UID mapping.

## Test plan
- [x] Spoofed `USER=...` no longer redirects the grant / fails closed in regression tests
- [x] Overlong duration is rejected
- [ ] CI green on rebase

### DIFF
--- a/bin/omarchy-sudo-passwordless
+++ b/bin/omarchy-sudo-passwordless
@@ -4,12 +4,22 @@
 # omarchy:args=[MINUTES]
 # omarchy:requires-sudo=true
 
-NOPASSWD_FILE="/etc/sudoers.d/99-omarchy-nopasswd-${USER}"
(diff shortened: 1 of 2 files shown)


Chessing234:

Chessing234:
folded into #12266 for easier review.

PR #9083, opened 2026-08-30
[Security] Escape TUI desktop entries like web apps

Chessing234:
## Summary
- `omarchy-tui-install` wrote Name/Exec/Icon without Freedesktop escaping (webapp install already did)
- A newline in the name injected a second `Exec=`; a slash wrote outside `applications/` (e.g. into autostart); an unquoted launch command reshaped Exec
- Reuse webapp escaping, refuse `/` in names, quote the launch command, and validate downloaded icons as `image/*`

## Test plan
- [x] `bash test/shell.d/tui-install-escaping-test.sh`
- [ ] Interactive install with a normal name still launches from the app launcher

### DIFF
--- a/bin/omarchy-tui-install
+++ b/bin/omarchy-tui-install
@@ -3,14 +3,26 @@
 # omarchy:summary=Create a desktop launcher for a terminal UI app
 # omarchy:args=[name command window-style icon-url-or-name]
 
-set -e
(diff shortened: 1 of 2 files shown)


Chessing234:

Chessing234:
folded into #12266 for easier review.

PR #9084, opened 2026-08-30
[Security] Keep the update transcript out of world-writable /tmp

Chessing234:
## Summary
- `omarchy update` always wrote `/tmp/omarchy-update.log`
- A local user who pre-created that path as a symlink could redirect `script(1)` into the victim's files
- Stage the log under `$XDG_RUNTIME_DIR/omarchy-update/` (or a 0700 cache under `~/.cache/omarchy/update`); analyze-logs follows the same path and refuses symlinks

## Test plan
- [x] `bash test/shell.d/update-log-path-test.sh`
- [ ] Run `omarchy update -y` (or a dry path) and confirm the transcript is not under `/tmp/omarchy-update.log`

### DIFF
--- a/bin/omarchy-update
+++ b/bin/omarchy-update
@@ -7,9 +7,30 @@
 
 set -e
 
+# Transcript must not land on a world-writable predictable path. A local user
+# who pre-creates /tmp/omarchy-update.log as a symlink can redirect script(1)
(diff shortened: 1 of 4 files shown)


Chessing234:

Chessing234:

PR #9085, opened 2026-08-30
[Security] Shell-quote install helpers before bash -c

Chessing234:
## Summary
- `omarchy-install-font` / `omarchy-install-app` interpolated names and packages into a string run by `bash -c`
- `omarchy-install-and-launch` quoted the message and desktop id but left packages raw
- A crafted package argument became arbitrary commands that `omarchy-pkg-add` then escalated via sudo
- `%q`-quote every interpolated piece

## Test plan
- [x] `bash test/shell.d/install-helpers-quote-test.sh`
- [ ] Menu-driven font/app installs still work with normal package names

Made with [Cursor](https://cursor.com)

### DIFF
--- a/bin/omarchy-install-and-launch
+++ b/bin/omarchy-install-and-launch
@@ -14,8 +14,9 @@ if [[ -z $name || -z $packages || -z $desktop_id ]]; then
 fi
 
 printf -v install_message '%q' "Installing ${name}..."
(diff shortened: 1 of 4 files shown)


Chessing234:
superseded by #7843 on quattro — the install helpers already %q-quote.

PR #9087, opened 2026-08-30
Update remaining basecamp/omarchy references to omacom/omarchy

PixelatedContinuum:
Follow-up to #9080, which fixed the same stale `basecamp/omarchy` → `omacom/omarchy`
reference in `default/agents/skills/omarchy/contributing.md`. A repo-wide grep
turned up eight more files with the same pre-rename org name:

- `manual/30-updates.md` — two release links
- `manual/49-omarchy-on.md` — three `discussions/<number>` links (numeric IDs
  are unaffected by the rename, only the org prefix changes)
- `default/systemd/user/omarchy-speaker-tuning.service` — the `Documentation=` URL
- `default/agents/skills/diagnose-crash/reporting.md` — five `gh` command examples
  (same pattern already fixed in `contributing.md`)
- `default/hypr/apps/jetbrains.lua` — a code comment linking the PR that explains
  the window rule
- `bin/omarchy-update-confirm` — the "What's new" release link shown before an update
- `.github/ISSUE_TEMPLATE/config.yml` — the Suggestion contact link
- `bin/omarchy-upgrade-to-quattro` — the archive-tarball URL in
  `populate_legacy_hypr_defaults`

## Why these and not the other two

Two more files reference the org (`bin/omarchy-channel-set`'s dev-channel
`git clone`, and `test/shell.d/channel-test.sh`, which asserts that exact
clone URL verbatim). Left those for a separate PR since the fix there has to
touch the test in the same commit or it breaks, which is a different shape of
change from a plain reference swap.

## Verified before changing anything

None of these are live breakage — confirmed via `curl` and `git ls-remote`
that GitHub's redirect covers the web UI, the `codeload` archive-tarball
endpoint, and the git smart-HTTP clone protocol identically today. So this is
future-proofing (the redirect could go away, or the old org name could get
reused), not a bug fix.

For `omarchy-upgrade-to-quattro` specifically: it fetches the `master` branch
by name (not the current default `quattro`), which looked like it might be a
second, unrelated bug — but the function is `populate_legacy_hypr_defaults`,
and `master` still exists upstream, so that's deliberately pulling legacy
content for a compatibility shim. Left the branch name alone; only the org
changed.

## Testing

`./test/all` — same 5 pre-existing, unrelated failures as on a clean `quattro`
checkout (3 need a sibling `omarchy-pkgs` checkout not present here, 2 look
environment-dependent). Nothing new.

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,7 +1,7 @@
 blank_issues_enabled: false
 contact_links:
   - name: Suggestion
-    url: https://github.com/basecamp/omarchy/discussions/categories/suggestions
+    url: https://github.com/omacom/omarchy/discussions/categories/suggestions
     about: Suggest a new feature, change to existing feature, or other ideas in Discussions.
   - name: Support
     url: https://omarchy.org/discord
--- a/bin/omarchy-update-confirm
+++ b/bin/omarchy-update-confirm
@@ -8,7 +8,7 @@ gum style --border normal --padding "1 2" \
   "• You cannot stop the update once you start!" \
   "• Make sure you're connected to power or have a full battery" \
   "" \
-  "What's new: https://github.com/basecamp/omarchy/releases/latest"
+  "What's new: https://github.com/omacom/omarchy/releases/latest"
 
 echo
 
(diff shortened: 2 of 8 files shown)


sprajs:
The current `quattro` branch already contains all eight files’ URL changes from this PR. The runtime/documentation URL cleanup landed in commit https://github.com/omacom/omarchy/commit/d78ea9e1, and the issue-template/agent-skill references are covered by merged #10532. I checked the complete PR patch in reverse against the current branch and it applies cleanly, so there is no remaining hunk to land. This looks safe to close as already addressed.

PixelatedContinuum:
Confirmed, and thanks for tracing it back. I checked all eight files against quattro and none of them carry a basecamp/omarchy reference any more, so d78ea9e1 and the merged #10532 cover everything this PR touched. Closing as already addressed.

PR #9088, opened 2026-08-30
Point the dev-channel checkout clone at omacom/omarchy

PixelatedContinuum:
Same rename as #9080 and #9087 (`basecamp/omarchy` → `omacom/omarchy`), split
out separately because this one is coupled to a test rather than a plain
reference swap.

`bin/omarchy-channel-set`'s `dev` channel clones the source repo directly to
`~/omarchy`:

```bash
[[ -d $checkout/.git ]] || git clone https://github.com/basecamp/omarchy.git "$checkout"
```

`test/shell.d/channel-test.sh` asserts that exact clone command, including the
URL, verbatim — so fixing the script without the test would fail CI, and
fixing the test without the script would just pin the stale URL harder. Both
move together in this commit.

## Verified before changing anything

`git ls-remote https://github.com/basecamp/omarchy.git HEAD` resolves
correctly today (GitHub's git smart-HTTP protocol follows the org rename), so
this isn't fixing a live clone failure — it's the same future-proofing as the
other two PRs.

## Testing

`bash test/shell.d/channel-test.sh` — all 33 assertions pass, including "dev
clones the source checkout to ~/omarchy" and "dev activates the checkout
before changing or updating packages," against the updated URL.

`./test/all` — same 5 pre-existing, unrelated failures as a clean `quattro`
checkout (see #9087 for detail); `channel-test.sh` is not among them.

### DIFF
--- a/bin/omarchy-channel-set
+++ b/bin/omarchy-channel-set
@@ -34,7 +34,7 @@ validate_dev_checkout() {
 
 link_dev_checkout() {
   local checkout="$1"
-  [[ -d $checkout/.git ]] || git clone https://github.com/basecamp/omarchy.git "$checkout"
+  [[ -d $checkout/.git ]] || git clone https://github.com/omacom/omarchy.git "$checkout"
 
   omarchy-dev-link "$checkout" --no-reboot
 }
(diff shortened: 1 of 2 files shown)


PixelatedContinuum:
Checked this against quattro and both changes are already in place, so there is nothing left for this PR to land. Closing as already addressed.

PR #9089, opened 2026-08-30
Restore lock password focus whenever the screen wakes

rdoupe:
## Symptom

After resuming a laptop from suspend, the lock screen sometimes ignores keystrokes. The password field looks normal but swallows everything typed until you click it. Locking manually is unaffected — it only shows up coming back from sleep.

## Cause

`omarchy-system-sleep-lock` locks *before* suspending and waits for `secure`, so the session lock is already up when the machine goes down. The journal shows the lock surface then outlives the suspend — there is no second `lock-requested`/`secure=true` on the way back:

```
16:36:35  lock-requested → secure=true     ← locked before sleep
16:38:33  Finished System Suspend
16:39:00  secure=false → unlocked          ← password finally entered
```

So `LockView` is never rebuilt, and its `Component.onCompleted` focus never runs again. Meanwhile the compositor tears down and rebuilds the keyboard across the resume (an xkb keymap recompile lands ~5s after wake), which is where the field's keyboard focus goes.

That left `MouseArea.onClicked` as the only thing that could restore it:

```qml
onClicked: { root.wakeRequested(); root.forcePasswordFocus() }
onPositionChanged: root.wakeRequested()      // wakes the screen, never refocuses
```

`idleBlankTimer` has DPMS'd the display off by then, so you have to interact to see anything at all. Move the mouse and the panel lights up with an unfocused field; click and it works. Hence the intermittence — it depended on whether you happened to click or just move.

## Fix

Handle the wake signal itself, so every wake path recovers focus instead of just one:

```qml
onWakeRequested: if (inputEnabled) forcePasswordFocus()
```

The click handler no longer needs to special-case focus. Guarding on `inputEnabled` keeps the lock preview — which reuses this view with input disabled — from focusing a field the user cannot type into.

## Testing

Adds `test/shell.d/lock-password-focus-test.sh`, following the existing headless `LockView` fixtures. It focuses the field, drops focus the way a resume does while the view stays alive, wakes, and asserts focus came back; it also asserts waking does not focus the field while `inputEnabled` is false. Verified it fails without the change:

```
"failures": [ "waking the lock screen restores focus to the password field" ]
```

`./test/shell` passes 211/216. The 5 failures (`config`, `snapper`, `theme-install-guards`, `unowned-system-paths`, `windows-vm-mount-boundary`) reproduce unchanged on a clean checkout of master on this machine and are unrelated.

Not verified on real hardware across a live suspend/resume yet, since that needs a locked session to reproduce in the act — the automated coverage stands in for it.

### DIFF
--- a/shell/plugins/lock/LockView.qml
+++ b/shell/plugins/lock/LockView.qml
@@ -67,6 +67,14 @@ Item {
     syncingPasswordText = false
   }
 
+  // The lock surface outlives a suspend, so the focus taken when the view was
+  // built can be gone by the time the machine resumes -- the compositor
+  // rebuilds the keyboard on the way back, and the view is never rebuilt to
+  // take focus again. Anything that wakes the screen has to leave the field
+  // ready to type into; previously only a click did, so waking the blanked
+  // screen with a mouse move left the password field swallowing keystrokes.
+  onWakeRequested: if (inputEnabled) forcePasswordFocus()
+
   onPasswordTextChanged: syncPasswordText()
   onInputEnabledChanged: {
     if (inputEnabled) Qt.callLater(forcePasswordFocus)
@@ -115,7 +123,7 @@ Item {
     MouseArea {
       anchors.fill: parent
       hoverEnabled: true
-      onClicked: { root.wakeRequested(); root.forcePasswordFocus() }
+      onClicked: root.wakeRequested()
       onPositionChanged: root.wakeRequested()
     }
 
@@ -131,6 +139,7 @@ Item {
 
       TextInput {
         id: passwordInput
+        objectName: "passwordInput"
         anchors.fill: parent
         anchors.topMargin: inputField.borderTop
         // Reserve the fingerprint icon's width on both sides so the centered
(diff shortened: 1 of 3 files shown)


rdoupe:
Closing as a duplicate — I should have listed the open PRs before opening this one. #7592, #8560 and #8869 all predate it and diagnose the same thing.

For anyone triaging: #8560 covers the most cases of the four. This PR only refocused on `wakeRequested`, which needs a pointer event first — resume with the screen already lit and type without touching the trackpad, and it does nothing. #8560's `onActiveFocusChanged` reclaims focus the moment it is lost, with no interaction, and without the 100ms poll that #7592/#8869 run for the whole lock.

Confirmed the underlying bug on real hardware (AMD laptop, single eDP panel, Hyprland, omarchy 4.0.1): `omarchy-system-sleep-lock` locks before suspending, so the lock surface outlives the sleep and `Component.onCompleted` never runs again. The journal shows no second `secure=true` on resume, which is also why #7164's `onSecureChanged` hook does not fire on this path. I will follow up on #8560 with a suspend/resume verification.

PR #7815, opened 2026-08-22
Add Wi-Fi password visibility toggle

AFOliveira:
Add an eye button to show or hide a Wi-Fi passphrase while typing. Passwords start masked. Closing the prompt, changing networks, submitting, or connecting successfully resets visibility; leaving a scan clears the secret after a fixed five-second grace period without extending the deadline on repeated refreshes.

The button supports Tab, Return, keypad Enter and Space, has accessible show/hide labels, and returns focus to the password field after activation. The complete branch is integrated with Quattro `e848f1df`.

Validation at `26dad29c30f973790ab5d2c72a90c4fb898996a7`: network tests and 112 CLI checks pass, and an independent Daybreak review found no code blocker. In a disposable 4.0.2 VM, the full running shell passed inspected masked/revealed screenshots, Tab/Return/Space interaction, submit remasking, disappearance expiry, and close/reset checks. That VM used deterministic synthetic Wi-Fi rows because it has no wireless radio; real association, enterprise authentication, and screen-reader interaction remain unverified. All temporary fixture changes were removed.

PR #9091, opened 2026-08-30
Tell agents to check open PRs before writing a fix

rdoupe:
## Problem

`AGENTS.md` says nothing about looking for existing work before opening a PR — no mention of pull requests, duplicates, or searching. With the backlog where it is, that omission is expensive: one bug draws several independent fixes that all have to be read and compared before any can be merged.

Two clusters found without looking hard:

**Lock screen losing password focus after suspend — four open PRs:**

| PR | Approach |
|---|---|
| #7164 | refocus on `secure` becoming true |
| #7592 | 100ms polling `Timer` while unfocused |
| #8560 | `onActiveFocusChanged` reclaims focus |
| #8869 | 100ms polling `Timer` + `focus: true` |

**Wi-Fi password reveal toggle — two open PRs:** #7815 and #8019 add the same toggle.

I am not reporting this from the outside. I hit the bug, diagnosed it, wrote a fix and opened #9089 — the fifth PR on that lock bug — before finding the other four. I have since closed it as a duplicate. This change is the fix for what actually went wrong, which was the process, not the patch.

## Change

Adds a `Pull Requests` section to `AGENTS.md`: search before writing, search twice because one query does not catch a PR that describes the same area in different words, and when an open PR already covers the change, add to it — confirm the bug, review the approach, contribute a missing test — rather than opening a competing one.

Docs only, no code or test changes. Written as full lines with a fenced `bash` block, matching the file's own Style section and the Refresh Pattern example.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -74,6 +74,19 @@ guidance does not drift from the router.
 - Commits should be atomic: include only one coherent change or fix, and do not mix unrelated work.
 - Commit messages should be succinct and describe the change being made.
 
+# Pull Requests
+
+Search the open pull requests before writing a fix. The backlog is large enough that one bug attracts several independent fixes: the lock screen losing password focus after suspend had four open PRs (#7164, #7592, #8560, #8869) before a fifth was filed, and the Wi-Fi password reveal toggle has two (#7815, #8019). A duplicate costs review time instead of saving it.
+
+Search by symptom, then again in a wider form, because a PR describing the same area in different words will not match the first query:
+
+```bash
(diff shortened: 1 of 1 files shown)


rdoupe:
Closing — wrong file. `default/agents/skills/omarchy/contributing.md` is titled "Reporting Issues and Submitting PRs" and already has a `## Submitting a PR` section, so that is where this belongs. It also ships in the `omarchy` skill, which is what an agent on a user's machine actually reads before contributing upstream — the duplicate gets written before anyone opens AGENTS.md. Reopening against that file.

PR #9100, opened 2026-08-30
fix(update): guard confirm_reboot in omarchy-update-restart under unattended runs

### LINKED ISSUE #9079: omarchy-update-restart: confirm_reboot has no unattended guard, still hangs under -y
Follow-up to #8780 — flagged separately per the note in that issue's timeline, since both open fixes (#8824, #8894) say "Fixes #8780" and neither touches this path.

## What happens

`omarchy update -y` can still block on interactive input, this time in `omarchy-update-restart` rather than the orphan-package step #8780 covers. `confirm_reboot()` in `bin/omarchy-update-restart` has no unattended guard at all — not even the (currently ineffective) TTY check `omarchy-update-orphan-pkgs` has:

```bash
confirm_reboot() {
  gum confirm "$1" && { omarchy-system-reboot; exit 0; }
}
```

It's called unconditionally from three sites in the same file, any one of which is a normal outcome of an update:

- the running kernel has no matching installed module directory (kernel was updated)
- `~/.local/state/omarchy/reboot-required` exists
- Hyprland's own binary was replaced while running (`readlink` on its `/proc/.../exe` shows `(deleted)`)

`bin/omarchy-update` runs `omarchy-update-restart` unconditionally at the end of every update, `-y` included, so this isn't an edge case — any update that touches the kernel or Hyprland hits it.

## Why this isn't just a repeat of #8780

It's the same failure mode with one less layer. `omarchy-update-orphan-pkgs` at least has `[[ ! -t 0 || ! -t 1 ]]`, which is exactly the check #8780 found doesn't fire under `-y` — `omarchy-update` re-execs the whole run through `script -qefc` for logging, and `script(1)` allocates a pty on both descriptors, so `-t 0`/`-t 1` read true even with nobody there to answer. `confirm_reboot` never had that check to begin with.

## What's already there for the fix

`omarchy-update` already exports `OMARCHY_UPDATE_UNATTENDED=1` on `-y`, before any pipeline step runs — including `omarchy-update-restart`. It's already inherited here with no plumbing needed; this looks like a local, single-function fix in `bin/omarchy-update-restart`, e.g.:

```bash
confirm_reboot() {
  if [[ ${OMARCHY_UPDATE_UNATTENDED:-} == 1 ]]; then
    echo "$1 Re-run omarchy update in a terminal to confirm."
    return
  fi
  gum confirm "$1" && { omarchy-system-reboot; exit 0; }
}
```

Report-and-continue rather than report-and-exit, since `confirm_reboot` is called from three separate sites and the shell-restart step still needs to run afterward — and continuing rather than auto-confirming keeps the same default #8824/#8894 chose for orphaned packages: an unattended run shouldn't take an action as disruptive as a reboot on its own say-so.

## Expected

`omarchy update -y` completes without blocking on any of the three reboot/restart prompts.

## System

- Omarchy 4.0.1-1

harshithnadig:
### Problem
In #9079, running `omarchy update -y` in unattended mode can still hang on interactive input if the Linux kernel, Hyprland binary, or `reboot-required` state file is present.

While `omarchy-update` exports `OMARCHY_UPDATE_UNATTENDED=1` on `-y`, `confirm_reboot()` in `bin/omarchy-update-restart` was calling `gum confirm "$1"` unconditionally without an unattended check.

### Solution
Check `${OMARCHY_UPDATE_UNATTENDED:-}` in `confirm_reboot()`. When unattended, log the reboot notification and return without blocking, allowing the remaining update and shell reload steps to complete cleanly.

Fixes #9079

### DIFF
--- a/bin/omarchy-update-restart
+++ b/bin/omarchy-update-restart
@@ -5,6 +5,11 @@
 echo
 
 confirm_reboot() {
+  if [[ ${OMARCHY_UPDATE_UNATTENDED:-} == 1 ]]; then
+    echo "$1 Re-run 'omarchy update' in a terminal to confirm reboot."
+    return
+  fi
+
   gum confirm "$1" && { omarchy-system-reboot; exit 0; }
 }
 


bjarneo:
Automated duplication check: this pull request looks similar to #8992, which covers the same update restart unattended guard. I keep that one open and close this one to consolidate review.

PR #9104, opened 2026-08-30
fix(service): enable Sunshine by canonical unit name and open admin UI port

### LINKED ISSUE #9099: omarchy install service sunshine always fails: enables 'sunshine.service' but package ships 'app-dev.lizardbyte.app.Sunshine.service'
### What happened

`omarchy install service sunshine` fails partway through. The package installs fine, then:

```
:: Running post-transaction hooks...
(1/5) Reloading user manager configuration...
...
```

Because `omarchy-install-service-sunshine` runs under `set -e`, the script aborts at that line and **silently skips every remaining step**:

- the ufw rules for the Moonlight ports (LAN + `tailscale0`) are never added
- the "Sunshine Admin" web app is never installed
- the `o.launch_on_start("sunshine")` autostart entry is never written to `~/.config/hypr/autostart.lua`

So the user is left with Sunshine installed but not running, not enabled at boot, and firewalled off — with no indication that most of the installer didn't run.

### Expected

Sunshine enabled and started, ports opened, admin web app installed.

### Cause

The installer hardcodes:

```bash
```

But the `sunshine` package ships the unit as `/usr/lib/systemd/user/app-dev.lizardbyte.app.Sunshine.service`. That unit declares:

```ini
[Install]
Alias=sunshine.service
```

`Alias=` is only materialized as a symlink **once the unit is enabled**, so `sunshine.service` does not exist on a fresh install. It is a chicken-and-egg: you cannot enable it by its alias before enabling it. This fails 100% of the time on a clean install.

Enabling it by its real name works and creates the alias:

```
Created symlink '/home/<user>/.config/systemd/user/sunshine.service' → '/usr/lib/systemd/user/app-dev.lizardbyte.app.Sunshine.service'.
Created symlink '/home/<user>/.config/systemd/user/graphical-session.target.wants/app-dev.lizardbyte.app.Sunshine.service' → '/usr/lib/systemd/user/app-dev.lizardbyte.app.Sunshine.service'.
```

### Steps to reproduce

1. On a system without Sunshine installed, run `omarchy install service sunshine`
2. Enter the sudo password, accept the pacman transaction
3. Observe `Failed to enable unit: Unit sunshine.service does not exist` and the script exiting before the firewall/webapp/autostart steps

### Suggested fix

In `omarchy-install-service-sunshine`, enable the unit by its real name:

```bash
```

Worth noting the unit is already `WantedBy=graphical-session.target`, so it autostarts on login under uwsm on its own — the additional `o.launch_on_start("sunshine")` Hyprland autostart entry looks like it would start a second instance.

### Minor: admin UI port not opened

Separately, `TCP_PORTS=(47984 47989 48010)` omits **47990**, the Sunshine configuration UI. Since pairing (setting the initial username/password and entering the Moonlight PIN) can only be done through that UI, and the installer explicitly opens the other ports on `tailscale0` for remote streaming, a user connecting over Tailscale can stream but cannot pair without separately opening 47990 or tunnelling it over SSH.

### System details

- Omarchy: `4.0.1-1`
- sunshine: `2026.516.143833-4` (omarchy repo)
- Kernel: `7.1.9-arch1-2`
- CPU: AMD Ryzen 7 9800X3D
- GPU: NVIDIA GeForce RTX 5090 (nvidia-open-dkms 610.57.04) + AMD Granite Ridge iGPU
- Compositor: Hyprland under uwsm, SDDM display manager

harshithnadig:
### Problem
In #9099, running `omarchy install service sunshine` fails on fresh installs with:
`Failed to enable unit: Unit sunshine.service does not exist`

### Cause
1. The `sunshine` package installs `/usr/lib/systemd/user/app-dev.lizardbyte.app.Sunshine.service`. The `sunshine.service` alias declared in the unit's `[Install]` section is only created once enabled; enabling by alias before it is enabled fails.
2. Port 47990 (the Sunshine Web Admin interface) was omitted from `TCP_PORTS`, preventing pairing over LAN/Tailscale.
3. `enable_hyprland_autostart` was creating a redundant launch entry when the systemd unit is already `WantedBy=graphical-session.target`.

### Solution
- Call `systemctl --user daemon-reload` and enable `app-dev.lizardbyte.app.Sunshine.service`.
- Add `47990` to `TCP_PORTS`.
- Remove redundant `autostart.lua` entry.

Fixes #9099

### DIFF
--- a/bin/omarchy-install-service-sunshine
+++ b/bin/omarchy-install-service-sunshine
@@ -5,7 +5,7 @@
 
 set -e
 
-TCP_PORTS=(47984 47989 48010)
+TCP_PORTS=(47984 47989 47990 48010)
 UDP_PORTS=(5353 47998 47999 48000 48002 48010)
 PRIVATE_CIDRS=(10.0.0.0/8 172.16.0.0/12 192.168.0.0/16)
 UFW_COMMENT="omarchy-sunshine"
@@ -13,8 +13,6 @@ SUNSHINE_ADMIN_APP="Sunshine Admin"
 SUNSHINE_ADMIN_URL="https://localhost:47990"
 SUNSHINE_ADMIN_EXEC="omarchy-launch-webapp $SUNSHINE_ADMIN_URL --ignore-certificate-errors"
 SUNSHINE_ICON_SOURCE="/usr/share/sunshine/web/images/logo-sunshine-45.png"
-HYPR_AUTOSTART_FILE="$HOME/.config/hypr/autostart.lua"
-HYPR_AUTOSTART_ENTRY='o.launch_on_start("sunshine")'
 
 open_ufw_port_for_private_lans() {
   local proto="$1"
@@ -60,18 +58,10 @@ install_admin_webapp() {
   omarchy-webapp-install "$SUNSHINE_ADMIN_APP" "$SUNSHINE_ADMIN_URL" "$SUNSHINE_ICON_SOURCE" "$SUNSHINE_ADMIN_EXEC"
 }
 
-enable_hyprland_autostart() {
-  mkdir -p "$(dirname "$HYPR_AUTOSTART_FILE")"
-  touch "$HYPR_AUTOSTART_FILE"
-
-  if ! grep -Fxq "$HYPR_AUTOSTART_ENTRY" "$HYPR_AUTOSTART_FILE"; then
-    printf '\n%s\n' "$HYPR_AUTOSTART_ENTRY" >>"$HYPR_AUTOSTART_FILE"
-  fi
-}
-
 echo "Installing Sunshine..."
 omarchy-pkg-add sunshine
-systemctl --user enable --now sunshine
+systemctl --user daemon-reload
+systemctl --user enable --now app-dev.lizardbyte.app.Sunshine.service
 
 echo "Opening Sunshine firewall ports..."
 open_ufw_ports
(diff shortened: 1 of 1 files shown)


bin/omarchy-install-service-sunshine
```diff
@@ -5,16 +5,14 @@
 
 set -e
 
-TCP_PORTS=(47984 47989 48010)
+TCP_PORTS=(47984 47989 47990 48010)
```
spuder, line 8: I think this will open the admin interface to anyone on the network. I don´t know if that is the correct default behavior. 

If this needs to be limited to just tailscale it would need to be a separate PRIVATE_CIDRS network

spuder:
Related: https://github.com/omacom/omarchy/pull/9003

harshithnadig:
Addressed the admin-plane exposure concern in commit `8a359197`:

- Port `47990` is no longer included in the private-LAN TCP allowlist.
- The Sunshine admin UI is opened only on an existing `tailscale0` interface; localhost access remains unchanged.
- Moonlight streaming ports remain available to the documented private LAN CIDRs and Tailscale.

The installer passes Bash syntax and a mocked UFW simulation verifying that no RFC1918 rule is emitted for `47990`.

bjarneo:
Automated duplication check: this pull request looks similar to #8879, which covers the same Sunshine canonical unit fix. I keep that one open and close this one to consolidate review. If you feel this is the wrong decision, please open the PR again with a note on the difference.

PR #9105, opened 2026-08-30
Use s2idle on T2 Macs with a discrete GPU

T1mmsquared:
`install/hardware/apple/fix-t2.sh` detects T2 Macs by PCI ID (`106b:180[12]`) and gives every one of them the same kernel command line, including `mem_sleep_default=deep`. That works on integrated-only models, but a T2 Mac with a discrete GPU can never complete a deep suspend.

On a MacBookPro16,1 (Radeon Pro 5500M), every suspend attempt fails and falls back:

```
PM: suspend entry (deep)
amdgpu 0000:03:00.0: GPU mode1 reset failed
amdgpu 0000:03:00.0: PM: pci_pm_suspend_noirq(): amdgpu_pmops_suspend_noirq [amdgpu] returns -22
amdgpu 0000:03:00.0: PM: dpm_run_callback(): pci_pm_suspend_noirq returns -22
amdgpu 0000:03:00.0: PM: failed to suspend noirq: error -22
PM: noirq suspend of devices failed
PM: suspend exit
PM: suspend entry (s2idle)   <- systemd retries, this one succeeds
```

Counted over the journal on this machine: **3 deep attempts, 3 `noirq suspend of devices failed`, 3 `GPU mode1 reset failed`** — a 100% failure rate. The machine still sleeps, because systemd retries with s2idle, so the symptom is easy to miss.

Two costs beyond the wasted transition:

1. Every lid close spends ~3 seconds suspending and then unwinding the entire device tree before the retry.
2. The failed pass tears down and recreates four Apple EFI variables each cycle — `AppleSecureBootFailureReason`, `AppleSecureBootKernelFailureReason`, `AppleSecureBootWindowsFailureReason`, `AppleFailedBootVolumeUUID` — so every suspend/resume writes NVRAM on hardware where that wear is not free.

## Approach

`mem_sleep_default=deep` was added deliberately in #6562, so this does not remove it. It selects the sleep state from the GPU configuration using the existing `omarchy-hw-hybrid-gpu` helper:

```bash
if omarchy-hw-hybrid-gpu; then
  sleep_state="s2idle"
else
  sleep_state="deep"
fi
```

Integrated-only T2 Macs produce a byte-identical config file to today, so #6562's fix is preserved exactly where it works. Only discrete-GPU models change, and they change to the state the kernel was already falling back to on its own.

## Included

- `install/hardware/apple/fix-t2.sh` — pick the sleep state from the GPU configuration
- `migrations/1788072498.sh` — move existing discrete-GPU installs off `deep` and rebuild the boot image, following the repair-marker pattern from `migrations/1785944594.sh` so a pre-reboot rerun stays idempotent
- `test/shell.d/t2-hardware-test.sh` — a stubbed `omarchy-hw-hybrid-gpu`, plus coverage for the discrete-GPU repair, its idempotency, the integrated-only path that must keep `deep`, and non-T2 hardware

## Testing

- `test/shell.d/t2-hardware-test.sh` — 10/10 pass (6 pre-existing, 4 new)
- `./test/all` — 4 failures, all pre-existing and environmental on this machine: three need a sibling `omarchy-pkgs` checkout that isn't present, one is a network-dependent theme URL check. None touch T2, suspend, or migrations.
- Verified on the affected hardware: MacBookPro16,1, Omarchy 4.0.1, `linux-t2` 7.1.8. After the change `/sys/power/mem_sleep` reads `[s2idle] deep` and the failing deep pass is gone.

## Scope of the evidence

I can only confirm this directly on a MacBookPro16,1. The other T2 models with discrete GPUs — MacBookPro15,1 / 15,3 / 16,4, iMacPro1,1, MacPro7,1 — should be affected by the same amdgpu failure, but I have not tested them. Happy to adjust the detection if a T2 owner reports otherwise.


jacobkjones:
Chiming in: 
MacBookPro16,1, Omarchy 4.0.2, linux-t2 7.1.8 t2bce, BCM4364, panel on AMD. `pm_async=off` already. Deep hung (forced off). systemd `MemorySleepMode=s2idle` still self-rebooted ~1 min later; no PM: suspend entry flushed, no pstore. Cmdline-only `apple_gmux.force_igd=1` muxed (Switching to IGD) then CATERR ~1 min after SDDM login.

Happy to retest a branch, would love a durable fix to get merged.

PR #9106, opened 2026-08-30
fix(battery): aggregate percentage and telemetry across dual-battery setups

### LINKED ISSUE #9026: Power panel percentage ignores second battery on dual-battery laptops
## What happened

On a dual-battery laptop (Lenovo ThinkPad, internal battery `BAT0` + swappable/slice battery `BAT1`), the power panel's percentage, size, cycles, time, and rate stats only reflect `BAT0` and silently ignore `BAT1`.

`omarchy battery status --shell` output:
```
percentage	78%
state	discharging
rate	9.1W
size	25Wh
time	2h 17m
cycles	17
threshold	75-80%
```

The true combined state (`upower -i /org/freedesktop/UPower/devices/DisplayDevice`):
```
state:       discharging
energy:      67.54 Wh
energy-full: 73.16 Wh
percentage:  92.3182%
```

So the panel shows **78% / 25Wh** when the real combined charge is **92% / 73Wh** — a large, user-visible discrepancy. Note the bar's battery *icon* and *progress-bar fill* are actually correct, because `Panel.qml` sources those from `UPower.displayDevice` (properly aggregated). Only the numeric stats are wrong, because they come from a separate script.

## Root cause

`/usr/share/omarchy/bin/omarchy-battery-status` picks a single battery like this:

```bash
battery=$(upower -e 2>/dev/null | grep BAT | head -n 1)
```

On a multi-battery system, `upower -e` lists all `battery_BAT*` devices, and `head -n 1` arbitrarily takes the first one (`BAT0`) rather than the combined `DisplayDevice` (or a proper sum across all present batteries). Everything downstream in the script (percentage, energy-full/size, cycles, rate, thresholds) is then computed from that single battery only.

## Expected

The power panel's percentage/size/cycles/time/rate should reflect the combined state of all present batteries, consistent with what the bar's icon/progress-bar already show via `UPower.displayDevice`.

## Suggested fix direction

`omarchy-battery-status` should query `upower -i $(upower -e | grep '/battery_DisplayDevice')` (or equivalent aggregation across all `battery_BAT*` devices) instead of `head -n 1` on the first battery, at least for percentage/energy/state. Cycle count and per-battery threshold could still reasonably be reported per-battery or summed, but the primary percentage must not silently drop a battery.

## Steps to reproduce

1. On a laptop with two batteries present under `/sys/class/power_supply/` (e.g. `BAT0` and `BAT1`), run `upower -e | grep BAT` — confirm both are listed.
2. Run `omarchy battery status --shell` and compare to `upower -i /org/freedesktop/UPower/devices/DisplayDevice`.
3. Observe the panel percentage/size/stats matches only the first-listed battery, not the combined total.

## System details

- Omarchy version: 4.0.1-1
- Kernel: 7.1.9-arch1-2
- Machine: Lenovo 20UROITORT (ThinkPad)
- CPU: Intel Core i7-8650U
- GPU: Intel UHD Graphics 620 (Kaby Lake-R)
- UPower: 1.91.3 (client and daemon)
- Batteries: BAT0 (LGC 01AV420, 24.51Wh design) + BAT1 (LGC 01AV491, 47.5Wh design)

harshithnadig:
### Problem
In #9026, on dual-battery laptops (such as ThinkPads with `BAT0` and `BAT1`), the power panel percentage, size, time, and rate stats only reflect `BAT0` and omit `BAT1`.

### Cause
`bin/omarchy-battery-status` selected `battery=$(upower -e | grep BAT | head -n 1)`, arbitrarily taking the first physical battery rather than the aggregated `DisplayDevice` or summing across all present batteries.

### Solution
1. Query UPower's aggregated `DisplayDevice` for combined percentage, full capacity, state, and time to empty/full.
2. Sum instantaneous sysfs power readings (`power_now`) and cycle counts across all present `BAT*` supplies.

Fixes #9026

### DIFF
--- a/bin/omarchy-battery-status
+++ b/bin/omarchy-battery-status
@@ -18,10 +18,20 @@ case "${1:-}" in
     ;;
 esac
 
-battery=$(upower -e 2>/dev/null | grep BAT | head -n 1)
-[[ -z $battery ]] && exit 0
+# Check for physical batteries
+batteries=($(upower -e 2>/dev/null | grep BAT))
+display_device=$(upower -e 2>/dev/null | grep -i 'DisplayDevice' | head -n 1)
 
-battery_info=$(upower -i "$battery")
+if (( ${#batteries[@]} == 0 )) && [[ -z $display_device ]]; then
+  exit 0
+fi
+
+# Prefer aggregated DisplayDevice if present and reporting a battery
+if [[ -n $display_device ]]; then
+  battery_info=$(upower -i "$display_device")
+else
+  battery_info=$(upower -i "${batteries[0]}")
+fi
 
 percentage=$(awk '/percentage/ { print int($2); exit }' <<<"$battery_info")
 capacity=$(awk '/energy-full:/ { printf "%d", $2; exit }' <<<"$battery_info")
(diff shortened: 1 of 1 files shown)


bjarneo:

PR #9119, opened 2026-08-30
Match Obsidian's dotted window class in the focus binding

### LINKED ISSUE #8902: Default Obsidian binding: focus = "^obsidian$" is anchored, so it can never match and SUPER+SHIFT+O never focuses an open window
### System details

Omarchy 4.0.0.alpha, Hyprland 0.56.2. Present at `main` as of today in
`default/hypr/bindings/applications.lua:18`.

### What's wrong?

The default Obsidian binding can never focus an already-open Obsidian window.
`SUPER + SHIFT + O` always takes the *launch* branch, Obsidian's single-instance
lock swallows the new process, the existing window never moves — and the key
looks dead.

```lua
o.bind("SUPER + SHIFT + O", "Obsidian", { launch = "obsidian", focus = "^obsidian$" })
```

`focus` is a **regex fragment, not a whole pattern**. `omarchy-launch-or-focus`
interpolates it:

```bash
test("\\b" + $p + "\\b";"i")
```

so `^obsidian$` becomes `\b^obsidian$\b`, which matches nothing, ever. The
pattern is also wrong on its own terms — Obsidian's class is
`md.obsidian.Obsidian` (its `StartupWMClass`), not `obsidian`.

### Steps to reproduce

1. Open Obsidian.
2. Focus any other window.
3. Press `SUPER + SHIFT + O`.

Nothing happens. Most visible with a fullscreen window up, where it reads as the
shortcut being broken rather than as a no-op.

Confirming the match fails, with Obsidian running:

```console
$ hyprctl clients -j | jq -r --arg p '^obsidian$' \
    '.[]|select((.class|test("\\b"+$p+"\\b";"i")))|.class'
                                      # (nothing)

$ hyprctl clients -j | jq -r --arg p 'md.obsidian.Obsidian' \
    '.[]|select((.class|test("\\b"+$p+"\\b";"i")))|.class'
md.obsidian.Obsidian
```

### Suggested fix

```diff
-o.bind("SUPER + SHIFT + O", "Obsidian", { launch = "obsidian", focus = "^obsidian$" })
+o.bind("SUPER + SHIFT + O", "Obsidian", { launch = "obsidian", focus = "md.obsidian.Obsidian" })
```

Verified on this machine: with the class pattern, focusing works and — because
`misc:on_focus_under_fullscreen = 1` is already a default — Obsidian correctly
takes over a fullscreen window on the same workspace.

### Why it is worth fixing rather than leaving to the app

It is tempting to assume Obsidian raises its own window and the binding is
redundant. It cannot, on Wayland: activating a window requires an
**xdg-activation token** passed from a real user interaction, and a process
spawned by `uwsm-app` from a keybinding has none, so the compositor is entitled
to ignore the request. The compositor has to be told explicitly, which is what
`omarchy-launch-or-focus` exists to do.

### Possibly worth a sweep

Anchored patterns fail silently, so they are easy to miss in review. It may be
worth grepping the defaults for `focus = "^`. Distinct from #4927 / #3669, which
are about `omarchy-launch-or-focus` matching too *broadly* (the `.title`
fallback) — this one is about a pattern that matches nothing at all. A pattern
that can never match would be cheap to reject at bind time.

ya-luotao:
## Problem

`SUPER + SHIFT + O` looks dead once Obsidian is running (#8902). The binding is `{ launch = "obsidian", focus = "^obsidian$" }`, which `omarchy-launch-or-focus` wraps as `\b^obsidian$\b` against each window's class and title. That matches a window whose class is exactly `obsidian`, but current Obsidian builds report the class `md.obsidian.Obsidian`, so the focus branch never matches, the script relaunches Obsidian every time, and Obsidian's single-instance lock swallows the new process without raising the existing window.

## Fix

Change the pattern to `^(md\.obsidian\.)?obsidian$` (Lua literal `"^(md\\.obsidian\\.)?obsidian$"`). It matches both the bare `obsidian` class and `md.obsidian.Obsidian`, case-insensitively, while keeping the anchors.

The anchors are deliberate: #1838 anchored this binding because a window titled "Obsidian is great!" stole the key, and ec593886 re-anchored it after the script went back to `\b` matching. Dropping them to a bare `obsidian` would fix #8902 but reintroduce that false match.

Fed through the exact jq expression from `bin/omarchy-launch-or-focus`:

| client | `^obsidian$` (old) | `^(md\.obsidian\.)?obsidian$` (new) |
|---|---|---|
| class `obsidian` | ✅ | ✅ |
| class `md.obsidian.Obsidian` | ❌ (the bug) | ✅ |
| firefox, title "Obsidian - Mozilla Firefox" | ❌ | ❌ |
| class `obsidianite` | ❌ | ❌ |
| class `org.obsidian.foo` | ❌ | ❌ |

## Verification

- jq matrix above against the expression in `bin/omarchy-launch-or-focus:13`.
- `applications.lua` is required from `$OMARCHY_PATH/default/hypr/bindings/` via `default.hypr.omarchy`, not copied into `~/.config`, so the change reaches existing installs on update; no migration needed.
- `test/shell.d/launch-1password-test.sh` and `test/shell.d/hyprland-default-config-test.sh` pass.
- Not exercised against a running Hyprland (developed on macOS). A quick `hyprctl clients -j | jq -r '.[].class' | grep -i obsidian` on an Omarchy box before/after would confirm the class value.

Fixes #8902

https://claude.ai/code/session_01RrGftF1S6Xymu7Z58TRd3M

### DIFF
--- a/default/hypr/bindings/applications.lua
+++ b/default/hypr/bindings/applications.lua
@@ -15,7 +15,7 @@ if o.preinstalled_bindings_enabled() then
   o.bind("SUPER + SHIFT + ALT + M", "Music TUI", { tui = "cliamp", focus = true })
   o.bind("SUPER + SHIFT + D", "Docker", { tui = "omarchy-launch-docker-tui" })
   o.bind("SUPER + SHIFT + G", "Signal", { omarchy = "signal" })
-  o.bind("SUPER + SHIFT + O", "Obsidian", { launch = "obsidian", focus = "^obsidian$" })
+  o.bind("SUPER + SHIFT + O", "Obsidian", { launch = "obsidian", focus = "^(md\\.obsidian\\.)?obsidian$" })
   o.bind("SUPER + SHIFT + W", "Omawrite", { launch = "omawrite" })
   o.bind("SUPER + SHIFT + SLASH", "Passwords", { omarchy = "1password" })
 


bjarneo:
Automated duplication check: this pull request looks similar to #7465, which covers the same Obsidian focus pattern. I keep that one open and close this one to consolidate review.

PR #9127, opened 2026-08-30
Add a notification for hyprpicker

Mrid22:
Replace the color picking action with a small script that adds a notification after you pick the color telling you what it is and that it has been copied to your clipboard. I opened a similar PR, but that was closed because it wasn't compatible with quattro. This one is.

### DIFF
--- a/bin/omarchy-capture-color
+++ b/bin/omarchy-capture-color
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+pkill hyprpicker || hyprpicker -a
+
+COLOR_HEX=$(wl-paste)
+
+omarchy-notification-send "Color copied" "$COLOR_HEX copied to clipboard"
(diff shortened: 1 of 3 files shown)


sprajs:
Cancelling an existing picker still sends “Color copied” with the previous clipboard contents. When `pkill hyprpicker` succeeds, the `||` branch skips `hyprpicker -a`, but the script then runs `wl-paste` and notifies unconditionally. I reproduced this with mocked commands: the picker was never launched, yet the stale clipboard text appeared in the notification.

Please return after stopping an existing picker, and only notify after a successful new selection. A cancellation-path test should assert that neither the clipboard nor the notification helper is used.


I implemented a fix with regression tests in https://github.com/omacom/omarchy/pull/12424, in the main Omarchy repository, preserving this PR’s original commits. The focused tests pass; the PR records the full-suite failures and remaining runtime validation.

PR #9128, opened 2026-08-30
Detect browser families without launching them

### LINKED ISSUE #8897: omarchy-launch-browser causes Brave SIGTRAP when man is unavailable
## Description

`omarchy-launch-browser` runs the selected browser with `--help` to detect
Mozilla-based browsers:

    if $browser_exec --help 2>/dev/null | grep -q MOZ_LOG; then

On Linux, Chromium-family browsers implement `--help` by executing `man`.
When `man` is unavailable, Brave calls PLOG(FATAL), raises SIGTRAP, and
systemd-coredump records a browser crash. Omarchy then launches the actual
browser successfully.

## Steps to reproduce

1. Use Brave as the default browser on a system without `man`.
2. Run `omarchy-launch-browser`.
3. Inspect `coredumpctl list brave`.

The failing command can be reproduced without writing a core:

    prlimit --core=0 -- /usr/bin/brave --help

Output:

    [FATAL:chrome/app/chrome_main_delegate.cc:404] execlp failed:
    No such file or directory (2)

Exit status: 133.

## Expected behavior

The browser launches without executing a probe that crashes or creates a
systemd-coredump entry.

## Actual behavior

Each launch creates a Brave SIGTRAP coredump from the `--help` probe. The real
browser starts afterward.

Three identical crashes were recorded on the same boot. Each command line ended
in `--help`, and each had the same stack offsets.

## Versions

Omarchy: 4.0.1-1
Brave package: brave-bin 1:1.94.117-1
Architecture: aarch64

## Suggested direction

Identify Mozilla-based browsers without executing the browser's `--help`
handler.

Filed by openai-codex/gpt-5.6-sol via Oh My Pi coding harness.

yashranaway:
Stops browser launches from probing Chromium-family binaries with `--help`. Fixes #8897.

<<<< AI wording below >>>>

## Problem

Chromium-family browsers may execute `man` for `--help`. Brave raises SIGTRAP when `man` is unavailable, so every normal launch leaves a coredump before starting the real browser.

## Fix

Choose the private-window flag from the resolved executable name instead of running the browser. Firefox-family, Edge, and Chromium-family browsers retain their existing flags.

## Verification

- `bash test/shell.d/launch-browser-test.sh`
- `bash -n bin/omarchy-launch-browser test/shell.d/launch-browser-test.sh`
- `git diff --check`
- The regression stubs Chromium and Firefox and fails if family detection executes either binary.


TANJlD:
Confirmed on Omarchy 4.0.2-1 with Google Chrome 152. Directly binding google-chrome-stable fixes it.

PR #9134, opened 2026-08-30
Constrain Omawrite's save dialog

### LINKED ISSUE #9046: Omawrite Save File dialog exceeds 2x logical display bounds
## System details

  Apple MacBook Air (Mac14,2 / M2), Asahi aarch64, Omarchy 4.0.1-2, Hyprland 0.56.1, Omawrite 0.5.0-1,
  Qt 6.11.2, GTK 3.24.52, xdg-desktop-portal-gtk 1.15.3-1. Built-in eDP-1 display at 2560x1664 with
  scale 2 (1280x832 logical pixels).

  ## What's wrong?

  Omawrite's **Save File** dialog opens larger than the logical display and is partly off-screen at 2x
  scaling. The title/upper content and lower action area are clipped, making the dialog difficult to
  use.

  ### Steps to reproduce

  1. Use the built-in 2560x1664 display with Hyprland scale 2.
  2. Open Omawrite and type into a new document.
  3. Press `Ctrl+S` to open **Save File**.
  4. Observe that the dialog extends beyond the top, right, and bottom of the display.

  ### Measured result

  - Logical display: 1280x832, with 28 pixels reserved at the top.
  - Save File window: 1231x950 at `(342, -45)`.
  - Right edge: 1573, or 293 logical pixels beyond the display.
  - Bottom edge: 905, while the display ends at 832.
  - Window class/title: `omawrite` / `Save File`.
  - The dialog PID is the Omawrite PID, rather than `xdg-desktop-portal-gtk`.

  GTK stored `org.gtk.Settings.FileChooser window-size` as `(1231, 903)`. Resetting it to the schema
  default `(-1, -1)` and restarting the portal services did not help; the same size was recreated
  immediately. Testing `GDK_SCALE=1` first on the GTK portal and then on a separate Omawrite process
  also reproduced the same 1231x950 dialog.

  ### Expected result

  The Save File dialog should fit within the current monitor's logical work area, with the filename,
  Cancel, Save, and file-type controls visible.

  ### Verified local workaround

  On this display, the following Hyprland 0.56 Lua rule constrains only Omawrite's Save File window:

  ```lua
  o.window(
    { class = "^omawrite$", title = "^Save File$" },
    { float = true, center = true, size = { 875, 600 } }
  )
  ```

  With the rule, the dialog consistently opens at 875x600 at `(203, 130)` and all controls are
  visible.

  This rule is only a local workaround for the reported display configuration. The underlying cause
  has not yet been established, and a fixed compositor-level size may not be appropriate across
  different monitors and scale factors.

  An earlier revision of this report included a `Main.qml:327` QML binding-loop warning. That warning
  originates from Omawrite's keyboard-shortcuts dialog and is unrelated to the Save File sizing
  problem.

  Possibly related, but distinct: #7021 (`GDK_SCALE` overflow) and #7944 (GTK file-chooser portal
  selection).

Image

Image

yashranaway:
Keeps Omawrite's save dialog inside scaled laptop displays. Fixes #9046.

<<<< AI wording below >>>>

## Problem

Omawrite can create a 1231x950 save dialog on a 1280x832 logical display, clipping its title and action controls beyond the monitor work area.

## Fix

Match only Omawrite's `Save File` window and apply the reporter-validated floating, centered 875x600 geometry.

## Verification

- `bash -n test/shell.d/omawrite-window-rule-test.sh`
- `git diff --check`
- The semantic Lua test checks the emitted class, title, float, center, and size properties.
- The focused test cannot run here because `lua` is not installed.
- Live verification at 2x scale is pending, so this PR is a draft.

### DIFF
--- a/default/hypr/apps/omawrite.lua
+++ b/default/hypr/apps/omawrite.lua
@@ -1,0 +1,4 @@
+o.window(
+  { class = "^omawrite$", title = "^Save File$" },
+  { float = true, center = true, size = { 875, 600 } }
+)
--- a/test/shell.d/omawrite-window-rule-test.sh
+++ b/test/shell.d/omawrite-window-rule-test.sh
@@ -1,0 +1,33 @@
+#!/bin/bash
+
+set -euo pipefail
+
+source "$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd)/base-test.sh"
+
+require_command lua
+
+rule=$(OMARCHY_PATH="$ROOT" lua <<'LUA'
+package.path = os.getenv("OMARCHY_PATH") .. "/?.lua;" .. package.path
+
+local emitted
+hl = { window_rule = function(value) emitted = value end }
+
+require("default.hypr.helpers")
+require("default.hypr.apps.omawrite")
+
+print(table.concat({
+  emitted.match.class,
+  emitted.match.title,
+  tostring(emitted.float),
+  tostring(emitted.center),
+  emitted.size[1] .. "x" .. emitted.size[2],
+}, "\t"))
+LUA
+)
+
+IFS=$'\t' read -r class title float center size <<<"$rule"
+[[ $class == "^omawrite$" && $title == "^Save File$" ]] ||
+  fail "the window rule targets only Omawrite's save dialog" "$rule"
+[[ $float == "true" && $center == "true" ]] || fail "the save dialog floats and centers" "$rule"
+[[ $size == "875x600" ]] || fail "the save dialog fits a scaled laptop display" "$rule"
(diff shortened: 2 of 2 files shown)


yuzuwoeser:
Thanks for looking into this. The 875x600 rule was only a workaround tested on my display, so I’m unsure it should become the default fix for everyone. Treat this as a mitigation and investigate the sizing issue in Omawrite/Qt instead?

yashranaway:
Agreed, 875x600 is too display-specific, so I am moving this fix into Omawrite's file-dialog implementation.

yashranaway:
Replaced by omacom/omawrite#39, which fixes the dialog in Omawrite instead of applying a display-specific Hyprland rule.

PR #9140, opened 2026-08-30
Center the Quake console in a 2:1 panel instead of spanning the screen.

ogarza:
## Summary
- The Quake console is a centered half-height panel, twice as wide as it is tall (`box = 2` next to `share = 0.5`), sized by leftover workspace gaps so it refits when the monitor or scale changes.
- One tiled window keeps that panel. A second tiled app gets the full width back. A floating window on top (btop, Super+O, etc.) does not count.
- It sizes to the output it actually opens on. A `follow_mouse` hop onto another screen does not steal the gaps. Quarter-turned monitors swap the work area so the panel is not a sliver.
- Membership is chased while the console is on screen: `window.open`, `window.destroy`, `window.move_to_workspace` (Super+Shift+1 / Super+Alt+S), and `window.update_rules` (Super+T / Super+O). `window.close` is not hooked; it still counts the window on its way out.

## Test plan
- [x] `./test/shell.d/hyprland-qconsole-test.sh`
- [x] Toggle the scratchpad on a 16:9 display: centered 2:1 panel, not full width.
- [x] Same on an ultrawide / mixed-DPI dock: panel stays 2:1 in logical pixels.
- [x] Open on one monitor, then open on another with a different aspect ratio without closing it: panel stays 2:1 for that output.
- [x] With the console open, Super+Alt+S a second tiled window onto it: full width. Super+Shift+1 that window off: panel again.
- [x] Super+Ctrl+T (btop) while the console is down: panel stays. Super+T that float: full width. Super+T again: panel.
- [x] Super+O a tiled window on a two-app console: panel re centers.

### DIFF
--- a/config/hypr/hyprland.lua
+++ b/config/hypr/hyprland.lua
@@ -10,6 +10,11 @@ dofile((os.getenv("OMARCHY_PATH") or "/usr/share/omarchy") .. "/default/hypr/boo
 -- keeping core window-manager bindings:
 -- omarchy_preinstalled_bindings = false
 
+-- Center the quake console in a tiling box. The number is how many times
+-- wider the box is than its height. 1 is a square. Comment this out for a
+-- full-width drop-down.
+omarchy_qconsole_ratio = 2
+
 -- Load Omarchy defaults.
 require("default.hypr.omarchy")
 
--- a/config/hypr/hyprland.lua
+++ b/config/hypr/hyprland.lua
@@ -10,11 +10,6 @@ dofile((os.getenv("OMARCHY_PATH") or "/usr/share/omarchy") .. "/default/hypr/boo
 -- keeping core window-manager bindings:
 -- omarchy_preinstalled_bindings = false
 
--- Center the quake console in a tiling box. The number is how many times
--- wider the box is than its height. 1 is a square. Comment this out for a
--- full-width drop-down.
-omarchy_qconsole_ratio = 2
-
 -- Load Omarchy defaults.
 require("default.hypr.omarchy")
 
--- a/default/hypr/qconsole.lua
+++ b/default/hypr/qconsole.lua
@@ -4,7 +4,12 @@
 
 -- How much of the usable screen the console covers, measured from the top.
 local share = 0.5
-local min_size = 64
+
+-- A console holding a single window is boxed into a centered panel this many
+-- times wider than it is tall, rather than stretched the width of the screen.
+-- A second app on the scratchpad gets the full width back: two windows splitting
+-- a half-width column is worse than the band this replaced.
+local box = 2
 
 local SCRATCHPAD = "special:scratchpad"
 
@@ -26,35 +31,25 @@ hl.config({
   },
 })
 
+-- The panel is always flush with the top and always centered, so two numbers
+-- describe it: the gap down each side and the gap underneath.
+--
 -- Refitting replaces the rule in place rather than stacking a new one, but it
 -- still schedules a monitor and window state refresh, and monitor.focused fires
 -- on every hop between screens. Most of those hops do not change the gaps, so
 -- only write the rule when it actually moves.
-local covering = nil
+local beside, below = nil, nil
 
-local function same_gaps(a, b)
-  return a
-    and b
-    and a.top == b.top
-    and a.right == b.right
-    and a.bottom == b.bottom
-    and a.left == b.left
-end
-
-local function cover(gaps_out)
-  if type(gaps_out) == "number" then
-    gaps_out = { top = 0, right = 0, bottom = gaps_out, left = 0 }
-  end
-
-  if same_gaps(covering, gaps_out) then
+local function cover(side, bottom)
+  if beside == side and below == bottom then
     return false
   end
-  covering = gaps_out
+  beside, below = side, bottom
 
   hl.workspace_rule({
     workspace = SCRATCHPAD,
     gaps_in = 0,
-    gaps_out = gaps_out,
+    gaps_out = { top = 0, right = side, bottom = bottom, left = side },
 
     -- Nothing to highlight in a console that is only ever focused when it is
     -- open, and the active border reads as a stray frame around a panel that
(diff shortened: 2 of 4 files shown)


ogarza:
I just tested these changes you pushed and it is much improved!

Love that opening a second window just expands it back.

ogarza:
One thing I noticed is if I have the console open with only the agent, it is centered.. if I open btop via SUPER + CTRL + T, I would expect it to not change the agent window, and I probably want to keep it on the bottom or drag it off to the side of the scratchpad.

changing lines 67-70 of `qconsole.lua` to:
```
local function alone()
  local ws = hl.get_workspace(SCRATCHPAD)
  return not ws or #hl.get_windows({ workspace = SCRATCHPAD, floating = false }) <= 1
end
```
Fixes this.

image

vs

image

ogarza:
Calling this as: count only tiled windows.

A floating window on the scratchpad is not a second app in the layout sense — it overlays the panel, it does not split it — so alone() now uses hl.get_windows({ workspace = scratchpad, floating = false }). Opening Activity while the console is down keeps the centered panel; a second tiled app still gets the full width back.

Toggling floating is not window.open / window.destroy, so tiling that float while the console is showing used to keep the old gaps until the console was closed. The float toggle and omarchy-hyprland-window-pop call o.qconsole_recount after they land.

ogarza:
The last commit skipped window.move_to_workspace because ws.windows was stale there, then we switched the count to get_windows() and never went back to check. So Super+Shift+1 / Super+Alt+S with the console already up didn't stretch or rebox — you had to close it and open it again.

Turns out moveToWorkspace writes m_workspace before it emits, so get_windows() is already right by then. Recount now hangs off that, plus window.update_rules for Super+T / Super+O (float toggle re-applies rules; Lua still has no window.floating). Dropped o.qconsole_recount from tiling.lua and window-pop. Those files don't need to know about the console.

ogarza:
Well this was much harder than I thought, but I've tested thoroughly with the two monitors I have, which are very different in size and it behaves nicely and feels polished, it did open up a can of worms where a lot of testing and math was needed so if this seems like too much to maintain, we could merge [a59afb9659ca0d2a33316d510fac70755d9752f1](https://github.com/omacom/omarchy/pull/9140/commits/a59afb9659ca0d2a33316d510fac70755d9752f1) instead.

## What you get

Super+S (or Super+`) drops the scratchpad down over whatever you're working on, Quake style. This PR makes it behave like a console instead of a full-width band:

- With one app in it, it's a centered panel — half the usable height, twice as wide as it is tall. Open a second tiled app and it gives the full width back, since at that point it's a workspace, not a console.
- Anything you float in there keeps its spot **relative to the edges** no matter which monitor you open it on. A corner stays a corner, centered stays centered.

So you can park three btop windows in the corners of the console on a 6K monitor, hit Super+S over on the 1080p, and they're still in the corners. Same across different scales, aspect ratios, and rotated outputs.

screenshot-2026-08-31_01-25-51
screenshot-2026-08-31_01-26-03

## What we did

Three things worth calling out, because none of them are the obvious approach.

**The panel is sized by its gaps, not by a window rule.** Hyprland resolves window-rule size expressions once, when the window maps. Size the console that way and rescaling the monitor afterwards leaves you with a console that's no longer half of anything. Instead the workspace rule sets the gaps around it and the layout re-applies them, so the console is whatever's left over — recomputed every time the monitor it's on changes.

**Floating windows get remapped, not just shifted.** Hyprland moves floats across a special-workspace hop with a plain origin shift — it slides them by the difference between the two monitors' positions and nothing else. Go from a 6144-wide screen to a 1920-wide one and anything past the first 1920px is simply off-screen. So we remap them ourselves: work out where each window sat in the old work area as a fraction of the room it had to slide, and put it at the same fraction of the new one. That's what makes an edge stay an edge when the two screens aren't the same shape.

**The reference monitor is recovered from the workspace when the config loads.** This is the actual bug that started this. Changing a monitor's scale rewrites `~/.config/hypr/monitors.lua`, Hyprland watches that file and reloads, and `bootstrap.lua` drops every `default.hypr.*` module from `package.loaded` — so `qconsole.lua` re-runs with the floats still sitting there but no memory of which monitor their coordinates belong to. The next hop had nothing to map from and left them wherever the origin shift dropped them. It only ever happened once, right after a scale change, which is what made it confusing. Now the module reads the output back off the scratchpad workspace on load.

The last commit is a cleanup pass with no behaviour change. `work_area` and `contains` each did their own transform swap and scale division, with a comment on `contains` promising it matched `work_area` — a claim the code couldn't enforce; both now take the output's logical box from one place. The three near-identical "work out which output the console is on, then remap and refit" handler bodies became one `reconcile()`, the re-resolve-by-name block that appeared twice became `freshen()`, and `visible_console_monitor` folded into `shown_monitor`, which took `console_monitor` from nine lines to one. The file is now grouped under headings: geometry, the panel, which output, where the floats were last placed, moving the floats, reacting.

## What was tested

`test/shell.d/hyprland-qconsole-test.sh`, 103 assertions against a mock of the Hyprland Lua API.

To check the tests were worth their length, we seeded plausible bugs into `qconsole.lua` — wrong scale arithmetic, dropped clamp, normalising by the work area instead of by the window's travel, the reload fix removed, and so on — and kept only assertions needed to catch them. The original 129 assertions caught 24 of 35 seeded bugs, and a greedy pass showed 119 of them could be deleted without losing a single one. The suite is now 103 assertions catching 34 of 41. What went were tautologies (`gaps are never negative` sitting under an assertion pinning them to exactly `0, 0, 960`), checks that only tested the mock, and four-corner blocks repeated six times over — x and y are mapped independently, so two opposite corners already pin both ends of both axes.

That exercise also turned up two real gaps, both now covered. `place_from_snap(last_mon)` in the layout handler was completely untested, because every scale test had the console *visible* — but a hidden console is this bug's own shape: nothing reports where it is, and the remembered output is the only handle left. And the decision that a float too wide for the destination keeps its *left* edge on screen rather than its right was silent, which reads like something a future reader would "simplify" into a plain `math.min(math.max(...))` and hang the window off the left of the screen.

For the cleanup commit, a differential harness drove the old and new files through the same randomised event sequences — four outputs including a fractional-scale laptop panel and a rotated one with side reserved area, random scale changes, hops, origin shifts, monitor tagging, window count changes — comparing every rule written and every float coordinate at every step. **1200 out of 1200 scenarios identical.**

Then live, on a 6144×2560 DP-1 next to a 1920×1080 HDMI-A-1, with three btop windows parked in the corners. Scaled DP-1 to 1.3333, waited for the reload, opened the console on the HDMI:

| window | before, on DP-1 | after the hop, on HDMI |
|---|---|---|
| top-left | 0.000, 0.000 | 0.000, 0.000 |
| bottom-right | 0.983, 0.970 | 0.982, 0.969 |
| bottom-left | 0.001, 0.986 | 0.001, 0.985 |

Those are fractions of the work area, 0 being flush with one edge and 1 flush with the other. The 0.001 is rounding to whole pixels. Before the fix the same hop gave 0.000, 1.518 and 3.036 — two of the three completely off-screen. After the cleanup commit, `hyprctl reload` plus another live hop, to catch any load-order slip the preloaded-module tests can't see.

## Found but not fixed

**`omarchy-hyprland-monitor-scaling` rescales every monitor, not just the focused one.** It applies the scale to the focused output via `hyprctl eval`, but persists it into the catch-all `hl.monitor({ output = "", ... })` in `monitors.lua` — so the reload a second later hands that scale to everything. You can see it in the run above: only DP-1 was scaled, and HDMI-A-1 came back at 1.3333334 too. Separate bug in that command, left alone here.

**Positions drift by about a pixel per hop.** Each remap reads the fraction back out of whole-pixel coordinates, so it loses a little each time — roughly 4px over a round trip on the 6K. Only fixable by remembering each window's fraction instead of re-deriving it, which is a bigger change than this needed.

**A window too big for one of the screens gets pinned to a corner there and doesn't recover.** If it can't fit there's nowhere else to put it, but the original position is lost rather than restored on the way back. They will just move to top-left.

**Right after a reload, a float can briefly land under the bar.** The bar's reserved area reads as zero for a moment while things settle, so a remap in that window measures against the full screen. It corrects itself on the next layout change — we watched it do so — but it's a real transient.

**Seven defensive paths have no test coverage,** and we checked each rather than assuming: `freshen`'s re-resolution (needs Hyprland to hand back a stale userdata; the mock returns the same table), `logical_box`'s `scale <= 0` guard, `map_along`'s and `place_floats`' degenerate-work-area guards, `work_area`'s left/right reserved (Omarchy's bar is on top, so it's always zero), `scratchpad_monitor`'s loop, and `float_output`'s name check — that last one is inert rather than uncaught, since removing it returns the snapshot's own output, which `place_from_snap` short-circuits anyway. We tried teaching the mock to reach `scratchpad_monitor` and it changed nothing, so those twelve lines of scaffolding got thrown away rather than kept for no coverage.

One incidental note for whoever touches this next: `same_geometry` deliberately ignores the monitor name, so a hop between two outputs at the same position with the same size is a no-op. That's only reachable with mirrored outputs, where the coordinates really are identical, so it's correct.

dhh:
Pushed d377569e on top, closing two gaps between the description and the code:

- **Only tiled windows count.** `alone()` read `ws.windows`, so a floating btop on the console blew it out to full width. The gaps only lay out tiled windows, so it now counts non-floating windows via `hl.get_workspace_windows`.
- **Moves and float toggles refit.** Only `window.open`/`window.destroy` were hooked, so Super+Alt+S a second app onto the console left both squeezed into the panel, and moving it off again left the console full width. `window.move_to_workspace` and `window.update_rules` are now hooked too. Measured on a live Hyprland 0.56.2 with the same dispatcher Super+Alt+S uses: `move_to_workspace` already carries the settled count, and every move and float toggle is trailed by several `update_rules`, the last of which always does. The earlier ones refit to the rule that is already there, which `cover()` skips.

Tests cover float-on-console, tile/float toggles, and moves on and off; each of the three fixes was reverted in turn and the suite failed every time. All 11 hyprland tests pass on the branch merged into current quattro, and loading the merged `qconsole.lua` into a live session applied gaps of `[0, 671, 833, 671]` on a 3008x1692 display with a 26px bar — the expected centered 1666x833 panel.

PR #9150, opened 2026-08-30
Harden tui launcher install and removal

rihadroshan:
Port webapp-install's name and desktop-value guards to tui-install.

Problem. omarchy-tui-install wrote its .desktop with raw values. A / in the app name became directory levels (unreachable by omarchy-tui-remove, and ../ escaped ~/.local/share/applications), and a newline could inject a second Exec= key. Its sibling omarchy-webapp-install already solves both with require_plain_name + desktop_string_escape.

Fix.

    bin/omarchy-tui-install: add require_plain_name (reject /) before any launcher/icon work, add desktop_string_escape for Name/Comment/Exec/Icon.
    bin/omarchy-tui-remove: index launchers and delete the found file (like webapp-remove), so legacy nested entries are reachable.
    test/shell.d/tui-install-hardening-test.sh: 7 assertions covering slash/../ rejection, early interactive refusal, newline injection, round-trip, nested legacy removal, quiet no-dir removal.

Verified: new test 7/7 pass; webapp-name, webapp-install-escaping, webapp-install, launcher-remove, desktop-entry-launch, refresh-config all green.


ErikMelton:
One issue: Explicit removal can delete the wrong launcher in omarchy-tui-remove, lines 54–55. The new lookup selects the first matching basename in the recursive scan. If both applications/Docker.desktop and applications/legacy/Docker.desktop exist, omarchy-tui-remove Docker can delete the legacy entry while leaving the requested top-level entry installed. This also affects removal through the app launcher.

ralyodio:
should we think about using hqtui for this?

PR #9152, opened 2026-08-30
Leave a locked session's display to the lock screen

Pillumz:
## What happens

Lock the session and let the lock screen blank the display. Later the panel comes back on at the lock screen with nobody at the machine, and stays on until someone touches the keyboard or mouse - on an OLED, that is the lock screen burning in for hours. `omarchy debug idle` shows a single `idle-monitor: active` → `idle-cycle-cancel: activity` → `process-start: wake omarchy-system-wake` with no input anywhere around it.

## Why

The idle service treats its `IdleMonitor` resuming as user activity and runs `omarchy-system-wake`, which turns DPMS back on. But the monitor is `respectInhibitors: true`, and Hyprland's ext-idle-notify implementation sends `resumed` to an idled notification the moment an idle inhibitor appears (`src/protocols/IdleNotify.cpp`: `update()` while inhibited calls `reset()`, which sends `resumed`). So any client taking an idle inhibitor - a browser starting media on a background workspace, a wake lock, an XWayland app suspending the screensaver - counts as activity with nobody there.

Unlocked that is harmless. Locked it is the whole bug: the lock screen only re-arms its blank timer for pointer, click or key events delivered to its own surface, and none happened, so the display stays lit indefinitely.

Reproduced on Hyprland 0.56.2 / Omarchy 4.0.1 with nobody at the desk: lock, wait for `idle-monitor: idle`, then start a silent `mpv` clip into the locked session from SSH. Hyprland logs `New idle inhibitor registered`, the idle service logs `active` and runs the wake in the same second, the panel lights, and it is still lit after mpv exits. An evdev witness on every input device saw nothing throughout. The same run *before* the monitor idles does nothing, as the Hyprland code predicts.

## Fix

While the session is locked, the display belongs to the lock screen, so `cancelIdleCycle()` no longer runs `omarchy-system-wake`; it logs `wake-skipped: session-locked` instead and still ends the cycle. The decision lives in `IdleModel.wakeAfterIdle()` so it is unit-testable, and `sessionLocked` is added to `omarchy-shell idle status` since `omarchy debug idle` prints that and not the lock's status.

The lock state comes from the lock service in-process, `shell.firstPartyServiceFor("omarchy.lock")`, the same call the Stay Awake, DND, night-light and media widgets use for their services. The `omarchy-shell lock isLocked` shell-out on the screensaver line answers from the same `locked` property, but only after a subprocess round trip, and it reads as unlocked whenever the IPC call fails; a synchronous decision needs the direct read. The idle service never blanks a display itself - the only DPMS-off in the tree is the lock's own blank timer - so skipping its wake cannot strand a display that nothing else can restore: Hyprland's `key_press_enables_dpms` / `mouse_move_enables_dpms` light the panel on keys, motion and gestures, the lock surface's own click and key handlers run the lock's wake, and `finishUnlock` runs it unconditionally (fingerprint unlock at a blanked panel included). Clamshell reconciliation, which `omarchy-system-wake` also does, has the lid switch bind, `omarchy-hyprland-monitor-watch` and the sleep lock as independent owners.

Why this layer: making the lock screen re-blank on a resume it did not ask for would still flash the panel; refusing inside `omarchy-system-wake` while locked would break the lock's own wake, which runs the same command; the Hyprland behaviour is protocol-legal and not ours to change; and `respectInhibitors: false` would bring the screensaver back over video playback.

One deliberate consequence: Hyprland does not enable DPMS for scroll (`axis`) or bare button events, only for keys, motion and gesture starts, so a scroll-wheel nudge at a blanked lock screen used to light the panel *only* through this accidental idle wake (and then never blanked again). After this change it does nothing; a click or any pointer motion still wakes. If a scroll wake is wanted, it belongs in `LockView` as a wheel handler that emits `wakeRequested`, which I am happy to follow up with.

#7131 adds a byte-identical `lockService` line with its own comment and a `lockInFlight` property; #7132 reads the same service through `firstPartyServiceFor` inside a function. Whichever lands second conflicts in the property block and in `statusJson`, and the two names should collapse into one.

## Tests

- `test/shell.d/idle-test.sh`: unit cases for `wakeAfterIdle`, plus source assertions that keep the wiring honest on machines where the compositor fixture has to skip.
- `test/shell.d/idle-lock-wake-test.sh` (new, compositor-backed like the lock fixtures): loads the real `Service.qml` into a throwaway quickshell with a stand-in shell whose services register by reassignment, as `shell.qml` does. The lock service is registered *after* the idle service and the fixture checks `sessionLocked` follows, so a future `serviceFor` that caches would fail here rather than silently disable the fix. Then it drives `handleActiveSignal()` twice: locked, no wake process is spawned and the skip is logged; unlocked, the wake process spawns and a PATH-shadowed `omarchy-system-wake` actually runs (marker file). The fake config uses day-long timeouts and the service is destroyed once the wake exits, so an orphaned fixture can never lock the developer's session; the test skips if the login shell would not resolve the shadow (a machine dev-linked to another checkout); the launch is under `timeout 15`. Against the unpatched service the fixture fails on all eight assertions.
- `./test/shell`: the same six pre-existing environmental failures with and without this change (packaging checkout, Claude usage collector, network guard, bar icon geometry); everything else passes.

### DIFF
--- a/shell/plugins/services/idle/IdleModel.js
+++ b/shell/plugins/services/idle/IdleModel.js
@@ -12,6 +12,12 @@ function eventParts(event, count) {
   return String(event && event.data ? event.data : "").split(",")
 }
 
+// The idle service wakes the display only for a cycle it put to sleep, and
+// never while the session is locked: the lock screen owns the display then.
+function wakeAfterIdle(idledThisCycle, sessionLocked) {
+  return !!idledThisCycle && !sessionLocked
+}
+
 function screensaverWindowsAfter(windows, address, visible) {
   var key = String(address || "")
   if (!key) {
@@ -47,6 +53,7 @@ if (typeof module !== "undefined") {
   module.exports = {
     secondsFromConfig: secondsFromConfig,
     eventParts: eventParts,
+    wakeAfterIdle: wakeAfterIdle,
     screensaverWindowsAfter: screensaverWindowsAfter
   }
 }
--- a/shell/plugins/services/idle/Service.qml
+++ b/shell/plugins/services/idle/Service.qml
@@ -25,6 +25,12 @@ Item {
   readonly property bool idleEnabled: stayAwakeStateLoaded && !stayAwake
   readonly property string screensaverClass: "org.omarchy.screensaver"
 
+  // The lock plugin is a service in this same process, so its state is readable
+  // directly: no subprocess round trip, and no reading as unlocked when an IPC
+  // call fails.
+  readonly property var lockService: shell && shell.firstPartyServiceFor ? shell.firstPartyServiceFor("omarchy.lock") : null
+  readonly property bool sessionLocked: !!(lockService && lockService.locked)
+
   property bool stayAwake: false
   property bool stayAwakeStateLoaded: false
   property bool hasPendingStayAwakePersist: false
@@ -103,7 +109,14 @@ Item {
     lockTimer.stop()
     screensaverLaunchGraceTimer.stop()
 
-    if (root.idledThisCycle) runProcess(wakeProcess, "wake", "omarchy-system-wake")
+    // A locked session's display belongs to the lock screen: it lights the
+    // panel for input that reaches it and blanks it again afterwards. The idle
+    // monitor resuming is not evidence of input while locked - Hyprland also
+    // reports activity when an idle inhibitor appears, with nobody at the
+    // machine - and waking on that lights the display with nothing left to
+    // blank it again.
+    if (IdleModel.wakeAfterIdle(root.idledThisCycle, root.sessionLocked)) runProcess(wakeProcess, "wake", "omarchy-system-wake")
+    else if (root.idledThisCycle) logEvent("wake-skipped", "session-locked")
 
     root.idledThisCycle = false
     root.screensaverStartedThisCycle = false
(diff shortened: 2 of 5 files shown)


v1b3m:
One gap in the "cannot strand a display" enumeration, from a clamshell machine - though having tested it, the gap argues *for* this PR rather than against it.

The list of independent owners assumes the output is still there to be lit. In clamshell (lid shut, external only, `eDP-1` disabled) after a blank, `HDMI-A-2` goes to `w:0 h:0` while the kernel still reports the connector `connected` with 35 modes available. A 2-second recorder trace of one blank:

```
08:09:36  HDMI-A-2  2560x1440  dpms:false    blanked
08:09:38  FALLBACK, HDMI-A-2 gone            the monitor drops its link
08:09:44  HDMI-A-2  w:0 h:0  dpms:true  lock=true/false
08:09:44  .. 08:10:25+   stuck at w:0 h:0, locked but never secure
```

Then:

- `realScreenCount()` is 0 (`plugins/lock/Service.qml:40-50`), so the lock never secures and the lock surface never exists - its click and key handlers and `finishUnlock` are unreachable rather than merely delayed.
- `key_press_enables_dpms` and `mouse_move_enables_dpms` have no mode to drive, so a keypress lights nothing.
- The clamshell reconciler can only `hyprctl reload`, and per the comment in `omarchy-hyprland-monitor-watch` a reload while the monitor is unpowered leaves it at 0x0.

So the owners that remain are exactly the ones this PR removes.

What that does **not** mean is that keeping the idle service's wake would rescue this state. `dpms:true` and `w:0 h:0` hold at the same time - the output is enabled with no mode to drive, so any DPMS-restoring wake is a no-op. The panel is dark for want of a mode, not for want of DPMS. If anything that argues for the PR: the owner being removed is not the owner that matters here.

What this state wants is an owner of its own - something that re-applies a mode (`hl.monitor(...)` via `hyprctl eval`) to an output reporting `0x0`, where `recover_modeless` only reloads. Flagging it because the argument as written reads as covering all cases, and this is one where none of the listed owners apply.

PR #9154, opened 2026-08-30
Tray: hide the drawer chevron when the drawer is empty

synjan:
## What

Six layout expressions in `shell/plugins/bar/widgets/Tray.qml` key on `root.allItems.length > 0` where they mean "is there anything in the drawer": `drawerBlockWidth`/`drawerBlockHeight`, the chevron's `visible`, and the `leftPadding`/`topPadding` join gap. `drawerCount` (`drawerItems.length`) already exists and is what the bar-level `visible: pinnedItems.length > 0 || drawerCount > 0` uses; this switches the six sites to it.

## Why

With every tray item pinned (`"pinned": ["1Password_status_icon_1"]`, nothing hidden) the drawer is empty, but the widget still reserves the drawer block and draws a chevron that opens onto an empty popup. After this change the chevron only appears once something actually lives in the drawer.

## Test

- All items pinned, drawer empty: no chevron, no reserved gap, pinned icons sit flush.
- One item unpinned: chevron and gap return, drawer opens with that item.
- Vertical bar: same via the `drawerBlockHeight`/`topPadding` twins.

https://claude.ai/code/session_01X2vGBcjauThiQrttbet9ZB

### DIFF
--- a/shell/plugins/bar/widgets/Tray.qml
+++ b/shell/plugins/bar/widgets/Tray.qml
@@ -232,7 +232,9 @@ BarWidget {
       id: horizontalTrayRoot
 
       readonly property int pinnedWidth: pinnedRow.implicitWidth
-      readonly property int drawerBlockWidth: root.allItems.length > 0 ? expandIcon.implicitWidth + root.drawerExtent : 0
+      // Only reserve the drawer block while something is actually in the drawer;
+      // an empty drawer would otherwise leave a chevron with nothing behind it.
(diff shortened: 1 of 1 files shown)


CtByte:
I have one small comment to this: how do you unpin an item if the chevron is hidden? Is there still a way to do so without a chevron?

synjan:
You're right — with everything pinned the chevron is the only way into the manage popup, so hiding it leaves no way to unpin. Closing; thanks for the review.

PR #9160, opened 2026-08-30
GPD Pocket 4: screen rotation, autorotation, and boot orientation

ekollof:
Implements the `omarchy hw gpd pocket 4` integration proposed in #9032. All pieces are verified working on real hardware (GPD Pocket 4, kernel 7.1.9-arch1-2 / Hyprland 0.56.2) and stay inert on every other machine.

The 8.8" 1600x2560 panel is physically portrait-mounted and used landscape (transform 3), so orientation has to be fixed at three layers:

## Session (autorotation)

- `bin/omarchy-hw-gpd-pocket-4-rotate`: watches `AccelerometerOrientation` on iio-sensor-proxy and applies matching monitor + touchscreen transforms via `hyprctl eval`. The touch device transform is always set explicitly alongside the monitor, because output-binding inheritance is not reliable (hyprwm/Hyprland#3538).
- `default/systemd/user/omarchy-gpd-pocket-4-rotate.service`: gated with `ExecCondition=/usr/bin/omarchy-hw-gpd-pocket-4` and `ConditionEnvironment=WAYLAND_DISPLAY`, so it ships in the first-run enable list everywhere and skips cleanly on other hardware instead of restart-looping.
- Calibration: the default mapping (`normal=0, left-up=1, bottom-up=2, right-up=3`) is empirically anchored on a stock unit whose laptop pose reports `right-up` — the ArchWiki mapping is off by 90° on this hardware. Units that anchor differently can shift the whole table with an offset written to `~/.config/omarchy/gpd-pocket-4-orientation` (0-3).

## libinput (touch digitizer)

- udev rule applying `LIBINPUT_CALIBRATION_MATRIX` to the NVTK0603 digitizer frame (270° clockwise per the libinput docs).

## Boot

- `/etc/limine-entry-tool.d/gpd-pocket4-orientation.conf`: `fbcon=rotate:1` fixes the text console and LUKS prompt; `video=eDP-1:panel_orientation=right_side_up` covers DRM-based renderers (plymouth). Hyprland ignores the DRM property, so there is no double rotation in the session.
- Idempotent one-time `interface_rotation: 90` prepend to `/boot/limine.conf` for Limine's graphical menu; the unmanaged global option survives `limine-update` regeneration (verified on hardware).
- Adds `iio-sensor-proxy` when the device is detected.

## Detection

- `bin/omarchy-hw-gpd-pocket-4`: DMI `G1628-04` (product name on real hardware), with a "Pocket 4" fallback match.

## What is intentionally left out

- **Two-finger touch scrolling** (userspace daemon injecting `REL_WHEEL_HI_RES` via uinput; MIT, stdlib-only): works, but the polished experience needs Hyprland's pointer emulation suppressed during drags, which currently requires the hyprgrass plugin — ABI-pinned to Hyprland and not packaged. Deferred until a packaged capture mechanism exists to point at.
- **FocalTech FT9362 fingerprint reader**: not upstreamable. The only working driver is FocalTech's proprietary libfprint blob (1.94.4+tod1), and its original distribution repo was DMCA'd, so neither Omarchy nor a comfortable AUR entry can ship it. The MIT libgusb version-node shim alone is not usable without that blob. The full recipe is documented in the discussion for users who want it on their own machine.

## Testing

- `./test/cli` passes (116 ok). The `./test/shell` failures around the omarchy-pkgs checkout are environmental and reproduce identically on pristine upstream.
- On hardware: detector exit codes, transient unit start through `ExecCondition`, daemon applies transform 3 with correct monitor/touch device detection, calibration offset override, `interface_rotation` prepend idempotency (two runs → one line), and `systemd-analyze verify` / `bash -n` / `py_compile` all clean.


ekollof:
Rebased onto current quattro (2fbac0c8). Dropped the two incidental keyboard-panel commits during the rebase, so this PR is now GPD-only (5 commits). gpd-pocket-4-test.sh and systemd-test.sh pass.

PR #9165, opened 2026-08-30
fix(hypr): route Super+J through split-toggle helper to prevent crash on scrolling workspaces

### LINKED ISSUE #9146: [Bug] Super+J dispatches unsupported togglesplit in Scrolling workspaces
## What happened

On a workspace using the Scrolling layout, pressing the default `Super+J` binding produces a runtime Lua error:

    no such layoutmsg for scrolling

The binding remains advertised as `Toggle window split`, but it cannot perform that action in the active layout.

## Steps to reproduce

1. On a workspace, press `Super+L` to switch the workspace layout to Scrolling.
2. Open at least two tiled windows.
3. Press `Super+J`.

## Expected behavior

The unsupported action should be handled gracefully: either show a clear notification that split toggling is only available in Dwindle, or dispatch an action supported by Scrolling. It should not surface a Lua runtime error.

For Dwindle, the intended behavior is to toggle the split between side-by-side and stacked orientations.

## Observations

Omarchy's stock binding is:

    o.bind("SUPER + J", "Toggle window split", hl.dsp.layout("togglesplit"))

`togglesplit` is a Dwindle layout message. Omarchy's `Super+L` command toggles layouts per workspace, so the global `general:layout = dwindle` setting does not prevent a workspace from being Scrolling. The keybinding is still installed unchanged in that workspace.

The affected machine had saved workspace layout overrides for workspaces 1 and 3:

    ~/.local/state/omarchy/workspace-layouts/1.lua -> scrolling
    ~/.local/state/omarchy/workspace-layouts/3.lua -> scrolling

## System

- Omarchy: 4.0.1-1
- Hyprland: 0.56.2
- Kernel: 7.1.9-arch1-2
- GPU: NVIDIA GeForce RTX 4070, driver 610.57.04
- Display: HDMI-A-1, 3840x2160

A diagnostic log was generated with `omarchy-debug --no-sudo --print`; the local file is `/tmp/omarchy-debug.log`.

harshithnadig:
### Problem
In #9146, pressing `Super + J` on a workspace using the `scrolling` layout produces a Lua runtime error (`no such layoutmsg for scrolling`) because `togglesplit` is only supported by the Dwindle layout engine.

### Solution
- Add `bin/omarchy-hyprland-window-split-toggle` to check the active workspace's `tiledLayout`.
- Dispatch `togglesplit` when in `dwindle` layout, and notify gracefully when on unsupported layouts.
- Point `Super + J` in `default/hypr/bindings/tiling.lua` to the new wrapper.

Fixes #9146

### DIFF
--- a/bin/omarchy-hyprland-window-split-toggle
+++ b/bin/omarchy-hyprland-window-split-toggle
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+# omarchy:summary=Toggle window split orientation on supported layouts
+# omarchy:hidden=true
+
+set -euo pipefail
+
+layout=$(hyprctl activeworkspace -j 2>/dev/null | jq -r '.tiledLayout // "dwindle"')
+
+if [[ $layout == "dwindle" ]]; then
+  hyprctl dispatch 'hl.dsp.layout("togglesplit")' >/dev/null 2>&1 || \
+    hyprctl dispatch layoutmsg togglesplit >/dev/null 2>&1 || true
+else
+  omarchy-notification-send -g 󱂬 "Split toggle is not supported on $layout layout" >/dev/null 2>&1 || true
+fi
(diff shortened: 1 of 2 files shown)


bjarneo:

PR #9170, opened 2026-08-30
Let idle screensaver and lock timings be disabled

### LINKED ISSUE #7178: No way to disable the idle lock while keeping the screensaver — and `idle.lock: 0` locks instantly instead of disabling
## Summary

There is no way to keep the screensaver while disabling the idle lock, and the value a user would naturally reach for to express that — `idle.lock: 0` — does the exact opposite: it locks **immediately** on idle.

Two asks in one report, because they are the same gap seen from both sides:

1. **Feature** — give `idle.lock` a real "never lock" value.
2. **Bug / footgun** — `0`, `null` and `false` all currently mean "lock instantly", which is the opposite of how every one of them reads.

## Use case

Desktop that runs long jobs unattended. The user wants the screensaver to kick in, but never wants to come back to a password prompt, and never wants the machine to suspend. Screensaver and lock are separate concepts in the shell (`omarchy-launch-screensaver` vs `omarchy-system-lock`), so wanting one without the other is coherent — there is just no way to express it.

## The footgun

`shell/plugins/services/idle/Service.qml`:

```qml
readonly property int lockTimeoutSeconds: secondsFromConfig(idleConfig.lock, defaultLockSeconds)
readonly property int firstIdleTimeoutSeconds: Math.min(screensaverTimeoutSeconds, lockTimeoutSeconds)
readonly property int lockDelaySeconds: Math.max(0, lockTimeoutSeconds - firstIdleTimeoutSeconds)
...
if (root.lockDelaySeconds === 0) lockSystem("lock-timeout-immediate")
```

With `"lock": 0`, `firstIdleTimeoutSeconds` becomes `min(screensaver, 0) = 0` and `lockDelaySeconds` is `0`, so `startIdleCycle()` calls `lockSystem("lock-timeout-immediate")` the moment idle is detected. The `IdleMonitor` timeout also collapses to `0`.

The other two intuitive spellings fail the same way, via `shell/plugins/services/idle/IdleModel.js`:

```js
  var n = Number(value)
  return Math.floor(n)
}
```

`Number(null)` and `Number(false)` are both `0` — finite and not negative — so they sail through the guard and land on the immediate-lock path. And a negative value is rejected into the 300s fallback, so there is no escape hatch there either.

Net result: of the four values a user would try to mean "off" (`0`, `-1`, `null`, `false`), three lock instantly and one silently keeps the default.

## Current workaround, and why it is not a fix

Set `idle.lock` to a value large enough that the timer never fires:

```json
{ "idle": { "screensaver": 150, "lock": 604800 } }
```

This works — the lock timer is armed for ~7 days and the cycle is cancelled the moment the screensaver is dismissed via `handleScreensaverWindowClosed` → `cancelIdleCycle`. But it is a workaround for three reasons:

- **It relies on an undocumented ceiling.** `Timer.interval` is a 32-bit int in milliseconds, so `lockDelaySeconds` must stay under ~2,147,483 (≈24.8 days) or it overflows. Nothing in `shell/README.md` tells a user this, and the natural instinct (`99999999`) is past the limit.
- **It does not express intent.** A reader of `shell.json` sees a magic number, not "locking is off".
- **`shell/README.md` §7 documents `idle.lock` as seconds only.** There is no supported spelling for the behaviour, so anyone who wants it has to read the QML.

## Suggested fix

Treat an explicit `false` (and/or `null`) as "never lock", and either accept `0` as the same thing or reject it loudly rather than treating it as instant:

```qml
readonly property bool lockEnabled: idleConfig.lock !== false
                                 && idleConfig.lock !== null
                                 && lockTimeoutSeconds > 0
```

then guard the two `lockSystem` call sites (the `lockDelaySeconds === 0` branch in `startIdleCycle`, and `lockTimer.onTriggered`) on `lockEnabled`, and skip arming `lockTimer` when it is false.

If instant-lock-on-idle is a behaviour someone actually relies on, it deserves an explicit spelling (`"lock": "immediate"`) rather than being what you get for typing the number that reads as "off".

Whatever is chosen, `shell/README.md` §7 should state it, since that section is the only place the idle schema is documented.

## Environment

- Omarchy 4.0.0.alpha
- Config: `~/.config/omarchy/shell.json` → `idle.screensaver: 150`, `idle.lock: 300` (default)
- Verified against the running shell with `omarchy-shell idle status`

### LINKED ISSUE #7506: idle.lock above ~24.8 days overflows the int32 timer ms: 1ms timer storm fills $XDG_RUNTIME_DIR and breaks all app launching
## Summary

Setting `idle.lock` in `shell.json` to a large value meant to express "never lock" — the natural `999999999` — overflows the 32-bit-millisecond `Timer.interval` of the idle service's `lockTimer`: `(lock − screensaver) × 1000` wraps to a negative millisecond count. Qt clamps to 1 ms and prints `WARN: QBasicTimer::start: negative intervals aren't allowed; the interval will be set to 1ms.` on every re-arm, which then continues for the **entire duration of every idle cycle** (~800 lines/s measured). The shell's instance log (`$XDG_RUNTIME_DIR/quickshell/by-id/<instance>/log.log`) grows ~2.7 MB per 30 s; on my 1.6 GB default `/run/user/1000` it filled the tmpfs in ~2.3 h, after which **no application could be launched** (critical "App failure: Invalid arguments:" on every launch — see system impact below).

Reproduced twice: the original incident (Aug 18 22:50 → Aug 19 05:07, 13.7 M log lines) and a controlled re-pro performed while filing this issue.

Relevant code, `shell/plugins/services/idle/Service.qml`:

```qml
readonly property int lockDelaySeconds: Math.max(0, lockTimeoutSeconds - firstIdleTimeoutSeconds)
...
Timer {
  id: lockTimer
  interval: root.lockDelaySeconds * 1000
```

With `lock: 999999999`, `screensaver: 150`: `999849999 × 1000 = 999,849,999,000 ms`, which as an int32 is `999849999000 mod 2^32 = 3,417,586,328` → signed = **−877,380,968 ms**. Negative interval → QBasicTimer warns and clamps to 1 ms on each start.

Note: the wrapped timer never managed to fire `lockSystem` once in 2.3 h of flooding (zero `lock-system` events in that window), so the user is *not* actually locked — the damage is purely the log flood.

## Expected behavior

An out-of-range (or absurdly large) idle value should be clamped to the int32-ms safe range or rejected loudly at config load — not wrap into a broken timer whose warning spam fills the runtime directory.

## Reproduction steps

1. `~/.config/omarchy/shell.json` → `{ "idle": { "screensaver": 150, "lock": 999999999 } }`
2. `omarchy restart shell`
3. Leave the machine with no input for 150 s
4. `tail -f $XDG_RUNTIME_DIR/quickshell/by-id/<newest>/log.log`

Observed (controlled re-pro, this filing):

```
16:48:39.224  DEBUG qml: omarchy idle ... idle-cycle-start: screensaver=150 lock=999999999
16:48:39.225  WARN: QBasicTimer::start: negative intervals aren't allowed; the interval will be set to 1ms.
16:48:39.227  WARN: QBasicTimer::start: negative intervals aren't allowed; ...
```

First warning lands ~1 ms after `idle-cycle-start`; ~24,102 warnings in the following 30 s, 2,800 KB of log. The storm persists for the whole idle cycle and resumes on every new cycle; activity cancels the cycle and the storm.

Safe values (`lock ≤ 2,147,483` s, i.e. delay × 1000 fits int32) do not warn; the lock timer simply arms for a very long time.

## System impact (why this is more than the undocumented ceiling)

#7178 already notes, in its workaround analysis, that the large-value workaround "relies on an undocumented ceiling" (~24.8 days). This report adds the demonstrated, session-breaking consequence of crossing it:

1. The instance log fills `/run/user/1000` to 100 %.
2. uwsm's `wayland-wm-app-daemon.service` crash-loops with `[Errno 28] No space left on device`. Each restart finds its `app_daemon_error` flag left truncated/empty by the ENOSPC'd write and immediately replies `error 'Invalid arguments: ' 2` to the next request (main.py `app_daemon()`, "argparse exit workaround").
3. Every `omarchy launch terminal` / `browser` / menu launch therefore fails with a critical "App failure: Invalid arguments:" notification. No application can be opened until the log is truncated and the value fixed.

uwsm-side evidence from the incident:

```
uwsm_app-daemon: [Errno 28] No space left on device
systemd[1022]: wayland-wm-app-daemon.service: Main process exited, status=1/FAILURE (restart counter 2→4)
uwsm_app-daemon: error flag /run/user/1000/uwsm/app_daemon_error exists
uwsm_app-daemon: sent: error 'Invalid arguments: ' 2
```

Incident instance-log evidence (13,673,933 lines total):

```
2026-08-18 22:50:38.580 DEBUG qml: omarchy idle ... idle-cycle-start: screensaver=150 lock=999999999
2026-08-18 22:50:38.580 DEBUG qml: omarchy idle ... process-start: screensaver ...
2026-08-18 22:50:38.582 WARN: QBasicTimer::start: negative intervals aren't allowed; ...
  ... ~800 lines/s until activity ...
2026-08-19 05:07:15.682 WARN: QBasicT   ← last line, cut mid-write as the tmpfs hit 100 %
```

## Workaround (applied on the affected system)

- `idle.lock ≤ 2,147,483` (safe ceiling). I use `lock: 2000000` (~23 days); verified with `omarchy-shell idle status` → `"lock":2000000,"lockDelay":1999850`.
- One-time recovery of an affected system: `truncate -s 0` the giant instance log(s), `omarchy restart shell`, `systemctl --user restart wayland-wm-app-daemon.service`, and confirm `omarchy launch terminal` works.

## Suggested fixes

- Validate `idle.lock` (and `idle.screensaver`) against the int32-ms ceiling — e.g. reject/clamp anything whose resulting `delay × 1000 > 2,147,483,647` — in `IdleModel.secondsFromConfig` or `Service.qml`. Falling back to the 300 s default (the current "negative value" behaviour) is a footgun for this case; clamping to the ceiling (or logging once at config load, not per re-arm) would be friendlier.
- Give a real "never lock" its first-class value (see #7178) so users don't need to reach for giant numbers at all.
- Optional hardening: cap/rotate the quickshell instance log so one misbehaving timer cannot exhaust `$XDG_RUNTIME_DIR` by itself.

## Related

- #7178 — no way to disable the idle lock; its workaround analysis documents the same ~24.8-day int32 ceiling ("the natural instinct (`99999999`) is past the limit").

## System details

- Omarchy 4.0.0-1 (stable channel)
- Hyprland 0.56.2, Quickshell 0.3.0 (quickshell-git, revision 28771c7)
- Kernel 7.1.8, 15 GiB RAM, Intel i7-9750H (UHD 630 + RX 5500)

voglster:
Closes #7178
Closes #7506

## The problem

`idle.screensaver` and `idle.lock` in `~/.config/omarchy/shell.json` have no way to say "off", and the values that *look* like off do the opposite. The parser is `secondsFromConfig()` in `shell/plugins/services/idle/IdleModel.js`:

```js
var n = Number(value)
return Math.floor(n)
```

`Number(null)`, `Number(false)` and `Number("")` are all `0`, and a `0` timing means *fire the moment the session goes idle*. Behaviour on `quattro` before this PR:

| config value | result | why it matters |
|---|---|---|
| `null` | **0 → fires immediately** | `Number(null) === 0` |
| `false` | **0 → fires immediately** | `Number(false) === 0` |
| `""` | **0 → fires immediately** | `Number("") === 0` |
| `-1` | 300 (silent default) | looks like "off", silently isn't |
| `"off"` | 300 (silent default) | NaN → fallback |
| missing | 300 (default) | correct |
| `999999999` | returned as-is | overflows the Timer int |

So the three most intuitive spellings of "off" lock the screen instantly, and the next two silently keep the default — the user thinks it worked until they get locked out mid-task. Both timings share the parser, so `"screensaver": null` launches the screensaver instantly too.

The overflow is separate: `lockDelaySeconds * 1000` feeds a QML `Timer.interval`, a signed 32-bit millisecond count (max 2147483647 ms ≈ 24.8 days). Anything above that wraps and can fire immediately — again the opposite of intent.

Real case that surfaced it: a stationary desktop that should never idle-lock, but still needs the screensaver and monitor blanking for burn-in. The only workaround today is `"lock": 604800` — 7 days, picked to stay under the 32-bit ceiling.

## The change

Distinguish *unset* from *explicitly disabled*:

- `null`, `false`, `"off"`, `"never"` (also `"none"`, `"no"`, `"disabled"`), and negative numbers → **disabled**, via a `DISABLED_SECONDS = -1` sentinel
- missing, `undefined`, `true`, `""`, and unparseable strings → existing default, unchanged
- `0` → still 0, still fires immediately — deliberately left alone, since changing it would be a behaviour change
- everything else → clamped to `MAX_SECONDS = 2147483`

A disabled timing is *skipped* when the idle cycle starts rather than pushed out to a huge interval, so its `Timer` is never started at all. `firstIdleTimeout()` arbitrates which timing arms the `IdleMonitor`, ignoring disabled ones, and the monitor itself is disabled when both are off.

`"lock": false` now does what it reads as: screensaver and monitor blanking keep working, the session never locks.

## Judgement calls

- **`""` falls back to the default rather than disabling.** It's listed above as one of the "looks like off" values, but an empty string reads more like *unset* than a deliberate choice, and defaulting is the conservative side of a security-relevant decision — a stray empty value shouldn't silently stop a laptop locking. Documented as unset.
- **`0` left as-is.** It's a valid, if unusual, "lock immediately" and some config could rely on it. Changing it would be the behaviour change this PR is trying to avoid.
- **Clamp rather than reject** oversized values, so the 7-day workaround above keeps working instead of breaking on upgrade.
- **`enabled` in `idle status` still means stay-awake.** I added separate `screensaverConfigured` / `lockConfigured` booleans rather than overloading it, so `omarchy-debug-idle` and the runtime smoke test's `.enabled | type == "boolean"` check are unaffected. The numeric `screensaver` / `lock` fields keep their type and report `-1` when disabled.

No behaviour change for any config that works today.

## Testing

`test/shell.d/idle-test.sh` covers the whole table at the parser level — each disabled spelling, the unset fallbacks, `0` staying immediate, the clamp, and `firstIdleTimeout()` including the both-disabled case. `./test/cli` passes.

`test/acceptance.d/idle-test.sh` is new and covers the config surface on a real desktop. I ran it in the disposable VM (`omarchy-iso-test` against the published `omarchy-4.0.1.iso`, which is the version this was reported from), with local `shell/` synced in:

```
ok - a lock disabled with false is reported as unconfigured
ok - a lock disabled with false reports the disabled sentinel
ok - a lock disabled with false leaves the screensaver configured
ok - a lock disabled with null is reported as unconfigured
ok - a lock disabled with null reports the disabled sentinel
ok - a lock disabled with null leaves the screensaver configured
ok - a lock disabled with "off" is reported as unconfigured
ok - a lock disabled with "off" reports the disabled sentinel
ok - a lock disabled with "off" leaves the screensaver configured
ok - an oversized lock clamps to the timer ceiling
ok - an omitted lock keeps the default timing
ok - an omitted lock stays configured
```

Reverting just `IdleModel.js` and `Service.qml` to their `quattro` versions in that same VM fails the test on its first assertion, and reproduces the bug on a real installed system — `"lock": false` yields:

```json
{"lock": 0, "lockDelay": 0}
```

`lockDelay: 0` is `lockSystem("lock-timeout-immediate")` the moment the session goes idle.

### What the acceptance test does not cover

The idle-*cycle* assertions — screensaver launches, lock timer never arms, session stays unlocked across a real idle transition — skip in that VM, because the shell's `IdleMonitor` never reports idle there. That is pre-existing and unrelated to this change: with plain `{"screensaver": 2, "lock": 300}` and no disabled value anywhere, the shell still never goes idle in the guest, while a standalone Quickshell `IdleMonitor` with the same 2s timeout idles immediately in the same session. So the test skips rather than failing for a reason that has nothing to do with the timings, and those assertions will run on a machine whose session does idle. Worth a second pair of eyes from someone who can reproduce a real idle transition.

### DIFF
--- a/manual/13-toggles-idle-screensaver.md
+++ b/manual/13-toggles-idle-screensaver.md
@@ -78,6 +78,19 @@ The Omarchy shell owns idle behavior, and the timings are a top-level `idle` blo
 
 Both numbers are seconds counted from the moment you went idle — not from each other. So with the defaults, the screensaver comes up after two and a half minutes and the lock screen takes over at five minutes, whether or not the screensaver ran. Save the file and the shell picks up the new timings right away.
 
+Either timing can be turned off on its own by setting it to `false` (`null` and `"off"` work too). On a desktop that never needs to lock but still wants the screensaver so the panel doesn't burn in:
+
+```json
+{
+  "idle": {
+    "screensaver": 150,
+    "lock": false
+  }
+}
+```
+
+Leave a key out entirely and it goes back to its default. Setting one to `0` is not "off" — it fires the instant you go idle.
+
 If you dismiss the screensaver before the lock deadline, that counts as activity and the pending lock is cancelled. You don't get locked out for glancing at your machine.
 
 To stop locking on idle entirely, `Super + Ctrl + I` — or `omarchy toggle idle` — flips stay awake on, and the coffee cup indicator appears in the bar. That's the one to hit before a long presentation or a build you want to watch. Hit it again to go back to normal. `omarchy toggle idle status` prints the current state as JSON if you need it from a script.
--- a/shell/README.md
+++ b/shell/README.md
@@ -276,7 +276,12 @@ becomes the authoritative file — we do **not** deep-merge defaults back in.
    entries with their own values.
 7. **Idle timings are top-level.** `idle.screensaver` and `idle.lock`
    are seconds since user idle began, so the default lock fires at 300s
-   even if the 150s screensaver starts first.
+   even if the 150s screensaver starts first. Either one can be turned off
+   on its own with `null`, `false`, or `"off"` (`"never"` also works) —
+   `"lock": false` keeps the screensaver and monitor blanking while never
+   locking. Omitting a key falls back to its default; `0` still means fire
+   the moment the session goes idle. Timings are clamped to 2147483 seconds,
+   the ceiling of the millisecond timer behind them.
 8. **`version: 1` is required** at the top level. The shell will fall back
    to defaults rather than load an unknown version.
 
--- a/shell/plugins/services/idle/IdleModel.js
+++ b/shell/plugins/services/idle/IdleModel.js
@@ -1,7 +1,38 @@
+var DISABLED_SECONDS = -1
+
+// Timer.interval is a signed 32-bit millisecond count, so anything past this
+// many seconds wraps around and fires immediately -- the opposite of what a
+// very long timeout asks for.
+var MAX_SECONDS = 2147483
+
+var DISABLED_WORDS = ["off", "never", "none", "no", "false", "disabled"]
+
+function isDisabled(seconds) {
+  return seconds === DISABLED_SECONDS
+}
+
 function secondsFromConfig(value, fallback) {
+  if (value === null || value === false) return DISABLED_SECONDS
+  if (value === undefined || value === true) return fallback
+
+  if (typeof value === "string") {
+    var word = value.trim().toLowerCase()
+    if (word === "") return fallback
+    if (DISABLED_WORDS.indexOf(word) !== -1) return DISABLED_SECONDS
+  }
+
   var n = Number(value)
-  if (!isFinite(n) || n < 0) return fallback
-  return Math.floor(n)
+  if (!isFinite(n)) return fallback
+  if (n < 0) return DISABLED_SECONDS
+  return Math.min(Math.floor(n), MAX_SECONDS)
+}
+
+// The idle monitor arms once, at whichever timeout comes first; a disabled
+// timing does not get a say in when that is.
+function firstIdleTimeout(screensaverSeconds, lockSeconds) {
+  if (isDisabled(screensaverSeconds)) return isDisabled(lockSeconds) ? 0 : lockSeconds
+  if (isDisabled(lockSeconds)) return screensaverSeconds
+  return Math.min(screensaverSeconds, lockSeconds)
 }
 
 function eventParts(event, count) {
@@ -45,7 +76,11 @@ function screensaverWindowsAfter(windows, address, visible) {
 
 if (typeof module !== "undefined") {
   module.exports = {
+    DISABLED_SECONDS: DISABLED_SECONDS,
+    MAX_SECONDS: MAX_SECONDS,
+    isDisabled: isDisabled,
     secondsFromConfig: secondsFromConfig,
+    firstIdleTimeout: firstIdleTimeout,
     eventParts: eventParts,
     screensaverWindowsAfter: screensaverWindowsAfter
   }
--- a/shell/plugins/services/idle/Service.qml
+++ b/shell/plugins/services/idle/Service.qml
@@ -19,9 +19,12 @@ Item {
   readonly property var idleConfig: shell && shell.shellConfig && shell.shellConfig.idle ? shell.shellConfig.idle : ({})
   readonly property int screensaverTimeoutSeconds: secondsFromConfig(idleConfig.screensaver, defaultScreensaverSeconds)
   readonly property int lockTimeoutSeconds: secondsFromConfig(idleConfig.lock, defaultLockSeconds)
-  readonly property int firstIdleTimeoutSeconds: Math.min(screensaverTimeoutSeconds, lockTimeoutSeconds)
-  readonly property int screensaverDelaySeconds: Math.max(0, screensaverTimeoutSeconds - firstIdleTimeoutSeconds)
-  readonly property int lockDelaySeconds: Math.max(0, lockTimeoutSeconds - firstIdleTimeoutSeconds)
+  readonly property bool screensaverConfigured: !IdleModel.isDisabled(screensaverTimeoutSeconds)
+  readonly property bool lockConfigured: !IdleModel.isDisabled(lockTimeoutSeconds)
+  readonly property bool anyIdleActionConfigured: screensaverConfigured || lockConfigured
+  readonly property int firstIdleTimeoutSeconds: IdleModel.firstIdleTimeout(screensaverTimeoutSeconds, lockTimeoutSeconds)
+  readonly property int screensaverDelaySeconds: screensaverConfigured ? Math.max(0, screensaverTimeoutSeconds - firstIdleTimeoutSeconds) : 0
+  readonly property int lockDelaySeconds: lockConfigured ? Math.max(0, lockTimeoutSeconds - firstIdleTimeoutSeconds) : 0
   readonly property bool idleEnabled: stayAwakeStateLoaded && !stayAwake
   readonly property string screensaverClass: "org.omarchy.screensaver"
 
@@ -38,6 +41,10 @@ Item {
 
   function secondsFromConfig(value, fallback) {
     return IdleModel.secondsFromConfig(value, fallback)
+  }
+
+  function describeTimeout(seconds) {
+    return IdleModel.isDisabled(seconds) ? "off" : String(seconds)
   }
 
   function nowIso() {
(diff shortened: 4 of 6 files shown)


voglster:
Heads up on an overlap I should surface rather than leave for a reviewer to find.

#7578 (opened Aug 20) already fixes the int32 overflow behind #7506, and it got there first. I only came across it today while reading through what else referenced #7506. This PR reached the same clamp independently — the disable work needed the same ceiling to be safe, so it fell out of the change rather than being the goal of it.

We landed on the same line in `secondsFromConfig`, with the same constant and the same reasoning for clamping instead of falling back to the default (a huge number means "never", and dropping to the 5 minute default would lock the user out):

```js
return Math.min(Math.floor(n), 2147483)
```

I've left `Closes #7506` on this PR because the fix genuinely is in here, but I'm not trying to claim that ground. Either sequencing works for me:

- merge #7578 first and I'll rebase this on top and drop my clamp, or
- merge this and #7578 closes as covered.

Maintainers' call entirely.

One thing worth keeping whichever way it goes: #7578 includes a test asserting that both timeouts actually reach the timers *through* `secondsFromConfig`, which guards against someone later reading a timeout straight off `idleConfig` and bringing the 1 ms timer storm back. This PR doesn't have that assertion and it's a good one.

MBemera:
@voglster thanks for the comment mentioning my previous PR.  Appreciated the acknowledgement and well written. Thanks again. -MB

PR #9171, opened 2026-08-30
Do not activate mise shims globally in UWSM session

### LINKED ISSUE #9158: Blender crashes at startup when UWSM exposes global Mise Python
## What happened

On Omarchy 4.0.1-1, Arch Blender 17:5.2.0-4 crashes with SIGSEGV
approximately two seconds after launching it from the desktop.

A global Mise Python 3.14.7 environment is installed through Omarchy.

## Steps to reproduce

1. Install Omarchy's Python development environment, making Python global in Mise.
2. Install Blender.
3. Launch Blender from its desktop entry.

## Expected

Blender opens normally.

## Actual

Blender's embedded Python imports its standard library from:

`~/.local/share/mise/installs/python/3.14.7`

Startup then reports:

```
ModuleNotFoundError: No module named 'math'
ModuleNotFoundError: No module named '_posixsubprocess'
```

The `bl_operators` startup package aborts before registering
`WM_OT_context_toggle`.

The symbolized crash stack is:

```
WM_operator_properties_create_ptr(ot=0x0)
WM_gizmo_operator_set(ot=0x0)
WIDGETGROUP_navigate_setup()
```

The immediate fault is an access to `ot->srna` while `ot` is null.

## Control test

Launching with a system-only PATH succeeds in registering the missing operator:

```bash
env PATH=/usr/bin:/bin blender
```

The normal Omarchy environment reproduces the missing Python modules.

## Omarchy involvement

`/usr/share/omarchy/default/uwsm/env.d/10-omarchy` activates Mise shims
globally for the UWSM session:

```bash
omarchy-cmd-present mise && eval "$(mise activate bash --shims)"
```

This exposes the developer Python to GUI applications with embedded Python
runtimes.

## System

- Omarchy: 4.0.1-1
- Blender: 17:5.2.0-4
- Mise Python: 3.14.7
- Signal: SIGSEGV
- GPU: NVIDIA RTX 3050 Mobile plus AMD Cezanne iGPU

`omarchy debug --no-sudo --print`, referenced by the crash-reporting
instructions, is no longer available in Omarchy 4.0.1.

Filed by GPT-5 via Codex.

harshithnadig:
## Summary

Fixes #9158.

### Root Cause
`default/uwsm/env.d/10-omarchy` previously executed:
```bash
omarchy-cmd-present mise && eval "$(mise activate bash --shims)"
```
`mise activate bash --shims` outputs `export PATH="$HOME/.local/share/mise/shims:$PATH"`, which prepended mise shims to `PATH` session-wide across the graphical session and systemd user services.

When a user installed a developer runtime like Python via Mise (e.g. `mise use -g python@latest`), native desktop applications with embedded Python interpreters (such as Blender) picked up the global Mise Python standard library rather than Arch's system Python runtime. This caused ABI mismatches (failing to import `math`, `_posixsubprocess`, etc.) and crashed Blender on startup with SIGSEGV.

### Resolution
- Removed `eval "$(mise activate bash --shims)"` from `default/uwsm/env.d/10-omarchy`.
- `default/bash/env-bootstrap` (sourced at session start) already appends `~/.local/share/mise/shims` and `~/.local/bin` to the end of `PATH`, ensuring system binaries in `/usr/bin` retain precedence for GUI application launches while keeping user tools accessible.
- Interactive shells continue to receive interactive Mise activation via `default/bash/init`.
- Added test coverage in `test/shell.d/browser-env-test.sh` to verify that UWSM session environment does not globally activate mise shims ahead of system `PATH`.

### Validation
- Validated test suite passes inside an unprivileged Bubblewrap sandbox container.

### DIFF
--- a/default/uwsm/env.d/10-omarchy
+++ b/default/uwsm/env.d/10-omarchy
@@ -14,6 +14,3 @@ fi
 
 # Optional user overrides kept for compatibility with older installs.
 [ -f "$HOME/.config/uwsm/default" ] && . "$HOME/.config/uwsm/default"
-
-# Activate mise if present on the system
-omarchy-cmd-present mise && eval "$(mise activate bash --shims)"
--- a/test/shell.d/browser-env-test.sh
+++ b/test/shell.d/browser-env-test.sh
@@ -24,3 +24,37 @@ pass "uwsm session env leaves BROWSER unset"
 
 ! grep -q "export BROWSER" "$uwsm_env" || fail "uwsm env.d fallback leaves BROWSER unset"
 pass "uwsm env.d fallback leaves BROWSER unset"
+
+# Global mise shims must not be activated session-wide in UWSM env, which would
+# place user-installed toolchains (e.g. Python) before system binaries and crash
+# native apps (like Blender) that link against system runtimes.
+! grep -q "mise activate" "$uwsm_env" || fail "uwsm session env leaves mise shims un-activated"
+pass "uwsm session env leaves mise shims un-activated"
+
+tmpdir=$(mktemp -d)
+trap 'rm -rf "$tmpdir"' EXIT
+
+mock_bin="$tmpdir/bin"
+mkdir -p "$mock_bin"
+cat >"$mock_bin/mise" <<'EOF'
+#!/bin/bash
+if [[ "$*" == *"activate"*"--shims"* ]]; then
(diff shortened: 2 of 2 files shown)


bjarneo:
Automated duplication check: this pull request looks similar to #8406, which covers the same mise shims PATH order. I keep that one open and close this one to consolidate review.

Issue #6880, opened 2026-08-14
SDDM greeter always uses US keyboard layout, ignoring the system layout


g-desoutter:
### System details

Omarchy 4.0.0-1, i5-12600K, RTX 5070,Keyboard layout: fr 

## Describe the issue

The SDDM greeter always uses the US/QWERTY layout, regardless of the system keyboard layout. On a FR/AZERTY system, the login password field is QWERTY, so any password containing layout-sensitive characters cannot be typed. There is no layout indicator or switcher on the greeter.

Note: the system XKB configuration is already correct, so `localectl set-x11-keymap` is not the fix. Hyprland ignores `/etc/X11/xorg.conf.d/00-keyboard.conf` entirely — it only honours `input { kb_layout }` in its own config — which is why the greeter is wrong while the rest of the system is right:

```
$ localectl status
  VC Keymap: fr
  X11 Layout: fr
  X11 Model: pc105

$ cat /etc/X11/xorg.conf.d/00-keyboard.conf
  Option "XkbLayout" "fr"
  Option "XkbModel" "pc105"

$ hyprctl getoption input:kb_layout
  str: fr
```

Cause: the greeter runs inside its own Hyprland instance.

```ini
# /etc/sddm.conf.d/10-wayland.conf
[Wayland]
CompositorCommand=start-hyprland -- --config /usr/share/sddm/hyprland.conf
```

Neither `/usr/share/sddm/hyprland.conf` nor the newer `/usr/share/sddm/hyprland.lua` declares an input block — both only set `misc{}` and `animations{}`. With no `kb_layout`, Hyprland falls back to its built-in default, `us`.

Confirmed: pointing `CompositorCommand` at a config that sources the shipped one and adds `input { kb_layout = fr }` makes the greeter AZERTY. The fix therefore belongs in the greeter's compositor config.

## Steps to recreate

1. Install with a non-US layout (FR here) and full-disk encryption.
2. Disable autologin by moving `/etc/sddm.conf.d/autologin.conf` out of the directory. SDDM reads every file in `conf.d` regardless of extension, so renaming it to `.disabled` in place is not enough — it is still parsed.
3. Reboot.
4. The LUKS decryption prompt is correctly AZERTY (vconsole `KEYMAP=fr`).
5. Seconds later, on the same physical keyboard, the SDDM greeter is US/QWERTY.

[...]

PR #9177, opened 2026-08-30
Apply the configured keyboard layout on the SDDM greeter

florentdestremau:
## Problem

The SDDM Wayland greeter starts its own minimal Hyprland from `default/sddm/hyprland.lua`, which only sets `misc`/`animations`. The layout resolution from `/etc/vconsole.conf` lives solely in `default/hypr/input.lua` (the user session), so **the greeter always falls back to `us`/QWERTY**.

Anyone whose configured layout differs from US (e.g. `fr`, `de`) gets a QWERTY password field on the login/lock greeter, even though `localectl`, the console and the session are all correct. Switching the greeter to `hyprland.lua` (as the current wayland conf does) does not help, because that file never applied a layout in the first place.

## Fix

- Extract the vconsole → `kb_layout`/`kb_variant`/`kb_options` resolution into a new pure module `default/hypr/keyboard.lua` (single source of truth).
- `default/hypr/input.lua` now consumes it (behavior unchanged for the session).
- `default/sddm/hyprland.lua` loads the same module and applies the layout to the greeter, so the password field matches the session.

The greeter loads the module **defensively**: it runs without `bootstrap.lua`, so it extends `package.path` by hand and guards every step (`package`/`require`/`pcall`). Any failure falls back to Hyprland's default layout rather than breaking the login screen.

## Tests

`test/shell.d/hyprland-keyboard-layout-test.sh`:
- the non-latin-list sync check now reads `keyboard.lua`;
- added three greeter cases: configured latin layout passes through, non-latin mirrors the session (`us,` prepended + `grp:alts_toggle`), and a missing `vconsole.conf` falls back to `us`.

All keyboard tests pass. The pre-existing failures on a bare clone (missing `omarchy-pkgs` sibling checkout, live-Quickshell IPC, VM mounts) are unrelated to this change.

### DIFF
--- a/default/hypr/input.lua
+++ b/default/hypr/input.lua
@@ -1,58 +1,15 @@
 -- https://wiki.hypr.land/Configuring/Basics/Variables/#input
 
-local function read_vconsole()
-  local values = {}
-  local file = io.open("/etc/vconsole.conf", "r")
-  if not file then
-    return values
-  end
-
-  for line in file:lines() do
-    local key, value = line:match("^%s*([%w_]+)%s*=%s*(.-)%s*$")
-    if key and value then
-      value = value:gsub("%s+#.*$", "")
-      value = value:gsub('^"(.*)"$', "%1")
-      value = value:gsub("^'(.*)'$", "%1")
-      values[key] = value
-    end
-  end
-
-  file:close()
-  return values
-end
-
--- Layouts that can't type Latin letters. Keep in sync with the list in
--- etc/mkinitcpio.conf.d/omarchy_hooks.conf.
-local non_latin_layouts =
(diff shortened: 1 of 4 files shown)


florentdestremau:
Closing as a duplicate — I missed that #6896 already fixes this (and is further along in review). Apologies for the noise.

For anyone landing here: the tracking issue is #6880 and the fix is #6896, which resolves the layout from `/etc/vconsole.conf` directly in `default/sddm/hyprland.lua`. That PR's thread also covers why the resolution is intentionally duplicated rather than shared via a module (a failed `require()` in the greeter would lock everyone out, and reaching the Omarchy Lua tree would mean hardcoding `/usr/share/omarchy`, breaking the `$OMARCHY_PATH` linked-checkout dev setup). Deferring to #6896.

PR #9181, opened 2026-08-30
Restore lock password focus after resume

### LINKED ISSUE #9180: [Quattro] Lock password field can lose focus after suspend resume
### System details

* Omarchy 4.0.1-1 (Quattro)
* Hyprland 0.56.2-1
* NVIDIA RTX 3080 Ti
* Two DisplayPort outputs: 2560x1440 at 180 Hz and 200 Hz
* Linux 7.1.9

### What is wrong?

After suspending while the native Omarchy session lock is active, a restored or recreated lock surface can show a password field that does not own keyboard focus. Focus can also be lost while PAM disables the field and then releases it after a failed attempt.

This is separate from the slow-display reblank problem in #7399. It is also separate from the first modifier event lost during input-device restoration, which was later isolated in hyprwm/aquamarine#375.

### Steps to reproduce

1. Suspend through the normal Omarchy path so the session is locked before sleep.
2. Resume and wait for the displays to return.
3. Inspect or interact with the visible password field after output recreation.
4. The field can lack active focus. The same loss can occur after PAM rejects a password and re-enables the field.

### Root cause

`LockView.qml` requested `passwordInput.forceActiveFocus()` only when the component completed or `inputEnabled` changed. During resume, ext-session-lock outputs can be recreated while `inputEnabled` remains true. `Service.qml` already observes `Quickshell.onScreensChanged`, but did not ask the recreated view to reacquire keyboard focus.

On a multi-monitor resume, outputs can return sequentially, so one immediate refocus is not sufficient. PAM also temporarily disables the input during authentication, requiring recovery to pause and restart when the field becomes enabled again.

### Expected

The password field owns keyboard focus whenever the visible lock surface is ready for input, including after output recreation and after a failed PAM attempt.

### Verified fix

A cloned `omarchy.lock` service sends a focus-generation request when the lock becomes secure, whenever screens change, and after PAM releases the input. `LockView` is a `FocusScope`, acknowledges `TextInput.activeFocus`, and the service retries briefly across sequential output recreation. Focus telemetry confirmed that the field reacquired active focus.

The separate case-sensitive first-password failure was not used to validate this fix; protocol traces attributed it to aquamarine#375.

A focused regression test covers the retry budget, screen-change trigger, failed-PAM transition, disabled-field guard, focus scope, request handoff, and focus acknowledgement. PR #9181 contains the fix.

ClGratton:
## Problem

After system resume, the visible password field can be left without keyboard focus. On multiple outputs, ext-session-lock surfaces may be recreated one after another. On a single output, the lock surface may survive suspend unchanged, so no screen-change callback occurs at all. A failed PAM attempt creates another focus-loss path because the password field is disabled during authentication and must be focused again after it is re-enabled.

Fixes #9180.

## Cause

`LockView` previously made one imperative focus request during construction. Quickshell constructs lock views before the compositor can grant keyboard focus, and recreated views have no declared focus scope to restore. The earlier version of this PR retried on secure state, screen changes, and PAM completion, but a single-output lock surface can survive sleep without any of those events firing after resume.

## Change

`LockView` is now a keyboard `FocusScope` with a declared focus target. The lock service sends explicit focus generations, acknowledges active focus, pauses while PAM owns a disabled field, and retries across sequential output recreation.

The sleep monitor now remains alive across the complete logind sleep cycle. After a known `PreparingForSleep=false`, it sends one generic `resumeFromSleep` IPC event to the lock service. Resume uses the complete twelve-second focus budget even if an early surface acknowledges focus. Unknown logind state never counts as resume, a vanished D-Bus producer is treated as failure, and the next delay inhibitor is armed before shell notification retries so a rapid second suspend remains protected.

The retry never submits or modifies password contents.

## Verification

The focused lock, monitor, and sleep tests pass. The complete shell suite ran all 226 files on the corrected #9181 branch; every affected test passed, with only the five existing host/baseline failures in `config-test`, `launch-about-test`, `runtime-smoke-test`, `snapper-test`, and `unowned-system-paths-test`.

A real dual-output S3 cycle on NVIDIA and DisplayPort exercised the explicit resume event. Both outputs remained active, the full focus budget ran after logind reported resume, and the first unlock completed normally. A single-output reporter independently showed that an external `PrepareForSleep=false` listener driving the same recovery method fixes the preserved-surface case; the corrected branch is now available for an exact retest of that topology.

### DIFF
--- a/bin/omarchy-system-sleep-monitor
+++ b/bin/omarchy-system-sleep-monitor
@@ -14,6 +14,8 @@ consume_sleep_events() {
       return 0
     fi
   done
+
+  return 1
 }
 
 monitor_sleep_events() {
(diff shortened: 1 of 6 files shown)


isho:
BTW. I tried this on my local environment by copying the change and it solved my issue.

freimer:
Additional reproduction closely matching this focus-loss issue.

On Omarchy 4.0.2 with Quickshell 0.3.1, Hyprland 0.56.2, NVIDIA driver 610.57.04, and four outputs, waking an idle-locked session left the lock UI partially rendered. The password field was visible on some monitors, but pointer/keyboard focus could not reliably be restored. Clicking the password field did not fix it in this case.

The local journal showed simultaneous output-rendering failures:

```text
qt.qpa.wayland: eglSwapBuffers failed with 0x3000
[nvidia-drm] Failed to allocate NVKMS memory for GEM object
qt.qpa.wayland: eglSwapBuffers failed with 0x3003
```

The installed LockView.qml already uses a full-surface MouseArea and calls forceActiveFocus() on click, so this appears to be beyond ordinary QML TextInput focus. It may involve Hyprland/session-lock input routing after output recreation, compounded here by the NVIDIA EGL allocation failure.
This was an idle display-blank wake, not a confirmed system suspend/resume cycle.

Henaneitor:
Verified the same bug on a single-output NVIDIA desktop (Omarchy 4.0.3, Hyprland 0.56.2): lock taken before suspend, surface survives, typing does nothing until pointer input. `forcePasswordFocus()` alone after resume was not enough; the lock window is inactive until something activates it.

A `PrepareForSleep=false` listener that recovers focus (without requiring a click) unblocked keyboard-only unlock on that machine. This PR's sleep-monitor `resumeFromSleep` path is the right place for that event — more complete than putting a D-Bus watcher inside the lock plugin.

Related: #8361, #9180.

adcuelloa:
Reproduced on current stable Omarchy 4.0.4-1.

Relevant environment:
- Hyprland 0.56.2-2
- Quickshell 0.3.1-1
- kernel 7.2.5-3-omarchy
- hybrid Intel/NVIDIA graphics
- single active display

Before the workaround, the password field required a pointer click before accepting keyboard input. I cloned `omarchy.lock` and added the 250 ms self-healing focus guard proposed in #8628, paused while password authentication is active.

After applying it, I verified keyboard-only unlock in all three cases:
1. Immediate manual lock: typed the password without pointer input.
2. DPMS blank after locking: woke with Shift, then typed the password without pointer input.
3. Suspend/lid resume: woke with Shift, then typed the password without pointer input.

This confirms the focus-loss symptoms and that repeated focus recovery covers this machine. I have not tested the #9181 branch itself, so this is supporting hardware/workaround evidence rather than a direct PR validation.

bjarneo:
Automated duplication check: this pull request looks similar to #8560, which covers the same lock focus restore on resume. I keep that one open and close this one to consolidate review.

PR #9189, opened 2026-08-30
Keep the backlight off while the laptop panel is disabled

lewapf:
## Problem

Disabling the internal laptop panel (`omarchy-hyprland-monitor-internal off`, or the DPMS-off path in `omarchy-brightness-display`) sets `hl.monitor({ disabled = true })` on the panel, but the backlight driver keeps powering it at its last brightness level. The panel shows a visible glow even though Hyprland reports it disabled — `dpmsStatus` stays `true`, and issuing a targeted `hl.dsp.dpms disable` against an already-disabled output is a no-op (verified: no change to `dpmsStatus` or backlight level).

Repro on a ThinkPad with `intel_backlight`, external monitor active:
```
$ hyprctl monitors all -j | jq '.[] | select(.name=="eDP-1") | {disabled,dpmsStatus}'
{"disabled": true, "dpmsStatus": true}
$ brightnessctl -d intel_backlight -m
intel_backlight,backlight,19393,100%,19393
```

## Fix

- `omarchy-hyprland-monitor-internal`: `off()` saves the current backlight level and zeroes it (`dim_backlight`); `on()`/`recover()` restore the saved level (`restore_backlight`).
- `omarchy-brightness-display`: same save/zero/restore around the `off`/`on` DPMS dispatch, gated to the non-Apple, non-DDC (internal panel) path.

Backlight device is resolved via the existing `omarchy-hw-display` heuristic, so this covers `intel_backlight`, `amdgpu_bl*`, `gmux_backlight`, `acpi_video*` consistently with the rest of the brightness tooling.

## Test plan

- [x] `omarchy-hyprland-monitor-internal off` → backlight drops to 0%, saved level persisted
- [x] `omarchy-hyprland-monitor-internal on` → backlight restored to prior level
- [x] `bash -n` on both changed scripts

### DIFF
--- a/bin/omarchy-brightness-display
+++ b/bin/omarchy-brightness-display
@@ -59,10 +59,29 @@ fi
 
 step="$1"
 
+state_file="${XDG_RUNTIME_DIR:-/tmp}/omarchy-brightness-display.saved"
+
 if [[ $step == "off" ]]; then
+  # DPMS disable only blanks the signal — the backlight driver keeps the
+  # panel lit at its last level, showing as a dim glow. Zero the backlight
+  # too, saving the prior value so "on" can restore it.
+  if ! use_apple_display && ! use_ddc_display; then
+    device="$(omarchy-hw-display)" || device=""
+    if [[ -n $device ]]; then
+      backlight_brightness "$device" > "$state_file" 2>/dev/null
+      brightnessctl -d "$device" set 0 >/dev/null 2>&1
+    fi
+  fi
   hyprctl dispatch 'hl.dsp.dpms({ action = "disable" })' >/dev/null 2>&1
   exit 0
 elif [[ $step == "on" ]]; then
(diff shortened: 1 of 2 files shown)


sprajs:
The new `off` path loses the saved brightness when called twice before `on`: the second call overwrites the original level with 0, then `on` restores 0 and deletes the saved state. I reproduced this with the current script and mocked brightnessctl/hyprctl: starting at 50%, `off; off; on` finishes at 0%.

Could the first saved value be preserved until a successful restore, with an `off → off → on` regression test? These commands previously allowed repeated DPMS requests without losing the backlight level.


I implemented a fix with regression tests in https://github.com/omacom/omarchy/pull/12423, in the main Omarchy repository, preserving this PR’s original commits. The focused tests pass; the PR records the full-suite failures and remaining runtime validation.

PR #9192, opened 2026-08-30
Add Yii3 development environment

samdark:
Adds [Yii3](https://www.yiiframework.com/) under Install > Development > PHP, provisions the shared PHP toolchain, and shows the official web, API, and console project creation commands.

For the menu I've used logo from NerdFont (Yii is there for a long time already) so it looks nice.

_// Yii is one of the oldest and most popular frameworks in PHP, and Yii3 is its latest modern major version._


gauravbhatia4601:
Damn, @samdark himself is here. Yii was the first framework I ever worked with.

PR #9206, opened 2026-08-30
Warn clearly when no Chromium browser can launch a web app

### LINKED ISSUE #9197: No warning when all Chromium-family browsers are removed breaks webapp launchers
### Summary
Launching any webapp (WhatsApp, Discord, YouTube, X, etc.) crashes with a
cryptic error when the default browser is **not** a Chromium-family browser
and **no** Chromium-family browser (Chrome/Brave/Edge/Opera/Vivaldi/Helium) is
installed. No warning is shown explaining that a supported browser is missing.

### What happened
```
$ omarchy-launch-webapp https://web.whatsapp.com/
Error: Path "--app=https:/web.whatsapp.com" does not exist!
```

Via the desktop launcher the webapp simply "crashes"/does nothing.

### What was expected
A clear warning that no supported (Chromium-family) browser is installed, and
a graceful error instead of a crash.

### Steps to reproduce
1. Set a non-Chromium browser as default, e.g. Zen:
   `xdg-settings set default-web-browser zen.desktop`
   (`zen.desktop` returns `Exec=/opt/zen-browser-bin/zen-bin %u`)
2. Ensure no Chromium-family browser is installed (no `chromium.desktop`,
   `chrome.desktop`, etc. in `/usr/share/applications/` or `~/.local/share/applications/`).
3. Run `omarchy-launch-webapp https://web.whatsapp.com/` (or open the WhatsApp
   desktop launcher / the `SUPER+SHIFT+ALT+G` keybind).

### Root cause
In `bin/omarchy-launch-webapp`:

```bash
case $browser in
esac
```

`xdg-settings get default-web-browser` returns `zen.desktop`, which is **not**
in the whitelist, so the script forces `chromium.desktop`. But `chromium.desktop`
does not exist (Chromium uninstalled), the `sed` emits an empty string, and the
command collapses to:

```
uwsm-app -- --app=https://web.whatsapp.com/
```

The Wayland app daemon has no binary to exec, interprets `--app=...` as a path,
and returns `Error: Path "--app=https:/web.whatsapp.com" does not exist!`.

There is no check that the resolved browser binary actually exists, and no
error/warning surfaced to the user.

### Suggested fix
- After resolving the `.desktop` file, verify a binary was found; if not, print
  a clear message that no supported (Chromium-family) browser is installed and
  exit with a non-zero code.
- Optionally detect whether the default browser is a supported one, and warn
  when falling back.

### Environment
- Omarchy: 4.0.1-1
- Default browser: Zen (`zen.desktop`), Firefox-based
- Kernel: 7.1.9-arch1-2
- Desktop: Hyprland (DM: SDDM), Omarchy
- Machine: Lenovo 300e 2nd Gen (Convertible, LNVNB161216)
- CPU: Intel Pentium Silver N5000 (4 cores, Goldmont Plus)
- GPU: Intel UHD Graphics 605 (i915), Gen-9.5
- Memory: 8 GiB
- Repos: pkgs.omarchy.org/stable (pacman, 1248 packages)

linuts:
## Summary

Fixes #9197

Two related problems made webapp launchers (`omarchy-launch-webapp` — WhatsApp, Discord, YouTube, X, etc.) fail:

1. When no Chromium-family browser was installed, the launcher silently ran a broken `--app=...` invocation with an empty browser binary, which the Wayland app daemon rejected at runtime with a cryptic `Error: Path "--app=https:/..." does not exist!`. From the GUI (desktop launcher / menu / keybind) it failed with no visible reason.
2. Even when a Chromium browser (e.g. Brave) *was* installed, if it wasn't the **default** browser the launcher forced `chromium.desktop` and gave up, pretending no browser existed.

This change:
- Resolves the browser binary first and, when none can be found, fails with a clear terminal message **and** a desktop notification instead of crashing.
- Falls back to any *installed* Chromium-family browser (preferring a whitelisted default), so Brave/Chrome/Edge work even when a Firefox-based browser (Zen) is the default.

## Behavior

No Chromium browser installed, previously:
```
Error: Path "--app=https:/web.whatsapp.com" does not exist!
```

Now:
```
omarchy-launch-webapp: cannot launch web app without a Chromium-based browser.
Install one (e.g. Brave, Chrome, Edge) or set it as the default browser.
```
plus a low-urgency desktop notification ("Web app failed to launch — No Chromium-based browser is installed"), and exit code 1.

Default is Zen but Brave is installed → the webapp now launches in Brave via `setsid uwsm-app -- <binary> --app=<url> <extra args...>`. Extra arguments after the URL are still forwarded.

## Notes

- Added `< /dev/null` to the `sed` resolution so it never blocks reading stdin when the desktop files are genuinely absent.
- Candidate order prefers the whitelisted default browser, then the known Chromium-family desktop IDs.
- Error-surfacing follows existing precedent: `omarchy-launch-screensaver` warns via `omarchy-notification-send` in an unsupported-browser case, and `omarchy-launch-config-editor` uses the same `-u low` urgency.

## Tests

New `test/shell.d/webapp-launch-browser-test.sh`:
- Chromium default browser → launches `uwsm-app -- <binary> --app=<url> <flags>`, forwards extra args.
- Non-Chromium default (Zen) with Brave installed → falls back to the installed Brave.

Verified: `./test/cli` passes; the new shell test passes; related webapp/launch shell tests pass.

### DIFF
--- a/bin/omarchy-launch-webapp
+++ b/bin/omarchy-launch-webapp
@@ -3,14 +3,33 @@
 # omarchy:summary=Launch a URL as a web app in the default supported browser
 # omarchy:args=<url>
 
-browser=$(xdg-settings get default-web-browser)
+browser=$(env -u BROWSER xdg-settings get default-web-browser)
 
 case $browser in
-google-chrome* | brave* | microsoft-edge* | opera* | vivaldi* | helium*) ;;
-*) browser="chromium.desktop" ;;
+google-chrome* | brave* | microsoft-edge* | opera* | vivaldi* | helium* | chromium*) ;;
+*) browser="" ;;
 esac
 
-bin=$(sed -n 's/^Exec=\([^ ]*\).*/\1/p' {~/.local,~/.nix-profile,/usr}/share/applications/$browser 2>/dev/null < /dev/null | head -1)
+candidates=(
+  ${browser:+$browser}
+  google-chrome.desktop
+  brave-browser.desktop
+  brave-origin.desktop
+  microsoft-edge.desktop
+  opera.desktop
+  vivaldi.desktop
+  helium.desktop
+  chromium.desktop
+)
+
+bin=""
+for candidate in "${candidates[@]}"; do
+  bin=$(sed -n 's/^Exec=\([^ ]*\).*/\1/p' \
+    ~/.local/share/applications/$candidate \
+    ~/.nix-profile/share/applications/$candidate \
+    /usr/share/applications/$candidate 2>/dev/null < /dev/null | head -1)
+  [[ -n $bin ]] && break
+done
 
 if [[ -z $bin ]]; then
   echo "omarchy-launch-webapp: cannot launch web app without a Chromium-based browser." >&2
--- a/bin/omarchy-launch-webapp
+++ b/bin/omarchy-launch-webapp
@@ -10,4 +10,13 @@ google-chrome* | brave* | microsoft-edge* | opera* | vivaldi* | helium*) ;;
 *) browser="chromium.desktop" ;;
 esac
 
-exec setsid uwsm-app -- $(sed -n 's/^Exec=\([^ ]*\).*/\1/p' {~/.local,~/.nix-profile,/usr}/share/applications/$browser 2>/dev/null | head -1) --app="$1" "${@:2}"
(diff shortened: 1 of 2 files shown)


JohnL404:
Nice — this is the same core approach as #9172, which now also includes the `env -u BROWSER` guard and the failure notification from here, plus case-insensitive default-browser matching and tests that mask `/usr/share/applications` so they pass on machines with Chromium-family browsers installed. Probably worth the maintainers consolidating these — happy either way.

PR #9219, opened 2026-08-30
Never let bar and panel error diagnostics block their fallbacks

### LINKED ISSUE #9116: Failed bar option leaves no bar at all: pluginBarLoader error handler throws ReferenceError before the fallback runs
### What happened

When a custom bar option fails to load, the shell is supposed to fall back to the default bar. Instead the error handler itself throws, the fallback never runs, and the session is left with **no bar at all** — the layer surface is gone and windows expand into the reclaimed space.

From the quickshell log (Omarchy 4.0.1-1):

```
WARN scene: @shell.qml[256:-1]: ReferenceError: errorString is not defined
```

### Why

`shell/shell.qml`, `pluginBarLoader.onStatusChanged` (line 257 on master):

```qml
onStatusChanged: {
  if (status === Loader.Error) {
    var detail = errorString && errorString() ? errorString() : ""
    console.warn("bar option " + shell.activeBarId + " failed to load, falling back to " + shell.defaultBarId + ":", detail)
    shell.failedBarId = shell.activeBarId
  }
}
```

`QtQuick.Loader` has no `errorString` member, and in QML an unqualified lookup of an undefined name throws `ReferenceError` — the `errorString &&` guard does not protect against that. The throw aborts the handler, so:

- the diagnostic `console.warn` never prints (the real load error is swallowed), and
- `shell.failedBarId` is never set, so `activeBarId` (line 181) never resolves back to `defaultBarId`, so `defaultBarLoader` never activates.

The same unguarded `errorString` pattern also appears at line 646 in the plugin-widget loader path and looks susceptible to the same failure.

### Reproduce

1. Switch to any bar option that fails to load — e.g. the verbatim clone from #9115, or `omarchy bar use x.bar` pointing at a bar QML with a syntax error.
2. `omarchy restart shell`
3. No bar on any monitor; log shows the ReferenceError above instead of the intended "falling back to omarchy.bar" warning.

### Expected

The failed bar option is recorded in `failedBarId` and the stock bar loads, as the surrounding code clearly intends.

### Possible fix direction

Guard the lookup (`typeof errorString === "function"`) or take the detail from the loader's `sourceComponent` when available — and consider wrapping the handler so a failure in diagnostics can never block `shell.failedBarId = shell.activeBarId`.

### System

- Omarchy 4.0.1-1, Arch Linux (kernel 7.1.9-arch1-2)
- CPU: Intel i7-10700K, GPU: NVIDIA Quadro P2200
- Verified the same code is on current master (`shell/shell.qml` lines 257 and 646)

DeanWahle:
Fixes #9116.

When a custom bar option fails to load, `pluginBarLoader.onStatusChanged` is supposed to record `failedBarId` so `activeBarId` resolves back to the default bar. But `QtQuick.Loader` has no `errorString` member, and in QML an unqualified lookup of an undefined name throws `ReferenceError` (the `errorString &&` guard can't help because the throw happens at the lookup itself). The handler aborted on its first line, so:

- the diagnostic never printed (the real load error was swallowed), and
- `failedBarId` was never set, so the fallback never ran and the session was left with **no bar at all**.

**Fix, in both the bar loader (line ~257) and the panel plugin loader (line ~646) which had the same unguarded pattern:**

1. Apply the recovery state change first (`shell.failedBarId` for the bar, `shell.hide(...)` for panels) so nothing thrown while gathering diagnostics can ever block it again.
2. Probe with `typeof errorString === \"function\"` before calling, falling back to `sourceComponent.errorString()` when available.

The bar handler captures `activeBarId` before assigning `failedBarId`, since the assignment reactively flips `activeBarId` to the default and the warning should name the bar that actually failed.

### DIFF
--- a/shell/shell.qml
+++ b/shell/shell.qml
@@ -253,9 +253,16 @@ ShellRoot {
     onActiveChanged: if (!active) shell.bar = null
     onStatusChanged: {
       if (status === Loader.Error) {
-        var detail = errorString && errorString() ? errorString() : ""
-        console.warn("bar option " + shell.activeBarId + " failed to load, falling back to " + shell.defaultBarId + ":", detail)
-        shell.failedBarId = shell.activeBarId
+        // Record the failure before any diagnostics: activeBarId resolves
+        // back to the default bar off failedBarId, so nothing thrown while
+        // gathering the error detail may block that assignment.
+        var failedId = shell.activeBarId
+        shell.failedBarId = failedId
+        // Loader has no errorString member; an unqualified lookup of an
+        // undefined name throws ReferenceError, so probe with typeof.
+        var detail = typeof errorString === "function" && errorString() ? errorString() : ""
+        if (!detail && sourceComponent) detail = sourceComponent.errorString()
+        console.warn("bar option " + failedId + " failed to load, falling back to " + shell.defaultBarId + ":", detail)
       }
     }
   }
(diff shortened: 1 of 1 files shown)


DeanWahle:
@ryanrhughes Fix for #9116: a failed custom bar option left the session with no bar at all, because the error handler threw a ReferenceError on the nonexistent Loader.errorString before it could set failedBarId and trigger the fallback. Same latent pattern fixed in the panel plugin loader.

PR #9222, opened 2026-08-30
Report every battery in the machine, not just the first

Vegasq:
## The bug

`omarchy-battery-status` picks its battery with `upower -e | grep BAT | head -n 1`, so on a machine with two packs it reports one. ThinkPads with an internal cell plus a hot-swap slice are the common case.

On my X280 the power panel read **5% · 22Wh · cycles 6 · 0W · no time left** while the machine was actually at **11% of 46Wh, drawing 20W, 15 minutes from empty** — the internal pack happened to enumerate first and happened to be idle.

The panel contradicted itself on the same surface, too: the hero percentage comes from this command, but the progress bar under it reads `UPower.displayDevice`, which is already the composite. The text said 5% next to a bar drawn at 11%.

## The fix

Take the headline figures from UPower's `DisplayDevice` — the composite of every pack, which UPower already maintains and which the bar icon and the low-battery service were using all along — and sum the instantaneous sysfs draw across all packs rather than reading one.

Two details do not come from the composite:

- **Charge thresholds** hang off the individual packs; `DisplayDevice` does not carry them, so they are still read from a real battery (falling back to sysfs as before).
- **Cycle counts** are reported per pack (`6 · 12`). Packs wear at their own pace and there is no honest way to collapse two counts into one number.

## Showing both

The status command now also emits each pack behind numbered keys (`batteries`, `battery.1.name`, `battery.1.percentage`, …), and the power panel breaks the composite down into a `BATTERIES` section when there is more than one:

image

Single-battery machines are untouched: one entry means no section, so the panel looks exactly as it does today. An older `omarchy-battery-status` that emits no entries at all is handled the same way, so a half-updated tree degrades rather than breaks.

## See also
[https://github.com/omacom/omarchy/pull/9106](https://github.com/omacom/omarchy/pull/9106)

## Verification

- `test/shell.d/battery-status-test.sh` — the existing single-battery scenario is unchanged and still passes; a new two-battery scenario stubs `upower` with `BAT0`/`BAT1`/`DisplayDevice` and asserts the composite figures, the joined cycle counts, the thresholds staying off the composite device, and every per-pack key.
- `test/shell.d/power-test.sh` — `Model.batteryList` parsing, fraction clamping, the empty case, and the panel bindings.
- `./test/shell` and `./test/cli` pass. Six shell files fail on this machine both with and without the change (`bar-icon-geometry`, `config`, `snapper`, `theme-install-guards`, `unowned-system-paths`, `windows-vm-mount-boundary`); they are environment checks against an unlinked checkout, not regressions here.
- Verified in the running UI on a two-battery ThinkPad: the shell was restarted against this checkout, the panel opened over IPC, and the result screenshotted. Hero percentage and progress bar now agree, capacity reads 46Wh, and both packs appear with correct fills.

### DIFF
--- a/bin/omarchy-battery-status
+++ b/bin/omarchy-battery-status
@@ -5,6 +5,7 @@
 
 shell_output=false
 power_supply_path="${OMARCHY_POWER_SUPPLY_PATH:-/sys/class/power_supply}"
+display_device="/org/freedesktop/UPower/devices/DisplayDevice"
 
 case "${1:-}" in
   "")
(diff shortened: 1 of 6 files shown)


Vegasq:
Closing this as a duplicate — #6845 covers the same ground and covers it better.

I wrote this before scanning the open PRs properly. There are nine of us on this bug, and #6845 is the one with maintainer commits on it and with #7182, #7295 and #7917 already closed in its favour.

For anyone triaging the field later, the two places this PR is concretely worse than #6845, both on the script side:

- **Dead packs.** I take UPower's `DisplayDevice` composite at face value. On the T460s from #7916 — internal cell at 0V/0Wh still advertising its old `energy-full` — that composite reads ~49% while the live slice is at 99%. #6845 splits packs by electrical liveness and rebuilds the headline figures from the live ones.
- **`ENODEV` on `power_now`.** My `$(<"$path/power_now")` returns empty on the HP firmware from #7182/#6895, awk coerces it to `0`, and my `[[ -n $rate ]]` fallback never fires because `"0"` is not empty — so the rate reads `0W`. Inherited from the packaged script rather than introduced here, but #6845 fixes it and this does not.

One thing that may still be worth taking from here, though not from this PR as written: for the panel half, #8904 reads per-pack data from `UPower.devices` in QML rather than through the status command. That composes with #6845 (or any other script-side fix) instead of conflicting with it. This PR invents numbered `battery.N.*` keys in `omarchy-battery-status`, which couples the panel to one particular implementation of the script — the wrong seam, in hindsight.

Verified on a ThinkPad X280 with two live packs (BAT0 22.96Wh / BAT1 23.79Wh) if a data point for that configuration is useful to #6845; most of the reports on this bug involve a dead pack rather than two working ones.

PR #9244, opened 2026-08-30
Install playerctl alongside voxtype so pause_media actually works

### LINKED ISSUE #9243: Voxtype pause_media silently fails: playerctl not installed by omarchy-voxtype-install
## What happened

Voxtype's `pause_media = true` (the default shipped in `$OMARCHY_PATH/default/voxtype/config.toml`) is supposed to auto-pause MPRIS media players while dictating, but it silently does nothing on a fresh `omarchy-voxtype-install` because `playerctl` — the binary Voxtype shells out to for this feature — is never installed.

Voxtype's own log confirms the failure:

```
WARN playerctl not found or failed to run: No such file or directory (os error 2)
```

## What was expected

Starting dictation (e.g. via hotkey) should pause any active MPRIS player (Chromium tab playing Google Music/YouTube Music, Spotify, etc.) since `pause_media = true` ships as the default.

## Steps to reproduce

1. Fresh install: `omarchy-voxtype-install` (installs `wtype voxtype-bin` only, does not install `playerctl`)
2. Play music in a browser tab that exposes MPRIS (confirmed via `dbus-send ... org.freedesktop.DBus.ListNames` showing `org.mpris.MediaPlayer2.chromium.instance...`)
3. Start dictation (hotkey / `voxtype record start`)
4. Music keeps playing; `journalctl --user -u voxtype` shows `WARN playerctl not found or failed to run: No such file or directory (os error 2)`
5. `which playerctl` → not found
6. After `sudo pacman -S playerctl`, the exact same dictation flow correctly pauses/resumes the browser player

## System details

- Omarchy version: 4.0.1-1
- Package: `voxtype-bin` 0.7.5-1 (repo: `omarchy`)
- `voxtype-bin` pacman deps: `alsa-lib curl gcc-libs glibc` — no `playerctl`, not even as an optional dep
- `omarchy-voxtype-install` runs `omarchy-pkg-add wtype voxtype-bin` — also doesn't pull in `playerctl`

## Proposed fix

`playerctl` should be added as an optional (or hard) dependency of `voxtype-bin`, and/or `omarchy-voxtype-install` should install it via `omarchy-pkg-add`, since the shipped default config already turns on `pause_media`. Will follow up with a PR.

montanasnow:
## Summary
- `default/voxtype/config.toml` ships `pause_media = true` by default, but Voxtype implements that by shelling out to `playerctl`, and `omarchy-voxtype-install` never installed it — so the feature silently failed on every fresh install.
- Confirmed via `journalctl --user -u voxtype`: `WARN playerctl not found or failed to run: No such file or directory (os error 2)`.
- Fix: add `playerctl` to the `omarchy-pkg-add` call in `bin/omarchy-voxtype-install`.

## Test plan
- [x] `./test/all` — same 5 pre-existing failures as on `quattro` before this change (all unrelated: missing `omarchy-pkgs` checkout / snapper / theme-install-guards / unowned-system-paths / windows-vm-mount-boundary), none touching voxtype
- [x] Manually installed `playerctl` and confirmed dictation now pauses a Chromium tab's MPRIS player (Google Music) and resumes it after

Fixes #9243

### DIFF
--- a/bin/omarchy-voxtype-install
+++ b/bin/omarchy-voxtype-install
@@ -8,7 +8,7 @@ set -e
 # Install voxtype and configure it for use.
 
 if gum confirm "Install Voxtype + AI model (~150MB) to enable dictation?"; then
-  omarchy-pkg-add wtype voxtype-bin
+  omarchy-pkg-add wtype playerctl voxtype-bin
 
   # Setup voxtype
   mkdir -p ~/.config/voxtype


686f6c61:
`pause_media` talks to MPRIS through playerctl. Quattro dropped playerctl with waybar, and `omarchy-voxtype-install` never put it back, so the shipped `pause_media = true` is a no-op. `omarchy-pkg-add` is the right helper.

One leftover: existing Voxtype installs do not get the package until they re-run the installer.

686f6c61:
Follow-up for existing Voxtype installs: https://github.com/omacom/omarchy/pull/9329 — same installer line as this PR, plus a migration that `omarchy-pkg-add playerctl` when `voxtype-bin` is already present. Happy to drop the installer hunk on rebase if this lands first.

PR #9248, opened 2026-08-30
Add managed account website allowlists

redreceipt:
## Summary

- Add `omarchy managed add|allow|deny|list|remove` and **Setup → Security → Managed Accounts**.
- Create non-admin logins with per-account hostname allowlists enforced by Squid and UID-scoped nftables; administrators stay unrestricted.
- Seed the Omarchy desktop, disable administrator autologin after setup, and preserve the Linux user and home when management is removed.

Initial scope is an administrator recovery account plus one or more managed accounts. Filtering is hostname-based and does not intercept TLS.

## Proof

Johnny starts with only `code.org`; the administrator later adds `scratch.mit.edu`.

https://github.com/user-attachments/assets/ebb13bed-201b-4a46-b87e-4adaab1f287f

## Validation

- ShellCheck, focused shell tests, and CLI metadata checks
- Arch VM end-to-end tests for allowed/blocked HTTPS, bypass prevention, allowlist updates, removal, and reboot persistence


redreceipt:
@codex review this

PR #9264, opened 2026-08-30
Include pahole in the offline ISO package set

ramirlm:
## Summary

Omarchy installations are required to work fully offline. During hardware finalization, installing kernel headers can require `pahole`, but it was not included in the package list used to build the ISO's offline mirror. The missing artifact caused the finalizer to stop even though the package is part of the normal Arch repositories.

## Changes

- Add `pahole` to `install/omarchy-other.packages`, so the ISO builder includes it in the offline package mirror.
- Keep the installer strictly offline; no online fallback is introduced.

## Validation

- `bash -n bin/omarchy-pkg-add`
- `grep -qx 'pahole' install/omarchy-other.packages`
- `bash test/shell.d/synaptic-touchpad-test.sh`
- `bash test/shell.d/update-pacman-guard-test.sh`
- `git diff --check`

### DIFF
--- a/bin/omarchy-pkg-add
+++ b/bin/omarchy-pkg-add
@@ -5,11 +5,42 @@
 # omarchy:examples=omarchy pkg add jq ripgrep
 # omarchy:requires-sudo=true
 
-if omarchy-pkg-missing "$@"; then
+pacman_install() {
   if (( EUID == 0 )); then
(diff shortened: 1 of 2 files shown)


ryanrhughes:
While an interesting solution, it goes against the requirement that the install is to be 100% offline. 

If there's a package missing, please submit a PR to add that to the pkgs list that gets built in.

PR #9276, opened 2026-08-31
Make Apple Magic Trackpad pointer feel closer to macOS

barmstrong:
On Linux, a Magic Trackpad treats a light resting thumb as a second finger. That becomes two-finger scroll or a right-click, and small pointer moves feel jumpy compared with macOS.

Two layers cause it:

- `hid-magicmouse` kernel two-finger scroll and 3-button emulation fire as soon as a second contact exists, before libinput can classify it as a thumb.
- Stock libinput `50-system-apple.quirks` sets `AttrTouchSizeRange=20:10` on these pads, which turns size-based palm/thumb filtering off.

This change:

- Installs `/etc/libinput/omarchy-magic-trackpad.quirks` so a light edge rest is ignored. Harmless on machines without the pad (match is by device name).
- Writes `/etc/modprobe.d/omarchy-magic-trackpad.conf` (`emulate_scroll_wheel=0`, `emulate_3button=0`) when a Magic Trackpad is present, so libinput owns two-finger scroll and click. Two-finger right-click still works through `clickfinger_behavior`.
- Skips those kernel options when an Apple Magic Mouse is also connected. That mouse still needs kernel surface-scroll emulation (same module).
- Pins Hyprland adaptive accel and sensitivity `0` on the common Magic Trackpad device names, so raising the global pointer speed does not make this large pad jumpy.
- Migrates existing installs and applies the kernel params live via sysfs when the module is loaded, so a reboot is not required for that half.

Tested on a USB Magic Trackpad 2 (05AC:0265) on Omarchy 4.0.1. Resting-thumb scroll/right-click stopped, and fine text selection became usable.

`test/shell.d/magic-trackpad-test.sh` covers install and migration for trackpad-only, mouse-present, no-device, and Bluetooth USB-C IDs.

### DIFF
--- a/bin/omarchy-hw-apple-magic-mouse
+++ b/bin/omarchy-hw-apple-magic-mouse
@@ -1,0 +1,14 @@
+#!/bin/bash
+
+# omarchy:summary=Detect an Apple Magic Mouse on USB or Bluetooth.
+
+hid_dir="${OMARCHY_HID_DEVICES_DIR:-/sys/bus/hid/devices}"
+shopt -s nullglob
+for uevent in "$hid_dir"/*/uevent; do
+  # USB 0003 / Bluetooth 0005. Apple 05AC and the later Bluetooth vendor 004C.
+  # Mouse v1 030D, v2 0269, USB-C 0323. Same hid-magicmouse module as the trackpad.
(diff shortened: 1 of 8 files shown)


spuder:
Is this only for the usb version? Looking for a fix for the built in trackpad on macbook air that has same symptom.

barmstrong:
yes it's only for the USB/bluetooth external magic trackpad

PR #9279, opened 2026-08-31
Tag Vivaldi's window class case-insensitively in browser.lua

### LINKED ISSUE #9274: Vivaldi misses the chromium-based-browser tag: browser.lua regex spells the class "Vivaldi-stable", real class is lowercase
## What happened

Vivaldi never receives the `chromium-based-browser` tag, so it falls through the
browser opacity exemption and gets the generic window opacity instead.

`default/hypr/apps/browser.lua:2` spells the class with a capital V:

```lua
o.window("((google-)?[cC]hrom(e|ium)|[bB]rave-browser|[mM]icrosoft-edge|Vivaldi-stable|helium)", { tag = "+chromium-based-browser" })
```

The actual window class is lowercase `vivaldi-stable`. The other alternatives in
the same regex explicitly cover both cases (`[cC]hrom`, `[bB]rave`, `[mM]icrosoft`),
so this looks like an oversight rather than intent.

Consequence: line 4 (`{ tag = "chromium-based-browser" }` -> `-default-opacity`,
`tile = true`, `opacity = "1.0 0.985"`) never applies to Vivaldi. It keeps the
`default-opacity` tag from `default/hypr/windows.lua:6` and ends up at
`opacity = "0.985 0.96"` instead of `"1.0 0.985"`. The difference is small when
focused but clearly noticeable when the window is not focused, and Vivaldi also
misses the `tile = true` the other Chromium browsers get.

## Expected

Vivaldi behaves like Chrome/Brave/Edge: `chromium-based-browser` tag applied,
`default-opacity` removed, `opacity = "1.0 0.985"`, `tile = true`.

## Steps to reproduce

1. Install Vivaldi (Arch package `vivaldi`) and open a window.
2. Inspect the tags Hyprland assigned:

```
$ hyprctl clients -j | jq -r '.[]|select(.class|test("vivaldi";"i"))|"\(.class) -> \(.tags|join(","))"'
vivaldi-stable -> default-opacity*
```

Expected `chromium-based-browser*` in that list; it is absent. For comparison, a
user rule that differs only in the capitalisation of the class does match:

```lua
o.window("vivaldi-stable", { tag = "+probe" })
```

```
vivaldi-stable -> default-opacity*,probe*
```

So the case of the class name is the only difference between matching and not.

## Suggested fix

One character class, consistent with the neighbouring alternatives:

```diff
-o.window("((google-)?[cC]hrom(e|ium)|[bB]rave-browser|[mM]icrosoft-edge|Vivaldi-stable|helium)", { tag = "+chromium-based-browser" })
+o.window("((google-)?[cC]hrom(e|ium)|[bB]rave-browser|[mM]icrosoft-edge|[vV]ivaldi-stable|helium)", { tag = "+chromium-based-browser" })
```

Local workaround for anyone hitting this, in `~/.config/hypr/looknfeel.lua`:

```lua
o.window("vivaldi-stable", { opacity = "1.0 0.985", tile = true })
```

## System details

| | |
|---|---|
| Omarchy | 4.0.0-1 |
| Hyprland | 0.56.2 (efb5099) |
| Kernel | 7.1.8-arch1-3 |
| CPU | Intel Core Ultra 7 258V |
| GPU | Intel Arc Graphics 130V/140V |
| Vivaldi | 8.1.4087.68-1 (Arch `vivaldi`) |

Note: `omarchy debug --no-sudo --print` is rejected as an unknown command on
4.0.0-1, so the system details above were collected manually.

reverb256:
Fixes #9274.

## Problem

Hyprland reports Vivaldi's window class in lowercase as `vivaldi-stable`, but the `chromium-based-browser` rule in `default/hypr/apps/browser.lua` spelled it with a capital V:

```lua
o.window("((google-)?[cC]hrom(e|ium)|[bB]rave-browser|[mM]icrosoft-edge|Vivaldi-stable|helium)", { tag = "+chromium-based-browser" })
```

Because the other alternatives already cover both cases (`[cC]hrom`, `[bB]rave`, `[mM]icrosoft`), Vivaldi never matches the rule, keeps the generic `-default-opacity` tag, and misses the `tile = true` + `opacity = "1.0 0.985"` treatment the other Chromium browsers get.

## Change

Match the class in both cases, consistent with its neighbouring alternatives:

```lua
o.window("((google-)?[cC]hrom(e|ium)|[bB]rave-browser|[mM]icrosoft-edge|[vV]ivaldi-stable|helium)", { tag = "+chromium-based-browser" })
```

## Test

Adds `test/shell.d/vivaldi-window-class-test.sh`, a regression guard that asserts `browser.lua` both tags a `chromium-based-browser` window class and matches Vivaldi case-insensitively (and fails if the case-sensitive spelling returns):

```
ok - browser.lua tags a chromium-based-browser window class
ok - Vivaldi's class is matched in both cases
```

- `bash -n test/shell.d/vivaldi-window-class-test.sh` — clean
- `bash test/shell.d/vivaldi-window-class-test.sh` — passes

Note: the test filename is scoped to Vivaldi to avoid colliding with #8968, which independently adds `test/shell.d/browser-tags-test.sh` for the Brave-origin rule.


686f6c61:
Hyprland reports `vivaldi-stable`, and the neighbouring Chromium matchers already take both cases. `[vV]ivaldi-stable` is the same pattern. The test locks the spelling so the upper-case-only form cannot sneak back.

Read against current `quattro` `default/hypr/apps/browser.lua`.

reverb256:
Went back over this with your note in mind — the matcher now uses [vV]ivaldi-stable and the new test locks both spellings so the upper-case-only form cannot return. Anything else you would like adjusted? Thanks!

PR #9281, opened 2026-08-31
Fit the presentation terminal to its logo instead of a fixed 875x600

londospark:
*Disclosure: this fix was built with AI assistance (Claude Code) -- the investigation, implementation, and this description were largely vibe-coded. I reviewed and tested everything below on my own machine before opening this, but wanted to be upfront about how it was made.*

## Summary

The branded floating terminal used for installs, wifi setup, fingerprint setup, and updates (`omarchy-launch-floating-terminal-with-presentation`) opens at a Hyprland float rule hardcoded to 875x600px (`default/hypr/apps/system.lua`, shared with the generic `floating-window` tag). That size is only wide enough for the 81-column logo at the default terminal font. At a bumped `omarchy display text-size`, the window is too narrow and the logo wraps mid-glyph:

**Before:**
before

**After:**
after

## Why this needs measuring, not a rule

The size that fits the logo depends on the terminal's actual cell metrics, which can only be read from inside a running terminal -- the same constraint `omarchy-launch-about` documents for its own window. I initially tried setting the size declaratively (both a static `o.window` rule in Hyprland config, and a `hl.window_rule` injected via `hyprctl eval` immediately before spawning, mirroring `omarchy-launch-about`'s `apply_size_rule`/`presize_window` pattern). Neither affected the size a new window actually opens at, in any ordering or timing I tried. The only thing that reliably resizes the window is an imperative resize dispatched *after* it exists, so that's what this does, self-relaunching with `--render` the same way `omarchy-launch-about` does.

## What's sized, and how

- **Columns**: the logo's real display width (`wc -L` on `logo.txt`) plus a little padding. Not hardcoded, so a rebranded logo is measured correctly.
- **Rows**: not a flat number, and not derived from the logo's own height either (its 10 lines have nothing to do with how tall this window should be). Instead: scale the pixel width the column target needs at the window's current cell size, apply the *old* 875x600 rule's own width:height ratio to get a pixel height, then convert that back to a row count at the current row height. The window ends up the same shape as the one it replaces, just as wide as the logo actually needs.
- The fitted size is cached (`~/.local/state/omarchy/windows/presentation.fit`) and re-applied via `presize_window`/`apply_size_rule`, mirroring `omarchy-launch-about`'s own caching -- though as noted above, I couldn't get this to actually change the window's opening size, so in practice the live nudge below still runs on every launch. Kept for consistency with the established pattern and in case it behaves differently elsewhere; happy to drop it if that's dead weight.

## A bug caught along the way

The window is matched via a small `own_client()` helper that walks this process's own ancestry (`$$` up through `ps -o ppid=`) to the Hyprland client Hyprland reports for this class, rather than `hyprctl activewindow`. I tried `activewindow` first, matching `omarchy-launch-about`'s pattern, but it silently returns nothing -- skipping the fit entirely -- whenever the popup doesn't hold focus, which happens whenever something else grabs focus while the terminal is still spawning. Caught this because a stray terminal from earlier testing had focus during a re-run and the fit didn't apply at all.

## Testing

- `bash -n bin/omarchy-launch-floating-terminal-with-presentation`
- `test/shell.d/floating-terminal-test.sh` -- passes
- `./test/all` -- no new failures (4 pre-existing failures unrelated to this file: `bar-icon-geometry-test.sh`, `config-test.sh`, `snapper-test.sh`, `unowned-system-paths-test.sh`; confirmed these fail identically on a clean `quattro` checkout)
- Manual verification on a live Hyprland/Omarchy session at both the default text size and a bumped one (`omarchy display text-size 14`): logo renders cleanly at both, final window shape matches the old box's aspect ratio (~1.48 vs. the original 875x600's 1.46), and the fit still applies correctly when the popup doesn't have focus (the bug described above)

## Note on scope

I run a small local Quickshell plugin as a stopgap on my own machine (external to this repo, since I can't presize the window without this fix) -- not part of this PR, just flagging that I've been dogfooding the underlying behavior beyond this one script for a while.

### DIFF
--- a/bin/omarchy-hyprland-window-resize-helpers
+++ b/bin/omarchy-hyprland-window-resize-helpers
@@ -1,0 +1,36 @@
+#!/bin/bash
+
+# omarchy:summary=Shared helpers for nudging a Hyprland window to a measured size (source this, don't run it).
+# omarchy:group=hyprland
+# omarchy:name=window-resize-helpers
+# omarchy:hidden=true
+
+# Both the About and presentation-terminal windows measure their own content
+# from inside the terminal and nudge themselves to fit it. This is the
+# plumbing they share for that: dispatching a resize through Hyprland's Lua
+# API where it exists, falling back to the legacy dispatch where it doesn't,
+# and waiting for the terminal's grid to hold still after either one. What to
+# measure and what size to ask for is each caller's own question.
+
+hypr_dispatch() {
+  local lua="$1"
+  shift
+  hyprctl dispatch "$lua" >/dev/null 2>&1 || hyprctl dispatch "$@" >/dev/null
+}
+
+# Hyprland animates a resize and the terminal reflows to every step of it, so
+# wait for the grid to hold still before measuring it.
+settle_grid() {
+  local current previous="" held=0
+
+  for _ in {1..20}; do
+    current=$(stty size)
+    if [[ $current == $previous ]]; then
+      (( ++held == 3 )) && break
+    else
+      held=0
+      previous=$current
+    fi
+    sleep 0.05
+  done
+}
--- a/bin/omarchy-launch-about
+++ b/bin/omarchy-launch-about
@@ -69,11 +69,7 @@ logo_dimensions() {
   printf '%s %s' "$(display_columns <"$LOGO_FILE")" "$(wc -l <"$LOGO_FILE")"
 }
 
-hypr_dispatch() {
-  local lua="$1"
-  shift
-  hyprctl dispatch "$lua" >/dev/null 2>&1 || hyprctl dispatch "$@" >/dev/null
-}
+source omarchy-hyprland-window-resize-helpers
 
 remember_fit() {
   local directory tmp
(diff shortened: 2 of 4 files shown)


londospark:
@dhh @ryanrhughes — this PR fixes the presentation terminal window sizing issue and makes it responsive to `omarchy display text-size` changes. The window now dynamically sizes to fit the logo instead of using a hardcoded 875x600px.

**What changed:**
- Dynamic window sizing based on logo width + padding
- Preserves the 875x600 aspect ratio (~1.46) regardless of text size
- Bug fix: proper window matching even when the popup doesn't have focus
- Shared window-fit helpers between the about/presentation launchers

**Testing:** passes all tests; manually verified at default and bumped text sizes.

Would love to get this into quattro — any feedback or concerns?

johnpippett:
The user reported a broken logo in the Spotify install popup while it waited for a sudo password. Their screenshot shows the logo wrapped across lines above the password prompt.

I reproduced the logo problem on Omarchy `4.0.4-1` without sudo or a package install:

```sh
omarchy-launch-floating-terminal-with-presentation 'sleep 2; stty size; sleep 10'
```

The live terminal reported `18 63` from `stty size`. The installed `logo.txt` needs 81 columns. Ghostty `1.3.1-2` uses a 12-point font on this system. The active `DP-1` display is 3840 × 2160 at scale `1.25`, and `GDK_SCALE=2`. Hyprland reports an 875 × 600 logical-pixel window.

This confirms the reported logo problem with these settings. I have not installed or tested the change in this PR.

PR #9282, opened 2026-08-31
Point XCompose at a home-local table for sandboxed apps

### LINKED ISSUE #9241: install/user/xcompose.sh writes an absolute /usr/share/omarchy include, breaking every compose sequence inside Steam's container
### System details

- Omarchy 4.0.1-1, Hyprland 0.56.2, libxkbcommon 1.13.2-1, steam 1.0.0.87-3
- AMD Ryzen 7 7800X3D, NVIDIA RTX 4090 (+ Raphael iGPU), Wayland/Hyprland, fcitx5 via `omarchy-fcitx5.service`

### What's wrong

`install/user/xcompose.sh` writes `~/.XCompose` with an **absolute** include into the packaged tree:

```bash
tee ~/.XCompose >/dev/null <<EOF
# Run omarchy-restart-xcompose to apply changes

# Include fast emoji access
include "/usr/share/omarchy/default/xcompose"
...
EOF
```

Steam runs games inside pressure-vessel, a container with its own `/usr`. `$HOME` is bind-mounted into it, so the container reads `~/.XCompose` — but `/usr/share/omarchy` does not exist there, so the include cannot resolve.

A failed include does not degrade gracefully. It aborts the whole compose table:

```
$ xkbcli compile-compose --file /tmp/bad.compose
xkbcommon: ERROR: (unknown file):1:9: failed to open included Compose file "/usr/share/omarchy/does-not-exist"
xkbcommon: ERROR: (unknown file):1:9: failed to parse file
ERROR: Couldn't create compose from file: /tmp/bad.compose
$ echo $?
1
```

So inside Steam and every Proton game, **no** compose sequence works — not the Omarchy emoji shortcuts, not the identity macros, not the `%L` locale defaults. Each affected process also logs the failure to the journal:

```
steam[80717]: xkbcommon: ERROR: /home/<user>/.XCompose:4:9: failed to open included Compose file
              "/usr/share/omarchy/default/xcompose": No such file or directory
steam[80717]: xkbcommon: ERROR: /home/<user>/.XCompose:4:9: failed to parse file
```

Confirmed the container genuinely cannot see the path, while it can see the file that references it:

```
$ ls /proc/$(pgrep -f SomeGame.exe)/root/usr/share/omarchy/default/xcompose
ls: cannot access '...': No such file or directory
$ ls /proc/$(pgrep -f SomeGame.exe)/root/home/<user>/.XCompose
-rw-r--r-- 1 <user> <user> 232 Aug 30 15:18 /proc/.../.XCompose
```

The same reasoning applies to any sandboxed app that sees `$HOME` but not the host `/usr` — Flatpaks, for instance — though I only verified Steam/pressure-vessel.

### This is a regression from Omarchy 3

Per #8173, Omarchy 3 wrote:

```
include "%H/.local/share/omarchy/default/xcompose"
```

`%H` expands to `$HOME`, which **is** bind-mounted into the container, so the Omarchy 3 form resolved correctly inside Steam. Moving the include to an absolute `/usr/share/omarchy` path is what broke it.

Worth flagging against #8175, which proposes repointing `~/.XCompose` at the packaged tree to fix the quattro upgrade path: as written that would give upgraded installs the same absolute include, and with it this bug.

### Steps to reproduce

1. Fresh Omarchy 4 install (or any install whose `~/.XCompose` has the absolute include).
2. Launch any Steam game, native or Proton.
3. Try a compose sequence in-game, e.g. CapsLock `m` `s`. Nothing is produced.
4. `journalctl --user -b | grep xkbcommon` shows the parse errors above.

### Suggested fix

Make the include home-relative so it resolves inside containers, writing a copy into `$HOME` at install time:

```bash
cp /usr/share/omarchy/default/xcompose ~/.XCompose.omarchy
# in ~/.XCompose:
include "%H/.XCompose.omarchy"
```

That keeps the packaged file authoritative at install/update time while leaving `~/.XCompose` self-contained from a container's point of view. It needs a re-sync step on update, which the current absolute include avoids — so if keeping the packaged path is preferred, the alternative is to accept that compose is simply unavailable in sandboxed apps and drop the include from the generated file rather than emit one that fails to parse.

### Note

I found this while chasing an unrelated missing-keyboard-input problem in a Proton game. Fixing this did **not** resolve that, and I want to be clear I have no evidence the broken include causes loss of key input — the verified impact is dead compose sequences plus journal error spam in every Steam/Proton process.

fresh3nough:
## Summary

Fixes #9241.

`install/user/xcompose.sh` wrote `include "/usr/share/omarchy/default/xcompose"` into `~/.XCompose`. Steam's pressure-vessel (and other sandboxes that bind `$HOME` but not host `/usr`) cannot open that path. xkbcommon does not degrade on a failed include — it aborts the entire compose table — so every sequence dies inside the container (emoji, identity macros, locale defaults), with journal spam on every game process.

Omarchy 3 used a `%H`-relative include. This restores a container-safe shape without depending on the retired `~/.local/share/omarchy` checkout symlink.

## Change

- Seed `~/.XCompose.omarchy` from `$OMARCHY_PATH/default/xcompose` and include `"%H/.XCompose.omarchy"` from install, upgrade-to-quattro, and `omarchy-restart-xcompose`.
- New `omarchy-refresh-xcompose` keeps the home copy current when the packaged table is newer; `omarchy-fcitx5.service` runs it as `ExecStartPre` so the normal login/update path refreshes without a migration.
- Migration `1788138200.sh` rewrites existing absolute `/usr/share/omarchy/...` and Omarchy 3 `%H/.local/share/omarchy/...` includes.
- Adjust migration `1788102906.sh` so its legacy rewrite targets the home-local include instead of an absolute packaged path.
- Neither migration restarts fcitx5 (#9541 / #9568); the next graphical login loads the repaired files. Manual `omarchy-restart-xcompose` keeps live-reload semantics.

## Test plan

- [x] Reproduced on GCP VM + Arch container: absolute missing include → `xkbcli compile-compose` fails
- [x] After fix: `%H/.XCompose.omarchy` under `$HOME` compiles and emits sequences
- [x] Migration rewrites absolute include and preserves user sequences without restarting fcitx5
- [x] Refresh copies a newer packaged table into `~/.XCompose.omarchy`; fcitx5 unit has `ExecStartPre`
- [x] `./test/shell.d/xcompose-home-include-test.sh` passes
- [x] `./test/shell.d/legacy-power-udev-rules-migration-test.sh` passes

```bash
./test/shell.d/xcompose-home-include-test.sh
./test/shell.d/legacy-power-udev-rules-migration-test.sh
```

### DIFF
--- a/bin/omarchy-restart-xcompose
+++ b/bin/omarchy-restart-xcompose
@@ -2,6 +2,15 @@
 
 # omarchy:summary=Restart the XCompose input method service (fcitx5) to apply new compose key settings.
 
+# Refresh the home-local emoji table from the packaged tree so package updates
+# reach sandboxed apps that include %H/.XCompose.omarchy rather than /usr.
+packaged_xcompose="${OMARCHY_PATH}/default/xcompose"
+home_xcompose="$HOME/.XCompose.omarchy"
+if [[ -f $packaged_xcompose ]]; then
+  cp "$packaged_xcompose" "$home_xcompose"
+  chmod 644 "$home_xcompose"
+fi
+
 # A fcitx5 running outside the unit owns org.fcitx.Fcitx5, so the unit's
 # instance would exit on arrival and the stale process would keep serving the
 # old ~/.XCompose -- a restart that reports success and changes nothing.
--- a/bin/omarchy-upgrade-to-quattro
+++ b/bin/omarchy-upgrade-to-quattro
@@ -2020,11 +2020,18 @@ if [[ -n ${OMARCHY_USER_EMAIL//[[:space:]]/} ]]; then
   git config --global user.email "$OMARCHY_USER_EMAIL" || true
 fi
 if [[ ! -f $HOME/.XCompose ]]; then
+  # Home-local copy so sandboxed apps that see $HOME but not /usr still parse
+  # the compose table (Steam pressure-vessel, Flatpaks). Absolute
+  # /usr/share/omarchy includes abort the whole file inside those containers.
+  if [[ -f $OMARCHY_PATH/default/xcompose ]]; then
+    cp "$OMARCHY_PATH/default/xcompose" "$HOME/.XCompose.omarchy"
+    chmod 644 "$HOME/.XCompose.omarchy"
+  fi
   cat >"$HOME/.XCompose" <<EOF
 # Run omarchy-restart-xcompose to apply changes
 
 # Include fast emoji access
-include "/usr/share/omarchy/default/xcompose"
+include "%H/.XCompose.omarchy"
 
 # Identification
 <Multi_key> <space> <n> : "$OMARCHY_USER_NAME"
--- a/install/user/xcompose.sh
+++ b/install/user/xcompose.sh
@@ -1,9 +1,24 @@
-# Set default XCompose that is triggered with CapsLock
+# Set default XCompose that is triggered with CapsLock.
+#
+# The emoji table ships under $OMARCHY_PATH, but sandboxed apps (Steam's
+# pressure-vessel, Flatpaks) bind-mount $HOME and not the host /usr. An
+# absolute include into /usr/share/omarchy fails to open there, and xkbcommon
+# aborts the whole compose file — every sequence dies, not only the Omarchy
+# ones. Keep a home-local copy and include it with %H so containers that see
+# $HOME still parse the table.
+packaged_xcompose="${OMARCHY_PATH}/default/xcompose"
+home_xcompose="$HOME/.XCompose.omarchy"
+
+if [[ -f $packaged_xcompose ]]; then
+  cp "$packaged_xcompose" "$home_xcompose"
+  chmod 644 "$home_xcompose"
+fi
+
 tee ~/.XCompose >/dev/null <<EOF
 # Run omarchy-restart-xcompose to apply changes
 
 # Include fast emoji access
-include "/usr/share/omarchy/default/xcompose"
+include "%H/.XCompose.omarchy"
 
 # Identification
 <Multi_key> <space> <n> : "$OMARCHY_USER_NAME"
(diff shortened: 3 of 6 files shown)


Marjinoz:
Thanks for picking this up, it matches what I ended up doing locally.

One gap: nothing in the normal update path calls `omarchy-restart-xcompose`, so `~/.XCompose.omarchy` only refreshes when a migration runs or the user runs the restart by hand. After the next package update that touches `default/xcompose`, everyone stays on the stale copy without noticing.

Suggestion: refresh the copy where the fcitx5 unit starts (e.g. an `ExecStartPre` that copies when the packaged file is newer), or from the post-update hook, rather than only inside the restart command.

Also, given #9541 / #9568, the new migration should probably drop its `omarchy-restart-xcompose` call and let the next login pick the file up.

fresh3nough:
Addressed the two follow-ups from @Marjinoz / @cking-bot:

1. **Normal update path refresh** — added `omarchy-refresh-xcompose` and wired it as `ExecStartPre=` on `omarchy-fcitx5.service`. After a package update that touches `default/xcompose`, the next graphical login (when fcitx5 starts) refreshes `~/.XCompose.omarchy` if the packaged file is newer. Install, upgrade-to-quattro, restart, and the migrations all go through the same helper.

2. **No migration restart** — dropped `omarchy-restart-xcompose` from both `1788138200.sh` and `1788102906.sh` so unattended repairs cannot invalidate XIM objects held by running X11 clients (#9541). Manual restart is unchanged for an intentional live reload.

```
bash test/shell.d/xcompose-home-include-test.sh
ok - install/user/xcompose.sh seeds a %H-relative home-local include
ok - migration rewrites absolute XCompose includes without restarting fcitx5
ok - refresh keeps ~/.XCompose.omarchy current on the fcitx5 start path
ok - xkbcli accepts home-local includes and rejects missing absolute ones

bash test/shell.d/legacy-power-udev-rules-migration-test.sh
# still green, including no-restart asserts on the XCompose rewrite
```

PR #9287, opened 2026-08-31
Fix invisible selection color when a theme has no [colors.selection]

### LINKED ISSUE #9266: Cloned theme without [colors.selection] gets selection == foreground (invisible highlights)

perryqh:
## Screenshots

**Before** — theme with no `[colors.selection]`: selection background equals the foreground color, so selected text is nearly unreadable.
selection-before

**After** — falls back to `color8` instead, so the selection is clearly visible.
selection-after

## Summary
- `omarchy-theme-colors-from-alacritty` fell back to `foreground` for `selection_background` when a theme's `alacritty.toml` had no `[colors.selection]` block, making selections/highlights identical to text color and effectively invisible (Neovim visual mode, gum confirm/choose prompts, btop's selected row). Now falls back to `color8` (bright black, itself falling back to normal black), matching the fallback chain `omarchy-theme-color` already uses elsewhere.
- `omarchy-theme-set` now also repairs that same fingerprint on the staged copy when an extra/git theme's repo already committed a `colors.toml` carrying it — theme-set skips regeneration once `colors.toml` exists, so such a theme would otherwise recopy the stale value forever no matter how many times it's refreshed. Only the staged copy is rewritten; the user's own theme source (which may be a git clone or a symlink into personal dotfiles) is left untouched, so the repair harmlessly repeats on every switch of an affected theme.
- Adds a migration so a theme already active before this fix gets refreshed automatically on next update. It also guards against a current theme that was later deleted: blindly refreshing it would hit `omarchy-theme-set`'s "Theme does not exist" and hard-fail under `omarchy-migrate`'s `set -euo pipefail`, stalling every migration queued behind it — so it seeds "Tokyo Night" instead, mirroring the existing `1787481315.sh` migration.

Fixes #9266

## Test plan
- [x] `test/cli`: a theme with `[colors.bright]` but no `[colors.selection]`, and one with neither, never get `selection == foreground` in the generated `colors.toml`
- [x] `test/shell.d/theme-staging-test.sh`: an extra/git theme with a baked-in stale `colors.toml` is repaired to `color8` (falling back to `color0`) on the staged copy, with the source left untouched; a healthy or official palette is never rewritten
- [x] `test/shell.d/theme-selection-color-migration-test.sh`: the migration refreshes an affected theme, leaves a healthy one alone, no-ops with no current theme, and — run against the real `omarchy-migrate` — does not stall the migration queue when the current theme was removed or `theme.name` is missing
- [x] `./test/cli` and `./test/shell` pass
- [x] `./test/all` shows no new failures (4 pre-existing failures on `quattro` unrelated to this change, reproduced independently with these commits removed: `config-test.sh`, `snapper-test.sh`, `theme-install-guards-test.sh`, `unowned-system-paths-test.sh`)
- [x] Verified visually on a live desktop: applied a scratch theme with an `alacritty.toml` lacking `[colors.selection]`, confirmed the generated `selection` differed from `foreground`, and checked Alacritty/Neovim/gum/btop selection highlights were visible before reverting to the original theme

### DIFF
--- a/bin/omarchy-theme-colors-from-alacritty
+++ b/bin/omarchy-theme-colors-from-alacritty
@@ -120,7 +120,7 @@ background=${background:-$color0}
 foreground=${foreground:-$color7}
 color0=$background
 color7=$foreground
-selection_background=${selection_background:-$foreground}
+selection_background=${selection_background:-$color8}
 accent=$color4
 
 mkdir -p "$THEME_SOURCE"
--- a/bin/omarchy-theme-set
+++ b/bin/omarchy-theme-set
@@ -174,6 +174,30 @@ stage_installed_colors_from_alacritty() {
   rm -rf "$scratch"
 }
 
+# Old omarchy-theme-colors-from-alacritty wrote selection = foreground when
+# alacritty.toml had no [colors.selection], which is invisible in Neovim, gum,
+# and btop. Theme-set skips regeneration when colors.toml already exists, so an
+# extra/git theme that committed that fingerprint would be recopied forever.
+# Rewrite only the generated `selection = "#hex"` line, and only when it equals
+# foreground, to color8 (bright black, falling back to normal black).
+repair_stale_generated_selection() {
+  local file="$1"
+  local selection foreground color8 color0 replacement
+
+  [[ -f $file && ! -L $file ]] || return 0
+
+  selection=$(awk -F'"' '/^selection = /{print $2; exit}' "$file")
+  foreground=$(awk -F'"' '/^foreground = /{print $2; exit}' "$file")
+  [[ -n $selection && $selection == "$foreground" ]] || return 0
+
+  color8=$(awk -F'"' '/^color8 = /{print $2; exit}' "$file")
+  color0=$(awk -F'"' '/^color0 = /{print $2; exit}' "$file")
+  replacement=${color8:-$color0}
+  [[ -n $replacement && $replacement != "$foreground" ]] || return 0
+
+  sed -i "s|^selection = \"$selection\"$|selection = \"$replacement\"|" "$file"
+}
+
 # Copies a directory without ever following a symlink: in an installed theme one
 # points wherever the theme author chose, which is how an unlock.png becomes a
 # copy of any file the session can read.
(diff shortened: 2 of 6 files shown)


bjarneo:
Automated duplication check: this pull request looks similar to #9270, which covers the same theme selection color fallback. I keep that one open and close this one to consolidate review.

PR #9312, opened 2026-08-31
Add responsive theme background support

acrogenesis:
## Summary

- add per-background metadata for fill mode, fill color, focal point, responsive SVG layout, and fitted-image backdrops
- preserve full compositions over either a cached sampled-edge color or a quiet blurred cover copy with `backdrop = "edge"` / `"blur"`; existing themes keep the historical solid behavior by default
- select same-stem aspect-ratio variants per display while keeping variants out of the picker and background cycle
- share background resolution across the desktop, lock screen, picker, and bar color detection
- document the theme format and credit Thomas Feichtinger's MIT-licensed Background Display plugin for the backdrop concepts
- cover the resolver, per-screen sampling, transition metadata, desktop renderer, and lock-screen parity with CLI and shell tests

## Security review

- treat installed theme images, SVG, and `backgrounds.toml` as untrusted data; no theme value is reconstructed as shell source or crosses a privilege boundary
- bound SVG rasterization to 16,384 pixels per dimension and 32 Mi pixels, and time-limit SVG, ImageMagick, and thumbnail conversion
- verify hostile intrinsic dimensions still produce a bounded cached raster, parent-directory image references remain inaccessible, malformed metadata fails to safe defaults, and option-shaped or control-character values cannot alter command or output framing
- verify every SVG in the stacked asset PR is XML-valid, self-contained, and free of scripts, event handlers, and external references

## Verification

- post-hardening: all 42 responsive SVG renders succeeded at 1920x1080 and 5120x1440; live Tokyo Night resolution and Hyprland config remained clean
- post-hardening aggregate rerun reached the changed background suites cleanly; its one reproducible unrelated failure is `agent-usage-update-test.sh` missing its `good.json` fixture
- `./test/cli`
- `bash test/shell.d/background-resolve-test.sh`
- `bash test/shell.d/background-test.sh`
- `./test/shell` exercised all 223 files; the changed background and runtime coverage passed. Unrelated aggregate failures were limited to sibling-checkout packaging coverage (`omarchy-pkgs` / absent `omarchy-iso`, plus an existing `omarchy-crash-watch.service` PKGBUILD mismatch) and the session's global `NO_COLOR=1`; `launch-about-test.sh` passed when rerun without that variable.
- visually verified sampled-edge and blurred backdrops through the live shell on the 5120x1440 display
- sampled-edge verification selected Catppuccin's actual `#181826` perimeter instead of a deliberately incorrect fallback color
- blurred verification preserved the complete Gruvbox painting over a subdued ambient cover
- restored Tokyo Night, `0-winding-road.webp`, workspace 1, and the normal checkout shell after verification


acrogenesis:
Pairs well with https://github.com/omacom/omarchy/pull/8720

OldJobobo:
Really impressive work here. I like the shared resolver approach, the backward compatibility, and the care given to security and multi-monitor behavior.

My only hesitation is the amount of custom machinery being introduced around metadata parsing, SVG processing, caching, and transition state. The AWK parser appears to support a restricted subset of TOML, while the responsive SVG handling rewrites XML attributes using sed. I wonder whether those boundaries should be documented more explicitly, or whether any of that could use existing parsers to reduce long-term maintenance risk.

I’m also curious about cache invalidation for SVGs that reference sibling assets, since the raster-cache key appears to include the SVG’s mtime but not the mtimes of its dependencies. A cache size limit might also be worth considering alongside the existing age-based cleanup.

None of these look like obvious blockers, and I may be missing some context. They’re mainly areas where I’d be interested in the maintainers’ perspective on complexity and long-term maintenance. Overall, this looks like a thoughtful and valuable improvement.

*Pi x GPT on behalf of OldJobobo.*

acrogenesis:
Thanks for the careful read — all four points are fair, and I've pushed changes for the concrete ones (a61e4ad, ce3ccf81):

- **Dependency cache invalidation:** good catch. Responsive SVGs pull in relative sibling assets at render time, so keying the raster cache only on the SVG's own mtime was a real staleness gap. The key now folds in a digest of the sibling files' names and mtimes, so editing a referenced asset re-rasterizes even when the SVG itself is untouched. The tradeoff — any change in a `backgrounds/` directory can invalidate that directory's SVG renders — is documented; those directories are small, so it's cheap.
- **Cache size:** added a count cap alongside the existing 30-day age prune, on both the render cache and the sampled-edge-color cache, so neither grows without bound.
- **TOML subset / sed-rewritten SVG:** both are deliberate zero-dependency choices — Omarchy's base package set ships no TOML or XML toolchain, so metadata is a small awk subset (`[section]` headers bare or quoted, `key = value`, `#` comments; no arrays, nested tables, or multi-line values) and the responsive `viewBox` rewrite stays in `sed`. I've documented the supported subset and the responsive-SVG root-attribute expectations in `docs/theming.md` so the boundaries are explicit.

On replacing the hand-rolled parsing with real parsers to cut long-term maintenance: that's genuinely a maintainers' call on the complexity-vs-dependency tradeoff, so I'll defer to their preference rather than pull a new runtime dependency into the base image. Appreciate the thoughtful review.

Geokec:
Seconding this — it solves a real ultrawide pain point.

I'm on a 3440x1440 display, and the Giants theme ships 3:2 photographic portraits. Today they get hard-cropped to cover, which loses most of each composition. As a stopgap I hand-re-encoded every background to 3440x1440 with the portrait fitted to screen height and the sidebars filled with the average color sampled from the image's own edges — which is essentially the `fill = "fit"` + `backdrop = "edge"` behavior this PR adds, done manually and per-image.

Having this as a first-party, per-theme (and eventually per-image) preference is exactly the right fix. The sampled-edge backdrop in particular produced seamless results on textured photographic images, so I can confirm the approach works well in practice. Happy to test on ultrawide once this lands.

PR #9316, opened 2026-08-31
Stop factory-reset from sudoing the path that started it

Adolanium:
`omarchy-system-factory-reset` raises itself like this:

```
if (( EUID != 0 )); then
  mapfile -t gum_env < <(env | grep '^GUM_' || true)
  exec sudo env "${gum_env[@]}" "$0" "$@"
fi
```

`$0` is whichever path started the script. On bash that is the full path even when the command was found by name on PATH. A file of the same name in a directory the user can write, sitting in front of `/usr/bin`, is then what sudo runs as root. `omarchy-dns` and `omarchy-theme-set-browser-policy` already avoid this by re-execing a fixed `/usr/bin/...` path. Factory-reset did not.

The prompt still asks for a password. The bug is that the password is for a path the caller can write, not for the packaged helper.

What changed:

- The privileged half always runs `/usr/bin/omarchy-system-factory-reset`.
- Gum theme vars are still forwarded the same way.

Tests: new `test/shell.d/factory-reset-packaged-path-test.sh`. A source check pins the packaged path. A writable copy, and a PATH lookup of that copy, both ask sudo to run `/usr/bin/omarchy-system-factory-reset` rather than the copy. Gum vars still ride along. All 4 pass. The elevation cases stub sudo, so nothing here starts a real reset.

### DIFF
--- a/bin/omarchy-system-factory-reset
+++ b/bin/omarchy-system-factory-reset
@@ -34,12 +34,17 @@ TOP_MNT=/run/omarchy-system-factory-reset/top
 NEXT_NAME=@omarchy-reset-next
 LOG_FILE=/var/log/omarchy-system-factory-reset.log
 
+# The path etc/sudoers.d names for other helpers, and the path this must
+# elevate even when a copy on PATH started the script. sudo runs this path,
+# not $0, so a file the user can write cannot become the root half.
+PACKAGED_PATH=/usr/bin/omarchy-system-factory-reset
+
 # Self-elevate so the menu entry (and a bare shell invocation) need no sudo
 # prefix. The caller's gum theme vars are forwarded as `env` arguments rather
 # than via `sudo -E`, so styling survives even under an env_reset sudoers.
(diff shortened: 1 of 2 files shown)


Adolanium:

The copy of the script that ends up in `$0` is already running as the user before sudo is called. It can call `sudo` on anything itself, or shadow `sudo` and take the password. Pinning the re-exec to `/usr/bin` only stops an attacker who chooses to go through the real script's re-exec line. There is no NOPASSWD rule for factory-reset, so the caller already has wheel and a password.

The change also breaks the dev-link case on purpose: `omarchy-dev-link` writes `secure_path` so that root runs the checkout copy, and this would make edits to factory-reset in a checkout silently run the packaged copy instead.

The `omarchy-dns` and browser-policy comparison in the description was also wrong. Those pin a packaged path because their NOPASSWD sudoers rules match an exact `/usr/bin` path, not for `$0` hygiene.

PR #9317, opened 2026-08-31
Stop privileged copies from following a caller-controlled OMARCHY_PATH

Adolanium:
`omarchy-refresh-limine`, `omarchy-refresh-pacman`, `omarchy-toggle-hybrid-gpu`, and `omarchy-hibernation-setup` expand `$OMARCHY_PATH` as the user, then `sudo cp` that path onto a root-owned target:

```
sudo cp "$OMARCHY_PATH/default/limine/limine.conf" /boot/limine.conf
sudo cp -f "$OMARCHY_PATH/default/pacman/pacman-$channel.conf" /etc/pacman.conf
sudo cp -p "$OMARCHY_PATH/default/systemd/system-sleep/force-igpu" /usr/lib/systemd/system-sleep/
sudo cp -p "$OMARCHY_PATH/default/systemd/system-sleep/keyboard-backlight" /usr/lib/systemd/system-sleep/
```

A caller who sets `OMARCHY_PATH` to a tree they can write, then types the sudo password for "refresh Limine" or "refresh pacman", installs their file as the bootloader config, the package config, or a sleep hook systemd later runs as root. `omarchy-plymouth-set` already refuses a user-owned source tree unless a root-owned `/etc/omarchy.conf` names that checkout. These copies did not.

The prompt still asks for a password. The bug is that the copy source is a path the caller chose.

What changed:

- All four helpers copy from `/usr/share/omarchy` by default.
- A linked checkout is used only when `/etc/omarchy.conf` is a root-owned regular file whose only line names that checkout, which is what `omarchy-dev-link` writes. Same rule as plymouth-set.

Tests: new `test/shell.d/privileged-source-copy-test.sh`. A source check pins every remaining sudo-copy. With a fake tree in `OMARCHY_PATH`, refresh-limine and refresh-pacman still ask sudo to copy from `/usr/share/omarchy`. All 3 pass. The elevation cases stub sudo, so nothing here writes `/boot` or `/etc`.

### DIFF
--- a/bin/omarchy-hibernation-setup
+++ b/bin/omarchy-hibernation-setup
@@ -4,6 +4,56 @@
 # omarchy:requires-sudo=true
 # omarchy:args=[--force] [--no-rebuild]
 
+# Privileged copies must not follow a caller-controlled OMARCHY_PATH. The
+# packaged tree is always trusted. A linked checkout is trusted only when
+# /etc/omarchy.conf is a root-owned regular file whose only line names that
+# checkout, which is what omarchy-dev-link writes. omarchy-plymouth-set uses
+# the same rule. Keep this block in step with omarchy-refresh-limine,
+# omarchy-refresh-pacman, and omarchy-toggle-hybrid-gpu.
+omarchy_privileged_source_root() {
+  local packaged=/usr/share/omarchy
+  local path=${OMARCHY_PATH:-$packaged}
+  local conf=/etc/omarchy.conf
+  local uid mode
+  local -a lines
+
+  if [[ $path == "$packaged" ]]; then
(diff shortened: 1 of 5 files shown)


Adolanium:

The caller who sets `OMARCHY_PATH` is the same wheel user who types the sudo password. That user can already `sudo cp` anything into `/boot` or `/etc/pacman.conf` directly. No NOPASSWD rule covers these four commands, so nothing is gained by redirecting the copy source.

There may be a case for a shared "resolve the trusted tree" helper for consistency with `omarchy-plymouth-set`, but that would be a refactor with one shared function and dev-link quoting handled properly, not a security fix. I will open that separately if there is interest.

PR #9319, opened 2026-08-31
Give the Secret portal a provider so Chromium can open its password store

wombatoperator:
## The symptom

Chrome on a fresh Omarchy greets the user with:

> Something went wrong when opening your profile. Some features may be unavailable.

## The cause

Chromium asks xdg-desktop-portal for `org.freedesktop.impl.portal.Secret` to get the key it encrypts saved passwords and cookies with. The only backend implementing that interface is `gnome-keyring.portal`, and it ships gated:

```ini
Interfaces=org.freedesktop.impl.portal.Secret
UseIn=gnome
```

The session is Hyprland, and `hyprland-portals.conf` prefers `hyprland;gtk` — neither of which implements Secret. So nothing answers, and Chromium records its own failure in `~/.config/google-chrome/Local State`:

```json
"os_crypt": {"portal": {"prev_desktop": "Hyprland", "prev_init_success": false}}
```

Migration `1784508556` already found this half of it —

> Chromium-based browsers auto-detect their os_crypt backend; on Hyprland the xdg-desktop-portal Secret backend has no provider and fails

— and pinned `--password-store=gnome-libsecret` so the backend stopped changing underneath people, which is what was costing them their cookies. This supplies the provider it was asking for. `gnome-keyring` is already in `omarchy-base.packages` and is already running as the `org.freedesktop.secrets` service; it was simply never wired to the portal.

## Measured

Omarchy 4.0.1-1, Hyprland 0.56.2, Chrome 141.

| | Secret portal | `prev_init_success` |
|---|---|---|
| before | no reply | `false` |
| after | `v u 1` | `true` |

```bash
busctl --user get-property org.freedesktop.portal.Desktop \
  /org/freedesktop/portal/desktop org.freedesktop.portal.Secret version
```

The secrets service itself was never broken — `secret-tool store` / `lookup` round-trips correctly either way. Only the portal route was unserved.

## The change

- `install/config/secret-portal.sh` for fresh installs, wired into `install/config/all.sh`
- a migration for existing machines

Both write only when `/etc/xdg-desktop-portal/portals.conf` is absent: a file already there is the user's, and changing the Secret assignment under them could move where existing secrets are read from. A config at `/etc` replaces the desktop-specific `hyprland-portals.conf` rather than merging with it, so the Hyprland defaults are restated alongside the new line.

## Caveat

Measurements above were taken with the same content at `~/.config/xdg-desktop-portal/portals.conf`, which takes precedence over `/etc`. The file contents and the resulting portal resolution are identical; only the path differs.

### DIFF
--- a/install/config/all.sh
+++ b/install/config/all.sh
@@ -1,5 +1,6 @@
 run_logged "$OMARCHY_INSTALL/config/theme-system.sh"
 run_logged "$OMARCHY_INSTALL/config/browser-policy.sh"
+run_logged "$OMARCHY_INSTALL/config/secret-portal.sh"
 run_logged "$OMARCHY_INSTALL/config/increase-lockout-limit.sh"
 run_logged "$OMARCHY_INSTALL/config/lockscreen-pam.sh"
 run_logged "$OMARCHY_INSTALL/config/fix-powerprofilesctl-shebang.sh"
--- a/install/config/secret-portal.sh
+++ b/install/config/secret-portal.sh
@@ -1,0 +1,26 @@
+# Chromium asks xdg-desktop-portal for org.freedesktop.impl.portal.Secret to get
+# the key it encrypts saved passwords and cookies with. The only backend that
+# implements Secret is gnome-keyring.portal, and it ships gated `UseIn=gnome`,
+# so in a Hyprland session nothing answers: the browser records
+# os_crypt.portal.prev_init_success = false and greets the user with "Something
+# went wrong when opening your profile. Some features may be unavailable."
+#
+# Migration 1784508556 already noted that "on Hyprland the xdg-desktop-portal
+# Secret backend has no provider and fails" and pinned --password-store so the
+# backend at least stops changing underneath people. This gives it the provider,
+# which is the other half: gnome-keyring is already a base package and is
+# running as the secrets service, it was simply never wired to the portal.
+#
+# A file at /etc replaces the desktop-specific hyprland-portals.conf rather than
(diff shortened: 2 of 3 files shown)


ml-tx:
This also fixes `slack-desktop`, which is worth noting because no `*-flags.conf` pin can reach it. That mechanism is Arch's wrapper script, not Electron itself — `/usr/bin/electron44` is bash that sources `~/.config/electron44-flags.conf` before exec'ing. `slack-desktop` bundles its own Electron 44 (`/usr/lib/slack/version`, with `libffmpeg.so` and `resources/app.asar` owned by the package, and no electron dependency), and `/usr/bin/slack` symlinks straight to the stripped ELF binary with no wrapper in between. So there is no flags file it will ever read, and routing the Secret portal is the only thing that helps it.

Before, on Hyprland 0.56.2 / slack-desktop 4.52.155, `~/.config/Slack/Local State` had no key and recorded the same failure you documented:

```json
{"os_crypt":{"portal":{"prev_desktop":"Hyprland","prev_init_success":false}}}
```

With no key the cookie store cannot be persisted, so Slack drops the workspace on every launch:

```
[AUTH] credentials are ready ({"removed":["T0XXXXXXX"],...})
[LOCAL-CONFIG] Removing expired team from localConfig teams: T0XXXXXXX
[AUTH] Not signed into any teams
```

After routing `org.freedesktop.impl.portal.Secret` to gnome-keyring, `prev_init_success` flips to `true`, an `Application key for slack` item appears in the default keyring, and the login survives a restart.

PR #9322, opened 2026-08-31
Clear special permission bits during Windows VM migration

pyrozhok:
## Problem

Existing Windows VM installs can fail to start when the shared directory has the setgid bit set, for example mode 2700. The migration hardens the directory with numeric `chmod 0700`, but numeric modes leave unspecified special bits unchanged. The subsequent exact-mode check sees 2700 instead of 700 and aborts the migration, causing the Quickshell Windows launcher to report only that startup failed.

## Solution

Use an explicit symbolic mode that sets owner permissions, removes group/other permissions, and clears all special bits before validating the mount sources. Apply the same hardening to both the user-side and privileged migration paths.

## Testing

- `bash test/shell.d/windows-vm-compose-test.sh`
- `./test/all` (the existing environment-dependent failures remain in `config-test.sh`, `snapper-test.sh`, `unowned-system-paths-test.sh`, and `windows-vm-mount-boundary-test.sh`)

The focused Windows VM test now covers migration from a setgid shared directory.

### DIFF
--- a/bin/omarchy-windows-vm
+++ b/bin/omarchy-windows-vm
@@ -580,7 +580,8 @@ prepare_caller_mounts() {
   # Privacy is an explicit preflight step for both already-pinned sources, not
   # a side effect halfway through the two-mount transaction. Old umask-022
   # installs are hardened together before either Docker-facing anchor changes.
-  chmod 0700 -- "/proc/$BASHPID/fd/$storage_fd" "/proc/$BASHPID/fd/$shared_fd" || {
+  # Clear special bits explicitly: numeric chmod modes leave them unchanged.
+  chmod u=rwx,go=,a-s -- "/proc/$BASHPID/fd/$storage_fd" "/proc/$BASHPID/fd/$shared_fd" || {
     exec {storage_fd}<&-
     exec {shared_fd}<&-
     return 1
@@ -1015,7 +1016,7 @@ prepare_user_mount_sources() {
     echo "omarchy-windows-vm: storage and shared must be different directories" >&2
     return 1
   }
-  chmod 0700 -- "$storage" "$shared"
+  chmod u=rwx,go=,a-s -- "$storage" "$shared"
 }
 
 storage_space_path() {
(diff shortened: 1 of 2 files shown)


686f6c61:
The numeric 0700 was leaving setgid in place, and the existing `%a == 700` check is exactly what would then fail. `u=rwx,go=,a-s` plus the `g+s` fixture is the right pair.

Read against current `quattro` `bin/omarchy-windows-vm` and `test/shell.d/windows-vm-compose-test.sh`.

slinn-01:
Independently reproduced this on a physical Omarchy 4.0.2-1 install upgraded from 4.0.1. Dockurr had left `~/Windows` with SGID (`2700`; it also became `2777` while the guest was running). The privileged launch trace reached `chmod 0700` successfully, then read `shared_mode=2700` and returned before Docker startup. Clearing SGID in place immediately restored launch, guest readiness, and RDP connectivity.

Validated this branch in a clean checkout:

- `bash test/shell.d/windows-vm-compose-test.sh`: pass, including the new `g+s` migration fixture and exact `700` assertion.
- `git diff --check quattro...HEAD`: pass.
- `bash test/shell.d/windows-vm-mount-boundary-test.sh`: fails at the existing protected-anchor identity assertion, but the same failure reproduces unchanged at base commit `b686ed89`, so it is not introduced by this patch.
- `./test/all`: CLI passes; shell has the same baseline/environment failures (missing `omarchy-pkgs`, live-shell IPC count, and the boundary result above). The Windows VM compose regression suite passes within the aggregate run.

The symbolic `chmod u=rwx,go=,a-s` is the right fix: it preserves the strict exact-mode postcondition while explicitly removing Dockurr-created special bits on every launch.

686f6c61:
Thanks — that 2700 → chmod 0700 → shared_mode=2700 trace is the live path I could not run here. The compose test covers the bits; your install shows Dockurr actually sets them.

Chessing234:
hey, this is already covered in https://github.com/omacom/omarchy/pull/12264 (clear setgid / special bits on windows vm harden). mind closing this as a duplicate?

PR #9328, opened 2026-08-31
Fix Obsidian -disable-gpu so CLI commands parse

### LINKED ISSUE #8538: Obsidian user-flags.conf default has single-dash typo (-disable-gpu), breaks Obsidian CLI

686f6c61:
## Problem

The shipped `config/obsidian/user-flags.conf` has `-disable-gpu`. Obsidian's CLI treats a single-dash token as a command, so `obsidian --version` fails with `Command "-disable-gpu" not found`.

#8700 is the one-character default fix. Copies already in `~/.config/obsidian/user-flags.conf` keep the typo until that file is refreshed.

## Change

Same default as #8700, plus a migration that rewrites an exact `-disable-gpu` line in the user's copy. Commented-out lines and symlinks are left alone.

If #8700 lands first, the default hunk here is a no-op on rebase. If this lands first, #8700's file change is the same line.

## How I verified

```text
bash test/shell.d/obsidian-user-flags-test.sh
# ok - the shipped Obsidian flags use --disable-gpu
# ok - a migration repairs existing ~/.config copies
# ok - migration turns the shipped typo into --disable-gpu
# ok - migration is idempotent
# ok - migration leaves a commented-out flag alone
# ok - migration succeeds when the user has no Obsidian flags
# ok - migration leaves a symlink alone
```

Not runtime-tested against a real Obsidian install.

Fixes #8538
Related: #8700

### DIFF
--- a/config/obsidian/user-flags.conf
+++ b/config/obsidian/user-flags.conf
@@ -1,3 +1,3 @@
 # Obsidian reads this file through the Arch package wrapper.
--disable-gpu
+--disable-gpu
 --enable-wayland-ime
--- a/migrations/1788166801.sh
+++ b/migrations/1788166801.sh
@@ -1,0 +1,9 @@
+echo "Fix Obsidian -disable-gpu so CLI commands parse"
+
+flags="$HOME/.config/obsidian/user-flags.conf"
+
+# A symlink is left alone: it already tracks whatever the package ships.
+[[ -f $flags && ! -L $flags ]] || exit 0
+grep -qxF -- '-disable-gpu' "$flags" || exit 0
+
+sed -i 's/^-disable-gpu$/--disable-gpu/' "$flags"
--- a/test/shell.d/obsidian-user-flags-test.sh
+++ b/test/shell.d/obsidian-user-flags-test.sh
@@ -1,0 +1,73 @@
+#!/bin/bash
+
+set -euo pipefail
+
+source "$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd)/base-test.sh"
+
+# Electron/Obsidian treats a single-dash token as a CLI command, so the shipped
+# `-disable-gpu` breaks `obsidian --version`. Complements #8700: that PR is the
+# default; this also repairs copies already in ~/.config.
+
+flags_src="$ROOT/config/obsidian/user-flags.conf"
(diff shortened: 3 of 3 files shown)


edschofield:
This fixed the CLI for me on a locally adapted Omarchy 4.0.2 setup, using Arch's Obsidian package (1.13.7-2).

Before the fix, `obsidian version` gave “Command "-disable-gpu" not found.” I changed `-disable-gpu` to `--disable-gpu` in `~/.config/obsidian/user-flags.conf`. After that, `obsidian version`, `obsidian help`, `obsidian vault`, and `obsidian files total` all worked. I didn't need to restart Obsidian.

I made the change by hand, so I haven't tested this PR's migration. Updating existing users' settings makes sense: fixing the default file alone would have left my CLI broken.

bjarneo:
Automated duplication check: this pull request looks similar to #8700, which covers the same Obsidian disable-gpu flag. I keep that one open and close this one to consolidate review.

PR #9329, opened 2026-08-31
Install playerctl for Voxtype pause_media

686f6c61:
## Problem

Voxtype ships `pause_media = true` and implements that by calling `playerctl`. `omarchy-voxtype-install` never installed it, so the journal shows `playerctl not found`.

#9244 adds `playerctl` to a fresh install. Anyone who already ran the installer still does not have the package.

## Change

Same installer line as #9244, plus a migration that `omarchy-pkg-add playerctl` when `voxtype-bin` is present.

If #9244 lands first, the installer hunk here is a no-op on rebase.

## How I verified

```text
bash test/shell.d/voxtype-playerctl-test.sh
# ok - voxtype-install installs playerctl with voxtype
# ok - a migration installs playerctl for existing Voxtype installs
# ok - migration no-ops when Voxtype is not installed
# ok - migration installs playerctl when Voxtype is installed
```

Not runtime-tested against a live Voxtype session.

Related: #9244

### DIFF
--- a/bin/omarchy-voxtype-install
+++ b/bin/omarchy-voxtype-install
@@ -8,7 +8,7 @@ set -e
 # Install voxtype and configure it for use.
 
 if gum confirm "Install Voxtype + AI model (~150MB) to enable dictation?"; then
-  omarchy-pkg-add wtype voxtype-bin
+  omarchy-pkg-add wtype playerctl voxtype-bin
 
   # Setup voxtype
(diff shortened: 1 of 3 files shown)


bjarneo:
Automated duplication check: this pull request looks similar to #9244, which covers the same voxtype playerctl install. I keep that one open and close this one to consolidate review.

PR #9332, opened 2026-08-31
Migrate hybrid GPU switching from supergfxctl to cardwire

yashwk:
## Migrate hybrid GPU switching from `supergfxctl` to `cardwire`

Implements the proposal in #5982. `supergfxctl`/`supergfxd` is deprecated and unmaintained, and it forces a reboot or logout on every mode switch because it unbinds PCI devices. `cardwire` is its actively maintained successor: it blocks access at the syscall level with eBPF/LSM hooks, so switching between integrated and hybrid is instant and the dGPU can stay in D3cold.

### What changes

- **`omarchy-toggle-hybrid-gpu`** now drives `cardwire`:
  - Installs `cardwire` and enables the `cardwired` service when missing.
  - When a leftover `supergfxd` is present, retires it so it can't fight cardwire over the dGPU: disables the service and removes `/etc/supergfxd.conf`, the `force-igpu` system-sleep hook, and the `supergfxd.service.d` delay-start drop-in.
  - Queries the current mode with `cardwire get` using the same bounded retry/timeout protection against a wedged daemon, and switches with `cardwire set hybrid | integrated`. No reboot required.
- **`omarchy-hw-hybrid-gpu`** gates the Hardware menu entry on cardwire reporting `integrated` as an available mode — only true on hybrid laptops (iGPU + offload dGPU); desktops get `hybrid`/`manual` alone — with the same lspci GPU-counting fallback when the daemon can't answer or cardwire isn't installed.
- **Drops** the supergfxd-only workarounds from the defaults: `default/systemd/system-sleep/force-igpu` and `default/systemd/system/supergfxd.service.d/delay-start.conf`. They existed to work around supergfxd unbinding the dGPU while the display subsystem initialized or during sleep/hibernate resume; cardwire's eBPF/LSM approach doesn't have those failure modes.
- **Updates** `docs/file-layout.md` and rewrites the two hybrid GPU shell test suites for `cardwire`, including coverage of the install and legacy-cleanup paths.

### Testing

- `./test/cli` passes.
- `./test/shell` passes; the only failing files are the pre-existing `config-test.sh`, `snapper-test.sh`, and `unowned-system-paths-test.sh`, which need a `~/omarchy-pkgs` checkout that isn't present on this machine.

### Note

The toggle installs `cardwire` via `omarchy-pkg-add`, mirroring how `supergfxctl` was handled — so the omarchy package repo needs to ship `cardwire` in place of the retired `supergfxctl`. cardwire is packaged upstream in AUR (`cardwire`) and the Open Gaming Collective repo.

Ref: discussion #5982


yashwk:
@ryanrhughes review required please

PR #9340, opened 2026-08-31
Stop theme-set-keyboard-f16 from hanging on wedged VIA boards

### LINKED ISSUE #8243: omarchy theme set / omarchy update hangs indefinitely in omarchy-theme-set-keyboard-f16 with a Keychron K8 Pro attached
### System details

AMD Ryzen 7 9700X, AMD Radeon RX 6800 XT, Omarchy 4.0.1-1, kernel 7.1.8-arch1-3, Hyprland 0.56.2, qmk-hid 0.1.11-3 (explicitly installed), keyboard: Keychron K8 Pro (QMK/VIA firmware)

### What's wrong?

# What's wrong?
With a Keychron K8 Pro (QMK/VIA keyboard) attached over USB, every `omarchy theme set <theme>` blocks forever after printing a single line (`Effect: 1`). Because the 4.0.1 update migration `1787481315` re-stages the current theme via `omarchy-theme-refresh`, `omarchy update` wedged at that migration on every retry until I unplugged the keyboard — it looked like a bricked update.

## Process tree while stuck (two independent instances, same spot)

```
bash -lc 'theme=$(omarchy-theme-switcher); ... omarchy-theme-set "$theme"'
└─ omarchy-theme-set tokyo-night
   └─ omarchy-theme-set-keyboard
      └─ omarchy-theme-set-keyboard-f16      <- STAT do_wait
         └─ qmk_hid via --rgb-hue 212        <- never exits
```

## Steps to reproduce

1. Attach a Keychron K8 Pro over USB
2. Run `omarchy theme set tokyo-night`
3. Observe: prints `Effect: 1`, then hangs forever
4. Unplug the keyboard, re-run: completes immediately

Direct reproduction of the blocking call:

```
$ timeout 5 qmk_hid via --rgb-hue 100   # K8 Pro attached
[killed at timeout, exit 124]
```

Side note for anyone debugging this from user reports: the f16 script only redirects stderr of the qmk_hid calls, so qmk_hid's own stdout response (`Effect: 1`) lands in the terminal and reads like output from whatever step preceded the hang.

Context: qmk-hid was not preinstalled on my system — I added it manually while configuring the K8 Pro (`pacman -S qmk-hid`). `omarchy-theme-set-keyboard-f16` then picked it up opportunistically and applied theme colors to the keyboard on every theme change. That is reasonable behavior; the problem is only that nothing bounds it.

## Analysis

Neither keyboard theming script detects actual hardware. The entire guard in `/usr/share/omarchy/bin/omarchy-theme-set-keyboard-f16` is:

```bash
if omarchy-cmd-present qmk_hid && [[ -f $FRAMEWORK16_THEME ]]; then
```

i.e. binary present + theme ships a `keyboard.rgb`, then fire five unbounded `qmk_hid` calls. Despite the name, nothing checks for a Framework 16 or any VIA device. The sibling `omarchy-theme-set-keyboard-asus-rog` has the identical pattern (`asusctl` present + rgb file) but gets away with it because `asusctl` fails fast on foreign hardware — `qmk_hid` instead blocks indefinitely on a non-responsive HID device.

(`omarchy-debug` is not available on my installed version to attach its output.)

## Expected

Theme application completes (or the keyboard step fails fast) regardless of what other HID devices are attached.

## Suggested fix

- Wrap the five `qmk_hid` invocations in `timeout`, matching the pattern used elsewhere in Omarchy's scripts (e.g. `shell_ipc()` uses `timeout 2`).
- Optionally verify an actual VIA-capable device responds before invoking `qmk_hid`.

Separately, qmk-hid itself hanging forever instead of erroring is arguably an upstream bug — happy to file there too.

Chessing234:
## Summary
- `qmk_hid via` can hang forever on non-Framework VIA keyboards (Keychron, etc.)
- That blocked `omarchy theme set` and migrations
- Bound the VIA sequence with `timeout`

Fixes #8243.

## Test plan
- [x] `bash test/shell.d/theme-set-keyboard-f16-timeout-test.sh`
- [ ] Theme set on a Keychron/VIA board no longer wedges

### DIFF
--- a/bin/omarchy-migrate
+++ b/bin/omarchy-migrate
@@ -32,6 +32,14 @@ OMARCHY_PATH="${OMARCHY_PATH:-/usr/share/omarchy}"
 STATE_DIR="${OMARCHY_MIGRATION_STATE:-$HOME/.local/state/omarchy/migrations}"
 MIGRATIONS_DIR="$OMARCHY_PATH/migrations"
 
+require_migrations_dir() {
+  if [[ ! -d $MIGRATIONS_DIR ]]; then
+    echo "omarchy-migrate: migrations directory missing: $MIGRATIONS_DIR" >&2
+    echo "Check OMARCHY_PATH (currently ${OMARCHY_PATH})." >&2
+    exit 2
+  fi
+}
+
 migration_entries() {
   [[ -d $MIGRATIONS_DIR ]] || return 0
 
@@ -56,6 +64,7 @@ pending_migrations() {
 }
 
 if [[ $mode == "pending" ]]; then
+  require_migrations_dir
   pending_output=$(pending_migrations)
   if [[ -n $pending_output ]]; then
     printf '%s\n' "$pending_output"
@@ -64,6 +73,8 @@ if [[ $mode == "pending" ]]; then
     exit 1
   fi
 fi
+
+require_migrations_dir
 
 wait_for_pacman_transaction() {
   local lock_file=/var/lib/pacman/db.lck
--- a/test/shell.d/migrate-pending-missing-dir-test.sh
+++ b/test/shell.d/migrate-pending-missing-dir-test.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+set -euo pipefail
+
+source "$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd)/base-test.sh"
+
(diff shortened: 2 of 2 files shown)


Chessing234:

Chessing234:
closing as a duplicate of #9343 (same commit).

PR #9342, opened 2026-08-31
Reprompt Wi-Fi passphrase when WPA3 auth fails as Network lost

### LINKED ISSUE #8620: Network panel: wrong Wi-Fi password on a WPA3 network fails silently as "Network lost" and never re-prompts
## What happens

If a known network's stored passphrase is wrong and the AP is WPA3/SAE (e.g. a Starlink router, `WPA2 WPA3` transition mode), clicking it in the bar's network panel shows **"Network lost"** and does nothing else. Every subsequent click does the same. The user has no way to enter a new password from the panel; the only recovery is `nmcli connection delete` and re-adding the network.

## Why

Under SAE a bad passphrase fails during 802.11 *authentication*, so wpa_supplicant never associates and NetworkManager reports the failure as `WifiNetworkLost` (`association took too long` → `ssid-not-found`) rather than `WifiAuthTimeout`/`NoSecrets`. Under plain WPA2 the same wrong password fails in the 4-way handshake and is reported as a key error, which the panel handles.

`shell/plugins/panels/network/Model.js` only re-opens the passphrase prompt for the two WPA2-shaped reasons:

```js
function shouldRepromptPassphrase(reason, needsCredentials, reasons) {
  var r = reasons || {}
  if (!needsCredentials) return false
  return reason === r.NoSecrets || reason === r.WifiAuthTimeout
}
```

Verified with `iw event -t -f`: SAE commit/confirm exchanged with `status 0`, then the station is torn down without an association request. Forcing the profile to WPA2 (`pmf 1`) turns the identical password into `WPA: 4-Way Handshake failed - pre-shared key may be incorrect` / `reason=WRONG_KEY`, confirming the stored key was wrong all along.

## Suggested fix

Re-prompt on any failure of a credentialed network except an explicit user disconnect, and label the ambiguous case honestly:

```diff
 function shouldRepromptPassphrase(reason, needsCredentials, reasons) {
   var r = reasons || {}
   if (!needsCredentials) return false
-  return reason === r.NoSecrets || reason === r.WifiAuthTimeout
+  // On WPA3/SAE a wrong passphrase is reported as WifiNetworkLost, not
+  // WifiAuthTimeout, so a bad saved key would otherwise never re-prompt.
+  return reason !== r.WifiClientDisconnected
 }

 function networkFailureReason(reason, needsCredentials, reasons) {
   ...
+  if (needsCredentials && reason === r.WifiNetworkLost) return "Wrong password or network lost"
   if (reason === r.WifiNetworkLost) return "Network lost"
```

Trade-off: a genuine out-of-range failure on a secured network also opens the prompt. That is far better than a silent dead click, and the prompt is one keypress to dismiss. Running this as a plugin clone (`omarchy plugin clone omarchy.network`) on Omarchy 4.0.1 for a day with no side effects.

## Repro

1. Connect to a WPA3 or WPA2/WPA3 network so a profile is saved.
2. Change the router password (or `nmcli connection modify <ssid> wifi-sec.psk wrong`).
3. Click the network in the panel → "Network lost", no prompt, repeatable.

## System details

Omarchy 4.0.1-1 · NetworkManager 1.x · wpa_supplicant · Intel BE213 (iwlmld) · kernel 7.1.9

Chessing234:
## Summary
- Wrong PSK on WPA3 often surfaces as `WifiNetworkLost` instead of `WifiAuthTimeout`
- The panel never asked again
- Treat NetworkLost as a credential failure when the network needs a passphrase

Fixes #8620.

## Test plan
- [x] `bash test/shell.d/network-test.sh` (new NetworkLost asserts)
- [ ] Wrong WPA3 password reopens the passphrase prompt

### DIFF
--- a/bin/omarchy-migrate
+++ b/bin/omarchy-migrate
@@ -32,6 +32,14 @@ OMARCHY_PATH="${OMARCHY_PATH:-/usr/share/omarchy}"
 STATE_DIR="${OMARCHY_MIGRATION_STATE:-$HOME/.local/state/omarchy/migrations}"
 MIGRATIONS_DIR="$OMARCHY_PATH/migrations"
 
+require_migrations_dir() {
+  if [[ ! -d $MIGRATIONS_DIR ]]; then
+    echo "omarchy-migrate: migrations directory missing: $MIGRATIONS_DIR" >&2
+    echo "Check OMARCHY_PATH (currently ${OMARCHY_PATH})." >&2
+    exit 2
+  fi
+}
+
 migration_entries() {
   [[ -d $MIGRATIONS_DIR ]] || return 0
 
@@ -56,6 +64,7 @@ pending_migrations() {
 }
 
 if [[ $mode == "pending" ]]; then
+  require_migrations_dir
   pending_output=$(pending_migrations)
   if [[ -n $pending_output ]]; then
     printf '%s\n' "$pending_output"
@@ -64,6 +73,8 @@ if [[ $mode == "pending" ]]; then
     exit 1
   fi
 fi
+
+require_migrations_dir
 
 wait_for_pacman_transaction() {
   local lock_file=/var/lib/pacman/db.lck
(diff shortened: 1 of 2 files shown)


Chessing234:

Chessing234:
closing as a duplicate of #9343 (same commit).

PR #9344, opened 2026-08-31
Keep speed tests loaded during process cleanup

ptqa:
## Problem

Starting speed test and closing before it's finished produces infinite download processes. 

## Summary

- keep the network and disk speed-test panels loaded after dismissal
- let their existing SIGTERM and onExited cleanup finish before Quickshell can destroy the Process object
- add a lifecycle manifest regression test for both panels

## Before and after

Reproduced on packaged Omarchy 4.0.2 by summoning the network speed test and dismissing it during the upload phase. Four `/bin/bash /usr/share/omarchy/bin/omarchy-network-speedtest up` workers remained reparented to the user systemd process after 4m18s.

Repeated against this branch with the same mid-upload dismissal. After three seconds, no network-speedtest Bash workers, Netflix curl processes, or dd producers remained.

## Test

- `bash test/shell.d/speedtest-lifecycle-test.sh`

### DIFF
--- a/shell/plugins/panels/disk-speedtest/manifest.json
+++ b/shell/plugins/panels/disk-speedtest/manifest.json
@@ -8,6 +8,7 @@
   "kinds": [
     "panel"
   ],
+  "keepLoaded": true,
   "entryPoints": {
     "panel": "Panel.qml"
   }
--- a/shell/plugins/panels/speedtest/manifest.json
+++ b/shell/plugins/panels/speedtest/manifest.json
@@ -8,6 +8,7 @@
   "kinds": [
     "panel"
   ],
+  "keepLoaded": true,
   "entryPoints": {
     "panel": "Panel.qml"
   }
--- a/test/shell.d/speedtest-lifecycle-test.sh
+++ b/test/shell.d/speedtest-lifecycle-test.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+set -euo pipefail
+
+source "$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd)/base-test.sh"
+
+run_node_test <<'JS'
+const fs = require('fs')
+
+for (const plugin of ['speedtest', 'disk-speedtest']) {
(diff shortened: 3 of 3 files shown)


estebanmarinramirez (commented):
Reviewed — this looks correct and well-targeted. 👍

**Root cause is addressed, not just the symptom.** On dismissal the panels set `proc.running = false` (SIGTERM), but `Process.running` stays true until the child actually exits and the reaping happens in `onExited` (`speedtest/Panel.qml:86-87,141`, `disk-speedtest/Panel.qml:56-57,111`). Without `keepLoaded`, the loader binding `active: … && (keepLoaded || openPanelIds[pluginId])` (`shell.qml:625`) flips false the moment the panel closes, so Quickshell destroys the `Process` object *inside that SIGTERM-in-flight window* and `onExited` never runs — which is exactly how the bash/curl/dd workers end up reparented to systemd. `keepLoaded: true` keeps the loader active long enough for the panels' own existing cleanup to finish.

**It's the idiomatic mechanism, used correctly:**
- `keepLoaded` is a first-class field (read at `shell.qml:600/620/625`, documented in `shell/README.md` and `docs/omarchy-shell.md` for exactly "outlive a single summon") and already used by ~11 core plugins (osd, notifications, media, lock, clipboard…).
- Passes the manifest validation that requires it be boolean when present (`plugins-test.sh:135`).
- Both panels share the identical dismiss → SIGTERM → `onExited` shape, so applying it to both is consistent and complete.
- It keeps the panel *loaded*, not *visible* (visibility is bound separately), so there's no dismissal-UX change.

**Two minor, non-blocking notes:**
- The regression test is manifest-only — it asserts `keepLoaded === true`, i.e. guards against the flag being removed, rather than exercising the process-lifecycle behavior itself. Reasonable given how hard Quickshell `Process` teardown is to test, and the before/after repro in the description is convincing — just worth naming that the net catches "flag present," not "leak fixed."
- Tradeoff worth acknowledging: both panels now stay memory-resident after first summon. Negligible while idle (`running: false`, no speedtest process alive) and clearly worth it versus orphaned worker chains.

Nice, minimal fix.

PR #9351, opened 2026-08-31
Read migrations and browser policies from the trusted Omarchy tree

Adolanium:
`omarchy-migrate` runs every pending migration from `$OMARCHY_PATH` with the caller's sudo credentials a prompt away:

```
OMARCHY_PATH="$OMARCHY_PATH" bash -euo pipefail "$file"
```

`omarchy-install-browser` sources its policy helper from the same caller-controlled path, and the helper installs the default Firefox policy as root:

```
source "$OMARCHY_PATH/install/helpers/browser-policy.sh"
as_root install -m 644 -o root -g root -T "$policies" "$distribution_dir/policies.json"
```

A caller who points `OMARCHY_PATH` at a tree they can write, then types the sudo password for an update or a browser install, runs their own script as root or installs their own mandatory browser policy. The session environment is user-writable by definition, and migration markers live under `~/.local/state`, so any migration can be made pending again.

`omarchy-plymouth-set` and the refresh helpers already refuse a user-owned source tree unless a root-owned `/etc/omarchy.conf` names that checkout. The update flow did not.

What changed:

- `omarchy-migrate` lists and runs migrations from the trusted source root, and passes that root to each migration.
- `omarchy-install-browser` sources the browser policy helper from the trusted source root.
- `browser-policy.sh` installs the default policy files from the trusted source root.
- The locate migration pins `/etc/updatedb.conf` instead of forwarding a caller-chosen `OMARCHY_UPDATEDB_CONF_PATH` into the root shell.

A linked checkout still works when `/etc/omarchy.conf` is a root-owned regular file whose only line names that checkout, which is what `omarchy-dev-link` writes. Setting `OMARCHY_PATH` by hand without `omarchy dev link` no longer moves the migration queue.

Tests: new `test/shell.d/privileged-source-update-test.sh` pins the trusted-tree behavior and shows a poisoned `OMARCHY_PATH` is ignored, with sudo stubbed so nothing touches the real system. `migrate-scope` and `migrate-wrapper` now authorize their scratch trees through a stand-in conf, the way `omarchy-dev-link` would. `locate-test` runs the pinned migration against a stand-in `updatedb.conf`. All pass, along with `migrate-notify` and `bin-style`.

### DIFF
--- a/bin/omarchy-install-browser
+++ b/bin/omarchy-install-browser
@@ -6,7 +6,59 @@
 
 set -e
 
-source "$OMARCHY_PATH/install/helpers/browser-policy.sh"
+# Browser policy setup installs files as root, so the helper and the policy
+# files it installs must come from a trusted tree. The packaged tree is always
+# trusted. A linked checkout is trusted only when /etc/omarchy.conf is a
+# root-owned regular file whose only line names that checkout, which is what
+# omarchy-dev-link writes. omarchy-plymouth-set uses the same rule. Keep this
+# block in step with omarchy-refresh-limine, omarchy-refresh-pacman,
+# omarchy-toggle-hybrid-gpu, omarchy-hibernation-setup, omarchy-migrate, and
+# install/helpers/browser-policy.sh.
+omarchy_privileged_source_root() {
+  local packaged=/usr/share/omarchy
+  local path=${OMARCHY_PATH:-$packaged}
+  local conf=/etc/omarchy.conf
+  local uid mode
+  local -a lines
+
+  if [[ $path == "$packaged" ]]; then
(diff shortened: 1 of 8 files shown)


Adolanium:

`omarchy-migrate` runs as the calling user. Every root step is a fresh `sudo` password prompt from that same user, who already has full wheel access. Someone who can set `OMARCHY_PATH` in the session can set `PATH` too and shim `sudo` directly. And `omarchy-dev-link` treats a user-controlled `OMARCHY_PATH` as trusted by design.

The description also said the refresh helpers already refuse a user-owned source tree. That was wrong. Only `omarchy-plymouth-set` has that check.

PR #9352, opened 2026-08-31
Refuse toggle and web app names that contain a slash

Adolanium:
`omarchy-hyprland-toggle` joins the flag name into two paths unchecked:

```
FLAG_FILE="$HOME/.local/state/omarchy/toggles/hypr/$FLAG_NAME.lua"
FLAG_SOURCE="$OMARCHY_PATH/default/hypr/toggles/$FLAG_NAME.lua"
```

`omarchy-hyprland-toggle ../../victim off` deletes `victim.lua` two levels above the toggle directory and exits 0. `on` with the same name copies to a mirrored path outside `toggles/hypr`. Hyprland never sources that file (`default/hypr/toggles.lua` only walks `toggles/hypr`), so `on` just leaves a stray copy. The delete is the part that matters.

`omarchy-webapp-remove` has the same shape in its fallback. When the name is not in the index, the path is rebuilt from the raw name:

```
rm -f "${desktop_file:-$DESKTOP_DIR/$APP_NAME.desktop}"
```

`omarchy-webapp-remove ../../victim` deletes `victim.desktop` outside the applications directory and exits 0.

This is not a security fix. Both commands run as the user on the user's own files, and no caller in the repo passes untrusted input: the toggle callers use fixed names, and `omarchy-webapp-remove` gets a basename from the picker or from `omarchy-remove-launcher-entry`. It is a footgun guard for a mistyped or scripted name, and it brings these two commands in line with `omarchy-webapp-install` and `omarchy-theme-set`, which already refuse names with a slash.

What changed:

- `omarchy-hyprland-toggle` rejects a flag name containing a slash before building its paths.
- `omarchy-webapp-remove` rejects an app name containing a slash, with the same message as `omarchy-webapp-install`.
- Both print the refusal on stderr.

Tests: new `test/shell.d/launcher-toggle-name-guard-test.sh`. Both refusals say why they refused and leave a file outside the guarded directory untouched, and plain names still toggle and remove. `webapp-name` (including the nested legacy launcher case), `webapp-install-escaping`, and `toggle` all pass. The new test fails on the pre-fix scripts.

### DIFF
--- a/bin/omarchy-hyprland-toggle
+++ b/bin/omarchy-hyprland-toggle
@@ -14,6 +14,16 @@ fi
 
 FLAG_NAME="$1"
 ACTION="${2:-toggle}"
+
+# The name becomes a filename in two paths. A slash would turn it into
+# directory levels, so a typo like ../foo would make off delete a .lua outside
+# the toggle directory. Refuse it, the same way omarchy-webapp-install and
+# omarchy-theme-set refuse names with a slash.
+if [[ $FLAG_NAME == */* ]]; then
+  echo "Flag name cannot contain '/': $FLAG_NAME" >&2
+  exit 1
+fi
+
 FLAG_FILE="$HOME/.local/state/omarchy/toggles/hypr/$FLAG_NAME.lua"
 FLAG_SOURCE="$OMARCHY_PATH/default/hypr/toggles/$FLAG_NAME.lua"
 
--- a/bin/omarchy-webapp-remove
+++ b/bin/omarchy-webapp-remove
@@ -50,6 +50,14 @@ if [[ -z $APP_NAME ]]; then
   exit 1
 fi
 
+# Same rule as omarchy-webapp-install: the name becomes a filename. A slash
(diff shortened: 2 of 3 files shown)


Adolanium:
Updated after a second look.

- The description said `on` copies into the directory Hyprland sources. It does not: the copy lands at a mirrored path outside `toggles/hypr`, which `toggles.lua` never walks. I fixed the text. The `off` and `webapp-remove` deletes are the real issue and are stated as such.
- Reframed the title, body, commit message, and comments: this is a footgun guard for consistency with `omarchy-webapp-install` and `omarchy-theme-set`, not a security fix. Both commands run as the user on the user's own files and no caller passes untrusted input.
- The refusal in `omarchy-webapp-remove` now goes to stderr like the one in `omarchy-hyprland-toggle`.
- Reran the new test plus `webapp-name`, `webapp-install-escaping`, and `toggle`. All pass; the new test still fails on the pre-fix scripts.

PR #9354, opened 2026-08-31
Add battery charge limit picker to the power panel

vincerhodes:
## What

Adds a charge limit picker (80 / 90 / 100%) to the power panel, so laptops that expose `charge_control_end_threshold` can cap charging with one click instead of dropping to a root shell.

## How

- **Capability gating**: new `omarchy-hw-battery-charge-limit` exits 0 only when a battery actually exposes `charge_control_end_threshold` (asus-wmi, thinkpad_acpi, Framework, etc.). The panel probes it once per open and hides the picker entirely on unsupported hardware and desktops.
- **Reading**: new hidden `omarchy-battery-limit-get` prints the current end threshold from the first capable battery. Direct sysfs read rather than upower, which can report a phantom "75-80%" start/end range.
- **Writing**: new `omarchy-battery-limit-set <80|90|100>` follows the `omarchy-dns` privilege pattern exactly — args validated before elevation, re-execs via `sudo` when a terminal is available, `pkexec` otherwise, PATH pinned to trusted directories when root. Writes every capable battery.
- **Passwordless grant**: `etc/sudoers.d/omarchy-battery-limit` covers exactly the three fixed presets, mirroring the existing `omarchy-dns` rule, so the one-click panel toggle never prompts. Any other value is rejected before elevation.
- **Persistence**: the setter saves the choice to `/etc/omarchy/battery-limit`, and a new `omarchy-battery-limit.service` oneshot re-applies it at boot for hardware whose embedded controller forgets thresholds across power cycles. No-op when no limit has been saved (`ConditionPathExists`).
- **Migration**: existing installs get the sudoers rule unconditionally and the systemd unit only on capable hardware.

## Tests

- `test/shell.d/battery-limit-test.sh` — sudoers rule scope and syntax, arg rejection (exit 2), sudo/pkexec elevation routing, sysfs fixture coverage for the hw check and getter.
- `test/shell.d/power-charge-limit-test.sh` — `parseChargeLimit` unit tests plus panel source assertions (correct commands, no hardcoded sysfs paths, no sudo in the shell, probe-once-per-open gating).
- `./test/cli` and `./test/shell` pass.

## Notes

- The panel picker is ported from a locally-run clone of the power panel that has been in daily use on an ASUS (asus-wmi) machine.
- Visual verification done: picker confirmed working in the running shell on asus-wmi hardware (see comment below).

### DIFF
--- a/bin/omarchy-battery-limit-get
+++ b/bin/omarchy-battery-limit-get
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+# omarchy:summary=Print the current battery charge limit percentage.
+# omarchy:hidden=true
+
+power_supply_path="${OMARCHY_POWER_SUPPLY_PATH:-/sys/class/power_supply}"
+
+for threshold in "$power_supply_path"/BAT*/charge_control_end_threshold; do
+  if [[ -r $threshold ]]; then
+    cat "$threshold"
+    exit 0
+  fi
+done
+
+exit 1
--- a/bin/omarchy-battery-limit-set
+++ b/bin/omarchy-battery-limit-set
@@ -1,0 +1,94 @@
+#!/bin/bash
+
+# omarchy:summary=Set the battery charge limit
+# omarchy:args=<80|90|100>
+# omarchy:examples=omarchy battery limit-set 80 | omarchy battery limit-set 100
+
+set -euo pipefail
+
+# Whenever this runs as root — invoked directly through the passwordless
+# sudoers rule, or re-execed by require_root below — sudo's secure_path decides
+# where a bare helper resolves, and a dev link (etc/sudoers.d/omarchy-dev-path)
+# prepends a user-writable checkout bin/ to it. Every helper this script calls
+# by bare name (install, cat) is a system tool, never an omarchy-* command, so
+# pin PATH to trusted system directories and keep root from resolving one out
+# of that checkout. The unprivileged wrapper phase keeps the caller's PATH so
+# it can still find sudo/pkexec.
+if (( EUID == 0 )); then
+  export PATH=/usr/local/sbin:/usr/local/bin:/usr/bin:/usr/sbin:/bin:/sbin
+fi
+
+STATE_FILE=/etc/omarchy/battery-limit
+
+usage() {
+  echo "Usage: omarchy-battery-limit-set <80|90|100>" >&2
+}
+
+# Validate before any elevation so a bad argument never reaches sudo/pkexec,
+# and so the check can be exercised without root.
+if (( $# != 1 )); then
+  usage
+  exit 2
+fi
+
+case "$1" in
+80 | 90 | 100)
+  limit="$1"
+  ;;
+*)
+  usage
+  exit 2
+  ;;
+esac
+
+# The path etc/sudoers.d/omarchy-battery-limit names. The privileged half
+# always runs from there rather than from whichever copy was invoked, so the
+# rule matches even where $OMARCHY_PATH points at a checkout.
+PACKAGED_PATH=/usr/bin/omarchy-battery-limit-set
+
+# True when sudo would run this exact command without stopping for a password.
+# `sudo -l` on its own reports whether a command is permitted, which the blanket
+# %wheel rule answers yes to for everything; the long listing prints the matched
+# entry's tags, so !authenticate is the grant in
+# etc/sudoers.d/omarchy-battery-limit and nothing else. Listing runs nothing
+# and, under -n, prompts for nothing, so a machine whose omarchy-settings
+# predates that file falls through to polkit instead of dying on a password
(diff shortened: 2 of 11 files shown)


vincerhodes:
Update: tested working on an ASUS Zenbook (asus-wmi).

- Panel picker renders correctly (header tracks current limit, all three glyphs fine, active state highlights)
- Each preset applies with no password prompt via the sudoers grant; `charge_control_end_threshold` follows the selection
- `/etc/omarchy/battery-limit` persists the choice and `omarchy-battery-limit.service` is enabled for boot restore
- Unsupported-hardware path is covered by the automated tests (fixture-based); this machine only exercises the supported path

Screenshot attached below.

screenshot-2026-09-01_00-49-06

flip-in:
Ran this locally on Apple Silicon and it works, but only after one change: **the `BAT*` glob excludes Apple Silicon entirely, so the picker never appears there.**

MacBookPro18,3 (M1 Pro), Asahi, kernel 7.1.6-1-1-ARCH, Omarchy 4.0.2-mac.1.

`omarchy-hw-battery-charge-limit`, `-get` and `-set` all scan `"$power_supply_path"/BAT*/charge_control_end_threshold`. There is no `BAT*` on this machine — the ACPI naming is x86's. Asahi's battery is `macsmc-battery`:

```console
$ ls -d /sys/class/power_supply/BAT*
ls: cannot access '/sys/class/power_supply/BAT*': No such file or directory

$ for d in /sys/class/power_supply/*/; do
    printf "%-26s type=%-8s scope=%-8s limit=%s\n" "$(basename $d)" \
      "$(cat $d/type 2>/dev/null)" "$(cat $d/scope 2>/dev/null || echo -)" \
      "$([ -r $d/charge_control_end_threshold ] && echo YES || echo no)"
  done
macsmc-battery             type=Battery  scope=System   limit=YES
hidpp_battery_2            type=Battery  scope=Device   limit=no
macsmc-ac                  type=Mains    scope=-        limit=no
tps6598x-source-psy-0-0038 type=USB      scope=-        limit=no
```

So `omarchy-hw-battery-charge-limit` exits 1, the panel sets `chargeLimitSupported = false`, and the picker is hidden on hardware that fully supports charge limiting. The capability gating degrades exactly as designed — it just gates the wrong way here.

### Suggested change

Drop the `BAT*` assumption and match on the attribute plus `scope`:

```bash
for threshold in "$power_supply_path"/*/charge_control_end_threshold; do
  [[ -r $threshold ]] || continue
  scope=$(cat "${threshold%/*}/scope" 2>/dev/null)
  [[ -z $scope || $scope == "System" ]] || continue
  ...
done
```

The `scope` test is worth having on its own merit: it keeps a peripheral battery from being mistaken for the laptop's. `hidpp_battery_2` above is a Logitech MX Master — `type=Battery`, `scope=Device`. It exposes no charge control today, but nothing in the current glob would stop a future peripheral that does from being written to, and on a ThinkPad `BAT*` would not have caught it either.

### Hardware note that may be worth a line in the setter

`macsmc-battery` derives the start threshold itself. Writing only the end value moves start automatically, keeping a 5-point gap:

```
end 80  ->  start 75
end 85  ->  start 80
```

So on this hardware writing just `charge_control_end_threshold` is correct and complete — which is what the PR already does. Worth knowing if start-threshold handling is ever added, since here it would fight the firmware.

### Verified

With only the glob changed, everything else in the PR working as written: the picker appears, applies through the sudoers grant with no password prompt, the value survives in `/etc/omarchy/battery-limit`, and the oneshot is enabled. `charge_behaviour` reads `[auto] inhibit-charge`, and with the cap set the pack reports `Discharging` while docked rather than being held — the SMC enforces it in firmware, so it also holds across suspend with nothing running.

Happy to push the glob change as a commit here if that is useful.

vincerhodes:
Go for it if you're happy to push, I don't have a mac to test on but all sounds sensible @flip-in

flip-in:
Pushed the aarch64 fix as vincerhodes/omarchy#1 against this branch — merging it updates this PR automatically. Thanks @vincerhodes.

It replaces the `BAT*` glob in all three commands with a match on any supply exposing `charge_control_end_threshold`, skipping peripheral batteries (`scope=Device`) and keeping unscoped older drivers. Four regression tests added; I verified they fail without the change.

One correction to my earlier comment: I implied `BAT*` was what kept a peripheral battery from being mistaken for the system one. It was not — a wireless mouse reports `type=Battery` and was excluded only by the naming accident. The `scope` check is what actually guards that, and it holds on any hardware, not just Apple Silicon.

Also worth flagging for whoever reviews: this branch is currently 198 commits behind `quattro`.

bjarneo:
Automated duplication check: this pull request looks similar to #11792, which covers the same battery charge-limit picker. I keep that one open and close this one to consolidate review.

PR #9367, opened 2026-08-31
Show the nighttime fog icon in the bar

Chessing234:
`omarchy-weather-icon` fetches the live sunrise and sunset from wttr.in and sets `night` from them, then maps the weather code to a glyph. Every day/night-aware arm of that case statement branches on `night` except the fog one:

    119|122) icon="" ;;
    143|248|260) icon="" ;;
    176|263|353) [[ $night == "true" ]] && icon="" || icon="" ;;

The panel disagrees. `shell/plugins/panels/weather/Model.js` resolves the same three wttr codes as `night ? "" : ""`, and `iconForOpenMeteoCode` routes Open-Meteo 45/48 through them. So after sunset in fog or mist the bar showed the daytime glyph while the panel showed the nighttime one, for the same condition at the same moment.

The fix branches the fog arm on `night` with the same two codepoints the panel uses.

Verification, with the repo's shell suite: the new sweep in `test/shell.d/weather-test.sh` stubs `curl` and `date` so the script sees a fixed wttr payload and a fixed clock, then runs the real `bin/omarchy-weather-icon` for one code from every arm of its case statement — plus an unmapped code for the fallback — once past sunrise and once past sunset, comparing each result to `Model.iconForCode` rather than to a second copy of the mapping.

    ok - weather icon matches the panel's daytime glyphs
    ok - weather icon matches the panel's nighttime glyphs

Reverting only the one-line `bin/` change and re-running gives:

    ok - weather icon matches the panel's daytime glyphs
    code 143:  != 
    code 248:  != 
    code 260:  != 
    not ok - weather icon matches the panel's nighttime glyphs

The whole `weather-test.sh` file passes with the fix in place.


Chessing234:

Chessing234:
folded into #12115 so the weather icon review stays with the other bar panel tweaks.

PR #9369, opened 2026-08-31
Stop inverting the on/off argument to omarchy toggle bar

### LINKED ISSUE #9323: omarchy toggle bar on/off arguments are inverted — 'on' hides the bar
## What happened

`omarchy toggle bar on` **hides** the bar, and `omarchy toggle bar off` shows it — the opposite of what the words say.

## Cause

`omarchy-toggle-bar` passes the user's `on|off` argument straight through to the underlying flag, which is named for the inverse state:

```bash
omarchy-toggle bar-off "${1:-toggle}"
```

So "bar **on**" sets the `bar-off` flag (bar hidden), and "bar **off**" clears it (bar shown). The bare `toggle` form behaves fine; only the explicit `on`/`off` arguments are inverted.

## Expected

`omarchy toggle bar on` → bar visible; `omarchy toggle bar off` → bar hidden. The wrapper should invert the argument before passing it to the `bar-off` flag (or the flag could be renamed).

## System details

Omarchy 4.0.1-1

(Noticed while recovering a stuck-hidden bar: separately, the bar can remain hidden after a fullscreen window closes, with no `bar-off` flag set — mentioning in case it's related, but the inverted argument is the concrete reproducible bug here.)

Chessing234:
Fixes #9323.

`omarchy-toggle-bar` forwarded its argument to a flag named for the opposite state:

    omarchy-toggle bar-off "${1:-toggle}"

`bar-off` present means the bar is hidden, so `omarchy toggle bar on` created the flag and hid the bar, and `omarchy toggle bar off` removed it and showed the bar. The bare `toggle` form was unaffected, which is why the menu entry (`trigger.toggle.top-bar`, no argument) always behaved correctly and only the explicit arguments were wrong.

Renaming the flag would strand the state of anyone who has already hidden their bar, and `shell/plugins/bar/Bar.qml` watches that exact filename, so the wrapper now translates the requested visibility into the flag state. An unknown argument is rejected here too instead of being forwarded to `omarchy-toggle`, which would have reported the inner flag name in its usage line.

Both suites had the inverted behavior written into them, so both move with the fix:

- `test/shell.d/toggle-test.sh` asserted `bar on enables bar-off toggle`. It now asserts `off` sets the flag and `on` clears it, each idempotent, plus the new unknown-argument rejection.
- `test/acceptance.d/session-test.sh` called `omarchy-toggle-bar on` and then waited for `hidden bar layer parks off screen`. It now hides with `off`, reveals with `on`, and restores visibility on exit with `on`.

Verification — `test/shell.d/toggle-test.sh` with the fix in place:

    ok - bar off sets the bar-off flag
    ok - bar off is idempotent
    ok - bar on clears the bar-off flag
    ok - bar on is idempotent
    ok - bar rejects an unknown argument

Reverting only `bin/omarchy-toggle-bar` and re-running the same test:

    not ok - bar off sets the bar-off flag

`bar-test.sh` and `crash-capture-test.sh` (the other users of the `bar-off` flag name) still pass. The acceptance suite runs in a VM and was not run here; its change is the same argument swap.


Chessing234:

Chessing234:
merged quattro. fullscreen desktop now hides the bar with off, since toggle-bar on/off is visibility.

bjarneo:
Automated duplication check: this pull request looks similar to #7023, which covers the same toggle-bar on/off inversion. I keep that one open and close this one to consolidate review.

PR #9372, opened 2026-08-31
Keep config migrations from taking their staging file's identity

### LINKED ISSUE #9326: Config-rewriting migrations reset file mode to 0600 via `mktemp` + `mv`
### What happens

Several migrations transform a user config file with this pattern:

```sh
tmp=$(mktemp)
jq '…' "$config_file" > "$tmp" && mv "$tmp" "$config_file"
```

`mktemp` always creates its temp file as `0600`. `mv` on the same filesystem is
a rename that keeps the temp file's inode and permissions, so `$config_file`
silently inherits `0600`, discarding whatever mode it had (normally `0644`).

### Reproduce

```sh
install -m 644 /dev/stdin /tmp/demo.json <<<'{"a":1}'
stat -c %a /tmp/demo.json          # 644
tmp=$(mktemp); stat -c %a "$tmp"    # 600
jq . /tmp/demo.json > "$tmp" && mv "$tmp" /tmp/demo.json
stat -c %a /tmp/demo.json          # 600  <-- mode changed
```

### Expected

A migration whose purpose is to edit file *contents* should preserve the
file's existing mode.

### Affected migrations

`mktemp` + `mv "$tmp" "$file"` over a user config:

- `1780057136.sh` (kitty.conf, foot.ini)
- `1780294774.sh` (shell.json)
- `1781063758.sh` (hyprland.lua)
- `1781587663.sh` (nvim options.lua)
- `1784989000.sh` (shell.json)
- `1785344985.sh` (shell.json)
- `1786099804.sh` (shell.json)

`1785189600.sh` already does it correctly for `shell.json`, using
`cat "$tmp" > "$config_file"` (truncate-in-place, mode preserved) — so the
codebase is inconsistent.

### Impact

Low. `0600` on a file the user owns in `$HOME` breaks nothing at runtime, and
the change can only ever be more restrictive. It surfaces when a dotfile
manager that records file modes (chezmoi, yadm, etc.) picks up the unexpected
`0600` — in my case chezmoi re-tracked `~/.config/omarchy/shell.json` as a
`private_` (0600) file after an update.

### Suggested fix

Either:

1. `chmod --reference="$config_file" "$tmp"` immediately before the `mv` (keeps
   the atomic replace), or
2. Standardise on the in-place form already used by `1785189600.sh`:
   `cat "$tmp" > "$config_file" && rm -f "$tmp"`, or
3. A small shared helper in the migration runtime, e.g.
   `omarchy-replace-file <src-tmp> <dest>` that preserves dest's mode.

### System

- Omarchy 4.0.2-1

Chessing234:
Fixes #9326.

Seven migrations rewrite a user config by staging the new contents in `mktemp` and renaming that file over the original:

    tmp=$(mktemp)
    jq '…' "$config_file" >"$tmp" && mv "$tmp" "$config_file"

A same-filesystem `mv` is a rename, so the config keeps the staging file's inode — and with it, the staging file's identity rather than its own.

The reported symptom is the mode: `mktemp` creates at 0600, so a config that was 0644 comes back 0600 and a dotfile manager that records modes re-tracks it as private. There is a second, louder consequence the issue does not mention: the rename also replaces a **symlink**. `1785189600.sh` already names it in a comment next to the correct pattern — "Write through the path instead of replacing it, so a tmux.conf symlinked out of a dotfiles repo keeps pointing where it pointed."

Ran against an unfixed `migrations/1781063758.sh` with `~/.config/hypr/hyprland.lua` symlinked into a dotfiles directory:

    unfixed: symlink? NO (replaced by a regular file)
    unfixed: dotfiles copy still original? UNCHANGED - edit lost to the dotfiles repo

    fixed: symlink? YES
    fixed: dotfiles copy rewritten? YES

So on those setups the migration silently did not migrate anything the user keeps: it wrote a detached copy into `~/.config` and left the tracked file behind, which the next `chezmoi apply` reverts.

`1781043107.sh`, `1785189600.sh` and `1785633225.sh` already write through the path with `cat "$tmp" >"$file"`. This applies that same pattern to the remaining ten sites across seven migrations (`1780057136.sh` ×4, `1780294774.sh`, `1781063758.sh`, `1781587663.sh`, `1784989000.sh`, `1785344985.sh`, `1786099804.sh`) and removes the staging file afterwards, which `mv` used to do implicitly. The `&&` guard is kept, so a failed `jq`/`awk` still leaves the config untouched.

Verification — `test/shell.d/migration-config-rewrite-test.sh` runs the Hyprland bootstrap and bar clock migrations against a temporary home, first on plain 0644 configs and then on configs owned by a dotfiles directory and symlinked into `~/.config`:

    ok - Hyprland migration keeps the config mode
    ok - clock migration keeps the config mode
    ok - Hyprland migration writes through a dotfiles symlink
    ok - clock migration writes through a dotfiles symlink
    ok - no migration replaces a config with its staging file

Reverting only `migrations/` and re-running:

    mode is now 600
    not ok - Hyprland migration keeps the config mode

The last assertion is a grep over the whole directory, so the five migrations not exercised directly are covered against a reintroduction.

`terminal-keybinding-migration-test.sh` — the existing test over `1780057136.sh`, the file with four of the ten sites — still passes, as do the other migration tests except five that fail identically on unmodified `quattro` here (`bluetooth`, `legacy-power-udev-rules`, `mise-wrapper-quiet`, `security-fido2`, `sshd-hardening`); those are GNU-vs-BSD artifacts of running the suite on macOS, not related to this change.

### DIFF
--- a/migrations/1780057136.sh
+++ b/migrations/1780057136.sh
@@ -71,7 +71,8 @@ ensure_kitty_binding() {
         }
         print
       }
-    ' "$config" >"$tmp" && mv "$tmp" "$config"
+    ' "$config" >"$tmp" && cat "$tmp" >"$config"
+    rm -f "$tmp"
   elif ! grep -Eq "^map[[:space:]]+$key_regex[[:space:]]" "$config"; then
     if grep -qxF 'map shift+insert paste_from_clipboard' "$config"; then
       tmp=$(mktemp)
@@ -87,7 +88,8 @@ ensure_kitty_binding() {
           print "map " key " " binding
           inserted = 1
         }
-      ' "$config" >"$tmp" && mv "$tmp" "$config"
+      ' "$config" >"$tmp" && cat "$tmp" >"$config"
+      rm -f "$tmp"
     else
       printf '\n%s\nmap %s %s\n' "$comment" "$key" "$binding" >>"$config"
     fi
@@ -112,7 +114,8 @@ ensure_foot_text_binding() {
       }
       index($0, sequence "=") == 1 { $0 = sequence "=" binding }
       { print }
-    ' "$config" >"$tmp" && mv "$tmp" "$config"
+    ' "$config" >"$tmp" && cat "$tmp" >"$config"
+    rm -f "$tmp"
   elif grep -qxF '[text-bindings]' "$config"; then
     tmp=$(mktemp)
     SEQUENCE="$sequence" BINDING="$binding" COMMENT="$comment" awk '
@@ -127,7 +130,8 @@ ensure_foot_text_binding() {
         print sequence "=" binding
         inserted = 1
       }
-    ' "$config" >"$tmp" && mv "$tmp" "$config"
+    ' "$config" >"$tmp" && cat "$tmp" >"$config"
+    rm -f "$tmp"
   else
     printf '\n[text-bindings]\n%s\n%s=%s\n' "$comment" "$sequence" "$binding" >>"$config"
   fi
--- a/migrations/1780294774.sh
+++ b/migrations/1780294774.sh
@@ -27,5 +27,6 @@ if [[ -f $config_file ]] && omarchy-cmd-present jq; then
         .
       end
     )
-  ' "$config_file" >"$tmp" && mv "$tmp" "$config_file" || rm -f "$tmp"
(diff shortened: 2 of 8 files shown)


Chessing234:

Chessing234:
folding into #12250 with the other config mode fixes

PR #9399, opened 2026-08-31
Add SSH Agent (gcr-ssh-agent) as an optional service install

carletex:
This PR is the optional service installation for everyone who doesn't run 1Password (or Bitwarden) as their SSH agent. 

Out of the box, a passphrase-protected SSH key only works in an interactive terminal: anything without one (IDEs, coding agents, GUI apps) has nowhere to ask for the passphrase, so SSH pushes just fail. #1661 proposed adding ssh-agent for this, and @dhh's take was:

> I use 1password as my ssh agent, so this wouldn't be a good fit for a default setup, but I'm open to the idea that we could have a way to add this on as an optional service installation.

Unlike #1661, which wires up plain ssh-agent, this uses gcr-ssh-agent (already installed on every Omarchy machine via gnome-keyring) which can prompt graphically from any context and needs no ~/.ssh/config changes

## What it does

_Install > Service > SSH Agent_ enables `gcr-ssh-agent` and points `SSH_AUTH_SOCK` at it via `environment.d`. On first use of a key, a dialog prompts for the passphrase:

screenshot-2026-08-31_18-41-56

The key stays unlocked in the agent for the rest of the session. The checkbox is opt-in: nothing is persisted unless the user chooses to remember the passphrase in the keyring (which unlocks at login). A matching _Remove > Service_ entry undoes everything.

## Testing

install/remove round-trips, menu show/hide conditions, and end-to-end `ssh -T git@github.com` from a context with no terminal attached. `test/cli` passes.


seanpk:
I arrived here because I have been fighting my agents all day about pushing with my key that has a passphrase. This is the solution we would have invented (and probably have to implement manually until this releases). I sure hope this merges.

seanpk (approved):
fwiw, and assuming that it won't be mergeable without a maintainer approving it, I approve.

PR #9402, opened 2026-08-31
Detect LUKS drives with lsblk so password change works without root

Chessing234:
## Summary
- `blkid -t TYPE=crypto_LUKS` as a normal user returns empty, so the menu claimed no encrypted drives before sudo cryptsetup.
- Discover LUKS devices with `lsblk` (same approach as other Omarchy scripts).

## Test plan
- [ ] `test/shell.d/drive-password-test.sh`
- [ ] On a LUKS install, Update/menu drive password lists the crypt device

### DIFF
--- a/bin/omarchy-drive-password
+++ b/bin/omarchy-drive-password
@@ -3,7 +3,7 @@
 # omarchy:summary=Set a new encryption password for a drive selected.
 # omarchy:requires-sudo=true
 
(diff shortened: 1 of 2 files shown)


Chessing234:

Chessing234:
folded into #12100 — same drive-password binary, one review is enough.

PR #9406, opened 2026-08-31
Let cups-browsed removal proceed when CUPS isn’t running

Chessing234:
## Summary
- Migration `1788009111` only treated `lpstat: No destinations added.` as empty.
- When cups.service is down, `lpstat -v` returns `Scheduler is not running.` and left the migration pending forever.

## Test plan
- [ ] With cups stopped and cups-browsed present, migrate completes and drops the package
- [ ] Real queues still block removal when lpstat succeeds with destinations


Chessing234:

Chessing234:
folded into #12101 — same cups-browsed migration, one review is enough.

PR #9408, opened 2026-08-31
Follow the icon theme inheritance chain when building the app icon index

### LINKED ISSUE #7552: AppLibrary icon index ignores the active icon theme: first filesystem hit wins, so app icons come from hicolor and change between rescans
### System details

Omarchy 4.0.0-1, Hyprland 0.56.2, `quickshell-git 0.3.0.r20.g28771c7-1`, Arch Linux, kernel 7.1.8-arch1-3, Intel i5-8350U / UHD 620. `/usr/share` on btrfs. Icon theme set to `Tela-circle` (`tela-circle-icon-theme-all 2026_07_07-1`).

### What is wrong?

`shell/services/AppLibrary.qml` builds its own icon index and consults it **before** the themed lookup, so the icon theme selected in `org.gnome.desktop.interface icon-theme` is largely ignored for application icons in the menu/launcher.

The scan (`iconIndexScanCommand()`) walks every XDG icon directory and takes every `*/apps/*` and `*/devices/*` file:

```bash
dirs="$HOME/.icons $HOME/.local/share/icons";
IFS=":"; for d in ${XDG_DATA_DIRS:-/usr/local/share:/usr/share}; do dirs="$dirs $d/icons"; done; unset IFS;
for ext in svg png; do
  for base in $dirs; do
    [[ -d $base ]] && find "$base" \( -path "*/apps/*" -o -path "*/devices/*" \) -name "*.$ext" 2>/dev/null;
  done
  find /usr/share/pixmaps -maxdepth 1 -name "*.$ext" 2>/dev/null;
done
```

`indexIconLine()` keys purely on the file's basename and keeps the first hit:

```js
if (name.length > 0 && root.pendingIconIndex[name] === undefined)
  root.pendingIconIndex[name] = value
```

and `iconSource()` gives that index precedence over `Quickshell.iconPath()`:

```js
var found = root.iconIndex[value]
if (found) return Util.fileUrl(found)
var themed = Quickshell.iconPath(value, true)
if (themed.length > 0) return themed
```

Nothing in this path knows which icon theme is active. Whichever theme on disk `find` happens to reach first wins.

**Consequence 1 — the selected theme loses to whatever is found first.** With `icon-theme` set to `Tela-circle`, only 20 of the 67 non-absolute `Icon=` names across my installed desktop entries resolved into `Tela-circle`; 42 came from `hicolor`/`breeze`/`Adwaita`, including names `Tela-circle` does provide:

```
NAME               IN-THEME INDEX PICKS (first hit wins)
foot               yes      /usr/share/icons/hicolor/scalable/apps/foot.svg
mpv                yes      /usr/share/icons/hicolor/scalable/apps/mpv.svg
btop               yes      /usr/share/icons/hicolor/scalable/apps/btop.svg
kdenlive           yes      /usr/share/icons/hicolor/scalable/apps/kdenlive.svg
libreoffice-calc   yes      /usr/share/icons/hicolor/scalable/apps/libreoffice-calc.svg
```

Since `hicolor` holds each application's own upstream icon, the practical effect is that installing an icon theme changes little in the launcher, and which icons *do* change is arbitrary.

**Consequence 2 — the winner is not stable across rescans.** `find` output order is not guaranteed. On btrfs here, five consecutive identical scans produced identical *content* in five different *orders* (`diff <(sort a) <(sort b)` is empty; `diff a b` is not). So the same name can resolve to a different theme on the next scan. Because `refreshIcons()` re-runs the scan whenever a consumer opens — and per #6806 it can fire roughly once a second — icons visibly change with no user action. That is how I noticed this: icons flipped between themed and stock while I was doing nothing.

Note that the code comment describes the index as a *fallback* for icons Qt's themed lookup misses (newly installed apps), but it is currently the primary resolver.

### Steps to recreate

```bash
# 1. Install any icon theme that overlaps hicolor, e.g.
sudo pacman -S tela-circle-icon-theme
gsettings set org.gnome.desktop.interface icon-theme Tela-circle
omarchy restart shell

# 2. Replay the scan exactly as AppLibrary.qml runs it
scan() {
  local dirs="$HOME/.icons $HOME/.local/share/icons"
  IFS=":"; for d in ${XDG_DATA_DIRS:-/usr/local/share:/usr/share}; do dirs="$dirs $d/icons"; done; unset IFS
  for ext in svg png; do
    for base in $dirs; do
      [[ -d $base ]] && find "$base" \( -path "*/apps/*" -o -path "*/devices/*" \) -name "*.$ext" 2>/dev/null
    done
    find /usr/share/pixmaps -maxdepth 1 -name "*.$ext" 2>/dev/null
  done
}
scan > /tmp/s1; scan > /tmp/s2

# 3. The active theme loses to hicolor even though it ships the icon
ls /usr/share/icons/Tela-circle/scalable/apps/foot.svg   # exists
grep -m1 '/foot\.' /tmp/s1                               # -> hicolor/scalable/apps/foot.svg

# 4. Same content, different order on every run
diff <(sort /tmp/s1) <(sort /tmp/s2) | wc -l   # 0
diff /tmp/s1 /tmp/s2 | wc -l                   # non-zero
```

Then open the Apps menu: icons come from a mix of themes, and change between openings.

### Expected

Application icons should come from the icon theme the user selected, and should not change on their own.

### Suggested fix

Either restrict the scan to the active theme plus its `Inherits` chain (then `hicolor`), or — probably simpler — restore the index to the fallback role its comment describes: try `Quickshell.iconPath()` first, since it already honours the icon theme, and fall back to the index only for names the themed lookup misses. Sorting each `find` (`find ... | sort`) would additionally make the outcome deterministic regardless of filesystem traversal order.

### Workaround (verified)

Force the wanted theme to the front of the scan by putting its app/device icons in `$HOME/.icons`, which the scan reads first:

```bash
mkdir -p ~/.icons/tela-priority/scalable/{apps,devices}
find /usr/share/icons/Tela-circle/scalable/apps    -maxdepth 1 -name '*.svg' -exec ln -sf {} ~/.icons/tela-priority/scalable/apps/    \;
find /usr/share/icons/Tela-circle/scalable/devices -maxdepth 1 -name '*.svg' -exec ln -sf {} ~/.icons/tela-priority/scalable/devices/ \;
omarchy restart shell
```

This raised the count from 20 to 46 of 67, deterministically. Two details matter: the scan's `find` has no `-L`, so a symlink to the theme *directory* is invisible to it and the links must be per-file; and the directory deliberately has no `index.theme`, so GTK/Qt do not treat it as a theme.

VykosMolt:
`AppLibrary` builds its icon index by walking every installed theme and taking the first matching basename. Since the index is consulted before the themed lookup, filesystem traversal order can select icons from a theme the user did not choose. With Papirus selected, only 2 of 150 desktop-entry icon names resolved into it; 118 names that Papirus supplied came from other themes.

The index now searches the active theme, its inherited themes in depth-first declaration order, and finally `hicolor`. It reads inheritance from the first `index.theme` in base-directory order and prevents cycles. `hicolor` stays last even when a theme lists it before other parents. Each theme is exhausted before the next, with SVGs before PNGs within a theme.

The index still takes precedence over the themed lookup and stays limited to apps/devices contexts. That avoids collisions with icons such as `zoom` in an actions context. Reverse sorting gives repeatable results and prefers scalable directories; symbolic icons are excluded. If reading the theme through `gsettings` fails, the index uses `hicolor`.

In the recorded comparison, 119 of the 150 names resolved into Papirus, 14 into `hicolor`, three into its parent `breeze`, 13 through `Quickshell.iconPath`, and one into pixmaps. The scan now costs what the active theme's chain costs rather than what happens to be installed: about 52,500 lines in 0.55 seconds here. The old full walk took 0.85 seconds for 104,959 lines when that comparison was recorded, and takes 0.29 seconds for 54,711 lines now that this machine has fewer unrelated themes, so with few themes installed the ordered scan is the slower of the two. Output is identical across runs.

`app-search-test.sh` runs the command embedded in the QML against fixture themes. It covers active and inherited theme precedence, inheritance depth and base-directory order, `hicolor` fallback, theme order across image formats, scalable and symbolic icons, cycles, and repeatable output. Removing the inheritance walk, changing it to breadth-first, or reading a shadowed `index.theme` each fails the corresponding ordering assertion.

`./test/shell` passed 217 of 221 files. The four failures (`config-test`, `runtime-smoke-test`, `snapper-test`, and `unowned-system-paths-test`) also failed on unmodified `quattro`. ShellCheck was clean on the generated command and reported only the existing `SC1091` finding on the test file.

Fixes #7552.

### DIFF
--- a/shell/services/AppLibrary.qml
+++ b/shell/services/AppLibrary.qml
@@ -17,9 +17,12 @@ Item {
   property var desktopHiddenEntryIds: ({})
 
   // Maps an icon name to a file on disk (e.g. "omacut" -> ".../apps/omacut.svg").
-  // Used as a fallback for icons that Qt's themed lookup misses because they were
-  // installed after this process started (its icon cache never re-scans). Refreshed
-  // whenever the app list changes, so newly installed apps get their icon live.
+  // Built by walking the icon theme's inheritance chain and limited to
+  // app/device icons, so a name resolves without the ambiguity an
+  // unconstrained themed lookup carries, and so icons installed after this
+  // process started are found at all (Qt's icon cache never re-scans).
+  // Refreshed whenever the app list changes, so newly installed apps get their
+  // icon live.
   property var iconIndex: ({})
   property var pendingIconIndex: ({})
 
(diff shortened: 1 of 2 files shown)


VykosMolt:
Force-pushed one correctness fix, no change in scope.

The scan had `for ext in svg png` as its outermost loop, so every SVG in the whole search order was emitted before any PNG. That let a scalable icon from an inherited theme — or from hicolor — outrank the configured theme's raster one, which is the same defect this PR is about, one level along. A theme is now exhausted before the next is consulted, SVGs first and then PNGs within it.

New assertion, which fails against the previous revision of this branch:

```
not ok - icon index exhausts a theme before the next one, whatever formats they carry
         .../icons/OmaParentA/scalable/apps/oma-test-format.svg
```

Thirteen assertions now. Output on my machine is unchanged — 52,556 lines, same 119/14/13/3/1 split across Papirus, hicolor, the themed lookup, breeze and pixmaps — because nothing in that chain has a format collision for those names. The fixture is what pins the rule.

I also prototyped resolving each theme's directories from its own `Directories=`/`ScaledDirectories=` instead of walking with `find`, since the spec makes the tree redundant. Measured here it was worse, not better: 119,769 lines and 2.8s against 52,556 and 0.85s, because 51,056 of those lines came from `@2x` directories re-emitting the same files through symlinks. That approach only pays when the scan resolves a fixed set of requested names rather than emitting everything, which is a different change with its own consequences — `iconSource` also serves notification icons for apps that may have no desktop entry. Leaving it out of this one.

PR #9411, opened 2026-08-31
Key Claude limits cache by CLAUDE_CONFIG_DIR

Chessing234:
## Summary
- Scan caches were already per-projects-path; the limits probe used a fixed `claude-limits.json`.
- Key the probe cache by a digest of `config_dir()` so multi-account setups do not reuse another account’s numbers.

## Test plan
- [ ] `test/shell.d/agent-usage-claude-limits-test.sh`
- [ ] Two `CLAUDE_CONFIG_DIR` values get distinct cache files


Chessing234:

Chessing234:
folded into #12265 for easier review.

PR #9417, opened 2026-08-31
Decode wallpapers at the size the screen can show

VykosMolt:
Desktop still wallpapers and their transition frames decode at native file resolution, even when the display can show much less. Their `sourceSize` is now capped at 1.6 times the screen dimensions, including `Screen.devicePixelRatio`, and both transition frames share the base wallpaper's decode size. The extra room accommodates aspect cropping without immediately scaling wide images back up.

The desktop now renders its base wallpaper through `BackgroundMedia`, so that component exposes a `sourceSize` property and forwards it to its still-image loader. Its default retains the existing lock-screen sizing behavior. Video playback, audio selection, and pause conditions continue to use the existing video path.

`background-test.sh` and `video-background-test.sh` pass, including checks for shared transition sizing, the media component's default sizing, and preserved video behavior. `git diff --check` passes. Rebased onto current `quattro`; live graphical verification has not been repeated for this revision.

### DIFF
--- a/shell/Ui/BackgroundMedia.qml
+++ b/shell/Ui/BackgroundMedia.qml
@@ -6,6 +6,7 @@ Item {
 
   property string path: ""
   property int version: 0
+  property size sourceSize: Qt.size(version > 0 ? width : 0, version > 0 ? height : 0)
   property bool playbackEnabled: true
   property bool audioEnabled: false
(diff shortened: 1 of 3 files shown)


bjarneo:
Automated duplication check: this pull request looks similar to #8324, which covers the same wallpaper decode size cap. I keep that one open and close this one to consolidate review.

PR #9440, opened 2026-08-31
Stop the workspace row rebuilding itself for no visible change

VykosMolt:
Creating or leaving an empty workspace rebuilds the whole workspace row, even when the visible IDs have not changed. The row always includes workspaces 1–5, but `workspaceIds()` returns a new array whenever Hyprland's workspace list changes. `Repeater` treats that as a new model and replaces every delegate. `columns` calls the function again.

The row now keeps the same array until its IDs change. Creating workspace 4 leaves `[1,2,3,4,5]` intact; creating a workspace outside that range replaces it. A map built once per update also replaces the lookup each delegate used to perform over the workspace list.

The ID and lookup logic lives in `WorkspacesModel.js`, following the existing model modules so it can run under Node. `workspaces-model-test.sh` checks the fixed 1–5 range, additional workspaces arriving and leaving, invalid entries, and array identity. It also checks that different IDs with the same length are treated as a change and that the QML uses the stable array and lookup map. The test fails on the baseline, which has no model module.

`./test/shell` passed 218 of 222 files. The four failures (`config-test`, `runtime-smoke-test`, `snapper-test`, and `unowned-system-paths-test`) also failed on unmodified `quattro`.

This touches `Workspaces.qml` alongside #9419's click-dispatch change, so whichever lands second may need rebasing.

### DIFF
--- a/shell/plugins/bar/widgets/Workspaces.qml
+++ b/shell/plugins/bar/widgets/Workspaces.qml
@@ -3,32 +3,30 @@ import QtQuick.Layouts
 import Quickshell.Hyprland
 import qs.Commons
 import qs.Ui
+import "WorkspacesModel.js" as WorkspacesModel
 
 BarWidget {
   id: root
   moduleName: "omarchy.workspaces"
 
-  function workspaceById(id) {
-    var values = Hyprland.workspaces.values
-    for (var i = 0; i < values.length; i++) {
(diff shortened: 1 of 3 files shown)


bjarneo:
Automated duplication check: this pull request looks similar to #8794, which covers the same workspace Repeater stability. I keep that one open and close this one to consolidate review.

PR #9454, opened 2026-08-31
Configure systemd-resolved through a drop-in instead of replacing its config

VykosMolt:
Choosing a DNS provider overwrites `/etc/systemd/resolved.conf`, losing any `DNSSEC=`, `Domains=`, `LLMNR=`, or other settings an administrator put there. Choosing DHCP writes another replacement file rather than restoring those settings.

The script now writes `/etc/systemd/resolved.conf.d/20-omarchy-dns.conf` and leaves the main file alone. Each provider clears `DNS=` and `FallbackDNS=` before assigning its servers, because resolved combines these lists across configuration files. Without the empty assignments, the new servers would be added to any old ones still in the main file. This also handles configurations written by earlier Omarchy versions without trying to guess whether the main file is safe to delete.

DHCP keeps a drop-in that clears both lists. Removing the drop-in would reactivate servers an older Omarchy wrote into the main file. Clearing `FallbackDNS=` also disables resolved's built-in fallback servers, so DHCP mode now relies on DNS supplied by the link. That is a behavior change. Clearing `DNS=` also disables reading `/etc/resolv.conf`; NetworkManager supplies per-link DNS over D-Bus, and `reload_dns_stack` requests that update with `nmcli general reload dns-full`.

Every drop-in also names `DNSOverTLS=`, because it decides how the servers selected here are reached: `opportunistic` for Cloudflare and Google, `no` for DHCP and Custom. The old full-file replacement left Custom on resolved's default of `no`; a drop-in that stayed silent would instead inherit an administrator's `DNSOverTLS=yes` and stop a custom server that does not speak DoT from answering. The test pins all four values.

`resolved_dns()` now reads the drop-in when present and uses its last non-empty `DNS=` assignment. Otherwise, a stale Cloudflare entry in the main file would still be reported after selecting DHCP. Machines without a drop-in continue to use the main file for reporting.

`dns-resolved-dropin-test.sh` checks DHCP over a stale main file, a drop-in overriding the old provider, reporting without a drop-in, and an unrecognized provider reporting Custom. These probes bind fixture directories over `/etc/systemd` and `/etc/NetworkManager` inside a user and mount namespace; they skip where unprivileged namespaces are unavailable. Write checks are static because the root path fixes `PATH`: they verify that the main file is never replaced and every provider clears both lists.

`./test/shell` passed 218 of 222 files. The four failures (`config-test`, `runtime-smoke-test`, `snapper-test`, and `unowned-system-paths-test`) also failed on unmodified `quattro`. `dns-sudoers-test.sh` passed. ShellCheck, `bash -n`, and `git diff --check` were clean.

This changes resolved's configuration files. The separate work on saved NetworkManager profile DNS settings is tracked in #8908 and touches the same script.

### DIFF
--- a/bin/omarchy-dns
+++ b/bin/omarchy-dns
@@ -19,6 +19,7 @@ if (( EUID == 0 )); then
 fi
 
 NM_DNS_CONF=/etc/NetworkManager/conf.d/20-omarchy-dns.conf
+RESOLVED_DNS_CONF=/etc/systemd/resolved.conf.d/20-omarchy-dns.conf
 
 provider_from_arg() {
   case "${1:-}" in
@@ -86,15 +87,25 @@ networkmanager_global_dns() {
 }
 
 resolved_dns() {
+  # Read our own drop-in where there is one, the way networkmanager_global_dns
+  # reads its own file. The main file is only consulted on a machine still
+  # carrying settings an older Omarchy wrote directly into it.
+  local file=/etc/systemd/resolved.conf
+  if [[ -f $RESOLVED_DNS_CONF ]]; then
+    file=$RESOLVED_DNS_CONF
+  fi
+
+  # An empty DNS= clears the servers before it, so the effective value is the
+  # last non-empty assignment rather than the first one seen.
   awk -F= '
     /^[[:space:]]*#/ { next }
     /^[[:space:]]*DNS[[:space:]]*=/ {
       value=$0
       sub(/^[^=]*=/, "", value)
-      print value
-      exit
+      if (value ~ /[^[:space:]]/) last = value
     }
-  ' /etc/systemd/resolved.conf 2>/dev/null || true
+    END { if (last != "") print last }
+  ' "$file" 2>/dev/null || true
 }
 
 current_dns_provider() {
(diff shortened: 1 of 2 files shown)


VykosMolt:
Force-pushed one behaviour fix caught in review.

The `Custom` drop-in did not name `DNSOverTLS=`, and that is not the same as the settings this PR deliberately leaves alone. The old code replaced the whole file with no `DNSOverTLS=` line, so Custom ran on resolved's default — `no` on the Arch build here, per the `#DNSOverTLS=no` in the shipped `resolved.conf`. A drop-in that stays silent instead *inherits* whatever the main file says, so an administrator with `DNSOverTLS=yes` would have had arbitrary custom servers contacted strictly over DoT, and a resolver that does not speak it would stop answering entirely.

`DNSOverTLS` governs how the servers selected here are reached, so it belongs to the provider — which is how the other three modes already treated it. Custom now pins `DNSOverTLS=no`, restoring the old effective behaviour while leaving `DNSSEC=`, `Domains=`, `LLMNR=` and the rest untouched.

Pinned in the test so all four modes are stated in one place rather than being read out of four heredocs:

```
Cloudflare=opportunistic
Google=opportunistic
DHCP=no
Custom=no
```

Against the previous revision of this branch that assertion reports `Custom=(unset)` and fails.

Suite unchanged: 218 of 222, same four pre-existing failures; `dns-sudoers-test.sh` passes; shellcheck on the script still clean.

VykosMolt:
Cross-reference for whoever queues these: #8908 also changes `bin/omarchy-dns`, removing the per-profile `nmcli connection modify` writes in favour of the global DNS configuration that already overrides them. Different invariant from this PR — that one is about saved NetworkManager profiles, this one about `/etc/systemd/resolved.conf` — but they touch the same file and will need rebasing against each other.

I had a change for the profile half written and withdrew it on finding #8908, which additionally keeps a guarded one-time cleanup for per-profile DNS left by earlier Quattro builds. Mine did not.

PR #9455, opened 2026-08-31
Fix inverted on/off arguments in omarchy toggle bar

### LINKED ISSUE #9323: omarchy toggle bar on/off arguments are inverted — 'on' hides the bar

perryqh:
## Screenshots

**`omarchy toggle bar on`** — bar visible:
toggle-bar-on

**`omarchy toggle bar off`** — bar hidden:
toggle-bar-off

## Summary
- `bin/omarchy-toggle-bar` forwarded the user's `on`/`off` straight through to the underlying `bar-off` flag, which is named for the hidden state. So `omarchy toggle bar on` set `bar-off` (bar hidden) and `omarchy toggle bar off` cleared it (bar shown) — the opposite of what the words say. The bare `toggle` form was unaffected. Now inverts `on`/`off` before forwarding to the flag.
- `test/shell.d/toggle-test.sh` had encoded the buggy behavior as the expected result; updated to assert the corrected mapping.
- `test/acceptance.d/session-test.sh` used the old mapping too: its EXIT trap restored visibility with `off`, the hide step used `on`, and the reveal step used `off`. Swapped all three so the trap does not hide the bar on exit.
Fixes #9323 
## Test plan
- [x] `test/shell.d/toggle-test.sh` passes with the corrected mapping
- [x] `./test/cli` passes
- [x] Verified visually on a live desktop: `omarchy-toggle-bar off` hides the real bar and `omarchy-toggle-bar on` restores it (screenshots above), state restored afterward

### DIFF
--- a/bin/omarchy-toggle-bar
+++ b/bin/omarchy-toggle-bar
@@ -4,7 +4,17 @@
 # omarchy:args=[toggle|on|off]
 # omarchy:examples=omarchy toggle bar | omarchy toggle bar off | omarchy toggle bar on
 
-omarchy-toggle bar-off "${1:-toggle}"
+ACTION="${1:-toggle}"
+
+# The underlying flag is named for the hidden state, so an explicit on/off
+# from the user is the inverse of what gets passed through. A bare toggle
+# flips the flag either way and needs no inversion.
+case "$ACTION" in
+  on) ACTION="off" ;;
+  off) ACTION="on" ;;
+esac
+
+omarchy-toggle bar-off "$ACTION"
 
 # The shell's watch on the toggles directory can miss flag changes that land in
 # quick succession, stranding the bar off screen until the shell restarts.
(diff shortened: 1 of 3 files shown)


bjarneo:
Automated duplication check: this pull request looks similar to #7023, which covers the same toggle-bar on/off inversion. I keep that one open and close this one to consolidate review.

PR #9460, opened 2026-08-31
[codex] OM-SEC-04: Require package authenticity during Quattro

AFOliveira:
Requires authenticated Omarchy packages from the first Quattro package transaction. The upgrader establishes and verifies the exact packaging-key fingerprint, writes `Required DatabaseOptional`, checks Pacman's effective package trust policy and removes the unauthenticated retry.

Finding: OM-SEC-04. [#8709](https://github.com/omacom/omarchy/pull/8709) overlaps, but its saved Quattro implementation tightens policy after the initial keyring installation. This change establishes the required key and policy before that transaction. Its Apple T2 work remains separate; #9461 handles the T2 provenance path. Keep the package ownership/release ordering in #9070. The closed #9473 is not a dependency and stays closed.

At current-base integration `2412feff`, the focused signature suite passed 5 groups, current Quattro suite 25 groups and CLI 112 groups. Native Pacman fixtures confirm that DatabaseOptional permits an unsigned database while packages remain required/trusted; an unsigned fixture package is rejected without changing the target. Current stable/edge keyring archives matched repository metadata, verified signatures and the pinned embedded key. Two Daybreak reviews found no code or integration blocker.

The exact runtime/settings/keyring artifacts for the coordinated release and the distributed legacy 3.x-to-Quattro entrypoint still need integration validation. No live legacy upgrade was performed here. This PR remains draft and claims only the focused package-authenticity boundary.

PR #9461, opened 2026-08-31
[codex] OM-SEC-05: Remove the unsigned Apple T2 package source

AFOliveira:
Retire Omarchy's unauthenticated Apple T2 package source and require all five replacements from the signed Omarchy repository before setup or migration can complete (OM-SEC-05).

The migration resolves effective policy through `pacman-conf`, including inheritance, includes and directive precedence. `DatabaseOptional` and other database-only exceptions do not cause safe package policy to be removed. An unsafe Omarchy-created `arch-mact2` section is privately backed up and disabled before the package transaction. An unsafe section wholly inside a custom include remains pending with explicit administrator recovery instructions; it is not silently marked repaired.

Affected machines use one fixed, packaged system operation under a machine lock. It rechecks shared state after acquiring the lock, runs without nested sudo prompts, and publishes completion only after a successful reinstall. Unaffected and already repaired machines need no authentication request. Discovery failures remain pending. Both queries and installation use explicit `omarchy/package` targets, and same-version packages are reinstalled because their existing bytes may come from the retired unsigned source.

Fresh setup follows the same policy and repository checks, propagates PCI discovery errors and writes hardware configuration only after authenticated package installation succeeds. The PR remains limited to four files.

Validation on `9fb88358e73e2fce9d6a0a1b67d8ad5707e0f95f`, integrated with quattro `e848f1df`:

- 12 focused test groups passed, including native policy resolution, real `flock` concurrency, query/discovery failures, retries, completion ordering and repository-bound transactions.
- Real isolated Pacman rejected an unsigned fixture under the final mandatory-package-signature policy.
- Six existing T2 hardware fixture groups, 112 CLI checks, Bash syntax and whitespace checks passed.
- Daybreak independently reviewed this revision and found no remaining local code blocker after the serialization and repository-target follow-ups.

This remains a draft. The current stable, rc and edge repository databases contain none of `linux-t2`, `linux-t2-headers`, `apple-t2-audio-config`, `apple-bcm-firmware` or `t2fanrd`. All five signed artifacts, production trust-chain acceptance, package lifecycle and real T2 boot/audio/connectivity/fan behavior remain release gates. Fixture coverage does not establish hardware readiness.

Related: #8709 also changes T2 signature policy, but optional package signatures still permit unsigned packages. This PR requires a complete authenticated replacement set. Quattro bootstrap authenticity is handled separately in #9460.


AFOliveira:
Narrowed and force-pushed 7b666471 after local re-review. The shared OM-SEC-04/05 test is gone; this PR now has a T2-only provenance regression and four files total (328 lines vs 495). It also explicitly handles reinstall failure before marker publication. Compared with #8709, this removes the unauthenticated source and fails closed rather than continuing to accept unsigned packages under PackageOptional. Local evidence on current quattro b71dcad9: real Pacman SigLevel=Never proof succeeds, repository-origin mutation detected, focused provenance and T2 hardware tests pass, CLI 116/116, shell 225/226 with only the base-identical launch-about failure. Keeping the PR draft: all five signed replacement packages are still a release blocker.

PR #9463, opened 2026-08-31
[codex] OM-SEC-08: Publish SSH only after proving key-only access

AFOliveira:
SSH setup now authorizes a valid key and proves effective key-only policy before starting/enabling the service or opening the firewall. Partial failure rolls back newly created state while preserving an existing daemon and administrator configuration. Finding: **OM-SEC-08**, part of the coordinated Omarchy security disclosure.

The fixed root migration validates machine-wide login accounts instead of treating the invoking user's missing key as evidence that nobody can log in. At least one account must have safe key files and be admitted by its effective user/group access controls. Setup proves admission for the specific account whose key it authorizes. Native policy checks cover global and matched authentication directives; service queries distinguish inactive/disabled from errors. Account names ending in `$` are supported.

Unknown administrator files remain untouched. An affected Omarchy-managed legacy service is disabled with recovery guidance when safe access cannot be proven. Exact AllowUsers/DenyUsers/AllowGroups/DenyGroups entries are supported; complex wildcard or host-qualified rules fail closed rather than being guessed. This is a deliberate compatibility limit, and custom PAM/account-expiry policy still needs administrator validation.

**Order: #9457 → this PR; #9474 integrates this corrected implementation after #9467.** Both branches retain the same key, policy, rollback and cold-authorization boundary.

At `1fc0511ca0637f6caa908deee61f7312eb759ddb`, independent Daybreak review accepted the service-state and account-admission corrections. Focused setup/migration fixtures, 112 CLI checks, syntax and diff checks pass. A disposable 4.0.2 VM with the installed candidate, an already keyed account and a locked/keyless second account passed native key-only checks, one-prompt cold migration, replay, active/enabled service retention, and actual fixture-key SSH login. The fixture used the safe final policy; it did not reproduce a vulnerable baseline. Temporary files, account and policy were restored afterward.

Keep draft until the full interactive setup, password-prompt cancellation, custom administrator policies and delivered package layout receive integration validation. The VM results establish the tested policy/replay path, not every possible SSH/PAM configuration.


ErikMelton:
[P1] Earlier per-user migration can disable an already-converted machine
Migration markers are per-user. After one account runs the new migration, it removes `10-omarchy-hardening.conf`. If another account still has `1788124236.sh` pending, that older migration runs first, does not recognize `00-omarchy-key-only.conf`, and disables sshd when the second account lacks an authorized key. The new migration then validates the machine but never re-enables the service, leaving SSH disabled with both migrations marked complete. Please update the older migration to recognize the new config, or retain another machine-wide completion marker.

 [P1] Restricted key records are incorrectly considered usable
`ssh-keygen -lf` only verifies that a line contains parseable key material. It also accepts records containing `cert-authority`, an expired `expiry-time`, `from="!*,*"`, or `command="false"`. With only such records, this function adds the account to `keyed_names` even though no usable administrative login has been proven. Setup uses the same validation and can consequently publish a key-only server without a usable login. Please reject or correctly evaluate authorization options that can prevent or redirect access.

 [P1] Irrelevant passwd entries can shut down SSH
The loop validates every account name and home path before filtering by UID and login shell. A legitimate non-login entry such as `svc.name:x:2:2:Service:/var/empty:/usr/bin/ nologin` therefore makes enumeration fail, which causes an affected active sshd service to be disabled. Please exclude non-login/system accounts before applying the strict validation needed for accounts passed to `sshd -T -C`.

 [P2] Setup cannot recover when the OpenSSH unit is absent
`record_existing_state` runs before `omarchy-pkg-add openssh`. When the package and unit are absent, the systemctl queries return an unrecognized blank or `not-found` state, so setup aborts before reaching the package installation. Please install OpenSSH before snapshotting the service state, or distinguish a missing unit from an actual systemctl query error.

Chessing234:
hey, the sshd key-before-start / key-only publish part overlaps https://github.com/omacom/omarchy/pull/12099. if that was the main goal here, mind closing this as a duplicate (or trimming to whatever is still unique)?

PR #9465, opened 2026-08-31
[codex] OM-SEC-10: Replace the world-writable installer log

AFOliveira:
Replaces the shared writable installer log with a fresh root-owned mode-0600 inode, validates the log and its parent directories, and migrates legacy installer-log state. Support uploads collect through one fixed root reader accepting no caller paths; both fixed log locations are validated before output and identical/prefix copies are deduplicated. Finding: OM-SEC-10.

Desktop collection clears existing sudo authorization, uses one `sudo -N` read phase and revokes on completion/error. The uploader now validates canonical privileged-Bash startup and its shared library before configurable collectors run. Root and nonroot read failures explicitly remove partial output and propagate failure; confirmed absence remains distinct. This also fixes a root-only error hidden by Bash's conditional errexit semantics.

Current candidate `78c26135` integrates Quattro and #9457 by ordinary full merges, with an identical shared helper. Independent Daybreak final review accepts the corrections. Focused logging/read/metadata/partial-failure/startup fixtures, CLI112, syntax and whitespace checks pass. No live logs were uploaded and no host installer log was changed.

#8169 overlaps fresh-inode/log ownership and #8370 overlaps private upload staging. Their overlapping hunks need rebasing or supersession against the final merged implementation; listing them as dependencies does not establish compatible code. Preserve their authors' credit when reconciling those PRs. Fresh ISO installation remains a release integration gate, and the settings-owned upload command requires the coordinated source/package release.


AFOliveira:
Reworked this draft after the review-wide complexity audit.

The branch is now additive to #8169, removes the unrelated `docs/file-layout.md` changes, uses the shared source-only helper from #9457, and reuses #8370's private staging model. I also added exact-head evidence for #8169's stale-descriptor gap and mutation-tested the desktop `sudo -N` reader locally.

Validated at `747ce813`: focused tests pass; CLI is 116/116; shell is 221/222 with only the exact-base `launch-about-test.sh` animation failure.

PR #9469, opened 2026-08-31
[codex] OM-SEC-14: Run update hooks without reusable sudo authority

AFOliveira:
Keeps update hooks, AUR builds, mise and channel hooks outside authorization created by the update workflow. It starts with canonical privileged-Bash/shared-helper validation and cold revocation, uses fixed no-update sudo dispatch, finishes privileged stages before user code, and revokes on errors, cancellation and completion. Finding: OM-SEC-14.

**Reporter credit:** Roger Piñol (R08) and Sean Huber reported earlier update/Pacman authorization-reuse cases; Afonso Oliveira independently disclosed OM-SEC-14 and authored this remediation. Sean also reported the separately tracked Stay Awake occurrence.

Candidate `75e58f03` builds on #9457's exact shared foundation and current Quattro. Earlier feedback about sudo option forwarding, unreachable tests, PATH preservation, hook timing and cleanup is reflected in the final implementation. The deferred pacman-refresh hook runs once after the completed channel transition.

The inhibitor identity protection is now a compatible child PR: #9467 is based on this exact candidate and preserves this updater/helper/wrapper while adding private state, full process identity and cancellation-safe launch. Order: #9457 → #9469 → #9467, then #9472/#9474. #9474 must also integrate #9463's SSH proof transaction. #9473 remains closed; its rejected 163-prompt legacy upgrade is not a successful validation or dependency.

Independent Daybreak review found no new defect in the update/channel authorization boundary. Thirteen focused suites (157 groups), CLI112, syntax and whitespace checks pass. The integrated inhibitor received separate reviews and tests in #9467.

Release gates remain: exact installed package-stack validation, measured normal update/channel prompt counts and cancellation, and #9474's migration security/UX integration. Command-scoped sudo can prompt for each privileged operation; `-y` does not bypass authentication. This source validation does not establish usable legacy upgrade authentication or account-wide isolation from unrelated concurrent sudo sessions.


sayem314:
@robosayem review and audit this properly, and if you have any opinion or suggestions then share it.

robosayem (commented):
Full audit done — I ran the whole thing, not just read it. This is a well-constructed security boundary and I found no exploitable gap. Verdict first: **no blocking findings**, four non-blocking notes, one of which is an interaction worth a decision.

**What the PR does:** update/channel-switch entrypoints start privileged Bash cold (`-p` enforced, decoy `-p` argument rejected), scrub `BASH_ENV`/`ENV`/`SHELLOPTS`/exported functions, pin `OMARCHY_PATH` to the canonical checkout or packaged `/usr/bin`, revoke any pre-existing sudo ticket, then scope all privileged work to `sudo -N` through a PATH-shim wrapper, with `sudo -k` traps on every exit path. The privilege ordering matters: system packages → migrations → AUR → hooks → mise, with a `sudo -k` before user code so a detached malicious child has no future ticket to wait for. This directly closes the R08/OM-SEC-14 class.

**Verification on head 0add43f9:**
- All 12 test suites pass here: security-source-root, security-entrypoint-symlink, update-hook-security, channel-sudo-boundary, update-user-path, update-lock, update-restart-phases, update-sequence, update-disk-space, channel, restart-shell, update-dev. The boundary fixture asserts real properties (no cached credential left, final `sudo -k` in the log, strict phase ordering, decoy `-p` rejection, BASH_ENV and exported-function non-propagation) — not just exit codes
- shellcheck `-S warning` is clean across all nine touched shell entrypoints
- Copilot's four Sep 4 findings are all resolved at this head: the wrapper no longer inserts `--` so it stays transparent to `sudo` options; docs no longer overclaim (the migration-standalone boundary is explicitly scoped out and honest about it); the hostile-yay section reads as fixture guidance inside the rewritten 123-line test, which runs to completion (I ran it); the kernel-compare `==` quoting nit stands — a kernel with glob metacharacters can't occur, cosmetic only

**Non-blocking notes:**

1. **UX consideration, not a bug:** stock Omarchy grants no NOPASSWD for update commands (only dns/theme/timezone), so `sudo -N` + cold start means password prompts **per privileged command** in a fresh-run update: pkg-prune, snapshot, system-pkgs, stay-awake inhibitor, keyring steps each can prompt. That is the security posture chosen and the docs are explicit about it, but the update just went from one password to potentially several. Worth a deliberate call before merge — either accept it in docs/release notes or reconsider after feedback. One mitigating detail I verified: the `sudo` wrapper still honors `-k`/`-K`/`-h`/`-V` transparently, and `omarchy-sudo-passwordless` users can re-enable NOPASSWD during updates.

2. `bin/omarchy-pkg-install`/`omarchy-pkg-aur-install` still source `omarchy-sudo-keepalive`, which runs a background `sudo -v` refresh loop *outside* the update boundary. That's outside this PR's stated scope (interactive installs, not update hooks), but the same OM-SEC-14 class — worth a follow-up card so it isn't lost.

3. The stay-awake rewrite now drops privilege via `setpriv --reuid` after a single foreground `sudo -N` authentication, writes the inhibit PID + start-time from inside the child, and `stop` revalidates both before killing — PID-reuse spoofing by a same-UID process is bounded correctly. The 5 s/100-attempt readiness polling caps are sensible. Nice work.

4. `omarchy-security-functions` includes an unused-but-exported `omarchy_security_sudo_supports_no_update` public path (used internally by the wrapper check) — harmless; the library's "internal only, exit 64 if sourced as a program" guard is a good touch.

Ship it. This is exactly the remediation pattern the other sudo-reuse findings (OM-SEC-17, 21, 23) should follow.

ErikMelton:
LGTM, pending hook doc fix.

ErikMelton (approved):

ErikMelton (changes requested):
- [P2] Notification readiness blocks lock recovery — bin/omarchy-restart-shell:86–87
    https://github.com/omacom/omarchy/blob/13a4306a8edd8708d8e2d5814593312a8f321921/bin/omarchy-restart-shell#L86

When recovering a locked session whose lock client died, the new condition prevents relock_session from running unless notifications also become ready. If that plugin fails or takes longer than the polling window, the command exits and leaves the user stranded behind Hyprland’s failsafe—even when the lock service works. Run lock recovery independently of notification readiness. I reproduced this with isolated stubs: the PR exits without restoring the lock; the previous condition restores it successfully.

AFOliveira:
Fixed in b34b117f, now at 689771bd. The restart waits for core IPC, re-secures the lock immediately, and only then waits for a previously running notification service, exiting 1 with its own message if that never returns; the restart suite covers the never-returning case.

Follow-up on the same head: cleanup traps are installed before the entry `sudo -k`, the notification probe is bounded to 1s instead of busctl's 25s default, and the branch carries the current #9457 head. All 15 owned suites and `./test/cli` pass; independent review is CLEAR at 43b91163 plus the #9457 merge.

PR #9471, opened 2026-08-31
[codex] OM-SEC-16: Validate source roots before privileged publication

AFOliveira:
## Summary

A caller-controlled `OMARCHY_PATH` could select files later copied or installed by root.

Finding: **OM-SEC-16**

## Security impact

Attacker-selected source bytes can cross into root-owned system state, allowing privileged configuration replacement or code execution.

## Remediation

Privileged publication uses canonical fixed or explicitly authorized roots with root-owned, non-writable ancestry; runtime scripts no longer publish generic static system files.

The reviewed implementation applies these concrete controls:

- Privileged workflows derive the source root from root-owned /etc/omarchy.conf or the fixed packaged tree.
- The exact config grammar, canonical path, ownership, modes, links, and ancestor chain are checked.
- Bash privileged startup suppresses BASH_ENV/exported-function injection where required.
- Root consumes validated source through inherited descriptors rather than reopening mutable paths.
- Package lists validate names and use an option terminator.
- Static executable root state is package-owned; runtime commands no longer publish the sleep hooks.
- omarchy-settings and omarchy-settings-dev both own the live static hooks.

## Validation

The baseline publisher accepted attacker paths; trusted-root, symlink, ownership, package-payload, and dev-link regressions pass.

**Detailed regression coverage:** trusted-source-publication-security-test.sh, unowned-system-paths-test.sh, hybrid-gpu-test.sh, quattro-ownership-bridge-test.sh, and the extracted archives pass.

Current status: Fixed. An explicitly root-authorized dev link intentionally trusts that user's checkout.

All changed shell files on this branch pass `bash -n`, and `git diff --check` passes.

## Coordination

This is one finding in a coordinated 21-finding disclosure sent to the Omarchy security team. The corresponding Markdown report contains the affected-code analysis and safe reproduction.

The matching `omarchy-settings` payload ownership change must ship in the coordinated release.

### DIFF
--- a/bin/omarchy-channel-set
+++ b/bin/omarchy-channel-set
@@ -1,10 +1,57 @@
-#!/bin/bash
+#!/bin/bash -p
 
 # omarchy:summary=Set the Omarchy package channel.
 # omarchy:args=<stable|rc|edge|dev>
 # omarchy:requires-sudo=true
 
+if [[ $- != *p* ]]; then
+  echo "Refusing an unsafe Bash startup for channel switching." >&2
+  exit 126
+fi
+
+require_privileged_bash_startup() {
+  [[ $- == *p* ]] || return 1
+  /usr/bin/env -i /usr/bin/bash -p -c '
+    [[ $1 =~ ^[1-9][0-9]*$ ]] || exit 1
+    mapfile -d "" -t argv <"/proc/$1/cmdline" || exit 1
+    executable=$(/usr/bin/readlink -e -- "/proc/$1/exe") || exit 1
+    [[ $executable == /usr/bin/bash ]]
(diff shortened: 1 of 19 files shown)


AFOliveira:
Withdrawing this as an independent security finding after re-checking the privilege boundary and the earlier #9317/#9351 analyses. The local proof confirms that caller-selected files reach sudo copy/install operands, but the caller is the same wheel user authorizing sudo and can already invoke the equivalent sudo operation directly. A cached sudo credential does not change that conclusion: any process running as that user can use sudo -n directly, so OMARCHY_PATH does not confer an additional capability. Trusted-source consolidation may still be useful defense-in-depth, but this 18-file security PR is not the right vehicle. I will keep the useful design notes locally and not count OM-SEC-16 among independently validated vulnerabilities.

PR #9472, opened 2026-08-31
[codex] OM-SEC-17: Allowlist update restart markers and commands

AFOliveira:
Writable restart markers select only supported, zero-argument component commands through an explicit allowlist. Commands resolve inside the validated source tree, including matching packaged links to `/usr/bin`. Unsupported names, generic dispatchers and non-regular markers cannot select a command. Failed restarts keep their marker for retry; a shell marker causes only the single normal shell refresh.

Finding: **OM-SEC-17**, reported and remediated by Afonso Oliveira. The #9469 prerequisite retains Roger Piñol's and Sean Huber's credits for their update/Pacman and Stay Awake reports.

**Order: #9457 → #9469 → #9467 → this PR.** The integration includes the actual #9467 `6af052fc` commit through an ordinary merge, including first-upgrade compatibility and the isolated idle-toggle fixture. Shared authorization code, the updater, inhibitor and their tests are byte-identical to that prerequisite. Erik's hook-documentation feedback is addressed by the current foundation's manual, reference and sample hook timing. #9474 remains a separate downstream migration change; #9473 is closed.

At `b9336ce8`, this PR's own delta over #9467 is four files: the restart dispatcher, its documentation and two restart test files. Seven focused suites pass both locally and in a fresh vanilla Arch VM: 111 groups, plus 112 CLI checks, with no skipped cross-account coverage. The VM had no installed Omarchy commands when the suites ran.

Native tests on the same VM exercised the actual restart dispatcher and tmux in both checkout and packaged symlink layouts. Supported markers reload the real tmux configuration; malicious PATH commands and unsupported markers are not dispatched; failed reloads retain their marker and succeed on retry; non-regular markers are refused; cached sudo authorization is revoked. This headless guest also exercises the non-blocking missing-desktop path; it does not establish graphical shell or every hardware component restart.

Daybreak's final scoped review of exact `b9336ce8` found no blocker; inherited SSH/AUR changes were excluded. Full historical update/migration prompt counts and the coordinated runtime/settings release remain separate validation work; earlier excessive prompt counts are not release-readiness evidence.


ErikMelton:
2 Changes:

1. The new sudo wrapper breaks option-bearing calls. default/omarchy/sudo-no-update/sudo executes:

  exec /usr/bin/sudo -N -- "$@"

The -- makes subsequent arguments command arguments rather than sudo options. For example, omarchy-update-stay-awake calls sudo -v, which becomes:

  /usr/bin/sudo -N -- -v

This attempts to execute a command named -v and aborts a normal terminal update under set -e. The wrapper needs to preserve supported sudo options, with regression coverage for sudo -v.


2. The original user PATH is lost before hooks and mise run. omarchy-update saves user_path, replaces and exports PATH, and then re-executes itself through script. Because user_path is not preserved across that re-exec, the new process captures the already-sanitized path. The later PATH="$user_path" calls therefore do not restore paths such as ~/.local/bin. Post-update hooks can silently lose commands, and a user-installed mise may be skipped.

ErikMelton:
LGTM, pending hook doc fix.

PR #9473, opened 2026-08-31
[codex] OM-SEC-18: Bound Quattro sudo before mixed-trust code

AFOliveira:
The legacy Quattro upgrader must complete its system phase without maintaining reusable sudo authorization across user-controlled or downloaded code. It now uses command-scoped authentication and an irreversible transition to user work.

Finding: **OM-SEC-18**.

**Reporter credit:** Reported and remediated by Afonso Oliveira. The included #9469/#9474 prerequisites retain Roger Piñol's and Sean Huber's credits for their respective update, migration/SSH and Stay Awake reports.

- Remove global sudo validation/keepalive; invalidate first and use `sudo -N` for each system operation.
- Create the pre-upgrade snapshot through fixed-path operations after initial revocation and before mutation, including legacy installations without the packaged snapshot helper.
- Finish system/package operations before closing the privilege gate. Later `as_root` calls or attempts to reopen it fail.
- Revoke on success, failure and EXIT/HUP/INT/TERM, including after user work. Failed final revocation prevents completion and reboot.
- Use the protected packaged migration and AUR paths after handoff; request reboot through systemd/logind.
- Retain the self-contained privileged-Bash bootstrap needed before the shared library is installed.

The independent Daybreak model review found no separate defect in the Quattro gate. Its channel and source-root findings were fixed in the prerequisites and merged here; the final reassessment accepts those changes for maintainer review. Erik's snapshot feedback is addressed.

Includes both prerequisites by normal merges. **Merge #9469, then #9474, then #9473.** Compare with #9474 for the Quattro-specific delta. Package authenticity (#9460), AUR consent (#9795) and package ownership (#9070) remain separate integration work.

Reviewed head: `e749a877ab49551c18f498ebf4cd7b06aa1e2d7d`.

Validation: the final combined revision passes `./test/all` (CLI and **242 shell test files**), 36 changed Bash files pass syntax checks, and the four PRs merge cleanly together. Focused coverage includes **16 channel-lifecycle cases**, **38 restart assertions**, source-root/package-layout/symlink checks, migration retry/failure/signals, SSH setup and the existing Quattro privilege/snapshot tests. The aggregate uses the packaging/ISO checkout overrides with `NO_COLOR` unset and Python bytecode disabled.

Signed candidate packages `4.0.2-9004` built from the combined source passed a real update on a disposable 4.0.2 VM and a complete **3.8.4 → Quattro** upgrade with reboot. The latter finished in 286.1 seconds, retained the pre-upgrade snapshot, had no pending migrations or failed system/user services, and booted a healthy desktop. Eleven installed security-related files match the candidate hashes in both guests. Installed and development-checkout restart checks also pass.

**Ready for maintainer review.** The legacy upgrade required **163 real password prompts** under the guest's normal authentication policy; the test driver supplied its fixture password. Maintainers must accept that usability cost or request a separate batching design before merging the complete upgrade experience. The hook timing change also needs acceptance. The protected workflow prevents its own authentication from refreshing sudo credentials; it does not provide account-wide isolation from unrelated concurrent authentication.


AFOliveira:
Addressed both review threads in d0b4d997:

- Removed the repository-wide `--overwrite` count. The regression now inspects only `omarchy-upgrade-to-quattro`, ignores comments, requires exactly one occurrence, verifies that it is the settings ownership bridge, and separately proves that the later system upgrade contains no overwrite.
- Documented the narrow `#!/bin/bash -p` convention exception in `AGENTS.md`. The boundary suite already rejects ordinary Bash startup with a decoy post-script `-p`, so the exception is both explained and regression-covered.

Local validation on d0b4d997:

- changed-shell `bash -n` and `git diff --check`: pass
- `upgrade-to-quattro-test.sh`: all 26 cases pass
- `quattro-sudo-boundary-security-test.sh`: pass, including its mutation checks
- CLI: all 116 cases pass
- full suite: 221/222 shell files pass; only `launch-about-test.sh` fails with the same `a roomy window animates` result reproduced on the exact base

This PR remains draft while its shared startup/sudo helpers are reduced onto the common foundation.

AFOliveira:
The reviewer-driven scope reduction is published in 9340e962. The PR now contains only the Quattro sudo-lifetime fix, its focused regression, and the documented privileged-Bash exception. Package-signature/T2/overwrite/SDDM/snapshot changes and the unused wrapper were removed (201 deletions vs 135 additions relative to the prior head). Local validation on the pushed commit: standalone baseline exploit confirmed; real boundary test passes; replacing `sudo -N` with ordinary sudo recreates detached credential theft; unchanged Quattro suite passes all 24 cases; CLI 116/116; full shell suite 221/222 with only the exact baseline-identical launch-about UI failure. The inline startup bootstrap remains because this upgrader must be streamable before a shared library exists. Draft status is retained for #9474 sequencing.

ErikMelton:
One last thing!

When /usr/bin/omarchy-snapshot exists, it uses ordinary sudo repeatedly, creating a reusable timestamp that a detached same-user process can consume before open_privilege_window invalidates it. On the intended Omarchy 3.8 upgrade path, commands instead live under $OMARCHY_PATH/bin, so the absolute /usr/bin path is absent and status 127 is accepted, which skips the snapshot entirely. The snapshot should be implemented using trusted fixed-path operations through as_root/sudo -N after cold invalidation but before the first mutation. The regression test should exercise this path explicitly.

ErikMelton:
Looks good to me, pending the other PRs.

PR #9474, opened 2026-08-31
[codex] OM-SEC-19: Protect migration and SSH setup authorization

AFOliveira:
Keep the ordered migration queue and SSH setup outside reusable sudo lifetime. Startup and source-root validation happen before selecting code; authentication is revoked before work, after each successful migration, and on failure or catchable signals. Completion is published only after both work and revocation succeed. Direct package-helper calls preserve the same boundary.

Finding: **OM-SEC-19**. Sean Huber reported inherited Bash-state authorization reuse in migrations and SSH setup. Afonso Oliveira independently disclosed the migration lifetime and direct package-helper paths as OM-SEC-19 and authored this remediation.

**Order: #9457 → #9469 → #9467 → #9472 → this PR, together with #9463.** This branch includes validated #9472 `b9336ce8` through an ordinary merge and retains the published #9463 `d72e607b` SSH implementation. No unpublished SSH redesign or later local SSH follow-ups are included. #9473 remains closed. Reconcile separate AUR/migration work in #9795 and override handling in #10022 when integrating them.

Four machine repairs use a fixed installed transaction each: T2 defaults, Bluetooth power persistence, FIDO2 authfile ownership and CUPS hardening. Each has literal targets, closed machine arguments, a clean environment, root-side serialization/recheck and no nested sudo. Discovery errors and failed mutations remain retryable. Missing service units and an inactive Bluetooth daemon now complete safely rather than blocking every later migration. AUR helpers pin yay's sudo, flags, credential loop and pacman/configuration under the boundary; an update stops if its package transaction removes that boundary. Kernel migration tests now contain absolute sudo/pacman calls within their fixtures.

At `cdaaeeb2`, 18 focused suites pass locally and in a dedicated vanilla Arch VM: 149 groups plus 112 CLI checks. Package lifecycle checks use the current published #257 `9b9a9c8e`. Actual installed Bluetooth/CUPS migrations reproduce the old failure and pass with the fix, both before and after stock packages are installed. The two first-time repairs need two password prompts; replays and a second account with no sudo membership need none. Separate fixture queues using the real migration runner and real sudo verify cold entry, command-scoped authentication, failure/TERM cleanup and pending markers. Native yay 13.0.1 installs a small repository package through the exact AUR helper despite hostile yay configuration, without running configured decoys or leaving a ticket. This is not a claim about an AUR build or the complete historical queue.

Daybreak caught a failed-service state being mistaken for inactive. The fix now checks exact states; a real failed systemd unit reproduces the old error and confirms refusal/recovery. Daybreak reviewed the correction at exact `cdaaeeb2`, closed the finding and approved the scoped follow-ups. No remaining scoped findings.

**Still draft.** The SSH dependency/design discussion is unresolved, and later local SSH fixes are not part of this head. The historical 58-prompt migration queue has not been remeasured on a complete legacy upgrade; the earlier 163-prompt Quattro run is an unacceptable UX result. Native T2/initramfs, FIDO2, physical Bluetooth/printer behavior, full-queue cancellation/retry and the coordinated runtime/settings release still need validation. These targeted passes do not establish release readiness or justify restoring reusable authorization.


AFOliveira:
Reworked after the shared-helper review: force-pushed 94a84d71. The PR now excludes the rejected trusted-source-root proposal from closed #9351 and is limited to the real migration sudo-lifetime boundary. The shared libraries, package helper, and no-update wrapper are byte-identical prerequisite copies from #9457/#9470/#9469. Local verification was rerun on the audit baseline and after a clean cherry-pick onto current quattro b71dcad9: exploit baseline succeeds/candidate blocks, focused 11/11, mutation detected, CLI 116/116, shell 224/225 with only the base-identical launch-about animation failure.

ErikMelton:
One issue I found:

The second revocation runs only after every migration succeeds. With set -e, a failed or interrupted migration exits before line 158. If migration code bypasses the PATH wrapper, for example through /usr/bin/sudo or another PATH-resetting helper, it can leave a reusable timestamp behind after the aborted queue. This preserves part of credential-leak scenario the PR is intended to close.

PR #9475, opened 2026-08-31
[codex] OM-SEC-21: Authenticate only after package picker code exits

AFOliveira:
Package selection runs with cold sudo authorization. After discovery and the configurable picker finish, only a validated selection reaches one fixed command-scoped Pacman transaction. Finding: OM-SEC-21.

Both install/remove entrypoints resolve their canonical shared library, validate privileged Bash startup, sanitize inherited Bash state and bind the source root before invoking configurable picker behavior. The shared library remains byte-identical to #9457's approved `ea617b91` version. Authorization is revoked on completion, cancellation and failure.

Query failures are checked before starting fzf. Empty selection, no match and picker cancellation are clean no-ops; partial output from a failed/cancelled picker is discarded. Only status 0 plus validated package names reaches the fixed Pacman argument array. #9795/#9227's AUR review/confirmation changes are separate scope; this picker uses repository packages.

Integrated current `quattro` (`8f324c90`) through an ordinary merge at `41e6882a`. The two conflicts preserve upstream's success/failure presentation: capture the command-scoped transaction result, pass it to `omarchy-show-done`, and return that result. No keepalive or xargs dispatch is restored. This reconciliation adds four net production lines and five net fixture lines; other changes in the merge come directly from upstream.

Validation at `41e6882a`: all four focused regression groups and 112 CLI checks pass, along with Bash syntax and whitespace checks. The regressions assert exact transaction arguments, success/failure display status, cleanup, cancellation, query/picker failures, initial/final revocation failures, source-root mismatch and inherited startup state. Fresh independent Daybreak review of the reconciliation returned PASS with no findings; its targeted status, cancellation and cleanup mutants were killed.

Native verification on 21 September used exact installed candidate commands in a disposable Omarchy VM with real sudo, fzf, Pacman and yay. Installing/removing `sl` through the CLI router uses one password prompt and shows Done. A controlled Pacman-lock failure in each command shows Failed with exit 1 and preserves failure. Cancelling authentication, Esc, Ctrl-C in the picker, and empty selections preserve cleanup. The background authorization-reuse witness never obtained root on the candidate. Earlier same-day native controls reproduced reuse when only `-N` was removed. The guest was cleaned up and shut down.

Depends on #9457 and its coordinated shared-helper release with omarchy-pkgs#257. These are installed-command tests on the disposable VM, not a build or certification of the final coordinated package pair. The broader shell run was not fully green: five cross-repository suites lacked their external omarchy-pkgs checkout, and 16 environment-dependent suites skipped. The scoped package-picker suite passed. AUR execution and alternative sudo policies remain outside this PR's scope.

History consolidated at `a2dbf117`: one complete package-picker fix with its regression tests, following the upstream integration merge `b2341a06`. The five inherited #9457 foundation commits are unchanged. The complete repository tree is byte-identical to the previously reviewed and VM-tested `41e6882a`; this is a history-only rewrite. The four focused regression groups and all 112 CLI checks were rerun successfully on the consolidated head.


AFOliveira:
Reworked in f1282ada after comparing the finding with current `quattro` and #9227. The PR is reduced from 10 files / 1,280 additions to a four-file, 421-addition diff, of which 276 lines are the isolated namespace regression and 87 are the shared #9457 library. All AUR/PKGBUILD, developer packaging, generic package-helper, wrapper, and unrelated-finding changes were removed. The original attack reproduces on the baseline, the fixed test passes, and its mutation restores the exploit. Full local validation details are now in the PR body. Keeping this draft until #9457 lands so the duplicate shared-library file can drop out cleanly.

PR #9478, opened 2026-08-31
Route Chromium notifications through the system center

Traap:
## Summary

- Enable Chromium's `SystemNotifications` feature in the shared browser flags so
  Chromium- family notifications use Omarchy's freedesktop notification service
  instead of opening a browser-owned tiled window.
- Migrate existing Chromium, Chrome, Edge, Brave, and Brave Origin flag files.
- Preserve custom enabled features and explicit `SystemNotifications` opt-outs.
- Add regression coverage for migration safety and idempotence.

## Verification

- `bash test/shell.d/chromium-system-notifications-migration-test.sh`
- `bash -n migrations/1788129995.sh test/shell.d/chromium-system-notifications-migration-
test.sh`
- `git diff --check`

## Diagnosis

In a running Quattro session, Quickshell owned `org.freedesktop.Notifications`,
but Chromium and Brave were started without Chromium's `SystemNotifications`
feature. Browser notifications consequently used Chromium's fallback
message-center window, which Hyprland tiled as a normal rowser window.

Explicitly enabling `SystemNotifications` routes these notifications through
Omarchy's existing notification center.

---
🤖 Built with [OpenAI Codex](https://developers.openai.com/codex/).

### DIFF
--- a/config/chromium-flags.conf
+++ b/config/chromium-flags.conf
@@ -1,5 +1,5 @@
 --ozone-platform=wayland
 --ozone-platform-hint=wayland
 --password-store=gnome-libsecret
---enable-features=TouchpadOverscrollHistoryNavigation
+--enable-features=TouchpadOverscrollHistoryNavigation,SystemNotifications
 --load-extension=/usr/share/omarchy/default/chromium/extensions/copy-url,/usr/share/omarchy/default/chromium/extensions/yt-dlp,/usr/share/omarchy/default/chromium/extensions/whatsapp-slim
--- a/migrations/1788129995.sh
+++ b/migrations/1788129995.sh
@@ -1,0 +1,21 @@
+echo "Route Chromium browser notifications through the system notification center"
+
+flag_files=(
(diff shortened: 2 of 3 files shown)


sprajs:
The append path corrupts an existing flags file when its last line has no trailing newline. I ran the exact added migration with an isolated HOME containing `chromium-flags.conf` with just `--ozone-platform=wayland` (no final newline). It exits 0 but produces `--ozone-platform=wayland--enable-features=SystemNotifications` on one line, altering the existing option instead of adding a new flag.

Please ensure a line boundary before appending and add a regression for an existing flags file without a final newline, including a second migration run.


I implemented a fix with regression tests in https://github.com/omacom/omarchy/pull/12421, in the main Omarchy repository, preserving this PR’s original commits. The focused tests pass; the PR records the full-suite failures and remaining runtime validation.

Traap:
I am closing this PR in favor of #12421.

Traap:
Please use #12421

PR #9482, opened 2026-08-31
Count GPUs without waking the one that is asleep

VykosMolt:
Opening the menu can wake a suspended discrete GPU because the Hybrid GPU row calls `omarchy-hw-hybrid-gpu`, whose fallback counts lines from `lspci`. Reading PCI config space resumes the card. The fallback now counts display-class devices (`0x03`) through cached sysfs values, as the NVIDIA detector already does.

The old `VGA|3D|Display` pattern also counted `Non-VGA unclassified device` as a graphics card. A machine with one GPU and one such device could therefore be reported as hybrid. Counting by PCI class excludes it. The supergfxctl path, its timeout handling, and the fallback when supergfxd cannot answer are unchanged.

On an Arrow Lake iGPU + RTX 5070 Ti Mobile laptop, starting with the discrete card suspended:

```
          verdict   time      card afterwards
quattro   hybrid    2659 ms   active
patched   hybrid       6 ms   suspended
```

The test now supplies a fake PCI tree instead of a stubbed `lspci`, including an unclassified device alongside the GPUs. It retains the eight existing cases and adds checks that one GPU beside an unclassified device is not hybrid and that the detector never calls `lspci`. The converted test fails on unmodified `quattro`.

`./test/shell` passed 217 of 221 files. The four failures (`config-test`, `runtime-smoke-test`, `snapper-test`, and `unowned-system-paths-test`) also failed on unmodified `quattro`.

### DIFF
--- a/bin/omarchy-hw-hybrid-gpu
+++ b/bin/omarchy-hw-hybrid-gpu
@@ -2,8 +2,29 @@
 
 # omarchy:summary=Detect whether the system has an active hybrid GPU configuration
 
+# Read the cached sysfs IDs rather than lspci, which reads PCI config space and
+# resumes a runtime-suspended GPU. This gate is what the Hybrid GPU menu row is
+# guarded by, so asking it used to wake the discrete card the row is about --
+# 2.6s on this laptop, and the card then stays powered. omarchy-hw-nvidia
+# already reads the same files for the same reason.
+#
+# Counting by class also stops "Non-VGA unclassified device" being counted as a
+# GPU: the old pattern matched the VGA inside Non-VGA, so a machine with one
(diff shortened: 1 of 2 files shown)


sprajs:
Cross-linking #11038: its current `omarchy-hw-hybrid-gpu` implementation is identical to this one apart from comments, including the sysfs `0x03*` scan and bounded supergfxctl fallback. That PR also changes the NVIDIA GLX environment handling, which is a separate part of its scope.

The detector fix only needs to land once. Keeping this focused PR and rebasing #11038 onto it, or consolidating the shared detector/tests there, would avoid maintaining two copies of the same change.

Issue #8989, opened 2026-08-29
Hybrid iGPU-primary laptops: nvidia.lua forces NVIDIA env session-wide — video corruption (root cause of #4901) and blocked dGPU runtime suspend


karluiz:
## System details

- **CPU:** Intel Core Ultra 9 275HX (Arrow Lake)
- **GPU (integrated):** Intel Arrow Lake-S — drives the internal eDP panel
- **GPU (discrete):** NVIDIA GeForce RTX 5080 Max-Q (no display connected; HDMI/one DP are wired to it)
- **Drivers:** nvidia-open-dkms 610.57.04, Mesa 26.2.1, intel-media-driver 26.2.4, libva-nvidia-driver 0.0.17
- **Omarchy:** 4.0.1
- **Kernel:** 7.1.9-arch1-2
- **Laptop:** Lenovo Legion 7i Pro (99.9 Wh battery)

## Problem

On hybrid laptops where the **iGPU drives the panel**, `default/hypr/nvidia.lua` detects the NVIDIA GPU and unconditionally sets, during config parse:

```lua
hl.env("NVD_BACKEND", "direct")
hl.env("LIBVA_DRIVER_NAME", "nvidia")
hl.env("__GLX_VENDOR_LIBRARY_NAME", "nvidia")
```

Then `default/hypr/autostart.lua` exports the resulting environment session-wide on `hyprland.start`:

```lua
hl.exec_cmd("systemctl --user import-environment $(env | cut -d'=' -f 1)")
hl.exec_cmd("dbus-update-activation-environment --systemd --all")
```

So every app (launcher, web apps, Electron apps, browsers) inherits `LIBVA_DRIVER_NAME=nvidia` and `__GLX_VENDOR_LIBRARY_NAME=nvidia`. On an iGPU-primary hybrid this has three consequences:

### 1. Cross-GPU video corruption (the root cause behind #4901 / #3899)

Chromium/Electron apps decode video on the NVIDIA GPU (NVDEC via libva-nvidia-driver) while presenting on the Intel GPU. Importing those DMA-BUFs into the iGPU fails (`eglCreateImage → EGL_BAD_MATCH`), producing corrupted video, ghosting while scrolling, and stuttering animations. #4901 documents the Chromium symptoms and workarounds in depth; the reason `vainfo` defaults to NVDEC in the first place is this `hl.env` in `nvidia.lua`.

[...]

Issue #8726, opened 2026-08-28
Hybrid AMD+NVIDIA laptop: forcing LIBVA_DRIVER_NAME=nvidia breaks all hardware-decoded video (black video players)


ArghyaRanjanDas:
## What happens

On a hybrid AMD + NVIDIA laptop where the internal panel is driven by the AMD iGPU,
Omarchy sets `LIBVA_DRIVER_NAME=nvidia`, which breaks **all hardware-decoded video**.
Video players render as a black rectangle. The media itself loads and decodes fine —
in Chromium the `<video>` element reports `readyState=4`, `error=null`, correct
`videoWidth/videoHeight` — the frames simply never reach the screen.

This affects any page with a `<video>` element (GitHub video attachments, YouTube, etc.)
and is present from install, on every boot.

## Expected

Hardware-decoded video renders.

## Root cause

`/usr/share/omarchy/default/hypr/nvidia.lua` sets, whenever a GSP-capable NVIDIA GPU
is detected:

```lua
hl.env("NVD_BACKEND", "direct")
hl.env("LIBVA_DRIVER_NAME", "nvidia")
hl.env("__GLX_VENDOR_LIBRARY_NAME", "nvidia")
```

The detector checks only *whether* an NVIDIA GPU exists — not whether it is the GPU the
compositor renders on. On this machine the connectors are split across both GPUs:

```
card2-eDP-1     (internal panel)  -> 0000:06:00.0  AMD HawkPoint2 iGPU
card1-HDMI-A-1  (external)        -> 0000:01:00.0  NVIDIA RTX 4050 Mobile
```

Hyprland composites on the AMD iGPU. Forcing `LIBVA_DRIVER_NAME=nvidia` routes VA-API
decode to NVDEC on the dGPU, and NVDEC output carries an NVIDIA-vendor DRM modifier that
`radeonsi` cannot import. The zero-copy path from decoder to compositor is therefore
structurally impossible, and decoded frames are dropped instead of displayed.

The same cross-GPU import failure shows up in Hyprland's logs as repeated failed
hardware-cursor blits (`GL_INVALID_OPERATION`), which force a software cursor.

## Steps to reproduce

1. Hybrid AMD + NVIDIA laptop, internal panel wired to the AMD iGPU.
2. Install Omarchy; confirm `LIBVA_DRIVER_NAME=nvidia` is exported into the session.
3. Open any page with an HTML5 `<video>` element in Chromium.
[...]

PR #9483, opened 2026-08-31
Only point the session at NVIDIA when NVIDIA is driving the screen

VykosMolt:
`nvidia.lua` sets `NVD_BACKEND`, `LIBVA_DRIVER_NAME`, and `__GLX_VENDOR_LIBRARY_NAME` whenever the machine has an NVIDIA GPU. On a hybrid laptop whose display is driven by the integrated GPU, those session-wide variables still send GL and VA-API clients to the discrete card. The first client has to wake it, and it stays powered while in use.

`nvidia.lua` now also checks `omarchy-hw-nvidia-drives-display`. The new detector answers yes when an NVIDIA connector reports `connected` and no built-in panel (`eDP`, `LVDS` or `DSI`) on another card does. A panel on another card means the compositor renders there, and `hl.env` is session-wide, so docking a laptop whose external ports are wired to the discrete card does not name NVIDIA. An NVIDIA-only machine, a laptop with its panel on the NVIDIA card, and a desktop with a monitor on NVIDIA still get the variables. It reads the driver's cached connector state; the measured check took 3 ms and left the suspended card asleep. PRIME offload remains available when the integrated GPU is driving the display.

The comparison used an Arrow Lake iGPU driving `eDP-1` and an RTX 5070 Ti Mobile with no connected output. With the variables explicitly removed using `env -u`, `glxinfo -B` reported the Intel renderer; with them set, it reported NVIDIA. Cold runs waited for the discrete card to return to suspend between measurements:

```
                        run 1     run 2    card afterwards
without the variables    75 ms     85 ms   suspended
with the variables     2675 ms   2733 ms   active
```

With the card already awake, the same comparison was 80 ms versus 153 ms.

Ten new assertions in `hw-nvidia-test.sh` cover an idle discrete card, the docked hybrid laptop, a panel on the NVIDIA card, desktops with an unused integrated card and with monitors on both cards, an NVIDIA-only machine, absent and connector-less NVIDIA cards, a DRM path containing a dash, and the requirement for both detectors in `nvidia.lua`. They fail on the baseline, which has no display-driving detector. The docked case was confirmed on live hardware in the thread below (Dell G15 5515, both external outputs on NVIDIA, panel on AMD).

`./test/shell` passed 217 of 221 files. The four failures (`config-test`, `runtime-smoke-test`, `snapper-test`, and `unowned-system-paths-test`) also failed on unmodified `quattro`.

This is independent of #9482, which changes `omarchy-hw-hybrid-gpu`; the two have no files in common.

### DIFF
--- a/bin/omarchy-hw-nvidia-drives-display
+++ b/bin/omarchy-hw-nvidia-drives-display
@@ -1,0 +1,41 @@
+#!/bin/bash
+
+# omarchy:summary=Detect whether an NVIDIA GPU is driving the screen the session renders on, rather than merely being present.
+
+# Read sysfs rather than shelling out to lspci or nvidia-smi, both of which
+# resume a runtime-suspended GPU. A connector's status file is served from the
+# driver's cached state, so asking does not power the card up.
+drm_class_path="${OMARCHY_DRM_CLASS_PATH:-/sys/class/drm}"
+
+shopt -s nullglob
+
+nvidia_connected=0
+foreign_panel=0
+
+for card in "$drm_class_path"/card[0-9]*; do
+  [[ ${card##*/} == *-* ]] && continue
+  [[ -r $card/device/vendor ]] || continue
+
+  vendor=$(<"$card/device/vendor")
+
+  for connector in "$card"-*; do
+    [[ -r $connector/status ]] || continue
+    [[ $(<"$connector/status") == "connected" ]] || continue
+
+    if [[ $vendor == "0x10de" ]]; then
+      nvidia_connected=1
+      continue
+    fi
+
+    # A built-in panel on another card means the compositor is rendering there,
+    # and these variables are session-wide. On a laptop whose external ports are
+    # wired to the discrete card, docking lights up an NVIDIA connector without
+    # moving the rendering off the integrated GPU.
+    name=${connector##*/}
+    case ${name#card*-} in
+      eDP* | LVDS* | DSI*) foreign_panel=1 ;;
+    esac
+  done
+done
+
+((nvidia_connected && !foreign_panel))
--- a/default/hypr/nvidia.lua
+++ b/default/hypr/nvidia.lua
@@ -3,11 +3,18 @@ local paths = require("default.hypr.paths")
 local nvidia = paths.omarchy_path .. "/bin/omarchy-hw-nvidia"
 local nvidia_gsp = paths.omarchy_path .. "/bin/omarchy-hw-nvidia-gsp"
 local nvidia_without_gsp = paths.omarchy_path .. "/bin/omarchy-hw-nvidia-without-gsp"
+local nvidia_drives_display = paths.omarchy_path .. "/bin/omarchy-hw-nvidia-drives-display"
 
 -- These detectors read cached sysfs IDs rather than shelling out to lspci.
 -- lspci reads PCI config space, which resumes a runtime-suspended GPU, and on a
 -- hybrid laptop that wake alone outlasts Hyprland's 1.5s config reload budget.
-if o.shell_succeeds(o.shell_quote(nvidia)) then
+--
+-- Naming NVIDIA here points every GL and VA-API client in the session at that
+-- card, so it has to be the card actually driving the screen. On a hybrid laptop
+-- the panel hangs off the integrated GPU while the discrete card sits in runtime
+-- suspend, and setting these there means the first GL application to start wakes
+-- it and holds it awake. Having an NVIDIA GPU is not the same as rendering on it.
+if o.shell_succeeds(o.shell_quote(nvidia)) and o.shell_succeeds(o.shell_quote(nvidia_drives_display)) then
   if o.shell_succeeds(o.shell_quote(nvidia_gsp)) then
     hl.env("NVD_BACKEND", "direct")
     hl.env("LIBVA_DRIVER_NAME", "nvidia")
--- a/test/shell.d/hw-nvidia-test.sh
+++ b/test/shell.d/hw-nvidia-test.sh
@@ -97,3 +97,96 @@ assert_detects "a non-display NVIDIA function is not a GPU" no no no
 
 write_pci_devices
 assert_detects "a machine with no PCI devices detects nothing" no no no
+
+# Whether an NVIDIA card is present and whether it is driving a screen are
+# different questions. On a hybrid laptop the panel hangs off the integrated GPU
+# and the discrete card sits in runtime suspend, and naming NVIDIA in the session
+# environment there wakes it for every GL client.
+#
+# Connector status is read from the driver's cached state, so the detector can
+# ask without powering the card up. That is the whole point of it.
+write_drm_cards() {
+  rm -rf "$tmp_dir/drm"
+  mkdir -p "$tmp_dir/drm"
+
+  local spec card vendor connector status
+  for spec in "$@"; do
+    IFS='|' read -r card vendor connector status <<<"$spec"
+    mkdir -p "$tmp_dir/drm/$card/device"
+    printf '%s\n' "$vendor" >"$tmp_dir/drm/$card/device/vendor"
+    if [[ -n $connector ]]; then
+      mkdir -p "$tmp_dir/drm/$card-$connector"
+      printf '%s\n' "$status" >"$tmp_dir/drm/$card-$connector/status"
+    fi
+  done
+}
+
+drives_display() {
+  OMARCHY_DRM_CLASS_PATH="$tmp_dir/drm" "$ROOT/bin/omarchy-hw-nvidia-drives-display"
+}
+
+# A hybrid laptop: the panel is on the Intel card, the NVIDIA card has an unused
+# output. This is the case that was setting the variables.
+write_drm_cards 'card0|0x10de|HDMI-A-1|disconnected' 'card1|0x8086|eDP-1|connected'
+if drives_display; then
+  fail "an NVIDIA card with no connected output is not driving the display"
+fi
+pass "an idle discrete NVIDIA card does not count as driving the display"
+
+# Docking that laptop lights up an NVIDIA connector, but the panel is still on
+# the Intel card and the compositor still renders there. hl.env is session-wide,
+# so naming NVIDIA would charge the internal display for the external one.
+write_drm_cards 'card0|0x10de|HDMI-A-1|connected' 'card1|0x8086|eDP-1|connected'
+if drives_display; then
+  fail "a built-in panel on another card keeps the session off NVIDIA"
+fi
+pass "docking a hybrid laptop does not move the session onto NVIDIA"
+
+# A laptop whose panel is wired to the NVIDIA card has nowhere else to render.
+write_drm_cards 'card0|0x10de|eDP-1|connected' 'card1|0x8086||'
+drives_display || fail "a panel on the NVIDIA card is driven by NVIDIA"
+pass "a panel on the NVIDIA card still names NVIDIA"
+
+# A desktop: an integrated card is present but drives nothing, and the monitor
+# hangs off NVIDIA. No built-in panel is involved, so nothing changes here.
+write_drm_cards 'card0|0x10de|DP-1|connected' 'card1|0x8086|DP-1|disconnected'
+drives_display || fail "an idle integrated card does not keep the session off NVIDIA"
+pass "a desktop with an unused integrated card still names NVIDIA"
+
+write_drm_cards 'card0|0x10de|DP-1|connected' 'card1|0x8086|DP-2|connected'
+drives_display || fail "a monitor on the integrated card is not a built-in panel"
+pass "a desktop driving monitors from both cards still names NVIDIA"
+
+write_drm_cards 'card0|0x10de|DP-1|connected'
+drives_display || fail "a single NVIDIA card with a connected output is driving the display"
+pass "an NVIDIA-only machine still names NVIDIA"
+
+# No NVIDIA at all, and an NVIDIA card carrying no connectors whatsoever.
+write_drm_cards 'card0|0x8086|eDP-1|connected'
+if drives_display; then
+  fail "a machine without an NVIDIA card is not driving a display with one"
+fi
+pass "a machine with no NVIDIA card answers no"
+
+write_drm_cards 'card0|0x10de||'
+if drives_display; then
+  fail "an NVIDIA card with no connectors is not driving a display"
+fi
+pass "an NVIDIA card with no connectors answers no"
+
+# Telling a card from a connector is a name comparison, so a dash further up the
+# path must not hide every card from it.
+mkdir -p "$tmp_dir/drm-alt/card0/device" "$tmp_dir/drm-alt/card0-DP-1"
+printf '0x10de\n' >"$tmp_dir/drm-alt/card0/device/vendor"
+printf 'connected\n' >"$tmp_dir/drm-alt/card0-DP-1/status"
+OMARCHY_DRM_CLASS_PATH="$tmp_dir/drm-alt" "$ROOT/bin/omarchy-hw-nvidia-drives-display" ||
+  fail "a dash in the sysfs path does not hide the card"
+pass "the detector reads cards from a path containing a dash"
+
+# The gate itself: nvidia.lua must require both answers before it names NVIDIA
+# for the whole session.
+grep -q 'omarchy-hw-nvidia-drives-display' "$ROOT/default/hypr/nvidia.lua" ||
+  fail "nvidia.lua asks whether NVIDIA drives the display"
+grep -qE 'shell_succeeds\(o\.shell_quote\(nvidia\)\) and o\.shell_succeeds\(o\.shell_quote\(nvidia_drives_display\)\)' "$ROOT/default/hypr/nvidia.lua" ||
+  fail "nvidia.lua requires both a present NVIDIA card and one driving the display"
+pass "nvidia.lua only names NVIDIA when NVIDIA drives the display"


kevinbsr:
Confirming this from the AMD+NVIDIA side (Dell G15 5515, Ryzen 7 5800H + RTX 3060 Mobile, iGPU-primary), and raising one hardware shape that I think the "drives a display" question answers differently than intended.

**The patch is the right shape.** Reading connector status out of the driver's cached state is the detail that matters — every other way of asking resumes the card, which is the trap the existing comment in `nvidia.lua` already warns about for `lspci`. Gating on it is strictly better than today on this machine.

**The measurement, for whoever needs the cost quantified.** I reported this in #8989: removing the two variables took the system from **23.7 W to 10.7 W**. One correction worth carrying over, because I got it wrong first: the env does **not** stop the dGPU from reaching D3cold when idle — I restored it and the card still slept 100% of the time. What it costs is waking the card for anything that decodes video, and then keeping it awake. It is "wakes on use", not "never sleeps", so the delta shows up under a video workload rather than at an idle desktop.

**The case I want to flag: laptops where every external output is muxed to the dGPU.** On this machine the internal panel is the only output the iGPU owns:

```
card0  vendor=0x10de  driver=nvidia
     card0-DP-1        disconnected
     card0-HDMI-A-1    disconnected
card1  vendor=0x1002  driver=amdgpu
     card1-eDP-1       connected
```

Undocked, the detector correctly answers no. **Docked, it answers yes** — DP-1 is an NVIDIA connector — and the session names NVIDIA again, including for the clients on the still-AMD internal panel. That is arguably correct for windows on the external screen, but `hl.env` is session-wide, so the laptop display pays for it too, and the ~10 W comes back whenever a monitor is plugged in.

This layout is common on gaming laptops: the built-in panel is wired to the iGPU and the external ports are hardwired to the discrete card, with the compositor still compositing on the iGPU and doing PRIME output offload.

So the question this answers is "is NVIDIA driving *an* output", where the thing the env should follow is closer to "is NVIDIA driving the output the compositor renders on". I don't think that blocks this PR — undocked hybrid laptops are the common case and this fixes them. But it may be worth either narrowing the detector to the primary/compositing GPU, or saying in the comment that docking deliberately opts the session back in, so the next person measuring 10 W while docked finds the answer instead of filing a fourth issue.

Happy to test a revision on this hardware, docked and undocked, if that would help.

---
Reported with the help of Claude Code.

VykosMolt:
You're right, and I'd put it more strongly than "arguably correct": `hl.env` is session-wide, so on that layout the internal panel pays for a card it never renders on. Thanks for the correction on the measurement too — "wakes on use", not "never sleeps", is the accurate framing and I've taken it.

Force-pushed a narrower detector. It now answers no if a built-in panel — `eDP`, `LVDS` or `DSI` — on a non-NVIDIA card is connected, whatever the NVIDIA connectors say. A panel on another card means the compositor is rendering there, and docking doesn't move it.

By hardware shape:

- your G15 undocked — no, unchanged
- your G15 docked — was yes, now no
- panel wired to the NVIDIA card — yes, unchanged
- desktop, monitor on NVIDIA, integrated card idle — yes, unchanged
- desktop driving monitors from both cards — yes, unchanged

Desktops are untouched, because the rule keys on a built-in panel and they don't have one. The last case is the one I left alone deliberately: there's no cheap sysfs answer to which card the compositor picked, and answering no there would be a regression for people who plugged their monitor into the NVIDIA card on purpose.

The docked case is now asserted in `hw-nvidia-test.sh` — it was asserted the other way round, which is the bug you found — alongside panel-on-NVIDIA, desktop-with-idle-integrated-card and monitors-on-both. 21 assertions, all passing; the previous detector fails the docked one.

The offer to test docked and undocked would be very welcome, if it still stands.

Hazakins:
The narrowed built-in-panel detector gives the right answer on an AMD+NVIDIA machine, and I want to add one thing the gate does not cover.

**Hardware:** Framework Laptop 16, Ryzen AI 9 HX 370 / Radeon 890M (`amdgpu`) driving `eDP-2` plus a Samsung G95C over a Thunderbolt dock; RTX 5070 graphics module present, its own connectors disconnected. Omarchy 4.0.3, Hyprland 0.56.2-2, kernel 7.2.3-arch1-3, mesa 26.2.2, nvidia 610.57.04. With stock `nvidia.lua`, `LIBVA_DRIVER_NAME=nvidia` + `__GLX_VENDOR_LIBRARY_NAME=nvidia` were exported session-wide; Chrome decoded video on the dGPU and handed frames to the amdgpu compositor (flashing playback, and a GPU-process SIGTRAP in Meet on 2026-09-09). Pinning the session to radeonsi/mesa fixed both. I have not built this PR — I run an equivalent local override — so this is a datapoint on the rule, not a patch test.

**The eDP-on-another-card rule asks the right question.** On this box both versions of the gate answer "no" (the NVIDIA connectors are disconnected), so it does not separate them here; but @TijmenvandenBan's case on #8215 (NVIDIA driving `DP-2` while Chromium still renders on `renderD129`) shows the display owner and the renderer can disagree, and keying on the built-in panel's card sidesteps that.

**What the env vars do not cover:** on this box the dGPU never reaches D3cold while any GTK4 app or Chrome is open, because the **Vulkan loader enumerates every ICD**. Measured with `gtk4-layer-demo` alone: 7 fds on `/dev/nvidia0`, 4 on `nvidiactl`, 2 on the NVIDIA render node, with `LIBVA_DRIVER_NAME` already set to `radeonsi`. Pinning `VK_DRIVER_FILES` to the RADV ICD takes it to zero with no renderer change. Worth considering alongside the VA-API/GLX selection.

---
*Disclosure: this diagnosis and write-up were produced by Claude (Fable 5.1, via Claude Code) working on my laptop; the measurements are from my machine and I reviewed the text before posting.*

kevinbsr:
Tested on the G15, both ways. The narrowed detector does what you said it would.

**Hardware, docked (real, not simulated):** two external displays plugged in, both on NVIDIA connectors, internal panel on the AMD card.

```
card0-DP-1        connected     (NVIDIA)
card0-HDMI-A-1    connected     (NVIDIA)
card1-eDP-1       connected     (amdgpu)
```

| | first version | current |
|---|---|---|
| **G15 docked** (real hardware) | **yes** | **no** |
| G15 undocked | no | no |

Undocked is the same machine with the two external connectors reading `disconnected` — that was its real state earlier today, and I re-ran it through `OMARCHY_DRM_CLASS_PATH` to get both detectors against the identical tree. Docked is live hardware with three displays actually lit.

Session env with the current detector's answer, docked: `LIBVA_DRIVER_NAME=radeonsi`, `__GLX_VENDOR_LIBRARY_NAME=mesa`, and `glxinfo` reports the compositor on `radeonsi/renoir`. That is the outcome the patch is for.

**One thing the docked case changes about my earlier power claim, which I should correct.** Docked, the dGPU sits at `runtime_status=active` / `power_state=D0` and stays there — it has to, it is driving both external outputs. So the 23.7 W → 10.7 W I quantified is an **undocked** number. Docked, the card is awake no matter what these two variables say, and the patch is not buying that back.

That does not weaken the change; it just narrows what it buys. Docked, the win is correctness rather than watts: clients render on the card the compositor actually composites on, instead of being pushed to a card whose frames then have to cross back. Undocked, it is both.

For completeness, the control case still answers correctly here: a fake tree with the panel on the NVIDIA card and the Intel connectors disconnected returns yes, so the rule does not regress laptops wired the other way.

---
Reported with the help of Claude Code.

kevinbsr:
@Hazakins I reproduced your `gtk4-layer-demo` measurement exactly, and then got the opposite answer on the part that matters. Posting both because I think the two halves of that finding travel differently.

**The fd count reproduces to the digit.** Dell G15 5515, Ryzen 7 5800H + Radeon Vega (Cezanne, `amdgpu`) driving the panel, RTX 3060 Mobile (GA106M) present. nvidia 610.57.04, mesa 26.2.2, Hyprland 0.56.2, kernel 7.2.4-arch1-2 — same driver and mesa as yours.

| | `/dev/nvidia0` | `/dev/nvidiactl` | NVIDIA render node | AMD render node |
|---|---|---|---|---|
| as shipped | 7 | 4 | 2 | 1 |
| `VK_DRIVER_FILES` → RADV | 0 | 0 | 0 | 1 |

7 / 4 / 2, exactly your numbers, with `LIBVA_DRIVER_NAME=radeonsi` already exported. Pinning the ICD zeroes them and does not move the renderer — the AMD node stays at 1 either way. So "the Vulkan loader opens the NVIDIA device even when nothing asks it to" is confirmed on a second machine.

**Where it diverges: here those fds do not cost D3cold.** Same session, undocked, with Hyprland, Brave, Electron and Spotify all holding `/dev/nvidiactl`, the card read `runtime_status=suspended` / `power_state=D3cold` with 397 minutes of accumulated `runtime_suspended_time`. An earlier pass on this machine (2026-08-31) reached the same conclusion from the other direction: with all four env vars restored to `nvidia`, the dGPU still slept 100% of the time.

So on GA106M an open fd is not a wakelock. On your RTX 5070 it evidently is. Same driver branch, different silicon — worth flagging so nobody generalises the sleep claim from the fd claim, because the fd claim does generalise and the sleep one apparently does not.

**And the fix has a cost worth naming.** Pinning `VK_DRIVER_FILES` to RADV makes the discrete card invisible to every Vulkan client, not just the ones you wanted to keep off it. On this machine that meant Steam titles silently rendering on the iGPU with no error and no indication — which is a worse failure than the thing it fixes, because it is quiet. Anyone applying this on a laptop that also games wants `nvidia-offload` (or an equivalent that re-exports the vars) on those launchers, or they will not notice until they wonder why the framerate halved.

None of this argues against your datapoint — on hardware where the card genuinely will not sleep, the trade flips. I just would not want the loader-enumerates observation and the never-sleeps conclusion shipped as one finding.

---
Measured with the help of Claude Code.

Hazakins:
@kevinbsr thanks for re-running it — and you're right that the two halves need to travel separately. I re-measured this morning so the sleep claim rests on a number rather than on my earlier wording.

**On the RTX 5070 Laptop the fds do hold the card awake.** `env -u VK_DRIVER_FILES gtk4-layer-demo`, reading only `power/runtime_status`: `suspended` → `active` within 2 s of launch, `active` for the whole two-minute run with `runtime_suspended_time` frozen, `suspended` again about 12 s after the demo exited. The demo was the only holder (`/dev/nvidia0` ×7, `nvidiactl` ×4, NVIDIA render node ×2). Same driver branch as yours (610.57.04, mesa 26.2.2, kernel 7.2.3).

One question that might reconcile the two machines: in your D3cold reading the holders had `/dev/nvidiactl` open. Did the 3060 also stay in D3cold **while `gtk4-layer-demo` itself was running** with its seven `/dev/nvidia0` handles? `nvidiactl` is the control node; `nvidia0` is the per-GPU one. If your card sleeps even then, it's silicon/firmware; if it wakes, it's about which node gets opened, and the "fd claim" and the "sleep claim" line up again.

**On the cost — agreed, and under this pin it's worse than a missing `nvidia-offload`.** Under the pin, the `prime-run` variables alone don't bring the card back: with `__NV_PRIME_RENDER_OFFLOAD=1 __VK_LAYER_NV_optimus=NVIDIA_only` set, `vulkaninfo --summary` still lists only the Radeon 890M, because `VK_DRIVER_FILES` keeps the loader from ever reading the NVIDIA ICD. The per-app override has to re-point `VK_DRIVER_FILES` at `nvidia_icd.json` as well. With that in a wrapper used as the Steam launch option (`<wrapper> %command%`), the setting survives the Steam runtime: pressure-vessel imports it as `/usr/lib/pressure-vessel/overrides/share/vulkan/icd.d/nvidia_icd.json` inside the container. (Honest limit: both titles I have installed are native OpenGL builds, so what I verified end to end is the GL side — `libGLX_nvidia` mapped, the game holding the 5070 — plus that env rewrite; I haven't yet run a Vulkan/Proton title through it.) We also list installed Steam games that lack the wrapper at login, so a forgotten one doesn't stay quiet.

So for anyone reading along: the pin is only worth it on hardware where the enumeration actually keeps the card awake, and only together with a per-app override that resets `VK_DRIVER_FILES`, not `prime-run` alone.

---
*Disclosure: this diagnosis and write-up were produced by Claude (Opus 5, via Claude Code) working on my laptop; the measurements are from my machine and I reviewed the text before posting.*

Hazakins:
Follow-up on the limit I mentioned: I've now run a Proton title through it. Among Us (Windows-only, Proton Experimental, SteamLinuxRuntime_4) with `<wrapper> %command%` as the launch option rendered on the RTX 5070 — DXVK's `d3d11.dll`/`dxgi.dll` → `winevulkan` → the NVIDIA ICD, `VK_DRIVER_FILES` inside the container rewritten to `/usr/lib/pressure-vessel/overrides/share/vulkan/icd.d/nvidia_icd.json`, and the game holding the card at ~56 % load. So the per-app override survives Proton as well as the native OpenGL titles.

One trap for anyone copying this: Steam writes a newly edited launch option to `localconfig.vdf` lazily (about two minutes here), so a script that reads the file right after you set it will report the game as not opted in. `~/.local/share/Steam/logs/console_log.txt` ("Game process added : AppID …") shows the command Steam actually launched.

---
*Disclosure: this diagnosis and write-up were produced by Claude (Opus 5, via Claude Code) working on my laptop; the measurements are from my machine and I reviewed the text before posting.*

VykosMolt:
@kevinbsr thanks for running both detectors on the G15 — docked on live hardware is exactly the case the tests can only simulate. The correction lands too: docked, the card is awake for its own outputs anyway, so there the change buys correctness rather than watts.

@Hazakins the loader opening every ICD is a real finding, but it's a separate mechanism from the one this PR gates, and your 5070 and kevinbsr's 3060 disagree on whether it costs D3cold. Pinning `VK_DRIVER_FILES` also hides the dGPU from every Vulkan client and fails quietly for games unless each launcher resets it. That trade depends on the hardware, so I'd rather keep it out of this PR and let it be its own issue.

kevinbsr:
@Hazakins ran your exact test on the 3060, with a clean sleeping baseline. **The card sleeps right through it.** So on this hardware it is silicon/firmware, not about which node gets opened — the two halves stay separate.

`env -u VK_DRIVER_FILES gtk4-layer-demo`, undocked (both NVIDIA connectors `disconnected`), reading `power/runtime_status` and `runtime_suspended_time`:

```
baseline, no demo :  30006 ms of sleep in 30 s   = 100 % asleep
demo running      :  72151 ms of sleep in 90 s   =  80 % asleep
  +15s  active
  +30s  active
  +45s  suspended     <- back in D3cold with the demo still up
  +60s  suspended
  +75s  suspended
  +90s  suspended
```

Descriptors held by the demo for that whole window: `/dev/nvidia0` ×7, `/dev/nvidiactl` ×4, NVIDIA render node ×2 — identical to your numbers. The card wakes for the enumeration, then returns to D3cold about 40 s later and stays there while every one of those handles is still open.

So your `nvidiactl` vs `nvidia0` hypothesis does not explain the difference here: the per-GPU node is open seven times and the GA106M sleeps anyway. Same driver branch (610.57.04), same mesa (26.2.2). The divergence looks like the silicon or its firmware, not the node.

**Why my earlier reply was not a clean measurement, for the record.** My first D3cold reading came from ordinary apps holding `nvidiactl`, not from the demo holding `nvidia0` — you were right that it was not comparable. My first two attempts at redoing it were also invalid: once docked, where the card is awake for its own outputs regardless, and once undocked but without a sleeping baseline. Which turned up something worth knowing on its own:

**Undocking does not restore runtime suspend on this machine.** After the external displays came out, both NVIDIA connectors read `disconnected`, yet `runtime_suspended_time` stayed frozen for over four minutes — the card had not slept for a single millisecond since the monitors were first plugged in. Only a logout/login brought it back. If anyone else measures a hybrid laptop that has been docked during the session, that alone will make the card look permanently awake.

One more confound for anyone reproducing this: `btop` with `show_gpu_info = "Auto"` polls the card through NVML and holds it up for as long as it is open.

---
Measured with the help of Claude Code.

remydw:
Verified on separate hardware — this detector gives the right answer where a `boot_vga`-based one does not.

**Machine:** ROG Zephyrus G14 GA403WW, Ryzen AI 9 HX 370 (Radeon 890M) + RTX 5080 Max-Q, Omarchy 4.0.3-1, Hyprland 0.56.2, mesa 26.2.2, Chrome 153.

```
NVIDIA  0000:64:00.0  class 0x030000  boot_vga=0
        DP-1, HDMI-A-1, eDP-1 -> all disconnected
AMD     0000:65:00.0  class 0x038000  boot_vga ABSENT
        eDP-2 + DP-3 connected (compositor renders on this card)
```

- `omarchy-hw-nvidia-drives-display` → **exit 1** here (no NVIDIA connector is connected, and the built-in panel belongs to another card). Correct.
- I then ran the session state this PR produces — all three variables unset — through headed Chrome with `--vmodule=vaapi*=2,video_decoder*=1` on a YouTube page: `VaapiVideoDecoder()` initialises three times, zero `vaInitialize` failures, **zero** `eglCreateImage failed` lines (the current default produces 2482 of them in ~15 s), the GPU process maps only `libgallium`, the video region shows real frames (0.0 % black, 63.9 % inter-frame change) and the dGPU stays `suspended`. CPU is 29.3 % of the Chrome tree versus 50.5 % for the software-decode fallback.
- Gating **all three** variables rather than only `LIBVA_DRIVER_NAME` is the part I would keep: leaving `__GLX_VENDOR_LIBRARY_NAME=nvidia` in place still points XWayland GLX clients at the discrete card.

One implementation note: not calling `hl.env` at all is the right shape for this fix, because Hyprland cannot unset a variable — `hl.env` requires a string, and an empty string is not equivalent to unset. `LIBVA_DRIVER_NAME=""` makes libva attempt to load a driver named `""` (`vaInitialize failed: unknown libva error`), after which Chromium silently software-decodes while `chrome://gpu` still reports "Video Decode: Hardware accelerated".

Independent context, measurements and the review of #10589 are in #8726.

VykosMolt:
@remydw thanks. The G14 is the third machine in this thread and the first where the NVIDIA card has its own `eDP-1` present but disconnected, which is the case the built-in-panel rule had to get right without leaning on `boot_vga`. Your Chrome trace is the user-visible half of what kevinbsr measured in watts: zero `eglCreateImage` failures against 2482 with the current default, and the dGPU staying `suspended` throughout.

Agreed on all three variables, and on unset rather than blank. `hl.env` cannot express "not set", and `LIBVA_DRIVER_NAME=""` sends libva looking for a driver named `""`, so the detector decides before `hl.env` is called and the NVIDIA branch is skipped entirely when it answers no.

Tally for whoever queues this, all iGPU-panel AMD+NVIDIA laptops, all answering no with the current detector: Dell G15 5515 docked and undocked (kevinbsr, live hardware both ways), Framework 16 with an RTX 5070 module (Hazakins, equivalent local override), ROG G14 GA403WW (remydw). The NVIDIA-drives-the-panel case is the one the tests cover.

srgonzale:
Tested on Intel+NVIDIA hybrid: TigerLake-H + RTX 3050 Mobile

**Hardware:**
- iGPU: Intel TigerLake-H GT1 UHD Graphics (`card2`, vendor `0x8086`) — `eDP-1` connected
- dGPU: NVIDIA GA107M GeForce RTX 3050 Mobile (`card1`, vendor `0x10de`) — no DRM connectors at all

**Detector results:**
- `omarchy-hw-nvidia` → exit 0 (NVIDIA GPU present)
- `omarchy-hw-nvidia-drives-display` → exit 1 (no NVIDIA connector connected, Intel panel is the built-in display)

**Before the patch:**
- `LIBVA_DRIVER_NAME=nvidia`, `NVD_BACKEND=direct`, `__GLX_VENDOR_LIBRARY_NAME=nvidia` set session-wide
- Chromium video playback issues (flashing, corruption, or decode failures)

**After the patch:**
- All three NVIDIA env vars unset — `env | grep -iE "libva|nvidia|glx|nvd"` returns nothing
- Chromium video playback works correctly, hardware decode on Intel/Mesa as expected
- Fix persists across logout/login

The detector correctly identifies this as a hybrid laptop where the iGPU drives the panel, and skips the NVIDIA session env. Video works as intended.

PR #9485, opened 2026-08-31
Honor keepLoaded for services during plugin hot-reload

barmstrong:
`omarchy.lock` already declares `keepLoaded: true`, but that flag was only consulted for summoned panels. `reloadPlugins()` still called `unloadPluginServices()`, which `destroy()`ed every service — including the lock client.

If an installed plugin file changes while the session is locked (a bar widget hot-reload is enough), the shell tears down `WlSessionLock`. Hyprland keeps the session locked and shows the crashed-lockscreen fallback ("Oopsie daisy, it looks like you locked your screen but the lockscreen app died").

This keeps `keepLoaded` services mounted across plugin hot-reload. Disabled or removed services are still dropped by `_syncServices()`.

## Test plan

- `bash test/shell.d/plugins-test.sh` — lock/idle/polkit/notifications stay `keepLoaded`, and `unloadPluginServices` consults that flag
- `bash test/shell.d/runtime-smoke-test.sh` — after `shell rescanPlugins`, lock `lastEventAt` is unchanged and is not `lock-stranded`

`./test/all` was run; CLI passed. The four shell files that failed here are preexisting environment gaps (`omarchy-pkgs` checkout missing, about-window animation), not this change.

### DIFF
--- a/docs/omarchy-shell.md
+++ b/docs/omarchy-shell.md
@@ -37,9 +37,7 @@ wait).
 
 Only one full bar option is active at a time. The built-in `omarchy.bar` is
 used when `bar.id` is omitted or when a selected third-party bar cannot load.
-Panels, overlays, and menus are loaded when summoned. Plugins can set the
-top-level manifest key `keepLoaded: true` to survive between summons.
-First-party services are loaded at startup.
+Panels, overlays, and menus are loaded when summoned. Plugins can set the top-level manifest key `keepLoaded: true` to survive between summons, and to keep a service mounted across plugin hot-reload (so `omarchy.lock` is not destroyed while Hyprland still holds the session lock). First-party services are loaded at startup.
 
 Entry points are QML `Item`s. Panel, overlay, and menu entry points expose
 `open(payloadJson)` and `close()` for summon/hide; on load the host injects
--- a/shell/README.md
+++ b/shell/README.md
@@ -86,8 +86,10 @@ Only one `bar` plugin is active at a time. Missing or invalid selections fall
 back to the built-in `omarchy.bar`, so users always have a safe path home.
 Panels, overlays, and menus are loaded when summoned. Plugins that need
 to outlive a single summon can set `keepLoaded: true` (e.g. the image
-picker keeps its overlay window mounted between summons). First-party
-services are loaded at startup.
+picker keeps its overlay window mounted between summons). The same flag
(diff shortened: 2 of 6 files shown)


barmstrong:
This is caused by having an agent work on a plugin, and then your lockscreen comes on, and the agent finishes making some changes to the plugin files, it apparently crashes/kills the lockscreen app and you get this nasty "blue screen of death" and almost feels like you've bricked your laptop.
image

I haven't extensively tested this fix, but this is what the agents came up with - so hoping it solves it! Surprised no one else has encountered this, since it happened a few times.

dhh:
Reviewed, and the fix is right — keeping `keepLoaded` services mounted is exactly what the lock client needs. I pushed three commits to tighten it up:

- The smoke-test assertion was vacuous: a freshly recreated lock service also reports an empty `lastEventAt`, so the before/after comparison passed even with the shell.qml change reverted (verified by running the suite both ways). It now uses a fixture `keepLoaded` service whose in-memory marker can only survive the rescan if the instance does.
- `_syncServices()` only checked installed + enabled when dropping, so a kept service whose plugin stopped declaring a service kind kept running as a zombie until shell restart. It now applies the same eligibility checks used at creation, and kept instances get the refreshed manifest after a rescan.
- Covered `omarchy.media` in the keepLoaded expectations and noted in the README that code changes to a kept service only land on shell restart.

Full smoke, plugins, and CLI suites pass, and both new assertions fail when their respective fix is reverted.

Rohansguliani:
Independent reproduction on Omarchy 4.0.2, Quickshell 0.3.1, and Hyprland 0.56.2:

1. The session was securely locked.
2. `omarchy-shell shell rescanPlugins` ran while locked.
3. The journal immediately recorded `lock-stranded: recovering`, followed by `lock-requested` and `lock-pending: screen-stabilizing`, but never returned to `secure=true`.
4. Hyprland displayed its crashed-lockscreen failsafe. There was no coincident OOM, GPU reset, segfault, or coredump.

I checked out the maintainer-updated head at `bd1464c7` on the affected machine. `plugins-test.sh` passes, and `runtime-smoke-test.sh` confirms the `keepLoaded` fixture service retains its in-memory marker across `rescanPlugins` and is still removed when it stops declaring a service kind.

This matches the failure exactly and the updated implementation covers it without special-casing `omarchy.lock`.

PR #9498, opened 2026-08-31
Add a Grok usage collector to the agents panel

Thunderhawke:
Adds a `grok` collector to the agents panel, so the Grok CLI gets a tab
alongside Claude, Codex and the rest instead of silently missing from the bar
for anyone whose subscription is Grok.

## What it reads

Grok ships no usage or rate-limit endpoint — the CLI has no `usage`
subcommand and never exposes the account allowance — so every number comes
from the CLI's own session logs at
`~/.grok/sessions/<encoded-cwd>/<session-id>/updates.jsonl`, where one
`turn_completed` update is written per prompt carrying that turn's token
usage broken down by model.

That makes the record local-stats-only: `hasLocalStats` and `hasPromptStats`
are true, `limits` stays empty, and `tierLabel`/`usageStatusText` stay empty
so `heroMeta()` renders the neutral "Subscription" line rather than a warning
card. This is the same shape Claude's collector already falls back to when it
can't reach Anthropic's endpoint, so the panel needs no changes to draw it.

## Details worth a reviewer's eye

**Token buckets are de-overlapped.** Grok's counters nest:
`cachedReadTokens` and `cacheCreationTokens` are both counted *inside*
`inputTokens` (records satisfy `totalTokens == inputTokens + outputTokens`).
The panel wants four disjoint buckets, so plain input is what remains after
subtracting the two cache buckets. Reasoning tokens stay folded into output,
matching how the Codex collector treats them. Verified against hand-summed
raw turns.

**Forked sessions are deduped.** `grok --fork-session` replays prior turns
into the new session's log, so turns are deduped on `prompt_id` and a fork
doesn't inflate prompt counts or token totals.

**One prompt per turn.** A `turn_completed` served by more than one model
counts once as a prompt, with its tokens split across models in
`todayTokensByModel` / `modelUsage`.

**Scan cache**, mirroring the Codex collector: flock-guarded, atomic write,
`{schemaVersion, scanDate, stats}` envelope so today's counters don't survive
midnight, and it degrades to a direct scan on any cache failure. Without it
the widget rereads every session file on each refresh. `--force` bypasses it.
The scan window is 30 days by mtime, so "all time" carries the same caveat
the README already documents for the other scanners.

`GROK_HOME` is honored, the way the Codex collector honors `CODEX_HOME`.

## Not included: the icon

I deliberately did not add `shell/plugins/agents/assets/grok.svg` — I wasn't
going to approximate xAI's mark by hand and ship it. Grok falls back to the
default bar glyph, which the plugin already supports. Happy to add the real
asset if a maintainer wants to drop one in, or to take a patch on the branch.

## Tests

`test/shell.d/agent-usage-grok-scanner-test.sh` — 10 assertions, all passing:
valid record with no sessions present, the bucket split, a second model in
its own bucket, today's prompts/sessions/tokens, the seven-day chart,
totals/active days, forked-session dedup, a multi-model turn counting as one
prompt, the scan-window skip, and surviving malformed lines. Fixtures are
built with `jq -nc` under `TZ=UTC` with `GROK_HOME`/`XDG_CACHE_HOME`
redirected, so nothing touches the runner's real state.

`./test/all` shows no new failures.

### DIFF
--- a/bin/omarchy-agent-usage-grok
+++ b/bin/omarchy-agent-usage-grok
@@ -1,0 +1,384 @@
+#!/usr/bin/python3
+# omarchy:summary=Print the Grok usage record as JSON
+# omarchy:args=[--force] [--limits-only]
+# omarchy:hidden=true
+"""Collect Grok CLI usage into one display-ready JSON record.
+
+Grok ships no usage or rate-limit endpoint: the CLI has no `usage`
+subcommand and never exposes the account's allowance, so this record carries
+local stats only -- the shape Claude's collector falls back to when it cannot
+reach Anthropic's endpoint. The panel draws tokens by day, tokens by model,
+prompts, and sessions; the limits section stays empty.
+
+Every number comes from ~/.grok/sessions/<encoded-cwd>/<session-id>/updates.jsonl,
+where the CLI writes one `turn_completed` update per prompt carrying that
+turn's token usage broken down by model.
+"""
+
+import argparse
+import fcntl
+import hashlib
+import json
+import os
+import sys
+import tempfile
+import time
+from datetime import datetime, timedelta, timezone
+from pathlib import Path
+
+AGENT_ID = "grok"
+AGENT_NAME = "Grok"
+AUTH_HELP = "Run `grok login` to sign in."
+
+# A scan this recent is only reused to dedup concurrent collector runs (the
+# update command backgrounds one per agent while the panel refreshes on its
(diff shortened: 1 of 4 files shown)


bjarneo:
Automated duplication check: this pull request looks similar to #6485, which covers the same Grok usage collector with broader scope (brand icons, weekly credit meter via ACP billing, live verifications). I keep that one open and close this one to consolidate review.

PR #9504, opened 2026-08-31
Clear setgid when securing Windows VM mounts

jmwall:
Fix protected-mount setup for existing Windows VM shared directories carrying the setgid bit.

GNU `chmod 0700` preserves setgid on directories. When the shared directory is mode `2700`, the launcher therefore reads it back as `2700`, rejects it because it is not exactly `700`, and aborts before starting the VM or FreeRDP.

Use `chmod =0700` during both user preflight and privileged mount preparation so inherited special bits are explicitly cleared.

Add regression coverage for setgid shared directories in the compose and privileged mount-boundary suites.

Tests:

- `bash test/shell.d/windows-vm-compose-test.sh`
- `bash test/shell.d/windows-vm-mount-boundary-test.sh`
- `OMARCHY_PKGS_PATH=/tmp/omarchy-pkgs-codex ./test/all`

### DIFF
--- a/bin/omarchy-windows-vm
+++ b/bin/omarchy-windows-vm
@@ -580,7 +580,9 @@ prepare_caller_mounts() {
   # Privacy is an explicit preflight step for both already-pinned sources, not
   # a side effect halfway through the two-mount transaction. Old umask-022
   # installs are hardened together before either Docker-facing anchor changes.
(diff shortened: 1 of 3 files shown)


bjarneo:
Automated duplication check: this pull request looks similar to #9322, which covers the same Windows VM mount set-ID hardening. I keep that one open and close this one to consolidate review. If you feel this is the wrong decision, please open the PR again with a note on the difference.

PR #9521, opened 2026-09-01
Cycle a theme's backgrounds with up and down in the theme picker

buffpesos:
### Problem

The theme picker shows one image per theme, so selecting a theme also selects an unseen wallpaper through the theme's normal rotation. To inspect a theme's other backgrounds, users must first apply it and then open the separate background picker.

### Solution

Up and down now cycle the highlighted theme's backgrounds inside the theme picker. Left and right continue to cycle themes. If the user stops on a background, that exact background is applied; leaving a theme on its preview preserves the existing rotation behavior.

theme-picker-1-preview

The vertical axis uses the same neighboring-image treatment as the horizontal carousel: bands above and below preview the previous and next background, and labels explain what each arrow pair controls.

theme-picker-2-background

### Implementation

- Image-selector rows gain an optional group column. Rows in the same group become variants of one carousel item; ungrouped callers retain existing behavior.
- `omarchy-menu-images --group-by-dir` groups each directory under its name. `--item-label` and `--variant-label` name the two axes.
- `omarchy-theme-switcher` stages one directory of links per theme: preview first, then the theme's backgrounds and matching user backgrounds. The selected theme and optional background are returned together.
- `omarchy-theme-set <theme> [background]` matches an explicit background against the staged theme. Unknown or omitted backgrounds retain the existing rotation behavior.
- Installed themes remain restricted to files that theme staging can safely carry; user-authored and stock themes retain their existing link behavior.
- Lazy still-image thumbnails drain through one bounded batch. Native video wallpapers remain supported and use their own narrower thumbnail queue so FFmpeg codec threads cannot fan out across every core.
- MP4, M4V, MOV, WebM, MKV, and AVI files participate in grouped theme variants and named previews.

### Compatibility

The image-selector IPC gains optional item and variant label arguments. The CLI and QML caller ship together, and the public IPC documentation is updated.

The picker is 705px tall to accommodate the vertical preview bands; it remains usable on 768px-tall displays.

### Testing

Affected tests pass:

- `test/shell.d/menu-images-test.sh`
- `test/shell.d/theme-switcher-test.sh`
- `test/shell.d/theme-set-background-test.sh`
- `test/shell.d/image-picker-test.sh`
- `test/shell.d/background-test.sh`
- `test/shell.d/runtime-smoke-test.sh`
- `test/shell.d/video-background-test.sh`

The full shell suite was also run. Its remaining failures are outside this diff: the existing bar-icon geometry check, three checks requiring a separate `omarchy-pkgs` checkout, and the locate integration test encountering a non-UTF-8 indexed file. All picker, theme, background, runtime, and video-wallpaper tests pass.

The picker was also verified in the running UI with stock themes, installed themes, and user-added per-theme backgrounds.

### DIFF
--- a/bin/omarchy-dev-benchmark-theme-switcher
+++ b/bin/omarchy-dev-benchmark-theme-switcher
@@ -17,7 +17,7 @@ Usage:
   omarchy dev benchmark theme switcher [--repeat=<count>] [--keep-cache]
 
 Measure the non-interactive parts of the theme switcher:
-- theme preview index build (omarchy-theme-switcher before UI handoff)
+- theme image index build (omarchy-theme-switcher before UI handoff)
 - lazy selector row prep used by the interactive theme switcher
 - full thumbnail cache warmup cost (omarchy-menu-images --cache-only)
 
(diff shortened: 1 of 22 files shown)


buffpesos:
@ryanrhughes @dhh please do take a look - this lets users see the background images at the time of picking a theme!

bjarneo:
Automated duplication check: this pull request looks similar to #10520, which covers the same theme background cycling. I keep that one open and close this one to consolidate review.

PR #9523, opened 2026-09-01
Refocus lock screen when session secures

### LINKED ISSUE #9418: Lock screen password field loses keyboard focus race — forcePasswordFocus() fires before WlSessionLock.secure becomes true
### System details

NVIDIA 4090, Omarchy Quattro, Apple Magic Keyboard (1st Gen), 
13th Gen Intel(R) Core(TM) i9-13900K (16+16) @ 5.80 GHz
GeForce RTX 4090
### What's wrong?

 Environment: Omarchy 4.0.1-1 (package version; /usr/share/omarchy/version reports 4.0.0.alpha)

  Description:
  When the lock screen appears, the password field sometimes doesn't receive keyboard input — no keystrokes register — until you click into the field with the mouse first.

  Root cause (confirmed via journalctl):

  In Service.qml, LockView.inputEnabled is bound to root.lockRequested, and when that becomes true, LockView.qml's onInputEnabledChanged handler calls forcePasswordFocus(). However,
  lockRequested becomes true ~500ms before the actual Wayland session lock (WlSessionLock.secure) actually hands keyboard input over to the lock surface:

  17:34:31.287Z  lock-requested       <- forcePasswordFocus() fires here
  17:34:31.808Z  secure=true          <- Wayland ACTUALLY routes keyboard input here

  Service.qml's sessionLock.onSecureStateChanged handler only calls root.startFingerprint() — it never re-focuses the password field. As a result, the QML-level focus request happens
  before real keyboard input is routed to the surface, so the first keystrokes are lost and the user has to click into the field with the mouse, which triggers forcePasswordFocus()
  again via MouseArea.onClicked.

  Relevant files:
  - shell/plugins/lock/LockView.qml — onInputEnabledChanged: { if (inputEnabled) Qt.callLater(forcePasswordFocus) }
  - shell/plugins/lock/Service.qml — sessionLock.onSecureStateChanged (starts fingerprint auth, doesn't refocus the password field)

  Suggested fix: Add a call to lockView.forcePasswordFocus() inside onSecureStateChanged (when secure becomes true), so the focus request is repeated at the moment keyboard input
  actually arrives.

  Steps to reproduce:
  1. Lock the screen (manually, or after resuming from suspend).
  2. As soon as the lock screen appears, try typing the password immediately.
  3. No characters appear; you have to click into the password field with the mouse first.

mlmrx:
## Summary

- refocus the password field when `WlSessionLock.secure` becomes true and keyboard input reaches the lock surface
- retain the existing early focus request while closing the secure-state race
- add focused regression coverage for the lock service's secure handler

Fixes #9418

## Testing

- `bash test/shell.d/lock-focus-test.sh`
- `bash test/shell.d/lock-blank-fingerprint-test.sh`
- `bash test/shell.d/lock-stranded-recovery-test.sh`
- `bash test/shell.d/monitor-recovery-test.sh`
- `git diff --cached --check`

### DIFF
--- a/shell/plugins/lock/Service.qml
+++ b/shell/plugins/lock/Service.qml
@@ -238,6 +238,7 @@ Item {
         root.pendingSessionLock = false
         sessionLockStabilizeTimer.stop()
         pendingSessionLockTimer.stop()
+        lockView.forcePasswordFocus()
         root.startFingerprint()
       }
     }
--- a/test/shell.d/lock-focus-test.sh
+++ b/test/shell.d/lock-focus-test.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+set -euo pipefail
+
+source "$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd)/base-test.sh"
+
+run_node_test <<'JS'
+const fs = require('fs')
+const serviceQml = fs.readFileSync(path.join(root, 'shell/plugins/lock/Service.qml'), 'utf8')
+
+const secureHandlerStart = serviceQml.indexOf('onSecureStateChanged:')
+const lockHandlerStart = serviceQml.indexOf('onLockStateChanged:', secureHandlerStart)
+const secureHandler = serviceQml.slice(secureHandlerStart, lockHandlerStart)
+const secureBlockMatch = secureHandler.match(/if \(secure\) \{([\s\S]*?)\n      \}/)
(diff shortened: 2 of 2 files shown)


isho:
I tried this on my own system and it didn't work. I also tried this solution: https://github.com/omacom/omarchy/pull/9181/changes and it did work. So I recommend closing this PR and accepting the other one (as long as it meets your requirements)

PR #9527, opened 2026-09-01
Wait for real output before shell launch

### LINKED ISSUE #9529: Blank screen after docked clamshell login
### System details

Omarchy 4.0.2-1, Hyprland, laptop docked in clamshell mode, internal eDP-1 disabled, external DP-7 Dell P2725DE 2560x1440 active.

## What happened

When logging in after reboot with a laptop docked to an external monitor and the laptop lid closed, login succeeds but the session shows a blank screen.

The same machine logs in normally if either:

- the laptop lid is open during login, or
- the laptop is undocked during login

After a successful login, closing the lid works normally and the external display continues working.

## Expected behavior

A docked laptop with the lid closed should show the Omarchy session on the external monitor after login.

## Steps to reproduce

1. Dock laptop to external monitor.
2. Close laptop lid.
3. Reboot.
4. Log in.
5. Observe blank screen after login.

## Root cause evidence

The user journal shows Omarchy shell starting before Hyprland has a real output available:

```text
omarchy-shell: qt.qpa.wayland: There are no outputs - creating placeholder screen
omarchy-shell: WARN: Layershell screen does not correspond to a real screen. Letting the compositor pick.
Once the session is recovered, hyprctl monitors all shows the external monitor active and the internal panel disabled for clamshell mode.
Workaround verified
Adding a startup guard that waits until hyprctl monitors all -j reports an enabled output with non-zero width and height, then reconciles clamshell state and launches/restarts the shell, resolves the issue.
System details
- Omarchy version: 4.0.2-1
- Session: Hyprland / Omarchy shell
- Setup: laptop docked to external monitor, lid closed
- Internal display: eDP-1
- External monitor: DP-7, Dell P2725DE, 2560x1440

tbasten:
## What changed

Launch Omarchy shell through a new `omarchy-launch-shell-ready` helper. The helper waits until Hyprland reports at least one enabled real output before starting the shell, then reconciles clamshell monitor state and launches `omarchy-launch-shell`.

If no real output appears after the bounded wait, it logs and falls back to launching the shell.

## Why

When logging in docked with the laptop lid closed, Hyprland can start before the external display is fully available. Omarchy shell may then start against Qt's placeholder screen:

`There are no outputs - creating placeholder screen`

That leaves the user logged in but looking at a blank screen. Waiting for a real output before shell launch fixes the docked clamshell login path.

fixes #9529

## Verification

- Reproduced locally: docked laptop, lid closed, reboot, login caused blank screen.
- Verified locally: same flow works after the change.
- `bash ./test/shell.d/launch-shell-ready-test.sh`
- `bash ./test/shell.d/bin-style-test.sh`
- `bash ./test/cli`
- `bash -n bin/omarchy-launch-shell-ready test/shell.d/launch-shell-ready-test.sh`
- `git diff --check`

`bash ./test/all` was also run. It still fails due to missing local `omarchy-pkgs` checkout / `OMARCHY_PKGS_PATH`, unrelated to this change.

### DIFF
--- a/bin/omarchy-launch-shell-ready
+++ b/bin/omarchy-launch-shell-ready
@@ -1,0 +1,28 @@
+#!/bin/bash
+
+# omarchy:summary=Launch the Omarchy shell after Hyprland has a real output
+# omarchy:hidden=true
+
+ready_attempts=${OMARCHY_HYPRLAND_READY_ATTEMPTS:-40}
+ready_delay=${OMARCHY_HYPRLAND_READY_DELAY:-0.25}
+
+real_output_ready() {
+  hyprctl monitors all -j | jq -e '
+    any(.[]; .disabled != true and (.width // 0) > 0 and (.height // 0) > 0)
+  ' >/dev/null 2>&1
+}
+
+for (( attempt = 1; attempt <= ready_attempts; attempt++ )); do
+  if real_output_ready; then
+    omarchy-hyprland-monitor-clamshell || true
+    exec omarchy-launch-shell
+  fi
+
+  if (( attempt < ready_attempts )); then
+    sleep "$ready_delay"
+  fi
+done
+
+logger -t omarchy-shell "Starting Omarchy shell before Hyprland reported a real output."
+omarchy-hyprland-monitor-clamshell || true
+exec omarchy-launch-shell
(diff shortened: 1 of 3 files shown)


tbasten:
After a week or so, the issue has re-occurred again. Closing PR until I have a verified solution.

PR #9528, opened 2026-09-01
Don't report Tailscale removal as done after a cancelled sudo

### LINKED ISSUE #9526: Remove Tailscale reports success after cancelling sudo, and still tears down plugin/webapp
### System details

Dell OptiPlex 7010, Intel Core i5-3570, AMD Radeon RX 580, Omarchy 4.0.2-1, Hyprland 0.56.2

Debug log: will-be-replaced

### What's wrong?

After installing Tailscale from the menu, searching "Tailscale" again only offers **Remove** (Install is hidden once the package is present). Choosing that launches `omarchy-remove-service-tailscale` in a presentation terminal.

Ctrl+C at the sudo password prompt still:

1. Prints `Tailscale has been removed.`
2. Shows the green **Done!** screen
3. Does **not** uninstall the package or disable `tailscaled`
4. **Does** run the unprivileged teardown: `tailscale down`, disable `omarchy-tailscale-receive.service`, disable the bar plugin, and delete the Tailscale web app

Reproduced twice on this machine. Journal for each attempt:

```
pam_unix(sudo:auth): conversation failed
pam_unix(sudo:auth): auth could not identify password for [mike]
```

twice per attempt (`sudo systemctl disable --now tailscaled`, then `sudo pacman -Rns`). Pacman log has only the install (`installed tailscale (1.102.3-1)`), never a remove. `tailscaled` stayed enabled and running.

#### Steps

1. Install Tailscale from the menu (Install → Service → Tailscale).
2. Open the launcher search and type `Tailscale`.
3. Select the matching row (it is Remove, because Install is gated on `! omarchy-pkg-present tailscale`).
4. At the sudo password prompt, press Ctrl+C.

#### Expected

Cancelling sudo aborts. No "removed" / **Done!** message. Plugin, web app, receive service, and `tailscale down` should not have already run.

#### Actual

The presentation terminal reports success. The package and `tailscaled` are still there. The Omarchy-side pieces are gone.

#### Why

`bin/omarchy-remove-service-tailscale` does user-level teardown *before* sudo, swallows the cancelled disable with `|| true`, then continues:

```bash
tailscale down 2>/dev/null || true
systemctl --user disable --now omarchy-tailscale-receive.service 2>/dev/null || true
sudo systemctl disable --now tailscaled.service 2>/dev/null || true
omarchy-plugin-disable omarchy.tailscale
omarchy-webapp-remove "Tailscale" 2>/dev/null || true
omarchy-pkg-drop tailscale

echo ""
echo "Tailscale has been removed."
```

Ctrl+C at a sudo password prompt is PAM exit 1, not 130. The presentation wrapper only skips **Done!** on 130:

```bash
omarchy-show-logo; $cmd; if (( $? != 130 )); then omarchy-show-done; fi
```

Related launcher footgun: once Tailscale is installed, search for "Tailscale" has no install/setup hit, only Remove, with no confirmation.

AtypicalMike:
Fixes #9526

Cancelling sudo during Remove → Tailscale still printed `Tailscale has been removed.`, showed green **Done!**, and disabled the bar plugin / web app / Taildrop receive service. The package and `tailscaled` stayed installed because the privileged steps never ran.

`sudo` at a password prompt exits 1 on Ctrl+C (PAM `conversation failed`), not 130. The remove script swallowed that with `|| true`, kept going, and always echoed success. The presentation wrapper then treated every non-130 exit as success.

**Remove script:** authenticate with `sudo -v` before any teardown, map a cancelled prompt to exit 130, and stop swallowing `systemctl disable` failure.

**Presentation:** show **Done!** only on exit 0; on other non-cancel failures show **Failed** and still wait for a key so the error stays readable.

```
bash test/shell.d/remove-service-tailscale-test.sh
bash test/shell.d/floating-terminal-test.sh
./test/cli
```

### DIFF
--- a/bin/omarchy-launch-floating-terminal-with-presentation
+++ b/bin/omarchy-launch-floating-terminal-with-presentation
@@ -8,6 +8,6 @@
 source omarchy-restart-gum
 
 cmd="$*"
-presentation_script="omarchy-show-logo; $cmd; if (( \$? != 130 )); then omarchy-show-done; fi"
+presentation_script="omarchy-show-logo; $cmd; status=\$?; if (( status == 0 )); then omarchy-show-done; elif (( status != 130 )); then omarchy-show-done --failed; fi"
 
 exec setsid uwsm-app -- xdg-terminal-exec --app-id=org.omarchy.terminal --title=Omarchy -e bash -c "$presentation_script"
--- a/bin/omarchy-remove-service-tailscale
+++ b/bin/omarchy-remove-service-tailscale
@@ -3,9 +3,17 @@
 # omarchy:summary=Remove Tailscale and its bar plugin.
 # omarchy:requires-sudo=true
 
+set -euo pipefail
+
+# Authenticate before any teardown. Ctrl+C at the password prompt is PAM
+# exit 1, not 130; map it so the presentation wrapper does not show Done.
+if ! sudo -v; then
+  exit 130
+fi
+
 tailscale down 2>/dev/null || true
 systemctl --user disable --now omarchy-tailscale-receive.service 2>/dev/null || true
-sudo systemctl disable --now tailscaled.service 2>/dev/null || true
+sudo systemctl disable --now tailscaled.service
 omarchy-plugin-disable omarchy.tailscale
 omarchy-webapp-remove "Tailscale" 2>/dev/null || true
 omarchy-pkg-drop tailscale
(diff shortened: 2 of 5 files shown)


bjarneo:
Automated duplication check: this pull request looks similar to #11482, which covers the same presentation failure status. I keep that one open and close this one to consolidate review.

PR #9539, opened 2026-09-01
Add cursor theme selection to the Style menu

TheLinuxITGuy:
## What

Adds a **Style → Cursor** submenu to choose and apply the system cursor theme.

- **`omarchy-cursor-set <theme>`** — applies a cursor theme system-wide. If the theme is a Bibata variant (`Bibata-Modern-*`) that isn't installed yet, it downloads it from the Bibata Cursor GitHub release. It writes the chosen theme into a managed `~/.config/hypr/cursor.lua` (Hyprland env), updates the X.org default index, applies it via `gsettings` and `hyprctl setcursor`, then fires the `cursor-set` hook.
- **`omarchy-cursor-current`** — prints the active cursor theme (used for the ✓ checkmark in the menu).
- Registers the new `cursor` command group.
- Ships a default `config/hypr/cursor.lua` and wires `require("hypr.cursor")` into the user Hyprland config so the managed file is loaded on login; `omarchy refresh hyprland` refreshes it too.
- Menu options: Bibata Modern Classic / Amber / Ice, Adwaita, Yaru.

## Why

Lets users change the cursor from the Omarchy Style menu, downloading Bibata cursors on demand instead of requiring manual installation.

## Testing

- `./test/cli` passes (command metadata, routing, group registration).
- Menu JSONC validated against the shell's `MenuModel`.
- Verified locally: each Bibata variant downloads and applies, and `omarchy-cursor-current` reports the active theme.

Note: the shell/acceptance suites require a running graphical session and were not run in the contributing environment.


jondkinney:
I tested this PR in a running Omarchy build and opened a one-commit follow-up against the feature branch: https://github.com/TheLinuxITGuy/omarchy/pull/3

It adds the macOS option and addresses the live-switching issue I found during testing. These bundled themes are XCursor themes, so the generated config disables Hyprcursor, then reloads Hyprland and calls setcursor to apply the selected theme immediately. It also validates downloaded cursor payloads and adds regression tests.

Both the cursor test and the full CLI suite pass.

PR #9545, opened 2026-09-01
Add Keychron RGB handler to the theme keyboard pipeline

ueltyx:
## Summary

Adds Keychron RGB support to the existing `keyboard.rgb` theme pipeline, following the `omarchy-theme-set-keyboard-asus-rog` / `-f16` handler pattern exactly.

Keychron's own firmware (V/Q/K "Ultra" and "Max" series, protocol v11+) implements the VIA raw-HID protocol on a vendor config channel (USB usage page `0xFF60`, usage `0x61`). It is the same lighting command set the Framework 16 handler already speaks via `qmk_hid`: solid effect (`[7,3,2,1]`), HSV color (`[7,3,4,h,s]`), NVRAM save (`[9,3]`). The protocol was extracted from Keychron's own Launcher web app (WebHID client) — no firmware flashing, no reverse-engineered vendor blobs.

## Changes

- `bin/omarchy-keychron-rgb` — transport: bash + embedded Python (stdlib only, no new packages, nothing to compile). Auto-discovers the config channel by vendor `3434` + usage-page signature and round-trip-probes it, so it works across models and connection modes (with cable and dongle both enumerated, one channel is a dead end depending on the keyboard's mode switch — the probe picks the live one).
- `bin/omarchy-hw-keychron` — hardware detect (USB vendor `3434`), `hw-*` pattern.
- `bin/omarchy-theme-set-keyboard-keychron` — the handler; no-op without a Keychron present.
- `bin/omarchy-theme-set-keyboard` — dispatch entry.
- `default/udev/keychron-rgb.rules` + `install/hardware/keychron-rgb.sh` + migration — vendor-wide uaccess rule (also unblocks the official Keychron Launcher web app), mirroring `framework/qmk-hid.sh`.
- `install/hardware/all.sh` — install wiring.

## Verified behavior

Tested on a Keychron V1 Ultra 8K (ZMK-based, `3434:0c10`) plus Ultra-Link 8K 2.4GHz dongle (`3434:d028`):

- All 22 stock themes switched back-to-back; keyboard color read back over HID and matched the theme color byte-exact every time, including Tokyo Night's hand-picked `keyboard.rgb` (`ff00ff`) overriding the accent template.
- User-created themes work automatically (accent template → staged `keyboard.rgb`).
- Connection matrix: wired works; 2.4GHz dongle works (it tunnels the config channel; needs a radio-wake timeout, handled); Bluetooth intentionally holds the last synced color — the stock firmware exposes no config channel over BT (confirmed via the BT HID descriptor, the GATT service tree, and the Launcher client being WebHID/USB-only).
- No-op cost on non-Keychron machines: one sysfs grep; the handler exits before spawning the transport.

## Testing

- `./test/cli` — 116 ok, 0 failures (includes metadata lint for the new commands).
- `./test/shell` — 2905 ok, 0 failures.
- End-to-end on real hardware as described above.

## Transparency

This contribution was developed end-to-end with AI assistance (an agentic CLI coding assistant): protocol extraction from the Launcher bundle, tool implementation, and the automated 22-theme verification. The submitter performed all hardware verification on their machine.

One natural follow-up (not in this PR to keep it scoped): a udev `RUN+` on the rule could invoke the handler on (re)plug, re-syncing the keyboard at boot and whenever it returns from wireless to wired. Happy to add if wanted.

### DIFF
--- a/bin/omarchy-hw-keychron
+++ b/bin/omarchy-hw-keychron
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+# omarchy:summary=Detect whether a Keychron keyboard or receiver is connected over USB.
+
+grep -q "^3434$" /sys/bus/usb/devices/*/idVendor 2>/dev/null
--- a/bin/omarchy-keychron-rgb
+++ b/bin/omarchy-keychron-rgb
@@ -1,0 +1,126 @@
+#!/bin/bash
+
+# omarchy:summary=Set or query Keychron keyboard RGB over the raw HID config channel (wired or 2.4GHz dongle)
+# omarchy:args=<set <RRGGBB>|get>
+# omarchy:examples=omarchy keychron-rgb set 89b4fa | omarchy keychron-rgb get
+# omarchy:hidden=true
+
+exec python3 - "$@" <<'PYTHON'
+import colorsys
+import glob
+import os
+import select
+import sys
+import time
+
+VID = 0x3434
+MAGIC = b"\x06\x60\xff\x09\x61"
+REPORT_LEN = 32
+REPLY_TIMEOUT = 1.0
+
+CMD_BACKLIGHT_GET = 8
+CMD_BACKLIGHT_SET = 7
+CMD_BACKLIGHT_SAVE = 9
+CH_LIGHTING = 3
+LID_EFFECT = 2
+LID_COLOR = 4
+
+
+def find_live_device():
+  for path in sorted(glob.glob("/sys/class/hidraw/hidraw*/device")):
+    try:
+      with open(path + "/uevent") as f:
+        hid_id = next((line for line in f if line.startswith("HID_ID=")), "")
+    except OSError:
+      continue
+    try:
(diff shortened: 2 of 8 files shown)


sprajs:
There is a migration-ID collision with #9478: both PRs add `migrations/1788129995.sh`, but that PR uses it to migrate Chromium notification flags and this one installs the Keychron udev rule. These are unrelated operations, so both need distinct completion markers.

Could one migration receive an unused timestamp before the PRs land? This avoids the add/add merge conflict and prevents either operation being lost when resolving it.

The same filename is also used by #9522 for nwg-displays workspace loading and #9562 for the XDG Desktop directory. Those four unrelated migrations need separate IDs.

ueltyx:
Good catch — fixed in 0f80aad2: the migration is now `migrations/1789739221.sh` (fresh timestamp at rename time; unused on `quattro`, across all open PRs, and in any existing branch or marker).

For what it's worth the collision was wider than the four PRs listed here: `1788129995.sh` was also carried by #9276, #9445, and #9446 — seven open PRs shared the name. The completion-marker point stands either way; thanks for flagging.

PR #9548, opened 2026-09-01
Add a Setup > Region toggle with Chinese language and input method

ZacharyZhang-NY:
## Summary

Replaces #9519.

_Setup > Region_ toggles between World and China. China switches the system language to zh_CN, installs @manateelazycat's [rime-ice-installer](https://github.com/manateelazycat/rime-ice-installer) and runs it, makes Rime the active input method with Ctrl+Space as the toggle, and swaps the menu to Chinese labels. World puts the language and menu back and leaves the input method installed.

Also fixes `mergeMenuSources` to only override the fields a user extension declares, as `docs/menu.md` documents.

Companion: omacom/omarchy-pkgs#251.

## Test plan

- [x] `region-test.sh` (new): language, menu and input-method round-trips, idempotent re-apply
- [x] `menu-test.sh` and `menu-guards-test.sh`
- [x] `./test/all` at the clean-tree baseline
- [x] Fresh VM install verified end to end


default/omarchy/omarchy-menu.zh-cn.jsonc
```diff
@@ -0,0 +1,169 @@
...
+  "trigger.tests.disk-speedtest": {"label":"磁盘测速"},
+  "style.theme": {"label":"主题"},
+  "style.background": {"label":"壁纸"},
+  "style.unlock": {"label":"解锁画面"},
+  "style.font": {"label":"字体"},
+  "style.bar": {"label":"菜单栏"},
+  "style.bar.position": {"label":"位置"},
+  "style.bar.transparency": {"label":"透明"},
+  "style.screensaver": {"label":"屏保"},
+  "style.about": {"label":"关于页"},
```
huacnlee: Here should "关于"

default/omarchy/omarchy-menu.zh-cn.jsonc
```diff
@@ -0,0 +1,169 @@
...
+  "update.channel": {"label":"通道"},
+  "update.config": {"label":"配置","title":"重置为默认"},
+  "update.themes": {"label":"附加主题"},
+  "update.process": {"label":"进程","title":"重启"},
+  "update.hardware": {"label":"硬件","title":"重启"},
+  "update.firmware": {"label":"固件"},
+  "update.password": {"label":"密码"},
+  "update.timezone": {"label":"时区"},
+  "update.time": {"label":"时间"},
+  "update.channel.stable": {"label":"稳定 Stable"},
```
Laotree: duplicate

huacnlee: Here just keep `Stable` be ok.

default/omarchy/omarchy-menu.zh-cn.jsonc
```diff
@@ -0,0 +1,169 @@
...
+  "setup.plugin.add": {"label":"添加插件"},
+  "setup.plugin.clone": {"label":"克隆插件"},
+  "setup.plugin.remove": {"label":"移除插件"},
+  "setup.security": {"label":"安全"},
+  "setup.config": {"label":"配置"},
+  "setup.security.fingerprint": {"label":"指纹"},
+  "setup.security.passwordless-sudo": {"label":"免密 sudo"},
+  "setup.security.sudoless-docker": {"label":"免 sudo Docker"},
+  "setup.direct-boot": {"label":"直接启动"},
+  "setup.reset": {"label":"恢复出厂"},
```
huacnlee: "恢复出厂设置"

default/omarchy/omarchy-menu.zh-cn.jsonc
```diff
@@ -0,0 +1,169 @@
...
+  "learn.keybindings": {"label":"快捷键"},
+  "learn.omarchy": {"label":"Omarchy 手册"},
+  "learn.community": {"label":"社区"},
+  "trigger.emoji": {"label":"表情"},
+  "trigger.reminder": {"label":"提醒"},
+  "trigger.capture": {"label":"截取"},
+  "trigger.capture.screenshot": {"label":"截图"},
+  "trigger.capture.screenrecord.stop": {"label":"停止录屏"},
+  "trigger.capture.screenrecord": {"label":"录屏"},
+  "trigger.capture.text": {"label":"取字 OCR"},
```
huacnlee: "OCR 取字"

Harry-Chen (changes requested):
As the administrator of mirrors.tuna.tsinghua.edu.cn, due to both technical and non-technical reasons, I hereby request the removal of any hardcoded TUNA mirrors from the official configuration of this distribution, for now and the future. This leads to traffic and ethical burden to our mirror service, which is strongly undesirable.

bin/omarchy-region
```diff
@@ -0,0 +1,251 @@
...
+  echo "Usage: omarchy-region <world|china>"
+  exit 1
+  ;;
+esac
+
+npm_registry="https://registry.npmmirror.com"
+pip_index="https://pypi.tuna.tsinghua.edu.cn/simple"
+cargo_registry="sparse+https://rsproxy.cn/index/"
+go_proxy="https://goproxy.cn,direct"
+gem_source="https://mirrors.tuna.tsinghua.edu.cn/rubygems/"
```
huacnlee: Remove all mirrors pelease. Current is a AI time. Omarchy is also the AI native system, all users that must be has proxy, unless that they not use Claude, Codex.

So, we don't need built-in have those mirrors. Even say that not need for every one.

We can built-in have VPN apps, just on-tap to install proxy subscription.

ZacharyZhang-NY: Actually built in vpn apps does not meet the Gov complience. Users will need to install by themselves.

Laotree: @ZacharyZhang-NY 

I think you may be going in the wrong direction here. We shouldn’t assume that users depend on domestic mirrors for their development tools.

There are many scenarios, including but not limited to:

1. Chinese-speaking users who live outside China
2. Users in China who use the system while traveling or working abroad
3. Users from China whose network environment is already outside China

None of these users need a system-wide built-in proxy or mirror configuration.

So, as @huacnlee  suggested, we should remove all mirror/proxy configuration and leave this decision and setup entirely to the user.

default/omarchy/omarchy-menu.zh-cn.jsonc
ZacharyZhang-NY: 非常感谢。

bin/omarchy-region
ZacharyZhang-NY: That makes sense. Would it be better to keep the region set to China, but make mirror configuration optional during setup so users can choose based on their network environment?
Or, we can just write a documentation on website showing the tutorial on source list and let user config themselves.

Laotree: I think the second option is better.

We can keep the region as `China`, but I don't think the setup process should make any assumption about package mirrors. Region and mirror configuration are separate concerns.

The safest default is to keep the official upstream sources and leave mirror selection entirely to the user. For users who need domestic mirrors, we can provide documentation showing how to configure them manually.

I also think the documentation should avoid explicitly recommending or naming any specific domestic mirror provider. This would prevent putting additional traffic, maintenance, or other burdens on individual mirror maintainers, such as TUNA.

This also avoids hardcoding or maintaining any specific mirror in the official configuration, while still making it easy for users in China to configure a mirror themselves if their network requires it.

ZacharyZhang-NY: I would agree with that. I removed all the sources and will update the aliyun mirror list on website. I will contact aliyun.

ZacharyZhang-NY:
> As the administrator of mirrors.tuna.tsinghua.edu.cn, due to both technical and non-technical reasons, I hereby request the removal of any hardcoded TUNA mirrors from the official configuration of this distribution, for now and the future. This leads to traffic and ethical burden to our mirror service, which is strongly undesirable.

I removed all the mirrors and will just show a aliyun source list on website for users to config manually. Give them the option.

bin/omarchy-region
Laotree: That sounds good. One small point I'd like to clarify: I don't mean that Aliyun is not a good mirror, nor do I think Omarchy should avoid working with Aliyun.

My concern is more about the order of things. I think we should first contact the mirror provider and reach an agreement with them about being listed as a recommended mirror, and only then add it to the Omarchy website.

In other words, the website should record an established relationship rather than pre-selecting a mirror before the provider has agreed to it.

So if Aliyun is willing to work with Omarchy and the maintainers agree on the arrangement, I'm totally fine with listing it. I just don't think we should present Aliyun as a recommended mirror before that agreement exists.

ZacharyZhang-NY: Yes, we are currently talking to aliyun right now.

PR #9550, opened 2026-09-01
fix(polkit) Attribute polkit prompts to coding agents

tcballard:
**What this is**

When Quattro's themed polkit dialog opens for a `pkexec` that came from the default coding-agent window (`org.omarchy.agent`), the chip changes from `Authorize running 'cmd'` to `Codex wants: 'cmd'`.

Same PAM flow. One-shot grant. No NOPASSWD. Not a sudo askpass (see #5617).

Quickshell's Polkit `AuthFlow` does not expose subject PID, and UWSM names the agent scope after `xdg-terminal-exec`, so this attributes via a live `pkexec` parent walk to Hyprland class `org.omarchy.agent`. Slug from `OMARCHY_AGENT` if set, else `~/.config/omarchy/defaults/agent`.

Test: `./test/shell.d/polkit-test.sh`. Manual: `omarchy agent` then `pkexec true` vs a normal terminal, and Settings polkit while an agent window is open (should stay stock).


parnoldx:
**Nice, and one step further I would suggest: not just *who* is asking, but *what and why*.**

What I miss in the polkit prompt is a way for the caller to say what it's about to do and why. My idea: an optional description argument on `pkexec`, e.g.

```bash
pkexec --description "Install Docker engine" -- /usr/bin/apt install docker
```

The dialog would then show that description above the command, so the user can decide based on intent, not just a raw command line. I prototyped this with a wrapper script plus a custom polkit action and it works nicely (screenshot attached).

polkit

PR #9562, opened 2026-09-01
Keep desktop shortcuts out of home

### LINKED ISSUE #9556: XDG_DESKTOP_DIR="$HOME" makes apps scatter .desktop shortcuts into the home directory
### Summary

`omarchy-provision-user` points `DESKTOP`, `TEMPLATES` and `PUBLICSHARE` at `$HOME` and removes the
folders:

```bash
# /usr/bin/omarchy-provision-user  (omarchy 4.0.2-1, lines ~98-101)
xdg-user-dirs-update --set TEMPLATES "$HOME"
xdg-user-dirs-update --set PUBLICSHARE "$HOME"
xdg-user-dirs-update --set DESKTOP "$HOME"
rmdir ~/Templates ~/Public ~/Desktop 2>/dev/null || true
```

The intent is clear and I agree with it — Hyprland has no desktop metaphor, so three permanently empty
folders are noise. Pointing an XDG dir at the home directory is also the freedesktop-documented way to
say "this directory doesn't exist here", so this isn't a hack.

**The problem is that apps don't interpret it that way.** Nothing reads `XDG_DESKTOP_DIR=$HOME` as
"there is no desktop, skip it". They read it as "the desktop is `~`" and write there. So the setting
that exists to keep `$HOME` tidy is the thing that untidies it.

### Impact

On one machine, after about a day of ordinary use, `~` had collected:

```
steam.desktop                                     # byte-identical to /usr/share/applications/steam.desktop
chrome-<id>-Default.desktop                       # Chrome PWA shortcut
chrome-<id>-Default.desktop                       # Chrome PWA shortcut
chrome-<id>-Default.desktop                       # Chrome PWA shortcut
```

Two things make this worse than ordinary clutter:

1. **The files are inert.** Walker and the app menu read `~/.local/share/applications/`, where correct
   copies already exist. The `~` copies are never used by anything.
2. **They are stale duplicates.** The Chrome PWA copies in `~` lacked the `%U` argument and the
   `MimeType=` line that the `~/.local/share/applications/` versions carry, so they were also *worse*
   copies — e.g. missing `x-scheme-handler/mailto` registration.

Users are unlikely to connect loose `.desktop` files in `$HOME` to an XDG setting, so they accumulate
indefinitely.

### Reproduce

On a stock Omarchy install:

1. `omarchy-install-steam` (or install Steam by any means) and let it create a desktop shortcut, and/or
2. In Chrome, open any site → **Cast, save and share → Install page as app**
3. `ls ~/*.desktop`

Expected: nothing in `$HOME`. Actual: one `.desktop` file per shortcut.

### Suggested fix

Keep the intent, drop the side effect — give `DESKTOP` a real but out-of-the-way target rather than
`$HOME`:

```bash
xdg-user-dirs-update --set DESKTOP "$HOME/.local/share/desktop"
```

No visible folder in `$HOME`, and anything that insists on writing a desktop shortcut lands somewhere
ignorable. `TEMPLATES` and `PUBLICSHARE` are rarely written to by anything, so they matter less — but
the same reasoning applies.

An alternative, if a hidden path feels wrong, is to leave the folders in place and accept three empty
directories; that at least confines the writes.

### Environment

- omarchy `4.0.2-1`
- `xdg-user-dirs 0.20-1`
- `/etc/xdg/user-dirs.defaults` ships the standard layout (`DESKTOP=Desktop`, `TEMPLATES=Templates`,
  `PUBLICSHARE=Public`); the `$HOME` values come from `omarchy-provision-user`, not the system default.

### Related

- #1331 — proposal that Omarchy avoid creating files in the user's home directory (same concern, other direction)

rookepoole:
## Summary

- redirect the XDG desktop from `$HOME` to `~/.local/share/desktop` for new users and Omarchy 3 upgrades
- migrate existing users only when their desktop directory still resolves exactly to `$HOME`
- preserve customized desktop directories and keep the migration idempotent
- document the hidden desktop target

## Testing

- `bash test/shell.d/provision-user-test.sh`
- `bash test/shell.d/xdg-desktop-dir-migration-test.sh`
- `./test/all` (the changed tests pass; six unrelated environment-dependent tests fail because this checkout lacks `omarchy-pkgs` fixtures or the sandbox lacks the required terminal, network, or privilege state)

Closes #9556

### DIFF
--- a/bin/omarchy-provision-user
+++ b/bin/omarchy-provision-user
@@ -95,10 +95,11 @@ for skill in "$OMARCHY_PATH"/default/agents/skills/*/; do
   ln -sfn "$skill" ~/.gemini/config/skills/"$name"
 done
 
-mkdir -p ~/Downloads ~/Pictures ~/Videos ~/.config/gtk-3.0
+desktop_dir="$HOME/.local/share/desktop"
+mkdir -p ~/Downloads ~/Pictures ~/Videos ~/.config/gtk-3.0 "$desktop_dir"
 xdg-user-dirs-update --set TEMPLATES "$HOME"
 xdg-user-dirs-update --set PUBLICSHARE "$HOME"
-xdg-user-dirs-update --set DESKTOP "$HOME"
+xdg-user-dirs-update --set DESKTOP "$desktop_dir"
 rmdir ~/Templates ~/Public ~/Desktop 2>/dev/null || true
 touch ~/.config/gtk-3.0/bookmarks
 for dir in Downloads Projects Pictures Videos; do
--- a/bin/omarchy-upgrade-to-quattro
+++ b/bin/omarchy-upgrade-to-quattro
@@ -2000,11 +2000,12 @@ mkdir -p "$HOME/.config/omarchy/branding"
 [[ -f $root/icon.txt ]] && cp "$root/icon.txt" "$HOME/.config/omarchy/branding/about.txt"
 [[ -f $root/logo.txt ]] && cp "$root/logo.txt" "$HOME/.config/omarchy/branding/screensaver.txt"
 
-mkdir -p "$HOME/Downloads" "$HOME/Pictures" "$HOME/Videos" "$HOME/Projects" "$HOME/.config/gtk-3.0"
+desktop_dir="$HOME/.local/share/desktop"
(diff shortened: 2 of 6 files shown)


bjarneo:
Automated duplication check: this pull request looks similar to #9563, which covers the same XDG desktop dir move. I keep that one open and close this one to consolidate review.

PR #9564, opened 2026-09-01
Clear set-ID bits from Windows VM mount sources

### LINKED ISSUE #9374: windows-vm: chmod 0700 preserves directory setgid bit, causing launch to fail safety check
### System details

- CPU: AMD Ryzen 7 5700X3D
- Omarchy: 4.0.2-1
- Kernel: 7.1.9-arch1-2
- GNU coreutils: 9.11

### What's wrong?

### Steps to reproduce

   ```bash
   chmod g+s ~/Windows
   ```

4. Approve the polkit prompt.

```bash
dir=$(mktemp -d)
chmod 2700 "$dir"
chmod 0700 "$dir"
chmod a-s,u=rwx,go= "$dir"
rmdir "$dir"
```

### Root cause

```bash

  return 1
fi
```

### Expected behavior

### Suggested fix

```diff
-  chmod 0700 -- "/proc/$BASHPID/fd/$storage_fd" "/proc/$BASHPID/fd/$shared_fd" || {
+  chmod a-s,u=rwx,go= -- "/proc/$BASHPID/fd/$storage_fd" "/proc/$BASHPID/fd/$shared_fd" || {
```

aiqubits:
## Summary

- Explicitly clear set-ID bits while normalizing the pinned Windows VM storage and shared directories to mode `700`.
- Preserve the existing strict permission verification before creating Docker-facing bind mounts.
- Extend the mount-boundary regression test to cover legacy directories starting with mode `2755`.

## Testing

- `bash test/shell.d/windows-vm-mount-boundary-test.sh`
- `bash test/shell.d/windows-vm-test.sh`
- `./test/cli`
- `bash -n bin/omarchy-windows-vm test/shell.d/windows-vm-mount-boundary-test.sh`
- Verified setuid, setgid, and sticky-bit mode combinations normalize to `700`.

Fixes #9374

### DIFF
--- a/bin/omarchy-windows-vm
+++ b/bin/omarchy-windows-vm
@@ -580,7 +580,7 @@ prepare_caller_mounts() {
   # Privacy is an explicit preflight step for both already-pinned sources, not
   # a side effect halfway through the two-mount transaction. Old umask-022
   # installs are hardened together before either Docker-facing anchor changes.
-  chmod 0700 -- "/proc/$BASHPID/fd/$storage_fd" "/proc/$BASHPID/fd/$shared_fd" || {
+  chmod a-s,u=rwx,go= -- "/proc/$BASHPID/fd/$storage_fd" "/proc/$BASHPID/fd/$shared_fd" || {
     exec {storage_fd}<&-
     exec {shared_fd}<&-
     return 1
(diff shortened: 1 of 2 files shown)


bjarneo:
Automated duplication check: this pull request looks similar to #9322, which covers the same Windows VM mount set-ID hardening. I keep that one open and close this one to consolidate review. If you feel this is the wrong decision, please open the PR again with a note on the difference.

PR #9577, opened 2026-09-01
Harden RetroArch desktop launcher paths

cmdr-chara:
## Summary

- Reject control characters in game and core paths before creating a desktop launcher.
- Escape Desktop Entry string values and Exec arguments according to their respective formats.
- Add regression coverage for injected newlines and special-character ROM names.

## Security impact

A newline in a ROM path could previously introduce additional Desktop Entry keys, including a second `Exec` directive. The launcher now refuses control characters and preserves quotes, percent signs, backslashes, spaces, and other supported path characters without changing the resulting RetroArch argument.

## Testing

- `bash test/shell.d/retro-install-desktop-entry-test.sh`
- `./test/cli`
- `bash -n bin/omarchy-games-retro-install test/shell.d/retro-install-desktop-entry-test.sh`
- `git diff --check`

`./test/shell` also ran; four unrelated environment checks remain unavailable because this checkout does not include the separate `omarchy-pkgs` repository (plus one independent launch-about check). The new focused test passes.

### DIFF
--- a/bin/omarchy-games-retro-install
+++ b/bin/omarchy-games-retro-install
@@ -46,6 +46,37 @@ if [[ ! -f $core_path ]]; then
   exit 1
 fi
 
+# Desktop Entry values are line-oriented, and Exec arguments have their own
+# quoting rules. Reject control characters before serializing either path.
+if [[ $game_path == *[[:cntrl:]]* || $core_path == *[[:cntrl:]]* ]]; then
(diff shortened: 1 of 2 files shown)


bjarneo:
Automated duplication check: this pull request looks similar to #9420, which covers the same retro desktop escaping. I keep that one open and close this one to consolidate review.

PR #9605, opened 2026-09-01
Clear set-ID bits when securing the Windows VM mount sources

### LINKED ISSUE #9374: windows-vm: chmod 0700 preserves directory setgid bit, causing launch to fail safety check

### LINKED ISSUE #9540: omarchy-windows-vm install: fails to write config when mount source dir has setgid bit (chmod 0700 doesn't clear S_ISGID)
## Bug

`omarchy-windows-vm install` fails with **"❌ Failed to write the Windows VM configuration."** when either mount source directory (`~/Windows` or `~/.windows`) carries the **setgid bit**.

## Root cause

In `prepare_caller_mounts()` (`/usr/bin/omarchy-windows-vm`, around lines 580-592):

```bash
chmod 0700 -- "/proc/$BASHPID/fd/$storage_fd" "/proc/$BASHPID/fd/$shared_fd" || { ...; return 1; }
storage_mode=$(stat -Lc '%a' "/proc/$BASHPID/fd/$storage_fd") || storage_mode=""
shared_mode=$(stat -Lc '%a' "/proc/$BASHPID/fd/$shared_fd") || shared_mode=""
  ...; return 1
fi
```

The code assumes `chmod 0700` guarantees mode `700`. It does not when the setgid bit is set: the Linux kernel **preserves `S_ISGID`** on `chmod` when the caller is in the file's group or has `CAP_FSETID` (which root always has). So a directory with mode `2700` stays `2700` after `chmod 0700` (`chmod` returns 0), `stat '%a'` yields `2700`, the `!= 700` check fails, `prepare_caller_mounts` returns 1, `write_compose` fails, and install aborts.

Reproduced with a plain dir: `mkdir /tmp/t && chmod 2700 /tmp/t && chmod 0700 /tmp/t && stat -c %a /tmp/t` → still `2700`.

## Steps to reproduce

1. `mkdir -p ~/Windows && chmod 2700 ~/Windows` (e.g. leftover from an older setup/tool)
2. `omarchy-windows-vm install` (any valid config)
3. After the pkexec authorization, install aborts: `❌ Failed to write the Windows VM configuration.`
4. Root-side stderr: the run dies in `prepare_caller_mounts` with no message (the `!= 700` branch returns silently), exit 2.

## Expected

Install should succeed. A setgid bit on the shared/storage folder (which the user legitimately owns) must not block VM setup.

## Suggested fix

Explicitly clear the setgid bit before the mode check (a plain `chmod 0700` cannot be relied on):

```bash
chmod 0700 -- "/proc/$BASHPID/fd/$storage_fd" "/proc/$BASHPID/fd/$shared_fd" || { ...; return 1; }
chmod g-s -- "/proc/$BASHPID/fd/$storage_fd" "/proc/$BASHPID/fd/$shared_fd" || { ...; return 1; }
```

or compare ignoring the special bits: `(( (8#$mode & 0777) == 700 ))`.

## System

- Omarchy 4.0.2-1 (Arch Linux)
- Kernel 7.1.9-arch1-2
- btrfs root, `@home` subvolume mounted at `/home`
- User uid 1000 in `docker` group; sudoless Docker off

### LINKED ISSUE #9567: omarchy-windows-vm launch fails silently when ~/Windows has the setgid bit (numeric chmod can't clear it)
# [BUG] omarchy-windows-vm launch fails silently when ~/Windows has the setgid bit set (numeric chmod cannot clear it)

## Environment

- Omarchy 4.0.2-1
- Arch Linux, kernel 7.1.9-arch1-2
- Windows VM installed via `omarchy windows vm install`, container `omarchy-windows` running and healthy (web console at 127.0.0.1:8006 shows a booted desktop; RDP answers X.224 on 3389)
- User not in the `docker` group (default), so `launch` elevates through pkexec

## What happened

```
$ omarchy windows vm launch
Starting Windows VM (this may prompt for authorization)...
❌ Failed to start Windows VM!
   Try checking: omarchy-windows-vm status
```

No reason is printed — not on the terminal, not in a notification. `status` reports RUNNING, the web console works, RDP is up; only `launch` refuses.

## Root cause

`prepare_caller_mounts()` in `omarchy-windows-vm` chmods the opened FDs of the caller's `~/.windows` / `~/Windows` and then verifies the mode:

```bash
chmod 0700 -- "/proc/$BASHPID/fd/$storage_fd" "/proc/$BASHPID/fd/$shared_fd" || ...
  exec {storage_fd}<&-
  exec {shared_fd}<&-
  return 1      # <-- silent
fi
```

In my case `~/Windows` carried the setgid bit (mode `2700`, `drwx--S---`). **A numeric `chmod 0700` cannot clear setgid on a directory** — the setgid bit is left as-is (setting it via numeric mode works, clearing it doesn't; `g-s` is required). So the chmod "succeeds", the verification still reads `2700`, and the function returns 1 without any message. `launch` surfaces only the generic failure box.

Confirmed with `bash -x` under pkexec:

```
+ chmod 0700 -- /proc/396250/fd/11 /proc/396250/fd/12
++ stat -Lc %a /proc/396582/fd/11
+ storage_mode=700
++ stat -Lc %a /proc/396584/fd/12
+ shared_mode=2700
+ [[ 700 != 700 ]]
+ [[ 2700 != 700 ]]
+ exec
+ return 1
```

## Workaround

```
chmod g-s ~/Windows
```

After this, `omarchy windows vm launch` works (verified).

## Suggested fix

1. Clear the setgid bit explicitly, e.g. `chmod 700,g-s` (symbolic `g-s`), at the FD chmod in `prepare_caller_mounts()` — and audit the other numeric `chmod 0700` calls on user-owned directories (there are two more) for the same trap.
2. Print a diagnostic before returning 1 in that failure path — the silent exit makes this (and any future mode mismatch) undiagnosable from the user side.

## Reproduce

1. Install the Windows VM normally.
2. `chmod g+s ~/Windows` (any source of setgid works; mine arrived with mode 2700).
3. `omarchy windows vm launch` → generic failure, no diagnostics.

omarchybot:
Closes #9374. Closes #9540. Closes #9567.

`omarchy-windows-vm launch` and `install` abort — the first with `❌ Failed to start Windows VM!`, the second with `❌ Failed to write the Windows VM configuration.` — when `~/Windows` or `~/.windows` carries the setgid bit. Nothing is written to stderr, so `status` reporting RUNNING and a working web console leave no way to tell what refused.

`prepare_caller_mounts()` hardens both pinned sources with `chmod 0700` and then asserts `stat -Lc '%a'` is exactly `700`. `chmod` with a numeric mode does not clear `S_ISGID` on a directory: the kernel preserves it whenever the caller is in the file's group or holds `CAP_FSETID`, which root always does. A directory at `2700` stays `2700`, `chmod` still returns 0, and the `!= 700` branch rejects a source the same function had just been asked to secure — then returns 1 without a message.

Symbolic modes clear the set-ID bits, so the check now sees the mode the chmod was meant to produce. The same substitution applies at the two other sites that make these directories private: `prepare_user_mount_sources()`, which prepares the very directories the elevated path re-checks, and `write_credentials()`, where a set-ID bit would hand the directory holding the plaintext RDP password a group it was never meant to have.

The mode-verification branch gains a diagnostic. It stays reachable for reasons a user can act on, and returning silently is what made this take three separate reports to characterise.

Reviewed by Codex at xhigh, which confirmed the mode arithmetic across every set-ID combination and pointed out that the existing suites start these directories at `0755` and so pass against the broken chmod too; the regression cases added here start them at `2700`/`6700` and fail without the fix.

### DIFF
--- a/bin/omarchy-windows-vm
+++ b/bin/omarchy-windows-vm
@@ -580,7 +580,9 @@ prepare_caller_mounts() {
   # Privacy is an explicit preflight step for both already-pinned sources, not
   # a side effect halfway through the two-mount transaction. Old umask-022
   # installs are hardened together before either Docker-facing anchor changes.
-  chmod 0700 -- "/proc/$BASHPID/fd/$storage_fd" "/proc/$BASHPID/fd/$shared_fd" || {
+  # Symbolic, because a numeric mode leaves S_ISGID standing on a directory and
+  # the mode check below then rejects a source we were asked to secure.
+  chmod u=rwx,go=,a-s -- "/proc/$BASHPID/fd/$storage_fd" "/proc/$BASHPID/fd/$shared_fd" || {
     exec {storage_fd}<&-
     exec {shared_fd}<&-
     return 1
@@ -588,6 +590,7 @@ prepare_caller_mounts() {
   storage_mode=$(stat -Lc '%a' "/proc/$BASHPID/fd/$storage_fd" 2>/dev/null) || storage_mode=""
   shared_mode=$(stat -Lc '%a' "/proc/$BASHPID/fd/$shared_fd" 2>/dev/null) || shared_mode=""
   if [[ $storage_mode != 700 || $shared_mode != 700 ]]; then
+    echo "omarchy-windows-vm: could not secure mount sources to mode 700 (storage=$storage_mode shared=$shared_mode)" >&2
     exec {storage_fd}<&-
     exec {shared_fd}<&-
     return 1
@@ -1015,7 +1018,7 @@ prepare_user_mount_sources() {
     echo "omarchy-windows-vm: storage and shared must be different directories" >&2
     return 1
   }
-  chmod 0700 -- "$storage" "$shared"
+  chmod u=rwx,go=,a-s -- "$storage" "$shared"
 }
 
 storage_space_path() {
@@ -1058,7 +1061,7 @@ write_credentials() {
   local username="$1" password="$2" old_umask dir tmp
   dir=$(dirname -- "$CREDENTIALS_FILE")
   mkdir -p "$dir" || return 1
-  chmod 0700 "$dir" || return 1
+  chmod u=rwx,go=,a-s -- "$dir" || return 1
   old_umask=$(umask)
   umask 077
   tmp=$(mktemp "$dir/.credentials.XXXXXX") || { umask "$old_umask"; return 1; }
--- a/test/shell.d/windows-vm-compose-test.sh
+++ b/test/shell.d/windows-vm-compose-test.sh
@@ -176,6 +176,22 @@ resolve_caller
 find "$CALLER_DATA_ROOT" -name 'rejected-*' -print -quit | grep -q . && fail "source was quarantined"
 pass "invalid second source leaves paths and anchors untouched and leaks no FD"
 
+# A set-ID bit on either source must not survive the privacy step. `chmod 0700`
+# leaves S_ISGID standing on a directory, and the exact mode 700 the elevated
+# path then demands would reject a source it had just been asked to secure.
+reset_case
+mkdir -p "$HOME/.windows" "$HOME/Windows"
+chmod 2700 "$HOME/.windows"
+chmod 6700 "$HOME/Windows"
+prepare_user_mount_sources
+[[ $(stat -Lc '%a' "$HOME/.windows") == 700 ]] || fail "setgid storage source was not cleared to 700"
+[[ $(stat -Lc '%a' "$HOME/Windows") == 700 ]] || fail "setuid/setgid shared source was not cleared to 700"
+chmod 2700 "$HOME/.windows"
+write 4G 2 64G setgid pw UTC
+resolve_caller
+[[ $(stat -Lc '%a' "$EXPECTED_STORAGE") == 700 && $(stat -Lc '%a' "$EXPECTED_SHARED") == 700 ]] || fail "mount leaves are not private after a set-ID source"
+pass "a set-ID bit on either source is cleared instead of failing the mount"
+
 # Distinct caller-owned symlink targets are supported and remain links.
 reset_case
 external_storage="$TMPDIR/external-storage"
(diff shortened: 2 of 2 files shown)


Zhruoshui:
+1 from an affected user — 4.0.3-1, and the analysis here matches what I found independently on my machine.

A few details that might be useful in the thread.

**`chmod g-s` isn't a sufficient workaround**, and it's the one people keep recommending across the duplicate issues. dockur's samba doesn't only set 2700 — on my box the share came back as `drwxrwsrwx` (2777) after a session. `g-s` on that gives you 777, which still trips the `!= 700` check, so the next launch fails again and it looks like the workaround "wore off". `u=rwx,go=,a-s` is the right form precisely because it lands on 700 from either variant. I've seen both 2700 and 2777 on the same install.

**The install path breaks the same way, not just launch.** `prepare_caller_mounts` is called from `__priv_write_compose` as well (line 762), so a first-time `omarchy-windows-vm install` dies at the same guard and only reports "Failed to write the Windows VM configuration." That one is nastier than the launch failure, because `remove_windows()` deletes `~/.windows` and `~/.config/windows` but leaves `~/Windows` alone — the set-ID bit survives a remove, so remove + reinstall is guaranteed to fail. Reinstalling is the first thing people try, which is probably feeding the duplicate count.

**On the numeric-mode behaviour**, for anyone still wondering whether it's the kernel or the filesystem: it's coreutils. 9.11 here, package checksums clean. Against an existing 2700 directory:

```
chmod 0700              -> 2700
chmod 700               -> 2700
install -d -m 0700      -> 2700
chmod 00700             -> 700
chmod u=rwx,go=,a-s     -> 700
python os.chmod(0o700)  -> 700
```

The raw syscall clears it, so it's the four-digits-or-fewer rule for directories — nothing btrfs-specific. Same applies to the other two `chmod 0700` sites this PR converts (1018 and 1061).

Lastly, the silent `return 1` at 590 is doing a lot of the damage on its own. Before I found the guard I had written off docker and polkit, since nothing in the output mentions permissions anywhere. The added echo is arguably worth as much as the chmod fix.

Issue #7374, opened 2026-08-18
Power panel shows UPower's hwdb charge limit (75-80%) instead of the real asusctl-set limit, and a bogus 0 cycle count


chickymonkeys:
### System details

ASUS ROG Zephyrus G14 GA402XV (AMD Ryzen 9 7940HS, NVIDIA RTX 4060 + AMD Radeon 780M), Omarchy 4.0.0-1, kernel 7.1.8-arch1-3, asusctl 6.3.8-1, upower 1.91.3-1. `SUPER+CTRL+P` power panel.

### What's wrong?

Two symptoms in the power panel's battery section, both traced to `omarchy-battery-status` preferring UPower over sysfs:

**1. Charge limit shows 75-80% while the real limit is 60%.**

`asusctl battery limit 60` sets the real limit:

```
$ cat /sys/class/power_supply/BAT0/charge_control_end_threshold
60
$ asusctl battery info # end threshold: 60
```
But UPower reports 75/80 and the widget shows that:
```
$ upower -i /org/freedesktop/UPower/devices/battery_BAT0 | grep threshold
    charge-start-threshold:   75%
    charge-end-threshold:     80%
```
Root cause: UPower imports the **udev hwdb default** `CHARGE_LIMIT=75,80` (from `/usr/lib/udev/hwdb.d/60-upower-battery.hwdb` via `/usr/lib/udev/rules.d/60-upower-battery.rules`), not the value asusctl/TLP actually set. The script reads thresholds from UPower first and only falls back to sysfs when UPower is empty, so the panel shows the hwdb default whenever it disagrees with the real limit. This affects any laptop where a tool sets a limit other than 75/80 (ASUS ROG, ThinkPads with TLP, etc.) — see e.g. Arch BBS #307781 and the systemd docs on `CHARGE_LIMIT`.

**2. "Charge cycles" shows 0, but the EC exposes no real cycle count.**
```
$ cat /sys/class/power_supply/BAT0/cycle_count
0
```
[...]

PR #9612, opened 2026-09-01
omarchy #9586 sysfs battery thresholds

### LINKED ISSUE #9586: omarchy-battery-status shows UPower's fixed ThinkPad default (75/80) instead of the real active charge threshold
## What happened

The battery panel (and `omarchy-battery-status`) shows "Holding at 75-80%" for the charge threshold, but the *actual* active kernel charge threshold on my ThinkPad is 45%-65%.

## What was expected

The panel/status string should reflect the real, active charge threshold.

## Root cause

`omarchy-battery-status` derives the threshold from `upower -i`'s `charge-start-threshold` / `charge-end-threshold` fields first, and only falls back to reading `/sys/class/power_supply/BAT*/charge_control_{start,end}_threshold` directly when those upower fields are empty:

https://github.com/basecamp/omarchy/blob/master/bin/omarchy-battery-status

On ThinkPad-class hardware, UPower's `ChargeStartThreshold`/`ChargeEndThreshold` D-Bus properties are a **fixed built-in default (75/80)** that UPower reports regardless of the real hardware state — they only reflect reality once `ChargeThresholdEnabled` has been toggled on via UPower's own `EnableChargeThreshold` method. Since UPower always returns *a* non-empty value for these fields (even when its own threshold management is inactive), the sysfs fallback in `omarchy-battery-status` never triggers, so real user-set thresholds (e.g. inherited from a prior TLP config, or set directly via sysfs) never show up.

Confirmed live on this machine:
```
$ cat /sys/class/power_supply/BAT0/charge_control_start_threshold
45
$ cat /sys/class/power_supply/BAT0/charge_control_end_threshold
65
$ gdbus introspect --system --dest org.freedesktop.UPower \
    --object-path /org/freedesktop/UPower/devices/battery_BAT0 | grep -i threshold
  readonly u ChargeStartThreshold = 75;
  readonly u ChargeEndThreshold = 80;
  readonly b ChargeThresholdEnabled = false;
  readonly b ChargeThresholdSupported = true;
$ omarchy-battery-status
Battery 64%  ·  Holding at 75-80%  ·  0W / 70Wh
```

The true active threshold (45/65) only shows up if I ignore upower and read sysfs directly — but the script never gets there because upower's fields aren't empty.

## Suggested fix

Prefer the live `/sys/class/power_supply/BAT*/charge_control_{start,end}_threshold` values over upower's fields when they disagree, or at minimum only trust upower's threshold fields when `ChargeThresholdEnabled` is true (readable via the same `upower -i` output as `charge-threshold-enabled` or via a D-Bus properties check).

## Steps to reproduce

1. On a ThinkPad (or other device with sysfs `charge_control_start_threshold`/`charge_control_end_threshold` support) with a custom threshold set directly via sysfs or a tool other than UPower's own `EnableChargeThreshold` (e.g. TLP, or a manual `echo N > charge_control_start_threshold`).
2. Run `omarchy-battery-status` or open the battery panel.
3. Observe it reports 75-80% (or whatever UPower's fixed default is) instead of the real active threshold.

## System details

- Omarchy: 4.0.2-1
- Kernel: 7.1.9-arch1-2
- Machine: Lenovo ThinkPad P15v Gen 1 (20TQ0045AD)
- Hyprland: 0.56.2
- UPower: 1.91.3

Full diagnostics: attached below.

<details>
<summary>omarchy-debug --no-sudo --print</summary>

```
$(cat /home/drligma/.claude/jobs/2bdf6353/tmp/omarchy-debug.log)
```

</details>

kvnloo:
## What

- omarchy-battery-status reads live sysfs charge_control_{start,end}_threshold before UPower.
- Falls back to BAT* glob, then UPower if sysfs is empty.
- test/shell.d/battery-status-test.sh: UPower 75/80 vs sysfs 45/65 reports Holding at 45-65%; no-sysfs still uses UPower; 5 ok.

## Why

Fixes omacom/omarchy#9586. ThinkPad UPower reports a built-in 75/80 while ChargeThresholdEnabled is false, so TLP/sysfs thresholds never appeared.

### DIFF
--- a/bin/omarchy-battery-status
+++ b/bin/omarchy-battery-status
@@ -62,11 +62,27 @@ power_rate=$(awk -v rate="${power_rate_raw:-0}" 'BEGIN {
   print rounded
 }')
 state=$(awk '/state/ { print $2; exit }' <<<"$battery_info")
-threshold_start=$(awk '/charge-start-threshold:/ { gsub(/%/, "", $2); print int($2); exit }' <<<"$battery_info")
-threshold_end=$(awk '/charge-end-threshold:/ { gsub(/%/, "", $2); print int($2); exit }' <<<"$battery_info")
+
+# Kernel sysfs is the live threshold. UPower's charge-*-threshold fields on
+# ThinkPads are a built-in 75/80 default until ChargeThresholdEnabled is set
+# through UPower itself, so they can disagree with TLP or a manual sysfs write.
+read_sysfs_threshold() {
+  local file="$1"
+  [[ -r $file ]] || return 1
+  local value
+  value=$(<"$file")
+  [[ -n $value ]] || return 1
+  printf '%s' "$value"
+}
+
+threshold_start=$(read_sysfs_threshold "$battery_path/charge_control_start_threshold" || true)
+threshold_end=$(read_sysfs_threshold "$battery_path/charge_control_end_threshold" || true)
 
 [[ -z $threshold_end ]] && threshold_end=$(cat "$power_supply_path"/BAT*/charge_control_end_threshold 2>/dev/null | head -1)
 [[ -z $threshold_start ]] && threshold_start=$(cat "$power_supply_path"/BAT*/charge_control_start_threshold 2>/dev/null | head -1)
+
+[[ -z $threshold_end ]] && threshold_end=$(awk '/charge-end-threshold:/ { gsub(/%/, "", $2); print int($2); exit }' <<<"$battery_info")
+[[ -z $threshold_start ]] && threshold_start=$(awk '/charge-start-threshold:/ { gsub(/%/, "", $2); print int($2); exit }' <<<"$battery_info")
 
 ac_online=false
 for supply in "$power_supply_path"/*; do
--- a/bin/omarchy-battery-status
+++ b/bin/omarchy-battery-status
@@ -81,8 +81,13 @@ threshold_end=$(read_sysfs_threshold "$battery_path/charge_control_end_threshold
 [[ -z $threshold_end ]] && threshold_end=$(cat "$power_supply_path"/BAT*/charge_control_end_threshold 2>/dev/null | head -1)
 [[ -z $threshold_start ]] && threshold_start=$(cat "$power_supply_path"/BAT*/charge_control_start_threshold 2>/dev/null | head -1)
 
-[[ -z $threshold_end ]] && threshold_end=$(awk '/charge-end-threshold:/ { gsub(/%/, "", $2); print int($2); exit }' <<<"$battery_info")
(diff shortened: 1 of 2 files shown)


sayem314:
@robosayem review this pr, and check if any dupe pr exist related to this. and list issues this pr will close.

robosayem (commented):
I reproduced the fix locally against a stub upower: with sysfs at 45/65 the script reports `threshold\t45-65%` over UPower's 75/80, without sysfs it falls back to UPower, and the holding line renders correctly. The core ordering (native device file, BAT* glob, UPower last) is right, and the moved fallback makes the previously dead glob code reachable. This fixes the ThinkPad case in #9586.

**One moderate fix requested before merge**

The fallbacks are per-field, which breaks end-only sysfs devices — the ASUS machines from #7374, where only `charge_control_end_threshold` exists. There, end=60 comes from sysfs while `threshold_start` falls back to UPower's fabricated 75, and the renderer takes the `start-end` branch, producing **75-60%** (an inverted range) instead of `60%`. The renderer already has a correct end-only branch, so the fix is to make UPower an all-or-nothing fallback: only read the UPower pair when sysfs exposed *neither* attribute. #7374's own proposed fix says the same ("don't fabricate a start threshold without a sysfs start file"). The PR's test only covers both-present and both-absent — a mixed case (end sysfs present, start absent) would have caught this.

**Nit, non-blocking**

Some drivers report `charge_control_end_threshold = 0` to mean disabled — that renders as `0-0%`. Treating 0 as absent (fall through to UPower) would be safer, but it can also be left to a follow-up.

**Duplicate check**

No duplicate: this is the only open PR changing the threshold source. Neighbouring open PRs touch other battery dimensions — #7915/#7895 (pending-charge hold heuristic), #7293 (charge *status* source), #6845/#7138/#7186/#8864/#9106 (multi-battery), #10076 (percentage source) — all complementary.

**What this closes**

- Fully: #9586.
- Partially: #7374 (fixes the 75-80-vs-real-limit symptom, not the bogus 0 cycle-count half) and #10344 (fixes the UPower-first source half, not the `Model.chargeThresholdActive` panel heuristic). Neither should auto-close on this PR alone unless the maintainers want to split them.

With the mixed-source fix, this is a clean approve from me.

kvnloo:
Addressed @robosayem's mixed-source fix on `0c9ad0b`:

- UPower thresholds are now an all-or-nothing fallback (only when sysfs exposed neither start nor end), so end-only ASUS sysfs (`charge_control_end_threshold=60`) renders `60%` instead of an inverted `75-60%`.
- Added a regression test for that end-only + UPower-start-present case.

Left for follow-up (non-blocking nit): treating sysfs `0` as absent / disabled.

kvnloo:

Scoped: battery-status-test (sysfs-first).

PR #9614, opened 2026-09-01
omarchy #9544 launch-or-focus agent titles (fork PR)

### LINKED ISSUE #9544: omarchy-launch-or-focus / omarchy-launch-spotify match window titles, so any window with the app name in its title hijacks the keybinding
## What happened

`SUPER+SHIFT+M` (Music) appeared to do nothing — Spotify never launched. Instead, focus silently jumped to an unrelated window on another workspace.

Root cause: `bin/omarchy-launch-spotify` looks for an existing Spotify window by matching the pattern against **both `.class` and `.title`** of every Hyprland client:

```bash
WINDOW_ADDRESS=$(hyprctl clients -j | jq -r '.[]|select((.class|test("\\bspotify\\b";"i")) or (.title|test("\\bspotify\\b";"i")))|.address' | head -n1)
```

Any window whose *title* merely contains the word "spotify" is treated as the Spotify app and gets focused instead of launching Spotify. In my case it was an Omarchy agent window (`class: org.omarchy.agent`) — a Claude Code session titled "Spotify player stoppt bei Fensterverschiebung". So Omarchy's own agent windows can hijack the binding. A browser window with "Spotify" in the page title does the same.

The generic `bin/omarchy-launch-or-focus` has the same class-or-title matcher, so every `omarchy launch or focus <pattern>` consumer is affected the same way (e.g. a webapp binding labelled "Calendar" being swallowed by an agent window whose task title contains "calendar" — I've hit that variant too).

## What I expected

`SUPER+SHIFT+M` launches Spotify (or focuses the actual Spotify app window if it is already running).

## Steps to reproduce

1. Make sure Spotify is not running.
2. Open any window whose title contains the word "Spotify" — e.g. start an Omarchy/Claude agent session about Spotify, or open a browser tab with "Spotify" in the title.
3. Press `SUPER+SHIFT+M`.
4. Focus jumps to that window; Spotify never launches.

## Suggested fix

For app launchers like `omarchy-launch-spotify`, match on `.class` (or `.initialClass`) only — the real app is reliably identified by class, and titles are user/content-controlled:

```bash
WINDOW_ADDRESS=$(hyprctl clients -j | jq -r '.[]|select(.class|test("\\bspotify\\b";"i"))|.address' | head -n1)
```

For the generic `omarchy-launch-or-focus`, where title matching may be intentional (TUIs/webapps), consider at least excluding `org.omarchy.agent` windows from the match so Omarchy's own agent sessions can't swallow app bindings.

## System

- Omarchy 4.0.1-1, Arch Linux, Hyprland (three 2560x1440 monitors, scale 1)
- Verified against current `master`: both `bin/omarchy-launch-spotify` and `bin/omarchy-launch-or-focus` contain the class-or-title matcher.
- Local workaround (class-only copy of the launcher + rebind) resolves it.

kvnloo:
## What

- omarchy-launch-or-focus skips windows with class org.omarchy.agent so agent titles cannot swallow app bindings.
- omarchy-launch-spotify matches window class only (not title).
- test/shell.d/launch-or-focus-test.sh stubs hyprctl: agent title no longer focuses; Spotify class still does; non-agent titles still match; 7 ok.

## Why

Fixes omacom/omarchy#9544. SUPER+SHIFT+M focused an Omarchy agent (or browser tab) whose title contained “Spotify” instead of launching Spotify.

### DIFF
--- a/bin/omarchy-launch-or-focus
+++ b/bin/omarchy-launch-or-focus
@@ -10,7 +10,7 @@ fi
 
 WINDOW_PATTERN="$1"
 LAUNCH_COMMAND="${2:-"uwsm-app -- $WINDOW_PATTERN"}"
-WINDOW_ADDRESS=$(hyprctl clients -j | jq -r --arg p "$WINDOW_PATTERN" '.[]|select((.class|test("\\b" + $p + "\\b";"i")) or (.title|test("\\b" + $p + "\\b";"i")))|.address' | head -n1)
+WINDOW_ADDRESS=$(hyprctl clients -j | jq -r --arg p "$WINDOW_PATTERN" '.[]|select(.class != "org.omarchy.agent")|select((.class|test("\\b" + $p + "\\b";"i")) or (.title|test("\\b" + $p + "\\b";"i")))|.address' | head -n1)
 
 if [[ -n $WINDOW_ADDRESS ]]; then
   hyprctl dispatch "hl.dsp.focus({ window = \"address:$WINDOW_ADDRESS\" })" >/dev/null 2>&1 || hyprctl dispatch focuswindow "address:$WINDOW_ADDRESS"
--- a/bin/omarchy-launch-spotify
+++ b/bin/omarchy-launch-spotify
@@ -4,7 +4,7 @@
 
 set -e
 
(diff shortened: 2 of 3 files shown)


kvnloo:

Scoped: launch-or-focus-test.

PR #9620, opened 2026-09-01
Make launch-or-focus execute argv directly

patroza:
## Problem

On `quattro`, `omarchy-launch-or-focus` accepts its launch command as one string and reparses it with `eval`. The web-app and TUI wrappers flatten their argument arrays to fit that interface:

```bash
LAUNCH_COMMAND="omarchy-launch-webapp $@"
```

That loses the original argument boundaries. For example, `--profile-directory=Profile 2` becomes `--profile-directory=Profile` plus a separate `2`, so Chromium opens the `Profile` profile instead of `Profile 2`. Shell metacharacters can also be interpreted instead of forwarded literally.

## Change

Make `omarchy-launch-or-focus` accept a window pattern followed by a command and its arguments. It stores the command as a Bash array and executes it directly, without `eval`.

All tracked callers are migrated in the same change:

- The web-app and TUI wrappers forward their arguments directly.
- The Hyprland `launch_sole` helper uses `bash -c` explicitly because it intentionally constructs shell source.
- The extension-menu TUI example passes `zsh`, `-c`, and the script as separate arguments.

The default launch behavior remains unchanged for ordinary single-name patterns.

## Compatibility

This intentionally breaks custom calls that pass an entire launch command as one argument:

```bash
omarchy-launch-or-focus pattern "command arg"
```

They must pass discrete arguments:

```bash
omarchy-launch-or-focus pattern command arg
```

Callers that intentionally need expansion, pipelines, redirections, or other shell syntax must request a shell explicitly:

```bash
omarchy-launch-or-focus pattern bash -c 'command "$VALUE" | other >output'
```

A multiword window pattern can also no longer double as the default launch command; provide the launch command separately.

## Existing backward-compatible alternatives

[#8171](https://github.com/omacom/omarchy/pull/8171) and [#9490](https://github.com/omacom/omarchy/pull/9490) preserve the current public interface. Both serialize each wrapper argument with `printf %q` before the existing `eval` reparses the command string.

Those PRs are the smallest backward-compatible fixes and also cover both web-app and TUI forwarding. They solve the immediate correctness and injection problems, but retain the command-string serialization and `eval` architecture.

## Backward-compatible alternative

Both interfaces could coexist by keeping the existing second argument as a command string and using `--` to select direct argv:

```bash
omarchy-launch-or-focus pattern "legacy command string"
omarchy-launch-or-focus pattern -- command args...
```

Applied on top of this PR, the compatibility layer would be:

```diff
diff --git a/bin/omarchy-launch-or-focus b/bin/omarchy-launch-or-focus
@@
-# omarchy:args=<window-pattern> [<command> [args...]]
+# omarchy:args=<window-pattern> [<launch-command> | -- <command> [args...]]
@@
-if (( $# > 0 )); then
+if [[ ${1:-} == "--" ]]; then
+  shift
+  if (( $# == 0 )); then
+    echo "omarchy-launch-or-focus: -- requires a command" >&2
+    exit 1
+  fi
   LAUNCH_ARGV=("$@")
+  LAUNCH_COMMAND=""
 else
-  LAUNCH_ARGV=(uwsm-app -- "$WINDOW_PATTERN")
+  LAUNCH_ARGV=()
+  LAUNCH_COMMAND="${1:-"uwsm-app -- $WINDOW_PATTERN"}"
 fi
@@
-  exec setsid "${LAUNCH_ARGV[@]}"
+  if (( ${#LAUNCH_ARGV[@]} > 0 )); then
+    exec setsid "${LAUNCH_ARGV[@]}"
+  else
+    eval exec setsid $LAUNCH_COMMAND
+  fi
 fi
diff --git a/bin/omarchy-launch-or-focus-webapp b/bin/omarchy-launch-or-focus-webapp
@@
-exec omarchy-launch-or-focus "$WINDOW_PATTERN" omarchy-launch-webapp "$@"
+exec omarchy-launch-or-focus "$WINDOW_PATTERN" -- omarchy-launch-webapp "$@"
diff --git a/bin/omarchy-launch-or-focus-tui b/bin/omarchy-launch-or-focus-tui
@@
-exec omarchy-launch-or-focus "$APP_ID" omarchy-launch-tui "$@"
+exec omarchy-launch-or-focus "$APP_ID" -- omarchy-launch-tui "$@"
diff --git a/default/hypr/helpers.lua b/default/hypr/helpers.lua
@@
-  return "omarchy-launch-or-focus " .. shell_quote(match) .. " bash -c " .. shell_quote(o.launch(command))
+  return "omarchy-launch-or-focus " .. shell_quote(match) .. " -- bash -c " .. shell_quote(o.launch(command))
```

This preserves external command-string callers while giving migrated callers exact argv semantics. Its cost is retaining `eval` and maintaining two interfaces.

## Tests

The focused shell test covers:

- web-app and TUI argument forwarding;
- arguments containing spaces and literal glob characters;
- explicit shell evaluation from the Hyprland helper;
- the default launch command.

All focused checks pass.

## Related upstream work

- [#1977](https://github.com/omacom/omarchy/issues/1977) reported missing browser-profile argument forwarding; [#1980](https://github.com/omacom/omarchy/pull/1980) introduced the current string flattening while addressing it.
- [#3390](https://github.com/omacom/omarchy/pull/3390) reported the same failure for a profile name containing a space.

### DIFF
--- a/bin/omarchy-launch-or-focus
+++ b/bin/omarchy-launch-or-focus
@@ -1,36 +1,25 @@
 #!/bin/bash
 
 # omarchy:summary=Launch an app or focus an existing window matching a pattern
-# omarchy:args=<window-pattern> [<launch-command> | -- <command> [args...]]
+# omarchy:args=<window-pattern> [<command> [args...]]
 
 if (($# == 0)); then
-  echo "Usage: omarchy-launch-or-focus <window-pattern> [<launch-command> | -- <command> [args...]]"
+  echo "Usage: omarchy-launch-or-focus <window-pattern> [<command> [args...]]"
   exit 1
 fi
 
 WINDOW_PATTERN="$1"
 shift
 
-if [[ ${1:-} == "--" ]]; then
-  shift
-  if (( $# == 0 )); then
-    echo "omarchy-launch-or-focus: -- requires a command" >&2
-    exit 1
-  fi
+if (( $# > 0 )); then
   LAUNCH_ARGV=("$@")
-  LAUNCH_COMMAND=""
 else
-  LAUNCH_ARGV=()
-  LAUNCH_COMMAND="${1:-"uwsm-app -- $WINDOW_PATTERN"}"
+  LAUNCH_ARGV=(uwsm-app -- "$WINDOW_PATTERN")
 fi
 WINDOW_ADDRESS=$(hyprctl clients -j | jq -r --arg p "$WINDOW_PATTERN" '.[]|select((.class|test("\\b" + $p + "\\b";"i")) or (.title|test("\\b" + $p + "\\b";"i")))|.address' | head -n1)
 
 if [[ -n $WINDOW_ADDRESS ]]; then
   hyprctl dispatch "hl.dsp.focus({ window = \"address:$WINDOW_ADDRESS\" })" >/dev/null 2>&1 || hyprctl dispatch focuswindow "address:$WINDOW_ADDRESS"
 else
-  if (( ${#LAUNCH_ARGV[@]} > 0 )); then
-    exec setsid "${LAUNCH_ARGV[@]}"
-  else
-    eval exec setsid $LAUNCH_COMMAND
-  fi
+  exec setsid "${LAUNCH_ARGV[@]}"
 fi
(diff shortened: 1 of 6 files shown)


bjarneo:
Automated duplication check: this pull request looks similar to #9490, which covers the same launch-or-focus arg quoting. I keep that one open and close this one to consolidate review.

PR #9625, opened 2026-09-01
Add the Perplexity desktop app to Install > AI

spencerbull:
Adds Perplexity to the desktop AI apps, following the T3 Code / Grok Bot flow: the _Install > AI_ entry runs `omarchy-install-and-launch`, so picking it installs the `perplexity` package on demand and launches the app once the install finishes; after that the entry greys out and a _Remove > AI_ entry appears.

`omarchy-remove-ai-perplexity` drops the package and always removes the app's own runtime: the Electron caches and the rpc-server downloads (llama.cpp builds and local models) under `~/.cache` and `~/.local/share`. What the user made with the app — the logins and session in `~/.config/Perplexity`, the secret vault and device identity in `~/.local/state/perplexity`, and the launcher flags — is kept by default and deleted only on an explicit yes at a terminal, default no, the same choice Remove OpenClaw offers for `~/.openclaw`. The prompt requires stderr on the terminal too, since that is where gum draws it; redirected stderr means the data stays rather than the removal blocking on an invisible question. Other `perplexity-*` cache dirs belong to Perplexity products that outlive the desktop app and always survive. Like the Hermes remover, it sets `-u` so an unset `HOME` is a refusal rather than a set of `rm -rf` paths rooted at `/`.

The menu mark is a new U+E90B glyph in the Omarchy icon font, from Simple Icons' monochrome Perplexity mark. The font is package-owned, so the mark reaches desktops with the next omarchy-settings release; until then the entry renders without an icon.

Sits on omacom/omarchy-pkgs#250, which packages the app itself.

🤖 Generated by Fable 5 in Claude Code. Reviewed by Fable 5, Codex XHigh, and Sonnet 5.


spencerbull:
Install View:
image

Remove View:
image

App View:
image

spencerbull:
Reviewed the diff against `quattro`, ran `test/shell` (`remove-ai-test.sh`, `menu-test.sh`) and `test/cli` in a worktree at e32745b0 — all green. Rendered the new U+E90B glyph standalone and it comes out as a clean asterisk silhouette; the Install/Remove screenshots above already show it live in the running menu, correctly colored, so that satisfies `agents/skills/icon-font.md`'s verification step. Cross-checked `install.ai.perplexity`'s desktop id and the `omarchy-pkg-present perplexity` guards against the packaging PR (omacom/omarchy-pkgs#250): the package name, `/usr/share/applications/perplexity.desktop`, and the `perplexity-flags.conf` path all line up.

Codex at xhigh reasoning found two real gaps in `omarchy-remove-ai-perplexity`, both inherited rather than introduced here. First, `omarchy-pkg-drop perplexity` runs before the script ever touches `$HOME`, so with `HOME` unset the package still gets dropped — verified by reproducing it — before the `rm -rf` line dies on the unbound variable; the `-u` guard stops the root-rooted delete, not the removal itself, so "refuses to run" overstates what actually happens. Second, the flags-file cleanup deletes `$HOME/.config/perplexity-flags.conf` unconditionally, but the launcher shipped in omarchy-pkgs#250 resolves that path through `${XDG_CONFIG_HOME:-$HOME/.config}`, so a user with `XDG_CONFIG_HOME` set elsewhere keeps a stale flags file after removal. Both gaps already exist unchanged in `omarchy-remove-ai-hermes` and `omarchy-remove-ai-t3-code` respectively — this PR copied the established pattern rather than creating a new one, so fixing only the Perplexity copy would leave the same behavior in the removers it was modeled on.

Nothing here blocks this PR. The two gaps are pre-existing and repo-wide; worth a follow-up that fixes the drop-before-HOME-check ordering and the `XDG_CONFIG_HOME` handling across all the AI-app removers at once rather than patching one copy at a time.

🤖 Generated by Sonnet 5 in Claude Code. Reviewed by Codex XHigh.

spencerbull:
Reviewed the new commit (919fd491, "Ask, default no, before Remove Perplexity deletes the user's data") against the previous head. Ran the full `./test/all`, not just the changed file — `test/shell.d/bin-style-test.sh` currently fails against this commit:

```
not ok - bin commands use command helpers
/tmp/omarchy-pr9625-review/bin/omarchy-remove-ai-perplexity
```

That test forbids raw `command -v` in `bin/` scripts outside the `omarchy-cmd-*`/`omarchy-pkg-*` helpers themselves, and the new `command -v gum >/dev/null` guard trips it. Confirmed this is a regression from this commit specifically: the prior commit (e32745b0) had no `command -v` calls and the test passed. `gum` is also in `install/omarchy-base.packages`, making it a runtime invariant under AGENTS.md's Helper Commands section ("do not add defensive... checks around them... use command-presence helpers only for genuinely optional dependencies"), and no other `gum confirm` call site in the repo (`omarchy-remove-preinstalls`, `omarchy-update-confirm`, `omarchy-system-factory-reset`) guards it this way — so the fix is to drop the check rather than swap it for `omarchy-cmd-present gum`. The other three files `test/all` reported failing (`config-test.sh`, `snapper-test.sh`, `unowned-system-paths-test.sh`) are unrelated — they need an `omarchy-pkgs` checkout at `OMARCHY_PKGS_PATH` that isn't present in this environment, and fail the same way on the base branch.

Codex at xhigh reasoning also flagged, and I verified directly: `[[ -t 0 ]]` alone doesn't establish that the confirm prompt is visible. gum's TUI renders on stderr, not stdout — confirmed by running `gum confirm` under `script` with stderr redirected to a file: the escape sequences land entirely in that file, nothing reaches the pty. So `omarchy-remove-ai-perplexity 2>some.log` (or anything else that redirects stderr away from the terminal while stdin stays attached) drops the package and always-owned caches, then blocks on a confirm the user can't see, with no visible cue to type y/n. The menu path is unaffected — `omarchy-launch-floating-terminal-with-presentation` doesn't redirect any of the child's stdio — but the command is also runnable directly, and this is where it bites. Worth checking `-t 2` alongside `-t 0`, since that's where gum actually writes.

One more, from reading the PR body against the code: it describes the ask-before-delete choice as "the same choice the Hermes remover offers." Checked `omarchy-remove-ai-hermes` directly — it has no `gum confirm` and no prompt at all; it unconditionally keeps `~/.hermes`. The comparison doesn't match what Hermes actually does.

Also checked, not a blocker: `omarchy-pkg-drop` is a plain `pacman -Rns` and doesn't stop a running instance first. If Perplexity or its rpc-server child is still open when this runs, the always-deleted cache dirs could have open handles at delete time — a lifecycle gap (app still running during removal), not something reordering the existing two steps would fix. Confirmed separately that `gum confirm --default=false` only enters the deletion branch on an explicit "Yes": bare Enter, Ctrl-C, SIGHUP, and closed input all resolve to "No" against the installed gum binary, so the default-no design itself is sound.

The HOME-unset-runs-pkg-drop-first and `XDG_CONFIG_HOME`-blind flags-file gaps from the last review are unchanged here — re-verified by reproduction — and remain pre-existing patterns shared with `omarchy-remove-ai-hermes`/`omarchy-remove-ai-t3-code`, not new.

The failing `bin-style-test.sh` is the one item here that blocks merge as-is; everything else is either already-covered pre-existing behavior or worth a judgment call rather than a fix.

🤖 Generated by Sonnet 5 in Claude Code. Reviewed by Codex XHigh.

spencerbull:
Reviewed the latest commit (eea678b4). Both issues from the previous pass are fixed and I re-verified each directly rather than trusting the diff alone.

`test/shell.d/bin-style-test.sh` now passes for `omarchy-remove-ai-perplexity`: the `command -v gum` guard is gone, replaced with `[[ -t 0 && -t 2 ]]`. Ran the full `./test/all` — the same 4 files fail as on `quattro` itself (`config-test.sh`, `snapper-test.sh`, `unowned-system-paths-test.sh` need `OMARCHY_PKGS_PATH`, which isn't set in this environment; `bin-style-test.sh` now fails on a *different*, already-merged file, `bin/omarchy-remove-ai-openclaw`, which has the identical `command -v gum` guard this PR just removed — pre-existing on `quattro`, not part of this diff, not something this PR needs to fix, but the same mistake independently in two AI-app removers is worth a follow-up across both once this lands).

Reproduced the original stderr-visibility bug directly against the new code to confirm the fix: `script -qec 'omarchy-remove-ai-perplexity 2>/dev/null'` on a real pty now completes without blocking and leaves `~/.config/Perplexity` untouched, where before it would drop the package/caches and then hang on a confirm prompt nobody could see. The new regression test for this (`remove-ai-test.sh`) passes and matches.

The PR body's design comparison is also fixed — it now says this mirrors "the same choice Remove OpenClaw offers for `~/.openclaw`," and I checked `omarchy-remove-ai-openclaw` directly: it does have the identical `gum confirm --default=false` pattern, so the comparison holds this time (the earlier Hermes comparison did not).

Codex at xhigh reasoning found no defect in the remover itself this round, only a test-coverage gap: none of the current tests independently pin `-t 0` and `-t 2` against each other — the "no terminal" case runs with stdin redirected but doesn't control whether the test process's own stderr happens to be a tty, and the "answer is no" cases run entirely inside a single `script` pty where all three streams move together. So a regression that dropped either half of `-t 0 && -t 2` (or swapped in `-t 1`) could pass the existing suite depending on incidental environment properties rather than being reliably caught. Worth a case that pins stdin to `/dev/null` while forcing stderr onto a pty (and the reverse), but this is a robustness improvement, not something blocking here.

The two pre-existing gaps from the first review (`omarchy-pkg-drop` running before `$HOME` is referenced, and the flags-file cleanup not honoring `XDG_CONFIG_HOME`) are unchanged, as expected — still shared with `omarchy-remove-ai-hermes`/`omarchy-remove-ai-t3-code`, not unique to this file.

Nothing here blocks this PR.

🤖 Generated by Sonnet 5 in Claude Code. Reviewed by Codex XHigh.

PR #9627, opened 2026-09-01
Merge quattro into hermes-agent

spencerbull:
Brings `hermes-agent` up to date with `quattro` (50 commits behind before this).

Beyond the routine sync, this picks up the current versions of two shell tests that fail on installed machines in the branch's pre-merge copies:

- `test/shell.d/windows-vm-mount-boundary-test.sh` — quattro's version copies the helper to a mount-safe path before the tmpfs mounts hide `/home`; the old copy dies silently when the checkout lives under `/home`.
- `test/shell.d/theme-install-guards-test.sh` — quattro's version shadows the packaged `omarchy-git-url-check` with a command-not-found stub instead of thinning PATH, so the missing-checker assertion holds on machines where the checker is installed in `/usr/bin`.

Merged clean, no conflicts. Full `./test/all` passes locally on this merge (225/225 test files, with `omarchy-pkgs` and `omarchy-iso` checkouts present for the packaging/installer coverage).


spencerbull:
Superseded: #7469 merged into quattro today (basecamp/omarchy@b71dcad9), which resolves the sync this PR existed for. Nothing here is missing from quattro.

PR #9629, opened 2026-09-01
Hide notifications during screensaver

### LINKED ISSUE #9623: [Quattro] Notifications render above the fullscreen screensaver
### System details

Omarchy 4.0.1 Quattro, Hyprland 0.56.2, Quickshell 0.3.1, NVIDIA RTX 3080 Ti, and two DisplayPort outputs.

### Problem

Notification toasts rendered above the fullscreen animated screensaver, exposing notification contents on a screen intended to conceal the session.

Pending Omarchy Migrations notification rendered over the animated screensaver

### Cause

The terminal screensaver is a normal fullscreen Wayland client. Notification popup windows are `WlrLayer.Overlay` layer-shell surfaces, so the compositor draws them above normal clients unless the notification service explicitly unmaps them.

### Verified solution

PR #9629 hides notification popup surfaces while the enabled idle service reports a live screensaver window and pauses popup lifetime timers while hidden. The idle service is resolved through the plugin registry, including user-cloned implementations.

On the dual-output hardware test, a persistent critical notification was visible before the screensaver, absent over both animations, and restored after dismissal with its remaining lifetime intact. Compositor inspection reported zero `omarchy-notifications` layer surfaces while both screensaver windows were mapped.

ClGratton:
## What changed

Notification popup surfaces are now hidden whenever the enabled idle service reports a live screensaver window. Popup countdowns pause while hidden, so a notification returns with its remaining visible lifetime after the screensaver is dismissed.

The idle service is resolved through the plugin registry, which keeps this working when someone has cloned `omarchy.idle` for customization.

Fixes #9623.

## Why

The fullscreen terminal screensaver is a normal Wayland client, while notification popups use the overlay layer. The compositor therefore draws notifications above the screensaver unless the notification service explicitly unmaps them. That can expose private notification contents on a screen intended to conceal the session.

## Testing

The notification and notification-send focused suites pass. On a two-DisplayPort system, a persistent critical notification was visible before the test, absent from both animated screensaver outputs, and restored after mouse input dismissed the screensaver. Compositor inspection also reported zero `omarchy-notifications` layer surfaces while both screensaver windows were mapped.

### DIFF
--- a/shell/plugins/notifications/Service.qml
+++ b/shell/plugins/notifications/Service.qml
@@ -17,6 +17,16 @@ Item {
   // Injected by omarchy-shell (the first-party service loader).
   property var shell: null
 
+  // Notification popups are layer-shell overlays, while the fullscreen
+  // screensaver is a normal Wayland client. Hide the overlays whenever the
+  // enabled idle service reports a screensaver window. Resolve through the
+  // registry so this also follows a user-cloned idle service.
+  readonly property string idleServiceId: shell && shell.pluginRegistry
+    ? shell.pluginRegistry.resolveEnabledId("omarchy.idle")
+    : "omarchy.idle"
+  readonly property var idleService: shell ? shell.serviceFor(idleServiceId) : null
+  readonly property bool screensaverActive: !!idleService && idleService.screensaverWindowCount > 0
+
   property string omarchyPath: Quickshell.env("OMARCHY_PATH")
   readonly property string home: Quickshell.env("HOME")
   // History + DND live under XDG_STATE_HOME: they're persistent user state
@@ -956,7 +966,7 @@ Item {
       id: popupWindow
       required property var modelData
       screen: modelData
-      visible: popupModel.count > 0
+      visible: popupModel.count > 0 && !service.screensaverActive
 
       WlrLayershell.namespace: "omarchy-notifications"
       WlrLayershell.layer: WlrLayer.Overlay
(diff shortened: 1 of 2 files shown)


dadofsambonzuki:
Reproduced the underlying issue on 4.0.4 and this looks right for the screensaver case — the surface hide plus the paused lifetime is the behaviour I'd want.

Same layer-shell ordering bites on ordinary fullscreen content, which the idle service can't see: a toast lands on top of a fullscreen video or game whenever DND isn't already on. Two notes in case they're useful, and no expectation that they belong in this PR:

- The hide-and-retain-lifetime shape here covers more than a DND-style fix would. Skipping the toast only handles notifications that respect DND; hiding the surface also handles the senders that deliberately bypass it (`omarchy-action` confirmations, critical `notify-send`).
- `Quickshell.Wayland.Toplevel.fullscreen` is a reactive per-window property and it is true for both client-requested fullscreen (`mpv --fullscreen`) and compositor-dispatched fullscreen (the `fullscreen` dispatcher on a foot window), false again when windowed — verified on 4.0.4 / Hyprland 0.56.2 / Quickshell 0.3.1. `Toplevel.screens` gives visibility, `Hyprland.toplevels` enumerates. So a `fullscreenActive` sibling to `screensaverActive` — `any over Hyprland.toplevels where wayland.fullscreen && wayland.screens.length > 0` — would extend these same `visible:` and `running:` guards to fullscreen apps with no new IPC and no polling.

Happy to write that as a follow-up commit on your branch, or as a separate PR if you'd rather keep this one narrow. Say which and I'll do it.

(I've written the wider ask up as a suggestion: https://github.com/omacom/omarchy/discussions/12951)

Issue #7559, opened 2026-08-20
[Bug] Fcitx5/Mozc candidate window is misplaced in native Wayland Chromium and Electron apps


simosako:
### System details

ThinkPad X1 Carbon Gen 12 / Intel Core Ultra 5 125U 

## Summary

When using Fcitx5 with Mozc on Omarchy Quattro, the Japanese IME candidate window is displayed at an incorrect position in native Wayland Chromium/Electron applications. The candidate window does not follow the text caret.

The issue is reproducible in both Chromium and Obsidian. It is not reproducible when either application is launched through XWayland.

Japanese input and conversion otherwise work correctly.

## Environment

- Omarchy: `4.0.0-1`
- Hardware: ThinkPad X1 Carbon Gen 12
- CPU: Intel Core Ultra 5 125U
- GPU: Intel Meteor Lake, `i915`
- Kernel: `7.1.8-arch1-3`
- Hyprland: `0.56.2-1`
- Hyprland commit: `efb50993780079460b0cbed1363e2166a2de1d9f`
- XWayland: `24.1.13`
- Session: Wayland
- Display: `1920x1200`, scale `1`
- Locale: `ja_JP.UTF-8`
- Keyboard layout: `jp106`
- Fcitx5: `5.1.21-1`
- Fcitx5 Qt: `5.1.14-1`
- Fcitx5 GTK: `5.1.7-1`
- Mozc: `3.34.6239.2-1`
- Chromium: `151.0.7922.137-1`
- Obsidian: `1.13.7-1`

Fcitx5 is managed by the stock `omarchy-fcitx5.service` and is running normally.

## Steps to reproduce

1. Start an Omarchy Quattro Wayland session.
2. Configure Fcitx5 with `keyboard-us` and Mozc.
3. Activate Mozc.
4. Launch Chromium using the standard Omarchy configuration.
5. Focus a text field, such as the Yahoo search field.
6. Type Japanese input and start conversion.
7. Move the caret or continue composing at a different position in the field.

The same behavior can be reproduced in Obsidian.

## Expected behavior

The candidate window should appear next to, or otherwise track, the text caret.

## Actual behavior

The candidate window appears at an incorrect position and does not follow the text caret. In the attached Chromium recording, the candidate list is rendered at an incorrect, relatively fixed position within the search field while Mozc conversion itself continues to work.

## A/B comparison

- Chromium running as native Wayland: **reproduces**
- Obsidian running as native Wayland: **reproduces**
[...]

PR #9634, opened 2026-09-01
Add one-step Japanese input setup

komagata:
## Summary

- add _Setup > Input Method > Mozc (Japanese)_ to the Omarchy menu
- install `fcitx5-mozc`, restart Fcitx 5, and append Mozc to the current input-method group
- document the setup and add shell plus desktop acceptance coverage

## Why

Japanese input is not optional for Japanese-speaking users. Windows and macOS finish setup with a working Japanese IME, while Omarchy currently requires users to discover the package and configure Fcitx 5 manually.

This keeps the first change deliberately small: it adds a one-step setup action rather than changing the main installer.

## Scope

This PR intentionally treats input-method setup separately from locale and UI internationalization. These concerns may be connected during initial setup, but users may need Japanese input regardless of their system locale or UI language.

Keeping this change focused also makes it possible to improve the input experience incrementally without first deciding the broader locale and i18n architecture.

Naming the menu item after the input engine rather than equating Japanese input with Mozc also leaves room to add options such as _Hazkey (Japanese)_ or _Rime (Chinese)_ alongside it in the future.

## Testing

- `bash test/shell.d/input-japanese-test.sh`
- `bash test/shell.d/menu-guards-test.sh`
- `./test/cli`
- QEMU/KVM desktop acceptance: installed Mozc, registered it with Fcitx 5, and converted `nihongo` to `日本語` in a real terminal

The broader VM suite still has unrelated existing environment/OCR failures; the new Japanese input acceptance test passes end to end.

## Related PRs

- #11719 adds one-step Korean input setup with Hangul.
- #12312 adds one-step Traditional Chinese input setup with Chewing.

These PRs share the Input Method parent menu and the Fcitx 5 registration flow. This PR remains focused on Mozc setup; shared setup logic can be extracted in a follow-up. If a sibling PR lands first, I will adjust the shared menu entry and documentation accordingly.

### DIFF
--- a/bin/omarchy-setup-input-mozc
+++ b/bin/omarchy-setup-input-mozc
@@ -1,0 +1,56 @@
+#!/bin/bash
+
+# omarchy:summary=Set up Mozc Japanese input with Fcitx 5
+
+set -euo pipefail
+
+omarchy-pkg-add fcitx5-mozc
+omarchy-restart-xcompose
+
+fcitx_ready=false
+for _ in {1..50}; do
+  if fcitx5-remote --check; then
+    fcitx_ready=true
+    break
+  fi
+  sleep 0.1
+done
+
+if [[ $fcitx_ready == false ]]; then
+  echo "Fcitx 5 is not running. Log in to an Omarchy desktop session and try again." >&2
+  exit 1
+fi
+
+group=$(fcitx5-remote -q)
+group_info=$(busctl --user --json=short call \
+  org.fcitx.Fcitx5 \
+  /controller \
+  org.fcitx.Fcitx.Controller1 \
+  InputMethodGroupInfo \
+  s "$group")
+
+if ! jq -e '.data[1] | any(.[0] == "mozc")' <<<"$group_info" >/dev/null; then
+  layout=$(jq -r '.data[0]' <<<"$group_info")
+  mapfile -t input_methods < <(jq -r '.data[1][] | @tsv' <<<"$group_info")
+
+  args=("$group" "$layout" "$((${#input_methods[@]} + 1))")
+  for input_method in "${input_methods[@]}"; do
+    IFS=$'\t' read -r name item_layout <<<"$input_method"
+    args+=("$name" "$item_layout")
+  done
+  args+=(mozc "")
+
+  busctl --user call \
+    org.fcitx.Fcitx5 \
+    /controller \
+    org.fcitx.Fcitx.Controller1 \
+    SetInputMethodGroupInfo \
+    'ssa(ss)' "${args[@]}" >/dev/null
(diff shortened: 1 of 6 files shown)


akitaonrails:
Ran this on Omarchy 4.0.4 (fcitx5 5.1.22, Hyprland 0.56.2) against an install that already had Mozc registered: idempotent, exit 0, no group rewrite. The D-Bus registration is the right call — editing `profile` directly loses the change, because fcitx5 rewrites that file from live state when it exits.

Two small things from using it, plus one gap that bites right after it succeeds.

**`fcitx5-remote --check` prints the input state to stdout.** It writes a digit (`1` here), so the readiness loop leaks it into the presentation terminal before the success message. `>/dev/null` covers it:

```bash
if fcitx5-remote --check >/dev/null 2>&1; then
```

#11719 hit the same thing in the Korean sibling.

**`fcitx5-configtool` is hidden from the launcher.** The manual edit still points other-language users at it:

> install an input engine like `fcitx5-chinese-addons` with `omarchy pkg add`, plus `fcitx5-configtool` to add the engine to your input methods

`default/omarchy/launcher.hides:11` lists `fcitx5-configtool`, so it never appears in the launcher after installing. Worth saying "run `fcitx5-configtool` from a terminal" so people do not go looking for it — that was a few confused minutes here.

**Chromium ignores the IME unless `--enable-wayland-ime` is set.** This is outside your scope, but it lands immediately after your script succeeds: the engine is installed and registered, every GTK/Qt app and terminal accepts Japanese, and the default browser silently does not. I opened #12139 to add the flag to `config/chromium-flags.conf` with a migration for existing installs, since the per-browser flags files are only copied at install time.

Unrelated to the setup path, for anyone who finds this thread: once input works in Chromium, the candidate window is misplaced there (#7559), and the popup is invisible in Hyprland true fullscreen (#11303).

For what it is worth, the shape of this PR — one menu action, engine named rather than language — reads right. A sibling `omarchy-theme-set-fcitx5` in #12142 makes the candidate window follow the active Omarchy theme, which pairs with this.

komagata:
@akitaonrails Thank you so much for testing this on your setup and taking the time to write such detailed feedback! It’s especially helpful to have confirmation that running it with Mozc already registered leaves the group unchanged.

I’ve addressed both points in f903b4e1:
- Suppressed the output from `fcitx5-remote --check`.
- Clarified that `fcitx5-configtool` should be run from a terminal, since it’s hidden from the launcher.

I also updated the test to catch the stray output. The related shell tests and CLI suite pass.

Thanks as well for opening #12139 and pointing out the related issues and theme integration. I really appreciate you looking beyond setup to what people encounter when they start using Japanese input.

PR #9638, opened 2026-09-01
Switch Bluetooth headsets to HFP when selecting their input

### LINKED ISSUE #9380: Selecting a Bluetooth headset input can leave it in A2DP with a silent microphone
### System details

- Omarchy: 4.0.0-1
- Machine: Lenovo ThinkPad P14s Gen 6 AMD (Ryzen AI 7 PRO 350 / Radeon 860M)
- Kernel: 7.1.8-arch1-3
- PipeWire: 1.6.8-1
- WirePlumber: 0.5.15-1
- BlueZ: 5.87-2
- Quickshell: 0.3.0.r20.g28771c7-1
- Headset used for reproduction: HUAWEI FreeBuds Pro 5
- Reproduced on two separate Omarchy machines with the same headset

### What is wrong?

The Omarchy audio panel lists the Bluetooth headset under **Input** and allows it to be selected. The row becomes highlighted and PipeWire records the Bluetooth loopback source as the default, but the headset can remain in playback-only A2DP. Applications then receive no microphone signal even though the panel visibly says the Bluetooth input is selected.

The headset advertises HFP correctly, and manually changing the card to `headset-head-unit` immediately makes the microphone work. This does not appear to be specific to Huawei; any classic Bluetooth headset that uses A2DP for playback and HFP/HSP for duplex audio may be affected.

### Steps to reproduce

1. Connect a Bluetooth headset that supports A2DP and HFP/HSP.
2. Leave it in its normal high-fidelity A2DP profile.
3. Start an application with an active capture stream, such as OBS, a browser call, or a microphone test.
4. Open the Omarchy audio panel.
5. Select the Bluetooth headset under **Input**.
6. Observe that the Bluetooth input is highlighted but the microphone remains silent.
7. Check the card profile with `pactl list cards`: it can still report `Active Profile: a2dp-sink`.
8. Run `pactl set-card-profile bluez_card.<MAC> headset-head-unit`. The microphone starts working immediately.

### Expected behavior

Selecting a Bluetooth headset input while a capture stream is active should result in a working input. WirePlumber should transition the card to the preferred duplex HFP profile, and return it to A2DP after capture ends.

### Actual behavior

The UI and default-source metadata change, but the profile transition does not reliably complete. The user is left with a selected but silent Bluetooth microphone until HFP is activated manually.

### Diagnostics

WirePlumber autoswitch is enabled:

```text
$ wpctl settings bluetooth.autoswitch-to-headset-profile
Value: true
```

The device exposes valid profiles:

```text
a2dp-sink: High Fidelity Playback (A2DP Sink, codec AAC) (sinks: 1, sources: 0, available: yes)
headset-head-unit-cvsd: Headset Head Unit (HSP/HFP, codec CVSD) (sinks: 1, sources: 1, available: yes)
headset-head-unit: Headset Head Unit (HSP/HFP, codec MSBC) (sinks: 1, sources: 1, available: yes)
```

After selecting the headset input, the stable WirePlumber loopback source is correctly saved as the default:

```text
Audio/Source  bluez_input.<MAC-with-colons>
```

At the time of the failed transition, WirePlumber logged this twice:

```text
wp-event-dispatcher: <WpAsyncEventHook:...> failed: <WpSiStandardLink:...> link failed: item deactivated before format was set
```

The relevant helper is `bin/omarchy-audio-input-set-default`. It sets both defaults and then immediately moves every active source output:

```bash
wpctl set-default "$node_id" 2>/dev/null || true
pactl set-default-source "$source_name" 2>/dev/null || true

pactl list short source-outputs 2>/dev/null | awk '{ print $1 }' | while read -r output; do
  [[ -n $output ]] && pactl move-source-output "$output" "$source_name" 2>/dev/null || true
done
```

A possible failure mode is a race between moving the active capture stream and WirePlumber replacing the A2DP graph with the HFP graph. The helper suppresses move/link errors and has no wait or retry, so the panel can report success even when the stream did not become usable.

### Workaround

```bash
pactl set-default-source bluez_input.<MAC-with-colons>
pactl set-card-profile bluez_card.<MAC-with-underscores> headset-head-unit
```

Return to high-quality playback afterward with:

```bash
pactl set-card-profile bluez_card.<MAC-with-underscores> a2dp-sink
```

### Related reports

- #1734 reported AirPods Pro 2 playback/microphone profile problems, but it predates the current Omarchy audio panel and is closed.
- #1818 covered Bluetooth devices missing from the audio device list rather than a listed and selected source failing to activate.
- wwmm/easyeffects#4878 describes a similar open WirePlumber limitation where access through a virtual/filter source does not trigger the hands-free profile.

fresh3nough:
## Summary

Selecting a Bluetooth headset under **Input** in the audio panel called `omarchy-audio-input-set-default`, which only set the default source. If the card was still on A2DP (playback-only), PipeWire metadata updated and the row highlighted, but the microphone stayed silent until the user manually ran:

```bash
pactl set-card-profile bluez_card.<MAC> headset-head-unit
```

## Fix

Before claiming a `bluez_*` source, `omarchy-audio-input-set-default` now:

1. Derives `bluez_card.<MAC>` from the source name
2. Reads the card's profiles via `pactl list cards`
3. If the active profile exposes **no sources**, switches to a duplex profile (`headset-head-unit*` preferred, otherwise any profile with `sources > 0`)
4. Then sets the default source and moves active source-outputs as before

Non-Bluetooth sources are unchanged. Cards already on an HFP/HSP profile are left alone.

## Tests

```
bash test/shell.d/audio-input-set-default-test.sh
bash test/shell.d/audio-test.sh
```

- non-Bluetooth input does not touch card profiles
- A2DP bluez input switches to `headset-head-unit`
- prefers `headset-head-unit*` when several duplex profiles exist
- already-on-HFP does not re-set the profile
- output-shaped bluez name still resolves the same card MAC

Fixes #9380

### DIFF
--- a/bin/omarchy-audio-input-set-default
+++ b/bin/omarchy-audio-input-set-default
@@ -12,9 +12,76 @@ if [[ -z $node_id || -z $source_name ]]; then
   exit 1
 fi
 
-wpctl set-default "$node_id" 2>/dev/null || true
-pactl set-default-source "$source_name" 2>/dev/null || true
+# Bluetooth headsets advertise capture only on an HFP/HSP profile. Selecting
+# their input while the card is still on A2DP changes the default source in
+# metadata but leaves the microphone silent (see #9380). Switch the owning
+# bluez card to a duplex headset profile before claiming the source.
+ensure_bluetooth_headset_profile() {
+  local source=$1
+  local mac card profiles active sources_on_active profile
 
-pactl list short source-outputs 2>/dev/null | awk '{ print $1 }' | while read -r output; do
-  [[ -n $output ]] && pactl move-source-output "$output" "$source_name" 2>/dev/null || true
+  # bluez_input.AA_BB_… / bluez_output.AA_BB_….a2dp-sink / bluez_card.AA_BB_…
+  [[ $source == bluez_* ]] || return 0
+
+  mac=$(printf '%s' "$source" | grep -oE '[0-9A-Fa-f]{2}(_[0-9A-Fa-f]{2}){5}' | head -n1)
+  [[ -n $mac ]] || return 0
+
+  card="bluez_card.$mac"
+
+  # pactl "Profiles:" block lists "name: Description (sinks: N, sources: M, …)".
+  # Prefer a profile that exposes a source (duplex headset). Fall back to any
+  # headset-head-unit* name WirePlumber still advertises.
+  profiles=$(
+    timeout 2 pactl list cards 2>/dev/null | awk -v card="$card" '
+      $1 == "Name:" && $2 == card { in_card = 1; next }
+      in_card && $1 == "Name:" { in_card = 0 }
+      in_card && $1 == "Active" && $2 == "Profile:" {
+        print "ACTIVE", $3
+        next
+      }
+      in_card && /^[\t ]+[a-z0-9-]+: / {
+        name = $1
+        sub(/:$/, "", name)
+        sources = 0
+        line = $0
+        sub(/.*sources: /, "", line)
+        sub(/[^0-9].*/, "", line)
+        if (line ~ /^[0-9]+$/) sources = line + 0
+        print "PROFILE", name, sources
+      }
+    '
+  )
+
+  active=$(printf '%s\n' "$profiles" | awk '$1 == "ACTIVE" { print $2; exit }')
+  # Already on a capture-capable profile — leave codec choice alone.
+  if [[ -n $active ]]; then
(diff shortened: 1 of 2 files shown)


filipechagas:
Thanks for putting this together. We confirmed that the root cause is in WirePlumber, and the upstream fix has now been merged in [WirePlumber MR !892](https://gitlab.freedesktop.org/pipewire/wireplumber/-/merge_requests/892). I think this PR should close so Omarchy does not carry a second Bluetooth profile switching policy. Current users will receive the fix when Arch packages a WirePlumber release containing commit `f5c3eb35`. Could the author or an Omarchy maintainer close it?

PR #9653, opened 2026-09-01
Show Bluetooth Alias ahead of the hardware Name

### LINKED ISSUE #9603: Bluetooth panel shows the device's raw hardware Name instead of the user-set Alias
### What happened
Renaming a paired Bluetooth device (via `bluetoothctl set-alias` or setting
the BlueZ `Alias` D-Bus property directly) does not change what the
Bluetooth bar panel displays. It keeps showing the device's original
hardware-advertised name.

### What was expected
The panel should show the custom alias, the same way GNOME/KDE Bluetooth
settings and `bluetoothctl devices` do — Alias is the whole point of a
user-facing rename.

### Root cause
In `shell/plugins/panels/bluetooth/Model.js`, `deviceLabel()` is:

    function deviceLabel(device) {
      if (!device) return ""
      return String(device.deviceName || device.name || "").trim()
    }

`device.name` is Quickshell's `Quickshell.Bluetooth.BluetoothDevice.name`
— writable (`setName`), backed by BlueZ's `Alias`. `device.deviceName` is
the read-only property backed by BlueZ's raw `Name`. The function checks
`deviceName` first, so any device with a real hardware name (i.e. every
device that has ever had a human-readable name) always shows that instead
of a custom alias, because `deviceName` is never empty.

### Suggested fix
Swap the priority:

    return String(device.name || device.deviceName || "").trim()

Verified locally: cloning the plugin and applying this one-line swap
makes the panel correctly reflect the alias.

### Steps to reproduce
1. Pair a Bluetooth device that has its own hardware name (e.g. headphones).
2. `bluetoothctl set-alias "My Custom Name"` (or set `Alias` via D-Bus).
3. Open the Bluetooth bar panel — the custom name never appears.

### System
- Omarchy 4.0.1-1 (stable channel)
- quickshell 0.3.1-1
- Kernel 7.1.9-arch1-2
- CPU: 13th Gen Intel Core i9-13900H
- GPU: Intel Iris Xe (Raptor Lake-P)

Chessing234:
## Summary
- The Bluetooth panel preferred BlueZ `Name` (`deviceName`) over `Alias` (`name`), so user renames never appeared.
- Prefer the alias, falling back to the hardware name when unset.

Fixes #9603

## Test plan
- [ ] `bluetoothctl set-alias "My Custom Name"` then open the Bluetooth panel — custom name shows
- [ ] A device without an alias still shows its hardware name

### DIFF
--- a/shell/plugins/panels/bluetooth/Model.js
+++ b/shell/plugins/panels/bluetooth/Model.js
@@ -1,6 +1,6 @@
 function deviceLabel(device) {
   if (!device) return ""
-  return String(device.deviceName || device.name || "").trim()
+  return String(device.name || device.deviceName || "").trim()
 }
 
 function toArray(values) {
--- a/test/shell.d/bluetooth-test.sh
+++ b/test/shell.d/bluetooth-test.sh
@@ -97,8 +97,16 @@ assertDeepEqual(
   { address: '1', name: 'Deadbeef', deviceName: '', connected: false, state: -1, batteryAvailable: false, battery: 0, pairing: false },
   'bluetooth projects device rows with primitives only'
 )
+// BlueZ Alias is the user-facing rename and arrives as Quickshell's writable
+// `name`; the read-only `deviceName` is the hardware-advertised Name, which is
+// only what to show when nothing renamed the device.
 assertEqual(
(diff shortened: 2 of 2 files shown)


Chessing234:

Chessing234:
folding into #12114 with the other bar status fixes

PR #9655, opened 2026-09-01
Install system-sleep hooks as executables

### LINKED ISSUE #9512: system-sleep hooks installed by omarchy-hibernation-setup and omarchy-toggle-hybrid-gpu are never executed (cp -p preserves mode 644)
## What happened

`omarchy-hibernation-setup` and `omarchy-toggle-hybrid-gpu` install scripts into
`/usr/lib/systemd/system-sleep/` with `cp -p`. The source files in
`default/systemd/system-sleep/` are mode `644`, so `-p` faithfully preserves a
non-executable mode onto the installed copy.

`systemd-sleep` only runs **executables** in that directory (`man 8 systemd-sleep`:
"will run all executables in /usr/lib/systemd/system-sleep/"). A mode-644 file there
is silently ignored — no warning, no journal entry. Both hooks have therefore never
run on any machine that installed them this way.

## Why it matters

Both hooks exist specifically to stop hibernation from failing:

- `keyboard-backlight` — "Turn off keyboard backlight before hibernate to prevent hang
  on power-off. The ASUS keyboard controller can block S4 shutdown if LEDs are active."
- `force-igpu` — "Before hibernating, switch to Vfio so the nvidia driver is detached
  from the dGPU. Without this, hibernate resume fails because the nvidia driver can't
  freeze a powered-off dGPU (returns -EIO), which aborts the entire resume."

So on the exact hardware these workarounds target, hibernate still hangs or fails to
resume, and the failure looks like a kernel/firmware problem rather than a missing
mode bit.

## Steps to reproduce

```
$ sudo omarchy hibernation setup      # or: omarchy toggle hybrid-gpu
$ stat -c '%a %n' /usr/lib/systemd/system-sleep/*
644 /usr/lib/systemd/system-sleep/keyboard-backlight
755 /usr/lib/systemd/system-sleep/unmount-fuse
```

`unmount-fuse` is the control: it is installed by the package itself and lands at
`755`, so it works. The two `cp -p` copies do not.

Confirming the mode is the whole bug:

```
$ pacman -Qo /usr/lib/systemd/system-sleep/keyboard-backlight
error: No package owns /usr/lib/systemd/system-sleep/keyboard-backlight   # written by the setup script
$ stat -c '%a' /usr/share/omarchy/default/systemd/system-sleep/keyboard-backlight
644                                                                      # the source is 644 too
```

## Root cause

Two call sites, same mistake:

- `bin/omarchy-hibernation-setup:92`
  `sudo cp -p "$OMARCHY_PATH/default/systemd/system-sleep/keyboard-backlight" /usr/lib/systemd/system-sleep/`
- `bin/omarchy-toggle-hybrid-gpu:63`
  `sudo cp -p "$OMARCHY_PATH/default/systemd/system-sleep/force-igpu" /usr/lib/systemd/system-sleep/`

Both `default/systemd/system-sleep/keyboard-backlight` and
`default/systemd/system-sleep/force-igpu` are `644` in the repo, and both start with
`#!/bin/bash`.

(`omarchy-toggle-hybrid-gpu:68` also uses `cp -p`, for
`supergfxd.service.d/delay-start.conf` — that one is fine, a drop-in does not need
the executable bit.)

## Suggested fix

Install with an explicit mode rather than inheriting the source's:

```diff
-  sudo cp -p "$OMARCHY_PATH/default/systemd/system-sleep/keyboard-backlight" /usr/lib/systemd/system-sleep/
+  sudo install -m 755 "$OMARCHY_PATH/default/systemd/system-sleep/keyboard-backlight" /usr/lib/systemd/system-sleep/
```

and likewise for `force-igpu`. Setting the executable bit on the two source files as
well would make the intent obvious at rest, but `install -m 755` is what actually
guarantees the outcome regardless of how the files are packaged.

Existing installs need a one-time `chmod +x` — they will not be repaired by an
update, since the stale copies are unowned by any package.

## System details

- Omarchy 4.0.1-1
- Kernel 7.1.9-arch1-2
- MacBookPro12,1 (Intel Broadwell, no dGPU) — found while setting up hibernation;
  this machine has an `smc::kbd_backlight`, so the `keyboard-backlight` hook is live
  here and was silently doing nothing.

Not the same as #9237 (hibernate on T2 Macs), which has a different root cause in the
T2 embedded controller. This one is hardware-independent.

Chessing234:
## Summary
- `cp -p` preserved mode `644` on `keyboard-backlight` and `force-igpu`, so systemd-sleep never ran them.
- Install with mode `755`, and mark the shipped sources executable.

Fixes #9512

## Test plan
- [ ] `omarchy hibernation setup` leaves `/usr/lib/systemd/system-sleep/keyboard-backlight` executable
- [ ] `omarchy toggle hybrid-gpu` leaves `force-igpu` executable
- [ ] `stat -c '%a'` on the installed hooks is `755`


Chessing234:

Chessing234:
closing — quattro already installs these hooks as 0755 via install_root_file.

PR #9656, opened 2026-09-01
Use subl for the Sublime Text default editor

### LINKED ISSUE #9643: Sublime Text install does not expose sublime_text on PATH, so Defaults → Editor never lists it
### System details
Intel Core i7-7820HQ, Intel HD Graphics 630 + AMD Radeon Pro 455, Apple MacBookPro14,3, Omarchy 4.0.1-1 (x86_64)

### What's wrong?

Installing Sublime Text from **Install → Editor → Sublime Text** succeeds, but the app never appears under **Setup → Defaults → Editor**. After install that menu still only listed Neovim.

The install path and the defaults path use two different identities, and only one of them matches what `sublime-text-4` actually ships.

**Install (this part works)** in `default/omarchy/omarchy-menu.jsonc`:

```jsonc
"install.editor.sublime": {
  "action": "omarchy-install-and-launch 'Sublime Text' sublime-text-4 sublime_text"
}
```

That installs package `sublime-text-4` and launches `gtk-launch sublime_text`, which is correct because the desktop file is `sublime_text.desktop`.

**Defaults (this part fails)** in the same file:

```jsonc
"setup.default.editor.sublime": {
  "when": "omarchy-cmd-present sublime_text",
  "checked": "[[ \"$(omarchy-default-editor)\" == \"sublime_text\" ]]",
  "action": "omarchy-default-editor sublime_text"
}
```

`omarchy-cmd-present` is `command -v`. The official package does **not** put `sublime_text` on `PATH`.

On a stock install of `sublime-text-4 4.4200-1`:

| What | Result |
|---|---|
| package | `sublime-text-4` installed |
| `command -v subl` | `/usr/bin/subl` (wrapper: `exec /opt/sublime_text/sublime_text "$@"`) |
| `command -v sublime_text` | missing |
| real binary | `/opt/sublime_text/sublime_text` (not on `PATH`) |
| desktop file | `/usr/share/applications/sublime_text.desktop` (`Exec=/usr/bin/subl %F`) |

So `omarchy-cmd-present sublime_text` fails and the Defaults row is hidden.

**Second failure if you force it:** `omarchy-default-editor` stores the name `sublime_text`, and `omarchy-launch-editor` then does `omarchy-cmd-present "$editor"` and falls back to `nvim` when that command is missing. Selecting Sublime as the default editor would not actually launch it.

Unlike VS Code / Zed / Helix / Emacs, Sublime has no dedicated `omarchy-install-editor-*` script, so install does not create a `sublime_text` wrapper either.

### Steps to reproduce

1. Install from **Install → Editor → Sublime Text** (package `sublime-text-4`).
2. Open **Setup → Defaults → Editor**.
3. Sublime Text is missing; only editors whose exact command is on `PATH` appear (Neovim / `nvim` in this case).
4. `omarchy-cmd-present sublime_text` exits 1; `omarchy-cmd-present subl` exits 0.

### Expected

After installing Sublime from the Omarchy menu, it should appear in Defaults → Editor, and choosing it should make `omarchy-launch-editor` open Sublime.

### Suggested fix

Keep `sublime_text` as the desktop-file id for `gtk-launch`, but treat `subl` as the PATH command (or install a `sublime_text` symlink/wrapper onto `PATH` as part of the Sublime install). `omarchy-default-editor` / `omarchy-launch-editor` need to use a command that actually exists after `sublime-text-4` is installed.

Related but distinct: #7810 (setting a GUI default editor does not update XDG MIME defaults).

`omarchy-debug --no-sudo --print` captured locally at `/tmp/omarchy-debug.log` (Omarchy 4.0.1-1). Happy to attach it if useful; the mismatch above reproduces from the package file list without extra logs.

Chessing234:
## Summary
- `sublime-text-4` puts `subl` on PATH, not `sublime_text`, so Defaults → Editor hid Sublime and `omarchy-launch-editor` fell back to Neovim.
- Store and launch `subl`, map a legacy `sublime_text` default, and gate the menu row on `subl`.

Fixes #9643

## Test plan
- [ ] After installing Sublime, Defaults → Editor lists Sublime Text
- [ ] Choosing it writes `subl` and `omarchy-launch-editor` opens Sublime
- [ ] An older `sublime_text` default file still launches via `subl`

### DIFF
--- a/bin/omarchy-default-editor
+++ b/bin/omarchy-default-editor
@@ -25,7 +25,7 @@ case "$1" in
 code) selection="code"; editor="code"; name="VSCode"; glyph= ;;
 cursor) selection="cursor"; editor="cursor"; name="Cursor"; glyph= ;;
 zed | zeditor) selection="zed"; editor="zeditor"; name="Zed"; glyph= ;;
-sublime_text) selection="sublime_text"; editor="sublime_text"; name="Sublime Text"; glyph= ;;
+sublime_text | subl) selection="sublime_text"; editor="subl"; name="Sublime Text"; glyph= ;;
 helix) selection="helix"; editor="helix"; name="Helix"; glyph= ;;
 vim) selection="vim"; editor="vim"; name="Vim"; glyph= ;;
 emacs) selection="emacs"; editor="emacs"; name="Emacs"; glyph= ;;
--- a/bin/omarchy-default-editor
+++ b/bin/omarchy-default-editor
@@ -17,6 +17,9 @@ if (($# == 0)); then
     read -r editor <"$editor_file"
   fi
 
+  # sublime-text-4 ships `subl`; older defaults stored `sublime_text`.
+  [[ $editor == sublime_text ]] && editor=subl
+
   [[ -n $editor ]] && echo "$editor" || echo "nvim"
   exit 0
 fi
--- a/bin/omarchy-launch-editor
+++ b/bin/omarchy-launch-editor
@@ -18,6 +18,9 @@ else
   editor="nvim"
 fi
 
+# sublime-text-4 ships `subl` on PATH, not `sublime_text`.
+[[ $editor == sublime_text ]] && editor=subl
+
 omarchy-cmd-present "$editor" || editor="nvim"
 
 case "${editor##*/}" in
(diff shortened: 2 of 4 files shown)


Chessing234:

Chessing234:
dropped the when: subl guard so sublime still shows up to install, and mapped stored sublime_text to subl for the checkmark.

laodzuh (commented):
Tested on real hardware: Omarchy 4.0.4-1, sublime-text-4 4.4200-1 from [omarchy], Intel i7-7700 / GTX 1070, with no `sublime_text` shim on PATH (confirmed `command -v sublime_text` fails, only `/usr/bin/subl` exists).

- `bash test/shell`: `default-apps-test.sh` passes all 11 cases. 4 unrelated files fail locally (3 need an omarchy-pkgs checkout, `runtime-smoke-test.sh` collides with my live shell) — same on `quattro`, not this branch.
- With the branch's `bin/` first on PATH: a pre-existing stored `sublime_text` reports as `subl` with no migration; `omarchy-default-editor sublime_text` persists `subl`; `omarchy-launch-editor <file>` opens Sublime instead of falling back to nvim.
- Loaded the changed `setup.default.editor.sublime` row via the user menu extension: Sublime Text appears under Setup > Default > Editor and shows the checkmark.
- Confirmed the `install.editor.sublime` third arg is correctly left as `sublime_text`: `omarchy-install-and-launch` hands it to `gtk-launch`, and the package ships `/usr/share/applications/sublime_text.desktop`, so that's the desktop ID, not the binary.

Also resolves #12787 (closed as a duplicate of #9643).

PR #9657, opened 2026-09-01
Back off fingerprint lock retries instead of spinning

### LINKED ISSUE #9648: Lock screen retries fingerprint auth every 250ms forever with no backoff — ~226 PAM sessions/minute while locked
## What happens

When the lock screen is up and `pam_fprintd` fails, `omarchy-shell` re-arms a fixed 250 ms timer and retries forever. There is no backoff and no attempt cap, so a persistent fingerprint failure becomes an unbounded retry storm that lasts as long as the screen is locked.

On my machine this produced **317,059 PAM sessions since boot** and a sustained **226 per minute** measured over a full minute of an overnight lock.

## The code

`/usr/share/omarchy/shell/plugins/lock/Service.qml`

```qml
onError: function(error) {
  root.fingerprintAuthenticating = false
  if (root.lockRequested && root.fingerprintConfigured) fingerprintRetryTimer.restart()
}

Timer {
  id: fingerprintRetryTimer
  interval: 250
  repeat: false
  onTriggered: root.startFingerprint()
}
```

Every failure re-arms the timer unconditionally. 1000 / 250 = 4 attempts per second, which matches the measured 226/min once process-spawn overhead is accounted for.

## What triggers it here

`fprintd` rejects the claim:

```
fprintd[...]: Authorization denied to :1.318060 to call method 'Claim' for device 'Synaptics Sensors': Device was already claimed
```

`fprintd-list` also shows **three device objects for one physical reader** — `Device/0`, `Device/1`, `Device/3`, with `Device/2` missing — suggesting fprintd leaks device objects as USB re-enumerates across suspend/resume. I can't prove the stale objects *cause* the failed claim rather than merely co-occurring with it, but the retry storm follows from any persistent claim failure regardless of cause.

## Impact

- ~226 process spawns per minute for the entire duration of every lock. An overnight lock is roughly 108,000 PAM subprocesses.
- Measurable battery drain on a laptop left locked.
- Journal growth — a large share of my 511 MB journal was this.
- `fprintd` accumulated 7 min 52 s of CPU time over ~5 days of uptime.

## Steps to reproduce

1. Set up fingerprint auth (`omarchy setup security fingerprint`) on hardware where `fprintd` can end up refusing the claim.
2. Lock the screen and leave it locked.
3. `journalctl --since "-1min" | grep -c omarchy-lock-fingerprint`

Expected: a small number of attempts, then backoff or a stop.
Actual: ~226 and constant for as long as the screen stays locked.

## Suggested fix

Exponential backoff on the retry timer, plus an attempt cap that stops retrying fingerprint and falls back to password-only for the remainder of the lock session. A repeated `Device was already claimed` in particular is not going to resolve by retrying 4×/second.

## Workaround

Deleting `/etc/pam.d/omarchy-lock-fingerprint` disables lock-screen fingerprint while leaving sudo and polkit fingerprint working — `Service.qml` gates the feature on that file existing. Verified: PAM sessions during a lock cycle dropped from ~226/min to zero, and the lock then used a single `omarchy-lock-password` session.

## System

- Omarchy 4.0.1
- Hyprland 0.56.2
- Linux 7.1.9-arch1-2
- Reader: Synaptics Prometheus (06cb:00fc), ThinkPad, Intel Iris Xe

Chessing234:
## Summary
- A failed fingerprint PAM attempt re-armed a fixed 250ms timer forever (~226 sessions/min while locked).
- Exponential backoff with a per-lock attempt cap; exhausted retries leave password unlock available.

Fixes #9648

## Test plan
- [ ] Lock with a working fingerprint still unlocks promptly
- [ ] With fingerprint claim failures, retries slow down and stop instead of hammering PAM
- [ ] Password unlock still works after fingerprint retries are exhausted

### DIFF
--- a/shell/plugins/lock/Service.qml
+++ b/shell/plugins/lock/Service.qml
@@ -22,6 +22,10 @@ Item {
   property bool fingerprintAuthenticating: false
   property bool passwordPamConfigured: false
   property bool fingerprintConfigured: false
+  property int fingerprintRetryAttempt: 0
+  readonly property int fingerprintRetryMax: 8
+  readonly property int fingerprintRetryBaseMs: 250
+  readonly property int fingerprintRetryCapMs: 30000
   property bool previewVisible: false
   property string enteredPassword: ""
   property string pendingPassword: ""
@@ -120,6 +124,7 @@ Item {
     failedAttempts = 0
     authenticatingPassword = false
     fingerprintAuthenticating = false
+    fingerprintRetryAttempt = 0
     fingerprintRetryTimer.stop()
     if (passwordPam.active) passwordPam.abort()
     if (fingerprintPam.active) fingerprintPam.abort()
@@ -223,7 +228,7 @@ Item {
     if (result === PamResult.Success) {
       finishUnlock()
     } else if (fingerprintConfigured) {
-      fingerprintRetryTimer.restart()
+      root.scheduleFingerprintRetry()
     }
   }
 
@@ -238,6 +243,7 @@ Item {
         root.pendingSessionLock = false
         sessionLockStabilizeTimer.stop()
         pendingSessionLockTimer.stop()
+        root.fingerprintRetryAttempt = 0
         root.startFingerprint()
       }
     }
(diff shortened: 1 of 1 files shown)


NeroSong:
This matches the mitigation I observed locally: backing off rapid fingerprint retries significantly improved recovery when an intermittent device failure would otherwise leave the reader unusable.

Chessing234:

Chessing234:
superseded by #11771 (same backoff, rebased onto tip for #9905 / #11462).

PR #9663, opened 2026-09-01
Fix the Hermes CLI readiness probe, and tear the CLI down on uninstall

### LINKED ISSUE #9705: omarchy-install-hermes-cli readiness probe never passes — --oneshot is top-level in hermes --help, not under hermes chat
**Title:** `omarchy-install-hermes-cli` readiness probe never passes — `--oneshot` lives at the top level of `hermes --help`, not under `hermes chat --help`

### Summary

`omarchy-install-hermes-cli --check` always exits 1 against the current Hermes Agent release, so Omarchy never recognizes Hermes as ready and reports "Hermes installed without the interactive seeded sessions Omarchy needs." Hermes does support interactive seeded sessions — the failure is in the probe.

### What I expected

With Hermes Agent installed and runnable, `omarchy-install-hermes-cli --check` (and the `--now` install path) should exit 0, letting the menu / floating terminal / `hermes` default agent launch properly.

### What happened

`hermes_prompt_ready()` probes only `hermes chat --help` for the string `--oneshot`:

```bash
# bin/omarchy-install-hermes-cli
hermes_prompt_ready() {
  local help
  hermes_runs &&
    help=$(timeout 15 "$HOME/.local/bin/hermes" chat --help 2>/dev/null) &&
    grep -qF -- '--oneshot' <<<"$help"
}
```

But Hermes Agent wires `--oneshot` (aliased `-z`) at the **top-level `hermes` command**, not under `hermes chat`. So the string never appears in `chat --help`, and the check always fails.

### Steps to reproduce

```bash
hermes --version
# Hermes Agent v0.19.0 (2026.7.20)  — latest release as of 2026-09-02

hermes chat --help | grep --oneshot   # -> no output
hermes --help     | grep --oneshot   # ->  -z, --oneshot PROMPT  One-shot mode: ...

omarchy-install-hermes-cli --check
echo $?   # -> 1
```

### System details

- Omarchy: `4.0.0.r2004.gb71dcad-1` (`VERSION_ID=4.0.0.r2004.gb71dcad`, Arch-based)
- Hermes Agent: `v0.19.0` (2026.7.20), installed via the Omarchy mise-backed wrapper in `~/.local/bin/hermes`
- Hardware: Dell Precision 5540, Intel i7-9750H, Intel UHD 630 + NVIDIA Quadro T1000 Mobile
- Presumably affects every current Hermes release, since the flag has been top-level in `hermes --help` in the installed and source versions inspected. The commit "Use Hermes native prompted sessions" (2026-08-28) appears to be what introduced the `chat --help`-only probe.

### Suggested fix

Check **either** help surface so the probe stays version-robust (top-level `-z/--oneshot` today, and a `chat`-level spelling in newer builds):

```bash
hermes_prompt_ready() {
  local help chat_help
  hermes_runs &&
    help=$(timeout 15 "$HOME/.local/bin/hermes" --help 2>/dev/null) &&
    chat_help=$(timeout 15 "$HOME/.local/bin/hermes" chat --help 2>/dev/null) &&
    grep -qF -- '--oneshot' <<<"$help$chat_help"
}
```

I've applied this locally (with a backup) so my machine works; this issue is so the fix lands upstream, since any local patch is overwritten by `omarchy update`.

The file also ships a test at `test/shell.d/hermes-cli-test.sh` that may need a companion case for the top-level flag.

**Workaround for other users:** patch the probe in the installer (installed as `/usr/bin/omarchy-install-hermes-cli`, symlinked from `/usr/share/omarchy/bin`) to check `hermes --help` too — but it's overwritten on update, hence this report.

spencerbull:
Two fixes to the Hermes CLI lifecycle.

**The readiness probe keyed on a flag no release ever carried.** `hermes_prompt_ready` in `omarchy-install-hermes-cli` grepped `hermes chat --help` for `--oneshot` — but no released Hermes defines that flag under `chat`: it lived at the top level through v0.20.6, and `chat` only gained its own in v0.21.0. So a fully bootstrapped install read as not ready on every release in circulation: `--check` failed forever, the default-agent flow looped back into the floating installer, and the install pass dead-ended with "Launch Hermes Desktop once to finish installing it" on a machine where it already had. The probe now asks for `--tui` and `--query`, the flags `omarchy-agent` actually passes to seed an interactive session, and a flag counts only when the help defines it — followed by the usage bracket, the gap before same-line help text, an uppercase metavar, or the line end — so prose like "With --tui:" in another option's description, or a `--tui-theme`/`--tui_mode` sibling, never answers for the option itself. It stays a help probe rather than a parse of a real invocation on purpose: a release that ignores unknown arguments would turn that probe into a live session.

**Remove Hermes left the mise CLI stranded.** `omarchy-remove-ai-hermes` dropped the desktop package and the `~/.hermes` runtime but never touched the mise-installed CLI, on the assumption the install-time desktop takeover had removed it; a CLI the app never superseded — an interrupted install, or the terminal CLI from before the app existed — stayed on PATH after uninstall. A new `--remove` mode in `omarchy-install-hermes-cli` tears it down and Remove Hermes calls it. The whole teardown turns on the ownership marker, exactly as replacement already does: without the marked stub nothing proves a mise environment is Omarchy's rather than one the user built against the same spec, so both are left alone. It attempts every step and then judges by what is left — the marked stub still present, or mise still resolving the tool — and a failure names the commands that finish the job by hand, since a rerun that no longer finds the marked stub would own nothing. Remove Hermes tolerates a failed teardown until the runtime and user data are handled, then carries the failure in its exit code. The tool spec and the ownership marker stay defined in one place, so the takeover and the teardown cannot drift.

Closes #9705.

🤖 Generated by Fable 5 in Claude Code. Reviewed by Codex XHigh.

### DIFF
--- a/bin/omarchy-install-hermes-cli
+++ b/bin/omarchy-install-hermes-cli
@@ -101,8 +101,8 @@ hermes_prompt_ready() {
   local help
   hermes_runs &&
     help=$(timeout 15 "$HOME/.local/bin/hermes" chat --help 2>/dev/null) &&
-    grep -qE -- '--tui([^[:alnum:]-]|$)' <<<"$help" &&
-    grep -qE -- '--query([^[:alnum:]-]|$)' <<<"$help"
+    grep -qE -- '--tui([^[:alnum:]_-]|$)' <<<"$help" &&
+    grep -qE -- '--query([^[:alnum:]_-]|$)' <<<"$help"
 }
 
 # --owns answers whether the wrapper on PATH is the one this command wrote, so
@@ -113,18 +113,20 @@ if [[ $mode == "--owns" ]]; then
 fi
 
 # --remove tears down a Hermes CLI this installer put in place -- the mise tool
-# it installs and the stub it marks -- so Remove Hermes clears a CLI the app
+# it installed and the stub it marked -- so Remove Hermes clears a CLI the app
 # never superseded (an interrupted install, or the terminal CLI from before the
-# app existed) rather than leaving it stranded on PATH. Scoped to what is ours:
-# the tool spec is this installer's, and the stub goes only when it carries the
-# marker, so a Hermes the user installed themselves is untouched. Idempotent --
-# nothing installed means nothing to do -- and it is the same teardown the
-# desktop takeover above performs, kept in one place so the two cannot drift.
+# app existed) rather than leaving it stranded on PATH. The whole teardown
+# turns on the marked stub, exactly as replacement does further down: without
+# it nothing proves the mise environment is Omarchy's rather than one the user
+# built against the same spec, and a user's stays theirs. The desktop takeover
+# removes the environment without asking, but that is its own bargain -- a
+# second Hermes has to go whoever built it, and the app still provides the
+# command afterwards; here nothing would. Idempotent: nothing owned, nothing
+# to do.
 if [[ $mode == "--remove" ]]; then
-  mise rm -g "$tool" >/dev/null 2>&1 || true
-  mise uninstall --all "$tool" >/dev/null 2>&1 || true
-
   if ours; then
+    mise rm -g "$tool" >/dev/null 2>&1 || true
+    mise uninstall --all "$tool" >/dev/null 2>&1 || true
     rm -f "$HOME/.local/bin/hermes"
   fi
 
--- a/bin/omarchy-install-hermes-cli
+++ b/bin/omarchy-install-hermes-cli
@@ -92,12 +92,17 @@ hermes_runs() {
 # session and --tui to keep it interactive -- rather than a flag that merely
 # shipped alongside them: the previous marker, --oneshot, was dropped in Hermes
 # v0.20 and left working installs stranded as "not ready".
+#
+# Matched at a flag boundary, not as a substring: a bare grep for --tui also
+# accepts a release that lists only --tui-theme while having dropped --tui
+# itself, which would call an install ready that omarchy-agent cannot drive --
+# the same false verdict this check exists to prevent, merely inverted.
 hermes_prompt_ready() {
   local help
   hermes_runs &&
     help=$(timeout 15 "$HOME/.local/bin/hermes" chat --help 2>/dev/null) &&
-    grep -qF -- '--tui' <<<"$help" &&
-    grep -qF -- '--query' <<<"$help"
+    grep -qE -- '--tui([^[:alnum:]-]|$)' <<<"$help" &&
+    grep -qE -- '--query([^[:alnum:]-]|$)' <<<"$help"
 }
 
 # --owns answers whether the wrapper on PATH is the one this command wrote, so
--- a/bin/omarchy-install-hermes-cli
+++ b/bin/omarchy-install-hermes-cli
@@ -88,13 +88,16 @@ hermes_runs() {
     timeout 15 "$HOME/.local/bin/hermes" --version >/dev/null 2>&1
 }
 
-# The chat subcommand's --oneshot opt-out arrived with native interactive -q,
-# so its presence is a stable capability check without relying on a version.
+# Probed for the flags omarchy-agent actually passes -- --query to seed the
+# session and --tui to keep it interactive -- rather than a flag that merely
+# shipped alongside them: the previous marker, --oneshot, was dropped in Hermes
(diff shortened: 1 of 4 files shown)


spencerbull:
## Review — head 46cfc4ad

**Verdict:** SHIP AFTER FIX (pushed). Every finding below is fixed on this branch; waiting on maintainer review.

**What it does:** Repoints the Hermes CLI readiness probe at the flags `omarchy-agent` actually invokes, and makes Remove Hermes tear down an Omarchy-owned mise CLI it previously left stranded on PATH.

**Findings** (reviewed at pre-fix head 36d52254, each verified against source and each fix proven by mutation — reverting it fails exactly its regression test):

- [high] bin/omarchy-install-hermes-cli — `--remove` ran `mise rm -g` and `mise uninstall --all` for the tool spec unconditionally, gating only the stub on ownership, so uninstalling Hermes Desktop could destroy a mise environment the user built against the same spec while sparing their wrapper — the command the removal claims to preserve, broken behind its back. Fixed in c462aad9: the whole teardown turns on the ownership marker, matching the doctrine the install path already enforces.
- [medium] bin/omarchy-install-hermes-cli — the probe matched `--tui`/`--query` anywhere in the help text, so a mention in another option's description ("With --tui:" already ships in `--dev`'s help) or an underscore sibling like `--tui_mode` answered for the option itself. Fixed in 46cfc4ad: a flag counts only in argparse definition shapes (usage bracket, pre-help gap, uppercase metavar, line end).
- [medium] bin/omarchy-remove-ai-hermes — a failed stub removal aborted the script under `set -e` before any `~/.hermes` handling, while mise failures vanished into `|| true` and the script announced success. Fixed in 46cfc4ad: `--remove` attempts every step then judges by what is left (marked stub still present, or mise still resolving the tool), and Remove Hermes tolerates the failure until runtime and user data are handled, then carries it in its exit code.
- [low] test/shell.d — either probe grep could be deleted without any test failing (fixtures always advertised both flags or neither), and the `--remove` wiring was asserted only when the app's runtime had landed, so moving the call inside that branch would silently drop the interrupted-install case this change targets. Both pinned in c462aad9.

**Tests:** `./test/all` green, 225/225, run on a local dev-linked machine, not on an omabot worker — the Proxmox host is unreachable, and this is an own-branch review with no contributor code, so worker isolation was not required; saying it plainly since a skip is not a pass. The probe fix was also verified live on the machine that reproduced the dead loop, against both the app-provisioned Hermes v0.20.4 runtime and a mise/PyPI v0.19.0 install, and `hermes chat --tui --query=…` was confirmed to seed a live interactive session rather than one-shot.

**Related:** closes #9705, which reports the same defect against the current release (its diagnosis differs — `--oneshot` was removed in v0.20 rather than moved to the top level — but the symptom and the fix are the same). Follows the #7469 merge that introduced this code.

**Codex:** second opinion at xhigh reasoning, read-only, pinned to 36d52254. The high finding and the deleted-grep test vacuity were its contributions rather than confirmations, and stand on their own. Where it agreed with conclusions already drawn, that is agreement only — its independence is not currently guaranteed, since it reads the same filesystem as the session it checks. Rejected from its fix directions: a parse-only probe that runs a real invocation with a sentinel flag, because a release that ignores unknown arguments would turn the probe into a live agent session; definition-shape matching covers the same false-accepts without that risk.

**For the maintainer:** `mise uninstall` leaves the tool's version-alias symlinks and caches behind under `~/.local/share/mise`; the teardown stops at mise's own commands rather than hardcoding mise's directory layout, so that inert residue is accepted. Say so if you want it pruned anyway.

🤖 Generated by Fable 5 in Claude Code. Reviewed by Codex XHigh.

spencerbull:
Both actionable findings above are fixed in 21470fd1: the `--remove` failure message now names the three commands that finish the teardown by hand — a rerun could never work once the stub was gone, exactly as described — and the probe's history is recorded straight in the code comment and the body: no released Hermes defined `--oneshot` under `chat`, so the old probe was keyed to a flag that never existed there, which is why a future hermes-desktop bump to v0.21+ must not be mistaken for the fix. The wrapped-prose and PYTHON_COLORS false-matches stay as accepted robustness debt (unreachable under the 3.13 pin and current help text), and the two mixed commits stand as pushed rather than rewriting public history.

🤖 Generated by Fable 5 in Claude Code.

PR #9667, opened 2026-09-01
fix: idle-inhibit Steam games matching steam_app_*

### LINKED ISSUE #9636: Screensaver fires over fullscreen Steam games every 150s: the idle_inhibit rule can never match a game
### System details

AMD Ryzen 7 9700X, NVIDIA RTX 5070 Ti (nvidia-utils 610.57.04), Omarchy 4.0.0.alpha, Hyprland 0.56.2, kernel 7.1.9-arch1

## Summary

`default/hypr/apps/steam.lua:1` gives Steam an idle inhibitor:

```lua
o.window("steam", { float = true, idle_inhibit = "fullscreen" })
```

Hyprland matches window-rule regexes with `re2::RE2::FullMatch`
(`src/desktop/rule/matchEngine/RegexMatchEngine.cpp:16`), so `"steam"` matches
**only** the literal class `steam` — the Steam client and Big Picture Mode. An
actual game has class `steam_app_<appid>`, so **no Steam game ever receives an
idle inhibitor.**

The result: the screensaver launches on top of a fullscreen game every
`idle.screensaver` seconds (default 150) and is dismissed about a second later.
It reads as a graphics bug — a brief flash where the game looks like it is
re-entering fullscreen, showing the desktop gaps and the gradient window border
— because the screensaver window carries `fullscreen = true`, `float = true`,
`animation = "slide"` (`default/hypr/apps/system.lua:35-37`).

It only affects gamepad play. Gamepads are not on the Wayland seat — SDL reads
them straight from evdev — so controller input never resets the idle timer.
Keyboard/mouse sessions are unaffected, which makes it look intermittent and
hard to pin down.

## Evidence

Correlating Steam's `logs/console_log.txt` game sessions against
`idle-cycle-start` in the user journal, on this machine:

| Session | Length | Screensaver fired |
|---|---|---|
| Elden Ring Nightreign, 2026-08-26 18:48–20:37 | 109 min | **42x**, every 151s |
| Elden Ring Nightreign, 2026-08-30 18:58–19:28 | 30 min | 11x |
| Elden Ring Nightreign, 2026-08-30 19:30–20:58 | 88 min | 34x |
| Elden Ring Nightreign, 2026-08-31 19:43–20:01 (kb+mouse) | 18 min | 0 |
| Elden Ring Nightreign, 2026-09-01 10:41–11:15 (**after fix**) | 33 min | **0** |

Measured interval across those 42 fires: min 151s, median 151s, max 152s,
against the configured `idle.screensaver = 150`.

One cycle, mid-game:

```
Aug 26 18:51:46.724  idle-monitor: idle
Aug 26 18:51:46.724  idle-cycle-start: screensaver=150 lock=300
Aug 26 18:51:46.724  process-start: screensaver ... omarchy-launch-screensaver
Aug 26 18:51:46.881  idle-monitor: active
Aug 26 18:51:46.884  process-exit: screensaver exitCode=0
Aug 26 18:51:47.932  idle-cycle-cancel: screensaver-dismissed
Aug 26 18:51:48.006  process-exit: wake exitCode=0
Aug 26 18:54:17.933  idle-monitor: idle          <- 151s later, again
```

Confirmed the game's class from the Hyprland event socket:

```
openwindow>>557641f17e50,2,steam_app_2622380,ELDEN RING NIGHTREIGN
fullscreen>>1
  -> fullscreen now: class='steam_app_2622380' title='ELDEN RING NIGHTREIGN'
```

Reproduced on two independent machines running the same Omarchy install, and
the one-line fix below was applied to both and resolved it on both.

## Why Big Picture Mode appears to work around it

Big Picture's window class *is* literally `steam`, so it is the one Steam window
the rule does match. Using Big Picture as a workaround happens to inhibit idle —
which is why this looks like a Big Picture vs. desktop-mode difference rather
than a rule-matching bug.

## Repro

1. Set `idle.screensaver` in `~/.config/omarchy/shell.json` (default 150 is fine)
   and leave stay-awake off.
2. Launch any Steam game and let it go fullscreen.
3. Play with a gamepad only — no keyboard or mouse input — for 3+ minutes.
4. The screensaver flashes over the game and is dismissed, repeating every 150s.
5. Confirm: `journalctl --user | grep idle-cycle-start` shows fires during the
   session.

## Suggested fix

The sibling app rules all match their own class correctly —
`geforce.lua:1` (`GeForceNOW`), `retroarch.lua:5`, `moonlight.lua:1`. Steam is
the outlier because its games do not share the launcher's class.

One added line in `default/hypr/apps/steam.lua`:

```diff
 o.window("steam", { float = true, idle_inhibit = "fullscreen" })
+-- Games are steam_app_<appid>, not "steam"; rule regexes are FullMatch, so the
+-- rule above never covers a running game.
+o.window("steam_app_.*", { idle_inhibit = "fullscreen" })
 o.window({ class = "steam", title = "Steam" }, { center = true, size = { 1100, 700 } })
```

Verified: 33 minutes of gameplay with zero idle cycles, against 42 fires in a
comparable pre-fix session — and confirmed independently on the second machine.

Worth considering more broadly: any fullscreen game played on a gamepad has this
problem, not only Steam ones. A general "inhibit idle while any window is
fullscreen" option, or a gamepad-aware idle reset, would cover Lutris, Heroic,
and standalone launchers too — but the one-line Steam fix above resolves the
reported case.

kvnloo:
Closes #9636

Hyprland FullMatchs class regexes, so literal `steam` never matches `steam_app_<id>`. Keep the client float+idle_inhibit rule and add `steam_app_.*` idle_inhibit fullscreen without float.

`test/shell.d/steam-idle-inhibit-test.sh` PASS on `346f98a5`.

### DIFF
--- a/default/hypr/apps/steam.lua
+++ b/default/hypr/apps/steam.lua
@@ -1,4 +1,6 @@
 o.window("steam", { float = true, idle_inhibit = "fullscreen" })
+-- Hyprland FullMatch: "steam" is the client only. Games are steam_app_<id>.
+o.window("steam_app_.*", { idle_inhibit = "fullscreen" })
 o.window({ class = "steam", title = "Steam" }, { center = true, size = { 1100, 700 } })
 o.window("steam.*", { tag = "-default-opacity", opacity = "1 1" })
 o.window({ class = "steam", title = "Friends List" }, { size = { 460, 800 } })
--- a/test/shell.d/steam-idle-inhibit-test.sh
+++ b/test/shell.d/steam-idle-inhibit-test.sh
@@ -9,15 +9,15 @@ const fs = require('fs')
 const steam = fs.readFileSync(root + '/default/hypr/apps/steam.lua', 'utf8')
 
 assert(
-  /o\.window\("steam_app_\.\*",\s*\{\s*idle_inhibit\s*=\s*"fullscreen"\s*\}/.test(steam),
+  /^o\.window\("steam_app_\.\*",\s*\{\s*idle_inhibit\s*=\s*"fullscreen"\s*\}/m.test(steam),
   'Steam games (steam_app_<id>) get a fullscreen idle inhibitor'
 )
 assert(
-  /o\.window\("steam",\s*\{[^}]*idle_inhibit\s*=\s*"fullscreen"/.test(steam),
+  /^o\.window\("steam",\s*\{[^}]*idle_inhibit\s*=\s*"fullscreen"/m.test(steam),
   'the Steam client still gets a fullscreen idle inhibitor'
 )
 assert(
-  !/o\.window\("steam_app_\.\*",\s*\{[^}]*float\s*=\s*true/.test(steam),
+  !/^o\.window\("steam_app_\.\*",\s*\{[^}]*float\s*=\s*true/m.test(steam),
   'Steam games are not forced floating by the idle-inhibit rule'
 )
 JS
--- a/test/shell.d/steam-idle-inhibit-test.sh
+++ b/test/shell.d/steam-idle-inhibit-test.sh
@@ -1,0 +1,23 @@
+#!/bin/bash
+
+set -euo pipefail
+
+source "$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd)/base-test.sh"
+
+run_node_test <<'JS'
+const fs = require('fs')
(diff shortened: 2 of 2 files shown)


kvnloo:
Follow-up: steam idle-inhibit test now anchors `o.window` rules to line start (`/^…/m`) so a commented-out rule fails. Addresses omarchybot comment-blind finding.

kvnloo:

Scoped: steam-idle-inhibit-test.

PR #9673, opened 2026-09-01
fix(editor): use subl as Sublime PATH command

### LINKED ISSUE #9643: Sublime Text install does not expose sublime_text on PATH, so Defaults → Editor never lists it

americos:
## Why
Package `sublime-text-4` puts `/usr/bin/subl` on PATH, not `sublime_text`. `omarchy-default-editor` stored and checked `sublime_text`, so after install Omarchy still treated Sublime as missing and `omarchy-launch-editor` fell back to nvim. Defaults → Editor never listed it. Fixes #9643.

## Scope
- Store/check PATH command `subl`; keep menu selection and gtk-launch desktop id `sublime_text` (install path is already correct — do not change it).
- Map stored `sublime_text` → `subl` in `omarchy-launch-editor` so existing defaults still launch.
- Menu checked-state compares to `subl`.
- Test mock: `sublime-text-4` → `subl` (the previous mock lied and mapped to `sublime_text`).

## Tradeoffs
Store the PATH command, not the desktop id. One extra alias in the case statement (`sublime_text | subl`) and a one-line launch remap. Avoids changing the install/gtk-launch id, which is correct.

## Blast radius
Default-editor selection, launch, and the Defaults menu check for Sublime only. Other editors unchanged. Existing stored `sublime_text` still launches via the remap.

## Verification
`bash test/shell.d/default-apps-test.sh` — all passed, including "editor defaults install every missing editor before selection". No Omarchy GUI repro here (GNOME box, not Omarchy/Hyprland).

Fixes #9643


americos:
Addressed the leftover checked-state: `omarchy-default-editor` with no args now remaps stored `sublime_text` to `subl`, same as launch. The Defaults row ticks without a migration. Covered in `default-apps-test.sh` (`legacy sublime_text default reports as subl`). Scripts kept mode 755. 1407271

bjarneo:
Automated duplication check: this pull request looks similar to #9656, which covers the same Sublime subl default. I keep that one open and close this one to consolidate review.

PR #9682, opened 2026-09-01
Replace YT6801 DKMS driver with upstream support

AFOliveira:
Retire `yt6801-dkms` for Motorcomm PCI ID `1f0a:6801` after proving the upstream `dwmac-motorcomm` driver has taken over every adapter. Successfully migrated systems no longer load the vendor module or expose its private device-control interface.

The migration distinguishes discovery failure from absence, verifies the exact replacement PCI alias, loads it, unloads the vendor module without force, reprobes each device, and checks its exact sysfs binding before removing the fallback package. Final checks require all target adapters to remain bound upstream with the vendor package and module absent. Failed or interrupted steps stay pending and can be retried, including when earlier DKMS removal deleted files for a still-loaded module. Completed machines perform no privileged work for later users.

Fresh installs no longer include the vendor package. The old hardware setup leaf is now an intentional no-op: manually rerunning hardware setup cannot remove an existing fallback without the migration's cutover proof. Package-removal queries propagate errors.

Validation at `36ea505e41451e6c839ebd96eaea245c99bbd459`: 41 focused migration/setup groups, four package-removal groups, 112 CLI checks, syntax and diff checks pass. Independent Daybreak review accepts the correction. Benign fixtures cover multiple devices, exact binding, all discovery failures, fallback retention, partial-state retry, final postconditions, real migration-marker ordering and manual setup reruns. Package and module operations are inert stand-ins; these tests do not establish real networking.

**Physical YT6801 validation remains required:** link, DHCP/traffic, NetworkManager/MAC configuration stability, reboot/initramfs and suspend/resume. A kernel without the replacement leaves the migration pending and retains the fallback; boot a supported Omarchy kernel before retrying. Coordinate omacom/omarchy-pkgs#263, which hardens that fallback for machines that cannot migrate. The two fixes are compatible and serve different machine states.


sayem314:
@robosayem review this and share your opinion

robosayem (approved):
Reviewed the full diff, cross-checked the upstream facts, and re-ran the reasoning on Copilot's two findings. This is one of the more carefully engineered hardware migrations in this repo — my detailed take below, and from my side it is approve-worthy.

**Upstream facts verified independently.** `dwmac-motorcomm` is a real mainline glue driver (Yao Zi's stmmac DWMAC4 glue for the YT6801, merged for kernel 7.0 per LKDDb, `drivers/net/ethernet/stmicro/stmmac/dwmac-motorcomm.c`, alias `pci:v00001F0Ad00006801sv*sd*bc*sc*i*`). The module name is distinct from the DKMS `yt6801` module, so the rmmod/alias-check separation in the migration is sound. The author's live `modinfo` claim on `7.1.9-arch1-2` matches the Arch package shipping this module as a tristate.

**The migration's failure discipline is the standout.** Discovery (`lspci`, `pacman`, `lsmod`) all happen before any state change and fail closed; the exact PCI MODALIAS is verified on the running kernel before the vendor module is unloaded; the upstream driver is loaded before removal; `rmmod` never forces a busy module; the reprobe rebinds through `drivers_probe`; the package is kept until the live cutover is verified; final checks re-assert package/module absence and binding; any failure leaves the migration pending and retriable, and the tests prove each fault preserves the installed fallback with retry-to-success. I checked Copilot's two findings against the current code and they do not hold: lspci failure aborts the migration rather than reading as absent hardware (only a clean empty result reads as absent, which correctly removes the leftover package), and the test explicitly asserts a completed machine runs zero privileged commands for later users.

**The `omarchy-pkg-drop` change fixes a real fail-open bug.** Previously a failed `pacman -Qq` produced an empty installed-set and silently treated every requested package as absent; now it exits with a clear error.

**Two notes, neither a blocker**

1. `install/hardware/fix-yt6801-ethernet-adapter.sh` drops `yt6801-dkms` on the adapter's presence without checking that the running kernel provides the alias — the check the migration performs. The normal path is fresh installs on the current ISO kernel, so this is a narrow window (an existing system on a pre-7.0 kernel re-running the installer path would lose its working vendor driver with no upstream replacement until a kernel update). Mirroring the migration's `modinfo` alias guard here would close that gap cheaply.
2. The author's own "before release" list is the right one: binding verified is not networking verified. Worth adding that the upstream glue generates a random MAC when the eFuse has none (per the v6 patch series), so MAC stability across reboots on affected machines may need NetworkManager attention once real hardware testing happens.

The DKMS retirement itself is the right call — the vendor module ships an unsafe private device-control interface, and dropping it removes that attack surface entirely rather than hardening it.

ErikMelton (approved):

PR #9697, opened 2026-09-01
Report unreadable Claude transcripts once per scan

### LINKED ISSUE #9674: omarchy-agent-usage-claude logs one stderr line per unreadable transcript, filling $XDG_RUNTIME_DIR and killing the compositor with SIGBUS
## Summary

`/usr/bin/omarchy-agent-usage-claude` scans `~/.claude/projects/**/*.jsonl` and writes one line to stderr for every file it cannot open:

```python
except Exception as exc:
  print(f"Ignoring unreadable Claude project file {path}: {exc}", file=sys.stderr)
```

(`scan_projects`, line 197)

There is no deduplication and no limit. The agents bar widget runs the command on a schedule (`refreshIntervalSec`, default 900), quickshell captures its stderr into the instance log, and the log grows without bound.

On my machine `~/.claude/projects/-app/` holds 90,105 root-owned transcripts, mode 600, left behind by Claude Code running as root inside a container with the home directory mounted. Every scan writes 90,105 lines. The instance log reached 3.35 GB across `log.log` and `log.qslog` in about 13 hours and filled `/run/user/1000` (3.2 GB).

## What breaks once the tmpfs is full

The next client that asks for a shm screen copy kills Hyprland. The compositor maps the client's shm buffer from `$XDG_RUNTIME_DIR` and writes the pixels into it. With no space left, the kernel cannot allocate a backing page, so the store raises SIGBUS.

Confirmed from the core dump. The fault address and the mapping it lands in:

```
$1 = (void *) 0x7f7952802010

0x00007f7952800000 0x00007f7953ae6000  /run/user/1000/.hyprpicker_uyi5bh
```

Signal `SIGBUS`, `si_code: BUS_ADRERR`. Stack:

```
#0  libnvidia-eglcore.so.610.57.04
...
#10 Render::GL::CGLFramebuffer::readPixels
#11 Screenshare::CScreenshareFrame::copyShm
#12 Screenshare::CScreenshareFrame::copy
#13 Screenshare::CScreenshareManager::onOutputCommit
```

The NVIDIA frames are the driver's copy loop performing the faulting write, not the cause.

This happened three times — Aug 30 21:52, Sep 1 08:57, Sep 1 22:15 — with identical stacks. The trigger each time was `SUPER + PRINT` (`hyprpicker -a`). Xwayland also fails to start in this state, because xkbcomp cannot write its keymap: `Cannot close "/run/user/1000/server-0.xkm" properly (not enough space?)`.

## Steps to reproduce

1. Create files under `~/.claude/projects/` that the user cannot read:
   ```bash
   sudo mkdir -p ~/.claude/projects/-test
   sudo bash -c 'for i in $(seq 1 50000); do : > ~/.claude/projects/-test/$i.jsonl; done'
   sudo chmod 600 ~/.claude/projects/-test/*.jsonl
   ```
2. Leave the agents bar widget enabled.
3. Watch `$XDG_RUNTIME_DIR/quickshell/by-id/*/log.log` grow by one line per file per refresh.
4. With the tmpfs full, press `SUPER + PRINT`. Hyprland dies with SIGBUS.

## Expected

The warning should be deduplicated, counted, or dropped after the first few. A directory the user cannot read is a normal condition, not something worth one log line per file per refresh.

Separately, the quickshell instance log should have a size cap. Issue #7506 is a different cause with the same result: an uncapped instance log filling `$XDG_RUNTIME_DIR`. There it broke app launching; here it kills the compositor. A cap would contain both.

## System

- omarchy 4.0.2-1
- omarchy-settings 4.0.2-1
- quickshell 0.3.1-1
- hyprland 0.56.2-1
- Linux 7.1.9-arch1-2, NVIDIA 610.57.04
- `/run/user/1000` is 3.2 GB

shaynhornik:
Fixes #9674

## Problem

`omarchy-agent-usage-claude` prints one stderr line for every `~/.claude/projects/**/*.jsonl` it cannot open. The agents widget runs the collector on every `refreshIntervalSec` tick (900 s by default, plus on start), `omarchy-agent-usage-update` captures only its stdout, and Quickshell keeps the collector's stderr in its instance log under `$XDG_RUNTIME_DIR`. A directory of root-owned transcripts (90,105 in the report, left by Claude Code running as root with the home mounted) writes 90,105 lines per refresh. The log filled the tmpfs in about 13 hours, and the next shm screencopy killed Hyprland with SIGBUS.

The sibling scanners already treat an unreadable file as unremarkable: `scan_pi_usage` in the same file and the Codex collector both skip it silently. This loop was the odd one out.

## Change

Count unreadable transcripts during the scan and report them once per scan, keeping the first path and error in the line so the culprit is still findable. Readable transcripts are counted exactly as before, and a scan with nothing unreadable prints nothing.

## Verification

- `test/shell.d/agent-usage-claude-scanner-test.sh` gains a case with three unreadable transcripts beside a readable one: exactly one stderr line, the count in it, and `todayTotalTokens` unchanged. It uses mode-000 files, or directories wearing the transcript suffix when the tests run as root. Passes locally, 9 ok.
- Reporter's shape, before and after: 25 unreadable entries under a scratch `HOME` produced 25 stderr lines at quattro HEAD and one line with this change. A projects tree with no unreadable entries prints nothing.
- `agent-usage-update-test.sh`, `agent-usage-claude-limits-test.sh`, and `agents-panel-test.sh` still pass.

The uncapped Quickshell instance log (#7506) is a separate change and is left alone here.

Written with Claude Code; I reviewed the change and ran the tests above locally.

https://claude.ai/code/session_01FTqr6ZjXfNthXT719ag7Qc

### DIFF
--- a/bin/omarchy-agent-usage-claude
+++ b/bin/omarchy-agent-usage-claude
@@ -132,6 +132,8 @@ def scan_projects(projects_path: Path) -> dict[str, Any]:
   prompts = 0
   today_prompt_count = 0
   today_token_total = 0
+  unreadable = 0
+  first_unreadable = ""
 
   files = projects_path.rglob("*.jsonl") if projects_path.is_dir() else []
   for path in files:
@@ -194,7 +196,21 @@ def scan_projects(projects_path: Path) -> dict[str, Any]:
             today_token_total += total
             today_tokens[model] = today_tokens.get(model, 0) + total
     except Exception as exc:
-      print(f"Ignoring unreadable Claude project file {path}: {exc}", file=sys.stderr)
+      # A transcript the user cannot read is a normal condition: Claude Code
+      # run as root with the home mounted leaves thousands behind. The shell
+      # keeps this collector's stderr in its tmpfs log, and a line per file per
+      # refresh filled $XDG_RUNTIME_DIR until the compositor died. Count them
+      # and say so once per scan.
+      unreadable += 1
+      if not first_unreadable:
+        first_unreadable = f"{path}: {exc}"
+
+  if unreadable:
+    noun = "file" if unreadable == 1 else "files"
+    print(
+      f"Ignoring {unreadable} unreadable Claude project {noun} under {projects_path}; first: {first_unreadable}",
+      file=sys.stderr,
+    )
 
   return {
     "todayPrompts": today_prompt_count,
--- a/bin/omarchy-agent-usage-claude
+++ b/bin/omarchy-agent-usage-claude
@@ -132,8 +132,8 @@ def scan_projects(projects_path: Path) -> dict[str, Any]:
   prompts = 0
   today_prompt_count = 0
   today_token_total = 0
-  unreadable = 0
-  first_unreadable = ""
+  skipped: dict[str, int] = {}
+  skipped_examples: dict[str, str] = {}
 
   files = projects_path.rglob("*.jsonl") if projects_path.is_dir() else []
   for path in files:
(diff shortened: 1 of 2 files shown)


StefanMaron:
Thanks for the fix — I filed #9674.

Two points on the change.

**Collapsing to the first failure loses information that matters.** The `except Exception` wraps the whole per-file body, so it catches more than permission errors: any failure while processing a transcript lands there. Before this change, each distinct failure printed its own exception text. Now 90,105 permission errors and one genuine parse bug collapse into a count plus whichever failure `rglob` reached first, and that order is arbitrary. Counting by exception type keeps the diagnostic value at the same cost:

```python
unreadable[type(exc).__name__] += 1
```

Reporting those counts still gives one line per scan, and it distinguishes 90,000 unreadable files from 89,999 unreadable files plus one real bug.

**The instance log cap is the part that prevents the crash.** Leaving #7506 as its own change is reasonable, but it is worth being explicit about what remains until it lands. This PR removes the one subprocess known to fill the log, but any chatty subprocess the shell spawns can still fill `$XDG_RUNTIME_DIR`, and a full runtime directory kills the compositor with SIGBUS instead of only losing logs. That is the step that turns a logging problem into a crash.

Two smaller notes: the singular `noun = "file"` branch has no test, and `unreadable` is not added to the returned dict, so the shell cannot show the count.

The root test case using a directory with the `.jsonl` suffix is a good detail. A `chmod 000` file would have passed silently under a root CI run while testing nothing.

Reviewed by Claude Opus 5 via Claude Code.

shaynhornik:
Thanks, all three taken in 412d5fd1.

- Counted per exception type now, still one line per scan, with the first path for each type: `Skipped 26 Claude project files under …: IsADirectoryError 25 (first …), PermissionError 1 (first …)`.
- `skippedFiles` and `skippedFilesByError` are in the record. They also survive the stats-cache fallback, which replaces the scan dict when nothing was readable, the case where the count matters most.
- Tests cover the singular line, and an unprivileged run stages a directory and a mode-000 file together and asserts the two errors are kept apart. Under root that case passes with a skip line, the same way the compositor-gated tests do.

On the log cap: agreed that the cap is what turns a logging problem into a crash, and this PR only removes the one known filler. I would rather not bundle a shell-wide log policy into a collector fix, so #7506 stays the place for that.

Issue #8993, opened 2026-08-29
SUPER + J errors on scrolling-layout workspaces


seth-wood:
### System details

AMD Ryzen 5 PRO 4650U, Radeon Vega (Renoir), Omarchy 4.0.1-1, Hyprland 0.56.2

### What's wrong?

`SUPER + J` errors on any workspace using the scrolling layout:

    error: =[C]:-1: no such layoutmsg for scrolling

Two stock bindings conflict. `SUPER + L` puts the active workspace into
Hyprland's scrolling layout, and `SUPER + J` dispatches a dwindle-only layout
message:

    -- default/hypr/bindings/tiling.lua:4
    o.bind("SUPER + J", "Toggle window split", hl.dsp.layout("togglesplit"))

Steps to reproduce, on stock config:

1. `SUPER + L` to switch the active workspace to scrolling
2. `SUPER + J`

Or without keybindings:

    hyprctl dispatch 'hl.dsp.layout("togglesplit")'

Expected `SUPER + J` to no-op on a scrolling workspace instead of surfacing a
Hyprland error.

Hyprland doesn't intend to silence unsupported layout messages -- in
hyprwm/Hyprland discussion #14237 the suggested resolution is for the caller
to check the active layout first. So the guard likely belongs here, e.g. on
`hyprctl activeworkspace -j | jq -r .tiledLayout`, which
omarchy-hyprland-workspace-layout-toggle already reads.

Happy to open a PR if the approach sounds right.

rookepoole:
Open PR #8939 addresses this Lua runtime error. It checks the active tiled layout before dispatching togglesplit: dwindle keeps the existing behavior, while layouts that do not support it receive a notification instead. The PR includes focused tests.

bjarneo:
Automated duplication check: this issue looks similar to #7556, which covers the same Super+J togglesplit errors on scrolling workspaces. I keep that one open and close this one to keep the discussion in one place.

If you feel this is the wrong decision, please open the issue again with a note on the difference.

PR #8939, opened 2026-08-29
fix(hypr): prevent SUPER+J runtime error on scrolling layout

heyssh:
## Problem
SUPER+J always sent the dwindle-only `togglesplit` layout message.
Once a workspace was on the scrolling layout (toggled via SUPER+L),
pressing SUPER+J threw a Lua runtime error popup instead of doing
anything:

    Runtime error in lua: ?.? no such layoutmsg for scrolling

## Fix
Added `bin/omarchy-hyprland-window-togglesplit`, a small script that
checks `hyprctl activeworkspace -j | jq -r .tiledLayout` before acting:

- **dwindle** → dispatches `hyprctl dispatch 'hl.dsp.layout("togglesplit")'`
  (unchanged behavior)
- **any other layout** → shows `omarchy-notification-send` with a short
  message instead of letting the runtime error surface

`default/hypr/bindings/tiling.lua` now points SUPER+J at this script,
the same pattern already used by SUPER+L
(`omarchy-hyprland-workspace-layout-toggle`).

## Testing
- SUPER+L → scrolling, SUPER+J → clean one-line notification, no error
- SUPER+L → back to dwindle, SUPER+J → split toggles normally

Before:
screenshot-2026-08-29_15-54-08
After:
image

PR #9704, opened 2026-09-01
Make Super + J flip the axis on scrolling workspaces

### LINKED ISSUE #7556: SUPER+J runs dwindle-only togglesplit on scrolling workspaces
### System details

Lenovo ThinkPad X13 Gen 2i; Intel Core i5-1145G7; Intel Iris Xe; Omarchy 4.0.0-1; Hyprland 0.56.2; kernel 7.1.8-arch1-3.

### What's wrong?

Omarchy's stock `SUPER + J` binding always sends the dwindle-only `togglesplit` layout message:

```lua
o.bind("SUPER + J", "Toggle window split", hl.dsp.layout("togglesplit"))
```

The stock `SUPER + L` binding can switch an individual workspace to the scrolling layout. Once a workspace is using scrolling, pressing `SUPER + J` produces a Lua runtime-error popup saying there is no such layout message for scrolling.

Steps to reproduce:

1. Use the stock Omarchy Hyprland bindings.
2. Press `SUPER + L` to switch the active workspace to the scrolling layout.
3. Press `SUPER + J`.
4. Observe the runtime error: `no such layoutmsg for scrolling`.

Expected: a stock shortcut should not produce a Lua runtime error on a layout enabled by another stock shortcut. It could dispatch a scrolling equivalent, such as `consume_or_expel prev`, or safely do nothing outside dwindle.

The failing binding is currently in `/usr/share/omarchy/default/hypr/bindings/tiling.lua`. A working local workaround is to inspect `hl.get_active_workspace().tiled_layout` and dispatch `togglesplit` only for dwindle, with a scrolling-specific action for scrolling.

I searched the Omarchy issue tracker for scrolling, SUPER+J, layoutmsg, and togglesplit before filing and did not find a report for this behavior.

xymbol:
Closes #7556.

## The choice this makes

`Super + J` dispatched the dwindle-only `togglesplit` layout message, so on a scrolling workspace the key did nothing except raise Hyprland's own error popup, Lua chunk info and all:

```
Runtime error in lua: =[C]:-1: no such layoutmsg for scrolling
```

The Lua binding layer reports before the result reaches `CKeybindManager::handleKeybinds`: `Internal::reportError` logs unconditionally and, at ERROR level, calls `CConfigManager::addError`, which pops the notification overlay. This is what #7556 and #8220 describe.

This is deliberately not just error suppression. The scrolling layout has a native axis, so `Super + J` can mean the same thing in both modes:

| layout | what `Super + J` flips |
|---|---|
| dwindle | whether the focused window and its neighbour sit **side by side** or **stacked** |
| scrolling | whether the workspace runs **rightward** or **downward** — where new windows land, and which way it scrolls |

One key, one meaning in both layouts: flip this arrangement between horizontal and vertical. The scope differs — dwindle flips one pair, scrolling flips the whole workspace — but the question the key asks is the same.

## How

`scrolling:direction` takes `right`/`left`/`up`/`down`, and `CScrollingAlgorithm::getDynamicDirection()` resolves it **per workspace** off the workspace rule before falling back to the global option — the same mechanism `master:orientation` uses through `getDynamicOrientation()`. There is no `layoutmsg` for it and no IPC to query it, so a workspace rule is the only way to flip the axis at all.

The binding routes through `omarchy-hyprland-window-split-toggle`:

- **dwindle** — keeps `togglesplit`, unchanged
- **scrolling** — flips the axis between `right` and `down`, with a notification
- **anything else** — an Omarchy notification instead of Hyprland's runtime-error popup
- **a Hyprland not reporting `tiledLayout`** — keeps the `togglesplit` this binding always sent

### Where the axis lives, and how long

While a workspace stays on scrolling, its axis sticks — across reloads and restarts — exactly like the layout choice it is an attribute of, which the manual already promises "sticks... after a restart".

**Changing layout forgets it**, matching dwindle, whose splits are destroyed on a mode change: `updateWorkspaceLayouts()` replaces the algorithm instance outright, so every `SDwindleNodeData::splitTop` goes with it. Scrolling's own runtime state (`SScrollingData` — columns, widths, tape offset) is discarded the same way, for free.

`Super + L` therefore saves no direction, and additionally re-applies the configured default **live**: `hl.workspace_rule` merges into an existing rule rather than replacing it, so a direction set earlier would otherwise linger until the next reload, making behaviour differ before and after one. The reset value is read from `scrolling:direction` at toggle time rather than hardcoded, so a default set in `looknfeel.lua` stays what a workspace falls back to, and nothing stale is ever written to disk.

Exact parity is not available: Hyprland offers config-backed values (survive everything) or eval-backed ones (die on *every* reload — theme changes, monitor hotplug). Dwindle's "survives a reload, dies on a mode change" is a third lifetime that does not exist for a config-typed value. Persisting both in parallel is left as the follow-up this makes room for.

## Building on the earlier attempts

This is the sixth attempt at this bug, and it reuses a lot of what came before. Credit where it is due:

- **#6924** (@seshna) — first to report and fix it, back on 15 Aug. Established that the dispatch has to be guarded by the active layout.
- **#7171** (@fayimora) — made the argument this PR is built on: that `Super + J` should *do* something layout-appropriate on scrolling rather than no-op, and that the manual needs updating alongside. Picked `consume_or_expel prev`. The reasoning here is an extension of that, not a departure from it.
- **#8548** (@gianmegantara) — introduced the `bin/omarchy-hyprland-window-split-toggle` helper shape dispatched from `tiling.lua`. This PR arrived at the same structure independently and keeps it, including the filename.
- **#8939** (@heyssh) — the layout check plus a notification for unsupported layouts, which is exactly what this PR does for the non-dwindle, non-scrolling case.
- **#9165** (@harshithnadig) — same helper path, with the `jq` fallback for a missing `tiledLayout`.

Also useful along the way: @SamSyntax's and @rsd's local workarounds in #8220 and #7556, @rookepoole's triage linking the duplicate reports to #8939, and @bierlingm's consolidation on #7556 which is what made the shape of the deadlock visible.

The one thing none of them had is the axis, and that is the only reason this PR exists rather than a +1 on one of theirs. Everything else here is their design.

#7556 is the original report and the thread @bierlingm consolidated on, so this closes that one. #8220, #8993 and #9146 are duplicates of it rather than separate bugs — flagging them rather than listing them as closed too, since that call is yours. Discussion and the parser evidence are on #7556.

**#8548 and #9165 create the same `bin/omarchy-hyprland-window-split-toggle` path**, so these conflict by construction — whichever lands first, I am happy to rebase onto it, fold the axis handling into that PR instead, or close this if a smaller fix is preferred. The error wants fixing regardless of which answer wins.

## Testing

`test/shell.d/hyprland-workspace-layout-test.sh`, fifteen assertions, one row each:

| assertion | what it holds down |
|---|---|
| the layout toggle persists and applies the chosen layout | `Super + L` still does its original job |
| it ignores broken `hyprctl` output | no workspace id means no rule file written |
| `togglesplit` on dwindle, saved rule untouched | the dwindle arm must not rewrite state on its way past |
| scrolling turns vertical | the axis is written **and** applied |
| then flips back to horizontal | it toggles, rather than only ever setting one value |
| a layout change forgets the axis | dropped from the rule **and** reset live, so a reload cannot change the answer |
| the reset honours a configured `scrolling:direction` | a default set in `looknfeel.lua` is what a workspace falls back to |
| an unusable default is left out of the rule | never send an answer `hyprctl` could not give to the compositor |
| an unsupported layout is explained | an Omarchy notification instead of Hyprland's runtime-error popup |
| the saved rules still load as Hyprland config | the file is replayed, not merely written |
| a duplicated rule reads as its first direction | a hand-edited file must not stop the axis flipping |
| an unrecognized direction is ignored | garbage in the file never propagates back out |
| the `keyword` fallback carries the direction | the string it builds is correct — but see the note below |
| a Hyprland not reporting `tiledLayout` keeps `togglesplit` | pre-Lua compositors keep the behaviour they had |
| the first press creates the state directory and rule | `Super + J` before `Super + L` ever ran, and without spilling to stderr |

Every assertion was checked to fail when the code it covers is deliberately broken. Two things came out of that pass rather than going in. A whitelist on the parsed direction proved observationally inert — the value is only ever compared against `down` and never written back — so it is gone; the equivalent whitelist on `Super + L` guards a value that comes from `hyprctl` rather than from this code, so it stays and now has a test. And the duplicated-rule assertion originally used `grep -F` against a seeded file whose second line was byte-identical to correct output, so a script that wrote *nothing* would have passed it; it now compares the whole file.

### A note on that `keyword` fallback

`hyprctl eval || hyprctl keyword` is inherited from `quattro`, and this change extends it to carry the direction. On 0.56.2 the two halves cannot combine: a Lua-config Hyprland prefixes a failed `eval` with `error:` so the fallback does fire, but `keyword` then refuses outright — *"keyword can't work with non-legacy parsers. Use eval."* — while a legacy-config Hyprland exits 0 from `eval`, so the fallback never runs at all. The test therefore proves the string this constructs is right, not that the path functions.

I have left it because it is the existing idiom here and removing it is a separate decision, but I am happy to drop it from both scripts if you would rather not carry dead code.

`./test/cli` passes. `./test/shell` fails the same four files on this branch and on a pristine `quattro` — three want an `omarchy-pkgs` checkout, one is a live-session shell smoke test — so there is no regression.

Verified in the running UI on Hyprland 0.56.2: both axes with real windows, the notifications, and that the compositor accepts the rules this writes.

One unrelated fix rides along, noted in the commit body: those Lua assertions never actually ran, because a block run as `lua <<'LUA'` exits 0 even when an `assert` inside it fails and nothing checked the status afterwards.

### DIFF
--- a/bin/omarchy-hyprland-window-split-toggle
+++ b/bin/omarchy-hyprland-window-split-toggle
@@ -1,0 +1,48 @@
+#!/bin/bash
+
+# omarchy:summary=Toggle the split direction on the current active workspace
+
+ACTIVE_WORKSPACE_JSON=$(hyprctl activeworkspace -j)
+WORKSPACE=$(jq -r '.id' <<<"$ACTIVE_WORKSPACE_JSON")
+[[ $WORKSPACE =~ ^-?[0-9]+$ ]] || exit 1
+
+LAYOUT=$(jq -r '.tiledLayout // "dwindle"' <<<"$ACTIVE_WORKSPACE_JSON")
+
+LAYOUTS_DIR="$HOME/.local/state/omarchy/workspace-layouts"
+LAYOUT_FILE="$LAYOUTS_DIR/$WORKSPACE.lua"
+
+# Hyprland has no IPC query for the axis, so the saved rule is the only record;
+# take the first match, or a two-rule file reads as both and stops flipping.
+saved_direction() {
+  [[ -f $LAYOUT_FILE ]] || return 0
+
+  sed -n 's/.*direction = "\([a-z]*\)".*/\1/p' "$LAYOUT_FILE" | head -1
+}
+
+case "$LAYOUT" in
+  scrolling)
+    # Scrolling has no togglesplit; flipping its axis is the same intent.
+    if [[ $(saved_direction) == "down" ]]; then
+      NEW_DIRECTION=right
+      HEADLINE="Scrolling horizontally"
+    else
+      NEW_DIRECTION=down
+      HEADLINE="Scrolling vertically"
+    fi
+
+    RULE="hl.workspace_rule({ workspace = \"$WORKSPACE\", layout = \"scrolling\", layout_opts = { direction = \"$NEW_DIRECTION\" } })"
+    mkdir -p "$LAYOUTS_DIR"
+    printf '%s\n' "$RULE" >"$LAYOUT_FILE"
+
+    hyprctl eval "$RULE" >/dev/null 2>&1 ||
+      hyprctl keyword workspace "$WORKSPACE, layout:scrolling, layoutopt:direction:$NEW_DIRECTION"
+
+    omarchy-notification-send -g 󱂬 "$HEADLINE"
+    ;;
+  dwindle)
+    hyprctl dispatch 'hl.dsp.layout("togglesplit")' >/dev/null
+    ;;
+  *)
+    omarchy-notification-send -g 󱂬 "No split to toggle" "The $LAYOUT layout has no split direction"
+    ;;
+esac
--- a/bin/omarchy-hyprland-workspace-layout-toggle
+++ b/bin/omarchy-hyprland-workspace-layout-toggle
@@ -16,6 +16,19 @@ esac
 mkdir -p "$LAYOUTS_DIR"
 printf 'hl.workspace_rule({ workspace = "%s", layout = "%s" })\n' "$ACTIVE_WORKSPACE" "$NEW_LAYOUT" >"$LAYOUT_FILE"
 
-hyprctl eval "hl.workspace_rule({ workspace = \"$ACTIVE_WORKSPACE\", layout = \"$NEW_LAYOUT\" })" >/dev/null 2>&1 || \
-  hyprctl keyword workspace "$ACTIVE_WORKSPACE, layout:$NEW_LAYOUT"
+# hl.workspace_rule merges rather than replaces, so a direction set earlier
+# lingers until the next reload; re-apply the configured default to reset now.
+DEFAULT_DIRECTION=$(hyprctl getoption scrolling:direction -j | jq -r '.str')
+case "$DEFAULT_DIRECTION" in
+  right | left | up | down)
+    RULE="hl.workspace_rule({ workspace = \"$ACTIVE_WORKSPACE\", layout = \"$NEW_LAYOUT\", layout_opts = { direction = \"$DEFAULT_DIRECTION\" } })"
+    KEYWORD="$ACTIVE_WORKSPACE, layout:$NEW_LAYOUT, layoutopt:direction:$DEFAULT_DIRECTION"
+    ;;
+  *)
+    RULE="hl.workspace_rule({ workspace = \"$ACTIVE_WORKSPACE\", layout = \"$NEW_LAYOUT\" })"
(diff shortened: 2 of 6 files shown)


bjarneo:

PR #9723, opened 2026-09-02
Add an embedded dev-env for ESP32 and Arduino boards

HugoluizMTB:
`omarchy install dev-env embedded` installs `arduino-cli`, `platformio-core`, `esptool`, `avrdude`, and `picocom`, then clears the two things that otherwise stand between a freshly plugged board and a working upload.

## Why

The 18 existing dev-envs each absorb their own environment papercuts — Composer on PATH, PHP extensions enabled, opam initialised. Hardware has none, and it has two that bite everyone on the first board:

**Serial devices belong to `uucp` on Arch.** Without group membership, every upload and every serial monitor needs sudo. Almost all advice online names `dialout`, which does not exist on Arch, so the search usually costs more time than the fix.

**ModemManager probes new serial devices on plug-in** and holds them open long enough to break the first upload. The symptom is a failed flash that succeeds on the second try, which reads as flaky hardware rather than a fixable cause.

A udev rule marks the common development-board vendors so ModemManager leaves them alone and the node lands in `uucp`:

| Vendor ID | Chip | Found on |
|---|---|---|
| `1a86` | CH340/CH341 | most low-cost ESP32 and Arduino clones |
| `10c4` | CP210x | ESP32 DevKit |
| `0403` | FTDI | official Arduino, many debug probes |
| `303a` | Espressif native USB | ESP32-S2/S3/C3 |
| `2341` | Arduino native USB | Leonardo, Micro |

## Notes

- Every package comes from the official repositories; no AUR.
- Group membership only applies on a new login, so the script reports that rather than appearing to have failed.
- Follows the group-membership pattern already used by `omarchy-install-gaming-xbox-controllers`.
- `test/cli` passes (116 tests).

Tested on Arch with the toolchain install and udev reload; I did not have a board on hand to confirm the end-to-end upload, so a second pair of eyes on the vendor list would be welcome.

### DIFF
--- a/bin/omarchy-install-dev-env
+++ b/bin/omarchy-install-dev-env
@@ -2,12 +2,12 @@
 
 # omarchy:summary=Install a supported development environment
 # omarchy:name=dev-env
-# omarchy:args=<ruby|node|bun|deno|go|laravel|symfony|php|python|elixir|phoenix|rust|java|zig|ocaml|dotnet|clojure|scala>
+# omarchy:args=<ruby|node|bun|deno|go|laravel|symfony|php|python|elixir|phoenix|rust|java|zig|ocaml|dotnet|clojure|scala|embedded>
 # omarchy:examples=omarchy install dev-env ruby | omarchy install dev-env node
 # omarchy:requires-sudo=true
 
 if [[ -z $1 ]]; then
-  echo "Usage: omarchy-install-dev-env <ruby|node|bun|deno|go|laravel|symfony|php|python|elixir|phoenix|rust|java|zig|ocaml|dotnet|clojure|scala>" >&2
+  echo "Usage: omarchy-install-dev-env <ruby|node|bun|deno|go|laravel|symfony|php|python|elixir|phoenix|rust|java|zig|ocaml|dotnet|clojure|scala|embedded>" >&2
   exit 1
 fi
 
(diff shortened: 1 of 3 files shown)


HugoluizMTB:
Pushed two follow-ups after re-reading `AGENTS.md`:

- **Menu entry** in `default/omarchy/omarchy-menu.jsonc`. *Install > Development* is built from that file, so without it the environment existed but was unreachable from the menu. It disables itself once `arduino-cli` is present, matching how the other environments detect themselves.
- **Manual** — `manual/18-development-tools.md` lists the environments by hand, so the new one is named there too.

`test/cli` (116) and `test/shell.d/menu-test.sh` both pass.

mykbren:
Hello @HugoluizMTB 👋

I found this PR while looking for CH340/CH341 and CP210x, so I'm happy to see it 👍

However, I'm not sure about putting all of this into the embedded dev-env namespace.

The arduino-cli, PlatformIO, esptool, and avrdude parts are clearly embedded-development related. But the serial-device configuration is more generic. CH340/CP210x/FTDI devices aren't necessarily development boards - you might buy a device that uses one of these chips and simply want to use it. You don't necessarily need to enter a dev mode/environment for that.

The Linux kernel also treats these as generic USB serial devices, with the drivers living under drivers/usb/serial, for example:
https://github.com/torvalds/linux/blob/master/drivers/usb/serial/ch341.c

So I'd probably separate the embedded toolchain from the USB/serial device setup.

To keep the distro thin and fast, I think it would be better UX to make these things installable on demand, more like plugins: search for CH340, install what is needed, and you're done.

That way, we don't need to put a growing list of hardware-specific serial configuration into the embedded namespace (and I believe there is much more that would eventually need to be added there), while still making everything easy to discover and install when needed.

The embedded CLI tools themselves make perfect sense to me as an embedded environment - I'm mainly questioning the generic USB/serial part.

My real case. I want to connect to my Solar Inverter via COM port. It needs CH340/CH341 and CP210x. But I do not need embedded coding environment. I just need an USB Serial Driver.

I believe there is much more devices in the world, that just need drivers.

Any thoughts? 🤣

PR #9725, opened 2026-09-02
Feature: Allow user to create floating bar by adding support for custom margin and bar radius.

Marc-AnthonyG:
Allow user to customize the emplacement of the bar without having to fork it. This PR is an iteration over #8975 

## What you get

`[bar] margin` and `radius` in `shell.toml`. `margin` takes the same CSS-style list the border widths already take.

```toml
[bar]
margin = "4 8"   # 4 from the anchored edge, 8 at the two it spans
radius = 16
```

Both default to `0` to replicate current state.

## What changed

- **`Style.qml`** — adds the `margin` and `radius` tokens; `margin` is read through `Geometry.parseWidthSpec`, the width spec the border widths already use, so there is no new vocabulary.
- **`Bar.qml`** — the surface carries the gap on each edge it touches and paints its background in a rounded `Rectangle` rather than the window clear colour, which is what lets a radius round it. Hiding parks by the bar's size *plus* the anchored edge's gap, so no sliver is left on screen.
- **`windowScreenPoint()`** — maps drag points from the bar's own origin. It assumed that origin was the screen corner, which left the drop marker and drag ghost a margin away from the cursor and handed `nearestScreenEdge()` a skewed point.
- **`omarchy-bar-text-color`** — takes an `--inset`. It crops the wallpaper at the screen edge to choose a legible transparent-mode foreground, and that is a strip a detached bar no longer covers.

Result:
screenshot-2026-09-01_21-39-59

### DIFF
--- a/bin/omarchy-bar-text-color
+++ b/bin/omarchy-bar-text-color
@@ -11,6 +11,10 @@ text_color=${3:-}
 background_color=${4:-}
 background_path=""
 screen_size=""
+inset_top=0
+inset_right=0
+inset_bottom=0
+inset_left=0
 
 shift 4 2>/dev/null || true
 while (($# > 0)); do
@@ -21,6 +25,10 @@ while (($# > 0)); do
     ;;
   --screen)
     screen_size=${2:-}
+    shift 2
+    ;;
+  --inset)
+    read -r inset_top inset_right inset_bottom inset_left <<<"${2:-}"
     shift 2
     ;;
   *)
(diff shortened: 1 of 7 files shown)


cristian-fleischer:
Hi @Marc-AnthonyG, thanks for this, and @ekrist1 for #8975. I've built a floating-bar follow-up on top of your two commits and would like to open it as a PR that includes them unchanged, authorship kept.

What it adds on top:

- `bar.floating` in `shell.json`, so users can switch it without editing a theme. Unset, your `margin` behaves exactly as it does here.
- With no theme margin, a floating bar uses Hyprland's `gaps_out` per side, so it lines up with the windows (#13027) and the gap above it matches the gap below.
- An unset `radius` follows Hyprland's rounding on a floating bar.

If you'd prefer, I can hold it until this merges and then open just the top commit. Either way is fine with me. It's part of a small set of opt-in bar changes: #13136

PR #9728, opened 2026-09-02
Fit popped windows to monitor work area

### LINKED ISSUE #9699: SUPER+O (window-pop) opens window larger than screen at non-1x scale
### System details

Hyprland 0.56.2, Omarchy 4, scale 1.6

### What's wrong?

Pressing SUPER+O (pop window out into floating/pinned mode) opens the
window larger than the visible screen area at monitor scale 1.6. The
window has to be manually resized every time to fit on screen.

At scale 1 (no scaling) the same window pops out correctly, smaller
than the screen, as expected.

Root cause: bin/omarchy-hyprland-window-pop hardcodes a default size
of 1300x900 logical pixels regardless of monitor scale:

    width=${1:-1300}
    height=${2:-900}
    ...
    hyprctl dispatch resizeactive exact $width $height address:$addr

hyprctl works in logical pixels (physical resolution / scale). At
higher scale factors the logical resolution can drop below 1300x900,
so the popped-out window ends up larger than the available screen
space.

Steps to Reproduce:
1. Set monitor scale to 1.6 (e.g. via Omarchy monitor settings).
2. Focus any tiled window.
3. Press SUPER+O.
4. Observe the window opens larger than the visible screen and must
   be manually resized to fit.
5. Set scale back to 1 and repeat: the window fits correctly.

Expected behavior:
The default pop-out size should scale with (or be clamped to) the
monitor's logical resolution instead of using a fixed 1300x900,
e.g. clamping to something like min(1300, logical_width * 0.9) and
min(900, logical_height * 0.9).

Workaround: override the keybind in ~/.config/hypr/bindings.lua with
an explicit smaller size, e.g.:

    hl.unbind("SUPER + O")
    o.bind("SUPER + O", "Pop window out (float & pin)", "omarchy-hyprland-window-pop 900 600")

imzihuailin:
Fixes #9699

Size popped windows from the active window's monitor work area when a width or height is omitted.

- Selects the monitor by active-window monitor ID.
- Accounts for monitor scale, rotation, and reserved edges.
- Keeps explicit width, height, and x/y arguments unchanged.
- Falls back quietly to 1300x900 when monitor geometry is unavailable or invalid.
- Reuses the calculated size for both Lua dispatch and the legacy dispatcher fallback.

Testing:

- `bash -n bin/omarchy-hyprland-window-pop`
- `./test/shell.d/hyprland-window-pop-test.sh`
- `./test/cli`
- Manual verification at 1.6x scaling: popped window stays centered within the usable display area.

`./test/shell` was also run. Four unrelated existing/environment-dependent suites failed:

- `config-test.sh`
- `launch-about-test.sh`
- `snapper-test.sh`
- `unowned-system-paths-test.sh`

before/after screenshots
image
image

### DIFF
--- a/bin/omarchy-hyprland-window-pop
+++ b/bin/omarchy-hyprland-window-pop
@@ -3,8 +3,8 @@
 # omarchy:summary=Toggle to pop-out a tile to stay fixed on a display basis.
 # omarchy:args=[width height x y]
 
-width=${1:-1300}
-height=${2:-900}
+width=${1:-}
+height=${2:-}
 x=${3:-}
 y=${4:-}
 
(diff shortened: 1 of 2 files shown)


imzihuailin:
Manual testing update: I tested this at 1x, 1.25x, 2x, 3x, and 4x display scaling. In every case, the popped window stayed within the usable monitor area and centered correctly. I did not encounter any issues.

omcular:
Independently checked the sizing logic here against a display neither the issue nor the tests cover: 2560x1440 at `scale=2`, `reserved=[0,26,0,0]`, Omarchy 4.0.2-1 / Hyprland 0.56.2.

Running this branch's `set_default_size` jq against that monitor's live `hyprctl monitors -j`:

```
usable = 1280x694  ->  pop size = 1152x624
```

which fits the work area with margin. (To be precise about scope: I exercised the sizing computation against real monitor output, not a full end-to-end pop on the patched branch. My bar is on the top edge, so this does not exercise the left/bottom `reserved` paths.)

Two things I like about the approach here, having looked at the alternatives:

- Selecting the monitor by the active window's `monitor` id rather than the focused monitor is the right call — the window being popped is not necessarily on the monitor holding focus.
- Clamping against the work area rather than the raw logical resolution is what makes the 1440x900 class of panel in #9731 come out right; that one is exactly 900 logical pixels tall and fails only because the bar takes 26 of them.

Confirming details for my display are in #9699.

bjarneo:
Automated duplication check: this pull request looks similar to #8517, which covers the same window pop monitor sizing. I keep that one open and close this one to consolidate review.

PR #9730, opened 2026-09-02
Expose image picker cursor control over IPC

sunny0826:
## Summary

- expose live Image Picker state and remote cursor control through the existing `image-selector` IPC target
- add optional semantic picker contexts while preserving the existing `open` and `preload` methods
- mark the theme and background pickers with `theme` and `background` contexts, and document the public contract

The new control surface is:

```bash
omarchy-shell image-selector state
omarchy-shell image-selector setCursor <path-or-name>
omarchy-shell image-selector apply
```

`setCursor` performs an exact match against the loaded path, file name, or extension-free name. It only moves the picker cursor. `apply` reuses the existing file-based selection flow. Calls made while the picker is closed return `closed`, and unknown cursor values return `unknown` without changing the current selection.

This first version intentionally uses polling through `state`. It does not add a runtime state file, event mechanism, or any Touch Bar-specific code.

## Testing

- `bash test/shell.d/image-picker-test.sh`
- `bash test/shell.d/menu-images-test.sh`
- `bash test/shell.d/runtime-smoke-test.sh` (real Wayland/Quickshell IPC)
- `./test/cli`
- `./test/shell` — affected tests pass; the full local run still reports unrelated environment failures from the missing `omarchy-pkgs` checkout and the About animation assertion. A transient Network QR failure passed on immediate rerun.

Related discussion: #9536

### DIFF
--- a/bin/omarchy-menu-images
+++ b/bin/omarchy-menu-images
@@ -1,9 +1,10 @@
 #!/bin/bash
 
 # omarchy:summary=Open a generic image selector menu
-# omarchy:args=[--selected <image>] [--print-name] [--show-labels] [--filterable] [--lazy-thumbnails] [--preload] [--cache-only] <image-dir>...
+# omarchy:args=[--selected <image>] [--context <name>] [--print-name] [--show-labels] [--filterable] [--lazy-thumbnails] [--preload] [--cache-only] <image-dir>...
 
 selected_image=""
+context=""
 print_name=false
 show_labels=false
 filterable=false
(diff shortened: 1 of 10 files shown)


sunny0826:
@dhh Could you take a look at this when you have a moment?

It exposes live Image Picker state and remote cursor control over the existing image-selector IPC (state / setCursor / apply), with optional theme and background contexts. More context and the task write-up are here: https://app.basecamp.com/5994298/buckets/48688864/card_tables/cards/10262905397

Would love your feedback.

Issue #5426, opened 2026-04-24
Feature idea: Window Restore — save and restore closed windows via keybindings


TimHerb2005:
## The Problem

On Omarchy, pressing `SUPER+W` closes a window permanently. If you accidentally close a browser window, a Terminal, or a PWA — it's gone. You have to reopen the app, navigate back to where you were, and potentially log in again.

## The Idea

Replace `SUPER+W` with a smarter close that saves the window to a history stack. A second keybinding restores the last closed window.

This is particularly useful for:
- Browser windows accidentally closed
- PWAs (Teams, ChatGPT, etc.) where state matters
- Any app where re-navigating is annoying

## Implementation

I built this as an AUR package and have been running it on Omarchy for a while:

- **GitHub:** https://github.com/TimHerb2005/omarchy-window-restore
- **AUR:** https://aur.archlinux.org/packages/omarchy-window-restore-git
- **Demo:** https://github.com/TimHerb2005/omarchy-window-restore/releases/download/v1.4.0/demo.gif
- **Wiki / Full docs:** https://github.com/TimHerb2005/omarchy-window-restore/wiki

### Keybindings

| Shortcut | Action |
|---|---|
| `SUPER+W` | Close window *(saved to history)* |
| `SUPER+R` | Restore last closed window |
| `SUPER+SHIFT+R` | Restore all closed windows |
| `SUPER+ALT+R` | Pick window from a menu |
| `SUPER+CTRL+R` | Settings |
| `SUPER+ALT+W` | Close permanently *(original killactive behavior)* |

### Two modes

**Relaunch** (default): kills the process, relaunches on restore — low memory, works for all apps.

**Hide**: moves the window to an invisible workspace — full state preserved, YouTube keeps playing, instant restore.

### Everything is opt-in and reversible

The feature is packaged separately (`omarchy-window-restore-git` on AUR) so it doesn't affect anyone who doesn't install it. An uninstall command removes all scripts, keybindings and cache files completely.

## PR

I've also opened a PR (#5425) that adds `omarchy-install-window-restore` to Omarchy's installer scripts, following the same pattern as `omarchy-install-tailscale`, `omarchy-install-vscode`, etc. This would make it installable from the Omarchy menu.

Happy to discuss the approach or adjust anything to fit Omarchy's conventions better.

j5onrf:
submap is lightweight 

```
Press SUPER+Q, then tap Q to kill. Any other key resets.
bind = SUPER, Q, submap, kill

[...]

PR #9747, opened 2026-09-02
Add a native Window Vault for undoable closes

mlmrx:
## Summary

- make Super+W an undoable safe close implemented entirely in native Hyprland Lua
- preserve each live window's workspace, floating geometry, pinned/fullscreen state, and group position
- add last-in-first-out restore, explicit permanent close, vault reveal, stale-entry cleanup, and an opt-out window tag
- keep Super+Q as the normal permanent close path
- document the behavior and add a focused state-machine test

## Motivation

This revisits #5426 and follows the maintainer direction in #5425 to wait for Hyprland Lua and implement the capability directly in Hyprland.

Credit to @TimHerb2005 for the original Window Restore concept and prototype in #5425. This implementation is independent and uses Omarchy's current Lua configuration stack.

## Behavior and scope

Super+W moves the active window to a hidden special workspace while leaving its application process alive. Super+Shift+T restores the most recently vaulted window, Super+Alt+W reveals the vault, and Super+Shift+Q visibly restores then gracefully closes the latest vaulted window so an unsaved-work prompt cannot be hidden.

The feature protects Omarchy's safe-close binding. It cannot intercept an application's own close action or recover a process after it exits. Windows tagged omarchy-no-window-vault bypass the vault.

## Tests

- bash test/shell.d/hyprland-window-vault-test.sh
- bash test/shell.d/hyprland-binding-conflicts-test.sh
- luac syntax validation for the module and binding file
- git diff --check

Live Hyprland visual verification was not available on this Windows host; the focused test covers state transitions and dispatched behavior with the real Lua 5.4 interpreter.


TimHerb2005:
Thanks for the Credit @mlmrx, just brought a Quattro version out but some things need to be fixed :D

PR #9749, opened 2026-09-02
Swallow the keystroke that wakes a blanked lock screen

LorenzGit:
## Bug Description

When the lock screen has blanked the display (idle blank, or after sleep), the first key that wakes it is inserted into the password field. Space is the obvious case — a lone dot before you have typed anything — but every character does it. Unlock then fails, or you backspace first.

Holding that key also auto-repeats into the field once the display is considered awake.

Lid-close suspend often wins the race against the lock's five-second blank timer, so the same first-key-is-password bug happens after resume even when Omarchy never blanked the panel itself.

## Root Cause

`LockView`'s `Keys.onPressed` always calls `wakeRequested()` and then lets the event fall through to the `TextInput`. Nothing records that the panel is dark, so a wake keystroke is indistinguishable from the first character of the password.

## Fix

- The lock tracks `displayBlanked` in `runBlank()` / `runWake()` and hands it to the view.
- The view samples that flag *before* `wakeRequested()` clears it, accepts the event, and keeps swallowing auto-repeats until that key is released.
- Escape and Ctrl+U still clear the field and return before the swallow.
- `omarchy-system-sleep-lock` blanks the panel once the session is secure, so the frozen QML state still says "blanked" on the first key after resume.

Mouse wake is unchanged. Typing after the display is on is unchanged.

Related: #7805. Adjacent open PRs #9157 (dirty fork) and #9211 (does not mark blanked on suspend, and does not swallow auto-repeats).

## How to Verify

1. Lock (`Super+Escape`), wait for the display to blank (~5s), press Space. The field should stay empty; the next character should be the first password dot.
2. Lock, wait for blank, hold a letter. The field should stay empty until you release and type again.
3. Close the lid (or suspend), open it, press a key. That key should wake the panel and not appear in the field.

## Test Plan

- [x] Added `test/shell.d/lock-wake-keystroke-test.sh` and sleep-lock coverage for the blank IPC
- [x] `./test/cli` passes
- [x] Focused lock/sleep tests pass (`lock-wake-keystroke`, `sleep-lock`, `lock-blank-fingerprint`)
- [ ] Manual verification of the fix on hardware (please try a lock-blank and a lid-close cycle)

## Risk Assessment

Low — only the lock password field and the pre-suspend blank. A key pressed while the panel is lit still types. A failed `lock blank` IPC is ignored so suspend is not blocked.

### DIFF
--- a/bin/omarchy-system-sleep-lock
+++ b/bin/omarchy-system-sleep-lock
@@ -113,7 +113,13 @@ sync_clamshell
 # inside the reserve derive_budget_ms already held back for logind.
 while (( $(remaining_ms) > 0 )); do
   case $(lock_state) in
-    secure) exit 0 ;;
+    secure)
+      # Lid-close suspend often wins the race against the lock's five-second
+      # blank timer. Mark the panel blanked before the machine goes down so
+      # the first key after resume wakes the display instead of typing.
+      lock_ipc "$status_timeout_ms" lock blank >/dev/null || true
+      exit 0
+      ;;
     locking) ;;
     *) request_lock ;;
   esac
--- a/shell/plugins/lock/LockView.qml
+++ b/shell/plugins/lock/LockView.qml
@@ -13,6 +13,8 @@ Item {
   property string failureMessage: ""
   property int failedAttempts: 0
   property bool inputEnabled: true
+  property bool displayBlanked: false
+  property bool swallowingWakeKey: false
   property bool loadBackground: true
   property string passwordText: ""
   property bool syncingPasswordText: false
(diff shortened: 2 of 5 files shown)


LorenzGit:
Closing for now.

This patch was never loaded in the reporter's running session (`OMARCHY_PATH=/usr/share/omarchy`). The session lock that required a reboot happened while `./test/shell` was instantiating extra Quickshell processes on the live compositor, which the suite is not supposed to do on a development machine.

Will reopen after VM verification, not live-session testing.

PR #9750, opened 2026-09-02
Kids mode: child installs with a kid password and a parent password

peterholko:
## Kids mode: a child install with a kid password and a parent password

Pairs with the installer half in omacom/omarchy-iso (branch `kids/child-profile`): the ISO asks "Who is this computer for? Me / Child / Another owner" first, and a Child answer takes both passwords and installs the child profile.

### What a child install gets

- **Two passwords.** The kid password is the account password: login, lock screen, and the disk at boot. The parent password is root's password, and it is what every privileged path asks for. Both unlock LUKS, and both open the lock screen.
- **`sudo` and system prompts ask for the parent password.** `Defaults rootpw` plus a parent-worded `passprompt` in `sudoers.d`, and a polkit admin rule naming `unix-user:root`, so the Quickshell polkit agent authenticates root ("Parent password" in the dialog).
- **The kid account is not an administrator.** It is kept out of `wheel` (least privilege, per the project direction), with an explicit per-account sudoers grant that still asks for the parent password, plus the NOPASSWD browser-accent re-grant so theme switching stays one click. `omarchy-menu-timezone` falls back to polkit where the wheel grant does not reach.
- **Wi-Fi asks the parent by default, and the parent can hand it to the kid.** Joining or changing a network is a NetworkManager settings change that Arch waves through for `wheel` only, so the kid account asks for the parent password. `omarchy-parent apply` writes `/etc/omarchy/parent.conf` (the parent's settings file, `wifi=parent` by default) and a polkit rule for the kid account from it; `sudo omarchy-parent wifi kid` lets the kid join the school's network alone, `wifi parent` puts the prompt back. Known networks, scanning, and the Wi-Fi switch never ask; global DNS stays the parent's.
- **The parent password opens the lock screen too.** On a child install the lock stack tries the kid's password first, then hands the typed password to a small helper, `omarchy-parent-unlock`, that asks `sudo` whether it is root's (which `rootpw` already makes it check), with the flag that keeps the credential cache out of it. So a parent can get into a locked session without the kid's password, the unlock leaves no passwordless `sudo` behind, and root's `faillock` limit applies to guesses. A normal unlock by the kid never touches `sudo`. The fingerprint stack and SDDM's login are unchanged.
- **Text consoles are closed** (`getty@tty2..6` masked) so the lock screen is the only way back into a locked session; `sudo omarchy-parent tty on` reopens them.
- **`sudo omarchy-parent`**, the parent's command: `password` (also under _Update > Password > Parent_ in the menu), `apply --user NAME [--remove]` (the plumbing the install leaf and first-boot provisioning call), `tty off|on`, and dispatch of feature commands: anything else runs `omarchy-parent-<feature>`, so later features (screen time, an app whitelist) plug in without editing this command. `install/helpers/parent.sh` shares the stage-`visudo`-install sudoers writer with them.
- **Profile plumbing.** `omarchy-apply-system --profile child` records `/etc/omarchy/profile`; `omarchy-profile-child` is the predicate the rest of the tree keys on; `install/omarchy-child.packages` is the (currently empty) child package list, wired for the ISO and `omarchy-reinstall-pkgs`.
- **Setup form and first boot.** `install/provisioning/setup-form.sh` gains the first question and both password prompts (parent must differ from kid); `omarchy-provision-owner` keeps root's password for the parent, adds the parent LUKS slot, and applies the posture on a child install after a factory reset. A child install asks for neither the full name nor the email: they only feed git identity and the compose shortcuts, and it is the parent at the keyboard, so an answer would have been theirs. The hostname prompt stays, with its `omarchy` default, since it is what the parent sees on the router and two kids' machines both named `omarchy` would collide on the home network.
- **Smaller adjustments.** `faillock` rate-limits root too; fingerprint setup words itself as unlock-only on a child install; FIDO2 setup is hidden there (it only ever covered sudo and polkit); manuals 02, 37, 48 and `docs/file-layout.md` document all of it.

Default installs are unchanged: every new path is gated on the child profile.

### What the kid and the parent see

The wording lives in the shared setup form, so the ISO and the first-boot flow show the same screens. On a child install the user step asks:

```
Username>         Alphanumeric without spaces (like dhh)
Kid password>     Used for login, unlocking, and disk encryption when enabled
Confirm>          Must match the password you just typed

Let's set the parent password...
sudo, updates, installs, and system prompts ask for this one; keep it from the kid.
Parent password>  Asked by sudo, updates, and installs (keep it from the kid)
Confirm>          Must match the parent password you just typed
```

The parent prompt rejects a blank, a mismatch, and a password equal to the kid's, and the confirmation table shows both as a fixed eight-character mask. Full name and email are not asked on a child install; the form goes on to hostname and timezone. The omarchy-iso PR carries the full screen-by-screen flow, starting with "Who is this computer for? Me / Child / Another owner".

After login, the kid sees no kids-mode banner; the one profile-aware onboarding text is the fingerprint invitation, which says "Unlock with your fingerprint." instead of promising sudo. `sudo` prompts read `[sudo] parent password:` and the polkit dialog's field reads "Parent password".

### What the kid account cannot do

- **Refused or hidden**: administrator standing (out of `wheel` and `docker`), the text consoles Ctrl+Alt+F2 through F6, _Setup > Security > Fido2_, root login and `su` without the parent password.
- **Asks for the parent password through sudo**: every _Install_ and _Remove_ entry; _Update > Omarchy_, _Firmware_, _Time_, _Channel_, _Config > Plymouth_, _Hardware > Trackpad_, _Password > Drive Encryption_, _Password > Parent_; _Setup > Direct Boot_, _Reset Computer_, _Security > Fingerprint_, _SSHD_, _Passwordless Sudo_, _Sudoless Docker_; _System > Hibernate_ setup; the hybrid GPU toggle; anything typed with `sudo` or `pacman`.
- **Asks for the parent password through polkit**: joining or changing a Wi-Fi network (unless `omarchy-parent wifi kid`), the DNS toggle, _Update > Timezone_, the Docker TUI, the Windows VM, and every other administrator-only polkit action.
- **Still hers**: login and unlock, fingerprint unlock, her own login password, known Wi-Fi networks, the Wi-Fi switch and scanning, Bluetooth, themes including the browser accent, screenshots, printing, volume and brightness, power off, reboot, suspend, USB sticks, the Hyprland, Hyprsunset, Shell, and Tmux config resets, and every installed app.
- **Not covered**: web content (the web filter is its own PR), software that needs no root, and a live USB with the kid's disk password, which the manual answers with a BIOS password. _Passwordless Sudo_ and _Sudoless Docker_ key on the invoking user, so a parent running them from the kid's session would grant the kid's account; both ask for the parent password first.

### Testing

- Shell suites: `parent`, `parent-unlock`, `profile`, `provision-owner`, `setup-form`, `polkit`, `timezone`, `fingerprint-invitation`, `security-fido2`, `sudoless-docker-posture` under `test/shell.d/` pass on the branch. They were run on macOS with bash 3.2, so the fake-root halves of `parent-test.sh` and `parent-unlock-test.sh` skip there; a Linux run of `./test/all` and `test/cli` remains to be done before merge.
- `provision-owner` checks that the child form skips the full name and email prompts and drops their rows from the summary; the omarchy-iso harness's child run types neither.
- `test/acceptance.d/parent-test.sh` covers the polkit dialog wording and the sudo prompt on a child install; it runs in the acceptance VM.
- Installed three times on a ThinkPad from ISOs built off this branch with the omarchy-iso branch, the last one on 2026-09-01 with the follow-up feature branches merged in for the trial:
  - The installer opens with "Who is this computer for?", the Child path asks for the kid and parent passwords, and the summary masks both.
  - Both passwords open the disk at boot; the kid password logs in.
  - `sudo` prompts with `[sudo] parent password:` and takes only the parent password; the polkit dialog says "Parent password".
  - The lock screen opens with either password. After a parent unlock, `sudo -n true` still asks, so the unlock cached no credential.
  - Joining a new Wi-Fi network asks for the parent password; after `sudo omarchy-parent wifi kid` a fresh join goes through without a prompt, and `wifi parent` puts the prompt back.
  - Ctrl+Alt+F2 through F6 give no login; `sudo omarchy-parent tty on` brings them back.
  - `/etc/omarchy/parent.conf` carries the documented `wifi=` key, and `omarchy-parent --help` lists the core commands.
- The first trial is what surfaced the Wi-Fi prompt, and the second the lock screen, both of which this branch now handles.

### Roadmap

- A web filter by domain (omarchy-parent dns),
- An app filter list (omarchy-parent apps), branch kids/apps, built on this dispatcher.

Plan with the direction this follows: `plans/kids-passwords.md`.


peterholko:
I asked Fable to built this PR from dhh's comment here:


1) Kids mode is triggered during initial install. Right now we let you setup for another owner when you go it Ctrl + C on the keyboard select screen. 
That changes such that the first Q is a full Who is this computer for? 1) Me, 2) Child,3) Another owner
If child, then we trigger a different install profile. That's what'll include the different apps and different themes.
We ask for both a kid password and an adult password. Adult password becomes sudo. Kid is normal login. Both unlock LUKS.
Once in, we have a sudo omarchy-parent bin that controls things.

kenhara:
Kids-mod smoke on this draft, paired with omacom/omarchy-iso#146 (`kids/child-profile`). Built the ISO, Child install, unencrypted, first boot.

## What landed

- First question is **Who is this computer for?** with Me / Child / Another owner. Child asks for a kid password and a parent password. The review table lists both.
- Kid password copy: `used for login, unlocking, and disk encryption when enabled`.
- Parent password copy: `sudo, updates, installs, and system prompts ask for this one; keep it from the kid`.
- After install, `id` is `uid=1000(omarchy) groups=1000(omarchy)` only. Not in `wheel`.
- `sudo -k id` prompts `[sudo] parent password:` (sudo lecture + renamed prompt). I could not finish typing into that raw field from the harness, so I did not prove success/fail of the two passwords at sudo. The prompt text is right.
- **SDDM: kid password got a desktop. Parent password was rejected** (field cleared, stayed on login). That is the important one.

## Please pick up

1. **Parent cannot reach a desktop at SDDM.** Logout / unencrypted boot lands on SDDM. Parent password fails there. `omarchy-apply-lock` / `bin/omarchy-parent-unlock` cover the *lock screen* (kid first, then parent via `sudo -k -S`). `plans/kids-passwords.md` still says fingerprint + SDDM are unchanged. Child also masks tty2–6, so if the lock path is not up yet, a parent is stuck. Product intent: parent should always get into the *same kid session*, not a second account. Wire SDDM the same way as the lock PAM stack, or do not mask ttys until that works.

2. **Installer copy vs behavior.** Kid password claims login + unlocking + disk. Parent claims sudo / updates / installs only. If parent is supposed to unlock the session, say so on the parent screen. If not, the SDDM gap is still the hole.

3. **Stale header.** After the parent password screens, `Let's set the parent password...` stays on Full name, Email, Hostname, Timezone, and the review table. It clears at disk selection. Look at the gum headers in `install/provisioning/setup-form.sh`.

4. **Unencrypted first boot** logged `ERROR: Failed to open encryption mapping: The device PARTUUID=… is not a LUKS volume and the crypto= parameter was not specified.` plus a failed resume from `/dev/vda2`. Child profile should not assume LUKS when the installer did the unencrypted path.

5. **Lock screen** reached an `Enter Password` dialog. Typing the parent password there black-screened the guest display (framebuffer stayed black). Treat as inconclusive on this VM. Please verify parent unlock of the lock screen on KVM or a real machine.

## Not tested here

- Encrypted install / both passwords at LUKS
- Polkit `Parent password` dialog (Update System did not raise one)
- `omarchy-parent wifi` / `tty on`
- Fingerprint

Happy to re-test a follow-up ISO.

kenhara:
*Updating to include Debian and QEMU versions, also deleted earlier redundant comment.*

Full pack (notes + every screenshot): https://gist.github.com/kenhara/f11a6667ff9f7cce290b502dc3cc3b30

Kids-mod smoke on this draft, paired with omacom/omarchy-iso#146 (`kids/child-profile` **73fdccf**). Omarchy tree **f17ab6c**. Local ISO, Child install, unencrypted, QEMU. Used the `omarchy-iso-test --child` defaults (kid `omarchy`, parent `omarchy-parent`). Specifically, I ran this test in Grok bot which is running Debian GNU/Linux 13.6 (trixie) and QEMU 10.0.11 (Debian 1:10.0.11+ds-0+deb13u1).

Product bar: parent password must always reach a desktop in the **same kid session**, not a second user.

## What I saw in the VM

- First question **Who is this computer for?** Me / Child / Another owner. Child asks for kid + parent passwords. Review table lists both.
- Kid copy: used for login, unlocking, and disk encryption when enabled.
- Parent copy: sudo, updates, installs, and system prompts; keep it from the kid.
- **Stale header:** `Let's set the parent password...` stayed on Full name, Email, Hostname, Timezone, and review. Cleared at disk selection. `install/provisioning/setup-form.sh` gum headers.
- Unencrypted first boot still logged `ERROR: Failed to open encryption mapping … is not a LUKS volume` plus failed resume from `/dev/vda2`.
- After install, `id` is `uid=1000(omarchy) groups=1000(omarchy)` only. Not in `wheel`.
- `sudo -k id` prompts `[sudo] parent password:`. Raw field not finished from the harness, so sudo auth success/fail not proven live. Prompt text is right.
- **SDDM: kid password got a desktop. Parent password was rejected** (field cleared, stayed on login). That is the important live finding.
- Lock reached an `Enter Password` dialog. Typing the parent password there black-screened the guest. **Inconclusive** on this VM. Please verify on KVM / hardware.
- Update System did not raise a polkit dialog. It opened the System… menu (Screensaver, Lock, Suspend, Hibernate, Logout, Reboot, Shutdown). Polkit copy not proven live.

Who is this computer for, Child selected

Parent password screen

SDDM after parent password rejected

sudo parent password prompt

id not in wheel

unencrypted first boot LUKS error

## Code (f17ab6c), nothing withheld

**Lock-screen parent unlock — PRESENT in tree, stub-tested only.** `omarchy-apply-lock` writes kid `pam_unix` then `pam_exec` → `omarchy-parent-unlock` (`sudo -k -S -u root -- true`). `test/shell.d/parent-unlock-test.sh` checks text, not a live PAM run.

**`pam_exec` has no `seteuid`.** Lock is accidentally safe because omarchy-shell is the kid. SDDM authenticates as root. Copying this stack onto SDDM without `seteuid` would make **any** password succeed after `pam_unix` fails.

**SDDM parent unlock — MISSING.** `install/login/sddm.sh` only strips gnome-keyring. `plans/kids-passwords.md` says SDDM is unchanged. Autologin is not a substitute (encrypted Autologin fires at SDDM start, not after logout; unencrypted removes it). ISO `establish_session` types the kid password only. `test/acceptance.d/parent-test.sh` never opens SDDM or the lock.

**sudo / polkit / wheel / omarchy-parent / masked ttys / LUKS second slot — PRESENT in tree.** Wheel absence proven live. sudo prompt text proven live. Polkit dialog not proven live. Ttys: `getty@tty2..6` masked; unmask is `sudo omarchy-parent tty on`, which already needs a session.

**Copy overclaims.** ISO configurator / getting-started: parent “opens the disk too so a parent can always get in.” True only for encrypted boot, not logout / SDDM / unencrypted.

### Gates

| Gate | Result |
|---|---|
| Lock | Code yes, same kid session. Live inconclusive. faillock on the kid is `required` *before* parent-unlock, so 10 failed kid tries also lock the parent out for 120s. |
| SDDM | **No. Kid only. Proven live.** |
| sudo | Code yes (`rootpw` + `[sudo] parent password:`). Live: prompt only. |
| polkit | Code yes (`unix-user:root`, dialog “Parent password”). Live not proven. |
| LUKS | Yes if encrypted. No recovery if unencrypted. |
| tty | No parent login path. 2–6 masked. |

Stuck-parent: logout → SDDM kid-only + ttys dead. Encrypted has reboot + LUKS parent + autologin. Unencrypted has **no reboot escape**. The plan treated SDDM as out of scope. The product intent does not.

## Please pick up

1. Wire parent unlock into SDDM as the **same kid user**. Reuse `omarchy-parent-unlock`. Do not add a parent account. Natural place: `omarchy-apply-lock` also ships the child auth stack into `/etc/pam.d/sddm` (or a shared include). Keep session/account stacks intact.

2. **Add `seteuid` on pam_exec for lock and SDDM.** Assert it in `parent-unlock-test.sh`.

3. Tests that would have caught this: apply-lock writes parent helper into the SDDM stack (default profile does not); seteuid present. Acceptance `iso-test --child` unencrypted: parent at SDDM lands in the kid session; kid still works. Encrypted: logout then parent at SDDM; also parent at LUKS. `establish_session` currently hard-codes `GUEST_PASSWORD`; add a parent path.

4. Do not “fix” SDDM with `Relogin=true` or a second user. Relogin skips the password after logout. A parent account violates one kid session.

5. Copy: stop saying a parent can always get in until SDDM works. Fix the stale parent-password gum header on later setup-form screens. Unencrypted first boot should not try a LUKS mapping.

6. After SDDM: either run parent-unlock before dying on kid faillock, or document the 120s wait. `even_deny_root` is already correct for sudo/polkit.

7. `omarchy-parent password` still does not update the LUKS parent slot. Fine if documented. Not the merge blocker.

8. Passwordless sudo / sudoless Docker menus are not child-gated. Enabling them requires the parent password, then the kid is passwordless root. Parent footgun, not a kid bypass.

9. Passing `test/acceptance.d/parent-test.sh` is not “parent can always reach a desktop.”

Happy to re-test a follow-up ISO.

elgevan:
### Testing feedback

Tested in a fresh QEMU/KVM install using:

- Omarchy PR [**Kids mode: child installs with a kid password and a parent password #9750**](https://github.com/omacom/omarchy/pull/9750) at `a6ebcde`
- Omarchy ISO PR [**Kids mode: ask who the computer is for and install the child profile #146**](https://github.com/omacom/omarchy-iso/pull/146) at `cbfea58`

A few thoughts:

- Does a child install need the same setup flow as a regular install? Full name and email are only used for Git, and the hostname could use a default. 

- WhatsApp was still available in the child environment. The default apps and shortcuts inherited by Child mode should probably be reviewed.

kenhara:
@peterholko @elgevan my understanding/two cents:

So, I think the parent/sudo account should be set first. When that's done, the hostname will be set and the parent will have the option to enter full name and email for their account. 

Then, a child account can be set up and both full name and email can be optional since they're for git. Hostname also skipped.

But I think it makes sense to do this in the same session. The kid screen show up after the parent account has been created. This isn't how it works right now but I think it makes sense. IDK. this is weird.

kenhara:
I think it would be odd to let kid be set up first, as non-sudo, and set the hostname. I have to sit down and think through it but that's not making sense to me at the moment. Unless you kick back to parent/sudo to complete the install during the bootup? idk.

peterholko:
I removed the Full Name and Email requirement, it can easily be done manually after installation.

hansent:
Setup omachi for my 10 year old today. Definitely would have loved to setup myself as su and then him as a user. Ended up resetting and setting it up just for him since this wasn't an option. 

Mainly just want to say +1 on the flow you have here; it's pretty much exactly what I wanted but didn't have today. 

Showed him a couple of keyboard commands and he picked it up way faster than I expected...already 10x happier about than he was with windows; especially given how easy steam and Minecraft were to install!

Maybe optionally allow adding multiple child accounts and have some account switcher for sibling purposes...although I can also see this turning into a couple of new computer setups for the other kids already 🤩

peterholko:
Patch: preserve the active Omarchy installation when parent commands elevate — [ea3f2255](https://github.com/peterholko/omarchy/commit/ea3f225541cb609e84cd9e4a57e7d90ebabfb646).

When `OMARCHY_PATH` points to a checkout outside `/usr/share/omarchy`, a bare command such as `omarchy-parent wifi kid --user kid` elevates itself through sudo. With `env_reset`, the elevated process previously lost that path and fell back to `/usr/share/omarchy`, potentially loading another installation's helpers or failing to find them.

The command now passes `OMARCHY_PATH` explicitly through `sudo env`, alongside the existing gum styling. Its existing default for direct sudo/root setup entry points remains intact.

Added `test/shell.d/parent-elevation-test.sh`. It runs the command through a sudo stub that clears the environment, using a custom installation path containing spaces. It verifies the elevated helper lookup, gum styling, and unchanged arguments. The regression fails on the previous PR code and passes with this patch. The test simulates the uid transition and touches only temporary fixtures; it does not change accounts or authenticate through the host's sudo/PAM.

peterholko:
Patch: serialize parent-settings writes and publish complete files atomically — [76bb1f2e](https://github.com/peterholko/omarchy/commit/76bb1f2efe2a64b2fb34ef64b752dfdec91645e9).

Default initialization and an explicit Wi-Fi change could race: `conf_document wifi parent` checked that the key was absent, `conf_set wifi kid` wrote the parent's choice, and the initializer then appended `wifi=parent`. Since reads use the last matching key, this silently replaced the explicit choice with the default. This reproduces using only the Wi-Fi control already included in this PR.

`conf_init`, `conf_set`, and `conf_document` now share one Bash writer. A persistent sibling `flock` covers the existence/key checks through publication. Complete contents are staged in the same directory, set to mode 0644, and renamed into place. Failed operations return failure and clean up their staging files; the subshell releases the lock while preserving the caller's traps and umask.

Added `test/shell.d/parent-config-test.sh`, covering the controlled concurrent Wi-Fi race, idempotent initialization, preserved comments/permissions, complete file generations for readers, and injected lock/content-generation/publication failures. The race test fails on the previous PR helper and passes with this patch.

Validation was local on macOS with Bash 5. The settings tests used real kernel `flock(2)` locks through a small local adapter for the descriptor-only util-linux `flock` invocation; that adapter is not part of the patch. The parent, parent-unlock, profile, and provision-owner checks also passed locally; the existing Linux user-namespace apply/password probes were skipped. No ISO or GitHub Actions runs were started.

kenhara:
While testing #11196 (which stacks on this branch) a few things surfaced —
not a full review, just flags with what I saw:

**1. The `parent-test.sh` sudo idiom caches nothing, so most of the file
never runs.** `sudo -S -k -v` validates but leaves no cached credential on
current sudo, so every subsequent `sudo -n` fails and the file dies at its
first `sudo -n` assertion ("sudo is configured to ask for root's password").
Verified live in a child VM: `sudo -S -v` caches, `sudo -S -k -v` doesn't.
I ran the remaining assertions manually on the same VM and they all pass
(per-account sudoers grant, polkit root auth, both PAM stacks,
kid-password-first, NOPASSWD scope, preserved sddm stack) — so it's the
idiom, not the install. Dropping the `-k` from the final validate should fix
it.

**2. Host-suite failures at this branch's head:**
- raw `command -v` in `bin/omarchy-parent` (bin-style-test)
- the `parent.conf` permissions check (`parent-test.sh:173` expects 644 —
  the kid-side tools read this policy — but it doesn't land that way)
- the unquoted `<<WIFI` heredoc feeding a privileged write embeds an
  install-time expansion (privileged-heredoc-test)

**3. `apply-lock-test.sh` fails inside the child VM only** — "the fully
hardened lock helper runs in an isolated root context" with `cp: command not
found` at the patched helper's line 23; passes on my host. I didn't chase it
further and it may be something about my guest setup, but worth a look if CI
starts hitting it.

Also: running the ISO harness at current quattro requires the greeter fix
from omarchy-iso#168 — possibly why the `--child` VM evidence hasn't been
rerun since the tagline change.

jankeesvw:
Opened peterholko/omarchy#2 against this branch with the labels @dhh asked for on X: Myself / My child / Another owner.

PR #9754, opened 2026-09-02
Prefer logind's LidClosed over raw ACPI for lid detection

### LINKED ISSUE #9753: Stale ACPI lid state turns external-monitor hotplug into clamshell mode, blanking the internal panel
## What happens

On some laptops the firmware's ACPI `_LID` method gets stuck reporting `closed` while the lid is open. When that happens, plugging in an external monitor blanks the internal panel: `omarchy-hyprland-monitor-clamshell` sees "lid closed + external active", writes the clamshell toggle, and disables eDP-1. Unplugging re-enables it, so the symptom looks like "my laptop screen dies whenever I connect my monitor". Workspaces also migrate to the external and back on each cycle, which adds to the confusion.

I hit this on a Razer Blade Stealth 13 (Late 2019). The desync is intermittent. In my captured case ACPI reported `closed` for over an hour of lid-open use, then recovered on its own:

```
$ cat /proc/acpi/button/lid/LID0/state
state:      closed                                # lid physically open

$ busctl get-property org.freedesktop.login1 /org/freedesktop/login1 \
    org.freedesktop.login1.Manager LidClosed
b false                                           # logind, correct

$ journalctl -b | grep -i "lid"
Sep 01 21:21:43 omarchy systemd-logind[744]: Lid opened.   # last real event
```

Hyprland's log at hotplug time shows the resulting disable:

```
drm: Modesetting DP-3 with 5120x2880@60.00Hz
drm: Disabling output eDP-1
drm: Connector eDP-1 enabledState changed true -> false
```

## Why

`omarchy-hw-laptop-closed` reads only `/proc/acpi/button/lid/*/state`. logind and libinput both derive lid state from the evdev `SW_LID` switch instead, with logind adding hardware quirk handling on top, which is why logind had the right answer the whole time. Omarchy already treats logind as the authority on lid policy elsewhere: it keeps logind's `HandleLidSwitch` defaults untouched, and `omarchy-system-lid-close`'s docked/undocked reasoning is written against logind's behavior. The detection script is the one place that trusts raw firmware state instead.

## Fix

Ask logind first, keep ACPI as the fallback:

- `busctl get-property org.freedesktop.login1 /org/freedesktop/login1 org.freedesktop.login1.Manager LidClosed`, and parse `b true`/`b false`.
- If busctl produces nothing (no running logind, e.g. installation chroots), fall back to the existing ACPI loop unchanged.

Pure bash, no new dependencies (busctl ships with systemd). PR incoming with a test that pins both branches.

## Blast radius

`omarchy-hw-laptop-closed` has five non-test call sites, so the change affects more than clamshell:

- `bin/omarchy-hw-clamshell` and `bin/omarchy-system-lid-close`: the display/lock paths this fixes.
- The PAM fingerprint gate (`auth [success=1 default=ignore] pam_exec.so quiet /usr/bin/omarchy-hw-laptop-closed` in `/etc/pam.d/sudo` and `/etc/pam.d/polkit-1`): runs as root at auth time. busctl talks to the system bus by default, so it works there; when D-Bus itself is unreachable the ACPI fallback answers, same as today.
- `shell/plugins/polkit/PolkitAgent.qml`: samples the command per auth request. Behavior unchanged apart from correctness.

The stuck-ACPI failure mode currently affects those consumers too: with ACPI stuck on `closed`, fingerprint auth is silently skipped on an open laptop, and the polkit dialog hides the fingerprint prompt. The fix corrects all of them at once.

bbishdotdev:
Fixes #9753.

`omarchy-hw-laptop-closed` trusted `/proc/acpi/button/lid/*/state` alone. Firmware `_LID` methods can report stale state. On my Razer Blade Stealth 13 (Late 2019), ACPI reported `closed` for over an hour of lid-open use, so every external-monitor hotplug entered clamshell mode and blanked the internal panel, while logind had the correct answer the whole time from the evdev lid switch:

```
$ cat /proc/acpi/button/lid/LID0/state
state:      closed                                # lid physically open

$ busctl get-property org.freedesktop.login1 /org/freedesktop/login1 \
    org.freedesktop.login1.Manager LidClosed
b false                                           # logind, correct
```

This asks logind first and keeps the ACPI loop as the fallback:

- `busctl get-property ... org.freedesktop.login1.Manager LidClosed`, parsed as `b true`/`b false`.
- Empty busctl output (no running logind, e.g. install chroots) falls through to the ACPI read, now behind an `OMARCHY_ACPI_LID_PATH` seam matching the other `hw-` probes (`OMARCHY_DRM_PATH`, `OMARCHY_USB_DEVICES_PATH`, ...).

No new dependencies; busctl ships with systemd.

`test/shell.d/hw-laptop-closed-test.sh` pins five behaviors from fixtures: logind-closed, logind-open despite stale ACPI (the bug), and the three ACPI fallback states (closed, open, no lid device). Against the old script the stale-ACPI scenario fails; against this change all five pass:

```
not ok - a lid logind reports closed is closed          # old script
ok - a lid logind reports closed is closed              # fixed
ok - a lid logind reports open is open despite stale ACPI
ok - without logind a closed ACPI lid is closed
ok - without logind an open ACPI lid is open
ok - no lid device at all means open
```

The source pin in `monitor-recovery-test.sh` is updated to match the new mechanism.

Callers affected beyond clamshell: the PAM fingerprint gate in `/etc/pam.d/sudo` and `/etc/pam.d/polkit-1` runs this as root at auth time (busctl uses the system bus by default; unreachable D-Bus falls back to ACPI as today), and `PolkitAgent.qml` samples it per auth request. The stale-ACPI failure mode mis-answers those consumers too, so they are corrected together.

Verified on the affected hardware: with ACPI stuck on closed and the lid open, the old script reported closed and the clamshell path disabled eDP-1 on hotplug; the new script follows logind and the panel stays up. `./test/cli` passes; `./test/shell` matches the base commit (the four failures on my machine fail identically on a clean checkout of the base and want a live Omarchy runtime).

### DIFF
--- a/bin/omarchy-hw-laptop-closed
+++ b/bin/omarchy-hw-laptop-closed
@@ -3,7 +3,22 @@
 # omarchy:summary=Returns true when the laptop lid is closed
 # omarchy:hidden=true
 
-for state in /proc/acpi/button/lid/*/state; do
+# logind reads the lid switch through evdev with hardware quirk handling, so
+# its answer beats raw ACPI, whose _LID method can report stale state.
+lid_closed=$(busctl get-property org.freedesktop.login1 /org/freedesktop/login1 org.freedesktop.login1.Manager LidClosed 2>/dev/null)
+if [[ -n $lid_closed ]]; then
+  if [[ $lid_closed == "b true" ]]; then
+    exit 0
+  else
+    exit 1
+  fi
+fi
+
+# Fall back to ACPI for contexts without a running logind, such as install
+# chroots.
+lid_path="${OMARCHY_ACPI_LID_PATH:-/proc/acpi/button/lid}"
+
+for state in "$lid_path"/*/state; do
   [[ -r $state ]] || continue
   [[ $(< "$state") == *"closed"* ]] && exit 0
 done
--- a/test/shell.d/hw-laptop-closed-test.sh
+++ b/test/shell.d/hw-laptop-closed-test.sh
@@ -1,0 +1,81 @@
+#!/bin/bash
+
+set -euo pipefail
+
+source "$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd)/base-test.sh"
+
+laptop_closed="$ROOT/bin/omarchy-hw-laptop-closed"
+tmpdir=$(mktemp -d)
+trap 'rm -rf "$tmpdir"' EXIT
+
+lid_path="$tmpdir/lid"
+
+# logind's LidClosed is the contract whenever logind is available, because it
+# reads the lid switch through evdev while raw ACPI can report stale state.
+# Each scenario pins what busctl reports and records the resulting exit.
+setup_busctl() {
+  mock_bin="$tmpdir/$1/bin"
+  mkdir -p "$mock_bin"
+
+  cat >"$mock_bin/busctl" <<SH
+#!/bin/bash
+$2
+SH
+  chmod +x "$mock_bin/busctl"
+}
+
+write_lid_state() {
+  rm -rf "$lid_path"
+  mkdir -p "$lid_path/LID0"
+  printf 'state:      %s\n' "$1" >"$lid_path/LID0/state"
+}
+
+run_laptop_closed() {
+  if OMARCHY_ACPI_LID_PATH="$lid_path" PATH="$mock_bin:$PATH" "$laptop_closed"; then
+    result=0
+  else
+    result=$?
+  fi
+}
+
+setup_busctl closed 'echo "b true"'
+write_lid_state open
+run_laptop_closed
+
+(( result == 0 )) ||
+  fail "a lid logind reports closed is closed" "exit: $result"
+pass "a lid logind reports closed is closed"
+
+# The stale-ACPI case this exists for: firmware says closed, logind knows
+# better from evdev.
+setup_busctl open 'echo "b false"'
+write_lid_state closed
+run_laptop_closed
+
+(( result == 1 )) ||
+  fail "a lid logind reports open is open despite stale ACPI" "exit: $result"
+pass "a lid logind reports open is open despite stale ACPI"
+
+# Without a running logind, as in an install chroot, busctl produces nothing
+# and the ACPI fallback decides.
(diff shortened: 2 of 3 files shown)


cristim:
Confirming this on real Apple Silicon hardware, since the failure it fixes is invisible without a lid to close.

**Machine:** MacBookPro17,1 (M1), Asahi, aarch64. No `/proc/acpi` at all, so the pre-fix helper's glob matches nothing and it reports the lid open permanently.

**logind's property is present and live here:**

```
$ busctl get-property org.freedesktop.login1 /org/freedesktop/login1 \
    org.freedesktop.login1.Manager LidClosed
b false
```

It is declared `emits-change`, and `HandleLidSwitch=suspend` / `HandleLidSwitchDocked=ignore` are both set as this PR's approach assumes.

**Closed case, against a real lid close rather than a stub.** Polling the property and the helper together at 5 Hz across an actual close-and-reopen:

```
  1 CLOSED helper=0     <- 20:27:56
126 open   helper=1     <- control samples either side
```

The `helper=0` line is the helper reporting the lid closed, which on this hardware it has never been able to do.

**The underlying source**, for the record: the lid is `input0`, "Apple SMC power/lid events" on `macsmc-input`, with `B: SW=1` (`SW_LID`). Reading it directly is not an option for these callers because `/dev/input/event0` is `root:input` and they run as the user, which is what makes going through a bus property the right call.

Preferring logind over ACPI rather than the other way round also looks correct to me beyond the no-ACPI case: I wrote the inverse first and it is weaker, since it cannot address the stale `_LID` behaviour in #9753.

One optional suggestion for the test, take it or leave it: two cases that fall out of this shape and are cheap to add are *logind outranks ACPI when both answer and disagree* (pins the precedence this PR establishes), and *neither source answers* (asserts it reports not-closed and stays quiet, which matters because `omarchy-setup-security-fingerprint` runs this helper from a PAM stack on every sudo).

bjarneo:
Automated duplication check: this pull request looks similar to #7951, which covers the same lid detection via logind. I keep that one open and close this one to consolidate review.

Issue #9029, opened 2026-08-29
Menu blocks on-screen keyboard input: Exclusive keyboard focus routes all pointer input to the menu (touchscreen devices)


ekollof:
### Summary

On touchscreen devices, opening the Omarchy menu makes any on-screen keyboard unresponsive: the first tap on the OSK dismisses the menu, and no keystrokes reach the menu's search field while it is open. The OSK keeps working normally in regular windows before and after.

### Environment

- Omarchy 4.0.1-1, Hyprland 0.56.2 (Arch packages)
- Hardware: GPD Pocket 4 (touchscreen, no tablet-mode sensor); reproduces in both portrait and landscape
- OSK under test: wvkbd 0.20 (layer-shell, `zwp_virtual_keyboard_v1`); wtype reproduces the same delivery path
- Debug log: see [comment below](https://github.com/omacom/omarchy/issues/9029#issuecomment-5465212545) (`omarchy debug --no-sudo --print`)

### Steps to reproduce

1. Show an on-screen keyboard (e.g. `wvkbd-mobintl`, or SUPER+SHIFT+K with a wvkbd autostart)
2. Open the Omarchy menu (SUPER+SPACE, `omarchy menu summon`, or the bar icon)
3. Tap keys on the on-screen keyboard

### Observed

- The menu closes on the first OSK tap
- While the menu is open, OSK taps never reach the keyboard (no key highlight, no text); the OSK remains fully functional in normal windows

### Expected

- The menu stays open and OSK keystrokes type into the menu's search filter, like a physical keyboard does

### Root cause

`plugins/menu/Menu.qml` creates its full-screen layer surface with:

```qml
WlrLayershell.keyboardFocus: WlrKeyboardFocus.Exclusive
exclusionMode: ExclusionMode.Ignore
```

Per the comment in `Ui/KeyboardPanel.qml`, Hyprland routes **every pointer event to an Exclusive-keyboard layer surface** (regardless of where the touch lands). So while the menu is open, OSK taps are delivered to the menu's own scrim (`MouseArea { onClicked: root.cancel() }`) and dismiss it. The `ExclusionMode.Ignore` additionally makes the menu overlap the OSK's reserved exclusive zone.

Instrumenting wvkbd with `-D` confirms the OSK itself emits all keystrokes correctly; delivery breaks because the pointer input is captured by the menu surface.

### Validated fix

[...]

PR #9757, opened 2026-09-02
Let on-screen keyboards reach bar panels

### LINKED ISSUE #9756: Bar panels dismiss when typing on an on-screen keyboard (wifi password, etc.)
### Summary

On a touchscreen, opening a bar popdown (Wi-Fi, Bluetooth, …) and typing on a layer-shell on-screen keyboard closes the panel on the first key tap. The passphrase field never receives the text. Same mechanism as #9029, but in `Ui/KeyboardPanel.qml` rather than the menu — **#9135 does not fix this**.

### Environment

- Omarchy 4.0.2-1, Hyprland 0.56.2
- Hardware: GPD Pocket 4 (touchscreen)
- OSK: omarchy-osk (`ekollof-osk` layer-shell, `WlrLayer.Top`, exclusive zone = keyboard height, `keyboardFocus: None`) — wvkbd would hit the same overlay

### Steps to reproduce

1. Show an on-screen keyboard (bottom exclusive zone)
2. Open the network bar applet and tap a secured SSID so the password field appears
3. Tap keys on the OSK

### Observed

- The Wi-Fi popdown closes on the first OSK tap
- Nothing is typed into the passphrase field

### Expected

- The panel stays open and OSK keystrokes go into the focused `TextField` (PanelKeyCatcher is already `blocked` while `passwordSsid` is set)

### Root cause

`shell/Ui/KeyboardPanel.qml` (used by network, bluetooth, audio, power, …) is a full-screen Overlay with:

```qml
exclusionMode: ExclusionMode.Ignore
WlrLayershell.keyboardFocus: open
  ? (focusPrimed ? WlrKeyboardFocus.OnDemand : WlrKeyboardFocus.Exclusive)
  : WlrKeyboardFocus.None
```

`Ignore` sets `exclusive_zone = -1`, so the overlay covers the OSK's reserved band. OSK taps are synthetic pointer clicks; Hyprland hit-tests the Overlay (`omarchy-keyboard-panel`) instead of the OSK (`Top`). The click is outside the card → `dismissArea.onClicked` → `close()`.

The 75ms Exclusive prime is the same pointer-capture issue as #9029, but after that the panel is already OnDemand — **covering the exclusive zone is enough to dismiss**.

#9029 / #9135 only change `Menu.qml`. KeyboardPanel is shared by every bar popdown.

### Fix

`ExclusionMode.Auto` on KeyboardPanel (and the other-output dismiss twins) so the overlay no longer covers the OSK exclusive zone. Keep the Exclusive→OnDemand prime so keyboard-summoned panels still take focus.

Related: #9029, #9135.

ekollof:
Fixes #9756. Same class of bug as #9029, but in `Ui/KeyboardPanel.qml` (wifi / bluetooth / every other bar popdown). #9135 only changes the menu.

## Problem

`ExclusionMode.Ignore` stretches the full-screen KeyboardPanel overlay over an on-screen keyboard's exclusive zone. The first OSK tap hits `dismissArea` and closes the panel, so a wifi passphrase never gets typed.

KeyboardPanel already settles on `OnDemand` after a short Exclusive prime, so Exclusive pointer capture is not the lasting cause here — covering the exclusive zone is.

## Fix

`ExclusionMode.Auto` on the panel and the other-output dismiss twins. The Exclusive → OnDemand prime is unchanged so keyboard-summoned panels still take focus.

## Verification

- `bash test/shell.d/keyboard-panel-test.sh`


ekollof:
Rebased onto current quattro (2fbac0c8). This pushes the local refinement (position via root.x/root.y, no windowOrigin derivation), superseding the previously pushed version of the second commit. keyboard-panel-test.sh passes.

ekollof:
Follow-up fix: anchor inset panels to the bar edge instead of the window position (7045244c).

The inset-overlay change parked every popdown at the top-left. Root cause: a Quickshell PanelWindow exposes no x/y, so translating the card origin via `x - root.x` / `y - root.y` evaluated to undefined and NaNed the origin for all panels (verified live: correct anchorScreenPos, but origin=NaN,NaN). It also trusted the pre-map 100x100 window geometry for the card budgets, collapsing cards to minimum width.

The fix derives the surface origin from the bar edge once the backing window maps (originX/originY, gated by insetOverlay), with fullscreen/pre-map behavior identical to before. The dismiss-area bar-click forwarding translates window-local input through the same origin.

Verified live on a single-monitor top bar (bluetooth under its icon at full width, clock centered, dismissal works), plus keyboard-panel-test.sh (now with regression guards) and the bar/clock/agents suites.

Issue #9421, opened 2026-08-31
SDDM greeter's Hyprland config ignores system keyboard layout, causing password mismatches for non-US layouts


winless-code:
Title: SDDM greeter's Hyprland config ignores system keyboard layout, causing password mismatches for non-US layouts

## What happened
The SDDM Wayland greeter runs its own minimal Hyprland instance via
`/usr/share/sddm/hyprland.lua` (referenced from
`/etc/sddm.conf.d/10-wayland.conf` as `CompositorCommand`). This config has
no `input` block, so it falls back to Hyprland's default keyboard layout
(US), regardless of the system's actually configured layout
(`localectl` correctly reports `de` for both console and X11).

This is invisible for most passwords, but any password containing a
character that sits on a different key between the US and configured
layout (e.g. `_` on a US vs. German keyboard) gets typed incorrectly at
the greeter — while working fine on a TTY or inside the real Hyprland
session, since those correctly use the system layout.

## What was expected
The SDDM greeter should use the system's configured keyboard layout
(as reported by `localectl status`), matching what the TTY and the
user's actual Hyprland session use — so password entry is consistent
across all three.

## Steps to reproduce
1. Install Omarchy on a system with a non-US keyboard layout (e.g. `de`)
   configured via `localectl` / `/etc/vconsole.conf`.
2. Set a password containing a character that differs in position between
   the US and configured layout (e.g. `_`, which is Shift+Minus on US but
   a different key combination on German layout).
3. Reboot to the SDDM login screen and enter the password.
4. Login fails — the greeter's default US-layout Hyprland instance
   produces the wrong character(s). Password verified independently
   (e.g. via `openssl passwd -6` against `/etc/shadow`) confirms it's
   correct; entering the same password on a TTY works.

## Workaround
Manually add an `input` block to `/usr/share/sddm/hyprland.lua`:
```lua
hl.config({
  input = { kb_layout = 'de' },
  ...
})
```
This is not durable: the file is owned by the `omarchy-settings` package
and has no pacman backup/pacnew protection, so it is silently overwritten
(and the bug reintroduced) on the next `omarchy update`.

## System details
- Omarchy version: 4.0.2-1
- Kernel: 7.1.9-arch1-2
- CPU: 11th Gen Intel(R) Core(TM) i5-1135G7 @ 2.40GHz
- GPU: Intel TigerLake-LP GT2 [Iris Xe Graphics] (i915 driver)
[...]

PR #9762, opened 2026-09-02
Give the SDDM greeter the system keyboard layout

### LINKED ISSUE #6880: SDDM greeter always uses US keyboard layout, ignoring the system layout

Yves848:
Fixes #6880. Also covers #9421, which reports the same bug independently.

## The bug

The SDDM Wayland greeter runs its own Hyprland instance:

```ini
# etc/sddm.conf.d/10-wayland.conf
CompositorCommand=start-hyprland -- --config /usr/share/sddm/hyprland.lua
```

`default/sddm/hyprland.lua` set only `misc` and `animations` — no `input` block — so Hyprland fell back to its built-in `us` default. A password set under any other layout cannot be typed at the login prompt, while the LUKS prompt (vconsole `KEYMAP`) and the user session (`default/hypr/input.lua`, which reads `XKBLAYOUT`) both get it right.

It is easy to miss because autologin skips the greeter at boot: the first time most users meet it is after an explicit logout, and then they simply cannot log back in.

## The fix

`default/hypr/input.lua` already resolved the layout from `/etc/vconsole.conf`. That block moves to `default/hypr/keyboard.lua`, which returns `{ layout, variant, options }`, and both configs read it — so the session and the greeter cannot drift apart. No behaviour change for the session; the test below pins that.

The greeter is started by sddm-helper rather than the user session, so it never runs `default/hypr/bootstrap.lua`. It sets up just the module path it needs, with the same `os.getenv("OMARCHY_PATH") or "/usr/share/omarchy"` idiom `config/hypr/hyprland.lua` already uses (deliberately not `bootstrap.lua`, which dereferences `$HOME` — not something to rely on in the greeter's environment).

The non-Latin fallback carries over to the greeter, and matters more there than in the session: passwords are Latin, so a non-Latin-only layout would leave the correct password untypeable. Same reasoning as the `vconsole.conf` guard in `etc/mkinitcpio.conf.d/omarchy_hooks.conf`.

`kb_model` and `numlock_by_default` are deliberately left alone — out of scope for this bug.

## Verification

`test/shell.d/hyprland-keyboard-layout-test.sh` gains four greeter assertions alongside the existing session ones, driven through the same stubbed `vconsole.conf`. The greeter is loaded the way SDDM loads it — by path, with no module path prepared for it — so the `package.path` line is exercised rather than assumed. The non-Latin-list sync check now reads `default/hypr/keyboard.lua`.

```
ok - greeter falls back to us without vconsole.conf
ok - greeter follows the system layout
ok - greeter keeps the variant
ok - greeter keeps a latin layout reachable for passwords
```

`./test/all`: 222 of 226 shell test files pass. The 4 failures (`config`, `snapper`, `unowned-system-paths`, `bar-icon-geometry`) fail identically on an unmodified checkout here — the first three want an `omarchy-pkgs` checkout, the last is sub-pixel monitor geometry.

Beyond the stubs, the greeter config was run as a real nested Hyprland instance (headless backend) on a `be` system:

```
$ hyprctl -i <nested> getoption input:kb_layout
str: be
$ hyprctl -i <nested> devices -j | jq '.keyboards[].active_keymap'
"Belgian"
$ hyprctl -i <nested> configerrors
(empty)
```

with `misc.disable_hyprland_logo` and `animations.enabled` still resolving to the values the file already set.

## Note for maintainers

The two open reports carry workarounds that no longer hold: #9421 edits the packaged `/usr/share/sddm/hyprland.lua`, which `omarchy update` silently reverts, and #6880 predates the move to Lua (it uses `source =` against the old `hyprland.conf`) and hardcodes one layout. Worth closing both on this rather than leaving those recipes as the answer.

Reported on Omarchy 4.0.2-1, Hyprland 0.56.2-1, sddm 0.21.0-7, layout `be`.

### DIFF
--- a/default/hypr/input.lua
+++ b/default/hypr/input.lua
@@ -1,58 +1,13 @@
 -- https://wiki.hypr.land/Configuring/Basics/Variables/#input
 
-local function read_vconsole()
-  local values = {}
-  local file = io.open("/etc/vconsole.conf", "r")
-  if not file then
-    return values
-  end
-
-  for line in file:lines() do
-    local key, value = line:match("^%s*([%w_]+)%s*=%s*(.-)%s*$")
-    if key and value then
-      value = value:gsub("%s+#.*$", "")
-      value = value:gsub('^"(.*)"$', "%1")
-      value = value:gsub("^'(.*)'$", "%1")
-      values[key] = value
-    end
-  end
-
-  file:close()
-  return values
-end
-
--- Layouts that can't type Latin letters. Keep in sync with the list in
--- etc/mkinitcpio.conf.d/omarchy_hooks.conf.
-local non_latin_layouts =
-  " af am ara bd bg by et ge gr il in iq ir kg kh kz la lk mk mm mn mv np rs ru sy th tj ua "
-
-local vconsole = read_vconsole()
-
-local kb_layout = vconsole.XKBLAYOUT or "us"
-local kb_variant = vconsole.XKBVARIANT or ""
--- CapsLock is the compose key, so Caps Lock itself has to live somewhere else.
--- Both Shifts together is the usual home for it, but it's easy to hit by
--- accident while typing. The _cancel variant sets Caps Lock the same way and
--- releases it on the next lone Shift, so a misfire clears itself.
-local kb_options = "compose:caps,shift:both_capslock_cancel"
-
--- Hyprland resolves keybindings against the first entry in kb_layout, not the
--- layout that's currently active, so Omarchy's Latin-keysym bindings (SUPER + W
--- and friends) only fire when a Latin layout leads. Installing with a non-Latin
--- one would otherwise leave the desktop unusable.
-if non_latin_layouts:find(" " .. kb_layout:match("^[^,]*") .. " ", 1, true) then
-  kb_layout = "us," .. kb_layout
-  kb_variant = "," .. kb_variant
-  -- Reach the original layout with Left Alt + Right Alt.
-  kb_options = kb_options .. ",grp:alts_toggle"
-end
+local keyboard = require("default.hypr.keyboard")
 
 hl.config({
   input = {
-    kb_layout = kb_layout,
-    kb_variant = kb_variant,
+    kb_layout = keyboard.layout,
+    kb_variant = keyboard.variant,
     kb_model = "",
-    kb_options = kb_options,
+    kb_options = keyboard.options,
     kb_rules = "",
     follow_mouse = 1,
     sensitivity = 0,
--- a/default/hypr/keyboard.lua
+++ b/default/hypr/keyboard.lua
@@ -1,0 +1,60 @@
+-- Keyboard layout resolved from the system's /etc/vconsole.conf.
+-- Shared by the user session (default/hypr/input.lua) and by the SDDM greeter
+-- (default/sddm/hyprland.lua). They are separate Hyprland instances, but a
+-- password is typed at one and set from the other, so both have to agree on
+-- the layout.
+
+local function read_vconsole()
+  local values = {}
+  local file = io.open("/etc/vconsole.conf", "r")
+  if not file then
+    return values
+  end
+
+  for line in file:lines() do
+    local key, value = line:match("^%s*([%w_]+)%s*=%s*(.-)%s*$")
+    if key and value then
+      value = value:gsub("%s+#.*$", "")
+      value = value:gsub('^"(.*)"$', "%1")
+      value = value:gsub("^'(.*)'$", "%1")
+      values[key] = value
+    end
+  end
+
+  file:close()
+  return values
+end
+
+-- Layouts that can't type Latin letters. Keep in sync with the list in
+-- etc/mkinitcpio.conf.d/omarchy_hooks.conf.
+local non_latin_layouts =
+  " af am ara bd bg by et ge gr il in iq ir kg kh kz la lk mk mm mn mv np rs ru sy th tj ua "
+
+local vconsole = read_vconsole()
+
+local kb_layout = vconsole.XKBLAYOUT or "us"
+local kb_variant = vconsole.XKBVARIANT or ""
+-- CapsLock is the compose key, so Caps Lock itself has to live somewhere else.
+-- Both Shifts together is the usual home for it, but it's easy to hit by
+-- accident while typing. The _cancel variant sets Caps Lock the same way and
+-- releases it on the next lone Shift, so a misfire clears itself.
+local kb_options = "compose:caps,shift:both_capslock_cancel"
+
+-- Hyprland resolves keybindings against the first entry in kb_layout, not the
+-- layout that's currently active, so Omarchy's Latin-keysym bindings (SUPER + W
+-- and friends) only fire when a Latin layout leads. Installing with a non-Latin
+-- one would otherwise leave the desktop unusable. The greeter needs the same
+-- fallback for a different reason: passwords are Latin, so a non-Latin-only
+-- layout would leave the correct one untypeable at the login prompt.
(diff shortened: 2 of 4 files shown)


IridescentRose:
Also closes related ticket: https://github.com/omacom/omarchy/issues/9859

bjarneo:
Automated duplication check: this pull request looks similar to #6896, which covers the same SDDM keyboard layout. I keep that one open and close this one to consolidate review.

PR #9769, opened 2026-09-02
Publish Xft.dpi to XWayland so fcitx5's candidate window scales

wulujia:
## Problem

On a HiDPI laptop (2x scale), typing Chinese with fcitx5 into an XWayland app such as Feishu shows the candidate window at half size. The same input into a Wayland-native app (terminal, Chromium) is fine.

## Cause

Omarchy runs XWayland with `force_zero_scaling`, so X11 windows draw at 1x and each client has to scale itself. GTK and Qt apps do that from `GDK_SCALE` and their own variables. Plain X11 clients read `Xft.dpi` from the root window's `RESOURCE_MANAGER` property, and nothing in the session sets it: Omarchy ships no `.Xresources`, does not install xrdb, and the property is empty on a fresh install.

fcitx5 draws its candidate window as an X11 window whenever the focused app is XWayland and sizes the font from `Xft.dpi` ([xcbui.cpp](https://github.com/fcitx/fcitx5/blob/master/src/ui/classic/xcbui.cpp), `forcedDpi`). With the property empty it assumes 96 dpi on a 192 dpi screen. Since Omarchy ships fcitx5 in every session and the manual points CJK users at it, every one of them with a 2x display hits this the first time they type into Feishu, WeChat, or any Electron app that fell back to X11.

## Fix

- `bin/omarchy-hyprland-xwayland-dpi` (hidden) writes `Xft.dpi = 96 * GDK_SCALE` into `RESOURCE_MANAGER` with xprop, which gtk3 already pulls in through at-spi2-core. It keeps other resources a user loaded with xrdb, leaves a user-set `Xft.dpi` alone, and replaces only the value it wrote itself (marked with a comment line), so a changed `omarchy_gdk_scale` takes effect on the next login.
- `omarchy-fcitx5.service` runs it as `ExecStartPre=-` so the value is published before the first candidate window. The unit already has `GDK_SCALE` and `DISPLAY` from the imported session environment. The leading `-` keeps a failed xprop from taking the input method down.
- A migration reloads the user manager and runs the command once. fcitx5 re-reads `RESOURCE_MANAGER` on `PropertyNotify`, so the running session is repaired without restarting fcitx5.

The 1x-monitor trade-off is the same one Omarchy already accepts for `GDK_SCALE` on X11 windows.

## Verification

- `test/shell.d/xwayland-dpi-test.sh` covers the scale math, the no-DISPLAY no-op, keeping foreign resources, respecting a user-set value, replacing its own, and the unit wiring.
- Ran on Omarchy 4.0.2 with fcitx5 5.1.21 and a 2880x1800 display at scale 2: after the unit change the property reads `Xft.dpi: 192` and running the command twice is idempotent.
- `./test/all` passes apart from `bar-icon-geometry-test.sh`, which fails on this machine before the change too because it launches Quickshell inside a live session.

Back to draft: with `GDK_SCALE=2`, GTK3 reads `Xft.dpi` as an unscaled value and multiplies it by the scale, so GTK X11 apps and Electron apps (via Chromium's GTK backend, `resolution * scale / 96`) render at 4x. Feishu doubled in size. Needs a different mechanism.

### DIFF
--- a/bin/omarchy-hyprland-xwayland-dpi
+++ b/bin/omarchy-hyprland-xwayland-dpi
@@ -1,0 +1,45 @@
+#!/bin/bash
+
+# omarchy:summary=Publish the desktop scale to XWayland clients as Xft.dpi
+# omarchy:hidden=true
+
+set -euo pipefail
+
+# Omarchy runs XWayland with force_zero_scaling, so X11 windows draw at 1x and
+# every client has to scale itself. GTK and Qt apps read GDK_SCALE and their
+# own variables for that, but plain X11 clients read Xft.dpi from the root
+# window's RESOURCE_MANAGER property, and nothing in the session set it. The
+# client every CJK user meets is fcitx5's candidate window, which it draws as
+# an X11 window whenever the focused app is XWayland: at 96 dpi it came up at
+# half size on a 2x display. Publish 96 * GDK_SCALE, the factor the rest of the
+# desktop already applies to X11 windows.
+
+[[ -n ${DISPLAY:-} ]] || exit 0
+
+scale=${GDK_SCALE:-1}
+if [[ ! $scale =~ ^[0-9]+$ ]]; then
+  scale=1
+fi
+dpi=$((96 * scale))
+
+marker="! Xft.dpi is managed by omarchy-hyprland-xwayland-dpi"
+
+# xprop prints the property as a C string literal; turn it back into text so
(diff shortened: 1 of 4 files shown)


wulujia:
Closing: with GDK_SCALE=2, GTK3 multiplies Xft.dpi by the scale, so this made GTK X11 apps and Electron apps render at 4x. The fix belongs in fcitx5, which already knows when it runs on XWayland; I'll take it there.

PR #9782, opened 2026-09-02
feat(lock): add Howdy face authentication, visual indicator, and setup wizard

edgseu:
## Summary

Adds Howdy (IR/webcam) face authentication to the Omarchy lock screen, wired to the real Howdy stack: howdy-git from the AUR and its compiled /lib/security/pam_howdy.so module.

Face authentication runs as a dedicated PAM context (omarchy-lock-howdy) in parallel with password and fingerprint — whichever matches first unlocks the desktop. An interactive setup wizard (omarchy setup security face) installs, configures, enrolls, and enables it; a matching remove wizard tears everything down, including enrolled biometric models.

## Changes

- **`bin/omarchy-setup-security-face`**: installs howdy-git (+ v4l-utils) via `omarchy-pkg-add`, aborting before anything is changed if the AUR build fails. Tunes whatever config home this Howdy build uses — probing /etc/howdy, /lib/security/howdy, and /usr/local/etc/howdy, since the location moved between builds (master r592+ uses /etc/howdy) — with detection notices off, timeout 3s, IR-friendly dark_threshold = 90, and face snapshots disabled (a documented spoofing hole). Idempotent: an existing face model skips enrollment so re-running to repair PAM never adds duplicate models. Mirrors the fingerprint enrollment flow's hardware checks: verifies the camera actually delivers frames before asking for a face, verifies enrollment stored a face model before writing any PAM, and aborts cleanly otherwise. Normalizes config/model paths after enrollment, since some Howdy builds write to /usr/local/etc/howdy while the PAM module only reads /lib/security/howdy. Writes /etc/pam.d/omarchy-lock-howdy (`auth sufficient /lib/security/pam_howdy.so`) only after successful enrollment and only when the module file exists.
- **`bin/omarchy-remove-security-face`**: removes the PAM context, drops howdy-git (or legacy howdy), disables the linux-enable-ir-emitter unit, and deletes enrolled face models (biometric data) from both known locations.
- **`shell/plugins/lock/Service.qml`**: omarchy-lock-howdy PAM context running in parallel with password and fingerprint; 5-second attempt timeout (battery guard); face attempts rate-limited by a cooldown so motion wake cannot turn every mouse movement into a camera-on PAM attempt; resume/lid-open detection re-triggers scanning and only runs when face auth is configured.
- **`shell/plugins/lock/LockView.qml`**: animated face indicator (pulsing while scanning) inside the field, a single hover area with motion wake (no click-swallowing overlay), error text keeps its error styling even while a parallel face scan runs, and indicator spacing is measured from the rendered glyphs.
- **`bin/omarchy-hw-face`**: hardware detector gating the wizard (exit codes per the hw- convention) — sysfs-precise: USB webcams matched by their uvcvideo driver binding, embedded MIPI/PCIe sensors by name, never just any /dev/video node.
- **`shell/plugins/lock/manifest.json`**: documents the face PAM flow.
- **`test/shell.d/lock-face-test.sh`**: 32 contract assertions — the real module and config paths, no phantom module references, howdy-git packaging (never plain howdy), path normalization, snapshot hardening, camera-frame and enrollment verification, complete teardown, the attempt cooldown, and the single hover area.

## Notes for reviewers

- The Howdy AUR dependency chain (python-dlib, opencv) occasionally breaks after Python version bumps; the wizard fails loudly and safely in that case. A pinned PKGBUILD in omarchy-pkgs would be the robust follow-up.
- Hardware behavior (IR emitters, camera detection, enrollment, the PAM conversation) needs a manual pass on real machines; the automated tests are contract-level by design.

## Verification

- `test/shell.d/lock-face-test.sh` — 32/32 assertions pass.
- `test/cli` — passes (116 assertions).
- `test/shell.d/bin-style-test.sh` — passes.


bjarneo:

edgseu:
Hey @bjarneo,

I understand the automated duplicate check consolidated review into #8336, but I would like to ask you to consider reopening this PR. This implementation is not a simple duplicate—it was built from the ground up to address several critical architectural, security, and performance issues present in #8336, and it would serve Omarchy much better:

### 1. Strict Security Boundaries
- **No sudo modification**: #8336 injects Howdy into `/etc/pam.d/sudo`. This PR confines PAM strictly to the lock screen context (`/etc/pam.d/omarchy-lock-howdy`).
- **Fail-safe PAM stack**: Uses `auth sufficient` (rather than `auth required`, which can break fallback auth) and verifies `/lib/security/pam_howdy.so` exists on disk before enabling the configuration.
- **Biometric privacy on removal**: Tears down enrolled face models (`rm -rf .../models`) and disables the `linux-enable-ir-emitter` systemd unit during removal, ensuring user biometric templates are never left behind.

### 2. Hardware Detection & Path Robustness
- **Dedicated `omarchy-hw-face` detector**: Follows the Omarchy `omarchy-hw-*` architecture and inspects sysfs (`/sys/class/video4linux`) for `uvcvideo` bindings and sensor names, preventing false-positive matches on capture cards, tuners, or loopback nodes.
- **Pre-flight frame & model verification**: Verifies that the camera can actually stream frames via `v4l2-ctl` before enrollment, checks that a model was saved before writing PAM, and normalizes paths across Howdy package revisions (`/etc/howdy`, `/lib/security/howdy`, `/usr/local/etc/howdy`).
- **Hardened settings**: Disables unencrypted face snapshots (`capture_failed = false`, `capture_successful = false`) and configures responsive IR thresholds.

### 3. Lock Screen Performance & Battery Protection (`Service.qml`)
- **Zero-fork reactive watching**: Uses Quickshell's native `FileView` instead of spawning `bash -c` subshells on lock/wake events.
- **Attempt timeout & cooldown**: Includes a 5-second attempt timeout so a stalled camera driver or recognizer cannot lock up the PAM context indefinitely, plus a 2-second cooldown to prevent mouse motion wake events from repeatedly re-powering the camera and draining battery.
- **Lid-open / resume detection**: Uses clock-delta monitoring to automatically trigger face scanning when resuming from sleep or opening the laptop lid.

### 4. Polished UI/UX (`LockView.qml`)
- **Symmetric layout**: Groups biometric indicators in a `Row` on the right edge, matching Hyprlock/Omarchy design conventions and preventing the input field margins from becoming distorted.
- **Active visual feedback**: Includes a smooth pulse animation, accent color highlighting, and dynamic "Scanning face…" status text while face authentication is running, with error messages properly prioritized.

### 5. Automated Testing & Clean Merge
- Comes with **32 contract assertions** in `test/shell.d/lock-face-test.sh` covering security, package targets, path normalization, teardown, cooldown, and glyph metrics.
- Merges cleanly on top of `quattro` without any conflicts.

Given the robustness, privacy, and performance improvements here, could we please reopen #9782 for review?

PR #9783, opened 2026-09-02
Fix Windows VM helper rejecting setgid source directories

### LINKED ISSUE #9334: Windows VM 4.0.2: launch fails silently after polkit auth (setgid chmod check can never pass)
> **Edit:** the original root-cause analysis below blamed a kernel regression. That was wrong, and I've rewritten this post. The bit is GNU `chmod` semantics, not the kernel — thanks to @murilocb for independently landing on the same conclusion in [a comment](https://github.com/omacom/omarchy/issues/9334). The failure itself and the code path are unchanged.

Windows VM launch silently fails after polkit auth on 4.0.2: the privileged mount-source check can never pass once the source dir has the setgid bit, because the `chmod` the script uses cannot clear it.

## What happened

After upgrading 4.0.1 → 4.0.2, launching the Windows VM (Apps menu → Windows, `omarchy-windows-vm launch`) shows the polkit password dialog and then nothing happens. No error, no window. Every retry asks for the password again with no result.

The elevated step exits non-zero and all of its stderr is discarded, because the desktop entry runs with `Terminal=false` — so from the user's perspective the password is accepted and then nothing happens at all.

## Root cause

`prepare_caller_mounts()` hardens both mount sources by running `chmod 0700` on the pinned fds and then requires the resulting mode to be exactly `700`:

```bash
chmod 0700 -- "/proc/$BASHPID/fd/$storage_fd" "/proc/$BASHPID/fd/$shared_fd" || { ...; return 1; }
[[ $storage_mode != 700 || $shared_mode != 700 ]] && { ...; return 1; }
```

**GNU `chmod` preserves a directory's setuid/setgid bits for numeric modes shorter than five octal digits.** This is documented, long-standing coreutils behavior. `chmod 0700` is therefore incapable of producing `700` from `2700`, and the check that follows can never pass. Measured on this box (`chmod (GNU coreutils) 9.11`, kernel `7.1.9-arch1-2`, btrfs):

| command on a `2777` directory | resulting mode |
| --- | --- |
| `chmod 700` | `2700` |
| `chmod 0700` | `2700` ← what the script does |
| `chmod 00700` | `700` |
| `chmod g-s` | `777` |

It behaves the same through the `/proc/$BASHPID/fd/N` indirection the script actually uses:

```
$ exec {fd}<~/Windows                          # mode 2777
$ chmod 0700 -- "/proc/$BASHPID/fd/$fd"
$ stat -Lc %a "/proc/$BASHPID/fd/$fd"          # 2700  → != 700 → return 1
```

Deterministic on/off proof that this single bit is what kills the launch:

```
$ stat -c %a ~/Windows                                 # 2700
$ pkexec /usr/bin/omarchy-windows-vm __priv up; echo $?  # 1, no output at all
$ chmod g-s ~/Windows                                  # 700
$ pkexec /usr/bin/omarchy-windows-vm __priv up; echo $?  # 0, container comes up
```

**This is not a one-off from an old install.** `dockurr/windows` chmods the shared mount to `2777` on every container start — after a successful `__priv up_wait` here, `~/Windows` is `drwxrwsrwx`. So the bit is re-applied by every launch, and since the same strict `mode == 700` comparison also lives in `mounted_leaf_matches()` (and therefore `mounts_ready`), *every* launch after the first one fails identically. Recreating the directory buys exactly one launch.

## Steps to reproduce

1. Have a Windows VM whose `~/Windows` carries the setgid bit — either from a ≤ 4.0.1 install (`drwx--S---`) or simply from having started the container once on 4.0.2 (`drwxrwsrwx`).
2. Apps menu → Windows, enter the polkit password.
3. Nothing happens, no error, no window.

Minimal repro of the underlying `chmod` behavior, no VM involved:

```bash
mkdir -m 2700 /tmp/sgdir
chmod 0700 /tmp/sgdir; stat -c %a /tmp/sgdir   # 2700
chmod 00700 /tmp/sgdir; stat -c %a /tmp/sgdir  # 700
```

## Workaround

Strip the bit with a form of `chmod` that actually clears it, before every launch:

```bash
chmod 00700 ~/Windows ~/.windows
```

If a launch already failed once, the root-owned bind anchor still pins the old inode when you recreate the directory instead of chmod'ing it, and the next attempt dies with `protected mount ... no longer matches its home source`. In that case unmount the anchor first — see the follow-up comment below. Chmod'ing in place avoids that entirely, since the inode never changes.

This has to be repeated after every container run, so in practice it needs to be wrapped around the launch.

## Expected

One or more of:

- `prepare_caller_mounts()` and `mounted_leaf_matches()` clear the bit with a mode that works — `chmod 00700`, or `chmod u-s,g-s` followed by `chmod 0700` — rather than `chmod 0700`.
- Alternatively, relax the comparison to what the hardening actually cares about: correct owner and no group/other access bits, rather than a strict `== 700`.
- The failure surfaces to the user (desktop notification, or a visible terminal) instead of dying silently *after* a successful password prompt. This is the part that makes it undiagnosable for a normal user.

## System details

- Omarchy 4.0.2-1 (upgraded from 4.0.1; VM originally installed in the 4.0.0 era)
- Kernel: 7.1.9-arch1-2
- coreutils 9.11
- Hyprland 0.56.2, SDDM
- Intel CoffeeLake-H (UHD Graphics 630)
- Filesystem: btrfs (`/@home` subvolume, zstd)

### LINKED ISSUE #9374: windows-vm: chmod 0700 preserves directory setgid bit, causing launch to fail safety check

### LINKED ISSUE #9540: omarchy-windows-vm install: fails to write config when mount source dir has setgid bit (chmod 0700 doesn't clear S_ISGID)

### LINKED ISSUE #9567: omarchy-windows-vm launch fails silently when ~/Windows has the setgid bit (numeric chmod can't clear it)

### LINKED ISSUE #9698: omarchy-windows-vm: setgid on ~/Windows silently blocks every privileged VM action
### What happened

Every privileged `omarchy windows vm` operation (`remove`, `launch`, `stop`) fails
when either VM source directory — `~/.windows` or `~/Windows` — has the setgid bit
set. The failure is silent: the privileged helper returns 1 without printing a
reason, so the user only sees the generic wrapper message.

```
$ omarchy windows vm remove
Removing Windows VM...
❌ Windows VM removal stopped before user-side cleanup; inspect the VM data before retrying.
```

`~/Windows` was mode `2700` (`drwx--S---`). No other diagnostic is printed, so
there is nothing pointing at the mode as the cause.

### Root cause

`omarchy-windows-vm`, `prepare_caller_mounts()`:

```bash
chmod 0700 -- "/proc/$BASHPID/fd/$storage_fd" "/proc/$BASHPID/fd/$shared_fd" || {
  ...
}
  exec {storage_fd}<&-
  exec {shared_fd}<&-
  return 1
fi
```

GNU `chmod` does not clear setuid/setgid on a **directory** for a numeric mode of
four digits or fewer — `0700` leaves the setgid bit untouched. `stat -Lc '%a'`
then reports `2700`, the `!= 700` comparison fails, and the function returns 1.

The hardening step can therefore never satisfy its own postcondition, and the
mode check is unreachable-by-success for any setgid source directory.

This blocks every privileged path, since they all funnel through
`assert_mounts_safe` → `prepare_caller_mounts`:

- `__priv_remove()` — `assert_mounts_safe || return 1`
- `__priv_up()` — `assert_mounts_safe && dc up -d`
- `__priv_up_wait()` — `assert_mounts_safe || return 1`

Nothing is destroyed (the check runs before `dc down` and the `find -delete`), but
the VM becomes permanently un-launchable and un-removable through the CLI.

### Minimal reproduction

Independent of the VM — this is just the predicate:

```bash
d=$(mktemp -d); chmod 2700 "$d"
exec {fd}<"$d/."
chmod 0700 -- "/proc/$BASHPID/fd/$fd"
stat -Lc '%a' "/proc/$BASHPID/fd/$fd"   # prints 2700, expected 700
exec {fd}<&-
```

Observed on coreutils 9.11-2:

| chmod form  | resulting mode |
|-------------|----------------|
| `0700`      | `2700` ❌ |
| `u=rwx,go=` | `2700` ❌ |
| `00700`     | `700` ✅ |
| `g-s`       | `700` ✅ |

Note that the symbolic form `u=rwx,go=` does not clear it either — GNU chmod
preserves setuid/setgid on directories unless they are named explicitly.

End-to-end:

1. `chmod g+s ~/Windows` (or have a pre-existing setgid `~/Windows` before install —
   `install_windows` only creates the directory when it is absent:
   `[[ -e $shared || -L $shared ]] || install -d -m 0700 -- "$shared"`, so an
   existing setgid directory is adopted as-is)
2. `omarchy windows vm install`
3. `omarchy windows vm remove` → fails with the generic message above
4. `omarchy windows vm launch` → fails the same way

### Expected

Either the hardening actually clears the special bits, or the failure explains
itself. Both would be better.

### Suggested fix

```diff
-  chmod 0700 -- "/proc/$BASHPID/fd/$storage_fd" "/proc/$BASHPID/fd/$shared_fd" || {
+  # Numeric modes of four digits or fewer leave setuid/setgid on a directory
+  # alone; the fifth digit is what clears them, and the mode check below
+  # requires exactly 700.
+  chmod 00700 -- "/proc/$BASHPID/fd/$storage_fd" "/proc/$BASHPID/fd/$shared_fd" || {
```

and give the mode check a diagnostic so it is not silent:

```diff
   if [[ $storage_mode != 700 || $shared_mode != 700 ]]; then
+    echo "omarchy-windows-vm: VM source directories must be mode 700 (storage=$storage_mode shared=$shared_mode)" >&2
     exec {storage_fd}<&-
     exec {shared_fd}<&-
     return 1
   fi
```

The same silent-`return 1` pattern appears elsewhere in `prepare_caller_mounts`
and in `__priv_remove` (the `EXPECTED_STORAGE`/`EXPECTED_SHARED` equality checks),
which is why the wrapper's generic message is all the user ever gets. Worth a
pass, though the setgid case is the one that actually bites.

### Workaround

```bash
chmod g-s ~/Windows ~/.windows
```

### System details

- Omarchy 4.0.2-1
- Kernel 7.1.9-arch1-2
- coreutils 9.11-2
- Docker 29.7.2
- Hyprland 0.56.2
- CPU: Intel Core i9-10885H
- GPU: Intel CometLake-H GT2 [UHD Graphics] / NVIDIA TU104GLM [Quadro RTX 5000 Mobile / Max-Q]
- Sudoless Docker: not enabled (privileged actions go through pkexec)

### LINKED ISSUE #9746: omarchy-windows-vm: launch fails when the dockur container re-adds setgid to ~/Windows
## Description

`omarchy-windows-vm launch` starts the VM only on the FIRST run. After the VM
has run once (or been launched via `up`/`up_wait`), every subsequent `launch`
fails with `❌ Failed to start Windows VM!` even though the VM is healthy when
started directly via pkexec or from the web console at `http://127.0.0.1:8006`.

## Root cause

Running the dockur/windows container chmods the shared-folder bind source
`~/Windows` to mode `2777` (setgid + world-writable). The mode persists after
the container stops.

On the next start, Omarchy's `prepare_caller_mounts` runs
`chmod 0700 ~/Windows`, then requires the mount to match mode exactly `700`
(`mounted_leaf_matches` tests `[[ $mode == 700 ]]`). But the kernel
**preserves the setgid bit on a directory for root/in-group chmod**, so the
mount ends up `2700 != 700` and start silently aborts.

`chmod 0700` cannot clear `S_ISGID` on a directory; the script must strip
special modes explicitly, e.g. `chmod u-s,g-s,o-s ~/Windows`.

Workaround on the affected system: `chmod u-s,g-s,o-s ~/Windows` while the VM
is stopped, so mode returns to `0700` before the next launch.

## Expected

- `prepare_caller_mounts` / `bind_mount_leaf` explicitly clear `S_ISGID`/`S_ISUID`
  (e.g. `chmod u-s,g-s,o-s`) in addition to `chmod 0700`, or
- the `== "700"` check tolerates/clears a leftover setgid bit.

## Steps to reproduce

1. `omarchy-windows-vm install` (configure a VM) then `omarchy-windows-vm launch`.
2. Let the VM run and stop it (or shut it down from inside the guest).
3. Run `omarchy-windows-vm launch` again.
4. Observe: `❌ Failed to start Windows VM!`.

Also confirm `stat -c %a ~/Windows` reads `2777` after step 2 and, while the VM
is stopped, `chmod g-s ~/Windows` makes `launch` succeed again.

## System

- Omarchy version: `4.0.2-1`
- Kernel: `7.1.9-arch1-2`
- CPU: AMD Ryzen 3 7320U (KVM `/dev/kvm` present)
- Filesystem for storage: btrfs

### LINKED ISSUE #9884: omarchy-windows-vm: launch/remove fail silently when ~/Windows has setgid (numeric chmod can't clear it since coreutils 9)
# omarchy-windows-vm: launch and remove fail silently when ~/Windows carries a setgid bit (numeric `chmod 0700` can't clear it since coreutils 9)

## System details

- CPU: Intel Core i5-8350U, GPU: Intel UHD Graphics 620 (Kaby Lake-R)
- Omarchy 4.0.2-1 (VM originally installed under the previous major version, ~Aug 27)
- Kernel 7.1.9-arch1-2, `/home` on btrfs subvolume `@home`
- coreutils 9.11

## What's wrong?

After updating to 4.0.2, both `omarchy windows vm launch` and `omarchy windows vm remove` fail:

```text
Starting Windows VM (this may prompt for authorization)...
❌ Failed to start Windows VM!
   Try checking: omarchy-windows-vm status
```

`remove` stops with `❌ Windows VM removal stopped before user-side cleanup; inspect the VM data before retrying.` `omarchy windows vm status` reports "container not found", so there is no obvious cause and no way out except manual surgery.

### Root cause

The security-hardened rewrite binds `~/.windows` and `~/Windows` into root-owned anchors. In `prepare_caller_mounts()` (`/usr/share/omarchy/bin/omarchy-windows-vm`, the "Privacy is an explicit preflight step" block) the privileged path does:

```bash
...
  ...
  return 1   # silent
fi
```

On my machine `~/Windows` was mode **2700** — a setgid leftover from the older omarchy version / the `dockurr/windows` Samba share (the running container sets the shared dir to 2777, and a stopped one leaves the setgid bit). Since GNU coreutils 9.0, **numeric chmod preserves set-user-ID/set-group-ID bits on directories unless explicitly specified** (see `man chmod`: "For directories chmod preserves set-user-ID and set-group-ID bits unless you explicitly specify otherwise. To clear these bits for directories with a numeric mode requires an additional leading zero like 00755, leading minus like -6000, or leading equals like =755."). So the script's own hardening step can never clear the leftover bit: `chmod 0700` succeeds (rc=0) but the very next `stat` still reads `2700`, and the check fails with a silent `return 1`.

Both `__priv_up_wait` and `__priv_remove` funnel through `assert_mounts_safe()` → `prepare_caller_mounts()`, so launch and remove are broken together. The same latent `chmod 0700` exists in the user-side `prepare_user_mount_sources()`.

Minimal evidence of the chmod behavior:

```console
$ mkdir t && chmod 2700 t
$ chmod 700 t && stat -c %a t        # coreutils, numeric: setgid SURVIVES
2700
$ chmod -s t && stat -c %a t
700
$ chmod 2700 t && chmod =700 t && stat -c %a t   # leading '=' does clear it
700
$ chmod 2700 t && python3 -c 'import os; os.chmod("t",0o700)' && stat -c %a t
700                                  # direct chmod(2) clears it; coreutils intentionally doesn't
```

### Steps to reproduce

Windows VM installed on an earlier version and updated (or a `~/Windows` created with the setgid bit), VM stopped:

```bash
chmod 2700 ~/Windows     # simulate the legacy/Samba setgid residue
omarchy windows vm launch   # ❌ Failed to start Windows VM!  (no reason printed)
omarchy windows vm remove   # ❌ ...stopped before user-side cleanup...
chmod -s ~/Windows       # clear setgid with something that actually clears it
omarchy windows vm launch   # works again
```

Tracing the privileged action shows the failure is silent — `bash -x` ends with `storage_mode=700`, `shared_mode=2700`, then an unlogged `return 1` at the mode check.

### Suggested fix

1. Use a mode that actually clears the bits on directories: `chmod =0700` (or `000700`) instead of `chmod 0700` in `prepare_caller_mounts()` and `prepare_user_mount_sources()`.
2. Don't fail silently: when the mode check (or `assert_mounts_safe` generally) bails, print an actionable message to stderr, e.g. `omarchy-windows-vm: refusing to continue: $path has mode $storage_mode, expected 700`. Right now `up_wait`/`remove` exit 1 with zero diagnostics and the user-facing wrapper only says "Failed to start Windows VM".

### Side note

`omarchy debug --no-sudo --print` (as documented) now returns `Unknown Omarchy command: omarchy debug --no-sudo --print`; only the `omarchy-debug` binary works. Possibly a lost dispatcher route.

Diagnostics: `/tmp/omarchy-debug.log` attached (`omarchy-debug --no-sudo --print` on Omarchy 4.0.2).

### LINKED ISSUE #9943: Windows VM: launch fails silently when the container leaves ~/Windows setgid (chmod 0700 cannot clear it)
### Summary

`omarchy-windows-vm launch` fails with `❌ Failed to start Windows VM!` and **no other output**, on a VM that is healthy and reachable over RDP. `omarchy-windows-vm status` simultaneously reports `RUNNING`.

The cause is a mode check in `prepare_caller_mounts` that the script itself cannot satisfy, and the `dockurr/windows` container re-creates the triggering condition on **every** start — so each successful start breaks the *next* launch.

### Environment

- Omarchy 4.0.2-1
- Kernel 7.1.9-arch1-2, GNU coreutils 9.11, Docker Compose 5.5.0
- Intel Core Ultra 9 275HX, Intel Arrow Lake-S + NVIDIA RTX 5070 Mobile

### Root cause

`bin/omarchy-windows-vm:583-593`, inside `prepare_caller_mounts`:

```bash
chmod 0700 -- "/proc/$BASHPID/fd/$storage_fd" "/proc/$BASHPID/fd/$shared_fd" || { ... }
  ...
  return 1
fi
```

The container sets `/shared` — bind-mounted from `~/Windows` — to `2777` on every start. **GNU chmod preserves setuid/setgid on directories when given an octal mode**, so the script's own normalisation cannot clear it:

```
start:        2777
after 0700:   2700   <- the script's chmod, mode "retained as 2700"
after g-s:    700
```

`stat -Lc '%a'` therefore returns `2700`, the comparison against the literal `700` fails, and the function returns 1. The callers at `:859` and `:888` propagate the bare `return 1`, and `launch_windows` prints only `❌ Failed to start Windows VM!`.

### Why this is unusually hard to diagnose

- **The failure is silent.** Every other refusal in `assert_mounts_safe` prints a reason to stderr; this branch prints nothing.
- **`status` still says RUNNING**, because `__priv_status` does not go through the mount checks.
- **It is self-perpetuating.** Clearing the bit by hand fixes exactly one launch; the next container start restores `2777`. Observed here: cleared 23:30, VM started 23:44, broken again by 23:47.
- **Through the desktop launcher the stderr is discarded** (`gtk-launch`), so even the branches that do print a reason surface nothing to the user.
- `chmod 0700 ~/Windows` — the obvious manual fix — reports `mode retained as 2700` and changes nothing, which sends you looking in the wrong place.

### Steps to reproduce

```bash
chmod 2777 ~/Windows        # what the container does on every start
omarchy-windows-vm launch   # ❌ Failed to start Windows VM!   (no other output)
omarchy-windows-vm status   # Windows VM Status: RUNNING
```

### Workaround

```bash
chmod 0700 ~/Windows && chmod g-s ~/Windows   # both calls are required
```

### Suggested fix

1. Normalise the mode in a way that actually clears the high bits — `chmod u=rwx,go=,g-s`, or `chmod 0700` followed by `chmod g-s` — or compare against the permission bits only rather than the literal string `700`.
2. Give this branch a message, matching the other refusals in `assert_mounts_safe`, e.g. `omarchy-windows-vm: refusing VM mount source with unexpected mode $shared_mode`. The silent `return 1` is what turns a one-line fix into a long debugging session.

ekollof:
Fixes #9698
Fixes #9374
Fixes #9334
Fixes #9540
Fixes #9567
Fixes #9884
Fixes #9943
Fixes #9746

GNU `chmod 0700` leaves setuid/setgid on directories when the numeric mode is four digits or fewer. `prepare_caller_mounts` then requires mode `700` exactly, so a leftover `g+s` on `~/Windows` (often `2700` or `2777`) makes every privileged VM action fail closed with no diagnostic.

This hardens those directories with `chmod a-s,u=rwx,go=` (which does clear the special bits) and prints the observed modes if the check still fails.

dockur `samba.sh` also `chmod 2777`s an empty `/shared` bind after the ISO download. The helper re-hardens the **protected bind anchor** (never `$HOME/Windows` by pathname) after the guest reports ready, on `up_wait` timeout, and on `stop`. Fresh `install` does not hold a polkit session for the download: it backgrounds an owner-side wait for `2777` and then restores.


ekollof:
Addressing the review by dropping privileged sentinel creation.

`ensure_shared_sentinel` ran as root in a caller-owned directory with a create/truncate and no `O_NOFOLLOW`, which is a write primitive the rest of `prepare_caller_mounts` exists to prevent. A missing sentinel also should not fail `remove`.

This revision:
- keeps `chmod_private_dir` and the exact-`700` checks
- removes all privileged (and user-side) sentinel creation
- restores share mode **after** `windows started successfully`, so it lands after dockur `samba.sh` rather than racing `dc up -d`
- treats that restore as best-effort so a chmod failure cannot report a successful start as `Failed to start Windows VM`

The world-writable window is now only during guest boot, not until the next privileged action.

ekollof:
Picked both of the remaining options.

`install` goes through `priv up`, so it never hit the guest-ready restore and left `~/Windows` at `2777` for the whole first-boot install (and forever if you never `launch`). Hardening immediately after `dc up -d` would race samba.sh again.

This revision:
- waits on `up` until the share is `2777` (or the shared-folder/samba log line, or a short timeout), then restores through the protected anchor
- restores on `down` after the container is gone, so samba cannot chmod it back
- restores when `up_wait` times out without the ready line

Still never chmod `$LEGACY_SHARED` by pathname, and still never fail a successful start/stop on that restore.

ekollof:
The 15s privileged wait on `up` was the wrong tool: samba.sh runs only after dockur's ISO download, `html "Initializing shared folder..."` does not go to stdout, and the unit test short-circuited before the loop.

This revision:
- leaves `priv up` as start-only, so install does not hold pkexec for 10–15 minutes
- backgrounds an **owner-side** watcher after install that chmods `~/Windows` once it becomes `2777` (user chmod of their own tree, not a root pathname)
- restores every `RUNTIME_DIR/mounts/users/*/shared` on `down`, so `sudo ... stop` without `PKEXEC_UID` and a second-user stop still hit the real share
- keeps the post-ready restore on `up_wait` (and on its timeout)

PR body updated: sentinel text is gone, and the extra setgid issues are listed.

ekollof:
Addressing the review at `e75738ec`:

**Sudoless-Docker restore** (the main open finding): took the non-root-branch option. `__priv_down` now restores by EUID — root walks the mounts tree (`restore_all_shared_privacy`, now root-only), and a sudoless caller goes through `resolve_caller` + `restore_shared_privacy` for its own anchor. That works unprivileged because the anchor sits under the root-owned 0711 tree the caller cannot rename, and while the bind exists it is the caller's own inode. The restore stays best-effort: a failed `resolve_caller` cannot fail a `dc down` that succeeded. A second user's unprivileged `stop` still restores only their own anchor — the most an unelevated caller can do without pkexec — and the next owner action covers the rest.

**Watcher**: exits only after verifying the mode is really 700, so a failed chmod no longer ends the watcher permanently; budget raised to 3600s, and on expiry it applies a final best-effort restore and logs via `logger -t omarchy-windows-vm` (the subshell's output goes to /dev/null, so the journal entry is the visible trace).

**Defence in depth**: `restore_all_shared_privacy` now applies the same `RUNTIME_DIR == /var/lib/omarchy/windows` refusal as `prepare_runtime_tree`.

Tests: the unit test now pins that an unprivileged `restore_all_shared_privacy` does nothing and that the owner-side restore fixes only the caller's own anchor; the boundary suite gains a root walk (both anchors to 700), a `setpriv --reuid=1000` case for the sudoless shape, and a planted non-standard-runtime tree proving the refusal — reverting the script makes the new cases fail. All suites green at this head: windows-vm-test (5), mount-boundary (11), compose (22), windows-key (7), ./test/cli.

ekollof:
Two additions while reproducing a launch failure on a dev-linked machine (`~/Windows` left at 2700, `Failed to start Windows VM!`):

**`e9d5d535` — refuse to elevate a mismatched packaged copy.** pkexec can only run `/usr/bin/omarchy-windows-vm`, which `omarchy dev link` never shadows: the unprivileged half runs the checkout while the elevated half runs the package. On the machine I reproduced on, the packaged copy still shipped `chmod 0700 -- "/proc/$BASHPID/fd/..."`, so root's `prepare_caller_mounts` turned a 2777 share into 2700 (setgid survives numeric modes), failed the exact-700 check, and left no diagnostic. `priv()` now compares the elevated target against the running script (same-inode fast path, sha256 otherwise) and refuses with an actionable message before any polkit prompt. Docker-group users are unaffected: their privileged half is the same file.

**`7a5fa2ee` — boundary suite is no longer host-state-sensitive.** The suite failed on that machine for an unrelated reason worth fixing on this branch: a host VM leaves anchors mounted at exactly the production paths the test re-creates. The user-ns copy of the mount table keeps them listed in /proc/self/mountinfo (they are MNT_LOCKED — the kernel refuses umount2 with EINVAL), the tmpfs over-mount hides them from the filesystem but not from mountinfo, and mountpoint(1) matches entries by path — so `mountpoint -q` reported the fresh anchor dirs as existing mounts, `prepare_mount_anchor` short-circuited, and the suite failed whenever the host had ever run the VM. The fixture uids are now chosen at runtime to avoid any uid the host already has anchors for.

All suites green at this head: windows-vm-test (6), mount-boundary (11), compose (22), windows-key (7).

ekollof:
Rebased onto current quattro (2fbac0c8). All 11 commits replayed cleanly, no conflicts. windows-vm-test.sh and windows-vm-mount-boundary-test.sh pass.

ekollof:
Fix pushed for a style violation this PR introduced: `bin-style-test.sh` ("bin commands use command helpers") flags the raw `command -v systemd-run` probe in `schedule_share_privacy_restore`. Found while triaging full `./test/shell` failures — upstream's copy of this file has no such probe, so this would have failed CI on merge.

`2270df6a` switches the probe to `omarchy-cmd-present systemd-run`. Behavior is unchanged: the background-subshell fallback for tests and stripped sessions without a user bus is preserved exactly as before.

Verified: `bin-style-test.sh`, `windows-vm-test.sh`, `windows-vm-mount-boundary-test.sh`, and `windows-vm-compose-test.sh` all green.

Aananda-giri:
Confirmed on a real Omarchy 4.0.4 host with dockurr/windows v6.05 and coreutils 9.11. After the VM starts, `~/Windows` is left at mode `2700`; the packaged `chmod 0700` preserves directory setgid, then the exact `700` check in `prepare_caller_mounts` fails and `omarchy-windows-vm launch --keep-alive` only reports “Failed to start Windows VM”. Explicitly clearing `g-s` immediately makes the same VM launch and connect over RDP.

I independently reproduced the mode behavior and added a local regression case using a `2700` shared directory. The focused `windows-vm-compose-test.sh` passes with symbolic special-bit clearing. This PR addresses both the relaunch failure and the longer-lived share-permission exposure, so it is preferable to another minimal duplicate PR.

ekollof:
Thanks for the thorough confirmation — the `2700`-preserved-through-`chmod 0700` behavior on coreutils 9.11 plus the exact-700 check failing is precisely the path this PR hardens, and good to know the symbolic special-bit clearing holds in your regression case too. Appreciate you weighing in against the minimal duplicate; the longer-lived share-permission exposure is exactly why this goes further than flipping the check.

Chessing234:
hey, this is already covered in https://github.com/omacom/omarchy/pull/12264 (clear setgid on windows vm harden). mind closing this as a duplicate?

ekollof:
Thanks for checking — not quite a duplicate, though the overlap is real.

#12264 patch 2/2 is a 4-line hardening of one `chmod 0700` call in `prepare_caller_mounts`.

#9783 (Fixes #9698/#9374/#9334/#9540/#9567/#9884/#9943/#9746) goes further:
- `chmod_private_dir` (`a-s,u=rwx,go=`) at all three hardening sites, with observed-mode diagnostics on failure
- dockur `samba.sh` `chmod 2777` re-harden via the protected bind anchor (after ready / `up_wait` timeout / `stop`), owner-side watcher for `install` so no long polkit hold
- EUID-aware restore for sudoless-Docker, packaged-copy mismatch refusal under `omarchy dev link`, plus `windows-vm-test.sh` + `windows-vm-mount-boundary-test.sh` coverage

Independently confirmed by @Aananda-giri on 4.0.4 (mode `2700` -> fail, clear `g-s` -> launch).

Happy to rebase onto #12264 if it lands first and shrink this to the delta, or close if maintainers prefer the minimal version — just let me know which direction you want.

PR #9794, opened 2026-09-02
Expose dropdown indicator font

villekivela:
Dropdown currently renders its Nerd Font chevron with the same font family as option labels. Consumers that use a separate icon font cannot override the indicator without also breaking label text.

Add indicatorText and indicatorFontFamily properties while preserving the existing glyph and font defaults. This is backward-compatible and lets hosts supply another icon font or glyph.

Test: bash test/shell.d/dropdown-indicator-test.sh


villekivela:
Closing this — it works on Omarchy as-is, so no change is needed.

PR #9795, opened 2026-09-02
Require interactive review for supported AUR flows

tonibergholm:
## Summary

- hold AUR changes during unattended and non-terminal updates, preserving each original prompt stream across the update logger's pseudo-terminal
- centralize prompt eligibility for AUR review, package-conflict resolution, and orphan removal
- centralize reviewed Yay execution so complete PKGBUILDs and available Yay/Git diffs open in the same blocking terminal viewer before final package-manager confirmation
- handle yay's empty-query status correctly and apply the transaction ignore list to the read-only report
- preserve explicit picker reinstalls without disabling Yay's build cache while routing them through the reviewed path and removing the sudo keepalive
- keep AUR-dependent migrations pending only when review cannot start, while propagating real install failures
- report query failures and declined or broken optional AUR updates without aborting the trusted system update
- preserve interactive migration review for curl-piped Quattro upgrades, and make `--yes` consistently unattended
- document the emergency review gate, its limitations, and the next trust boundaries

This is a compatibility-phase review gate. It does not sandbox PKGBUILDs, bind recipes and sources to immutable hashes, inspect built artifacts independently, or produce signed attestations.

## Testing

- `bash test/shell.d/update-aur-pkgs-test.sh`
- `bash test/shell.d/pkg-aur-review-test.sh`
- `bash test/shell.d/pkg-aur-run-reviewed-test.sh`
- `bash test/shell.d/update-can-prompt-test.sh`
- existing update sequence, lock, disk-space, conflict, orphan, migration, channel, and Quattro upgrade shell tests
- `./test/cli`
- `bash -n` for changed shell scripts and tests
- `git diff --check`

The complete 230-file shell suite exercised the new tests successfully. Its five remaining failures were unrelated existing/environment cases: three require a sibling `omarchy-pkgs` checkout, plus `launch-about-test` and `network-qr-test`.


bjarneo:
Automated duplication check: this pull request looks similar to #9227, which covers the same AUR interactive gate. I keep that one open and close this one to consolidate review.

Issue #7439, opened 2026-08-19
No Wi-Fi on Apple Silicon Macs: the Broadcom quirk's chip-ID gate includes BCM4378/BCM4387


thejamescollins:
### System details

2021 16" MacBook Pro, Apple M1 Max, 32GB RAM (MacBookPro18,2 / apple,j316c), Broadcom BCM4387 [14e4:4433], Asahi Linux on Arch Linux ARM aarch64, linux-asahi 7.1.6.asahi1-1, Omarchy 4.0.0

### What's wrong?

With Wi-Fi firmware loading and `wlan0` present, no network can be joined. Scanning lists everything in range. Connecting hangs, times out, and NetworkManager reports it like a wrong password. After a couple of attempts the chip stops answering the driver, so scanning dies too, and only a driver reload or a reboot brings it back.

Not to be confused with the other Apple Silicon Wi-Fi problem, which gives no `wlan0` at all: that one comes from `omarchy_hooks.conf` reassigning `HOOKS` and dropping Asahi's `asahi` initcpio hook, already reported in #6876. Everything here is with firmware loading correctly.

The cause is `install/hardware/apple/fix-brcmfmac-supplicant.sh`, which writes `options brcmfmac feature_disable=0x82000` for any Apple machine matching a `brcmfmac` PCI ID from `brcm_hw_ids.h`. Two of those nine IDs are Apple Silicon parts, not Intel Macs:

| ID | Constant | Hardware |
|---|---|---|
| `4425` | `BRCM_PCIE_4378_DEVICE_ID` | BCM4378, M1 |
| `4433` | `BRCM_PCIE_4387_DEVICE_ID` | BCM4387, M1 Pro/Max/Ultra, M2 |

On a BCM4387 the flag does the opposite of its job. It is meant to fix a WPA handshake that never completes, and instead it stops association happening at all. Without the flag the same machine joins the same AP normally.

#6652 flagged this risk itself, with "The chip-ID list is derived, not tested", and asked for confirmation from an `iMac19,x` or a 2017 MacBook Pro. This is confirmation, from the other end of the list.

#### Reproduce

1. Apple Silicon Mac on Asahi, with vendor firmware loading so `wlan0` exists and scanning works.
2. Confirm `/etc/modprobe.d/brcmfmac.conf` contains `options brcmfmac feature_disable=0x82000`, written by the script at install time or by `migrations/1786391100.sh` on an existing install. Reboot.
[...]

PR #9803, opened 2026-09-02
Fix WPA3 on MacBookPro16,1

### LINKED ISSUE #9802: MacBookPro16,1 Broadcom quirk disables WPA3-SAE
# MacBookPro16,1 Broadcom quirk disables WPA3-SAE

### System details

2019 16-inch Intel MacBook Pro (`MacBookPro16,1`), Broadcom BCM4364B3 `[14e4:4464]`, Omarchy 4.0.1-1 Quattro, `linux-t2` 7.1.8, brcmfmac firmware `9.30.514.0.32.5.94` (`FWID 01-68d7ff80`).

### What's wrong?

Omarchy writes this module option on Apple systems using the BCM4364:

```text
options brcmfmac feature_disable=0x82000
```

On this MacBookPro16,1, that option prevents connections to WPA3-only networks. Scanning still finds the AP, but authentication/association times out and NetworkManager eventually reports `ssid-not-found`.

The option's current comment says `0x80000` disables firmware authentication (`FWAUTH`). In the current brcmfmac feature enum, however, bit 19 (`0x80000`) is `BRCMF_FEAT_SAE`; `FWAUTH` is bit 20 (`0x100000`). Disabling `0x82000` therefore disables both firmware supplication and SAE. cfg80211 only advertises SAE offload when FWSUP and SAE remain enabled, and this firmware does not expose a separate host SAE path.

`feature_disable=0x82000` never disabled FWAUTH: it disabled FWSUP and SAE. FWAUTH was enabled under the old configuration and remains enabled with the driver defaults. The behavioral change on `MacBookPro16,1` is restoring FWSUP and SAE, which were verified against WPA2, transition-mode, and WPA3 networks. There is no evidence that disabling the separate FWAUTH capability is necessary or beneficial for this client-mode issue.

Reproduction:

1. Confirm `/etc/modprobe.d/brcmfmac.conf` contains `options brcmfmac feature_disable=0x82000` and load brcmfmac with that value.
2. Attempt to join a WPA3-SAE network with PMF required. The connection times out.
3. Reload brcmfmac with `feature_disable=0`, or remove the Omarchy-owned option and reboot.
4. `iw phy` now reports `SAE_OFFLOAD`; the same profile joins successfully, obtains DHCP, and passes gateway and internet ping tests.

This was reproduced with the same adapter and firmware in one session. WPA3-SAE with PMF required connected after the live reload and again after persisting the driver default. A WPA2/WPA3 transition AP was then tested through both authentication paths: its normal WPA-PSK profile and a temporary SAE profile with PMF required both associated, obtained DHCP, and passed gateway and internet ping tests with 0% packet loss. The WPA-PSK profile also passed a disconnect/reconnect cycle. A separate WPA2-only AP was not available, but the WPA2 authentication path on the transition AP is verified.

This is related to #7439 and #7577, but those concern Apple Silicon BCM4378/BCM4387 systems. The proposed architecture guard in #7577 deliberately retains the quirk on x86_64 and therefore does not restore WPA3 on this Intel Mac.

Suggested narrow fix: exempt the verified `MacBookPro16,1` from the quirk, correct the bit documentation, and migrate existing installs by removing only the exact Omarchy-owned configuration blocks. Other Apple models should retain the transition-mode workaround until they are tested.

kazeshini178:
## Summary

- keep brcmfmac firmware SAE offload enabled on `MacBookPro16,1`
- correct the `feature_disable` documentation: `0x80000` is SAE, not FWAUTH
- remove the exact Omarchy-owned `0x82000` block from existing `MacBookPro16,1` installs and request a reboot
- leave other Mac models and administrator-customized configurations unchanged

Fixes #9802.

Related: #7439, #7577, #6652.

## Why

The BCM4364B3 firmware on a tested 2019 16-inch Intel Mac needs its advertised SAE offload path to join WPA3-only networks. Omarchy's `feature_disable=0x82000` transition-mode workaround disables bit 19 (`BRCMF_FEAT_SAE`) as well as firmware supplication, so WPA3-SAE authentication is unavailable on this model.

`feature_disable=0x82000` never disabled FWAUTH: it disabled FWSUP and SAE. FWAUTH is bit 20 (`0x100000`) and was enabled both before and after this fix. The behavioral change on `MacBookPro16,1` is restoring FWSUP and SAE, which were verified against WPA2, transition-mode, and WPA3 networks. There is no evidence that disabling FWAUTH is needed for this client-mode issue.

With the quirk enabled, a WPA3-only connection timed out. Reloading the same driver and firmware with `feature_disable=0` exposed `SAE_OFFLOAD`; the same NetworkManager profile then associated, obtained DHCP, and passed gateway and internet ping checks. The persistent configuration was verified after a second reload.

The model exemption is intentionally narrow. The workaround remains useful for older Broadcom firmware with broken WPA2/WPA3 transition-mode handling, and only `MacBookPro16,1` has been verified here.

## Test plan

- `bash test/shell.d/brcmfmac-supplicant-test.sh`
- `bash test/shell.d/brcmfmac-wpa3-cleanup-test.sh`
- `bash -n` on all changed shell files
- `git diff --check`
- live WPA3-SAE/PMF-required association, DHCP, gateway ping, and internet ping on BCM4364B3
- live WPA2/WPA3 transition AP through both WPA-PSK and SAE/PMF-required profiles, including DHCP and 0%-loss gateway/internet pings
- disconnect/reconnect cycle on the transition AP's WPA-PSK profile

`./test/all` was also run. The changed Wi-Fi tests passed; five unrelated environment/baseline tests failed. Three require a separate `omarchy-pkgs` checkout, while `launch-about-test.sh` and `network-qr-test.sh` fail in this non-Wayland test environment. No failure references a changed file.

A separate WPA2-only AP was not available. The WPA2 authentication path itself was verified by pinning the transition AP profile to WPA-PSK; NetworkManager logged `WPA-PSK WPA-PSK-SHA256 FT-PSK`, and the connection passed association, DHCP, reconnect, and traffic tests.

### DIFF
--- a/install/hardware/apple/fix-brcmfmac-supplicant.sh
+++ b/install/hardware/apple/fix-brcmfmac-supplicant.sh
@@ -3,9 +3,9 @@
 # WPA2/WPA3 transition mode: the client associates, the four-way handshake never
 # completes, and NetworkManager reports the password as wrong.
 #
-# feature_disable turns off the firmware supplicant (FWSUP, 0x2000) and firmware
-# authenticator (FWAUTH, 0x80000), handing the handshake back to wpa_supplicant
-# in software.
+# feature_disable turns off the firmware supplicant (FWSUP, 0x2000) and SAE
+# authentication (SAE, 0x80000), handing WPA2 handshakes back to wpa_supplicant
+# in software and preventing WPA3-only authentication.
 #
 # This quirk already shipped for T2 Macs, detected by the T2 PCI ID that every
 # one of them carries. The bug is in the Broadcom firmware, not in the T2 bridge,
@@ -19,17 +19,20 @@
 # from the T2 era on. The BCM4360 in 2013-2015 Macs is deliberately absent: it
 # runs the out-of-tree wl driver, which never reads a brcmfmac option.
 sys_vendor="$(cat /sys/class/dmi/id/sys_vendor 2>/dev/null || true)"
+product_name="$(cat /sys/class/dmi/id/product_name 2>/dev/null || true)"
 
-if lspci -nn | grep "106b:180[12]" >/dev/null ||
+if [[ $product_name == "MacBookPro16,1" ]]; then
+  echo "Detected a MacBookPro16,1; keeping WPA3 firmware offload enabled"
+elif lspci -nn | grep "106b:180[12]" >/dev/null ||
   { [[ $sys_vendor == Apple* ]] &&
     lspci -nn | grep -E "14e4:(43ba|43bb|43bc|43a3|43dc|4464|4488|4425|4433)" >/dev/null; }; then
   echo "Detected a Mac with Broadcom Wi-Fi; running the WPA handshake in software"
 
   mkdir -p /etc/modprobe.d
   cat > /etc/modprobe.d/brcmfmac.conf <<'EOF'
-# Broadcom's firmware supplicant and authenticator fail the WPA four-way
-# handshake on Apple hardware, which surfaces as a rejected password. Disable
-# both so wpa_supplicant performs the handshake instead.
+# Broadcom's firmware WPA handling fails on affected Apple hardware, which
+# surfaces as a rejected password. Disable firmware supplication and SAE so
+# transition-mode networks use the software WPA2 handshake instead.
 options brcmfmac feature_disable=0x82000
 EOF
 fi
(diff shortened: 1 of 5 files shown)


thedavidweng:
Can confirm this exact issue also affects **`MacBookPro15,2`** (and many more models) and the fix will work too.

The model check `product_name == "MacBookPro16,1"` is too narrow:
1. All T2 Intel Macs (`MacBookPro15,x`, `MacBookPro16,x`, `MacBookAir8,x`/`9,x`, `Macmini8,1`, `iMac20,x`, `MacPro7,1`) use the same **BCM4364 (`14e4:4464`)** controller and 2021-2023 modern firmware lineage that natively supports SAE offload.
2. The original bug that inspired #6652 was on legacy pre-T2 hardware (BCM43602, 2015-era firmware). Applying the quirk to BCM4364 was an accidental overshoot.

PR #9806, opened 2026-09-02
Support browser flags in web app bindings

fabiokr:
## Summary
- allow web app bindings to pass browser flags as an argument list
- preserve argument boundaries for focused web app launches
- document selecting a Chromium browser profile

## Tests
- not run after final changes (requested)


fabiokr:
Alternatively, maybe we could define a default browser profile system wide, and when opening web apps we forward that?

PR #9812, opened 2026-09-02
Don't pin VA-API to NVIDIA on hybrid GPU systems

ifigueroap:
## Problem

On a hybrid GPU laptop, video flickers while scrolling — reproducible on Twitter/X, where videos scroll in and out of the viewport continuously.

## Root cause

`default/hypr/nvidia.lua` exports `LIBVA_DRIVER_NAME=nvidia` for the whole session whenever it finds an NVIDIA GPU with GSP firmware. The systemd user manager inherits it, so desktop launches get it too.

That is correct on an NVIDIA-only machine. On a hybrid laptop the iGPU normally drives the panel, so pinning VA-API to the NVIDIA driver makes every app decode video on the dGPU, and each decoded frame then crosses the PRIME boundary to reach an iGPU-composited display.

Measured on the affected machine before the fix — Edge's GPU process:

```
/usr/lib/dri/nvidia_drv_video.so
/usr/lib/libEGL_nvidia.so.610.57.04
23 fds on /dev/nvidia0
```

while the panel is on the Intel iGPU (`card2-eDP-2`, `i915`).

## Fix

Gate the pin on a new `omarchy-hw-nvidia-only` detector. Left unset, libva picks the driver matching whichever GPU an app actually renders on — the per-device answer a hybrid system needs. `NVD_BACKEND` and `__GLX_VENDOR_LIBRARY_NAME` are left alone.

The detector reads sysfs rather than reusing `omarchy-hw-hybrid-gpu`, which shells out to `lspci`. `nvidia.lua` already documents why that matters: lspci reads PCI config space, resuming a runtime-suspended GPU, and the wake alone outlasts Hyprland's 1.5s config reload budget.

## Verification

Hardware: Intel Raptor Lake-P `[8086:a7ab]` + NVIDIA GA107BM RTX 3050 Laptop `[10de:25ac]`, panel on the iGPU, Omarchy 4.0.2, `nvidia-open-dkms` 610.57.04, Hyprland/Wayland.

Applying the change (via an equivalent per-app override) and restarting the browser:

| | Before | After |
|---|---|---|
| Video decode | `nvidia_drv_video.so` | `iHD_drv_video.so` |
| EGL | `libEGL_nvidia.so` | `libEGL_mesa.so` |
| `/dev/nvidia0` fds | 23 | 0 |

Flicker is gone. Hardware decode is retained, now on the iGPU:

```
$ env -u LIBVA_DRIVER_NAME vainfo
vainfo: Driver version: Intel iHD driver for Intel(R) Gen Graphics - 26.2.4
```

Worth noting that unsetting the variable was sufficient on its own — EGL moved to Mesa as a consequence, without needing `__EGL_VENDOR_LIBRARY_FILENAMES`. That is why this PR changes only the one export.

Note this requires an iGPU VA-API driver to be present (`intel-media-driver` here). Without one, decode falls back to software instead of decoding on the wrong GPU — which is the safer failure, and no worse than the current behaviour on machines where the NVIDIA VA-API driver is absent.

## Tests

`test/shell.d/hw-nvidia-test.sh` covers the new detector across every existing fixture, plus a new Intel + NVIDIA hybrid case. `test/cli` passes. `test/shell` has 3 failures (`config-test`, `snapper-test`, `unowned-system-paths-test`) that reproduce identically on unmodified `quattro`.

https://claude.ai/code/session_01BT3H7mj63CxzHWKLd4apWk

### DIFF
--- a/bin/omarchy-hw-nvidia-only
+++ b/bin/omarchy-hw-nvidia-only
@@ -1,0 +1,23 @@
+#!/bin/bash
+
+# omarchy:summary=Detect whether NVIDIA is the only GPU in the computer.
+
+# Read the cached sysfs IDs rather than lspci, which reads PCI config space and
+# resumes runtime-suspended GPUs.
+pci_devices_path="${OMARCHY_PCI_DEVICES_PATH:-/sys/bus/pci/devices}"
+
+shopt -s nullglob
+
+nvidia_found=false
+
+for device in "$pci_devices_path"/*; do
+  [[ $(< "$device/class") == 0x03* ]] || continue
+
+  if [[ $(< "$device/vendor") == "0x10de" ]]; then
+    nvidia_found=true
+  else
+    exit 1
+  fi
+done
+
+[[ $nvidia_found == true ]]
--- a/default/hypr/nvidia.lua
+++ b/default/hypr/nvidia.lua
@@ -3,6 +3,7 @@ local paths = require("default.hypr.paths")
 local nvidia = paths.omarchy_path .. "/bin/omarchy-hw-nvidia"
 local nvidia_gsp = paths.omarchy_path .. "/bin/omarchy-hw-nvidia-gsp"
 local nvidia_without_gsp = paths.omarchy_path .. "/bin/omarchy-hw-nvidia-without-gsp"
+local nvidia_only = paths.omarchy_path .. "/bin/omarchy-hw-nvidia-only"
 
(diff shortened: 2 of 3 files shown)


bjarneo:
Automated duplication check: this pull request looks similar to #7851, which covers the same LIBVA-only NVIDIA gating on hybrid-GPU systems. I keep that one open and close this one to consolidate review.

Issue #9817, opened 2026-09-02
Launch feedback OSD remains stuck after Cursor and Grok Bot launch


aj-ebels:
### System details

Omarchy: 4.0.2-1   Hyprland: 0.56.2-1   Quickshell: 0.3.1-1   Kernel: 7.1.9-arch1-2   Desktop/session: Wayland / Hyprland   Hardware: Dell XPS 13 DX13260   CPU: Intel Core 5 320   GPU: Intel Wildcat Lake, xe driver   Display: 2560x1600 internal display

### What's wrong?

The Omarchy launch-feedback OSD remains stuck indefinitely after launching applications from the Omarchy application menu.

This has occurred with both Cursor and Grok Bot. The applications launch successfully and their windows appear normally, but the OSD remains visible showing:

Launching Grok Bot…

The same behavior previously occurred with Cursor.

The OSD should disappear when the application window appears, or after the documented 15-second failsafe timeout. Instead, it remains visible indefinitely.

I confirmed that both applications are running and registered correctly in Hyprland:

- Cursor: class `cursor`
- Grok Bot: class `grok-bot`

The stuck OSD can be manually dismissed with: omarchy-shell osd close

The installed `/usr/share/omarchy/shell/services/AppLibrary.qml` shows the launch OSD with `duration: 0`, then attempts to close it when a new toplevel appears or after 15 seconds.
Neither automatic cleanup path appears to work in this case.

The attached debug log and screenshot provide additional details.

<img width="578" height="180" alt="Image" src="https://github.com/user-attachments/assets/be3693a7-9c9e-467a-833d-508da3b3d669" />

[omarchy-debug.log](https://github.com/user-attachments/files/31741006/omarchy-debug.log)

xjerster:
Confirming on another machine with Grok Bot (AppImage 0.39.0) launched from the Omarchy app menu.

**System:** Omarchy 4.0.2-1, Hyprland 0.56.2 (efb5099), kernel 7.1.9-arch1-2, AMD Ryzen 7 7435HS, NVIDIA GeForce RTX 4060 Max-Q.

[...]

Issue #8242, opened 2026-08-25
Launching… OSD can remain indefinitely due to unordered show/close IPC calls


brianmsm:
## System details

- AMD Ryzen AI 9 365 / Radeon 880M
- Omarchy 4.0.1-1
- Hyprland 0.56.2
- Quickshell 0.3.1-1

## What's wrong?

The `Launching …` OSD can remain visible indefinitely after the launched application has already mapped and become active. I have seen this twice; the latest occurrence was launching TIDAL Hi-Fi from the Omarchy app menu.

Expected: the OSD closes when the app maps, or at worst after the 15-second launch timeout.

Actual: the app launches successfully, but the OSD remains until manually closed with:

```bash
omarchy-shell osd close
```

A privacy-cropped screenshot is available if needed. The full `omarchy-debug` output is intentionally not attached because it contains unrelated local identifiers and network logs; the system details above and deterministic reproduction contain the relevant evidence.

## Probable root cause: unordered detached `show` / `close` IPC calls

In `shell/services/AppLibrary.qml`, `launchDelay` waits 2 seconds, sets `launchOsdOpen = true`, and dispatches this asynchronously:

```qml
Quickshell.execDetached(["omarchy-shell", "osd", "show", JSON.stringify({
  icon: "󱓞",
  message: root.launchOsdMessage,
  duration: 0
})])
```

If the app maps at approximately the same time, `onValuesChanged()` or `onActiveToplevelChanged()` calls `closeLaunchFeedback()`. That stops `launchTimeout`, dispatches a separate detached `osd close` process, and sets `launchOsdOpen = false`.

The two detached IPC client processes are not ordered. The `close` request can reach Quickshell first, followed by the delayed `show`. In that ordering:

1. `close` closes an OSD that is not open yet.
2. AppLibrary marks the OSD closed and stops the 15-second timeout.
3. `show` arrives afterward with `duration: 0`.
4. Nothing remains that can close it, so it stays indefinitely.

This timing explains why the problem is intermittent and more likely for apps that map around the 2-second launch-feedback threshold.

## Deterministic IPC-level reproduction

Running the same `show` and `close` commands concurrently reproduces the unordered final state:

```bash
payload='{"icon":"󱓞","message":"Race test","duration":0}'
open=0
closed=0
for i in $(seq 1 100); do
  omarchy-shell osd show "$payload" >/dev/null &
  omarchy-shell osd close >/dev/null &
  wait
  state=$(omarchy-shell osd state)
[...]

PR #9819, opened 2026-09-02
Drive the launch OSD in-process so a lost close cannot strand it

Yiin:
The "Launching Slack…" OSD stayed on screen after Slack had opened. Nothing removed it until a shell restart.

The launcher opens this OSD with duration 0. Only an explicit close removes it. It sends both the show and the close by spawning `omarchy-shell`, a bash script that forwards to `qs ipc` with a 2s timeout and no retry. If the close process times out, or runs before the show process, the OSD stays forever. Both cases are easy to hit while the launched app loads the CPU.

The shell already has `summon()` and `hide()`. Other plugins use them for the OSD. This change hands the shell to `AppLibrary` and calls them directly. The calls are in-process and ordered, so a close cannot be lost or overtake a show.

Tested by running the shell from this branch and launching a desktop entry that sleeps 5s, then opens a terminal. The OSD appeared after 2s and closed when the window appeared. `./test/all` passes except `config-test.sh`, which needs an `omarchy-pkgs` checkout.

https://claude.ai/code/session_017J3tUnyMTMmWF6MDp96LZC

### DIFF
--- a/shell/services/AppLibrary.qml
+++ b/shell/services/AppLibrary.qml
@@ -12,6 +12,9 @@ Item {
   id: root
 
   property string omarchyPath: Quickshell.env("OMARCHY_PATH")
+  // The shell that owns this library; launch feedback drives the OSD plugin
+  // through it in-process rather than through a spawned omarchy-shell call.
+  property var shell: null
 
   property var configuredHiddenEntryIds: ({})
   property var desktopHiddenEntryIds: ({})
@@ -171,7 +174,7 @@ Item {
     launchDelay.stop()
     launchTimeout.stop()
     if (root.launchOsdOpen) {
-      Quickshell.execDetached(["omarchy-shell", "osd", "close"])
+      root.shell.hide("omarchy.osd")
       root.launchOsdOpen = false
     }
   }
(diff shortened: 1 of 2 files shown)


paulogeyer:
Looks good — this is the right fix for the sticky “Launching…” OSD.

**Why:** AppLibrary currently show/closes the launch OSD via unordered `execDetached` → `omarchy-shell osd …`, with `duration: 0`. If close runs before show (easy when an Electron app maps around the 2s delay), the toast sticks forever and the 15s timer is already stopped. Driving `shell.summon` / `shell.hide` in-process matches how volume/media already use the OSD and removes that race. Keeping `duration: 0`, the 15s fallback, and the `launchOsdOpen` guard is correct — a duration-only backstop alone would leave a stale toast after the app is open.

**Laptop check:** On my laptop (ASUS Vivobook, Intel Arc) I hit the sticky toast once before with Grok Bot; I couldn’t re-stick it today (toast showed and cleared on stock), so this isn’t a live stock-fail → PR-pass soak — just code review + agreement with the race reports on #8242 / #9817 / #12123.

**Nit (optional):** null-guard `shell` before summon/hide if AppLibrary can ever be constructed without the shell wiring.

Happy to re-test if maintainers want a forced IPC race harness; otherwise 👍 from me.

PR #9833, opened 2026-09-02
Switch Claude accounts from the agents panel

tfournet:
The agents widget already switches providers. It still has one Claude login at a time, so personal and work means swapping `CLAUDE_CONFIG_DIR` by hand. See https://github.com/omacom/omarchy/discussions/9821.

`omarchy agent account` registers the live `~/.claude` session, adds further logins as isolated config dirs, and writes a pointer. Switching copies the selected login onto `~/.claude/.credentials.json` as a regular file so tools that hardcode that path follow. The saved copy is a different file; a live Claude blank does not destroy it, empty writes are not harvested back, and last-good restores a blanked saved file. `omarchy agent` and the Claude usage collector follow the pointer. The Claude tab of the agents panel shows email chips and a `+` to add another account.

The usage collector still probes when the saved token's `expiresAt` has lapsed. Only an empty token or HTTP 401/403 is treated as signed out.

Codex is out of this change. A later `omarchy-agent-account-codex` collector would light the same chips.

## Test plan

- `bash test/shell.d/agent-account-claude-test.sh` — list/add/use against a fake `$HOME`, original credentials left in place, live file is not a symlink onto the saved copy, blanking live does not wipe the saved login, usage collector follows the pointer, unknown providers fail
- `bash test/shell.d/agent-usage-claude-limits-test.sh` — lapsed `expiresAt` still probes; 401 is signed out
- `bash test/shell.d/agents-panel-test.sh` — right-click still launches the agent; chips are Claude-tab only
- `./test/cli` — metadata for the new binaries

Visual: adding a second account uses `omarchy-launch-tui` and leaves the original `~/.claude` login intact. New `omarchy agent` windows follow the pointer. Already-open Claude windows do not.

### DIFF
--- a/bin/omarchy-agent
+++ b/bin/omarchy-agent
@@ -75,6 +75,10 @@ crush)
   fi
   ;;
 claude)
+  if [[ -x ${OMARCHY_PATH:-}/bin/omarchy-agent-account ]]; then
+    claude_dir=$("$OMARCHY_PATH/bin/omarchy-agent-account" dir claude 2>/dev/null || true)
+    [[ -n $claude_dir ]] && export CLAUDE_CONFIG_DIR=$claude_dir
+  fi
   command=(claude --permission-mode auto)
   [[ -n ${prompt:-} ]] && command+=(-- "$prompt")
   ;;
--- a/bin/omarchy-agent-account
+++ b/bin/omarchy-agent-account
@@ -1,0 +1,90 @@
+#!/bin/bash
+
+# omarchy:summary=List or switch AI coding-agent accounts
+# omarchy:args=<list|use|add|dir|current> [claude] [id]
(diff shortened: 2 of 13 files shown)


tfournet:
Follow-up from live use (email chips covered):

**Selected login, token graphs, limit meters**

Agents panel on one Claude login with chips redacted

**Other login after a switch — graphs change; exhausted-limit toast**

Agents panel after switching logins, chips redacted, used-up notification

Pushed:
- `~/.claude/.credentials.json` points at the active login (Hermes hardcodes that path)
- toast once per used-up window, and whether another login is saved
- usage record reload so the Claude tab stays put when that login has no local transcripts

bjarneo:
Automated duplication check: this pull request looks similar to #7272, which covers the same agent account switching. I keep that one open and close this one to consolidate review.

PR #506, opened 2025-08-05
Enable accelerated video decode for Chromium

shayne:
Add AcceleratedVideoDecodeLinuxGL feature flag to chromium-flags.conf to improve video playback performance and power consumption.

Issue #277, opened 2025-07-22
hyprshot adds 1px border to screenshots


al3rez:
when you take screenshots hyprshot adds 1px border to screenshots, which we can fix by adding these:

```
layerrule = noanim, hyprpicker
layerrule = noanim, selection
```

to hyprland.conf

al3rez:
Before:
<img width="1021" height="603" alt="Image" src="https://github.com/user-attachments/assets/6037371f-6a7e-4f1a-b901-ddf660d9cf39" />

After:

<img width="1153" height="811" alt="Image" src="https://github.com/user-attachments/assets/a8fbd196-e8e3-45b6-8649-9d4de4f88ddc" />

dhh:
Seems like we just need the one on selection. But this is great. I've added it and given you the commit credit 👍

PR #9835, opened 2026-09-02
Run Omarchy on Apple Silicon under Asahi Linux

maralcbr:
## Summary

Preliminary runtime integration for running Omarchy on Apple Silicon under Asahi Linux, coordinated with the package repository and subsequent image-builder work.

- Detect Apple displays, trackpads and system batteries; configure Apple HID, Vulkan, audio scheduling and the appropriate platform exclusions. Tap-to-click is disabled only for the named built-in Apple touchpads.
- Make signing-key updates, conflict recovery and toolchain setup architecture-aware. Keyring recovery reinstalls missing payloads while preserving existing trust requirements, checks the pinned Omarchy fingerprint, and verifies the key before reporting success. Apple refreshes the Omarchy, Arch Linux ARM and Asahi repository keyrings. A wiped platform trust store still needs trusted bootstrap media.
- Express optional-install package/architecture guards directly in menu rows, with one cached availability checker supporting provider/version resolution. Chrome and Edge remain x86-only. Xbox support uses upstream's matching-header base guarantee.
- Share architecture-aware pacman configuration staging between refresh and offline finalization. Preserve upstream's developer-checkout-before-refresh behavior, transaction protection and channel-switch recovery guidance.

## Companion package and remaining integration

**Wi-Fi package: [omacom/omarchy-pkgs#506](https://github.com/omacom/omarchy-pkgs/pull/506).** `omarchy-settings-asahi` owns `/usr/lib/NetworkManager/conf.d/wifi_backend.conf` with the iwd backend. Generic settings advertise it as an optional aarch64 dependency; the Apple image profile must select it before hardware setup. Administrator `/etc` configuration retains precedence. Existing Apple installations receive it through a migration using the protected update transaction wrapper.

The earlier package foundation is omacom/omarchy-pkgs#275, #276 and #277. This remains coordinated source integration: **the add-on must be signed/published and included in the Apple image profile before this dependent runtime is distributed.** Creating the companion PR does not publish a package repository or an image.

**ARM channel activation remains deferred.** The checked-in FIXME and empty qualification allowlist preserve the installed repositories until channel publication/aliases and signed upgrade/downgrade/reboot qualification are complete. Configuration consolidation is implemented; channel enablement is not being claimed as complete.

## Latest review corrections

Rebased onto quattro `9c5482c58dbe4974de337450754885083c91eada`, resolving all seven conflicting files. Retained upstream keyring failure checks, developer-channel order, protected update wrappers and current GitHub URLs. Removed redundant Xbox header selection/guard and `docs/apple-review-integration.md`. The SPI one-liner and removal of the network heredoc remain present; old review-thread hunks can still show their previous versions.

## Validation

Runtime commit: `97668c5641f973e3678a05c6142ca903593f00aa`.
Package commit: `097cc7f0adee0adf4cef86494ac2bbfeb70667a5`.

- Focused keyring, Apple settings migration, developer-channel order, pacman staging, conflict recovery, optional availability/transaction drift and SPI tests pass as unprivileged users in both aarch64 and x86_64 containers.
- CLI checks pass for both upstream and this head on both architectures.
- Full shell comparison: 244 upstream files versus 260 on this head. All 16 added test files pass on both architectures. The x86 run has exactly the same 24 failing files on base and head. The final ARM run has 13 baseline failures plus an intermittent update-lock failure in unchanged code; that test passes for both base and head on an isolated rerun. Existing environment/test failures are retained in the evidence, so the overall shell suite is not described as green. All runs use disposable containers and non-root users; do not run this suite as root on a live installation.
- The ARM runtime and development settings packages were rebuilt from this exact runtime commit, with archive contents verified. Runtime build-only dependency checks were skipped; these are unsigned review artifacts, not release packages.
- The companion add-on passes makepkg checksum verification and a native M2 Max build with dependency checks enabled. NetworkManager's isolated print-config test confirms the vendor iwd default and administrator override precedence. The M2 also has the matching Aurora header tree for its running kernel. No live networking or boot state was changed.
- Companion PR #506 passes both build-isolation and self-tests CI jobs. Two Claude CLI Fable 5.1 High reviews reported no blockers in the supplied implementation excerpts; the Asahi keyring and protected-migration suggestions were implemented and retested.

These checks cover source integration and package construction. No fresh-install, signed channel transition or reboot qualification was performed in this follow-up. No shared feed or runtime release was published.


ryanrhughes:
Reviewed all five commits at 1b35de5, read against the surrounding code and quattro's `AGENTS.md`, with a second opinion from Codex at xhigh reasoning as an independent reviewer. Where Codex agreed with what I had already concluded, its independence is not guaranteed (it runs read-only but can read this session); its two findings below that I had not reached on my own stand on their own terms.

**Two defects, both fixed and pushed to this branch** (a623a31 and d67ef02), each with a test that fails against 1b35de5 and passes with the fix:

- The Install menu availability guard asked `pacman -Si` for every name in a row's transaction, which cannot see AUR packages or provided names. On a stock x86_64 install Chrome, Edge, Brave, Brave Origin and Zen (built by `omarchy-pkg-aur-add`) and Dropbox (`libappindicator-gtk3` is only ever provided, by `libappindicator`) all disappeared from Install, which the commit said could not happen. The five AUR rows now live in `optional-aur-packages.tsv` beside NordVPN and lose the sync guard; availability asks `pacman -Sp`, which resolves the way `-S` will, so provided names and constraints both count. Codex found this independently.
- Widening the battery helpers from `BAT*` to any `Battery` let a wireless mouse or keyboard stand in for the machine's battery: the kernel registers those as `type=Battery` with `scope=Device`, and UPower lists them as `battery_hidpp_battery_0` with `power supply: no`. `omarchy-battery-present` then answered yes on a desktop with a wireless mouse (which `lpmd.sh` and `thermald.sh` gate on), and `omarchy-battery-status` could report the mouse's charge on a laptop. `battery-present` now skips `scope=Device`; `battery-status` takes the first UPower device marked a power supply. Codex found this independently.

**Still open, for you:**

- The branch conflicts with quattro in `default/omarchy/omarchy-menu.jsonc`, `install/config/all.sh` (quattro dropped the `locate.sh` line) and `test/shell.d/network-manager-transition-test.sh` (quattro dropped the migration assertions). It needs a rebase before it can land.
- With the AUR rows unguarded, Edge is offered on aarch64 again and its AUR recipe declares `x86_64` only, so it fails at build there. Chrome, Brave, Brave Origin and Zen all declare `aarch64`. If you want Edge hidden on Apple Silicon, `optional-aur-packages.tsv` could carry an architecture column that `omarchy-install-available` and the menu prelude answer from `uname -m`; I left that to you since it extends the manifest you designed.
- The drift test excludes `omazed` and the kernel-headers packages by name, so a row whose recipe installs one of those can pass the test with an incomplete transaction. On the current repositories `omazed` is in the omarchy repo for both architectures, so nothing is hidden today.

**What was checked and where.** `test/cli` and the full `test/shell` suite ran on a credential-free Docker worker built from Arch, against this branch and against the merge base d3d23fd: the same 16 environment-bound files fail on both (fonts, docker, sudo, hybrid-GPU, provisioning), every one of the 12 new test files passes, and no file that passes on the base fails here. That worker has no compositor, so nothing compositor-backed ran, and nothing here was exercised on Apple Silicon hardware; the hardware, migration and keyring paths were read, not executed. The `omarchy-update-keyring` fingerprint read works as an unprivileged user (`/etc/pacman.d/gnupg/pubring.gpg` is world-readable on a stock install), and `omarchy-state set reboot-required` exists on quattro.

**Codex findings rejected:** it flagged the two added lines in `AGENTS.md` as an attempt to steer an automated reviewer; they document the two new helpers in the existing helper list, in the same shape as the entries around them, and I read the house rules from quattro rather than from this branch.

The tap-to-click default (commit 5) is a product call and I have not touched it. Overlaps worth knowing about: #8131 carries the same `apple-panel-bl` hunk, #7488 the same `multi-touch` hunk, #7577 gates the Broadcom quirk on `uname -m` and adds a repair migration for machines the old gate already reached (this branch gates on `omarchy-hw-apple-silicon` and has no repair), and #8039 touches `install/post-install/pacman.sh` and `install/user/mise-work.sh` for the same reason.

— 🤖 Claude, posting on behalf of @ryanrhughes

scottjones:
Detector follow-up (Naeem's architecture split: helpers invoked next to the script, match only `apple,*` device-tree entries, not a `pineapple,board` substring): omacom/omarchy#11325. Unique commit is on top of this overlay. Close that and cherry-pick onto here if you would rather keep a single PR.

scottjones:
Correction: the detector follow-up is stacked on this overlay's head, not a competing quattro PR. maralcbr/omarchy-mx-mac#99 — merge there and it lands in this PR.

ryanrhughes:
Re-reviewed at c260643. This answers what moved since d67ef02: the rebase onto quattro 31bd80d and the seven commits on top, again with Codex at xhigh as an independent reviewer of those commits only.

**Everything left open last time is resolved on this head.** The branch is mergeable-clean against quattro. Edge is gated off aarch64 through the architecture column in `optional-aur-packages.tsv`, and the shared helper in `install/helpers/optional-packages.sh` gives the CLI and the menu batch one implementation. Tap-to-click is now off only for `apple-mtp-multi-touch` and `apple-spi-trackpad` via `hl.device`, which is the same call `default/hypr/disabled-input-device.lua` already makes on quattro, so no other machine's default changes. The drift test no longer excludes `omazed` or the kernel headers, and the Xbox row derives its header from the same predicate the installer uses. Migration `1789172112` takes the Broadcom quirk back off Apple Silicon machines the old gate reached; its block text matches what the leaf and migration `1786391100` write byte for byte, and the pending marker survives a failed `mkinitcpio` and a second user. The two fixes I pushed came through the rebase unchanged apart from menu key order.

**Nothing new to fix.** I read the seven diffs against the surrounding code, checked that a failed `source` in the guard prelude makes the menu keep its last results rather than hide rows, that `pacman -Sp` works offline against the local sync database, and that every name in `optional-packages.tsv` except `libappindicator-gtk3` is in core, extra, multilib or the omarchy repo on stable, rc and edge (that one resolves through `-Sp` as intended).

**Codex finding rejected:** it reported `google-chrome` as x86_64-only on the AUR. The recipe declares `aarch64` with a `source_aarch64` pointing at Google's arm64 deb, so the table's `x86_64 aarch64` is right.

**What ran where.** `test/cli` and the full `test/shell` suite on a credential-free Docker worker, against this head and against the new merge base: the same 20 environment-bound files fail on both (docker, sudo, fonts, kitty, locate, hybrid GPU, provisioning), all 14 new test files pass, and no file that passes on the base fails here. The worker has no compositor, so the compositor-backed tests skipped and the Hyprland change rests on the Lua test plus the author's M1 Pro run. Nothing here executed on Apple Silicon.

For the maintainer: the follow-up scottjones stacked on this branch (maralcbr/omarchy-mx-mac#99) is closed unmerged, so the detector on this head is still the `apple,` substring match. No shipping device-tree vendor prefix contains that substring, so it is a tidiness point rather than a defect.

— 🤖 Claude, posting on behalf of @ryanrhughes

ryanrhughes:
Pushed da68d5e: Chrome is now x86_64 only in `optional-aur-packages.tsv`. Google ships an arm64 deb and the AUR recipe declares aarch64, but Omarchy does not build or test Chrome for aarch64, so the maintainer's call is that the row stays hidden there until it does. Chromium, the shipped default, is unaffected on both architectures. The availability test covers Chrome the same way it covers Edge and fails against the previous table.

Nothing else is open.

— 🤖 Claude, posting on behalf of @ryanrhughes

bin/omarchy-channel-set
```diff
@@ -41,6 +41,13 @@ link_dev_checkout() {
 
 (( $# > 0 )) || { usage; exit 1; }
 
+# Channels are pacman.conf/mirrorlist pairs, and every one of them is x86_64.
+# Until there is an aarch64 repository per channel, switching would swap the
+# Arch Linux ARM and Asahi repositories out from under the kernel.
+if omarchy-hw-apple-silicon; then
+  fail "There is no per-channel aarch64 repository for Apple Silicon yet; preserving the existing Arch Linux ARM repositories."
+fi
+
```
ryanrhughes: @maralcbr Even though this isn't live yet, should we just make it work properly? We could easily alias the other locations if need be.

bin/omarchy-update-keyring
```diff
@@ -3,20 +3,41 @@
 # omarchy:summary=Ensure the Omarchy and Arch keyring packages are installed and populated
 # omarchy:requires-sudo=true
 
-if omarchy-pkg-missing omarchy-keyring || ! sudo pacman-key --list-keys 40DFB630FF42BCFFB047046CF0134EE680CAC571 &>/dev/null; then
-  sudo pacman-key --recv-keys 40DFB630FF42BCFFB047046CF0134EE680CAC571 --keyserver keys.openpgp.org
-  sudo pacman-key --lsign-key 40DFB630FF42BCFFB047046CF0134EE680CAC571
+trusted_key=40DFB630FF42BCFFB047046CF0134EE680CAC571
+
+if omarchy-hw-apple-silicon; then
+  # Apple Silicon images ship omarchy-keyring with the installed system, so the
```
ryanrhughes: Can we simplify this? I believe it's the same on x86 as well. We ship the key with the ISO but this is meant to update it should you ever need. 

It just needs to install the pkgs and trust the key.

install/hardware/apple/fix-brcmfmac-supplicant.sh
```diff
@@ -18,6 +18,12 @@
 # machines including the T2-less iMac19,1 and iMac19,2, and BCM4377/4378/4387
 # from the T2 era on. The BCM4360 in 2013-2015 Macs is deliberately absent: it
 # runs the out-of-tree wl driver, which never reads a brcmfmac option.
+#
+# Apple Silicon is excluded even though its BCM4387 is in that list: there the
+# firmware supplicant is what works, and disabling it makes firmware commands
+# time out and breaks scanning entirely.
+omarchy-hw-apple-silicon && return 0
```
ryanrhughes: Won't this always just return 0? Feels like maybe it was meant to be `||`

install/hardware/apple/fix-spi-keyboard.sh
```diff
@@ -1,5 +1,11 @@
-# Detect MacBook models that need SPI keyboard modules
-product_name="$(cat /sys/class/dmi/id/product_name 2>/dev/null)"
+# Detect MacBook models that need SPI keyboard modules.
+# Apple Silicon has no DMI at all, and the install runs under set -e, so a
+# failed read has to land in an if rather than abort the whole hardware pass.
+if product_name=$(cat /sys/class/dmi/id/product_name 2>/dev/null); then
+  product_name=${product_name:-}
+else
+  product_name=""
+fi
```
ryanrhughes: What's the point of all of this when the comparison doesn't change?

install/hardware/network.sh
```diff
@@ -1,4 +1,15 @@
...
+# Apple Silicon Wi-Fi (brcmfmac on BCM4377/4378/4387) only works through iwd:
+# wpa_supplicant cannot drive the firmware's offloaded handshake. Keep
+# NetworkManager in charge, with iwd as its backend, and leave an existing
+# choice alone.
+if omarchy-hw-apple-silicon && [[ ! -e /etc/NetworkManager/conf.d/wifi_backend.conf ]]; then
+  install -Dm644 /dev/stdin /etc/NetworkManager/conf.d/wifi_backend.conf <<'EOF'
+[device]
+wifi.backend=iwd
+EOF
+fi
```
ryanrhughes: This should be shipped as a drop-in file so it just comes from omarchy-settings

install/helpers/optional-packages.sh
```diff
@@ -0,0 +1,92 @@
+# Shared by the standalone availability commands and the menu's Bash batch.
```
ryanrhughes: Do we need this?

install/post-install/pacman.sh
```diff
@@ -1,7 +1,11 @@
 # Configure pacman after package installation completes. Offline target package
 # installs use the live ISO's offline pacman.conf until this final restore.
-cp -f "$OMARCHY_PATH/default/pacman/pacman-${OMARCHY_MIRROR:-stable}.conf" /etc/pacman.conf
-cp -f "$OMARCHY_PATH/default/pacman/mirrorlist-${OMARCHY_MIRROR:-stable}" /etc/pacman.d/mirrorlist
+# Apple Silicon keeps the Arch Linux ARM and Asahi repositories that own its
+# kernel and firmware: the Omarchy x86_64 mirrorlist would leave it unbootable.
+if ! omarchy-hw-apple-silicon; then
```
ryanrhughes: Can we consolidate apple silicon here as well? 

If we can't we should leave a FIXME to address later

install/optional-aur-packages.tsv
ryanrhughes, line 1: WHy does this exist?

test/fixtures/optional-aarch64-required
ryanrhughes, line 1: There has to be a better way to handle these so we don't need these files

migrations/1788200002.sh
```diff
@@ -0,0 +1,8 @@
+echo "Install the PipeWire realtime scheduling provider on Apple Silicon"
+
+# PipeWire asks rtkit for realtime priority; without it audio threads run at
+# normal priority and drop out under load. The x86_64 package set already
+# carries it, the Arch Linux ARM install an Apple Silicon Mac started from
+# does not.
+omarchy-hw-apple-silicon || exit 0
+omarchy-pkg-add rtkit
```
malik-na: **[bug]** rtkit is installed only by this migration. `omarchy-provision-user --first-install` stamps every `migrations/*.sh` marker without running the scripts (`bin/omarchy-provision-user:124-127`), so a fresh Apple Silicon image never executes this. rtkit is not in `install/*.packages` and there is no install-time `omarchy-pkg-add rtkit` (omarchy-mac does that in `install/hardware/apple/audio.sh`). PipeWire's rtkit dependency is optional, so new Asahi machines keep PipeWire at normal priority. The test only exercises the migration as a standalone script and cannot see the first-install stamp.

**Suggestion:** Add a gated `omarchy-pkg-add rtkit` in an install hardware script (same pattern as `vulkan-asahi`), and keep the migration only for machines already installed. Alternatively put rtkit in the package set the image actually installs.

install/hardware/network.sh
```diff
@@ -1,4 +1,15 @@
...
 systemctl disable iwd.service 2>/dev/null || true
```
malik-na: **[bug]** On Apple Silicon the script writes `wifi.backend=iwd`, then the next line still runs the unconditional `systemctl disable iwd.service` that every Omarchy machine gets. omarchy-mac hit this on Asahi Alarm and documented the failure in `install/hardware/network.sh`: disabling iwd while NetworkManager is pointed at it leaves NM with a backend that never starts and no Wi-Fi devices. This PR does not skip that disable, does not `omarchy-pkg-add iwd`, and the "leave an existing choice alone" guard only looks at the exact path `wifi_backend.conf` — Asahi's drop-in often uses another name, so iwd.service is still disabled on machines that already had a working iwd backend.

**Suggestion:** Do not disable `iwd.service` when `omarchy-hw-apple-silicon` is true (or enable/start it after selecting the backend). Walk every `conf.d` file for an existing `wifi.backend` the way omarchy-mac does. Install `iwd` the same way `install/hardware/vulkan.sh` installs `vulkan-asahi`.

bin/omarchy-update-keyring
```diff
@@ -3,20 +3,41 @@
...
+  # Omarchy key is already in pacman's keyring. Fetching it from a keyserver
+  # would silently accept whatever the network handed back, so instead verify
+  # the fingerprint that is already trusted and fail loudly when it is not.
+  # Reading the keyring directly avoids another sudo prompt for a check.
+  installed_fingerprint=$(GNUPGHOME=/etc/pacman.d/gnupg gpg --batch --with-colons --fingerprint "$trusted_key" 2>/dev/null |
```
malik-na: **[bug]** The Apple Silicon branch reads the pacman keyring as the calling user (`GNUPGHOME=/etc/pacman.d/gnupg gpg ...`) specifically to avoid a sudo prompt. `omarchy-update` invokes this script unprivileged (`bin/omarchy-update:42`). pacman-key's homedir is created mode 700; gpg then fails, `installed_fingerprint` is empty, and the script exits 1 telling the user to reinstall `omarchy-keyring` — which blocks the rest of `omarchy update` on every Asahi machine. The x86 path already uses `sudo pacman-key --list-keys`. The new test stubs `gpg` on PATH, so it never sees the real directory mode.

**Suggestion:** Use `sudo pacman-key --list-keys` / `sudo gpg --homedir /etc/pacman.d/gnupg` for the fingerprint check, or fail open to the platform-keyring refresh when the homedir is unreadable rather than treating "cannot read" as "wrong key".

install/hardware/apple/fix-asahi-hid-race.sh
```diff
@@ -0,0 +1,33 @@
...
+echo "Detected Apple Silicon Mac: early-loading the Apple HID drivers"
+
+sudo mkdir -p /etc/mkinitcpio.conf.d
+
+# mkinitcpio fails the whole image over a MODULES entry it cannot find, which
+# would take every later rebuild with it, including the ones a kernel upgrade
+# triggers. So each driver is named only on kernels that build it as a module;
+# ending on unset keeps the exit status zero, which mkinitcpio reads as a
+# config that sources cleanly.
+sudo tee /etc/mkinitcpio.conf.d/apple_hid_modules.conf >/dev/null <<'EOF'
```
malik-na: **[suggestion]** The install leaf writes `/etc/mkinitcpio.conf.d/apple_hid_modules.conf` and stops. The only `mkinitcpio -P` is in migration `1788200001`, which first-install stamps as already applied. Hardware setup runs after kernel packages, and this PR skips snapper/Limine on Apple Silicon, so the usual later rebuild may never fire. New installs get a drop-in that does not affect the running initramfs until the next kernel upgrade; the race the docs describe remains on first boot.

**Suggestion:** Rebuild the initramfs in the leaf after writing the drop-in (as the migration already does), or have the image builder guarantee a rebuild and document that `install.sh` alone is not enough.

bin/omarchy-hw-apple-silicon
```diff
@@ -0,0 +1,11 @@
...
+
+# omarchy:summary=Detect whether the computer is an Apple Silicon Mac.
+
+# Apple Silicon Macs have no DMI, so omarchy-hw-match cannot see them. The
+# device tree Asahi's m1n1 hands the kernel names every board "apple,<model>",
+# which is the one identifier that survives across M1 through M4.
+proc_root="${OMARCHY_PROC_ROOT:-/proc}"
+
+[[ $(uname -m) == "aarch64" ]] &&
+  grep -aq 'apple,' "$proc_root/device-tree/compatible" 2>/dev/null
```
malik-na: **[suggestion]** `grep -aq 'apple,'` is an unanchored substring match on a NUL-separated compatible blob. A non-Apple aarch64 board whose compatible contained `pineapple,` would pass. Scottjones already called this out and stacked maralcbr/omarchy-mx-mac#99 / omacom/omarchy#11325 to match only `apple,*` entries.

**Suggestion:** Land that follow-up (NUL-split / `grep -z` / match `apple,` only at a token start) before this is the sole Apple Silicon detector everything else gates on.

default/hypr/input.lua
```diff
@@ -74,6 +74,10 @@ hl.config({
   },
 })
 
+-- Avoid stray taps while typing on the built-in Apple touchpad.
+hl.device({ name = "apple-mtp-multi-touch", tap_to_click = false })
+hl.device({ name = "apple-spi-trackpad", tap_to_click = false })
```
malik-na: **[suggestion]** `hl.device({ name = "apple-spi-trackpad", tap_to_click = false })` is not Apple-Silicon-only. Intel MacBook8,1 / MacBookPro13,x / MacBookPro14,x use the applespi driver and Hyprland names that device the same way. The PR text says no other machine's default changes; this line does.

**Suggestion:** Gate the SPI override on `omarchy-hw-apple-silicon` if it is only wanted on Asahi, or document that Intel SPI MacBooks are included on purpose.

install/hardware/apple/fix-asahi-hid-race.sh
```diff
@@ -0,0 +1,33 @@
+# Apple Silicon exposes the internal keyboard and trackpad as HID devices
```
malik-na: **[suggestion]** The 16-line header restates `docs/apple-silicon-trackpad.md` (dockchannel-hid, hid-generic rebind, ENOENT, libinput never retries). It narrates design history rather than a constraint the next reader cannot see from the drop-in itself.

**Suggestion:** Keep a one-line pointer to the doc and the `modinfo`/unset constraint; drop the duplicated cause/fix narrative.

install/hardware/apple/fix-brcmfmac-supplicant.sh
maralcbr: Agree. Rewrote it as an explicit if omarchy-hw-apple-silicon; then return 0; fi to make the intent clearer. Intel and Apple fixture tests pass.

install/hardware/apple/fix-spi-keyboard.sh
maralcbr: Agreed. Simplified it to product_name=$(cat /sys/class/dmi/id/product_name 2>/dev/null) || product_name="". The comparison stays unchanged; the fallback prevents missing or unreadable DMI from aborting hardware setup under errexit. Tests cover that failure path and matching Intel models.

install/hardware/network.sh
maralcbr: Agreed on package ownership. Removed the installer-written configuration and prepared a companion omarchy-settings-asahi package owning /usr/lib/NetworkManager/conf.d/wifi_backend.conf. The Apple-only add-on avoids changing x86 or generic ARM defaults and preserves administrator /etc overrides. This PR checks for the package during setup and adds a migration for existing installations. The companion package/profile changes still need to land before this runtime ships. Add-on installation, Wi-Fi reconnect and reboot passed on the M2 Max.

install/helpers/optional-packages.sh
maralcbr: Removed this helper, the generic install-availability wrapper and the transaction manifest. Menu conditions now name the complete package targets directly. Kept one small sourceable omarchy-pkg-available checker shared by the CLI and menu so caching and provider/version handling remain consistent. Tests cover secondary dependencies, source failures and CLI/menu parity.

install/post-install/pacman.sh
maralcbr: Consolidated configuration staging into a shared helper used by refresh and post-install. Finalization reuses only the configuration step, avoiding an online sync/upgrade during offline installation. Until ARM channels are qualified, it preserves the installed Apple configuration. Added the explicit FIXME for repository publication/aliases and qualification.

install/optional-aur-packages.tsv
maralcbr: Removed it. Architecture conditions now live directly in the menu: Chrome/Edge remain x86-only, and the other curated AUR entries retain their x86_64/aarch64 restrictions. Their existing installers remain responsible for package installation; AUR targets aren’t checked against sync repositories.

test/fixtures/optional-aarch64-required
maralcbr: Removed it from the installation tree and removed the count-only assertion. Availability now comes from explicit menu targets. I retained the expected ARM support cases as a test-only fixture so losing support doesn’t silently redefine the baseline. Tests also check secondary dependencies and compare menu targets with installer recipes. There is no production reader for the list anymore.

bin/omarchy-channel-set
maralcbr: Agreed. Added architecture-aware channel templates and shared configuration staging, with a repository/target preflight before changing live configuration. The ARM templates retain the Arch Linux ARM and Asahi repositories. Activation remains gated until the aliases are available and signed upgrade/downgrade/reboot testing passes; the FIXME names that remaining dependency.

bin/omarchy-update-keyring
maralcbr: Agreed. Removed the Apple presence-only shortcut and unified the update flow, varying only the platform keyring. Both paths now update and populate the keyring packages while retaining the pinned Omarchy fingerprint check. Also fixed failure propagation: previously a failed pacman command could still print “Keys are correct”; the regression test now verifies that it exits with the failure.

ryanrhughes:
Re-reviewed at 980ce7e, the five commits answering the nine review threads, with Codex at xhigh as an independent reviewer of those commits only. Six of the nine are resolved. One is not, and the way it was answered breaks every Apple Silicon machine, so this cannot land as it stands.

**Resolved:** the brcmfmac exclusion is an explicit `if` (3); the SPI DMI read is the one-liner (4); the shared helper, `omarchy-install-available`, both manifests and the aarch64-required list are gone from the install tree, and menu rows name their packages or architecture directly (6, 8, 9); the pacman configuration lives in one helper with the FIXME (7); the keyring script is one path for both architectures (2).

**Not resolved, and a regression (5).** `install/hardware/network.sh` now aborts hardware setup on Apple Silicon unless `omarchy-settings-asahi` is installed, and `migrations/1789275235.sh` installs that package on every existing Apple machine. The package does not exist: not in omacom/omarchy-pkgs, not in the fork's pkgbuilds, not in the aarch64 edge repository (which carries `omarchy-settings` and `omarchy-settings-dev` only). So a fresh Apple install fails in `omarchy-apply-hardware` under `set -e` with no iwd configuration written, and every existing Apple user's `omarchy update` stops at this migration and retries it forever. The previous head wrote the drop-in itself and worked. Until the companion package is published, this branch has to keep writing the file, or the package must land first; a runtime that depends on a package nobody can install is not a fix. Codex found this independently.

**Not really resolved (1).** `omarchy-channel-set` no longer refuses outright, but `qualified_arm_channels` is empty, so it still refuses every channel on Apple Silicon, and the new aarch64 templates are unreachable code. Two of the three channels they point at return 404 (`pkgs.omarchy.org/{stable,rc}/aarch64`). That is the same behaviour as before with more machinery around it, which is fine if the maintainer accepts "not yet" as the answer, but the thread should say so rather than "addressed".

**Two things the reorder and the rebase introduce:**

- `omarchy-channel-set` now runs `omarchy-refresh-pacman` before `link_dev_checkout`, so `omarchy-channel-set dev` refreshes from the installed release's templates instead of the checkout it is about to link. The test was changed to lock in the new order. That is a behaviour change on x86_64 for developers, unrelated to Apple Silicon, and should be its own decision.
- quattro has moved under the branch again (7 conflicting files). Two of them matter: #7807 already added `set -euo pipefail` and a final key check to `omarchy-update-keyring`, and 24417bf1 made matching kernel headers a base guarantee and dropped `linux-headers` from the Xbox installer. After the rebase the PR's header selection and the dynamic header target in the Xbox menu guard are obsolete, and the keyring change should be reconciled with quattro's version rather than replace it. The keyring script also runs `pacman-key --populate` before the keyring package is installed, which aborts under errexit if that package's keyring file is missing.

**Drop `docs/apple-review-integration.md`.** It is release-process narration and review bookkeeping, not reference on how the system is shaped, which is what `docs/` is for. Its last line, "Prepare a response for all nine PR review threads", is an instruction addressed to whoever is reading the repository. Fold anything durable into the commit messages and remove the file.

**What ran where.** `test/cli` and the full `test/shell` suite on a credential-free Docker worker at 980ce7e: the same 20 environment-bound files fail as on the merge base, every PR test passes. The network.sh and migration tests pass because they stub pacman; they prove the code fails closed when the package is absent, which is exactly the shipping condition. Nothing ran on Apple Silicon hardware. Codex's other findings agreed with what I had already established, so independence is not guaranteed there.

— 🤖 Claude, posting on behalf of @ryanrhughes

PR #9846, opened 2026-09-02
Clamp idle timeouts to QTimer-safe seconds

### LINKED ISSUE #9832: idle.lock above ~24.8 days overflows the QTimer interval negative: ~400 warnings/sec exhausts shmem and crash-loops wayland-wm-app-daemon
### What happened

A large `idle.lock` value in `~/.config/omarchy/shell.json` overflows the idle service's QTimer interval, producing a runaway Qt warning loop that eventually exhausts shmem/tmpfs and takes the desktop down.

`plugins/services/idle/Service.qml:267` binds the interval in milliseconds:

```qml
Timer {
  id: lockTimer
  interval: root.lockDelaySeconds * 1000
  repeat: false
  onTriggered: if (root.idleEnabled && root.idledThisCycle) root.lockSystem("lock-timeout")
}
```

`QTimer`/`QQmlTimer` intervals are a signed 32-bit int, so any `lockDelaySeconds` above `2147483` (~24.8 days) wraps negative. With `idle.lock: 2592000` (30 days — a natural way to approximate "never lock"):

```
lockDelaySeconds = 2592000 - 150 = 2591850
2591850 * 1000   = 2591850000 ms
INT32_MAX        = 2147483647
wraps to         = -1702967296
```

Qt clamps the negative interval to 1ms and logs a warning on every timer start:

```
omarchy-shell[1194]: WARN: QBasicTimer::start: negative intervals aren't allowed; the interval will be set to 1ms.
```

Measured on my machine: **~424 warnings/second**, `9,101,214` lines in a single session.

### Impact

This is not just log noise. The flood escalates into a system-wide failure:

1. Journald hits its 4G cap; weeks of real system history are rotated away.
2. Quickshell's log memfd fills, and `shmem`/tmpfs allocation starts failing:
   ```
   omarchy-shell: ERROR quickshell.logging: Failed to copy log from memfd with error code 28 "No space left on device"
   voxtype:       WARN Failed to write PID file: No space left on device (os error 28)
   uwsm_app-daemon: [Errno 28] No space left on device
   ```
3. `wayland-wm-app-daemon.service` crashes on the ENOSPC and enters a restart loop. Each restarted daemon receives a truncated request:
   ```
   uwsm_app-daemon: received:
   uwsm_app-daemon: sent: error 'Invalid arguments: ' 2
   ```
4. Every app launch then fails with a critical `App failure / Invalid arguments:` desktop notification (`urgency: 2`, `expireTimeout: 0`, so they never expire and persist across restarts).

The machine became unusable and needed a hard restart. Note that `df` shows plenty of free disk throughout — the ENOSPC is shmem exhaustion, which makes this quite hard to diagnose from the symptom.

### Steps to reproduce

1. Set in `~/.config/omarchy/shell.json`:
   ```json
   "idle": { "screensaver": 150, "lock": 2592000 }
   ```
2. Leave the machine idle for 150s so the idle cycle starts and `lockTimer` arms.
3. Watch the flood:
   ```bash
   journalctl -f | grep QBasicTimer
   omarchy-shell idle status   # "lockDelay": 2591850, "timers": { "lock": true }
   ```

### Expected

Either the timer is armed correctly for the configured duration, or the value is clamped/rejected — not silently wrapped into a 1ms busy loop.

### Suggested fix

`secondsFromConfig` in `plugins/services/idle/IdleModel.js` already guards the low end but not the high end:

```js
  var n = Number(value)
  return Math.floor(n)
}
```

Clamping to the largest value that survives the `* 1000` conversion would fix it:

```js
// Timer intervals are milliseconds in a signed 32-bit int.
var MAX_TIMER_SECONDS = 2147483  // floor(INT32_MAX / 1000)

  var n = Number(value)
  return Math.min(Math.floor(n), MAX_TIMER_SECONDS)
}
```

`screensaverDelaySeconds` (`Service.qml:260`) has the same exposure and would be covered by the same clamp.

Separately, it may be worth giving `idle.lock` an explicit "never" sentinel (e.g. `0` or `false`), since approximating it with a large number is what leads users into this.

### System details

| | |
|---|---|
| Omarchy | 4.0.2-1 |
| Kernel | 7.1.9-arch1-2 |
| CPU | Intel Core i5-10400F |
| GPU | NVIDIA GeForce GTX 1660 Ti (TU116) |
| quickshell | 0.3.1-1 |
| qt6-base | 6.11.2-2 |

### Related

- #9361 — critical notifications never auto-expire and have no close button. That's why the resulting `App failure` toasts pile up and survive restarts; this bug is one way to generate them in bulk.
- #9745 — different cause, same class of consequence (runtime tmpfs filling up and taking the session down).
- #9496 — also concerns `idle.screensaver` / `idle.lock` handling in `shell.json`.

fresh3nough:
## Summary

Fixes #9832.

`idle.lock` / `idle.screensaver` values above ~24.8 days overflow the idle service's QTimer interval. Qt stores timer intervals as a signed 32-bit millisecond int, so a 30-day "never lock" config (`2592000`) wraps negative, is clamped to 1ms, and floods ~400 `QBasicTimer::start: negative intervals aren't allowed` warnings per second until shmem/tmpfs is exhausted and the session dies.

## Change

`secondsFromConfig` in `shell/plugins/services/idle/IdleModel.js` already rejected non-finite and negative values. Clamp the high end to `floor(INT32_MAX / 1000)` (`2147483`) so both `lockTimer` and `screensaverTimer` intervals stay representable:

```js
var MAX_TIMER_SECONDS = 2147483

  var n = Number(value)
  return Math.min(Math.floor(n), MAX_TIMER_SECONDS)
}
```

Both timers bind `interval: *DelaySeconds * 1000` after this helper, so one clamp covers lock and screensaver.

## Test plan

- [x] Extended `test/shell.d/idle-test.sh` with ceiling, ceiling+1, and the 30-day overflow case from the issue
- [x] `bash test/shell.d/idle-test.sh` — all assertions pass, including `clamped idle timeouts stay within a signed 32-bit millisecond interval`
- [x] `bash -n test/shell.d/idle-test.sh` — clean
- [x] Confirmed pre-fix math: `2591850 * 1000` wraps to a negative int32; post-fix clamp yields `2147483000 <= INT32_MAX`

```
ok - idle floors configured seconds
ok - idle rejects negative seconds
ok - idle rejects invalid seconds
ok - idle accepts a zero timeout
ok - idle keeps the largest QTimer-safe timeout
ok - idle clamps just past the QTimer-safe ceiling
ok - idle clamps multi-week lock timeouts that would wrap QTimer intervals negative
ok - clamped idle timeouts stay within a signed 32-bit millisecond interval
```

### DIFF
--- a/shell/plugins/services/idle/IdleModel.js
+++ b/shell/plugins/services/idle/IdleModel.js
@@ -1,7 +1,12 @@
+// QTimer/QQmlTimer intervals are milliseconds in a signed 32-bit int.
+// Values above floor(INT32_MAX / 1000) wrap negative, Qt clamps them to 1ms,
+// and the idle service floods the journal until shmem is exhausted.
+var MAX_TIMER_SECONDS = 2147483
+
 function secondsFromConfig(value, fallback) {
   var n = Number(value)
   if (!isFinite(n) || n < 0) return fallback
-  return Math.floor(n)
+  return Math.min(Math.floor(n), MAX_TIMER_SECONDS)
 }
 
 function eventParts(event, count) {
@@ -45,6 +50,7 @@ function screensaverWindowsAfter(windows, address, visible) {
 
 if (typeof module !== "undefined") {
   module.exports = {
+    MAX_TIMER_SECONDS: MAX_TIMER_SECONDS,
     secondsFromConfig: secondsFromConfig,
     eventParts: eventParts,
     screensaverWindowsAfter: screensaverWindowsAfter
--- a/test/shell.d/idle-test.sh
+++ b/test/shell.d/idle-test.sh
@@ -10,6 +10,27 @@ const idle = requireFromRoot('shell/plugins/services/idle/IdleModel.js')
 assertEqual(idle.secondsFromConfig('42.9', 10), 42, 'idle floors configured seconds')
 assertEqual(idle.secondsFromConfig('-1', 10), 10, 'idle rejects negative seconds')
 assertEqual(idle.secondsFromConfig('nope', 10), 10, 'idle rejects invalid seconds')
+assertEqual(idle.secondsFromConfig(0, 10), 0, 'idle accepts a zero timeout')
+assertEqual(
+  idle.secondsFromConfig(idle.MAX_TIMER_SECONDS, 10),
+  idle.MAX_TIMER_SECONDS,
+  'idle keeps the largest QTimer-safe timeout'
+)
+assertEqual(
+  idle.secondsFromConfig(idle.MAX_TIMER_SECONDS + 1, 10),
+  idle.MAX_TIMER_SECONDS,
+  'idle clamps just past the QTimer-safe ceiling'
+)
+// 30 days is a common "never lock" approximation and overflows INT32 ms.
+assertEqual(
+  idle.secondsFromConfig(2592000, 10),
+  idle.MAX_TIMER_SECONDS,
+  'idle clamps multi-week lock timeouts that would wrap QTimer intervals negative'
+)
+assert(
+  idle.secondsFromConfig(2592000, 10) * 1000 <= 2147483647,
+  'clamped idle timeouts stay within a signed 32-bit millisecond interval'
+)
 
 assertDeepEqual(idle.eventParts({ data: 'a,b,c' }, 2), ['a', 'b', 'c'], 'idle parses raw event data')
 assertDeepEqual(


bjarneo:
Automated duplication check: this pull request looks similar to #7578, which covers the same idle timeout int32 clamp. I keep that one open and close this one to consolidate review.

PR #9849, opened 2026-09-02
Validate hook type and name before installing

Chessing234:
## Summary
- `omarchy-hook-install` joined unsanitized type/basename into `~/.config/omarchy/hooks`
- `../` segments could escape the hooks tree
- require a safe token shape before mkdir/cp

## Test plan
- [ ] `omarchy-hook-install '../x' ./hook` fails
- [ ] valid `post-update` install still works

### DIFF
--- a/bin/omarchy-hook-install
+++ b/bin/omarchy-hook-install
@@ -15,8 +15,24 @@ fi
 
 HOOK_TYPE=$1
 HOOK_FILE=$2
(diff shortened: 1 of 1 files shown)


Chessing234:

Chessing234:
folded into #12104 — same hook name validation, one review is enough.

PR #9851, opened 2026-09-02
Sanitize USER before building plugin clone ids

Chessing234:
## Summary
- clone ids are `$USER.$source` under the plugins dir
- a USER with path separators / `..` escapes that directory
- reject unsafe usernames before mkdir

## Test plan
- [ ] normal clone still creates `user.source` under plugins


Chessing234:

Chessing234:
folded into #12718 for easier review.

PR #9854, opened 2026-09-02
Keep the update lock out of world-writable /tmp

Chessing234:
## Summary
- update lock used `${XDG_RUNTIME_DIR:-/tmp}`
- fall back to `~/.local/state/omarchy` instead and chmod the directory

## Test plan
- [ ] `omarchy-update-lock run true` creates lock under runtime/state

### DIFF
--- a/bin/omarchy-update-lock
+++ b/bin/omarchy-update-lock
@@ -6,7 +6,9 @@
 
 set -e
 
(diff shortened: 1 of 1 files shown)


Chessing234:

Chessing234:

PR #9858, opened 2026-09-02
Preserve shell.json mode across atomic updates

Chessing234:
## Summary
- `omarchy-shell-config` replaced via mktemp+mv without preserving mode
- copy the existing mode (or 0600 for new files) before replace

## Test plan
- [ ] editing shell.json keeps prior permissions


Chessing234:

Chessing234:
folding into #12250 with the other config mode fixes

PR #9860, opened 2026-09-02
Hibernation: fail when unsupported; refuse empty resume= device

Chessing234:
## Summary
- Exit non-zero when the host cannot hibernate
- Folded #12729: refuse / repair drop-ins with empty `resume=` device

## Test plan
- [ ] Existing hibernation checks still apply
- [ ] `bash test/shell.d/hibernation-empty-resume-test.sh`


Chessing234:

PR #9862, opened 2026-09-02
Stage clipboard shares privately and clean them up

Chessing234:
## Summary
- clipboard share wrote into world-readable `/tmp` and never cleaned up
- stage under runtime/state at 0600, `--wait` for LocalSend, then remove

## Test plan
- [ ] `omarchy share clipboard` no longer leaves a file in `/tmp`


Chessing234:

Chessing234:

PR #9868, opened 2026-09-02
Restart fprintd after resume so fingerprint auth survives suspend

balazsorban44:
## Problem

On a laptop with fingerprint authentication configured, fingerprint stops working after the first suspend and stays broken until the next reboot — for `sudo`, for polkit, and for the lock screen. There is no error; PAM just falls through to the password prompt, so it reads as "the reader is broken".

`fprintd`/`libfprint` do not survive the USB re-enumeration a suspend causes. The daemon keeps its pre-sleep handle for the reader, and every later `Claim` fails:

```
fprintd[…]: Unexpected error while suspending device: The device is still busy with another operation, please try again later.
fprintd[…]: Async command sending failed: USB error on device 06cb:00fc : No such device (it may have been disconnected) [-4]
fprintd[…]: Authorization denied to :1.3883 to call method 'Claim' for device 'Synaptics Sensors': Device was already claimed
```

Reproduced on a ThinkPad X1 with a Synaptics Prometheus reader (`06cb:00fc`), Omarchy 4.0.2, systemd 261. The daemon had been stuck for three days; `systemctl restart fprintd` clears it immediately. `omarchy setup security fingerprint` configures the three PAM stacks but installs nothing that recovers the daemon on resume.

## Changes

Two atomic commits.

**1. `Install system-sleep hooks as executable`**

Found while following the existing convention for shipping a sleep hook. `systemd-sleep` runs only *executables* in `/usr/lib/systemd/system-sleep/` (`man 8 systemd-suspend.service`), but `omarchy-hibernation-setup` and `omarchy-toggle-hybrid-gpu` install their hooks with `cp -p`, which preserves the repository's 644 mode and the invoking user's ownership:

```
$ ls -l /usr/lib/systemd/system-sleep/
-rw-r--r-- 1 balazs balazs  472 Mar 26 16:52 keyboard-backlight   # cp -p, never runs
-rwxr-xr-x 1 root   root   1148 Aug 31 05:11 unmount-fuse         # packaged, runs
```

So `keyboard-backlight` and `force-igpu` are skipped on every suspend, silently. The commit marks the shipped hooks executable, installs them with an explicit mode and owner, and adds a migration that repairs an existing non-executable copy.

**2. `Restart fprintd after resume to keep fingerprint auth working`**

Ships `default/systemd/system-sleep/fprintd-reset`, installed by `omarchy-setup-security-fingerprint`, removed by `omarchy-remove-security-fingerprint`, plus a migration for installs that already enrolled a print.

The restart runs on `post`, not `pre`. A pre-suspend stop loses the race: `fprintd` is D-Bus activated, so the shell polkit and lock plugins re-activate it before the machine goes down. `try-restart` is a no-op while the daemon is idle, which is its normal state.

## Testing

- `./test/all` — no new failures. The 4 pre-existing failures reproduce identically on unmodified `quattro`: `config`, `snapper` and `unowned-system-paths` need an `omarchy-pkgs` checkout (`OMARCHY_PKGS_PATH`) that is not public, and `runtime-smoke` is the graphical Quickshell test.
- `/usr/lib/systemd/system-sleep` is already in the `allowed` set of `unowned-system-paths-test.sh`, under the "hardware-conditional sleep hooks, installed only on the machines that need them" rationale, which is exactly how this hook is installed.
- Verified on the affected machine that restarting `fprintd` on resume is what clears the stale claim: before the restart `fprintd-verify` fails with `AlreadyInUse`; after it, `Verify started!` and the reader works for sudo, polkit and hyprlock.


bjarneo:
Automated duplication check: this pull request looks similar to #7158, which covers the same fprintd resume hook. I keep that one open and close this one to consolidate review.

PR #9869, opened 2026-09-02
Add OpenClaw as an AI agent option

mbot11:
# Add OpenClaw as an AI agent option

## Summary

Adds [OpenClaw](https://openclaw.ai/) (formerly Clawdbot/Moltbot) to the AI
coding-agent roster on `quattro`, following the exact conventions of the
existing agents:

- lazy-loaded launcher via mise (`agent_package="npm:openclaw"`)
- a row in **Setup > Defaults > Agent** with the standard checkmark guard
- an unattended launch mapping in `omarchy-agent` (`openclaw tui`, prompt
  seeded via `--message`)
- an agents-panel usage collector matching the packaged collector record
  contract (`schemaVersion: 1`), writing `openclaw.json` next to
  `claude.json` and `codex.json`

## Why

OpenClaw is one of the most widely used agent harnesses, and its terminal UI
(`openclaw tui`) is a full agent surface with tool approvals. Users who run it
alongside Claude Code / Codex / OpenCode currently get no agents-panel usage
tracking and no default-agent option. This patch closes both gaps the same way
every other agent is wired.

## Changes

1. `bin/omarchy-default-agent` — `openclaw` / `open-claw` case, usage strings
2. `bin/omarchy-agent` — launch case: `openclaw tui` (gateway-connected TUI),
   `--message "$prompt"` for prompt seeding
3. `default/omarchy/omarchy-menu.jsonc` — `setup.default.agent.openclaw` row
4. `bin/omarchy-agent-usage-openclaw` — new usage collector (same record
   contract as the packaged collectors; discovers sessions via
   `openclaw sessions --json --all-agents` and prompt counts from gateway
   audit events)

## Notes for review

- OpenClaw is a gateway-first tool; `openclaw tui` talks to the local Gateway
  (loopback) and works headless, so the launch mapping needs no special flags.
- The usage collector shells out to the `openclaw` CLI rather than reading its
  state SQLite directly (the store uses the `sqlite-vec` extension), which
  also keeps it stable across OpenClaw schema changes.
- The collector only reports metadata (counts, model names, token totals) —
  no message content, credentials, or provider keys are read or written.

## Testing

- `bash -n` on both modified scripts; JSONC parse on the modified menu
- `openclaw tui` launches against a running gateway and accepts `--message`
- collector verified against a live gateway: correct today/session/token
  aggregates, atomic write to `~/.local/state/omarchy/agents/usage/openclaw.json`

### DIFF
--- a/bin/omarchy-agent
+++ b/bin/omarchy-agent
@@ -104,6 +104,10 @@ ori)
   command=(ori code)
   [[ -n ${prompt:-} ]] && command+=(--interactive --prompt "$prompt")
   ;;
+openclaw)
+  command=(openclaw tui)
+  [[ -n ${prompt:-} ]] && command+=(--message "$prompt")
+  ;;
 pi)
   command=(pi)
   [[ -n ${prompt:-} ]] && command+=("$prompt")
--- a/bin/omarchy-agent-usage-openclaw
+++ b/bin/omarchy-agent-usage-openclaw
@@ -1,0 +1,178 @@
+#!/usr/bin/python3
+# omarchy:summary=Print the OpenClaw Agent usage record as JSON
+# omarchy:args=[--write]
+"""Collect OpenClaw Gateway usage into one display-ready JSON record.
+
+Everything the agents panel shows for OpenClaw comes from this one command:
+session stats (sessions, per-model input/output tokens) from
+`openclaw sessions --json --all-agents` and run counts from the gateway audit
+log (`openclaw audit`, agent.run.started events). It prints the same record
+contract as the packaged claude/codex collectors; with --write it updates the
+record file atomically.
+
+Chat sessions that arrived over messaging gateways count too: the gateway
+treats them as first-class sessions and so does this record.
+"""
+
+from __future__ import annotations
+
+import json
+import os
+import shutil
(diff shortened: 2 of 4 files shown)


smfworks:
Cross-referencing with #7919, which wires the same agent entries (`bin/omarchy-agent`, `bin/omarchy-default-agent`, `omarchy-menu.jsonc`) plus the full install/remove/onboarding flow: these two PRs edit the same files from different bases and will conflict.

The unique piece here is real and needed: there is no `omarchy-agent-usage-openclaw` on main, and the packaged collectors (`claude`, `codex`, `fireworks`) are what feed the agents usage panel — without one, OpenClaw shows up in the panel with no stats. I verified this collector's CLI calls against openclaw 2026.9.4: `sessions --json --all-agents --limit all` and `audit --json` + `--cursor` pagination are all valid flags, and the record matches the packaged contract (schemaVersion 1, same field set).

Suggestion: drop the duplicated agent wiring from this PR (that's #7919's territory) and either rebase this collector on top of #7919's wiring once it lands or offer it to #7919 directly. The collector is worth keeping; the overlapping wiring will just burn review time on conflicts.

PR #9870, opened 2026-09-02
Fix monospace font preference ordering

derluke:
## Summary

- keep JetBrainsMono as the packaged monospace default while leaving the generic family in the pattern for a later user override
- scope both default and selected-font rules to requests where `monospace` is the first family, preserving explicit application fonts
- migrate only the exact Fontconfig file generated by the previous `omarchy font set` implementation

## Why

The package rule currently assigns JetBrainsMono before `50-user.conf` loads. That removes `monospace` from the pattern, so the selected-font rule never matches. Appending the packaged default keeps the generic family available for the user rule; qualifying the first family prevents a recurrence of the app-specific font override fixed in #5700.

## Testing

- `bash test/shell.d/fontconfig-test.sh`
- `./test/cli`


bjarneo:
Automated duplication check: this pull request looks similar to #8449, which covers the same font mono preference. I keep that one open and close this one to consolidate review.

PR #9876, opened 2026-09-02
Fix nonfunctional brightness sliders on some DDC monitors

### LINKED ISSUE #8931: Display panel brightness slider does not work with LG UltraGear evo 27GM950B-B
## System details

AMD Ryzen 7 8845HS with Radeon 780M Graphics (amdgpu), LG UltraGear evo 27GM950B-B connected via DisplayPort (DP-1), Omarchy 4.0.1-1, Hyprland 0.56.2, Linux 7.1.9-arch1-2, ddcutil 2.2.7.

Diagnostic log: https://logs.omarchy.org/6F0h7h

## What's wrong?

The Display panel detects the external monitor and shows a brightness slider, but dragging the slider does not change the monitor's brightness. The panel also reports success because the underlying helper exits successfully, although the monitor's VCP brightness value remains unchanged.

This is reproducible with the packaged DDC helper:

```text
$ ddcutil --bus 9 getvcp 10 --brief
VCP 10 C 100 100

$ omarchy-brightness-display --no-osd --monitor DP-1 70%
$ echo $?
0

$ ddcutil --bus 9 getvcp 10 --brief
VCP 10 C 100 100
```

The same monitor responds correctly when the DDC checks are not skipped:

```text
$ ddcutil --bus 9 setvcp 10 70
$ ddcutil --bus 9 getvcp 10 --brief
VCP 10 C 70 100
```

The fast path used by Omarchy reproduces the failure:

```text
$ ddcutil --bus 9 --skip-ddc-checks --noverify setvcp 10 70
$ ddcutil --bus 9 getvcp 10 --brief
VCP 10 C 100 100
```

The packaged helper `/usr/share/omarchy/bin/omarchy-brightness-display-ddc` currently invokes:

```bash
ddcutil --bus "$bus" --skip-ddc-checks --noverify setvcp 10 "$raw_target"
```

It appears that this monitor accepts the command without an error but ignores the write when `--skip-ddc-checks` is used. The normal `ddcutil` path successfully changes VCP 0x10. A possible fix would be to retry without `--skip-ddc-checks` when the requested value is not applied, or otherwise allow a per-monitor fallback.

The monitor is identified by `ddcutil` as LG ULTRAGEAR+, product code 23800, serial 605NTUWKV886, on I2C bus 9; VCP 0x10 is reported as writable with a range of 0–100.

Expected: dragging the Display panel brightness slider changes the physical monitor brightness.

Actual: the slider moves in the UI, but the physical brightness remains unchanged.

Murderlon:
Fixes #8931

Upstream investigation: https://github.com/rockowitz/ddcutil/issues/626

> [!Note] 
> The AI reviews in the comments evaluated the earlier two-process fallback, which this version supersedes now that an upstream fix happened

## Summary

1. Verify external DDC brightness writes before reporting success.
2. Retry one rejected set-and-verify cycle in the same ddcutil process with `--max-setvcp-and-verify-tries 2`.
3. Keep rejected displays readable, snap the slider back to the last confirmed value, and suppress the OSD when a write fails.

The affected LG ignores the first fast write. Its verification read primes the display, so the second write succeeds when it reuses the same display handle. The public retry option shipped in ddcutil 3.0.0 as the replacement for the hidden `--i5` option; Omarchy targets ddcutil 3.0.1+, now packaged in Arch Extra.

A display that rejects every write still returns a failure without hiding its brightness slider. The helper remains stateless and uses one ddcutil process per write.

## Out of scope

No ddcutil call has a timeout, so a blocked ddcutil holds the brightness key lock for as long as it blocks. That predates this change and is left for a follow-up.

## Testing

1. `./test/shell.d/brightness-display-test.sh`
2. `./test/shell.d/brightness-panel-test.sh`
3. `bash test/shell.d/bin-style-test.sh`
4. `bash test/shell.d/manifest-entrypoints-test.sh`
5. `./test/cli`
6. Verified the equivalent same-process retry (`--i5 2`) on the affected LG with ddcutil 2.2.7 before upstream published the named option.

### DIFF
--- a/bin/omarchy-brightness-display-ddc
+++ b/bin/omarchy-brightness-display-ddc
@@ -12,14 +12,12 @@ step="${2:-}"
 cache_dir="${XDG_RUNTIME_DIR:-/tmp}/omarchy-brightness-display-ddc"
 cache_name="${monitor//[^[:alnum:]_.-]/_}"
 cache_file="$cache_dir/$cache_name.bus"
-full_checks_cache_file="$cache_dir/$cache_name.full-checks"
 unavailable_cache_seconds=60
 range_cache_seconds=10
 
 cache_unavailable() {
   mkdir -p "$cache_dir" 2>/dev/null || true
   printf 'unavailable %s\n' "$(date +%s)" >"$cache_file" 2>/dev/null || true
-  rm -f "$full_checks_cache_file"
 }
 
 detect_bus() {
@@ -103,28 +101,11 @@ read_brightness() {
   printf '%s %s\n' "$bus" "$values"
 }
 
+# Some displays ignore a write on the fast path but accept it once the initial
+# DDC checks run, so retry with full checks before reporting a rejected write.
 write_vcp() {
-  local bus="$1"
-  local raw_target="$2"
-  local full_checks_bus=""
-
-  if [[ -r $full_checks_cache_file ]]; then
-    read -r full_checks_bus <"$full_checks_cache_file" || true
-  fi
-
-  if [[ $full_checks_bus == "$bus" ]]; then
-    if ddcutil --bus "$bus" --verify setvcp 10 "$raw_target" >/dev/null 2>&1; then
-      return 0
-    fi
-  elif ddcutil --bus "$bus" --skip-ddc-checks --verify setvcp 10 "$raw_target" >/dev/null 2>&1; then
-    return 0
-  elif ddcutil --bus "$bus" --verify setvcp 10 "$raw_target" >/dev/null 2>&1; then
-    printf '%s\n' "$bus" >"$full_checks_cache_file" 2>/dev/null || true
-    return 0
-  fi
-
-  cache_unavailable
-  return 1
+  ddcutil --bus "$1" --skip-ddc-checks --verify setvcp 10 "$2" >/dev/null 2>&1 ||
+    ddcutil --bus "$1" --verify setvcp 10 "$2" >/dev/null 2>&1
 }
 
 bus=""
--- a/bin/omarchy-brightness-display-ddc
+++ b/bin/omarchy-brightness-display-ddc
@@ -12,12 +12,14 @@ step="${2:-}"
 cache_dir="${XDG_RUNTIME_DIR:-/tmp}/omarchy-brightness-display-ddc"
 cache_name="${monitor//[^[:alnum:]_.-]/_}"
 cache_file="$cache_dir/$cache_name.bus"
+full_checks_cache_file="$cache_dir/$cache_name.full-checks"
 unavailable_cache_seconds=60
 range_cache_seconds=10
 
 cache_unavailable() {
   mkdir -p "$cache_dir" 2>/dev/null || true
   printf 'unavailable %s\n' "$(date +%s)" >"$cache_file" 2>/dev/null || true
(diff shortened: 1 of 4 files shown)


Murderlon:
Everything except the timeout is addressed in the two commits above.

Finding 1: I chose to keep the display readable. A rejected write now simply fails. On a monitor that rejects every write the slider stays visible and snaps back to the real value, and no OSD is shown. No write only negative cache, since the key lock and the single in flight write in the panel already bound repeat attempts.

Finding 3: I removed the marker entirely. Every write probes the fast path and retries with full checks, so nothing can latch. Displays that need full checks pay one extra failed attempt per write, a fraction of a second on a path that is slow either way.

Finding 2: The wheel path now shows the OSD from the exit handler only when the helper exits zero. A failed write drops any queued follow up and refreshes so the slider snaps back.

Finding 5: The tests assert that a rejected fast write is retried with full checks and that a read still succeeds afterwards. Each mutation now fails a named assertion.

Finding 4: Left out as pre existing and noted in the description.

Murderlon:
A rejected write made the panel read its state again. An empty read clears `brightnessAvailable` and hides the slider until the next poll. The panel now keeps the last percentage that the hardware confirmed. A rejected write puts the slider back to that percentage and does not read the state again.

A write in flight now owns `brightnessPercent`. The 5-second poll does not change that value during a write. Therefore `pendingBrightnessPercent` and `brightnessSetQueued` are not necessary. The panel has four brightness properties, not six.

`test/shell.d/brightness-panel-test.sh` is new. It has four assertions for the behaviour of the three commits: the OSD comes only after a verified write, and a rejected write puts the slider back without a new read.

## Hardware test

I ran the commands from this branch on the affected display (LG UltraGear 27GM950B-B, DisplayPort). After each command I read the actual value with `ddcutil getvcp 10`.

| command              | exit | hardware |
| -------------------- | ---- | -------- |
| `60%`                | 0    | 80 -> 60 |
| `+5%`                | 0    | 60 -> 65 |
| `5%-`                | 0    | 65 -> 60 |
| `35%`                | 0    | 60 -> 35 |
| `50%` (with the OSD) | 0    | 35 -> 50 |

I did not run the QML in a live shell, because this desktop uses the packaged shell. The helper is verified on the hardware. The panel behaviour is covered by the tests only.

## Upstream status (rockowitz/ddcutil#626)

The maintainer replied on 7 September. He added the option `--max-setvcp-and-verify-tries` to branch 2.2.8-dev. It replaces the hidden option `--i5`.

We cannot use the new option now. Arch Linux supplies ddcutil 2.2.7, and 2.2.7 stops with `Unknown option --max-setvcp-and-verify-tries`. ddcutil 2.2.8 is not released. Therefore this PR keeps the retry with the full DDC checks, which is correct on 2.2.7 and on 2.2.8.

Murderlon:
Great news! The upstream issue paid off and ddcutil 3.0.1 is now released and packaged in Arch Extra, so this PR now uses the public `--max-setvcp-and-verify-tries 2` replacement for `--i5 2`. The two-process full-checks fallback is removed, verification and its one retry now share the same display handle.

Note: this does expect the new version to be present for this to work. Currently `ddcutil` is unpinned, we can optionally pin `ddcutil>=3.0.1`

PR #9878, opened 2026-09-02
Re-apply hardware pacman repos after a refresh restore

### LINKED ISSUE #9853: omarchy-refresh-pacman drops hardware-specific repos (arch-mact2 on T2 Macs), silently orphaning linux-t2
**System details:** Apple MacBookPro16,1 (T2, `106b:1801`/`106b:1802`), Intel i7-9750H + AMD Navi 14, Omarchy 4.0.2-1, kernel `linux-t2` 7.1.8.arch1-3, Hyprland 0.56.2, bootloader Limine (UKI).

## Summary

`omarchy-refresh-pacman` overwrites `/etc/pacman.conf` from the channel template but never re-applies the hardware-specific repository additions that `install/hardware/pacman.sh` is responsible for. On Apple T2 Macs this silently drops the `[arch-mact2]` repository, orphaning `linux-t2` and the other T2 packages so they can no longer receive updates — including the kernel.

The install path does this correctly; the refresh/upgrade path does not.

## Impact

After upgrading to 4.x on a T2 Mac, these packages remain installed but have no update source:

- `linux-t2` — **the running kernel**
- `t2fanrd`
- `apple-t2-audio-config`
- `apple-bcm-firmware`

The machine keeps booting and working, so there is no visible symptom. The failure is silent: the kernel simply stops receiving security and bugfix updates, and the user has no indication anything is wrong. `pacman -Syu` reports nothing missing because the packages resolve locally.

## Root cause

`bin/omarchy-refresh-pacman` overwrites the config and then only runs the user hook:

```bash
sudo cp -f "$OMARCHY_PATH/default/pacman/pacman-$channel.conf" /etc/pacman.conf
sudo cp -f "$OMARCHY_PATH/default/pacman/mirrorlist-$channel" /etc/pacman.d/mirrorlist

# Allow user customization of /etc/pacman.conf before the upgrade runs
omarchy-hook pre-refresh-pacman

sudo env OMARCHY_UPDATE_PACMAN=1 pacman -Syyuu --noconfirm
```

`install/post-install/pacman.sh` does the same overwrite but *does* re-apply hardware repos:

```bash
cp -f "$OMARCHY_PATH/default/pacman/pacman-${OMARCHY_MIRROR:-stable}.conf" /etc/pacman.conf
...
source "$OMARCHY_INSTALL/hardware/pacman.sh"
```

And `install/hardware/pacman.sh` states the intent explicitly:

```bash
# Hardware-specific pacman repository extensions that must survive the final
# pacman.conf restore.
if lspci -nn | grep "106b:180[12]" >/dev/null; then
  if ! grep -q '^\[arch-mact2\]' /etc/pacman.conf; then
    cat >> /etc/pacman.conf <<'EOF'

[arch-mact2]
Server = https://github.com/NoaHimesaka1873/arch-mact2-mirror/releases/download/release
SigLevel = Never
EOF
  fi
fi
```

So the repo is added deliberately for this hardware and is meant to survive a template restore — but the refresh path is missing the `source "$OMARCHY_INSTALL/hardware/pacman.sh"` line. No migration re-adds it either (`grep -rn arch-mact2 migrations/` returns nothing).

Note the ordering hazard: `pacman -Syyuu --noconfirm` runs immediately afterwards, so the full upgrade proceeds with the repo already missing.

## Reproduction

1. Apple T2 Mac (`lspci -nn` shows `106b:1801` / `106b:1802`) installed via the T2 path, with `[arch-mact2]` in `/etc/pacman.conf`
2. Upgrade to 4.x (or run `omarchy refresh pacman`)
3. `grep arch-mact2 /etc/pacman.conf` → gone
4. `pacman -Si linux-t2` → no repository

## Evidence from the affected machine

`/etc/pacman.conf` before the upgrade (`/etc/pacman.conf.bak`, 15:53) ended with:

```
[omarchy]
SigLevel = Optional TrustAll
Server = https://pkgs.omarchy.org/stable/$arch

[arch-mact2]
Server = https://github.com/NoaHimesaka1873/arch-mact2-mirror/releases/download/release
SigLevel = Never
```

The upgrade's own backup (`/etc/pacman.conf.omarchy-upgrade-to-quattro.20260902155457.bak`, 15:54) already ends at `[omarchy]` — `[arch-mact2]` is gone.

Resulting state:

```
$ pacman -Q linux-t2
linux-t2 7.1.8.arch1-3

$ pacman -Si linux-t2
error: package 'linux-t2' was not found
```

`arch-mact2-mirrorlist` was also removed.

## Suggested fix

Re-apply hardware repository additions in `omarchy-refresh-pacman` after the template copy and before `pacman -Syyuu`, mirroring what `install/post-install/pacman.sh` already does:

```bash
sudo cp -f "$OMARCHY_PATH/default/pacman/pacman-$channel.conf" /etc/pacman.conf
sudo cp -f "$OMARCHY_PATH/default/pacman/mirrorlist-$channel" /etc/pacman.d/mirrorlist

source "$OMARCHY_INSTALL/hardware/pacman.sh"   # <-- add

omarchy-hook pre-refresh-pacman

sudo env OMARCHY_UPDATE_PACMAN=1 pacman -Syyuu --noconfirm
```

`install/hardware/pacman.sh` is already idempotent (it guards on `grep -q '^\[arch-mact2\]'`), so this is safe to run unconditionally on every refresh.

A migration to repair already-upgraded machines would also help, since affected users have no symptom to prompt them to look.

## Workaround

Re-append the stanza to `/etc/pacman.conf` manually, or add a
`~/.config/omarchy/hooks/pre-refresh-pacman.d/` hook that re-adds it on each refresh.

## Note on scope

The removal of `tiny-dfr` by migration `1785944594.sh` is **not** part of this report — that is clearly deliberate and well-reasoned ("the optional daemon holds stale device descriptors across suspend with t2bce"), and matches this machine's experience. Only the loss of the repository appears unintended.

---

*This issue was researched and written by Claude (Claude Code) running on the affected machine, and filed by the repository account owner. All command output above was collected from that machine.*

hudsonwa:
Fixes #9853

## What broke

`omarchy refresh pacman` overwrites `/etc/pacman.conf` with the channel template and never re-applies hardware-specific repository additions. On Apple T2 Macs that silently drops the `[arch-mact2]` stanza, so `linux-t2` (the running kernel), `t2fanrd`, `apple-t2-audio-config` and `apple-bcm-firmware` lose their update source with no visible symptom.

The install path does this correctly: `install/post-install/pacman.sh` sources `install/hardware/pacman.sh`, which appends the T2 stanza and is idempotent (it guards on `grep -q '^\[arch-mact2\]'`). That same file states the intent in its own comment: the extension "must survive the final pacman.conf restore". The refresh path never copies it.

## Fix

After the template restore, source `$OMARCHY_PATH/install/hardware/pacman.sh` the same way the install path does:

```bash
if [[ -f $OMARCHY_PATH/install/hardware/pacman.sh ]]; then
  sudo bash -c "source '$OMARCHY_PATH/install/hardware/pacman.sh'"
fi
```

Notes on why this shape:

- The issue's suggested `source "$OMARCHY_INSTALL/hardware/pacman.sh"` cannot work in `bin/omarchy-refresh-pacman`: `$OMARCHY_INSTALL` is not defined at runtime (only `omarchy-apply-hardware`/`omarchy-apply-system` set it). At runtime the tree lives at `$OMARCHY_PATH/install/` (docs/file-layout.md maps `install/**` to `/usr/share/omarchy/install/`).
- The file appends to `/etc/pacman.conf`, so it must run as root; the copy steps just above already use `sudo`, so the script runs as the user and elevates.
- Placed before the `pre-refresh-pacman` user hook, matching the issue's ordering, so user hooks see the final repo set.
- The `[[ -f ]]` guard keeps the refresh a no-op on anything where the install tree is absent, and if `install/hardware/pacman.sh` is ever removed the refresh simply skips it.

## Security context

`install/hardware/pacman.sh` is also in the sights of open security PRs #8709 (arms verification on the T2 stanza) and #9461 (draft: remove the unsigned source). This PR deliberately does not take a side: it sources whatever the packaged file contains at runtime, so if #8709 lands the refresh re-applies the armed stanza, and if a removal lands the `-f` guard no-ops. Only the "refresh drops hardware repos" bug is fixed here.

## Tests

New `test/shell.d/refresh-pacman-test.sh` runs the command against a sandboxed `/etc/pacman.conf` (absolute paths rewritten) with stubbed `lspci`/`sudo`/`pacman`/`omarchy-hook`:

- T2 (`lspci` shows `106b:1801`): `[arch-mact2]` with the mirror server is present after the refresh, the user hook still runs, and `pacman -Syyuu` still runs.
- Second run: the stanza is not duplicated.
- Non-T2: nothing is added.

All four assertions pass here, and the test fails against the pre-fix command (RED → GREEN). Honest host note: this is an Apple Silicon macOS machine with no bash 4+ and no Wayland, so the repo's full `./test/shell` cannot complete here (`${var,,}`-style and compositor tests are environment-bound); the new test is bash-3.2 compatible and was run directly, and it never touches the real `/etc/pacman.conf`. Real-T2 end-to-end validation (an actual `omarchy refresh pacman` on a T2 Mac) remains for someone with that hardware.

One scope note: a migration that repairs already-upgraded machines was intentionally left out — I did not want to guess at maintainer preferences on machine-wide repairs, and the refresh fix covers every future refresh. Happy to add one if you want it.

### DIFF
--- a/bin/omarchy-refresh-pacman
+++ b/bin/omarchy-refresh-pacman
@@ -23,8 +23,13 @@ sudo cp -f "$OMARCHY_PATH/default/pacman/mirrorlist-$channel" /etc/pacman.d/mirr
 # Macs) must survive the template restore above; the install path re-applies
 # them from this same file. It is idempotent and no-ops on other hardware,
 # but appends to /etc/pacman.conf, so it has to run as root like the copies.
+# A failure here is exactly the silent-failure class this issue is about:
+# the refresh must stop rather than run the upgrade with hardware repos gone.
 if [[ -f $OMARCHY_PATH/install/hardware/pacman.sh ]]; then
-  sudo bash -c "source '$OMARCHY_PATH/install/hardware/pacman.sh'"
+  if ! sudo bash -c "source '$OMARCHY_PATH/install/hardware/pacman.sh'"; then
+    echo "Failed to re-apply hardware pacman repositories; aborting refresh" >&2
+    exit 1
+  fi
 fi
 
 # Allow user customization of /etc/pacman.conf before the upgrade runs
--- a/bin/omarchy-refresh-pacman
+++ b/bin/omarchy-refresh-pacman
@@ -19,6 +19,14 @@ echo
 sudo cp -f "$OMARCHY_PATH/default/pacman/pacman-$channel.conf" /etc/pacman.conf
 sudo cp -f "$OMARCHY_PATH/default/pacman/mirrorlist-$channel" /etc/pacman.d/mirrorlist
 
+# Hardware-specific repository additions (e.g. [arch-mact2] on Apple T2
+# Macs) must survive the template restore above; the install path re-applies
+# them from this same file. It is idempotent and no-ops on other hardware,
+# but appends to /etc/pacman.conf, so it has to run as root like the copies.
+if [[ -f $OMARCHY_PATH/install/hardware/pacman.sh ]]; then
+  sudo bash -c "source '$OMARCHY_PATH/install/hardware/pacman.sh'"
+fi
+
 # Allow user customization of /etc/pacman.conf before the upgrade runs
 omarchy-hook pre-refresh-pacman
 
--- a/test/shell.d/refresh-pacman-test.sh
+++ b/test/shell.d/refresh-pacman-test.sh
@@ -61,7 +61,7 @@ run_refresh() {
     HOOK_LOG="$tmp_dir/hook.log" \
     OMARCHY_PATH="$tmp_dir/omarchy" \
     PATH="$tmp_dir/bin:$PATH" \
-    bash -euo pipefail "$tmp_dir/leaf.sh"
+    bash "$tmp_dir/leaf.sh"
 }
 
 # On Apple T2 hardware (lspci shows 106b:1801) the [arch-mact2] repository must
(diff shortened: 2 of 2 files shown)


hudsonwa:
Thanks for the worker verification, it is exactly the scrutiny this needed. I took the exit-status point further: since the whole bug class here is silent failure, a failed re-apply now aborts the refresh before the upgrade instead of only echoing. The command has no set -e, so the step is checked explicitly; test/shell.d/refresh-pacman-test.sh gained a failing-hardware-script case that fails on the unchecked version and passes now. The repair-migration question I am leaving to the maintainer as agreed: a migration that re-adds the unsigned stanza would collide with #8709 and #9461, so I did not touch it. The pre-refresh-pacman double-append note is worth a release line.

PR #9880, opened 2026-09-02
Stop installing the obsolete SPI keyboard DKMS package

### LINKED ISSUE #9831: fix-spi-keyboard.sh installs macbook12-spi-driver-dkms, which cannot build on kernels shipping applespi in-tree
`install/hardware/apple/fix-spi-keyboard.sh` installs `macbook12-spi-driver-dkms`
on every MacBook it matches. On current kernels that package is obsolete and
**cannot build** — `applespi` has been mainlined, and the out-of-tree copy still
includes a header that no longer exists.

The keyboard works anyway, because the mkinitcpio drop-in the same script writes
pulls in the in-tree module. So the only effect of the package is a failing DKMS
build on every kernel update.

### Evidence

`applespi` ships in the `linux` package:

```
$ pacman -Qo "$(modinfo -n applespi)"
/usr/lib/modules/7.1.9-arch1-2/kernel/drivers/input/keyboard/applespi.ko.zst is owned by linux 7.1.9.arch1-2

$ tar -tf linux-7.1.9.tar.xz | grep applespi
linux-7.1.9/drivers/input/keyboard/applespi.c
linux-7.1.9/drivers/input/keyboard/applespi.h
linux-7.1.9/drivers/input/keyboard/applespi_trace.h
```

It is also the only `applespi` on disk — the DKMS build never produces one:

```
$ find /lib/modules/7.1.9-arch1-2 -name 'applespi*'
/lib/modules/7.1.9-arch1-2/kernel/drivers/input/keyboard/applespi.ko.zst
```

The build failure, from `/var/lib/dkms/macbook12-spi-driver/0+git.315/build/make.log`:

```
   64 | #include <asm/unaligned.h>
      |          ^~~~~~~~~~~~~~~~~
compilation terminated.
make[4]: *** [scripts/Makefile.build:289: applespi.o] Error 1
make[4]: *** [scripts/Makefile.build:289: apple-ibridge.o] Error 1
```

`asm/unaligned.h` was removed in 6.12; it is `linux/unaligned.h` now, and the
kernel tree confirms there is no `arch/x86/include/asm/unaligned.h` in 7.1.9.

As seen during a pacman transaction:

```
(3/3) Install DKMS modules
==> dkms install --no-depmod macbook12-spi-driver/0+git.315 -k 7.1.9-arch1-2
Error! Bad return status for module build on kernel: 7.1.9-arch1-2 (x86_64)
==> WARNING: `dkms install --no-depmod macbook12-spi-driver/0+git.315 -k 7.1.9-arch1-2' exited 10
```

### Why this is currently invisible

On a stock install `linux-headers` isn't present, so Arch's DKMS hook globs
`usr/lib/modules/*/build`, matches nothing, builds nothing and exits 0. The
package sits at `added` and never errors:

```
$ dkms status
macbook12-spi-driver/0+git.315: added
```

The failure only surfaces the moment `linux-headers` gets installed for any
reason.

**This interacts with #8285.** That issue proposes installing `linux-headers`
explicitly (for the CS8409 audio driver) and describes the same zero-kernel
glob behaviour. If it lands, every matched MacBook starts surfacing this failed
build on each kernel update. Fixing this first would avoid shipping that
regression.

### Steps to reproduce

1. Install Omarchy on a machine matching the regex (MacBookPro14,1 here).
2. `sudo pacman -S linux-headers`
3. Observe the DKMS build fail in the pacman output, and on every kernel update after.

### Expected

Either don't install `macbook12-spi-driver-dkms` on kernels that ship `applespi`
in-tree, or drop it entirely and rely on the mkinitcpio drop-in.

### Important: the drop-in must stay

The other half of the script is load-bearing and should not be removed along
with the package:

```
/etc/mkinitcpio.conf.d/macbook_spi_modules.conf
MODULES=(applespi intel_lpss_pci spi_pxa2xx_platform)
```

On a LUKS root the keyboard has to be in the initramfs to type the passphrase
(cf. #1313). Removing only the package is safe — pacman owns just
`/usr/src/macbook12-spi-driver-0+git.315/`, not the drop-in — and the keyboard
and trackpad keep working:

```
$ sudo pacman -R macbook12-spi-driver-dkms
$ grep '^N: Name' /proc/bus/input/devices | grep -i apple
N: Name="Apple SPI Keyboard"
N: Name="Apple SPI Touchpad"
```

### System

| | |
|---|---|
| Omarchy | 4.0.2-1 (stable) |
| Model | MacBookPro14,1 |
| CPU | Intel Core i5-7360U |
| GPU | Intel Iris Plus Graphics 640 [8086:5926] |
| Kernel | 7.1.9-arch1-2 |
| dkms | 3.4.3-2 |
| package | macbook12-spi-driver-dkms 0+git.315-1 |

hudsonwa:
Fixes #9831

## What broke

`install/hardware/apple/fix-spi-keyboard.sh` installs `macbook12-spi-driver-dkms` on every MacBook it matches. That package is obsolete: `applespi` is mainlined (it ships in the `linux` package), and the out-of-tree copy includes `asm/unaligned.h`, which was removed in kernel 6.12. The DKMS build therefore fails on every kernel update once `linux-headers` is present (reproduced in the issue with the full build log). The failure is currently invisible on stock installs only because Arch's DKMS hook finds no `usr/lib/modules/*/build` without headers; issue #8285 proposes installing `linux-headers`, which would surface this failed build on every matched MacBook.

The keyboard itself is unaffected: the script also writes `/etc/mkinitcpio.conf.d/macbook_spi_modules.conf`, which loads the in-tree `applespi` module into the initramfs. That half is load-bearing (LUKS passphrase entry), and the issue verifies removing only the package leaves the keyboard and touchpad working.

## Fix

Drop the `omarchy-pkg-add macbook12-spi-driver-dkms` line and keep everything else, with a comment recording why the package no longer belongs. The initramfs drop-in is unchanged, so initramfs behavior is identical.

## Why install/hardware

This is an install-time hardware leaf run via `run_logged` from `install/hardware/all.sh`. The package install belongs there (or nowhere), and the drop-in must stay there, so the fix is confined to the one file.

## Tests

New `test/shell.d/spi-keyboard-test.sh` runs a sandboxed copy of the leaf (DMI path and mkinitcpio dir rewritten, stubbed `omarchy-pkg-add`/`sudo`):

- Matched MacBook (MacBookPro14,1): the DKMS package is NOT requested, and the initramfs drop-in is still written with the SPI modules.
- Unmatched machine (MacBookPro15,2): nothing happens at all.

All pass here; the test fails against the pre-fix leaf (RED to GREEN). Honest host note: this machine cannot run a real pre-T2 MacBook, so the model regex is covered by the sandbox rather than hardware; the match set itself is unchanged.

## Not included

A migration that removes the package from already-installed machines was left out deliberately (machine-wide package removal is the maintainers' call; the script fix prevents new installs). Happy to add one if wanted. This also removes the regression #8285 would have surfaced.

### DIFF
--- a/install/hardware/apple/fix-spi-keyboard.sh
+++ b/install/hardware/apple/fix-spi-keyboard.sh
@@ -3,7 +3,11 @@ product_name="$(cat /sys/class/dmi/id/product_name 2>/dev/null)"
 if [[ $product_name =~ MacBook[89],1|MacBook1[02],1|MacBookPro13,[123]|MacBookPro14,[123] ]]; then
   echo "Detected MacBook with SPI keyboard"
 
-  omarchy-pkg-add macbook12-spi-driver-dkms
+  # macbook12-spi-driver-dkms is obsolete: applespi is mainlined (it ships in
+  # the linux package, and its header that the out-of-tree copy needs was
+  # removed in 6.12), so the DKMS build fails on every kernel update once
+  # linux-headers is present. Only the initramfs drop-in below is needed to
+  # get the in-tree module into the initramfs (load-bearing for LUKS roots).
   sudo mkdir -p /etc/mkinitcpio.conf.d
   if [[ $product_name == "MacBook8,1" ]]; then
     echo "MODULES=(applespi spi_pxa2xx_platform spi_pxa2xx_pci)" | \
--- a/install/omarchy-other.packages
+++ b/install/omarchy-other.packages
@@ -25,7 +25,6 @@ linux-firmware
 linux-headers
 linux-ptl
 linux-ptl-headers
-macbook12-spi-driver-dkms
 nvidia-580xx-dkms
 nvidia-dkms
 nvidia-open-dkms
--- a/migrations/1788476400.sh
+++ b/migrations/1788476400.sh
@@ -1,0 +1,12 @@
+echo "Remove the obsolete MacBook SPI keyboard DKMS package"
+
+# applespi is mainlined (it ships with the linux package), so the out-of-tree
+# driver cannot build on current kernels: its include was removed in kernel
+# 6.12 and the DKMS build fails on every kernel update once linux-headers is
+# present. The initramfs drop-in written by
+# install/hardware/apple/fix-spi-keyboard.sh loads the in-tree module instead,
+# so the package is pure failure surface. Idempotent: nothing happens when the
+# package is not installed.
+if omarchy-pkg-present macbook12-spi-driver-dkms; then
+  omarchy-pkg-drop macbook12-spi-driver-dkms
+fi
--- a/test/shell.d/spi-keyboard-test.sh
+++ b/test/shell.d/spi-keyboard-test.sh
@@ -58,4 +58,9 @@ run_leaf "MacBookPro15,2"
   fail "an unmatched machine does not write the drop-in"
 pass "unmatched hardware is untouched"
 
+# Nothing installs the package now, so the ISO has no reason to cache it.
+! grep -qx 'macbook12-spi-driver-dkms' "$ROOT/install/omarchy-other.packages" ||
+  fail "the ISO no longer caches macbook12-spi-driver-dkms"
+pass "the obsolete DKMS package is gone from the ISO package cache list"
+
 pass "SPI keyboard detection writes only the needed initramfs drop-in"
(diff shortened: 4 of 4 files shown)


hudsonwa:
Thanks for the review and the ISO fix; caching a package nothing installs was the right catch. I completed the migration half: migrations/1788476400.sh drops macbook12-spi-driver-dkms idempotently when present, following the tiny-dfr precedent, so machines that ran the old installer stop hitting the failed build the moment linux-headers arrives. spi-keyboard-test.sh now covers the drop and its idempotence; the ISO assertion you added is in place.

PR #9881, opened 2026-09-02
Strip WhatsApp Slim extension flags when the web app is removed

### LINKED ISSUE #9856: Remove WhatsApp Slim extension from Brave Origin when uninstalling WhatsApp webapp
### System details

Omarchy 4

### What's wrong?

I'm not a WhatsApp user and do not want the app on my Omarchy. After removing the web app I discovered that Brave Origin has an out of the box WhatsApp Slim extension installed. Even after manually disabling and removing it from Brave, it came back after the next restart.
I think the extension should be removed when the web app is uninstalled.

hudsonwa:
Fixes #9856

## What broke

The WhatsApp Slim extension is loaded into every Chromium-family browser through `--load-extension=.../whatsapp-slim` entries in `~/.config/<browser>-flags.conf` (migrations 1785543725 and 1785591762 add them for eight profiles: chromium, chrome, google-chrome, brave, brave-beta, brave-nightly, brave-origin, microsoft-edge-stable). Removing the WhatsApp web app removes the launcher and its icons, but not those flags, so the extension keeps loading on every browser restart even after the user disables it manually.

## Fix

`bin/omarchy-webapp-remove` now also strips the WhatsApp Slim entry (name matched case-insensitively) from the same flags files the migrations touch, as the inverse of the migration's append:

- a line that only loads the slim extension is deleted;
- `--load-extension=/path/to/others,$ext` becomes `--load-extension=/path/to/others` (sibling extensions such as copy-url and yt-dlp are kept);
- files without a whatsapp-slim entry are left alone.

The sed expressions avoid `/` as a delimiter because the extension path contains slashes, and the collapse-then-strip-then-delete ordering keeps every case and a re-run idempotent.

## Why bin/

This is the removal path for a web app; the flags are user-level config under `$HOME/.config`, and the command already operates on user-level launchers/icons.

## Tests

New `test/shell.d/webapp-remove-test.sh` runs the real command against a sandbox HOME:

- removing WhatsApp deletes the launcher, strips the slim entry from `chromium-flags.conf` while keeping `copy-url`, and empties a brave-origin flags file that had only the slim extension;
- removing an unrelated app (Slack) leaves every flags file untouched.

Honest host note: this machine is macOS with bash 3.2 and BSD sed, and the command is a bash 4+/GNU sed script (mapfile in the no-arg path, `sed -i --follow-symlinks`), so the full test file is written for the repo's Arch/CI environment. Locally I validated the exact sed expressions with BSD sed against the three real file shapes (comma-suffixed list, sole extension line, unrelated file) and a repeat run: all four pass, and the unmodified command leaves the flags in place, which is the bug reproduced.

### DIFF
--- a/bin/omarchy-webapp-remove
+++ b/bin/omarchy-webapp-remove
@@ -62,7 +62,11 @@ rm -f "$ICON_DIR/$icon_name.png" "$ICON_DIR/$APP_NAME.png" "$OLD_ICON_DIR/$APP_N
 # every browser restart (issue #9856).
 app_lc=$(printf '%s' "$APP_NAME" | tr '[:upper:]' '[:lower:]')
 if [[ $app_lc == "whatsapp" ]]; then
-  whatsapp_slim_ext="$OMARCHY_PATH/default/chromium/extensions/whatsapp-slim"
+  # Matched by path suffix, not by the current $OMARCHY_PATH: the migration
+  # wrote whichever value was live when it ran, and `omarchy dev link` changes
+  # it. Only one Omarchy owns a flags file, so any entry ending in this path is
+  # the one the migration added.
+  whatsapp_slim_ext='[^,]*/default/chromium/extensions/whatsapp-slim'
   for conf in chromium chrome google-chrome brave brave-beta brave-nightly brave-origin microsoft-edge-stable; do
     flag_file="$HOME/.config/$conf-flags.conf"
     [[ -f $flag_file ]] || continue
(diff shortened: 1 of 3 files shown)


hudsonwa:
Thanks for the review; both real defects. The first-position gap was one migration away from biting and the prefix corruption was worse than I want in a flags rewrite. Your commit is in. I also closed the remove-all gap: omarchy-webapp-remove-all now routes each app through omarchy-webapp-remove (notifications suppressed) so Remove -> Preinstalls removes the paired flags too, with webapp-remove-test.sh covering that path. I am leaving config/chromium-flags.conf alone: whether the extension is default-on is a product call, and the issue is about removal, so I did not want to widen this PR.

Issue #9823, opened 2026-09-02
Chromium main Meet tab loses borders after session restore - over-broad PiP rule pip.lua:15


0xleelu:
**Omarchy version:** 4.0.2-1 (Hyprland 0.56.2, eDP-1 1920x1200@1.25)
**Repro 100% - only when in a Meet meeting, not on landing page:** Open `chromium` with an *active* `meet.google.com` meeting (title `Meet - ... - Chromium`), close Chromium with tabs open (session `restore_on_startup=1`, `Preferences` `browser.window_placement.maximized=True`, `exit_type=Crashed`), reopen with restored tabs -> main Meet window loses Hyprland borders and is offset. Does **NOT** happen on the base `https://meet.google.com` landing page (title `Google Meet` - no `Meet - ` prefix, doesn't match `^Meet - .+`), only once you're in a meeting where title becomes `Meet - <meeting name> - Chromium`.

**What happened:**
- Main Meet *meeting* tab after restore shows at `[10,36] size [1516,914] floating:false` vs expected tiled with borders at `[12,38] [1512,910]` (like `foot` and `chrome://newtab` window). `hyprctl clients -j` shows `tags: chromium-based-browser*` but `border_size` appears 0, `hyprctl decorations` empty. Visual: no 2px border from `default/hypr/looknfeel.lua:10`, gaps look wrong, window behaves half-tiled/half-pinned.
- Landing page `meet.google.com` without a meeting stays correctly bordered `12,38 1512x910` after restore - confirms over-broad `^Meet - .+` match.

**What expected:**
Main Meet *meeting* tab should stay tiled with `border_size 2` `gaps_out 10`, only the actual PiP overlay should be floating borderless `600x338` pinned.

**Root cause (read-only diagnosis, no local changes):**
- `/usr/share/omarchy/default/hypr/apps/pip.lua:15`:
```lua
o.window({ tag = "chromium-based-browser", title = "^Meet - .+" }, {
  float=true, pin=true, size={600,338}, border_size=0, move={...}
})
```
[...]

PR #9882, opened 2026-09-02
Tag Chromium --app web apps as chromium-based browsers

### LINKED ISSUE #9784: Web apps never get the chromium-based-browser tag: browser.lua patterns assume substring matching, Hyprland uses RE2::FullMatch
## Summary

Hyprland matches window rule `class` with `RE2::FullMatch`, but the chromium tag rule in `default/hypr/apps/browser.lua` is written as if matching were a substring search. As a result **no web app window ever receives `+chromium-based-browser`**, on any Chromium-family browser, so web apps miss the tile and opacity rules that follow the tag.

This is adjacent to but not covered by #8968, which fixes the *browser* class (`brave-origin`) on the same line. Web app classes stay unmatched even with that PR applied.

## The rule

```lua
o.window("((google-)?[cC]hrom(e|ium)|[bB]rave-browser|[mM]icrosoft-edge|Vivaldi-stable|helium)", { tag = "+chromium-based-browser" })
```

Every alternative is a complete browser class, so the rule works for real browser windows. Web apps launched through `omarchy-launch-webapp` get a Chromium `--app` class of the form `<browser>-<host>__<path>-Default`, which no alternative can full-match.

Matching semantics, for reference (`src/desktop/rule/matchEngine/RegexMatchEngine.cpp` in Hyprland):

```cpp
bool CRegexMatchEngine::match(const std::string& other) {
    return re2::RE2::FullMatch(other, *m_regex) != m_negative;
}
```

## Steps to reproduce

```bash
omarchy-webapp-install "Example" "https://example.com/" "https://example.com/favicon.ico"
gio launch ~/.local/share/applications/Example.desktop
hyprctl clients -j | jq -r '.[] | select(.class | test("example.com")) | "\(.class)  tags=\(.tags)"'
```

Observed, with Chromium as the default browser:

```
chrome-example.com__-Default  tags=["default-opacity*"]
```

And with Brave Origin as the default browser (an Outlook web app):

```
brave-outlook.office.com__mail_-Default  tags=["default-opacity*"]
```

**Expected:** both carry `chromium-based-browser`.
**Actual:** neither does. They fall through to the generic `default-opacity` rule.

Note the first case uses plain Chromium, whose class does contain the literal substring `chrome`. It still misses the tag, which isolates the cause to full-match semantics rather than to any particular browser's naming.

## Consequences

1. Web apps miss `tile = true` and the browser opacity pair `1.0 0.985`, and instead inherit `default-opacity` at `0.985 0.96`. Web app windows are visibly more translucent than browser windows when unfocused, which is most noticeable on text-dense apps such as web mail.
2. These lines further down the same file are dead code, because they remove a tag that was never applied:

   ```lua
   o.window("(^.+-youtube\\.com__.*$|^.+-app\\.zoom\\.us__wc_home.*$)", { tag = "-chromium-based-browser" })
   ```

   Their existence, and their deliberately browser-agnostic `^.+-` prefix, suggests web apps were intended to carry the tag. YouTube and Zoom still end up correct only because the following line drops `default-opacity` from them directly.

## Suggested direction

Match the web app class form alongside the bare browser classes, in the same browser-agnostic spirit as the existing YouTube and Zoom rules, for example by allowing an optional `-<appid>-Default` suffix on the browser alternation. I have not opened a PR since #8968 already touches this line and adds `test/shell.d/browser-tags-test.sh`; folding a web app case into that test seems tidier than a competing change.

## System

- Omarchy 4.0.2-1
- Hyprland 0.56.2
- Chromium 151.0.7922.173
- Brave Origin 152.1.94.117 (`brave-origin-bin` 1:1.94.117-1)

fresh3nough:
## Summary

Fixes #9784.

Hyprland matches window-rule `class` with `RE2::FullMatch`. The chromium tag rule in `default/hypr/apps/browser.lua` listed bare browser classes only, so Chromium `--app` web apps never received `+chromium-based-browser`.

Observed classes from the issue:

```
chrome-example.com__-Default
brave-outlook.office.com__mail_-Default
```

Both missed the tag and fell through to `default-opacity` (more translucent when unfocused) instead of the browser tile + opacity pair.

## Change

```lua
o.window("((google-)?[cC]hrom(e|ium)|[bB]rave(-browser|-origin)?|[mM]icrosoft-edge|Vivaldi-stable|helium)(-.+-Default)?", { tag = "+chromium-based-browser" })
```

- Optional `(-.+-Default)` covers the Chromium `--app` class form used by `omarchy-launch-webapp`.
- `brave(-browser|-origin)?` covers bare Brave / Brave Origin windows and their webapp classes (adjacent to #8968 on the same line).
- Existing YouTube/Zoom strip rules already use a browser-agnostic `^.+-youtube.com__.*` form and start working once the tag is applied.

## Test plan

- [x] Added `test/shell.d/webapp-browser-tags-test.sh` using Python `re.fullmatch` (bash `=~` is unanchored and would hide this bug)
- [x] Bare Chromium-family classes still match
- [x] Webapp classes from the issue match (`chrome-example.com__-Default`, `brave-outlook.office.com__mail_-Default`, plus origin/edge/helium variants)
- [x] Firefox and incomplete suffixes stay untagged
- [x] `bash test/shell.d/webapp-browser-tags-test.sh` — all pass
- [x] Confirmed pre-fix FullMatch fails the issue classes; post-fix succeeds

```
ok - chromium rule full-matches bare Chromium-family browser classes
ok - chromium rule full-matches Chromium --app webapp classes
ok - chromium rule leaves non-Chromium and incomplete classes untagged
ok - chromium rule includes the Chromium --app webapp class suffix
```

### DIFF
--- a/default/hypr/apps/browser.lua
+++ b/default/hypr/apps/browser.lua
@@ -1,12 +1,22 @@
 -- Browser tags and styling.
-o.window("((google-)?[cC]hrom(e|ium)|[bB]rave-browser|[mM]icrosoft-edge|Vivaldi-stable|helium)", { tag = "+chromium-based-browser" })
+-- Hyprland matches class with RE2 FullMatch, so bare browser ids alone miss
+-- Chromium --app webapp classes. The --app class uses a product id prefix, not
+-- the browser window class: chrome- / brave- / msedge- / vivaldi- / opera- /
+-- helium-, then <host>__<path>-<Profile> (see omarchy-launch-webapp).
+-- Suffix is (-.+__.*-.+)? so non-Default profiles still match while extension
+-- popouts like chrome-<extension-id>-Default (no host/__) stay untagged.
+-- [oO]pera also tags Opera's main window class for the first time.
+o.window("((google-)?[cC]hrom(e|ium)|[bB]rave(-browser|-origin)?|[mM]icrosoft-edge|msedge|[vV]ivaldi(-stable)?|[oO]pera|helium)(-.+__.*-.+)?", { tag = "+chromium-based-browser" })
 o.window("([fF]irefox|zen|librewolf)", { tag = "+firefox-based-browser" })
-o.window({ tag = "chromium-based-browser" }, { tag = "-default-opacity", tile = true, opacity = "1.0 0.985" })
-o.window({ tag = "firefox-based-browser" }, { tag = "-default-opacity", opacity = "1.0 0.985" })
 
--- Video apps: remove chromium browser tag so they don't get opacity applied.
+-- Video apps: remove the chromium browser tag so they don't get opacity applied.
+-- This has to precede the rules below, because removing a tag does not undo a
+-- rule that already matched on it.
 o.window("(^.+-youtube\\.com__.*$|^.+-app\\.zoom\\.us__wc_home.*$)", { tag = "-chromium-based-browser" })
 o.window("(^.+-youtube\\.com__.*$|^.+-app\\.zoom\\.us__wc_home.*$)", { tag = "-default-opacity" })
+
+o.window({ tag = "chromium-based-browser" }, { tag = "-default-opacity", tile = true, opacity = "1.0 0.985" })
+o.window({ tag = "firefox-based-browser" }, { tag = "-default-opacity", opacity = "1.0 0.985" })
 
 -- Hide screen sharing notification windows.
 o.window({ title = ".*is sharing.*" }, { workspace = "special silent" })
--- a/test/shell.d/webapp-browser-tags-test.sh
+++ b/test/shell.d/webapp-browser-tags-test.sh
@@ -1,0 +1,114 @@
+#!/bin/bash
+
+# Chromium --app web apps must receive +chromium-based-browser under Hyprland's
+# RE2 FullMatch class matching. Bare browser ids alone never full-match the
+# product-prefixed <id>-<host>__<path>-<Profile> form from omarchy-launch-webapp
+# (chrome-/brave-/msedge-/vivaldi-/opera-/helium-, not the bare window class).
+#
+# Regression guard for omacom/omarchy#9784. Uses Python re.fullmatch so the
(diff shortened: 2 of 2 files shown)


fresh3nough:
Fixed (bot “smaller, unfixed”)

Real --app class prefixes in browser.lua: msedge, vivaldi/Vivaldi-stable, opera (kept bare browser classes; product prefixes for webapps).

Tests: realistic webapp ids only; dropped brittle (-.+-Default)? text assert; added require_command python3.

Not touched (as instructed)

#9823 Meet PiP — left alone (maintainer call).
Non-Default profiles — left unmatched.

UI inspection
Not required. Static regex + re.fullmatch shell test only. webapp-browser-tags-test.sh → 3 ok.

fresh3nough:
Addressed https://github.com/omacom/omarchy/pull/9882#issuecomment-5551859520.

Rebased onto current `quattro` and pushed `b5816578`.

What changed

1. Webapp suffix is now `(-.+__.*-.+)?` instead of `(-.+-Default)?`.
   - Non-Default profiles match (`chrome-example.test__-Profile_1`).
   - Extension popouts without a host/`__` stay untagged (`chrome-nngceckbapebfimnlniiiahkandclblb-Default` for Bitwarden).
   - Chromium `--app` ids always include `<host>__<path>` before the profile; that is the separator.

2. Tests
   - Profile_1 / non-Default webapp classes added as positives.
   - Bitwarden-style extension id added as a negative.
   - Rule-order guard: YouTube/Zoom `-chromium-based-browser` removal must precede the tile/opacity consumer (Hyprland does not undo a rule that already matched).

3. Noted in code comments: `[oO]pera` also tags Opera's main window class (beyond the pure webapp fix; intentional).

Left alone (maintainer calls from the review)

- #9823 Meet PiP widening once web apps carry the tag.
- Collision with #8968 on the same line.

`bash test/shell.d/webapp-browser-tags-test.sh` — all ok.

Commit is DCO-signed (`Signed-off-by: fresh3nough <anonwurcod@proton.me>`) and SSH-signed (verified).

PR #9885, opened 2026-09-02
Notify when agent limits reset

szalikdev:
## Summary

- schedule each future agent rate-limit reset independently of usage refreshes
- show an Omarchy desktop notification when the window becomes available again
- expose an enabled-by-default `notifyOnLimitReset` bar-widget setting

## Testing

- `./test/shell.d/agents-panel-test.sh`
- `./test/shell` (agent-related tests passed; full run later stops at the existing `omarchy-pkgs checkout found for PKGBUILD coverage` environment prerequisite)

### DIFF
--- a/shell/plugins/agents/Main.qml
+++ b/shell/plugins/agents/Main.qml
@@ -76,8 +76,70 @@ Item {
 
   function recordsChanged() {
     dataRevision++
+    scheduleLimitResetNotifications()
     scheduleLimitsRetry()
     scheduleSync()
+  }
+
+  // A reset deadline is stable for the lifetime of a rate-limit window. Keep
+  // the deadlines we first see in memory and announce them when they pass;
+  // this stays accurate even when the normal usage refresh is deliberately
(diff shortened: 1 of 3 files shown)


sprajs:
Queued resets survive a provider being disabled. `scheduleLimitResetNotifications()` copies the previous queue and skips disabled providers, while `announcePassedLimitResets()` never checks `providerEnabled()`. I reproduced this with the added functions and a fake clock: queue a future Codex reset, disable Codex, advance past the deadline, and a “Codex limit reset” notification still fires.

Could pending entries retain the provider ID and be removed or ignored when that provider is disabled? A behavioral test covering schedule → disable → deadline would catch this; the new tests currently check that the functions exist without exercising that transition.


I implemented a fix with regression tests in https://github.com/omacom/omarchy/pull/12425, in the main Omarchy repository, preserving this PR’s original commits. The focused tests pass; the PR records the full-suite failures and remaining runtime validation.

PR #9894, opened 2026-09-02
Fix Tailscale plugin failing to reconnect when accept-routes is enabled

llstrk:
## Summary

Reconnecting through the Tailscale plugin currently runs `tailscale up`, which requires restating every persisted option that differs from its CLI defaults. This fails when settings such as `accept-routes` are enabled.

Replace it with a targeted LocalAPI preference update that changes only `WantRunning`, preserving existing configuration. Interactive login uses the LocalAPI login endpoint and opens only explicit `AuthURL` or IPN `BrowseToURL` values.

Subscribe to Tailscale's IPN event stream for prompt connection-state updates, with the existing periodic refresh retained as a fallback. If a stopped daemon reveals an expired key only after a resume request, the same user action continues into interactive login without requiring another toggle.

## Lifecycle and timeout handling

- Bound finite CLI commands (resume/login request, disconnect, account switch, and exit-node selection) to 20 seconds, with a two-second forced-termination grace period. The event watcher and interactive authorization prompt are deliberately excluded.
- Bound the asynchronous login-link wait to 25 seconds. Suppress repeated login attempts, ignore late URLs after timeout or cancellation, and refresh status at the deadline so a manual retry can use a newly available URL even if the watcher is reconnecting.
- Clear optimistic and pending state after command failure, preserve useful timeout feedback, and clear old errors when starting a retry. No automatic command retry or daemon restart is introduced. Terminating a CLI does not imply cancellation of work already accepted by `tailscaled`.
- Retain the 20-second optimistic connection deadline and re-check status when it expires.
- Reveal machine sections only after status confirms the connection. Reserve stable message space for login-required states, while keeping ordinary disconnects compact.
- Close the panel when handing authentication to the browser. Do **not** automatically reopen it when login finishes: the user may already be typing in another application, and reopening would steal keyboard focus.

## Watcher and error hardening

- Use IPN mask `18` (`NotifyInitialState | NotifyNoPrivateKeys`) so the stream does not carry node private keys. Discard the login process's captured preferences body after completion.
- Back off watcher restarts from two seconds to one minute while the daemon is unreachable, resetting after a notification object parses.
- Strip LocalAPI CLI request commentary from login errors and provide a fallback when the command produces no useful error text.

## Testing

Latest checks:

- `bash test/shell.d/tailscale-test.sh` passed, including tests executing the actual QML function/handler bodies for timeout cleanup, retry feedback, expired-key continuation, cancellation, and late authentication URLs.
- `bash test/shell.d/qml-text-format-test.sh` passed.
- `git diff --check` passed.
- Tested the unmodified candidate code in fresh nested shells with isolated CLI fixtures: hung commands and commands ignoring SIGTERM terminated at approximately 20/22 seconds, all four actions accepted subsequent retries, and pending/busy state cleared.
- Verified expiry discovered before and after the resume command exits, watcher-unavailable deadline refresh, cached-URL retry without an unnecessary login POST, and suppression of late URLs after disconnect.
- Inspected screenshots and recorded transitions for stable login layout, compact disconnects, timeout feedback, and retry behavior.
- Verified typing in another application continued uninterrupted across simulated authentication completion; the bar updated while the panel remained closed.

Earlier manual verification covered reconnecting with `RouteAll=true`, preservation of persisted preferences, prompt external state updates, watcher restart, and explicit authentication URL handling on Tailscale 1.102.3. The earlier CLI suite passed; shell/aggregate runs encountered three unrelated failures due to an unavailable `omarchy-pkgs` checkout. Those full suites were not rerun for this update.

Fable reviewed the local follow-up changes through Claude Code; the actionable review items were addressed and regression-tested. Local plugin deployment/cache workarounds and diagnostic tooling are not included in the PR.


llstrk:
## Automated AI review

> **Community review:** Independent automated community review, unaffiliated with the Omarchy team, intended to help prepare PRs for their review.

**Verified:** Resuming through `PATCH /localapi/v0/prefs` with `{"WantRunning":true,"WantRunningSet":true}` changes only `WantRunning`, so accept-routes and other saved preferences no longer block reconnecting. Tailscale's source (v1.102.3 and main) confirms the endpoint, the masked-prefs semantics, a non-zero `tailscale debug localapi` exit on non-2xx responses, and that turning `WantRunning` on starts the control login. The `timeout --foreground --kill-after=2s 20s` wrapper returns 124 for a child that exits on TERM and 137 after about 22 s for a child that ignores it, and ordinary exit codes pass through. The PR's `tailscale-test.sh` passes on the head and fails against the merge-base plugin sources.

One issue was found in the interactive-login path.

### The IPN watcher does not receive the login URL requested by `login-interactive`

In tailscaled, `StartLoginInteractiveAs` records the requesting connection's actor, and `setAuthURLLocked` later sends `BrowseToURL` only to that recipient. On Linux the recipient's `ClientID` is derived from the connecting process's PID ([`ipn/ipnserver/actor.go`](https://github.com/tailscale/tailscale/blob/v1.102.3/ipn/ipnserver/actor.go), `newActor`; [`ipn/ipnlocal/local.go`](https://github.com/tailscale/tailscale/blob/v1.102.3/ipn/ipnlocal/local.go), `setAuthURLLocked` and `notificationTarget.match`). The plugin sends the POST from a short-lived `tailscale debug localapi` process and watches the bus from a separate long-lived one, so the watcher's session never matches:

```text
POST process (PID A) ── login-interactive ──> tailscaled: authActor = A
                                               control returns URL
                                               BrowseToURL -> clients matching A only
watcher process (PID B) ── watch-ipn-bus ────> no BrowseToURL (B != A)
                                               state stays NeedsLogin, no State change

URL reaches the plugin only through `tailscale status --json` AuthURL:
  +0.6 s after POST exits    delayedRefresh (often before control has replied)
  periodic refreshTimer      default 30 s, independent phase
  25 s deadline refresh      caches the URL, does not open it
```

The base ran `tailscale up`, which requests login and receives the URL in the same process, and parsed the link from its output. This does not affect the accept-routes resume fix, or the case where status already carries an `AuthURL` (then `loginPlan` opens it without a POST).

**Impact:** When the daemon has no login URL yet (for example after `tailscale logout`), the browser opens only if a periodic poll lands inside the 25 s window. Otherwise the panel reports "No login link received within 25 seconds" although a link exists, and a second click opens it. This is established from source; how often each outcome occurs on a live daemon was not measured. The tests do not catch it because they inject `BrowseToURL` directly into `handleStateWatchData`.

**Suggested change:** While `_loginInProgress` is true, poll `tailscale status --json` at a short interval (for example 1 s) until an `AuthURL` appears or the deadline passes. Alternatively, open a URL found by the deadline refresh instead of reporting that no link was received. The code comment and PR description stating that the watcher opens `BrowseToURL` for login could be qualified accordingly.

**Documentation note:** The mask comment says Prefs and NetMap notifications "otherwise carry the node's private key". In v1.102.3 and main, `NotifyNoPrivateKeys` is documented as a no-op because keys are always stripped. Mask `18` remains valid.

---

## Review information

**Test scope:** Source inspection of the PR head, the merge-base and Tailscale v1.102.3 and main. The PR's test script was run on the head and against the merge-base plugin sources, and the timeout wrapper was exercised with synthetic children, in an isolated sandbox. No live `tailscaled`, real login, QML runtime or UI was used, so the login-URL finding is source-level and its user-visible frequency is untested.

**AI process:** Opus 5.5 Medium coordination and synthesis, independent Opus 5.5 High and GPT 6 Sol Xhigh technical assessments, Opus 5.5 Medium editorial check.

**Opt out:** To stop receiving these reviews, reply to this comment saying so.

PR #9910, opened 2026-09-03
Persist apple-gmux panel brightness across reboots

shawnyeager:
## Why

Panel brightness on pre-T2 dual-GPU Macs (`gmux_backlight`) resets on reboot. Keyboard backlight on the same machine persists.

Stock systemd udev (`/usr/lib/udev/rules.d/99-systemd.rules`) attaches `systemd-backlight@` with `IMPORT{builtin}="path_id"` on the same line as `SYSTEMD_WANTS`. `apple-gmux` lives on PNP `APP000b`; `path_id` fails (`No such file or directory`), so the rest of the rule is skipped. No unit, no file under `/var/lib/systemd/backlight/`, nothing to restore.

The keyboard LED is on PCI/SPI, so `path_id` works and `systemd-backlight@leds:spi::kbd_backlight` already saves/loads.

`systemd-backlight` itself can persist without `ID_PATH` (it names the state file `backlight:gmux_backlight`). It is never asked.

Verified on MacBookPro14,3. Detector is the sysfs node, not a DMI list — T2 Macs expose PNP `APP000b` but never create `gmux_backlight`. Independent of Touch Bar, CS8409, and the 15" lid/idle work.

## What

- `omarchy-hw-apple-gmux` matches `/sys/class/backlight/gmux_backlight`.
- Packaged udev rule tags that device for systemd **without** importing `path_id`.
- Install hook + migration copy the rule, reload udev, start the unit, and snapshot the current level so the next boot has something to restore.

## Test plan

- [x] `test/shell.d/gmux-backlight-test.sh`
- [x] `./test/cli` metadata for the new hidden detector
- [x] MacBookPro14,3: after installing the rule, `systemd-backlight@backlight:gmux_backlight` is active and `/var/lib/systemd/backlight/backlight:gmux_backlight` matches live sysfs (471). Reboot still needed to confirm load-on-boot.

---

Diagnosed with Compound Engineering `/ce-debug`.

### DIFF
--- a/bin/omarchy-hw-apple-gmux
+++ b/bin/omarchy-hw-apple-gmux
@@ -1,0 +1,10 @@
+#!/bin/bash
+
+# omarchy:summary=Detect the old IO-port apple-gmux panel backlight
+# omarchy:hidden=true
+
+# Pre-T2 dual-GPU Macs register gmux_backlight on PNP APP000b. T2 Macs also
+# expose that HID, but apple-gmux does not bind there (no IORESOURCE_IO) and
+# never creates this class device — so the sysfs node is the discriminator,
+# not the PNP id. Observed on MacBookPro14,3.
+[[ -e ${OMARCHY_GMUX_BACKLIGHT:-/sys/class/backlight/gmux_backlight} ]]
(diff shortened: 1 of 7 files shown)


bjarneo:
Automated duplication check: this pull request looks similar to #8371, which covers the same GMUX backlight persist. I keep that one open and close this one to consolidate review.

PR #9921, opened 2026-09-03
Add SUPER+A universal select all keybinding

AlanRoman117:
## Summary
- SUPER+C (copy), SUPER+V (paste), and SUPER+X (cut) all have universal keybindings, but select all still required reaching for CTRL+A.
- Adds `SUPER + A` → "Universal select all" in `default/hypr/bindings/clipboard.lua`, forwarding CTRL+A to the focused window using the same `send_key_state` down/up pattern as the existing clipboard bindings.
- Skips terminal windows (reusing the existing `active_window_is_terminal` check) since CTRL+A there is readline's "move to start of line", not select-all — forwarding it would silently move the cursor instead.

## Test plan
- [x] `./test/all` passes
- [x] Verified locally: `SUPER + A` selects all text in GUI apps (browser, text editor), and is a no-op in terminal windows (leaves CTRL+A's normal readline behavior untouched)
- [x] `omarchy menu keybindings --print` shows `SUPER + A → Universal select all`

https://claude.ai/code/session_012FgNnd3XE7c2jL478nscw6

### DIFF
--- a/default/hypr/bindings/clipboard.lua
+++ b/default/hypr/bindings/clipboard.lua
@@ -46,13 +46,4 @@ o.bind("SUPER + C", "Universal copy", universal_clipboard_shortcut("CTRL", "C", 
 o.bind("SUPER + V", "Universal paste", universal_clipboard_shortcut("CTRL", "V", "SHIFT", "Insert"))
 o.bind("SUPER + X", "Universal cut", send_shortcut_once("CTRL", "X"))
 o.bind("SUPER + CTRL + V", "Clipboard manager", "omarchy-shell shell toggle omarchy.clipboard")
-
--- Skipped in terminals: CTRL+A there is readline's "move to start of line",
--- not select-all, so forwarding it would silently move the cursor instead.
-o.bind("SUPER + A", "Universal select all", function()
-  if active_window_is_terminal() then
-    return
-  end
-
-  send_shortcut_once("CTRL", "A")()
-end)
+o.bind("SUPER + A", "Universal select all", universal_clipboard_shortcut("CTRL", "A", "CTRL SHIFT", "A"))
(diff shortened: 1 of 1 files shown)


axelfontaine:
Strongly disagree about the no-op. To stay consistent, this shortcut should absolutely make it universal and send Ctrl+shift+a to terminals.

AlanRoman117:
Sounds good, let me update the PR. I also didn't want to get into how some folks may prefer their CLI, however it could indeed improve the experience. Please let me know if this is good and I'll proceed accordingly.

AlanRoman117:
Updated — SUPER+A now forwards CTRL+SHIFT+A in terminals instead of no-op'ing, using the existing `universal_clipboard_shortcut` helper (same one SUPER+C/V already use), so it's consistent with the rest of the file:

```lua
o.bind("SUPER + A", "Universal select all", universal_clipboard_shortcut("CTRL", "A", "CTRL SHIFT", "A"))
```

`./test/all` passes and I verified locally that the binding resolves cleanly (no `hyprctl configerrors`) and shows up as `SUPER + A → Universal select all` in `omarchy menu keybindings --print`.

https://claude.ai/code/session_01UByBunyAcPCN8Wyo2fwXpM

axelfontaine:
I may have spoken too soon. It appears only Ghostty supports a "select all" action 😬

AlanRoman117:
You're right, and it's actually worse than a missing feature — I installed all four terminals and tested what CTRL+SHIFT+A actually does in each by capturing the raw bytes their pty receives:

| Terminal | CTRL+SHIFT+A byte(s) | Result if forwarded |
|---|---|---|
| foot | `^A` (0x01, same as plain CTRL+A) | Reintroduces the original bug — readline still runs "move to start of line" |
| Alacritty | `^A` (0x01, same as plain CTRL+A) | Same — collapses to the same byte as CTRL+A |
| kitty | `^[[97;6u` (distinct CSI-u sequence) | Harmless no-op — readline doesn't recognize it — but still not select-all |
| Ghostty | consumed entirely by the terminal | Actually selects all — Ghostty ships `ctrl+shift+a=select_all` as a **default keybind** (`ghostty +list-keybinds --default`) and intercepts the chord before it ever reaches the shell |

Also confirmed via each terminal's own action/keybind listing that foot, kitty, and Alacritty have no bindable "select all" action at all (only mouse-driven selection primitives like `select-word`/`ExpandSelection`/etc.) — there's nothing for a forwarded keystroke to hook into on those three, regardless of which chord we pick.

So I've updated the binding to check specifically for Ghostty (`window.class == "com.mitchellh.ghostty"`, matching the regex already used to tag terminals in `default/hypr/apps/terminals.lua`) and only forward CTRL+SHIFT+A there. foot/kitty/Alacritty get a no-op, which is strictly better than the alternative of silently moving the cursor.

`./test/all` passes (82/82).

https://claude.ai/code/session_0117QTgSe8i9Sp8gry3KzW27

PR #9924, opened 2026-09-03
Add a Grok (xAI) collector to the agents bar widget

jpage77:
## Summary

- Adds `bin/omarchy-agent-usage-grok`, a collector for the `omarchy.agents` bar widget covering Grok/xAI, following the same display-ready JSON contract as `claude`/`codex`/`fireworks`.
- Primary data source is the Grok CLI's own local log (`~/.grok/logs/unified.jsonl`), which already carries the SuperGrok weekly allowance and prepaid balance on every `grok` run — no credentials needed. This is the *only* source for the weekly allowance: xAI's Management API only sees team pay-as-you-go billing, a different pool of money from a personal SuperGrok subscription.
- Falls back to that Management API before `grok` has ever been run locally. Getting a working key for it took three non-obvious steps (documented in `shell/plugins/agents/README.md`): a separate **management key** is required since the CLI's own OAuth token is inference-scoped only, that key is scoped to one team which need not be the CLI's default team, and its ACL must explicitly include billing read access. That fallback path has no per-model token breakdown available — the usage-analytics endpoint only accepts a cost/line-item query, not a token-count one — so it populates a balance only.
- Adds `grok` to the widget's `manifest.json` provider defaults and description, and documents both paths in the plugin's `README.md`.
- Adds `test/shell.d/agent-usage-grok-scanner-test.sh`, covering: a valid record with neither a log nor a key, building the record from a fixture log (percent-fraction scaling, cents-to-dollars balance, on-demand limit, reset time), picking the newest log line by timestamp regardless of file order, a staleness note on an old snapshot vs. none on a fresh one, the configured-key-wins-with-CLI-team_id credential precedence, the Management API's plain-decimal-string balance parse, the cost-series summation, and the CLI-token hint on a 403.

All findings above came from testing against a real xAI team and a real Grok CLI session, not from the docs alone (the docs don't mention the management-key/team-scoping/ACL requirements, and the usage-analytics endpoint's accepted field names aren't documented either — both were found by testing against the real API).

## Test plan

- [x] `./test/all` — passing except 3 pre-existing failures unrelated to this change (`config-test.sh`, `snapper-test.sh`, `unowned-system-paths-test.sh` all fail the same way on a clean checkout of this branch's base, needing an `OMARCHY_PKGS_PATH` checkout not present in this environment)
- [x] `bash test/shell.d/agent-usage-grok-scanner-test.sh` — all 8 assertions pass
- [x] Ran the collector standalone against a real Grok CLI session and a real xAI Management API team; installed it locally (`$OMARCHY_PATH/bin/`) and confirmed it renders correctly in the live bar panel

https://claude.ai/code/session_01Webm6BX1U8ZBfzuwQhEyyS

### DIFF
--- a/bin/omarchy-agent-usage-grok
+++ b/bin/omarchy-agent-usage-grok
@@ -1,0 +1,449 @@
+#!/usr/bin/python3
+# omarchy:summary=Print the Grok (xAI) usage record as JSON
+# omarchy:args=[--force] [--limits-only]
+# omarchy:hidden=true
+"""Collect xAI/Grok usage into one display-ready JSON record for the
+omarchy.agents bar widget.
+
+Primary source: the Grok CLI's own local log, ~/.grok/logs/unified.jsonl.
+Every `grok` run logs a "billing: fetched credits config" line carrying the
+same numbers the CLI itself displays — the SuperGrok weekly allowance and
+prepaid balance. This is deliberately not sourced from xAI's Management
+API: that API only sees team pay-as-you-go billing, a different pool of
+money from a personal SuperGrok subscription, with no visibility into the
+weekly allowance at all. Reading the CLI's log needs no credentials and
+reflects reality as of the last `grok` invocation.
+
+Fallback source, only used when no matching log line exists yet (e.g.
+right after install, before ever running `grok`): xAI's Management API
+(https://management-api.x.ai). Getting a working key for this takes three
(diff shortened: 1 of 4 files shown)


bjarneo:
Automated duplication check: this pull request looks similar to #9498, which covers the same Grok usage collector. I keep that one open and close this one to consolidate review.

PR #9926, opened 2026-09-03
Add Super+Z shortcut to reopen browser tabs

juan-LARRAYA:
## Summary

- add `Super+Z` to reopen the last closed tab in recognized Chromium- and Firefox-based browsers
- reuse application tags so the shortcut does nothing outside supported browser windows
- extract reusable Hyprland helpers for tag detection and safe synthetic shortcuts
- document the shortcut in the manual

## Testing

- `bash test/shell.d/keybindings-menu-test.sh`
- `./test/all` (221/226 files passed locally; the persistent environment-dependent failures also reproduce on a clean `origin/quattro` worktree, while the transient network QR test passes when rerun)


axelfontaine:
Disagree. They already share a common shortcut. No need to needlessly fill up the SUPER namespace for this.

PR #9929, opened 2026-09-03
Delegate bar transparency to the active plugin

GrantAbell:
## Summary

- route `omarchy bar transparent true|false|toggle` through the active bar over shell IPC
- add an explicit transparency hook while preserving the built-in bar's current boolean behavior
- fall back to editing `shell.json` when the shell is stopped or predates the IPC contract
- document the optional full-bar hooks and cover delegation plus fallback behavior

## Why

Full bar plugins can replace `omarchy.bar`, but the CLI currently writes `bar.transparent` directly. That prevents a custom bar from distinguishing a toggle request from an explicit `true` or `false` request. Delegating those operations lets each active bar define its own transparency modes without changing the existing config schema.

## Testing

- `env -u NO_COLOR ./test/all` — all 226 test files passed

### DIFF
--- a/bin/omarchy-bar
+++ b/bin/omarchy-bar
@@ -204,16 +204,26 @@ cmd_position() {
   echo "Bar position set to $position"
 }
 
+try_bar_transparency_ipc() {
+  local result
+  result=$(omarchy-shell shell "$@" 2>/dev/null) || return 1
+  [[ $result == "ok" ]]
+}
+
 cmd_transparent() {
   local transparent="${1:-}"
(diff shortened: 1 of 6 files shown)


GrantAbell:
### Use Case

My plugin [Three-State-Bar](https://github.com/GrantAbell/Omarchy-Three-State-Bar) would utilize this by allowing the `omarchy bar transparent toggle` command to cycle the bar through the three transparency states the plugin provides, rather than only the default 'off' and 'on' states which result in inconsistencies when used with the plugin.

PR #9933, opened 2026-09-03
Let users disable the bar transparency double-click gesture

itchyfeetleech:
Follows up on #9932.

Double-left-clicking empty bar space toggles transparency with no way to opt out, and it's easy to trigger accidentally. This adds a `transparencyToggle` bar option (default `true`) so `{"bar": {"transparencyToggle": false}}` leaves the bar as configured.

- shell/plugins/bar/Bar.qml: new property, read in applyBarConfig via `config.transparencyToggle !== false` (no deep-merge, so the default lives in code), honored by the CenterGestureArea double-click handler
- config/omarchy/shell.json, shell/README.md, docs/omarchy-shell.md: documented in the bar config shape
- shell/plugins/bar/README.md, manual/05-the-top-bar.md: gesture docs mention the opt-out
- test/shell.d/bar-test.sh: asserts the gesture is gated and defaults on

Tests: bar-test.sh fully green, including the two new assertions. ./test/all shows only the 5 failures that also fail on untouched quattro (agent-usage collectors x2, plus 3 needing an omarchy-pkgs checkout). Not visually verified in a running shell beyond config-level checks; the change is one boolean guard on an existing gesture.

### DIFF
--- a/config/omarchy/shell.json
+++ b/config/omarchy/shell.json
@@ -7,6 +7,7 @@
   "bar": {
     "position": "top",
     "transparent": false,
+    "transparencyToggle": true,
     "centerAnchor": "omarchy.clock",
     "layout": {
       "left": [
--- a/docs/omarchy-shell.md
+++ b/docs/omarchy-shell.md
@@ -140,6 +140,7 @@ string on a miss.
     "id": "omarchy.bar",
     "position": "top",
     "transparent": false,
+    "transparencyToggle": true,
     "centerAnchor": "omarchy.clock",
     "layout": {
       "left":   [ { "id": "omarchy.menu" } ],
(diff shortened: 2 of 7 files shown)


davidbrownell:
+1 on this — I hit the accidental toggle regularly on 4.0.3-1 and went looking for a config key before finding this PR.

For anyone else who lands here searching: there is no way to disable it today. `bar.transparent: false` in `shell.json` only sets the current state; the gesture rewrites that same key, so it doesn't stick. The only local workaround is `omarchy plugin clone omarchy.bar` and deleting the `onDoubleClicked` body, which freezes the whole ~2000-line bar engine at your installed version — a steep price for turning off one gesture.

The `transparencyToggle` shape here looks right to me: default `true` in code (correct, given there's no deep-merge on `shell.json`), and a single boolean guard rather than a broader gesture-config surface.

Worth noting this also overlaps with #8111, where tray clicks fall through to `CenterGestureArea` and double-clicking a tray icon toggles transparency. That's a separate routing bug, but an opt-out would defuse the most annoying symptom of it in the meantime.

PR #9942, opened 2026-09-03
Workspaces widget: middle-click pulls a tag onto the clicked monitor

ekollof:
Implements the workflow proposed in #9941.

On multi-monitor setups, left-clicking a workspace number keeps Hyprland's default semantics (focus warps to the monitor where the tag lives). This adds **middle-click** on a workspace number as the Qtile/Xmonad-style inverse: the tag is pulled onto the monitor whose bar was clicked, swapping it away from another monitor when occupied. Left-click behavior is unchanged.

### How it works

The widget resolves the monitor of the clicked bar instance via `QsWindow.window.screen.name`, then runs two dispatches:

```
hyprctl dispatch 'hl.dsp.focus({ monitor = "<name>" })'
hyprctl dispatch 'hl.dsp.focus({ workspace = "<id>", on_current_monitor = true })'
```

Two dispatches are required because `hl.dsp.focus` takes `monitor` or `workspace` but not both (`monitor` wins), and `on_current_monitor` resolves against the focused monitor — so the clicked monitor is focused first, then the tag is pulled onto what is now the current monitor. If the screen name cannot be resolved, it falls back to a single `on_current_monitor = true` dispatch against the currently focused monitor.

### Why middle-click instead of shift-click

Shift-click was the original idea, but Wayland does not deliver keyboard modifier state to layer surfaces without keyboard focus (`MouseEvent.modifiers` is always 0 on the bar, which is a `WlrKeyboardFocus: None` layer surface), and there is no global modifier API in Quickshell. Middle-click is part of the pointer event itself and passes through the bar's module drag overlay, which only grabs left clicks. Middle-click is currently unused on workspace buttons.

### Testing

- `./test/shell` passes (the five failing test files fail identically on pristine `quattro` in this environment: missing `omarchy-pkgs` checkout and a pre-existing rendered-geometry failure in `bar-icon-geometry`).
- Verified live on a four-monitor Hyprland session: middle-click pulls the tag onto the clicked monitor and swaps when it lives elsewhere; left-click behavior unchanged; middle-click on the already-active tag of that monitor is a no-op.


ekollof:
Rebased onto current quattro (2fbac0c8). All 3 commits replayed cleanly, no conflicts. workspaces-middle-click-test.sh passes.

PR #9944, opened 2026-09-03
agents: add OpenCode usage collector

ranjithrajv:
## Summary
Adds an `omarchy-agent-usage-opencode` collector so the `omarchy.agents` bar panel shows OpenCode usage alongside Claude, Codex, and Fireworks.

- Reads `~/.local/share/opencode/opencode.db` (any provider) — input/output/reasoning/cache tokens, grouped by day and model.
- Follows the existing collector contract (`schemaVersion: 1`, `--force`/`--limits-only`, file-locked cache in `~/.cache/omarchy/agent-usage`, `ready`/`hasLocalStats` based on actual sessions).
- No rate-limits endpoint — `limits: []`, same as other local-only views.
- Manifest defaults and README updated to include `opencode` (enabled by default, `enabled` still defaults to `true` for any discovered agent).

Automatically discovered by `omarchy-agent-usage-update` (ships as `bin/omarchy-agent-usage-opencode`) — no plugin code changes needed.

## Testing
- `python3 -m py_compile bin/omarchy-agent-usage-opencode` — pass
- `bin/omarchy-agent-usage-opencode --force | jq` — produces valid record with `id: opencode`, `recentDays`, `modelUsage`, `todayTokensByModel`
- `omarchy-agent-usage-update --force && ls ~/.local/state/omarchy/agents/usage/` — `opencode.json` appears and the bar panel picks up a new tab on refresh

## Notes
- Respects `XDG_DATA_HOME` for the DB path; read-only SQLite open to avoid blocking a running opencode.

### DIFF
--- a/bin/omarchy-agent-usage-opencode
+++ b/bin/omarchy-agent-usage-opencode
@@ -1,0 +1,272 @@
+#!/usr/bin/python3
+# omarchy:summary=Print the OpenCode usage record as JSON
+# omarchy:args=[--force] [--limits-only]
+# omarchy:hidden=true
+"""Collect OpenCode usage into one display-ready JSON record.
+
+Reads session data from OpenCode's SQLite database and computes per-day
+and per-model token usage. The agents panel only ever reads the JSON
+this prints.
+"""
+
+import argparse
+import fcntl
+import hashlib
+import json
+import os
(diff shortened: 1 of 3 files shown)


bjarneo:

PR #9958, opened 2026-09-03
Add minimax agent usage collector for opencode

axisg4:
Adds a usage collector for MiniMax (opencode) to the agents widget in the top bar. The collector reads session data from opencode's SQLite database and reports token usage, sessions, and daily activity.

The collector follows the same pattern as existing collectors (claude, codex, fireworks) and produces the required JSON schema for the omarchy.agents bar widget.

Fixes: Users with only MiniMax/opencode agents couldn't see the agents icon in the bar since no collector existed.


axisg4:
https://github.com/user-attachments/assets/25b10f9b-312c-48a1-8cd0-add549dc10ee

Issue #9914, opened 2026-09-03
quickshell SIGSEGV in VDMListDelegateDataType::createMissingProperties when PipeWire restarts with cachedAudioSinks


dorsetineb:
### System details
- Omarchy version: 4.0.2-1
- Quickshell: 0.3.1-1
- Qt: 6.11.2-1
- PipeWire: 1.6.8-1

### What happened?
When PipeWire restarts or disconnects while `quickshell` is running, `quickshell` crashes with `SIGSEGV` (signal 11) inside `libQt6QmlModels.so.6`.

### What was expected?
Quickshell should handle the PipeWire disconnect/reconnect gracefully without crashing.

### Steps to reproduce
1. Run Quickshell with the Omarchy shell: `quickshell -n -p /usr/share/omarchy/shell`
2. Open the audio panel (or have audio sinks populated in `cachedAudioSinks`)
3. Restart PipeWire: `systemctl --user restart pipewire wireplumber`
4. Quickshell crashes with SIGSEGV.

### Stack trace
```text
#0  VDMListDelegateDataType::createMissingProperties (this=..., row=...)
    at /usr/src/debug/qt6-declarative/qtdeclarative/src/qmlmodels/qqmldmlistaccessordata_p.h:222
#1  VDMListDelegateDataType::toDynamicMetaObject (this=..., object=...)
    at /usr/src/debug/qt6-declarative/qtdeclarative/src/qmlmodels/qqmldmlistaccessordata.cpp:157
#2  qQmlPropertyCacheProperty<QV4::String const*> (...)
    at /usr/src/debug/qt6-declarative/qtdeclarative/src/qml/qml/qqmlpropertycache.cpp:1059
#3  QQuickRepeater::regenerate() at /usr/lib/libQt6Quick.so.6
#4  QQuickRepeater::setModel(...) at /usr/lib/libQt6Quick.so.6
#5  QQmlTimer::event(QEvent*) at /usr/lib/libQt6QmlMeta.so.6
```

### Root Cause
In `shell/plugins/panels/audio/Panel.qml`:
1. `audioSinks` falls back to `cachedAudioSinks` when `rawAudioSinks.length === 0`:
   ```qml
   readonly property var audioSinks: rawAudioSinks.length > 0 ? rawAudioSinks : cachedAudioSinks
   ```
2. When PipeWire disconnects or restarts, `rawAudioSinks` becomes empty, causing `audioSinks` to fall back to `cachedAudioSinks`.
[...]

PR #9808, opened 2026-09-02
Deduplicate audio nodes and filter portless ALSA sinks

gladimdim:
### Problem

When external displays (e.g. Type-C/DisplayPort/HDMI) are connected or the system wakes from sleep, WirePlumber/PipeWire may re-enumerate audio endpoints without promptly unbinding previous nodes, leaving stale or duplicate nodes in the PipeWire graph.

This caused several issues in Omarchy:
1. **Duplicate outputs in the audio panel:**
   `shell/plugins/panels/audio/Panel.qml` gathered all candidate sinks from PipeWire without deduplicating by `node.name`. When duplicate nodes for the same sink name existed, the panel rendered each one as a separate duplicate entry (e.g. multiple identical HDMI entries).
2. **False availability of zombie/unplugged ALSA sinks:**
   `omarchy-audio-sink-availability` used `((port_count == 0 || available) ? 1 : 0)` to check sink availability, intending to treat virtual sinks without ports (like EasyEffects) as available. However, disconnected or orphaned ALSA sinks (`alsa_output.*`) also report 0 ports in `pactl`, causing them to be falsely marked as available (`1`). Furthermore, duplicate entries with the same sink name were emitted repeatedly.
3. **Output switcher cycling onto dead sinks:**
   `omarchy-audio-output-switch` used `((.ports | length == 0) or ([.ports[]? | .availability != "not available"] | any))` and did not deduplicate by name, causing the switcher hotkey to cycle onto dead sinks.

### Solution

- **`shell/plugins/panels/audio/Model.js` & `Panel.qml`:**
  - Added `Model.deduplicateNodes(nodes, preferredNode)` to deduplicate candidate sinks and sources by unique `node.name`, prioritizing the current active default sink/source (or newest ID).
- **`bin/omarchy-audio-sink-availability`:**
  - Differentiated ALSA hardware sinks (`alsa_output.*`) from virtual sinks: ALSA sinks require at least one active port (`port_count > 0 && available`), while virtual sinks without ports remain available (`port_count == 0`).
  - Aggregated availability by sink name so each unique sink is emitted cleanly once.
- **`bin/omarchy-audio-output-switch`:**
  - Applied the ALSA port requirement and piped through `unique_by(.name)`.
- **`test/shell.d/audio-test.sh`:**
  - Added unit tests for `Model.deduplicateNodes`.

Issue #9647, opened 2026-09-01
quickshell crash (SIGSEGV) during PipeWire audio device switch — QML Repeater binding update


kczimm:
## What happened

Quickshell 0.3.1-1 crashed with SIGSEGV at 12:32:51 CDT while running the Omarchy shell config (`-p /usr/share/omarchy/shell`). The signal was self-inflicted (`si_code: SI_TKILL`) — quickshell's crash handler or a Qt assertion raised SIGSEGV, not a hardware memory fault.

## Trigger

A Bluetooth audio device disconnected, causing PipeWire to tear down and re-enumerate devices. The default sink switched from `bluez_output.08_DF_1F_35_07_90.1` to `alsa_output.pci-0000_01_00.1.hdmi-stereo`. The crash occurred ~6 seconds after the initial PipeWire error (code -14).

## Stack trace

The crashing thread was in a `QQuickRepeater::setModel` → `regenerate` cycle, incubating a new QML item. During module evaluation, `TypeofName::call` resolved a property and the engine hit an invalid state:

```
#0  raise (libc.so.6)
#2  quickshell + 0x124389
#4  QV4::QObjectWrapper::getQmlProperty
#7  QV4::Runtime::TypeofName::call
#13 QV4::Module::evaluate
#16 QQmlIncubatorPrivate::incubate
#19 QQuickRepeater::regenerate
#20 QQuickRepeater::setModel
#22 QQmlPropertyPrivate::write
#26 QQmlBinding::slowWrite
```

## System details

- Omarchy 4.0.2-1, Arch Linux rolling
- Quickshell 0.3.1-1, Qt 6.11.2
- NVIDIA GPU (10de:2d04), nvidia-drm driver
- Memory healthy: 7.6 GB used / 15 GB total, no OOM
- One-off crash — no previous quickshell crashes in coredumpctl
- Full crash report saved at `/home/kevin/.cache/quickshell/crashes/0x2cu0pkt/`

## Notes

This is a one-off crash, not a repeated pattern. It was triggered by a Bluetooth headset disconnect causing a PipeWire device switch. The QML binding update in the audio widget likely hit an invalid state during repeater model regeneration.

ggregoro:
Hit the same crash on a different trigger, and pulled a symbolized backtrace off the core that pins the root cause.

## Same failure

- quickshell 0.3.1-1, Qt 6.11.2, Omarchy 4.0.2-1, Arch rolling
[...]

Issue #9256, opened 2026-08-30
Restarting audio services crashes QuickShell in QQuickRepeater::regenerate


kstimpson85:
## What happened

Running `omarchy restart audio` caused the Omarchy QuickShell process to crash with SIGSEGV. QuickShell restarted automatically, so no user data was lost.

The journal shows the audio restart stopping PipeWire, WirePlumber, and PipeWire PulseAudio. During QuickShell's PipeWire disconnect/reconnect handling, QuickShell crashed while regenerating a QML `Repeater`.

Crash run ID: `pvsrllkt`

## Steps to reproduce

1. Run `omarchy restart audio`.
2. Observe the Omarchy shell crash and automatically restart.

This was observed once. I have not repeatedly reproduced it to avoid unnecessarily disrupting the desktop session.

## Expected behavior

Audio services restart without terminating the Omarchy shell.

## Actual behavior

QuickShell receives SIGSEGV and restarts.

## Relevant timeline

```text
17:18:11 omarchy-restart-audio invoked
17:18:11 PipeWire, WirePlumber, and PipeWire PulseAudio stopped
17:18:11 QuickShell reported PipeWire connection error
17:18:11 PipeWire services restarted
17:18:11 QuickShell destroyed and recreated its PipeWire node model
17:18:11 QuickShell crashed with SIGSEGV
```

## Relevant stack frames

```text
libQt6QmlModels.so.6
QV4::QObjectWrapper::getQmlProperty
QV4::Module::evaluate
QQmlObjectCreator::createInstance
QQmlIncubatorPrivate::incubate
QQuickRepeater::regenerate
QQuickRepeater::setModel
QQmlBinding::update
QQmlTimer::event
```

The shipped audio panel already snapshots PipeWire-backed lists before feeding them to `Repeater`, with a comment explaining that rebuilding a Repeater during PipeWire node removal has previously crashed QuickShell. This crash indicates that the current protection may not cover every model update during a full audio-service restart.

## System details

- Omarchy: `4.0.1-1`
- QuickShell: `0.3.1-1`
- Qt: `6.11.2`
- PipeWire: `1.6.8`
- WirePlumber: `0.5.15`
- Kernel: `7.1.9-arch1-2`
- Shell config: `/usr/share/omarchy/shell/shell.qml`
- Hardware: MacBookPro13,3, Intel HD 530 + AMD Radeon Pro 455

The complete local QuickShell report is stored at:

[...]

PR #9959, opened 2026-09-03
Drop destroyed PipeWire nodes before they reach the audio panel's repeaters

tomputer22:
Fixes the SIGSEGV reported in #6952, #9914, #9647 and #9256 — four open issues that are the same crash seen from four different triggers (a USB DAC timing out, a PipeWire restart, an ALSA capture device going busy, and plain graph churn).

## The crash

When PipeWire restarts, or any graph churn re-enumerates devices, Quickshell destroys the `PwNode` objects backing the audio panel. `rawAudioSinks` goes empty, `audioSinks` falls back to `cachedAudioSinks`, and that cache still references the destroyed nodes. `listSnapshot()` was a bare `list.slice()`, so those dangling references were copied straight into `displayAudioSinks` and handed to a `Repeater`. Regenerating it dereferences them as a null `QObject*` during delegate incubation:

```
VDMListDelegateDataType::createMissingProperties
  qtdeclarative/src/qmlmodels/qqmldmlistaccessordata_p.h:222

  const QMetaObject *metaObject = row->value<QObject *>()->metaObject();
```

That line has no null guard, so the shell segfaults. Upstream could harden it, but the dangling reference is ours.

The 75 ms snapshot timer debounces the *list* mutation, which is what the comment at `Panel.qml:103` is about — but it never detached the *elements*. That is the gap this closes.

## The fix

Snapshots are taken against the live PipeWire node list, so a node that is already gone cannot reach a delegate. Membership is an identity comparison and never dereferences the candidate, which is what makes it safe to test on a node that may already be destroyed.

## Behaviour change

The cached fallback no longer serves nodes PipeWire has dropped, so on a full restart the device list is briefly empty rather than showing stale entries. It repopulates on the next refresh. Showing devices that no longer exist is what crashed, so I think empty is the correct trade — but it is a visible change and worth a maintainer's opinion.

## Testing

**Crash reproduced, and confirmed fixed.** On Omarchy 4.0.1-1 / Quickshell 0.3.1 / Qt 6.11.2, in an isolated shell config that mirrors the panel's shape — a `Repeater` over a plain array of `PwNode`s whose delegate declares `required property var modelData` — with `systemctl --user restart pipewire wireplumber` as the trigger:

| | before restart | after restart | assigning stale cache | result |
|---|---|---|---|---|
| `listSnapshot()` (old) | 9 nodes cached | 11 new live nodes | — | **SIGSEGV**, execution never reaches the next statement |
| `liveNodeSnapshot()` (new) | 11 nodes cached | 11 new live nodes | 0 rows survive the filter | clean exit, no crash |

The old run logged `Quickshell has crashed under pid …`; the new run logged `assigned_rows: 0` and exited 0.

<details>
<summary>Harness used (drop in a directory as <code>shell.qml</code> alongside <code>Model.js</code>, run <code>quickshell -n -p &lt;dir&gt;</code>)</summary>

```qml
import QtQuick
import Quickshell
import Quickshell.Services.Pipewire
import "Model.js" as Model

ShellRoot {
  id: root
  property var cache: []
  property var displayList: []

  PwObjectTracker { objects: Pipewire.nodes ? Pipewire.nodes.values : [] }

  Item {
    Repeater {
      model: root.displayList
      delegate: Item { required property var modelData }
    }
  }

  Timer { interval: 3000; running: true; onTriggered: {
    root.cache = Pipewire.nodes.values.slice()
    console.log("PWTEST cached_nodes:", root.cache.length)
  } }

  Timer { interval: 4000; running: true; onTriggered: {
    Quickshell.execDetached(["systemctl", "--user", "restart", "pipewire", "wireplumber"])
  } }

  Timer { interval: 9000; running: true; onTriggered: {
    console.log("PWTEST live_nodes_now:", Pipewire.nodes.values.length)
    root.displayList = Model.liveNodeSnapshot(root.cache, Pipewire.nodes.values)
    console.log("PWTEST assigned_rows:", root.displayList.length)
  } }

  Timer { interval: 11000; running: true; onTriggered: { console.log("PWTEST SURVIVED"); Qt.callLater(Qt.quit) } }
}
```

Swap the assignment for `Model.listSnapshot(root.cache)` against the pre-patch `Model.js` to reproduce the crash.
</details>

**Unit tests.** `test/shell.d/audio-test.sh` gains six cases: live nodes, destroyed nodes, a fully emptied graph, null entries, and both missing-argument paths. Whole file passes.

**Suite.** `test/shell` — 223/226 files pass. The three failures (`config`, `snapper`, `unowned-system-paths`) fail identically on a clean checkout here, so they are pre-existing and environment-sensitive.

**Also verified** that `Pipewire.nodes.values` returns identical JS wrappers for the same `PwNode` across separate accesses, and rejects foreign objects — the assumption the membership test rests on.

**Not done:** verification inside the running shell via `omarchy dev link`, which needs a reboot. The harness above exercises the same code path, delegate shape and trigger.

## Origin

Diagnosed from a `systemd-coredump` core after `pipewire`, `wireplumber` and `pipewire-pulse` were restarted with the audio panel open — the exact reproduction in #9914. Symbolized stack matches #9647 and #6952 frame for frame.

## Related PRs

#7783 targets the same defect from the other direction — primitive `{id, name}` rows resolved through a `nodeFor()` lookup, the pattern `plugins/bluetooth/Model.js` already uses. That is immune by construction rather than by call-path analysis, at the cost of a much larger diff. Which shape the repo wants is a maintainer's call; I am not touching that PR.

#9808 deduplicates `rawAudioSinks`/`rawAudioSources` by name. It is upstream of this snapshot rather than in competition with it, but it will conflict textually in `Model.js`'s exports block and at the end of `audio-test.sh`. Happy to rebase behind it if it lands first.

> Authored by Claude Opus 5 via Claude Code.

### DIFF
--- a/shell/plugins/panels/audio/Model.js
+++ b/shell/plugins/panels/audio/Model.js
@@ -18,8 +18,34 @@ function isAudioSource(node) {
     || mediaClass.indexOf("Source") !== -1
 }
 
-function listSnapshot(list) {
-  return list && list.slice ? list.slice() : []
+// PipeWire can destroy a PwNode while the panel still holds a reference to it:
+// during a service restart, or on any graph churn that re-enumerates devices.
+// A destroyed node left in a Repeater's model is dereferenced as a null
+// QObject* while the delegate is incubated -- qtdeclarative's
+// qqmldmlistaccessordata_p.h:222 calls row->value<QObject *>()->metaObject()
+// with no null guard -- and the shell segfaults.
+//
+// Keeping only the nodes PipeWire still lists drops those references before
+// they can reach a Repeater. Membership is an identity comparison, so it is
+// safe to test on a node that is already gone.
+function liveNodeSnapshot(list, liveNodes) {
+  if (!list || !list.length) return []
+  if (!liveNodes || !liveNodes.length) return []
+
+  var out = []
+  for (var i = 0; i < list.length; i++) {
+    var node = list[i]
+    if (!node) continue
+    // Indexed access only: the live list arrives as a QML sequence type, which
+    // is not guaranteed to carry Array.prototype methods such as indexOf.
+    for (var j = 0; j < liveNodes.length; j++) {
+      if (liveNodes[j] === node) {
+        out.push(node)
+        break
+      }
+    }
+  }
+  return out
 }
 
 function outputVolumeName(volume, muted) {
@@ -237,7 +263,7 @@ if (typeof module !== "undefined") {
   module.exports = {
     isPlaybackStream: isPlaybackStream,
     isAudioSource: isAudioSource,
-    listSnapshot: listSnapshot,
+    liveNodeSnapshot: liveNodeSnapshot,
     outputVolumeName: outputVolumeName,
     parseSinkAvailability: parseSinkAvailability,
     friendlyDeviceLabel: friendlyDeviceLabel,
--- a/shell/plugins/panels/audio/Panel.qml
+++ b/shell/plugins/panels/audio/Panel.qml
@@ -324,8 +324,12 @@ Panel {
   onAudioSourcesChanged: scheduleDisplayAudioModelRefresh()
   onAudioStreamsChanged: scheduleDisplayAudioModelRefresh()
 
+  // Snapshots are taken against the live PipeWire node list so that a node
+  // destroyed since the list was built -- most often one still held by
+  // cachedAudioSinks/cachedAudioSources after a PipeWire restart -- never
+  // reaches a Repeater delegate. See Model.liveNodeSnapshot.
   function listSnapshot(list) {
-    return Model.listSnapshot(list)
+    return Model.liveNodeSnapshot(list, nodes)
   }
 
   function refreshDisplayAudioModels() {
--- a/test/shell.d/audio-test.sh
+++ b/test/shell.d/audio-test.sh
@@ -46,4 +46,18 @@ assertEqual(audio.matchingMprisStreamLabel('Chromium', players), 'Chromium', 'au
 assertEqual(audio.unmatchedMprisStreamLabel('audio-src', players, streams), 'Spotify', 'audio uses unmatched MPRIS player for generic streams')
 assertEqual(audio.streamLabel(streams[1], players, streams), 'Spotify', 'audio labels generic streams from MPRIS')
 assert(audio.streamRepresentsPlayer(streams[1], players[0], players, streams), 'audio links generic streams to active player')
+
+// A PwNode destroyed by PipeWire must never survive into a Repeater's model:
+// Qt dereferences it as a null QObject* during delegate incubation and the
+// shell segfaults (basecamp/omarchy#9914, #9647, #9256).
+const speakers = { name: 'alsa_output' }
+const headset = { name: 'bluez_output' }
+const destroyed = { name: 'raop_sink' }
+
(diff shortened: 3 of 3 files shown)


andresleecom:
Confirming this still reproduces on omarchy 4.0.3-1 / quickshell 0.3.1-1 / qt6-declarative 6.11.2-1, and that this PR's approach addresses it.

### Different trigger from the reports cited here

#9914 and #10371 reproduce through `systemctl --user restart pipewire` / `omarchy-restart-audio`. Mine is physical: hot-unplugging a USB hub carrying two USB audio devices (Beats Studio Pro, RØDE PodMic USB). Three crashes in three days, each on a hub unplug or replug:

```
Sep  8 15:03:33  usb 1-12: USB disconnect  ->  SIGSEGV at 15:03:33  (same second)
Sep  9 12:28:23  usb 1-12: USB disconnect  ->  SIGSEGV at 12:28:23  (same second)
Sep 10 10:05:26  usb 1-12: USB disconnect
Sep 10 10:06:22  usb 1-12: reconnect       ->  SIGSEGV at 10:06:33
```

All three cores carry byte-identical stacks, so it is deterministic rather than a narrow race. Worth recording because it needs no user interaction and no service restart: pulling the cable is enough, as long as the audio panel is open (`refreshDisplayAudioModels()` returns early on `!opened`).

### Memory-level confirmation of the null dereference

The existing reports correctly deduce that `row->value<QObject *>()` is null. Reading it out of the core makes it explicit:

```
(gdb) frame 5
#5  VDMListDelegateDataType::createMissingProperties (this=0x7f3645002800, row=0x7f363e7fd780)
    at qtdeclarative/src/qmlmodels/qqmldmlistaccessordata_p.h:222

(gdb) p row->d.is_shared                        $1 = 0
(gdb) p/x row->d.data.data                      $2 = {0x0 <repeats 24 times>}
(gdb) info symbol 0x7f366aa78800
QtPrivate::QMetaTypeInterfaceWrapper<QObject*>::metaType
(gdb) p/x *(unsigned int*)(0x7f366aa78800+8)    $3 = 0x80c
```

`is_shared = 0` means the payload is stored inline, so those 24 zero bytes are the pointer itself. Flags `0x80c` are `IsPointer | PointerToQObject | RelocatableType`, so the guard on line 221 passes and line 222 dereferences null.

The metatype being erased to plain `QObject*` rather than `PwNode*` is the signature of a wrapper whose object has been destroyed, which is precisely what `liveNodeSnapshot` filters out. `propertyCache = {o = 0x0}` with zero dynamic properties built also shows it dies on the first row of a freshly assigned model, not partway through.

### One suggestion beyond this PR

The unguarded dereference is generic: any `Repeater` fed a plain JS array of QObjects is exposed, not only the audio panel. The same shape appears in the tray (`root.allItems`, `root.pinnedItems`), media (`root.sourcePlayers`), and the bluetooth, network and tailscale panels, all holding objects whose lifetime belongs to an external service.

`liveNodeSnapshot` fixes the panel that crashes today. Promoting it to a shared helper in `Commons`, filtering an array against its owning service's live collection, would head off the next one. Happy to open that separately if you would rather keep this PR narrow.

Upstream, `qqmldmlistaccessordata_p.h:222` looks worth a QTBUG in its own right: every sibling branch in `createMissingProperties` is defensive, and only the `PointerToQObject` branch dereferences unchecked.

bjarneo:
Automated duplication check: this pull request looks similar to #7783, which covers the same audio Repeater node filter. I keep that one open and close this one to consolidate review.

PR #9962, opened 2026-09-03
Theme mpv with the active Omarchy palette

paulnunezm:
mpv and mpv-mpris already ship in the base packages, but nothing repaints mpv on theme changes. This wires mpv into the existing theming pipeline so the player bar, menus, and OSD follow `omarchy theme set` like every other app.

## What it adds

- `default/themed/mpv-osc.conf.tpl` — oscc bar palette (`bg/fg/accent/track/pressed/tooltip_outline` + bottombar layout), rendered from `colors.toml` by `omarchy-theme-set-templates`
- `default/themed/mpv-console.conf.tpl` / `mpv-context_menu.conf.tpl` — console and context-menu colors
- `bin/omarchy-theme-set-mpv` (hidden) — copies the staged files to `~/.config/mpv/script-opts/`, maintains a managed `osc=no` + OSD-color block in `mpv.conf` (preserving user settings), and seeds the oscc script/font for users who predate the shipped config
- `config/mpv/scripts/oscc.lua` + `config/mpv/fonts/oscc.ttf` — oscc player bar (vendored from https://github.com/0dist/oscc, Omarchy palette wiring marked with OMARCHY comments) seeded to new users via /etc/skel
- `bin/omarchy-theme-set` — runs `omarchy-theme-set-mpv` in `post_theme_commands`, so theme changes apply automatically with no manual hook install
- `migrations/1788435827.sh` — seeds the bar + applies the theme once for existing installs (idempotent)
- `test/shell.d/theme-staging-test.sh` — classifies the new templates as colour-only, as the denylist test requires

## Automatic theme changes (no user hook needed)

The previous standalone version required `omarchy hook install theme-set`. Here that is replaced by the native mechanism: `omarchy-theme-set` renders the templates, then runs `omarchy-theme-set-mpv` alongside the other app retints. The vendored oscc polls `script-opts/osc.conf` every 0.5s and rebuilds its styles in-process, so a running player repaints within ~1s of `omarchy theme set` with no restart; menus and OSD apply on next launch.

The managed `mpv.conf` markers (`# >>> omarchy-mpv-theme >>>`) are kept identical to the standalone project, so existing users of it migrate cleanly — the old block is replaced, not duplicated.

## Verification

- Rendered templates against tokyo-night colors and applied via `omarchy-theme-set-mpv` in a scratch HOME: staged + applied colors match, user `mpv.conf` content preserved, managed block idempotent across reruns
- `./test/cli` — pass
- `test/shell.d/theme-staging-test.sh` — pass (including the new colour classification)
- Full `./test/shell`: 5 failures, all confirmed pre-existing on the clean tree (missing omarchy-pkgs checkout, bar-icon geometry, plymouth, snapper, config PKGBUILD coverage)

## Notes for reviewers

- Upstream oscc has no declared license; happy to drop the vendored bar and fall back to stock-OSC theming (OSD + menus only) if you prefer a smaller footprint.
- No `install.sh`/`uninstall.sh`: installation is via the normal config-seed + migration path per the repo layout.


paulnunezm:
Here is a preview.


https://github.com/user-attachments/assets/163c388d-ecf9-4290-9752-cd327ade02cd

PR #9966, opened 2026-09-03
Default OMARCHY_PATH in update-dev for sudo env_reset

### LINKED ISSUE #9953: sudo omarchy update aborts with "OMARCHY_PATH: unbound variable" in omarchy-update-dev
### Summary

`sudo omarchy update` aborts right after the system snapshot with:

```
Prune package cache
==> no candidate packages found for pruning

Create system snapshot
/usr/bin/omarchy-update-dev: line 7: OMARCHY_PATH: unbound variable

```

The update never reaches `omarchy-update-system-pkgs`, so no packages are upgraded.

### What happens

`omarchy-update-dev` runs with `set -euo pipefail` and its first check dereferences `OMARCHY_PATH`:

```bash
set -euo pipefail

[[ $OMARCHY_PATH != "/usr/share/omarchy" ]] || exit 0   # line 7
```

`OMARCHY_PATH` is normally exported by `/etc/profile.d/omarchy.sh` (login shells) and the uwsm session env, but **`sudo` resets the environment** (`env_reset`), so when the update runs as root the variable is unset. Under `set -u` the unbound reference kills the script, the parent `omarchy-update` ERR trap fires, and the whole update is aborted.

The script clearly intends to be a no-op on stock installs (it exits when the path points at `/usr/share/omarchy`), it just doesn't handle the case where the variable is not set at all.

### Steps to reproduce

1. On a stock (non dev-linked) Omarchy install, from a normal terminal:
   `sudo omarchy update`
2. Update proceeds through cache prune + snapper snapshot, then dies at `omarchy-update-dev` with `OMARCHY_PATH: unbound variable`.

### Expected

The update should continue normally. `omarchy-update-dev` should treat an unset `OMARCHY_PATH` as the stock install location.

### Suggested fix

```bash
[[ ${OMARCHY_PATH:-/usr/share/omarchy} != "/usr/share/omarchy" ]] || exit 0
```

(Alternatively, `omarchy-update` could set `OMARCHY_PATH=${OMARCHY_PATH:-/usr/share/omarchy}` before calling into sub-scripts.)

### Workaround

```bash
sudo env OMARCHY_PATH=/usr/share/omarchy omarchy update
```

### System details

- Omarchy package: `omarchy 4.0.2-1`, `aether 4.29.8-1`
- Kernel: 7.1.9-arch1-2
- Hyprland 0.56.2
- Hardware: HUAWEI KPL-W0X laptop, AMD Ryzen 5 2500U, amdgpu

Full diagnostic output is available at `/tmp/omarchy-debug.log` (can attach on request).

fresh3nough:
## Summary

`sudo omarchy update` aborts right after the system snapshot with:

```
/usr/bin/omarchy-update-dev: line 7: OMARCHY_PATH: unbound variable
```

`omarchy-update-dev` runs under `set -euo pipefail` and dereferences `OMARCHY_PATH` before anything else. `sudo` resets the environment, so the variable is unset on stock installs and the step kills the whole update before packages are upgraded.

## Fix

Default a missing path to the packaged install, matching `omarchy-migrate`, `omarchy-version`, and the other privileged helpers:

```bash
OMARCHY_PATH="${OMARCHY_PATH:-/usr/share/omarchy}"
```

Stock installs still exit 0 immediately. Dev checkouts keep pulling when the path is set.

Fixes #9953

## Test plan

- [x] `bash test/shell.d/update-dev-test.sh` — all cases pass, including new ones:
  - unset `OMARCHY_PATH` (sudo `env_reset`) skips without unbound-variable error
  - empty `OMARCHY_PATH` skips as stock
  - stock path still skips git
  - custom checkout still `pull --ff-only`
  - invalid checkout still fails with a useful error
  - no-upstream checkout still skips safely
- [x] `bash test/shell.d/update-sequence-test.sh` — update step order unchanged
- [x] Manual repro:
  - **Before:** `env -u OMARCHY_PATH bin/omarchy-update-dev` → `OMARCHY_PATH: unbound variable` (exit 1)
  - **After:** same command exits 0 with no git activity
  - Custom `OMARCHY_PATH` still runs `git -C … pull --ff-only`

### DIFF
--- a/bin/omarchy-update-dev
+++ b/bin/omarchy-update-dev
@@ -4,6 +4,10 @@
 
 set -euo pipefail
 
+# sudo env_reset drops session exports. Treat a missing path as the packaged
+# install so this step is a no-op there instead of aborting under `set -u`.
+OMARCHY_PATH="${OMARCHY_PATH:-/usr/share/omarchy}"
+
 [[ $OMARCHY_PATH != "/usr/share/omarchy" ]] || exit 0
 
 if ! git -C "$OMARCHY_PATH" rev-parse --is-inside-work-tree >/dev/null 2>&1; then
--- a/test/shell.d/update-dev-test.sh
+++ b/test/shell.d/update-dev-test.sh
@@ -53,10 +53,36 @@ run_dev_update() {
     "$ROOT/bin/omarchy-update-dev"
 }
 
+# sudo env_reset (and similar clean environments) leave OMARCHY_PATH unset.
+# The stock path must still be treated as a no-op so `sudo omarchy update` can
+# continue past this step instead of dying under `set -u`.
+run_dev_update_without_path() {
+  env -u OMARCHY_PATH \
+    TEST_GIT_LOG="$git_log" \
+    PATH="$stub_bin:$PATH" \
+    "$ROOT/bin/omarchy-update-dev"
+}
+
 : >"$git_log"
 run_dev_update /usr/share/omarchy
 [[ ! -s $git_log ]] || fail "package-backed updates do not invoke git" "$(cat "$git_log")"
 pass "package-backed updates skip the dev checkout step"
+
+: >"$git_log"
+run_dev_update_without_path >"$test_tmp/unset.out" 2>"$test_tmp/unset.err" ||
+  fail "unset OMARCHY_PATH must not abort under set -u" "$(cat "$test_tmp/unset.err")"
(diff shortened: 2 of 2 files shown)


bjarneo:
Automated duplication check: this pull request looks similar to #8590, which covers the same OMARCHY_PATH default in update-dev. I keep that one open and close this one to consolidate review.

PR #9969, opened 2026-09-03
Add omarchy font cjk commands for locale-aware CJK fallback

### LINKED ISSUE #9580: CJK font fallback always selects Korean variant regardless of locale
## System details

```
Omarchy v4.0.2-1, Foot, JetBrainsMono Nerd Font
LANG=en_US.UTF-8 (default)
```

## What's wrong?

CJK characters (e.g., "全", "骨", "円") always render with Korean variant (Noto Sans CJK KR) regardless of the user's locale. The sans-serif fallback chain in 65-nonlatin.conf only includes Noto Sans CJK KR, but not other CJK variants (JP, SC, TC, HK).

### Steps to reproduce

1. Use default settings on Omarchy v4.0.2-1 (LANG=en_US.UTF-8)
2. Display "全" in Foot terminal → Korean variant
3. Type "全" in Chrome omnibox input field → Korean variant

usutani:
Fixes #9580.

CJK characters always resolve to the Korean variant (Noto Sans CJK KR) regardless of the user's locale, because the sans-serif chain in the system `65-nonlatin.conf` only lists KR and Omarchy has no locale-aware selection.

## What this adds (post-install only)

- `omarchy font cjk list` — installed CJK variants (jp/kr/sc/tc/hk)
- `omarchy font cjk set <variant|auto>` — writes `~/.config/fontconfig/conf.d/50-omarchy-cjk.conf`, owned outright by this command; the user's `fonts.conf` is left alone by construction
- `omarchy font cjk current` — shows the active variant (`unset` when the file is absent)
- `omarchy font cjk unset` — removes the file (`rm -f`)
- `auto` maps LANG (ja→jp, ko→kr, zh_CN/SG→sc, zh_TW/HK/MO→tc); `en_*` and unknown locales prompt instead of guessing. `zh_HK` maps to `tc` by design, so `hk` is reachable only by naming it explicitly. A stock install only generates `C.UTF-8` and `en_US.UTF-8`, so `auto` always prompts there until additional locales exist
- `omarchy font set` is untouched by this PR (no shared file, no cooperation needed)
- Style > Font > CJK Variant menu backed by a `cjk-variants` provider
- The `prepend` rules cover sans-serif/serif/monospace plus a `lang` rule for the Chromium per-character path that ignores `lang`. The `lang` rule uses `mode="append"` so a monospace query never gets a proportional CJK face at the head of its chain; charset matching still outranks family matching, so the bare-charset path keeps working

## Design notes

- Explicit stored preference wins over implicit locale tracking: `auto` only proposes from LANG on first run; later LANG changes do not move the selection.
- `ko` mapping keeps KR (status quo), so Korean users see no behavior change.
- `conf.d` loads before `fonts.conf`, so the `prepend_first` monospace rule `omarchy font set` writes keeps winning the head of the chain.

## Verification (author)

| Check | Result |
|---|---|
| `cjk-set jp/kr` flips `fc-match -s` CJK order | Pass |
| `cjk-current` / `cjk-list` | Pass (`jp`, all 5 variants) |
| `auto` on `LANG=en_US.UTF-8` prompts | Pass |
| `menu-test`, `menu-guards-test`, `./test/cli` | Pass |
| Foot/Chromium rendering | Verified JA locale only |

Revision `a71d175` (review feedback, sandbox `HOME` + `fc-list` stub):

| Check | Result |
|---|---|
| `set tc` → `current tc`, `lang` rule `mode="append"`, XML parses | Pass |
| `unset` → file removed, `current unset` | Pass |
| `auto` on `zh_MO` → `tc`, `zh_HK` → `tc`, `bogus` → exit 1 | Pass |
| `fonts.conf` never created or modified | Pass |
| `./test/cli` | 116 ok / 0 fail |

## Not verified by author

ko, zh_CN, zh_TW rendering — requesting native speakers to test. ko mapping is unchanged from status quo, so Korean users see no behavior change. (zh_TW verified by reviewer `linyiru`; re-verification of the `append` revision requested.)

## Out of scope

- Install-time locale prompt (needs a fresh implementation in the current ISO/Quattro flows; follow-up, see #10949)
- Changes to the Arch `fontconfig` package (`65-nonlatin.conf` is intentionally locale-agnostic upstream)
- XDG directory renaming

## User-config side effects

- Only `~/.config/fontconfig/conf.d/50-omarchy-cjk.conf` is written.
- `~/.config/fontconfig/fonts.conf` is untouched by `cjk-set`.
- Terminal configs (`foot.ini` etc.) are untouched by `cjk-set`.

- Per-application tuning (e.g. Foot's fallback font sizes in `foot.ini`)

### DIFF
--- a/bin/omarchy-font-cjk-current
+++ b/bin/omarchy-font-cjk-current
@@ -1,0 +1,22 @@
+#!/bin/bash
+
+# omarchy:summary=Show current CJK font variant preference
+# omarchy:examples=omarchy font cjk current
+
+# The CJK variant marker lives in fonts.conf (the only per-user fontconfig
+# file this system loads), managed by `omarchy font cjk set`.
+cjk_file="$HOME/.config/fontconfig/fonts.conf"
+
+if [[ ! -f "$cjk_file" ]]; then
+  echo "unset"
+  exit 0
+fi
+
+# The variant marker is written by cjk-set as "<!-- omarchy-cjk-variant: <name> -->".
+current=$(sed -n 's/.*omarchy-cjk-variant: \([a-z]*\).*/\1/p' "$cjk_file" | tail -n1)
+
+if [[ -z "$current" ]]; then
+  echo "unset"
+else
+  echo "$current"
+fi
--- a/bin/omarchy-font-cjk-list
+++ b/bin/omarchy-font-cjk-list
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+# omarchy:summary=List available CJK font variants
+# omarchy:examples=omarchy font cjk list
+
+# Each variant is a Noto CJK sub-family installed by noto-fonts-cjk.
+# We check for the sans-serif variant as a proxy — if it is present,
+# the serif and mono siblings are too.
+variants=("jp" "kr" "sc" "tc" "hk")
+
+for v in "${variants[@]}"; do
+  upper=$(echo "$v" | tr '[:lower:]' '[:upper:]')
+  if fc-list | grep -Fqi "Noto Sans CJK $upper"; then
+    echo "$v"
+  fi
+done
--- a/bin/omarchy-font-cjk-set
+++ b/bin/omarchy-font-cjk-set
@@ -1,0 +1,187 @@
+#!/bin/bash
+
+# omarchy:summary=Set the preferred CJK font variant
+# omarchy:args=<variant>
+# omarchy:examples=omarchy font cjk set jp | omarchy font cjk set auto
+
+usage() {
+  cat <<'EOF'
+Usage: omarchy font cjk set <variant>
+
+Variants:
+  jp    Japanese (Noto Sans CJK JP)
+  kr    Korean (Noto Sans CJK KR)
+  sc    Simplified Chinese (Noto Sans CJK SC)
+  tc    Traditional Chinese (Noto Sans CJK TC)
+  hk    Hong Kong Chinese (Noto Sans CJK HK)
+  auto  Detect from LANG environment variable (prompts if LANG is en_*)
+EOF
+}
+
+variant="${1:-}"
+
+case "$variant" in
+  -h|--help)
+    usage
+    exit 0
+    ;;
+  "")
+    usage >&2
+    exit 1
+    ;;
+esac
+
+# --- Resolve variant name ---
+
+resolve_auto() {
+  local lang="${LANG:-en_US.UTF-8}"
+  local base="${lang%%.*}"
+  base="${base%%@*}"
+  local territory="${base##*_}"
+  local language="${base%%_*}"
+
+  case "$language" in
+    ja)
+      echo "jp"
+      ;;
+    ko)
+      echo "kr"
+      ;;
+    zh)
+      case "$territory" in
+        CN|SG) echo "sc" ;;
+        TW|HK) echo "tc" ;;  # zh_HK defaults to TC per design decision
+        *)     echo "sc" ;;
+      esac
+      ;;
+    *)
+      # en_* or unknown — prompt the user
+      echo ""
+      ;;
+  esac
+}
+
+case "$variant" in
+  auto)
+    resolved=$(resolve_auto)
+    if [[ -z "$resolved" ]]; then
+      echo "LANG=$LANG does not map to a CJK variant."
+      echo ""
+      echo "Select CJK variant:"
+      echo "  1) jp  — Japanese (Noto Sans CJK JP)"
+      echo "  2) kr  — Korean (Noto Sans CJK KR)"
+      echo "  3) sc  — Simplified Chinese (Noto Sans CJK SC)"
+      echo "  4) tc  — Traditional Chinese (Noto Sans CJK TC)"
+      echo "  5) hk  — Hong Kong Chinese (Noto Sans CJK HK)"
+      echo ""
+      read -rp "Number [1-5]: " choice
+      case "$choice" in
+        1) resolved="jp" ;;
+        2) resolved="kr" ;;
+        3) resolved="sc" ;;
+        4) resolved="tc" ;;
+        5) resolved="hk" ;;
+        *)
+          echo "Invalid choice." >&2
+          exit 1
+          ;;
+      esac
+    fi
+    variant="$resolved"
+    ;;
+  jp|kr|sc|tc|hk)
+    ;;
+  *)
+    echo "Unknown variant: $variant" >&2
+    usage >&2
+    exit 1
+    ;;
+esac
+
+# --- Map variant to font family names ---
+
+upper=$(echo "$variant" | tr '[:lower:]' '[:upper:]')
+sans_family="Noto Sans CJK $upper"
+serif_family="Noto Serif CJK $upper"
+mono_family="Noto Sans Mono CJK $upper"
+
+# For the lang match, map variant back to BCP 47 language tag.
+case "$variant" in
+  jp) lang_tag="ja" ;;
+  kr) lang_tag="ko" ;;
+  sc) lang_tag="zh-cn" ;;
+  tc) lang_tag="zh-tw" ;;
+  hk) lang_tag="zh-hk" ;;
+esac
+
+# --- Check that the font is installed ---
+
+if ! fc-list | grep -Fqi -- "$sans_family"; then
+  echo "Font '$sans_family' not found. Install noto-fonts-cjk first." >&2
+  exit 1
+fi
+
+# --- Write the CJK section into fonts.conf ---
+#
+# fonts.conf is the only per-user fontconfig file this system loads, so the
+# CJK rules live here in a marker-delimited section. The monospace section
+# managed by `omarchy font set` is preserved.
+
+section_file=$(mktemp)
+cat >"$section_file" <<XML
+  <!-- omarchy:begin-cjk -->
+  <!-- omarchy-cjk-variant: $variant -->
+  <match target="pattern">
+    <test name="family" qual="any">
+      <string>sans-serif</string>
+    </test>
+    <edit name="family" mode="prepend" binding="strong">
+      <string>$sans_family</string>
+    </edit>
+  </match>
+  <match target="pattern">
+    <test name="family" qual="any">
+      <string>serif</string>
+    </test>
+    <edit name="family" mode="prepend" binding="strong">
+      <string>$serif_family</string>
+    </edit>
+  </match>
+  <match target="pattern">
+    <test name="family" qual="any">
+      <string>monospace</string>
+    </test>
+    <edit name="family" mode="prepend" binding="strong">
+      <string>$mono_family</string>
+    </edit>
+  </match>
+  <!-- Backup for Chromium/Electron: they resolve glyphs one character at a
+       time without lang on the pattern, so the rules above never fire.
+       The lang-targeted rule covers that path. -->
+  <match target="pattern">
+    <test name="lang" compare="contains">
+      <string>$lang_tag</string>
+    </test>
+    <edit name="family" mode="prepend" binding="strong">
+      <string>$sans_family</string>
+    </edit>
+  </match>
+  <!-- omarchy:end-cjk -->
+XML
+
+omarchy-fontconfig-replace-section cjk "$section_file"
+rm -f -- "$section_file"
+
+echo "CJK variant set to: $variant ($sans_family)"
+
+omarchy-restart-shell 2>/dev/null || true
+
+if pgrep -x foot; then
+  omarchy-notification-send -g "You must restart Foot to see CJK font change" 2>/dev/null || true
+fi
+
+if pgrep -x ghostty; then
+  omarchy-notification-send -g "You must restart Ghostty to see CJK font change" 2>/dev/null || true
+fi
+
+omarchy-hook font-set "cjk-$variant" 2>/dev/null || true
--- a/bin/omarchy-font-set
+++ b/bin/omarchy-font-set
@@ -45,15 +45,12 @@ if [[ -f ~/.config/foot/foot.ini ]]; then
 fi
 
 # fontconfig is the canonical source of truth — the omarchy shell, Qt apps,
-# and anything resolving "monospace" all read from here. This file is loaded
-# after the package-owned default, and prepend_first puts the chosen family at
-# the head of the list so it wins over the family that default prefers.
-fontconfig_file="$HOME/.config/fontconfig/fonts.conf"
-mkdir -p "$(dirname "$fontconfig_file")"
-cat >"$fontconfig_file" <<XML
-<?xml version="1.0"?>
-<!DOCTYPE fontconfig SYSTEM "fonts.dtd">
-<fontconfig>
+# and anything resolving "monospace" all read from here. The monospace choice
+# lives in a marker-delimited section of fonts.conf so the CJK section managed
+# by `omarchy font cjk set` survives a font change (and vice versa).
+section_file=$(mktemp)
+cat >"$section_file" <<XML
+  <!-- omarchy:begin-mono -->
   <match target="pattern">
     <test name="family" qual="any">
       <string>monospace</string>
@@ -62,8 +59,11 @@ cat >"$fontconfig_file" <<XML
       <string>$font_name</string>
     </edit>
   </match>
-</fontconfig>
+  <!-- omarchy:end-mono -->
 XML
+
+omarchy-fontconfig-replace-section mono "$section_file"
+rm -f -- "$section_file"
 
 omarchy-restart-shell
 
--- a/bin/omarchy-fontconfig-replace-section
+++ b/bin/omarchy-fontconfig-replace-section
@@ -1,0 +1,81 @@
+#!/bin/bash
+
+# omarchy:hidden=true
+# omarchy:summary=Replace a marker-delimited section in fonts.conf (internal helper)
+
+# Usage: omarchy-fontconfig-replace-section <name> <section-file>
+#
+# <name> is "mono" or "cjk". <section-file> holds the complete replacement
+# block, INCLUDING its begin/end marker lines:
+#   <!-- omarchy:begin-<name> --> ... <!-- omarchy:end-<name> -->
+# The other section (if any) is preserved byte-for-byte.
+#
+# fonts.conf is the only per-user fontconfig file this system loads
+# (~/.config/fontconfig/conf.d/ is not read), so both `font set` and
+# `font cjk set` cooperate through marker sections in this one file.
+#
+# A fonts.conf with no omarchy markers is treated as legacy (hand-written):
+# it is backed up once and replaced.
+
+if (( $# != 2 )); then
+  echo "Usage: omarchy-fontconfig-replace-section <name> <section-file>" >&2
+  exit 1
+fi
+
+name="$1"
+section_file="$2"
+fonts_conf="$HOME/.config/fontconfig/fonts.conf"
+begin="<!-- omarchy:begin-$name -->"
+end="<!-- omarchy:end-$name -->"
+
+if [[ ! -f $section_file ]]; then
+  echo "Section file '$section_file' not found." >&2
+  exit 1
+fi
+
+mkdir -p "$(dirname "$fonts_conf")"
+
+write_skeleton() {
+  {
+    echo '<?xml version="1.0"?>'
+    echo '<!DOCTYPE fontconfig SYSTEM "fonts.dtd">'
+    echo '<fontconfig>'
+    cat -- "$section_file"
+    echo '</fontconfig>'
+  } >"$fonts_conf"
+}
+
+if [[ ! -f $fonts_conf ]]; then
+  write_skeleton
+  exit 0
+fi
+
+if ! grep -q "omarchy:begin-" "$fonts_conf"; then
+  backup="$fonts_conf.bak-$(date +%Y%m%d%H%M%S)"
+  cp -- "$fonts_conf" "$backup"
+  echo "Backed up legacy fonts.conf to $backup"
+  write_skeleton
+  exit 0
+fi
+
+tmp=$(mktemp "$(dirname "$fonts_conf")/.fonts.conf.XXXXXX")
+# Marker lines may carry leading indentation; compare stripped.
+if grep -qF -- "$begin" "$fonts_conf"; then
+  awk -v begin="$begin" -v end="$end" -v section="$section_file" '
+    { stripped = $0; sub(/^[ \t]+/, "", stripped) }
+    stripped == begin { while ((getline line < section) > 0) print line; skip = 1; next }
+    stripped == end { skip = 0; next }
+    !skip { print }
+  ' "$fonts_conf" >"$tmp" && mv -- "$tmp" "$fonts_conf"
+else
+  if ! grep -q "</fontconfig>" "$fonts_conf"; then
+    echo "fonts.conf is malformed (no </fontconfig>); aborting." >&2
+    rm -f -- "$tmp"
+    exit 1
+  fi
+  awk -v section="$section_file" '
+    { stripped = $0; sub(/^[ \t]+/, "", stripped) }
+    stripped == "</fontconfig>" { while ((getline line < section) > 0) print line }
+    { print }
+  ' "$fonts_conf" >"$tmp" && mv -- "$tmp" "$fonts_conf"
+fi
--- a/default/agents/skills/omarchy/theming.md
+++ b/default/agents/skills/omarchy/theming.md
@@ -76,4 +76,7 @@ omarchy theme set catppuccin-custom
 omarchy font list               # Available fonts
 omarchy font current            # Current font
 omarchy font set <name>         # Change font
+omarchy font cjk list           # Available CJK variants
+omarchy font cjk current        # Current CJK variant
+omarchy font cjk set <variant>  # Change CJK variant (jp|kr|sc|tc|hk|auto)
 ```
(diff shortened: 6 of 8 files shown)


linyiru:
## zh-TW verification

You asked for native speakers to check `ko`, `zh_CN` and `zh_TW` rendering. I ran the `zh_TW` half.

**Rendering is correct — `tc` does exactly what it claims.** Along the way I hit four things that only
surface on a CJK locale, so they would not have appeared in your own testing on `en_US`. One of them
means you can drop a file from the PR entirely. Repro for everything below.

Host: Arch, Omarchy 4.0.3-1, `omarchy-settings 4.0.3-1`, `noto-fonts-cjk 20240730-1`,
`fontconfig 2.18.3`, `pango 1.58.2`, all five Noto CJK variants installed. Method: the five `bin/`
scripts from `feat/font-cjk` run against a sandbox `HOME`/`XDG_CONFIG_HOME`, with
`omarchy-restart-shell`, `omarchy-notification-send` and `omarchy-hook` stubbed, so the live session
was never touched. Every number below is from a real run.

### The problem it fixes is real

On stock quattro the `lang` tag has no effect whatsoever — all four CJK locales get the same chain,
and the first Han-covering font is Korean:

```
$ fc-match -s "sans-serif:lang=zh-tw" | grep -m1 CJK
NotoSansCJK-Regular.ttc: "Noto Sans CJK KR" "Regular"
$ fc-match -s "sans-serif:lang=ja"    → Noto Sans CJK KR
$ fc-match -s "sans-serif:lang=ko"    → Noto Sans CJK KR
$ fc-match -s "sans-serif:lang=zh-cn" → Noto Sans CJK KR
```

Noto Sans CJK TC is installed the entire time and simply never selected.

I rendered `骨 直 令 次 內 說` through the generic `sans-serif` alias with
`pango-view --language=zh-tw`, before and after. The PNGs differ, and they differ in the right
places: `骨` top-right component, `直` base, `令` final stroke, `次` left radical, `內` inner `人` vs
`入`, `說` upper `兌`. Before the change all six render in their Korean standard shapes, which to a
Taiwanese reader reads as a font substitution failure rather than a style choice. After
`omarchy font cjk set tc` they are correct Traditional Chinese forms.

The five variants each produce a distinct rendering with a distinct checksum, so the selection is
genuinely reaching the shaper rather than relabelling one face — `令` and `次` are the easiest tells.
Happy to send the rendered comparisons if useful.

Confirmed correct for `zh_TW`.

### What works

| Check | Result |
|---|---|
| `cjk current` with no `fonts.conf` | `unset` |
| `cjk list` | `jp kr sc tc hk` |
| `cjk set tc` → `cjk current` | `tc` |
| `sans-serif` / `serif` under `tc` | `Noto Sans CJK TC` / `Noto Serif CJK TC` |
| `cjk set jp` after `tc` | cjk section replaced, `mono` section preserved byte-for-byte |
| `omarchy font set` after `cjk set` | mono section written, cjk section and marker preserved |
| Generated XML | parses; `FC_DEBUG` reports 0 fontconfig errors or warnings |
| Legacy hand-written `fonts.conf` | backed up to `fonts.conf.bak-<ts>`, path printed |
| Second run on an omarchy-managed file | no second backup |
| `cjk set bogus` / no argument | usage, exit 1 |
| `auto` for `ja_JP` / `ko_KR` / `zh_CN` / `zh_SG` / `zh_TW` | `jp` / `kr` / `sc` / `sc` / `tc` |
| Chromium-style bare charset lookup | `LANG=zh_TW` + `:charset=4e00` → TC (the lang rule earns its keep) |

The marker machinery itself is sound. The two sections really do survive each other, in both orders.

### 1. `~/.config/fontconfig/conf.d/` *is* read — the whole helper can go

`omarchy-fontconfig-replace-section` and the change to `omarchy-font-set` both rest on this premise:

> `fonts.conf` is the only per-user fontconfig file this system loads
> (`~/.config/fontconfig/conf.d/` is not read)

That is not the case here. The `fontconfig` package ships
`/usr/share/fontconfig/conf.default/50-user.conf` and symlinks it into `/etc/fonts/conf.d/` from its
own pacman hook (`/usr/share/libalpm/hooks/40-fontconfig-config.hook`), so it is present on every
Arch install:

```xml
  <description>Load per-user customization files</description>
	<include ignore_missing="yes" prefix="xdg">fontconfig/conf.d</include>
	<include ignore_missing="yes" prefix="xdg">fontconfig/fonts.conf</include>
```

Verified empirically — a file dropped in `~/.config/fontconfig/conf.d/` with **no** `fonts.conf`
present takes effect:

```
$ cat ~/.config/fontconfig/conf.d/99-probe.conf   # prepend Noto Sans CJK TC to sans-serif
$ fc-match -s sans-serif | head -2
LiberationSans-Regular.ttf: "Liberation Sans" "Regular"
NotoSansCJK-Regular.ttc: "Noto Sans CJK TC" "Regular"
```

So `omarchy-font-cjk-set` can own `~/.config/fontconfig/conf.d/50-omarchy-cjk.conf` outright. That
drops all 81 lines of `omarchy-fontconfig-replace-section`, leaves `bin/omarchy-font-set` completely
unmodified, and removes three failure modes with it: no marker parsing, no malformed-file path, and
no reason to ever back up and overwrite a user's hand-written `fonts.conf` — theirs is left alone by
construction. `cjk current` becomes `[[ -f the-file ]]` plus the same `sed`, and `cjk unset` becomes
`rm -f`, which there is no clean way to offer today.

Note the include order: `conf.d` loads *before* `fonts.conf`, so the `prepend_first` monospace rule
`omarchy font set` writes keeps winning the head of the chain. That ordering is what makes issue 2
below go away for the case where the user has set a font — see the table.

### 2. The `lang` rule puts a proportional font at the head of the `monospace` chain

`omarchy-font-cjk-set:161-168` tests only `lang`, not `family`, so the rule also fires for
`monospace` patterns and prepends the **proportional** `Noto Sans CJK TC` — not
`Noto Sans Mono CJK TC`, which the rule three blocks above already handles correctly.

That lands on the shell. `shell/Commons/Style.qml:469` picks the shell font with
`fc-match -f "%{family[0]}" monospace`, and `Style.fontFamily` defaults to `"monospace"` precisely so
that "the bar and every qs.Ui component follows the fontconfig alias `omarchy-font-set` writes"
(`Style.qml:265-269`). So whatever that one query returns is what the bar, the menu and every
`qs.Ui` component draw in — Latin text included:

```
                                              LANG=en_US        LANG=zh_TW
stock quattro, no fonts.conf                  JetBrainsMono     JetBrainsMono
after `omarchy font set "JetBrainsMono …"`     JetBrainsMono     JetBrainsMono
after `omarchy font cjk set tc` only          JetBrainsMono     Noto Sans CJK TC   ←
`font set`, then `cjk set tc`                 JetBrainsMono     Noto Sans CJK TC   ←
`cjk set tc`, then `font set`                 JetBrainsMono     JetBrainsMono

$ fc-match -f '%{spacing}' "Noto Sans CJK TC"        →  (empty: proportional)
$ fc-match -f '%{spacing}' "Noto Sans Mono CJK TC"   →  100
$ fc-match -f '%{spacing}' "JetBrainsMono Nerd Font" →  100
```

Two things worth drawing out. It is order-dependent: `prepend_first` in the mono section only wins
when that section sits *after* the cjk section, and which is last depends on which command the user
ran last. And rows 3 and 4 include the default state — quattro ships no
`~/.config/fontconfig/fonts.conf` at all (`bin/omarchy-upgrade-to-quattro:1598` retires the
pre-quattro one, and `bin/omarchy-font-set` is the only thing in the repo that creates it), so a
user who has never changed their font is in the broken row.

**The fix is one word.** `50-omarchy.conf` already documents the right pattern for exactly this
situation, in its own comment about the Arabic last-resort rule: appending "keeps it behind every
family the app actually asked for, and charset matching outranks family matching, so a font that
does not cover the codepoint can never be pulled in." Changing the lang rule from
`mode="prepend"` to `mode="append"` does the same thing here. Measured, with the file in `conf.d`:

```
                              lang rule mode=  prepend            append
no fonts.conf, LANG=zh_TW      shell font      Noto Sans CJK TC   JetBrainsMono ✓
with fonts.conf, LANG=zh_TW    shell font      JetBrainsMono      JetBrainsMono ✓
either,          LANG=zh_TW    sans-serif      Noto Sans CJK TC   Noto Sans CJK TC ✓
either,          LANG=zh_TW    :charset=4e00   Noto Sans CJK TC   Noto Sans CJK TC ✓
either,          LANG=en_US    :charset=4e00   Noto Sans CJK JP   Noto Sans CJK JP
```

`append` costs nothing: the Chromium-style bare-charset path the rule exists for still lands on TC.
(Scoping the rule with a `spacing` test instead also works, but it takes two rules and does not match
anything already in the tree.)

### 3. A failed write is reported as success

`omarchy-fontconfig-replace-section` exits 1 on a malformed `fonts.conf` (line 74), but neither
caller checks the status. `omarchy-font-cjk-set:172` calls it and line 175 prints success
unconditionally:

```
$ omarchy-font-cjk-set tc
fonts.conf is malformed (no </fontconfig>); aborting.
CJK variant set to: tc (Noto Sans CJK TC)
$ echo $?
0
$ omarchy-font-cjk-current
unset
```

Nothing was written, `omarchy-hook font-set cjk-tc` still fired, and the exit status is 0.
`omarchy-font-set` behaves identically. `omarchy-fontconfig-replace-section … || exit 1` in both
callers covers it — or it disappears with issue 1.

### 4. `auto` cannot resolve on a stock install

Quattro generates only `C.UTF-8` and `en_US.UTF-8` — `/etc/locale.gen` has a single uncommented
line, `/etc/locale.conf` is `LANG=en_US.UTF-8` — and no Omarchy command generates another. So on any
unmodified install `auto` takes the `en_*` branch and drops into the interactive 1-5 prompt every
time; the `LANG` detection path is unreachable until locale selection exists, which is what #10949 is
about. Graceful, not broken, but worth a line in the usage text so `auto` does not read as detection
that silently never fires.

Two smaller things in `resolve_auto`:

- `zh_MO` falls through line 54 to `sc`. Macau uses Traditional Chinese; it belongs with `TW|HK`.
- `zh_HK` → `tc` (line 53) means `auto` never selects `hk`, even though `cjk list` advertises it and
  `Noto Sans CJK HK` is installed. Deliberate, going by the comment — but then `hk` is reachable only
  by typing it, and the usage block is the place to say so.

Bare `zh` → `sc` is right and matches CLDR likely-subtags.

### Minor

- `cjk set` prints raw PIDs, because `pgrep -x foot` / `pgrep -x ghostty` are used as conditions
  without redirecting stdout. Pre-existing style — `bin/omarchy-font-set` already does this at
  lines 76 and 80 — so not introduced here, but it is now on three code paths.
- One variant is global, so a `zh_TW` user reading Japanese gets TC kanji shapes
  (`sans-serif:lang=ja` → `Noto Sans CJK TC`). Same trade-off the upstream Noto config makes, and the
  right one for a single setting; noting it only so it is a known limit rather than a later bug
  report.

Happy to re-run any of this against a revision.

linyiru:
## Correction to my review above

Two things in my comment are wrong, both in the "the problem it fixes is real" section. The rest of the review stands. One of the corrections sharpens the case for this PR and also bounds it, so it is worth more than the errata.

### 1. `內` does not discriminate KR from TC — drop it from the diagnostic set

I listed `內` (inner `人` vs `入`) as one of six characters whose shape changes. It does not. KR and TC map it to the *same* glyph in the same `.ttc`, so nothing about it changes before or after `cjk set tc`:

```
$ F=/usr/share/fonts/noto-cjk/NotoSansCJK-Regular.ttc
$ for i in 1 3 2; do printf "idx%s: " $i; \
    hb-shape --font-file=$F --face-index=$i --no-clusters --no-positions --text="骨直令次內說"; done
idx1 (KR): [gid45132|gid27873|gid9808|gid62443|gid10896|gid37887]
idx3 (TC): [gid45134|gid27874|gid9810|gid22497|gid10896|gid37889]
idx2 (SC): [gid45133|gid27874|gid9809|gid22496|gid10897|gid37888]
                                               ^^^^^^^^ KR == TC, SC differs
```

Per-character pixel diff at 80 px confirms it (`magick compare -metric AE`, 22,225 px per cell):

| KR vs TC | 骨 | 直 | 令 | 次 | 內 | 說 | 產 | 雪 | 過 |
|---|---|---|---|---|---|---|---|---|---|
| differing px | 955 | 2194 | 2266 | 1101 | **0** | 1113 | 1488 | 1313 | 716 |

So `內` discriminates SC only. `產`, `雪` and `過` are all usable replacements for a KR-vs-TC argument — measurements above.

### 2. My method line is wrong, and reproducing it as written produces a null

I wrote that I rendered through the `sans-serif` alias "with `pango-view --language=zh-tw`". That is not what the script did: it set `LANG=zh_TW.UTF-8 LC_ALL=zh_TW.UTF-8` in the environment on a box where that locale had never been generated, so **no language reached the shaper at all**. The numbers remain valid for the state I meant to test — a stock install is precisely a box with no CJK locale — but the flag as written inverts the result, and anyone re-running it will conclude the feature does nothing:

```
$ for f in JP KR SC TC HK; do pango-view -q --language=zh-tw \
    --font="Noto Sans CJK $f 40" --text="骨直令次內說" -o /tmp/pv.png 2>/dev/null; \
    md5sum /tmp/pv.png; done
fcff50719717c054f71f138e9a00a27d      ← JP
fcff50719717c054f71f138e9a00a27d      ← KR   all five faces
fcff50719717c054f71f138e9a00a27d      ← SC   byte-identical
fcff50719717c054f71f138e9a00a27d      ← TC
fcff50719717c054f71f138e9a00a27d      ← HK

$ # identical loop, no --language
b11719068241be129b44035fdac3e254      ← JP
c78e2d8d73e7c06b821af5e33f74993f      ← KR
a3b90004fbc8e943b06cd08f0c6625dc      ← SC
fcff50719717c054f71f138e9a00a27d      ← TC   note: equals the --language=zh-tw hash
d6221b9d0b550875503f45c5d7d40e1e      ← HK
```

That is HarfBuzz's `locl` feature. Given a language of `zh-TW`, all five faces produce byte-identical Traditional Chinese output, and that output is exactly what the TC face produces with no language at all — the face stops mattering. Use `hb-shape` glyph ids as ground truth rather than pango PNG checksums; pango's bytes also move when only the primary font's metrics change, which is a second way to get a false positive out of that method.

### The part worth your attention

Read together, those two make a stronger case for this PR than my original review did, and they also bound it.

**Today the fix is load-bearing, and `locl` is why.** Nothing in the repo generates a locale — `grep -rn 'locale-gen\|localedef\|locale\.gen' .` returns zero hits; the only locale reference anywhere is `default/bash/envs` sourcing `/etc/locale.conf`. So a Taiwanese user has no `zh_TW.UTF-8` to select, no application can request `zh-TW`, `locl` never fires, and the Korean forms carried by the KR face are what they actually see. Your variant preference is the only lever that exists.

**But the lever's value is conditional on exactly that.** With the locale present versus absent, same face, same text:

```
$ env LANG=zh_TW.UTF-8 LOCPATH=$HOME/.local/locale \
    pango-view -q --font="Noto Sans CJK KR 40" --text="骨直令次內說" ...
    → byte-identical to the TC face

$ env LANG=zh_TW.UTF-8      # locale absent, silently falls back
    pango-view -q --font="Noto Sans CJK KR 40" --text="骨直令次內說" ...
    → Korean forms
```

For any surface that honours the session locale, the correct fix is the locale and the fontconfig preference is redundant. For surfaces that do not, the preference is the only fix. I have not measured which of Quickshell/Qt, Chromium and foot fall on which side, so I am not going to assert it — but it decides whether `cjk set` is a permanent setting or a bridge until locale selection (#10949) lands, and that is cheaper to settle before this merges than after. My issue 4 above is the same finding from the other end: `auto` cannot resolve on a stock install because the locale it detects does not exist.

Everything else in the review is unaffected and re-verified where relevant: `~/.config/fontconfig/conf.d/` is read, the `prepend` → `append` fix, the unchecked exit status, and the `zh_MO` / `zh_HK` notes.

usutani:
Thanks for the thorough verification. Addressed in `a71d175`:

1. Moved to `~/.config/fontconfig/conf.d/50-omarchy-cjk.conf` owned outright; dropped `omarchy-fontconfig-replace-section` (81 lines); `bin/omarchy-font-set` is now unmodified vs `quattro`; `cjk current`/`unset` use the conf.d path.
2. `lang` rule changed to `mode="append"`.
3. Gone with the helper — direct write, no unchecked exit path remains.
4. `zh_MO` → `tc`; usage notes that `auto` always prompts on stock installs and that `hk` is explicit-only.

`pgrep` silenced in `cjk-set`; left the pre-existing `font-set` occurrences alone to keep it unmodified. Could you re-run the `monospace`/`sans-serif` measurements against this revision?

linyiru:
## Re-ran against `a71d175`

Same box and method as before (Arch, `fontconfig 2:2.18.3-2`, `harfbuzz 14.4.0-1`, `noto-fonts-cjk 20240730-1`, `pango 1:1.58.2-1`, omarchy `4.0.4-1`, all five Noto CJK variants installed), the four `bin/` scripts run against a sandbox `HOME` so the live session is never touched. Issues 1, 2, 3 and 4 are all fixed, and issue 2's broken row is gone rather than papered over. One behaviour change is worth a section, and my conclusion there is that it needs no code.

### A method correction first, because it changes how you read `fc-match`

`fc-match monospace` and `fc-match sans-serif` cannot answer this question, and my own issue-2 table leaned on them. With the PR applied both still report the Latin font, because that is genuinely what heads the chain:

```
$ fc-match sans-serif   → Liberation Sans           (identical before and after `cjk set tc`)
$ fc-match monospace    → CaskaydiaMono Nerd Font   (identical)
```

Read alone, that says the feature does nothing. What changes is which face renders a Han codepoint, so the measurement has to walk the sorted chain and find the first face that actually covers one:

```bash
covers() {  # true only if this face has a real glyph for 骨
  s=$(hb-shape --font-file="$1" --face-index="${2:-0}" --no-clusters --no-positions --text=骨)
  [[ -n $s && $s != *".notdef"* && $s != *"gid0"* ]]
}
fc-match -s -f '%{file}|%{index}|%{family[0]}\n' "$pattern" | while IFS='|' read -r f i fam; do
  covers "$f" "$i" && { echo "$fam"; break; }
done
```

Note the `gid0` test. An uncovered character does not always come back as `.notdef`: `NotoColorEmoji.ttf` returns `[gid0]` for 骨, so a `.notdef`-only predicate names Noto Color Emoji as the covering face and everything downstream is garbage. That cost me a pass.

### Results

First Han-covering face and its position in the sorted chain. Fixture: `fonts.conf` exactly as `omarchy font set` writes it, user mono font `CaskaydiaMono Nerd Font`.

| pattern | baseline, no preference | `cjk set tc` |
|---|---|---|
| `sans-serif` | Noto Sans CJK KR (pos 7) | **Noto Sans CJK TC (pos 2)** |
| `serif` | Noto Serif CJK KR (pos 6) | **Noto Serif CJK TC (pos 3)** |
| `monospace` | Noto Sans Mono CJK KR (pos 6) | **Noto Sans Mono CJK TC (pos 3)** |
| `:charset=4e00:lang=zh-tw` | Noto Sans CJK JP (pos 1) | **Noto Sans CJK TC (pos 1)** |
| head of those three chains | Liberation Sans / Liberation Serif / CaskaydiaMono | all three unchanged |

All five variants, all three generics, and the restore:

```
state      sans-serif           serif                monospace
set jp     Noto Sans CJK JP     Noto Serif CJK JP    Noto Sans Mono CJK JP
set kr     Noto Sans CJK KR     Noto Serif CJK KR    Noto Sans Mono CJK KR
set sc     Noto Sans CJK SC     Noto Serif CJK SC    Noto Sans Mono CJK SC
set tc     Noto Sans CJK TC     Noto Serif CJK TC    Noto Sans Mono CJK TC
set hk     Noto Sans CJK HK     Noto Serif CJK HK    Noto Sans Mono CJK HK
unset      Noto Sans CJK KR     Noto Serif CJK KR    Noto Sans Mono CJK KR   (baseline, exactly)
```

Sans, serif and mono all work, every variant is reachable, `unset` is clean, and row 4 of the table is the Chromium bare-charset path still earning the appended `lang` rule its keep (TC where baseline gives JP).

**Issue 2 specifically.** The row that was broken was the default state, no `~/.config/fontconfig/fonts.conf` at all, which is what any user who has never run `omarchy font set` has. The shell's own query is `fc-match -f '%{family[0]}' monospace` at `shell/Commons/Style.qml:469`:

```
                                    monospace resolves to
no cjk preference                   JetBrainsMono Nerd Font (spacing=100)
cjk set tc, no fonts.conf           JetBrainsMono Nerd Font (spacing=100)   ← was Noto Sans CJK TC, proportional
cjk set jp, no fonts.conf           JetBrainsMono Nerd Font (spacing=100)
cjk set tc, LANG=zh_TW.UTF-8        JetBrainsMono Nerd Font (spacing=100)
```

Both orders are fine too (`font set` then `cjk set`, and `cjk set` then `font set`), which now follows structurally from the two commands owning separate files instead of racing inside one. The order dependence I complained about is gone.

### The one behaviour change, and why I think it needs no code

The three family rules carry no `lang` test, so with `tc` selected, explicitly tagged Japanese, Korean and Simplified text also resolves to the TC face:

```
sans-serif:lang=ja      baseline Noto Sans CJK KR        →  cjk set tc  Noto Sans CJK TC
sans-serif:lang=ko      baseline Noto Sans CJK KR        →  cjk set tc  Noto Sans CJK TC
sans-serif:lang=zh-cn   baseline Noto Sans CJK KR        →  cjk set tc  Noto Sans CJK TC
monospace:lang=ja       baseline Noto Sans Mono CJK KR   →  cjk set tc  Noto Sans Mono CJK TC
```

I went into this expecting to hand you a patch. Then I measured what the face choice is worth once a language tag is present, and the answer is nothing:

```
$ TTC=/usr/share/fonts/noto-cjk/NotoSansCJK-Regular.ttc
$ for i in 0 1 2 3 4; do hb-shape --font-file=$TTC --face-index=$i --language=ja \
    --no-clusters --no-positions --text=骨直令次說產; done
[gid45132|gid27873|gid9808|gid22495|gid37887|gid27042]   ← JP face
[gid45132|gid27873|gid9808|gid22495|gid37887|gid27042]   ← KR face
[gid45132|gid27873|gid9808|gid22495|gid37887|gid27042]   ← SC face
[gid45132|gid27873|gid9808|gid22495|gid37887|gid27042]   ← TC face
[gid45132|gid27873|gid9808|gid22495|gid37887|gid27042]   ← HK face
```

All five faces produce identical Japanese forms, and the same holds for `--language=ko`, `zh-TW` and `zh-CN`. This is `locl` again, the same mechanism as in my earlier correction. A surface that put `lang=ja` on the fontconfig pattern is a surface that hands the language to the shaper as well (Pango does both), so it gets JP forms out of the TC face. Baseline's KR face was equally not-Japanese and equally corrected by `locl`, so this is one wrong-on-paper face traded for another with no rendered difference.

My recommendation is therefore to leave it as pushed. For completeness, what I tried and what it would cost:

- Nine lang-tested rules (ja/ko/zh-cn × sans/serif/mono) placed **after** the existing block is a no-op. Both rules prepend relative to the same matched `sans-serif` element, so the unconditional TC stays in front of the language's own face. I measured that before I understood it.
- The same nine rules placed **before** the block do work: `lang=ja` → JP, `lang=ko` → KR, `lang=zh-cn` → SC, `lang=zh-tw` and `lang=zh-hk` → TC, untagged → TC, Latin heads unchanged, charset path unchanged, `xmllint` clean, zero fontconfig warnings under `FC_DEBUG`.
- But they move territory-less `lang=zh` from the user's TC to SC, because `compare="contains"` treats a bare tag as contained by a territory-qualified one. That is defensible by CLDR likely-subtags, and it is what I argued for in `resolve_auto`, but it is not what someone who typed `cjk set tc` asked for.

Nine rules plus a new edge case, to fix something `locl` already fixes. Not worth it. The honest bound on that conclusion: it holds because Noto CJK is a single `.ttc` with `locl` coverage across all five faces, so an application that tags the fontconfig pattern but not the shaper run would still show the chosen variant's forms. I could not find one to name, and if you know of one this trade flips.

### Two small things

- `omarchy-font-cjk-current:16` derives the variant from the `<!-- omarchy-cjk-variant: -->` comment. Delete that one line by hand and `current` reports `unset` while the rules stay live and keep working. Now that the file is owned outright, reading the family out of the sans rule instead (`sed -n 's/.*Noto Sans CJK \([A-Z]*\).*/\1/p'`, lowercased) is a source of truth that cannot drift from what fontconfig reads. `unset` is unaffected, it tests `-f`.
- `cjk set TC` exits 1 with `Unknown variant: TC`. The usage block only advertises lowercase, so this is defensible; a `variant="${variant,,}"` before the validating `case` is the one-line version if you want it.

And a note so nobody re-raises it from the diff view: `git diff quattro <pr-head> -- bin/omarchy-font-set` does show changes, but that is quattro moving underneath the branch in `4cb9c75a` ("Move Kitty defaults into the system config"). Against the merge base `f99d33a8` the file is untouched, so your claim holds.

`pgrep` output is silenced, `zh_MO` → `tc`, and the usage block now says both of the things it needed to.

usutani:
Both small things addressed in `a7d4681`:

1. `cjk-current` now reads the variant from the sans rule family (`Noto Sans CJK TC` → `tc`) instead of the marker comment, so it cannot drift from what fontconfig reads. Verified: deleting the comment line by hand still reports `tc`. The `unset` path is untouched (still `-f`). The writer still emits the comment for human readability.
2. `cjk set` lowercases input first (`variant="${variant,,}"`), so `TC` and `Auto` work; unknown input (e.g. `BOGUS`) still exits 1 after normalization. Lowercase behavior unchanged.

`./test/cli`: 116 ok / 0 fail.

PR #9978, opened 2026-09-03
Load personal hypr.envs after package Hyprland defaults

### LINKED ISSUE #9902: 4.0.x: user ~/.config/hypr/envs.lua is never loaded — user env overrides (and the updater's own iHD fix) are dead code
### System details

Intel UHD Graphics (Raptor Lake-P, iGPU-primary) + NVIDIA RTX 4050 hybrid laptop, Omarchy 4.0.1, Hyprland / Wayland

### What's wrong?

`~/.config/hypr/envs.lua` — the personal counterpart of `default/hypr/envs.lua` — is never loaded. Nothing ever `require`s it, so env vars a user sets there are silently ignored.

**Evidence**

1. Neither the migrated user config, the shipped template (`config/hypr/hyprland.lua`), nor current `main` requires `hypr.envs`. The personal requires are only `hypr.monitors`, `hypr.input`, `hypr.bindings`, `hypr.looknfeel`, `hypr.autostart` — and the comment block above them ("Put your personal overrides in these files") does not mention envs either.
2. `grep -rn 'hypr\.envs' /usr/share/omarchy ~/.config/hypr` finds only `default.hypr.envs` (`omarchy.lua:16`). No reference to the user module.
3. `default/hypr/bootstrap.lua` only adds `~/.config/?.lua` to `package.path` (making the module *requireable*); it does not load it. `require_all` is used only for apps, bindings, toggles, and workspace-layouts — never for `~/.config/hypr`.
4. Meanwhile the defaults do set hardware envs: `omarchy.lua` → `default.hypr.envs` → `default.hypr.nvidia` → on a hybrid laptop with GSP, `LIBVA_DRIVER_NAME=nvidia` is what actually lands in the session.

**Concrete failure on this hybrid laptop**

An Omarchy update (Aug 29, 4.0.x) itself rewrote the user's `envs.lua` from `LIBVA_DRIVER_NAME=nvidia` to `iHD`, leaving a timestamped `.bak` and this comment:

```lua
-- Video decode (VAAPI) forced to Intel iGPU: the panel is wired to Intel,
-- so decoding on NVIDIA required an extra cross-GPU frame copy that caused
-- browser video glitches. NVIDIA still handles OpenGL/Vulkan (games) above.
```

But since nothing loads `envs.lua`, the updater's own fix never takes effect. Browsers inherit no `LIBVA_DRIVER_NAME`, libva auto-detects a driver per render node, selects the NVDEC VAAPI shim on the dGPU node, and hardware-decoded video in Chromium/Brave paints a checkerboard (or never paints at all).

**Steps to reproduce**

1. Omarchy 4.0.1 on an iGPU-primary Intel+NVIDIA hybrid laptop.
2. Set `hl.env("LIBVA_DRIVER_NAME", "iHD")` in `~/.config/hypr/envs.lua`, relogin.
3. `systemctl --user show-environment | grep LIBVA` → variable not applied; browser video still broken.
4. Launch a browser directly with `LIBVA_DRIVER_NAME=iHD brave` → video plays fine (confirming the env is the whole fix).
5. Add `require("hypr.envs")` to the personal block of `hyprland.lua`, relogin → `LIBVA_DRIVER_NAME=iHD` is exported and video works without the manual env.

**Regression note:** in 3.x the personal env file *was* loaded — `hyprland.conf` sourced `~/.config/hypr/envs.conf` in the personal overrides section (see the load order in #6068). The 3.x → 4.x Lua migration dropped it from the personal require list, while the updater still patches the file as if it were live.

**Suggested fix:** add `require_optional.module("hypr.envs")` (or `require("hypr.envs")`) alongside the other personal requires in `config/hypr/hyprland.lua`, and have the updater insert that require into migrated `hyprland.lua` files that predate 4.0.

Related but distinct: #8989, #9890, #9096, #8328, #8215, #9708 report the video corruption caused by `nvidia.lua` forcing NVIDIA VA-API on hybrid laptops. This report is about the user env override channel itself being dead code — which also silently disables the updater's own `envs.lua` fix.

fresh3nough:
## Summary

`~/.config/hypr/envs.lua` was never required. The 3.x → 4.x Lua migration dropped the personal envs load while the updater still patched that file, so user `hl.env()` overrides (and the updater's own iHD fix) never reached the session.

## Fix

- `require_optional.module("hypr.envs")` in `config/hypr/hyprland.lua` after package defaults
- Shipped `config/hypr/envs.lua` template
- Migration `1788439000.sh` inserts the require into existing entrypoints (idempotent)

Fixes #9902

## Test plan

- [x] Confirmed no `hypr.envs` require on quattro entrypoint before the change
- [x] `bash test/shell.d/hypr-envs-load-test.sh` — entrypoint load order, runtime evaluation of personal envs, migration insert + idempotency

### DIFF
--- a/config/hypr/envs.lua
+++ b/config/hypr/envs.lua
@@ -1,0 +1,5 @@
+-- Personal environment overrides.
+-- Loaded after Omarchy's default.hypr.envs so these hl.env() calls win.
+--
+-- Example:
+-- hl.env("LIBVA_DRIVER_NAME", "iHD")
--- a/config/hypr/hyprland.lua
+++ b/config/hypr/hyprland.lua
@@ -21,6 +21,10 @@ require("hypr.input")
 require("hypr.bindings")
 require("hypr.looknfeel")
 require("hypr.autostart")
+-- Personal env overrides. After package defaults so user hl.env() wins. Optional
+-- so installs without ~/.config/hypr/envs.lua are unchanged (#9902).
+local require_optional = require("default.hypr.require_optional")
+require_optional.module("hypr.envs")
 
 -- Toggle config flags dynamically.
 require("default.hypr.toggles")
--- a/migrations/1788439000.sh
+++ b/migrations/1788439000.sh
@@ -1,0 +1,34 @@
+echo "Load personal hypr.envs from the Hyprland entrypoint"
+
+# 4.x dropped the personal envs require that 3.x had via envs.conf. The updater
+# still patches ~/.config/hypr/envs.lua, but nothing loaded it (#9902).
+
+hyprland_config="$HOME/.config/hypr/hyprland.lua"
+[[ -f $hyprland_config ]] || exit 0
+
+if grep -Eq 'require(_optional\.module)?\(["'\'']hypr\.envs["'\'']\)|require_optional\.module\(["'\'']hypr\.envs["'\'']\)' "$hyprland_config"; then
+  exit 0
+fi
+
+tmp=$(mktemp)
+awk '
+  BEGIN { inserted = 0 }
+  {
+    print
+    if (!inserted && $0 ~ /^require\("hypr\.autostart"\)/) {
+      print "-- Personal env overrides (after package defaults so user hl.env wins)."
+      print "local require_optional = require(\"default.hypr.require_optional\")"
+      print "require_optional.module(\"hypr.envs\")"
+      inserted = 1
+    }
+  }
+  END {
+    if (!inserted) {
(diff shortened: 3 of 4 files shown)


sprajs:
#10024 addresses the same #9902 missing `hypr.envs` load with a different integration point: it loads the optional module from package-owned `default.hypr.envs` after the NVIDIA defaults, whereas this PR updates and migrates the user entrypoint.

Could the two be linked and one loading strategy selected before merging? The package-level approach can reach existing users without patching their customized entrypoint; the load-order and missing-file tests from both are useful to preserve when consolidating.

PR #9979, opened 2026-09-03
Enable Sunshine by its real systemd user unit name

### LINKED ISSUE #9710: omarchy install service sunshine aborts before firewall/webapp/autostart: systemctl --user enable --now sunshine cannot resolve the Alias= on a fresh install
### What happened

On a clean Omarchy install, `omarchy install service sunshine` installs the
package and then aborts. Sunshine ends up installed but **not enabled, not
started, with no firewall rules, no admin web app, and no autostart entry**.
The only visible output is:

```
Installing Sunshine...
```

followed by the failure of the `systemctl` call.

### Root cause

`omarchy-install-service-sunshine` runs:

```bash
```

But the `sunshine` package ships its user unit as
`/usr/lib/systemd/user/app-dev.lizardbyte.app.Sunshine.service`, whose
`[Install]` section declares:

```ini
[Install]
Alias=sunshine.service
```

A systemd `Alias=` is only materialised as a symlink **when the unit is
enabled**. On a fresh install the unit has never been enabled, so
`sunshine.service` does not resolve yet and the enable call fails:

```
$ systemctl --user status sunshine
Unit sunshine.service could not be found.

$ systemctl --user list-unit-files | grep -i sunshine
app-dev.lizardbyte.app.Sunshine.service    disabled  enabled
```

Because the script is `set -e`, it exits at that point, so
`open_ufw_ports`, `install_admin_webapp` and `enable_hyprland_autostart`
never run.

### Why this is easy to miss when reproducing

Enabling the unit by its real name creates the alias symlink:

```
/home/<user>/.config/systemd/user/sunshine.service
  -> /usr/lib/systemd/user/app-dev.lizardbyte.app.Sunshine.service
```

After that, `systemctl --user enable --now sunshine` resolves fine and a
re-run of the installer succeeds. **The bug only reproduces on a machine
where the unit has never been enabled.**

### Steps to reproduce

1. Fresh Omarchy install with no prior Sunshine
2. `omarchy install service sunshine`
3. Observe the script stops after "Installing Sunshine..."
4. `systemctl --user is-enabled sunshine` -> `not-found`
5. `sudo ufw status | grep omarchy-sunshine` -> no rules

### Expected

Sunshine enabled and running, Moonlight ports opened for private LANs and
`tailscale0`, admin web app installed, autostart configured.

### Suggested fix

Enable the unit by the name the package actually ships:

```bash
```

Worth noting for the autostart step: that unit is already
`WantedBy=graphical-session.target`, so it starts with the session on its
own. Adding `o.launch_on_start("sunshine")` to `autostart.lua` on top of the
enabled unit launches a second instance that contends for the same ports —
so `enable_hyprland_autostart` may be redundant once the unit is enabled
correctly.

### System details

- Omarchy 4.0.2-1
- `sunshine` 2026.516.143833-4
- CPU: Intel Core i7-6700
- GPU: Intel HD Graphics 530 + NVIDIA Quadro K1200
- Hyprland 0.56.2, session via uwsm, SDDM autologin

### LINKED ISSUE #9775: omarchy install service sunshine fails: enabling by Alias= before the unit is enabled can never resolve
### What happened

`omarchy install service sunshine` aborts partway through on a fresh install:

```
Installing Sunshine...
...
(4/4) installing sunshine                                    [====] 100%
```

The package installs fine, but the script exits at that point and never completes.

### Root cause

`bin/omarchy-install-service-sunshine` line 74 runs:

```bash
```

The `omarchy/sunshine` package does not ship a `sunshine.service`. It ships:

```
/usr/lib/systemd/user/app-dev.lizardbyte.app.Sunshine.service
```

That unit declares `Alias=sunshine.service` in its `[Install]` section — but a
`systemd` `Alias=` is only materialised as a symlink **when the unit is enabled**.
So on a fresh install `sunshine.service` does not yet exist, and enabling *by the
alias* is a chicken-and-egg that can never succeed.

Enabling by the real unit name works and creates the alias, after which
`sunshine.service` resolves normally:

```
Created symlink '~/.config/systemd/user/sunshine.service' → '/usr/lib/systemd/user/app-dev.lizardbyte.app.Sunshine.service'
Created symlink '~/.config/systemd/user/graphical-session.target.wants/app-dev.lizardbyte.app.Sunshine.service' → ...

$ systemctl --user is-active sunshine.service
active
```

### Impact

The script has `set -e`, so the failed `enable` aborts everything after line 74:

- Sunshine is **not** enabled or started
- UFW streaming ports (47984/47989/48010 TCP, 5353/47998/47999/48000/48002/48010 UDP) are **not** opened for private LANs or `tailscale0`
- The "Sunshine Admin" web app is **not** installed
- `o.launch_on_start("sunshine")` is **not** added to `~/.config/hypr/autostart.lua`

The user is left with a Sunshine package installed but entirely non-functional, and
the only visible symptom is the one-line `Failed to enable unit` message.

### Steps to reproduce

1. On a system without Sunshine installed, run `omarchy install service sunshine`
2. Observe `Failed to enable unit: Unit sunshine.service does not exist`
3. `systemctl --user is-enabled app-dev.lizardbyte.app.Sunshine.service` → `disabled`
4. `ls ~/.local/share/applications | grep -i sunshine` → nothing
5. `tail ~/.config/hypr/autostart.lua` → no `launch_on_start` entry

### Expected

The install completes: the service is enabled and running, ports are opened, and the
admin web app is installed.

### Suggested fix

Enable the unit by the name the package actually ships:

```bash
```

Worth noting: that unit is already `WantedBy=graphical-session.target`, so it starts
with the session on its own. The separate `o.launch_on_start("sunshine")` autostart
entry the script adds will spawn a second instance competing for the same ports —
possibly worth dropping as part of the same fix.

### System details

| | |
|---|---|
| Omarchy | 4.0.2-1 |
| sunshine | 2026.516.143833-4 (omarchy repo) |
| systemd | 261.2-1 |
| Kernel | 7.1.9-arch1-2 |
| CPU | Intel Core Ultra 9 285K |
| GPU | Intel Arrow Lake-S (integrated) |

fresh3nough:
## Summary

`omarchy install service sunshine` ran `systemctl --user enable --now sunshine`. The package only ships `app-dev.lizardbyte.app.Sunshine.service` with `Alias=sunshine.service` in `[Install]`. That alias is created on enable, so enabling by the alias on a fresh install fails with `Unit sunshine.service does not exist` and `set -e` aborts before firewall, webapp, and autostart.

## Fix

Enable/disable the real unit name. Remove still falls back to the alias if present.

Fixes #9775  
Fixes #9710

## Test plan

- [x] Reproduced: enable by `sunshine` fails; enable by package unit succeeds
- [x] `bash test/shell.d/sunshine-service-test.sh` — install stubs enable real unit, does not use alias, completes autostart

### DIFF
--- a/bin/omarchy-install-service-sunshine
+++ b/bin/omarchy-install-service-sunshine
@@ -13,9 +13,11 @@ SUNSHINE_ADMIN_APP="Sunshine Admin"
 SUNSHINE_ADMIN_URL="https://localhost:47990"
 SUNSHINE_ADMIN_EXEC="omarchy-launch-webapp $SUNSHINE_ADMIN_URL --ignore-certificate-errors"
 SUNSHINE_ICON_SOURCE="/usr/share/sunshine/web/images/logo-sunshine-45.png"
+# Package unit name. Alias=sunshine.service only appears after enable, so the
+# first enable must use this name rather than the alias (#9775 / #9710).
+SUNSHINE_UNIT="app-dev.lizardbyte.app.Sunshine.service"
 HYPR_AUTOSTART_FILE="$HOME/.config/hypr/autostart.lua"
 HYPR_AUTOSTART_ENTRY='o.launch_on_start("sunshine")'
-
 open_ufw_port_for_private_lans() {
   local proto="$1"
   local port="$2"
@@ -71,8 +73,7 @@ enable_hyprland_autostart() {
 
 echo "Installing Sunshine..."
 omarchy-pkg-add sunshine
-systemctl --user enable --now sunshine
-
+systemctl --user enable --now "$SUNSHINE_UNIT"
 echo "Opening Sunshine firewall ports..."
 open_ufw_ports
 
--- a/bin/omarchy-remove-service-sunshine
+++ b/bin/omarchy-remove-service-sunshine
@@ -59,9 +59,10 @@ disable_hyprland_autostart() {
   fi
 }
 
+# Prefer the real unit; also clear the alias name if only that remains.
+systemctl --user disable --now app-dev.lizardbyte.app.Sunshine.service 2>/dev/null || true
 systemctl --user disable --now sunshine 2>/dev/null || true
-omarchy-pkg-drop sunshine
-omarchy-webapp-remove "$SUNSHINE_ADMIN_APP" 2>/dev/null || true
+omarchy-pkg-drop sunshineomarchy-webapp-remove "$SUNSHINE_ADMIN_APP" 2>/dev/null || true
 disable_hyprland_autostart
 close_ufw_ports
 
--- a/test/shell.d/sunshine-service-test.sh
+++ b/test/shell.d/sunshine-service-test.sh
@@ -1,0 +1,89 @@
+#!/bin/bash
+
+set -euo pipefail
+
+source "$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd)/base-test.sh"
+
+test_tmp=$(mktemp -d)
+trap 'rm -rf "$test_tmp"' EXIT
+
+stub_bin="$test_tmp/bin"
+home="$test_tmp/home"
+log="$test_tmp/systemctl.log"
+mkdir -p "$stub_bin" "$home/.config/hypr"
+
+# Fresh install: sunshine.service alias does not exist until the real unit is
+# enabled. Enabling by the alias must fail; enabling by the package unit works.
+cat >"$stub_bin/systemctl" <<'SH'
+#!/bin/bash
+printf '%s\n' "$*" >>"$TEST_SYSTEMCTL_LOG"
+if [[ ${1:-} == "--user" && ${2:-} == "enable" && ${3:-} == "--now" ]]; then
+  unit=${4:-}
+  case "$unit" in
+    app-dev.lizardbyte.app.Sunshine.service)
+      exit 0
+      ;;
+    sunshine|sunshine.service)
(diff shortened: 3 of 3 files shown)


sprajs:
The removal hunk accidentally joins two commands into `omarchy-pkg-drop sunshineomarchy-webapp-remove "$SUNSHINE_ADMIN_APP" ... || true`. That passes the concatenated name to the package remover and never invokes `omarchy-webapp-remove`; the trailing `|| true` then hides the failure. Please restore the two separate commands and assert both package and webapp removal in the test.

Related #11352 covers the same canonical-unit/alias fix, including both names during removal. It would be worth consolidating these implementations rather than maintaining two fixes for the same install failure.


Reproduced on `9d1d462e` with isolated command stubs and a temporary HOME: the parent calls `omarchy-pkg-drop sunshine` and `omarchy-webapp-remove "Sunshine Admin"`; this head instead calls only `omarchy-pkg-drop sunshineomarchy-webapp-remove "Sunshine Admin"`. Even when the package stub rejects that argument, the script exits 0 and prints “Sunshine has been removed”. A removal regression test should assert both exact helper calls as well as the exit status; the current install-only test does not cover this path. No packages, firewall rules, or live services were changed.

bjarneo:
Automated duplication check: this pull request looks similar to #8879, which covers the same Sunshine canonical unit fix. I keep that one open and close this one to consolidate review. If you feel this is the wrong decision, please open the PR again with a note on the difference.

PR #9980, opened 2026-09-03
Survive an unset OMARCHY_PATH in the update pipeline

### LINKED ISSUE #8369: omarchy-update-dev crashes with "unbound variable" on $OMARCHY_PATH under sudo
### What happened
`sudo omarchy update` fails partway through with:

    /usr/bin/omarchy-update-dev: line 7: OMARCHY_PATH: unbound variable

### Steps to reproduce
1. Run `sudo omarchy update` on a standard (non-dev-checkout) Omarchy install.
2. It fails during the dev-checkout-update step.

### Cause
`omarchy-update-dev` runs under `set -euo pipefail` and references
`$OMARCHY_PATH` directly on line 7:

    [[ $OMARCHY_PATH != "/usr/share/omarchy" ]] || exit 0

without a fallback default. Every other Omarchy script sets one
(`: "${OMARCHY_PATH:=/usr/share/omarchy}"` — see e.g.
default/bash/rc), but this one doesn't. Since sudo's default
`env_reset` strips OMARCHY_PATH from the environment before the
update script runs as root, the variable is unset when
omarchy-update-dev (invoked from omarchy-update) checks it, and
`set -u` turns that into a fatal error.

### Expected
Same fallback default as the rest of the codebase, e.g.:

    : "${OMARCHY_PATH:=/usr/share/omarchy}"

added near the top of omarchy-update-dev (or OMARCHY_PATH exported
explicitly by omarchy-update before invoking it).

### Workaround
    sudo OMARCHY_PATH=/usr/share/omarchy omarchy update

### System
- Omarchy 4.0.1-1
- Kernel 7.1.9-arch1-2
- Dell Latitude 7450, Intel Core Ultra 7 165U
- sudoers: default `env_reset` (no OMARCHY_PATH in env_keep)

### LINKED ISSUE #9953: sudo omarchy update aborts with "OMARCHY_PATH: unbound variable" in omarchy-update-dev

omarchybot:
`sudo omarchy update` aborts right after the system snapshot with `/usr/bin/omarchy-update-dev: line 7: OMARCHY_PATH: unbound variable`, before any package is upgraded.

`OMARCHY_PATH` is established by `default/bash/env-bootstrap`, which runs only from a login shell, an interactive rc, or the uwsm session environment. `sudo` execs the router directly under its default `env_reset`, and `omarchy-update:12` re-execs itself through `script -qefc`, so none of those run and the variable is unset. `omarchy-update-dev:5,7` is `set -euo pipefail` with a bare `$OMARCHY_PATH`, so `set -u` kills it and the ERR trap at `omarchy-update:19` aborts the whole update. `omarchy-update-available:5,9` carries the identical defect and is reached from the same run through `omarchy-update-status`; it is quieter only because its caller puts it in an `if` condition, where the death reads as "no updates".

Both now fall back to `/usr/share/omarchy`, which is the value `env-bootstrap:13-14` itself picks when no dev-link config is present, and which `omarchy-migrate:31` already defaults to one step later in this same pipeline. A dev-linked checkout is not pulled under `sudo`: `omarchy-dev-link` grants sudo a `secure_path` and no `env_keep`, and pulling a user-owned checkout as root would leave root-owned objects behind.

This stops the abort. It does not make a root-run update correct, and it is not meant to: migrations, hooks, AUR and mise updates read root's `$HOME`, `omarchy-restart-shell:9` resolves an empty path, and `omarchy-shell -q` turns the missing variable into quiet success. Whether `sudo omarchy update` should be supported at all, refused with a message, or made to drop back to the invoking user is a separate decision, and this change does not settle it either way.

Fixes #8369
Fixes #9953

### DIFF
--- a/bin/omarchy-update-available
+++ b/bin/omarchy-update-available
@@ -4,6 +4,10 @@
 
 set -euo pipefail
 
+# Reached from `sudo omarchy update` through omarchy-update-status, where sudo's
+# env_reset has already dropped the session's OMARCHY_PATH.
+OMARCHY_PATH="${OMARCHY_PATH:-/usr/share/omarchy}"
+
 updates=()
 
 if [[ $OMARCHY_PATH != "/usr/share/omarchy" ]]; then
--- a/bin/omarchy-update-dev
+++ b/bin/omarchy-update-dev
@@ -4,6 +4,10 @@
 
 set -euo pipefail
 
+# `sudo omarchy update` runs no shell that sources default/bash/env-bootstrap,
+# and sudo's env_reset has already dropped the session's copy.
+OMARCHY_PATH="${OMARCHY_PATH:-/usr/share/omarchy}"
+
 [[ $OMARCHY_PATH != "/usr/share/omarchy" ]] || exit 0
 
 if ! git -C "$OMARCHY_PATH" rev-parse --is-inside-work-tree >/dev/null 2>&1; then
--- a/test/shell.d/update-available-test.sh
+++ b/test/shell.d/update-available-test.sh
@@ -211,3 +211,17 @@ grep -Fx 'omarchy-dev-checkout 1 new commit on origin/quattro' "$stdout" >/dev/n
   fail "update checker reports cached dev commits after a fetch failure" "$(cat "$stdout")"
 [[ ! -s $stderr ]] || fail "update checker keeps dev fetch failures quiet" "$(cat "$stderr")"
 pass "update checker uses cached dev state when fetching is unavailable"
+
+: >"$git_log"
+if TEST_CHECKUPDATES=updates TEST_INSTALLED_PACKAGE=omarchy TEST_GIT_LOG="$git_log" \
+  PATH="$stub_bin:$PATH" env -u OMARCHY_PATH \
+  "$ROOT/bin/omarchy-update-available" >"$stdout" 2>"$stderr"; then
+  status=0
+else
+  status=$?
+fi
+[[ ! -s $stderr ]] || fail "sudo's stripped environment does not break the update checker" "$(cat "$stderr")"
+[[ $status -eq 0 ]] || fail "update checker still reports package updates without OMARCHY_PATH"
+grep -q '^omarchy ' "$stdout" || fail "update checker still reports package updates without OMARCHY_PATH" "$(cat "$stdout")"
+[[ ! -s $git_log ]] || fail "an unset OMARCHY_PATH checks no dev checkout" "$(cat "$git_log")"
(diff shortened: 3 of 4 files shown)


bjarneo:
Automated duplication check: this pull request looks similar to #8590, which covers the same OMARCHY_PATH default in update-dev. I keep that one open and close this one to consolidate review.

PR #9981, opened 2026-09-03
Run executable hooks via shebang instead of always bash

### LINKED ISSUE #9845: omarchy-hook always runs .d hook scripts with bash, breaking non-bash hooks despite correct shebang + exec bit
**What happened**
A `theme-set.d` hook script written in Python (`#!/usr/bin/env python3` shebang, executable via `chmod +x`) fails every time a theme is applied. The errors are bash trying to parse Python syntax as shell commands — e.g. `import sys` resolves to ImageMagick's `import` binary and fails with "unable to grab mouse", `from pathlib import Path` fails with "from: command not found", and the script dies on the first real assignment with a syntax error.

**Root cause**
`/usr/share/omarchy/bin/omarchy-hook` invokes every file in `<hook>.d/` unconditionally with `bash "$hook" "$@"`, regardless of the file's shebang or executable bit:

```bash
if [[ -d $HOOK_DIR ]]; then
  for hook in "$HOOK_DIR"/*; do
    [[ -f $hook ]] || continue
    [[ $hook == *.sample ]] && continue
    bash "$hook" "$@" || echo "Hook failed: $hook"
  done
fi
```

This silently discards the shebang, so any hook not written in bash (Python, in this case) is unusable even though it's installed exactly as `hooks.md` describes (executable, in the right `.d` directory).

**Expected**
An executable hook script with a valid shebang should be invoked directly (`"$hook" "$@"`), letting the kernel/shebang pick the interpreter — falling back to `bash "$hook"` only for non-executable files, matching how `omarchy hook install` already makes scripts executable.

**Steps to reproduce**
1. Create `~/.config/omarchy/hooks/theme-set.d/test-hook.py`:
   ```python
   #!/usr/bin/env python3
   print("hello from python")
   ```
2. `chmod +x ~/.config/omarchy/hooks/theme-set.d/test-hook.py`
3. `omarchy theme set <any-theme>` or `omarchy-hook theme-set <slug>`
4. Observe bash errors instead of `hello from python`.

This is not theme-specific — it reproduces on every theme switch, for any theme, since the runner never inspects the theme at all before dispatching the hook.

**System**
- Omarchy version: 4.0.2-1 (stable channel)
- Kernel: Linux 7.1.9-arch1-2 x86_64
- GPU: NVIDIA GeForce RTX 3060 (GA106)

**Note:** `omarchy debug --no-sudo --print` from the current docs doesn't exist on this installed version (`omarchy commands --all` has no `debug` group), so I couldn't attach the standard diagnostic log — worth checking if that command was renamed/removed since 4.0.2-1.

fresh3nough:
## Summary

`omarchy-hook` always ran `bash "$hook"`, ignoring shebang and the executable bit. Python (and other non-bash) hooks installed as documented failed as shell syntax on every theme-set and other events.

## Fix

- If the hook is executable, run it directly so the kernel honors the shebang
- Otherwise keep `bash "$hook"` for plain non-executable shell snippets

Fixes #9845

## Test plan

- [x] Reproduced: executable `#!/usr/bin/env python3` hook → `import: command not found`
- [x] `bash test/shell.d/hook-runner-test.sh` — python +x, bash fallback, flat path, failure reporting

### DIFF
--- a/bin/omarchy-hook
+++ b/bin/omarchy-hook
@@ -15,14 +15,27 @@ HOOK_PATH="$HOME/.config/omarchy/hooks/$1"
 HOOK_DIR="$HOOK_PATH.d"
 shift
 
+# Executable hooks keep their shebang (python, etc.). Non-executable files stay
+# on bash so plain shell snippets without +x still work (#9845).
+run_hook() {
+  local hook=$1
+  shift
+
+  if [[ -x $hook ]]; then
+    "$hook" "$@" || echo "Hook failed: $hook"
+  else
+    bash "$hook" "$@" || echo "Hook failed: $hook"
+  fi
+}
+
 if [[ -f $HOOK_PATH ]]; then
-  bash "$HOOK_PATH" "$@" || echo "Hook failed: $HOOK_PATH"
+  run_hook "$HOOK_PATH" "$@"
 fi
 
 if [[ -d $HOOK_DIR ]]; then
   for hook in "$HOOK_DIR"/*; do
     [[ -f $hook ]] || continue
     [[ $hook == *.sample ]] && continue
-    bash "$hook" "$@" || echo "Hook failed: $hook"
+    run_hook "$hook" "$@"
   done
 fi
--- a/test/shell.d/hook-runner-test.sh
+++ b/test/shell.d/hook-runner-test.sh
@@ -1,0 +1,64 @@
+#!/bin/bash
+
+set -euo pipefail
+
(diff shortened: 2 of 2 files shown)


paulogeyer:
Same fix as #8493 (verified locally too). Prefer merging #8493 and closing this as a duplicate — #8493 has slightly stronger shebang/flat-hook test coverage.

paulogeyer:
Closing as duplicate of #8493 (same shebang/+x behavior; prefer that PR).

bjarneo:
Automated duplication check: this pull request looks similar to #8493, which covers the same hook shebang exec. I keep that one open and close this one to consolidate review.

PR #9414, opened 2026-08-31
Clear setgid when hardening Windows VM storage directories

Chessing234:
## Summary
- `chmod 0700` on a setgid directory leaves mode `2700`, so the exact `700` privacy check failed and launch aborted.
- Use `chmod a-s,u=rwx,go=` when hardening storage/shared leaves.

## Test plan
- [ ] Launch Windows VM from a setgid-inherited storage path
- [ ] Existing windows-vm mount/compose privacy tests still pass

PR #9989, opened 2026-09-03
Clear setgid when hardening Windows VM mount modes

### LINKED ISSUE #9943: Windows VM: launch fails silently when the container leaves ~/Windows setgid (chmod 0700 cannot clear it)

fresh3nough:
## Summary

`prepare_caller_mounts` ran `chmod 0700` then required `stat -c %a` to equal `700`. The Windows container sets `~/Windows` to `2777`; GNU chmod keeps directory setgid, leaving `2700`, so launch failed with only `Failed to start Windows VM!` while status still said RUNNING. The failure self-renewed on every container start.

## Fix

- After `chmod 0700`, run `chmod g-s,o-t` on storage/shared FDs and user mount sources
- Print unexpected modes on refusal instead of a silent `return 1`

Fixes #9943

## Test plan

- [x] Reproduced: `2777` → `chmod 0700` → `2700`; `g-s,o-t` → `700`
- [x] `bash test/shell.d/windows-vm-setgid-test.sh`

### DIFF
--- a/bin/omarchy-windows-vm
+++ b/bin/omarchy-windows-vm
@@ -580,14 +580,18 @@ prepare_caller_mounts() {
   # Privacy is an explicit preflight step for both already-pinned sources, not
   # a side effect halfway through the two-mount transaction. Old umask-022
   # installs are hardened together before either Docker-facing anchor changes.
+  # dockurr/windows sets ~/Windows to 2777; GNU chmod 0700 keeps setgid (2700),
+  # so clear g-s explicitly or the literal 700 check fails forever (#9943).
   chmod 0700 -- "/proc/$BASHPID/fd/$storage_fd" "/proc/$BASHPID/fd/$shared_fd" || {
     exec {storage_fd}<&-
     exec {shared_fd}<&-
     return 1
   }
+  chmod g-s,o-t -- "/proc/$BASHPID/fd/$storage_fd" "/proc/$BASHPID/fd/$shared_fd" 2>/dev/null || true
   storage_mode=$(stat -Lc '%a' "/proc/$BASHPID/fd/$storage_fd" 2>/dev/null) || storage_mode=""
   shared_mode=$(stat -Lc '%a' "/proc/$BASHPID/fd/$shared_fd" 2>/dev/null) || shared_mode=""
   if [[ $storage_mode != 700 || $shared_mode != 700 ]]; then
+    echo "omarchy-windows-vm: refusing VM mount source with unexpected mode (storage=$storage_mode shared=$shared_mode; want 700)" >&2
     exec {storage_fd}<&-
     exec {shared_fd}<&-
     return 1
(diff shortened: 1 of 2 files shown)


spuder:
Possible duplicate of
- #9414 
- #9564
- #9605

